PR #6023, opened 2025-09-25
Add `device_ref` parameter to `get_device_address`

### LINKED ISSUE #6020: [BUG]: `cuda::get_device_address` relies on current device
### Is this a duplicate?

### Type of Bug

Something else

### Component

libcu++

### Describe the bug

`cuda::get_device_address` uses `cudaGetSymbolAddress` to retrieve the device address of an object. The problem is that it works with the current device, which is something we don't want in cccl-rt

### How to Reproduce

See https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__MEMORY.html#group__CUDART__MEMORY_1g4f513be54d3794667c2017146b3d6a2b

### Expected behavior

The function should take a `device_ref` parameter.

### Reproduction link

_No response_

### Operating System

_No response_

### nvidia-smi output

_No response_

### NVCC version

_No response_

davebayer:
Fixes #6020.

### DIFF
--- a/libcudacxx/include/cuda/__memory/get_device_address.h
+++ b/libcudacxx/include/cuda/__memory/get_device_address.h
@@ -23,7 +23,9 @@
 
 #if _CCCL_HAS_CTK()
 
+#  include <cuda/__device/device_ref.h>
 #  include <cuda/__runtime/api_wrapper.h>
+#  include <cuda/__runtime/ensure_current_context.h>
 #  include <cuda/std/__memory/addressof.h>
 
 #  include <nv/target>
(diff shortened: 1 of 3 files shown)


pciolkosz (approved):
I like the new overload, but lets confirm the deprecation with @miscco

I was thinking how we could make this usable in host device functions without using the current device, maybe we could introduce a placeholder device_ref getter in device code, something like:
`__device__ device_ref currently_executing_device() { return device_ref{-1}; }`

Then someone could do:
```
__host__ __device__ void my_host_dev_func(device_ref dev, ...) {
  // ...

  auto addr = get_device_address(var, dev);
}

// host call:
my_host_dev_func(device_ref{0}, ...);

// device call:
my_host_dev_func(currently_executing_device(), ...);
```

Then we would document that a call to `get_device_address` in device code is UB unless `device_ref` argument is this dummy and inside it we ignore the `device_ref`

davebayer:
I was also thinking about something similar, I had `cuda::device::this_device()` in mind. We could use `cudaGetDevice` to retrieve the device id in device code I believe. I would go for `__host__` only for now, we can always add it later :)

libcudacxx/include/cuda/__memory/get_device_address.h
```diff
@@ -36,19 +38,41 @@ _CCCL_BEGIN_NAMESPACE_CUDA
 //! @param __device_object the object residing in device memory
 //! @return Valid pointer to the device object
 template <class _Tp>
-[[nodiscard]] _CCCL_API inline _Tp* get_device_address(_Tp& __device_object)
+[[nodiscard]] CCCL_DEPRECATED _CCCL_API inline _Tp* get_device_address(_Tp& __device_object)
```
miscco: I am not to big of a fan of deprecating this. A lof of users are still in a single device stup and we are making life much harder for them. I would love to have an additional doc string at the beginning

```
//! @warn The user must ensure that the current device is properly set to the device the object was allocated on
```

davebayer: That's a fair point. I suppose we don't expect users to jump right away to the explicit device APIs that quickly. @pciolkosz  wouldn't it make sense to provide something like `cuda::current_device()` that would work both `__host__` and `__device__` code that would be a solution for projects that either want to keep using the old current device design or need something for the transition period?

pciolkosz: Yes, we discussed an API like that and it might be a good idea to provide it. We just need to be careful to provide it in a way that makes it clear this is more of a stopgap thing and we recommend the explicit device APIs instead

PR #6029, opened 2025-09-25
Provide escape hatch for CTK compatability check

miscco:
Some (internal) users might want to build a newer CTK with an older compiler. This is not officially suported but we can unbreak them by providing an escape hatch.

This is obviously an unsupported use case


davebayer (approved):

m3vaz:
Looks good, Thanks!

bernhardmgruber (approved):

PR #6032, opened 2025-09-25
Fix addressof shadowing issue with libc++

wmaxey:
## Description

We can't test libc++ yet entirely in CCCL. So I'm changing this from 'add libc++ testing' to 'fix specific shadowing issue when using libc++'.

`addressof(Ty&&)` is deleted by modern standards, we don't need to do that if we're importing the function into our namespace. 

For cases where `addressof` is defined by CCCL we should delete the overload.

I'll keep the infrastructure for firing up builds with libc++. There may be ways to bypass issues seen with the crt configs, but it would likely be fragile.

## Checklist


cmake/CCCLBuildCompilerTargets.cmake
```diff
@@ -100,6 +100,11 @@ function(cccl_build_compiler_targets)
     list(APPEND cxx_compile_definitions "CCCL_DISABLE_RTTI")
   endif()
 
+  #  if (CCCL_USE_LIBCXX)
+  #    list(APPEND cxx_compile_options "-stdlib=libc++")
+  #    list(APPEND cxx_compile_definitions "_ALLOW_UNSUPPORTED_LIBCPP=1")
+  #  endif()
+
```
miscco, line 107: Why is this commented? We should at least reference an issue

libcudacxx/include/cuda/std/__memory/addressof.h
```diff
@@ -42,19 +42,22 @@ template <class _Tp>
...
-[[nodiscard]] _CCCL_API inline _CCCL_NO_CFI _Tp* addressof(_Tp& __x) noexcept
+[[nodiscard]] _CCCL_API inline _CCCL_NO_CFI constexpr _Tp* addressof(_Tp& __x) noexcept
 {
   return reinterpret_cast<_Tp*>(const_cast<char*>(&reinterpret_cast<const volatile char&>(__x)));
 }
 
-#endif // defined(_CCCL_BUILTIN_ADDRESSOF)
-
 template <class _Tp>
 _Tp* addressof(const _Tp&&) noexcept = delete;
```
miscco, line 57: The delete is already there, please just move it out

libcudacxx/include/cuda/std/__memory/addressof.h
```diff
@@ -48,7 +48,7 @@ _Tp* addressof(const _Tp&&) noexcept = delete;
 #else
 
 template <class _Tp>
-[[nodiscard]] _CCCL_API inline _CCCL_NO_CFI _Tp* addressof(_Tp& __x) noexcept
+[[nodiscard]] _CCCL_API inline _CCCL_NO_CFI constexpr _Tp* addressof(_Tp& __x) noexcept
```
miscco: This cannot be constexpr

wmaxey: Ah oops, didn't mean to leave that. I was trying to resolve a different issue. 

I was encountering an error in some build where with NVCC not including device builtins there was no constexpr addressof.

libcudacxx/include/cuda/std/__memory/addressof.h
wmaxey: What do you mean already there?

wmaxey: The point is that it can't be shadowing the host libraries' version.

wmaxey:
/ok to test [1109973](https://github.com/NVIDIA/cccl/pull/6032/commits/110997380fa3ed25cc8f1593f3947fdd31d299a2)

cmake/CCCLBuildCompilerTargets.cmake
wmaxey: CUDA header bits fail to compile on x86 because all the libcxx paths are guarded by Android detection.

miscco: WTF?

PR #6034, opened 2025-09-26
Provide `cuda::in_range`

fbusato:
briefly discussed with @bernhardmgruber, @miscco, @davebayer. @bernhardmgruber pointed out that [is_clamped](http://wg21.link/p1440) has been rejected.
## Description

Checks if a value is within a range is a very common functionality. The common approach `x >= start && x < end` is error-prone due to implicit conversion, and triggers compiler warnings for the same reason. Also, the functionality can be optimized when is `x` is an unsigned type. Finally, we can verify the range validity.

```cpp
  template <typename T, typename R>
  __host__ __device__
  [[nodiscard]] constexpr bool in_range(T value, R start, R end) noexcept;
```

### DIFF
--- a/libcudacxx/test/libcudacxx/cuda/utility/in_range.pass.cpp
+++ b/libcudacxx/test/libcudacxx/cuda/utility/in_range.pass.cpp
@@ -39,36 +39,33 @@ __host__ __device__ constexpr void test()
   assert(!cuda::in_range(T{5}, cuda::std::numeric_limits<T>::min(), T{cuda::std::numeric_limits<T>::min() + T{1}}));
   assert(cuda::in_range(T{5}, cuda::std::numeric_limits<T>::min(), cuda::std::numeric_limits<T>::max()));
 
-  if constexpr (cuda::is_floating_point_v<T>)
+  if constexpr (cuda::std::numeric_limits<T>::has_infinity)
   {
-    if constexpr (cuda::std::numeric_limits<T>::has_infinity)
+    constexpr auto inf = cuda::std::numeric_limits<T>::infinity();
+    assert(cuda::in_range(inf, -inf, inf));
+    assert(cuda::in_range(-inf, -inf, inf));
+    assert(!cuda::in_range(inf, T{0}, T{10}));
+    assert(cuda::in_range(T{1}, T{-1}, inf));
+  }
+  if constexpr (cuda::std::numeric_limits<T>::has_quiet_NaN)
+  {
+    constexpr auto nan = cuda::std::numeric_limits<T>::quiet_NaN();
+    if (!cuda::std::__cccl_default_is_constant_evaluated())
     {
-      constexpr auto inf = cuda::std::numeric_limits<T>::infinity();
-      assert(cuda::in_range(inf, -inf, inf));
-      assert(cuda::in_range(-inf, -inf, inf));
-      assert(!cuda::in_range(inf, T{0}, T{10}));
-      assert(cuda::in_range(T{1}, T{-1}, inf));
+      assert(!cuda::in_range(nan, T{0}, T{10}));
+      assert(!cuda::in_range(T{1}, nan, T{10}));
+      assert(!cuda::in_range(T{1}, T{10}, nan));
     }
-    if constexpr (cuda::std::numeric_limits<T>::has_quiet_NaN)
+  }
+  if constexpr (cuda::std::numeric_limits<T>::has_quiet_NaN && cuda::std::numeric_limits<T>::has_infinity)
+  {
+    constexpr auto inf = cuda::std::numeric_limits<T>::infinity();
+    constexpr auto nan = cuda::std::numeric_limits<T>::quiet_NaN();
+    if (!cuda::std::__cccl_default_is_constant_evaluated())
     {
-      constexpr auto nan = cuda::std::numeric_limits<T>::quiet_NaN();
-      if (!cuda::std::__cccl_default_is_constant_evaluated())
-      {
-        assert(!cuda::in_range(nan, T{0}, T{10}));
-        assert(!cuda::in_range(T{1}, nan, T{10}));
-        assert(!cuda::in_range(T{1}, T{10}, nan));
-      }
-    }
-    if constexpr (cuda::std::numeric_limits<T>::has_quiet_NaN && cuda::std::numeric_limits<T>::has_infinity)
-    {
-      constexpr auto inf = cuda::std::numeric_limits<T>::infinity();
-      constexpr auto nan = cuda::std::numeric_limits<T>::quiet_NaN();
-      if (!cuda::std::__cccl_default_is_constant_evaluated())
-      {
-        assert(!cuda::in_range(nan, -inf, inf));
-        assert(!cuda::in_range(T{1}, nan, inf));
-        assert(!cuda::in_range(T{1}, inf, nan));
-      }
+      assert(!cuda::in_range(nan, -inf, inf));
+      assert(!cuda::in_range(T{1}, nan, inf));
+      assert(!cuda::in_range(T{1}, inf, nan));
     }
   }
 }


docs/libcudacxx/extended_api/utility/in_range.rst
```diff
@@ -0,0 +1,79 @@
...
+
+Defined in the ``<cuda/utility>`` header.
+
+.. code:: cuda
+
+    namespace cuda {
+
+    template <typename T, typename R>
+    __host__ __device__
+    [[nodiscard]] constexpr bool in_range(T value, R start, R end) noexcept;
```
davebayer:
```suggestion
    [[nodiscard]] __host__ __device__ constexpr 
    bool in_range(T value, R start, R end) noexcept;
```
Let's be consistent

docs/libcudacxx/extended_api/utility/in_range.rst
```diff
@@ -0,0 +1,79 @@
...
+
+    } // namespace cuda
+
+Checks whether a value ``value`` is within the range ``[start, end)``.
```
davebayer: Shouldn't the end be included to in the range? Because if not, you have no native way to specify range `[0, ULONGLONG_MAX]`. I think it should be `[start, end]`

davebayer: [`std::clamp`](https://en.cppreference.com/w/cpp/algorithm/clamp.html) also does this, so let's stay consistent

docs/libcudacxx/extended_api/utility/in_range.rst
```diff
@@ -0,0 +1,79 @@
...
+
+Example
+-------
+
+.. code:: cuda
+
+    #include <cuda/utility>
+    #include <cuda/std/cassert>
+
+    __global__ void kernel() {
```
davebayer:
```suggestion
    __global__ void in_range_kernel() {
```

libcudacxx/include/cuda/__utility/in_range.h
```diff
@@ -0,0 +1,62 @@
...
+    // if __end > __start, we know that the range is always positive. Similarly, __v is positive if unsigned.
+    // this optimization is useful when __start and __end are compile-time constants, or when in_range is used multiple
+    // times with the same range
+    using _CommonType         = ::cuda::std::common_type_t<_Tp, _Up, unsigned>; // at least 32-bit
+    using _UnsignedCommonType = ::cuda::std::make_unsigned_t<_CommonType>;
+    const auto __start1       = static_cast<_UnsignedCommonType>(__start);
+    const auto __end1         = static_cast<_UnsignedCommonType>(__end);
+    const auto __v1           = static_cast<_UnsignedCommonType>(__v);
+    const auto __range        = __end1 - __start1;
+    return (__v1 - __start1) < __range;
```
davebayer: Are you sure this is correct? I'm thinking about cases when `_Up` is signed, will it behave correctly?

libcudacxx/include/cuda/__utility/in_range.h
```diff
@@ -0,0 +1,62 @@
...
+#include <cuda/std/__type_traits/make_unsigned.h>
+#include <cuda/std/__utility/cmp.h>
+
+#include <cuda/std/__cccl/prologue.h>
+
+_CCCL_BEGIN_NAMESPACE_CUDA
+
+_CCCL_TEMPLATE(typename _Tp, typename _Up)
+_CCCL_REQUIRES(::cuda::std::__cccl_is_integer_v<_Tp> _CCCL_AND ::cuda::std::__cccl_is_integer_v<_Up>)
+[[nodiscard]] _CCCL_API constexpr bool in_range(_Tp __v, _Up __start, _Up __end) noexcept
```
miscco: I am not a fan of the different template arguments. Especially integer promotion could do wonders here.

miscco (changes requested):

docs/libcudacxx/extended_api/utility/in_range.rst
fbusato: ah ok, so you are suggesting an inclusive range similarly of [P1440](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1440r0.html).
I'm not sure about that. I'm thinking about checking if an index falls in the right range. I would write `x < size`

libcudacxx/include/cuda/__utility/in_range.h
fbusato: I'm not against it, but one of the benefits of the function is indeed to avoid implicit promotion. I'm not understanding the problem from the user side

libcudacxx/include/cuda/__utility/in_range.h
fbusato: Correct me if I'm wrong:

- `__v1` and `__range` are always positive
- if `_Up` is signed

   -  `__start` is positive, the function behaves in the same way as unsigned
   - if `_Up` is signed and `__start` is negative, `__v1 - __start1` represents the 'distance' between them. If `__v1` is closer to `__start` than `__end`, it is in the range, `false` otherwise.

libcudacxx/include/cuda/__utility/in_range.h
miscco: The problem is whether a user can oversee whether there is integer promotion happening. 

The question I have is whether there is a benefit to having different types here, we *alway* want to compare the same range, so different types provide no benefit.

On the other hand they add a lot of complexity and compile time cost

fbusato: that's a good motivation. I will go ahead with your suggestion

docs/libcudacxx/extended_api/utility/in_range.rst
```diff
@@ -0,0 +1,72 @@
...
+
+- ``true`` if ``value`` is in the range ``[start, end]``, ``false`` otherwise.
+
+**Constraints**
+
+- ``T`` must be an integer type.
+
+**Preconditions**
+
+- ``end`` must be greater than ``start``.
```
davebayer:
```suggestion
- ``end`` must be greater than or equal to ``start``.
```
`[4, 4]` is still a valid range

libcudacxx/include/cuda/__utility/in_range.h
```diff
@@ -0,0 +1,60 @@
...
+
+#include <cuda/std/__cccl/prologue.h>
+
+_CCCL_BEGIN_NAMESPACE_CUDA
+
+_CCCL_TEMPLATE(typename _Tp)
+_CCCL_REQUIRES(::cuda::std::__cccl_is_integer_v<_Tp>)
+[[nodiscard]] _CCCL_API constexpr bool in_range(_Tp __v, _Tp __start, _Tp __end) noexcept
+{
+  _CCCL_ASSERT(__end > __start, "in_range: __end must be greater than __start");
```
davebayer:
```suggestion
  _CCCL_ASSERT(__end >= __start, "in_range: __end must be greater than or equal to __start");
```

davebayer (changes requested):

fbusato:
what about set the default values of `start` and `end` with `numeric_limits<T>`?

davebayer:
> what about set the default values of `start` and `end` with `numeric_limits<T>`?

No, it will lead to cases like:
1. `in_range(value, 0)`, which is not obvious what does it actually check
2. `in_range(value)` which would always result in `true`
3. and the worst one - `in_range<unsigned>(value)` which would cast the value to `unsigned` and then returned `true`

libcudacxx/include/cuda/__utility/in_range.h
```diff
@@ -0,0 +1,60 @@
...
+
+#include <cuda/std/__cccl/prologue.h>
+
+_CCCL_BEGIN_NAMESPACE_CUDA
+
+_CCCL_TEMPLATE(typename _Tp)
+_CCCL_REQUIRES(::cuda::std::__cccl_is_integer_v<_Tp>)
+[[nodiscard]] _CCCL_API constexpr bool in_range(_Tp __v, _Tp __start, _Tp __end) noexcept
+{
+  _CCCL_ASSERT(__end >= __start, "in_range: __end must be greater than or equal to __start");
```
bernhardmgruber: Important: From the documentation, it seems that `in_range(x, 13, 13)` should be valid (the range has a single element), but it would trigger the assertion here. I think we should allow single item ranges.

libcudacxx/include/cuda/__utility/in_range.h
```diff
@@ -0,0 +1,60 @@
...
+#include <cuda/std/__type_traits/is_integer.h>
+#include <cuda/std/__type_traits/is_unsigned.h>
+
+#include <cuda/std/__cccl/prologue.h>
+
+_CCCL_BEGIN_NAMESPACE_CUDA
+
+_CCCL_TEMPLATE(typename _Tp)
+_CCCL_REQUIRES(::cuda::std::__cccl_is_integer_v<_Tp>)
+[[nodiscard]] _CCCL_API constexpr bool in_range(_Tp __v, _Tp __start, _Tp __end) noexcept
```
bernhardmgruber: Q: Why can't we support floating point types in this function? The signed integer path looks like it could work for FP types.

fbusato: great point. we can extend it

libcudacxx/include/cuda/__utility/in_range.h
fbusato: `__end >= __start` follows the documentation where we say 
```
Checks whether a value ``value`` is within the range ``[start, end]``, inclusive.
```

fbusato:
> Does the standard define it like that or is it our choice?

not really, more our choice. The closest match is `std::clamp` which uses the inclusive range

libcudacxx/include/cuda/__utility/in_range.h
```diff
@@ -0,0 +1,60 @@
...
+  {
+    // if __end > __start, we know that the range is always positive. Similarly, __v is positive if unsigned.
+    // this optimization is useful when __start and __end are compile-time constants, or when in_range is used multiple
+    // times with the same range
+    using _Up = ::cuda::std::conditional_t<(sizeof(_Tp) <= sizeof(unsigned)), unsigned, _Tp>; // at least 32-bit
+    const auto __start1 = static_cast<_Up>(__start);
+    const auto __end1   = static_cast<_Up>(__end);
+    const auto __v1     = static_cast<_Up>(__v);
+    const auto __range  = __end1 - __start1;
+    return (__v1 - __start1) <= __range;
```
davebayer, line 53: Is this something we want to do for floating point types? There will be rounding errors. This branch should be unsigned integer only

libcudacxx/test/libcudacxx/cuda/utility/in_range.pass.cpp
```diff
@@ -0,0 +1,68 @@
...
+  test<int>();
+  test<unsigned long>();
+  test<long>();
+  test<unsigned long long>();
+  test<long long>();
+#if _CCCL_HAS_INT128()
+  test<__int128_t>();
+  test<__uint128_t>();
+#endif // _CCCL_HAS_INT128()
+  return true;
```
davebayer, line 105: I am missing floating point types tests

libcudacxx/include/cuda/__utility/in_range.h
```diff
@@ -0,0 +1,60 @@
...
+#include <cuda/std/__type_traits/conditional.h>
+#include <cuda/std/__type_traits/is_arithmetic.h>
+#include <cuda/std/__type_traits/is_unsigned.h>
+
+#include <cuda/std/__cccl/prologue.h>
+
+_CCCL_BEGIN_NAMESPACE_CUDA
+
+_CCCL_TEMPLATE(typename _Tp)
+_CCCL_REQUIRES(::cuda::std::is_arithmetic_v<_Tp>)
```
davebayer: Don't we want to `= delete` the `bool` overload?

libcudacxx/include/cuda/__utility/in_range.h
fbusato: classic floating-point types are signed only but yes, we can make it more precise

libcudacxx/include/cuda/__utility/in_range.h
fbusato: let also exclude `char`

libcudacxx/include/cuda/__utility/in_range.h
```diff
@@ -0,0 +1,61 @@
...
+
+#include <cuda/std/__cccl/prologue.h>
+
+_CCCL_BEGIN_NAMESPACE_CUDA
+
+_CCCL_TEMPLATE(typename _Tp)
+_CCCL_REQUIRES(::cuda::std::__cccl_is_integer_v<_Tp> || ::cuda::is_floating_point_v<_Tp>)
+[[nodiscard]] _CCCL_API constexpr bool in_range(_Tp __v, _Tp __start, _Tp __end) noexcept
+{
+  _CCCL_ASSERT(__end >= __start, "in_range: __end must be greater than or equal to __start");
```
davebayer: What about `NaN`s? Can they make a valid range?

libcudacxx/test/libcudacxx/cuda/utility/in_range.pass.cpp
```diff
@@ -0,0 +1,86 @@
...
+  {
+    assert(!cuda::in_range(T{-5}, T{0}, T{10}));
+    assert(cuda::in_range(T{5}, T{-10}, T{10}));
+    assert(cuda::in_range(T{5}, T{-1}, T{10}));
+    assert(cuda::in_range(T{0}, T{-1}, T{1}));
+  }
+  assert(!cuda::in_range(T{5}, T{cuda::std::numeric_limits<T>::max() - T{1}}, cuda::std::numeric_limits<T>::max()));
+  assert(!cuda::in_range(T{5}, cuda::std::numeric_limits<T>::min(), T{cuda::std::numeric_limits<T>::min() + T{1}}));
+  assert(cuda::in_range(T{5}, cuda::std::numeric_limits<T>::min(), cuda::std::numeric_limits<T>::max()));
+}
```
davebayer, line 71: We should check NaNs and infinities for fp types I believe

libcudacxx/include/cuda/__utility/in_range.h
fbusato: NaN comparison is always false, so that's fine for the contract of the function

libcudacxx/test/libcudacxx/cuda/utility/in_range.pass.cpp
fbusato: Added, `NaN` cannot be used in the range bounds because the assertion directly fails (as expected)

docs/libcudacxx/extended_api/utility/in_range.rst
```diff
@@ -0,0 +1,73 @@
...
+
+- ``true`` if ``value`` is in the range ``[start, end]``, ``false`` otherwise.
+
+**Constraints**
+
+- ``T`` must be an integer or floating point type.
+
+**Preconditions**
+
+- ``end`` must be greater than or equal to ``start``.
```
davebayer, line 40: Maybe we can document be behaviour of fp types for `NaN` and other values?

libcudacxx/test/libcudacxx/cuda/utility/in_range.pass.cpp
```diff
@@ -0,0 +1,102 @@
...
+      assert(!cuda::in_range(nan, -inf, inf));
+      assert(!cuda::in_range(T{1}, nan, inf));
+      assert(!cuda::in_range(T{1}, inf, nan));
+    }
+  }
+}
+
+__host__ __device__ constexpr bool test()
+{
+  static_assert(noexcept(cuda::in_range(1, 0, 10)));
```
davebayer: nitpick: This should be checked for each type individually

libcudacxx/test/libcudacxx/cuda/utility/in_range.pass.cpp
```diff
@@ -0,0 +1,102 @@
...
+    assert(!cuda::in_range(T{-5}, T{0}, T{10}));
+    assert(cuda::in_range(T{5}, T{-10}, T{10}));
+    assert(cuda::in_range(T{5}, T{-1}, T{10}));
+    assert(cuda::in_range(T{0}, T{-1}, T{1}));
+  }
+  assert(!cuda::in_range(T{5}, T{cuda::std::numeric_limits<T>::max() - T{1}}, cuda::std::numeric_limits<T>::max()));
+  assert(!cuda::in_range(T{5}, cuda::std::numeric_limits<T>::min(), T{cuda::std::numeric_limits<T>::min() + T{1}}));
+  assert(cuda::in_range(T{5}, cuda::std::numeric_limits<T>::min(), cuda::std::numeric_limits<T>::max()));
+
+  if constexpr (cuda::std::is_floating_point_v<T>)
```
davebayer: This doesn't test the extended floating point types

libcudacxx/test/libcudacxx/cuda/utility/in_range.pass.cpp
```diff
@@ -0,0 +1,102 @@
...
+  {
+    constexpr auto nan = cuda::std::numeric_limits<T>::quiet_NaN();
+    constexpr auto inf = cuda::std::numeric_limits<T>::infinity();
```
davebayer: could we guard these tests by `numeric_limits::has_[infinity,quiet_NaN]`? It will save us some time in the future

davebayer (changes requested):

libcudacxx/test/libcudacxx/cuda/utility/in_range.pass.cpp
fbusato: they are in the `constexpr` test function

libcudacxx/test/libcudacxx/cuda/utility/in_range.pass.cpp
```diff
@@ -15,14 +15,18 @@
...
 #include <cuda/utility>
 
 #include "test_macros.h"
 
 template <typename T>
 __host__ __device__ constexpr void test()
 {
+  if constexpr (!cuda::std::is_same_v<T, __half> && !cuda::std::is_same_v<T, __nv_bfloat16>)
+  {
+    static_assert(noexcept(cuda::in_range(T{1}, T{0}, T{10})));
```
davebayer:
```suggestion
    static_assert(noexcept(cuda::in_range(cuda::std::declval<T>(), cuda::std::declval<T>(), cuda::std::declval<T>())));
```

libcudacxx/test/libcudacxx/cuda/utility/in_range.pass.cpp
davebayer: I meant separate inf/nan completely - we have types that have only nan for example

libcudacxx/test/libcudacxx/cuda/utility/in_range.pass.cpp
davebayer: And we can get rid of the `if constexpr`

davebayer (approved):

miscco (approved):

PR #6036, opened 2025-09-26
Improve `zip_iterator`

miscco:
We were experiencing siutations where nvc++ had issues recognizing that it would need to generate a device function for a class static function

Futhermore, we found that `zip_iterator::operator-` would not work if one of the base iterators was not default constructible. 

That seems like a tough constraint, so also allow any random_access_traversal because that need operator- to work

### DIFF
--- a/libcudacxx/include/cuda/__iterator/zip_iterator.h
+++ b/libcudacxx/include/cuda/__iterator/zip_iterator.h
@@ -128,6 +128,55 @@ _CCCL_API constexpr auto __get_zip_view_iterator_tag()
   _CCCL_UNREACHABLE();
 }
 
+//! @note Not static functions because nvc++ sometimes has issues with class static functions in device code
+struct __zip_op_star
+{
+  template <class... _Iterators>
+  using reference = __tuple_or_pair<::cuda::std::iter_reference_t<_Iterators>...>;
+
+  _CCCL_EXEC_CHECK_DISABLE
+  template <class... _Iterators>
+  [[nodiscard]] _CCCL_API constexpr reference<_Iterators...> operator()(const _Iterators&... __iters) const
+    noexcept(noexcept(reference<_Iterators...>{*__iters...}))
+  {
+    return reference<_Iterators...>{*__iters...};
+  }
+};
+
+struct __zip_op_increment
+{
+  _CCCL_EXEC_CHECK_DISABLE
+  template <class... _Iterators>
+  _CCCL_API constexpr void operator()(_Iterators&... __iters) const noexcept(noexcept(((void) ++__iters, ...)))
+  {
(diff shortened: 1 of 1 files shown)


libcudacxx/include/cuda/__iterator/zip_iterator.h
```diff
@@ -126,6 +128,55 @@ _CCCL_API constexpr auto __get_zip_view_iterator_tag()
   _CCCL_UNREACHABLE();
 }
 
+//! @note Not static functions because nvc++ sometimes has issues with class static functions in device code
+struct __zip_op_star
```
miscco, line 132: Note to reviewers, I pulled the structs out because otherwise we would have to instantiate a type for each of them for every `zip_iterator` instantiation, whereas this is effectively free because its just exactly one per function and only when used

libcudacxx/include/cuda/__iterator/zip_iterator.h
```diff
@@ -126,6 +128,55 @@ _CCCL_API constexpr auto __get_zip_view_iterator_tag()
...
+  _CCCL_EXEC_CHECK_DISABLE
+  template <class... _Iterators>
+  [[nodiscard]] _CCCL_API constexpr reference<_Iterators...> operator()(const _Iterators&... __iters) const
+    noexcept(noexcept(reference<_Iterators...>{*__iters...}))
+  {
+    return reference<_Iterators...>{*__iters...};
+  }
+};
+
+struct __zip_op_increment
```
davebayer, line 146: Are the structs necessary? Can't we have just free functions?

miscco: The original bug was passing function pointers to `apply`, so yes we need structs

davebayer: My brain isn't working yet

bernhardmgruber (approved):

libcudacxx/test/libcudacxx/cuda/iterators/zip_iterator/arithmetic.pass.cpp
```diff
@@ -91,6 +91,16 @@ __host__ __device__ constexpr bool test()
     static_assert(cuda::std::is_same_v<decltype(iter2 - iter1), cuda::std::iter_difference_t<Iter>>);
   }
 
+  { // One of the iterators is not default constructible but sized
+    cuda::zip_iterator iter1{IterNotDefaultConstructibleSized{0}, b};
+    cuda::zip_iterator iter2{IterNotDefaultConstructibleSized{5}, b + 5};
+    using Iter = decltype(iter1);
```
bernhardmgruber: Not used:
```suggestion
```

PR #6037, opened 2025-09-26
Move `for_each_canceled_block` to `cuda::device::`

davebayer:
It is a device only function, so it should be placed in `cuda::device::` namespace.

### DIFF
--- a/libcudacxx/include/cuda/__functional/for_each_canceled.h
+++ b/libcudacxx/include/cuda/__functional/for_each_canceled.h
@@ -147,14 +147,14 @@ _CCCL_DEVICE_API void __for_each_canceled_block_sm100(::dim3 __block_idx, bool _
     }
 
     // Read new thread block dimensions
-    ::dim3 __b(_CUDA_DEVICE::__cluster_get_dim<0>(__result), 1, 1);
+    ::dim3 __b(::cuda::device::__cluster_get_dim<0>(__result), 1, 1);
     if constexpr (__ThreadBlockRank >= 2)
     {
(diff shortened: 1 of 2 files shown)


libcudacxx/include/cuda/__functional/for_each_canceled.h
```diff
@@ -148,14 +147,14 @@ __for_each_canceled_block_sm100(dim3 __block_idx, bool __is_leader, __UnaryFunct
     }
 
     // Read new thread block dimensions
-    dim3 __b(::cuda::__cluster_get_dim<0>(__result), 1, 1);
+    ::dim3 __b(_CUDA_DEVICE::__cluster_get_dim<0>(__result), 1, 1);
```
bernhardmgruber: Q: Why can't we use?
```suggestion
    ::dim3 __b(::cuda::device::__cluster_get_dim<0>(__result), 1, 1);
```

davebayer: Aaah, sorry, I thought we use this for `cuda::device::`..

bernhardmgruber: We decided a couple weeks ago that using the namespace directly is less confusing for people. And not including the inline ABI namespace should be fine.

davebayer: Yeah I know, I just thought we still use `_CUDA_DEVICE` for some reason

bernhardmgruber (approved):

PR #6039, opened 2025-09-26
Unwrap cuda::zip_iterator/zip_function in thrust::transform

bernhardmgruber:
This allows `cub::DeviceTransform` to optimize better.

### DIFF
--- a/libcudacxx/include/cuda/__iterator/zip_iterator.h
+++ b/libcudacxx/include/cuda/__iterator/zip_iterator.h
@@ -593,6 +593,16 @@ public:
   {
     return __zip_apply(__zip_op_iter_swap{}, __lhs.__current_, __rhs.__current_);
   }
+
+  [[nodiscard]] _CCCL_API constexpr __tuple_or_pair<_Iterators...>& __iterators() noexcept
+  {
+    return __current_;
+  }
+
+  [[nodiscard]] _CCCL_API constexpr const __tuple_or_pair<_Iterators...>& __iterators() const noexcept
+  {
+    return __current_;
+  }
 };
 
 template <class... _Iterators>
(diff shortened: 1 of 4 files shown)


thrust/thrust/system/cuda/detail/transform.h
```diff
@@ -202,6 +205,17 @@ OutputIt _CCCL_API _CCCL_FORCEINLINE cub_transform_many(
     policy, get<0>(firsts).get_iterator_tuple(), result, num_items, transform_op.underlying_function());
 }
 
+template <class Derived, class Offset, class... InputIts, class OutputIt, class TransformOp>
+OutputIt _CCCL_API _CCCL_FORCEINLINE cub_transform_many(
+  execution_policy<Derived>& policy,
+  ::cuda::std::tuple<::cuda::zip_iterator<InputIts...>> firsts,
+  OutputIt result,
+  Offset num_items,
+  ::cuda::zip_function<TransformOp> transform_op)
```
miscco, line 214: I believe we should reuse the `__is_zip_iterator` and `__is_zip_function `traits that are defined in `__fwd/zip_iterator.h`

That would allow us to also catch accidental issues like mixing thrust and cuda facilities

libcudacxx/include/cuda/__iterator/zip_iterator.h
```diff
@@ -593,6 +593,16 @@ class zip_iterator : public __zv_iter_category_base<_Iterators...>
   {
     return __zip_apply(__zip_op_iter_swap{}, __lhs.__current_, __rhs.__current_);
   }
+
+  [[nodiscard]] _CCCL_API constexpr __tuple_or_pair<_Iterators...>& __iterators() noexcept
```
miscco, line 597: This will require an implicit conversion from `const pair<T, U>&` to `tuple<T, U>`

I would say it is much safer to explicitly return a tuple here

miscco (approved):

PR #6040, opened 2025-09-26
Replace `__all` with fold expression

davebayer:
This PR replaces `__all` trait with fold expression if possible. I've kept the `__all_v` trait for cases when `__all` was used inside `noexcept`, because compilers tend to have problems with fold expressions inside `noexcept`.

### DIFF
--- a/libcudacxx/include/cuda/std/__type_traits/all.h
+++ b/libcudacxx/include/cuda/std/__type_traits/all.h
@@ -1,0 +1,35 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of libcu++, the C++ Standard Library for your entire system,
+// under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef _CUDA_STD___TYPE_TRAITS_ALL_H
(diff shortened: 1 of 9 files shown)


libcudacxx/include/cuda/std/detail/libcxx/include/tuple
```diff
@@ -944,9 +944,6 @@ public:
       : __base_(allocator_arg_t(), __a, ::cuda::std::forward<_Tuple>(__t))
   {}
 
-  using _CanCopyAssign = __all<is_copy_assignable_v<_Tp>...>;
-  using _CanMoveAssign = __all<is_move_assignable_v<_Tp>...>;
```
davebayer: These were unused

miscco (changes requested):
Unfortunately this is not easily possible

there are a bunch of compilers that cannot handle fold expressions in template headers. or need to explicitly put the result of the fold into a struct at least

libcudacxx/include/cuda/std/__type_traits/all.h
```diff
@@ -0,0 +1,35 @@
...
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_MSVC)
+#  pragma system_header
+#endif // no system header
+
+#include <cuda/std/__cccl/prologue.h>
+
+_CCCL_BEGIN_NAMESPACE_CUDA_STD
+
+template <bool... _Preds>
+inline constexpr bool __all_v = (_Preds && ...);
```
miscco: I believe what we actually should do to avoid the "fold expression in template headers" is to do
```suggestion
inline constexpr bool __all_v = bool_constant<(_Preds && ...)>::value;
```

This should work without insantiating any more types, because there is already `true_type`

fbusato:
I still think that having some documentation for internal traits would be valuable...

PR #6041, opened 2025-09-26
Fix nullptr dereference in vector_base::fill_insert

### LINKED ISSUE #2964: [BUG] Thrust `vector_base::fill_insert` does not consider empty initial state
We are unconditionally copying from `m_storage` even if there is no allocation leading to a nullptr dereference later on.

We should check whether there is an allocation and if not just copy input over

charan-003:
[BUG FIX] Fix nullptr dereference in vector_base::fill_insert for empty vectors

## Summary
This PR fixes a nullptr dereference that occurs when calling [fill_insert](cci:1://file:///home/scharan/mine/cccl/thrust/thrust/detail/vector_base.inl:878:0-1028:1) on an empty `thrust::vector` or `thrust::device_vector`. The issue was that the function would attempt to access [begin()](cci:1://file:///home/scharan/mine/cccl/thrust/thrust/detail/vector_base.inl:444:0-448:1), [end()](cci:1://file:///home/scharan/mine/cccl/thrust/thrust/detail/vector_base.inl:480:0-486:1), or `position` on unallocated storage.

## Changes
1. Added a special case for empty vectors that:
   - Handles allocation if needed
   - Uses `uninitialized_fill_n` directly to avoid iterator operations on unallocated storage

Fixes #2964


charan-003:
Hi @miscco ,

Thank you for your guidance on this issue! I assume that the suggested approach of filling the entire vector with x would change the intended behavior of 
fill_insert from `"inserting at position"` to `"resizing and filling`.

thrust/thrust/detail/vector_base.inl
```diff
@@ -879,105 +879,155 @@ void vector_base<T, Alloc>::append(size_type n)
 template <typename T, typename Alloc>
 void vector_base<T, Alloc>::fill_insert(iterator position, size_type n, const T& x)
 {
-  if (n != 0)
+  // Early return if nothing to insert
+  if (n == 0)
   {
-    if (capacity() - size() >= n)
+    return;
```
miscco, line 885: Sorry, I was not fully clear on the early return. We do need to clear the existing elements, so that the vector has a new size of 0

thrust/thrust/detail/vector_base.inl
```diff
@@ -879,105 +879,155 @@ void vector_base<T, Alloc>::append(size_type n)
...
+  }
+
+  // Handle the simple case: empty vector
+  if (size() == 0)
```
miscco: This is a rare case that should not need a special case, I believe we should only check for `capacity < n` and then handle all other cases after that

charan-003: Done! Now clears existing elements and sets size to 0 when n == 0  
updated the code to destroy all elements before returning.

charan-003: Hii @miscco 
 
Done! Now clears existing elements and sets size to 0 when n == 0
updated the code to destroy all elements before returning.

charan-003:
@bernhardmgruber @miscco 
Just wondering if there are any changes needed or anything else I should do for this PR?

bernhardmgruber:
I will let @miscco review this PR next week.

charan-003:
Hi @miscco 
Please let me know if any changes are needed for this PR.

thrust/thrust/detail/vector_base.inl
```diff
@@ -879,105 +879,110 @@ void vector_base<T, Alloc>::append(size_type n)
...
+    // Clear existing elements so the vector has size 0
+    if (size() > 0)
     {
-      // we've got room for all of them
-      // how many existing elements will we displace?
-      const size_type num_displaced_elements = end() - position;
-      iterator old_end                       = end();
+      m_storage.destroy(begin(), end());
+    }
+    m_size = 0;
```
miscco: Nitpick: 
```suggestion
    if (size() > 0)
    {
       m_storage.destroy(begin(), end());
       m_size = 0;
     }
```

thrust/thrust/detail/vector_base.inl
```diff
@@ -879,105 +879,110 @@ void vector_base<T, Alloc>::append(size_type n)
...
 
-        // finally, fill the range to the insertion point
-        thrust::fill_n(position, n, x);
-      } // end if
-      else
-      {
-        // construct new elements at the end of the vector
-        m_storage.uninitialized_fill_n(end(), n - num_displaced_elements, x);
+    // do not exceed maximum storage
+    new_capacity = ::cuda::std::min<size_type>(new_capacity, max_size());
```
miscco: I believe we should rather use `::cuda::std::clamp`
```suggestion
    // Ensure allocation grows exponentially wiithin bounds
    new_capacity = ::cuda::std::clamp<size_type>(new_capacity, static_cast<size_type>(2 * capacity()), max_size());
```

thrust/thrust/detail/vector_base.inl
```diff
@@ -879,105 +879,110 @@ void vector_base<T, Alloc>::append(size_type n)
...
 
-        // extend the size
-        m_size += n - num_displaced_elements;
+    if (new_capacity > max_size())
+    {
+      throw std::length_error("insert(): insertion exceeds max_size().");
+    } // end if
```
miscco: Important: This cannot happen due to the clamp above

thrust/thrust/detail/vector_base.inl
```diff
@@ -879,105 +879,110 @@ void vector_base<T, Alloc>::append(size_type n)
...
-      if (new_capacity > max_size())
-      {
-        throw std::length_error("insert(): insertion exceeds max_size().");
-      } // end if
+    // call destructors on the elements in the old storage
+    m_storage.destroy(begin(), end());
 
-      storage_type new_storage(copy_allocator_t(), m_storage, new_capacity);
+    // record the vector's new state
+    m_storage.swap(new_storage);
```
miscco: Nitpick: we should first swap then destroy

thrust/thrust/detail/vector_base.inl
```diff
@@ -879,105 +879,110 @@ void vector_base<T, Alloc>::append(size_type n)
...
+    m_size = old_size + n;
+  }
+  else
+  {
```
miscco: Nitpick: I would prefer if we would return here and reduce indentation afterwards

```suggestion
    m_size = old_size + n;
    return;
  }

```

thrust/thrust/detail/vector_base.inl
```diff
@@ -879,105 +879,110 @@ void vector_base<T, Alloc>::append(size_type n)
...
-        // remember [position, end()) refers to the old storage
-        new_end = m_storage.uninitialized_copy(position, end(), new_end);
-      } // end try
-      catch (...)
-      {
-        // something went wrong, so destroy & deallocate the new storage
-        m_storage.destroy(new_storage.begin(), new_end);
-        new_storage.deallocate();
+      // finally, fill the range to the insertion point
+      thrust::fill_n(position, n, x);
```
miscco: Critical: Missing include

miscco (changes requested):

thrust/thrust/detail/vector_base.inl
charan-003: Thank you for the comments

charan-003:
@bernhardmgruber, I hope the failed cases are unrelated to the changes I made

bernhardmgruber:
> @bernhardmgruber, I hope the failed cases are unrelated to the changes I made

Looks like a network issue. I restarted the job.

charan-003:
@bernhardmgruber, can you please take a look at this when possible?
Thanks!

bernhardmgruber:
> @bernhardmgruber, can you please take a look at this when possible? Thanks!

@miscco can you have a look please. You already digged into this :)

thrust/thrust/detail/vector_base.inl
```diff
@@ -878,105 +880,100 @@ void vector_base<T, Alloc>::append(size_type n)
 template <typename T, typename Alloc>
 void vector_base<T, Alloc>::fill_insert(iterator position, size_type n, const T& x)
 {
-  if (n != 0)
+  if (n == 0)
   {
-    if (capacity() - size() >= n)
+    if (size() > 0)
```
miscco: I believe we should just access `m_size` directly here given that we write it below
```suggestion
    if (m_size > 0)
```

thrust/thrust/detail/vector_base.inl
```diff
@@ -878,105 +880,100 @@ void vector_base<T, Alloc>::append(size_type n)
...
+    }
+    return;
+  }
 
-      if (num_displaced_elements > n)
-      {
-        // construct copy n displaced elements to new elements
-        // following the insertion
-        m_storage.uninitialized_copy(end() - n, end(), end());
+  if (capacity() - size() < n)
```
miscco: I would prefer if we moved the "simple" case of enough space here with
```suggestion
  if (n < capacity()) 
  {
    ...
  }
```

thrust/thrust/detail/vector_base.inl
```diff
@@ -878,105 +880,100 @@ void vector_base<T, Alloc>::append(size_type n)
...
-        m_size += n;
+    // compute the new capacity after the allocation
+    size_type new_capacity = old_size + ::cuda::std::max THRUST_PREVENT_MACRO_SUBSTITUTION(old_size, n);
 
-        // copy num_displaced_elements - n elements to existing elements
-        // this copy overlaps
-        const size_type copy_length = (old_end - n) - position;
-        thrust::detail::overlapped_copy(position, old_end - n, old_end - copy_length);
+    // Ensure allocation grows exponentially within bounds
+    new_capacity = ::cuda::std::clamp<size_type>(new_capacity, static_cast<size_type>(2 * capacity()), max_size());
```
miscco: This is unnecessary, if we are here, we already know that `size() < n`, so `new_capacity == n` and we can just use that instead


```suggestion
    // Ensure allocation grows exponentially within bounds
    const size_type new_capacity = ::cuda::std::clamp(n, static_cast<size_type>(2 * capacity()), max_size());
```

thrust/thrust/detail/vector_base.inl
```diff
@@ -878,105 +880,100 @@ void vector_base<T, Alloc>::append(size_type n)
...
-      const size_type old_size = size();
+      // something went wrong, so destroy & deallocate the new storage
+      new_storage.destroy(new_storage.begin(), new_end);
+      new_storage.deallocate();
 
-      // compute the new capacity after the allocation
-      size_type new_capacity = old_size + ::cuda::std::max THRUST_PREVENT_MACRO_SUBSTITUTION(old_size, n);
+      // rethrow
+      throw;
+    } // end catch
```
miscco, line 960: This is wrong, we are replacing all existing elements, so this should just call `uninitialized_fill`

This logic is only needed for the "enough capacity" branch

charan-003: got it, 

Thanks for the review.

charan-003:
@miscco can you please take a look at this now.
Thank you

thrust/thrust/detail/vector_base.inl
```diff
@@ -878,105 +880,99 @@ void vector_base<T, Alloc>::append(size_type n)
...
+  }
 
-        // finally, fill the range to the insertion point
-        thrust::fill_n(position, n, x);
-      } // end if
-      else
-      {
-        // construct new elements at the end of the vector
-        m_storage.uninitialized_fill_n(end(), n - num_displaced_elements, x);
+  if (m_size + n <= capacity())
```
miscco: This is not fully correct, because we could potentially overflow the addition `m_size + n`
```suggestion
  if (n <= capacity() - m_size)
```

thrust/thrust/detail/vector_base.inl
```diff
@@ -878,105 +880,99 @@ void vector_base<T, Alloc>::append(size_type n)
...
-
-        // copy num_displaced_elements - n elements to existing elements
-        // this copy overlaps
-        const size_type copy_length = (old_end - n) - position;
-        thrust::detail::overlapped_copy(position, old_end - n, old_end - copy_length);
+      m_storage.destroy(begin(), end());
+      m_size = 0;
+    }
+    return;
+  }
```
miscco, line 886: Oh boy, I really need to better read:

This is fill_insert so we are indeed adding elements, so this can just be 
```suggestion
  if (n == 0)
{
  return;
}
```

thrust/thrust/detail/vector_base.inl
```diff
@@ -878,105 +880,99 @@ void vector_base<T, Alloc>::append(size_type n)
...
-      } // end if
-      else
-      {
-        // construct new elements at the end of the vector
-        m_storage.uninitialized_fill_n(end(), n - num_displaced_elements, x);
+  if (m_size + n <= capacity())
+  {
+    // we've got room for all of them
+    const size_type num_displaced_elements = end() - position;
+    iterator old_end                       = end();
```
miscco, line 892: I would suggest to declare a variable that defines the insertion point of the exsiting elements that are pused back
```suggestion
    iterator old_end                       = end();
    iterator mid                           = position + n; 
```

thrust/thrust/detail/vector_base.inl
```diff
@@ -878,105 +880,99 @@ void vector_base<T, Alloc>::append(size_type n)
...
 
-        // extend the size
-        m_size += n - num_displaced_elements;
+    if (num_displaced_elements > n)
```
miscco, line 895: This condition is not really clear to me I would prefer
```suggestion
    if (mid < end())
```

With that we know immediately if either all existing elements are pushed into uninitialized storage, or we need to do the two step dance

It also makes the sequence of operations much clearer

```
if (mid < end())
// uninitialized_copy old elements into unused storage
// copy old elements into now freed storage
// fill(position, mid, x)
else
// uninitialized_copy(position, old_end, mid) old elements into unused storage
// uninitialized_fill(old_end, mid, x)
// fill(position, old_end, x)
```

thrust/thrust/detail/vector_base.inl
```diff
@@ -878,105 +880,99 @@ void vector_base<T, Alloc>::append(size_type n)
...
-        new_end = m_storage.uninitialized_copy(position, end(), new_end);
-      } // end try
-      catch (...)
-      {
-        // something went wrong, so destroy & deallocate the new storage
-        m_storage.destroy(new_storage.begin(), new_end);
-        new_storage.deallocate();
+      // Fill n elements with x at insertion point
+      new_storage.uninitialized_fill_n(new_end, n, x);
+      new_end += n;
```
miscco, line 947: This is not the right order of operations. We want to fill first then copy, because we do not want to touch the initial elements at all until we know we need to move them

thrust/thrust/detail/vector_base.inl
miscco: Sorry, I totally misread the algorithm as fill, not fill_insert

charan-003:
Hi @miscco, I've applied all your review feedback. Here's my understanding of the changes:

1. Changed the capacity check to avoid potential integer overflow when adding m_size + n

2. Simplified the n==0 case: When inserting zero elements, we now just return early instead of unnecessarily destroying and clearing existing elements

3. Added a mid variable: This clearly marks where the insertion ends and makes it easier to understand which existing elements get pushed back

4. Clearer condition: Using mid < end() instead of comparing counts makes it immediately obvious whether all existing elements fit in uninitialized storage or if we need to do the two-step copy

5. Reallocation order: Kept the current approach (copy elements before insertion point, fill new elements, copy elements after) since you confirmed this is the correct order for insert operations

Please let me know if I understood everything correctly, and I'll push the changes. Thanks for the detailed review!

miscco:
/ok to test [887653c](https://github.com/NVIDIA/cccl/pull/6041/commits/887653cd3a000d0510465cc412a223c5bcd11400)

charan-003:
> I pushed some minor changes, thanks a lot for working on this. These are some of the more complex operations where it is really easy to get something wrong

thanks for that.
I just removed the unused type alias in TestVectorFillInsert. The CI fix is now pushed and should resolve the CI failure.

miscco:
/ok to test [abdece1](https://github.com/NVIDIA/cccl/pull/6041/commits/abdece1bc37a479eb7eaef4d739a9def32a1b48c)

thrust/thrust/detail/vector_base.inl
```diff
@@ -878,105 +880,93 @@ void vector_base<T, Alloc>::append(size_type n)
...
+    const size_type num_displaced_elements = end() - position;
+    iterator old_end                       = end();
+    iterator mid                           = position + n;
 
-        // extend the size
-        m_size += n - num_displaced_elements;
+    if (mid < old_end)
+    {
+      // construct copy n displaced elements to new elements following the insertion
+      m_storage.uninitialized_copy(end() - n, end(), end());
```
charan-003:
```suggestion
      m_storage.uninitialized_copy(old_end - n, old_end, old_end);
```

Hopefully this might fix it?

charan-003:
@miscco, Can you please take a look when possible?
Thanks

miscco:
/ok to test [8442999](https://github.com/NVIDIA/cccl/pull/6041/commits/844299985b8cf782f976963097d7325bfb0920b1)

miscco:
Thanks a lot for working on this 🎉 

Sorry that it took so long to review

charan-003:
@miscco should we guard `_CCCL_BUILTIN_MEMMOVE` or  is there a better fix

thrust/thrust/detail/vector_base.inl
```diff
@@ -878,105 +880,93 @@ void vector_base<T, Alloc>::append(size_type n)
...
+  {
+    // we've got room for all of them
+    const size_type num_displaced_elements = end() - position;
+    iterator old_end                       = end();
+    iterator mid                           = position + n;
 
-        // extend the size
-        m_size += n - num_displaced_elements;
+    if (mid < old_end)
+    {
```
miscco, line 896: I found this to be the source of the GCC miscompilation. For whatever reason the compiler cannot compare the pointers and then goes crazy

miscco:
/ok to test [c8b4648](https://github.com/NVIDIA/cccl/pull/6041/commits/c8b4648be4a3ed781252d4960f217cc240c4c0ca)

miscco:
Thanks again for the effort

PR #6045, opened 2025-09-26
define `cuda::std::declval` in terms of new `__declfn_t` alias

ericniebler:
## Description

Currently, the default implementation of `cuda::std::declval` is implemented in terms of `cuda::std::type_identity_t`. That causes a needless class template instantiation, `cuda::std::type_identity`, when the whole point of the "fast" `declval` implementation is to avoid class template instantiations.

Instead of `type_identity_t`, use a new `__declfn_t` alias template. `__declfn_t` has other uses besides. We could use it to rewrite the `invocable` concept from:

```cpp
template <class _Fn, class... _Args>
concept invocable = requires(_Fn&& __fn, _Args&&... __args) {
  ::cuda::std::__invoke(::cuda::std::forward<_Fn>(__fn), ::cuda::std::forward<_Args>(__args)...);
};
```

to:

```cpp
template <class _Fn, class... _Args>
concept invocable = requires(__declfn_t<_Fn&&> __fn, __declfn_t<_Args&&>... __args) {
  ::cuda::std::__invoke(__fn(), __args()...);
};
```

## Checklist

### DIFF
--- a/libcudacxx/include/cuda/std/__utility/declval.h
+++ b/libcudacxx/include/cuda/std/__utility/declval.h
@@ -20,25 +20,38 @@
 #  pragma system_header
 #endif // no system header
 
-#include <cuda/std/__type_traits/type_identity.h>
 #include <cuda/std/__type_traits/void_t.h>
 
 #include <cuda/std/__cccl/prologue.h>
 
 _CCCL_BEGIN_NAMESPACE_CUDA_STD
 
+#ifndef _CCCL_DOXYGEN_INVOKED
+// Handy in concept definitions for turning:
+//   requires (T&& t) { ... std::forward<T>(t) ...; }
+// into:
+//   requires (__declfn_t<T> t) { ... t() ...; }
+//
+// NOTE: Unlike std::declval, the return type of __declfn_t is not
+// rvalue-reference-qualified. In most cases this does not matter, except when _Tp is an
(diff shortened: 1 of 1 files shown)


libcudacxx/include/cuda/std/__utility/declval.h
```diff
@@ -20,25 +20,38 @@
...
+template <class _Tp, bool _Noexcept = true>
+using __declfn_t _CCCL_NODEBUG_ALIAS = _Tp (*)() noexcept(_Noexcept);
+#endif // _CCCL_DOXYGEN_INVOKED
+
 // When variable templates and noexcept function types are available, a faster
 // implementation of declval is available. It compiles approximately 2x faster
 // than the fallback. NOTE: this implementation causes nvcc < 12.4 to ICE and
 // MSVC < 19.39 to miscompile so we use the fallback instead. The use of the
-// `type_identity_t` alias is help MSVC parse the declaration correctly.
+// `__declfn_t` alias is help MSVC parse the declaration correctly.
```
miscco, line 47: Q: Is that still the case?

miscco (approved):

libcudacxx/include/cuda/std/__utility/declval.h
ericniebler: Yes, msvc pukes if I don't use the alias

Issue #6049, opened 2025-09-28
Introduce vsmem utility to `cub::DeviceReduce`


bernhardmgruber:

bernhardmgruber:
I discussed this with @gevtushenko today in the context of my PR #6050. We believe that `cub::DeviceReduce` very likely will not run out of SMEM for very large element sizes. We concluded to not implement support for vsmem and try to add a static assert that tells users nicely if they ever run out of SMEM (instead of the compiler telling them later).

PR #6050, opened 2025-09-28
Add vsmem support to cub::DeviceReduce

bernhardmgruber:
Merge first:

- [x] #6051

Fixes: #6049

### DIFF
--- a/cub/cub/device/dispatch/dispatch_reduce.cuh
+++ b/cub/cub/device/dispatch/dispatch_reduce.cuh
@@ -859,8 +859,8 @@ struct DispatchSegmentedReduce
         static_cast<::cuda::std::int64_t>(::cuda::std::numeric_limits<::cuda::std::int32_t>::max());
       const ::cuda::std::int64_t num_invocations = ::cuda::ceil_div(num_segments, num_segments_per_invocation);
 
-      // If we need multiple passes over the segments but the iterators do not support the + operator, we cannot use
-      // the streaming approach and have to fail, returning cudaErrorInvalidValue. This is because c.parallel passes
+      // If we need multiple passes over the segments but the iterators do not support the + operator, we cannot use the
+      // streaming approach and have to fail, returning cudaErrorInvalidValue. This is because c.parallel passes
       // indirect_arg_t as the iterator type, which does not support the + operator.
       // TODO (elstehle): Remove this check once https://github.com/NVIDIA/cccl/issues/4148 is resolved.
       if (num_invocations > 1


cub/test/catch2_test_device_reduce_vsmem.cu
```diff
@@ -0,0 +1,60 @@
...
+// huge_data<x> -> sizeof(AgentReduce<...>::TempStorage) on sm86
+// 16 -> 224
+// 32 -> 368
+// 64 -> 656
+// 4Ki -> 8216
+// 16Ki -> 32792
+// 24Ki -> 49176
+// 32Ki -> 65560 (but causes Error: Formal parameter space overflowed)
+using value_types =
+  c2h::type_list<c2h::custom_type_t<c2h::equal_comparable_t, c2h::accumulateable_t, c2h::huge_data<24 * 1024>::type>>;
```
bernhardmgruber, line 26: Compile times will be interesting

bernhardmgruber: Even better, [one CI run](https://github.com/NVIDIA/cccl/actions/runs/18075194835/job/51430644070?pr=6050#step:4:7722) fails with:
```
  /home/coder/cccl/lib/cmake/cub/../../../cub/cub/thread/thread_load.cuh:123:48: fatal error: instantiating fold expression with 3073 arguments exceeded expression nesting limit of 256
    123 |   ((dst[Is] = ThreadLoad<MODIFIER>(src + Is)), ...);
        |   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~
```

bernhardmgruber: Another [one](https://github.com/NVIDIA/cccl/actions/runs/18075194835/job/51430644288?pr=6050#step:4:1216):
```
reduce.cuh(243): Error: Formal parameter space overflowed (24614 bytes required, max 4096 bytes allowed)
```

bernhardmgruber: On my machine, I failed because I ran out of 128GiB RAM.

bernhardmgruber:
So, how should we test the vsmem code path? No compiler could compile it :D Should we override the policy hub to use a block reduction using more SMEM? That feels like we are not testing the real thing.

cub/test/catch2_test_device_reduce_vsmem.cu
miscco: :this_is_fine:

bernhardmgruber:
Discussed with @gevtushenko and we will not pursue this. Rational in: #6049

PR #6061, opened 2025-09-29
Refactor DeviceSegmentedReduce

bernhardmgruber:


cub/cub/device/device_segmented_reduce.cuh
```diff
@@ -261,24 +197,29 @@ public:
...
-      d_out,
-      num_segments,
-      d_begin_offsets,
-      d_end_offsets,
-      reduction_op,
-      initial_value, // zero-initialize
-      stream);
+    using OffsetT = detail::common_iterator_value_t<BeginOffsetIteratorT, EndOffsetIteratorT>;
+    static_assert(::cuda::std::is_integral_v<OffsetT>, "Offset iterator value type should be integral.");
+    if constexpr (::cuda::std::is_integral_v<OffsetT>)
```
miscco, line 202: Those two are not sensible together, we should have the static assert and no if constexpr

bernhardmgruber: I agree. The `static_assert` should suffice. But @gevtushenko told me in the past that we have those workarounds in place so users would not see any follow-up errors after the static assert. I don't share this opinion, but that's why we do it this way here.

miscco: I am fine with keeping it as is to keep the error messages clean

miscco (approved):

PR #6065, opened 2025-09-29
Refactor `thrust::[try_]unwrap_contiguous_iterator[_t]`

bernhardmgruber:


thrust/thrust/type_traits/unwrap_contiguous_iterator.h
```diff
@@ -25,72 +25,39 @@
...
-  using type = Iterator;
-
-  static _CCCL_HOST_DEVICE type get(Iterator it)
-  {
-    return it;
-  }
-};
+//! Converts a contiguous iterator type to its underlying raw pointer type.
+template <typename ContiguousIterator>
+using unwrap_contiguous_iterator_t = decltype(unwrap_contiguous_iterator(::cuda::std::declval<ContiguousIterator>()));
```
miscco, line 40:
```suggestion
using unwrap_contiguous_iterator_t = decltype(thrust::unwrap_contiguous_iterator(::cuda::std::declval<ContiguousIterator>()));
```

miscco (approved):
I believe the whole `unwrap` machinery is not how we want to move forward.

The standard knows `pointer_traits<Iter>::to_address` as the way of a contiguous iterator to transform into a `T*` I have a branch that goes into that direction.

Thats said, the refactor here is nice, so lets bring it in anyway

bernhardmgruber:
> I believe the whole `unwrap` machinery is not how we want to move forward.
> 
> The standard knows `pointer_traits<Iter>::to_address` as the way of a contiguous iterator to transform into a `T*` I have a branch that goes into that direction.

That's a good point and we should definitely move towards that! Mind that `thrust::[try_]unwrap...` are public APIs so we must deprecate them as well before removal.

thrust/thrust/type_traits/unwrap_contiguous_iterator.h
bernhardmgruber: Yeah, maybe in the next refactoring :)

PR #6067, opened 2025-09-29
Update Catch2 to 3.10

bernhardmgruber:


bernhardmgruber:
nvc++ is failing with:
```
NVC++-S-0000-Internal compiler error. unrecognized enum base type       6  (/home/coder/cccl/libcudacxx/test/libcudacxx/cuda/ccclrt/stream/stream_smoke.cu: 140)
```
if the line number is to be trusted, we are comparing two `stream_id`s here, which are defined as:
```c++
enum class stream_id : unsigned long long
{
};
```
Unfortunately, I cannot reproduce the failure in the same devcontainer locally :S

miscco:
@dkolsen-pgi for your interest

miscco (approved):

dkolsen-pgi:
Thanks for bringing the internal error to my attention.  I'll get it fixed.  (To reproduce the failure, the enum with an underlying type of `long long` or `unsigned long long` needs to be a member of an aggregate struct and an object of that struct type needs to be copied.  I only know that by looking at the code; I never would have figured that out by trial and error.)

bernhardmgruber:
Compiler engineers are like:

> I only know that by looking at the code

image

bernhardmgruber:
Nice!

PR #6073, opened 2025-09-30
Replace CUDA Runtime calls with Driver calls in libcu++

davebayer:
We are trying to move away from CUDA Runtime APIs to Driver APIs. This PR replaces CUDA Runtime calls with Driver calls in libcu++.

Also, I added `cuInit(0)` call, so we make sure we initialize the Driver.


libcudacxx/include/cuda/__stream/stream_ref.h
```diff
@@ -238,10 +238,10 @@ class stream_ref
   //! @throws cuda_error if device check fails
   _CCCL_HOST_API device_ref device() const
   {
-    CUcontext __stream_ctx = ::cuda::__driver::__streamGetCtx(__stream);
+    ::CUcontext __stream_ctx = ::cuda::__driver::__streamGetCtx(__stream);
     __ensure_current_context __setter(__stream_ctx);
     int __id;
-    _CCCL_TRY_CUDA_API(cudaGetDevice, "Could not get device from a stream", &__id);
+    _CCCL_TRY_CUDA_API(cudaGetDevice, "Could not get device from a stream", &__id); // todo: replace with driver call
```
davebayer: Q: is there any API that can be used to get ordinal from `CUdevice`? I couldn't fine any :(

cub/cub/util_device.cuh
```diff
@@ -104,7 +103,34 @@ CUB_RUNTIME_FUNCTION inline int CurrentDevice()
 
 //! @brief RAII helper which saves the current device and switches to the specified device on construction and switches
 //! to the saved device on destruction.
-using SwitchDevice = ::cuda::__ensure_current_device;
```
davebayer, line 107: `cuda::__ensure_current_device` is no longer used in libcu++, so I moved the implementation to `cub`. It also now uses `CubDebug(...)` to check the CUDA Runtime calls

libcudacxx/include/cuda/std/__exception/cuda_error.h
```diff
@@ -109,14 +109,7 @@ class cuda_error : public ::std::runtime_error
...
 {
-#  if _CCCL_CUDA_COMPILATION()
-  NV_IF_ELSE_TARGET(NV_IS_HOST,
-                    (::cudaGetLastError(); // clear CUDA error state
-                     throw ::cuda::cuda_error(__status, __msg, __api, __loc);), //
-                    (::cuda::std::terminate();))
-#  else // ^^^ _CCCL_CUDA_COMPILATION() ^^^ / vvv !_CCCL_CUDA_COMPILATION() vvv
-  throw ::cuda::cuda_error(__status, __msg, __api, __loc);
-#  endif // !_CCCL_CUDA_COMPILATION()
+  NV_IF_TARGET(NV_IS_HOST, (throw ::cuda::cuda_error(__status, __msg, __api, __loc);), (::cuda::std::terminate();))
```
davebayer, line 112: I've changed `__throw_cuda_error` to no longer clear the CUDA Runtime error state. It was moved to `_CCCL_TRY_CUDA_API`.

davebayer (commented):
Notes for reviewers

libcudacxx/include/cuda/__stream/stream_ref.h
```diff
@@ -238,11 +238,17 @@ class stream_ref
...
   _CCCL_HOST_API device_ref device() const
   {
-    CUcontext __stream_ctx = ::cuda::__driver::__streamGetCtx(__stream);
-    __ensure_current_context __setter(__stream_ctx);
-    int __id;
-    _CCCL_TRY_CUDA_API(cudaGetDevice, "Could not get device from a stream", &__id);
-    return device_ref{__id};
+    ::CUdevice __device{};
+#  if _CCCL_CTK_AT_LEAST(12, 8)
+    __device = ::cuda::__driver::__streamGetDevice(__stream);
```
pciolkosz, line 243: We need a driver version check here

libcudacxx/include/cuda/__driver/driver_api.h
```diff
@@ -239,8 +325,70 @@ _CCCL_HOST_API void __memsetAsync(void* __dst, _Tp __value, size_t __count, ::CU
   }
 }
 
+// Unified Addressing
+
+template <::CUpointer_attribute _Attr>
+[[nodiscard]] _CCCL_API _CCCL_CONSTEVAL auto __pointer_attribute_value_type_t_impl() noexcept
```
pciolkosz, line 332: I think I would prefer the casting logic to be in some other file, this file is rather long as it is. But we also don't have where to put it yet, so maybe just TODO would be fine?

pciolkosz (approved):

cub/cub/util_device.cuh
bernhardmgruber: Let's just make sure the APIs stay the same!

libcudacxx/include/cuda/__stream/stream_ref.h
davebayer: Aaaah, but that means I would have to implement the fallback implementation here.. I will do what you did in the past - require CUDA 13.0, we can enable it in CUDA 12 once we have implemented the checks if an API is available with the current CUDA driver

cub/cub/util_device.cuh
```diff
@@ -104,7 +103,34 @@ CUB_RUNTIME_FUNCTION inline int CurrentDevice()
...
+class SwitchDevice
+{
+  int target_device_;
+  int original_device_;
+
+public:
+  //! @brief Queries the current device and if that is different than \p __target_device sets the current device to
```
miscco:
```suggestion
  //! @brief Queries the current device and if that is different than @p __target_device sets the current device to
```

libcudacxx/include/cuda/__driver/driver_api.h
```diff
@@ -323,31 +483,52 @@ _CCCL_HOST_API inline void __streamWaitEvent(::CUstream __stream, ::CUevent __ev
...
+}
+
 [[nodiscard]] _CCCL_HOST_API inline ::cudaError_t __eventDestroyNoThrow(::CUevent __evnt)
 {
   static auto __driver_fn = _CCCLRT_GET_DRIVER_FUNCTION(cuEventDestroy);
   return static_cast<::cudaError_t>(__driver_fn(__evnt));
 }
 
-_CCCL_HOST_API inline void __eventElapsedTime(::CUevent __start, ::CUevent __end, float* __ms)
+[[nodiscard]] _CCCL_HOST_API inline float __eventElapsedTime(::CUevent __start, ::CUevent __end)
```
miscco, line 508: Question should this be a `chrono::duration`?

libcudacxx/include/cuda/std/__exception/cuda_error.h
miscco: Not sure I like that too much but fine with me

libcudacxx/include/cuda/__driver/driver_api.h
davebayer: These wrappers should be really minimal and fast to include. I would go with `float` here definitely.

libcudacxx/include/cuda/std/__exception/cuda_error.h
davebayer: That's because CUDA Driver calls don't set CUDA Runtime error state. And throwing `cuda_error` would cause clearing CUDA runtime error state that was not set by our calls + we would initialize CUDA Runtime for no reason.

PR #6083, opened 2025-09-30
Add clang21 to CI.

alliepiper:


miscco:
We love our unsupported compiler thingy https://github.com/NVIDIA/cccl/actions/runs/18144608121/job/51643523655?pr=6083#step:4:533

I dont know how we can circumvent that for CMAKE

davebayer:
> I dont know how we can circumvent that for CMAKE

I made it work locally in the past by setting `CUDAFLAGS=-allow-unsupported-compiler`

alliepiper:
Yeah, we can disable those checks, but at the same time, I see little value in spending CI time testing configurations that are unsupported 🙂

davebayer:
@alliepiper I think it would make sense, because from what I've seen from the bug reports, the support is being tested internally and if test new compilers in advance, we can catch errors earlier.

Plus we can test newer version of clang-cuda which comes with support for newer CUDA features, too :)

bernhardmgruber:
> the support is being tested internally and if test new compilers in advance, we can catch errors earlier.

Could we test them as part of the cuda 99.8 CI images? Do we even run them? They appear in `matrix.yaml` but seem to not be used by any job.

alliepiper:
> Could we test them as part of the cuda 99.8 CI images? Do we even run them? They appear in matrix.yaml but seem to not be used by any job.

We only test publicly supported compiler pairings in our public CI. Internal testing is carried out by internal SWQA, we do not test the cuda99.* images on github. The matrix.yaml info for these are only used by `make_devcontainers.sh` to associate the right host compilers when regenerating the devcontainer files.

PR #6084, opened 2025-09-30
Provide `cuda::sub_overflow`

fbusato:
## Description

Add `cuda::sub_overflow` similarly to [`cuda::add_overflow` ](https://nvidia.github.io/cccl/libcudacxx/extended_api/numeric/add_overflow.html#libcudacxx-extended-api-numeric-add-overflow)

Notes:

- `cuda::add_overflow` has a very high cyclomatic complexity ( ~30 IIRC), so I tried a different approach with less control flow paths. If the new approach will get positive feedback, I will port it to `cuda::add_overflow` in another PR.
- The documentation of `cuda::add_overflow` has been slight modified to be aligned with `cuda::sub_overflow`.
- The PR also fixes a bug in `libcudacxx/include/cuda/__numeric/overflow_cast.h` : `__is_integer_representable_v` which caused `add_overflow` to fail.

### DIFF
--- a/libcudacxx/include/cuda/__numeric/add_overflow.h
+++ b/libcudacxx/include/cuda/__numeric/add_overflow.h
@@ -180,7 +180,9 @@ template <typename _Tp>
 
 template <typename _Result, typename _Lhs, typename _Rhs>
 inline constexpr bool __is_add_representable_v =
-  (sizeof(_Result) > sizeof(_Lhs) && sizeof(_Result) > sizeof(_Rhs) && ::cuda::std::is_signed_v<_Result>);
+  sizeof(_Result) > sizeof(_Lhs) && sizeof(_Result) > sizeof(_Rhs)
+  && (::cuda::std::is_signed_v<_Result>
+      || (::cuda::std::is_unsigned_v<_Lhs> && ::cuda::std::is_unsigned_v<_Rhs> && ::cuda::std::is_unsigned_v<_Result>) );
 
 /***********************************************************************************************************************
  * Public interface
@@ -193,8 +195,8 @@ _CCCL_TEMPLATE(typename _Result = void,
                typename _ActualResult = ::cuda::std::conditional_t<::cuda::std::is_void_v<_Result>, _Common, _Result>)
 _CCCL_REQUIRES((::cuda::std::is_void_v<_Result> || ::cuda::std::__cccl_is_integer_v<_Result>)
                  _CCCL_AND ::cuda::std::__cccl_is_integer_v<_Lhs> _CCCL_AND ::cuda::std::__cccl_is_integer_v<_Rhs>)
-[[nodiscard]] _CCCL_API constexpr overflow_result<_ActualResult>
-add_overflow(const _Lhs __lhs, const _Rhs __rhs) noexcept
+[[nodiscard]]
+_CCCL_API constexpr overflow_result<_ActualResult> add_overflow(const _Lhs __lhs, const _Rhs __rhs) noexcept
 {
 // (1) __builtin_add_overflow is not available in a constant expression with gcc + nvcc
 // (2) __builtin_add_overflow generates suboptimal code with nvc++ and clang-cuda for device code


docs/libcudacxx/extended_api/numeric/sub_overflow.rst
```diff
@@ -0,0 +1,88 @@
...
+
+**Parameters**
+
+- ``result``: The result of the subtraction (2).
+- ``lhs``: The left-hand side operand (1, 2).
+- ``rhs``: The right-hand side operand (1, 2).
+
+**Return value**
+
+1. Returns an :ref:`overflow_result <libcudacxx-extended-api-numeric-overflow_result>` object containing the computed  difference and a boolean flag indicating whether an overflow or underflow occurred. If the ``Result`` type is specified, it will be used as the type of the result, otherwise the common type of ``Lhs`` and ``Rhs`` is used.
```
bernhardmgruber, line 35: Q: If an overflow occurs, will the difference be the correct result in modular arithmetic? Or is it an unspecified value?

docs/libcudacxx/extended_api/numeric/sub_overflow.rst
```diff
@@ -0,0 +1,88 @@
...
+.. code:: cpp
+
+   namespace cuda {
+
+   template <class T>
+   struct overflow_result;
+
+   template <class Result = /*unspecified*/, class Lhs, class Rhs>
+   [[nodiscard]] __host__ __device__ constexpr
+   overflow_result</*see-below*/> sub_overflow(Lhs lhs, Rhs rhs) noexcept; // (1)
```
bernhardmgruber, line 17: Suggestion: Should this API return an `optional<Result>` instead?

docs/libcudacxx/extended_api/numeric.rst
```diff
@@ -34,3 +35,8 @@ Numeric
      - Performs addition with overflow checking
      - CCCL 3.2.0
      - CUDA 13.2
+
+   * - :ref:`cuda::sub_overflow <libcudacxx-extended-api-numeric-sub_overflow>`
+     - Performs subtraction with overflow checking
```
bernhardmgruber:
```suggestion
     - Performs subtraction with over- and underflow checking
```

docs/libcudacxx/extended_api/numeric/sub_overflow.rst
davebayer: It should behave as if you did the computation in an infinite precision and then casted the value to the type

docs/libcudacxx/extended_api/numeric/sub_overflow.rst
davebayer: No, I've considered this in the past and there are 2 reasons why `overflow_result` is better:
1. it has no dependency, fast to compile
2. sometimes you would like to still get the result even if the overflow happened which is something that `optional` cannot provide

docs/libcudacxx/extended_api/numeric/add_overflow.rst
```diff
@@ -58,9 +62,9 @@ Example
 
         // cuda::add_overflow(lhs, rhs) returning common type of lhs and rhs
         // 'result' is evaluated to true if an overflow occurred, false otherwise
-        if (auto result = cuda::add_overflow(1, int_max))
+        if (auto result = cuda::add_overflow(1, 2))
         {
-            assert(result.value == int_min);
+            assert(result.value == 3);
         }
```
davebayer, line 64: What was wrong with this example? I think it shows that the behaviour is well defined even when the overflow happens

docs/libcudacxx/extended_api/numeric/add_overflow.rst
```diff
@@ -58,9 +62,9 @@ Example
 
         // cuda::add_overflow(lhs, rhs) returning common type of lhs and rhs
         // 'result' is evaluated to true if an overflow occurred, false otherwise
-        if (auto result = cuda::add_overflow(1, int_max))
+        if (auto result = cuda::add_overflow(1, 2))
         {
-            assert(result.value == int_min);
+            assert(result.value == 3);
```
miscco: This is incorrect, there is no overflow in `cuda::add_overflow(1, 2)`

docs/libcudacxx/extended_api/numeric/sub_overflow.rst
```diff
@@ -0,0 +1,88 @@
...
+**Parameters**
+
+- ``result``: The result of the subtraction (2).
+- ``lhs``: The left-hand side operand (1, 2).
+- ``rhs``: The right-hand side operand (1, 2).
+
+**Return value**
+
+1. Returns an :ref:`overflow_result <libcudacxx-extended-api-numeric-overflow_result>` object containing the computed  difference and a boolean flag indicating whether an overflow or underflow occurred. If the ``Result`` type is specified, it will be used as the type of the result, otherwise the common type of ``Lhs`` and ``Rhs`` is used.
+2. Returns ``true`` if an overflow or underflow occurred. When ``false`` is returned, the computed difference is stored in ``result``.
```
miscco: Shouldnt we always store the result?

AFAIK that was one of the use cases of @davebayer

docs/libcudacxx/extended_api/numeric/sub_overflow.rst
```diff
@@ -0,0 +1,88 @@
...
+    __global__ void kernel()
+    {
+        constexpr auto int_max = cuda::std::numeric_limits<int>::max();
+        constexpr auto int_min = cuda::std::numeric_limits<int>::min();
+
+        // cuda::sub_overflow(lhs, rhs) returning the common type of lhs and rhs
+        if (auto result = cuda::sub_overflow(3, 1))
+        {
+            assert(result.value == 2);
+        }
```
miscco: I dont believe this is correct, their is neither overflow nor underflow in `cuda::sub_overflow(3, 1)`

libcudacxx/include/cuda/__numeric/sub_overflow.h
```diff
@@ -0,0 +1,336 @@
...
+#  include <intrin.h>
+#endif // _CCCL_COMPILER(MSVC) && _CCCL_ARCH(X86_64)
+
+#include <cuda/std/__cccl/prologue.h>
+
+_CCCL_BEGIN_NAMESPACE_CUDA
+
+// subtraction with unsigned types to avoid UB, return an unsigned type
+template <typename _Result, typename _Lhs, typename _Rhs>
+[[nodiscard]] _CCCL_API constexpr _Result __safe_sub(_Lhs __lhs, _Rhs __rhs) noexcept
```
miscco: I would rather call this `__sub_as_unsigned` or similar

docs/libcudacxx/extended_api/numeric/sub_overflow.rst
bernhardmgruber: That makes more sense now that I know that the value is meaningful even if an overflow occured!

docs/libcudacxx/extended_api/numeric/sub_overflow.rst
fbusato: the idea is to provide the result if the output type is unsigned, even when overflow occurs. While we are free to return any value when the result in signed (UB)

miscco (approved):

docs/libcudacxx/extended_api/numeric/sub_overflow.rst
davebayer: That's not true, see the [gcc overflow builtins](https://gcc.gnu.org/onlinedocs/gcc/Integer-Overflow-Builtins.html), they have the result always well defined:

_.. As the addition is performed in infinite signed precision, these built-in functions have fully defined behavior for all argument values. .._

docs/libcudacxx/extended_api/numeric/add_overflow.rst
```diff
@@ -28,19 +32,19 @@ The function ``cuda::add_overflow`` performs addition of two values ``lhs`` and
 
 **Return value**
 
-1. Returns an :ref:`overflow_result <libcudacxx-extended-api-numeric-overflow_result>` object that contains the result of the addition and a boolean indicating whether an overflow occurred. If the ``Result`` type is specified, it will be used as the type of the result, otherwise the common type of ``Lhs`` and ``Rhs`` is used.
-2. Returns a boolean indicating whether an overflow occurred, and if no overflow occurred, the result is stored in the provided ``result`` variable.
+1. Returns an :ref:`overflow_result <libcudacxx-extended-api-numeric-overflow_result>` object  containing the result of the addition and a boolean indicating whether an overflow or underflow occurred. If the ``Result`` type is specified, it will be used as the type of the result, otherwise the common type of ``Lhs`` and ``Rhs`` is used.
+2. Returns ``true`` if an overflow or underflow occurred. When ``false`` is returned, the computed difference is stored in ``result``.
```
davebayer: This is not valid as well, we should fix the behaviour of `add_overflow` to avoid the UB.

libcudacxx/include/cuda/__numeric/sub_overflow.h
```diff
@@ -0,0 +1,336 @@
...
+  using ::cuda::std::make_signed_t;
+  using ::cuda::std::make_unsigned_t;
+  using ::cuda::std::numeric_limits;
+  // shortcut for the case where inputs are representable with the result type
+  if constexpr (__is_sub_representable_v<_ActualResult, _Lhs, _Rhs>)
+  {
+    const auto __lhs1 = static_cast<_ActualResult>(__lhs);
+    const auto __rhs1 = static_cast<_ActualResult>(__rhs);
+    const auto __sub  = static_cast<_ActualResult>(__lhs1 - __rhs1);
+    return overflow_result<_ActualResult>{__sub, false};
```
davebayer: This will create sub optimal code for cases when `_ActualResult` is e. g. `__int128_t` and the operands are of a smaller type. But do we care?

libcudacxx/include/cuda/std/__cccl/builtin.h
```diff
@@ -361,6 +361,10 @@
 #  define _CCCL_BUILTIN_ADD_OVERFLOW(...) __builtin_add_overflow(__VA_ARGS__)
 #endif // _CCCL_CHECK_BUILTIN(__builtin_add_overflow)
 
+#if _CCCL_CHECK_BUILTIN(__builtin_sub_overflow) || _CCCL_COMPILER(GCC)
+#  define _CCCL_BUILTIN_SUB_OVERFLOW(...) __builtin_sub_overflow(__VA_ARGS__)
+#endif // _CCCL_CHECK_BUILTIN(__builtin_sub_overflow)
```
davebayer: Let's move these builtins directly to the file

libcudacxx/include/cuda/__numeric/sub_overflow.h
fbusato: yeah, not really important, but pretty easy to write

docs/libcudacxx/extended_api/numeric/add_overflow.rst
fbusato: the behavior is correct. The documentation was not.

docs/libcudacxx/extended_api/numeric/add_overflow.rst
fbusato: my intent was to simplify the example but, yes, even the overflow is fine

docs/libcudacxx/extended_api/numeric/sub_overflow.rst
fbusato: David is right. Fixed the documentation.

libcudacxx/include/cuda/__numeric/add_overflow.h
```diff
@@ -174,6 +178,10 @@ template <typename _Tp>
   return ::cuda::__add_overflow_generic_impl(__lhs, __rhs);
 }
 
+template <typename _Result, typename _Lhs, typename _Rhs>
+inline constexpr bool __is_add_representable_v =
+  (sizeof(_Result) > sizeof(_Lhs) && sizeof(_Result) > sizeof(_Rhs) && ::cuda::std::is_signed_v<_Result>);
```
davebayer: Is this condition correct? Shouldn't it also be `true` for unsinged types?

libcudacxx/include/cuda/__numeric/sub_overflow.h
```diff
@@ -0,0 +1,344 @@
...
+_CCCL_BEGIN_NAMESPACE_CUDA
+
+// subtraction with unsigned types to avoid UB, return an unsigned type
+template <typename _Result, typename _Lhs, typename _Rhs>
+[[nodiscard]] _CCCL_API constexpr _Result __sub_as_unsigned(_Lhs __lhs, _Rhs __rhs) noexcept
+{
+  using _UnsignedResult = ::cuda::std::make_unsigned_t<_Result>;
+  const auto __lhs1     = static_cast<_UnsignedResult>(__lhs);
+  const auto __rhs1     = static_cast<_UnsignedResult>(__rhs);
+  return static_cast<_Result>(__lhs1 - __rhs1);
```
davebayer: I don't think the description matches the behaviour, the result is being casted to the original type

libcudacxx/include/cuda/__numeric/add_overflow.h
fbusato: Unfortunately, no. Suppose `3 + (-4)` this is not representable with `unsigned`

davebayer: But if I have unsigned + unsigned and the result is unsigned wide enough, this trait should be also true, right?

fbusato: good point. I thought this case was already covered in the branches.

PR #6086, opened 2025-10-01
Provide `cuda::pointer_in_range`

fbusato:
## Description

Checking if a pointer is in a given range is very common. The utility has already been proposed in [WG21 P3234](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2024/p3234r0.html). It is already provided by [Boost Core](https://www.boost.org/doc/libs/develop/libs/core/doc/html/core/pointer_in_range.html) and part of libc++/libstdc++ as internal function.

```cpp
template <typename T>
[[nodiscard]] _CCCL_API inline bool pointer_in_range(T* ptr, T* start,  T* end) noexcept
```

### DIFF
--- a/docs/libcudacxx/extended_api/memory/pointer_in_range.rst
+++ b/docs/libcudacxx/extended_api/memory/pointer_in_range.rst
@@ -1,0 +1,61 @@
+.. _libcudacxx-extended-api-memory-pointer_in_range:
+
+``cuda::pointer_in_range``
+=========================
+
+.. code:: cuda
+
+   template <typename T>
+   [[nodiscard]] __host__ __device__ inline
+   T* pointer_in_range(T* ptr, T* start, T* end) noexcept
+
+Checks whether ``ptr`` lies inside the half-open interval ``[start, end)``. The interval bounds are interpreted using the standard pointer comparison semantics for ``T*``.
+
+**Template parameters**
+
+- ``T``: The type of the pointer being tested.
+
+**Parameters**
+
+- ``ptr``: The pointer being tested.
+- ``start``: Pointer to the first element in the range.
+- ``end``: Pointer one past the last element in the range.
+
+**Return value**
+
+- ``true`` when the pointer lies in ``[start, end)``, ``false`` otherwise.
+
+**Preconditions**
+
+- ``end`` must be strictly greater than ``start``.
+
+Example
+-------
+
+.. code:: cuda
+
+    #include <cuda/memory>
+
+    __global__ void kernel(float* data, size_t count) {
+        float* first = data;
+        float* last  = data + count;
+
+        float* elem_ptr = data + threadIdx.x;
+        if (cuda::pointer_in_range(elem_ptr, first, last)) {
+            *elem_ptr = static_cast<float>(threadIdx.x);
+        }
+    }
+
+    int main() {
+        size_t N          = 32;
+        float* device_ptr = nullptr;
+        cudaMalloc(&device_ptr, N * sizeof(float));
+
+        kernel<<<1, N>>>(device_ptr, N);
+        cudaDeviceSynchronize();
+
+        cudaFree(device_ptr);
+        return 0;
+    }
+
+`See it on Godbolt 🔗 <https://godbolt.org/z/EPfMErjGK>`_
--- a/libcudacxx/include/cuda/__memory/pointer_in_range.h
+++ b/libcudacxx/include/cuda/__memory/pointer_in_range.h
@@ -8,8 +8,8 @@
 //
 //===----------------------------------------------------------------------===//
 
-#ifndef _CUDA_MEMORY
-#define _CUDA_MEMORY
+#ifndef _CUDA___MEMORY_POINTER_IN_RANGE_H
+#define _CUDA___MEMORY_POINTER_IN_RANGE_H
 
 #include <cuda/std/detail/__config>
 
@@ -21,15 +21,22 @@
 #  pragma system_header
 #endif // no system header
 
-#include <cuda/__memory/address_space.h>
-#include <cuda/__memory/align_down.h>
-#include <cuda/__memory/pointer_in_range.h>
-#include <cuda/__memory/align_up.h>
-#include <cuda/__memory/aligned_size.h>
-#include <cuda/__memory/discard_memory.h>
-#include <cuda/__memory/get_device_address.h>
-#include <cuda/__memory/is_aligned.h>
-#include <cuda/__memory/ptr_rebind.h>
-#include <cuda/std/memory>
+#include <cuda/std/__functional/operations.h>
 
-#endif // _CUDA_MEMORY
+#include <cuda/std/__cccl/prologue.h>
+
+_CCCL_BEGIN_NAMESPACE_CUDA
+
+template <typename _Tp>
+[[nodiscard]] _CCCL_API inline bool pointer_in_range(_Tp* __ptr, _Tp* __start, _Tp* __end) noexcept
+{
+  _CCCL_ASSERT(::cuda::std::greater<const _Tp*>{}(__end, __start),
+               "pointer_in_range: __end must be greater than __start");
+  return ::cuda::std::less_equal<const _Tp*>{}(__start, __ptr) && ::cuda::std::less<const _Tp*>{}(__ptr, __end);
+}
+
+_CCCL_END_NAMESPACE_CUDA
+
+#include <cuda/std/__cccl/epilogue.h>
+
+#endif // _CUDA___MEMORY_POINTER_IN_RANGE_H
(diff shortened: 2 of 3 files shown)


fbusato:
next step is to aggregate memory/pointer utilities in `cuda/memory`. Right now, they are spread out throughout the library, e.g. ptr overlapping in `memcpy.h`, or address space in `host/device` mdspan.

docs/libcudacxx/extended_api/memory/pointer_in_range.rst
```diff
@@ -0,0 +1,61 @@
+.. _libcudacxx-extended-api-memory-pointer_in_range:
+
+``cuda::pointer_in_range``
+=========================
+
+.. code:: cuda
+
+   template <typename T>
+   [[nodiscard]] __host__ __device__ inline
+   T* pointer_in_range(T* ptr, T* start, T* end) noexcept
```
davebayer: I don't like that we have `ptr_rebind` but `pointer_in_range`. I would prefer to stay consistent and go with `ptr_in_range`

docs/libcudacxx/extended_api/memory/pointer_in_range.rst
```diff
@@ -0,0 +1,61 @@
...
+- ``start``: Pointer to the first element in the range.
+- ``end``: Pointer one past the last element in the range.
+
+**Return value**
+
+- ``true`` when the pointer lies in ``[start, end)``, ``false`` otherwise.
+
+**Preconditions**
+
+- ``end`` must be strictly greater than ``start``.
```
davebayer: Why? If `end == start`, it's just an empty range. I would go for `mustn't be less`

davebayer (changes requested):

docs/libcudacxx/extended_api/memory/pointer_in_range.rst
fbusato: sorry, I don't understand the relation with [ptr_rebind](https://nvidia.github.io/cccl/libcudacxx/extended_api/memory/ptr_rebind.html#libcudacxx-extended-api-memory-ptr-rebind). The function is for another purpose

davebayer: Yeah, but we should be consistent with the naming

fbusato: oh, my bad. I misunderstood your comment. I'm fine with that

fbusato:
@davebayer @miscco  open question, we don't do anything special in `cuda::std::less_equal` and similar. Would make sense to use builtin operators and `const void*` instead of a template to decrease the impact on compile time?

docs/libcudacxx/extended_api/memory/ptr_in_range.rst
```diff
@@ -0,0 +1,61 @@
+.. _libcudacxx-extended-api-memory-ptr_in_range:
+
+``cuda::ptr_in_range``
+======================
+
+.. code:: cuda
+
+   template <typename T>
+   [[nodiscard]] __host__ __device__ inline
+   T* ptr_in_range(T* ptr, T* start, T* end) noexcept
```
davebayer:
```suggestion
   bool ptr_in_range(T* ptr, T* start, T* end) noexcept
```

davebayer: And yeah, let's go with `const void*` as you said :) I wouldn't care about `volatile` in this case

fbusato:
I was also tempted to optimize the function in the same way of `cuda::in_range` but I'm scared about UB with pointers difference.

davebayer:
pre-commit.ci autofix

docs/libcudacxx/extended_api/memory/ptr_in_range.rst
```diff
@@ -0,0 +1,62 @@
...
+======================
+
+Defined in ``<cuda/memory>`` header.
+
+.. code:: cuda
+
+   namespace cuda {
+
+   [[nodiscard]] __host__ __device__ inline
+   bool ptr_in_range(const void* ptr, const void* start, const void* end) noexcept
```
davebayer: Can this be `constexpr` or will it not work?

davebayer (approved):

docs/libcudacxx/extended_api/memory.rst
```diff
@@ -64,3 +65,8 @@ Memory
      - Rebind a pointer to a different type
      - CCCL 3.1.0
      - CUDA 13.1
+
+   * - :ref:`ptr_in_range <libcudacxx-extended-api-memory-ptr_in_range>`
+     - Check if a pointer is in a range
+     - CCCL 3.1.0
+     - CUDA 13.1
```
davebayer: And we should either backport it to 3.1 or update the docs to 3.2

fbusato: thanks, copy&paste mistake

docs/libcudacxx/extended_api/memory/ptr_in_range.rst
fbusato: it should work in an ideal world :). I want to check if nvcc also agrees with that.

fbusato:
I got scared enough by https://quuxplusone.github.io/blog/2019/01/20/std-less-nightmare/ to use a safer version of the function.

docs/libcudacxx/extended_api/memory/ptr_in_range.rst
```diff
@@ -0,0 +1,67 @@
...
+
+**Template parameters**
+
+- ``T``: The type of the pointer.
+
+**Parameters**
+
+- ``ptr``: The pointer being tested.
+- ``start``: Pointer to the first element in the range.
+- ``end``: Pointer one past the last element in the range.
```
gonidelis:
```suggestion
- ``end``: Pointer to one past the last element in the range.
```

libcudacxx/include/cuda/__memory/ptr_in_range.h
```diff
@@ -0,0 +1,91 @@
...
+// - https://pvs-studio.com/en/blog/posts/cpp/1199/
+// - https://releases.llvm.org/20.1.0/tools/clang/docs/ReleaseNotes.html#resolutions-to-c-defect-reports
+
+#if _CCCL_HOST_COMPILATION()
+
+template <typename _Tp>
+[[nodiscard]] _CCCL_API bool __ptr_in_range_host(_Tp* __ptr, _Tp* __start, _Tp* __end) noexcept
+{
+  _CCCL_ASSERT(::std::greater_equal<>{}(__end, __start), "__ptr_in_range_host: __end must be greater than __start");
+  return ::std::greater_equal<>{}(__ptr, __start) && ::std::less<>{}(__ptr, __end);
```
gonidelis, line 53: just out of curiosity, is our common practice to use `::std::*` within `HOST_COMPILATION` instead of universally relying on `::cuda::std::less` (in this case) to do the host/device dispatching?

bernhardmgruber: In this case we really rely on a special blessing of `std::greater_equal` etc. in the C++ standard that allows pointers to be compared and establishes a total order. `operator<` on raw pointers only yields a partial order, which the compiler abuses in some situations. E.g.:
```c++
int a[10];
int b[10];
int* p = a +5;
if (a < p && p < a + 10) // this is legal
if (a < p && p < b) // this is NOT legal, pointers from different arrays cannot be compared. invokes UB
 if (std::less{}(a, p) && std::less{}(p, b)) // this IS legal, pointer addresses are numerically compared
```
Don't ask me why :D I really don't know.

bernhardmgruber: Here is the wording: https://eel.is/c++draft/comparisons#general-2

libcudacxx/test/libcudacxx/cuda/memory/ptr_in_range.pass.cpp
```diff
@@ -0,0 +1,77 @@
...
+{
+  constexpr auto nullptr_int = static_cast<int*>(nullptr);
+  static_assert(noexcept(cuda::ptr_in_range(nullptr_int, nullptr_int, nullptr_int)));
+  using ret_type = decltype(cuda::ptr_in_range(nullptr_int, nullptr_int, nullptr_int));
+  static_assert(::cuda::std::is_same_v<bool, ret_type>);
+
+  test_variants<int>();
+  test_variants<const int>();
+  return true;
+}
```
bernhardmgruber, line 71: Important: I think the facility should work for `void*`, can you please add tests for that? Thanks!

bernhardmgruber (commented):
LGTM, just:

libcudacxx/include/cuda/__memory/ptr_in_range.h
gonidelis: I read about it before reviewing here. My questions was why we don't use `cuda::std::*` instead of just `std::`. Not why we do not use the `<` operator. But I guess you answered by saying "_special blessing of `std::greater_equal`_" which kinda implies that the `cuda::std` counterpart does not partake of that blessing?

fbusato: yes, correct. The host compiler recognizes `std::greater_equal` but not `cuda::std::greater_equal`

Issue #3790, opened 2025-02-12
[FEA]: implement virtual shared memory in c.parallel and cuda.parallel merge_sort


NaderAlAwar:
### Is this a duplicate?

- [x] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cccl/issues) for this request and that I agree to the [Code of Conduct](CODE_OF_CONDUCT.md)

### Area

cuda.parallel (Python)

### Is your feature request related to a problem? Please describe.

`cub::DeviceMergeSort` uses virtual shared memory in order to support larger data types. This is currently not implemented in c.parallel and cuda.parallel, see https://github.com/NVIDIA/cccl/blob/5857e3d0893e9767504a9944da3d5b7ea82b4ff5/c/parallel/src/merge_sort.cu#L261-L265

### Describe the solution you'd like

Implement virtual shared memory for large data types

PR #6087, opened 2025-10-01
c.parallel: enable dynamic policies in unique_by_key.

### LINKED ISSUE #4362: Refactor unique_by_key in c.parallel to use json magic for reusing CUB tuning policies

griwes:
## Description

~~Note: the "big keys shouldn't compile" case started compiling because we now actually do the proper math, presumably that's fine? I'd like some confirmation on that one.~~

Resolves #4362 

## Checklist


c/parallel/test/test_unique_by_key.cpp
```diff
@@ -641,59 +641,6 @@ struct large_key_pair
   }
 };
 
-C2H_TEST("DeviceSelect::UniqueByKey fails to build for large types due to no vsmem", "[device][select_unique_by_key]")
-{
-  const int num_items = 1;
-
-  operation_t op = make_operation(
-    "op",
-    "struct large_key_pair { int a; char c[100]; };\n"
```
NaderAlAwar, line 650: Important: instead of deleting this, let us keep it but increase the size of the dtype. Use `char c[500]` here instead. The reason the build succeeds after your changes is that the selected policy uses a smaller block size with less items per thread, which allows us to use a type of this size without running out of shared memory.

VSmem is still not supported, if you search for `VSMemPerBlock` in `unique_by_key.cu` you can see we return 0.

This issue is tracked here https://github.com/NVIDIA/cccl/issues/3790

NaderAlAwar (approved):
Looks good, we should keep that test instead of deleting it and just increase the size for now.

c/parallel/test/test_unique_by_key.cpp
NaderAlAwar: You will also want to update the size here

bernhardmgruber (approved):
CUB part loooks good

PR #6092, opened 2025-10-01
Simplify type traits implementation pt.3

davebayer:


davebayer:
pre-commit.ci autofix

libcudacxx/include/cuda/std/__type_traits/is_signed.h
```diff
@@ -22,44 +22,55 @@
...
-_CCCL_DIAG_SUPPRESS_MSVC(4197) //  top-level volatile in cast is ignored
+#if _CCCL_CHECK_BUILTIN(is_signed)
+#  define _CCCL_BUILTIN_IS_SIGNED(...) __is_signed(__VA_ARGS__)
+#endif // _CCCL_CHECK_BUILTIN(is_signed)
+
+// if we put this trait to cuda::std::, it colides with libstdc++, putting it to to global namespace seems to work fine
+#if defined(_CCCL_BUILTIN_IS_SIGNED)
+template <class _Tp>
+inline constexpr bool __cccl_is_signed_v = _CCCL_BUILTIN_IS_SIGNED(_Tp);
+#endif // _CCCL_BUILTIN_IS_SIGNED
```
miscco, line 37: I am deeplpy scared by this and would prefer if we did not go that way.

The problem is that clang is hijacking at the parser level, which GCC does not recognize

PR #6094, opened 2025-10-01
Remove iterator workarounds for lack of operator+=

bernhardmgruber:
Those are no longer needed since `indirect_iterator_t` supports `operator+=`


bernhardmgruber:
> I want to note that this is a potentially breaking change

Techncially yes. We can still backport to 3.1, so I looked what was already present in 3.0. There, we only handled iterators without `operator+=` in segmented reductions. The documentation there says the input types must be random access iterators. In the implementation, if they were not, then the kernel would return `cudaErrorInvalidValue`. The new behavior for 3.1 and beyond would be a compilation error. That is technically a breaking change for users that violated the documented requirements of the function. I think it's ok for us to go forward with this change, but let's ask internally.

NaderAlAwar (approved):

cub/cub/device/dispatch/dispatch_radix_sort.cuh
```diff
@@ -1435,8 +1426,8 @@ struct DispatchSegmentedRadixSort
 
       if (invocation_index + 1 < num_invocations)
       {
-        detail::advance_iterators_inplace_if_supported(begin_offsets_current_it, num_current_segments);
-        detail::advance_iterators_inplace_if_supported(end_offsets_current_it, num_current_segments);
+        begin_offsets_current_it += num_current_segments;
+        end_offsets_current_it += num_current_segments;
```
xwang233, line 1430: pytorch builds failed on this,
```
#36 1147.7 /usr/local/cuda/include/cub/device/dispatch/dispatch_radix_sort.cuh(1430): error: no operator "+=" matches these operands
#36 1147.7             operand types are: at::native::<unnamed>::offset_t += const int64_t
#36 1147.7           end_offsets_current_it += num_current_segments;
#36 1147.7                                  ^
```
cc @eqy @nweidia  @bernhardmgruber

jrhemstad: @xwang233 thanks for the heads up!

This failure means that the `at::native::offset_t` iterator type doesn't have an `operator+=` that is required for a valid random access iterator. 

This is usually a pretty trivial thing to add, so the fix should be easy. 

Can you confirm if this is the correct `offset_t` in question? 

https://github.com/pytorch/pytorch/blob/cc71ab86a6985e85645424b727c766e031047ff6/aten/src/ATen/native/cuda/SortStable.cu#L21-L25

xwang233: Thank you @jrhemstad . This is the internal build log that just failed this morning http://nv/eXH. It contains the full compiler trace.

... and I guess your pointer to the source code is correct

bernhardmgruber: Should be just:
```c++
  __device__ offset_t& operator+=(int i) {
    begin += i;
    return *this;
  }
```

bernhardmgruber: >  This is the internal build log that just failed this morning http://nv/eXH.

This link does not work for me. Even with VPN.

bernhardmgruber: https://github.com/pytorch/pytorch/pull/164570

PR #6100, opened 2025-10-01
Provide utility to check pointer ranges overlapping

fbusato:
## Description

Small utility to check if two ranges overlap. 

```cpp
template<typename T>
[[nodiscard]] _CCCL_API constexpr bool ranges_overlap(
  T __lhs_start,
  T __lhs_end,
  T __rhs_start,
  T __rhs_end) noexcept
```

### DIFF
--- a/docs/libcudacxx/extended_api/memory.rst
+++ b/docs/libcudacxx/extended_api/memory.rst
@@ -16,6 +16,7 @@ Memory
    memory/is_aligned
    memory/ptr_rebind
    memory/ptr_in_range
+   memory/ranges_overlap
 
 .. list-table::
    :widths: 25 45 30 30
--- a/docs/libcudacxx/extended_api/memory/ranges_overlap.rst
+++ b/docs/libcudacxx/extended_api/memory/ranges_overlap.rst
@@ -32,6 +32,10 @@ Returns ``true`` when the half-open byte ranges ``[lhs_start, lhs_end)`` and ``[
 
 - ``true`` when the two ranges overlap, ``false`` otherwise.
 
+**Performance considerations**
+
+- The function is optimized when the ranges are contiguous and random access iterators.
+
 Example
 -------
 
@@ -53,4 +57,4 @@ Example
         return 0;
     }
 
-`See it on Godbolt 🔗 <https://godbolt.org/z/GPc4T4h7x>`_
+`See it on Godbolt 🔗 <https://godbolt.org/z/xP343GKea>`_


docs/libcudacxx/extended_api/memory/ptr_ranges_overlap.rst
```diff
@@ -0,0 +1,52 @@
...
+
+Defined in ``<cuda/memory>`` header.
+
+.. code:: cuda
+
+   namespace cuda {
+
+   [[nodiscard]] __host__ __device__ constexpr
+   bool ptr_ranges_overlap(const void* ptr1_start, const void* ptr1_end,
+                           const void* ptr2_start, const void* ptr2_end) noexcept
```
davebayer: I am thinking about whether this is a good design. When I see this API, it looks like a C API. It will not work with iterators, though it has a `range` in name.

Don't we want to name this `ranges_overlap` which would take contiguous iterators as arguments?

fbusato: that's a great point. I like it. contiguous iterators are semantically similar to pointers. Their conversion is trivial, but even better if we express the API in C++ terms.

davebayer: Yeah.. but there is a problem - the contiguous iterator is only available in C++20. I'd like to know @miscco's thoughts on this

miscco: In general there is nothing here that prevents us from taking any `forward_iterator` because we can just walk the range

We can optimize this for `contiguous_iterator` by calling `cuda::std::to_address` and then do the checks with that

miscco: I would prefer a different name

`ranges_overlap` would be more generic

fbusato: ok, now I'm thinking again about `cuda::ptr_in_range`.  Would be the case to extend it to `forward_iterator` too? I'm also fine with the current version.

libcudacxx/include/cuda/__memory/ranges_overlap.h
```diff
@@ -0,0 +1,126 @@
...
+      const auto __ptr_lhs_end   = ::cuda::std::to_address(__lhs_end);
+      const auto __ptr_rhs_begin = ::cuda::std::to_address(__rhs_begin);
+      const auto __ptr_rhs_end   = ::cuda::std::to_address(__rhs_end);
+      NV_IF_ELSE_TARGET(
+        NV_IS_HOST,
+        (return ::cuda::__ptr_ranges_overlap_host(__ptr_lhs_begin, __ptr_lhs_end, __ptr_rhs_begin, __ptr_rhs_end);),
+        (return ::cuda::__ptr_ranges_overlap_device(__ptr_lhs_begin, __ptr_lhs_end, __ptr_rhs_begin, __ptr_rhs_end);));
+    }
+  }
+  else if constexpr (::cuda::std::random_access_iterator<_Tp>)
```
fbusato, line 94: C++20 `random_access_iterator` ensures total ordering, so this path should be fine

libcudacxx/include/cuda/__memory/ranges_overlap.h
```diff
@@ -0,0 +1,126 @@
...
+ranges_overlap(_Tp __lhs_begin, _Tp __lhs_end, _Tp __rhs_begin, _Tp __rhs_end) noexcept
+{
+  if constexpr (::cuda::std::contiguous_iterator<_Tp>)
+  {
+    if (::cuda::std::__cccl_default_is_constant_evaluated())
+    {
+      // UB is not possible in a constant expression
+      _CCCL_ASSERT(__lhs_begin <= __lhs_end, "lhs range is invalid");
+      _CCCL_ASSERT(__rhs_begin <= __rhs_end, "rhs range is invalid");
+      return __lhs_begin < __rhs_end && __rhs_begin < __lhs_end;
```
miscco, line 80: I believe this is not actually checking what the name of the function says:

This check whether `__rhs` is a subset of `__lhs`. From the name of the function I would think that it  would also return true if `__rhs` partially overlaps `__lhs`

Should we rename to `is_subrange_of`

fbusato: I'm more in favor of `ranges_overlap` but I'm open to suggestion. @davebayer any thought?

davebayer: I don't think @miscco is right here, this condition checks even for partial overlaps indeed. I would go for `ranges_overlap`, too.

But we can introduce `is_subrange_of`, too

fbusato: I misread the first comment. `ranges_overlap` and  `is_subrange_of` are two different things. Agree with @davebayer

PR #6110, opened 2025-10-02
Cache device name and peers

davebayer:
Currently `device_ref::name()` and `device_ref::peer_devices()` return `std::string` and `std::vector`. I think we should avoid mixing `std::` and `cuda::std::` namespaces in our public APIs, so I've changed the implementation.

Now, both values are cached and only a non-owning const view is returned.

### DIFF
--- a/libcudacxx/include/cuda/__device/all_devices.h
+++ b/libcudacxx/include/cuda/__device/all_devices.h
@@ -207,7 +207,7 @@ inline const arch::traits_t& device_ref::arch_traits() const
   return devices[get()].__name();
 }
 
-[[nodiscard]] inline ::cuda::std::span<const device_ref> device_ref::peer_devices() const
+[[nodiscard]] inline ::cuda::std::span<const device_ref> device_ref::peers() const
 {
   return devices[get()].__peers();
 }
--- a/libcudacxx/include/cuda/__device/device_ref.h
+++ b/libcudacxx/include/cuda/__device/device_ref.h
@@ -159,7 +159,7 @@ public:
   //! The device on which this API is called is not included in the vector.
   //!
   //! @throws cuda_error if any peer access query fails
-  [[nodiscard]] ::cuda::std::span<const device_ref> peer_devices() const;
+  [[nodiscard]] ::cuda::std::span<const device_ref> peers() const;
 };
 
 _CCCL_END_NAMESPACE_CUDA
--- a/cudax/examples/simple_p2p.cu
+++ b/cudax/examples/simple_p2p.cu
@@ -200,7 +200,7 @@ try
   std::vector<cuda::device_ref> peers;
   for (auto& dev : cuda::devices)
   {
-    const auto dev_peers = dev.peer_devices();
+    const auto dev_peers = dev.peers();
     if (dev_peers.size() != 0)
     {
       peers.assign(dev_peers.begin(), dev_peers.end());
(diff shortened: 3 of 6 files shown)


libcudacxx/include/cuda/__device/all_devices.h
```diff
@@ -201,31 +202,14 @@ inline const arch::traits_t& device_ref::arch_traits() const
...
-      // it is possible to build a system with one directional peer access, check
-      // both ways here just to be safe
-      if (has_peer_access_to(__other_dev) && __other_dev.has_peer_access_to(*this))
-      {
-        __result.push_back(__other_dev);
-      }
-    }
-  }
-  return __result;
+[[nodiscard]] inline ::cuda::std::span<const device_ref> device_ref::peer_devices() const
```
davebayer: @pciolkosz don't we want to rename it to just `peers()`?

libcudacxx/include/cuda/__device/device_ref.h
```diff
@@ -166,7 +160,7 @@ class device_ref
   //! if a full group of peer devices is needed, it needs to be pushed_back separately.
   //!
   //! @throws cuda_error if any peer access query fails
-  ::std::vector<device_ref> peer_devices() const;
+  [[nodiscard]] ::cuda::std::span<const device_ref> peer_devices() const;
```
pciolkosz: I am a bit torn if this should be `span` or `vector`.
We are returning a set of peers which excludes the device this was called on. I can see in a lot of cases you would like to include that device in the set and then you need a `vector` so you can modify it.
At the same time, maybe we can return `span` and the caller is supposed to just use it to create their own vector.
Last option would be to expose two spans, one with the calling device and one without, so no vectors are needed, but then we would need to come up how to expose both versions. But then we also need to store two versions somewhere

pciolkosz (commented):
I was not sure if these things are going to be queried usually just once or multiple times and decided to just return an owning object to let the user decide if they need to store it or not. But I guess these things are small enough that it's not really important, so if the argument is caching lets us stay in `cuda::std` namespace with return types, maybe that is good enough?

libcudacxx/include/cuda/__device/all_devices.h
pciolkosz: It started as `get_peers` and then got renamed when `get_` was generally removed in https://github.com/NVIDIA/cccl/pull/4984, but I don't remember why I went with `peer_devices` instead of `peers`, I think it just sounded better?
But I don't have a strong opinion on this one

libcudacxx/include/cuda/__device/device_ref.h
pciolkosz: Also comment still says `vector`

davebayer: I see. I would propose to always return the device it was called on on the first place. The user can easily call `.subspan(1)` to get the view without the device the peers were queried for

libcudacxx/include/cuda/__device/all_devices.h
davebayer: Neither do I, but I see that we name all of the variables `peers` anyway, so I feel like why not to also name the method that way. And it's shorter to type :D

libcudacxx/include/cuda/__device/device_ref.h
davebayer: Or we can `peers_inclusive()` (or something like that) that would return the view over `[device, peer1, ...]` and implement the `peers()` method simply as `peers_inclusive().subspan(1)`.

Because I asked Copilot and after a deep think he said it's not expected to have the queried device included in the list

PR #6114, opened 2025-10-02
Add some ramblings about symbol visibility

miscco:
Symbol visibility is a huge issue that we need to address.

The current way of doing so has considerable downsides, but also gurantees some form of safety. This tries to expand the internal developer guide a bit to describe the three different problems we have.

Personally I believe that there is not possibility for us as library authors to ever solve the problem of different architectures for all possible scenarios.

The moment a user puts any function or type that calls a kernel into a library the issue reappears and there is nothing we can do about it.

### DIFF
--- a/docs/cub/developer/visibility.rst
+++ b/docs/cub/developer/visibility.rst
@@ -48,30 +48,16 @@ between the two TUs, but the kernel itself will be mangled as the same symbol, w
 
 A more detailed description can be found :ref:`here <cub-developer-guide-visibility-different-architectures>`.
 
-Possible Solutions
--------------------
+Proposed Solutions:
+--------------------
 
-For a while, the solution to these issues consisted of wrapping CUB/Thrust namespaces with
-the ``THRUST_CUB_WRAPPED_NAMESPACE`` macro so that different shared libraries have different symbols.
-This solution has poor discoverability, since issues present themselves in forms of segmentation faults, hangs,
-wrong results, etc. To eliminate the symbol visibility issues on our end, we follow the following rules:
+We can solve the first two problems by marking all kernels as ``hidden`` This ensures that we do not leak kernels.
 
-    #. Hiding symbols accepting kernel pointers:
-       it's important that an API accepting kernel pointers (e.g. ``triple_chevron``) always resides in the same
-       library as the code taking this pointers.
+However, there is no proper solution for the architecture problem. We can internally ensure that all our symbols are
+properly mangled, but even that falls short the moment we need to consider architecture families.
 
-    #. Hiding all kernels:
-       it's important that kernels always reside in the same library as the API using these kernels.
-
-    #. Incorporating GPU architectures into symbol names:
-       it's important that kernels compiled for a given GPU architecture are always used by the host
-       API compiled for that architecture.
-
-To satisfy (1), the visibility of ``thrust::cuda_cub::detail::triple_chevron`` is hidden.
-
-To satisfy (2), instead of annotating kernels as ``__global__`` we annotate them as
-``CUB_DETAIL_KERNEL_ATTRIBUTES``. Apart from annotating a kernel as global function, the macro also
-contains an attribute to set the visibility to hidden.
-
-To satisfy (3), CUB symbols are placed inside an inline namespace containing the set of
-GPU architectures for which the TU is being compiled.
+Furthermore, we cannot require our users to use all the same workarounds we do. What we *can* do is ensure that all
+kernels are unique symbols, so that if we call a kernel internally we kow that it is always the correct one.
+For that we can either use an inline namespace or preferably a deduced non-type template argument that uniquely
+represents the architectures a library is compiled for. The latter has the clear benefit that it is much shorter than #
+the current inline namespace name, as mangling of an integer is relatively short and simple.
--- a/docs/cub/developer/visibility.rst
+++ b/docs/cub/developer/visibility.rst
@@ -9,6 +9,7 @@ Symbol Visibility
 
    visibility/host_stub_visibility
    visibility/device_kernel_visibility
+   visibility/different_architectures
 
 Using CUB/Thrust in shared libraries is a known source of issues. This relates to the visibility of the kernel functions
 but also of the stub functions NVCC generates to actually call the kernel.
@@ -33,6 +34,19 @@ inside ``lib_b`` instead of ``lib_b::foo``, or vice versa. The CUDA runtime from
 the kernel function pointer we passed from ``lib_b``.
 
 A more detailed description can be found :ref:`here <cub-developer-guide-visibility-device-kernel-visibility>`.
+
+Problem 3: linking TUs compiled for different architectures
+------------------------------------------------------------
+
+This is orthogonal to the visibility of the functions themself but relates to ODR violations in case a TU is compiled
+for different architectures. As new architectures come out, we adopt new features to provide the best possible
+performance for all existing architectures.
+
+However, consider a kernel that contains an `NV_IF_TARGET` that conditionally selects different architecture features.
+If we build 2 TUs for different architectures then we will end up in a situation that the kernel implementation differs
+between the two TUs, but the kernel itself will be mangled as the same symbol, which is a clear ODR violation.
+
+A more detailed description can be found :ref:`here <cub-developer-guide-visibility-different-architectures>`.
 
 Possible Solutions
 -------------------
--- a/docs/cub/developer/visibility.rst
+++ b/docs/cub/developer/visibility.rst
@@ -8,6 +8,7 @@ Symbol Visibility
    :maxdepth: 1
 
    visibility/host_stub_visibility
+   visibility/device_kernel_visibility
 
 Using CUB/Thrust in shared libraries is a known source of issues. This relates to the visibility of the kernel functions
 but also of the stub functions NVCC generates to actually call the kernel.
@@ -21,6 +22,17 @@ That stub function has weak linkage, so if both libraries try to launch ``kernel
 and the other kernel launch ^might silently fail.
 
 A more detailed description can be found :ref:`here <cub-developer-guide-visibility-host-stub-visibility>`.
+
+Problem 2: Calling kernels from inside a shared library
+--------------------------------------------------------
+
+This is quite similar to Problem 1 above. Again a project links two shared libraries ``lib_a`` and ``lib_b``. However,
+this time we call a library function ``foo`` that takes a function pointer to a kernel as an argument and invokes it.
+If ``foo`` has weak external linkage (e.g ``thrust::triple_chevron``) we might end up calling ``lib_a::foo`` from a
+inside ``lib_b`` instead of ``lib_b::foo``, or vice versa. The CUDA runtime from ``lib_a`` will not be able to call
+the kernel function pointer we passed from ``lib_b``.
+
+A more detailed description can be found :ref:`here <cub-developer-guide-visibility-device-kernel-visibility>`.
 
 Possible Solutions
 -------------------
--- a/docs/cub/developer/visibility.rst
+++ b/docs/cub/developer/visibility.rst
@@ -3,12 +3,32 @@
 Symbol Visibility
 ==================
 
-Using CUB/Thrust in shared libraries is a known source of issues.
+.. toctree::
+   :glob:
+   :maxdepth: 1
+
+   visibility/host_stub_visibility
+
+Using CUB/Thrust in shared libraries is a known source of issues. This relates to the visibility of the kernel functions
+but also of the stub functions NVCC generates to actually call the kernel.
+
+Problem 1: Selecting the right kernel stub
+-------------------------------------------
+
+Consider a project that links two shared libraries ``lib_a`` and ``lib_b`` that involve a kernel call of some global
+``kernel``. The compiler will generate a stub function that handles actually launching the kernel via the CUDA runtime.
+That stub function has weak linkage, so if both libraries try to launch ``kernel`` only one host stub will be selected
+and the other kernel launch ^might silently fail.
+
+A more detailed description can be found :ref:`here <cub-developer-guide-visibility-host-stub-visibility>`.
+
+Possible Solutions
+-------------------
+
 For a while, the solution to these issues consisted of wrapping CUB/Thrust namespaces with
 the ``THRUST_CUB_WRAPPED_NAMESPACE`` macro so that different shared libraries have different symbols.
-This solution has poor discoverability,
-since issues present themselves in forms of segmentation faults, hangs, wrong results, etc.
-To eliminate the symbol visibility issues on our end, we follow the following rules:
+This solution has poor discoverability, since issues present themselves in forms of segmentation faults, hangs,
+wrong results, etc. To eliminate the symbol visibility issues on our end, we follow the following rules:
 
     #. Hiding symbols accepting kernel pointers:
        it's important that an API accepting kernel pointers (e.g. ``triple_chevron``) always resides in the same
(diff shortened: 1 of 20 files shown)


docs/cub/developer/visibility.rst
```diff
@@ -3,29 +3,61 @@
...
+
+Proposed Solutions:
+--------------------
+
+We can solve the first two problems by marking all kernels as ``hidden`` This ensures that we do not leak kernels.
+
+However, there is no proper solution for the architecture problem. We can internally ensure that all our symbols are
+properly mangled, but even that falls short the moment we need to consider architecture families.
+
+Furthermore, we cannot require our users to use all the same workarounds we do. What we *can* do is ensure that all
```
bernhardmgruber:
```suggestion
Furthermore, we cannot require our users to use all the same workarounds we do. What we *can* do is to ensure that all
```

docs/cub/developer/visibility/different_architectures.rst
```diff
@@ -0,0 +1,185 @@
...
+Lets look at the prime example ``thrust::device_vector``, which uses a kernel for initialization. What happens if we
+accidentally run the kernel from another shared library compiled with different architectures? Worst case we are
+eating some performance regressions because the kernel will not utilize advanced features of a new architecture,
+but in the end the result of calling that kernel will not change the outcome.
+
+This is because the kernel call is consistent *within* each library. As long as the user facing API does not rely on
+specific internals of a kernel to be called -which it should not-, then any of the two libraries will do.
+
+Finally, the architectures that are passed around in ``__CUDA_ARCH_LIST__`` do *not* discriminate architecture families.
+There is currently not programmatic way to discriminate a library that has been compiled for ``SM90a`` from one that was
```
pciolkosz: "There is currently no programatic way"

docs/cccl/development/visibility/device_kernel_visibility.rst
```diff
@@ -0,0 +1,133 @@
...
+and so the right stub function will always be selected.
+
+Now imagine that both libraries are going to defer launching of their kernels to a function ``foo`` common to both
+``lib_a`` and ``lib_b``, which has weak external linkage. This might happen in ``CUB``, because it launches
+kernels through the ``thrust::triple_chevron`` helper.
+
+Similar to :ref:`Problem 1 <cub-developer-guide-visibility-host-stub-visibility>` the linker will pick one of the two
+weak symbols and subsequently ``lib_a`` will try to pass its own kernel ``d_kernel<int>_a`` to ``lib_b::foo``.
+
+However, the CUDA runtime in ``lib_b`` will not find any kernel registered at the address of ``d_kernel<int>_a`` and
```
pciolkosz, line 89: A bit off topic but I was wondering if we should mention that CUDA Runtime now has `cudaKernel_t` that you could pass around instead and that will allow you to control where the stub -> kernel lookup happens vs where the kernel launch happens

docs/cub/developer/visibility.rst
```diff
@@ -3,29 +3,61 @@
...
+A more detailed description can be found :ref:`here <cub-developer-guide-visibility-host-stub-visibility>`.
 
-To satisfy (3), CUB symbols are placed inside an inline namespace containing the set of
-GPU architectures for which the TU is being compiled.
+Problem 2: Calling kernels from inside a shared library
+--------------------------------------------------------
+
+This is quite similar to Problem 1 above. Again a project links two shared libraries ``lib_a`` and ``lib_b``. However,
+this time we call a library function ``foo`` that takes a function pointer to a kernel as an argument and invokes it.
+If ``foo`` has weak external linkage (e.g ``thrust::triple_chevron``) we might end up calling ``lib_a::foo`` from a
```
pauleonix:
```suggestion
If ``foo`` has weak external linkage (e.g ``thrust::triple_chevron``) we might end up calling ``lib_a::foo`` from
```

docs/cub/developer/visibility.rst
```diff
@@ -3,29 +3,61 @@
...
 
-    #. Incorporating GPU architectures into symbol names:
-       it's important that kernels compiled for a given GPU architecture are always used by the host
-       API compiled for that architecture.
+Problem 1: Selecting the right kernel stub
+-------------------------------------------
 
-To satisfy (1), the visibility of ``thrust::cuda_cub::detail::triple_chevron`` is hidden.
+Consider a project that links two shared libraries ``lib_a`` and ``lib_b`` that involve a kernel call of some global
+``kernel``. The compiler will generate a stub function that handles actually launching the kernel via the CUDA runtime.
```
gevtushenko: imporant: only kernel templates lead to weak symbols, just kernels are fine:
```suggestion
Consider a project that links two shared libraries ``lib_a`` and ``lib_b`` that involve a kernel call of some global
``kernel`` template. The compiler will generate a stub function that handles actually launching the kernel via the CUDA runtime.
```

docs/cub/developer/visibility.rst
```diff
@@ -3,29 +3,61 @@
...
+That stub function has weak linkage, so if both libraries try to launch ``kernel`` only one host stub will be selected
+and the other kernel launch ^might silently fail.
```
gevtushenko: suggestion: this doesn't look like a RST syntax, doesn't render well:
```suggestion
and the other kernel launch might silently fail.
```

jrhemstad:
> The moment a user puts any function or type that calls a kernel into a library the issue reappears and there is nothing we can do about it.

This is true, unless the library author knows about this problem and is taking the necessary precautions, e.g., RAPIDS. 

For those cases, in order for them to do the right thing, they still depend on us to do the right thing with our symbols. 

In other words, for a well-formed user program, it is necessary, but not _sufficient_, for us to properly guard our symbols.

docs/cub/developer/visibility.rst
fbusato: I would mention https://developer.nvidia.com/blog/cuda-c-compiler-updates-impacting-elf-visibility-and-linkage/ and highlights the differences between CTK 12.x and 13.x

docs/cub/developer/visibility.rst
```diff
@@ -3,29 +3,61 @@
...
+
+This is quite similar to Problem 1 above. Again a project links two shared libraries ``lib_a`` and ``lib_b``. However,
+this time we call a library function ``foo`` that takes a function pointer to a kernel as an argument and invokes it.
+If ``foo`` has weak external linkage (e.g ``thrust::triple_chevron``) we might end up calling ``lib_a::foo`` from a
+inside ``lib_b`` instead of ``lib_b::foo``, or vice versa. The CUDA runtime from ``lib_a`` will not be able to call
+the kernel function pointer we passed from ``lib_b``.
+
+A more detailed description can be found :ref:`here <cub-developer-guide-visibility-device-kernel-visibility>`.
+
+Problem 3: linking TUs compiled for different architectures
```
fbusato: suggestion: minimize abbreviations

docs/cub/developer/visibility.rst
```diff
@@ -3,29 +3,61 @@
...
+
+We can solve the first two problems by marking all kernels as ``hidden`` This ensures that we do not leak kernels.
+
+However, there is no proper solution for the architecture problem. We can internally ensure that all our symbols are
+properly mangled, but even that falls short the moment we need to consider architecture families.
+
+Furthermore, we cannot require our users to use all the same workarounds we do. What we *can* do is ensure that all
+kernels are unique symbols, so that if we call a kernel internally we kow that it is always the correct one.
+For that we can either use an inline namespace or preferably a deduced non-type template argument that uniquely
+represents the architectures a library is compiled for. The latter has the clear benefit that it is much shorter than #
```
fbusato: typo?

docs/cub/developer/visibility/host_stub_visibility.rst
```diff
@@ -0,0 +1,110 @@
...
+Host Stub Visibility Issue
+---------------------------
+
+Consider the following simple TU:
+
+.. code:: cpp
+
+    template <class T>
+    __global__ void kernel(T *val) {
+        ::printf("kernel: set val = 42\n");
```
fbusato:
```suggestion
        printf("kernel: set val = 42\n");
```

docs/cccl/development/visibility/host_stub_visibility.rst
```diff
@@ -0,0 +1,110 @@
+.. _cub-developer-guide-visibility-host-stub-visibility:
+
+
+Host Stub Visibility Issue
+---------------------------
+
+Consider the following simple TU:
+
+.. code:: cpp
+
```
fbusato, line 10:
```suggestion
#include <cstdio>
```

docs/cub/developer/visibility/host_stub_visibility.rst
```diff
@@ -0,0 +1,110 @@
...
+        *val = 42;
+    }
+
+   int main() {
+     int *ptr{};
```
fbusato: not initialized

docs/cub/developer/visibility/host_stub_visibility.rst
```diff
@@ -0,0 +1,110 @@
...
+     kernel<<<1, 1>>>(ptr);
+   }
+
+The cuda compiler frontend will turn this into:
```
fbusato:
```suggestion
The CUDA compiler frontend will turn this into:
```

docs/cub/developer/visibility/host_stub_visibility.rst
bernhardmgruber: It's initialized to zero. I guess for the demo, this is fine? But the program could not be run this way, true.

ericniebler:
i learned recently about the GNU [`__abi_tag__`](https://gcc.gnu.org/onlinedocs/gcc/C_002b_002b-Attributes.html#index-abi_005ftag-function-attribute) attribute which we can use to make symbols unique where it is supported. i'm not certain if our toolchain supports it.

docs/cub/developer/visibility.rst
```diff
@@ -3,29 +3,61 @@
...
+the current inline namespace name, as mangling of an integer is relatively short and simple.
```
gevtushenko: critical: the document is in the developer overview section and should target contributors. This document used to provide specific instructions on how to address symbol visibility issue. After modification, it's mostly listing problems. Let's:
- move the solution section to the top of the document and leave problems that it solves as a detailed references
- list specific guidance for CCCL developers to address symbol visibility issues on CCCL end
- split the solution into two sections: algorithms (how do we solve visibility issues for free and static member functions), and containers (how do we solve visibility issue for vectors etc.)

miscco: Moved the proposed solution to top and dropped the ramblings ;)

docs/cccl/development/visibility/device_kernel_visibility.rst
miscco: I would like to leave that as a followup, because I am not too deep into that topic

docs/cub/developer/visibility/host_stub_visibility.rst
miscco: Moved to a `__shared__` variable

docs/cccl/development/visibility.rst
```diff
@@ -0,0 +1,63 @@
...
+and their host stubs, but also ODR violations that arise from compiling with different CUDA architectures.
+
+To alleviate those issues we have derived the following rules to ensure that users can safely rely on CCCL features in
+shared libraries.
+
+1. Every kernel should be annotated as ``hidden`` through ``CCCL_DETAIL_KERNEL_ATTRIBUTES``
+2. Every function or type that eventually calls a kernel in a subsequent function call or member function must be put in
+   a namespace that disambiguates the CUDA architectures the library was compiled with.
+and the other kernel launch might silently fail.
+   disambiguated free functions
```
gevtushenko: important: these seems to be unrelated / leftover? 

```suggestion
```

docs/cccl/development/visibility.rst
```diff
@@ -0,0 +1,63 @@
...
+
```
gevtushenko, line 25: critical: one step is missing. The namespace is not enough, we also need to hide functions accepting kernel pointers:

```suggestion
  3. Hiding symbols accepting kernel pointers:
     it's important that an API accepting kernel pointers (e.g. ``triple_chevron``) always resides in the same
     library as the code taking this pointers.
```

gevtushenko (approved):
Left a few comments that we should address before merging, otherwise looks good!

docs/cccl/development/visibility/host_stub_visibility.rst
```diff
@@ -0,0 +1,114 @@
...
+  #include <cstdio>
+
+  template <class T>
+  __global__ void kernel(T *val) {
+      printf("kernel: set val = 42\n");
+      *val = 42;
+  }
+
+  __shared__ int val;
```
fbusato: this is not correct. Replace with `__device__`

docs/cccl/development/visibility/host_stub_visibility.rst
```diff
@@ -0,0 +1,114 @@
...
+
+  int main() {
+
+     kernel<<<1, 1>>>(cuda::std::addressof(val));
```
fbusato: this is also wrong. Should use `cudaGetSymbolAddress()` https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__MEMORY.html#group__CUDART__MEMORY_1g4f513be54d3794667c2017146b3d6a2b

docs/cccl/development/visibility/host_stub_visibility.rst
```diff
@@ -0,0 +1,114 @@
+.. _cccl-development-visibility-host-stub-visibility:
+
+
+Host Stub Visibility Issue
+---------------------------
+
+Consider the following simple TU:
```
fbusato: prefer Translation Unit (TU)

docs/cccl/development/visibility/host_stub_visibility.rst
```diff
@@ -0,0 +1,114 @@
...
+
+.. code-block:: cpp
```
fbusato, line 9: consider adding file name(s)

docs/cccl/development/visibility/different_architectures.rst
```diff
@@ -0,0 +1,185 @@
...
+      return (Archs + ... + 0);
+    }
+
+    // kernel with architecture dependent symbol name and functionality
+    template <class T, auto Archs = sum_archs<__CUDA_ARCH_LIST__>()>
+    __global__ void kernel(T *val) {
+      *val = sum_archs<__CUDA_ARCH_LIST__>();
+    }
+
+    __attribute__((visibility("hidden"))) __forceinline__ int use_kernel() {
```
fbusato: `inline`

docs/cccl/development/visibility/different_architectures.rst
```diff
@@ -0,0 +1,185 @@
...
+      int *d_val{};
+      cudaMalloc(&d_val, sizeof(size_t));
```
fbusato:
```suggestion
      cudaMalloc(&d_val, sizeof(d_val));
```

docs/cccl/development/visibility/different_architectures.rst
```diff
@@ -0,0 +1,185 @@
...
+      kernel<<<1, 1>>>(d_val);
+      int ret;
+      if (cudaMemcpy(&ret, d_val, sizeof(size_t), cudaMemcpyDeviceToHost) !=
+          cudaSuccess) {
+        std::printf("c: FAILED to copy from device to host\n");
+      }
+      return ret;
+    }
+
+    template <class T = int> struct some_class_with_kernel {
```
fbusato:
```suggestion
    template <class T = int>
    struct some_class_with_kernel {
```

PR #6122, opened 2025-10-03
Implement `cudax::cufile`

davebayer:
This PR implements wrappers of [`CUfileHandle_t` and related APIs](https://docs.nvidia.com/gpudirect-storage/api-reference-guide/index.html#cufile-file-handle-apis).

It introduces:
1. `cudax::cufile` type which handles opening/closing the native file handle and registering/deregistering of the cuFile file handle. It owns both of these resources.
2. `cudax::cufile_ref` type that is a lightweight non-owning wrapper for the cuFile file handle. It doesn't own the resources (with exception below).
3. `cuda::cufile_driver.[de]register_native_handle(...) -> cuda::cufile_ref` that can be used for manual registration of the native file handle. It has basically the same semantics as `new`/`delete`.

So, right now these are the scenarios are covered:
1. the user has some legacy code from where he gets the `CUfileHandle_t` -> he can use `cuda::cufile_ref` with our APIs.
2. the user has some code that gives him the native file handle, but he doesn't own the native file handle -> he can use `cuda::cufile_driver.register_native_handle(...)`.
3. the user wants us to open the native file handle for him as well -> he can use `cuda::cufile`

We might introduce a third option: `cuda::scoped_cufile` or something similar, that would be constructible from `cuda::cufile_ref` and would deregister the handle when going out of scope, but I would wait with it.


cudax/include/cuda/experimental/__cufile/cufile_ref.cuh
```diff
@@ -0,0 +1,81 @@
...
+#endif // no system header
+
+#include <cuda/std/__utility/swap.h>
+
+#include <cufile.h>
+
+namespace cuda::experimental
+{
+
+class cufile_ref
```
pciolkosz, line 31: Do we need the `_ref` type if it does not offer a constructor from some outside native handle?

davebayer: This is a bit more complicated. Since `cufile_ref` implements `is_open()` method, just `CUfileHandle_t` is not enough, I need also the `native_handle`.

So, we could:
1. move the `is_open()` method to `cufile` exclusively (which I don't like)
2. throw if one uses `is_open()` on `cufile_ref` that was not created together with the `native_handle`
3. keep it as is

I'm not sure what's the best option here

davebayer: Or we can use `CUfileDescr_t` and do the registration in the `cufile_ref` constructor, but that doesn't seem to follow the design we have in other components

pciolkosz: I guess what I was suggesting is to remove `cufile_ref` completely, users can use `cufile&` instead. The reason we have `_ref` files is that we want to be compatible with old CUDA interfaces where the ownership of the underlying object stays on the old API side. So I believe we need to make the `cufile_ref` constructor from native handle and `CUfileHandle_t` public or otherwise we can remove this class

cudax/test/cufile/driver.cu
```diff
@@ -0,0 +1,290 @@
...
+  // 19. Close the driver.
+  test_close();
+
+  // 20. Test that the driver is closed.
+  test_is_open(false);
+
+  // 21. Close the driver again.
+  test_close();
+
+  // 22. Test that the driver is closed.
```
jvera-nvidia, line 287: we need to check if opening -> closing -> reopening should always work (it should but driver was not designed with this possiblity in mind AFAIK)

cudax/include/cuda/experimental/__cufile/cufile_ref.cuh
jvera-nvidia: there are two use cases in the original cufile:
a file handle "owning" the underlying file descriptor and a file handle which does not own the underlying file descriptor. 

So if we want to remove cufile_ref; then this owning concept won't be there anymore. We could still make references to "owning fd" cufile but we could not create non-owning cufile from native handles.

In my opinion the later doesn't make a lot of sense and in cases its needed the user can always use the underlying cufile API directly.

I'd vote to remove the cufile_ref as well.

davebayer: I've reworked the `cufile_ref`. Now it behaves as other `cuda::meow_ref` wrappers. For now it just implements the `.get()` method, but we might add other, like `.write(...)` or `.read(...)`. I removed the `.is_open()` method.

cudax/include/cuda/experimental/__cufile/cufile.cuh
```diff
@@ -0,0 +1,324 @@
...
+
+    ::CUfileHandle_t __handle{};
+    _CCCL_TRY_CUFILE_API(::cuFileHandleRegister, "Failed to register cuFile handle", &__handle, &__desc);
+    return __handle;
+  }
+
+  //! @brief Wrapper for retrieving the open mode.
+  [[nodiscard]] static _CCCL_HOST_API cufile_open_mode __open_mode(native_handle_type __native_handle)
+  {
+    int __oflags = ::fcntl(__native_handle, F_GETFL);
```
jvera-nvidia, line 110: do we really need to retrieve the opening mode? 
I'm a bit worried on non-standard things in the bindings for when eventually cufile is windows supported.

cudax/include/cuda/experimental/__cufile/cufile.cuh
```diff
@@ -0,0 +1,324 @@
...
+    try
+    {
+      __cufile_handle_ = __register_cufile_handle(__native_handle_);
+    }
+    catch (...)
+    {
+      __close_file(::cuda::std::exchange(__native_handle_, __invalid_native_handle));
+      throw;
+    }
+  }
```
jvera-nvidia, line 274: I'm not sure we need this... its two ways of doing the same (this vs constructor) and this actually is forbidden if the constructor was already called. I'd vote to remove this one.

cudax/include/cuda/experimental/__cufile/cufile.cuh
```diff
@@ -0,0 +1,324 @@
...
+  _CCCL_HOST_API void close()
+  {
+    if (!is_open())
+    {
+      return;
+    }
+
+    ::cuFileHandleDeregister(::cuda::std::exchange(__cufile_handle_, nullptr));
+    __close_file(::cuda::std::exchange(__native_handle_, __invalid_native_handle));
+  }
```
jvera-nvidia, line 290: having these manual state management methods can be confusing, I'd very much prefer relying on C++ ctors and dtors alone and just have a single way of doing things.

davebayer: But this allows us to simply change the `__meow(...)` implementation when coming to different platform. And also, I can have the thing implemented once and reuse it in constructor and the `.open(...)` for example.

I would like to keep it as is

cudax/include/cuda/experimental/__cufile/cufile.cuh
davebayer: But for example `std::fstream` also implements the `.open(...)` method. I'd like to stay as close as what is common in the standard as possible

jvera-nvidia: but we can always implement it later if needed... I don't see why this is needed; we are not trying to stay close to fstream either, are we?

davebayer: That's true, but I'd like to provide tools users are used to have. Consider this:
```cpp
cuda::cufile file;
// ...
file = cuda::cufile{filename, open_mode}; // open the file
// ...
file = cuda::cufile{}; // close the file
```
You have a situation like this, when you want to have the object in not opened state. Then, to open the file, you must create another object and move assign it to the original object. And if you want to close the file with detection whether it was closed or not, you must (again) move assign a new default constructed object.

In my opinion this is *not* a good design. I'd like to do:
```cpp
cuda::cufile file;
// ...
file.open(filename, open_mode); // open the file
// ...
file.close(); // close the file
```

jvera-nvidia:
Fair enough, those are valid points. I was unsure we needed it but as you said the easier of use and known methods are totally valid and I think it's good to have!

Enviado desde Outlook para Android<https://aka.ms/AAb9ysg>
________________________________

cudax/test/cufile/common.h
```diff
@@ -0,0 +1,32 @@
...
+#pragma once
+
+#include <cstdio>
+
+#include <fcntl.h>
+#include <unistd.h>
+
+void test_check_fd_is_valid(int fd)
+{
+  CUDAX_REQUIRE(fcntl(fd, F_GETFD) == 0);
```
pciolkosz: Could correct flags be non-zero here? Internet says to do:
`fcntl(fd, F_GETFD) != -1 || errno != EBADF;`
For this use case probably just the first part is ok

pciolkosz (approved):

cudax/test/cufile/common.h
davebayer: Oh yeah, you are right

davebayer:
Waiting for https://github.com/nv-gha-runners/roadmap/issues/318

davebayer:
> Waiting for https://github.com/nv-gha-runners/roadmap/issues/318

As discussed internally, we will just build the tests for now, we will run those tests once the support is added. I tested it locally with CUDA 12.9 and 13.0 and both compiled & ran fine.

cudax/CMakeLists.txt
```diff
@@ -40,7 +40,12 @@ option(
   "Enable STF tests/examples that use cublas/cusolver."
   OFF
 )
-option(cudax_ENABLE_CUFILE "Enable cuFile in CUDA Experimental" OFF)
+option(cudax_ENABLE_CUFILE "Enable cuFile in CUDA Experimental" ON)
+option(
+  cudax_ENABLE_CUFILE_TESTS_RUN
+  "Enable cuFile tests run on top of compilation"
+  ON
+)
```
alliepiper: Q: This is unconventional for our build system. Is there a reason to specifically toggle execution (but not building) of these tests?

cudax/test/CMakeLists.txt
```diff
@@ -51,6 +51,15 @@ function(cudax_add_catch2_test target_name_var test_name cudax_target) # ARGN=te
...
 
+## cudax_add_catch2_cufile_test
+#
+# Same as cudax_add_catch2_test but links CUDA::cuFile to the target.
+function(cudax_add_catch2_cufile_test target_name_var test_name cudax_target) # ARGN=test sources
+  cudax_add_catch2_test(${target_name_var} ${test_name} ${cudax_target} ${ARGN})
+  target_link_libraries(${${target_name_var}} PRIVATE CUDA::cuFile)
+  set(${target_name_var} ${target_name_var} PARENT_SCOPE)
+endfunction()
+
```
alliepiper: Unused

cudax/test/CMakeLists.txt
```diff
@@ -154,21 +163,14 @@ foreach (cudax_target IN LISTS cudax_TARGETS)
...
-  if (cudax_ENABLE_CUFILE)
-    cudax_add_catch2_test(cufile_driver_target cufile_driver ${cudax_target}
-      cufile/driver.cu
-    )
-
-    set(cufile_test_targets ${cufile_driver_target})
-    foreach (cufile_test_target IN LISTS cufile_test_targets)
-      target_link_libraries(${cufile_test_target} PRIVATE CUDA::cuFile)
-    endforeach()
-  endif()
```
alliepiper, line 167: This looks closer to what we'd usually do -- call the existing `cudax_add_catch2_test` and then call `target_link_libraries`:

```cmake
  if (cudax_ENABLE_CUFILE)
    cudax_add_catch2_test(test_target cufile_driver_attributes ${cudax_target}
      cufile/driver_attributes.cu
    )
    target_link_libraries(${test_target} PRIVATE CUDA::cuFile)

    cudax_add_catch2_test(test_target cufile_driver_register ${cudax_target}
      cufile/driver_register.cu
    )
    target_link_libraries(${test_target} PRIVATE CUDA::cuFile)

    cudax_add_catch2_test(test_target cufile_cufile ${cudax_target}
      cufile/cufile.cu
    )
    target_link_libraries(${test_target} PRIVATE CUDA::cuFile)

    cudax_add_catch2_test(test_target cufile_cufile_ref ${cudax_target}
      cufile/cufile_ref.cu
    )
    target_link_libraries(${test_target} PRIVATE CUDA::cuFile)

    cudax_add_catch2_test(test_target cufile_open_mode ${cudax_target}
      cufile/open_mode.cu
    )
    target_link_libraries(${test_target} PRIVATE CUDA::cuFile)
  endif() # cudax_ENABLE_CUFILE
```

This would remove a lot of the extra complexity around these new tests and fit with the current design more consistently.

alliepiper (changes requested):

cudax/CMakeLists.txt
alliepiper: Just saw this:

> we will just build the tests for now, we will run those tests once the support is added. I tested it locally with CUDA 12.9 and 13.0 and both compiled & ran fine.

In that case, let's just exclude running the tests from the presets with a regex, similar to how we exclude certain STF tests from CI here: https://github.com/NVIDIA/cccl/blob/main/CMakePresets.json#L719-L727

We can drop the special CMake option and special handling this way.

PR #6128, opened 2025-10-03
Provide `cuda::div_overflow`

fbusato:
## Description

Add `cuda::div_overflow` similarly to [cuda::add_overflow](https://nvidia.github.io/cccl/libcudacxx/extended_api/numeric/add_overflow.html#libcudacxx-extended-api-numeric-add-overflow) and `cuda::sub_overflow` #6084

### DIFF
--- a/docs/libcudacxx/extended_api/numeric/div_overflow.rst
+++ b/docs/libcudacxx/extended_api/numeric/div_overflow.rst
@@ -1,0 +1,92 @@
+.. _libcudacxx-extended-api-numeric-div_overflow:
+
+``cuda::div_overflow``
+======================
+
+Defined in ``<cuda/numeric>`` header.
+
+.. code:: cpp
+
+   namespace cuda {
+
+   template <class T>
+   struct overflow_result;
+
+   template <class Result = /*unspecified*/, class Lhs, class Rhs>
+   [[nodiscard]] __host__ __device__ constexpr
+   overflow_result</*see-below*/> div_overflow(Lhs lhs, Rhs rhs) noexcept; // (1)
+
+   template <class Result, class Lhs, class Rhs>
+   [[nodiscard]] __host__ __device__ constexpr
+   bool div_overflow(Result& result, Lhs lhs, Rhs rhs) noexcept; // (2)
+
+   } // namespace cuda
+
+The function ``cuda::div_overflow`` performs integer division of ``lhs`` by ``rhs`` with overflow and error detection. The result is the same as if the operands were first promoted to an infinite precision signed type, divided, and the result truncated to the type of the return value.
+
+**Parameters**
+
+- ``result``: Receives the quotient when no overflow is detected (2).
+- ``lhs``: The dividend (1, 2).
+- ``rhs``: The divisor (1, 2).
+
+**Return value**
+
+1. Returns an :ref:`overflow_result <libcudacxx-extended-api-numeric-overflow_result>` object containing the computed quotient and a boolean flag indicating whether an overflow or underflow occurred. If the ``Result`` type is specified, it will be used as the type of the result, otherwise the common type of ``Lhs`` and ``Rhs`` is used.
+2. Returns ``true`` if an overflow or underflow occurred. When ``false`` is returned, the computed quotient is stored in ``result``.
+
+**Constraints**
+
+- ``Result``, ``Lhs``, and ``Rhs`` must be integer types.
+
+**Notes**
+
+- When ``Result`` is a wider signed type than both operands, the quotient is always representable unless ``rhs`` is zero.
+- For signed types, ``-1`` divided by ``numeric_limits<Rhs>::min()`` triggers overflow when ``Result`` matches the operand types.
+- Division by zero is always reported as overflow and yields a default-initialized result value.
+
+**Performance considerations**
+
+- No overflow checking if ``Lhs / Rhs`` is always  representable with the ``Result`` type.
+- The computation with unsigned types is faster than signed types.
+  
+Example
+-------
+
+.. code:: cuda
+
+    #include <cuda/numeric>
+    #include <cuda/std/cassert>
+    #include <cuda/std/limits>
+
+    __global__ void kernel()
+    {
+        constexpr auto int_min = cuda::std::numeric_limits<int>::min();
+
+        // cuda::div_overflow(lhs, rhs) returning the common type of the operands
+        if (auto result = cuda::div_overflow(-1, int_min))
+        {
+            assert(result.overflow);
+        }
+
+        // cuda::div_overflow<Result>(lhs, rhs) with an explicitly wider result type
+        auto wide = cuda::div_overflow<long long>(-1, int_min);
+        assert(!wide.overflow);
+        assert(wide.value == 0);
+
+        assert(cuda::div_overflow(3, 0));
+
+        unsigned quotient{};
+        // cuda::div_overflow(result, lhs, rhs) with bool return type
+        bool overflow = cuda::div_overflow(quotient, 10u, 2u);
+        assert(!overflow);
+        assert(quotient == 5u);
+    }
+
+    int main()
+    {
+        kernel<<<1, 1>>>();
+        cudaDeviceSynchronize();
+    }
+
+`See it on Godbolt 🔗 <https://godbolt.org/z/GhEqWGeb4>`_
--- a/libcudacxx/include/cuda/__numeric/div_overflow.h
+++ b/libcudacxx/include/cuda/__numeric/div_overflow.h
@@ -1,0 +1,98 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of the libcu++ Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef _CUDA___NUMERIC_DIV_OVERFLOW_H
+#define _CUDA___NUMERIC_DIV_OVERFLOW_H
+
+#include <cuda/std/detail/__config>
+
+#if defined(_CCCL_IMPLICIT_SYSTEM_HEADER_GCC)
+#  pragma GCC system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_CLANG)
+#  pragma clang system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_MSVC)
+#  pragma system_header
+#endif // no system header
+
+#include <cuda/__numeric/overflow_cast.h>
+#include <cuda/__numeric/overflow_result.h>
+#include <cuda/std/__concepts/concept_macros.h>
+#include <cuda/std/__limits/numeric_limits.h>
+#include <cuda/std/__type_traits/common_type.h>
+#include <cuda/std/__type_traits/conditional.h>
+#include <cuda/std/__type_traits/is_integer.h>
+#include <cuda/std/__type_traits/is_signed.h>
+#include <cuda/std/__type_traits/is_unsigned.h>
+#include <cuda/std/__type_traits/is_void.h>
+
+#include <cuda/std/__cccl/prologue.h>
+
+_CCCL_BEGIN_NAMESPACE_CUDA
+
+template <typename _Result, typename _Lhs, typename _Rhs>
+inline constexpr bool __is_div_representable_v =
+  (sizeof(_Result) > sizeof(_Lhs) && sizeof(_Result) > sizeof(_Rhs) && ::cuda::std::is_signed_v<_Result>)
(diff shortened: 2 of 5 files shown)


docs/libcudacxx/extended_api/numeric/div_overflow.rst
```diff
@@ -0,0 +1,92 @@
...
+- ``rhs``: The divisor (1, 2).
+
+**Return value**
+
+1. Returns an :ref:`overflow_result <libcudacxx-extended-api-numeric-overflow_result>` object containing the computed quotient and a boolean flag indicating whether an overflow or underflow occurred. If the ``Result`` type is specified, it will be used as the type of the result, otherwise the common type of ``Lhs`` and ``Rhs`` is used.
+2. Returns ``true`` if an overflow or underflow occurred. When ``false`` is returned, the computed quotient is stored in ``result``.
+
+**Constraints**
+
+- ``Result``, ``Lhs``, and ``Rhs`` must be integer types.
```
gonidelis, line 44: `integer types` or `integral types`?

docs/libcudacxx/extended_api/numeric/div_overflow.rst
```diff
@@ -0,0 +1,92 @@
...
+
+**Performance considerations**
+
+- No overflow checking if ``Lhs / Rhs`` is always  representable with the ``Result`` type.
+- The computation with unsigned types is faster than signed types.
+
+Example
+-------
+
+.. code:: cuda
```
gonidelis, line 58: suggestion: not sure what the standard practice is for libcu++, but in CUB/Thrust we are shifting towards writing examples in separate files for two reasons:

1. They are offered within the codebase as standalone executables.
2. We make sure they are actually compiling properly (especially if any changes/updates happen - which btw reflect immediately when chunk of code is simply included).

We use `.. literalinclude::` for that reason.

gonidelis: I see there is a godbolt link which basically serves the two points above

docs/libcudacxx/extended_api/numeric/div_overflow.rst
fbusato: `integer type` is fine, @miscco suggested the introduction of the category. `integral` also includes enumerators

libcudacxx/test/libcudacxx/cuda/numeric/overflow.arithmetic/div_overflow.pass.cpp
```diff
@@ -0,0 +1,189 @@
...
+    const auto result = cuda::div_overflow<Result>(lhs, rhs);
+    if (!overflow)
+    {
+      if (special_case)
+      {
+        assert(result.value == expected);
+      }
+      else
+      {
+        assert(result.value == static_cast<Result>(static_cast<Result>(lhs) / static_cast<Result>(rhs)));
```
gonidelis, line 36: nit: Are the  `static_cast<Result>`s on `lhs` and `rhs` necessary here? I see that our test cases are concrete at the moment e.g:

  ```cpp
 // 3. 0 / 1 -> should not overflow
  test_div_overflow<Result>(Lhs{0}, Rhs{1}, false);
```

but wouldn't a test case like 

```cpp
  test_div_overflow<Result>(Lhs{0}, Rhs{2^31 + 1}, false);
```
lead to undefined behavior due potential truncation ?

libcudacxx/test/libcudacxx/cuda/numeric/overflow.arithmetic/div_overflow.pass.cpp
```diff
@@ -0,0 +1,189 @@
...
+  test_div_overflow<Result>(Lhs{0}, Rhs{0}, true);
+
+  // 3. 0 / 1 -> should not overflow
+  test_div_overflow<Result>(Lhs{0}, Rhs{1}, false);
+
+  // 4. 1 / 1 -> should not overflow
+  test_div_overflow<Result>(Lhs{1}, Rhs{1}, false);
+
+  // 5. 1 / -1 -> should overflow if the destination type is unsigned
+  if constexpr (is_signed_v<Rhs>)
```
gonidelis: just for the shake of coverage would it make sense to add a test case where `lhs` is negative and `rhs` is  positive to see if unsigned result triggers the overflow?

libcudacxx/test/libcudacxx/cuda/numeric/overflow.arithmetic/div_overflow.pass.cpp
```diff
@@ -0,0 +1,189 @@
...
+  {
+    if constexpr (is_unsigned_v<Result>)
+    {
+      test_div_overflow<Result>(Lhs{1}, Rhs{-1}, true);
+    }
+    else
+    {
+      test_div_overflow<Result>(Lhs{1}, Rhs{-1}, false, true, Result{-1});
```
gonidelis, line 95: why do we need the special case here? isn't `1/-1`a standard, constexpr-safe operation? Or are we just "pretending as if it weren't"

libcudacxx/test/libcudacxx/cuda/numeric/overflow.arithmetic/div_overflow.pass.cpp
fbusato: we have all the cases with `lhs_max`. `rhs_max` is not very interesting. `static_cast<Result>` is used to avoid all kind of implicit conversions what could happen

libcudacxx/test/libcudacxx/cuda/numeric/overflow.arithmetic/div_overflow.pass.cpp
fbusato: yes, that's a good point

gonidelis: feel free to resolve when done

libcudacxx/test/libcudacxx/cuda/numeric/overflow.arithmetic/div_overflow.pass.cpp
gonidelis: If these implicit conversions are not part of what `div_overflow` is trying to mitigate feel free to resolve this comment

docs/libcudacxx/extended_api/numeric/div_overflow.rst
```diff
@@ -0,0 +1,92 @@
...
+
+**Constraints**
+
+- ``Result``, ``Lhs``, and ``Rhs`` must be integer types.
+
+**Notes**
+
+- When ``Result`` is a wider signed type than both operands, the quotient is always representable unless ``rhs`` is zero.
+- For signed types, ``-1`` divided by ``numeric_limits<Rhs>::min()`` triggers overflow when ``Result`` matches the operand types.
+- Division by zero is always reported as overflow and yields a default-initialized result value.
```
davebayer: Division by zero should just trap. See the [paper](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2025/p3161r3.html#bad_div) that proposes the overflow arithmetic in C++. Also [rust](https://doc.rust-lang.org/std/primitive.i32.html#method.overflowing_div) does the same thing here

fbusato: what about `_CCCL_VERIFY`?

davebayer: I would go for `_CCCL_ASSERT`

libcudacxx/include/cuda/__numeric/div_overflow.h
```diff
@@ -0,0 +1,149 @@
...
+               typename _Lhs,
+               typename _Rhs,
+               typename _Common       = ::cuda::std::common_type_t<_Lhs, _Rhs>,
+               typename _ActualResult = ::cuda::std::conditional_t<::cuda::std::is_void_v<_Result>, _Common, _Result>)
+_CCCL_REQUIRES((::cuda::std::is_void_v<_Result> || ::cuda::std::__cccl_is_integer_v<_Result>)
+                 _CCCL_AND ::cuda::std::__cccl_is_integer_v<_Lhs> _CCCL_AND ::cuda::std::__cccl_is_integer_v<_Rhs>)
+[[nodiscard]]
+_CCCL_API constexpr overflow_result<_ActualResult> div_overflow(const _Lhs __lhs, const _Rhs __rhs) noexcept
+{
+  _CCCL_ASSERT(__rhs != _Rhs{0}, "division by zero");
```
miscco, line 71: Devils advocate here, is there *any* situation other than `-1 / numeric_limits::min` that could result in an oderflow?

If not, should we just explicitly check for this combination and and otherwise return the division?

fbusato: yes, e.g. `1000 / 2 -> int8_t` or `1000 / -2 -> unsigned`

gonidelis: also mixed sign with result type too small

libcudacxx/test/libcudacxx/cuda/numeric/overflow.arithmetic/div_overflow.pass.cpp
```diff
@@ -0,0 +1,188 @@
...
+  //--------------------------------------------------------------------------------------------------------------------
+  //  trivial cases
+  // 1. 0 / 1 -> should not overflow
+  test_div_overflow<Result>(Lhs{0}, Rhs{1}, false);
+
+  // 2. 1 / 1 -> should not overflow
+  test_div_overflow<Result>(Lhs{1}, Rhs{1}, false);
+
+  // nvcc 12.0 doesn't support this case in a constexpr context
+#if _CCCL_CUDA_COMPILER(NVCC, !=, 12, 0)
```
miscco, line 100: I believe we should then wrap the whole test in a `!__cccl_default_is_constant_evaluated()` so that we at least test runtime with 12.0

docs/libcudacxx/extended_api/numeric/div_overflow.rst
```diff
@@ -0,0 +1,92 @@
...
+
+- ``rhs != 0``
+
+**Constraints**
+
+- ``Result``, ``Lhs``, and ``Rhs`` must be integer types.
+
+**Notes**
+
+- For signed types, ``-1`` divided by ``numeric_limits<Rhs>::min()`` triggers overflow when ``Result`` matches the operand types.
```
davebayer:
```suggestion
- For signed types, ``numeric_limits<Lhs>::min()`` divided by ``-1`` triggers overflow when ``Result`` matches the operand types.
```

libcudacxx/include/cuda/__numeric/div_overflow.h
```diff
@@ -0,0 +1,149 @@
...
+      const auto __rhs1           = static_cast<_UnsignedCommon>(__rhs);
+      const auto __result         = __lhs1 / __rhs1;
+      const auto __is_overflow    = ::cuda::std::cmp_greater(__result, __result_max);
+      return overflow_result<_ActualResult>{static_cast<_ActualResult>(__result), __is_overflow};
+    }
+    else // lhs and rhs are mixed positive/negative -> negative result
+    {
+      if constexpr (is_unsigned_v<_ActualResult>)
+      {
+        return overflow_result<_ActualResult>{_ActualResult{}, __lhs != 0}; // the result is negative
```
davebayer: We should still return the result of the division here

libcudacxx/test/libcudacxx/cuda/numeric/overflow.arithmetic/div_overflow.pass.cpp
```diff
@@ -0,0 +1,197 @@
...
+  test_type<Lhs, Rhs, cuda::std::common_type_t<Lhs, Rhs>>();
+  test_type<Lhs, Rhs, unsigned>();
+  test_type<Lhs, Rhs, int>();
+  // instantiation of all tests is very expensive. clang hits "constexpr evaluation hit maximum step limit"
+#if _CCCL_COMPILER(GCC)
+  test_type<Lhs, Rhs, long long>();
+  test_type<Lhs, Rhs, unsigned long long>();
+  test_type<Lhs, Rhs, signed char>();
+  test_type<Lhs, Rhs, unsigned char>();
+#endif
```
miscco: Nitpick: We do add the corresponding comments at the end of conditional compilation blocks

libcudacxx/test/libcudacxx/cuda/numeric/overflow.arithmetic/div_overflow.pass.cpp
```diff
@@ -0,0 +1,197 @@
...
+}
+
+template <typename Lhs, typename Rhs>
+__host__ __device__ constexpr void test_type()
+{
+  test_type<Lhs, Rhs, cuda::std::common_type_t<Lhs, Rhs>>();
+  test_type<Lhs, Rhs, unsigned>();
+  test_type<Lhs, Rhs, int>();
+  // instantiation of all tests is very expensive. clang hits "constexpr evaluation hit maximum step limit"
+#if _CCCL_COMPILER(GCC)
```
miscco: Nitpick: This talks about clang but checks for GCC

fbusato: 🤦‍♂️great catch

PR #6130, opened 2025-10-05
Remove fully qualified ::cuda::std:: from examples

### LINKED ISSUE #4479: [BUG]: Don't fully qualify `cuda::std` replacements in examples
Recent changes have replaced things like `thrust::plus` with `cuda::std::plus`. Within Thrust's implementation, `cuda::std` is fully qualified (e.g. `::cuda::std::plus`).

However, the fully qualified names have also been used in updates to the examples. This is unnecessary, undesirable, and ugly. It can be confusing to end users, who may wonder what that end syntax is and why only certain things in the examples use it. Please fix this so that in the examples, we use the qualified (`cuda::std::plus`) but not fully qualified (`::cuda::std::plus`) names.

charan-003:
This PR removes fully qualified `::cuda::std::` names from Thrust examples, replacing them with the simpler qualified form `cuda::std::`.

**Changes:**
- Replaced `::cuda::std::` with `cuda::std::` in 19 example files
- Removed legacy `#include "../include/host_device.h"`

Closes #4479

### DIFF
--- a/thrust/examples/constant_iterator.cu
+++ b/thrust/examples/constant_iterator.cu
@@ -12,8 +12,7 @@ int main()
   thrust::device_vector<int> data{3, 7, 2, 5};
 
   // add 10 to all values in data
-  thrust::transform(
-    data.begin(), data.end(), thrust::constant_iterator<int>(10), data.begin(), cuda::std::plus<int>());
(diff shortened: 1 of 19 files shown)


davebayer (approved):
Thanks for working on this 🎊

PR #6127, opened 2025-10-03
Improve fully cached build times.

alliepiper:
Switches to the rapids fork of sccache (https://github.com/rapidsai/devcontainers/pull/588) and sets up a persistent sccache preprocessor cache using Github Cache.

Replaces the fragile sccache output parsing with just dumping the raw `sccache --show-adv-stats` -- no more 1063% hit rates 😅

Drops the time for a fully-cached windows build from >1hr to ~10-15min:

image

Similarly reduces the fully-cache linux build times from ~15m to <5m.

PR #5725, opened 2025-08-30
Enable MSVC GPU jobs for CCCL projects

alliepiper:
## Summary

Enables Windows (MSVC) GPU CI jobs across CCCL projects: workflow action, PowerShell build/test scripts, and matrix compiler/GPU entries. Prerequisite fixes land first so each is independently bisectable.

**Prerequisite fixes:**
- Close temp file handle before removing it (`c/parallel`)
- Guard `__device__` lambdas with `EXTENDED_LAMBDA` in `for_each_in_extents` and `for_each_in_layout` tests (`cub`)
- Shorten long pstl scan test filenames (`libcudacxx`)
- Skip two failing tests on MSVC with issue refs — `no_unique_address` (#8953), `tuple_size_structured_bindings` (#8952) (`libcudacxx`)

**Windows CI infrastructure:**
- `.github/actions/workflow-run-job-windows/action.yml` — PowerShell runner action with colored output
- `ci/windows/build_common.psm1`, `test_libcudacxx.ps1`, `test_packaging.ps1` — build/test scripts
- `ci/matrix.yaml` — MSVC compiler entries (`msvc2019`, `msvc14.39`, `msvc2022`) across CTK versions; A100 GPU spec; cudax excluded until #8903 is resolved
- `cmake/CCCLUtilities.cmake` — `_CCCL_WSL` / `_CCCL_WINDOWS` detection helpers
- `cudax/test/CMakeLists.txt` — unified-memory tests guarded against Windows/WSL

PR #6131, opened 2025-10-05
Initial batch of changes to setup GPU windows runners.

alliepiper:
This does not enable any builds on them, but sets-up the infra while we fix library issues (#5725) and improve caching behavior (#6127).

The `ci/util/[workflow|artifact]/*` and `ci/upload_*_artifacts.sh` scripts are now cross platform and reused between linux and windows.

This also switches artifact compression from bzip2 to zstd. On our windows runners:

image


.github/actions/workflow-run-job-windows/action.yml
```diff
@@ -23,83 +57,166 @@ runs:
...
         role-duration-seconds: 43200 # 12 hours
+
     - name: Checkout repo
       uses: actions/checkout@v4
       with:
         path: ${{github.event.repository.name}}
         persist-credentials: false
-    - name: Fetch ${{ inputs.image }}
+
+    - name: Install jq on host vm
```
leofang: IIRC Justin said `jq` is already installed on the self-hosted runners.

alliepiper: Ah, nice. It was missing from our devcontainers, I'll double check the host VMs.

alliepiper: Yep! Looks like it's already there 👍

ci/windows/build_cub.ps1
```diff
@@ -1,4 +1,4 @@
-Param(
```
alliepiper, line 1: These seemingly empty changes fix the BOM to UTF-8.

ci/matrix.yaml
```diff
@@ -269,6 +269,8 @@ workflows:
   exclude:
     # GPU runners are not available on Windows.
     - {jobs: ['test', 'test_gpu', 'test_nolid', 'test_lid0', 'test_lid1', 'test_lid2'], cxx: ['msvc2019', 'msvc14.39', 'msvc2022']}
+    # cudax doesn't support C++17 on msvc:
+    - {project: 'cudax', std: 17, cxx: ['msvc2019', 'msvc14.39', 'msvc2022']}
```
alliepiper, line 273: Fun fact, we've been running msvc+cudax+cpp17 jobs for months, and they've been silently failing due to error propagation being dumb in powershell. That's fixed now, so we need this exclusion.

.github/actions/workflow-run-job-linux/action.yml
```diff
@@ -179,19 +179,15 @@ runs:
...
-            --env "GITHUB_ENV=$GITHUB_ENV" \
             --env "GITHUB_SHA=$GITHUB_SHA" \
-            --env "GITHUB_PATH=$GITHUB_PATH" \
-            --env "GITHUB_OUTPUT=$GITHUB_OUTPUT" \
             --env "GITHUB_ACTIONS=$GITHUB_ACTIONS" \
             --env "GITHUB_REF_NAME=$GITHUB_REF_NAME" \
             --env "GITHUB_RUN_ID=$GITHUB_RUN_ID" \
-            --env "GITHUB_WORKSPACE=$GITHUB_WORKSPACE" \
             --env "GITHUB_REPOSITORY=$GITHUB_REPOSITORY" \
-            --env "GITHUB_STEP_SUMMARY=$GITHUB_STEP_SUMMARY" \
```
alliepiper, line 191: These vars are removed as a) they're not used, and b) many of them refer to paths on the host that aren't mounted in the container.

.github/actions/workflow-run-job-windows/install_gpu_driver.ps1
leofang, line 1: The update looks so nice that I wanna steal it and use it in CUDA Python too! 👍  We'll need this for https://github.com/NVIDIA/cuda-python/issues/985 (cc @cryos for vis)

alliepiper: Feel free! I stole the initial implementation from y'all in the first place 😄

ci/windows/build_cub.ps1
bernhardmgruber: Nice, I wondered already yesterday whether it was an encoding thing!

examples/CMakeLists.txt
```diff
@@ -32,6 +32,11 @@ set(cmake_cpm_opts
   -D "CCCL_TAG=${CCCL_EXAMPLE_CPM_TAG}"
 )
 
+if (MSVC)
+  message(STATUS "Skipping example build-and-test CTests on MSVC")
+  return()
+endif()
+
```
bernhardmgruber, line 39: Q: Don't these examples work on Windows? Should we try to get them working? If yes, we let's create an issue for that.

alliepiper: It's not worth it. This is a limitation of the compile-and-test machinery implemented in CMake/CTest.

The examples do work when manually configured, and we have other tests that verify that the packaging scripts we use here are working, and those do run on windows.

PR #6134, opened 2025-10-06
Refactor Thrust `destroy_range` and `device_[new|delete|free]`

bernhardmgruber:


miscco (approved):
We should definitely consolidate with the std conformant parts in libcu++

PR #6135, opened 2025-10-06
Remove `cuda::physical_device` from public API

### LINKED ISSUE #6137: Don't expose `cuda::physical_device`
`cuda::physical_device` is a type that is the `value_type` of `cuda::devices`. It represents a device and is used to store cached device properties. It looks something like:
```cpp
class physical_device : public device_ref
{
public:
  const arch::traits_t& arch_traits() const noexcept;
  ::CUcontext primary_context() const;
  ~physical_device();
  [[nodiscard]] ::cuda::std::string_view __name() const;
  [[nodiscard]] ::cuda::std::span<const device_ref> __peers() const;

private:
  mutable ::CUcontext __primary_ctx = nullptr;
  mutable ::CUdevice __device{};
  mutable ::std::once_flag __init_once;
  static constexpr ::cuda::std::size_t __max_name_length{256};
  mutable ::std::once_flag __init_once_name_{};
  mutable char __name_[__max_name_length]{};
  mutable ::cuda::std::size_t __name_length_{};
  mutable ::std::once_flag __init_once_peers_{};
  mutable ::std::vector<device_ref> __peers_{};
  arch::traits_t __traits;

  explicit physical_device(int __id);

  // `device` objects are not movable or copyable.
  physical_device(physical_device&&)                 = delete;
  physical_device(const physical_device&)            = delete;
  physical_device& operator=(physical_device&&)      = delete;
  physical_device& operator=(const physical_device&) = delete;
};
```
The type is not intended to be used by users, it's single purpose only. It's not copyable, not movable. It also contains many members that are `mutable` because they are set on first use of any of the methods.

Since it serves only the single purpose, I'd like to not expose the type and make it internal only as an implementation detail. We can replace it's use with `cuda::device_ref` instead. It will make the API more clear and the implementation as well.

davebayer:
With @pciolkosz we decided that we needn't to expose this class and we will keep it as an implementation detail.

Closes #6137.

### DIFF
--- a/libcudacxx/include/cuda/__device/all_devices.h
+++ b/libcudacxx/include/cuda/__device/all_devices.h
@@ -23,10 +23,11 @@
 
 #if _CCCL_HAS_CTK() && !_CCCL_COMPILER(NVRTC)
 
+#  include <cuda/__device/device_ref.h>
 #  include <cuda/__device/physical_device.h>
 #  include <cuda/__driver/driver_api.h>
-#  include <cuda/std/cassert>
-#  include <cuda/std/detail/libcxx/include/stdexcept>
+#  include <cuda/__fwd/devices.h>
+#  include <cuda/std/__cstddef/types.h>
 #  include <cuda/std/span>
 
 #  include <vector>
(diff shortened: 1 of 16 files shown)


libcudacxx/include/cuda/__device/all_devices.h
```diff
@@ -23,141 +23,71 @@
...
 
-  bool operator==(const __initializer_iterator& __other) const noexcept
+  [[nodiscard]] _CCCL_HOST_API iterator end() const
   {
-    return __id_ == __other.__id_;
+    return ::cuda::__devices().end();
   }
 
-  bool operator!=(const __initializer_iterator& __other) const noexcept
+  _CCCL_HOST_API operator ::cuda::std::span<const device_ref>() const
```
pciolkosz: The operator might be not needed now, because we should be able to use the span constructor from a range

libcudacxx/include/cuda/__device/physical_device.h
```diff
@@ -150,63 +146,42 @@ class physical_device : public device_ref
...
   }
+};
 
-private:
-  // TODO: put a mutable thread-safe (or thread_local) cache of device
-  // properties here.
-
-  friend class device_ref;
-  friend struct __detail::__emplace_device;
+[[nodiscard]] _CCCL_HOST_API inline ::std::unique_ptr<__physical_device[]>
```
pciolkosz, line 151: Why unique_ptr to an array instead of a vector?

libcudacxx/include/cuda/__device/physical_device.h
```diff
@@ -24,107 +24,102 @@
...
+    {
+      [[maybe_unused]] const auto __ignore = ::cuda::__driver::__primaryCtxReleaseNoThrow(__device_);
+    }
   }
 
   //! @brief Retrieve the primary context for this device.
   //!
   //! @return A reference to the primary context for this device.
-  ::CUcontext primary_context() const
+  [[nodiscard]] _CCCL_HOST_API ::CUcontext __primary_context()
```
pciolkosz, line 85: Is this something we would like to expose in a public API? I am not sure, especially since it's a driver type, but I also thought it could be useful. It also served as an explicit initializer, we might want to provide at least `init()` on `device_ref` now

libcudacxx/include/cuda/__device/physical_device.h
davebayer: There is a problem with the type not being copyable

libcudacxx/include/cuda/__device/physical_device.h
davebayer: I've added the `init()` method

libcudacxx/include/cuda/std/__cccl/visibility.h
```diff
@@ -44,7 +44,7 @@
 #endif // !_CCCL_COMPILER(NVRTC)
 
 #if _CCCL_COMPILER(MSVC)
-#  define _CCCL_VISIBILITY_DEFAULT __declspec(dllimport)
```
pciolkosz, line 47: Should this be a separate PR?

davebayer: Hm, you are probably right. I will keep the functions without any attribute for now and we can add them later

pciolkosz (approved):

libcudacxx/include/cuda/__device/arch_traits.h
```diff
@@ -182,10 +184,10 @@ struct traits_t
 // @brief Architecture traits
 // Template function that returns the traits for an architecture with a given id.
 template <id _Id>
-[[nodiscard]] _CCCL_HOST_DEVICE constexpr traits_t traits();
+[[nodiscard]] _CCCL_API constexpr traits_t traits();
 
 template <>
-[[nodiscard]] _CCCL_HOST_DEVICE inline constexpr traits_t traits<id::sm_60>()
+[[nodiscard]] _CCCL_API inline constexpr traits_t traits<id::sm_60>()
```
bernhardmgruber, line 190: Remark: Those changes do not look like they are needed to remove `cuda::physical_device` from the public API. Please separate those next time. It will make bisecting and backporting etc. easier.

libcudacxx/include/cuda/__device/device_ref.h
```diff
@@ -103,30 +89,35 @@ class device_ref
...
   //!
   //! @return The memory location of this device
-  [[nodiscard]] operator memory_location() const noexcept
+  [[nodiscard]] _CCCL_HOST_API operator memory_location() const noexcept
   {
     return memory_location{::cudaMemLocationTypeDevice, get()};
   }
 
+  //! @brief Initializes the primary context of the device.
+  _CCCL_HOST_API void init() const; // implemented in <cuda/__device/physical_device.h> to avoid circular dependency
```
bernhardmgruber, line 113: Suggestion: please document what this function does, or give it a better name.

bernhardmgruber (approved):
CUB is fine for me.

libcudacxx/include/cuda/__device/all_devices.h
```diff
@@ -23,140 +23,74 @@
...
+    __ret.emplace_back(static_cast<int>(__i));
   }
+  return __ret;
+}
 
-  __emplace_device operator->() const noexcept
-  {
-    return __emplace_device{__id_};
-  }
+[[nodiscard]] inline ::cuda::std::span<const device_ref> __devices()
```
fbusato, line 50: `_CCCL_HOST_API`?

libcudacxx/include/cuda/__device/all_devices.h
```diff
@@ -23,140 +23,74 @@
...
-  __initializer_iterator& operator++() noexcept
-  {
-    ++__id_;
-    return *this;
-  }
+//! @brief A random-access range of all available CUDA devices
+class __all_devices
+{
+public:
+  using value_type = ::cuda::std::span<const device_ref>::value_type;
```
fbusato, line 60: should not be `__value_type` now that it is internal only?

libcudacxx/include/cuda/__device/all_devices.h
```diff
@@ -23,140 +23,74 @@
...
-    return __tmp;
+    if (__i >= size())
+    {
+      ::cuda::std::__throw_out_of_range("device index out of range");
+    }
+    return ::cuda::__devices()[__i];
   }
 
-  bool operator==(const __initializer_iterator& __other) const noexcept
+  [[nodiscard]] _CCCL_HOST_API size_type size() const
```
fbusato, line 79: `noexcept`?

libcudacxx/include/cuda/__device/arch_traits.h
```diff
@@ -556,7 +558,7 @@ _CCCL_API inline constexpr id __special_id_for_compute_capability(int value)
 }
 
 //! @brief Provides architecture traits of the architecture matching __CUDA_ARCH__ macro
-[[nodiscard]] _CCCL_DEVICE inline constexpr arch::traits_t current_traits()
+[[nodiscard]] _CCCL_DEVICE_API inline constexpr arch::traits_t current_traits()
```
fbusato, line 561: `inline constexpr` is redundant

libcudacxx/include/cuda/__device/device_ref.h
```diff
@@ -103,30 +89,35 @@ class device_ref
...
     return __attr(*this);
   }
 
   //! @overload
   template <::cudaDeviceAttr _Attr>
-  [[nodiscard]] auto attribute() const
+  [[nodiscard]] _CCCL_HOST_API auto attribute() const
   {
-    return attribute(__detail::__dev_attr<_Attr>());
+    return attribute(__dev_attr<_Attr>());
```
fbusato, line 101: `__dev_attr<_Attr>{}` makes easier to understand that `__dev_attr` is class, not a function

libcudacxx/include/cuda/__device/physical_device.h
```diff
@@ -150,63 +146,47 @@ class physical_device : public device_ref
     });
     return ::cuda::std::span<const device_ref>{__peers_};
   }
+};
 
-private:
-  // TODO: put a mutable thread-safe (or thread_local) cache of device
-  // properties here.
+[[nodiscard]] _CCCL_HOST_API inline ::std::unique_ptr<__physical_device[]>
+__make_physical_devices(::cuda::std::size_t __device_count)
```
fbusato, line 152: not sure if `size_t` for `__device_count` is the best choice

Issue #6132, opened 2025-10-06
[BUG]: Compile-time error when using `device_delete` on a derived class


asmelko:
### Is this a duplicate?

- [x] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cccl/issues) for this bug and that I agree to the [Code of Conduct](CODE_OF_CONDUCT.md)

### Type of Bug

Compile-time Error

### Component

Thrust

### Describe the bug

Attempting to compile a program with `thrust::device_delete` that tries to deallocate an object of a derived class with a virtual destructor.

### How to Reproduce

0. See the godbolt link for the exact reproducer
1. Create a base class with a virtual destructor
2. Create a derived class
3. Assign return value of `device_new<derived>()` to `device_ptr<base> ptr`
4. Call `device_delete(ptr)`
5. Compilation fails with:
```
/opt/compiler-explorer/cuda/13.0.1/bin/../targets/x86_64-linux/include/cccl/thrust/detail/allocator/allocator_traits.h(217): error: class "thrust::THRUST_300001_SM_750_NS::detail::device_delete_allocator" has no member "value_type"
    using value_type = typename allocator_type::value_type;
```

### Expected behavior

Compilation should run to completion, and the virtual destructor of both `derived` and `base` objects should be called.

### Reproduction link

https://godbolt.org/z/68PssqfvY

### Operating System

Ubuntu 24.04.3 LTS

### nvidia-smi output
```
+-----------------------------------------------------------------------------------------+
| NVIDIA-SMI 580.82.10              Driver Version: 581.29         CUDA Version: 13.0     |
+-----------------------------------------+------------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id          Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |           Memory-Usage | GPU-Util  Compute M. |
|                                         |                        |               MIG M. |
|=========================================+========================+======================|
|   0  NVIDIA GeForce RTX 4090        On  |   00000000:01:00.0  On |                  Off |
| 30%   33C    P0             51W /  450W |    2004MiB /  24564MiB |      0%      Default |
|                                         |                        |                  N/A |
+-----------------------------------------+------------------------+----------------------+

+-----------------------------------------------------------------------------------------+
| Processes:                                                                              |
|  GPU   GI   CI              PID   Type   Process name                        GPU Memory |
|        ID   ID                                                               Usage      |
[...]

PR #6140, opened 2025-10-07
Test polymorphic types in `thrust::device_delete`

bernhardmgruber:
- [x] Merge before: #6198

Fixes: #6132

### DIFF
--- a/thrust/testing/device_delete.cu
+++ b/thrust/testing/device_delete.cu
@@ -10,7 +10,7 @@
 struct Foo
 {
   _CCCL_DEVICE Foo()
-      : set_me_upon_destruction{nullptr}
+      : destroyed{nullptr}
   {}
 
   Foo(const Foo&)            = delete;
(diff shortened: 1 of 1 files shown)


thrust/testing/device_delete.cu
```diff
@@ -21,24 +21,39 @@ struct Foo
   bool* set_me_upon_destruction;
 };
 
-#if !defined(__QNX__)
```
bernhardmgruber: I have no idea whether we still have to skip the test on QNX, so let's try.

bernhardmgruber:
I am stuck at:
```
  /home/coder/cccl/lib/cmake/thrust/../../../thrust/thrust/detail/reference.h(171): error: function returning abstract class "base" is not allowed:
              function "base::f" is a pure virtual function
```
The code in question is:
```c++
  _CCCL_HOST_DEVICE operator value_type() const { ... }
```
I tried working around this with:
```c++
  template <typename VT = value_type>
  _CCCL_HOST_DEVICE operator VT() const
```
However, this breaks some unit tests using `ASSERT_EQUAL(false, destructor_flag[0]);`, where `destructor_flag[0]` is a `device_reference<bool>` and we expect the compiler to call the implicit conversion operator to make a `bool`, and only a `bool`.

I also tried concepts:
```
  _CCCL_HOST_DEVICE operator value_type() const requires (!::cuda::std::is_abstract_v<value_type>)
```
But this gives us the original error message that `value_type` is abstract.

Is there another way to constrain `operator value_type()`? Alternatives would be to lift the offending members into a base class and conditionally inherit them. Or partially specialize `thrust::reference`.

thrust/testing/device_delete.cu
```diff
@@ -9,36 +9,50 @@
...
+struct derived : public base
+{
+  _CCCL_HOST_DEVICE void f() override {}
+};
+
+void TestDeviceDeleteVirtualDestructorInvocation()
+{
+  thrust::device_ptr<base> ptr = thrust::device_new<derived>();
+  thrust::device_delete(ptr);
+}
```
asmelko, line 92: Can we also make sure that the base destructor was called here?

bernhardmgruber: Added.

bernhardmgruber:
I had to change the test slightly because Thrust really has troubles with abstract types. I therefore made base a concrete class.

miscco (approved):

PR #6077, opened 2025-09-30
[CUB] Use `BlockLoadToShared` in `DeviceMerge`

pauleonix:
## Description

closes #6005 (previous PR #5926 was wrongly closed)

Since `DeviceMerge` needs its data in shared memory either way, it is a good candidate to demonstrate the new `BlockLoadToShared` interface in the real world. It also comes with some complications since we know the size of the output tile at compile-time, but how many elements to load from each of the ranges to be merged (calculated at runtime using merge-path). This can be solved by introducing additional padding (size known at compile-time) for in between the two dynamically sized input buffers.

## Checklist
- [x] I am familiar with the [Contributing Guidelines]().
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #6142, opened 2025-10-07
Cache modify iterators locally in agent_merge.cuh

bernhardmgruber:
This avoids needing to specify the cache modifier in two places and eases integration of #6077

No relevant SASS changes on sm86 and sm120 for `cub.cpp17.test.device_merge.lid_0`. Just a few symbol changes and constants like:
```diff
<         /*0110*/                   HFMA2 R8, -RZ, RZ, 0, 6.258487701416015625e-06 ;           /* 0x00000069ff087431 */
---
>         /*0110*/                   HFMA2 R8, -RZ, RZ, 0, 5.900859832763671875e-06 ;           /* 0x00000063ff087431 */
```

### DIFF
--- a/cub/cub/agent/agent_merge.cuh
+++ b/cub/cub/agent/agent_merge.cuh
@@ -53,14 +53,8 @@ struct agent_t
   using policy = Policy;
 
   // key and value type are taken from the first input sequence (consistent with old Thrust behavior)
-  using key_type  = it_value_t<KeysIt1>;
-  using item_type = it_value_t<ItemsIt1>;
-
-  using keys_load_it1  = try_make_cache_modified_iterator_t<Policy::LOAD_MODIFIER, KeysIt1>;
-  using keys_load_it2  = try_make_cache_modified_iterator_t<Policy::LOAD_MODIFIER, KeysIt2>;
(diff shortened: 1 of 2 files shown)


cub/cub/agent/agent_merge.cuh
```diff
@@ -128,10 +122,14 @@ struct agent_t
     }
 
     key_type keys_loc[items_per_thread];
-    merge_sort::gmem_to_reg<threads_per_block, IsFullTile>(
-      keys_loc, keys1_in + keys1_beg, keys2_in + keys2_beg, keys1_count_tile, keys2_count_tile);
-    merge_sort::reg_to_shared<threads_per_block>(&storage.keys_shared[0], keys_loc);
-    __syncthreads();
+    {
+      auto keys1_in_cm = try_make_cache_modified_iterator<Policy::LOAD_MODIFIER>(keys1_in);
```
fbusato, line 126: nit. `cm` for "cache modifier` is a really short abbreviation and not very intuitive

bernhardmgruber: I agree, but the use of the variable is so local that I didn't bother with a long and descriptive name.

PR #6109, opened 2025-10-02
Add a philox PRNG engine

RAMitchell:
Implements part of https://github.com/NVIDIA/cccl/issues/5679

This implementation follows the C++ standard https://en.cppreference.com/w/cpp/numeric/random/philox_engine/set_counter.html

I will additionally implement an optimised discard operator for this engine for use in parallel.

This engine depends on a `mulhilo` operation for good performance. We will have to implement a few versions of this for different platforms.

PR #6146, opened 2025-10-07
Expose `cuda::multiply_half_high`

fbusato:
## Description

Motivated by [Add a philox PRNG engine to thrust](https://github.com/NVIDIA/cccl/pull/6109) #6109.

Multiply and keep the high half of the result is a common primitive in random generator, cryptography, and modulo arithmetic.
We already use the functionality in `fast_mod_div`.
This PR exposes the functionality, improving a few aspects, and adding a fast windows path for 64-bit operands.

### DIFF
--- a/docs/libcudacxx/extended_api/math/multiply_high_half.rst
+++ b/docs/libcudacxx/extended_api/math/multiply_high_half.rst
@@ -1,0 +1,68 @@
+.. _libcudacxx-extended-api-math-multiply-high-half:
+
+``cuda::multiply_half_high``
+============================
+
+Defined in ``<cuda/cmath>`` header.
+
+.. code:: cuda
+
+   namespace cuda {
+
+   template <typename T>
+   [[nodiscard]] constexpr
+   T multiply_half_high(T lhs, T rhs) noexcept;
+
+   } // namespace cuda
+
+Computes the high half of the product of two non-negative integers ``lhs`` and ``rhs``.
+
+**Parameters**
+
+- ``lhs``: First multiplicand.
+- ``rhs``: Second multiplicand.
+
+**Return value**
+
+Upper half of ``lhs * rhs`` returned as ``T``.
+
+**Constraints**
+
+- ``T`` is an integer type.
+
+**Preconditions**
+
+- ``lhs >= 0`` when ``T`` is signed.
+- ``rhs >= 0`` when ``T`` is signed.
+
+**Remarks**
+
+- Uses ``__umulhi`` or ``__umul64hi`` instructions on device when available.
+- Uses a double-width intermediate type when possible.
+- Relies on a manual decomposition fallback when 128-bit intermediates are unavailable.
+
+Example
+-------
+
+.. code:: cuda
+
+   #include <cuda/cmath>
+   #include <cuda/std/cassert>
+   #include <cuda/std/cstdint>
+
+   __global__ void multiply_high_kernel()
+   {
+       uint32_t lhs = 0xABCD1234;
+       uint32_t rhs = 1 << 16;
+       uint32_t high_half = cuda::multiply_half_high(lhs, rhs);
+       assert(high_half == 0xAB);
+   }
+
+   int main()
+   {
+       multiply_high_kernel<<<1, 1>>>();
+       cudaDeviceSynchronize();
+       return 0;
+   }
+
+`See it on Godbolt 🔗 <https://godbolt.org/z/dPPzsEaGM>`_
--- a/docs/libcudacxx/extended_api/math/multiply_high_half.rst
+++ b/docs/libcudacxx/extended_api/math/multiply_high_half.rst
@@ -52,8 +52,8 @@ Example
 
    __global__ void multiply_high_kernel()
    {
-       uint32_t lhs = 0xABCD1234;
-       uint32_t rhs = 1 << 16;
+       uint32_t lhs       = 0xABCD1234;
+       uint32_t rhs       = 1 << 16; // 2^16
        uint32_t high_half = cuda::multiply_half_high(lhs, rhs);
        assert(high_half == 0xAB);
    }
--- a/docs/libcudacxx/extended_api/math.rst
+++ b/docs/libcudacxx/extended_api/math.rst
@@ -17,6 +17,7 @@ Math
    math/neg
    math/uabs
    math/fast_mod_div
+   math/multiply_half_high
 
 .. list-table::
    :widths: 25 45 30 30
@@ -91,3 +92,9 @@ Math
      - Fast Modulo/Division
      - CCCL 3.1.0
      - CUDA 13.1
+
+   * - :ref:`multiply_half_high <libcudacxx-extended-api-math-multiply-half-high>`
+     - Half high of the product
+     - CCCL 3.2.0
+     - CUDA 13.2
+     
--- a/docs/libcudacxx/extended_api/math.rst
+++ b/docs/libcudacxx/extended_api/math.rst
@@ -97,4 +97,3 @@ Math
      - Half high of the product
      - CCCL 3.2.0
      - CUDA 13.2
-     
--- a/libcudacxx/include/cuda/__cmath/multiply_high.h
+++ b/libcudacxx/include/cuda/__cmath/multiply_high.h
@@ -1,0 +1,110 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of libcu++, the C++ Standard Library for your entire system,
+// under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef _CUDA___CMATH_MULTIPLY_HIGH_HALF_H
+#define _CUDA___CMATH_MULTIPLY_HIGH_HALF_H
+
+#include <cuda/std/detail/__config>
+
+#if defined(_CCCL_IMPLICIT_SYSTEM_HEADER_GCC)
+#  pragma GCC system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_CLANG)
+#  pragma clang system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_MSVC)
+#  pragma system_header
+#endif // no system header
+
+#include <cuda/std/__type_traits/is_constant_evaluated.h>
+#include <cuda/std/__type_traits/is_integer.h>
+#include <cuda/std/__type_traits/is_signed.h>
+#include <cuda/std/__type_traits/make_nbit_int.h>
+#include <cuda/std/__type_traits/make_unsigned.h>
+#include <cuda/std/__type_traits/num_bits.h>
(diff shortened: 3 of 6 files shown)


docs/libcudacxx/extended_api/math/multiply_high_half.rst
```diff
@@ -0,0 +1,68 @@
...
+
+   namespace cuda {
+
+   template <typename T>
+   [[nodiscard]] constexpr
+   T multiply_half_high(T lhs, T rhs) noexcept;
+
+   } // namespace cuda
+
+Computes the high half of the product of two non-negative integers ``lhs`` and ``rhs``.
```
bernhardmgruber: Suggestion:
```suggestion
Computes the most significant half of the bits of the product of two non-negative integers ``lhs`` and ``rhs``.
```

docs/libcudacxx/extended_api/math/multiply_high_half.rst
```diff
@@ -0,0 +1,68 @@
...
+
+**Parameters**
+
+- ``lhs``: First multiplicand.
+- ``rhs``: Second multiplicand.
+
+**Return value**
+
+Upper half of ``lhs * rhs`` returned as ``T``.
```
bernhardmgruber: Suggestion: apply above suggestion here as well

docs/libcudacxx/extended_api/math/multiply_high_half.rst
```diff
@@ -0,0 +1,68 @@
...
+============================
+
+Defined in ``<cuda/cmath>`` header.
+
+.. code:: cuda
+
+   namespace cuda {
+
+   template <typename T>
+   [[nodiscard]] constexpr
```
davebayer:
```suggestion
   [[nodiscard]] __host__ __device__ constexpr
```

docs/libcudacxx/extended_api/math/multiply_high_half.rst
```diff
@@ -0,0 +1,68 @@
...
+   T multiply_half_high(T lhs, T rhs) noexcept;
```
davebayer: Uhh, can't we just name it `mul_hi`?

docs/libcudacxx/extended_api/math/multiply_high_half.rst
```diff
@@ -0,0 +1,68 @@
...
+Upper half of ``lhs * rhs`` returned as ``T``.
+
+**Constraints**
+
+- ``T`` is an integer type.
+
+**Preconditions**
+
+- ``lhs >= 0`` when ``T`` is signed.
+- ``rhs >= 0`` when ``T`` is signed.
```
davebayer: Why? We can just use `__mulhi` and `__mul64hi` for signed types. I'd like to have no preconditions for this function

docs/libcudacxx/extended_api/math/multiply_high_half.rst
fbusato: yes, I considered that. `multiply_half_high` is very long

docs/libcudacxx/extended_api/math/multiply_high_half.rst
fbusato: that's a good point. Signed types are rarely used for this kind of operation, but would be nice to not have restrictions

libcudacxx/test/libcudacxx/cuda/cmath/mul_hi.pass.cpp
```diff
@@ -0,0 +1,107 @@
...
+  assert(cuda::mul_hi(max_value, T{1}) == T{0});
+  assert(cuda::mul_hi(T{0}, max_value) == T{0});
+  assert(cuda::mul_hi(T{1}, max_value) == T{0});
+  if constexpr (cuda::std::is_signed_v<T>)
+  {
+    assert(cuda::mul_hi(T{-1}, T{0}) == T{0});
+    assert(cuda::mul_hi(T{-1}, T{-1}) == T{0});
+    assert(cuda::mul_hi(T{-1}, T{1}) == T{-1});
+    assert(cuda::mul_hi(T{1}, T{-1}) == T{-1});
+  }
```
davebayer, line 43: can we have some more advanced tests for multiplication of different arguments with different signs?

fbusato: there are, e.g. https://github.com/NVIDIA/cccl/pull/6146/files#diff-f04a3fffd8c5fb3b3121e56cb1db58cd5ea93fd5641bfed76ccf156139b05a69R54
```cpp
if constexpr (sizeof(T) >= 2)
  {
    constexpr auto mask3 = static_cast<T>(0x7BCD);
    assert(cuda::mul_hi(mask3, T{4096}) == (mask3 >> (bits - 12)));
    assert(cuda::mul_hi(T{4096}, mask3) == (mask3 >> (bits - 12)));
  }
```
but I can add more complex cases

RAMitchell:
Thank you for updating this, I will switch over to this in my PR when this gets merged.

Btw this is what GCC does https://github.com/gcc-mirror/gcc/blob/d2ad7e90834d9c18e0c7104796f3131594e7bbfb/libstdc%2B%2B-v3/include/bits/random.tcc#L916

RAMitchell:
One curious thing I noticed as philox uses std::uint_fast32_t:
```
// std::uint_fast32_t is actually 64 bits and this performs __umul64hi()
mul_hi(std::uint_fast32_t(a), std::uint_fast32_t(b))
```

Not sure if you should do anything about that.

libcudacxx/include/cuda/__cmath/mul_hi.h
```diff
@@ -0,0 +1,138 @@
...
+// Part of libcu++, the C++ Standard Library for your entire system,
+// under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef _CUDA___CMATH_MULTIPLY_HIGH_HALF_H
+#define _CUDA___CMATH_MULTIPLY_HIGH_HALF_H
```
davebayer:
```suggestion
#ifndef _CUDA___CMATH_MUL_HI_H
#define _CUDA___CMATH_MUL_HI_H
```
detail..

libcudacxx/include/cuda/__cmath/mul_hi.h
```diff
@@ -0,0 +1,138 @@
...
+{
+  using ::cuda::std::int32_t;
+  using ::cuda::std::int64_t;
+  using ::cuda::std::is_same_v;
+  using ::cuda::std::is_signed_v;
+  using ::cuda::std::uint32_t;
+  using ::cuda::std::uint64_t;
+  if (!::cuda::std::__cccl_default_is_constant_evaluated())
+  {
+    if constexpr (is_same_v<_Tp, int32_t>)
```
davebayer:
```suggestion
    if constexpr (sizeof(_Tp) == sizeof(int32_t)>)
```
We wouldn't use the builtins for `long`, applies to the rest of the cases as well

fbusato: uhmm, probably the best solution is to use `is_signed_v<_Tp> && sizeof(_Tp) ==  sizeof(int32_t))`

fbusato:
@RAMitchell I think we never used `std::uint_fast32_t` in our codebase. Anyway, I generalized the implementation in the last commit to dispatch to intrinsics even with non-fixed width types.

libcudacxx/include/cuda/__cmath/mul_hi.h
```diff
@@ -0,0 +1,146 @@
...
+  auto __lhs_high           = static_cast<__half_bits_t>(__lhs >> __half_bits); // 32-bit
+  auto __rhs_low            = static_cast<__half_bits_t>(__rhs); // 32-bit
+  auto __rhs_high           = static_cast<__half_bits_t>(__rhs >> __half_bits); // 32-bit
+  auto __po_half            = (static_cast<_Tp>(__lhs_low) * __rhs_low) >> __half_bits;
+  auto __p1                 = static_cast<_Tp>(__lhs_low) * __rhs_high; // 64-bit
+  auto __p2                 = static_cast<_Tp>(__lhs_high) * __rhs_low; // 64-bit
+  auto __p3                 = static_cast<_Tp>(__lhs_high) * __rhs_high; // 64-bit
+  auto __p1_half            = static_cast<__half_bits_t>(__p1); // 32-bit
+  auto __p2_half            = static_cast<__half_bits_t>(__p2); // 32-bit
+  auto __carry              = (__po_half + __p1_half + __p2_half) >> __half_bits; // 64-bit
```
miscco: Nitpick: could all be `const`

libcudacxx/include/cuda/__cmath/mul_hi.h
```diff
@@ -0,0 +1,146 @@
...
+  using ::cuda::std::int64_t;
+  using ::cuda::std::is_signed_v;
+  if (!::cuda::std::__cccl_default_is_constant_evaluated())
+  {
+    if constexpr (sizeof(_Tp) == sizeof(int))
+    {
+      if constexpr (is_signed_v<_Tp>)
+      {
+        const auto __lhs1 = static_cast<int>(__lhs);
+        const auto __rhs1 = static_cast<int>(__rhs);
```
miscco: We might need to mark those as `[[maybe_unused]]` because the host compiler will not see `__mulhi`

docs/libcudacxx/extended_api/math/mul_hi.rst
```diff
@@ -0,0 +1,64 @@
...
+- The most significant half of ``lhs * rhs`` returned as ``T``.
+
+**Constraints**
+
+- ``T`` is an integer type.
+
+**Remarks**
+
+- Uses ``__mulhi``, ``__umulhi``, ``__mul64hi``, ``__umul64hi`` instructions on device when available.
+- Uses ``__mulh``, ``__umulh`` instructions on Windows host code when available.
```
davebayer:
```suggestion
- Uses ``__mulhi``, ``__umulhi``, ``__mul64hi``, ``__umul64hi`` intrinsics on device when available.
- Uses ``__mulh``, ``__umulh`` intrinsics on Windows host code when available.
```

libcudacxx/test/libcudacxx/cuda/cmath/mul_hi.pass.cpp
```diff
@@ -0,0 +1,140 @@
...
+  else if constexpr (sizeof(T) == 8)
+  {
+    assert(cuda::mul_hi(T{2'305'878'454'534}, T{31'852'876'355'863}) == T{0x3CC166});
+    if constexpr (cuda::std::is_same_v<T, int64_t>)
+    {
+      assert(cuda::mul_hi(T{-2'305'878'454'534}, T{-31'852'876'355'863}) == T{0x3CC166});
+      assert(cuda::mul_hi(T{-2'305'878'454'534}, T{31'852'876'355'863}) == T{-0x3CC167});
+      assert(cuda::mul_hi(T{2'305'878'454'534}, T{-31'852'876'355'863}) == T{-0x3CC167});
+    }
```
davebayer, line 112: Can we add int128 tests? You can use the `_i128`/`_u128` literals from `test_integer_literals` namespace in `"literal.h"`

PR #6150, opened 2025-10-08
Refactor `cuda::arch_traits`

davebayer:
This PR refactors the CUDA architecture traits.

1. the `cuda::arch::` namespace is removed and the functions and types are prefixed with `arch_` directly
2. the `cuda::current_arch_traits()` was moved to `cuda::device::` namespace, because it's device-only function
3. the `cuda::device::current_arch_id()` function was added

### DIFF
--- a/libcudacxx/include/cuda/__device/arch_traits.h
+++ b/libcudacxx/include/cuda/__device/arch_traits.h
@@ -296,8 +296,11 @@ template <>
 template <>
 [[nodiscard]] _CCCL_API constexpr arch_traits_t arch_traits<arch_id::sm_88>() noexcept
 {
-  auto __traits    = ::cuda::arch_traits<arch_id::sm_86>();
-  __traits.arch_id = arch_id::sm_88;
+  auto __traits                     = ::cuda::arch_traits<arch_id::sm_86>();
+  __traits.arch_id                  = arch_id::sm_88;
+  __traits.compute_capability_major = 8;
+  __traits.compute_capability_minor = 8;
+  __traits.compute_capability       = compute_capability{88};
   return __traits;
 };
 


libcudacxx/include/cuda/__device/arch_traits.h
```diff
@@ -408,50 +407,50 @@ template <>
...
   __traits.compute_capability       = 110;
   return __traits;
 };
 
 template <>
-[[nodiscard]] _CCCL_API inline constexpr traits_t traits<id::sm_110a>()
+[[nodiscard]] _CCCL_API inline constexpr arch_traits_t arch_traits<arch_id::sm_110a>() noexcept
 {
-  return ::cuda::arch::traits<id::sm_110>();
+  return ::cuda::arch_traits<arch_id::sm_110>();
```
davebayer: @pciolkosz Shouldn't we overwrite the returned `.arch_id`?

libcudacxx/include/cuda/__device/arch_traits.h
```diff
@@ -604,12 +579,69 @@ __arch_traits_might_be_unknown(int __device, unsigned int __compute_capability)
...
+#    endif // ^^^ __CUDA_ARCH_SPECIFIC__ ^^^
+}
+
+extern "C" _CCCL_DEVICE_API ::cuda::arch_traits_t __unknown_architecture_cannot_be_queried_for_arch_traits();
+
+//! @brief Returns the \c cuda::arch_trait_t of the architecture that is currently being compiled.
+//!
+//! @note If the current architecture is unknown (too new/old), the compilation will fail.
+template <class _Dummy = void>
+[[nodiscard]] _CCCL_DEVICE_API inline ::cuda::arch_traits_t current_arch_traits() noexcept
```
davebayer: I've made the function non-`constexpr`. As was described in the test file, it doesn't fully serve it's purpose due to missing function signatures. The advantage is that now we can be compatible with nvhpc in CUDA mode.

My idea is that we should try to go without `constexpr` and if users complain, we should add `constexpr` if possible (non nvhpc compilers).

If you don't agree with this, we can make it `constexpr` now, but then we will allow users to write code that is not portable among all of the CUDA compilers. And as I said, we can always add it later.

libcudacxx/test/libcudacxx/cuda/ccclrt/device/arch_traits.c2h.cu
```diff
@@ -17,55 +18,62 @@ __device__ int foo(const int& x)
...
-        asm volatile("mov.u32 %0, %%cluster_ctarank;" : "=r"(dummy));
-      } if constexpr (cuda::arch::current_traits().redux_intrinisic) {
-        [[maybe_unused]] int dummy1 = 0, dummy2 = 0;
-        asm volatile("redux.sync.add.s32 %0, %1, 0xffffffff;" : "=r"(dummy1) : "r"(dummy2));
-      } if constexpr (cuda::arch::current_traits().cp_async_supported) { asm volatile("cp.async.commit_group;"); }
-
-      // Confirm trait value is defined device code and usable as a reference
-      foo(cuda::arch::traits<Arch>().compute_capability);
-      foo(cuda::arch::current_traits().compute_capability);))
+  assert(Arch == cuda::device::current_arch_id());
```
davebayer, line 25: todo: make it `static_assert` if we decide to go with `constexpr`

libcudacxx/test/libcudacxx/cuda/ccclrt/device/arch_traits.c2h.cu
```diff
@@ -17,55 +18,62 @@ __device__ int foo(const int& x)
...
+  // I will try to pack something like this into an API
+  if (cuda::arch_traits<Arch>().compute_capability != cuda::device::current_arch_traits().compute_capability)
+  {
+    return;
+  }
+
+  [[maybe_unused]] __shared__ int array[cuda::arch_traits<Arch>().max_shared_memory_per_block / sizeof(int)];
+
+  // constexpr is useless and I can't use intrinsics here :(
+  if (cuda::device::current_arch_traits().cluster_supported)
```
davebayer, line 36: todo: make it `if constexpr` if we decide to go with `constexpr`

libcudacxx/include/cuda/__device/arch_traits.h
```diff
@@ -472,118 +471,94 @@ template <>
...
+    case arch_id::sm_110:
+      return ::cuda::arch_traits<arch_id::sm_110>();
+    case arch_id::sm_110a:
+      return ::cuda::arch_traits<arch_id::sm_110a>();
+    case arch_id::sm_120:
+      return ::cuda::arch_traits<arch_id::sm_120>();
+    case arch_id::sm_120a:
+      return ::cuda::arch_traits<arch_id::sm_120a>();
     default:
       ::cuda::__throw_cuda_error(::cudaErrorInvalidValue, "Traits requested for an unknown architecture");
```
davebayer, line 494: Do we want to throw `cuda_error` here? Don't we want to use `cuda::std::invalid_argument` instead?

libcudacxx/include/cuda/__device/arch_traits.h
pciolkosz: Oops, yes we should across all the `sm_XXa`, thanks for spotting that

davebayer: I'll fix it in this PR

libcudacxx/include/cuda/__device/arch_traits.h
pciolkosz: Technically nvhpc CUDA compilation is not officially supported, I am not sure we should make the nvcc case worst because of something we don't support.
But at the same time it is true that the constexpr traits are not as useful as I hoped, so maybe we don't lose much.
The question I would ask is: Later we can add a second API that serves one of those cases, would it be better to special case the `constexpr` traits getter or to special case traits getter for nvhpc. I thought it made more sense to special case for nvhpc, but I am fine either way. Maybe we could provide both cases already?

libcudacxx/include/cuda/__device/arch_traits.h
```diff
@@ -181,16 +180,16 @@ struct traits_t
...
-template <id _Id>
-[[nodiscard]] _CCCL_API constexpr traits_t traits();
+//! @brief Architecture traits
+//! Template function that returns the traits for an architecture with a given id.
+template <arch_id _Id>
+[[nodiscard]] _CCCL_API constexpr arch_traits_t arch_traits() noexcept;
 
 template <>
-[[nodiscard]] _CCCL_API inline constexpr traits_t traits<id::sm_60>()
+[[nodiscard]] _CCCL_API inline constexpr arch_traits_t arch_traits<arch_id::sm_60>() noexcept
```
fbusato: maybe I'm overthinking, but all these template specializations are compiled for all TUs that include this header. Could not be a concern?

libcudacxx/include/cuda/__device/arch_traits.h
```diff
@@ -472,118 +471,94 @@ template <>
...
+  return ::cuda::arch_traits<arch_id::sm_120>();
 };
 
+inline constexpr int __lowest_known_arch  = 60;
 inline constexpr int __highest_known_arch = 120;
 
-[[nodiscard]] _CCCL_API inline constexpr traits_t traits_for_id(id __id)
+[[nodiscard]] _CCCL_API inline constexpr bool __is_known_compute_capability(int __cc)
+{
+  return __cc >= __lowest_known_arch && __cc <= __highest_known_arch;
```
fbusato: `::cuda::in_range` 😅 when ready

libcudacxx/include/cuda/__device/arch_traits.h
```diff
@@ -472,118 +471,94 @@ template <>
...
       ::cuda::__throw_cuda_error(::cudaErrorInvalidValue, "Traits requested for an unknown architecture");
       break;
   }
 }
 
-[[nodiscard]] _CCCL_API inline constexpr id id_for_compute_capability(int compute_capability)
+[[nodiscard]] _CCCL_API inline constexpr arch_id arch_id_for_compute_capability(int __cc)
 {
-  if (compute_capability < 60 || compute_capability > __highest_known_arch)
+  if (!::cuda::__is_known_compute_capability(__cc))
```
fbusato: `__is_known_compute_capability` checks the range but not the value, e.g. sm109 is not valid but it is in range. The name is misleading and the behavior could be more accurate

libcudacxx/include/cuda/__device/arch_traits.h
davebayer: I know it's not officially supported now, but the thing is that it might change in the future. And users will have already their code written in a way that is not compatible with nvhpc.

I'm not against adding another API that would be `constexpr`, but I would like to make this the default.

And I believe that if you use the `cuda::arch_trait().property` in an `if (...)`, I believe it will get optimized away, so we would be only paying the additional compilation cost.

libcudacxx/include/cuda/__device/arch_traits.h
```diff
@@ -472,118 +471,94 @@ template <>
...
-  return ::cuda::arch::traits_for_compute_capability(__CUDA_ARCH__ / 10);
-#    endif // __CUDA_ARCH_SPECIFIC__
-#  else // __CUDA_ARCH__
-  // Should be unreachable in __device__ function
-  return ::cuda::arch::traits_t{};
-#  endif // __CUDA_ARCH__
-}
-
-[[nodiscard]] _CCCL_HOST_API inline constexpr arch::traits_t
-__arch_traits_might_be_unknown(int __device, unsigned int __compute_capability)
```
fbusato, line 580: +1, prefer signed integer where possible.

libcudacxx/include/cuda/__device/arch_traits.h
```diff
@@ -604,12 +579,69 @@ __arch_traits_might_be_unknown(int __device, unsigned int __compute_capability)
...
+#  endif // !_CCCL_COMPILER(NVRTC)
 
 _CCCL_END_NAMESPACE_CUDA
 
+#  if _CCCL_CUDA_COMPILATION()
+
+_CCCL_BEGIN_NAMESPACE_CUDA_DEVICE
+
+//! @brief Returns the compute capability of the architecture that is currently being compiled.
+[[nodiscard]] _CCCL_DEVICE_API inline int __current_arch_cc() noexcept
```
fbusato: this could be `constexpr`, not sure for nvc++

libcudacxx/include/cuda/__device/arch_traits.h
pciolkosz: Ok, I am fine with that, we just need to figure out the name for the `constexpr` version then

davebayer: What about `current_arch_traits_static` or `current_static_arch_traits` (similar to `static_extents`)?

libcudacxx/include/cuda/__device/arch_traits.h
davebayer: Not if the specializations are `inline`

libcudacxx/include/cuda/__device/arch_traits.h
davebayer: That's the point, it cannot be `constexpr` with nvc++ in CUDA mode..

fbusato: as we don't officially support nvc++ in CUDA mode, would be too crazy to disable `constexpr` only in this mode?

libcudacxx/include/cuda/__device/arch_traits.h
davebayer: I've introduced a `cuda::compute_capability` type

libcudacxx/include/cuda/__device/arch_traits.h
davebayer: I've added a `__current_arch_cc_constexpr()` function

libcudacxx/include/cuda/__device/physical_device.h
```diff
@@ -188,6 +188,13 @@ _CCCL_HOST_API inline void device_ref::init() const
   return ::cuda::__physical_devices()[__id_].__peers();
 }
 
+// arch_traits functions dependent on __physical_device
+
+[[nodiscard]] _CCCL_HOST_API inline const arch_traits_t& arch_traits_for(device_ref __device)
```
pciolkosz: Do we need both this and `device_ref::arch_traits()`? We could just keep one of them

libcudacxx/include/cuda/__device/arch_traits.h
pciolkosz: I don't mind

davebayer:
pre-commit.ci autofix

libcudacxx/include/cuda/__device/arch_traits.h
davebayer: Fixed

davebayer:
We discussed changes in this PR with @pciolkosz and we decided to remove `.arch_traits()` from `device_ref` for now. Also, I gave up on making the `cuda::device::current_mewo()` functions non `constexpr`, so everything is simplified.

libcudacxx/include/cuda/__device/arch_id.h
```diff
@@ -0,0 +1,176 @@
...
+  sm_87   = 87,
+  sm_88   = 88,
+  sm_89   = 89,
+  sm_90   = 90,
+  sm_100  = 100,
+  sm_103  = 103,
+  sm_110  = 110,
+  sm_120  = 120,
+  sm_121  = 121,
+  sm_90a  = 90 * __arch_specific_id_multiplier,
```
fbusato, line 53: for my personal knowledge, how we support `f` variants? e.g. `sm_110f`

davebayer: This is a question for @pciolkosz, we can add them in a separate PR, I need to merge this now

PR #6151, opened 2025-10-08
Use `size_t` for byte count in device attributes

davebayer:


fbusato:
I'm not against it, but I would like to understand the motivation behind using size_t. The shared memory space is 32-bit; I'm not sure it's worth moving the attributes to 64-bit. I'm thinking of use cases where users experience worse performance (on the device) due to implicit promotion.

davebayer:
> I'm not against it, but I would like to understand the motivation behind using size_t. The shared memory space is 32-bit; I'm not sure it's worth moving the attributes to 64-bit. I'm thinking of use cases where users experience worse performance (on the device) due to implicit promotion.

Whenever you have number of bytes in C++, it's data type is `size_t`, the type returned by `sizeof` operator. I believe we should be consistent with what's commonly used.

I understand the point about performance, but I don't believe anyone would use any of these parameters in performance critical path, but tell me if I'm wrong :)

And the returned values are known at compile time so I don't believe there will be any overhead in the end

PR #6153, opened 2025-10-08
Fix typos in segmented reduce

oleksandr-pavlyk:
## Description

Edited some comments in dispatch_reduce to fix typos in comments in segmented reduction code.
Also moved `segmented_reduce_config` construction to debug-guarded section and removed `[[maybe_unsued]]` attribute.

Corrected typos in doxygen strings of `DeviceSegmentedReduce` methods where reference to non-existing `d_keys_*` and `d_values_*` were used for algorithms with no such parameters in the signature.

closes 

## Checklist

### DIFF
--- a/cub/cub/device/device_segmented_reduce.cuh
+++ b/cub/cub/device/device_segmented_reduce.cuh
@@ -156,14 +156,14 @@ struct DeviceSegmentedReduce
   //!   @rst
   //!   Random-access input iterator to the sequence of beginning offsets of
   //!   length ``num_segments``, such that ``d_begin_offsets[i]`` is the first
-  //!   element of the *i*\ :sup:`th` data segment in ``d_keys_*`` and ``d_values_*``
+  //!   element of the *i*\ :sup:`th` data segment in ``d_in``
   //!   @endrst
   //!
   //! @param[in] d_end_offsets
   //!   @rst
   //!   Random-access input iterator to the sequence of ending offsets of length
   //!   ``num_segments``, such that ``d_end_offsets[i] - 1`` is the last element of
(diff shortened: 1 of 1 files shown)


cub/cub/device/dispatch/dispatch_reduce.cuh
```diff
@@ -809,15 +809,6 @@ struct DispatchSegmentedReduce
...
 
-      // Init kernel configuration
-      [[maybe_unused]] detail::KernelConfig segmented_reduce_config;
-      error =
-        CubDebug(segmented_reduce_config.Init(segmented_reduce_kernel, policy.SegmentedReduce(), launcher_factory));
-      if (cudaSuccess != error)
-      {
-        break;
-      }
-
```
miscco, line 820: I believe that is not correct. With your code change we would not call `segmented_reduce_config.Init` if `CUB_DEBUG_LOG` is not defined

oleksandr-pavlyk: Yes, it was unused otherwise and `[[maybe_unused]]` was used to confirm that.

Does `segmented_reduce_config` have side effects?

oleksandr-pavlyk: @miscco Yes, this was the intent of this change. We used to call `detail::KernelConfig::Init` but only used struct members it initialized in code sections fenced in `#ifdef CUB_DEBUG_LOG`.

The proposed change shaves off the host-side execution of segmented reduce device-wide algorithm execution.

oleksandr-pavlyk: This change would only be externally observable to library user if the call to `detail::KernelConfig::Init` used to return non-success error code, which should not nominally happen.

Issue #6156, opened 2025-10-08
[BUG]: error: namespace "thrust" has no member "host_vector"


zkhatami:
### Is this a duplicate?

- [x] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cccl/issues) for this bug and that I agree to the [Code of Conduct](CODE_OF_CONDUCT.md)

### Type of Bug

Compile-time Error

### Component

Thrust

### Describe the bug

This test began failing at compile-time:
```
#include <execution>
#include <thrust/host_vector.h>
int main() {
  thrust::host_vector<int> hv;
}
```
It fails with this error:
```
thrust/mr/disjoint_pool.h", line 180: error: namespace "thrust" has no member "host_vector"
    using chunk_vector = thrust::host_vector<chunk_descriptor, allocator<chunk_descriptor, Bookkeeper>>;
```
I suspect this commit is the root cause:  50c7731bc8a0dd3711c8e973cc0cb04f541d4a28

### How to Reproduce

nvc++ -stdpar --c++17 test_host_vector.cpp

### Expected behavior

After the fix, it should compile fine.

zkhatami:
@davebayer could you please take a look at this issue? Thanks!

dkolsen-pgi:
> I suspect this commit is the root cause: https://github.com/NVIDIA/cccl/commit/50c7731bc8a0dd3711c8e973cc0cb04f541d4a28

In particular, the cause is most likely this line in `libcudacxx/include/cuda/__device/physical_device.h`:
```
#  include <memory>
```
See the file `thrust/thrust/detail/memory_wrapper.h` (and `algorithm_wrapper.h` and `numeric_wrapper.h`) and the uses of those files for more information.

alliepiper:
We should probably run some of these checks on libcudacxx as well (we currently have versions of this check in Thrust and CUB): https://github.com/NVIDIA/cccl/blob/ba9e12af08c7e4fc184de67a0f5d8b98715e564d/thrust/testing/cmake/check_source_files.cmake#L52-L61

davebayer:
Oops, sorry, I didn't know, should be fixed by #6165

alliepiper:
#6167 moves the wrapper headers to libcudacxx and sets up the `check_source_files` scans for it.

bernhardmgruber:
Is this solved now?

davebayer:
Yes, fixed by #6167

PR #6167, opened 2025-10-09
Move nvhpc header wrappers to libcudacxx.

alliepiper:
Also sets up the `check_source_files` test to catch uses of the unwrapped headers before they land.

Fixes: #6156


libcudacxx/include/cuda/std/__cccl/memory_wrapper.h
```diff
@@ -0,0 +1,34 @@
...
+
+// When nvc++ uses CCCL components as part of its implementation of
+// Standard C++ algorithms, a cycle of included files may result when CCCL code
+// tries to use a standard algorithm. The THRUST_INCLUDING_ALGORITHMS_HEADER macro
+// is defined only when CCCL is including an algorithms-related header, giving
+// the compiler a chance to detect and break the cycle of includes.
+
+#define THRUST_INCLUDING_ALGORITHMS_HEADER
+#include <memory>
+#undef THRUST_INCLUDING_ALGORITHMS_HEADER
```
miscco, line 32: Those should be guarded against NVRTC

bernhardmgruber: Addressed by: #6195

PR #5941, opened 2025-09-19
Implement `cudax::cufile_driver`

davebayer:

PR #6168, opened 2025-10-09
Add version_compare script, minor build_common updates.

alliepiper:
Added to simplify some logic in #5941.

```
Usage: ci/util/version_compare.sh X.Y[.Z[.W[...]]] <compare_op> A.B[.C[.D[...]]]

Compares two version strings with the specified operator.

compare_ops:
  lt - less than
  le - less than or equal to
  eq - equal to
  ne - not equal to
  ge - greater than or equal to
  gt - greater than
```

Also:
- Restore `NVCC_VERSION` parsing in build_common.sh
- Print sccache version in build_common.sh.

### DIFF
--- a/ci/build_common.sh
+++ b/ci/build_common.sh
@@ -113,6 +113,13 @@ CUDA_COMPILER=$(validate_and_resolve_compiler "CUDA compiler" "${CUDA_COMPILER}"
 
 if [[ "$(basename "$CUDA_COMPILER")" == nvcc* ]]; then
     NVCC_VERSION=$("$CUDA_COMPILER" --version | grep "release" | sed 's/.*, V//')
+    # Verify that we have an X.Y.Z version in case the output format changes:
+    if ! [[ "$NVCC_VERSION" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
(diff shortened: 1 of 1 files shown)


ci/build_common.sh
```diff
@@ -111,6 +111,11 @@ function validate_and_resolve_compiler() {
 HOST_COMPILER=$(validate_and_resolve_compiler "Host compiler" "${HOST_COMPILER}")
 CUDA_COMPILER=$(validate_and_resolve_compiler "CUDA compiler" "${CUDA_COMPILER}")
 
+if [[ "$(basename "$CUDA_COMPILER")" == nvcc* ]]; then
+    NVCC_VERSION=$("$CUDA_COMPILER" --version | grep "release" | sed 's/.*, V//')
```
davebayer, line 115: Q: Don't you want to add some kind of check that the version is >= 12.0 and < 14.0 for case that the format gets changed for some unknown reason?

alliepiper: We'd have to parse the version first to know that :)

If that happens, we'll deal with it then. Hopefully the compiler team won't be messing with this too much. I'll add a check that ensures that we end up with an X.Y.Z string for now.

PR #6170, opened 2025-10-09
Remove the lite PR matrix pending more testing.

alliepiper:
There are issues with some of these jobs. Removing them all to unblock CI.


alliepiper:
Issues are fixed in https://github.com/NVIDIA/cccl/pull/6169

PR #6175, opened 2025-10-10
Fixing cudax::execution CUDA stream scheduler

### LINKED ISSUE #5564: [BUG]: CUDA stream scheduler in cudax execution library is broken
### Is this a duplicate?

### Type of Bug

Runtime Error

### Component

CUDA Experimental (cudax)

### Describe the bug

due to design flaws in the sender algorithm customization scheme, the transitions between the CPU and GPU are not always orchestrated correctly by the CUDA stream scheduler in cudax, leading to hangs or crashes; hence, its tests have been disabled for some time which i come to grips with the issue. a fix must be made quickly to `std::execution` for C++26.

here are my current design thoughts/directions:

* now that `ensure_started` and `split` have been removed, there isn't much argument anymore for early customization.
* we have `get_scheduler` and `get_completion_scheduler<SetTag>`, and we have `get_domain` but no `get_completion_domain<SetTag>`. i think this is an oversight.
* `get_completion_[scheduler|domain]<SetTag>` needs the receiver's environment in order to properly answer the query. `just()` can only know where it will complete when it knows where it is started.
* although not strictly necessary, it would be helpful to adopt [P3206](https://wg21.link/P3206), "A sender query for completion behaviour". if a sender is known to complete inline, then its completion scheduler/domain is the scheduler/domain on which it is started.
* every sender has _two_ domains: the one it is started on, and the one it completes on, and they could be different. in `connect` and `get_completion_signatures` we can know both. the question is how to use them.

the last bullet is the most interesting. i can imagine a scheduler wanting to do something special, say, whenever a `foo` sender is started, and i can imagine another scheduler wanting to do something special when a `foo` sender completes. this suggest to me that `transform_sender` might need to apply _two_ transforms to a sender in `connect`, one for each domain (if they are different). Q: does it matter which order they are applied?

so a given domain might want to provide two different transforms for each sender: the "start" transform and the "complete" transform. if we had such a thing, we no longer need `schedule_from` to be a different algorithm from `continues_on`. domain `A` can provide a "start" `continues_on` transform for transfers _off of_ context `A`, and domain `B` can provide a "complete" `continues_on` transform for transfers _onto_ context `B`.

my intention is to implement this design in cudax and then update [P3718](https://wg21.link/P3718) for (fingers crossed) inclusion in C++26.

### How to Reproduce

n/a

### Expected behavior

n/a

### Reproduction link

_No response_

### Operating System

_No response_

### nvidia-smi output

_No response_

### NVCC version

_No response_

ericniebler:
this pr ports cudax's stream scheduler to the new sender algorithm customization framework.

fixes #5564

### DIFF
--- a/cudax/test/execution/test_stream_context.cu
+++ b/cudax/test/execution/test_stream_context.cu
@@ -8,8 +8,6 @@
 //
 //===----------------------------------------------------------------------===//
 
-// BUGBUG BUGBUG
-
 // Include this first
 #include <cuda/experimental/execution.cuh>
 


cudax/include/cuda/experimental/__container/async_buffer.cuh
```diff
@@ -220,6 +228,27 @@ public:
       : __buf_(__resource, __stream, __size)
   {}
 
+  //! @brief Constructs a async_buffer of size \p __size using a memory resource and initializes all elements to
+  //! \p __initial_value.
+  //! @param __env The environment used to query the memory resource.
+  //! @param __size The size of the async_buffer.
+  //! @param __value The initial value for all elements in the async_buffer.
+  //! @warning This constructor does *NOT* initialize any elements. It is the user's responsibility to ensure that the
+  //! elements within `[vec.begin(), vec.end())` are properly initialized, e.g with `cuda::std::uninitialized_copy`.
```
ericniebler:
```suggestion
```

cudax/include/cuda/experimental/__execution/stream/sequence.cuh
```diff
@@ -29,18 +29,18 @@
...
+// template <>
+// struct stream_domain::__apply_t<sequence_t>
+// {
+//   template <class _Sndr, class _Env>
+//   _CCCL_API auto operator()(_Sndr __sndr, const _Env& __env) const
+//   {
+//     static_assert(::cuda::std::__always_false_v<_Sndr>,
+//                   "The CUDA stream scheduler does not yet support the 'sequence' algorithm.");
+//   }
+// };
```
miscco, line 43: That looks like it was accidental

cudax/include/cuda/experimental/__execution/stream/starts_on.cuh
```diff
@@ -39,67 +39,65 @@ namespace cuda::experimental::execution
...
 } // namespace __stream
 
-// Start work on the GPU
-template <>
-struct stream_domain::__apply_t<starts_on_t> : __stream::__starts_on_t
-{};
+// // Start work on the GPU
+// template <>
+// struct stream_domain::__apply_t<starts_on_t> : __stream::__starts_on_t
+// {};
```
miscco: Should we just remove that?

cudax/include/cuda/experimental/__container/async_buffer.cuh
```diff
@@ -220,6 +228,25 @@ public:
...
+  template <class _Env = ::cuda::std::execution::env<>>
+  _CCCL_HIDE_FROM_ABI explicit async_buffer(
+    ::cuda::stream_ref __stream,
+    __resource_t __resource,
+    const size_type __size,
+    const value_type __value,
+    [[maybe_unused]] const _Env& __env = {})
+      : __buf_(__resource, __stream, __size)
+  {
+    ::cuda::experimental::__fill_n<_Tp, !::cuda::mr::__is_device_accessible<_Properties...>>(
```
pciolkosz: We can't have a member function that calls a kernel due to the ABI namespace issues. We had to remove this constructor, but there is a corresponding `make_async_buffer` function

ericniebler: what is the ABI namespace issue? this code path is being tested, it seems to work.

miscco: That will launch a kernel.

If someone links libraries containing this function compiled with different architectures you get *fun*

cudax/include/cuda/experimental/__execution/stream/sequence.cuh
ericniebler: it wasn't accidental. by commenting out this specialization, `sequence` uses its default implementation on the stream scheduler instead of `static_assert`-ing. the default implementation works on the stream scheduler, but it isn't optimal.

i'd like to come back to this and fix the `sequence` algorithm's cuda perf in a separate pr.

cudax/test/execution/test_on.cu
```diff
@@ -73,25 +73,27 @@ void simple_continue_on_stream_test()
...
   auto [result] = ex::sync_wait(std::move(sndr)).value();
   CUDAX_CHECK(result == 42);
 }
 
-C2H_TEST("simple on(sch, sndr) thread test", "[on]")
+namespace
 {
-  simple_start_on_thread_test();
-}
+// C2H_TEST("simple on(sch, sndr) thread test", "[on]")
```
mhoemmen: Is there a plan to restore these tests?

cudax/test/execution/test_stream_context.cu
```diff
@@ -8,13 +8,16 @@
 //
 //===----------------------------------------------------------------------===//
 
+// BUGBUG BUGBUG
```
mhoemmen: What's the bug?

cudax/include/cuda/experimental/__execution/bulk.cuh
```diff
@@ -152,7 +161,7 @@ struct _CCCL_TYPE_VISIBILITY_DEFAULT __bulk_t
       execution::set_stopped(static_cast<_Rcvr&&>(__state_->__rcvr_));
     }
 
-    [[nodiscard]] _CCCL_NODEBUG_API constexpr auto get_env() const noexcept -> __fwd_env_t<env_of_t<_Rcvr>>
```
mhoemmen, line 155: I do not understand the consequences of this macro change.

cudax/include/cuda/experimental/__execution/continues_on.cuh
```diff
@@ -424,7 +423,7 @@ struct _CCCL_TYPE_VISIBILITY_DEFAULT continues_on_t::__sndr_t
     return __attrs_t<_Sch, _Sndr>{this};
   }
 
-  _CCCL_NO_UNIQUE_ADDRESS continues_on_t __tag_;
+  /*_CCCL_NO_UNIQUE_ADDRESS*/ continues_on_t __tag_;
```
mhoemmen, line 426: Eric explained that tests were failing without removing this macro.  I'm treating this removal as a temporary phenomenon.  There are other ways to get this effect, e.g., the "compressed pair" pattern that the reference implementation of mdspan classically used.

cudax/include/cuda/experimental/__execution/cpos.cuh
```diff
@@ -37,13 +37,13 @@ template <__disposition _Disposition>
 struct __completion_tag
 {
   template <__disposition _OtherDisposition>
-  _CCCL_NODEBUG_API constexpr auto operator==(__completion_tag<_OtherDisposition>) const noexcept -> bool
+  _CCCL_TRIVIAL_API constexpr auto operator==(__completion_tag<_OtherDisposition>) const noexcept -> bool
```
mhoemmen, line 40: I do not understand the consequences of this change.

cudax/include/cuda/experimental/__execution/bulk.cuh
ericniebler: It changes the functions' attributes so that debuggers won't step over them. Otherwise it has no effect.

cudax/test/execution/test_on.cu
ericniebler: done

cudax/test/execution/test_stream_context.cu
ericniebler: the bug is that i forgot to remove this comment. :-P

cudax/test/execution/test_stream_context.cu
```diff
@@ -8,13 +8,16 @@
...
+
```
ericniebler:
```suggestion
```

cudax/include/cuda/experimental/__execution/continues_on.cuh
ericniebler: i need to track down the compiler bug, file it, and then selectively disable `_CCCL_NO_UNIQUE_ADDRESS` acrosss all of CCCL wherever the bug can potentially manifest. then i can start using `_CCCL_NO_UNIQUE_ADDRESS` again.

cudax/include/cuda/experimental/__execution/cpos.cuh
ericniebler: a "trivial" api is one that is force-`inline`-ed and annotated so that debuggers step over them. `execution::set_stopped(move(rcvr))` just calls `move(rcvr).set_stopped()` for example. it is never interesting to step into `execution::set_stopped`; you'd rather step directly into `move(rcvr).set_stopped()`.

these annotations make debugging easier and shorten stack traces to just the good parts.

cudax/include/cuda/experimental/__execution/bulk.cuh
mhoemmen: @ericniebler You mean, it stops inlining.  We do need to restore this at some point : - )

cudax/include/cuda/experimental/__execution/continues_on.cuh
mhoemmen: @ericniebler The other option is to use a different programming technique that guarantees that empty members occupy zero bytes.  A classic one is a ["compressed pair"](https://github.com/kokkos/mdspan/blob/3c1aa524617f272836a7222e69d17574ce998dd8/include/experimental/__p0009_bits/compressed_pair.hpp) or "compressed tuple" for storing the members, that does not store empty members.

cudax/include/cuda/experimental/__execution/cpos.cuh
mhoemmen: @ericniebler wrote:
> these annotations make debugging easier and shorten stack traces to just the good parts.

We're gonna put these macros back like they were after the nightlies run, right?

cudax/include/cuda/experimental/__execution/bulk.cuh
ericniebler: i would like to keep these changes to the function attributes. i learned many things while stepping through the experimental `execution` code in cuda-gdb. it turns out i rarely want "nodebug" apis. i very occasionally want "trivial" (force-inline nodebug) apis, but most apis should just be `_CCCL_API`.

cudax/include/cuda/experimental/__execution/cpos.cuh
ericniebler: that isn't my intention. these changes are intentional. they improve the debugability of the code.

cudax/include/cuda/experimental/__execution/continues_on.cuh
ericniebler: that's true. it will be important when this code is no longer experimental to have all the space optimizations before we lock in the ABI.

miscco: I want to note that I am on the verge of completely dropping any use of `_CCCL_NO_UNIQUE_ADDRESS`

Its just a complete trainwreck in the waiting

mhoemmen: @miscco wrote:
> Its just a complete trainwreck in the waiting

100% agree; as a language feature it means well, but in practice for us it's nothing but trouble.

cudax/include/cuda/experimental/__execution/cpos.cuh
mhoemmen: Ah, got it.  Thanks for explaining!

PR #6182, opened 2025-10-10
`cuda.compute`: Add PermutationIterator 

shwina:
Closes https://github.com/NVIDIA/cccl/issues/5690

## Description

This PR adds an implementation of `PermutationIterator` to `cuda.compute`. It works as follows:

* The state of a `PermutationIterator` is a struct composed of the states of the `values` and `index` iterators
* The `advance` method advances the `index` state
* The `dereference` method:
  * dereference the `index` state to get the index value
  * advances a copy of the `value` state (always at position 0) by the index value
  * dereferences the copy of the `value` state

## Checklist


python/cuda_cccl/cuda/compute/iterators/_iterators.py
```diff
@@ -610,3 +618,200 @@ def _get_last_element_ptr(device_array) -> int:
...
+        def codegen(context, builder, sig, args):
+            struct_ptr = args[0]
+            # Use GEP to get pointer to field at index 0 (value_state)
+            field_ptr = builder.gep(
+                struct_ptr,
+                [ir.Constant(ir.IntType(32), 0), ir.Constant(ir.IntType(32), 0)],
+            )
+            return field_ptr
+
+        from numba.core.datamodel.registry import default_manager
```
tpn, line 706: Looks like this hasn't been vendored yet in numba-cuda; we'll need to keep an eye on it for when it does.

PR #6183, opened 2025-10-10
`cuda.compute`: Use annotations when available to determine signature of user-defined transform operation

shwina:
## Description

This PR uses type annotations when provided by the user to determine the signature of the user-defined operation for `unary_transform` and `binary_transform`. This avoids the need to JIT to do type inference, and also provides an escape hatch in case of any possible problems when inferring the result type. It is especially important when struct types are involved.

## Checklist

### DIFF
--- a/python/cuda_cccl/cuda/compute/numba_utils.py
+++ b/python/cuda_cccl/cuda/compute/numba_utils.py
@@ -39,10 +39,10 @@ def signature_from_annotations(func) -> numba.core.typing.Signature:
     argspec = inspect.getfullargspec(func)
     num_args = len(argspec.args)
     try:
-        retty = to_numba_type(argspec.annotations["return"])
+        ret_ann = argspec.annotations["return"]
     except KeyError:
         raise ValueError("Function has incomplete annotations: missing return type")
-
+    retty = to_numba_type(ret_ann)
     if num_args != len(argspec.annotations) - 1:  # -1 for the return type
         raise ValueError("One or more arguments are missing type annotations")
     argtys = tuple(


python/cuda_cccl/cuda/compute/algorithms/_transform.py
```diff
@@ -32,16 +32,20 @@ def __init__(
...
         # For well-known operations, we don't need a signature
         if isinstance(op, OpKind):
             self.op_wrapper = cccl.to_cccl_op(op, None)
         else:
-            if not out_value_type.is_internal:
-                out_value_type = get_inferred_return_type(op, (in_value_type,))
-            sig = out_value_type(in_value_type)
+            try:
+                sig = signature_from_annotations(op)
+            except ValueError:
```
rwgk, line 42: General remark, not for this PR:

This pattern of reusing a **built-in** exception has a non-zero risk of masking bugs deeper down the call stack and generating confusion that can easily take hours or even days to resolve.

I looked at the existing `signature_from_annotations()` implementation, it has that same pattern already.

I see there:

```python
    try:
        retty = to_numba_type(argspec.annotations["return"])
    except KeyError:
        raise ValueError("Function has incomplete annotations: missing return type")
```

Quite plausibly, when working on a new feature, a bug might slip into e.g. `to_numba_type()` or even deeper down that generates a `KeyError`. The exception will be converted to a `ValueError` in `signature_from_annotations()`, without exception chaining (`from e`), i.e. the connection to the root cause bug is silently lost. Then here the `ValueError` is silently ignored, and the behavior is to fall back to an alternative algorithm, even though there are actually valid annotation. This might go unnoticed. Only later when some subtle difference between the two ways to compute `sig` triggers something else, someone has to work back from that, through two levels of silently remapping exceptions, to the root cause bug. (I've been there a few times; it was very frustrating.)

Alternatives to avoid setting up such traps:
* Use custom exceptions (super easy, doesn't need refactoring).
* Do not (ab)use exceptions for handling expected alternative code paths (this would require deeper changes to the existing code).

shwina: I think that's very fair. However, what do you think of scoping the `KeyError` more tightly?

```python
try:
    ret_ann = argspec.annotations["return"]
except KeyError:
    raise ValueError(...)
retty = to_numba_type(ret_ann)
```

shwina: I fixed the scoping of the `KeyError`.

I do need to think a bit more about the alternatives suggested and whether/how they apply more generally. A custom exception for this specific case seems overkill, and in general Python suggests EAFP v/s LBYL. I appreciate the feedback/discussion!

rwgk: > I fixed the scoping of the `KeyError`.

That's a good step!

I wrote my previous comment mostly just hoping to add perspective for future developments. I didn't mean so much to suggest retrofitting existing code, although small and easy steps like this one tend to add up in a good way.

PR #6184, opened 2025-10-10
Improve `string_view` interoperability `std::` counterpart and `string`

davebayer:


libcudacxx/include/cuda/std/string_view
```diff
@@ -157,6 +157,20 @@ public:
       , __size_{::cuda::std::ranges::size(__r)}
   {}
 
+#if !_CCCL_COMPILER(NVRTC)
```
miscco, line 203: This needs to be guarded by [__cpp_lib_string_view](https://en.cppreference.com/w/cpp/experimental/feature_test.html#cpp_lib_string_view) because there is no guarantee that the host STL supports `string_view` even in C++17 mode

libcudacxx/test/libcudacxx/cuda/strings/std_interop/string.view.cons.pass.cpp
```diff
@@ -0,0 +1,119 @@
+// <cuda/std/string_view>
+
+//  constexpr basic_string_view(std::basic_string_view sv);
+
+#include <cuda/std/cassert>
+#include <cuda/std/string_view>
+#include <cuda/std/type_traits>
+
+#include <string_view>
```
miscco, line 20: Ditto , needs guard

libcudacxx/test/libcudacxx/cuda/strings/std_interop/string.view.cons.pass.cpp
```diff
@@ -0,0 +1,119 @@
+// <cuda/std/string_view>
+
+//  constexpr basic_string_view(std::basic_string_view sv);
+
```
miscco: I believe we can just do
```suggestion
//  constexpr basic_string_view(std::basic_string_view sv);

// UNSUPPORTED: nvrtc
```

PR #6186, opened 2025-10-10
Assert deallocation is noexcept

bdice:
## Description
Now that RMM requires `noexcept` in its `deallocate` functions, this TODO assertion should be safe to enforce. We _do_ want to enforce this in RMM, too.

xref:
- https://github.com/rapidsai/rmm/issues/2076
- https://github.com/rapidsai/rmm/pull/2077
- https://github.com/rapidsai/rmm/pull/2085
  - I verified that RMM will build successfully with this change to CCCL once this PR lands.

## Checklist

### DIFF
--- a/libcudacxx/test/libcudacxx/cuda/memory_resource/async_resource_ref/inheritance.pass.cpp
+++ b/libcudacxx/test/libcudacxx/cuda/memory_resource/async_resource_ref/inheritance.pass.cpp
@@ -29,7 +29,7 @@ struct async_resource_base
 
   virtual void* allocate(cuda::stream_ref, std::size_t, std::size_t) = 0;
 
-  virtual void deallocate(cuda::stream_ref, void* ptr, std::size_t, std::size_t) = 0;
+  virtual void deallocate(cuda::stream_ref, void* ptr, std::size_t, std::size_t) noexcept = 0;
 
   bool operator==(const async_resource_base& other) const
   {
@@ -73,7 +73,7 @@ struct async_resource_derived_first : public async_resource_base<Properties...>
     return &_val;
   }
 
-  void deallocate(cuda::stream_ref, void* ptr, std::size_t, std::size_t) override {}
+  void deallocate(cuda::stream_ref, void* ptr, std::size_t, std::size_t) noexcept override {}
 
   bool operator==(const async_resource_derived_first& other) const
   {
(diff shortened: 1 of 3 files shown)


libcudacxx/include/cuda/__memory_resource/resource_ref.h
```diff
@@ -161,10 +161,7 @@ struct _Resource_vtable_builder
   template <class _Resource>
   static void _Dealloc(void* __object, void* __ptr, size_t __bytes, size_t __alignment) noexcept
   {
-    // TODO: this breaks RMM because their memory resources do not declare their
-    // deallocate_sync functions to be noexcept. Comment out the check for now until
-    // we can fix RMM.
-    // static_assert(noexcept(static_cast<_Resource*>(__object)->deallocate(__ptr, __bytes, __alignment)));
+    static_assert(noexcept(static_cast<_Resource*>(__object)->deallocate(__ptr, __bytes, __alignment)));
```
bdice: Should there be a similar assertion in `_Dealloc_async`?

fbusato: (personal opinion) yes, `_Dealloc_async` should be marked `noexcept` and adds a similar assertion.
An exception in an asynchronous context could be also more problematic to handle

fbusato:
related to `noexcept`, should the other methods be decorated with `noexcept(noexcept())` depending on their internal operation?

libcudacxx/include/cuda/__memory_resource/resource_ref.h
pciolkosz: The entire content of `resource_ref` will be replaced soon with the experimental implementation, so I am not sure if its worth spending time refactoring the old version

bdice: I filed this PR because this is what RMM is using today if I build against CCCL `main`. I don't know how to build against the experimental implementation.

bdice: I would like to start testing the experimental implementation ASAP so that we can ensure CCCL 3.2 is usable in RAPIDS, so please feel free to migrate it here and replace this code!

pciolkosz: Ok, I will look into it next week

bdice:
@miscco @fbusato I had to add a few more `noexcept` declarations to make the build pass. I think I have it correct now -- CI will tell.

@fbusato I didn't fully understand your suggestion "should the other methods be decorated with `noexcept(noexcept())` depending on their internal operation?" If you want me to make further changes, please let me know! Enforcing this properly is a good idea, and ideally we only have to make this kind of change once.

fbusato:
> @fbusato I didn't fully understand your suggestion "should the other methods be decorated with noexcept(noexcept()) depending on their internal operation?" If you want me to make further changes, please let me know! Enforcing this properly is a good idea, and ideally we only have to make this kind of change once.

I see that the current implementation doesn't have this issue. I was just referring to `noexcept(noexcept(op))`  when `op` is the internal operation, e.g. for the comparison operators.

Issue #6104, opened 2025-10-02
[FEA]: Drop duplicated math functions from Thrust


miscco:
### Is this a duplicate?

- [x] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cccl/issues) for this request and that I agree to the [Code of Conduct](CODE_OF_CONDUCT.md)

### Area

Thrust

### Is your feature request related to a problem? Please describe.

We have some duplicated functions in `thrust/thrust/detail/integer_math.h` that are already either in `cuda/std` or in cuda

Those functions are internal helper, so we should drop them.

This includes: 

* is_power_of_2 -> `::cuda::is_power_of_two`
* integer log2    -> `::cuda::ilog2`
* divide_ri          -> `::cuda::ceil_div`
* divide_rz ?
* round_i            -> `::cuda::round_up`
* round_z           -> `::cuda::round_down`

### Describe the solution you'd like

Replace all internal uses and remove those functions

viralbhadeshiya:
@miscco I can take care of this if it's okay with you. can you assign this to me?

fbusato:
I wrote most of these functions. I can help you/review the code for sure (@miscco is OOTO now)

viralbhadeshiya:
Thanks @fbusato, let me work on it. I'll get back to you if I have any question.

viralbhadeshiya:
@fbusato just opened PR for changes, can you help with review?

fbusato:
sure, either I or someone else from the team will help you with the review.

PR #6188, opened 2025-10-10
Dropped duplicated math function from Thrust

viralbhadeshiya:
## Description

closes https://github.com/NVIDIA/cccl/issues/6104

- Removed duplicated math function from Thrust
- Moved all duplicated math function call from Thrust to cuda/std or cuda

## Checklist


thrust/thrust/detail/integer_math.h
```diff
@@ -27,6 +27,9 @@
 #endif // no system header
 #include <thrust/detail/type_deduction.h>
 
+#include <cuda/__cmath/ceil_div.h>
+#include <cuda/__cmath/round_down.h>
+#include <cuda/__cmath/round_up.h>
```
bernhardmgruber: Suggestion: I don't think this is needed, since we do not use any new functions in this header file.
```suggestion
```

cub/cub/device/dispatch/dispatch_merge_sort.cuh
miscco, line 44: as @bernhardmgruber said, please remove the unused include from files

cub/cub/device/dispatch/dispatch_merge_sort.cuh
```diff
@@ -282,7 +283,7 @@ struct DispatchMergeSort
         break;
       }
 
-      const int num_passes = static_cast<int>(THRUST_NS_QUALIFIER::detail::log2_ri(num_tiles));
+      const int num_passes = static_cast<int>(::cuda::ceil_ilog2(num_tiles));
```
miscco: The `static_cast` is not necessary, as `cuda::ceil_ilog2` returns `int`
```suggestion
      const int num_passes = ::cuda::ceil_ilog2(num_tiles);
```

thrust/thrust/detail/execute_with_allocator.h
```diff
@@ -50,7 +52,7 @@ get_temporary_buffer(thrust::detail::execute_with_allocator<Allocator, BaseSyste
 
   // How many elements of type value_type do we need to accommodate n elements
   // of type T?
-  size_type num_elements = divide_ri(sizeof(T) * n, sizeof(value_type));
+  size_type num_elements = ::cuda::ceil_div(sizeof(T) * n, sizeof(value_type));
```
miscco: pre-existing: This might lead to conversion warnings, depending on what type `size_type` denotes. We should static_cast
```suggestion
  const size_type num_elements = static_cast<size_type>(::cuda::ceil_div(sizeof(T) * n, sizeof(value_type)));
```

Applies throughout

thrust/thrust/mr/disjoint_pool.h
```diff
@@ -370,7 +372,7 @@ class disjoint_unsynchronized_pool_resource final
   [[nodiscard]] void_ptr do_allocate_impl(std::size_t bytes, std::size_t alignment)
   {
     bytes = (std::max) (bytes, m_options.smallest_block_size);
-    assert(detail::is_power_of_2(alignment));
+    assert(::cuda::ceil_ilog2(alignment));
```
miscco: Critical: This is not correct
```suggestion
    assert(::cuda::is_power_of_two(alignment));
```

thrust/thrust/detail/integer_math.h
viralbhadeshiya: Fixed.

cub/cub/device/dispatch/dispatch_merge_sort.cuh
viralbhadeshiya: Fixed.

thrust/thrust/detail/execute_with_allocator.h
viralbhadeshiya: Fixed.

thrust/thrust/mr/disjoint_pool.h
viralbhadeshiya: Fixed.

viralbhadeshiya:
@bernhardmgruber @miscco all changes that you guys suggested is done, lmk if there are more Changes. :)

miscco:
@viralbhadeshiya This looks great 🎉 

Thanks a lot for working on this.

I added a few minor changes:
* dropped unused `is_odd`
* improved `is_negative`
* dropped two more includes

thrust/thrust/detail/integer_math.h
```diff
@@ -25,89 +25,29 @@
...
-  if (!is_power_of_2(x))
+  if constexpr (::cuda::std::is_unsigned_v<T>)
   {
-    ++result;
+    return false;
+  }
+  else
+  {
+    return x < 0;
   }
```
bernhardmgruber: Suggestion: we can simplify this even further:
```c++
return ::cuda::std::is_signed_v<T> && x < 0;
```

miscco: That is not a simplification, because the compiler most often cannot elide the `::cuda::std::is_signed_v<T>`

The if constexpr is the most efficient way

bernhardmgruber: Wat, really? The compiler just has to fold a constant. But I believe you if you say so.

viralbhadeshiya:
Seems like couple of jobs are failing, @miscco lmk if it requires any changes for it, happy to do it.

thrust/thrust/detail/integer_math.h
```diff
@@ -25,89 +25,29 @@
...
-_CCCL_HOST_DEVICE _CCCL_FORCEINLINE bool is_odd(Integer x)
-{
-  return 1 & x;
-}
-
-template <typename Integer>
-_CCCL_HOST_DEVICE _CCCL_FORCEINLINE Integer log2(Integer x)
+_CCCL_TEMPLATE(typename T)
+_CCCL_REQUIRES(::cuda::std::is_arithmetic_v<T>)
+_CCCL_HOST_DEVICE _CCCL_FORCEINLINE constexpr bool is_negative([[maybe_unused]] const T x) noexcept
```
fbusato:
```suggestion
[[nodiscard]] _CCCL_API constexpr bool is_negative([[maybe_unused]] T x) noexcept
```

thrust/thrust/detail/integer_math.h
```diff
@@ -25,89 +25,29 @@
 #elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_MSVC)
 #  pragma system_header
```
fbusato: would make sense to also update the license because the file is practically entirely new.

viralbhadeshiya: @fbusato how can I do that?

miscco: Honestly, this is effectively only ever used in `thrust/system/cuda/detail/dispatch.h`, so lets move the function there and drop the file?

viralbhadeshiya: @miscco I see couple of usage of ```is_negative``` in ```thrust/thrust/system/cuda/detail/scan.h``` too, do we need to worry about that?

miscco: It is already including `dispatch.h`

bernhardmgruber (approved):

miscco:
@fbusato There is an issue with `thrust::pool_options`

I suspect there is a differing behavior for `is_power_of_two` could you please inevestigate why it differs?

fbusato:
`::cuda::is_power_of_two` correctly evaluates 0 as non-power of two, while `thrust::detail::is_power_of_2()` misses the check `x != 0`.

This causes a different behavior in `!::cuda::is_power_of_two(alignment)`, which is initialized to 0 by default.
https://github.com/NVIDIA/cccl/pull/6188/files#diff-d85bf89d70ec2f2a069c4c0ffeccd2722fe2a108a012566503392885900f40a1R115

fbusato:
@viralbhadeshiya my suggestion is just to add the condition `alignment != 0 && !::cuda::is_power_of_two(alignment)`

viralbhadeshiya:
> `::cuda::is_power_of_two` correctly evaluates 0 as non-power of two, while `thrust::detail::is_power_of_2()` misses the check `x != 0`.
> 
> This causes a different behavior in `!::cuda::is_power_of_two(alignment)`, which is initialized to 0 by default. https://github.com/NVIDIA/cccl/pull/6188/files#diff-d85bf89d70ec2f2a069c4c0ffeccd2722fe2a108a012566503392885900f40a1R115

@fbusato would this similar changes also required in ```!::cuda::is_power_of_two(smallest_block_size)``` & ```!::cuda::is_power_of_two(largest_block_size)```?

fbusato:
yes, but I'm not sure that this edge case is evaluated in the test case. You can also run the test locally to make sure it works

viralbhadeshiya:
> yes, but I'm not sure that this edge case is evaluated in the test case. You can also run the test locally to make sure it works

Okay made changes, unfortunately my remote server which had NVIDIA device is facing some issue so could not run tests. but changes are done.

PR #6192, opened 2025-10-13
[libcu++] Implement `cuda::ffs`

### LINKED ISSUE #6108: [FEA]: Implement a proper `__ffs` replacement
### Is this a duplicate?

### Area

libcu++

We use `__ffs` in some places, but that intrinsic expects integers of a given size. We should provide a type safe replacement `::cuda::bit_ffs` or similar that works on all supported platforms and all supported integers

Implement the function in `<cuda/bit>`

_No response_

### Additional context

_No response_

Aminsed:
Implement type-safe `cuda::std::ffs` function as replacement for `__ffs` intrinsic.

* Returns 1-based index of first set bit (0 for no bits set)
* Works on all platforms and integer types
* Handles `x == 0` correctly (returns 0)

Fixes #6108

### DIFF
--- a/docs/libcudacxx/extended_api/bit.rst
+++ b/docs/libcudacxx/extended_api/bit.rst
@@ -11,6 +11,7 @@ Bit
    bit/bit_reverse
    bit/bitfield_insert
    bit/bitfield_extract
+   bit/ffs
 
 .. list-table::
    :widths: 25 45 30 30
@@ -40,3 +41,8 @@ Bit
      - Extract a bitfield
      - CCCL 3.0.0
      - CUDA 13.0
+
+   * - :ref:`ffs <libcudacxx-extended-api-bit-ffs>`
+     - Find first set bit
+     - CCCL 3.2.0
+     - CUDA 13.2
--- a/docs/libcudacxx/extended_api/bit/ffs.rst
+++ b/docs/libcudacxx/extended_api/bit/ffs.rst
@@ -1,0 +1,68 @@
+.. _libcudacxx-extended-api-bit-ffs:
+
+``cuda::ffs``
+=============
+
+Defined in the ``<cuda/bit>`` header.
+
+.. code:: cpp
+
+   namespace cuda {
+
+   template <class T>
+   [[nodiscard]] __host__ __device__ constexpr
+   int ffs(T value) noexcept;
+
+   } // namespace cuda
+
+The function finds the first (least significant) set bit in ``value`` and returns its 1-based index. If ``value`` is 0, returns 0.
+
+**Parameters**
+
+- ``value``: Input value
+
+**Return value**
+
+- The 1-based index of the first set bit, or 0 if ``value`` is 0
+
+**Constraints**
+
+- ``T`` must be an unsigned integer type. Supported types include all standard unsigned integer types and ``__uint128_t`` when available.
+
+**Relationship with other functions**
+
+- For non-zero values: ``ffs(x) == countr_zero(x) + 1``
+
+**Performance considerations**
+
+The function performs the following operations:
+
+- Device:
+
+  - ``uint8_t``, ``uint16_t``, ``uint32_t``: ``BREV``, ``FLO``, ``IADD3``
+
+- Host:
+
+  - GCC/Clang: ``__builtin_ffs`` / ``__builtin_ffsll``
+  - MSVC: ``_BitScanForward`` / ``_BitScanForward64``
+
+Example
+-------
+
+.. code:: cpp
+
+    #include <cuda/bit>
+    #include <cuda/std/cassert>
+
+    __global__ void ffs_kernel() {
+        assert(cuda::ffs(0u) == 0);
+        assert(cuda::ffs(1u) == 1);
+        assert(cuda::ffs(0b1100u) == 3);
+        assert(cuda::ffs(0x80000000u) == 32);
+    }
+
+    int main() {
+        ffs_kernel<<<1, 1>>>();
+        cudaDeviceSynchronize();
+        return 0;
+    }
--- a/libcudacxx/include/cuda/__bit/ffs.h
+++ b/libcudacxx/include/cuda/__bit/ffs.h
@@ -1,0 +1,175 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of libcu++, the C++ Standard Library for your entire system,
+// under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef _CUDA___BIT_FFS_H
+#define _CUDA___BIT_FFS_H
+
+#include <cuda/std/detail/__config>
+
+#if defined(_CCCL_IMPLICIT_SYSTEM_HEADER_GCC)
+#  pragma GCC system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_CLANG)
+#  pragma clang system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_MSVC)
+#  pragma system_header
+#endif // no system header
+
+#include <cuda/std/__type_traits/enable_if.h>
+#include <cuda/std/__type_traits/is_unsigned_integer.h>
+#include <cuda/std/__type_traits/remove_cv.h>
+#include <cuda/std/cstdint>
+
+#include <nv/target>
+
+#if _CCCL_COMPILER(MSVC)
+#  include <intrin.h>
+#endif // _CCCL_COMPILER(MSVC)
+
+#include <cuda/std/__cccl/prologue.h>
+
+#if _CCCL_HAS_BUILTIN(__builtin_ffs) || _CCCL_COMPILER(GCC)
+#  define _CCCL_BUILTIN_FFS(...)   __builtin_ffs(__VA_ARGS__)
+#  define _CCCL_BUILTIN_FFSLL(...) __builtin_ffsll(__VA_ARGS__)
+#endif // _CCCL_HAS_BUILTIN(__builtin_ffs) || _CCCL_COMPILER(GCC)
+
+_CCCL_BEGIN_NAMESPACE_CUDA
+
+namespace __detail
+{
+// Generic constexpr implementation - used for constexpr evaluation and as fallback
+template <class _Tp>
+[[nodiscard]] _CCCL_HOST_DEVICE constexpr int __ffs_constexpr_impl(_Tp __v) noexcept
+{
+  static_assert(::cuda::std::__cccl_is_unsigned_integer_v<_Tp>, "_Tp must be unsigned");
+  if (__v == 0)
+  {
+    return 0;
+  }
+  int __pos = 1;
+  while ((__v & 1) == 0)
+  {
+    __v >>= 1;
+    ++__pos;
+  }
+  return __pos;
+}
+
+#if _CCCL_CUDA_COMPILATION()
+// Device-specific implementations using CUDA intrinsics
+[[nodiscard]] _CCCL_DEVICE_API inline int __ffs_device32(::cuda::std::uint32_t __v) noexcept
+{
+  return ::__ffs(static_cast<int>(__v));
+}
+
+[[nodiscard]] _CCCL_DEVICE_API inline int __ffs_device64(::cuda::std::uint64_t __v) noexcept
+{
+  return ::__ffsll(static_cast<long long>(__v));
+}
+
+#  if _CCCL_HAS_INT128()
+[[nodiscard]] _CCCL_DEVICE_API inline int __ffs_device128(__uint128_t __v) noexcept
+{
+  const auto __lo = static_cast<::cuda::std::uint64_t>(__v);
+  if (const int __result = ::cuda::__detail::__ffs_device64(__lo))
+  {
+    return __result;
+  }
+  const auto __hi = static_cast<::cuda::std::uint64_t>(__v >> 64);
+  if (const int __result = ::cuda::__detail::__ffs_device64(__hi))
+  {
+    return __result + 64;
+  }
+  return 0;
+}
+#  endif // _CCCL_HAS_INT128()
+
+#endif // _CCCL_CUDA_COMPILATION()
+} // namespace __detail
+
+template <class _Tp, ::cuda::std::enable_if_t<::cuda::std::__cccl_is_cv_unsigned_integer_v<_Tp>, int> = 0>
+[[nodiscard]] _CCCL_API constexpr int ffs(_Tp __v) noexcept
+{
+  using _Unsigned = ::cuda::std::remove_cv_t<_Tp>;
+
+  _CCCL_IF_NOT_CONSTEVAL_DEFAULT
+  {
+#if _CCCL_HAS_INT128()
+    if constexpr (sizeof(_Unsigned) == sizeof(__uint128_t))
+    {
+      NV_IF_ELSE_TARGET(
+        NV_IS_HOST,
+        (
+          // Host path for 128-bit: split into two 64-bit parts
+          const auto __lo = static_cast<::cuda::std::uint64_t>(__v);
+          const auto __hi = static_cast<::cuda::std::uint64_t>(static_cast<__uint128_t>(__v) >> 64);
+#  if defined(_CCCL_BUILTIN_FFSLL)
+          if (const int __result = _CCCL_BUILTIN_FFSLL(static_cast<long long>(__lo))) {
+            return __result;
+          } if (const int __result = _CCCL_BUILTIN_FFSLL(static_cast<long long>(__hi))) {
+            return __result + 64;
+          } return 0;
+#  else
+          if (const int __result = ::cuda::__detail::__ffs_constexpr_impl(__lo)) {
+            return __result;
+          } if (const int __result = ::cuda::__detail::__ffs_constexpr_impl(__hi)) { return __result + 64; } return 0;
+#  endif
+          ),
+        (return ::cuda::__detail::__ffs_device128(static_cast<__uint128_t>(__v));))
+    }
+#endif // _CCCL_HAS_INT128()
+
+    if constexpr (sizeof(_Unsigned) <= sizeof(::cuda::std::uint32_t))
+    {
+      NV_IF_ELSE_TARGET(
+        NV_IS_HOST,
+        (
+          // Host path for 32-bit
+          const auto __v32 = static_cast<::cuda::std::uint32_t>(__v);
+#if defined(_CCCL_BUILTIN_FFS)
+          return _CCCL_BUILTIN_FFS(static_cast<int>(__v32));
+#elif _CCCL_COMPILER(MSVC)
+          unsigned long __where{};
+          const unsigned char __res = ::_BitScanForward(&__where, __v32);
+          return __res ? static_cast<int>(__where) + 1 : 0;
+#else
+          return ::cuda::__detail::__ffs_constexpr_impl(__v32);
+#endif
+          ),
+        (return ::cuda::__detail::__ffs_device32(static_cast<::cuda::std::uint32_t>(__v));))
+    }
+    else if constexpr (sizeof(_Unsigned) <= sizeof(::cuda::std::uint64_t))
+    {
+      NV_IF_ELSE_TARGET(
+        NV_IS_HOST,
+        (
+          // Host path for 64-bit
+          const auto __v64 = static_cast<::cuda::std::uint64_t>(__v);
+#if defined(_CCCL_BUILTIN_FFSLL)
+          return _CCCL_BUILTIN_FFSLL(static_cast<long long>(__v64));
+#elif _CCCL_COMPILER(MSVC)
+          unsigned long __where{};
+          const unsigned char __res = ::_BitScanForward64(&__where, __v64);
+          return __res ? static_cast<int>(__where) + 1 : 0;
+#else
+          return ::cuda::__detail::__ffs_constexpr_impl(__v64);
+#endif
+          ),
+        (return ::cuda::__detail::__ffs_device64(static_cast<::cuda::std::uint64_t>(__v));))
+    }
+  }
+
+  return ::cuda::__detail::__ffs_constexpr_impl(static_cast<_Unsigned>(__v));
+}
+
+_CCCL_END_NAMESPACE_CUDA
+
+#include <cuda/std/__cccl/epilogue.h>
+
+#endif // _CUDA___BIT_FFS_H
--- a/libcudacxx/test/libcudacxx/cuda/bit/ffs.pass.cpp
+++ b/libcudacxx/test/libcudacxx/cuda/bit/ffs.pass.cpp
@@ -1,0 +1,129 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of the libcu++ Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES.
+//
+//===----------------------------------------------------------------------===//
+
+// int ffs(T x) noexcept;
+//
+// Returns: The 1-based index of the first (least significant) set bit, or 0 if x == 0.
+
+#include <cuda/bit>
+#include <cuda/std/cassert>
+#include <cuda/std/cstdint>
+#include <cuda/std/type_traits>
+
+#include "literal.h"
+#include "test_macros.h"
+
+using namespace test_integer_literals;
+
+template <typename T>
+__host__ __device__ constexpr bool constexpr_test()
+{
+  assert(cuda::ffs(T(0)) == 0);
+  assert(cuda::ffs(T(1)) == 1);
+  assert(cuda::ffs(T(2)) == 2);
+  assert(cuda::ffs(T(3)) == 1);
+  assert(cuda::ffs(T(4)) == 3);
+  assert(cuda::ffs(T(5)) == 1);
+  assert(cuda::ffs(T(6)) == 2);
+  assert(cuda::ffs(T(7)) == 1);
+  assert(cuda::ffs(T(8)) == 4);
+  assert(cuda::ffs(T(9)) == 1);
+  assert(cuda::ffs(T(127)) == 1);
+  assert(cuda::ffs(T(128)) == 8);
+
+  // Test relationship with countr_zero: ffs(x) == countr_zero(x) + 1 for x != 0
+  assert(cuda::ffs(T(1)) == cuda::std::countr_zero(T(1)) + 1);
+  assert(cuda::ffs(T(2)) == cuda::std::countr_zero(T(2)) + 1);
+  assert(cuda::ffs(T(4)) == cuda::std::countr_zero(T(4)) + 1);
+  assert(cuda::ffs(T(8)) == cuda::std::countr_zero(T(8)) + 1);
+
+  // Test MSB for different sizes (compile-time)
+  if constexpr (sizeof(T) >= 4)
+  {
+    assert(cuda::ffs(T(0x80000000u)) == 32);
+  }
+  if constexpr (sizeof(T) >= 8)
+  {
+    assert(cuda::ffs(T(0x8000000000000000ull)) == 64);
+  }
+
+  return true;
+}
+
+template <typename T>
+__host__ __device__ inline void assert_ffs(T val, int expected)
+{
+  volatile auto v = val;
+  assert(cuda::ffs(v) == expected);
+}
+
+template <typename T>
+__host__ __device__ void runtime_test()
+{
+  static_assert(cuda::std::is_same_v<int, decltype(cuda::ffs(T(0)))>, "");
+  static_assert(noexcept(cuda::ffs(T(0))), "");
+
+  assert_ffs(T(0), 0);
+  assert_ffs(T(1), 1);
+  assert_ffs(T(121), 1);
+  assert_ffs(T(122), 2);
+  assert_ffs(T(124), 3);
+
+  if constexpr (sizeof(T) > 1)
+  {
+    assert_ffs(T(128), 8);
+    assert_ffs(T(256), 9);
+    assert_ffs(T(512), 10);
+    assert_ffs(T(1024), 11);
+  }
+
+  if constexpr (sizeof(T) >= 4)
+  {
+    assert_ffs(T(0x80000000u), 32);
+  }
+
+  if constexpr (sizeof(T) >= 8)
+  {
+    assert_ffs(T(0x8000000000000000ull), 64);
+  }
+}
+
+int main(int, char**)
+{
+  static_assert(constexpr_test<unsigned char>(), "");
+  static_assert(constexpr_test<unsigned short>(), "");
+  static_assert(constexpr_test<unsigned>(), "");
+  static_assert(constexpr_test<unsigned long>(), "");
+  static_assert(constexpr_test<unsigned long long>(), "");
+
+#if _CCCL_HAS_INT128()
+  static_assert(constexpr_test<__uint128_t>(), "");
+#endif // _CCCL_HAS_INT128()
+
+  runtime_test<unsigned char>();
+  runtime_test<unsigned short>();
+  runtime_test<unsigned>();
+  runtime_test<unsigned long>();
+  runtime_test<unsigned long long>();
+
+#if _CCCL_HAS_INT128()
+  runtime_test<__uint128_t>();
+
+  // Additional 128-bit tests with literals
+  assert_ffs(0_u128, 0);
+  assert_ffs(1_u128, 1);
+  assert_ffs(0x8000000000000000_u128, 64);
+  assert_ffs((1_u128 << 64), 65);
+  assert_ffs((1_u128 << 65), 66);
+  assert_ffs((1_u128 << 100), 101);
+  assert_ffs((1_u128 << 127), 128);
+#endif // _CCCL_HAS_INT128()
+
+  return 0;
+}
--- a/libcudacxx/include/cuda/bit
+++ b/libcudacxx/include/cuda/bit
@@ -24,6 +24,7 @@
 #include <cuda/__bit/bit_reverse.h>
 #include <cuda/__bit/bitfield.h>
 #include <cuda/__bit/bitmask.h>
+#include <cuda/__bit/ffs.h>
 #include <cuda/std/bit>
 
 #endif // _CUDA_BIT
--- a/libcudacxx/include/cuda/pipeline
+++ b/libcudacxx/include/cuda/pipeline
@@ -28,6 +28,7 @@
 #include <cuda/__ptx/instructions/get_sreg.h>
 #include <cuda/atomic>
 #include <cuda/barrier>
+#include <cuda/bit>
 #include <cuda/std/__algorithm/max.h>
 #include <cuda/std/__bit/popcount.h>
 #include <cuda/std/__chrono/duration.h>
@@ -111,7 +112,7 @@ public:
       NV_IS_DEVICE,
       const uint32_t __match_mask =
         ::__match_any_sync(::__activemask(), reinterpret_cast<uintptr_t>(__shared_state_get_refcount()));
-      const uint32_t __elected_id = ::__ffs(__match_mask) - 1;
+      const uint32_t __elected_id = ::cuda::ffs(__match_mask) - 1;
       __elected                   = (::cuda::ptx::get_sreg_laneid() == __elected_id);
       __sub_count                 = ::cuda::std::popcount(__match_mask);
       , __elected = true;
@@ -133,11 +134,6 @@ public:
     }
     __active = false;
     return __released;
-  }
-
-  _CCCL_API inline bool __is_active() const
-  {
-    return __active;
   }
 
   _CCCL_API inline void producer_acquire()
@@ -305,7 +301,7 @@ make_pipeline(const _Group& __group, pipeline_shared_state<_Scope, _Stages_count
       NV_IS_DEVICE,
       const uint32_t __match_mask =
         ::__match_any_sync(::__activemask(), reinterpret_cast<uintptr_t>(&__shared_state->__refcount));
-      const uint32_t __elected_id = ::__ffs(__match_mask) - 1;
+      const uint32_t __elected_id = ::cuda::ffs(__match_mask) - 1;
       __elected                   = (::cuda::ptx::get_sreg_laneid() == __elected_id);
       __add_count                 = ::cuda::std::popcount(__match_mask);
       , __elected = true;
@@ -473,15 +469,15 @@ _CCCL_API inline pipeline<thread_scope_thread> make_pipeline()
 }
 
 template <uint8_t _Prior>
-_CCCL_API inline void pipeline_consumer_wait_prior([[maybe_unused]] pipeline<thread_scope_thread>& __pipeline)
+_CCCL_API inline void pipeline_consumer_wait_prior(pipeline<thread_scope_thread>& __pipeline)
 {
   NV_IF_TARGET(NV_PROVIDES_SM_80, ::cuda::device::__pipeline_consumer_wait<_Prior>(__pipeline);
                __pipeline.__tail = __pipeline.__head - _Prior;)
 }
 
 template <thread_scope _Scope>
-_CCCL_API inline void pipeline_producer_commit([[maybe_unused]] pipeline<thread_scope_thread>& __pipeline,
-                                               [[maybe_unused]] barrier<_Scope>& __barrier)
+_CCCL_API inline void
+pipeline_producer_commit([[maybe_unused]] pipeline<thread_scope_thread>& __pipeline, barrier<_Scope>& __barrier)
 {
   NV_IF_TARGET(NV_PROVIDES_SM_80,
                ((void) __memcpy_completion_impl::__defer(
@@ -492,11 +488,6 @@ template <typename _Group, class _Tp, typename _Size, thread_scope _Scope>
 _CCCL_API inline async_contract_fulfillment __memcpy_async_pipeline(
   _Group const& __group, _Tp* __destination, _Tp const* __source, _Size __size, pipeline<_Scope>& __pipeline)
 {
-  if constexpr (_Scope != thread_scope_thread)
-  {
-    _CCCL_ASSERT(__pipeline.__is_active(), "The pipeline used for memcpy_async must be active (not quitted)");
-  }
-
   // 1. Set the completion mechanisms that can be used.
   //
   //    Do not (yet) allow async_bulk_group completion. Do not allow


libcudacxx/include/cuda/std/__bit/ffs.h
```diff
@@ -0,0 +1,144 @@
...
+  return (__res) ? static_cast<int>(__where) + 1 : 0;
+#  else
+  return ::cuda::std::__cccl_ffs_impl_constexpr(__v);
+#  endif // _CCCL_COMPILER(MSVC)
+}
+#endif // !_CCCL_COMPILER(NVRTC)
+
+#if _CCCL_CUDA_COMPILATION()
+template <typename _Tp>
+[[nodiscard]] _CCCL_HIDE_FROM_ABI _CCCL_DEVICE int __cccl_ffs_impl_device(_Tp __v) noexcept
```
miscco: We gravitate towards always using `_CCCL_DEVICE_API` instead of `_CCCL_HIDE_FROM_ABI _CCCL_DEVICE`

Applies throughout

libcudacxx/include/cuda/std/__bit/ffs.h
```diff
@@ -0,0 +1,144 @@
...
+}
+
+_CCCL_TEMPLATE(class _Tp)
+_CCCL_REQUIRES(::cuda::std::__cccl_is_unsigned_integer_v<_Tp>)
+[[nodiscard]] _CCCL_API constexpr int ffs(_Tp __v) noexcept
+{
+  int __result{};
+
+#if defined(_CCCL_BUILTIN_FFSG)
+  __result = _CCCL_BUILTIN_FFSG(__v);
```
miscco: According to the [GCC documentation](https://gcc.gnu.org/onlinedocs/gcc/Bit-Operation-Builtins.html)

This takes a *signed* integer, so we need to cast to `make_signed_t<_Tp>`

libcudacxx/include/cuda/__barrier/barrier_block_scope.h
```diff
@@ -167,7 +167,7 @@ class barrier<thread_scope_block, ::cuda::std::__empty_completion> : public __bl
         unsigned int __active  = __activeA & __activeB;
         int __inc              = ::__popc(__active) * __update;
 
-        int __leader = ::__ffs(__active) - 1;
+        int __leader = ::cuda::std::ffs(__active) - 1;
```
davebayer: The `ffs` function should be our extension, so it should go to `cuda::` namespace

libcudacxx/include/cuda/std/__bit/ffs.h
```diff
@@ -0,0 +1,144 @@
...
+// Part of libcu++, the C++ Standard Library for your entire system,
+// under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef _CUDA_STD___BIT_FFS_H
+#define _CUDA_STD___BIT_FFS_H
```
davebayer:
```suggestion
#ifndef _CUDA___BIT_FFS_H
#define _CUDA___BIT_FFS_H
```
and move this file to `<cuda/__bit/ffs.h>`, please

libcudacxx/include/cuda/std/__bit/ffs.h
```diff
@@ -0,0 +1,144 @@
...
+{
+  int __result{};
+
+#if defined(_CCCL_BUILTIN_FFSG)
+  __result = _CCCL_BUILTIN_FFSG(__v);
+#else // ^^^ _CCCL_BUILTIN_FFSG ^^^ // vvv !_CCCL_BUILTIN_FFSG vvv
+  if constexpr (sizeof(_Tp) <= sizeof(uint64_t))
+  {
+    using _Sp = _If<sizeof(_Tp) <= sizeof(uint32_t), uint32_t, uint64_t>;
+    __result  = ::cuda::std::__cccl_ffs_impl(static_cast<_Sp>(__v));
```
davebayer: I would rather move the just call `__cccl_ffs_impl` and casted everything that has less or same size as `int` to `int` there, we don't need the `_Sp` type.

libcudacxx/include/cuda/std/__cccl/builtin.h
```diff
@@ -158,6 +158,20 @@
...
+#endif // _CCCL_CHECK_BUILTIN(builtin_ffs)
+
+#if _CCCL_CHECK_BUILTIN(builtin_ffsg)
+#  define _CCCL_BUILTIN_FFSG(...) __builtin_ffsg(__VA_ARGS__)
+#endif // _CCCL_CHECK_BUILTIN(builtin_ffsg)
+
+// nvcc cannot handle __builtin_ffsg
+#if _CCCL_CUDA_COMPILER(NVCC)
+#  undef _CCCL_BUILTIN_FFSG
+#endif // _CCCL_CUDA_COMPILER(NVCC)
```
davebayer: We can move these builtins directly to the `<cuda/__bit/ffs.h>` file

Aminsed:
Addressed all feedback:

* Moved to cuda/__bit/ffs.h
* Changed to cuda:: namespace
* Updated header guards
* Using _CCCL_DEVICE_API
* Fixed signed cast with make_signed_t
* Simplified implementation
* Builtin definitions moved to ffs.h
* Added comprehensive tests

libcudacxx/include/cuda/__bit/ffs.h
```diff
@@ -0,0 +1,137 @@
...
+#if _CCCL_COMPILER(MSVC)
+#  include <intrin.h>
+#endif // _CCCL_COMPILER(MSVC)
+
+#if _CCCL_COMPILER(GCC) || _CCCL_COMPILER(CLANG) || _CCCL_COMPILER(NVHPC)
+#  define _CCCL_BUILTIN_FFS(...)   __builtin_ffs(__VA_ARGS__)
+#  define _CCCL_BUILTIN_FFSLL(...) __builtin_ffsll(__VA_ARGS__)
+#endif // _CCCL_COMPILER(GCC) || _CCCL_COMPILER(CLANG) || _CCCL_COMPILER(NVHPC)
+
+#include <cuda/std/__cccl/prologue.h>
```
davebayer:
```suggestion
#include <cuda/std/__cccl/prologue.h>

#if _CCCL_HAS_BUILTIN(__builtin_ffs) || _CCCL_COMPILER(GCC)
#  define _CCCL_BUILTIN_FFS(...)   __builtin_ffs(__VA_ARGS__)
#  define _CCCL_BUILTIN_FFSLL(...) __builtin_ffsll(__VA_ARGS__)
#endif // _CCCL_HAS_BUILTIN(__builtin_ffs) || _CCCL_COMPILER(GCC)
```
Let's put the check inside the prologue/epilogue scope. Also, it should be enough to use the combination of `_CCCL_HAS_BUILTIN(__builtin_ffs)` and `_CCCL_COMPILER(GCC)` :)

libcudacxx/include/cuda/__bit/ffs.h
```diff
@@ -0,0 +1,137 @@
...
+#  if defined(_CCCL_BUILTIN_FFS)
+  using _Signed = ::cuda::std::make_signed_t<_Tp>;
+  if constexpr (sizeof(_Tp) <= sizeof(int))
+  {
+    return _CCCL_BUILTIN_FFS(static_cast<_Signed>(__v));
+  }
+  else
+  {
+    return _CCCL_BUILTIN_FFSLL(static_cast<_Signed>(__v));
+  }
```
davebayer:
```suggestion
  if constexpr (sizeof(_Tp) <= sizeof(int))
  {
    return _CCCL_BUILTIN_FFS(static_cast<int>(__v));
  }
  else
  {
    return _CCCL_BUILTIN_FFSLL(static_cast<long long>(__v));
  }
```
we can simplify it a bit further

libcudacxx/include/cuda/__bit/ffs.h
```diff
@@ -0,0 +1,137 @@
...
+[[nodiscard]] _CCCL_API constexpr int ffs(_Tp __v) noexcept
+{
+  if constexpr (sizeof(_Tp) <= sizeof(int))
+  {
+    return ::cuda::__ffs_impl(static_cast<uint32_t>(__v));
+  }
+  else
+  {
+    return ::cuda::__ffs_impl(static_cast<uint64_t>(__v));
+  }
```
davebayer: This is unnecessary I believe, we can just move contents of `__ffs_impl` here directly

libcudacxx/include/cuda/__bit/ffs.h
```diff
@@ -0,0 +1,137 @@
...
+  {
+    return ::__ffsll(static_cast<long long>(__v));
+  }
+}
+#endif // _CCCL_CUDA_COMPILATION()
+
+template <typename _Tp>
+[[nodiscard]] _CCCL_API constexpr int __ffs_impl(_Tp __v) noexcept
+{
+  static_assert(::cuda::std::is_same_v<_Tp, uint32_t> || ::cuda::std::is_same_v<_Tp, uint64_t>);
```
davebayer: This can be removed

libcudacxx/include/cuda/__bit/ffs.h
davebayer: We also support 128-bit integers if available, so we should handle them, too. You can just split them into 2x 64-bit parts and handle them separately

libcudacxx/test/libcudacxx/cuda/bit/ffs.pass.cpp
```diff
@@ -0,0 +1,77 @@
...
+  test<uint16_t>();
+  test<uint32_t>();
+  test<uint64_t>();
+  test<size_t>();
+  test<uintmax_t>();
+  test<uintptr_t>();
+
+#if _CCCL_HAS_INT128()
+  test<__uint128_t>();
+#endif // _CCCL_HAS_INT128()
```
davebayer:
```suggestion
  test<unsigned char>();
  test<unsigned short>();
  test<unsigned>();
  test<unsigned long>();
  test<unsigned long long>();
#if _CCCL_HAS_INT128()
  test<__uint128_t>();
#endif // _CCCL_HAS_INT128()
```
Those are not necessary, they are just aliases to the standard integer types

libcudacxx/test/libcudacxx/cuda/bit/ffs.pass.cpp
```diff
@@ -0,0 +1,77 @@
...
+#include <cuda/std/cassert>
+#include <cuda/std/cstdint>
+#include <cuda/std/type_traits>
+
+#include "test_macros.h"
+
+template <typename T>
+__host__ __device__ constexpr bool test()
+{
+  static_assert(cuda::ffs(T(0)) == 0);
```
davebayer: These tests should use just `assert(...)` instead of `static_assert(...)`. The values will be checked in `constexpr` context when we call `static_assert(test()` from `main`

libcudacxx/test/libcudacxx/cuda/bit/ffs.pass.cpp
```diff
@@ -0,0 +1,77 @@
...
+  static_assert(cuda::ffs(T(2)) == 2);
+  static_assert(cuda::ffs(T(3)) == 1);
+  static_assert(cuda::ffs(T(4)) == 3);
+  static_assert(cuda::ffs(T(5)) == 1);
+  static_assert(cuda::ffs(T(6)) == 2);
+  static_assert(cuda::ffs(T(7)) == 1);
+  static_assert(cuda::ffs(T(8)) == 4);
+  static_assert(cuda::ffs(T(9)) == 1);
+  static_assert(cuda::ffs(T(127)) == 1);
+  static_assert(cuda::ffs(T(128)) == 8);
```
davebayer: I'd like to see tests of greater values than `128`. That's also why these tests wouldn't catch that the function won't work for 128-bit integers.

For 128-bit integer literals, you can include `#include "literal.h" and use `_u128` user defined literals from `test_integer_literals` namespace

libcudacxx/test/libcudacxx/cuda/bit/ffs.pass.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+  // Test relationship with countr_zero: ffs(x) == countr_zero(x) + 1 for x != 0
+  static_assert(cuda::ffs(T(1)) == cuda::std::countr_zero(T(1)) + 1);
+  static_assert(cuda::ffs(T(2)) == cuda::std::countr_zero(T(2)) + 1);
+  static_assert(cuda::ffs(T(4)) == cuda::std::countr_zero(T(4)) + 1);
+  static_assert(cuda::ffs(T(8)) == cuda::std::countr_zero(T(8)) + 1);
+
+  // Test return type
+  static_assert(cuda::std::is_same_v<int, decltype(cuda::ffs(T(0)))>);
+  static_assert(noexcept(cuda::ffs(T(0))));
```
miscco: Important, this will only test at compile time, we want all those `static_assert` to be plain `assert`

At compile time that would trigger a compile error but do the right thing at runtime

libcudacxx/test/libcudacxx/cuda/bit/ffs.pass.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+#include <cuda/bit>
+#include <cuda/std/cassert>
+#include <cuda/std/cstdint>
+#include <cuda/std/type_traits>
+
+#include "test_macros.h"
+
+template <typename T>
+__host__ __device__ constexpr bool test()
```
miscco: NitpicK: This function could just be void

libcudacxx/test/libcudacxx/cuda/bit/ffs.pass.cpp
```diff
@@ -0,0 +1,77 @@
...
+  test<__uint128_t>();
+#endif // _CCCL_HAS_INT128()
+  return true;
+}
+
+int main(int, char**)
+{
+  assert(test());
+  static_assert(test());
+  return 0;
```
miscco: Important, we had issues with the compiler being too clever and optimizing the whole tests out.

to alleviate that we do
```suggestion
  volatile int val = 0;
  assert(test(val));
  static_assert(test(0));
  return 0;
```

And then inside `test(const int)` we static cast to the respective type.

Its a pain, but it ensures that the optimizer does not constant fold the function out

Aminsed:
* Builtins use _CCCL_HAS_BUILTIN inside prologue
* Simplified to direct int/long long casts
* Flattened ffs() implementation
* Added 128-bit integer support with hi/lo split
* Tests use assert in constexpr context
* Added large value tests with _u128 literals

libcudacxx/include/cuda/__bit/ffs.h
```diff
@@ -0,0 +1,165 @@
...
+  }
+  else
+#endif // _CCCL_HAS_INT128()
+  {
+    if constexpr (sizeof(_Tp) <= sizeof(int))
+    {
+      if (!::cuda::std::__cccl_default_is_constant_evaluated())
+      {
+        NV_IF_ELSE_TARGET(NV_IS_HOST,
+                          (return ::cuda::__ffs_impl_host(static_cast<uint32_t>(__v));),
```
fbusato: `static_cast` is not needed for `__ffs_impl_host` and `__ffs_impl_device`

libcudacxx/include/cuda/__bit/ffs.h
```diff
@@ -0,0 +1,165 @@
...
+                          (return ::cuda::__ffs_impl_device(static_cast<uint32_t>(__v));));
+      }
+      return ::cuda::__ffs_impl_constexpr(static_cast<uint32_t>(__v));
+    }
+    else
+    {
+      if (!::cuda::std::__cccl_default_is_constant_evaluated())
```
fbusato: would be possible to merge this branch with the previous one?

libcudacxx/include/cuda/__bit/ffs.h
```diff
@@ -0,0 +1,165 @@
...
+    if (__lo != 0)
+    {
+      if (!::cuda::std::__cccl_default_is_constant_evaluated())
+      {
+        NV_IF_ELSE_TARGET(
+          NV_IS_HOST, (return ::cuda::__ffs_impl_host(__lo);), (return ::cuda::__ffs_impl_device(__lo);));
+      }
+      return ::cuda::__ffs_impl_constexpr(__lo);
+    }
+    const auto __hi = static_cast<uint64_t>(__v >> 64);
```
fbusato: prefer an unsigned type. Signed shift is undefined behavior (before C++20)

libcudacxx/include/cuda/__bit/ffs.h
```diff
@@ -0,0 +1,165 @@
...
+    if (__hi != 0)
```
fbusato: this condition is not needed.  `__ffs_impl_host/device` already return 0 for input `0`

libcudacxx/include/cuda/__bit/ffs.h
```diff
@@ -0,0 +1,165 @@
...
+  {
+    const auto __lo = static_cast<uint64_t>(__v);
+    if (__lo != 0)
+    {
+      if (!::cuda::std::__cccl_default_is_constant_evaluated())
+      {
+        NV_IF_ELSE_TARGET(
+          NV_IS_HOST, (return ::cuda::__ffs_impl_host(__lo);), (return ::cuda::__ffs_impl_device(__lo);));
+      }
+      return ::cuda::__ffs_impl_constexpr(__lo);
```
fbusato: consider also merging this pattern in a separate function

libcudacxx/include/cuda/__bit/ffs.h
```diff
@@ -0,0 +1,165 @@
...
+  unsigned char __res{};
+  if constexpr (sizeof(_Tp) <= sizeof(uint32_t))
+  {
+    __res = ::_BitScanForward(&__where, static_cast<uint32_t>(__v));
+  }
+  else
+  {
+    __res = ::_BitScanForward64(&__where, static_cast<uint64_t>(__v));
+  }
+  return (__res) ? static_cast<int>(__where) + 1 : 0;
```
fbusato:
```suggestion
  return tatic_cast<int>(__where) + (__res != 0);
```

libcudacxx/include/cuda/__bit/ffs.h
```diff
@@ -0,0 +1,165 @@
...
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_CLANG)
+#  pragma clang system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_MSVC)
+#  pragma system_header
+#endif // no system header
+
+#include <cuda/std/__type_traits/is_constant_evaluated.h>
+#include <cuda/std/__type_traits/is_unsigned_integer.h>
+#include <cuda/std/cstdint>
+#include <cuda/std/limits>
```
fbusato: add `nvtarget`

libcudacxx/include/cuda/__bit/ffs.h
```diff
@@ -0,0 +1,165 @@
...
+
+_CCCL_BEGIN_NAMESPACE_CUDA
+
+#if _CCCL_HAS_BUILTIN(__builtin_ffs) || _CCCL_COMPILER(GCC)
+#  define _CCCL_BUILTIN_FFS(...)   __builtin_ffs(__VA_ARGS__)
+#  define _CCCL_BUILTIN_FFSLL(...) __builtin_ffsll(__VA_ARGS__)
+#endif // _CCCL_HAS_BUILTIN(__builtin_ffs) || _CCCL_COMPILER(GCC)
+
+template <typename _Tp>
+[[nodiscard]] _CCCL_API constexpr int __ffs_impl_constexpr(_Tp __v) noexcept
```
fbusato: `_Tp` must be an unsigned type to avoid UB with signed bitwise ops

libcudacxx/include/cuda/__bit/ffs.h
```diff
@@ -0,0 +1,165 @@
...
+    return _CCCL_BUILTIN_FFS(static_cast<int>(__v));
+  }
+  else
+  {
+    return _CCCL_BUILTIN_FFSLL(static_cast<long long>(__v));
+  }
+#  elif _CCCL_COMPILER(MSVC)
+  unsigned long __where{};
+  unsigned char __res{};
+  if constexpr (sizeof(_Tp) <= sizeof(uint32_t))
```
fbusato: I would also add
```cpp
using ::cuda::std::uint32_t
using ::cuda::std::uint64_t
```

libcudacxx/include/cuda/__bit/ffs.h
```diff
@@ -0,0 +1,165 @@
...
+#if defined(_CCCL_IMPLICIT_SYSTEM_HEADER_GCC)
+#  pragma GCC system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_CLANG)
+#  pragma clang system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_MSVC)
+#  pragma system_header
+#endif // no system header
+
+#include <cuda/std/__type_traits/is_constant_evaluated.h>
+#include <cuda/std/__type_traits/is_unsigned_integer.h>
```
fbusato, line 25: looks unused

libcudacxx/include/cuda/__bit/ffs.h
```diff
@@ -0,0 +1,165 @@
...
+    }
+    else
+    {
+      if (!::cuda::std::__cccl_default_is_constant_evaluated())
+      {
+        NV_IF_ELSE_TARGET(NV_IS_HOST,
+                          (return ::cuda::__ffs_impl_host(static_cast<uint64_t>(__v));),
+                          (return ::cuda::__ffs_impl_device(static_cast<uint64_t>(__v));));
+      }
+      return ::cuda::__ffs_impl_constexpr(static_cast<uint64_t>(__v));
```
fbusato: please also add `_CCCL_ASSUME(result >= 0 && result <=32 )`

Aminsed:
@fbusato 

* Removed unnecessary casts
* Fixed unsigned shift (avoid UB)
* Added static_assert and _CCCL_ASSUME
* Simplified MSVC ternary

Skipped:
* Using declarations (bit_reverse.h doesn't use them)
* Helper extraction (conflicts with @davebayer's flatten request)

Aminsed:
@fbusato @miscco @davebayer Is there any changes I could make to make this PR merge ready?

miscco:
@Aminsed sorry about the delay, this is just a matter of prioritization

fbusato:
@Aminsed please ping us when you make progress. It is very easy to miss updates on the PR from our side. 
Please also mark the comments as resolved after you address them. Thanks again for the contribution.

libcudacxx/include/cuda/__bit/ffs.h
```diff
@@ -0,0 +1,175 @@
...
+
+    if (__lo != 0)
+    {
+      return ::cuda::__ffs_impl_constexpr(__lo);
+    }
+    if (__hi != 0)
+    {
+      return ::cuda::__ffs_impl_constexpr(__hi) + 64;
+    }
+    return 0;
```
davebayer:
```suggestion
    const auto __lo = static_cast<::cuda::std::uint64_t>(__v);
    const auto __hi = static_cast<::cuda::std::uint64_t>(static_cast<__uint128_t>(__v) >> 64);
    
    if (const auto __result = ::cuda::ffs(__lo) != 0)
    {
      return __result;
    }
    if (const auto __result = ::cuda::ffs(__hi) != 0)
    {
      return __result + 64;
    }
    return 0;
```
can be simplified to

libcudacxx/include/cuda/__bit/ffs.h
```diff
@@ -0,0 +1,175 @@
...
+  }
+  else
+#endif // _CCCL_HAS_INT128()
+  {
+    int __result{};
+    if constexpr (sizeof(_Tp) <= sizeof(int))
+    {
+      const auto __v32 = static_cast<uint32_t>(__v);
```
davebayer:
```suggestion
      const auto __v32 = static_cast<::cuda::std::uint32_t>(__v);
```
Please, fully qualify these types

docs/libcudacxx/extended_api/bit/ffs.rst
```diff
@@ -0,0 +1,67 @@
+.. _libcudacxx-extended-api-bit-ffs:
+
+``cuda::ffs``
+=============
+
+.. code:: cpp
+
+   template <typename T>
+   [[nodiscard]] constexpr int
+   ffs(T value) noexcept;
```
davebayer:
```suggestion
   namespace cuda {
   
   template <class T>
   [[nodiscard]] __host__ __device__ constexpr
   int ffs(T value) noexcept;
   
   } // namespace cuda
```

docs/libcudacxx/extended_api/bit.rst
```diff
@@ -40,3 +41,8 @@ Bit
      - Extract a bitfield
      - CCCL 3.0.0
      - CUDA 13.0
+
+   * - :ref:`ffs <libcudacxx-extended-api-bit-ffs>`
+     - Find first set bit
+     - CCCL 3.0.0
+     - CUDA 13.0
```
davebayer:
```suggestion
     - CCCL 3.2.0
     - CUDA 13.2
```

libcudacxx/include/cuda/__bit/ffs.h
```diff
@@ -0,0 +1,171 @@
...
+#  include <intrin.h>
+#endif // _CCCL_COMPILER(MSVC)
+
+#include <cuda/std/__cccl/prologue.h>
+
+_CCCL_BEGIN_NAMESPACE_CUDA
+
+using ::cuda::std::uint32_t;
+using ::cuda::std::uint64_t;
+
```
davebayer:
```suggestion
```
We don't want to import these types to `cuda::` namespace

libcudacxx/include/cuda/__bit/ffs.h
```diff
@@ -0,0 +1,171 @@
...
+  {
+    __v >>= 1;
+    ++__pos;
+  }
+  return __pos;
+}
+
+#if !_CCCL_COMPILER(NVRTC)
+template <typename _Tp>
+[[nodiscard]] _CCCL_HIDE_FROM_ABI _CCCL_HOST int __ffs_impl_host(_Tp __v) noexcept
```
davebayer:
```suggestion
[[nodiscard]] _CCCL_HOST_API int __ffs_impl_host(_Tp __v) noexcept
```

libcudacxx/include/cuda/__bit/ffs.h
```diff
@@ -0,0 +1,171 @@
...
+  return __res ? (static_cast<int>(__where) + 1) : 0;
+#  else
+  return ::cuda::__ffs_impl_constexpr(__v);
+#  endif // _CCCL_COMPILER(MSVC)
+}
+#endif // !_CCCL_COMPILER(NVRTC)
+
+#if _CCCL_CUDA_COMPILATION()
+template <typename _Tp>
+[[nodiscard]] _CCCL_HIDE_FROM_ABI _CCCL_DEVICE int __ffs_impl_device(_Tp __v) noexcept
```
davebayer:
```suggestion
[[nodiscard]] _CCCL_DEVICE_API int __ffs_impl_device(_Tp __v) noexcept
```

libcudacxx/include/cuda/__bit/ffs.h
```diff
@@ -0,0 +1,171 @@
...
+      if (!::cuda::std::__cccl_default_is_constant_evaluated())
+      {
+        NV_IF_ELSE_TARGET(
+          NV_IS_HOST, (__result = ::cuda::__ffs_impl_host(__v64);), (__result = ::cuda::__ffs_impl_device(__v64);));
+      }
+      else
+      {
+        __result = ::cuda::__ffs_impl_constexpr(__v64);
+      }
+    }
```
davebayer: You do something like `using _Up = ::cuda::std::conditional_t<sizeof(_Tp) == sizeof(::cuda::std::uint64_t), ::cuda::std::uint64_t, ::cuda::std::uint32_t>` and merge these branches together

docs/libcudacxx/extended_api/bit/ffs.rst
```diff
@@ -0,0 +1,69 @@
...
+
+- For non-zero values: ``ffs(x) == countr_zero(x) + 1``
+
+**Performance considerations**
+
+The function performs the following operations in device code:
+
+- ``uint32_t``: Single ``FFS`` instruction
+- ``uint64_t``: Single ``FFSLL`` instruction
+- ``__uint128_t``: Two ``FFSLL`` instructions with conditional logic
```
davebayer: This is not true, there are no such SASS instructions

Aminsed:
@fbusato TIA for your feedback

docs/libcudacxx/extended_api/bit/ffs.rst
fbusato: this suggestion has not been applied

docs/libcudacxx/extended_api/bit/ffs.rst
```diff
@@ -0,0 +1,69 @@
+.. _libcudacxx-extended-api-bit-ffs:
+
+``cuda::ffs``
+=============
+
```
fbusato, line 5: please add "Defined in the ``<cuda/bit>`` header.

docs/libcudacxx/extended_api/bit/ffs.rst
```diff
@@ -0,0 +1,69 @@
...
+- For non-zero values: ``ffs(x) == countr_zero(x) + 1``
+
+**Performance considerations**
+
+The function performs the following operations:
+
+- Device:
+
+  - ``uint8_t``, ``uint16_t``, ``uint32_t``: ``FFS``
```
fbusato:
```suggestion
  - ``uint8_t``, ``uint16_t``, ``uint32_t``: ``BREV``, ``FLO``, ``IADD3``
```
I would omit the other cases

docs/libcudacxx/extended_api/bit/ffs.rst
```diff
@@ -0,0 +1,69 @@
...
+  - ``uint64_t``: ``FFSLL``
+  - ``uint128_t``: ``FFSLL`` x2 with conditional logic
+
+- Host:
+
+  - GCC/Clang: ``__builtin_ffs`` / ``__builtin_ffsll``
+  - MSVC: ``_BitScanForward`` / ``_BitScanForward64``
+  - Other: Portable constexpr loop implementation
```
fbusato:
```suggestion
```

docs/libcudacxx/extended_api/bit/ffs.rst
```diff
@@ -0,0 +1,69 @@
...
+
+.. note::
```
fbusato: I would suggest removing this note. It doesn't look aligned with other similar functions

libcudacxx/include/cuda/__bit/ffs.h
```diff
@@ -0,0 +1,156 @@
...
+#if _CCCL_HAS_BUILTIN(__builtin_ffs) || _CCCL_COMPILER(GCC)
+#  define _CCCL_BUILTIN_FFS(...)   __builtin_ffs(__VA_ARGS__)
+#  define _CCCL_BUILTIN_FFSLL(...) __builtin_ffsll(__VA_ARGS__)
+#endif // _CCCL_HAS_BUILTIN(__builtin_ffs) || _CCCL_COMPILER(GCC)
+
+template <typename _Tp>
+[[nodiscard]] _CCCL_HIDE_FROM_ABI constexpr int __ffs_impl_constexpr(_Tp __v) noexcept
+{
+  static_assert(::cuda::std::__cccl_is_unsigned_integer_v<_Tp>, "_Tp must be unsigned");
+
```
fbusato:
```suggestion
```

libcudacxx/include/cuda/__bit/ffs.h
```diff
@@ -0,0 +1,156 @@
...
+  if (__v == 0)
+  {
+    return 0;
+  }
+
```
fbusato:
```suggestion
```

libcudacxx/include/cuda/__bit/ffs.h
```diff
@@ -0,0 +1,156 @@
...
+  while ((__v & 1) == 0)
+  {
+    __v >>= 1;
+    ++__pos;
+  }
+  return __pos;
+}
+
+#if !_CCCL_COMPILER(NVRTC)
+template <typename _Tp>
```
fbusato:
```suggestion

template <typename _Tp>
```

libcudacxx/include/cuda/__bit/ffs.h
```diff
@@ -0,0 +1,156 @@
...
+  else
+  {
+    __res = ::_BitScanForward64(&__where, static_cast<::cuda::std::uint64_t>(__v));
+  }
+  return __res ? (static_cast<int>(__where) + 1) : 0;
+#  else
+  return ::cuda::__ffs_impl_constexpr(__v);
+#  endif // _CCCL_COMPILER(MSVC)
+}
+#endif // !_CCCL_COMPILER(NVRTC)
```
fbusato:
```suggestion

#endif // !_CCCL_COMPILER(NVRTC)
```

libcudacxx/include/cuda/__bit/ffs.h
```diff
@@ -0,0 +1,156 @@
...
+
+#if _CCCL_CUDA_COMPILATION()
```
fbusato, line 64:
```suggestion
#if _CCCL_CUDA_COMPILATION()

```

libcudacxx/include/cuda/__bit/ffs.h
```diff
@@ -0,0 +1,156 @@
...
+  if constexpr (sizeof(_Tp) <= sizeof(int))
+  {
+    return ::__ffs(static_cast<int>(__v));
+  }
+  else
+  {
+    return ::__ffsll(static_cast<long long>(__v));
+  }
+}
+#endif // _CCCL_CUDA_COMPILATION()
```
fbusato, line 93:
```suggestion

#endif // _CCCL_CUDA_COMPILATION()
```

libcudacxx/include/cuda/__bit/ffs.h
```diff
@@ -0,0 +1,156 @@
...
+    {
+      return __result + 64;
+    }
+    return 0;
+  }
+  else
+#endif // _CCCL_HAS_INT128()
+  {
+    using _Up = ::cuda::std::
+      conditional_t<sizeof(_Tp) == sizeof(::cuda::std::uint64_t), ::cuda::std::uint64_t, ::cuda::std::uint32_t>;
```
fbusato: `conditional_t` is expensive. Is it really needed?

libcudacxx/include/cuda/std/__bit/integral.h
```diff
@@ -22,6 +22,7 @@
 #endif // no system header
 
 #if _CCCL_CUDA_COMPILATION()
+#  include <cuda/__bit/ffs.h>
```
fbusato: The right header is cuda/bit only. Remove this line

Aminsed:
@fbusato Thank you for the feedback! ran into a bit of an issue with clang-format lint and had to remove some blank lines. Happy to to revert but lint check might fail.

Aminsed:
@fbusato had to make these [changes](https://github.com/NVIDIA/cccl/pull/6192/commits/2b54a3d1e90eef15a6af7fe03d22d7c2347ad5e3) to be able to build successfully locally. Thank you for your help

libcudacxx/include/cuda/__bit/ffs.h
```diff
@@ -0,0 +1,165 @@
...
+  {
+    __res = ::_BitScanForward(&__where, static_cast<uint32_t>(__v));
+  }
+  else
+  {
+    __res = ::_BitScanForward64(&__where, static_cast<uint64_t>(__v));
+  }
+  return __res ? (static_cast<int>(__where) + 1) : 0;
+#  else
+  return __ffs_impl_constexpr(__v);
```
fbusato: this needs to be fully qualified

libcudacxx/include/cuda/__bit/ffs.h
```diff
@@ -0,0 +1,165 @@
...
+_CCCL_REQUIRES(::cuda::std::__cccl_is_unsigned_integer_v<_Tp>)
+[[nodiscard]] _CCCL_API constexpr int ffs(_Tp __v) noexcept
+{
+#if _CCCL_HAS_INT128()
+  if constexpr (sizeof(_Tp) == sizeof(__uint128_t))
+  {
+    const auto __lo = static_cast<uint64_t>(__v);
+    const auto __hi = static_cast<uint64_t>(static_cast<__uint128_t>(__v) >> 64);
+
+    if (const auto __result = ffs(__lo))
```
fbusato: same here

libcudacxx/include/cuda/__bit/ffs.h
```diff
@@ -0,0 +1,165 @@
...
+  }
+  else
+#endif // _CCCL_HAS_INT128()
+  {
+    int __result{};
+    if (!::cuda::std::__cccl_default_is_constant_evaluated())
+    {
+      if constexpr (sizeof(_Tp) <= sizeof(uint32_t))
+      {
+        const uint32_t __vu = static_cast<uint32_t>(__v);
```
fbusato: also `uint32_t` and similar need full qualification. You can also use `using ::cuda::std::uint32_t;`

Aminsed:
@fbusato 

For verification, I pulled nvidia/cuda:13.0.2-devel-ubuntu22.04 (NVCC 13.0.88) and built/ran libcudacxx/test/libcudacxx/cuda/bit/ffs.pass.cpp; it passes, so the new cuda::ffs path behaves as expected with the latest toolchain.

What I've learned from the CI run: If we re-run the full libcudacxx presets, CMake still aborts with “cuda_std_20 is not known to CUDA compiler” (NVCC 12.5 and even 13.0 don’t expose that compile feature yet).

jrhemstad:
pre-commit.ci autofix

jrhemstad:
/ok to test [fd5edab](https://github.com/NVIDIA/cccl/pull/6192/commits/fd5edaba402465618ee28ba32c1b3e9f75ad2bdc)

Aminsed:
Quick update: I re-ran the full libcudacxx, Thrust, CUB, and cudax presets in the rapidsai/devcontainers:25.12-cpp-llvm20-cuda13.0 image and they’re all green now. I also kicked off the c-parallel preset, but it still fails on existing issues (Thrust warnings treated as errors for unused parameters, plus lambda/index_sequence bugs in c/parallel/src/jit_templates). Those failures happen on current main as well, so I didn’t try to patch them here. Could you please rerun CI when you get a chance?
@fbusato @jrhemstad Thanks

Aminsed:
The latest CI failures were NVRTC-related: the __ffs_host* functions were missing _CCCL_DEVICE annotations for JIT mode, and I had accidentally used <cstddef> instead of <cuda/std/cstddef> which NVRTC can't find. 
@fbusato @miscco Would you mind triggering another CI run when you get a chance?

libcudacxx/include/cuda/__barrier/barrier_block_scope.h
```diff
@@ -101,7 +103,10 @@ class barrier<thread_scope_block, ::cuda::std::__empty_completion> : public __bl
   {
     NV_IF_TARGET(NV_PROVIDES_SM_80,
                  (if (::cuda::device::is_object_from(__barrier, ::cuda::device::address_space::shared)) {
-                   ::cuda::ptx::mbarrier_inval(__native_handle());
+                   // TODO(bgruber): expose mbarrier.inval.shared in cuda::ptx
```
fbusato: why are we changing the code of this function? it looks out of scope

libcudacxx/include/cuda/__bit/ffs.h
```diff
@@ -0,0 +1,213 @@
...
+  while ((__v & 1) == 0)
+  {
+    __v >>= 1;
+    ++__pos;
+  }
+  return __pos;
+}
+
+#if _CCCL_COMPILER(NVRTC)
+[[nodiscard]] _CCCL_DEVICE inline int __ffs_host32(::cuda::std::uint32_t __v) noexcept
```
fbusato: I think that merging all `__ffs_host*` functions with `if constexpr (sizeof(T) == X)` would result in a shorted/cleaner code, or at least the constant evaluation part

libcudacxx/include/cuda/__bit/ffs.h
```diff
@@ -0,0 +1,213 @@
...
+{
+  return ::cuda::__detail::__ffs_constexpr_impl(__v);
+}
+#else
+[[nodiscard]] _CCCL_HOST_API inline int __ffs_host32(::cuda::std::uint32_t __v) noexcept
+{
+  if (::cuda::std::__cccl_default_is_constant_evaluated())
```
fbusato: please use the new macro based on `consteval`

libcudacxx/include/cuda/__bit/ffs.h
```diff
@@ -0,0 +1,213 @@
...
+  {
+    return __result + 64;
+  }
+  return 0;
+}
+#  endif // _CCCL_COMPILER(NVRTC)
+#endif // _CCCL_HAS_INT128()
+
+#if _CCCL_CUDA_COMPILATION()
+[[nodiscard]] _CCCL_DEVICE_API inline int __ffs_device32(::cuda::std::uint32_t __v) noexcept
```
fbusato, line 66: same here. We could merge the function bodies

libcudacxx/include/cuda/__bit/ffs.h
```diff
@@ -0,0 +1,213 @@
...
+#  endif // _CCCL_HAS_INT128()
+#endif // _CCCL_CUDA_COMPILATION()
+} // namespace __detail
+
+template <class _Tp, ::cuda::std::enable_if_t<::cuda::std::__cccl_is_cv_unsigned_integer_v<_Tp>, int> = 0>
+[[nodiscard]] _CCCL_API constexpr int ffs(_Tp __v) noexcept
+{
+  using _Unsigned = ::cuda::std::remove_cv_t<_Tp>;
+
+#if !defined(_CCCL_BUILTIN_IS_CONSTANT_EVALUATED)
```
fbusato: we should not check the built-in here. We can use `consteval` @davebayer

libcudacxx/include/cuda/__bit/ffs.h
```diff
@@ -0,0 +1,213 @@
...
+  return __detail::__ffs_constexpr_impl(static_cast<_Unsigned>(__v));
```
fbusato: please use the full namespace

libcudacxx/include/cuda/__bit/ffs.h
```diff
@@ -0,0 +1,213 @@
...
+  {
+    NV_IF_ELSE_TARGET(NV_IS_HOST,
+                      (return __detail::__ffs_host32(static_cast<::cuda::std::uint32_t>(__v));),
+                      (return __detail::__ffs_device32(static_cast<::cuda::std::uint32_t>(__v));))
+  }
+  else
+  {
+    NV_IF_ELSE_TARGET(NV_IS_HOST,
+                      (return __detail::__ffs_host64(static_cast<::cuda::std::uint64_t>(__v));),
+                      (return __detail::__ffs_device64(static_cast<::cuda::std::uint64_t>(__v));))
```
fbusato: all these functions could be merged after the previous suggestions

libcudacxx/include/cuda/std/__cccl/builtin.h
```diff
@@ -147,6 +147,10 @@
 #  define _CCCL_BUILTIN_IS_CONSTANT_EVALUATED(...) __builtin_is_constant_evaluated(__VA_ARGS__)
 #endif // _CCCL_CHECK_BUILTIN(builtin_is_constant_evaluated)
 
+#if _CCCL_HOST_COMPILATION() && _CCCL_COMPILER(GCC, <, 9)
```
fbusato: please don't change the built-in checks

libcudacxx/include/cuda/std/__type_traits/is_constant_evaluated.h
```diff
@@ -24,23 +24,33 @@
 
 _CCCL_BEGIN_NAMESPACE_CUDA_STD
 
-[[nodiscard]] _CCCL_API constexpr bool is_constant_evaluated() noexcept
-{
 #if defined(_CCCL_BUILTIN_IS_CONSTANT_EVALUATED)
+_CCCL_API constexpr bool is_constant_evaluated() noexcept
```
fbusato: this file should not be modified

libcudacxx/include/cuda/std/__type_traits/is_trivially_destructible.h
```diff
@@ -35,18 +36,42 @@
...
-inline constexpr bool is_trivially_destructible_v = is_destructible_v<_Tp> && _CCCL_BUILTIN_HAS_TRIVIAL_DESTRUCTOR(_Tp);
+template <class _Tp, class>
+struct __is_trivially_destructible_base
+    : bool_constant<is_destructible_v<_Tp> && _CCCL_BUILTIN_HAS_TRIVIAL_DESTRUCTOR(_Tp)>
+{};
 
 #endif // ^^^ !_CCCL_BUILTIN_IS_TRIVIALLY_DESTRUCTIBLE ^^^
 
+template <class _Tp>
+struct __is_trivially_destructible_impl : __is_trivially_destructible_base<_Tp>
```
fbusato: why you need this change?

libcudacxx/test/libcudacxx/cuda/bit/ffs.pass.cpp
```diff
@@ -0,0 +1,114 @@
...
+  {
+    assert_ffs(T(128), 8);
+    assert_ffs(T(256), 9);
+    assert_ffs(T(512), 10);
+    assert_ffs(T(1024), 11);
+  }
+
+  if constexpr (sizeof(T) >= 4)
+  {
+    assert_ffs(T(0x80000000u), 32);
```
fbusato, line 88: I also would like to see a test for the msb in both run-time and compile-time

fbusato (changes requested):
I see there are multiple changes not related to `ffs`. Let me update the branch to align the code with the latest version

libcudacxx/test/libcudacxx/force_include.h
```diff
@@ -102,10 +102,14 @@ int main(int argc, char** argv)
     };
     CUDA_CALL(err, cudaLaunchKernelEx(&config, fake_main_kernel, cuda_ret));
   }
-  else
+  else if (cuda_thread_count > 0)
```
fbusato: I guess this is also unrelated

libcudacxx/test/libcudacxx/std/iterators/predef.iterators/move.iterators/move.sentinel/base.pass.cpp
```diff
@@ -13,6 +13,8 @@
 
 // constexpr S base() const;
 
+#include <cuda/std/detail/__config>
```
fbusato: same here

libcudacxx/include/cuda/__bit/ffs.h
```diff
@@ -0,0 +1,174 @@
...
+    return __result;
+  }
+  const auto __hi = static_cast<::cuda::std::uint64_t>(__v >> 64);
+  if (const int __result = ::cuda::__detail::__ffs_device64(__hi))
+  {
+    return __result + 64;
+  }
+  return 0;
+}
+#  endif // _CCCL_HAS_INT128()
```
fbusato, line 91: can we try to merge these functions?

Aminsed:
@fbusato Addressed your feedback:

- Reverted `force_include.h` changes (unrelated test infrastructure)
- Reverted `base.pass.cpp` changes (unrelated test workaround)

The PR now only contains changes directly related to `cuda::ffs`:
- `docs/libcudacxx/extended_api/bit.rst` - doc index
- `docs/libcudacxx/extended_api/bit/ffs.rst` - new docs
- `libcudacxx/include/cuda/__bit/ffs.h` - implementation
- `libcudacxx/include/cuda/bit` - include
- `libcudacxx/include/cuda/pipeline` - usage of cuda::ffs
- `libcudacxx/test/libcudacxx/cuda/bit/ffs.pass.cpp` - tests

All tests verified locally:
- bitfield 
- bitmask 
- bit_reverse 
- ffs 

Regarding the function merging suggestion on line 132 - the current pattern follows the existing structure in the codebase. Happy to refactor if you prefer the `if constexpr` template pattern used in `bit_reverse.h`.

Ready for CI when you have a chance!

fbusato (approved):
looks good! I think the refactoring is not mandatory. The code in `ffs.h` is well-organized at the end.

PR #6195, opened 2025-10-13
We should guard the host library include wrappers so that we can unconditionally include the headers with NVRTC

miscco:


davebayer:
pre-commit.ci autofix

PR #6198, opened 2025-10-13
Bypass allocator in `thrust::device_delete`

bernhardmgruber:
This allows us to re-enable a test


bernhardmgruber:
clang in CUDA mode correctly observed that we construct `Foo` on the host, even though the constructor is only `__device__`. I fixed this as well.

Issue #5604, opened 2025-08-20
[FEA]: Annotate cub device functions with `nodiscard`.


trivialfis:
### Is this a duplicate?

- [x] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cccl/issues) for this request and that I agree to the [Code of Conduct](CODE_OF_CONDUCT.md)

### Area

CUB

### Is your feature request related to a problem? Please describe.

It's for me to forget about checking the return status `cudaErrort_t` from cub device functions like `cub::DeviceSegmentedReduce::Sum`, and most of the [snippets](https://nvidia.github.io/cccl/cub/api/structcub_1_1DeviceReduce.html#id1) in the documents don't check the return. It would be great if cub could annotate its device functions with `[[nodiscard]]`.

### Describe the solution you'd like

- Annotate the device functions with `[[nodiscard]]`
- Update document snippets to check the return.

miscco:
As discussed internally, I am strongly against breaking user code for this. 

We can potentially add a macro like `_CCCL_NODISCARD_ERROR` that the user can disable via a feature flag.

bernhardmgruber:
I fully support this idea! I also understand that this may break some users relying on `-Werror`. I think we should definitely address this for CCCL 4.0.

In the meantime, I am open to review and merge a PR that uses a macro containing `[[no_discard]]` to annotate all CUB functions which can be disabled by the user.

PR #6204, opened 2025-10-13
Add environment overloads for DeviceTransform

bernhardmgruber:
Fixes: #6203

### DIFF
--- a/cub/test/catch2_test_device_transform_env.cu
+++ b/cub/test/catch2_test_device_transform_env.cu
@@ -13,6 +13,13 @@ namespace stdexec = cuda::std::execution;
 
 using namespace thrust::placeholders;
 
+auto make_stream_env(cudaStream_t stream)
+{
+  // MSVC has trouble nesting two aggregate initializations with CTAD
+  auto stream_prop = stdexec::prop{cuda::get_stream, cuda::stream_ref{stream}};
+  return stdexec::env{cuda::std::move(stream_prop)};
+}
+
 C2H_TEST("DeviceTransform::Transform custom stream", "[device][transform]")
 {
   using type          = int;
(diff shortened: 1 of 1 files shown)


cub/cub/device/device_transform.cuh
```diff
@@ -66,18 +75,28 @@ private:
...
+             num_items,
+             ::cuda::std::move(predicate),
+             ::cuda::std::move(transform_op),
+             stream);
+  }
+
+  template <typename Env>
+  CUB_RUNTIME_FUNCTION static auto get_stream(Env env) -> cudaStream_t
+  {
+    return _CUDA_STD_EXEC::__query_or(env, ::cuda::get_stream, ::cuda::stream_ref{cudaStream_t{}}).get();
```
NaderAlAwar: Suggestion:

```suggestion
    return ::cuda::std::execution::__query_or(env, ::cuda::get_stream, ::cuda::stream_ref{cudaStream_t{}}).get();
```

cub/cub/device/device_transform.cuh
```diff
@@ -108,18 +127,28 @@ public:
...
-  //! @param stream **[optional]** CUDA stream to launch kernels within. Default is stream\ :sub:`0`.
-  template <typename... RandomAccessIteratorsIn, typename RandomAccessIteratorOut, typename NumItemsT, typename TransformOp>
+  //! @param env Execution environment, or cudaStream_t. Default is ``cuda::std::execution::env{}``, which will run on
+  //! stream\ :sub:`0`
+  template <typename... RandomAccessIteratorsIn,
+            typename RandomAccessIteratorOut,
+            typename NumItemsT,
+            typename TransformOp,
+            typename Env = ::cuda::std::execution::env<>>
   CUB_RUNTIME_FUNCTION static cudaError_t Transform(
```
NaderAlAwar, line 137: Question: are we replacing the old non-env API with the new one completely? I thought we were keeping both. If we are keeping both, then the new API should be marked with `[[nodiscard]]`

cub/cub/device/device_transform.cuh
bernhardmgruber: Agreed! I just took code from DeviceReduce without thinking.

cub/cub/device/device_transform.cuh
bernhardmgruber: Long story short: it saves a lot of code duplication to route the `cudaStream_t` through the same overload as the environment. 

Would you be fine if I omit the `[[nodiscard]]` in this PR and we just bulk introduce it as part of #5604?

NaderAlAwar: Makes sense, yeah that's fine

PR #6208, opened 2025-10-13
[DOC] Add temp_storage_bytes usage guide

### LINKED ISSUE #847: Documentation (& code) regarding determining temp_storage_bytes not very clear
I would like to be able to determine my device-wide primitive's temp_storage_bytes before I have all of the primitive's arguments ready. The interface for obtaining it ostensibly requires everything to be ready for the actual run - and the documentation does not make it clear what arguments it actually needs and what it's going to do with them (e.g. will it look at the input at all, in any way? Probably not, but who knows)

Also, delving into the source (in my case: `dispatch_select_if.cuh`), I see:

```
if (d_temp_storage == NULL)
{
    // Return if the caller is simply requesting the size of the storage allocation
    break;
}
```

which is super-weird. What if I passed a pointer to a size_t containing 2^64-1? Would my code be led to believe it has to allocate that much? Hmm.

Aminsed:
Add documentation explaining temp_storage_bytes query pattern.

* Clarify two-phase usage pattern
* Document what arguments are required vs optional
* Explain that pointers can be nullptr during query
* Add example showing proper usage

Fixes #847

### DIFF
--- a/docs/cub/api_docs/device_wide.rst
+++ b/docs/cub/api_docs/device_wide.rst
@@ -44,7 +44,7 @@ Environment-based overloads are rolling out across CUB device-wide primitives. T
 Key properties of the environment argument:
 
 - It is defaulted and appears as the last argument.
-- Streams can be forwarded by specifying ``cuda::get_stream`` properties.
+- Streams can be specified with ``cuda::get_stream`` properties.
 - You can select the memory resource (CCCL-provided or custom) used for internal allocations.
 - Supported algorithms accept determinism requirements (for example, ``cuda::execution::determinism::gpu_to_gpu``).
 - Multiple properties compose into a single centralized argument.


docs/cub/api_docs/device_wide.rst
```diff
@@ -11,6 +11,40 @@ Device-Wide Primitives
    ../api/device
 
 
+Determining Temporary Storage Requirements
```
fbusato, line 16: would be nice to also mention the single-phase API

docs/cub/api_docs/device_wide.rst
```diff
@@ -11,6 +11,47 @@ Device-Wide Primitives
...
+
+**What arguments are needed during the query phase?**
+
+* **Required**: Data types (via template parameters and iterator types) and problem size (``num_items``)
+* **Can be nullptr/uninitialized**: All input/output pointers (``d_in``, ``d_out``, etc.)
+* **Note**: The algorithm does not access input data during the query phase
+
+Example pattern:
+
+.. code-block:: c++
```
gonidelis, line 54: we stopped using `code-block` and rather use `literalinclude`s now linked to standalone .cu source examples.

docs/cub/api_docs/device_wide.rst
```diff
@@ -11,6 +11,47 @@ Device-Wide Primitives
...
+
+   // Run the actual algorithm with real pointers
+   cub::DeviceReduce::Sum(
+     d_temp_storage, temp_storage_bytes,
+     d_in, d_out, num_items);
+
+**Single-Phase API** (Environment-Based)
+
+Some algorithms provide environment-based overloads that eliminate the two-phase call pattern.
+These APIs accept an execution environment parameter. See the individual algorithm documentation for availability.
```
gonidelis: `Some algorithms`: we are in the process of adding these APIs for all algorithms. Let's not make it sound like we selectively picked some algorithms to support this feature.

`eliminate the two-phase call pattern`: they eliminate the explicit allocation requirements which relives us from a) two phase call b) the two legacy storage arguments in the beginning of the parameter list of every device premitive.

`These APIs accept an execution environment parameter.`: I feel like there is much more to be said here. There is a blogpost coming out soon that can be used as inspo. Trying to sum it up briefly: 

1.  the execution environment arg is at the end **and is defaulted** which means users do not have to pass something explicitly to it.

2.  the execution environment arg can be used to select on which memory pool the primitive will be executed:

       i) there are existing memory resources provided by CCCL or

       ii) the user can create their own custom memory resource

3. in *some* algorithms the argument can be used to specify the deterministic requirements (*expand on det reqs*) of the algorithm

4. the user can still pass the stream as before in the templated execution environment arg

5. all of the above can be passed simultaneously in a centralized control argument manner (*show how*). 

There are more to be said but are not feature complete so we can avoid them for now.

gonidelis (changes requested):
single phase APIs open up a whole new space of explanation for the documentation - the two phase clarifications though are much needed. Thanks a lot for taking the time to provide docs for it.

Aminsed:
@gonidelis

docs/cub/api_docs/device_wide.rst
```diff
@@ -11,6 +11,73 @@ Device-Wide Primitives
...
+Most CUB device-wide algorithms follow a two-phase usage pattern:
+
+1. **Query Phase**: Call the algorithm with ``d_temp_storage = nullptr`` to determine required temporary storage size
+2. **Execution Phase**: Allocate storage and call the algorithm again to perform the actual operation
+
+**What arguments are needed during the query phase?**
+
+* **Required**: Data types (via template parameters and iterator types) and problem size (``num_items``)
+* **Can be nullptr/uninitialized**: All input/output pointers (``d_in``, ``d_out``, etc.)
+* **Note**: The algorithm does not access input data during the query phase
```
bernhardmgruber: Are we actually providing this guarantee? Can you point at a place from which you derive this fact?

Hypothetical scenario: we could determine the temporary storage based on the alignment of another input pointer. AFAIK we don't do that, but currently, we could.

However, since we seem to be vage about what's required on all parameters that are not taking part in the query phase, maybe we should just define what's being suggested here. But that requires some broader approval and probably a review of all existing APIs.

@gevtushenko what do you think?

Aminsed: @bernhardmgruber Thanks for flagging this great point. I re-audited the device-wide dispatch layer to make sure we're not overpromising. Every dispatcher we ship 

(`dispatch_reduce*.cuh`, `dispatch_scan*.cuh`, `dispatch_select_if.cuh`, 
`dispatch_histogram.cuh`, `dispatch_radix_sort.cuh`, `dispatch_merge*.cuh`, 
`dispatch_rle.cuh`, `dispatch_unique_by_key.cuh`, `dispatch_three_way_partition.cuh`, 
`dispatch_topk.cuh`, `dispatch_adjacent_difference.cuh`, `dispatch_batch_memcpy.cuh`) 

exits immediately when `d_temp_storage == nullptr`- no kernels launch and no user pointers are dereferenced. I've updated the “_What arguments are needed during the query phase?_” bullets to call that out explicitly and list the audited dispatchers. Please let me know if you’d like me to add anything else or tighten the wording further.

bernhardmgruber: I don't think we should list the specific implementations, but rather provide a general guarantee.

@gevtushenko can we agree that any arguments, except for the temporary storage pointer and size reference, are not inspected during a size query call of a CUB device API?

Aminsed: @bernhardmgruber Thanks! I’ve trimmed the doc to state the general guarantee. Happy to adjust further once we hear back from @gevtushenko.

gevtushenko: @Aminsed thank you for the contribution! 

It's a long overdue on our end to provide an overview of two-phase interface. We do receive questions regarding it, and I'm glad there'll be a single source of truth going forward. Access to the iterators during temp storage query stage is one of these frequent questions, so we should document it. 

## Iterator Access 

> Are we actually providing this guarantee? Can you point at a place from which you derive this fact?

Each device-level algorithm says something like "the required allocation size is written to `temp_storage_bytes` and no work is done". The only reasonable way to define "work" in the context of CUB is something involving driver: kernel invocation, data movement, memory space query etc. 

Users frequently have to estimate the temporary storage size before they have data, so it's common and valid use to estimate temp storage with `int *d_in;`(can be uninitialized, can be `nullptr` etc. in user code), which would break if we tried accessing iterators. So even if accessing these iterator made sense, we don't have a freedom to start doing that for existing algorithms. If we discover use case requiring access to iterators at query stage at any point, I'm confident that we'd have to go through guarantees API for user to explicitly tell us it's fine to do so. 

So the only comment I'd have is that current phrasing is a bit limiting: "The algorithm does not access input data during the query phase". It shouldn't be specific to input data, since we won't access output data as well. Maybe we can rephrase it in the lines of driver use / work etc. or at least relax phrazing to "does not access iterators". 

## Suggestions

### Doxygen Alias
To make this section more easily discoverable, consider adding a reference to it in the doxygen aliases:

https://github.com/NVIDIA/cccl/blob/93e1f87c25249cdf391091e0e56d2d52a6f00aac/docs/cub/Doxyfile#L74

This way, a link to this section will appear on each algorithm using `@devicestorage`.  If you'd like to have the PR scope smaller, feel free to file an issue so that this'd be addressed later, just make sure to link the issue here. 

### Current GPU
Another underspecified aspect is that current GPU probably shouldn't change between the phases. Say, in reduce, temp storage size depends on the occupancy, which might vary between GPUs. For radix sort, some architectures might use onesweep approach, while others would use legacy scheme. I think it's safer to have a requirement that same current GPU is used between phases by default and relax it on per-algorithm basis when needed. @elstehle, @bernhardmgruber what do you think?

bernhardmgruber: Thank you for the clarifications @gevtushenko! I would then word the requirement on the query phase something like this:

The query phase must use the same template instantiation than the execution phase of a two-step CUB API. That is, all template arguments are equal. Except for the first two function arguments (temporary storage pointer and size) as well as any problem size arguments (e.g. `num_items`), no arguments are accessed during the query phase and their values may be indeterminate.

bernhardmgruber: The requirement that the GPU is not allowed to be changed between query and execution phase is definitely should definitely be added as well! So let's also add:

The temporary storage size queried for a CUB algorithm is only valid for a CUB algorithm execution that is using the same current CUDA device as the query call.

Aminsed: Thanks @gevtushenko @bernhardmgruber for the detailed guidance. I’ve adopted that wording, query calls use the same template instantiation, only touch temp-storage/problem-size args (others may be indeterminate), and now note the same-current-device requirement. Happy to adjust further if needed.

elstehle: Agree with Georgii's assessment here. This, to me, is a very strong argument:
> Users frequently have to estimate the temporary storage size before they have data, so it's common and valid use to estimate temp storage with int *d_in;(can be uninitialized, can be nullptr etc. in user code)

Thanks, Bernhard. I think this is very concise. Just a nit, I would replace `than` with `as`:
> The query phase must use the same template instantiation ~~than~~ as the execution phase of a two-step CUB API. That is, all template arguments are equal. Except for the first two function arguments (temporary storage pointer and size) as well as any problem size arguments (e.g. num_items), no arguments are accessed during the query phase and their values may be indeterminate.

**Regarding current GPU:**
Agreed, that we should definitely document that we require the same current device being set during query and execution phase.

docs/cub/api_docs/device_wide.rst
```diff
@@ -11,6 +11,72 @@ Device-Wide Primitives
...
+.. _device-temp-storage:
+
+Determining Temporary Storage Requirements
+++++++++++++++++++++++++++++++++++++++++++++++++
+
+**Two-Phase API** (Traditional)
+
+Most CUB device-wide algorithms follow a two-phase usage pattern:
+
+1. **Query Phase**: Call the algorithm with ``d_temp_storage = nullptr`` to determine required temporary storage size
```
bernhardmgruber:
```suggestion
1. **Query Phase**: Call the algorithm with ``d_temp_storage = nullptr`` to determine the required temporary storage size
```

docs/cub/api_docs/device_wide.rst
```diff
@@ -11,6 +11,72 @@ Device-Wide Primitives
...
+2. **Execution Phase**: Allocate storage and call the algorithm again to perform the actual operation
+
+**What arguments are needed during the query phase?**
+
+* **Template instantiation**: The query call must use the same template arguments as the execution call; only ``d_temp_storage``, ``temp_storage_bytes``, and problem-size parameters (for example ``num_items``) differ.
```
bernhardmgruber: Critical: I believe this is not correct. All template arguments must be the same. `d_temp_storage` and `temp_storage_bytes` have fixed types, so we can remove them here. The type for `num_items` must be the same, since we may use more temporary storage for 64-bit sizes vs. 32-bit sizes.

```suggestion
* **Template instantiation**: The query call must use the same template arguments as the execution call.
```

docs/cub/api_docs/device_wide.rst
```diff
@@ -11,6 +11,72 @@ Device-Wide Primitives
...
+* **Argument access**: Aside from ``d_temp_storage``, ``temp_storage_bytes``, and the problem-size arguments, no parameters are inspected during the query phase, so their values may be indeterminate. The dispatch layer returns before launching kernels or touching user storage.
```
bernhardmgruber: Suggestion: I like `accessed` more than `inspected`:
```suggestion
* **Argument access**: Aside from ``d_temp_storage``, ``temp_storage_bytes``, and the problem-size arguments, no parameters are accessed during the query phase, so their values may be indeterminate. The dispatch layer returns before launching kernels or touching user storage.
```

docs/cub/api_docs/device_wide.rst
```diff
@@ -11,6 +11,72 @@ Device-Wide Primitives
...
+* **Current device**: The computed temporary storage size is valid only when the execution call runs on the same current CUDA device as the query. Re-run the query if the current device changes between phases.
```
bernhardmgruber, line 30: This sounds good to me.

docs/cub/api_docs/device_wide.rst
```diff
@@ -11,6 +11,72 @@ Device-Wide Primitives
...
+
+.. literalinclude:: ../../../cub/examples/device/example_device_reduce.cu
+   :language: c++
+   :dedent:
+   :start-after: example-begin temp-storage-query
+   :end-before: example-end temp-storage-query
+
+**Single-Phase API** (Environment-Based)
+
+Environment-based overloads are rolling out across CUB device-wide primitives. They remove the manual query/execute split by deriving temporary storage from an execution environment argument.
```
bernhardmgruber: I am not sure "deriving temporary storage" is accurate enough. Maybe:

```suggestion
Environment-based overloads are rolling out across CUB device-wide primitives. They remove the manual query/execute split by obtaining the temporary storage from a memory resource queried from the execution environment argument.
```

docs/cub/api_docs/device_wide.rst
```diff
@@ -11,6 +11,72 @@ Device-Wide Primitives
...
+
+Key properties of the environment argument:
+
+- It is defaulted and appears last, so traditional two-phase call sites remain valid without changes.
+- Streams can be forwarded by specifying ``cuda::get_stream`` properties.
```
bernhardmgruber: I am uncertain what "forwarding" should mean in this context. Can we use "passed" or "specified"?

docs/cub/api_docs/device_wide.rst
```diff
@@ -11,6 +11,72 @@ Device-Wide Primitives
...
+   :start-after: example-begin temp-storage-query
+   :end-before: example-end temp-storage-query
+
+**Single-Phase API** (Environment-Based)
+
+Environment-based overloads are rolling out across CUB device-wide primitives. They remove the manual query/execute split by deriving temporary storage from an execution environment argument.
+
+Key properties of the environment argument:
+
+- It is defaulted and appears last, so traditional two-phase call sites remain valid without changes.
```
bernhardmgruber: Important: I think "traditional two-phase call sites remain valid without changes." is not correct here, since we also drop the first two arguments so there should not be an ambiguity with two-phase calls.
```suggestion
- It is defaulted and appears as the last argument.
```

docs/cub/api_docs/device_wide.rst
```diff
@@ -11,6 +11,72 @@ Device-Wide Primitives
...
+- Streams can be forwarded by specifying ``cuda::get_stream`` properties.
+- You can select the memory resource (CCCL-provided or custom) used for internal allocations.
+- Supported algorithms accept determinism requirements (for example, ``cuda::execution::determinism::gpu_to_gpu``).
+- Multiple properties compose into a single centralized control point.
```
bernhardmgruber: Maybe:
```suggestion
- Multiple properties compose into a single centralized argument.
```
Since it's not clear to me what a "control point" is.

docs/cub/api_docs/device_wide.rst
```diff
@@ -11,6 +11,72 @@ Device-Wide Primitives
...
+
+Example (centralized control via a single environment argument):
+
+.. code-block:: c++
+
+   #include <cub/cub.cuh>
```
bernhardmgruber: Suggestion: `cub.cuh` is expensive to compile, so maybe show users that they can and should use smaller headers:
```suggestion
   #include <cub/device/device_reduce.cuh>
```

docs/cub/api_docs/device_wide.rst
```diff
@@ -11,6 +11,72 @@ Device-Wide Primitives
...
+   auto mr_env = cuda::std::execution::prop{cuda::mr::__get_memory_resource_t{}, mr};
+
+   auto det_env = cuda::execution::require(cuda::execution::determinism::gpu_to_gpu);
+
+   auto env = cuda::std::execution::env{stream_env, mr_env, det_env};
+
+   // Single-phase API (no explicit temp storage, environment last and defaulted)
+   cub::DeviceReduce::Reduce(d_in, d_out, num_items, cuda::std::plus<>{}, init, env);
+
+This page focuses on the traditional two-phase pattern; see individual algorithm documentation for the
```
bernhardmgruber: Suggestion:
```suggestion
The remainder of this page focuses on the traditional two-phase pattern; see individual algorithm documentation for the
```

docs/cub/api_docs/device_wide.rst
Aminsed: Done [c0f595b](https://github.com/NVIDIA/cccl/pull/6208/commits/c0f595bac1bb9ea5ada34c6bf1bf23eeb2ef3dcd)

bernhardmgruber:
@gonidelis please re-review and unblock

Aminsed:
Heads-up: the only red check here is the optional MatX build. I tracked down the root cause to MatX's argmin/argmax comparators not accepting the tuple/pair accumulator mix that current CUB emits and opened NVIDIA/MatX#1095 to describe it. I also sent NVIDIA/MatX#1096 to normalize those comparators; once that lands, the optional workflow should go green again.

In the meantime, could whoever is holding the lock on this PR please re-review the docs patch? All mandatory CCCL jobs are green.

bernhardmgruber:
@gonidelis please re-review and unblock

gonidelis (approved):

PR #6225, opened 2025-10-14
[on hold] Drop experimental TMA exposure in cuda::barrier

bernhardmgruber:
This is all exposed by `cuda::ptx`. It was originally introduced in #379

This is a breaking change so we wait for CCCL 4.0.


libcudacxx/include/cuda/barrier
```diff
@@ -38,230 +38,4 @@
...
-// Forward-declare CUtensorMap for use in cp_async_bulk_tensor_* PTX wrapping
-// functions. These functions take a pointer to CUtensorMap, so do not need to
-// know its size. This type is defined in cuda.h (driver API) as:
-//
-//     typedef struct CUtensorMap_st {  [ .. snip .. ] } CUtensorMap;
-//
-// We need to forward-declare both CUtensorMap_st (the struct) and CUtensorMap
-// (the typedef):
-struct CUtensorMap_st;
-typedef struct CUtensorMap_st CUtensorMap;
```
bernhardmgruber, line 50: This is a borderline breaking change. In principle, we require users to include the headers for what they should use, so I think they need to ensure they have the right header to use `CUtensorMap`. But I am also fine leaving the forward declaration in.

libcudacxx/include/cuda/barrier
```diff
@@ -38,230 +38,4 @@
...
-               "Source must be global memory address.");
-
-  ::cuda::ptx::cp_async_bulk(
-    ::cuda::ptx::space_cluster,
-    ::cuda::ptx::space_global,
-    __dest,
-    __src,
-    __size,
-    ::cuda::device::barrier_native_handle(__bar));
-}
```
bernhardmgruber: I find it a bit sad to leave some of the assertions behind here. But those functions were not used anyway in our `barrier` and `memcpy_async` implementation etc. I think we should consider adding some assertions to the PTX exposure in cuda::ptx. @ahendriksen do you think we can add such assertions there?

bernhardmgruber:
@ahendriksen I would like your review as well, since I think you added this functionality.

miscco (approved):

libcudacxx/include/cuda/barrier
fbusato: Adding assertions in the code generator could be very hard because they are specific for each instruction. For example, I wrote the code for `warp_shuffle` by hand for this reason

fbusato (approved):

bernhardmgruber:
Switching to draft to prevent accidential merging. Waiting for approval from @ahendriksen.

ahendriksen:
This looks great. But it is a breaking change. These APIs where used in the programming guide until recently (or are still up in the 13.0 release?). 

Can we mark these APIs as deprecated and remove them in the CTK 14 release?

bernhardmgruber:
> Can we mark these APIs as deprecated and remove them in the CTK 14 release?

Yes. Deprecation here:  #6305

Issue #6172, opened 2025-10-09
[FEA]: Have utility function to take a DLPack and emit a TMA descritor


nouiz:
### Is this a duplicate?

- [ ] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cccl/issues) for this request and that I agree to the [Code of Conduct](CODE_OF_CONDUCT.md)

### Area

General CCCL

### Is your feature request related to a problem? Please describe.

Using TMA should help kernels go faster.
But it is harder to use then normal load/store operation.

### Describe the solution you'd like

What about providing utilities to help?
For example, a function like: `CreateTMADescriptor(DLPack* tensor, void**tma_descriptor)`
Can be used like this:
```
CUtensorMap tma_map;
DLPack* dltensor;
CreateTMADescriptor(dltensor, &tma_map);
```

This is when we create the TMA descriptor from the CPU.
DLPack is interesting as many frameworks support it. It is a de-facto standard n-dim tensor descriptor.

nouiz:
Note, there is related issues/epic like:
https://github.com/NVIDIA/cccl/issues/58
https://github.com/NVIDIA/cccl/issues/39
https://github.com/NVIDIA/cccl/issues/38
https://github.com/NVIDIA/cccl/issues/36

This one isn't purely a duplicate as it include a DLPack interface.
But when the above issues are done, this one would be less work.

jrhemstad:
Thanks @nouiz I think this is absolutely something we can and should provide in CCCL.

There'd already been some discussions around building some DLPack <-> mdspan interop utilities in CCCL, and I think that a utility to go to DLPack -> TMA will be very valuable to have alongside things like mdspan -> TMA.

I'm envisioning what this could look like is...
- We'd add a new `cuda::tma_descriptor` owning type for constructing/cleaning up a TMA descriptor 
- We'd add factories:
   - `cuda::tma_descriptor cuda::make_tma_descriptor(DLTensor* tensor);` 
   -  `cuda::tma_descriptor cuda::make_tma_descriptor(DLManagedTensor* tensor);` 

The lifetime semantics will need more thinking.

[...]

PR #6237, opened 2025-10-14
Provide `make_tma_descriptor`, `DLPack` ->  `CUtensorMap`

fbusato:
Address https://github.com/NVIDIA/cccl/issues/6172

## Description

Provide `make_tma_descriptor` to convert `DLTensor` and `DLManagedTensor` to `CUtensorMap`.

Aspects discussed with the JAX team:
- Interface: current APIs are fine.
- Error handling: Exceptions should be avoided.

Aspects to discuss:
- `CUtensorMap` opaque type is enough, or do we need a higher-level object type.

Todo:
- [x] extend unit test
- [x] documentation


docs/libcudacxx/extended_api/tma/make_tma_descriptor.rst
```diff
@@ -0,0 +1,191 @@
...
+        bytes128_atom_32B,
+        bytes128_atom_32B_flip_8B,
+        bytes128_atom_64B
+    };
+
+    } // namespace cuda
+
+The functions construct a `CUDA Tensor Memory Accelerator (TMA) descriptor <https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#using-tma-to-transfer-multi-dimensional-arrays>`__ from a ``DLTensor`` or ``DLManagedTensor`` `descriptors <https://dmlc.github.io/dlpack/latest/c_api.html>`__. The resulting ``CUtensorMap`` can be bound to TMA-based copy instructions to efficiently stage multi-dimensional tiles in shared memory on Compute Capability 9.0 and newer GPUs.
+
+The API is available when ``dlpack/dlpack.h`` (DLPack v1) is discoverable at compile time.
```
bernhardmgruber:
```suggestion
The API is available when ``dlpack/dlpack.h`` (DLPack v1) is discovered at compile time.
```

libcudacxx/include/cuda/__tma/make_tma_descriptor.h
```diff
@@ -0,0 +1,499 @@
...
+
+#if defined(_CCCL_IMPLICIT_SYSTEM_HEADER_GCC)
+#  pragma GCC system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_CLANG)
+#  pragma clang system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_MSVC)
+#  pragma system_header
+#endif // no system header
+
+#if _CCCL_HAS_INCLUDE("dlpack/dlpack.h")
```
bernhardmgruber: Why not:
```suggestion
#if _CCCL_HAS_INCLUDE(<dlpack/dlpack.h>)
```
?

libcudacxx/include/cuda/__tma/make_tma_descriptor.h
```diff
@@ -0,0 +1,499 @@
...
+
+_CCCL_BEGIN_NAMESPACE_CUDA
+
+static_assert(DLPACK_MAJOR_VERSION == 1, "DLPACK_MAJOR_VERSION must be 1");
+
+/***********************************************************************************************************************
+ * Public Enums
+ ***********************************************************************************************************************/
+
+enum class TmaOOBfill
```
bernhardmgruber: Important: IIRC, `snake_case` also applies to enumerations in our coding conventions. Please update all such identifiers.

libcudacxx/include/cuda/__tma/make_tma_descriptor.h
fbusato: I don't know if there is a difference for `__has_include`

miscco: Actually there is because it uses the different search path

libcudacxx/test/libcudacxx/CMakeLists.txt
```diff
@@ -160,6 +165,7 @@ set_tests_properties(libcudacxx.test.lit PROPERTIES
 # Add c2h tests:
 cccl_get_c2h()
 
+
```
alliepiper: nit: extra whitespace
```suggestion
```

libcudacxx/test/libcudacxx/CMakeLists.txt
```diff
@@ -43,6 +47,7 @@ endif()
 
 # enable exceptions and assertions in tests
 string(APPEND LIBCUDACXX_TEST_COMPILER_FLAGS " -DCCCL_ENABLE_ASSERTIONS")
+string(APPEND LIBCUDACXX_FORCE_INCLUDE " -I ${dlpack_SOURCE_DIR}/include")
```
alliepiper: Minor: The force include is for force including specific headers, not include paths. This would make more sense with the compiler flags.
```suggestion
string(APPEND LIBCUDACXX_COMPILER_FLAGS " -I ${dlpack_SOURCE_DIR}/include")
```

libcudacxx/test/libcudacxx/CMakeLists.txt
```diff
@@ -43,6 +47,7 @@ endif()
 
 # enable exceptions and assertions in tests
 string(APPEND LIBCUDACXX_TEST_COMPILER_FLAGS " -DCCCL_ENABLE_ASSERTIONS")
+string(APPEND LIBCUDACXX_COMPILER_FLAGS " -I ${dlpack_SOURCE_DIR}/include")
```
alliepiper: D'oh, just realized I made a typo in my earlier suggested fix. It should be:


```suggestion
string(APPEND LIBCUDACXX_TEST_COMPILER_FLAGS " -I ${dlpack_SOURCE_DIR}/include")
```

Curious that CI is currently working without this being added to the lit flags...

alliepiper: > Curious that CI is currently working without this being added to the lit flags...

Ahh nvm, I see the fallback path now 👍🏻

libcudacxx/test/libcudacxx/cuda/tma/make_tma_descriptor.pass.cpp
```diff
@@ -0,0 +1,330 @@
...
+  tensor.strides     = const_cast<int64_t*>(strides_storage);
+  tensor.byte_offset = 0;
+  unused(cuda::make_tma_descriptor(tensor, box_sizes, elem_strides, no_interleave));
+  unused(cuda::make_tma_descriptor(tensor, box_sizes, elem_strides, cuda::tma_interleave_layout::bytes16));
+  return true;
+}
+
+__host__ bool test_enums()
+{
+  constexpr cuda::tma_oob_fill tma_oob_fill_array[]           = {cuda::tma_oob_fill::none, cuda::tma_oob_fill::nan};
```
oleksandr-pavlyk, line 190: NIT: perhaps `static constexpr`?

oleksandr-pavlyk:
@fbusato I could not find a test covering making TMA descriptor with `tma_oob_fill:nan`.

docs/libcudacxx/extended_api/tma/make_tma_descriptor.rst
```diff
@@ -0,0 +1,229 @@
...
+
+**Function signatures**
+
+.. code:: cuda
+
+    namespace cuda {
+
+    template <size_t BoxDimSize, size_t ElemStrideSize>
+    [[nodiscard]] inline CUtensorMap
+    make_tma_descriptor(
```
oleksandr-pavlyk: Should these signatures explicitly state that they are `__host__` API only?

libcudacxx/include/cuda/__tma/make_tma_descriptor.h
```diff
@@ -0,0 +1,575 @@
...
+
+/***********************************************************************************************************************
+ * Public Enums
+ ***********************************************************************************************************************/
+
+enum class tma_oob_fill
+{
+  none,
+  nan
+};
```
bernhardmgruber, line 53: Q: Should we assign the enumerators to the same values as the driver macros?
```suggestion
enum class tma_oob_fill
{
  none = ::CU_TENSOR_MAP_FLOAT_OOB_FILL_NONE,
  nan = ::CU_TENSOR_MAP_FLOAT_OOB_FILL_NAN_REQUEST_ZERO_FMA
};
```
?

libcudacxx/test/libcudacxx/cuda/tma/make_tma_descriptor.pass.cpp
fbusato: no really. It has compile-time indexing and GCC generates better code for small array

docs/libcudacxx/extended_api/tma/make_tma_descriptor.rst
fbusato: yes, makes sense

libcudacxx/include/cuda/__tma/make_tma_descriptor.h
fbusato: that's a good point. Hope I'm not missing any implication about this dependency

docs/libcudacxx/extended_api/tma/make_tma_descriptor.rst
```diff
@@ -0,0 +1,229 @@
...
+Defined in the ``<cuda/tma>`` header.
+
+**Function signatures**
+
+.. code:: cuda
+
+    namespace cuda {
+
+    template <size_t BoxDimSize, size_t ElemStrideSize>
+    [[nodiscard]] inline __host__
```
miscco: We should not have naked `__host__` as that has no real meaning

Also the inline is more or less noise as that is a function template

libcudacxx/include/cuda/__driver/driver_api.h
```diff
@@ -823,6 +824,37 @@ __graphKernelNodeSetAttribute(::CUgraphNode __node, ::CUkernelNodeAttrID __id, c
 }
 #  endif // _CCCL_CTK_AT_LEAST(13, 0)
 
+[[nodiscard]] _CCCL_HOST_API inline ::cuda::std::pair<::CUtensorMap, ::cudaError_t> __tensorMapEncodeTiledNoThrow(
```
miscco: I know its a bit more complex, but should this be `cuda::std::expected<::CUtensorMap, ::cudaError_t> `

libcudacxx/include/cuda/__tma/make_tma_descriptor.h
```diff
@@ -0,0 +1,575 @@
...
+
+#if defined(_CCCL_IMPLICIT_SYSTEM_HEADER_GCC)
+#  pragma GCC system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_CLANG)
+#  pragma clang system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_MSVC)
+#  pragma system_header
+#endif // no system header
+
+#if _CCCL_HAS_INCLUDE(<dlpack/dlpack.h>)
```
miscco: Should this also be guarded by `_CCCL_HAS_CTK() && !_CCCL_COMPILER(NVRTC)`

libcudacxx/include/cuda/__tma/make_tma_descriptor.h
```diff
@@ -0,0 +1,575 @@
...
+  switch (__oobfill)
+  {
+    case tma_oob_fill::none:
+      return ::CU_TENSOR_MAP_FLOAT_OOB_FILL_NONE;
+    case tma_oob_fill::nan:
+      return ::CU_TENSOR_MAP_FLOAT_OOB_FILL_NAN_REQUEST_ZERO_FMA;
+    default:
+      _CCCL_UNREACHABLE();
+  }
+  _CCCL_UNREACHABLE();
```
miscco: I believe this should not be necessary or desirable. Modern compiler warn if not all cases are handled and with the default we might miss some

libcudacxx/include/cuda/__tma/make_tma_descriptor.h
```diff
@@ -0,0 +1,575 @@
...
+/***********************************************************************************************************************
+ * DLTensor Internal API
+ ***********************************************************************************************************************/
+
+_CCCL_HOST_API inline void __check_device(const ::DLTensor& __tensor, tma_swizzle __swizzle) noexcept
+{
+  _CCCL_VERIFY(__tensor.device.device_type == ::kDLCUDA || __tensor.device.device_type == ::kDLCUDAManaged,
+               "Device type must be kDLCUDA or kDLCUDAManaged");
+  _CCCL_VERIFY(__tensor.device.device_id >= 0, "Device ID must be a valid GPU device ordinal");
+  auto __current_device = ::cuda::__driver::__deviceGet(static_cast<int>(__tensor.device.device_id));
```
miscco: Should we use explicit `cuda::device_ref`?

libcudacxx/include/cuda/std/__memory/is_sufficiently_aligned.h
```diff
@@ -25,6 +25,7 @@
 #include <cuda/__cmath/pow2.h>
 #include <cuda/std/cstddef> // size_t
 #include <cuda/std/cstdint> // uintptr_t
+#include <cuda/std/type_traits> // is_void_v
```
miscco: Critical: We only include the required subheaders

libcudacxx/test/libcudacxx/cuda/tma/make_tma_descriptor.pass.cpp
```diff
@@ -0,0 +1,330 @@
...
+#include <cuda/std/cstdint>
+#include <cuda/std/span>
+
+#include <cuda_runtime_api.h>
+
+#include "test_macros.h"
+
+constexpr auto no_interleave = cuda::tma_interleave_layout::none;
+
+__host__ bool test_ranks()
```
miscco: Ditto: Drop unnecessary `__host__`

miscco (changes requested):
Looks good, mostly nitpicks

libcudacxx/include/cuda/__tma/make_tma_descriptor.h
fbusato: great suggestion

libcudacxx/include/cuda/__tma/make_tma_descriptor.h
fbusato: I'm always confused which compiler supports one case or the other

libcudacxx/include/cuda/__device/device_ref.h
```diff
@@ -42,7 +43,10 @@ class device_ref
   //! @brief Create a `device_ref` object from a native device ordinal.
   /*implicit*/ _CCCL_HOST_API constexpr device_ref(int __id) noexcept
       : __id_(__id)
-  {}
+  {
+    const auto __device_count = ::cuda::__driver::__deviceGetCount();
+    _CCCL_VERIFY(__id >= 0 && __id < __device_count, "Device ID must be a valid GPU device ordinal");
```
fbusato: @davebayer can we add this check in `device_ref`?

docs/libcudacxx/extended_api/tma/make_tma_descriptor.rst
```diff
@@ -0,0 +1,229 @@
...
+        tensor.dtype.code  = static_cast<uint8_t>(kDLFloat);
+        tensor.dtype.bits  = 32;
+        tensor.dtype.lanes = 1;
+        tensor.shape       = const_cast<int64_t*>(shape_storage);
+        tensor.strides     = const_cast<int64_t*>(strides_storage);
+        tensor.byte_offset = 0;
+
+        // Define shared memory box sizes and element strides.
+        int box_sizes_storage[2]    = {16, 16};
+        int elem_strides_storage[2] = {1, 1};
```
oleksandr-pavlyk: I am a little perplexed by `{1, 1}` strides for the tile in shared memory. Does it mean that shared memory is assumed to be array of arrays `smem[DimX][DimY]` and `{1, 1}` means we increment each index by 1?

fbusato: this is a special case indeed

> When all elements of elementStrides array is one, boxDim specifies the number of elements to load.

https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__TENSOR__MEMORY.html#group__CUDA__TENSOR__MEMORY_1ga7c7d2aaac9e49294304e755e6f341d7

not sure if this should be in the example or not

oleksandr-pavlyk: What if my shared memory is allocated as 1D array `smem[DimX * DimY]` would I them specify `box_sizes = {DimX, DimY}` and `elem_strides = {DimY, 1}` ?

docs/libcudacxx/extended_api/tma/make_tma_descriptor.rst
```diff
@@ -0,0 +1,229 @@
+.. _libcudacxx-extended-api-tma-make_tma_descriptor:
+
+``cuda::make_tma_descriptor``
+=============================
+
+Defined in the ``<cuda/tma>`` header.
```
oleksandr-pavlyk, line 6: The header does not exist at the moment. Just making sure it is not an oversight

docs/libcudacxx/extended_api/tma/make_tma_descriptor.rst
fbusato: yes, probably better to document your case and add a comment for `{1, 1}`

docs/libcudacxx/extended_api/tma/make_tma_descriptor.rst
fbusato: thanks! this would have been a big mistake

docs/libcudacxx/extended_api/tma/make_tma_descriptor.rst
```diff
@@ -0,0 +1,231 @@
...
+
+        // Define shared memory box sizes and element strides.
+        constexpr int BoxSizeX      = 16; // rows
+        constexpr int BoxSizeY      = 16; // columns
+        int box_sizes_storage[2]    = {BoxSizeX, BoxSizeY};
+        int elem_strides_storage[2] = {BoxSizeY, 1}; // {1, ..., 1} is also valid to specify contiguous memory
+
+        return cuda::make_tma_descriptor(
+            tensor,
+            cuda::std::span{box_sizes_storage},
```
oleksandr-pavlyk: This code fails to compile since deduced span typename is `int`, not `const int` as expected by `make_tma_descriptor` signature

fbusato: ahhh, that's painful. The function uses a template parameter for the size that breaks automatic conversion.
Let me see I can use dynamic extent everywhere

oleksandr-pavlyk:
I was able to use `CUtensorMap` produced by `make_tma_descriptor` in a kernel that transposes an array using `cuda::ptx::cp_async_bulk_tensor` to load/store tiles.

fbusato:
> I was able to use CUtensorMap produced by make_tma_descriptor in a kernel that transposes an array using cuda::ptx::cp_async_bulk_tensor to load/store tiles.

Amazing 🎉

libcudacxx/include/cuda/__driver/driver_api.h
```diff
@@ -836,6 +837,41 @@ __graphKernelNodeSetAttribute(::CUgraphNode __node, ::CUkernelNodeAttrID __id, c
 }
 #  endif // _CCCL_CTK_AT_LEAST(13, 0)
 
+[[nodiscard]] _CCCL_HOST_API inline ::cuda::std::expected<::CUtensorMap, ::cudaError_t> __tensorMapEncodeTiledNoThrow(
```
fbusato: @miscco @davebayer @pciolkosz any thought about using `expected` in `driver_api.h`?

pciolkosz: I am neutral, I don't think I personally need it for the `NoThrow` interfaces, but I have a hard time estimating how that would impact compile times and if I care about it

davebayer: I would prefer to just make it a very lightweight wrapper over the called function. What benefits does the `expected` bring here? Instead of:
```cpp
int value;
const auto status = ::cuda::__driver::__meow(&value, CU_MEOW_PARAM);
```
you get
```cpp
const auto result = ::cuda::__driver::__meow(&value, CU_MEOW_PARAM);
```
which you still need to check for errors. I don't think it brings anything too valuable to the table.

@miscco will probably like this due to not having in/out arguments which I don't see as a problem since these are internal only apis.

libcudacxx/include/cuda/__device/device_ref.h
davebayer: I don't know. Maybe if it was a `_CCCL_ASSERT`? @pciolkosz what do you think?

oleksandr-pavlyk:
My original example created DLTensor for square matrix with C contiguous layout, data type `float`, and shape `(256, 256)`.

When passing `DLTensor` with shape `{57, 57}`, strides `{57, 1}` (C-contiguous layout), the assertion ``make_tma_descriptor.h:398: __get_tensor_strides: Assertion `Stride in bytes is not a multiple of the alignment (32 or 16)' failed.`` is triggered.

Replacing 57 with 58 also trips it, but 56 works:

```
>>> [ (n, (n * 4) % 16) for n in [55, 56, 57, 58] ]
[(55, 12), (56, 0), (57, 4), (58, 8)]
```

Passing `tma_oob_fill::nan` does not help.

All tests appear to use shape values that are a whole multiple of itemsize. 

While this may be what is expected by the driver, I feel we need to provide an example/suggestion of how to handle layouts were the stride in GMEM is not a multiple of 16 bytes.

oleksandr-pavlyk:
> When passing `DLTensor` with shape `{57, 57}`, strides `{57, 1}` (C-contiguous layout), the assertion `` make_tma_descriptor.h:398: __get_tensor_strides: Assertion `Stride in bytes is not a multiple of the alignment (32 or 16)' failed. `` is triggered.
> 

The `__CCCL_VERIFY` should use matching quotation symbols. Note how the output above uses \` and \' instead of \' and \'.

fbusato:
> The __CCCL_VERIFY should use matching quotation symbols. Note how the output above uses ` and ' instead of ' and '.

`_CCCL_VERIFY` relies on `__assert_fail` so it is no a related to CCCL

fbusato:
> My original example created DLTensor for square matrix with C contiguous layout, data type float, and shape (256, 256).
>
> When passing DLTensor with shape {57, 57}, strides {57, 1} (C-contiguous layout), the assertion make_tma_descriptor.h:398: __get_tensor_strides: Assertion `Stride in bytes is not a multiple of the alignment (32 or 16)' failed. is triggered.
>
> Replacing 57 with 58 also trips it, but 56 works:
>
> [ (n, (n * 4) % 16) for n in [55, 56, 57, 58] ]
> [(55, 12), (56, 0), (57, 4), (58, 8)]
>Passing tma_oob_fill::nan does not help.

The behavior seems correct. Note the message `Stride in BYTES is not a multiple of the alignment ... `: `56 * sizeof(int) % 16 == 0`.

> All tests appear to use shape values that are a whole multiple of itemsize.
>
> While this may be what is expected by the driver, I feel we need to provide an example/suggestion of how to handle layouts were the stride in GMEM is not a multiple of 16 bytes.

This is not possible. 16 bytes is a constraint of the HW

libcudacxx/include/cuda/__device/device_ref.h
pciolkosz: Definitely not as a driver call for device count. I could see it being an assert comparing it with `devices.size()`

But that might run into a circular dependency with the type that would need to be resolved somehow

docs/libcudacxx/extended_api/tma.rst
```diff
@@ -0,0 +1,24 @@
+.. _libcudacxx-extended-api-tma:
+
+Tensor Memory Accelerator (TMA)
+===============================
+
```
bernhardmgruber, line 5: Meta question: I wondered whether I should also add a section on TMA exposure to the docs, where we would point out the various APIs we have for targetting TMA. That would include `cuda::memcpy_async`, `cuda::device::memcpy_async_tx`, `cub::BlockLoadToShared`, etc. Do you think this content should be part of this page?

fbusato: That's a great idea, actually. Would be nice to have a common place for all utilities related to TMA.
Let think about it after this PR

bernhardmgruber:
@miscco this PR is blocked by you. Can you please re-review and unblock? Thx!

libcudacxx/include/cuda/__tma/make_tma_descriptor.h
```diff
@@ -0,0 +1,574 @@
...
+_CCCL_HOST_API inline void __check_device(const ::DLTensor& __tensor, tma_swizzle __swizzle) noexcept
+{
+  _CCCL_VERIFY(__tensor.device.device_type == ::kDLCUDA || __tensor.device.device_type == ::kDLCUDAManaged,
+               "Device type must be kDLCUDA or kDLCUDAManaged");
+  auto __current_device     = ::cuda::device_ref(__tensor.device.device_id);
+  auto __compute_capability = __current_device.attribute<::cudaDevAttrComputeCapabilityMajor>();
+  // auto __current_device = ::cuda::__driver::__deviceGet(__tensor.device.device_id);
+  // auto __compute_capability =
+  //   ::cuda::__driver::__deviceGetAttribute(::CU_DEVICE_ATTRIBUTE_COMPUTE_CAPABILITY_MAJOR, __current_device);
+  //_CCCL_VERIFY(__compute_capability >= 9, "Compute capability 9.0 or higher is required");
```
miscco, line 205: Please drop dead code

libcudacxx/include/cuda/__tma/make_tma_descriptor.h
```diff
@@ -0,0 +1,574 @@
...
+  if (__compute_capability >= 10)
```
miscco, line 206: previously `__compute_capability` was 90. now this is 10 are we missing a zero or did something change

miscco (approved):

libcudacxx/include/cuda/__device/physical_device.h
```diff
@@ -139,9 +139,15 @@ __make_physical_devices(::cuda::std::size_t __device_count)
   return __devices;
 }
 
-[[nodiscard]] inline ::cuda::std::span<__physical_device> __physical_devices()
+[[nodiscard]] inline ::cuda::std::size_t __physical_devices_count()
 {
   static const auto __device_count = static_cast<::cuda::std::size_t>(::cuda::__driver::__deviceGetCount());
+  return __device_count;
```
davebayer, line 145: Why was this function necessary? You could just use `cuda::__driver::__deviceGetCount()` directly

libcudacxx/include/cuda/__tma/make_tma_descriptor.h
```diff
@@ -0,0 +1,574 @@
...
+ **********************************************************************************************************************/
+
+[[nodiscard]] _CCCL_HOST_API inline ::CUtensorMap make_tma_descriptor(
+  const ::DLTensor& __tensor,
+  ::cuda::std::span<const int> __box_sizes,
+  ::cuda::std::span<const int> __elem_strides,
+  tma_interleave_layout __interleave_layout = tma_interleave_layout::none,
+  tma_swizzle __swizzle                     = tma_swizzle::none,
+  tma_l2_fetch_size __l2_fetch_size         = tma_l2_fetch_size::none,
+  tma_oob_fill __oobfill                    = tma_oob_fill::none) noexcept
```
davebayer, line 516: Q: I don't know anything about tma, but if I want to specify only l2 fetch size, then I have to type `make_tma_descriptor(..., tma_interleave_layout::none, tma_swizzle::none, tma_l2_fetch_size::some_size)`?

If yes, my question is whether it wouldn't be better to implement this as a separate `struct tma_descriptor_params` (or some other name) and in C++20 you could do: `make_tma_descriptor(..., {.l2_fetch_size = tma_l2_fetch_size::some_size})`

bernhardmgruber: That looks cool!

davebayer: Actually, with all compilers (except for msvc) it works even in C++17, see https://godbolt.org/z/jh3vT4fc4

fbusato: this is a very nice idea. Unfortunate, C++20 is not really adopted in the API context, so the modification would make the API less user-friendly. We can think about it in the future.

libcudacxx/include/cuda/__device/physical_device.h
fbusato: IIRC @pciolkosz suggested to not directly use the function directly to allow caching.

libcudacxx/include/cuda/__tma/make_tma_descriptor.h
fbusato: it was automatically merged. Let me open a new PR

libcudacxx/include/cuda/__tma/make_tma_descriptor.h
davebayer: But as I said, nvcc has no problem with the code, nor do any of the other compilers with exception of msvc. So most of the time, you can just use it in C++17 as well. And if not, you just have to fallback to:
```cpp
make_tma_descriptor(..., {tma_interleave_layout::none, tma_swizzle::none, tma_l2_fetch_size::some_size});
```
which is basically the same thing.

Please, reconsider this

PR #6244, opened 2025-10-15
Add host standard library detection

davebayer:
@ericniebler implemented some host standard library detection for using intrinsics for some simple functions such as `move` or `forward`.

I've improved the design a bit further, moved it to a separate header and added host standard library namespaces. The idea is that we could forward declare things like `std::string` instead of including the whole header to reduce compilation times.

The user would be required to include those headers if he wants to use APIs including these types.

Btw. `_GLIBCXX_VERSION` doesn't exist, I've changed it to `__GLIBCXX__`.


libcudacxx/include/cuda/std/__cccl/builtin.h
```diff
@@ -609,25 +610,18 @@
...
 // Bring in the bits of the STL we need
-#  if defined(_GLIBCXX_VERSION)
+#  if _CCCL_HOST_STD_LIB(LIBSTDCXX)
 #    include <bits/move.h> // for move, forward, forward_like, and addressof
-#  elif defined(_LIBCPP_VERSION)
+#  elif _CCCL_HOST_STD_LIB(LIBCXX)
 #    include <__memory/addressof.h>
 #    include <__utility/as_const.h>
 #    include <__utility/forward.h>
 #    include <__utility/forward_like.h>
```
miscco: As discussed internally, we need to guard this on availability of `__cpp_lib_forward_like`

libcudacxx/include/cuda/std/__internal/namespaces.h
```diff
@@ -104,6 +104,18 @@
 #  endif // not defined(_NVHPC_CUDA)
 #endif // not defined(CCCL_DISABLE_ARCH_DEPENDENT_NAMESPACE)
 
+// Host standard library namespaces
+#if _CCCL_HOST_STD_LIB(LIBSTDCXX)
```
miscco, line 108: As discussed internally, we need to respect `_GLIBCXX_INLINE_VERSION ` to verify that there is an inline namespace

libcudacxx/include/cuda/std/__cccl/builtin.h
```diff
@@ -608,54 +609,47 @@
...
+// #    endif
+
+// // std::forward_like builtin
+// // Leaving out MSVC for now because it is hard for forward-declare std::forward_like.
+// #    if (_CCCL_COMPILER(CLANG, >=, 17) || _CCCL_COMPILER(GCC, >=, 15)) && defined(__cpp_lib_forward_like) \
+//       && (__cpp_lib_forward_like >= 202217L)
+// #      define _CCCL_HAS_BUILTIN_STD_FORWARD_LIKE() 1
+// #    endif
+// #  endif // defined(_GLIBCXX_VERSION) || defined(_LIBCPP_VERSION) || defined(_MSVC_STL_VERSION)
+// #endif // defined(__cplusplus)
```
miscco, line 655: That looks accidental

libcudacxx/include/cuda/std/__internal/namespaces.h
```diff
@@ -104,6 +104,27 @@
...
+  // We apply the visibility attribute on each type individually, so the _GLIBCXX_VISIBILITY(default) is not used here
```
miscco: This is something I really want to change

libcudacxx/include/cuda/std/__cccl/builtin.h
davebayer: I'll add it back later

libcudacxx/include/cuda/std/__complex/nvbf16.h
```diff
@@ -295,21 +296,21 @@ struct __get_complex_impl<__nv_bfloat16>
 };
 
 #  if !_CCCL_COMPILER(NVRTC)
-template <class _CharT, class _Traits>
+template <class _CharT, class _Traits, class _Tp = float>
 ::std::basic_istream<_CharT, _Traits>&
 operator>>(::std::basic_istream<_CharT, _Traits>& __is, complex<__nv_bfloat16>& __x)
 {
-  ::std::complex<float> __temp;
+  ::std::complex<_Tp> __temp;
```
davebayer: This is necessary to prevent errors about the type not being defined, but only forward declared

libcudacxx/include/cuda/std/__complex/nvfp16.h
```diff
@@ -158,7 +158,8 @@ class _CCCL_TYPE_VISIBILITY_DEFAULT _CCCL_ALIGNAS(alignof(__half2)) complex<__ha
     return *this;
   }
 
-  _CCCL_HOST operator ::std::complex<value_type>() const
+  template <class _Up = value_type>
+  _CCCL_HOST operator ::std::complex<_Up>() const
```
davebayer: Again..

libcudacxx/include/cuda/std/string_view
```diff
@@ -855,9 +855,10 @@ _CCCL_HOST_API ::std::basic_ostream<_CharT>&
 operator<<(::std::basic_ostream<_CharT>& __os, basic_string_view<_CharT> __str)
 {
 #  if __cpp_lib_string_view >= 201606L
-  return __os << ::std::basic_string_view<_CharT>{__str};
+  return __os << ::std::basic_string_view<_CharT, ::std::char_traits<_CharT>>{__str};
```
davebayer, line 862: Because we only forward declare the types, we don't have access to default template parameters, so we have to pass all of the template paramters here.

libcudacxx/include/cuda/std/__complex/nvbf16.h
```diff
@@ -295,21 +296,21 @@ struct __get_complex_impl<__nv_bfloat16>
...
 }
 
-template <class _CharT, class _Traits>
+template <class _CharT, class _Traits, class _Tp = float>
 ::std::basic_ostream<_CharT, _Traits>&
 operator<<(::std::basic_ostream<_CharT, _Traits>& __os, const complex<__nv_bfloat16>& __x)
 {
-  return __os << complex<float>{__x};
+  return __os << complex<_Tp>{__x};
 }
```
miscco: This one should not be necessary because we are converting to a `cuda::std::complex` which is defined

libcudacxx/include/cuda/std/__complex/nvbf16.h
miscco: I believe we could get away with the extended floating point types always casting to `::cuda::std::complex<float>` and only do the smart thing for that

libcudacxx/include/cuda/std/__cccl/builtin.h
```diff
@@ -607,55 +608,51 @@
...
-
-// std::forward_like builtin
-// Leaving out MSVC for now because it is hard for forward-declare std::forward_like.
-#    if (_CCCL_COMPILER(CLANG, >=, 17) || _CCCL_COMPILER(GCC, >=, 15)) && defined(__cpp_lib_forward_like) \
-      && (__cpp_lib_forward_like >= 202217L)
-#      define _CCCL_HAS_BUILTIN_STD_FORWARD_LIKE() 1
-#    endif
-#  endif // defined(_GLIBCXX_VERSION) || defined(_LIBCPP_VERSION) || defined(_MSVC_STL_VERSION)
-#endif // defined(__cplusplus)
+// todo: re-enable std builtins
```
davebayer, line 611: They were not working anyway, so I'd like to leave this for a different PR

libcudacxx/include/cuda/std/__complex/nvfp16.h
```diff
@@ -294,7 +295,7 @@ struct __get_complex_impl<__half>
   }
 };
 
-#  if !defined(_LIBCUDACXX_HAS_NO_LOCALIZATION) && !_CCCL_COMPILER(NVRTC)
```
davebayer, line 297: Is never defined

libcudacxx/include/cuda/std/__cccl/builtin.h
miscco: I would love to know what is not woorking. Generally I would just replace the host STL detection and leave the rest in

PR #6250, opened 2025-10-15
Replace inline PTX by cuda::ptx in cuda::barrier<thread_scope_block>

bernhardmgruber:
Due to lack of a good example for a SASS test, I used this simple example:
```c++
#include <cuda/barrier>
#include <cuda/ptx>

// selects a single leader thread from the block
__device__ bool elect_one() {
  // elect_sync is important to help the optimizer generate a uniform datapath
  return cuda::ptx::elect_sync(~0) && threadIdx.x < 32;
}

__global__ void example_kernel(int* gmem1, double* gmem2) {
  constexpr int tile_size = 1024;
  __shared__ alignas(16)    int smem1[tile_size];
  __shared__ alignas(16) double smem2[tile_size];
  #pragma nv_diag_suppress static_var_with_dynamic_init
  using barrier_t = cuda::barrier<cuda::thread_scope_block>;
  __shared__  barrier_t bar;
  // setup the barrier where only the leader thread arrives
  if (elect_one()) {
    init(&bar, 1);
    // issue two TMA bulk copy operations
    cuda::device::memcpy_async_tx(smem1, gmem1, cuda::aligned_size_t<16>(tile_size * sizeof(int)   ), bar);
    cuda::device::memcpy_async_tx(smem2, gmem2, cuda::aligned_size_t<16>(tile_size * sizeof(double)), bar);
    // arrive and update the barrier's expect_tx with the **total** number of loaded bytes
    (void)cuda::device::barrier_arrive_tx(bar, 1, tile_size * (sizeof(int) + sizeof(double)));
  }
  __syncthreads(); // need to sync so the barrier is set up when the other threads arrive and wait
  // wait for the current barrier phase to complete
  bar.wait_parity(0);
  // process data in smem ...
}
```
Compiled for sm100, the SASS stays identical.

### DIFF
--- a/libcudacxx/include/cuda/__barrier/barrier_block_scope.h
+++ b/libcudacxx/include/cuda/__barrier/barrier_block_scope.h
@@ -69,7 +69,7 @@ class barrier<thread_scope_block, ::cuda::std::__empty_completion> : public __bl
   __barrier_base __barrier;
 
   _CCCL_DEVICE friend inline ::cuda::std::uint64_t* ::cuda::device::_LIBCUDACXX_ABI_NAMESPACE::barrier_native_handle(
-    barrier& __b);
+    barrier<thread_scope_block>& __b);
 
   _CCCL_DEVICE ::cuda::std::uint64_t* __native_handle() const
   {


libcudacxx/include/cuda/__barrier/barrier_block_scope.h
```diff
@@ -141,20 +144,17 @@ class barrier<thread_scope_block, ::cuda::std::__empty_completion> : public __bl
...
           return __barrier.arrive(__update);
-        } else if (!::cuda::device::is_object_from(__barrier, ::cuda::device::address_space::shared)) { ::__trap(); }
-        // Cannot use cuda::device::barrier_native_handle here, as it is
-        // only defined for block-scope barriers. This barrier may be a
-        // non-block scoped barrier.
-        auto __bh = reinterpret_cast<::cuda::std::uint64_t*>(&__barrier);
-        __token   = ::cuda::ptx::mbarrier_arrive(__bh, __update);),
+        } else if (!::cuda::device::is_object_from(__barrier, ::cuda::device::address_space::shared)) {
+          ::__trap();
+        } __token = ::cuda::ptx::mbarrier_arrive(__native_handle(), __update);),
```
miscco: This formatting is weird, do we want to pull it out into a separate function?

libcudacxx/include/cuda/__barrier/barrier_block_scope.h
```diff
@@ -179,44 +179,21 @@ class barrier<thread_scope_block, ::cuda::std::__empty_completion> : public __bl
...
           ::__trap();
-        } asm volatile("{\n\t"
-                       ".reg .pred p;\n\t"
-                       "mbarrier.try_wait.shared.b64 p, [%1], %2;\n\t"
-                       "selp.b32 %0, 1, 0, p;\n\t"
-                       "}" : "=r"(__ready) : "r"(
-                         static_cast<::cuda::std::uint32_t>(::__cvta_generic_to_shared(&__barrier))),
-                       "l"(__token) : "memory");
-        return __ready;),
+        } return ::cuda::ptx::mbarrier_try_wait(__native_handle(), __token);),
```
miscco: Wound an else here help?
```suggestion
        } 
        else
        {
          return ::cuda::ptx::mbarrier_try_wait(__native_handle(), __token);
        }),
```

miscco (commented):
looks technically correct, but the formatting is atrocious. Could we add an `else` to the conditions, as all early branches are returning?

libcudacxx/include/cuda/__barrier/barrier_block_scope.h
bernhardmgruber: I created lots of separate functions now where the dispatch was big.

libcudacxx/include/cuda/__barrier/barrier_block_scope.h
```diff
@@ -87,140 +94,184 @@ class barrier<thread_scope_block, ::cuda::std::__empty_completion> : public __bl
...
+  _CCCL_DEVICE_API _CCCL_FORCEINLINE arrival_token __arrive_sm90(::cuda::std::ptrdiff_t __update)
+  {
+    if (!::cuda::device::is_object_from(__barrier, ::cuda::device::address_space::cluster_shared))
+    {
+      return __barrier.arrive(__update);
+    }
+    if (!::cuda::device::is_object_from(__barrier, ::cuda::device::address_space::shared))
+    {
+      ::__trap();
+    }
```
miscco: This is strange, because the first condition takes anything but `cluster_shared`, so the second one seems wrong

bernhardmgruber: Anything that is `shared` is also `cluster_shared`. Because the shared memory address space is part of the cluster shared memory space.

bernhardmgruber: Here, we trap for any barrier that is in cluster shared memory, but not in the shared memory of the current CTA.

miscco: could we turn that into an else if

libcudacxx/include/cuda/__barrier/barrier_block_scope.h
```diff
@@ -87,140 +94,184 @@ class barrier<thread_scope_block, ::cuda::std::__empty_completion> : public __bl
...
-        } else if (::cuda::device::is_object_from(__barrier, ::cuda::device::address_space::cluster_shared)) {
-          ::__trap();
-        }),
-      NV_PROVIDES_SM_80,
-      (if (::cuda::device::is_object_from(__barrier, ::cuda::device::address_space::shared)) {
-        asm volatile("mbarrier.inval.shared.b64 [%0];" ::"r"(static_cast<::cuda::std::uint32_t>(
-          ::__cvta_generic_to_shared(&__barrier)))
-                     : "memory");
-      }))
+      (if (::cuda::device::is_object_from(__barrier, ::cuda::device::address_space::cluster_shared)) { ::__trap(); }))
```
fbusato: `cuda::std::terminate()` ?

libcudacxx/include/cuda/__barrier/barrier_block_scope.h
```diff
@@ -87,140 +94,184 @@ class barrier<thread_scope_block, ::cuda::std::__empty_completion> : public __bl
...
-         "}" : "=r"(__ready) : "r"(static_cast<::cuda::std::uint32_t>(::__cvta_generic_to_shared(&__barrier))),
-         "l"(__token) : "memory");))
-    return __ready;
+    if (!::cuda::device::is_object_from(__barrier, ::cuda::device::address_space::cluster_shared))
+    {
+      return ::cuda::std::__call_try_wait(__barrier, ::cuda::std::move(__token));
+    }
+    if (!::cuda::device::is_object_from(__barrier, ::cuda::device::address_space::shared))
+    {
+      ::__trap();
```
fbusato: could be a `_CCCL_ASSERT` instead?

libcudacxx/include/cuda/__barrier/barrier_block_scope.h
```diff
@@ -87,140 +94,184 @@ class barrier<thread_scope_block, ::cuda::std::__empty_completion> : public __bl
...
+    if (!::cuda::device::is_object_from(__barrier, ::cuda::device::address_space::shared))
+    {
+      return __barrier.arrive(__update);
+    }
+
+    unsigned int __mask    = ::__activemask();
+    unsigned int __activeA = ::__match_any_sync(__mask, __update);
+    unsigned int __activeB = ::__match_any_sync(__mask, reinterpret_cast<::cuda::std::uintptr_t>(&__barrier));
+    unsigned int __active  = __activeA & __activeB;
+    int __inc              = ::__popc(__active) * __update;
```
fbusato: Probably not worth to move to their C++ versions

libcudacxx/include/cuda/__barrier/barrier_block_scope.h
fbusato: could not be a precondition instead?

libcudacxx/include/cuda/__barrier/barrier_block_scope.h
bernhardmgruber: I agree, we should probably assert here. And check the documentation whether we make it clear that barriers ought not live in cluster shared memory.

libcudacxx/include/cuda/__barrier/barrier_block_scope.h
bernhardmgruber: Replaced by `_CCCL_ASSERT`

libcudacxx/include/cuda/__barrier/barrier_block_scope.h
bernhardmgruber: Agreed. I don't want to touch too much of the existing code, especially if I don't understand what's going on :)

libcudacxx/include/cuda/__barrier/barrier_block_scope.h
bernhardmgruber: Replaced by `_CCCL_ASSERT`

bernhardmgruber:
- [x] affer introducing all the assertions, we should check whether SASS did not change. Maybe the compiler uses some of the traps to optimize.

PR #6251, opened 2025-10-15
Fix type parameter in `cub::ThreadReduce` function

fbusato:
## Description

Fix a typo related `cub::ThreadReduce`  call.

### DIFF
--- a/cub/cub/thread/thread_reduce.cuh
+++ b/cub/cub/thread/thread_reduce.cuh
@@ -533,7 +533,7 @@ ThreadReduce(const Input& input, ReductionOp reduction_op, PrefixT prefix)
   {
     array[i + 1] = input[i];
   }
-  return cub::ThreadReduce<decltype(array), ReductionOp, AccumT, AccumT>(array, reduction_op);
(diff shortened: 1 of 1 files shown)


bernhardmgruber (approved):

cub/cub/thread/thread_reduce.cuh
```diff
@@ -533,7 +533,7 @@ ThreadReduce(const Input& input, ReductionOp reduction_op, PrefixT prefix)
   {
     array[i + 1] = input[i];
   }
-  return cub::ThreadReduce<decltype(array), ReductionOp, AccumT, AccumT>(array, reduction_op);
```
miscco, line 536: I do not believe this is correct, because we copied `Input` into `array`, so the value type of `array` is AccumT

What I was overlooking is that the `ThreadReduce` function is forward declared, so I was thinking that a user would have to explicitly define the template arguments.

However, this is not the case, so this use case is fine

fbusato: you are right. The third parameter is the type of the `array` element which is `AccumT` not `ValueT`

PR #6252, opened 2025-10-15
[libcu++] Enable complex literals

davebayer:


libcudacxx/include/cuda/std/__complex/literals.h
```diff
@@ -27,22 +27,18 @@
...
 
 inline namespace literals
 {
 inline namespace complex_literals
 {
-#  if !_CCCL_CUDA_COMPILER(NVCC) && !_CCCL_COMPILER(NVRTC)
-// NOTE: if you get a warning from GCC <7 here that "literal operator suffixes not preceded by ‘_’ are reserved for
-// future standardization" then we are sorry. The warning was implemented before GCC 7, but can only be disabled since
-// GCC 7. See also: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=69523
 _CCCL_API constexpr complex<long double> operator""il(long double __im)
```
fbusato, line 46: my understanding is that we don't support GCC <7, we can get rid of

davebayer: Yeah

libcudacxx/include/cuda/std/__complex/literals.h
```diff
@@ -21,28 +21,28 @@
 #  pragma system_header
 #endif // no system header
 
-#include <cuda/std/__complex/complex.h>
+// gcc < 8 warns about it's extended literals being shadowed by the implementation, so let's just disable the complex
+// literals
+#if !_CCCL_COMPILER(GCC, <, 8)
```
davebayer, line 26: I had to disable the literals for gcc 7 and older..

PR #6254, opened 2025-10-15
Windows CI: CCCL C Parallel

alliepiper:


c/parallel/test/test_radix_sort.cpp
```diff
@@ -314,7 +314,7 @@ C2H_TEST("DeviceRadixSort::SortPairs works", "[radix_sort]", test_params_tuple)
   if (is_descending)
   {
     std::sort(expected_keys.begin(), expected_keys.end(), std::greater<KeyT>());
-    std::sort(expected_items.begin(), expected_items.end(), std::greater<KeyT>());
```
tpn, line 317: Oof, how was that working on Linux?!

c2h/generators_vector.cu
```diff
@@ -83,12 +83,11 @@ VEC_SPECIALIZATION(int4);
...
 VEC_SPECIALIZATION(long2);
 VEC_SPECIALIZATION(long3);
-_CCCL_SUPPRESS_DEPRECATED_PUSH
-VEC_SPECIALIZATION(long4);
-_CCCL_SUPPRESS_DEPRECATED_POP
 #  if _CCCL_CTK_AT_LEAST(13, 0)
 VEC_SPECIALIZATION(long4_16a);
 VEC_SPECIALIZATION(long4_32a);
+#  else
+VEC_SPECIALIZATION(long4);
```
bernhardmgruber, line 90: Remark: This makes `long4` no longer usable for test data generation in unit tests on CTK >= 13.0. I previously changed all uses of `long4` for those versions to the `_16a` and `_32a` variants to avoid warnings, but wanted to revisit this.

On second thought, maybe we are fine with no longer testing with `long4` in CTK >= 13.0. So this is not a blocker.

alliepiper: Agreed, I think this is the cleanest way around the headache of trying to actually suppress these warnings. MSVC likes to still complain even though they were suppressed >.<

PR #6262, opened 2025-10-16
Pass a device array or `None` as the initial value to cuda.compute scan

### LINKED ISSUE #6261: [cuda.compute]: Allow passing a device array or `None` as the initial value in scan
### Is this a duplicate?

### Area

cuda.compute (Python)

CUB includes variants of DeviceScan that accept as the initial value either a `FutureValue` or no initial value at all. We should expose those variants to cuda.compute instead of just a host initial value, which is the only option that exists in cuda.compute today.

Allow the user to pass in either a device array that implements `__cuda_array_interface__` or `None`, which will internally use either a `FutureValue` or a `NullType` respectively.

_No response_

### Additional context

_No response_

NaderAlAwar:
## Description

closes #6261 

## Checklist

### DIFF
--- a/python/cuda_cccl/cuda/compute/_bindings_impl.pyx
+++ b/python/cuda_cccl/cuda/compute/_bindings_impl.pyx
@@ -591,7 +591,6 @@ cdef class Iterator:
     cdef object state_obj
     cdef object host_advance_obj
     cdef cccl_iterator_t iter_data
-    cdef TypeInfo value_type
 
     def __cinit__(self,
         int alignment,
@@ -657,7 +656,6 @@ cdef class Iterator:
         self.iter_data.type = <cccl_iterator_kind_t> it_kind
         self.iter_data.advance = self.advance.op_data
         self.iter_data.dereference = self.dereference.op_data
-        self.value_type = value_type
         self.iter_data.value_type = value_type.type_info
 
     @property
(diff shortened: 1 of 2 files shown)


python/cuda_cccl/cuda/compute/_bindings_impl.pyx
```diff
@@ -947,7 +958,8 @@ cdef extern from "cccl/c/scan.h":
         cccl_iterator_t,
         cccl_iterator_t,
         cccl_op_t,
-        cccl_value_t,
+        cccl_type_info,
+        _Bool,
         _Bool,
```
oleksandr-pavlyk: In the new signature, should arguments be `cccl_type_info, _Bool, InitKind` instead?

python/cuda_cccl/cuda/compute/_bindings_impl.pyx
```diff
@@ -586,6 +591,7 @@ cdef class Iterator:
     cdef object state_obj
     cdef object host_advance_obj
     cdef cccl_iterator_t iter_data
+    cdef TypeInfo value_type
```
oleksandr-pavlyk: There is no need to add this class member, as it is duplicated in `self.iter_data.value_type`.

We can implement the property as follows without this field:

```python
    @property
    def value_type(self):
        cdef cccl_type_info _ti = self.iter_data.iter_data.value_type 
        return TypeInfo(_ti.size, _ti.alignment, _ti.type)
```

python/cuda_cccl/cuda/compute/_bindings_impl.pyx
NaderAlAwar: Good catch, fixed

oleksandr-pavlyk (approved):
LGTM @NaderAlAwar. Support for initial values being a device arrays is super-useful.

PR #6263, opened 2025-10-16
Do no use `bit_cast` to work around initialization issues with barrier

### LINKED ISSUE #6255: [BUG]: barrier tests fail to compile w/ clang:
### Is this a duplicate?

### Type of Bug

Compile-time Error

### Component

libcu++

### Describe the bug

A handful of barrier tests fail to build with clang due to their attempts to use `bit_cast<barrier>` which fails to compile because `barrier` is not trivially copyable.

https://godbolt.org/z/M6qjz1hsh

E.g:
https://github.com/NVIDIA/cccl/blob/813e75eb761abf1be641d89c2d8f0a1ecb9a48e2/libcudacxx/test/libcudacxx/cuda/barrier/cp_async_bulk.pass.cpp#L44-L48

```
<source>:16:18: error: no matching function for call to 'bit_cast'
   16 |   barrier& bar = cuda::std::bit_cast<barrier>(barrier_data);
      |                  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
/opt/compiler-explorer/libs/cccl/trunk/libcudacxx/include/cuda/std/__bit/bit_cast.h:52:61: note: candidate template ignored: requirement 'is_trivially_copyable_v<cuda::barrier<cuda::std::thread_scope_block, cuda::std::__empty_completion>> || __is_extended_floating_point_v<cuda::barrier<cuda::std::thread_scope_block, cuda::std::__empty_completion>>' was not satisfied [with _To = barrier, _From = char[8], $2 = 0]
   52 | [[nodiscard]] _CCCL_API inline _CCCL_CONSTEXPR_BIT_CAST _To bit_cast(const _From& __from) noexcept

```

### How to Reproduce

https://godbolt.org/z/M6qjz1hsh

### Expected behavior

Test should be compileable.

### Reproduction link

https://godbolt.org/z/M6qjz1hsh

### Operating System

_No response_

### nvidia-smi output

_No response_

### NVCC version

_No response_

miscco:
None of the tests is constexpr so we can just use `reinterpret_cast`

Fixes #6255

### DIFF
--- a/libcudacxx/test/libcudacxx/cuda/barrier/cp_async_bulk.pass.cpp
+++ b/libcudacxx/test/libcudacxx/cuda/barrier/cp_async_bulk.pass.cpp
@@ -43,7 +43,7 @@ __device__ void test()
   alignas(16) __shared__ int smem_buffer[buf_len];
 #if _CCCL_CUDA_COMPILER(CLANG)
   __shared__ char barrier_data[sizeof(barrier)];
-  barrier& bar = cuda::std::bit_cast<barrier>(barrier_data);
+  barrier& bar = reinterpret_cast<barrier&>(barrier_data);
 #else // ^^^ _CCCL_CUDA_COMPILER(CLANG) ^^^ / vvv !_CCCL_CUDA_COMPILER(CLANG)
   __shared__ barrier bar;
 #endif // !_CCCL_CUDA_COMPILER(CLANG)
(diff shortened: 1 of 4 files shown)


libcudacxx/test/libcudacxx/cuda/barrier/cp_async_bulk.pass.cpp
```diff
@@ -43,7 +43,7 @@ __device__ void test()
   alignas(16) __shared__ int smem_buffer[buf_len];
 #if _CCCL_CUDA_COMPILER(CLANG)
   __shared__ char barrier_data[sizeof(barrier)];
-  barrier& bar = cuda::std::bit_cast<barrier>(barrier_data);
+  barrier& bar = *reinterpret_cast<barrier*>(&barrier_data);
```
davebayer:
```suggestion
  barrier& bar = reinterpret_cast<barrier&>(barrier_data);
```

davebayer: How could the original solution even work? `bit_cast` returns a new object, it doesn't reference the old one..

davebayer (approved):

PR #6266, opened 2025-10-16
Deprecate `<cuda/stream_ref>` header

davebayer:
Users should use `<cuda/stream>` instead


davebayer:
pre-commit.ci autofix

miscco (approved):

PR #6281, opened 2025-10-17
Cleanup and simplify structured bindings support

miscco:
We do not need to pull in all of `<utility>` it suffices to just forward declare the std struct

### DIFF
--- a/libcudacxx/include/cuda/std/__tuple_dir/structured_bindings.h
+++ b/libcudacxx/include/cuda/std/__tuple_dir/structured_bindings.h
@@ -31,6 +31,10 @@ _CCCL_DIAG_SUPPRESS_CLANG("-Wmismatched-tags")
 #include <cuda/std/__fwd/tuple.h>
 #include <cuda/std/__tuple_dir/tuple_element.h>
 #include <cuda/std/__tuple_dir/tuple_size.h>
+#include <cuda/std/__type_traits/enable_if.h>
+#include <cuda/std/__type_traits/integral_constant.h>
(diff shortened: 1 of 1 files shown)


libcudacxx/include/cuda/std/__tuple_dir/structured_bindings.h
```diff
@@ -23,193 +23,137 @@
...
-struct tuple_size<volatile ::cuda::std::array<_Tp, _Size>> : ::cuda::std::tuple_size<::cuda::std::array<_Tp, _Size>>
-{};
-
-template <class _Tp, size_t _Size>
-struct tuple_size<const volatile ::cuda::std::array<_Tp, _Size>>
-    : ::cuda::std::tuple_size<::cuda::std::array<_Tp, _Size>>
-{};
+struct tuple_size<::cuda::std::array<_Tp, _Size>>
+{
+  static constexpr size_t value = _Size;
```
fbusato, line 102: considering the compile time and that we more likely use `tuple_size_v`, would not be better to specialize the template variable and define the structure relying on it?

davebayer: No, because the standard says users are able to specialize the structure.

miscco: Its even worse, we cannot internally use `tuple_size_v` because it is only valid if `tuple_size` is specialized

miscco:
pre-commit.ci autofix

PR #6283, opened 2025-10-17
Implement `cudax::read`

davebayer:
Follow up to #6122. Please, filter only the last commit.

This PR is a preview of the `cudax::read` API.

### DIFF
--- a/cudax/include/cuda/experimental/__cufile/read.cuh
+++ b/cudax/include/cuda/experimental/__cufile/read.cuh
@@ -1,0 +1,370 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of CUDA Experimental in CUDA C++ Core Libraries,
+// under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES.
+//
+//===----------------------------------------------------------------------===//
+#pragma once
+
+#include <cuda/std/detail/__config>
+
+#if defined(_CCCL_IMPLICIT_SYSTEM_HEADER_GCC)
+#  pragma GCC system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_CLANG)
+#  pragma clang system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_MSVC)
+#  pragma system_header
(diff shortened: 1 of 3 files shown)


cudax/include/cuda/experimental/__cufile/read.cuh
```diff
@@ -0,0 +1,457 @@
...
+  }
+
+  _CCCL_TRY_CUFILE_API(
+    ::cuFileReadAsync, "Failed to submit asynchronous read.", __file.get(), __dst, __nbytes_p, __foffset_p, __doffset_p);
+
+  // Ignore the result of the callback. If the call fails, there will be a small memory leak. We cannot do anything
+  // about it, because cuFIle will write the results there.
+  // todo: replace with ::cuda::__driver:: call
+  const auto __ignore = ::cuStreamAddCallback(
+    __stream.get(), __cufile_io_submit_result_data::__maybe_destroy_callback, __result.__data_, 0);
```
jvera-nvidia: what is this __result ? i don't see the definition thi this __data_ field

cudax/include/cuda/experimental/__cufile/read.cuh
```diff
@@ -0,0 +1,457 @@
...
+  ::off_t __foffset_{};
+  ::cuda::std::size_t __nbytes_{};
+  ::off_t __doffset_{};
+  ::ssize_t __result_{};
+};
+
+template <class _FOffset, class _NBytes, class _DOffset>
+class cufile_io_submit_result
+{
+  __cufile_io_submit_result_data* __data_{};
```
jvera-nvidia: can't we use unique_ptr here instead of implementing the ref counting?

davebayer: We cannot. To prevent the variables going out of scope, I implemented the reference counting.

It works like this:
1. The structure containing the in/out data is allocated.
2. The known values are set.
3. The operation is enqueued in the stream with values pointing to the allocated structure.
4. A callback that will delete the structure is enqueued into the same stream.
5. The result object is created from the allocated structure and returned.

The idea is that the allocated structure contains a reference counter. At the beginning, both the callback and the returned object reference the object.

Once the callback was executed and the result object goes out of scope, the data structure is deleted.

This mechanism prevents writes to variables that are already out of scope and at the same time enables to delay the arguments definition.

cudax/include/cuda/experimental/__cufile/read.cuh
davebayer: Oops it should be just the `__data` pointer

davebayer:
Closing for now, we can reopen this once we get back to the cuFile APIs

PR #6284, opened 2025-10-17
Implement `cudax::cufile_batch`

davebayer:
Follow up to #6122. Please, filter only the last commit.

This PR is a preview of the `cudax::cufile_batch` API.

### DIFF
--- a/cudax/include/cuda/experimental/__cufile/batch.cuh
+++ b/cudax/include/cuda/experimental/__cufile/batch.cuh
@@ -116,6 +116,14 @@ public:
            ? ::cuda::std::optional{__io_events_.ret}
            : ::cuda::std::nullopt;
   }
+
+  //! @brief Gets the cookie the batch request was created with.
+  //!
+  //! @return The cookie.
+  [[nodiscard]] _CCCL_HOST_API void* cookie() const noexcept
+  {
+    return __io_events_.cookie;
+  }
 };
 
 //! @brief Type wrapping the cuFile batch query operations.
--- a/cudax/include/cuda/experimental/__cufile/batch.cuh
+++ b/cudax/include/cuda/experimental/__cufile/batch.cuh
@@ -1,0 +1,295 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of CUDA Experimental in CUDA C++ Core Libraries,
+// under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES.
+//
+//===----------------------------------------------------------------------===//
+#pragma once
+
+#include <cuda/std/detail/__config>
+
+#if defined(_CCCL_IMPLICIT_SYSTEM_HEADER_GCC)
+#  pragma GCC system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_CLANG)
+#  pragma clang system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_MSVC)
+#  pragma system_header
+#endif // no system header
+
+#include <cuda/__stream/stream_ref.h>
+#include <cuda/std/__cccl/memory_wrapper.h>
+#include <cuda/std/__chrono/duration.h>
+#include <cuda/std/__memory/uninitialized_algorithms.h>
+#include <cuda/std/__utility/exchange.h>
+#include <cuda/std/cstddef>
+#include <cuda/std/optional>
+
+#include <cuda/experimental/__cufile/cufile_ref.cuh>
+#include <cuda/experimental/__cufile/exception.cuh>
+
+#include <vector>
+
+#include <cufile.h>
+
+namespace cuda::experimental
+{
+
+//! @brief Enum for the cuFile batch request state.
+enum class cufile_batch_request_state
+{
+  waiting, //!< Job hasn't been submitted yet.
+  pending, //!< Job was enqueued.
+  invalid, //!< The request was ill-formed or could not be enqueued.
+  canceled, //!< The request was successfully canceled.
+  complete, //!< The request was successfully completed.
+  timeout, //!< The request timed out.
+  failed, //!< The request was unable to complete.
+};
+
+//! @brief Class wrapping the cuFile batch request state operations.
+class cufile_batch_request
+{
+  friend class cufile_batch;
+
+  ::CUfileIOEvents_t __io_events_;
+
+  _CCCL_HIDE_FROM_ABI cufile_batch_request() noexcept = default;
+
+public:
+  cufile_batch_request(const cufile_batch_request&)            = delete;
+  cufile_batch_request(cufile_batch_request&&)                 = delete;
+  cufile_batch_request& operator=(const cufile_batch_request&) = delete;
+  cufile_batch_request& operator=(cufile_batch_request&&)      = delete;
+
+  //! @brief Gets the request state.
+  //!
+  //! @return The request state.
+  [[nodiscard]] _CCCL_API cufile_batch_request_state state() const noexcept
+  {
+    switch (__io_events_.status)
+    {
+      case ::CUFILE_WAITING:
+        return cufile_batch_request_state::waiting;
+      case ::CUFILE_PENDING:
+        return cufile_batch_request_state::pending;
+      case ::CUFILE_INVALID:
+        return cufile_batch_request_state::invalid;
+      case ::CUFILE_CANCELED:
+        return cufile_batch_request_state::canceled;
+      case ::CUFILE_COMPLETE:
+        return cufile_batch_request_state::complete;
+      case ::CUFILE_TIMEOUT:
+        return cufile_batch_request_state::timeout;
+      case ::CUFILE_FAILED:
+        return cufile_batch_request_state::failed;
+      default:
+        _CCCL_UNREACHABLE();
+    }
+  }
+
+  //! @brief Gets the number of transferred bytes by the request, written or read.
+  //!
+  //! @return The number of transferred bytes. If the state is not \c cuda::cufile_batch_request_state::complete, an
+  //!         empty optional is returned.
+  [[nodiscard]] _CCCL_API ::cuda::std::optional<::cuda::std::size_t> transferred_nbytes() const noexcept
+  {
+    return (state() == cufile_batch_request_state::complete)
+           ? ::cuda::std::optional{__io_events_.ret}
+           : cuda::std::nullopt;
+  }
+};
+
+//! @brief Type wrapping the cuFile batch query operations.
+class cufile_batch
+{
+  friend class cufile_batch_requests;
+
+  ::CUfileBatchHandle_t __batch_{};
+  unsigned __nrequests_{};
+  ::std::unique_ptr<cufile_batch_request[]> __requests_{};
+
+  _CCCL_HOST_API cufile_batch(::CUfileBatchHandle_t __batch, unsigned __nrequests) noexcept
+      : __batch_{__batch}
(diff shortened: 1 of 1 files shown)


cudax/include/cuda/experimental/__cufile/batch.cuh
```diff
@@ -0,0 +1,295 @@
...
+        return cufile_batch_request_state::failed;
+      default:
+        _CCCL_UNREACHABLE();
+    }
+  }
+
+  //! @brief Gets the number of transferred bytes by the request, written or read.
+  //!
+  //! @return The number of transferred bytes. If the state is not \c cuda::cufile_batch_request_state::complete, an
+  //!         empty optional is returned.
```
jvera-nvidia, line 112: why not just return 0? that would be conceptually accurate and simpler

cudax/include/cuda/experimental/__cufile/batch.cuh
```diff
@@ -0,0 +1,295 @@
...
+  friend class cufile_batch_requests;
+
+  ::CUfileBatchHandle_t __batch_{};
+  unsigned __nrequests_{};
+  ::std::unique_ptr<cufile_batch_request[]> __requests_{};
+
+  _CCCL_HOST_API cufile_batch(::CUfileBatchHandle_t __batch, unsigned __nrequests) noexcept
+      : __batch_{__batch}
+      , __nrequests_{__nrequests}
+      , __requests_{new cufile_batch_request[__nrequests]}
```
jvera-nvidia: why not make_unique? it doesnt make a hell of a difference here (actually none) but its educative to avoid those new/deletes

cudax/include/cuda/experimental/__cufile/batch.cuh
```diff
@@ -0,0 +1,295 @@
...
+  //! @brief Checks whether the object is valid.
+  [[nodiscard]] _CCCL_HOST_API bool __valid() const noexcept
+  {
+    return __batch_ != nullptr;
+  }
+
+  //! @brief Updates the tracked requests.
+  //!
+  //! @return The number of active requests.
+  [[nodiscard]] _CCCL_HOST_API unsigned __update_requests(::timespec* __timeout = nullptr) const
```
jvera-nvidia: i'd like to avoid exposing brittle timespec* and use chrono instead

cudax/include/cuda/experimental/__cufile/batch.cuh
```diff
@@ -0,0 +1,295 @@
...
+  //!
+  //! todo: params
+  _CCCL_API void add_read_bytes(
+    cufile_ref __file,
+    cufile_ref::off_type __foffset,
+    ::cuda::std::byte* __dst,
+    ::cuda::std::size_t __nbytes,
+    ::cuda::std::ptrdiff_t __doffset = 0)
+  {
+    ::CUfileIOParams_t& __new   = __io_params_.emplace_back();
```
jvera-nvidia: any reason for no using auto here?

cudax/include/cuda/experimental/__cufile/batch.cuh
davebayer: Because:
1. `new` is completely fine here
2. `make_unique` instantiates a function without any additional benefit in this case
3. I would need `make_unique_for_overwrite` here which is only available since C++23

cudax/include/cuda/experimental/__cufile/batch.cuh
davebayer: Because 0 could be a valid value. I want the result to be unavailable until the state is in `complete` state

cudax/include/cuda/experimental/__cufile/batch.cuh
davebayer: This is an internal function. I haven't implemented the waiting for the requests yet

cudax/include/cuda/experimental/__cufile/batch.cuh
davebayer: Yes, you cannot see clearly the returned value, so I find this more readable

cudax/include/cuda/experimental/__cufile/batch.cuh
```diff
@@ -0,0 +1,295 @@
...
+  pending, //!< Job was enqueued.
+  invalid, //!< The request was ill-formed or could not be enqueued.
+  canceled, //!< The request was successfully canceled.
+  complete, //!< The request was successfully completed.
+  timeout, //!< The request timed out.
+  failed, //!< The request was unable to complete.
+};
+
+//! @brief Class wrapping the cuFile batch request state operations.
+class cufile_batch_request
```
davebayer: 1. Do we want to be able to call `.update()` method on a single request?
2. Do we want to have `.is_done()` method (completed/failed/canceled/...)?

cudax/include/cuda/experimental/__cufile/batch.cuh
```diff
@@ -0,0 +1,295 @@
...
+  //! @return The span of tracked requests.
+  [[nodiscard]] _CCCL_HOST_API ::cuda::std::span<const cufile_batch_request> requests() const noexcept
+  {
+    __update_requests();
+    return ::cuda::std::span{__requests_.get(), static_cast<::cuda::std::size_t>(__nrequests_)};
+  }
+};
+
+//! @brief Class wrapping the setup of the \c cuda::cufile_batch.
+class cufile_batch_requests
```
davebayer: 1. Is this class even needed? Wouldn't be a better design choice to rename `cufile_batch_request` to `cufile_batch_job` and add a new class `cufile_batch_request` that would represent a single request and would be set up by the user. The user would create a contiguous range of the requests and then would pass it to a new api, for example `cufile_commit_batch_requests(/*request-range*/)` which would return a `cufile_batch` object.

     I'd really like to separate the requests and the committed batch, because it will remove cases when `.sync()` method would be called on a non-committed batch.

2. How to expose all of the read/write variants? If we decide to go for the solution I proposed in (1), I would suggest either:
      - create an `inline constexpr` object for each signature in a similar style as are the objects `cufile_driver_attributes` namespace that would be put in the `cufile_batch_request` constructor as `cufile_batch_request{cufile_batch_op::read, ...}`
      - implement free functions such as `make_cufile_read_batch_request(...)`

    I am not quite sure what approach would be more suitable.

cudax/include/cuda/experimental/__cufile/batch.cuh
jvera-nvidia: i'd like to use vector here instead of nrequests + requests

cudax/include/cuda/experimental/__cufile/batch.cuh
jvera-nvidia: I agree, I'd separate what is a request (should be a POD or nearly a POD) and what a batch is and what a job is.

I think I'd prefer second option, since its easier for the API clients to explicitly create the requests they want and more friendly for IDEs

cudax/include/cuda/experimental/__cufile/batch.cuh
```diff
@@ -0,0 +1,295 @@
...
+  }
+
+  //! @brief Cancels all tracked requests.
+  _CCCL_HOST_API void cancel() const
+  {
+    if (__valid())
+    {
+      _CCCL_TRY_CUFILE_API(::cuFileBatchIOCancel, "Failed to cancel IO batch.", __batch_);
+    }
+  }
```
jvera-nvidia: this shouldn't be const since its changing internal state

cudax/include/cuda/experimental/__cufile/batch.cuh
```diff
@@ -0,0 +1,295 @@
...
+
+  //! @brief Checks if all of the tracked requests are done.
+  //!
+  //! @return \c true if done, \c false otherwise.
+  [[nodiscard]] _CCCL_HOST_API bool is_done() const
+  {
+    return __update_requests() == 0;
+  }
```
jvera-nvidia: this can't be const since its changing internal state (__update_requests() indeed modifies __requests object and don't do it in a thread safe manner)

cudax/include/cuda/experimental/__cufile/batch.cuh
```diff
@@ -0,0 +1,295 @@
...
+
+  //! @brief Gets the span of the tracked requests.
+  //!
+  //! @return The span of tracked requests.
+  [[nodiscard]] _CCCL_HOST_API ::cuda::std::span<const cufile_batch_request> requests() const noexcept
+  {
+    __update_requests();
+    return ::cuda::std::span{__requests_.get(), static_cast<::cuda::std::size_t>(__nrequests_)};
+  }
+};
```
jvera-nvidia: same here, this can not be const

cudax/include/cuda/experimental/__cufile/batch.cuh
jvera-nvidia: this can't be const since cuFileBatchIOGetStatus modifies `__requests_` array (the `__io_events_ `member).

cudax/include/cuda/experimental/__cufile/batch.cuh
davebayer: vector stores the `capacity` member and we don't need to resize the array once allocated, I would prefer to use this approach

cudax/include/cuda/experimental/__cufile/batch.cuh
davebayer: Good catch

cudax/include/cuda/experimental/__cufile/batch.cuh
jvera-nvidia: 1) I don't think so, if its a batch we should care only about the batch. The underlying API works also on the whole batch (not individual requests/jobs)
2) I think it could be convenient but I would avoid calling update from within, since then i'd found the naming  missleading. 

I'd rather have an update explicit public call instead of implicitly calling it from within these accessors; avoid to have side effects in these queries.

cudax/include/cuda/experimental/__cufile/batch.cuh
```diff
@@ -0,0 +1,307 @@
...
+  [[nodiscard]] _CCCL_HOST_API unsigned __update_jobs(::timespec* __timeout = nullptr)
+  {
+    if (!__valid())
+    {
+      return 0;
+    }
+
+    auto __nr = __njobs_;
+    _CCCL_TRY_CUFILE_API(
+      ::cuFileBatchIOGetStatus,
```
davebayer, line 160: @jvera-nvidia can you find out if `nullptr` can be passed as the timeout parameter here, meaning no timeout? The documentation doesn't say.

cudax/include/cuda/experimental/__cufile/batch.cuh
```diff
@@ -0,0 +1,307 @@
...
+      &__jobs_.get()->__io_events_,
+      __timeout);
+    return __nr;
+  }
+
+public:
+  //! @brief Makes a read request.
+  //!
+  //! todo: params
+  [[nodiscard]] static _CCCL_HOST_API cufile_batch_request make_read_bytes_request(
```
davebayer, line 174: Does `cufile_batch::make_read_bytes_request(...)` look good? I think it's better then `make_cufile_batch_read_bytes_request(...)`

cudax/include/cuda/experimental/__cufile/batch.cuh
davebayer: It can

cudax/include/cuda/experimental/__cufile/batch.cuh
jvera-nvidia: yes indeed, I'd prefer the first one; btw what is your take in the prefix build vs make ? I find build more readable

davebayer: I would definitely go for `make_`, C++ standard library uses that for example in `std::make_unique(...)` or `std::make_optional(...)`

davebayer:
Closing for now, we can reopen this once we get back to the cuFile APIs

PR #6288, opened 2025-10-18
Extract BlockScan code-block examples to literalinclude 1/3

### LINKED ISSUE #6285: [DOC]: Typos in example code, will not compile without edits
### Is this a duplicate?

New

From Tim Davis:
Davis, Timothy A 11:20 AM
https://nvidia.github.io/cccl/cub/api/classcub_1_1BlockScan.html#_CPPv4N3cub9BlockScan12InclusiveSumE1TR1T
Search for “incluisve” typo

There are also two more typos in the BlockLoad example:  one is use of ellipses in the declaration of the example kernel, the other is use of a pointer type instead of the element type <int* vs <int  in the template args. 

https://nvidia.github.io/cccl/cub/api/classcub_1_1BlockScan.html#_CPPv4N3cub9BlockScan12InclusiveSumE1TR1T
Search for “incluisve” typo

gonidelis:
fixes #6285

and takes inspiration from it to start moving the examples in the `block_scan.cuh` header in separate example executable `example_block_scan.cu` section so that they compile and work out of the box without the user having to tweak the code to make it run.

To make the PR review friendly I split the work over 3 PRs

### DIFF
--- a/cub/cub/block/block_scan.cuh
+++ b/cub/cub/block/block_scan.cuh
@@ -172,25 +172,11 @@ enum BlockScanAlgorithm
 //! are partitioned in a :ref:`blocked arrangement <flexible-data-arrangement>` across 128 threads
 //! where each thread owns 4 consecutive items.
 //!
-//! .. code-block:: c++
-//!
-//!    #include <cub/cub.cuh>   // or equivalently <cub/block/block_scan.cuh>
-//!
-//!    __global__ void ExampleKernel(...)
-//!    {
-//!        // Specialize BlockScan for a 1D block of 128 threads of type int
-//!        using BlockScan = cub::BlockScan<int, 128>;
-//!
-//!        // Allocate shared memory for BlockScan
-//!        __shared__ typename BlockScan::TempStorage temp_storage;
-//!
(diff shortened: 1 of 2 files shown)


miscco (approved):
This is awesome, I love that it makes product code shorten and also ensures the examples compile

bernhardmgruber:
I am confused. Isn't this why we have dedicated API example unit tests? For example, `cub/test/catch2_test_block_scan_api.cu` contains the tested API examples for `cub::BlockScan`. How are the examples touched in this PR different from other CUB API examples?

gonidelis:
@bernhardmgruber 
>  How are the examples touched in this PR different from other CUB API examples?

Indeed there is not much of a difference. I don't see a reason of why these "example tests" couldn't just be the "API tests". We could move onto that subsequently to avoid having too many tests that *achieve almost the same thing.* I guess maybe the only thing is API tests do not guarantee there is going to be an API test case for **every single overload**. But as soon as you add an overload and you have to provide docs + example then you are forced to have an API test for everything.

Anyways, the purpose of this work remains and is strong: "Documentation examples are not guaranteed to compile when we just hand them out with pseudo-code in the comments sections". Merging them with API tests can follow up.

PR #6289, opened 2025-10-20
Provide `cuda::ptx::enable_smem_spilling()`

davebayer:
In PTX ISA 9.0, the `.pragma "enable_smem_spilling"` pragma was introduced. The current way to enable the register spilling to shared memory (as showed in the [blog post](https://developer.nvidia.com/blog/how-to-improve-cuda-kernel-performance-with-shared-memory-register-spilling/)) is by writing:
```cuda
__global__ void kernel(...)
{
  // ...
  asm volatile (".pragma \"enable_smem_spilling\";");
  // ...
}
```
I think we can provide a more convenient way in our `cuda::ptx::` namespace. I've wrapped the pragma into a function which must be `static` and force inlined.

### DIFF
--- a/.github/actions/workflow-run-job-windows/action.yml
+++ b/.github/actions/workflow-run-job-windows/action.yml
@@ -30,6 +30,8 @@ runs:
       with:
         path: ${{github.event.repository.name}}
         persist-credentials: false
+        fetch-depth: 0
+        fetch-tags: true
 
     - name: Define and log job details
       shell: bash --noprofile --norc -euo pipefail {0}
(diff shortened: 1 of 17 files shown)


fbusato:
would not be better to add this feature as part of the PTX generator? in this way, it could be better supported in the future

davebayer:
> would not be better to add this feature as part of the PTX generator? in this way, it could be better supported in the future

What would be the benefit? It's literally a one liner which I don't expect to ever change

fbusato:
not a big issue, but we have a systematic way to add PTX instructions, which automates the whole process, including docs and tests. This is a simple instruction, but for more complex ones, it makes sense to rely on this tool

PR #6228, opened 2025-10-14
[Backport 3.1] Backport #6184 and #6224

davebayer:

PR #6291, opened 2025-10-20
Fix `string_view` construction from `std::string_view`

davebayer:
Fixes the problem reported in #6228.

The problem is that we instantiated the `std::basic_string_view` even though we don't know whether `std::char_traits` are specialized for that type.

### DIFF
--- a/libcudacxx/include/cuda/std/string_view
+++ b/libcudacxx/include/cuda/std/string_view
@@ -138,6 +138,13 @@ _CCCL_CONCEPT __cccl_basic_sv_compatible_range = _CCCL_REQUIRES_EXPR((_Range, _C
                    || __cccl_basic_sv_compatible_range_check_traits_std_ext<_Range, _CharT, _Traits>) )) //
   );
 
+template <class _Traits, class _HostTraits>
+_CCCL_CONCEPT __cccl_basic_sv_is_std_to_cuda_std_char_traits = _CCCL_REQUIRES_EXPR((_Traits, _HostTraits)) //
+  ( //
+    requires(__is_cuda_std_char_traits_v<_Traits>), //
+    requires(__is_std_char_traits_v<_HostTraits>) //
+  );
+
 template <class _CharT, class _Traits>
 class _CCCL_TYPE_VISIBILITY_DEFAULT basic_string_view
 {
@@ -206,7 +213,7 @@ public:
 
 #if !_CCCL_COMPILER(NVRTC)
   _CCCL_TEMPLATE(class _Traits2, class _Alloc)
-  _CCCL_REQUIRES(__is_cuda_std_char_traits_v<_Traits> _CCCL_AND __is_std_char_traits_v<_Traits2>)
+  _CCCL_REQUIRES(__cccl_basic_sv_is_std_to_cuda_std_char_traits<_Traits, _Traits2>)
   _CCCL_HOST_API constexpr basic_string_view(const ::std::basic_string<_CharT, _Traits2, _Alloc>& __sv) noexcept
       : __data_{__sv.data()}
       , __size_{__sv.size()}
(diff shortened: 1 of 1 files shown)


libcudacxx/include/cuda/std/string_view
```diff
@@ -205,30 +205,34 @@ public:
   }
 
 #if !_CCCL_COMPILER(NVRTC)
-  _CCCL_TEMPLATE(class _Traits2 = _Traits, class _Alloc)
-  _CCCL_REQUIRES(is_same_v<_Traits2, char_traits<_CharT>>)
-  _CCCL_HOST_API constexpr basic_string_view(
-    const ::std::basic_string<_CharT, ::std::char_traits<_CharT>, _Alloc>& __sv) noexcept
+  _CCCL_TEMPLATE(class _CharT2, class _Traits2, class _Alloc)
+  _CCCL_REQUIRES(is_same_v<_CharT, _CharT2> _CCCL_AND is_same_v<_Traits, char_traits<_CharT>> _CCCL_AND
+                   is_same_v<_Traits2, ::std::char_traits<_CharT2>>)
```
miscco: rather than this, I would like to introduce `__is_std_char_traits_v` in `__fwd/char_traits` and use that.

davebayer: Leaving this to another PR

libcudacxx/include/cuda/std/__fwd/char_traits.h
```diff
@@ -55,6 +55,18 @@ template <>
 struct char_traits<wchar_t>;
 #endif // _CCCL_HAS_WCHAR_T()
 
+template <class _Tp>
+inline constexpr bool __is_std_char_traits_v = false;
+#if _CCCL_HAS_HOST_STD_LIB()
+template <class _Tp>
+inline constexpr bool __is_std_char_traits_v<::std::char_traits<_Tp>> = true;
```
Artem-B, line 62: Just a FYI. In my naive backport of this patch to v3.1 branch use of `::std::char_traits` here failed due to the missing inclusion of `<string>`.

```
In file included from third_party/gpus/cccl/v3_1/libcudacxx/test/libcudacxx/std/strings/char.traits/char.traits.specializations/char.traits.specializations.char16_t/types.compile.pass.cpp:10:
In file included from blaze-out/k8-opt/bin/third_party/gpus/cuda/_virtual_includes/_stage/third_party/gpus/cuda/include/cuda/std/__string_:24:
In file included from blaze-out/k8-opt/bin/third_party/gpus/cuda/_virtual_includes/_stage/third_party/gpus/cuda/include/cuda/std/__string/char_traits.h:24:
blaze-out/k8-opt/bin/third_party/gpus/cuda/_virtual_includes/_stage/third_party/gpus/cuda/include/cuda/std/__fwd/char_traits.h:49:46: error: no template named 'char_traits' in namespace 'std'; did you mean simply 'char_traits'?
   49 | inline constexpr bool __is_std_char_traits_v<::std::char_traits<_Tp>> = true;
      |                                              ^~~~~~~
blaze-out/k8-opt/bin/third_party/gpus/cuda/_virtual_includes/_stage/third_party/gpus/cuda/include/cuda/std/__fwd/char_traits.h:29:38: note: 'char_traits' declared here
   29 | struct _CCCL_TYPE_VISIBILITY_DEFAULT char_traits;
      |                                      ^

```

Once `<string>` was included, the patch does appear to fix the original compilation problem I reported in https://github.com/NVIDIA/cccl/pull/6228

davebayer: Yes, it's because we started to forward declare types from `std::` namespace right after the 3.1 release. I will have to fix it in the backport later :)

I am glad to hear that it fixes the problem, thanks for reporting it :) Now I just to need to fight msvc 19.29 to accept the code..

Issue #5679, opened 2025-08-27
[FEA]: Implement new random number generators


RAMitchell:
### Is this a duplicate?

- [x] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cccl/issues) for this request and that I agree to the [Code of Conduct](CODE_OF_CONDUCT.md)

### Area

Thrust

### Is your feature request related to a problem? Please describe.

Thrust currently implements the following random number generators:

[thrust::random::minstd_rand](https://nvidia.github.io/cccl/thrust/api/group__predefined__random_1ga1c89beeeb2681026be374c4e0d6609f3.html)
[thrust::random::ranlux24_base](https://nvidia.github.io/cccl/thrust/api/group__predefined__random_1ga44ec779aead9b140b3fb332ea4c349bc.html)
[thrust::random::ranlux48](https://nvidia.github.io/cccl/thrust/api/group__predefined__random_1ga67a4f4c57c6d60aea077a947285cb7c5.html)
[thrust::random::default_random_engine](https://nvidia.github.io/cccl/thrust/api/group__predefined__random_1ga8450c92f9d27a16992ca377f7d8dc75c.html)
[thrust::random::ranlux24](https://nvidia.github.io/cccl/thrust/api/group__predefined__random_1gaacaf3beafd00b7087532c98d9062778a.html)
[thrust::random::taus88](https://nvidia.github.io/cccl/thrust/api/group__predefined__random_1gac781a23cdf5b0cdb48ec3fbe7cb4318c.html)
[thrust::random::ranlux48_base](https://nvidia.github.io/cccl/thrust/api/group__predefined__random_1gaeb4bf97731c920bdf7cc80d2c2321f2f.html)
[thrust::random::minstd_rand0](https://nvidia.github.io/cccl/thrust/api/group__predefined__random_1gaf8dffb2a8bfaaa20d5c916d74c07dc98.html)

These leave a little to be desired in either performance (either generation or advancing) or robustness in comparison to some recent developments in the field.

### Describe the solution you'd like

I propose adding:
https://en.cppreference.com/w/cpp/numeric/random/philox_engine.html
https://en.wikipedia.org/wiki/Permuted_congruential_generator (PCG64 as used in numpy).

Furthermore, I think documentation could be improved in the following areas:
- Example code of how generators are expected to be used in a parallel environment (e.g. generate/for_each).
[...]

PR #6292, opened 2025-10-20
Implement PCG64 as extension

RAMitchell:
## Description

Part of #5679

Depends on #6109

### DIFF
--- a/libcudacxx/include/cuda/__random/pcg_engine.h
+++ b/libcudacxx/include/cuda/__random/pcg_engine.h
@@ -22,12 +22,13 @@
 #endif // no system header
 
 #include <cuda/std/__bit/rotate.h>
+#include <cuda/std/__limits/numeric_limits.h>
 #include <cuda/std/__random/is_seed_sequence.h>
 #include <cuda/std/__type_traits/enable_if.h>
+#include <cuda/std/__type_traits/integral_constant.h>
+#include <cuda/std/__utility/pair.h>
 #include <cuda/std/array>
 #include <cuda/std/cstdint>
-#include <cuda/std/limits>
-#include <cuda/std/utility>
 
 #include <cuda/std/__cccl/prologue.h>
 
@@ -63,14 +64,15 @@ private:
   static constexpr __uint128_t __multiplier = (static_cast<__uint128_t>(_AHi) << 64) | _ALo;
   static constexpr __uint128_t __increment  = (static_cast<__uint128_t>(_CHi) << 64) | _CLo;
 
-  [[nodiscard]] _CCCL_API constexpr result_type __output_transform(__uint128_t __internal) noexcept
+  [[nodiscard]] _CCCL_API static constexpr result_type __output_transform(__uint128_t __internal) noexcept
   {
-    const int __rot = static_cast<__bitcount_t>(__internal >> 122);
+    const auto __rot = static_cast<__bitcount_t>(__internal >> 122);
     __internal ^= __internal >> 64;
     return ::cuda::std::rotr(result_type(__internal), __rot);
   }
 
-  [[nodiscard]] _CCCL_API constexpr ::cuda::std::pair<__uint128_t, __uint128_t> __power_mod(__uint128_t __delta) noexcept
+  [[nodiscard]] _CCCL_API static constexpr ::cuda::std::pair<__uint128_t, __uint128_t>
+  __power_mod(__uint128_t __delta) noexcept
   {
     __uint128_t __acc_mult = 1;
     __uint128_t __acc_plus = 0;
@@ -177,13 +179,13 @@ public:
     return __x.__x_ == __y.__x_;
   }
 
-#  if _CCCL_STD_VER == 2017
+#  if _CCCL_STD_VER <= 2017
   //! @brief Inequality comparison for two engines.
   [[nodiscard]] _CCCL_API constexpr friend bool operator!=(const pcg64_engine& __x, const pcg64_engine& __y) noexcept
   {
     return !(__x == __y);
   }
-#  endif
+#  endif // _CCCL_STD_VER <= 2017
 
 #  if !_CCCL_COMPILER(NVRTC)
 
@@ -262,7 +264,7 @@ public:
 //!
 //! @note This class requires compiler support for 128-bit integers.
 using pcg64 =
-  pcg64_engine<2549297995355413924ULL, 4865540595714422341ULL, 6364136223846793005ULL, 1442695040888963407ULL>;
+  pcg64_engine<2549297995355413924ull, 4865540595714422341ull, 6364136223846793005ull, 1442695040888963407ull>;
 
 #endif // _CCCL_HAS_INT128()
 _CCCL_END_NAMESPACE_CUDA


libcudacxx/include/cuda/__random/pcg_engine.h
```diff
@@ -0,0 +1,251 @@
...
+
+    // restore flags
+    __is.flags(__flags);
+
+    return __is;
+  }
+#endif // !_CCCL_COMPILER(NVRTC)
+
+private:
+  using __uint128_type = unsigned __int128;
```
davebayer: Just use `__uint128_t`

libcudacxx/include/cuda/__random/pcg_engine.h
```diff
@@ -0,0 +1,251 @@
...
+  using __bitcount_t   = ::cuda::std::uint8_t;
+
+  static constexpr __uint128_type __multiplier =
+    ((__uint128_type) 2549297995355413924ULL << 64) | 4865540595714422341ULL;
+  static constexpr __uint128_type __increment =
+    ((__uint128_type) 6364136223846793005ULL << 64) | 1442695040888963407ULL;
+  [[nodiscard]] _CCCL_API constexpr result_type __rotr(result_type __value, __bitcount_t __rot) noexcept
```
davebayer: you can use `cuda::std::rotr` instead

libcudacxx/include/cuda/__random/pcg_engine.h
```diff
@@ -0,0 +1,251 @@
...
+  {
+    __bitcount_t __rot = __bitcount_t(__internal >> 122);
+    __internal ^= __internal >> 64;
+    return __rotr(result_type(__internal), __rot);
+  }
+
+  [[nodiscard]] _CCCL_API constexpr auto __power_mod(__uint128_type __delta) noexcept
+  {
+    constexpr __uint128_type __ZERO = 0u;
+    constexpr __uint128_type __ONE  = 1u;
```
davebayer: I would get rid of these completely

libcudacxx/include/cuda/__random/pcg_engine.h
```diff
@@ -0,0 +1,251 @@
...
+
+    // save old flags & fill character
+    const typename ios_base::fmtflags __flags = __os.flags();
+    const _CharT __fill                       = __os.fill();
+
+    __os.flags(ios_base::dec | ios_base::fixed | ios_base::left);
+    __os.fill(__os.widen(' '));
+    // Write 64 bits at a time
+    ::cuda::std::uint64_t low = static_cast<::cuda::std::uint64_t>(__e.__x_);
+    ::cuda::std::uint64_t hi  = static_cast<::cuda::std::uint64_t>(__e.__x_ >> 64);
```
davebayer:
```suggestion
    ::cuda::std::uint64_t __low = static_cast<::cuda::std::uint64_t>(__e.__x_);
    ::cuda::std::uint64_t __hi  = static_cast<::cuda::std::uint64_t>(__e.__x_ >> 64);
```

libcudacxx/include/cuda/__random/pcg_engine.h
RAMitchell: I still need to implement a fallback int128.

davebayer: Do you need this on windows right now? I think it would be easier to guard it by ` #if _CCCL_HAS_INT128()` for now and allow this feature on windows once we have our  internal `int128` fallback

RAMitchell: From slack it sounded like we wanted to have a complete implementation. I don't think its too hard so I think we can do it here.

davebayer: Please, do it in a separate PR

RAMitchell:
I am a bit unsure about the naming yet and if we would support variants of this engine. 

I'm thinking as follows:

pcg64_engine - PCG XSL RR 128/64 algorithm templated on the multiplier/increment
pcg64 - named instance of the above engine with default multiplier/increment
pcg64_discard - a version of pcg64 that can discard a compile time or runtime number of values with each operator(), allowing higher performing "leapfrogging" PRNG in GPU kernels.

@fbusato maybe you have some thoughts?

libcudacxx/include/cuda/__random/pcg_engine.h
```diff
@@ -0,0 +1,243 @@
...
+  }
+
+  /// @brief Seed the engine from a SeedSequence-like object.
+  /// @tparam _Sseq A SeedSequence-like type providing entropy words.
+  /// @param __seq A SeedSequence-like object providing 128 bits of entropy.
+  _CCCL_TEMPLATE(class _Sseq)
+  _CCCL_REQUIRES(::cuda::std::__is_seed_sequence<_Sseq, pcg64_engine>)
+  constexpr _CCCL_API void seed(_Sseq& __seq)
+  {
+    ::cuda::std::array<uint32_t, 4> data = {};
```
fbusato:
```suggestion
    ::cuda::std::array<::cuda::std::uint32_t, 4> data = {};
```

libcudacxx/include/cuda/__random/pcg_engine.h
```diff
@@ -0,0 +1,243 @@
...
+  {
+    __x_ = __x_ * __multiplier + __increment;
+    return __output_transform(__x_);
+  }
+
+  /// @brief Advance the engine state by `__z` steps, discarding outputs.
+  /// @param __z Number of values to discard.
+  constexpr _CCCL_API void discard(unsigned long long __z) noexcept
+  {
+    auto [__mult, __plus] = __power_mod(__z);
```
fbusato:
```suggestion
    const auto [__mult, __plus] = __power_mod(__z);
```

libcudacxx/include/cuda/__random/pcg_engine.h
```diff
@@ -0,0 +1,243 @@
...
+  }
+
+  /// @brief Equality comparison for two engines.
+  /// @return True if both engines have identical internal state.
+  [[nodiscard]] _CCCL_API constexpr friend bool operator==(const pcg64_engine& __x, const pcg64_engine& __y) noexcept
+  {
+    return __x.__x_ == __y.__x_;
+  }
+  /// @brief Inequality comparison for two engines.
+  [[nodiscard]] _CCCL_API constexpr friend bool operator!=(const pcg64_engine& __x, const pcg64_engine& __y) noexcept
```
fbusato, line 184: automatically generated in C++20

libcudacxx/include/cuda/__random/pcg_engine.h
```diff
@@ -0,0 +1,243 @@
...
+  {
+    return !(__x == __y);
+  }
+#  if !_CCCL_COMPILER(NVRTC)
```
fbusato, line 190: nit
```suggestion

#  if !_CCCL_COMPILER(NVRTC)

```

libcudacxx/include/cuda/__random/pcg_engine.h
```diff
@@ -0,0 +1,243 @@
...
+    __is.flags(__flags);
+
+    return __is;
+  }
+#  endif // !_CCCL_COMPILER(NVRTC)
+
+private:
+  using __bitcount_t = ::cuda::std::uint8_t;
+
+  static constexpr __uint128_t __multiplier = ((__uint128_t) 2549297995355413924ULL << 64) | 4865540595714422341ULL;
```
fbusato:
```suggestion
  static constexpr __uint128_t __multiplier = (__uint128_t{2549297995355413924ULL} << 64) | 4865540595714422341ULL;
```

libcudacxx/include/cuda/__random/pcg_engine.h
```diff
@@ -0,0 +1,243 @@
...
+  static constexpr __uint128_t __increment  = ((__uint128_t) 6364136223846793005ULL << 64) | 1442695040888963407ULL;
+  [[nodiscard]] _CCCL_API constexpr result_type __output_transform(__uint128_t __internal) noexcept
+  {
+    __bitcount_t __rot = __bitcount_t(__internal >> 122);
```
fbusato: `__rot` should be `int` as for `std::rotr` input
```suggestion
    const int __rot = static_cast<__bitcount_t>(__internal >> 122);
```

libcudacxx/test/libcudacxx/cuda/random/pcg64.pass.cpp
```diff
@@ -0,0 +1,52 @@
...
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES.
+//
+//===----------------------------------------------------------------------===//
+
+// <random>
+
+#include <cuda/__random/pcg_engine.h>
+
+#include "../../std/random/engine/test_engine.h"
```
fbusato: prefer absolute paths

libcudacxx/test/libcudacxx/libcxx/random/pcg_engine.pass.cpp
```diff
@@ -0,0 +1,52 @@
...
+// <random>
+
+#include <cuda/__random/pcg_engine.h>
+
+#include "../../std/random/engine/test_engine.h"
+
+__host__ __device__ constexpr bool test_against_reference()
+{
+  // reference values obtained from other library implementations
+  const int seeds[]                            = {10823018, 0, 23};
```
fbusato:
```suggestion
  constexpr int seeds[]                            = {10823018, 0, 23};
```
or `static constexpr`

fbusato: same below

libcudacxx/include/cuda/__random/pcg_engine.h
davebayer: This is not correct, you must cast the value to `__uint128_t` first and then do the shift

libcudacxx/test/libcudacxx/cuda/random/pcg64.pass.cpp
davebayer: Btw this test should go to `libcudacxx/test/libcudacxx/cuda/random`

libcudacxx/include/cuda/__random/pcg_engine.h
fbusato: right, wrong parenthesis. Fixed

fbusato:
> `pcg64_engine` - PCG XSL RR 128/64 algorithm templated on the multiplier/increment
`pcg64` - named instance of the above engine with default multiplier/increment
`pcg64_discard` - a version of pcg64 that can discard a compile time or runtime number of values with each operator(), allowing higher performing "leapfrogging" PRNG in GPU kernels.

these names are perfectly fine. I'm thinking if there are better names for `pcg64_discard` but I'm not sure

libcudacxx/test/libcudacxx/cuda/random/pcg64.pass.cpp
RAMitchell: This test helper would need to be somewhere else to use absolute paths. Let me know if I should move it.

libcudacxx/include/cuda/__random/pcg_engine.h
```diff
@@ -0,0 +1,267 @@
...
+///
+/// Most users should use the predefined `pcg64` type alias instead of this class directly.
+///
+/// @tparam __a The multiplier constant for the LCG.
+/// @tparam __c The increment constant for the LCG.
+///
+/// @note This class requires compiler support for 128-bit integers.
+///
+/// @see https://www.pcg-random.org/ for details on the PCG family of generators.
+template <__uint128_t __a, __uint128_t __c>
```
davebayer: In case we implement our own `cuda::std::uint128_t`, these arguments will not be usable in C++20, because passing `struct` as an NTTP is not supported in C++17. We should come up with a way to provide this even in C++17.

libcudacxx/include/cuda/__random/pcg_engine.h
```diff
@@ -0,0 +1,267 @@
...
+  /// @brief Returns the largest value the engine can produce.
+  /// @return The maximum representable `result_type`.
+  [[nodiscard]] _CCCL_API static constexpr result_type max() noexcept
+  {
+    return ::cuda::std::numeric_limits<result_type>::max();
+  }
+
+  // constructors and seeding functions
+  /// @brief Default-constructs the engine using `default_seed`.
+  constexpr _CCCL_API pcg64_engine() noexcept
```
davebayer:
```suggestion
  _CCCL_API constexpr pcg64_engine() noexcept
```
applies to other places, too

davebayer (changes requested):

libcudacxx/include/cuda/__random/pcg_engine.h
RAMitchell: Maybe it should be 4 64-bit ints and internally we create a static cuda::std::uint128_t member?

RAMitchell:
@davebayer @fbusato please re-review thanks :)

libcudacxx/include/cuda/__random/pcg_engine.h
```diff
@@ -0,0 +1,269 @@
...
+  {
+    return ::cuda::std::numeric_limits<result_type>::max();
+  }
+
+  // constructors and seeding functions
+  /// @brief Default-constructs the engine using `default_seed`.
+  _CCCL_API constexpr pcg64_engine() noexcept
+      : pcg64_engine(default_seed)
+  {}
+  /// @brief Constructs the engine and seeds it with `__seed`.
```
miscco: Nitpick: We use `//!` for doxygen comments

libcudacxx/include/cuda/__random/pcg_engine.h
```diff
@@ -0,0 +1,269 @@
...
+  /// @param __seed The seed value used to initialize the engine state.
+  _CCCL_API constexpr explicit pcg64_engine(result_type __seed) noexcept
+  {
+    seed(__seed);
+  }
+
+  _CCCL_TEMPLATE(class _Sseq)
+  /// @brief Constructs the engine and seeds it from a SeedSequence-like object.
+  /// @tparam _Sseq A SeedSequence-like type satisfying the project's seed concept.
+  /// @param __seq The seed sequence used to initialize the internal state.
```
miscco: Important: The docs come before the template header
```suggestion
  /// @brief Constructs the engine and seeds it from a SeedSequence-like object.
  /// @tparam _Sseq A SeedSequence-like type satisfying the project's seed concept.
  /// @param __seq The seed sequence used to initialize the internal state.
  _CCCL_TEMPLATE(class _Sseq)
```

libcudacxx/include/cuda/__random/pcg_engine.h
```diff
@@ -0,0 +1,269 @@
...
+    return __is;
+  }
+#  endif // !_CCCL_COMPILER(NVRTC)
+
+private:
+  using __bitcount_t = ::cuda::std::uint8_t;
+
+  static constexpr __uint128_t __multiplier = (static_cast<__uint128_t>(_AHi) << 64) | _ALo;
+  static constexpr __uint128_t __increment  = (static_cast<__uint128_t>(_CHi) << 64) | _CLo;
+  [[nodiscard]] _CCCL_API constexpr result_type __output_transform(__uint128_t __internal) noexcept
```
miscco: Readability:
```suggestion

  [[nodiscard]] _CCCL_API constexpr result_type __output_transform(__uint128_t __internal) noexcept
```

libcudacxx/include/cuda/__random/pcg_engine.h
```diff
@@ -0,0 +1,269 @@
...
+    // Read engine state from stream: low 64 bits then high 64 bits.
+    __e.__x_ = (static_cast<__uint128_t>(__hi) << 64) | __low;
+    // restore flags
+    __is.flags(__flags);
+
+    return __is;
+  }
+#  endif // !_CCCL_COMPILER(NVRTC)
+
+private:
```
miscco: Style: 

When possible I prefer to move private parts of the implementation to the top. 

I many cases the members and type aliases are important to understand the implementations, so scrolling back and forth is unfortunate

libcudacxx/test/libcudacxx/cuda/random/pcg64.pass.cpp
miscco: I believe we should move it into the `libcudacxx/test/support` folder and then included via `"test_engine.h"` as we do with other test headers

libcudacxx/include/cuda/__random/pcg_engine.h
```diff
@@ -0,0 +1,272 @@
...
+public:
+  using result_type = ::cuda::std::uint64_t;
+
+private:
+  using __bitcount_t = ::cuda::std::uint8_t;
+
+  static constexpr __uint128_t __multiplier = (static_cast<__uint128_t>(_AHi) << 64) | _ALo;
+  static constexpr __uint128_t __increment  = (static_cast<__uint128_t>(_CHi) << 64) | _CLo;
+
+  [[nodiscard]] _CCCL_API constexpr result_type __output_transform(__uint128_t __internal) noexcept
```
fbusato: seems a `static` function

libcudacxx/include/cuda/__random/pcg_engine.h
```diff
@@ -0,0 +1,272 @@
...
+  {
+    const int __rot = static_cast<__bitcount_t>(__internal >> 122);
```
fbusato:
```suggestion
    const auto __rot = static_cast<__bitcount_t>(__internal >> 122);
```

libcudacxx/include/cuda/__random/pcg_engine.h
```diff
@@ -0,0 +1,272 @@
...
+    __internal ^= __internal >> 64;
+    return ::cuda::std::rotr(result_type(__internal), __rot);
+  }
+
+  [[nodiscard]] _CCCL_API constexpr ::cuda::std::pair<__uint128_t, __uint128_t> __power_mod(__uint128_t __delta) noexcept
```
fbusato: this also seems `static`

libcudacxx/include/cuda/__random/pcg_engine.h
```diff
@@ -0,0 +1,272 @@
...
+//! @note This class requires compiler support for 128-bit integers.
+//!
+//! @see https://www.pcg-random.org/ for details on the PCG family of generators.
+template <::cuda::std::uint64_t _AHi, ::cuda::std::uint64_t _ALo, ::cuda::std::uint64_t _CHi, ::cuda::std::uint64_t _CLo>
+class pcg64_engine
+{
+public:
+  using result_type = ::cuda::std::uint64_t;
+
+private:
```
fbusato, line 61: would be better to keep private members at the end of the class

libcudacxx/include/cuda/__random/pcg_engine.h
```diff
@@ -0,0 +1,272 @@
...
+      __cur_plus = (__cur_mult + 1) * __cur_plus;
+      __cur_mult *= __cur_mult;
+      __delta >>= 1;
+    }
+    return ::cuda::std::pair{__acc_mult, __acc_plus};
+  }
+  __uint128_t __x_{};
+
+public:
+  static constexpr result_type default_seed = 0xcafef00dd15ea5e5ULL;
```
fbusato, line 97:
```suggestion
  static constexpr result_type default_seed = 0xCAFEF00DD15EA5E5ull;
```

libcudacxx/include/cuda/__random/pcg_engine.h
```diff
@@ -0,0 +1,272 @@
...
+  }
+
+  //! @brief Equality comparison for two engines.
+  //! @return True if both engines have identical internal state.
+  [[nodiscard]] _CCCL_API constexpr friend bool operator==(const pcg64_engine& __x, const pcg64_engine& __y) noexcept
+  {
+    return __x.__x_ == __y.__x_;
+  }
+
+#  if _CCCL_STD_VER == 2017
```
fbusato:
```suggestion
#  if _CCCL_STD_VER <= 2017
```
for consistency

libcudacxx/include/cuda/__random/pcg_engine.h
```diff
@@ -0,0 +1,272 @@
...
+  //! @brief Inequality comparison for two engines.
+  [[nodiscard]] _CCCL_API constexpr friend bool operator!=(const pcg64_engine& __x, const pcg64_engine& __y) noexcept
+  {
+    return !(__x == __y);
+  }
+#  endif
```
fbusato:
```suggestion
#  endif // _CCCL_STD_VER == 2017
```

libcudacxx/include/cuda/__random/pcg_engine.h
```diff
@@ -0,0 +1,272 @@
...
+//!
+//!   cuda::pcg64 eng;                 // default seed
+//!   uint64_t v = eng();                     // draw value
+//!   eng.seed(42);                           // reseed
+//!   eng.discard(10);                        // skip 10 outputs
+//! @endcode
+//!
+//! @note This class requires compiler support for 128-bit integers.
+using pcg64 =
+  pcg64_engine<2549297995355413924ULL, 4865540595714422341ULL, 6364136223846793005ULL, 1442695040888963407ULL>;
```
fbusato:
```suggestion
  pcg64_engine<2549297995355413924ull, 4865540595714422341ull, 6364136223846793005ull, 1442695040888963407ull>;
```

libcudacxx/include/cuda/__random/pcg_engine.h
```diff
@@ -0,0 +1,272 @@
...
+#  pragma system_header
+#endif // no system header
+
+#include <cuda/std/__bit/rotate.h>
+#include <cuda/std/__random/is_seed_sequence.h>
+#include <cuda/std/__type_traits/enable_if.h>
+#include <cuda/std/array>
+#include <cuda/std/cstdint>
+#include <cuda/std/limits>
+#include <cuda/std/utility>
```
fbusato: probably we can just use `pair.h` header

libcudacxx/include/cuda/__random/pcg_engine.h
```diff
@@ -0,0 +1,272 @@
...
+
+  //! @brief Generate the next pseudo-random value.
+  //!
+  //! Advances the internal LCG state and applies the PCG output
+  //! permutation to produce a 64-bit result.
+  //! @return A 64-bit pseudo-random value.
+  _CCCL_API constexpr result_type operator()() noexcept
+  {
+    __x_ = __x_ * __multiplier + __increment;
+    return __output_transform(__x_);
```
fbusato, line 164: important. The function should return the previous state

fbusato (changes requested):

libcudacxx/include/cuda/__random/pcg_engine.h
fbusato: in addition, the unit test didn't captured this problem

libcudacxx/include/cuda/__random/pcg_engine.h
miscco: Actually not the general advice is to move them to the front, because they are crucial to understand what is in the class

libcudacxx/include/cuda/__random/pcg_engine.h
miscco: yeah for internal headers we always use the smallest available one, here this would be `cuda/std/__utiltiy/pair.h`

libcudacxx/include/cuda/__random/pcg_engine.h
davebayer: why? Is this a rule?

libcudacxx/include/cuda/__random/pcg_engine.h
RAMitchell: I normally belong to the "end of class" church, but we can do whatever here :)

miscco: The point being that usually a class implementation starts with constructors and other SMF, so I need to know what are the actual data members. In that case and others I have to jump around to the back of the potentially long definition to know what I am working with

libcudacxx/include/cuda/__random/pcg_engine.h
RAMitchell: Why? Isn't this an implementation detail? The output sequence is correct against numpy and the PCG64 reference implementation.

We could increment the state by 1 after seeding and then change the order of operator() but I don't think it changes anything.

libcudacxx/include/cuda/__random/pcg_engine.h
davebayer: Hm, I don't know. Thinking about this a bit more, there will only be problems with msvc an nvrtc on windows in C++17, all other compilers will be fine with the code. How much do we care about those?

libcudacxx/include/cuda/__random/pcg_engine.h
miscco: yeah lets do this as a followup

libcudacxx/include/cuda/__random/pcg_engine.h
RAMitchell: I think its fine. FYI I have a follow up PR with a minimal __pcg64_uint128_fallback in about 100 lines of code. With this we support everything, no exceptions.

libcudacxx/include/cuda/__random/pcg_engine.h
fbusato: definitely not a rule, but suggested by some c++ secure coding guidelines, e.g. Autosar C++14 Rule A2-13-5 https://www.autosar.org/fileadmin/standards/R18-03_R1.4.0/AP/AUTOSAR_RS_CPP14Guidelines.pdf. `ull` lowercase to have a clear distinction with uppercase digits

libcudacxx/include/cuda/__random/pcg_engine.h
fbusato: I check the original implementation. You are right. I got confused because other generators return the previous state

libcudacxx/include/cuda/__random/pcg_engine.h
fbusato: interesting, I always thought about user perspective. Users are interested to the interface, not implementation details. Also, the implementation rarely changes.
Anyway, I'm fine with both approaches.

davebayer (approved):

RAMitchell:
@fbusato waiting on you :)

PR #6296, opened 2025-10-20
Do not include API references into toc tree

miscco:
This reduces our docs size by about 200MB

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -50,6 +50,7 @@ extensions = [
     "nbsphinx",
     # "rst_processor",  # Disabled - breathe handles embed:rst natively
     "auto_api_generator",  # Automatically generate API reference pages from Doxygen XML
+    "sphinx_remove_toctrees",
 ]
 
 # Breathe configuration for Doxygen integration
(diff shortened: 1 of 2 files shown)


ericniebler (approved):

miscco:
pre-commit.ci autofix

oleksandr-pavlyk:
Note that while the size is indeed reduced, a fair amount of documentation becomes nondiscoverable (for better or for worse). For example, `cub::DeviceTransform` page is still generated and available, but no longer discoverable from the table of content (TOC) sidebar.

image

Perhaps a fair amount of space can be saved by pruning those pages too. 

Or leave them there and reference them in the TOC.

bernhardmgruber:
> For example, `cub::DeviceTransform` page is still generated and available, but no longer discoverable from the table of content (TOC) sidebar.

I think it's fine as long as we can reach the page through other pages.

> Perhaps a fair amount of space can be saved by pruning those pages too.

I think we would want a larger reorganization of the docs, but we definitely don't want to drop any of the documentations of each overload a user can call. We could group them better, or differently though.

miscco:
Closing because it hurts more than it offers

PR #6299, opened 2025-10-20
[cudax -> libcudacxx] Move type-erased resource wrappers to libcudacxx

pciolkosz:
This PR moves the experimental implementation of `any_resource` and `resource_ref` (with synchronous variants) to libcu++. This new implementation replaces the previous `resource_ref` implementation.

This is mostly just copying the `any_resource.cuh` file, but I had to make some adjustments:
- I had to remove the compatibility parts with the old `resource_ref` implementation, since that header is now gone
- I had to add `has_property` requirement on `__get_property` function, otherwise the reference type was wrongly reported to be constructible from a resource with wrong properties (even while actual construction failed to compile)
- In the tests I removed the bits testing construction from a pointer, it seemed not well compatible with the new implementation of resources and also fairly low value

For now I left aliases in the experimental namespace and experimental tests where they are. I will remove those aliases and move the tests in a separate PR

### DIFF
--- a/c/parallel/include/cccl/c/radix_sort.h
+++ b/c/parallel/include/cccl/c/radix_sort.h
@@ -41,6 +41,7 @@ typedef struct cccl_device_radix_sort_build_result_t
   CUkernel exclusive_sum_kernel;
   CUkernel onesweep_kernel;
   cccl_sort_order_t order;
+  void* runtime_policy;
 } cccl_device_radix_sort_build_result_t;
 
 CCCL_C_API CUresult cccl_device_radix_sort_build(
--- a/c/parallel/src/radix_sort.cu
+++ b/c/parallel/src/radix_sort.cu
@@ -20,6 +20,7 @@
 #include "kernels/operators.h"
 #include "util/context.h"
 #include "util/indirect_arg.h"
+#include "util/runtime_policy.h"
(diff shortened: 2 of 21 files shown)


libcudacxx/include/cuda/__memory_resource/any_resource.h
```diff
@@ -0,0 +1,877 @@
...
+// Part of the CUDA C++ Core Libraries,
+// under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef _CUDA__MEMORY_RESOURCE_ANY_RESOURCE_H
+#define _CUDA__MEMORY_RESOURCE_ANY_RESOURCE_H
```
davebayer:
```suggestion
#ifndef _CUDA___MEMORY_RESOURCE_ANY_RESOURCE_H
#define _CUDA___MEMORY_RESOURCE_ANY_RESOURCE_H
```
detail

PR #6304, opened 2025-10-21
Ignore OOM failures for large size unique thrust test.

alliepiper:


thrust/testing/cuda/unique.cu
```diff
@@ -414,12 +414,18 @@ void TestUniqueWithMagnitude(int magnitude)
...
 
 void TestUniqueWithLargeNumberOfItems()
+try
 {
   for (int mag : {30, 31, 32, 33})
   {
     TestUniqueWithMagnitude(mag);
   }
 }
+catch (std::bad_alloc&)
```
bernhardmgruber, line 424:
```suggestion
catch (const std::bad_alloc&)
```

PR #6307, opened 2025-10-21
Fix `is_sufficiently_aligned` with `const void*`

fbusato:
## Description

`alignof(_ElementType)` fails when `_ElementType* == const void` or (just `void`)

### DIFF
--- a/libcudacxx/include/cuda/std/__memory/is_sufficiently_aligned.h
+++ b/libcudacxx/include/cuda/std/__memory/is_sufficiently_aligned.h
@@ -23,9 +23,9 @@
 #endif // no system header
 
 #include <cuda/__cmath/pow2.h>
+#include <cuda/std/__type_traits/is_void.h>
 #include <cuda/std/cstddef> // size_t
 #include <cuda/std/cstdint> // uintptr_t
-#include <cuda/std/type_traits> // is_void_v
 
 #include <cuda/std/__cccl/prologue.h>
 


libcudacxx/include/cuda/std/__memory/is_sufficiently_aligned.h
```diff
@@ -33,9 +34,13 @@ _CCCL_BEGIN_NAMESPACE_CUDA_STD
 template <size_t _ByteAlignment, class _ElementType>
 [[nodiscard]] _CCCL_API inline bool is_sufficiently_aligned(_ElementType* __ptr) noexcept
 {
+  using ::cuda::std::uintptr_t;
   static_assert(::cuda::is_power_of_two(_ByteAlignment), "alignment must be a power of two");
-  static_assert(_ByteAlignment % alignof(_ElementType) == 0,
-                "the alignment must be a multiple of the element alignment");
+  if constexpr (!::cuda::std::is_void_v<_ElementType>)
+  {
+    static_assert(_ByteAlignment % alignof(_ElementType) == 0,
```
pciolkosz, line 41: ~Why not put `is_void_v` as part of the assert, I think it would be easier to parse than the if breaking up the conditions into two parts~

Edit: NVM, I realized `alignof` fails to compile in that case

pciolkosz (approved):

libcudacxx/include/cuda/std/__memory/is_sufficiently_aligned.h
```diff
@@ -25,6 +25,7 @@
 #include <cuda/__cmath/pow2.h>
 #include <cuda/std/cstddef> // size_t
 #include <cuda/std/cstdint> // uintptr_t
+#include <cuda/std/type_traits> // is_void_v
```
pciolkosz: Should this be just `is_void.h`?

fbusato: you are right

wmaxey:
Can we get a test for some degenerate cases? I won't block for that. I disabled auto-merge in case you want to address this.

bernhardmgruber (commented):
Let's add tests for such issues in the future when we propose fixes.

PR #6310, opened 2025-10-22
Always include `<new>` when we need `operator new` for clang-cuda

### LINKED ISSUE #6309: [BUG]: clang requires '#include <new>' in order to use `operator new` in GPU code
### Is this a duplicate?

### Type of Bug

Compile-time Error

### Component

libcu++

### Describe the bug

Compilation of some libcu++ headers fails with clang due to the headers using `operator new` without `#include <new>`.

Clang provides `__device__` variants of the allocator/deallocator functions via the wrapper headers for  `#include <new>`, and if `<new>` is not included, `__device__ operator new(...)` is not available during compilation, which leads to compilation failures.

E.g.:
```
In file included from third_party/gpus/cccl/v3_1/libcudacxx/test/libcudacxx/cuda/pipeline/pipeline_memcpy_async_producer_consumer.pass.cpp:13:
In file included from blaze-out/k8-opt/bin/third_party/gpus/cuda/_virtual_includes/_stage/third_party/gpus/cuda/include/cuda/pipeline:27:
In file included from blaze-out/k8-opt/bin/third_party/gpus/cuda/_virtual_includes/_stage/third_party/gpus/cuda/include/cuda/__memcpy_async/memcpy_async_barrier.h:27:
blaze-out/k8-opt/bin/third_party/gpus/cuda/_virtual_includes/_stage/third_party/gpus/cuda/include/cuda/__barrier/barrier_block_scope.h:118:18: error: reference to __host__ function 'operator new' in __host__ __device__ function
  118 |         } else { new (&__b->__barrier) __barrier_base(__expected); }),
      |                  ^

```

Currently libcu++ includes `new` from `cuda/std/__new/allocate.h` only if `_LIBCUDACXX_HAS_ALIGNED_ALLOCATION()` is true.

https://github.com/NVIDIA/cccl/blob/2d263e1cdd970e336c2e83429405e24de6ac510e/libcudacxx/include/cuda/std/__new/allocate.h#L27-L29

### How to Reproduce

Compile libcu++ tests that include .../__barrier/...` headers  with clang.

### Expected behavior

Headers should be compileable.

### Reproduction link

_No response_

### Operating System

_No response_

### nvidia-smi output

_No response_

### NVCC version

_No response_

miscco:
Fixes #6309

### DIFF
--- a/libcudacxx/include/cuda/std/__new/allocate.h
+++ b/libcudacxx/include/cuda/std/__new/allocate.h
@@ -28,6 +28,11 @@
 #  include <new> // for align_val_t
 #endif // _LIBCUDACXX_HAS_ALIGNED_ALLOCATION() !_CCCL_COMPILER(NVRTC)
 
+// clang-cuda only provides device flavors of operator new, so we need to pull in <new> here
+#if _CCCL_CUDA_COMPILER(CLANG)
+#  include <new>
+#endif // _CCCL_CUDA_COMPILER(CLANG)
+
 #if !defined(__cpp_sized_deallocation) || __cpp_sized_deallocation < 201309L
 #  define _LIBCUDACXX_HAS_SIZED_DEALLOCATION() 0
 #else
--- a/cub/cub/block/block_load.cuh
+++ b/cub/cub/block/block_load.cuh
@@ -46,6 +46,8 @@
 #include <cub/util_ptx.cuh>
 #include <cub/util_type.cuh>
 
+#include <cuda/std/__new/device_new.h>
+
 CUB_NAMESPACE_BEGIN
 
 //! @name Blocked arrangement I/O (direct)
--- a/cub/cub/detail/uninitialized_copy.cuh
+++ b/cub/cub/detail/uninitialized_copy.cuh
@@ -37,6 +37,7 @@
 #  pragma system_header
 #endif // no system header
 
+#include <cuda/std/__new/device_new.h>
 #include <cuda/std/__type_traits/enable_if.h>
 #include <cuda/std/__type_traits/is_trivially_copyable.h>
 #include <cuda/std/__utility/forward.h>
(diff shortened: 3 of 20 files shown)


miscco:
I believe you are right

miscco:
> Don't we want to use `cuda::std::__construct_at` in many of these places?

Thinking about it a bit more I believe we do not want this.

The problem is that `__construct_at` forces us to pull in all of `<memory>` to get the blessed `std::construct_at`

So this is a much heavier include than just `<new>` and also necessary for all compilers and not just for `clang-cuda`

oleksandr-pavlyk:
The #6309 reporter encountered the issue compiling our test file: `libcudacxx/cuda/pipeline/pipeline_memcpy_async_producer_consumer.pass.cpp`

I wonder why this issue was not detected by our CI. We are not using the compiler he was using?

libcudacxx/include/cuda/std/__new/allocate.h
```diff
@@ -28,6 +28,11 @@
 #  include <new> // for align_val_t
 #endif // _LIBCUDACXX_HAS_ALIGNED_ALLOCATION() !_CCCL_COMPILER(NVRTC)
 
+// clang-cuda only provides device flavors of operator new, so we need to pull in <new> here
+#if _CCCL_CUDA_COMPILER(CLANG)
+#  include <new>
+#endif // _CCCL_CUDA_COMPILER(CLANG)
```
bernhardmgruber, line 34: Q: Should we include `<cuda/std/__new/device_new.h>` here instead?

miscco: urgh I thought I did that

Artem-B:
> The #6309 reporter encountered the issue compiling our test file: `libcudacxx/cuda/pipeline/pipeline_memcpy_async_producer_consumer.pass.cpp`
> 
> I wonder why this issue was not detected by our CI. We are not using the compiler he was using?

These are the major differences between our builds that account for most of the issues I've found:
- our build uses clang built from the sources close to LLVM/clang HEAD, while CCCL builds with released versions.
- we use libc++ (also from recent sources), which may expose/introduce different issues not tested by CCCL.

Some of the issues I report are the issues that would eventually be discovered by CCCL CI when it would update clang to the next official release.

Issue #6308, opened 2025-10-21
[BUG]: some thrust headers need include file hints


Artem-B:
### Is this a duplicate?

- [x] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cccl/issues) for this bug and that I agree to the [Code of Conduct](CODE_OF_CONDUCT.md)

### Type of Bug

Compile-time Error

### Component

Thrust

### Describe the bug

A handful of thrust include files that rely on including a header defined by preprocessor macro need hints to help the build system find the actual header file names.

Affected headers:
- https://github.com/NVIDIA/cccl/blob/2d263e1cdd970e336c2e83429405e24de6ac510e/thrust/thrust/execution_policy.h#L26
- https://github.com/NVIDIA/cccl/blob/2d263e1cdd970e336c2e83429405e24de6ac510e/thrust/thrust/iterator/detail/host_system_tag.h#L31
- https://github.com/NVIDIA/cccl/blob/2d263e1cdd970e336c2e83429405e24de6ac510e/thrust/thrust/mr/host_memory_resource.h#L31
- https://github.com/NVIDIA/cccl/blob/2d263e1cdd970e336c2e83429405e24de6ac510e/thrust/thrust/universal_allocator.h#L36
- https://github.com/NVIDIA/cccl/blob/2d263e1cdd970e336c2e83429405e24de6ac510e/thrust/thrust/universal_vector.h#L37

All these files need include hints like these:
https://github.com/NVIDIA/cccl/blob/2d263e1cdd970e336c2e83429405e24de6ac510e/thrust/thrust/system/detail/adl/scan.h#L35-L43

### How to Reproduce

Custom build with bazel, but the issue may potentially surface with SCons and other build systems that scan source files for used headers.

### Expected behavior

It would help if the headers would be more friendly to the build systems that scan sources for used headers.

PR #6311, opened 2025-10-22
Fix thrust system dependend includes

miscco:
Some build systems have issues with our preprocessor based includes, so give them a hint.

Fixes [BUG]: some thrust headers need include file hints #6308

### DIFF
--- a/thrust/thrust/mr/host_memory_resource.h
+++ b/thrust/thrust/mr/host_memory_resource.h
@@ -26,6 +26,13 @@
 #  pragma system_header
 #endif // no system header
 
+// Some build systems need a hint to know which files we actually include
+#if 0
+#  include <thrust/system/cpp/memory_resource.h>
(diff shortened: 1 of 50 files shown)


davebayer (approved):

thrust/thrust/mr/host_memory_resource.h
```diff
@@ -26,6 +26,13 @@
 #  pragma system_header
 #endif // no system header
 
+// Some build systems need a hint to know which files we actually include
+#if 0
+#  include <thrust/system/cpp/memory_resource.h>
+#  include <thrust/system/omp/memory_resource.h>
+#  include <thrust/system/tbb/memory_resource.h>
+#endif
```
bernhardmgruber, line 34: Q: Isn't this missing the cuda header?

miscco: No because that file only uses the `HOST` backends

PR #6313, opened 2025-10-22
Implement interact with legacy library cccl-rt example

### LINKED ISSUE #5715: [FEA]: Example Mixing CUDA Old Runtime and C++ Runtime
### Is this a duplicate?

- [ ] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cccl/issues) for this request and that I agree to the [Code of Conduct](CODE_OF_CONDUCT.md)

### Area

General CCCL

Pick a CUDA sample and reimplement part of it using the C++ runtime, showing how old and new runtime features can be mixed and interoperate.

Provide code and explanation for integrating C++ runtime with legacy CUDA APIs in a sample, documenting the integration points and any limitations.

Alternative is to keep samples separate, but mixing shows migration and interoperability paths.

### Additional context

Useful for users migrating to new runtime or maintaining legacy code.

davebayer:
Closes #5715

### DIFF
--- a/cudax/examples/cccl-rt/legacy_interact/CMakeLists.txt
+++ b/cudax/examples/cccl-rt/legacy_interact/CMakeLists.txt
@@ -1,0 +1,69 @@
+find_package(cudax REQUIRED) # already found, bring in version info.
+find_package(CUDAToolkit REQUIRED)
+
+set(legculib_target cudax.example.cccl-rt.legacy_interact.legculib)
+add_library(${legculib_target} SHARED legculib.cpp)
+cccl_configure_target(${legculib_target} DIALECT 17)
+target_link_libraries(${legculib_target}
+  PUBLIC
+    CUDA::cuda_driver
+  PRIVATE
+    CUDA::nvrtc
+    cudax.compiler_interface_cpp17)
+
+function(cudax_add_cccl_rt_legacy_interact_example target_name_var cudax_target)
+  cudax_get_target_property(config_prefix ${cudax_target} PREFIX)
+  cudax_get_target_property(config_dialect ${cudax_target} DIALECT)
+
+  set(example_name legacy_interact)
+
+  # The actual name of the test's target:
+  set(example_target ${config_prefix}.example.cccl-rt.${example_name})
+  set(${target_name_var} ${example_target} PARENT_SCOPE)
+
+  # Related target names:
+  set(config_meta_target ${config_prefix}.examples)
+  set(example_meta_target cudax.all.example.cccl-rt.${example_name})
+
+  add_executable(${example_target} "main.cpp")
+  cccl_configure_target(${example_target} DIALECT ${config_dialect})
+  target_link_libraries(${example_target} PRIVATE
+    ${cudax_target}
+    libcudacxx::libcudacxx
+    ${legculib_target}
+    CUDA::cudart
+  )
+  target_compile_options(${example_target} PRIVATE
+    $<$<COMPILE_LANG_AND_ID:CUDA,NVIDIA>:--extended-lambda>
+  )
+  target_compile_definitions(${example_target} PRIVATE
+    "LIBCUDACXX_ENABLE_EXPERIMENTAL_MEMORY_RESOURCE"
+  )
+
+  cudax_clone_target_properties(${example_target} ${cudax_target})
+  target_include_directories(${example_target} PRIVATE "${CUB_SOURCE_DIR}/examples")
+
+  # Add to the active configuration's meta target
+  add_dependencies(${config_meta_target} ${example_target})
+
+  # Meta target that builds examples with this name for all configurations:
+  if (NOT TARGET ${example_meta_target})
+    add_custom_target(${example_meta_target})
+  endif()
+  add_dependencies(${example_meta_target} ${example_target})
+
+  add_test(NAME ${example_target}
+    COMMAND "$<TARGET_FILE:${example_target}>"
+  )
+endfunction()
+
+foreach(cudax_target IN LISTS cudax_TARGETS)
+  cudax_get_target_property(config_prefix ${cudax_target} PREFIX)
+
+  # Metatarget for the current configuration's tests:
+  set(config_meta_target ${config_prefix}.examples.cccl-rt)
+  add_custom_target(${config_meta_target})
+  add_dependencies(${config_prefix}.all ${config_meta_target})
+
+  cudax_add_cccl_rt_legacy_interact_example(example_target ${cudax_target})
+endforeach()
--- a/cudax/examples/cccl-rt/legacy_interact/legculib.cpp
+++ b/cudax/examples/cccl-rt/legacy_interact/legculib.cpp
@@ -1,0 +1,373 @@
+#include "legculib.h"
+
+#include <cuda.h>
+#include <nvrtc.h>
+#include <stddef.h>
+#include <stdint.h>
+#include <stdio.h>
+
+//--------------------------------------------------------------------------------------------------------------------//
+// Library private data
+//--------------------------------------------------------------------------------------------------------------------//
+
+//! @brief The CUDA device handle.
+static CUdevice device{};
+
+//! @brief The device compute capability.
+static int device_cc{};
+
+//! @brief The device primary context.
+static CUcontext context{};
+
+//! @brief The device memory pool.
+static CUmemoryPool device_memory_pool{};
+
+//! @brief The cache of modules containing the kernels for sum_of_products.
+static CUmodule sum_of_products_modules[LEGCULIB_DATATYPE_COUNT]{};
+
+//! @brief The cache of kernel handles for sum_of_products.
+static CUfunction sum_of_products_kernels[LEGCULIB_DATATYPE_COUNT]{};
+
+//! @brief Array of datatype names. Should be indexed by \c legculib_datatype.
+static constexpr const char* datatype_name[]{
+  "int",
+  "long long",
+  "unsigned",
+  "unsigned long long",
+  "float",
+  "double",
+};
+
+//--------------------------------------------------------------------------------------------------------------------//
+// Error handling
+//--------------------------------------------------------------------------------------------------------------------//
+
+//! @brief Gets the error string for a given \c error.
+//! @note Public API.
+extern "C" const char* legculib_get_error_string(legculib_error error)
+{
+  switch (error)
+  {
+    case LEGCULIB_SUCCESS:
+      return "operation was successful";
+    case LEGCULIB_ERROR_NOT_INITIALIZED:
+      return "library was not initialized before calling the API";
+    case LEGCULIB_ERROR_INVALID_ARGUMENT:
+      return "an invalid argument was passed to the API call";
+    case LEGCULIB_ERROR_KERNEL_COMPILATION:
+      return "a kernel compilation failed";
+    case LEGCULIB_ERROR_MALLOC:
+      return "failed to allocate memory from the heap";
+    case LEGCULIB_ERROR_CUDA_DRIVER:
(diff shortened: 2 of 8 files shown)


cudax/examples/cccl-rt/legacy_interact/CMakeLists.txt
```diff
@@ -0,0 +1,65 @@
+cmake_minimum_required(VERSION 3.26)
+
+find_package(cudax) # already found, bring in version info.
+find_package(CUDAToolkit)
+
+add_library(cudax.all.example.cccl-rt.legacy_interact.legculib SHARED legculib.cpp)
```
alliepiper: The CCCL, the `.all.` metatargets are used to build all configurations (e.g. something like `project.all.headers` builds `project.cpp17.headers` and `project.cpp20.headers`). Here, it's being used for a utility library, which breaks this convention.

Let's rename this to something like `cudax.example.cccl-rt.legacy_interact.legculib`.

cudax/examples/cccl-rt/legacy_interact/CMakeLists.txt
```diff
@@ -0,0 +1,65 @@
...
+target_link_libraries(cudax.all.example.cccl-rt.legacy_interact.legculib PUBLIC CUDA::cuda_driver PRIVATE CUDA::nvrtc)
```
alliepiper: Also link with `cudax.compiler_interface_cpp17` to bring in the project's standard warning flags, etc.

```suggestion
target_link_libraries(${legculib_target}
  PUBLIC 
    CUDA::cuda_driver
  PRIVATE
    CUDA::nvrtc
    cudax.compiler_interface_cpp17
)
```

cudax/examples/cccl-rt/legacy_interact/CMakeLists.txt
```diff
@@ -0,0 +1,65 @@
...
+target_compile_features(cudax.all.example.cccl-rt.legacy_interact.legculib PRIVATE cxx_std_17)
```
alliepiper: The `cccl_configuure_target` utility will take care of this (and more):

```suggestion
cccl_configure_target(${legculib_target} DIALECT 17)
```

cudax/examples/cccl-rt/legacy_interact/CMakeLists.txt
alliepiper: Also consider pulling this target name out into a variable for readability / typo-avoidance:

```suggestion
set(legculib_target cudax.example.cccl-rt.legacy_interact.legculib)
add_library(${legculib_target} SHARED legculib.cpp)
```

cudax/examples/cccl-rt/legacy_interact/CMakeLists.txt
```diff
@@ -0,0 +1,65 @@
...
+  # Related target names:
+  set(config_meta_target ${config_prefix}.examples)
+  set(example_meta_target cudax.all.example.cccl-rt.${example_name})
+
+  add_executable(${example_target} "main.cpp")
+  cccl_configure_target(${example_target} DIALECT ${config_dialect})
+  target_link_libraries(${example_target} PRIVATE
+    ${cudax_target}
+    libcudacxx::libcudacxx
+    cudax.all.example.cccl-rt.legacy_interact.legculib
```
alliepiper:
```suggestion
    ${legculib_target}
```

cudax/examples/cccl-rt/legacy_interact/CMakeLists.txt
```diff
@@ -0,0 +1,65 @@
...
+    CUDA::cudart
+  )
```
alliepiper, line 35: Also the project flags:

```suggestion
    cudax.compiler_interface_cpp${config_dialect}
  )
```

cudax/examples/cccl-rt/legacy_interact/CMakeLists.txt
```diff
@@ -0,0 +1,65 @@
+cmake_minimum_required(VERSION 3.26)
```
alliepiper: Is this necessary? I might be missing something, but everything here looks like it should be available with our normal default of 3.21, in which case this line can be removed and inherit from the parent directory.

alliepiper (changes requested):
Reviewed the cmake and left some comments.

It also might be nice to add a README describing what this is testing and why it's separated from cudax's other examples.

cudax/examples/cccl-rt/legacy_interact/CMakeLists.txt
alliepiper: Actually nvm -- we provide this through the `cudax_target` linked earlier.

cudax/examples/cccl-rt/legacy_interact/CMakeLists.txt
davebayer: It's because I initially wanted the legculib target to link to `CUDA::nvrtc_static` which is available in CMake 3.26+. The idea was that the legculib should only need libcuda to work.. But I am scared of the windows static/dynamic runtime mixing problems, so I just switched to `CUDA::nvrtc`, I don't think anyone will care :)

cudax/examples/cccl-rt/legacy_interact/main.cpp
```diff
@@ -0,0 +1,143 @@
...
+  cuda::device_ref device = cuda::devices[0];
+
+  // Initialize the legacy library for the device.
+  safe_call(legculib_init(device.get()));
+
+  // Create a stream for the device.
+  cuda::stream stream{device};
+
+  // Get the legculib's sum of products kernel compile.
+  safe_call(legculib_sum_of_products(nullptr, nullptr, nullptr, nullptr, 0, LEGCULIB_I32));
```
pciolkosz: Is it intentional this says `I32` and usage below says `F32`?

cudax/examples/cccl-rt/legacy_interact/legculib.cpp
```diff
@@ -0,0 +1,340 @@
...
+  CALL_CUDA(cuDeviceGetAttribute(&cc_major, CU_DEVICE_ATTRIBUTE_COMPUTE_CAPABILITY_MAJOR, device));
+  int cc_minor{};
+  CALL_CUDA(cuDeviceGetAttribute(&cc_minor, CU_DEVICE_ATTRIBUTE_COMPUTE_CAPABILITY_MINOR, device));
+  device_cc = 10 * cc_major + cc_minor;
+
+  // Retain the primary context.
+  CALL_CUDA(cuDevicePrimaryCtxRetain(&context, device));
+
+  return success();
+}
```
leofang, line 151: FYI, only the three of us will do it like this 😆 Many library folks just do
```c++
int orig_dev;
CALL_CUDA( cudaGetDevice(&orig_dev) );
CALL_CUDA( cudaSetDevice(ordinal) );
CALL_CUDA( cudaFree(0) );
CALL_CUDA( cudaSetDevice(orig_dev) );
```
I think this example should be made as cudart-based.

cudax/examples/cccl-rt/legacy_interact/legculib.cpp
```diff
@@ -0,0 +1,340 @@
...
+{
+  if constexpr (block_size >= 64) sdata[tid] += sdata[tid + 32];
+  if constexpr (block_size >= 32) sdata[tid] += sdata[tid + 16];
+  if constexpr (block_size >= 16) sdata[tid] += sdata[tid + 8];
+  if constexpr (block_size >= 8) sdata[tid] += sdata[tid + 4];
+  if constexpr (block_size >= 4) sdata[tid] += sdata[tid + 2];
+  if constexpr (block_size >= 2) sdata[tid] += sdata[tid + 1];
+}
+
+extern "C" __global__ void kernel(T* result, const T* a, const T* b, unsigned n)
```
leofang, line 224: Using NVRTC in this example is overly complicated. Most libraries just have a `__global__` kernel that's compiled by NVCC and launched using `kernel<<< ... >>>(...);`.

leofang (changes requested):
We need to show mixing cccl-rt with cudaart-based libraries (which are the majority of them). It is less complex, closer to real-life, and in turn helps cut many lines.

cudax/examples/cccl-rt/legacy_interact/main.cpp
davebayer: No, it's a bug, thanks

cudax/examples/cccl-rt/legacy_interact/legculib.cpp
davebayer: Do you find this disturbing?

The idea was to create a library that is completely independent on CUDA Runtime, but uses CUDA Runtime types to interact with other libraries. I thought it is quite interesting and explanatory.

If you don't like how complicated this is, I would suggest just to add a another example interacting with CUDA Runtime exclusively.

cudax/examples/cccl-rt/legacy_interact/legculib.cpp
davebayer: Hm, again, I thought it would be a bit more interesting like this :D I can rework it if you want.

cudax/examples/cccl-rt/legacy_interact/legculib.cpp
leofang: I think the main purpose of #5715 is to show how cccl-rt interacts with a library built against the "old runtime" (aka cudart), and that means a library using both types and APIs from cudart, not just types. This is where the majority of the developers and uses cases are. 

Due to the trajectory of CUDA development, for many years we have been teaching internal and external CUDA programmers to avoid driver APIs in favor of the (old) runtime APIs, and this has become internalized to seasoned CUDA developers. We want to leverage examples (and not explode the number of them) to educate them how a transition path to the "new runtime" would look like.

cudax/examples/cccl-rt/legacy_interact/legculib.cpp
leofang: It is interesting, but is a bit more complicated than necessary 😅 Most CUDA C++ developers do not use nor are they aware of NVRTC. NVCC is the overwhelming majority.

cudax/examples/cccl-rt/legacy_interact/CMakeLists.txt
leofang: By removing the dependency on NVRTC as I suggested below, this becomes moot.

Statically linking to NVRTC is a terrible idea (ping me if you want to know why) that requires the developer to be fully aware of the tradeoff. We should not make it a default for no reason.

cudax/examples/cccl-rt/legacy_interact/legculib.cpp
pciolkosz: I agree, the example should be simpler and moving to use the Runtime API would help with it

cudax/examples/cccl-rt/legacy_interact/CMakeLists.txt
```diff
@@ -0,0 +1,69 @@
+find_package(cudax REQUIRED) # already found, bring in version info.
```
alliepiper, line 1: I don't see this being used -- can likely be removed if we aren't using the cudax version variables.

alliepiper (approved):
Signing off on cmake changes with a minor nit.

cudax/examples/cccl-rt/legacy_interact/legculib.cpp
pciolkosz: I would say the NVRTC stuff would be interesting if we had a modern replacement for it to show the difference. For now we don't have it, we should start simpler and maybe add NVRTC later?

PR #6018, opened 2025-09-25
Refactor thrust::mismatch to use CUDA iterators

bernhardmgruber:
Revisits: #3591

- [x] Merge before: #6019
- [x] nvc++ successfully compiles the test from #3591
- [x] Merge before: #6323

PR #6323, opened 2025-10-23
Ensure that we can instantiate `zip_function` with a type that is not non-const invocable

miscco:
I did not constraint the call operators properly, so when the compiler tried to build the overload set for an
instantiation of a `const zip_function` that holds a function with only a mutable call operator it would fail.


bernhardmgruber (approved):

bernhardmgruber:
I confirm it fixes the issue in #6018

davebayer (approved):

PR #6325, opened 2025-10-23
Provide utilities to check pointer memory space (host/device/managed)

fbusato:
## Description

Provide the following functions to check the pointer memory space.
`cuda::` namespace, `cuda/memory` header.

```cpp
 template <typename Pointer>
 [[nodiscard]] __host__
 bool is_host_accessible(Pointer ptr);

 template <typename Pointer>
 [[nodiscard]] __host__
 bool is_device_accessible(Pointer ptr, device_ref device);

 template <typename Pointer>
 [[nodiscard]] __host__
 bool is_managed_pointer(Pointer ptr);
```


docs/libcudacxx/extended_api/memory/is_pointer_accessible.rst
```diff
@@ -0,0 +1,103 @@
...
+
+Defined in ``<cuda/memory>`` header.
+
+.. code:: cuda
+
+   namespace cuda {
+
+   template <typename Pointer>
+   [[nodiscard]] __host__ constexpr
+   bool is_host_accessible_pointer(Pointer ptr) noexcept;
```
davebayer: Can't it be just `is_host_accessible`? It would be a more versatile symbol name, which could also be used for e. g. memory resources (applies to all other functions)

docs/libcudacxx/extended_api/memory/is_pointer_accessible.rst
```diff
@@ -0,0 +1,103 @@
...
+Determines whether the memory referenced by ``ptr`` is accessible from the host.
+
+----
+
+.. code:: cuda
+
+   namespace cuda {
+
+   template <typename Pointer>
+   [[nodiscard]] __host__ constexpr
```
davebayer: I wouldn't make these APIs `constexpr`. I believe the motivation behind this is so we can use them in `constexpr` functions in `_CCCL_ASSERT` statements, but it doesn't make any sense.

I think it will do more harm than good.

libcudacxx/include/cuda/__memory/is_pointer_accessible.h
```diff
@@ -0,0 +1,97 @@
...
+#include <cuda/std/__cccl/prologue.h>
+
+_CCCL_BEGIN_NAMESPACE_CUDA
+
+#if _CCCL_HAS_CTK() && !_CCCL_COMPILER(NVRTC)
+
+_CCCL_TEMPLATE(typename _Pointer)
+_CCCL_REQUIRES(::cuda::std::contiguous_iterator<_Pointer> || ::cuda::std::is_pointer_v<_Pointer>)
+[[nodiscard]]
+_CCCL_HOST_API constexpr bool is_managed_pointer(_Pointer __p) noexcept
```
davebayer: Shouldn't these functions `throw` if the driver call fails? It is good enough for asserts, because you are just trying to verify a precondition, but if you write a code that really needs a precise answer, I don't think it's a good idea to return a possibly untrue value.

libcudacxx/include/cuda/__memory/is_pointer_accessible.h
```diff
@@ -0,0 +1,97 @@
...
+        || (__status == ::cudaSuccess && ((__type == ::CU_MEMORYTYPE_HOST) || (__type == ::CU_MEMORYTYPE_UNIFIED)))
+        || ::cuda::is_managed_pointer(__p);
+  }
+  return true; // cannot be verified
+}
+
+_CCCL_TEMPLATE(typename _Pointer)
+_CCCL_REQUIRES(::cuda::std::contiguous_iterator<_Pointer> || ::cuda::std::is_pointer_v<_Pointer>)
+[[nodiscard]]
+_CCCL_HOST_API constexpr bool is_device_accessible_pointer(_Pointer __p) noexcept
```
davebayer: This API should take a `device_ref` parameter to specify the device we are querying the accessibility for.

libcudacxx/include/cuda/__memory/is_pointer_accessible.h
```diff
@@ -0,0 +1,97 @@
...
+{
+  if (!cuda::std::__cccl_default_is_constant_evaluated())
+  {
+    auto __p1 = ::cuda::std::to_address(__p);
+    ::CUmemorytype __type{};
+    const auto __status =
+      ::cuda::__driver::__pointerGetAttributeNoThrow<::CU_POINTER_ATTRIBUTE_MEMORY_TYPE>(__type, __p1);
+    return (__status == ::cudaErrorInvalidValue)
+        || (__status == ::cudaSuccess && ((__type == ::CU_MEMORYTYPE_DEVICE) || (__type == ::CU_MEMORYTYPE_UNIFIED)))
+        || ::cuda::is_managed_pointer(__p);
```
davebayer: We are missing pinned memory here

libcudacxx/test/libcudacxx/cuda/memory/is_pointer_accessible.pass.cpp
```diff
@@ -0,0 +1,79 @@
...
+    assert(cuda::is_managed_pointer(ptr + 1) == is_managed_accessible);
+  }
+}
+
+bool test()
+{
+  [[maybe_unused]] int host_ptr1[] = {1, 2, 3, 4};
+  [[maybe_unused]] auto host_ptr2  = new int[2];
+  [[maybe_unused]] int* host_ptr3  = nullptr;
+  cudaMallocHost(&host_ptr3, sizeof(int) * 2);
```
davebayer: Should be tested for device accessibility

docs/libcudacxx/extended_api/memory/is_pointer_accessible.rst
fbusato: yes, `is_host_accessible_pointer` is pretty long name

docs/libcudacxx/extended_api/memory/is_pointer_accessible.rst
fbusato: I don't have a strong opinion on that.

libcudacxx/include/cuda/__memory/is_pointer_accessible.h
fbusato: In the current status, is a driver call fails for an unexpected problem, it always returns false.
I'm very conservative regarding exceptions. @miscco any thought?

libcudacxx/include/cuda/__memory/is_pointer_accessible.h
fbusato: that's a good idea! can it have a default value?

libcudacxx/test/libcudacxx/cuda/memory/is_pointer_accessible.pass.cpp
fbusato: they are. `test_accessible_pointer` tests for all spaces

libcudacxx/include/cuda/__memory/is_pointer_accessible.h
fbusato: pinned memory is not directly accessible from device. I could add a `cudaHostAllocMapped` test

libcudacxx/include/cuda/__memory/is_pointer_accessible.h
davebayer: I would say no for now. We already started discussing something like `cuda::current_device()` or something similar in #6023, but haven't resolved it yet. I think it should be an explicit parameter anyway

libcudacxx/include/cuda/__memory/is_pointer_accessible.h
davebayer: Yeah, I think we could test it querying `CU_POINTER_ATTRIBUTE_DEVICE_POINTER` attribute and checking the return value

davebayer: Or is this something we don't want, because yes it is device accessible memory, but not under this address?

fbusato:
@miscco any idea how to fix `contiguous_iterator` with `void*` with GCC7?
```
cuda/std/__iterator/concepts.h:371:387: error: invalid use of ‘void’
_CCCL_CONCEPT_FRAGMENT(
```

```
cuda/std/__iterator/concepts.h: In substitution of ‘template<class ... _As> char (& cuda::std::__4::__sized_sentinel_for__CCCL_CONCEPT_FRAGMENT_(__cccl_tag<_As ...>*, decltype (& __sized_sentinel_for__CCCL_CONCEPT_FRAGMENT_impl_<_As ...>)))[1] [with _As = {void*, void*}]’:

cuda/std/__iterator/concepts.h:380:81:   required from ‘constexpr const bool cuda::std::__4::sized_sentinel_for<void*, void*>’
cuda/std/__iterator/concepts.h:459:319:   required by substitution of ‘template<class ... _Args, class> static constexpr bool cuda::std::__4::__cccl_requires_expr_detail_24::__cccl_is_satisfied(__cccl_tag<_As ...>*, int) [with _Args = {void, void*}; <template-parameter-1-2> = <missing>]’
cuda/std/__iterator/concepts.h:459:122:   required from ‘constexpr const bool cuda::std::__4::random_access_iterator<void*>’
cuda/std/__iterator/concepts.h:468:156:   required by substitution of ‘template<class ... _As> char (& cuda::std::__4::__contiguous_iterator__CCCL_CONCEPT_FRAGMENT_(__cccl_tag<_As ...>*, decltype (& __contiguous_iterator__CCCL_CONCEPT_FRAGMENT_impl_<_As ...>)))[1] [with _As = {void*}]’
cuda/std/__iterator/concepts.h:478:83:   required from ‘constexpr const bool cuda::std::__4::contiguous_iterator<void*>’
````

File: docs/libcudacxx/extended_api/memory/is_pointer_accessible.rst [hunk 7]
```diff
@@ -0,0 +1,111 @@
...
+
+----
+
+.. code:: cuda
+
+   namespace cuda {
+
+   template <typename Pointer>
+   [[nodiscard]] __host__
+   bool is_device_accessible(Pointer ptr);
```
davebayer: You didn't add the `device_ref` parameter

docs/libcudacxx/extended_api/memory/is_pointer_accessible.rst
```diff
@@ -0,0 +1,111 @@
...
+
+        assert(cuda::is_host_accessible(host_ptr));
+        assert(!cuda::is_device_accessible(host_ptr));
+
+        assert(cuda::is_device_accessible(device_ptr));
+        assert(!cuda::is_host_accessible(device_ptr));
+
+        assert(cuda::is_host_accessible(managed_ptr));
+        assert(cuda::is_device_accessible(managed_ptr));
+        assert(cuda::is_managed_pointer(managed_ptr));
```
davebayer:
```suggestion
        cuda::device_ref device = cuda::devices[0];
        cudaSetDevice(device.get());

        cudaMallocHost(&host_ptr, 1024);
        cudaMalloc(&device_ptr, 1024);
        cudaMallocManaged(&managed_ptr, 1024);

        assert(cuda::is_host_accessible(host_ptr));
        assert(!cuda::is_device_accessible(host_ptr, device));

        assert(cuda::is_device_accessible(device_ptr, device));
        assert(!cuda::is_host_accessible(device_ptr));

        assert(cuda::is_host_accessible(managed_ptr));
        assert(cuda::is_device_accessible(managed_ptr, device));
        assert(cuda::is_managed_pointer(managed_ptr));
```

libcudacxx/include/cuda/__memory/is_pointer_accessible.h
```diff
@@ -0,0 +1,107 @@
...
+    bool __is_managed{};
+    const auto __status =
+      ::cuda::__driver::__pointerGetAttributeNoThrow<::CU_POINTER_ATTRIBUTE_IS_MANAGED>(__is_managed, __p1);
+    if (__status != ::cudaErrorInvalidValue && __status != ::cudaSuccess)
+    {
+      ::cuda::__throw_cuda_error(__status, "is_managed_pointer failed()", _CCCL_BUILTIN_PRETTY_FUNCTION());
+    }
+    return (__status == ::cudaErrorInvalidValue) || __is_managed;
+  }
+  return true; // cannot be verified
```
davebayer:
```suggestion
  const auto __p1 = ::cuda::std::to_address(__p);
  bool __is_managed{};
  const auto __status =
    ::cuda::__driver::__pointerGetAttributeNoThrow<::CU_POINTER_ATTRIBUTE_IS_MANAGED>(__is_managed, __p1);
  if (__status != ::cudaErrorInvalidValue && __status != ::cudaSuccess)
  {
    ::cuda::__throw_cuda_error(__status, "is_managed_pointer failed()", _CCCL_BUILTIN_PRETTY_FUNCTION());
  }
  return (__status == ::cudaErrorInvalidValue) || __is_managed;
```
We don't need this anymore

docs/libcudacxx/extended_api/memory/is_pointer_accessible.rst
```diff
@@ -0,0 +1,111 @@
...
+
+Defined in ``<cuda/memory>`` header.
+
+.. code:: cuda
+
+   namespace cuda {
+
+   template <typename Pointer>
+   [[nodiscard]] __host__
+   bool is_host_accessible(Pointer ptr);
```
davebayer: Do we really need to support the contiguous iterators? Is there any benefit? In my opinion the API now takes some unknown type that is named `Pointer`, but needn't to be a pointer.

I would suggest taking a `const void*` instead. If anyone wants to use a contiguous iterator, he can do `cuda::std::to_address(it)`.

docs/libcudacxx/extended_api/memory/is_pointer_accessible.rst
fbusato: There is no need to set or switch the device because the pointers are visible in all cuda contexts.

docs/libcudacxx/extended_api/memory/is_pointer_accessible.rst
fbusato: not really.  I also would prefer `void*`. IIRC contiguous iterators were suggested by @miscco when we worked on host/device mdspan.

docs/libcudacxx/extended_api/memory/is_pointer_accessible.rst
fbusato: yes, I removed it. Pointers are visible in all cuda contexts.

davebayer: Yeah, but what about the case when you have 2 GPUs and the memory is allocated on the first one. Then it's not accessible from the second device directly

libcudacxx/include/cuda/__memory/is_pointer_accessible.h
davebayer: Applies to other the functions as well..

davebayer: Applies to other the functions as well..

docs/libcudacxx/extended_api/memory/is_pointer_accessible.rst
fbusato: this distinction could be confusing. What about another overload with `device_ref` to check if the pointer is accessible by a specific GPU?

davebayer: @pciolkosz @miscco what do you think?

pciolkosz: What is the envisioned use case for these APIs? I think I agree with @davebayer that device accessibility would be more useful with a `device_ref` argument. Currently the API returns if the pointer is accessible from some device in the system, which is why I ask about the use case, because it seems to be a partial solution. We might want to provide just the `device_ref` taking version.
Or we could just be consistent with `is_managed_pointer` and expose `is_host_pointer` and `is_device_pointer`? Then user can combine multiple of those calls to ask the exact question they need? This way the `device_ref` argument is not needed and the semantics is more clear.

I would also be curious what would be a long term vision for the exposure other pointer attributes, because that could help decide what to do with the ones in this PR. If we want to expose something more inline with what other attribute queries we exposed so far, the APIs proposed here might become redundant (which might be fine as well, not sure). But we could also try a new approach with pointer attributes and expose a collection of free query functions, especially since we don't have a pointer object we expose. But it would also be nice to make sure that new approach extends well to all relevant pointer attributes

fbusato: `is_device_accessible(Pointer, device_ref);` is definitely valuable IMO. However, strictly requiring a device_ref for all usage looks too strong. I would prefer to provide both versions.

For the long term, we could propose more sophisticated API that can query several attributes. For now, I would prefer just these three simple functions

davebayer: I agree with Piotr. I would keep `is_device_accessible(Pointer, device_ref)` as the only overload for this function. I would rather see additional `is_device_ptr(Pointer)` and `is_device_ptr(Pointer, device_ref)` functions (or `is_device_address`). Querying address information and accessibility are 2 separate things in my opinion.

libcudacxx/include/cuda/__memory/is_pointer_accessible.h
```diff
@@ -79,13 +79,34 @@ _CCCL_HOST_API bool is_device_accessible(_Pointer __p)
...
+    ::cuda::__driver::__pointerGetAttributeNoThrow<::CU_POINTER_ATTRIBUTE_DEVICE_ORDINAL>(__ptr_dev_id, __p1);
+  if (__status != ::cudaErrorInvalidValue && __status != ::cudaSuccess)
+  {
+    ::cuda::__throw_cuda_error(__status, "is_device_accessible() failed", _CCCL_BUILTIN_PRETTY_FUNCTION());
+  }
+  if (__status == ::cudaErrorInvalidValue) // unregistered pointer
+  {
+    return true;
+  }
+  const bool __can_access_peer = ::cuda::__driver::__deviceCanAccessPeer(__device.get(), __ptr_dev_id);
```
davebayer: This will not cover all of the cases. We need to query the memory pool the pointer was allocated from using `CU_POINTER_ATTRIBUTE_MEMPOOL_HANDLE` and then query the access mode using `cuMemPoolGetAccess`

davebayer: I would suggest returning `true` if any of `CU_MEM_ACCESS_FLAGS_PROT_READ` or `CU_MEM_ACCESS_FLAGS_PROT_READWRITE` is set

docs/libcudacxx/extended_api/memory/is_pointer_accessible.rst
davebayer: Let's go for `const void*`, we use it in all other APIs

libcudacxx/include/cuda/__memory/is_pointer_accessible.h
davebayer: Okay, so it gets even more complicated. If the query for mempool of the pointer returns `nullptr`, it wasn't allocated from a pool and we should check for `cuDeviceCanAccessPeer`. Otherwise use the `cuMemPoolGetAccess`..

fbusato:
implemented most of the suggestions. I would like a second opinion on exceptions and input type (pointer vs. contiguous iterator)  from @miscco

libcudacxx/include/cuda/__driver/driver_api.h
```diff
@@ -324,6 +325,20 @@ _CCCL_HOST_API void __memsetAsync(void* __dst, _Tp __value, size_t __count, ::CU
   }
 }
 
+_CCCL_HOST_API inline ::cudaError_t
```
miscco: Nitpick: should be `[[nodiscard]]`
```suggestion
[[nodiscard]] _CCCL_HOST_API inline ::cudaError_t
```

libcudacxx/include/cuda/__driver/driver_api.h
```diff
@@ -324,6 +325,20 @@ _CCCL_HOST_API void __memsetAsync(void* __dst, _Tp __value, size_t __count, ::CU
...
+__memRetainAllocationHandleNoThrow(::CUmemGenericAllocationHandle* __handle, void* __addr)
+{
+  static auto __driver_fn = _CCCLRT_GET_DRIVER_FUNCTION(cuMemRetainAllocationHandle);
+  return static_cast<::cudaError_t>(__driver_fn(__handle, __addr));
+}
+
+_CCCL_HOST_API inline ::cudaError_t
```
miscco: Ditto:
```suggestion
[[nodiscard]] _CCCL_HOST_API inline ::cudaError_t
```

libcudacxx/include/cuda/__driver/driver_api.h
```diff
@@ -324,6 +325,20 @@ _CCCL_HOST_API void __memsetAsync(void* __dst, _Tp __value, size_t __count, ::CU
...
+__memGetAllocationPropertiesFromHandleNoThrow(::CUmemAllocationProp* __prop, ::CUmemGenericAllocationHandle __handle)
+{
+  static auto __driver_fn = _CCCLRT_GET_DRIVER_FUNCTION(cuMemGetAllocationPropertiesFromHandle);
+  return static_cast<::cudaError_t>(__driver_fn(__prop, __handle));
+}
+
 _CCCL_HOST_API inline ::cudaError_t __mempoolCreateNoThrow(::CUmemoryPool* __pool, ::CUmemPoolProps* __props)
```
miscco: Here too while we are at it
```suggestion
[[nodiscard]] _CCCL_HOST_API inline ::cudaError_t __mempoolCreateNoThrow(::CUmemoryPool* __pool, ::CUmemPoolProps* __props)
```

libcudacxx/include/cuda/__memory/is_pointer_accessible.h
```diff
@@ -0,0 +1,132 @@
...
+  const auto __p1                  = ::cuda::std::to_address(__p);
+  ::CUpointer_attribute __attrs[2] = {::CU_POINTER_ATTRIBUTE_MEMORY_TYPE, ::CU_POINTER_ATTRIBUTE_IS_MANAGED};
+  ::CUmemorytype __memory_type     = static_cast<::CUmemorytype>(0);
+  int __is_managed                 = 0;
+  void* __results[2]               = {&__memory_type, &__is_managed};
+  const auto __status              = ::cuda::__driver::__pointerGetAttributesNoThrow(
+    ::cuda::std::span<::CUpointer_attribute, 2>{__attrs}, ::cuda::std::span<void*, 2>{__results}, __p1);
+  if (__status != ::cudaSuccess)
+  {
+    ::cuda::__throw_cuda_error(__status, "__query_host_or_device_accessible() failed", _CCCL_BUILTIN_PRETTY_FUNCTION());
```
miscco: Should this be 
```suggestion
    ::cuda::__throw_cuda_error(__status, "is_host_accessible() failed", _CCCL_BUILTIN_PRETTY_FUNCTION());
```

libcudacxx/test/libcudacxx/cuda/memory/is_pointer_accessible.pass.cpp
```diff
@@ -0,0 +1,143 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of the libcu++ Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES.
+//
+//===----------------------------------------------------------------------===//
+// UNSUPPORTED: nvrtc
+#include <cuda/__runtime/ensure_current_context.h>
```
miscco, line 11: Nitpick: add newline
```suggestion

#include <cuda/__runtime/ensure_current_context.h>
```

libcudacxx/test/libcudacxx/cuda/memory/is_pointer_accessible.pass.cpp
```diff
@@ -0,0 +1,143 @@
...
+#include <cuda/std/cassert>
+
+#include <cuda_runtime_api.h>
+
+#include "test_macros.h"
+
+__device__ int device_ptr3[]              = {1, 2, 3, 4};
+__device__ __managed__ int managed_ptr2[] = {1, 2, 3, 4};
+
+int host_ptr5[] = {1, 2, 3, 4};
```
miscco: I am deeply confused by the numbering scheme

libcudacxx/include/cuda/__driver/driver_api.h
```diff
@@ -482,6 +488,15 @@ __pointerGetAttributeNoThrow(__pointer_attribute_value_type_t<_Attr>& __result,
...
 }
 
+[[nodiscard]] _CCCL_HOST_API inline ::cudaError_t __pointerGetAttributesNoThrow(
+  ::cuda::std::span<::CUpointer_attribute> __attrs, ::cuda::std::span<void*> __results, const void* __ptr)
+{
+  _CCCL_ASSERT(__attrs.size() == __results.size(), "size mismatch in __pointerGetAttributesNoThrow");
+  static const auto __driver_fn = _CCCLRT_GET_DRIVER_FUNCTION(cuPointerGetAttributes);
+  return static_cast<::cudaError_t>(__driver_fn(
+    static_cast<unsigned>(__attrs.size()), __attrs.data(), __results.data(), reinterpret_cast<::CUdeviceptr>(__ptr)));
+}
```
davebayer: From what I remember we wanted to keep `driver_api.h` minimal, I'd prefer to have pointers + size here..

davebayer:
```suggestion
template <::cuda::std::size_t _Np>
[[nodiscard]] _CCCL_HOST_API inline ::cudaError_t __pointerGetAttributesNoThrow(
  ::CUpointer_attribute (&__attrs)[N], void* (&__results)[N], const void* __ptr)
{
  static const auto __driver_fn = _CCCLRT_GET_DRIVER_FUNCTION(cuPointerGetAttributes);
  return static_cast<::cudaError_t>(__driver_fn(
    static_cast<unsigned>(N), __attrs, __results, reinterpret_cast<::CUdeviceptr>(__ptr)));
}
```

libcudacxx/include/cuda/__memory/is_pointer_accessible.h
```diff
@@ -0,0 +1,118 @@
...
+_CCCL_HOST_API inline bool is_managed(const void* __p)
+{
+  bool __is_managed{};
+  const auto __status =
+    ::cuda::__driver::__pointerGetAttributeNoThrow<::CU_POINTER_ATTRIBUTE_IS_MANAGED>(__is_managed, __p);
+  if (__status != ::cudaErrorInvalidValue && __status != ::cudaSuccess)
+  {
+    ::cuda::__throw_cuda_error(__status, "is_managed() failed", _CCCL_BUILTIN_PRETTY_FUNCTION());
+  }
+  return (__status == ::cudaErrorInvalidValue) || __is_managed;
```
davebayer: I think we should return the pessimistic result. Also, I think switch will look better here
```suggestion
  switch (__status)
  {
    case ::cudaSuccess:
      return __is_managed;
    case ::cudaErrorInvalidValue:
      return false;
    default:
      ::cuda::__throw_cuda_error(__status, "is_managed() failed", _CCCL_BUILTIN_PRETTY_FUNCTION());
  }
```

libcudacxx/include/cuda/__memory/is_pointer_accessible.h
```diff
@@ -0,0 +1,118 @@
...
+
+[[nodiscard]]
+_CCCL_HOST_API inline bool is_host_accessible(const void* __p)
+{
+  ::CUpointer_attribute __attrs[2] = {::CU_POINTER_ATTRIBUTE_MEMORY_TYPE, ::CU_POINTER_ATTRIBUTE_IS_MANAGED};
+  ::CUmemorytype __memory_type     = static_cast<::CUmemorytype>(0);
+  int __is_managed                 = 0;
+  void* __results[2]               = {&__memory_type, &__is_managed};
+  const auto __status              = ::cuda::__driver::__pointerGetAttributesNoThrow(
+    ::cuda::std::span<::CUpointer_attribute, 2>{__attrs}, ::cuda::std::span<void*, 2>{__results}, __p);
```
davebayer: Should work fine if you accept the driver api suggestion
```suggestion
  const auto __status              = ::cuda::__driver::__pointerGetAttributesNoThrow(__attrs, __results, __p);
```

libcudacxx/include/cuda/__memory/is_pointer_accessible.h
```diff
@@ -0,0 +1,118 @@
...
+    ::CU_POINTER_ATTRIBUTE_IS_MANAGED,
+    ::CU_POINTER_ATTRIBUTE_DEVICE_ORDINAL,
+    ::CU_POINTER_ATTRIBUTE_MEMPOOL_HANDLE};
+  ::CUmemorytype __memory_type = static_cast<::CUmemorytype>(0);
+  int __is_managed             = 0;
+  int __ptr_dev_id             = 0;
+  ::CUmemoryPool __ptr_mempool = nullptr;
+  void* __results[4]           = {&__memory_type, &__is_managed, &__ptr_dev_id, &__ptr_mempool};
+  const auto __status          = ::cuda::__driver::__pointerGetAttributesNoThrow(
+    ::cuda::std::span<::CUpointer_attribute, 4>{__attrs}, ::cuda::std::span<void*, 4>{__results}, __p);
```
davebayer:
```suggestion
  const auto __status          = ::cuda::__driver::__pointerGetAttributesNoThrow(__attrs, __results, __p);
```

libcudacxx/include/cuda/__memory/is_pointer_accessible.h
```diff
@@ -0,0 +1,118 @@
...
+    ::cuda::std::span<::CUpointer_attribute, 2>{__attrs}, ::cuda::std::span<void*, 2>{__results}, __p);
+  if (__status != ::cudaSuccess)
+  {
+    ::cuda::__throw_cuda_error(__status, "is_host_accessible() failed", _CCCL_BUILTIN_PRETTY_FUNCTION());
+  }
+  if (__memory_type == static_cast<::CUmemorytype>(0)) // check if the pointer is unregistered
+  {
+    return true;
+  }
+  return (__is_managed || __memory_type == ::CU_MEMORYTYPE_UNIFIED || __memory_type == ::CU_MEMORYTYPE_HOST);
```
davebayer: I am not sure if we are handling the errors correctly, the docs say:
_Instead, the attributes are assigned default NULL values and CUDA_SUCCESS is returned._

I am not sure if it doesn't mean that the `__attrs` array values will be `0`

davebayer (changes requested):

libcudacxx/include/cuda/__driver/driver_api.h
fbusato: this is also my feeling, but I would like to hear from @pciolkosz  and @miscco. I recently added `expected` which is very heavy

libcudacxx/include/cuda/__memory/is_pointer_accessible.h
fbusato: pretty sure this is correct. Anyway, let me double-check with the driver team

libcudacxx/include/cuda/__memory/is_pointer_accessible.h
fbusato: not sure about that. We cannot use these functions for validation anymore because we don't know the reason of `false`.

I'm actually thinking about a three-state (true, false, unregister) solution.

libcudacxx/include/cuda/__memory/is_pointer_accessible.h
fbusato: checked. It means array values will be 0

libcudacxx/include/cuda/__memory/is_pointer_accessible.h
```diff
@@ -0,0 +1,148 @@
...
+
+[[nodiscard]]
+_CCCL_HOST_API inline bool is_host_accessible(const void* __p)
+{
+  if (__p == nullptr)
+  {
+    return false;
+  }
+  [[maybe_unused]] void* __tmp_ptr = nullptr;
+  const auto __status1             = ::cudaGetSymbolAddress(&__tmp_ptr, __p);
```
fbusato: any concern of using cuda runtime api here?

pciolkosz: For this thing you have to use the runtime and I think it would be fine. That said, I don't think this API should try to do the translation for the user. Especially the `is_device_accessible` one is wrong right now. If the user just pass this pointer to a kernel it's not going to be accessible unless they do the translation themselves or use the original symbol (not the pointer we are testing). I think I would prefer to expect the user call `cudaGetSymbolAddress` if needed and then pass the  result to `is_*_accessible`. We could provide a separate utility wrapping it for convenience

fbusato: true, the pointer has a different value on the host. I'm realizing that doing the translation for the user is too far.
I'm thinking to remove it then

docs/libcudacxx/extended_api/memory/is_pointer_accessible.rst
```diff
@@ -0,0 +1,108 @@
...
+.. code:: cuda
+
+   namespace cuda {
+
+   [[nodiscard]] inline
+   bool is_managed(const void* ptr);
+
+   } // namespace cuda
+
+Determines whether the memory referenced by ``ptr`` is backed by Unified Memory (managed memory).
```
davebayer: Suggestion: don't we want to bring them together as:

```suggestion
Defined in the ``<cuda/memory>`` header.

.. code:: cuda

   namespace cuda {

   [[nodiscard]] inline
   bool is_host_accessible(const void* ptr); // (1)
   
   [[nodiscard]] inline
   bool is_device_accessible(const void* ptr, device_ref device); // (2)
   
   [[nodiscard]] inline
   bool is_managed(const void* ptr); // (3)

   } // namespace cuda

-- describe the behaviour for (1), (2) and (3) ---
```

libcudacxx/include/cuda/__memory/is_pointer_accessible.h
davebayer: I also think the function shouldn't try to get the address of a global variable symbol

libcudacxx/include/cuda/__memory/is_pointer_accessible.h
```diff
@@ -0,0 +1,139 @@
...
+#include <cuda/std/__exception/cuda_error.h>
+
+#include <cuda/std/__cccl/prologue.h>
+
+_CCCL_BEGIN_NAMESPACE_CUDA
+
+#if _CCCL_HAS_CTK() && !_CCCL_COMPILER(NVRTC)
+
+[[nodiscard]]
+_CCCL_HOST_API inline bool is_managed(const void* __p)
```
davebayer, line 41: Can we add doxygen docs to all public functions?

libcudacxx/test/libcudacxx/cuda/memory/is_pointer_accessible.pass.cpp
```diff
@@ -0,0 +1,152 @@
...
+  return true;
+}
+
+bool test_memory_pool()
+{
+  cuda::device_ref dev{0};
+  cudaMemPoolProps pool_prop = {};
+  pool_prop.allocType        = cudaMemAllocationTypePinned;
+  pool_prop.location.id      = dev.get();
+  pool_prop.location.type    = cudaMemLocationTypeDevice;
```
davebayer: Can we also add tests for host pinned memory pool and managed memory pool? Managed mempool is CUDA 13.0+ I believe

davebayer (changes requested):

libcudacxx/include/cuda/__memory/is_pointer_accessible.h
fbusato: here I'm not 100% sure. We could handle an additional case, which is good. Otherwise, global variable symbols will be recognized as host pointers.

libcudacxx/include/cuda/__memory/is_pointer_accessible.h
fbusato: I don't see it as a common practice in our codebase but I can add it

libcudacxx/include/cuda/__memory/is_pointer_accessible.h
davebayer: If we handled it in `cuda::launch`, too, I would agree, but I believe we don't do that. I think that if we do this only in one API, it could rather lead to errors.

We can always fix this behaviour in the future.

fbusato: that's fine in this case. I prefer to have a uniform user experience rather than handle edge cases separately.

libcudacxx/include/cuda/__memory/is_pointer_accessible.h
```diff
@@ -0,0 +1,161 @@
...
+  {
+    return false;
+  }
+  ::CUpointer_attribute __attrs[3] = {
+    ::CU_POINTER_ATTRIBUTE_MEMORY_TYPE, ::CU_POINTER_ATTRIBUTE_IS_MANAGED, ::CU_POINTER_ATTRIBUTE_MEMPOOL_HANDLE};
+  ::CUmemorytype __memory_type = static_cast<::CUmemorytype>(0);
+  int __is_managed             = 0;
+  ::CUmemoryPool __mempool     = nullptr;
+  void* __results[3]           = {&__memory_type, &__is_managed, &__mempool};
+  const auto __status2         = ::cuda::__driver::__pointerGetAttributesNoThrow(__attrs, __results, __p);
```
davebayer:
```suggestion
  const auto __status          = ::cuda::__driver::__pointerGetAttributesNoThrow(__attrs, __results, __p);
```

libcudacxx/include/cuda/__memory/is_pointer_accessible.h
```diff
@@ -0,0 +1,161 @@
...
+  if (__memory_type == static_cast<::CUmemorytype>(0))
+  {
+    return true;
+  }
+  // (2) check if a memory pool is associated with the pointer
+  if (__mempool != nullptr)
+  {
+    ::CUmemLocation __prop{::CU_MEM_LOCATION_TYPE_HOST, 0};
+    const auto __pool_flags = ::cuda::__driver::__mempoolGetAccess(__mempool, &__prop);
+    return (__pool_flags == ::CU_MEM_ACCESS_FLAGS_PROT_READ || __pool_flags == ::CU_MEM_ACCESS_FLAGS_PROT_READWRITE);
```
davebayer:
```suggestion
    return (__pool_flags & ::CU_MEM_ACCESS_FLAGS_PROT_READWRITE) != 0;
```

libcudacxx/include/cuda/__memory/is_pointer_accessible.h
```diff
@@ -0,0 +1,161 @@
...
+  int __is_managed             = 0;
+  ::CUmemoryPool __mempool     = nullptr;
+  void* __results[3]           = {&__memory_type, &__is_managed, &__mempool};
+  const auto __status2         = ::cuda::__driver::__pointerGetAttributesNoThrow(__attrs, __results, __p);
+  if (__status2 != ::cudaSuccess)
+  {
+    ::cuda::__throw_cuda_error(__status2, "is_host_accessible() failed", _CCCL_BUILTIN_PRETTY_FUNCTION());
+  }
+  // (1) check if the pointer is unregistered
+  if (__memory_type == static_cast<::CUmemorytype>(0))
```
davebayer: I think we should move the check if the pointer is managed here:
```suggestion
  if (__memory_type == static_cast<::CUmemorytype>(0) || __is_managed)
```

libcudacxx/include/cuda/__memory/is_pointer_accessible.h
```diff
@@ -0,0 +1,161 @@
...
+    return true;
+  }
+  // (2) check if a memory pool is associated with the pointer
+  if (__mempool != nullptr)
+  {
+    ::CUmemLocation __prop{::CU_MEM_LOCATION_TYPE_HOST, 0};
+    const auto __pool_flags = ::cuda::__driver::__mempoolGetAccess(__mempool, &__prop);
+    return (__pool_flags == ::CU_MEM_ACCESS_FLAGS_PROT_READ || __pool_flags == ::CU_MEM_ACCESS_FLAGS_PROT_READWRITE);
+  }
+  return (__is_managed || __memory_type == ::CU_MEMORYTYPE_UNIFIED || __memory_type == ::CU_MEMORYTYPE_HOST);
```
davebayer: I believe the check for unified memory type is not needed here. Also, the docs doesn't say that the `CU_MEMORYTYPE_UNIFIED` value can be returned from the function.
```suggestion
  return __memory_type == ::CU_MEMORYTYPE_HOST;
```

libcudacxx/include/cuda/__memory/is_pointer_accessible.h
```diff
@@ -0,0 +1,161 @@
...
+  {
+    ::cuda::__throw_cuda_error(__status2, "is_device_accessible() failed", _CCCL_BUILTIN_PRETTY_FUNCTION());
+  }
+  // (1) check if the pointer is unregistered
+  if (__memory_type == static_cast<::CUmemorytype>(0))
+  {
+    return false;
+  }
+  // (2) check if the pointer is a device accessible pointer or managed memory
+  if (!__is_managed && __memory_type != ::CU_MEMORYTYPE_DEVICE && __memory_type != ::CU_MEMORYTYPE_UNIFIED)
```
davebayer:
```suggestion
  if (!__is_managed && __memory_type != ::CU_MEMORYTYPE_DEVICE)
```

libcudacxx/include/cuda/__memory/is_pointer_accessible.h
```diff
@@ -0,0 +1,161 @@
...
+  {
+    return false;
+  }
+  // (3) check if a memory pool is associated with the pointer
+  if (__mempool != nullptr)
+  {
+    ::CUmemLocation __prop{::CU_MEM_LOCATION_TYPE_DEVICE, __device.get()};
+    const auto __pool_flags = ::cuda::__driver::__mempoolGetAccess(__mempool, &__prop);
+    return (__pool_flags == ::CU_MEM_ACCESS_FLAGS_PROT_READ || __pool_flags == ::CU_MEM_ACCESS_FLAGS_PROT_READWRITE);
```
davebayer:
```suggestion
    return (__pool_flags & ::CU_MEM_ACCESS_FLAGS_PROT_READWRITE) != 0;
```

davebayer (changes requested):
some more details

libcudacxx/include/cuda/__memory/is_pointer_accessible.h
fbusato: I was tempted to do the same, but I'm a bit worried about long term maintainability.

davebayer: Why? What's wrong with this? We are really only checking the read/write bits

libcudacxx/include/cuda/__memory/is_pointer_accessible.h
fbusato: we can move everything on top to avoid a driver call

libcudacxx/include/cuda/__memory/is_pointer_accessible.h
fbusato: what about if they change in the future?

fbusato: Oh, wait. I misinterpreted your comment.

fbusato: we cannot do that because
```cpp
CU_MEM_ACCESS_FLAGS_PROT_MAX         = 0x7FFFFFFF
```

davebayer: And what does this value mean, it's undocumented. The type has `_FLAGS_` in it's name, so I would expect it to behave like a flag type.

libcudacxx/include/cuda/__memory/is_pointer_accessible.h
```diff
@@ -0,0 +1,163 @@
...
+    ::CU_POINTER_ATTRIBUTE_MEMORY_TYPE,
+    ::CU_POINTER_ATTRIBUTE_IS_MANAGED,
+    ::CU_POINTER_ATTRIBUTE_DEVICE_ORDINAL,
+    ::CU_POINTER_ATTRIBUTE_MEMPOOL_HANDLE};
+  ::CUmemorytype __memory_type = static_cast<::CUmemorytype>(0);
+  int __is_managed             = 0;
+  int __ptr_dev_id             = 0;
+  ::CUmemoryPool __mempool     = nullptr;
+  void* __results[4]           = {&__memory_type, &__is_managed, &__ptr_dev_id, &__mempool};
+  const auto __status2         = ::cuda::__driver::__pointerGetAttributesNoThrow(__attrs, __results, __p);
```
davebayer:
```suggestion
  const auto __status          = ::cuda::__driver::__pointerGetAttributesNoThrow(__attrs, __results, __p);
```

libcudacxx/include/cuda/__memory/is_pointer_accessible.h
fbusato: > And what does this value mean, it's undocumented. The type has _FLAGS_ in it's name, so I would expect it to behave like a flag type.

my guess is that the flag is only used to mark the last value. Anyway, I would prefer the verbose but safer version

davebayer: Tell me, what's the risk here? Like when would be the max value returned?

The current approach is not correct.

libcudacxx/test/libcudacxx/cuda/memory/is_pointer_accessible.pass.cpp
```diff
@@ -0,0 +1,167 @@
...
+  test_accessible_pointer(ptr, is_host_accessible, is_device_accessible, is_managed_accessible, dev);
+}
+
+bool test_memory_pool()
+{
+  test_memory_pool_impl(cudaMemAllocationTypePinned, cudaMemLocationTypeHost, true, false, false);
+  test_memory_pool_impl(cudaMemAllocationTypePinned, cudaMemLocationTypeDevice, false, true, false);
+
+#if _CCCL_CTK_AT_LEAST(13, 0)
+  test_memory_pool_impl(cudaMemAllocationTypeManaged, cudaMemLocationTypeHost, true, false, true);
```
pciolkosz, line 131: The managed pool should be accessible everywhere I believe. The set access only affects the preferred location, but managed memory should always be accessible everywhere

libcudacxx/test/libcudacxx/cuda/memory/is_pointer_accessible.pass.cpp
```diff
@@ -0,0 +1,167 @@
...
+  test_memory_pool_impl(cudaMemAllocationTypeManaged, cudaMemLocationTypeDevice, false, true, true);
+#endif // _CCCL_CTK_AT_LEAST(13, 0)
+  return true;
+}
+
+bool test_multiple_devices()
+{
+  if (cuda::devices.size() < 2)
```
pciolkosz, line 139: It would be nice to test mempools in multi-gpu case too

libcudacxx/include/cuda/__memory/is_pointer_accessible.h
fbusato: why it is not correct? do you expect overlapping flags?

>  Like when would be the max value returned?

if max value is defined, it could be potentially used.

fbusato: anyway, I can use a stricter bitwise and check

libcudacxx/test/libcudacxx/cuda/memory/is_pointer_accessible.pass.cpp
fbusato: discussed offline. We will address the issue in future releases

libcudacxx/include/cuda/__memory/is_pointer_accessible.h
davebayer: > why it is not correct? do you expect overlapping flags?

Because if any other flag is added to the enum, your solution won't work, because you check for equality, but we need to check just the relevant bits, which are any of the bits `0` and `1` in our case.

> if max value is defined, it could be potentially used.

1. I strongly doubt that.
2. Then it's not our problem, because it had the relevant bits set here.

libcudacxx/include/cuda/__memory/is_pointer_accessible.h
```diff
@@ -0,0 +1,166 @@
...
+    return true;
+  }
+  // (2) check if a memory pool is associated with the pointer
+#  if _CCCL_CTK_AT_LEAST(12, 2)
+  if (__mempool != nullptr)
+  {
+    ::CUmemLocation __prop{::CU_MEM_LOCATION_TYPE_HOST, 0};
+    const unsigned __pool_flags = ::cuda::__driver::__mempoolGetAccess(__mempool, &__prop);
+    return !(__pool_flags & ~unsigned{::CU_MEM_ACCESS_FLAGS_PROT_READWRITE})
+        && (__pool_flags & unsigned{::CU_MEM_ACCESS_FLAGS_PROT_READ});
```
davebayer: I have no idea what this is trying to achieve. Please, don't overcomplicate this. From both the name and the enum definition it is clear that this type is a flag type where each bit represents a specific memory access.
```cpp
CU_MEM_ACCESS_FLAGS_PROT_NONE      = 0x0, // 0b00
CU_MEM_ACCESS_FLAGS_PROT_READ      = 0x1, // 0b01
CU_MEM_ACCESS_FLAGS_PROT_READWRITE = 0x3, // 0b11
CU_MEM_ACCESS_FLAGS_PROT_MAX       = 0x7FFFFFFF
```
Bit at position `0` controls the read access, bit at position `1` controls the write access. The `_MAX` value probably exists so the C++ compiler is forced to make this enum based on `int` and not any smaller type.

fbusato: ook, let do it. Hopefully `_MAX` will never be used.

PR #6332, opened 2025-10-23
Fix transform.cu size_t/int issue.

tpn:
This PR does two things:

1. Ensures we build with `/WX` (warning-as-error) in our default Windows CI configuration (MSVC, CMake Visual Studio generator).
2. Fixes a size_t/int warning in transform.cu that causes the build to fail due to `/WX`.

### DIFF
--- a/c/parallel/src/transform.cu
+++ b/c/parallel/src/transform.cu
@@ -209,7 +209,7 @@ struct transform_kernel_source
   }
 
 private:
-  static auto is_pointer_aligned(const indirect_iterator_t& it, int alignment)
+  static auto is_pointer_aligned(const indirect_iterator_t& it, ::cuda::std::size_t alignment)
   {
     return it.value_size != 0 && ::cuda::is_aligned(*static_cast<char**>(it.ptr), alignment);
   }


leofang (approved):

c/parallel/CMakeLists.txt
```diff
@@ -63,6 +63,13 @@ target_compile_options(cccl.c.parallel PRIVATE $<$<COMPILE_LANG_AND_ID:CUDA,NVID
 target_include_directories(cccl.c.parallel PUBLIC "include")
 target_include_directories(cccl.c.parallel PRIVATE "src")
 
+# Ensure warnings-as-errors are applied on MSVC builds for this target when
+# requested, regardless of generator or configuration type.
+if(CCCL_ENABLE_WERROR AND CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
```
miscco: I am torn whether we should just always build with /WX, we at least do that on the C++ side

tpn: I think we should.  Short term pain (depending on how many current warnings there are), but better in the long run.

alliepiper: I don't think this will work. `cccl.c.parallel` is already linked to `cccl.compiler_interface_cpp20`, which includes these flags. However, these are C targets, so the genexes like `COMPILE_LANGUAGE:CXX` disable these flags -- and the ones below will fail for the same reason.

I'll update the logic that builds the compiler interface targets to include C flags as well.

alliepiper (approved):
Changing to approved to unblock merging the fix. Please remove the CMake updates as they won't work as intended -- I'll work on this in https://github.com/NVIDIA/cccl/pull/6340.

c/parallel/CMakeLists.txt
alliepiper: I take that back -- this project only contains cuda and cxx objects:

```
file(GLOB_RECURSE srcs
  RELATIVE "${CMAKE_CURRENT_LIST_DIR}"
  CONFIGURE_DEPENDS
  "src/*.cu" "src/*.cpp"
)
```

So C language flags aren't required. Still, we should already be passing `/WX` etc as part of the cccl compiler interface target for CUDA and CXX targets. I'll do some more investigation -- adding them here should not be needed.

c/parallel/CMakeLists.txt
alliepiper, line 43: This should be bringing the new flags in already.

c/parallel/CMakeLists.txt
alliepiper: To update this from internal conversations, the issue is that the windows python builds use the Visual Studio CMake generator. Our CI and build system have been designed around single-config CMake generators (make, ninja), and using the multiconfig generators in our CI would require additional complexity and maintenance / testing cost without providing any meaningful benefit.

While our projects should still function if a user decides to use our packages with a multiconfig CMake generator, we should stick with Ninja in our CI and other internal testing to make sure that the expected flags, etc (including `/WX`) are added.

PR #6335, opened 2025-10-24
Move the `throw_meow_error` functions into their own header and drop the `stdexcept` include

miscco:
We want this to be an internal header


davebayer (approved):

davebayer:
pre-commit.ci autofix

bernhardmgruber (approved):

PR #6336, opened 2025-10-24
Move the implementation of `tuple` into its own file

miscco:
Modernize and modularize a lot of the tuple implementation


libcudacxx/include/cuda/std/__ranges/movable_box.h
```diff
@@ -141,7 +141,7 @@ struct __mb_optional_copy_assign : __mb_optional_destruct_base<_Tp>
 {
   _LIBCUDACXX_DELEGATE_CONSTRUCTORS(__mb_optional_copy_assign, __mb_optional_destruct_base, _Tp);
 
-  _CCCL_HIDE_FROM_ABI constexpr __mb_optional_copy_assign(const __mb_optional_copy_assign&) = default;
```
davebayer, line 144: It wouldn't be your PR without fixing something in movable box :)

libcudacxx/include/cuda/std/__tuple_dir/tuple.h
```diff
@@ -0,0 +1,321 @@
...
+  _CCCL_API explicit constexpr tuple() noexcept(_Constraints::__nothrow_default_constructible)
+  {}
+
+  _CCCL_HIDE_FROM_ABI tuple(tuple const&) = default;
+  _CCCL_HIDE_FROM_ABI tuple(tuple&&)      = default;
+
+  template <class _Alloc,
+            class _Constraints                                               = __tuple_constraints<_Tp...>,
+            enable_if_t<_Constraints::__implicit_default_constructible, int> = 0>
+  _CCCL_API inline tuple(allocator_arg_t, _Alloc const& __a) noexcept(_Constraints::__nothrow_default_constructible)
```
davebayer, line 99: Don't we want to use our concept macros here?

miscco: I am wondering about that too

But I wanted to keep the commits that move stuff relatively clean

bernhardmgruber: Yes, please just focus on moving code in this PR and not introducing any new changes.

bernhardmgruber (approved):

libcudacxx/include/cuda/std/__tuple_dir/apply.h
```diff
@@ -0,0 +1,82 @@
...
+template <class _Fn, class... _Types>
+inline constexpr bool __can_apply_impl<_Fn, __tuple_types<_Types...>> = is_invocable_v<_Fn, _Types...>;
+
+template <class _Fn, class _Tuple, bool = __tuple_like<_Tuple>>
+inline constexpr bool __can_apply = false;
+
+template <class _Fn, class _Tuple>
+inline constexpr bool __can_apply<_Fn, _Tuple, true> = __can_apply_impl<_Fn, __make_tuple_types_t<_Tuple>>;
+
+#define _LIBCUDACXX_NOEXCEPT_RETURN(...) \
```
davebayer, line 48: Don't you want to use this macro to the config? It could be used in other places, too

miscco: Actually I want to get rid of it but I want to do that in a followup PR, because this is already sketch with the minor changes I did

PR #6337, opened 2025-10-24
Improve exception macros

davebayer:


libcudacxx/include/cuda/std/__cccl/exceptions.h
```diff
@@ -25,67 +25,12 @@
 
 #if defined(CCCL_DISABLE_EXCEPTIONS) // Escape hatch for users to manually disable exceptions
 #  define _CCCL_HAS_EXCEPTIONS() 0
-#elif _CCCL_COMPILER(NVRTC) // NVRTC has no exceptions
+#elif _CCCL_DEVICE_COMPILATION() && !_CCCL_CUDA_COMPILER(NVHPC) // nvc++ traps if an exception is thrown in device code
```
davebayer: @ericniebler is this change fine? Won't I break the cudax constexpr exceptions? I don't know if they can be thrown in device code

davebayer: Looks like it breaks cudax.. How should `constexpr` exceptions behave when compiling device code? Should they just be disabled?

davebayer: Discussed internally, we kept the original definition.

docs/cccl/development/macro.rst
```diff
@@ -330,6 +330,77 @@ Usage example:
 | ``_CCCL_PRAGMA_NOUNROLL()``    | Portable ``#pragma nounroll`` pragma      |
 +--------------------------------+-------------------------------------------+
 
+**Exception Macros**
+
+CUDA doesn't support exceptions in device code, however, sometimes we need to write host/device functions that use exceptions on host and ``__trap()`` on device. CCCL provides set of macros that should be used in place of the standard C++ keywords to make the code compile.
```
bernhardmgruber:
```suggestion
CUDA doesn't support exceptions in device code, however, sometimes we need to write host/device functions that use exceptions on host and ``__trap()`` on device. CCCL provides a set of macros that should be used in place of the standard C++ keywords to make the code compile in both, host and device code.
```

docs/cccl/development/macro.rst
```diff
@@ -330,6 +330,77 @@ Usage example:
...
+| ``_CCCL_CATCH_ALL``        | Replacement for the ``catch (...)`` statement                     |
++----------------------------+-------------------------------------------------------------------+
+| ``_CCCL_CATCH_FALLTHOUGH`` | End of ``try``/``catch`` block if ``_CCCL_CATCH_ALL`` is not used |
++----------------------------+-------------------------------------------------------------------+
+| ``_CCCL_THROW``            | Replacement for the ``throw /*arg*/`` expression                  |
++----------------------------+-------------------------------------------------------------------+
+| ``_CCCL_RETHROW``          | Replacement for the plain ``throw`` expression                    |
++----------------------------+-------------------------------------------------------------------+
+
+*Note*: The ``catch`` clause must always bind to a named variable. If omitted, there will be an expected identifier error.
```
bernhardmgruber: I don't understand what the `catch` clause is. Do you mean the argument to `_CCCL_CATCH`? Then maybe:
```
The ``_CCCL_CATCH`` clause must always introduce a named variable, like: ``_CCCL_CATCH(const exception_type& var)``

File: docs/cccl/development/macro.rst [hunk 4]
```diff
@@ -330,6 +330,77 @@ Usage example:
...
+    {
+        int* buff{};
+
+        _CCCL_TRY
+        {
+            buff = reinterpret_cast<int*>(alloc(n * sizeof(int)));
+
+            do_something(buff);
+        }
+        _CCCL_CATCH ([[maybe_unused]] std::bad_alloc& e) // must be always named
```
bernhardmgruber: I have learned a long time ago that exceptions should be caught by `const` reference. Is there a reason to not do that here?
```suggestion
        _CCCL_CATCH ([[maybe_unused]] const std::bad_alloc& e) // must be always named
```

docs/libcudacxx/extended_api/streams/stream_ref.rst
```diff
@@ -10,7 +10,7 @@ However, as this is just an alias for a plain pointer type it carries with it a
 conversions from e.g ``nullptr`` or a literal ``0``.
 
 These hard to spot bugs can be avoided through ``cuda::stream_ref``, which is a simple wrapper around a ``cudaStream_t``
-that prevents implicit conversions. It also provides the ``wait()`` and ``ready()`` member functions to facilitate
+that prevents implicit conversions. It also provides the ``wait()`` and ``is_done()`` member functions to facilitate
```
bernhardmgruber: Is this related to exception macros? If not, can you please file the changes in this file as a separate PR? Thx!

bernhardmgruber (approved):

docs/libcudacxx/extended_api/streams/stream_ref.rst
davebayer: Moved these changes to #6376

docs/cccl/development/macro.rst
davebayer: There isn't, it's a typo thanks

docs/cccl/development/macro.rst
davebayer: Yeah, makes much more sense

PR #6341, opened 2025-10-25
c.parallel: single-stage runtime compilation.

### LINKED ISSUE #5366: c.parallel: investigate making ptx-json generate valid PTX
Currently, ptx-json emits invalid PTX, which means that regardless of whether we need to do a two C++ frontend pass compilation or not when building kernels, we have to do it. #5355 serves as a proof of concept for extracting data from PTX generated from LTOIR, and adapting that technique in c.parallel would allow us to eliminate one of the C++ frontend passes, speeding the build step considerably. Investigate making this possible.

griwes:
## Description

This PR replaces the current two-stage approach to obtaining tuning policies in c.parallel with a single stage approach, by switching the JSON output from appearing in the PTX output to appearing as a string in a CUBIN.

Timing comparison on my machine (ignoring `for` and `histogram`, as they currently don't use dynamic policies):

<table>
<tr>
<th>Algorithm</th>
<th>Before</th>
<th>After</th>
</tr>
<tr>
<td>Merge sort</td>
<td>119.69 sec</td>
<td>93.02 sec</td>
</tr>
<tr>
<td>Radix sort</td>
<td>54.74 sec</td>
<td>49.32 sec</td>
</tr>
<tr>
<td>Reduce</td>
<td>81.17 sec</td>
<td>60.78 sec</td>
</tr>
<tr>
<td>Scan</td>
<td>121.72 sec</td>
<td>92.77 sec</td>
</tr>
<tr>
<td>Segmented reduce</td>
<td>84.78 sec</td>
<td>72.65 sec</td>
</tr>
<tr>
<td>Three way partition</td>
<td>35.36 sec</td>
<td>27.78 sec</td>
</tr>
<tr>
<td>Transform</td>
<td>95.37 sec</td>
<td>70.11 sec</td>
</tr>
<tr>
<td>Unique by key</td>
<td>88.01 sec</td>
<td>66.95 sec</td>
</tr>
</table>

Resolves #5366

## Checklist


griwes:
I'd be very surprised if anything changed in the selected policies, because the policy hub expressions did not actually change from what's in main. The only real difference is that we are now compiling them without -default-device.

griwes:
The CI failures were because I was too optimistic that I got the changes to ptx_json test flags right without testing locally :sweat_smile:

griwes:
I believe the remaining CI failures are unrelated to this PR.

bernhardmgruber:
> I believe the remaining CI failures are unrelated to this PR.

#6357 will save us

alliepiper (approved):
CMake lgtm, so long as this has been tested with `-lineinfo` (that was breaking the compute sanitizer jobs).

griwes:
@alliepiper this new approach doesn't do any of the funky codegen stuff the old approach was doing, so I see no way for it to break with debug info.

PR #6329, opened 2025-10-23
Improve `cuda::barrier` TMA examples and `elect_one` in `DeviceTransform`

bernhardmgruber:
@ahendriksen is back from parental leave, so here are some new micro optimizations. These are especially beneficial for small problem sizes where the time-to-UBLKCP in SASS matters a lot:

`cub.bench.transform.babelstream.base` on B200:
```c++
# mul

## [0] NVIDIA B200

|  T{ct}  |  OffsetT{ct}  |  Elements{io}  |   Ref Time |   Ref Noise |   Cmp Time |   Cmp Noise |      Diff |   %Diff |  Status  |
|---------|---------------|----------------|------------|-------------|------------|-------------|-----------|---------|----------|
|   I8    |      I32      |      2^16      |   6.312 us |       5.35% |   6.354 us |       6.04% |  0.042 us |   0.67% |   SAME   |
|   I8    |      I32      |      2^20      |   7.812 us |       6.67% |   8.211 us |       3.30% |  0.400 us |   5.12% |   SLOW   |
|   I8    |      I32      |      2^24      |  15.509 us |       6.09% |  15.538 us |       6.47% |  0.028 us |   0.18% |   SAME   |
|   I8    |      I32      |      2^28      | 123.889 us |       0.15% | 123.440 us |       0.69% | -0.450 us |  -0.36% |   FAST   |
|   I8    |      I64      |      2^16      |   7.163 us |       1.51% |   6.233 us |       3.25% | -0.929 us | -12.97% |   FAST   |
|   I8    |      I64      |      2^20      |   8.043 us |       4.96% |   8.119 us |       5.99% |  0.076 us |   0.95% |   SAME   |
|   I8    |      I64      |      2^24      |  15.946 us |       4.92% |  15.768 us |       5.87% | -0.178 us |  -1.11% |   SAME   |
|   I8    |      I64      |      2^28      | 124.553 us |       0.76% | 125.065 us |       0.33% |  0.512 us |   0.41% |   SLOW   |
|   I16   |      I32      |      2^16      |   7.153 us |       1.55% |   6.270 us |       5.34% | -0.883 us | -12.34% |   FAST   |
|   I16   |      I32      |      2^20      |   8.181 us |       2.59% |   8.218 us |       2.10% |  0.037 us |   0.45% |   SAME   |
|   I16   |      I32      |      2^24      |  18.448 us |       1.61% |  18.493 us |       0.80% |  0.044 us |   0.24% |   SAME   |
[...]

PR #6344, opened 2025-10-27
Apply cuda::barrier and elect_one feedback

bernhardmgruber:
This is a follow-up to #6329 after feedback from @ahendriksen

### DIFF
--- a/libcudacxx/include/cuda/__memcpy_async/elect_one.h
+++ b/libcudacxx/include/cuda/__memcpy_async/elect_one.h
@@ -22,9 +22,7 @@
 #  pragma system_header
 #endif // no system header
 
-#if _CCCL_CUDA_COMPILATION() && __cccl_ptx_isa >= 800
-#  include <cuda/__ptx/instructions/elect_sync.h>
-#endif // _CCCL_CUDA_COMPILATION() && __cccl_ptx_isa >= 800
+#include <cuda/__ptx/instructions/elect_sync.h>
 
 #include <cuda/std/__cccl/prologue.h>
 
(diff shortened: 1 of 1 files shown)


libcudacxx/include/cuda/__memcpy_async/elect_one.h
```diff
@@ -22,37 +22,36 @@
...
-  NV_IF_ELSE_TARGET(
+#if _CCCL_CUDA_COMPILATION() && __cccl_ptx_isa >= 800
+  NV_IF_TARGET(
     NV_PROVIDES_SM_90,
     (const auto tid             = threadIdx.x; //
      const auto warp_id         = tid / 32;
      const auto uniform_warp_id = __shfl_sync(~0, warp_id, 0); // broadcast from lane 0
      return uniform_warp_id == 0 && cuda::ptx::elect_sync(~0); // elect a leader thread among warp 0
-     ),
-    (::cuda::device::__cuda_elect_sync_is_not_supported_before_SM_90__(); _CCCL_UNREACHABLE();));
```
ahendriksen, line 48: Shouldn't the 
```
  return threadIdx.x == 0;
```
go in the else here? 

AFAIK PTX ISA is CTK 12.0. CCCL support 12.0 and up, so there is no need for the `__cccl_ptx_isa` ifdef.

bernhardmgruber: Then I would need to add it twice, as else branch of the `NV_IF_TARGET` and as else branch of the `_CCCL_CUDA_COMPILATION() && __cccl_ptx_isa >= 800`

Regarding PTX ISA. I don't know whether the clang CUDA versions we test already support __cccl_ptx_isa >= 800

bernhardmgruber: ok, let's just try

PR #6346, opened 2025-10-27
Various CMake cleanups.

alliepiper:
1. Replace `cn_` with `cudax_` in various places (cudax used to be "cuda next", these are just leftovers from a bygone era).
2. Ensure that project-wide compiler targets are linked into tests and examples so they build with expected flags.
3. Add missing `cccl_configure_target` that configure things like extra dialect settings, binary output locations, etc.
4. Remove unnecessary CMAKE_CUDA_ARCHITECTURES property setting.
5. Build CMake cudax example with 'native' by default.
6. Fix some docstrings.
7. Remove checks for obsolete architectures.

### DIFF
--- a/nvbench_helper/nvbench_helper/nvbench_helper.cu
+++ b/nvbench_helper/nvbench_helper/nvbench_helper.cu
@@ -16,7 +16,7 @@
 #include <thrust/tabulate.h>
 
 #include <cuda/functional>
-#include <cuda/std/__bit/bit_cast.h>
+#include <cuda/std/bit>
 
 #include <cstdint>
 #include <random>


cudax/cmake/cudaxSTFConfigureTarget.cmake
```diff
@@ -6,7 +6,6 @@ function(cudax_stf_configure_target target_name)
   cmake_parse_arguments(CSCT "${options}" "${oneValueArgs}" "${multiValueArgs}" ${ARGN})
 
   target_link_libraries(${target_name} PRIVATE
-    ${cn_target}
```
alliepiper, line 9: This was undefined here, and is handled at a higher scope.

libcudacxx/cmake/LibcudacxxPublicHeaderTesting.cmake
```diff
@@ -40,16 +32,6 @@ function(libcudacxx_create_public_header_test header_name headertest_src)
...
-  # Ensure that if this is an atomic header, we only include the right architectures
-  string(REGEX MATCH "atomic|barrier|latch|semaphore|annotated_ptr|pipeline" match "${header}")
-  if(match)
-    # Ensure that we only compile the header when we have some architectures enabled
-    if (NOT architectures_at_least_sm70)
-      return()
-    endif()
-    set_target_properties(public_headertest_${header_name} PROPERTIES CUDA_ARCHITECTURES "${architectures_at_least_sm70}")
-  endif()
-
```
miscco, line 52: I am not in favor of those changes, there is a high probability that someone will run our tests against a pre-sm70 architecture

This is not doing anything currently, but we should keep it until no supported CTK supports SM70

alliepiper: @jrhemstad Thoughts on this? AIUI, SMs <75 don't exist as far as CCCL is concerned.

jrhemstad: @miscco shouldn't all of these headers be includeable in a TU compiled for <sm70 anyways such that there's no need to filter them out anymore?

miscco: We still error out in e.g barrier
> #if _CCCL_DEVICE_COMPILATION() && _CCCL_PTX_ARCH() < 700 && !_CCCL_CUDA_COMPILER(NVHPC)
> #  error "CUDA synchronization primitives are only supported for sm_70 and up."
> #endif // _CCCL_DEVICE_COMPILATION() && _CCCL_PTX_ARCH() < 700 && !_CCCL_CUDA_COMPILER(NVHPC)

miscco: What a formatting ^^

alliepiper: AIUI, we no longer support sm < 75 as of CCCL 3+, and will not test these -- even on older CTK. This is explicitly called out in the README:

https://github.com/NVIDIA/cccl?tab=readme-ov-file#gpu-architectures

> To be clear, while CCCL can be compiled with both the current and previous CTK major versions, we do not test or validate architectures that were only supported in the older CTK.
>
> For example, CCCL 3.0 supports compiling with CTK 12.x and 13.x where
>
> - CUDA Toolkit 13.x supports >=sm_75
> - CUDA Toolkit 12.x supports >=sm_50
>
> In this scenario, compiling CCCL 3.0 with CTK 12.x targeting architectures below sm_75 may work, but those configurations are not part of our regular testing.

Either way, I restored the filtering logic in CMake to unblock this PR.

bernhardmgruber (approved):

nvbench_helper/nvbench_helper/nvbench_helper.cu
```diff
@@ -16,6 +16,7 @@
 #include <thrust/tabulate.h>
 
 #include <cuda/functional>
+#include <cuda/std/__bit/bit_cast.h>
```
miscco: Nitpick: this is a public test file, so we should use 
```suggestion
#include <cuda/std/bit>
```

miscco (approved):

PR #6347, opened 2025-10-27
Retry sccache startup on windows to WAR random auth issues.

alliepiper:


trxcllnt:
@alliepiper Any idea why [this file](https://github.com/NVIDIA/cccl/actions/runs/18857541169/job/53825885949#step:4:2271) compiles OK here, but [fails](https://github.com/NVIDIA/cccl/actions/runs/18858176086/job/53813524359#step:4:2375) in the sccache-dist PR? As best I can tell, they should be identical.

PR #6348, opened 2025-10-27
Fix return value of scan in OMP backend

### LINKED ISSUE #6317: [BUG]: stdpar multicore scan tests fail at runtime after commit 1238f287f4f654bff12e4d4a29b7dfa1c40a12fe
### Is this a duplicate?

### Type of Bug

Runtime Error

### Component

Thrust

### Describe the bug

stdpar multicore scan tests have started failing in correctness after the following commit: 1238f287f4f654bff12e4d4a29b7dfa1c40a12fe
For this example:
```
#include <execution>
#include <algorithm>
#include <numeric>
#include <cassert>

constexpr int N = 10;

int main() {
  int* data = new int[N];
  int* output = new int[N];
  std::iota(data, data + N, 0);
  auto r0 = std::exclusive_scan(std::execution::par, data, data + N, output, 10);
  assert(r0 == output + N);
  delete[] output;
  delete[] data;
}
```
The program fails at runtime with the following assertion error:
```
int main(): Assertion `r0 == output + N' failed.
```

### How to Reproduce

```
nvc++ -stdpar=multicore -fast --c++17 scan.cpp
```

### Expected behavior

The program should pass the assertion (r0 == output + N) and complete successfully.

### Reproduction link

_No response_

### Operating System

_No response_

### nvidia-smi output

_No response_

### NVCC version

_No response_

miscco:
Fixes #6317

### DIFF
--- a/thrust/thrust/system/omp/detail/scan.inl
+++ b/thrust/thrust/system/omp/detail/scan.inl
@@ -77,7 +77,7 @@ OutputIterator scan_impl(
     {
       ::cuda::std::exclusive_scan(first, last, result, init, binary_op);
     }
-    return result;
+    return result + n;
   }
 
   thrust::detail::temporary_array<ValueType, DerivedPolicy> block_sums(exec, num_threads);


bernhardmgruber:
> Please also add a test for this. Thx!

We do have tests for this case though. It seems there is a bug in the dispatching when the device system is OMP the unit tests run the sequential sort and not the OMP one.

bernhardmgruber:
#6359 is the reason our unit tests did not discover this issue

bernhardmgruber:
I would like to supersede this PR with #6361

PR #6351, opened 2025-10-27
[cudax->libcu++] Move any_resource tests and remove experimental aliases

pciolkosz:
`any_resource` and the family was moved to libcu++ in https://github.com/NVIDIA/cccl/pull/6299, this PR moves the tests and removes the experimental alias. I also had to adjust one test to not use a cudax type and added some checks.


libcudacxx/include/cuda/__memory_resource/any_resource.h
```diff
@@ -217,7 +217,7 @@ struct _CCCL_DECLSPEC_EMPTY_BASES any_synchronous_resource
       : __base(::cuda::std::move(__other.__get_base()))
   {}
 
-  using default_queries = properties_list<_Properties...>;
+  using default_queries = ::cuda::mr::properties_list<_Properties...>;
```
davebayer, line 220: Q: shouldn't this just be called `property_list`?

PR #6352, opened 2025-10-27
[cudax] Add pointer attributes fallback to async buffer initialization

pciolkosz:
Currently we use device side initialization when buffer is both host and device accessible. This is probably a mistake for host pinned memory and we should use host side instead, but it's not possible to tell that case apart from managed memory case. This PR adds pointer attribute check when buffer is both host and device accessible and if its not managed memory the host path is used instead.

### DIFF
--- a/cudax/include/cuda/experimental/__container/async_buffer.cuh
+++ b/cudax/include/cuda/experimental/__container/async_buffer.cuh
@@ -14,15 +14,15 @@
 #include <cuda/std/detail/__config>
 
 #if defined(_CCCL_IMPLICIT_SYSTEM_HEADER_GCC)
-#pragma GCC system_header
+#  pragma GCC system_header
 #elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_CLANG)
-#pragma clang system_header
+#  pragma clang system_header
 #elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_MSVC)
-#pragma system_header
+#  pragma system_header
 #endif // no system header
 
 #if _CCCL_HAS_CUDA_COMPILER()
-#include <cub/device/device_transform.cuh>
+#  include <cub/device/device_transform.cuh>
 #endif // _CCCL_HAS_CUDA_COMPILER()
 
 #include <cuda/__container/heterogeneous_iterator.h>
@@ -54,8 +54,10 @@
 #include <cuda/std/__cccl/prologue.h>
 
 //! @file The \c buffer class provides a container of contiguous memory
-namespace cuda::experimental {
-enum class __memory_accessability {
+namespace cuda::experimental
+{
+enum class __memory_accessability
+{
   __device,
   __host,
   __device_and_host,
@@ -64,20 +66,20 @@ enum class __memory_accessability {
 // This assumes one of host or device accessible properties is present, needs to
 // be updated when we relax that
 template <typename... _Properties>
-struct __memory_accessability_from_properties {
+struct __memory_accessability_from_properties
+{
   static constexpr __memory_accessability value =
-      ::cuda::mr::__is_device_accessible<_Properties...>
-          ? ::cuda::mr::__is_host_accessible<_Properties...>
-                ? __memory_accessability::__device_and_host
-                : __memory_accessability::__device
-          : __memory_accessability::__host;
+    ::cuda::mr::__is_device_accessible<_Properties...>
+      ? ::cuda::mr::__is_host_accessible<_Properties...>
+        ? __memory_accessability::__device_and_host
+        : __memory_accessability::__device
+      : __memory_accessability::__host;
 };
 
 // Once we add support from options taken from the env we can list them here in
 // addition to using is_same_v
 template <class _Env>
-inline constexpr bool __buffer_compatible_env =
-    ::cuda::std::is_same_v<_Env, ::cuda::std::execution::env<>>;
+inline constexpr bool __buffer_compatible_env = ::cuda::std::is_same_v<_Env, ::cuda::std::execution::env<>>;
 
 //! @rst
 //! .. _cudax-containers-async-vector:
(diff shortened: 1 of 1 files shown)


cudax/include/cuda/experimental/__container/async_buffer.cuh
```diff
@@ -592,7 +611,28 @@ __fill_n(cuda::stream_ref __stream, _Tp* __first, ::cuda::std::size_t __count, c
...
 
-  if constexpr (_IsHostOnly)
+  // We don't know what to do with both device and host accessible buffers, so we need to check the attributes
+  if constexpr (_Accessability == __memory_accessability::device_and_host)
+  {
+    __driver::__pointer_attribute_value_type_t<CU_POINTER_ATTRIBUTE_MEMORY_TYPE> __type;
+    bool __is_managed{};
+    auto __status1 = ::cuda::__driver::__pointerGetAttributeNoThrow<CU_POINTER_ATTRIBUTE_MEMORY_TYPE>(__type, __first);
+    auto __status2 =
+      ::cuda::__driver::__pointerGetAttributeNoThrow<CU_POINTER_ATTRIBUTE_IS_MANAGED>(__is_managed, __first);
```
davebayer: Can we use [`cuGetPointerAttributes`](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__UNIFIED.html#group__CUDA__UNIFIED_1gf65e9ea532e311dd049166e4894955ad) to get both values at the same time?

cudax/include/cuda/experimental/__container/async_buffer.cuh
```diff
@@ -57,6 +57,25 @@
...
+template <typename... _Properties>
+struct __memory_accessability_from_properties
+{
+  static constexpr __memory_accessability value =
+    ::cuda::mr::__is_device_accessible<_Properties...>
+      ? ::cuda::mr::__is_host_accessible<_Properties...>
+        ? __memory_accessability::device_and_host
+        : __memory_accessability::device
+      : __memory_accessability::host;
+};
```
davebayer, line 77: Let's use a template variable here

pciolkosz: It started as a variable template, but I can't use `properties_list::rebind` on it, had to change it to a struct

cudax/include/cuda/experimental/__container/async_buffer.cuh
```diff
@@ -674,7 +714,7 @@ async_buffer<_Tp, _Properties...> make_async_buffer(
   [[maybe_unused]] const _Env& __env = {})
 {
   auto __res = async_buffer<_Tp, _Properties...>{__stream, __mr, __size, no_init};
-  __fill_n<_Tp, !::cuda::mr::__is_device_accessible<_Properties...>>(
+  __fill_n<_Tp, __memory_accessability_from_properties<_Properties...>::value>(
```
jrhemstad, line 750: question: Will the use of a kernel or host initialization eventually be controllable via the `env`?

pciolkosz: Yes, we should have a way to override whatever default we pick

cudax/include/cuda/experimental/__container/async_buffer.cuh
pciolkosz: Lets table it for now, its a minor optimization

cudax/include/cuda/experimental/__container/async_buffer.cuh
```diff
@@ -591,7 +610,28 @@ __fill_n(cuda::stream_ref __stream, _Tp* __first, ::cuda::std::size_t __count, c
     return;
   }
 
-  if constexpr (_IsHostOnly)
+  // We don't know what to do with both device and host accessible buffers, so we need to check the attributes
+  if constexpr (_Accessability == __memory_accessability::device_and_host)
```
miscco: I would prefer if this would be 
```suggestion
  if constexpr ( ::cuda::mr::__is_device_accessible<_Properties...> && ::cuda::mr::__is_host_accessible<_Properties...>)
```

The additional enumeration does not add anything here

cudax/include/cuda/experimental/__container/async_buffer.cuh
```diff
@@ -591,7 +610,28 @@ __fill_n(cuda::stream_ref __stream, _Tp* __first, ::cuda::std::size_t __count, c
...
+    if (__type == ::CU_MEMORYTYPE_HOST && !__is_managed)
+    {
+      __fill_n<_Tp, __memory_accessability::host>(__stream, __first, __count, __value);
+    }
+    else
+    {
+      __fill_n<_Tp, __memory_accessability::device>(__stream, __first, __count, __value);
+    }
+  }
+  else if constexpr (_Accessability == __memory_accessability::host)
```
miscco: Ditto this could just be
```suggestion
  else if constexpr (::cuda::mr::__is_host_accessible<_Properties...>)
```

cudax/include/cuda/experimental/__container/async_buffer.cuh
pciolkosz: I don't have `Properties` pack here and it's hard to provide it, because it might come from the `properties_list` structure. But I later realized this enum is sort of redundant with `_ExecutionSpace` we already have, I am just moving it in https://github.com/NVIDIA/cccl/pull/6489/ I will unify them after that PR gets merged

cudax/include/cuda/experimental/__container/async_buffer.cuh
```diff
@@ -56,6 +56,25 @@
 //! @file The \c async_buffer class provides a container of contiguous memory
 namespace cuda::experimental
 {
+enum class __memory_accessability
+{
+  device,
+  host,
+  device_and_host,
```
davebayer: should be uglified
```suggestion
  __device,
  __host,
  __device_and_host,
```

miscco (approved):

PR #6353, opened 2025-10-27
Support nested structs in `cuda.compute`

shwina:
## Description

Closes https://github.com/NVIDIA/cccl/issues/6633.

This PR enables arbitrary nesting of `gpu_struct` and `ZipIterator`. Example usage:

```python
# Define nested structs
@gpu_struct
class Stats:
    count: np.int32
    sum: np.float32


@gpu_struct
class DataPoint:
    value: np.int64
    stats: Stats


def sum_with_tuples(d1, d2):
    return DataPoint(
        d1.value + d2.value,
        # Tuple syntax for constructing the nested Stats struct
        (d1.stats.count + d2.stats.count, d1.stats.sum + d2.stats.sum),
    )

...

cuda.compute.reduce_into(d_input, d_output, sum_with_tuples, num_items, h_init)
```

Other examples are included in the PR and can be viewed [here](https://github.com/shwina/cccl/tree/nested-structs/python/cuda_cccl/tests/compute/examples/struct).

Broadly speaking, the changes in this PR are:

* generalization of `gpu_struct` to allow nesting
  * includes ability to construct gpu_structs from dicts ("JSON-like syntax ")
* generalization of `ZipIterator` to allow nesting
* reorganization of code in `struct.py` to cleanly separate "bare" struct creation (for internal use) and adding Python-side features (for convenience/end-user)
* Added test and examples

## Checklist

### DIFF
--- a/python/cuda_cccl/cuda/compute/iterators/_zip_iterator.py
+++ b/python/cuda_cccl/cuda/compute/iterators/_zip_iterator.py
@@ -92,7 +92,7 @@ def {func_name}(context, struct_ptr_type):
     field_type = struct_model._members[{field_idx}]
     return types.CPointer(field_type)(struct_ptr_type), codegen
 """
-        # Execute the code to create the intrinsic function in global namespace
+        # Execute the code to create the intrinsic function in local namespace
         exec(intrinsic_code, local_ns)
 
     # Now we can define the advance and dereference methods of this iterator,
--- a/python/cuda_cccl/cuda/compute/struct.py
+++ b/python/cuda_cccl/cuda/compute/struct.py
@@ -298,7 +298,7 @@ def make_struct_type(name, field_names, field_types):
             return
 
         # Compile-time constant index
-        if isinstance(idx, (types.IntegerLiteral, types.Literal)):
+        if isinstance(idx, (types.IntegerLiteral)):
             idx_val = getattr(idx, "literal_value", getattr(idx, "value", None))
 
             if idx_val is None or not (0 <= idx_val < len(field_names_list)):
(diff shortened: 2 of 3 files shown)


python/cuda_cccl/cuda/compute/iterators/_zip_iterator.py
```diff
@@ -89,21 +95,19 @@ def codegen(context, builder, sig, args):
     return types.CPointer(field_type)(struct_ptr_type), codegen
 """
         # Execute the code to create the intrinsic function in global namespace
```
NaderAlAwar: Important: this comment needs to be updated

python/cuda_cccl/cuda/compute/typing.py
```diff
@@ -29,9 +26,11 @@ class StreamLike(Protocol):
     def __cuda_stream__(self) -> tuple[int, int]: ...
 
 
-# TODO: type GpuStruct appropriately. It should be any type that has
-# been decorated with `@gpu_struct`.
-GpuStruct = Any
-GpuStruct.__doc__ = """\
-    Type of instances of classes decorated with @gpu_struct.
-"""
+class GpuStruct(Protocol):
```
shwina, line 29: Finally, we have proper typing for `GpuStruct`!

python/cuda_cccl/cuda/compute/struct.py
```diff
@@ -53,96 +250,102 @@ def can_convert_from(self, typingctx, other):
...
-        # Add the error case
-        impl_lines.append("    else:")
-        impl_lines.append("        raise IndexError('Index out of range')")
-
-        impl_code = f"""
-def struct_getitem_impl(struct_val, idx):
-{chr(10).join(impl_lines)}
-"""
+        # Compile-time constant index
+        if isinstance(idx, (types.IntegerLiteral, types.Literal)):
```
NaderAlAwar: Question: why do we need to check for `types.Literal` as well?

python/cuda_cccl/cuda/compute/struct.py
```diff
@@ -183,154 +386,127 @@ def tuple_to_struct_cast(context, builder, fromty, toty, val):
...
+            values = args[0]
+        elif args:
+            assert len(args) == len(field_spec), (
+                f"Expected {len(field_spec)} arguments, got {len(args)}"
+            )
+            values = dict(zip(field_spec.keys(), args))
+        else:
+            values = kwargs
+
+        assert values.keys() == field_spec.keys()
```
NaderAlAwar: Suggestion: the branching logic here is hard to follow. I'm not sure if it can be improved, but it would definitely help to move this to a new method, maybe `get_fields_as_dict` or something more appropriate.

python/cuda_cccl/tests/compute/examples/struct/nested_struct_tuple_construction.py
```diff
@@ -0,0 +1,80 @@
...
+    Reduction operation using tuple syntax for nested struct construction.
+
+    Instead of writing: Stats(d1.stats.count + d2.stats.count, ...)
+    We can use tuple syntax: (d1.stats.count + d2.stats.count, ...)
+    """
+    return DataPoint(
+        d1.value + d2.value,
+        # Tuple syntax for constructing the nested Stats struct
+        (d1.stats.count + d2.stats.count, d1.stats.sum + d2.stats.sum),
+    )
```
NaderAlAwar, line 45: Question: is it possible to also get rid of the `DataPoint` constructor call? It might be helpful to keep it for readability, but I am just curious.

python/cuda_cccl/tests/compute/test_zip_iterator.py
```diff
@@ -284,3 +284,167 @@ def add_pairs(p1, p2):
...
+
+
+@pytest.mark.parametrize("num_items", [10, 1000])
+@pytest.mark.parametrize(
+    "dtype_map",
+    [
+        {"x": np.float32, "y": np.float32},
+        pytest.param(
+            {"x": np.float64, "y": np.float32},
+            marks=pytest.mark.xfail(reason="Fails due to ODR violation"),
```
NaderAlAwar: Suggestion: would be helpful to link to an issue if we have one

python/cuda_cccl/cuda/compute/struct.py
shwina: I've removed the check for Literal - we don't need it.

python/cuda_cccl/cuda/compute/struct.py
shwina: I extracted this all to a helper method.

python/cuda_cccl/tests/compute/examples/struct/nested_struct_tuple_construction.py
shwina: Yes, it's possible (I tested locally)!

python/cuda_cccl/tests/compute/test_zip_iterator.py
shwina: Done

PR #6357, opened 2025-10-28
Fix `iter_move` constraints for MSVC

miscco:
Seems like we have issues in cases where the iterator prooduces a PR value that does not bind to an lvalue reference.

Investigate workarounds for this

### DIFF
--- a/libcudacxx/include/cuda/std/__iterator/iter_swap.h
+++ b/libcudacxx/include/cuda/std/__iterator/iter_swap.h
@@ -26,6 +26,7 @@
 #include <cuda/std/__iterator/iter_move.h>
 #include <cuda/std/__iterator/iterator_traits.h>
 #include <cuda/std/__iterator/readable_traits.h>
+#include <cuda/std/__type_traits/add_lvalue_reference.h>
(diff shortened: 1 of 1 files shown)


bernhardmgruber:
@miscco the tests passed with the override of the CI matrix, so I re-enabled the full CI.

bernhardmgruber:
I see some test failures around a change of `noexcept`:
```
 /home/coder/cccl/libcudacxx/test/libcudacxx/cuda/iterators/permutation_iterator/iter_swap.pass.cpp(41): error: static assertion failed
     static_assert(!noexcept(cuda::std::ranges::iter_swap(iter1, iter2)));
     ^
```

miscco:
> I see some test failures around a change of `noexcept`:
> 
> ```
>  /home/coder/cccl/libcudacxx/test/libcudacxx/cuda/iterators/permutation_iterator/iter_swap.pass.cpp(41): error: static assertion failed
>      static_assert(!noexcept(cuda::std::ranges::iter_swap(iter1, iter2)));
>      ^
> ```

Unfortuntely GCC prior to GCC14 is unable to use `noexcept` within a requires clause, which is ...

PR #6358, opened 2025-10-28
Implement std::seed_seq

RAMitchell:
## Description
Implements part of #5679

Implements std::seed_seq

## Checklist

### DIFF
--- a/libcudacxx/include/cuda/std/__random/seed_seq.h
+++ b/libcudacxx/include/cuda/std/__random/seed_seq.h
@@ -152,7 +152,8 @@ public:
       result_type __k_p_mod_n = (__k + __p) % __n;
       result_type __k_q_mod_n = (__k + __q) % __n;
       // 2.1
-      const result_type __r1 = 1664525 * __T(__begin[__k_mod_n] ^ __begin[__k_p_mod_n] ^ __begin[(__k - 1) % __n]);
+      const result_type __r1 =
+        1664525 * __generate_T(__begin[__k_mod_n] ^ __begin[__k_p_mod_n] ^ __begin[(__k - 1) % __n]);
       // 2.2
       result_type __r2 = (__k == 0)   ? __r1 + __z
                        : (__k <= __z) ? __r1 + __k_mod_n + __data_[__k - 1]
@@ -172,7 +173,8 @@ public:
       result_type __k_p_mod_n = (__k + __p) % __n;
       result_type __k_q_mod_n = (__k + __q) % __n;
       // 3.1
-      const result_type __r3 = 1566083941 * __T(__begin[__k_mod_n] + __begin[__k_p_mod_n] + __begin[(__k - 1) % __n]);
+      const result_type __r3 =
+        1566083941 * __generate_T(__begin[__k_mod_n] + __begin[__k_p_mod_n] + __begin[(__k - 1) % __n]);
       // 3.2
       const result_type __r4 = __r3 - __k_mod_n;
       // 3.3
@@ -187,7 +189,7 @@ public:
   }
 
 private:
-  [[nodiscard]] _CCCL_API static constexpr result_type __T(result_type __x) noexcept
+  [[nodiscard]] _CCCL_API static constexpr result_type __generate_T(result_type __x) noexcept
   {
     return (__x ^ (__x >> 27));
   }


libcudacxx/include/cuda/std/__random/seed_seq.h
```diff
@@ -0,0 +1,196 @@
...
+/// or is not mutable, the behavior is undefined.
+class _CCCL_TYPE_VISIBILITY_DEFAULT seed_seq
+{
+public:
+  using result_type = ::cuda::std::uint_least32_t;
+
+  /// @brief Default-construct an empty seed sequence.
+  ///
+  /// `size()` will return 0.
+  seed_seq() = default;
```
davebayer:
```suggestion
  _CCCL_HIDE_FROM_ABI constexpr seed_seq() noexcept = default;
```

libcudacxx/include/cuda/std/__random/seed_seq.h
```diff
@@ -0,0 +1,196 @@
...
+
+  /// @brief Construct from an iterator range of seed values.
+  /// @tparam _InputIt Input iterator type yielding values convertible to `result_type`.
+  /// @param __begin Iterator pointing to the first seed value.
+  /// @param __end Iterator one-past-the-last seed value.
+  template <class _InputIt>
+  _CCCL_API constexpr seed_seq(_InputIt __begin, _InputIt __end)
+  {
+    auto __n = ::cuda::std::distance(__begin, __end);
```
davebayer, line 66: You cannot call `distance` on input iterators. The problem is that you can iterate over input iterator only once. You can keep it like this for forward iterator, but for input iterator we will need to walk the iterator and enlarge the array step by step :(

libcudacxx/include/cuda/std/__random/seed_seq.h
```diff
@@ -0,0 +1,196 @@
...
+      __begin[__k_mod_n] = __r4 & 0xFFFFFFFF;
+    }
+  }
+
+private:
+  [[nodiscard]] _CCCL_API static constexpr result_type __T(result_type __x) noexcept
+  {
+    return (__x ^ (__x >> 27));
+  }
+  ::cuda::std::span<result_type> __v_{};
```
davebayer: This is not a good idea. `span` is a view, it shouldn't own the resources. Please, add `__data_` and `__size_` members instead.

libcudacxx/include/cuda/std/__random/seed_seq.h
```diff
@@ -0,0 +1,196 @@
...
+      __ptr[__i] = static_cast<result_type>(*__begin) & 0xFFFFFFFF;
+    }
+    __v_ = ::cuda::std::span<result_type>(__ptr, __n);
+  }
+
+  /// @brief Construct from an initializer list of seed values.
+  /// @tparam _InitT Element type convertible to `result_type`.
+  /// @param __il The list of seed values.
+  template <typename _InitT,
+            typename = typename ::cuda::std::enable_if<::cuda::std::is_convertible<_InitT, result_type>::value>::type>
```
davebayer: The standard says the `_InitT` must be an integer type, not just convertible to `result_type`

libcudacxx/include/cuda/std/__random/seed_seq.h
```diff
@@ -0,0 +1,196 @@
...
+  _CCCL_API constexpr seed_seq(std::initializer_list<_InitT> __il)
+      : seed_seq(__il.begin(), __il.end())
```
davebayer, line 93: This is not correct, we must store the values in the array

libcudacxx/include/cuda/std/__random/seed_seq.h
```diff
@@ -0,0 +1,196 @@
...
+  // Constexpr destructor from C++20 onwards
+#if _CCCL_STD_VER > 2017
+  constexpr
+#endif // _CCCL_STD_VER > 2017
+    _CCCL_API ~seed_seq()
+  {
+    if (!__v_.empty())
+    {
+      delete[] __v_.data();
+    }
```
davebayer:
```suggestion
    delete[] __v_.data();
```
Delete can be called with `nullptr`

libcudacxx/include/cuda/std/__random/seed_seq.h
```diff
@@ -0,0 +1,196 @@
...
+  ///
+  /// `size()` will return 0.
+  seed_seq() = default;
+
+  /// @brief Construct from an iterator range of seed values.
+  /// @tparam _InputIt Input iterator type yielding values convertible to `result_type`.
+  /// @param __begin Iterator pointing to the first seed value.
+  /// @param __end Iterator one-past-the-last seed value.
+  template <class _InputIt>
+  _CCCL_API constexpr seed_seq(_InputIt __begin, _InputIt __end)
```
davebayer:
```suggestion
  _CCCL_API _CCCL_CONSTEXPR_CXX20 seed_seq(_InputIt __begin, _InputIt __end)
```

libcudacxx/include/cuda/std/__random/seed_seq.h
```diff
@@ -0,0 +1,196 @@
...
+  {}
+
+  /// @brief seed_seq is not copyable.
+  seed_seq(const seed_seq&) = delete;
+
+  // Constexpr destructor from C++20 onwards
+#if _CCCL_STD_VER > 2017
+  constexpr
+#endif // _CCCL_STD_VER > 2017
+    _CCCL_API ~seed_seq()
```
davebayer:
```suggestion
    _CCCL_API _CCCL_CONSTEXPR_CXX20 ~seed_seq()
```

libcudacxx/include/cuda/std/__random/seed_seq.h
```diff
@@ -0,0 +1,196 @@
...
+  {
+    if (!__v_.empty())
+    {
+      delete[] __v_.data();
+    }
+  }
+
+  /// @brief seed_seq is not copy-assignable.
+  _CCCL_API auto operator=(const seed_seq&) = delete;
```
davebayer:
```suggestion
  seed_seq& operator=(const seed_seq&) = delete;
```

libcudacxx/include/cuda/std/__random/seed_seq.h
RAMitchell: Ok is __end - __begin  ok in that case?

libcudacxx/include/cuda/std/__random/seed_seq.h
RAMitchell: Not sure I understand sorry.

libcudacxx/include/cuda/std/__random/seed_seq.h
davebayer: Aaah, I am sorry, I am wrong, you can use `distance` event for input iterators, can stay as is :)

libcudacxx/include/cuda/std/__random/seed_seq.h
davebayer: Again, sorry, it's fine as is - should have had the lunch first before reviewing stuff

libcudacxx/include/cuda/std/__random/seed_seq.h
```diff
@@ -114,7 +107,7 @@ class _CCCL_TYPE_VISIBILITY_DEFAULT seed_seq
   template <class _OutputIt>
   _CCCL_API constexpr void param(_OutputIt __dest) const
   {
-    ::cuda::std::copy(__v_.begin(), __v_.end(), __dest);
+    ::cuda::std::copy(__data_, __data_ + __size_, __dest);
```
davebayer: You can use `copy_n` here

libcudacxx/include/cuda/std/__random/seed_seq.h
```diff
@@ -0,0 +1,192 @@
...
+  /// @param __end Iterator one-past-the-last seed value.
+  template <class _InputIt>
+  _CCCL_API _CCCL_CONSTEXPR_CXX20 seed_seq(_InputIt __begin, _InputIt __end)
+  {
+    auto __n = ::cuda::std::distance(__begin, __end);
+    if (__n <= 0)
+    {
+      return;
+    }
+    __data_ = new result_type[__n];
```
davebayer:
```suggestion
    __data_ = new result_type[__n]{};
```
You need to initialize the array, so it can work in `constexpr` context

libcudacxx/include/cuda/std/__random/seed_seq.h
```diff
@@ -0,0 +1,192 @@
...
+    __size_ = __n;
```
davebayer:
```suggestion
    __size_ = static_cast<size_t>(__n);
```
Let's explicitly convert signed to unsigned

libcudacxx/include/cuda/std/__random/seed_seq.h
```diff
@@ -0,0 +1,192 @@
...
+    for (::cuda::std::size_t __i = 0; __i < static_cast<::cuda::std::size_t>(__n); ++__i, ++__begin)
```
davebayer:
```suggestion
    for (size_t __i = 0; __i < __size_; ++__i, ++__begin)
```
We are in `cuda::std::` namespace, so the `size_t` from `cuda::std::` will be used

libcudacxx/include/cuda/std/__random/seed_seq.h
```diff
@@ -0,0 +1,192 @@
...
+    {
+      __data_[__i] = static_cast<result_type>(*__begin) & 0xFFFFFFFF;
```
davebayer: Can't we just do

```suggestion
      __data_[__i] = static_cast<uint32_t>(*__begin);
```

libcudacxx/include/cuda/std/__random/seed_seq.h
davebayer: Wasn't applied yet

libcudacxx/include/cuda/std/__random/seed_seq.h
```diff
@@ -0,0 +1,192 @@
...
+    }
+  }
+
+  /// @brief Construct from an initializer list of seed values.
+  /// @tparam _InitT Element type convertible to `result_type`.
+  /// @param __il The list of seed values.
+  template <typename _InitT,
+            typename = typename ::cuda::std::enable_if_t<
+              ::cuda::std::is_convertible_v<_InitT, result_type>&& ::cuda::std::is_integral_v<_InitT>>>
```
davebayer: Should be:

```suggestion
  _CCCL_TEMPLATE(typename _InitT)
  _CCCL_REQUIRES(__cccl_is_integer_v<_InitT>)
```
The standard says integer, not integral. Integral types are integers + character types + bool

Also, please add includes: `<cuda/std/__type_traits/is_integer.h>` and `<cuda/std/__concepts/concept_macros.h>`

libcudacxx/include/cuda/std/__random/seed_seq.h
```diff
@@ -0,0 +1,192 @@
...
+  _CCCL_API _CCCL_CONSTEXPR_CXX20 seed_seq(std::initializer_list<_InitT> __il)
```
davebayer: Should be just:
```suggestion
  _CCCL_API _CCCL_CONSTEXPR_CXX20 seed_seq(initializer_list<_InitT> __il)
```
We import `std::initializer_list` to `cuda::std::` namespace

libcudacxx/include/cuda/std/__random/seed_seq.h
```diff
@@ -0,0 +1,192 @@
...
+  {
+    delete[] __data_;
+  }
+
+  /// @brief seed_seq is not copy-assignable.
+  seed_seq& operator=(const seed_seq&) = delete;
+
+  /// @brief Returns the number of seed values stored.
+  /// @return Number of internal seed values (may be 0).
+  [[nodiscard]] _CCCL_API constexpr ::cuda::std::size_t size() const noexcept
```
davebayer:
```suggestion
  [[nodiscard]] _CCCL_API constexpr size_t size() const noexcept
```
applies to other places, too

libcudacxx/include/cuda/std/__random/seed_seq.h
RAMitchell: That works :)

libcudacxx/include/cuda/std/__random/seed_seq.h
```diff
@@ -0,0 +1,193 @@
...
+/// the values are produced by the standard seed sequence mixing algorithm.
+///
+/// If `std::iterator_traits<RandomIt>::value_type` is not an unsigned integer
+/// type, or its width is less than 32 bits, the program is ill-formed.
+/// If `RandomIt` does not meet the requirements of a LegacyRandomAccessIterator
+/// or is not mutable, the behavior is undefined.
+class _CCCL_TYPE_VISIBILITY_DEFAULT seed_seq
+{
+public:
+  using result_type = ::cuda::std::uint_least32_t;
```
davebayer:
```suggestion
  using result_type = uint_least32_t;
```

libcudacxx/include/cuda/std/__random/seed_seq.h
```diff
@@ -0,0 +1,193 @@
...
+  _CCCL_API _CCCL_CONSTEXPR_CXX20 seed_seq(_InputIt __begin, _InputIt __end)
+  {
+    auto __n = ::cuda::std::distance(__begin, __end);
+    if (__n <= 0)
+    {
+      return;
+    }
+    __data_ = new result_type[__n]{};
+    __size_ = static_cast<size_t>(__n);
+    for (size_t __i = 0; __i < static_cast<size_t>(__n); ++__i, ++__begin)
```
davebayer: Why don't you just use `__size_` here?

```suggestion
    for (size_t __i = 0; __i < __size_; ++__i, ++__begin)
```

libcudacxx/test/libcudacxx/std/random/seed_seq/equal.fail.cpp
```diff
@@ -0,0 +1,25 @@
...
+//===----------------------------------------------------------------------===//
+
+#include <cuda/std/__random_>
+#include <cuda/std/cassert>
+
+__host__ __device__ void test()
+{
+  ::cuda::std::seed_seq seq1{1, 2, 3};
+  ::cuda::std::seed_seq seq2;
+  bool _ = (seq2 == seq1);
```
davebayer:
```suggestion
  ::cuda::std::seed_seq seq1{1, 2, 3};
  ::cuda::std::seed_seq seq2;
  [[maybe_unused]] bool _ = (seq2 == seq1);
```

libcudacxx/test/libcudacxx/std/random/seed_seq/ctor.pass.cpp
```diff
@@ -0,0 +1,49 @@
...
+  ::cuda::std::array<::cuda::std::uint32_t, 5> reference = {4, 5, 6, 7, 8};
+  assert(init_seeds_copy == reference);
+  return true;
+}
+
+int main(int, char**)
+{
+  test();
+#if TEST_STD_VER >= 2020
+  static_assert(test(), "");
```
davebayer:
```suggestion
  static_assert(test());
```
We are C++17, so we needn't to have static assert message :)

libcudacxx/test/libcudacxx/std/random/seed_seq/ctor.pass.cpp
```diff
@@ -0,0 +1,49 @@
...
+  assert(empty_seeds_copy.empty());
+
+  // Initializer list constructor
+  ::cuda::std::seed_seq seq3{4, 5, 6, 7, 8};
+  assert(seq3.size() == 5);
+  ::cuda::std::array<::cuda::std::uint32_t, 5> init_seeds_copy{};
+  seq3.param(init_seeds_copy.begin());
+  ::cuda::std::array<::cuda::std::uint32_t, 5> reference = {4, 5, 6, 7, 8};
+  assert(init_seeds_copy == reference);
+  return true;
```
davebayer:
```suggestion
  cuda::std::array<cuda::std::uint32_t, 3> seeds{1, 2, 3};
  
  // 1. Default constructor
  {
    cuda::std::seed_seq seq;
    assert(seq.size() == 0);
  }
  
  // 2. Iterator constructor
  {
    cuda::std::seed_seq seq(seeds.begin(), seeds.end());
    assert(seq.size() == 3);
    
    cuda::std::array<cuda::std::uint32_t, 3> seeds_copy{};
    seq.param(seeds_copy.begin());
    assert(seeds_copy == seeds);
  }

  // 3. Initializer list constructor
  {
    cuda::std::seed_seq seq{4, 5, 6, 7, 8};
    assert(seq.size() == 5);
    
    cuda::std::array<cuda::std::uint32_t, 5> init_seeds_copy{};
    seq.param(init_seeds_copy.begin());
    assert((init_seeds_copy == cuda::std::array<cuda::std::uint32_t, 5>{4, 5, 6, 7, 8}));
  }
  
  return true;
```
1. Could we improve the structure of the test?
2. Don't fully qualify types nor calls in tests

libcudacxx/test/libcudacxx/std/random/seed_seq/generate.pass.cpp
```diff
@@ -0,0 +1,61 @@
...
+  static_assert(noexcept(::cuda::std::seed_seq{}));
+  cuda::std::seed_seq seq{1, 2, 3, 4, 5};
+  cuda::std::uint32_t out[5] = {};
+  seq.generate(::cuda::std::begin(out), ::cuda::std::end(out));
+  static_assert(cuda::std::is_void_v<decltype(seq.generate(::cuda::std::begin(out), ::cuda::std::end(out)))>);
+  return true;
+}
+
+#if !_CCCL_COMPILER(NVRTC)
+#  include <random>
```
davebayer: Move the include up, please

RAMitchell:
Looking at llvms implementation I notice two differences:
1) They carefully optimise the mod operations. I don't think this class is designed to be used in any performance context (i.e. inside a loop), so I haven't done this but can if needed.
2) They don't mask like `&= 0xFFFFFFFF`. This is specified in the standard and its not clear that their implementation is correct without it. Without this the generate function could modify the higher 32 bits in a 64 bit input, which should be zero?

RAMitchell:
Checking another implementation, it does mask the outputs and looks closer to my version.

davebayer:
> 1. They carefully optimise the mod operations. I don't think this class is designed to be used in any performance context (i.e. inside a loop), so I haven't done this but can if needed.

I would leave that as is for now. We can always implement it later when someone asks for it.

> 2. They don't mask like `&= 0xFFFFFFFF`. This is specified in the standard and its not clear that their implementation is correct without it. Without this the generate function could modify the higher 32 bits in a 64 bit input, which should be zero?

Maybe it's because they know that the `uint_least32_t` is always `uint32_t`, like it's in our case.

> Checking another implementation, it does mask the outputs and looks closer to my version.

I think that the current version is good :)

libcudacxx/include/cuda/std/__random/seed_seq.h
```diff
@@ -0,0 +1,193 @@
...
+    if (__n <= 0)
+    {
+      return;
+    }
+    __data_ = new result_type[__n]{};
+    __size_ = static_cast<size_t>(__n);
+    for (size_t __i = 0; __i < __size_; ++__i, ++__begin)
+    {
+      __data_[__i] = static_cast<uint32_t>(*__begin);
+    }
```
davebayer, line 84: I am thinking about whether there could be a memory leak if `++__begin` or `*__begin` throw and exception. Maybe we should rather do:

```suggestion
    __data_ = new result_type[__n]{};
    __size_ = static_cast<size_t>(__n);
    _CCCL_TRY
    {
      for (size_t __i = 0; __i < __size_; ++__i, ++__begin)
      {
        __data_[__i] = static_cast<uint32_t>(*__begin);
      }
    }
    _CCCL_CATCH_ALL
    {
      delete[] __data_;
      _CCCL_RETHROW;
    }
```

libcudacxx/include/cuda/std/__random/seed_seq.h
```diff
@@ -0,0 +1,193 @@
...
+    return __size_;
+  }
+
+  /// @brief Copy stored seed values into the output iterator `__dest`.
+  /// @tparam _OutputIt Output iterator type accepting `result_type` values.
+  /// @param __dest Destination iterator where stored values will be written.
+  template <class _OutputIt>
+  _CCCL_API constexpr void param(_OutputIt __dest) const
+  {
+    copy_n(__data_, __size_, __dest);
```
davebayer:
```suggestion
    ::cuda::std::copy_n(__data_, __size_, __dest);
```
I should have left a note before.. We needn't to fully qualify types in the same namespace, but we need to fully qualify the function calls

libcudacxx/test/libcudacxx/std/random/engine/test_engine.h
```diff
@@ -156,6 +147,7 @@ __host__ __device__ constexpr bool test_min_max()
 }
 
 #if !_CCCL_COMPILER(NVRTC)
+#  include <sstream>
```
davebayer: Please, move the include to the top of the file

libcudacxx/test/libcudacxx/std/random/seed_seq/assign.compile.pass.cpp
```diff
@@ -0,0 +1,25 @@
...
+
+#include <cuda/std/__random_>
+#include <cuda/std/cassert>
+
+__host__ __device__ void test()
+{
+  ::cuda::std::seed_seq seq1{1, 2, 3};
+  ::cuda::std::seed_seq seq2;
+  seq2 = seq1;
+}
```
davebayer, line 20: I think we should make this test be just:
```cpp
static_assert(!cuda::std::is_copy_assignable_v<cuda::std::seed_seq>);
static_assert(!cuda::std::is_move_assignable_v<cuda::std::seed_seq>);
```
And this test be can be a `.pass` test

libcudacxx/test/libcudacxx/std/random/seed_seq/copy.fail.cpp
```diff
@@ -0,0 +1,24 @@
...
+//===----------------------------------------------------------------------===//
+
+#include <cuda/std/__random_>
+#include <cuda/std/cassert>
+
+__host__ __device__ void test()
+{
+  ::cuda::std::seed_seq seq1{1, 2, 3};
+  ::cuda::std::seed_seq seq2 = seq1;
+}
```
davebayer, line 20: Let's just get rid of the test completely and just check for:
```cpp
static_assert(!cuda::std::is_copy_constructible_v<cuda::std::seed_seq>);
static_assert(!cuda::std::is_move_constructible_v<cuda::std::seed_seq>);
```
inside `ctor.pass.cpp`

libcudacxx/test/libcudacxx/std/random/seed_seq/param.pass.cpp
```diff
@@ -0,0 +1,34 @@
...
+__host__ __device__ TEST_CONSTEXPR_CXX20 bool test()
+{
+  ::cuda::std::array<::cuda::std::uint32_t, 3> seeds_copy{};
+  ::cuda::std::array<::cuda::std::uint32_t, 3> seeds = {1, 2, 3};
+  ::cuda::std::seed_seq seq1(seeds.begin(), seeds.end());
+  seq1.param(seeds_copy.begin());
+  assert(seeds_copy == seeds);
+  static_assert(cuda::std::is_void_v<decltype(seq1.param(seeds_copy.begin()))>);
+  return true;
+}
```
davebayer, line 25: Can we add a test for _InputIterator_ as well? You can use `std::istringstream` for example. Because right now we test only _ContiguousIterator_.

RAMitchell: Is this for the ctor or the param method?

RAMitchell: I am not able to use std::istream_iterator because ` calling a __host__ function from a __host__ __device__ function is not allowed`. Do we have an example of ContiguousIterator internally?

davebayer: Yeah I meant the constructor, it should be implemented in `ctor.pass.cpp`

To solve the error, just put it in a separate function that will be `__host__` only and it will work fine.

libcudacxx/include/cuda/std/__random/seed_seq.h
RAMitchell: This _CCCL_RETHROW macro doesn't seem to exist.

libcudacxx/test/libcudacxx/std/random/seed_seq/param.pass.cpp
RAMitchell: The error comes from calling ::cuda::std::distance on `std::istream_iterator` inside seed_seq as opposed to in the tests.

RAMitchell: I'll make my own from https://en.cppreference.com/w/cpp/iterator/input_iterator.html

libcudacxx/include/cuda/std/__random/seed_seq.h
```diff
@@ -0,0 +1,201 @@
...
+  _CCCL_HIDE_FROM_ABI constexpr seed_seq() noexcept = default;
+
+  /// @brief Construct from an iterator range of seed values.
+  /// @tparam _InputIt Input iterator type yielding values convertible to `result_type`.
+  /// @param __begin Iterator pointing to the first seed value.
+  /// @param __end Iterator one-past-the-last seed value.
+  template <class _InputIt>
+  _CCCL_API _CCCL_CONSTEXPR_CXX20 seed_seq(_InputIt __begin, _InputIt __end)
+  {
+    auto __n = ::cuda::std::distance(__begin, __end);
```
miscco, line 66: Yeah distance takes by reference, that said it should have been guarded by `sized_sentinel_for<_InputIt _InputIt >`

This is fine here I believe

libcudacxx/include/cuda/std/__random/seed_seq.h
```diff
@@ -0,0 +1,201 @@
...
+    if (__n <= 0)
+    {
+      return;
+    }
+    __data_ = new result_type[__n]{};
```
miscco, line 71: We do have `cuda::std::unique_ptr` albeit that is indeed something slightl more costly

libcudacxx/include/cuda/std/__random/seed_seq.h
```diff
@@ -0,0 +1,201 @@
...
+      // 3.4
+      __begin[__k_q_mod_n] ^= __r4;
+      __begin[__k_q_mod_n] &= 0xFFFFFFFF;
+      // 3.5
+      __begin[__k_mod_n] = __r4 & 0xFFFFFFFF;
+    }
+  }
+
+private:
+  [[nodiscard]] _CCCL_API static constexpr result_type __T(result_type __x) noexcept
```
miscco: Please give the operation a better name

libcudacxx/test/libcudacxx/std/random/seed_seq/assign.compile.pass.cpp
```diff
@@ -0,0 +1,26 @@
...
+// SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES.
+//
+//===----------------------------------------------------------------------===//
+
+#include <cuda/std/__random_>
+#include <cuda/std/cassert>
+
+__host__ __device__ void test()
+{
+  static_assert(!cuda::std::is_copy_assignable_v<cuda::std::seed_seq>);
```
miscco, line 16: Nitpick: this test does not need to be actually run, so we can change the file ending to `assign.compile.pass.cpp`

libcudacxx/include/cuda/std/__random/seed_seq.h
davebayer: Since this type is not copyable/movable nor copy/move assignable, I would rather go for `new`/`delete`

libcudacxx/include/cuda/std/__random/seed_seq.h
davebayer: I also wanted to suggest that, but if you have a look at the `generate()` [docs](https://en.cppreference.com/w/cpp/numeric/random/seed_seq/generate.html), there is the `T(...)` operation described. Maybe we can call it `__generate_T`

RAMitchell: It is defined as precisely T in the standard. I think its more confusing to rename it.

libcudacxx/include/cuda/std/__random/seed_seq.h
```diff
@@ -0,0 +1,202 @@
...
+  /// @param __end Iterator one-past-the-end of the destination range.
+  template <class _RandomIt>
+  _CCCL_API constexpr void generate(_RandomIt __begin, _RandomIt __end) const
+  {
+    if (__begin == __end)
+    {
+      return;
+    }
+    // https://en.cppreference.com/w/cpp/numeric/random/seed_seq/generate.html
+    const result_type __z = __size_;
```
davebayer:
```suggestion
    const result_type __z = static_cast<result_type>(__size_);
```
to fix msvc

libcudacxx/test/libcudacxx/std/random/seed_seq/generate.pass.cpp
```diff
@@ -0,0 +1,64 @@
...
+  std::vector<std::vector<::std::uint32_t>> sequences = {
+    {1, 2, 3, 4, 5},
+    {42, 43, 44, 45, 46, 47, 48, 49, 50},
+    {123456789, 987654321, 555555555, 333333333, 111111111},
+    {::std::numeric_limits<::std::uint32_t>::max()},
+    {0}};
+  for (const auto& seq_values : sequences)
+  {
+    cuda::std::seed_seq cuda_seq{seq_values.data(), seq_values.data() + seq_values.size()};
+    std::seed_seq std_seq{std::begin(seq_values), std::end(seq_values)};
```
davebayer:
```suggestion
    std::seed_seq std_seq(seq_values.begin(), seq_values.end());
```

libcudacxx/test/libcudacxx/std/random/seed_seq/generate.pass.cpp
```diff
@@ -0,0 +1,64 @@
...
+#if !_CCCL_COMPILER(NVRTC)
+void test_against_std()
+{
+  ::std::size_t n                                     = 100;
+  std::vector<std::vector<::std::uint32_t>> sequences = {
+    {1, 2, 3, 4, 5},
+    {42, 43, 44, 45, 46, 47, 48, 49, 50},
+    {123456789, 987654321, 555555555, 333333333, 111111111},
+    {::std::numeric_limits<::std::uint32_t>::max()},
+    {0}};
```
davebayer:
```suggestion
  cuda::std::size_t n                                     = 100;
  std::vector<std::vector<cuda::std::uint32_t>> sequences = {
    {1, 2, 3, 4, 5},
    {42, 43, 44, 45, 46, 47, 48, 49, 50},
    {123456789, 987654321, 555555555, 333333333, 111111111},
    {cuda::std::numeric_limits<cuda::std::uint32_t>::max()},
    {0}};
```

libcudacxx/test/libcudacxx/std/random/seed_seq/generate.pass.cpp
```diff
@@ -0,0 +1,64 @@
...
+  for (const auto& seq_values : sequences)
+  {
+    cuda::std::seed_seq cuda_seq{seq_values.data(), seq_values.data() + seq_values.size()};
+    std::seed_seq std_seq{std::begin(seq_values), std::end(seq_values)};
+
+    std::vector<::std::uint32_t> cuda_output(n);
+    std::vector<std::uint32_t> std_output(n);
```
davebayer:
```suggestion
    std::vector<cuda::std::uint32_t> cuda_output(n);
    std::vector<cuda::std::uint32_t> std_output(n);
```

libcudacxx/include/cuda/std/__random/seed_seq.h
miscco: We cannot use `__T` though because AFAIK that is a predefined macro on MSVC wrt to different string types

Issue #6302, opened 2025-10-21
[BUG]: cuda::memcopy_async{_tx} has inefficient codegen on SM120


ahendriksen:
### Is this a duplicate?

- [x] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cccl/issues) for this bug and that I agree to the [Code of Conduct](CODE_OF_CONDUCT.md)

### Type of Bug

Performance

### Component

libcu++

### Describe the bug

`cuda::memcpy_async{_tx}` emits the `cp.async.bulk.shared::cluster.global.mbarrier::complete_tx::bytes` instruction. This instruction has inefficient code-gen on SM120. It provides a fallback path when the shared memory pointer points to remote shared memory.

When available, the code should emit the PTX instruction with `.shared::cta` modifier, which has better codegen on SM120:

`cp.async.bulk.shared::cta.global.mbarrier::complete_tx::bytes`

### How to Reproduce

https://godbolt.org/z/P641vnYM3

### Expected behavior

Using `.shared::cta` as `.space` has better codegen on sm120: 

https://godbolt.org/z/1xx5eP1Ms

### Reproduction link

https://godbolt.org/z/P641vnYM3

### NVCC version

13.0

ahendriksen:
The `.shared::cta` option was added in PTX ISA 8.6: 

https://nvidia.github.io/cccl/libcudacxx/ptx/instructions/cp_async_bulk.html#cp-async-bulk-shared-cta-global-mbarrier-complete-tx-bytes

bernhardmgruber:
I am certain we have this in other places too! I think @pauleonix also noticed this while working on `cub::[detail::]BlockLoadToShared`

PR #6362, opened 2025-10-28
Use `cp_async_bulk(space_shared,...)` when available

bernhardmgruber:
Fixes: #6302

### DIFF
--- a/cub/cub/device/dispatch/kernels/transform.cuh
+++ b/cub/cub/device/dispatch/kernels/transform.cuh
@@ -653,11 +653,7 @@ _CCCL_DEVICE void bulk_copy_maybe_unaligned(
       _CCCL_ASSERT(aligned_bytes_to_copy % bulk_copy_size_multiple == 0, "");
 
       ::cuda::ptx::cp_async_bulk(
-#if __cccl_ptx_isa >= 860
-        ::cuda::ptx::space_shared,
-#else // __cccl_ptx_isa >= 860
-        ::cuda::ptx::space_cluster,
-#endif // __cccl_ptx_isa >= 860
+        ::cuda::std::conditional_t<__cccl_ptx_isa >= 860, ::cuda::ptx::space_shared_t, ::cuda::ptx::space_cluster_t>{},
         ::cuda::ptx::space_global,
         dst_ptr + head_bytes,
         src_ptr + head_bytes,
@@ -792,11 +788,7 @@ _CCCL_DEVICE void transform_kernel_ublkcp(
         }
 
         ::cuda::ptx::cp_async_bulk(
-#if __cccl_ptx_isa >= 860
-          ::cuda::ptx::space_shared,
-#else // __cccl_ptx_isa >= 860
-          ::cuda::ptx::space_cluster,
-#endif // __cccl_ptx_isa >= 860
+          ::cuda::std::conditional_t<__cccl_ptx_isa >= 860, ::cuda::ptx::space_shared_t, ::cuda::ptx::space_cluster_t>{},
           ::cuda::ptx::space_global,
           dst,
           src,


cub/cub/device/dispatch/kernels/transform.cuh
```diff
@@ -653,7 +653,11 @@ _CCCL_DEVICE void bulk_copy_maybe_unaligned(
       _CCCL_ASSERT(aligned_bytes_to_copy % bulk_copy_size_multiple == 0, "");
 
       ::cuda::ptx::cp_async_bulk(
+#if __cccl_ptx_isa >= 860
+        ::cuda::ptx::space_shared,
```
fbusato: the original issue only describes slow performance for SM120. What are the implications of this change for other archs?

libcudacxx/include/cuda/__memcpy_async/cp_async_bulk_shared_global.h
```diff
@@ -54,15 +54,20 @@ inline _CCCL_DEVICE void __cp_async_bulk_shared_global_and_expect_tx(
   const _Group& __g, char* __dest, const char* __src, ::cuda::std::size_t __size, ::cuda::std::uint64_t* __bar_handle)
 {
   // https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#data-movement-and-conversion-instructions-cp-async-bulk
+#    if __cccl_ptx_isa >= 860
```
miscco: I believe we should move that out to a define that we can use in all places

bernhardmgruber: Found a better way. Using `conditional_t` now.

cub/cub/device/dispatch/kernels/transform.cuh
bernhardmgruber: Trying that with https://godbolt.org/z/1xx5eP1Ms, the codegen is identical on sm90 and sm100 for `::cuda::ptx::space_shared` and ` ::cuda::ptx::space_cluster`.

I would still go with `::cuda::ptx::space_shared` regardless, since it is semantically what we do (the destination is CTA local SMEM)

libcudacxx/include/cuda/__memcpy_async/memcpy_async_tx.h
```diff
@@ -75,7 +76,7 @@ _CCCL_DEVICE inline async_contract_fulfillment memcpy_async_tx(
       if (::cuda::device::is_address_from(__dest, ::cuda::device::address_space::shared)
           && ::cuda::device::is_address_from(__src, ::cuda::device::address_space::global)) {
         ::cuda::ptx::cp_async_bulk(
-          ::cuda::ptx::space_cluster,
+          ::cuda::std::conditional_t<__cccl_ptx_isa >= 860, ::cuda::ptx::space_shared, ::cuda::ptx::space_cluster>,
```
fbusato: would not be better to use a global `constexpr` variable?

cub/cub/device/dispatch/kernels/transform.cuh
fbusato: perfect

libcudacxx/include/cuda/__memcpy_async/memcpy_async_tx.h
bernhardmgruber: I think the small duplication is fine for the two occurences here. Also, I would not know where to put the global variable except for a new header and I feel that's overkill for this small branch.

cub/cub/device/dispatch/kernels/transform.cuh
```diff
@@ -787,7 +791,17 @@ _CCCL_DEVICE void transform_kernel_ublkcp(
           bytes_to_copy = int{sizeof(T)} * tile_size;
         }
 
-        ::cuda::ptx::cp_async_bulk(::cuda::ptx::space_cluster, ::cuda::ptx::space_global, dst, src, bytes_to_copy, &bar);
+        ::cuda::ptx::cp_async_bulk(
+#if __cccl_ptx_isa >= 860
+          ::cuda::ptx::space_shared,
+#else // __cccl_ptx_isa >= 860
+          ::cuda::ptx::space_cluster,
+#endif // __cccl_ptx_isa >= 860
```
miscco: Nitpick: those two should also use the `conditional_t`

bernhardmgruber: I considered and thought that the preprocessor might compile faster. But I don't mind. Chaning to `conditional_t`.

PR #6364, opened 2025-10-28
Add missing header to random engine tests

RAMitchell:


libcudacxx/test/libcudacxx/std/random/engine/test_engine.h
```diff
@@ -156,6 +156,7 @@ __host__ __device__ constexpr bool test_min_max()
 }
 
 #if !_CCCL_COMPILER(NVRTC)
+#  include <sstream>
```
davebayer: Can we move the include to the top of the file, please?

davebayer (changes requested):

PR #6365, opened 2025-10-28
Implement Stream Ordered Mixed Host and Device Execution Example with `host_launch`

### LINKED ISSUE #5708: [FEA]: Stream Ordered Mixed Host and Device Execution Example with `host_launch`
### Is this a duplicate?

- [ ] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cccl/issues) for this request and that I agree to the [Code of Conduct](CODE_OF_CONDUCT.md)

### Area

General CCCL

Implement an example demonstrating stream-ordered execution mixing device and host launches. The host launch should prepare data for a later kernel launch and finally consume it with another host launch, showcasing the sequencing and data dependency.

Provide a sample illustrating `host_launch` usage to prepare, launch, and consume data, with clear comments and minimal code.

Alternative is to document the API, but an executable example better shows the intent and possibilities.

### Additional context

Should highlight interplay between host and device in stream ordering.

davebayer:
Closes #5708.

### DIFF
--- a/examples/ccclrt/heat_diffusion/heat_diffusion.cu
+++ b/examples/ccclrt/heat_diffusion/heat_diffusion.cu
@@ -1,0 +1,357 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of the libcu++ Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES.
+//
+//===----------------------------------------------------------------------===//
+
+//===----------------------------------------------------------------------===//
+//
+// MIT License
+//
+// Copyright (c) 2016 CSC Training
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+//
+//===----------------------------------------------------------------------===//
+
+#include <cuda/algorithm>
+#include <cuda/buffer>
+#include <cuda/cmath>
+#include <cuda/hierarchy>
+#include <cuda/launch>
+#include <cuda/mdspan>
+#include <cuda/memory_pool>
+#include <cuda/std/span>
+#include <cuda/std/utility>
+#include <cuda/stream>
+
+#include <cstdio>
+#include <exception>
+#include <fstream>
+#include <stdexcept>
+#include <string>
+
+//
+// Based on CSC materials from:
+//
+// https://github.com/csc-training/openacc/tree/master/exercises/heat
+//
+
+//! @brief Initializes the data with a pattern of a disk of radius of 1/6 of the width.
+void init_grid(cuda::host_mdspan<float, cuda::std::dims<2, int>> grid)
+{
+  const float radius2 = (grid.extent(0) / 6.f) * (grid.extent(1) / 6.f);
+
+  for (int i = 0; i < grid.extent(0); ++i)
+  {
+    for (int j = 0; j < grid.extent(1); ++j)
+    {
+      const float distance2 =
+        (i - grid.extent(0) / 2) * (i - grid.extent(0) / 2) + (j - grid.extent(1) / 2) * (j - grid.extent(1) / 2);
+      grid(i, j) = (distance2 < radius2) ? 65.f : 5.f;
+    }
+  }
+}
+
+//! @brief Kernel functor for heat diffusion computation.
+class HeatDiffusionKernelFunctor
+{
+  int nx_; //!< Number of grid points in X axis.
+  int ny_; //!< Number of grid points in Y axis.
+  float dx2_; //!< Horizontal grid spacing squared.
+  float dy2_; //!< Vertical grid spacing squared.
+  float a_times_dt_; //!< Diffusion coefficient multiplied by time difference.
+
+public:
+  //! @brief The number of threads per dimension in a thread block.
+  static constexpr int threads_per_side = 16;
+
+  //! @brief Constructor.
+  //!
+  //! @param nx Number of grid points in X axis.
+  //! @param ny Number of grid points in Y axis.
+  //! @param a The diffusion coefficient.
+  //! @param dx Horizontal grid spacing.
+  //! @param dy Vertical grid spacing.
+  //! @param dt Time difference.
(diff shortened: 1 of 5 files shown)


cudax/examples/cccl-rt/heat_diffusion.cu
```diff
@@ -0,0 +1,331 @@
...
+      , a_times_dt_{a * dt}
+  {}
+
+  //! @brief Gets the default configuration with filled block dimensions. Should be combined with grid dimensions
+  //!        config.
+  [[nodiscard]] constexpr auto default_config() const noexcept
+  {
+    const auto blocks_per_grid_x = cuda::ceil_div(nx_, threads_per_side);
+    const auto blocks_per_grid_y = cuda::ceil_div(ny_, threads_per_side);
+    return cudax::make_config(cudax::make_hierarchy(cudax::block_dims<threads_per_side, threads_per_side>(),
```
pciolkosz: You can directly pass dims to `make_config`, no need to call `make_hierarchy` directly

cudax/examples/cccl-rt/heat_diffusion.cu
```diff
@@ -0,0 +1,331 @@
...
+  for (int ti = 0; ti <= nt; ti++)
+  {
+    if (should_output(ti))
+    {
+      // Record an event to main stream signaling that the data is ready and the copy operation can start.
+      copy_output_event.record(main_stream);
+    }
+
+    // Enqueue the computation to the main stream.
+    cudax::launch(main_stream, {}, kernel_functor, device_buffer_1.data(), device_buffer_2.data());
```
pciolkosz: You can take buffers as spans and then you don't need to call `.data()` on buffer, they will transform to spans automatically

cudax/examples/cccl-rt/heat_diffusion.cu
```diff
@@ -0,0 +1,331 @@
...
+    if (should_output(ti))
+    {
+      // Make the output stream wait for the data to be ready.
+      output_stream.wait(copy_output_event);
+
+      // Enqueue the copy from device to host buffer.
+      cudax::copy_bytes(output_stream, device_buffer_1, host_buffer);
+
+      // Record an event signaling that the copy operation has finished.
+      copy_done_event.record(output_stream);
```
pciolkosz: We could also do `main_stream.wait(output_stream)` here, its technically allocating an event every time and discards it, but it would be nice to show both explicit and implicit event cases, I expect we will get `cudaStreamWaitStream` in some near future to optimize the implicit event case

cudax/examples/cccl-rt/heat_diffusion.cu
davebayer: Yeah, but I don't think there is any benefit to it, I don't need the size, I know what the dimensions are inside the functor

cudax/examples/cccl-rt/heat_diffusion.cu
davebayer: I don't think we can do that, because the whole compute loop is asynchronous and all of the work is launched to streams at once. The problem is that the event will be always destroyed before the synchronization shall happen.

If I am right, I believe the `stream1.wait(stream2)` API is quite dangerous..

cudax/examples/cccl-rt/heat_diffusion.cu
pciolkosz: I don't like raw pointers, I don't think storing the size in span makes any measurable difference. I would prefer if we promote passing buffers to kernels as spans

cudax/examples/cccl-rt/heat_diffusion.cu
pciolkosz: > The problem is that the event will be always destroyed before the synchronization shall happen.

Once you call `cudaStreamWaitEvent` you can discard the event without problems

davebayer: Whaaat, that's amazing! I will use wait stream in that case

cudax/examples/cccl-rt/heat_diffusion.cu
```diff
@@ -0,0 +1,337 @@
...
+#include <cuda/experimental/container.cuh>
+#include <cuda/experimental/hierarchy.cuh>
+#include <cuda/experimental/launch.cuh>
+#include <cuda/experimental/memory_resource.cuh>
+
+#include <cstdio>
+#include <exception>
+#include <fstream>
+
+/*
```
jrhemstad: Suggestion: It would be helpful for every example to have at the top a succinct but complete TL;DR of what it does.

examples/ccclrt/heat_diffusion/heat_diffusion.cu
```diff
@@ -0,0 +1,337 @@
...
+  // Allocate a pinned host buffer.
+  auto host_buffer =
+    cudax::make_async_buffer<float>(main_stream, cudax::pinned_default_memory_pool(), ngrid_points, cudax::no_init);
+  cuda::std::mdspan<float, cuda::std::dims<2, int>> host_buffer_view{host_buffer.data(), ny, nx};
+
+  // Wait for the allocation to complete before initializing the buffer.
+  main_stream.sync();
+
+  // Initialize the host buffer with values.
+  init_grid(host_buffer_view);
```
jrhemstad, line 276: concern: This feels odd to be teaching people to sync a stream immediately after allocating data on the stream. Can we find a different way to do this?

cudax/examples/cccl-rt/heat_diffusion.cu
```diff
@@ -0,0 +1,337 @@
...
+ *
+ * https://github.com/csc-training/openacc/tree/master/exercises/heat
+ *
+ */
+#include <algorithm>
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <time.h>
```
leofang: suggestion: Change to C++ style stdlib include. Also, `cstdio` is already imported.

examples/ccclrt/heat_diffusion/heat_diffusion.cu
leofang: Indeed. it seems the sync can be removed safely

cudax/examples/cccl-rt/heat_diffusion.cu
```diff
@@ -0,0 +1,337 @@
...
+                  const_cast<const float*>(device_buffer_1.data()),
+                  device_buffer_2.data());
+
+    if (should_output(ti))
+    {
+      // Make the output stream wait for the data to be ready.
+      output_stream.wait(copy_output_event);
+
+      // Enqueue the copy from device to host buffer.
+      cudax::copy_bytes(output_stream, device_buffer_1, host_buffer);
```
leofang: Q: Is it expected that in the 0-th iteration we see the same data (because `device_buffer_1` was just copied to `host_buffer` and it is not changed by the kernel since it's const)?

examples/ccclrt/heat_diffusion/heat_diffusion.cu
```diff
@@ -0,0 +1,337 @@
...
+
+  // Record the end of the computation.
+  const auto end = main_stream.record_timed_event();
+
+  // Wait for all of the asynchronous work to complete.
+  main_stream.sync();
+  output_stream.sync();
+
+  // Get the elapsed time.
+  const auto elapsed_time = end - start;
```
leofang, line 338: Q: We don't need to sync over `end` first?

cudax/examples/cccl-rt/heat_diffusion.cu
```diff
@@ -0,0 +1,337 @@
...
+
+  // Wait for the allocation to complete before initializing the buffer.
+  main_stream.sync();
+
+  // Initialize the host buffer with values.
+  init_grid(host_buffer_view);
+
+  // Allocate device buffers.
+  auto device_buffer_1 = cudax::make_async_buffer<float>(
+    main_stream, cudax::device_default_memory_pool(device), ngrid_points, cudax::no_init);
```
pciolkosz: Suggestion: In case this was missed, there is a `make_async_buffer` overload that takes another buffer:
```
auto make_async_buffer(
  stream_ref __stream, _Resource&& __mr, const async_buffer<_Tp, _SourceProperties...>& __source, const _Env& __env = {})
```
But I could see explicit copies being more educational, so up to you, just wanted to make sure that overload was not missed

cudax/examples/cccl-rt/heat_diffusion.cu
davebayer: Yes! Do you think I should fix that? I didn't want to make it more complicated

examples/ccclrt/heat_diffusion/heat_diffusion.cu
davebayer: No, because we sync the both streams before calculating the elapsed time.

cudax/examples/cccl-rt/heat_diffusion.cu
davebayer: I missed it, thanks!

PR #6366, opened 2025-10-28
change concepts portability macros to avoid use of macro `EXPAND`

ericniebler:
## Description

cusparse is currently not building with cccl trunk. this is because of a conflict with the `EXPANDS` macro that cusparse defines and which appears in cccl's concepts portability macros.

this pr changes the concepts portability macros to avoid using non-uglified macro names.

## Checklist


davebayer:
Do we even need this workaround for MSVC? We require the conforming preprocessor anyway (`/Zc:preprocessor')

ericniebler:
> Do we even need this workaround for MSVC? We require the conforming preprocessor anyway (`/Zc:preprocessor')

users can still use msvc's legacy preprocessor if they define `CCCL_IGNORE_MSVC_TRADITIONAL_PREPROCESSOR_WARNING`. i don't want to be the one that breaks it.

Issue #6369, opened 2025-10-28
[FEA]: Add `max_pool_size` property to `memory_pool_properties`


nirandaperera:
### Is this a duplicate?

- [x] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cccl/issues) for this request and that I agree to the [Code of Conduct](CODE_OF_CONDUCT.md)

### Area

CUDA Experimental (cudax)

### Is your feature request related to a problem? Please describe.

cuda mempool properties has a `maxSize` [property](https://docs.nvidia.com/cuda/cuda-runtime-api/structcudaMemPoolProps.html#structcudaMemPoolProps_12d90e82afb8d577a2aee847d082e145d) but this is not currently being exposed by cccl `memory_pool_properties` struct. 

### Describe the solution you'd like

add a `size_t max_pool_size` field to `memory_pool_properties` struct. 

### Describe alternatives you've considered

NA

PR #6370, opened 2025-10-28
Enabling max pool size for memory pools

nirandaperera:
## Description

This PR enables setting a max pool size for async memory pools. 

Adds a `max_pool_size` field to `memory_pool_properties` struct. 
https://github.com/NVIDIA/cccl/issues/6369

## Checklist

### DIFF
--- a/c/parallel/src/histogram.cu
+++ b/c/parallel/src/histogram.cu
@@ -75,6 +75,7 @@ struct histogram_kernel_source
 {
   cccl_device_histogram_build_result_t& build;
 
+  template <typename PolicyT>
   CUkernel HistogramInitKernel() const
   {
     return build.init_kernel;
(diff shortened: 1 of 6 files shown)


cudax/include/cuda/experimental/__memory_resource/memory_resource_base.cuh
```diff
@@ -267,6 +268,11 @@ struct memory_pool_properties
   __pool_properties.allocType   = __allocation_type;
   __pool_properties.handleTypes = ::CUmemAllocationHandleType(__properties.allocation_handle_type);
   __pool_properties.location    = __location;
+  __pool_properties.maxSize     = __properties.max_pool_size;
```
pciolkosz: I think `maxSize` member was added in 12.2, we need a `#if _CCCL_CTK_AT_LEAST(12, 2)` around the usage. We should probably leave it in the `memory_pool_properties`, just throw if its not 0 for < 12.2

miscco: yeah we should just mark this as something invalid like 0 int his case and then guard all use by `#if _CCCL_CTK_AT_LEAST(12, 2)`

libcudacxx/include/cuda/__memory_resource/memory_resource_base.h
```diff
@@ -253,19 +253,42 @@ struct memory_pool_properties
...
+    if (__properties.initial_pool_size > __properties.max_pool_size)
+    {
+      ::cuda::std::__throw_invalid_argument("Initial pool size must be less than the max pool size");
+    }
+  }
+  __pool_properties.maxSize = __properties.max_pool_size;
+#else
+  if (__properties.max_pool_size != 0)
+  {
+    ::cuda::std::__throw_invalid_argument("Max pool size is not supported on this CUDA driver version");
```
pciolkosz: I would just say CUDA version here, because this is CTK version check

Testing/Temporary/LastTest.log
```diff
@@ -0,0 +1,3 @@
+Start testing: Oct 29 10:52 PDT
```
pciolkosz: Seems like some extra file added by mistake

Testing/Temporary/CTestCostData.txt
```diff
@@ -0,0 +1 @@
+---
```
pciolkosz: Same here, seems like added by mistake

PR #6372, opened 2025-10-29
Refactor Thrust OMP system headers

bernhardmgruber:
I finally had a bit of time to try Cursor, so here are some refactorings, done in three steps:
1. Use C++17 nested namespace definitions
2. Move all function implementations in *.inl files to their headers
3. Update all copyright headers to use the SPDX style

In step 1, a detail namespace got messed up, that I had to fix manually. I diffed all the changes in step 2 and I think the function bodies were correctly moved. I diffed all the copyright header changes and noticed there was one BSD-3 header, which got rewritten to Apache2. But that was on me, my prompt only had an Apache2 example. I fixed it manually.

### DIFF
--- a/thrust/thrust/system/omp/detail/copy_if.h
+++ b/thrust/thrust/system/omp/detail/copy_if.h
@@ -12,6 +12,7 @@
 #elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_MSVC)
 #  pragma system_header
 #endif // no system header
+
 #include <thrust/system/detail/generic/copy_if.h>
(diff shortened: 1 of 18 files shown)


thrust/thrust/system/omp/detail/copy_if.h
```diff
@@ -25,14 +12,11 @@
 #elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_MSVC)
 #  pragma system_header
 #endif // no system header
+#include <thrust/system/detail/generic/copy_if.h>
```
miscco, line 16: Nitpick: missing newline before the header guard

PR #6376, opened 2025-10-29
Use non deprecated methods for `stream_ref` in docs

davebayer:
Use `.sync()` and `.is_done()` instead of deprecated `.wait()` and `.ready()`. I also improved and added tests.

### DIFF
--- a/libcudacxx/test/libcudacxx/cuda/stream_ref/stream_ref.is_done.pass.cpp
+++ b/libcudacxx/test/libcudacxx/cuda/stream_ref/stream_ref.is_done.pass.cpp
@@ -19,28 +19,19 @@
 
 #include "test_macros.h"
 
-_CCCL_SUPPRESS_DEPRECATED_PUSH
-
-void CUDART_CB callback(cudaStream_t, cudaError_t, void* flag)
-{
-  std::chrono::milliseconds sleep_duration{1000};
-  std::this_thread::sleep_for(sleep_duration);
-  assert(!reinterpret_cast<std::atomic_flag*>(flag)->test_and_set());
-}
-
-void test_wait(cuda::stream_ref& ref)
+void test_is_done(cuda::stream_ref& ref)
 {
 #if TEST_HAS_EXCEPTIONS()
   try
   {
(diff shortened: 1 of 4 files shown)


libcudacxx/test/libcudacxx/cuda/stream_ref/stream_ref.is_done.pass.cpp
```diff
@@ -0,0 +1,53 @@
...
+  catch (...)
+  {
+    assert(false && "Should not have thrown");
+  }
+#else
+  assert(ref.is_done());
+#endif // TEST_HAS_EXCEPTIONS()
+}
+
+bool test()
```
bernhardmgruber, line 38: Q: Since the returned value is not used, should this function return `void`?

Applies to other test as well

davebayer: I don't know, I'd like to always return a `bool` since it's useful in constexpr test evaluated as `static_assert(test())`. But here we will probably never have a `constexpr` CUDA stream.

I don't really care too much here, if you want, I can make it return `void`.

libcudacxx/test/libcudacxx/cuda/stream_ref/stream_ref.is_done.pass.cpp
```diff
@@ -0,0 +1,53 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of libcu++, the C++ Standard Library for your entire system,
+// under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-FileCopyrightText: Copyright (c) 2023 NVIDIA CORPORATION & AFFILIATES.
```
miscco: time flies
```suggestion
// SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES.
```

PR #6377, opened 2025-10-29
Inline the Thrust ADL layer

bernhardmgruber:


thrust/thrust/detail/adjacent_difference.inl
```diff
@@ -25,10 +25,23 @@
...
-#include <thrust/system/detail/adl/adjacent_difference.h>
-#include <thrust/system/detail/generic/adjacent_difference.h>
+
 #include <thrust/system/detail/generic/select_system.h>
 
+// Include all active backend system implementations (generic, sequential, host and device)
+#include <thrust/system/detail/generic/adjacent_difference.h>
+#include <thrust/system/detail/sequential/adjacent_difference.h>
+#include __THRUST_HOST_SYSTEM_ALGORITH_HEADER_INCLUDE(detail / adjacent_difference.h)
+#include __THRUST_DEVICE_SYSTEM_ALGORITH_HEADER_INCLUDE(detail / adjacent_difference.h)
```
davebayer: I am not sure if this will work, I believe you will have to add `__THRUST_DEVICE_SYSTEM_ALGORITH_HEADER_INCLUDE` to whitespace sensitive macros in `.clang-format`. I had to deal with the same thing when implementing `_CCCL_HAS_INCLUDE(...)`

bernhardmgruber: I replaced this by a new macro now that already includes the `detail/`. Good idea about adding that to clang-format! I may add that.

davebayer: Okay! The problem with `__has_include` was that it was whitespace sensitive, I am not sure if it is the same case here :)

PR #6383, opened 2025-10-29
Prefix CUB kernel headers with `kernel_`

bernhardmgruber:
This makes the filenames more consistent with the rest of CUB

### DIFF
--- a/c/parallel/src/transform.cu
+++ b/c/parallel/src/transform.cu
@@ -285,7 +285,7 @@ CUresult cccl_device_unary_transform_build_ex(
     std::string final_src = std::format(
       R"XXX(
 #include <cub/device/dispatch/tuning/tuning_transform.cuh>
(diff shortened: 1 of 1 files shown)


fbusato (approved):

bernhardmgruber:
> I am not in favor given that we already have `/kernels/meow.cuh`
> 
> `kernel/kernel_meow..cuh` feels like an unnecessary overkill

We already have:
`agent/agent_meow.cuh`
`block/block_meow.cuh`
`device/device_meow.cuh`
`device/dispatch/dispatch_meow.cuh`
`device/dispatch/tuning/tuning_meow.cuh`
`grid/grid_meow.cuh`
`thread/thread_meow.cuh`
`warp/warp_meow.cuh`

I would agree with your reasoning in isolation. But with the status quo, the kernel headers just violate the established convention.

miscco (approved):
Fine by me ;)

Issue #5160, opened 2025-07-04
Add missing Thrust algorithms


bernhardmgruber:
Thrust currenty misses implementations for the following algorithms operating on ranges of elements. Based on C++23.

From the `<algorithm>` header:
 - [ ] find_end(I, I, I, I)
 - [ ] find_end(I, I, I, I, P)
 - [ ] find_first_of(I, I, I, I)
 - [ ] find_first_of(I, I, I, I, P)
 - [ ] adjacent_find(I, I)
 - [ ] adjacent_find(I, I, P)
 - [ ] mismatch(I, I, I, I)
 - [ ] mismatch(I, I, I, I, P)
 - [ ] equal(I, I, I, I)
 - [ ] equal(I, I, I, I, P)
 - [ ] search
 - [ ] search_n(I, I, C, V)
 - [ ] search_n(I, I, C, V, P)
 - [ ] move(I, I, I)
 - [ ] rotate
 - [ ] rotate_copy
 - [ ] shift_left
 - [ ] shift_right
 - [ ] partial_sort(I, I, I)
 - [ ] partial_sort(I, I, I, P)
 - [ ] partial_sort_copy(I, I, I, I)
 - [ ] partial_sort_copy(I, I, I, I, P)
 - [ ] nth_element(I, I, I)
 - [ ] nth_element(I, I, I, P)
 - [ ] inplace_merge(I, I, I)
 - [ ] inplace_merge(I, I, I, P)
 - [ ] includes(I, I, I, I)
 - [ ] includes(I, I, I, I, P)
 - [ ] is_heap(I, I)
 - [ ] is_heap(I, I, P)
 - [ ] is_heap_until(I, I)
 - [ ] is_heap_until(I, I, P)
 - [ ] lexicographical_compare(I, I, I, I)
 - [ ] lexicographical_compare(I, I, I, I, P)

From the `<numeric>` header:
 - [ ] transform_reduce(I, I, I, T)
 - [ ] transform_reduce(I, I, I, T, B, B)
 - [ ] transform_exclusive_scan(I, I, I, T, B, U)

From the `<memory>` header:
 - [ ] uninitialized_default_construct
 - [ ] uninitialized_default_construct_n
 - [ ] uninitialized_value_construct
 - [ ] uninitialized_value_construct_n
 - [ ] uninitialized_move
 - [ ] uninitialized_move_n
 - [ ] destroy
 - [ ] destroy_n

Algorithms without an execution policy in the standard:
[...]

PR #6388, opened 2025-10-30
Add missing Thrust Search algorithms

charan-003:
Thrust Range-Based Algorithm Implementation Checklist (C++23) based on https://en.cppreference.com/w/cpp/algorithm.html

### Search 
- [x] search(...)

Implements only the search algorithm for now #5160


gonidelis:
Is this PR intended to check all the boxes? If not (hopefully) you can just leave the algorithm(s) that is intended to work on and we can add a link to the EPIC issue. Keep in mind of the PR size - reviewability inverse proportionality theorem if you plan to add more.

bernhardmgruber:
@charan-003 thank you for trying to contribute to CCCL! Could you pleaes focus on one algorithm at a time, so we have an easier time reviewing? Thx!

thrust/thrust/system/detail/generic/search.inl
bernhardmgruber, line 1: Please don't use any `inl` files any more. Just define the function body right in the header with the declaration.

charan-003:
> @charan-003 thank you for trying to contribute to CCCL! Could you pleaes focus on one algorithm at a time, so we have an easier time reviewing? Thx!

@gonidelis @bernhardmgruber 
Thanks for the suggestion! I’ll focus on the search algorithms in this PR and create a new one for the other algorithms.

thrust/thrust/system/detail/generic/search.h
```diff
@@ -0,0 +1,98 @@
...
+      {
+        break;
+      }
+    }
+
+    if (it2 == s_last)
+    {
+      return it;
+    }
+  }
```
bernhardmgruber: Unless I am missing something, this is a fully serial implementation. Please try to come up with a parallel implementation.

bernhardmgruber (changes requested):
Thank you for this attempt! I would really like to see a parallel implementation though, since users would be very surprised to find that `thrust::search` is slower than `std::search` (which uses better than `O(m*n)` algorithms like Boyer Moore).

bernhardmgruber:
I started an internal discussion about this PR a few days ago, because I am not convinced about the proposed algorithm. IIUC, the code will perform N*M reads in the worst case to find the search string of length M in the input of length N. The cache is probably mitigating a lot of the redundant accesses, but I don't know. I like this PR because it adds missing functionality and I can't come up with a good argument for rejecting it, but I also feel like accepting it does not living up to our standard of offering SOL algorithms. I also don't have the time to work on a better implementation.

We concluded that we want to have more data on the performance of the implementation. Fortunately, we seem to already have some benchmarks for the parallel `std::search` implementation of NVHPC (@dkolsen-pgi). I will try to add them.

charan-003:
> I started an internal discussion about this PR a few days ago, because I am not convinced about the proposed algorithm. IIUC, the code will perform N*M reads in the worst case to find the search string of length M in the input of length N. The cache is probably mitigating a lot of the redundant accesses, but I don't know. I like this PR because it adds missing functionality and I can't come up with a good argument for rejecting it, but I also feel like accepting it does not living up to our standard of offering SOL algorithms. I also don't have the time to work on a better implementation.
> 
> We concluded that we want to have more data on the performance of the implementation. Fortunately, we seem to already have some benchmarks for the parallel `std::search` implementation of NVHPC (@dkolsen-pgi). I will try to add them.

I agree the worst-case behaviour is concerning, but without data it’s hard to judge the impact. Benchmarking seems like the right next step. If it helps, I can try adding or extending the NVHPC `std::search` benchmarks and propose a draft, hopefully by next week.

bernhardmgruber:
@charan-003 I just want to let you know that I have not forgotten about this PR! I just didn't have bandwidth yet to add the benchmark.

charan-003:
> @charan-003 I just want to let you know that I have not forgotten about this PR! I just didn't have bandwidth yet to add the benchmark.

@bernhardmgruber  Thank you for your response.

I will be waiting the benchmarks, and in the meantime. I'm focusing on improving my CUDA skills.

PR #6396, opened 2025-10-30
Move the libcu++ specific config

miscco:
This moves the internal libcu++ configuration into its own file

It also splits some of the configurations for e.g thread API and atomimcs into their own file


oleksandr-pavlyk:
One of the modified files, `python/cuda_cccl/build.log`, file was first added by @shwina in 3f37825cad0675eaf3d17b11f0420cd14d757483. What purpose does it serve in the repo?

davebayer:
pre-commit.ci autofix

PR #6399, opened 2025-10-30
Replace `_CCCL_HAS_CUDA_COMPILER()` with `_CCCL_CUDA_COMPILATION()`

davebayer:
In the past, we used to use `_CCCL_HAS_CUDA_COMPILER()` to check whether we are compiling a CUDA source file, which is not what `_CCCL_HAS_CUDA_COMPILER()` actually does. I've replaced the use with mostly `_CCCL_CUDA_COMPILATION()`.

Please, carefully check whether I didn't do a mistake :)

### DIFF
--- a/cub/cub/util_device.cuh
+++ b/cub/cub/util_device.cuh
@@ -754,6 +754,6 @@ private:
 };
 CUB_NAMESPACE_END
 
-#if _CCCL_HAS_CUDA_COMPILER() && !_CCCL_COMPILER(NVRTC)
+#if _CCCL_CUDA_COMPILATION() && !_CCCL_COMPILER(NVRTC)
 #  include <cub/detail/launcher/cuda_runtime.cuh> // to complete the definition of TripleChevronFactory
-#endif // _CCCL_HAS_CUDA_COMPILER() && !_CCCL_COMPILER(NVRTC)
+#endif // _CCCL_CUDA_COMPILATION() && !_CCCL_COMPILER(NVRTC)
--- a/thrust/thrust/random/detail/normal_distribution_base.h
+++ b/thrust/thrust/random/detail/normal_distribution_base.h
@@ -146,11 +146,12 @@ private:
 template <typename RealType>
 struct normal_distribution_base
 {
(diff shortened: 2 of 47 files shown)


cub/cub/util_device.cuh
```diff
@@ -797,6 +797,6 @@ private:
 };
 CUB_NAMESPACE_END
 
-#if _CCCL_HAS_CUDA_COMPILER() && !_CCCL_COMPILER(NVRTC)
+#if _CCCL_HAS_CTK() && !_CCCL_COMPILER(NVRTC)
```
miscco: I am not sure whether that is correct, tripple chevron requires a cuda compiler otherwise there is no sense in launching a kernel

thrust/thrust/random/detail/normal_distribution_base.h
```diff
@@ -148,7 +148,7 @@ class normal_distribution_portable
 template <typename RealType>
 struct normal_distribution_base
 {
-#if _CCCL_HAS_CUDA_COMPILER() && !_CCCL_CUDA_COMPILER(NVHPC)
+#if _CCCL_CUDA_COMPILER(NVCC)
```
fbusato: what about clang-cuda?

cub/cub/util_device.cuh
fbusato: I guess `<cub/detail/launcher/cuda_runtime.cuh>` doesn't rely on CUDA specific syntax. It should be only related to cuda runtime api. The modification is fine in this case

davebayer: @miscco is actually right here, the header include the triple chevron stuff which fail to compile with C++ only compilers

thrust/thrust/system/cuda/detail/core/agent_launcher.h
```diff
@@ -36,7 +36,7 @@
 #  pragma system_header
 #endif // no system header
 
-#if _CCCL_HAS_CUDA_COMPILER()
+#if _CCCL_CUDA_COMPILATION()
```
miscco, line 39: I observe that this is not changing the comment at the end of the block, please adopt

PR #6402, opened 2025-10-30
Update linear_congruential_generator with constexpr, tests and a fast discard

RAMitchell:
## Description

closes 
Part of #5679 

Update linear_congruential_generator with constexpr, stream operators, tests and a fast discard operator for common parameters.


libcudacxx/include/cuda/std/__random/linear_congruential_engine.h
```diff
@@ -236,108 +238,151 @@ class _CCCL_TYPE_VISIBILITY_DEFAULT linear_congruential_engine
...
+      for (; __z; --__z)
+      {
+        (void) operator()();
+      }
+    }
+    else if constexpr (increment == 0)
     {
-      (void) operator()();
+      // minstd_rand and minstd_rand0 can use this efficient implementation
+      uint64_t __acc_mult = 1;
```
fbusato: `::cuda::std::uint64_t` or better  in this case `using ::cuda::std::uint64_t;`

RAMitchell: It was suggested in another PR to not use fully qualified namespaces e.g. for ::cuda::std::size_t because we are already in the cuda::std:: namespace. Is something different here?

RAMitchell: https://github.com/NVIDIA/cccl/pull/6358#discussion_r2469489220

libcudacxx/include/cuda/std/__random/linear_congruential_engine.h
```diff
@@ -190,15 +194,13 @@ struct __lce_ta<__a, __c, __m, static_cast<uint16_t>(~0), __b>
...
-operator<<(basic_ostream<_CharT, _Traits>& __os, const linear_congruential_engine<_Up, _Ap, _Cp, _Np>&);
+_CCCL_API ::std::basic_ostream<_CharT, _Traits>&
+operator<<(::std::basic_ostream<_CharT, _Traits>& __os, const linear_congruential_engine<_Up, _Ap, _Cp, _Np>&);
 
 template <class _CharT, class _Traits, class _Up, _Up _Ap, _Up _Cp, _Up _Np>
-_CCCL_API basic_istream<_CharT, _Traits>&
-operator>>(basic_istream<_CharT, _Traits>& __is, linear_congruential_engine<_Up, _Ap, _Cp, _Np>& __x);
-#endif //
+_CCCL_API ::std::basic_istream<_CharT, _Traits>&
+operator>>(::std::basic_istream<_CharT, _Traits>& __is, linear_congruential_engine<_Up, _Ap, _Cp, _Np>& __x);
```
miscco: Those need to be guarded by 
```suggestion
#if !_CCCL_COMPILER(NVRTC)
template <class _CharT, class _Traits, class _Up, _Up _Ap, _Up _Cp, _Up _Np>
_CCCL_API ::std::basic_ostream<_CharT, _Traits>&
operator<<(::std::basic_ostream<_CharT, _Traits>& __os, const linear_congruential_engine<_Up, _Ap, _Cp, _Np>&);

template <class _CharT, class _Traits, class _Up, _Up _Ap, _Up _Cp, _Up _Np>
_CCCL_API ::std::basic_istream<_CharT, _Traits>&
operator>>(::std::basic_istream<_CharT, _Traits>& __is, linear_congruential_engine<_Up, _Ap, _Cp, _Np>& __x);
#endif // !_CCCL_COMPILER(NVRTC)

```

libcudacxx/include/cuda/std/__random/linear_congruential_engine.h
miscco: Yeah, I believe for anything in `cuda::std::` its better to not use the qualifiers, but in `cuda` we need them

libcudacxx/include/cuda/std/__random/linear_congruential_engine.h
```diff
@@ -236,108 +238,151 @@ class _CCCL_TYPE_VISIBILITY_DEFAULT linear_congruential_engine
...
     {
-      (void) operator()();
+      // minstd_rand and minstd_rand0 can use this efficient implementation
+      uint64_t __acc_mult = 1;
+      uint64_t __cur_mult = multiplier;
+      while (__z > 0)
+      {
+        if (__z & 1)
+        {
+          __acc_mult = (__acc_mult * __cur_mult) % modulus;
```
miscco, line 285: I would prefer if we had this as
```suggestion
          __acc_mult = (__acc_mult * __cur_mult) % modulus;
          if constexpr(increment != 0) {
            __acc_plus = (__acc_plus * __cur_mult + __cur_plus) % modulus;          
          }
```

We would need to annotate `__acc_plus ` with `[[maybe_unused]]` but that seems to be much cleaner

libcudacxx/include/cuda/std/__random/linear_congruential_engine.h
```diff
@@ -236,108 +238,151 @@ class _CCCL_TYPE_VISIBILITY_DEFAULT linear_congruential_engine
...
     return __x.__x_ == __y.__x_;
   }
-  [[nodiscard]] _CCCL_API friend bool
+  [[nodiscard]] _CCCL_API friend constexpr bool
   operator!=(const linear_congruential_engine& __x, const linear_congruential_engine& __y) noexcept
   {
     return !(__x == __y);
   }
 
+  template <typename _CharT, typename _Traits>
```
miscco, line 314: Ditto, needs to be guarded

libcudacxx/test/libcudacxx/std/random/engine/lcg.pass.cpp
```diff
@@ -0,0 +1,75 @@
...
+  E2 e2(seq);
+  for (int i = 0; i < 100; ++i)
+  {
+    auto v1 = e1();
+    auto v2 = e2();
+    assert(v1 == v2);
+  }
+}
+
+void test_against_std()
```
miscco: We should not test against the host library. 

There is a lot of implementation wiggle room that could break us without any benefit

miscco (changes requested):

libcudacxx/include/cuda/std/__random/linear_congruential_engine.h
RAMitchell: I think I can actually remove these.

libcudacxx/include/cuda/std/__random/linear_congruential_engine.h
RAMitchell: Yeah this is nicer.

libcudacxx/include/cuda/std/__random/linear_congruential_engine.h
```diff
@@ -236,158 +230,164 @@ class _CCCL_TYPE_VISIBILITY_DEFAULT linear_congruential_engine
...
+      {
+        (void) operator()();
+      }
+    }
+    else
+    {
+      uint64_t __acc_mult                  = 1;
+      [[maybe_unused]] uint64_t __acc_plus = 0;
+      uint64_t __cur_mult                  = multiplier;
+      uint64_t __cur_plus                  = increment;
```
miscco:
```suggestion
      [[maybe_unused]] uint64_t __cur_plus                  = increment;
```

libcudacxx/include/cuda/std/__random/linear_congruential_engine.h
```diff
@@ -236,158 +230,164 @@ class _CCCL_TYPE_VISIBILITY_DEFAULT linear_congruential_engine
...
+      {
+        (void) operator()();
+      }
+    }
+    else
+    {
+      uint64_t __acc_mult                  = 1;
+      [[maybe_unused]] uint64_t __acc_plus = 0;
+      uint64_t __cur_mult                  = multiplier;
+      [[maybe_unused]] uint64_t __cur_plus                  = increment;
```
miscco:
```suggestion
      uint64_t __acc_mult                  = 1;
      [[maybe_unused]] uint64_t __acc_plus = 0;
      uint64_t __cur_mult                  = multiplier;
      [[maybe_unused]] uint64_t __cur_plus = increment;
```

RAMitchell:
@miscco test_engine has the constexpr evaluations inside it. Its already testing constexpr for c++20 - if I understand you correctly.

libcudacxx/include/cuda/std/__random/linear_congruential_engine.h
fbusato: you are right. I missed it

libcudacxx/include/cuda/std/__random/linear_congruential_engine.h
```diff
@@ -236,158 +230,164 @@ class _CCCL_TYPE_VISIBILITY_DEFAULT linear_congruential_engine
...
+  _CCCL_API constexpr void seed(_Sseq& __q) noexcept
   {
     __seed(__q,
            integral_constant<uint32_t,
                              1 + (__m == 0 ? (sizeof(result_type) * CHAR_BIT - 1) / 32 : (__m > 0x100000000ull))>());
   }
 
   // generating functions
-  [[nodiscard]] _CCCL_API result_type operator()() noexcept
+  _CCCL_API constexpr result_type operator()() noexcept
```
fbusato, line 259: are we intentionally dropping `[[nodiscard]]`? is it only because we are using it internally? or is an expected usage from users?

libcudacxx/include/cuda/std/__random/linear_congruential_engine.h
fbusato: could we please use the same convention for all parameters?

libcudacxx/include/cuda/std/__random/linear_congruential_engine.h
RAMitchell: I was not a fan of nodiscard on operator() because a valid use is to advance the state of the engine without using the result. E.g. knuth has algorithms where a generator discards values after seeding. It was pointed out that we can also use discard for this use case. My view is that nodiscard here is overly strict but I don't mind too much - if we do nodiscard I will do it for every engine.

libcudacxx/include/cuda/std/__random/linear_congruential_engine.h
RAMitchell: Not sure I understand sorry, which convention and which parameters specifically?

fbusato: sorry, we really need to add the conventions to the CCCL Development Guide. 
Anyway, CCCL uses `_PascalStyle` for template parameters.

miscco:
> approved. One last suggestion, could you please change `__A` to single underscore `_A` for template parameters?

I am not in favor, single letter `_A` macros are frequently used by the compiler. Dont ask me why, but for example `_T` is used by MSVC to convert string literals

RAMitchell:
Will merge as is according to @miscco `s comment.

PR #6404, opened 2025-10-30
Fix clang 21 issues

davebayer:
Fixes nvbug 5535419

### DIFF
--- a/libcudacxx/test/libcudacxx/std/strings/string.view/string.view.deduct/range.pass.cpp
+++ b/libcudacxx/test/libcudacxx/std/strings/string.view/string.view.deduct/range.pass.cpp
@@ -26,8 +26,7 @@ __host__ __device__ constexpr void test_range_deduct()
 {
   // 1. Test construction of a string_view from an cuda::std::array
   {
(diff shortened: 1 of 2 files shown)


bernhardmgruber (approved):

libcudacxx/test/libcudacxx/std/strings/string.view/string.view.deduct/range.pass.cpp
```diff
@@ -26,8 +26,7 @@ __host__ __device__ constexpr void test_range_deduct()
 {
   // 1. Test construction of a string_view from an cuda::std::array
   {
-    constexpr auto str = TEST_STRLIT(CharT, "test");
-    cuda::std::array<CharT, 4> val{str[0], str[1], str[2], str[3]};
+    cuda::std::array<CharT, 4> val{};
```
miscco, line 29: This is not equivalent, we want something non empty

davebayer: But is, we don't need the members to have any specific value, we care about the range size and address.

Issue #6298, opened 2025-10-20
[FEA]: Disable NVTX ranges by default for `thrust::seq` Thrust algorithms


jrhemstad:
### Is this a duplicate?

- [x] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cccl/issues) for this request and that I agree to the [Code of Conduct](CODE_OF_CONDUCT.md)

### Area

Thrust

### Is your feature request related to a problem? Please describe.

As a user of CCCL's `thrust::` algorithms and containers with a non-CUDA backend, it is unlikely I am interested in capturing NVTX ranges for the execution of that algorithm. Furthermore, capturing an NVTX range is much more likely to introduce overhead as the cost starting/stopping the range is small relative to a kernel launch, but can be large compared to CPU-only work.

### Describe the solution you'd like

I'd like NVTX ranges to not be captured by default when a `thrust::` algorithm is invoked with an execution policy other than `thrust::cuda::par`. 

### Additional context

We could consider adding a compile-time flag to enable capturing Thrust host ranges by default, but I think YAGNI applies here.

bernhardmgruber:
I assume we want to have NVTX ranges for any execution policy derived from `thrust::cuda_cub::execution_policy`. That includes `cuda::tag{}`, `cuda::par_nosync`, `cuda::par.on(...)` (which is a different type than `decltype(cuda::par)`).

gonidelis:
@bernhardmgruber  
>  I assume we want have NVTX ranges

you mean "we want to have", right? and not "we won't have".

bernhardmgruber:
> you mean "we want to have", right? and not "we won't have"

Yes. I want to remain with the status quo and emit NVTX ranges for any execution policy that targets the CUDA backend.

gonidelis:
@bernhardmgruber perf the fixing PR is doing just that by disabling only `thrust::seq` and its derivatives

PR #6415, opened 2025-10-31
MVP for disabling nvtx ranges for `thrust::seq`

gonidelis:
fixes https://github.com/NVIDIA/cccl/issues/6298

by disabling NVTX ranges for `thrust::seq`. This PR specifically tracks the thrust algorithms involved in `host_vector::push_back()` operations and disables NVTX ranges just for these. 

The disabling basically evaluates at compile time whether the policy passed to the thrust algorithm (and subsequently to `should_enable_nvtx_for_policy()`) is `thrust::seq` or any derivatives using CRTP.

Upon approval of the utility provided here, I will add all the remaining thrust algorithms.

_I did isolate and run profiles on the [cudf code that pointed to the problem](https://github.com/rapidsai/cudf/blob/2ce0a1d185e9daf72aa1982e494d3466cb320eb1/cpp/src/io/parquet/reader_impl_chunking_utils.cu#L547-L550) and after the changes I got zero nvtx ranges for these lines in the profile._

### DIFF
--- a/thrust/thrust/detail/nvtx_policy.h
+++ b/thrust/thrust/detail/nvtx_policy.h
@@ -20,14 +20,12 @@ THRUST_NAMESPACE_BEGIN
 // Forward declarations
 namespace system::detail::sequential
 {
-struct tag;
 template <class>
 struct execution_policy;
 } // namespace system::detail::sequential
 
 namespace detail
 {
-
 // Helper to determine if NVTX should be enabled for a given policy
 // NVTX is DISABLED only for thrust::seq and any policy derived from sequential::execution_policy
 // ENABLED for all other policies (CUDA, OMP, TBB, etc.)
@@ -38,7 +36,6 @@ inline constexpr bool should_enable_nvtx_for_policy()
   // This catches thrust::seq, cpp::tag, and any other sequential-based policy
   return !::cuda::std::is_base_of_v<thrust::system::detail::sequential::execution_policy<Policy>, Policy>;
 }
-
 } // namespace detail
 
 THRUST_NAMESPACE_END


thrust/thrust/detail/nvtx_policy.h
```diff
@@ -0,0 +1,48 @@
+// SPDX-FileCopyrightText: Copyright (c) 2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: BSD-3-Clause
```
bernhardmgruber: Critical: new files must be:
```suggestion
// SPDX-FileCopyrightText: Copyright (c) 2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
```

thrust/thrust/detail/nvtx_policy.h
```diff
@@ -0,0 +1,48 @@
...
+// ENABLED for all other policies (CUDA, OMP, TBB, etc.)
+template <typename DerivedPolicy>
+inline constexpr bool should_enable_nvtx_for_policy()
+{
+  using Policy = typename ::cuda::std::remove_cv<typename ::cuda::std::remove_reference<DerivedPolicy>::type>::type;
+
+  // Check if Policy is derived from sequential::execution_policy
+  // This catches thrust::seq, cpp::tag, and any other sequential-based policy
+  return !::cuda::std::is_base_of<thrust::system::detail::sequential::execution_policy<Policy>, Policy>::value
+      && !::cuda::std::is_same<Policy, thrust::system::detail::sequential::tag>::value;
```
bernhardmgruber: Suggestion: I think this should just be:
```suggestion
  using Policy = ::cuda::std::decay_t<DerivedPolicy>;

  // Check if Policy is derived from sequential::execution_policy
  // This catches thrust::seq, cpp::tag, and any other sequential-based policy
  return !::cuda::std::is_base_of_v<thrust::system::detail::sequential::execution_policy<Policy>, Policy>;
```

thrust/thrust/detail/nvtx_policy.h
```diff
@@ -0,0 +1,48 @@
...
+namespace detail
+{
+
+// Helper to determine if NVTX should be enabled for a given policy
+// NVTX is DISABLED only for thrust::seq and any policy derived from sequential::execution_policy
+// ENABLED for all other policies (CUDA, OMP, TBB, etc.)
+template <typename DerivedPolicy>
+inline constexpr bool should_enable_nvtx_for_policy()
+{
+  using Policy = typename ::cuda::std::remove_cv<typename ::cuda::std::remove_reference<DerivedPolicy>::type>::type;
```
bernhardmgruber: Q: Why do we need to strip reference and CV qualifiers? I don't understand how `DerivedPolicy` could be qualified.

thrust/thrust/detail/copy.inl
```diff
@@ -79,7 +80,7 @@ _CCCL_HOST_DEVICE OutputIterator two_system_copy(
   InputIterator last,
   OutputIterator result)
 {
-  _CCCL_NVTX_RANGE_SCOPE("thrust::two_system_copy");
+  _CCCL_NVTX_RANGE_SCOPE_IF(should_enable_nvtx_for_policy<System1>(), "thrust::two_system_copy");
```
bernhardmgruber: Important: If multiple systems are in use, please only disable the range if all systems are sequential. For example:
```suggestion
  _CCCL_NVTX_RANGE_SCOPE_IF(should_enable_nvtx_for_policy<System1>() || should_enable_nvtx_for_policy<System2>(), "thrust::two_system_copy");
```
Applies to all APIs taking multiple systems.

thrust/thrust/detail/nvtx_policy.h
gonidelis: It was a form of guarding against custom policy wrappers that might come from the users. Not needed if our in house policies are used per our programming guide though, indeed.

thrust/thrust/detail/copy.inl
gonidelis: good catch

thrust/thrust/detail/nvtx_policy.h
bernhardmgruber: I think if `DerivedPolicy` was qualified then `thrust::derived_cast` may no longer work (or so something completely buggy). I think we are save in assuming `DerivedPolicy` is not qualified.

thrust/thrust/detail/copy.inl
```diff
@@ -115,8 +118,8 @@ _CCCL_HOST_DEVICE OutputIterator two_system_copy_n(
 template <typename InputIterator, typename OutputIterator>
 OutputIterator copy(InputIterator first, InputIterator last, OutputIterator result)
 {
-  _CCCL_NVTX_RANGE_SCOPE("thrust::copy");
   using System1 = typename thrust::iterator_system<InputIterator>::type;
+  _CCCL_NVTX_RANGE_SCOPE_IF(detail::should_enable_nvtx_for_policy<System1>(), "thrust::copy");
   using System2 = typename thrust::iterator_system<OutputIterator>::type;
```
bernhardmgruber, line 120: Important: Please regard both systems here as well.

thrust/thrust/detail/copy.inl
```diff
@@ -128,8 +131,8 @@ OutputIterator copy(InputIterator first, InputIterator last, OutputIterator resu
 template <typename InputIterator, typename Size, typename OutputIterator>
 OutputIterator copy_n(InputIterator first, Size n, OutputIterator result)
 {
-  _CCCL_NVTX_RANGE_SCOPE("thrust::copy_n");
   using System1 = typename thrust::iterator_system<InputIterator>::type;
+  _CCCL_NVTX_RANGE_SCOPE_IF(detail::should_enable_nvtx_for_policy<System1>(), "thrust::copy_n");
   using System2 = typename thrust::iterator_system<OutputIterator>::type;
```
bernhardmgruber, line 135: Here as well.

bernhardmgruber (commented):
Please fix the remaining two comments, otherwise LGTM.

thrust/thrust/detail/nvtx_policy.h
```diff
@@ -0,0 +1,45 @@
...
+#  pragma GCC system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_CLANG)
+#  pragma clang system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_MSVC)
+#  pragma system_header
+#endif // no system header
+
+#include <cuda/std/__type_traits/is_base_of.h>
+#include <cuda/std/__type_traits/remove_cv.h>
+#include <cuda/std/__type_traits/remove_reference.h>
```
bernhardmgruber:
```suggestion
#include <cuda/std/__type_traits/decay.h>
```

thrust/thrust/detail/copy.inl
gonidelis: shoot missed them cause they no template args

gonidelis:
@bernhardmgruber @miscco now that I am thinking. Is a test expected for these facilities? And should it be under libcu++ better of maybe? Could be in `libcudacxx/include/cuda/__nvtx/nvtx.h` 👀

bernhardmgruber:
> @bernhardmgruber @miscco now that I am thinking. Is a test expected for these facilities?

Good point! Please add a test if you come up with a good idea on how to test it.

PR #6420, opened 2025-10-31
Prepare `cudax::host_lauch` migration to libcu++

davebayer:


cudax/test/launch/host_launch.cu
```diff
@@ -0,0 +1,226 @@
...
+  void operator()(MoveOnlyArg) {}
+
+private:
+  MoveOnlyCallable() = default;
+};
+
+C2H_TEST("Host launch", "")
+{
+  cuda::device_ref device{0};
+  device.init();
```
pciolkosz, line 113: Does it fail without an explicit init? If yes we might need to fix this

davebayer: There was a failing test that held a reference to a local variable that went out of scope. This is what left from the debugging. I'll keep it for now as we will touch the tests soon

PR #6424, opened 2025-10-31
Use `if consteval` in libcu++

davebayer:
In C++23, `if consteval` construct was added to the language which should replace `if (std::is_constant_evaluated())`.

This PR adds set of `_CCCL_IF_[NOT_]CONSTEVAL_[DEFAULT]` macros to be used instead of `cuda::std::is_constant_evaluated()` and `cuda::std::__default_is_constant_evaluated()`.

Even thought this is a C++23 feature, many compilers support it in C++20 as an extension. NVCC tolerates `if consteval` since 13.0.


libcudacxx/include/cuda/std/__cccl/dialect.h
```diff
@@ -22,6 +22,8 @@
 #  pragma system_header
 #endif // no system header
 
+#include <cuda/std/__cccl/dialect.h>
+
```
miscco: That looks circular ^^

libcudacxx/include/cuda/__memory/ptr_in_range.h
```diff
@@ -73,7 +72,7 @@ template <typename _Tp>
 template <typename _Tp>
 [[nodiscard]] _CCCL_API constexpr bool ptr_in_range(_Tp* __ptr, _Tp* __start, _Tp* __end) noexcept
 {
-  if (::cuda::std::__cccl_default_is_constant_evaluated())
+  _CCCL_IF_CONSTEVAL
```
davebayer:
```suggestion
  _CCCL_IF_CONSTEVAL_DEFAULT
```

libcudacxx/include/cuda/std/__type_traits/is_constant_evaluated.h
```diff
@@ -24,25 +24,28 @@
...
 {
-  return _CCCL_BUILTIN_IS_CONSTANT_EVALUATED();
+  _CCCL_IF_CONSTEVAL
+  {
+    return true;
+  }
+  else
+  {
+    return false;
+  }
```
miscco: I do not want this. The standard is pretty clear that we should use the builtin

davebayer: Reverted the changes

libcudacxx/include/cuda/std/__cccl/host_std_lib.h
```diff
@@ -23,9 +23,9 @@
 #  pragma system_header
 #endif // no system header
 
-#define _CCCL_HOST_STD_LIB_LIBSTDCXX() 0
-#define _CCCL_HOST_STD_LIB_LIBCXX()    0
-#define _CCCL_HOST_STD_LIB_STL()       0
+#define _CCCL_HOST_STD_LIB_LIBSTDCXX() _CCCL_VERSION_INVALID()
```
fbusato, line 26: would be better to make these changes in a new PR

libcudacxx/include/cuda/std/__cccl/dialect.h
```diff
@@ -127,4 +130,65 @@
...
+#elif defined(_CCCL_BUILTIN_IS_CONSTANT_EVALUATED)
+#  if _CCCL_HOST_COMPILATION()
+#    define _CCCL_BEGIN_IF_CONSTEVAL_SUPPRESS() _CCCL_DIAG_PUSH _CCCL_DIAG_SUPPRESS_GCC("-Wtautological-compare")
+#    define _CCCL_END_IF_CONSTEVAL_SUPPRESS()   _CCCL_DIAG_POP
+#  else // ^^^ _CCCL_HOST_COMPILATION() ^^^ / vvv !_CCCL_HOST_COMPILATION() vvv
+#    define _CCCL_BEGIN_IF_CONSTEVAL_SUPPRESS()
+#    define _CCCL_END_IF_CONSTEVAL_SUPPRESS()
+#  endif // ^^^ !_CCCL_HOST_COMPILATION() ^^^
+
+#  define _CCCL_IF_CONSTEVAL \
```
fbusato, line 165: what be possible to have a function macro?

libcudacxx/include/cuda/std/__type_traits/is_constant_evaluated.h
```diff
@@ -24,25 +24,23 @@
...
-}
 #else // ^^^ _CCCL_BUILTIN_IS_CONSTANT_EVALUATED ^^^ / vvv !_CCCL_BUILTIN_IS_CONSTANT_EVALUATED vvv
-_CCCL_API constexpr bool is_constant_evaluated() noexcept
-{
   return false;
+#endif // ^^^ !_CCCL_BUILTIN_IS_CONSTANT_EVALUATED ^^^
 }
-_CCCL_API constexpr bool __cccl_default_is_constant_evaluated() noexcept
+
+[[nodiscard]] _CCCL_API constexpr bool __cccl_default_is_constant_evaluated() noexcept
```
fbusato, line 36: do we still use this function after `consteval` introduction?

fbusato (approved):
looks great. Would be nice to have a function macro.

minor comment (I made the same mistake many times). It's fine to have small, unrelated changes this time, but please try to submit independent PRs.

libcudacxx/include/cuda/std/__cccl/host_std_lib.h
davebayer: Yeaaah, I know, I can move it to a separate PR

libcudacxx/include/cuda/std/__cccl/dialect.h
davebayer: Hm, I intentionally kept it like this to mimic the `if consteval {}` syntax. And we don't call it inside any macro preprocessor condition, so it should be safe to use, too.

Would you prefer it to be the function macro?

libcudacxx/include/cuda/std/__type_traits/is_constant_evaluated.h
davebayer: These functions are still used in tests, maybe we can just move it to `test/support`

libcudacxx/include/cuda/std/__cccl/dialect.h
fbusato: yeah, I would feel safer with the function macro (if not too much work)

libcudacxx/include/cuda/std/__type_traits/is_constant_evaluated.h
davebayer: I'll keep it for now, it's still in use in `cuda::std::__string_view` in some weird construct

libcudacxx/include/cuda/std/__cccl/dialect.h
davebayer: @fbusato I decided to go for the variant without the curly braces for now, I think it's more readable. We can always change it later :)

PR #6426, opened 2025-10-31
Refactor Thrust generic system headers

bernhardmgruber:
* move `*.inl` file content to their headers
* use SPDX license headers


bernhardmgruber:
Alright, a big problem here is that many generic thrust algorithm implementation headers include themselves again transitively. So the current split in header+forward declaration and inc file+lots of includes+implementation avoids the problem by declaring the generic implementation early enough.

bernhardmgruber:
I will replace this PR with a different approach.

bernhardmgruber:
I'll leave this for now.

PR #5371, opened 2025-07-25
Complex sqrt accuracy/speed improvements

s-oboyle:
## Updated the complex sqrt function to avoid numerical issues.

There are some issues with the current complex sqrt function.

It does work not around overflow or underflow, so anywhere `|x|` under/overflows `sqrt` will too, while mathematically all finite values of `x` should lead to finite values of `sqrt(x)`.

It also does not account for catastrophic cancellation issues that can arise in normal/non-extremal circumstances, where there can be large accuracy losses.

For example, in fp64 we have `sqrt(-1) = 6.123233996e-17 + i`

This new version fixes these accuracy issues, and gives a speedup.

## Perf
On all tested systems, H100 and Q8000, CPU, there was speedups.

Using the math-teams standard `math_bench` test we have the following on H100 (averaged summary):
**Operations/SM/cycle:**
`csqrt():`
| H100 |  old | new | new/old |
| --- | --- | --- | --- |
| fp64 | 0.35614 | 0.94882 | 2.664 | 
| fp32 | 0.7396 | 1.78346 | 2.411 |
| half | 0.54516 | 1.62066 | 2.973 |
| bf16 | 0.54958 | 1.58898 | 2.891 |

(See https://docs.google.com/spreadsheets/d/1TdOpEbLgoL1QOWKjeO3pwFDDMoPr4iqnPjz_XrL7UEA for raw/non-averaged info. NV internal.)

# GPU Correctness
For the new version, an intensive bracket and bisect search, along with testing special hard values, gives:
```
GPU fp64:
Max ulp real error (3.277,1.172) @ (-1.63385718e+151,-2.682543403e+154) (0xdf53f71f22de2422,0xe000017febfc2db5)
        Ours = (1.157780213e+77,-1.158485597e+77)    Ref = (1.157780213e+77,-1.158485597e+77)
        Ours = (0x4fefff0135f075a6,0xcff001ff60eba13c)               Ref = (0x4fefff0135f075a9,0xcff001ff60eba13b)

Max ulp imag error (1.154,3.218) @ (7.565676833e+183,5.473795433e+185)  (0x661c7d11e8ada443,0x66801a50ab08f572)
        Ours = (5.267815024e+92,5.195508392e+92)    Ref = (5.267815024e+92,5.195508392e+92)
        Ours = (0x533029a1a415d8b3,0x532fe1acf50ce33d)               Ref = (0x533029a1a415d8b2,0x532fe1acf50ce340)
```

```
GPU fp32:
[...]

PR #6428, opened 2025-10-31
Complex asinh accuracy refinement

s-oboyle:
## Update the complex asinh function to avoid numerical issues.

The current complex asinh function loses accuracy in several places.
These mostly relate to over/underflow, and catastrophic cancellation for tough values.

This new version fixes these accuracy issues while retaining it's perf.

## Perf
On GH100 we don't have much perf difference. (There used to be a much large perf gap until #5371 got merged, which the current version is availing of).
Using the math-teams standard math_bench test we have the following:

**Operations/SM/cycle:**
`casinh():`
| --- | --- | --- | --- |
| fp64 | 0.2531 | 0.2549 | 1.01 | 
| fp32 | 0.6072 | 0.6334 | 1.04 |

## Correctness
Apart from the usual over/underflow suspects, there is also some very subtle intervals where accuracy gets badly thrown out, especially by catastrophic cancellation very close to `+-i`.

### GPU Correctness
```
GPU fp64:
Max ulp real error (4.867,1.742) @ (0.007757045272,-0.0002247045536)    (0x3f7fc5d9fc1f5662,0xbf2d73d56affd72d)
        Ours = (0.007756967678,-0.0002246977954)    Ref = (0.007756967678,-0.0002246977954)
        Ours = (0x3f7fc5c527dd1d58,0xbf2d739b5d7a3961)               Ref = (0x3f7fc5c527dd1d53,0xbf2d739b5d7a3963)

Max ulp imag error (0.1719,5.453) @ (7.198570162e+103,-5.623976789e+101)        (0x558011effdb5a3ad,0xd510120190e898df)
        Ours = (239.8333247,-0.007812471425)    Ref = (239.8333247,-0.007812471425)
        Ours = (0x406dfaaa988c99ba,0xbf7ffff854599f01)               Ref = (0x406dfaaa988c99ba,0xbf7ffff854599efc)
```

```
GPU fp32:
Max ulp real error (6.619,2.232) @ (0.007812378462,-0.0007928675623)    (0x3bfffefb,0xba4fd871)
        Ours = (0.007812298369,-0.0007928435807)    Ref = (0.007812301628,-0.0007928434643)
        Ours = (0x3bfffe4f,0xba4fd6d5)               Ref = (0x3bfffe56,0xba4fd6d3)

Max ulp imag error (3.732,5.528) @ (0.007806597743,3.029832988e-05)     (0x3bffce7d,0x37fe292b)
        Ours = (0.007806516718,3.029741674e-05)    Ref = (0.007806518581,3.029740583e-05)
        Ours = (0x3bffcdcf,0x37fe2735)               Ref = (0x3bffcdd3,0x37fe272f)
```

### CPU Correctness
```
CPU fp64:
Max ulp real error (4.125,0) @ (0.01542893159,-0)       (0x3f8f993424afec00,0x8000000000000000)
        Ours = (0.01542831951,-0)    Ref = (0.01542831951,-0)
        Ours = (0x3f8f98e1fdb37251,0x8000000000000000)               Ref = (0x3f8f98e1fdb3724d,0x8000000000000000)

Max ulp imag error (0.5078,3.484) @ (0.8869326854,-1.12001698e-254)     (0x3fec61c0a7b18800,0x8b3505783ad41800)
        Ours = (0.7991224705,-8.379245502e-255)    Ref = (0.7991224705,-8.379245502e-255)
        Ours = (0x3fe99269498d3a37,0x8b2f742347588681)               Ref = (0x3fe99269498d3a36,0x8b2f742347588684)
```

```
CPU fp32:
Max ulp real error (4.827,0.5125) @ (0.001131535857,0.9893865585)       (0x3a94500b,0x3f7d4870)
        Ours = (0.007776218932,1.424767017)    Ref = (0.007776216604,1.424766898)
        Ours = (0x3bfecfa7,0x3fb65ec4)               Ref = (0x3bfecfa2,0x3fb65ec3)

Max ulp imag error (0.498,4.695) @ (-5.695841894e+11,3.510264218e+10)   (0xd3049ddd,0x5102c47d)
        Ours = (-27.76321411,0.06155067682)    Ref = (-27.76321411,0.06155069545)
        Ours = (0xc1de1b10,0x3d7c1c90)               Ref = (0xc1de1b10,0x3d7c1c95)
```


libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
```diff
@@ -21,69 +21,352 @@
...
+  //     x_(n+1) = x_n - 0.5*x_n*((__hi + __lo)(x_n^2) - 1)
+
+  // __initial_guess^2:
+  const _Tp __init_sq_hi = __initial_guess * __initial_guess;
+  const _Tp __init_sq_lo = ::cuda::std::fma(__initial_guess, __initial_guess, -__init_sq_hi);
+
+  // Times (__hi + __lo).
+  // We need to add the -1.0 here in an fma, or different compilers (eg host vs device)
+  // optimize differently and give (sometimes very) different results.
+  const _Tp _hi_hi_hi = ::cuda::std::fma(__hi, __init_sq_hi, -1.0);
```
davebayer:
```suggestion
  const _Tp _hi_hi_hi = ::cuda::std::fma(__hi, __init_sq_hi, _Tp(-1.0));
```
otherwise we would get a conversion to `double`

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
```diff
@@ -21,69 +21,352 @@
...
+  // Low part not needed for fp32/fp64, but might be if this is extended to other types:
+  // _Tp _hi_hi_lo = fma(__hi, __init_sq_hi, -_hi_hi_hi);
+
+  // Add all terms
+  const _Tp __full_term = _hi_hi_hi + (::cuda::std::fma(__lo, __init_sq_hi, __hi * __init_sq_lo) /*+ _hi_hi_lo*/);
+
+  const _Tp __correction_term = -0.5 * __initial_guess * __full_term;
```
davebayer:
```suggestion
  const _Tp __correction_term = _Tp(-0.5) * __initial_guess * __full_term;
```

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
```diff
@@ -21,69 +21,352 @@
...
+  // We can have two slightly different paths depending on whether rsqrt is available
+  // or not, aka are we on device or host.
+
+  const _Tp __recip_sqrt = __internal_rsqrt_inverse_hyperbloic<_Tp>(__inside_sqrt_term);
+  _Tp __pos_evaluation_real;
+
+  // This reuses the sqrt calculated on CPU already in __recip_sqrt,
+  // And gets sqrt quickly on device using the rsqrt already calculated.
+  NV_IF_ELSE_TARGET(NV_IS_DEVICE, __pos_evaluation_real = (__recip_sqrt * __inside_sqrt_term);
+                    , __pos_evaluation_real             = ::cuda::std::sqrt(__inside_sqrt_term);)
```
davebayer:
```suggestion
  NV_IF_ELSE_TARGET(NV_IS_DEVICE, (__pos_evaluation_real = (__recip_sqrt * __inside_sqrt_term);), (__pos_evaluation_real = ::cuda::std::sqrt(__inside_sqrt_term);))
```

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
miscco: Should we rather do
```suggestion
  const _Tp _hi_hi_hi = ::cuda::std::fma(__hi, __init_sq_hi, _Tp{-1.0});
```

Because we prefer `direct-list-initialization` when possible.

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
```diff
@@ -21,69 +21,352 @@
...
+{
+  if constexpr (is_same_v<_Tp, float>)
+  {
+    NV_IF_TARGET(NV_IS_DEVICE, (return ::rsqrtf(__x);))
+  }
+  if constexpr (is_same_v<_Tp, double>)
+  {
+    NV_IF_TARGET(NV_IS_DEVICE, (return ::rsqrt(__x);))
+  }
+  return _Tp(1) / ::cuda::std::sqrt(__x);
```
miscco:
```suggestion
  return _Tp{1} / ::cuda::std::sqrt(__x);
```

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
```diff
@@ -21,69 +21,352 @@
...
+[[nodiscard]] _CCCL_API inline _Tp __internal_rsqrt_inverse_hyperbloic(_Tp __x) noexcept
+{
+  if constexpr (is_same_v<_Tp, float>)
+  {
+    NV_IF_TARGET(NV_IS_DEVICE, (return ::rsqrtf(__x);))
+  }
+  if constexpr (is_same_v<_Tp, double>)
+  {
+    NV_IF_TARGET(NV_IS_DEVICE, (return ::rsqrt(__x);))
+  }
```
miscco, line 57: We can only use those when compiling with a cuda compiler and moreso in cuda compilation mode
```suggestion
#if _CCCL_CUDA_COMPILATION()
  if constexpr (is_same_v<_Tp, float>)
  {
    NV_IF_TARGET(NV_IS_DEVICE, (return ::rsqrtf(__x);))
  }
  if constexpr (is_same_v<_Tp, double>)
  {
    NV_IF_TARGET(NV_IS_DEVICE, (return ::rsqrt(__x);))
  }
#endif // _CCCL_CUDA_COMPILATION()
```

Another question, do we need to include some cuda headers for that or are those compiler intrinsics?

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
```diff
@@ -21,69 +21,352 @@
...
+  return _Tp(1) / ::cuda::std::sqrt(__x);
+}
+
+// An unsafe sqrt(_Tp + _Tp) extended precision sqrt.
+template <typename _Tp>
+static void __device__ __host__ __forceinline__
```
miscco: We cannot use `__device__ __host__`, so this needs to be
```suggestion
static void _CCCL_API _CCCL_FORCEINLINE
```

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
```diff
@@ -21,69 +21,352 @@
...
+__internal_double_Tp_sqrt_unsafe(_Tp __hi, _Tp __lo, _Tp* __out_hi, _Tp* __out_lo) noexcept
```
miscco: Nitpick: I would strongly prefer if we would return a simple struct here instead of inout parameters

```cpp
template<class _Tp>
struct _CCCL_ALIGNAS(2 * sizeof(_Tp)) __cccl_sqrt_return_hilo {
  __Tp __hi;
  __Tp __low;
};
```

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
s-oboyle: I didn't realize this is a use case, there is also existing `NV_IF_TARGET` code[ in sqrt](https://github.com/NVIDIA/cccl/blob/93e1f87c25249cdf391091e0e56d2d52a6f00aac/libcudacxx/include/cuda/std/__complex/roots.h#L47).

`rsqrt `is kinda an intrinsic, though perhaps technically not. The math definition header basically get silently included by nvcc in every compilation.

miscco: This is less about `nvcc` and more about `clang-cuda`

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
s-oboyle: clang-format might say no.

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
s-oboyle: done

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
```diff
@@ -21,69 +21,368 @@
...
-#include <cuda/std/__cmath/isnan.h>
-#include <cuda/std/__cmath/trigonometric_functions.h>
-#include <cuda/std/__complex/complex.h>
-#include <cuda/std/__complex/exponential_functions.h>
-#include <cuda/std/__complex/logarithms.h>
 #include <cuda/std/__complex/nvbf16.h>
 #include <cuda/std/__complex/nvfp16.h>
 #include <cuda/std/__complex/roots.h>
-#include <cuda/std/limits>
-#include <cuda/std/numbers>
```
davebayer, line 36: Why are all of these includes dropped?

s-oboyle: My prejudice basically, in math we get compile time build bugs because of header parsing so I usually try take them out if they're not needed. I was over zealous here it seems, not sure why this even built on my machine. Added them back, like I would probably have to do later after changing the other functions also.

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
```diff
@@ -39,51 +39,360 @@
...
+template <class _Tp>
+struct _CCCL_ALIGNAS(2 * sizeof(_Tp)) __cccl_asinh_sqrt_return_hilo
+{
+  _Tp __hi;
+  _Tp __lo;
+};
+
+// An unsafe sqrt(_Tp + _Tp) extended precision sqrt.
+template <typename _Tp>
+static _CCCL_API _CCCL_FORCEINLINE __cccl_asinh_sqrt_return_hilo<_Tp>
```
s-oboyle: @miscco The `_CCCL_API _CCCL_FORCEINLINE` combined with the `noexcept` seems to be giving errors:
```
inverse_hyperbolic_functions.h:72:54: error: ‘visibility’ attribute ignored [-Werror=attributes]
72 | __internal_double_Tp_sqrt_unsafe(_Tp __hi, _Tp __lo) noexcept 
   |                                                      ^~~~~~~~
```
is there a best way to fix this or will I just remove the noexcept?

s-oboyle:
/ok to test [ba49bcb](https://github.com/NVIDIA/cccl/pull/6428/commits/ba49bcb56885a164b7cf3827dab46c91a99d0745)

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
```diff
@@ -39,51 +39,360 @@
...
 
 // We have performance issues with some trigonometric functions with extended floating point types
 #if _LIBCUDACXX_HAS_NVBF16()
 template <>
 _CCCL_API inline complex<__nv_bfloat16> asinh(const complex<__nv_bfloat16>& __x)
 {
-  return complex<__nv_bfloat16>{::cuda::std::asinh(complex<float>{__x})};
+  const complex<__nv_bfloat16> ans = complex<__nv_bfloat16>{::cuda::std::asinh(complex<float>{__x})};
+
+  return ans;
```
miscco: What is this needed for?

s-oboyle:
/ok to test [e128ba2](https://github.com/NVIDIA/cccl/pull/6428/commits/e128ba2d0eff049c41946428b9fc170873f316fa)

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
s-oboyle: Good catch, I have no memory of changing that, I must have been experimenting. Reverting.

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
miscco: We have gotten *really* good at only including what we use, so there is a pretty high chance that we need to someplace

also its relatively easy to search for the content of a file because in most cases its the name of the file, like `isinf`

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
```diff
@@ -39,57 +39,364 @@
...
+  if constexpr (is_same_v<_Tp, float>)
+  {
+    NV_IF_TARGET(NV_IS_DEVICE, (return ::rsqrtf(__x);))
+  }
+  if constexpr (is_same_v<_Tp, double>)
+  {
+    NV_IF_TARGET(NV_IS_DEVICE, (return ::rsqrt(__x);))
+  }
+#endif // _CCCL_CUDA_COMPILATION()
+  return _Tp{1} / ::cuda::std::sqrt(__x);
```
miscco, line 59: This could be slightly improved
```suggestion
#if _CCCL_CUDA_COMPILATION()
  if constexpr (is_same_v<_Tp, float>)
  {
    NV_IF_TARGET(NV_IS_DEVICE, (return ::rsqrtf(__x);))
  }
  else if constexpr (is_same_v<_Tp, double>)
  {
    NV_IF_TARGET(NV_IS_DEVICE, (return ::rsqrt(__x);))
  }
  else
#endif // _CCCL_CUDA_COMPILATION()
  {
    return _Tp{1} / ::cuda::std::sqrt(__x);
  }
```

s-oboyle:
/ok to test [fa3bd08](https://github.com/NVIDIA/cccl/pull/6428/commits/fa3bd08b034b91bb9d1f23d18110d53db5200a89)

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
s-oboyle: I think I could remove those files because they were also included by eg `#include <cuda/std/__complex/roots.h>`, but after the other functions get updated `roots.h` will probably be taken out and we'd need all those headers back.

I've added them back in, and will remove `roots.h` when the time comes (which is the most correct way to have things anyway).

miscco (approved):

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
s-oboyle: This won't work, the final return needs to be a catch all for specializations that don't exist otherwise.

With this, if you try to build for eg fp32 on host with nvcc, this is just an empty function with no return value.
(So when ` if constexpr (is_same_v<_Tp, float>)` is true and `NV_IF_TARGET is false`)

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
```diff
@@ -39,43 +39,350 @@
...
-  return complex<_Tp>(::cuda::std::copysign(__z.real(), __x.real()), ::cuda::std::copysign(__z.imag(), __x.imag()));
+
+  // It is a little involved to account for large inputs in an inlined-into-existing-code fashion.
+  // The easiest place to account for large values appears to be here at the start.
+  // Use asinh(x) ~ log(2x) for large x.
+  // Get the largest exponent that passes through the algorithm without issue.
+  // This is ~(max_exponent / 4), with a small bias to make sure edge cases get caught
+  // ~254 for double, ~30 for float
+  constexpr int32_t __max_allowed_exponent     = (__exp_max / 4) - 2;
+  constexpr __uint_t __max_allowed_val_as_uint = __uint_t(__max_allowed_exponent + __exp_max) << __mant_nbits;
```
davebayer: We need to set the explicit bit here. @miscco haven't you added something like `__fp_construct` in the past?

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
```diff
@@ -39,43 +39,350 @@
...
+
+  //  Check if the largest component of __x is > 2^__max_allowed_exponent:
+  _Tp __x_big_factor = _Tp{0};
+  const _Tp __max    = (__realx > __imagx) ? __realx : __imagx;
+  const bool __x_big = ::cuda::std::bit_cast<__uint_t>(__max) > __max_allowed_val_as_uint;
```
davebayer: Why don't you rather bitcast the `__max_allowed_val_as_uint` to `_Tp` and then do the comparison?

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
davebayer: Shouldn't the bias be added here instead of `__exp_max`?

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
```diff
@@ -39,43 +39,350 @@
...
+  _Tp __lo;
+};
+
+// An unsafe sqrt(_Tp + _Tp) extended precision sqrt.
+template <typename _Tp>
+[[nodiscard]] _CCCL_API _CCCL_FORCEINLINE __cccl_asinh_sqrt_return_hilo<_Tp>
+__internal_double_Tp_sqrt_unsafe(_Tp __hi, _Tp __lo) noexcept
+{
+  // rsqrt
+  const _Tp __initial_guess = __internal_rsqrt_inverse_hyperbloic<_Tp>(__hi);
```
fbusato, line 75:
```suggestion
  const _Tp __initial_guess = ::cuda::std::__internal_rsqrt_inverse_hyperbloic<_Tp>(__hi);
```

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
```diff
@@ -39,43 +39,350 @@
...
+  // We need to add the -1.0 here in an fma, or different compilers (eg host vs device)
+  // optimize differently and give (sometimes very) different results.
+  const _Tp _hi_hi_hi = ::cuda::std::fma(__hi, __init_sq_hi, _Tp{-1.0});
+  // Low part not needed for fp32/fp64, but might be if this is extended to other types:
+  // _Tp _hi_hi_lo = fma(__hi, __init_sq_hi, -_hi_hi_hi);
+
+  // Add all terms
+  const _Tp __full_term = _hi_hi_hi + (::cuda::std::fma(__lo, __init_sq_hi, __hi * __init_sq_lo) /*+ _hi_hi_lo*/);
+
+  const _Tp __correction_term = _Tp(-0.5) * __initial_guess * __full_term;
```
fbusato: `_Tp{-0.5}`?

fbusato: other cases as well

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
```diff
@@ -39,43 +39,350 @@
...
+  // Use asinh(x) ~ log(2x) for large x.
+  // Get the largest exponent that passes through the algorithm without issue.
+  // This is ~(max_exponent / 4), with a small bias to make sure edge cases get caught
+  // ~254 for double, ~30 for float
+  constexpr int32_t __max_allowed_exponent     = (__exp_max / 4) - 2;
+  constexpr __uint_t __max_allowed_val_as_uint = __uint_t(__max_allowed_exponent + __exp_max) << __mant_nbits;
+
+  //  Check if the largest component of __x is > 2^__max_allowed_exponent:
+  _Tp __x_big_factor = _Tp{0};
+  const _Tp __max    = (__realx > __imagx) ? __realx : __imagx;
```
fbusato: prefer `fmax` for perfomance reasons

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
```diff
@@ -39,43 +39,350 @@
...
+  const bool __x_big = ::cuda::std::bit_cast<__uint_t>(__max) > __max_allowed_val_as_uint;
+
+  if (__x_big)
+  {
+    // We need __max to be <= ~(2^__max_allowed_exponent),
+    // but not small enough that the asinh(x) ~ log(2x) estimate does
+    // not break down. We are not able to reduce this with a single simple reduction,
+    // so we do a fast/inlined frexp/ldexp:
+    const int32_t __exp_biased = int32_t(::cuda::std::bit_cast<__uint_t>(__max) >> __mant_nbits);
```
fbusato:
```suggestion
    const auto __exp_biased = static_cast<int32_t>(::cuda::std::bit_cast<__uint_t>(__max) >> __mant_nbits);
```

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
```diff
@@ -39,43 +39,350 @@
...
+  const _Tp __x_abs_sq_p1_hi = (__x_abs_sq_hi + _Tp{1});
+  const _Tp __x_abs_sq_p1_lo = __x_abs_sq_lo - ((__x_abs_sq_p1_hi - _Tp{1}) - __x_abs_sq_hi);
+
+  // square:
+  const _Tp __x_abs_sq_p1_sq_hi = __x_abs_sq_p1_hi * __x_abs_sq_p1_hi;
+  _Tp __x_abs_sq_p1_sq_lo       = ::cuda::std::fma(__x_abs_sq_p1_hi, __x_abs_sq_p1_hi, -__x_abs_sq_p1_sq_hi);
+
+  // Add in the lower square terms, all needed
+  __x_abs_sq_p1_sq_lo =
+    ::cuda::std::fma(__x_abs_sq_p1_lo, (_Tp{2} * __x_abs_sq_p1_hi + __x_abs_sq_p1_lo), __x_abs_sq_p1_sq_lo);
```
fbusato, line 240: `fma`?

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
```diff
@@ -39,43 +39,350 @@
...
+
+  // Get __x_abs_sq_p1_sq_hi/lo - 4.0*__imagx_sq_hi/lo:
+  // Subtract high parts:
+  _Tp __inner_most_term_hi = __x_abs_sq_p1_sq_hi - _Tp{4} * __imagx_sq_hi;
```
fbusato, line 244: `fma`? need to swap the operand sign

fbusato: same below

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
```diff
@@ -39,43 +39,350 @@
...
+  _Tp __inner_most_term_lo = ((__x_abs_sq_p1_sq_hi - __inner_most_term_hi) - _Tp{4} * __imagx_sq_hi);
+  // lo parts, all needed:
+  __inner_most_term_lo += __x_abs_sq_p1_sq_lo - _Tp{4} * __imagx_sq_lo;
+
+  // We can have some slightly bad cases here due to catastrohip cancellation that can't be fixed easily.
+  // We still need to to the extended-sqrt on these values, so we fix them now.
+  // It occurs around "real ~= small" and "imag ~= (1 - small)", and imag < 1.
+  // Worked out through targeted testing on fp64 and fp32.
+  _Tp __realx_small_bound = _Tp{1.0e-13};
```
fbusato, line 253: why not a `constexpr`? e.g. `__realx_small_bound = is_same_v<_Tp, float> ? X : Y;

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
```diff
@@ -39,43 +39,350 @@
...
+  // Normalize the above (assumed in the extended-sqrt function):
+  const _Tp __norm_hi  = __inner_most_term_hi + __inner_most_term_lo;
+  const _Tp __norm_lo  = -((__norm_hi - __inner_most_term_hi) - __inner_most_term_lo);
+  __inner_most_term_hi = __norm_hi;
+  __inner_most_term_lo = __norm_lo;
+
+  // Extended sqrt function:
+  // (__extended_sqrt_hi + __extended_sqrt_lo) = sqrt(__inner_most_term_hi + __inner_most_term_lo)
+  const __cccl_asinh_sqrt_return_hilo<_Tp> __extended_sqrt_hilo =
+    __internal_double_Tp_sqrt_unsafe<_Tp>(__inner_most_term_hi, __inner_most_term_lo);
```
fbusato: `::cuda::std`

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
```diff
@@ -39,43 +39,350 @@
...
+
+  _Tp __extended_sqrt_hi = __extended_sqrt_hilo.__hi;
+  _Tp __extended_sqrt_lo = __extended_sqrt_hilo.__lo;
+
+  // 0.0, and some very particular values, do not survive this unsafe sqrt function.
+  // This case occurs when (1 + x^2) is zero or denormal. (and rsqrt(x)*rsqrt(x) become inf).
+  constexpr __uint_t __min_normal_bits = __uint_t(0x1) << __mant_nbits;
```
fbusato:
```suggestion
  constexpr __uint_t __min_normal_bits = __uint_t{1} << __mant_nbits;
```

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
```diff
@@ -39,43 +39,350 @@
...
+  // We instead use the equivalent
+  //     (__realx*__imagx) / sqrt(0.5*(__extended_sqrt_hi - __diffx_m1))
+
+  const _Tp __inside_sqrt_term = _Tp{0.5} * (::cuda::std::fabs(__diffx_m1) + __extended_sqrt_hi);
+
+  // Allow for rsqrt optimization:
+  // We can have two slightly different paths depending on whether rsqrt is available
+  // or not, aka are we on device or host.
+
+  const _Tp __recip_sqrt = __internal_rsqrt_inverse_hyperbloic<_Tp>(__inside_sqrt_term);
```
fbusato:
```suggestion
  const _Tp __recip_sqrt = ::cuda::std::__internal_rsqrt_inverse_hyperbloic<_Tp>(__inside_sqrt_term);
```

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
```diff
@@ -98,7 +405,7 @@ _CCCL_API inline complex<__half> asinh(const complex<__half>& __x)
 // acosh
 
 template <class _Tp>
-[[nodiscard]] _CCCL_API inline complex<_Tp> acosh(const complex<_Tp>& __x)
+_CCCL_API inline complex<_Tp> acosh(const complex<_Tp>& __x)
```
fbusato: why removing `[[nodiscard]]``?

fbusato: also is the function `noexcept`?

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
```diff
@@ -160,7 +467,7 @@ _CCCL_API inline complex<__half> acosh(const complex<__half>& __x)
 // atanh
 
 template <class _Tp>
-[[nodiscard]] _CCCL_API inline complex<_Tp> atanh(const complex<_Tp>& __x)
+_CCCL_API inline complex<_Tp> atanh(const complex<_Tp>& __x)
```
fbusato: same here

fbusato (changes requested):

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
s-oboyle: Good catch, not sure how [[nodiscard]] got removed. noexcept is optional, you can but you don't have to. But I'll be doing `acosh` and `atanh` next anyway

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
s-oboyle: Maybe, but this version was heavily heavily numerically tested and I'm not going to change the numerics now.
This function was written nearly half a year ago (every new maths function seems to take at least a month to be reviewed so it's quite slow) so I've forgotten exactly if this had meaning or not. I guess in this situation it should be numerically the same. (When the accuracy doesn't change I do usually let the compiler decide if it wants to fma or not).

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
s-oboyle: Done, though I really dislike `auto` in math functions like these.

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
s-oboyle: done

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
s-oboyle: done.

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
s-oboyle: done

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
s-oboyle: I couldn't cast to `_Tp` in a `constexpr` way so I was using this, maybe just my issue, but it ended up turning out that integer comparison pipelines everything better in fp heavy code so I left it in.

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
s-oboyle: Good point, we have used `__fp_explicit_bit_mask_of_v<_Tp>` before for explicit bits. Added

s-oboyle: I think the fp formats where `__exp_max` is not equal to `__exp_bias` are basically only the tiny ones right? Like 8/6/4 bit types? I've changed it to `__exp_bias` to be clearer anyway.

s-oboyle:
/ok to test [684c44c](https://github.com/NVIDIA/cccl/pull/6428/commits/684c44c5d959c3cb612b70ae9eac5e58c0ca21f2)

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
s-oboyle: Anticipating a future when other types (eg fp128) will need their own values.

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
```diff
@@ -39,57 +39,367 @@
...
+  // This is ~(max_exponent / 4), with a small bias to make sure edge cases get caught
+  // ~254 for double, ~30 for float
+  constexpr int32_t __max_allowed_exponent = (__exp_max / 4) - 2;
+  constexpr __uint_t __max_allowed_val_as_uint =
+    (__uint_t(__max_allowed_exponent + __exp_bias) << __mant_nbits) | __fp_explicit_bit_mask_of_v<_Tp>;
+
+  //  Check if the largest component of __x is > 2^__max_allowed_exponent:
+  _Tp __x_big_factor = _Tp{0};
+  const _Tp __max    = ::cuda::std::fmax(__realx, __imagx);
+  const bool __x_big = ::cuda::std::bit_cast<__uint_t>(__max) > __max_allowed_val_as_uint;
```
davebayer:
```suggestion
  const bool __x_big = ::cuda::std::__fp_get_storage(__max) > __max_allowed_val_as_uint;
```

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
```diff
@@ -39,57 +39,367 @@
...
+    // but not small enough that the asinh(x) ~ log(2x) estimate does
+    // not break down. We are not able to reduce this with a single simple reduction,
+    // so we do a fast/inlined frexp/ldexp:
+    const auto __exp_biased = static_cast<int32_t>(::cuda::std::bit_cast<__uint_t>(__max) >> __mant_nbits);
+
+    // Get a factor such that (__max * __exp_mul_factor) <= __max_allowed_exponent
+    const __uint_t __exp_reduce_factor =
+      (__uint_t((2 * __exp_max) + __max_allowed_exponent - __exp_biased) << __mant_nbits)
+      | __fp_explicit_bit_mask_of_v<_Tp>;
+    const _Tp __exp_mul_factor = ::cuda::std::bit_cast<_Tp>(__exp_reduce_factor);
```
davebayer:
```suggestion
    const _Tp __exp_mul_factor = ::cuda::std::__fp_from_storage<_Tp>(__exp_reduce_factor);
```

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
```diff
@@ -39,57 +39,367 @@
...
+
+  _Tp __extended_sqrt_hi = __extended_sqrt_hilo.__hi;
+  _Tp __extended_sqrt_lo = __extended_sqrt_hilo.__lo;
+
+  // 0.0, and some very particular values, do not survive this unsafe sqrt function.
+  // This case occurs when (1 + x^2) is zero or denormal. (and rsqrt(x)*rsqrt(x) become inf).
+  constexpr __uint_t __min_normal_bits = __uint_t{0x1} << __mant_nbits;
+  const _Tp __min_normal               = ::cuda::std::bit_cast<_Tp>(__min_normal_bits);
+
+  if (__inner_most_term_hi <= _Tp{2} * __min_normal)
```
davebayer, line 303:
```suggestion
  if (__inner_most_term_hi <= _Tp{2} * ::cuda::std::__fp_min<_Tp>())
```

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
```diff
@@ -39,57 +39,367 @@
...
+  const _Tp __max    = ::cuda::std::fmax(__realx, __imagx);
+  const bool __x_big = ::cuda::std::bit_cast<__uint_t>(__max) > __max_allowed_val_as_uint;
+
+  if (__x_big)
+  {
+    // We need __max to be <= ~(2^__max_allowed_exponent),
+    // but not small enough that the asinh(x) ~ log(2x) estimate does
+    // not break down. We are not able to reduce this with a single simple reduction,
+    // so we do a fast/inlined frexp/ldexp:
+    const auto __exp_biased = static_cast<int32_t>(::cuda::std::bit_cast<__uint_t>(__max) >> __mant_nbits);
```
davebayer:
```suggestion
    const auto __exp_biased = ::cuda::std::__fp_get_exp_biased(__max);
```

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
```diff
@@ -39,57 +39,367 @@
...
+
+  // It is a little involved to account for large inputs in an inlined-into-existing-code fashion.
+  // The easiest place to account for large values appears to be here at the start.
+  // Use asinh(x) ~ log(2x) for large x.
+  // Get the largest exponent that passes through the algorithm without issue.
+  // This is ~(max_exponent / 4), with a small bias to make sure edge cases get caught
+  // ~254 for double, ~30 for float
+  constexpr int32_t __max_allowed_exponent = (__exp_max / 4) - 2;
+  constexpr __uint_t __max_allowed_val_as_uint =
+    (__uint_t(__max_allowed_exponent + __exp_bias) << __mant_nbits) | __fp_explicit_bit_mask_of_v<_Tp>;
```
davebayer, line 185:
```suggestion
  constexpr __uint_t __max_allowed_val_as_uint = ::cuda::std::__fp_set_exp(__uint_t{}, __max_allowed_exponent);
```

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
```diff
@@ -39,57 +39,367 @@
...
+    // We need __max to be <= ~(2^__max_allowed_exponent),
+    // but not small enough that the asinh(x) ~ log(2x) estimate does
+    // not break down. We are not able to reduce this with a single simple reduction,
+    // so we do a fast/inlined frexp/ldexp:
+    const auto __exp_biased = static_cast<int32_t>(::cuda::std::bit_cast<__uint_t>(__max) >> __mant_nbits);
+
+    // Get a factor such that (__max * __exp_mul_factor) <= __max_allowed_exponent
+    const __uint_t __exp_reduce_factor =
+      (__uint_t((2 * __exp_max) + __max_allowed_exponent - __exp_biased) << __mant_nbits)
+      | __fp_explicit_bit_mask_of_v<_Tp>;
```
davebayer, line 203:
```suggestion
    const auto __exp_reduce_factor = ::cuda::std::__fp_set_exp_biased(__uint_t{}, (2 * __exp_max) + __max_allowed_exponent - __exp_biased);
```

davebayer (changes requested):
Please, try to use more of the `__fp_meow` machinery. In #6749, I am fixing the behaviour of `__fp_set_exp`, so it automatically sets the explicit bit

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
```diff
@@ -39,57 +39,367 @@
...
 _CCCL_BEGIN_NAMESPACE_CUDA_STD
 
+// Specialization of 1/sqrt(x), uses device-only rsqrtf/rsqrt when viable.
+// Sometimes compilers can optimize 1/sqrt(x) so we don't suffer the slowdown of the divide.
+// It can happen that the double 1/sqrt() on device gets optimized better than CUDA rsqrt(),
+// but it depends on the calling function. We needs to optimize on an individual basis.
+template <class _Tp>
+[[nodiscard]] _CCCL_API _CCCL_FORCEINLINE _Tp __internal_rsqrt_inverse_hyperbloic(_Tp __x) noexcept
+{
+#if _CCCL_CUDA_COMPILATION()
```
fbusato, line 49: is `_CCCL_CUDA_COMPILATION` strictly needed here?

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
fbusato: no, it is not :)

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
```diff
@@ -39,57 +39,367 @@
...
+
+  // All terms needed, allow the compiler to pick which way to
+  // optimize this to fma, same accuracy.
+  __ans_hi_lo += __initial_guess * __lo + __correction_term * __hi;
+
+  __cccl_asinh_sqrt_return_hilo<_Tp> __retval_hilo;
+  __retval_hilo.__hi = __ans_hi_hi;
+  __retval_hilo.__lo = __ans_hi_lo;
+
+  return __retval_hilo;
```
fbusato:
```suggestion
  return __cccl_asinh_sqrt_return_hilo<_Tp>{__ans_hi_hi, __ans_hi_lo};
```

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
```diff
@@ -39,57 +39,367 @@
 
 _CCCL_BEGIN_NAMESPACE_CUDA_STD
 
+// Specialization of 1/sqrt(x), uses device-only rsqrtf/rsqrt when viable.
+// Sometimes compilers can optimize 1/sqrt(x) so we don't suffer the slowdown of the divide.
+// It can happen that the double 1/sqrt() on device gets optimized better than CUDA rsqrt(),
+// but it depends on the calling function. We needs to optimize on an individual basis.
+template <class _Tp>
+[[nodiscard]] _CCCL_API _CCCL_FORCEINLINE _Tp __internal_rsqrt_inverse_hyperbloic(_Tp __x) noexcept
```
fbusato, line 47: replace it with `inline` for consistency

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
```diff
@@ -39,57 +39,367 @@
...
+  constexpr int32_t __exp_bias   = __fp_exp_bias_v<__fp_format_of_v<_Tp>>;
+
+  constexpr _Tp __pi  = __numbers<_Tp>::__pi();
+  constexpr _Tp __ln2 = __numbers<_Tp>::__ln2();
+
+  _Tp __realx = ::cuda::std::fabs(__x.real());
+  _Tp __imagx = ::cuda::std::fabs(__x.imag());
+
+  // Special cases that do not pass through:
+  if (!::cuda::std::isfinite(__realx) || !::cuda::std::isfinite(__imagx))
```
fbusato, line 128: this section repeats `isfinite`, `isinf` `isnan` multiple times. Would not be better to create variables for them. The compiler should be able to optimize them anyway...

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
```diff
@@ -39,57 +39,367 @@
...
+    const __uint_t __exp_reduce_factor =
+      (__uint_t((2 * __exp_max) + __max_allowed_exponent - __exp_biased) << __mant_nbits)
+      | __fp_explicit_bit_mask_of_v<_Tp>;
+    const _Tp __exp_mul_factor = ::cuda::std::bit_cast<_Tp>(__exp_reduce_factor);
+
+    // Scale down to a working range.
+    __realx *= __exp_mul_factor;
+    __imagx *= __exp_mul_factor;
+
+    __x_big_factor = _Tp((__exp_biased - __exp_max) - __max_allowed_exponent) * __ln2;
```
fbusato: `fma` could be also used here. Let skip it if it could introduce accuracy issues.

fbusato: also, `_Tp()` to `static_cast`

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
```diff
@@ -39,57 +39,367 @@
...
+  const _Tp __ans_imag = ::cuda::std::atan2(__atan2_input1, __atan2_input2);
+
+  // The small |real| case, as mentioned above.
+  // Bounds found by testing.
+  _Tp __realx_small_bound_override = _Tp{2.220446e-16};
+
+  if constexpr (is_same_v<_Tp, float>)
+  {
+    __realx_small_bound_override = _Tp{6.0e-08f};
+  }
```
fbusato, line 374:
```suggestion
  _Tp __realx_small_bound_override = is_same_v<_Tp, float> ? _Tp(6.0e-08f) : _Tp(2.220446e-16);
```

miscco: This is not equivalent, as you flipped the condition. 

Also that transformation might introduce compiler warnings about non-representable floating point values, as the double branch is not elided away

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
miscco: It uses cuda runtime functions on device so I would assume yes

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
fbusato: fixed. Should work for the conversion

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
s-oboyle: We want forceinline here and just inline for the functions, had previously talked about it with @miscco.

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
s-oboyle: Was going to, but this is how it is written in all other functions so left it for some consistency. Don't think it should matter too much alright.

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
s-oboyle: Could `__fp_min `be renamed to `__fp_min_normal`, if possible? To me `__fp_min `could mean the min normal or denormal value (and could even potentially depend on the compilation flags, like `--ftz`).

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
s-oboyle: Could you create a version of `__fp_get_exp_biased` for values that are known to be positive? Like `__fp_get_exp_biased_pos` or whatever you want.
The small little extra operations and mask value needed tend to have a measurable perf effect on the math functions when they are called in a loop unrolled fashion, if the compiler can't work the stuff it can optimize out.

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
s-oboyle: Same with the extra mask and operation used in this function.

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
s-oboyle: Wouldn't be needed here, but a version of `__fp_set_exp_biased` for fp-types that need an explicit bit, but that doesn't set it would be good. Just to remove the unneeded ops.

s-oboyle:
/ok to test [eacbe9e](https://github.com/NVIDIA/cccl/pull/6428/commits/eacbe9e3c58dea18c634b5ccb75da36d8f38e2d1)

s-oboyle:
/ok to test [39dbc1c](https://github.com/NVIDIA/cccl/pull/6428/commits/39dbc1c30c99d721218784dc2cd6d627f0955f6d)

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
miscco: we should keep it as is for now and investigate improving it in a consistent maner

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
davebayer: I copied the `numeric_limits` style, `__fp_min` is normal min and `__fp_denorm_min` is the denorm min. Do you find that confusing?

davebayer (approved):

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
s-oboyle: Ah I did not know the `numeric_limits` had it defined this way, it's not so much confusing as it is vague to me.
I'm little surprised that it is defined this way, if I was using it to write fp-code I would be looking up if it was the min-normal or min-denormal every second day. Thanks for enlightening me!

PR #6431, opened 2025-10-31
Add `WarpReduce` Device-Side Benchmarks

fbusato:
## Description

Historically, we have benchmarked all CUB functionalities by evaluating the performance of host-side calls. While this remains the appropriate method for benchmarking host-side APIs, it is not the most effective method for evaluating device-side functionalities.

The main reason is that optimizations for device-side functionalities do not always improve the overall performance of the host-side API. Even a small modification to the device-side code can result in a different reordering of SASS instructions, code layout, and cache hits. This can lead to lower overall performance, even if the individual functionality has been optimized and results in the expected SASS code.

----

In this PR, we evaluate the performance of `WarpReduce` by isolating and benchmarking the device-side code directly. The target here is throughput, rather than latency.

The following aspects are considered to ensure reliable and reproducible results:

- Maximize GPU utilization.
- Avoid grid quantization.
- Minimize benchmark noise of initialization and epilogue. 
- Isolate warp workloads to different warp chunks.
- Create false dependencies between loop iterations.
- Forced unrolling (`cuda::static_for`) instead of relying on weaker `pragma unroll`.
- The final SASS code has been validated. 

Finally, the code has been generalized enough to be easily extended to other functionalities in the future.

### DIFF
--- a/cub/benchmarks/bench/reduce/warp_reduce_min.cu
+++ b/cub/benchmarks/bench/reduce/warp_reduce_min.cu
@@ -11,10 +11,10 @@ using value_types =
 #if NVBENCH_HELPER_HAS_I128
                      int128_t,
 #endif
-#if _CCCL_HAS_NVFP16()
+#if _CCCL_HAS_NVFP16() && _CCCL_CTK_AT_LEAST(12, 2)
                      __half,
 #endif
-#if _CCCL_HAS_NVBF16()
+#if _CCCL_HAS_NVBF16() && _CCCL_CTK_AT_LEAST(12, 2)
                      __nv_bfloat16,
 #endif
                      float,
--- a/cub/benchmarks/bench/reduce/warp_reduce_sum.cu
+++ b/cub/benchmarks/bench/reduce/warp_reduce_sum.cu
@@ -10,18 +10,18 @@ using value_types = nvbench::type_list<
 #if NVBENCH_HELPER_HAS_I128
   int128_t,
 #endif
-#if _CCCL_HAS_NVFP16()
+#if _CCCL_HAS_NVFP16() && _CCCL_CTK_AT_LEAST(12, 2)
   __half,
 #endif
-#if _CCCL_HAS_NVBF16()
+#if _CCCL_HAS_NVBF16() && _CCCL_CTK_AT_LEAST(12, 2)
   __nv_bfloat16,
 #endif
   float,
   double,
-#if _CCCL_HAS_NVFP16()
+#if _CCCL_HAS_NVFP16() && _CCCL_CTK_AT_LEAST(12, 2)
   cuda::std::complex<__half>,
 #endif
-#if _CCCL_HAS_NVBF16()
+#if _CCCL_HAS_NVBF16() && _CCCL_CTK_AT_LEAST(12, 2)
   cuda::std::complex<__nv_bfloat16>,
 #endif
   cuda::std::complex<float>,


nvbench_helper/nvbench_helper/device_side_benchmark.cuh
```diff
@@ -0,0 +1,43 @@
...
+
+__device__ static int device_var[16];
+
+template <typename T>
+__device__ __forceinline__ static void sink(T value)
+{
+  if (cuda::ptx::get_sreg_smid() == static_cast<uint32_t>(-1))
+  {
+    *reinterpret_cast<T*>(device_var) = value;
+  }
```
bernhardmgruber, line 32: Could we just do a fake read here, with something like:
```c++
  asm("" : "+r"(value));
  ```
  ? I have seen this in benchmarks from @ahendriksen

File: nvbench_helper/nvbench_helper/device_side_benchmark.cuh [hunk 2]
```diff
@@ -0,0 +1,43 @@
+// SPDX-FileCopyrightText: Copyright (c) 2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: BSD-3-Clause
```
bernhardmgruber: Important: any new file should be:
```suggestion
// SPDX-FileCopyrightText: Copyright (c) 2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
```
applies to all new files.

cub/benchmarks/bench/reduce/warp_reduce_base.cuh
```diff
@@ -0,0 +1,41 @@
...
+void warp_reduce(nvbench::state& state, nvbench::type_list<T>)
+{
+  constexpr int block_size    = 256;
+  constexpr int unroll_factor = 128; // compromise between compile time and noise
+  const auto& kernel          = benchmark_kernel<block_size, unroll_factor, benchmark_op_t, T>;
+  int num_SMs                 = state.get_device().value().get_number_of_sms();
+  int device                  = state.get_device().value().get_id();
+  int max_blocks_per_SM       = 0;
+  NVBENCH_CUDA_CALL_NOEXCEPT(cudaOccupancyMaxActiveBlocksPerMultiprocessor(&max_blocks_per_SM, kernel, block_size, 0));
+  int grid_size = max_blocks_per_SM * num_SMs;
```
bernhardmgruber:
```suggestion
  const int num_SMs                 = state.get_device().value().get_number_of_sms();
  const int device                  = state.get_device().value().get_id();
  const int max_blocks_per_SM       = 0;
  NVBENCH_CUDA_CALL_NOEXCEPT(cudaOccupancyMaxActiveBlocksPerMultiprocessor(&max_blocks_per_SM, kernel, block_size, 0));
  const int grid_size = max_blocks_per_SM * num_SMs;
```

nvbench_helper/nvbench_helper/device_side_benchmark.cuh
ahendriksen: > `asm("" : "+r"(value));`

That will ensure NVVM believes that the value is used, but from PTX -> SASS, this is still likely to be optimized out.

bernhardmgruber: Understood, thx!

cub/benchmarks/bench/reduce/warp_reduce_min.cu
bernhardmgruber, line 1: Question: Should we add a comment to this file, that this benchmark is not intended for tuning? Or should this be inherently clear, since we cannot tune warp algorithms?

fbusato: I don't think it is needed. If the issue will come up later, we can add a note

nvbench_helper/nvbench_helper/nvbench_helper.cuh
```diff
@@ -33,7 +33,17 @@ NVBENCH_DECLARE_TYPE_STRINGS(uint128_t, "U128", "uint128_t");
 
 using complex = cuda::std::complex<float>;
 
-NVBENCH_DECLARE_TYPE_STRINGS(complex, "C64", "complex");
+#if _CCCL_HAS_NVFP16()
+NVBENCH_DECLARE_TYPE_STRINGS(__half, "Half", "half");
```
alliepiper, line 37: Nit: I'm late to the party on this, but I'd recommend `F16` and `BF16` to better match the other short identifiers and keep the table widths under control. Food for thought in a future PR 🙂

PR #6432, opened 2025-10-31
[cudax] Add synchronous_resource_adapter and use it in async_buffer

pciolkosz:
We have the legacy non-stream ordered memory resources, but we don't have a way to use them in a buffer type.
While we could add a `synchronous_buffer` type along with `async_buffer` we currently have, this approach has a number of issues:
- In a lot of cases user would like to initialize the `synchronous_buffer` and most likely that would be a stream ordered initialization. It makes no difference if the allocation was stream ordered or not, after the initialization is submitted into a stream the buffer has a dependency on that stream and it would be nice if we could express that in the buffer type, like we do in `async_buffer`
- Conversion from `async_buffer` to `synchronous_buffer` (and back?) on synchronization or launching new stream ordered work is not the most convenient. Instead https://github.com/NVIDIA/cccl/pull/6416 makes the stream in `async_buffer` optional, so you can add and remove stream dependency from `async_buffer` easily.

The direction we would like to take instead is to rename `async_buffer` to just `buffer` and make it work with legacy resources. This PR adds `synchronous_resource_adapter`, which is a generic utility wrapper that can extend `synchronous_resource` to a full `resource`. It forwards `allocate_sync` and `deallocate_sync` to the wrapper resource, while for `allocate` and `deallocate`, it forwards it if possible, otherwise uses the `_sync` variants. In order to make `deallocate` correct it also synchronizes the stream in case forwarding was not possible. The adapter is generic and will work for any custom `synchronous_resource`.

This PR also makes `async_buffer` automatically apply the added adapter to a resource used to create a buffer if it does not have the stream ordered allocation/deallocation. This way legacy resources can be used with that type just like the memory pools.

After this and https://github.com/NVIDIA/cccl/pull/6416 is merged a future change will rename `async_buffer` to just `buffer`

### DIFF
--- a/.github/actions/workflow-run-job-windows/install_gpu_driver.ps1
+++ b/.github/actions/workflow-run-job-windows/install_gpu_driver.ps1
@@ -13,7 +13,7 @@ function Install-Driver {
 
     # extract gpu type from gha runner label:
     # Labels are in the form: <os>-<cpu>-gpu-<gpu>-<driver>-<n>
-    $gha_runner_label=$env:RUNNER
+    $gha_runner_label=$env:JOB_RUNNER
     $gpu_type = $gha_runner_label.Split('-')[3]
 
     $data_center_gpus = @('a100', 'h100', 'l4', 't4', 'v100', 'rtxa6000', 'rtx6000ada')
--- a/cub/benchmarks/bench/reduce/deterministic.cu
+++ b/cub/benchmarks/bench/reduce/deterministic.cu
@@ -51,7 +51,7 @@ void deterministic_sum(nvbench::state& state, nvbench::type_list<T>)
   using accum_t     = T;
   using transform_t = ::cuda::std::identity;
 
-  using dispatch_t = cub::detail::DispatchReduceDeterministic<
+  using dispatch_t = cub::detail::rfa::dispatch_t<
     input_it_t,
     output_it_t,
     int,
--- a/cub/cub/device/device_reduce.cuh
+++ b/cub/cub/device/device_reduce.cuh
@@ -197,7 +197,7 @@ private:
       __accumulator_t<ReductionOpT, ::cuda::std::invoke_result_t<TransformOpT, detail::it_value_t<InputIteratorT>>, T>;
     using policy_t = typename reduce_tuning_t::template fn<accum_t, offset_t, ReductionOpT>;
     using dispatch_t =
-      detail::DispatchReduceDeterministic<InputIteratorT, OutputIteratorT, offset_t, T, TransformOpT, accum_t, policy_t>;
+      detail::rfa::dispatch_t<InputIteratorT, OutputIteratorT, offset_t, T, TransformOpT, accum_t, policy_t>;
 
     return dispatch_t::Dispatch(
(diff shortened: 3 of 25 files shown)


cudax/include/cuda/experimental/__container/async_buffer.cuh
```diff
@@ -194,11 +196,12 @@ public:
   //! @brief Constructs an empty async_buffer using an environment
   //! @param __env The environment providing the needed information
   //! @note No memory is allocated.
-  _CCCL_TEMPLATE(class _Env = ::cuda::std::execution::env<>)
-  _CCCL_REQUIRES(__buffer_compatible_env<_Env>)
+  _CCCL_TEMPLATE(class _Resource, class _Env = ::cuda::std::execution::env<>)
+  _CCCL_REQUIRES(
+    ::cuda::mr::synchronous_resource<::cuda::std::decay_t<_Resource>> _CCCL_AND __buffer_compatible_env<_Env>)
```
davebayer: Isn't this condition wrong? It will always require the resource to be synchronous

cudax/include/cuda/experimental/__memory_resource/synchronous_resource_adapter.cuh
```diff
@@ -0,0 +1,129 @@
...
+  _CCCL_REQUIRES_EXPR((_Resource), _Resource& __res, ::cuda::stream_ref __stream, size_t __bytes, size_t __alignment)(
+    _Same_as(void*) __res.allocate(__stream, __bytes, __alignment));
+
+template <class _Resource>
+_CCCL_CONCEPT __has_member_deallocate = _CCCL_REQUIRES_EXPR(
+  (_Resource), _Resource& __res, ::cuda::stream_ref __stream, void* __ptr, size_t __bytes, size_t __alignment)(
+  _Same_as(void) __res.deallocate(__stream, __ptr, __bytes, __alignment));
+
+template <class _Resource>
+struct synchronous_resource_adapter : ::cuda::mr::__copy_default_queries<_Resource>
```
davebayer, line 51: Q: should the adapter own the resource?

cudax/include/cuda/experimental/__memory_resource/synchronous_resource_adapter.cuh
```diff
@@ -0,0 +1,129 @@
...
+{
+  synchronous_resource_adapter(const _Resource& __resource) noexcept
+      : __resource(__resource)
+  {}
+
+  synchronous_resource_adapter(_Resource&& __resource) noexcept
```
davebayer:
```suggestion
  _CCCL_HOST_API synchronous_resource_adapter(const _Resource& __resource) noexcept
      : __resource(__resource)
  {}

  _CCCL_HOST_API synchronous_resource_adapter(_Resource&& __resource) noexcept
```

davebayer (approved):

cudax/include/cuda/experimental/__memory_resource/synchronous_resource_adapter.cuh
pciolkosz: I think it should, otherwise you run into lifetime issues

davebayer:
pre-commit.ci autofix

PR #414, opened 2023-09-07
64-bit indexing for DeviceSegmentedReduce

jecs:
## Description

 

* Added NumSegmentsT parameter to DeviceSegmentedReduce methods, so now num_segments can be of any integral type
* Added OffsetT type resolution in a manner similar to DeviceReduce: set to std::common_type_t of BeginOffsetIterT::value_type and EndOffsetIterT::value_type. (see detail/choose_offset.cuh)
* In a manner similar to DeviceReduce, we use static_cast to convert num_segments to OffsetT before dispatching
* OffsetT type resolution uses std::common_type_t, std::void_t, and one inline constexpr variable (is_iterator_v).

## Checklist

- [x] New or existing tests cover these changes.
- [ ] The documentation is up to date with these changes.

PR #6434, opened 2025-11-01
Add `cccl_add_xfail_compile_target_test` CMake function

alliepiper:
- CUB's usages of XFAIL compilation are updated to use the new utility.
- test_device_segment_reduce_offset_type_fail test is fixed; the regexes weren't detected and some cases were failing for unintended reasons.
- test_param_general_fail is added to test edge case.

Supports CUB-style in-source error regex extraction as defined the docs:

```cmake
# cccl_add_xfail_compile_target_test(
#   <target_name>
#   [TEST_NAME <test_name>]
#   [ERROR_REGEX <regex>]
#   [SOURCE_FILE <source_file>]
#   [ERROR_REGEX_LABEL <error_string>]
#   [ERROR_NUMBER <error_number>]
#   [ERROR_NUMBER_TARGET_NAME_REGEX <regex>]
# )
#
# Given a configured build target that is expected to fail to compile:
# - Mark the target as excluded from the `all` target.
# - Create a CTest test that compiles the target. If TEST_NAME is provided, it is used.
#   Otherwise, the target_name is used as the test name.
# - When the test runs, it passes if exactly one of the following conditions is met:
#   - A provided / detected error regex matches the compilation output, ignoring exit code.
#   - No error regex is provided / detected, and the compilation fails.
#
# An error regex may be explicitly provided via ERROR_REGEX, or it may be
# detected by scanning the SOURCE_FILE for a specially formatted comment.
#
# If ERROR_REGEX_LABEL is provided, the SOURCE_FILE will read, looking for a comment of the form:
#
# // <ERROR_REGEX_LABEL> {{"error_regex"}}
#
# An error number may be appended to the ERROR_REGEX_LABEL in the comment:
#
# // <ERROR_REGEX_LABEL>-<error_number> {{"error_regex"}}
#
# If ERROR_NUMBER_TARGET_NAME_REGEX is specified, the regex is used to capture
# the error_number from the target name. If target_name is
# "cccl.test.my_test.err_5.foo_3" and ERROR_NUMBER_TARGET_NAME_REGEX is
# "\\.err_([0-9]+)", the captured error number "5."
#
# // <ERROR_REGEX_LABEL>-<captured_error_number> {{"error_regex"}}
#
# If ERROR_NUMBER is provided, ERROR_NUMBER_TARGET_NAME_REGEX is ignored.
# If ERROR_NUMBER_TARGET_NAME_REGEX is provided but does not match, a plain ERROR_REGEX_LABEL is used.
```


cub/cub/device/device_segmented_reduce.cuh
```diff
@@ -830,10 +830,13 @@ struct DeviceSegmentedReduce
 
     using InputValueT  = detail::it_value_t<InputIteratorT>;
     using OutputTupleT = detail::non_void_value_t<OutputIteratorT, KeyValuePair<OffsetT, InputValueT>>;
+    using OutputKeyT   = typename OutputTupleT::Key;
     using OutputValueT = typename OutputTupleT::Value;
     using AccumT       = OutputTupleT;
     using InitT        = detail::reduce::empty_problem_init_t<AccumT>;
 
+    static_assert(::cuda::std::is_same_v<int, OutputKeyT>, "Output key type must be int.");
```
bernhardmgruber, line 814: @elstehle does the output iterator's value type's key really have to be `int` or can it be another integral type (a larger offset type)? The documentation for `OutputIteratorT` says `int`.

alliepiper: FWIW, I originally tried `is_convertible` to `int`, but the compilation fails farther in if this isn't exactly `int`. The new static assert just gives a nicer error message that we can consistently check with the fail test.

alliepiper: @bernhardmgruber Addressing this issue is out of scope. The static assert here just gets the `xfail` test passing and only adds a clearer message for an existing failure mode.

Can we approve this and clean up the API in a later PR? This PR is blocking #4996.

bernhardmgruber: But the status quo of this PR introduces a breaking change if the user was allowed to pass an output key type that is not `int`. I need to know whether this was allowed before.

miscco: Shouldnt it rather be an `is_signed_v`?

elstehle: `DeviceSegmentedReduce::Arg{Min,MAx}` is the only algorithm missing support for large number of items because of [this issue](https://github.com/NVIDIA/cccl/pull/414#discussion_r1330632615). For this reason, `int` as the `OutputIteratorT` type is currently correct. 

I would like to deprecate and drop returning `cub::KeyValuePair` and replace it with `cuda::std::pair`, but I'm afraid that'll have to wait for the next major release.

bernhardmgruber (approved):

PR #6435, opened 2025-11-01
Use `__is_floating_point` builtin

davebayer:


davebayer:
Clang considers return `true` for `__float128`, which is not something we want..

Issue #6430, opened 2025-10-31
[BUG]: False-positive memcheck failure when using `atomic_ref` on small data types (e.g., int8_t)


PointKernel:
### Is this a duplicate?

- [x] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cccl/issues) for this bug and that I agree to the [Code of Conduct](CODE_OF_CONDUCT.md)

### Type of Bug

Something else

### Component

libcu++

### Describe the bug

When using `atomic_ref::min` (or other atomic operations) on small data types such as `int8_t`, `compute-sanitizer` (memcheck) reports an **invalid global read of size 4 bytes**.

This occurs when the underlying memory allocation is tightly packed (i.e., no 4-byte padding). The issue stems from CCCL’s use of a 4-byte CAS loop to emulate atomic operations on types smaller than 4 bytes.

While the sanitizer reports an out-of-bounds load, the operation is known to be **safe on all supported hardware**—making this a **false positive**.

libcudf currently works around this by rounding small-type column buffer sizes to a multiple of 4 bytes to silence the failures in our nightly CI.

### How to Reproduce

Compile and run the following kernel under `compute-sanitizer` (memcheck) with input data size of 5:

```cpp
__global__ void test(int8_t* data) {
  cuda::atomic_ref<int8_t, cuda::thread_scope_system> ref(*data);
  ref.fetch_min(1);
}
```

**Observed output:**

```
========= Invalid __global__ read of size 4 bytes
=========     at void cuda::std::__cuda_atomic_load<unsigned int>(...)
```

This only happens when `data` points to a buffer without 4-byte padding.

### Expected behavior

* No invalid global memory access should be reported by `compute-sanitizer`.
* The sanitizer should recognize that the 4B CAS operation does not modify bytes outside the target value.

### Additonal notes 
* The operation is safe and behaves correctly on all supported GPU architectures.
* CCCL maintainers confirmed this is not a bug in `atomic_ref`, but a sanitizer false positive.
* Potential solution: NVVM intrinsic or PTX pragma to annotate these safe atomic operations so memcheck can suppress the warning.

PR #6442, opened 2025-11-02
Fix atomic_ref narrow-type memcheck false positive

Aminsed:
## Summary
- introduce `__atomic_ref_small_tag` and `reference_small.h` so `cuda::atomic_ref` on sub-4-byte types performs byte-granular locking instead of widened RMWs  
- update atomic dispatch plumbing to route narrow references through the new implementation without disturbing existing owners  
- add a focused `atomic_ref<int8_t>` regression test covering host and device access patterns (sanitizer-friendly)

## Motivation
[`#6430`](https://github.com/NVIDIA/cccl/issues/6430) reports that `cuda::atomic_ref` on narrow types trips compute-sanitizer memcheck because the implementation promotes byte writes to 32-bit RMWs. The sanitizer flags that widened access as an invalid global read, which blocks users (e.g., libcudf) from running their pipelines under memcheck.

## Explanation
The patch introduces a new storage tag for narrow `atomic_ref` instances and backs it with a small device-side lock table. Each operation acquires a byte-granular lock, performs the necessary load/store/update, and releases the lock. Host execution still relies on the existing libatomic wrappers. This approach keeps behavior identical while preventing memcheck from seeing out-of-bounds accesses.

## Rationale
- **Minimally invasive**: the change only touches the `atomic_ref` storage layer; existing owning atomics and the dispatch macros stay untouched.  
- **Deterministic under sanitizer**: the lock path guarantees single-byte reads/writes, so memcheck sees the exact footprint users expect.  
- **Regression safeguards**: the new test exercises `atomic_ref<int8_t>` on both host and device, proving the narrow path works end-to-end (and keeps sanitizer latency low).

## Testing
- `nvcc -std=c++20 -x cu libcudacxx/test/libcudacxx/std/atomics/atomics.types.generic/integral/8b_integral_ref.pass.cpp -Ilibcudacxx/include -Ilibcudacxx/test/support -o /tmp/atomic_ref_small_test`
- `compute-sanitizer --tool memcheck /tmp/atomic_ref_small_test`

@miscco @griwes

### DIFF
--- a/libcudacxx/include/cuda/std/__atomic/types.h
+++ b/libcudacxx/include/cuda/std/__atomic/types.h
@@ -24,7 +24,9 @@
 #include <cuda/std/__atomic/types/base.h>
 #include <cuda/std/__atomic/types/locked.h>
 #include <cuda/std/__atomic/types/reference.h>
-#include <cuda/std/__atomic/types/reference_small.h>
+#if defined(_LIBCUDACXX_ATOMIC_REF_ENABLE_MEMCHECK_SAFE)
+#  include <cuda/std/__atomic/types/reference_small.h>
+#endif // defined(_LIBCUDACXX_ATOMIC_REF_ENABLE_MEMCHECK_SAFE)
 #include <cuda/std/__atomic/types/small.h>
 #include <cuda/std/__type_traits/conditional.h>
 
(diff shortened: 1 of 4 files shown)


miscco:
Thanks @Aminsed for looking into this.

We need to discuss this internally. atomics are highly performance sensitive and already a compile time sink. Given that this is something we as the implementation know to work properly, we might want to eat / disable the warning instead of adding complexity.

@wmaxey please give this a look and investigate any performance / code size issues

Aminsed:
@miscco Thanks for raising the concern. I agree we shouldn't make the byte-wide path the default. I've reworked the series so the lock-backed implementation only comes into play when `_LIBCUDACXX_ATOMIC_REF_ENABLE_MEMCHECK_SAFE` is defined. Without that opt-in the codegen stays identical to the existing 4-byte CAS path, so there shouldn't be any runtime or compile-time impact on normal builds. Defining the macro lets us flip the byte-granular path on for targeted memcheck runs; I re-ran the regression test + memcheck with it enabled to make sure the warning still disappears. Let me know if you'd prefer a different toggle shape or additional measurements.

Aminsed:
After benchmarking the opt-in macro and some thoughts I agree with @miscco that is hauling in extra code for minimal gain. I’m going to close this out, added complexity isn’t justified here.

Issue #5668, opened 2025-08-26
[BUG]: GPU Scan uses wrong aggregation type


pauleonix:
### Is this a duplicate?

- [x] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cccl/issues) for this bug and that I agree to the [Code of Conduct](CODE_OF_CONDUCT.md)

### Type of Bug

Silent Failure

### Component

Thrust

### Describe the bug

I post this as a Thrust bug as I found it there, but I'm pretty sure that the underlying bug (assuming there is only one) is in `cub::BlockScan`. In the multiple-elements-per-thread case `cub::BlockScan` will use `cub::ThreadReduce` but not specify its `AccumT`. By default this type is chosen as whatever is returned by the binary operator applied to the values of the input type. This is inconsistent with `BlockScan` using the input/output/value type `T` for accumulation everywhere else.

Example member of `BlockScan` (not the one used by Thrust in the reproducer, but they all have this problem):
https://github.com/NVIDIA/cccl/blob/d52ffc7393a35abf15d06c6fd6a96c681a0d186d/cub/cub/block/block_scan.cuh#L1062-L1074

Signature of `ThreadReduce`:
https://github.com/NVIDIA/cccl/blob/d52ffc7393a35abf15d06c6fd6a96c681a0d186d/cub/cub/thread/thread_reduce.cuh#L169-L173

I was honestly not able to find a reproducer that does not rely on an artificially overloaded custom functor. But as it is supposedly very hard to prove that the current implementation is safe under any realistic circumstances this still needs to be fixed.

### How to Reproduce

0. See godbolt link below for exact reproducer
1. Define a custom binary operator with differing input and output types that has an overload with special behavior in the case that the first argument is of the output type:
   ```c++
   struct custom_op {
       __host__ __device__ op_t operator()(value_t lhs, value_t rhs) {
           return cuda::std::plus<op_t>{}(lhs, rhs);
       }
       // This one should never be called!
       __host__ __device__ op_t operator()(op_t lhs, value_t rhs) {
           return op_t{999};
       }
   };
   ```
2. Choose a big enough input range (16 is the smallest value that shows the bug for `sm_75`).
3. Compare the output to a host scan (either Thrust or STL).

### Expected behavior

I expect the second overload of the above functor not be be called and therefore the result on host and device to be the same.

### Reproduction link

https://godbolt.org/z/TGr4M6W6h

PR #6443, opened 2025-11-02
Fix BlockScan accumulator type handling

Aminsed:
## Summary
- keep `ThreadReduce` accumulator types pinned to the block value type across `BlockScan` and `BlockReduce`
- apply the same accumulator fix to the raking specialization so all paths use the intended type
- add a regression test that exercises `BlockScan` with a functor returning a wider type

## Motivation
[`#5668`](https://github.com/NVIDIA/cccl/issues/5668) shows that `BlockScan` widens the accumulator when the scan functor returns a wider type than the block value. That implicit widening breaks user code that relies on the original type and can even hit deleted overloads.

## Explanation
`ThreadReduce` was deducing its accumulator type from the functor instead of the block value `T`. The patch explicitly instantiates `ThreadReduce` with `AccumT = T` everywhere `BlockScan` and `BlockReduce` dispatch through it, including the raking specialization. The new unit test exercises an operator that returns `long long` for `int` inputs and verifies the accumulator remains `int`.

## Rationale
- **Minimal surface area**: the change touches only the `ThreadReduce` call sites; public APIs and template parameters stay the same.  
- **Consistent behavior**: every BlockScan reduction path now uses the same accumulator type, avoiding divergent code paths.  
- **Regression coverage**: the new Catch2 test guards against future regressions triggered by wider returning ops.

## Testing
- `pre-commit run --files cub/cub/block/block_scan.cuh cub/cub/block/block_reduce.cuh cub/cub/block/specializations/block_reduce_raking_commutative_only.cuh cub/test/catch2_test_block_scan.cu`

### DIFF
--- a/.github/actions/workflow-run-job-windows/install_gpu_driver.ps1
+++ b/.github/actions/workflow-run-job-windows/install_gpu_driver.ps1
@@ -11,42 +11,13 @@ function Install-Driver {
     # This driver is picked to support Windows 11 & CUDA 12.8
     $version = '581.15'
 
+    # extract gpu type from gha runner label:
+    # Labels are in the form: <os>-<cpu>-gpu-<gpu>-<driver>-<n>
+    $gha_runner_label=$env:JOB_RUNNER
+    $gpu_type = $gha_runner_label.Split('-')[3]
+
     $data_center_gpus = @('a100', 'h100', 'l4', 't4', 'v100', 'rtxa6000', 'rtx6000ada')
     $desktop_gpus = @('rtx2080', 'rtx4090')
-
-    # Extract the gpu type from JOB_RUNNER (preferred) or the legacy RUNNER host name.
-    # Labels are in the form: <os>-<cpu>-gpu-<gpu>-<driver>-<n>
-    $gha_runner_label = if ($env:JOB_RUNNER) { $env:JOB_RUNNER } elseif ($env:RUNNER) { $env:RUNNER } else { $null }
-
-    if (-not $gha_runner_label)
-    {
-        Write-Output "Unable to determine GPU type: neither JOB_RUNNER nor RUNNER is set."
-        exit 1
-    }
-
-    $segments = $gha_runner_label.Split('-', [System.StringSplitOptions]::RemoveEmptyEntries)
-    $gpu_type = $null
-
-    foreach ($segment in $segments)
-    {
-        $normalized = $segment.ToLowerInvariant()
-        if ($data_center_gpus -contains $normalized -or $desktop_gpus -contains $normalized)
-        {
-            $gpu_type = $normalized
-            break
-        }
-    }
-
-    if (-not $gpu_type -and $segments.Length -gt 3)
-    {
-        $gpu_type = $segments[3].ToLowerInvariant()
-    }
-
-    if (-not $gpu_type)
-    {
-        Write-Output "Unknown GPU type in runner label: $gha_runner_label"
-        exit 1
-    }
 
     if ($data_center_gpus -contains $gpu_type) {
         Write-Output "Data center GPU detected: $gpu_type"
--- a/cub/cub/thread/thread_reduce.cuh
+++ b/cub/cub/thread/thread_reduce.cuh
@@ -156,31 +156,25 @@ template <typename Input,
 namespace detail
 {
 template <typename PreferredT, typename ValueT, typename ReductionOp, typename L, typename R>
-_CCCL_DEVICE _CCCL_FORCEINLINE auto thread_reduce_apply(ReductionOp&& reduction_op, L&& lhs, R&& rhs)
+_CCCL_DEVICE _CCCL_FORCEINLINE auto thread_reduce_apply(ReductionOp reduction_op, L lhs, R rhs)
 {
-  using _ReductionOpRef = ReductionOp&;
-  auto&& _op            = ::cuda::std::forward<ReductionOp>(reduction_op);
+  using ReductionOpRef = ReductionOp&;
 
-  if constexpr (::cuda::std::is_invocable_v<_ReductionOpRef, PreferredT, PreferredT>)
+  if constexpr (::cuda::std::is_invocable_v<ReductionOpRef, PreferredT, PreferredT>)
   {
-    return ::cuda::std::invoke(_op,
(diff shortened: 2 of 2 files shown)


cub/cub/block/block_reduce.cuh
```diff
@@ -443,7 +443,8 @@ public:
   _CCCL_DEVICE _CCCL_FORCEINLINE T Reduce(T (&inputs)[ITEMS_PER_THREAD], ReductionOp reduction_op)
   {
     // Reduce partials
-    T partial = cub::ThreadReduce(inputs, reduction_op);
+    T partial =
+      cub::ThreadReduce<::cuda::std::remove_reference_t<decltype(inputs)>, ReductionOp, T, T>(inputs, reduction_op);
```
fbusato: this looks a partial solution. It could regress for small integer types. For example, reduction/scan over `int8_t`. It is better to perform the computation with 32-bit and cast back at the end

cub/cub/block/block_reduce.cuh
```diff
@@ -601,7 +602,8 @@ public:
   _CCCL_DEVICE _CCCL_FORCEINLINE T Sum(T (&inputs)[ITEMS_PER_THREAD])
   {
     // Reduce partials
-    T partial = cub::ThreadReduce(inputs, ::cuda::std::plus<>{});
+    T partial = cub::ThreadReduce<::cuda::std::remove_reference_t<decltype(inputs)>, ::cuda::std::plus<>, T, T>(
```
fbusato: nit. please isolate the first template parameter with `using` to improve readability

fbusato (changes requested):

cub/cub/block/block_reduce.cuh
Aminsed: Good catch. I dropped the explicit T and taught ThreadReduce to keep its `__accumulator_t` promotion, so `int8_t` still widens to 32-bit.

cub/cub/block/block_reduce.cuh
Aminsed: @fbusato Thanks! I reverted that spot to plain `ThreadReduce(inputs, …)`, so there’s nothing left to alias. Let me know if you’d still like a `using` helper there.

fbusato:
pre-commit.ci autofix

Aminsed:
@fbusato  I fixed `thread_reduce_apply` so the reduction functor is forwarded and invoked via `cuda::std::invoke`, which keeps the `PreferredT` casting while still honoring const functors (this unblocks Thrust’s `key_flag_scan_op`). I re-ran the full cub-cpp20 preset inside rapidsai/devcontainers:25.12-cpp-llvm20-cuda13.0; all 1,427 build/test/bench targets completed successfully. Could you please rerun CI when convenient?

Aminsed:
Hi @fbusato  quick update before another CI spin. 
I only touched two spots: (1) install_gpu_driver.ps1 now pulls the GPU type from JOB_RUNNER (falling back to RUNNER) and scans the label segments, so the Windows jobs shouldn’t bail out before nvidia-smi; and (2) the new block-scan regression test marks its algorithm constant [[maybe_unused]], eliminating the NVHPC/MSVC /WX failure without changing behavior. I built cub.cpp17.test.block_scan.alg_0.mode_1 locally to double-check the warning is gone, but I can’t exercise the Windows runner script here. Could you please kick CI once more when convenient?

Aminsed:
@fbusato could you please run the test? Thanks

.github/actions/workflow-run-job-windows/install_gpu_driver.ps1
fbusato, line 1: please remove this file

cub/cub/thread/thread_reduce.cuh
```diff
@@ -155,6 +155,34 @@ template <typename Input,
 /// Internal namespace (to prevent ADL mishaps between static functions when mixing different CUB installations)
 namespace detail
 {
+template <typename PreferredT, typename ValueT, typename ReductionOp, typename L, typename R>
+_CCCL_DEVICE _CCCL_FORCEINLINE auto thread_reduce_apply(ReductionOp&& reduction_op, L&& lhs, R&& rhs)
```
fbusato: Universal references make little sense in device code. Please use  pass-by-value/const-value

.github/actions/workflow-run-job-windows/install_gpu_driver.ps1
fbusato, line 1: not sure why the file is still be visible even if it has been deleted.

Aminsed:
@fbusato The Dec 1 feedback has been addressed in commit e01a0bd2ca:

- Removed `install_gpu_driver.ps1` (out of scope)
- Changed `thread_reduce_apply` to use pass-by-value instead of universal references

Local tests all passing:
- `cub.test.block.scan`: 
- `cub.test.block.reduce`

@pauleonix Could you please confirm if this PR addresses all cases from issue #5668? The maintainers would like verification before merging.

Ready for CI when convenient! 🙏

Aminsed:
@fbusato 
Previous CI failures were infrastructure-related (missing script, numba-cuda version bug), not caused by this PR. Rebased onto current main to resolve.
Ready for CI

pauleonix:
> look fine. The original issue was opened by @pauleonix. It would be great to know if the PR fixes all cases of the original issues

Sorry that it took me so long to reply. This is not really what I had in mind when I wrote the issue. My view was that the problem is not with `ThreadReduce` but with how it is used by `BlockScan`. Therefore my solution would have been to just explicitly call the correct instantiation of `ThreadReduce` e.g. in the example from my issue:
```c++
T thread_prefix = cub::ThreadReduce<cuda::std::span<T, ITEMS_PER_THREAD>, ScanOp, T, T>(input, scan_op); 
```
On the other hand this is quite ugly/verbose and one can argue that this bug was introduced with changes to `ThreadReduce` (1e98f6f) and therefore should be fixed there. My main problem with that is that `ThreadReduce` is a public API and afaiu is currently working as intended. So changing its behavior could be a breaking change while changing `BlockScan` to use `ThreadReduce` "properly" would just recover documented behavior. 

Additionally this new (compile-time) logic seems to me like it is complicating things for the user, i.e. it becomes harder to argue about which overload of the functor is actually called in the end if there are multiple (an edge case but still).

PR #6448, opened 2025-11-03
Relax constexpr complex ops requirements

davebayer:
Just testing what will fail..


miscco (changes requested):
I would like to see a full CI matrix run on this, @alliepiper is there an easy way to test all compiler in C++17 for just libcu++?

alliepiper:
> I would like to see a full CI matrix run on this, @alliepiper is there an easy way to test all compiler in C++17 for just libcu++?

https://github.com/NVIDIA/cccl/blob/main/ci-overview.md#temporarily-overriding-the-pull-request-matrix

You can copy and modify the relevant lines from the nightly workflow -> `override:` to achieve this.

PR #6454, opened 2025-11-03
Drop build.log

bernhardmgruber:
This file seems to have been comitted by accident


bernhardmgruber:
> @bernhardmgruber please also add it to `python/cuda_cccl/.gitignore` to avoid this in the future.

#6473

PR #6461, opened 2025-11-03
Implement `__is_fully_bounded_array` trait

davebayer:
This PR introduces a trait which checks whether the passed type is a fully bounded array. I need this for `cudax::dynamic_shared_memory` launch option implementation.

The behaviour can be observed from:
```cpp
static_assert(!cuda::std::__is_all_bounded_array_v<int>);
static_assert(!cuda::std::__is_all_bounded_array_v<int[]>);
static_assert(cuda::std::__is_all_bounded_array_v<int[1]>);
static_assert(cuda::std::__is_all_bounded_array_v<int[1][2]>);
static_assert(!cuda::std::__is_all_bounded_array_v<int[][2]>);
static_assert(!cuda::std::__is_all_bounded_array_v<int[][1][1]>);
static_assert(cuda::std::__is_all_bounded_array_v<int[1][1][1]>);
```

### DIFF
--- a/libcudacxx/include/cuda/std/__type_traits/is_fully_bounded_array.h
+++ b/libcudacxx/include/cuda/std/__type_traits/is_fully_bounded_array.h
@@ -1,0 +1,47 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of libcu++, the C++ Standard Library for your entire system,
+// under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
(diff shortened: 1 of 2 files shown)


libcudacxx/include/cuda/std/__type_traits/is_fully_bounded_array.h
```diff
@@ -0,0 +1,46 @@
...
+#  pragma system_header
+#endif // no system header
+
+#include <cuda/std/__cstddef/types.h>
+
+#include <cuda/std/__cccl/prologue.h>
+
+_CCCL_BEGIN_NAMESPACE_CUDA_STD
+
+template <class _Tp>
```
miscco, line 30: This definitely needs some comments on what it wants to achieve, I am not sure whether its a good name, maybe `__is_fully_bounded_array` ?

PR #6469, opened 2025-11-04
[cudax->libcudacxx] Move device_transform to libcu++

pciolkosz:
This PR moves `device_transform` from cudax to libcu++. I had to adjust some things it used from cudax. One of them was `cuda::experimental::__variant`, which I replaced with `cuda::std::optional`. I believe it was just a simpler type we can later go back to.

With `device_transform` moved to libcu++ there was no longer a need to keep the experimental version of `copy_bytes` and `fill_bytes`, I removed them and added device transform to the non-experimental version.

I had to still leave experimental tests of the `algorithm` header, because of `uninitialized_buffer` used in them. We can move them when we move that type.


libcudacxx/include/cuda/__type_traits/is_instantiable_with.h
```diff
@@ -0,0 +1,47 @@
...
+
+template <template <class...> class _Fn, class... _Ts>
+_CCCL_CONCEPT __is_instantiable_with = requires { typename _Fn<_Ts...>; };
+
+#else // ^^^ _CCCL_HAS_CONCEPTS() ^^^ / vvv !_CCCL_HAS_CONCEPTS() vvv
+
+template <template <class...> class _Fn, class... _Ts>
+_CCCL_CONCEPT __is_instantiable_with = ::cuda::std::_IsValidExpansion<_Fn, _Ts...>::value;
+
+#endif // ^^^ !_CCCL_HAS_CONCEPTS() ^^^
```
davebayer, line 41: Can't we just do this:
```suggestion
template <template <class...> class _Fn, class... _Ts>
_CCCL_CONCEPT __is_instantiable_with = _CCCL_REQUIRES_EXPR(_Fn, variadic _Ts)( typename(_Fn<_Ts...>) );
```

libcudacxx/include/cuda/__stream/device_transform.h
```diff
@@ -152,7 +142,7 @@ struct _CCCL_TYPE_VISIBILITY_DEFAULT __device_transform_t
   [[nodiscard]] _CCCL_HOST_API auto operator()(_Stream&& __stream, _Arg&& __arg) const -> decltype(auto)
   {
     // Calls to transform_device_argument are intentionally unqualified so as to use ADL.
-    if constexpr (__is_instantiable_with<__transformed_argument_t, __transform_result_t<_Arg>>)
+    if constexpr (::cuda::__is_instantiable_with<__transformed_argument_t, __transform_result_t<_Arg>>)
```
davebayer: No need to fully qualify a concept
```suggestion
    if constexpr (__is_instantiable_with<__transformed_argument_t, __transform_result_t<_Arg>>)
```

libcudacxx/include/cuda/__stream/device_transform.h
```diff
@@ -120,26 +103,33 @@ struct _CCCL_TYPE_VISIBILITY_DEFAULT __device_transform_t
...
   // but sadly that is not valid C++.
+  // TODO move to use __variant type once cuda/experimental/execution/__variant is moved to libcudacxx
+  // NOTE: The above seems to only apply if the type is not trivially destructible. To use the optional here I had to
+  // add a destructor.
+  template <typename _Tp>
+  struct __optional_with_a_destructor : ::cuda::std::optional<_Tp>
+  {
+    using ::cuda::std::optional<_Tp>::optional;
+    ~__optional_with_a_destructor() {}
+  };
```
miscco, line 123: I do not understand why this needed, the destructor will still run at the same time, its just that there is nothing to destroy there

libcudacxx/include/cuda/__type_traits/is_instantiable_with.h
pciolkosz: I don't know if the C++17 version of `_CCCL_REQUIRES_EXPR` is compatible with `template <class...>`, I tried it and it fails, because inside the macro `_Fn` needs to have its template arguments list, but that would be basically the same expression as inside the require

libcudacxx/include/cuda/__stream/device_transform.h
davebayer, line 56: Design question: is device transform a good name for this object? Whenever I see the name, I am immediately thinking about some `cub` algorithm

bernhardmgruber: Same here.

pciolkosz: It started as `launch_transform` because it was used for `cuda::launch` API. Later we added it for `copy_bytes` and I believe some things Eric is working on. In general the idea was anytime execution is transferred to the device, arguments passed are going to this transformation. This is how `device_transform` name came up. Would `device_argument_transform` help with this concern? I can't think of a fundamentally different name for this

libcudacxx/include/cuda/__stream/device_transform.h
pciolkosz: I think for a trivially destructible type the compiler was reusing the stack before the entire expression was finished, corrupting the data held by the transformed argument

libcudacxx/include/cuda/__stream/device_transform.h
pciolkosz: I decided to merge the change as is and file an issue to rename this functionality https://github.com/NVIDIA/cccl/issues/6574

PR #6470, opened 2025-11-04
[Thrust]: New single-pass run length decoding example

brycelelbach:
Adds a new Thrust example demonstrating single-pass run-length decoding using inclusive_scan with a transform_output_iterator.

Conor thought it couldn't be done, but once again I have prevailed!

[Episode 255: 🇩🇰 C++ Copenhagen Meetup & Replicate](https://adspthepodcast.com/2025/10/10/Episode-255.html)
[Episode 256: 🇩🇰 Algorithms: Replicate, Scatter, Gather & RLD (Part 2)](https://adspthepodcast.com/2025/10/17/Episode-256.html)
[Episode 257: 🇳🇴 Live from Norway! Replicate, Scatter, Gather & RLD (Part 3)](https://adspthepodcast.com/2025/10/24/Episode-257.html)

If you code review this PR, you might be featured in part 4!

### DIFF
--- a/thrust/examples/run_length_decoding_single_pass.cu
+++ b/thrust/examples/run_length_decoding_single_pass.cu
@@ -14,10 +14,10 @@
 #include <thrust/scan.h>
 #include <thrust/sequence.h>
 
-#include <cuda/std/__algorithm/fill.h>
-#include <cuda/std/__algorithm/equal.h>
 #include <cuda/functional>
 #include <cuda/iterator>
+#include <cuda/std/__algorithm/equal.h>
+#include <cuda/std/__algorithm/fill.h>
 #include <cuda/std/cstdint>
 #include <cuda/std/ranges>
 


brycelelbach:
Here's a reproducer for the NVCC 13 + GCC 14 bug. I filed it upstream.

Examples should be exemplars, so I think we should leave the lovely ranges code and live with masking off the bad config instead of doing a raw for loop or something.

```
// Minimal reproducer for CUDA 13.0 nvcc + GCC 14 libstdc++ ranges incompatibility
//
// BUG DESCRIPTION:
// ================
// CUDA 13.0's nvcc fails to compile C++20 ranges code involving join_view with
// nested transform_view when using GCC 14's libstdc++. The same code compiles
// successfully with GCC 13's libstdc++, and GCC 14 alone can compile it without nvcc.
//
// ENVIRONMENT:
// ============
// - CUDA: 13.0.88
// - Host Compiler: GCC 14.2.0
// - Platform: Linux x86_64
// - libstdc++: GCC 14.2.0
//
// COMPILATION COMMANDS:
// =====================
// FAILS with nvcc + GCC 14:
//   nvcc -std=c++20 nvcc_ranges_join_view_bug.cu
//
// WORKS with nvcc + GCC 13:
//   nvcc -std=c++20 nvcc_ranges_join_view_bug.cu
//   (using GCC 13.3.0 as host compiler)
//
// WORKS with GCC 14 alone (not through nvcc):
//   g++ -std=c++20 nvcc_ranges_join_view_bug.cu
//
// ERROR MESSAGE:
// ==============
// /usr/include/c++/14/ranges(3164): error: no instance of constructor
// "std::ranges::join_view<_Vp>::_Iterator<_Const>::_Iterator [with
// _Vp=std::ranges::transform_view<std::ranges::iota_view<unsigned long,
// std::unreachable_sentinel_t>, lambda [](auto)->auto>, _Const=false]"
// matches the argument list
//
// The error indicates that nvcc's concepts evaluation fails to match any of the
// constrained constructors for join_view::_Iterator, even though one should match
// (the one with `requires forward_range<_Base>`).
//
// ROOT CAUSE:
// ===========
// CUDA 13.0's nvcc has incomplete C++20 concepts support. When evaluating the
// constraints on join_view::_Iterator constructors with complex nested types
// (transform_view containing lambdas), nvcc's concepts subsumption logic fails.
// GCC 14's libstdc++ has a more sophisticated ranges implementation than GCC 13,
// which triggers this bug.

#include <ranges>
#include <iostream>

int main() {
    using CountType = unsigned long;
    constexpr CountType repeat = 4;
    
    // Create a nested range: each element of the outer range is itself a range
    // This creates a 2D structure that needs to be flattened with join_view
    auto nested = std::views::iota(CountType{0}, CountType{8})
                  | std::views::transform([=](auto x) {
                      return std::views::iota(CountType{0}, repeat)
                          | std::views::transform([=](auto){ return x; });
                    });
    
    // join_view should flatten the nested ranges
    // This line triggers the compilation error with nvcc + GCC 14
    auto flattened = nested | std::views::join;
    
    // Attempting to get begin() iterator instantiates the problematic constructor
    auto it = flattened.begin();
    
    std::cout << "First element: " << *it << std::endl;
    return 0;
}

// EXPECTED BEHAVIOR:
// ==================
// The code should compile successfully and print "First element: 0"
//
// ACTUAL BEHAVIOR:
// ================
// Compilation fails with nvcc + GCC 14 due to failed concepts constraint evaluation
// in join_view::_Iterator constructor selection.
//
// WORKAROUND:
// ===========
// Use GCC 13 or earlier as the host compiler with CUDA 13.0.
```

thrust/examples/run_length_decoding_single_pass.cu
```diff
@@ -0,0 +1,169 @@
...
+      : out(out)
+      , out_size(out_size)
+      , runs_size(runs_size)
+      , expanded_size(expanded_size)
+  {}
+
+  template <typename ValueType>
+  __host__ __device__ CountType operator()(run<ValueType, CountType> r) const
+  {
+    cuda::std::size_t end = cuda::minimum()(r.offset + r.count, out_size);
```
miscco: This could just be 
```suggestion
    cuda::std::size_t end = cuda::std::min(r.offset + r.count, out_size);
```

thrust/examples/run_length_decoding_single_pass.cu
```diff
@@ -0,0 +1,169 @@
...
+  CountType run_id; // 1-index of the run in the input sequence; subtract by 1 to get the 0-index.
+
+  __host__ __device__ run(ValueType value = ValueType{}, CountType count = 0, CountType offset = 0, CountType run_id = 1)
+      : value(value)
+      , count(count)
+      , offset(offset)
+      , run_id(run_id)
+  {}
+  run(run const& other)            = default;
+  run& operator=(run const& other) = default;
```
miscco: Nitpick: What is the purpose here? do you want to delete move operations?

thrust/examples/run_length_decoding_single_pass.cu
```diff
@@ -0,0 +1,169 @@
...
+OutputIterator run_length_decode(
+  ValueIterator values, CountIterator counts, CountType runs_size, OutputIterator out, CountType out_size)
+{
+  using ValueType = typename ValueIterator::value_type;
+
+  // Zip the values and counts together and convert the resulting tuples to a named struct.
+  auto runs = thrust::make_transform_iterator(
+    thrust::make_zip_iterator(values, counts), [] __host__ __device__(thrust::tuple<ValueType, CountType> tup) {
+      return run{thrust::get<0>(tup), thrust::get<1>(tup)};
+    });
```
miscco: We do have `cuda::zip_transform_iterator`, just saying

thrust/examples/run_length_decoding_single_pass.cu
```diff
@@ -0,0 +1,169 @@
...
+  using CountType = cuda::std::size_t;
+
+  CountType size   = 8192;
+  CountType repeat = 4;
+
+  thrust::device_vector<ValueType> values(size);
+  thrust::sequence(thrust::device, values.begin(), values.end(), 0);
+
+  thrust::device_vector<CountType> counts(size);
+  thrust::fill(thrust::device, counts.begin(), counts.end(), repeat);
```
miscco, line 110: This should rather be 
```suggestion
  thrust::device_vector<CountType> counts(size, repeat);
```

thrust/examples/run_length_decoding_single_pass.cu
```diff
@@ -0,0 +1,169 @@
...
+  {
+    std::cout << observed[i];
+    if (i < 31)
+    {
+      std::cout << " ";
+    }
+  }
+  std::cout << std::endl;
+
+  auto gold = std::views::iota(CountType{0}) | std::views::transform([=](auto x) {
```
miscco: We do have `cuda::std::iota_view` and `cuda::std::transform_view` at home

miscco (changes requested):

thrust/examples/run_length_decoding_single_pass.cu
miscco: They are even available in C++17

thrust/examples/run_length_decoding_single_pass.cu
bernhardmgruber: Suggestion: delete special ops and just use aggregate init

thrust/examples/run_length_decoding_single_pass.cu
```diff
@@ -0,0 +1,169 @@
...
+  CountType runs_size;
+  ExpandedSizeIterator expanded_size;
+
+  __host__ __device__
+  expand(OutputIterator out, CountType out_size, CountType runs_size, ExpandedSizeIterator expanded_size)
+      : out(out)
+      , out_size(out_size)
+      , runs_size(runs_size)
+      , expanded_size(expanded_size)
+  {}
```
bernhardmgruber: Same here

thrust/examples/run_length_decoding_single_pass.cu
bernhardmgruber: Suggestion: use `zip_function` to unpack the tuple
```suggestion
  auto runs = thrust::make_transform_iterator(
    thrust::make_zip_iterator(values, counts), thrust::make_zip_function([] __host__ __device__(ValueType v CountType c) {
      return run{v, c};
    }));
```

thrust/examples/run_length_decoding_single_pass.cu
```diff
@@ -0,0 +1,169 @@
...
+      return run{thrust::get<0>(tup), thrust::get<1>(tup)};
+    });
+
+  // Allocate storage for the expanded size. If we were using CUB directly, we could read this
+  // from the final ScanTileState.
+  thrust::device_vector<CountType> expanded_size(1);
+
+  // Iterator that writes the expanded sequence to the output and discards the actual scan results.
+  auto expand_out = thrust::make_transform_output_iterator(
+    thrust::make_discard_iterator(), expand(out, out_size, runs_size, expanded_size.begin()));
```
bernhardmgruber: Suggestion: I would find a braced-init list easier to read here:
```suggestion
  auto expand_out = thrust::make_transform_output_iterator(
    thrust::make_discard_iterator(), expand{out, out_size, runs_size, expanded_size.begin()});
```
Plus, you can drop the ctor of `expand` now.

thrust/examples/run_length_decoding_single_pass.cu
```diff
@@ -0,0 +1,169 @@
...
+  }
+  std::cout << std::endl << std::endl;
+
+  thrust::device_vector<ValueType> output(size * repeat);
+
+  auto out_end = run_length_decode(values.begin(), counts.begin(), values.size(), output.begin(), output.size());
+
+  if (static_cast<CountType>(out_end - output.begin()) != size * repeat)
+  {
+    throw int{};
```
bernhardmgruber: Important: let's not show bad practices like throwing exceptions not derived from `std::exception`. Since this is in `main`, can we just print an error message and return something nonzero?

Applies one more time below.

thrust/examples/run_length_decoding_single_pass.cu
brycelelbach: Nice. Cursor, just use that instead.

brycelelbach: Cursor, make it so.

thrust/examples/run_length_decoding_single_pass.cu
brycelelbach: Cursor, just print an error message and return non-zero as suggested.

thrust/examples/run_length_decoding_single_pass.cu
brycelelbach: Ah, maybe they don't have the bug that NVCC / GCC14 has. Cursor, let's try using the cuda::std:: versions.

thrust/examples/run_length_decoding_single_pass.cu
brycelelbach: I prefer to be explicit. The device_vector constructor doesn't make it as clear what's going on.

thrust/examples/run_length_decoding_single_pass.cu
brycelelbach: Cursor, delete the `run` constructor and special member functions and just rely on aggregate initialization. You'll need to update places in the code that construct `run` objects. Prefer `run{}` syntax to `run()` syntax.

brycelelbach: Will do.

thrust/examples/run_length_decoding_single_pass.cu
brycelelbach: Cursor, delete the constructor for `expand`.

thrust/examples/run_length_decoding_single_pass.cu
brycelelbach: Cursor, make it so.

thrust/examples/run_length_decoding_single_pass.cu
brycelelbach: Cursor, make it so.

thrust/examples/run_length_decoding_single_pass.cu
brycelelbach: Just used `zip_transform_iterator` instead.

thrust/examples/run_length_decoding_single_pass.cu
bernhardmgruber: > I prefer to be explicit. The device_vector constructor doesn't make it as clear what's going on.

While this may be a somewhat valid argument, the code now writes to memory twice. The vector constructor zeros it, then the fill overwrites all the zeros. I think this is bad practice as well.

thrust/examples/run_length_decoding_single_pass.cmake
```diff
@@ -0,0 +1,41 @@
+# This example uses a single-pass algorithm that performs side effects during scan.
+# The algorithm relies on the scan functor being called exactly once per element with the final
+# scan value, which is guaranteed by CUDA but not by CPU-parallel backends like OMP/TBB where
+# the functor may be called multiple times with intermediate values during parallel reduction.
```
bernhardmgruber, line 4: Suggestion: I think we should also put this disclaimer into the example source code.

thrust/examples/run_length_decoding_single_pass.cu
```diff
@@ -0,0 +1,156 @@
...
+    values,
+    counts);
+
+  // Allocate storage for the expanded size. If we were using CUB directly, we could read this
+  // from the final ScanTileState.
+  thrust::device_vector<CountType> expanded_size(1);
+
+  // Iterator that writes the expanded sequence to the output and discards the actual scan results.
+  auto expand_out = thrust::make_transform_output_iterator(
+    thrust::make_discard_iterator(), expand{out, out_size, runs_size, expanded_size.begin()});
```
bernhardmgruber, line 89: Let's advertise the new CUDA iterators:
```suggestion
  auto expand_out = cuda::make_transform_output_iterator(
    cuda::make_discard_iterator(), expand{out, out_size, runs_size, expanded_size.begin()});
```

thrust/examples/run_length_decoding_single_pass.cmake
brycelelbach: That comment may not be entirely accurate, it's a question of how many times the output iterator is written to, because that's what invokes the functor.

thrust/examples/run_length_decoding_single_pass.cu
```diff
@@ -0,0 +1,156 @@
...
+  OutputIterator out;
+  CountType out_size;
+  CountType runs_size;
+  ExpandedSizeIterator expanded_size;
+
+  template <typename ValueType>
+  __host__ __device__ CountType operator()(run<ValueType, CountType> r) const
+  {
+    cuda::std::size_t end = cuda::std::min(r.offset + r.count, out_size);
+    cuda::std::fill(out + r.offset, out + end, r.value);
```
AlexanderSinn, line 56: Assuming this does something like
```C++
for (auto it = out + r.offset; it != out + end; ++it) {
    *it = out.value;
}
```
for each thread, then this will result in load imbalance and a poor memory access pattern for some inputs. If instead one would have full control of the thread block after the device-wide inclusive scan finishes, multiple block-wide max scans could be used to fix the memory access pattern and do load balancing across the threads within a block.

PR #6471, opened 2025-11-04
Add python wrappers for c.parallel segmented_sort

### LINKED ISSUE #5497: Add python wrappers for segmented sort

NaderAlAwar:
## Description

closes #5497

## Checklist

### DIFF
--- a/python/cuda_cccl/cuda/compute/algorithms/__init__.py
+++ b/python/cuda_cccl/cuda/compute/algorithms/__init__.py
@@ -13,13 +13,13 @@ from ._scan import make_exclusive_scan as make_exclusive_scan
 from ._scan import make_inclusive_scan as make_inclusive_scan
 from ._segmented_reduce import make_segmented_reduce as make_segmented_reduce
 from ._segmented_reduce import segmented_reduce
-from ._sort._merge_sort import make_merge_sort as make_merge_sort
-from ._sort._merge_sort import merge_sort as merge_sort
-from ._sort._radix_sort import make_radix_sort as make_radix_sort
-from ._sort._radix_sort import radix_sort as radix_sort
-from ._sort._segmented_sort import make_segmented_sort as make_segmented_sort
-from ._sort._segmented_sort import segmented_sort as segmented_sort
-from ._sort._sort_common import DoubleBuffer, SortOrder
+from ._sort import DoubleBuffer, SortOrder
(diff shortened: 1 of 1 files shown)


python/cuda_cccl/cuda/compute/_bindings_impl.pyx
```diff
@@ -2271,3 +2271,150 @@ cdef class DeviceThreeWayPartitionBuildResult:
...
+        cccl_iterator_t d_keys_in,
+        cccl_iterator_t d_keys_out,
+        cccl_iterator_t begin_offset_in,
+        cccl_iterator_t end_offset_in,
+        int,
+        int,
+        const char *,
+        const char *,
+        const char *,
+        const char *
```
shwina: For consistency/completeness can we give names to all parameters?

python/cuda_cccl/cuda/compute/algorithms/__init__.py
```diff
@@ -18,6 +13,13 @@
 from ._scan import make_inclusive_scan as make_inclusive_scan
 from ._segmented_reduce import make_segmented_reduce as make_segmented_reduce
 from ._segmented_reduce import segmented_reduce
+from ._sort._merge_sort import make_merge_sort as make_merge_sort
+from ._sort._merge_sort import merge_sort as merge_sort
+from ._sort._radix_sort import make_radix_sort as make_radix_sort
+from ._sort._radix_sort import radix_sort as radix_sort
+from ._sort._segmented_sort import make_segmented_sort as make_segmented_sort
+from ._sort._segmented_sort import segmented_sort as segmented_sort
+from ._sort._sort_common import DoubleBuffer, SortOrder
```
shwina: Let's import these names in `_sort/__init__.py`, so that we can just do here (and elsewhere):

```python
from ._sort import merge_sort, radix_sort, segmented_sort
```

The organization of submodules in `_sort` then becomes an "implementation detail" of the `_sort` package.

python/cuda_cccl/cuda/compute/_bindings_impl.pyx
NaderAlAwar: We don't give names to these parameters for the other algorithms, these are things like include paths and compute capabilities, but for consistency I will format them so that they are all on the same line like the others

PR #6476, opened 2025-11-04
Implement CUDA backend for parallel `cuda::std::reduce`

miscco:
This implements a cuda backend for `cuda::std::reduce`

### DIFF
--- a/libcudacxx/include/cuda/std/__pstl/cuda/reduce.h
+++ b/libcudacxx/include/cuda/std/__pstl/cuda/reduce.h
@@ -1,0 +1,139 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of libcu++, the C++ Standard Library for your entire system,
+// under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef _CUDA_STD___PSTL_CUDA_REDUCE_H
+#define _CUDA_STD___PSTL_CUDA_REDUCE_H
+
+#include <cuda/std/detail/__config>
+
+#if defined(_CCCL_IMPLICIT_SYSTEM_HEADER_GCC)
+#  pragma GCC system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_CLANG)
+#  pragma clang system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_MSVC)
+#  pragma system_header
+#endif // no system header
+
+#if _CCCL_HAS_BACKEND_CUDA()
+
+_CCCL_DIAG_PUSH
+_CCCL_DIAG_SUPPRESS_CLANG("-Wshadow")
+
+#  include <cub/device/device_reduce.cuh>
+
+_CCCL_DIAG_POP
+
+#  include <cuda/__execution/policy.h>
+#  include <cuda/__runtime/api_wrapper.h>
+#  include <cuda/std/__exception/cuda_error.h>
+#  include <cuda/std/__execution/env.h>
+#  include <cuda/std/__execution/policy.h>
+#  include <cuda/std/__iterator/distance.h>
+#  include <cuda/std/__iterator/iterator_traits.h>
+#  include <cuda/std/__memory/addressof.h>
+#  include <cuda/std/__new/bad_alloc.h>
+#  include <cuda/std/__numeric/reduce.h>
+#  include <cuda/std/__pstl/dispatch.h>
+#  include <cuda/std/__type_traits/always_false.h>
+#  include <cuda/std/__utility/move.h>
+
+#  include <cuda_runtime.h>
+
+#  include <cuda/std/__cccl/prologue.h>
+
+_CCCL_BEGIN_NAMESPACE_CUDA_STD_EXECUTION
+
+_CCCL_BEGIN_NAMESPACE_ARCH_DEPENDENT
+
+template <>
+struct __pstl_dispatch<__pstl_algorithm::__reduce, __execution_backend::__cuda>
+{
+  template <class _Policy, class _Iter, class _Tp, class _BinaryOp>
+  [[nodiscard]] _CCCL_HOST_API static _Tp
+  __par_impl(_Policy __policy, _Iter __first, _Iter __last, _Tp __init, _BinaryOp __func)
+  {
+    // Allocate memory for result
+    _Tp* __device_ret_ptr = nullptr;
+
+    _CCCL_TRY_CUDA_API(
+      ::cudaMalloc, "__pstl_cuda_reduce: allocation failed", reinterpret_cast<void**>(&__device_ret_ptr), sizeof(_Tp));
+
+    const auto __count = ::cuda::std::distance(__first, __last);
+    _Tp __ret;
+
+    _CCCL_TRY_CUDA_API(
+      ::cub::DeviceReduce::Reduce,
+      "__pstl_cuda_reduce: cub::DeviceReduce::Reduce failed",
+      ::cuda::std::move(__first),
+      __device_ret_ptr,
+      __count,
+      ::cuda::std::move(__func),
+      ::cuda::std::move(__init),
+      ::cuda::std::move(__policy));
+
+    _CCCL_TRY_CUDA_API(
+      ::cudaMemcpy,
+      "__pstl_cuda_reduce: copy of result from device to host failed",
+      ::cuda::std::addressof(__ret),
+      __device_ret_ptr,
+      sizeof(_Tp),
+      ::cudaMemcpyDeviceToHost);
+
+    _CCCL_TRY_CUDA_API(::cudaFree, "__pstl_cuda_reduce: deallocate failed", __device_ret_ptr);
+
+    return __ret;
+  }
+
+  template <class _Policy, class _Iter, class _Tp, class _BinaryOp>
+  [[nodiscard]] _CCCL_HOST_API _Tp
+  operator()([[maybe_unused]] _Policy __policy, _Iter __first, _Iter __last, _Tp __init, _BinaryOp __func) const
+  {
+    if constexpr (::cuda::std::__has_random_access_traversal<_Iter>)
+    {
+      try
+      {
+        return __par_impl(
+          ::cuda::std::move(__policy),
+          ::cuda::std::move(__first),
+          ::cuda::std::move(__last),
+          ::cuda::std::move(__init),
+          ::cuda::std::move(__func));
+      }
+      catch (const ::cuda::cuda_error& __err)
+      {
+        if (__err.status() == cudaErrorMemoryAllocation)
+        {
+          ::cuda::std::__throw_bad_alloc();
+        }
+        else
+        {
+          throw __err;
+        }
+      }
+    }
+    else
+    {
+      static_assert(__always_false_v<_Policy>,
+                    "__pstl_dispatch: CUDA backend of cuda::std::reduce requires at least random access iterators");
+      return ::cuda::std::reduce(
+        ::cuda::std::move(__first), ::cuda::std::move(__last), ::cuda::std::move(__init), ::cuda::std::move(__func));
+    }
+  }
+};
+
+_CCCL_END_NAMESPACE_ARCH_DEPENDENT
+
+_CCCL_END_NAMESPACE_CUDA_STD_EXECUTION
+
+#  include <cuda/std/__cccl/epilogue.h>
+
+#endif /// _CCCL_HAS_BACKEND_CUDA()
+
+#endif // _CUDA_STD___PSTL_CUDA_REDUCE_H
--- a/libcudacxx/include/cuda/std/__pstl/reduce.h
+++ b/libcudacxx/include/cuda/std/__pstl/reduce.h
@@ -1,0 +1,115 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of libcu++, the C++ Standard Library for your entire system,
+// under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef _CUDA_STD___PSTL_REDUCE_H
+#define _CUDA_STD___PSTL_REDUCE_H
+
+#include <cuda/std/detail/__config>
+
+#if defined(_CCCL_IMPLICIT_SYSTEM_HEADER_GCC)
+#  pragma GCC system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_CLANG)
+#  pragma clang system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_MSVC)
+#  pragma system_header
+#endif // no system header
+
+#if !_CCCL_COMPILER(NVRTC)
+
+#  include <cuda/std/__concepts/concept_macros.h>
+#  include <cuda/std/__execution/policy.h>
+#  include <cuda/std/__functional/invoke.h>
+#  include <cuda/std/__functional/operations.h>
+#  include <cuda/std/__iterator/iterator_traits.h>
+#  include <cuda/std/__iterator/readable_traits.h>
+#  include <cuda/std/__numeric/reduce.h>
+#  include <cuda/std/__pstl/dispatch.h>
+#  include <cuda/std/__type_traits/always_false.h>
+#  include <cuda/std/__type_traits/is_convertible.h>
+#  include <cuda/std/__type_traits/is_move_constructible.h>
+#  include <cuda/std/__utility/move.h>
+
+#  if _CCCL_HAS_BACKEND_CUDA()
+#    include <cuda/std/__pstl/cuda/reduce.h>
+#  endif // _CCCL_HAS_BACKEND_CUDA()
+
+#  include <cuda/std/__cccl/prologue.h>
+
+_CCCL_BEGIN_NAMESPACE_CUDA_STD
+
+template <class _Iter, class _Tp, class _BinaryOp>
+_CCCL_CONCEPT __indirect_binary_function = _CCCL_REQUIRES_EXPR((_Iter, _Tp, _BinaryOp))(
+  requires(is_convertible_v<invoke_result_t<_BinaryOp&, iter_reference_t<_Iter>, _Tp>, _Tp>),
+  requires(is_convertible_v<invoke_result_t<_BinaryOp&, _Tp, iter_reference_t<_Iter>>, _Tp>),
+  requires(is_convertible_v<invoke_result_t<_BinaryOp&, _Tp, _Tp>, _Tp>),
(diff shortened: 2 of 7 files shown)


libcudacxx/include/cuda/std/__pstl/cuda/reduce.h
```diff
@@ -0,0 +1,128 @@
...
+  template <class _Policy, class _Iter, class _Tp, class _BinaryOp>
+  [[nodiscard]] _CCCL_HOST_API static _Tp
+  __par_impl(const _Policy& __policy, _Iter __first, _Iter __last, _Tp __init, _BinaryOp __func)
+  {
+    // Allocate memory for result
+    _Tp* __device_ret_ptr = nullptr;
+    try
+    {
+      _CCCL_TRY_CUDA_API(
+        ::cudaMalloc, "__pstl_dispatch: allocation failed", reinterpret_cast<void**>(&__device_ret_ptr), sizeof(_Tp));
```
bernhardmgruber: Critical: we should use `cudaMallocAsync`.

libcudacxx/include/cuda/std/__pstl/cuda/reduce.h
```diff
@@ -0,0 +1,128 @@
...
+    }
+    catch (...)
+    {
+      ::cuda::std::__throw_bad_alloc();
+    }
```
bernhardmgruber: So we are checking the return code of a C API to throw an exception if it's not `cudaSuccess`, only to catch this exception, discard it and then throw `bad_allow`.

Suggestion: just check return code and throw:
```suggestion
    if(::cudaMalloc(reinterpret_cast<void**>(&__device_ret_ptr), sizeof(_Tp)) != cudaSuccess)
      ::cuda::std::__throw_bad_alloc();
```

libcudacxx/include/cuda/std/__pstl/cuda/reduce.h
```diff
@@ -0,0 +1,128 @@
...
+
+    const auto __count = ::cuda::std::distance(__first, __last);
+    _Tp __ret;
+
+    try
+    {
+      _CCCL_TRY_CUDA_API(
+        ::cub::DeviceReduce::Reduce,
+        "__pstl_dispatch: kernel launch failed",
```
bernhardmgruber: Important: cub::DeviceReduce can fail due to a lot of reasons, some are not kernel launches. Let's fix the error message:
```suggestion
      _CCCL_TRY_CUDA_API(
        ::cub::DeviceReduce::Reduce,
        "__pstl_dispatch: calling cub::DeviceReduce::Reduce failed",
```

libcudacxx/include/cuda/std/__pstl/cuda/reduce.h
```diff
@@ -0,0 +1,128 @@
...
+        "__pstl_dispatch: copy of result from device to host failed",
+        ::cuda::std::addressof(__ret),
+        __device_ret_ptr,
+        sizeof(_Tp),
+        ::cudaMemcpyDeviceToHost);
+    }
+    catch (...)
+    {
+      ::cuda::std::terminate();
+    }
```
bernhardmgruber: Again, if we terminate in case any of the CUDA or CUB APIs fail, then we should just check their error codes and terminate. No need to throw an exception and catch it ourselves.

bernhardmgruber (commented):
LGTM, except for the error handling strategy. Are we not allowed to throw any exceptions out of a PSTL call?

In the C++ draft I see: https://eel.is/c++draft/algorithms.parallel#exceptions-2

> During the execution of a parallel algorithm, if the invocation of an element access function exits via an uncaught exception, the behavior is determined by the execution policy[.](https://eel.is/c++draft/algorithms.parallel#exceptions-2.sentence-1)

So if we are implementing any other policy than the standard ones, I think we can just throw exceptions, right? So why terminate and provide little help to the user?

libcudacxx/include/cuda/std/__pstl/cuda/reduce.h
davebayer: I don't think it's important right now, but we should change this once we start to care about streams in the pstl implementation.

I would leave it as is and add `todo` comment.

davebayer: also this should be replaced by CUDA Driver call in the future

libcudacxx/include/cuda/std/__pstl/cuda/reduce.h
davebayer: That's actually not quite correct. I would throw `std::bad_alloc` only if `cudaErrorMemoryAllocation` is returned, otherwise, we should terminate for standard execution policies.

libcudacxx/include/cuda/std/__pstl/cuda/reduce.h
miscco: We currently do not have a stream and I was told that that is not part of the MVP

libcudacxx/include/cuda/std/__pstl/cuda/reduce.h
```diff
@@ -0,0 +1,128 @@
...
+        __policy);
+
+      _CCCL_TRY_CUDA_API(
+        ::cudaMemcpy,
+        "__pstl_dispatch: copy of result from device to host failed",
+        ::cuda::std::addressof(__ret),
+        __device_ret_ptr,
+        sizeof(_Tp),
+        ::cudaMemcpyDeviceToHost);
+    }
```
davebayer: I am missing `cudaFree` of the allocated memory

libcudacxx/include/cuda/std/__pstl/cuda/reduce.h
miscco: I believe this hinges on whether we want to to actually propagate exceptions out.

If we do, then we can just drop the whole try-catch block

davebayer: @bernhardmgruber it's how the behaviour is defined for standard execution policies. But I agree that we should throw for any other policies, I'd like to see it handled here.

libcudacxx/include/cuda/std/__pstl/cuda/reduce.h
bernhardmgruber: Alright. We should have a stream in the execution policy at some point, which we can use here.

libcudacxx/include/cuda/std/__pstl/cuda/reduce.h
bernhardmgruber: Will any standard execution policy ever go to the CUDA backend? Maybe if NVHPC ever uses this?

cub/cub/device/device_reduce.cuh
```diff
@@ -1281,30 +1283,32 @@ public:
...
     using requirements_t =
-      _CUDA_STD_EXEC::__query_result_or_t<EnvT, _CUDA_EXEC::__get_requirements_t, _CUDA_STD_EXEC::env<>>;
+      ::cuda::std::execution::__query_result_or_t<EnvT, _CUDA_EXEC::__get_requirements_t, ::cuda::std::execution::env<>>;
     using requested_determinism_t =
-      _CUDA_STD_EXEC::__query_result_or_t<requirements_t, //
-                                          _CUDA_EXEC::determinism::__get_determinism_t,
-                                          _CUDA_EXEC::determinism::run_to_run_t>;
+      ::cuda::std::execution::__query_result_or_t<requirements_t, //
+                                                  _CUDA_EXEC::determinism::__get_determinism_t,
+                                                  _CUDA_EXEC::determinism::run_to_run_t>;
```
davebayer: Wouldn't be better to move these changes to a separate PR?

libcudacxx/include/cuda/std/__pstl/cuda/reduce.h
```diff
@@ -0,0 +1,123 @@
...
+      ::cub::DeviceReduce::Reduce,
+      "__pstl_cuda_for_each: cub::DeviceReduce::Reduce failed",
+      ::cuda::std::move(__first),
+      __device_ret_ptr,
+      __count,
+      ::cuda::std::move(__func),
+      ::cuda::std::move(__init),
+      __policy);
+
+    _CCCL_TRY_CUDA_API(
```
davebayer, line 81: If this function should take standard execution policies, the execution should terminate in case of an error

libcudacxx/include/cuda/std/__pstl/cuda/reduce.h
```diff
@@ -0,0 +1,123 @@
...
+{
+  template <class _Policy, class _Iter, class _Tp, class _BinaryOp>
+  [[nodiscard]] _CCCL_HOST_API static _Tp
+  __par_impl(const _Policy& __policy, _Iter __first, _Iter __last, _Tp __init, _BinaryOp __func)
+  {
+    // Allocate memory for result
+    _Tp* __device_ret_ptr = nullptr;
+
+    _CCCL_TRY_CUDA_API(
+      ::cudaMalloc, "__pstl_dispatch: allocation failed", reinterpret_cast<void**>(&__device_ret_ptr), sizeof(_Tp));
```
davebayer: Should throw `std::bad_alloc` as you had it before

cub/cub/device/device_reduce.cuh
davebayer: Aaah, already moved to #6522

libcudacxx/include/cuda/std/__pstl/cuda/reduce.h
miscco: Currently this function can never take standard execution policy because for now its always an implementation defined custom cuda specific one

libcudacxx/include/cuda/std/__pstl/cuda/reduce.h
miscco: The current answer is no

libcudacxx/include/cuda/std/__pstl/cuda/reduce.h
miscco: I am kind of torn, it needlessly complicates the implementation

davebayer: But that's what the standard requires

libcudacxx/include/cuda/std/__pstl/dispatch.h
```diff
@@ -0,0 +1,123 @@
...
+[[nodiscard]] _CCCL_API _CCCL_CONSTEVAL auto __pstl_select_dispatch() noexcept
+{
+  // First extract the desired backend from the policy
+  constexpr __execution_backend __backend = _Policy::__get_backend();
+
+  // If the user requests a unique backends, we must take that
+  if constexpr (::cuda::std::execution::__has_unique_backend(__backend))
+  {
+    return __pstl_dispatch<_Algorithm, __backend>{};
+  }
```
bernhardmgruber, line 109: I would strongly recommend to statically assert this property. I don't see a use case for allowing multiple backends and having us choose one. If a user wants to dispatch to multiple backends, then they should implement this logic.
```suggestion
  static_assert(::cuda::std::execution::__has_unique_backend(__backend));
  return __pstl_dispatch<_Algorithm, __backend>{};
```

libcudacxx/include/cuda/std/__pstl/for_each.h
```diff
@@ -0,0 +1,67 @@
...
+
+#  include <cuda/std/__cccl/prologue.h>
+
+_CCCL_BEGIN_NAMESPACE_CUDA_STD
+
+_CCCL_BEGIN_ARCH_VISIBILITY_NAMESPACE
+
+_CCCL_TEMPLATE(class _Policy, class _Iter, class _Fn)
+_CCCL_REQUIRES(__has_forward_traversal<_Iter> _CCCL_AND is_execution_policy_v<_Policy>)
+_CCCL_HOST_API void for_each(const _Policy& __pol, _Iter __first, _Iter __last, _Fn __func)
```
bernhardmgruber: Are policies passed by `const&`? Shouldn't we pass them by value and move them into the next call?

libcudacxx/include/cuda/std/__pstl/dispatch.h
miscco: I am a bit torn here, because that would break all used of standard execution policies without a fall back.

libcudacxx/include/cuda/std/__pstl/cuda/reduce.h
miscco: Changed it so that we catch `cudaErrorMemoryAllocation` in the call operator

libcudacxx/include/cuda/std/__pstl/cuda/reduce.h
```diff
@@ -0,0 +1,139 @@
...
+{
+  template <class _Policy, class _Iter, class _Tp, class _BinaryOp>
+  [[nodiscard]] _CCCL_HOST_API static _Tp
+  __par_impl(_Policy __policy, _Iter __first, _Iter __last, _Tp __init, _BinaryOp __func)
+  {
+    // Allocate memory for result
+    _Tp* __device_ret_ptr = nullptr;
+
+    _CCCL_TRY_CUDA_API(
+      ::cudaMalloc, "__pstl_dispatch: allocation failed", reinterpret_cast<void**>(&__device_ret_ptr), sizeof(_Tp));
```
bernhardmgruber: Important: this should allocate in stream order:
```suggestion
    _CCCL_TRY_CUDA_API(
      ::cudaMallocAsync, "__pstl_dispatch: allocation failed", reinterpret_cast<void**>(&__device_ret_ptr), sizeof(_Tp), /* stream */ 0);
```

libcudacxx/include/cuda/std/__pstl/reduce.h
```diff
@@ -0,0 +1,115 @@
...
+_CCCL_BEGIN_NAMESPACE_ARCH_DEPENDENT
+
+_CCCL_TEMPLATE(class _Policy, class _Iter, class _Tp, class _BinaryOp)
+_CCCL_REQUIRES(__has_forward_traversal<_Iter> _CCCL_AND is_execution_policy_v<_Policy>)
+[[nodiscard]] _CCCL_HOST_API _Tp
+reduce([[maybe_unused]] _Policy __policy, _Iter __first, _Iter __last, _Tp __init, _BinaryOp __func)
+{
+  static_assert(__indirect_binary_function<_Iter, _Tp, _BinaryOp>,
+                "cuda::std::reduce: The return value of BinaryOp is not convertible to T.");
+  static_assert(is_move_constructible_v<_Tp>, "cuda::std::reduce: T must be move constructible.");
```
bernhardmgruber, line 63: Important: In the error message `T` is mentioned, but the function takes a `_Tp`. Please use `_Tp` in the error message as well.

libcudacxx/include/cuda/std/__pstl/cuda/reduce.h
```diff
@@ -0,0 +1,139 @@
...
+
+    _CCCL_TRY_CUDA_API(
+      ::cudaMemcpy,
+      "__pstl_cuda_for_each: copy of result from device to host failed",
+      ::cuda::std::addressof(__ret),
+      __device_ret_ptr,
+      sizeof(_Tp),
+      ::cudaMemcpyDeviceToHost);
+
+    _CCCL_TRY_CUDA_API(::cudaFree, "__pstl_cuda_for_each: deallocate failed", __device_ret_ptr);
```
bernhardmgruber:
```suggestion
    _CCCL_TRY_CUDA_API(::cudaFreeAsync, "__pstl_cuda_for_each: deallocate failed", __device_ret_ptr, /* stream */ 0);
```

libcudacxx/include/cuda/std/__pstl/cuda/reduce.h
```diff
@@ -0,0 +1,139 @@
...
+
+template <>
+struct __pstl_dispatch<__pstl_algorithm::__reduce, __execution_backend::__cuda>
+{
+  template <class _Policy, class _Iter, class _Tp, class _BinaryOp>
+  [[nodiscard]] _CCCL_HOST_API static _Tp
+  __par_impl(_Policy __policy, _Iter __first, _Iter __last, _Tp __init, _BinaryOp __func)
+  {
+    // Allocate memory for result
+    _Tp* __device_ret_ptr = nullptr;
```
davebayer, line 63:
```suggestion
    _Tp* __device_ret_ptr = nullptr;
    ::cuda::stream_ref __stream{::cudaStream_t{}};
```
Can we add a stream we will launch the CUDA calls in? Let it be just the default stream for now.

libcudacxx/include/cuda/std/__pstl/cuda/reduce.h
davebayer:
```suggestion
      ::cudaMallocAsync, "__pstl_dispatch: allocation failed", reinterpret_cast<void**>(&__device_ret_ptr), sizeof(_Tp), __stream.get());
```

libcudacxx/include/cuda/std/__pstl/cuda/reduce.h
```diff
@@ -0,0 +1,139 @@
...
+
+    _CCCL_TRY_CUDA_API(
+      ::cub::DeviceReduce::Reduce,
+      "__pstl_cuda_for_each: cub::DeviceReduce::Reduce failed",
+      ::cuda::std::move(__first),
+      __device_ret_ptr,
+      __count,
+      ::cuda::std::move(__func),
+      ::cuda::std::move(__init),
+      ::cuda::std::move(__policy));
```
davebayer, line 79:
```suggestion
      ::cuda::std::move(__init),
      ::cuda::std::move(__policy),
      __stream.get());
```

libcudacxx/include/cuda/std/__pstl/cuda/reduce.h
```diff
@@ -0,0 +1,139 @@
...
+
+    _CCCL_TRY_CUDA_API(
+      ::cudaMemcpy,
+      "__pstl_cuda_for_each: copy of result from device to host failed",
+      ::cuda::std::addressof(__ret),
+      __device_ret_ptr,
+      sizeof(_Tp),
+      ::cudaMemcpyDeviceToHost);
```
davebayer, line 87:
```suggestion
      ::cudaMemcpyAsync,
      "__pstl_cuda_for_each: copy of result from device to host failed",
      ::cuda::std::addressof(__ret),
      __device_ret_ptr,
      sizeof(_Tp),
      ::cudaMemcpyDeviceToHost,
      __stream.get());
```

libcudacxx/include/cuda/std/__pstl/cuda/reduce.h
davebayer:
```suggestion
    _CCCL_TRY_CUDA_API(::cudaFreeAsync, "__pstl_cuda_for_each: deallocate failed", __device_ret_ptr, __stream.get());
```

libcudacxx/include/cuda/std/__pstl/cuda/reduce.h
```diff
@@ -0,0 +1,139 @@
...
+
+    _CCCL_TRY_CUDA_API(::cudaFree, "__pstl_cuda_for_each: deallocate failed", __device_ret_ptr);
+
```
davebayer, line 90:
```suggestion
    __stream.sync();
```

libcudacxx/include/cuda/std/__pstl/reduce.h
miscco: This is a bit icky

We conventionally use the template arguments that would appear in the documentation, except when dealing with STL APIs because in the docs for those its called `T`, see https://en.cppreference.com/w/cpp/algorithm/reduce.html

libcudacxx/include/cuda/std/__pstl/cuda/reduce.h
miscco: I explicitly do not want to deal with streams here at all, that is what #6503 is about

libcudacxx/include/cuda/std/__pstl/cuda/reduce.h
miscco: I explicitly do not want to deal with streams here at all, that is what #6503 is about

PR #6487, opened 2025-11-04
Enable `__int128_t` as difference type in `counting_iterator`

### LINKED ISSUE #6485: [BUG]: mismatch fails with "Found integer-like type that is bigger than largest integer like type."
### Is this a duplicate?

### Type of Bug

Compile-time Error

### Component

Thrust

### Describe the bug

For this mismatch test:
```
#include <execution>
#include <algorithm>
#include <vector>
#include <ranges>

int main() {
  std::vector<long> v(100);
  auto i = std::views::iota(0L, 100L);
  std::mismatch(std::execution::par, v.begin(), v.end(), i.begin(), i.end());
}
```

it has started failing at compile time with the following error:
```
cuda/__iterator/counting_iterator.h", line 87: error: static assertion failed with "Found integer-like type that is bigger than largest integer like type."
      static_assert(sizeof(_Int) <= sizeof(long long),
      ^
```
I suspect this regression may have been introduced by commit 8ca3dc31e9075e038d2

### How to Reproduce

```
nvc++ -std=c++23 --gcc-toolchain=/home/sw/thirdparty/gcc/gcc-13.1.0/Linux_x86_64 -stdpar=gpu mismatch_min.cpp
```

### Expected behavior

The test should compile fine after the fix. 

### Reproduction link

_No response_

### Operating System

_No response_

### nvidia-smi output

_No response_

### NVCC version

_No response_

miscco:
Fixes #6485

### DIFF
--- a/libcudacxx/include/cuda/__iterator/counting_iterator.h
+++ b/libcudacxx/include/cuda/__iterator/counting_iterator.h
@@ -41,6 +41,7 @@
 #include <cuda/std/__ranges/enable_borrowed_range.h>
 #include <cuda/std/__ranges/movable_box.h>
 #include <cuda/std/__ranges/view_interface.h>
+#include <cuda/std/__type_traits/always_false.h>
 #include <cuda/std/__type_traits/conditional.h>
 #include <cuda/std/__type_traits/enable_if.h>
 #include <cuda/std/__type_traits/is_constructible.h>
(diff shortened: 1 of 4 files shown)


libcudacxx/include/cuda/__iterator/counting_iterator.h
```diff
@@ -80,12 +80,26 @@ struct __get_wider_signed
       return ::cuda::std::type_identity<long>{};
     }
     else
+#if _CCCL_HAS_INT128()
```
fbusato, line 83: confused. Why we need `_CCCL_HAS_INT128()` with `long long`?

miscco: because you need the difference type to be larger than your integer type

miscco: It needs to hold `LONG_LONG_MIN - LONG_LONG_MAX`

libcudacxx/include/cuda/__iterator/counting_iterator.h
```diff
@@ -79,13 +80,21 @@ struct __get_wider_signed
     {
       return ::cuda::std::type_identity<long>{};
     }
-    else
+#if _CCCL_HAS_INT128()
+    else if constexpr (sizeof(_Int) < sizeof(long long))
     {
       return ::cuda::std::type_identity<long long>{};
```
davebayer, line 86: These can be just `(long long){}`, do you want to change this?

miscco: I believe this might be a good first issue

PR #6488, opened 2025-11-04
Implement `cuda::invalid_stream`

davebayer:
In #6416, @pciolkosz tried to make `stream` optional in `cuda::buffer`. After an internal discussion, we decided that it won't be the case and we will have a `cuda::invalid_stream` instead, which can be compared with a `cuda::stream_ref` to check whether the stream is a valid stream or not.

It was not decided what type should the `invalid_stream` have. I decided to go for a separate type for these reasons:
1. `cuda::stream_ref` cannot be a global variable accessible from device.
2. I don't think that user should be able to easily pass `invalid_stream` as a parameter taking `cuda::stream_ref`.

Also, I don't think we should introduce a `stream_ref::is_valid()` method, because the API could suggest that the operation actually checks the validity of the stream, not only comparing it to an invalid stream.

So, I basically implemented a custom `cuda::invalid_stream_t` type that can be compared with `cuda::stream_ref` for (in)equality. Also I added an explicit constructor from `cuda::invalid_stream_t` to `cuda::stream_ref` in case it's desired.

### DIFF
--- a/libcudacxx/include/cuda/__stream/invalid_stream.h
+++ b/libcudacxx/include/cuda/__stream/invalid_stream.h
@@ -4,12 +4,12 @@
 // under the Apache License v2.0 with LLVM Exceptions.
 // See https://llvm.org/LICENSE.txt for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
-// SPDX-FileCopyrightText: Copyright (c) 2023 NVIDIA CORPORATION & AFFILIATES.
+// SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES.
 //
 //===----------------------------------------------------------------------===//
 
-#ifndef _CUDA_STREAM
-#define _CUDA_STREAM
+#ifndef _CUDA___STREAM_INVALID_STREAM_H
+#define _CUDA___STREAM_INVALID_STREAM_H
 
 #include <cuda/std/detail/__config>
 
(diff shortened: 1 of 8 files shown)


libcudacxx/include/cuda/__stream/stream_ref.h
```diff
@@ -294,6 +294,9 @@ _CCCL_HOST_API inline __ensure_current_context::__ensure_current_context(stream_
   ::cuda::__driver::__ctxPush(__ctx);
 }
 
+//! @brief Invalid stream instance.
+inline const cuda::stream_ref invalid_stream{__detail::__invalid_stream};
```
miscco: Nitpick: The stream_ref constructor is constexpr, so this should be an inline variable and we technically allow stream ref to be passed to device, so this should probably be a `_CCCL_GLOBAL_CONSTANT`

miscco (changes requested):

libcudacxx/include/cuda/__stream/stream_ref.h
davebayer: It's not that simple. Yes, the constructor is `constexpr`, but I have no way of doing `(cudaStream_t)(~0ull)` in a constant expression.

We could have a workaround that `stream_ref` would store a `uintptr_t` and the `get()` method would be non-constexpr casting the value to the pointer.

miscco: urgh that would be right

libcudacxx/include/cuda/__stream/invalid_stream.h
```diff
@@ -0,0 +1,47 @@
...
+
+enum class invalid_stream_t : unsigned char
+{
+};
+
+_CCCL_GLOBAL_CONSTANT invalid_stream_t invalid_stream{};
+
+[[nodiscard]] _CCCL_API _CCCL_FORCEINLINE ::cudaStream_t __invalid_stream() noexcept
+{
+  return reinterpret_cast<::cudaStream_t>(~0ull);
```
bdice, line 38: Is `cudaStream_t` a pointer type? Is it possible to use a null pointer as the sentinel value here?

bdice (commented):
I got a notification on this that might have been a mistake (I see I was unassigned immediately after). However, I already took a look over the PR so I decided to share feedback anyway.

I have no context on why this is desired. Can you write a PR description and/or link to a descriptive issue? I do like to watch CCCL development and keep up to speed on what is coming soon. Having some visibility into the motivation and requirements helps a lot.

libcudacxx/include/cuda/__stream/invalid_stream.h
davebayer: It is, but `NULL` is a valid (default) stream, so the driver considers `0xff...ff` pointer value to be the invalid stream handle.

davebayer:
> I got a notification on this that might have been a mistake (I see I was unassigned immediately after). However, I already took a look over the PR so I decided to share feedback anyway.

Thanks for giving a look! It probably happend due to me force pushing the branch.

> I have no context on why this is desired. Can you write a PR description and/or link to a descriptive issue?

I updated the description.

PR #6489, opened 2025-11-05
[cudax->libcu++] Move uninitialized_async_buffer and heterogeneous_iterator to libcu++

pciolkosz:
This PR move some of the `async_buffer` dependencies from cudax to libcu++

We don't plan to expose the `uninitialized_async_buffer` to the users, so I added `__` to it

### DIFF
--- a/cudax/test/CMakeLists.txt
+++ b/cudax/test/CMakeLists.txt
@@ -114,7 +114,6 @@ foreach(cudax_target IN LISTS cudax_TARGETS)
 
   cudax_add_catch2_test(test_target containers ${cudax_target}
     containers/uninitialized_buffer.cu
-    containers/uninitialized_async_buffer.cu
     containers/async_buffer/access.cu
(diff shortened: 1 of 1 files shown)


libcudacxx/include/cuda/__container/heterogeneous_iterator.h
davebayer, line 43: We should simplify this to:
```suggestion
enum class _IsConstIter : bool
{};
```

davebayer (approved):

libcudacxx/include/cuda/__container/heterogeneous_iterator.h
miscco: We should not do any significant code changes in a PR that is supposed to move things. 

We should rather do that as a followup

davebayer: Yeah, That's why I approved it

PR #6493, opened 2025-11-05
Use `__byte_perm` intrinsic rather then inline asm in `cuda::std::byteswap`

davebayer:


miscco (approved):
Verified that the intrinsic is available in CTK 12.0

PR #5535, opened 2025-08-13
[DRAFT] Shared Memory Span

fbusato:
## Description

This is a follow-up of internal `#467`.
The PR proposes a modified `cuda::std::span` to efficiently handle static and dynamic shared memory.

The motivations come from the limitations of  `cuda::std::span`:

1. It doesn't propagate the memory address space.
2. It is based on 64-bit pointers, which are less efficient than 32-bit used by shared memory.
3. Memory footprint is 128-bit, while a shared memory span requires 64-bit of local memory.

The PR is a **DRAFT** to collect feedback.

PR #6495, opened 2025-11-05
Improve `cudax::dynamic_shared_memory` implementation

davebayer:
Some time ago, I had an idea that we could copy the syntax of `new` or `std::make_unique` when creating the dynamic shared memory launch option and do:
```cpp
auto opt1 = cuda::dynamic_shared_memory<int>();
auto opt2 = cuda::dynamic_shared_memory<int[]>();
auto opt3 = cuda::dynamic_shared_memory<int[10]>();
```
In dependance of the type passed to the function, a `dynamic_shared_memory_option::view_type` type alias is set. The mappings are:
```cpp
T    -> T&
T[]  -> cuda::std::span<T>
T[N] -> cuda::std::span<T, N>
```
The view type is then returned from the `cuda::device::dynamic_shared_memory_view(config)` call on device.

I was also considering supporting multidimensional arrays as `T[N0][N1]` or allowing `cuda::std::extents` to be passed to the factory function resulting in a `cuda::std::mdspan` type, but I don't think it will be used.

### DIFF
--- a/cudax/include/cuda/experimental/__launch/configuration.cuh
+++ b/cudax/include/cuda/experimental/__launch/configuration.cuh
@@ -12,6 +12,13 @@
 #define _CUDAX__LAUNCH_CONFIGURATION_CUH
 
 #include <cuda/__driver/driver_api.h>
+#include <cuda/__numeric/overflow_cast.h>
+#include <cuda/__ptx/instructions/get_sreg.h>
+#include <cuda/std/__cstddef/types.h>
+#include <cuda/std/__type_traits/is_const.h>
+#include <cuda/std/__type_traits/is_reference.h>
+#include <cuda/std/__type_traits/is_unbounded_array.h>
+#include <cuda/std/__type_traits/rank.h>
 #include <cuda/std/span>
 #include <cuda/std/tuple>
 
@@ -32,17 +39,7 @@ struct launch_option
 {
   static constexpr bool needs_attribute_space = false;
   static constexpr bool is_relevant_on_device = false;
-
-protected:
-  [[nodiscard]] cudaError_t apply(CUlaunchConfig&, CUfunction) const noexcept
-  {
-    return cudaSuccess;
-  }
 };
-
-template <typename Dimensions, typename... Options>
-cudaError_t apply_kernel_config(
-  const kernel_config<Dimensions, Options...>& config, CUlaunchConfig& cuda_config, CUfunction kernel) noexcept;
 
 // Might need to go to the main namespace?
 enum class launch_option_kind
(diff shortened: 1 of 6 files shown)


cudax/test/launch/dynamic_shared_memory.cu
```diff
@@ -0,0 +1,107 @@
...
+  {
+    static_assert(cuda::std::is_same_v<View, decltype(cudax::device::dynamic_shared_memory_view(config))>);
+    static_assert(noexcept(cudax::device::dynamic_shared_memory_view(config)));
+
+    write_smem(cudax::device::dynamic_shared_memory_view(config));
+  }
+
+  __device__ void write_smem(cuda::std::reference_wrapper<T> view)
+  {
+    view.get() = T{};
```
davebayer: I am not quite sure about the `reference_wrapper`, this syntax is just sooo awkward

cudax/include/cuda/experimental/__launch/configuration.cuh
```diff
@@ -197,18 +262,19 @@ private:
  * @tparam NonPortableSize
  *  Needs to be enabled to exceed the portable limit of 48kB of shared memory per block
  */
-template <typename Content, std::size_t Extent = 1, bool NonPortableSize = false>
-struct dynamic_shared_memory_option : public __detail::launch_option
+template <class _Tp>
```
pciolkosz, line 229: Why remove the `NonPortableSize`? I will need to make a follow up change to make it an enum, but I would still keep the bool version for now

davebayer: Yeah I will bring it back. I just think that it needn't to be a template parameter, but should be an ordinary member. The value is only being used before launch anyway

cudax/test/launch/dynamic_shared_memory.cu
davebayer: @pciolkosz I found out I am strongly against using `reference_wrapper` here.
1. I believe when you write a kernel, you always know if you are expecting and array or a single object in the dynamic shared memory.
2. This simple example I expect everyone to use 99 % of the time doesn't work, because it will try to get the reference to the `reference_wrapper` instead of the underlying dynamic smem. You need to do `.get()` which I think is very counterintuitive.
```cpp
auto& dyn_smem = cuda::dynamic_shared_memory_view(config);
```

I would prefer to return a plain reference in this case.

pciolkosz: I am ok with that

cudax/include/cuda/experimental/__launch/configuration.cuh
```diff
@@ -197,96 +224,129 @@ private:
...
+  _CCCL_TEMPLATE(class _Tp2 = _Tp)
+  _CCCL_REQUIRES((!::cuda::std::is_unbounded_array_v<_Tp2>) )
+  _CCCL_HOST_API constexpr dynamic_shared_memory(non_portable_t) noexcept {}
+
+  _CCCL_TEMPLATE(class _Tp2 = _Tp)
+  _CCCL_REQUIRES(::cuda::std::is_unbounded_array_v<_Tp2>)
+  _CCCL_HOST_API constexpr dynamic_shared_memory(::cuda::std::size_t __n)
+      : __base_type{__n}
+  {
+    if (__n * sizeof(value_type) > __max_portable_dyn_smem_size)
```
pciolkosz, line 275: This changes the semantics a little bit, because before if the shared memory limit function attribute was set correctly we would allow non-portable size to go through without an explicit opt-in. That said, I think it's fine to always require the non-portable to be specified, even if the function attribute was set. There is little downside to provide the argument and I want people to move away from function attributes. I would only expand the throw message to include the name of the option in, like `"It is required to pass non_portable to opt-in to dynamic shared memory exceeding the portable limit"

pciolkosz: Ahh, I just realized this approach won't work. The 48 kB portable limit applies to static + dynamic shared. So the initial value of the shared memory limit function attribute is 48 kB - size of static shared: https://godbolt.org/z/8cEWfh4WG
This means there is no way to check if dynamic shared size is portable without knowing the kernel, which happens after the option is created. I think we need to note somewhere if non-portable size was opted-into, but not necessarily as a template argument like it was before

fbusato:
I'm a bit confused how this could interact with a shared memory-specific span, see https://github.com/NVIDIA/cccl/pull/5535

pciolkosz:
We could return `shared_mem_span` from `dynamic_shared_memory_view` function. But we would need to finalize the decision if there is enough motivation behind `shared_mem_span` and finalize the implementation for 3.2 release. I don't think we want to slip `cuda::launch` to the next release for it

fbusato:
I don't see big problems for targeting CCCL 3.2. It is not a huge task

cudax/include/cuda/experimental/__launch/configuration.cuh
```diff
@@ -197,96 +226,151 @@ private:
...
     {
-      // TODO since 12.6 there is a per launch option available, we should switch once compatibility is not an issue
-      // TODO should we validate the max amount with device props or just pass it through and rely on driver error?
-      status = ::cuda::__driver::__functionSetAttributeNoThrow(
-        kernel, CU_FUNC_ATTRIBUTE_MAX_DYNAMIC_SHARED_SIZE_BYTES, size_needed);
-      if (status != cudaSuccess)
+      if (!::cuda::std::__cccl_default_is_constant_evaluated())
       {
-        return status;
+        NV_IF_TARGET(NV_IS_DEVICE, (return ::cuda::ptx::get_sreg_dynamic_smem_size();))
```
pciolkosz, line 295: I am not sure if it's better to query the special register, but lets leave it as is

PR #6496, opened 2025-11-05
Expose `ptx::mbarrier_inval` and use it

bernhardmgruber:


cub/cub/block/block_load_to_shared.cuh
```diff
@@ -248,12 +248,7 @@ public:
     __syncthreads();
     if (elected)
     {
-      NV_IF_TARGET(NV_PROVIDES_SM_90,
-                   (
-                     // Borrowed from cuda::barrier
-                     // TODO Make this available through cuda::ptx::
-                     asm volatile("mbarrier.inval.shared.b64 [%0];" ::"r"(static_cast<::cuda::std::uint32_t>(
-                       ::__cvta_generic_to_shared(&temp_storage.mbarrier_handle))) : "memory");));
+      NV_IF_TARGET(NV_PROVIDES_SM_90, ::cuda::ptx::mbarrier_inval(&temp_storage.mbarrier_handle););
```
miscco, line 251: Why is this SM_90 when we support it from SM_80 onwards?

bernhardmgruber: Because here, only the SM90 code path used a barrier for async copy synchronization. The SM80 code path uses a commit group and thus no barrier.

PR #6499, opened 2025-11-05
Allow passing in `None`  as init value for scan when using an iterator as input

### LINKED ISSUE #6498: [FEA]: Allow passing in an iterator as `d_in` with `None` as init value in scan
### Is this a duplicate?

### Area

cuda.compute (Python)

Currently, using an iterator with `None` as the initial value is not supported. 

We should just get the datatype of the initial value from the `d_in` iterator's `value_type`

_No response_

### Additional context

_No response_

NaderAlAwar:
## Description

closes #6498

## Checklist


python/cuda_cccl/cuda/compute/algorithms/_scan.py
```diff
@@ -61,14 +61,18 @@ def __init__(
 
         match self.init_kind:
             case _bindings.InitKind.NO_INIT:
-                # TODO: we just need to extract the dtype from the input iterator
-                if not isinstance(d_in, DeviceArrayLike):
+                self.init_value_cccl = None
+
+                # Extract value type from input (either device array or iterator)
```
shwina: We should use https://github.com/NVIDIA/cccl/blob/bcb2bcb106dac16dee49e46e0fa7df21dae6a24a/python/cuda_cccl/cuda/compute/_cccl_interop.py#L315 here.

PR #4928, opened 2025-06-06
Added f32/fp64 specializations for complex exp function.

s-oboyle:
## Updated the complex exp function to avoid under/overflow issues.

There are a few issues with the bounds in the cexp function relating to under/overflow.

In the current version, it can be the case that the real/imag parts of the answer, which is implemented as:
`(e^real * cos(imag), e^real * sin(imag))`,

can have the `e^real` part overflow, while the next sin/cos multiplication would bring it back into the fp range, however we return `INF`.

It can also generate `NaN` results, when `imag == 0.0` -> `sin(imag) == 0`, and exp overflows.
Here the correct `imag` result is 0, but the naive implementation returns `NaN`.

This new version fixes all these possible under/overflow issues, and through inlining some of the function calls and stripping out unneeded checks does so with minimal perf disruption.

## Perf
For fp64 precision performance was faster on H100 and Quadro RTX 8000.
For fp32 precision performance was only faster on the Quadro RTX 8000.

Using the math-teams standard `math_bench` test we have the following on H100 (averaged summary):
**Operations/SM/cycle:**
| H100 |  old | new |
| --- | --- | ---|
| fp64 | 0.66082 | 0.70204 | 
| fp32 | 1.61528 | 1.53386 | 

(See https://docs.google.com/spreadsheets/d/1TdOpEbLgoL1QOWKjeO3pwFDDMoPr4iqnPjz_XrL7UEA for raw/non-averaged info.)

## Correctness GPU/CPU
The before / after accuracy accuracy is similar on regions where the current function does not have issues.
An intensive bracket and bisect search, along with testing special hard values, gives:
```
GPU fp64:
Max ulp real error (2.98,0.9139) @ (5.217369075,0.849946261)    (0x4014de95ffae7653,0x3feb32c280518909)
        Ours = (121.7401997,138.5658332)    Ref = (121.7401997,138.5658332)
        Ours = (0x405e6f5f6e98d3e1,0x4061521b4e34f7e3)               Ref = (0x405e6f5f6e98d3de,0x4061521b4e34f7e2)

Max ulp imag error (0.666,2.852) @ (8.78730102e-12,3.14673129e+293)     (0x3da352cf8a03ca22,0x7cdf886be6d24ce9)
        Ours = (-0.9685123725,-0.2489654278)    Ref = (-0.9685123725,-0.2489654278)
[...]

PR #6502, opened 2025-11-05
Catchs NaN's before they make it to static_cast and creating UB

s-oboyle:
Fix UB detailed here:
https://github.com/NVIDIA/cccl/pull/4928#issuecomment-3488235048

No perf changes.


s-oboyle:
/ok to test [17b47ab](https://github.com/NVIDIA/cccl/pull/6502/commits/17b47ab4fe3683cd3bcfde9f601e178b0b1f03d4)

Artem-B:
Thank you for the fix!
I can confirm that the change fixes the reported UB instance in my build.

PR #6518, opened 2025-11-05
Allow using ZipIterator as an output in cuda.compute

### LINKED ISSUE #6505: [FEA]: Enable using ZipIterator as an output iterator in cuda.compute
### Is this a duplicate?

### Area

cuda.compute (Python)

We should be able to use ZipIterator as an output

We could either use ZipIterator or create a new ZipOutputIterator like we do for transform https://github.com/NVIDIA/cccl/pull/5934#issue-3431367833, although that was likely due to differing type annotations.

_No response_

### Additional context

_No response_

NaderAlAwar:
## Description

closes #6505

credit to @shwina for the implicit conversion between from tuple to ZipValue in the return of the operator

## Checklist


python/cuda_cccl/cuda/compute/iterators/_zip_iterator.py
```diff
@@ -194,6 +242,18 @@ def advance(self):
...
+        def output_dereference(self):
+            if not all_support_output:
+                raise AttributeError(
+                    "ZipIterator cannot be used as output iterator when not all "
+                    "underlying iterators support output"
+                )
+            return output_dereference_func
+
+        # Make the ZipValue type accessible for constructing return values
+        value_struct_type = ZipValue
```
shwina: Is this needed even after adding implicit tuple conversion?

python/cuda_cccl/tests/compute/test_zip_iterator.py
```diff
@@ -239,3 +240,49 @@ def min_pairs(p1, p2):
...
+    zip_it = ZipIterator(d_in1, d_in2)
+
+    d_out1 = cp.empty_like(d_in1)
+    d_out2 = cp.empty_like(d_in2)
+
+    zip_out_it = ZipIterator(d_out1, d_out2)
+
+    def add_pairs(p1, p2):
+        a, b = tuple(p1)
+        c, d = tuple(p2)
```
shwina: This part seems a bit icky. For now, can I suggest just:

```python
def add_pairs(p1, p2):
    return p1[0] + p2[0], p2[1] + p2[1]
```

We should figure out how to make straight unpacking work:

```python
a, b = p1
c, d = p2
```

Can we also then get rid of the convesion-TO-tuple in this PR? Let's keep the conversion-FROM-tuple (which is what allows the `return` to work).

python/cuda_cccl/cuda/compute/iterators/_zip_iterator.py
NaderAlAwar: no, removed

python/cuda_cccl/tests/compute/test_zip_iterator.py
NaderAlAwar: Deleted the explicit `tuple` construction from this PR, will create an issue

NaderAlAwar: Tracked in #6610

PR #6519, opened 2025-11-05
[DO NOT MERGE] Fix Host `isnan`, `isinf`, etc.  handling in some `math.h` implementations

fbusato:
## Description

`isnan`, `isinf` and other utilities in  `math.h`  can be implemented as a macro or a function, depending on the implementation. 
Most x86 implementations provide it as a function. Some Arm64 implementations provide it as a macro.

This PR adds the host fallback with `isnan` instead of `::isnan` when the utility is implemented as a macro.

### DIFF
--- a/libcudacxx/include/cuda/std/__cmath/isfinite.h
+++ b/libcudacxx/include/cuda/std/__cmath/isfinite.h
@@ -46,11 +46,7 @@ template <class _Tp>
   static_assert(is_floating_point_v<_Tp>, "Only standard floating-point types are supported");
   if (!::cuda::std::__cccl_default_is_constant_evaluated())
   {
-#if defined(isfinite)
-    NV_IF_TARGET(NV_IS_DEVICE, (return ::isfinite(__x);), (return isfinite(__x);));
-#else // ^^^ macro ^^^ / vvv function vvv
-    NV_IF_TARGET(NV_IS_DEVICE, (return ::isfinite(__x);), (return ::isfinite(__x);));
-#endif // function
+    return isfinite(__x);
   }
   return !::cuda::std::isnan(__x) && !::cuda::std::isinf(__x);
 }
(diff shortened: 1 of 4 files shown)


libcudacxx/include/cuda/std/__cmath/isinf.h
```diff
@@ -70,7 +74,7 @@ template <class _Tp>
 #elif _CCCL_HAS_CONSTEXPR_BIT_CAST()
   if (!::cuda::std::__cccl_default_is_constant_evaluated())
   {
-    return ::isinf(__x);
+    return ::cuda::std::__isinf_impl(__x);
```
miscco: We really want the math library function here

libcudacxx/include/cuda/std/__cmath/isfinite.h
```diff
@@ -58,7 +62,7 @@ template <class _Tp>
 #else // ^^^ _CCCL_BUILTIN_ISFINITE ^^^ / vvv !_CCCL_BUILTIN_ISFINITE vvv
   if (!::cuda::std::__cccl_default_is_constant_evaluated())
   {
-    return ::isfinite(__x);
+    return ::cuda::std::__isfinite_impl(__x);
```
miscco: Ditto we explicitly want the host function here

davebayer:
@miscco can't we just include `<cmath>` and use `::std::meow` functions instead?

libcudacxx/include/cuda/std/__cmath/isfinite.h
```diff
@@ -46,7 +46,11 @@ template <class _Tp>
   static_assert(is_floating_point_v<_Tp>, "Only standard floating-point types are supported");
   if (!::cuda::std::__cccl_default_is_constant_evaluated())
   {
-    return ::isfinite(__x);
+#if defined(isfinite)
+    NV_IF_TARGET(NV_IS_DEVICE, (return ::isfinite(__x);), (return isfinite(__x);));
```
davebayer: Are you sure the macro is not defined in device code? Because then this will not work on device, we would have to push/pop the macro

miscco:
> @miscco can't we just include `<cmath>` and use `::std::meow` functions instead?

That would potentially add considerable code that we do not need

davebayer (changes requested):
Can you please confirm the macros are not defined in device code?

libcudacxx/include/cuda/std/__cmath/isfinite.h
fbusato: the situation is even more complex. `crt/math_functions.h` imported by nvcc uses `#undef isnan`.
In addition, the user that reported the bug is compiling a `.cpp` file.

PR #6522, opened 2025-11-06
Fix some warnings in cub headers that are picked up by the libcu++ tests

miscco:


cub/cub/device/device_reduce.cuh
```diff
@@ -994,14 +994,14 @@ public:
   {
     _CCCL_NVTX_RANGE_SCOPE("cub::DeviceReduce::Min");
 
-    static_assert(!_CUDA_STD_EXEC::__queryable_with<EnvT, _CUDA_EXEC::determinism::__get_determinism_t>,
+    static_assert(!::cuda::std::execution::__queryable_with<EnvT, _CUDA_EXEC::determinism::__get_determinism_t>,
```
miscco: Drive by fix, because we dropped the macros

cub/cub/device/device_reduce.cuh
```diff
@@ -1281,30 +1283,32 @@ public:
...
+                                                  _CUDA_EXEC::determinism::__get_determinism_t,
+                                                  _CUDA_EXEC::determinism::run_to_run_t>;
 
     // Static assert to reject gpu_to_gpu determinism since it's not properly implemented
     static_assert(!::cuda::std::is_same_v<requested_determinism_t, ::cuda::execution::determinism::gpu_to_gpu_t>,
                   "gpu_to_gpu determinism is not supported");
 
     // TODO(NaderAlAwar): Relax this once non-deterministic implementation for min / max is available
-    using determinism_t = ::cuda::execution::determinism::run_to_run_t;
+    // using determinism_t = ::cuda::execution::determinism::run_to_run_t;
```
miscco: The tunings arent used here, I disabled them

miscco: @NaderAlAwar can you please give this a look, because that is your code

bernhardmgruber (approved):

cub/cub/device/device_reduce.cuh
NaderAlAwar: I think @gonidelis wrote this code. I'm actually not sure why we don't pass the tuning policy to the dispatch class.

cub/cub/device/dispatch/kernels/kernel_reduce.cuh
```diff
@@ -335,12 +335,12 @@ __launch_bounds__(int(ChainedPolicyT::ActivePolicy::ReducePolicy::BLOCK_THREADS)
...
 
   _CCCL_PRAGMA_UNROLL_FULL()
-  for (::cuda::std::uint32_t i = tid; i < num_items; i += (n_threads * items_per_thread))
+  for (int i = tid; i < num_items; i += (n_threads * items_per_thread))
   {
     ftype items[items_per_thread] = {};
     for (int j = 0; j < items_per_thread; j++)
     {
-      const ::cuda::std::uint32_t idx = i + j * n_threads;
+      const int idx = i + j * n_threads;
```
NaderAlAwar: @srinivasyadav18 you should take a look at this

miscco: Oh that is totally fine, all inputs are int, so lets just roll with that

srinivasyadav18: The change looks fine to me.

davebayer (approved):

cub/cub/device/dispatch/kernels/kernel_reduce.cuh
```diff
@@ -335,12 +335,12 @@ __launch_bounds__(int(ChainedPolicyT::ActivePolicy::ReducePolicy::BLOCK_THREADS)
   int n_threads = reduce_grid_size * block_threads;
 
   _CCCL_PRAGMA_UNROLL_FULL()
-  for (::cuda::std::uint32_t i = tid; i < num_items; i += (n_threads * items_per_thread))
+  for (int i = tid; i < num_items; i += (n_threads * items_per_thread))
```
fbusato: weird that we used `uint32_t` in the original version

fbusato (approved):

miscco:
This consistently hangs `cub.cpp20.test.device_reduce_deterministic.lid_0`

cub/cub/device/dispatch/kernels/kernel_reduce.cuh
miscco:
```suggestion
  for (::cuda::std::uint32_t i = static_cast<::cuda::std::uint32_t>(tid); i < static_cast<::cuda::std::uint32_t>(num_items); i += static_cast<::cuda::std::uint32_t>(n_threads * items_per_thread))
```

Check sass

fbusato: ~I don't think this is the right solution~

fbusato: discussed offline with @gevtushenko. There could be alternatives to the proposed approach but definitely for another PR

bernhardmgruber: > Check sass

Is that a command? To whom? Or do you mean you checked the SASS? Then what executable and for which SM arch?

gonidelis: I think he just wants to make sure sass for `cub.cpp20.test.device_reduce_deterministic.lid_0` doesn't change after this change @bernhardmgruber

bernhardmgruber: I think the `i += (n_threads * items_per_thread)` can overflow if `num_items` is very close to `INT_MAX`. I think it would be safer to stay in the realm of `u32`.

bernhardmgruber: I confirm that the current state of the PR loops infinitely.

cub/cub/device/dispatch/kernels/kernel_reduce.cuh
bernhardmgruber: Unfortunately, the value of `idx` goes beyond `INT_MAX`, so using `int` overflows here.

bernhardmgruber:
I dropped the commented out code and rebased on `main`.

There are no SASS differences for `cub.test.device_reduce_deterministic.lid_0` on sm120.

bernhardmgruber (approved):

PR #6524, opened 2025-11-06
Split `DeviceSegmentedReduce` in its own file

bernhardmgruber:
The file was a bit big and the segmented reduce dispatch is logically independent.

### DIFF
--- a/cub/cub/device/dispatch/dispatch_reduce.cuh
+++ b/cub/cub/device/dispatch/dispatch_reduce.cuh
@@ -34,7 +34,7 @@
 #include <cuda/std/cstdint>
 
 // TODO(bgruber): included to not break users when moving DeviceSegmentedReduce to its own file. Remove in CCCL 4.0.
-#include <cub/device/dispatch/dispatch_reduce.cuh>
+#include <cub/device/dispatch/dispatch_segmented_reduce.cuh>
 
 CUB_NAMESPACE_BEGIN
 


cub/cub/device/dispatch/dispatch_segmented_reduce.cuh
```diff
@@ -0,0 +1,730 @@
+// SPDX-FileCopyrightText: Copyright (c) 2011, Duane Merrill. All rights reserved.
+// SPDX-FileCopyrightText: Copyright (c) 2011-2024, NVIDIA CORPORATION. All rights reserved.
+// SPDX-License-Identifier: BSD-3
```
bernhardmgruber, line 3: License retained from file this was copied from.

cub/cub/device/dispatch/dispatch_reduce.cuh
```diff
@@ -21,26 +21,20 @@
...
-#include <cuda/std/__algorithm/min.h>
 #include <cuda/std/__functional/identity.h>
 #include <cuda/std/__functional/invoke.h>
-#include <cuda/std/__iterator/iterator_traits.h>
-#include <cuda/std/__utility/pair.h>
 #include <cuda/std/cstdint>
-#include <cuda/std/limits>
+
+// TODO(bgruber): included to not break users when moving DeviceSegmentedReduce to its own file. Remove in CCCL 4.0.
+#include <cub/device/dispatch/dispatch_reduce.cuh>
```
miscco: This should be 
```suggestion
#include <cub/device/dispatch/dispatch_segmented_reduce.cuh>
```

bernhardmgruber: What would I do without you! Thx!

PR #6527, opened 2025-11-06
Replace `__throw_meow` by `_CCCL_THROW(meow)`

davebayer:
Since we have the `_CCCL_THROW` macro, we can remove the `__thow_meow` functions which only conditionally throw if exceptions are enabled.

This change has a very important consequence: if we `throw` in a function that is marked as `noexcept`, we get a compiler warning.

### DIFF
--- a/cudax/include/cuda/experimental/__launch/launch.cuh
+++ b/cudax/include/cuda/experimental/__launch/launch.cuh
@@ -121,7 +121,7 @@ _CCCL_HOST_API auto __launch_impl(_Dst&& __dst, _Config __conf, ::CUfunction __k
   ::cudaError_t __status = cuda::__detail::apply_kernel_config(__conf, __config, __kernel);
   if (__status != ::cudaSuccess)
   {
-    _CCCL_THROW(cuda::cuda_error, __status, "Failed to prepare a launch configuration");
+    _CCCL_THROW(::cuda::cuda_error, __status, "Failed to prepare a launch configuration");
   }
 
   __config.gridDimX  = block.dims(grid, __conf).x;
--- a/libcudacxx/include/cuda/__algorithm/copy.h
+++ b/libcudacxx/include/cuda/__algorithm/copy.h
@@ -76,7 +76,7 @@ _CCCL_HOST_API void __copy_bytes_impl(
 
   if (__src.size_bytes() > __dst.size_bytes())
   {
-    _CCCL_THROW(std::invalid_argument, "Copy destination is too small to fit the source data");
+    _CCCL_THROW(::std::invalid_argument, "Copy destination is too small to fit the source data");
   }
   if (__src.size_bytes() == 0)
   {
@@ -121,12 +121,12 @@ _CCCL_HOST_API void __copy_bytes_impl(
   // Check only destination, because the layout of destination is the same as source
   if (!__dst.is_exhaustive())
   {
-    _CCCL_THROW(std::invalid_argument, "copy_bytes supports only exhaustive mdspans");
+    _CCCL_THROW(::std::invalid_argument, "copy_bytes supports only exhaustive mdspans");
   }
 
   if (__src.extents() != __dst.extents())
   {
-    _CCCL_THROW(std::invalid_argument, "Copy destination size differs from the source");
+    _CCCL_THROW(::std::invalid_argument, "Copy destination size differs from the source");
   }
 
   ::cuda::__detail::__copy_bytes_impl(
--- a/libcudacxx/include/cuda/__algorithm/fill.h
+++ b/libcudacxx/include/cuda/__algorithm/fill.h
@@ -55,7 +55,7 @@ _CCCL_HOST_API void __fill_bytes_impl(stream_ref __stream,
   // Check if the mdspan is exhaustive
   if (!__dst.is_exhaustive())
   {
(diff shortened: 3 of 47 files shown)


libcudacxx/include/cuda/__algorithm/copy.h
```diff
@@ -26,9 +26,12 @@
 #  include <cuda/__algorithm/common.h>
 #  include <cuda/__stream/stream_ref.h>
 #  include <cuda/std/__concepts/concept_macros.h>
+#  include <cuda/std/__exception/exception_macros.h>
 #  include <cuda/std/mdspan>
 #  include <cuda/std/span>
 
+#  include <stdexcept>
```
miscco: This is a bug, we cannot include that file with nvrtc. This needs to be handled centrally in `<cuda/std/__exception/exception_macros.h>`

miscco (changes requested):

libcudacxx/include/cuda/__algorithm/copy.h
davebayer: The contents are guarded by `#if _CCCL_HAS_CTK() && !_CCCL_COMPILER(NVRTC)`

libcudacxx/include/cuda/__barrier/barrier_block_scope.h
```diff
@@ -44,6 +44,10 @@
 #include <cuda/std/__new/device_new.h>
 #include <cuda/std/cstdint>
 
+#if !_CCCL_COMPILER(NVRTC)
+#  include <new>
+#endif // !_CCCL_COMPILER(NVRTC)
```
miscco: Didnt we want to move that somewhere else?

libcudacxx/include/cuda/std/__pstl/cuda/for_each_n.h
```diff
@@ -60,7 +60,7 @@ struct __pstl_dispatch<__pstl_algorithm::__for_each_n, __execution_backend::__cu
 
     auto __stream = ::cuda::__call_or(::cuda::get_stream, ::cuda::stream_ref{cudaStreamPerThread}, __policy);
 
-    _CCCL_TRY_CUDA_API(
+    _CCCL_TERMINATE_CUDA_API(
```
fbusato: the name is not very intuitive.

libcudacxx/include/cuda/__runtime/api_wrapper.h
```diff
@@ -45,6 +47,20 @@
     _CCCL_ASSERT(__status == cudaSuccess, _MSG);                        \
   } while (0)
 
+#define _CCCL_TERMINATE_CUDA_API(_NAME, _MSG, ...)                                 \
```
fbusato: the macro name is not very clear. Would be better to find a more intuitive name. `_CCCL_REQUIRE_CUDA_API` or `_CCCL_CHECK_CUDA_API` or `_CCCL_CUDA_API_FATAL`, etc.

fbusato (approved):

libcudacxx/include/cuda/__runtime/api_wrapper.h
davebayer: What about `_CCCL_ABORT_CUDA_API`? I don't like that the alternatives that you proposed don't actually say what it would do.

fbusato: `TERMINATE` better than `ABORT` because it uses `cuda::std::terminate`. We could keep `_CCCL_TERMINATE_CUDA_API`.
My point is that the API name seems to say "terminate the program" instead of "terminate if the API fails".

davebayer: I got rid of it completely

cudax/include/cuda/experimental/__launch/launch.cuh
```diff
@@ -120,7 +121,7 @@ _CCCL_HOST_API auto __launch_impl(_Dst&& __dst, _Config __conf, ::CUfunction __k
   ::cudaError_t __status = cuda::__detail::apply_kernel_config(__conf, __config, __kernel);
   if (__status != ::cudaSuccess)
   {
-    ::cuda::__throw_cuda_error(__status, "Failed to prepare a launch configuration");
+    _CCCL_THROW(cuda::cuda_error, __status, "Failed to prepare a launch configuration");
```
miscco: Shouldnt all those be 
```suggestion
    _CCCL_THROW(::cuda::cuda_error, __status, "Failed to prepare a launch configuration");
```

libcudacxx/include/cuda/__algorithm/fill.h
```diff
@@ -53,7 +55,7 @@ _CCCL_HOST_API void __fill_bytes_impl(stream_ref __stream,
   // Check if the mdspan is exhaustive
   if (!__dst.is_exhaustive())
   {
-    ::cuda::std::__throw_invalid_argument("fill_bytes supports only exhaustive mdspans");
+    _CCCL_THROW(std::invalid_argument, "fill_bytes supports only exhaustive mdspans");
```
miscco: We should also use `::std::` always

miscco (commented):
As discussed internally, we should not hide the global qualification inside the macro. 

It is *really* important and we should always use it

PR #6529, opened 2025-11-06
Implement a way to provide a memory_resource through the execution policy

miscco:


libcudacxx/test/libcudacxx/cuda/execution/execution_policy/get_stream.pass.cpp
```diff
@@ -69,6 +61,9 @@ void test()
   static_assert(!execution::__queryable_with<execution::unsequenced_policy, ::cuda::get_stream_t>);
 
   test(cuda::execution::__cub_par_unseq);
+
+  // Ensure that all works even if we have a memory resource
+  test(cuda::execution::__cub_par_unseq.set_memory_resource(::cuda::device_default_memory_pool(::cuda::device_ref{0})));
```
bernhardmgruber, line 66: Suggestion: I would intuitively name this:
```suggestion
  test(cuda::execution::__cub_par_unseq.with_memory_resource(::cuda::device_default_memory_pool(::cuda::device_ref{0})));
```

libcudacxx/include/cuda/std/__pstl/cuda/reduce.h
```diff
@@ -61,15 +62,9 @@ struct __pstl_dispatch<__pstl_algorithm::__reduce, __execution_backend::__cuda>
...
     auto __stream         = __policy.query(::cuda::get_stream);
-
-    _CCCL_TRY_CUDA_API(
-      ::cudaMallocAsync,
-      "__pstl_dispatch: allocation failed",
-      reinterpret_cast<void**>(&__device_ret_ptr),
-      sizeof(_Tp),
-      __stream.get());
+    auto __resource       = __policy.query(::cuda::mr::get_memory_resource);
+    _Tp* __device_ret_ptr = static_cast<_Tp*>(__resource.allocate(__stream, sizeof(_Tp), alignof(_Tp)));
```
bernhardmgruber, line 67: Important: we should really use an asynchronous, stream ordered allocation here.

libcudacxx/include/cuda/std/__pstl/cuda/reduce.h
```diff
@@ -93,7 +88,7 @@ struct __pstl_dispatch<__pstl_algorithm::__reduce, __execution_backend::__cuda>
       ::cudaMemcpyDeviceToHost,
       __stream.get());
 
-    _CCCL_TRY_CUDA_API(::cudaFreeAsync, "__pstl_cuda_for_each: deallocate failed", __device_ret_ptr, __stream.get());
+    __resource.deallocate(__stream, __device_ret_ptr, sizeof(_Tp), alignof(_Tp));
```
bernhardmgruber, line 91: Important: should be async free

libcudacxx/test/libcudacxx/cuda/execution/execution_policy/get_stream.pass.cpp
miscco: This is a more general comment, because we need to be consistent. we cannot have `set_stream` and `with_memory_resource`. We need to settle on one spelling

libcudacxx/include/cuda/std/__pstl/cuda/reduce.h
miscco: The memory resource concept changed, now the default is asynchronous and we have `allocate_sync`

libcudacxx/include/cuda/std/__pstl/cuda/reduce.h
miscco: ditto that is the async version 🙈

libcudacxx/include/cuda/std/__pstl/cuda/reduce.h
bernhardmgruber: alright. not obvious :D

libcudacxx/test/libcudacxx/cuda/execution/execution_policy/get_stream.pass.cpp
bernhardmgruber: I would use `with_xxx` for any kind of attachment. thrust uses `on(stream)`, which is also possible, but then `operator()` for the MR, which I don't think fits a new design.

miscco: yeah, I am totally fine with **any** naming scheme as long as we use it everywhere

PR #6535, opened 2025-11-06
Improvements and testing for inspect_changes CI functionality.

alliepiper:
1. Rewrote `inspect_changes.sh` to python.
2. Split out project name, path, dependency information into new yaml file.
3. Simplified dependency specification (transitive deps are now inferred).
4. Split dependency specification into two types: full (use the pull_request matrix) and lite (use the pull_request_lite matrix).
5. Use new features to split some projects with expensive dependency chains into 'public' (public headers, etc) and `internal` (tests/examples/infra, etc). This allows finer control over dependencies (eg. downstream projects don't need to retest when an upstream project's test files change).
6. Update the ignored paths to include newer additions.
7. Add tests for inspect_changes to make it easier to test and validate modifications.


alliepiper:
@wmaxey The rest is handled during matrix parsing  / workflow building: https://github.com/NVIDIA/cccl/blob/abfdbd67d930eb0af438af6a0a20d4a86c8e1544/.github/actions/workflow-build/build-workflow.py#L1173-L1174

Issue #6528, opened 2025-11-06
[BUG]: Device `cuda::align_up` keeps NVCC from tracking memory space


pauleonix-bak:
### Is this a duplicate?

- [x] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cccl/issues) for this bug and that I agree to the [Code of Conduct](CODE_OF_CONDUCT.md)

### Type of Bug

Performance

### Component

libcu++

### Describe the bug

Basically the same problem as in #4836 but for `cuda::align_[up|down]`. Even though we discussed this problem when these were implemented  in #5037 and #5151 even explicitly documented the absence of this problem,  we still seem to be loosing memory space information.

### How to Reproduce

0. See godbolt link for exact reproducer (Can replace `align_up` with `align_down` in the uppermost source code field).
1. Use the specified functions to align a pointer to shared memory.
2. Access shared memory through that pointer and take a look at the generated SASS. It should contain LDS?STS but does instead use generic load/store instructions (which affect performance)

### Expected behavior

As documented, these function templates should allow the compiler to keep track of memory space information. This can be achieved by only using `uintptr_t` for calculating the offset to be added/subtracted to/from the original pointer instead of casting back from `uintptr_t` to the pointer type as demonstrated in the godbolt reproducer.

### Reproduction link

https://cuda.godbolt.org/z/6oG851G6P

pauleonix-bak:
@fbusato Maybe a commit got lost one the way or similar?

fbusato:
We explicitly check the memory space in the unit test, and we describe the behavior in [docs](https://nvidia.github.io/cccl/libcudacxx/extended_api/memory/align_up.html#libcudacxx-extended-api-memory-align-up). 
I will check if we are missing something.

fbusato:
that's a double failure. `cuda::align_up` loses the memory space AND `__isShared` states that the memory space is shared memory even it is not, or the compiler loses the property. I will open a bug.

pauleonix:
@fbusato 
> `__isShared` states that the memory space is shared memory even it is not, or the compiler loses the property.

Isn't that the expected behavior? It seems like `__isShared` is a runtime check that should also be true for a generic smem pointer. I.e. By design, even when the compiler looses track of the memory space, this `__isShared` should still be able to determine it at runtime. Citing the [docs](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#isshared) (emphasis mine):

[...]

PR #6541, opened 2025-11-07
`cuda::align_up/down` workaround for memory space

fbusato:
Address #6528

Ensure that `align_down` and `align_up` return pointers in the same memory space of the input, while keeping optimal code.
Took me a while, but I was able to find the right pattern that the compiler is able to optimize for both functions.

Kept the opportunity to improve related documentation and unit test.

### DIFF
--- a/docs/libcudacxx/extended_api/memory/align_up.rst
+++ b/docs/libcudacxx/extended_api/memory/align_up.rst
@@ -29,13 +29,15 @@ The function returns the original pointer or closest pointer larger than ``ptr``
 **Performance considerations**
 
 - The function is optimized for compile-time values of ``alignment``.
+- The function does not perform any operations if ``alignment == alignof(T)``.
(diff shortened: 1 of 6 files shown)


docs/libcudacxx/extended_api/memory/align_up.rst
```diff
@@ -29,13 +29,15 @@ The function returns the original pointer or closest pointer larger than ``ptr``
...
+- The function does not perform any operations if ``alignment == alignof(T)``.
+- The function is translated to ``LOP3.LUT`` + ``IADD.64`` instructions for other values of ``alignment``.
 - The returned pointer is decorated with ``__builtin_assume_aligned`` to help the compiler generate better code.
 - The returned pointer maintains the same memory space, for example shared memory, as the input pointer.
 
 Example
 -------
 
-.. code:: cuda
+.. code:: cudas
```
bernhardmgruber, line 40: Q: Is this a typo?

Issue #6368, opened 2025-10-28
[EPIC] Design a public tuning API for CUB


bernhardmgruber:
Background: CUB internally uses tunings to pick parameters for a given workload and GPU architecture. Some users exploit this mechanism and provide their own tunings, attempting to get even better performance than CUB, which is reasonable in case they know more their workload than what CUB was tuned for. However, this either requires calling internal APIs or calling the CUB dispatch layer instead of the device API entry points. This is one reason why we have not moved the dispatch layer into a `detail` namespace.

We should have a well defined public API that our users and we ourselves can use to specify tunings for CUB algoriths.

- [ ] Review existing use cases of our internal tuning APIs
- [ ] Gather requirements for a new public tuning API
- [ ] Design the new tuning API
- [ ] Implement a MVP for a selected CUB algorithm
- [ ] Change/Extend all CUB algorithms to provide access to the new tuning API
- [ ] Replace all internal dispatch layer access by public CUB API calls including the new tuning API
- [ ] Move the CUB dispatch layer into a detail namespace, leaving deprecated aliases in the public (dropped at the next major release)

bernhardmgruber:
All prototyping and design tasks around the tuning API have concluded. Future progress of the CUB-wide roll out is tracked in #7165

PR #6544, opened 2025-11-07
Implement the new tuning API for `DeviceReduce`

bernhardmgruber:
Part of #6368, which was design approved yesterday. The goal is to merge refacttorings like the one here continuously, but avoid any public exposure of the tuning APIs for now. We can turn them live once we completed the rewrite.

- [x] No SASS difference for `cub.bench.reduce.sum.base` on sm120
- [x] No SASS difference for `cub.bench.transform_reduce.sum.base` on sm120
- [x] Compile-time comparison between before and after this PR

Running 3 times (excluding 1 warmup)
```
touch ../cub/test/catch2_test_device_reduce.cu
time make cub.test.device.reduce.lid_0.types_0
```
before:
```
34.716s
34.758s
34.842s
```
after:
```
34.697s
34.477s
34.500s
```

Quick benchmark of `cub.bench.reduce.sum.base` on my RTX 5090, since the SASS diff would not cover regressions in host code. LGTM:
```c++
## [0] NVIDIA GeForce RTX 5090

|   I8    |      I32      |      2^16      |   6.127 us |       1.74% |   6.183 us |       3.24% |  0.056 us |   0.91% |   SAME   |
|   I8    |      I32      |      2^20      |   6.325 us |       9.10% |   6.529 us |      12.28% |  0.204 us |   3.23% |   SAME   |
|   I8    |      I32      |      2^24      |  18.513 us |       2.25% |  18.575 us |       2.91% |  0.062 us |   0.34% |   SAME   |
|   I8    |      I32      |      2^28      | 186.257 us |       0.67% | 185.953 us |       0.70% | -0.304 us |  -0.16% |   SAME   |
|   I8    |      I64      |      2^16      |   6.138 us |       2.74% |   6.144 us |       0.38% |  0.006 us |   0.09% |   SAME   |
|   I8    |      I64      |      2^20      |   6.219 us |       6.64% |   6.288 us |       8.30% |  0.069 us |   1.11% |   SAME   |
|   I8    |      I64      |      2^24      |  18.408 us |       0.64% |  18.437 us |       1.12% |  0.029 us |   0.16% |   SAME   |
|   I8    |      I64      |      2^28      | 185.568 us |       0.73% | 186.777 us |       0.68% |  1.209 us |   0.65% |   SAME   |
|   I16   |      I32      |      2^16      |   6.214 us |       6.61% |   6.236 us |       6.93% |  0.022 us |   0.35% |   SAME   |
|   I16   |      I32      |      2^20      |   8.185 us |       0.99% |   8.188 us |       0.58% |  0.002 us |   0.03% |   SAME   |
|   I16   |      I32      |      2^24      |  28.724 us |       1.37% |  28.770 us |       1.52% |  0.047 us |   0.16% |   SAME   |
|   I16   |      I32      |      2^28      | 347.629 us |       0.55% | 347.639 us |       0.55% |  0.009 us |   0.00% |   SAME   |
|   I16   |      I64      |      2^16      |   6.115 us |       2.25% |   6.112 us |       2.62% | -0.003 us |  -0.05% |   SAME   |
|   I16   |      I64      |      2^20      |   8.173 us |       1.21% |   8.188 us |       0.59% |  0.015 us |   0.18% |   SAME   |
|   I16   |      I64      |      2^24      |  29.304 us |       3.49% |  29.576 us |       3.91% |  0.271 us |   0.93% |   SAME   |
|   I16   |      I64      |      2^28      | 347.087 us |       0.56% | 347.213 us |       0.56% |  0.127 us |   0.04% |   SAME   |
|   I32   |      I32      |      2^16      |   6.126 us |       2.07% |   6.132 us |       1.67% |  0.005 us |   0.08% |   SAME   |
|   I32   |      I32      |      2^20      |   8.183 us |       0.83% |   8.186 us |       0.70% |  0.003 us |   0.04% |   SAME   |
|   I32   |      I32      |      2^24      |  51.217 us |       0.53% |  51.215 us |       0.38% | -0.002 us |  -0.00% |   SAME   |
|   I32   |      I32      |      2^28      | 672.272 us |       0.46% | 672.407 us |       0.50% |  0.135 us |   0.02% |   SAME   |
|   I32   |      I64      |      2^16      |   6.121 us |       1.83% |   6.139 us |       1.16% |  0.018 us |   0.29% |   SAME   |
|   I32   |      I64      |      2^20      |   8.190 us |       0.40% |   8.172 us |       1.28% | -0.018 us |  -0.22% |   SAME   |
|   I32   |      I64      |      2^24      |  51.259 us |       0.74% |  51.325 us |       1.02% |  0.066 us |   0.13% |   SAME   |
|   I32   |      I64      |      2^28      | 671.143 us |       0.61% | 670.702 us |       0.46% | -0.441 us |  -0.07% |   SAME   |
|   I64   |      I32      |      2^16      |   6.131 us |       1.63% |   6.135 us |       1.25% |  0.005 us |   0.08% |   SAME   |
|   I64   |      I32      |      2^20      |  12.444 us |       4.49% |  12.376 us |       3.37% | -0.068 us |  -0.55% |   SAME   |
|   I64   |      I32      |      2^24      |  95.237 us |       1.13% |  95.110 us |       1.12% | -0.127 us |  -0.13% |   SAME   |
|   I64   |      I32      |      2^28      |   1.312 ms |       0.14% |   1.312 ms |       0.15% |  0.045 us |   0.00% |   SAME   |
|   I64   |      I64      |      2^16      |   6.135 us |       1.99% |   6.136 us |       2.53% |  0.001 us |   0.02% |   SAME   |
|   I64   |      I64      |      2^20      |  14.277 us |       2.22% |  14.311 us |       1.59% |  0.035 us |   0.24% |   SAME   |
|   I64   |      I64      |      2^24      |  95.506 us |       1.02% |  95.617 us |       1.03% |  0.112 us |   0.12% |   SAME   |
|   I64   |      I64      |      2^28      |   1.313 ms |       0.15% |   1.312 ms |       0.11% | -0.291 us |  -0.02% |   SAME   |
|  I128   |      I32      |      2^16      |   6.565 us |      12.21% |   6.615 us |      12.69% |  0.050 us |   0.76% |   SAME   |
|  I128   |      I32      |      2^20      |  18.460 us |       1.82% |  18.444 us |       1.16% | -0.016 us |  -0.09% |   SAME   |
|  I128   |      I32      |      2^24      | 188.893 us |       1.50% | 189.003 us |       1.52% |  0.110 us |   0.06% |   SAME   |
|  I128   |      I32      |      2^28      |   2.581 ms |       0.14% |   2.581 ms |       0.13% |  0.101 us |   0.00% |   SAME   |
|  I128   |      I64      |      2^16      |   6.165 us |       2.94% |   6.211 us |       5.45% |  0.046 us |   0.75% |   SAME   |
|  I128   |      I64      |      2^20      |  18.637 us |       3.38% |  18.854 us |       4.38% |  0.216 us |   1.16% |   SAME   |
|  I128   |      I64      |      2^24      | 183.319 us |       0.97% | 183.897 us |       1.05% |  0.578 us |   0.32% |   SAME   |
|  I128   |      I64      |      2^28      |   2.573 ms |       0.11% |   2.573 ms |       0.10% |  0.118 us |   0.00% |   SAME   |
|   F32   |      I32      |      2^16      |   6.216 us |       6.39% |   6.144 us |       0.00% | -0.072 us |  -1.17% |   ????   |
|   F32   |      I32      |      2^20      |   8.189 us |       0.40% |   8.188 us |       0.58% | -0.001 us |  -0.01% |   SAME   |
|   F32   |      I32      |      2^24      |  51.843 us |       2.05% |  51.733 us |       1.98% | -0.110 us |  -0.21% |   SAME   |
|   F32   |      I32      |      2^28      | 666.607 us |       0.27% | 666.752 us |       0.25% |  0.145 us |   0.02% |   SAME   |
|   F32   |      I64      |      2^16      |   6.126 us |       1.63% |   6.136 us |       1.10% |  0.010 us |   0.17% |   SAME   |
|   F32   |      I64      |      2^20      |   9.863 us |       7.95% |   9.680 us |       9.39% | -0.183 us |  -1.86% |   SAME   |
|   F32   |      I64      |      2^24      |  51.605 us |       1.66% |  51.636 us |       1.68% |  0.031 us |   0.06% |   SAME   |
|   F32   |      I64      |      2^28      | 668.767 us |       0.43% | 668.438 us |       0.39% | -0.329 us |  -0.05% |   SAME   |
|   F64   |      I32      |      2^16      |  10.204 us |       1.37% |  10.222 us |       1.13% |  0.018 us |   0.18% |   SAME   |
|   F64   |      I32      |      2^20      |  14.522 us |       4.02% |  14.466 us |       3.40% | -0.057 us |  -0.39% |   SAME   |
|   F64   |      I32      |      2^24      |  96.754 us |       1.11% |  96.618 us |       1.11% | -0.137 us |  -0.14% |   SAME   |
|   F64   |      I32      |      2^28      |   1.317 ms |       0.24% |   1.317 ms |       0.24% | -0.008 us |  -0.00% |   SAME   |
|   F64   |      I64      |      2^16      |  10.189 us |       1.81% |  10.233 us |       1.45% |  0.044 us |   0.43% |   SAME   |
|   F64   |      I64      |      2^20      |  14.373 us |       2.37% |  14.337 us |       0.20% | -0.036 us |  -0.25% |   FAST   |
|   F64   |      I64      |      2^24      |  96.030 us |       0.86% |  96.017 us |       0.86% | -0.013 us |  -0.01% |   SAME   |
|   F64   |      I64      |      2^28      |   1.313 ms |       0.18% |   1.313 ms |       0.17% | -0.124 us |  -0.01% |   SAME   |
|   C32   |      I32      |      2^16      |   6.142 us |       1.25% |   6.133 us |       1.57% | -0.009 us |  -0.15% |   SAME   |
|   C32   |      I32      |      2^20      |  12.470 us |       4.63% |  12.505 us |       5.04% |  0.035 us |   0.28% |   SAME   |
|   C32   |      I32      |      2^24      |  95.110 us |       1.09% |  94.948 us |       1.09% | -0.161 us |  -0.17% |   SAME   |
|   C32   |      I32      |      2^28      |   1.311 ms |       0.13% |   1.311 ms |       0.13% | -0.088 us |  -0.01% |   SAME   |
|   C32   |      I64      |      2^16      |   6.125 us |       1.67% |   6.144 us |       0.00% |  0.019 us |   0.31% |   ????   |
|   C32   |      I64      |      2^20      |  14.251 us |       2.79% |  14.206 us |       3.52% | -0.045 us |  -0.31% |   SAME   |
|   C32   |      I64      |      2^24      |  95.332 us |       1.13% |  95.240 us |       1.12% | -0.091 us |  -0.10% |   SAME   |
|   C32   |      I64      |      2^28      |   1.312 ms |       0.13% |   1.312 ms |       0.11% | -0.170 us |  -0.01% |   SAME   |
```

Fixes: #6565

### DIFF
--- a/c/parallel/src/reduce.cu
+++ b/c/parallel/src/reduce.cu
@@ -177,7 +177,7 @@ CUresult cccl_device_reduce_build_ex(
       }
       else if (accum_t.type == CCCL_FLOAT64)
       {
-        accum_type = accum_type::double32;
+        accum_type = accum_type::float64;
       }
 
       auto operation_t = op_type::unknown;
--- a/cub/cub/device/dispatch/tuning/tuning_reduce.cuh
+++ b/cub/cub/device/dispatch/tuning/tuning_reduce.cuh
@@ -271,7 +271,7 @@ struct sm100_tuning<double, OffsetT, op_type::plus, offset_size::_4, accum_size:
 enum class accum_type
 {
   float32,
-  double32,
+  float64,
   other,
 };
 
@@ -280,7 +280,7 @@ _CCCL_HOST_DEVICE constexpr accum_type classify_accum_type()
 {
   return ::cuda::std::is_same_v<AccumT, float> ? accum_type::float32
        : ::cuda::std::is_same_v<AccumT, double>
-         ? accum_type::double32
+         ? accum_type::float64
          : accum_type::other;
 }
 
@@ -303,7 +303,7 @@ _CCCL_API constexpr auto get_sm100_tuning(accum_type accum_t, op_type operation_
   {
     return sm100_tuning_values{16, 512, 2};
   }
-  if (accum_t == accum_type::double32 && offset_size == 4 && accum_size == 8)
+  if (accum_t == accum_type::float64 && offset_size == 4 && accum_size == 8)
   {
     return sm100_tuning_values{16, 640, 1};
   }


cub/cub/detail/launcher/cuda_driver.cuh
```diff
@@ -81,7 +81,7 @@ struct CudaDriverLauncherFactory
 
   ::cudaError_t PtxVersion(int& version) const
   {
-    version = cc;
+    version = cc * 10;
```
bernhardmgruber, line 87: I think that's a bug in `main`.

bernhardmgruber: #6567

bernhardmgruber:
Found an issue with the way the accumulator type is specified in the benchmarks, which explains the regressions I currently observe when using the public tuning API: #6576

cub/benchmarks/bench/reduce/base.cuh
```diff
@@ -5,88 +5,188 @@
...
-  };
-
-  using MaxPolicy = policy_t;
+    const auto policy = cub::agent_reduce_policy{
+      .block_threads      = TUNE_THREADS_PER_BLOCK,
+      .items_per_thread   = TUNE_ITEMS_PER_THREAD,
+      .vector_load_length = 1 << TUNE_ITEMS_PER_VEC_LOAD_POW2,
+      .block_algorithm    = cub::BLOCK_REDUCE_WARP_REDUCTIONS,
+      .load_modifier      = cub::LOAD_DEFAULT};
+    return {policy, policy, policy, policy};
```
miscco, line 21: I love those quadruplets

cub/cub/device/dispatch/tuning/tuning_reduce.cuh
```diff
@@ -17,8 +17,56 @@
...
 
+struct agent_reduce_policy // equivalent of AgentReducePolicy
+{
+  int block_threads;
+  int items_per_thread;
+  int vector_load_length;
+  BlockReduceAlgorithm block_algorithm;
+  CacheLoadModifier load_modifier;
+
+#if !_CCCL_COMPILER(NVRTC)
```
miscco, line 57: Nitpick: this is missing `ìnclude <ostream>`

cub/cub/device/dispatch/tuning/tuning_reduce.cuh
```diff
@@ -184,8 +224,57 @@ struct sm100_tuning<double, OffsetT, op_type::plus, offset_size::_4, accum_size:
   static constexpr int items_per_vec_load = 1;
 };
 
+enum class accum_type
+{
+  _float,
+  _double,
+  other,
```
miscco: how about 
```suggestion
  float32,
  float64,
  other,
```

cub/cub/device/dispatch/tuning/tuning_reduce.cuh
```diff
@@ -184,8 +224,57 @@ struct sm100_tuning<double, OffsetT, op_type::plus, offset_size::_4, accum_size:
...
+}
+
+struct sm100_tuning_values
+{
+  int items;
+  int threads;
+  int items_per_vec_load;
+};
+
+_CCCL_API constexpr auto get_sm100_tuning(accum_type at, op_type ot, offset_size os, accum_size as)
```
miscco: nitpick: early return

cub/cub/device/dispatch/tuning/tuning_reduce.cuh
```diff
@@ -279,6 +368,95 @@ struct policy_hub
...
+{
+  accum_type at;
+  op_type ot;
+  offset_size os;
+  int accum_size;
+
+  // IDEA(bgruber): instead of the constexpr function, we could also provide a map<int, reduce_arch_policy> and move the
+  // selection mechanism elsewhere
+
+  _CCCL_API constexpr auto operator()(int arch) const -> reduce_arch_policy
```
miscco: Nitpick: 
```suggestion
  [[nodiscard]] _CCCL_API constexpr auto operator()(int arch) const -> reduce_arch_policy
```

cub/cub/device/dispatch/tuning/tuning_reduce.cuh
```diff
@@ -279,6 +368,95 @@ struct policy_hub
...
+  {
+    if (arch >= 1000)
```
miscco: I like this a lot

cub/cub/util_arch.cuh
```diff
@@ -112,28 +114,61 @@ inline constexpr bool prefer_conflict_over_padding = CUB_PTX_PREFER_CONFLICT_OVE
...
 static constexpr ::cuda::std::size_t max_smem_per_block = 48 * 1024;
 
+struct scaling_result
+{
+  int items_per_thread;
+  int block_threads;
+};
+
+_CCCL_API constexpr auto
+scale_reg_bound(int nominal_4B_block_threads, int nominal_4B_items_per_thread, int target_type_size) -> scaling_result
```
miscco:
```suggestion
[[nodiscard]] _CCCL_API inline constexpr auto
scale_reg_bound(int nominal_4B_block_threads, int nominal_4B_items_per_thread, int target_type_size) -> scaling_result
```

cub/cub/util_arch.cuh
```diff
@@ -112,28 +114,61 @@ inline constexpr bool prefer_conflict_over_padding = CUB_PTX_PREFER_CONFLICT_OVE
...
-                          * 32);
+private:
+  static constexpr auto result = scale_reg_bound(Nominal4ByteBlockThreads, Nominal4ByteItemsPerThread, int{sizeof(T)});
+
+public:
+  static constexpr int ITEMS_PER_THREAD = result.items_per_thread;
+  static constexpr int BLOCK_THREADS    = result.block_threads;
 };
 
+_CCCL_API constexpr auto
```
miscco:
```suggestion
[[nodiscard]] _CCCL_API inline constexpr auto
```

cub/benchmarks/bench/reduce/base.cuh
bernhardmgruber: I think this is the result of ordinary reduce, segmented reduce and non-deterministic reduce sharing the same tuning hub for no reason. It should be refactored at some point.

cub/cub/device/dispatch/tuning/tuning_reduce.cuh
bernhardmgruber: I don't mind, so I take it.

cub/cub/device/dispatch/tuning/tuning_reduce.cuh
bernhardmgruber: Sorry, can you please elaborate? It's not clear what you mean.

cub/cub/device/dispatch/tuning/tuning_reduce.cuh
bernhardmgruber: Yeah, the complexity reduction of a constexpr function vs. the previous TMP logic is enormous. I hope we can stay with this, since we lose the ability to ask for which arches the policy provides tunings. This could easily be added with an additional array data member, but then there are two pieces which can get out of sync.

cub/cub/device/dispatch/tuning/tuning_reduce.cuh
miscco: This was attached at the wrong place, I meant
```cpp
  if (ot != op_type::plus)
  {
    return {};
  }
  ....
```

cub/cub/util_arch.cuh
bernhardmgruber: Is that really meaningful? `inline` is implied by `constexpr` and since the function has no side effects, not using the return value is pointless.

miscco: We do have seen issues with functions that are not marked as inline being flagged by the linker. nodiscard aint hurt nobody yet

bernhardmgruber: Fine.

bernhardmgruber: Applied to #6575.

c/parallel/src/reduce.cu
```diff
@@ -206,13 +181,7 @@ struct __align__({2}) storage_t {{
 {3}
 {4}
 {5}
-using device_reduce_policy = {6}::MaxPolicy;
-
-#include <cub/detail/ptx-json/json.h>
-__device__ consteval auto& policy_generator() {{
-  return ptx_json::id<ptx_json::string("device_reduce_policy")>()
-    = cub::detail::reduce::ReducePolicyWrapper<device_reduce_policy::ActivePolicy>::EncodedPolicy();
-}};
```
griwes, line 215: Maybe it'd be just a debug mode thing, but I'd like us to have some sort of a check here that the policy we've computed on the host is, in fact, the same policy that we computed in the JITed code. Right now that's by construction, but I worry slightly about running into weird esoteric bugs thanks to nvrtc behaving differently and/or there being some weird header mismatch somewhere, and then having the worst time trying to figure out what's going on when we could catch it here.

c/parallel/src/reduce.cu
```diff
@@ -286,18 +254,40 @@ __device__ consteval auto& policy_generator() {{
...
+        ot = op_type::min_or_max;
+        break;
+      default:
+        break;
+    }
 
-    using cub::detail::RuntimeReduceAgentPolicy;
-    auto reduce_policy = RuntimeReduceAgentPolicy::from_json(runtime_policy, "ReducePolicy");
-    auto st_policy     = RuntimeReduceAgentPolicy::from_json(runtime_policy, "SingleTilePolicy");
+    auto os = offset_size::_8; // sizeof(uint64_t)
```
griwes: This should be centralized. Not just for c.parallel (so that we can avoid re-stating this over and over again in mimicry of the CUB `classify` calls), but also for CUB itself so that c.parallel can just do this per category (op_type, accum_type) instead of doing it per algorithm.

c/parallel/src/reduce.cu
```diff
@@ -330,30 +320,20 @@ CUresult cccl_device_reduce(
...
-        {},
-        {build},
-        cub::detail::CudaDriverLauncherFactory{cu_device, build.cc},
-        *reinterpret_cast<reduce::reduce_runtime_tuning_policy*>(build.runtime_policy));
+    using namespace cub::detail::reduce;
+    auto exec_status = ::cub::detail::reduce::dispatch<void>(
+      d_temp_storage,
+      *temp_storage_bytes,
+      indirect_arg_t{d_in}, // TODO(bgruber): why not indirect_iterator_t?
+      indirect_arg_t{d_out}, // TODO(bgruber): why not indirect_iterator_t?
```
griwes: The answer is that reduce does not need the extra features of `indirect_iterator_t`; think of `indirect_arg_t` as just a type erased pointer, and `indirect_iterator_t` as a type erased pointer that can also be incremented. Reduce doesn't do that on the host, so we don't care here.

cub/cub/device/dispatch/kernels/kernel_reduce.cuh
```diff
@@ -121,29 +119,33 @@ finalize_and_store_aggregate(OutputIteratorT d_out, ReductionOpT, empty_problem_
...
 {
+  static constexpr agent_reduce_policy policy = ArchPolicies{}(CUB_PTX_ARCH).reduce_policy;
+  using agent_policy_t =
+    AgentReducePolicy<policy.block_threads,
+                      policy.items_per_thread,
+                      AccumT,
+                      policy.vector_load_length,
+                      policy.block_algorithm,
+                      policy.load_modifier,
+                      NoScaling<policy.block_threads, policy.items_per_thread, AccumT>>;
```
griwes: Can't wait for C++20.

cub/cub/device/dispatch/tuning/tuning_reduce.cuh
```diff
@@ -17,8 +17,60 @@
...
+  BlockReduceAlgorithm block_algorithm;
+  CacheLoadModifier load_modifier;
+
+#if !_CCCL_COMPILER(NVRTC)
+  friend ::std::ostream& operator<<(::std::ostream& os, const agent_reduce_policy& p)
+  {
+    return os << "agent_reduce_policy { block_threads = " << p.block_threads
+              << ", items_per_thread = " << p.items_per_thread << ", vector_load_length = " << p.vector_load_length
+              << ", block_algorithm = " << p.block_algorithm << ", load_modifier = " << p.load_modifier << " }";
+  }
```
griwes, line 63: Maybe turn this into a print of valid C++ (with designated initializers and whatnot), and then we could easily reuse this in c.parallel to resolve my earlier comment about checking that we computed the same policy on both sides? (If we then provide `operator==` for all of these, the requested debug check would just turn into `static_assert(policy_hub{}(ptx_ver) == {});` in the format string.)

cub/cub/device/dispatch/tuning/tuning_reduce.cuh
```diff
@@ -17,8 +17,60 @@
...
+  agent_reduce_policy segmented_reduce_policy;
+  agent_reduce_policy reduce_nondeterministic_policy;
+
+#if !_CCCL_COMPILER(NVRTC)
+  friend ::std::ostream& operator<<(::std::ostream& os, const reduce_arch_policy& p)
+  {
+    return os << "reduce_arch_policy { reduce_policy = " << p.reduce_policy << ", single_tile_policy = "
+              << p.single_tile_policy << ", segmented_reduce_policy = " << p.segmented_reduce_policy
+              << ", reduce_nondeterministic_policy = " << p.reduce_nondeterministic_policy << " }";
+  }
```
griwes, line 93: Same here.

cub/cub/device/dispatch/tuning/tuning_reduce.cuh
```diff
@@ -17,8 +17,60 @@
...
+};
+
+#if _CCCL_STD_VER >= 2020
+consteval void needs_a_constexpr_value(auto) {}
+
+template <typename T>
+concept reduce_policy_hub = requires(T hub, int arch) {
+  requires ::cuda::std::is_empty_v<T>;
+  { hub(arch) } -> ::cuda::std::same_as<reduce_arch_policy>;
+  { needs_a_constexpr_value(hub(0)) }; // also checks that we return a value for PTX version 0 (host code)
```
griwes: ...this gives me an idea for a C++ core language feature...

cub/cub/device/dispatch/tuning/tuning_reduce.cuh
```diff
@@ -279,6 +372,94 @@ struct policy_hub
...
+        rp = (*this)(600).reduce_policy;
+      }
+
+      const auto rp_nondet = agent_reduce_policy{
+        rp.block_threads,
+        rp.items_per_thread,
+        rp.vector_load_length,
+        BLOCK_REDUCE_WARP_REDUCTIONS_NONDETERMINISTIC,
+        rp.load_modifier};
+      return {rp, rp, rp, rp_nondet};
```
griwes: Move this into the if above and then we can get rid of that ugly `else` branch. Doing the `(*this)(fallback)` dance feels icky, but as long as it can be avoided except in some really weird cases, it's fine. If it turns out we somehow need it more often, we should look for some more declarative abstraction to that I think.

cub/cub/device/dispatch/tuning/tuning_reduce.cuh
```diff
@@ -184,8 +228,57 @@ struct sm100_tuning<double, OffsetT, op_type::plus, offset_size::_4, accum_size:
...
+  int threads;
+  int items_per_vec_load;
+};
+
+_CCCL_API constexpr auto get_sm100_tuning(accum_type at, op_type ot, offset_size os, accum_size as)
+  -> ::cuda::std::optional<sm100_tuning_values>
+{
+  if (ot == op_type::plus)
+  {
+    if (at == accum_type::float32 && os == offset_size::_4 && as == accum_size::_4)
```
griwes: I do hate this part. I think I like the overall structure of the whole thing, but these arch-specific parts could really use some (any) pattern matching sugar on top of them.

cub/cub/device/dispatch/tuning/tuning_reduce.cuh
```diff
@@ -279,6 +372,94 @@ struct policy_hub
 
   using MaxPolicy = Policy1000;
 };
+
+struct arch_policies // equivalent to the policy_hub, holds policies for a bunch of CUDA architectures
+{
+  accum_type at;
+  op_type ot;
+  offset_size os;
+  int accum_size;
```
griwes: I'm a little bit worried about this. The way we have constructed it here, it is an ABI (and possibly API in some cases?) break to add "traits" to the selection process; same for exposing more knobs to the user. I wonder if instead of doing this, having a user / the dispatch code actually pass in something more resembling an environment that provides the queries. Yes, that'd turn this back into a template, but if we consistently passed in an equivalently-typed environment, that wouldn't be too bad, *and* it'd provide us ways of extending the knobs without breaks in the future.

cub/cub/device/dispatch/tuning/tuning_reduce.cuh
```diff
@@ -279,6 +372,94 @@ struct policy_hub
...
+};
+
+// stateless version which can be passed to kernels
+template <typename AccumT, typename OffsetT, typename ReductionOpT>
+struct arch_policies_from_types
+{
+  [[nodiscard]] _CCCL_API constexpr auto operator()(int arch) const -> reduce_arch_policy
+  {
+    constexpr auto policies = arch_policies{
+      classify_accum_type<AccumT>(), classify_op<ReductionOpT>(), classify_offset_size<OffsetT>(), int{sizeof(AccumT)}};
```
griwes: Re: the above comment about potential ABI breaks - this would then just turn into a construction of an environment from 4 "runtime"-valued queries, with the solution I proposed above.

cub/cub/device/dispatch/tuning/tuning_reduce.cuh
```diff
@@ -279,6 +372,94 @@ struct policy_hub
 
   using MaxPolicy = Policy1000;
 };
+
+struct arch_policies // equivalent to the policy_hub, holds policies for a bunch of CUDA architectures
```
griwes, line 422: This is an internal type, but one that still materializes when users invoke the algorithms, right? I wonder if this should turn into a template and its data members should be turned into an environment returning those values by queries, because as is, any change to the layout would be an ABI break...

jrhemstad:
@bernhardmgruber can we see a comparison in compile time between this approach and the new one for the DeviceReduce tests? I want to see if there is any impact (for better or worse) on compile time with the new tuning machinery.

c/parallel/src/reduce.cu
bernhardmgruber: That's a good point. I try to think what we can do in the new design.

c/parallel/src/reduce.cu
bernhardmgruber: Agreed!

c/parallel/src/reduce.cu
bernhardmgruber: Thanks, that makes a lot of sense! I updated the comment.

cub/benchmarks/bench/reduce/base.cuh
```diff
@@ -5,88 +5,188 @@
...
+  }
 };
 #endif // !TUNE_BASE
 
+#if 0
+struct caching_last_alloc_mr
```
bernhardmgruber: I think I will drop this for this PR before merging, but let it sit a bit in case we change our minds.

cub/cub/device/dispatch/kernels/kernel_reduce.cuh
bernhardmgruber: I fully agree! I would love to just pass `policy` as template argument to the agent.

cub/cub/device/dispatch/tuning/tuning_reduce.cuh
bernhardmgruber: I agree that pattern matching is the right tool for this job here, but we don't have it in constant evaluation in C++17 :/ There are third-party pattern matching libraries we may want to reach out too.

For now, I would like to keep the code as is, because it's simple and (probably) compiles fast. But we can revisit this design. I already started working on radix sort and there the story is a lot more complex.

cub/cub/device/dispatch/tuning/tuning_reduce.cuh
bernhardmgruber: A very appealing aspect of the current design is that tuning information is expressed very simply as structs with data members, so I would love if we could keep that.

Regarding API breaks, we do allow those at every release. This is pointed out in our [README](https://github.com/NVIDIA/cccl?tab=readme-ov-file#application-binary-interface-abi):

> Symbols in the `thrust::` and `cub::` namespaces may break ABI at any time without warning.

c/parallel/src/reduce.cu
```diff
@@ -286,18 +254,40 @@ __device__ consteval auto& policy_generator() {{
...
-      cub::detail::ptx_json::parse("device_reduce_policy", {result.data.get(), result.size});
+    // convert type information to CUB arch_policies
+    using namespace cub::detail::reduce;
+
+    auto accum_type = accum_type::other;
+    if (accum_t.type == CCCL_FLOAT32)
+    {
+      accum_type = accum_type::float32;
+    }
+    if (accum_t.type == CCCL_FLOAT64)
```
NaderAlAwar: NIT
```suggestion
    else if (accum_t.type == CCCL_FLOAT64)
```

c/parallel/src/reduce.cu
```diff
@@ -286,18 +254,40 @@ __device__ consteval auto& policy_generator() {{
...
+        operation_t = op_type::min_or_max;
+        break;
+      default:
+        break;
+    }
 
-    using cub::detail::RuntimeReduceAgentPolicy;
-    auto reduce_policy = RuntimeReduceAgentPolicy::from_json(runtime_policy, "ReducePolicy");
-    auto st_policy     = RuntimeReduceAgentPolicy::from_json(runtime_policy, "SingleTilePolicy");
+    const int offset_size = int{sizeof(uint64_t)};
```
NaderAlAwar: Important:
```suggestion
    const int offset_size = int{sizeof(OffsetT)};
```
This way if we change `OffsetT` later it will be easier to do so

c/parallel/src/reduce.cu
```diff
@@ -330,30 +320,19 @@ CUresult cccl_device_reduce(
...
-        {},
-        {build},
-        cub::detail::CudaDriverLauncherFactory{cu_device, build.cc},
-        *reinterpret_cast<reduce::reduce_runtime_tuning_policy*>(build.runtime_policy));
+    auto exec_status = cub::detail::reduce::dispatch<void>(
+      d_temp_storage,
+      *temp_storage_bytes,
+      indirect_arg_t{d_in}, // could be indirect_iterator_t, but CUB does not need to increment it
+      indirect_arg_t{d_out}, // could be indirect_iterator_t, but CUB does not need to increment it
+      static_cast<::cuda::std::size_t>(num_items),
```
NaderAlAwar: Important:
```suggestion
      static_cast<OffsetT>(num_items),
```
This existed before but it would be good to fix it here

cub/benchmarks/bench/reduce/base.cuh
```diff
@@ -5,88 +5,188 @@
...
   // Retrieve axis parameters
-  const auto elements = static_cast<std::size_t>(state.get_int64("Elements{io}"));
+  const auto elements = static_cast<offset_t>(state.get_int64("Elements{io}"));
 
   thrust::device_vector<T> in = generate(elements);
   thrust::device_vector<T> out(1);
 
-  input_it_t d_in   = thrust::raw_pointer_cast(in.data());
-  output_it_t d_out = thrust::raw_pointer_cast(out.data());
+  auto d_in  = thrust::raw_pointer_cast(in.data());
```
NaderAlAwar, line 38: Question: prior to this change `d_in` had the type `const T*`, but with this the `const` is gone. Could this affect performance?

cub/cub/device/dispatch/kernels/kernel_reduce.cuh
```diff
@@ -167,9 +170,6 @@ __launch_bounds__(int(ChainedPolicyT::ActivePolicy::ReducePolicy::BLOCK_THREADS)
  *        to aggregate privatized thread block reductions from a previous
  *        multi-block reduction pass.
  *
- * @tparam ChainedPolicyT
- *   Chained tuning policy
- *
```
NaderAlAwar, line 172: Suggestion: add back the doc for `ArchPolicies`

cub/cub/device/dispatch/tuning/tuning_reduce.cuh
bernhardmgruber: That is an amazing idea!

c/parallel/src/reduce.cu
```diff
@@ -206,21 +216,17 @@ struct __align__({2}) storage_t {{
...
 {5}
-using device_reduce_policy = {6}::MaxPolicy;
-
-#include <cub/detail/ptx-json/json.h>
-__device__ consteval auto& policy_generator() {{
-  return ptx_json::id<ptx_json::string("device_reduce_policy")>()
-    = cub::detail::reduce::ReducePolicyWrapper<device_reduce_policy::ActivePolicy>::EncodedPolicy();
-}};
+using device_reduce_policy = {6};
+static_assert(device_reduce_policy()(__CUDA_ARCH__) == {7}, "Host generated and JIT compiled policy mismatch");
```
bernhardmgruber: @griwes I added this check here to verify that the serialized tuning from the host (`{7}`) matches the tuning selected in device code. It works, really neat! Thx for the suggestion!

bernhardmgruber:
```
/home/bgruber/dev/cccl/c/parallel/src/jit_templates/jit_entry.h(16): error: static assertion failed with "Host generated and JIT compiled policy mismatch"
  static_assert(device_reduce_policy()(__CUDA_ARCH__) == reduce_arch_policy { .reduce_policy = agent_reduce_policy { .block_threads = 256, .items_per_thread = 20, .vector_load_length = 4, .block_algorithm = BLOCK_REDUCE_WARP_REDUCTIONS, .load_modifier = LOAD_LDG }, .single_tile_policy = agent_reduce_policy { .block_threads = 256, .items_per_thread = 20, .vector_load_length = 4, .block_algorithm = BLOCK_REDUCE_WARP_REDUCTIONS, .load_modifier = LOAD_LDG }, .segmented_reduce_policy = agent_reduce_policy { .block_threads = 256, .items_per_thread = 20, .vector_load_length = 4, .block_algorithm = BLOCK_REDUCE_WARP_REDUCTIONS, .load_modifier = LOAD_LDG }, .reduce_nondeterministic_policy = agent_reduce_policy { .block_threads = 256, .items_per_thread = 20, .vector_load_length = 4, .block_algorithm = BLOCK_REDUCE_WARP_REDUCTIONS_NONDETERMINISTIC, .load_modifier = LOAD_LDG } }, "Host generated and JIT compiled policy mismatch");
  ^
```

cub/cub/device/dispatch/tuning/tuning_reduce.cuh
bernhardmgruber: Totally! If we could require 
```c++
requires(T hub, int arch) {
  { hub(arch) } constexpr;
```
like we can do for `noexcept` today. That would be amazing! I think it's a hole in the current design. Let me know if want to draft a proposal! ;)

c/parallel/src/reduce.cu
bernhardmgruber: I messed up converting major and minor CC to an arch int again. I will replace the logic by the strongly typed `cuda::arch_id` next.

bernhardmgruber: Converted to use `arch_id`

cub/benchmarks/bench/reduce/base.cuh
bernhardmgruber: I would say it shouldn't. But I have to do a benchmark or SASS test anyway, so we will see.

cub/benchmarks/bench/reduce/base.cuh
bernhardmgruber: Dropped

cub/cub/device/dispatch/tuning/tuning_reduce.cuh
griwes: I agree; however, it'd be nice to have the ABI break manifest as a linker error instead of being entirely silent.

cub/cub/device/dispatch/kernels/kernel_reduce.cuh
```diff
@@ -125,16 +126,16 @@ template <typename ArchPolicies,
...
-__launch_bounds__(int(ArchPolicies{}(CUB_PTX_ARCH).reduce_policy.block_threads)) void DeviceReduceKernel(
-  InputIteratorT d_in,
-  AccumT* d_out,
-  OffsetT num_items,
-  GridEvenShare<OffsetT> even_share,
-  ReductionOpT reduction_op,
-  TransformOpT transform_op)
+CUB_DETAIL_KERNEL_ATTRIBUTES __launch_bounds__(int(
+  ArchPolicies{}(::cuda::arch_id{CUB_PTX_ARCH / 10})
+    .reduce_policy.block_threads)) void DeviceReduceKernel(InputIteratorT d_in,
```
griwes, line 137: ...this formatting got really bad...

griwes: Could we move the expression into an `auto BlockThreads = ...` at the end of the template parameter list? That should let clang format this in a saner manner.

cub/cub/device/dispatch/tuning/tuning_reduce.cuh
bernhardmgruber: I agree that for this case here, we can simplify the code. What I wanted to demonstrate here is the pattern we have sometimes, where we try to lookup a tuning for a certain architecture, and if there is none, we either fall through or modify a tuning of a previous architecture.

I will refactor the code now.

cub/cub/device/dispatch/kernels/kernel_reduce.cuh
```diff
@@ -121,29 +123,34 @@ finalize_and_store_aggregate(OutputIteratorT d_out, ReductionOpT, empty_problem_
...
           typename InputIteratorT,
           typename OffsetT,
           typename ReductionOpT,
           typename AccumT,
-          typename TransformOpT>
-CUB_DETAIL_KERNEL_ATTRIBUTES
-__launch_bounds__(int(ChainedPolicyT::ActivePolicy::ReducePolicy::BLOCK_THREADS)) void DeviceReduceKernel(
+          typename TransformOpT,
+          int BlockThreads = ArchPolicies{}(::cuda::arch_id{CUB_PTX_ARCH / 10}).reduce_policy.block_threads>
+CUB_DETAIL_KERNEL_ATTRIBUTES __launch_bounds__(BlockThreads) void DeviceReduceKernel(
```
bernhardmgruber: Critical: this does not work, since now the value of `CUB_PTX_ARCH` influences the symbol generated from the template, so host and device symbol no longer match.

bernhardmgruber: Reverted

griwes: Dang. But also grrrr.

cub/cub/device/dispatch/tuning/tuning_reduce.cuh
bernhardmgruber: Done

cub/cub/device/dispatch/tuning/tuning_reduce.cuh
bernhardmgruber: Doesn't this already happen automatically, since each CCCL release will have all CUB and Thrust entities in a different inline namespace? Like, now the type is called `cub::_V_300300_SM120::detail::reduce`. With the next release, the `300300` changes to `300400`. What more is needed?

cub/benchmarks/bench/reduce/base.cuh
bernhardmgruber: No difference observed.

cub/cub/device/dispatch/tuning/tuning_reduce.cuh
```diff
@@ -17,12 +17,99 @@
...
+
+// TODO(bgruber): bikeshed name before we make the tuning API public
+template <typename T>
+concept reduce_policy_hub = requires(T hub, ::cuda::arch_id arch) {
+// nvcc < 12.2 cannot handle a fully qualified name to a concept
+#  if _CCCL_CUDA_COMPILER(NVCC, <, 12, 2)
+  { hub(arch) } -> cuda::std::same_as<reduce_arch_policy>;
+#  else // _CCCL_CUDA_COMPILER(NVCC, <, 12, 2)
+  { hub(arch) } -> ::cuda::std::same_as<reduce_arch_policy>;
+#  endif // _CCCL_CUDA_COMPILER(NVCC, <, 12, 2)
```
NaderAlAwar: Important: we have a macro `_CCCL_CONCEPT_VSTD` that handles this issue

```suggestion
  { hub(arch) } -> _CCCL_CONCEPT_VSTD::same_as<reduce_arch_policy>;
```

cub/cub/device/dispatch/tuning/tuning_reduce.cuh
```diff
@@ -184,8 +272,61 @@ struct sm100_tuning<double, OffsetT, op_type::plus, offset_size::_4, accum_size:
   static constexpr int items_per_vec_load = 1;
 };
 
+// TODO(bgruber): we should have a more central enum for types, like cccl_type_enum in CCCL.C
+enum class accum_type
+{
+  float32,
+  double32,
```
NaderAlAwar: Important: this looks like a typo, should be `float64`

cub/cub/device/dispatch/dispatch_streaming_reduce.cuh
```diff
@@ -206,28 +206,9 @@ struct dispatch_streaming_arg_reduce_t
...
-    // Per-partition DispatchReduce template specialization
-    using dispatch_reduce_t =
-      DispatchReduce<arg_index_input_iterator_t,
-                     accumulating_transform_out_it_t,
-                     PerPartitionOffsetT,
-                     ReductionOpT,
-                     empty_problem_init_t,
-                     per_partition_accum_t,
-                     ::cuda::std::identity,
-                     PolicyChainT>;
```
bernhardmgruber, line 229: Critical: `PolicyChainT` was passed here to `dispatch_reduce_t` and is later picked up by `dispatch_reduce_t::Dispatch`. The replacement call `reduce::dispatch<per_partition_accum_t>` no longer carries forward this information.

bernhardmgruber: Issue: #7804

PR #6550, opened 2025-11-07
Generate `numeric_limits` for floating points from `__fp_meow` machinery

davebayer:
This PR simplifies the `numeric_limits` implementation for floating point types by reusing our already existing `__fp_meow` machinery.


libcudacxx/include/cuda/std/limits
```diff
@@ -22,7 +22,6 @@
 #endif // no system header
 
 #include <cuda/std/__limits/numeric_limits.h>
-#include <cuda/std/__limits/numeric_limits_ext.h>
```
fbusato: many times we use `numeric_limits` for builtin types.
Separate the implementations is good thing to minimize included code.

libcudacxx/include/cuda/std/__floating_point/constants.h
```diff
@@ -369,6 +369,58 @@ template <class _Tp>
...
+{
+  using _Storage = __fp_storage_t<_Fmt>;
+  if constexpr (_Fmt == __fp_format::__fp4_nv_e2m1 || _Fmt == __fp_format::__fp6_nv_e2m3)
+  {
+    return _Storage{0x1u};
+  }
+  else
+  {
+    return static_cast<_Storage>(
+      (static_cast<_Storage>(__fp_exp_bias_v<_Fmt> + 1 - __fp_digits_v<_Fmt>) << __fp_mant_nbits_v<_Fmt>)
```
fbusato, line 385: is `_Storage` always unsigned right?

libcudacxx/include/cuda/std/__limits/numeric_limits.h
```diff
@@ -302,282 +301,177 @@ class __numeric_limits_impl<bool, __numeric_limits_type::__bool>
...
-  {
-#  if defined(LDBL_TRUE_MIN)
-    return LDBL_TRUE_MIN;
-#  else // ^^^ LDBL_TRUE_MIN ^^^ // vvv !LDBL_TRUE_MIN vvv
-    return __LDBL_DENORM_MIN__;
-#  endif // ^^^ !LDBL_TRUE_MIN ^^^
-  }
-
-  static constexpr bool is_iec559  = true;
+  static constexpr int radix       = 2;
```
fbusato, line 377: should we not add at least a `static_assert` for `FLT_RADIX`?

libcudacxx/include/cuda/std/__limits/numeric_limits.h
```diff
@@ -302,282 +301,177 @@ class __numeric_limits_impl<bool, __numeric_limits_type::__bool>
...
+    else
+    {
+      using _Storage = __fp_storage_t<__fmt>;
+      return static_cast<_Storage>((static_cast<_Storage>(__fp_exp_bias_v<__fmt> - 1) << __fp_mant_nbits_v<__fmt>)
+                                   | __fp_explicit_bit_mask_v<__fmt>);
+    }
+  }
+
+  static constexpr int min_exponent   = __fp_exp_min_v<__fmt> + 1;
+  static constexpr int min_exponent10 = (30103ll * __fp_exp_min_v<__fmt>) / 100000ll;
```
fbusato, line 407: please comment magic numbers

libcudacxx/test/libcudacxx/libcxx/numerics/floating.point/constants/epsilon.pass.cpp
```diff
@@ -0,0 +1,102 @@
...
+  assert(cuda::std::__fp_get_storage(cuda::std::__fp_epsilon<T>()) == cuda::std::__fp_epsilon<fmt>());
+  static_assert(((void) cuda::std::__fp_epsilon<T>(), true));
+}
+
+__host__ __device__ bool test()
+{
+  using namespace test_integer_literals;
+
+  // 1. Test formats
+  test_fp_epsilon<cuda::std::__fp_format::__binary16>(0x1400u);
```
fbusato, line 42: would not be better to compute the epsilon instead of fixed value? e.g. (1 + "last_bit") - 1

fbusato (changes requested):
the PR removes several lines from `numeric_limits.h` and delete `numeric_limits_ext.h`, which is great! On the other hand, many times we use `numeric_limits` without the need of extended types. Would not be better to separate the implementations?

libcudacxx/include/cuda/std/__floating_point/constants.h
davebayer: Yes

PR #6553, opened 2025-11-07
Optimize `cuda::is_address_space`  by forcing the memory space

fbusato:
From the documentation

"If the memory space of the input pointer matches the requested address space, the function marks the pointer as belonging to that address space. For example, a subsequent store to a generic address that maps to shared memory emits an ``STS`` SASS instruction rather than the generic ``ST`` instruction."


fbusato:
do not merge yet. The evaluation of `__isLocal()` behavior is ongoing

alliepiper:
> do not merge yet. The evaluation of __isLocal() behavior is ongoing

Moving to draft to clear review requests.

davebayer (approved):

Issue #6163, opened 2025-10-08
Refactor `lit`-based testing infrastructure to eliminate redudnancy with cmake-provided information


jrhemstad:
There's a lot of stuff that `lit` does that is redundant with what cmake already does for us.

We should boil away as much of that as we can and just use cmake for that such that `lit` just does:  
- Build/Test files for a given directory
- Test file meta-language 
   - `.pass` or `.fail` file name convention for pass/fail at compile/runtime
   - Compiler/CTK version/feature specific comments 

From first principles, we want using `lit` to look like:

1. Invoke `cmake` to generate site config file
2. Invoke lit using config file (`SITE_CONFIG=*config file* lit /path/to/file(s)`)
     - It will then compile/run all files in the given path 

For files following the convention:
- X.pass.cpp -> should pass compile & runtime
- X.fail.cpp -> should fail compile
- X.runfail.cpp -> should fail at runtime

With support in the test files themselves for "meta-comments", e.g., things like the `UNSUPPORTED:` key
- https://github.com/NVIDIA/cccl/blob/050b57c3199088674e0d7ddabdcdd32001a4c032/libcudacxx/test/libcudacxx/cuda/barrier/arrive_tx_cta.pass.cpp#L11-L12

davebayer:
Would it be possible to add `X.compile.pass.cpp` for tests that should be just compiled? This is a common case for e. g. testing deduction guides

alliepiper:
Attempting to document lit behavior here: https://github.com/NVIDIA/cccl/issues/6556

PR #6554, opened 2025-11-07
Fix oversubscription issue with lit precompile, label hack

alliepiper:
Moves the C2H tests earlier in the config so we can add the fake dependency to lit.
Explains and labels the hack we're using to avoid oversubscription.


libcudacxx/test/libcudacxx/CMakeLists.txt
```diff
@@ -60,9 +99,6 @@ if (NOT MSVC AND NOT ${CMAKE_CUDA_COMPILER_ID} STREQUAL "Clang")
 endif()
 
 if (MSVC)
-  # sccache cannot handle the -Fd option generationg pdb files
-  set(CMAKE_MSVC_DEBUG_INFORMATION_FORMAT Embedded)
```
alliepiper, line 90: Drive-by fix -- we define this at the CCCL level, no need to set it here. It doesn't have any effect on the lit tests defined in this file anyway.

libcudacxx/test/libcudacxx/CMakeLists.txt
```diff
@@ -136,10 +172,26 @@ if (NOT LIBCUDACXX_TEST_WITH_NVRTC)
...
+    # FIXME: This has nasty side effects:
+    # - It's fragile and must be updated every time we add new targets to libcudacxx
+    # - It oversubs `-dev` presets that configure libcudacxx alongside other CCCL projects
+    # - It makes it impossible to just build this target alone since it brings in the world
+    # See related issue https://github.com/NVIDIA/cccl/issues/6163.
+    DEPENDS
+      libcudacxx.test.public_headers
+      libcudacxx.test.internal_headers
+      libcudacxx.test.public_headers_host_only
+      libcudacxx.test.c2h_all
```
miscco, line 244: Question: Why arent we just limiting the number of parallel compilations via the `-j` flag?

miscco (changes requested):

libcudacxx/test/libcudacxx/CMakeLists.txt
alliepiper: That would force the entire build to be serial. We can have parallelism in both ninja and lit, the issue is having them overlap.

alliepiper: The long-term solution for this issue is NVIDIA/cccl#6163. This PR just labels the hack that already exists.

PR #6555, opened 2025-11-07
Update NVBench type string declarations for FP16 and BF16

fbusato:


nvbench_helper/nvbench_helper/nvbench_helper.cuh
```diff
@@ -34,11 +34,11 @@ NVBENCH_DECLARE_TYPE_STRINGS(uint128_t, "U128", "uint128_t");
 using complex = cuda::std::complex<float>;
 
 #if _CCCL_HAS_NVFP16()
-NVBENCH_DECLARE_TYPE_STRINGS(__half, "Half", "half");
+NVBENCH_DECLARE_TYPE_STRINGS(__half, "FP16", "half");
```
alliepiper: `F16` would be more consistent with the `F32` / `F64` strings built into NVBench.

```suggestion
NVBENCH_DECLARE_TYPE_STRINGS(__half, "F16", "half");
```

PR #6557, opened 2025-11-08
Add gersemi CMake formatter

alliepiper:
Adds configuration and precommit hooks for `gersemi`, a cmake formatter:

https://github.com/BlankSpruce/gersemi

I've reviewed all of the changes in this PR and am pleased overall with the results. The upstream maintainer has been responsive to questions, suggestions, and issues I've had in this process.

We have a total of 135 files with 20816 lines of CMake in our repo, and there were very few instances where I felt the need to manually format using the `# gersemi: [off|on]` hints -- pretty much all of them were for `COMMAND` argument formatting which I don't expect a formatter to get correct.


.gersemirc
```diff
@@ -0,0 +1,15 @@
...
+#  pip3 install gersemi
+## Run only on git-tracked files:
+#  gersemi -i -- $(git ls-files "*.cmake" "CMakeLists.txt" "**/CMakeLists.txt")
+
+# See "outcome configuration" section of gersemi --help for more info on these options.
+definitions: []
+disable_formatting: false
+extensions: [".gersemi/ext/builtins.py", ".gersemi/ext/cccl.py", ".gersemi/ext/thirdparty.py"]
+indent: 2
+line_length: 80
```
bernhardmgruber, line 11: Let's use the same 120 as the `.clang-format` for consistency. This way we can get more code on the screen as well.

alliepiper: I tried that at first -- I also highly value consistency, but I found the results much less readable.

More code on the screen isn't a benefit with cmake -- having things laid out in well-formatted / nested lists so they can be groked at a glance is far preferable to having things squeezed into as few lines as possible.

alliepiper:
> Are there any options to keep the code more dense?

Dense CMake quickly becomes unreadable, and I greatly prefer readability over compactness for the sake of compactness. Newlines are cheap :)

My biggest gripe about gersemi's formatting is that when it splits lines, it moves the first args to a new line, eg.

```cmake
message(STATUS
  "Line one"
  "Line two"
)
```

becomes

```cmake
message(
  STATUS
  "Line one"
  "Line two"
)
```

I do greatly prefer the first example, but I'm willing to concede that to avoid having to think about formatting going forward 🙂 This particular refactoring is the main contributor to the increase in size.

Overall, this PR adds ~2000 LOCs. Our repo has >20,000 lines of cmake total, so it's more like a 10% increase instead of 50% taken as a whole.

.gersemirc
alliepiper: I just reran locally with a line limit of 120 to see the specific changes it makes and found the results significantly worse (subjectively, ofc).

bernhardmgruber: I did the same experiment now locally and I think 120 looks a LOT better. Many function calls can now again stay on a single line, which greatly improves the readability of the code, since we can fit more on a screen.

I agree this is subjectively though. Maybe we should vote on this.

bernhardmgruber:
> > Are there any options to keep the code more dense?
> 
> Dense CMake quickly becomes unreadable, and I greatly prefer readability over compactness for the sake of compactness. Newlines are cheap :)

Compactness up to some degree does strongly correlate with readability for me. But I guess this is a personal preference.

> I'm willing to concede that to avoid having to think about formatting going forward 🙂

I agree!

.gersemirc
bernhardmgruber: Btw, the line width of 120 results in the following diff again:
```
 109 files changed, 835 insertions(+), 4022 deletions(-)
```
It takes out more than 3kLOCs again. That's a huge win IMO.

bernhardmgruber:
> Overall, this PR adds ~2000 LOCs. Our repo has >20,000 lines of cmake total, so it's more like a 10% increase instead of 50% taken as a whole.

Yeah, that's a better reasoning. Still, if we go with 120, there is no increase :)

alliepiper:
The issue with CMake is that there is a lot of structure _within_ those function arguments, have having them broken out makes that structure more apparent and obvious -- this is the readability gain that we get here. Having all of the args squeezed into a single line loses that structure.

Maybe it's because I've spent hundreds or thousands of hours reading and writing CMake at this point, but I "just know" what eg. `target_link_libraries` or `set_target_properties` "looks like" when it's properly formatted. I can grok it at a glance if it's split out and formatted, and know exactly what it's doing with very little cognitive work. I can spot bugs and issues much easier in code where the internal structure of a call is shown through its formatting. The keywords and options are obvious, the lists of libraries / options are laid out in a row and easy to inspect...it's overall very easy and simple to read. It's delightful -- or at least as close as CMake gets to it 😉

If all of the keywords, options, and values are compressed into a single line, it takes _much_ more cognitive effort to read through them, mentally parse them, group them, find what I'm looking for, catch bugs...each line, each token must be inspected one-by-one, rather being able to just skim / glance, and subtle bugs like quoting errors get lost in the soup. It's a *lot* more work to actually read and understand long, dense lines of CMake.

For these reasons I prefer the shorter, less dense lines of CMake. I'll also point out that I'm the one doing at least 98% of the CMake work for CCCL, so I'd like to ask that my opinions and subjective takes are allowed to carry a little more weight here 🙂

----

That said, there are some expansions I don't like and feel are unnecessary -- for instance `function(...)` / `macro(...)` definitions and calls to `cmake_parse_arguments()`. These could all go on a single line with no loss of readability, but gersemi currently splits all function calls that have more than 4 arguments in a group. I've asked for the ability to customize this per-command (https://github.com/BlankSpruce/gersemi/issues/78), and the maintainer seems receptive and is considering it.

If they decide that's worth implementing and maintaining, we'll have much more control over when expansion vs inlining occurs on a per-command basis. This will let me set up gersemi to expand the commands that are useful to have expanded, but inline those that we don't really care about (message, function, macro, etc). At this point we can revisit the line-length settings, once we're able to ensure that the commands that should be expanded for readability are never inlined.

Sound like a good compromise for now?

.gersemirc
alliepiper: Fewer LOC at the expense of readability is not a win in my book. I'm evaluating these changes through "as a CMake developer, what results are the easiest to work with?" not "What gives us fewer lines?" Those brief stats say nothing about what effect it actually has on the changes to the content.

Per @jrhemstad's comment on the Slack thread, let's continue with the current config and revisit as more features get added upstream that will provide better control 👍

PR #6560, opened 2025-11-09
fix omp scan bug

### LINKED ISSUE #6317: [BUG]: stdpar multicore scan tests fail at runtime after commit 1238f287f4f654bff12e4d4a29b7dfa1c40a12fe

charan-003:
using the first element of each block as the init value

closes #6317

### DIFF
--- a/thrust/thrust/system/omp/detail/scan.h
+++ b/thrust/thrust/system/omp/detail/scan.h
@@ -123,12 +123,9 @@ OutputIterator scan_impl(
   else
   {
     // For no init inclusive scan: exclusive_scan starting at second element with first element as init
-    if (num_threads > 1)
-    {
-      accum_t first_block_sum = block_sums[0];
-      ::cuda::std::exclusive_scan(
-        block_sums.begin() + 1, block_sums.end(), block_sums.begin() + 1, first_block_sum, wrapped_binary_op);
-    }
+    accum_t first_block_sum = block_sums[0];
+    ::cuda::std::exclusive_scan(
+      block_sums.begin() + 1, block_sums.end(), block_sums.begin() + 1, first_block_sum, wrapped_binary_op);
   }
 
   // Step 3: Scan each block with offset (N reads/writes)


thrust/thrust/system/omp/detail/scan.h
```diff
@@ -106,7 +106,25 @@ OutputIterator scan_impl(
...
+        }
+      }
+      else
+      {
+        // No init value: reduce uses first element as init
+        if (end > start)
+        {
+          block_sums[tid] = ::cuda::std::reduce(first + start + 1, first + end, *(first + start), wrapped_binary_op);
+        }
+      }
```
dkolsen-pgi: You want to always do the `else` branch.  The true branch is incorrect.  The initial value doesn't get incorporated into the result until step 2.  And you still have `accum_t{}` in there, which is not correct.  Get rid of the `if` statement entirely and just replace
```c++
block_sums[tid] = ::cuda::std::reduce(first + start, first + end, accum_t{}, wrapped_binary_op);
```
with
```c++
block_sums[tid] = ::cuda::std::reduce(first + start + 1, first + end, *(first + start), wrapped_binary_op);
```

thrust/thrust/system/omp/detail/scan.h
```diff
@@ -106,7 +106,25 @@ OutputIterator scan_impl(
...
+        }
+        else
+        {
+          block_sums[tid] = ::cuda::std::reduce(first + start, first + end, accum_t{}, wrapped_binary_op);
+        }
+      }
+      else
+      {
+        // No init value: reduce uses first element as init
+        if (end > start)
```
dkolsen-pgi: This check is unnecessary.  If `start < n`, then `end > start` is always true.

thrust/thrust/system/omp/detail/scan.h
```diff
@@ -117,7 +135,18 @@ OutputIterator scan_impl(
...
+    {
+      temporary_array<accum_t, DerivedPolicy> block_sums_copy(exec, num_threads);
+      for (int i = 0; i < num_threads; ++i)
+      {
+        block_sums_copy[i] = block_sums[i];
+      }
+      ::cuda::std::inclusive_scan(
+        block_sums_copy.begin(), block_sums_copy.begin() + num_threads - 1, block_sums.begin() + 1, wrapped_binary_op);
+      // block_sums[0] stays as-is (no prefix for first block)
+    }
```
dkolsen-pgi: This is more complex than necessary.  There is no need for `block_sums_copy`.  For the inclusive_scan with no init, just do an exclusive_scan starting at the second element where the initial value is the first element.
```
::cuda::std::exclusive_scan(block_sums.begin() + 1, block_sums.end(), block_sums.begin() + 1,
                            block_sums[0], wrapped_binary_op);
```

thrust/thrust/system/omp/detail/scan.h
```diff
@@ -130,13 +159,30 @@ OutputIterator scan_impl(
...
+          if (tid == 0)
+          {
+            ::cuda::std::inclusive_scan(first + start, first + end, result + start, wrapped_binary_op);
+          }
+          else
+          {
+            const accum_t prefix = block_sums[tid];
+            ::cuda::std::inclusive_scan(first + start, first + end, result + start, wrapped_binary_op, prefix);
+          }
+        }
```
dkolsen-pgi, line 160: This part of the change looks correct to me.  There is no way to avoid the extra branching.

dkolsen-pgi (changes requested):

bernhardmgruber:
@charan-003 I added a few more tests.

bernhardmgruber:
@charan-003 I will review this next week.

thrust/testing/transform_scan.cu
```diff
@@ -400,3 +403,145 @@ void TestValueCategoryDeduction()
...
+void TestTransformScanEdgeCases()
+{
+  // Test 1: Large array with transform_inclusive_scan, multiplies, and init
+  {
+    const int n = 10000;
+    thrust::device_vector<int> d_input(n);
+    for (int i = 0; i < n; ++i)
+    {
+      d_input[i] = (i % 5) + 1;
+    }
```
bernhardmgruber: Important: This is **really** slow, since this will do a `cudaMemcpyAsync` for each subscription of `d_input`, so 10000 times. Please use `thrust::tabulate` and a functor to initialize the data.

thrust/testing/transform_scan.cu
```diff
@@ -400,3 +403,145 @@ void TestValueCategoryDeduction()
...
+  }
+
+  // Test 2: transform_exclusive_scan with multiplies on large array
+  {
+    const int n = 10000;
+    thrust::device_vector<int> d_input(n);
+    for (int i = 0; i < n; ++i)
+    {
+      d_input[i] = (i % 3) + 1;
+    }
```
bernhardmgruber: Same here. Applies to more places (also in the other test file)

thrust/thrust/system/omp/detail/scan.h
```diff
@@ -106,7 +106,9 @@ OutputIterator scan_impl(
 
     if (start < n)
     {
-      block_sums[tid] = ::cuda::std::reduce(first + start, first + end, accum_t{}, wrapped_binary_op);
+      // For both has_init and no-init cases: reduce each block using first element as init
+      accum_t first_elem = *(first + start);
+      block_sums[tid]    = ::cuda::std::reduce(first + start + 1, first + end, first_elem, wrapped_binary_op);
```
bernhardmgruber: Important: Let's avoid an extra copy:
```suggestion
      block_sums[tid]    = ::cuda::std::reduce(first + start + 1, first + end, *(first + start), wrapped_binary_op);
```

bernhardmgruber, line 114: I am wondering what happens if `num_threads` becomes larger than `parallel_scan_threshold`. In this case `block_size` becomes zero and `start` and `end` are the same. I think we should catch this sooner where we have:
```c++
if (static_cast<size_t>(n) < parallel_scan_threshold || num_threads <= 1)
```
and do:
```c++
if (static_cast<size_t>(n) < ::cuda::std::max(parallel_scan_threshold, num_threads) || num_threads <= 1)
```

thrust/thrust/system/omp/detail/scan.h
```diff
@@ -117,7 +119,13 @@ OutputIterator scan_impl(
   }
   else
   {
-    ::cuda::std::exclusive_scan(block_sums.begin(), block_sums.end(), block_sums.begin(), accum_t{}, wrapped_binary_op);
+    // For no init inclusive scan: exclusive_scan starting at second element with first element as init
+    if (num_threads > 1)
```
bernhardmgruber: I think this condition is always true, since we handle `num_threads <= 1` serially above. I think we can remove the `if` and just use the body directory. You can of course replace it by an assertion instead:
```suggestion
    _CCCL_ASSERT (num_threads > 1, "");
```

charan-003:
@bernhardmgruber I was wondering if it's possible to perform CI tests locally without using Docker?

bernhardmgruber:
> @bernhardmgruber I was wondering if it's possible to perform CI tests locally without using Docker?

Sure! CCCL is just an ordinary cmake project so you can:
```
mkdir my_build
cd my_build
cmake .. -DTHRUST_ENABLE_MULTICONFIG=OFF
ccmake . # enable what you need, for example Thrust. make sure to set the THRUST_DEVICE_SYSTEM=OMP
make thrust.test.scan
./bin/thrust.test.scan
```

thrust/testing/transform_scan.cu
```diff
@@ -400,3 +404,142 @@ void TestValueCategoryDeduction()
...
+    ASSERT_EQUAL(vec, expected);
+  }
+
+  // Test 5: Boundary case at threshold (1024 elements)
+  {
+    const int n = 1024;
+    thrust::device_vector<int> d_input(n);
+    thrust::tabulate(d_input.begin(), d_input.end(), [](int i) {
+      return i + 1;
+    });
```
bernhardmgruber: I am not sure we enable the use of lambdas in tests by default. We have a few options:
1. turn the lambdas into function objects
2. use `thrust::placeholders` like:
```suggestion
    using thrust::placeholders::_1;
    thrust::tabulate(d_input.begin(), d_input.end(), _1 + 1);
```
3. add an equally named `transform_scan.cmake` file with the content:
```
target_compile_options(
  ${test_target}
  PRIVATE $<$<COMPILE_LANG_AND_ID:CUDA,NVIDIA>:--extended-lambda>
)
```
I am leaning towards 2 or 3.

charan-003: sure, i will be using `thrust::placeholders`.
let me make those changes asap 
thank you :))

bernhardmgruber (approved):

bernhardmgruber:
Thank you so much for taking the time to push the fix through!

bernhardmgruber:
@dkolsen-pgi please re-review and unblock. Thx!

thrust/thrust/system/omp/detail/scan.h
```diff
@@ -117,7 +122,13 @@ OutputIterator scan_impl(
...
```
dkolsen-pgi: You added the assertion as suggested, but forgot to remove this now unnecessary check for `num_threads > 1`.

charan-003: Right, thanks for bringing this up.

PR #6562, opened 2025-11-09
[libcu++] Add initial cccl-runtime docs for 3.1

pciolkosz:
We need some documentation for cccl-runtime APIs released in CCCL 3.1

This PR adds minimal docs for devices, stream, event and algorithm headers. It also adds those headers to Doxyfile.
I wanted to link to the Doxygen API docs, but I couldn't figure out how to do it. That can be fixed with a future change

### DIFF
--- a/docs/libcudacxx/runtime/device.rst
+++ b/docs/libcudacxx/runtime/device.rst
@@ -1,11 +1,11 @@
-.. _libcudacxx-extended-api-runtime-device:
+.. _cccl-runtime-device:
 
 Devices
 =======
 
 ``cuda::device_ref``
 ---------------------
-.. _libcudacxx-extended-api-runtime-device-device-ref:
+.. _cccl-runtime-device-device-ref:
 
 ``cuda::device_ref`` is a lightweight, non-owning handle to a CUDA device ordinal.
 It offers:
@@ -21,7 +21,7 @@ Availability: CCCL 3.1.0 / CUDA 13.1
 
 ``cuda::devices``
 ------------------
-.. _libcudacxx-extended-api-runtime-device-devices:
+.. _cccl-runtime-device-devices:
 
 ``cuda::devices`` is a random-access view of all available CUDA devices in form of ``cuda::device_ref`` objects`. It provides indexing, size, and iteration for use
 in range-based loops.
@@ -43,7 +43,7 @@ Example:
 
 Device attributes
 -----------------
-.. _libcudacxx-extended-api-runtime-device-attributes:
+.. _cccl-runtime-device-attributes:
 
 ``cuda::device_attributes`` provides strongly-typed attribute query objects usable with
 ``device_ref::attribute``. Selected examples:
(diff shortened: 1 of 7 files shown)


docs/libcudacxx/runtime/device.rst
```diff
@@ -0,0 +1,104 @@
...
+
+Availability: CCCL 3.1.0 / CUDA 13.1
+
+Example:
+
+.. code:: cpp
+
+   #include <cuda/devices>
+
+   int get_max_blocks_on_device(cuda::device_ref dev) {
```
pciolkosz, line 65: This example is a bit weak, but I don't know what's a good attributes example

docs/libcudacxx/extended_api/runtime/algorithm.rst
```diff
@@ -0,0 +1,56 @@
+.. _libcudacxx-extended-api-runtime-algorithm:
```
miscco: Important:

This is not an extension to the C++ standard library but a fully new set of features, I believe it would be reasonable to put them into `ccclrt` instead of `extended_api`

I would also change the module name to 
```suggestion
.. _cccl-ccclrt-algorithm:
```

docs/libcudacxx/extended_api/runtime/algorithm.rst
```diff
@@ -0,0 +1,56 @@
...
+
+Algorithms
+==========
+
+The runtime algorithms provide stream-ordered, byte-wise primitives that operate on ``cuda::std::span`` and
```
miscco: I believe we should name the runtime here explicitly
```suggestion
The ``ccclrt`` algorithms provide stream-ordered, byte-wise primitives that operate on ``cuda::std::span`` and
```

docs/libcudacxx/extended_api/runtime/algorithm.rst
```diff
@@ -0,0 +1,56 @@
...
+``cuda::std::mdspan``-compatible types. They require a ``cuda::stream_ref`` to enqueue work.
+
+``cuda::copy_bytes``
+---------------------
+.. _libcudacxx-extended-api-runtime-algorithm-copy_bytes:
+
+Launch a byte-wise copy from source to destination into the provided stream.
```
miscco:
```suggestion
Launches a byte-wise copy from source to destination into the provided stream.
```

Is `into` the right word or should this rather be `on the provided stream`

docs/libcudacxx/extended_api/runtime.rst
```diff
@@ -0,0 +1,78 @@
+.. _libcudacxx-extended-api-runtime:
```
miscco: Lets move all of this one level up into `ccclrt`

docs/libcudacxx/extended_api/runtime/algorithm.rst
pciolkosz: I think algorithm header might provide more things in the future not necessarily from the ccclrt part, so I changed it to highlight this is just a runtime part of that header documented on that page.

In general I don't think we want to mention `ccclrt` or `cccl-runtime` explicitly in the docs. Its runtime subsection of unified CCCL, not a separate library.

PR #6563, opened 2025-11-10
Fix merge conflicts from dropping headers

miscco:


cub/cub/device/dispatch/dispatch_streaming_reduce.cuh
```diff
@@ -192,9 +192,6 @@ struct dispatch_streaming_arg_reduce_t
     InitT init,
     cudaStream_t stream)
   {
-    // Constant iterator to provide the offset of the current partition for the user-provided input iterator
-    using constant_offset_it_t = THRUST_NS_QUALIFIER::constant_iterator<GlobalOffsetT>;
```
davebayer, line 196: Is this not a public type alias? (applies to all other, too)

miscco: Oh wait, I was on the wrong branch

Issue #6558, opened 2025-11-08
[BUG]: Thrust fails to compile when the `THRUST` macro is defined


bernhardmgruber:
### Is this a duplicate?

- [x] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cccl/issues) for this bug and that I agree to the [Code of Conduct](CODE_OF_CONDUCT.md)

### Type of Bug

Compile-time Error

### Component

Thrust

### Describe the bug

Compile with `-DTHRUST=1` and include any Thrust header. BabelStream does that for example.

### How to Reproduce

https://godbolt.org/z/Evea9d7ET

### Expected behavior

Thrust should compile.

PR #6564, opened 2025-11-10
Make Thrust/CUB ABI namespace resilient against user-defined macros

bernhardmgruber:
Example for new ABI namespace name: `thrust::_V_300200_SM_1200::host_vector`

Fixes: #6558

### DIFF
--- a/libcudacxx/include/cuda/std/__internal/namespaces.h
+++ b/libcudacxx/include/cuda/std/__internal/namespaces.h
@@ -95,12 +95,12 @@
 #else // not defined(CCCL_DISABLE_ARCH_DEPENDENT_NAMESPACE)
 #  if defined(_NVHPC_CUDA)
 #    define _CCCL_BEGIN_NAMESPACE_ARCH_DEPENDENT                                                    \
-      inline namespace _CCCL_PP_SPLICE_WITH(_, CCCL, SM, NV_TARGET_SM_INTEGER_LIST, NVHPC) \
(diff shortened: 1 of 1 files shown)


cub/cub/util_namespace.cuh
```diff
@@ -102,13 +102,14 @@
 #  define CUB_DETAIL_MAGIC_NS_END
 #else // not defined(CUB_DISABLE_NAMESPACE_MAGIC)
 #  if defined(_NVHPC_CUDA)
-#    define CUB_DETAIL_MAGIC_NS_BEGIN                                                                  \
-      inline namespace _CCCL_PP_SPLICE_WITH(_, CUB, CUB_VERSION, SM, NV_TARGET_SM_INTEGER_LIST, NVHPC) \
+#    define CUB_DETAIL_MAGIC_NS_BEGIN                                                                                  \
+      inline namespace _CCCL_PP_CAT(                                                                                   \
+        _CCCL_PP_CAT(_CCCL_PP_CAT(_V_, CUB_VERSION), _CCCL_PP_SPLICE_WITH(_, _SM, NV_TARGET_SM_INTEGER_LIST)), _NVHPC) \
```
miscco, line 107: I would say the "simpler" fix would be to just use `_CUB` and not `CUB`

That wont change the actual string in any recognizable way and is a single character change. 

Note we should also change the CCCL macro in `namespace.h`

bernhardmgruber: But a user could also define `SM` or `NVHPC` as macros, which I am trying to handle here as well.

miscco: Oh we can happily add `_` there too

miscco (approved):

PR #6571, opened 2025-11-10
Cleanup dependencies between internal targets.

alliepiper:
1. Add the missing Thrust dependency to CUB.
2. Remove fallback paths that allowed non-sibling CCCL packages to be used when searching for dependencies.
3. Change version checks to exact (rather than minimum).


alliepiper:
@miscco Can you check that this fixes the issue you found in your PSTL PR?

miscco (approved):

PR #4355, opened 2025-04-05
[Thrust]  `reduce_into`

brycelelbach:
```
template <typename InputIt, typename T, typename Op>
T reduce(InputIt first, InputIt last, T init, Op op);

template <typename InputIt, typename OutputIt, typename T, typename Op>
void reduce(InputIt first, InputIt last, OutputIt output, T init, Op op);
```

`reduce_into` is a variant of `reduce` that assigns the reduction result to an iterator. Because it's result is not computation dependent, it is asynchronous when invoked with `par_nosync`. It can be used to avoid transferring the reduction result off the device.

This is the first step in addressing #4056.

PR #6572, opened 2025-11-10
Add missing test for thrust::reduce_into

Pansysk75:
Adds test to `thrust/tests/cuda` for thrust::reduce_into (mentioned here: https://github.com/NVIDIA/cccl/pull/4355#discussion_r2048784124)

### DIFF
--- a/thrust/testing/cuda/reduce_into.cu
+++ b/thrust/testing/cuda/reduce_into.cu
@@ -1,0 +1,123 @@
+#include <thrust/execution_policy.h>
+#include <thrust/iterator/constant_iterator.h>
+#include <thrust/reduce.h>
+
+#include <unittest/unittest.h>
+
+template <typename ExecutionPolicy, typename InputIter, typename OutputIter, typename T>
+__global__ void reduce_into_kernel(ExecutionPolicy exec, InputIter first, InputIter last, OutputIter result, T init)
+{
+  thrust::reduce_into(exec, first, last, result, init);
+}
+
+#ifdef THRUST_TEST_DEVICE_SIDE
+template <typename T, typename ExecutionPolicy>
+void TestReduceIntoDevice(ExecutionPolicy exec, const size_t n)
+{
+  thrust::host_vector<T> h_data   = unittest::random_integers<T>(n);
+  thrust::device_vector<T> d_data = h_data;
+
+  thrust::host_vector<T> h_result(1);
+  thrust::device_vector<T> d_result(1);
+
+  T init = 13;
+
+  thrust::reduce_into(h_data.begin(), h_data.end(), h_result.begin(), init);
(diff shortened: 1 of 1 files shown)


miscco:
Thanks a lot for adding those tests 🎉

thrust/testing/cuda/reduce_into.cu
```diff
@@ -0,0 +1,123 @@
...
+};
+VariableUnitTest<TestReduceIntoDeviceNoSync, IntegralTypes> TestReduceIntoDeviceNoSyncInstance;
+#endif
+
+template <typename ExecutionPolicy>
+void TestReduceIntoCudaStreams(ExecutionPolicy policy)
+{
+  using Vector = thrust::device_vector<int>;
+
+  Vector v = {1, -2, 3};
```
gonidelis, line 70: no need for an intermediate `Vector` alias. I think you can just go with a straightforward `thrust::device_vector<int> v = {1, -2, 3};` instead

thrust/testing/cuda/reduce_into.cu
```diff
@@ -0,0 +1,123 @@
...
+  thrust::device_vector<T> d_data = h_data;
+
+  thrust::host_vector<T> h_result(1);
+  thrust::device_vector<T> d_result(1);
+
+  T init = 13;
+
+  thrust::reduce_into(h_data.begin(), h_data.end(), h_result.begin(), init);
+
+  reduce_into_kernel<<<1, 1>>>(exec, d_data.begin(), d_data.end(), d_result.begin(), init);
```
gonidelis, line 27: curious, we do we need to wrap the thrust call in a kernel? I summon @bernhardmgruber

bernhardmgruber: I think this is testing device side use of the API. This should be fine.

bernhardmgruber (commented):
Thank you for providing new tests! I see they have been added for the CUDA backend only. Can you please also add a few tests in the `thrust/testing` directory, which are backend independent? Thx!

Pansysk75:
> Thank you for providing new tests! I see they have been added for the CUDA backend only. Can you please also add a few tests in the `thrust/testing` directory, which are backend independent? Thx!

If I'm not mistaken, those already exist here:
https://github.com/NVIDIA/cccl/blob/main/thrust/testing/reduce_into.cu

bernhardmgruber:
> > Can you please also add a few tests in the `thrust/testing` directory
> 
> If I'm not mistaken, those already exist here: https://github.com/NVIDIA/cccl/blob/main/thrust/testing/reduce_into.cu

Oh sorry, I missed that. Great!

bernhardmgruber:
Thank you for the contribution!

Issue #6565, opened 2025-11-10
Implement a MVP for `cub::DeviceReduce` using the new tuning API


bernhardmgruber:
Preliminary changes:

- [ ] ~~#6566~~ (not strictly required, we can still go through the dispatcher in the benchmarks)
- [x] #6575
- [x] #6577

PR #6575, opened 2025-11-11
Test and refactor `[Mem|Reg]BoundScaling`

bernhardmgruber:
Split out of #6544, works towards #6565

### DIFF
--- a/cub/cub/device/dispatch/tuning/tuning_transform.cuh
+++ b/cub/cub/device/dispatch/tuning/tuning_transform.cuh
@@ -20,6 +20,7 @@
 #include <thrust/type_traits/is_contiguous_iterator.h>
 #include <thrust/type_traits/is_trivially_relocatable.h>
 
+#include <cuda/__cmath/pow2.h>
 #include <cuda/__cmath/round_up.h>
 #include <cuda/__functional/address_stability.h>
 #include <cuda/std/__algorithm/max.h>


cub/test/catch2_test_util_arch.cu
```diff
@@ -0,0 +1,117 @@
...
+          int ExpectedBlockThreads,
+          int ExpectedItemsPerThread>
+void check_mem_bound_scaling()
+{
+  using mbs = cub::detail::MemBoundScaling<Nominal4ByteBlockThreads, Nominal4ByteItemsPerThread, ComputeT>;
+
+  if constexpr (mbs::ITEMS_PER_THREAD != ExpectedItemsPerThread)
+  {
+    show<mbs::ITEMS_PER_THREAD>::asdf();
+  }
```
bernhardmgruber, line 23: This was mainly used to get the value from the compiler in case the static assertion below fires, but I thought it may be helpful in the future. Let me know if I should remove it.

miscco: Nitpick: I commonly use a deleted function for that

```cpp
template<class>
void print() = delete,
```

The this could be
```suggestion
  if constexpr (mbs::ITEMS_PER_THREAD != ExpectedItemsPerThread)
  {
    print<mbs::ITEMS_PER_THREAD>();
  }
```

PR #6579, opened 2025-11-11
Refactor cub::detail::for_each::dispatch_t

bernhardmgruber:


cub/cub/device/dispatch/dispatch_for.cuh
```diff
@@ -126,45 +113,46 @@ struct dispatch_t
...
+          THRUST_NS_QUALIFIER::cuda_cub::detail::triple_chevron(
+            static_cast<unsigned int>(num_tiles), static_cast<unsigned int>(block_threads), 0, stream)
+            .doit(detail::for_each::static_kernel<typename PolicyHubT::MaxPolicy, OffsetT, OpT>, num_items, op)))
     {
       return error;
     }
 
-    error = CubDebug(detail::DebugSyncStream(stream));
-    if (cudaSuccess != error)
+    if (auto error = CubDebug(detail::DebugSyncStream(stream)))
```
miscco, line 124: Nitpick, we are inconsistent with `if (const auto error` and `if (auto error`

bernhardmgruber: This is on purpose, becaues we assign to `error` again inside the branch. This is to retain the original behavior, which I don't understant why we need that.

PR #6585, opened 2025-11-11
Implement std::normal_distribution

RAMitchell:


bernhardmgruber:
Side note: We also have some implementations of `<random>` in Thrust under `thrust/thrust/random`. I would love if we could replace those by new facilities in libcu++! You could try to replace the Thrust implementation by your new one for testing.

RAMitchell:
https://github.com/NVIDIA/cccl/issues/6595 is preventing us from making this constexpr. We can defer implementing constexpr or I will use another algorithm to generate the normal distribution.

libcudacxx/include/cuda/std/__random/uniform_real_distribution.h
```diff
@@ -45,25 +45,25 @@ class uniform_real_distribution
   public:
     using distribution_type = uniform_real_distribution;
 
-    _CCCL_API explicit param_type(result_type __a = 0, result_type __b = 1) noexcept
+    _CCCL_API constexpr explicit param_type(result_type __a = 0, result_type __b = 1) noexcept
```
fbusato, line 48: nit. As you adding `constexpr` refactoring, prefer `__a_{__a}` syntax

libcudacxx/include/cuda/std/__random/normal_distribution.h
```diff
@@ -0,0 +1,233 @@
...
+//===----------------------------------------------------------------------===//
+
+#ifndef _CUDA_STD___RANDOM_NORMAL_DISTRIBUTION_H
+#define _CUDA_STD___RANDOM_NORMAL_DISTRIBUTION_H
+
+#include <cuda/std/detail/__config>
+
+#include <cuda/std/__random/is_valid.h>
+#include <cuda/std/__random/uniform_real_distribution.h>
+#include <cuda/std/cmath>
```
fbusato: can we use more precise headers?

libcudacxx/include/cuda/std/__random/normal_distribution.h
```diff
@@ -0,0 +1,233 @@
...
+  };
+
+private:
+  param_type __p_{};
+  result_type __v_{};
+  bool __v_hot_{false};
+
+public:
+  _CCCL_API constexpr normal_distribution() noexcept
+      : normal_distribution(0)
```
fbusato:
```suggestion
      : normal_distribution{0}
```

libcudacxx/include/cuda/std/__random/normal_distribution.h
```diff
@@ -0,0 +1,233 @@
...
+  {}
+  _CCCL_API constexpr explicit normal_distribution(result_type __mean, result_type __stddev = 1) noexcept
```
fbusato:
```suggestion
  _CCCL_API constexpr explicit normal_distribution(result_type __mean, result_type __stddev = result_type{1}) noexcept
```

libcudacxx/include/cuda/std/__random/normal_distribution.h
```diff
@@ -0,0 +1,233 @@
...
+      result_type __u = __uni(__g);
+      result_type __v = __uni(__g);
+      result_type __s = __u * __u + __v * __v;
+      while (__s > 1 || __s == 0)
+      {
+        __u = __uni(__g);
+        __v = __uni(__g);
+        __s = __u * __u + __v * __v;
+      }
+      result_type __fp = cuda::std::sqrt(-2 * cuda::std::log(__s) / __s);
```
fbusato:
```suggestion
      const result_type __fp = ::cuda::std::sqrt(-2 * ::cuda::std::log(__s) / __s);
```

libcudacxx/include/cuda/std/__random/normal_distribution.h
```diff
@@ -0,0 +1,233 @@
...
+    static_assert(__cccl_random_is_valid_urng<_URng>, "");
+    result_type __up = 0;
+    if (__v_hot_)
+    {
+      __v_hot_ = false;
+      __up     = __v_;
+    }
+    else
+    {
+      cuda::std::uniform_real_distribution<result_type> __uni(-1, 1);
```
fbusato:
```suggestion
      uniform_real_distribution<result_type> __uni(-1, 1);
```

libcudacxx/include/cuda/std/__random/normal_distribution.h
```diff
@@ -0,0 +1,233 @@
...
+    return __p_;
+  }
+  _CCCL_API constexpr void param(const param_type& __p) noexcept
+  {
+    __p_ = __p;
+  }
+
+  [[nodiscard]] _CCCL_API constexpr result_type min() const noexcept
+  {
+    return -numeric_limits<result_type>::infinity();
```
fbusato, line 158: we don't support extented floating points, right? otherwise, this is problematic

libcudacxx/test/libcudacxx/std/random/distribution/normal.pass.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+template <class T>
+__host__ __device__ void test()
+{
+  // Can be true if/when cuda::std::log is constexpr
+  const bool test_constexpr = false;
+  using D                   = cuda::std::normal_distribution<T>;
+  using P                   = typename D::param_type;
+  using G                   = cuda::std::philox4x64;
+  auto cdf                  = [] __host__ __device__(double x, const P& p) {
```
fbusato: lambdas don't work with NVRTC

RAMitchell: That would explain all the test failures :)

libcudacxx/include/cuda/std/__random/normal_distribution.h
RAMitchell: I've change this to lowest and max. According to the standard min should be "the minimum potentially generated value" so this seems right.

fbusato: `infinity` looks right https://github.com/llvm/llvm-project/blob/main/libcxx/include/__random/normal_distribution.h.
My point was related to the `-` operator that doesn't work in older CTK for `__half/__nv_bfloat16`. I guess these types are out of scope.

libcudacxx/include/cuda/std/__random/normal_distribution.h
```diff
@@ -0,0 +1,234 @@
...
+    return __p_;
+  }
+  _CCCL_API constexpr void param(const param_type& __p) noexcept
+  {
+    __p_ = __p;
+  }
+
+  [[nodiscard]] _CCCL_API constexpr result_type min() const noexcept
+  {
+    return -numeric_limits<result_type>::lowest();
```
miscco: Is that `-` here correct? 
```suggestion
    return numeric_limits<result_type>::lowest();
```

libcudacxx/include/cuda/std/__random/uniform_real_distribution.h
```diff
@@ -45,25 +45,25 @@ class uniform_real_distribution
...
-    [[nodiscard]] _CCCL_API friend bool operator==(const param_type& __x, const param_type& __y) noexcept
+    [[nodiscard]] _CCCL_API friend constexpr bool operator==(const param_type& __x, const param_type& __y) noexcept
     {
       return __x.__a_ == __y.__a_ && __x.__b_ == __y.__b_;
     }
-    [[nodiscard]] _CCCL_API friend bool operator!=(const param_type& __x, const param_type& __y) noexcept
+    [[nodiscard]] _CCCL_API friend constexpr bool operator!=(const param_type& __x, const param_type& __y) noexcept
     {
       return !(__x == __y);
     }
```
miscco, line 70: Nitpick 
```suggestion
#if _CCCL_STD_VER <= 2017
    [[nodiscard]] _CCCL_API friend constexpr bool operator!=(const param_type& __x, const param_type& __y) noexcept
    {
      return !(__x == __y);
    }
#endif // _CCCL_STD_VER <= 2017
```

libcudacxx/include/cuda/std/__random/uniform_real_distribution.h
miscco, line 145:
```suggestion
#if _CCCL_STD_VER <= 2017
  [[nodiscard]] _CCCL_API friend constexpr bool
operator!=(const uniform_real_distribution& __x, const uniform_real_distribution& __y) noexcept
  {
    return !(__x == __y);
  }
#endif // _CCCL_STD_VER <= 2017
```

libcudacxx/test/libcudacxx/std/random/distribution/bernoulli/ctor_param.pass.cpp
```diff
@@ -1,38 +0,0 @@
...
-// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
-// See https://llvm.org/LICENSE.txt for license information.
-// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
-// SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES
-//
-//===----------------------------------------------------------------------===//
-
-// <random>
-
-// class bernoulli_distribution
```
miscco, line 12: Why are those tests deleted?

miscco (changes requested):

libcudacxx/test/libcudacxx/std/random/distribution/bernoulli/ctor_param.pass.cpp
RAMitchell: I've put all tests in test_distribution.h. I can now just specify a CDF and test param sets and it will test everything. This speeds up the 15 or so distributions still to implement, improves consistency and means over 100 source files less.

libcudacxx/include/cuda/std/__random/normal_distribution.h
RAMitchell: Good catch this is incorrect. We are changing it back to inifinity though.

libcudacxx/include/cuda/std/__random/normal_distribution.h
RAMitchell: Changed it back.

RAMitchell:
Not sure what is up with CI.

Question for my reviewers:

I've been adding std::iostream operators so far. Does this make sense? I am not so sure any more, it might be better to just remove them all.

fbusato:
CI had troubles. I reran it.

> I've been adding std::iostream operators so far. Does this make sense? I am not so sure any more, it might be better to just remove them all.

I would really want to remove them, but they are part of the standard. @davebayer @miscco any thought?

libcudacxx/include/cuda/std/__random/normal_distribution.h
```diff
@@ -0,0 +1,234 @@
...
+#endif // !_CCCL_COMPILER(NVRTC)
+
+#include <cuda/std/__cccl/prologue.h>
+
+_CCCL_BEGIN_NAMESPACE_CUDA_STD
+
+template <class _RealType = double>
+class normal_distribution
+{
+  static_assert(__libcpp_random_is_valid_realtype<_RealType>, "RealType must be a supported floating-point type");
```
davebayer, line 40: use `__cccl` prefix, please, `__libcpp` belongs to libc++

libcudacxx/include/cuda/std/__random/normal_distribution.h
```diff
@@ -0,0 +1,234 @@
...
+  result_type __v_{};
+  bool __v_hot_{false};
+
+public:
+  _CCCL_API constexpr normal_distribution() noexcept
+      : normal_distribution{0}
+  {}
+  _CCCL_API constexpr explicit normal_distribution(result_type __mean, result_type __stddev = result_type{1}) noexcept
+      : __p_{param_type(__mean, __stddev)}
+      , __v_hot_{false}
```
davebayer: This is unnecessary since you define the member with a default value
```suggestion
```

libcudacxx/include/cuda/std/__random/normal_distribution.h
```diff
@@ -0,0 +1,234 @@
...
+  {}
+  _CCCL_API constexpr explicit normal_distribution(const param_type& __p) noexcept
+      : __p_{__p}
+      , __v_hot_{false}
```
davebayer:
```suggestion
```

libcudacxx/include/cuda/std/__random/normal_distribution.h
```diff
@@ -0,0 +1,234 @@
...
+  // generating functions
+  template <class _URng>
+  _CCCL_API constexpr result_type operator()(_URng& __g)
+  {
+    return (*this)(__g, __p_);
+  }
+  template <class _URng>
+  _CCCL_API constexpr result_type operator()(_URng& __g, const param_type& __p)
+  {
+    static_assert(__cccl_random_is_valid_urng<_URng>, "");
```
davebayer: Can we provide a better error message?

libcudacxx/test/support/random_utilities/test_distribution.h
```diff
@@ -0,0 +1,281 @@
...
+//===----------------------------------------------------------------------===//
+#ifndef LIBCUDACXX_TEST_SUPPORT_RANDOM_UTILITIES_TEST_DISTRIBUTION_H
+#define LIBCUDACXX_TEST_SUPPORT_RANDOM_UTILITIES_TEST_DISTRIBUTION_H
+#include <cuda/std/__algorithm/partial_sort.h>
+#include <cuda/std/__memory_>
+#include <cuda/std/array>
+#include <cuda/std/cstddef>
+#if !_CCCL_COMPILER(NVRTC)
+#  include <sstream>
+#endif // !_CCCL_COMPILER(NVRTC)
```
davebayer, line 20:
```suggestion
#define LIBCUDACXX_TEST_SUPPORT_RANDOM_UTILITIES_TEST_DISTRIBUTION_H

#include <cuda/std/__algorithm/partial_sort.h>
#include <cuda/std/__memory_>
#include <cuda/std/array>
#include <cuda/std/cstddef>

#if !_CCCL_COMPILER(NVRTC)
#  include <sstream>
#endif // !_CCCL_COMPILER(NVRTC)
```

davebayer:
> I would really want to remove them, but they are part of the standard. @davebayer @miscco any thought?

We implement them for all other `cuda::std::` stuff, so I think we should stay consistent and implement them

libcudacxx/include/cuda/std/__random/normal_distribution.h
RAMitchell: I am just using whats already in libcudacxx/include/cuda/std/__random/is_valid.h. I don't think this PR is the right place to change everything in there.

davebayer: Okay, we should do it in a separate PR then :)

davebayer (approved):

RAMitchell:
> > I would really want to remove them, but they are part of the standard. @davebayer @miscco any thought?
> 
> We implement them for all other `cuda::std::` stuff, so I think we should stay consistent and implement them

I will continue implementing them then.

@miscco @fbusato any further changes?

This PR is a template for how I intend to do the remaining distributions, in particular with regards to testing. Once this is merged I will open several concurrent PRs for other distributions, which can be reviewed in parallel.

fbusato (approved):

PR #6586, opened 2025-11-11
Disable clang-cuda with libc++  tests for now

miscco:


ci/matrix.yaml
```diff
@@ -78,7 +78,7 @@ workflows:
     - {jobs: ['build'], cudacxx: 'clang', ctk: 'clang-cuda', cxx: 'clang-cuda',  std: 'all', sm: '75;80;90;100'}
     # libc++
     # - arm64 for now as it's closest to android.
-    - {jobs: ['build'], cpu: 'arm64', project: 'libcudacxx', std: 'all', cxx: 'clang', cmake_options: '-DCCCL_USE_LIBCXX=ON', sm: '75;80;90;100'}
+    # - {jobs: ['build'], cpu: 'arm64', project: 'libcudacxx', std: 'all', cudacxx: 'clang', ctk: 'clang-cuda', cxx: 'clang-cuda', cmake_options: '-DCCCL_USE_LIBCXX=ON', sm: '75;80;90;100'}
```
bernhardmgruber, line 81: Q: Are the changes appart from commenting out correct or should those be reverted?

bernhardmgruber, line 81: Important: Please add a comment that this is disabled until **something** with an appropriate something. Please also open a tracking issue.

miscco: Yes those are correct, I tested locally and we cannot include anything that might include the runtime or any other STL header with nvcc and libc++

The reason being that modern libc++ only supports the last 2 major clang versions.

With that it requires availability of a lot f compiler builtins for type traits that are not available in nvcc

miscco: I have an explorative [PR ](https://github.com/NVIDIA/cccl/pull/6592) open, but that will require some work

bernhardmgruber: Can you still please add a comment why the run is disabled? So nobody deletes it by accident.

bernhardmgruber (approved):

PR #6591, opened 2025-11-12
Refactor rfa dispatcher

bernhardmgruber:
Moved it from `cub::detail` to `cub::detail::rfa` and renamed it to `dispatch_t`. Also condensed error handling a bit.

### DIFF
--- a/cub/cub/device/dispatch/dispatch_reduce_deterministic.cuh
+++ b/cub/cub/device/dispatch/dispatch_reduce_deterministic.cuh
@@ -193,18 +193,12 @@ struct dispatch_t
     THRUST_NS_QUALIFIER::cuda_cub::detail::triple_chevron(1, ActivePolicyT::SingleTilePolicy::BLOCK_THREADS, 0, stream)
       .doit(single_tile_kernel, d_in, d_out, static_cast<int>(num_items), reduction_op, init, transform_op);
     // Check for failure to launch
-    auto error = CubDebug(cudaPeekAtLastError());
-    if (cudaSuccess != error)
+    if (const auto error = CubDebug(cudaPeekAtLastError()))
     {
       return error;
     }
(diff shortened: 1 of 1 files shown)


cub/cub/device/dispatch/dispatch_reduce_deterministic.cuh
```diff
@@ -46,9 +46,7 @@
 
 CUB_NAMESPACE_BEGIN
 
-namespace detail
-{
-namespace rfa
+namespace detail::rfa
```
NaderAlAwar, line 49: Suggestion: It is a little annoying to me that in #6593 we moved the dispatcher to `detail::reduce::`, but rfa uses its own namespace. Could this be moved to `detail::reduce::` as well? Might be some name collisions I haven't considered

bernhardmgruber: I don't mind. We already have the namespace `cub::detail::rfa` so it seemed like the fitting solution. The tunings are also split into the `cub::detail::reduce` namespace for the general reduction, segmented reduction, and non-deterministic reduction. Tunings for the deterministic reduction use the namespace `cub::detail::rfa`. There is also `cub::detail::fixed_size_segmented_reduce` for tunings for the fixed size version of the segmented reduction. Moving `DispatchReduceDeterministic` into the `rfa` namespace is thus increasing consistency, because the dispatcher now lives in the same namespace as the tunings.

Do you propose to remove the `cub::detail::rfa` namespace entirely and move everything into `cub::detail::reduce`? Then we need to rename the tunings and policy hubs for rfa to avoid collisions with the other reduction algorithms.

NaderAlAwar: On second thought this is not really necessary

PR #6593, opened 2025-11-12
Refactor deterministic reduce dispatcher

bernhardmgruber:
Moved it from `cub::detail` to `cub::detail::reduce` and renamed it to `dispatch_t`. Also condensed error handling a bit.

### DIFF
--- a/cub/cub/device/dispatch/dispatch_reduce_nondeterministic.cuh
+++ b/cub/cub/device/dispatch/dispatch_reduce_nondeterministic.cuh
@@ -182,7 +182,7 @@ struct dispatch_nondeterministic_t
 
     const int reduce_device_occupancy = reduce_config.sm_occupancy * sm_count;
     // Even-share work distribution
-    int max_blocks = reduce_device_occupancy * subscription_factor;
+    const int max_blocks = reduce_device_occupancy * subscription_factor;
     GridEvenShare<OffsetT> even_share;
     even_share.DispatchInit(num_items, max_blocks, reduce_config.tile_size);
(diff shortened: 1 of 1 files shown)


cub/benchmarks/bench/reduce/nondeterministic.cu
```diff
@@ -58,7 +58,7 @@ void nondeterministic_sum(nvbench::state& state, nvbench::type_list<T, OffsetT>)
   using output_t    = T;
   using op_t        = cuda::std::plus<>;
   using init_t      = T;
-  using dispatch_t  = cub::detail::DispatchReduceNondeterministic<
+  using dispatch_t  = cub::detail::reduce::dispatch_nondet_t<
```
miscco: I would prefer 
```suggestion
  using dispatch_t  = cub::detail::reduce::dispatch_nondeterministic_t<
```

But nothing I am crazy about

cub/cub/device/dispatch/dispatch_reduce_nondeterministic.cuh
```diff
@@ -185,7 +182,7 @@ struct DispatchReduceNondeterministic
 
     const int reduce_device_occupancy = reduce_config.sm_occupancy * sm_count;
     // Even-share work distribution
-    int max_blocks = reduce_device_occupancy * detail::subscription_factor;
+    int max_blocks = reduce_device_occupancy * subscription_factor;
```
NaderAlAwar: Suggestion: could be made `const`

PR #6596, opened 2025-11-12
Refactor DispatchScanByKey

bernhardmgruber:


cub/cub/device/dispatch/dispatch_scan_by_key.cuh
```diff
@@ -332,142 +332,122 @@ struct DispatchScanByKey
...
-      }
+      return error;
+    }
 
-      // Number of input tiles
-      int tile_size = Policy::BLOCK_THREADS * Policy::ITEMS_PER_THREAD;
-      int num_tiles = static_cast<int>(::cuda::ceil_div(num_items, tile_size));
+    // Number of input tiles
+    const int tile_size = Policy::BLOCK_THREADS * Policy::ITEMS_PER_THREAD;
+    const int num_tiles = static_cast<int>(::cuda::ceil_div(num_items, tile_size));
```
miscco, line 344: Nitpick: should be constexpr?

bernhardmgruber: They could be for now, but will fall back to `const` again when we introduce the new tuning API. I would let them be for now.

PR #6600, opened 2025-11-12
Refactor DispatchUniqueByKey

bernhardmgruber:


cub/cub/device/dispatch/dispatch_unique_by_key.cuh
```diff
@@ -236,183 +236,161 @@ struct DispatchUniqueByKey
...
     {
-      // Get device ordinal
-      int device_ordinal;
-      error = CubDebug(cudaGetDevice(&device_ordinal));
-      if (cudaSuccess != error)
-      {
-        break;
-      }
+      return error;
+    }
```
miscco: We should definitely use the new cuda runtime for that @pciolkosz

elstehle: I'd be ok to drop querying `device_ordinal` here altogether. It's used to later on query `cudaDevAttrMaxGridDimX`. Iirc, that's consistently `INT_MAX` so far(?) and our agent implementation is limited to supporting `int` as a linearized block id. So, I'd be ok dropping these parts.

elstehle (approved):

cub/cub/device/dispatch/dispatch_unique_by_key.cuh
bernhardmgruber: I just discovered something even better:
```c++
const int max_dim_x = cuda::arch_traits_for(cuda::compute_capability{ptx_version}).max_grid_dim_x;
```
arch traits <3

bernhardmgruber: Which are nice, but wrong, since we are interested in the current device and not the one we compiled for.

bernhardmgruber: Changed to use just `INT_MAX`.

PR #6605, opened 2025-11-12
Rework native hierarchy levels and queries

davebayer:
This PR implements the first part of hierarchy levels support in libcu++. It implements the native hierarchy levels:
- `cuda::device::grid`
- `cuda::device::cluster`
- `cuda::device::block`
- `cuda::device::warp`
- `cuda::device::thread`

and operations `dims`, `static_dims`, `extents`, `count`, `index` and `rank` on them.

### DIFF
--- a/libcudacxx/include/cuda/__hierarchy/block_level.h
+++ b/libcudacxx/include/cuda/__hierarchy/block_level.h
@@ -1,0 +1,78 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of libcu++, the C++ Standard Library for your entire system,
+// under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef _CUDA___HIERARCHY_BLOCK_LEVEL_H
+#define _CUDA___HIERARCHY_BLOCK_LEVEL_H
+
+#include <cuda/std/detail/__config>
+
+#if defined(_CCCL_IMPLICIT_SYSTEM_HEADER_GCC)
+#  pragma GCC system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_CLANG)
+#  pragma clang system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_MSVC)
+#  pragma system_header
+#endif // no system header
+
+#include <cuda/__fwd/hierarchy.h>
+#include <cuda/__hierarchy/hierarchy_query_result.h>
+#include <cuda/__hierarchy/native_hierarchy_level_base.h>
+#include <cuda/std/__mdspan/extents.h>
+
+#include <cuda/std/__cccl/prologue.h>
+
+_CCCL_BEGIN_NAMESPACE_CUDA
+
+struct block_level : __native_hierarchy_level_base<block_level>
+{
+  using __next_level = cluster_level;
+
+  using __base_type = __native_hierarchy_level_base<block_level>;
+  using __base_type::extents;
+  using __base_type::index;
+
+  // interactions with cluster level
+
+  [[nodiscard]] _CCCL_DEVICE_API static ::cuda::std::dims<3, unsigned> extents(const cluster_level&) noexcept
+  {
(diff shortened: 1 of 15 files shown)


libcudacxx/test/libcudacxx/cuda/hierarchy/thread_level/implicit_hierarchy.pass.cpp
```diff
@@ -0,0 +1,196 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of the libcu++ Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES.
+//
+//===----------------------------------------------------------------------===//
+
+#include <cuda/__hierarchy_>
```
davebayer: I implemented tests for thread level only for now, to get feedback on the test structure. Once approved, I'll add tests for all other levels as well.

libcudacxx/include/cuda/__hierarchy/hierarchy_query_result.h
```diff
@@ -0,0 +1,284 @@
...
+#include <cuda/std/__cstddef/types.h>
+#include <cuda/std/__type_traits/is_convertible.h>
+#include <cuda/std/__type_traits/is_nothrow_convertible.h>
+#include <cuda/std/__type_traits/is_same.h>
+
+#include <cuda/std/__cccl/prologue.h>
+
+_CCCL_BEGIN_NAMESPACE_CUDA
+
+template <class _Tp, ::cuda::std::size_t _Np>
```
pciolkosz, line 35: I don't know how useful this is right now. We either have to support dimensionality other than 3 in all operations and return appropriately from the level factory functions, or we just continue to return 3 dimensions from factories. The changes here are insufficient.
Also is it important to support different number of dimensions instead of marking unused ones as statically 1? I don't know

libcudacxx/include/cuda/__hierarchy/thread_level.h
```diff
@@ -0,0 +1,74 @@
...
+#include <cuda/__hierarchy/native_hierarchy_level_base.h>
+#include <cuda/std/__mdspan/extents.h>
+
+#include <cuda/std/__cccl/prologue.h>
+
+_CCCL_BEGIN_NAMESPACE_CUDA
+
+struct thread_level : __native_hierarchy_level_base<thread_level>
+{
+  using __next_level = block_level;
```
pciolkosz, line 35: Next level will be insufficient to validate hierarchies

libcudacxx/include/cuda/__hierarchy/warp_level.h
```diff
@@ -0,0 +1,62 @@
...
+  using __base_type::extents;
+  using __base_type::index;
+
+  [[nodiscard]] _CCCL_DEVICE_API static ::cuda::std::dims<1, unsigned> extents(const block_level&) noexcept
+  {
+    return ::cuda::std::dims<1, unsigned>{::cuda::ptx::get_sreg_nwarpid()};
+  }
+  [[nodiscard]] _CCCL_DEVICE_API static hierarchy_query_result<unsigned, 1> index(const warp_level&) noexcept
+  {
+    return {::cuda::ptx::get_sreg_warpid()};
```
pciolkosz, line 48: We shouldn't use ptx for any hierarchy related things, because its a black box, the compiler can't reuse subexpressions or optimize the code based on results

libcudacxx/include/cuda/__hierarchy/block_level.h
```diff
@@ -0,0 +1,78 @@
...
+  }
+  [[nodiscard]] _CCCL_DEVICE_API static hierarchy_query_result<unsigned, 3> index(const grid_level&) noexcept
+  {
+    return {blockIdx.z, blockIdx.y, blockIdx.x};
+  }
+};
+
+_CCCL_END_NAMESPACE_CUDA
+
+_CCCL_BEGIN_NAMESPACE_CUDA_DEVICE
```
pciolkosz, line 70: You might need to refactor all of the hierarchy testing to this new namespace only to refactor it back in a week

pciolkosz (commented):
It's in general hard to review this reimplementation from scratch. I think I would either need the changes integrated to the old implementation and testing to see if we don't regress, which would also mean extra work to make each step integrated to the old implementation.
Or you could replace everything in one go, but then we you would need to implement the whole thing in one go.
Otherwise I see a different implementation, but is it better than the old? Hard to say

Issue #6601, opened 2025-11-12
[BUG] `cuda::memcpy_async` hangs in some examples


bernhardmgruber:
Some (internal) `cuda::memcpy_async` examples began to hang after #5996.

Example: https://godbolt.org/z/5n37zsvs7
```c++
__global__ void kernel_pipeline(double* __restrict__ ary) {
    __shared__ alignas(16) double val[16];
    __shared__ cuda::barrier<cuda::thread_scope_block> bar;

    if (threadIdx.x == 0) {
        init(&bar, 32);
    }
    __syncthreads();

    if (threadIdx.x == 0) {
        cuda::memcpy_async(val, ary,
            cuda::aligned_size_t<16>(16 * sizeof(double)),
            bar);
    }

    bar.arrive_and_wait();
}
```

bernhardmgruber:
@ahendriksen found the root cause of the hang, which stems from ` cuda::memcpy_async` performing an `cuda::ptx::elect_sync` with mask `~0` internally, implying all threads in the warp participate. However, `cuda::memcpy_async` is only called by a single thread.

The problematic line is here: https://github.com/NVIDIA/cccl/pull/5996/files#diff-653687f3174c2a4fc5f8fc4f318aaefe94a023d6fd6bca5b7924feb59e0632bcR48

The problem does not seem to occur on all nvcc versions, but it does occur with ToT nvcc.

bernhardmgruber:
The higher level problem I have with this bug is that I wasn't sure whether the example above is actually legal. We have zero tests in libcu++, where `cuda::memcpy_async` is not called with all threads in the block. I assumed that the algorithm has to be called cooperatively. However, I have since been explained that only the overloads taking a group parameter are cooperative, and the overloads without are not. This should be better documented and tested.

Issue #5995, opened 2025-09-23
`cuda::memcpy_async` with `cuda::barrier` implementation is inefficient on sm90+


bernhardmgruber:
I received the following piece of code (shortened) using `cuda::memcpy_async` and its author claims the codegen is inefficient:
```c++
#include <cuda/barrier>
#include <cuda/ptx>
#include <cooperative_groups.h>

__device__ static inline bool elect_sync(const ::cuda::std::uint32_t& __membermask) { ... }

constexpr int elem_per_block = 128;
constexpr int num_bytes = elem_per_block * sizeof(float);

using barrier = cuda::barrier<cuda::thread_scope_block>;
using aligned_16 = cuda::aligned_size_t<16>;
namespace cg = cooperative_groups;

__global__ void tma_add(
  const float* a, const float* b, float* c, size_t len) {
    __shared__ alignas(16) float smem_a[elem_per_block];
    __shared__ alignas(16) float smem_b[elem_per_block];
    __shared__ barrier bar;
    auto block = cg::this_thread_block();

    if (threadIdx.x == 0) {
        init(&bar, blockDim.x);
    }
    __syncthreads();

    cuda::memcpy_async(
        block,
        smem_a,
        a,
        aligned_16(num_bytes),
        bar
    );
    bar.arrive_and_wait();
}
```
Full version here: https://godbolt.org/z/jGz884cE5

If the call to `cuda::memcpy_async` is replaced with
```c++
    int warp_id = threadIdx.x / 32;
    bool elected = warp_id == 0 && elect_sync(~0);
    if (elected) {
      cuda::ptx::cp_async_bulk(cuda::ptx::space_cluster, cuda::ptx::space_global, 
        smem_a, a, num_bytes, cuda::device::barrier_native_handle(bar));
      cuda::ptx::mbarrier_expect_tx(cuda::ptx::sem_relaxed, cuda::ptx::scope_cta, cuda::ptx::space_shared,
        cuda::device::barrier_native_handle(bar), num_bytes);
    }
```
the codegen is much better. So what's wrong with `cuda::memcpy_async`?

After some investigation I noticed the following problems:

- [x] our `__cp_async_bulk_shared_global` elects with `__g.thread_rank() == 0` and not with `elect_sync`, leading to bad codegen because the compiler does not recognize the uniform data path

[...]

PR #6608, opened 2025-11-12
Fix `cuda::memcpy async` edge cases and add more tests

bernhardmgruber:
- [x] Broken example in #6601 does not hang anymore
- [x] Codegen for the example in #5995 is still optimal, we just have more code now for computing the thread rank of the CG group
- [x] Codegen of the above using a custom 1D thread block and `is_thread_block_group_v` optimal

I pulled the core fix out into #6710, so it can ship on time for 3.2.

Fixes: #6601

### DIFF
--- a/libcudacxx/test/libcudacxx/CMakeLists.txt
+++ b/libcudacxx/test/libcudacxx/CMakeLists.txt
@@ -281,7 +281,7 @@ add_test(
 set_tests_properties(
   libcudacxx.test.lit
   PROPERTIES
-    # 3h30, some CI machines are slow
-    TIMEOUT 12600
+    # 3hr, some CI machines are slow
+    TIMEOUT 10800
     RUN_SERIAL TRUE
 )


libcudacxx/include/cuda/__memcpy_async/cp_async_bulk_shared_global.h
```diff
@@ -37,18 +37,52 @@
...
+__elect_from_group(const cooperative_groups::thread_block& __g) noexcept
 {
-  // cooperative groups maps a multidimensional thread id into the thread rank the same way as warps do
-  const unsigned int tid             = __g.thread_rank();
+  // Cannot call __g.thread_rank(), because we only forward declared the thread_block type
+  // cooperative groups (and we here) maps a multidimensional thread id into the thread rank the same way as warps do
+  const unsigned int tid             = threadIdx.z * blockDim.y * blockDim.x + threadIdx.y * blockDim.x + threadIdx.x;
   const unsigned int warp_id         = tid / 32;
   const unsigned int uniform_warp_id = __shfl_sync(0xFFFFFFFF, warp_id, 0); // broadcast from lane 0
   return uniform_warp_id == 0 && ::cuda::ptx::elect_sync(0xFFFFFFFF); // elect a leader thread among warp 0
```
bernhardmgruber: The old logic is wrong for any `_Group` that is not a full thread block.

libcudacxx/include/cuda/__memcpy_async/cp_async_bulk_shared_global.h
```diff
@@ -37,18 +37,52 @@
...
+
+// elect from the whole thread block
+[[nodiscard]] _CCCL_DEVICE _CCCL_FORCEINLINE bool
+__elect_from_group(const cooperative_groups::thread_block& __g) noexcept
 {
-  // cooperative groups maps a multidimensional thread id into the thread rank the same way as warps do
-  const unsigned int tid             = __g.thread_rank();
+  // Cannot call __g.thread_rank(), because we only forward declared the thread_block type
+  // cooperative groups (and we here) maps a multidimensional thread id into the thread rank the same way as warps do
+  const unsigned int tid             = threadIdx.z * blockDim.y * blockDim.x + threadIdx.y * blockDim.x + threadIdx.x;
```
bernhardmgruber: @pciolkosz if we had a `cooperative_groups::thread_block<1>` or some other way to detect that the block is 1D, we could save a lot of special register reads here!

bernhardmgruber: Alternatively, we could just add a `cuda::thread_block_group<1>` which would fulfill the `Group` concept and give us efficient codegen here. @miscco  and @pciolkosz what do you think?

libcudacxx/include/cuda/__memcpy_async/cp_async_bulk_shared_global.h
```diff
@@ -37,16 +37,50 @@
 
 #    include <cuda/std/__cccl/prologue.h>
 
+namespace cooperative_groups
+{
+namespace __v1
```
miscco: Is that meant to be an inline namespace?

libcudacxx/include/cuda/__memcpy_async/cp_async_bulk_shared_global.h
```diff
@@ -37,16 +37,50 @@
...
+using namespace __v1;
+} // namespace cooperative_groups
+
 _CCCL_BEGIN_NAMESPACE_CUDA
 
+template <typename _Group>
+[[nodiscard]] _CCCL_DEVICE _CCCL_FORCEINLINE unsigned int __thread_rank(const _Group& __g)
+{
+  return __g.thread_rank();
+}
```
miscco: Is that a leftover from an earlier version or intended to replace the calculation of `__tid` below?

libcudacxx/include/cuda/__memcpy_async/cp_async_bulk_shared_global.h
```diff
@@ -37,16 +37,50 @@
...
+  return __uniform_warp_id == 0 && ::cuda::ptx::elect_sync(0xFFFFFFFF); // elect a leader thread among warp 0
+}
+
+// elect from a single warp
+template <typename Parent>
+[[nodiscard]] _CCCL_DEVICE _CCCL_FORCEINLINE bool
+__elect_from_group(const cooperative_groups::thread_block_tile<32, Parent>&) noexcept
+{
+  return ::cuda::ptx::elect_sync(0xFFFFFFFF); // elect a leader thread among warp 0
+}
```
miscco: question: is there a trait that we could use to sniff out a thread group and then use that inside an `if constexpr`?

libcudacxx/test/libcudacxx/cuda/memcpy_async/group_memcpy_async.h
```diff
@@ -65,10 +68,10 @@ struct storage
...
 
 #if !TEST_COMPILER(NVRTC) && !TEST_COMPILER(CLANG)
-static_assert(std::is_trivially_copy_constructible<storage<int8_t>>::value, "");
-static_assert(std::is_trivially_copy_constructible<storage<uint16_t>>::value, "");
-static_assert(std::is_trivially_copy_constructible<storage<int32_t>>::value, "");
-static_assert(std::is_trivially_copy_constructible<storage<uint64_t>>::value, "");
+static_assert(std::is_trivially_copy_constructible_v<storage<int8_t>>, "");
+static_assert(std::is_trivially_copy_constructible_v<storage<uint16_t>>, "");
+static_assert(std::is_trivially_copy_constructible_v<storage<int32_t>>, "");
+static_assert(std::is_trivially_copy_constructible_v<storage<uint64_t>>, "");
```
miscco: Please use `cuda::std`
```suggestion
static_assert(cuda::std::is_trivially_copy_constructible_v<storage<int8_t>>, "");
static_assert(cuda::std::is_trivially_copy_constructible_v<storage<uint16_t>>, "");
static_assert(cuda::std::is_trivially_copy_constructible_v<storage<int32_t>>, "");
static_assert(cuda::std::is_trivially_copy_constructible_v<storage<uint64_t>>, "");
```

libcudacxx/include/cuda/__memcpy_async/cp_async_bulk_shared_global.h
bernhardmgruber: No

From: `cooperative_groups/details/info.h`:
```c++
#define _CG_CONCAT_INNER(x, y) x ## y
#define _CG_CONCAT_OUTER(x, y) _CG_CONCAT_INNER(x, y)
#define _CG_NAMESPACE _CG_CONCAT_OUTER(__v, _CG_ABI_VERSION)

#define _CG_BEGIN_NAMESPACE \
    namespace cooperative_groups { namespace _CG_NAMESPACE {
#define _CG_END_NAMESPACE \
    }; using namespace _CG_NAMESPACE; };
```

libcudacxx/include/cuda/__memcpy_async/cp_async_bulk_shared_global.h
```diff
@@ -37,16 +37,55 @@
...
+class thread_block_tile;
+} // namespace __v1
+using namespace __v1;
+} // namespace cooperative_groups
+
 _CCCL_BEGIN_NAMESPACE_CUDA
 
+//! Trait to detect whether a group represents a CUDA thread block, for example: ``cooperative_groups::thread_block``.
+template <typename T>
+_CCCL_GLOBAL_CONSTANT bool is_thread_block_group_v = false;
```
bernhardmgruber: New public trait. Should we name it `is_thread_block_group` instead? Bikeshedding also welcome.

libcudacxx/test/libcudacxx/cuda/memcpy_async/group_memcpy_async.h
```diff
@@ -79,52 +82,90 @@ template <class T,
...
+  if (threadIdx.x < 8)
+  {
+    static_assert(thread_block_size >= 8);
+    printf("%u copying 8 items at meta group rank %u\n", threadIdx.x, tiled_groups.meta_group_rank());
+    cuda::memcpy_async(
+      tiled_groups,
+      &dest->data[tiled_groups.meta_group_rank() * 16],
+      &source->data[tiled_groups.meta_group_rank() * 16],
+      sizeof(T) * 8,
+      *bar);
```
bernhardmgruber, line 144: Remark: the possibility of this is incredibly clever and unholy at the same time.

libcudacxx/test/libcudacxx/cuda/memcpy_async/group_memcpy_async_16b.pass.cpp
```diff
@@ -14,9 +14,9 @@
 
 int main(int argc, char** argv)
 {
-  NV_IF_TARGET(NV_IS_HOST, cuda_thread_count = 4;)
+  NV_IF_TARGET(NV_IS_HOST, cuda_thread_count = thread_block_size;)
 
-  test_select_source<storage<uint16_t>>();
+  test_select_source<uint16_t>();
```
bernhardmgruber: Just for my understanding: libcu++ unit tests are called twice, once with the real `main` to execute on the host, and once where `main` is executed inside a kernel on the device. And the host execution comes first, so when setting the global variable `cuda_thread_count` it influences how many threads the following device execution will  use?

miscco: that is correct and totally normal

libcudacxx/include/cuda/__memcpy_async/cp_async_bulk_shared_global.h
miscco: I would question whether we want to make it public, but I believe the `_v` better marks it as an inline variable as opposed to the old school traits

docs/libcudacxx/extended_api/asynchronous_operations/memcpy_async.rst
```diff
@@ -133,6 +144,26 @@ Parameters
...
+--------------
+
+.. code:: cuda
+
+   template <typename T>
+   constexpr inline bool is_thread_block_group_v;
+
+This trait is ``true`` if ``T`` represents a CUDA thread block.
+For example, ``cooperative_groups::thread_block`` satisfies this trait.
+Users are encouraged to specialize this trait for their own groups.
```
miscco, line 157: I believe we should make clear that this talks about a full thread group and not just a single thread?

This was the original bug wasnt it?

libcudacxx/include/cuda/__memcpy_async/cp_async_bulk_shared_global.h
miscco: 🙀

libcudacxx/include/cuda/__memcpy_async/cp_async_bulk_shared_global.h
```diff
@@ -37,16 +37,55 @@
...
+template <>
+_CCCL_GLOBAL_CONSTANT bool is_thread_block_group_v<cooperative_groups::thread_block> = true;
+
+//! Trait to detect whether a group represents a CUDA warp, for example:
+//! ``cooperative_groups::thread_block_tile<32, ...>``.
+template <typename T>
+_CCCL_GLOBAL_CONSTANT bool is_warp_group_v = false;
+
+template <typename Parent>
+_CCCL_GLOBAL_CONSTANT bool is_warp_group_v<cooperative_groups::thread_block_tile<32, Parent>> = true;
```
miscco: I would prefer it we would pull this out into its own header.

I could imagine a world were we want this in other places too

libcudacxx/test/libcudacxx/cuda/memcpy_async/group_memcpy_async.h
```diff
@@ -65,10 +68,10 @@ struct storage
 };
 
 #if !TEST_COMPILER(NVRTC) && !TEST_COMPILER(CLANG)
-static_assert(std::is_trivially_copy_constructible<storage<int8_t>>::value, "");
-static_assert(std::is_trivially_copy_constructible<storage<uint16_t>>::value, "");
-static_assert(std::is_trivially_copy_constructible<storage<int32_t>>::value, "");
-static_assert(std::is_trivially_copy_constructible<storage<uint64_t>>::value, "");
+static_assert(cuda::std::is_trivially_copy_constructible_v<storage<int8_t>>, "");
```
miscco, line 72: Missing `#include <cuda/std/type_traits>`

docs/libcudacxx/extended_api/asynchronous_operations/memcpy_async.rst
bernhardmgruber: Yes. How would you like to improve the documentation? I am already happy with it. What's missing or unclear?

libcudacxx/include/cuda/__memcpy_async/cp_async_bulk_shared_global.h
bernhardmgruber: Moved to a new header.

libcudacxx/test/libcudacxx/cuda/memcpy_async/group_memcpy_async_16b.pass.cpp
bernhardmgruber: Thx!

libcudacxx/include/cuda/__memcpy_async/cp_async_bulk_shared_global.h
bernhardmgruber: I added this trait now.

libcudacxx/include/cuda/__memcpy_async/cp_async_bulk_shared_global.h
bernhardmgruber: yes, removed.

docs/libcudacxx/extended_api/asynchronous_operations/memcpy_async.rst
miscco: Something like `full CUDA thread block` or something that indicates that we need all dimensions?

bernhardmgruber: I updated the wording to spell `if Group represents the full CUDA thread block`. It does not matter which dimensionality the thread block has, this is abstracted by `__g.thread_rank()`.

docs/libcudacxx/extended_api/asynchronous_operations/memcpy_async.rst
```diff
@@ -91,14 +91,25 @@ namely:
      the behavior is undefined.
    - If the objects are not of `TriviallyCopyable <https://en.cppreference.com/w/cpp/named_req/TriviallyCopyable>`_
      type the program is ill-formed, no diagnostic required.
+
+Additionally:
+
    - If *Shape* is :ref:`cuda::aligned_size_t <libcudacxx-extended-api-memory-aligned-size>`, ``source``
```
fbusato, line 97: question. Are these constraints evaluated in assertions?

libcudacxx/include/cuda/__memcpy_async/elect_one.h
```diff
@@ -45,6 +48,24 @@ _CCCL_BEGIN_NAMESPACE_CUDA_DEVICE
...
+template <typename _Group>
+[[nodiscard]] _CCCL_DEVICE _CCCL_FORCEINLINE bool __group_elect_one(const _Group& __g) noexcept
+{
+  NV_IF_TARGET(NV_PROVIDES_SM_90,
+               (
+                 if constexpr (is_thread_block_group_v<_Group>) {
+                   // cooperative groups maps a multidimensional thread id into the thread rank the same way as warps do
+                   const unsigned int __tid             = __g.thread_rank();
+                   const unsigned int __warp_id         = __tid / 32;
+                   const unsigned int __uniform_warp_id = __shfl_sync(~0, __warp_id, 0); // broadcast from lane 0
```
fbusato:
```suggestion
                   const unsigned int __uniform_warp_id = ::__shfl_sync(~0, __warp_id, 0); // broadcast from lane 0
```

libcudacxx/include/cuda/__memcpy_async/elect_one.h
```diff
@@ -45,6 +48,24 @@ _CCCL_BEGIN_NAMESPACE_CUDA_DEVICE
...
 }
 
+template <typename _Group>
+[[nodiscard]] _CCCL_DEVICE _CCCL_FORCEINLINE bool __group_elect_one(const _Group& __g) noexcept
+{
+  NV_IF_TARGET(NV_PROVIDES_SM_90,
+               (
+                 if constexpr (is_thread_block_group_v<_Group>) {
+                   // cooperative groups maps a multidimensional thread id into the thread rank the same way as warps do
+                   const unsigned int __tid             = __g.thread_rank();
```
fbusato:
```suggestion
                   const unsigned __tid             = __g.thread_rank();
```

libcudacxx/include/cuda/__memcpy_async/elect_one.h
```diff
@@ -45,6 +48,24 @@ _CCCL_BEGIN_NAMESPACE_CUDA_DEVICE
     (return threadIdx.x == 0;));
 }
 
+template <typename _Group>
+[[nodiscard]] _CCCL_DEVICE _CCCL_FORCEINLINE bool __group_elect_one(const _Group& __g) noexcept
```
fbusato:
```suggestion
[[nodiscard]] _CCCL_DEVICE_API _CCCL_FORCEINLINE bool __group_elect_one(const _Group& __g) noexcept
```

libcudacxx/include/cuda/__memcpy_async/group_traits.h
```diff
@@ -0,0 +1,61 @@
...
+} // namespace cooperative_groups
+
+_CCCL_BEGIN_NAMESPACE_CUDA
+
+//! Trait to detect whether a group represents a CUDA thread block, for example: ``cooperative_groups::thread_block``.
+template <typename _Group>
+inline constexpr bool is_thread_block_group_v = false;
+
+template <>
+inline constexpr bool is_thread_block_group_v<cooperative_groups::thread_block> = true;
```
fbusato:
```suggestion
inline constexpr bool is_thread_block_group_v<::cooperative_groups::thread_block> = true;
```

docs/libcudacxx/extended_api/asynchronous_operations/memcpy_async.rst
bernhardmgruber: We already assert that pointers are aligned. I added now that the pipeline is not quit.

I cannot easily check whether the parameters are the same across all threads of a group and whether all threads of that group also called the API. It may be possible with some block-wide operations, but seems a bit much for an assertion.

bernhardmgruber:
I am afraid some tests are timeouting with NVRTC on Turing :S

fbusato:
> I am afraid some tests are timeouting with NVRTC on Turing :S

Turing + NVRTC looks more an edge case

bernhardmgruber:
I ran the tests on an RTX 2060 (sm75) and the libcu++ tests succeed normally when not run with nvrtc. When run with nvrtc, the tests hang consuming 100% CPU per test executable.

libcudacxx/test/libcudacxx/cuda/memcpy_async/group_memcpy_async_16b.pass.cpp
```diff
@@ -14,9 +14,9 @@
 
 int main(int argc, char** argv)
 {
-  NV_IF_TARGET(NV_IS_HOST, cuda_thread_count = 4;)
+  NV_IF_TARGET(NV_IS_HOST, cuda_thread_count = thread_block_size;)
```
bernhardmgruber: I could finally reproduce and hunt down this bug, and the problematic line is here. `nvrtcc` (a driver executable for nvrtc) searches the input source for a line like `cuda_thread_count = ...` where `...` is supposed to be an interger literal. Because I put a named constant here, nvrtc ran the tests with a block size of 1, which lead to the hang in the kernel.

bernhardmgruber: Here is a PR to save us such a long hunt next time: #7035

PR #4674, opened 2025-05-11
Introduce `<cuda/vector_types>`

davebayer:
This PR introduces CUDA vector types to libcu++ in `<cuda/vector_types>` header. Read the docs for more info :)

PR #6619, opened 2025-11-13
Implement internal vector type traits

davebayer:
The PR introduces interanl `cuda::__vector_type_t` and `cuda::__has_vector_type_v` traits for crating CUDA vector types from type and size.

I will need this type for hierarchy query result type.

### DIFF
--- a/libcudacxx/include/cuda/__type_traits/vector_type.h
+++ b/libcudacxx/include/cuda/__type_traits/vector_type.h
@@ -193,7 +193,7 @@ template <class _Tp, ::cuda::std::size_t _Size>
     else if constexpr (_Size == 4)
     {
 #  if _CCCL_CTK_AT_LEAST(13, 0)
-      return ::long4_16a{};
+      return ::long4_32a{};
 #  else // ^^^ _CCCL_CTK_AT_LEAST(13, 0) ^^^ / vvv _CCCL_CTK_BELOW(13, 0) vvv
       return ::long4{};
 #  endif // ^^^ _CCCL_CTK_BELOW(13, 0) ^^^
@@ -220,7 +220,7 @@ template <class _Tp, ::cuda::std::size_t _Size>
     else if constexpr (_Size == 4)
     {
 #  if _CCCL_CTK_AT_LEAST(13, 0)
-      return ::ulong4_16a{};
+      return ::ulong4_32a{};
 #  else // ^^^ _CCCL_CTK_AT_LEAST(13, 0) ^^^ / vvv _CCCL_CTK_BELOW(13, 0) vvv
       return ::ulong4{};
 #  endif // ^^^ _CCCL_CTK_BELOW(13, 0) ^^^
@@ -247,7 +247,7 @@ template <class _Tp, ::cuda::std::size_t _Size>
     else if constexpr (_Size == 4)
     {
 #  if _CCCL_CTK_AT_LEAST(13, 0)
-      return ::longlong4_16a{};
+      return ::longlong4_32a{};
 #  else // ^^^ _CCCL_CTK_AT_LEAST(13, 0) ^^^ / vvv _CCCL_CTK_BELOW(13, 0) vvv
       return ::longlong4{};
 #  endif // ^^^ _CCCL_CTK_BELOW(13, 0) ^^^
@@ -274,7 +274,7 @@ template <class _Tp, ::cuda::std::size_t _Size>
     else if constexpr (_Size == 4)
     {
 #  if _CCCL_CTK_AT_LEAST(13, 0)
-      return ::ulonglong4_16a{};
+      return ::ulonglong4_32a{};
 #  else // ^^^ _CCCL_CTK_AT_LEAST(13, 0) ^^^ / vvv _CCCL_CTK_BELOW(13, 0) vvv
       return ::ulonglong4{};
 #  endif // ^^^ _CCCL_CTK_BELOW(13, 0) ^^^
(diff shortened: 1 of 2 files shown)


libcudacxx/include/cuda/__type_traits/vector_type.h
```diff
@@ -0,0 +1,355 @@
...
+  {
+    return;
+  }
+}
+
+template <class _Tp, size_t _Size>
+using __vector_type_t = decltype(::cuda::__cccl_vector_type_t_impl<_Tp, _Size>());
+
+template <class _Tp, size_t _Size>
+inline constexpr bool __has_vector_type_v = !::cuda::std::is_same_v<__vector_type_t<_Tp, _Size>, void>;
```
miscco, line 347: Isnt this the same as __cccL_integer && `is_floating_point`?

davebayer: No, because it checks the size as well

davebayer: And excludes long double

libcudacxx/include/cuda/__type_traits/vector_type.h
```diff
@@ -0,0 +1,355 @@
...
+    {
+      return ::double3{};
+    }
+    else if constexpr (_Size == 4)
+    {
+#  if _CCCL_CTK_AT_LEAST(13, 0)
+      return ::double4_16a{};
+#  else // ^^^ _CCCL_CTK_AT_LEAST(13, 0) ^^^ / vvv _CCCL_CTK_BELOW(13, 0) vvv
+      return ::double4{};
+#  endif // ^^^ _CCCL_CTK_BELOW(13, 0) ^^^
```
bernhardmgruber, line 330: I am not sure this is the right choice. What is the rational to pick the 16 byte aligned type and not the 32? We usually want the 32 byte aligned one whenever we use vector types for load vectorization.

davebayer: I don't really care about those `type4` specializations, I need just those of size 1-3. The reason I chose the `_a16` variant is because it's equivalent to the original `type4` variant.

If you would prefer the `_a32` variant, I we can go for it :)

davebayer:
> I think most places where CubVector<...>::Type is used, we can use this new facility now.

Will it improve anything?

> We may also want to revive some more parts of #4674.

I am not quite sure. Adding the `_a16` and `_a32` variants adds much more complexity to the design. I would go for the exclusive internal exposure only

libcudacxx/include/cuda/__type_traits/vector_type.h
```diff
@@ -0,0 +1,355 @@
...
+      return ::longlong2{};
+    }
+    else if constexpr (_Size == 3)
+    {
+      return ::longlong3{};
+    }
+    else if constexpr (_Size == 4)
+    {
+#  if _CCCL_CTK_AT_LEAST(13, 0)
+      return ::longlong4_16a{};
```
fbusato: this should be `::longlong4_32a{};`

libcudacxx/include/cuda/__type_traits/vector_type.h
```diff
@@ -0,0 +1,355 @@
...
+      return ::ulonglong2{};
+    }
+    else if constexpr (_Size == 3)
+    {
+      return ::ulonglong3{};
+    }
+    else if constexpr (_Size == 4)
+    {
+#  if _CCCL_CTK_AT_LEAST(13, 0)
+      return ::ulonglong4_16a{};
```
fbusato: same here

libcudacxx/include/cuda/__type_traits/vector_type.h
```diff
@@ -0,0 +1,355 @@
...
+      return;
+    }
+  }
+  else
+  {
+    return;
+  }
+}
+
+template <class _Tp, size_t _Size>
```
fbusato: for consistency
```suggestion
template <class _Tp, ::cuda::std::size_t _Size>
```

libcudacxx/include/cuda/__type_traits/vector_type.h
```diff
@@ -0,0 +1,355 @@
...
+#  else // ^^^ _CCCL_CTK_AT_LEAST(13, 0) ^^^ / vvv _CCCL_CTK_BELOW(13, 0) vvv
+      return ::ulonglong4{};
+#  endif // ^^^ _CCCL_CTK_BELOW(13, 0) ^^^
+    }
+    else
+    {
+      return;
+    }
+  }
+  else if constexpr (::cuda::std::is_same_v<_Tp, float>)
```
fbusato, line 287: should we take the opportunity to also handle half/bfloat?

libcudacxx/include/cuda/__type_traits/vector_type.h
bernhardmgruber: Let's do the `_a32` variants! Because then we can use this new factility in some more places in CUB.

bernhardmgruber:
> > I think most places where CubVector<...>::Type is used, we can use this new facility now.
> 
> Will it improve anything?

Yes, because then I can deprecate and eventually remove `CubVector<...>`, or we replace it by something better.

libcudacxx/include/cuda/__type_traits/vector_type.h
davebayer: I don't care about these types right now, if you need them, I can add them. I added the fp types just to cover all types coming from `<vector_types.h>`

davebayer:
> Yes, because then I can deprecate and eventually remove CubVector<...>, or we replace it by something better.

But the problem is that you cannot just implement the arithmetic operators for these types as `CubVector` does

fbusato (approved):

PR #6622, opened 2025-11-13
Extract environment boilerplate code from within the device interfaces to a separate header

### LINKED ISSUE #5606: Refactor and cleanup env-based CUB APIs and underlying implementation
The new cub env based APIs reuse a lot of template code that can be extracted to a common header so that we avoid code duplication as discussed [here](https://github.com/NVIDIA/cccl/pull/5143#discussion_r2255387506). Put the common lines of code under a single header and use it as common functionality under the APIs implementation.

Also try consolidating the two underlying env `cub::DeviceReduce` implementations by template away the Dispatch type.

gonidelis:
fixes #5606

Boilerplate code for extracting types information (`stream`, `mr`, `tuning_t` etc.) is too big and repetitive across the new device environment based interfaces we introduced. This PR extracts the code into a separate function and re-uses it in the existing environment based device APIs that we have (`DeviceScan` and `DeviceReduce`).

Some consideration about the design for the reviewers:

- Each device primitive has its own quirks regarding which `deterministm_t` is supported. For example `DeviceReduce::Reduce` can support both `gpu_to_gpu` and `run_to_run` determinism, while `DeviceReduce::ArgMax/Min` or `DeviceScan` only support `run_to_run` at the moment. That means the determinism heuristics cannot be incorporated into the boilerplate code. Future environment-based APIs must individually evaluate each algorithm to determine and support the appropriate deterministic types.
- The existing boilerplate code uses a lambda callable to pass the specific deterministic algorithm implementation by packing the arguments.

```cpp
      auto reduce_callable = [&](auto tuning, void* storage, size_t& bytes, auto... args) {
        using tuning_t = decltype(tuning);
        return reduce_impl<tuning_t>(storage, bytes, args...);
      };

      // Dispatch with environment - handles all boilerplate
      return detail::dispatch_with_env(
        env, determinism_t{}, reduce_callable, d_in, d_out, num_items, reduction_op, ::cuda::std::identity{}, init);
    }
```

I need some feedback on whether this interface on the `dispatch_with_env()` looks sane.

### DIFF
--- a/cub/cub/detail/env_dispatch.cuh
+++ b/cub/cub/detail/env_dispatch.cuh
@@ -46,14 +46,21 @@ CUB_RUNTIME_FUNCTION static cudaError_t dispatch_with_env(EnvT env, AlgorithmCal
     ::cuda::std::execution::__query_or(env, ::cuda::mr::__get_memory_resource, detail::device_memory_resource{});
 
   // Query tuning from environment
-  using tuning_t =
-    ::cuda::std::execution::__query_result_or_t<EnvT, ::cuda::execution::__get_tuning_t, ::cuda::std::execution::env<>>;
+
+  // clang 15-19 ICE when using __query_result_or_t, so we just use the underlying machinery as a workaround
+#if _CCCL_COMPILER(CLANG, >=, 15) && _CCCL_COMPILER(CLANG, <, 19)
+  const auto tuning = ::cuda::std::execution::__detail::__query_or_t{}(
+    env, ::cuda::execution::__get_tuning_t{}, ::cuda::std::execution::env<>{});
+#else
+  const auto tuning =
+    ::cuda::std::execution::__query_result_or_t<EnvT, ::cuda::execution::__get_tuning_t, ::cuda::std::execution::env<>>{};
+#endif
 
   void* d_temp_storage      = nullptr;
   size_t temp_storage_bytes = 0;
 
   // Phase 1: Query temporary storage size
-  if (const auto error = algorithm_callable(tuning_t{}, d_temp_storage, temp_storage_bytes, stream.get()))
+  if (const auto error = algorithm_callable(tuning, d_temp_storage, temp_storage_bytes, stream.get()))
   {
     return error;
   }
@@ -65,7 +72,7 @@ CUB_RUNTIME_FUNCTION static cudaError_t dispatch_with_env(EnvT env, AlgorithmCal
   }
 
   // Phase 2: Execute algorithm
-  const auto error = algorithm_callable(tuning_t{}, d_temp_storage, temp_storage_bytes, stream.get());
+  const auto error = algorithm_callable(tuning, d_temp_storage, temp_storage_bytes, stream.get());
 
   // Deallocate temporary storage (always attempt, even on error)
   const auto deallocate_error =
--- a/cub/cub/device/device_reduce.cuh
+++ b/cub/cub/device/device_reduce.cuh
@@ -907,7 +907,6 @@ public:
 
     using InitT    = OutputT;
     using limits_t = ::cuda::std::numeric_limits<InitT>;
-
     // Dispatch with environment - handles all boilerplate
     return detail::dispatch_with_env(env, [&]([[maybe_unused]] auto tuning, void* storage, size_t& bytes, auto stream) {
       using tuning_t = decltype(tuning);


cub/cub/device/device_reduce.cuh
```diff
@@ -2104,7 +1956,7 @@ public:
     using InitT           = OutputExtremumT;
 
     // Initial value
-    OutputExtremumT initial_value{::cuda::std::numeric_limits<InputValueT>::max()};
+    OutputExtremumT initial_value{::cuda::std::numeric_limits<InputValueT>::lowest()};
```
gonidelis, line 1917: I need to figure out why this bug was not caught from the tests

cub/cub/device/device_reduce.cuh
```diff
@@ -506,65 +507,15 @@ public:
...
-
-      if (error != cudaSuccess)
-      {
-        // Reduction error takes precedence over deallocation error since it happens first
-        return error;
-      }
-
-      return deallocate_error;
+      // Lambda that calls reduce_impl with the right overload based on determinism
+      auto reduce_callable = [&](auto tuning, void* storage, size_t& bytes, auto... args) {
```
NaderAlAwar: Suggestion: it seems that we are not using anything from the outside scope, so we don't need to capture by reference `[&]`. Comment applies throughout PR

NaderAlAwar: Question: do we need a lambda to make this work? Can we just pass `reduce_impl<tuning_t>` directly?

srinivasyadav18: Might not be possible as `reduce_impl` is a function template, hence we are wrapping in a callable lambda.

cub/cub/detail/env_dispatch.cuh
```diff
@@ -0,0 +1,98 @@
+// SPDX-FileCopyrightText: Copyright (c) 2025, NVIDIA CORPORATION. All rights reserved.
+// SPDX-License-Identifier: BSD-3-Clause
```
bernhardmgruber, line 2: Remark: I assume we are retaining BSD-3 here, cause that's what we had in the file where we extracted the below code from.

cub/cub/detail/env_dispatch.cuh
```diff
@@ -0,0 +1,98 @@
...
+  {
+    return error;
+  }
+
+  // Allocate temporary storage
+  error = CubDebug(detail::temporary_storage::allocate(stream, d_temp_storage, temp_storage_bytes, mr));
+  if (error != cudaSuccess)
+  {
+    return error;
+  }
```
bernhardmgruber: Suggestion: I got really fond of this style of error handling recently:
```suggestion
  if (const auto error = CubDebug(detail::temporary_storage::allocate(stream, d_temp_storage, temp_storage_bytes, mr)))
  {
    return error;
  }
```
This way, we don't leak an `error` variable anywhere else, so we are always reporting the error we just got.

cub/cub/device/device_scan.cuh
```diff
@@ -404,12 +329,43 @@ struct DeviceScan
...
+
+    using InitT = cub::detail::it_value_t<InputIteratorT>;
     InitT init_value{};
 
-    return scan_impl_env(d_in, d_out, ::cuda::std::plus<>{}, detail::InputValue<InitT>(init_value), num_items, env);
+    // Lambda that calls scan_impl_determinism
+    auto scan_callable = [&](auto tuning, void* storage, size_t& bytes, auto... args) {
+      using tuning_t = decltype(tuning);
+      return scan_impl_determinism<tuning_t>(storage, bytes, args...);
+    };
```
bernhardmgruber: Strong suggestion: That's a lot of duplication with the overload below. I think we should retain `scan_impl_env` to share the implementation.

gonidelis: done

gonidelis:
LLVMs 15-18 are not friendly with complex C++20 aggregate expressions codegen hence this PR breaks CCCL for these compilers/C++ version.

Investigating what's the best course of action. I see three options:

a) simplify the code in this pr
b) file a bug to llvm
c) reject pr

digging deeper to see what's the exact issue.

cub/cub/detail/env_dispatch.cuh
```diff
@@ -0,0 +1,98 @@
...
+    d_temp_storage,
+    temp_storage_bytes,
+    ::cuda::std::forward<AlgorithmArgs>(algorithm_args)...,
+    determinism,
+    stream.get());
+
+  if (error != cudaSuccess)
+  {
+    return error;
+  }
```
bernhardmgruber: Suggestion: I got really fond of this pattern:
```suggestion
  if (const auto error = algorithm_callable(
    tuning_t{},
    d_temp_storage,
    temp_storage_bytes,
    ::cuda::std::forward<AlgorithmArgs>(algorithm_args)...,
    determinism,
    stream.get()))
  {
    return error;
  }
```
I would apply it everywhere.

cub/cub/detail/env_dispatch.cuh
```diff
@@ -0,0 +1,98 @@
...
+
+  void* d_temp_storage      = nullptr;
+  size_t temp_storage_bytes = 0;
+
+  // Phase 1: Query temporary storage size
+  cudaError_t error = algorithm_callable(
+    tuning_t{},
+    d_temp_storage,
+    temp_storage_bytes,
+    ::cuda::std::forward<AlgorithmArgs>(algorithm_args)...,
```
bernhardmgruber: Important: We must not forward `algorithm_args` here. If it would move any argument, then the argument is gone for the next call to `algorithm_callable` below. This must be:
```suggestion
    algorithm_args...,
```

cub/cub/device/device_reduce.cuh
```diff
@@ -496,65 +497,15 @@ public:
...
-      return deallocate_error;
+      // Lambda that calls reduce_impl with the right overload based on determinism
+      auto reduce_callable = [](auto tuning, void* storage, size_t& bytes, auto... args) {
+        using tuning_t = decltype(tuning);
+        return reduce_impl<tuning_t>(storage, bytes, args...);
+      };
+
+      // Dispatch with environment - handles all boilerplate
+      return detail::dispatch_with_env(
+        env, determinism_t{}, reduce_callable, d_in, d_out, num_items, reduction_op, ::cuda::std::identity{}, init);
```
bernhardmgruber: Suggestion: I would remove the determinism arg from `dispatch_with_env` and, if possible, the variadic arguments. The function should just handle tuning, allocation, and the stream. Then just capture all args by reference:
```suggestion
      // Lambda that calls reduce_impl with the right overload based on determinism
      auto reduce_callable = [](auto tuning, void* storage, size_t& bytes, cudaStream_t stream) {
        using tuning_t = decltype(tuning);
        return reduce_impl<tuning_t>(storage, bytes, d_in, d_out, num_items, reduction_op, ::cuda::std::identity{}, init, determinism, stream);
      };

      // Dispatch with environment - handles all boilerplate
      return detail::dispatch_with_env(
        env, determinism_t{}, reduce_callable);
```
This way, `dispatch_with_env` should be a lot simpler and maybe more reusable (since it does not need to deal with determinism).

gonidelis: i love that and it was a point of discussion with @srinivasyadav18: how much should we abstract away for the facility to be reusable, yet comply to the peculiarities of every algorithm (especially the deterministic ones)

gonidelis: @bernhardmgruber in your suggested code you are still passing `determinism_t{}` despite your suggestion though:

``` cpp
return detail::dispatch_with_env(
        env, determinism_t{}, reduce_callable);
```

gonidelis: ~Also I expect the variadic template arg removal to cure the Clang15-19 complains~ it didn't:

`tuning_t{}` is still a SFINAE-dependent alias and the call to that lambda is still through a template. Clang 15-18 code generator has a bug in how it handles this specific combination, leading to the ICE in `EmitMaterializeTemporaryExpr`. :(((

bernhardmgruber: > @bernhardmgruber in your suggested code you are still passing `determinism_t{}` despite your suggestion though:
> 
> ```c++
> return detail::dispatch_with_env(
>         env, determinism_t{}, reduce_callable);
> ```

Yeah sorry, just drop it.

bernhardmgruber: Pushed the refactoring myself. Compiles with clang-20. Let's see what the CI says.

cub/cub/detail/env_dispatch.cuh
```diff
@@ -46,14 +46,21 @@ CUB_RUNTIME_FUNCTION static cudaError_t dispatch_with_env(EnvT env, AlgorithmCal
     ::cuda::std::execution::__query_or(env, ::cuda::mr::__get_memory_resource, detail::device_memory_resource{});
 
   // Query tuning from environment
-  using tuning_t =
-    ::cuda::std::execution::__query_result_or_t<EnvT, ::cuda::execution::__get_tuning_t, ::cuda::std::execution::env<>>;
+
+  // clang 15-19 ICE when using __query_result_or_t, so we just use the underlying machinery as a workaround
```
gonidelis, line 50: i would promise we tried the same thing with @miscco

gonidelis:
Once green this should be good to go

cub/cub/detail/env_dispatch.cuh
```diff
@@ -0,0 +1,87 @@
...
+//! Generic environment-based algorithm dispatch wrapper
+//!
+//! Handles common boilerplate for all env-based algorithms:
+//! - Query stream, memory resource, and tuning from environment
+//! - Two-phase call (query temp storage size, then execute)
+//! - Temporary storage allocation/deallocation
+//! - Memory resource querying from environment
+//!
+//! @param env The execution environment
+//! @param algorithm_callable Callable that invokes the algorithm implementation with determinism specified
```
bernhardmgruber, line 37:
```suggestion
//! @param algorithm_callable Callable that invokes the algorithm
```

cub/cub/device/device_reduce.cuh
```diff
@@ -1294,6 +1152,9 @@ public:
     void* d_temp_storage      = nullptr;
     size_t temp_storage_bytes = 0;
 
+    using tuning_t =
+      ::cuda::std::execution::__query_result_or_t<EnvT, ::cuda::execution::__get_tuning_t, ::cuda::std::execution::env<>>;
+
```
bernhardmgruber, line 1157: Why do we need this here?

cub/cub/device/device_reduce.cuh
```diff
@@ -2084,6 +1901,9 @@ public:
...
```
bernhardmgruber, line 1906: Also, I think this is unused.

bernhardmgruber (approved):

cub/cub/device/device_reduce.cuh
gonidelis: will patch it up

PR #6623, opened 2025-11-13
streamline the implementation of `cuda::std::__tuple`

ericniebler:
## Description

`cuda::std::__tuple` is intended to be a lighter-weight alternative to `cuda::std::tuple`. it is also a structural type, so unlike `cuda::std::tuple`, a function could return a `__tuple` of immovable types.

the implementation of `__tuple` got more complicated than necessary. i since have found a simpler, shorter, and more efficient way to implement it.

## Checklist

### DIFF
--- a/libcudacxx/include/cuda/std/__utility/pod_tuple.h
+++ b/libcudacxx/include/cuda/std/__utility/pod_tuple.h
@@ -24,6 +24,7 @@
 #include <cuda/std/__concepts/concept_macros.h>
 #include <cuda/std/__type_traits/decay.h>
 #include <cuda/std/__type_traits/is_callable.h>
+#include <cuda/std/__type_traits/remove_reference.h>
 #include <cuda/std/__utility/declval.h>
 #include <cuda/std/__utility/integer_sequence.h>
 #include <cuda/std/__utility/undefined.h>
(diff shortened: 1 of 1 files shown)


libcudacxx/include/cuda/std/__utility/pod_tuple.h
```diff
@@ -103,11 +67,35 @@ _CCCL_BEGIN_NAMESPACE_CUDA_STD
...
+template <class... _Ts>
+struct _CCCL_TYPE_VISIBILITY_DEFAULT _CCCL_DECLSPEC_EMPTY_BASES __tuple;
+
+namespace __detail
+{
+template <class _Fn, class _Tuple, class... _Us>
+extern __undefined<_Fn, _Tuple, _Us...> __applicable_v;
+
+template <class _Fn, class... _Ts, class... _Us>
+constexpr bool __applicable_v<_Fn, __tuple<_Ts...>, _Us...> = __is_callable_v<_Fn, _Us..., _Ts...>;
```
davebayer:
```suggestion
inline constexpr bool __applicable_v<_Fn, __tuple<_Ts...>, _Us...> = __is_callable_v<_Fn, _Us..., _Ts...>;
```
applies to all other template variables as well

libcudacxx/include/cuda/std/__utility/pod_tuple.h
davebayer, line 106: Q: wouldn't be better to specialize this class for cases when `_Ty` is empty rather than using `_CCCL_NO_UNIQUE_ADDRESS`? I remember some data corruption bugs when using `[[no_unique_address]]`

libcudacxx/include/cuda/std/__utility/pod_tuple.h
```diff
@@ -121,284 +109,120 @@ struct __tupl_base;
...
+    return static_cast<_Fn&&>(__fn)(
+      static_cast<_Us&&>(__us)..., static_cast<_Self&&>(__self)._CCCL_CWG1835_TEMPLATE __box<_Index, _Ts>::__value...);
+  }
 };
+} // namespace __detail
 
 template <class... _Ts>
-struct _CCCL_TYPE_VISIBILITY_DEFAULT _CCCL_DECLSPEC_EMPTY_BASES __tuple //
-    : __tupl_base<index_sequence_for<_Ts...>, _Ts...>
+struct __tuple : __detail::__tupl_base<index_sequence_for<_Ts...>, _Ts...>
```
davebayer, line 130: Shouldn't we have `_CCCL_DECLSPEC_EMPTY_BASES` here, too?

ericniebler: the attribute is specified when the struct is forward-declared at the top of the file.

libcudacxx/include/cuda/std/__utility/pod_tuple.h
ericniebler: my preference is to control this via the definition of `_CCCL_NO_UNIQUE_ADDRESS` itself. on buggy compilers, it should expand to nothing.

by doing it this way, i can uniformly access the I-th value as `tup.__box<I, Ti>::value`.

libcudacxx/include/cuda/std/__utility/pod_tuple.h
ericniebler: why? their address is never taken.

davebayer: Is the compiler in that case forced to not include the variable in the object file?

ericniebler: my assumption has always been that `inline` variables are more work for the compiler/linker, but i have never measured.

davebayer: That could be true, but the symbol could be defined in multiple TUs and cause link error, if I am correct

miscco: the NVCC compiler *always* ODR uses variables

miscco: I agree with @davebayer we should have an explicit reason to diverge from inline variable and a clear benefit.

The goal is that any reader can immediately understand the code and not wondering why in this situation it is not an inline variable and the ramifications of that.

ericniebler: > the symbol could be defined in multiple TUs and cause link error

`const` objects at namespace scope have internal linkage. but it doesn't really matter. i can make them `inline`.

davebayer: Just out of curiosity - why are all traits in the standard library `inline constexpr` then?

ericniebler: since they are public, it is possible someone might care about the address of one of them.

PR #6626, opened 2025-11-13
Add `__version_(at_least|below)` utilities to CUDA Driver wrappers

davebayer:
We already have these utilities for header's version checks, this PR brings them to `cuda::__driver::` namespace, so we can use them to check the version of the CUDA driver in runtime and prevent misspelled versions.

### DIFF
--- a/libcudacxx/include/cuda/__driver/driver_api.h
+++ b/libcudacxx/include/cuda/__driver/driver_api.h
@@ -61,6 +61,14 @@ _CCCL_SUPPRESS_DEPRECATED_PUSH
 
 _CCCL_SUPPRESS_DEPRECATED_POP
 
+//! @brief Makes the driver version from major and minor version.
+[[nodiscard]] _CCCL_HOST_API constexpr int __make_version(int __major, int __minor) noexcept
+{
+  _CCCL_ASSERT(__major >= 12, "invalid major CUDA Driver version");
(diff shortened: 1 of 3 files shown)


libcudacxx/include/cuda/__driver/driver_api.h
```diff
@@ -171,6 +171,16 @@ __get_driver_entry_point(const char* __name, [[maybe_unused]] int __major = 12,
   return __version;
 }
 
+[[nodiscard]] _CCCL_HOST_API inline bool __version_at_least(int __major, int __minor)
+{
+  return ::cuda::__driver::__getVersion() >= __major * 1000 + __minor * 10;
```
miscco: I am missing verification here
```suggestion
  _CCCL_ASSERT(__major >= 12 && __major < 100, "Invalid major CUDA version");
  _CCCL_ASSERT(__minor >= 0 && __minor < 100, "Invalid minor  CUDA version");
  return ::cuda::__driver::__getVersion() >= __major * 1000 + __minor * 10;
```

miscco (approved):

libcudacxx/include/cuda/__driver/driver_api.h
davebayer: Moved it to a separate function

PR #6627, opened 2025-11-14
[cudax->libcu++] Move buffer type from cudax to libcu++

pciolkosz:
PR moving last bits of the buffer type and de-duplicating the `__memory_accessibility` enum

### DIFF
--- a/libcudacxx/test/libcudacxx/cuda/containers/buffer/properties.cu
+++ b/libcudacxx/test/libcudacxx/cuda/containers/buffer/properties.cu
@@ -50,7 +50,7 @@ C2H_CCCLRT_TEST("cuda::buffer properties", "[container][buffer]", test_types)
   static_assert(cuda::std::is_same_v<cuda::std::reverse_iterator<iterator>, typename Buffer::reverse_iterator>, "");
   static_assert(
     cuda::std::is_same_v<cuda::std::reverse_iterator<const_iterator>, typename Buffer::const_reverse_iterator>, "");
-  // TODO fix Clang not recongnizing device accessible buffers as contiguous range
+  // TODO fix Clang not recognizing device accessible buffers as contiguous range
 #if !_CCCL_CUDA_COMPILER(CLANG)
   static_assert(cuda::std::ranges::contiguous_range<Buffer>);
 #endif // ^^^ !_CCCL_CUDA_COMPILER(CLANG) ^^^


cudax/test/CMakeLists.txt
```diff
@@ -119,15 +119,6 @@ foreach (cudax_target IN LISTS cudax_TARGETS)
 
   cudax_add_catch2_test(test_target containers ${cudax_target}
     containers/uninitialized_buffer.cu
```
miscco, line 92: arent we moving `uninitialized_buffer` too

libcudacxx/include/cuda/__container/buffer.h
```diff
@@ -1,15 +1,15 @@
...
 // under the Apache License v2.0 with LLVM Exceptions.
 // See https://llvm.org/LICENSE.txt for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 // SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES.
 //
 //===----------------------------------------------------------------------===//
 
-#ifndef __CUDAX__CONTAINER_ASYNC_BUFFER__
-#define __CUDAX__CONTAINER_ASYNC_BUFFER__
+#ifndef _CUDA___CONTAINER_ASYNC_BUFFER_H
```
miscco: Nitpick: this is only 
```suggestion
#ifndef _CUDA___CONTAINER_BUFFER_H
```

libcudacxx/include/cuda/__container/uninitialized_async_buffer.h
```diff
@@ -71,7 +75,8 @@ class __uninitialized_async_buffer
 {
 private:
   static_assert(::cuda::mr::__contains_execution_space_property<_Properties...>,
-                "The properties of cuda::__uninitialized_async_buffer must contain at least one "
+                "The properties of cuda::__uninitialized_async_buffer must "
+                "contain at least one "
                 "execution space property!");
```
miscco: Please wrap the strings

libcudacxx/include/cuda/__memory_resource/synchronous_resource_adapter.h
```diff
@@ -28,8 +28,8 @@
 
 #include <cuda/std/__cccl/prologue.h>
 
-namespace cuda::experimental
-{
+_CCCL_BEGIN_NAMESPACE_CUDA
```
miscco: Should this be 
```suggestion
_CCCL_BEGIN_NAMESPACE_CUDA_MR
```

libcudacxx/test/libcudacxx/cuda/containers/buffer/conversion.cu
```diff
@@ -1,13 +1,14 @@
 //===----------------------------------------------------------------------===//
 //
-// Part of CUDA Experimental in CUDA C++ Core Libraries,
+// Part of libcu++, the C++ Standard Library for your entire system,
```
miscco, line 3: Nitpick: all those should be in the folder `buffer` not `async_buffer`

libcudacxx/include/cuda/__container/buffer.h
```diff
@@ -692,7 +666,7 @@ __fill_n(cuda::stream_ref __stream, _Tp* __first, ::cuda::std::size_t __count, c
     else
     {
 #if _CCCL_HAS_CUDA_COMPILER()
-      ::cuda::experimental::__ensure_current_device __guard(__stream);
+      ::cuda::__ensure_current_context __guard(__stream);
       ::cub::DeviceTransform::Fill(__first, __count, __value, __stream.get());
```
davebayer, line 672: Can't we implement this as (in pseudocode):
```cpp
void* __tmp;
cuda_malloc_host_async(&__tmp, sizeof(__value), __cccl_alloc_stream());
__cccl_alloc_stream().sync();
memcpy(__tmp, &__value, sizeof(__value));
::cuMemcpy2DAsync(__tmp __first, __count, src_pitch=0, __first, dst_pitch=sizeof(__value));
cuda_free_host_async(__tmp);
```
We would:
1. get rid of the cub include
2. enable filling without CUDA compiler

pciolkosz: We could, but the performance would be considerably worse

davebayer: But still better than nothing right. It would allow us to merge the PR and we can improve the performance later, once cub can be included from libcu++

pciolkosz: Ahh, I see. Yeah that would be an option. But I think the blocker should be resolved soon in https://github.com/NVIDIA/cccl/pull/6694, lets wait a bit and see how it goes

davebayer: Yeah, but still, we will need this implementation so we needn't to `static_assert(false)` for the larger value sizes, so I think you can add the implementation anyway

pciolkosz: I think I would rather `static_assert` than implement the slow path, it might be surprising for the users and I believe they should just use something else in that case

davebayer: I will try to benchmark it tomorrow, I think suboptimal solution is better than no solution :)

cudax/test/CMakeLists.txt
pciolkosz: We don't really use it anywhere, I think it can stay experimental until we have a usage for it

bernhardmgruber: What's the new replacement for `thrust::device_vector<T>(n, thrust::no_init)`, which is used everywhere in benchmarks and tests?

pciolkosz: `cuda::make_buffer<T>(stream, cuda::device_default_memory_pool(cuda::devices[i]), n, cuda::no_init)`. We might consider a version of it in the future that does not take a stream

libcudacxx/include/cuda/__container/heterogeneous_iterator.h
```diff
@@ -229,13 +229,16 @@ class heterogeneous_iterator
...
+  // TODO fix sfinae issue here for older MSVC
+#if !_CCCL_COMPILER(MSVC) || _CCCL_COMPILER(MSVC, >=, 14, 30)
   //! @brief Constructs an immutable \c heterogeneous_iterator from a mutable one
   //! @param __other The mutable \c heterogeneous_iterator
   _CCCL_TEMPLATE(class _OtherTp)
   _CCCL_REQUIRES((::cuda::std::is_same_v<_OtherTp, value_type>) _CCCL_AND(::cuda::std::is_const_v<_CvTp>))
   _CCCL_API constexpr heterogeneous_iterator(heterogeneous_iterator<_OtherTp, _Properties...> __other) noexcept
       : __base(__other.__ptr_)
   {}
+#endif // ^^^ !_CCCL_COMPILER(MSVC) ^^^
```
davebayer:
```suggestion
//! @brief Constructs an immutable \c heterogeneous_iterator from a mutable one
//! @param __other The mutable \c heterogeneous_iterator
_CCCL_TEMPLATE(class _OtherTp, class _CvTp2 = _CvTp)
_CCCL_REQUIRES((::cuda::std::is_same_v<_OtherTp, value_type>) _CCCL_AND(::cuda::std::is_const_v<_CvTp2>))
_CCCL_API constexpr heterogeneous_iterator(heterogeneous_iterator<_OtherTp, _Properties...> __other) noexcept
: __base(__other.__ptr_)
{}
```

davebayer:
pre-commit.ci autofix

libcudacxx/test/libcudacxx/cuda/containers/buffer/properties.cu
```diff
@@ -51,5 +50,8 @@ C2H_CCCLRT_TEST("cudax::buffer properties", "[container][buffer]", test_types)
   static_assert(cuda::std::is_same_v<cuda::std::reverse_iterator<iterator>, typename Buffer::reverse_iterator>, "");
   static_assert(
     cuda::std::is_same_v<cuda::std::reverse_iterator<const_iterator>, typename Buffer::const_reverse_iterator>, "");
+  // TODO fix Clang not recongnizing device accessible buffers as contiguous range
```
davebayer:
```suggestion
  // TODO fix Clang not recognizing device accessible buffers as contiguous range
```

davebayer:
pre-commit.ci autofix

libcudacxx/include/cuda/__container/buffer.h
davebayer: Okay, this solution doesn't work, because the pitch must be at least the size of the copied data..

cudax/test/CMakeLists.txt
davebayer: I think it would look much better if we could just do:
```cpp
cuda::make_buffer<T>(stream, cuda::devices[i], n, cuda::no_init);
```
We could add a constructor from `cuda::memory_location` which would mean the default memory pool

libcudacxx/include/cuda/__container/buffer.h
```diff
@@ -1,16 +1,18 @@
...
 //
 //===----------------------------------------------------------------------===//
 
-#ifndef __CUDAX__CONTAINER_ASYNC_BUFFER__
-#define __CUDAX__CONTAINER_ASYNC_BUFFER__
+#ifndef _CUDA___CONTAINER_BUFFER_H
+#define _CUDA___CONTAINER_BUFFER_H
 
+// Temporary workaround to not trigger issues in CUB headers missing prologue
+#define _CCCL_WAIVE_PROLOGUE_INCLUDE_CHECK
```
davebayer, line 15: What was the problem? I don't like that we are disabling this

pciolkosz: CUB didn't properly include prologue headers and I didn't want to try to fix that as part of this PR

davebayer: I understand but then there are these workarounds coming out of nowhere. I think it would be better to fix this problem in another PR before merging this one.

It seems that we will include more cub headers, so we need a systematic solution.

davebayer: Fixed by #6844

pciolkosz: I still think it would be nice to have an opt-out like that available for similar future cases, especially since the prologue thing is not essential, but I don't mind removing it

davebayer: No, it is absolutely essential. All code in libcu++ should be guarded by the prologue/epilogue include. We do many important stuff there to prevent other macros breaking our code

pciolkosz: But now we also include CUB in libcu++ and CUB doesn't use it, so I would be cautious to call it essential

davebayer: And not only that, we suppress compiler warnings and other stuff. That's why the checks for the prologue/epilogue wrapping are so strict (and annoying)

davebayer: CUB doesn't do many of the things we do in libcu++, but that doesn't mean we should relax requirements on what we do in libcu++

miscco: I am with Dave here, this is potentially breaking libcu++ which we must not do

PR #6628, opened 2025-11-14
implement `indeterminate_domain` from P3826R2

ericniebler:
## Description

my focus recently has been shoring up `std::execution`'s algorithm customization mechanism ahead of the C++26 release. in that mechanism, the domains on which an operation starts and completes are used to dispatch to the correct algorithm implementation.

a thorny issue has been: what to do when an operation can complete on domain `A` _or_ domain `B`? that can sometimes happen. without the ability to pick _one_ domain, the customization mechanism gives up, and no custom implementation is used.

during last week's committee meeting i had a happy thought. most domains do not customize most algorithms. it shouldn't be an error if `then(sndr, fn)` can complete on domain `A` or domain `B` _if neither domain customizes the `then` algorithm_.

this led me to the idea of using a set of domains as a domain. `indeterminate_domain<A, B>` captures the fact that a sender completes on one of two domains.

dispatching with `indeterminate_domain<A, B>` proceeds as follows:

- compute the types of `D().transform_sender(set_value, sndr, env)` for `D` in `A`, `B`, and `default_domain`.
- if the 3 transformed sender types are the same, then we use the `default_domain` transform.
- otherwise, it is a hard error.

## Checklist

### DIFF
--- a/cudax/include/cuda/experimental/__execution/bulk.cuh
+++ b/cudax/include/cuda/experimental/__execution/bulk.cuh
@@ -198,18 +198,30 @@ struct _CCCL_TYPE_VISIBILITY_DEFAULT __bulk_t
   struct _CCCL_TYPE_VISIBILITY_DEFAULT __closure_base_t
   {
     template <class _Sndr>
-    [[nodiscard]] _CCCL_API friend constexpr auto operator|(_Sndr&& __sndr, __closure_base_t __self)
+    [[nodiscard]] _CCCL_API constexpr auto operator()(_Sndr&& __sndr) &&
     {
       static_assert(__is_sender<_Sndr>);
 
       if constexpr (!dependent_sender<_Sndr>)
       {
         using __sndr_t = typename _BulkTag::template __sndr_t<_Sndr, _Policy, _Shape, _Fn>;
-        __assert_valid_completion_signatures(get_completion_signatures<__sndr_t>());
+        __assert_valid_completion_signatures(execution::get_completion_signatures<__sndr_t>());
       }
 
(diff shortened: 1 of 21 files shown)


miscco:
Question: If only Domain A and *NOT* Domain B specializes an algorithm, is the result an error?

ericniebler:
Yes

miscco:
> Yes

I am not sure whether that is a correct approach. I dont know enough about domains, but I could imagine that specializing algorithms for a specific domain is a valid use case. 

With the proposed design, how would I as Domain A author specialize algorithms, if I have no access / knowledge about Domain B? In the end It is the user connecting those domains?

ericniebler:
> With the proposed design, how would I as Domain A author specialize algorithms, if I have no access / knowledge about Domain B? In the end It is the user connecting those domains?

a sender has an indeterminate domain when it can complete in different places. consider:

```cpp
auto sndr = when_all(starts_on(cpu, my_task1),
                     starts_on(gpu, my_task2))
          | bulk(par, shape, fn);
```

because of the semantics of the `when_all` algorithm, the function `fn` could end up running on the cpu _or_ the gpu. if there is a customization of `bulk` for the gpu, should we use it? no, because the `bulk` operation may run on the cpu; a gpu implementation would not be appropriate.

so do we pick the cpu's (default) implementation? no, because that might be inappropriate should the `bulk` execute on the gpu.

so it is an error. the user has to transfer to a determinate domain for algorithm dispatch to work again, like this:

```c++
auto sndr = when_all(starts_on(cpu, my_task1),
                     starts_on(gpu, my_task2))
          | contunues_on(gpu)
          | bulk(par, shape, fn);
```

PR #6632, opened 2025-11-14
Refactor driver APIs

davebayer:
Currently, we have 2 types of CUDA Driver calls:
1. `__meow(...)` and 
2. `__meowNoThrow(...)`

I don't like the design, it forces us to repeat the implementations and forces us to have in/out parameters.

This PR refactors the driver APIs and keeps only 1 definition for each API. Each function is `noexcept` and returns `std::expected`-like type that returns the error + return values.

In case that throwing behaviour is demanded, `_CCCL_TRY_DRIVER_API` macro is provided, which checks the return value of the CUDA call and returns the output parameters. Alternatively, `_CCCL_ASSERT_DRIVER_API` can be used to assert that a call succeeded.

### DIFF
--- a/cudax/include/cuda/experimental/__launch/configuration.cuh
+++ b/cudax/include/cuda/experimental/__launch/configuration.cuh
@@ -325,33 +325,34 @@ template <class _Tp>
 [[nodiscard]] ::cudaError_t __apply_launch_option(
   const dynamic_shared_memory<_Tp>& __opt, ::CUlaunchConfig& __config, ::CUfunction __kernel) noexcept
 {
-  ::cudaError_t __status = ::cudaSuccess;
-
-  // Since CUDA 12.4, querying CU_FUNC_ATTRIBUTE_SHARED_SIZE_BYTES requires the
-  // function to be loaded.
-  if (::cuda::__driver::__version_at_least(12, 4))
+  // Since CUDA 12.4, querying CU_FUNC_ATTRIBUTE_SHARED_SIZE_BYTES requires the function to be loaded.
+  const auto __needs_load = ::cuda::__driver::__version_at_least(12, 4);
+  if (__needs_load.__error_ != ::CUDA_SUCCESS)
   {
-    __status = ::cuda::__driver::__functionLoadNoThrow(__kernel);
-    if (__status != ::cudaSuccess)
+    return static_cast<::cudaError_t>(__needs_load.__error_);
+  }
+
+  if (__needs_load.__value_)
+  {
+    const auto __status = ::cuda::__driver::__functionLoad(__kernel).__error_;
+    if (__status != ::CUDA_SUCCESS)
     {
-      return __status;
+      return static_cast<::cudaError_t>(__status);
     }
   }
 
-  int __static_smem_size{};
-  __status = ::cuda::__driver::__functionGetAttributeNoThrow(
-    __static_smem_size, ::CU_FUNC_ATTRIBUTE_SHARED_SIZE_BYTES, __kernel);
-  if (__status != ::cudaSuccess)
+  const auto __static_smem_size =
+    ::cuda::__driver::__functionGetAttribute(::CU_FUNC_ATTRIBUTE_SHARED_SIZE_BYTES, __kernel);
+  if (__static_smem_size.__error_ != ::CUDA_SUCCESS)
   {
-    return __status;
+    return static_cast<::cudaError_t>(__static_smem_size.__error_);
   }
 
-  int __max_dyn_smem_size{};
-  __status = ::cuda::__driver::__functionGetAttributeNoThrow(
-    __max_dyn_smem_size, ::CU_FUNC_ATTRIBUTE_SHARED_SIZE_BYTES, __kernel);
-  if (__status != ::cudaSuccess)
+  const auto __max_dyn_smem_size =
+    ::cuda::__driver::__functionGetAttribute(::CU_FUNC_ATTRIBUTE_SHARED_SIZE_BYTES, __kernel);
+  if (__max_dyn_smem_size.__error_ != ::CUDA_SUCCESS)
   {
-    return __status;
+    return static_cast<::cudaError_t>(__max_dyn_smem_size.__error_);
   }
 
   const auto __dyn_smem_size = ::cuda::overflow_cast<int>(__opt.size_bytes());
@@ -360,19 +361,21 @@ template <class _Tp>
     return ::cudaErrorInvalidValue;
   }
 
-  const int __smem_size = __static_smem_size + __dyn_smem_size.value;
+  const int __smem_size = __static_smem_size.__value_ + __dyn_smem_size.value;
   if (static_cast<::cuda::std::size_t>(__smem_size) > __max_portable_dyn_smem_size && !__opt.__non_portable_)
   {
     return ::cudaErrorInvalidValue;
   }
 
-  if (__max_dyn_smem_size < __dyn_smem_size.value)
+  if (__max_dyn_smem_size.__value_ < __dyn_smem_size.value)
   {
-    __status = ::cuda::__driver::__functionSetAttributeNoThrow(
-      __kernel, ::CU_FUNC_ATTRIBUTE_MAX_DYNAMIC_SHARED_SIZE_BYTES, __dyn_smem_size.value);
-    if (__status != ::cudaSuccess)
+    const auto __status =
+      ::cuda::__driver::__functionSetAttribute(
+        __kernel, ::CU_FUNC_ATTRIBUTE_MAX_DYNAMIC_SHARED_SIZE_BYTES, __dyn_smem_size.value)
+        .__error_;
+    if (__status != ::CUDA_SUCCESS)
     {
-      return __status;
+      return static_cast<::cudaError_t>(__status);
     }
   }
 
--- a/libcudacxx/include/cuda/__driver/driver_api.h
+++ b/libcudacxx/include/cuda/__driver/driver_api.h
@@ -27,12 +27,54 @@
 #  include <cuda/std/__internal/namespaces.h>
 #  include <cuda/std/__type_traits/always_false.h>
 #  include <cuda/std/__type_traits/is_same.h>
+#  include <cuda/std/__utility/pod_tuple.h>
+#  include <cuda/std/source_location>
 
 #  include <cuda.h>
 
 #  include <cuda/std/__cccl/prologue.h>
 
 _CCCL_BEGIN_NAMESPACE_CUDA_DRIVER
+
+template <class _Tp>
+struct [[nodiscard]] __driver_call_result
+{
+  _Tp __value_;
+  ::CUresult __error_;
+};
+
+template <>
+struct [[nodiscard]] __driver_call_result<void>
+{
+  ::CUresult __error_;
+};
+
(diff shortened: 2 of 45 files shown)


libcudacxx/include/cuda/__driver/driver_api.h
```diff
@@ -165,281 +210,287 @@ __get_driver_entry_point(const char* __name, [[maybe_unused]] int __major = 12,
...
+  __ret.__error_ = __driver_fn(&__ret.__value_, __attr, __device);
+  return __ret;
 }
 
-[[nodiscard]] _CCCL_HOST_API inline int __deviceGetCount()
+_CCCL_HOST_API inline __call_result<int> __deviceGetCount() noexcept
 {
   static auto __driver_fn = _CCCLRT_GET_DRIVER_FUNCTION(cuDeviceGetCount);
-  int __result;
-  ::cuda::__driver::__call_driver_fn(__driver_fn, "Failed to get device count", &__result);
```
davebayer, line 211: With this implementation, we are loosing the API name and error messages. I am wondering whether we could just add a `__api_` and `__error_msg_` members to the `__call_result` structure and use it in `__check_and_extract_result` function, it should be really easy for the compiler to optimize it out

fbusato:
ok, now I'm confused. In [another PR](https://github.com/NVIDIA/cccl/pull/6325#discussion_r2492026820) you suggested me to avoid `std::span` and use raw arrays to keep the implementation minimal. In this PR, we are introducing `std::expected` which is way more complex.

leofang:
Non-blocking comments:
- I discussed with @pciolkosz a while back about this, my opinion is that the entire `driver_api.h` should just be auto-generated. We can already do this same thing for `cuda-bindings` ([here](https://github.com/NVIDIA/cuda-python/blob/main/cuda_bindings/cuda/bindings/_bindings/cydriver.pyx.in)), so this is trivial, and we have [a plan](https://github.com/NVIDIA/cuda-python/issues/1228) to expand it further 
- it has occurred to me that `_CCCLRT_GET_DRIVER_FUNCTION` does not cache the retrieved function pointers. There will be a small performance penalty. (It's the reason that we cached them in `cuda-bindings`.)

davebayer:
> ok, now I'm confused. In [another PR](https://github.com/NVIDIA/cccl/pull/6325#discussion_r2492026820) you suggested me to avoid `std::span` and use raw arrays to keep the implementation minimal. In this PR, we are introducing `std::expected` which is way more complex.

_... returns `std::expected`-like type ..._

> I discussed with @pciolkosz a while back about this, my opinion is that the entire `driver_api.h` should just be auto-generated. 

I don't think the file to auto generate the header would contain less code, because we still want to adapt the API to our needs.

> it has occurred to me that `_CCCLRT_GET_DRIVER_FUNCTION` does not cache the retrieved function pointers

We store the function addresses in `static` members inside the functions, `_CCCLRT_GET_DRIVER_FUNCTION` is called only once per program launch.

leofang:
> I don't think the file to auto generate the header would contain less code, because we still want to adapt the API to our needs.

It is about avoiding hand-writing `driver_api.h` in a future-proof, safer fashion. The call sites still needs to be hand-written. 

Hand-writing bindings are error prone and not scalable as we expand the project scope and coverage.   

> We store the function addresses in `static` members inside the functions, `_CCCLRT_GET_DRIVER_FUNCTION` is called only once per program launch.

Function-local `static` variables are what I missed, thanks!

cudax/include/cuda/experimental/__launch/configuration.cuh
```diff
@@ -319,32 +319,34 @@ template <class _Tp>
...
 {
-  ::cudaError_t __status = ::cudaSuccess;
-
   // Since CUDA 12.4, querying CU_FUNC_ATTRIBUTE_SHARED_SIZE_BYTES requires the function to be loaded.
-  if (::cuda::__driver::__version_at_least(12, 4))
+  const auto __needs_load = ::cuda::__driver::__version_at_least(12, 4);
+  if (__needs_load.__error_ != ::CUDA_SUCCESS)
+  {
+    return static_cast<::cudaError_t>(__needs_load.__error_);
+  }
```
miscco, line 333: Why are these not `_CCCL_TRY_DRIVER_API`?

libcudacxx/test/libcudacxx/cuda/ccclrt/device/device_smoke.c2h.cu
```diff
@@ -36,7 +36,7 @@ C2H_CCCLRT_TEST("init", "[device]")
 {
   cuda::device_ref dev{0};
   dev.init();
-  CCCLRT_REQUIRE(cuda::__driver::__isPrimaryCtxActive(cuda::__driver::__deviceGet(0)));
+  CCCLRT_REQUIRE(_CCCL_TRY_DRIVER_API(__isPrimaryCtxActive(_CCCL_TRY_DRIVER_API(__deviceGet(0)))));
```
miscco: Critical: We do not want to use internal macros code in tests.

Those should be written as a user would write them

libcudacxx/include/cuda/__driver/driver_api.h
```diff
@@ -27,13 +27,55 @@
...
+#  include <cuda/std/source_location>
 
 #  include <cuda.h>
 
 #  include <cuda/std/__cccl/prologue.h>
 
 _CCCL_BEGIN_NAMESPACE_CUDA_DRIVER
 
+template <class _Tp>
+struct [[nodiscard]] __call_result
```
miscco: Nitpick: I believe this name is too generic
```suggestion
struct [[nodiscard]] __driver_call_result
```

libcudacxx/include/cuda/__driver/driver_api.h
```diff
@@ -27,13 +27,55 @@
...
+
+template <>
+struct [[nodiscard]] __call_result<void>
+{
+  ::CUresult __error_;
+};
+
+template <class _Tp>
+[[nodiscard]] _CCCL_HOST_API constexpr _Tp __check_and_extract_result(
+  __call_result<_Tp> __result, ::cuda::std::source_location __loc = ::cuda::std::source_location::current())
```
miscco: Question should this take by 
```suggestion
  __call_result<_Tp>&& __result, ::cuda::std::source_location __loc = ::cuda::std::source_location::current())
```

and `forward_like` below?

libcudacxx/include/cuda/__driver/driver_api.h
```diff
@@ -27,13 +27,55 @@
...
 #  include <cuda/std/__cccl/prologue.h>
 
 _CCCL_BEGIN_NAMESPACE_CUDA_DRIVER
 
+template <class _Tp>
+struct [[nodiscard]] __call_result
+{
+  ::CUresult __error_;
+  _Tp __value_;
+};
```
miscco, line 44: I believe this is missing at least one feature

```suggestion
  _CCCL_API constexpr operator bool() const noexcept {
    return __error_ == ::CUDA_SUCCESS;
  }
};
```

We might also want to think about at least one monadic interface:
```suggestion
  _CCCL_API constexpr _Tp or_else(_Tp __alternative) const noexcept {
    if (__error_ == ::CUDA_SUCCESS) {
      return __value_;
    } 
    else
    {
      return __alternative;
    }
  }
};
```
This would be handy if there is a fallible API with a sensible default value

miscco (changes requested):

libcudacxx/include/cuda/__driver/driver_api.h
```diff
@@ -132,7 +174,10 @@ _CCCL_HOST_API inline void __call_driver_fn(Fn __fn, const char* __err_msg, Args
 {
   auto __driver_fn = reinterpret_cast<decltype(::cuInit)*>(
     ::cuda::__driver::__get_driver_entry_point_impl(__get_proc_addr_fn, "cuInit", 12, 0));
-  ::cuda::__driver::__call_driver_fn(__driver_fn, "Failed to initialize CUDA Driver", 0);
+  if (auto __status = __driver_fn(0); __status != ::CUDA_SUCCESS)
+  {
+    ::cuda::__throw_cuda_error(static_cast<::cudaError_t>(__status), "Failed to initiazlie CUDA Driver");
```
fbusato: spelling

libcudacxx/include/cuda/__stream/stream.h
```diff
@@ -48,7 +48,7 @@ struct stream : stream_ref
       : stream_ref(::cuda::__invalid_stream())
   {
     [[maybe_unused]] __ensure_current_context __ctx_setter(__dev);
-    __stream = ::cuda::__driver::__streamCreateWithPriority(cudaStreamNonBlocking, __priority);
```
fbusato, line 51: minor concern. This changes the original behavior of the driver call wrt. exceptions. A developer could still think that direct driver calls throw

libcudacxx/include/cuda/__driver/driver_api.h
davebayer: I had it like this before, but changed it to this name, because it's in the `cuda::__driver::` namespace. Can rename it back

libcudacxx/include/cuda/__driver/driver_api.h
davebayer: All of the types returned by the driver APIs are trivial, so I don't think it makes any difference

cudax/include/cuda/experimental/__launch/configuration.cuh
davebayer: That's a question for @pciolkosz, all of the functions applying the launch options to a given kernel launch are marked as `noexcept`

libcudacxx/include/cuda/__stream/stream.h
davebayer: That's true. On the other a developer using `cuda::__driver::__meow(...)` call in a `noexcept` function thinking it just does what `cuMeow(...)` would do is even worse in my opinion

libcudacxx/include/cuda/__driver/driver_api.h
davebayer: I don't agree. I don't think the `operator bool` does make things more clean. Examples from the code:
1. Not ideally named variables
```cpp
const auto __needs_load = _CCCL_TRY_DRIVER_API(__version_at_least(12, 4));
if (!__needs_load)
{
  return __needs_load.__error_;
}
```
2. What's actually being checked here?
```cpp
const auto __mem_type = ::cuda::__driver::__pointerGetAttribute<::CU_POINTER_ATTRIBUTE_MEMORY_TYPE>(__p1);
return __mem_type || __mem_type.__value_ == ::CU_MEMORYTYPE_HOST;
```
I expect us to use the macros in 99% of the time. I would suggest just using the `result.__error_` and `result.__value_` members directly for other cases.

cudax/include/cuda/experimental/__launch/configuration.cuh
pciolkosz: This is a leftover consequence of an early design, where I had only the top level interface throwing and internal function always `noexcept` in case we wanted to implement a exception-free alternative top level interface. But we decided its not a goal anymore, so we could just throw here

davebayer:
Closing for now, we can get back to it once it becomes a priority

Issue #4573, opened 2025-04-29
Use device functions with `void*` argument types for iterator advance/dereference methods in c.parallel


shwina:
Similar to the work done in https://github.com/NVIDIA/cccl/pull/4570 for user-defined operators, we need to modify device functions for iterator advance/dereference methods to also accept `void*` arguments.

PR #6634, opened 2025-11-14
Use wrapper with `void*` argument types for iterator advance/dereference signature

shwina:
## Description

Merge after #6353.

Closes https://github.com/NVIDIA/cccl/issues/4573. This PR is shown to resolve issues due to ODR violation, by unmarking a test that was previously marked `xfail`.

## Checklist


c/parallel/test/test_merge_sort.cpp
```diff
@@ -432,12 +432,16 @@ C2H_TEST("DeviceMergeSort::SortKeys works with output iterators", "[merge_sort]"
     make_iterator<TestType, random_access_iterator_state_t>(
       {"random_access_iterator_state_t", "struct random_access_iterator_state_t { int* d_input; };\n"},
       {"advance",
-       "extern \"C\" __device__ void advance(random_access_iterator_state_t* state, unsigned long long offset) {\n"
-       "  state->d_input += offset;\n"
+       "extern \"C\" __device__ void advance(void* state, const void* offset) {\n"
+       "  auto* typed_state = static_cast<random_access_iterator_state_t*>(state);\n"
+       "  auto offset_val = *static_cast<const unsigned long long*>(offset);\n"
+       "  typed_state->d_input += offset_val;\n"
```
bernhardmgruber: Question: why not use a raw string literal?

shwina: We can. Changed to using raw strings in https://github.com/NVIDIA/cccl/pull/6634/commits/ac415014c58353f012ab5776b7f8ec3588f35f12.

python/cuda_cccl/cuda/compute/iterators/_iterators.py
```diff
@@ -162,22 +163,30 @@ def get_output_dereference_signature(self):
         )
```
NaderAlAwar: Important: let us delete these if they are no longer used

python/cuda_cccl/tests/compute/examples/iterator/transform_output_iterator.py
```diff
@@ -27,7 +27,7 @@ def sqrt(x: np.float32) -> np.float32:
 
 
 # Create transform output iterator
-d_out_it = TransformIterator(d_output, sqrt)
+d_out_it = TransformOutputIterator(d_output, sqrt)
```
NaderAlAwar, line 30: Question: why is this change needed now?

shwina: No idea, and honestly no idea how this ever worked before...

Issue #6466, opened 2025-11-04
Turn --extended-lambda on by default for all examples


brycelelbach:
It's 2025. We should be switching our examples, especially the Thrust ones, to use lambdas instead of function objects. Let's turn --extended-lambda on by default when building all of our examples so we don't need to have a sidecar CMake file for every modern example (see #4462).

fbusato:
I think this is feasible. `--extended-lambda` is supported by CTK 12.x and we don't need NVRTC for examples.
This is could be a "good first issue"

miscco:
![Image](https://github.com/user-attachments/assets/86cac922-d69f-4c1a-a4e5-5b76ab887614)

bernhardmgruber:
I am fully in favor of this!

viralbhadeshiya:
Did someone said "good first issue", I'll take it 😂. 

No but seriously,  I can work on this, if you guys are okay with it. lmk. Thanks.

fbusato:
@viralbhadeshiya would be great. We really appreciate your help.

PR #6635, opened 2025-11-15
Turn --extended-lambda on by default for all examples

viralbhadeshiya:
## Description
- Convert all functor to lambda in examples for thrust, cub and libcudacxx
- Changed exmaple cmake files to build example with --extended-lambda on by default

closes https://github.com/NVIDIA/cccl/issues/6466

## Checklist


cudax/examples/stf/thrust_zip_iterator.cu
```diff
@@ -72,7 +61,13 @@ void thrust_algorithm(context& ctx, ZippedIt& first, ZippedIt& last, OutIt& outp
     // Create a device pointer from the raw pointer
     thrust::device_ptr<int> dout = thrust::device_pointer_cast(dC.data_handle());
 
-    thrust::transform(thrust::cuda::par_nosync.on(stream), dfirst, dlast, dout, my_transform_functor());
+    // Lambda to apply the transformation
+    thrust::transform(
+      thrust::cuda::par_nosync.on(stream), dfirst, dlast, dout, [] __device__(const thrust::tuple<int, char>& t) {
+        int a  = thrust::get<0>(t);
+        char b = thrust::get<1>(t);
+        return a + static_cast<int>(b); // Example operation
```
bernhardmgruber: Suggestion: just destructure the tuple
```suggestion
        const auto [a, b]  = t;
        return a + static_cast<int>(b); // Example operation
```

thrust/examples/arbitrary_transformation.cu
```diff
@@ -69,23 +50,31 @@ int main()
   // apply the transformation
   thrust::for_each(thrust::make_zip_iterator(A.begin(), B.begin(), C.begin(), D1.begin()),
                    thrust::make_zip_iterator(A.end(), B.end(), C.end(), D1.end()),
-                   arbitrary_functor1());
+                   [] __device__(auto t) {
+                     // D[i] = A[i] + B[i] * C[i];
+                     thrust::get<3>(t) = thrust::get<0>(t) + thrust::get<1>(t) * thrust::get<2>(t);
+                   });
```
bernhardmgruber, line 67: Consider:
```suggestion
                   [] __device__(auto t) {
                     // D[i] = A[i] + B[i] * C[i];
                     auto& [a, b, c, d] = t;
                     d = a +b * c;
                   });
```

thrust/examples/repeated_range.cu
```diff
@@ -21,19 +21,12 @@ class repeated_range
...
-    __host__ __device__ difference_type operator()(const difference_type& i) const
-    {
+  // lambda that implements the repeat operation
+  using repeat_functor = decltype([](difference_type repeats) {
+    return [=] __device__(const difference_type& i) {
       return i / repeats;
-    }
-  };
+    };
+  }(difference_type{}));
```
bernhardmgruber: A lamdba in an unevaluated context will require C++20. We discussed turning that on for some examples, but I think here we can rewrite the code a bit to compile under C++17 still. Please make it work in C++17.

thrust/examples/sort.cu
```diff
@@ -148,7 +127,20 @@ int main()
...
+        return x < y;
+      }
+      else if (x % 2)
+      {
+        return false;
+      }
+      else
+      {
+        return true;
+      }
```
bernhardmgruber, line 138: Suggestion: this could be shortened:
```suggestion
      if (x % 2 == y % 2)
      {
        return x < y;
      }
      else
      {
        return !(x % 2);
      }
```

thrust/examples/stream_compaction.cu
```diff
@@ -48,23 +38,28 @@ int main()
   // allocate output storage, here we conservatively assume all values will be copied
   Vector output(values.size());
 
+  // lambda returns true if the argument is odd, and false otherwise
+  auto is_odd = [] __device__(int x) {
+    return x % 2;
+  };
```
bernhardmgruber: Important: the original functor forced the return type to `bool`, but the lambda here isn't. Please add `-> bool` as return type or produce one in the return expression: `x % 2 == 1`

thrust/examples/transform_iterator.cu
```diff
@@ -8,19 +8,11 @@
...
-      : lo(_lo)
-      , hi(_hi)
-  {}
-
-  __host__ __device__ T operator()(T x)
-  {
+  return [=] __device__(T x) {
     if (x < lo)
     {
       return lo;
```
bernhardmgruber: Suggestion: Let's just use `cuda::std::clamp as implementation.

bernhardmgruber:
> CMake LGTM -- @bernhardmgruber can you take another look re: your review comments?

My above comments have not been addressed yet. @viralbhadeshiya please address them. Thx!

viralbhadeshiya:
Thanks @bernhardmgruber @alliepiper , let me take a look at those changes comment. Will make a change and let you know when I am done.

alliepiper:
> Will make a change and let you know when I am done.

Thanks! I'll mark this as a draft for now. Just click "Ready for Review" when you're ready for us to take another look.

viralbhadeshiya:
@bernhardmgruber can you take a look at the changes that I pushed for review, lmk if I got those right, or if you want me to change/add anything.

viralbhadeshiya:
@bernhardmgruber I have no idea, why this is failing, do you have any suggestions where should I take a look to solve this?

bernhardmgruber:
> @bernhardmgruber I have no idea, why this is failing, do you have any suggestions where should I take a look to solve this?

Seems like another artifact of a stale sccache object after updating Catch2. Let's see if updating the branch helps.

cub/examples/device/example_device_partition_if.cu
```diff
@@ -158,7 +143,11 @@ int main(int argc, char** argv)
 
   // Initialize problem and solution
   Initialize(h_in, num_items, max_segment);
-  GreaterThan select_op(h_in[pivot_index]);
+
+  // Lambda for selection operation (greater-than)
+  auto select_op = [compare = h_in[pivot_index]] __device__(const int& a) {
```
gevtushenko, line 148: minor suggestion: in some corner cases, we can disable optimizations if operator takes a reference. To avoid spreading ref-based examples, I'd rather accept `a` by value here and on other places. Reference doesn't add much here, just takes more space leading to same code gen.

cudax/examples/stf/thrust_zip_iterator.cu
```diff
@@ -72,7 +61,12 @@ void thrust_algorithm(context& ctx, ZippedIt& first, ZippedIt& last, OutIt& outp
     // Create a device pointer from the raw pointer
     thrust::device_ptr<int> dout = thrust::device_pointer_cast(dC.data_handle());
 
-    thrust::transform(thrust::cuda::par_nosync.on(stream), dfirst, dlast, dout, my_transform_functor());
+    // Lambda to apply the transformation
+    thrust::transform(
+      thrust::cuda::par_nosync.on(stream), dfirst, dlast, dout, [] __device__(const thrust::tuple<int, char>& t) {
```
gevtushenko, line 66: suggestion: here and in thrust examples, consider using `thrust::make_zip_function` to make lambda accept `(int a, char b)`.

gevtushenko: suggestion: there are some caveats to device lambdas. In some cases, thrust has to introspect the return type for user operator, which is not allowed for extended device lambdas. It's not the case for host-device lambdas or lambdas with trailing return type. To simplify users lives, I'd suggest using `__host__ __device__` here and the rest of the examples. This way the code is copy paste-able for other execution policies, but also doesn't break on users in other algorithms.

thrust/examples/arbitrary_transformation.cu
```diff
@@ -69,23 +60,32 @@ int main()
   // apply the transformation
   thrust::for_each(thrust::make_zip_iterator(A.begin(), B.begin(), C.begin(), D1.begin()),
                    thrust::make_zip_iterator(A.end(), B.end(), C.end(), D1.end()),
-                   arbitrary_functor1());
+                   [] __device__(auto t) {
```
gevtushenko, line 63: suggestion: same here and the rest of the thrust examples, let's use host-device.

thrust/examples/norm.cu
```diff
@@ -17,28 +17,22 @@
...
+  };
+
   // setup arguments
-  square<float> unary_op;
   cuda::std::plus<float> binary_op;
   float init = 0;
 
   // compute norm
-  float norm = std::sqrt(thrust::transform_reduce(d_x.begin(), d_x.end(), unary_op, init, binary_op));
+  float norm = std::sqrt(thrust::transform_reduce(d_x.begin(), d_x.end(), square, init, binary_op));
```
gevtushenko, line 35: suggestion: not required for this PR, but I'd consider changing this example to use `reduce_into` with transform iterator and transform output iterator. Found this to be helpful for people more frequently while being less discoverable.

thrust/examples/discrete_voronoi.cu
```diff
@@ -22,102 +22,89 @@
...
-};
+  if (y + k < n)
+  {
+    v = minVoro(x, y, v, cuda::std::get<2>(t));
+  }
+
+  // global return
+  return v;
+}
+}
```
gevtushenko, line 106: question: is this `operator()` a left over?

thrust/examples/scan_matrix_by_rows.cu
```diff
@@ -50,6 +36,9 @@ __host__ void scan_matrix_by_rows1(thrust::device_vector<int>& u, int n, int m)
   // We construct a `thrust::transform_iterator` which applies the `which_row`
   // function object to the index of each element.
   thrust::transform_iterator<which_row, thrust::counting_iterator<int>> t_first(c_first, which_row(m));
+  auto t_first = thrust::make_transform_iterator(c_first, [m] __device__(int idx) {
```
gevtushenko, line 39: question: both are called `t_first`, how does it compile?

thrust/examples/strided_range.cu
```diff
@@ -21,19 +21,12 @@ class strided_range
...
-    difference_type stride;
-
-    stride_functor(difference_type stride)
-        : stride(stride)
-    {}
-
-    __host__ __device__ difference_type operator()(const difference_type& i) const
-    {
+  // lambda that implements the stride operation
+  using stride_functor = decltype([](difference_type stride) {
```
gevtushenko, line 25: questoin: If I understand correctly, all lambdas have unique type, not sure how this using would work

thrust/examples/minmax.cu
```diff
@@ -64,9 +33,22 @@ int main()
   }
   thrust::device_vector<int> data = host_data;
 
-  // setup arguments
-  minmax_unary_op<int> unary_op;
-  minmax_binary_op<int> binary_op;
+  // lambda that takes in a value x and returns a minmax_pair whose minimum and maximum values are initialized to x
+  auto unary_op = [] __device__(const int& x) {
```
gevtushenko, line 37: important: this operator is invoked on host side (below), must be `__host__ __device__`

cub/examples/device/example_device_partition_if.cu
gevtushenko: important: this lambda is invoked on the host side (in the solution below), must be `__host__ __device__`

thrust/examples/remove_points2d.cu
```diff
@@ -58,7 +58,12 @@ int main()
   size_t new_size =
     thrust::remove_if(thrust::make_zip_iterator(x.begin(), y.begin()),
                       thrust::make_zip_iterator(x.end(), y.end()),
-                      is_outside_circle<float>())
+                      [] __device__(const auto& tuple) {
```
gevtushenko, line 61: important: this is called on host vectors, must be `__host__ __device__` or with no execution space specifier

thrust/examples/remove_points2d.cu
```diff
@@ -58,7 +58,12 @@ int main()
...
+                        // unpack the tuple into x and y coordinates
+                        const float x = thrust::get<0>(tuple);
+                        const float y = thrust::get<1>(tuple);
+                        return x * x + y * y > 1;
+                      })
```
gevtushenko, line 66: important: `is_outside_circle` is a dead code now, please remove

thrust/examples/padded_grid_reduction.cu
```diff
@@ -104,11 +104,40 @@ int main()
   std::cout << "\n";
 
   // compute min & max over valid region of the 2d grid
+  using InputTuple  = cuda::std::tuple<int, float>;
+  using OutputTuple = cuda::std::tuple<bool, float, float>;
   using result_type = cuda::std::tuple<bool, float, float>;
 
+  // lambda to transform a tuple (int,value) into a tuple (bool,value,value)
+  // where the bool is true for valid grid values and false for values in the padded region
+  auto unary_op = [n, N] __device__(const InputTuple& t) {
```
gevtushenko, line 113: important: `transform_tuple` and `reduce_tuple` are dead code now, please remove

thrust/examples/arbitrary_transformation.cu
gevtushenko: important: `arbitrary_functor1` is dead code now, please remove

cudax/examples/stf/thrust_zip_iterator.cu
bernhardmgruber: > suggestion: here and in thrust examples, consider using `thrust::make_zip_function` to make lambda accept `(int a, char b)`.

And even better: `cuda::make_zip_function`.

PR #6637, opened 2025-11-16
[on hold] Drop tests for `thrust::pair` and `thrust::tuple`

bernhardmgruber:
Those are alias to `cuda::std` anyway, which is tested already

@miscco please let me know if you would like to move any Thrust tests to libcu++.


davebayer (approved):

miscco (changes requested):
I would like to keep them for as long as we have `thrust::pair` and `thrust::tuple` in the library

Issue #6482, opened 2025-11-04
[FEA]: Enable `__grid_constant__` parameters in CUB kernels


fbusato:
### Is this a duplicate?

- [x] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cccl/issues) for this request and that I agree to the [Code of Conduct](CODE_OF_CONDUCT.md)

### Area

CUB

### Is your feature request related to a problem? Please describe.

`__grid_constant__` prevents the compiler from creating a per-thread copy of a parameter. Grid constant annotation allows threads to access the parameter through a single address, which can improve performance.

Considering CUB, I don't expect big performance improvements because most (if not all) functions are inlined. Still, some cases could benefit from it and it is considered a good practice.

see also, https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#grid-constant

### Describe the solution you'd like

Apply `const __grid_constant__` (`_CCCL_GRID_CONSTANT` for portability) to all non-mutable parameters in CUB kernels.

toxicteddy00077:
Hi I think I can help out. May I give this a try?

fbusato:
@toxicteddy00077 would be great. Let me assign the issue to you.

PR #6642, opened 2025-11-17
Fea/6482 enable grid constant for all non-mutable CUB kernel parameters

toxicteddy00077:
## Description

closes  #6482 

Just added `_CCCL_GRID_CONSTANT const` to the immutable params for the necessary kernels


bernhardmgruber:
Thank you for this contribution! I think we should also do a benchmark and a SASS check somewhere, just because I am curious about the impact of this change. @toxicteddy00077 you can leave this to us.

cub/cub/device/dispatch/kernels/kernel_for_each.cuh
```diff
@@ -86,7 +86,8 @@ using can_regain_copy_freedom =
 
 // This kernel is used when the block size is not known at compile time
 template <class ChainedPolicyT, class OffsetT, class OpT>
-CUB_DETAIL_KERNEL_ATTRIBUTES void dynamic_kernel(OffsetT num_items, OpT op)
+CUB_DETAIL_KERNEL_ATTRIBUTES void
+dynamic_kernel(_CCCL_GRID_CONSTANT const OffsetT num_items, _CCCL_GRID_CONSTANT const OpT op)
```
fbusato: @bernhardmgruber Rare but possible. I guess this is problematic if the operator has an internal state (non-`const`).

cub/cub/device/dispatch/kernels/kernel_histogram.cuh
```diff
@@ -445,17 +445,17 @@ template <typename ChainedPolicyT,
           typename OffsetT>
 __launch_bounds__(int(ChainedPolicyT::ActivePolicy::AgentHistogramPolicyT::BLOCK_THREADS))
   CUB_DETAIL_KERNEL_ATTRIBUTES void DeviceHistogramSweepKernel(
-    SampleIteratorT d_samples,
+    _CCCL_GRID_CONSTANT const SampleIteratorT d_samples,
     ::cuda::std::array<int, NumActiveChannels> num_output_bins_wrapper,
```
fbusato: why `_CCCL_GRID_CONSTANT` is skipped on `cuda::std::array` parameters?

cub/cub/device/dispatch/kernels/kernel_merge_sort.cuh
```diff
@@ -272,15 +272,15 @@ __launch_bounds__(
...
-    OffsetT keys_count,
+    _CCCL_GRID_CONSTANT const bool ping,
+    _CCCL_GRID_CONSTANT const KeyIteratorT keys_ping,
+    _CCCL_GRID_CONSTANT const ValueIteratorT items_ping,
+    _CCCL_GRID_CONSTANT const OffsetT keys_count,
     KeyT* keys_pong,
     ValueT* items_pong,
-    CompareOpT compare_op,
+    _CCCL_GRID_CONSTANT const CompareOpT compare_op,
     OffsetT* merge_partitions,
```
fbusato: is it a *write* memory location? or `_CCCL_GRID_CONSTANT` was not applied to pointers

cub/cub/device/dispatch/kernels/kernel_radix_sort.cuh
```diff
@@ -85,11 +85,11 @@ __launch_bounds__(int((ALT_DIGIT_BITS) ? int(ChainedPolicyT::ActivePolicy::AltUp
   CUB_DETAIL_KERNEL_ATTRIBUTES void DeviceRadixSortUpsweepKernel(
     const KeyT* d_keys,
```
fbusato: `_CCCL_GRID_CONSTANT` is should be fine on `const` pointers (if not modified)

fbusato (changes requested):
thanks @toxicteddy00077. It is great to see this feature in CUB.

Initial feedbacl: This work could be extended to more parameters, e.g. `cuda::std::array` and pointers. Also, we need to double-check the implications for custom operators

cub/cub/device/dispatch/kernels/kernel_histogram.cuh
toxicteddy00077: I tried it earlier, but when i tested the benches I ran into errors regarding `AgentHistogram` since the `.data()` function in it is expected to return type `CounterT**` but with `_CCCL_GRID_CONSTANT const` seems to return something else. Please let me know if I've missed something

cub/cub/device/dispatch/kernels/kernel_for_each.cuh
bernhardmgruber: Good catch! I think we definitely should not mark operators as `const`. Lot's of them have a mutable `operator()`. Please remove `_CCCL_GRID_CONSTANT const` from all operators.

cub/cub/device/dispatch/kernels/kernel_merge_sort.cuh
toxicteddy00077: From what i understand the `_CCCL_GRID_CONSTANT const` was applied to the outer pointers, not the inner pointers which actually point to the *write* memory.

cub/cub/device/dispatch/kernels/kernel_radix_sort.cuh
toxicteddy00077: For this too, I ran into errors when I ran the benches after I added `_CCCL_GRID_CONSTANT`

cub/cub/device/dispatch/kernels/kernel_for_each.cuh
toxicteddy00077: I have removed `_CCCL_GRID_CONSTANT const` from the operators

cub/cub/device/dispatch/kernels/kernel_histogram.cuh
fbusato: I didn't check the code. It is a bit unexpected that we modify internal `cuda::std::array` pointers. Maybe @bernhardmgruber knows more about this point.

Excluding pointers, other `::cuda::std::array` parameters should work.

cub/cub/device/dispatch/kernels/kernel_merge_sort.cuh
```diff
@@ -167,15 +167,15 @@ __launch_bounds__(
...
+    _CCCL_GRID_CONSTANT const bool ping,
+    _CCCL_GRID_CONSTANT const KeyInputIteratorT keys_in,
+    _CCCL_GRID_CONSTANT const ValueInputIteratorT items_in,
+    _CCCL_GRID_CONSTANT const KeyIteratorT keys_out,
+    _CCCL_GRID_CONSTANT const ValueIteratorT items_out,
+    _CCCL_GRID_CONSTANT const OffsetT keys_count,
     KeyT* tmp_keys_out,
     ValueT* tmp_items_out,
-    CompareOpT compare_op,
+    _CCCL_GRID_CONSTANT const CompareOpT compare_op,
```
fbusato: side effects here are extremely rare, but I would remove `_CCCL_GRID_CONSTANT` from `compare_op`

cub/cub/device/dispatch/kernels/kernel_radix_sort.cuh
```diff
@@ -243,11 +243,11 @@ __launch_bounds__(int((ALT_DIGIT_BITS) ? int(ChainedPolicyT::ActivePolicy::AltDo
     const ValueT* d_values_in,
```
fbusato: question. Why `_CCCL_GRID_CONSTANT` cannot be applied to `const T*` pointer?
note, for const-correctness you have to write
```cpp
_CCCL_GRID_CONSTANT const ValueT* const d_values_in,
```
even better if you introduce a new macro for pointers

fbusato (changes requested):

cub/cub/device/dispatch/kernels/kernel_histogram.cuh
bernhardmgruber: I assume this had historical reasons, because we could not change `AgentHistogram` to take pointers by `const` before CCCL 3.0, where we moved them to a `detail` namespace.

Please add `_CCCL_GRID_CONSTANT const` to the `::cuda::std::array` for histogram and propagatge `const` throughout the agent.

cub/cub/device/dispatch/kernels/kernel_merge_sort.cuh
fbusato: yes, this is the reason I asked. It doesn't matter if the pointer is for read or write

cub/cub/device/dispatch/kernels/kernel_radix_sort.cuh
fbusato: got it. What errors? is `d_keys` modified, e.g. `d_keys += value`? or const propagation?

cub/cub/device/dispatch/kernels/kernel_merge_sort.cuh
toxicteddy00077: done

cub/cub/device/dispatch/kernels/kernel_radix_sort.cuh
toxicteddy00077: i have made the changes

cub/cub/device/dispatch/kernels/kernel_radix_sort.cuh
toxicteddy00077: sorry this was an oversight on my part. I wasn't applying `const` correctly to the pointers. i have made the necessary improvements.

toxicteddy00077:
Apologies for the delay. I will be making changes and adding `_CCCL_GRID_CONSTANT const` to `::cuda::std::array` for histogram and make sure it propogates `const`

cub/cub/device/dispatch/kernels/kernel_histogram.cuh
```diff
@@ -445,17 +445,17 @@ template <typename ChainedPolicyT,
           typename OffsetT>
 __launch_bounds__(int(ChainedPolicyT::ActivePolicy::AgentHistogramPolicyT::BLOCK_THREADS))
   CUB_DETAIL_KERNEL_ATTRIBUTES void DeviceHistogramSweepKernel(
```
fbusato, line 446: the previous kernel `DeviceHistogramInitKernel` misses `_CCCL_GRID_CONSTANT`

cub/cub/device/dispatch/kernels/kernel_radix_sort.cuh
```diff
@@ -327,14 +327,14 @@ template <typename ChainedPolicyT,
           typename DecomposerT = identity_decomposer_t>
 __launch_bounds__(int(ChainedPolicyT::ActivePolicy::SingleTilePolicy::BLOCK_THREADS), 1)
   CUB_DETAIL_KERNEL_ATTRIBUTES void DeviceRadixSortSingleTileKernel(
-    const KeyT* d_keys_in,
+    _CCCL_GRID_CONSTANT const KeyT* const d_keys_in,
     KeyT* d_keys_out,
-    const ValueT* d_values_in,
+    _CCCL_GRID_CONSTANT const ValueT* const d_values_in,
     ValueT* d_values_out,
     OffsetT num_items,
```
fbusato, line 343: @bernhardmgruber what do you think about `num_items`? `_CCCL_GRID_CONSTANT` should prevent the issue described in the comment below

cub/cub/device/dispatch/kernels/kernel_radix_sort.cuh
```diff
@@ -466,15 +473,15 @@ CUB_DETAIL_KERNEL_ATTRIBUTES void __launch_bounds__(ChainedPolicyT::ActivePolicy
     AtomicOffsetT* d_lookback,
     AtomicOffsetT* d_ctrs,
     OffsetT* d_bins_out,
-    const OffsetT* d_bins_in,
+    _CCCL_GRID_CONSTANT const OffsetT* const d_bins_in,
```
fbusato, line 486: `DeviceRadixSortExclusiveSumKernel` (below) is missing

cub/cub/device/dispatch/kernels/kernel_reduce.cuh
```diff
@@ -214,11 +214,12 @@ template <typename ChainedPolicyT,
           typename TransformOpT = ::cuda::std::identity>
 CUB_DETAIL_KERNEL_ATTRIBUTES __launch_bounds__(
   int(ChainedPolicyT::ActivePolicy::SingleTilePolicy::BLOCK_THREADS),
-  1) void DeviceReduceSingleTileKernel(InputIteratorT d_in,
-                                       OutputIteratorT d_out,
-                                       OffsetT num_items,
+  1) void DeviceReduceSingleTileKernel(_CCCL_GRID_CONSTANT const InputIteratorT d_in,
+<<<<<<< HEAD
```
fbusato:
```suggestion
```

cub/cub/device/dispatch/kernels/kernel_scan.cuh
```diff
@@ -140,13 +141,13 @@ template <typename ChainedPolicyT,
...
   CUB_DETAIL_KERNEL_ATTRIBUTES void DeviceScanKernel(
-    InputIteratorT d_in,
-    OutputIteratorT d_out,
+    _CCCL_GRID_CONSTANT const InputIteratorT d_in,
+    _CCCL_GRID_CONSTANT const OutputIteratorT d_out,
     ScanTileStateT tile_state,
-    int start_tile,
+    _CCCL_GRID_CONSTANT const int start_tile,
     ScanOpT scan_op,
     InitValueT init_value,
```
fbusato:
```suggestion
    _CCCL_GRID_CONSTANT const InitValueT init_value,
```

cub/cub/device/dispatch/kernels/kernel_merge_sort.cuh
```diff
@@ -167,12 +167,12 @@ __launch_bounds__(
...
-    KeyIteratorT keys_out,
-    ValueIteratorT items_out,
-    OffsetT keys_count,
+    _CCCL_GRID_CONSTANT const bool ping,
+    _CCCL_GRID_CONSTANT const KeyInputIteratorT keys_in,
+    _CCCL_GRID_CONSTANT const ValueInputIteratorT items_in,
+    _CCCL_GRID_CONSTANT const KeyIteratorT keys_out,
+    _CCCL_GRID_CONSTANT const ValueIteratorT items_out,
+    _CCCL_GRID_CONSTANT const OffsetT keys_count,
     KeyT* tmp_keys_out,
```
fbusato:
```suggestion
    _CCCL_GRID_CONSTANT KeyT* const tmp_keys_out,
```
does it work?

cub/cub/device/dispatch/kernels/kernel_transform.cuh
```diff
@@ -984,12 +984,12 @@ template <typename MaxPolicy,
           typename... RandomAccessIteartorsIn>
 __launch_bounds__(get_block_threads<typename MaxPolicy::ActivePolicy>)
   CUB_DETAIL_KERNEL_ATTRIBUTES void transform_kernel(
-    Offset num_items,
-    int num_elem_per_thread,
+    _CCCL_GRID_CONSTANT const Offset num_items,
+    _CCCL_GRID_CONSTANT const int num_elem_per_thread,
     [[maybe_unused]] bool can_vectorize,
```
fbusato:
```suggestion
    _CCCL_GRID_CONSTANT [[maybe_unused]] bool can_vectorize,
```

cub/cub/device/dispatch/kernels/kernel_merge_sort.cuh
fbusato: if yes, please propage it to other occurences

fbusato (changes requested):

cub/cub/device/dispatch/kernels/kernel_scan.cuh
toxicteddy00077: I'm not sure we can apply ` _CCCL_GRID_CONSTANT` to a non `const` type .  Even when I did apply the `_CCCL_GRID_CONSTANT const` it resulted in a error, since `InitValueT` seems to be type converted, and as the type conversion isn't for `const` it fails. The change we can probably make is in https://github.com/NVIDIA/cccl/blob/b8a0a9598f43612fb5d661c6f896fd8db6a34e5d/cub/cub/util_type.cuh#L246C1-L246C51 where we make this conversion operator take in `const`

cub/cub/device/dispatch/kernels/kernel_merge_sort.cuh
toxicteddy00077: Yes this actually worked, I have added it over to multiple types `KeyT`, `ValueT`, `OffsetT`, etc

cub/cub/device/dispatch/kernels/kernel_reduce.cuh
toxicteddy00077: accident sorry.

cub/cub/device/dispatch/kernels/kernel_transform.cuh
toxicteddy00077: done

cub/cub/device/dispatch/kernels/kernel_scan.cuh
fbusato: suggestion updated

cub/cub/device/dispatch/kernels/kernel_segmented_sort.cuh
```diff
@@ -425,15 +425,15 @@ template <SortOrder Order,
           typename OffsetT>
 __launch_bounds__(ChainedPolicyT::ActivePolicy::LargeSegmentPolicy::BLOCK_THREADS)
   CUB_DETAIL_KERNEL_ATTRIBUTES void DeviceSegmentedSortKernelLarge(
-    const local_segment_index_t* d_segments_indices,
+    _CCCL_GRID_CONSTANT const local_segment_index_t* const d_segments_indices,
     const KeyT* d_keys_in_orig,
```
fbusato, line 470: can we also apply _`CCCL_GRID_CONSTANT` here?

fbusato (changes requested):
good progress! Most of the kernel parameters now have the grid constant attribute. There still a couple of open issues. After that, we can merge the PR

cub/cub/device/dispatch/kernels/kernel_scan.cuh
toxicteddy00077: Yes I tried this earlier, but `InitValueT` is type converted and I think it requires me to change the operator in https://github.com/NVIDIA/cccl/blob/b8a0a9598f43612fb5d661c6f896fd8db6a34e5d/cub/cub/util_type.cuh#L246C1-L246C51 to make it take `const` type

cub/cub/device/dispatch/kernels/kernel_segmented_sort.cuh
toxicteddy00077: yes i have applied it successfully in the last commit

cub/cub/device/dispatch/kernels/kernel_radix_sort.cuh
toxicteddy00077: @fbusato `num_items` is modified in this kernel

cub/cub/device/dispatch/kernels/kernel_histogram.cuh
toxicteddy00077: I have added `_CCCL_GRID_CONSTANT const` to the `::cuda::std::array`, just had to make the necessary `AgentHistogram` members `const`(I marked them as read-only) and also make some kernel_histogram.cuh methods in `Transform` which are utilized as `const`

cub/cub/device/dispatch/kernels/kernel_histogram.cuh
toxicteddy00077: made corrections

cub/cub/device/dispatch/kernels/kernel_radix_sort.cuh
toxicteddy00077: made corrections

toxicteddy00077:
@fbusato I think I have made most of the changes as specified. If there are any changes I missed or need to make further, do let me know, I'll get to it ASAP

toxicteddy00077:
Hi @fbusato @bernhardmgruber just curious what is the current state of this PR? Anything to improve on?

toxicteddy00077:
Fixed a small oversight on my part sorry. should work now @fbusato

toxicteddy00077:
@bernhardmgruber I had a look at the CI tests but Im not exactly sure why they're failing. Is it just because of standard incompatibility or something I'm missing?

fbusato:
about 
```cpp
// error: an attribute list cannot appear here
_CCCL_GRID_CONSTANT [[maybe_unused]] const bool can_vectorize,
```
probably should be `[[maybe_unused]] _CCCL_GRID_CONSTANT`.

toxicteddy00077:
@fbusato thanks I was not aware if this. I've made the changes

bernhardmgruber:
Please rebase and fix the merge conflict.

toxicteddy00077:
I'll just get to it

toxicteddy00077:
Hi @bernhardmgruber anything further to be done?

fbusato:
@toxicteddy00077 Bernhard is OOTO this week. I don't see a clear reason why the CI is failing. Let me investigate a bit.

fbusato (approved):

bernhardmgruber:
> Hi @bernhardmgruber anything further to be done?

Yes, please rebase and resolve conflicts ;)

> I don't see a clear reason why the CI is failing

I think we are hitting timeouts because the changeset leads to too many recompilations. This should be addressed with a rebase as well.

toxicteddy00077:
All done @bernhardmgruber

cub/cub/agent/agent_histogram.cuh
```diff
@@ -226,12 +226,14 @@ struct AgentHistogram
   _TempStorage& temp_storage;
   WrappedSampleIteratorT d_wrapped_samples; // with cache modifier applied, if possible
   SampleT* d_native_samples; // possibly nullptr if unavailable
-  int* num_output_bins; // one for each channel
-  int* num_privatized_bins; // one for each channel
+  const int* num_output_bins; // one for each channel (read-only)
+  const int* num_privatized_bins; // one for each channel (read-only)
```
bernhardmgruber: Please remove the added comment, since it does not convey any additional information that `const` does not already spell.
```suggestion
  const int* num_output_bins; // one for each channel
  const int* num_privatized_bins; // one for each channel
```

cub/cub/agent/agent_histogram.cuh
```diff
@@ -226,12 +226,14 @@ struct AgentHistogram
...
   CounterT* d_privatized_histograms[NumActiveChannels]; // one for each channel
   CounterT** d_output_histograms; // in global memory
-  OutputDecodeOpT* output_decode_op; // determines output bin-id from privatized counter index, one for each channel
-  PrivatizedDecodeOpT* privatized_decode_op; // determines privatized counter index from sample, one for each channel
+  const OutputDecodeOpT* output_decode_op; // determines output bin-id from privatized counter index, one for each
+                                           // channel (read-only)
+  const PrivatizedDecodeOpT* privatized_decode_op; // determines privatized counter index from sample, one for each
+                                                   // channel (read-only)
```
bernhardmgruber: Same here

bernhardmgruber (approved):

cub/cub/agent/agent_histogram.cuh
toxicteddy00077: must have missed that. sorry

toxicteddy00077:
done @bernhardmgruber

bernhardmgruber:
Some CI jobs are getting stuck compiling now. I have seen this previously with the warpspeed scan, where adding `_CCCL_GRID_CONSTANT` to one specific kernel parameter caused nvcc < 12.8 to get stuck. See:

https://github.com/NVIDIA/cccl/blob/94bd6e40d36a765a9bc6db69a60619092b41e228/cub/cub/device/dispatch/kernels/kernel_scan.cuh#L187-L190

@fbusato what do you want to do? Use `_CCCL_GRID_CONSTANT` only for nvcc >= 12.8 or start finding out which line(s) of this PR cause an issue?

fbusato:
thanks, @bernhardmgruber. This is very helpful to address the problem. Yes, I strongly suggest to enable it only for nvcc >= 12.8. I will track the potential bug offline and chat with the compiler team.

toxicteddy00077:
@fbusato any update on this? thanks

fbusato:
@toxicteddy00077 sorry again, this not a known issue. Please go ahead and enable `_CCCL_GRID_CONSTANT` only for nvcc >= 12.8

toxicteddy00077:
@fbusato there's still one test failing, what could be the reason? Anything i can fix?

fbusato:
this is unrelated to your changes. It has been fixed quite recently. Let me try to rebase the PR.

fbusato (approved):

fbusato:
Python nvcc GCC / SQ / [CTK13.1 GCC13 py3.13] Test cuda.compute(amd64, L4)
> AssertionError: LDL instruction found in SASS

@NaderAlAwar  could you please check?

NaderAlAwar:
@fbusato Looked into this. The LDL/STL instructions appear even with NVCC, this is not specific to `cuda.compute`. I ran some benchmarks and the performance after the changes in this PR are 0-2% better, so this is not an issue. I will open a PR to disable the checks on the python side. I think we should disable all the LDL/STL checks until I spend some more time looking into #7978.

NaderAlAwar:
Disabled sass checks in https://github.com/NVIDIA/cccl/pull/8053. Once it lands, you should merge from main to fix CI

toxicteddy00077:
@NaderAlAwar @fbusato seems to be still failing. please take a look

fbusato:
@toxicteddy00077 thanks again for the contribution! We thought creating the issue would be a relatively easy task. However, it ended up being really tough.

toxicteddy00077:
@fbusato thanks for the opportunity. looking forward to helping around CCCL  more

elstehle:
I've been running into some issues related to `__grid_constant__` annotated parameters (specifically annotated pointers) recently. I wanted to check whether anybody has run a full benchmark sweep from before and after this PR yet. I just want to be sure we would catch performance regressions, if there were any.

The `LDL`s flagged by CI may have a symptom of this (though it seems that in this specific case it didn't materialize in a performance delta). @NaderAlAwar do you remember which benchmark(s) you ran?

NaderAlAwar:
@elstehle looking at the CI history https://github.com/NVIDIA/cccl/actions/runs/23012608221/job/66829176437?pr=6642, it seems that merge sort pairs with key type = `__half` and value type = `float32` was what triggered this. I investigated it at the time and IIRC, prior to this change, there were `LDL`s for when `OffsetT` was 32 bits but not 64 bits. Following this change, the reverse happened, `LDL`s for 64 bits but not 32 bits (which is what triggered the CI failure at the time since we hardcode offset types to 64 bits in python).

That said, the performance of merge sort seemed unaffected by the presence or absence of `LDL`s in this specific case.

bernhardmgruber:
QA reports several regressions with CCCL 3.4 (compared to 3.3) and I could already pin two of them down to this PR (see #9763 and #9765). I think we need to revert these changes. @fbusato for vis.

bernhardmgruber:
We had to revert this entire PR in #9858, because of several regressions reported by QA. The cause seems to be a compiler bug that optimizes differently in the presence of grid constant annotations. This will likely be fixed in the future and we will be able to re-introduce grid annotations at some point.

PR #6678, opened 2025-11-18
Add host std lib version detection

davebayer:
This PR adds support for host's standard library version detection. I need it for #6424.

### DIFF
--- a/libcudacxx/include/cuda/std/__cccl/host_std_lib.h
+++ b/libcudacxx/include/cuda/std/__cccl/host_std_lib.h
@@ -23,9 +23,9 @@
 #  pragma system_header
 #endif // no system header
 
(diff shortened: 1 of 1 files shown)


fbusato (approved):
not sure how it is related to the other PR. Anyway, the changes make sense

davebayer:
> not sure how it is related to the other PR. Anyway, the changes make sense

nvc++ requires libstdc++ 12+ to support `if consteval`

fbusato:
...things that I will never understand the reasons...

davebayer:
> ...things that I will never understand the reasons...

It's because EDG frontend wants to be compatible gcc which added support for `if consteval` in version 12. Makes no sense to me, but here we are.

PR #6682, opened 2025-11-18
Add `cuda::execution::guarantee` and `cuda::execution::segment_size::max_segment_size`

### LINKED ISSUE #6620: [FEA]: Design Guarantees API
### Is this a duplicate?

### Area

CUB

As of https://github.com/NVIDIA/cccl/pull/4877 we have requirements API. This interface allows user to make an algorithm guarantee something:

```c++
cuda::execution::require(cuda::execution::determinism::run_to_run);
```

In the example above, user requres run-to-run determinism and algorithm guarantees it. 

On the implementation side, we sometimes need this relationship reversed. We want an algorithm to require something and user to guarantee that. For instance, we have multiple implementations of segmented reduction. Some implementations require max segment size to be less than a certain value. If we had an interface that allows user to guarantee max segment size, we'd be able to dispatch to optimal algorithm implementation. 

Design gaurantees API:

```c++
cuda::execution::guarantee(max seg size < 42);
```

- [ ] Add `cuda::execution::guarantee`
- [ ] Design max segment size guarantee
- [ ] Leverage guarantees API to optimize segmented reduce

_No response_

### Additional context

_No response_

srinivasyadav18:
## Description

Add `cuda::execution::guarantee` API and `cuda::execution::segment_size::max_segment_size` 

closes https://github.com/NVIDIA/cccl/issues/6620

## Checklist

- [x] Choose proper namespace for `max_segment_size`


libcudacxx/include/cuda/__execution/max_segment_size.h
```diff
@@ -0,0 +1,85 @@
...
+namespace segment_size
+{
+struct __get_max_segment_size_t;
+
+// TODO : add dynamic extent specialization later, when stateful guarantees (env) are supported
+
+//! A class template that can be used to specify the maximum segment size
+//! for segmented algorithms.
+//! \tparam _N The maximum segment size.
+template <size_t _N>
```
jrhemstad: question: Is specifying a maximum segment size only useful when known at compile time? Or is it worth considering enabling specifying a maximum segment size as either a dynamic or static value?

gevtushenko: > // TODO : add dynamic extent specialization later, when stateful guarantees (env) are supported

I think the idea was to add a specialization `template <> max_segment_size<dyn> : __guarantee { size_t value; }` later. There were some compiler bugs that prevent states at the moment. We should design the guarantee in a way that allows adding runtime value.

libcudacxx/include/cuda/__execution/max_segment_size.h
```diff
@@ -0,0 +1,85 @@
...
+#include <cuda/std/__concepts/concept_macros.h>
+#include <cuda/std/__execution/env.h>
+#include <cuda/std/__type_traits/is_one_of.h>
+
+#include <cuda/std/__cccl/prologue.h>
+
+_CCCL_BEGIN_NAMESPACE_CUDA_EXECUTION
+
+namespace segment_size
+{
```
bernhardmgruber: Q: Do we really want a sub namespace here? This get's really long (and redundant to spell):
```
cuda::execution::segment_size::max_segment_size
```
I would prefer
```
cuda::execution::max_segment_size
```

jrhemstad: I suspect this may be because we may want to add other segment-size related guarantees and want to have some way to group them? But otherwise I agree, this is a bit wordy.

srinivasyadav18: Or we could do something like `cuda::execution::guarantees::max_segment_size` and also move `determinism*` requirements into `requirments` sub-namespace ? The naming maybe still longer, but it is more precise in its meaning IMO.

bernhardmgruber: I think as long as the new guarantees carry `segment_size` in their name, we should be fine.

I would also like to point you to the [guidance](http://wg21.link/P0816) of Titus Winters, a previous LEWG chair, arguing against the use of nested namespaces in library design.

libcudacxx/include/cuda/__execution/max_segment_size.h
```diff
@@ -0,0 +1,85 @@
...
+struct __get_max_segment_size_t;
+
+// TODO : add dynamic extent specialization later, when stateful guarantees (env) are supported
+
+//! A class template that can be used to specify the maximum segment size
+//! for segmented algorithms.
+//! \tparam _Size The maximum segment size.
+template <size_t _Size>
+struct max_segment_size : __guarantee
```
bernhardmgruber: Q: Has it been discussed whether we should also specify a minimum segment size? Does that make any sense? Like, whether we should have `segment_bounds<int Min, int Max>`, where each value can be a compile-time value, or `dyn` (and a nested data member).

libcudacxx/include/cuda/__execution/max_segment_size.h
```diff
@@ -0,0 +1,85 @@
...
+_CCCL_BEGIN_NAMESPACE_CUDA_EXECUTION
+
+namespace segment_size
+{
+struct __get_max_segment_size_t;
+
+// TODO : add dynamic extent specialization later, when stateful guarantees (env) are supported
+
+//! A class template that can be used to specify the maximum segment size
+//! for segmented algorithms.
```
bernhardmgruber, line 38: Remark: I assume that if you specify a max segment size, then calling a CUB algorithm with segmentes larger than this results in undefined behavior? Can we state this universially? If yes, then we should maybe document it here. We can also document it at the CUB algorithms accepting a max segment size though.

libcudacxx/test/libcudacxx/cuda/execution/guarantee.pass.cpp
```diff
@@ -0,0 +1,26 @@
...
+
+#include <cuda/__execution/max_segment_size.h>
+
+__host__ __device__ void test()
+{
+  static_assert(
+    cuda::std::is_same_v<decltype(cuda::execution::segment_size::__get_max_segment_size(cuda::execution::__get_guarantees(
+                           cuda::execution::guarantee(cuda::execution::segment_size::max_segment_size<42>{})))),
+                         cuda::execution::segment_size::max_segment_size<42>>);
+}
```
bernhardmgruber: Remark: I am not a fan of having a nested `guarantees` environment in an environment. I think it unnecessarily complicates usage. I made the same comment for the design of tunings. IMO, we should have a flat environment with properties.

Q: Is there a design document for the naming and nestings of CUB environments? I would like to read up on the design rational.

libcudacxx/include/cuda/__execution/max_segment_size.h
srinivasyadav18: Yes, there are use cases where `min_segment_size` is needed, and we planned to add it in future. But, `segment_bounds` seems like great idea and may be a perfect fit.

libcudacxx/include/cuda/__execution/max_segment_size.h
srinivasyadav18: > I think as long as the new guarantees carry segment_size in their name, we should be fine.

Makes sense. 
Thanks for the suggestion and the reference :) .

libcudacxx/include/cuda/__execution/max_segment_size.h
srinivasyadav18: `segment_bounds` seems to be great, but often users might have to provide information only about min or max and not both. With `segment_bounds<int Min, int Max>` it would be hard to design a simpler interface where we can default min or max in the template arguments, because currently we default the template arguments to coordinate between compile-time and run-time values.

libcudacxx/test/libcudacxx/cuda/execution/guarantee.pass.cpp
srinivasyadav18: I would prefer flat environment too for the simplicity, but I am totally uncertain if we can do that. 

> Q: Is there a design document for the naming and nestings of CUB environments? I would like to read up on the design rational.

I am not sure of that.

> but I am totally uncertain if we can do that. 

 CC @gevtushenko, can you please help here ?

bernhardmgruber: I talked to @gevtushenko about this recently and he said it was decided to have nested environments in an Alg & Perf WG meeting. So no issue on my side here (I still don't like the nesting though).

srinivasyadav18: Got it. Thanks!

libcudacxx/include/cuda/__execution/max_segment_size.h
srinivasyadav18: > Can we state this universially? 

I do not think we can state this universally. It is true when `max_segment_size` is used with guarnatee's API, and user's **guarantee** the max segment size. In that case any segment size greater than the `max_segment_size` will make the algorithm U.B by breaking the guarantee itself, provided by the user. 

This is more a helper class template that can be used with guarantee's, and maybe adapted to be used in requirements too in future.

libcudacxx/include/cuda/__execution/max_segment_size.h
elstehle: When we'll add runtime constraints, we want to consider that users may want to express **_both_**, compile time and runtime constraints: E.g., at compile time I know that my segment sizes will be below 5k items (so we can compile one-worker-per-segment kernels only). Then, at run time, I may have further information on the segment size, e.g., that it is at most 512 items, which could allow us to avoid having to filter segment sizes going to `BlockReduce` versus `WarpReduce` on an per-algorithm-invocation basis versus on a per-segment basis.

This is what I currently designed `DeviceSegmentedTopK` for.

bernhardmgruber (approved):

elstehle (approved):

gevtushenko:
The max segment size would likely go through the parameter framework that @elstehle is working on. We probably still need `guarantee`, but for a different use case. We'd like to specify total number of items for segmented problem. Given that this value doesn't describe any parameter and is optional, it feels like part of env.

elstehle:
Fyi, I have opened https://github.com/NVIDIA/cccl/pull/9278 to add a `max_total_num_items` guarantee. If you could take a look that would greatly be appreciated.

PR #6684, opened 2025-11-19
Improve and apply `_CCCL_THROW` 

fbusato:
## Description

The CCCL host API should use the exception mechanism to report errors. 

Changes introduced by this PR:

- Adds `_CCCL_THROW_IF` to throw an exception when preconditions are not meet. 
- Applies `_CCCL_THROW_IF` to `make_tma_descriptor` to replace `_CCCL_VERIFY`.


libcudacxx/include/cuda/__tma/make_tma_descriptor.h
```diff
@@ -157,8 +157,8 @@ __to_cutensor_map(tma_interleave_layout __interleave_layout) noexcept
   }
 }
 
-[[nodiscard]] _CCCL_HOST_API inline ::cuda::std::size_t
-__to_cutensor_map_size(::cuda::std::size_t __num_items, ::CUtensorMapDataType __data_type) noexcept
+[[nodiscard]] _CCCL_HOST_API inline ::cuda::std::size_t __to_cutensor_map_size(
+  ::cuda::std::size_t __num_items, ::CUtensorMapDataType __data_type) noexcept(!_CCCL_HAS_EXCEPTIONS())
```
davebayer: Let's just keep this function non noexcept

libcudacxx/include/cuda/std/__exception/exception_macros.h
```diff
@@ -84,10 +99,13 @@ _CCCL_END_NAMESPACE_CUDA_STD
     else                          \
     {                             \
     }
-#  define _CCCL_THROW(...) ::cuda::std::terminate()
-#  define _CCCL_RETHROW    ::cuda::std::terminate()
+#  define _CCCL_THROW(...)                             ::cuda::std::terminate()
+#  define _CCCL_RETHROW                                ::cuda::std::terminate()
+#  define _CCCL_THROW_IF(_CONDITION, _EXCEPTION, _MSG) _CCCL_VERIFY(!(_CONDITION), _MSG)
```
davebayer: No, this is not consistent. We should either change the `_CCCL_THROW` definition to `_CCCL_VERIFY`, too, or make this just `terminate()`, I don't want 2 macros with the same name doing different things.

What we might want to do is change the `_CCCL_THROW` definition to:
```cpp
#  define _CCCL_THROW(_EXCEPTION) _CCCL_VERIFY(false, _EXCEPTION.what())
```
but I'd like to discuss this with @miscco first.

Maybe we can introduce a public macro that would control the behaviour.

libcudacxx/include/cuda/std/__exception/exception_macros.h
```diff
@@ -84,10 +99,13 @@ _CCCL_END_NAMESPACE_CUDA_STD
...
 #endif // ^^^ no exceptions ^^^
 
+#define _CCCL_THROW_INVALID_ARG_IF(_CONDITION, _MSG) _CCCL_THROW_IF(_CONDITION, ::std::invalid_argument, _MSG)
```
davebayer: I don't think this macro is an improvement over `_CCCL_THROW_IF(cond, ::std::invalid_argument, "message")`

libcudacxx/include/cuda/std/__exception/throw_error.h
```diff
@@ -31,6 +35,26 @@
...
+                            const char* __condition,
+                            const char* __msg,
+                            ::cuda::std::source_location __loc = ::cuda::std::source_location::current()) noexcept
+{
+  _CCCL_ASSERT(__condition != nullptr, "Condition is null");
+  _CCCL_ASSERT(__msg != nullptr, "Message is null");
+  _CCCL_ASSERT(
+    ::snprintf(__msg_buffer.__buffer, 512, "%s:%d %s: %s", __loc.file_name(), __loc.line(), __condition, __msg) > 0,
+    "Failed to format error");
+  _CCCL_ASSERT(__msg_buffer.__buffer[511] == '\0', "Error message is not null-terminated");
```
davebayer: What's the point of asserting inside a error message formatter function? The function should try its best to format the error and find workarounds for invalid arguments.

If you want to make sure the function behaves correctly, please, write a test.

docs/libcudacxx/extended_api/tma/make_tma_descriptor.rst
```diff
@@ -87,6 +87,8 @@ Return value
 Preconditions
 -------------
 
+The functions throw ``std::invalid_argument`` or ``std::runtime_error`` if the preconditions are not met. The exceptions can be disabled by defining ``CCCL_DISABLE_EXCEPTIONS`` before including the header. If exceptions are disabled, the functions will terminate the program instead.
```
davebayer: Do we want to comment everywhere how the exceptions can be disabled? Personally, I would just keep the description of which exceptions can be thrown from the API.

davebayer:
Also, docs update is missing

libcudacxx/include/cuda/__tma/make_tma_descriptor.h
fbusato: this was my feeling too :)

libcudacxx/include/cuda/std/__exception/exception_macros.h
fbusato: slightly shorter name

libcudacxx/include/cuda/std/__exception/throw_error.h
fbusato: I don't see problems on doing that. Extra checks are always fine.

docs/libcudacxx/extended_api/tma/make_tma_descriptor.rst
fbusato: or we can point out a common reference in the documentation.

davebayer: Yeah, why not :)

libcudacxx/include/cuda/std/__exception/exception_macros.h
fbusato: I'm in favor of exposing assertions macros. Users should be able to replace them if they want, similarly to [Boost.Assert](https://www.boost.org/doc/libs/latest/libs/assert/doc/html/assert.html)

libcudacxx/include/cuda/std/__exception/throw_error.h
```diff
@@ -31,6 +35,26 @@
...
+{
+static char* __format_error(::cuda::__detail::__msg_storage& __msg_buffer,
+                            const char* __condition,
+                            const char* __msg,
+                            ::cuda::std::source_location __loc = ::cuda::std::source_location::current()) noexcept
+{
+  _CCCL_ASSERT(__condition != nullptr, "Condition is null");
+  _CCCL_ASSERT(__msg != nullptr, "Message is null");
+  _CCCL_ASSERT(
+    ::snprintf(__msg_buffer.__buffer, 512, "%s:%d %s: %s", __loc.file_name(), __loc.line(), __condition, __msg) > 0,
```
pciolkosz: Nit: buffer size should come from `__msg_storage` type, probably static member? But could be just a member too

libcudacxx/include/cuda/std/__exception/throw_error.h
davebayer: I don't agree, because if there is an error, you want to end up with the best possible message describing the error, not `"Message is null"` or `Condition is null`, or `"Failed to format error"`, you want even a partially formatted string.

davebayer: And we need to set the last element to `\0`, not assert that it's zero terminated.

libcudacxx/include/cuda/std/__exception/throw_error.h
```diff
@@ -22,7 +22,11 @@
 #  pragma system_header
 #endif // no system header
 
+#include <cuda/std/__exception/cuda_error.h>
 #include <cuda/std/__exception/terminate.h>
+#include <cuda/std/source_location>
+
+#include <cstdio>
```
davebayer: Will not work with nvrtc, must be guarded.

libcudacxx/include/cuda/std/__exception/exception_macros.h
davebayer: We cannot just change the behaviour like this. Why would `_CCCL_THROW` do something different than `_CCCL_THROW_IF`?

libcudacxx/include/cuda/std/__exception/exception_macros.h
```diff
@@ -68,6 +73,16 @@ _CCCL_END_NAMESPACE_CUDA_STD
...
 #  define _CCCL_RETHROW    throw
+#  define _CCCL_THROW_IF(_CONDITION, _EXCEPTION, _MSG)                                            \
+    do                                                                                            \
+    {                                                                                             \
+      if (_CONDITION)                                                                             \
+      {                                                                                           \
+        ::cuda::std::__detail::__msg_storage __msg_buffer{};                                      \
+        throw _EXCEPTION(::cuda::std::__detail::__format_error(__msg_buffer, #_CONDITION, _MSG)); \
+      }                                                                                           \
+    } while (false)
```
davebayer: This will not work for example for `cuda::cuda_error`. It takes the error code as the first argument

libcudacxx/include/cuda/std/__exception/exception_macros.h
davebayer: But there are many other exceptions, so should we add all of them? And what about `cuda::cuda_error` thrown with `cudaInvalidArgument`? We also do it in some APIs.

No, we need a generic solution.

libcudacxx/include/cuda/std/__exception/exception_macros.h
fbusato: honestly, I don't care of using `_CCCL_VERIFY` or not. The goal is to help the user to understand where the problem is. It would be even a simple print

libcudacxx/include/cuda/std/__exception/exception_macros.h
fbusato: I don't see any problem. It is an internal only macro. We can add `CUDA` in the name if it is a CUDA exception.

libcudacxx/include/cuda/std/__exception/throw_error.h
fbusato: suppose the macro is called with `nullptr` for some reason, e.g. in the cuda version of this function, `nullptr` is a valid parameter. This helps to quickly fix the error. I can slightly improve the message but the idea is the same.
If the last element is not `\0` it means that the message is too long, which is a check that makes sense. We can also set the last element to `\0` for a partial message.

libcudacxx/include/cuda/std/__exception/exception_macros.h
davebayer: I think we *should* care. `assert` calls `std::abort`, but we should call `std::terminate` in these cases

libcudacxx/include/cuda/std/__exception/throw_error.h
fbusato: that's fine. my only concern is that it will use 512B even if it is never used...

libcudacxx/include/cuda/std/__exception/exception_macros.h
davebayer: use `::fprintf(stderr, ...)` instead

libcudacxx/include/cuda/std/__exception/exception_macros.h
davebayer: I see it as a big problem. These macros should be simple and generic. If I would follow your logic, we would also have `_CCCL_THROW_INVALID_ARG`, `_CCCL_THROW_RUNTIME_ERROR`, `_CCCL_THROW_RUNTIME_ERROR_IF`, etc.

This is not readable, nor is it generic. For each exception we would have 2 macros.

libcudacxx/include/cuda/std/__exception/throw_error.h
davebayer: Then I would expect to see something like `/home/david/my_src.cpp:255 "__v > 0": <unknown_error>`. The problem I see is that if the user gets: `"Message is null"` for *any* reason, it's just wrong.

From what the docs says, the message is always zero terminated, so you can completely remove the check for `'\0'`.

libcudacxx/include/cuda/std/__exception/exception_macros.h
fbusato: I don't want to overcomplicate exception handling here. We have dedicated macros for cuda exceptions.
If `_CCCL_THROW_IF` will be used in many other places, we can think to a more generic solution

libcudacxx/include/cuda/std/__exception/throw_error.h
pciolkosz: I would still prefer `__msg_storage::size()` here instead of random 512 and 511 below, especially if we want to adjust the size later

fbusato: good point. Let me fix it

libcudacxx/include/cuda/std/__exception/exception_macros.h
davebayer: Understood. I have a PR #6527 opened for changing that. But this is good for now, I will try to improve this in a future PR, I cannot come up with a better solution now :)

libcudacxx/include/cuda/std/__exception/exception_macros.h
```diff
@@ -86,6 +104,18 @@ _CCCL_END_NAMESPACE_CUDA_STD
...
 #  define _CCCL_RETHROW    ::cuda::std::terminate()
+#  define _CCCL_THROW_IF(_CONDITION, _EXCEPTION, _MSG)                          \
+    do                                                                          \
+    {                                                                           \
+      if (_CONDITION)                                                           \
+      {                                                                         \
+        ::cuda::__detail::__msg_storage __msg_buffer{};                         \
+        ::cuda::std::__detail::__format_error(__msg_buffer, #_CONDITION, _MSG); \
+        ::fprintf(stderr, "%s\n", __msg_buffer.__buffer);                       \
+        ::fflush(stderr);                                                       \
```
davebayer: This must be inside `NV_IF_TARGET(NV_IS_HOST, (...))`. These macros must work in both host and device code, otherwise this is good

libcudacxx/include/cuda/std/__exception/throw_error.h
```diff
@@ -31,6 +37,25 @@
...
+static char* __format_error(::cuda::__detail::__msg_storage& __msg_buffer,
+                            const char* __condition,
+                            const char* __msg,
+                            ::cuda::std::source_location __loc = ::cuda::std::source_location::current()) noexcept
+{
+  _CCCL_ASSERT(
+    ::snprintf(
+      __msg_buffer.__buffer, __msg_buffer.__size, "%s:%d %s: %s", __loc.file_name(), __loc.line(), __condition, __msg)
+      > 0,
+    "Failed to format error");
```
davebayer: Actually asserting on the formatting result is the part I am the most scared of. I'd like to see something like:

```suggestion
  ::snprintf(
    __msg_buffer.__buffer, __msg_buffer.__size, "%s(%u): %s: %s", __loc.file_name(), __loc.line(), __condition, (__msg != nullptr) ? __msg : "<unknown_error>");
```

libcudacxx/include/cuda/std/__exception/exception_macros.h
```diff
@@ -86,6 +104,18 @@ _CCCL_END_NAMESPACE_CUDA_STD
...
+        ::cuda::std::terminate();                                               \
+      }                                                                         \
+    } while (false)
```
davebayer: The solution I would like the most would be something like:
```suggestion
#  define _CCCL_THROW(...) \
  do \
  { \
    NV_IF_TARGET(NV_IS_HOST, ({ \
      ::cuda::__detail::__msg_storage __msg_buffer{};                         \
      ::cuda::std::__detail::__format_error(__msg_buffer, (__VA_ARGS__).what()); \
      ::fprintf(stderr, "%s\n", __msg_buffer.__buffer);                       \
      ::fflush(stderr);                                                       \
    }) \
    ::cuda::std::terminate();                                               \
  } while (0)
#  define _CCCL_RETHROW    ::cuda::std::terminate()
#  define _CCCL_THROW_IF(_CONDITION, ...)                          \
    do                                                                          \
    {                                                                           \
      if (_CONDITION)                                                           \
      {                                                                         \
        _CCCL_THROW(__VA_ARGS__); \
      }                                                                         \
    } while (false)
```
which could be simply called as `_CCCL_THROW_IF(i < size, ::std::out_of_range{})` for example. What do you think?

davebayer: We would be paying the cost of constructing the exception object, but it basically doesn't matter, because we would terminate anyway..

fbusato: I like the idea. The more information we provide, the more likely we are to address exception-related errors.

fbusato:
Reorganized the headers and a few functions to avoid circular dependencies.

libcudacxx/include/cuda/std/__exception/exception_macros.h
```diff
@@ -84,8 +96,26 @@ _CCCL_END_NAMESPACE_CUDA_STD
...
+    } while (0)
+#  define _CCCL_RETHROW ::cuda::std::terminate()
+#  define _CCCL_THROW_IF(_CONDITION, ...) \
+    do                                    \
+    {                                     \
+      if (_CONDITION)                     \
+      {                                   \
+        _CCCL_THROW(__VA_ARGS__);         \
+      }                                   \
+    } while (false)
```
davebayer: We can just use this definition for compilation both with and without exceptions

libcudacxx/include/cuda/std/__exception/exception_macros.h
```diff
@@ -84,8 +96,26 @@ _CCCL_END_NAMESPACE_CUDA_STD
...
     {                             \
     }
-#  define _CCCL_THROW(...) ::cuda::std::terminate()
-#  define _CCCL_RETHROW    ::cuda::std::terminate()
+#  define _CCCL_THROW(...)                                                                      \
+    do                                                                                          \
+    {                                                                                           \
+      NV_IF_TARGET(NV_IS_HOST, ({                                                               \
+                     ::cuda::__detail::__msg_storage __msg_buffer{};                            \
+                     ::cuda::std::__detail::__format_error(__msg_buffer, (__VA_ARGS__).what()); \
```
davebayer: Hm, it would be probably also nice to provide the type of the exception, something like: 
```suggestion
                     ::cuda::std::__detail::__format_error(__msg_buffer, typid(decltype(__VA_ARGS__)).name(), (__VA_ARGS__).what()); \
```

libcudacxx/include/cuda/std/__exception/exception_macros.h
```diff
@@ -68,6 +71,15 @@ _CCCL_END_NAMESPACE_CUDA_STD
 #  define _CCCL_CATCH_FALLTHROUGH
 #  define _CCCL_THROW(...) throw __VA_ARGS__
 #  define _CCCL_RETHROW    throw
+#  define _CCCL_THROW_IF(_CONDITION, ...) \
```
miscco: I dont like this being implemented with elipsis. This hould be
```suggestion
#  define _CCCL_THROW_IF(_CONDITION, _ERROR) \
```

libcudacxx/include/cuda/std/__exception/msg_storage.h
```diff
@@ -0,0 +1,43 @@
...
+#  pragma system_header
+#endif // no system header
+
+#include <cuda/std/__cstddef/types.h>
+
+#include <cuda/std/__cccl/prologue.h>
+
+_CCCL_BEGIN_NAMESPACE_CUDA
+
+namespace __detail
```
miscco: That namespace does not serve any purpose, as this is an internal type

libcudacxx/include/cuda/std/__exception/throw_error.h
```diff
@@ -32,6 +38,23 @@
 
 _CCCL_BEGIN_NAMESPACE_CUDA_STD
 
+namespace __detail
+{
```
miscco: Drop the namespace

libcudacxx/include/cuda/__tma/make_tma_descriptor.h
```diff
@@ -168,21 +168,25 @@ __to_cutensor_map_size(::cuda::std::size_t __num_items, ::CUtensorMapDataType __
     case ::CU_TENSOR_MAP_DATA_TYPE_UINT16:
     case ::CU_TENSOR_MAP_DATA_TYPE_BFLOAT16:
     case ::CU_TENSOR_MAP_DATA_TYPE_FLOAT16:
-      _CCCL_VERIFY(__num_items <= __max_size_t / 2, "Number of items must be less than or equal to 2^64 / 2");
+      _CCCL_THROW_IF(__num_items > __max_size_t / 2,
```
miscco: This is missing to bring in `stdexcept` or better `<cuda/std/__exception/throw_error.h>`

libcudacxx/include/cuda/std/__exception/msg_storage.h
fbusato: we need to align the implementation with https://github.com/NVIDIA/cccl/pull/6759

fbusato: @davebayer

libcudacxx/include/cuda/std/__exception/throw_error.h
fbusato: let keep for another PR to avoid conflicts with David's PR

libcudacxx/include/cuda/std/__exception/exception_macros.h
```diff
@@ -84,10 +87,29 @@ _CCCL_END_NAMESPACE_CUDA_STD
...
     {                             \
     }
-#  define _CCCL_THROW(...) ::cuda::std::terminate()
-#  define _CCCL_RETHROW    ::cuda::std::terminate()
+#  define _CCCL_THROW(...)                                                                                        \
+    do                                                                                                            \
+    {                                                                                                             \
+      NV_IF_TARGET(NV_IS_HOST, ({                                                                                 \
+                     ::cuda::__detail::__msg_storage __msg_buffer{};                                              \
+                     ::cuda::std::__format_error(__msg_buffer, typeid(__VA_ARGS__).name(), (__VA_ARGS__).what()); \
```
fbusato: `typeid(__VA_ARGS__).name()` is not quite right. We need something like `::cuda::demangle`

davebayer: Yeah, I meant `::cuda::std::__pretty_nameof`

libcudacxx/include/cuda/__tma/make_tma_descriptor.h
```diff
@@ -194,85 +199,87 @@ __to_cutensor_map_size(::cuda::std::size_t __num_items, ::CUtensorMapDataType __
...
-  _CCCL_VERIFY(__tensor.device.device_type == ::kDLCUDA || __tensor.device.device_type == ::kDLCUDAManaged,
-               "Device type must be kDLCUDA or kDLCUDAManaged");
-  auto __current_device     = ::cuda::device_ref(__tensor.device.device_id);
-  auto __compute_capability = __current_device.attribute<::cudaDevAttrComputeCapabilityMajor>();
-  _CCCL_VERIFY(__compute_capability >= 9, "Compute capability 9.0 or higher is required");
+  _CCCL_THROW_IF(__tensor.device.device_type != ::kDLCUDA && __tensor.device.device_type != ::kDLCUDAManaged,
+                 ::std::invalid_argument{"Device type must be kDLCUDA or kDLCUDAManaged"});
+  const auto __current_device     = ::cuda::device_ref(__tensor.device.device_id);
+  const auto __compute_capability = __current_device.attribute<::cudaDevAttrComputeCapabilityMajor>();
+  _CCCL_THROW_IF(__compute_capability < 9, ::std::runtime_error{"Compute capability 9.0 or higher is required"});
```
davebayer: I don't like that the code is so dense, it tries to do many things on the same line, I think this is much easier to read:
```suggestion
_CCCL_HOST_API inline void __check_device(const ::DLTensor& __tensor, tma_swizzle __swizzle)
{
  if (__tensor.device.device_type != ::kDLCUDA && __tensor.device.device_type != ::kDLCUDAManaged)
  {
    _CCCL_THROW(::std::invalid_argument{"Device type must be kDLCUDA or kDLCUDAManaged"});
  }

  const auto __current_device     = ::cuda::device_ref(__tensor.device.device_id);
  const auto __compute_capability = __current_device.attribute<::cudaDevAttrComputeCapabilityMajor>();
  
  if (__compute_capability < 9)
  {
    _CCCL_THROW(::std::runtime_error{"Compute capability 9.0 or higher is required"});
  }
```
but ok.

libcudacxx/include/cuda/__tma/make_tma_descriptor.h
```diff
@@ -194,85 +199,87 @@ __to_cutensor_map_size(::cuda::std::size_t __num_items, ::CUtensorMapDataType __
...
+      _CCCL_THROW_IF(__tensor.dtype.bits != 32 && __tensor.dtype.bits != 64,
+                     ::std::invalid_argument{"Int data type must be 32 or 64 bits"});
+      _CCCL_THROW_IF(__oobfill != tma_oob_fill::none,
+                     ::std::invalid_argument{"tma_oob_fill::nan is only supported for floating-point data types"});
       return (__tensor.dtype.bits == 32) ? ::CU_TENSOR_MAP_DATA_TYPE_INT32 : ::CU_TENSOR_MAP_DATA_TYPE_INT64;
     case ::kDLUInt: {
-      _CCCL_VERIFY(__oobfill == tma_oob_fill::none,
-                   "tma_oob_fill::nan is only supported for floating-point data types");
+      _CCCL_THROW_IF(__oobfill != tma_oob_fill::none,
+                     ::std::invalid_argument{"tma_oob_fill::nan is only supported for floating-point data types"});
```
davebayer: The same here, I am just lost in this

davebayer (commented):
I have a mixed feelings about this, I'd like to have another opinion whether this is an improvement.

@miscco could you review this, please?

libcudacxx/include/cuda/__tma/make_tma_descriptor.h
fbusato: the scope of this PR is to only apply `_CCCL_THROW`/`_CCCL_THROW_IF`. It only requires reversing the condition of `_CCCL_VERIFY`. There is no change in logic in `make_tma_descriptor.h`.

libcudacxx/include/cuda/__tma/make_tma_descriptor.h
fbusato: `_CCCL_VERIFY` is actually an assertion. It verifies that `oobfill` is `none` in this point. `_CCCL_THROW_IF` raises an exception (`std::invalid_argument`) if the condition is not true (the reverse condition of `_CCCL_VERIFY`).

libcudacxx/include/cuda/__tma/make_tma_descriptor.h
davebayer: Yes, I understand. I just want to say that I find this code hard to read, these are like 10 lines of basically full 120 characters per line. And that I would probably prefer to have `if (cond) { _CCCL_THROW(...) }` instead of `_CCCL_THROW_IF(...)`

davebayer: I mean I won't block this PR, I just want a third opinion :)

fbusato:
@davebayer I replaced all `_CCCL_THROW_IF` with `if` + `_CCCL_THROW`. If the result is nicer, I'm also fine with removing `_CCCL_THROW_IF` entirely

libcudacxx/include/cuda/std/__exception/exception_macros.h
```diff
@@ -84,10 +88,30 @@ _CCCL_END_NAMESPACE_CUDA_STD
...
+                       __msg_buffer, ::cuda::std::__pretty_nameof<decltype(__VA_ARGS__)>(), (__VA_ARGS__).what()); \
+                     ::fprintf(stderr, "%s\n", __msg_buffer.__buffer);                                             \
+                     ::fflush(stderr);                                                                             \
+                   }));                                                                                            \
+      ::cuda::std::terminate();                                                                                    \
+    } while (0)
+#  define _CCCL_RETHROW ::cuda::std::terminate()
 #endif // ^^^ no exceptions ^^^
 
+#define _CCCL_THROW_IF(_CONDITION, _ERROR) \
```
miscco: In the PR we only ever use the `_CCCL_THROW` macro, any reason to add this?

fbusato: Based on the discussion above, I agree to remove it. Thanks for the reminder.

PR #6686, opened 2025-11-19
Investigate nvhpc crash

davebayer:


davebayer:
Reduced to https://godbolt.org/z/MTEoaf1b9

PR #6689, opened 2025-11-19
Simplify `cuda::host_launch` API

davebayer:
Previously, we had a special overload for cases when the user passed `cuda::std::reference_wrapper` as the callable without any arguments.

This PR removes this overload and handles it inside the generic implementation. In addition, also functions returning `void` without arguments are launched using `cuLaunchHostFunc` which doesn't require memory allocation.

### DIFF
--- a/libcudacxx/include/cuda/__launch/host_launch.h
+++ b/libcudacxx/include/cuda/__launch/host_launch.h
@@ -26,14 +26,24 @@
 #  include <cuda/__driver/driver_api.h>
 #  include <cuda/__stream/stream_ref.h>
 #  include <cuda/std/__functional/reference_wrapper.h>
+#  include <cuda/std/__memory/addressof.h>
+#  include <cuda/std/__tuple_dir/apply.h>
+#  include <cuda/std/__tuple_dir/tuple.h>
 #  include <cuda/std/__type_traits/decay.h>
+#  include <cuda/std/__type_traits/is_function.h>
 #  include <cuda/std/__type_traits/is_move_constructible.h>
-#  include <cuda/std/__utility/forward.h>
-#  include <cuda/std/tuple>
+#  include <cuda/std/__type_traits/is_pointer.h>
+#  include <cuda/std/__utility/move.h>
 
 #  include <cuda/std/__cccl/prologue.h>
 
 _CCCL_BEGIN_NAMESPACE_CUDA
+
+template <class _Callable>
+_CCCL_HOST_API inline void CUDA_CB __host_func_launcher(void* __callable_ptr)
+{
+  (*(_Callable*) __callable_ptr)();
+}
 
 template <class _Callable, class... _Args>
 struct __stream_callback_data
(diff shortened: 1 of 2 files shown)


libcudacxx/include/cuda/__launch/host_launch.h
```diff
@@ -73,37 +82,29 @@ _CCCL_HOST_API void host_launch(stream_ref __stream, _Callable __callable, _Args
...
-//! or moved by the time it is asynchronously called the behavior is undefined.
-//!
-//! Callable must not call any APIs from cuda, thrust or cub namespaces.
-//! It must not call into CUDA Runtime or Driver APIs. It also can't depend on another
-//! thread that might block on any asynchronous CUDA work.
-//!
-//! @param __stream Stream to launch the host function on
-//! @param __callable A reference to a host function or callable object to call in stream order
-template <class _Callable>
-_CCCL_HOST_API void host_launch(stream_ref __stream, ::cuda::std::reference_wrapper<_Callable> __callable)
```
pciolkosz, line 103: I would prefer to keep the separate overload, it's easier to document this mode. Otherwise with one overload you need to describe the set of conditions to avoid the allocation, where here you have them expressed in code

libcudacxx/include/cuda/__launch/host_launch.h
```diff
@@ -73,37 +82,29 @@ _CCCL_HOST_API void host_launch(stream_ref __stream, _Callable __callable, _Args
...
                 "All callback arguments must be move constructible");
 
-  using _CallbackData                = __stream_callback_data<_Callable, _Args...>;
-  _CallbackData* __callback_data_ptr = new _CallbackData{::cuda::std::move(__callable), {::cuda::std::move(__args)...}};
+  constexpr auto __has_args = sizeof...(_Args) > 0;
 
-  // We use the callback here to have it execute even on stream error, because it needs to free the above allocation
-  ::cuda::__driver::__streamAddCallback(__stream.get(), __stream_callback_launcher<_CallbackData>, __callback_data_ptr);
-}
+  if constexpr (!__has_args && ::cuda::std::is_function_v<_Callable> && ::cuda::std::is_pointer_v<_Callable>)
```
pciolkosz, line 90: I like this

libcudacxx/include/cuda/__launch/host_launch.h
davebayer: I don't agree, actually I think that it makes everything much easier. We would have a single function. We can simply document the behaviour without talking about memory allocations, pointing out the option to use `cuda::std::reference_wrapper` for cases when the user wants to pass a reference to a callable or an argument.

Then, we usually have *Performance Considerations* section where we would describe that if there are no parameters passed to the function and the function is either a free function or a `cuda::std::reference_wrapper` we use `cuLaunchHostFunc` without memory allocations and `cuStreamAddCallback` otherwise.

I think this makes everything much cleaner.

pciolkosz: People usually just glance the documentation and they will immediately notice two overloads, a very small subset will read the performance considerations section.
I actually think the overload you are removing is more important than the other one and should be used more often, that's why I want it to be as visible as possible.

pciolkosz:
I think we should try to merge this for 3.2. But I will insist to keep the two overloads. Especially since the `reference_wrapper` argument only makes sense with no function arguments, but single overload can't express that. The proposed implementation doesn't handle that, but even if it did, explaining it in the docs seems more complex than just having that separate overload.

davebayer:
> I think we should try to merge this for 3.2. But I will insist to keep the two overloads. Especially since the reference_wrapper argument only makes sense with no function arguments, but single overload can't express that. The proposed implementation doesn't handle that, but even if it did, explaining it in the docs seems more complex than just having that separate overload.

The implementation handles that. there is `!__has_args` inside the condition. I repeat: I don't see any sense in keeping the implementation separated, especially now when we optimize the case when there are no arguments passed to a free function. I am all for separating the overloads in the documentation.

pciolkosz:
> The implementation handles that. there is !__has_args inside the condition.

I guess it depends how you define it handles it. If you pass in a reference wrapper and arguments I think there will be some compilation error from the callable wrapper, which is not ideal. But this is also fixable with a simple static assert, so as long as we document separate overloads, I am fine with the code changes

pciolkosz (approved):

Issue #6573, opened 2025-11-10
Remove support for multiple C++ standards in the same CMake build


bernhardmgruber:
Our CMake scripts support creating targets for multiple different C++ standard versions in the same build directory. This ignores common practice of setting a single C++ standard in CMake using `CMAKE_CXX_STANDARD` (or `CMAKE_CUDA_STANDARD`), which is misleading for new comers expecting a standard CMake workflow, and incurs maintainance cost and overhead on our side, whenever we have to touch this machinery or configure builds ourselves outside presets.

Because there is no longer a need to configure for multiple C++ standards in the same build, I propose to remove this "feature" and just depend on `CMAKE_CXX_STANDARD` and `CMAKE_CUDA_STANDARD`.

As a starting point, this concerns at least the following cmake variables (to be removed):
`CUB_ENABLE_DIALECT_CPP17`
`CUB_ENABLE_DIALECT_CPP20`
`CUB_CPP_DIALECT_OPTIONS`
`THRUST_CPP_DIALECT`
`CCCL_KNOWN_CXX_DIALECTS`
`cudax_CPP_DIALECT_OPTIONS`
`CUB_ENABLE_CPP_DIALECT_IN_NAMES`

Any `xxx.compiler_interface_cpp${dialect}` can be collapsed to a single target as well.

Grep for any use of the word `DIALECT` or `17` and `20` in the CMake scripts.

PR #6693, opened 2025-11-19
Update libcudacxx C++ dialect handling.

alliepiper:
Ref #6573 

This is roughly the vision I have for updating std handling for the other projects.

1. Simplify the presets (remove per dialect presets)
2. Set the dialect via CMake options in the CI scripts
3. Enforce that CXX / CUDA dialects match for dev builds
4. Use CMAKE_CXX_STANDARD / CMAKE_CUDA_STANDARD for CCCL targets.

This PR just updates libcudacxx since it was most of the way there already. The other projects will need more surgery to remove the multi-config standard options/implementation, but this paints a picture of where I see this going.


alliepiper:
> I am not a fan of dropping existing configurations that are present in the other libraries
> we still have cub-cpp17 and thrust-cpp17 so I dont want to have a difference here with just libcudacxx and an additional cmake option

The other libraries will be updated to match this pattern incrementally. This is just step one.

alliepiper:
@miscco I updated the description with more details.

PR #6702, opened 2025-11-19
Remove special dialect handling from cudax build system.

alliepiper:
Ref: #6573.

Depends on #6705 .


ci/windows/test_cudax.ps1
```diff
@@ -2,8 +2,8 @@ Param(
     [Parameter(Mandatory = $false)]
     [Alias("std")]
     [ValidateNotNullOrEmpty()]
-    [ValidateSet(17, 20)]
-    [int]$CXX_STANDARD = 17,
+    [ValidateSet(20)]
```
bernhardmgruber, line 5: Q: Was 17 removed on purpose?

alliepiper: Yeah, cudax doesn't support C++17 on MSVC. This config is invalid.

PR #6710, opened 2025-11-20
Fix electing leader from any group in `cuda::memcpy_async`

bernhardmgruber:
We cannot rely that the passed group represents a thread block, it could be any group. So the best we can do is select the first thread. This will not lead to a uniform data path and results in worse codegen for TMA copies.

This fix is pulled out of #6608, which retains good codegen for TMA copies, but gets stuck with NVRTC and Turning GPUs and we run out of time for the 3.2 branch.

Fixes: #6601


elstehle (approved):

bernhardmgruber:
> > This fix is pulled out of #6608, which retains good codegen for TMA copies, but gets stuck with NVRTC and Turning GPUs and we run out of time for the 3.2 branch.
> 
> can we isolate these cases and generate better code in the others?

That's what #6608 should do. We get stuck in the newly added tests in the fallback if we do not have TMA and no LDGSTS. So we get stuck in the ordinary LDG STG path. I just have to debug this on an actual Turing GPU.

PR #6712, opened 2025-11-20
Segmented scan agent with multiple segments per block

### LINKED ISSUE #6783: [CUB]: Support processing multiple segments per block in agent_segment_scan
Currently, `cub::detail::segmented_scan::agent_segmented_scan` processes one segment per block.

This approach does not perform well when segments are short.

This issue is to support processing statically requested number of segments per block though using Schwarz reformulation of segmented scan as an ordinary scan over augmented type and using augmented scan operator.

Having such a support would allow `DeviceSegmentedScan` to process small size segments more efficiently.

This issue can be closed by merged implementation. The interface for enabling it will come as a follow up work after we have parameter guarantees framework.

### LINKED ISSUE #8424: Implement the new tuning API for `detail::segmented_scan::dispatch_segmented_scan`

oleksandr-pavlyk:
## Description

Closes #6783 
Closes #8424

  - extend `agent_segmented_scan` with ability to to consume multiple segments per block
     - Introduce new agent template parameter `MaxSegmentsPerBlock` which controls the size of `OffsetT` array in shared memory dedicated to storing inclusive-scan of segment sizes. The actual number of segments to process is passed dynamically to `consume_ranges` method. It can not exceed `MaxSegmentsPerBlock`, however.
     - Multiple-segments are processed by reconstructing head-flags while loading values using `multi_segment_iterator` class, which is read- and write- transform iterator.  
        - Read transformation reconstructs head-flag and forms value-flag pair, used as `AccumT` by `BlockScan`, and  
        - write-transform unpacks value-flag pair and only writes the value to the output iterator.
     - Mapping from `element_id` to `segment_id` and `relative_offset` (or rank) is encapsulated into searcher types.
     - Three searcher types are implemented:
        - When segment sizes are all the same, `segment_id = element_id / segment_size`.
        - Generically, binary search in the array of inclusive-scan of segment sizes is performed. Two implementations (types) are provided:
           - When static bound on array size of known (as it is in our case), branch-less binary search provides best performance
           - When this bound is unknown, `cuda::std::upper_bound` can be used for larger number of sizes, and linear search using dynamic for-loop for  smaller sizes
     - While computing inclusive-scan of segment sizes, we check if all sizes to be processed by block are the same, and use fixed-size searcher, otherwise use branch-less generic searcher.
     - When dynamic number of segments equals 1, simpler `consume_range` call is made, which does not reconstruct head-flags 

  - introduce `agent_warp_segmented_scan` which closely follows the logic of `agent_segmented_scan` that computes are block-granularity, but using warps to process given number of segments
    - This agent is advantageous when segment sizes are medium-sized, but not too small.
  - introduce `agent_thread_segmented_scan` which implement thread-granularity work. Each thread processes requested number of segments.
     - multi-segment iterator here is implemented sequentially, iterating within segment, then over segments.
     - Thread-granularity agent is ideal when segments have only a few elements each.
   - Dispatch layer is modified to take 2 additional parameters (with default values), the number of segments to be processed by a worker (defaults to 1), and the type of worker requested (defaults to block).

<hr>

<strike>
This PR implements `cub::detail::segmented_scan::agent_segmented_scan_multiple_segments_per_block`.
The agent processes statically known `int SegmentsPerBlock` segments per block in `consume_ranges` method, and implements a tail version as well. 
</strike>

## Checklist


cub/cub/agent/agent_segmented_scan.cuh
```diff
@@ -580,7 +580,7 @@ private:
         }
         else
         {
-          const projector_iv<AccumT> projector_op{static_cast<AccumT>(initial_value)};
+          const projector_iv<AccumT> projector_op{initial_value};
```
miscco: I believe this and the previous code is wrong. There is no requirement on the initial value being convertible to AccumT

oleksandr-pavlyk: Do you have an instance of `InitValueT` and `AccumT` types where this is the case?

oleksandr-pavlyk: Perhaps an example is `AccumT` which is a pair of integers `pair_t`, and scan operator is defined with signatures `auto scan_op(pair_t, pair_t) -> pair_t` and `auto scan_op(pair_t, int) -> pair_t` and the initial value is an integer. 

What I do not understand though is how the `ExclusiveScan` defines the first element in output sequence.

cub/test/catch2_test_device_segmented_scan_multi_segment.cu
```diff
@@ -648,7 +648,8 @@ std::tuple<std::vector<OffsetT>, std::vector<OffsetT>> make_in_out_offsets(const
   std::vector<OffsetT> offsets_with_gaps;
   offsets_with_gaps.resize(segment_count + 1);
   offsets_with_gaps[0] = OffsetT{0};
-  std::inclusive_scan(sizes_with_gaps.begin(), sizes_with_gaps.end(), offsets_with_gaps.begin() + 1);
```
miscco: why not ::cuda::std::inclusive_scan

oleksandr-pavlyk: This code works on `std::vector`. To use `cuda::std::inclusive_scan` I'd need to copy data into `c2h::host_vector`.  So I am going to use `compute_inclusive_scan_reference` instead.

miscco: I was talking about the serial algorithm that is running on host

cub/cub/agent/agent_segmented_scan.cuh
oleksandr-pavlyk: Actually, I think `DeviceScan` already makes the assumption that `InitValueT` is convertible to `AccumT`.

In `AgentScan::ScanFirstTile` method, during invocations of `blockScan.InclusiveScan(items, items, init_value, scan_op, block_aggregate);`/`blockScan.ExclusiveScan(items, items, init_value, scan_op, block_aggregate);` the `init_value` is implicitly being converted to `AccumT`.

This is because `using BlockScanT = BlockScan<AccumT, ...>` and relevant `InclusiveScan`/`ExclusiveScan` signatures are `IncluveScan(AccumT (&input)[ITEMS_PER_THREAD], AccumT (&output)[ITEMS_PER_THREAD], AccumT init_value, ...)`.

cub/test/catch2_test_device_segmented_scan_multi_segment.cu
oleksandr-pavlyk: Using 

```cpp
template <typename OffsetT>
std::tuple<std::vector<OffsetT>, std::vector<OffsetT>> 
make_in_out_offsets(const std::vector<OffsetT> sizes, OffsetT gap)
{
  std::vector<OffsetT> offsets;

  std::size_t segment_count = sizes.size();

  static constexpr OffsetT zero{0};

  offsets.resize(segment_count + 1);
  offsets[0] = zero;

  cuda::std::inclusive_scan(sizes.begin(), sizes.end(), offsets.begin() + 1);

  std::vector<OffsetT> sizes_with_gaps;
  sizes_with_gaps.resize(segment_count);
  for (std::size_t i = 0; i < segment_count; ++i)
  {
    const auto s       = sizes[i];
    sizes_with_gaps[i] = (s == 0) ? gap : s;
  }

  std::vector<OffsetT> offsets_with_gaps;
  offsets_with_gaps.resize(segment_count + 1);
  offsets_with_gaps[0] = zero;
  cuda::std::inclusive_scan(
    sizes_with_gaps.begin(),
    sizes_with_gaps.end(),
    offsets_with_gaps.begin() + 1);

  return {offsets, offsets_with_gaps};
}
```

I am getting the following errors with CTK 12.0:

<details>
<summary>Compilation errors</summary>

```
/home/coder/cccl/lib/cmake/libcudacxx/../../../libcudacxx/include/cuda/std/__numeric/inclusive_scan.h(51): error #20014-D: calling a __host__ function from a __host__ __device__ function is not allowed

Remark: The warnings can be suppressed with "-diag-suppress <warning-number>"

/home/coder/cccl/lib/cmake/libcudacxx/../../../libcudacxx/include/cuda/std/__numeric/inclusive_scan.h(52): error #20014-D: calling a __host__ function from a __host__ __device__ function is not allowed

/home/coder/cccl/lib/cmake/libcudacxx/../../../libcudacxx/include/cuda/std/__numeric/inclusive_scan.h(52): error #20014-D: calling a __host__ function from a __host__ __device__ function is not allowed

/home/coder/cccl/lib/cmake/libcudacxx/../../../libcudacxx/include/cuda/std/__numeric/inclusive_scan.h(53): error #20014-D: calling a __host__ function from a __host__ __device__ function is not allowed

4 errors detected in the compilation of "/home/coder/cccl/cub/test/catch2_test_device_segmented_scan_multi_segment.cu".
ninja: build stopped: subcommand failed.
```
</details>

At the same time, compiling the following snippet based on the above code with `nvcc inclusive_scan.cu --std=c++17 -arch=sm_86 -I libcudacxx/include/ -I cub/ -I thrust/` works fine.

<details>
<summary>C++ snippet</summary>

```cpp
// filename: inclusive_scan.cu
#include <cuda/std/numeric>

#include <iostream>
#include <tuple>
#include <vector>

template <typename OffsetT>
std::tuple<std::vector<OffsetT>, std::vector<OffsetT>> make_in_out_offsets(const std::vector<OffsetT> sizes, OffsetT gap)
{
  std::vector<OffsetT> offsets;

  std::size_t segment_count = sizes.size();

  static constexpr OffsetT zero{0};

  offsets.resize(segment_count + 1);
  offsets[0] = zero;

  // cuda::std::plus<> plus_t{};

  cuda::std::inclusive_scan(sizes.begin(), sizes.end(), offsets.begin() + 1);

  std::vector<OffsetT> sizes_with_gaps;
  sizes_with_gaps.resize(segment_count);
  for (std::size_t i = 0; i < segment_count; ++i)
  {
    const auto s       = sizes[i];
    sizes_with_gaps[i] = (s == 0) ? gap : s;
  }

  std::vector<OffsetT> offsets_with_gaps;
  offsets_with_gaps.resize(segment_count + 1);
  offsets_with_gaps[0] = zero;
  cuda::std::inclusive_scan(sizes_with_gaps.begin(), sizes_with_gaps.end(), offsets_with_gaps.begin() + 1);

  return {offsets, offsets_with_gaps};
}

int main()
{
  std::vector<unsigned int> sizes = {2, 4, 0, 5, 1, 0, 0, 1};

  auto [a, b] = make_in_out_offsets(sizes, 4u);

  for (auto el : b)
  {
    std::cout << el << " ";
  }

  std::cout << "\n";

  return 0;
}
```
</details>

cub/benchmarks/bench/segmented_scan/base.cuh
```diff
@@ -0,0 +1,300 @@
...
+  return bench_impl<0, T, OffsetT>(state, tl);
+}
+
+template <typename T, typename OffsetT>
+static void varying_segment_size_bench(nvbench::state& state, nvbench::type_list<T, OffsetT> tl)
+{
+  return bench_impl<1, T, OffsetT>(state, tl);
+}
+
+#if !(_CCCL_CUDA_COMPILER(NVCC, ==, 12, 0))
```
bernhardmgruber:
```suggestion
#if _CCCL_CUDA_COMPILER(NVCC, >=, 12, 1)
```

cub/benchmarks/bench/segmented_scan/base.cuh
```diff
@@ -0,0 +1,300 @@
...
+using benched_value_types = all_types;
+#else
+// WAR for excessive time CTK 12.0 CICC takes to compile these benchmarks for int128_t
+using benched_value_types = nvbench::type_list<int8_t, int16_t, int32_t, int64_t, float, double, complex>;
```
bernhardmgruber: Please apply the optimization described here: https://nvidia.github.io/cccl/unstable/cub/tuning.html#nvbench-attributes, even if we probably never run the benchmarks with nvcc 12.0.

Alternatively, we can just disable the benchmark entirely for nvcc 12.0 (e.g. by wrapping a big `#if/#else` around it

cub/cub/agent/agent_segmented_scan.cuh
```diff
@@ -17,13 +17,21 @@
...
 
 #include <cub/block/block_load.cuh>
+#include <cub/block/block_reduce.cuh>
 #include <cub/block/block_scan.cuh>
 #include <cub/block/block_store.cuh>
+#include <cub/detail/segmented_scan_helpers.cuh>
 #include <cub/iterator/cache_modified_input_iterator.cuh>
+#include <cub/util_arch.cuh> // MemBoundScaling
 
+#include <cuda/__cmath/ceil_div.h>
```
miscco: Question: @bernhardmgruber Do we want to go ahead with the split of agent + dispatch + kernel in the context of the new tuning API?

cub/cub/detail/segmented_scan_helpers.cuh
```diff
@@ -0,0 +1,445 @@
+// SPDX-FileCopyrightText: Copyright (c) 2025, NVIDIA CORPORATION. All rights reserved.
+// SPDX-License-Identifier: BSD-3
```
miscco:
```suggestion
// SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION. All rights reserved.
// SPDX-License-Identifier: BSD-3
```

cub/cub/detail/segmented_scan_helpers.cuh
```diff
@@ -0,0 +1,445 @@
...
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_CLANG)
+#  pragma clang system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_MSVC)
+#  pragma system_header
+#endif // no system header
+
+#include <cuda/std/__algorithm/upper_bound.h>
+#include <cuda/std/__bit/integral.h>
+#include <cuda/std/__cccl/execution_space.h>
+#include <cuda/std/__cccl/visibility.h>
```
miscco: This should be included through `config.cuh` We should never need to include `<cuda/std/__cccl/moew.h>`

cub/cub/detail/segmented_scan_helpers.cuh
```diff
@@ -0,0 +1,445 @@
...
+_CCCL_DEVICE _CCCL_FORCEINLINE constexpr augmented_value_t<ValueT, FlagT> make_value_flag(ValueT v, FlagT f) noexcept
+{
+  return {v, f};
+}
+
+template <typename BinaryOpT, typename ValueT, typename FlagT = bool>
+struct schwarz_scan_op
+{
+  using fv_t = augmented_value_t<ValueT, FlagT>;
+  BinaryOpT& scan_op;
```
miscco: Is there a need for a reference?

Do we have a requirement on internal state of the algorithm? Otherwise, this should be a value

cub/cub/detail/segmented_scan_helpers.cuh
```diff
@@ -0,0 +1,445 @@
...
+  _CCCL_HOST_DEVICE _CCCL_FORCEINLINE constexpr auto operator()(V v, F f) const
+  {
+    return make_value_flag(v, f);
+  }
+};
+
+template <typename ScanOp, typename V, typename F = bool>
+struct packer_iv
+{
+  ScanOp& op;
```
miscco: Ditto: Make value

cub/cub/detail/segmented_scan_helpers.cuh
```diff
@@ -0,0 +1,445 @@
...
+  }
+};
+
+// Given a sequence of segments, specified by cumulative sum of its sizes
+// and iterator of offsets to beginning of each segment in some allocation
+// the bag_of_segments struct maps a logical identifier of an element,
+// 0 <= elem_id < m_offsets[m_offsets.size()-1], to segment id and relative
+// offset within the segment and produces offset of the corresponding element
+// in the underlying allocation.
+template <typename SpanT, unsigned int LinearBinarySearchThreshold = 20>
```
miscco:
```suggestion
template <typename ValueT, unsigned int LinearBinarySearchThreshold = 20>
```

jrhemstad:
From offline discussion with @oleksandr-pavlyk:

The current state of this PR is that it introduces 2 axii for type of worker processing a segment (thread, warp, block) and number of segments per worker (runtime value fixed for every worker). These choices are currently manual at the non-public Dispatch API, i.e., there's no load balancing logic similar to segmented sort that attempts to bin segments based on their size and intelligently use different worker/segments per worker. 

The existing implementation before this PR only exposes block worker type and 1 segment per worker.

One way of breaking this up is to incrementally add the following:
- Adding support for >1 segments per worker (block)
- Adding support for warp as a worker
- Adding support for thread as a worker

The above will require writing the tests against the non-public Dispatch API because there is no other way to select the relevant code paths. To allow selecting these code paths via the public API, another step would be:
- Add support for new tuning API that allows picking worker & segments per worker

cub/benchmarks/bench/segmented_scan/sum.cu
```diff
@@ -0,0 +1,15 @@
+// SPDX-FileCopyrightText: Copyright (c) 2011-2023, NVIDIA CORPORATION. All rights reserved.
+// SPDX-License-Identifier: BSD-3
```
bernhardmgruber: Important: Any new file added to CCCL should have the following license:

```suggestion
// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
```
Comment applies to all new files.

cub/benchmarks/bench/segmented_scan/base.cuh
```diff
@@ -0,0 +1,300 @@
...
+      // since shared memory is allocated per warp, scale it down so that total amount of shared memory
+      // per CTA is the same as in block-level segmented scan
+      ::cuda::ceil_div(TUNE_MAX_SEGMENTS_PER_BLOCK* cub::detail::warp_threads, TUNE_THREADS)>;
+
+    using thread_segmented_scan_policy_t =
+      thread_level_agent_policy_t<TUNE_THREADS, TUNE_ITEMS, AccumT, TUNE_LOAD_MODIFIER>;
+  };
+
+  using MaxPolicy = policy_t;
+};
```
bernhardmgruber: Important: Policy hubs are phased out and should replaced by policy selectors. We translated most of CUB already, so you should fined ample examples. AI is also good at making this translation.

If you need more background, here is the design document: https://docs.google.com/document/d/1KtrSsMk2LAXSB5cdAUnmn3BMow5y6B8BhC5-XVA434g

cub/benchmarks/bench/segmented_scan/base.cuh
```diff
@@ -0,0 +1,300 @@
...
+
+  const auto elements     = static_cast<offset_t>(state.get_int64("Elements{io}"));
+  const auto segment_size = static_cast<offset_t>(state.get_int64("SegmentSize{io}"));
+  const auto num_segments = cuda::ceil_div(elements, segment_size);
+  auto& summary           = state.add_summary("NumSegments");
+  summary.set_string("name", "Number of Segments");
+  summary.set_int64("value", num_segments);
+
+  thrust::device_vector<T> input = generate(elements);
+  thrust::device_vector<T> output(elements);
```
bernhardmgruber:
```suggestion
  thrust::device_vector<T> output(elements, thrust::no_init);
```

cub/benchmarks/bench/segmented_scan/base.cuh
```diff
@@ -0,0 +1,300 @@
...
+
+  thrust::device_vector<offset_t> offsets(num_segments + 1, thrust::no_init);
+  thrust::tabulate(offsets.begin(), offsets.end(), to_offsets_functor<offset_t>{elements, segment_size, Wobble});
+
+  T* d_input          = thrust::raw_pointer_cast(input.data());
```
bernhardmgruber:
```suggestion
  const T* d_input          = thrust::raw_pointer_cast(input.data());
```

cub/cub/agent/agent_segmented_scan.cuh
bernhardmgruber: I don't have strong opinions on this. I think the split of dispatch + kernel makes somewhat sense, but I don't see a good reason to split off agents from kernels.

cub/cub/agent/agent_segmented_scan.cuh
```diff
@@ -129,25 +150,69 @@ struct agent_segmented_scan
...
-  using block_store_t = BlockStore<AccumT, block_threads, items_per_thread, AgentSegmentedScanPolicyT::store_algorithm>;
-  using block_scan_t  = BlockScan<AccumT, block_threads, AgentSegmentedScanPolicyT::scan_algorithm>;
+  using block_load_aug_t    = BlockLoad<augmented_accum_t, block_threads, items_per_thread, load_algorithm>;
+  using block_store_aug_t   = BlockStore<augmented_accum_t, block_threads, items_per_thread, store_algorithm>;
+  using block_scan_aug_t    = BlockScan<augmented_accum_t, block_threads, scan_algorithm>;
+  using block_offset_scan_t = BlockScan<OffsetT, block_threads, scan_algorithm>;
+  using block_reduce_t      = BlockReduce<unsigned int, block_threads>;
 
-  union _TempStorage
+  using _multiple_segment_algorithms_storage_t = union
```
bernhardmgruber: Q: What's the advantage of an alias to an unnamed union vs. naming the union directly?
```suggestion
  union _multiple_segment_algorithms_storage_t
```

cub/cub/agent/agent_thread_segmented_scan.cuh
```diff
@@ -0,0 +1,488 @@
...
+
+template <int Nominal4ByteBlockThreads,
+          int Nominal4BytesItemsPerThread,
+          typename ComputeT,
+          CacheLoadModifier LoadModifier,
+          typename ScalingType = detail::MemBoundScaling<Nominal4ByteBlockThreads, Nominal4BytesItemsPerThread, ComputeT>>
+struct agent_thread_segmented_scan_policy_t : ScalingType
+{
+  static constexpr CacheLoadModifier load_modifier = LoadModifier;
+};
```
bernhardmgruber: Important: Please just define a simple struct as tuning policy, as most other CUB algorithms do by now. The struct should already contain mem-bound scaled values. The scaling should be done in the default policy selector.

cub/cub/device/dispatch/tuning/tuning_segmented_scan.cuh
```diff
@@ -30,6 +32,142 @@ CUB_NAMESPACE_BEGIN
...
+  CUB_RUNTIME_FUNCTION static constexpr CacheLoadModifier LoadModifier()
+  {
+    return StaticPolicyT::block_segmented_scan_policy_t::load_modifier;
+  }
+
+  CUB_RUNTIME_FUNCTION static constexpr int WorkersPerBlock()
+  {
+    return 1;
+  }
+};
```
bernhardmgruber: Important: Since there is no CCCL.C exposure for the new segmented scan, all of those policy wrappers should not be needed. Furthermore, with the new tuning API design (policy selector-based), they should also not be needed anymore. Please refactor them away when migrating to policy selectors.

bernhardmgruber (commented):
Here are some comments from a quick look. I think the first thing to attach is to migrate to the new tuning API design, which most of CUB has by now. Since we already have a segmented scan implementation, that is not ported yet, you could port that yourself or let me port it. You can also try to integrate this PR using the old policy hub style and then migrate in a follow-up PR. What's important to me is that we finish all this work in time before CCCL 3.4 branches, which is in 3-4 weeks. Let me know what you think!

cub/benchmarks/bench/segmented_scan/base.cuh
oleksandr-pavlyk: Using `no_init` tag requires `T` to be trivially constructible, and `T=complex` is not.  So my choices are to either be paying the price of performing unnecessary initialization for all types, or create utility and place it into `nvbench_helper`, or make it available even broader as this issue often arises in tests as well

```cpp
template <typename T, typename = void>
auto make_output_device_vector(cuda::std::size_t elements) {
   return thrust::device_vector<T>(elements);
}

template <typename T, std::enable_if_t<cuda::std::is_trivially_constructible_v<T>, void>>
auto make_output_device_vector(cuda::std::size_t elements) {
   return thrust::device_vector<T>(elements, thrust::no_init);
}
```

What do you think?

bernhardmgruber: Ah, I see. That's what I built `thrust::default_init` for.

cub/test/catch2_test_device_segmented_scan_multi_segment.cu
bernhardmgruber: @miscco I think we need to disable the exec check for all our STL algorithms

oleksandr-pavlyk: Opened https://github.com/NVIDIA/cccl/issues/8469 to track the issue

davebayer: @bernhardmgruber it's even worse, I think even incrementing a host-only iterator from `__host__ __device__` function is problematic

cub/cub/agent/agent_segmented_scan.cuh
oleksandr-pavlyk: Right now, there are 3 agent implementation header files, and one kernel header file. I think the right split is to create 3 kernel header file, and move agent definition into respective kernel header files. 

If you agree to the split, I will reorganize sources

cub/cub/agent/agent_segmented_scan.cuh
```diff
@@ -17,13 +17,21 @@
 #endif // no system header
 
 #include <cub/block/block_load.cuh>
+#include <cub/block/block_reduce.cuh>
 #include <cub/block/block_scan.cuh>
 #include <cub/block/block_store.cuh>
+#include <cub/detail/segmented_scan_helpers.cuh>
 #include <cub/iterator/cache_modified_input_iterator.cuh>
+#include <cub/util_arch.cuh> // MemBoundScaling
```
oleksandr-pavlyk, line 25: MemBoundScaling is no longer used in this file. Remove the include

oleksandr-pavlyk:
This PR now uses new tuning API. 

Thank you @bernhardmgruber 🙏 for guidance with this transition.

cub/cub/agent/agent_segmented_scan.cuh
oleksandr-pavlyk: @bernhardmgruber @miscco `"cub/agent/agent_*segmented_scan.cuh"` files each define agent and agent policy. The tuning header includes agent header for the policy definition, and kernel header includes it for the agent definition. 

By moving agent definition into kernel file we reduce compilation time, but avoiding compiling agent definition twice (once compiling tuning, and once compiling kernel).

We would still need to keep dedicated files for agent policy structs though.

cub/cub/device/dispatch/dispatch_segmented_scan.cuh
```diff
@@ -66,213 +76,253 @@ struct device_segmented_scan_kernel_source
...
+  ::cuda::arch_id arch_id{};
+  if (const auto error = CubDebug(launcher_factory.PtxArchId(arch_id)))
+  {
+    return error;
+  }
 
-  /// Iterator to the input sequence of data items
-  InputIteratorT d_in;
+  return dispatch_arch(policy_selector, arch_id, [&](auto policy_getter) {
+    const segmented_scan_policy active_policy = policy_getter();
```
bernhardmgruber: Important: If you never need the tuning policy at compile-time in host code, you should just query it directly like:
```suggestion
    const segmented_scan_policy active_policy = policy_selector(arch_id);
```
Then you don't need to call `dispatch_arch` and save some compile-time.

cub/cub/agent/agent_segmented_scan.cuh
oleksandr-pavlyk: I moved definition of agents in the same translation units that defined the kernel that uses the agent.

cub/cub/device/dispatch/dispatch_segmented_scan.cuh
```diff
@@ -15,16 +15,19 @@
...
 #include <cub/device/dispatch/dispatch_common.cuh>
 #include <cub/device/dispatch/kernels/kernel_segmented_scan.cuh>
+#include <cub/device/dispatch/kernels/kernel_thread_segmented_scan.cuh>
+#include <cub/device/dispatch/kernels/kernel_warp_segmented_scan.cuh>
 #include <cub/device/dispatch/tuning/tuning_segmented_scan.cuh>
-#include <cub/thread/thread_operators.cuh>
 #include <cub/util_debug.cuh>
 #include <cub/util_device.cuh>
 #include <cub/util_math.cuh>
+#include <cub/util_temporary_storage.cuh>
```
oleksandr-pavlyk: Neither `"cub/util_temporary_storage.cuh"` nor `"cub/util_math.cuh"` are necessary in this translation unit.

cub/cub/device/dispatch/tuning/tuning_segmented_scan.cuh
```diff
@@ -13,95 +13,298 @@
...
-struct segmented_scan_policy_wrapper : PolicyT
+struct block_segmented_scan_policy
 {
-  CUB_RUNTIME_FUNCTION segmented_scan_policy_wrapper(PolicyT base)
-      : PolicyT(base)
-  {}
+  int block_threads;
+  int items_per_thread;
+  CacheLoadModifier load_modifier;
+  BlockLoadAlgorithm load_algorithm;
```
bernhardmgruber, line 42: Important: Please swap those two members for consistency with other tuning policies.

cub/cub/device/dispatch/tuning/tuning_segmented_scan.cuh
```diff
@@ -13,95 +13,298 @@
...
 {
-  CUB_RUNTIME_FUNCTION segmented_scan_policy_wrapper(StaticPolicyT base)
-      : StaticPolicyT(base)
-  {}
+  int block_threads;
+  int items_per_thread;
+  CacheLoadModifier load_modifier;
+  WarpLoadAlgorithm load_algorithm;
+  WarpStoreAlgorithm store_algorithm;
+  int max_segments_per_warp;
```
bernhardmgruber, line 92: Q: Should we rename this to just `max_segments` here and in `block_segmented_scan_policy`?

cub/cub/device/dispatch/tuning/tuning_segmented_scan.cuh
```diff
@@ -13,95 +13,298 @@
...
+        max_segments_per_block},
+      warp_segmented_scan_policy{
+        warp_scaled.block_threads,
+        warp_scaled.items_per_thread,
+        LOAD_DEFAULT,
+        WARP_LOAD_TRANSPOSE,
+        WARP_STORE_TRANSPOSE,
+        max_segments_per_warp},
+      thread_segmented_scan_policy{thread_scaled.block_threads, thread_scaled.items_per_thread, LOAD_DEFAULT}};
+  }
```
bernhardmgruber, line 307: Important: let's avoid repetition:
```suggestion
  [[nodiscard]] _CCCL_API constexpr auto operator()(::cuda::arch_id arch) const -> segmented_scan_policy
  {
    return policy_selector{int{sizeof(AccumT)}}(arch);
  }
```

oleksandr-pavlyk:
I merged main to pick up env overloads added by @gonidelis in #8610 . I resolved conflicts and updated calls to `dispatch` since they changed in this PR.

cub/cub/device/dispatch/kernels/kernel_warp_segmented_scan.cuh
```diff
@@ -0,0 +1,664 @@
...
+                  "Unexpected iterator type");
+
+    static_assert(MaxNumSegments <= max_segments,
+                  "Template parameter MaxNumSegments must not exceed size of shared array");
+
+    _CCCL_ASSERT(n_segments > 0, "Number of segments should be greater than zero");
+    _CCCL_ASSERT(n_segments <= max_segments, "Number of segments should not exceed statically provisioned storage");
+
+    // cooperatively compute inclusive scan of sizes of segments to be processed by this block
+    {
```
NaderAlAwar: Suggestion: if possible, move this block of code to a separate device function, maybe named something like `compute_segment_offsets_and_fixed_size_mask`

cub/benchmarks/bench/segmented_scan/base.cuh
NaderAlAwar: Suggestion: you probably want to add `"warp"` and `"thread"`

cub/cub/device/dispatch/kernels/kernel_thread_segmented_scan.cuh
```diff
@@ -0,0 +1,542 @@
...
+          m_output_idx = m_output_begin_idx_it[m_segment_id];
+          m_is_head    = head_flag;
+          return;
+        }
+      }
+    }
+  }
+
+public:
+  _CCCL_HOST_DEVICE _CCCL_FORCEINLINE multi_segmented_seq_iterator(
```
NaderAlAwar: Important: this should likely be `_CCCL_DEVICE` only, same for the other member functions. It also calls `set_first_nonempty`, which is device only

cub/benchmarks/bench/segmented_scan/base.cuh
oleksandr-pavlyk: Using worker "thread" for larger segment size 4417 would be quite slow, so I did not work to exercise it by default.

bernhardmgruber:
@coderabbitai full review

oleksandr-pavlyk:
@coderabbitai full review

oleksandr-pavlyk:
@coderabbitai full review

oleksandr-pavlyk:
@coderabbitai full review

oleksandr-pavlyk:
@coderabbitai full review

oleksandr-pavlyk:
@coderabbitai full review

cub/cub/device/dispatch/tuning/tuning_segmented_scan.cuh
```diff
@@ -105,33 +174,46 @@ struct policy_selector
...
+        BLOCK_SCAN_WARP_SCANS,
+        max_segments_per_block},
+      warp_segmented_scan_policy{
+        warp_scaled.threads_per_block,
+        warp_scaled.items_per_thread,
+        WARP_LOAD_TRANSPOSE,
+        LOAD_DEFAULT,
+        WARP_STORE_TRANSPOSE,
+        max_segments_per_warp},
+      thread_segmented_scan_policy{thread_scaled.threads_per_block, thread_scaled.items_per_thread, LOAD_DEFAULT}};
```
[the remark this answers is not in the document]
oleksandr-pavlyk: 🤷🏻. I do not see a problem here. Algorithm tuning will be performed before the algorithms moves out of `detail` namespace. But this is for a follow up future PR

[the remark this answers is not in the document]
oleksandr-pavlyk: I do not see any issue with this tuning. I am going to add a comment to the code to make it clear.

oleksandr-pavlyk:
@coderabbitai review

oleksandr-pavlyk:
@coderabbitai full review

oleksandr-pavlyk:
@coderabbitai full review

oleksandr-pavlyk:
@coderabbitai full review

PR #6717, opened 2025-11-20
[libcu++] Use resource test fixture members through this

pciolkosz:
NVBUG 5676106

For some reason on gcc 15 resource fixture member function `bytes` and `align` come up as undefined. Use them through `this->` to fix that issue

### DIFF
--- a/libcudacxx/test/libcudacxx/cuda/memory_resource/shared_resource.cu
+++ b/libcudacxx/test/libcudacxx/cuda/memory_resource/shared_resource.cu
@@ -87,12 +87,12 @@ TEMPLATE_TEST_CASE_METHOD(test_fixture, "shared_resource", "[container][resource
       ++expected.object_count;
       CHECK(this->counts == expected);
 
-      void* ptr = mr.allocate_sync(bytes(50), align(8));
+      void* ptr = mr.allocate_sync(this->bytes(50), this->align(8));
       CHECK(ptr == this);
       ++expected.allocate_count;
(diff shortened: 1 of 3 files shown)


libcudacxx/test/libcudacxx/cuda/memory_resource/shared_resource.cu
```diff
@@ -146,15 +146,16 @@ TEMPLATE_TEST_CASE_METHOD(test_fixture, "shared_resource", "[container][resource
...
         cuda::__uninitialized_async_buffer<int, ::cuda::mr::host_accessible> other_buffer{
           buffer.memory_resource(), stream, 1337};
         ++expected.allocate_async_count;
         CHECK(this->counts == expected);
       }
 
-      // The original resource is still alive, but the second allocation was released
-      bytes(42 * sizeof(int));
+      // The original resource is still alive, but the second allocation was
+      // released
```
davebayer, line 157: You should fix your clang-format to stop wrapping comments after 80 characters, I've seen it in other PRs, too

davebayer (approved):

libcudacxx/test/libcudacxx/cuda/memory_resource/shared_resource.cu
pciolkosz: Yeah, for some reason if I touch a file outside of a dev container it happens, but I don't know why, I assumed it would be the same clang-format that is in the repo

PR #6718, opened 2025-11-20
[libcu++] Automatically bump up the release threshold of default mempools

pciolkosz:
CUDA memory pools by default have 0 release threshold, which means they don't cache any memory. It is considered a mistake and might change in the future. We changed that default for memory pools we create, but we want users to prefer the default memory pools. We should on first access to a default memory pool check if user didn't change that default 0 and if not, we will bump it up to size_t max.

This PR also fixes missing rename of the `__memory_resource_base` to `__memory_pool_base`.

### DIFF
--- a/libcudacxx/include/cuda/__memory_resource/memory_pool_base.h
+++ b/libcudacxx/include/cuda/__memory_resource/memory_pool_base.h
@@ -216,7 +216,7 @@ inline void __verify_device_supports_export_handle_type(
   }
 }
 
-inline cudaMemPool_t
+[[nodiscard]] _CCCL_HOST_API inline cudaMemPool_t
 __get_default_memory_pool(const CUmemLocation __location, [[maybe_unused]] const CUmemAllocationType __allocation_type)
 {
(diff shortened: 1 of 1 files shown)


libcudacxx/include/cuda/__memory_resource/memory_pool_base.h
```diff
@@ -211,12 +216,33 @@ inline void __verify_device_supports_export_handle_type(
   }
 }
 
+inline cudaMemPool_t
+__get_default_memory_pool(const CUmemLocation __location, [[maybe_unused]] const CUmemAllocationType __allocation_type)
```
davebayer:
```suggestion
[[nodiscard]] _CCCL_HOST_API inline cudaMemPool_t
__get_default_memory_pool(const CUmemLocation __location, [[maybe_unused]] const CUmemAllocationType __allocation_type)
```

davebayer:
pre-commit.ci autofix

davebayer (approved):

PR #6722, opened 2025-11-21
Remove special handling for dialect in Thrust's build system.

### LINKED ISSUE #6573: Remove support for multiple C++ standards in the same CMake build

alliepiper:
Closes #6573

This ended up including several rabbit holes from knock-on effects:

1. The python CCCL install relied on CMake variables that are not provided by `find_package`, and just happened to be available for _reasons_. The refactorings in this PR removed those variables and broke the install.
2. Fixed the python wheel build CMake to reuse CCCL's existing install rules instead of rolling their own with unreliable state.
3. Fixed the linux python wheel builds to setup the compilers consistently for CCCL.
4. Fixed the windows python wheel builds to always use Ninja, which always sets up compilers consistently and applies our desired build flags.
5. Added some python build artifacts to `.gitignore` files discovered after testing the python wheel build changes.
6. Discovered that some CUB headers were not being installed due to not following CUB header naming conventions. Fixed these.

### DIFF
--- a/cub/cmake/CubHeaderTesting.cmake
+++ b/cub/cmake/CubHeaderTesting.cmake
@@ -7,7 +7,17 @@
 function(cub_add_header_test label definitions)
   set(headertest_target cub.headers.${label})
 
-  cccl_generate_header_tests(${headertest_target} cub GLOBS "cub/*.cuh")
+  cccl_generate_header_tests(
+    ${headertest_target}
+    cub
+    GLOBS "cub/*.cuh"
+    # These headers have additional dependencies and strict compiler reqs.
+    # They're effectively an implementation detail of cccl.c.parallel and
+    # have their own testing.
+    EXCLUDES #
+      "cub/detail/*ptx-json*"
+      "cub/detail/ptx-json/*.cuh"
+  )
   cub_configure_cuda_target(${headertest_target} RDC ${CUB_FORCE_RDC})
   target_link_libraries(${headertest_target} PUBLIC cub.compiler_interface)
   target_compile_definitions(${headertest_target} PRIVATE ${definitions})
--- a/cub/test/ptx-json/CMakeLists.txt
+++ b/cub/test/ptx-json/CMakeLists.txt
@@ -26,15 +26,14 @@ endif()
 
 cccl_get_json()
 
-add_executable(ptx-json-test-filter filter.cpp)
-cccl_configure_target(ptx-json-test-filter DIALECT 20)
+set(filter_target "cub.test.ptx_json.filter")
+cccl_add_executable(${filter_target} SOURCES filter.cpp)
 target_link_libraries(
-  ptx-json-test-filter
+  ${filter_target}
   PRIVATE #
+    cub.compiler_interface
     nlohmann_json::nlohmann_json
     CUDA::cudart
-    Thrust::Thrust
-    CUB::CUB
 )
 
 function(cub_detail_ptx_json_add_test target_name_var source)
@@ -63,7 +62,7 @@ function(cub_detail_ptx_json_add_test target_name_var source)
     # gersemi: off
     COMMAND
       "${CMAKE_CURRENT_SOURCE_DIR}/dump_and_check.bash"
-        $<TARGET_FILE:ptx-json-test-filter>
+        $<TARGET_FILE:${filter_target}>
         $<TARGET_OBJECTS:${target_name}>
         "${CMAKE_CURRENT_SOURCE_DIR}/${source}"
         "test-json-id"
@@ -81,3 +80,21 @@ file(
 foreach (test_src IN LISTS test_srcs)
   cub_detail_ptx_json_add_test(test_target "${test_src}")
 endforeach()
+
+if (CUB_ENABLE_HEADER_TESTING)
+  set(headertest_target cub.headers.ptx_json)
+  cccl_generate_header_tests(
+    ${headertest_target}
+    cub
+    GLOBS #
+      "cub/detail/*ptx-json*"
+      "cub/detail/ptx-json/*.cuh"
+  )
+  target_link_libraries(
+    ${headertest_target}
+    PUBLIC #
+      cub.compiler_interface
+      nlohmann_json::nlohmann_json
+      CUDA::cudart
+  )
+endif()


cmake/CCCLBuildCompilerTargets.cmake
```diff
@@ -230,27 +213,4 @@ function(cccl_build_compiler_targets)
...
-      cccl.compiler_interface_cpp${dialect}
-      INTERFACE cccl.compiler_interface
-    )
-  endforeach()
-
-  # Some of our unit tests unconditionally throw exceptions, and compilers will
-  # detect that the following instructions are unreachable. This is intentional
-  # and unavoidable in these cases. This target can be used to silence
-  # unreachable code warnings.
-  add_library(cccl.silence_unreachable_code_warnings INTERFACE)
```
alliepiper, line 247: This is no longer used, drive-by cleanup.

cub/CMakeLists.txt
```diff
@@ -1,26 +1,11 @@
...
-endif()
-
-project(CUB LANGUAGES NONE)
-
-# This must appear before the installation rules, as it is required by the
-# GNUInstallDirs CMake module.
-enable_language(CXX)
-
-# Support adding CUB to a parent project via add_subdirectory.
-# See examples/cmake/add_subdir/CMakeLists.txt for details.
```
alliepiper, line 18: Made this cleanup across several projects -- these comments refer to hacks we no longer rely on, and the overall pattern can be cleaned up significantly as shown.

cub/CMakeLists.txt
```diff
@@ -40,12 +25,6 @@ if (CUB_ENABLE_HEADER_TESTING)
   include(cmake/CubHeaderTesting.cmake)
 endif()
 
-# Both testing and examples use ctest
-if (CUB_ENABLE_TESTING OR CUB_ENABLE_EXAMPLES)
-  include(CTest)
-  enable_testing()
-endif()
```
alliepiper, line 47: Made this refactoring across several projects. We always enable this in the CCCL root, so these have no effect.

libcudacxx/cmake/LibcudacxxInternalHeaderTesting.cmake
```diff
@@ -45,8 +45,8 @@ list(FILTER internal_headers EXCLUDE REGEX "__ptx/instructions/generated")
 
 function(libcudacxx_create_internal_header_test header_name headertest_src)
   # Create the default target for that file
-  set(internal_headertest_${header_name} verify_${header_name})
```
alliepiper, line 48: Drive-by clean-up -- this has no effect, the created variable is never used. Same for the other libcu++ header test files.

thrust/cmake/ThrustHeaderTesting.cmake
```diff
@@ -79,42 +75,6 @@ function(thrust_add_header_test thrust_target label definitions)
   )
   list(REMOVE_ITEM headers ${headers_exclude_details})
 
-  # List of headers that aren't implemented for all backends, but are implemented for CUDA.
```
alliepiper, line 82: This block is a leftover from `thrust::async::` test special handling. No longer needed.

ci/windows/build_cuda_cccl_python.ps1
```diff
@@ -23,12 +23,6 @@
     **Required.** The Python version to use for building the wheel, expressed
     as `<major>.<minor>` (e.g. `3.11`).
 
-.PARAMETER UseNinja
```
alliepiper, line 26: Removed the ability to use anything but ninja here. The default VS generator doesn't add all of the flags we need, and ninja is always available in our devcontainers.

This was a knock-on effect of the wheel packaging refactor that was necessary to fix how python installs CCCL headers.

ci/build_cuda_cccl_wheel.sh
```diff
@@ -32,13 +32,17 @@ echo "Using package version ${package_version}"
...
 export SETUPTOOLS_SCM_PRETEND_VERSION_FOR_CUDA_CCCL="${package_version}"
 
-
 cd /workspace/python/cuda_cccl
 
 # Determine CUDA version from nvcc
 cuda_version=$(nvcc --version | grep -oP 'release \K[0-9]+\.[0-9]+' | cut -d. -f1)
 echo "Detected CUDA version: ${cuda_version}"
 
+# Configure compilers:
```
alliepiper, line 41: Required now that python is properly installing our headers without relying on CMake variables that aren't reliable.

python/cuda_cccl/CMakeLists.txt
```diff
@@ -15,37 +15,24 @@ message(
   "Building for CUDA ${CUDA_VERSION_MAJOR}, output directory: ${CUDA_VERSION_DIR}"
 )
 
+# Build cccl.c.parallel and add CCCL's install rules
```
alliepiper, line 18: Refactored to reuse CCCL's existing install rules, and avoid using `<subproject>_SOURCE_DIR` vars, which are not provided by `find_package` and only happened to be defined here randomly. These variables are no longer available here after other refactorings in this PR, so fixing this now make sense.

cub/cub/detail/ptx-json/array.cuh
```diff
@@ -3,7 +3,7 @@
 
 #pragma once
 
-#include <cub/detail/ptx-json/object.h>
+#include <cub/detail/ptx-json/object.cuh>
```
alliepiper, line 6: Another knock-on effect of fixing python's CCCL install, it exposed that our installs were missing some CUB headers due to some headers not following conventions. CUB's headers end in `.cuh`, not `.h`

bernhardmgruber:
> Discovered that some CUB headers were not being installed due to not following CUB header naming conventions. Fixed these.

This may actually be on purpose. I think I saw that the extension of some JSON headers were changed. Those are only needed when CUB is compiled from CCCL.C, and will hopefully go away over time when we roll out the new tuning API design. I think those headers are not meant to be installed or packaged with the CTK. @wmaxey and @griwes may know more here.

alliepiper:
Gotcha. If that's the case we should exclude them from the install, but keep the header conventions.

I checked and they aren't included anywhere other than the c libraries and cub, and the cub usages are gated behind a macro, so installing them shouldn't hurt anything. Fixing the extensions did break the header tests since they're being picked up for that now, but I'm fixing that now, commit incoming!

PR #6451, opened 2025-11-03
Fixes non-default-constructible iterators for large number of items types in `DeviceRunLengthEncode::Encode`

elstehle:
## Description

Closes https://github.com/NVIDIA/cccl/issues/6419

PR #5440, opened 2025-08-06
Add support for virtual shared memory to `DispatchReduceByKey`

elstehle:
## Description

Closes https://github.com/NVIDIA/cccl/issues/5438

PR #5252, opened 2025-07-14
Adds support for large number of items to `DeviceRunLengthEncode::NonTrivialRuns`

elstehle:
## Description

Draft PR because this PR depends on https://github.com/NVIDIA/cccl/pull/5251 and https://github.com/NVIDIA/cccl/pull/5248.

This PR uses the streaming approach to add support for large number of items to `DeviceRunLengthEncode::NonTrivialRuns`. 

Closes https://github.com/NVIDIA/cccl/issues/2520  

### Benchmark results:

Comparing streaming approach (this PR) versus simply instantiating the kernel with a given offset type:

**The following table summarizes the speed-ups we get with the streaming approach:**

| Statistic            | Throughput Factor |
|----------------------|------------------|
| **Maximum speed-up** | 2.40×            |
| **Worst-case**       | 0.96×            |
| **Average speed-up** | 1.42×            |

**For 2^28 Elements Only**

| Statistic            | Throughput Factor |
|----------------------|------------------|
| **Maximum speed-up** | 2.40×            |
| **Worst-case**       | 1.00×            |
| **Average speed-up** | 1.56×            |

PR #6727, opened 2025-11-21
Backport changes for XGBoost compatibility

bdice:
## Summary
Backporting changes from PRs #5252, #5440, and #6451. These are needed to compile xgboost with CCCL 3.1.x.

## Test plan
- [x] Build RMM artifacts with these changes
- [x] Validate XGBoost compilation with CCCL 3.1.x and new RMM artifacts

### DIFF
--- a/cub/benchmarks/bench/run_length_encode/non_trivial_runs.cu
+++ b/cub/benchmarks/bench/run_length_encode/non_trivial_runs.cu
@@ -70,16 +70,19 @@ struct device_rle_policy_hub
 };
 #endif // !TUNE_BASE
 
-template <class T, class OffsetT>
-static void rle(nvbench::state& state, nvbench::type_list<T, OffsetT>)
+template <class T, class OffsetT, class RunLengthT>
+static void rle(nvbench::state& state, nvbench::type_list<T, OffsetT, RunLengthT>)
 {
-  using offset_t                   = OffsetT;
+  // Offset type large enough to represent any offset into the input sequence
+  using offset_t = cub::detail::choose_signed_offset_t<OffsetT>;
+  // Offset type large enough to represent the longest run in the sequence
+  using run_length_t = RunLengthT;
+
   using keys_input_it_t            = const T*;
   using offset_output_it_t         = offset_t*;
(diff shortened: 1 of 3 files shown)


bernhardmgruber:
> Backporting changes from PRs #5252, #5440, and #6451. These are needed to compile xgboost with CCCL 3.1.x.

That's fine. What confuses me is that I should see 3 commits that are cherry-picked from the `main` branch. Can you please check the commits you are proposing? Something is odd.

bdice:
I got these patches from @trxcllnt. @trxcllnt Can you answer @bernhardmgruber's question?

I believe the answer is that the commits on `main` don't apply cleanly, so Paul pieced this together from a partial rebase, but I'm not sure.

bdice:
Answering the above question: the only difference is that the patches I got earlier did not include a few changes in `cub/benchmarks` or `cub/tests`. I think those should be included here, though they weren't included in the original source of these patches (tests and benchmarks weren't being run there, so a minimal patch was acceptable). I am going to force-push this PR with the original commits as cherry-picks.

bdice:
This is ready for review.

I validated downstream that this fixes xgboost builds.

I am still verifying that all of RAPIDS 25.12 builds and that cuDF passes tests with these changes, but I expect that to be done shortly.

bernhardmgruber:
Thank you for changing the PR to use proper cherry-picks. May I ask why the last commit, [f72b1c7](https://github.com/NVIDIA/cccl/pull/6727/commits/f72b1c7a8e40f928e608367c3d7eda61f78ce907), is necessary? This seems to not come from a commit on `main`.

bdice:
@bernhardmgruber The cherry-picked commits are missing a number of other changes on the `main` branch that are needed to compile the tests and benchmarks that were added in the original PRs. I have added all the relevant library/header changes but CI fails without commit f72b1c7. I do not know how to adapt those tests/benchmarks to compile properly so I removed them. This PR’s total diff is exactly equivalent to what @trxcllnt has used in internal builds of XGBoost and other libraries so we believe it is correct and safe to merge.

PR #6733, opened 2025-11-22
Refactor `mdspan` `cuda::std::__detectably_invalid`

fbusato:
## Description

Replace old code in host/device/managed mdspan with https://nvidia.github.io/cccl/libcudacxx/extended_api/memory/is_pointer_accessible.html


libcudacxx/include/cuda/__mdspan/host_device_accessor.h
```diff
@@ -278,8 +267,12 @@ class __device_accessor : public _Accessor
...
   _CCCL_API static constexpr void __check_device_pointer([[maybe_unused]] __data_handle_type __p) noexcept
   {
-    NV_IF_TARGET(NV_IS_HOST,
-                 (_CCCL_ASSERT(__is_device_accessible_pointer_from_host(__p), "The pointer is not device accessible");))
+    if (!::cuda::std::__cccl_default_is_constant_evaluated())
+    {
+      NV_IF_TARGET(
+        NV_IS_HOST,
+        (_CCCL_ASSERT(__is_device_accessible_pointer_from_host(__p), "The pointer is not device accessible");))
+    }
```
davebayer:
```suggestion
    _CCCL_IF_NOT_CONSTEVAL_DEFAULT
    {
      NV_IF_TARGET(
        NV_IS_HOST,
        (_CCCL_ASSERT(__is_device_accessible_pointer_from_host(__p), "The pointer is not device accessible");))
    }
```
Applies to other places, too

libcudacxx/include/cuda/__mdspan/host_device_accessor.h
```diff
@@ -23,10 +23,12 @@
 
 #include <cuda/__driver/driver_api.h>
 #include <cuda/__memory/address_space.h>
+#include <cuda/__memory/is_pointer_accessible.h>
 #include <cuda/std/__concepts/concept_macros.h>
+#include <cuda/std/__cstddef/types.h>
 #include <cuda/std/__iterator/concepts.h>
 #include <cuda/std/__memory/pointer_traits.h>
-#include <cuda/std/__type_traits/always_false.h>
+#include <cuda/std/__type_traits/is_constant_evaluated.h>
```
davebayer:
```suggestion
```

libcudacxx/include/cuda/__mdspan/host_device_accessor.h
```diff
@@ -242,29 +234,26 @@ class __device_accessor : public _Accessor
...
       return true; // cannot be verified
     }
   }
 
 #if _CCCL_DEVICE_COMPILATION()
 
-  [[nodiscard]] _CCCL_HIDE_FROM_ABI _CCCL_DEVICE static constexpr bool
-  __is_device_accessible_pointer_from_device(__data_handle_type __p) noexcept
+  [[nodiscard]] _CCCL_HIDE_FROM_ABI
+  _CCCL_DEVICE static constexpr bool __is_device_accessible_pointer_from_device(__data_handle_type __p) noexcept
```
davebayer:
```suggestion
  [[nodiscard]] _CCCL_DEVICE_API static constexpr bool __is_device_accessible_pointer_from_device(__data_handle_type __p) noexcept
```

davebayer (changes requested):

fbusato:
@davebayer @miscco I'm starting thinking that we are doing the wrong thing for host/device/managed.
There are two problems:

1. We cannot check the pointer validity in respect to the memory space for EVERY access. This involves multiple driver calls. This is too expensive even in debug mode.
2. We cannot check the device accessibility on the host because its access is always wrong on the host side.

Ideally, `mdspan` itself should check the pointer during the creating, not during run-time. Two potential solutions:

- `cuda:std::mdpan` checks `__detectably_invalid`.
- Modify the `host/device/managed_mdspan` constructs to check the memory space.

libcudacxx/include/cuda/__mdspan/host_device_accessor.h
```diff
@@ -105,21 +104,13 @@ class __host_accessor : public _Accessor
     noexcept(::cuda::std::declval<_Accessor>().offset(::cuda::std::declval<__data_handle_type>(), 0));
 
 #if !_CCCL_COMPILER(NVRTC)
-  [[nodiscard]] _CCCL_HOST_API static constexpr bool
-  __is_host_accessible_pointer([[maybe_unused]] __data_handle_type __p) noexcept
+  [[nodiscard]]
+  _CCCL_HOST_API static constexpr bool __is_host_accessible_pointer([[maybe_unused]] __data_handle_type __p) noexcept
```
davebayer:
```suggestion
  _CCCL_HOST_API static bool __is_host_accessible_pointer([[maybe_unused]] __data_handle_type __p) noexcept
```

libcudacxx/include/cuda/__mdspan/host_device_accessor.h
```diff
@@ -242,29 +235,26 @@ class __device_accessor : public _Accessor
...
+#if _CCCL_HAS_CTK() && !_CCCL_COMPILER(NVRTC)
     if constexpr (::cuda::std::contiguous_iterator<__data_handle_type>)
     {
-      auto __p1 = ::cuda::std::to_address(__p);
-      ::CUmemorytype __type{};
-      const auto __status =
-        ::cuda::__driver::__pointerGetAttributeNoThrow<::CU_POINTER_ATTRIBUTE_MEMORY_TYPE>(__type, __p1);
-      return (__status != ::cudaSuccess) || __type == ::CU_MEMORYTYPE_DEVICE;
+      static const auto __dev_id = static_cast<int>(::cuda::__driver::__ctxGetDevice());
+      return ::cuda::is_device_accessible(::cuda::std::to_address(__p), ::cuda::device_ref{__dev_id});
```
davebayer: This is not correct, we don't want to check for the current device. We should just try to see whether it's accessible from any device here

libcudacxx/include/cuda/std/__mdspan/concepts.h
```diff
@@ -130,6 +129,33 @@ _CCCL_CONCEPT __index_pair_like = _CCCL_REQUIRES_EXPR((_Tp, _IndexType))(
...
+  _AccessorPolicy,
+  ::cuda::std::void_t<decltype(::cuda::std::declval<_AccessorPolicy>().__has_detect_invalidity(
+    ::cuda::std::declval<typename _AccessorPolicy::data_handle_type>(), ::cuda::std::declval<size_t>()))>>
+    : ::cuda::std::true_type
+{};
+
+template <typename _AccessorPolicy>
+inline constexpr bool __has_detect_invalidity = __has_detect_invalidity_s<_AccessorPolicy>::value;
+
+#endif // ^^^ _CCCL_HAS_CONCEPTS() ^^^ / vvv !_CCCL_HAS_CONCEPTS() vvv
```
davebayer: Can be simplified to:
```suggestion
template <class _AccessorPolicy>
_CCCL_CONCEPT __has_detect_invalidity = _CCCL_REQUIRES_EXPR((_AccessorPolicy), _AccessorPolicy __ap)
{
  __ap.__has_detect_invalidity(
    ::cuda::std::declval<typename _AccessorPolicy::data_handle_type>(), ::cuda::std::declval<::cuda::std::size_t>());
};
```

libcudacxx/include/cuda/std/__mdspan/mdspan.h
```diff
@@ -165,7 +165,14 @@ class mdspan
     is_nothrow_default_constructible_v<data_handle_type> && is_nothrow_default_constructible_v<mapping_type>
     && is_nothrow_default_constructible_v<accessor_type>)
       : __base()
-  {}
+  {
+    if constexpr (::cuda::std::__has_detect_invalidity<accessor_type>)
+    {
+      const auto __tmp = mapping(); // workaround for clang with nodiscard
```
davebayer: Will cause problems when assertions are disabled
```suggestion
      [[maybe_unused]] const auto __tmp = mapping(); // workaround for clang with nodiscard
```

libcudacxx/include/cuda/std/__mdspan/mdspan.h
```diff
@@ -174,52 +181,106 @@ class mdspan
...
   _CCCL_REQUIRES(__mdspan_detail::__matches_dynamic_rank<extents_type, _Size> _CCCL_AND
                    __constraints::template __is_constructible_from_index_type<_OtherIndexType>)
   _CCCL_API constexpr mdspan(data_handle_type __p, span<_OtherIndexType, _Size> __exts)
       : __base(::cuda::std::move(__p), extents_type{__exts})
-  {}
+  {
+    if constexpr (::cuda::std::__has_detect_invalidity<accessor_type>)
+    {
+      _CCCL_ASSERT(!accessor().__has_detect_invalidity(data_handle(), mapping().required_span_size()),
+                   "mdspan: invalid data handle");
```
davebayer: Here is clang fine with the code?

libcudacxx/include/cuda/std/__mdspan/mdspan.h
```diff
@@ -448,22 +519,25 @@ class mdspan
   [[nodiscard]] _CCCL_API constexpr bool is_unique() const
     noexcept(noexcept(::cuda::std::declval<const mapping_type&>().is_unique()))
   {
-    return mapping().is_unique();
+    const auto __tmp = mapping(); // workaround for clang with nodiscard
+    return __tmp.is_unique();
```
davebayer: Btw what's the issue with clang?

fbusato: clang reports that `mapping()` is discarded even if it is used by `is_unique()`

libcudacxx/include/cuda/__mdspan/host_device_accessor.h
fbusato: you suggested adding `device_ref` at the time. It makes sense, but it also makes sense to have a relaxed version without it, as I initially proposed.

libcudacxx/include/cuda/std/__mdspan/mdspan.h
fbusato: yes, looks fine

libcudacxx/include/cuda/__mdspan/host_device_accessor.h
fbusato: created https://github.com/NVIDIA/cccl/pull/6918

libcudacxx/include/cuda/__mdspan/host_device_accessor.h
```diff
@@ -202,25 +193,27 @@ class __host_accessor : public _Accessor
...
-  _CCCL_API constexpr reference access(data_handle_type __p, size_t __i) const noexcept(__is_access_noexcept)
+  _CCCL_API constexpr reference access(data_handle_type __p, ::cuda::std::size_t __i) const
+    noexcept(__is_access_noexcept)
   {
-    NV_IF_ELSE_TARGET(
-      NV_IS_DEVICE,
-      (_CCCL_VERIFY(false, "cuda::__host_accessor cannot be used in DEVICE code");),
-      (_CCCL_ASSERT(__is_host_accessible_pointer(__p), "cuda::__host_accessor data handle is not a HOST pointer");))
+    NV_IF_TARGET(NV_IS_DEVICE, (_CCCL_VERIFY(false, "cuda::__host_accessor cannot be used in DEVICE code");))
     return _Accessor::access(__p, __i);
```
miscco, line 200: Why are we dropping the check here? that seems like a regression

libcudacxx/include/cuda/std/__mdspan/mdspan.h
```diff
@@ -174,52 +181,106 @@ class mdspan
   _CCCL_REQUIRES(__constraints::template __can_construct_from_handle_and_variadic<_OtherIndexTypes...>)
   _CCCL_API explicit constexpr mdspan(data_handle_type __p, _OtherIndexTypes... __exts)
       : __base(::cuda::std::move(__p), extents_type(static_cast<index_type>(::cuda::std::move(__exts))...))
-  {}
+  {
+    if constexpr (::cuda::std::__has_detect_invalidity<accessor_type>)
+    {
+      const auto __tmp = mapping(); // workaround for clang with nodiscard
```
miscco: Above you are declaring this `[[maybe_unused]]` I believe we should do that in all places

libcudacxx/include/cuda/std/__mdspan/mdspan.h
miscco: oh yeah that bug has been one of my favorites

libcudacxx/include/cuda/__mdspan/host_device_accessor.h
fbusato: yes, this is on purpose. We don't want to call a driver API for each access of a mdspan. This is overkill

libcudacxx/include/cuda/std/__mdspan/mdspan.h
fbusato: yes, `[[maybe_unused]]` is correct because it is not used in release mode. The code is terrible for doing a trivial check, but I don't have better workaround for the clang warning

libcudacxx/include/cuda/__mdspan/host_device_accessor.h
miscco: Suggestion:  there are different ways to store that information during construction. We could either make it a boolean that is set during construction or turn the pointer into a tagged pointer which would be much more onvolved though

fbusato: I don't think we need to store any extra information. Checking the pointer validity during `mdspan` (not `accessor`) construction is enough.

fbusato:
added non-throwing overloads of pointer access functions https://github.com/NVIDIA/cccl/pull/6988

libcudacxx/include/cuda/__mdspan/host_device_accessor.h
```diff
@@ -202,26 +193,33 @@ class __host_accessor : public _Accessor
...
 
-  _CCCL_API constexpr reference access(data_handle_type __p, size_t __i) const noexcept(__is_access_noexcept)
+  _CCCL_API constexpr reference access(data_handle_type __p, ::cuda::std::size_t __i) const
+    noexcept(__is_access_noexcept)
   {
-    NV_IF_ELSE_TARGET(
-      NV_IS_DEVICE,
-      (_CCCL_VERIFY(false, "cuda::__host_accessor cannot be used in DEVICE code");),
-      (_CCCL_ASSERT(__is_host_accessible_pointer(__p), "cuda::__host_accessor data handle is not a HOST pointer");))
+    NV_IF_TARGET(NV_IS_DEVICE, (_CCCL_VERIFY(false, "cuda::__host_accessor cannot be used in DEVICE code");))
```
miscco, line 199: This puts the check into every access call, shouldnt we at least turn this into an `_CCCL_ASSERT`

libcudacxx/test/libcudacxx/cuda/containers/views/mdspan/device_mdspan/ctor.dh_extents.pass.cpp
```diff
@@ -22,6 +22,7 @@
 //   - Direct-non-list-initializes ptr_ with cuda::std::move(p),
 //   - direct-non-list-initializes map_ with ext, and
 //   - value-initializes acc_.
+#define _CCCL_DISABLE_MDSPAN_ACCESSOR_DETECT_INVALIDITY
```
miscco, line 25: Why do we need that?

fbusato: it is very hard to reimplement all tests to work with the right accessor respect to the memory space.
I added this macro to disable the checks

libcudacxx/include/cuda/__mdspan/host_device_accessor.h
fbusato: there is no check here. `_CCCL_VERIFY(false)` terminates the kernel.

PR #6739, opened 2025-11-23
libcudacxx: guard invocability traits against extended lambdas

### LINKED ISSUE #1984: [FEA]: Consider adding a static_assert to cuda::std::is_invocable that fails where its used in host code on an extended lambda
### Is this a duplicate?

- [X] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cccl/issues) for this request and that I agree to the [Code of Conduct](CODE_OF_CONDUCT.md)

### Area

libcu++

Similar to how cuda::std::invoke_result has a static_assert warning about accessing extended lambda return type from host where its not available, it would make sense to add such assert to cuda::std::is_invocable. Right now it just always returns true, no matter what argument types are passed into it

It would be nice if is_invocable would fail to compile when used in host code on an extended lambda

_No response_

### Additional context

_No response_

Aminsed:
## Summary
- reuse the existing extended-lambda detection logic from `invoke_result` to block `is_invocable{,_r}` and `is_nothrow_invocable{,_r}` when host builds introspect extended lambdas
- rely on a shared helper so the diagnostic is consistent for both `__device__` and `__host__ __device__` closures
- add NVCC-only `.fail.cpp` tests that instantiate each trait with extended lambdas to prove the static assertions fire

## Testing
- ./ci/util/build_and_test_targets.sh --preset libcudacxx-cpp20 --cmake-options "-DCMAKE_CUDA_HOST_COMPILER=/usr/bin/g++ -DCMAKE_CXX_COMPILER=/usr/bin/g++" --lit-tests "std/utilities/meta/meta.rel/is_invocable.extended_device_lambda.fail.cpp std/utilities/meta/meta.rel/is_invocable.extended_host_device_lambda.fail.cpp std/utilities/meta/meta.rel/is_nothrow_invocable.extended_device_lambda.fail.cpp"
- pre-commit run --files libcudacxx/include/cuda/std/__functional/invoke.h libcudacxx/test/libcudacxx/std/utilities/meta/meta.rel/is_invocable.extended_device_lambda.fail.cpp libcudacxx/test/libcudacxx/std/utilities/meta/meta.rel/is_invocable.extended_host_device_lambda.fail.cpp libcudacxx/test/libcudacxx/std/utilities/meta/meta.rel/is_nothrow_invocable.extended_device_lambda.fail.cpp

Fixes #1984.


libcudacxx/include/cuda/std/__functional/invoke.h
```diff
@@ -214,17 +238,35 @@ struct _CCCL_TYPE_VISIBILITY_DEFAULT invoke_result //
...
 struct _CCCL_TYPE_VISIBILITY_DEFAULT is_invocable : bool_constant<__is_invocable<_Fn, _Args...>>
-{};
+{
+#if _CCCL_CUDA_COMPILER(NVCC) && defined(__CUDACC_EXTENDED_LAMBDA__) && !_CCCL_DEVICE_COMPILATION() \
+  && (defined(__nv_is_extended_device_lambda_closure_type)                                          \
+      || defined(__nv_is_extended_host_device_lambda_closure_type))
+  static_assert(!__detail::__disallow_extended_lambda_invocability_v<_Fn>,
+                "Attempt to use an extended __device__ or __host__ __device__ lambda in a context "
+                "that requires querying its invocability in host code. Use a named function object or "
+                "cuda::proclaim_return_type instead.");
```
pciolkosz: I think we shouldn't mention `cuda::proclaim_return_type`, because it won't help here

pciolkosz:
>If we only want to invoke the lambda, there is no reason to error out. This is especially true, given that this is a nontrivial source breaking change.

I don't agree. I had a use case like this:
```
if constexpr (is_invocable<Fn, Args...>) {
   // call Fn
}
else {
   // do something else
}
```
There is no way to express that with `invoke_result` and currently if `Fn` is an extended lambda the if statement always evaluates to true regardless of what the argument types are, leading to errors. It might lead to missed errors if someone is not aware of this bizarre behavior.

The source breaking is a valid point, but in the case we would error out we return a false result, which I would say is even worse

miscco:
> The source breaking is a valid point, but in the case we would error out we return a false result, which I would say is even worse

If you care about the return type you need to query it through `is_invocable_r`  and not break other users that dont

Aminsed:
@miscco Thanks for the feedback. I've removed the static_asserts from the invocability traits and kept the diagnostic only in `invoke_result` where the return type is actually queried. Please let me know if there's anything else that needs adjustment.

libcudacxx/include/cuda/std/__functional/invoke.h
```diff
@@ -243,23 +243,23 @@ template <class _Ret, class _Fp, class... _Args>
 inline constexpr bool __nothrow_invocable_r_imp<true, true, _Ret, _Fp, _Args...> =
   noexcept(::cuda::std::__invoke(::cuda::std::declval<_Fp>(), ::cuda::std::declval<_Args>()...));
 
-template <class _Fp, class... _Args>
-inline constexpr bool is_nothrow_invocable_v =
-  __nothrow_invocable_r_imp<__is_invocable<_Fp, _Args...>, true, void, _Fp, _Args...>;
-
-template <class _Ret, class _Fp, class... _Args>
-inline constexpr bool is_nothrow_invocable_r_v =
-  __nothrow_invocable_r_imp<__is_invocable_r<_Ret, _Fp, _Args...>, is_void_v<_Ret>, _Ret, _Fp, _Args...>;
```
miscco, line 252: This is actually really bad because it now forces us to instantiate types when we do not want to do that

pciolkosz:
> If you care about the return type you need to query it through is_invocable_r and not break other users that don't

Let's step back, this has nothing to do with the return type. Currently in host code `is_invokable` will always return true for an extended lambda, regardless of what arguments types you pass into it, as described in https://github.com/NVIDIA/cccl/issues/1984

Example: https://godbolt.org/z/zs3MGexf5

I believe there was a misunderstanding of the underlying issue, leading to this PR being closed

Issue #2768, opened 2024-11-11
type trait alias templates are needlessly forcing the instantiation of class templates


ericniebler:
The `cuda::std::decay_t` type alias is currently defined as:

```c++
template <class _Tp>
using decay_t = typename decay<_Tp>::type;
```

This forces the instantiation of the `cuda::std::decay` class template, which is not necessary when the `__decay` compiler intrinsic is available.

```c++
#if defined(_CCCL_BUILTIN_DECAY)
template <class _Tp>
using decay_t = _CCCL_BUILTIN_DECAY(_Tp);
#else
template <class _Tp>
using decay_t = typename decay<_Tp>::type;
```

There are similar efficiency wins to be had for many of the other type trait alias templates.

PR #6740, opened 2025-11-23
libcudacxx: streamline simple trait aliases

Aminsed:
## Summary
- emit `add_const_t`, `add_volatile_t`, `add_cv_t`, and `type_identity_t` directly so callers no longer have to instantiate their class templates when compiler builtins can produce the result immediately
- keep the existing class templates in place for compatibility, but make the aliases return the transformed type even when the builtin is unavailable

Addresses #2768.


libcudacxx/include/cuda/std/__type_traits/type_identity.h
```diff
@@ -31,7 +31,7 @@ struct type_identity
 };
 
 template <class _Tp>
-using type_identity_t _CCCL_NODEBUG_ALIAS = typename type_identity<_Tp>::type;
+using type_identity_t _CCCL_NODEBUG_ALIAS = _Tp;
```
davebayer: I've already tried this before and some compilers are not quite happy about it unfortunately

miscco: The point here is that `type_identity_t` is a *depenent type*

This is unfortunately necessary

Aminsed:
@davebayer 
Looked into the CI failures, none seem related to this PR:
- MSVC CTK12.9: sccache S3 auth hiccup 
- Python/PyTorch: Affecting other PRs as well

davebayer:
Thanks the contribution!

PR #6742, opened 2025-11-23
Implement `cuda::sincos`

davebayer:
If once wants to compute sin and cos of a number, computing both operations at the same time allows some optimizations.

This PR introduces `cuda::sincos` that tries to use compiler builtins and device intrinsics to provide this functionality and fallbacks to separate sin and cos calculation.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -213,7 +213,7 @@ Requirements:
 * Python 3.9+
 * CUDA Toolkit 12.x
 * NVIDIA GPU (CC 6.0+)
-* Dependencies: `numba>=0.60.0`, `numpy`, `cuda-bindings>=12.9.1`, `cuda-core`, `numba-cuda>=0.18.0`
+* Dependencies: `numba>=0.60.0`, `numpy`, `cuda-bindings>=12.9.1`, `cuda-core`, `numba-cuda>=0.18.0,<0.21.2`
 
 ### Usage Examples
 
--- a/nvbench_helper/nvbench_helper/nvbench_helper.cu
+++ b/nvbench_helper/nvbench_helper/nvbench_helper.cu
@@ -398,13 +398,17 @@ void generator_t::generate(
   {
     case bit_entropy::_1_000: {
       const double* uniform_distribution = dist.new_uniform_distribution(seed, span.size());
-      thrust::for_each_n(
-        exec, thrust::make_counting_iterator(0), span.size(), set_real_t{min, max, span.data(), uniform_distribution});
+      thrust::for_each_n(exec,
+                         thrust::make_counting_iterator(std::size_t{0}),
+                         span.size(),
+                         set_real_t{min, max, span.data(), uniform_distribution});
       ++seed;
 
       uniform_distribution = dist.new_uniform_distribution(seed, span.size());
-      thrust::for_each_n(
-        exec, thrust::make_counting_iterator(0), span.size(), set_imag_t{min, max, span.data(), uniform_distribution});
+      thrust::for_each_n(exec,
+                         thrust::make_counting_iterator(std::size_t{0}),
+                         span.size(),
+                         set_imag_t{min, max, span.data(), uniform_distribution});
       ++seed;
       return;
     }
@@ -419,13 +423,17 @@ void generator_t::generate(
     }
     default: {
       const double* uniform_distribution = dist.new_uniform_distribution(seed, span.size());
-      thrust::for_each_n(
-        exec, thrust::make_counting_iterator(0), span.size(), set_real_t{min, max, span.data(), uniform_distribution});
+      thrust::for_each_n(exec,
+                         thrust::make_counting_iterator(std::size_t{0}),
+                         span.size(),
+                         set_real_t{min, max, span.data(), uniform_distribution});
       ++seed;
 
       uniform_distribution = dist.new_uniform_distribution(seed, span.size());
-      thrust::for_each_n(
-        exec, thrust::make_counting_iterator(0), span.size(), set_imag_t{min, max, span.data(), uniform_distribution});
+      thrust::for_each_n(exec,
+                         thrust::make_counting_iterator(std::size_t{0}),
+                         span.size(),
+                         set_imag_t{min, max, span.data(), uniform_distribution});
       ++seed;
 
       const int number_of_steps = static_cast<int>(entropy);
--- a/python/cuda_cccl/pyproject.toml
+++ b/python/cuda_cccl/pyproject.toml
@@ -31,7 +31,7 @@ dependencies = [
   "numpy",
   "cuda-pathfinder>=1.2.3",
   "cuda-core",
-  "numba-cuda>=0.20.0",
+  "numba-cuda>=0.20.0,<0.21.2",
   "typing_extensions",
 ]
 
@@ -42,12 +42,12 @@ readme = { file = "README.md", content-type = "text/markdown" }
 cu12 = [
   "cuda-bindings>=12.9.1,<13.0.0",
   "cuda-toolkit[nvrtc,nvjitlink,cudart,nvcc]==12.*",
-  "numba-cuda[cu12]",
+  "numba-cuda[cu12]>=0.20.0,<0.21.2",
 ]
 cu13 = [
   "cuda-bindings>=13.0.0,<14.0.0",
   "cuda-toolkit[nvrtc,nvjitlink,cudart,nvcc]==13.*",
-  "numba-cuda[cu13]",
+  "numba-cuda[cu13]>=0.20.0,<0.21.2",
 ]
 test-cu12 = [
   # an undocumented way to inherit the dependencies of the cu12 extra.


docs/libcudacxx/extended_api/math/sincos.rst
```diff
@@ -0,0 +1,69 @@
...
+
+.. code:: cuda
+
+    #include <cuda/cmath>
+    #include <cuda/std/cassert>
+    #include <cuda/std/numbers>
+
+    __global__ void sincos_kernel() {
+        auto [sin_pi, cos_pi] = cuda::sincos(cuda::std::numbers::pi_v<float>);
+        assert(sin_pi == 0.f);
```
s-oboyle, line 58: Small PSA `sin(pi) == 0.0f` shouldn't be true just because of the error in the floating point value of `pi`.
Envelope estimate would be `sin(fp_pi) ~= -8.742278... × 10^-8`.

`cos(pi) == 1.0f` is probably ok. (`cos(fp_pi) = -0.999999999999996178628768210...`)

Same for double etc.

[(Small bonus fun fact, `sin(fp_pi)` is actually a really good estimate for the floating-point error in the value of `fp_pi` vs `pi`)](https://randomascii.wordpress.com/2014/10/09/intel-underestimates-error-bounds-by-1-3-quintillion/)

davebayer: Oh yeah, you are right.. I changed the example to do `cuda::sincos(0.f)` which results in `sin == 0.f` and `cost == 1.f` precisely

libcudacxx/include/cuda/std/__complex/exponential_functions.h
```diff
@@ -21,6 +21,7 @@
 #  pragma system_header
 #endif // no system header
 
+#include <cuda/__cmath/sincos.h>
```
s-oboyle, line 24: There should also be this sin/cos mutual calculation present in `hyperbolic_functions.h`, which would give a nice immediate perf boost if you changed it also. I do have numerical updates for those functions so it would eventually get replaced, but I would estimate ~6 months before that could actually be integrated.

Up to you whether you'd like the perf update in the meanwhile.

davebayer: Thanks for the suggestion, fixed :)

davebayer:
@s-oboyle I've applied the function to even more places, could you confirm it is fine?

docs/libcudacxx/extended_api/math/sincos.rst
```diff
@@ -0,0 +1,68 @@
...
+
+.. code:: cuda
+
+    #include <cuda/cmath>
+    #include <cuda/std/cassert>
+
+    __global__ void sincos_kernel() {
+        auto [sin_pi, cos_pi] = cuda::sincos(0.f);
+        assert(sin_pi == 0.f);
+        assert(cos_pi == 1.f);
```
davebayer, line 59:
```suggestion
        auto [sin_zero, cos_zero] = cuda::sincos(0.f);
        assert(sin_zero == 0.f);
        assert(cos_zero == 1.f);
```

docs/libcudacxx/extended_api/math/sincos.rst
```diff
@@ -0,0 +1,68 @@
...
+        assert(cos_zero == 1.f);
+    }
+
+    int main() {
+        sincos_kernel<<<1, 1>>>();
+        cudaDeviceSynchronize();
+        return 0;
+    }
+
+`See it on Godbolt 🔗 <https://godbolt.org/z/99PP9s1z6>`__
```
davebayer, line 68:
```suggestion
`See it on Godbolt 🔗 <https://godbolt.org/z/5TTdPMzjz>`__
```

docs/libcudacxx/extended_api/math/sincos.rst
```diff
@@ -0,0 +1,68 @@
...
+     T cos;
+   };
+
+   template </*floating-point-type*/ T>
+   [[nodiscard]] __host__ __device__
+   sincos_result<T> sincos(T value) noexcept; // (1)
+
+   template <class Integral>
+   [[nodiscard]] __host__ __device__
+   sincos_result<double> sincos(Integral value) noexcept; // (A)
```
fbusato: using `1` and `A` looks a typo
```suggestion
   sincos_result<double> sincos(Integral value) noexcept; // (2)
```

docs/libcudacxx/extended_api/math/sincos.rst
```diff
@@ -0,0 +1,68 @@
...
+- ``cuda::sincos_result`` object with both values set to ``NaN`` if the input value is :math:`\pm\infty` or ``NaN`` and to results of :math:`\sin value` and :math:`\cos value` otherwise. (1)
+- if ``T`` is an integral type, the input value is treated as ``double``. (A)
+
+**Constraints**
+
+- ``T`` is an arithmetic type.
+
+**Performance considerations**
+
+- If available, the functionality is implemented by compiler builtins, otherwise fallbacks to ``cuda::std::sin(value)`` and ``cuda::std::cos(value)``.
```
fbusato, line 46: we can be more precise about the types where the optimization applies

libcudacxx/test/libcudacxx/cuda/cmath/sincos.pass.cpp
```diff
@@ -0,0 +1,106 @@
...
+  if constexpr (cuda::std::numeric_limits<T>::has_quiet_NaN)
+  {
+    auto pos_result = cuda::sincos(cuda::std::numeric_limits<T>::quiet_NaN());
+    assert(cuda::std::isnan(pos_result.sin));
+    assert(cuda::std::isnan(pos_result.cos));
+
+    auto neg_result = cuda::sincos(-cuda::std::numeric_limits<T>::quiet_NaN());
+    assert(cuda::std::isnan(neg_result.sin));
+    assert(cuda::std::isnan(neg_result.cos));
+  }
```
fbusato, line 97: can we add another test case with a generic values, e.g. 0.3 and test if it matches (or approx) separate calls to `cos/sin`

libcudacxx/include/cuda/__cmath/sincos.h
```diff
@@ -0,0 +1,122 @@
...
+#include <cuda/std/__concepts/concept_macros.h>
+#include <cuda/std/__type_traits/conditional.h>
+#include <cuda/std/__type_traits/is_extended_arithmetic.h>
+#include <cuda/std/__type_traits/is_integral.h>
+#include <cuda/std/__type_traits/is_same.h>
+
+#include <cuda/std/__cccl/prologue.h>
+
+#if _CCCL_HAS_BUILTIN(__builtin_sincosf) || _CCCL_COMPILER(GCC)
+#  define _CCCL_BUILTIN_SINCOSF(...) __builtin_sincosf(__VA_ARGS__)
```
fbusato, line 34: question. Would make sense to `#undef` the macros at the end of the file

libcudacxx/include/cuda/__cmath/sincos.h
```diff
@@ -0,0 +1,122 @@
...
+
+//! @brief Computes sin and cos operation of a value.
+//!
+//! @param __v The value.
+//!
+//! @return The \c cuda::sincos_result with the results of sin and cos operations.
+_CCCL_TEMPLATE(class _Tp)
+_CCCL_REQUIRES(::cuda::std::__is_extended_arithmetic_v<_Tp>)
+[[nodiscard]] _CCCL_API auto sincos(_Tp __v) noexcept
+  -> sincos_result<::cuda::std::conditional_t<::cuda::std::is_integral_v<_Tp>, double, _Tp>>
```
fbusato, line 70: I see little benefits of using trailing return type but it is fine

libcudacxx/include/cuda/__cmath/sincos.h
```diff
@@ -0,0 +1,122 @@
...
+        NV_IF_TARGET(NV_IS_DEVICE, (::sincosf(__v, &__ret.sin, &__ret.cos); return __ret;))
+      }
+      if constexpr (::cuda::std::is_same_v<_Tp, double>)
+      {
+        NV_IF_TARGET(NV_IS_DEVICE, (::sincos(__v, &__ret.sin, &__ret.cos); return __ret;))
+      }
+    }
+    __ret.sin = ::cuda::std::sin(__v);
+    __ret.cos = ::cuda::std::cos(__v);
+    return __ret;
```
fbusato: not sure if all compilers like it
```suggestion
    return {::cuda::std::sin(__v), ::cuda::std::cos(__v)};
```

fbusato (changes requested):

docs/libcudacxx/extended_api/math/sincos.rst
davebayer: It is not, it's how they define these overloads on cppreference. Will (2) be better?

libcudacxx/include/cuda/__cmath/sincos.h
davebayer: It's because clang has problems with determining the return type for integral overload. And I used the trailing return type to have the return type on one line.

libcudacxx/include/cuda/__cmath/sincos.h
davebayer: I don't think this is an improvement, the return structure is defined anyway

libcudacxx/include/cuda/__cmath/sincos.h
davebayer: I don't think so, we don't do it anywhere else. This would be a question for another PR anyway

libcudacxx/test/libcudacxx/cuda/cmath/sincos.pass.cpp
davebayer: Hahah, for sure we can, I wanted not to deal with the floating point comparisons and just test whether it works 😅

docs/libcudacxx/extended_api/math/sincos.rst
fbusato: yes, using 1. and A. is not very clear. I never noted in cppreference honestly

libcudacxx/include/cuda/__cmath/sincos.h
fbusato: it is an improvement in the sense the code is shorter 😅. Anyway, up to you

docs/libcudacxx/extended_api/math/sincos.rst
davebayer: What exactly do you have in mind? Right now, the implementation uses the buitlins for all types if available

fbusato: I would like to see the list of types, e.g. `float`, `__half`, etc. where `sincos` is better than two separate calls

davebayer: With exception of `__float128` (for which we don't have cmath support yet), the implementation always uses `sincos` variant if available. `__half` and `__nv_bfloat16` are converted to `float`, computed using `sincosf` and converted back as we do when computing any other trigonometric functions.

So the list would basically just contain all types with a note that all of them use the `sincos`. I don't see too much value in that

fbusato: ok, then. There is no value in adding practically all types

libcudacxx/test/libcudacxx/cuda/cmath/sincos.pass.cpp
```diff
@@ -0,0 +1,145 @@
...
+
+#if _LIBCUDACXX_HAS_NVFP16()
+__host__ __device__ bool is_about(__half x, __half y)
+{
+  return (cuda::std::fabs((x - y) / (x + y)) <= __half(1e-3));
+}
+#endif // _LIBCUDACXX_HAS_NVFP16()
+
+#if _LIBCUDACXX_HAS_NVBF16()
+__host__ __device__ bool is_about(__nv_bfloat16 x, __nv_bfloat16 y)
```
fbusato, line 43: I always think that one day we should provide these functions in libcu++ because they are used practically everywhere

PR #6757, opened 2025-11-24
Fix `cuda::std::reference_wrapper` noexcept test with gcc-8

davebayer:


miscco (approved):
I wonder whyy its failing now, but not enough to worry

PR #6758, opened 2025-11-24
`cuda.compute`: Fixes and improvements to function caching

shwina:
## Description

The `CachableFunction` utility helps with caching of build objects involving user-defined functions, by wrappping UDFs in a hashable object.

Thie PR:

* Fixes the behaviour for functions that capture by global reference (ensures that two otherwise identical function that capture different globals don't have the same hash)
* Enables caching for functions that call other functions
* Adds some tests for the above
* Adds some documentation explaining gotchas when using user-defined functions

## Checklist

### DIFF
--- a/docs/python/compute.rst
+++ b/docs/python/compute.rst
@@ -102,25 +102,34 @@ Here are some important gotchas to be aware of:
 * Functions capturing by global reference may not work as intended.
   Prefer using closures in these situations.
 
-  Here is an example of a function that captures a global variable by reference.
+  Here is an example of a function that captures a global variable by reference,
+  which is then used in a loop with ``unary_transform``.
 
   .. code-block:: python
 
-     i = 1
-     def func(x):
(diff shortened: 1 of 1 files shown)


NaderAlAwar (approved):

docs/python/compute.rst
```diff
@@ -74,11 +74,54 @@ The ``cuda.compute`` library supports defining custom data types,
...
+
+  To avoid this, capture the variable in a closure:
+
+  .. code-block:: python
+
+     def make_func(i):
+        def func(x):
+              return x + i  # i is captured as a closure variable
+        return func
+
```
leofang, line 133: Q: Can we say a few more words on how users should invoke `make_func` in their workloads? ex: do they pass `make_func(42)` to where a user-defined function is expected?

shwina: I tried to make it more concrete with an example.

PR #6759, opened 2025-11-24
Fix `__throw_cuda_error` availability with nvrtc

davebayer:
In #6337, I made `__throw_cuda_error` unavailable for nvrtc, but the function should be visible even for nvrtc.

The question is, why the tests didn't catch this earlier.. Are the catch2 tests compiled with nvrtc?

### DIFF
--- a/libcudacxx/include/cuda/std/__exception/cuda_error.h
+++ b/libcudacxx/include/cuda/std/__exception/cuda_error.h
@@ -1,11 +1,10 @@
-// -*- C++ -*-
 //===----------------------------------------------------------------------===//
 //
 // Part of libcu++, the C++ Standard Library for your entire system,
 // under the Apache License v2.0 with LLVM Exceptions.
 // See https://llvm.org/LICENSE.txt for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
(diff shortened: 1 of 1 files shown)


libcudacxx/include/cuda/std/__exception/cuda_error.h
```diff
@@ -22,27 +21,28 @@
...
+#if _CCCL_HAS_CTK()
 using __cuda_error_t = ::cudaError_t;
-#  else
+#else
 using __cuda_error_t = int;
-#  endif
+#endif
 
+#if !_CCCL_COMPILER(NVRTC)
 namespace __detail
```
fbusato, line 46: would be better to remove the `__detail` namespace to make it uniform with the `_CCCL_THROW_IF` PR

davebayer: Let's leave it for a separate PR where we remove all `__detail` namespaces, I'd like to keep this one simple for the backport

davebayer: Or not? I don't know :D

miscco: Fine by me

Issue #6743, opened 2025-11-24
`cuda::arch_id::sm_62` is missing


bernhardmgruber:
While porting the tunings of `DeviceRadixSort` I noticed that `sm_62` is missing from our `cuda::arch_id`s. This should be added, or the radix sort tunings for this architecture removed.

PR #6772, opened 2025-11-25
[libcu++] Add sm_62 arch traits

pciolkosz:
CUB has tunings for `sm_62`, we should also expose arch traits for it, until all sm_6X traits are removed in the future

Fixes https://github.com/NVIDIA/cccl/issues/6743

Latest programming guide doesn't have information about sm_6X, older version needs to be used for validation https://docs.nvidia.com/cuda/archive/12.9.1/cuda-c-programming-guide/index.html#features-and-technical-specifications

### DIFF
--- a/libcudacxx/include/cuda/__device/arch_id.h
+++ b/libcudacxx/include/cuda/__device/arch_id.h
@@ -37,6 +37,7 @@ enum class arch_id : int
 {
   sm_60   = 60,
   sm_61   = 61,
+  sm_62   = 62,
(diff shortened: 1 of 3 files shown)


bernhardmgruber (approved):

bernhardmgruber:
Thank you @pciolkosz!

davebayer:
We should have backported this PR

pciolkosz:
Ignore the above, I missed these changes were already back ported

PR #6774, opened 2025-11-25
Add `operator<<` to some CUB enums

bernhardmgruber:


cub/cub/block/block_radix_rank.cuh
```diff
@@ -36,6 +36,10 @@
 #include <cuda/std/limits>
 #include <cuda/std/span>
 
+#if !_CCCL_COMPILER(NVRTC)
+#  include <ostream>
+#endif
```
miscco: Nitpick:
```suggestion
#if !_CCCL_COMPILER(NVRTC)
#  include <ostream>
#endif // !_CCCL_COMPILER(NVRTC)
```

PR #6776, opened 2025-11-25
Add queries for CUB ptx version as `arch_id`

bernhardmgruber:


cub/cub/util_device.cuh
davebayer, line 307: You should be able to replace this by `cuda::device::current_arch_id()` or `cuda::device::current_compute_capability()` :)

davebayer (approved):

cub/cub/util_device.cuh
bernhardmgruber: I think we could. But all we could replace is `ptx_version = CUB_TEMP_GET_PTX;` We still have to retain the instantiation of the empty kernel to keep the host path working.

I would love to see a host version of `cuda::device::current_arch_id()` that abstracts:
```
cudaFuncAttributes empty_kernel_attrs;
result = CubDebug(cudaFuncGetAttributes(&empty_kernel_attrs, reinterpret_cast<void*>(empty_kernel)));
ptx_version = empty_kernel_attrs.ptxVersion * 10;
```
But then, we kind of have that already with `PtxVersionUncached`. I also think reporting the CUDA arch that's going to be used based on what your binary contains is maybe not what `cuda::device::current_arch_id()` would suggest, when being run on the host.

So maybe the status quo is fine.

davebayer: Yeah, I meant just the device side. I have that implemented in cuda::experimental::kernel_ref, you will have to wait for it

PR #6777, opened 2025-11-25
Implement parallel `cuda::std::reduce`

miscco:
This implements `cuda::std::reduce` utilizing a CUB backend

### DIFF
--- a/cub/test/catch2_test_device_adjacent_difference_custom_policy_hub.cu
+++ b/cub/test/catch2_test_device_adjacent_difference_custom_policy_hub.cu
@@ -1,0 +1,75 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include "insert_nested_NVTX_range_guard.h"
+
+#include <cub/device/device_adjacent_difference.cuh>
+
+#include <thrust/detail/raw_pointer_cast.h>
+
+#include <cuda/std/functional>
+#include <cuda/std/numeric>
+
+#include <c2h/catch2_test_helper.h>
+
+using namespace cub;
+
+// TODO(bgruber): drop this test with CCCL 4.0 when we drop the adjacent difference dispatcher after publishing the
+// tuning API
+
+template <typename InputIteratorT>
+struct my_policy_hub
+{
+  using ValueT = cub::detail::it_value_t<InputIteratorT>;
+
+  // from Policy500 of the CUB adjacent difference tunings
+  struct MaxPolicy : ChainedPolicy<500, MaxPolicy, MaxPolicy>
+  {
+    using AdjacentDifferencePolicy =
+      AgentAdjacentDifferencePolicy<128,
+                                    Nominal8BItemsToItems<ValueT>(7),
+                                    BLOCK_LOAD_WARP_TRANSPOSE,
+                                    LOAD_LDG,
+                                    BLOCK_STORE_WARP_TRANSPOSE>;
+  };
+};
+
+C2H_TEST("DispatchAdjacentDifference::Dispatch: custom policy hub", "[device][adjacent_difference]")
+{
+  using value_t            = int;
+  using offset_t           = unsigned;
+  using difference_op_t    = cuda::std::minus<>;
+  const offset_t num_items = 12345;
+
+  c2h::device_vector<value_t> in_items(num_items);
+  c2h::device_vector<value_t> out_items(num_items);
+  c2h::gen(C2H_SEED(1), in_items);
+
+  c2h::host_vector<value_t> host_in(in_items);
+  c2h::host_vector<value_t> expected(num_items);
+  cuda::std::adjacent_difference(host_in.begin(), host_in.end(), expected.begin(), cuda::std::minus<value_t>{});
+
+  using policy_hub_t = my_policy_hub<value_t*>;
+  using dispatch_t =
+    DispatchAdjacentDifference<value_t*, value_t*, difference_op_t, offset_t, MayAlias::No, ReadOption::Left, policy_hub_t>;
+  size_t temp_size = 0;
+  dispatch_t::Dispatch(
+    nullptr,
+    temp_size,
+    thrust::raw_pointer_cast(in_items.data()),
+    thrust::raw_pointer_cast(out_items.data()),
+    num_items,
+    difference_op_t{},
+    /* stream */ nullptr);
+  c2h::device_vector<std::uint8_t> temp_storage(temp_size, thrust::no_init);
+  dispatch_t::Dispatch(
+    thrust::raw_pointer_cast(temp_storage.data()),
+    temp_size,
+    thrust::raw_pointer_cast(in_items.data()),
+    thrust::raw_pointer_cast(out_items.data()),
+    num_items,
+    difference_op_t{},
+    /* stream */ nullptr);
+
+  REQUIRE(out_items == expected);
+}
--- a/cub/test/catch2_test_device_batch_memcpy_custom_policy_hub.cu
+++ b/cub/test/catch2_test_device_batch_memcpy_custom_policy_hub.cu
@@ -1,0 +1,115 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include "insert_nested_NVTX_range_guard.h"
+
+#include <cub/device/dispatch/dispatch_batch_memcpy.cuh>
+
+#include <thrust/detail/raw_pointer_cast.h>
+
+#include <cuda/std/array>
+#include <cuda/std/cstdint>
+
+#include <c2h/catch2_test_helper.h>
+
+using namespace cub;
+
+// TODO(bgruber): drop this test with CCCL 4.0 when we drop the batch memcpy dispatcher after publishing the tuning API
+
+template <class BufferOffsetT, class BlockOffsetT>
+struct my_policy_hub
+{
+  static constexpr uint32_t BLOCK_THREADS         = 128U;
+  static constexpr uint32_t BUFFERS_PER_THREAD    = 4U;
+  static constexpr uint32_t TLEV_BYTES_PER_THREAD = 8U;
+
+  static constexpr uint32_t LARGE_BUFFER_BLOCK_THREADS    = 256U;
+  static constexpr uint32_t LARGE_BUFFER_BYTES_PER_THREAD = 32U;
+
+  static constexpr uint32_t WARP_LEVEL_THRESHOLD  = 128;
+  static constexpr uint32_t BLOCK_LEVEL_THRESHOLD = 8 * 1024;
+
+  using buff_delay_constructor_t  = cub::detail::default_delay_constructor_t<BufferOffsetT>;
+  using block_delay_constructor_t = cub::detail::default_delay_constructor_t<BlockOffsetT>;
+
+  // from Policy500 of the CUB batch memcpy tunings
+  struct MaxPolicy : ChainedPolicy<500, MaxPolicy, MaxPolicy>
+  {
+    using AgentSmallBufferPolicyT = cub::detail::batch_memcpy::AgentBatchMemcpyPolicy<
+      BLOCK_THREADS,
+      BUFFERS_PER_THREAD,
+      TLEV_BYTES_PER_THREAD,
+      /* PREFER_POW2_BITS */ true,
+      LARGE_BUFFER_BLOCK_THREADS * LARGE_BUFFER_BYTES_PER_THREAD,
+      WARP_LEVEL_THRESHOLD,
+      BLOCK_LEVEL_THRESHOLD,
+      buff_delay_constructor_t,
+      block_delay_constructor_t>;
+
+    using AgentLargeBufferPolicyT =
+      cub::detail::batch_memcpy::agent_large_buffer_policy<LARGE_BUFFER_BLOCK_THREADS, LARGE_BUFFER_BYTES_PER_THREAD>;
+  };
+};
+
+C2H_TEST("DispatchBatchMemcpy::Dispatch: custom policy hub", "[device][memcpy]")
+{
+  using value_t         = cuda::std::uint8_t;
+  using buffer_size_t   = cuda::std::uint32_t;
+  using block_offset_t  = cuda::std::uint32_t;
+  using buffer_offset_t = cub::detail::batch_memcpy::per_invocation_buffer_offset_t;
+
+  const cuda::std::array<buffer_size_t, 5> buffer_sizes{3, 128, 512, 4096, 9000};
+
+  c2h::host_vector<c2h::device_vector<value_t>> in_buffers(buffer_sizes.size());
+  c2h::host_vector<c2h::device_vector<value_t>> out_buffers(buffer_sizes.size());
+
+  c2h::host_vector<value_t*> h_in_ptrs(buffer_sizes.size());
+  c2h::host_vector<value_t*> h_out_ptrs(buffer_sizes.size());
+  c2h::host_vector<buffer_size_t> h_sizes(buffer_sizes.size());
+
+  for (buffer_size_t i = 0; i < buffer_sizes.size(); ++i)
+  {
+    const auto bytes = buffer_sizes[i];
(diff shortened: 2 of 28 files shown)


libcudacxx/include/cuda/std/__pstl/cuda/reduce.h
```diff
@@ -0,0 +1,139 @@
...
+
+    _CCCL_TRY_CUDA_API(
+      ::cub::DeviceReduce::Reduce,
+      "__pstl_cuda_reduce: cub::DeviceReduce::Reduce failed",
+      ::cuda::std::move(__first),
+      __device_ret_ptr,
+      __count,
+      ::cuda::std::move(__func),
+      ::cuda::std::move(__init),
+      ::cuda::std::move(__policy));
```
bernhardmgruber: Important: In the `for_each_n` implementation we create a `::cuda::stream_ref __stream{cudaStreamPerThread};` and pass the stream instead of the policy here. I think we need to add `__stream` to `__policy` here.

libcudacxx/include/cuda/std/__pstl/cuda/reduce.h
```diff
@@ -0,0 +1,139 @@
...
+{
+  template <class _Policy, class _Iter, class _Tp, class _BinaryOp>
+  [[nodiscard]] _CCCL_HOST_API static _Tp
+  __par_impl(_Policy __policy, _Iter __first, _Iter __last, _Tp __init, _BinaryOp __func)
+  {
+    // Allocate memory for result
+    _Tp* __device_ret_ptr = nullptr;
+
+    _CCCL_TRY_CUDA_API(
+      ::cudaMalloc, "__pstl_cuda_reduce: allocation failed", reinterpret_cast<void**>(&__device_ret_ptr), sizeof(_Tp));
```
bernhardmgruber: Important: This must be `cudaMallocAsync`.

libcudacxx/include/cuda/std/__pstl/cuda/reduce.h
```diff
@@ -0,0 +1,139 @@
...
+
+    _CCCL_TRY_CUDA_API(
+      ::cudaMemcpy,
+      "__pstl_cuda_reduce: copy of result from device to host failed",
+      ::cuda::std::addressof(__ret),
+      __device_ret_ptr,
+      sizeof(_Tp),
+      ::cudaMemcpyDeviceToHost);
+
+    _CCCL_TRY_CUDA_API(::cudaFree, "__pstl_cuda_reduce: deallocate failed", __device_ret_ptr);
```
bernhardmgruber: Important: this should be `cudaMemcpyAsync` and `cudaFreeAsync`, followed by a sync of the stream.

libcudacxx/include/cuda/std/__pstl/cuda/reduce.h
gevtushenko: I don't think `cudaMallocAsync` is enough. Many calls below throw without ever freeing allocated memory. We need a RAII abstraction, like `async_buffer`, at least internally.

libcudacxx/include/cuda/std/__pstl/cuda/reduce.h
```diff
@@ -0,0 +1,139 @@
...
+  template <class _Policy, class _Iter, class _Tp, class _BinaryOp>
+  [[nodiscard]] _CCCL_HOST_API static _Tp
+  __par_impl(_Policy __policy, _Iter __first, _Iter __last, _Tp __init, _BinaryOp __func)
+  {
+    // Allocate memory for result
+    _Tp* __device_ret_ptr = nullptr;
+
+    _CCCL_TRY_CUDA_API(
+      ::cudaMalloc, "__pstl_cuda_reduce: allocation failed", reinterpret_cast<void**>(&__device_ret_ptr), sizeof(_Tp));
+
```
gevtushenko: important: this might lead to an issue if user type has a non-trivial constructor. Thrust hanles that by invoking a kernel only when a constructor is needed. We might do a bit better. Consider having a fancy iterator that does in-place new for non-trivial types and a raw pointer otherwise.

libcudacxx/include/cuda/std/__pstl/cuda/reduce.h
```diff
@@ -0,0 +1,139 @@
...
+struct __pstl_dispatch<__pstl_algorithm::__reduce, __execution_backend::__cuda>
+{
+  template <class _Policy, class _Iter, class _Tp, class _BinaryOp>
+  [[nodiscard]] _CCCL_HOST_API static _Tp
+  __par_impl(_Policy __policy, _Iter __first, _Iter __last, _Tp __init, _BinaryOp __func)
+  {
+    // Allocate memory for result
+    _Tp* __device_ret_ptr = nullptr;
+
+    _CCCL_TRY_CUDA_API(
```
gevtushenko: important: given that we throw below, this looks like a memory leak. Consider a RAII abstraction

libcudacxx/test/libcudacxx/std/numerics/numeric.ops/reduce/pstl_reduce.cu
```diff
@@ -0,0 +1,87 @@
...
+  };
+};
+
+C2H_TEST("cuda::std::reduce(Iter, Iter, Tp, Fn)", "[parallel algorithm]")
+{
+  thrust::device_vector<int> data(size);
+  thrust::sequence(data.begin(), data.end(), 1);
+
+  const auto policy  = cuda::execution::__cub_par_unseq;
+  decltype(auto) res = cuda::std::reduce(policy, data.begin(), data.end(), 42, plus_two{});
```
gevtushenko, line 80: suggestion: consider adding a check in the binary operator to see if it's actually invoked on GPU

libcudacxx/include/cuda/std/__pstl/cuda/reduce.h
gevtushenko: Agreed, we should pass a stream. Regarding adding stream to policy, it's a complicated subject that we are deferring until after reduction is merged.

libcudacxx/include/cuda/std/__pstl/cuda/reduce.h
bernhardmgruber: How about a `unique_ptr`?

libcudacxx/include/cuda/std/__pstl/cuda/reduce.h
bernhardmgruber: Yep, I think the iterator handling the output should do placement new.

libcudacxx/include/cuda/std/__pstl/cuda/reduce.h
miscco: Done

libcudacxx/include/cuda/std/__pstl/cuda/reduce.h
miscco: Done

libcudacxx/include/cuda/std/__pstl/cuda/reduce.h
miscco: As discussed, this will come in a later PR

libcudacxx/include/cuda/std/__pstl/cuda/reduce.h
miscco: went with a simple guard for now until we can actually provide a proper  memory resource

libcudacxx/include/cuda/std/__pstl/cuda/reduce.h
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    _CCCL_HOST_API ~__allocation_guard()
+    {
+      _CCCL_TRY_CUDA_API(::cudaFree, "__pstl_cuda_reduce: deallocate failed", __ptr_);
+    }
+
+    [[nodiscard]] _CCCL_HOST_API auto __get_result_iter()
+    {
+      return ::cuda::tabulate_output_iterator{__construct_result{__ptr_}};
```
gevtushenko: suggestion: for trivial types I'd just return `__ptr_` to let compiler optimize reduction better.

libcudacxx/include/cuda/std/__pstl/cuda/reduce.h
```diff
@@ -0,0 +1,177 @@
...
+      const auto __count = ::cuda::std::distance(__first, __last);
+      _CCCL_TRY_CUDA_API(
+        ::cub::DeviceReduce::Reduce,
+        "__pstl_cuda_reduce: cub::DeviceReduce::Reduce failed",
+        ::cuda::std::move(__first),
+        __guard.__get_result_iter(),
+        __count,
+        ::cuda::std::move(__func),
+        ::cuda::std::move(__init),
+        ::cuda::std::move(__policy));
```
gevtushenko: suggestion: this leads to two allocations / deallocations as written. We first allocate output memory, then allocate temp storage separately within the env-based CUB API. Later we deallocate in reverse order. This is rather expensive. Thrust works around this by querying temporary storage and allocating it once for temp storage + output data.

miscco: Yeah i did the two step dance

libcudacxx/include/cuda/std/__pstl/cuda/reduce.h
```diff
@@ -0,0 +1,204 @@
...
+      _CCCL_TRY_CUDA_API(::cudaMalloc,
+                         "__pstl_cuda_reduce: allocation failed",
+                         reinterpret_cast<void**>(&__ptr_),
+                         __num_bytes + sizeof(_Tp));
+    }
+
+    _CCCL_HOST_API ~__allocation_guard()
+    {
+      _CCCL_TRY_CUDA_API(::cudaFree, "__pstl_cuda_reduce: deallocate failed", __ptr_);
+    }
```
bernhardmgruber, line 96: Important: If `cudaFree` fails here, we throw an exception and immediately terminate since we are in a destructor. Should we log here or just swallow the error instead?

libcudacxx/include/cuda/std/__pstl/cuda/reduce.h
```diff
@@ -0,0 +1,204 @@
...
+      {}
+
+      template <class _Index, class _Up>
+      _CCCL_DEVICE_API void operator()(_Index, _Up&& __value)
+      {
+        ::cuda::std::__construct_at(__ptr_, ::cuda::std::forward<_Up>(__value));
+      }
+    };
+
+    void* __ptr_;
```
bernhardmgruber, line 81: Suggestion: You may want to store this as a `char*` or `cuda::std::byte*` to save a cast below.

libcudacxx/include/cuda/std/__pstl/cuda/reduce.h
```diff
@@ -0,0 +1,204 @@
...
+
+    _CCCL_HOST_API __allocation_guard(size_t __num_bytes)
+        : __ptr_(nullptr)
+    {
+      // Add the temporary storage from DeviceReduce to the allocation
+      _CCCL_TRY_CUDA_API(::cudaMalloc,
+                         "__pstl_cuda_reduce: allocation failed",
+                         reinterpret_cast<void**>(&__ptr_),
+                         __num_bytes + sizeof(_Tp));
```
bernhardmgruber:
```suggestion
      _CCCL_TRY_CUDA_API(::cudaMalloc,
                         "__pstl_cuda_reduce: allocation failed",
                         reinterpret_cast<void**>(&__ptr_),
                          sizeof(_Tp) + __num_bytes);
```
Would reflect better how we intend to use the storage

libcudacxx/include/cuda/std/__pstl/cuda/reduce.h
```diff
@@ -0,0 +1,204 @@
...
+        }
+      }
+    }
+    else
+    {
+      static_assert(__always_false_v<_Policy>,
+                    "__pstl_dispatch: CUDA backend of cuda::std::reduce requires at least random access iterators");
+      return ::cuda::std::reduce(
+        ::cuda::std::move(__first), ::cuda::std::move(__last), ::cuda::std::move(__init), ::cuda::std::move(__func));
+    }
```
bernhardmgruber, line 192: Simplify this to:
```suggestion
      static_assert(::cuda::std::__has_random_access_traversal<_Iter>,
                    "__pstl_dispatch: CUDA backend of cuda::std::reduce requires at least random access iterators");
      try
      {
        return __par_impl(
          ::cuda::std::move(__policy),
          ::cuda::std::move(__first),
          ::cuda::std::move(__last),
          ::cuda::std::move(__init),
          ::cuda::std::move(__func));
      }
      catch (const ::cuda::cuda_error& __err)
      {
        if (__err.status() == cudaErrorMemoryAllocation)
        {
          ::cuda::std::__throw_bad_alloc();
        }
        else
        {
          throw __err;
        }
      }
```

libcudacxx/include/cuda/std/__pstl/reduce.h
```diff
@@ -0,0 +1,116 @@
...
+      ::cuda::std::move(__last),
+      ::cuda::std::move(__init),
+      ::cuda::std::move(__func));
+  }
+  else
+  {
+    static_assert(__always_false_v<_Policy>, "Parallel cuda::std::reduce requires at least one selected backend");
+    return ::cuda::std::reduce(
+      ::cuda::std::move(__first), ::cuda::std::move(__last), ::cuda::std::move(__init), ::cuda::std::move(__func));
+  }
```
bernhardmgruber, line 81: Q: is there a reason why do need the `if constexpr`?

libcudacxx/include/cuda/std/__pstl/for_each.h
```diff
@@ -42,16 +43,14 @@ _CCCL_BEGIN_NAMESPACE_ARCH_DEPENDENT
...
   [[maybe_unused]] auto __dispatch =
     ::cuda::std::execution::__pstl_select_dispatch<::cuda::std::execution::__pstl_algorithm::__for_each_n, _Policy>();
   if constexpr (::cuda::std::execution::__pstl_can_dispatch<decltype(__dispatch)>)
   {
-    (void) __dispatch(::cuda::std::move(__policy),
-                      ::cuda::std::move(__first),
-                      ::cuda::std::distance(__first, __last),
-                      ::cuda::std::move(__func));
+    (void) __dispatch(
+      __policy, ::cuda::std::move(__first), ::cuda::std::distance(__first, __last), ::cuda::std::move(__func));
```
bernhardmgruber: Critical: There is a potential use-after-move bug here regarding `__first`, so:
```suggestion
    const auto __n = ::cuda::std::distance(__first, __last);
    (void) __dispatch(
      __policy, ::cuda::std::move(__first), __n, ::cuda::std::move(__func));
```

libcudacxx/include/cuda/std/__pstl/cuda/reduce.h
miscco: I am not in favor of this, technically forward iterator are supported by the standard, so we might be inclined to add support for those

bernhardmgruber: But just for my understanding, we do `static_assert` and abort compilation if the iterators are not random access. And this code:
```c++
      return ::cuda::std::reduce(
        ::cuda::std::move(__first), ::cuda::std::move(__last), ::cuda::std::move(__init), ::cuda::std::move(__func));
```
is never executed?

libcudacxx/include/cuda/std/__pstl/cuda/reduce.h
miscco: Moved to _CCCL_ASSERT_CUDA_API

libcudacxx/include/cuda/std/__pstl/cuda/reduce.h
miscco: yes

libcudacxx/include/cuda/std/__pstl/reduce.h
miscco: Oh this is for when there is no backend available

miscco: If we compile without a cuda compiler, then it would fail to find the `__dispatch` codepath

bernhardmgruber: But we only have one backend ;) So I assume you wrote the code in a forward compatible manner when we add new backends?

miscco: I did but we also have no backend, because we must also compile with just a C++ compiler

PR #6780, opened 2025-11-25
Add a type and operation enum to CUB

bernhardmgruber:
And use it in the scan tunings to test it.

- [x] No SASS diff for `cub.test.device.scan.lid_0.types_0` on sm120

### DIFF
--- a/cub/cub/device/dispatch/tuning/tuning_scan.cuh
+++ b/cub/cub/device/dispatch/tuning/tuning_scan.cuh
@@ -18,6 +18,7 @@
 #include <cub/block/block_load.cuh>
 #include <cub/block/block_scan.cuh>
 #include <cub/block/block_store.cuh>
+#include <cub/device/dispatch/tuning/common.cuh>
 #include <cub/thread/thread_load.cuh>
 #include <cub/util_device.cuh>
 #include <cub/util_type.cuh>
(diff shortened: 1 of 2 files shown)


cub/cub/device/dispatch/tuning/tuning_scan.cuh
```diff
@@ -139,14 +117,14 @@ constexpr _CCCL_HOST_DEVICE offset_size classify_offset_size()
...
-          op_type OpTypeT,
+          op_kind_t OpTypeT,
           primitive_accum PrimitiveAccumulator = is_primitive_accum<AccumT>(),
           offset_size OffsetSize               = classify_offset_size<OffsetT>(),
           value_size ValueSize                 = classify_value_size<ValueT>()>
 struct sm75_tuning;
 
 template <class ValueT, class AccumT, class OffsetT>
-struct sm75_tuning<ValueT, AccumT, OffsetT, op_type::plus, primitive_accum::yes, offset_size::_8, value_size::_4>
+struct sm75_tuning<ValueT, AccumT, OffsetT, op_kind_t::plus, primitive_accum::yes, offset_size::_8, value_size::_4>
```
fbusato, line 127: question. Most of the code is not in the detail namespace. Is this something that we can change without affecting users?

bernhardmgruber: The changes entities are in the namespace `cub::detail::scan`. We should be fine.

Issue #1244, opened 2023-12-28
[BUG]: Low performance of sorting with OMP backend


dexter2206:
### Is this a duplicate?

- [X] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cccl/issues) for this bug and that I agree to the [Code of Conduct](CODE_OF_CONDUCT.md)

### Type of Bug

Performance

### Component

Thrust

### Describe the bug

Performance of `thrust::sort` with OpenMP backend seems to be far worse than even standard `std::sort`. The code provided in "How to reproduce" section compares timings of sorting with `std` and `thrust`. I tested it on two machines, one with AMD Ryzen 5950X and 32GB DDR4 memory, and the other with Intel i7-1365U and 32GB DDR5 memory, both on Fedora 38 operating system. In the first case, `std::sort` manages to sort `100 000 000` floats in 32s and thrust performs the same task in ~87s. In the second case, the timings are ~32s and ~87s respectively. Looking at htop it seems that thrust indeed uses some parallelization as all cores being used.

### How to Reproduce

1. Save the following file as `sorting_comparison.cpp`:
```C++
#include <algorithm>
#include <chrono>
#include <iostream>
#include <random>
#include <vector>

#include <thrust/sort.h>
#include <thrust/device_vector.h>

using namespace std;
using namespace std::chrono;

const int VEC_SIZE = 100'000'000;

vector<float> random_vector(int size) {
    vector<float> data(size);
    random_device rnd_device;
    mt19937 mersenne_engine{rnd_device()};
    uniform_real_distribution<float> dist{-10.0f, 10.0f};

    generate(begin(data), end(data),
             [&dist, &mersenne_engine]() { return dist(mersenne_engine); });

    return data;
}

int main() {
    vector<float> data = random_vector(VEC_SIZE);
    thrust::device_vector<float> data_for_thrust(data);

    auto start = high_resolution_clock::now();
    sort(begin(data), end(data));
    auto stop = high_resolution_clock::now();
    cout << "std took: " << duration_cast<seconds>(stop-start).count() << "s" << endl;

    start = high_resolution_clock::now();
    thrust::sort(begin(data_for_thrust), end(data_for_thrust));
    stop = high_resolution_clock::now();
    cout << "thrust took: " << duration_cast<seconds>(stop-start).count() << "s" << endl;

    return 0;
}
```
2. Make sure thrust is available in include path and compile the file with:
   ```bash
[...]

PR #6784, opened 2025-11-25
Thrust: Optimize OMP sort

Pansysk75:
Addresses issue https://github.com/NVIDIA/cccl/issues/1244

Takes advantage of `inplace_merge` from libcudacxx, using >50% less memory for the merge (plus some additional optimizations).
Reduces `thrust::sort` execution time by ~20% on my system.

(Performance was already improved since the Issue was opened, as arithmetic types now dispatch to `radix_sort`. So, perhaps the referenced issue can be closed).

Question:
A similar optimization can be made in the implementation of [stable_merge_sort](https://github.com/NVIDIA/cccl/blob/b8a0a9598f43612fb5d661c6f896fd8db6a34e5d/thrust/thrust/system/detail/sequential/stable_merge_sort.h#L44-L58). However, `thrust::merge` is not necessarily called sequentially there, so I didn't feel comfortable forcing that by directly dispatching to `cuda::std::inplace_merge`
A solution would be to provide `thrust::inplace_merge` directly, and let the dispatch be controlled by the execution policy, as usual. Would you be interested in me contributing towards that?

### DIFF
--- a/thrust/thrust/system/omp/detail/sort.h
+++ b/thrust/thrust/system/omp/detail/sort.h
@@ -42,10 +42,8 @@ void inplace_merge(execution_policy<DerivedPolicy>& exec,
 {
   using value_type = thrust::detail::it_value_t<RandomAccessIterator>;
 
-  thrust::detail::temporary_array<value_type, DerivedPolicy> a(exec, first, middle);
-  thrust::detail::temporary_array<value_type, DerivedPolicy> b(exec, middle, last);
-
-  thrust::merge(thrust::seq, a.begin(), a.end(), b.begin(), b.end(), first, comp);
+  thrust::detail::temporary_array<value_type, DerivedPolicy> buf(exec, first, middle);
+  thrust::merge(thrust::seq, buf.begin(), buf.end(), middle, last, first, comp);
 }
 
 template <typename DerivedPolicy,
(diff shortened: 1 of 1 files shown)


gonidelis:
I love this. My local 32 core Threadripper reported an average of ~35% speedup with these changes so seems like there was even more leeway for SOL.

Cleaner code ✅ 
faster code ✅
👇👇👇
**mega approve**

Not sure what the execution semantics for the libcu++ algorithms are. @bernhardmgruber @miscco  is there a way we can call `::cuda::std::inplace_merge` in parallel? I know that in the long term all should be under `cuda::std` but that doesn't help much now.

@Pansysk75  Feel free to proceed with `stable_merge_sort` in a new pr once path is clear from the experts.

thrust/thrust/system/omp/detail/sort.h
```diff
@@ -131,8 +114,7 @@ void stable_sort(
 
       if ((p_i % h) == 0 && c > b)
       {
-        sort_detail::inplace_merge(
-          exec, first + decomp[a].begin(), first + decomp[b].end(), first + decomp[c].end(), comp);
+        ::cuda::std::inplace_merge(first + decomp[a].begin(), first + decomp[b].end(), first + decomp[c].end(), comp);
```
bernhardmgruber: Important: This is a break in behavior since before, if the execution policy specified an allocator, we would use it. Now we ignore the execution policy (and a potential user provided allocator) and use the allocator for `libcu++`. 

There are a couple of options. We could check whether `exec` is derived from `thrust::detail::execute_with_allocator` and call the old implementation if it is. Or we could call `cuda::std::__inplace_merge` and do the allocation using `thrust::get_temporary_buffer<T>(exec, ...)` ourselves. I would strongly prefer the latter. I hope the benchmark is still good afterwards.

So basically, we would replace the above line with the implementation: https://github.com/NVIDIA/cccl/blob/fba9f83410301d03bc70bdaa502985a4abe62447/libcudacxx/include/cuda/std/__algorithm/inplace_merge.h#L252-L267.

bernhardmgruber (changes requested):
Thx for providing this great fix! One issue though: we have to respect any user-provided allocator that comes as part of the execution policy.

thrust/thrust/system/omp/detail/sort.h
Pansysk75: Thank you for the feedback! I completely missed that role of the execution policy.

In addition to what you mentioned, it would also be wise to avoid depending too much on the specifics of the internal implementation. I will try to find a clean way to go about it.

Pansysk75:
Hello, I played around with different implementations of this, but had way too many questions, so I ended up with the simplest possible change.

Here are some measurements, from a simple sort benchmark on random vectors of floats.
The previous ones were on my local machine, and turns out were very inconsistent
These are on a 64-core Intel Xeon, on an isolated cluster node. The compiler / standard library is GCC 14.3 libstdc++:
```
Size    Impl           Avg (ms)  Median (ms)  StdDev (ms)   Reps
----------------------------------------------------------------
10^4    std                0.54         0.54         0.01   7311
        thrust(old)        0.15         0.15         0.04   7263
        thrust(new)        0.14         0.14         0.21   7311
----------------------------------------------------------------
10^5    std                6.76         6.77         0.06    624
        thrust(old)        1.60         1.17         3.66    598
        thrust(new)        1.26         1.17         1.41    624
----------------------------------------------------------------
10^6    std               79.51        79.54         0.50     56
        thrust(old)       10.66        10.49         0.52     56
        thrust(new)       10.62        10.57         0.24     56
----------------------------------------------------------------
10^7    std              921.20       918.89         3.61      5
        thrust(old)      114.30       113.34         2.45      5
        thrust(new)      107.58       105.67         2.60      5
----------------------------------------------------------------
10^8    std            10235.73     10239.27        34.73      3
        thrust(old)     1043.86      1045.03         2.65      3
        thrust(new)      992.17       993.52         2.75      3
----------------------------------------------------------------
10^9    std           103277.37    103317.55        60.10      3
        thrust(old)     5961.19      5979.49        31.15      3
        thrust(new)     5084.92      5114.79        67.83      3
----------------------------------------------------------------
```

PR #6791, opened 2025-11-26
Improve CMake package handling, add MSVC compat flags to libcudacxx's public interface.

alliepiper:
Fixes some edge cases in the internal dependency graphs that were broken by the libcudacxx/thrust/cub interdependency.

Adds `/Zc:preprocessor` and `/Zc:__cplusplus` to `libcudacxx::libcudacxx`'s public interface. Re-check enabled languages / compilers for repeated `find_package` calls to CCCL projects, since `CUDA` / `CUDA_HOST` compilers may be unknown during the first pass through the find module.

Changes the `/Zc:preprocessor` check into an `#error` rather than a `#pragma message`. MSVC will encounter preprocessor issues and exit with non-obvious errors before handling the `#pragma message`, so users were not seeing the message when it was needed.

Misc:
- Adds `cccl_get_<project>` helpers for `cccl`, `thrust`, `cub`, `libcudacxx`, `cudax` and `CUDAToolkit` to reduce noise and keep find_package logic consistent.
- Ensure that all internal projects are finding and linking their interdependencies correctly using new helpers.
- Set `<Project>_DIR` in the CMake cache when including CCCL with `add_subdirectory`/CPM. Ensures that future calls to `find_package(CCCL)` will resolve to the same config files, but won't override user settings.
- Cleaned up `enable_language(CUDA)` usages.
- Update `ci/test_packaging.sh` to:
  - Use the `HEAD` commit of the local repository when run locally.
  - Always use the local repo checkout as upstream when fetching from CPM in examples both locally and in CI. Much faster than network clones.

### DIFF
--- a/cub/cub/util_namespace.cuh
+++ b/cub/cub/util_namespace.cuh
@@ -25,8 +25,6 @@
 
 #include <cub/detail/detect_cuda_runtime.cuh>
 
-#include <cuda/std/__cccl/preprocessor.h>
-
 // Prior to 1.13.1, only the PREFIX/POSTFIX macros were used. Notify users
 // that they must now define the qualifier macro, too.
 #if (defined(CUB_NS_PREFIX) || defined(CUB_NS_POSTFIX)) && !defined(CUB_NS_QUALIFIER)
--- a/thrust/thrust/detail/config/namespace.h
+++ b/thrust/thrust/detail/config/namespace.h
@@ -29,8 +29,6 @@
 #include <thrust/detail/config/device_system.h>
 #include <thrust/version.h>
 
-#include <cuda/std/__cccl/preprocessor.h>
-
 /**
  * \file namespace.h
  * \brief Utilities that allow `thrust::` to be placed inside an


alliepiper:
pre-commit.ci autofix

cub/cub/util_namespace.cuh
```diff
@@ -25,6 +25,8 @@
 
 #include <cub/detail/detect_cuda_runtime.cuh>
 
+#include <cuda/std/__cccl/preprocessor.h>
+
```
miscco:
```suggestion
```

thrust/thrust/detail/config/namespace.h
```diff
@@ -29,6 +29,8 @@
 #include <thrust/detail/config/device_system.h>
 #include <thrust/version.h>
 
+#include <cuda/std/__cccl/preprocessor.h>
+
```
miscco:
```suggestion
```

lib/cmake/cccl/cccl-config.cmake
```diff
@@ -29,7 +29,7 @@ if (NOT TARGET CCCL::CCCL)
 endif()
 
 message(STATUS "Finding CCCL components: ${components}")
-message(STATUS "cccl_cmake_dir: ${cccl_cmake_dir}")
+message(VERBOSE "cccl_cmake_dir: ${cccl_cmake_dir}")
```
bernhardmgruber, line 32: Suggestion: This seems like a remainder of some testing. Can we revert it?

examples/CMakeLists.txt
```diff
@@ -58,34 +60,31 @@ cccl_add_compile_test(
...
+  ${cmake_cpm_opts}
+)
 
 foreach (DEVICE_SYSTEM IN ITEMS CUDA OMP TBB CPP)
   cccl_add_compile_test(
     test_name
     cccl.example
     thrust_flexible_device_system
     "${DEVICE_SYSTEM}"
+    --log-level=VERBOSE
```
bernhardmgruber, line 87: Q: Do we need this or was this used for testing?

lib/cmake/cccl/cccl-config.cmake
alliepiper: It's helpful for troubleshooting, but noisy, so I moved it to the `VERBOSE` stream so it won't spam users.

examples/CMakeLists.txt
alliepiper: This is intentional -- we always do verbose configures in CI to get more detailed target setup info.

PR #6797, opened 2025-11-27
Implement std::weibull_distribution

RAMitchell:


libcudacxx/include/cuda/std/__random/weibull_distribution.h
```diff
@@ -0,0 +1,189 @@
...
+#if _CCCL_STD_VER <= 2017
+    [[nodiscard]] _CCCL_API friend constexpr bool operator!=(const param_type& __x, const param_type& __y) noexcept
+    {
+      return !(__x == __y);
+    }
+#endif // _CCCL_STD_VER <= 2017
+  };
+
+private:
+  param_type __p_;
```
miscco: Q: Should this be value initialized?
```suggestion
  param_type __p_{};
```

libcudacxx/include/cuda/std/__random/weibull_distribution.h
```diff
@@ -0,0 +1,189 @@
...
+      : __p_{param_type{__a, __b}}
+  {}
+  _CCCL_API constexpr explicit weibull_distribution(const param_type& __p) noexcept
+      : __p_{__p}
+  {}
+  _CCCL_API constexpr void reset() noexcept {}
+
+  // generating functions
+  template <class _URng>
+  [[nodiscard]] _CCCL_API result_type operator()(_URng& __g)
```
miscco, line 96: Q: Should the call operator be `nodiscard`? I believe we do not do that for the other distributions

RAMitchell: Should be nodiscard for distributions not for engines.

libcudacxx/include/cuda/std/__random/weibull_distribution.h
```diff
@@ -0,0 +1,189 @@
...
+  {
+    return (*this)(__g, __p_);
+  }
+  template <class _URng>
+  [[nodiscard]] _CCCL_API result_type operator()(_URng& __g, const param_type& __p)
+  {
+    return __p.b() * cuda::std::pow(exponential_distribution<result_type>{}(__g), result_type{1} / __p.a());
```
fbusato:
```suggestion
    return __p.b() * ::cuda::std::pow(exponential_distribution<result_type>{}(__g), result_type{1} / __p.a());
```

fbusato: maybe the compiler is able to perform this optimization but, would make sense to more `result_type{1} / __p.a()` outside?

RAMitchell: Seems like premature optimisation. I would leave it unless we have a reason to benchmark it and improve it.

fbusato (approved):

PR #6810, opened 2025-11-28
Replace inline ptx with intrinsics

davebayer:
Found 2 occurrences

### DIFF
--- a/libcudacxx/include/cuda/std/__thread/threading_support_cuda.h
+++ b/libcudacxx/include/cuda/std/__thread/threading_support_cuda.h
@@ -33,7 +33,7 @@ _CCCL_API inline void __cccl_thread_yield() {}
 
 _CCCL_API inline void __cccl_thread_sleep_for(::cuda::std::chrono::nanoseconds __ns)
 {
-  NV_IF_TARGET(NV_IS_DEVICE,
+  NV_IF_TARGET(NV_PROVIDES_SM_70,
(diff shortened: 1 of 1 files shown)


bernhardmgruber (approved):

miscco (approved):

libcudacxx/include/cuda/std/__thread/threading_support_cuda.h
```diff
@@ -33,9 +33,9 @@ _CCCL_API inline void __cccl_thread_yield() {}
 
 _CCCL_API inline void __cccl_thread_sleep_for(::cuda::std::chrono::nanoseconds __ns)
 {
-  NV_IF_TARGET(NV_IS_DEVICE,
+  NV_IF_TARGET(NV_PROVIDES_SM_70,
                (auto const __step = __ns.count(); assert(__step < numeric_limits<unsigned>::max());
```
fbusato, line 37: should not be `_CCCL_ASSERT(...)`?

davebayer: Yeah, I wanted to make a PR that replaces all uses of `assert` with `_CCCL_ASSERT` in libcu++

davebayer: Addressed in #6825

PR #6814, opened 2025-11-30
Improve our `WarpReduce` implementation

miscco:
For the warpspeed scan PR we need a common implementation of `__reduce_or_sync`

I wanted to add a specialization for that to `WarpReduce` and found some ... historical artifacts.

This cleanes those up and centralizes the optimizations into a single place. Finally it adds the specializations for `__reduce_and_sync` and `__reduce_or_sync`

There is no standard functor for `logical_xor` though :(

### DIFF
--- a/cub/cub/warp/specializations/warp_reduce_shfl.cuh
+++ b/cub/cub/warp/specializations/warp_reduce_shfl.cuh
@@ -87,28 +87,39 @@ template <class T, class ReductionOp, class = void>
 inline constexpr bool can_use_reduce_and_sync = false;
 
 template <class T>
-inline constexpr bool can_use_reduce_and_sync<T,
-                                              ::cuda::std::logical_and<>,
-                                              ::cuda::std::void_t<decltype(__reduce_and_sync(0xFFFFFFFF, T{}))>> =
-  ::cuda::std::is_same_v<unsigned int, T>;
+inline constexpr bool
+  can_use_reduce_and_sync<T, ::cuda::std::bit_and<>, ::cuda::std::void_t<decltype(__reduce_and_sync(0xFFFFFFFF, T{}))>> =
+    ::cuda::std::is_same_v<unsigned int, T>;
 
 template <class T>
-inline constexpr bool can_use_reduce_and_sync<T,
-                                              ::cuda::std::logical_and<T>,
-                                              ::cuda::std::void_t<decltype(__reduce_and_sync(0xFFFFFFFF, T{}))>> =
-  ::cuda::std::is_same_v<unsigned int, T>;
+inline constexpr bool
+  can_use_reduce_and_sync<T, ::cuda::std::bit_and<T>, ::cuda::std::void_t<decltype(__reduce_and_sync(0xFFFFFFFF, T{}))>> =
+    ::cuda::std::is_same_v<unsigned int, T>;
 
 template <class T, class ReductionOp, class = void>
 inline constexpr bool can_use_reduce_or_sync = false;
 
 template <class T>
 inline constexpr bool
-  can_use_reduce_or_sync<T, ::cuda::std::logical_or<>, ::cuda::std::void_t<decltype(__reduce_or_sync(0xFFFFFFFF, T{}))>> =
+  can_use_reduce_or_sync<T, ::cuda::std::bit_or<>, ::cuda::std::void_t<decltype(__reduce_or_sync(0xFFFFFFFF, T{}))>> =
     ::cuda::std::is_same_v<unsigned int, T>;
 
 template <class T>
 inline constexpr bool
-  can_use_reduce_or_sync<T, ::cuda::std::logical_or<T>, ::cuda::std::void_t<decltype(__reduce_or_sync(0xFFFFFFFF, T{}))>> =
+  can_use_reduce_or_sync<T, ::cuda::std::bit_or<T>, ::cuda::std::void_t<decltype(__reduce_or_sync(0xFFFFFFFF, T{}))>> =
+    ::cuda::std::is_same_v<unsigned int, T>;
+
+template <class T, class ReductionOp, class = void>
+inline constexpr bool can_use_reduce_xor_sync = false;
+
+template <class T>
+inline constexpr bool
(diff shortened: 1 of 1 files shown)


miscco:
Note to reviewers, its simplest to go by commit

cub/cub/warp/specializations/warp_reduce_shfl.cuh
```diff
@@ -43,29 +43,60 @@ CUB_NAMESPACE_BEGIN
...
+template <class T, class ReductionOp, class = void>
+inline constexpr bool reduce_max_exists = false;
+
+template <class T>
+inline constexpr bool reduce_max_exists<T, ::cuda::maximum<>, decltype(__reduce_max_sync(0xFFFFFFFF, T{}))> =
+  (::cuda::std::is_same_v<int, T> || ::cuda::std::is_same_v<unsigned int, T>);
+
+template <class T>
+inline constexpr bool reduce_max_exists<T, ::cuda::maximum<T>, decltype(__reduce_max_sync(0xFFFFFFFF, T{}))> =
+  (::cuda::std::is_same_v<int, T> || ::cuda::std::is_same_v<unsigned int, T>);
```
miscco: Note to self:

I believe we can relax that constrain to `is_integral<T> && sizeof(T) <= sizeof(int)` and then in the optimization  pass below cast to `using promoted_t = ::cuda::std::conditional_t<::cuda::std::is_signed_v<T>, int, unsigned>;`

Then casting back to the original T should be fine. That said, I believe this to be a rather invasiev change so I leave that for a followup

cub/cub/warp/specializations/warp_reduce_shfl.cuh
```diff
@@ -441,57 +460,6 @@ struct WarpReduceShfl
...
-   *   Up-offset to pull from
-   *
-   * @param[in] is_small_unsigned
-   *   Marker type indicating whether T is a small unsigned integer
-   */
-  template <typename _Tp, typename ReductionOp>
-  _CCCL_DEVICE _CCCL_FORCEINLINE _Tp ReduceStep(
-    _Tp input, ReductionOp reduction_op, int last_lane, int offset, ::cuda::std::true_type /*is_small_unsigned*/)
-  {
-    return ReduceStep(input, reduction_op, last_lane, offset);
```
miscco, line 466: "Dispatch to specialized implementations"

oleksandr-pavlyk:
Per https://docs.nvidia.com/cuda/cuda-c-programming-guide/#warp-reduce-synopsis  `__reduce_xor_sync` is there.

miscco:
> Per https://docs.nvidia.com/cuda/cuda-c-programming-guide/#warp-reduce-synopsis `__reduce_xor_sync` is there.

That is true, but we do not have a function object that does logical `XOR` in the library

cub/cub/warp/specializations/warp_reduce_shfl.cuh
miscco, line 357: Note to reviewer, I believe we can also considerably clean those up, but that would make the PR quite unwieldly

oleksandr-pavlyk:
> > Per https://docs.nvidia.com/cuda/cuda-c-programming-guide/#warp-reduce-synopsis `__reduce_xor_sync` is there.
> 
> That is true, but we do not have a function object that does logical `XOR` in the library

I think we should map `cuda::std::bit_xor` for unsigned integral types to `__reduce_xor_sync`.

cub/cub/warp/specializations/warp_reduce_shfl.cuh
bernhardmgruber: :rofl:

bernhardmgruber:
Unfortunately, this is creating a lot more instructions on SM90 (and 100) for test `cub.test.block.reduce.dimx_32.dimyz_1` (compiled for 75,90,100)

For example:
image
We previously generated a `REDUX.MAX.S32` and now have a lot of shuffles.

This also affects `REDUX.SUM.S32` and `CREDUX.MAX.S32`

miscco:
Thanks @bernhardmgruber for checking, I found that I missed a `void_t` in the move to inline variables. With that I cannot find any SASS changes in the test

cub/cub/warp/specializations/warp_reduce_shfl.cuh
```diff
@@ -653,7 +543,50 @@ struct WarpReduceShfl
   template <bool ALL_LANES_VALID, typename ReductionOp>
   _CCCL_DEVICE _CCCL_FORCEINLINE T Reduce(T input, int valid_items, ReductionOp reduction_op)
   {
-    return ReduceImpl(bool_constant_v<ALL_LANES_VALID>, input, valid_items, reduction_op);
+    if constexpr (ALL_LANES_VALID)
```
miscco, line 546: Question: Why do we only do the `__reduce_meow_sync` optimization when having a full lane, it explicitly takes the valid lanes as arguments?

cub/cub/warp/specializations/warp_reduce_shfl.cuh
```diff
@@ -43,29 +44,83 @@ CUB_NAMESPACE_BEGIN
...
-struct reduce_add_exists : ::cuda::std::false_type
-{};
+template <class T, class ReductionOp, class = void>
+inline constexpr bool can_use_reduce_add_sync = false;
 
 template <class T>
-struct reduce_add_exists<T, decltype(__reduce_add_sync(0xFFFFFFFF, T{}))> : ::cuda::std::true_type
-{};
+inline constexpr bool
+  can_use_reduce_add_sync<T, ::cuda::std::plus<>, ::cuda::std::void_t<decltype(__reduce_add_sync(0xFFFFFFFF, T{}))>> =
```
fbusato, line 52: can we also add integer promotion here?

fbusato: already handled in the other PR

miscco: https://github.com/NVIDIA/cccl/pull/6819

This PR tries hard to not change SASS, because we want it in relatively fast and I leave the optimization that greatly affects SASS for the other PR

PR #2779, opened 2024-11-12
Optimize implementation of `cuda::std::conditional_t`

miscco:
The current implementation always instantiates `conditional` at least once.

However, we can get aways with a total of 2 class instantiations by using the implementation of the internal `_If` alias template.

This not only makes `conditional_t` more efficient but also simplifies usage, because no one has to think about what to take

PR #6820, opened 2025-12-01
Optimize `cuda::std::conditional_t`

davebayer:
For `_If` we use an implementation that instantiates less types. This PR uses the implementation for `cuda::std::conditional_t`, too, and removes the `_If` completely.


davebayer:
Duplicate of #2779

PR #6824, opened 2025-12-01
Implement std::piecewise_constant_distribution

RAMitchell:
This distribution is somewhat more difficult than the others due to dynamic memory allocation.

- Memory is allocated using unique_ptr and storing lengths
- Copying is deleted for both param and the distribution
- Methods that return std::vector now return cuda::std::span
- the param() getter returns a const ref instead of a copy

If we agree on the implementation here, the same will be used for:
- piecewise_linear_distribution
- discrete_distribution

### DIFF
--- a/libcudacxx/include/cuda/std/__random/piecewise_constant_distribution.h
+++ b/libcudacxx/include/cuda/std/__random/piecewise_constant_distribution.h
@@ -20,13 +20,15 @@
 #  pragma system_header
 #endif // no system header
 
+#include <cuda/std/__algorithm/copy_n.h>
 #include <cuda/std/__algorithm/upper_bound.h>
-#include <cuda/std/__memory_>
+#include <cuda/std/__iterator/distance.h>
+#include <cuda/std/__memory/unique_ptr.h>
+#include <cuda/std/__numeric/accumulate.h>
+#include <cuda/std/__numeric/partial_sum.h>
 #include <cuda/std/__random/is_valid.h>
 #include <cuda/std/__random/uniform_real_distribution.h>
-#include <cuda/std/__utility/swap.h>
 #include <cuda/std/initializer_list>
-#include <cuda/std/numeric>
 #include <cuda/std/span>
 
 #if !_CCCL_COMPILER(NVRTC)
--- a/libcudacxx/include/cuda/std/__random_
+++ b/libcudacxx/include/cuda/std/__random_
@@ -27,6 +27,7 @@
 #include <cuda/std/__random/linear_congruential_engine.h>
 #include <cuda/std/__random/normal_distribution.h>
 #include <cuda/std/__random/philox_engine.h>
+#include <cuda/std/__random/piecewise_constant_distribution.h>
 #include <cuda/std/__random/seed_seq.h>
 #include <cuda/std/__random/uniform_int_distribution.h>
 #include <cuda/std/__random/uniform_real_distribution.h>


libcudacxx/include/cuda/std/__random/piecewise_constant_distribution.h
```diff
@@ -0,0 +1,402 @@
...
+#  pragma clang system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_MSVC)
+#  pragma system_header
+#endif // no system header
+
+#include <cuda/std/__algorithm/upper_bound.h>
+#include <cuda/std/__memory_>
+#include <cuda/std/__random/is_valid.h>
+#include <cuda/std/__random/uniform_real_distribution.h>
+#include <cuda/std/__utility/swap.h>
```
fbusato: looks unused

libcudacxx/include/cuda/std/__random/piecewise_constant_distribution.h
```diff
@@ -0,0 +1,402 @@
...
+#include <cuda/std/initializer_list>
+#include <cuda/std/numeric>
```
fbusato: can we use more precise headers?

libcudacxx/include/cuda/std/__random/piecewise_constant_distribution.h
```diff
@@ -0,0 +1,402 @@
...
+  static_assert(__libcpp_random_is_valid_realtype<_RealType>, "RealType must be a supported floating-point type");
+
+public:
+  // types
+  using result_type = _RealType;
+
+  class param_type
+  {
+  private:
+    unique_ptr<result_type[]> __b_;
```
fbusato, line 54: header missing?

libcudacxx/include/cuda/std/__random/piecewise_constant_distribution.h
```diff
@@ -0,0 +1,402 @@
...
+    {
+      // __densities_ contains non-normalized areas
+      result_type __total_area =
+        ::cuda::std::accumulate(__densities_.get(), __densities_.get() + __size_densities_, result_type());
+      for (size_t __i = 0; __i < __size_densities_; ++__i)
+      {
+        __densities_[__i] /= __total_area;
+      }
+      // __densities_ contains normalized areas
+      __areas_.reset(new result_type[__size_densities_]);
```
fbusato, line 71: uhmm that's a problem for a host-device implementation. Dynamic memory is not great on device side. Secondly, we cannot initialize on the host and move to the device. @miscco for visibility

libcudacxx/include/cuda/std/__random/piecewise_constant_distribution.h
miscco: :+1: We should only ever include precisely what we need, some headers like tuple implement a feature, but for conglomerates we should include what is in `cuda/std/__numeric/...`

libcudacxx/include/cuda/std/__random/piecewise_constant_distribution.h
miscco: yes this is an issue and I am not sure how to properly address this as there is no standard conforming way to achieve this

RAMitchell: I know performance would be bad but I think we should aim to implement it for completeness. Is it sufficient to delete the copy constructor to prevent it from being moved from host to device?

RAMitchell:
One idea is to implement only the intializer list type constructor from https://en.cppreference.com/w/cpp/numeric/random/piecewise_constant_distribution/piecewise_constant_distribution.html
then add a new hidden template parameter N for array member size, which would be inferred.

This way we create a distribution only usable with constant size but it can be copied between device/host.

libcudacxx/include/cuda/std/__random/piecewise_constant_distribution.h
fbusato: how many distributions do we have with dynamic memory? they are too problematic IMO. I'm not sure that we want them. Alternatively, we need a different interface in `cuda::` to be efficiently used on device code.

RAMitchell: There are 3 sampling distributions. As mentioned below, we could maybe just support the intializer list constructor and use array storage.

fbusato: I agree in principle. We could document the constraint. @miscco what do you think?

miscco: fine by me

RAMitchell:
Current consensus seems to be implementing a variant of this class under the cuda:: namespace with documented limitations, closing this for now.

PR #4460, opened 2025-04-15
Prefer `_CCCL_UNREACHABLE_WITH_CHECK` over `_CCCL_UNREACHABLE`

fbusato:
## Description

Related to https://github.com/NVIDIA/cccl/pull/4443. ~`cuda::std::unreachable()`~  `_CCCL_UNREACHABLE_WITH_CHECK ` uses an assertion to check if the path is actually reached

PR #6825, opened 2025-12-01
Replace internal `assert` with `_CCCL_ASSERT` in libcu++

davebayer:


libcudacxx/codegen/generators/fence.h
```diff
@@ -88,7 +88,7 @@ static inline _CCCL_DEVICE void __atomic_thread_fence_cuda(int __memorder, _Sco)
         case __ATOMIC_ACQ_REL: [[fallthrough]];
         case __ATOMIC_RELEASE: __cuda_atomic_fence(_Sco{}, __atomic_cuda_acq_rel{}); break;
         case __ATOMIC_RELAXED: break;
-        default: assert(0);
```
fbusato, line 91: `assert(0)` sounds like the path is never reached rather than a real assertion. Anyway, that's fine for this PR

davebayer: Would you prefer `_CCCL_UNREACHABLE()` in this case?

fbusato: `_CCCL_UNREACHABLE` should be the right solution but it doesn't provide any debug info. I already tried to improve it but I didn't want to spent time on this https://github.com/NVIDIA/cccl/pull/4460

PR #6838, opened 2025-12-02
Add ptxas local memory usage warnings to cub builds

davebayer:
`ptxas` provides some warnings on local memory usage and register spilling to local memory.

This PR adds these warnings to cub Release builds. They are not treated as errors.

### DIFF
--- a/cub/cmake/CubBuildCompilerTargets.cmake
+++ b/cub/cmake/CubBuildCompilerTargets.cmake
@@ -10,9 +10,30 @@ function(cub_build_compiler_targets)
 
   thrust_create_target(cub.thrust HOST CPP DEVICE CUDA)
 
+  set(ptxas_compile_options)
+  if (CCCL_ENABLE_PTXAS_WARNINGS)
+    list(
+      APPEND ptxas_compile_options
+      "--warn-on-spills"
+      "--warn-on-local-memory-usage"
+    )
+  endif()
+
   set(cuda_compile_options)
   set(cxx_compile_options)
   set(cxx_compile_definitions)
+
+  # append ptxas compile options to cuda_compile_options with compiler specific prefix
+  foreach (ptxas_compile_option ${ptxas_compile_options})
+    if (
+      "${CMAKE_CUDA_COMPILER_ID}" STREQUAL "NVIDIA"
+      OR "${CMAKE_CUDA_COMPILER_ID}" STREQUAL "NVHPC"
+    )
+      list(APPEND cuda_compile_options "-Xptxas=${ptxas_compile_option}")
+    elseif ("${CMAKE_CUDA_COMPILER_ID}" STREQUAL "CLANG")
+      list(APPEND cuda_compile_options "-Xcuda-ptxas ${ptxas_compile_option}")
+    endif()
+  endforeach()
 
   cccl_build_compiler_interface(
     cub.compiler_flags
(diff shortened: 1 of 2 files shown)


cub/cmake/CubBuildCompilerTargets.cmake
```diff
@@ -10,10 +10,27 @@ function(cub_build_compiler_targets)
...
+    if (
+      "${CMAKE_CUDA_COMPILER_ID}" STREQUAL "NVIDIA"
+      OR "${CMAKE_CUDA_COMPILER_ID}" STREQUAL "NVHPC"
+    )
+      list(APPEND cuda_compile_options "-Xptxas=${ptxas_compile_option}")
+    elseif ("${CMAKE_CUDA_COMPILER_ID}" STREQUAL "CLANG")
+      list(APPEND cuda_compile_options "-Xcuda-ptxas ${ptxas_compile_option}")
+    endif()
+  endforeach()
+
```
miscco, line 37: @alliepiper should we move that into the `cccl_build_compiler_interface` facility or is hat too special?

davebayer:
Okay! And do we want this for Release only or for ReWithDebInfo, too?

davebayer:
I see all of these warnings are generated in the CI, maybe we will want to suppress these warnings there and keep them for local builds only

cub/cmake/CubBuildCompilerTargets.cmake
```diff
@@ -10,10 +10,27 @@ function(cub_build_compiler_targets)
...
   set(cxx_compile_options)
   set(cxx_compile_definitions)
 
+  # append ptxas compile options to cuda_compile_options with compiler specific prefix
+  foreach (ptxas_compile_option ${ptxas_compile_options})
+    if (
+      "${CMAKE_CUDA_COMPILER_ID}" STREQUAL "NVIDIA"
+      OR "${CMAKE_CUDA_COMPILER_ID}" STREQUAL "NVHPC"
+    )
+      list(APPEND cuda_compile_options "-Xptxas=${ptxas_compile_option}")
```
alliepiper, line 32: We'll need the generator expression that checks for release here, wrapping the `-Xptxas` bits, too. Otherwise this will add a dangling `-Xptxas=` with no arg on non-release builds.

cub/cmake/CubBuildCompilerTargets.cmake
```diff
@@ -10,10 +10,27 @@ function(cub_build_compiler_targets)
 
   thrust_create_target(cub.thrust HOST CPP DEVICE CUDA)
 
+  set(
+    ptxas_compile_options
+    $<$<CONFIG:Release>:"--warn-on-spills">
+    $<$<CONFIG:Release>:"--warn-on-local-memory-usage">
```
alliepiper: Move the genex to the spot below -- also, quote the entire genex, not just the payload, otherwise cmake can get cranky :)

cub/cmake/CubBuildCompilerTargets.cmake
alliepiper: Let's leave it in CUB for now and see how it goes. We can move it to CCCL-wide later if it goes well and we want it other places. I'm not sure if eg. cudax would care about this since it's experimental/sandboxy.

alliepiper (changes requested):
Left some initial comments inline.

Will these warnings be promoted to errors with our current warming promotion flags?

cub/cmake/CubBuildCompilerTargets.cmake
davebayer: I always forget that generator expression is not evaluated immediately.. thanks

ci/build_cub.sh
```diff
@@ -33,6 +34,7 @@ CMAKE_OPTIONS=(
     -DCMAKE_CUDA_STANDARD=$CXX_STANDARD
     -DCCCL_ENABLE_BENCHMARKS=$ENABLE_CCCL_BENCHMARKS
     -DCUB_ENABLE_RDC_TESTS=$ENABLE_CUB_RDC
+    -DCCCL_ENABLE_PTXAS_WARNINGS=$ENABLE_PTXAS_WARNINGS
```
alliepiper: We'd need to make this change in the `ci/windows/` script as well.

It might make more sense to leave it off (by default in the CMake scripts and in the presets) so folks can just turn it on when they want this. That way we won't have to special case this in the CI scripts at all.

alliepiper (changes requested):

alliepiper (approved):

PR #6844, opened 2025-12-02
Avoid use of cccl namespace macros in cub

davebayer:
This PR replaces the uses `_CCCL_(BEGIN|END)_NAMESPACE_MEOW` macros in cub so we needn't to find workarounds for libcu++'s prologue/epilogue includes.

### DIFF
--- a/cub/cub/device/device_transform.cuh
+++ b/cub/cub/device/device_transform.cuh
@@ -39,11 +39,9 @@ struct __return_constant
 } // namespace detail
 CUB_NAMESPACE_END
 
-_CCCL_BEGIN_NAMESPACE_CUDA
 template <typename T>
-struct proclaims_copyable_arguments<CUB_NS_QUALIFIER::detail::__return_constant<T>> : ::cuda::std::true_type
+struct ::cuda::proclaims_copyable_arguments<CUB_NS_QUALIFIER::detail::__return_constant<T>> : ::cuda::std::true_type
 {};
-_CCCL_END_NAMESPACE_CUDA
 
 CUB_NAMESPACE_BEGIN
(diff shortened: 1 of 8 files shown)


cub/cub/device/device_transform.cuh
```diff
@@ -39,11 +39,9 @@ struct __return_constant
 } // namespace detail
 CUB_NAMESPACE_END
 
-_CCCL_BEGIN_NAMESPACE_CUDA
 template <typename T>
-struct proclaims_copyable_arguments<CUB_NS_QUALIFIER::detail::__return_constant<T>> : ::cuda::std::true_type
+struct ::cuda::proclaims_copyable_arguments<CUB_NS_QUALIFIER::detail::__return_constant<T>> : ::cuda::std::true_type
 {};
-_CCCL_END_NAMESPACE_CUDA
```
bernhardmgruber, line 46: I love how we go back and forth between those two version. I did `struct ::cuda::...` initially, but IIRC @fbusato changed that back in later PRs because nvc++ hung itself on the syntax. Let's see what happens now (we updated nvc++ in the meantime I think).

bernhardmgruber (approved):

cub/cub/device/device_transform.cuh
fbusato: excellent memory! yes, nvc++ was problematic. I would love to get rid of these macros

davebayer: hopefully `namespace cuda { struct ... }` will work if the problem is still there

davebayer: Seems to be fixed :)

bernhardmgruber (approved):

PR #6848, opened 2025-12-02
Add an option to use CCCL from CTK for C2H

bernhardmgruber:
This avoids the need to recompile C2H every time we change something in CCCL, which tremendously speeds up the edit->compile->test cycle

The problem I have seen so far is that now C2H needs to work with very old CTKs (e.g. 12.0), which is potentially a lot of work. So maybe it would be better to just add a cmake option that allows us to specify whether CCCL for C2H is taken from the CTK or from the working copy. And then C2H would need to support only somewhat recent CTKs that developers use. This would work for me.

### DIFF
--- a/c2h/include/c2h/custom_type.h
+++ b/c2h/include/c2h/custom_type.h
@@ -154,7 +154,7 @@ public:
 } // namespace c2h
 
 template <template <typename> class... Policies>
-class ::cuda::std::numeric_limits<c2h::custom_type_t<Policies...>>
+class cuda::std::numeric_limits<c2h::custom_type_t<Policies...>>
 {
 public:
   static constexpr bool is_specialized = true;


c2h/CMakeLists.txt
```diff
@@ -41,3 +37,7 @@ endif()
 
 add_library(cccl.c2h.main OBJECT catch2_runner.cpp catch2_runner_helper.cu)
 target_link_libraries(cccl.c2h.main PUBLIC cccl.c2h)
+target_include_directories(
+  cccl.c2h.main
+  PUBLIC ${CUDAToolkit_INCLUDE_DIRS}/cccl
```
miscco: Isnt that what we explicitly tell users not to do ;) @wmaxey 

It should be sufficient to always use just ` ${CUDAToolkit_INCLUDE_DIRS}` because the CTK already packages it correctly

bernhardmgruber: I think the better way would be a `find_package(CCCL ...)` but only allow it to find the one from the CTK. 

`${CUDAToolkit_INCLUDE_DIRS}` is just `/usr/local/cuda/targets/x86_64-linux/include` on my machine. It's set by CMake and we have no control over it.

bernhardmgruber: @alliepiper how can I find CCCL from the CTK? I tried various combinations of:
```
find_package(CUDAToolkit REQUIRED)
find_package(CCCL
        CONFIG
        REQUIRED
        NO_DEFAULT_PATH
        HINTS ${CUDAToolkit_INCLUDE_DIRS})
```
but it keeps finding the CCCL from the current working copy.

alliepiper: It's not really possible AFAIK to bring in different versions of the same targets to the same CMake build.

You should be able to just avoid linking c2h to any CCCL targets and let the default CTK cccl get picked up. We definitely shouldn't include `<CTK include>/cccl`, it sets a bad example and won't do anything on CTK 12.X.

alliepiper: To be more specific, if CCCL is found once, later calls to `find_package(CCCL)` will only refresh the already-found CCCL, regardless of hints.

alliepiper: What do the compile commands for something like c2h's generator.cu look like with `CCCL::CCCL` removed from the `target_link_libraries` for `cccl.c2h`? That should be all that's needed.

bernhardmgruber: Thx for the tips! The only problem I encountered was that we have a single `.cpp` file (not a `.cu` file) which was not having the includes setup correctly. I can also just rename the file extension.

alliepiper: Ahh, gotcha, in that case you're right, we do need to add the include dir for the CXX targets. I'd do something like this:

```cmake
target_include_directories(cccl.c2h.main
  SYSTEM
  PRIVATE "$<COMPILE_LANGUAGE:CXX>:${CUDAToolkit_INCLUDE_DIRS}"
)
```

Don't append `cccl`, that variable is a list of multiple directories, and already includes CCCL if needed (`/usr/local/cuda/targets/x86_64-linux/include;/usr/local/cuda/targets/x86_64-linux/include/cccl` for me locally on 13.X).

`PRIVATE` will keep it contained to the `cccl.c2h.main` target so it won't be transitive downstream, where the CCCL targets should be used.

bernhardmgruber: Interesting. For me (cmake `3.31.6` and CTK 13.0) it only expands to `/usr/local/cuda/targets/x86_64-linux/include`, and fails to find the CCCL headers.

I went with renaming the `cpp` file extension to `cu` now.

miscco (approved):

PR #6850, opened 2025-12-03
CMake backports for 3.2

alliepiper:
Loads of build system / infra changes that the libcudacxx/thrust/cub interdependency update depends on.


bernhardmgruber:
Q: Why do we need the backport? Do we have a bug that those are supposed to fix?

alliepiper:
> Q: Why do we need the backport? Do we have a bug that those are supposed to fix?

The main motivation is that @pciolkosz needs the libcudacxx dependency on CUB for some work that he's backporting, and that is tied up / dependent on a bunch of the other CMake refactorings that have gone in recently.

Another benefit is that this adds the MSVC compat flags to our public packages in 3.2, which has already been reported as a frustration for some users trying out the RCs.

PR #6852, opened 2025-12-03
[libcu++] Add memory_pool header and correct legacy resources namespace

pciolkosz:
This PR adds a new user-facing `cuda/memory_pool` header. It's a result of us putting memory pools into `cuda::` namespace to separate them from the `cuda::mr::` namespace, they should also have a separate header.

This PR also adjusts legacy resources namespace to `cuda::mr::`, because unlike the memory pools, those types are only resources and they should go to `cuda::mr::` namespace.

### DIFF
--- a/libcudacxx/include/cuda/memory_resource
+++ b/libcudacxx/include/cuda/memory_resource
@@ -26,6 +26,9 @@
 //!
 //!@endrst
 
+#include <cuda/__memory_pool/device_memory_pool.h>
+#include <cuda/__memory_pool/managed_memory_pool.h>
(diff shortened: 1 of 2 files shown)


libcudacxx/include/cuda/memory_resource
```diff
@@ -27,13 +27,10 @@
 //!@endrst
 
 #include <cuda/__memory_resource/any_resource.h>
```
miscco, line 32: I still want the pools here

miscco (approved):

PR #6856, opened 2025-12-03
Implement std::chi_squared_distribution

RAMitchell:


libcudacxx/include/cuda/std/__random/chi_squared_distribution.h
```diff
@@ -0,0 +1,179 @@
...
+    result_type __n_ = result_type{1};
+
+  public:
+    using distribution_type = chi_squared_distribution;
+
+    constexpr param_type() = default;
+
+    _CCCL_API constexpr explicit param_type(result_type __n)
+        : __n_{__n}
+    {}
```
fbusato:
```suggestion
    constexpr param_type() noexcept = default;

    _CCCL_API constexpr explicit param_type(result_type __n) noexcept 
        : __n_{__n}
    {}
```

fbusato (changes requested):

RAMitchell:
I've changed all the existing limit headers in this PR.

PR #6861, opened 2025-12-03
Allow numpy struct types as initial value for Zipiterator inputs

shwina:
## Description

When writing up an example for our docs, I realized that we don't gracefully handle when a numpy struct type is used as the initial value corresponding to ZipIterator input.

Specifically, it enables the following:

```python
it = ZipIterator(arr1, arr2)
h_init = np.asarray((0, 0), dtype=np.dtype([('x', np.int32), ('y', np.int32)], align=True))
reduce_into(it, d_out, op, h_init, num_items)
```

Where previously it required one to define a `@gpu_struct`:

```python
@gpu_struct
class MyStruct:
    x: np.int32
    y: np.int32

it = ZipIterator(arr1, arr2)
h_init = MyStruct(0, 0)
reduce_into(it, d_out, op, h_init, num_items)
```

Note: the `gpu_struct` approach still works with this PR.

## Checklist

### DIFF
--- a/python/cuda_cccl/cuda/compute/algorithms/_reduce.py
+++ b/python/cuda_cccl/cuda/compute/algorithms/_reduce.py
@@ -46,7 +46,7 @@ class _Reduce:
         self.h_init_cccl = cccl.to_cccl_value(h_init)
         value_type = get_value_type(h_init)
 
-        # For well-known operations, we don't neeId a signature
+        # For well-known operations, we don't need a signature
         if isinstance(op, OpKind):
             self.op_wrapper = cccl.to_cccl_op(op, None)
         else:


python/cuda_cccl/cuda/compute/algorithms/_reduce.py
```diff
@@ -40,12 +44,9 @@ def __init__(
...
         self.d_out_cccl = cccl.to_cccl_output_iter(d_out)
         self.h_init_cccl = cccl.to_cccl_value(h_init)
-        if isinstance(h_init, np.ndarray):
-            value_type = numba.from_dtype(h_init.dtype)
-        else:
-            value_type = numba.typeof(h_init)
+        value_type = get_value_type(h_init)
 
-        # For well-known operations, we don't need a signature
+        # For well-known operations, we don't neeId a signature
```
NaderAlAwar: typo

NaderAlAwar (approved):

PR #6868, opened 2025-12-03
In test_device_segmented_scan_api change type from int to unsigned

oleksandr-pavlyk:
This works-around an issue where scan operator may be invoked on uninitialized value by the algorithm on certain architectures, but the operator value is unused.

Some of these unitialized values trip assertions embedded in fast_mod_div implementation, tripping assertion failures in test runs.

This PR replaces use of signed integral input type with unsigned integral type instead, as all uninitialized values are valid in the eye of fast_mod_div algorithm.

## Description

closes 

## Checklist


bernhardmgruber (approved):

bernhardmgruber:
Thx for the quick fix!

The problem was that because `BlockScan` calls the scan operator on garbage values, if we are calling `fast_mod_div` on garbage `int`s, we may assert on negative values. If we pick `unsigned`, any garbage value is within the valid range of `fast_mod_div`.

PR #6879, opened 2025-12-04
Implement std::negative_binomial_distribution

RAMitchell:


libcudacxx/include/cuda/std/__random/negative_binomial_distribution.h
```diff
@@ -0,0 +1,212 @@
...
+#  pragma clang system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_MSVC)
+#  pragma system_header
+#endif // no system header
+
+#include <cuda/std/__random/bernoulli_distribution.h>
+#include <cuda/std/__random/gamma_distribution.h>
+#include <cuda/std/__random/is_valid.h>
+#include <cuda/std/__random/poisson_distribution.h>
+#include <cuda/std/limits>
```
fbusato: here we are using integer types only. I think we can restrict the header inclusion

libcudacxx/include/cuda/std/__random/negative_binomial_distribution.h
```diff
@@ -0,0 +1,212 @@
...
+  [[nodiscard]] _CCCL_API result_type operator()(_URng& __g)
+  {
+    return (*this)(__g, __p_);
+  }
+  template <class _URng>
+  [[nodiscard]] _CCCL_API result_type operator()(_URng& __urng, const param_type& __pr)
+  {
+    static_assert(__cccl_random_is_valid_urng<_URng>, "URng must meet the UniformRandomBitGenerator requirements");
+    result_type __k = __pr.k();
+    double __p      = __pr.p();
```
fbusato:
```suggestion
    const result_type __k = __pr.k();
    const double __p      = __pr.p();
```

fbusato (approved):

PR #6880, opened 2025-12-04
Add missing nvrtc nv target archs

davebayer:
> I found a way to force nvrtc to include the version of `<nv/target>` that comes with our CCCL.

> The idea is that we pass the `-D__NV_TARGET_H` option to nvrtc through commandline which eliminates makes the builtin `<nv/target>` header to not be included. Then I `#undef __NV_TARGET_H` which allows our version of `<nv/target>` to be included.~~

> I'd like to know your opinion on this. We could handle everything inside `nvrtcc` - both the define and undefine.

> Alternatively, we could change the implementation of `<nv/target>` to not be included if there is a CCCL available in the include path and the version is newer than what nvrtc has. But that would require much larger changes.

> It would look something like:
```cpp
// <nv/target>

// If compiling with nvrtc and there is another <nv/target> in the include path, we skip the builtin header and use the
// other <nv/target> instead.
#if !defined(__NV_TARGET_H) && !defined(_NV_TARGET_SKIPPING_NVRTC_BUILTIN_HEADER)
#  if defined(__CUDACC_RTC__) && __has_include(<cuda/__cccl_config>) && __has_include(<nv/target>)
#    include <cuda/__cccl_config>
#    if (CCCL_MAJOR_VERSION > (__CUDACC_VER_MAJOR__  - 10)) || (CCCL_MAJOR_VERSION == (__CUDACC_VER_MAJOR__  - 10) && CCCL_MINOR_VERSION > __CUDACC_VER_MINOR__)
#      define _NV_TARGET_SKIPPING_NVRTC_BUILTIN_HEADER
#      include <nv/target>
#      undef _NV_TARGET_SKIPPING_NVRTC_BUILTIN_HEADER
#    endif // ^^^ use <nv/target> include ^^^
#  endif // 
#endif // !__NV_TARGET_H && !_NV_TARGET_SKIPPING_NVRTC_BUILTIN_HEADER

#ifndef __NV_TARGET_H
#define __NV_TARGET_H

// ...

#endif // __NV_TARGET_H
```

> Edit: implemented this in #6881

Edit 2: Just adding missing nv target macros for archs 103, 110 and 120.


davebayer:
Hm, this is probably not safe, because if nvrtc expects the nv target macros in another builtin headers, the compilation will fail

bernhardmgruber:
I guess for tests, this is fine. Please also add a comment to each
```
// ADDITIONAL_COMPILE_DEFINITIONS: __NV_TARGET_H
```
explaining how this disables nvrtc's builtin header.

davebayer:
With @wmaxey we are working on #6881, so we needn't to deal with this problem ever again. I changed this PR to just add the missing architectures. Hopefully we do this for the very last time.

@alliepiper this should fix our libcu++ nightlies.

davebayer:
> I hate that we have to do this. LGTM.

This fix is on the way, but I believe we will have to keep doing this till CUDA 15 :D

PR #6881, opened 2025-12-04
Use newer `<nv/target>` with nvrtc if available

davebayer:
This PR adds prologue to `<nv/target>` that is intended to be used during compilation with nvrtc when the header is builtin. The idea is to check whether there is a newer version of the header available in the include path. If there is, we use that one instead of the builtin one


nv_target_macros/a/nv/target
```diff
@@ -0,0 +1,67 @@
+
+// Identifies this header as versioned nv target header.
+#define __NV_TARGET_VERSIONED
+
+// This guard has 2 purposes - preventing multiple inclusions and allowing to detect whether the header is versioned or
+// or not without including the contents.
+#if !defined(__NV_TARGET_H)
+
+// Hardcoded version. We cannot use something like __NV_TARGET_VERSION_CURRENT, because there would be collisions.
+#  define __NV_TARGET_VERSION_03_01 301
```
miscco: This version is different than the one in the other file is that intentional?

PR #6884, opened 2025-12-04
[cuda.compute]: Add fast path to extract PyTorch array pointer

NaderAlAwar:
## Description

Should be merged after #6882

## Checklist


NaderAlAwar:
>I wonder by how much does this slow down non-Torch containers?

From my local benchmarks on reduce (three arrays) it seems like it is only 1-2us. The corresponding speedup for pytorch is 5-7us.

Issue #5970, opened 2025-09-22
[BUG]: CCCL 3.1.x breaks when building with CUDA 12.9 and running on CUDA 12.0/12.2


bdice:
### Is this a duplicate?

- [x] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cccl/issues) for this bug and that I agree to the [Code of Conduct](CODE_OF_CONDUCT.md)

### Type of Bug

Runtime Error

### Component

libcu++

### Describe the bug

I'm seeing a CUDA build/runtime version issue with CCCL 3.1.0-rc6.

`undefined symbol: cudaGetDriverEntryPointByVersion, version libcudart.so.12`

I'm building cuDF with CUDA 12.9 and then running tests with CUDA 12.0 / 12.2. Both of these configurations fail. It seems like [this macro](https://github.com/NVIDIA/cccl/blob/9d7333b964f61c0de20042bb44cfa2b6486e4e16/libcudacxx/include/cuda/__driver/driver_api.h#L53) is guarding for build-time CUDA > 12.5 we also need to have CUDA runtime >= 12.5 in order to have this API available.

The `cudaGetDriverEntryPointByVersion` API was added in CUDA 12.5:
https://docs.nvidia.com/cuda/archive/12.5.0/cuda-runtime-api/group__CUDART__DRIVER__ENTRY__POINT.html#group__CUDART__DRIVER__ENTRY__POINT_1gda6c90a43afbe8043128285afbaf8c04

### How to Reproduce

I'm not sure what is calling this code path yet.

### Expected behavior

Building with CUDA 12.9 and running with CUDA 12.0/12.2 should work without runtime errors.

bdice:
Fixed by https://github.com/NVIDIA/cccl/pull/5976.

#5987 tracks testing this kind of compatibility explicitly for the future.

PR #6895, opened 2025-12-05
[libcu++] Fix minor version compatibility in 13.X

pciolkosz:
`cudaGetDriverEntryPoint` will always request the version of a function that matches the current CUDA Runtime version. This is not minor version compatibility friendly and older driver will just respond with invalid value error.

We made a switch to use `cudaGetDriverEntryPoint` in order to fix https://github.com/NVIDIA/cccl/issues/5970, but that broke minor version compatibility. This PR brings back `cudaGetDriverEntryPointByVersion` at least for CUDA 13.X CTK.

There will probably need to be a follow-up change to instead switch to dlopen + dlsym in order to fix this in 12.X

### DIFF
--- a/libcudacxx/include/cuda/__driver/driver_api.h
+++ b/libcudacxx/include/cuda/__driver/driver_api.h
@@ -51,7 +51,12 @@ _CCCL_SUPPRESS_DEPRECATED_PUSH
   // TODO switch to dlopen of libcuda.so instead of the below
   void* __fn;
   ::cudaDriverEntryPointQueryResult __result;
+#  if _CCCL_CTK_AT_LEAST(13, 0)
+  ::cudaError_t __status =
+    ::cudaGetDriverEntryPointByVersion("cuGetProcAddress", &__fn, 13000, ::cudaEnableDefault, &__result);
+#  else
   ::cudaError_t __status = ::cudaGetDriverEntryPoint("cuGetProcAddress", &__fn, ::cudaEnableDefault, &__result);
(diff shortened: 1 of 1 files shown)


libcudacxx/include/cuda/__driver/driver_api.h
```diff
@@ -51,7 +51,12 @@ _CCCL_SUPPRESS_DEPRECATED_PUSH
   // TODO switch to dlopen of libcuda.so instead of the below
   void* __fn;
   ::cudaDriverEntryPointQueryResult __result;
+#  if _CCCL_CTK_AT_LEAST(13, 0)
+  ::cudaError_t __status =
+    ::cudaGetDriverEntryPointByVersion("cuGetProcAddress", &__fn, 13000, ::cudaEnableDefault, &__result);
```
miscco, line 56: Should this ask for the actual driver version instead of 1300

leofang: @pciolkosz i have the same question

pciolkosz: Do you mean if we should check the driver version and always ask for the version matching it? It doesn't matter here, its asking for the version of `cuGetProcAddress` from CUDA 13.0, so on all drivers we will get the same thing, even if in some later CUDA version it was changed. Passing the exact current driver version here would mean we always get the latest version of the API, which we don't want, because if it ever changes we would need to handle a new signature of the function

PR #6897, opened 2025-12-05
Consolidate and fix RDC toggles.

alliepiper:
Removes per-project helpers that toggle RDC state and replaces them with a CCCL-level utility.

Removes sm87 from targets with RDC enabled.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -19,6 +19,22 @@ The repository uses **CMake** with the **Ninja** generator and provides standard
 
 ---
 
+## Iteration Cycles
+
+For a given task, you should:
+
+1. Research. Search the web, read existing code, look up system/dependency headers / implementations of related functionality. Figure out best practices and common pitfalls. Look for existing tests of the functionality; if none exist, plan a new test that integrates with the relevant existing testing frameworks.
+2. Plan. Create a high-level plan to implement the requested feature.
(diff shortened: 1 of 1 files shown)


bernhardmgruber (approved):

cmake/CCCLUtilities.cmake
```diff
@@ -13,6 +13,38 @@
...
 
+# Enable or disable relocatable device code (RDC) for a given target.
+function(cccl_set_rdc target_name rdc_enabled)
+  if (rdc_enabled)
+    set_target_properties(
+      ${target_name}
+      PROPERTIES #
+        CUDA_SEPARABLE_COMPILATION ON
+        CUDA_RESOLVE_DEVICE_SYMBOLS ON
+        POSITION_INDEPENDENT_CODE ON
```
davebayer, line 24: Q: Should really `POSITION_INDEPENDENT_CODE` be turned on? It influences only host code, so `-Xcompiler=-fPIC` is passed to the host compiler when compiling with `nvcc`. If this should force the `ptxas` to generate position independent code, we need to pass `-Xptxas -pic`

alliepiper: I'm not sure why that is there tbh. I left it in place because it shouldn't hurt anything and I don't want to risk breaking anything unexpectedly. We can look into removing it in a future PR, but this PR is blocking some other work so I'd like to not add scope here.

davebayer: Opened #7237 for that

PR #6899, opened 2025-12-05
[libcu++] Dynamically load CUDA library instead of using the runtime

pciolkosz:
We can't support minor version compatibility in 12.X releases without breaking dynamic runtime use case described here: https://github.com/NVIDIA/cccl/issues/5970. Its because in older 12.X releases there is no `cudaGetDriverEntryPointByVersion` and the non versioned one can't support MVC.
This PR switches to instead dynamically load the CUDA library and fetch `cuGetProcAddress` from it, instead of using `cudaGetDriverEntryPoint`
On the windows side it uses `LibraryLoad/GetProcAddress`, which requires `windows.h` include. It's a pretty heavy header, but should also be commonly included anyway. Long term we can think about an alternative

### DIFF
--- a/libcudacxx/include/cuda/__driver/driver_api.h
+++ b/libcudacxx/include/cuda/__driver/driver_api.h
@@ -53,16 +53,17 @@ _CCCL_SUPPRESS_DEPRECATED_PUSH
 //! @brief Gets the cuGetProcAddress function pointer.
 [[nodiscard]] _CCCL_PUBLIC_HOST_API inline auto __getProcAddressFn() -> decltype(cuGetProcAddress)*
 {
+  const char* __fn_name = "cuGetProcAddress_v2";
 #  if _CCCL_OS(WINDOWS)
   static auto __driver_library = ::LoadLibraryExA("nvcuda.dll", nullptr, LOAD_LIBRARY_SEARCH_SYSTEM32);
   if (__driver_library == nullptr)
   {
     ::cuda::__throw_cuda_error(::cudaErrorUnknown, "Failed to load nvcuda.dll");
   }
-  static void* __fn = ::GetProcAddress(__driver_library, "cuGetProcAddress_v2");
+  static void* __fn = ::GetProcAddress(__driver_library, __fn_name);
   if (__fn == nullptr)
   {
-    ::cuda::__throw_cuda_error(::cudaErrorUnknown, "Failed to get cuGetProcAddress_v2 from nvcuda.dll");
+    ::cuda::__throw_cuda_error(::cudaErrorInitializationError, "Failed to get cuGetProcAddress from nvcuda.dll");
   }
 #  else // ^^^ _CCCL_OS(WINDOWS) ^^^ / vvv !_CCCL_OS(WINDOWS) vvv
 #    if _CCCL_OS(ANDROID)
@@ -75,10 +76,10 @@ _CCCL_SUPPRESS_DEPRECATED_PUSH
   {
     ::cuda::__throw_cuda_error(::cudaErrorUnknown, "Failed to load libcuda.so.1");
   }
-  static void* __fn = ::dlsym(__driver_library, "cuGetProcAddress_v2");
+  static void* __fn = ::dlsym(__driver_library, __fn_name);
   if (__fn == nullptr)
   {
-    ::cuda::__throw_cuda_error(::cudaErrorUnknown, "Failed to get cuGetProcAddress_v2 from libcuda.so.1");
+    ::cuda::__throw_cuda_error(::cudaErrorInitializationError, "Failed to get cuGetProcAddress from libcuda.so.1");
   }
 #  endif // ^^^ !_CCCL_OS(WINDOWS) ^^^
   return reinterpret_cast<decltype(cuGetProcAddress)*>(__fn);


libcudacxx/include/cuda/__driver/driver_api.h
```diff
@@ -48,19 +53,34 @@ _CCCL_SUPPRESS_DEPRECATED_PUSH
...
 {
-  // TODO switch to dlopen of libcuda.so instead of the below
-  void* __fn;
-  ::cudaDriverEntryPointQueryResult __result;
-#  if _CCCL_CTK_AT_LEAST(13, 0)
-  ::cudaError_t __status =
-    ::cudaGetDriverEntryPointByVersion("cuGetProcAddress", &__fn, 13000, ::cudaEnableDefault, &__result);
+#  if _CCCL_OS(WINDOWS)
+  HMODULE m_cudaDriverLibrary = LoadLibraryEx("nvcuda.dll", nullptr, LOAD_LIBRARY_SEARCH_SYSTEM32);
+  if (m_cudaDriverLibrary == nullptr)
```
miscco: Critical: This is loading every single time the function is called, should this be a function local static?

libcudacxx/include/cuda/__driver/driver_api.h
```diff
@@ -48,19 +53,34 @@ _CCCL_SUPPRESS_DEPRECATED_PUSH
...
 #  else
-  ::cudaError_t __status = ::cudaGetDriverEntryPoint("cuGetProcAddress", &__fn, ::cudaEnableDefault, &__result);
-#  endif
-  if (__status != ::cudaSuccess || __result != ::cudaDriverEntryPointSuccess)
+#    if _CCCL_OS(ANDROID)
+  const char* m_cudaDriverLibraryName = "libcuda.so";
+#    else
+  const char* m_cudaDriverLibraryName = "libcuda.so.1";
+#    endif
+  void* m_cudaDriverLibrary = dlopen(m_cudaDriverLibraryName, RTLD_NOW);
```
miscco: ditto should this be function local static?

libcudacxx/include/cuda/__driver/driver_api.h
pciolkosz: We only call this function once to initialize a function local static `__get_proc_addr_fn` in `__get_driver_entry_point`. I don't think we will ever have to call this function more than once, so I don't think there is a reason to store the loaded library handle. But its also not expensive, so I would be fine to store it as well if there is an argument to do so

miscco: I dont think we need to store it, just make it
```suggestion
#  if _CCCL_OS(WINDOWS)
  static auto m_cudaDriverLibrary = ::LoadLibraryExA("nvcuda.dll", nullptr, LOAD_LIBRARY_SEARCH_SYSTEM32);
  if (m_cudaDriverLibrary == nullptr)
```

davebayer: Shouldn't this function also be `_CCCL_PUBLIC_HOST_API` so we really query it only once across TUs?

libcudacxx/include/cuda/__driver/driver_api.h
```diff
@@ -48,19 +53,34 @@ _CCCL_SUPPRESS_DEPRECATED_PUSH
...
 #  else
-  ::cudaError_t __status = ::cudaGetDriverEntryPoint("cuGetProcAddress", &__fn, ::cudaEnableDefault, &__result);
-#  endif
-  if (__status != ::cudaSuccess || __result != ::cudaDriverEntryPointSuccess)
+#    if _CCCL_OS(ANDROID)
+  const char* m_cudaDriverLibraryName = "libcuda.so";
+#    else
+  const char* m_cudaDriverLibraryName = "libcuda.so.1";
+#    endif
+  void* m_cudaDriverLibrary = ::dlopen(m_cudaDriverLibraryName, RTLD_NOW);
```
davebayer: Please, use `__my_var` style

libcudacxx/include/cuda/__driver/driver_api.h
```diff
@@ -46,21 +51,36 @@ _CCCL_BEGIN_NAMESPACE_CUDA_DRIVER
...
-  ::cudaError_t __status = ::cudaGetDriverEntryPoint("cuGetProcAddress", &__fn, ::cudaEnableDefault, &__result);
-#  endif
-  if (__status != ::cudaSuccess || __result != ::cudaDriverEntryPointSuccess)
+#  if _CCCL_OS(WINDOWS)
+  static auto __driver_library = ::LoadLibraryExA("nvcuda.dll", nullptr, LOAD_LIBRARY_SEARCH_SYSTEM32);
+  if (__driver_library == nullptr)
+  {
+    ::cuda::__throw_cuda_error(::cudaErrorUnknown, "Failed to load nvcuda.dll");
+  }
+  static void* __fn = ::GetProcAddress(__driver_library, "cuGetProcAddress_v2");
```
davebayer: Can we have a separate variable for the symbol name? It's used twice and I can see us changing it only in one place in the future..

libcudacxx/include/cuda/__driver/driver_api.h
```diff
@@ -46,21 +51,36 @@ _CCCL_BEGIN_NAMESPACE_CUDA_DRIVER
...
+#  else // ^^^ _CCCL_OS(WINDOWS) ^^^ / vvv !_CCCL_OS(WINDOWS) vvv
+#    if _CCCL_OS(ANDROID)
+  const char* __driver_library_name = "libcuda.so";
+#    else // ^^^ _CCCL_OS(ANDROID) ^^^ / vvv !_CCCL_OS(ANDROID) vvv
+  const char* __driver_library_name = "libcuda.so.1";
+#    endif // ^^^ !_CCCL_OS(ANDROID) ^^^
+  static void* __driver_library = ::dlopen(__driver_library_name, RTLD_NOW);
+  if (__driver_library == nullptr)
+  {
+    ::cuda::__throw_cuda_error(::cudaErrorUnknown, "Failed to load libcuda.so.1");
```
davebayer, line 77: Shouldn't we throw `cudaErrorInitializationError`?

davebayer (approved):

libcudacxx/include/cuda/__driver/driver_api.h
jrhemstad: Make sure we aren't depending on function local statics being shared across DLL boundaries 

https://stackoverflow.com/questions/62336104/are-function-local-static-objects-of-inline-functions-shared-across-shared-objec

pciolkosz: I think we should be good, for any function local static we use (in this PR or in other places), I think it would only be a resource consumption / performance difference if they are not shared, but not a correctness issue

PR #6900, opened 2025-12-06
[libcu++] Add as_ref() to memory pool types

pciolkosz:
The owning memory pool types are not copyable (for obvious reasons), but we require the resource to be copyable in `any_resource` and `buffer`. User needs to either pack the pool into `shared_resource` or pass it as a memory pool reference.

This PR adds a convenient API `as_ref()` to the owning memory pool types to opt-out of pool ownership when it's passed to an API. It's basically a `static_cast<X_memory_pool_ref>(this)`.

I also added a test for constructing the buffer with `shared_resource` to show the two different approaches to custom memory pools

### DIFF
--- a/cmake/CCCLBuildCompilerTargets.cmake
+++ b/cmake/CCCLBuildCompilerTargets.cmake
@@ -12,6 +12,7 @@ set(CMAKE_MSVC_DEBUG_INFORMATION_FORMAT Embedded)
 option(CCCL_ENABLE_EXCEPTIONS "Enable exceptions within CCCL libraries." ON)
 option(CCCL_ENABLE_RTTI "Enable RTTI within CCCL libraries." ON)
 option(CCCL_ENABLE_WERROR "Treat warnings as errors for CCCL targets." ON)
+option(CCCL_ENABLE_PTXAS_WARNINGS "Enable ptxas warnings" OFF) # currently used only in CUB
 
 function(
   cccl_build_compiler_interface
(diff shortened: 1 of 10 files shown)


libcudacxx/include/cuda/__memory_pool/managed_memory_pool.h
```diff
@@ -130,6 +130,13 @@ struct managed_memory_pool : managed_memory_pool_ref
     return managed_memory_pool(__pool);
   }
 
+  //! @brief Returns a \c managed_memory_pool_ref for this \c managed_memory_pool.
+  //! The result is the same as if this object was cast to a \c managed_memory_pool_ref.
+  _CCCL_HOST_API managed_memory_pool_ref as_ref() noexcept
```
miscco: Should be
```suggestion
  [[nodiscard]] _CCCL_HOST_API managed_memory_pool_ref as_ref() noexcept
```

libcudacxx/include/cuda/__memory_pool/device_memory_pool.h
```diff
@@ -139,6 +139,13 @@ struct device_memory_pool : device_memory_pool_ref
     return device_memory_pool(__pool);
   }
 
+  //! @brief Returns a \c device_memory_pool_ref for this \c device_memory_pool.
+  //! The result is the same as if this object was cast to a \c device_memory_pool_ref.
+  _CCCL_HOST_API device_memory_pool_ref as_ref() noexcept
```
miscco:
```suggestion
  [[nodiscard]] _CCCL_HOST_API device_memory_pool_ref as_ref() noexcept
```

PR #6902, opened 2025-12-07
Allow `if consteval` in device code with nvcc 13.1

davebayer:
`if consteval` in C++20 mode seems to finally work with nvcc 13.1.

Requires #6887.


bernhardmgruber:
> `if consteval` in C++20 mode seems to finally work with nvcc 13.1.

I thought this is a C++23 feature. Was it DR-ed against C++20?

davebayer:
> I thought this is a C++23 feature. Was it DR-ed against C++20?

It is and it was not DR-ed against C++20 as far as I know. But most compilers support it even in C++20 as an extension. NVCC accepts `if consteval` since 13.0, but emitted an error when used in device code. This seems to be fixed in 13.1.

libcudacxx/include/cuda/std/__cccl/dialect.h
```diff
@@ -141,8 +141,10 @@
 #  define _CCCL_HAS_IF_CONSTEVAL_IN_CXX20() 0
 #endif
 
-#if _CCCL_CUDA_COMPILER(NVCC) \
-  && (_CCCL_CUDA_COMPILER(NVCC, <, 13) || _CCCL_DEVICE_COMPILATION() || _CCCL_COMPILER(CLANG))
+// nvcc before 13 doesn't support if consteval at all. Since 13, it accepts if consteval in host code (clang doesn't
+// work) and since 13.1 it works in device code, too.
+#if _CCCL_CUDA_COMPILER(NVCC, <, 13) || (_CCCL_CUDA_COMPILER(NVCC, <, 13, 1) && _CCCL_DEVICE_COMPILATION()) \
```
fbusato, line 146: why we cannot merge both conditions?

fbusato: Never mind. I missed the initial parenthesis.

bernhardmgruber (approved):

PR #6904, opened 2025-12-08
Make sure we actually use overflow builtins

davebayer:
We should be careful about removing the `__` prefix when detecting builtins by `_CCCL_CHECK_BUILTIN(...)`..

I've also found some bugs in 13.1 `__builtin_add_overflow` and `__builtin_sub_overflow` implementation (nvbug 5724197).

### DIFF
--- a/libcudacxx/include/cuda/__numeric/add_overflow.h
+++ b/libcudacxx/include/cuda/__numeric/add_overflow.h
@@ -48,6 +48,12 @@
 #  define _CCCL_BUILTIN_ADD_OVERFLOW(...) __builtin_add_overflow(__VA_ARGS__)
 #endif // _CCCL_CHECK_BUILTIN(builtin_add_overflow)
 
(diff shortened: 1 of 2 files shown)


libcudacxx/include/cuda/__numeric/add_overflow.h
```diff
@@ -44,9 +44,9 @@
 
 #include <cuda/std/__cccl/prologue.h>
 
-#if _CCCL_CHECK_BUILTIN(__builtin_add_overflow) || _CCCL_COMPILER(GCC)
+#if _CCCL_CHECK_BUILTIN(builtin_add_overflow) || _CCCL_COMPILER(GCC)
 #  define _CCCL_BUILTIN_ADD_OVERFLOW(...) __builtin_add_overflow(__VA_ARGS__)
-#endif // _CCCL_CHECK_BUILTIN(__builtin_add_overflow)
+#endif // _CCCL_CHECK_BUILTIN(builtin_add_overflow)
```
miscco, line 49: 🙀

PR #6909, opened 2025-12-08
[cuda.compute] Add dependency on nvidia-nvvm

shwina:
## Description

Adding this dependency explicitly should help ensure that `nvidia-nvvm` has the same minor version as all the other dependencies.

## Checklist


shwina:
/ok to test https://github.com/NVIDIA/cccl/pull/6909/commits/cb2ba1a367283418456022372946afc37f936cc3

PR #6910, opened 2025-12-08
Add missing OpKind docs entries

ktaletsk:
## Summary
- Document the missing `cuda.compute.op.OpKind` enum members (`IDENTITY`, `MINIMUM`, `MAXIMUM`) in the Python API reference.

## Context
Found during the CUDA Python Kernel Authoring tutorial at PyData Boston 2025: `OpKind.MINIMUM` worked in code but was absent from the docs.

image

CC @kriehl

## Testing
- Doc-only change; no tests run.

### DIFF
--- a/python/cuda_cccl/pyproject.toml
+++ b/python/cuda_cccl/pyproject.toml
@@ -46,7 +46,7 @@ cu12 = [
 ]
 cu13 = [
   "cuda-bindings>=13.0.0,<14.0.0",
-  "cuda-toolkit[nvrtc,nvjitlink,cudart,nvcc]==13.*",
+  "cuda-toolkit[nvrtc,nvjitlink,cudart,nvcc,nvvm]==13.*",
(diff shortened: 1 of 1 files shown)


shwina:
Thanks for the PR @ktaletsk!

At the time of writing, I found that using `OpKind.MINIMUM/MAXIMUM` with e.g., `reduce_into` was giving me build errors. However, testing it now, it seems to work.

I will run tests on this PR after merging and rebasing against https://github.com/NVIDIA/cccl/pull/6909, which unblocks CI.

Thank you again!

shwina:
I also see we already have a test for this, so clearly this has been working for a while now :)

https://github.com/NVIDIA/cccl/blob/aafe8b6274ca087d9b6db88c60fe030602aa6e10/python/cuda_cccl/tests/compute/test_reduce.py#L628

PR #6912, opened 2025-12-08
Smoke test for exclusive_scan

viralbhadeshiya:
## Description
- Include basic smoke tests for exclusive_scan

See 

## Checklist

### DIFF
--- a/test/stdpar/tests/exclusive_scan.cpp
+++ b/test/stdpar/tests/exclusive_scan.cpp
@@ -43,17 +43,11 @@ int main()
 
   // exclusive_scan with initial value & custom binary ops
   const int init2 = 7;
-  std::exclusive_scan(
-    std::execution::par,
-    in.begin(),
-    in.end(),
-    out.begin(),
-    init2,
-    [](int lhs, int rhs) {
-      return lhs + rhs + 1;
-    });
+  std::exclusive_scan(std::execution::par, in.begin(), in.end(), out.begin(), init2, [](int lhs, int rhs) {
+    return lhs + rhs + 1;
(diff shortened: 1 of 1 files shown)


viralbhadeshiya:
@miscco this is the PR for exclusive_scan algorithm :)

test/stdpar/tests/exclusive_scan.cpp
```diff
@@ -0,0 +1,46 @@
...
+  std::vector<int> in(N);
+  std::vector<int> out(N);
+
+  for (std::size_t i = 0; i < N; ++i)
+  {
+    in[i] = static_cast<int>(i + 1);
+  }
+
+  // exclusive_scan with default initial value (0)
+  std::exclusive_scan(std::execution::par, in.begin(), in.end(), out.begin(), 0);
```
miscco: We should replace this test by one that provides an init value and a suitable binary operator

test/stdpar/tests/exclusive_scan.cpp
```diff
@@ -0,0 +1,46 @@
...
+{
+  constexpr std::size_t N = 1 << 16;
+
+  std::vector<int> in(N);
+  std::vector<int> out(N);
+
+  for (std::size_t i = 0; i < N; ++i)
+  {
+    in[i] = static_cast<int>(i + 1);
+  }
```
miscco, line 28: This should just be
```suggestion
  std::iota(in.begin(), in.end(), 1);
```

viralbhadeshiya:
@miscco check the new once, I added some test cases for binary ops.

Issue #6486, opened 2025-11-04
[FEA]: Improve NVHPC smoke tests


miscco:
### Is this a duplicate?

- [x] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cccl/issues) for this request and that I agree to the [Code of Conduct](CODE_OF_CONDUCT.md)

### Area

libcu++

### Is your feature request related to a problem? Please describe.

NVHPC relies on our implementation for their `stdpar` backend.

We have *one* smoke test in https://github.com/NVIDIA/cccl/tree/main/test that serves as the sole backstop before we are breaking their internal CI

We want to improve the coverage spanning more algorithms. Note that compiling the tests is sufficient for now.

### Describe the solution you'd like

Implement more small test cases that verify that we do not break NVHPC

- [x] adjacent_difference
- [ ] exclusive_scan
- [ ] inclusive_scan
- [ ] transform_exclusive_scan
- [ ] transform_inclusive_scan
- [ ] transform_reduce
- [x] all_of
- [ ] any_of
- [ ] count
- [ ] copy
- [ ] copy_if
- [ ] copy_n
- [ ] count_if
- [ ] equal
- [ ] fill
- [ ] fill_n
- [ ] find
- [ ] find_if
- [ ] find_if_not
- [ ] for_each
- [ ] for_each_n
- [ ] generate
- [ ] generate_n
- [ ] is_partitioned
- [ ] is_sorted
- [ ] is_sorted_until
- [ ] max_element
- [ ] min_element
- [ ] minmax_element
- [ ] merge
- [ ] mismatch
- [ ] none_of
- [ ] partition
- [ ] partition_copy
- [ ] remove
- [ ] remove_copy
- [ ] remove_copy_if
- [ ] remove_if
- [ ] replace
- [ ] replace_copy
- [ ] replace_copy_if
- [ ] replace_if
- [ ] reverse
- [ ] reverse_copy
- [ ] rotate_copy
- [ ] set_difference
- [ ] set_intersection
- [ ] set_symmetric_difference
- [ ] set_union
- [ ] sort
- [ ] stable_partition
- [ ] stable_sort
- [ ] swap_ranges
- [ ] transform
- [ ] unique
- [ ] unique_copy

### Additional context

The reduce test lives [here](https://github.com/NVIDIA/cccl/blob/main/test/stdpar/tests/reduce.cpp)

We should take this test as a sample and expand it to more accelerated algorithms

[...]

PR #6913, opened 2025-12-08
inclusive_scan smoke test

viralbhadeshiya:
## Description
- Added smoke test for inclusive_scan algorithm

https://github.com/NVIDIA/cccl/issues/6486

## Checklist


viralbhadeshiya:
@miscco this PR contains smoke test for inclusive_scan :)

test/stdpar/tests/inclusive_scan.cpp
```diff
@@ -0,0 +1,44 @@
...
+{
+  constexpr std::size_t N = 1 << 16;
+
+  std::vector<int> in(N);
+  std::vector<int> out(N);
+
+  for (std::size_t i = 0; i < N; ++i)
+  {
+    in[i] = 1;
+  }
```
miscco: Should be `iota`

test/stdpar/tests/inclusive_scan.cpp
```diff
@@ -0,0 +1,44 @@
...
+    const int expected = static_cast<int>(i + 1);
+    if (out[i] != expected)
+    {
+      return 1;
+    }
+  }
+
+  // inclusive_scan with non-zero initial value
+  const int init = 42;
+  std::inclusive_scan(std::execution::par, in.begin(), in.end(), out.begin(), init);
```
miscco, line 29: Does this compile? This is missing the binary operator

We also need tests for only a binary operator and no init

viralbhadeshiya: Yes, I think so because default operator is + so even if there is no binary ops specified it will perform addition by default.

viralbhadeshiya:
@miscco added a test case for only binary ops and no init value. lmk if need any extra cases for inclusive_scan.

viralbhadeshiya:
@miscco can we run test for this and exclusive_scan PR?

test/stdpar/tests/inclusive_scan.cpp
```diff
@@ -0,0 +1,55 @@
...
+  std::vector<int> out(N);
+
+  std::iota(in.start(), in.end(), 1);
+
+  // inclusive_scan with default initial value
+  std::inclusive_scan(std::execution::par, in.begin(), in.end(), out.begin());
+
+  for (std::size_t i = 0; i < N; ++i)
+  {
+    const int expected = static_cast<int>(i + 1);
```
miscco, line 20: I dont believe this is the right expected value, did you run the code locally?

viralbhadeshiya: I did before the changes, but let me run it once more. After I finish all changes.

Issue #5057, opened 2025-06-24
Make `cub::DeviceTransform` tunable


bernhardmgruber:
After the following changes land

- [x] #2394 
- [x] #4976
- [x]  my internal rewrite of memcpy_async

we should make sure `cub::DeviceTransform` can be tuned.

Should include solving:

- [x] #3017

bernhardmgruber:
#6914 probably solves this, but we should verify it.

Issue #3017, opened 2024-12-03
Expose `cub::DeviceTransform` BIF to tuning


bernhardmgruber:
We currently rely on @ahendriksen to provide us with magical values for the target bytes-in-flight for various architectures in `cub::DeviceTransform`. Since the tuning space for `cub::DeviceTransform` is rather small, we could consider exposing this parameter to the CUB tuning framework to confirm that those values are indeed the best ones.

PR #6914, opened 2025-12-08
Implement the new tuning API for `DeviceTransform`

bernhardmgruber:
Fixes: #6919
Fixes: #5057 
Fixes: #3017

- [x] CUB transform tests pass
- [x]  CCCL.C transform tests pass
- [x] Rebase on top of #6916
- [x] Update benchmarks
- [x] Rename anything `arch_policy` related to `policy_selector`
- [x] Merge before: #7103
- [x] No SASS difference for `cub.bench.transform.babelstream.base` on SM `75;80;90;100;120`
- [x] No relevant SASS difference for `cub.test.device.transform.lid_0` on SM `75;80;90;100;120`
- [x]  Compile-time comparison between before and after this PR (including clang in CUDA, because we don't have `__CUDA_ARCH_LIST__`.)
- [x] Resolve decision on changing `value_type` of CCCL.C output iterators from `void` to the type they write to. See also: #7169

Compile time of `cub.test.device.transform.lid_0` using nvcc 13.0 and clang 20 for sm86, sm120

branch:
1m49.900s
1m50.615s
1m50.255s

main:
1m56.917s
1m57.378s
1m59.371s

Compile time of `cub.test.device.transform.lid_0` for sm86, sm120 using clang 20 in CUDA mode:

branch:
real    1m40.627s
real    1m40.675s
real    1m40.912s

main:
real    1m39.273s
real    1m39.669s
real    1m39.835s

### DIFF
--- a/cub/benchmarks/bench/transform/common.h
+++ b/cub/benchmarks/bench/transform/common.h
@@ -34,7 +34,7 @@ struct policy_selector
       cub::detail::transform::arch_to_min_bytes_in_flight(::cuda::arch_id{__CUDA_ARCH_LIST__ / 10}) + TUNE_BIF_BIAS;
 #  if TUNE_ALGORITHM == 0
     constexpr auto algorithm = cub::detail::transform::Algorithm::prefetch;
-    constexpr auto policy    = prefetch_policy{
+    constexpr auto policy    = cub::detail::transform::prefetch_policy{
       TUNE_THREADS
 #    ifdef TUNE_ITEMS_PER_THREAD_NO_INPUT
       ,
@@ -44,7 +44,7 @@ struct policy_selector
     return {min_bytes_in_flight, algorithm, policy, {}, {}};
 #  elif TUNE_ALGORITHM == 1
     constexpr auto algorithm = cub::detail::transform::Algorithm::vectorized;
-    constexpr auto policy    = vectorized_policy{
+    constexpr auto policy    = cub::detail::transform::vectorized_policy{
       TUNE_THREADS,
       (1 << TUNE_VEC_SIZE_POW2) * TUNE_VECTORS_PER_THREAD,
       (1 << TUNE_VEC_SIZE_POW2)
@@ -56,11 +56,12 @@ struct policy_selector
     return {min_bytes_in_flight, algorithm, {}, policy, {}};
 #  elif TUNE_ALGORITHM == 2
     constexpr auto algorithm = cub::detail::transform::Algorithm::memcpy_async;
-    constexpr auto policy    = async_copy_policy{TUNE_THREADS, cub::detail::transform::ldgsts_size_and_align};
+    constexpr auto policy =
+      cub::detail::transform::async_copy_policy{TUNE_THREADS, cub::detail::transform::ldgsts_size_and_align};
     return {min_bytes_in_flight, algorithm, {}, {}, policy};
 #  elif TUNE_ALGORITHM == 3
     constexpr auto algorithm = cub::detail::transform::Algorithm::ublkcp;
-    constexpr auto policy    = async_copy_policy{
+    constexpr auto policy    = cub::detail::transform::async_copy_policy{
       TUNE_THREADS, cub::detail::transform::bulk_copy_alignment(::cuda::arch_id{__CUDA_ARCH_LIST__ / 10})};
     return {min_bytes_in_flight, algorithm, {}, {}, policy};
 #  else // TUNE_ALGORITHM


c/parallel/src/transform.cu
```diff
@@ -238,6 +196,14 @@ public:
     return (is_pointer_aligned(its, its.value_size * vec_size) && ...);
   }
 };
+
+auto make_iterator_info(cccl_iterator_t input_it) -> cdt::iterator_info
+{
+  return {static_cast<int>(input_it.value_type.size),
+          static_cast<int>(input_it.value_type.alignment),
+          /* trivially_relocatable */ true, // TODO(bgruber): how to check this properly?
+          input_it.type == CCCL_POINTER}; // TODO(bgruber): how to check this properly?
```
bernhardmgruber: I would appreciate some cccl.c maintainer input here. How I do know whether the iterator's value type is trivially relocatable and the iterator is contiguous?

c/parallel/src/transform.cu
```diff
@@ -692,10 +630,12 @@ CUresult cccl_device_transform_cleanup(cccl_device_transform_build_result_t* bui
...
     }
-    std::unique_ptr<char[]> cubin(reinterpret_cast<char*>(build_ptr->cubin));
-    std::unique_ptr<transform::runtime_tuning_policy> rtp(
-      reinterpret_cast<transform::runtime_tuning_policy*>(build_ptr->runtime_policy));
-    std::unique_ptr<transform::cache> cache(reinterpret_cast<transform::cache*>(build_ptr->cache));
+    using namespace cub::detail::transform;
+    std::unique_ptr<char[]> cubin(static_cast<char*>(build_ptr->cubin));
+    std::unique_ptr<arch_policies<1>> rtp(static_cast<arch_policies<1>*>(build_ptr->runtime_policy)); // FIXME(bgruber):
+                                                                                                      // handle <2> as
+                                                                                                      // well
```
bernhardmgruber: Is there any way in this function to distinguish whether we build the unary or binary transform?

cub/cub/device/dispatch/kernels/kernel_transform.cuh
```diff
@@ -559,7 +560,7 @@ _CCCL_DEVICE void transform_kernel_ldgsts(
   static_assert(ldgsts_size_and_align <= 16);
   _CCCL_ASSERT(reinterpret_cast<uintptr_t>(smem) % ldgsts_size_and_align == 0, "");
 
-  constexpr int block_threads = LdgstsPolicy::block_threads;
+  constexpr int block_threads = Policy.async_copy_policy.block_threads;
```
miscco: I assume that the previous name of the policy was historically obsolete?

cub/cub/device/dispatch/kernels/kernel_transform.cuh
```diff
@@ -950,88 +951,93 @@ _CCCL_HOST_DEVICE auto make_aligned_base_ptr_kernel_arg(It ptr, int alignment) -
   return arg;
 }
 
-template <typename ActivePolicy>
+template <typename ArchPolicies>
 _CCCL_API constexpr int get_block_threads_helper()
 {
-  if constexpr (ActivePolicy::algorithm == Algorithm::prefetch)
+  constexpr transform_arch_policy policy = ArchPolicies{}(::cuda::arch_id{CUB_PTX_ARCH / 10});
```
miscco: I hate the arcane `/ 10 ` here with a passion

cub/cub/device/dispatch/kernels/kernel_transform.cuh
```diff
@@ -950,88 +951,93 @@ _CCCL_HOST_DEVICE auto make_aligned_base_ptr_kernel_arg(It ptr, int alignment) -
...
-    int num_elem_per_thread,
-    [[maybe_unused]] bool can_vectorize,
-    Predicate pred,
-    F f,
-    RandomAccessIteratorOut out,
-    kernel_arg<RandomAccessIteartorsIn>... ins)
+          typename... RandomAccessIteratorsIn>
+#if _CCCL_HAS_CONCEPTS()
+  requires transform_policy_hub<ArchPolicies>
+#endif // _CCCL_HAS_CONCEPTS()
```
miscco, line 1014: Nitpick: I believe we should either use the concept emulation or plain SFINAE in C++17 too

cub/cub/device/dispatch/kernels/kernel_transform.cuh
```diff
@@ -950,88 +951,93 @@ _CCCL_HOST_DEVICE auto make_aligned_base_ptr_kernel_arg(It ptr, int alignment) -
...
+  [[maybe_unused]] bool can_vectorize,
+  Predicate pred,
+  F f,
+  RandomAccessIteratorOut out,
+  kernel_arg<RandomAccessIteratorsIn>... ins)
 {
   _CCCL_ASSERT(blockDim.y == 1 && blockDim.z == 1, "transform_kernel only supports 1D blocks");
 
-  if constexpr (MaxPolicy::ActivePolicy::algorithm == Algorithm::prefetch)
+  static constexpr const transform_arch_policy policy = ArchPolicies{}(::cuda::arch_id{CUB_PTX_ARCH / 10});
```
miscco:
```suggestion
  static constexpr transform_arch_policy policy = ArchPolicies{}(::cuda::arch_id{CUB_PTX_ARCH / 10});
```

cub/cub/device/dispatch/tuning/tuning_transform.cuh
```diff
@@ -66,137 +67,185 @@ enum class Algorithm
...
-struct async_copy_policy_t
-{
-  static constexpr int block_threads = BlockThreads;
-  // items per tile are determined at runtime. these (inclusive) bounds allow overriding that value via a tuning policy
-  static constexpr int min_items_per_thread = 1;
-  static constexpr int max_items_per_thread = 32;
+  _CCCL_API constexpr friend bool operator!=(const prefetch_policy& lhs, const prefetch_policy& rhs)
+  {
+    return !(lhs == rhs);
+  }
```
miscco: Nitpick: could be
```suggestion
#if _CCCL_STD_VER <= 2017
  _CCCL_API constexpr friend bool operator!=(const prefetch_policy& lhs, const prefetch_policy& rhs)
  {
    return !(lhs == rhs);
  }
#endif // _CCCL_STD_VER <= 2017
```

Applies throughout

cub/cub/device/dispatch/tuning/tuning_transform.cuh
```diff
@@ -66,137 +67,185 @@ enum class Algorithm
...
-  static constexpr int vec_size                    = Tuning::vec_size;
+  int block_threads;
+  // items per tile are determined at runtime. these (inclusive) bounds allow overriding that value via a tuning policy
+  int items_per_thread_no_input = 2; // when there are no input iterators, the kernel is just filling
+  int min_items_per_thread      = 1;
+  int max_items_per_thread      = 32;
 
-  using not_a_vectorized_policy = void; // TODO: remove with C++20, shadows the variable in prefetch_policy_t
-};
+  _CCCL_API constexpr friend bool operator==(const prefetch_policy& lhs, const prefetch_policy& rhs)
```
miscco: Nitpick: Missing 
```suggestion
  [[nodiscard]] _CCCL_API constexpr friend bool operator==(const prefetch_policy& lhs, const prefetch_policy& rhs)
```

cub/cub/device/dispatch/tuning/tuning_transform.cuh
```diff
@@ -66,137 +67,185 @@ enum class Algorithm
...
+#endif // !_CCCL_COMPILER(NVRTC)
 };
 
-template <typename PolicyT>
-_CCCL_HOST_DEVICE TransformPolicyWrapper<PolicyT> MakeTransformPolicyWrapper(PolicyT base)
+#if _CCCL_HAS_CONCEPTS()
+// TODO(bgruber): bikeshed name before we make the tuning API public
+template <typename T>
+concept transform_policy_hub = policy_hub<T, transform_arch_policy>;
+#endif // _CCCL_HAS_CONCEPTS()
```
miscco, line 214: I believe we should reuse `_CCCL_CONCEPT` and `_CCCL_REQUIRES_EXPR` here

cub/cub/device/dispatch/tuning/tuning_transform.cuh
```diff
@@ -66,137 +67,185 @@ enum class Algorithm
...
+inline constexpr size_t size_of<void> = 0;
+
+template <typename T>
+inline constexpr size_t align_of = alignof(T);
+
+template <>
+inline constexpr size_t align_of<void> = 0;
+
+template <typename It>
+_CCCL_API constexpr auto make_iterator_info() -> iterator_info
```
miscco: Nitpick:
```suggestion
[[nodiscard]] _CCCL_API constexpr auto make_iterator_info() -> iterator_info
```

cub/cub/device/dispatch/tuning/tuning_transform.cuh
```diff
@@ -207,130 +256,225 @@ _CCCL_HOST_DEVICE constexpr auto loaded_bytes_per_iteration() -> int
...
+  for (const auto& input : inputs)
   {
-    smem_size += tile_padding + static_cast<int>(vt_size) * tile_size;
+    smem_size += tile_padding + input.value_type_size * tile_size;
   }
   return smem_size;
 }
 
-_CCCL_HOST_DEVICE constexpr int arch_to_min_bytes_in_flight(int sm_arch)
+_CCCL_HOST_DEVICE constexpr int arch_to_min_bytes_in_flight(::cuda::arch_id arch)
```
miscco:
```suggestion
[[nodiscard]] _CCCL_HOST_DEVICE constexpr int arch_to_min_bytes_in_flight(::cuda::arch_id arch)
```

cub/cub/device/dispatch/tuning/tuning_transform.cuh
```diff
@@ -207,130 +256,225 @@ _CCCL_HOST_DEVICE constexpr auto loaded_bytes_per_iteration() -> int
...
-_CCCL_HOST_DEVICE static constexpr auto make_sizes_alignments()
-{
-  return ::cuda::std::array<::cuda::std::pair<::cuda::std::size_t, ::cuda::std::size_t>,
-                            sizeof...(RandomAccessIteratorsIn)>{
-    {{sizeof(it_value_t<RandomAccessIteratorsIn>), alignof(it_value_t<RandomAccessIteratorsIn>)}...}};
-}
-
-template <int PtxVersion, int StoreSize, int... LoadSizes>
-struct tuning_vec
+[[nodiscard]] _CCCL_API constexpr auto vectorized_policy_for_filling(::cuda::arch_id arch, int store_size)
```
miscco: This is using the blessed `_CCCL_API ` which is not used above. I believe we should be consistent

cub/cub/device/dispatch/tuning/tuning_transform.cuh
```diff
@@ -207,130 +256,225 @@ _CCCL_HOST_DEVICE constexpr auto loaded_bytes_per_iteration() -> int
...
+    bool can_memcpy_contiguous_inputs           = true;
+    bool all_value_types_have_power_of_two_size = ::cuda::is_power_of_two(output.value_type_size);
+    for (const auto& input : inputs)
+    {
+      all_inputs_contiguous &= input.is_contiguous;
+      all_input_values_trivially_reloc &= input.value_type_is_trivially_relocatable;
+      // the vectorized kernel supports mixing contiguous and non-contiguous iterators
+      can_memcpy_contiguous_inputs &= !input.is_contiguous || input.value_type_is_trivially_relocatable;
+      all_value_types_have_power_of_two_size &= ::cuda::is_power_of_two(input.value_type_size);
+    }
```
miscco, line 347: Nitpick: While it is technically more efficient, I believe it would improve readability if we did
```c++
    const bool all_inputs_contiguous = ::cuda::std::all_of(input.begin(), input.end(), [](const auto& input) { return input.is_contiguous; })
```

cub/cub/device/dispatch/tuning/tuning_transform.cuh
```diff
@@ -207,130 +256,225 @@ _CCCL_HOST_DEVICE constexpr auto loaded_bytes_per_iteration() -> int
...
+      return transform_arch_policy{
+        min_bif,
+        algorithm,
+        prefetch,
+        vectorized,
+        async,
+      };
+    }
+
+    if (arch >= ::cuda::arch_id::sm_80)
```
miscco: Nitpick: the above condition returns unconditionally, I believe it would improve clarity if we turn this into an `else if`

cub/cub/device/dispatch/kernels/kernel_transform.cuh
bernhardmgruber: yes

cub/cub/device/dispatch/kernels/kernel_transform.cuh
bernhardmgruber: I would love to call `::cuda::current_arch_id()` but it's not `constexpr` on NVHPC by design.

cub/cub/device/dispatch/kernels/kernel_transform.cuh
bernhardmgruber: Hmm. We could also `static_assert`, but `ArchPolicies` is already used in the kernel attributes before we reach the body. And using a `static_assert` would only be evaluated in the device path. 

How would I write that using concept emulation and have the concept check before the `__launch_bounds__`?

cub/cub/device/dispatch/kernels/kernel_transform.cuh
bernhardmgruber: I had this before when I was still passing a `static constexpr transform_arch_policy*` as the template argument. The `const` was needed in addition to `constexpr`, for a reason that is beyond me.

cub/cub/device/dispatch/tuning/tuning_transform.cuh
bernhardmgruber: I may want to generally replace more `_CCCL_HOST_DEVICE` by `_CCCL_API`, but outside this PR. I will fix this one occurrence.

cub/cub/device/dispatch/tuning/tuning_transform.cuh
bernhardmgruber: Nah, I feel this is a bit too much. We should just upgrade to C++20 and replace all of this by a defaulted spaceship.

bernhardmgruber:
I see tiny changes in the generated SASS for `cub.bench.transform.babelstream.base`, notable in the filling kernels (no inputs) for `complex<float>`. The compiler now generates `STG.E.ENL2.256`, which it didn't do before.  

The fill lernel for `int128` seems to have degraded from generating `STG.E.128` to a lot more `STG.E`.

All kernels with a functor marked as `__callable_permitting_copied_arguments` show no changes. That's good.

It feels a bit like the items per thread changed for the fill kernels.

bernhardmgruber:
> It feels a bit like the items per thread changed for the fill kernels.

They did. Before we had a tuning policy for sm_120, that was not taken into account :D This PR now uses it.

bernhardmgruber:
I disabled the sm120 fill policy and now the only SASS diff for filling is on:
```
void cub::_V_300300_SM_1200::detail::transform::transform_kernel<cub::_V_300300_SM_1200::detail::transform::policy_hub<false, true, cuda::std::__4::tuple<cuda::__4::counting_iterator<long, 0, 0>>, unsigned long*>::policy1000, long, cub::_V_300300_SM_1200::detail::transform::always_true_predicate, cuda::__4::__callable_permitting_copied_arguments<(anonymous namespace)::lognormal_adjust_t<unsigned long>>, unsigned long*, cuda::__4::counting_iterator<long, 0, 0>>(long, int, bool, cub::_V_300300_SM_1200::detail::transform::always_true_predicate, cuda::__4::__callable_permitting_copied_arguments<(anonymous namespace)::lognormal_adjust_t<unsigned long>>, unsigned long*, cub::_V_300300_SM_1200::detail::transform::kernel_arg<cuda::__4::counting_iterator<long, 0, 0>>)
```
which is a `thrust::tabulate` of a `counting_iterator<long>` and an `unsigned long*`.

bernhardmgruber:
Found the final issue with the fill kernels. Disabled the vectorized tunings when we have input streams (they were tuned for output only use cases). SASS of `cub.bench.transform.fill.base` now matches baseline on sm120.

cub/cub/device/dispatch/kernels/kernel_transform.cuh
```diff
@@ -201,11 +202,18 @@ _CCCL_HOST_DEVICE _CCCL_CONSTEVAL auto load_store_type()
...
-template <typename VectorizedPolicy,
-          typename Offset,
-          typename F,
-          typename RandomAccessIteratorOut,
-          typename... RandomAccessIteratorsIn>
+// FIXME(bgruber): nvcc 12.0 - 13.1 crash with `error: Internal Compiler Error (codegen): "unexpected error in codegen
+// for function: found previous definition of same function!"` when we pass a const& as template parameter (and the
+// function template body contains a lambda). As a workaround, we pass the parts of the policy by value.
+// TODO(bgruber): In C++20, we should just pass transform_arch_policy by value.
+template < // const transform_arch_policy& Policy,
```
miscco: Is there a reason to pass `transform_arch_policy` by `const&` in the template arguments?

cub/cub/device/dispatch/tuning/tuning_transform.cuh
```diff
@@ -66,137 +67,187 @@ enum class Algorithm
...
 {
-  return policy;
-}
+  int block_threads;
+  int items_per_thread_vectorized;
+  int vec_size;
+  // if we have to fall back to prefetching, use these values:
+  int prefetch_items_per_thread_no_input = 2;
+  int prefetch_min_items_per_thread      = 1;
+  int prefetch_max_items_per_thread      = 32;
```
miscco, line 124: Should this rather hold a `prefetch_policy` instead of the individual members?

cub/cub/device/dispatch/tuning/tuning_transform.cuh
```diff
@@ -66,137 +67,187 @@ enum class Algorithm
...
+              << policy.bulk_copy_alignment << ", .min_items_per_thread = " << policy.min_items_per_thread
+              << ", .max_items_per_thread = " << policy.max_items_per_thread << " }";
   }
+#endif // !_CCCL_COMPILER(NVRTC)
+};
 
-  _CCCL_HOST_DEVICE static constexpr auto VectorizedPolicy()
+struct transform_arch_policy
+{
+  int min_bif;
```
miscco: We have `min_items_per_thread ` so should this be `min_bytes_in_flight`

cub/cub/device/dispatch/tuning/tuning_transform.cuh
```diff
@@ -207,139 +258,253 @@ _CCCL_HOST_DEVICE constexpr auto loaded_bytes_per_iteration() -> int
...
+    //   return vectorized_policy{256, 8, 4};
+    // }
+    // manually tuned fill on B200, same as H200
+    if (arch >= ::cuda::arch_id::sm_90)
+    {
+      return vectorized_policy{
+        store_size > 4 ? 128 : 256, 16, ::cuda::std::max(8 / store_size, 1) /* 64-bit instructions */};
+    }
+    // manually tuned fill on A100
+    if (arch >= ::cuda::arch_id::sm_90) // TODO(bgruber): this should be sm_80
```
miscco: Why is this then not changed? This code path cannot be taken currently

cub/cub/util_device.cuh
```diff
@@ -25,11 +25,16 @@
...
 #include <cuda/__device/compute_capability.h>
+#include <cuda/std/__concepts/regular.h>
 #include <cuda/std/__type_traits/conditional.h>
 #include <cuda/std/__utility/forward.h>
 #include <cuda/std/array>
 #include <cuda/std/cassert>
 
+#if _CCCL_HAS_CONCEPTS()
+#  include <cuda/std/concepts>
+#endif // _CCCL_HAS_CONCEPTS()
```
miscco: We should only include the concepts that we need

cub/cub/device/dispatch/kernels/kernel_transform.cuh
bernhardmgruber: `transform_arch_policy` is not a literal type in C++17, so we can only pass a pointer or reference to a `static constexpr` instance of it. However, nvcc dies if I do this, so I had to pass the relevant members of `transform_arch_policy` instead.

cub/cub/device/dispatch/tuning/tuning_transform.cuh
bernhardmgruber: Because enabling it caused sudden compilation errors that I didn't understand and I had to make progress elsewhere. But this needs to be fixed before merging.

cub/cub/device/dispatch/tuning/tuning_transform.cuh
bernhardmgruber: I think I initially inherited from a `prefetch_policy`, but then designated initializers don't work anymore. Then I did a `prefetch_policy` member, which was awkward again, because now you had to write `policy.prefetch_policy.block_threads` to get the block threads for the vectorized policy in case the fall back is not needed. The current state avoids both issues, but I agree, it's not nice.

bernhardmgruber:
I pulled out the arch dispatching logic into: #7093

c/parallel/src/transform.cu
bernhardmgruber: Discussed internally with @griwes.

c/parallel/src/transform.cu
bernhardmgruber: obsolete

c/parallel/src/transform.cu
```diff
@@ -357,23 +331,16 @@ __device__ consteval auto& policy_generator() {{
...
-      algorithm,
-      min_bif,
-      transform::cdt::RuntimeTransformAgentPrefetchPolicy::from_json(runtime_policy, "prefetch_policy"),
-      transform::cdt::RuntimeTransformAgentVectorizedPolicy::from_json(runtime_policy, "vectorized_policy"),
-      transform::cdt::RuntimeTransformAgentAsyncPolicy::from_json(runtime_policy, "async_policy")};
-    build_ptr->cache = new transform::cache();
+    build_ptr->cache                      = new transform::cache();
+
+    // avoid new and delete which requires the allocated and freed types to match
+    static_assert(std::is_trivially_copyable_v<decltype(policy_sel)>);
```
miscco:
```suggestion
    static_assert(cuda::std::is_trivially_copyable_v<decltype(policy_sel)>);
```

c/parallel/src/transform.cu
```diff
@@ -357,23 +331,16 @@ __device__ consteval auto& policy_generator() {{
...
-    build_ptr->runtime_policy             = new transform::runtime_tuning_policy{
-      algorithm,
-      min_bif,
-      transform::cdt::RuntimeTransformAgentPrefetchPolicy::from_json(runtime_policy, "prefetch_policy"),
-      transform::cdt::RuntimeTransformAgentVectorizedPolicy::from_json(runtime_policy, "vectorized_policy"),
-      transform::cdt::RuntimeTransformAgentAsyncPolicy::from_json(runtime_policy, "async_policy")};
-    build_ptr->cache = new transform::cache();
+    build_ptr->cache                      = new transform::cache();
+
+    // avoid new and delete which requires the allocated and freed types to match
```
miscco: 🙀

cub/benchmarks/bench/transform/common.h
```diff
@@ -23,32 +23,48 @@
...
-  struct max_policy : cub::ChainedPolicy<500, max_policy, max_policy>
+  _CCCL_API constexpr auto operator()(cuda::arch_id) const -> cub::detail::transform::transform_policy
   {
-    static constexpr int min_bif = cub::detail::transform::arch_to_min_bytes_in_flight(__CUDA_ARCH_LIST__);
+    const int min_bytes_in_flight =
+      cub::detail::transform::arch_to_min_bytes_in_flight(__CUDA_ARCH_LIST__) + TUNE_BIF_BIAS;
 #  if TUNE_ALGORITHM == 0
-    static constexpr auto algorithm = cub::detail::transform::Algorithm::prefetch;
+    const auto algorithm = cub::detail::transform::Algorithm::prefetch;
+    const auto policy    = prefetch_policy{
```
miscco: Q: could those be constexpr?

cub/cub/device/dispatch/kernels/kernel_transform.cuh
miscco: We could write:

```suggestion
_CCCL_TEMPLATE(typename PolicySelector,
          typename Offset,
          typename Predicate,
          typename F,
          typename RandomAccessIteratorOut,
          typename... RandomAccessIteratorsIn)
_CCCL_REQUIRES(transform_policy_selector<PolicySelector>)
```

c/parallel/src/transform.cu
bernhardmgruber: I think CCCL.C by convention uses type traits from `std`, so I stayed with it.

cub/benchmarks/bench/transform/common.h
bernhardmgruber: They could. Fixed.

cub/cub/device/dispatch/tuning/tuning_transform.cuh
bernhardmgruber: Fixed by now.

cub/cub/device/dispatch/kernels/kernel_transform.cuh
bernhardmgruber: Yeah, but as discussed on Slack before, we would need to get `transform_policy_selector` and then `policy_selector` working, which we couldn't of the `is_constant_expression` check. Let's leave it.

c/parallel/src/kernels/iterators.cpp
```diff
@@ -114,7 +114,7 @@ struct {0}_proxy_t {{
 struct {0} {{
   using iterator_category = cuda::std::random_access_iterator_tag;
   using difference_type   = DIFF_T;
-  using value_type        = void;
+  using value_type        = VALUE_T;
```
bernhardmgruber, line 117: TODO: Requires pending decision on #7169

bernhardmgruber: @gevtushenko was fine with this change here.

cub/cub/device/dispatch/kernels/kernel_transform.cuh
```diff
@@ -201,11 +202,18 @@ _CCCL_HOST_DEVICE _CCCL_CONSTEVAL auto load_store_type()
...
-template <typename VectorizedPolicy,
-          typename Offset,
-          typename F,
-          typename RandomAccessIteratorOut,
-          typename... RandomAccessIteratorsIn>
+// FIXME(bgruber): nvcc 12.0 - 13.1 crash with `error: Internal Compiler Error (codegen): "unexpected error in codegen
+// for function: found previous definition of same function!"` when we pass a const& as template parameter (and the
+// function template body contains a lambda). As a workaround, we pass the parts of the policy by value.
+// TODO(bgruber): In C++20, we should just pass transform_policy by value.
+template < // const transform_policy& Policy,
```
bernhardmgruber, line 209: I wonder whether I should keep comments like these, since we will probably never implement passing policies by `const&` template parameters.

cub/cub/device/dispatch/tuning/tuning_transform.cuh
bernhardmgruber: Can I do this later? Maybe we have `std::ranges::all_of` by then.

c/parallel/src/transform.cu
bernhardmgruber: It's not pretty. Do you know any other way to write `new T` and then later delete a `void*` pointing to the `T`, without having the type `T` available?

c/parallel/src/transform.cu
bernhardmgruber: I could also use a variant, since I know the kinds of types involved. But that also seems overkill.

cub/cub/util_device.cuh
```diff
@@ -831,6 +833,22 @@ private:
...
+{
+#if _CCCL_HAS_CONCEPTS()
+_CCCL_API consteval void __needs_a_constexpr_value(auto) {}
+
+// TODO(bgruber): should we either drop the Policy template argument or rename it to policy_selector_for?
+template <typename T, typename Policy>
+concept policy_selector = requires(T hub, ::cuda::arch_id arch) {
+  requires ::cuda::std::regular<Policy>;
+  { hub(arch) } -> _CCCL_CONCEPT_VSTD::same_as<Policy>;
+  { __needs_a_constexpr_value(hub(arch)) };
```
gevtushenko: question: what was the intention for `__needs_a_constexpr_value` here? Do you want to check if hub's `operator()` can return compile-time information? If so, I don't think this works. Let's add a test for non-constexpr operator to cover this. For instance, the following type satisfies this concept:

```c++
struct policy_selector_all {
  auto operator()(arch_id id) const -> a_policy {
    int r = rand() % 5;
    return a_policy{static_cast<arch_id>(r)};
  }
};
```

maybe something along the following lines would fix this:

```c++
template <auto> inline constexpr bool __needs_a_constexpr_value = true;
// ...
requires __needs_a_constexpr_value<T{}(arch_id{})>;
```

bernhardmgruber: While fixing this I got this error and I am massively impressed:
```
/home/bgruber/dev/cccl/lib/cmake/cub/../../../cub/cub/device/dispatch/tuning/tuning_reduce.cuh(474): error: expression must have a constant value
  static_assert(__needs_a_constexpr_value2<policy_selector{}(::cuda::arch_id::sm_60)>);
                                           ^
/home/bgruber/dev/cccl/lib/cmake/cub/../../../cub/cub/util_arch.cuh(150): note #61-D: integer operation result is out of range
      ::cuda::std::clamp(nominal_4B_items_per_thread * 4 / target_type_size, 1, nominal_4B_items_per_thread * 2);
                                                         ^
/home/bgruber/dev/cccl/lib/cmake/cub/../../../cub/cub/device/dispatch/tuning/tuning_reduce.cuh(440): note #2693-D: called from:
        auto [scaled_items, scaled_threads] = scale_mem_bound(threads_per_block, items_per_thread, accum_size);
                                                             ^
```
The concept check did actually validate whether the computation of the tuning policy was sound. The reason it fails here is because a default constructed policy selector like `reduce::policy_selector{}` has a bunch of zero data members now, like `accum_size`, leading to the division by zero later. So we cannot test from just the type of a policy selector whether it returns a policy at compile-time. I will drop the constexpr test from the concept.

I was btw doing this investigation while I had Cursor (with claude-4.5-opus-high) trying to fix it as well. Cursor was fast in figuring out how to run the tests, but got really lost when the first two attempts of changing the concept definition didn't fix the problem.

cub/benchmarks/bench/transform/common.h
```diff
@@ -23,32 +23,48 @@
...
-struct policy_hub_t
+#if !TUNE_BASE
+struct policy_selector
 {
-  struct max_policy : cub::ChainedPolicy<500, max_policy, max_policy>
+  _CCCL_API constexpr auto operator()(cuda::arch_id) const -> cub::detail::transform::transform_policy
   {
-    static constexpr int min_bif = cub::detail::transform::arch_to_min_bytes_in_flight(__CUDA_ARCH_LIST__);
+    const int min_bytes_in_flight =
+      cub::detail::transform::arch_to_min_bytes_in_flight(__CUDA_ARCH_LIST__) + TUNE_BIF_BIAS;
```
gevtushenko: question: does this require cuda arch list to have exactly one value? Check above triggers for multiple architectures only when algorithm != 3. Should we move the check of number of architectures outside of algorithm check?

gevtushenko: important: `arch_to_min_bytes_in_flight` expects `cuda::arch_id` not `int`. Also, arch list would provide a value like 1200, but `arch_id` expects 120. Please, try building a tuning variant.

cub/benchmarks/bench/transform/common.h
```diff
@@ -23,32 +23,48 @@
...
+      TUNE_ITEMS_PER_THREAD_NO_INPUT
+#    endif
+    };
+    return {min_bytes_in_flight, algorithm, policy, {}, {}};
 #  elif TUNE_ALGORITHM == 1
-    static constexpr auto algorithm = cub::detail::transform::Algorithm::ublkcp;
+    constexpr auto algorithm = cub::detail::transform::Algorithm::vectorized;
+    constexpr auto policy    = vectorized_policy{
+      TUNE_THREADS,
+      TUNE_VEC_SIZE * TUNE_VECTORS_PER_THREAD,
```
gevtushenko: question: the field on this policy is called `items_per_thread_vectorized`, which is a bit confusing. It's not clear if this expects number of vectorized items or not. Should we rename this to just `items_per_thread`?

cub/benchmarks/bench/transform/babelstream.cu
```diff
@@ -1,11 +1,13 @@
 // SPDX-FileCopyrightText: Copyright (c) 2024, NVIDIA CORPORATION. All rights reserved.
 // SPDX-License-Identifier: BSD-3-Clause
 
-// Because CUB cannot inspect the transformation function, we cannot add any tunings based on the results of this
-// benchmark. Its main use is to detect regressions.
-
+// %RANGE% TUNE_BIF_BIAS alg -16:16:4
+// %RANGE% TUNE_ALGORITHM alg 1:4:1
```
gevtushenko: important: here and in other benchmarks, code in `common.h` expects algorithm index to start from zero `#if TUNE_ALGORITHM == 0`. Why do we ignore the first algorithm here? To my understanding, we are exclusing prefetch from consideration this way.

cub/benchmarks/bench/transform/babelstream.cu
```diff
@@ -1,11 +1,13 @@
...
 // %RANGE% TUNE_THREADS tpb 128:1024:128
-// %RANGE% TUNE_ALGORITHM alg 0:2:1
+
+// TODO(bgruber): those parameters only apply if TUNE_ALGORITHM == 1 (vectorized)
+// %RANGE% TUNE_VEC_SIZE ipt 1:32:1
+// %RANGE% TUNE_VECTORS_PER_THREAD vpt 1:4:1
```
gevtushenko: suggestion: maybe we can start the range with zero and add a preprocessor check in such a way that if algorithm is `!= 1` we error out and shortcircuit the process, and if algorithm `== 1`, we error out if these parameters are `== 0`.

cub/cub/device/dispatch/tuning/tuning_transform.cuh
```diff
@@ -66,137 +67,183 @@ enum class Algorithm
...
+#endif // _CCCL_HAS_CONCEPTS()
+
+struct iterator_info
 {
-  return TransformPolicyWrapper<PolicyT>(base);
+  int value_type_size;
+  int value_type_alignment;
+  bool value_type_is_trivially_relocatable;
+  bool is_contiguous;
+};
```
gevtushenko: suggestion: not for this PR, but we should re-use types that already exist to describe iterators from cccl.c, or extract a common part into a common header and extend it on cccl.c end. Consider filing an issue for a follow up refactoring.

c/parallel/src/transform.cu
```diff
@@ -563,20 +529,16 @@ __device__ consteval auto& policy_generator() {{
...
   build_ptr->cubin_size                 = result.size;
-  build_ptr->runtime_policy             = new transform::runtime_tuning_policy{
-    algorithm,
-    min_bif,
-    transform::cdt::RuntimeTransformAgentPrefetchPolicy::from_json(runtime_policy, "prefetch_policy"),
-    transform::cdt::RuntimeTransformAgentVectorizedPolicy::from_json(runtime_policy, "vectorized_policy"),
-    transform::cdt::RuntimeTransformAgentAsyncPolicy::from_json(runtime_policy, "async_policy")};
-  build_ptr->cache = new transform::cache();
+  build_ptr->cache                      = new transform::cache();
+
```
gevtushenko, line 536: important: to my understanding, we no longer have to parse policy from the PTX. Let's drop this.

bernhardmgruber: This is for caching the items per thread we computed based on the occupancy in the transform dispatch, which involves a bunch of CUDA API calls and that's why we cache it. In ordinary CUB we can do this in a `static` variable, since there is one such variable for each template instantiation of the CUB algo, and each template instantiation with a distinct set of types may have a different config. In CCCL.C, we type-erase the iterators so the caching mechanism has to work differently, which is handled by the above cache. In my understanding, this is still needed.

cub/cub/device/dispatch/tuning/tuning_transform.cuh
bernhardmgruber: Agreed!

cub/benchmarks/bench/transform/common.h
bernhardmgruber: Both comments point out issues, will fix!

I didn't verify if I could compile tuning benchmarks, since the did not work before this PR already. I planned to postpone this work to #5057, soon after this PR.

cub/benchmarks/bench/transform/babelstream.cu
bernhardmgruber: That understanding is true, because I presumed that other transform implementations are faster than prefetch. But it would leave us blind in case prefetch would really be better, so you are right, we should include `TUNE_ALGORITHM == 0` in tuning.

cub/benchmarks/bench/transform/common.h
bernhardmgruber: Good point! I had `items_per_thread_vectorized` in the past because the `vectorized_policy` inherited from the `prefetch_policy` (since the vectorized kernel can fall back to the prefetch). Because the `prefetch_policy` had a member `items_per_thread` I had to use a different name. This is no longer required.

cub/cub/device/dispatch/tuning/tuning_transform.cuh
bernhardmgruber: Done.

cub/benchmarks/bench/transform/babelstream.cu
bernhardmgruber: I really have no good idea what's the best way here, so I will take your recommendation.

bernhardmgruber:
I see EVO correctly failing to build invalid parameter configurations:
```
2026-01-15 12:55:46,259: starting build for cub.bench.transform.babelstream.bif_-8.alg_3.tpb_384.vsp2_2.vpt_2: cmake --build . --target cub.bench.transform.babelstream.variant
2026-01-15 12:55:52,387: finished build for cub.bench.transform.babelstream.bif_-8.alg_3.tpb_384.vsp2_2.vpt_2 (exit code: 2) in 6.128s
2026-01-15 12:55:53,348: found cached base build for cub.bench.transform.babelstream.base
2026-01-15 12:55:53,352: found cached base build for cub.bench.transform.babelstream.base
2026-01-15 12:55:58,146: starting build for cub.bench.transform.babelstream.bif_-16.alg_0.tpb_512.vsp2_4.vpt_3: cmake --build . --target cub.bench.transform.babelstream.variant
2026-01-15 12:56:05,124: finished build for cub.bench.transform.babelstream.bif_-16.alg_0.tpb_512.vsp2_4.vpt_3 (exit code: 2) in 6.978s
2026-01-15 12:56:06,090: found cached base build for cub.bench.transform.babelstream.base
2026-01-15 12:56:06,094: found cached base build for cub.bench.transform.babelstream.base
2026-01-15 12:56:07,406: starting build for cub.bench.transform.babelstream.bif_0.alg_2.tpb_128.vsp2_2.vpt_1: cmake --build . --target cub.bench.transform.babelstream.variant
2026-01-15 12:56:14,887: finished build for cub.bench.transform.babelstream.bif_0.alg_2.tpb_128.vsp2_2.vpt_1 (exit code: 2) in 7.481s
2026-01-15 12:56:47,128: found cached base build for cub.bench.transform.babelstream.base
2026-01-15 12:56:47,134: found cached base build for cub.bench.transform.babelstream.base
2026-01-15 12:56:55,717: starting build for cub.bench.transform.babelstream.bif_-4.alg_1.tpb_512.vsp2_5.vpt_1: cmake --build . --target cub.bench.transform.babelstream.variant
2026-01-15 12:57:23,938: finished build for cub.bench.transform.babelstream.bif_-4.alg_1.tpb_512.vsp2_5.vpt_1 (exit code: 0) in 28.221s
```
What's a bit worrying is that in the course of 4h and across 8 GPUs not a single algorithm other than 1 succeeded in building. This may suggest we need a different approach, but we can address this later.

c/parallel/src/transform.cu
gevtushenko: The code lines must've shifted due to force push. This comment was about:
```c++
  nlohmann::json runtime_policy =
    cub::detail::ptx_json::parse("device_transform_policy", {result.data.get(), result.size});
```

bernhardmgruber: Oh, I see it now. That may have slipped in during a rebase. I will file a PR to remove it.

bernhardmgruber: #7264

PR #6915, opened 2025-12-08
Remove need for hardcoded `LevelT` for histogram in c.parallel and cuda.compute

### LINKED ISSUE #5479: c.parallel: replace hardcoded types in histogram with actual types passed by user
In `histogram.cu`, we currently instantiate `cub::detail::histogram::Transforms` with hardcoded types:

```c++
cub::detail::histogram::Transforms<
  LevelT, // LevelT
  OffsetT, // OffsetT
  LevelT // SampleT
>
```

`OffsetT` is typically hardcoded in c.parallel so it is fine for now. `LevelT`, which is the datatype of the values we compare to, and `SampleT`, the datatype of the input samples, are hardcoded to `double`. We should type erase them instead so the user can set them from python.

NaderAlAwar:
## Description

closes #5479

- [x] should only be merged after #6908

No SASS differences on `sm100` for all histogram benchmarks

## Checklist

### DIFF
--- a/cub/cub/device/dispatch/dispatch_histogram.cuh
+++ b/cub/cub/device/dispatch/dispatch_histogram.cuh
@@ -54,7 +54,7 @@ CUB_NAMESPACE_BEGIN
 namespace detail::histogram
 {
 // Maximum number of bins per channel for which we will use a privatized smem strategy
-static constexpr int MAX_PRIVATIZED_SMEM_BINS = 256;
+static constexpr int max_privatized_smem_bins = 256;
 
 template <int NUM_CHANNELS,
           int NUM_ACTIVE_CHANNELS,
@@ -142,11 +142,11 @@ struct DeviceHistogramKernelSource
     [[maybe_unused]] const LowerLevelArrayT& lower_level,
     [[maybe_unused]] int channel)
   {
-    using CommonT        = typename TransformsT::ScaleTransform::CommonT;
-    using IntArithmeticT = typename TransformsT::ScaleTransform::IntArithmeticT;
+    using CommonT = typename TransformsT::ScaleTransform::CommonT;
 
     if constexpr (::cuda::std::is_integral_v<CommonT>)
     {
+      using IntArithmeticT = typename TransformsT::ScaleTransform::IntArithmeticT;
       return static_cast<IntArithmeticT>(upper_level[channel] - lower_level[channel])
            > (::cuda::std::numeric_limits<IntArithmeticT>::max() / static_cast<IntArithmeticT>(num_bins));
     }
@@ -562,7 +562,7 @@ public:
       int max_num_output_bins = max_levels - 1;
 
       // Dispatch
-      if (max_num_output_bins > detail::histogram::MAX_PRIVATIZED_SMEM_BINS)
+      if (max_num_output_bins > detail::histogram::max_privatized_smem_bins)
       {
         // Too many bins to keep in shared memory.
         constexpr int PRIVATIZED_SMEM_BINS = 0;
@@ -608,7 +608,7 @@ public:
       else
       {
         // Dispatch shared-privatized approach
-        constexpr int PRIVATIZED_SMEM_BINS = detail::histogram::MAX_PRIVATIZED_SMEM_BINS;
+        constexpr int PRIVATIZED_SMEM_BINS = detail::histogram::max_privatized_smem_bins;
 
         detail::histogram::dispatch_histogram<
           NUM_CHANNELS,
@@ -922,7 +922,7 @@ public:
       }
       int max_num_output_bins = max_levels - 1;
 
-      if (max_num_output_bins > detail::histogram::MAX_PRIVATIZED_SMEM_BINS)
+      if (max_num_output_bins > detail::histogram::max_privatized_smem_bins)
       {
         // Dispatch shared-privatized approach
         constexpr int PRIVATIZED_SMEM_BINS = 0;
@@ -968,7 +968,7 @@ public:
       else
       {
         // Dispatch shared-privatized approach
-        constexpr int PRIVATIZED_SMEM_BINS = detail::histogram::MAX_PRIVATIZED_SMEM_BINS;
+        constexpr int PRIVATIZED_SMEM_BINS = detail::histogram::max_privatized_smem_bins;
 
         detail::histogram::dispatch_histogram<
           NUM_CHANNELS,
@@ -1242,7 +1242,7 @@ public:
    * @param stream
    *   CUDA stream to launch kernels within. Default is stream<sub>0</sub>.
    */
-  template <typename MaxPolicyT = typename ::cuda::std::_If<
+  template <typename MaxPolicyT = typename ::cuda::std::conditional_t<
               ::cuda::std::is_void_v<PolicyHub>,
               /* fallback_policy_hub */
               detail::histogram::policy_hub<SampleT, CounterT, NUM_CHANNELS, NUM_ACTIVE_CHANNELS, /* isEven */ 0>,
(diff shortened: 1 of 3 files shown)


cub/cub/device/dispatch/kernels/kernel_histogram.cuh
```diff
@@ -239,38 +227,14 @@ struct Transforms
...
 
-      // Check whether accurate bin computation for an integral sample type may overflow
-      if (MayOverflow(static_cast<CommonT>(num_levels - 1)))
-      {
-        return cudaErrorInvalidValue;
-      }
-
       m_scale = this->ComputeScale(num_levels, m_max, m_min);
-      return cudaSuccess;
     }
```
bernhardmgruber, line 238: Q: Why is this no longer needed?

cub/cub/device/dispatch/dispatch_histogram.cuh
```diff
@@ -585,14 +690,16 @@ public:
               ::cuda::std::is_void_v<PolicyHub>,
               /* fallback_policy_hub */
               detail::histogram::policy_hub<SampleT, CounterT, NUM_CHANNELS, NUM_ACTIVE_CHANNELS, /* isEven */ 0>,
-              PolicyHub>::MaxPolicy>
-  CUB_RUNTIME_FUNCTION static cudaError_t DispatchRange(
+              PolicyHub>::MaxPolicy,
+            typename NumOutputLevelsArrayT = ::cuda::std::array<int, NUM_ACTIVE_CHANNELS>,
+            typename LevelsArrayT          = ::cuda::std::array<const LevelT*, NUM_ACTIVE_CHANNELS>>
+  CUB_RUNTIME_FUNCTION static cudaError_t DispatchRangeDeviceInit(
```
bernhardmgruber: Suggestion: please add new APIs below the existing ones so I have an easier time seeing whether there are any diffs on the existing APIS.

bernhardmgruber: Critical: please keep the new APIs internal by spelling them `__dispatch_range_device_init`. We are going to refactor a lot of this for the new tuning APIs and anything public is a pain.

cub/cub/device/dispatch/kernels/kernel_histogram.cuh
```diff
@@ -493,5 +467,176 @@ __launch_bounds__(int(ChainedPolicyT::ActivePolicy::AgentHistogramPolicyT::BLOCK
...
+          typename CounterT,
+          typename FirstLevelArrayT, // Upper level array for DispatchEven; Number of output levels array for
+                                     // DispatchRange
+          typename SecondLevelArrayT, // Lower level array for DispatchEven; Levels array for DispatchRange
+          typename PrivatizedDecodeOpT,
+          typename OutputDecodeOpT,
+          typename OffsetT,
+          bool IsEven>
+__launch_bounds__(int(ChainedPolicyT::ActivePolicy::AgentHistogramPolicyT::BLOCK_THREADS))
+  CUB_DETAIL_KERNEL_ATTRIBUTES void DeviceHistogramSweepKernelDeviceInit(
```
bernhardmgruber: Important: Please use `snake_case` for all function names.

cub/cub/device/dispatch/kernels/kernel_histogram.cuh
NaderAlAwar: I move the `MayOverflow` check out of this, so `Init` no longer needs to return an error code in case of overflow. The point is to make it possible to call this in the kernel

cub/cub/device/dispatch/kernels/kernel_histogram.cuh
NaderAlAwar: This is a kernel, and we've always used `PascalCase` for these. I guess I can rename it to `DeviceHistogramSweepDeviceInitKernel`, or if you want I can use snake_case

python/cuda_cccl/tests/compute/test_histogram.py
```diff
@@ -73,9 +73,11 @@ def compute_reference_histogram(h_samples, num_levels, lower_level, upper_level)
     if len(valid_samples) == 0:
         return np.zeros(num_levels - 1, dtype=np.int32)
 
-    # Compute bin indices for valid samples
+    # Compute bin indices for valid samples using float arithmetic
     bin_indices = (
-        (valid_samples - lower_level) * (num_levels - 1) / (upper_level - lower_level)
+        (valid_samples.astype(np.float64) - lower_level)
```
shwina, line 78: This change gives me a bit of pause. Considering we're using `/` (floating-point division), we shouldn't need this change strictly. I'm curious what happens if we undo this change. Is there a minimal repro of the behaviour?

shwina (approved):
Approving, pending a minor comment.

cub/cub/device/dispatch/dispatch_histogram.cuh
```diff
@@ -40,29 +40,40 @@
...
 
 #include <nv/target>
 
 CUB_NAMESPACE_BEGIN
 
 namespace detail::histogram
 {
-template <int NUM_CHANNELS, int NUM_ACTIVE_CHANNELS, typename SampleIteratorT, typename CounterT, typename OffsetT>
+// Maximum number of bins per channel for which we will use a privatized smem strategy
+static constexpr int MAX_PRIVATIZED_SMEM_BINS = 256;
```
fbusato: use macro style only for macros. You can use snake style here

cub/cub/device/dispatch/dispatch_histogram.cuh
```diff
@@ -78,20 +89,87 @@ struct DeviceHistogramKernelSource
...
+
+  template <typename NumBinsT, typename UpperLevelArrayT, typename LowerLevelArrayT>
+  CUB_RUNTIME_FUNCTION static constexpr bool MayOverflow(
+    [[maybe_unused]] NumBinsT num_bins,
+    [[maybe_unused]] const UpperLevelArrayT& upper_level,
+    [[maybe_unused]] const LowerLevelArrayT& lower_level,
+    [[maybe_unused]] int channel)
+  {
+    using CommonT        = typename TransformsT::ScaleTransform::CommonT;
+    using IntArithmeticT = typename TransformsT::ScaleTransform::IntArithmeticT;
```
fbusato: move `IntArithmeticT` within the loop

cub/cub/device/dispatch/dispatch_histogram.cuh
```diff
@@ -330,18 +433,18 @@ struct dispatch_histogram
...
+template <
+  int NUM_CHANNELS,
+  int NUM_ACTIVE_CHANNELS,
+  typename SampleIteratorT,
+  typename CounterT,
+  typename LevelT,
+  typename OffsetT,
+  typename PolicyHub    = void, // if user passes a custom Policy this should not be void
+  typename SampleT      = cub::detail::it_value_t<SampleIteratorT>, /// The sample value type of the input iterator
+  typename KernelSource = detail::histogram::
```
fbusato, line 445: I don't think is needed here detail::histogram::

cub/cub/device/dispatch/dispatch_histogram.cuh
```diff
@@ -430,6 +539,8 @@ public:
         break;
       }
 
+      using TransformsT = detail::histogram::Transforms<LevelT, OffsetT, SampleT>;
```
fbusato, line 542: same here

cub/cub/device/dispatch/dispatch_histogram.cuh
```diff
@@ -451,7 +562,7 @@ public:
       int max_num_output_bins = max_levels - 1;
 
       // Dispatch
-      if (max_num_output_bins > TransformsT::MAX_PRIVATIZED_SMEM_BINS)
+      if (max_num_output_bins > detail::histogram::MAX_PRIVATIZED_SMEM_BINS)
```
fbusato: same for a variable

cub/cub/device/dispatch/dispatch_histogram.cuh
```diff
@@ -1010,9 +1160,703 @@ public:
...
+   * @param num_rows
+   *   The number of rows in the region of interest
+   *
+   * @param row_stride_samples
+   *   The number of samples between starts of consecutive rows in the region of interest
+   *
+   * @param stream
+   *   CUDA stream to launch kernels within. Default is stream<sub>0</sub>.
+   */
+  template <typename MaxPolicyT = typename ::cuda::std::_If<
```
fbusato: I don't think we have still advantages of using `_If`

cub/cub/device/dispatch/dispatch_histogram.cuh
```diff
@@ -1010,9 +1160,703 @@ public:
...
+    OffsetT row_stride_samples,
+    cudaStream_t stream,
+    ::cuda::std::false_type /*is_byte_sample*/,
+    KernelSource kernel_source             = {},
+    KernelLauncherFactory launcher_factory = {},
+    MaxPolicyT max_policy                  = {})
+  {
+    cudaError error = cudaSuccess;
+
+    do
```
fbusato: why the do/while(0) loop?

cub/cub/device/dispatch/dispatch_histogram.cuh
```diff
@@ -1010,9 +1160,703 @@ public:
...
+      if (cudaSuccess != error)
+      {
+        break;
+      }
+    } while (0);
+
+    return error;
+  }
+
+  // Dispatch routines for device-side decode operator initialization. These
```
fbusato, line 1184: from here to the end of the file I have the impression that there is a lot of redundancy. Not sure if it can be reduced without affecting readability

cub/cub/device/dispatch/kernels/kernel_histogram.cuh
```diff
@@ -493,5 +467,176 @@ __launch_bounds__(int(ChainedPolicyT::ActivePolicy::AgentHistogramPolicyT::BLOCK
...
+    SampleIteratorT d_samples,
+    ::cuda::std::array<int, NumActiveChannels> num_output_bins_wrapper,
+    ::cuda::std::array<int, NumActiveChannels> num_privatized_bins_wrapper,
+    ::cuda::std::array<CounterT*, NumActiveChannels> d_output_histograms_wrapper,
```
fbusato, line 572: question. If it compiles it should be fine but I saw `::cuda::std::array<const LevelT*, NumActiveChannels>` in the dispatch file that implies that array pointers are read-only.

cub/cub/device/dispatch/kernels/kernel_histogram.cuh
```diff
@@ -493,5 +467,176 @@ __launch_bounds__(int(ChainedPolicyT::ActivePolicy::AgentHistogramPolicyT::BLOCK
...
+    ::cuda::std::array<CounterT*, NumActiveChannels> d_privatized_histograms_wrapper,
+    FirstLevelArrayT first_level_array,
```
fbusato: I would start using `_CCCL_GRID_CONSTANT` for all input parameters

cub/cub/device/dispatch/kernels/kernel_histogram.cuh
```diff
@@ -493,5 +467,176 @@ __launch_bounds__(int(ChainedPolicyT::ActivePolicy::AgentHistogramPolicyT::BLOCK
...
+    OffsetT num_rows,
+    OffsetT row_stride_samples,
+    int tiles_per_row,
+    GridQueue<int> tile_queue)
+{
+  OutputDecodeOpT output_decode_op[NumActiveChannels];
+  PrivatizedDecodeOpT privatized_decode_op[NumActiveChannels];
+  if constexpr (IsEven)
+  {
+    for (int channel = 0; channel < NumActiveChannels; ++channel)
```
fbusato, line 587: please add `_CCCL_PRAGMA_UNROLL_FULL`. Same in the other path

cub/cub/device/dispatch/kernels/kernel_histogram.cuh
```diff
@@ -493,5 +467,176 @@ __launch_bounds__(int(ChainedPolicyT::ActivePolicy::AgentHistogramPolicyT::BLOCK
...
+    {
+      const int num_levels   = num_output_bins_wrapper[channel] + 1;
+      const auto upper_level = first_level_array[channel];
+      const auto lower_level = second_level_array[channel];
+      privatized_decode_op[channel].Init(num_levels, upper_level, lower_level);
```
fbusato, line 592: question. `upper_level, lower_level` looks swapped, is it expected?

fbusato (changes requested):

cub/cub/device/dispatch/kernels/kernel_histogram.cuh
NaderAlAwar: The `const LevelT*` is for a different array, this one is for the output histogram. The `const LevelT*` is handled by the `FirstLevelArrayT` and `SecondLevelArrayT` template parameters below

cub/cub/device/dispatch/kernels/kernel_histogram.cuh
NaderAlAwar: Yes the ordering is correct

cub/cub/device/dispatch/dispatch_histogram.cuh
NaderAlAwar: It actually is needed since `DispatchHistogram` is not in `detail::histogram`

cub/cub/device/dispatch/dispatch_histogram.cuh
NaderAlAwar: It is hard to reuse code between the code that follows and the code above. The main reason is that the code above instantiates the `Transforms` template with `LevelT` and `SampleT`, which are not known in c.parallel since we type erase them. Previously we hardcoded them so the reason for this PR is basically to stop doing that.

The reason I introduced a new dispatch path is because this change also affects the kernel and this leads to SASS changes and performance regressions (and some improvements actually). I will file an issue once this PR is merged to look into if we can merge the two paths again.

PR #6916, opened 2025-12-08
Implement `cuda::__all_arch_ids` and `cuda::__is_specific_arch`

davebayer:


bernhardmgruber:
Thank you @davebayer!

PR #6922, opened 2025-12-09
Enable `<cuda/device>` APIs to be used on device

davebayer:
This PR enables `<cuda/device>` APIs to be used on device by calling CUDA device runtime APIs.

Once this PR is merged, the user will be able to query:
- number of CUDA devices
- current device
- device attributes
directly inside a kernel.

This change brings more complexity to our testing, because using CUDA device runtime requires linking `libcudadevrt` and separable compilation.

For that reason I introduced `WANTS_CUDADEVRT.` tag that informs `lit` to link `libcudadevrt`.

I don't know if `clang-cuda` supports linking with `libcudadevrt`, will have to find out.


pciolkosz:
Is there some specific use case for these changes? They seem complicated with the `libcudadevrt` while I think the use case is rare and narrow

Issue #5450, opened 2025-08-06
cccl.parallel: Replace `gpu_struct` with plain numpy struct types


shwina:
https://github.com/NVIDIA/cccl/pull/3218 introduced the ability to use struct types as inputs to `cuda.cccl.parallel` algorithms.

As part of the design was a `@gpu_struct` decorator, which decorates a dataclass representing the struct. The resulting class could be used to define device functions that operate on structs.

Quoting from that PR:

> #### Why not use structured dtypes?
> 
> Instead of having the user define a custom dataclass-like struct type, we could have them pass numpy [structured dtypes instead](https://numpy.org/doc/stable/user/basics.rec.html#structured-datatypes). However, this was rejected. The primary reason is that for Python functions accepting/returning structured dtypes, numba generates device functions accepting/returning pointers to structs. This is problematic for two reasons: (1) In the context of CCCL, binary/unary ops are expected to take arguments of type T rather than T*, and (2) memory management - who will keep the T* returned by the binary operation of a reduction alive?

The time has come to revisit this design decision. We have since decided to use device functions accepting pointer arguments (see github.com/NVIDIA/cccl/pull/4249) to avoid ODR violations between Python and C++ definitions. In terms of memory management, memory for the result is allocated on the C++ side and passed to the user-defined device functions by pointer, so that is no longer a concern.

leofang:
From https://github.com/NVIDIA/cccl/pull/6923#issuecomment-3633470708:
> After a discussion with @NaderAlAwar we decided to punt on this for now. Both approaches have their pros and cons. `@gpu_struct` is modeled after Warp's [approach](https://nvidia.github.io/warp/modules/runtime.html#structured-arrays) which is at least proven, so let's keep going with this for now.

@shwina what would be the pros and cons?

shwina:
Mainly the considerations were purely from an API perspective. The advantages of `gpu_struct` are:

1. It gives the user a constructor to use within a device function; i.e., one can do:
  
   ```python
   @gpu_struct
   class MyStruct:
       x: np.int32
       y: np.int32

    def my_operator(x: MyStruct) -> MyStruct:
        return MyStruct(x=0, y=0)  # silly, just an example
   ```
  
   whereas the best solution I came up with for numpy struct types is returning by tuple:

   ```python
   MyStruct = np.dtype([('x', np.int32), ('y', np.int32)], align=True)

   def my_operator(x: MyStruct) -> MyStruct:
[...]

PR #6923, opened 2025-12-09
[cuda.compute] Eliminate `gpu_struct` from public API 

shwina:
## Description

Closes https://github.com/NVIDIA/cccl/issues/5450.

This PR eliminates `gpu_struct` from the API of `cuda.compute`, in favor of numpy struct dtypes. This simplifies working with structured data.

Previously, working with structured data required the use of both `gpu_struct` _and_ numpy structured dtypes, in a way that was confusing and error prone:

```python
# Before this PR:

# Define a custom data type to store the pixel values.
@gpu_struct
class Pixel:
    r: np.int32
    g: np.int32
    b: np.int32

# Define a reduction operation that returns the pixel with the maximum green value.
def max_g_value(x, y):
    return x if x.g > y.g else y

# Prepare the input and output arrays.
d_rgb = cp.random.randint(0, 256, (10, 3), dtype=np.int32).view(Pixel.dtype)
d_out = cp.empty(1, Pixel.dtype)

# Prepare the initial value for the reduction.
h_init = Pixel(0, 0, 0)

# Perform the reduction.
cuda.compute.reduce_into(d_rgb, d_out, max_g_value, d_rgb.size, h_init)
```

Now, the `gpu_struct` is completely hidden from the user, who only needs to define and pass numpy dtypes:

```python
# After this PR:

# Define a custom data type to store the pixel values using numpy structured dtype.
pixel_dtype = np.dtype([("r", np.int32), ("g", np.int32), ("b", np.int32)], align=True)

# Define a reduction operation that returns the pixel with the maximum green value.
# Return type is inferred from the output array dtype.
def max_g_value(x, y):
    return x if x.g > y.g else y

# Prepare the input and output arrays.
d_rgb = cp.random.randint(0, 256, (10, 3), dtype=np.int32).view(pixel_dtype)
d_out = cp.empty(1, pixel_dtype)

# Prepare the initial value for the reduction using np.void.
h_init = np.void((0, 0, 0), dtype=pixel_dtype)

# Perform the reduction.
cuda.compute.reduce_into(d_rgb, d_out, max_g_value, d_rgb.size, h_init)
```

## Checklist

### DIFF
--- a/docs/python/compute.rst
+++ b/docs/python/compute.rst
@@ -70,9 +70,9 @@ of a reduction before writing to the underlying array.
 Custom Types
 ------------
 
-The ``cuda.compute`` library supports defining custom data types,
-using the :func:`gpu_struct <cuda.compute.struct.gpu_struct>` decorator.
-Here are some examples showing how to define and use custom types:
+The ``cuda.compute`` library supports custom data types using NumPy structured
+dtypes. Use ``align=True`` when creating the dtype to ensure proper memory
+layout for GPU operations. Here's an example:
 
 .. literalinclude:: ../../python/cuda_cccl/tests/compute/examples/struct/struct_reduction.py
    :language: python
--- a/docs/python/compute_api.rst
+++ b/docs/python/compute_api.rst
@@ -60,9 +60,3 @@ Operators
    .. py:attribute:: NEGATE
    .. py:attribute:: MINIMUM
    .. py:attribute:: MAXIMUM
-
-Utilities
----------
-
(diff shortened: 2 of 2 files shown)


shwina:
After a discussion with @NaderAlAwar we decided to punt on this for now. Both approaches have their pros and cons. `@gpu_struct` is modeled after Warp's [approach](https://nvidia.github.io/warp/modules/runtime.html#structured-arrays) which is at least proven, so let's keep going with this for now.

PR #6926, opened 2025-12-09
Expose not guaranteed determinism to reduce in cuda.compute

NaderAlAwar:
## Description

No differences in sass on `sm100` for `cub.bench.reduce.nondeterministic.base`

## Checklist

### DIFF
--- a/c/parallel/test/test_reduce.cpp
+++ b/c/parallel/test/test_reduce.cpp
@@ -123,7 +123,7 @@ struct reduce_run
   {
     if (determinism == CCCL_NOT_GUARANTEED)
     {
-      return cccl_device_reduce_not_guaranteed_determinism(build, d_temp_storage, temp_storage_bytes, args...);
+      return cccl_device_reduce_nondeterministic(build, d_temp_storage, temp_storage_bytes, args...);
     }
     else
     {


python/cuda_cccl/cuda/compute/_cccl_interop.py
```diff
@@ -511,8 +511,8 @@ def _check_compile_result(cubin: bytes):
         sass = "nvdiasm not found, skipping SASS validation"
         warnings.warn(sass)
 
-    assert "LDL" not in sass, "LDL instruction found in SASS"
-    assert "STL" not in sass, "STL instruction found in SASS"
+    # assert "LDL" not in sass, "LDL instruction found in SASS"
+    # assert "STL" not in sass, "STL instruction found in SASS"
```
shwina: Should we be skipping sass check for a specific test?

NaderAlAwar: No I commented these out mistakenly. Undid the change.

cub/cub/device/dispatch/dispatch_reduce_nondeterministic.cuh
```diff
@@ -289,9 +298,184 @@ struct dispatch_nondeterministic_t
...
+            OffsetT,
+            ReductionOpT,
+            InitT,
+            AccumT,
+            TransformOpT>,
+          typename KernelLauncherFactory = TripleChevronFactory>
+#if _CCCL_HAS_CONCEPTS()
+  requires reduce_policy_hub<ArchPolicies>
+#endif // _CCCL_HAS_CONCEPTS()
+CUB_RUNTIME_FUNCTION _CCCL_FORCEINLINE auto dispatch_nondeterministic(
```
bernhardmgruber, line 165: Important: this seems like a full duplication of the logic in `dispatch_nondeterministic_t` for a reason I don't understand yet. Why is this necessary?

Furthermore, I see it uses `ArchPolicies`, a concept from the new tuning API design. So it seems this PR does a partial rewrite from the old to the new tuning interface. This is good, but I think it's incomplete. There are missing rewrites from the old `dispatch_nondeterministic_t` dispatcher in some environment tests and the benchmarks. Those should be completed together with the rewrite and `dispatch_nondeterministic_t` should be removed afterwards (since it's an internal type). I could not do that for `DispatchReduce` since it's a public API type. Furthermore, as a follow-up, I think the arch policies for the ordinary and non-deterministic reduction should be split.

I can help you do some of this work, but it should be definitely split off from this PR and merged as a complete separate change.

NaderAlAwar: Moved rewrite of nondeterministic reduce dispatch to #6932

PR #6927, opened 2025-12-09
[libcu++] Rename device_transform back to launch_transform

pciolkosz:
We renamed `launch_transform` to `device_transform`, because it's applicable in more places, for example `copy_bytes`. But the new name clashes with the CUB algorithm, this PR restores the old name back


davebayer (approved):

bernhardmgruber:
That makes a lot of sense, thank you!

PR #6931, opened 2025-12-10
Properly specialize cub functions for `__nv_bfloat16`

### LINKED ISSUE #1766: CUB fails to build when __CUDA_NO_HALF_OPERATORS__ is defined.
### Discussed in https://github.com/NVIDIA/cccl/discussions/1727

<div type='discussions-op-text'>

<sup>Originally posted by **Artem-B** May  8, 2024</sup>
We've encountered another interesting conundrum, trying to build torch.

Torch builds with a bunch of `__CUDA_NO_{HALF/BFLOAT16}...` defines https://github.com/pytorch/pytorch/blob/faf0015052ee37db718bc5efa6673e0c25be1e8d/cmake/Dependencies.cmake#L1609

It works well enough for them, at least up until the CUB version included with cuda-12.4 (v2.3.0?).
However, attempting to build it with cub v2.3.2 runs into an issue. Apparently the newer version of CUB needs some of the fp16/bf16 overloads disabled by the torch build flags.

```
In file included from third_party/py/torch/aten/src/ATen/cuda/cub.cu:2:
In file included from third_party/py/torch/aten/src/ATen/cuda/cub.cuh:15:
In file included from third_party/gpus/cccl/cub/cub/cub.cuh:64:
In file included from third_party/gpus/cccl/cub/cub/device/device_histogram.cuh:52:
third_party/gpus/cccl/cub/cub/device/dispatch/dispatch_histogram.cuh:779:35: error: invalid operands to binary expression ('__half' and '__half')
  779 |                    (return sample >= min_level && sample < max_level;),
      |                            ~~~~~~ ^  ~~~~~~~~~
...
blaze-out/k8-opt-cuda12/bin/third_party/gpus/cuda/_virtual_includes/_stage/third_party/gpus/cuda/include/cuda_bf16.hpp:310:52: note: candidate function not viable: no known conversion from '__half' to 'const __nv_bfloat16' for 1st argument
  310 | __CUDA_HOSTDEVICE__ __CUDA_BF16_FORCEINLINE__ bool operator>=(const __nv_bfloat16 &lh, const __nv_bfloat16 &rh) { return __hge(lh, rh); }
      |                                                    ^          ~~~~~~~~~~~~~~~~~~~~~~~
blaze-out/k8-opt-cuda12/bin/third_party/gpus/cuda/_virtual_includes/_stage/third_party/gpus/cuda/include/cuda_bf16.hpp:367:52: note: candidate function not viable: no known conversion from '__half' to 'const __nv_bfloat162' for 1st argument
  367 | __CUDA_HOSTDEVICE__ __CUDA_BF16_FORCEINLINE__ bool operator>=(const __nv_bfloat162 &lh, const __nv_bfloat162 &rh) { return __hbge2(lh, rh); }
      |                                                    ^          ~~~~~~~~~~~~~~~~~~~~~~~~```
```

Note that the ambiguous overload here is a red herring. The real problem is that the correct overloads were not found because they were disabled.

If CUDA-provided overloads are enabled, then torch build fails:
https://github.com/pytorch/pytorch/actions/runs/8991128640/job/24733299645?pr=125707

```
[ 89%] Building CUDA object caffe2/CMakeFiles/torch_cuda.dir/__/aten/src/ATen/native/cuda/ActivationHardshrinkKernel.cu.o
/var/lib/jenkins/workspace/aten/src/ATen/native/cuda/ActivationHardshrinkKernel.cu(23): error: more than one operator ">=" matches these operands:
            built-in operator "arithmetic >= arithmetic"
            function "operator>=(const __half &, const __half &)" (declared at line 311 of /usr/local/cuda/include/cuda_fp16.hpp)
            operand types are: scalar_t >= c10::Half
```

So, we somehow need to have the cake (have CUDA-provided overloads), and eat it (have those overloads disabled), to keep both cub and torch happy.

Ideally it would be great if CUDA headers would stash the overloads in some known namespace, which would allow cub to pull them in into its own namespace, but it's not an option, those overloads are just preprocessed away. 

Edit: Not sure if it's possible, though, as the `__CUDA_NO_*` macros also disable member functions, not just free-standing ones.

Fixing torch would also be nice, but at the moment things work fine for them, and as far as torch is concerned, it's not a supported build configuration. They are still building with CUDA-12.1 and very recent versions of CUDA and related libraries are not their problem yet.

Considering that we're missing a relatively few overload functions, I wonder if it would make sense for CUB to carry its own set, and either always use them, or fall back to them if it's used in a build which disables those overloads in CUDA headers.

@jrhemstad @miscco @voznesenskym @malfet</div>

miscco:
We currently  have some places where we handle `__half` specially because a user might have defined `__CUDA_NO_NVBFLOAT_OPERATORS__`

We need  to do the same for `__nv_bfloat16`

Fixes CUB fails to build when __CUDA_NO_HALF_OPERATORS__ is defined.
Fixes #1766

### DIFF
--- a/cub/cub/agent/agent_sub_warp_merge_sort.cuh
+++ b/cub/cub/agent/agent_sub_warp_merge_sort.cuh
@@ -127,7 +127,7 @@ class AgentSubWarpSort
 #if _CCCL_HAS_NVBF16()
     _CCCL_DEVICE bool operator()(__nv_bfloat16 lhs, __nv_bfloat16 rhs) const noexcept
     {
-      // Need to explicitly cast to float for SM <= 80.
+      // Need to explicitly cast to float for SM < 80.
       if constexpr (IS_DESCENDING)
       {
         NV_IF_TARGET(
@@ -154,7 +154,7 @@ class AgentSubWarpSort
 #if _CCCL_HAS_NVBF16()
   _CCCL_DEVICE static bool equal(__nv_bfloat16 lhs, __nv_bfloat16 rhs)
   {
-    // Need to explicitly cast to float for SM <= 80.
+    // Need to explicitly cast to float for SM < 80.
     NV_IF_TARGET(NV_PROVIDES_SM_80, (return __heq(lhs, rhs);), (return __bfloat162float(lhs) == __bfloat162float(rhs);));
   }
 #endif // _CCCL_HAS_NVBF16()


bernhardmgruber (approved):

davebayer (approved):

cub/cub/agent/agent_sub_warp_merge_sort.cuh
```diff
@@ -123,6 +123,24 @@ class AgentSubWarpSort
...
 #endif // _CCCL_HAS_NVFP16()
+
+#if _CCCL_HAS_NVBF16()
+    _CCCL_DEVICE bool operator()(__nv_bfloat16 lhs, __nv_bfloat16 rhs) const noexcept
+    {
+      // Need to explicitly cast to float for SM <= 52.
+      if constexpr (IS_DESCENDING)
+      {
+        NV_IF_TARGET(
+          NV_PROVIDES_SM_53, (return __hgt(lhs, rhs);), (return __bfloat162float(lhs) > __bfloat162float(rhs);));
```
davebayer: You need at least sm80 for operations on bfloat16

cub/cub/agent/agent_sub_warp_merge_sort.cuh
```diff
@@ -123,6 +123,24 @@ class AgentSubWarpSort
       _CCCL_UNREACHABLE();
     }
 #endif // _CCCL_HAS_NVFP16()
+
+#if _CCCL_HAS_NVBF16()
+    _CCCL_DEVICE bool operator()(__nv_bfloat16 lhs, __nv_bfloat16 rhs) const noexcept
+    {
+      // Need to explicitly cast to float for SM <= 52.
```
davebayer: The comments don't match the code

cub/cub/agent/agent_sub_warp_merge_sort.cuh
```diff
@@ -123,6 +123,24 @@ class AgentSubWarpSort
       _CCCL_UNREACHABLE();
     }
 #endif // _CCCL_HAS_NVFP16()
+
+#if _CCCL_HAS_NVBF16()
+    _CCCL_DEVICE bool operator()(__nv_bfloat16 lhs, __nv_bfloat16 rhs) const noexcept
+    {
+      // Need to explicitly cast to float for SM <= 80.
```
davebayer:
```suggestion
      // Need to explicitly cast to float for SM < 80.
```

cub/cub/agent/agent_sub_warp_merge_sort.cuh
```diff
@@ -133,6 +151,14 @@ class AgentSubWarpSort
   }
 #endif // _CCCL_HAS_NVFP16()
 
+#if _CCCL_HAS_NVBF16()
+  _CCCL_DEVICE static bool equal(__nv_bfloat16 lhs, __nv_bfloat16 rhs)
+  {
+    // Need to explicitly cast to float for SM <= 80.
```
davebayer:
```suggestion
    // Need to explicitly cast to float for SM < 80.
```

PR #6938, opened 2025-12-10
Unify operator handling in cuda.compute

shwina:
## Description

This PR is a refactor, in preparation for supporting stateful ops in `cuda.compute`.

### Duplicated logic for handling `OpKind` v/s callables

Currently, user-defined operations (e.g., predicates, transformations, etc.,) can be provided either as "built-in" ops (`OpKind`) or custom functions (callables). These differ in the way they are (1) cached and (2) "compiled" into LTOIR. To handle these differences, we currently have a bunch of `if-else` statements everywhere, like:

```python
if isinstance(op, OpKind):
    self.op_wrapper = cccl.to_cccl_op(op, None)
else:
    self.op_wrapper = cccl.to_cccl_op(op, value_type(value_type, value_type))
```

and:

```python
    # Handle well-known operations differently
    op_key: Union[tuple[str, int], CachableFunction]
    if isinstance(op, OpKind):
        op_key = (op.name, op.value)
    else:
        op_key = CachableFunction(op)
```

### Determining signatures from annotations

If provided, type annotations offer a faster way to determine the return type of a user-defined callable, compared to using numba type inference. We take advantage of this in., e.g., [`transform`](https://github.com/NVIDIA/cccl/blob/04e095592cee1513e042e5878079d1f95682b957/python/cuda_cccl/cuda/compute/algorithms/_transform.py#L40-L47), but it would be nice to do this for all ops. Ideally, we don't want to repeat the logic everywhere.

## Solution

This PR solves the above by introducing an (internal) `OpAdaptor` type that encapsulates the logic for caching, signature determination, and compiling. Furthermore, it will make adding support for stateless ops much easier.

## Checklist

### DIFF
--- a/python/cuda_cccl/cuda/compute/algorithms/_reduce.py
+++ b/python/cuda_cccl/cuda/compute/algorithms/_reduce.py
@@ -45,11 +45,10 @@ class _Reduce:
         self.d_in_cccl = cccl.to_cccl_input_iter(d_in)
         self.d_out_cccl = cccl.to_cccl_output_iter(d_out)
         self.h_init_cccl = cccl.to_cccl_value(h_init)
-        self.op = op
 
         # Compile the op with value types
         value_type = get_value_type(h_init)
-        self.op_cccl = self.op.compile((value_type, value_type), value_type)
+        self.op_cccl = op.compile((value_type, value_type), value_type)
 
         self.build_result = call_build(
             _bindings.DeviceReduceBuildResult,
--- a/python/cuda_cccl/cuda/compute/algorithms/_select.py
+++ b/python/cuda_cccl/cuda/compute/algorithms/_select.py
@@ -40,7 +40,7 @@ class _Select:
         d_in: DeviceArrayLike | IteratorBase,
         d_out: DeviceArrayLike | IteratorBase,
         d_num_selected_out: DeviceArrayLike,
-        cond: Callable | OpAdapter,  # Raw callable or Operator
+        cond: OpAdapter,
     ):
         # Create discard iterators for unused outputs, using d_out as reference
         # to match the input/output type
--- a/python/cuda_cccl/cuda/compute/op.py
+++ b/python/cuda_cccl/cuda/compute/op.py
@@ -56,7 +56,7 @@ class _WellKnownOp(_OpAdapter):
         self._kind = kind
 
     def get_cache_key(self) -> Hashable:
-        return (self.__class__.__name__, self._kind.name, self._kind.value)
+        return (self._kind.name, self._kind.value)
 
     def compile(self, input_types, output_type=None) -> Op:
         return Op(
@@ -83,7 +83,7 @@ class _StatelessOp(_OpAdapter):
         self._cachable = CachableFunction(func)
 
     def get_cache_key(self) -> Hashable:
-        return (self.__class__.__name__, self._cachable)
+        return self._cachable
 
     def compile(self, input_types, output_type=None) -> Op:
         from . import _cccl_interop as cccl
--- a/python/cuda_cccl/cuda/compute/algorithms/_scan.py
+++ b/python/cuda_cccl/cuda/compute/algorithms/_scan.py
@@ -60,7 +60,6 @@ class _Scan:
     ):
         self.d_in_cccl = cccl.to_cccl_input_iter(d_in)
         self.d_out_cccl = cccl.to_cccl_output_iter(d_out)
-        self.op = op
 
         self.init_kind = get_init_kind(init_value)
 
@@ -88,7 +87,7 @@ class _Scan:
                 init_value_type_info = self.init_value_cccl.type
 
         # Compile the op with value types
-        self.op_cccl = self.op.compile((value_type, value_type), value_type)
+        self.op_cccl = op.compile((value_type, value_type), value_type)
 
         self.build_result = call_build(
             _bindings.DeviceScanBuildResult,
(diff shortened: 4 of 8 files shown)


shwina:
/ok to test https://github.com/NVIDIA/cccl/pull/6938/changes/2d3f64727e7e9b3060b44ea2f88cb80ea4d5d2e2

shwina:
/ok to test https://github.com/NVIDIA/cccl/pull/6938/changes/2d3f64727e7e9b3060b44ea2f88cb80ea4d5d2e2

shwina:
/ok to test https://github.com/NVIDIA/cccl/pull/6938/changes/2d3f64727e7e9b3060b44ea2f88cb80ea4d5d2e2

python/cuda_cccl/cuda/compute/algorithms/_sort/_merge_sort.py
```diff
@@ -146,14 +142,25 @@ def __call__(
...
+            self.op_cccl,
             stream_handle,
         )
 
         return temp_storage_bytes
 
 
-@cache_with_key(make_cache_key)
+@cache_with_key(_make_cache_key)
+def _make_merge_sort_cached(
```
shwina, line 153: Note to reviewers: this approach _does_ introduce a layer of indirection here in the caching.

I do have some ideas for unifying caching across all algorithms which should simplify this, but I'll do that in a subsequent PR.

python/cuda_cccl/cuda/compute/op.py
```diff
@@ -1,3 +1,140 @@
...
+    def __init__(self, kind: OpKind):
+        if not _is_well_known_op(kind):
+            raise ValueError(
+                f"OpKind.{kind.name} is not a well-known operation. "
+                "Use OpKind.PLUS, OpKind.MAXIMUM, etc."
+            )
+        self._kind = kind
+
+    def get_cache_key(self) -> Hashable:
+        return (self.__class__.__name__, self._kind.name, self._kind.value)
```
NaderAlAwar: Question: prior to this change, we only return `(op.name, op.value)`. Why do we need to include `self.__class__.__name__`?

python/cuda_cccl/cuda/compute/op.py
```diff
@@ -1,3 +1,140 @@
...
+    """Internal wrapper for stateless callables."""
+
+    __slots__ = ["_func", "_cachable"]
+
+    def __init__(self, func: Callable):
+        self._func = func
+        self._cachable = CachableFunction(func)
+
+    def get_cache_key(self) -> Hashable:
+        return (self.__class__.__name__, self._cachable)
```
NaderAlAwar: Same question as above, why do we need `self.__class__.__name__`?

python/cuda_cccl/cuda/compute/algorithms/_sort/_merge_sort.py
```diff
@@ -88,23 +88,19 @@ def __init__(
         self.d_in_items_cccl = cccl.to_cccl_input_iter(d_in_items)
         self.d_out_keys_cccl = cccl.to_cccl_output_iter(d_out_keys)
         self.d_out_items_cccl = cccl.to_cccl_output_iter(d_out_items)
+        self.op_adapter = op
```
NaderAlAwar, line 91: Question: why do we store `op_adapter` as a member variable? This also applies to other algorithms

python/cuda_cccl/cuda/compute/algorithms/_reduce.py
```diff
@@ -38,24 +39,23 @@ def __init__(
...
         d_in: DeviceArrayLike | IteratorBase,
         d_out: DeviceArrayLike | IteratorBase,
-        op: Callable | OpKind,
+        op: OpAdapter,
         h_init: np.ndarray | GpuStruct,
     ):
         self.d_in_cccl = cccl.to_cccl_input_iter(d_in)
         self.d_out_cccl = cccl.to_cccl_output_iter(d_out)
         self.h_init_cccl = cccl.to_cccl_value(h_init)
+        self.op = op
```
NaderAlAwar: Important: this is named `op_adapter` in `merge_sort`. We should use consistent names.

python/cuda_cccl/cuda/compute/algorithms/_select.py
```diff
@@ -38,7 +40,7 @@ def __init__(
         d_in: DeviceArrayLike | IteratorBase,
         d_out: DeviceArrayLike | IteratorBase,
         d_num_selected_out: DeviceArrayLike,
-        cond: Callable,
+        cond: Callable | OpAdapter,  # Raw callable or Operator
```
NaderAlAwar: Question: it is not clear to me why this is annotated differently than the other algorithms? Also the comment seems unnecessary

NaderAlAwar (approved):
Looks good, left a few comments. We should also be careful not to introduce too much overhead to the single phase API

python/cuda_cccl/cuda/compute/algorithms/_sort/_merge_sort.py
shwina: When we _do_ introduce stateful operators, the `op_adapter` is what will hold the state arrays. That being said, let me remove this change from this PR and introduce it (or something else) in the subsequent PR.

python/cuda_cccl/cuda/compute/algorithms/_reduce.py
shwina: I'll use `op` everywhere.

python/cuda_cccl/cuda/compute/algorithms/_select.py
shwina: Fixed

python/cuda_cccl/cuda/compute/op.py
shwina: Fixed - ditto below

PR #5975, opened 2025-09-22
implement `task_scheduler` from C++26 ([exec.task.scheduler])

ericniebler:
## Description

`task_scheduler` is a type-erased scheduler that is used by the `std::execution::task` coroutine type ([exec.task]). If the type-erased scheduler has customized the `bulk` algorithms, the `task_scheduler` will use the customized implementations to accelerate any `bulk` senders scheduled on it.

## Checklist

- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #6939, opened 2025-12-10
clean up some allocator and memory utilities

ericniebler:
## Description

while working on #5975, i found it necessary to make a few small changes to libcudacxx's allocator machinery, mostly to work around MSVC compiler bugs. this PR collects those changes into one package.

* replace uses of `enable_if` with concepts portability macros and `if constexpr`
* replace conditional compilation of calls to `construct_at` and `destroy_at` with calls to `__construct_at` and `__destroy_at`
* suppress deprecation warnings that `cuda::std::allocator` is triggering on itself.

## Checklist

### DIFF
--- a/libcudacxx/include/cuda/std/__memory/allocator_traits.h
+++ b/libcudacxx/include/cuda/std/__memory/allocator_traits.h
@@ -332,7 +332,7 @@ struct _CCCL_TYPE_VISIBILITY_DEFAULT allocator_traits
   }
 
   [[nodiscard]] _CCCL_API inline _CCCL_CONSTEXPR_CXX20 static pointer
-  allocate(allocator_type& __a, size_type __n, const_void_pointer __hint)
+  allocate(allocator_type& __a, size_type __n, [[maybe_unused]] const_void_pointer __hint)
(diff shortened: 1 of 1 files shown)


davebayer (approved):

libcudacxx/include/cuda/std/__memory/construct_at.h
```diff
@@ -102,9 +103,8 @@ _CCCL_CONCEPT __can_optimize_construct_at = _CCCL_REQUIRES_EXPR((_Tp, variadic _
 #if _CCCL_STD_VER >= 2020
 
 _CCCL_EXEC_CHECK_DISABLE
-template <class _Tp,
-          class... _Args,
-          class = decltype(::new(::cuda::std::declval<void*>()) _Tp(::cuda::std::declval<_Args>()...))>
+_CCCL_TEMPLATE(class _Tp, class... _Args)
+_CCCL_REQUIRES(is_constructible_v<_Tp, _Args...>)
```
miscco, line 107: We should change that in all overloads and `__construct_at`

PR #4249, opened 2025-03-24
Use device functions that accept pointer arguments in ccc.cl and cuda.parallel

shwina:
## Description

Initial review of this PR can be done by looking at the [individual commits](https://github.com/NVIDIA/cccl/pull/4249/commits).

Closes https://github.com/NVIDIA/cccl/issues/4372
Closes https://github.com/NVIDIA/cccl/issues/3789

This PR makes it so that user-defined operators declared in both `cuda.parallel` and `c.parallel` have the exact same signature, to ensure the one-definition rule (ODR) is not broken. Note that on the `c.parallel` side, operators are declared with "extern C", and then actually defined on the `cuda.parallel` (Python) side.

This is achieved by declaring the operator as having `void*` arguments on the C side, and correspondingly, wrapping the user-provided function in a device function with `void*` arguments on the Python side. The return value is also passed as a `void*` (final) argument.

Additional notes:

- With these changes, bugs like https://github.com/NVIDIA/cccl/issues/3789 disappear - although that's for [orthogonal](https://github.com/NVIDIA/cccl/issues/4248) reasons. :heavy_check_mark: 
- With these changes, a large majority of the failures noted in https://github.com/NVIDIA/cccl/pull/4472 disappear :heavy_check_mark:
  - the remaining (two) failures are likely due to a compiler bug that is being tracked internally

## Benchmark comparison with `main`

```bash
------------------------------------------------------------------------------------------------------------------------ benchmark: 88 tests ------------------------------------------------------------------------------------------------------------------------
Name (time in us)                                                    Min                       Max                      Mean                  StdDev                    Median                     IQR            Outliers          OPS            Rounds  Iterations
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
bench_binary_transform_iterator[1000000] (fix-struct-t)          67.2230 (7.81)         1,586.1630 (53.51)           75.3450 (5.41)          74.9515 (69.97)           68.6150 (7.43)           2.5762 (11.71)      63;748  13,272.2789 (0.18)      10801           1
bench_binary_transform_iterator[1000000] (main-benchma)          67.7940 (7.88)         2,446.3780 (82.52)           81.1264 (5.82)         117.7801 (109.95)          69.8970 (7.57)           0.6600 (3.00)     100;1315  12,326.4363 (0.17)      10287           1
[...]

PR #6941, opened 2025-12-10
[cuda.compute] Refactor code for creating void* wrappers

shwina:
## Description

This PR is a refactor, moving the logic for generating `void*` wrapper functions introduced in PRs https://github.com/NVIDIA/cccl/pull/6634 and https://github.com/NVIDIA/cccl/pull/4249 into a new module `_odr_helpers.py`. The module docstring clearly describes why we need the wrappers.

Secondly, the code implementing the logic is refactored:

- All the `create_*_wrapper` functions followed a similar pattern, differing only in the way they handled different argument types. The pattern is basically (1) bitcast arguments from void* to the appropriate argument type, (2) call the original function, (3) store the result in the return value if needed. Thus, I introduced a helper implementing this pattern.

- Previously, all of the above code was defined as a string and exec'd. However, we only truly need `exec` for the definition of the numba intrinsic. I've included this change to aid debugging. I've also added a comment as to why the `exec` is necessary. 

## Checklist

### DIFF
--- a/python/cuda_cccl/cuda/compute/_odr_helpers.py
+++ b/python/cuda_cccl/cuda/compute/_odr_helpers.py
@@ -84,9 +84,7 @@ def _codegen_void_ptr_wrapper(
     input_vals = []
     ret_ptr = None
 
-    for i, spec in enumerate(arg_specs):
-        arg = args[i]
-
+    for i, (arg, spec) in enumerate(zip(args, arg_specs)):
         match spec.mode:
             case _ArgMode.LOAD:
                 # Cast void* to typed pointer and load value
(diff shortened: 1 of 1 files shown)


python/cuda_cccl/cuda/compute/_odr_helpers.py
```diff
@@ -0,0 +1,240 @@
...
+
+    Returns:
+        LLVM dummy value (for void return)
+    """
+
+    input_vals = []
+    ret_ptr = None
+
+    for i, spec in enumerate(arg_specs):
+        arg = args[i]
```
NaderAlAwar: Suggestion: can you just use `zip`?

python/cuda_cccl/cuda/compute/_odr_helpers.py
```diff
@@ -0,0 +1,240 @@
...
+def create_advance_void_ptr_wrapper(advance_fn, state_ptr_type):
+    """Creates a wrapper function for iterator advance method.
+
+    The wrapper takes 2 void* arguments:
+    - state pointer
+    - offset pointer (points to uint64 value)
+    """
+    arg_specs = [
+        _ArgSpec(state_ptr_type, _ArgMode.PTR),
+        _ArgSpec(types.uint64, _ArgMode.LOAD),
```
NaderAlAwar: Suggestion: add a comment to explain why we hardcode this to `uint64`

NaderAlAwar (approved):
This is great! Thanks @shwina

shwina:
/ok to test https://github.com/NVIDIA/cccl/commit/5920c48089a2d8492b32b004ab87a8ccb7406a11

PR #6948, opened 2025-12-11
Avoid using `_CCCL_UNREACHABLE()` unless it's necessary

davebayer:
This PR tries to remove as many uses of `_CCCL_UNREACHABLE()` from libcu++ as possible. I believe many of them are there because of msvc 2017 which we don't support anymore.


libcudacxx/include/cuda/std/__variant/visit.h
```diff
@@ -78,12 +78,7 @@ struct __variant_binary_visitor
     {
       return __op(::cuda::std::get<0>(__lhs), ::cuda::std::get<0>(__rhs));
     }
-    // We already checked that every variant has a value, so we should never reach this line
-#if _CCCL_COMPILER(MSVC) // MSVC needs this to be wrapped in a function or it will error
-    ::cuda::std::unreachable();
-#else // ^^^ _CCCL_COMPILER(MSVC) ^^^ / vvv !_CCCL_COMPILER(MSVC) vvv
     _CCCL_UNREACHABLE();
-#endif // !_CCCL_COMPILER(MSVC)
```
miscco, line 86: We cannot remove that

This is necessary to tell the compiler that it is fine to not return anything here

libcudacxx/test/libcudacxx/std/utilities/optional/optional.monadic/or_else.pass.cpp
```diff
@@ -91,11 +87,7 @@ __host__ __device__ constexpr bool test()
     }) == 42);
     opt = val;
     opt.or_else([] {
-#if TEST_COMPILER(GCC, <, 9)
-      _CCCL_UNREACHABLE();
-#else
```
miscco, line 96: pretty sure thisis going to blow up

libcudacxx/include/cuda/std/__variant/visit.h
davebayer: I left the `_CCCL_UNREACHABLE()` there

libcudacxx/test/libcudacxx/std/utilities/optional/optional.monadic/or_else.pass.cpp
davebayer: Seems to be passing

miscco: Are we testing 12.0 with GCC < 7?

davebayer: [pull_request / libcu++ nvcc GCC / [CTK12.0 GCC7 C++17] Build(amd64) (push)](https://github.com/NVIDIA/cccl/actions/runs/20129165645/job/57766134085?pr=6948)

I guess you mean gcc < 9, seems to work

PR #6951, opened 2025-12-11
Cleanups for random module

RAMitchell:
- Add generate_canonical to random header, add basic test.
- Rename libcpp assertions to cccl

PR #6952, opened 2025-12-11
Make `__cccl_is_floating_point_v` consistent with `__cccl_is_integer_v`

davebayer:
We have `__cccl_is_integer_v` which checks if a type is cv unqualified integer type. Let's have the same thing for floating point types.

### DIFF
--- a/libcudacxx/include/cuda/std/__type_traits/is_floating_point.h
+++ b/libcudacxx/include/cuda/std/__type_traits/is_floating_point.h
@@ -28,22 +28,22 @@
 _CCCL_BEGIN_NAMESPACE_CUDA_STD
 
 template <class _Tp>
-inline constexpr bool __cccl_is_floating_point_helper_v = false;
+inline constexpr bool __cccl_is_floating_point_v = false;
 
 template <>
-inline constexpr bool __cccl_is_floating_point_helper_v<float> = true;
+inline constexpr bool __cccl_is_floating_point_v<float> = true;
 
 template <>
-inline constexpr bool __cccl_is_floating_point_helper_v<double> = true;
+inline constexpr bool __cccl_is_floating_point_v<double> = true;
 
 template <>
-inline constexpr bool __cccl_is_floating_point_helper_v<long double> = true;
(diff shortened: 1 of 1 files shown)


bernhardmgruber:
Does the PR actually change anything beyond a few renamings?

davebayer:
@bernhardmgruber no, I think it could be useful in #6951 if we want cv unqualified `is_floating_point_v`, we already do the same thing with `__cccl_is_integer_v`

libcudacxx/include/cuda/std/__type_traits/is_floating_point.h
```diff
@@ -28,22 +28,22 @@
...
-inline constexpr bool __cccl_is_floating_point_helper_v<double> = true;
+inline constexpr bool __cccl_is_floating_point_v<double> = true;
 
 template <>
-inline constexpr bool __cccl_is_floating_point_helper_v<long double> = true;
+inline constexpr bool __cccl_is_floating_point_v<long double> = true;
 
 template <class _Tp>
-inline constexpr bool is_floating_point_v = __cccl_is_floating_point_helper_v<remove_cv_t<_Tp>>;
+inline constexpr bool is_floating_point_v = __cccl_is_floating_point_v<remove_cv_t<_Tp>>;
```
bernhardmgruber, line 43: @davebayer but the old implementation already did `remove_cv_t`, so what's the new behavior of this PR?

fbusato: I think the purpose of the PR is to make the naming consistent, which makes sense

davebayer: It's just to make `__cccl_is_floating_point_v` usable in the codebase

bernhardmgruber: Ahh, I thought you meant `is_floating_point_v`. Sorry, all good!

davebayer: It's just a cosmetic change..

PR #6954, opened 2025-12-11
[cuda.compute] Add support for ops that capture state

shwina:
## Description

This PR adds support for ops that can capture state. This can be used to implement algorithms wide side-effects:

```python
from cuda.compute import select
import numpy as np
import cupy as cp
import numba.cuda as numba_cuda


# Create device state for counting rejected items
reject_counter = cp.zeros(1, dtype=np.int32)

# Define condition that references state as closure
def count_rejects(x):
    if x > 50:
        return True
    else:
        numba_cuda.atomic.add(reject_counter, 0, 1)
        return False

d_in = cp.arange(100, dtype=np.int32)
d_out = cp.empty_like(d_in)
d_num_selected = cp.empty(2, dtype=np.uint64)

select(
    d_in,
    d_out,
    d_num_selected,
    count_rejects,
    len(d_in),
)

print(reject_counter)
```

prints:

```
[51]
```

## Checklist

### DIFF
--- a/python/cuda_cccl/benchmarks/compute/bench_reduce.py
+++ b/python/cuda_cccl/benchmarks/compute/bench_reduce.py
@@ -91,10 +91,7 @@ def bench_reduce_pointer(bench_fixture, request, size):
         reduce_pointer(input_array, build_only=(bench_fixture == "compile_benchmark"))
 
     fixture = request.getfixturevalue(bench_fixture)
-    if bench_fixture == "compile_benchmark":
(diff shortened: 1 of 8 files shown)


shwina:
/ok to test https://github.com/NVIDIA/cccl/pull/6954/commits/686268f642275a6eb707608e91eb1f726020ae22

shwina:
There's a much simpler solution that involves hacking numba-cuda directly. I'll open a PR implementing that.

PR #6956, opened 2025-12-11
Provide thrust::find_if benchmark

gonidelis:
Provides `thrust::find_if` benchmark that will be needed when new `cub::FindIf` is introduced and `thrust::find` ports to that.

### DIFF
--- a/thrust/benchmarks/bench/find_if/basic.cu
+++ b/thrust/benchmarks/bench/find_if/basic.cu
@@ -21,7 +21,7 @@ void find_if(nvbench::state& state, nvbench::type_list<T>)
   thrust::fill(dinput.begin(), dinput.begin() + mismatch_point, T{0});
   thrust::fill(dinput.begin() + mismatch_point, dinput.end(), val);
 
-  state.add_global_memory_reads<T>(mismatch_point);
+  state.add_global_memory_reads<T>(mismatch_point + 1);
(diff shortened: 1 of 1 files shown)


thrust/benchmarks/bench/find_if/basic.cu
```diff
@@ -0,0 +1,36 @@
...
+  // set up input
+  const auto elements       = static_cast<std::size_t>(state.get_int64("Elements"));
+  const auto common_prefix  = state.get_float64("MismatchAt");
+  const auto mismatch_point = static_cast<std::size_t>(elements * common_prefix);
+
+  thrust::device_vector<T> dinput(elements, thrust::no_init);
+  thrust::fill(dinput.begin(), dinput.begin() + mismatch_point, T{0});
+  thrust::fill(dinput.begin() + mismatch_point, dinput.end(), val);
+
+  state.add_global_memory_reads<T>(mismatch_point);
```
bernhardmgruber: Thinking about it, if you scan `dinput` contiguously from the beginning, you need to read `mismatch_point` many zeros, and at least one more element to detect the mismatch. So it shoul dbe:
```suggestion
  state.add_global_memory_reads<T>(mismatch_point + 1);
```

bernhardmgruber (approved):

PR #6957, opened 2025-12-11
Rework hierarchy levels

### LINKED ISSUE #6885: [FEA]: Improve device hierarchy queries
### Is this a duplicate?

### Area

libcu++

Current way of using hierarchy queries is:
```cpp
cuda::hierarchy_dimensions<...> dims;
dims.extents(cuda::block, cuda::grid);
dims.index(cuda::block, cuda::grid);
```
which is a bit hard to read and can look weird for the `.index(...)` for example.

I'd like to do something that is more natural to a CUDA user:
```cpp
cuda::hierarchy_dimensions<...> dims;
cuda::block.extents(cuda::grid, dims);
cuda::thread.index(cuda::grid, dims);
```
I think it's easier to read: _block index in grid for given dims_

The nice thing about this approach is that a user needn't to use hierarchy dimensions at all and do:
```cpp
cuda::block.extents(cuda::grid);
cuda::thread.index(cuda::grid);
```

I will also provide `.meow_as<int>(...)` variants which can override the type that is returned by the query.

_No response_

### Additional context

_No response_

davebayer:
Closes #6885

### DIFF
--- a/c2h/include/c2h/test_util_vec.h
+++ b/c2h/include/c2h/test_util_vec.h
@@ -49,198 +49,199 @@ inline int CoutCast(signed char val)
 /**
  * Vector1 overloads
  */
-#  define C2H_VEC_OVERLOAD_1(T)                                               \
-    /* Ostream output */                                                      \
-    inline std::ostream& operator<<(std::ostream& os, const T& val)           \
-    {                                                                         \
-      os << '(' << CoutCast(val.x) << ')';                                    \
-      return os;                                                              \
-    }                                                                         \
-    /* Inequality */                                                          \
-    inline __host__ __device__ bool operator!=(const T& a, const T& b)        \
-    {                                                                         \
-      return (a.x != b.x);                                                    \
-    }                                                                         \
-    /* Equality */                                                            \
-    inline __host__ __device__ bool operator==(const T& a, const T& b)        \
-    {                                                                         \
-      return (a.x == b.x);                                                    \
-    }                                                                         \
-    /* Max */                                                                 \
-    inline __host__ __device__ bool operator>(const T& a, const T& b)         \
-    {                                                                         \
-      return (a.x > b.x);                                                     \
-    }                                                                         \
-    /* Min */                                                                 \
-    inline __host__ __device__ bool operator<(const T& a, const T& b)         \
-    {                                                                         \
-      return (a.x < b.x);                                                     \
-    }                                                                         \
-    /* Summation (non-reference addends for VS2003 -O3 warpscan workaround */ \
-    inline __host__ __device__ T operator+(T a, T b)                          \
-    {                                                                         \
-      T retval = make_##T(a.x + b.x);                                         \
-      return retval;                                                          \
+#  define C2H_VEC_OVERLOAD_1(T)                                                  \
+    /* Ostream output */                                                         \
+    inline std::ostream& operator<<(std::ostream& os, const T& val)              \
+    {                                                                            \
+      os << '(' << CoutCast(val.x) << ')';                                       \
+      return os;                                                                 \
+    }                                                                            \
+    /* Inequality */                                                             \
+    inline __host__ __device__ constexpr bool operator!=(const T& a, const T& b) \
+    {                                                                            \
+      return (a.x != b.x);                                                       \
+    }                                                                            \
+    /* Equality */                                                               \
+    inline __host__ __device__ constexpr bool operator==(const T& a, const T& b) \
+    {                                                                            \
+      return (a.x == b.x);                                                       \
+    }                                                                            \
+    /* Max */                                                                    \
+    inline __host__ __device__ constexpr bool operator>(const T& a, const T& b)  \
+    {                                                                            \
+      return (a.x > b.x);                                                        \
+    }                                                                            \
+    /* Min */                                                                    \
+    inline __host__ __device__ constexpr bool operator<(const T& a, const T& b)  \
+    {                                                                            \
+      return (a.x < b.x);                                                        \
+    }                                                                            \
+    /* Summation (non-reference addends for VS2003 -O3 warpscan workaround */    \
+    inline __host__ __device__ constexpr T operator+(T a, T b)                   \
+    {                                                                            \
+      using V = decltype(T::x);                                                  \
+      return T{static_cast<V>(a.x + b.x)};                                       \
     }
 
 /**
  * Vector2 overloads
  */
-#  define C2H_VEC_OVERLOAD_2(T)                                               \
-    /* Ostream output */                                                      \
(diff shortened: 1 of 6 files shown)


libcudacxx/include/cuda/__hierarchy/dimensions.h
```diff
@@ -22,7 +32,7 @@ template <class _Tp, size_t... _Extents>
 using dimensions = ::cuda::std::extents<_Tp, _Extents...>;
 
 // not unsigned because of a bug in ::cuda::std::extents
```
pciolkosz: Remove the comment

libcudacxx/include/cuda/__hierarchy/dimensions.h
```diff
@@ -55,19 +64,19 @@ using dimensions_index_type = int;
  *   Extents of the result
  */
 template <class _Tp, size_t... _Extents>
-struct hierarchy_query_result : public dimensions<_Tp, _Extents...>
+struct hierarchy_query_result_org : public dimensions<_Tp, _Extents...>
```
pciolkosz, line 67: Should this type be removed?

libcudacxx/include/cuda/__hierarchy/block_level.h
```diff
@@ -0,0 +1,141 @@
...
+    const auto __dims = dims_as<_Tp>(__level);
+    const auto __idx  = index_as<_Tp>(__level);
+    return static_cast<_Tp>((__idx.z * __dims.y + __idx.y) * __dims.x + __idx.x);
+  }
+
+  // interactions with grid level in hierarchy
+
+  _CCCL_TEMPLATE(class _Tp, class _Hierarchy)
+  _CCCL_REQUIRES(::cuda::std::__cccl_is_integer_v<_Tp> _CCCL_AND __is_hierarchy_v<_Hierarchy>)
+  [[nodiscard]] _CCCL_DEVICE_API static _Tp rank_as(const grid_level& __level, const _Hierarchy& __hier) noexcept
```
pciolkosz, line 124: Is this needed just as an optimization?

libcudacxx/include/cuda/__hierarchy/hierarchy_level_base.h
```diff
@@ -0,0 +1,330 @@
...
+                   __is_hierarchy_v<_Hierarchy>)
+  [[nodiscard]] _CCCL_API static constexpr auto dims_as(const _InLevel& __level, const _Hierarchy& __hier) noexcept
+  {
+    return __dims_as_impl<_Tp>(__level, __hier);
+  }
+
+  _CCCL_TEMPLATE(class _Tp, class _InLevel, class _Hierarchy)
+  _CCCL_REQUIRES(::cuda::std::__cccl_is_integer_v<_Tp> _CCCL_AND __is_hierarchy_level_v<_InLevel> _CCCL_AND
+                   __is_hierarchy_v<_Hierarchy>)
+  [[nodiscard]] _CCCL_API static constexpr auto extents_as(const _InLevel& __in_level, const _Hierarchy& __hier) noexcept
```
pciolkosz, line 184: Why not just return __hier.extents(_Level{}, _InLevel{})?

libcudacxx/include/cuda/__hierarchy/hierarchy_query_result.h
```diff
@@ -0,0 +1,149 @@
...
+    }
+    else
+    {
+      return z;
+    }
+  }
+
+  _CCCL_TEMPLATE(class _Tp2 = _Tp)
+  _CCCL_REQUIRES(::cuda::std::is_same_v<_Tp2, signed char>)
+  _CCCL_API constexpr operator char3() const noexcept
```
pciolkosz, line 76: we really need all those operators?

libcudacxx/test/libcudacxx/cuda/hierarchy/traits/is_natively_reachable_hierarchy_level_v.compile.pass.cpp
```diff
@@ -0,0 +1,59 @@
...
+#include <cuda/hierarchy>
+
+struct MyLevel : cuda::hierarchy_level_base<MyLevel>
+{};
+
+template <class FromLevel, class ToLevel>
+inline constexpr bool trait_v = cuda::__is_natively_reachable_hierarchy_level_v<FromLevel, ToLevel>;
+
+static_assert(!trait_v<cuda::thread_level, cuda::thread_level>);
+static_assert(!trait_v<cuda::thread_level, cuda::warp_level>);
```
pciolkosz, line 22: think a query like that should work no?

libcudacxx/include/cuda/__hierarchy/dimensions.h
davebayer: In next PR, this one is already huge

libcudacxx/include/cuda/__hierarchy/hierarchy_level_base.h
davebayer: Because I thought we agreed that we will be removing these queries.. Or not?

libcudacxx/include/cuda/__hierarchy/hierarchy_query_result.h
davebayer: Yes, because if I do just conversion operator to `__vector_type_t<_Tp, 3>`, there are some problems with implicit conversions with older compilers. And because it's only 1 type, I decided to just do it manually.

libcudacxx/test/libcudacxx/cuda/hierarchy/traits/is_natively_reachable_hierarchy_level_v.compile.pass.cpp
davebayer: Oh it works, but only for the native queries, that means: `cuda::gpu_thread.index(cuda::warp)` without the hierarchy. In the future, I'd like to add a feature to allow the user to query all native levels in a hierarchy, if the hierarchy has the other hierarchy levels needed for the computation.

For example, you would be able to query `cuda::warp.index(cuda::block, hier)` in a hierarchy that has block and thread level in it

libcudacxx/include/cuda/__hierarchy/block_level.h
davebayer: Yes, cg does this, too

libcudacxx/include/cuda/__hierarchy/hierarchy_level_base.h
pciolkosz: We could just make them internal for now if it's easier. But I guess this is already implemented so 🤷 . As you prefer

PR #6961, opened 2025-12-11
Disable `__builtin_structured_binding_size` with nvcc

davebayer:
It fails with clang-21 + nvcc 13.1


bernhardmgruber:
> clang-21 + nvcc 13.1

So can we just disable it for the offending compiler versions?

davebayer:
> So can we just disable it for the offending compiler versions?

You mean for nvcc < 13.1? The builtin is completely new, I don't expect nvcc to support this builtin anytime soon

bernhardmgruber:
> The builtin is completely new, I don't expect nvcc to support this builtin anytime soon

That's the part I was missing. Thx!

PR #6963, opened 2025-12-11
Add support for ops that capture device arrays from enclosing scope

shwina:
## Description

This PR adds support for ops that can capture state. This can be used to implement algorithms wide side-effects:

```python
from cuda.compute import select
import numpy as np
import cupy as cp
import numba.cuda as numba_cuda


# Create device state for counting rejected items
reject_counter = cp.zeros(1, dtype=np.int32)

# Define condition that references state as closure
def count_rejects(x):
    if x > 50:
        return True
    else:
        numba_cuda.atomic.add(reject_counter, 0, 1)
        return False

d_in = cp.arange(100, dtype=np.int32)
d_out = cp.empty_like(d_in)
d_num_selected = cp.empty(2, dtype=np.uint64)

select(
    d_in,
    d_out,
    d_num_selected,
    count_rejects,
    len(d_in),
)

print(reject_counter)
```

prints:

```
[51]
```

It works by extending numba-CUDA to recognize device-array-like objects, and teaching it how to capture them from enclosing scope using `lower_constant`.

## Checklist

### DIFF
--- a/python/cuda_cccl/cuda/compute/_caching.py
+++ b/python/cuda_cccl/cuda/compute/_caching.py
@@ -69,6 +69,14 @@ def clear_all_caches():
         cached_func.cache_clear()
 
 
+def _hash_device_array_like(value):
+    # hash based on pointer, shape, and dtype
+    ptr = value.__cuda_array_interface__["data"][0]
+    shape = value.__cuda_array_interface__["shape"]
+    dtype = value.__cuda_array_interface__["typestr"]
+    return hash((ptr, shape, dtype))
+
+
 def _make_hashable(value):
     import numba.cuda.dispatcher
 
@@ -77,7 +85,7 @@ def _make_hashable(value):
     if isinstance(value, numba.cuda.dispatcher.CUDADispatcher):
         return CachableFunction(value.py_func)
     elif isinstance(value, DeviceArrayLike):
-        return id(value)
+        return _hash_device_array_like(value)
     elif isinstance(value, (list, tuple)):
         return tuple(_make_hashable(v) for v in value)
     elif isinstance(value, dict):


python/cuda_cccl/cuda/compute/_caching.py
```diff
@@ -77,7 +77,11 @@ def __eq__(self, other):
         return self._identity == other._identity
 
     def __hash__(self):
-        return hash(self._identity)
+        try:
+            return hash(self._identity)
+        except TypeError:
+            # if we can't hash the identity, use its id
+            return id(self._identity)
```
bdice: I recommend hashing the resulting `id` value.
```suggestion
            return hash(id(self._identity))
```

python/cuda_cccl/cuda/compute/_device_array_capture.py
```diff
@@ -0,0 +1,197 @@
...
+            ("itemsize", types.intp),
+            ("data", types.CPointer(fe_type.dtype)),
+            ("shape", types.UniTuple(types.intp, ndim)),
+            ("strides", types.UniTuple(types.intp, ndim)),
+        ]
+        super().__init__(dmm, fe_type, members)
+
+
+# =============================================================================
+# Step 5: Register lower_constant for our type
```
bdice, line 143: These step numbers don't align with the 1, 2, 3 at the top of the file. Maybe these comments aren't essential?

python/cuda_cccl/cuda/compute/_caching.py
oleksandr-pavlyk: In CPython these two things give identical values, although spec does not require them to be:

```
Python 3.13.3 | packaged by conda-forge | (main, Apr 14 2025, 20:44:03) [GCC 13.3.0]
Type 'copyright', 'credits' or 'license' for more information
IPython 9.4.0 -- An enhanced Interactive Python. Type '?' for help.
Tip: You can use LaTeX or Unicode completion, `\alpha<tab>` will insert the α symbol.

In [1]: o = list()

In [2]: id(o), hash(id(o))
Out[2]: (138960802540160, 138960802540160)

In [3]: %timeit hash(id(o))
36.7 ns ± 0.17 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [4]: %timeit id(o)
21.7 ns ± 0.909 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```

python/cuda_cccl/cuda/compute/_device_array_capture.py
```diff
@@ -0,0 +1,197 @@
...
+    Type for any DeviceArrayLike object.
+
+    By subclassing Array and overriding copy() to ignore readonly,
+    we ensure these arrays remain mutable when captured from globals.
+    """
+
+    def __init__(self, dtype, ndim, layout, readonly=False, aligned=True):
+        type_name = "device_array"
+        if readonly:
+            type_name = "readonly " + type_name
```
gmarkall, line 37: For the parent class, "unaligned" also appears in the name if `aligned` is `False`. Is there a guarantee that things of `DeviceArrayType` are aligned?

python/cuda_cccl/cuda/compute/_device_array_capture.py
```diff
@@ -0,0 +1,197 @@
...
+# =============================================================================
+# Step 3: Patch typeof_impl to handle CAI objects
+# =============================================================================
+
+# Get the original generic fallback
+_original_typeof_impl = typeof_module.typeof_impl
+_original_generic = _original_typeof_impl.dispatch(object)
+
+
+def _patched_generic(val, c):
```
gmarkall, line 117: Numba-CUDA already has a mechanism for typing CAI object, but it lived in the CUDA dispatcher: https://github.com/NVIDIA/numba-cuda/blob/94321646a046ea5dd1b64e8dc078e27121552e23/numba_cuda/numba/cuda/dispatcher.py#L1626-L1643

This was because `typeof` came from upstream Numba, but I think we can push this check into `typeof` in Numba-CUDA now that it contains its own `typeof` implementation.

python/cuda_cccl/cuda/compute/_device_array_capture.py
```diff
@@ -0,0 +1,197 @@
...
+            aligned=self.aligned,
+        )
+
+
+# =============================================================================
+# Step 2: Define _typeof_device_array to type DeviceArrayLike objects
+# =============================================================================
+
+
+def _typeof_device_array(val, c):
```
gmarkall, line 62: When typing an argument, this will be called with `c.purpose == Purpose.constant` (from `numba.cuda.typing.typeof.Purpose`). Perhaps we don't want to affect typing of arguments - in which case we should return `None` when `c.purpose == Purpose.argument`.

gmarkall: Also - I haven't read through this implementation, but it's probably faster than the implementation in `Dispatcher.typeof_pyval`, so we should probably try if upstreaming it into `typeof_pyval` to see if it makes kernel launch a bit faster for CAI arrays (cc @cpcloud)

python/cuda_cccl/cuda/compute/_device_array_capture.py
```diff
@@ -0,0 +1,197 @@
...
+        members = [
+            ("meminfo", types.MemInfoPointer(fe_type.dtype)),
+            ("parent", types.pyobject),
+            ("nitems", types.intp),
+            ("itemsize", types.intp),
+            ("data", types.CPointer(fe_type.dtype)),
+            ("shape", types.UniTuple(types.intp, ndim)),
+            ("strides", types.UniTuple(types.intp, ndim)),
+        ]
+        super().__init__(dmm, fe_type, members)
```
gmarkall: This looks the same as the `ArrayModel`, so I think we could just register it:

```suggestion
register_default(DeviceArrayType)(ArrayModel)
```

python/cuda_cccl/cuda/compute/_device_array_capture.py
shwina: Hmm, maybe this is a question for @leofang ? Do CAI (and DLPack) make any guarantees about alignment?

leofang: I don't think either one guarantees or constrains about alignment.

shwina:
/ok to test https://github.com/NVIDIA/cccl/pull/6963/commits/12f18992ed0aa0fa899d2e76302be2bf1319c5d6

shwina:
/ok to test https://github.com/NVIDIA/cccl/pull/6963/commits/8d6faa0383fff5899f9e047551e29c626a0247d2

python/cuda_cccl/cuda/compute/_device_array_capture.py
shwina: I've defaulted to `align=False` and added "unaligned" to the type name.

python/cuda_cccl/cuda/compute/_device_array_capture.py
shwina: Fixed

python/cuda_cccl/cuda/compute/_device_array_capture.py
shwina: Done

python/cuda_cccl/cuda/compute/_device_array_capture.py
```diff
@@ -0,0 +1,198 @@
...
+
+@builtin_registry.lower_constant(DeviceArrayType)
+def lower_device_array(context, builder, ty, pyval):
+    """
+    Lower DeviceArrayLike objects by embedding the device pointer as a constant.
+    """
+
+    interface = pyval.__cuda_array_interface__
+
+    # hold on to the device-array-like object
```
shwina, line 155: @gmarkall - added your code snippet for keeping references to the arrays here.

python/cuda_cccl/cuda/compute/_device_array_capture.py
```diff
@@ -0,0 +1,198 @@
...
+            # Check F-contiguous
+            f_strides = []
+            stride = itemsize
+            for i in range(ndim):
+                f_strides.append(stride)
+                stride *= shape[i]
+            if tuple(strides) == tuple(f_strides):
+                layout = "F"
+            else:
+                layout = "A"
```
oleksandr-pavlyk, line 102: Perhaps one can do a quick rejection test to avoid reconstruction of entire `c_strides` or `f_strides`. 
If neither last nor first element of `strides` sequence equals `1`, we can set `layout = "A"`.

shwina:
Closing this PR, as I've upstreamed the bulk of the work to numba-cuda (merged): https://github.com/NVIDIA/numba-cuda/pull/666

I opened https://github.com/NVIDIA/cccl/pull/7008 as a follow-up to that, to add the minor changes needed to `cuda-compute`.

PR #6964, opened 2025-12-11
Add `_CCCL_DEDUCTION_GUIDE_EXSPACE`

davebayer:
clang-cuda has always forced us to add `__host__ __device__` annotations to deduction guides. However, since clang-22, this is no longer the case + clang emits ton of warnings about this behaviour being deprecated.

This PR introduces `_CCCL_DEDUCTION_GUIDE_EXSPACE` macro that expands to `_host__ __device__` for clang < 22, otherwise is empty.

### DIFF
--- a/cub/cub/detail/segmented_scan_helpers.cuh
+++ b/cub/cub/detail/segmented_scan_helpers.cuh
@@ -32,7 +32,7 @@ struct augmented_value_t
 };
 
 template <typename ValueT, typename FlagT = bool>
-_CCCL_HOST_DEVICE augmented_value_t(ValueT, FlagT) -> augmented_value_t<ValueT, FlagT>;
+_CCCL_DEDUCTION_GUIDE_ATTRIBUTES augmented_value_t(ValueT, FlagT) -> augmented_value_t<ValueT, FlagT>;
 
 template <typename ComputeT, int MaxSegmentsPerWorker>
 using agent_segmented_scan_compute_t =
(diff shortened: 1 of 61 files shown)


fbusato:
I'm a bit hesitant because llvm-22 is still under development. If something change, we need to address the issue. Is the behavior reported on some docs at least?

davebayer:
> I'm a bit hesitant because llvm-22 is still under development. If something change, we need to address the issue. Is the behavior reported on some docs at least?

Aaah, you are right, the change was added to clang last week :D https://github.com/llvm/llvm-project/commit/03e42c97c7c0f6df51e3298cafa60b1f1e59c0b9

But it's well documented and I don't expect it to be changed.

cub/cub/detail/segmented_scan_helpers.cuh
```diff
@@ -32,7 +32,7 @@ struct augmented_value_t
 };
 
 template <typename ValueT, typename FlagT = bool>
-_CCCL_HOST_DEVICE augmented_value_t(ValueT, FlagT) -> augmented_value_t<ValueT, FlagT>;
+_CCCL_DEDUCTION_GUIDE_EXSPACE augmented_value_t(ValueT, FlagT) -> augmented_value_t<ValueT, FlagT>;
```
bernhardmgruber: What does the term `EXSPACE` mean here?

Could we name this: `_CCCL_DEDUCTION_GUIDE_ATTRIBUTES`?

miscco (approved):

cub/cub/detail/segmented_scan_helpers.cuh
fbusato: +1 for `_CCCL_DEDUCTION_GUIDE_ATTRIBUTES`.
 I'm not sure what `EXSPACE` means

davebayer: Execution space

fbusato: not so intuitive. Maybe `EXEC_SPACE` would be better. Also, could please add it to the documentation

davebayer: Btw I went with Bernhard's suggestion :) good point, I will do that in another pr

PR #6979, opened 2025-12-15
Improve `auto(expr)` support

davebayer:
gcc and nvhpc support `auto(expr)` even before c++23 as an extension, nvcc supports this since 13.0.

This renames `_LIBCUDACXX_AUTO_CAST` to `_CCCL_AUTO_CAST` and allows more efficient implementation via `auto(expr)` even in earlier c++ modes.

### DIFF
--- a/libcudacxx/include/cuda/std/__cccl/prologue.h
+++ b/libcudacxx/include/cuda/std/__cccl/prologue.h
@@ -345,4 +345,9 @@ _CCCL_DIAG_SUPPRESS_NVCC(3206) // "if consteval" and "if not consteval" are mean
 _CCCL_DIAG_SUPPRESS_NVCC(3060) // call to __builtin_is_constant_evaluated appearing in a non-constexpr function always
                                // produces "false"
 
+// suppress warnings about auto(expr)
+_CCCL_DIAG_SUPPRESS_NVHPC(auto_cast_is_cpp23)
+
+_CCCL_DIAG_SUPPRESS_NVCC(3347)
+
 // NO include guards here (this file is included multiple times)
--- a/libcudacxx/include/cuda/std/__utility/auto_cast.h
+++ b/libcudacxx/include/cuda/std/__utility/auto_cast.h
@@ -1,10 +1,9 @@
-// -*- C++ -*-
 //===----------------------------------------------------------------------===//
 //
 // Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
 // See https://llvm.org/LICENSE.txt for license information.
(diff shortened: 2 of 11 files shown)


libcudacxx/include/cuda/std/__cccl/prologue.h
```diff
@@ -288,25 +288,29 @@ _CCCL_NV_DIAG_PUSH()
 // warning C4996: 'meow': was declared deprecated
 _CCCL_DIAG_SUPPRESS_MSVC(4100 4127 4180 4197 4296 4324 4455 4503 4522 4668 4800 4996)
 
-// disable warnings about using C++23 features in C++20 (needed for if consteval)
-#if _CCCL_HAS_IF_CONSTEVAL_IN_CXX20()
```
ericniebler: i wonder if we even need the `_CCCL_HAS_IF_CONSTEVAL_IN_CXX20` macro anymore. its only remaining use will be in `__cccl/dialect.h`.

ericniebler (approved):

libcudacxx/include/cuda/std/__utility/auto_cast.h
```diff
@@ -23,12 +22,24 @@
...
+#endif // ^^^ no auto(expr) before c++23 ^^^
+
+// nvcc < 13 does not support auto(expr), nvhpc fails to use auto(expr) in noexcept(noexcept(...)) (nvbug 5742468)
+#if _CCCL_CUDA_COMPILER(NVCC, <, 13) || _CCCL_COMPILER(NVHPC)
+#  undef _CCCL_HAS_AUTO_CAST_BEFORE_CXX23
+#  define _CCCL_HAS_AUTO_CAST_BEFORE_CXX23() 0
+#endif // _CCCL_CUDA_COMPILER(NVCC, <, 13)
+
+#if (_CCCL_STD_VER >= 2023 && __cpp_auto_cast >= 202110L) || _CCCL_HAS_AUTO_CAST_BEFORE_CXX23()
+#  define _CCCL_AUTO_CAST(expr) auto(expr)
```
davebayer: @ericniebler this doesn't seem to be 100 % equivalent to `static_cast<::cuda::std::decay_t<decltype((expr))>>(expr)`, I'm getting some failures with gcc. I had a look at libc++ and they use the cast, too, even though they have access to `auto(x)`. I think we should disable it completely, since we will enable C++23 one day and this would be blocking the CI

ericniebler: i can make a pr to change `_CCCL_AUTO_CAST` to always do the cast.

Issue #6864, opened 2025-12-03
Add `DispatchSegmentedTopK` that's flexible enough to cover arbitrary per-segment and uniform parameter combinations


elstehle:
### Context
[https://github.com/NVIDIA/cccl/issues/5673](https://github.com/NVIDIA/cccl/issues/5673) lists an extensive feature matrix for `DeviceSegmentedTopK`. We need a `DispatchSegmentedTopK` layer that is flexible enough to cover this wide range of use cases across different domains, while simultaneously allowing optimum performance for targeted workloads.

### Goals
There are two main goals we want to achieve with a flexible `DispatchSegmentedTopK` design:
1.  **Arbitrary Parameter Mixtures:** Support any combination of segment-specific parameters and parameters that are uniform across all segments.
2.  **Targeted Optimization:** Enable peak performance for workloads with narrow constraints (e.g., a fixed segment size of 512 and `K=128`).

### Segment-Specific vs. Uniform vs. Static Parameters
Segmented Top-K depends on three key parameters: `k` (items to select), `segment size`, and `sort direction`. To support the goals above, each of these must be specifiable in three ways:

1.  **Uniform:** The same value applies to all segments (e.g., "Get top-5 for all segments").
2.  **Segment-specific:** Values vary per segment (e.g., "Segment 0 needs top-5, Segment 1 needs top-2").
3.  **Static:** Values are known at compile-time (e.g., "K is always 32").

Additionally, users should be able to provide the `number of segments` as either a device-accessible iterator or a host value. Furthermore, if users can guarantee an upper bound on the total number of items (summed across all segments), we can optimize the allocation of temporary buffers for intermediate results (similar to CUB's existing sorting algorithms).

### Optimal Performance for Targeted Workloads
LLMs (e.g., MoE and Sparse Attention models) often operate under narrow constraints, such as uniform segment sizes of exactly 512 items or a fixed `K=128`. We need a mechanism for users to communicate these constraints to the algorithm.

By establishing a flexible interface now, we allow users to provide "hints" (e.g., `MaxK=1024`) to benefit from future kernel specializations without needing to refactor their algorithm invocations later.

**Examples of Constraint-Based Optimization:**
* **Small K (Reduce-based):** Knowing that `K` is in the range `[1, 8]` suggests a reduce-based Top-K algorithm is optimal. If this information is available at compile-time, we can compile *only* the reduce-based code path, reducing compilation time, binary size, and potentially register pressure.
[...]

Issue #6863, opened 2025-12-03
Add tests for `DeviceSegmentedTopK` with fixed-size segments


elstehle:
The goal here is to have an initial set of tests so we can verify correctness of `DeviceSegmentedTopK` implementation with fixed-size segments. To enable these tests, the idea is to back the implementation with a stub implementation that uses `BlockRadixSort`.

Issue #6508, opened 2025-11-05
Prototype benchmarks for `DeviceSegmentedTopK` with fixed `k` and `segment_size` across segments


elstehle:
`DeviceSegmentedTopK` with a fixed `k` and fixed `segment_size` across segments is a very common use case in LLMs for mixture-of-experts (MoE) and, more recently, also sparse attention, e.g., in DeepSeek-V3.2. 

In the context of LLMs, `segment_size`  usually falls into the range of 4k to 128k items, which requires both, a block-level as well as a multi-block approach. 

In order to determine an efficient implementation, we need to have proper benchmarking in place. The goal of this issue is to have a dev branch that allows comparing various state-of-the-art segmented top-k implementations.

Issue #6290, opened 2025-10-20
Add support for non-fundamental types to `DeviceTopK`


elstehle:
Our current implementation of `DeviceTopK` is radix partitioning-based and, similar to `DeviceRadixSort` relies on the `cub::Traits` for bit-twiddling and related functionality. As a result, the implementation is currently limited to support only fundamental types.

There's existing interest from internal and external users to extend support for custom types.

PR #6980, opened 2025-12-15
Initial version of `DeviceSegmentedTopk` for fixed-size segments

elstehle:
Closes https://github.com/NVIDIA/cccl/issues/6508
Closes https://github.com/NVIDIA/cccl/issues/6863
Closes https://github.com/NVIDIA/cccl/issues/6864

### DIFF
--- a/cub/cub/agent/agent_batched_topk.cuh
+++ b/cub/cub/agent/agent_batched_topk.cuh
@@ -199,7 +199,22 @@ struct agent_batched_topk_worker_per_segment
     __syncthreads();
 
     // Perform Block Top-K
-    if constexpr (!is_keys_only)
+    if constexpr (is_keys_only)
+    {
+      const bool is_successful_dispatch =
+        detail::params::dispatch_discrete(select_directions, segment_id, [this, &thread_keys, k](auto direction_tag) {
+          if constexpr (decltype(direction_tag)::value == detail::topk::select::max)
+          {
+            block_topk_t(temp_storage.topk).max_keys(thread_keys, k);
+          }
+          else
+          {
+            block_topk_t(temp_storage.topk).min_keys(thread_keys, k);
+          }
+        });
+      _CCCL_ASSERT(is_successful_dispatch, "Error: Unsupported select direction");
+    }
+    else
     {
       // Pass both keys and values
       const bool is_successful_dispatch = detail::params::dispatch_discrete(
@@ -211,21 +226,6 @@ struct agent_batched_topk_worker_per_segment
           else
           {
             block_topk_t(temp_storage.topk).min_pairs(thread_keys, thread_values, k);
-          }
-        });
-      _CCCL_ASSERT(is_successful_dispatch, "Error: Unsupported select direction");
-    }
-    else
-    {
-      const bool is_successful_dispatch =
-        detail::params::dispatch_discrete(select_directions, segment_id, [this, &thread_keys, k](auto direction_tag) {
-          if constexpr (decltype(direction_tag)::value == detail::topk::select::max)
-          {
-            block_topk_t(temp_storage.topk).max_keys(thread_keys, k);
-          }
-          else
-          {
-            block_topk_t(temp_storage.topk).min_keys(thread_keys, k);
           }
         });
       _CCCL_ASSERT(is_successful_dispatch, "Error: Unsupported select direction");


cub/cub/agent/agent_segmented_topk.cuh
```diff
@@ -0,0 +1,247 @@
...
+
+  using BlockTopkT = BlockTopK<key_t, block_threads, items_per_thread, value_t>;
+
+  using BlockStoreKeysT = BlockStore<key_t, block_threads, items_per_thread, BLOCK_STORE_WARP_TRANSPOSE>;
+  using BlockStoreValsT = BlockStore<value_t, block_threads, items_per_thread, BLOCK_STORE_WARP_TRANSPOSE>;
+
+  // -------------------------------------------------------------------------
+  // Shared Memory Storage
+  // -------------------------------------------------------------------------
+  struct TempStorage
```
oleksandr-pavlyk: In other parts of CUB, we define `union _TempStorage {...};` and `using TempStorage = Uninitialized<_TempStorage>;`. 

Should it be done here as well?

cub/cub/agent/agent_segmented_topk.cuh
```diff
@@ -0,0 +1,247 @@
...
+  // -------------------------------------------------------------------------
+  // Primitive Types
+  // -------------------------------------------------------------------------
+  using BlockLoadKeysT = BlockLoad<key_t, block_threads, items_per_thread, BLOCK_LOAD_WARP_TRANSPOSE>;
+  using BlockLoadValsT = BlockLoad<value_t, block_threads, items_per_thread, BLOCK_LOAD_WARP_TRANSPOSE>;
+
+  using BlockTopkT = BlockTopK<key_t, block_threads, items_per_thread, value_t>;
+
+  using BlockStoreKeysT = BlockStore<key_t, block_threads, items_per_thread, BLOCK_STORE_WARP_TRANSPOSE>;
+  using BlockStoreValsT = BlockStore<value_t, block_threads, items_per_thread, BLOCK_STORE_WARP_TRANSPOSE>;
```
oleksandr-pavlyk: Per coding guideline, these should follow snake case naming convention.

cub/cub/device/dispatch/dispatch_batched_topk.cuh
```diff
@@ -0,0 +1,499 @@
...
+  void* d_temp_storage;
+
+  /// Reference to size in bytes of `d_temp_storage` allocation
+  size_t& temp_storage_bytes;
+
+  /// d_key_segments_it[segment_index] -> iterator to the input sequence of key data for segment `segment_index`
+  KeyInputItItT d_key_segments_it;
+
+  /// d_key_segments_out_it[segment_index] -> iterator to the output sequence of key data for segment `segment_index`
+  KeyOutputItItT d_key_segments_out_it;
```
gevtushenko, line 167: important: this feels closer to semantics of batched algorithm rather then segmented one. To me, "segmented" means that there a single entity (contiguous input) that's divided into parts (segments). Other interfaces like segmented reduce, sort, scan, etc. use begin / end offsets to express this. On the other hand, "batched" means that there's a set of independent problems that we try to solve together. If your use case depands iterator of iterators, I'd consider renaming the algorithm to batched top k. If begin / end offsets suffice, I'd rather have a single input / output iterator accompanied by begin / end offsets.

cub/benchmarks/bench/segmented_topk/keys.cu
```diff
@@ -0,0 +1,179 @@
...
+  const auto elements          = num_segments * segment_size;
+  const auto total_num_items   = total_num_items_guarantee_t{static_cast<::cuda::std::int64_t>(elements)};
+  const bit_entropy entropy    = str_to_entropy(state.get_string("Entropy"));
+
+  // Skip workloads where k exceeds the segment size
+  if (selected_elements >= segment_size)
+  {
+    state.skip("Skipping workload where K >= SegmentSize.");
+    return;
+  }
```
gevtushenko, line 104: question: I'm not sure if skipping is supported by tuning infrastructure. Any chance we could re formulate num selected to % like 10%, 50% etc?

cub/cub/agent/agent_segmented_topk.cuh
```diff
@@ -0,0 +1,247 @@
+// SPDX-FileCopyrightText: Copyright (c) 2025, NVIDIA CORPORATION. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+//! @file
+//! cub::AgentTopK implements a stateful abstraction of CUDA thread blocks for participating in device-wide topK.
```
gevtushenko: suggestion: not sure how doxygen treats duplicates here, let's rename to `cub::AgentSegmentedTopK...`

cub/cub/agent/agent_segmented_topk.cuh
```diff
@@ -0,0 +1,247 @@
...
+
+#include <cub/block/block_load.cuh>
+#include <cub/block/block_scan.cuh>
+#include <cub/block/block_store.cuh>
+#include <cub/block/block_topk.cuh>
+#include <cub/detail/segmented_params.cuh>
+#include <cub/device/dispatch/dispatch_common.cuh>
+#include <cub/util_type.cuh>
+
+#include <cuda/atomic>
```
gevtushenko: question: atomic header might be limiting pre-volta architectures for CTK 12.0. Is anything in this file using it? I see non-segmented TopK also including this but using `atomicAdd`. Any chance we can drop the include?

cub/benchmarks/bench/segmented_topk/keys.cu
```diff
@@ -0,0 +1,179 @@
...
+#endif
+
+#if !TUNE_BASE
+template <class KeyInT, class OffsetT>
+struct policy_hub_t
+{
+  struct policy_t : cub::ChainedPolicy<300, policy_t, policy_t>
+  {
+    static constexpr int nominal_4b_items_per_thread = TUNE_ITEMS_PER_THREAD;
+    static constexpr int items_per_thread = cuda::std::max(1, (nominal_4b_items_per_thread * 4 / sizeof(KeyInT)));
```
gevtushenko: suggestion: agent and tuning for segmented top k do not scale items per thread. This might lead to a confusion, when tuning says that `ITP=12` is optimal, and then 12 is passed to the agent instead of scaled version of it. As a rule of thumb, we can avoid confusion going forward if make sure to use wantever parameter is used in the tuning process in the encoding of the tuning as well. We can either remove scaling here, or move it inside the agent to achieve this.

cub/benchmarks/bench/segmented_topk/keys.cu
```diff
@@ -0,0 +1,179 @@
...
+  // Static selection direction (max)
+  using select_direction_value_t = cub::detail::segmented_topk::select_direction_static<cub::detail::topk::select::max>;
+
+  // Number of segments is a host-accessible value
+  using num_segments_uniform_t = cub::detail::segmented_topk::num_segments_uniform<>;
+
+  // Total number of items guarantee type
+  using total_num_items_guarantee_t =
+    cub::detail::segmented_topk::total_num_items_guarantee<min_num_total_items, max_num_total_items>;
+  using offset_t     = ::cuda::std::int32_t;
```
gevtushenko: suggestion: unused? Also, there's `OffsetT`

cub/cub/block/block_topk.cuh
```diff
@@ -0,0 +1,83 @@
...
+
+public:
+  struct TempStorage
+  {
+    typename BlockRadixSortT::TempStorage sort_storage;
+  };
+
+private:
+  TempStorage& temp_storage;
+  int linear_tid;
```
gevtushenko: question: can we drop it?

cub/cub/block/block_topk.cuh
```diff
@@ -0,0 +1,83 @@
...
+public:
+  _CCCL_DEVICE _CCCL_FORCEINLINE BlockTopK(TempStorage& temp_storage)
+      : temp_storage(temp_storage)
+      , linear_tid(threadIdx.x)
+  {}
+
+  _CCCL_DEVICE _CCCL_FORCEINLINE void
+  Max(KeyT (&keys)[ItemsPerThread],
+      ValueT (&values)[ItemsPerThread],
+      int k,
```
gevtushenko: suggestion: to avoid warnings about unused parameters here and below:
```suggestion
      int /* k */,
```

cub/cub/agent/agent_segmented_topk.cuh
```diff
@@ -0,0 +1,247 @@
...
+
+  // -------------------------------------------------------------------------
+  // Primitive Types
+  // -------------------------------------------------------------------------
+  using BlockLoadKeysT = BlockLoad<key_t, block_threads, items_per_thread, BLOCK_LOAD_WARP_TRANSPOSE>;
+  using BlockLoadValsT = BlockLoad<value_t, block_threads, items_per_thread, BLOCK_LOAD_WARP_TRANSPOSE>;
+
+  using BlockTopkT = BlockTopK<key_t, block_threads, items_per_thread, value_t>;
+
+  using BlockStoreKeysT = BlockStore<key_t, block_threads, items_per_thread, BLOCK_STORE_WARP_TRANSPOSE>;
```
gevtushenko: question: we sometimes know statically how many items we are going to store back. Is there any perf benefit in using this knowledge here instead of passing runtime num valid or compiler manages to optimize this out? If so, it might deserve a comment

cub/cub/agent/agent_segmented_topk.cuh
```diff
@@ -0,0 +1,247 @@
...
+  _CCCL_DEVICE _CCCL_FORCEINLINE void select_topk_keys(
+    key_t (&keys)[items_per_thread], KValueT k, ::cuda::std::integral_constant<detail::topk::select, Direction>)
+  {
+    if constexpr (Direction == detail::topk::select::max)
+    {
+      BlockTopkT(temp_storage.topk).Max(keys, k);
+    }
+    else
+    {
+      BlockTopkT(temp_storage.topk).Min(keys, k);
```
gevtushenko: question: to my understanding, this always reverses the order of elements in the output. Is there any way to forward "stability" parameter here and static assert with a comment of current behavior?

cub/cub/detail/segmented_params.cuh
```diff
@@ -0,0 +1,250 @@
...
+CUB_NAMESPACE_BEGIN
+
+namespace detail::params
+{
+// -----------------------------------------------------------------------------
+// Parameter Mixins and Helpers
+// -----------------------------------------------------------------------------
+
+// Allows providing constrains on parameter values at compile time
+template <typename T, T Min = ::cuda::std::numeric_limits<T>::min(), T Max = ::cuda::std::numeric_limits<T>::max()>
```
gevtushenko: question: should it be lowest instead of min here?

cub/cub/device/dispatch/dispatch_segmented_topk.cuh
```diff
@@ -0,0 +1,499 @@
...
+template <typename ChainedPolicyT,
+          typename KeyInputItItT,
+          typename KeyOutputItItT,
+          typename ValueInputItItT,
+          typename ValueOutputItItT,
+          typename SegmentSizeParameterT,
+          typename KParameterT,
+          typename SelectDirectionParameterT,
+          typename NumSegmentsParameterT>
+__launch_bounds__(int()) __global__ void DeviceSegmentedTopKKernel(
```
gevtushenko: question: I'm not familiar with specifying 0 for launch bounds. Is the behavior documented anywhere? What's the intention here?

cub/cub/device/dispatch/dispatch_segmented_topk.cuh
```diff
@@ -0,0 +1,499 @@
...
+
+    // TODO (elstehle): support number of segments provided by device-accessible iterator
+    // Only uniform number of segments are supported (i.e., we  need to resolve the number of segments on the host)
+    static_assert(!params::is_per_segment_param_v<NumSegmentsParameterT>,
+                  "Only uniform segment sizes are currently supported.");
+
+    // TODO (elstehle): support larger number of segments through multiple kernel launches
+    int grid_dim = static_cast<int>(resolve_param(num_segments, 0));
+
+    THRUST_NS_QUALIFIER::cuda_cub::detail::triple_chevron(grid_dim, block_dim, 0, stream)
```
gevtushenko: suggestion: tripple chevron can return an error that's coming from a launch failure and isn't reported by subsequent keep at last error. I'd suggest to catch it with `auto error = CubDebug(THRUST_NS_QUALIFIER::cuda_cub...`.

cub/cub/device/dispatch/dispatch_batched_topk.cuh
```diff
@@ -0,0 +1,499 @@
...
+        select_directions,
+        num_segments);
+
+    // Check for failure to launch
+    if (const auto error = CubDebug(cudaPeekAtLastError()))
+    {
+      return error;
+    }
+
+    return cudaSuccess;
```
gevtushenko, line 263: suggestion: missing:

```
      // Sync the stream if specified to flush runtime errors
      error = CubDebug(detail::DebugSyncStream(stream));
      if (cudaSuccess != error)
      {
        return error;
      }
```

cub/test/catch2_test_device_segmented_topk_keys.cu
```diff
@@ -0,0 +1,258 @@
...
+      segment_sizes,
+      k,
+      select_directions,
+      num_segments,
+      total_num_items_guarantee,
+      stream);
+}
+
+// %PARAM% TEST_LAUNCH lid 0:1:2
+DECLARE_LAUNCH_WRAPPER(dispatch_segmented_topk_keys, segmented_topk_keys);
```
gevtushenko: question: are we missing a test for pairs?

cub/cub/agent/agent_segmented_topk.cuh
bernhardmgruber: My latest info is that we do not support any architecture below sm75. We just try not to break older arches if we don't have to.

cub/cub/detail/segmented_params.cuh
elstehle: We have decided to go with the naming convention `Min` and `Max` in [#5674](https://github.com/NVIDIA/cccl/issues/5674), when agreeing on the interface for `DeviceTopK`. The primary motivation is consistency with other _algorithms_:

* Generalization of Reduce: Just as `DeviceReduce::{Min,Max}` selects the single item with the lowest/highest value (`k=1`), `DeviceTopK::{Min,Max}` generalizes this to select the `k` items with the lowest/highest values.
* STL precedence: `std::min_element`

cub/cub/device/dispatch/dispatch_batched_topk.cuh
elstehle: The idea here is to have the possibility of serving both (segmented and batched) use cases through a single `Dispatch`. Segmented would wrap the items in a `cuda::make_strided_iterator(cuda::make_counting_iterator(items_it), segment_size)`, e.g., on the `Device` interface.

I will rename the `DispatchSegmentedTopK` to `DispatchBatchedTopK` to reflect the nature of the `Dispatch` part.

cub/benchmarks/bench/segmented_topk/keys.cu
elstehle: Mh, some of the workloads we want to focus on for the LLM use cases (Sparse Attention and MoE) have fixed and well-known `k` values. So, I definitely want to reflect those. 

How can I figure out if skipping is supported by tuning?

cub/cub/detail/segmented_params.cuh
bernhardmgruber: I think @gevtushenko meant:
```suggestion
template <typename T, T Min = ::cuda::std::numeric_limits<T>::lowest(), T Max = ::cuda::std::numeric_limits<T>::max()>
```

elstehle: Oh, thanks Bernhard, that definitely makes sense. Sorry, I had missed the code part the suggestion referred to.

cub/cub/agent/agent_segmented_topk.cuh
elstehle: I ignored the `requirements` part for block-level top-k and would like to defer it - and more generally the exact `BlockTopK` interface - until we expose this primitive publicly, when moving `cub::detail::BlockTopK` to `cub::BlockTopK`. We want to give the block interface some more thought. 

The priority for now is to be able to use this primitive in the `DeviceSegmentedTopK`, so we do not need an immediate answer for the interface yet.

cub/test/catch2_test_device_segmented_topk_keys.cu
elstehle: Yes, that's right. I decided to postpone tests for pairs until there was sufficient feedback for `keys`, so I wouldn't duplicate revision work ;-)

cub/cub/agent/agent_segmented_topk.cuh
elstehle: Good catch. Yes, definitely. Fixed.

cub/cub/agent/agent_segmented_topk.cuh
elstehle: Thanks! Fixed it.

cub/cub/device/dispatch/dispatch_segmented_topk.cuh
elstehle: This wasn't intentional. I fixed it.

cub/cub/device/dispatch/dispatch_segmented_topk.cuh
elstehle: Good catch! This seems to missing in most of CUB's uses of `triple_chevron`. I will follow up with an issue (or maybe a PR straight).

cub/benchmarks/bench/segmented_topk/keys.cu
gevtushenko: If this is intentional - it's fine to keep it this way! Consider adding a comment saying that these numbers are not random.

> How can I figure out if skipping is supported by tuning?

We'll see it when we start tuning it, so fine to keep like that for this PR :)

cub/cub/agent/agent_segmented_topk.cuh
elstehle: That's a great remark. I added an optimized `if constexpr` for loading items and I've added a `TODO` for storing items. Storing is a bit more involved, because currently, `BlockTopK` will return items in a blocked arrangement. I generally want to revisit the interfaces when adding performant `BlockTopK` implementations, also to cover shared memory as input (for better TMA support) and output.

cub/benchmarks/bench/segmented_topk/keys.cu
elstehle: Thanks for noticing, I've settled on just removing scaling. I've neglected the tuning parts a bit, since I expect that part to need some overhaul for the final tuning parameters, once we have a better understanding of all the approaches that we want to cover.

cub/cub/device/dispatch/tuning/tuning_batched_topk.cuh
```diff
@@ -0,0 +1,43 @@
...
+#include <cub/agent/agent_batched_topk.cuh>
+#include <cub/util_device.cuh>
+
+CUB_NAMESPACE_BEGIN
+namespace detail::segmented_topk
+{
+template <typename KeyT, typename ValueT, typename SegmentSizeT, ::cuda::std::int64_t MaxK>
+struct policy_hub
+{
+  struct Policy900 : ChainedPolicy<900, Policy900, Policy900>
```
bernhardmgruber, line 27: Remark: I assume I cannot convince you to update the PR to use the new tuning API design? ;) Staying with the status quo will definitely allow this work to land sooner.

cub/cub/device/dispatch/tuning/tuning_batched_topk.cuh
```diff
@@ -0,0 +1,43 @@
...
+#  pragma clang system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_MSVC)
+#  pragma system_header
+#endif // no system header
+
+#include <cub/agent/agent_batched_topk.cuh>
+#include <cub/util_device.cuh>
+
+CUB_NAMESPACE_BEGIN
+namespace detail::segmented_topk
```
bernhardmgruber: Suggestion: Is the algorithm a batched or a segmented implementation? I think the namespace and the file name should agree on this.

cub/cub/agent/agent_batched_topk.cuh
```diff
@@ -0,0 +1,269 @@
...
+namespace detail::segmented_topk
+{
+template <int BlockThreads, int ItemsPerThread>
+struct AgentBatchedTopKWorkerPerSegmentPolicy
+{
+  /// Threads per thread block
+  static constexpr int BLOCK_THREADS = BlockThreads;
+
+  /// Items per thread (per tile of input)
+  static constexpr int ITEMS_PER_THREAD = ItemsPerThread;
```
bernhardmgruber: Important: Please adhere to the naming conventions.
```suggestion
struct agent_batched_topk_worker_per_segment_policy
{
  /// Threads per thread block
  static constexpr int block_threads = BlockThreads;

  /// Items per thread (per tile of input)
  static constexpr int items_per_thread = ItemsPerThread;
```
Applies to more entities in this PR.

cub/cub/agent/agent_batched_topk.cuh
```diff
@@ -0,0 +1,269 @@
...
+
+  using key_t   = typename ::cuda::std::iterator_traits<key_it_t>::value_type;
+  using value_t = typename ::cuda::std::iterator_traits<value_it_t>::value_type;
+
+  static constexpr int block_threads    = ActivePolicyT::BLOCK_THREADS;
+  static constexpr int items_per_thread = ActivePolicyT::ITEMS_PER_THREAD;
+  static constexpr int tile_size        = block_threads * items_per_thread;
+
+  // Check if we are dealing with keys-only or key-value pairs
+  static constexpr bool is_keys_only = ::cuda::std::is_same<value_t, cub::NullType>::value;
```
bernhardmgruber:
```suggestion
  static constexpr bool is_keys_only = ::cuda::std::is_same_v<value_t, cub::NullType>;
```

cub/cub/block/block_topk.cuh
```diff
@@ -0,0 +1,81 @@
...
+#include <cub/block/block_radix_sort.cuh>
+#include <cub/util_ptx.cuh>
+#include <cub/util_type.cuh>
+
+CUB_NAMESPACE_BEGIN
+
+namespace detail
+{
+template <typename KeyT, int BlockDimX, int ItemsPerThread, typename ValueT = NullType>
+class BlockTopK
```
bernhardmgruber: Suggestion: despite existing practice, this should be called:
```suggestion
class block_topk
```

cub/cub/device/dispatch/dispatch_batched_topk.cuh
```diff
@@ -0,0 +1,519 @@
...
+      , d_value_segments_it(d_value_segments_it)
+      , d_value_segments_out_it(d_value_segments_out_it)
+      , segment_sizes(segment_sizes)
+      , k(k)
+      , select_directions(select_directions)
+      , num_segments(num_segments)
+      , total_num_items_guarantee(total_num_items_guarantee)
+      , stream(stream)
+      , ptx_version(ptx_version)
+  {}
```
bernhardmgruber: Suggestion: We can probably skip this constructor and use aggregate init.

cub/test/catch2_test_device_segmented_topk_keys.cu
elstehle: I've added the tests for pairs.

cub/cub/device/dispatch/tuning/tuning_batched_topk.cuh
elstehle: It's definitely on my agenda and I want to land the new tuning API design for batched top-k by the time we have the full implementation in place and ready to be tuned. I would like to defer it for one of the follow-up PRs of this line of work, though.

cub/cub/device/dispatch/tuning/tuning_batched_topk.cuh
elstehle: I renamed the namespace to `batched_topk`. The implementation is generic enough to support batched and segmented workloads. Since batched is more generic than segmented, I already renamed the implementation details (`Dispatch`, `Agent`) to batched and the namespace now followed.

cub/cub/agent/agent_batched_topk.cuh
elstehle: I renamed some of the remaining entities and member function, where applicable, to follow `snake_case`. As for the static policy member constants, that was actually a conscious decision: there are helpers, like `policy_wrapper_t` used within the `vsmem_helper` that rely on the policy member variables being `ALL_CAPS`. I don't think these constants are important right now, since I will revisit this part anyways.

cub/cub/agent/agent_batched_topk.cuh
```diff
@@ -0,0 +1,269 @@
+// SPDX-FileCopyrightText: Copyright (c) 2025, NVIDIA CORPORATION. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+//! @file
+//! cub::AgentBatchedTopK implements a stateful abstraction of CUDA thread blocks for participating in device-wide topK.
```
bernhardmgruber: Suggestion: I don't see a class named `cub::AgentBatchedTopK`. Please adjust this comment.

cub/cub/agent/agent_batched_topk.cuh
```diff
@@ -0,0 +1,269 @@
...
+          typename SelectDirectionParameterT,
+          typename NumSegmentsParameterT>
+struct agent_batched_topk_worker_per_segment
+{
+  // -------------------------------------------------------------------------
+  // Types and Constants
+  // -------------------------------------------------------------------------
+  // Derive inner types from Iterator of Iterators
+  using key_it_t   = typename ::cuda::std::iterator_traits<KeyInputItItT>::value_type;
+  using value_it_t = typename ::cuda::std::iterator_traits<ValueInputItItT>::value_type;
```
bernhardmgruber: Remark: we have `[cub::detail::]it_value_t` which is often a bit shorter to spell. No need to change anything here.

cub/cub/agent/agent_batched_topk.cuh
```diff
@@ -0,0 +1,269 @@
...
+      }
+    }
+
+    __syncthreads();
+
+    // Perform Block Top-K
+    if constexpr (!is_keys_only)
+    {
+      // Pass both keys and values
+      bool is_successful_dispatch = detail::params::dispatch_discrete(
```
bernhardmgruber:
```suggestion
      const bool is_successful_dispatch = detail::params::dispatch_discrete(
```
Applies another time below.

cub/cub/block/block_topk.cuh
```diff
@@ -0,0 +1,81 @@
+// SPDX-FileCopyrightText: Copyright (c) 2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+//! @file
+//! The @c cub::block_topk class provides a :ref:`collective <collective-primitives>` method for selecting the top-k
```
bernhardmgruber:
```suggestion
//! The @c cub::detail::block_topk class provides a :ref:`collective <collective-primitives>` method for selecting the top-k
```

cub/cub/block/block_topk.cuh
```diff
@@ -0,0 +1,81 @@
...
+  TempStorage& temp_storage;
+
+public:
+  _CCCL_DEVICE _CCCL_FORCEINLINE block_topk(TempStorage& temp_storage)
+      : temp_storage(temp_storage)
+  {}
+
+  _CCCL_DEVICE _CCCL_FORCEINLINE void
+  Max(KeyT (&keys)[ItemsPerThread],
+      ValueT (&values)[ItemsPerThread],
```
bernhardmgruber: Suggestion: name the overloads for keys `max_keys` and the ones for values `max_values`. I don't see a benefit of calling both the key and the values overload the same way. It's probably only confusing.

elstehle: Changed to your suggestions. Exact interface is still tbd, when making these public. I previously oriented along other block-level primitives, like `BlockRadixSort`, which provide both overloads for keys and pairs under the same name. I'd be in favor of giving different names to avoid future overload conflicts. Also, I named it `{max,min}_{keys,pairs}` following our other conventions in device-level primitives.

cub/cub/agent/agent_batched_topk.cuh
elstehle: Agreed. I've previously been choosing at random, tbh, but since `it_value_t` is briefer, I will switch to consistently using that one from now on. Also, since I was at it, I changed it here as well.

cub/cub/agent/agent_batched_topk.cuh
elstehle: I've removed the comment altogether.

cub/cub/device/dispatch/tuning/tuning_batched_topk.cuh
```diff
@@ -0,0 +1,43 @@
...
+#include <cub/agent/agent_batched_topk.cuh>
+#include <cub/util_device.cuh>
+
+CUB_NAMESPACE_BEGIN
+namespace detail::batched_topk
+{
+template <typename KeyT, typename ValueT, typename SegmentSizeT, ::cuda::std::int64_t MaxK>
+struct policy_hub
+{
+  struct Policy900 : ChainedPolicy<900, Policy900, Policy900>
```
bernhardmgruber, line 27: Q: Doesn't 900 mean that this algorithm cannot run below Hopper?

elstehle: My understanding of `ChainedPolicy` is that it invokes the last policy in the chain if it doesn't find an "appropriate" one while traversing the chain (see `have_previous_policy` in `ChainedPolicy`).

bernhardmgruber: Oh yeah, just checked it. Alright then!

cub/benchmarks/bench/segmented_topk/keys.cu
```diff
@@ -0,0 +1,174 @@
...
+
+  // Static selection direction (max)
+  using select_direction_value_t = cub::detail::batched_topk::select_direction_static<cub::detail::topk::select::max>;
+
+  // Number of segments is a host-accessible value
+  using num_segments_uniform_t = cub::detail::batched_topk::num_segments_uniform<>;
+
+  // Total number of items guarantee type
+  using total_num_items_guarantee_t =
+    cub::detail::batched_topk::total_num_items_guarantee<min_num_total_items, max_num_total_items>;
```
bernhardmgruber, line 72: Critical: `total_num_items_guarantee` is defined as:
```
template <::cuda::std::int64_t MaxNumItems, ::cuda::std::int64_t MinNumItemsT>
struct total_num_items_guarantee;
```
The arguments here appear swapped.

cub/cub/detail/segmented_params.cuh
```diff
@@ -0,0 +1,254 @@
...
+  _CCCL_HOST_DEVICE constexpr per_segment_param(IteratorT iter)
+      : iterator(iter)
+  {}
+
+  // Constructor 2: Explicit runtime bounds
+  _CCCL_HOST_DEVICE constexpr per_segment_param(IteratorT iter, T min_v, T max_v)
+      : iterator(iter)
+      , min_value(min_v)
+      , max_value(max_v)
+  {}
```
bernhardmgruber: Suggestion: could use a single ctor
```suggestion
  _CCCL_HOST_DEVICE constexpr per_segment_param(IteratorT iter, T min_v = Min, T max_v = Max)
      : iterator(iter)
  {}
```

cub/cub/detail/segmented_params.cuh
```diff
@@ -0,0 +1,254 @@
...
+      : value(v)
+  {}
+
+  uniform_param() = default;
+};
+
+// Deduction Guide:
+// Allows: uniform_param{5} -> uniform_param<int, INT_MIN, INT_MAX>
+template <typename T>
+uniform_param(T) -> uniform_param<T>;
```
bernhardmgruber: Remark: I would not need the comment above, since it does not add any new information.

bernhardmgruber, line 92: Remark: I remember clang in CUDA mode to require host/device annotations on deduction guides. Maybe those days are gone.

cub/cub/detail/segmented_params.cuh
```diff
@@ -0,0 +1,254 @@
...
+inline constexpr bool is_static_param_v = is_static_param<T>::value;
+
+template <typename T>
+using is_uniform_param = ::cuda::std::is_same<typename T::param_tag, tag_uniform>;
+
+template <typename T>
+inline constexpr bool is_uniform_param_v = is_uniform_param<T>::value;
+
+template <typename T>
+using is_per_segment_param = ::cuda::std::is_same<typename T::param_tag, tag_per_segment>;
```
bernhardmgruber: Suggestion: The type aliases are not used anywhere, except for the definitions of the variable templates. I would remove them.

cub/cub/detail/segmented_params.cuh
```diff
@@ -0,0 +1,254 @@
...
+};
+
+// -----------------------------------------------------------------------------
+// Parameter Type Helpers
+// -----------------------------------------------------------------------------
+template <typename T>
+using is_static_param = ::cuda::std::is_same<typename T::param_tag, tag_static>;
+
+template <typename T>
+inline constexpr bool is_static_param_v = is_static_param<T>::value;
```
bernhardmgruber: Suggestion: Instead of relying on `::param_tag` I would just match the template specialization
```suggestion
template <typename T>
inline constexpr bool is_static_param_v = false

template <typename T, T Value>
inline constexpr bool is_static_param_v<static_constant_param<T, Value>> = true;
```
and drop the tags. This requires less types involved.

cub/cub/detail/segmented_params.cuh
```diff
@@ -0,0 +1,254 @@
...
+    // Case 2: Runtime uniform.
+    return p.value;
+  }
+  else
+  {
+    // Case 3: Per-segment.
+    static_assert(is_per_segment_param_v<ParamT>, "Unknown parameter type");
+    return p.iterator[segment_id];
+  }
+}
```
bernhardmgruber: Suggestion: It may be less code if instead of `resolve_param`, each parameter type would just have a member function `auto get([[maybe_unused]] SegmentIndexT segment_id)` and do whatever is in the `if constexpr `branches here.

cub/cub/detail/segmented_params.cuh
```diff
@@ -0,0 +1,254 @@
...
+  // Fold expression over the supported options.
+  // This generates code equivalent to:
+  // if (val == Opt1) f(integral_constant<Opt1>);
+  // else if (val == Opt2) f(integral_constant<Opt2>);
+  // ...
+  bool match_found = ((val == Opts ? (f(::cuda::std::integral_constant<T, Opts>{}), true) : false) || ...);
+
+  // Optional: Handling cases where the runtime value was not in the supported
+  // list. In a release build, we assume the user respected the contract.
+  return match_found;
```
bernhardmgruber, line 235: Q: should we add an assertion here that we found a match?

cub/cub/detail/segmented_params.cuh
```diff
@@ -0,0 +1,254 @@
...
+// options.
+template <typename T, T... Opts, typename Functor>
+_CCCL_HOST_DEVICE bool dispatch_impl(T val, supported_options<T, Opts...>, Functor&& f)
+{
+  // Fold expression over the supported options.
+  // This generates code equivalent to:
+  // if (val == Opt1) f(integral_constant<Opt1>);
+  // else if (val == Opt2) f(integral_constant<Opt2>);
+  // ...
+  bool match_found = ((val == Opts ? (f(::cuda::std::integral_constant<T, Opts>{}), true) : false) || ...);
```
bernhardmgruber:
```suggestion
  const bool match_found = ((val == Opts ? (f(::cuda::std::integral_constant<T, Opts>{}), true) : false) || ...);
```

cub/cub/device/dispatch/tuning/tuning_batched_topk.cuh
```diff
@@ -0,0 +1,43 @@
...
+#  pragma system_header
+#endif // no system header
+
+#include <cub/agent/agent_batched_topk.cuh>
+#include <cub/util_device.cuh>
+
+CUB_NAMESPACE_BEGIN
+namespace detail::batched_topk
+{
+template <typename KeyT, typename ValueT, typename SegmentSizeT, ::cuda::std::int64_t MaxK>
```
bernhardmgruber, line 24: Strong suggestion: Unless there is a good reason that I am not seeing, those template parameters are unused. Please remove them.

cub/cub/device/dispatch/tuning/tuning_batched_topk.cuh
```diff
@@ -0,0 +1,43 @@
...
+    // Policies selected based on optimal performance for different segment sizes
+    // The list below will be checked to determine if each policy can support the one-worker-per-segment approach
+    // within available shared memory limits. Policies must be ordered by decreasing segment size
+    using worker_per_segment_policies =
+      ::cuda::std::tuple<agent_batched_topk_worker_per_segment_policy<256, 64>,
+                         agent_batched_topk_worker_per_segment_policy<256, 32>,
+                         agent_batched_topk_worker_per_segment_policy<256, 16>,
+                         agent_batched_topk_worker_per_segment_policy<256, 8>,
+                         agent_batched_topk_worker_per_segment_policy<256, 4>,
+                         agent_batched_topk_worker_per_segment_policy<128, 2>>;
```
bernhardmgruber: Q: I don't understand where the segment sizes are prescribed here. The integers here are block sizes and items per thread.

Edit: I have found how this is used. Please reword the comment to say, that **the first policy** that fits SMEM is taken.

cub/cub/device/dispatch/dispatch_batched_topk.cuh
```diff
@@ -0,0 +1,490 @@
...
+          ::cuda::std::int64_t MinNumSegments = 1,
+          ::cuda::std::int64_t MaxNumSegments = ::cuda::std::numeric_limits<::cuda::std::int64_t>::max()>
+using num_segments_per_segment =
+  params::per_segment_param<NumSegmentsItT, ::cuda::std::int64_t, MinNumSegments, MaxNumSegments>;
+
+// ------------ TOTAL NUMBER OF ITEMS PARAMETER TYPES ------------
+
+// Number of items guarantee
+template <::cuda::std::int64_t MaxNumItems  = ::cuda::std::numeric_limits<::cuda::std::int64_t>::max(),
+          ::cuda::std::int64_t MinNumItemsT = 1>
```
bernhardmgruber: Strong suggestion: It's massively surprising if all other parameters take the template parameters in the order `Min, Max` and this one here is reversed. Please swap the template parameters.

cub/cub/device/dispatch/dispatch_batched_topk.cuh
```diff
@@ -0,0 +1,490 @@
...
+          template <typename...> class WorkerPerSegmentAgentT,
+          typename... AgentParamsT>
+struct find_valid_policy_impl
+{
+  // Inspect the current policy
+  using current_policy_t = ::cuda::std::tuple_element_t<Index, PoliciesT>;
+
+  // Instantiate agent to check temporary storage size
+  using current_agent_t      = WorkerPerSegmentAgentT<current_policy_t, AgentParamsT...>;
+  static constexpr bool fits = (sizeof(typename current_agent_t::TempStorage) <= 48 * 1024);
```
bernhardmgruber: Suggestion: please use the global constant we have for this:
```suggestion
  static constexpr bool fits = (sizeof(typename current_agent_t::TempStorage) <= max_smem_per_block);
```

cub/cub/device/dispatch/tuning/tuning_batched_topk.cuh
bernhardmgruber: Suggestion: since this is basically an array of pairs, can we just make that a `static constexpr` array of a simple struct? Than the whole `find_valid_policy_impl` could just be implemented as a `constexpr` function.

If out of scope for this PR, we will probably have to do this when we migrate to the new tuning API, which is constexpr value based.

cub/cub/device/dispatch/dispatch_batched_topk.cuh
```diff
@@ -0,0 +1,490 @@
...
+    KeyInputItItT,
+    KeyOutputItItT,
+    ValueInputItItT,
+    ValueOutputItItT,
+    SegmentSizeParameterT,
+    KParameterT,
+    SelectDirectionParameterT,
+    NumSegmentsParameterT>;
+
+  using agent_t = typename find_valid_policy_t::worker_per_segment_agent_t;
```
bernhardmgruber: Important: `worker_per_segment_agent_t` may be `void` here. What happens then? I guess it's an implementation limit for now? We should add a static assert here that gives a proper error message here. Otherwise, `agent_t::tile_size` will give a cryptic error message in case `agent_t` is `void` later

cub/cub/device/dispatch/dispatch_batched_topk.cuh
```diff
@@ -0,0 +1,490 @@
...
+      return cudaSuccess;
+    }
+
+    // TODO (elstehle): support number of segments provided by device-accessible iterator
+    // Only uniform number of segments are supported (i.e., we  need to resolve the number of segments on the host)
+    static_assert(!params::is_per_segment_param_v<NumSegmentsParameterT>,
+                  "Only uniform segment sizes are currently supported.");
+
+    // TODO (elstehle): support larger number of segments through multiple kernel launches
+    int grid_dim = static_cast<int>(resolve_param(num_segments, 0));
```
bernhardmgruber:
```suggestion
    const int grid_dim = static_cast<int>(resolve_param(num_segments, 0));
```

cub/cub/device/dispatch/dispatch_batched_topk.cuh
```diff
@@ -0,0 +1,490 @@
...
+      return error;
+    }
+
+    return cudaSuccess;
+  }
+
+  template <typename ActivePolicyT>
+  CUB_RUNTIME_FUNCTION _CCCL_FORCEINLINE cudaError_t Invoke()
+  {
+    using max_policy_t = typename SelectedPolicy::max_policy;
```
bernhardmgruber: Important: this is unused. Please remove.

cub/cub/agent/agent_batched_topk.cuh
```diff
@@ -0,0 +1,267 @@
...
+    __syncthreads();
+
+    // Perform Block Top-K
+    if constexpr (!is_keys_only)
+    {
+      // Pass both keys and values
+      const bool is_successful_dispatch = detail::params::dispatch_discrete(
+        select_directions, segment_id, [this, &thread_keys, &thread_values, k](auto direction_tag) {
+          select_topk_pairs(thread_keys, thread_values, k, direction_tag);
+        });
```
bernhardmgruber: Suggestion: I would inline the logic of `select_topk_pairs` here to avoid an extra function and keep the code shorter:
```suggestion
        select_directions, segment_id, [this, &thread_keys, &thread_values, k](auto direction_tag) {
              block_topk_t block_alg(temp_storage.topk);
              if constexpr (direction_tag == detail::topk::select::max)
                block_alg.max_pairs(keys, values, k);
              else
                block_alg.min_pairs(keys, values, k);
        });
```
And delete `select_topk_pairs`.

The same refactoring would apply to `select_topk_keys`

cub/cub/agent/agent_batched_topk.cuh
```diff
@@ -0,0 +1,267 @@
...
+    else
+    {
+      block_topk_t(temp_storage.topk).min_pairs(keys, values, k);
+    }
+  }
+
+  _CCCL_DEVICE _CCCL_FORCEINLINE void Process()
+  {
+    // Identify Segment
+    int segment_id = blockIdx.x;
```
bernhardmgruber:
```suggestion
    const int segment_id = blockIdx.x;
```

cub/cub/agent/agent_batched_topk.cuh
```diff
@@ -0,0 +1,267 @@
...
+      return;
+    }
+
+    // Resolve Segment Parameters
+    const auto segment_size = resolve_param(segment_sizes, segment_id);
+    const auto k            = resolve_param(k_param, segment_id);
+    const auto direction    = resolve_param(select_directions, segment_id);
+
+    // Determine padding key based on direction
+    key_t padding_key = (direction == detail::topk::select::max)
```
bernhardmgruber:
```suggestion
    const key_t padding_key = (direction == detail::topk::select::max)
```

cub/cub/agent/agent_batched_topk.cuh
```diff
@@ -0,0 +1,267 @@
...
+    else
+    {
+      // Potentially partial final load with padding
+      block_load_keys_t(temp_storage.load_keys).Load(block_keys_in, thread_keys, segment_size, padding_key);
+    }
+
+    // Load Values (if applicable)
+    [[maybe_unused]] value_t thread_values[items_per_thread];
+
+    if constexpr (!is_keys_only)
```
bernhardmgruber, line 180: Suggestion: I would invert the condition and swap the branches below. Applies above as well.

cub/cub/block/block_topk.cuh
```diff
@@ -0,0 +1,81 @@
...
+  {
+    BlockRadixSortT(temp_storage.sort_storage).SortDescending(keys, values, begin_bit, end_bit);
+  }
+
+  _CCCL_DEVICE _CCCL_FORCEINLINE void
+  max_keys(KeyT (&keys)[ItemsPerThread], int /*k*/, int begin_bit = 0, int end_bit = sizeof(KeyT) * 8)
+  {
+    BlockRadixSortT(temp_storage.sort_storage).SortDescending(keys, begin_bit, end_bit);
+  }
+  _CCCL_DEVICE _CCCL_FORCEINLINE void min_pairs(
```
bernhardmgruber, line 63:
```suggestion
  }
  
  _CCCL_DEVICE _CCCL_FORCEINLINE void min_pairs(
```

cub/test/catch2_test_device_segmented_topk_keys.cu
```diff
@@ -0,0 +1,168 @@
...
+
+#include <thrust/detail/raw_pointer_cast.h>
+#include <thrust/sort.h>
+
+#include <cuda/iterator>
+#include <cuda/std/type_traits>
+
+#include "catch2_test_device_topk_common.cuh"
+#include "catch2_test_launch_helper.h"
+#include "cuda/std/__algorithm/min.h"
```
bernhardmgruber:
```suggestion
#include <cuda/std/__algorithm/min.h>
```

cub/test/catch2_test_device_segmented_topk_keys.cu
```diff
@@ -0,0 +1,168 @@
...
+using max_segment_size_list = c2h::enum_type_list<cuda::std::size_t, 4 * 1024>;
+
+// Segment size: static, uniform
+using max_num_k_list = c2h::enum_type_list<cuda::std::size_t, 32, 4 * 1024>;
+
+using key_types =
+  c2h::type_list<cuda::std::uint8_t,
+                 float,
+                 cuda::std::uint64_t
+// clang-format off
```
bernhardmgruber, line 82: Important: Please turn clang-format on again after the type list.

cub/test/catch2_test_device_segmented_topk_keys.cu
```diff
@@ -0,0 +1,168 @@
...
+                                                  , half_t
+                                 #endif // TEST_HALF_T()
+                                 #if TEST_BF_T()
+                                                  , bfloat16_t
+                                 #endif // TEST_BF_T()
+                                 >;
+
+
+C2H_TEST("DeviceSegmentedTopK::{Min,Max}Keys work with small fixed-size segments",
+         "[keys][segmented][topk][device]",
```
bernhardmgruber: Suggestion: replace "segmented" with "batched"

cub/test/catch2_test_device_segmented_topk_keys.cu
```diff
@@ -0,0 +1,168 @@
...
+          c2h::type_name<segment_index_t>(),
+          static_max_segment_size,
+          static_max_k,
+          segment_size,
+          k,
+          num_segments,
+          direction);
+
+  // Prepare input & output
+  c2h::device_vector<key_t> keys_in_buffer(num_segments * segment_size);
```
bernhardmgruber:
```suggestion
  c2h::device_vector<key_t> keys_in_buffer(num_segments * segment_size, thrust::no_init);
```
since we overwrite the buffer anyway below

cub/test/catch2_test_device_segmented_topk_pairs.cu
```diff
@@ -0,0 +1,255 @@
...
+using max_segment_size_list = c2h::enum_type_list<cuda::std::size_t, 4 * 1024>;
+
+// Segment size: static, uniform
+using max_num_k_list = c2h::enum_type_list<cuda::std::size_t, 32, 4 * 1024>;
+
+using key_types =
+  c2h::type_list<cuda::std::uint8_t,
+                 float,
+                 cuda::std::uint64_t
+// clang-format off
```
bernhardmgruber, line 103: Same here, please turn on again.

cub/test/catch2_test_device_segmented_topk_pairs.cu
```diff
@@ -0,0 +1,255 @@
...
+    auto d_sorted_values = thrust::raw_pointer_cast(sorted_values.data());
+    auto num_items = sorted_values.size();
+
+    flag_duplicates_in_segment<ValueT*> flag_op{d_sorted_values, k};
+
+    auto num_duplicates = thrust::count_if(
+        thrust::make_counting_iterator(size_t{0}),
+        thrust::make_counting_iterator(num_items - 1), flag_op);
+
+    return num_duplicates == 0;
```
bernhardmgruber: Suggestion: Technically, `thrust::find_if` would be faster, but both algorithms are expected to traverse the whole range, so it shouldn't matter.

cub/test/catch2_test_device_segmented_topk_pairs.cu
```diff
@@ -0,0 +1,255 @@
...
+          c2h::type_name<segment_index_t>(),
+          static_max_segment_size,
+          static_max_k,
+          segment_size,
+          k,
+          num_segments,
+          direction);
+
+  // Prepare keys input & output
+  c2h::device_vector<key_t> keys_in_buffer(num_segments * segment_size);
```
bernhardmgruber:
```suggestion
  c2h::device_vector<key_t> keys_in_buffer(num_segments * segment_size, thrust::no_init);
```

cub/test/catch2_test_device_topk_common.cuh
```diff
@@ -149,3 +154,92 @@ public:
...
+
+  // Query temporary storage size
+  size_t temp_storage_bytes = 0;
+  if (direction == cub::detail::topk::select::min)
+  {
+    cub::DeviceSegmentedSort::SortKeys(
+      nullptr, temp_storage_bytes, d_keys, num_items, num_segments, segment_offsets_it, (segment_offsets_it + 1));
+
+    // Allocate temporary storage
+    c2h::device_vector<cuda::std::uint8_t> d_temp_storage(temp_storage_bytes);
```
bernhardmgruber:
```suggestion
    c2h::device_vector<cuda::std::uint8_t> d_temp_storage(temp_storage_bytes, thrust::no_init);
```

cub/test/catch2_test_device_topk_common.cuh
```diff
@@ -149,3 +154,92 @@ public:
...
+      segment_offsets_it,
+      (segment_offsets_it + 1));
+  }
+  else
+  {
+    cub::DeviceSegmentedSort::SortKeysDescending(
+      nullptr, temp_storage_bytes, d_keys, num_items, num_segments, segment_offsets_it, (segment_offsets_it + 1));
+
+    // Allocate temporary storage
+    c2h::device_vector<cuda::std::uint8_t> d_temp_storage(temp_storage_bytes);
```
bernhardmgruber:
```suggestion
    c2h::device_vector<cuda::std::uint8_t> d_temp_storage(temp_storage_bytes, thrust::no_init);
```

cub/test/catch2_test_device_topk_common.cuh
```diff
@@ -149,3 +154,92 @@ public:
...
+  // Prepare segment offsets
+  auto segment_offsets_it =
+    cuda::make_strided_iterator(cuda::make_counting_iterator<cuda::std::int64_t>(0), segment_size);
+
+  // Query temporary storage size
+  size_t temp_storage_bytes = 0;
+  if (direction == cub::detail::topk::select::min)
+  {
+    cub::DeviceSegmentedSort::SortKeys(
+      nullptr, temp_storage_bytes, d_keys, num_items, num_segments, segment_offsets_it, (segment_offsets_it + 1));
```
bernhardmgruber, line 206: Q: How terrible would it be to use `thrust::sort_by_key` instead, just to make the code shorter?

oleksandr-pavlyk:
During code review session, I noted that for floating point input one often needs to disregard NaN values from consideration, and only select top-k among finite values.  This would require passing a custom decomposer instance to radix sort.

We need to make sure that this task is solvable with this algorithm, e.g., by applying decomposed to values passed to the algorithm.

cub/cub/block/block_topk.cuh
```diff
@@ -0,0 +1,81 @@
...
+
+  _CCCL_DEVICE _CCCL_FORCEINLINE void max_pairs(
+    KeyT (&keys)[ItemsPerThread],
+    ValueT (&values)[ItemsPerThread],
+    int /*k*/,
+    int begin_bit = 0,
+    int end_bit   = sizeof(KeyT) * 8)
+  {
+    BlockRadixSortT(temp_storage.sort_storage).SortDescending(keys, values, begin_bit, end_bit);
+  }
```
davebayer, line 56: I suppose currently the top K elements are stored sequentially filling first `ItemsPerThread / k` thead's storage.

Do we want to provide an alternative that stores the top K elements in distributing them sequentially among the threads? That means each thread gets `k / ItemsPerThread` elements

cub/cub/agent/agent_batched_topk.cuh
```diff
@@ -0,0 +1,267 @@
...
+
+CUB_NAMESPACE_BEGIN
+
+namespace detail::batched_topk
+{
+template <int BlockThreads, int ItemsPerThread>
+struct agent_batched_topk_worker_per_segment_policy
+{
+  /// Threads per thread block
+  static constexpr int BLOCK_THREADS = BlockThreads;
```
gevtushenko: question: can we have a lower case `block_threads` here?

cub/cub/agent/agent_batched_topk.cuh
```diff
@@ -0,0 +1,267 @@
...
+  static constexpr int items_per_thread = ActivePolicyT::ITEMS_PER_THREAD;
+  static constexpr int tile_size        = block_threads * items_per_thread;
+
+  // Check if we are dealing with keys-only or key-value pairs
+  static constexpr bool is_keys_only = ::cuda::std::is_same_v<value_t, cub::NullType>;
+
+  // -------------------------------------------------------------------------
+  // Primitive Types
+  // -------------------------------------------------------------------------
+  using block_load_keys_t = BlockLoad<key_t, block_threads, items_per_thread, BLOCK_LOAD_WARP_TRANSPOSE>;
```
oleksandr-pavlyk: Algorithm choices for `block_load_keys_t`, `block_load_vals_t`, etc. should be controlled via policy, as they can be involved in tuning the algorithm.

cub/cub/agent/agent_batched_topk.cuh
```diff
@@ -0,0 +1,267 @@
...
+
+  // -------------------------------------------------------------------------
+  // Members
+  // -------------------------------------------------------------------------
+  TempStorage_& temp_storage;
+  KeyInputItItT d_key_segments_it;
+  KeyOutputItItT d_key_segments_out_it;
+  ValueInputItItT d_value_segments_it;
+  ValueOutputItItT d_value_segments_out_it;
+  SegmentSizeParameterT segment_sizes;
```
gevtushenko, line 107: suggestion: minor nit to rename segment to batch

cub/cub/agent/agent_batched_topk.cuh
gevtushenko: important: I believe `blockIdx.x` is unsigned, so it can fit 4B segment indices, but here we cast that to `int`. I don't see streaming approach or any special handling for that on the dispatch end. Should we use `unsigned` type here and maybe add a test for TopK of larger than int max number of segments?

cub/cub/device/dispatch/dispatch_batched_topk.cuh
```diff
@@ -0,0 +1,490 @@
...
+// K via iterator
+template <typename KItT,
+          ::cuda::std::int64_t MinK = 1,
+          ::cuda::std::int64_t MaxK = ::cuda::std::numeric_limits<::cuda::std::int64_t>::max()>
+using k_per_segment = params::per_segment_param<KItT, ::cuda::std::int64_t, MinK, MaxK>;
+
+// ------------ TOTAL NUMBER OF SEGMENTS ------------
+// Number of segments known at compile time
+template <::cuda::std::int64_t StaticNumSegments>
+using num_segments_static = params::static_constant_param<::cuda::std::int64_t, StaticNumSegments>;
```
gevtushenko, line 95: question: we use `blockIdx.x` for segment index. Here, num segments are `int64_t`. Are you going to add streaming approach in the dispatch layer later on?

cub/cub/agent/agent_batched_topk.cuh
```diff
@@ -0,0 +1,267 @@
...
+    if constexpr (params::has_single_static_value_v<SegmentSizeParameterT>
+                  && params::static_min_value_v<SegmentSizeParameterT> == tile_size)
+    {
+      // No padding needed
+      block_load_keys_t(temp_storage.load_keys).Load(block_keys_in, thread_keys);
+    }
+    else
+    {
+      // Potentially partial final load with padding
+      block_load_keys_t(temp_storage.load_keys).Load(block_keys_in, thread_keys, segment_size, padding_key);
```
oleksandr-pavlyk, line 174: It may be worth to add dynamic check, and call `block_load_keys_t(temp_storage.load_keys).load(block_keys_in, thread_keys)` is `segment_size == tile_size`.

cub/cub/agent/agent_batched_topk.cuh
```diff
@@ -0,0 +1,267 @@
...
+  _CCCL_DEVICE _CCCL_FORCEINLINE void Process()
+  {
+    // Identify Segment
+    int segment_id = blockIdx.x;
+
+    // Boundary check
+    if (segment_id >= resolve_param(num_segments, 0))
+    {
+      return;
+    }
```
gevtushenko: question: why do we need this check? Can we somehow launch more thead blocks than we need?

cub/cub/device/dispatch/tuning/tuning_batched_topk.cuh
```diff
@@ -0,0 +1,43 @@
...
+template <typename KeyT, typename ValueT, typename SegmentSizeT, ::cuda::std::int64_t MaxK>
+struct policy_hub
+{
+  struct Policy900 : ChainedPolicy<900, Policy900, Policy900>
+  {
+    // Policies selected based on optimal performance for different segment sizes
+    // The list below will be checked to determine if each policy can support the one-worker-per-segment approach
+    // within available shared memory limits. Policies must be ordered by decreasing segment size
+    using worker_per_segment_policies =
+      ::cuda::std::tuple<agent_batched_topk_worker_per_segment_policy<256, 64>,
```
gevtushenko: suggestion: let's include tuple header

cub/cub/device/dispatch/dispatch_batched_topk.cuh
oleksandr-pavlyk: Suggestion: replace `48 * 1024` with `cub::detail::max_smem_per_block`

cub/cub/block/block_topk.cuh
elstehle: Thanks for the suggestion! Probably, yes. Most likely, we will also need an overload allowing to provide inputs/outputs via shared memory. The `BlockTopK` interface is really just a stub implementation for now, the interfaces will be determined once we actually expose `BlockTopK` as a block-level primitive.

elstehle:
> During code review session, I noted that for floating point input one often needs to disregard NaN values from consideration, and only select top-k among finite values. This would require passing a custom decomposer instance to radix sort.
> 
> We need to make sure that this task is solvable with this algorithm, e.g., by applying decomposed to values passed to the algorithm.

Agreed. We already have an issue tracking this: https://github.com/NVIDIA/cccl/issues/6290

cub/test/catch2_test_device_topk_common.cuh
elstehle: I don't think that's a good idea. `thrust::sort_by_key` is  a `Device[Radix]Sort::SortPairs`. So, now, we would have to sort not only on the keys but on also on an array of segment tags.

cub/cub/agent/agent_batched_topk.cuh
elstehle: Good point! I added a comment for this to be benchmarked once we have a full-stack implementation.

cub/cub/agent/agent_batched_topk.cuh
elstehle: Thanks for highlighting this. I've added them. There will be many more parameters for tuning and I actually wanted to defer this until the actual segmented implementation is further along and we understand the parameters we want to include in tuning.

cub/cub/agent/agent_batched_topk.cuh
elstehle: I don't follow. Can you please elaborate.

cub/cub/device/dispatch/dispatch_batched_topk.cuh
elstehle: Agree, this was just was just a leftover.

cub/cub/device/dispatch/tuning/tuning_batched_topk.cuh
elstehle: Comment addded

cub/cub/device/dispatch/tuning/tuning_batched_topk.cuh
elstehle: That is because we have not tuned yet. These are the parameters currently judged relevant for looking up the  right tuning policy. If I would remove them now, I would have to re-add them once we tune. So I will keep them around for now.

cub/cub/detail/segmented_params.cuh
elstehle: Good idea! Added an assert.

cub/benchmarks/bench/segmented_topk/keys.cu
elstehle: Good catch, thanks for noticing.

cub/cub/device/dispatch/tuning/tuning_batched_topk.cuh
elstehle: I added a TODO to have this as part of moving over to the new tuning API.

cub/cub/device/dispatch/dispatch_batched_topk.cuh
elstehle: Currently, it would run into a `static_assert` in the kernel. I've added another static assert here in the dispatch.

cub/cub/detail/segmented_params.cuh
elstehle: I integrated your feedback. I called the getter member function `get_param`, because I think it will make maintenance easier if we can code search in some distant future without seeing tons of false positives.

cub/cub/device/dispatch/dispatch_batched_topk.cuh
```diff
@@ -0,0 +1,488 @@
...
+                    agent_batched_topk_worker_per_segment,
+                    KeyInputItItT,
+                    KeyOutputItItT,
+                    ValueInputItItT,
+                    ValueOutputItItT,
+                    SegmentSizeParameterT,
+                    KParameterT,
+                    SelectDirectionParameterT,
+                    NumSegmentsParameterT>::worker_per_segment_policy_t::block_threads)) __global__
+  void device_segmented_topk_kernel(
```
gevtushenko: suggestion: we can move kernels into `kernel_*` file to avoid moving a lot of code later when we enable it for cccl.c

gevtushenko (approved):
A few minor nits except for offset one. Since interface is still in detail namespace, it should be safe to merge and address in a follow-up PR.

cub/cub/agent/agent_batched_topk.cuh
elstehle: This has been changed in the meanwhile. I had it `ALL_CAPS` mostly just to align with the vsmem helper utilities.

cub/cub/device/dispatch/dispatch_batched_topk.cuh
elstehle: Yup, that's the plan. We already have a TODO there:
> `// TODO (elstehle): support larger number of segments through multiple kernel launches`

cub/cub/agent/agent_batched_topk.cuh
elstehle: At some point `num_segments` may also be device-accessible-only, at which point we cannot rely on having the grid the right size. I added a TODO to skip the check if the parameter can be read from the host and, hence, we safely assume the right grid size.

cub/cub/device/dispatch/tuning/tuning_batched_topk.cuh
elstehle: Good catch. Added header include.

cub/cub/device/dispatch/dispatch_batched_topk.cuh
elstehle: Done 👍

cub/cub/agent/agent_batched_topk.cuh
elstehle: I wouldn't want to call it `BatchSizeParameterT` and `batch_size` here, as other libraries use this naming to describe the number of sub-problems ("number of segments"). I find `segment_sizes` easier to understand, even though it's strictly not correct for a batched algorithm. We could just call it `sizes` or `num_items`. Any preference?

cub/cub/agent/agent_batched_topk.cuh
elstehle: I've added a `static_cast<int>` to avoid potential unsigned/signed comparisons in the rest of the implementation. The idea is to add a streaming approach, partitioning segments across kernel invocations (there's already a `TODO`, see https://github.com/NVIDIA/cccl/pull/6980#discussion_r2752581824).

bernhardmgruber: @gevtushenko meta question: are we concerned about `blockIdx.x` getting larger than `INT_MAX`? I see lot's of CUB algorithms handling block and tile indices in `int`.

cub/cub/agent/agent_batched_topk.cuh
bernhardmgruber: Please apply the transformation:
```c++
if constexpr (!cond)
  a
else
  b
```
to 
```c++
if constexpr (cond)
  b
else
  a
```

elstehle: Thanks for clarifying. Sorry, I had looked at the wrong conditional, which didn't have an else branch. I should start checking the line numbers people comment on.

cub/cub/agent/agent_batched_topk.cuh
gevtushenko: @elstehle I missed the comment, thanks! It's a good idea to have streaming as a follow up PR.

@bernhardmgruber previously, `num_segments` was `int` in many algorithms. In some kernels, we kept it as `int` but implement streaming approach where we process at most `INT_MAX` segments at a time, so that `int` is sufficient on the kernel side. In other algorithms, we launch at most occupancy kernels, so we know the `blockIdx.x` always fits `int`. If you find other cases, where tile index is an int, I'd suggest filing an issue to investigate large problem sizes. Specifically here, today we accept `int64_t` as `num_segments`, so kernel can be launched with more blocks than what `int` can fit.

cub/cub/agent/agent_batched_topk.cuh
gevtushenko: I see your point now! Now I don't have a preference

PR #6981, opened 2025-12-15
Add a way to pass a stream to an execution policy

miscco:
We want to pass streams through the execution policy to allow customization on the user side


miscco:
closing in favor of #7299

PR #6984, opened 2025-12-15
Remove `_CCCL_HAS_CUDA_COMPILER()`

davebayer:
I've removed almost all uses of `_CCCL_HAS_CUDA_COMPILER()` and now it's the time to remove the macro completely. Basically it's just almost always not the right tool for the job.

Most of the uses were/are replaced by `_CCCL_CUDA_COMPILATION()` and the rest was/is almost always replacement for `_CCCL_CUDA_COMPILER(NVCC)` or `_CCCL_HAS_CTK()`.

The compiler detection is now wrapped by `_CCCL_CUDA_COMPILATION()`, so if we are not compiling CUDA code, there is not CUDA compiler. This was not always the case, because if the user does:
```bash
$ nvcc -x c++ my_src.cpp
```
we would still set the `_CCCL_CUDA_COMPILER_NVCC()` even though we are compiling a c++ file. In my opinion this is not correct, because nvcc just controls the compilation trajectory and doesn't compile any source code.

This means that many of the builtins that `cudafe++` doesn't support are not used even though the c++ compiler could use them without any problem.

So, this PR removes `_CCCL_HAS_CUDA_COMPILER()` completely and puts all compiler/compilation macros to a consistent state.

### DIFF
--- a/libcudacxx/include/cuda/std/__internal/features.h
+++ b/libcudacxx/include/cuda/std/__internal/features.h
@@ -54,7 +54,7 @@
 #endif // _LIBCUDACXX_HAS_NO_INCOMPLETE_RANGES
 
 // libcu++ requires host device support for its tests. Until then restrict usage to at least 12.2
-#if _CCCL_HAS_NVFP16() && _CCCL_CTK_AT_LEAST(12, 2) && (_CCCL_HAS_CTK() || defined(LIBCUDACXX_ENABLE_HOST_NVFP16))
+#if _CCCL_HAS_NVFP16() && _CCCL_CTK_AT_LEAST(12, 2)
 #  define _LIBCUDACXX_HAS_NVFP16() 1
 #else
 #  define _LIBCUDACXX_HAS_NVFP16() 0


libcudacxx/include/cuda/std/__internal/atomic.h
```diff
@@ -22,7 +22,7 @@
 
 #include <cuda/std/__internal/features.h>
 
-#if _CCCL_HAS_CUDA_COMPILER()
+#if _CCCL_CUDA_COMPILATION()
 #  define _CCCL_ATOMIC_ALWAYS_LOCK_FREE(size, ptr) (size <= 8)
```
davebayer, line 26: I am not sure about this, @wmaxey can you confirm this is correct? Shouldn't this be `_CCCL_DEVICE_COMPILATION()`?

libcudacxx/include/cuda/std/__cccl/compiler.h
```diff
@@ -109,31 +115,21 @@
...
-#else // ^^^ has cuda compiler ^^^ / vvv no cuda compiler vvv
-#  define _CCCL_HAS_CUDA_COMPILER() 0
-#endif // ^^^ no cuda compiler ^^^
-
-#if defined(__CUDACC__) || _CCCL_CUDA_COMPILER(NVHPC)
-#  define _CCCL_CUDA_COMPILATION() 1
-#else // ^^^ compiling .cu file ^^^ / vvv not compiling .cu file vvv
-#  define _CCCL_CUDA_COMPILATION() 0
-#endif // ^^^ not compiling .cu file ^^^
+#if _CCCL_CUDA_COMPILATION()
```
fbusato, line 118: This looks correct also in the edge cases:
nvc++ -> `_NVHPC_CUDA`
clang, nvrtc -> `__CUDACC__`

libcudacxx/include/cuda/std/__cccl/dialect.h
```diff
@@ -188,10 +188,10 @@
 
 // Fixme: replace the condition with (!_CCCL_DEVICE_COMPILATION())
 // FIXME: Enable this for clang-cuda in a followup
-#if !_CCCL_HAS_CUDA_COMPILER()
+#if !_CCCL_CUDA_COMPILATION()
```
fbusato, line 191: ok, so here we are saying `long double` can be used in a .cpp file compiled with nvcc

libcudacxx/include/cuda/std/__internal/features.h
```diff
@@ -54,8 +54,7 @@
 #endif // _LIBCUDACXX_HAS_NO_INCOMPLETE_RANGES
 
 // libcu++ requires host device support for its tests. Until then restrict usage to at least 12.2
-#if _CCCL_HAS_NVFP16() && _CCCL_CTK_AT_LEAST(12, 2) \
-  && (_CCCL_HAS_CUDA_COMPILER() || defined(LIBCUDACXX_ENABLE_HOST_NVFP16))
+#if _CCCL_HAS_NVFP16() && _CCCL_CTK_AT_LEAST(12, 2) && (_CCCL_HAS_CTK() || defined(LIBCUDACXX_ENABLE_HOST_NVFP16))
```
fbusato: is `_CCCL_HAS_CTK()` redundant considering `_CCCL_CTK_AT_LEAST(12, 2)`?

libcudacxx/include/cuda/std/__cccl/dialect.h
fbusato: (to be extended in the future)

libcudacxx/include/cuda/std/__cccl/compiler.h
davebayer: Yes, `nvcc` also defines `__CUDACC__` when compiling CUDA source files

libcudacxx/include/cuda/std/__cccl/dialect.h
davebayer: Yeah, our `long double` support is.. very poor.. and we don't actually run tests with it, we just check if we can compile header files..

libcudacxx/include/cuda/std/__internal/features.h
davebayer: Yeah, somehow missed it, thanks!

davebayer: Actually, the condition is redundant, we already check this in `_CCCL_HAS_NVFP16()`

fbusato: even better :)

libcudacxx/include/cuda/std/__internal/atomic.h
wmaxey: This is fine, I have some plans to change this macro in the future anyways.

wmaxey (approved):

PR #6985, opened 2025-12-15
Workaround for a potential bug in the driver related to TMA descriptor

fbusato:
## Description

bug/5736804 workaround

### DIFF
--- a/libcudacxx/include/cuda/__driver/driver_api.h
+++ b/libcudacxx/include/cuda/__driver/driver_api.h
@@ -962,9 +962,8 @@ __cutensormap_size_bytes(::cuda::std::size_t __num_items, ::CUtensorMapDataType 
     case ::CU_TENSOR_MAP_DATA_TYPE_16U4_ALIGN16B:
       return __num_items / 2;
 #  endif // _CCCL_CTK_AT_LEAST(12, 8)
-    default:
-      _CCCL_UNREACHABLE();
   }
+  return 0; // MSVC workaround
 }
 
 [[nodiscard]] _CCCL_HOST_API inline ::CUtensorMap __tensorMapEncodeTiled(


libcudacxx/include/cuda/__driver/driver_api.h
```diff
@@ -946,6 +946,14 @@ __graphKernelNodeSetAttribute(::CUgraphNode __node, ::CUkernelNodeAttrID __id, c
     __swizzle,
     __l2Promotion,
     __oobFill));
+#  if _CCCL_CTK_BELOW(13, 2)
+  if (::cuda::__driver::__version_below(13, 2))
+  {
+    const auto __tensorMapPtr = reinterpret_cast<::cuda::std::uint64_t*>(static_cast<void*>(&__tensorMap));
+    __tensorMapPtr[1]         = ~(::cuda::std::uint64_t{1} << 21);
+  }
+#  endif // _CCCL_CTK_BELOW(13, 2)
```
davebayer:
```suggestion
  if (::cuda::__driver::__version_below(13, 2))
  {
    const auto __tensorMapPtr = reinterpret_cast<::cuda::std::uint64_t*>(static_cast<void*>(&__tensorMap));
    __tensorMapPtr[1]         = ~(::cuda::std::uint64_t{1} << 21);
  }
```
We should just check the version at runtime. Can you also add a comment what this is and the nvbug number?

fbusato: well, we can skip for newer CTK. Why making a driver call when it is not needed

fbusato: > Can you also add a comment what this is and the nvbug number?

sorry, the PR was not finalized yet

libcudacxx/include/cuda/__driver/driver_api.h
```diff
@@ -946,6 +995,32 @@ __graphKernelNodeSetAttribute(::CUgraphNode __node, ::CUkernelNodeAttrID __id, c
     __swizzle,
     __l2Promotion,
     __oobFill));
+  // workaround for nvbug 5736804
+#  if _CCCL_CTK_BELOW(13, 2)
+  if (::cuda::__driver::__version_below(13, 2))
+  {
+    const auto __tensor_req_size                = __globalDim[__tensorRank - 1] * __globalStrides[__tensorRank - 1];
+    ::cuda::std::size_t __tensor_req_size_bytes = 0;
+    try
```
fbusato: here also a termination is fine

libcudacxx/include/cuda/__driver/driver_api.h
miscco: I am with @fbusato here, we should avoid driver calls if we can

davebayer: I agree, but we can prevent the case when the user compiles with CUDA 13.2, but then runs the program with driver version 13.1. We do this already in some other APIs. We cache the driver version, so there should be almost no overhead at all.

I think preventing a bug has greater value than 1 version comparison.

fbusato: I double-checked. The driver version is encoded, so there is practically zero overhead when calling this function. I'm fine with this modification.
Anyway, I don't think it is possible to compile and run a code compiled with CUDA 13.2 with lower driver version. However, considering our support for non-conventional platforms, edge cases could still  be possible.

PR #6987, opened 2025-12-16
[libcu++] Correctly handle extended lambda in cuda::launch

pciolkosz:
There is no way to inspect arguments of an extended lambda in the host code, so we can't automatically dispatch to either pass the kernel config into the lambda or not.
For now the safe option is to require the user to explicitly pass in the config twice, which is not ideal, but lambda kernels are expected to be short and might not need the config object.
Added a test for extended lambda, had to move to lit test to be able to specify the nvcc flag.

This PR also changes some of the common testing headers, because I wanted to use the pinned type in lit test without including the c2h headers.

### DIFF
--- a/cub/cub/device/dispatch/tuning/tuning_transform.cuh
+++ b/cub/cub/device/dispatch/tuning/tuning_transform.cuh
@@ -326,6 +326,15 @@ struct tuning_vec<1200, StoreSize>
   static constexpr int items_per_thread = 8;
 };
 
+// manually tuned triad on A100
+template <int StoreSize, int LoadSize0, int... LoadSizes>
+struct tuning_vec<800, StoreSize, LoadSize0, LoadSizes...>
+{
+  static constexpr int block_threads    = 128;
+  static constexpr int vec_size         = 4;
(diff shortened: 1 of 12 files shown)


libcudacxx/test/libcudacxx/cuda/ccclrt/launch/extended_lambda.pass.cpp
```diff
@@ -0,0 +1,46 @@
...
+  auto config           = cuda::block_dims<32>() & cuda::grid_dims<1>();
+  auto assign_42_lambda = [] __device__(int* pi) {
+    *pi = 42;
+  };
+  cuda::launch(stream, config, assign_42_lambda, i.get());
+  stream.sync();
+  assert(*i == 42);
+
+  auto assign_1337_lambda = [] __device__(auto config, int* pi) {
+    static_assert(config.dims.count(cuda::thread, cuda::block) == 32);
```
davebayer:
```suggestion
    static_assert(config.dims.count(cuda::gpu_thread, cuda::block) == 32);
```

davebayer (approved):

PR #6988, opened 2025-12-16
Add non-throwing overloads to `is_pointer_accessible`

fbusato:
## Description

`is_host_accessible`, `is_device_accessible`, and `is_managed` throw exceptions when the underlying driver calls fail.
On the other hand, such functions are mostly used in assertion where side effects are not allowed.

This PR introduces non-throwing overloads via `cuda::std::nothrow_t` tag dispatch.

### DIFF
--- a/libcudacxx/include/cuda/__driver/driver_api.h
+++ b/libcudacxx/include/cuda/__driver/driver_api.h
@@ -887,7 +887,8 @@ __graphKernelNodeSetAttribute(::CUgraphNode __node, ::CUkernelNodeAttrID __id, c
   return static_cast<bool>(__result);
 }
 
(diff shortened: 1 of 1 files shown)


libcudacxx/include/cuda/__memory/is_pointer_accessible.h
```diff
@@ -58,6 +59,17 @@ _CCCL_HOST_API inline bool is_managed(const void* __p)
...
+[[nodiscard]]
+_CCCL_HOST_API inline bool is_managed(const void* __p, ::cuda::std::nothrow_t) noexcept
+try
+{
+  return ::cuda::is_managed(__p);
+}
+catch (...)
+{
+  return false;
+}
```
davebayer: This won't work if exceptions are disabled.

davebayer (approved):

PR #6995, opened 2025-12-16
[cuda.compute]: fix alignment not being set properly for `gpu_struct` types

NaderAlAwar:
## Description

Prior to this PR, the alignment for a gpu_struct like this

```python
@gpu_struct
class pair:
    first: np.float32
    second: np.float32
```

is being passed as 1 to c.parallel when it is supposed to be 4. This PR fixes that.

## Checklist


oleksandr-pavlyk (approved):
LGTM. A test would be nice to have. Perhaps no the correctness test, but a test than retrieves iterator and checks that alignment is as expected

PR #6997, opened 2025-12-17
[libcu++] Remove _view from the shared memory getter name

pciolkosz:
Currently we have `dynamic_shared_memory` launch option type and `dynamic_shared_memory_view` to get it from the config. We would like to move to two overloads of `dynamic_shared_memory` function, one that creates the option given type/size and one to get it from the config. This simplifies the interface and the overloads will never be ambitious.

Because `dynamic_shared_memory` is currently a type, I had to rename the type to `dynamic_shared_memory_option` with private constructors and instead introduce factory functions

### DIFF
--- a/c2h/include/c2h/test_util_vec.h
+++ b/c2h/include/c2h/test_util_vec.h
@@ -49,198 +49,199 @@ inline int CoutCast(signed char val)
 /**
  * Vector1 overloads
  */
-#  define C2H_VEC_OVERLOAD_1(T)                                               \
-    /* Ostream output */                                                      \
-    inline std::ostream& operator<<(std::ostream& os, const T& val)           \
-    {                                                                         \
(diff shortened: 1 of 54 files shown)


libcudacxx/include/cuda/__launch/configuration.h
```diff
@@ -258,48 +258,11 @@ class _CCCL_DECLSPEC_EMPTY_BASES dynamic_shared_memory
   using typename __base_type::value_type; //!< Value type of the dynamic
                                           //!< shared memory elements.
   using typename __base_type::view_type; //!< The view type returned by the
-                                         //!< cuda::dynamic_shared_memory_view(config).
```
davebayer, line 261: We should rename this member, too

pciolkosz: We agreed to get back to `view_type`

PR #6998, opened 2025-12-17
Small fixes around warpspeed scan

bernhardmgruber:
This extracts some early fixes around the incoming warpspeed scan implementation

### DIFF
--- a/cub/test/catch2_test_device_scan.cu
+++ b/cub/test/catch2_test_device_scan.cu
@@ -76,20 +76,14 @@ C2H_TEST("Device scan works with all device interfaces", "[scan][device]", full_
   using offset_t = int32_t;
 
   constexpr offset_t min_items = 1;
-  constexpr offset_t max_items = 1'000'000;
+  constexpr offset_t max_items = 10'000'000;
 
   // Generate the input sizes to test for
   const offset_t num_items = GENERATE_COPY(
     10,
-    1000,
     1337,
-    3000,
     1 * 31 * 128, // int64 tile size for lookahead
-    10'000,
-    100'000,
-    1'000'000,
-    10'000'000,
-    100'000'000,
+    10'000, // a handful of tiles for lookahead
     take(3, random(min_items, max_items)),
     values({
       min_items,


cub/benchmarks/bench/scan/exclusive/base.cuh
```diff
@@ -101,7 +101,14 @@ try
...
   size_t tmp_size;
   dispatch_t::Dispatch(
-    nullptr, tmp_size, d_input, d_output, op_t{}, wrapped_init_t{T{}}, static_cast<int>(input.size()), 0 /* stream */);
+    nullptr,
+    tmp_size,
+    d_input,
+    d_output,
+    op_t{},
+    wrapped_init_t{T{}},
+    static_cast<offset_t>(input.size()),
```
bernhardmgruber, line 110: Here, we used to cast to `int` and then promote again to `offset_t` (argument type of `Dispatch`). This is wrong for sizes larger than `INT_MAX`.

cmake/CCCLGetDependencies.cmake
```diff
@@ -87,7 +87,7 @@ endmacro()
 
 set(
   CCCL_NVBENCH_SHA
-  "0c24f0250bf4414ab5ad19709090c6396e76516b"
+  "34f1e2a7eeeeec7bb2965a3ec9667d297b479cfe"
```
bernhardmgruber, line 90: This is the version we used during all measurements. Not the most recent one though.

cub/cub/util_type.cuh
```diff
@@ -243,7 +243,7 @@ struct InputValue
 {
   using value_type    = T;
   using iterator_type = IterT;
-  _CCCL_HOST_DEVICE _CCCL_FORCEINLINE operator T()
+  _CCCL_HOST_DEVICE _CCCL_FORCEINLINE operator T() const
```
bernhardmgruber, line 246: `m_future_value` has a conversion operator that is already `const` and `m_immediate_value` already has type `T`, so making the current `operator T()` `const` should not be a breaking change.

elstehle (approved):

cub/test/catch2_test_device_scan.cu
```diff
@@ -75,19 +75,29 @@ C2H_TEST("Device scan works with all device interfaces", "[scan][device]", full_
...
+    1000,
+    1337,
+    3000,
+    1 * 31 * 128, // int64 tile size for lookahead
+    10'000,
+    100'000,
+    1'000'000,
+    10'000'000,
+    100'000'000,
     take(3, random(min_items, max_items)),
```
bernhardmgruber, line 87: The additional sizes change the test runtime from 2.5s to 2min 5s, which is a bit excessive, I think we should cut some.

bernhardmgruber: Reduced the test sizes, we are now taking 14.7s which is more acceptable.

PR #7002, opened 2025-12-17
[thrust] Ignore CUDA free errors in thrust memory resource

pciolkosz:
NVBUG 5744986

It seems cub tests do some cleanup after main exits which is technically not supported. We should ignore the failure in memory resource deallocate, so even if the call fails because CUDA Runtime is being cleaned up, the test won't fail

### DIFF
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -106,7 +106,7 @@ jobs:
 
       - name: Ensure Jekyll is disabled for GitHub Pages
         if: ${{ github.event_name == 'push' && !inputs.pr_number }} # Only deploy .nojekyll for main docs, not PR previews
-        uses: peaceiris/actions-gh-pages@v3
+        uses: peaceiris/actions-gh-pages@v4
         with:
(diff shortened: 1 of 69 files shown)


thrust/thrust/system/cuda/memory_resource.h
```diff
@@ -67,12 +67,7 @@ class cuda_memory_resource final : public mr::memory_resource<Pointer>
 
   void do_deallocate(Pointer p, [[maybe_unused]] std::size_t bytes, [[maybe_unused]] std::size_t alignment) override
   {
-    cudaError_t status = Dealloc(thrust::detail::pointer_traits<Pointer>::get(p));
-
-    if (status != cudaSuccess)
-    {
-      thrust::cuda_cub::throw_on_error(status, "CUDA free failed");
-    }
+    [[maybe_unused]] auto status = Dealloc(thrust::detail::pointer_traits<Pointer>::get(p));
```
bernhardmgruber, line 72: Suggestion: Add a comment why we omit error handling here.

bernhardmgruber (approved):

bernhardmgruber (approved):

PR #7006, opened 2025-12-18
Use cudaMemcpyDefault for trivial copies

bdice:
## Description

In some instances, libcudf and/or rapidsmpf may call Thrust algorithms with an execution policy with a custom allocator that allocates device-accessible host memory. This was observed to fail on some systems, because `trivial_copy_from_device` uses `cudaMemcpyDeviceToHost` which becomes a `cudaErrorInvalidValue`. If we instead use `cudaMemcpyDefault`, the CUDA runtime correctly infers the source/destination locations and no error is observed.

Sample error output from @nirandaperera:

```
[13:01:35.917][140065985130496][CUDA][E] Returning 1 (CUDA_ERROR_INVALID_VALUE) from cuMemcpyDtoHAsync_v2
terminate called after throwing an instance of 'thrust::system::system_error'
  what():  __copy:: D->H: failed: cudaErrorInvalidValue: invalid argument
```

I added a basic test that calls `trivial_copy_(from|to)_device` with host pointers. This should cover the failure case, though it is actually triggered by some more complex Thrust algorithms in `cudf::pack` that I did not think were worth replicating as a unit test. Being able to use (pinned) host memory more easily is becoming important for performance, so we'd like to support this behavior.

## Checklist


bdice:
@pciolkosz Thanks for helping us analyze this issue. Could this PR be a backport candidate for CCCL 3.2?

nirandaperera:
@bdice thank you for doing this. Can we revert the fixes in `utils.h` and make sure that the test case catches the error? I saw this specifically when using pinned memory.

nirandaperera:
BTW this may well fix https://github.com/rapidsai/cudf/issues/20886.

thrust/thrust/system/cuda/detail/util.h
```diff
@@ -145,7 +145,7 @@ trivial_copy_from_device(Type* dst, Type const* src, size_t count, cudaStream_t
     return status;
   }
 
-  status = ::cudaMemcpyAsync(dst, src, sizeof(Type) * count, cudaMemcpyDeviceToHost, stream);
+  status = ::cudaMemcpyAsync(dst, src, sizeof(Type) * count, cudaMemcpyDefault, stream);
```
miscco, line 148: I believe we cannot unconditionally do this, because `cudaMemcpyDefault` has the following limitations:

```
However, [cudaMemcpyDefault](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__TYPES.html#group__CUDART__TYPES_1gg18fa99055ee694244a270e4d5101e95b715aff8fb2b8f4f1bb553fee802db57a) is only allowed on systems that support unified virtual addressing.
```

@davebayer could you give this a look and see whether we can sniff the supported cuda runtime version as well as whether unified virtual addressing is available?

miscco: More specifically https://developer.download.nvidia.com/compute/DevZone/docs/html/C/doc/html/group__CUDA__UNIFIED.html

davebayer: If I am right, we always suppose UVA support, because otherwise we wouldn't be able to use `cuPointerGetAttributes` APIs

davebayer: I think this should be fine, @pciolkosz can you please confirm

bdice:
> @bdice thank you for doing this. Can we revert the fixes in `utils.h` and make sure that the test case catches the error? I saw this specifically when using pinned memory.

Yes, I verified it. This tests just does a plain host->host copy but this is sufficient to cover your use case.

thrust/thrust/system/cuda/detail/util.h
bdice: If it's helpful to know, cuDF has been using `cudaMemcpyDefault` in all copies for quite a while and hasn't encountered any issues.

jrhemstad: This is a non-issue. Every system we care about has had unified virtual addressing for a _very_ long time.

pciolkosz (approved):

PR #7007, opened 2025-12-18
cub, c.parallel: {lower,upper}_bound

### LINKED ISSUE #6695: [FEA]: Add vectorized cub::DeviceFind::UpperBound and cub::DeviceFind::LowerBound algorithms
### Is this a duplicate?

### Area

CUB

The current CUB API lacks vectorized equivalents of upper_bound and lower_bound algorithms, which are present in Thrust. I would like to be able to use vectorized cub::DeviceFind::UpperBound and cub::DeviceFind::LowerBound similar to thrust::upper_bound and thrust::lower_bound algorithms. This would help with feature parity and enable these capabilities in cuda.compute.

Introduce new vectorized cub::DeviceFind::UpperBound and cub::DeviceFind::LowerBound APIs, mirroring the interface and behavior provided by thrust::upper_bound and thrust::lower_bound.

_No response_

### Additional context

End goal is to expose upper_bound and lower_bound functionality in cuda.compute.

griwes:
## Description

This PR adds {lower,upper}_bound device algorithms to both CUB and c.parallel.

In the case of CUB, the implementation is very straightforward and directly follows current implementation in Thrust (which I have cleaned up as a drive by change).

In the case of c.parallel, because of how CUB's `for_each` passes in kernel arguments, repeating the slight madness of the current `for` operator construction for the for_each algorithm itself felt beyond annoying, but I needed a kernel pointer; so instead of reusing the kernels available in CUB, I adapted the static-block-size `for_each` kernel to accept all the necessary arguments as separate kernel arguments, then construct the for operator expected by CUB inside the kernel, and finally invoke the CUB `for_each` agent with that operator. So, it's a manually constructed kernel, but it reuses both the agent code and binary search helper types from CUB.

Resolves #6695 

## Checklist

### DIFF
--- a/cub/benchmarks/bench/find_if/base.cu
+++ b/cub/benchmarks/bench/find_if/base.cu
@@ -1,7 +1,7 @@
 // SPDX-FileCopyrightText: Copyright (c) 2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
-#include <cub/device/device_find.cuk>
+#include <cub/device/device_find.cuh>
 
 #include <thrust/count.h>
 #include <thrust/detail/internal_functional.h>
--- a/cub/test/catch2_test_device_find_if.cu
+++ b/cub/test/catch2_test_device_find_if.cu
@@ -3,7 +3,7 @@
 
 #include "insert_nested_NVTX_range_guard.h"
 
-#include <cub/device/device_find.cuk>
+#include <cub/device/device_find.cuh>
 
 #include <thrust/detail/raw_pointer_cast.h>
 
--- a/cub/test/catch2_test_device_find_if_api.cu
+++ b/cub/test/catch2_test_device_find_if_api.cu
@@ -3,7 +3,7 @@
 
 #include "insert_nested_NVTX_range_guard.h"
 
-#include <cub/device/device_find.cuk>
+#include <cub/device/device_find.cuh>
 
 #include <thrust/device_vector.h>
 
--- a/thrust/thrust/system/cuda/detail/find.h
+++ b/thrust/thrust/system/cuda/detail/find.h
@@ -39,7 +39,7 @@
 #if _CCCL_CUDA_COMPILATION()
 #  include <thrust/system/cuda/config.h>
 
-#  include <cub/device/device_find.cuk>
+#  include <cub/device/device_find.cuh>
 
 #  include <thrust/detail/temporary_array.h>
 #  include <thrust/system/cuda/detail/cdp_dispatch.h>


c/parallel/src/binary_search.cu
```diff
@@ -0,0 +1,359 @@
...
+
+  if (num_values == 0)
+  {
+    return error;
+  }
+
+  void* args[] = {&d_in, &num_items, &d_values, &num_values, &d_out, &op};
+
+  const unsigned int thread_count = 256;
+  const size_t items_per_block    = 512;
```
shwina, line 60: Are these essentially hardcoded somewhere in the CUB as well?

griwes: Yes. There's no tunings for for, and this just follows that.

I'm planning to start working on a warp level binary search algorithm in the new year, and then build a device wide one on top of that, as a replacement of the current approach - we'll do actual tunings then.

cub/cub/device/device_merge_sort.cuh
```diff
@@ -303,7 +303,7 @@ public:
    *
    * @param[in,out] temp_storage_bytes
    *   Reference to size in bytes of `d_temp_storage` allocation
-   *
+   :
```
bernhardmgruber: Important: Typo

cub/cub/detail/binary_search_helpers.cuh
```diff
@@ -0,0 +1,65 @@
...
+{
+template <typename RangeIteratorT, typename CompareOpT, typename Mode>
+struct comp_wrapper_t
+{
+  RangeIteratorT first;
+  RangeIteratorT last;
+  CompareOpT op;
+
+  template <typename Value, typename Output>
+  _CCCL_DEVICE _CCCL_FORCEINLINE void operator()(cuda::std::tuple<Value, Output> args) const
```
bernhardmgruber: Important: Please fully qualify all references to entities in the `cuda` namespace:
```suggestion
  _CCCL_DEVICE _CCCL_FORCEINLINE void operator()(::cuda::std::tuple<Value, Output> args) const
```
Applies to several other places in this PR.

cub/cub/device/device_find.cuh
```diff
@@ -0,0 +1,190 @@
...
+  //!
+  //! @param[in] stream
+  //!   **[optional]** CUDA stream to launch kernels within.
+  //!   Default is stream<sub>0</sub>.
+  //!
+  //! [Random Access Iterator]: https://en.cppreference.com/w/cpp/iterator/random_access_iterator
+  //! [Strict Weak Ordering]: https://en.cppreference.com/w/cpp/concepts/strict_weak_order
+  //! [Relation]: https://en.cppreference.com/w/cpp/concepts/relation
+  template <typename RangeIteratorT, typename ValuesIteratorT, typename OutputIteratorT, typename CompareOpT>
+  CUB_RUNTIME_FUNCTION static cudaError_t LowerBound(
```
bernhardmgruber, line 170: Should we call this function `LowerBounds` (plural) since it's computing several at the same time?

docs/cub/Doxyfile
```diff
@@ -14,6 +14,7 @@ INPUT                  = ../../cub/cub \
                          ../../cub/cub/warp \
                          ../../cub/cub/block \
                          ../../cub/cub/device \
+                         ../../cub/cub/device/device_find.cuh \
```
bernhardmgruber: Q: Do we need to explicitly add new files here? I thought this is done automatically, since we add the entire `device` directory.

griwes: Ah, I was dealing with some nonsense and this was an attempt to sanity check some things, I'll clean it up.

shwina (approved):
The C side looks good to me. Thanks!

c/parallel/src/binary_search.cu
```diff
@@ -0,0 +1,359 @@
...
+      arch.c_str(),
+      cub_path,
+      thrust_path,
+      libcudacxx_path,
+      ctk_path,
+      "-std=c++20",
+      "-rdc=true",
+      "-dlto",
+      "-DCUB_DISABLE_CDP",
+      "-G"};
```
oleksandr-pavlyk: The `-G` is likely a debugging leftover. Is there a reason to keep it?

griwes: It is indeed a leftover, will fix.

cub/cub/device/device_find.cuh
griwes: This name is consistent with both @jrhemstad's issue, and with the current naming in Thrust. I'm gonna leave it as is.

cub/cub/device/device_merge_sort.cuh
griwes: Oops.

c/parallel/src/binary_search.cu
```diff
@@ -0,0 +1,358 @@
...
+  CUstream stream)
+{
+  bool pushed    = false;
+  CUresult error = CUDA_SUCCESS;
+
+  try
+  {
+    pushed = try_push_context();
+    auto exec_status =
+      Invoke(d_data, num_items, d_values, num_values, d_out, op, build.cc, (CUfunction) build.kernel, stream);
```
NaderAlAwar, line 293: Suggestion: can we use `static_cast` here? If not, maybe `reinterpret_cast`.

griwes: This mirrors what for does, I can go clean it up a little in both places later but I'd rather not multiply the drive by fixes here.

cub/cub/detail/binary_search_helpers.cuh
```diff
@@ -0,0 +1,65 @@
...
+{
+  RangeIteratorT first;
+  RangeIteratorT last;
+  CompareOpT op;
+
+  template <typename Value, typename Output>
+  _CCCL_DEVICE _CCCL_FORCEINLINE void operator()(::cuda::std::tuple<Value, Output> args) const
+  {
+    ::cuda::std::get<1>(args) = Mode::Invoke(first, last, ::cuda::std::get<0>(args), op);
+  }
```
miscco: Nitpick: I would have preferred an enumeration and `if constexpr` but that is purely personal preference, so no change requested

cub/cub/device/device_find.cuh
```diff
@@ -0,0 +1,275 @@
...
+    ValuesIteratorT values_last,
+    OutputIteratorT output,
+    CompareOpT comp,
+    cudaStream_t stream = 0)
+  {
+    _CCCL_NVTX_RANGE_SCOPE("cub::DeviceFind::UpperBound");
+    return DeviceFor::ForEach(
+      d_temp_storage,
+      temp_storage_bytes,
+      ::cuda::make_zip_iterator(values_first, output),
```
miscco: Nitpick: We have `::cuda::zip_transform_iterator` that does not need to instantiate a tuple but can directly apply the dereferenced iterators to the function, might be useful here

PR #7012, opened 2025-12-18
Allow all public headers to be included with host compilers only

davebayer:
This PR makes all public (`<cuda/meow>` and `<cuda/std/woof>`) headers includable with host compilers only. I've also created cmake target that compiles all public headers with host compiler only + CTK, to make sure we don't use `_CCCL_HAS_CTK()` in place of `_CCCL_CUDA_COMPILATION()`

### DIFF
--- a/libcudacxx/include/cuda/__memcpy_async/is_local_smem_barrier.h
+++ b/libcudacxx/include/cuda/__memcpy_async/is_local_smem_barrier.h
@@ -38,7 +38,7 @@ template <thread_scope _Sco,
           typename _CompF,
           bool _Is_mbarrier = (_Sco == thread_scope_block)
                            && ::cuda::std::is_same_v<_CompF, ::cuda::std::__empty_completion>>
-_CCCL_API inline bool __is_local_smem_barrier(barrier<_Sco, _CompF>& __barrier)
+_CCCL_API inline bool __is_local_smem_barrier([[maybe_unused]] barrier<_Sco, _CompF>& __barrier)
 {
   NV_IF_ELSE_TARGET(
     NV_IS_DEVICE,
--- a/libcudacxx/include/cuda/__memory_pool/device_memory_pool.h
+++ b/libcudacxx/include/cuda/__memory_pool/device_memory_pool.h
@@ -21,18 +21,15 @@
 #  pragma system_header
 #endif // no system header
 
-#if _CCCL_CUDA_COMPILER(CLANG)
-#  include <cuda_runtime.h>
-#  include <cuda_runtime_api.h>
-#endif // _CCCL_CUDA_COMPILER(CLANG)
+#if _CCCL_HAS_CTK()
 
-#include <cuda/__memory_pool/memory_pool_base.h>
-#include <cuda/__memory_resource/get_property.h>
-#include <cuda/__memory_resource/properties.h>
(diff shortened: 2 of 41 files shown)


libcudacxx/include/cuda/__memcpy_async/is_local_smem_barrier.h
```diff
@@ -38,7 +38,7 @@ template <thread_scope _Sco,
           typename _CompF,
           bool _Is_mbarrier = (_Sco == thread_scope_block)
                            && ::cuda::std::is_same_v<_CompF, ::cuda::std::__empty_completion>>
-_CCCL_API inline bool __is_local_smem_barrier(barrier<_Sco, _CompF>& __barrier)
+_CCCL_API inline bool __is_local_smem_barrier([[maybe_unused]] barrier<_Sco, _CompF>& __barrier)
```
fbusato, line 41: this change makes sense, but I'm wondering why the warning is triggered by this PR and not before

fbusato: same for similar cases in the PR

libcudacxx/include/cuda/__memory_pool/device_memory_pool.h
```diff
@@ -21,18 +21,15 @@
 #  pragma system_header
 #endif // no system header
 
-#if _CCCL_CUDA_COMPILER(CLANG)
-#  include <cuda_runtime.h>
```
fbusato, line 25: question. clang automatically pulls *both* headers, right?

davebayer: During CUDA compilation, the CUDA compiler includes these headers. If C++ is being compiled, we include `<cuda_runtime.h>` in our config, because we need to get the version of CUDA Toolkit headers. So, they are always included, if available.

`<cuda_runtime.h>` includes `<cuda_runtime_api.h>`

libcudacxx/include/cuda/__memcpy_async/is_local_smem_barrier.h
davebayer: I was asking the same question. I don't know.

pciolkosz (approved):

davebayer:
pre-commit.ci autofix

bernhardmgruber:
> This PR makes all public (`<cuda/meow>` and `<cuda/std/woof>`) headers includable with host compilers only.

I am a bit confused by this statement, since why shouldn't I be able to use `<cuda/std/optional>` only in device code or from nvrtc? Looking at the diff, I hope the above statement is just wrong and only those headers require a host compiler now that did so in the past.

davebayer:
Sorry, the description is misleading. This PR aims to make *all* our public headers includable from C++ only source files, not only selected ones.

So, after this PR, you should be able to `#include <cuda/memory_pool>` without any problems even though you don't have CTK installed. We will just not include the contents.

On top of that I added header tests for situation when you are compiling with CTK installed, but not compiling with CUDA compiler. Just to make sure our headers don't error out in such a case.

bernhardmgruber:
Thank you for the clarification!

bernhardmgruber (approved):

PR #7023, opened 2025-12-18
Clean up hierarchy

davebayer:


libcudacxx/include/cuda/__hierarchy/dimensions.h
```diff
@@ -30,142 +30,8 @@
 
 _CCCL_BEGIN_NAMESPACE_CUDA
```
pciolkosz: We should remove this header

PR #7027, opened 2025-12-19
`mdspan` to `DLPack`

fbusato:
## Description

The PR implements conversion utilities that take a (host/device/managed) `mdspan` and produce a [DLTensor view](https://dmlc.github.io/dlpack/latest/c_api.html) of the same underlying memory.

The main issue is how to handle the memory associated with the `shape` and `stride` array of DLPack.
The implementation provides a small wrapper that owns the shape and strides arrays and stores a DLTensor whose shape/strides pointers refer into those arrays.

Todo:

- [x] documentation

### DIFF
--- a/libcudacxx/include/cuda/__mdspan/mdspan_to_dlpack.h
+++ b/libcudacxx/include/cuda/__mdspan/mdspan_to_dlpack.h
@@ -37,6 +37,7 @@
 #  include <cuda/std/__type_traits/num_bits.h>
 #  include <cuda/std/__type_traits/remove_cv.h>
 #  include <cuda/std/__utility/cmp.h>
+#  include <cuda/std/array>
 #  include <cuda/std/cstdint>
 #  include <cuda/std/mdspan>
 
@@ -153,8 +154,8 @@ template <typename _ElementType>
 template <::cuda::std::size_t _Rank>
 struct __dlpack_tensor
 {
-  ::cuda::std::array<::cuda::std::int64_t, _Rank> __shape[_Rank]{};
-  ::cuda::std::array<::cuda::std::int64_t, _Rank> __strides[_Rank]{};
+  ::cuda::std::array<::cuda::std::int64_t, _Rank> __shape{};
+  ::cuda::std::array<::cuda::std::int64_t, _Rank> __strides{};
   ::DLTensor __tensor{};
 
   [[nodiscard]] _CCCL_HOST_API ::DLTensor get() const& noexcept _CCCL_LIFETIMEBOUND


libcudacxx/include/cuda/__mdspan/mdspan_to_dlpack.h
```diff
@@ -0,0 +1,496 @@
...
+  else if constexpr (::cuda::std::is_same_v<_ElementType, unsigned long long>)
+  {
+    return ::DLDataType{::kDLUInt, 64, 1};
+  }
+#  if _CCCL_HAS_INT128()
+  else if constexpr (::cuda::std::is_same_v<_ElementType, __uint128_t>)
+  {
+    return ::DLDataType{::kDLUInt, 128, 1};
+  }
+#  endif // _CCCL_HAS_INT128()
```
davebayer:
```suggestion
  else if constexpr (::cuda::std::__cccl_is_integer_v<_ElementType>)
  {
    return ::DLDataType{(::cuda::std::is_signed_v<_ElementType>) ? ::kDLInt : ::kDLUInt, ::cuda::std::__num_bits_v<_ElementType>, 1};
  }
```
let's simplify this

libcudacxx/include/cuda/__mdspan/mdspan_to_dlpack.h
```diff
@@ -0,0 +1,496 @@
...
+  {
+    return ::DLDataType{::kDLFloat6_e3m2fn, 6, 1};
+  }
+#  endif // _CCCL_HAS_NVFP6_E3M2()
+#  if _CCCL_HAS_NVFP4_E2M1()
+  else if constexpr (::cuda::std::is_same_v<_ElementType, ::__nv_fp4_e2m1>)
+  {
+    return ::DLDataType{::kDLFloat4_e2m1fn, 4, 1};
+  }
+#  endif // _CCCL_HAS_NVFP4_E2M1()
```
davebayer, line 109: We can do the same thing for fp types

libcudacxx/include/cuda/__mdspan/mdspan_to_dlpack.h
```diff
@@ -0,0 +1,496 @@
...
+  }
+#  endif // _CCCL_HAS_NVFP16()
+  else if constexpr (::cuda::std::is_same_v<_ElementType, ::cuda::std::complex<float>>)
+  {
+    return ::DLDataType{::kDLComplex, 64, 1};
+  }
+  else if constexpr (::cuda::std::is_same_v<_ElementType, ::cuda::std::complex<double>>)
+  {
+    return ::DLDataType{::kDLComplex, 128, 1};
+  }
```
davebayer: and complex

libcudacxx/include/cuda/__mdspan/mdspan_to_dlpack.h
```diff
@@ -0,0 +1,496 @@
...
+#    if _CCCL_CTK_AT_LEAST(13, 0)
+  else if constexpr (::cuda::std::is_same_v<_ElementType, ::double4_32a>)
+  {
+    return ::DLDataType{::kDLFloat, 64, 4};
+  }
+#    else // ^^^ _CCCL_CTK_AT_LEAST(13, 0) ^^^ / vvv _CCCL_CTK_BELOW(13, 0) vvv
+  else if constexpr (::cuda::std::is_same_v<_ElementType, ::double4>)
+  {
+    return ::DLDataType{::kDLFloat, 64, 4};
+  }
```
davebayer, line 143: and here, just use `tuple_size` for the element count

libcudacxx/include/cuda/__mdspan/mdspan_to_dlpack.h
```diff
@@ -0,0 +1,496 @@
...
+  //--------------------------------------------------------------------------------------------------------------------
+  // Unsupported types
+  else
+  {
+    static_assert(::cuda::std::__always_false_v<_ElementType>, "Unsupported type");
+  }
+}
+
+template <::cuda::std::size_t _Rank>
+class DLPackWrapper
```
davebayer: Hm, do we want to be inconsistent with the naming? Shouldn't this be `dlpack_wrapper`?

libcudacxx/include/cuda/__mdspan/mdspan_to_dlpack.h
```diff
@@ -0,0 +1,496 @@
...
+ * Public API
+ **********************************************************************************************************************/
+
+template <typename _ElementType, typename _Extents, typename _Layout, typename _Accessor>
+[[nodiscard]] _CCCL_HOST_API DLPackWrapper<_Extents::rank()>
+mdspan_to_dlpack(const ::cuda::host_mdspan<_ElementType, _Extents, _Layout, _Accessor>& __mdspan)
+{
+  using __mdspan_type = ::cuda::std::mdspan<_ElementType, _Extents, _Layout, _Accessor>;
+  return ::cuda::__mdspan_to_dlpack(__mdspan_type{__mdspan}, ::kDLCPU, 0);
+}
```
davebayer: I think `to_dlpack` would make more sense, it's consistent with `std::to_string` or `cuda::to_arch_id`

davebayer (changes requested):

libcudacxx/include/cuda/__mdspan/mdspan_to_dlpack.h
fbusato: great suggestion

libcudacxx/include/cuda/__mdspan/mdspan_to_dlpack.h
fbusato: yes, that's also fine. Keep in mind that I also need to add the opposite direction, `dlpack` -> `mdspan`, so we will have `to_mdspan`

libcudacxx/include/cuda/__mdspan/mdspan_to_dlpack.h
fbusato: We should document these utilities; otherwise, no one will know about them.

libcudacxx/include/cuda/__mdspan/mdspan_to_dlpack.h
fbusato: ah sorry, used the convention of my previous project

libcudacxx/include/cuda/__mdspan/mdspan_to_dlpack.h
davebayer: Aaah. Why don't we just add a constructor from mdspan to dlpack?

davebayer:
I was also wondering whether we don't want to introduce a special namespace (something like `cuda::interop`) for interoperability with other APIs apart from CTK

libcudacxx/include/cuda/__mdspan/mdspan_to_dlpack.h
davebayer: And why is it not called `dlpack_tensor` when it's a wrapper fro `DLTensor`?

fbusato: that's ok

libcudacxx/include/cuda/__mdspan/mdspan_to_dlpack.h
```diff
@@ -0,0 +1,446 @@
...
+  {
+    static_assert(::cuda::std::__always_false_v<_ElementType>, "Unsupported type");
+  }
+}
+
+template <::cuda::std::size_t _Rank>
+class DLPackWrapper
+{
+  ::cuda::std::array<::cuda::std::int64_t, _Rank> __shape{};
+  ::cuda::std::array<::cuda::std::int64_t, _Rank> __strides{};
```
davebayer, line 158: Don't we just want to use `std::vector` to get rid of the template parameter? It will introduce some overhead, but I think it would simplify everything

libcudacxx/include/cuda/__mdspan/mdspan_to_dlpack.h
fbusato: do you mean a constructor in `dlpack_tensor` (previously `DLPackWrapper`)?

libcudacxx/include/cuda/__mdspan/mdspan_to_dlpack.h
```diff
@@ -0,0 +1,446 @@
...
+
+  _CCCL_HOST_API ::DLTensor* operator->() noexcept
+  {
+    return &__tensor;
+  }
+
+  _CCCL_HOST_API const ::DLTensor* operator->() const noexcept
+  {
+    return &__tensor;
+  }
```
davebayer: I don't think it's a good idea to implement these operators. IMO they should be used exclusively for pointer wrappers and iterators

libcudacxx/include/cuda/__mdspan/mdspan_to_dlpack.h
davebayer: Yes

libcudacxx/include/cuda/__mdspan/mdspan_to_dlpack.h
fbusato: resolved per offline discussion

libcudacxx/include/cuda/__type_traits/vector_type.h
```diff
@@ -346,10 +345,200 @@ using __vector_type_t = decltype(::cuda::__cccl_vector_type_t_impl<_Tp, _Size>()
...
+template <>
+inline constexpr bool __is_vector_type_v<::long3> = true;
+#  if _CCCL_CTK_AT_LEAST(13, 0)
+template <>
+inline constexpr bool __is_vector_type_v<::long4_16a> = true;
+template <>
+inline constexpr bool __is_vector_type_v<::long4_32a> = true;
+#  else // ^^^ _CCCL_CTK_AT_LEAST(13, 0) ^^^ / vvv _CCCL_CTK_BELOW(13, 0) vvv
+template <>
+inline constexpr bool __is_vector_type_v<::long4> = true;
```
davebayer, line 420: These types exist even in CUDA 13.0, it's just deprecated

docs/libcudacxx/extended_api/mdspan/mdspan_to_dlpack.rst
```diff
@@ -0,0 +1,140 @@
...
+Conversion functions
+--------------------
+
+.. code:: cuda
+
+   namespace cuda {
+
+   template <typename T, typename Extents, typename Layout, typename Accessor>
+   [[nodiscard]] __dlpack_tensor<Extents::rank()>
+   to_dlpack_tensor(const cuda::host_mdspan<T, Extents, Layout, Accessor>& mdspan);
```
davebayer: C++ standard (for example [this](https://en.cppreference.com/w/cpp/utility/format/make_format_args.html)) specifies implementation defined types as `/*the-type*/`. I've used it in other places in the docs, too. Do you think we could continue doing so?

Or do you think it's a bad idea?

docs/libcudacxx/extended_api/mdspan/mdspan_to_dlpack.rst
```diff
@@ -0,0 +1,140 @@
...
+   template <typename T, typename Extents, typename Layout, typename Accessor>
+   [[nodiscard]] __dlpack_tensor<Extents::rank()>
+   to_dlpack_tensor(const cuda::managed_mdspan<T, Extents, Layout, Accessor>& mdspan);
+
+   } // namespace cuda
+
+Types
+-----
+
+``__dlpack_tensor`` is an internal class that stores a ``DLTensor`` and owns the backing storage for its ``shape`` and ``strides`` pointers. The class does not use any heap allocation.
```
davebayer: Maybe we should point out that the class name is a placeholder and shouldn't be used by the user.

docs/libcudacxx/extended_api/mdspan/mdspan_to_dlpack.rst
```diff
@@ -0,0 +1,140 @@
...
+  class __dlpack_tensor {
+  public:
+      __dlpack_tensor() noexcept;
+      __dlpack_tensor(const __dlpack_tensor&) noexcept;
+      __dlpack_tensor(__dlpack_tensor&&) noexcept;
+      __dlpack_tensor& operator=(const __dlpack_tensor&) noexcept;
+      __dlpack_tensor& operator=(__dlpack_tensor&&) noexcept;
+      ~__dlpack_tensor() noexcept = default;
+
+      DLTensor&       get() & noexcept       [[lifetimebound]];
```
davebayer: Why do we want this overload? The user should never modify the object

docs/libcudacxx/extended_api/mdspan/mdspan_to_dlpack.rst
```diff
@@ -0,0 +1,140 @@
...
+  - ``{kDLCPU, 0}`` for ``cuda::host_mdspan``
+  - ``{kDLCUDA, /*device_id*/}`` for ``cuda::device_mdspan``
+  - ``{kDLCUDAManaged, 0}`` for ``cuda::managed_mdspan``
+
+Element types are mapped to ``DLDataType`` according to the DLPack conventions, including:
+
+- ``bool``.
+- Signed and unsigned integers.
+- IEEE-754 Floating-point and extended precision floating-point, including ``__half``, ``__nv_bfloat16``, ``__float128``, FP8, FP6, FP4 when available.
+- Complex: ``cuda::std::complex<__half>``, ``cuda::std::complex<float>``, and ``cuda::std::complex<double>``.
```
davebayer, line 82: Aren't we missing `__nv_bfloat16`, too? And maybe we can just say that `cuda::std::complex` instantiations with supported `value_type` are accepted

docs/libcudacxx/extended_api/mdspan/mdspan_to_dlpack.rst
```diff
@@ -0,0 +1,140 @@
...
+
+  int main() {
+    using extents_t = cuda::std::extents<size_t, 2, 3>;
+
+    int data[6] = {0, 1, 2, 3, 4, 5};
+    cuda::host_mdspan<int, extents_t> md{data, extents_t{}};
+
+    auto dl              = cuda::to_dlpack_tensor(md);
+    const auto& dltensor = dl.get();
+    // auto dltensor = dl.get(); is incorrect; it returns a reference to a temporary object that will be destroyed at the end of the statement.
```
davebayer: this is correct I believe, I think it would be better to demonstrate this on an example where the returned handle goes out of scope. Maybe we could also demonstrate the moves invalidate the previously obtained handles.

libcudacxx/include/cuda/__mdspan/mdspan_to_dlpack.h
```diff
@@ -0,0 +1,311 @@
...
+{
+  using __mdspan_type              = ::cuda::std::mdspan<_ElementType, _Extents, _Layout, _Accessor>;
+  ::CUpointer_attribute __attrs[1] = {::CU_POINTER_ATTRIBUTE_DEVICE_ORDINAL};
+  int __ptr_dev_id                 = 0;
+  void* __results[1]               = {&__ptr_dev_id};
+  const auto __status = ::cuda::__driver::__pointerGetAttributesNoThrow(__attrs, __results, __mdspan.data_handle());
+  if (__status != ::cudaSuccess)
+  {
+    _CCCL_THROW(::std::invalid_argument{"Failed to get device ordinal of a pointer"});
+  }
```
davebayer, line 231: Shouldn't this be a `cuda::cuda_error`?

libcudacxx/include/cuda/__type_traits/vector_type.h
```diff
@@ -340,16 +339,204 @@ template <class _Tp, ::cuda::std::size_t _Size>
...
+_CCCL_SUPPRESS_DEPRECATED_PUSH
+
 template <class _Tp, ::cuda::std::size_t _Size>
 using __vector_type_t = decltype(::cuda::__cccl_vector_type_t_impl<_Tp, _Size>());
 
 template <class _Tp, ::cuda::std::size_t _Size>
 inline constexpr bool __has_vector_type_v = !::cuda::std::is_same_v<__vector_type_t<_Tp, _Size>, void>;
 
+template <class _Tp>
+inline constexpr bool __is_vector_type_v = false;
```
davebayer, line 351: Can we move these to a separate header `<cuda/__type_traits/is_vector_type.h>`?

libcudacxx/include/cuda/__type_traits/vector_type.h
```diff
@@ -340,16 +339,204 @@ template <class _Tp, ::cuda::std::size_t _Size>
...
+
+template <typename _Tp>
+inline constexpr bool __is_extended_fp_vector_type_v = false;
+
+#  if _CCCL_HAS_NVFP8()
+template <>
+inline constexpr bool __is_extended_fp_vector_type_v<::__nv_bfloat162> = true;
+template <>
+inline constexpr bool __is_extended_fp_vector_type_v<::__half2> = true;
+template <>
```
davebayer, line 502: Maybe we should also add these overloads to `__vector_type_t`, so these traits stay consistent. Maybe we can move this to a separate PR?

docs/libcudacxx/extended_api/mdspan/mdspan_to_dlpack.rst
fbusato: I had the same idea in mind. It is not very readable, but it is aligned with C++ standard

docs/libcudacxx/extended_api/mdspan/mdspan_to_dlpack.rst
fbusato: the documentation is correct. DLPack supports a subset of complex types

libcudacxx/include/cuda/__type_traits/vector_type.h
fbusato: yes, better in another PR. We would need to update the include paths of other files

libcudacxx/include/cuda/__type_traits/vector_type.h
fbusato: I will address the issue in another PR.

libcudacxx/include/cuda/__mdspan/mdspan_to_dlpack.h
```diff
@@ -0,0 +1,308 @@
...
+
+  _CCCL_HOST_API __dlpack_tensor(__dlpack_tensor&& __other) noexcept
+      : __shape{::cuda::std::move(__other.__shape)}
+      , __strides{::cuda::std::move(__other.__strides)}
+      , __tensor{__other.__tensor}
+  {
+    __other.__tensor = ::DLTensor{};
+    __update_tensor();
+  }
```
davebayer: Do we need move constructor + assignment operator? What are they good for? They just invalidate the old object for no reason. I would keep just the copy variants :)

libcudacxx/include/cuda/__mdspan/mdspan_to_dlpack.h
```diff
@@ -0,0 +1,308 @@
...
+  [[nodiscard]] _CCCL_HOST_API const ::DLTensor& get() const&& noexcept = delete;
+};
+
+template <typename _ElementType, typename _Extents, typename _Layout, typename _Accessor>
+[[nodiscard]] _CCCL_HOST_API __dlpack_tensor<_Extents::rank()>
+__to_dlpack(const ::cuda::std::mdspan<_ElementType, _Extents, _Layout, _Accessor>& __mdspan,
+            ::DLDeviceType __device_type,
+            int __device_id)
+{
+  static_assert(::cuda::std::is_pointer_v<typename _Accessor::data_handle_type>, "data_handle_type must be a pointer");
```
davebayer, line 178: Maybe in the future we can relax this to *ContiguousIterator* if anyone needs it

libcudacxx/include/cuda/__mdspan/mdspan_to_dlpack.h
fbusato: the reason is to give more freedom/flexibility, but yes, it could be error-prone. The copy operator is not heavy, so we can keep only it.

libcudacxx/include/cuda/__mdspan/mdspan_to_dlpack.h
```diff
@@ -0,0 +1,247 @@
...
+  {
+    static_assert(::cuda::std::__always_false_v<_ElementType>, "Unsupported type");
+    return ::DLDataType{};
+  }
+}
+
+template <::cuda::std::size_t _Rank>
+struct __dlpack_tensor
+{
+  ::cuda::std::array<::cuda::std::int64_t, _Rank> __shape[_Rank]{};
```
fbusato: `cuda::std::array` is need for Rank == 0

libcudacxx/include/cuda/__mdspan/mdspan_to_dlpack.h
```diff
@@ -0,0 +1,247 @@
...
+
+  [[nodiscard]] _CCCL_HOST_API ::DLTensor get() const& noexcept _CCCL_LIFETIMEBOUND
+  {
+    auto __tensor1    = __tensor;
+    __tensor1.shape   = _Rank > 0 ? const_cast<::cuda::std::int64_t*>(__shape.data()) : nullptr;
+    __tensor1.strides = _Rank > 0 ? const_cast<::cuda::std::int64_t*>(__strides.data()) : nullptr;
+    return __tensor1;
+  }
+
+  ::DLTensor get() const&& = delete;
```
fbusato, line 169: without explicit `= delete` the compiler doesn't raise an error

davebayer (approved):

PR #7034, opened 2025-12-19
[libcu++] Make kernel_config member private and allow it in hierarchy queries

pciolkosz:
Currently members of `kernel_config` are public, make them private and provide a reference getter functions.

We also want the hierarchy queries that accept a hierarchy to optimize them to work with kernel_config, so the user doesn't have to use `config.hierarchy()` and instead the config directly. This PR adds a function that will either pass-through a hierarchy or unpack it from a config. This function was added to hierarchy queries to make them handle a config argument

### DIFF
--- a/cub/cub/block/block_adjacent_difference.cuh
+++ b/cub/cub/block/block_adjacent_difference.cuh
@@ -213,7 +213,7 @@ public:
       , linear_tid(RowMajorTid(BlockDimX, BlockDimY, BlockDimZ))
   {}
 
-  //! @} end member group
+  //! @}
   //! @name Read left operations
   //! @{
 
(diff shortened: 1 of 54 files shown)


libcudacxx/include/cuda/__hierarchy/hierarchy_level_base.h
davebayer:
```suggestion
      __ret[__i] = static_cast<_Index>(_Ret::static_extent(__i));
```

davebayer:
```suggestion
      __ret[__i] = static_cast<_Index>(_OrgExts::static_extent(__i));
```

PR #7047, opened 2025-12-23
`DLPack` to `mdspan`

fbusato:
## Description

The PR implements conversion utilities that take a [DLTensor view](https://dmlc.github.io/dlpack/latest/c_api.html) and produce a (host/device/managed) mdspan of the same underlying memory.

The opposite conversion is implemented in [mdspan to DLPack #7027](https://github.com/NVIDIA/cccl/pull/7027). #7027 is also a prerequisite of this PR. 

Todo:

- [x] documentation

### DIFF
--- a/libcudacxx/include/cuda/__mdspan/dlpack_to_mdspan.h
+++ b/libcudacxx/include/cuda/__mdspan/dlpack_to_mdspan.h
@@ -35,7 +35,9 @@
 #  include <cuda/std/cstdint>
 #  include <cuda/std/mdspan>
 
-#  include <stdexcept>
+#  if !_CCCL_COMPILER(NVRTC)
+#    include <stdexcept>
+#  endif // !_CCCL_COMPILER(NVRTC)
 
 #  include <dlpack/dlpack.h>
 //
@@ -185,10 +187,8 @@ __to_mdspan(const ::DLTensor& __tensor)
       {
         _CCCL_THROW(::std::invalid_argument{"DLTensor shape must be non-null"});
       }
-      using ::cuda::std::int64_t;
-      using ::cuda::std::size_t;
-      ::cuda::std::array<int64_t, _Rank> __extents_array{};
-      for (size_t __i = 0; __i < _Rank; ++__i)
+      ::cuda::std::array<::cuda::std::int64_t, _Rank> __extents_array{};
+      for (::cuda::std::size_t __i = 0; __i < _Rank; ++__i)
       {
         if (__tensor.shape[__i] < 0)
         {
@@ -199,8 +199,8 @@ __to_mdspan(const ::DLTensor& __tensor)
       ::cuda::__validate_dlpack_strides<_LayoutPolicy>(__tensor, _Rank);
       if constexpr (__is_layout_stride)
       {
-        ::cuda::std::array<int64_t, _Rank> __strides_array{};
-        for (size_t __i = 0; __i < _Rank; ++__i)
+        ::cuda::std::array<::cuda::std::int64_t, _Rank> __strides_array{};
+        for (::cuda::std::size_t __i = 0; __i < _Rank; ++__i)
         {
           const bool __has_strides = __tensor.strides != nullptr;
           __strides_array[__i] =


oleksandr-pavlyk:
I feel very strongly that if `std::layout_stride` can not represent strides that can be encountered in `DLPack`, i.e., non-positive strides, then CCCL must implement a layout object that can. 

@mhoemmen Are there reasons this can not be done?

mhoemmen:
@oleksandr-pavlyk wrote:
> I feel very strongly that if `std::layout_stride` can not represent strides that can be encountered in DLPack, i.e., non-positive strides, then CCCL must implement a layout object that can.

`std::layout_stride::mapping` CANNOT represent nonpositive strides.

The design intent of `std::layout_stride::mapping` is to represent any layout mapping resulting from one or more applications of `submdspan` to a `layout_left` or `layout_right` mapping.  It's NOT a general strided layout mapping.  It does NOT support broadcasting or negative strides.

This was always part of the design.  It's something `mdspan`'s layouts inherited from `Kokkos::View`.  It's also part of the reason why the layout mapping requirements define a strided layout mapping (see `is_strided()`) separately from `layout_stride`.  Relaxing this would break `submdspan`.

Therefore, I agree that if you want a layout mapping that can represent everything that DLPack can represent, then you'll need a custom layout mapping.

You'll have this issue and worse with NumPy's `ndarray` format, because that uses byte strides instead of element strides.  Supporting that natively would require a custom accessor as well.  PR https://github.com/kokkos/mdspan/pull/249 shows an example.

leofang:
I agree with @oleksandr-pavlyk:
> I feel very strongly that if `std::layout_stride` can not represent strides that can be encountered in `DLPack`, i.e., non-positive strides, then CCCL must implement a layout object that can.

@mhoemmen If strides must be positive, the whole conversion DLPack <-> mdspan is useless to the Python land (which this and the other PR are targeting, IIUC), it's even dangerous to use, because errors due to nonpositive strides would only happen at run-time, not at compile-time. Not that I need these two PRs merged asap, but knowing this I wouldn't even bother considering it for CuPy or cuda-core.

What is the solution, then? For example
> Therefore, I agree that if you want a layout mapping that can represent everything that DLPack can represent, then you'll need a custom layout mapping.

What does a custom layout mapping entail, exactly? Maybe there is something obvious that I am missing here, due to my lack of full understanding in mdspan 😛 Also, 

> Relaxing this would break `submdspan`.

Is `submdspan` a must-have for DLPack / Python users? I get that it is a thing that C++ users need (due to the standardization), but I can't see how we'd use it in Python. Assuming we are allowed to not worry about `submdspan`, is there any other reason that requires positive strides?

docs/libcudacxx/extended_api/mdspan/dlpack_to_mdspan.rst
```diff
@@ -0,0 +1,129 @@
...
+
+Conversion functions
+--------------------
+
+.. code:: cuda
+
+   namespace cuda {
+
+   template <typename ElementType, size_t Rank, typename LayoutPolicy = cuda::std::layout_stride>
+   [[nodiscard]] cuda::host_mdspan<ElementType, cuda::std::dextents<int64_t, Rank>, LayoutPolicy>
```
davebayer:
```suggestion
   [[nodiscard]] cuda::host_mdspan<ElementType, cuda::std::dims<Rank, int64_t>, LayoutPolicy>
```
Let's use the modern approach

docs/libcudacxx/extended_api/mdspan/mdspan_to_dlpack.rst
```diff
@@ -0,0 +1,134 @@
...
+Conversion functions
+--------------------
+
+.. code:: cuda
+
+   namespace cuda {
+
+   template <typename T, typename Extents, typename Layout, typename Accessor>
+   [[nodiscard]] dlpack_tensor<Extents::rank()>
+   to_dlpack(const cuda::host_mdspan<T, Extents, Layout, Accessor>& mdspan);
```
davebayer: I believe this should be `to_dlpack_tensor`, right?

docs/libcudacxx/extended_api/mdspan/mdspan_to_dlpack.rst
```diff
@@ -0,0 +1,134 @@
...
+
+   template <typename T, typename Extents, typename Layout, typename Accessor>
+   [[nodiscard]] dlpack_tensor<Extents::rank()>
+   to_dlpack(const cuda::device_mdspan<T, Extents, Layout, Accessor>& mdspan,
+                    cuda::device_ref device = cuda::device_ref{0});
```
davebayer: What do we need the `device` parameter for? We should be able to find out the device from the pointer attributes

davebayer: And this should definitely not be defaulted to device 0

docs/libcudacxx/extended_api/mdspan/mdspan_to_dlpack.rst
```diff
@@ -0,0 +1,134 @@
...
+      const DLTensor& get() const noexcept;
+  };
+
+  } // namespace cuda
+
+``cuda::dlpack_tensor`` stores a ``DLTensor`` and owns the backing storage for its ``shape`` and ``strides`` pointers. The class does not use any heap allocation.
+
+.. note:: **Lifetime**
+
+  The ``DLTensor`` associated with ``cuda::dlpack_tensor`` must not outlive the wrapper. If the wrapper is destroyed, the returned ``DLTensor::shape`` and ``DLTensor::strides`` pointers will dangle.
```
davebayer: Not just destroyed, even move invalidates the pointers.

docs/libcudacxx/extended_api/mdspan/mdspan_to_dlpack.rst
```diff
@@ -0,0 +1,134 @@
...
+
+Constraints
+-----------
+
+- The accessor ``data_handle_type`` must be a pointer type.
+
+Runtime errors
+--------------
+
+- If any ``extent(i)`` or ``stride(i)`` cannot be represented in ``int64_t``, the conversion raises an exception.
```
davebayer: We should specify which exception is thrown

libcudacxx/include/cuda/__mdspan/dlpack_to_mdspan.h
```diff
@@ -0,0 +1,261 @@
...
+  // TODO: add support for layout_right_padded and layout_left_padded
+  if constexpr (!__is_layout_right && !__is_layout_left && !__is_layout_stride)
+  {
+    static_assert(::cuda::std::__always_false_v<_LayoutPolicy>, "Unsupported layout policy");
+    _CCCL_UNREACHABLE();
+    return __mdspan_type{};
+  }
+  else
+  {
+    if (__tensor.ndim != int{_Rank})
```
davebayer: you can use `cuda::std::cmp_not_equal` here

libcudacxx/include/cuda/__mdspan/mdspan_to_dlpack.h
```diff
@@ -0,0 +1,303 @@
...
+
+#if defined(_CCCL_IMPLICIT_SYSTEM_HEADER_GCC)
+#  pragma GCC system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_CLANG)
+#  pragma clang system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_MSVC)
+#  pragma system_header
+#endif // no system header
+
+#if !_CCCL_COMPILER(NVRTC) && _CCCL_HAS_INCLUDE(<dlpack/dlpack.h>)
```
davebayer: I think we should provide a way to disable dlpack if someone doesn't want to use it for some reason

docs/libcudacxx/extended_api/mdspan/dlpack_to_mdspan.rst
fbusato: I don't see any difference...but ok

docs/libcudacxx/extended_api/mdspan/mdspan_to_dlpack.rst
fbusato: > We should be able to find out the device from the pointer attributes

that's a nice idea

docs/libcudacxx/extended_api/mdspan/mdspan_to_dlpack.rst
fbusato: this is a comment for the other PR (opposite direction)

libcudacxx/include/cuda/__mdspan/mdspan_to_dlpack.h
fbusato: these APIs are not available if `<dlpack/dlpack.h>` is not present. It looks not necessary.

miscco: The request is the opposite, we should not include it if the user does not want it

mhoemmen:
@leofang wrote:
> What does a custom layout mapping entail, exactly? 

A "custom layout mapping" is a user-defined type that meets the [layout mapping requirements](https://eel.is/c++draft/mdspan.layout.reqmts).

[This file](https://github.com/kokkos/mdspan/blob/5d4eb209c77f4744980c0b0c2af44636cc81b08b/tests/libcxx-backports/CustomTestLayouts.h) in the reference mdspan implementation's tests has examples.

It should be possible for us to write a custom layout mapping that supports arbitrary DLPack layouts.   It would need to store an offset as well as strides, so that negative strides would still result in a nonnegative mapping result.

> Is submdspan a must-have for DLPack / Python users? I get that it is a thing that C++ users need (due to the standardization), but I can't see how we'd use it in Python. Assuming we are allowed to not worry about submdspan, is there any other reason that requires positive strides?

* `layout_stride` needs nonzero strides if all the extents are nonzero, because otherwise the mapping would not be unique.  The C++ Standard specifies that `layout_stride::mapping` is always unique.  This would have nothing to do with `submdspan`.

* `layout_stride` needs nonnegative strides because otherwise evaluating the mapping could give a negative result.  That would violate the layout mapping requirements.  This would have nothing to do with `submdspan`.

The following paper explains these issues: https://isocpp.org/files/papers/P3959R0.html .

mhoemmen:
Here is a rough draft of a custom layout mapping that would support DLPack's layout (including zero and negative strides): https://godbolt.org/z/WEYazcsxT .  I've commented out `submdspan_mapping` so the type doesn't yet support `submdspan`, but it's still a legal layout mapping.  Getting it to support `submdspan` wouldn't be too hard.

```c++
namespace std {
  // work around issue in single header reference implementation
  using ::std::experimental::dims; 

  namespace impl {
    // [mdspan.layout.stride.expo] 2 defines OFFSET(m).
    // It's only ever applied to strided layout mappings.
    template<class Mapping>
      requires(Mapping::is_always_strided())
    constexpr auto offset(const Mapping& m) {
      if constexpr (typename Mapping::extents_type::rank() == 0) {
        return m();
      }
      else {
        using index_type = typename Mapping::index_type;
        constexpr auto rank = typename Mapping::extents_type::rank();     
        bool any_zero = false;
        for (::std::size_t r = 0; r < rank; ++r) {
          any_zero = any_zero || (m.extents().extent(r) == 0);
        }
        if (any_zero) {
          return index_type(0);
        }
        else {
          constexpr auto zeros =
            []< ::std::size_t... Rs>(std::index_sequence<Rs...>) {
              return std::tuple{((void) Rs, index_type(0))...};
            } (std::make_index_sequence<rank>());
          return std::apply(m, zeros);
        }
      }
    }
  }

  class layout_stride_relaxed {
  public:
    template<class Extents>
    class mapping;
  };

  template<class Extents>
  class layout_stride_relaxed::mapping {
  public:
    using extents_type = Extents;
    using index_type = extents_type::index_type;
    using size_type = extents_type::size_type;
    using rank_type = extents_type::rank_type;
    using layout_type = layout_stride_relaxed;

  private:
    static constexpr rank_type rank_ = extents_type::rank();

  public:
    constexpr mapping() noexcept {
      const layout_right::mapping<extents_type> map{};
      for (std::size_t d = 0; d < rank_; ++d) {
        strides_[d] = map.stride(d);
      }
    }

    constexpr mapping(const mapping&) noexcept = default;

    template<class OtherIndexType>
    requires(
      ::std::is_convertible_v<const OtherIndexType&, ::std::intptr_t> &&
      ::std::is_nothrow_constructible_v<::std::intptr_t, const OtherIndexType&>
    )
    constexpr mapping(
        const extents_type& e,
        ::std::span<OtherIndexType, rank_> s,
        ::std::size_t offset = 0) noexcept
      : extents_(e), offset_(offset)
    {
      for (::std::size_t d = 0; d < rank_; ++d) {
        strides_[d] = s[d];
      }
    }

    template<class OtherIndexType>
    requires(
      is_convertible_v<const OtherIndexType&, ::std::intptr_t> &&
      is_nothrow_constructible_v<::std::intptr_t, const OtherIndexType&>
    )
    constexpr mapping(
        const extents_type& e,
        const ::std::array<OtherIndexType, rank_>& s,
        ::std::size_t offset = 0) noexcept
      : extents_(e), offset_(offset)
    {
      for (::std::size_t d = 0; d < rank_; ++d) {
        strides_[d] = s[d];
      }
    }

    // m IS a layout_stride_relaxed::mapping
    template<class OtherMapping>
    requires(
      detail::layout_mapping_alike<OtherMapping> &&
      ::std::is_constructible_v<
        extents_type,
        typename OtherMapping::extents_type
      > &&
      ::std::is_same_v<
        layout_type,
        typename OtherMapping::layout_type
      >
    )
    constexpr explicit(
      ! (
        ::std::is_convertible_v<
          typename OtherMapping::extents_type, extents_type
        >
      )
    )
    mapping(const OtherMapping& m) noexcept
      : extents_(m.extents()), offset_(m.offset_)
    {
      for (std::size_t d = 0; d < rank_; ++d) {
        strides_[d] = m.stride(d);
      }
    }

    // m is NOT a layout_stride_relaxed::mapping
    template<class StridedLayoutMapping>
    requires(
      detail::layout_mapping_alike<StridedLayoutMapping> &&
      ::std::is_constructible_v<
        extents_type,
        typename StridedLayoutMapping::extents_type> &&
      StridedLayoutMapping::is_always_unique() &&
      StridedLayoutMapping::is_always_strided()
    )
    constexpr explicit(
      ! (
        ::std::is_convertible_v<
          typename StridedLayoutMapping::extents_type, extents_type
        > && (
          detail::is_mapping_of<layout_left, StridedLayoutMapping> ||
          detail::is_mapping_of<layout_right, StridedLayoutMapping> ||
          experimental::detail::is_layout_left_padded_mapping<
            StridedLayoutMapping>::value ||
          experimental::detail::is_layout_right_padded_mapping<
            StridedLayoutMapping>::value ||
          detail::is_mapping_of<layout_stride, StridedLayoutMapping>
        )
      )
    )
    mapping(const StridedLayoutMapping& m) noexcept
      : extents_(m.extents())
    {
      for (std::size_t d = 0; d < rank_; ++d) {
        strides_[d] = m.stride(d);
      }
    }
 
    constexpr mapping& operator=(const mapping&) noexcept = default;

    // [mdspan.layout.stride.obs], observers
    constexpr const extents_type& extents() const noexcept {
      return extents_;
    }
    constexpr ::std::array<index_type, rank_> strides() const noexcept {
      return strides_;
    }
    constexpr ::std::intptr_t offset() const noexcept {
      return offset_;
    }

    constexpr index_type required_span_size() const noexcept {
      // The dot product of indices and strides is linear.
      // Thus, over all valid indices, the max value of the
      // dot product is achieved at the extrema: either the
      // min index (0) if the stride is negative, or the max
      // index (extent(r) - 1) if the stride is nonnegative.
      std::array<index_type, rank_> max_indices{};
      for (std::size_t r = 0; r < rank_; ++r) {
        const index_type ext = extents_.extent(r);
        const index_type ext_minus_1 =
          ext == 0 ? index_type(0) : ext - index_type(1);
        max_indices[r] = strides_[r] < 0 ? index_type(0) : ext_minus_1;
      }
      index_type dot = 0;
      for (std::size_t r = 0; r < rank_; ++r) {
        dot += max_indices[r] * strides_[r];
      }
      return offset() + dot;
    }

    template<class... Indices>
    requires(
      sizeof...(Indices) == rank_ &&
      (::std::is_convertible_v<Indices, index_type> && ...) &&
      (::std::is_nothrow_constructible_v<index_type, Indices> && ...)
    )
    constexpr index_type operator()(Indices... inds) const noexcept {
      return offset() +
        [&, this]<::std::size_t... Rs>(::std::index_sequence<Rs...>) {
          return ((inds...[Rs] * strides_[Rs]) + ... + index_type(0));
        } (::std::make_index_sequence<rank_>());
    }

    static constexpr bool is_always_unique() noexcept { return false; }
    static constexpr bool is_always_exhaustive() noexcept { return false; }
    // It's technically NOT always strided, because of the offset
    // (to accommodate negative strides)
    static constexpr bool is_always_strided() noexcept { return false; }

    constexpr bool is_unique() noexcept {
      // The Standard doesn't require that this be exact.
      // Possibility of negative strides with an offset
      // makes that harder to figure out.
      return false;  
    }
    constexpr bool is_exhaustive() const noexcept {
      // The Standard doesn't require that this be exact.
      // Possibility of negative strides with an offset
      // makes that harder to figure out.
      return false;  
    }
    constexpr bool is_strided() noexcept {
      return offset_ == 0;
    }

    constexpr index_type stride(rank_type i) const noexcept {
      return strides_[i];
    }

    // y is also a layout_stride_relaxed::mapping
    template<class OtherMapping>
    requires(
      detail::layout_mapping_alike<OtherMapping> &&
      rank_ == OtherMapping::extents_type::rank() &&
      ::std::is_same_v<layout_type, typename OtherMapping::layout_type>
    )
    friend constexpr bool
    operator==(const mapping& x, const OtherMapping& y) noexcept {
      return x.extents() == y.extents() &&
      x.offset_ == y.offset_ &&
      [&]<::std::size_t...Rs> (::std::index_sequence<Rs...>) {
        return ((x.stride(Rs) == y.stride(Rs)) && ...);
      } (::std::make_index_sequence<rank_>());
    }

    // y is NOT a layout_stride_relaxed::mapping but is strided.
    template<class OtherMapping>
    requires(
      detail::layout_mapping_alike<OtherMapping> &&
      rank_ == OtherMapping::extents_type::rank() &&
      OtherMapping::is_always_strided()
    )
    friend constexpr bool
    operator==(const mapping& x, const OtherMapping& y) noexcept {
      return x.extents() == y.extents() &&
      impl::offset(y) == x.offset_ &&
      [&]<::std::size_t...Rs> (::std::index_sequence<Rs...>) {
        return ((x.stride(Rs) == y.stride(Rs)) && ...);
      } (::std::make_index_sequence<rank_>());
    }

  private:
    extents_type extents_{};
    std::intptr_t offset_ = 0;
    array<std::intptr_t, rank_> strides_{};

#if 0
    // [mdspan.sub.map], submdspan mapping specialization
    template<class... SliceSpecifiers>
      constexpr auto submdspan-mapping-impl(SliceSpecifiers...) const
        -> /* see-below */;

    template<class... SliceSpecifiers>
      friend constexpr auto submdspan_mapping(
        const mapping& src, SliceSpecifiers... slices) {
          return src.submdspan-mapping-impl(slices...);
      }
#endif // 0
  };
} // namespace std

int main() {
  std::dims<3> exts(3, 5, 11);  
  std::array<std::intptr_t, 3> strides{0, 1, 5}; // broadcasting
  std::layout_stride_relaxed::mapping<std::dims<3>> map(exts, strides);

  assert(map(0, 1, 1) == map(1, 1, 1));

  return 0;
}
```

docs/libcudacxx/extended_api/mdspan/dlpack_to_mdspan.rst
```diff
@@ -0,0 +1,138 @@
...
+Conversion functions
+--------------------
+
+.. code:: cuda
+
+   namespace cuda {
+
+   template <typename ElementType, size_t Rank, typename LayoutPolicy = cuda::std::layout_stride>
+   [[nodiscard]] cuda::host_mdspan<ElementType, cuda::std::dims<Rank, int64_t>, LayoutPolicy>
+   to_host_mdspan(const DLTensor& tensor);
```
davebayer, line 19: Q: would it make sense to take `_Extents` instead of `_Rank` as the second parameter? Could it be the case that you know some dimensions at compile time and only want to fill the dynamic dimensions?

It could be the same story with the layout.

It would look as:
```cpp
// 1. same as is now
auto my_mdspan1 = cuda::to_host_mdspan<int>(tensor);

// 2. same as now, but uses 32 bits for dimensions
auto my_mdspan2 = cuda::to_host_mdspan<int, cuda::std::dims<3, int32_t>>(tensor);

// 3. with dimensions known at compile time
auto my_mdspan3 = cuda::to_host_mdspan<int, cuda::std::extents<int64_t, 2, cuda::std::dynamic_extent, 128>>(tensor);

// 4. with custom dimensions and forced layout left in 32-bit 
auto my_mdspan4 = cuda::to_host_mdspan<int, cuda::std::extents<int32_t, 2, cuda::std::dynamic_extent, 128>, cuda::std::layout_left>(tensor);
```

What do you think, is it too complicated? I'm also not sure if someone would want to use it. @leofang would there be a usecase for this?

docs/libcudacxx/extended_api/mdspan/dlpack_to_mdspan.rst
```diff
@@ -0,0 +1,138 @@
...
+- ``DLTensor::strides`` are not compatible with the requested ``layout_right`` or ``layout_left``.
+- ``DLTensor::device.device_type`` does not match the target mdspan type.
+- Data pointer is not properly aligned for the element type.
+
+Availability notes
+------------------
+
+- This API is available only when DLPack header is present, namely ``<dlpack/dlpack.h>`` is found in the include path.
+- Requires DLPack major version 1.
+
```
davebayer, line 93: Can we also add the macro to disable this as we do in mdspan -> dlpack

libcudacxx/include/cuda/__internal/dlpack.h
```diff
@@ -26,7 +26,7 @@
 #  include <dlpack/dlpack.h>
 
 #  define _CCCL_DLPACK_AT_LEAST(_MAJOR, _MINOR) \
-    (DLPACK_MAJOR_VERSION > (_MAJOR) || (DLPACK_MAJOR_VERSION == (_MAJOR) && DLPACK_VERSION_MINOR >= (_MINOR)))
```
davebayer, line 29: omg 🤦

docs/libcudacxx/extended_api/mdspan/dlpack_to_mdspan.rst
fbusato: I don't think it is worth it. Dynamic extents is the general case. The user could always convert the output to a compile-time extents in a second step...but I think it is rarer.

libcudacxx/include/cuda/__internal/dlpack.h
fbusato: ahah, I was extremely lucky to catch this bug

docs/libcudacxx/extended_api/mdspan/dlpack_to_mdspan.rst
davebayer: Okay! We can always add it later if needed.

libcudacxx/include/cuda/__internal/dlpack.h
davebayer: Thanks

libcudacxx/include/cuda/__mdspan/dlpack_to_mdspan.h
```diff
@@ -0,0 +1,268 @@
...
+}
+
+[[nodiscard]]
+_CCCL_HOST_API inline ::cuda::std::int64_t
+__get_layout_right_stride(const ::cuda::std::int64_t* __shapes, ::cuda::std::size_t __pos, ::cuda::std::size_t __rank)
+{
+  ::cuda::std::int64_t __stride = 1;
+  for (auto __i = __pos + 1; __i < __rank; ++__i)
+  {
+    if (__stride * __shapes[__i] < 0 || ::cuda::mul_hi(__stride, __shapes[__i]) != 0) // TODO: replace with mul_overflow
```
davebayer: I believe this multiplication can overflow in a way that `__stride * __shapes[__i] < 0` would be `false`, this should be:
```suggestion
    if (const auto __hi = ::cuda::mul_hi(__stride, __shapes[__i]); __hi != 0 && __hi != -1) // TODO: replace with mul_overflow
```

fbusato: `__stride` and `__shapes[__i]` are both `int64_t`.

- if `mul_hi() != 0`, it is clearly an overflow.
- The subtle case is when  `__stride * __shapes[__i] > 2^63-1`. In this case, the product is `<= 0` and `__hi != -1`.

libcudacxx/include/cuda/__mdspan/dlpack_to_mdspan.h
```diff
@@ -0,0 +1,270 @@
...
+#  include <cuda/__memory/is_aligned.h>
+#  include <cuda/std/__cstddef/types.h>
+#  include <cuda/std/__exception/exception_macros.h>
+#  include <cuda/std/__type_traits/is_same.h>
+#  include <cuda/std/__utility/cmp.h>
+#  include <cuda/std/array>
+#  include <cuda/std/cstdint>
+#  include <cuda/std/mdspan>
+
+#  include <stdexcept>
```
miscco: Please do *not* include standard library headers unguarded. 

EDIT: I talked to @davebayer and he will create a followup to centralize the `<stdexcept>` include. We do expect this to be needed to work with NVRTC soon, so a open std-include is not possible

libcudacxx/include/cuda/__mdspan/dlpack_to_mdspan.h
```diff
@@ -0,0 +1,270 @@
...
+      return __mdspan_type{__data, __mapping_type{}};
+    }
+    else // Rank > 0
+    {
+      if (__tensor.shape == nullptr)
+      {
+        _CCCL_THROW(::std::invalid_argument{"DLTensor shape must be non-null"});
+      }
+      using ::cuda::std::int64_t;
+      using ::cuda::std::size_t;
```
miscco: Nitpick: I would prefer if we would not pull in those, especially if they are only used once or twice

PR #7048, opened 2025-12-25
Policy_with_memory_resource

miscco:
This implements a way to pass a stream and a memory resource along with an execution policy


miscco:
closing in favor of #7299

PR #7049, opened 2025-12-26
Enhance RDC detection and add `_CCCL_HAS_DEVICE_RUNTIME()` macro

davebayer:
Currently, we are mixing several things together. This PR is trying to put things in order and extend RDC support.

Changes:
1. The `_CCCL_HAS_RDC()` macro just reflects whether RDC is being generated. This macro will no longer be disabled by defining `CCCL_DISABLE_CDP` or `CUB_DISABLE_CDP`.
2. The `_CCCL_HAS_EWP()` macro expands to `1` if the `-ewp` (extensive whole program) option was passed to the compiler.
3. New `_CCCL_HAS_DEVICE_RUNTIME()` macro that expands to `1` if device runtime (`cuda_device_runtime_api.h`) stuff can be used by CCCL. This option requires RDC or EWP and can be disabled by the user by defining `CCCL_DISABLE_DEVICE_RUNTIME`.
4. New `_CCCL_HAS_CDP()` macro that expands to `1` if CUDA dynamic parallelism is enabled. Requires `_CCCL_HAS_DEVICE_RUNTIME()` to be `1`

### DIFF
--- a/libcudacxx/include/cuda/__nvtx/nvtx.h
+++ b/libcudacxx/include/cuda/__nvtx/nvtx.h
@@ -94,19 +94,26 @@ _CCCL_END_NAMESPACE_CUDA
 #    define _CCCL_BEFORE_NVTX_RANGE_SCOPE(name)
 #  endif // !CCCL_DETAIL_BEFORE_NVTX_RANGE_SCOPE
 
+#  if _CCCL_HOST_COMPILATION()
 // Conditionally inserts a NVTX range starting here until the end of the current function scope in host code. Does
 // nothing in device code.
 // The optional is needed to defer the construction of an NVTX range (host-only code) and message string registration
 // into a dispatch region running only on the host, while preserving the semantic scope where the range is declared.
-#  define _CCCL_NVTX_RANGE_SCOPE_IF(condition, name)                                                               \
-    _CCCL_BEFORE_NVTX_RANGE_SCOPE(name)                                                                            \
-    ::cuda::std::optional<::nvtx3::v1::scoped_range_in<::cuda::detail::NVTXCCCLDomain>> __cuda_nvtx3_range;        \
-    NV_IF_TARGET(                                                                                                  \
-      NV_IS_HOST,                                                                                                  \
-      static const ::nvtx3::v1::registered_string_in<::cuda::detail::NVTXCCCLDomain> __cuda_nvtx3_func_name{name}; \
-      static const ::nvtx3::v1::event_attributes __cuda_nvtx3_func_attr{__cuda_nvtx3_func_name};                   \
-      if (condition) __cuda_nvtx3_range.emplace(__cuda_nvtx3_func_attr);                                           \
-      (void) __cuda_nvtx3_range;)
+#    define _CCCL_NVTX_RANGE_SCOPE_IF(condition, name)                                                                 \
+      _CCCL_BEFORE_NVTX_RANGE_SCOPE(name)                                                                              \
+      ::cuda::std::optional<::nvtx3::v1::scoped_range_in<::cuda::detail::NVTXCCCLDomain>> __cuda_nvtx3_range;          \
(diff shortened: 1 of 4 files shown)


libcudacxx/include/cuda/std/__cccl/cuda_capabilities.h
```diff
@@ -48,4 +48,45 @@
...
+#if defined(__CUDACC_EWP__)
+#  define _CCCL_HAS_EWP() 1
+#else // ^^^ has EWP ^^^ / vvv no EWP vvv
+#  define _CCCL_HAS_EWP() 0
+#endif // ^^^ no EWP ^^^
+
+// Control whether device runtime APIs can be used, because they require libcudadevrt to be linked. Defaults to true
+// when RDC or EWP are enabled. Can be disabled by defining CCCL_DISABLE_DEVICE_RUNTIME.
+#if (_CCCL_HAS_RDC() || _CCCL_HAS_EWP()) && !defined(CCCL_DISABLE_DEVICE_RUNTIME)
+#  define _CCCL_HAS_DEVICE_RUNTIME() 1
```
fbusato, line 68: relocatable device code does not strictly require `libcudadevrt`, but only for CDP

davebayer: I don't understand what do you mean. Using CUDA device runtime APIs require either RDC or EWP + link `libcudadevrt`. This macro returns `1` if device runtime (stuff from `<cuda_device_runtime_api.h>`) can be used by CCCL, that means if RDC or EWP are enabled. The user has the option to explicitly disable device runtime (in CCCL) by defining `CCCL_DISABLE_DEVICE_RUNTIME`. CDP calls `cudaDeviceLaunch` which is one of the device runtime APIs, thus CDP depends on `_CCCL_HAS_DEVICE_RUNTIME()`

fbusato: maybe I'm reading it in the wrong way.

> - The user has the option to explicitly disable device runtime (in CCCL) by defining CCCL_DISABLE_DEVICE_RUNTIME. 
> - CDP depends on _CCCL_HAS_DEVICE_RUNTIME()

these are perfectly fine.

My issue is about RDC, relocatable device code. If we have RDC, this doesn't automatically translate in having `_CCCL_HAS_DEVICE_RUNTIME`. Am I missing something?

davebayer: It does not, because the user needn't to link `libcudadevrt`

libcudacxx/include/cuda/__nvtx/nvtx.h
```diff
@@ -94,19 +94,26 @@ _CCCL_END_NAMESPACE_CUDA
...
+          static const ::nvtx3::v1::registered_string_in<::cuda::detail::NVTXCCCLDomain> __cuda_nvtx3_func_name{name}; \
+          static const ::nvtx3::v1::event_attributes __cuda_nvtx3_func_attr{__cuda_nvtx3_func_name};                   \
+          if (condition)                                                                                               \
+          {                                                                                                            \
+            __cuda_nvtx3_range.emplace(__cuda_nvtx3_func_attr);                                                        \
+          }                                                                                                            \
+        }))
+#  else // ^^^ _CCCL_HOST_COMPILATION() ^^^ / vvv !_CCCL_HOST_COMPILATION() vvv
+#    define _CCCL_NVTX_RANGE_SCOPE_IF(condition, name)
+#  endif // ^^^ !_CCCL_HOST_COMPILATION() ^^^
```
bernhardmgruber, line 116: Remark: this looks a bit weird, but I think it's correct. For nvc++, `_CCCL_HOST_COMPILATION` is defined for the combined host/device path, so we still need the `NV_IF_TARGET` inside.

bernhardmgruber (approved):

PR #7051, opened 2025-12-30
Initial `nvrtcc` implementation

### LINKED ISSUE #7178: [FEA]: Initial `nvrtcc` implementation
Create a working prototype that compiles simple examples.

davebayer:
Closes #7178.

### DIFF
--- a/nvrtcc/bin/nvrtcc
+++ b/nvrtcc/bin/nvrtcc
@@ -294,7 +294,6 @@ class Nvrtcc:
 
         return parser.parse_known_args(nvcc_opts)[0]
 
-
     def __init__(self, opts: list[str], env):
         args, nvcc_opts = Nvrtcc._parse_nvrtcc_opts(opts)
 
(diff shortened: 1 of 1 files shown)


nvrtcc/src/nvrtcc_device_compiler.cpp
```diff
@@ -0,0 +1,378 @@
...
+          add_symbol(prog, symbol);
+        }
+      }
+
+      ptx_input.remove_prefix(line.size());
+    }
+  }
+}
+
+int main(int argc, const char* const* argv)
```
wmaxey, line 289: Check that argc has an expected minimum value.

davebayer: I thought it isn't necessary since we call it only from the script, but I can add this for sure

wmaxey: Eh, I guess if we're pretending to be NVCC someone could just poke us with `-v` or maybe nothing at all.

PR #7059, opened 2025-12-31
[BUG] use references for mdspan internal methods

### LINKED ISSUE #7057: [BUG]: mdspan property checks are no longer constexpr
### Is this a duplicate?

### Type of Bug

Compile-time Error

### Component

libcu++

### Describe the bug

The `is_unique`, `is_exhaustive`, `is_strided`, and `stride` methods of `mdspan` (defined [here](https://github.com/NVIDIA/cccl/blob/38351635dc71f28655c0c036e54491cbf3f69eaf/libcudacxx/include/cuda/std/__mdspan/mdspan.h#L449)) are marked constexpr. However, the clang workaround introduced to `is_exhaustive` in https://github.com/NVIDIA/cccl/pull/5939 and propagated to the others in https://github.com/NVIDIA/cccl/pull/6703 means that these methods are no longer actually safe to call in a constexpr context. The code now looks like this:
```cpp
    auto __tmp = mapping(); // workaround for clang with nodiscard  <<< This is a problem
    return __tmp.is_unique();
```
Assigning to `__tmp` requires runtime storage, resulting in errors like 
```bash
/path/to/libcudacxx/include/cuda/std/__mdspan/mdspan.h(463): note #2701-D: attempt to access run-time storage
      auto __tmp = mapping();
```

### How to Reproduce

The godbolt link contains a complete reproduction. Here is the code:
```cpp
#include <cuda/std/mdspan>
  
int main() {  
    constexpr char data[] = {'H', 'O', 'P', 'P', 'E', 'R'};  
    cuda::std::mdspan md{data, cuda::std::layout_left::mapping{cuda::std::dims<1>{6}}};    
    const auto slice = cuda::std::pair{1, 4};
    cuda::std::mdspan sub = cuda::std::submdspan(md, slice);  
    static_assert(sub.is_strided());
}
```
This fails on the trunk of CCCL but passes on older versions. Specifically in godbolt you can select the 3.1 release and see that it passes, but any commit prior to https://github.com/NVIDIA/cccl/commit/bae1a71f2174cabe86f8bc0c6424c10773f7d159 should work as well.

### Expected behavior

These functions should be safe to call in a constexpr context as advertised.

### Reproduction link

https://godbolt.org/z/9WP6GT6Ez

### Operating System

Ubuntu 24.04

### nvidia-smi output

_No response_

### NVCC version

$ nvcc --version
nvcc: NVIDIA (R) Cuda compiler driver
Copyright (c) 2005-2025 NVIDIA Corporation
Built on Tue_May_27_02:21:03_PDT_2025
Cuda compilation tools, release 12.9, V12.9.86
Build cuda_12.9.r12.9/compiler.36037853_0

fbusato:
Fixes #7057

## Description

Refactor mdspan methods to use const references for temporary mapping objects to fix https://github.com/NVIDIA/cccl/issues/7057

### DIFF
--- a/libcudacxx/include/cuda/std/__mdspan/mdspan.h
+++ b/libcudacxx/include/cuda/std/__mdspan/mdspan.h
@@ -448,24 +448,24 @@ public:
   [[nodiscard]]
   _CCCL_API constexpr bool is_unique() const noexcept(noexcept(::cuda::std::declval<const mapping_type&>().is_unique()))
   {
-    auto __tmp = mapping(); // workaround for clang with nodiscard
+    const auto& __tmp = mapping(); // workaround for clang with nodiscard
     return __tmp.is_unique();
   }
   [[nodiscard]] _CCCL_API constexpr bool is_exhaustive() const
     noexcept(noexcept(::cuda::std::declval<const mapping_type&>().is_exhaustive()))
   {
-    auto __tmp = mapping(); // workaround for clang with nodiscard
+    const auto& __tmp = mapping(); // workaround for clang with nodiscard
     return __tmp.is_exhaustive();
   }
   [[nodiscard]] _CCCL_API constexpr bool is_strided() const
     noexcept(noexcept(::cuda::std::declval<const mapping_type&>().is_strided()))
   {
-    auto __tmp = mapping(); // workaround for clang with nodiscard
+    const auto& __tmp = mapping(); // workaround for clang with nodiscard
     return __tmp.is_strided();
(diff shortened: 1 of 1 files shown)


bernhardmgruber (approved):

bdice:
@fbusato If you’re able to fix the backport for branch/3.2.x, I’ll test it in RAFT and revert the changes I made to remove constexpr usage.

fbusato:
yes, this is my intention

PR #7062, opened 2026-01-01
Add shuffle iterator implementation, tests, and example

shwina:
## Description

This PR adds a `ShuffleIterator` implementation that uses a [Feistel network](https://en.wikipedia.org/wiki/Feistel_cipher) based permutation function.

Note - the implementation is **completely vibe-coded**, but I was pleased to find that it more or less matches the [libcudacxx approach](https://github.com/NVIDIA/cccl/blob/main/libcudacxx/include/cuda/__random/feistel_bijection.h) .

## Performance

As reported in https://github.com/cupy/cupy/issues/9320, `cp.random.choice(N)` allocates a temporary of size `N` which can be slow and memory-inefficient when randomly drawing a small number of values from a large range. The stateless approach used by `ShuffleIterator` solves that problem:

```python
import cupy as cp
import numpy as np

def run_cupy_choice(N: int, K: int, seed: int):
    out = cp.random.choice(N, size=K, replace=False)
    cp.cuda.runtime.deviceSynchronize()
    return out


def run_cuda_compute_choice(N: int, K: int, seed: int):
    from cuda.compute import ShuffleIterator, unary_transform
    shuffle_it = ShuffleIterator(N, seed)
    out = cp.empty(K, dtype=np.int64)
    unary_transform(shuffle_it, out, lambda x: x, K)
    cp.cuda.runtime.deviceSynchronize()
    return out
```

```python
%timeit run_cupy_choice(100_000_000, 1_000, 42)
34.1 ms ± 16.3 μs per loop (mean ± std. dev. of 7 runs, 10 loops each)

%timeit run_cuda_compute_choice(100_000_000, 1_000, 42)
15.3 μs ± 112 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```

## Checklist

### DIFF
--- a/python/cuda_cccl/cuda/compute/iterators/_shuffle_iterator.py
+++ b/python/cuda_cccl/cuda/compute/iterators/_shuffle_iterator.py
@@ -44,12 +44,6 @@ def _splitmix64_next(state):
 def _feistel_bijection(val, seed, left_bits, right_bits, left_mask, right_mask):
     """
     Feistel bijection matching libcudacxx __feistel_bijection.
-
-    The C++ implementation uses a struct with members named __low and __high,
-    but these names refer to struct position, not bit position. We use the
-    same naming here to match the C++ logic exactly:
-      - state_low  corresponds to C++ __state.__low
-      - state_high corresponds to C++ __state.__high
     """
     M0 = uint64(FEISTEL_M0)
 


python/cuda_cccl/cuda/compute/iterators/_shuffle_iterator.py
```diff
@@ -0,0 +1,216 @@
...
+    - **Deterministic**: the same ``num_items`` and ``seed`` always produce the
+      same ordering.
+    - **Stateless**: no per-element or per-thread state is required.
+    - **Lazy**: indices are computed on demand; no permutation buffer is stored.
+    - **Device-friendly**: implemented using simple integer arithmetic and
+      device-callable functions.
+
+    Limitations
+    -----------
+    - The resulting permutation is *not* uniformly sampled from all
```
RAMitchell: This statement kind of feels like it might be correct but I think its nonsense.

python/cuda_cccl/cuda/compute/iterators/_shuffle_iterator.py
```diff
@@ -0,0 +1,216 @@
...
+      ``num_items!`` permutations. It is drawn from a large, structured family
+      of permutations induced by the Feistel construction.
+    - Cycle-walking may apply the Feistel permutation more than once per element
```
RAMitchell: I think the user doesn't care about the implementation detail of cycle walking. One consequence of this so called "cycle-walking" is the the worst case runtime for generating an element is O(n) where n is the permutation length - it is just vanishing unlikely. I am not sure we even need to mention this.

python/cuda_cccl/cuda/compute/iterators/_shuffle_iterator.py
```diff
@@ -0,0 +1,216 @@
...
+        # Standard Feistel step
+        new_L = R
+        new_R = (L ^ F) & half_mask
+        L = new_L
+        R = new_R
+
+    return (R << hb) | L
+
+
+def _splitmix64_host(x: int) -> int:
```
RAMitchell, line 93: I think using splitmix is in theory fine, but we should use the VariablePhilox algorithm from this paper: https://arxiv.org/abs/2106.06161. This implemention needs to round up the sequence to the nearest power of 4, the one from the paper is the nearest power of 2. And its tested.

python/cuda_cccl/tests/compute/test_shuffle_iterator.py
```diff
@@ -0,0 +1,180 @@
...
+
+    shuffle_it = ShuffleIterator(num_items, seed)
+
+    d_output = cp.empty(num_items, dtype=np.int64)
+    cuda.compute.unary_transform(shuffle_it, d_output, lambda x: x, num_items)
+
+    result = d_output.get()
+
+    # Should be a valid permutation
+    assert len(set(result)) == num_items
```
RAMitchell, line 82: Check the sorted permutation against cp.arange for a faster test.

shwina:
Thanks for the great feedback, @RAMitchell.

I've updated the implementation to use the VariablePhilox algorithm as mentioned; and to match the libcu++ implementation, we do 24 Feistel rounds.

However, I still don't see strictly uniform distribution of permutations using the test you suggested. For `n=4` (24 different permutations), and `samples=24000` (expected 1000 samples per permutation), here's what the distribution looks like:

Screenshot 2026-01-05 at 3 24 18 PM

And as another case, for `n=5`, and `samples=120000`:

Screenshot 2026-01-05 at 3 43 14 PM

---

So, while "good enough for government work", it's not as rigorous as maybe you would like.

Now, I also noted that the C++ implementation suffers from a similar issue.

The validation scripts, along with instructions for how to run them are in this gist: https://gist.github.com/shwina/2508ed08bc7c257dcf1834dfa574d7e6

python/cuda_cccl/cuda/compute/iterators/_factories.py
```diff
@@ -219,14 +220,34 @@ def PermutationIterator(values, indices):
     return make_permutation_iterator(values, indices)
 
 
+def ShuffleIterator(num_items, seed):
```
leofang, line 223: API design: 
- As a library developer, being able to pass my internal `seed` is great
- As an end user, I sometimes don't want to think about it and would prefer if `seed=None` (default), cuda.compute generates a random seed for me internally.

leofang:
I can reproduce the observation in https://github.com/NVIDIA/cccl/pull/7062#issuecomment-3712125756 with https://github.com/cupy/cupy/pull/9570 (re-implementation of @RAMitchell's `thrust::shuffle_iterator`), by reusing @shwina's `validation.py` and replacing this
```python
shuffle_it = ShuffleIterator(n, seed)
cuda.compute.unary_transform(shuffle_it, d_perm, lambda x: x, n)
```
by this
```python
keys = np.random.randint(
    0, 0xFFFFFFFF + 1, size=24, dtype=np.uint32)
bijection = FeistelBijection(n, keys)
d_perm = bijection(n)
```
Since we use different RNGs, this verifies that the observation is **NOT** due to the choice of RNG.

leofang:
(updated my comment above)

RAMitchell:
Thanks @leofang, I've done the same experiment and something is definitely not quite right, either with our Philox implementation or there are simply too few bits at low sizes for this hash to work as intended. I will do some more investigation.

RAMitchell:
I've tried a number of different approaches today, including strengthening the hash functions in the feistel cipher. The thing that matters in the end is just the minimum size of the sequence. 4 bits as a minimum seems to be too low, increasing it to as few as 6 bits seems to give the cipher enough to work with and passes uniformity tests. This means that we need to throw away more values for small sequences but this seems ok.

shwina:
@RAMitchell Thanks for taking a look. To be clear, does this mean we are OK with non strict-uniformity at lower sequence sizes; or do we need changes to the implementation at the C++ level?

Tangentially, I've realized that we don't strictly need to reimplement the functionality in Python. Similar to what we do in `cuda.coop`, it's possible to reuse the existing functionality provided in `feistel_bijection.h` from Python to implement `ShuffleIterator`. So any changes to the C++ implementation will be picked up here automatically. I'll push an update to the PR that uses this implementation soon.

RAMitchell:
No, we can have uniformity. We just have to use e.g. a bijection for 256 elements to generate a permutation of length 5. The tiny bijections don't work well with the feistel cipher.

shwina:
OK, would you like to push a fix for that in a separate PR, or would you like me to do that in this one? (it will probably take me more cycles)

RAMitchell:
I will do it in a separate PR thanks. I would like to add a bunch more tests as well.

python/cuda_cccl/tests/compute/test_shuffle_iterator.py
```diff
@@ -0,0 +1,180 @@
...
+    d_output1 = cp.empty(num_items, dtype=np.int64)
+    d_output2 = cp.empty(num_items, dtype=np.int64)
+
+    cuda.compute.unary_transform(shuffle_it1, d_output1, lambda x: x, num_items)
+    cuda.compute.unary_transform(shuffle_it2, d_output2, lambda x: x, num_items)
+
+    cp.testing.assert_array_equal(d_output1, d_output2)
+
+
+def test_shuffle_iterator_different_seeds():
```
djns99, line 52: Some extra tests I would consider:

1. Check the result != identity permutation
2. Check a few different seeds and assert every value is moved in at least some of them

Lets have at least a couple of basic uniformity test:

- Check all permutations of 5-6 elements is uniformly selected (the only true test)
- Pick one element and check it lands at every location uniformly (ok sanity check for long sequences)

Otherwise lets reuse C++ and rely on their tests

djns99 (commented):
> I've realized that we don't strictly need to reimplement the functionality in Python

Yes please try to reuse the C++ implementation so we don't have to duplicate the correctness tests in python. 
C++ supports custom bijections but I don't think we would need to expose that functionality here

djns99:
> However, I still don't see strictly uniform distribution of permutations using the test you suggested.

I'll note that we don't expect to see a perfectly flat line, we should see a slight S shape. But over 100 difference is definitely high. e.g this is a plot of randomly generated values between 0, 120

random_plot

shwina:
Thank you @djns99 - I can confirm I do see better S shaped curves when reusing the (new, fixed) underlying C++ implementation, with less variance.

I will mark this PR as draft, as I'd like to land some other bits of functionality that will allow us to better reuse the underlying C++ implementation first.

shwina:
Closing this in favor of https://github.com/NVIDIA/cccl/pull/7721.

Issue #6422, opened 2025-10-31
Support free-threaded Python


shwina:

NaderAlAwar:
Free threading requirements:

- [x] Add marker to our cython bindings saying we assert free threading is supported. If we don't do this, importing cuda.compute will reenable the GIL:
  - [x] This requires cython >= 3.1.0. Confirm with Ashwin that this is fine
  - [x] Audit bindings and C code to make sure that threading is safe, specifically shared mutable state in Python, C++, and Cython code: Caches, registries, global handles, JIT/build paths, no gil regions (in cython)
    - [x] Remember transform has that cache for ublkcp
- [x] Add tests:
  - [x] Smoke test that ensures when we import cuda.compute we can `assert not sys._is_gil_enabled()`
  - [x] Threaded stress tests. The CuPy CI uses two threads for 3.14t jobs. It also has explicit threaded tests
    - Kernel cache threaded stress tests in `cupy/tests/cupy_tests/core_tests/test_multithreading.py:57`
    - Memory allocator/thread-local behavior in `cupy/tests/cupy_tests/cuda_tests/test_memory.py:833`
    - Random state thread-safety tests in `cupy/tests/cupy_tests/random_tests/test_generator.py:1373`
- [x] Check dependency readiness:
  - [x] Binary packages we depend on like Numba, numba-cuda, cuda-bindings, cuda-core, NumPy, etc. must support free threading
  - [x] If issue with numba-cuda bring up with Leo
- [x] Make sure that performance is not hurt due to thread safety
- [ ] Replace `numba-cuda` with `numba-cuda-mlir` #9408

PR #7072, opened 2026-01-04
cuda.compute: Allow multiple uses of the same function in single compilation

### LINKED ISSUE #6768: [BUG]: `cuda.compute.three_way_partition` fails to build when the same predicate is used for both first and second part
When the same predicate is used for both selectors in `three_way_partition`, it fails to build with:

```python
test: link error: error: Linking globals named 'wrapped_always_true': symbol multiply defined!

ERROR 9 in nvvmCompileProgram callback
```

shwina:
## Description

Closes #6768 

When linking together LTOIRs, care must be taken to ensure that we don't have multiple symbols with the same name. If we _just_ used the name of a function as its corresponding symbol name, this is very easy to run into. For example, we could have an algorithm with two input arguments - both iterators - and the corresponding iterator advance methods would both be named `advance`.

To avoid this, we were previously using the function `id` to compute a unique suffix for the corresponding symbol name. As seen in #6768, this is still problematic when the exact same function is used more than once in the same compilation (for example, as both selectors of a `three_way_partition`).

This PR makes it so that we use a simple counter for computing the suffix instead, guaranteeing a unique suffix for every symbol.

To ensure this doesn't inadvertently cause any performance regressions, I ran our existing Python benchmarks both before  (`be`) and after (`af`) this PR. The results show no difference.

Comparison results for posterity:
[results.txt](https://github.com/user-attachments/files/24421620/results.txt)

## Checklist

### DIFF
--- a/python/cuda_cccl/cuda/compute/_odr_helpers.py
+++ b/python/cuda_cccl/cuda/compute/_odr_helpers.py
@@ -27,6 +27,7 @@ from __future__ import annotations
 import enum
 import itertools
 import textwrap
+import threading
 from typing import TYPE_CHECKING
 
 from numba import types
@@ -40,6 +41,7 @@ if TYPE_CHECKING:
 # Global counter to generate unique symbol names even when the same function
 # is used multiple times (e.g., as both selectors in `three_way_partition`).
 _wrapper_name_counter = itertools.count()
+_wrapper_name_lock = threading.Lock()
 
 __all__ = [
     "create_op_void_ptr_wrapper",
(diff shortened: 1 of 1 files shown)


python/cuda_cccl/cuda/compute/_odr_helpers.py
```diff
@@ -36,6 +37,10 @@
 if TYPE_CHECKING:
     from numba.core.typing import Signature
 
+# Global counter to generate unique symbol names even when the same function
+# is used multiple times (e.g., as both selectors in `three_way_partition`).
+_wrapper_name_counter = itertools.count()
```
oleksandr-pavlyk, line 43: Perhaps this global  counter needs a lock to avoid creating race condition in free-threaded interpreter

shwina: Fixed, although I will say that `cuda.compute` as a whole is probably not thread-safe today. All the caching mechanisms etc., haven't taken thread safety into account thus far (#6422).

PR #7074, opened 2026-01-05
Guard against user defined `major` and `minor` macros

miscco:
We need to avoid using those identifiers as they are frequently used by users

Addresses nvbug5781562

### DIFF
--- a/libcudacxx/include/cuda/std/__cccl/epilogue.h
+++ b/libcudacxx/include/cuda/std/__cccl/epilogue.h
@@ -352,4 +352,20 @@ _CCCL_DIAG_POP
 #  undef _CCCL_POP_MACRO_clang
 #endif
 
+// major and minor are defined in sys/sysmacros.h in libc
+#if defined(major)
+#  error \
+    "cccl internal error: macro `major` was redefined between <cuda/std/__cccl/prologue.h> and <cuda/std/__cccl/epilogue.h>"
+#elif defined(_CCCL_POP_MACRO_major)
+#  pragma pop_macro("major")
+#  undef _CCCL_POP_MACRO_major
+#endif
+
+#if defined(minor)
+#  error \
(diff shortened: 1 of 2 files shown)


davebayer:
Uuuh, I don't know whether this is a good idea, we could end up adding more and more of these macros. And I don't think it will fix the problem, because then if the user wants to use `cc.major()` somewhere, it will blow up again.

miscco:
The other option would be to change the member function to `major_version` or `version_major`

@pciolkosz what do you prefer?

davebayer:
> The other option would be to change the member function to `major_version` or `version_major`

This API is already released, we cannot change that. I really think that this is user's problem, because if you have a look for example at `<nvrtc.h>`, they use `major` and `minor` as names for function parameters and the compilation would fail if the user defines the macros.

It's not that I don't want to help the user, but this solution is not sustainable. We are already trying our best to prevent the user to modify our internals, but I don't think we should do that for everything.

fbusato (approved):

jrhemstad:
@davebayer @miscco the "user" in this situation is just one of the CUDA samples, specifically:
https://github.com/NVIDIA/cuda-samples/tree/master/Samples/3_CUDA_Features/cdpAdvancedQuicksort
https://github.com/NVIDIA/cuda-samples/tree/master/Samples/3_CUDA_Features/cdpQuadtree

That said, I don't see any macro definition of `major` or `minor` here that would cause the issue @miscco described.

miscco:
Yeah I am really wondering how this could happen, but that the error in question is:

```
Command output: /usr/local/cuda-safe-13.2/targets/aarch64-qnx/include/cccl/cuda/__device/compute_capability.h(87): error: expected an identifier
                 __attribute__((__visibility__("hidden"))) constexpr int ((int)((() >> 10) & 0x3f)) const noexcept
```
This clearly hints at `major()` being replaced by `((int)((() >> 10) & 0x3f))`

davebayer:
I've done some research and those macros come from `sys/sysmacros.h`. Could you please update the description for these macros?

miscco:
closing in favor of #7099

PR #7077, opened 2026-01-05
Avoid invalid compiler warning with VS2026

miscco:
internally we are seeing invalid compiler warnings about a missing `template` keyword which is definitely there.

To unblock internal CI disable that warning until the actual compiler issue is found

Addresses nvbug5765092


davebayer (approved):

davebayer:
Maybe it has to be typed in capitals, so msvc can see it, it's getting old now :)

PR #7078, opened 2026-01-05
[DOCS] Clarifies `DeviceTopK` docs that inputs and output ranges may not overlap

elstehle:
Since `DeviceTopK` is a multi-pass algorithm, user-provided inputs and output ranges may not overlap.

### DIFF
--- a/cub/cub/device/device_topk.cuh
+++ b/cub/cub/device/device_topk.cuh
@@ -143,7 +143,8 @@ struct DeviceTopK
   //! Overview
   //! +++++++++++++++++++++++++++++++++++++++++++++
   //!
-  //! Finds the largest K keys and their corresponding values from an unordered input sequence of key-value pairs.
+  //! - Finds the largest K keys and their corresponding values from an unordered input sequence of key-value pairs.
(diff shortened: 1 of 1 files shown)


cub/cub/device/device_topk.cuh
```diff
@@ -143,7 +143,8 @@ struct DeviceTopK
   //! Overview
   //! +++++++++++++++++++++++++++++++++++++++++++++
   //!
-  //! Finds the largest K keys and their corresponding values from an unordered input sequence of key-value pairs.
+  //! - Finds the largest K keys and their corresponding values from an unordered input sequence of key-value pairs.
+  //! - Note that the behavior is undefined if the input and output ranges overlap in any way.
```
bernhardmgruber, line 147: Suggestion: let's not call it a "note":
```suggestion
  //! - The behavior is undefined if the input and output ranges overlap in any way.
```
or:
```suggestion
  //! - The input and output ranges must not overlap in any way.
```
Applies to all comments.

bernhardmgruber (approved):

PR #7088, opened 2026-01-06
Disable reference_wrapper test for VS2026

miscco:
VS2026 seems to have a bug where reference wrapper picks up a `result_type` it should not.

Disable that test until we find a proper workaround

Fixes nvbug5765035


libcudacxx/test/libcudacxx/std/utilities/function.objects/refwrap/weak_result.pass.cpp
```diff
@@ -96,7 +96,9 @@ int main(int, char**)
     "");
   static_assert((cuda::std::is_same<cuda::std::reference_wrapper<C()>::result_type, C>::value), "");
   static_assert(has_result_type<cuda::std::reference_wrapper<functor3>>::value, "");
+#if !TEST_COMPILER(MSVC, >=, 19, 50) // MSVC 2026 fails this
```
bernhardmgruber: Remark:
```suggestion
#if !TEST_COMPILER(MSVC, >=, 19, 50) // Visual Studio 2026 fails this
```
no action needed

PR #7090, opened 2026-01-06
Add PyTorch Conference 2025 `cuda.compute` poster

NaderAlAwar:


rwgk:
Nice Poster, it reads very well!

There was only this one place where I felt a little confused:

___
Explicit Fusion: Since iterators
are "lazy", all computations are
fused into a single kernel call
here.
___


1. From the perspective of a user, the fusion doesn't appear to be "Explicit". WDYT about "Automatic Fusion" instead?

2. I had to look around a bit to make up my mind what "here" refers to. I assume it's the `# Perform the scan` box right underneath. A small arrow pointing there would be nice.

NaderAlAwar:
Posted here https://naderalawar.github.io/files/High-Performance%20CUDA%20Ops%20in%20Python%2C%20JIT-Compiling%20CUB%20with%20cuda.compute.pdf

PR #7093, opened 2026-01-06
Implement new tuning API arch dispatching

bernhardmgruber:
This implements the new runtime arch_id -> compile-time tuning policy dispatching logic of the new tuning API. It is analogous to `cub::ChainedPolicy`.


cub/test/catch2_test_arch_dispatch.cu
```diff
@@ -0,0 +1,150 @@
...
+{
+  arch_id id;
+
+  template <typename PolicyGetter>
+  CUB_RUNTIME_FUNCTION auto operator()(PolicyGetter policy_getter) const ->
+#ifdef CUDA_SM_LIST
+    typename check<PolicyGetter{}().value, CUDA_SM_LIST>::type
+#else // CUDA_SM_LIST
+    cudaError_t
+#endif // CUDA_SM_LIST
```
miscco, line 60: Honestly I would rather have the static_assert in the function body and always return cudaError_t

bernhardmgruber: Done

Issue #724, opened 2019-12-18
Examples can (and should) be exemplary C++


xtofl:
Obviously, the primary goal of the examples is to show how one uses the library.

However, they are also adding to the overall C++ knowledge: people using `thrust` are very likely to also learn the standard library algorithms, and take over the style presented (if not even copy-paste-adapt the examples).

So there would be value in making the example code exemplary, 'modern' C++.

Some instances of example code that could  be modernized:

```
    thrust::device_vector<int> stencil(8);
    stencil[0] = 0;
    stencil[1] = 1;
...
```
could make use of `std::initializer_list<int>` to direct initialize the vector.

```
std::cout << "found " << (indices_end - indices.begin()) ...
```
Should really use `std::distance(begin(indices), indices_end)`

```
// reduce a pair of bounding boxes (a,b) to a bounding box containing a and b
struct bbox_reduction : public thrust::binary_function<bbox,bbox,bbox>
```
Is a poor choice of name: it tells you what function it's going to be used in (reduce), not what it does (bounding box of bounding boxes); it may be replaced by a lambda expression.

...

I think it would be a good idea to set up an 'example-wide' code review and try to exemplify that code.

(cf. also the very limited pull request NVIDIA/thrust#753)

charan-003:
Hi, I'd like to work on this issue. Could you please assign it to me? I hope no one else is currently working on it. Looking at the thrust examples, I assume the changes I need to make include:
- Using `std::initializer_list` to directly initialize vectors (e.g., `thrust::device_vector<int> stencil = {0, 1, ...};`).
- Replacing manual iterator subtraction with `std::distance` (e.g., `std::distance(indices.begin(), indices_end)`).
- Replacing named functors like `bbox_reduction` with lambda expressions for clarity.
and further continue with other files.
Please let me know if I’m on the right track or if there’s anything else to consider!

charan-003:
@bernhardmgruber @xtofl  ?

bernhardmgruber:
@charan-003 I removed the assignment of this issue to @vrajvaghela89. We are happy to receive a PR with improvements. There is no need to be assigned to this issue.

Please create small PRs with limited scope, like one PR will just improve all the use of thrust vectors etc. Another PR will move to more `cuda::std` algorithms etc. This way, we avoid long review times. We are looking forward to your suggestions!

charan-003:
@bernhardmgruber thank you for the reply!!
[...]

PR #7094, opened 2026-01-06
Fix/modernize thrust examples

Flawxd:
Following #724

Modernizes the thrust/examples/word_count.cu example by replacing the functor struct is_word_start with an extended lambda
to align with modern C++ practices and to make it easier to read

Changes:

Replaced struct is_word_start with a __host__ __device__ lambda.

Replaced C-style array pointer arithmetic with std::begin() and std::end() for safer iterator access.

The code was verified using a T4 GPU and ran perfectly

### DIFF
--- a/thrust/examples/word_count.cmake
+++ b/thrust/examples/word_count.cmake
@@ -3,12 +3,19 @@ target_compile_options(
   PRIVATE $<$<COMPILE_LANG_AND_ID:CUDA,NVIDIA>: --extended-lambda>
 )
 
-# When clang >= 13 is used as host compiler, we get the following warning:
-#   nvcc_internal_extended_lambda_implementation:312:22: error: definition of implicit copy constructor for '__nv_hdl_wrapper_t<false, true, false, __nv_dl_tag<void (*)(), &TestAddressStabilityLambda, 2>, int (const int &)>' is deprecated because it has a user-declared copy assignment operator [-Werror,-Wdeprecated-copy]
-#   312 | __nv_hdl_wrapper_t & operator=(const __nv_hdl_wrapper_t &in) = delete;
-#       |                      ^
-# Let's suppress it until NVBug 4980157 is resolved.
-target_compile_options(
-  ${example_target}
-  PRIVATE $<$<COMPILE_LANG_AND_ID:CUDA,NVIDIA>: -Wno-deprecated-copy>
+# This check is actually not correct, because we must check the host compiler, not the CXX compiler.
+# We rely on these usually being the same ;)
+if (
+  "Clang" STREQUAL "${CMAKE_CXX_COMPILER_ID}"
+  AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 13
 )
+  # When clang >= 13 is used as host compiler, we get the following warning:
+  #   nvcc_internal_extended_lambda_implementation:312:22: error: definition of implicit copy constructor for '__nv_hdl_wrapper_t<false, true, false, __nv_dl_tag<void (*)(), &TestAddressStabilityLambda, 2>, int (const int &)>' is deprecated because it has a user-declared copy assignment operator [-Werror,-Wdeprecated-copy]
+  #   312 | __nv_hdl_wrapper_t & operator=(const __nv_hdl_wrapper_t &in) = delete;
+  #       |                      ^
+  # Let's suppress it until NVBug 4980157 is resolved.
+  target_compile_options(
+    ${example_target}
+    PRIVATE $<$<COMPILE_LANG_AND_ID:CUDA,NVIDIA>: -Wno-deprecated-copy>
+  )
+endif()


thrust/examples/word_count.cu
```diff
@@ -35,14 +27,21 @@ int word_count(const thrust::device_vector<char>& input)
     return 0;
   }
 
+  // determines whether the right character begins a new word
+  // Define lambda locally for safety and clarity
```
bernhardmgruber: Important: Please remove superfluous comments that explain C++ syntax or explain design decisions that don't concern CUDA. They distract IMO.
```suggestion
```

thrust/examples/word_count.cu
```diff
@@ -35,14 +27,21 @@ int word_count(const thrust::device_vector<char>& input)
...
+
   // compute the number characters that start a new word
   int wc = thrust::inner_product(
     input.begin(),
     input.end() - 1, // sequence of left characters
     input.begin() + 1, // sequence of right characters
     0, // initialize sum to 0
-    cuda::std::plus<int>(), // sum values together
-    is_word_start()); // how to compare the left and right characters
+    cuda::std::plus<int>{}, // sum values together (use {} for uniform init)
```
bernhardmgruber:
```suggestion
    cuda::std::plus<int>{}, // sum values together
```

thrust/examples/word_count.cu
```diff
@@ -35,14 +27,21 @@ int word_count(const thrust::device_vector<char>& input)
...
+    is_word_start // Pass the lambda object directly (NO parentheses)
```
bernhardmgruber: Important: please retain the previous comment

thrust/examples/word_count.cu
```diff
@@ -65,16 +64,17 @@ int main()
...
     "  Then the bird said, `Nevermore.'\n";
 
-  std::cout << "Text sample:" << std::endl;
-  std::cout << raw_input << std::endl;
+  std::cout << "Text sample:\n";
+  std::cout << raw_input << "\n";
 
   // transfer to device
-  thrust::device_vector<char> input(raw_input, raw_input + sizeof(raw_input));
+  // Use std::begin/end for safer C++ iterator access
```
bernhardmgruber:
```suggestion
```

thrust/examples/word_count.cu
```diff
@@ -5,6 +5,7 @@
 #include <thrust/reduce.h>
 
 #include <iostream>
+#include <iterator> // Required for std::begin/std::end
```
miscco: This should be
```suggestion
#include <cuda/std/iterator> // Required for std::begin/std::end
```

thrust/examples/word_count.cu
```diff
@@ -65,16 +63,16 @@ int main()
...
     "  Then the bird said, `Nevermore.'\n";
 
-  std::cout << "Text sample:" << std::endl;
-  std::cout << raw_input << std::endl;
+  std::cout << "Text sample:\n";
+  std::cout << raw_input << "\n";
 
   // transfer to device
-  thrust::device_vector<char> input(raw_input, raw_input + sizeof(raw_input));
+  thrust::device_vector<char> input(std::begin(raw_input), std::end(raw_input));
```
miscco:
```suggestion
  thrust::device_vector<char> input(cuda::std::begin(raw_input), cuda::std::end(raw_input));
```

thrust/examples/word_count.cu
bernhardmgruber: I am not so strict about this for examples. There is no benefit for users to prefer `cuda::std::begin` in host code. At worst, it may confuse them.

We should use it ourselves in tests and implementation to increase test coverage.

bernhardmgruber: Talked to @miscco offline and he wants to have it. So let's use it.

bernhardmgruber:
pre-commit.ci autofix

bernhardmgruber:
From the CI:
```
 /home/coder/cccl/thrust/examples/word_count.cu(31): error: __host__ or __device__ annotation on lambda requires --extended-lambda nvcc flag
      auto is_word_start = [] __attribute__((host)) __attribute__((device))(const char& left, const char& right) {
                                                                        ^
```
Please add a new file `thrust/examples/word_count.cmake` with content:
```cmake
target_compile_options(
  ${example_target}
  PRIVATE $<$<COMPILE_LANG_AND_ID:CUDA,NVIDIA>: --extended-lambda>
)
```

bernhardmgruber:
> The code was verified using a T4 GPU and ran perfectly

@Flawxd I wonder how you verfied this, because the example in the PR cannot be compiled without `--extended-lambda`. Please make sure you test compiling the example with our build system!

Flawxd:
> @Flawxd I wonder how you verfied this, because the example in the PR cannot be compiled without `--extended-lambda`. Please make sure you test compiling the example with our build system!

Yes, I did test it using --extend-lambda, it was an oversight from me to not include it in a cmake file, I'll push the new file

bernhardmgruber:
pre-commit.ci autofix

bernhardmgruber:
Please also add this to the cmake file:
```
  # When clang >= 13 is used as host compiler, we get the following warning:
  #   nvcc_internal_extended_lambda_implementation:312:22: error: definition of implicit copy constructor for '__nv_hdl_wrapper_t<false, true, false, __nv_dl_tag<void (*)(), &TestAddressStabilityLambda, 2>, int (const int &)>' is deprecated because it has a user-declared copy assignment operator [-Werror,-Wdeprecated-copy]
  #   312 | __nv_hdl_wrapper_t & operator=(const __nv_hdl_wrapper_t &in) = delete;
  #   |                      ^
  # Let's suppress it until NVBug 4980157 is resolved.
  target_compile_options(
    ${test_target}
    PRIVATE $<$<COMPILE_LANG_AND_ID:CUDA,NVIDIA>:-Wno-deprecated-copy>
  )
```
we already have this fix in other places.

Flawxd:
Hello, isnt the correct variable example_target instead of test_target ?
Also, from what I can see the warning only appears for clang >=13 so shouldnt we test and apply only after seeing that the version is greater than 12 ?  so something like
```
target_compile_options(
  ${example_target}
  PRIVATE $<$<COMPILE_LANG_AND_ID:CUDA,NVIDIA>: --extended-lambda>
)

if (
  "Clang" STREQUAL "${CMAKE_CXX_COMPILER_ID}"
  AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 13
)
  # When clang >= 13 is used as host compiler, we get the following warning:
  #   nvcc_internal_extended_lambda_implementation:312:22: error: definition of implicit copy constructor for '__nv_hdl_wrapper_t<false, true, false, __nv_dl_tag<void (*)(), &TestAddressStabilityLambda, 2>, int (const int &)>' is deprecated because it has a user-declared copy assignment operator [-Werror,-Wdeprecated-copy]
  #   312 | __nv_hdl_wrapper_t & operator=(const __nv_hdl_wrapper_t &in) = delete;
  #       |                      ^
  # Let's suppress it until NVBug 4980157 is resolved.
  target_compile_options(
    ${example_target}
    PRIVATE $<$<COMPILE_LANG_AND_ID:CUDA,NVIDIA>: -Wno-deprecated-copy>
  )
endif()
```

bernhardmgruber:
> Hello, isnt the correct variable example_target instead of test_target

Correct. I copied this from a test.

> Also, from what I can see the warning only appears for clang >=13 so shouldnt we test and apply only after seeing that the version is greater than 12 ?

Yes, but this may be trickier than it sounds. `CMAKE_CXX_COMPILER_ID` is CMake's C++ compiler and this could be a different compiler than the `CMAKE_CUDA_HOST_COMPILER`. We guard against this somewhere in CMake, but if you take the example CMake scripts out of context, they could be different. Recent versions of CMake allow better inspection of the host compiler, but then we need to handle different versions of CMake. It's just a lot easier to just disable the warning for nvcc.

Flawxd:
Alright then, I'll do what you suggested with the correct variable

thrust/examples/word_count.cmake
```diff
@@ -0,0 +1,14 @@
...
+
+# When clang >= 13 is used as host compiler, we get the following warning:
+#   nvcc_internal_extended_lambda_implementation:312:22: error: definition of implicit copy constructor for '__nv_hdl_wrapper_t<false, true, false, __nv_dl_tag<void (*)(), &TestAddressStabilityLambda, 2>, int (const int &)>' is deprecated because it has a user-declared copy assignment operator [-Werror,-Wdeprecated-copy]
+#   312 | __nv_hdl_wrapper_t & operator=(const __nv_hdl_wrapper_t &in) = delete;
+#       |                      ^
+# Let's suppress it until NVBug 4980157 is resolved.
+target_compile_options(
+  ${example_target}
+  PRIVATE $<$<COMPILE_LANG_AND_ID:CUDA,NVIDIA>: -Wno-deprecated-copy>
+)
```
bernhardmgruber:
```suggestion
# This check is actually not correct, because we must check the host compiler, not the CXX compiler.
# We rely on these usually being the same ;)
if (
  "Clang" STREQUAL "${CMAKE_CXX_COMPILER_ID}"
  AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 13
)
  # When clang >= 13 is used as host compiler, we get the following warning:
  #   nvcc_internal_extended_lambda_implementation:312:22: error: definition of implicit copy constructor for '__nv_hdl_wrapper_t<false, true, false, __nv_dl_tag<void (*)(), &TestAddressStabilityLambda, 2>, int (const int &)>' is deprecated because it has a user-declared copy assignment operator [-Werror,-Wdeprecated-copy]
  #   312 | __nv_hdl_wrapper_t & operator=(const __nv_hdl_wrapper_t &in) = delete;
  #       |                      ^
  # Let's suppress it until NVBug 4980157 is resolved.
  target_compile_options(
    ${example_target}
    PRIVATE $<$<COMPILE_LANG_AND_ID:CUDA,NVIDIA>: -Wno-deprecated-copy>
  )
endif()
```

bernhardmgruber:
> > Also, from what I can see the warning only appears for clang >=13 so shouldnt we test and apply only after seeing that the version is greater than 12 ?
> 
> Yes, but this may be trickier than it sounds. `CMAKE_CXX_COMPILER_ID` is CMake's C++ compiler and this could be a different compiler than the `CMAKE_CUDA_HOST_COMPILER`. We guard against this somewhere in CMake, but if you take the example CMake scripts out of context, they could be different. Recent versions of CMake allow better inspection of the host compiler, but then we need to handle different versions of CMake. It's just a lot easier to just disable the warning for nvcc.

It seems I confused something, since we still have not exclude the flag for MSVC. Looking into other files, I see that we do limit the flag to when the CXX compiler is clang >=13. I pushed the change. Sorry for the noise!

Flawxd:
No problem, thank you for the heads-up !

miscco:
Thanks a lot for improving the examples 🎉

Flawxd:
NP, Thanks for your guidance !

PR #7096, opened 2026-01-07
[libcu++] Check if managed pools are accessible in is_pointer_accessible test

pciolkosz:
is_pointer_accessible test uses managed memory pools without checking if they are supported, this PR adds the required device attribute checks.

It also does a fly-by fix of `__deviceGetAttribute` accidentally returning `CUdevice` instead of `int`

### DIFF
--- a/docs/cub/api_docs/device_wide.rst
+++ b/docs/cub/api_docs/device_wide.rst
@@ -90,6 +90,7 @@ CUB device-level single-problem parallel algorithms:
 * :cpp:struct:`cub::DeviceRunLengthEncode` demarcating "runs" of same-valued items within a sequence residing within device-accessible memory
 * :cpp:struct:`cub::DeviceScan` computes a prefix scan across a sequence of data items residing within device-accessible memory
 * :cpp:struct:`cub::DeviceSelect` compacts data residing within device-accessible memory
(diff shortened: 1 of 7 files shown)


libcudacxx/test/libcudacxx/cuda/memory/is_pointer_accessible.pass.cpp
```diff
@@ -129,15 +129,22 @@ void test_memory_pool_impl(
...
-  test_memory_pool_impl(cudaMemAllocationTypeManaged, cudaMemLocationTypeDevice, false, true, true);
+    // TODO(fbusato): check if this can be improved in future releases
+    if (cuda::__driver::__deviceGetAttribute(::CU_DEVICE_ATTRIBUTE_CONCURRENT_MANAGED_ACCESS, 0))
+    {
+      // TODO(fbusato): check if this can be improved in future releases
+      test_memory_pool_impl(cudaMemAllocationTypeManaged, cudaMemLocationTypeHost, true, false, true);
+      test_memory_pool_impl(cudaMemAllocationTypeManaged, cudaMemLocationTypeDevice, false, true, true);
+    }
+  }
 #endif // _CCCL_CTK_AT_LEAST(13, 0)
```
davebayer:
```suggestion
#endif // _CCCL_CTK_AT_LEAST(13, 0)
  }
```

PR #7099, opened 2026-01-07
Generator for prologue/epilogue

davebayer:
I am scared we will introduce a typo one day, so let's use a generator for our prologue/epilogue.

### DIFF
--- a/ci/matrix.yaml
+++ b/ci/matrix.yaml
@@ -80,7 +80,7 @@ workflows:
     - {jobs: ['test'], project: 'nvbench_helper', ctk: '13.0', cxx: ['gcc',   'clang'],   gpu: 'rtx2080'}
     - {jobs: ['test'], project: 'nvbench_helper', ctk: '13.X', cxx: ['gcc',   'clang'],   gpu: 'rtx2080'}
     # NVHPC build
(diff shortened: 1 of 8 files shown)


miscco (approved):
Please mark nvbug5781562 in the commit message

libcudacxx/codegen/generate_prologue_epilogue.py
bernhardmgruber, line 1: Suggestion: we already have code generating facilities in `libcudacxx/codegen` (for atomic and new PTX instructions). Please consider adding the script there, or at least moving the `add_ptx_instruction.py` script to your new `utils` directory as well, if we determine that `codegen` belongs to atomics alone.

davebayer: That makes sense. I placed it to `utils`, because that's what libc++ does with their scripts. Maybe we should put the atomics generator to `codegen/atomics` to better separate each generator. Maybe we can do that in another PR?

bernhardmgruber: Fine for me!

davebayer: I've decided to fix it right away :) Thanks for the suggestion

PR #7108, opened 2026-01-07
Move DLPack include to separate file

davebayer:
We will be using DLPack in multiple files, so I think it should be better to separate it in a separate include. I've also added several macros for detecting and version control.


libcudacxx/include/cuda/std/__internal/features.h
```diff
@@ -101,4 +101,12 @@
 #  define _CCCL_DIAGNOSE_ERROR(_COND, _MSG)
 #endif
 
+// Third party libraries
+
+#if _CCCL_HAS_INCLUDE(<dlpack/dlpack.h>) && !_CCCL_COMPILER(NVRTC) && !defined(CCCL_DISABLE_DLPACK)
```
fbusato, line 106: I'm a puzzled about `NVRTC`. It makes sense to disable DLPack on host conversion API, but I would not exclude `DLPack` entirely. Suppose that a user what to use `DLTensor` on a function directly with NVRTC enabled.

libcudacxx/include/cuda/__internal/dlpack.h
```diff
@@ -0,0 +1,38 @@
+//===----------------------------------------------------------------------===//
```
fbusato, line 1: I would not use `cuda/std` path

libcudacxx/include/cuda/std/__internal/features.h
davebayer: Discussed offline: `<dlpack/dlpack.h>` includes `<stdlib.h>` and `<stddef.h>` which are not available with nvrtc

libcudacxx/include/cuda/__internal/dlpack.h
davebayer: Moved to `<cuda/__internal/dlpack.h>`

PR #7109, opened 2026-01-07
Disable _CCCL_NO_UNIQUE_ADDRESS for CTK 12.9

### LINKED ISSUE #7111: [BUG]: `cuda::std::tuple` or empty classes with `[[no_unique_address]]` attribute segfaults on CTK 12.9
### Is this a duplicate?

### Type of Bug

Silent Failure

### Component

libcu++

### Describe the bug

`cuda::std::tuple` or empty classes with `[[no_unique_address]]` attribute segfaults on CTK 12.9 with both GCC and Clang host compiler's with both C++ 17 and 20 dialects. 

### How to Reproduce

https://godbolt.org/z/GKrE43xsK

```#include <cuda/std/__utility/pod_tuple.h>

struct max_segment_size_42
{};

__host__ __device__ void test_tuple_empty()
{
  [[maybe_unused]] ::cuda::std::__tuple<max_segment_size_42> t{};
}

int main(int, char**)
{
  test_tuple_empty();
  return 0;
}
```

### Expected behavior

Should compile sucessfully

### Reproduction link

_No response_

### Operating System

_No response_

### nvidia-smi output

_No response_

### NVCC version

_No response_

srinivasyadav18:
## Description

closes #7111 

`::cuda::std::__tuple` fails to compile with empty classes when `_CCCL_NO_UNIQUE_ADDRESS` is defined on CTK 12.9 with both GCC and Clang with both C++17,20. 

We might have to disable `_CCCL_NO_UNIQUE_ADDRESS` macro for CTK 12.9, inorder to make stateless env work properly.

### DIFF
--- a/libcudacxx/include/cuda/std/__cccl/attributes.h
+++ b/libcudacxx/include/cuda/std/__cccl/attributes.h
@@ -157,7 +157,7 @@
 
 // _CCCL_NO_UNIQUE_ADDRESS
 
-#if _CCCL_COMPILER(MSVC) || _CCCL_HAS_CPP_ATTRIBUTE(no_unique_address) < 201803L || _CCCL_CTK_EQUAL(12, 9)
+#if _CCCL_COMPILER(MSVC) || _CCCL_HAS_CPP_ATTRIBUTE(no_unique_address) < 201803L || _CCCL_CUDA_COMPILER(NVCC, <, 13)
 // MSVC implementation has lead to multiple issues with silent runtime corruption when passing data into kernels
 // CTK 12.9 seg faults when using [[no_unique_address]] with instantiations of empty base classes
 #  define _CCCL_HAS_ATTRIBUTE_NO_UNIQUE_ADDRESS() 0
--- a/libcudacxx/include/cuda/std/__cccl/cuda_toolkit.h
+++ b/libcudacxx/include/cuda/std/__cccl/cuda_toolkit.h
@@ -52,6 +52,5 @@
 #define _CCCL_CTK_MAKE_VERSION(_MAJOR, _MINOR) ((_MAJOR) * 1000 + (_MINOR) * 10)
 #define _CCCL_CTK_BELOW(...)                   _CCCL_VERSION_COMPARE(_CCCL_CTK_, _CCCL_CTK, <, __VA_ARGS__)
 #define _CCCL_CTK_AT_LEAST(...)                _CCCL_VERSION_COMPARE(_CCCL_CTK_, _CCCL_CTK, >=, __VA_ARGS__)
-#define _CCCL_CTK_EQUAL(...)                   _CCCL_VERSION_COMPARE(_CCCL_CTK_, _CCCL_CTK, ==, __VA_ARGS__)
 
 #endif // __CCCL_CUDA_TOOLKIT_H


libcudacxx/include/cuda/std/__cccl/attributes.h
```diff
@@ -156,8 +157,9 @@
 
 // _CCCL_NO_UNIQUE_ADDRESS
 
-#if _CCCL_COMPILER(MSVC) || _CCCL_HAS_CPP_ATTRIBUTE(no_unique_address) < 201803L
+#if _CCCL_COMPILER(MSVC) || _CCCL_HAS_CPP_ATTRIBUTE(no_unique_address) < 201803L || _CCCL_CTK_EQUAL(12, 9)
 // MSVC implementation has lead to multiple issues with silent runtime corruption when passing data into kernels
+// CTK 12.9 seg faults when using [[no_unique_address]] with instantiations of empty base classes
```
miscco: We should just disable it for anything less than 13.0

libcudacxx/include/cuda/std/__cccl/cuda_toolkit.h
```diff
@@ -52,5 +52,6 @@
 #define _CCCL_CTK_MAKE_VERSION(_MAJOR, _MINOR) ((_MAJOR) * 1000 + (_MINOR) * 10)
 #define _CCCL_CTK_BELOW(...)                   _CCCL_VERSION_COMPARE(_CCCL_CTK_, _CCCL_CTK, <, __VA_ARGS__)
 #define _CCCL_CTK_AT_LEAST(...)                _CCCL_VERSION_COMPARE(_CCCL_CTK_, _CCCL_CTK, >=, __VA_ARGS__)
+#define _CCCL_CTK_EQUAL(...)                   _CCCL_VERSION_COMPARE(_CCCL_CTK_, _CCCL_CTK, ==, __VA_ARGS__)
```
bernhardmgruber: I am not sure this works, since `_CCCL_CTK_EQUAL(12, 9)` would compare against 12.9 exactly and probably not match patch releases. @miscco probably knows what to do instead.

miscco: We should just disable it completely for _CCCL_CTK_BELOW(13, 0)

davebayer: Even with clang-cuda and nvhpc? Then it should be `_CCCL_CUDACC_BELOW(13,0)`. If only `nvcc`, then `_CCCL_CUDA_COMPILER(NVCC, <, 13)`

libcudacxx/include/cuda/std/__cccl/attributes.h
davebayer: Isn't this an ABI break?

miscco: Disabling at all is an ABI break 🙈

davebayer: :this-is-fine-2:

libcudacxx/include/cuda/std/__cccl/cuda_toolkit.h
srinivasyadav18: @davebayer @miscco only NVCC 12.9 is failing, clang-cuda and nvhpc work fine. So, now should I disable only for nvcc through `_CCCL_CUDA_COMPILER(NVCC, <, 13)` or everything i.e.  `_CCCL_CTK_BELOW(13, 0)`

davebayer: This whole thing is just scary, there is not good solution for this problem. If we disable `[[no_unique_address]]` with `nvcc` and the user has some host only code compiled without `nvcc` (nor has include path to CTK), then we are basically breaking the ABI compatiblity between C++ and CUDA/C++ code.

I don't know what we can do about this.

davebayer: The only solution I see is either use it everywhere or nowhere.

miscco: The only issue we would have is if the user cross links libraries compiled with different cuda compiles, which seems odd.

I believe we should disable for `_CCCL_CUDA_COMPILER(NVCC, <, 13)` and be done with it

davebayer (changes requested):
Block this PR until we decide what to do about the `[[no_unique_address]]` attribute

PR #7113, opened 2026-01-08
[libcu++] Make resource an environment when queried for a resource

pciolkosz:
We introduced environment based APIs in CUB, but right now you need to use:
```
auto memory_prop  = cuda::std::execution::prop{cuda::mr::get_memory_resource, mr};
auto env = cuda::std::execution::env{ memory_prop, other_things};
```
It would be better to just do
```
auto env = cuda::std::execution::env{mr, other_things};
```
like we do with a stream. This PR adds an overload accepting a resource to `cuda::mr::get_memory_resource`, like we do with `get_stream`. It would be a burden to require every resource to declare `cuda::mr::get_memory_resource` query, so instead this PR changes `cuda::std::execution::__query_or` to also try to call the supplied query with the env as an argument, to utilize the newly added overload in `cuda::mr::get_memory_resource`

### DIFF
--- a/libcudacxx/include/cuda/std/__execution/env.h
+++ b/libcudacxx/include/cuda/std/__execution/env.h
@@ -422,7 +422,7 @@ struct __query_or_t
   _CCCL_EXEC_CHECK_DISABLE
   _CCCL_TEMPLATE(class _Env, class _Query, class _Default, class... _Args)
   _CCCL_REQUIRES(__queryable_with<_Env, _Query, _Args...>)
-  [[nodiscard]] _CCCL_API constexpr auto operator()(const _Env& __env, _Query&&, _Default&&, _Args&&... __args) const
+  [[nodiscard]] _CCCL_API constexpr auto operator()(const _Env& __env, _Query, _Default&&, _Args&&... __args) const
     noexcept(__nothrow_queryable_with<_Env, _Query, _Args...>) -> __query_result_t<_Env, _Query, _Args...>
   {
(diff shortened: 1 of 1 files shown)


libcudacxx/include/cuda/__memory_resource/get_memory_resource.h
```diff
@@ -46,11 +46,11 @@ _CCCL_CONCEPT __has_query_get_memory_resource = _CCCL_REQUIRES_EXPR((_Env))(
   requires(!__has_member_get_resource<_Env>),
   requires(
     resource<
-      ::cuda::std::remove_cvref_t<::cuda::std::execution::__query_result_t<const _Env&, __get_memory_resource_t>>>));
+      ::cuda::std::remove_cvref_t<::cuda::std::execution::__query_result_t<const _Env&, get_memory_resource_t>>>));
 
-//! @brief `__get_memory_resource_t` is a customization point object that queries a type `T` for an associated memory
+//! @brief `get_memory_resource_t` is a customization point object that queries a type `T` for an associated memory
 //! resource
-struct __get_memory_resource_t
```
miscco, line 53: My memory is fuzzy here, why did we keep this internal last release?

pciolkosz: I am not sure, but I think we can just remove the internal name usage

pciolkosz:
Closing in favor of https://github.com/NVIDIA/cccl/pull/7266

Issue #6865, opened 2025-12-03
Improve performance of Device Segmented Reduce on large and very large segment sizes


srinivasyadav18:
The current implementation of device_segmented_reduce only reaches upto 30% in average for segment sizes between 1M and 100M. We could possibliy optimize it by using two phase segmented reduction by using multiple blocks to reduce a single segment. 

Bandwidth utilization of current implementation on H100:
<img width="1200" height="800" alt="Image" src="https://github.com/user-attachments/assets/f7e4b031-ce3f-4197-a012-8990096eb069" />

Some quick optimization show's promising results:

<img width="1200" height="800" alt="Image" src="https://github.com/user-attachments/assets/b8b0ed3b-f6f6-4d1e-9480-201cbc68cda4" />

Leaving with speed ups of

<img width="1200" height="800" alt="Image" src="https://github.com/user-attachments/assets/17ffb214-78f7-4bd7-8ba1-1371c146dad9" />

PR #7114, opened 2026-01-08
Two-phase reduction for fixed size segmented reduction for very large segment sizes

srinivasyadav18:
## Description

closes https://github.com/NVIDIA/cccl/issues/6865

The current implementation uses single block to reduce segments. This creates a performance bottleneck when segments are very large. 
This PR implements two-phase segmented reduction where each segment is processed by multiple blocks, by dividing the single large segment into mulitple fixed size chunks and assigning each chunk to a block. 

## Checklist


cub/cub/device/dispatch/dispatch_fixed_size_segmented_reduce.cuh
```diff
@@ -249,11 +260,125 @@ struct DispatchFixedSizeSegmentedReduce
...
+
+    // if single chunk or if two phase cannot be completed with single invocation, use single-phase reduction
+    if (segment_size < seg_chunk_size || two_phase_num_blocks >= ::cuda::std::numeric_limits<uint32_t>::max())
+    {
+      return InvokePasses<ActivePolicyT>(kernel_source.FixedSizeSegmentedReduceKernel());
+    }
+    // if multiple chunks, use two-phase reduction by reducing large segments in chunks, by assigning each chunk to a
+    // block
+    return InvokeTwoPhase<ActivePolicyT>(
+      kernel_source.FixedSizeSegmentedReduceKernelPartial(), kernel_source.FixedSizeSegmentedReduceKernelFinal());
```
bernhardmgruber: Important: We bent over backward in the past to not increase the amount of emitted kernels of a single CUB algorithm instantiation, because we have several users which are severely impacted by increases in binary size. This change here will emit 3 kernels instead of 1 for each instantiation of `Invoke`. This is an issue given the standard which we upheld in the past.

Now, working around this and channeling new algorithms through a single kernel binary is not only tedious and error prone, it's also really ugly and hard to maintain. I would like to have an opinion from @gevtushenko whether we can weaken this guideline in general, or make an exception here.

cub/cub/device/dispatch/dispatch_fixed_size_segmented_reduce.cuh
```diff
@@ -126,6 +134,9 @@ struct DispatchFixedSizeSegmentedReduce
 
   KernelLauncherFactory launcher_factory;
 
+  // Segment chunk size for two-phase reduction for large segments
+  static constexpr int seg_chunk_size = 1u << 12;
```
bernhardmgruber: Suggestion: this looks like a tuning parameter and should probably go into the policy hub.

cub/cub/device/dispatch/dispatch_fixed_size_segmented_reduce.cuh
```diff
@@ -249,11 +260,125 @@ struct DispatchFixedSizeSegmentedReduce
...
+  CUB_RUNTIME_FUNCTION _CCCL_VISIBILITY_HIDDEN _CCCL_FORCEINLINE cudaError_t
+  InvokeTwoPhase(DeviceFixedSizeSegmentedReduceKernelPartialT fixed_size_segmented_reduce_kernel_partial,
+                 DeviceFixedSizeSegmentedReduceKernelFinalT fixed_size_segmented_reduce_kernel_final)
+  {
+    constexpr auto small_items_per_tile  = ActivePolicyT::SmallReducePolicy::ITEMS_PER_TILE;
+    constexpr auto medium_items_per_tile = ActivePolicyT::MediumReducePolicy::ITEMS_PER_TILE;
+
+    static_assert((small_items_per_tile < medium_items_per_tile),
+                  "small items per tile must be less than medium items per tile");
+    int blocks_per_segment = ::cuda::ceil_div(segment_size, seg_chunk_size);
```
bernhardmgruber:
```suggestion
    const int blocks_per_segment = ::cuda::ceil_div(segment_size, seg_chunk_size);
```

bernhardmgruber: Question: `ceil_div` returns an `OffsetT` which is narrowed to `int` here. Is this intended?

cub/cub/device/dispatch/dispatch_fixed_size_segmented_reduce.cuh
```diff
@@ -249,11 +260,125 @@ struct DispatchFixedSizeSegmentedReduce
...
+            d_block_reductions,
+            segment_size,
+            local_seg_chunk_size,
+            blocks_per_segment,
+            num_current_blocks,
+            reduction_op,
+            init);
+
+    error = CubDebug(cudaPeekAtLastError());
+    if (cudaSuccess != error)
```
bernhardmgruber: Important: please use this pattern for error checking for this PR and the future:
```suggestion
    if (const auto error = CubDebug(cudaPeekAtLastError()))
```
This also eliminates the local `error` variable in this scope.

cub/cub/device/dispatch/dispatch_fixed_size_segmented_reduce.cuh
```diff
@@ -249,11 +260,125 @@ struct DispatchFixedSizeSegmentedReduce
...
+    }
+
+    // Sync the stream if specified to flush runtime errors
+    error = CubDebug(detail::DebugSyncStream(stream));
+    if (cudaSuccess != error)
+    {
+      return error;
+    }
+
+    int final_segment_size       = blocks_per_segment;
```
bernhardmgruber:
```suggestion
    const int final_segment_size       = blocks_per_segment;
```

cub/cub/device/dispatch/kernels/kernel_segmented_reduce.cuh
```diff
@@ -290,6 +290,54 @@ __launch_bounds__(int(ChainedPolicyT::ActivePolicy::ReducePolicy::BLOCK_THREADS)
...
+
+  // Calculate partial segment size
+  const int paritial_segment_size_ =
+    ((full_segment_size % seg_chunk_size != 0) && ((bid % blocks_per_segment) == (blocks_per_segment - 1))
+       ? (full_segment_size % seg_chunk_size)
+       : seg_chunk_size);
+
+  const int nth_segment            = bid / blocks_per_segment;
+  const auto segment_begin         = static_cast<::cuda::std::int64_t>(nth_segment) * full_segment_size;
+  const auto block_segment_offset  = static_cast<::cuda::std::int64_t>(bid % blocks_per_segment) * seg_chunk_size;
```
bernhardmgruber: Question: I have a suspicion that this should cast to `OffsetT` and not `int64_t`. Why is `int64_t` needed here?

cub/cub/device/dispatch/dispatch_fixed_size_segmented_reduce.cuh
srinivasyadav18: I have replaced that with tile_size, which is part of tuning

cub/cub/device/dispatch/dispatch_fixed_size_segmented_reduce.cuh
srinivasyadav18: Yes, `segment_size` is fixed to `int` type from `Device` layer

cub/cub/device/dispatch/kernels/kernel_segmented_reduce.cuh
srinivasyadav18: `num_segments` is of type `int64_t` which determines the index of the segment begin.

cub/cub/device/dispatch/dispatch_fixed_size_segmented_reduce.cuh
```diff
@@ -234,27 +243,162 @@ struct DispatchFixedSizeSegmentedReduce
...
+      launcher_factory(num_current_blocks, ActivePolicyT::ReducePolicy::BLOCK_THREADS, 0, stream)
+        .doit(fixed_size_segmented_reduce_kernel_partial,
+              d_in,
+              d_block_reductions,
+              segment_size,
+              seg_chunk_size,
+              tiles_per_segment,
+              num_current_blocks,
+              reduction_op,
+              init);
```
bernhardmgruber: Important:

```suggestion
      if (const auto error = CubDebug(launcher_factory(num_current_blocks, ActivePolicyT::ReducePolicy::BLOCK_THREADS, 0, stream)
        .doit(fixed_size_segmented_reduce_kernel_partial,
              d_in,
              d_block_reductions,
              segment_size,
              seg_chunk_size,
              tiles_per_segment,
              num_current_blocks,
              reduction_op,
              init)) { return error; }
```

cub/cub/device/dispatch/kernels/kernel_segmented_reduce.cuh
bernhardmgruber: Right now, `num_segments` is of type `int`. Something is not right here ;)

srinivasyadav18: sorry for the confusion. Its not `num_segments`, but both `nth_segment` and `full_segment_size`. consider a case where we have 2^30 segments, and each segment of size 2^14, then `segment_begin` could go beyond 2^31, so we we need `int64_t` for proper indexing.

cub/cub/device/dispatch/dispatch_fixed_size_segmented_reduce.cuh
srinivasyadav18: We can reduce the number of instantions of same kernel from 2 to 1, by adding extra template parameter and parameter to the kernel and using it conditionally. But, not sure how it impacts performance. I will file an issue and try it later.

cub/cub/device/dispatch/dispatch_fixed_size_segmented_reduce.cuh
```diff
@@ -234,27 +243,170 @@ struct DispatchFixedSizeSegmentedReduce
...
+
+    for (::cuda::std::int64_t invocation_index = 0; invocation_index < num_invocations; invocation_index++)
+    {
+      const auto current_seg_offset = invocation_index * num_segments_per_invocation;
+
+      const auto num_current_segments =
+        ::cuda::std::min(num_segments_per_invocation, num_segments - current_seg_offset);
+      const auto num_current_blocks = static_cast<::cuda::std::int32_t>(num_current_segments * tiles_per_segment);
+
+      constexpr int seg_chunk_size = tile_size;
```
NaderAlAwar: Nit: this line feels unnecessary

cub/cub/device/dispatch/dispatch_fixed_size_segmented_reduce.cuh
```diff
@@ -234,27 +243,170 @@ struct DispatchFixedSizeSegmentedReduce
...
+            launcher_factory(num_current_blocks, ActivePolicyT::ReducePolicy::BLOCK_THREADS, 0, stream)
+              .doit(fixed_size_segmented_reduce_kernel_partial,
+                    d_in,
+                    d_block_reductions,
+                    segment_size,
+                    seg_chunk_size,
+                    tiles_per_segment,
+                    num_current_blocks,
+                    reduction_op,
+                    init)))
```
NaderAlAwar: Question: something feels off about passing `init` to both this kernel and the final kernel. In the normal reduce, we only include `init` in the final single tile kernel. I checked the tests and I don't think we have one with a non-zero init value. Could we add a test with a non-zero init value and see if it passes?

cub/cub/device/dispatch/kernels/kernel_segmented_reduce.cuh
```diff
@@ -309,6 +309,56 @@ __launch_bounds__(int(ChainedPolicyT::ActivePolicy::ReducePolicy::BLOCK_THREADS)
...
+          typename InitT,
+          typename AccumT>
+CUB_DETAIL_KERNEL_ATTRIBUTES __launch_bounds__(int(
+  ChainedPolicyT::ActivePolicy::ReducePolicy::
+    BLOCK_THREADS)) void DeviceFixedSizeSegmentedReducePartialKernel(InputIteratorT d_in,
+                                                                     OutputIteratorT d_out,
+                                                                     OffsetT segment_size,
+                                                                     int full_chunk_size,
+                                                                     int blocks_per_segment,
+                                                                     int num_segments,
```
NaderAlAwar: Important: `num_segments` is unused, we should remove it

cub/cub/device/dispatch/kernels/kernel_segmented_reduce.cuh
```diff
@@ -309,6 +309,56 @@ __launch_bounds__(int(ChainedPolicyT::ActivePolicy::ReducePolicy::BLOCK_THREADS)
...
+  const auto segment_id            = bid / blocks_per_segment;
+  const auto segment_begin         = static_cast<::cuda::std::int64_t>(segment_id) * segment_size;
+  const auto chunk_offset          = chunk_id * full_chunk_size;
+  const auto partial_segment_begin = segment_begin + chunk_offset;
+
+  AccumT block_aggregate =
+    AgentReduceT(temp_storage, d_in + partial_segment_begin, reduction_op).ConsumeRange({}, chunk_size);
+  if (tid == 0)
+  {
+    finalize_and_store_aggregate(d_out + bid, reduction_op, reduce::empty_problem_init_t<InitT>{init}, block_aggregate);
```
NaderAlAwar: Question: I believe `init` will be added to the result in the other kernel, is it correct to add `init` here as well?

cub/cub/device/dispatch/dispatch_fixed_size_segmented_reduce.cuh
srinivasyadav18: yes, you are right. But current code works, as kernel might still take `init`, but we wrap it in empty_problem and its ignored. So init is not accumulated twice. But I have cleaned up the code in anycase, as partial kernel accepting init is redundant.

cub/cub/device/dispatch/kernels/kernel_segmented_reduce.cuh
srinivasyadav18: its ignored, as I wrap it in `reduce::empty_problem_init_t<InitT>{init}`

cub/cub/device/dispatch/kernels/kernel_segmented_reduce.cuh
```diff
@@ -309,6 +309,55 @@ __launch_bounds__(int(ChainedPolicyT::ActivePolicy::ReducePolicy::BLOCK_THREADS)
...
+          typename InitT,
+          typename AccumT>
+CUB_DETAIL_KERNEL_ATTRIBUTES __launch_bounds__(int(
+  ChainedPolicyT::ActivePolicy::ReducePolicy::
+    BLOCK_THREADS)) void DeviceFixedSizeSegmentedReducePartialKernel(InputIteratorT d_in,
+                                                                     OutputIteratorT d_out,
+                                                                     OffsetT segment_size,
+                                                                     int full_chunk_size,
+                                                                     int blocks_per_segment,
+
```
NaderAlAwar: Nit: weird newline here

cub/cub/device/dispatch/dispatch_fixed_size_segmented_reduce.cuh
NaderAlAwar: I approved since we have a similar approach in DeviceReduce already. @srinivasyadav18 did you file an issue to reduce the instantiations?

srinivasyadav18: I just did a quick experiment ([b65f5bc](https://github.com/NVIDIA/cccl/pull/7114/commits/b65f5bc372f20420abef8bfe862951c0626fca89)) in reducing number kernels/instantiations from 2/3 to 1/2, and it had no performance impact. Further simplicification would be quite impossible.

srinivasyadav18:
### Performance report 

<details>
<summary>Base bandwidth utilization</summary>
base_single_phase
</details>

<details>
<summary>Optimized two-phase using two kernels</summary>
opt_two_phase
</details>

<details>
<summary>Optimized two-phase using single kernel</summary>
single_kernel_two_phase
</details>

<details>
<summary>
Extermely minimal regressions/improvements in using single kernel </summary>
speedup_base_vs_single_kernel_I32
</details>

PR #7122, opened 2026-01-08
Fix build issues with documentation

miscco:
Our docs currently  do not build. 

I tried to address as many probloms as possible, however there are still some strange errors complains about a missing :end-after: string, but only sometimes :shrug:

Also for whatever reason doxygen sometimes does not find the right module label of the any_resource files?


python/cuda_cccl/cuda/compute/algorithms/_sort/_segmented_sort.py
```diff
@@ -229,14 +229,14 @@ def segmented_sort(
     Example:
         Below, ``segmented_sort`` is used to perform a segmented sort. It also rearranges the values according to the keys' order.
 
-        .. literalinclude:: ../../python/cuda_cccl/tests/parallel/examples/sort/segmented_sort_basic.py
+        .. literalinclude:: ../../python/cuda_cccl/tests/compute/examples/sort/segmented_sort_basic.py
```
miscco, line 232: This is giving a warning about missing `:end-after:`

The same for radix_sort, maybe because there are 2 examples after each other? @shwina

cub/cub/agent/agent_radix_sort_onesweep.cuh
```diff
@@ -58,7 +58,7 @@ enum RadixSortStoreAlgorithm
   RADIX_SORT_STORE_ALIGNED
 };
 
-#if !_CCCL_COMPILER(NVRTC)
+#if !_CCCL_COMPILER(NVRTC) && !defined(_CCCL_DOXYGEN_INVOKED)
 inline ::std::ostream& operator<<(::std::ostream& os, RadixSortStoreAlgorithm algo)
```
bernhardmgruber, line 62: Question: I don't understand why Doxygen cannot handle `operator<<`. I think it should be publicly documented it would be sad to see it go. What's the problem here?

miscco: I dont know either, from the error message I guess it is that it tries to putt the  `<<` into the section header and that does things

bernhardmgruber: Super annoying! I hope we can bring the docs back at some point.

PR #7129, opened 2026-01-08
Add support for `__tuple_like` for `::std::{array, pair, tuple}`

### LINKED ISSUE #7110: [BUG]: CUDA 13.1 cuda::std::mdspan does not implement P3663
### Is this a duplicate?

### Type of Bug

Compile-time Error

### Component

libcu++

### Describe the bug

A customer reports that with CUDA 13.1, `cuda::std::submdspan` does not work with `std::pair` slices (in host code, so `__device__` markings are not relevant).  This means that the implementation does not implement [P3663](https://wg21.link/p3663), which was voted into C++26.  P3663 requires that `submdspan` accept as a (begin, end) slice any type that can be destructured into two values that are both convertible to `index_type`.

For a sample implementation of an earlier draft of P3663, please see [this pull request](https://github.com/kokkos/mdspan/pull/408) on the reference implementation of mdspan.

### How to Reproduce

Attempt to compile the following.

```c++
    cuda::std::mdspan<int, cuda::std::dextents<int, 4>> mds; 
    auto sub = cuda::std::submdspan(mds
                         ,std::pair<int,int>(0,1)
                         ,std::pair<int,int>(0,1)
                         ,std::pair<int,int>(0,1)
                         ,std::pair<int,int>(0,1)
                        );
```

Error looks like this.

```
/local-projects/spack-install/linux-skylake_avx512/cuda-13.1.0-rwyxgxh5a5uxxhzoqojenuffp72b5eue/targets/x86_64-linux/include/cccl/cuda/std/__mdspan/submdspan_extents.h(180): error: static assertion failed with "[mdspan.sub.extents] For each rank index k of src.extents(), exactly one of the following is true:"
    static_assert(((__is_valid_subextents<typename _Extents::index_type, _Slices>) && ... && true),
    ^
          detected during:
            instantiation of "auto cuda::std::__4::submdspan_extents(const _Extents &, _Slices...) [with _Extents=cuda::std::__4::extents<int, 18446744073709551615UL, 18446744073709551615UL, 18446744073709551615UL, 18446744073709551615UL>, _Slices=<std::pair<int, int>, std::pair<int, int>, std::pair<int, int>, std::pair<int, int>>]" at line 187
            instantiation of type "cuda::std::__4::__get_subextents_t<cuda::std::__4::extents<int, 18446744073709551615UL, 18446744073709551615UL, 18446744073709551615UL, 18446744073709551615UL>, std::pair<int, int>, std::pair<int, int>, std::pair<int, int>, std::pair<int, int>>" at line 272 of /local-projects/spack-install/linux-skylake_avx512/cuda-13.1.0-rwyxgxh5a5uxxhzoqojenuffp72b5eue/targets/x86_64-linux/include/cccl/cuda/std/__mdspan/submdspan_mapping.h
            instantiation of "auto cuda::std::__4::__submdspan_mapping_impl(const cuda::std::__4::layout_right::mapping<_Extents> &, _Slices...) [with _Extents=cuda::std::__4::extents<int, 18446744073709551615UL, 18446744073709551615UL, 18446744073709551615UL, 18446744073709551615UL>, _Slices=<std::pair<int, int>, std::pair<int, int>, std::pair<int, int>, std::pair<int, int>>]" at line 319 of /local-projects/spack-install/linux-skylake_avx512/cuda-13.1.0-rwyxgxh5a5uxxhzoqojenuffp72b5eue/targets/x86_64-linux/include/cccl/cuda/std/__mdspan/submdspan_mapping.h
            instantiation of "auto cuda::std::__4::submdspan_mapping(const _LayoutMapping &, _Slices...) [with _LayoutMapping=cuda::std::__4::layout_right::mapping<cuda::std::__4::extents<int, 18446744073709551615UL, 18446744073709551615UL, 18446744073709551615UL, 18446744073709551615UL>>, _Slices=<std::pair<int, int>, std::pair<int, int>, std::pair<int, int>, std::pair<int, int>>]" at line 24 of .../kokkos/core/unit_test/default/TestDefaultDeviceDevelop.cpp

1 error detected in the compilation of ".../kokkos/core/unit_test/default/TestDefaultDeviceDevelop.cpp".
```

### Expected behavior

The example code above should compile.

### Reproduction link

_No response_

### Operating System

_No response_

### nvidia-smi output

_No response_

### NVCC version

_No response_

miscco:
This allows us to use standard library types with submdspan

Fixes [BUG]: CUDA 13.1 cuda::std::mdspan does not implement P3663 Fixes #7110

### DIFF
--- a/libcudacxx/include/cuda/std/__complex/tuple.h
+++ b/libcudacxx/include/cuda/std/__complex/tuple.h
@@ -33,11 +33,11 @@
 _CCCL_BEGIN_NAMESPACE_CUDA_STD
 
 template <class _Tp>
-struct tuple_size<complex<_Tp>> : ::cuda::std::integral_constant<size_t, 2>
+struct tuple_size<complex<_Tp>> : integral_constant<size_t, 2>
 {};
 
 template <size_t _Index, class _Tp>
-  struct tuple_element<_Index, complex<_Tp>> : ::cuda::std::enable_if < _Index<2, _Tp>
+struct tuple_element<_Index, complex<_Tp>> : enable_if<(_Index < 2), _Tp>
 {};
 
 template <class _Tp>
--- a/libcudacxx/include/cuda/std/__fwd/array.h
+++ b/libcudacxx/include/cuda/std/__fwd/array.h
@@ -24,6 +24,24 @@
 
 #include <cuda/std/__cccl/prologue.h>
 
+// std:: forward declarations
+
+#if _CCCL_HAS_HOST_STD_LIB()
+_CCCL_BEGIN_NAMESPACE_STD
+
+#  if _CCCL_HOST_STD_LIB(STL)
+template <class _Tp, size_t _Size>
+class array;
+#  else // ^^^ _CCCL_HOST_STD_LIB(STL) ^^^ / vvv !_CCCL_HOST_STD_LIB(STL) vvv
+template <class _Tp, size_t _Size>
+struct array;
+#  endif // !_CCCL_HOST_STD_LIB(STL)
+
+_CCCL_END_NAMESPACE_STD
+#endif // _CCCL_HAS_HOST_STD_LIB()
+
+// ::cuda::std:: forward declaration
+
 _CCCL_BEGIN_NAMESPACE_CUDA_STD
 
 template <class _Tp, size_t _Size>
--- a/libcudacxx/include/cuda/std/__mdspan/submdspan_helper.h
+++ b/libcudacxx/include/cuda/std/__mdspan/submdspan_helper.h
@@ -30,6 +30,7 @@
 #include <cuda/std/__type_traits/is_same.h>
 #include <cuda/std/__type_traits/is_signed.h>
 #include <cuda/std/__type_traits/is_unsigned.h>
+#include <cuda/std/__type_traits/type_list.h>
 #include <cuda/std/__utility/integer_sequence.h>
 #include <cuda/std/array>
 #include <cuda/std/tuple>
@@ -112,15 +113,24 @@ template <class _IndexType, class _From>
   }
 }
 
+_CCCL_EXEC_CHECK_DISABLE
 template <size_t _Index, class... _Slices>
 [[nodiscard]] _CCCL_API constexpr decltype(auto) __get_slice_at(_Slices&&... __slices) noexcept
 {
-  return ::cuda::std::get<_Index>(::cuda::std::forward_as_tuple(::cuda::std::forward<_Slices>(__slices)...));
+  // Pull in `::std::get` via ADL for host library types
+  using ::cuda::std::get;
+#if _CCCL_COMPILER(MSVC)
+  tuple<_Slices...> __tuple{::cuda::std::forward<_Slices>(__slices)...};
+  return get<_Index>(::cuda::std::move(__tuple));
+#else // ^^^ _CCCL_COMPILER(MSVC) ^^^ / vvv !_CCCL_COMPILER(MSVC) vvv
+  return get<_Index>(::cuda::std::forward_as_tuple(::cuda::std::forward<_Slices>(__slices)...));
+#endif // !_CCCL_COMPILER(MSVC)
 }
 
 template <size_t _Index, class... _Slices>
-using __get_slice_type = tuple_element_t<_Index, __tuple_types<_Slices...>>;
+using __get_slice_type = __type_at_c<_Index, __type_list<_Slices...>>;
 
+_CCCL_EXEC_CHECK_DISABLE
 template <class _IndexType, size_t _Index, class... _Slices>
 [[nodiscard]] _CCCL_API constexpr _IndexType __first_extent_from_slice(_Slices... __slices) noexcept
 {
@@ -136,7 +146,9 @@ template <class _IndexType, size_t _Index, class... _Slices>
   {
     if constexpr (__index_pair_like<_SliceType, _IndexType>)
     {
-      return ::cuda::std::__index_cast<_IndexType>(::cuda::std::get<0>(__slice));
+      // Pull in `::std::get` via ADL for host library types
+      using ::cuda::std::get;
+      return ::cuda::std::__index_cast<_IndexType>(get<0>(__slice));
     }
     else if constexpr (__is_strided_slice<_SliceType>)
     {
@@ -149,6 +161,7 @@ template <class _IndexType, size_t _Index, class... _Slices>
   }
 }
 
+_CCCL_EXEC_CHECK_DISABLE
 template <size_t _Index, class _Extents, class... _Slices>
 [[nodiscard]] _CCCL_API constexpr typename _Extents::index_type
 __last_extent_from_slice(const _Extents& __src, _Slices... __slices) noexcept
@@ -166,7 +179,9 @@ __last_extent_from_slice(const _Extents& __src, _Slices... __slices) noexcept
   {
     if constexpr (__index_pair_like<_SliceType, _IndexType>)
     {
-      return ::cuda::std::__index_cast<_IndexType>(::cuda::std::get<1>(__slice));
+      // Pull in `::std::get` via ADL for host library types
+      using ::cuda::std::get;
+      return ::cuda::std::__index_cast<_IndexType>(get<1>(__slice));
     }
     else if constexpr (__is_strided_slice<_SliceType>)
     {
--- a/libcudacxx/include/cuda/std/__tuple_dir/tuple_element.h
+++ b/libcudacxx/include/cuda/std/__tuple_dir/tuple_element.h
@@ -20,12 +20,17 @@
 #  pragma system_header
 #endif // no system header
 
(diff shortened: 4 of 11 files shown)


libcudacxx/include/cuda/std/__complex/tuple.h
```diff
@@ -100,6 +100,16 @@ template <size_t _Index, class _Tp>
   return __get_complex_impl<_Tp>::template get<_Index>(::cuda::std::move(__z));
 }
 
+#if _CCCL_HAS_HOST_STD_LIB()
+template <class _Tp>
+struct tuple_size<::std::complex<_Tp>> : ::cuda::std::integral_constant<size_t, 2>
+{};
+
+template <size_t _Index, class _Tp>
+  struct tuple_element<_Index, ::std::complex<_Tp>> : ::cuda::std::enable_if < _Index<2, _Tp>
```
davebayer:
```suggestion
struct tuple_element<_Index, ::std::complex<_Tp>> : enable_if<(_Index < 2), _Tp>
```
To get the formatting right

libcudacxx/include/cuda/std/__complex/tuple.h
```diff
@@ -100,6 +100,16 @@ template <size_t _Index, class _Tp>
   return __get_complex_impl<_Tp>::template get<_Index>(::cuda::std::move(__z));
 }
 
+#if _CCCL_HAS_HOST_STD_LIB()
+template <class _Tp>
+struct tuple_size<::std::complex<_Tp>> : ::cuda::std::integral_constant<size_t, 2>
```
davebayer:
```suggestion
struct tuple_size<::std::complex<_Tp>> : integral_constant<size_t, 2>
```
Should be enough

libcudacxx/include/cuda/std/__complex/tuple.h
```diff
@@ -100,6 +100,16 @@ template <size_t _Index, class _Tp>
   return __get_complex_impl<_Tp>::template get<_Index>(::cuda::std::move(__z));
 }
 
+#if _CCCL_HAS_HOST_STD_LIB()
```
miscco: @davebayer do we want to centralize those into a single file?

libcudacxx/include/cuda/std/__fwd/array.h
```diff
@@ -24,6 +24,19 @@
...
 
+// std:: forward declarations
+
+#if _CCCL_HAS_HOST_STD_LIB()
+_CCCL_BEGIN_NAMESPACE_STD
+
+template <class _Tp, size_t _Size>
+struct array;
+
+_CCCL_END_NAMESPACE_STD
```
davebayer, line 40: Did you check whether libstdc++ doesn't put it in some different namespace? It does that for example with `std::string` if I remember correctly. We must be very careful here. Applies to other forward declarations, too

libcudacxx/include/cuda/std/__tuple_dir/tuple_element.h
```diff
@@ -70,6 +70,14 @@ struct _CCCL_TYPE_VISIBILITY_DEFAULT tuple_element<_Ip, tuple<_Tp...>>
   using type _CCCL_NODEBUG_ALIAS = tuple_element_t<_Ip, __tuple_types<_Tp...>>;
 };
 
+#if _CCCL_HAS_HOST_STD_LIB()
+template <size_t _Ip, class... _Tp>
+struct _CCCL_TYPE_VISIBILITY_DEFAULT tuple_element<_Ip, ::std::tuple<_Tp...>>
+{
+  using type _CCCL_NODEBUG_ALIAS = tuple_element_t<_Ip, __tuple_types<_Tp...>>;
+};
+#endif // _CCCL_HAS_HOST_STD_LIB()
```
davebayer, line 97: Do we want to test these specializations?

davebayer (commented):
Shouldn't we add tests for

libcudacxx/include/cuda/std/__complex/tuple.h
davebayer: No, I don't think so

davebayer: I would put them right next to the `cuda::std::` variants

libcudacxx/include/cuda/std/__fwd/array.h
miscco: I believe string is different because it has shared library components

libcudacxx/include/cuda/std/__tuple_dir/tuple_element.h
miscco: yes

davebayer: Would you add the tests then? 😃

libcudacxx/include/cuda/std/__mdspan/submdspan_helper.h
```diff
@@ -112,15 +112,19 @@ template <class _IndexType, class _From>
   }
 }
 
+_CCCL_EXEC_CHECK_DISABLE
 template <size_t _Index, class... _Slices>
 [[nodiscard]] _CCCL_API constexpr decltype(auto) __get_slice_at(_Slices&&... __slices) noexcept
 {
-  return ::cuda::std::get<_Index>(::cuda::std::forward_as_tuple(::cuda::std::forward<_Slices>(__slices)...));
+  // Pull in `::std::get` via ADL for host library types
```
mhoemmen, line 120: P3663 means that any type for which structured binding into two things, both convertible to `index_type`, is well-formed will work as a slice specifier.  For example, the following type works.

```c++
struct Foo {
  int bar;
  int xyzzy;
};
```

Note that this type is not _`tuple-like`_ and that `std::get` does not work on it.

libcudacxx/include/cuda/std/__complex/tuple.h
miscco: I wrote the test and the issue with that is that we would need to include the right folder for that and not only just `<cuda/std/tuple>` but also `<cuda/std/array>`

libcudacxx/include/cuda/std/__tuple_dir/tuple_element.h
miscco: yes

libcudacxx/include/cuda/std/__mdspan/submdspan_helper.h
miscco: I am not sure whether this is something we can easily  implement in C++17 will for now just add this to make sure we can handle the standard types

libcudacxx/include/cuda/std/__fwd/array.h
```diff
@@ -24,6 +24,24 @@
...
+#if _CCCL_HAS_HOST_STD_LIB()
+_CCCL_BEGIN_NAMESPACE_STD
+
+#  if _CCCL_HOST_STD_LIB(STL)
+template <class _Tp, size_t _Size>
+class array;
+#  else // ^^^ _CCCL_HOST_STD_LIB(STL) ^^^ / vvv !_CCCL_HOST_STD_LIB(STL) vvv
+template <class _Tp, size_t _Size>
+struct array;
+#  endif // !_CCCL_HOST_STD_LIB(STL)
```
davebayer, line 38: Yeah, why should anyone bother following the standard..

davebayer (approved):

libcudacxx/include/cuda/std/__mdspan/submdspan_helper.h
mhoemmen: > ... just add this to make sure we can handle the standard types

That's not a complete fix for the bug.

If you go to the _`canonical-slice`_ function definition and look at the implementation following paragraph 7 ("*Effects*: Equivalent to"), you can make it valid C++17 code by not using designated initializers.

The point of _`canonical-slice`_ is to turn every "pair of indices" slice into `strided_slice` with compile-time stride of 1.  The implementation shown in paragraph 7 implements the Mandates in paragraph 6 ("*Mandates*: `S` is a `submdspan` slice type for `IndexType`).  It's certainly possible in C++17 to test whether some type `T` is a specialization of `strided_slice`.

miscco: yeah I need to rewrite the commit message for github to not force close the issue

mhoemmen: I'll say more concisely: _`canonical-slice`_ exists so that you don't need a _`tuple-like`_ test.

miscco: yeah, but that is a much larger change and I am not 100% sure we can emulate it in Cüü17, so I want this in because it is something we would want anyway

mhoemmen: @miscco That's fine; just please be aware that this is not a complete implementation of P3663, as the branch name might otherwise suggest.  Thanks!

PR #7130, opened 2026-01-08
Fixes for shuffle_iterator

RAMitchell:
Fixes the issues described in https://github.com/NVIDIA/cccl/pull/7062#issuecomment-3712125756 and adds many more tests.

### DIFF
--- a/libcudacxx/test/libcudacxx/cuda/iterators/shuffle_iterator/uniformity.pass.cpp
+++ b/libcudacxx/test/libcudacxx/cuda/iterators/shuffle_iterator/uniformity.pass.cpp
@@ -22,6 +22,8 @@
 #include "test_macros.h"
 #include "types.h"
 
+TEST_DIAG_SUPPRESS_MSVC(4146) // unary minus operator applied to unsigned type, result still unsigned
+
 // A lehmer code is a unique index for a permutation
 template <size_t N>
 __host__ __device__ size_t lehmer_code(const cuda::std::array<int, N>& perm)


libcudacxx/include/cuda/__random/feistel_bijection.h
```diff
@@ -57,44 +51,49 @@ class __feistel_bijection
   template <class _RNG>
   _CCCL_API __feistel_bijection(uint64_t __num_elements, _RNG&& __gen)
   {
-    const uint64_t __total_bits = static_cast<uint64_t>(::cuda::std::max(4, ::cuda::std::bit_width(__num_elements)));
+    const uint64_t __total_bits =
+      (::cuda::std::max) (uint64_t{8}, static_cast<uint64_t>(::cuda::std::bit_width(__num_elements)));
```
davebayer:
```suggestion
      ::cuda::std::max(uint64_t{8}, static_cast<uint64_t>(::cuda::std::bit_width(__num_elements)));
```
Should be fine now :) you can leave it as is

davebayer (approved):

libcudacxx/test/libcudacxx/cuda/iterators/shuffle_iterator/uniformity.pass.cpp
```diff
@@ -0,0 +1,390 @@
...
+  double EK2 = expected_K2(n, lambda);
+
+  double varK   = EK2 - EK * EK;
+  double varMMD = 2.0 * varK / double(M);
+
+  return cuda::std::sqrt(2.0 * varMMD) * inverse_erf(1.0 - alpha);
+}
+
+template <size_t N>
+_CCCL_HOST_DEVICE void test_mmd()
```
miscco: In tests we currently still use `__host__ __device__` plainly I would prefer to keep it as such until we change it globally

libcudacxx/include/cuda/__random/feistel_bijection.h
```diff
@@ -57,44 +51,49 @@ class __feistel_bijection
...
 
   [[nodiscard]] _CCCL_API constexpr uint64_t operator()(const uint64_t __val) const noexcept
   {
-    __decomposed __state = {static_cast<uint32_t>(__val >> __right_side_bits),
-                            static_cast<uint32_t>(__val & __right_side_mask)};
-    for (uint32_t i = 0; i < __num_rounds; i++)
+    // Mitchell, Rory, et al. "Bandwidth-optimal random shuffling for GPUs." ACM Transactions on Parallel Computing 9.1
+    // (2022): 1-20.
+    uint32_t __L = __val >> __R_bits_;
+    uint32_t __R = __val & __R_mask_;
```
leofang: Q: Is this part the actual fix, that we swap the definitions of the original `high` and `low`? Otherwise the rest looks the same to me.

miscco (approved):

miscco:
pre-commit.ci autofix

libcudacxx/include/cuda/__random/feistel_bijection.h
RAMitchell: Actually there was a significant bug. The high and low values were being initialised swapped from what they should have been. This meant that with odd number of bits one of the bits was getting removed -  as a consequence the tests would run forever because the bijection was incorrect (i.e. values collided).

I wrote it from scratch again to find the bug as I couldn't find it at first! I relabeled everything with the notation from the paper which is less confusing to me.

RAMitchell:
One thing I would also like to try in a subsequent PR is using a splitmix64 (or similar PRNG sequence) inside the round function to generate the key sequence from a 64 bit starting key instead of carrying around 24 registers with the PRNG keys. The philox PRNG just uses a weyl sequence but according to my tests it is not random enough.

RAMitchell:
Further note: number of rounds is set to 24. Anything less than 12 starts to fail these tests. So 24 is probably a good conservative value for now.

miscco:
/ok to test [013af58](https://github.com/NVIDIA/cccl/pull/7130/commits/013af58a16a4227521924da57c13327b1b558e07)

Issue #4610, opened 2025-05-01
Make `numba-cuda` an optional dependency


leofang:
(@shwina here, hijacking the issue description)

# Making numba-(cuda) an optional dependency of `cuda.compute`

`cuda.compute` relies on `numba.cuda` to JIT compile Python functions passed as operators for algorithms (such as binary operators, unary operators, predicates). For example, the lambda below is compiled into device code with Numba:

```python
# Example 1

reduce_into(d_input, d_output, lambda x, y: x + y, len(d_input), h_init)
```

It is also possible to use pre-defined ops in place of lambdas/functions. The following invocation is equivalent in terms of functionality:

```python
# Example 2

reduce_into(d_input, d_output, OpKind.PLUS, len(d_input), h_init)
```

What's _not_ possible today - and something we would like to enable is the ability to pass pre-compiled device functions:

```python
# Example 3 (NOT POSSIBLE today)

add_op = CompiledOp(my_ltoir, "my_add_i64")  # here, `my_ltoir` is a  bytes object representing a pre-compiled device function

reduce_into(d_input, d_output, add_op, len(d_input), h_init)
```

## What do we want?

* `numba-cuda` should only be required if the user is passing Python functions that need to be JIT compiled into device code. In other words, if the user only wants to use the library as shown in Examples 2 and 3 above (but not Example 1), we shouldn't require the presence of `numba-cuda`.

* For a user with their own compiler (i.e., bringing pre-compiled device code), we want to enable them to use all features of the library. In other words, we want to enable Example 3 above.

## Challenges of making `numba-cuda` optional

At the time of writing, `numba-cuda` is deeply embedded in the internals of `cuda.compute`:

- While NumPy dtypes are used as the user-facing type system, we use `numba-cuda` for our **internal type system**, which means our internal functions and classes pass around `numba-cuda` types to each other. ([example](https://github.com/NVIDIA/cccl/blob/9e72637aa35665dc382e43583201bfd77ee7a4ac/python/cuda_cccl/cuda/compute/_cccl_interop.py#L245)).

[...]

PR #7132, opened 2026-01-08
`cuda.compute`: Bring-your-own-LTOIR

shwina:
## Description

Closes https://github.com/NVIDIA/cccl/issues/4611.

## The problem - tight coupling of `cuda.compute` & `numba(-cuda)`

Currently, the `cuda.compute` functionality and API are tightly coupled with Numba, making Numba and numba-cuda required dependencies.

* Internally, we use the numba type system (see for example [`get_value_type`](https://github.com/NVIDIA/cccl/blob/3f77dd30fe236b4b7405f4a4f2becee05a5391a7/python/cuda_cccl/cuda/compute/_cccl_interop.py#L245)).
* User-defined operations are compiled into LTOIR via Numba, and that LTOIR is passed down to the CCCL C library.
* User-defined types (structs) are created using a `@gpu_struct` decorator which registers the user's types with Numba, enabling them to be used within user-defined operations.
* Iterators are implemented in pure Python, with Numba being used to compile the advance and dereference methods of each iterator ([example](https://github.com/NVIDIA/cccl/blob/3f77dd30fe236b4b7405f4a4f2becee05a5391a7/python/cuda_cccl/cuda/compute/iterators/_iterators.py#L138)).

## What do we want?

We want the ability to use `cuda.compute` without `numba(-cuda)`. Some use cases for this may be:

* Downstream libraries (such as CuPy) may want to be able to use the functionality provided by `cuda.compute`, but not want to take on numba as a dependency
* Users may have pre-existing code (in C++ or another language) that they can compile (using some compiler other than Numba) into LTOIR, and may want to use this with `cuda.compute`. This can be used to implement arbitrary iterators, or ops that may not be implementable with `numba.cuda` today.

## How to achieve this - in an API compatible manner?

This is achieved in an API compatible manner in the following way:

* Two new types, `CompiledOp` and `CompiledIterator` are introduced.
  * `CompiledOp` can be passed by the user anywhere that we currently accept a user-defined Python callable - e.g., as the binary op for a reduction. The `CompiledOp` has:
    * Pre-compiled LTOIR for the operation
    * Symbol name for the function implementing the operation
    * Argument and return types
  * `CompiledIterator` can be passed anywhere that we currently accept iterators. The `CompiledIterator` has:
    * The iterator state (as raw bytes) and its alignment
    * The value type of the iterator (the return type of the dereference operation)
    * Pre-compiled LTOIRs for the advance and dereference operations
* Numba is never imported until it is actually used. This is achieved in the following way:
  * The internal type system is changed to a native one, rather than using numba types
  * Iterators that depend on numba (all the iterators we support currently) are importable, but raise upon construction - complaining that numba is required
  * `gpu_struct` is usable without numba. When numba is present, creating a `gpu_struct` will create and register a corresponding numba type. When numba is not present, the struct is used to figure out the size/type alignment of the type descriptor that will be passed down to the C library. See an example of using `gpu_struct` with a pre-compiled op [here](https://github.com/shwina/cccl/blob/e7bd42425db395e1b9748f2c372506894d42ecf0/python/cuda_cccl/tests/compute/test_compiled.py#L337).

## Performance

TBD

## Future Work

- **Make numba actually optional**: this PR doesn't remove numba as a required dependency, although it gets us 90% of the way there. The only blockers are (1) figuring out what the UX should be for installing and using the library without Numba, and (2) `DiscardIterator` which is [used by `select`](https://github.com/NVIDIA/cccl/blob/c156e907e6b25b62e2b28b37d160fedefc967933/python/cuda_cccl/cuda/compute/algorithms/_select.py#L11), and currently requires Numba. The latter issue is possible to work around by just implementing `DiscardIterator` without numba (trivial).

- **Clearly indicate expected op types for every algorithm**: The user provided (compiled) device functions have all `void*` signature, and we basically assume that the arguments are cast to the expected types within the definition. So for a binary op, we assume that the arguments (and return type) are all cast to `T`, where `T` is the accumulator type - same as `h_init`. With JIT compiled functions, we were largely able to avoid needing to be too explicit about this, as _we_ compiled the user-provided binary op with the appropriate argument and return types. With pre-compiled functions, we have to be explicit. So, it's best for every algorithm to be clear about what the argument and return types are for every op.

## Checklist

### DIFF
--- a/python/cuda_cccl/cuda/compute/algorithms/_select.py
+++ b/python/cuda_cccl/cuda/compute/algorithms/_select.py
@@ -3,6 +3,7 @@
 #
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
+
 from typing import Callable
 
 from .._caching import cache_with_key
--- a/python/cuda_cccl/cuda/compute/compiled/iterator.py
+++ b/python/cuda_cccl/cuda/compute/compiled/iterator.py
@@ -9,10 +9,45 @@ This enables users to bring their own compiler (BYOC) by providing
 pre-compiled LTOIR rather than relying on Numba for JIT compilation.
 """
 
-from typing import Callable
+from typing import Union
+
+import numpy as np
 
 from .._bindings import Iterator, IteratorKind, IteratorState, Op, OpKind
 from ..types import _TypeDescriptor
+from .op import CompiledOp
+
+# Type alias for state input - can be bytes or numpy scalar/array
+StateInput = Union[bytes, np.ndarray, np.generic]
+
+
+def _convert_state(state: StateInput) -> tuple[bytes, int | None]:
+    """
+    Convert state input to (bytes, alignment) tuple.
+
+    Args:
+        state: Raw bytes, numpy scalar, or numpy array
+
+    Returns:
+        Tuple of (state_bytes, alignment)
+    """
+    if isinstance(state, bytes):
+        # Raw bytes - caller must provide alignment separately
+        return state, None
+
+    if isinstance(state, (np.ndarray, np.generic)):
+        # Numpy scalar or array - convert to bytes and infer alignment
+        arr = np.asarray(state)
+        if arr.ndim > 0 and arr.size != 1:
+            # For arrays, ensure contiguous and get bytes
+            arr = np.ascontiguousarray(arr)
+        state_bytes = arr.tobytes()
+        alignment = arr.dtype.alignment
+        return state_bytes, alignment
+
+    raise TypeError(
+        f"state must be bytes, numpy scalar, or numpy array, got {type(state).__name__}"
+    )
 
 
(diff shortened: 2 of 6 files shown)


shwina:
/ok to test https://github.com/NVIDIA/cccl/pull/7132/commits/e7bd42425db395e1b9748f2c372506894d42ecf0

shwina:
/ok to test https://github.com/NVIDIA/cccl/pull/7132/changes/ae29d9890963ad428cb75f7cc66795bfa72e4d60

shwina:
/ok to test https://github.com/NVIDIA/cccl/pull/7132/changes/01f32324dea4c69e9761c2f16296f431af3b7af2

shwina:
Stale PR; closing as all the required functionality is implemented (#4610)

Issue #6761, opened 2025-11-24
[BUG]: Scan incorrectly promises run-to-run determinism


gevtushenko:
### Is this a duplicate?

- [x] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cccl/issues) for this bug and that I agree to the [Code of Conduct](CODE_OF_CONDUCT.md)

### Type of Bug

Silent Failure

### Component

CUB

### Describe the bug

Our scan implementation is non-deterministic in the presense of floating point or unknown operators. We should fail to compile if user requests `run-to-run` determinism, but currently the build succeeds. On the other hand, weakest determinism fails to compile `cuda::execution::determinism::not_guaranteed`.  

### How to Reproduce

```c++
  float* d_input;
  float *d_output;
  int num_items = 0;
  auto env = cuda::execution::require(cuda::execution::determinism::run_to_run);
  cub::DeviceScan::ExclusiveSum(d_input, d_output, num_items, env);
```

### Expected behavior

- static assert if run-to-run determinism cannot be guaranteed
- non-deterministic requirement should always compile

PR #7133, opened 2026-01-08
fix determinism rejection logic for scan

srinivasyadav18:
## Description

closes https://github.com/NVIDIA/cccl/issues/6761

## Checklist

### DIFF
--- a/cub/cub/device/device_scan.cuh
+++ b/cub/cub/device/device_scan.cuh
@@ -174,18 +174,20 @@ struct DeviceScan
                                                     cub::detail::it_value_t<InputIteratorT>,
                                                     typename InitValueT::value_type>>;
 
-    // static assert to reject `run_to_run` determinism if `accum_t` is not integral or op type is not well-known
-    static_assert(!(!::cuda::std::is_same_v<requested_determinism_t, ::cuda::execution::determinism::not_guaranteed_t>
-                    && (!::cuda::std::is_integral_v<accum_t> || !detail::is_cuda_binary_operator<ScanOpT>) ),
-                  "run_to_run or gpu_to_gpu is only supported for integral types with known operators");
+    constexpr bool is_determinism_required =
+      !::cuda::std::is_same_v<requested_determinism_t, ::cuda::execution::determinism::not_guaranteed_t>;
+    constexpr bool is_safe_integral_op =
+      ::cuda::std::is_integral_v<accum_t> && detail::is_cuda_binary_operator<ScanOpT>;
 
-    using determinism_t = ::cuda::execution::determinism::not_guaranteed_t;
+    // Logic: If determinism is required, we must have a safe integral operator.
+    static_assert(!is_determinism_required || is_safe_integral_op,
+                  "run_to_run or gpu_to_gpu is only supported for integral types with known operators");
 
     // Dispatch with environment - handles all boilerplate
     return detail::dispatch_with_env(env, [&]([[maybe_unused]] auto tuning, void* storage, size_t& bytes, auto stream) {
       using tuning_t = decltype(tuning);
       return scan_impl_determinism<tuning_t>(
-        storage, bytes, d_in, d_out, scan_op, init, num_items, determinism_t{}, stream);
+        storage, bytes, d_in, d_out, scan_op, init, num_items, requested_determinism_t{}, stream);
     });
   }
   //! @endcond


cub/cub/device/device_scan.cuh
```diff
@@ -165,15 +165,18 @@ struct DeviceScan
...
       ::cuda::std::execution::__query_result_or_t<requirements_t,
                                                   ::cuda::execution::determinism::__get_determinism_t,
-                                                  ::cuda::execution::determinism::run_to_run_t>;
+                                                  ::cuda::execution::determinism::not_guaranteed_t>;
+
+    // static assert to reject `run_to_run` determinism if input type is not integral or op type is not well-known
+    static_assert(!(::cuda::std::is_same_v<requested_determinism_t, ::cuda::execution::determinism::run_to_run_t>
+                    && (!::cuda::std::is_integral_v<cub::detail::it_value_t<InputIteratorT>>
+                        || !detail::is_cuda_binary_operator<ScanOpT>) ),
+                  "run_to_run determinism is only supported for integral types with well-known operators");
```
miscco: That would never provide any run-to-run determinism anymore, is that intentional?

I would have expected a second conditional

srinivasyadav18: > That would never provide any run-to-run determinism anymore, is that intentional?

No, it means `run_to_run` is **only valid/available** to use for `integral_types` with **well-known operators**.

srinivasyadav18:
TODO: Add a fail test

cub/cub/device/device_scan.cuh
```diff
@@ -165,16 +165,21 @@ struct DeviceScan
...
+                                   ::cuda::std::_If<::cuda::std::is_same_v<InitValueT, NullType>,
+                                                    cub::detail::it_value_t<InputIteratorT>,
+                                                    typename InitValueT::value_type>>;
 
-    static_assert(!::cuda::std::is_same_v<requested_determinism_t, ::cuda::execution::determinism::not_guaranteed_t>,
-                  "not_guaranteed determinism is not supported");
+    // static assert to reject `run_to_run` determinism if `accum_t` is not integral or op type is not well-known
+    static_assert(!(!::cuda::std::is_same_v<requested_determinism_t, ::cuda::execution::determinism::not_guaranteed_t>
+                    && (!::cuda::std::is_integral_v<accum_t> || !detail::is_cuda_binary_operator<ScanOpT>) ),
+                  "run_to_run or gpu_to_gpu is only supported for integral types with known operators");
```
elstehle: question: Wouldn't a scan operator like `cuda::std::max` over `float` still be run-to-run deterministic?

elstehle: This logic is a bit hard to parse with double negations. Can we simplify this to, e.g.,:
```
constexpr bool is_determinism_required = !::cuda::std::is_same_v<requested_determinism_t, ::cuda::execution::determinism::not_guaranteed_t>;
constexpr bool is_safe_integral_op = ::cuda::std::is_integral_v<accum_t> && detail::is_cuda_binary_operator<ScanOpT>;

// Logic: If determinism is required, we must have a safe integral operator.
static_assert(!is_determinism_required || is_safe_integral_op,
              "run_to_run or gpu_to_gpu is only supported for integral types with known operators");
```

cub/cub/device/device_scan.cuh
```diff
@@ -165,16 +165,21 @@ struct DeviceScan
...
 
-    using determinism_t = ::cuda::execution::determinism::run_to_run_t;
+    using determinism_t = ::cuda::execution::determinism::not_guaranteed_t;
```
elstehle: question: Why can't we just resume with `requested_determinism_t` here?

cub/cub/device/device_scan.cuh
srinivasyadav18: > question: Wouldn't a scan operator like cuda::std::max over float still be run-to-run deterministic?

No we reject it.

cub/cub/device/device_scan.cuh
srinivasyadav18: Yes, cleaned up in 68a392c

cub/cub/device/device_scan.cuh
elstehle: I understand that, from this logic, we reject it. But aren't we rejecting it, even though it would produce deterministic results?

PR #7138, opened 2026-01-08
Use dynamic cudart for c2h

davebayer:
Trying to fix cudax compilation. I think the problem is that c2h is compiled with static cudart, but everything else with dynamic.


davebayer:
Doesn't help

PR #7139, opened 2026-01-08
Disable cudax with msvc in CI for now

pciolkosz:
We see some MSVC build failure because cudart is linked twice, disable it for now to not block other PRs


pciolkosz:
First step for https://github.com/NVIDIA/cccl/issues/7140

bernhardmgruber (approved):

PR #7144, opened 2026-01-09
libcu++: silence msvc+nvcc12.9 warning plaguing c.parallel.

griwes:
## Description

Seems like there's shenanigans in nvcc 12.9's output w.r.t. parentheses, and that makes msvc unhappy: https://github.com/NVIDIA/cccl/actions/runs/20770250706/job/59644951459?pr=7007#step:4:1321. Silence that warning.

## Checklist

### DIFF
--- a/libcudacxx/include/cuda/std/__complex/roots.h
+++ b/libcudacxx/include/cuda/std/__complex/roots.h
@@ -78,10 +78,10 @@ template <class _Tp>
 
   // NVCC 12.9 seems to be eliminating some parentheses which makes MSVC unhappy.
   _CCCL_DIAG_PUSH
-#if _CCCL_CUDACC_BELOW(13, 0)
+#if _CCCL_CUDA_COMPILER(NVCC, <, 13, 0)
   _CCCL_DIAG_SUPPRESS_MSVC(4554) // warning C4554: '<<': check operator precedence for possible error; use parentheses
                                  // to clarify precedence
-#endif
+#endif // _CCCL_CUDA_COMPILER(NVCC, <, 13, 0)
 
   // Get some bounds where __re +- |__x| won't overflow.
   // Doesn't need to be too exact, enough to cover extremal cases.


libcudacxx/include/cuda/std/__complex/roots.h
```diff
@@ -76,6 +76,13 @@ template <class _Tp>
   // pre-check to see if we over/underflow:
   _Tp __x_abs_sq = ::cuda::std::fma(__re, __re, __im * __im);
 
+  // NVCC 12.9 seems to be eliminating some parentheses which makes MSVC unhappy.
+  _CCCL_DIAG_PUSH
+#if _CCCL_CUDACC_BELOW(13, 0)
+  _CCCL_DIAG_SUPPRESS_MSVC(4554) // warning C4554: '<<': check operator precedence for possible error; use parentheses
+                                 // to clarify precedence
+#endif
```
davebayer:
```suggestion
  _CCCL_DIAG_PUSH
#if _CCCL_CUDA_COMPILER(NVCC, <, 13, 0)
  _CCCL_DIAG_SUPPRESS_MSVC(4554) // warning C4554: '<<': check operator precedence for possible error; use parentheses
                                 // to clarify precedence
#endif // _CCCL_CUDA_COMPILER(NVCC, <, 13, 0)
```

libcudacxx/include/cuda/std/__complex/roots.h
```diff
@@ -87,6 +94,8 @@ template <class _Tp>
     (static_cast<__uint_t>((static_cast<__uint_t>(-__max_exponent + __mant_nbits) >> 1) + __exp_bias) << __mant_nbits)
     | __fp_explicit_bit_mask_of_v<_Tp>;
 
+  _CCCL_DIAG_POP
```
miscco, line 97: Please move the warning suppression outof the function. We have multiple similar bit operations and I dont want that to reappear

griwes: These were the only ones triggering it and I'd rather be narrow with silencing like this, to avoid getting false negatives without inspection.

bernhardmgruber:
I think this PR did not entirely fix the root cause. I still see the error on my PR which includes the change here: https://github.com/NVIDIA/cccl/actions/runs/20848102214/job/59896493770?pr=7103#step:4:1367

PR #7152, opened 2026-01-09
Fix `make_tma_descriptor()` unit test

fbusato:
## Description

Refactor memory allocation in TMA descriptor tests to use `cudaMalloc` instead of stack allocation to prevent `is_device_accessible()` failure

Fixes nvbug 5778303

### DIFF
--- a/libcudacxx/test/libcudacxx/cuda/tma/make_tma_descriptor.pass.cpp
+++ b/libcudacxx/test/libcudacxx/cuda/tma/make_tma_descriptor.pass.cpp
@@ -51,6 +51,7 @@ bool test_ranks()
   // test 3D tensor + interleave layout 32B + swizzle 32B
   unused(
     cuda::make_tma_descriptor(tensor, box_sizes_3D, cuda::tma_interleave_layout::bytes32, cuda::tma_swizzle::bytes32));
+  assert(cudaFree(data) == cudaSuccess);
   return true;
 }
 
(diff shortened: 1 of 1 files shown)


libcudacxx/test/libcudacxx/cuda/tma/make_tma_descriptor.pass.cpp
```diff
@@ -23,7 +23,8 @@ constexpr auto no_interleave = cuda::tma_interleave_layout::none;
 
 bool test_ranks()
 {
-  alignas(128) float data[64]{};
+  float* data = nullptr;
+  assert(cudaMalloc(&data, 64 * sizeof(float)) == cudaSuccess);
```
bernhardmgruber, line 27: Question: where is this memory deallocated?

fbusato: I didn't add a free because I thought it was not necessary for a test. On the other hand, I think the sanitizer could fail.

fbusato: better to add it

bernhardmgruber: Yes, please add it! It makes debugging easier if the program behaves well.

bernhardmgruber (approved):

PR #7155, opened 2026-01-09
Add support for `[[lifetimebound]]`

fbusato:
## Description

The PR introduces the `_CCCL_LIFETIMEBOUND` macro to handle the `lifetimebound` attribute across different compilers, including Clang and MSVC.

The macro will be useful to trigger compile-time warning for potentially dangling references. 
One example is `mdspan to DLPack` #7027

Thanks @gevtushenko for the idea

### DIFF
--- a/libcudacxx/test/libcudacxx/libcxx/macros/lifetime_bound.compile.fail.cpp
+++ b/libcudacxx/test/libcudacxx/libcxx/macros/lifetime_bound.compile.fail.cpp
@@ -6,6 +6,7 @@
 // SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES.
 //
 //===----------------------------------------------------------------------===//
+// FORCE_ALL_WARNINGS.
 
 #include <cuda/std/__cccl/attributes.h>
 


libcudacxx/include/cuda/std/__cccl/attributes.h
```diff
@@ -154,6 +154,18 @@
 #define _CCCL_NO_SPECIALIZATIONS \
   _CCCL_NO_SPECIALIZATIONS_BECAUSE("Users are not allowed to specialize this cccl entity")
 
+// _CCCL_LIFETIMEBOUND
+
+#if _CCCL_HAS_CPP_ATTRIBUTE(clang::lifetimebound) || _CCCL_COMPILER(CLANG)
+#  define _CCCL_LIFETIMEBOUND [[clang::lifetimebound]]
+#elif _CCCL_HAS_CPP_ATTRIBUTE(msvc::lifetimebound) && _CCCL_COMPILER(MSVC, >=, 19, 37)
+#  define _CCCL_LIFETIMEBOUND [[msvc::lifetimebound]]
```
davebayer, line 162: Do the compilers misreport the support?

libcudacxx/include/cuda/std/__cccl/attributes.h
```diff
@@ -154,6 +154,18 @@
...
+#elif _CCCL_HAS_CPP_ATTRIBUTE(lifetimebound)
+#  define _CCCL_LIFETIMEBOUND [[lifetimebound]]
```
davebayer: Is there any compiler actually supporting this? I don't think so, so I would just remove this branch. It's not a standard c++ attribute

libcudacxx/include/cuda/std/__cccl/attributes.h
```diff
@@ -154,6 +154,18 @@
 #define _CCCL_NO_SPECIALIZATIONS \
   _CCCL_NO_SPECIALIZATIONS_BECAUSE("Users are not allowed to specialize this cccl entity")
 
+// _CCCL_LIFETIMEBOUND
+
+#if _CCCL_HAS_CPP_ATTRIBUTE(clang::lifetimebound) || _CCCL_COMPILER(CLANG)
```
davebayer, line 159: We should add `lifetimebound` to push/pop macros

libcudacxx/test/libcudacxx/libcxx/macros/lifetime_bound.compile.pass.cpp
```diff
@@ -0,0 +1,25 @@
...
+  __host__ __device__ int get() const _CCCL_LIFETIMEBOUND
+  {
+    return 3;
+  }
+};
+
+int main(int, char**)
+{
+  return 0;
+}
```
davebayer, line 25: What do we need this test for? If the compiler cannot handle the attribute, the header test compilation would fail, too, so I am not sure if this has any value

libcudacxx/test/libcudacxx/libcxx/macros/lifetime_bound.compile.fail.cpp
```diff
@@ -0,0 +1,39 @@
...
+//===----------------------------------------------------------------------===//
+
+#include <cuda/std/__cccl/attributes.h>
+
+#include "test_macros.h"
+
+#if (_CCCL_LIFETIMEBOUND + 0)
+#else
+#  error "lifetimebound attribute not supported"
+#endif
```
davebayer, line 19: This doesn't work, see https://godbolt.org/z/8hvPvn9fj

libcudacxx/include/cuda/std/__cccl/attributes.h
fbusato: yes, unfortunately.

libcudacxx/include/cuda/std/__cccl/attributes.h
fbusato: right. I will remove it

libcudacxx/test/libcudacxx/libcxx/macros/lifetime_bound.compile.pass.cpp
fbusato: I want to prevent unexpected behavior. It is a non-standard attribute and not very well-supported. I would prefer to keep a very simple test for it.

libcudacxx/include/cuda/std/__cccl/epilogue.h
```diff
@@ -105,6 +105,14 @@ _CCCL_DIAG_POP
...
 #endif
 
+#if defined(lifetimebound)
+#  error \
+    "cccl internal error: macro `lifetimebound` was redefined between <cuda/std/__cccl/prologue.h> and <cuda/std/__cccl/epilogue.h>"
+#elif defined(_CCCL_POP_MACRO_lifetimebound)
+#  pragma pop_macro("lifetimebound")
+#  undef _CCCL_POP_MACRO_lifetimebound
+#endif
+
```
davebayer: There is a generator now for prologue/epilogue in `libcudacxx/codegen/generate_prologue_epilogue.py`, add it there and regenerate the files, please :)

PR #7172, opened 2026-01-12
Fixes for thrust::shuffle

### LINKED ISSUE #7159: [BUG]: Fix randomness in thrust::shuffle_iterator
### Is this a duplicate?

### Type of Bug

Runtime Error

### Component

Thrust

### Describe the bug

The fix landed in libcudacxx (#7130), which is copied/pasted from Thrust. We need to backport the fix to Thrust, and then back port to v3.2.x.

### How to Reproduce

n/a

### Expected behavior

n/a

### Reproduction link

_No response_

### Operating System

_No response_

### nvidia-smi output

_No response_

### NVCC version

_No response_

RAMitchell:
Fix #7159

Another fix is that bit_with(256)=9. We only need 8, so the calculation becomes total_bits = bit_width(n - 1). The previous implementation is wasting compute for powers of two.

There are some errors in the thrust shuffle randomness tests.
- Chi squared tests being applied to continous quantities, which are also not independant
- Large numbers of chi squared tests being run in a loop. The significance is incorrect when you run multiple tests.

### DIFF
--- a/cub/cub/agent/agent_radix_sort_onesweep.cuh
+++ b/cub/cub/agent/agent_radix_sort_onesweep.cuh
@@ -58,7 +58,7 @@ enum RadixSortStoreAlgorithm
   RADIX_SORT_STORE_ALIGNED
 };
 
-#if !_CCCL_COMPILER(NVRTC)
+#if !_CCCL_COMPILER(NVRTC) && !defined(_CCCL_DOXYGEN_INVOKED)
 inline ::std::ostream& operator<<(::std::ostream& os, RadixSortStoreAlgorithm algo)
 {
   switch (algo)
@@ -71,7 +71,7 @@ inline ::std::ostream& operator<<(::std::ostream& os, RadixSortStoreAlgorithm al
       return os << "<unknown RadixSortStoreAlgorithm: " << static_cast<int>(algo) << ">";
   }
 }
-#endif // !_CCCL_COMPILER(NVRTC)
+#endif // !_CCCL_COMPILER(NVRTC) && !_CCCL_DOXYGEN_INVOKED
 
 template <int NominalBlockThreads4B,
           int NominalItemsPerThread4B,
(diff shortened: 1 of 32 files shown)


libcudacxx/include/cuda/__random/feistel_bijection.h
```diff
@@ -51,7 +51,10 @@ class __feistel_bijection
   template <class _RNG>
   _CCCL_API __feistel_bijection(uint64_t __num_elements, _RNG&& __gen)
   {
-    const uint64_t __total_bits = static_cast<uint64_t>(::cuda::std::max(8, ::cuda::std::bit_width(__num_elements)));
+    // Calculate number of bits needed to represent num_elements - 1
+    // Prevent zero
+    const uint64_t __max_index  = ::cuda::std::max(static_cast<uint64_t>(1), __num_elements) - 1;
```
miscco, line 56: Technically this would regress the case of 0 elements

RAMitchell: Doesn't `__num_elements=0` result in `__total_bits=8`? I don't see the mistake. There is also a test for this case which results in size 256.

miscco: oh yeah you are correct

thrust/thrust/detail/random_bijection.h
```diff
@@ -34,25 +35,22 @@ namespace detail
 //! \brief A Feistel cipher for operating on power of two sized problems
 class feistel_bijection
```
miscco, line 36: Just for me, why cannot we just do
```suggestion
using feistel_bijection = ::cuda::__feistel_bijection;
```

RAMitchell: `::cuda::__feistel_bijection` uses a URBG inside and when we use the thrust one it fails to compile. As you mentioned one option would be to allow uniform_int_distribution to accept an incorrect URBG. Otherwise we have to duplicate.

leofang:
Much appreciated, @RAMitchell 🙏

PR #7174, opened 2026-01-12
Add msvc2026 to ci matrix.

wmaxey:
## Description

Adds support for testing MSVC 2026 with 14.50.

## Checklist


miscco:
/ok to test [d73dcff](https://github.com/NVIDIA/cccl/pull/7174/commits/d73dcff6c14ffab82d97fbb998b2a7e828f31ba8)

.github/actions/workflow-build/build-workflow.py
```diff
@@ -501,6 +501,21 @@ def generate_dispatch_job_image(matrix_job, job_type):
...
+
+    env = "'["
+    comma = ""
+
+    for environ in matrix_job["environment"]:
+        env += f'{comma}"{environ}"'
+        comma = ","
+
+    env += "]'"
+    return env
```
alliepiper: This function is used to build a python object that will get converted to a json string later. It should just return a python list containing the environment strings.

alliepiper: I think this should be sufficient?

```suggestion
    env = matrix_job["environment"]
    return env if env else []
```

ci/matrix.yaml
```diff
@@ -387,9 +390,11 @@ host_compilers:
     container_tag: 'cl'
     exe: cl
     versions:
-      14.29: { stds: [17,   ], alias: '2019' }
-      14.39: { stds: [17, 20] } # CTK 12.0 doesn't recognize >14.39 as MSVC 2022.
-      14.44: { stds: [17, 20], alias: '2022' }
+      '14.29': { stds: [17,   ], alias: '2019' }
+      '14.39': { stds: [17, 20] } # CTK 12.0 doesn't recognize >14.39 as MSVC 2022.
+      '14.44': { stds: [17, 20], alias: '2022' }
+      '14.50': { stds: [17, 20], alias: '2026' }
```
alliepiper, line 433: Suggestion: to avoid having to change all of the `msvc` -> `msvc2022` while testing this, you can define a new host compiler `msvc_preview` or something in this `host_compilers` block:

```
  msvc_preview:
    name: 'MSVC (Preview)'
    container_tag: 'cl'
    exe: cl
    versions:
      '14.50': { stds: [17, 20], alias: '2026' }
```

Then just define an override matrix with targeted jobs that use `cxx: 'msvc_preview'`.

alliepiper (dismissed):

.github/actions/workflow-build/build-workflow.py
wmaxey: I obviously have completely forgotten everything about this by now, but I think that it wasn't sufficient because it needed extra quoting. I'll try to look back at this.

ci/matrix.yaml
wmaxey: Ah! Nice! I was struggling with some of the surgery to prevent extra builds kicking off.

alliepiper (approved):

davebayer (approved):

Jacobfaib (approved):

PR #7175, opened 2026-01-13
Provide `mdspan` `layout_stride_relaxed`

fbusato:
## Description

Standard library `mdspan` layouts have two main constraints that prevent the adoption on a wider number of use cases:

- **Only allow strictly positive strides** -> (a) prevents broadcasting, (b) fully-compatibility with other tensor formats, such as [DLPack ](https://github.com/dmlc/dlpack).
- **Cannot mix static and dynamic extends** -> while `layout_right/left` derives the strides from the extends, allowing any kind of combinations, `layout_stride` enforces dynamic-only strides. Padded layouts specify a static stride for only one dimension.

The PR proposes a _relaxed_ `layout_stride`, where the strides are derived through an extent type. This greatly improves flexibility by enabling the combination of static and dynamic extents. Secondly, the new layout does not enforce constraints on zero or negative strides.

Credit to @mhoemmen for the initial implementation.


libcudacxx/include/cuda/__mdspan/layout_stride_relaxed.h
```diff
@@ -0,0 +1,445 @@
...
+  using size_type    = typename extents_type::size_type;
+  using rank_type    = typename extents_type::rank_type;
+  using layout_type  = layout_stride_relaxed;
+
+private:
+  static constexpr rank_type __rank_    = extents_type::rank();
+  static constexpr auto __rank_sequence = ::cuda::std::make_index_sequence<extents_type::rank()>();
+
+  using __offset_type     = ::cuda::std::intptr_t;
+  using __stride_array    = ::cuda::std::array<__offset_type, __rank_>;
```
fbusato: using `__offset_type` (`intptr_t`) for the stride array looks inconsistent with the other standard layouts.

libcudacxx/include/cuda/__mdspan/layout_stride_relaxed.h
```diff
@@ -0,0 +1,445 @@
...
+
+  _CCCL_HIDE_FROM_ABI constexpr mapping(const mapping&) noexcept            = default;
+  _CCCL_HIDE_FROM_ABI constexpr mapping& operator=(const mapping&) noexcept = default;
+
+  _CCCL_TEMPLATE(class _OtherIndexType)
+  _CCCL_REQUIRES(::cuda::std::is_convertible_v<const _OtherIndexType&, __offset_type> _CCCL_AND(
+    ::cuda::std::is_nothrow_constructible_v<__offset_type, const _OtherIndexType&>))
+  _CCCL_API constexpr mapping(const extents_type& __ext,
+                              ::cuda::std::span<_OtherIndexType, __rank_> __span_strides,
+                              __offset_type __offset = 0) noexcept
```
fbusato: Adopted `__offset_type` instead of `size_t`

libcudacxx/include/cuda/__mdspan/layout_stride_relaxed.h
```diff
@@ -0,0 +1,445 @@
...
+    else
+    {
+      // The dot product of indices and strides is linear.
+      // Thus, over all valid indices, the max value of the dot product is achieved at the extrema: either the min
+      // index (0) if the stride is negative, or the max index (extent(r) - 1) if the stride is non-negative.
+      index_type __max_indices[__rank_];
+      for (rank_type r = 0; r < __rank_; ++r)
+      {
+        const index_type __ext         = __extents_.extent(r);
+        const index_type __ext_minus_1 = __ext == 0 ? index_type{0} : __ext - index_type{1};
```
fbusato: if an extent is zero, should not `required_span_size` also return zero?

libcudacxx/include/cuda/__mdspan/layout_stride_relaxed.h
```diff
@@ -0,0 +1,445 @@
...
+  using rank_type    = typename extents_type::rank_type;
+  using layout_type  = layout_stride_relaxed;
+
+private:
+  static constexpr rank_type __rank_    = extents_type::rank();
+  static constexpr auto __rank_sequence = ::cuda::std::make_index_sequence<extents_type::rank()>();
+
+  using __offset_type     = ::cuda::std::intptr_t;
+  using __stride_array    = ::cuda::std::array<__offset_type, __rank_>;
+  using __index_compute_t = decltype(index_type{} + __offset_type{});
```
fbusato: `cuda::std::common_type` is probably a better choice

libcudacxx/include/cuda/__mdspan/layout_stride_relaxed.h
```diff
@@ -0,0 +1,445 @@
...
+//! - Negative strides (for reverse iteration)
+//! - Zero strides (for broadcasting)
+//! - A base offset (to accommodate negative strides)
+//!
+//! @note This layout is NOT always unique, exhaustive, or strided in the standard sense.
+struct layout_stride_relaxed
+{
+  template <class _Extents>
+  class mapping;
+};
```
miscco: Please move  this into `<cuda/__fwd/mdspan.h>`

libcudacxx/include/cuda/__mdspan/layout_stride_relaxed.h
```diff
@@ -0,0 +1,445 @@
...
+__layout_stride_relaxed_compute_offset(const _StridedMapping& __mapping, ::cuda::std::index_sequence<_Pos...>) noexcept
+{
+  using __index_type = typename _StridedMapping::index_type;
+  return static_cast<__index_type>(__mapping((static_cast<void>(_Pos), __index_type{0})...));
+}
+
+template <class _StridedMapping>
+[[nodiscard]] _CCCL_API constexpr auto __layout_stride_relaxed_compute_offset(const _StridedMapping& __mapping) noexcept
+{
+  using __stride_extents = typename _StridedMapping::extents_type;
```
miscco: In the existing mdspan implementation this would be:
```suggestion
  using _Extents   = typename _StridedMapping::extents_type;
```

libcudacxx/include/cuda/__mdspan/layout_stride_relaxed.h
```diff
@@ -0,0 +1,445 @@
...
+  if constexpr (__stride_extents::rank() == 0)
+  {
+    return static_cast<typename _StridedMapping::index_type>(__mapping());
+  }
+  else
+  {
+    // Check if any extent is zero
+    using __stride_rank = typename _StridedMapping::rank_type;
```
miscco: This should just be
```suggestion
    using rank_type = typename _StridedMapping::rank_type;
```

libcudacxx/include/cuda/__mdspan/layout_stride_relaxed.h
miscco: In `layout_stride` we use `index_type` as the type

Also this is storing all the strides, even those that are statically known. We should probably do something like extents where we only store the dynamic ones

libcudacxx/include/cuda/__mdspan/layout_stride_relaxed.h
miscco: This should just be `index_type`

libcudacxx/include/cuda/__mdspan/layout_stride_relaxed.h
```diff
@@ -0,0 +1,445 @@
...
+      }
+    }
+  }
+
+  _CCCL_HIDE_FROM_ABI constexpr mapping(const mapping&) noexcept            = default;
+  _CCCL_HIDE_FROM_ABI constexpr mapping& operator=(const mapping&) noexcept = default;
+
+  _CCCL_TEMPLATE(class _OtherIndexType)
+  _CCCL_REQUIRES(::cuda::std::is_convertible_v<const _OtherIndexType&, __offset_type> _CCCL_AND(
+    ::cuda::std::is_nothrow_constructible_v<__offset_type, const _OtherIndexType&>))
```
miscco: Why would this need to be nothrow_constructible?

libcudacxx/include/cuda/__mdspan/layout_stride_relaxed.h
```diff
@@ -0,0 +1,445 @@
...
+        _CCCL_ASSERT(!::cuda::overflow_cast<__offset_type>(__span_strides[__d]),
+                     "layout_stride_relaxed::mapping: stride is out of range");
+        __strides_[__d] = static_cast<__offset_type>(__span_strides[__d]);
+      }
+    }
+  }
+
+  _CCCL_TEMPLATE(class _OtherIndexType)
+  _CCCL_REQUIRES(::cuda::std::is_convertible_v<const _OtherIndexType&, __offset_type> _CCCL_AND(
+    ::cuda::std::is_nothrow_constructible_v<__offset_type, const _OtherIndexType&>))
```
miscco: Ditto should be 
```suggestion
    ::cuda::std::is_constructible_v<__offset_type, const _OtherIndexType&>))
```

libcudacxx/include/cuda/__mdspan/layout_stride_relaxed.h
```diff
@@ -0,0 +1,445 @@
...
+  _CCCL_API constexpr mapping(const extents_type& __ext,
+                              const ::cuda::std::array<_OtherIndexType, __rank_>& __array_strides,
```
miscco: This constructor should just forward to the span one

miscco (commented):
Looks great already, we will need some work to make it work efficiently with static strides though

libcudacxx/include/cuda/__mdspan/layout_stride_relaxed.h
davebayer: I know it's the same type in the end, but wouldn't make sense it was a signed `size_t`, so `cuda::std::make_signed_t<cuda::std::size_t>`?

libcudacxx/include/cuda/__mdspan/layout_stride_relaxed.h
davebayer: Or just
```suggestion
    using typename _StridedMapping::rank_type;
```

libcudacxx/include/cuda/__mdspan/layout_stride_relaxed.h
mhoemmen: @fbusato wrote:
> using `__offset_type` (`intptr_t`) for the stride array looks inconsistent with the other standard layouts.

That's a good point.  The layout mapping's `operator()` always returns `index_type`, so there's no reason to use a larger integer type.

There's a good reason to make the strides signed: With a positive offset and negative strides, it's still possible to get a nonnegative result.  The implementation of `mapping::operator()` needs to cast intermediate results correctly, though.  I don't think I did that in the prototype.

This suggests using `make_signed_t<index_type>` as the type of the strides and the offset.

libcudacxx/include/cuda/__mdspan/layout_stride_relaxed.h
mhoemmen: That's a good point; thanks!

libcudacxx/include/cuda/__mdspan/layout_stride_relaxed.h
miscco: Shouldnt index_type just be a signed integer if using negative strides?

mhoemmen: @miscco wrote:
> Shouldn't `index_type` just be a signed integer if using negative strides?

One motivation for permitting negative strides with unsigned `index_type` is that the default `index_type` is `size_t`.  It's easy to get that by using `dims` or CTAD with `extents`.

```c++
// extents<size_t, dynamic_extent, dynamic_extent, dynamic_extent>
std::extents e{3, 5, 7};
std::layout_stride_relaxed::mapping m{e, std::array{-1, -3, -15}, 104};
```

mhoemmen: Another example:
```c++
std::layout_stride_relaxed::mapping m{
  std::dims<3>(3, 5, 7), std::array{-1, -3, -15}, 104};
```

libcudacxx/include/cuda/__mdspan/layout_stride_relaxed.h
fbusato: the function is marked `noexcept` and `_OtherIndexType` could be an object too. I think it makes sense

libcudacxx/include/cuda/__mdspan/layout_stride_relaxed.h
```diff
@@ -0,0 +1,450 @@
...
+template <class _Extents>
+class layout_stride_relaxed::mapping
+{
+public:
+  static_assert(::cuda::std::__is_cuda_std_extents_v<_Extents>,
+                "layout_stride_relaxed::mapping template argument must be a specialization of extents.");
+
+  using extents_type = _Extents;
+  using index_type   = typename extents_type::index_type;
+  using offset_type  = ::cuda::std::make_signed_t<index_type>;
```
fbusato: I think `offset_type` should be exposed because it is used in the public method `stride()` and in the ctors

libcudacxx/include/cuda/std/__mdspan/extents.h
```diff
@@ -684,6 +703,31 @@ _CCCL_HOST_DEVICE extents(_IndexTypes...) -> extents<size_t, __to_dynamic_extent
 
 namespace __mdspan_detail
 {
+// ------------------------------------------------------------------
+// ------------ __required_span_size_is_representable ---------------
+// ------------------------------------------------------------------
+
+// Checks if the product of extents is representable as index_type without overflow
+template <class _Extents>
+[[nodiscard]] _CCCL_API constexpr bool __required_span_size_is_representable(const _Extents& __ext) noexcept
```
fbusato, line 650: This change is no longer necessary. I could undo it, but I would prefer to keep it.

libcudacxx/include/cuda/__mdspan/layout_stride_relaxed.h
```diff
@@ -0,0 +1,450 @@
...
+    return false;
+  }
+
+  //! @brief Returns true if offset is zero (standard strided behavior)
+  [[nodiscard]] _CCCL_API constexpr bool is_strided() const noexcept
+  {
+    return __offset_ == 0;
+  }
+
+  [[nodiscard]] _CCCL_API constexpr offset_type stride(rank_type __r) const noexcept
```
fbusato, line 370: NOTE: `stride(r)` returns `offset_type` not `index_type` as the other standard layouts

libcudacxx/include/cuda/__mdspan/layout_stride_relaxed.h
fbusato: > We should probably do something like extents where we only store the dynamic ones

that's a good point. Could we reuse `extents` here? what about a second PR for that?

libcudacxx/include/cuda/__mdspan/layout_stride_relaxed.h
fbusato: I think we can directly use `extents` instead of reimplementing all the machinery

fbusato: `extents`  doesn't work because the guard value is `numeric_limits<size_t>::max()` which is unsigned. I need to create a similar class for strides

libcudacxx/include/cuda/__mdspan/layout_stride_relaxed.h
mhoemmen: > We should probably do something like extents where we only store the dynamic ones.

That's an excellent idea.  It didn't occur to me because I figured you were always pulling in strides from DLPack at run time, but it's harmless to make the class support mixed compile-time and run-time strides as well.

libcudacxx/include/cuda/__mdspan/layout_stride_relaxed.h
mhoemmen: That's right -- `extents` represents compile-time extents as `size_t`.

mhoemmen: > the function is marked `noexcept` and `_OtherIndexType` could be an object too. I think it makes sense

It's not required to make the constructor `noexcept`; that's a design choice.

libcudacxx/include/cuda/__mdspan/layout_stride_relaxed.h
mhoemmen: @fbusato That's a good point!  The layout mapping requirements ask that `m.stride(r)` return `index_type`.  It's a fun legal question whether a non-strided layout (a layout for which `is_strided()` always returns `false`) can have a `stride` member function whose return type is not `index_type`.

mhoemmen: @fbusato I'm pretty sure, though, that is if `is_strided()` returns `true`, then the return type of `stride` should be `index_type` in order for the layout mapping to be considered "strided."  That doesn't mean the strides need to be stored _internally_ as `index_type`, though.

libcudacxx/include/cuda/__mdspan/layout_stride_relaxed.h
```diff
@@ -0,0 +1,424 @@
...
+/***********************************************************************************************************************
+ * layout_stride_relaxed::mapping
+ **********************************************************************************************************************/
+
+template <class _Extents, class _Stride>
+class layout_stride_relaxed::mapping
+{
+public:
+  static_assert(::cuda::std::__is_cuda_std_extents_v<_Extents>,
+                "layout_stride_relaxed::mapping template argument must be a specialization of extents.");
```
griwes, line 105: There should also be a similar check for `_Stride`, no?

libcudacxx/include/cuda/__mdspan/layout_stride_relaxed.h
```diff
@@ -0,0 +1,424 @@
...
+    requires(::cuda::std::is_convertible_v<typename _StridedLayoutMapping::extents_type, _Extents>),
+    requires(::cuda::std::__mdspan_detail::__is_mapping_of<::cuda::std::layout_left, _StridedLayoutMapping>
+             || ::cuda::std::__mdspan_detail::__is_mapping_of<::cuda::std::layout_right, _StridedLayoutMapping>
+             || ::cuda::std::__mdspan_detail::__is_mapping_of<::cuda::std::layout_stride, _StridedLayoutMapping>));
+
+/***********************************************************************************************************************
+ * layout_stride_relaxed::mapping
+ **********************************************************************************************************************/
+
+template <class _Extents, class _Stride>
```
griwes: `s/_Stride/_Strides/`?

mhoemmen:
@fbusato It occurs to me to check whether there is an `explicit` conversion to `layout_stride`, and an implicit conversion from `layout_stride` (and possibly also `layout_left*` and `layout_right*`).

libcudacxx/include/cuda/__mdspan/layout_stride_relaxed.h
fbusato: maybe `is_strided()` could also check if all strides are strictly positive

fbusato:
I added the explicit conversion to `layout_stride` [here](https://github.com/NVIDIA/cccl/pull/7175/changes#diff-d50565d2a32c964c160d62c4026eaae6722c4da819540d938bacd38cfbdc9d4fR393).
All implicit conversions are listed [here](https://github.com/NVIDIA/cccl/pull/7175/changes#diff-d50565d2a32c964c160d62c4026eaae6722c4da819540d938bacd38cfbdc9d4fR88)

libcudacxx/include/cuda/__mdspan/layout_stride_relaxed.h
fbusato: good point, yes!

libcudacxx/include/cuda/__mdspan/layout_stride_relaxed.h
fbusato: `_Strides`

docs/libcudacxx/extended_api/mdspan/layout_stride_relaxed.rst
```diff
@@ -0,0 +1,416 @@
...
+.. list-table::
+    :widths: 40 60
+    :header-rows: 1
+
+    * - **Non-member Functions**
+      - Description
+    * - ``operator==``
+      - ``true`` if ranks are equal and all stride values compare equal; ``false`` otherwise.
+    * - ``operator!=``
+      - ``false`` if ranks are equal and all stride values compare equal; ``true`` otherwise.
```
mhoemmen:
```suggestion
      - ``false`` if ranks are equal and all extent and stride values compare equal; ``true`` otherwise.
```

docs/libcudacxx/extended_api/mdspan/layout_stride_relaxed.rst
```diff
@@ -0,0 +1,416 @@
...
+      - Returns the stride at dimension ``r``.
+
+.. list-table::
+    :widths: 40 60
+    :header-rows: 1
+
+    * - **Non-member Functions**
+      - Description
+    * - ``operator==``
+      - ``true`` if ranks are equal and all stride values compare equal; ``false`` otherwise.
```
mhoemmen:
```suggestion
      - ``true`` if ranks are equal and all extent and stride values compare equal; ``false`` otherwise.
```

docs/libcudacxx/extended_api/mdspan/layout_stride_relaxed.rst
```diff
@@ -0,0 +1,416 @@
...
+      - ``layout_stride_relaxed``
+
+.. list-table::
+    :widths: 40 60
+    :header-rows: 1
+
+    * - **Static Member Functions**
+      - Description
+    * - ``is_always_unique()``
+      - Returns ``false``. Uniqueness is not guaranteed due to zero/negative strides.
```
mhoemmen, line 237: It occurs to me that we could guarantee uniqueness or stridedness, if we knew the offset value at compile time.  That would mean templating the layout (or layout mapping) on the offset type as well, and letting that be either integral or _`integral-constant-like`_.

docs/libcudacxx/extended_api/mdspan/layout_stride_relaxed.rst
```diff
@@ -0,0 +1,416 @@
...
+      - Returns the required span size to cover all valid indices.
+    * - ``operator()(Indices... indices)``
+      - Maps multidimensional indices to a linear index.
+    * - ``is_unique()``
+      - Returns ``false`` (conservative).
+    * - ``is_exhaustive()``
+      - Returns ``false`` (conservative).
+    * - ``is_strided()``
+      - Returns ``true`` if offset is zero, ``false`` otherwise.
+    * - ``stride(rank_type r)``
```
mhoemmen, line 265: The [layout mapping requirements](https://eel.is/c++draft/mdspan.layout.reqmts) say that if the mapping is a strided mapping, then `stride` returns `index_type`.  It's an interesting question whether returning an integer type convertible to `index_type` is acceptable for requirements written in this way.  I'm fine with this design, though.

docs/libcudacxx/extended_api/mdspan/layout_stride_relaxed.rst
```diff
@@ -0,0 +1,416 @@
...
+- **(3)** Direct-non-list-initializes the extents, strides, and offset with the provided arguments.
+
+  - *Preconditions*:
+
+    - ``offset`` is nonnegative.
+    - ``required_span_size()`` is representable as ``index_type``.
+
+- **(4)** Constructs the mapping by copying extents and strides from ``other``. For ``layout_stride_relaxed`` sources, also copies the offset.
+
+  - The constructor is ``explicit`` if ``OtherMapping::extents_type`` is not convertible to ``extents_type``, or (for non-``layout_stride_relaxed`` mappings) the source is not a ``layout_left``, ``layout_right``, or ``layout_stride`` mapping.
```
mhoemmen, line 309: What about `layout_left_padded` and `layout_right_padded`?  For those two layouts, it should also be straightforward to extract a compile-time stride if available.

docs/libcudacxx/extended_api/mdspan/layout_stride_relaxed.rst
```diff
@@ -0,0 +1,416 @@
...
+
+**Mixed strides**
+
+.. code:: cpp
+
+    #include <cuda/mdspan>
+    #include <cassert>
+
+    int main() {
+        // 2D array with column-major layout but reversed rows
```
mhoemmen, line 423: Excellent example!

mhoemmen (changes requested):

docs/libcudacxx/extended_api/mdspan/layout_stride_relaxed.rst
fbusato: yes, it is but we don't have these two layouts yet :(. I don't think it is a good idea to mention them directly.

docs/libcudacxx/extended_api/mdspan/layout_stride_relaxed.rst
fbusato: that's a great idea. In this way, we will get the properties at compile-time. I have to think about if using `integral-constant-like` doesn't break something.

libcudacxx/include/cuda/__fwd/mdspan.h
```diff
@@ -0,0 +1,91 @@
...
+//! @brief Tag value indicating a dynamic stride (similar to dynamic_extent for extents)
+inline constexpr ::cuda::std::ptrdiff_t dynamic_stride = (::cuda::std::numeric_limits<::cuda::std::ptrdiff_t>::min)();
+
+namespace __strides_detail
+{
+template <class _OffsetType, class _Seq>
+struct __make_dstrides_impl;
+
+template <class _OffsetType, ::cuda::std::size_t... _Idx>
+struct __make_dstrides_impl<_OffsetType, ::cuda::std::index_sequence<_Idx...>>
```
miscco: Critical: This does not match with `strides` For strides you use `ptrdiff_t` and for `make_dstrides_impl` you use size_t

We should use `ptrdiff_t` consistently here

This also implies lack of testing

libcudacxx/include/cuda/__fwd/mdspan.h
```diff
@@ -0,0 +1,91 @@
...
+};
+} // namespace __strides_detail
+
+//! @brief Alias template for strides with all dynamic stride values
+template <class _OffsetType, ::cuda::std::size_t _Rank>
+using dstrides =
+  typename __strides_detail::__make_dstrides_impl<_OffsetType, ::cuda::std::make_index_sequence<_Rank>>::type;
+
+template <::cuda::std::size_t _Rank, class _OffsetType = ::cuda::std::ptrdiff_t>
+using steps = dstrides<_OffsetType, _Rank>;
```
miscco, line 64: Ditto should use `ptrdiff_t`

libcudacxx/include/cuda/__mdspan/layout_stride_relaxed.h
```diff
@@ -0,0 +1,464 @@
...
+#include <cuda/std/array>
+
+#include <cuda/std/__cccl/prologue.h>
+
+_CCCL_BEGIN_NAMESPACE_CUDA
+
+//! @brief Computes the effective offset of a strided mapping by calling mapping(0, 0, ...).
+//! Standard layouts (layout_left, layout_right, layout_stride) don't have an explicit offset() member,
+//! but layout_stride_relaxed does. This helper enables comparing offsets in operator==.
+template <class _StridedMapping, ::cuda::std::size_t... _Pos>
```
miscco, line 53: I am again wondering why we are not using an `integer_sequence<ptrdiff_t, ...>`

libcudacxx/include/cuda/__mdspan/layout_stride_relaxed.h
```diff
@@ -0,0 +1,464 @@
...
+template <class _StridedMapping>
+[[nodiscard]] _CCCL_API constexpr auto __layout_stride_relaxed_compute_offset(const _StridedMapping& __mapping) noexcept
+{
+  using _Extents        = typename _StridedMapping::extents_type;
+  using _RankType       = typename _StridedMapping::rank_type;
+  constexpr auto __rank = _Extents::rank();
+  if constexpr (__rank > 0) // avoid pointless comparison of unsigned integer with zero warning
+  {
+    // Check if any extent is zero - can't call mapping(0,...) in that case
+    for (_RankType __r = 0; __r < __rank; ++__r)
```
miscco: Informational: We can avoid the warning by comparing for unequal
```suggestion
    for (_RankType __r = 0; __r != __rank; ++__r)
```

libcudacxx/include/cuda/__mdspan/layout_stride_relaxed.h
```diff
@@ -0,0 +1,464 @@
...
+      }
+      return strides_type(__init_strides);
+    }
+  }
+
+  [[nodiscard]] _CCCL_API constexpr bool __has_positive_strides() const noexcept
+  {
+    if constexpr (__rank_ > 0) // avoid pointless comparison of unsigned integer with zero warning
+    {
+      for (rank_type __r = 0; __r < __rank_; ++__r)
```
miscco: Ditto we can avoid all such clauses by comparing rank against `!=` if the loop starts at 0

miscco (changes requested):
Looks great already, we will need some work to make it work efficiently with static strides though

docs/libcudacxx/extended_api/mdspan/layout_stride_relaxed.rst
```diff
@@ -0,0 +1,416 @@
...
+    template <class OffsetType, ptrdiff_t... Strides>
+    class strides;
+
+    // Alias for all-dynamic strides
+    template <class OffsetType, size_t Rank>
+    using dstrides = strides<OffsetType, /* Rank dynamic_stride values */>;
+
+    // Alias for steps (synonym for dstrides)
+    template <size_t Rank, class OffsetType = ptrdiff_t>
+    using steps = dstrides<OffsetType, Rank>;
```
davebayer: I've never header `steps` to be used in this context, I am not a fan of this alias

libcudacxx/include/cuda/__fwd/mdspan.h
fbusato: great catch! true, I didn't test `dstrides`

libcudacxx/include/cuda/__mdspan/layout_stride_relaxed.h
fbusato: true but the compiler optimizes the code better with `<`. Probably no differences here

libcudacxx/include/cuda/__mdspan/layout_stride_relaxed.h
fbusato: `size_t` looks correct here because it is used for rank positions. `ptrdiff_t` should be only used for strides

docs/libcudacxx/extended_api/mdspan/layout_stride_relaxed.rst
mhoemmen: @fbusato We'll want to check that _`integral-constant-like`_ has an up-to-date definition ( https://eel.is/c++draft/span.syn#concept:integral-constant-like ), given that it's changed in C++26.

mhoemmen (approved):

docs/libcudacxx/extended_api/mdspan/layout_stride_relaxed.rst
fbusato: Definitely, we use it at multiple points.

libcudacxx/include/cuda/__fwd/mdspan.h
fbusato: just realized that this doesn't make any difference because `strides` uses `ptrdiff_t`.
I can slightly improve it by using `integer_sequence`

libcudacxx/include/cuda/__fwd/mdspan.h
fbusato: `Rank` of type `ptrdiff_t` is correct

libcudacxx/include/cuda/std/__type_traits/integral_constant_like.h
```diff
@@ -0,0 +1,50 @@
...
+#include <cuda/std/__type_traits/is_integral.h>
+#include <cuda/std/__type_traits/is_same.h>
+#include <cuda/std/__type_traits/remove_cvref.h>
+
+#include <cuda/std/__cccl/prologue.h>
+
+_CCCL_BEGIN_NAMESPACE_CUDA_STD
+
+template <class _Tp>
+_CCCL_CONCEPT __integral_constant_like = _CCCL_REQUIRES_EXPR((_Tp)) //
```
fbusato, line 37: @miscco this concept is used in several points, so I moved it to a common location.

libcudacxx/include/cuda/std/__type_traits/integral_constant_like.h
```diff
@@ -0,0 +1,50 @@
+//===----------------------------------------------------------------------===//
```
mhoemmen, line 1: @fbusato Should we consider calling this file `integral_constant_like.h` instead of `integer_constant_like.h`?

libcudacxx/include/cuda/std/__type_traits/integer_constant_like.h
```diff
@@ -0,0 +1,50 @@
...
+
+template <class _Tp>
+_CCCL_CONCEPT __integral_constant_like = _CCCL_REQUIRES_EXPR((_Tp)) //
+  ( //
+    requires(is_integral_v<remove_cvref_t<decltype(_Tp::value)>>), //
+    requires(!is_same_v<bool, remove_cvref_t<decltype(_Tp::value)>>), //
+    requires(convertible_to<_Tp, decltype(_Tp::value)>), //
+    requires(equality_comparable_with<_Tp, decltype(_Tp::value)>), //
+    requires(_Tp() == _Tp::value), //
+    requires(static_cast<decltype(_Tp::value)>(_Tp()) == _Tp::value) //
```
mhoemmen: This last bit isn't _quite_ right, per https://eel.is/c++draft/span.syn#concept:integral-constant-like .

```
bool_constant<static_cast<decltype(T::value)>(T()) == T::value>::value;
```

The difference is that we want `operator==` to be `constexpr` and to return something convertible to `bool`.  Just requiring that `operator==` be well-formed isn't enough here.

fbusato: thanks for the explanation, @mhoemmen. I thought that `bool_constant` is used for compatibility reasons.

libcudacxx/include/cuda/std/__type_traits/integral_constant_like.h
miscco: yeah definitely

libcudacxx/include/cuda/__mdspan/layout_stride_relaxed.h
mhoemmen: `mdspan` generally uses `extents::rank_type = size_t` for rank positions.

libcudacxx/include/cuda/std/__type_traits/integral_constant_like.h
```diff
@@ -0,0 +1,51 @@
...
+// Part of libcu++, the C++ Standard Library for your entire system,
+// under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef _CUDA_STD___TYPE_TRAITS_INTEGER_CONSTANT_LIKE_H
+#define _CUDA_STD___TYPE_TRAITS_INTEGER_CONSTANT_LIKE_H
```
miscco: Please fix the header guard

libcudacxx/test/libcudacxx/cuda/containers/views/mdspan/layout_stride_relaxed/assertions.pass.cpp
```diff
@@ -0,0 +1,72 @@
...
+//
+// Part of libcu++, the C++ Standard Library for your entire system,
+// under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES.
+//
+//===----------------------------------------------------------------------===//
+
+// REQUIRES: has-unix-headers
```
miscco, line 11: What would we need unix for?

libcudacxx/test/libcudacxx/cuda/containers/views/mdspan/layout_stride_relaxed/comparison.pass.cpp
```diff
@@ -0,0 +1,263 @@
...
+  [[maybe_unused]] constexpr size_t D = cuda::std::dynamic_extent;
+
+  // Rank-0 cases
+  test_comparison(
+    true,
+    cuda::std::extents<T1>(),
+    cuda::std::extents<T2>(),
+    cuda::std::array<intptr_t, 0>{},
+    cuda::std::array<intptr_t, 0>{});
+  test_comparison(
```
miscco, line 99: Please add some spacing, its hard to read
```suggestion

  test_comparison(
```

libcudacxx/test/libcudacxx/cuda/containers/views/mdspan/layout_stride_relaxed/comparison.pass.cpp
```diff
@@ -0,0 +1,263 @@
...
+  cuda::std::array<intptr_t, E1::rank()> strides,
+  intptr_t offset,
+  E2 e2,
+  OtherArgs... other_args)
+{
+  using layout_type  = cuda::layout_stride_relaxed::mapping<E1>;
+  using strides_type = typename layout_type::strides_type;
+  using offset_type  = typename layout_type::offset_type;
+  layout_type map(e1, strides_type(strides), static_cast<offset_type>(offset));
+  typename OtherLayout::template mapping<E2> other_map(e2, other_args...);
```
miscco: I believe that should have an alias too
```suggestion
  using other_mapping = typename OtherLayout::template mapping<E2>;
  other_mapping other_map(e2, other_args...);
```

libcudacxx/test/libcudacxx/cuda/containers/views/mdspan/layout_stride_relaxed/comparison.pass.cpp
```diff
@@ -0,0 +1,263 @@
...
+  auto dest_offset1      = static_cast<offset_type_dest>(dest_offset);
+  auto src_offset1       = static_cast<offset_type_src>(src_offset);
+  dest_mapping dest(dest_exts, strides_type_dst(dest_strides), dest_offset1);
+  src_mapping src(src_exts, strides_type_src(src_strides), src_offset1);
+  static_assert(noexcept(dest == src));
+  assert((dest == src) == equal);
+  assert((dest != src) == !equal);
+}
+
+template <class T1, class T2>
```
miscco: Here we have T1 and T2, above its To, From which seems clearer

libcudacxx/test/libcudacxx/cuda/containers/views/mdspan/layout_stride_relaxed/comparison.pass.cpp
```diff
@@ -0,0 +1,263 @@
...
+  intptr_t src_offset  = 0)
+{
+  using dest_mapping     = cuda::layout_stride_relaxed::mapping<To>;
+  using src_mapping      = cuda::layout_stride_relaxed::mapping<From>;
+  using strides_type_dst = typename dest_mapping::strides_type;
+  using strides_type_src = typename src_mapping::strides_type;
+  using offset_type_dest = typename dest_mapping::offset_type;
+  using offset_type_src  = typename src_mapping::offset_type;
+  auto dest_offset1      = static_cast<offset_type_dest>(dest_offset);
+  auto src_offset1       = static_cast<offset_type_src>(src_offset);
```
miscco: I find this test setup super complicated and hard to follow, especially given that all this is repeated boilerplate.

Can you inline that function and see whether that improves the tests?

libcudacxx/test/libcudacxx/cuda/containers/views/mdspan/layout_stride_relaxed/conversion.layout_stride.pass.cpp
```diff
@@ -0,0 +1,77 @@
...
+  using RelaxedMapping = cuda::layout_stride_relaxed::template mapping<E>;
+  using StrideMapping  = cuda::std::layout_stride::template mapping<E>;
+  using strides_type   = typename RelaxedMapping::strides_type;
+  RelaxedMapping src(ext, strides_type(input_strides), 0); // zero offset
+  static_assert(!cuda::std::is_convertible_v<RelaxedMapping, StrideMapping>,
+                "conversion should be explicit, not implicit");
+  static_assert(cuda::std::is_constructible_v<StrideMapping, RelaxedMapping>,
+                "conversion should be possible via explicit cast");
+  static_assert(noexcept(static_cast<StrideMapping>(src)), "conversion should be noexcept");
+
```
miscco, line 36: Please space the tests out, they are really hard to read
```suggestion

  RelaxedMapping src(ext, strides_type(input_strides), 0); // zero offset
  static_assert(!cuda::std::is_convertible_v<RelaxedMapping, StrideMapping>,
                "conversion should be explicit, not implicit");
  static_assert(cuda::std::is_constructible_v<StrideMapping, RelaxedMapping>,
                "conversion should be possible via explicit cast");
  static_assert(noexcept(static_cast<StrideMapping>(src)), "conversion should be noexcept");

```

libcudacxx/test/libcudacxx/cuda/containers/views/mdspan/layout_stride_relaxed/ctor.default.pass.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+#include "test_macros.h"
+
+template <class E, cuda::std::enable_if_t<E::rank() != 0, int> = 0>
+__host__ __device__ constexpr void test_construction()
+{
+  using M = cuda::layout_stride_relaxed::mapping<E>;
+  static_assert(noexcept(M{}));
+  M m;
+  E e;
```
miscco: please initialize, a lot of compilers will complain otherwise
```suggestion
  M m{};
  E e{};
```

libcudacxx/test/libcudacxx/cuda/containers/views/mdspan/layout_stride_relaxed/ctor.default.pass.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+  // check strides: uses layout_right strides by default
+  auto strides_obj = m.strides();
+  static_assert(noexcept(m.strides()));
+  cuda::std::layout_right::mapping<E> m_right{};
+  for (typename E::rank_type r = 0; r < E::rank(); r++)
+  {
+    assert(cuda::std::cmp_equal(m.stride(r), m_right.stride(r)));
+    assert(cuda::std::cmp_equal(strides_obj.stride(r), m_right.stride(r)));
+  }
```
miscco: We can merge both functions
```suggestion
  static_assert(noexcept(m.strides()));
  
  if constexpr (E::rank() != 0) {
    auto strides_obj = m.strides();
    cuda::std::layout_right::mapping<E> m_right{};
    for (typename E::rank_type r = 0; r < E::rank(); r++)
    {
      assert(cuda::std::cmp_equal(m.stride(r), m_right.stride(r)));
      assert(cuda::std::cmp_equal(strides_obj.stride(r), m_right.stride(r)));
    }
  }
```

libcudacxx/test/libcudacxx/cuda/containers/views/mdspan/layout_stride_relaxed/ctor.extents_strides.pass.cpp
```diff
@@ -0,0 +1,161 @@
...
+  assert(m.offset() == offset);
+
+  // check strides
+  auto strides_obj = m.strides();
+  static_assert(noexcept(m.strides()));
+  for (typename E::rank_type r = 0; r < E::rank(); r++)
+  {
+    assert(cuda::std::cmp_equal(m.stride(r), s[r]));
+    assert(cuda::std::cmp_equal(strides_obj.stride(r), s[r]));
+  }
```
miscco: Ditto merge both functions
```suggestion
  // check strides
  static_assert(noexcept(m.strides()));
  
  if constexpr (E::rank() != 0) {
    auto strides_obj = m.strides();
    for (typename E::rank_type r = 0; r < E::rank(); r++)
    {
      assert(cuda::std::cmp_equal(m.stride(r), s[r]));
      assert(cuda::std::cmp_equal(strides_obj.stride(r), s[r]));
    }
  }
```

libcudacxx/test/libcudacxx/cuda/containers/views/mdspan/layout_stride_relaxed/ctor.extents_strides.pass.cpp
```diff
@@ -0,0 +1,161 @@
...
+  test_construction(cuda::std::extents<int, 0>(), cuda::std::array<intptr_t, 1>{1});
+  test_construction(cuda::std::extents<int, D>(0), cuda::std::array<intptr_t, 1>{1});
+  test_construction(cuda::std::extents<int, 0>(), cuda::std::array<intptr_t, 1>{1}, 10);
+
+  // All extents zero (multiple dimensions)
+  test_construction(cuda::std::extents<int, 0, 0>(), cuda::std::array<intptr_t, 2>{1, 1});
+  test_construction(cuda::std::extents<int, D, D>(0, 0), cuda::std::array<intptr_t, 2>{8, 1});
+  test_construction(cuda::std::extents<int, 0, 0, 0>(), cuda::std::array<intptr_t, 3>{1, 1, 1});
+
+  // All extents zero with non-zero offset
```
miscco, line 93: Love those tests, great job

libcudacxx/test/libcudacxx/cuda/containers/views/mdspan/layout_stride_relaxed/ctor.strided_mapping.pass.cpp
```diff
@@ -0,0 +1,208 @@
...
+template <class FromL,
+          class FromExt,
+          cuda::std::enable_if_t<cuda::std::is_same<FromL, cuda::std::layout_stride>::value, int> = 0>
+__host__ __device__ constexpr auto get_strides(FromExt src_exts)
+{
+  using From = typename FromL::template mapping<FromExt>;
+
+  // just construct some strides which aren't layout_left/layout_right
+  cuda::std::array<size_t, FromExt::rank()> strides{};
+  size_t stride = 2;
```
miscco: Would prefer if if constexpr and one function

libcudacxx/test/libcudacxx/cuda/containers/views/mdspan/layout_stride_relaxed/ctor.strided_mapping.pass.cpp
```diff
@@ -0,0 +1,208 @@
...
+  if constexpr (ToExt::rank() > 0)
+  {
+    for (typename ToExt::rank_type r = 0; r < ToExt::rank(); r++)
+    {
+      assert(cuda::std::cmp_equal(dest.stride(r), src.stride(r)));
+    }
+  }
+
+  To dest_implicit = src;
+  assert(dest_implicit.extents() == src.extents());
```
miscco: Ditto `if constexpr`

libcudacxx/test/libcudacxx/cuda/containers/views/mdspan/layout_stride_relaxed/index_operator.pass.cpp
```diff
@@ -0,0 +1,162 @@
...
+
+template <class Mapping, class... Indices>
+_CCCL_CONCEPT operator_constraints = _CCCL_REQUIRES_EXPR((Mapping, variadic Indices), Mapping m, Indices... idxs)(
+  _Same_as(typename Mapping::index_type) m(idxs...));
+
+_CCCL_TEMPLATE(class Mapping, class... Indices)
+_CCCL_REQUIRES(operator_constraints<Mapping, Indices...>)
+__host__ __device__ constexpr bool check_operator_constraints(Mapping m, Indices... idxs)
+{
+  (void) m(idxs...);
```
miscco: if constexpr

libcudacxx/test/libcudacxx/cuda/containers/views/mdspan/layout_stride_relaxed/negative_stride_unsigned_index.fail.cpp
```diff
@@ -0,0 +1,49 @@
...
+    const auto expected = int64_t{offset} + static_cast<int64_t>(i) * -2;
+    const auto got      = static_cast<int64_t>(mapping(i));
+    assert(got == expected);
+  }
+  return true;
+}
+
+int main(int, char**)
+{
+  assert(test());
```
miscco:
```suggestion
  test();
```

libcudacxx/test/libcudacxx/cuda/containers/views/mdspan/layout_stride_relaxed/negative_stride_unsigned_index.pass.cpp
```diff
@@ -0,0 +1,50 @@
...
+    const auto expected = int64_t{offset} + static_cast<int64_t>(i) * -2;
+    const auto got      = static_cast<int64_t>(mapping(i));
+    assert(got == expected);
+  }
+  return true;
+}
+
+int main(int, char**)
+{
+  assert(test());
```
miscco, line 47:
```suggestion
  test();
```

libcudacxx/test/libcudacxx/cuda/containers/views/mdspan/layout_stride_relaxed/offset.pass.cpp
```diff
@@ -0,0 +1,175 @@
...
+__host__ __device__ constexpr void test_2d_partial_reverse()
+{
+  // 2D array 3x4 where the first dimension is reversed
+  // Layout: physical[offset - i*stride0 + j*stride1]
+  //
+  // With strides = {-4, 1} and offset = 8:
+  // (0,0) -> 8 - 0*4 + 0*1 = 8
+  // (1,0) -> 8 - 1*4 + 0*1 = 4
+  // (2,0) -> 8 - 2*4 + 0*1 = 0
+  // (0,1) -> 8 - 0*4 + 1*1 = 9
```
miscco: Can we move that above the asserts where its actually tested?

libcudacxx/test/libcudacxx/cuda/containers/views/mdspan/layout_stride_relaxed/properties.pass.cpp
```diff
@@ -0,0 +1,198 @@
...
+#include <cuda/std/type_traits>
+
+#include "test_macros.h"
+
+using cuda::std::intptr_t;
+
+template <class E, class M, cuda::std::enable_if_t<(E::rank() > 0), int> = 0>
+__host__ __device__ constexpr void test_strides(E ext, M& m, const M& c_m, cuda::std::array<intptr_t, E::rank()> strides)
+{
+  for (typename E::rank_type r = 0; r < E::rank(); r++)
```
miscco: Do we actually need the two functions at all?

libcudacxx/include/cuda/__mdspan/layout_stride_relaxed.h
```diff
@@ -0,0 +1,480 @@
...
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef _CUDA___MDSPAN_LAYOUT_STRIDE_RELAXED_H
+#define _CUDA___MDSPAN_LAYOUT_STRIDE_RELAXED_H
+
+#include <cuda/std/detail/__config>
+
+#include <cuda/std/__type_traits/is_integer.h>
```
miscco: Critical: Move to the right place

libcudacxx/include/cuda/__mdspan/layout_stride_relaxed.h
```diff
@@ -0,0 +1,480 @@
...
+  using _RankType       = typename _StridedMapping::rank_type;
+  constexpr auto __rank = _Extents::rank();
+  // Check if any extent is zero - can't call mapping(0,...) in that case
+  for (_RankType __r = 0; __r != __rank; ++__r)
+  {
+    if (__mapping.extents().extent(__r) == 0)
+    {
+      return typename _StridedMapping::index_type{0};
+    }
+  }
```
miscco, line 75: Question: Would it be more efficient to only check the static ranks in a constexpr function? Fot the rank to be zero it *must* be static???

libcudacxx/include/cuda/__mdspan/layout_stride_relaxed.h
```diff
@@ -0,0 +1,480 @@
...
+    {
+      return _OffsetType::value;
+    }
+    else
+    {
+      return _OffsetType{};
+    }
+  }
+
+  using __offset_t = decltype(__get_offset_t());
```
miscco: isnt that just `decltype(cuda::std::__de_ice(_OffsetType))`

libcudacxx/include/cuda/__mdspan/layout_stride_relaxed.h
```diff
@@ -0,0 +1,480 @@
...
+
+  static constexpr rank_type __rank_    = extents_type::rank();
+  static constexpr auto __rank_sequence = ::cuda::std::make_index_sequence<extents_type::rank()>();
+
+  extents_type __extents_{};
+  strides_type __strides_{};
+  offset_type __offset_{};
```
miscco: Critical: We should use empty base class optimization as in the other layouts

libcudacxx/include/cuda/__mdspan/layout_stride_relaxed.h
```diff
@@ -0,0 +1,480 @@
...
+        {
+          return index_type{0};
+        }
+        _CCCL_ASSERT(!::cuda::overflow_cast<index_type>(::cuda::uabs(__strides_.stride(__r))),
+                     "layout_stride_relaxed::mapping: stride is out of range");
+        const auto __max_index = __strides_.stride(__r) < 0 ? index_type{0} : static_cast<index_type>(__ext - 1);
+        const auto __stride    = static_cast<index_type>(__strides_.stride(__r));
+        _CCCL_ASSERT(!::cuda::std::__mdspan_detail::__mul_overflow(__max_index, __stride)
+                       && !::cuda::add_overflow(__max_index * __stride, __dot),
+                     "layout_stride_relaxed::mapping: required_span_size is not representable as index_type");
```
miscco: This really feels like we should have the overflow functions in place and just check the result of overflow detection

miscco (commented):
Overall this already looks great

I do have some product code questions and request for improvement of the tests.

mdspan is a beast and some tests are really complicated to understand

libcudacxx/test/libcudacxx/cuda/containers/views/mdspan/layout_stride_relaxed/assertions.pass.cpp
fbusato: it is needed for `TEST_CCCL_ASSERT_FAILURE`, or am I wrong?

libcudacxx/test/libcudacxx/cuda/containers/views/mdspan/layout_stride_relaxed/comparison.pass.cpp
fbusato: I pratically adapted the other layout tests. Anyway, I will try to simplify it

libcudacxx/test/libcudacxx/cuda/containers/views/mdspan/layout_stride_relaxed/ctor.default.pass.cpp
fbusato: all tests passed. I will apply it anyway.

libcudacxx/include/cuda/__mdspan/layout_stride_relaxed.h
fbusato: not really. A dynamic extent can be zero at runtime (`>=0` constraint).
I don't think it is worth to split the function to check static and dynamic extents separaterly. `__mapping.extents().extent(__r) == 0` is already evaluated at compile-time if `extent(__r)` is a static

libcudacxx/include/cuda/__mdspan/layout_stride_relaxed.h
fbusato: calling these functions in release mode can be much more expensive than just the builtin operator, even if only check the overflow in `_CCCL_ASSERT`

libcudacxx/include/cuda/__mdspan/layout_stride_relaxed.h
fbusato: this was quite hard to address. 

`__mdspan_ebco<_Extents, _Strides, _OffsetType>` with `_OffsetType` equal to `integral_constant` causes a recursion.
one argument `is_constructible_v<_Elem1, const __mdspan_ebco&>` checks `__mdspan_ebco` again.

fbusato:
@miscco thanks. great feedback. I tried to further simplify the unit tests.

libcudacxx/include/cuda/std/__mdspan/extents.h
```diff
@@ -395,6 +395,25 @@ _CCCL_REQUIRES(integral<_To>)
...
+  *__res = __x * __y;
+  return __x && ((*__res / __x) != __y);
+}
+
+template <class _Tp>
+[[nodiscard]] _CCCL_API constexpr bool __mul_overflow(_Tp __x, _Tp __y) noexcept
+{
+  const auto __res = __x * __y;
+  return __x && ((__res / __x) != __y);
+}
```
davebayer: Use `cuda::mul_overflow` now, please :)

libcudacxx/include/cuda/__fwd/mdspan.h
```diff
@@ -0,0 +1,92 @@
...
+//! Similar to extents, but for strides. Supports both static (compile-time known)
+//! and dynamic (runtime) stride values. Uses dynamic_stride as the tag for dynamic values.
+//!
+//! @tparam _OffsetType The signed integer type for stride values (supports negative strides)
+//! @tparam _Strides... The stride values, where dynamic_stride indicates a runtime value
+template <class _OffsetType, ::cuda::std::ptrdiff_t... _Strides>
+class strides;
+
+//! @brief Tag value indicating a dynamic stride (similar to dynamic_extent for extents)
+inline constexpr ::cuda::std::ptrdiff_t dynamic_stride = (::cuda::std::numeric_limits<::cuda::std::ptrdiff_t>::min)();
```
davebayer:
```suggestion
inline constexpr ::cuda::std::ptrdiff_t dynamic_stride = ::cuda::std::numeric_limits<::cuda::std::ptrdiff_t>::min();
```

libcudacxx/include/cuda/__mdspan/layout_stride_relaxed.h
```diff
@@ -0,0 +1,458 @@
...
+             || ::cuda::std::__mdspan_detail::__is_mapping_of<::cuda::std::layout_right, _StridedLayoutMapping>
+             || ::cuda::std::__mdspan_detail::__is_mapping_of<::cuda::std::layout_stride, _StridedLayoutMapping>));
+
+/***********************************************************************************************************************
+ * layout_stride_relaxed::mapping
+ **********************************************************************************************************************/
+
+template <class _Extents, class _Strides, class _OffsetType>
+class _CCCL_DECLSPEC_EMPTY_BASES layout_stride_relaxed::mapping
+    : private ::cuda::std::__mdspan_ebco<_Extents, _Strides, _OffsetType>
```
fbusato, line 100: @miscco it is essential to keep the offset as the last field for nvmath Python compatibility. I want to double check that `__mdspan_ebco` keeps the order.

leofang: It's CuPy not nvmath-python, but tomayto, tomahto 😆 

It'd be great to ensure the memory layout is tested in the CI.

libcudacxx/include/cuda/__mdspan/layout_stride_relaxed.h
miscco: Its fine that it is generally used, but its a bug if a user uses a different type, I believe we should be defensive and make this generic. It wont cost anything if its size_t and work with other rank types

libcudacxx/include/cuda/__mdspan/layout_stride_relaxed.h
miscco: Technically, I believe there is no strict requirement on the layout of multiple inheritance classes. That said, I believe it is universally in order of definition

libcudacxx/include/cuda/__mdspan/layout_stride_relaxed.h
```diff
@@ -0,0 +1,458 @@
...
+  [[nodiscard]] _CCCL_API static constexpr strides_type __make_strides(const _StridedLayoutMapping& __other) noexcept
+  {
+    if constexpr (__rank_ == 0)
+    {
+      return strides_type{};
+    }
+    else
+    {
+      ::cuda::std::array<__offset_t, __rank_> __init_strides{};
+      for (rank_type __d = 0; __d < __rank_; ++__d)
```
miscco, line 133: Even if this is in the else branch I would prefer if this was
```suggestion
      for (rank_type __d = 0; __d != __rank_; ++__d)
```

libcudacxx/include/cuda/__mdspan/layout_stride_relaxed.h
```diff
@@ -0,0 +1,458 @@
...
+
+  _CCCL_HIDE_FROM_ABI constexpr mapping(const mapping&) noexcept            = default;
+  _CCCL_HIDE_FROM_ABI constexpr mapping& operator=(const mapping&) noexcept = default;
+
+  _CCCL_API constexpr mapping(
+    const extents_type& __ext, const strides_type& __strides, offset_type __offset = offset_type{}) noexcept
+      : __base(__ext, __strides, __offset)
+  {
+    _CCCL_ASSERT(::cuda::std::cmp_greater_equal(static_cast<__offset_t>(offset()), __offset_t{0}),
+                 "layout_stride_relaxed::mapping: offset must be nonnegative");
```
miscco: I believe this is an overkill, we could simply do
```suggestion
    _CCCL_ASSERT(::cuda::std::is_unsigned_v<__offset_t> ? true : __offset >= 0),
                 "layout_stride_relaxed::mapping: offset must be nonnegative");
```

Applies throughout

libcudacxx/include/cuda/__mdspan/layout_stride_relaxed.h
```diff
@@ -0,0 +1,458 @@
...
+
+  //! @brief Returns true if offset is zero (standard strided behavior)
+  [[nodiscard]] _CCCL_API constexpr bool is_strided() const noexcept
+  {
+    return static_cast<__offset_t>(offset()) == __offset_t{0};
+  }
+
+  [[nodiscard]] _CCCL_API constexpr __offset_t stride(rank_type __r) const noexcept
+  {
+    _CCCL_ASSERT(__r < __rank_, "layout_stride_relaxed::mapping::stride(): invalid rank index");
```
miscco: This might warn if `__rank_ == 0`

libcudacxx/include/cuda/__mdspan/layout_stride_relaxed.h
```diff
@@ -0,0 +1,458 @@
...
+  using strides_type = _Strides;
+  using index_type   = typename extents_type::index_type;
+  using size_type    = typename extents_type::size_type;
+  using rank_type    = typename extents_type::rank_type;
+  using offset_type  = _OffsetType;
+  using layout_type  = layout_stride_relaxed;
+  using __base       = ::cuda::std::__mdspan_ebco<_Extents, _Strides, _OffsetType>;
+
+private:
+  using __offset_t = decltype(::cuda::std::__de_ice(_OffsetType{}));
```
miscco: I believe this should be required to be the same as `offset_type`

libcudacxx/include/cuda/__mdspan/layout_stride_relaxed.h
```diff
@@ -0,0 +1,458 @@
...
+  template <::cuda::std::size_t... _Pos, class... _Indices>
+  [[nodiscard]] _CCCL_API constexpr index_type
+  __compute_index(::cuda::std::index_sequence<_Pos...>, _Indices... __indices) const noexcept
+  {
+    _CCCL_ASSERT((__is_valid_index(__indices) && ...),
+                 "layout_stride_relaxed::mapping: index is not representable as index_type");
+    _CCCL_ASSERT(((static_cast<index_type>(__indices) < extents().extent(_Pos)) && ...),
+                 "layout_stride_relaxed::mapping: index is out of bounds");
+    return (static_cast<index_type>(offset()) + ...
+            + (static_cast<index_type>(__indices) * static_cast<index_type>(strides().stride(_Pos))));
```
miscco, line 311: We probably should worry about overflow here

libcudacxx/include/cuda/std/__mdspan/empty_base.h
```diff
@@ -241,13 +266,38 @@ struct _CCCL_DECLSPEC_EMPTY_BASES __mdspan_ebco<_Elem1, _Elem2, _Elem3>
   {}
 
   template <class _Arg1>
-  static constexpr bool __is_constructible_from_one_arg =
-    is_constructible_v<_Elem1, _Arg1> && is_default_constructible_v<_Elem2> && is_default_constructible_v<_Elem3>;
+  _CCCL_API static constexpr bool __is_constructible_from_one_arg_fn() noexcept
```
miscco: I believe we can avoid the need for that condition by defaulting the copy / move assignment / constructors

libcudacxx/include/cuda/__mdspan/layout_stride_relaxed.h
fbusato: `cmp_greater_equal` practically translates to the same expression. I would keep it because it is more readable

libcudacxx/include/cuda/__mdspan/layout_stride_relaxed.h
fbusato: wait, why? `__indices` is an user value, it could excess the extents and wrapper around/UB and `::cuda::std::cmp_less(__index, required_span_size())` could fail

fbusato: we can probably drop `::cuda::std::cmp_less(__index, required_span_size())` because we already check it in the constructor

libcudacxx/include/cuda/__mdspan/layout_stride_relaxed.h
fbusato: tried, but `stride(__d)` triggers a compile-time error with rank=0

libcudacxx/include/cuda/std/__mdspan/empty_base.h
fbusato: this was pretty hard to address due to clang failures

miscco (approved):

NaderAlAwar (approved):
Approved python part

PR #6767, opened 2025-11-24
Implement the new tuning API for `DeviceRadixSort`

bernhardmgruber:
PR is not fully ready yet, but ready to take a first round of review.

- [x] `cub.test.device.radix_sort_keys.lid_0.key_bits_16` passes
- [x] CCCL.C tests pass
- [x] Retain the `if constexpr` on the onesweep algorithm in the dispatcher
- [x] implement `policy_selector_from_hub`
- [x] Drop `_policy`
- [x] Merge before: #7103 
- [x] Merge before: #7282
- [x] tests for supplying a custom policy hub
- [x] No SASS changes for `cub.bench.radix_sort.keys.base` for SMs `70;80;90;100`
- [x] No SASS changes for `cub.bench.radix_sort.pairs.base` for SMs `70;80;90;100`

I cannot SASS check benchmarks for < SM70, because `nvbench_helper.cu` does not compile there with:
```
/home/coder/cccl/lib/cmake/cub/../../../cub/cub/agent/agent_find.cuh(177): error #20305-D: atomic operations' argument of memory order is supported on architecture sm_70 or above. Fall back to use membar.
          __nv_atomic_load(found_pos_ptr, &temp_storage.global_result, __NV_ATOMIC_RELAXED, __NV_THREAD_SCOPE_DEVICE);
                           ^
```
But it's somewhat ok, since we don't officially support <SM75 anyway

Fixes: #6676

PR #7193, opened 2026-01-13
Refactor some bits of `DeviceRadixSort`

bernhardmgruber:
Adds an assertion and simplifies static loop unrolling.

- [x] No SASS changes for `cub.bench.radix_sort.keys.base` on sm120

### DIFF
--- a/c/parallel/src/radix_sort.cu
+++ b/c/parallel/src/radix_sort.cu
@@ -378,7 +378,7 @@ __device__ consteval auto& policy_generator() {{
     "-rdc=true",
     "-dlto",
     "-default-device",
-      "-DCUB_DISABLE_CDP",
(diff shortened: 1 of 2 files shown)


cub/cub/agent/agent_radix_sort_upsweep.cuh
```diff
@@ -258,6 +230,7 @@ struct AgentRadixSortUpsweep
 
     // Get row offset
     uint32_t row_offset = digit >> LOG_PACKING_RATIO;
+    _CCCL_ASSERT(row_offset < COUNTER_LANES, "");
```
bernhardmgruber, line 233: I hit this a couple of times in #6767, so I decided to keep it.

shwina (approved):

PR #7213, opened 2026-01-14
Run build/test commands with 5h30m timeout on CI.

### LINKED ISSUE #7211: [CI] Use `timeout` wrapper in CI scripts
Our CI scripts should timeout before GHA kills them so we can upload logs, etc.

Probably easiest to just wrap the put like 5h30m timeouts on every ninja / ctest invocation the build_common.sh script, only when in running in github actions.

alliepiper:
Fixes #7211.

### DIFF
--- a/ci/build_common.sh
+++ b/ci/build_common.sh
@@ -177,6 +177,12 @@ export CXX_STANDARD
 
 source ./pretty_printing.sh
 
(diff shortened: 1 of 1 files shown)


alliepiper (commented):
/ok to test

PR #7217, opened 2026-01-14
Try and work around NVHPC issue with `is_metafunction_defined`

miscco:
Obviously the addition of comparison operators  broke NVHPC which is currently disabled in CI

This tries to work around it be reworking `is_thrust_pointer_v`

This allows us to actually just drop `is_metafunction_defined` entirely and rely on a simple inline variable.

I had to adapt some includes that were transitive


bernhardmgruber:
I am happy to see `is_metafunction_defined` go!

Issue #7173, opened 2026-01-12
[RFE]: Provide compile-time utilities to describe operator properties


fbusato:
### Is this a duplicate?

- [x] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cccl/issues) for this request and that I agree to the [Code of Conduct](CODE_OF_CONDUCT.md)

### Area

libcu++

### Is your feature request related to a problem? Please describe.

Operator associativity/commutativity directly determines whether results or CUB/Thrust algorithm are well-defined, reproducible, or deterministic.

Adding compile-time “operator property” utilities would provide a way to validate these assumptions explicitly, simplify dispatch logic, and would even allow users to describe the behavior of custom operators.

### Describe the solution you'd like

- Type traits: `cuda::is_associative_operator_v<Op>`.
- Numeric limits-style: `cuda::operator_property<Op>::associativity()`

bernhardmgruber:
I think this is a good idea! We already have traits that give us the identity, so this is a natural extension.

fbusato:
done in #7240

PR #7218, opened 2026-01-14
Drop `thrust::detail::is_commutative`

miscco:
its unused and implemented inefficiently


miscco:
@gevtushenko I believe we discussed something similar in the review meeting. We cann bring that back in libcu+++ when we want it

bernhardmgruber:
@fbusato opened this issue for operator traits: #7173

miscco:
yeah but `is_commutative` is the wrong trait and its also defined inefficiently in the wrong place

PR #7222, opened 2026-01-14
 Implement `__no_unique_member` and use it in ranges

davebayer:
We are having problems with `[[no_unique_address]]` causing data corruption. This PR tries to provide a solution to this problem by introducing `__no_unique_member` class wrapper that should be used as a base class with combination with `_CCCL_DECLSPEC_EMPTY_BASES`.

I tried to demonstrate how it should be used on some simple ranges.


davebayer:
pre-commit.ci autofix

libcudacxx/include/cuda/std/__utility/no_unique.h
```diff
@@ -0,0 +1,95 @@
...
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_CLANG)
+#  pragma clang system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_MSVC)
+#  pragma system_header
+#endif // no system header
+
+#include <cuda/std/__cstddef/types.h>
+#include <cuda/std/__type_traits/enable_if.h>
+#include <cuda/std/__type_traits/is_class.h>
+#include <cuda/std/__type_traits/is_union.h>
```
miscco: unused

PR #5787, opened 2025-09-05
Refactor docs deployment to support future versioned docs

jrhemstad:
This PR updates the docs build/deploy flow so Sphinx emits versioned output and the version switcher resolves correctly on forks and GitHub Pages.

As a result, the docs built from `main` will be published under https://nvidia.github.io/cccl/unstable and future builds from release branches will be under https://nvidia.github.io/cccl/X.Y.

Key changes
- docs/gen_docs.bash now reorganizes the Sphinx build into a versioned directory, and writes root assets (index.html redirect, nv-versions.json, versions.json, .nojekyll, and objects.inv) driven by SPHINX_CCCL_VER, CCCL_DOCS_BASE_URL, and CCCL_DOCS_IS_LATEST.
- docs/conf.py derives the release version from SPHINX_CCCL_VER (fallbacks to VERSION.md/unstable) and configures the theme switcher to fetch nv-versions.json from html_baseurl, which is set via CCCL_DOCS_BASE_URL.
- docs-deploy.yml computes the base URL from the repo owner/name plus the Pages prefix, derives the deployment destination and latest flag, passes those into the docs build, and deploys under deploy/<publish_prefix> while preserving existing versions with keep_files + force_orphan.

Example versioned docs on my fork: https://jrhemstad.github.io/cccl/unstable/

PR #7229, opened 2026-01-14
docs: Simplify docs workflow build

wence-:
## Description

The workflow build only runs on main, so cull a load of unnecessary steps.

Additionally, adapt the upload action to:

1. Generate the requisite .nojekyll file
2. Force push to the gh-pages branch

The latter will ensure (eventually, once git gc runs on the server) that the CCCL repo doesn't massively bloat with each docs deployment.

Similar to #1194 

## Checklist

### DIFF
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -16,20 +16,6 @@ on:
         required: false
         default: "docs"
         type: string
-      pr_number:
-        description: "PR number for preview URL generation"
-        required: false
-        default: ""
-        type: string
-      remove_pr_preview:
-        description: "If true, deploy empty folder to remove PR preview"
-        required: false
-        default: false
-        type: boolean
-    outputs:
-      preview_url:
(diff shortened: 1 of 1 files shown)


miscco:
We once dreamed of a time where each PR would generate a documentation @jrhemstad what do you say?

wence-:
> We once dreamed of a time where each PR would generate a documentation @jrhemstad what do you say?

In rapids we do this by pushing the built site into an s3 bucket that we expose via html. e.g. see https://github.com/rapidsai/cudf/actions/runs/20975326181/job/60292097495?pr=21032#step:13:7664

That avoids polluting the gh-pages branch with commits for every PR.

The official github actions `deploy-pages` action has a `preview` option (https://github.com/actions/deploy-pages/blob/main/action.yml) but it is github internal only and has been in alpha preview for at least 4 years. (You can see it working e.g. https://github.com/github/opensource.guide/pull/3583)

wence-:
#5787 is a better version of this.

PR #7233, opened 2026-01-14
Avoid deallocate throwing

gonidelis:
tries to fix https://nvbugspro.nvidia.com/bug/5017720

### DIFF
--- a/thrust/thrust/detail/allocator/temporary_allocator.h
+++ b/thrust/thrust/detail/allocator/temporary_allocator.h
@@ -33,6 +33,7 @@
 #include <thrust/memory.h>
 #include <thrust/system/detail/bad_alloc.h>
 
+#include <cuda/std/__exception/exception_macros.h>
 #include <cuda/std/__memory/allocator_traits.h>
 #include <cuda/std/__utility/pair.h>
 #include <cuda/std/cassert>
@@ -99,7 +100,21 @@ public:
 
   _CCCL_HOST_DEVICE void deallocate(pointer p, size_type n) noexcept
   {
-    return thrust::return_temporary_buffer(system(), p, n);
+    _CCCL_TRY
+    {
+      thrust::return_temporary_buffer(system(), p, n);
+    }
+    _CCCL_CATCH_ALL
+    {
+      _CCCL_ASSERT(false, "Exception thrown in deallocate");
+      // Swallow all exceptions to maintain noexcept contract per C++ allocator requirements.
+      // Deallocate must be noexcept to be safe in destructors and during exception unwinding.
+      // Clear CUDA error state and leak the memory rather than propagating exception.
+      // Memory is leaked, but this matches standard allocator behavior when deallocation fails.
+#if _CCCL_CUDA_COMPILATION()
+      NV_IF_TARGET(NV_IS_HOST, cudaGetLastError();)
+#endif // _CCCL_CUDA_COMPILATION()
+    }
   }
 
   _CCCL_HOST_DEVICE inline System& system()


thrust/thrust/system/cuda/detail/temporary_buffer.h
```diff
@@ -78,7 +78,12 @@ _CCCL_HOST void return_temporary_buffer(par_nosync_t&, Pointer ptr, ::cuda::std:
 
     if (status != cudaSuccess)
     {
-      throw system::detail::bad_alloc(cuda_category().message(status).c_str());
+      // Cannot throw from deallocate per C++ allocator requirements.
+      // Deallocate must be noexcept to be safe in destructors and during exception unwinding.
+      // Clear the error state and leak the memory rather than throwing.
+      // This matches standard allocator behavior.
+      cudaGetLastError(); // Clear error
+      return; // Leak rather than throw
```
bernhardmgruber: Important: I don't understand the rationale here. The current functions is not marked `noexcept` so we should be fine throwing here. The function called out in the NVBug is `deallocate`, which is in a different file:
https://github.com/NVIDIA/cccl/blob/91434d76ab3cf2e85fa7f72a70deb844f064c639/thrust/thrust/detail/allocator/temporary_allocator.h#L100

Here, I would try to remove the `noexcept` and see how far we get. `std::allocator` also does not have a `noexcept` on `deallocate`, but I am not an expert on allocators: https://en.cppreference.com/w/cpp/memory/allocator/deallocate.html

gonidelis: i would like to avoid removing `noexcept`. It's just a matter of opinion. If resistance is hard I can just remove the `noexcept`.

I summon @miscco. Enlighten us with your standards wisdomry.

gonidelis:
For the record this is what I have been working upon https://eel.is/c++draft/allocator.traits.members#5

bernhardmgruber:
> For the record this is what I have been working upon https://eel.is/c++draft/allocator.traits.members#5

The `Throws: Nothing` clause just tells you that it shouldn't through an exception, not that it should be `noexcept`. The distinction is often made because a Throws: Nothing` API may have a pre condition that a contract check may eventually turn into an exception with a corresponding contract violation handler.

In any case, if we want to avoid `deallocate` to throw an exception, we can also just catch it ourselves from the underlying temporary buffer APIs, log it, and throw it away.

But this is allocator design now and I am not experienced enough on the matter. @griwes and @miscco may help here.

thrust/thrust/detail/allocator/temporary_allocator.h
```diff
@@ -99,7 +99,35 @@ class temporary_allocator : public thrust::detail::tagged_allocator<T, System, t
...
+    {
+      thrust::return_temporary_buffer(system(), p, n);
+    }
+    catch (...)
+    {
+      // Swallow all exceptions to maintain noexcept contract per C++ allocator requirements.
+      // Deallocate must be noexcept to be safe in destructors and during exception unwinding.
+      // Memory is leaked, but this matches standard allocator behavior when deallocation fails.
+    }
+#endif
```
miscco: This is not the right approach. We have macro functionality for exactly that
```suggestion
    _CCCL_TRY 
    {
      thrust::return_temporary_buffer(system(), p, n);
    }
    _CCCL_CATCH_ALL 
    {
      // Swallow all exceptions to maintain noexcept contract per C++ allocator requirements.
      // Deallocate must be noexcept to be safe in destructors and during exception unwinding.
      // Memory is leaked, but this matches standard allocator behavior when deallocation fails.
      NV_IF_TARGET(NV_IS_HOST, cudaGetLastError();))
    }
```

gonidelis: so this *was* the right approach, just not the right implementation. Question was whether it was right that we keep the `noexcept` at the expense of more sophisticated branch code. I will push your changes and we can merge.

gonidelis: I love that we have a thing for that thing btw

gonidelis:
@miscco  @bernhardmgruber please approve once green

thrust/thrust/detail/allocator/temporary_allocator.h
```diff
@@ -99,7 +100,20 @@ class temporary_allocator : public thrust::detail::tagged_allocator<T, System, t
...
+      thrust::return_temporary_buffer(system(), p, n);
+    }
+    _CCCL_CATCH_ALL
+    {
+      // Swallow all exceptions to maintain noexcept contract per C++ allocator requirements.
+      // Deallocate must be noexcept to be safe in destructors and during exception unwinding.
+      // Clear CUDA error state and leak the memory rather than propagating exception.
+      // Memory is leaked, but this matches standard allocator behavior when deallocation fails.
+#if defined(__CUDACC__)
+      NV_IF_TARGET(NV_IS_HOST, (cudaGetLastError();), ())
```
bernhardmgruber: Remark: could be shortened:
```suggestion
      NV_IF_TARGET(NV_IS_HOST, cudaGetLastError();)
```

bernhardmgruber (approved):

thrust/thrust/detail/allocator/temporary_allocator.h
```diff
@@ -99,7 +100,20 @@ class temporary_allocator : public thrust::detail::tagged_allocator<T, System, t
...
+    {
+      thrust::return_temporary_buffer(system(), p, n);
+    }
+    _CCCL_CATCH_ALL
+    {
+      // Swallow all exceptions to maintain noexcept contract per C++ allocator requirements.
+      // Deallocate must be noexcept to be safe in destructors and during exception unwinding.
+      // Clear CUDA error state and leak the memory rather than propagating exception.
+      // Memory is leaked, but this matches standard allocator behavior when deallocation fails.
+#if defined(__CUDACC__)
```
miscco:
```suggestion
#if _CCCL_CUDA_COMPILATION()
```

thrust/thrust/detail/allocator/temporary_allocator.h
```diff
@@ -99,7 +100,20 @@ class temporary_allocator : public thrust::detail::tagged_allocator<T, System, t
...
+      NV_IF_TARGET(NV_IS_HOST, cudaGetLastError();)
+#endif
```
miscco:
```suggestion
#endif // _CCCL_CUDA_COMPILATION()
```

thrust/thrust/detail/allocator/temporary_allocator.h
```diff
@@ -99,7 +100,20 @@ class temporary_allocator : public thrust::detail::tagged_allocator<T, System, t
...
   _CCCL_HOST_DEVICE void deallocate(pointer p, size_type n) noexcept
   {
-    return thrust::return_temporary_buffer(system(), p, n);
+    _CCCL_TRY
+    {
+      thrust::return_temporary_buffer(system(), p, n);
+    }
+    _CCCL_CATCH_ALL
+    {
+      // Swallow all exceptions to maintain noexcept contract per C++ allocator requirements.
```
miscco, line 110: I am wondering whether we want to add a 
```suggestion
      _CCCL_ASSERT(false, "Exception thrown in deallocate");
      // Swallow all exceptions to maintain noexcept contract per C++ allocator requirements.
```

bernhardmgruber: That's a good idea!

gonidelis: good point. swallow on release, throw on debug mode. sits better with me.

PR #7239, opened 2026-01-14
Fix extracting CUDA stream in `cub::DeviceTransform`

bernhardmgruber:
#6204 changed `cub::DeviceTransform` APIs from taking `cudaStream_t` to environments. Special handling preserved support for `cudaStream_t`. However, user-provided stream types with conversion operators to `cudaStream_t` were now queried as environments, failing to return a stream.

This PR treats any type that is convertible to a `cudaStream_t` not as an environment and extracts the underlying stream.

Fixes NVBug 5813928

### DIFF
--- a/cub/cub/device/device_transform.cuh
+++ b/cub/cub/device/device_transform.cuh
@@ -109,7 +109,7 @@ private:
   }
 
   template <typename Env>
-  CUB_RUNTIME_FUNCTION static auto get_stream(Env env) -> cudaStream_t
+  CUB_RUNTIME_FUNCTION static auto get_stream([[maybe_unused]] Env env) -> cudaStream_t
   {
     if constexpr (::cuda::std::is_invocable_v<::cuda::get_stream_t, Env>)
     {


cub/cub/device/device_transform.cuh
```diff
@@ -489,7 +495,7 @@ public:
       num_items,
       ::cuda::std::move(predicate),
       ::cuda::std::move(transform_op),
-      get_stream(env));
+      ::cuda::std::move(env));
```
bernhardmgruber, line 516: Drive-by fix

cub/cub/device/device_transform.cuh
```diff
@@ -633,7 +639,7 @@ public:
       ::cuda::std::move(output),
       num_items,
       ::cuda::std::move(transform_op),
-      get_stream(env));
+      ::cuda::std::move(env));
```
bernhardmgruber, line 659: Drive-by fix

cub/test/catch2_test_device_transform_env.cu
```diff
@@ -31,14 +41,20 @@ C2H_TEST("DeviceTransform::Transform custom stream", "[device][transform]")
...
+    cub::DeviceTransform::Transform(cuda::std::make_tuple(a, b), result.begin(), num_items, _1 + _2, streamish);
+  };
   SECTION("raw stream")
   {
-    cub::DeviceTransform::Transform(cuda::std::make_tuple(a, b), result.begin(), num_items, _1 + _2, stream);
+    run(stream);
+  }
+  SECTION("custom stream")
+  {
+    run(custom_stream{stream});
```
bernhardmgruber, line 53: I verified that the stream is extracted in the debugger, but I wonder if I could write the unit test in a way to detect if the default stream was taken anywhere. Does anybody know if I can query the `stream` whether something was really enqueued there?

pciolkosz: One solution would be to start a graph capture on a stream and see if anything was captured, but that might have some limitations, not sure if its applicable here

cub/cub/device/device_transform.cuh
```diff
@@ -89,12 +89,18 @@ private:
...
+    {
+      return env;
+    }
+    else if constexpr (::cuda::std::is_convertible_v<Env, cudaStream_t>)
+    {
+      return static_cast<cudaStream_t>(env);
+    }
+    else
+    {
+      return ::cuda::std::execution::__query_or(env, ::cuda::get_stream, ::cuda::stream_ref{cudaStream_t{}}).get();
```
miscco: This is very close to what we have in `get_stream` 

I believe the issue is that we are checking whether tis convertible to `stream_ref` and not `cudaStream_t`

But at the same time `cudaStream_t` should be convertible to `stream_ref`

miscco (commented):
I would love to better understand where the issue with the conversions from `cudaStream_t` lies so that we can extent our `get_stream` CPO

What was the exact case failing? from what I can see it the second and third branch in the if constexpr should already work

Maybe we just need another clause for get_stream

bernhardmgruber:
Discussed this PR with @miscco and we concluded that the environment query for `cuda::get_stream` misses a check whether the environment is convertible to a `cudaStream_t`. He will add this and then we can replace the `get_stream` in this PR by just the environment query.

miscco:
/ok to test [cc8f29e](https://github.com/NVIDIA/cccl/pull/7239/commits/cc8f29eece46209f9267ae9413a230b2827443b1)

miscco:
I have updated our `get_stream` CPO to ensure that it also works with types that are convertible to `::cudaStream_t` as apposed to `::cuda::stream_ref`

bernhardmgruber:
I started the backport already from the current state: #7263, since I need to rebase and change this PR again, since #6914 was merged and we have conflicts.

PR #7240, opened 2026-01-15
Provide Operator Properties

fbusato:
## Description

Address https://github.com/NVIDIA/cccl/issues/7173

The PR provides a standardized way to query various properties of operators. For example, `cuda::std::plus` is associative with `unsigned` but not for `float`.

### DIFF
--- a/.devcontainer/launch.py
+++ b/.devcontainer/launch.py
@@ -86,11 +86,25 @@ def load_devcontainer_json(path):
 
 
 def load_devcontainer_meta(tag):
+    import time
+
     # Always pull the latest copy of the image
     if os.environ.get("GITHUB_ACTIONS", None) is not None:
         print(f"::group::Pulling Docker image {tag}", file=sys.stderr)
 
-    subprocess.run(["docker", "pull", tag], stdout=sys.stderr).check_returncode()
+    attempts = 0
+    while True:
+        try:
+            subprocess.run(
+                ["docker", "pull", tag], stdout=sys.stderr
+            ).check_returncode()
+            break
+        except Exception as e:
+            if attempts < 10:
+                attempts += 1
+                time.sleep(5)
+            else:
+                raise e
 
     if os.environ.get("GITHUB_ACTIONS", None) is not None:
         print("::endgroup::", file=sys.stderr)
--- a/ci/upload_job_result_artifacts.sh
+++ b/ci/upload_job_result_artifacts.sh
@@ -44,16 +44,21 @@ fi
 
 # Finds a matching file in the root and copies it to the artifact directory.
 find_and_copy_job_artifact_from() {
-  filename="$1"
-  root="$2"
-  filepath="$(find "$root" -name "$filename" -print -quit)"
-  if [[ -z "$filepath" ]]; then
-    echo "$filename does not exist in repo directory."
+  root="$1"
+  name="$2"
+  if find "$root"/ -maxdepth 4 -name "$name" -type f -printf '' -quit 2>/dev/null; then
+    find "$root"/ -maxdepth 4 -name "$name" -type f -print0 | xargs -0 -P4 -I% cp -v % "$job_artifacts"/
+  else
+    echo "No file matching '$name' found in '$root'."
     return 1
   fi
-  cp -v "$filepath" "$job_artifacts/"
 }
 
-find_and_copy_job_artifact_from "sccache_stats.json" build/ || : # Nonfatal if not found
+find_and_copy_job_artifact_from /tmp  "sccache*.log"       || : # Nonfatal if not found
+find_and_copy_job_artifact_from build "sccache_stats.json" || : # Nonfatal if not found
+find_and_copy_job_artifact_from build ".ninja_log"         || : # Nonfatal if not found
+find_and_copy_job_artifact_from build "build.ninja"        || : # Nonfatal if not found
+find_and_copy_job_artifact_from build "rules.ninja"        || : # Nonfatal if not found
+find_and_copy_job_artifact_from build "ctest.log"          || : # Nonfatal if not found
 
 ci/util/artifacts/upload/register.sh "zz_jobs-$job_id" "$jobs_artifact_dir"
--- a/cub/cub/block/block_adjacent_difference.cuh
+++ b/cub/cub/block/block_adjacent_difference.cuh
@@ -201,12 +201,23 @@ public:
   //! @{
 
   //! @brief Collective constructor using a private static allocation of shared memory as temporary storage
+  //!
+  //! @rst
+  //! .. versionadded:: 2.2.0
+  //!    First appears in CUDA Toolkit 12.3.
+  //! @endrst
   _CCCL_DEVICE _CCCL_FORCEINLINE BlockAdjacentDifference()
       : temp_storage(PrivateStorage())
       , linear_tid(RowMajorTid(BlockDimX, BlockDimY, BlockDimZ))
   {}
 
   //! @brief Collective constructor using the specified memory allocation as temporary storage
+  //!
+  //! @rst
+  //! .. versionadded:: 2.2.0
+  //!    First appears in CUDA Toolkit 12.3.
+  //! @endrst
+  //!
   //! @param[in] temp_storage Reference to memory allocation having layout type TempStorage
   _CCCL_DEVICE _CCCL_FORCEINLINE BlockAdjacentDifference(TempStorage& temp_storage)
       : temp_storage(temp_storage.Alias())
@@ -219,6 +230,9 @@ public:
 
   //! @rst
   //! Subtracts the left element of each adjacent pair of elements partitioned across a CUDA thread block.
+  //!
+  //! .. versionadded:: 2.2.0
+  //!    First appears in CUDA Toolkit 12.3.
   //!
   //! - @rowmajor
   //! - @smemreuse
@@ -303,6 +317,9 @@ public:
 
   //! @rst
   //! Subtracts the left element of each adjacent pair of elements partitioned across a CUDA thread block.
+  //!
+  //! .. versionadded:: 2.2.0
+  //!    First appears in CUDA Toolkit 12.3.
   //!
   //! - @rowmajor
   //! - @smemreuse
@@ -403,6 +420,9 @@ public:
 
   //! @rst
   //! Subtracts the left element of each adjacent pair of elements partitioned across a CUDA thread block.
+  //!
+  //! .. versionadded:: 2.2.0
+  //!    First appears in CUDA Toolkit 12.3.
   //!
   //! - @rowmajor
   //! - @smemreuse
@@ -513,6 +533,9 @@ public:
 
   //! @rst
   //! Subtracts the left element of each adjacent pair of elements partitioned across a CUDA thread block.
+  //!
+  //! .. versionadded:: 2.2.0
+  //!    First appears in CUDA Toolkit 12.3.
   //!
   //! - @rowmajor
   //! - @smemreuse
@@ -646,6 +669,9 @@ public:
   //!
   //! Subtracts the right element of each adjacent pair of elements partitioned across a CUDA thread block.
   //!
+  //! .. versionadded:: 2.2.0
+  //!    First appears in CUDA Toolkit 12.3.
+  //!
   //! - @rowmajor
   //! - @smemreuse
   //!
@@ -731,6 +757,9 @@ public:
 
   //! @rst
   //! Subtracts the right element of each adjacent pair of elements partitioned across a CUDA thread block.
+  //!
+  //! .. versionadded:: 2.2.0
+  //!    First appears in CUDA Toolkit 12.3.
   //!
   //! - @rowmajor
   //! - @smemreuse
@@ -829,6 +858,9 @@ public:
 
   //! @rst
   //! Subtracts the right element of each adjacent pair in range of elements partitioned across a CUDA thread block.
+  //!
+  //! .. versionadded:: 2.2.0
+  //!    First appears in CUDA Toolkit 12.3.
   //!
   //! - @rowmajor
   //! - @smemreuse
--- a/cub/cub/block/block_discontinuity.cuh
+++ b/cub/cub/block/block_discontinuity.cuh
@@ -221,6 +221,11 @@ public:
   /**
    * @brief Collective constructor using a private static allocation of shared memory as temporary
    *        storage.
+   *
+   * @rst
+   * .. versionadded:: 2.2.0
+   *    First appears in CUDA Toolkit 12.3.
+   * @endrst
    */
   _CCCL_DEVICE _CCCL_FORCEINLINE BlockDiscontinuity()
       : temp_storage(PrivateStorage())
@@ -229,6 +234,11 @@ public:
 
   /**
    * @brief Collective constructor using the specified memory allocation as temporary storage.
+   *
+   * @rst
+   * .. versionadded:: 2.2.0
+   *    First appears in CUDA Toolkit 12.3.
+   * @endrst
    *
    * @param[in] temp_storage
    *   Reference to memory allocation having layout type TempStorage
@@ -327,6 +337,9 @@ public:
   //! Sets head flags indicating discontinuities between items partitioned across the thread
   //! block, for which the first item has no reference and is always flagged.
   //!
+  //! .. versionadded:: 2.2.0
+  //!    First appears in CUDA Toolkit 12.3.
+  //!
   //! - The flag ``head_flags[i]`` is set for item ``input[i]`` when ``flag_op(previous-item, input[i])`` returns
   //!   ``true`` (where ``previous-item`` is either the preceding item in the same thread or the last item in
   //!   the previous thread).
@@ -400,6 +413,9 @@ public:
 
   //! @rst
   //! Sets head flags indicating discontinuities between items partitioned across the thread block.
+  //!
+  //! .. versionadded:: 2.2.0
+  //!    First appears in CUDA Toolkit 12.3.
   //!
   //! - The flag ``head_flags[i]`` is set for item ``input[i]`` when ``flag_op(previous-item, input[i])``
   //!   returns ``true`` (where ``previous-item`` is either the preceding item in the same thread or the last item
@@ -490,6 +506,9 @@ public:
   //! Sets tail flags indicating discontinuities between items partitioned across the thread
   //! block, for which the last item has no reference and is always flagged.
   //!
+  //! .. versionadded:: 2.2.0
+  //!    First appears in CUDA Toolkit 12.3.
+  //!
   //! - The flag ``tail_flags[i]`` is set for item ``input[i]`` when
   //!   ``flag_op(input[i], next-item)``
   //!   returns ``true`` (where `next-item` is either the next item
@@ -578,6 +597,9 @@ public:
 
   //! @rst
   //! Sets tail flags indicating discontinuities between items partitioned across the thread block.
+  //!
+  //! .. versionadded:: 2.2.0
+  //!    First appears in CUDA Toolkit 12.3.
   //!
   //! - The flag ``tail_flags[i]`` is set for item ``input[i]`` when ``flag_op(input[i], next-item)``
   //!   returns ``true`` (where ``next-item`` is either the next item in the same thread or the first item in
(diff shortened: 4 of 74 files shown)


libcudacxx/include/cuda/__functional/operator_properties.h
```diff
@@ -0,0 +1,404 @@
...
+ * cuda::std::multiplies specializations
+ **********************************************************************************************************************/
+
+template <class T>
+struct operator_properties<::cuda::std::multiplies<T>, T>
+{
+  static constexpr bool is_commutative        = true;
+  static constexpr bool is_associative        = ::cuda::std::is_integral_v<T>;
+  static constexpr bool has_identity_element  = true;
+  static constexpr bool has_absorbing_element = true;
```
miscco: This is only true for arithmetic types A user can define what ever

libcudacxx/include/cuda/__functional/operator_properties.h
```diff
@@ -0,0 +1,404 @@
...
+  {
+    return T{};
+  }
+
+  [[nodiscard]] _CCCL_API static constexpr T absorbing_element() noexcept
+  {
+    static_assert(has_absorbing_element, "cuda::std::plus does not have an absorbing element");
+    return T{};
+  }
+};
```
bernhardmgruber: Suggestion: I vaguely remember that `std::numeric_limits` was used as example for a bad design for accumulating multiple traits together, and individual traits would be superior. Unfortunately, I forgot the reason. Maybe @andralex knows it. Maybe it's related to that some traits don't make sense for all types.

Given this  background, I would recommend providing individual traits instead of a single one containing a bunch of subinformation.

fbusato: that's a good point. This is also something that I considered.

fbusato: I see there are critics related to `numeric_limits` [p0437](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0437r1.pdf), [p1370](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1370r1.html)

docs/libcudacxx/extended_api/functional.rst
```diff
@@ -44,3 +45,8 @@ Functional
      - Returns a valid address to a device object
      - CCCL 2.8.0
      - CUDA 12.9
+
+   * - :ref:`cuda::is_associative <libcudacxx-extended-api-functional-operator-properties>`, :ref:`cuda::is_commutative <libcudacxx-extended-api-functional-operator-properties>`, :ref:`cuda::identity_element <libcudacxx-extended-api-functional-operator-properties>`, :ref:`cuda::absorbing_element <libcudacxx-extended-api-functional-operator-properties>`
+     - Determines if an operator is associative for a type
+     - CCCL 3.2.0
+     - CUDA 13.2
```
bernhardmgruber: I assume we won't get the backport in time :)
```suggestion
     - CCCL 3.3.0
     - CUDA 13.3
```

docs/libcudacxx/extended_api/functional/operator_properties.rst
```diff
@@ -0,0 +1,343 @@
...
+   * - ``cuda::minimum``
+     - ``true``
+     - ``true``
+   * - ``cuda::maximum``
+     - ``true``
+     - ``true``
+
+.. note::
+
+   In the strictest sense of the term, the operations of plus and multiplication for integral values may result in undefined behaviour due to overflow. However, in the context of parallel algorithms, they are considered to be associative.
```
bernhardmgruber, line 75: Remark: I am not worried, if `op(op(a, b), c)` for plus/mul and integral types overflows, `op(a, op(b, c))` will overflow as well :) If one is UB, so is the other.

You may consider just removing the sentence.

libcudacxx/include/cuda/__functional/operator_properties.h
```diff
@@ -0,0 +1,613 @@
...
+template <class _Tp>
+struct __is_commutative<::cuda::std::modulus<>, _Tp> : __is_commutative<::cuda::std::modulus<_Tp>, _Tp>
+{};
+
+template <class _Op, class _Tp>
+struct is_commutative : __is_commutative<_Op, _Tp>
+{};
+
+template <class _Op, class _Tp>
+inline constexpr bool is_commutative_v = is_commutative<_Op, _Tp>::value;
```
bernhardmgruber: Suggestion: I think it's more efficient to define all the specializations for the variable template and only define one trait struct via the variable template. If you never use the struct, the compiler saves some compile time and memory, since structs are more expensive than variable templates.

docs/libcudacxx/extended_api/functional.rst
fbusato: you are right. No rush

docs/libcudacxx/extended_api/functional/operator_properties.rst
fbusato: true, but UB could lead to different results -> non-associative.  It is definitely non-critical. I'm also not sure if it is the case to remove the note

libcudacxx/include/cuda/__functional/operator_properties.h
fbusato: that's a good point. Should not be a problem for user-specialization, right?

docs/libcudacxx/extended_api/functional/operator_properties.rst
```diff
@@ -0,0 +1,347 @@
...
+   template <class Op, class T>
+   struct is_associative;
+
+   template <class Op, class T>
+   inline constexpr bool is_associative_v = is_associative<Op, T>::value;
+
+   } // namespace cuda
+
+Determines whether a binary operator ``Op`` is associative for type ``T``, meaning ``op(op(a, b), c) == op(a, op(b, c))``
+for all values ``a``, ``b``, ``c`` of type ``T``.
```
miscco:
```suggestion
for all values ``a``, ``b``, ``c`` of type ``T``. This allows the implementation to reorder operations.
```

docs/libcudacxx/extended_api/functional/operator_properties.rst
```diff
@@ -0,0 +1,347 @@
...
+A helper trait that evaluates to ``true`` if the given trait template ``Trait`` instantiated with types ``Tp...``
+has a valid ``value`` member.
+
+.. code:: cuda
+
+   // Check if identity_element is defined for plus<int> and int
+   static_assert(cuda::element_exists<cuda::identity_element, cuda::std::plus<int>, int>);
+
+   // Check if absorbing_element is defined for plus<int> and int (it's not)
+   static_assert(!cuda::element_exists<cuda::absorbing_element, cuda::std::plus<int>, int>);
```
miscco: I find this overly complex, and would rather just have `has_identity_element `

docs/libcudacxx/extended_api/functional/operator_properties.rst
```diff
@@ -0,0 +1,347 @@
...
+   namespace cuda {
+
+   template <class Op, class T>
+   struct identity_element;
+
+   template <class Op, class T>
+   constexpr auto get_identity_element() noexcept;
+
+   template <class Op, class T>
+   inline constexpr bool has_identity_element = /* see below */;
```
miscco: In the document we suffix inline variables with `_v`
```suggestion
   inline constexpr bool has_identity_element_v = /* see below */;
```

docs/libcudacxx/extended_api/functional/operator_properties.rst
```diff
@@ -0,0 +1,347 @@
...
+
+Identity Element
+----------------
+
+.. code:: cuda
+
+   namespace cuda {
+
+   template <class Op, class T>
+   struct identity_element;
```
miscco: Mathematically this is called the 
```suggestion
   struct neutral_element;
```

docs/libcudacxx/extended_api/functional/operator_properties.rst
```diff
@@ -0,0 +1,347 @@
...
+
+Commutativity
+-------------
+
+.. code:: cuda
+
+   namespace cuda {
+
+   template <class Op, class T>
+   struct is_commutative;
```
miscco: We should not provide struct of those types. Those are inefficient and make specializing the traits cumbersome

libcudacxx/include/cuda/__functional/operator_properties.h
```diff
@@ -0,0 +1,609 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef _CUDA_FUNCTIONAL_OPERATOR_PROPERTIES_H
+#define _CUDA_FUNCTIONAL_OPERATOR_PROPERTIES_H
+
+#include <cuda/std/detail/__config>
+
+#include <cuda/std/__cccl/execution_space.h>
+
```
miscco:
```suggestion
```

libcudacxx/include/cuda/__functional/operator_properties.h
```diff
@@ -0,0 +1,609 @@
...
+template <class _Tp>
+inline constexpr bool __is_associative_v<::cuda::std::modulus<>, _Tp> =
+  __is_associative_v<::cuda::std::modulus<_Tp>, _Tp, void>;
+
+template <class _Op, class _Tp>
+inline constexpr bool is_associative_v = __is_associative_v<_Op, _Tp>;
+
+template <class _Op, class _Tp>
+struct is_associative : ::cuda::std::bool_constant<is_associative_v<_Op, _Tp>>
+{};
```
miscco: We should just define `is_associative_v ` instead of `__is_associative_v ` and drop the struct

A user specializing the struct would be a bug because we would never realize it

Applies throughout

libcudacxx/include/cuda/__functional/operator_properties.h
```diff
@@ -0,0 +1,609 @@
...
+ **********************************************************************************************************************/
+
+template <template <class...> class _Trait, class... _Tp>
+_CCCL_HOST_DEVICE auto __element_exists_test(int) -> decltype(_Trait<_Tp...>::value, ::cuda::std::true_type{});
+
+template <template <class...> class _Trait, class... _Tp>
+_CCCL_HOST_DEVICE auto __element_exists_test(...) -> ::cuda::std::false_type;
+
+template <template <class...> class _Trait, class... _Tp>
+inline constexpr bool element_exists = decltype(__element_exists_test<_Trait, _Tp...>(0))::value;
```
miscco: We should just drop this. A user can use `has_identity_element_v` and friends

docs/libcudacxx/extended_api/functional/operator_properties.rst
fbusato: `identity` is more common in English https://en.wikipedia.org/wiki/Identity_element

libcudacxx/include/cuda/__functional/operator_properties.h
fbusato: I followed the proposal [P0437](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2016/p0437r0.pdf)

libcudacxx/include/cuda/__functional/operator_properties.h
fbusato: why? partial and full specialization have higher priority over fully generic structure

docs/libcudacxx/extended_api/functional/operator_properties.rst
fbusato: see the proposal https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2016/p0437r0.pdf

libcudacxx/include/cuda/__functional/operator_properties.h
fbusato: we need a structure for several reasons:

- Follows  C++ standard library convention.
- Can be used with `std::conjunction` or `std::negation`.
- Tag dispatching (caller function).
- Loses `integral_constant` members, e.g. `operator()` to make the object callable.

libcudacxx/include/cuda/__functional/operator_properties.h
miscco: we want to only use inline variables in product code because those are much more efficient than the types.

However, what happens if a user specializes the struct? We would not pick that up.

The same is true for the opposite, we would have to always use the slow struct and forbid users to specialize the inline variable.

So then the point is why do we even have a struct when its just a C++11 relict

libcudacxx/include/cuda/__functional/operator_properties.h
miscco: 1. is a bad precedent, because the standard only has the struct due to C++11 legacy
2. This is a non issue, we can either use concepts or just wrap in an integral constant. Also the only situation where one would need conjunction is when short-circuiting is necessary for the evaluation of a followup condition which seems unlikely for these traits which are purely optimization hints
3. We never want to use tag dispatch nowadays, `if constexpr` is strictly better
4. why would we want to call it? its a compile-time optimization trait

docs/libcudacxx/extended_api/functional/operator_properties.rst
miscco: fine by me

docs/libcudacxx/extended_api/functional/operator_properties.rst
miscco: The proposal predates C++17 so I would give that as a reason why it did not use inline variables

libcudacxx/include/cuda/__functional/operator_properties.h
fbusato: makes sense. you convinced me 😊

fbusato:
@miscco I refactored the code. I don't think we can get rid of structure for `absorbing_element` and `identity_element` due to limitations of cudafe to handle types like `__half`

bernhardmgruber:
> Please merge this new functionality with the existing cub::detail::identity_v

Ping. Please replace `cub::detail::identity_v` with the new facility.

fbusato:
> Please replace cub::detail::identity_v with the new facility.

I would prefer to do the task in another PR, but I can try here. Please note that `identity_v` also handle vector types for SIMD operations, which are out of scope.

fbusato:
@bernhardmgruber done. everything works

libcudacxx/include/cuda/__functional/operator_properties.h
```diff
@@ -0,0 +1,572 @@
...
+inline constexpr bool
+  has_identity_element_v<_Op, _Tp, ::cuda::std::void_t<decltype(identity_element<_Op, _Tp>::value)>> = true;
+
+/***********************************************************************************************************************
+ * Absorbing Element
+ **********************************************************************************************************************/
+
+template <class _Op, class _Tp, class Enable = void>
+struct absorbing_element
+{};
```
bernhardmgruber: Strong suggestion: I think we should really aim to make the APIs for absorbing and identity element variable templates. The API would look and feel just a lot better if I could write `cuda::identity_element<cuda::maximum, int>`. Then the query and the template to specialize for any user provided type would also be the same entity.

For the non-existing case, you could just add a marker type:
```c++
struct __no_element{};

template <class _Op, class _Tp, class Enable = void>
inline constexpr auto absorbing_element = __no_element;

... // all your specializations

template <class _Op, class _Tp>
inline constexpr bool has_absorbing_element_v = !::cuda::std::is_same_v<absorbing_element<_Op, _Tp>, __no_element>;
```

cub/test/thread_reduce/catch2_test_thread_reduce_partial.cu
```diff
@@ -197,7 +198,7 @@ C2H_TEST("ThreadReduce Narrow PrecisionType Tests",
   constexpr int num_items      = c2h::get<2, TestType>::value;
   using dist_param             = dist_interval<value_t, op_t, num_items>;
   constexpr auto reduce_op     = unwrap_op(std::true_type{}, op_t{});
-  const auto operator_identity = cub_operator_to_identity<accum_t, op_t>::value();
+  const auto operator_identity = identity_v<op_t, accum_t>;
```
miscco, line 201: Shouldnt this be
```suggestion
  const auto operator_identity =cuda::get_identity_element<op_t, accum_t>;
```

cub/test/catch2_test_thread_scan_exclusive_partial.cu
```diff
@@ -274,7 +275,7 @@ C2H_TEST("ThreadScanExclusive Narrow PrecisionType Tests",
   using dist_param             = dist_interval<value_t, op_t, num_items, accum_t, output_t>;
   using filler_dist_param      = dist_interval<accum_t, op_t, num_items, accum_t, output_t>;
   constexpr auto scan_op       = unwrap_op(std::true_type{}, op_t{});
-  const auto operator_identity = cub_operator_to_identity<accum_t, op_t>::value();
+  const auto operator_identity = identity_v<op_t, accum_t>;
```
miscco, line 278: Ditto why isnt this `cuda::get_identity_element<op_t, accum_t>`

libcudacxx/include/cuda/__functional/operator_properties.h
miscco: I would like to second that.

We already use a constexpr function for `is_associative` so we should do the same, that would always leave the inline variable defined and we can provide a clear error message

fbusato: I would prefer `constexpr` template variable too. The problem is that they are not supported on Windows for all types that have conversion operator, like `__half`. Please see the internal discussion with the compiler team.

cub/test/thread_reduce/catch2_test_thread_reduce_partial.cu
fbusato: no, `identity_v` also handles vector types

cub/test/catch2_test_thread_scan_exclusive_partial.cu
fbusato: see above

libcudacxx/include/cuda/__functional/operator_properties.h
fbusato: I tried a middle ground. Use `inline constexpr` for all traits, except for `__half` and `__nv_bfloat16`

libcudacxx/include/cuda/__functional/operator_properties.h
```diff
@@ -0,0 +1,715 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef _CUDA_FUNCTIONAL_OPERATOR_PROPERTIES_H
+#define _CUDA_FUNCTIONAL_OPERATOR_PROPERTIES_H
+
+#include <cuda/std/detail/__config>
+
+#include <cuda/std/__type_traits/is_extended_floating_point.h>
+
```
miscco: Please move to the right position
```suggestion
```

libcudacxx/include/cuda/__functional/operator_properties.h
```diff
@@ -0,0 +1,715 @@
...
+#include <cuda/__functional/maximum.h>
+#include <cuda/__functional/minimum.h>
+#include <cuda/__type_traits/is_floating_point.h>
+#include <cuda/std/__floating_point/arithmetic.h>
+#include <cuda/std/__floating_point/constants.h>
+#include <cuda/std/__functional/operations.h>
+#include <cuda/std/__limits/numeric_limits.h>
+#include <cuda/std/__type_traits/always_false.h>
+#include <cuda/std/__type_traits/enable_if.h>
+#include <cuda/std/__type_traits/is_integer.h>
```
miscco, line 34:
```suggestion
#include <cuda/std/__type_traits/is_extended_floating_point.h>
#include <cuda/std/__type_traits/is_integer.h>
```

libcudacxx/include/cuda/__functional/operator_properties.h
```diff
@@ -0,0 +1,715 @@
...
+                        _Tp,
+                        ::cuda::std::enable_if_t<::cuda::std::__is_cv_extended_floating_point_v<_Tp>>> =
+    ::cuda::std::__fp_inf<_Tp>();
+
+template <class _Tp>
+_CCCL_GLOBAL_CONSTANT auto
+  __absorbing_element_v<::cuda::maximum<>,
+                        _Tp,
+                        ::cuda::std::enable_if_t<::cuda::std::__is_cv_extended_floating_point_v<_Tp>>> =
+    __absorbing_element_v<::cuda::maximum<_Tp>, _Tp>;
```
miscco: Looking at this more, I am strongly considering to move the values to functions 

That would make sure that they are available even for `__half` and friends and also allow us to avoid storing them somewhere

bernhardmgruber: Remark: As a user, I don't care how it's implemented, but I would like to spell the public API as a variable template: `cuda::absorbing_element_v<Op, T>`.

fbusato: we all agree that variable template is the best solution but it doesn't work in multiple cases, for example `__half`.
I'm also considering `contexpr` functions because it solves all issues

PR #7241, opened 2026-01-15
Add a test for <nv/target> to validate old dialect support.

### LINKED ISSUE #7196: [FEA]:  [CI] Add tests for nv/target in pre-C++17 dialects
### Is this a duplicate?

### Area

Infrastructure

`nv/target` has different requirements from the rest of libcudacxx as it is used in CTK headers that support pre-C++17 dialects. We currently do not have regression tests for these configurations.

Add a simple test to libcudacxx that just includes <nv/target> and does a simple `NV_IF_TARGET` dispatch. Add build targets for all supported CTK dialects. Ensure that these pass on all CCCL supported compilers.

_No response_

### Additional context

_No response_

wmaxey:
## Description

closes #7196

## Checklist


libcudacxx/test/nvtarget/CMakeLists.txt
```diff
@@ -0,0 +1,59 @@
...
+      CXX_STANDARD "${dialect}"
+      CXX_STANDARD_REQUIRED ON
+  )
+  set_target_properties(
+    "nvtarget_cuda${dialect}"
+    PROPERTIES
+      COMPILE_OPTIONS "${headertest_warning_levels_device}"
+      CUDA_STANDARD "${dialect}"
+      CUDA_STANDARD_REQUIRED ON
+  )
```
davebayer: Wouldn't this be a better solution:
```cmake
  if ("cxx_std_${dialect}" IN_LIST CMAKE_CXX_COMPILE_FEATURES)
    add_library("nvtarget_cpp${dialect}" OBJECT nvtargettest.cpp)
    target_compile_features("nvtarget_cpp${dialect}" PRIVATE "cxx_std_${dialect}")
  endif()

  if ("cuda_std_${dialect}" IN_LIST CMAKE_CUDA_COMPILE_FEATURES)
    add_library("nvtarget_cuda${dialect}" OBJECT nvtargettest.cu)
    target_compile_features("nvtarget_cuda${dialect}" PRIVATE "cuda_std_${dialect}")
  endif()
```
It will check whether the compiler even supports the dialect first

libcudacxx/test/nvtarget/CMakeLists.txt
```diff
@@ -0,0 +1,59 @@
...
+else()
+  set(
+    nvtarget_dialects
+    11
+    14
+    17
+    20
+  )
+endif()
```
miscco: I believe we want to run this in all configurations and C++20 is only available in some, so we should make sure this is supported

davebayer: This would be solved by checking the available compiler features

libcudacxx/test/nvtarget/CMakeLists.txt
wmaxey: Yeah, I'll swap it out for that, I'm operating off the assumption that we don't have to deal with incompatible compilers by now.

wmaxey: `CMAKE_CXX_STANDARD` overrides `target_compile_features`. I can't clobber it at a directory level, so I had to keep the `set_target_properties` bit, but overall used the structure you described here.

libcudacxx/test/nvtarget/CMakeLists.txt
```diff
@@ -0,0 +1,40 @@
...
+    add_library("nvtarget_cuda${dialect}" OBJECT nvtargettest.cu)
+    # This is required to override CMAKE_CXX_STANDARD - target_compile_features is insufficient.
+    set_target_properties(
+      "nvtarget_cuda${dialect}" PROPERTIES
+      CUDA_STANDARD "${dialect}"
+      CUDA_STANDARD_REQUIRED ON
+    )
+  endif()
+
+  add_dependencies(libcudacxx.test.nvtarget "nvtarget_cpp${dialect}" "nvtarget_cuda${dialect}")
```
davebayer: Nitpick: if `CMAKE_CXX_COMPILE_FEATURES` and `CMAKE_CUDA_COMPILE_FEATURES` don't match, one of these targets would be missing

libcudacxx/test/nvtarget/nvtargettest.cu
```diff
@@ -0,0 +1,6 @@
+#include <nv/target>
+
+int main()
+{
+  return 0;
+}
```
davebayer, line 6: Suggestion: we could get rid of one of the files by just keeping the `.cpp` file and setting the `LANGUAGE` property of the target

libcudacxx/test/nvtarget/CMakeLists.txt
davebayer: Oh, I didn't know that, that explains a lot of things that happened to me in the past :D and it makes the compile features pretty useless then

libcudacxx/test/nvtarget/CMakeLists.txt
```diff
@@ -0,0 +1,40 @@
...
+    add_library("nvtarget_cpp${dialect}" OBJECT nvtargettest.cpp)
+    # This is required to override CMAKE_CXX_STANDARD - target_compile_features is insufficient.
+    set_target_properties(
+      "nvtarget_cpp${dialect}" PROPERTIES
+      CXX_STANDARD "${dialect}"
+      CXX_STANDARD_REQUIRED ON
+    )
+  endif()
+
+  if ("cuda_std_${dialect}" IN_LIST CMAKE_CUDA_COMPILE_FEATURES)
```
davebayer: It seems that `CMAKE_CUDA_COMPILE_FEATURES` doesn't reflect the standard version supported by the host compiler.. We will have to add the CUDA target only if we can add the C++ target for that standard version

libcudacxx/test/nvtarget/nvtargettest.cu
wmaxey: I don't think the property of the target would work, it would be a source file property as a single target can make use of multiple of languages. 

If we change the source file's property I'm not sure what the effects would be. It might bleed into the other targets.

libcudacxx/test/nvtarget/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
...
+  11
+  14
+  17
+  20
+)
+
+set_directory_properties(
+  PROPERTIES
+  INCLUDE_DIRECTORIES "${libcudacxx_SOURCE_DIR}/include"
+)
```
alliepiper, line 7: Heads up that this API is a bit of a footgun and will clobber any existing directories rather than append. Prefer `set_property(DIRECTORY APPEND PROPERTY ...)` or `include_directories(...)`.

For consistency with the rest of CCCL, can we move this to a target property?

```suggestion
```

libcudacxx/test/nvtarget/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
...
+
+foreach(dialect ${nvtarget_dialects})
+  if ("cxx_std_${dialect}" IN_LIST CMAKE_CXX_COMPILE_FEATURES)
+    add_library("nvtarget_cpp${dialect}" OBJECT nvtargettest.cpp)
+    # This is required to override CMAKE_CXX_STANDARD - target_compile_features is insufficient.
+    set_target_properties(
+      "nvtarget_cpp${dialect}" PROPERTIES
+      CXX_STANDARD "${dialect}"
+      CXX_STANDARD_REQUIRED ON
+    )
```
alliepiper, line 26:
```suggestion
    )
    target_include_directories(nvtarget_cpp${dialect} PRIVATE "${libcudacxx_SOURCE_DIR}/include")
```

libcudacxx/test/nvtarget/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
...
+  endif()
+
+  if ("cuda_std_${dialect}" IN_LIST CMAKE_CUDA_COMPILE_FEATURES AND "cxx_std_${dialect}" IN_LIST CMAKE_CXX_COMPILE_FEATURES)
+    add_library("nvtarget_cuda${dialect}" OBJECT nvtargettest.cu)
+    # This is required to override CMAKE_CXX_STANDARD - target_compile_features is insufficient.
+    set_target_properties(
+      "nvtarget_cuda${dialect}" PROPERTIES
+      CUDA_STANDARD "${dialect}"
+      CUDA_STANDARD_REQUIRED ON
+    )
```
alliepiper, line 46:
```suggestion
    )
    target_include_directories(nvtarget_cuda${dialect} PRIVATE "${libcudacxx_SOURCE_DIR}/include")
```

libcudacxx/test/nvtarget/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
...
+set_directory_properties(
+  PROPERTIES
+  INCLUDE_DIRECTORIES "${libcudacxx_SOURCE_DIR}/include"
+)
+
+add_custom_target(libcudacxx.test.nvtarget)
+
+foreach(dialect ${nvtarget_dialects})
+  if ("cxx_std_${dialect}" IN_LIST CMAKE_CXX_COMPILE_FEATURES)
+    add_library("nvtarget_cpp${dialect}" OBJECT nvtargettest.cpp)
```
alliepiper: nit: Consider naming the object lib targets `libcudacxx.test.nvtarget.<LANG><STD>` so they appear nested under the parent in the target lists.

libcudacxx/test/nvtarget/CMakeLists.txt
wmaxey: I was kind of hoping for the clobbering behavior.

wmaxey:
pre-commit.ci autofix

wmaxey:
/ok to test [b46f4a1](https://github.com/NVIDIA/cccl/pull/7241/commits/b46f4a17ab326080a8836fdc8ccec60a5ee24ac9)

PR #7245, opened 2026-01-15
Fix `is_address_from` for `cluster_shared` for pre-sm_90

### LINKED ISSUE #7243: `::cuda::device::is_object_from(..., ::cuda::device::address_space::cluster_shared)`may be wrong on < SM90
Any address in the shared state space is also part of the cluster shared state space. That is, the cluster shared state space is a superset of the shared state space.

Consider the following program:
```c++
#include <cuda/memory>

__global__ void k() {
    __shared__ int i;
    _CCCL_VERIFY(::cuda::device::is_object_from(i, ::cuda::device::address_space::cluster_shared), "");
}

int main() {
    k<<<1, 1>>>();
}
```
This runs successfully on SM90+, but will fail below SM90, because `::cuda::device::is_object_from(..., ::cuda::device::address_space::cluster_shared)` just returns `false` below SM90. I believe this limits the usability of this function and we should just downgrade the check from `cluster_shared` to `shared` below SM90.

davebayer:
Fixes #7243.


libcudacxx/test/libcudacxx/cuda/memory/address_space.pass.cpp
```diff
@@ -33,14 +33,16 @@ __global__ void test_kernel(const _CCCL_GRID_CONSTANT MyStruct grid_constant_var
     assert(is_address_from(&constant_var, address_space::constant));
     assert(is_address_from(&local_var, address_space::local));
     assert(is_address_from(&grid_constant_var, address_space::grid_constant) == _CCCL_HAS_GRID_CONSTANT());
-    // todo: test address_space::cluster_shared
+    assert(is_address_from(&shared_var, address_space::cluster_shared));
+    // todo: test address_space::cluster_shared for __shared__ in different different block
```
bernhardmgruber:
```suggestion
    // todo: test address_space::cluster_shared for __shared__ in different block
```
Applies multiple times

miscco (commented):
We might want to add that its pre-SM90

libcudacxx/test/libcudacxx/cuda/memory/address_space.pass.cpp
davebayer: I wanted a very different block :)

fbusato (approved):

Issue #6688, opened 2025-11-19
Provide `lower_bound` in `cuda.compute`


shwina:
Specifically, I need vectorized lower bound: Specifically, I need vectorized lower bound: https://nvidia.github.io/cccl/thrust/api/group__binary__search_1gad2ac5a1c5091366144d6fdbf2ff5551e.html#lower-bound-exec-first-last-values-first-values-last-result

griwes:
Binary search algorithms for c.parallel are now up for review: #7007

griwes:
#7007 has now been merged.

shwina:
Thanks! I'll get on Python exposure.

PR #7250, opened 2026-01-15
Add cuda.compute APIs for `upper_bound` and `lower_bound`

shwina:
## Description

Closes https://github.com/NVIDIA/cccl/issues/6688.

This Python part of this PR is pretty standard. However, on the C side, I found the need to extend the binary search APIs to return indices rather than pointers. The `cub::DeviceFind::Lower/UpperBound` APIs follow C++ standard library semantics, returning an iterator representing the lower/upper bound positions, rather than indices. On the Python side, we want indices instead, similar to the vectorized versions `thrust::lower_bound` and `thrust::upper_bound`.

I couldn't make a `TransformOutputIterator` to make this work. I think this is because of an assumption that the output is always a pointer (not an iterator) [here](https://github.com/NVIDIA/cccl/blob/90a30e28e4dcd8a26a61a92f9e6a9f49ce1d864a/c/parallel/src/binary_search.cu#L167).

## Checklist


python/cuda_cccl/cuda/compute/algorithms/_binary_search.py
```diff
@@ -0,0 +1,269 @@
...
+        stream=None,
+    ):
+        if protocols.get_data_pointer(d_data) != self.data_ptr:
+            raise ValueError(
+                "d_data must match the array used to construct the binary search object."
+            )
+        if protocols.get_data_pointer(d_out) != self.out_ptr:
+            raise ValueError(
+                "d_out must match the array used to construct the binary search object."
+            )
```
NaderAlAwar: Question: why does it have to match the array? This is not a requirement for other algorithms.

shwina: Right you are.

griwes:
See #7320 for a proper fix to the binary search return type issue.

PR #7252, opened 2026-01-15
[STF] Make data place more extensible 

caugonnet:
Make data place more extensible by introducing an "extension" class to implement custom places. This avoids numerous ugly dispatch based on types, and makes it possible to design standalone places.

To improve modularity, we also provide helpers to allocate data on a data place, even if we are not using tasks.

We also support making a composite allocation from different data places (a VMM large reservation with physical allocations on the different places) instead of just plain devices.

## Description

closes 

## Checklist

### DIFF
--- a/cudax/test/stf/places/affinity_gc.cu
+++ b/cudax/test/stf/places/affinity_gc.cu
@@ -66,7 +66,7 @@ int main()
 #else // ^^^ _CCCL_CTK_BELOW(12, 4) ^^^ / vvv _CCCL_CTK_AT_LEAST(12, 4) vvv
   // Test both data place modes
   test_green_ctx_affinity(false); // Device data place (default)
-  test_green_ctx_affinity(true);  // Green context data place extension
+  test_green_ctx_affinity(true); // Green context data place extension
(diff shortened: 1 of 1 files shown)


cudax/include/cuda/experimental/__stf/places/exec/green_context.cuh
```diff
@@ -179,9 +334,9 @@ public:
   class impl : public exec_place::impl
   {
   public:
-    /* Note that we are using the data place of the device here as the affine data place */
+    /* Note that we are using the green context data place as the affine data place */
     impl(green_ctx_view gc_view)
-        : exec_place::impl(data_place::green_ctx(gc_view))
```
caugonnet, line 184: It is a little unclear whether it's a good idea for now to make a difference between data_place::green_ctx and data_place::device if the implementation is exactly the same without leveraging locality...

caugonnet: This was fixed

caugonnet:
pre-commit.ci autofix

PR #7266, opened 2026-01-16
Fix `__query_or` CPO

miscco:
The `__query_or` currently  requires that the query is always of the form `env.query(CPO, args...)`

However, we also want that to work with types that are their own query, which would satisfy `cpo(env, args...)`

This is the case for `::cuda::get_stream(::cudaStream_t)` and also for our memory resources

### DIFF
--- a/cub/test/catch2_test_env_launch_helper.h
+++ b/cub/test/catch2_test_env_launch_helper.h
@@ -185,7 +185,7 @@ struct device_memory_resource : cub::detail::device_memory_resource
     return nullptr;
   }
 
-  void deallocate_sync(void* /* ptr */, size_t /* bytes */)
+  void deallocate_sync(void* /* ptr */, size_t /* bytes */, size_t /* alignment */)
   {
     FAIL("CUB shouldn't use synchronous deallocation");
   }
@@ -206,6 +206,11 @@ struct device_memory_resource : cub::detail::device_memory_resource
     return cub::detail::device_memory_resource::allocate(stream, bytes);
   }
 
+  void deallocate(const cuda::stream_ref stream, void* ptr, size_t bytes, size_t /* alignment */)
+  {
+    deallocate(stream, ptr, bytes);
+  }
+
   void deallocate(const cuda::stream_ref stream, void* ptr, size_t bytes)
   {
     REQUIRE(target_stream == stream.get());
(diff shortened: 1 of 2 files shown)


davebayer (approved):

davebayer:
pre-commit.ci autofix

ericniebler:
today, given a query `q` and an `env`, there are two reasons why `q(env)` might be well-formed:

1. `env.query(q)` is well-formed, or
2. `q` has a fallback that returns a default value

today, `__query_or` is useful because it distinguishes between these cases. this change breaks that property.

the current semantics of `__query_or` also make it useful for implementing queries. `get_allocator_t` might look like this:

```c++
struct get_allocator_t {
  template <class Env>
  auto query(const Env& env, auto&&...) const noexcept {
    return __query_or(env, *this, std::allocator<std::byte>());
  }
};
```

the change proposed here would make this infinitely recurse.

would a `__call_or` or `__invoke_or` function, or a `__with_default` callable adaptor, meet your needs?

miscco:
> would a `__call_or` or `__invoke_or` function, or a `__with_default` callable adaptor, meet your needs?

I am impartial to what solution we use, but we need one that we can use uniformly in CCCL

libcudacxx/include/cuda/std/__execution/env.h
```diff
@@ -421,11 +422,12 @@ struct __query_or_t
 {
   _CCCL_EXEC_CHECK_DISABLE
   _CCCL_TEMPLATE(class _Env, class _Query, class _Default, class... _Args)
-  _CCCL_REQUIRES(__queryable_with<_Env, _Query, _Args...>)
-  [[nodiscard]] _CCCL_API constexpr auto operator()(const _Env& __env, _Query, _Default&&, _Args&&... __args) const
-    noexcept(__nothrow_queryable_with<_Env, _Query, _Args...>) -> __query_result_t<_Env, _Query, _Args...>
+  _CCCL_REQUIRES(__queryable_with<const _Env&, _Query, _Args...>)
```
miscco, line 425: @ericniebler I believe that is a bugfix, because a query with a non-const query method would satisfy `__queryable_with<_Env, _Query, _Args...>` but cannot be called here

ericniebler: true. a different fix would be for `__queryable_with` and friends to automatically const & qualify the `Env` argument.

pciolkosz:
> the change proposed here would make this infinitely recurse.

This is exactly what happened when I was trying a similar approach in https://github.com/NVIDIA/cccl/pull/7113

bernhardmgruber:
Btw, #7278 will add a bunch of tests for extracting streams from environments in CUB.

PR #7267, opened 2026-01-16
[Backport 3.2] Fix `__query_or` CPO

miscco:
The `__query_or` currently  requires that the query is always of the form `env.query(CPO, args...)`

However, we also want that to work with types that are their own query, which would satisfy `cpo(env, args...)`

This is the case for `::cuda::get_stream(::cudaStream_t)` and also for our memory resources

### DIFF
--- a/cub/test/catch2_test_env_launch_helper.h
+++ b/cub/test/catch2_test_env_launch_helper.h
@@ -180,7 +180,7 @@ struct device_memory_resource : cub::detail::device_memory_resource
     return nullptr;
   }
 
-  void deallocate_sync(void* /* ptr */, size_t /* bytes */)
+  void deallocate_sync(void* /* ptr */, size_t /* bytes */, size_t /* alignment */)
   {
     FAIL("CUB shouldn't use synchronous deallocation");
   }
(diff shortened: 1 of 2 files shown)


bernhardmgruber (approved):

libcudacxx/include/cuda/__functional/call_or.h
```diff
@@ -0,0 +1,65 @@
...
+#include <cuda/std/__concepts/concept_macros.h>
+#include <cuda/std/__tuple_dir/ignore.h>
+#include <cuda/std/__type_traits/is_callable.h>
+#include <cuda/std/__type_traits/is_nothrow_move_constructible.h>
+
+#include <cuda/std/__cccl/prologue.h>
+
+_CCCL_BEGIN_NAMESPACE_CUDA
+
+//! @brief `__call_or` is an higher-order function that accepts a function, a default
```
bernhardmgruber:
```suggestion
//! @brief `__call_or` is a higher-order function that accepts a function, a default
```

bernhardmgruber (approved):

cub/cub/device/device_transform.cuh
```diff
@@ -18,6 +18,7 @@
 #include <cub/util_namespace.cuh>
 
 #include <cuda/__functional/address_stability.h>
+#include <cuda/__functional/call_or.h>
```
bernhardmgruber, line 21: Remark: I think this is not strictly necessary, since there are no other changes in this file.

Alternatively, you can also use `call_or` in the implementation below.

miscco: Yeah that is a merge conflict, because the environment APIs came after 3.2

That said I would like to not run another CI run for this

bdice:
Cross-linking the main PR: https://github.com/NVIDIA/cccl/pull/7266

PR #7271, opened 2026-01-16
Replace `typedef` with `using`

### LINKED ISSUE #1747: Rewrite alias declarations from `typedef` to `using`
I got so used to alias declarations by now my parser hickups when encountering a `typedef` these days. Given the general better readability of the former, let's run the clang-tidy refactor on CCCL to rewrite the typedefs to using declarations.

_Originally posted by @bernhardmgruber in https://github.com/NVIDIA/cccl/pull/1721#discussion_r1593966840_

davebayer:
Closes #1747.=


libcudacxx/include/nv/target
```diff
@@ -41,7 +41,7 @@ namespace target
 {
 namespace detail
 {
-typedef unsigned long long base_int_t;
```
davebayer, line 44: @wmaxey is this change fine?

bernhardmgruber:
> I would like to block large scale refactoring's until we are sure we are done with the 3.2 backports

That's a good reason! Let's give it a (quiet) week or two. Thx @davebayer for addressing the issue so promptly, since I reassigned it just today.

bernhardmgruber (approved):

Issue #6400, opened 2025-10-30
[BUG]: Using cuda::std::zip_iterator with std::stable_sort runs into compilation error


oleksandr-pavlyk:
### Is this a duplicate?

- [x] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cccl/issues) for this bug and that I agree to the [Code of Conduct](CODE_OF_CONDUCT.md)

### Type of Bug

Compile-time Error

### Component

libcu++

### Describe the bug

Call to `std::stable_sort` with host-accessible `thrust::zip_iterator` works, but replacing it with `cuda::std::zip_iterator` runs into compilation errors.

### How to Reproduce

The reproducer is as follows:

```cu
#ifdef USE_CUDA_ITERATOR
# include <cuda/std/tuple>
# include <cuda/std/iterator>
#else
# include <thrust/iterator/zip_iterator.h>
#endif

#include <thrust/host_vector.h>
#include <algorithm>

struct compare_first_lt_op_t
{
 template <typename LhsT, typename RhsT>
  __host__ __device__ bool operator()(const LhsT& lhs, const RhsT& rhs) const
  {
    return thrust::get<0>(lhs) < thrust::get<0>(rhs);
  }
};

int main(void) {
  using key_t = int;
  using data_t = long;

  size_t num_items = 128;
  thrust::host_vector<key_t> keys_expected(num_items);
  thrust::host_vector<data_t> values_expected(num_items);

#ifdef USE_CUDA_ITERATOR
  auto zipped_expected_it = cuda::make_zip_iterator(keys_expected.begin(), values_expected.begin());
#else
  auto zipped_expected_it = thrust::make_zip_iterator(keys_expected.begin(), values_expected.begin());
#endif

  std::stable_sort(zipped_expected_it, zipped_expected_it + num_items, compare_first_lt_op_t{});

  return 0;
}
```

Compiling with `-UUSE_CUDA_ITERATOR` to use `thrust::zip_iterator` compiles fine. Switching to `-DUSE_CUDA_ITERATOR` to use `cuda::std::zip_iterator` runs into compilation errors.

Reproduced with CTK 12.9 and CTK 13.0.1, also with GCC 13 host compiler and LLVM 19 host compiler.

The compilation error log is below:

<details>
<summary>Compilation log</summary>

```
[...]

PR #7279, opened 2026-01-17
Add binary `transform_reduce` to Thrust

### LINKED ISSUE #4086: [FEA] Missing binary-binary `thrust::transform_reduce` overload
We seem to be missing the [binary-binary `thrust::transform_reduce` overload](https://wg21.link/P0452).

maifeeulasad:
## Description

related to #5160
Fixes: #4086

## Checklist


thrust/thrust/transform_reduce.h
```diff
@@ -181,6 +181,179 @@ template <typename InputIterator, typename UnaryFunction, typename OutputType, t
...
+          typename BinaryOp1,
+          typename BinaryOp2>
+_CCCL_HOST_DEVICE T transform_reduce(
+  const thrust::detail::execution_policy_base<DerivedPolicy>& exec,
+  InputIterator1 first1,
+  InputIterator1 last1,
+  InputIterator2 first2,
+  T init,
+  BinaryOp1 binary_op1,
+  BinaryOp2 binary_op2);
```
bernhardmgruber: Please use better names for the binary operations.
```suggestion
  BinaryOp1 reduce,
  BinaryOp2 transform);
```
Please also update documentation and other overloads as well.

thrust/thrust/transform_reduce.h
```diff
@@ -181,6 +181,179 @@ template <typename InputIterator, typename UnaryFunction, typename OutputType, t
...
+ *  \see \c reduce
+ *  \see \c inner_product
+ */
+template <typename DerivedPolicy, typename InputIterator1, typename InputIterator2, typename T>
+_CCCL_HOST_DEVICE T transform_reduce(
+  const thrust::detail::execution_policy_base<DerivedPolicy>& exec,
+  InputIterator1 first1,
+  InputIterator1 last1,
+  InputIterator2 first2,
+  T init);
```
bernhardmgruber: Important: Let's just default `BinaryOp1` and `BinaryOp2` on the other overload added in this PR, and drop the dedicated function here.

thrust/thrust/system/detail/generic/transform_reduce.inl
```diff
@@ -50,5 +55,50 @@ _CCCL_HOST_DEVICE OutputType transform_reduce(
...
+  T init,
+  BinaryOp1 binary_op1,
+  BinaryOp2 binary_op2)
+{
+  // Create a zip iterator to iterate over both input ranges simultaneously
+  const auto first = thrust::make_zip_iterator(first1, first2);
+  const auto last  = thrust::make_zip_iterator(last1, first2); // only first iterator matters
+
+  // Use the unary transform_reduce with the zipped iterators and a zip_function
+  return thrust::transform_reduce(exec, first, last, thrust::make_zip_function(binary_op2), init, binary_op1);
```
bernhardmgruber: Suggestion: let's use the newer iterators from `cuda`:
```suggestion
  const auto first = ::cuda::make_zip_iterator(first1, first2);
  const auto last  = ::cuda::make_zip_iterator(last1, first2); // only first iterator matters

  // Use the unary transform_reduce with the zipped iterators and a zip_function
  return thrust::transform_reduce(exec, first, last, ::cuda::make_zip_function(binary_op2), init, binary_op1);
```
Please update includes to `#include <cuda/iterator>`

bernhardmgruber (commented):
Thank you for the PR!

When I checked cppreference for `transform_exclusive_scan` I found out that there is no binary overload of that function and #5160 wrongly listed the algorithm. I assume the reason was that Thrust uses a different order of arguments, so when I prepared the list of missing algorithms, I didn't find an exact match and concluded it must be missing. Sorry for that! Please remove all code related to `transform_exclusive_scan` from this PR.

thrust/thrust/system/detail/generic/transform_reduce.inl
maifeeulasad: Tried but getting some segmentation fault. Maybe related to https://github.com/NVIDIA/cccl/issues/6400

maifeeulasad: But I am open for discussion on this one, and want to resolve this.

thrust/thrust/system/detail/generic/transform_reduce.inl
```diff
@@ -26,8 +26,13 @@
 #  pragma system_header
 #endif // no system header
 #include <thrust/iterator/transform_iterator.h>
+#include <thrust/iterator/zip_iterator.h>
 #include <thrust/reduce.h>
 #include <thrust/system/detail/generic/transform_reduce.h>
+#include <thrust/tuple.h>
+#include <thrust/zip_function.h>
```
miscco: For new APIs we should use `<cuda/std/tuple>` also this should probably use `cuda::zip_transfrom_iterator`

thrust/thrust/transform_reduce.h
```diff
@@ -189,6 +193,129 @@ template <typename InputIterator, typename UnaryFunction, typename OutputType, t
...
+ *  \see \c reduce
+ *  \see \c inner_product
+ */
+template <
+  typename InputIterator1,
+  typename InputIterator2,
+  typename T,
+  typename BinaryOp1 = ::cuda::std::plus<T>,
+  typename BinaryOp2 = ::cuda::std::multiplies<
+    ::cuda::std::common_type_t<::cuda::std::iter_value_t<InputIterator1>, ::cuda::std::iter_value_t<InputIterator2>>>>
```
miscco: Nitpick: I believe we should just use
```suggestion
  typename BinaryOp1 = ::cuda::std::plus<>,
  typename BinaryOp2 = ::cuda::std::multiplies<>>
```

thrust/thrust/transform_reduce.h
```diff
@@ -189,6 +193,129 @@ template <typename InputIterator, typename UnaryFunction, typename OutputType, t
...
+ *  \see \c inner_product
+ */
+template <
+  typename DerivedPolicy,
+  typename InputIterator1,
+  typename InputIterator2,
+  typename T,
+  typename BinaryOp1 = ::cuda::std::plus<T>,
+  typename BinaryOp2 = ::cuda::std::multiplies<
+    ::cuda::std::common_type_t<::cuda::std::iter_value_t<InputIterator1>, ::cuda::std::iter_value_t<InputIterator2>>>>
```
miscco: Ditto just use
```suggestion
  typename BinaryOp1 = ::cuda::std::plus<>,
  typename BinaryOp2 = ::cuda::std::multiplies<>>
```

thrust/thrust/transform_reduce.h
```diff
@@ -30,6 +30,10 @@
 #  pragma system_header
 #endif // no system header
 #include <thrust/detail/execution_policy.h>
+#include <thrust/iterator/iterator_traits.h>
+
+#include <cuda/std/functional>
+#include <cuda/std/type_traits>
```
miscco: We moved away from using compound headers in thrust.
```suggestion

#include <thrust/detail/execution_policy.h>
#include <thrust/iterator/iterator_traits.h>

#include <cuda/std/__functional/operations.h>
```

thrust/thrust/system/detail/generic/transform_reduce.inl
```diff
@@ -26,8 +26,13 @@
...
+
+#include <cuda/std/functional>
```
miscco: We moved away from compound headers for product code, also what is this header used for?
```suggestion
```

thrust/thrust/detail/transform_reduce.inl
```diff
@@ -78,4 +78,42 @@ OutputType transform_reduce(
...
+          typename BinaryOp1,
+          typename BinaryOp2>
+_CCCL_HOST_DEVICE T transform_reduce(
+  const thrust::detail::execution_policy_base<DerivedPolicy>& exec,
+  InputIterator1 first1,
+  InputIterator1 last1,
+  InputIterator2 first2,
+  T init,
+  BinaryOp1 reduce,
+  BinaryOp2 transform)
```
miscco: Nitpick: 

We consistently suffix operators with `_op` in the other APIs, so this should rather be
```suggestion
  BinaryOp1 reduce_op,
  BinaryOp2 transform_op)
```

Applies throughout

miscco (changes requested):

thrust/thrust/system/detail/generic/transform_reduce.inl
bernhardmgruber: There is a comment above that this apparently segfaults.

miscco: we should definitely investigate that 😿

bernhardmgruber: I need another clone of myself.

thrust/thrust/system/detail/generic/transform_reduce.inl
```diff
@@ -50,5 +55,28 @@ _CCCL_HOST_DEVICE OutputType transform_reduce(
...
+  InputIterator1 first1,
+  InputIterator1 last1,
+  InputIterator2 first2,
+  T init,
+  BinaryOp1 reduce,
+  BinaryOp2 transform)
+{
+  // Create a zip iterator to iterate over both input ranges simultaneously
+  const auto first = thrust::make_zip_iterator(first1, first2);
+  const auto last  = thrust::make_zip_iterator(last1, first2); // only first iterator matters
```
bernhardmgruber: The `// only first iterator matters` holds only true for Thrust's zip iterator, since it only checks the first iterator in the tuple when comparing two zip iterators.

Please try using `cuda::zip_iterator` and using:
```suggestion
  const auto last  = cuda::make_zip_iterator(last1, first2 + ::cuda::std::distance(first1, last1));
```
That may fix the issue.

miscco: `cuda::zip_iterator`does the right thing and check all zipped iterators for their distance,

so if one knows that the smallest distance is in `last1, first1` then the best way would be
```suggestion
  const auto last  = first + cuda::std::distance(first1, last1);
```

The suggestion by @bernhardmgruber is also valid

bernhardmgruber: How can this work? We only have an overload of `thrust::transform_reduce` that takes begin and end iterators. There is no begin and size overload.

miscco: I meant to use `zip_iterator::operator+`

maifeeulasad: I will be testing out changing all the header to `cuda`, instead of `thrust`. Maybe the issue was due to some mixing of headers. I will try to test first thing in the morning, tomorrow.

maifeeulasad: @bernhardmgruber @miscco  I saw that in `thrust/thrust/system/cuda/detail/transform_reduce.h` file we can import `cuda` related headers. And on the other hand in non cuda files, it breaks the executable.

Care to add some remarks on this please?

bernhardmgruber: You can use `<cuda/std/*>` headers everywhere, and a lot of `<cuda/*>` headers work in non-CUDA code as well.

maifeeulasad:
Sorry for the delay. I have tried addressing all the reviews. And I am open for upcoming reivews as well.

miscco (changes requested):
Please also add some benchmarks to thrust/benchmark

maifeeulasad:
@miscco I missed that, extremely sorry for that. Care to take a look now?

miscco (approved):

thrust/benchmarks/bench/transform_reduce/binary.cu
```diff
@@ -0,0 +1,70 @@
...
+// This computes: reduce(init, transform(*first1, *first2), transform(*(first1+1), *(first2+1)), ...)
+
+template <typename T>
+static void basic(nvbench::state& state, nvbench::type_list<T>)
+{
+  const auto elements = static_cast<std::size_t>(state.get_int64("Elements"));
+
+  auto generator               = generate(elements);
+  thrust::device_vector<T> lhs = generator;
+  thrust::device_vector<T> rhs = generator;
```
bernhardmgruber: Suggestion: Would look a bit better:
```suggestion
  thrust::device_vector<T> lhs = generate(elements);
  thrust::device_vector<T> rhs = generate(elements);
```

thrust/benchmarks/bench/transform_reduce/binary.cu
```diff
@@ -0,0 +1,70 @@
...
+  .add_int64_power_of_two_axis("Elements", nvbench::range(16, 28, 4));
+
+template <typename T>
+static void custom_ops(nvbench::state& state, nvbench::type_list<T>)
+{
+  const auto elements = static_cast<std::size_t>(state.get_int64("Elements"));
+
+  auto generator               = generate(elements);
+  thrust::device_vector<T> lhs = generator;
+  thrust::device_vector<T> rhs = generator;
```
bernhardmgruber: Same here.

thrust/thrust/detail/transform_reduce.inl
```diff
@@ -78,4 +78,61 @@ OutputType transform_reduce(
...
+          typename InputIterator1,
+          typename InputIterator2,
+          typename T,
+          typename BinaryOp1,
+          typename BinaryOp2>
+_CCCL_HOST_DEVICE
+// SFINAE: only participate when InputIterator2 is actually an iterator
+typename ::cuda::std::enable_if<
+  ::cuda::std::is_convertible<typename ::cuda::std::iterator_traits<InputIterator2>::iterator_category*, void*>::value,
+  T>::type
```
bernhardmgruber: Important: Please use the C++17 forms:

```suggestion
typename ::cuda::std::enable_if_t<
  ::cuda::std::is_convertible_v<typename ::cuda::std::iterator_traits<InputIterator2>::iterator_category*, void*>,
  T>
```
Also please put the constraint into the template head as an `enable_if_t<constraint, int> = 0` instead of the return type.

Applies to other overloads as well.

thrust/thrust/system/cuda/detail/transform_reduce.h
```diff
@@ -134,6 +135,28 @@ T _CCCL_HOST_DEVICE transform_reduce(
        cvt_to_seq(derived_cast(policy)), first, first + num_items, transform_op, init, reduce_op);));
   return init;
 }
+
+// Binary transform_reduce with CUDA-specific iterators
```
bernhardmgruber: Suggestion: That comment does not make sense. What is a CUDA-specific iterator? I would drop it.

thrust/thrust/system/cuda/detail/transform_reduce.h
```diff
@@ -134,6 +135,28 @@ T _CCCL_HOST_DEVICE transform_reduce(
...
+  TransformOp transform_op)
+{
+  // Calculate the number of elements
+  const auto n = ::cuda::std::distance(first1, last1);
+
+  // Create a zip_transform_iterator to iterate over both input ranges simultaneously
+  const auto first = ::cuda::make_zip_transform_iterator(transform_op, first1, first2);
+
+  // Delegate to the unary transform_reduce which handles CDP dispatch
+  return thrust::cuda_cub::transform_reduce(policy, first, first + n, ::cuda::std::identity{}, init, reduce_op);
```
bernhardmgruber, line 157: Question: Why can't we just call the regular `reduce` here? Why does it have to be `transform_reduce` with an identity transformation?

maifeeulasad: I thought it increases maintainability. And also reduces the call for costly `THRUST_CDP_DISPATCH`, as we are  just delegating this way we don't create or use another boilerplate. 

But I can change it if you want, all the other comments has been addressed so far.

ref: https://github.com/maifeeulasad/cccl/blob/575e784c51a74b3a1620b2b650bc9e0dbb79de21/cub/cub/device/device_reduce.cuh#L500-L505

thrust/thrust/detail/transform_reduce.inl
```diff
@@ -78,4 +78,67 @@ OutputType transform_reduce(
...
+  typename DerivedPolicy,
+  typename InputIterator1,
+  typename InputIterator2,
+  typename T,
+  typename BinaryOp1,
+  typename BinaryOp2,
+  // SFINAE: only participate when InputIterator2 is actually an iterator
+  ::cuda::std::enable_if_t<
+    ::cuda::std::is_convertible_v<typename ::cuda::std::iterator_traits<InputIterator2>::iterator_category*, void*>,
+    int>>
```
bernhardmgruber:
```suggestion
  ::cuda::std::enable_if_t<
    ::cuda::std::is_convertible_v<typename ::cuda::std::iterator_traits<InputIterator2>::iterator_category*, void*>,
    int> = 0>
```

thrust/thrust/detail/transform_reduce.inl
```diff
@@ -78,4 +78,67 @@ OutputType transform_reduce(
...
+template <
+  typename InputIterator1,
+  typename InputIterator2,
+  typename T,
+  typename BinaryOp1,
+  typename BinaryOp2,
+  // SFINAE: only participate when InputIterator2 is actually an iterator
+  ::cuda::std::enable_if_t<
+    ::cuda::std::is_convertible_v<typename ::cuda::std::iterator_traits<InputIterator2>::iterator_category*, void*>,
+    int>>
```
bernhardmgruber:
```suggestion
  ::cuda::std::enable_if_t<
    ::cuda::std::is_convertible_v<typename ::cuda::std::iterator_traits<InputIterator2>::iterator_category*, void*>,
    int> = 0>
```

maifeeulasad:
bump @bernhardmgruber

thrust/thrust/system/cuda/detail/transform_reduce.h
bernhardmgruber: > And also reduces the call for costly `THRUST_CDP_DISPATCH`, as we are just delegating this way we don't create or use another boilerplate.

I don't follow, sorry. `THRUST_CDP_DISPATCH` should not be costly. Did you measure?

I would just call
```suggestion
  return thrust::cuda_cub::reduce_n(policy, first, n, init, reduce_op);
```
here.

thrust/thrust/detail/transform_reduce.inl
```diff
@@ -78,4 +78,67 @@ OutputType transform_reduce(
...
+  typename DerivedPolicy,
+  typename InputIterator1,
+  typename InputIterator2,
+  typename T,
+  typename BinaryOp1,
+  typename BinaryOp2,
+  // SFINAE: only participate when InputIterator2 is actually an iterator
+  ::cuda::std::enable_if_t<
+    ::cuda::std::is_convertible_v<typename ::cuda::std::iterator_traits<InputIterator2>::iterator_category*, void*>,
+    int> = 0>
```
bernhardmgruber, line 92: @miscco is there a better way to check if `InputIterator2` is an iterator?

thrust/thrust/transform_reduce.h
```diff
@@ -189,6 +193,135 @@ template <typename InputIterator, typename UnaryFunction, typename OutputType, t
...
+ *  \param transform The binary transformation operation (defaults to <tt>::cuda::std::multiplies<common_type></tt>).
+ *  \return The result of the transformed reduction.
+ *
+ *  \tparam DerivedPolicy The name of the derived execution policy.
+ *  \tparam InputIterator1 is a model of <a href="https://en.cppreference.com/w/cpp/iterator/input_iterator">Input
+ * Iterator</a>. \tparam InputIterator2 is a model of <a
+ * href="https://en.cppreference.com/w/cpp/iterator/input_iterator">Input Iterator</a>. \tparam T is convertible to \p
+ * BinaryOp1's first and second argument types. \tparam BinaryOp1 is a binary function whose return type is convertible
+ * to \p T (defaults to <tt>::cuda::std::plus<T></tt>). \tparam BinaryOp2 is a binary function whose return type is
+ * convertible to \p BinaryOp1's second argument type (defaults to <tt>::cuda::std::multiplies<common_type></tt>).
```
bernhardmgruber, line 220: Important: Please fix the formatting of the `\tparam`s here

thrust/thrust/transform_reduce.h
```diff
@@ -189,6 +193,135 @@ template <typename InputIterator, typename UnaryFunction, typename OutputType, t
...
+ *  \code
+ *  #include <thrust/transform_reduce.h>
+ *  #include <thrust/functional.h>
+ *  #include <thrust/execution_policy.h>
+ *  ...
+ *  int data1[6] = {1, 0, 2, 2, 1, 3};
+ *  int data2[6] = {4, 1, 5, 3, 2, 1};
+ *
+ *  // Explicit operations:
+ *  int result1 = thrust::transform_reduce(thrust::host, data1, data1 + 6, data2, 0,
```
bernhardmgruber, line 231: Suggestion:

```suggestion
 *  int result1 = thrust::transform_reduce(thrust::device, data1, data1 + 6, data2, 0,
```

thrust/thrust/transform_reduce.h
```diff
@@ -189,6 +193,135 @@ template <typename InputIterator, typename UnaryFunction, typename OutputType, t
...
+    ::cuda::std::is_convertible_v<typename ::cuda::std::iterator_traits<InputIterator2>::iterator_category*, void*>,
+    int> = 0>
+_CCCL_HOST_DEVICE T transform_reduce(
+  const thrust::detail::execution_policy_base<DerivedPolicy>& exec,
+  InputIterator1 first1,
+  InputIterator1 last1,
+  InputIterator2 first2,
+  T init,
+  BinaryOp1 reduce    = BinaryOp1(),
+  BinaryOp2 transform = BinaryOp2());
```
bernhardmgruber, line 262: Question: Hmm. The order of parameters here does not fit to the unary overload of `transform_reduce` we have in Thrust. It fits the C++ standard library though. I am more leaning towards retaining the Thrust order here for consistency. That is, ordering the parameters as first1, last1, first2, transform_op, init, reduce_op. What do you think?

@miscco what do you think as well?

_deps/fmt-src
miscco, line 1: Do these really need to be changed?

thrust/thrust/detail/transform_reduce.inl
miscco: I would just test against `::cuda::std::__has_random_access_traversal<InputIterator2>`

thrust/thrust/transform_reduce.h
miscco: I would keep the thrust ordering we already have the binary overload in libcu++ and they differe anyhow

PR #7281, opened 2026-01-18
cuda.compute: Consolidate caching logic across all algorithms

shwina:
## Description

This PR is purely a refactor.

In https://github.com/NVIDIA/cccl/pull/6938#discussion_r2608031116, I promised to simplify how we do caching. This PR consolidates all of the caching logic by introducing a registry that maps types to key functions (defined in `_caching.py`). Thus, we don't have to write a cache key function for every individual algorithm (which was incredibly repetitive).

As a nice side-effect, this also removes the indirection from the referenced comment.

## Checklist

### DIFF
--- a/python/cuda_cccl/cuda/compute/_caching.py
+++ b/python/cuda_cccl/cuda/compute/_caching.py
@@ -4,6 +4,7 @@
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 import functools
+import types
 from typing import Any, Callable, Hashable
 
 import numpy as np
@@ -13,6 +14,8 @@ try:
 except ImportError:
     from cuda.core.experimental import Device
 
+from ._utils.protocols import get_dtype
+from .typing import DeviceArrayLike, GpuStruct
 
 # Registry thet maps type -> key function for extracting cache key
 # from a value of that type.
@@ -47,32 +50,6 @@ def _key_for(value: Any) -> Hashable:
         if isinstance(value, registered_type):
             return keyer(value)
 
-    # For numpy dtypes, return directly (they're already hashable)
-    if isinstance(value, np.dtype):
-        return value
-
-    # For None, return directly
-    if value is None:
-        return None
-
-    # Check if it's a DeviceArrayLike (duck-typed via __cuda_array_interface__)
-    if hasattr(value, "__cuda_array_interface__"):
-        # It's a device array-like, extract dtype and fully-qualified type name
-        from ._utils.protocols import get_dtype
-
-        # Use fully-qualified type name (e.g., 'numpy.ndarray' v/s 'cupy.ndarray')
-        type_fqn = f"{type(value).__module__}.{type(value).__name__}"
-        return (type_fqn, get_dtype(value))
-
-    # Check if it has a dtype attribute
-    if hasattr(value, "dtype"):
-        type_fqn = f"{type(value).__module__}.{type(value).__name__}"
-        return (type_fqn, value.dtype)
-
-    # For callables, wrap in CachableFunction
-    if callable(value):
-        return CachableFunction(value)
-
     # Fallback: use value directly (assumes it's hashable)
     return value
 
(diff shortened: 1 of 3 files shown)


python/cuda_cccl/cuda/compute/op.py
```diff
@@ -17,11 +17,10 @@ class _OpAdapter:
     Provides a unified interface for operators, whether they are:
     - Well-known operations (OpKind.PLUS, OpKind.MAXIMUM, etc.)
     - Stateless user-provided callables
-    """
 
-    def get_cache_key(self) -> Hashable:
-        """Return a hashable cache key for this operator."""
-        raise NotImplementedError("Subclasses must implement this method")
+    Note: Cache key extraction is handled by the cache key registry,
```
shwina: Just delete this mehtod

python/cuda_cccl/cuda/compute/_caching.py
```diff
@@ -4,56 +4,211 @@
...
+            type_: The type to register
+            key_function: A callable that takes an instance of type_ and returns a hashable cache key
+        """
+        _KEY_FUNCTIONS[type_] = key_function
+
+    def __call__(self, func_or_key_function=None):
+        """
+        Decorator to cache the result of the decorated function.
+
+        If `key` is provided, it should be a function that computes the cache key
```
NaderAlAwar:
```suggestion
        If `func_or_key_function` is provided, it should be a function that computes the cache key
```

python/cuda_cccl/cuda/compute/_caching.py
```diff
@@ -4,56 +4,211 @@
...
+        from the function arguments. Otherwise, the cache key is automatically
```
NaderAlAwar:
```suggestion
        from the decorated function's arguments. Otherwise, the cache key is automatically
```

python/cuda_cccl/cuda/compute/_caching.py
```diff
@@ -4,56 +4,211 @@
...
+        The CUDA compute capability of the current device is appended to
+        the cache key.
+
+        The decorated function is automatically registered in the central
+        cache registry for easy cache management.
+
+        Args:
+            func_or_key_function: The function to decorate, or a custom key function.
+        """
+        # Check if this is being used without parentheses (@cache_with_registered_key_functions)
```
NaderAlAwar: Suggestion: if `func_or_key_function` is always supposed to be a callable, can we just add a type check at the start and fail fast if the passed value is not a callable? Then remove the checks for callable later

python/cuda_cccl/cuda/compute/_caching.py
```diff
@@ -4,56 +4,211 @@
...
+
+        A key function is a function that extracts a hashable cache key from a value.
+
+        Args:
+            type_: The type to register
+            key_function: A callable that takes an instance of type_ and returns a hashable cache key
+        """
+        _KEY_FUNCTIONS[type_] = key_function
+
+    def __call__(self, func_or_key_function=None):
```
NaderAlAwar: Suggestion: add a type annotation to `func_or_key_function`

python/cuda_cccl/cuda/compute/_caching.py
```diff
@@ -4,56 +4,211 @@
...
 
-        # Register the cache in the central registry
-        cache_name = func.__qualname__
-        _cache_registry[cache_name] = inner
+    # For callables, wrap in CachableFunction
+    if callable(value):
+        return CachableFunction(value)
+
+    # Fallback: use value directly (assumes it's hashable)
+    return value
```
NaderAlAwar, line 54: Suggestion: lets try not falling back, and raise an error or exception instead

python/cuda_cccl/cuda/compute/_caching.py
```diff
@@ -4,56 +4,211 @@
...
+        return _KEY_FUNCTIONS[value_type](value)
+
+    # Check for instance match (handles inheritance)
+    for registered_type, keyer in _KEY_FUNCTIONS.items():
+        if isinstance(value, registered_type):
+            return keyer(value)
+
+    # For numpy dtypes, return directly (they're already hashable)
+    if isinstance(value, np.dtype):
+        return value
```
NaderAlAwar: Suggestion: let us register a key function for np.dtype (if we don't already)

python/cuda_cccl/cuda/compute/_caching.py
```diff
@@ -4,56 +4,211 @@
...
+    if hasattr(value, "dtype"):
+        type_fqn = f"{type(value).__module__}.{type(value).__name__}"
+        return (type_fqn, value.dtype)
 
-        # Register the cache in the central registry
-        cache_name = func.__qualname__
-        _cache_registry[cache_name] = inner
+    # For callables, wrap in CachableFunction
+    if callable(value):
+        return CachableFunction(value)
```
NaderAlAwar: Suggestion: register key functions instead of implementing logic here

python/cuda_cccl/cuda/compute/_caching.py
```diff
@@ -4,56 +4,211 @@
...
+        sorted_kwargs = sorted(kwargs.items())
+        kwarg_keys = tuple((k, _key_for(v)) for k, v in sorted_kwargs)
+        return positional_keys + (kwarg_keys,)
+
+    return positional_keys
+
+
+# Register built-in types
+# Include fully-qualified type name to distinguish np.ndarray from cp.ndarray from GpuStruct
+_KEY_FUNCTIONS[np.ndarray] = lambda arr: ("numpy.ndarray", arr.dtype)
```
NaderAlAwar: Suggestion: use `register()` instead if possible

python/cuda_cccl/cuda/compute/_caching.py
```diff
@@ -4,56 +4,211 @@
...
+        Args:
+            func_or_key_function: The function to decorate, or a custom key function.
+        """
+        # Check if this is being used without parentheses (@cache_with_registered_key_functions)
+        if callable(func_or_key_function) and not hasattr(
+            func_or_key_function, "__self__"
+        ):
+            # Direct decoration: @cache_with_registered_key_functions
+            # func_or_key is the actual function being decorated
+            return self._make_wrapper(None)(func_or_key_function)
```
NaderAlAwar: Suggestion: let's not accept custom key functions for now since none of the existing algorithms need it

python/cuda_cccl/cuda/compute/_caching.py
```diff
@@ -4,56 +4,211 @@
...
+            return inner
+
+        return deco
+
+
+# Create the singleton instance
+cache_with_registered_key_functions = _CacheWithKeyFunctionsDecorator()
+
+# Keep old name for backwards compatibility
+cache_with_key = cache_with_registered_key_functions
```
NaderAlAwar: Suggestion: remove this

python/cuda_cccl/cuda/compute/_caching.py
shwina: We take no arguments in the decorator now, so this has gone away.

python/cuda_cccl/cuda/compute/_caching.py
shwina: Done (clarified whose arguments in the docstring).

python/cuda_cccl/cuda/compute/_caching.py
shwina: Since we don't take `func_or_key_function` as an argument anymore (just `func`), this is a non-issue

python/cuda_cccl/cuda/compute/_caching.py
shwina: Ditto as above.

python/cuda_cccl/cuda/compute/_caching.py
shwina: So - this becomes tedious, because we'll then have to register all builtin types (like ints, floats, strings, enum values), which are all used as arguments today.

I think this is a reasonable behaviour, and in the worst case we lose caching. Which should be visible in benchmarks. Which we don't run in CI though :(

python/cuda_cccl/cuda/compute/_caching.py
shwina: Done.

python/cuda_cccl/cuda/compute/_caching.py
shwina: Done.

python/cuda_cccl/cuda/compute/_caching.py
shwina: Done.

python/cuda_cccl/cuda/compute/_caching.py
shwina: Done.

python/cuda_cccl/cuda/compute/_caching.py
shwina: Done.

PR #7283, opened 2026-01-18
Fix reference to `cccl-version.json`

### LINKED ISSUE #6438: [BUG]: CMake: Reference to non-existent `cccl_version` file
### Is this a duplicate?

### Type of Bug

Something else

### Component

General CCCL

### Describe the bug

CMake fails to configure the CCCL project if it's not part of a Git repository; the following snippet fails to find a `cccl_version` file:

https://github.com/NVIDIA/cccl/blob/ebd60639e898a8c37147d4b1b66869fae95d8fad/benchmarks/cmake/CCCLBenchmarkRegistry.cmake#L22-L29

### How to Reproduce

Remove the `.git` folder and attempt to configure the project.

### Expected behavior

CMake configuration should complete successfully.

### Reproduction link

_No response_

### Operating System

_No response_

### nvidia-smi output

_No response_

### NVCC version

_No response_

Pansysk75:
## Description

Fixes #6438

Modified `CCCLBenchmarkRegistry.cmake` to pull version from `cccl-version.json` (at root of monorepo), instead of the now non-existent `cccl-version` file.
Without this change, CMake would fail to configure if outside of a Git repository.
This resolves the issue, hopefully preserving the original intended behavior.

@alliepiper

### DIFF
--- a/benchmarks/cmake/CCCLBenchmarkRegistry.cmake
+++ b/benchmarks/cmake/CCCLBenchmarkRegistry.cmake
@@ -20,11 +20,10 @@ if(GIT_FOUND)
 endif()
 
 # Sometimes this script is used outside of a Git repository.
-# In this case, we read the revision from cccl/cccl_version instead.
+# In this case, we read the revision from cccl/cccl-version.json instead.
 if ("${cccl_revision}" STREQUAL "")
-  file(READ "${CMAKE_SOURCE_DIR}/cccl_version" cccl_revision)
-  string(STRIP "${cccl_revision}" cccl_revision)
-  string(REPLACE "\n" "" cccl_revision "${cccl_revision}")
+  file(READ "${CMAKE_SOURCE_DIR}/cccl-version.json" cccl_revision_json)
+  string(JSON cccl_revision GET "${cccl_revision_json}" "full")
 endif()
 message(STATUS "Git revision: ${cccl_revision}")
 


bernhardmgruber:
I think the PR is fine, but @gevtushenko wrote the affected code in #1305 to cover NVIDIA internal uses where we benchmark outside a git repo. I don't know whether a full clone (including `cccl-version.json`) will be available there. I would like to have his opinion, whether this breaks any internal workflows.

alliepiper:
Good catch! I thought this was just a typo on a rarely-if-ever-used code path, but looking closer the file contents are also different -- the file that's currently read isn't json, but just contains a string with the version. This string is then directly passed to some other utilities, which may break if they get the json payload instead of just X.Y.Z.

Let's see what @gevtushenko says, and we might just need to add a comment here to clarify that this is for a different file than the json one -- or migrate internal workflows to align on the existing json file before we merge.

Pansysk75:
Thank you, I only came across this due to an unusual IDE setup, and I assumed it was referring to a now obsolete file that used to be in this repo.

gevtushenko:
Hello! The expectations of `cccl-version` are different from that of `cccl-version.json`. `cccl-version` is created internally as `git describe > cccl-version`, so it's not a json file. We need a human readable version, like `3.2` followed by a number that uniquely represent a commit. `cccl-version.json` doesn't have enough information for that, so we shouldn't rely on it.

PR #7285, opened 2026-01-19
[nvrtcc] Custom nvrtcc cmdline parser

### LINKED ISSUE #7179: [BUG]: Fix option parsing with `nvrtcc`
Currently, we use `argparse` to parse `nvrtcc` and `nvcc` options. However, the original `nvcc`'s options parser is a bit different and the current implementation doesn't support taking options as an argument for another option, for example:
```sh
nvrtcc -Xcompiler -my_option ...
```
fails with `nvrtcc`, but works fine with `nvcc`. There are other problems, too, for example `argparse` does _short options completions_ which `nvcc` doesn't do.

We can either find a different parser or write our own.

davebayer:
Closes #7179

### DIFF
--- a/nvrtcc/bin/nvrtcc
+++ b/nvrtcc/bin/nvrtcc
@@ -12,6 +12,8 @@
 
 import argparse
 import copy
+import dataclasses
+import enum
 import os
 import re
 import shlex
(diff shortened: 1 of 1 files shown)


davebayer:
Closing for now, we can get back to it once it becomes a priority

Issue #7298, opened 2026-01-20
Profile cmake generation


bernhardmgruber:
I recently learned you can profile cmake using `--profiling-format=google-trace --profiling-output=cmake_profile.json`. Since running cmake is not fast, I wondered what we can find. CMake configure time matters especially on clusters with network file systems, were cmake generation can take a couple of minutes.

So I tried it using a cmake command I commonly use to get the `all-dev` preset:

```bash
cccl/build_prof$ CXX=clang++-20 cmake .. --preset all-dev -B. -G "Unix Makefiles" -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_CUDA_HOST_COMPILER=clang++-20 -DTHRUST_ENABLE_MULTICONFIG=OFF -DCCCL_ENABLE_BENCHMARKS=OFF -DCMAKE_CUDA_FLAGS="--ftemplate-backtrace-limit=0 -lineinfo" -DCMAKE_CUDA_ARCHITECTURES=120 --profiling-format=google-trace --profiling-output=cmake_profile.json
```
results in this graph:
<img width="2066" height="366" alt="Image" src="https://github.com/user-attachments/assets/412eb40a-1a86-4fa1-bb83-211f7d364855" />
We can see that fetching dependencies via `fetchcontent` is expensive.

On a second run of the above command, where we already have an existing binary directory with fetched dependencies:
<img width="2066" height="366" alt="Image" src="https://github.com/user-attachments/assets/53029509-8543-4cff-a918-e533f9121c2b" />
We now see that generating the build system takes the most time. I assume because we have a LOT of targets.

This is for CUB tuning, which we run very frequently on clusters:
```
cccl/build_prof$ cmake .. --preset=cub-tune
```
<img width="2067" height="432" alt="Image" src="https://github.com/user-attachments/assets/a68a6ccb-3537-4a86-b92f-2daa731b0042" />
We can see that `try_compile` from CMake takes some time (I guess we cannot change that), and the other large block is getting nvbench, which pulls in catch2 (which we wouldn't need as consumers of nvbench), and libfmt and nlohmann_json (both via rapids_cpm_find).

We could consider not fetching Catch2 when nvbench is consumed as sub project, and refactoring libfmt to `std::format` so we can lose the dependency.

In any case, profiling cmake is surprisingly easy and insightful and there are probably some low hanging fruits on how to configure faster that we could try.

PR #7300, opened 2026-01-20
Remove fmtlib from CCCL

davebayer:
We were using it only for atomics codegen in libcu++ and we can just use `std::format` instead as was suggested in #7298.


bernhardmgruber:
We should also ask for @wmaxey's opinion.

libcudacxx/codegen/generators/compare_and_swap.h
```diff
@@ -126,15 +126,31 @@ static inline _CCCL_DEVICE bool __cuda_atomic_compare_exchange(
           {
             continue;
           }
-          out << fmt::format(
-            (size == 128) ? asm_intrinsic_format_128 : asm_intrinsic_format,
```
davebayer, line 130: These are not constant expressions, so I need to split it up into 2 calls

PR #7312, opened 2026-01-21
Test passing a custom policy to DispatchFixedSizeSegmentedReduce

bernhardmgruber:
Generated using Cursor+gpt-5.2-codex. Then reviewed on a high level and applied a few fixed.


bernhardmgruber:
Turns out `cub::detail::reduce::DispatchFixedSizeSegmentedReduce` is not a public dispatcher. I got confused by the naming convention.

PR #5516, opened 2025-08-13
`cuda::clamp`

fbusato:
## Description

`cuda::std::clamp` is a widely used utility in CCCL and in general code. Unfortunately, it doesn't produce the optimal code for device code for the most common use cases: _integer_ and _floating-point types_. The main reason is that `cuda::std::clamp` returns a reference instead of a value, which is not needed for primitive types.

The PR provides `cuda::clamp` to generate optimal code for integer and floating-point types.

Requires `constexpr` `fmin/fmax`.

PR #3449, opened 2025-01-19
Implement saturation arithmetic from C++26

davebayer:
This PR implements saturation arithmetic from C++26 ([P0543R3](https://wg21.link/P0543R3)).

PR #7325, opened 2026-01-22
Implement internal `min`/`max`/`clamp` returning value

davebayer:
I've heard several times already that `cuda::std::(min|max|clamp)` are problematic because they return reference and not value.

This PR introduces internal variants that always return value.

### DIFF
--- a/libcudacxx/include/cuda/__tma/make_tma_descriptor.h
+++ b/libcudacxx/include/cuda/__tma/make_tma_descriptor.h
@@ -479,7 +479,7 @@ _CCCL_HOST_API inline __tma_box_sizes_array_t __get_box_sizes(
   size_t __total_size = 1;
   for (int __i = 0; __i < __rank; ++__i)
   {
-    const auto __max_box_size = static_cast<int>(::cuda::std::__vmin(__tensor_sizes[__i], uint64_t{256}));
+    const auto __max_box_size = static_cast<int>(::cuda::std::__min(__tensor_sizes[__i], uint64_t{256}));
     const auto __box_size     = __box_sizes[__rank - 1 - __i];
     if (__box_size <= 0 || __box_size > __max_box_size)
     {
@@ -569,7 +569,7 @@ _CCCL_HOST_API inline __tma_elem_strides_array_t __get_elem_strides(
   const int __init_index = (__interleave_layout == tma_interleave_layout::none) ? 1 : 0;
   for (int __i = __init_index; __i < __rank; ++__i)
   {
-    const auto __max_elem_stride = static_cast<int>(::cuda::std::__vmin(__tensor_sizes[__i], uint64_t{8}));
+    const auto __max_elem_stride = static_cast<int>(::cuda::std::__min(__tensor_sizes[__i], uint64_t{8}));
     const auto __elem_stride     = __elem_strides[__rank - 1 - __i];
     if (__elem_stride <= 0 || __elem_stride > __max_elem_stride)
     {
--- a/libcudacxx/include/cuda/__tma/make_tma_descriptor.h
+++ b/libcudacxx/include/cuda/__tma/make_tma_descriptor.h
@@ -479,7 +479,7 @@ _CCCL_HOST_API inline __tma_box_sizes_array_t __get_box_sizes(
   size_t __total_size = 1;
   for (int __i = 0; __i < __rank; ++__i)
   {
-    const auto __max_box_size = static_cast<int>(::cuda::std::min(__tensor_sizes[__i], uint64_t{256}));
+    const auto __max_box_size = static_cast<int>(::cuda::std::__vmin(__tensor_sizes[__i], uint64_t{256}));
     const auto __box_size     = __box_sizes[__rank - 1 - __i];
     if (__box_size <= 0 || __box_size > __max_box_size)
     {
@@ -569,7 +569,7 @@ _CCCL_HOST_API inline __tma_elem_strides_array_t __get_elem_strides(
   const int __init_index = (__interleave_layout == tma_interleave_layout::none) ? 1 : 0;
   for (int __i = __init_index; __i < __rank; ++__i)
   {
-    const auto __max_elem_stride = static_cast<int>(::cuda::std::min(__tensor_sizes[__i], uint64_t{8}));
+    const auto __max_elem_stride = static_cast<int>(::cuda::std::__vmin(__tensor_sizes[__i], uint64_t{8}));
     const auto __elem_stride     = __elem_strides[__rank - 1 - __i];
     if (__elem_stride <= 0 || __elem_stride > __max_elem_stride)
     {
--- a/libcudacxx/include/cuda/std/__algorithm/clamp.h
+++ b/libcudacxx/include/cuda/std/__algorithm/clamp.h
@@ -21,6 +21,8 @@
 #endif // no system header
 
 #include <cuda/std/__algorithm/comp.h>
+#include <cuda/std/__algorithm/max.h>
+#include <cuda/std/__algorithm/min.h>
 
 #include <cuda/std/__cccl/prologue.h>
 
@@ -39,6 +41,12 @@ template <class _Tp>
 {
   _CCCL_ASSERT(!(__hi < __lo), "Bad bounds passed to cuda::std::clamp");
   return __v < __lo ? __lo : __hi < __v ? __hi : __v;
+}
+
+template <class _Tp>
+[[nodiscard]] _CCCL_API constexpr _Tp __vclamp(_Tp __v, _Tp __lo, _Tp __hi) noexcept
+{
+  return ::cuda::std::__vmax(__lo, ::cuda::std::__vmin(__v, __hi));
 }
 
 _CCCL_END_NAMESPACE_CUDA_STD
--- a/libcudacxx/include/cuda/std/__algorithm/clamp.h
+++ b/libcudacxx/include/cuda/std/__algorithm/clamp.h
@@ -44,7 +44,7 @@ template <class _Tp>
 }
 
 //! @brief Internal version of clamp that works with values instead of references. Can be used with extended arithmetic
-//!        types. Should be preferred as it produces better optimized code with nvcc.
+//!        types. Should be preferred as it produces better optimized code with nvcc (see nvbug 5455679).
 template <class _Tp>
 [[nodiscard]] _CCCL_API constexpr _Tp __clamp(_Tp __v, _Tp __lo, _Tp __hi) noexcept
 {
--- a/libcudacxx/include/cuda/std/__algorithm/clamp.h
+++ b/libcudacxx/include/cuda/std/__algorithm/clamp.h
@@ -43,10 +43,12 @@ template <class _Tp>
   return __v < __lo ? __lo : __hi < __v ? __hi : __v;
 }
 
+//! @brief Internal version of clamp that works with values instead of references. Can be used with extended arithmetic
+//!        types. Should be preferred as it produces better optimized code with nvcc.
 template <class _Tp>
-[[nodiscard]] _CCCL_API constexpr _Tp __vclamp(_Tp __v, _Tp __lo, _Tp __hi) noexcept
+[[nodiscard]] _CCCL_API constexpr _Tp __clamp(_Tp __v, _Tp __lo, _Tp __hi) noexcept
 {
-  return ::cuda::std::__vmax(__lo, ::cuda::std::__vmin(__v, __hi));
+  return ::cuda::std::__max(__lo, ::cuda::std::__min(__v, __hi));
 }
 
 _CCCL_END_NAMESPACE_CUDA_STD
--- a/libcudacxx/include/cuda/std/__algorithm/min.h
+++ b/libcudacxx/include/cuda/std/__algorithm/min.h
@@ -70,17 +70,17 @@ template <class _Tp>
   {
     _CCCL_IF_NOT_CONSTEVAL_DEFAULT
     {
-      NV_IF_TARGET(NV_IS_DEVICE({
-        if constexpr (sizeof(_Tp) < sizeof(int))
-        {
-          using _Up = __make_nbit_int_t<32, is_signed_v<_Tp>>;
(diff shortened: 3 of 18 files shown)


fbusato:
I'm 100% in favor of that. Just for reference, this is a missed compiler optimization, see nvbug 5455679. I will close https://github.com/NVIDIA/cccl/pull/5516 then

libcudacxx/include/cuda/std/__algorithm/clamp.h
```diff
@@ -41,6 +43,14 @@ template <class _Tp>
   return __v < __lo ? __lo : __hi < __v ? __hi : __v;
 }
 
+//! @brief Internal version of clamp that works with values instead of references. Can be used with extended arithmetic
+//!        types. Should be preferred as it produces better optimized code with nvcc.
```
fbusato:
```suggestion
//!        types. Should be preferred as it produces better optimized code with nvcc (see nvbug 5455679)
```

libcudacxx/include/cuda/__tma/make_tma_descriptor.h
```diff
@@ -479,7 +479,7 @@ _CCCL_HOST_API inline __tma_box_sizes_array_t __get_box_sizes(
   size_t __total_size = 1;
   for (int __i = 0; __i < __rank; ++__i)
   {
-    const auto __max_box_size = static_cast<int>(::cuda::std::min(__tensor_sizes[__i], uint64_t{256}));
```
fbusato, line 482: `make_tma_descriptor` is host-only. It is not worth to update it

libcudacxx/include/cuda/std/__algorithm/min.h
```diff
@@ -55,6 +58,22 @@ template <class _Tp>
...
 
+//! @brief Internal version of min that works with values instead of references. Can be used with extended arithmetic
+//!        types. Should be preferred as it produces better optimized code with nvcc.
+template <class _Tp>
+[[nodiscard]] _CCCL_API constexpr _Tp __min(_Tp __a, _Tp __b) noexcept
+{
+  static_assert(__is_extended_arithmetic_v<_Tp>);
+  if constexpr (is_integral_v<_Tp>)
+  {
+    return (__b < __a) ? __b : __a;
```
fbusato, line 85: please use

```suggestion
    return ::min(__a, __b);
```
it produces better code for `__clamp`, see https://godbolt.org/z/xrP1WYc6a
Same for `__max`

libcudacxx/include/cuda/std/__format/format_integral.h
```diff
@@ -150,7 +150,7 @@ template <class _Tp, class _CharT, class _FmtCtx>
     __out_it                  = ::cuda::std::__fmt_copy(__array, __first, ::cuda::std::move(__out_it));
     __specs.__alignment_      = ::cuda::std::to_underlying(__fmt_spec_alignment::__right);
     __specs.__fill_.__data[0] = _CharT{'0'};
-    __specs.__width_ -= ::cuda::std::min(static_cast<uint32_t>(__last - __first), __specs.__width_);
+    __specs.__width_ -= ::cuda::std::__min(static_cast<uint32_t>(__last - __first), __specs.__width_);
```
fbusato, line 153: it looks that we never used `cuda::std::min/max` with floating-point. Probably, we directly used `fmin/fmax`.
question. Does make sense to add both paths in `__min/__max`?

fbusato (changes requested):

pauleonix:
At C++ Under the Sea I saw the talk [In Search of the Perfect Return: A C++ Programming Journey](https://youtu.be/sIZUg9tN4sk?si=i-03bytf5lT_Oin1&t=3263) by Dr. Walter Brown which might be of interest in this context (link skips to the most interesting part). He proposed a way to implement "perfect return" similar to the existing "perfect forwarding" method specifically for functions like `min` and `max`.  At the time I thought we might want to add this as e.g. `cuda::[min|max|clamp]` as always returning a reference (as per the standard) isn't only a performance issue but also introduces bugs.

@bernhardmgruber you were interested in this as well.

I guess this might warrant it's own issue though since this issue is rather about cases where we always want to return by value?

libcudacxx/include/cuda/__tma/make_tma_descriptor.h
miscco: We should always be consistent

libcudacxx/include/cuda/std/__algorithm/min.h
miscco: should be `::cuda::std::min`

davebayer: @miscco no, it must be `::min`, because `nvcc` provides these as intrinsics

miscco (commented):
I would like those to be public functions like `cuda::max`

libcudacxx/include/cuda/std/__format/format_integral.h
davebayer: I don't know, maybe not. @miscco what do you think?

davebayer:
Can we please provide an internal version, so I can use it in #3449 and then decide what steps do we want to take next?

bernhardmgruber:
> I would like those to be public functions like `cuda::max`

I discussed this already on #5516, users will be confused what the benefit of `cuda::max` is over `cuda::std::max`. We also advertise in conference talks etc. that entities in `cuda::` are extended versions of `cuda::std::` ones. The C++ standard defines the interface of `std::max`, so we know how `cuda::std::max` must look like and `cuda::max` should definitely follow the interface.

I think we are implementing a different function here with a new behavior and should just give it a dedicated name. I suggested `cuda::max_value` etc. in #5516 to hint at the return by value. 

> At C++ Under the Sea I saw the talk [In Search of the Perfect Return: A C++ Programming Journey](https://youtu.be/sIZUg9tN4sk?si=i-03bytf5lT_Oin1&t=3263) by Dr. Walter Brown ...

This is the essential slide:
image

And applied to `min`:
image

I have seen this before in a talk from @andralex. Maybe that is a better interface. But even then it should have a different name, since the name `max` is taken and specified by the C++ standard. I am even fine with `cuda::perfect_max` :D

miscco:
I am fine with any name you folks come up with ;)

andralex:
Haha, nice. I indeed mention the implementation at https://youtu.be/H3IdVM4xoCU?t=2621. Having these functions in a popular library would close a nice story arc.

I'd call them `safe_min` and `safe_max` because they eliminate the risk of dangling references.

davebayer:
But does this fix the problem with nvcc optimizations? If not, can we please focus on that? This is the purpose of this PR.

If you want to provide a public solution, please, open an issue or start a discussion on slack and agree on the design first. I will be happy to remove this function once we have the public alternative that has the same performance. Thank you :)

fbusato:
agree. Let's focus on the optimization issue first. Then, we can work on creating a public version.

libcudacxx/include/cuda/std/__algorithm/min.h
```diff
@@ -55,6 +60,36 @@ template <class _Tp>
...
+      NV_IF_TARGET(NV_IS_DEVICE, ({
+                     if constexpr (sizeof(_Tp) < sizeof(int))
+                     {
+                       using _Up = __make_nbit_int_t<32, is_signed_v<_Tp>>;
+                       return static_cast<_Tp>(::cuda::std::__min(static_cast<_Up>(__a), static_cast<_Up>(__b)));
+                     }
+                     else if constexpr (sizeof(_Tp) <= sizeof(long long))
+                     {
+                       return ::min(__a, __b);
+                     }
```
fbusato, line 82: I think we can merge both branches with
```
::min(__a, __b)
```

libcudacxx/include/cuda/std/__cccl/prologue.h
```diff
@@ -244,12 +244,24 @@
 #  define _CCCL_POP_MACRO_min
 #endif // defined(min)
 
+#if defined(__min)
```
fbusato, line 247: do you expect any user or library to define this macro?

davebayer: Our enemies are always one step ahead of us.. Windows defines it in <stdlib.h> 😅

libcudacxx/include/cuda/std/__algorithm/min.h
davebayer: I don't think it would work, there are missing specialisations for char and short

fbusato: it seems to work https://godbolt.org/z/GcMcne7oc

andralex:
@davebayer interesting, thanks very much. I looked around a bit, allow me to share a few findings and questions.

First, looking at https://nvbugspro.nvidia.com/bug/5455679 I wonder whether that's a red herring. Indeed the code generated at https://ce.nvidia.com/z/Y7nrxb for the out-of-line `clamp1` (the bad function taking reference parameters and returning by reference) is larger than the by-value version. But that's expected because, well, you need to dereference stuff so there's more work to do. The important point is that `clamp` will virtually always be inline, so what we should look at is `test1` and `test3`. Here the code should compare apples with apples, i.e. the bodies of clamp1 and clamp3 should be identical to isolate the matter of pass/return by reference. If you do change the body, you will notice indeed that the code generated for `test1` and `test3` is the same: https://ce.nvidia.com/z/TxdPEK

```
        VIMNMX R5, PT, PT, R4, R5, !PT                    ?WAIT4_END_GROUP
        VIMNMX R4, PT, PT, R5, R6, PT                     ?WAIT1
```

Now consider the code at https://ce.nvidia.com/z/fboe15, which uses the `safe_(min|max)` functions. The code generated for `test1` and `test3` is identical.

I'd posit that the code generated for the out-of-line clamp, min, and max is irrelevant because those small functions make sense if and only if inlined.

(It's quite possible that I misunderstood something or that I'm missing some context, so please disabuse me as appropriate. Thanks!)

andralex:
@davebayer @fbusato also looking at the reproducer (https://godbolt.org/z/GTEqznbTa) in the now closed https://github.com/NVIDIA/cccl/pull/5516, quite interestingly indeed the kernels using those inlined functions (which I'd argue is the only relevant place to look at) are different. The germane portions of code are:

```
std_clamp_kernel:
	@%p2 bra 	$L__BB4_3;
	setp.gt.s32 	%p3, %r2, %r6;
	shl.b64 	%rd8, %rd1, 2;
	add.s64 	%rd9, %rd3, %rd8;
	selp.b64 	%rd11, %rd4, %rd9, %p3;

cuda_clamp_kernel:
	setp.lt.s32 	%p2, %r8, %r2;
	min.s32 	%r9, %r8, %r3;
	selp.b32 	%r10, %r2, %r9, %p2;

temporary_clamp_kernel:
	max.s32 	%r9, %r8, %r2;
	min.s32 	%r10, %r9, %r3;

minmax_clamp_kernel:
	min.s32 	%r9, %r8, %r3;
	max.s32 	%r10, %r9, %r2;
```

In that code, replacing `minmax_clamp` with

```C++
template <class A, class B, class C>
__device__ decltype(auto) minmax_clamp(A&& value, B&& lo, C&& hi)
{
    return max(lo, min(value, hi));
}
```

produces the same optimal code.

fbusato:
yes, I did the same experiment. We could also introduce `safe_clamp` https://cuda.godbolt.org/z/Ecj38b3x3. It works perfectly.

one minor note about https://ce.nvidia.com/z/TxdPEK, `::min` and `::max` input/output are by-value, so `const T&` is a dangling reference.

andralex:
> yes, I did the same experiment. We could also introduce `safe_clamp` https://cuda.godbolt.org/z/Ecj38b3x3. It works perfectly.

Cool!

> one minor note about https://ce.nvidia.com/z/TxdPEK, `::min` and `::max` input/output are by-value, so `const T&` is a dangling reference.

I see, thanks.

Wondering whether we should close https://nvbugspro.nvidia.com/bug/5455679 as not a defect?

fbusato:
If the functions are inlined, I would prefer that NVVM optimizes all cases in the same way. Here the code with better names
```cpp
using T = int;

__device__ __forceinline__
const T& clamp_byref(const T& value, const T& low, const T& high) {
    return value < low ? low : high < value ? high : value; // ISETP, VIMNMX, IMAD
}

__device__ __forceinline__
T clamp_byvalue(T value, T low, T high) {
    return value < low ? low : high < value ? high : value; // ISETP, VIMNMX, IMAD
}

__device__ __forceinline__
T clamp_minmax(T value, T low, T high) {
    return ::min(::max(value, low), high);  // VIMNMX, VIMNMX
}

// .. use the output by-value
```
https://cuda.godbolt.org/z/jn4WP6qch

andralex:
One more thing before I forget... `safe_(min|max)` work with more types than `std::min` and `std::max`. Too many, in fact - they allow comparing signed with unsigned values, which is poor form. So I suggest adding a static assertion as in https://cuda.godbolt.org/z/oErhMoM18.

fbusato:
thanks for the suggestion! Technically, we could even use `cuda::std::cmq_less/greater` etc. (maybe I'm going too far)

bernhardmgruber:
> Wondering whether we should close https://nvbugspro.nvidia.com/bug/5455679 as not a defect?

@davebayer and @fbusato please have a quick look at the codegen of the latest nvcc. It LGTM. In light of this, do we still need the feature proposed in this PR?

Of course, older nvcc versions will still have worse codegen for `clamp` etc. Do we care enough to roll out the changes of this PR? If yes, we should document that codegen of `cuda::std::clamp` is fine one nvcc X.YZ.

fbusato:
Took a look. There is still one unneeded instruction that bothers me. Secondly, `std::min/max/clamp` are unsafe. I'm very in favor of following what Andrei suggested with `safe_*` API.

bernhardmgruber:
> There is still one unneeded instruction that bothers me. Secondly, `std::min/max/clamp` are unsafe. I'm very in favor of following what Andrei suggested with `safe_*` API.

Alright, that's fine for me then!

fbusato:
considering the recent discussion on nvbug 5455679, it is not worth modifying the compiler behavior regarding this functionality. @davebayer  the best option is to provide the `safe_` version of these functions

PR #7331, opened 2026-01-23
Fix double destroy in vector

miscco:
We were not resetting the `vector::m_size` member after destroying the old storage, only `vector::m_storage::m_size`. Then in `range_assign` we would call `m_storage.destroy(begin(), end());` but `end()` takes into account `vector::m_size` and not `vector::m_storage::m_size`.

Consequently we would try to delete the old elements again, even if the storage was already deallocated, passing a `nullptr` to `destroy_at`

This started to fail because with the rework of `allocator_traits` in another PR we are actually checking in `cuda::std::__destroy_at` whether the passed pointer is valid.

### DIFF
--- a/thrust/thrust/detail/vector_base.inl
+++ b/thrust/thrust/detail/vector_base.inl
@@ -165,8 +165,8 @@ vector_base<T, Alloc>& vector_base<T, Alloc>::operator=(const vector_base& v)
   {
     m_storage.destroy_on_allocator_mismatch(v.m_storage, begin(), end());
     m_storage.deallocate_on_allocator_mismatch(v.m_storage);
-
     m_storage.propagate_allocator(v.m_storage);
+    m_size = 0;
 
     assign(v.begin(), v.end());
   } // end if
--- a/thrust/thrust/detail/vector_base.inl
+++ b/thrust/thrust/detail/vector_base.inl
@@ -165,8 +165,14 @@ vector_base<T, Alloc>& vector_base<T, Alloc>::operator=(const vector_base& v)
   {
     m_storage.destroy_on_allocator_mismatch(v.m_storage, begin(), end());
     m_storage.deallocate_on_allocator_mismatch(v.m_storage);
+    if constexpr (thrust::detail::allocator_traits<Alloc>::propagate_on_container_copy_assignment::value)
+    {
+      if (m_storage.get_allocator() != v.m_storage.get_allocator())
+      {
+        m_size = 0;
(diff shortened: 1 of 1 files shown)


elstehle (approved):

davebayer (approved):

thrust/thrust/detail/vector_base.inl
```diff
@@ -165,8 +165,8 @@ vector_base<T, Alloc>& vector_base<T, Alloc>::operator=(const vector_base& v)
   {
     m_storage.destroy_on_allocator_mismatch(v.m_storage, begin(), end());
     m_storage.deallocate_on_allocator_mismatch(v.m_storage);
-
     m_storage.propagate_allocator(v.m_storage);
+    m_size = 0;
 
     assign(v.begin(), v.end());
```
bernhardmgruber, line 177: Critical: I think this is not correct. If the allocators do not propagate on assignment or both vectors have the same allocator, then
```
    m_storage.destroy_on_allocator_mismatch(v.m_storage, begin(), end());
    m_storage.deallocate_on_allocator_mismatch(v.m_storage);

    m_storage.propagate_allocator(v.m_storage);
```
do nothing. No deallocation, no resize, etc. Setting `m_size = 0` would effectively mark the vector empty but all elements are within their lifetime. This is certainly not correct.

bernhardmgruber (changes requested):

thrust/thrust/detail/vector_base.inl
miscco: You are correct, we need to check for `propagate_on_container_copy_assignment`

thrust/thrust/detail/vector_base.inl
```diff
@@ -165,8 +165,14 @@ vector_base<T, Alloc>& vector_base<T, Alloc>::operator=(const vector_base& v)
...
     m_storage.deallocate_on_allocator_mismatch(v.m_storage);
-
     m_storage.propagate_allocator(v.m_storage);
+    if constexpr (thrust::detail::allocator_traits<Alloc>::propagate_on_container_copy_assignment::value)
+    {
+      if (m_allocator != other.m_allocator)
+      {
+        m_size = 0;
+      }
+    }
```
bernhardmgruber, line 174: Since `m_storage.propagate_allocator` changes `this->m_allocator`, we need to move this check before it.

bernhardmgruber (approved):

PR #7334, opened 2026-01-23
Implement the new tuning API for `DeviceSegmentedReduce`

bernhardmgruber:
I started with this prompt to Cursor+codex:
```
Let's look at commit f5ddc3c, it introduces a refactoring to a new tuning API design for DispatchReduce.
Please apply this rewrite to DispatchSegmentedReduce as well.
Do not introduce any API breaking changes into the DispatchSegmentedReduce class.
Run the tests and fix any errors.
```
Except CCCL.C, all worked, but it was not particularly clean. I would say it got 85% there. I am still positively surprised.

I added a lot of refactoring and cleanup on top. And I debugged some segfaults in CCCL.C myself.

- [x] CUB tests pass
- [x] CCCL.C tests pass
- [x] No SASS diff for `cub.bench.segmented_reduce.sum.base` on SM75/SM100

### DIFF
--- a/cub/cub/device/dispatch/dispatch_segmented_reduce.cuh
+++ b/cub/cub/device/dispatch/dispatch_segmented_reduce.cuh
@@ -66,6 +66,23 @@ struct DeviceSegmentedReduceKernelSource
       InitT,
       AccumT>)
 };
+
+template <typename PolicyHub>
+struct policy_selector_from_hub
+{
+  // this is only called in device code, so we can ignore the arch parameter
+  _CCCL_DEVICE_API constexpr auto operator()(::cuda::arch_id /*arch*/) const -> segmented_reduce_policy
+  {
+    using p = typename PolicyHub::MaxPolicy::ActivePolicy::SegmentedReducePolicy;
+    return segmented_reduce_policy{{
+      p::BLOCK_THREADS,
+      p::ITEMS_PER_THREAD,
+      p::VECTOR_LOAD_LENGTH,
+      p::BLOCK_ALGORITHM,
+      p::LOAD_MODIFIER,
+    }};
+  }
+};
 } // namespace detail::segmented_reduce
 
 // TODO(bgruber): deprecate once we publish the tuning API
--- a/cub/cub/device/dispatch/dispatch_reduce.cuh
+++ b/cub/cub/device/dispatch/dispatch_reduce.cuh
@@ -88,6 +88,55 @@ struct DeviceReduceKernelSource
     return sizeof(AccumT);
   }
 };
+
+// TODO(bgruber): remove in CCCL 4.0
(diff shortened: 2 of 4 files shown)


cub/cub/device/dispatch/kernels/kernel_segmented_reduce.cuh
```diff
@@ -99,18 +102,32 @@ template <typename ChainedPolicyT,
...
-  BeginOffsetIteratorT d_begin_offsets,
-  EndOffsetIteratorT d_end_offsets,
-  ReductionOpT reduction_op,
-  InitT init)
+#if _CCCL_HAS_CONCEPTS()
+  requires segmented_reduce_policy_selector<PolicySelector>
+#endif // _CCCL_HAS_CONCEPTS()
+CUB_DETAIL_KERNEL_ATTRIBUTES __launch_bounds__(int(
+  PolicySelector{}(::cuda::arch_id{CUB_PTX_ARCH / 10})
+    .segmented_reduce.block_threads)) void DeviceSegmentedReduceKernel(InputIteratorT d_in,
```
miscco: That is some cursed formatting

cub/cub/device/dispatch/dispatch_segmented_reduce.cuh
```diff
@@ -25,12 +27,17 @@
 #include <cuda/std/__algorithm/min.h>
 #include <cuda/std/cstdint>
 #include <cuda/std/limits>
+#include <cuda/std/type_traits>
```
miscco: We should not pull in all of `<cuda/std/type_traits>` but only those headers we need

cub/cub/device/dispatch/dispatch_segmented_reduce.cuh
```diff
@@ -407,4 +413,154 @@ struct DispatchSegmentedReduce
...
+  if (const auto error = CubDebug(launcher_factory.PtxArchId(arch_id)))
+  {
+    return error;
+  }
+
+  const segmented_reduce_policy active_policy = policy_selector(arch_id);
+#if !_CCCL_COMPILER(NVRTC) && defined(CUB_DEBUG_LOG)
+  NV_IF_TARGET(
+    NV_IS_HOST,
+    (std::stringstream ss; ss << active_policy;
```
miscco: Nitpick: We should always fully qualify host library code in production

bernhardmgruber: Oh really, `::std::stringstream` as well here?

miscco: Big proponent of simple rules. Always is easier to follow than a list of bespoke exceptions

PR #7335, opened 2026-01-23
Formatters for `cuda::arch_id` and `cuda::compute_capability`

davebayer:
This PR specializes `std::formatter` for `cuda::arch_id` and `cuda::compute_capability`

### DIFF
--- a/libcudacxx/test/libcudacxx/cuda/ccclrt/device/arch_id_fmt.pass.cpp
+++ b/libcudacxx/test/libcudacxx/cuda/ccclrt/device/arch_id_fmt.pass.cpp
@@ -1,0 +1,60 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of libcu++, the C++ Standard Library for your entire system,
+// under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES.
+//
+//===----------------------------------------------------------------------===//
+
+#include <cuda/devices>
+
+#if __cpp_lib_format >= 201907L
+#  include <format>
+#endif // __cpp_lib_format >= 201907L
+
+#include "literal.h"
+
+#if __cpp_lib_format >= 201907L
+template <class C>
+void test()
+{
+  assert(std::format(TEST_STRLIT(C, "{}"), cuda::arch_id::sm_60) == TEST_STRLIT(C, "sm_60"));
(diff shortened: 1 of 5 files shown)


libcudacxx/test/libcudacxx/cuda/ccclrt/device/arch_id_fmt.pass.cpp
```diff
@@ -0,0 +1,60 @@
...
+  assert(std::format(TEST_STRLIT(C, "{}"), cuda::arch_id::sm_103) == TEST_STRLIT(C, "103"));
+  assert(std::format(TEST_STRLIT(C, "{}"), cuda::arch_id::sm_110) == TEST_STRLIT(C, "110"));
+  assert(std::format(TEST_STRLIT(C, "{}"), cuda::arch_id::sm_120) == TEST_STRLIT(C, "120"));
+  assert(std::format(TEST_STRLIT(C, "{}"), cuda::arch_id::sm_121) == TEST_STRLIT(C, "121"));
+  assert(std::format(TEST_STRLIT(C, "{}"), cuda::arch_id::sm_90a) == TEST_STRLIT(C, "90a"));
+  assert(std::format(TEST_STRLIT(C, "{}"), cuda::arch_id::sm_100a) == TEST_STRLIT(C, "100a"));
+  assert(std::format(TEST_STRLIT(C, "{}"), cuda::arch_id::sm_103a) == TEST_STRLIT(C, "103a"));
+  assert(std::format(TEST_STRLIT(C, "{}"), cuda::arch_id::sm_110a) == TEST_STRLIT(C, "110a"));
+  assert(std::format(TEST_STRLIT(C, "{}"), cuda::arch_id::sm_120a) == TEST_STRLIT(C, "120a"));
+  assert(std::format(TEST_STRLIT(C, "{}"), cuda::arch_id::sm_121a) == TEST_STRLIT(C, "121a"));
```
davebayer: @bernhardmgruber is formatting `cuda::arch_id` as `"NM[a]"` what you want? We could make it `"sm_NM[a]"`, too.

Or alternatively, we can make it part of the formatting options as `std::format("{:p}", arch_id)` no add the `sm_` prefix.

davebayer: Another option would be:
- `b` for binary architecture, the output would look as `sm_NM[a]`
- `v` for virtual architecture, the output would look as `compute_NM[a]`

davebayer: Maybe it would be better to keep it as is, you can always prepend the right prefix that works for your compiler.

bernhardmgruber: I would love if it was just spelled like the enum, so `sm_103a` for example :)

PR #7337, opened 2026-01-23
Ensure `device_find_if` works with non-default-constructible types

### LINKED ISSUE #7336: [BUG]: cub/agent/agent_find.cuh, line 106: error: no default constructor exists
### Is this a duplicate?

### Type of Bug

Compile-time Error

### Component

CUB

### Describe the bug

This test has started failing after this commit: c156e907e6b25b62e2b28b37d160fedefc967933
```
#include <execution>
#include <algorithm>
#include <vector>

template <typename T>
struct LocalWrapper
{
    explicit LocalWrapper(std::size_t k) : my_val(k) {}

  private:
    T my_val;
};

int32_t
main()
{
    typedef LocalWrapper<double> T;
    auto pred = [](const T& x) { return true; };
    const std::size_t max_n = 1000000;
    std::vector<T> in;
    in.reserve(max_n);
    const bool actual = std::is_partitioned(std::execution::par_unseq, in.begin(), in.begin() + 100, pred);
    return 0;
}
```
it fails with:
```
cub/agent/agent_find.cuh", line 106: error: no default constructor exists for class "LocalWrapper<TestUtils::float64_t>" (aka "LocalWrapper<double>")
      InputT input_items[items_per_thread];
             ^

```

### How to Reproduce

```
nvc++ -stdpar --c++17   test_is_partitioned.cpp;
```

### Expected behavior

Test should compile fine after the fix.

### Reproduction link

_No response_

### Operating System

_No response_

### nvidia-smi output

_No response_

### NVCC version

_No response_

miscco:
We load into a C-array of InputT, which requires initialization of its elements.

Fixes [BUG]: cub/agent/agent_find.cuh, line 106: error: no default constructor exists Fixes #7336

### DIFF
--- a/cub/test/catch2_test_device_find_if.cu
+++ b/cub/test/catch2_test_device_find_if.cu
@@ -255,14 +255,12 @@ C2H_TEST("Device find_if works with non default constructible types", "[device][
 
   CAPTURE(num_items, val_to_find);
 
-  const auto expected_if_found = cuda::std::min(static_cast<offset_t>(val_to_find), num_items);
-
   // counting_iterator input
   auto c_it = cuda::make_transform_iterator(cuda::make_counting_iterator(input_t{0}), converter{});
   {
     c2h::device_vector<offset_t> out_result(1, thrust::no_init);
     auto predicate = thrust::detail::equal_to_value<NotDefaultConstructible>{NotDefaultConstructible{val_to_find}};
     find_if(c_it, thrust::raw_pointer_cast(out_result.data()), predicate, num_items);
-    REQUIRE(expected_if_found == out_result[0]);
+    REQUIRE(val_to_find == out_result[0]);
   }
 }


cub/cub/agent/agent_find.cuh
```diff
@@ -51,7 +52,8 @@ struct agent_t
   // Can vectorize according to the policy if the input iterator is a native pointer to a primitive type
   static constexpr bool attempt_vectorization =
     (vector_load_length > 1) && (items_per_thread % vector_load_length == 0)
-    && (::cuda::std::contiguous_iterator<InputIteratorT>) && THRUST_NS_QUALIFIER::is_trivially_relocatable_v<InputT>;
+    && (::cuda::std::contiguous_iterator<InputIteratorT>) && THRUST_NS_QUALIFIER::is_trivially_relocatable_v<InputT> //
+    && ::cuda::std::is_default_constructible_v<InputT>; // We load into an C-array, which requires initialization
```
davebayer: Q: shouldn't we require trivial default constructability?

bernhardmgruber: Hmm, I wasn't sure if this is correct, but maybe it is. This is what the code does:
```c++
    InputT input_items[items_per_thread];
    auto* vec_items = reinterpret_cast<VectorT*>(input_items);
    for (int i = 0; i < number_of_vectors; ++i)
      vec_items[i] = d_vec_in[block_threads * i];
```
So `InputT` must be default constructible, but I can just memcpy over it (which is what load vectorization does in some sense). So `InputT` must be trivially destructible. But this is given by `is_trivially_relocatable_v`, so I think the check is actually fine.

However, it's not too hard to also implement support for types with default constructors:
```c++
    alignas(InputT) char input_items_raw[items_per_thread * sizeof(InputT)];
    auto* vec_items = reinterpret_cast<VectorT*>(input_items);
    for (int i = 0; i < number_of_vectors; ++i)
      vec_items[i] = d_vec_in[block_threads * i];
    InputT* input_tems = reinterpret_cast<InputT*>(input_items_raw);
```
That should be valid since `Inputt` is `is_trivially_relocatable_v` and it would include types which are not default constructible.

@miscco can we please add the second version instead?

miscco: I dont think that this is easily done, because we do access `input_items` below

miscco: yeah I moved to the suggestion

cub/cub/agent/agent_find.cuh
```diff
@@ -15,6 +15,7 @@
 #include <cuda/__memory/is_aligned.h>
 #include <cuda/atomic>
 #include <cuda/std/__type_traits/integral_constant.h>
+#include <cuda/std/__type_traits/is_default_constructible.h>
```
bernhardmgruber:
```suggestion
```

cub/test/catch2_test_device_find_if.cu
```diff
@@ -202,3 +202,67 @@ C2H_TEST("Device find_if works with non primitive iterator", "[device][find_if]"
...
+  }
+};
+
+C2H_TEST("Device find_if works with non default constructible types", "[device][find_if]")
+{
+  using input_t  = int32_t;
+  using offset_t = int32_t;
+
+  constexpr offset_t min_items = 1;
+  constexpr offset_t max_items = 10000000; // 10M items for reasonable test time
```
bernhardmgruber: Suggestion: this could be a lot lower, but should probably span more than one tile.
```suggestion
  constexpr offset_t max_items = 10'000;
```

cub/test/catch2_test_device_find_if.cu
```diff
@@ -202,3 +202,67 @@ C2H_TEST("Device find_if works with non primitive iterator", "[device][find_if]"
...
+
+  input_t val_to_find = static_cast<input_t>(GENERATE_COPY(take(1, random(min_items, max_items))));
```
bernhardmgruber: We can simplify this a bit by setting `val_to_find` to `num_items - 1` and expect it to always be found. Is there any advantage of randomizing whether the valid is in the input or not?

cub/test/catch2_test_device_find_if.cu
```diff
@@ -202,3 +202,65 @@ C2H_TEST("Device find_if works with non primitive iterator", "[device][find_if]"
     REQUIRE(-expected_if_found == out_result[0]);
   }
 }
+
+struct NotDefaultConstructible
+{
+  int value_;
+
+  __host__ __device__ constexpr explicit NotDefaultConstructible(int value = 0)
```
gonidelis, line 210: late chime in but (despite it's name) I think this type is `DefaultConstructible`, not `NotDefaultConstructible`. This compiles on my side:

```cpp

template <typename T>
void foo(T arg){
    T arr[10];
}

int main()
{
    NotDefaultConstructible ndc;
    func(ndc);

    return 0;
}
```

gonidelis: Param should not be default initialized. 

`  __host__ __device__ constexpr explicit NotDefaultConstructible(int value) {}`

cub/test/catch2_test_device_find_if.cu
```diff
@@ -202,3 +202,65 @@ C2H_TEST("Device find_if works with non primitive iterator", "[device][find_if]"
...
+    values({
+      min_items,
+      max_items,
+    }));
+  const input_t val_to_find = static_cast<input_t>(num_items - 1);
+
+  CAPTURE(num_items, val_to_find);
+
+  // counting_iterator input
+  auto c_it = cuda::make_transform_iterator(cuda::make_counting_iterator(input_t{0}), converter{});
```
gonidelis, line 259: Also this creates a non-raw iterator which takes the non vectorized path `ConsumeTile(..., ::cuda::std::integral_constant<bool, false> /*CAN_VECTORIZE*/)` which doesn't include my bugus code.

Pushing a fix

PR #7339, opened 2026-01-23
[libcu++] Add runtime check if memory pools are supported

pciolkosz:
We lack some runtime checks if memory pools are supported, which came up in some Tegra testing, which currently does not support managed or pinned memory pools.

This PR adds missing device attribute to check for host pool support and adds the missing checks in the tests. The tests were slightly refactored to avoid repetition. Runtime support checks were also added to memory pool implementation to provide more meaningful error message

### DIFF
--- a/libcudacxx/test/libcudacxx/cuda/memory_resource/resources/managed_memory_resource.cu
+++ b/libcudacxx/test/libcudacxx/cuda/memory_resource/resources/managed_memory_resource.cu
@@ -90,8 +90,8 @@ C2H_CCCLRT_TEST_LIST("managed_memory_resource allocation", "[memory_resource]", 
 {
   using managed_resource = TestType;
 #if _CCCL_CTK_AT_LEAST(13, 0)
-  if (cuda::std::is_same_v<managed_resource, cuda::managed_memory_pool_ref>
-      && !cuda::device_attributes::concurrent_managed_access(cuda::devices[0]))
+  if (!cuda::device_attributes::concurrent_managed_access(cuda::devices[0])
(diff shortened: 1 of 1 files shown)


miscco (approved):

libcudacxx/test/libcudacxx/cuda/memory_resource/resources/memory_pools.cu
```diff
@@ -639,6 +646,11 @@ C2H_CCCLRT_TEST("device_memory_pool with allocation handle", "[memory_resource]"
 #  if _CCCL_CTK_AT_LEAST(12, 6)
 C2H_CCCLRT_TEST("pinned_memory_pool with allocation handle", "[memory_resource]")
 {
+  if (!cuda::device_attributes::memory_pool_supported_handle_types(cuda::devices[0])
+      & cudaMemHandleTypePosixFileDescriptor)
```
bernhardmgruber, line 650: @pciolkosz clang gives me a warning here:
```
/home/coder/cccl/libcudacxx/test/libcudacxx/cuda/memory_resource/resources/memory_pools.cu:619:7: warning: logical not is only applied to the left hand side of this bitwise operator [-Wlogical-not-parentheses]
  619 |   if (!cuda::device_attributes::memory_pool_supported_handle_types(cuda::devices[0])
      |       ^
  620 |       & cudaMemHandleTypePosixFileDescriptor)
      |       ~
```
Did you mean to bitmask `cudaMemHandleTypePosixFileDescriptor` out of the supported types BEFORE negating them?

Issue #7317, opened 2026-01-21
Benchmarking `cuda.compute`


shwina:
We currently have some benchmarks for `cuda.compute` in https://github.com/NVIDIA/cccl/tree/main/python/cuda_cccl/benchmarks/compute.

We should

PR #7341, opened 2026-01-23
bench: Migrate Python cuda.compute benchmarks to nvbench

danielfrg:
## Description

closes https://github.com/NVIDIA/cccl/issues/7317

Migrating the Python `cuda.compute` benchmarks from pytest to nvbench.
- Adds equivalent C++ benchmarks to compare Python vs C++


danielfrg:
@shwina I migrated one benchmark to nvbench here and added C++ equivalent. If you can review and see if this is what you were thinking i can continue with the rest.

I added some instructions and doing the C++ part with cmake but let me know if you have other ideas on how to make that part better/easier.

On my dev GPU (4090) I got:

```
# bench_unary_transform_pointer

GPU Time: Mean GPU execution time (cold start, pure kernel)
  CUDA events (nvbench tag: nv/cold/time/gpu/mean)
CPU Time: Mean CPU (host) latency
  Host clock (nvbench tag: nv/cold/time/cpu/mean)

## [0] NVIDIA GeForce RTX 4090

| Elements   | C++ GPU    | Py GPU     | % Slower   | C++ CPU    | Py CPU     | CPU Ovhd   |
|------------|------------|------------|------------|------------|------------|------------|
| 2^1        | 3.097 us   | 3.392 us   | 9.51%      | 15.209 us  | 22.101 us  | 6.892 us   |
| 2^10       | 3.126 us   | 4.525 us   | 44.72%     | 15.202 us  | 23.571 us  | 8.369 us   |
| 2^14       | 3.307 us   | 3.607 us   | 9.07%      | 15.399 us  | 22.187 us  | 6.788 us   |
| 2^17       | 4.902 us   | 6.229 us   | 27.08%     | 16.972 us  | 25.270 us  | 8.298 us   |
| 2^20       | 12.139 us  | 12.571 us  | 3.56%      | 24.200 us  | 30.918 us  | 6.718 us   |
| 2^24       | 149.988 us | 150.786 us | 0.53%      | 162.432 us | 169.373 us | 6.941 us   |
| 2^26       | 586.332 us | 584.699 us | -0.28%     | 598.789 us | 603.301 us | 4.511 us   |
```

shwina:
Thanks @danielfrg - I think this is a really great start!

With regards to comparing C++ and Python benchmarks, I wonder if we can't use the existing `nvbench_compare` script - https://github.com/NVIDIA/nvbench/blob/main/scripts/nvbench_compare.py? IIRC @NaderAlAwar has had some previous experience doing this, so it would be great to have his feedback as well.

python/cuda_cccl/benchmarks/compute/nvbench/nvbench_transform.py
```diff
@@ -0,0 +1,66 @@
...
+        exec_stream = as_cupy_stream(launch.get_stream())
+        with exec_stream:
+            transform(inp, out, size)
+
+    state.exec(launcher)
+
+
+if __name__ == "__main__":
+    b1 = bench.register(bench_unary_transform_pointer)
+    b1.add_int64_power_of_two_axis("Elements", [1, 10, 14, 17, 20, 24, 26])
```
NaderAlAwar: Suggestion:
```suggestion
    b1.add_int64_power_of_two_axis("Elements", range(12, 29, 4))
```

python/cuda_cccl/benchmarks/compute/nvbench/nvbench_transform.py
```diff
@@ -0,0 +1,66 @@
+# Copyright (c) 2025, NVIDIA CORPORATION & AFFILIATES. ALL RIGHTS RESERVED.
```
NaderAlAwar: Suggestion:

```suggestion
# Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. ALL RIGHTS RESERVED.
```

danielfrg:
1. Agree, the nvbench scripts are here at the moment for convenience, if you are able to expose them in the wheels then we can remove. Those work great but I think the python vs c++ one has a but more convenient reports for this case in general, the summary table I have on my first comment is from that script.
2. Makes sense on trying to reuse the existing benchmarks, for this example one I was just converting the existing python one from pytest. Do we want to convert the ones we have instead? Lets say starting with this one: https://github.com/NVIDIA/cccl/blob/main/cub/benchmarks/bench/transform/fill.cu ?

NaderAlAwar:
I see now, yeah the python vs cpp summary script should stay. I think it would be more valuable to convert the existing C++ ones, and yes `fill.cu` seems like a good start.

NaderAlAwar:
Also, take a look at https://github.com/NVIDIA/cccl/blob/main/CONTRIBUTING.md#code-formatting-pre-commit-hooks to fix the pre-commit ci issue

danielfrg:
I updated the one benchmark here to be based on `fill.cu` and I got this that in general it looks good to me: For large data sizes we between the 5% target:

```
…/compute/nvbench cuda-compute-nvbench ? ❯ python analysis/python_vs_cpp_summary.py results/fill_py.json results/fill_cpp.json
# bench_fill

GPU Time: Mean GPU execution time (cold start, pure kernel)
  CUDA events (nvbench tag: nv/cold/time/gpu/mean)
CPU Time: Mean CPU (host) latency
  Host clock (nvbench tag: nv/cold/time/cpu/mean)

## [0] NVIDIA GeForce RTX 4090

### Type: I16

| Elements   | C++ GPU    | Py GPU     | % Slower   | C++ CPU    | Py CPU     | CPU Ovhd   |
|------------|------------|------------|------------|------------|------------|------------|
| 2^16       | 2.754 us   | 4.278 us   | 55.35%     | 14.796 us  | 23.269 us  | 8.474 us   |
| 2^20       | 3.532 us   | 4.972 us   | 40.76%     | 15.715 us  | 23.905 us  | 8.190 us   |
| 2^24       | 34.311 us  | 40.369 us  | 17.65%     | 46.510 us  | 58.837 us  | 12.327 us  |
| 2^28       | 558.510 us | 565.441 us | 1.24%      | 570.950 us | 584.637 us | 13.687 us  |
| 2^32       | 8.952 ms   | 8.958 ms   | 0.06%      | 8.964 ms   | 8.978 ms   | 13.978 us  |

### Type: I32

| Elements   | C++ GPU   | Py GPU    | % Slower   | C++ CPU   | Py CPU    | CPU Ovhd   |
|------------|-----------|-----------|------------|-----------|-----------|------------|
| 2^16       | 2.861 us  | 3.365 us  | 17.64%     | 14.940 us | 21.710 us | 6.770 us   |
| 2^20       | 4.511 us  | 4.837 us  | 7.24%      | 16.391 us | 23.517 us | 7.127 us   |
| 2^24       | 69.059 us | 67.794 us | -1.83%     | 81.073 us | 85.855 us | 4.782 us   |
| 2^28       | 1.118 ms  | 1.120 ms  | 0.21%      | 1.130 ms  | 1.139 ms  | 8.893 us   |

### Type: I64

| Elements   | C++ GPU    | Py GPU     | % Slower   | C++ CPU    | Py CPU     | CPU Ovhd   |
|------------|------------|------------|------------|------------|------------|------------|
| 2^16       | 2.881 us   | 4.293 us   | 48.99%     | 14.974 us  | 23.148 us  | 8.174 us   |
| 2^20       | 8.536 us   | 9.735 us   | 14.04%     | 20.425 us  | 28.462 us  | 8.037 us   |
| 2^24       | 140.137 us | 140.560 us | 0.30%      | 152.051 us | 159.512 us | 7.460 us   |
| 2^28       | 2.239 ms   | 2.239 ms   | -0.00%     | 2.252 ms   | 2.259 ms   | 7.814 us   |

### Type: I8

| Elements   | C++ GPU    | Py GPU     | % Slower   | C++ CPU    | Py CPU     | CPU Ovhd   |
|------------|------------|------------|------------|------------|------------|------------|
| 2^16       | 2.727 us   | 3.350 us   | 22.85%     | 14.533 us  | 21.981 us  | 7.448 us   |
| 2^20       | 3.160 us   | 3.842 us   | 21.59%     | 15.063 us  | 22.308 us  | 7.245 us   |
| 2^24       | 17.222 us  | 22.981 us  | 33.45%     | 29.363 us  | 41.180 us  | 11.817 us  |
| 2^28       | 278.926 us | 285.592 us | 2.39%      | 291.525 us | 304.250 us | 12.725 us  |
| 2^32       | 4.476 ms   | 4.482 ms   | 0.14%      | 4.488 ms   | 4.502 ms   | 13.767 us  |
```

I had claude to analyze the APIs and give me a mapping for the transform benchmarks:

| CUB Benchmark | CUB API | Python API | Status | Notes |
|---------------|---------|------------|--------|-------|
| `transform/fill.cu` | `cub::detail::transform::dispatch` with `return_constant` | `unary_transform` + `ConstantIterator` | ✅ **DONE** | Using ConstantIterator(42) as 0-ary generator |
| `transform/babelstream.cu` | `cub::detail::transform::dispatch` (mul, add, triad, nstream) | `unary_transform`, `binary_transform` | ✅ **Ready** | Multiple operations: a*scalar, a+b, a+b*scalar, a+b+c*scalar |
| `transform/heavy.cu` | `cub::detail::transform::dispatch` with heavy compute | `unary_transform` | ✅ **Ready** | Custom function with configurable compute intensity |
| `transform/fib.cu` | `cub::detail::transform::dispatch` | `unary_transform` | ✅ **Ready** | Fibonacci computation |
| `transform/grayscale.cu` | `cub::detail::transform::dispatch` with struct input | `unary_transform` + `@gpu_struct` | ✅ **Ready** | RGB struct → grayscale, need gpu_struct for RGB type |
| `transform/complex_cmp.cu` | `cub::detail::transform::dispatch` | `unary_transform` | ✅ **Ready** | Complex comparison logic |
| `transform_reduce/sum.cu` | `cub::detail::reduce::dispatch` with transform | `reduce_into` with `TransformIterator` | ⚠️ **Workaround** | Need to use TransformIterator or pre-transform |

They look good to me on a quick look. Let me know what you guys prefer in terms of PRs, do we want to keep them all in a single PR? If the structure looks good now I will start with more transform benchmarks.

python/cuda_cccl/benchmarks/compute/nvbench/nvbench_fill.py
```diff
@@ -0,0 +1,80 @@
...
+    # Match C++ metrics:
+    state.add_element_count(size)
+    state.add_global_memory_reads(0)
+    state.add_global_memory_writes(size * out.dtype.itemsize)
+
+    # Execute benchmark
+    def launcher(launch: bench.Launch):
+        exec_stream = as_cupy_stream(launch.get_stream())
+        with exec_stream:
+            transform(constant_it, out, size)
```
NaderAlAwar: Critical: we cannot use a cupy context manager for our kernels. Instead, we need to pass in a stream object that implements the cuda stream protocol as the stream argument. The stream type in pynvbench should support it https://github.com/NVIDIA/nvbench/blob/a681e2185d57a1af9aefa59b25132f6067c20684/python/cuda/bench/__init__.pyi#L38, so this should work instead


```suggestion
        transform(constant_it, out, size, stream=launch.get_stream())
```

python/cuda_cccl/benchmarks/compute/nvbench/CMakeLists.txt
NaderAlAwar, line 1: Important: instead of building the CUB benchmarks this way, we should be just building them from the repo using the ci scripts. So this would be `ci/build_cub.sh -arch 89`.

I think we should just add a bash script here, maybe call it `run_python_vs_cpp.sh`. It should build the benchmarks using the ci script and should include a list of benchmarks that we currently support in cuda.compute (only fill for now), and then automatically run both cpp and python versions, and compare them with `python_vs_cpp_summary.py`.

danielfrg:
Changed the C++ build script to use the one in the ci folder and added a single script to run the benchmarks.

I added two extra benchmarks to show how that worked. The `babelstream` one looks ok in my GPU but the `transform_heavy` one the performance has big differences:

```
# heavy

GPU Time: Mean GPU execution time (cold start, pure kernel)
  CUDA events (nvbench tag: nv/cold/time/gpu/mean)
CPU Time: Mean CPU (host) latency
  Host clock (nvbench tag: nv/cold/time/cpu/mean)

## [0] NVIDIA GeForce RTX 4090

### Heaviness=128

| Elements   | C++ GPU    | Py GPU     | % Slower   | C++ CPU    | Py CPU     | CPU Ovhd   |
|------------|------------|------------|------------|------------|------------|------------|
| 2^16       | 6.171 us   | 12.469 us  | 102.07%    | 18.244 us  | 28.769 us  | 10.525 us  |
| 2^20       | 48.865 us  | 145.804 us | 198.38%    | 60.683 us  | 162.066 us | 101.383 us |
| 2^24       | 613.619 us | 2.187 ms   | 256.37%    | 626.126 us | 2.204 ms   | 1.578 ms   |
| 2^28       | 9.894 ms   | 34.658 ms  | 250.30%    | 9.906 ms   | 34.677 ms  | 24.770 ms  |

### Heaviness=256

| Elements   | C++ GPU    | Py GPU     | % Slower   | C++ CPU    | Py CPU     | CPU Ovhd   |
|------------|------------|------------|------------|------------|------------|------------|
| 2^16       | 9.700 us   | 384.247 us | 3861.21%   | 21.617 us  | 400.758 us | 379.141 us |
| 2^20       | 105.618 us | 9.129 ms   | 8543.15%   | 117.345 us | 9.147 ms   | 9.029 ms   |
| 2^24       | 1.541 ms   | 149.250 ms | 9586.40%   | 1.553 ms   | 149.271 ms | 147.718 ms |
| 2^28       | 24.399 ms  | 2.416 s    | 9800.04%   | 24.412 ms  | 2.416 s    | 2.391 s    |

### Heaviness=32

| Elements   | C++ GPU    | Py GPU     | % Slower   | C++ CPU    | Py CPU     | CPU Ovhd   |
|------------|------------|------------|------------|------------|------------|------------|
| 2^16       | 4.316 us   | 5.651 us   | 30.94%     | 16.286 us  | 22.287 us  | 6.001 us   |
| 2^20       | 15.354 us  | 36.968 us  | 140.77%    | 27.240 us  | 53.072 us  | 25.832 us  |
| 2^24       | 174.735 us | 495.926 us | 183.82%    | 186.898 us | 512.593 us | 325.695 us |
| 2^28       | 2.889 ms   | 7.961 ms   | 175.57%    | 2.902 ms   | 7.979 ms   | 5.077 ms   |

### Heaviness=64

| Elements   | C++ GPU    | Py GPU    | % Slower   | C++ CPU    | Py CPU    | CPU Ovhd   |
|------------|------------|-----------|------------|------------|-----------|------------|
| 2^16       | 4.989 us   | 7.757 us  | 55.49%     | 16.932 us  | 24.156 us | 7.224 us   |
| 2^20       | 25.499 us  | 76.779 us | 201.10%    | 37.433 us  | 92.815 us | 55.382 us  |
| 2^24       | 317.748 us | 1.005 ms  | 216.32%    | 329.875 us | 1.022 ms  | 692.030 us |
| 2^28       | 5.204 ms   | 15.931 ms | 206.15%    | 5.216 ms   | 15.949 ms | 10.733 ms  |


Comparison saved to: /home/danielfrg/Documents/cccl/python/cuda_cccl/benchmarks/compute/nvbench/results/heavy_comparison.txt
```

Any ideas on why this might be happening? My guess would be mostly on the numba kernels or compilation of those?

python/cuda_cccl/benchmarks/compute/environment.yml
```diff
@@ -0,0 +1,26 @@
...
+  # nvbench
+  - cmake
+  - ninja
+  - fmt
+
+  - pip
+  - pip:
+    - cuda-bench[cu12]>=0.2.0
+    - cuda-cccl[cu12]
+    - cupy-cuda12x
```
NaderAlAwar: Question: is there a way to not hardcode this to cu12?

NaderAlAwar:
pre-commit.ci autofix

NaderAlAwar:
>Any ideas on why this might be happening? My guess would be mostly on the numba kernels or compilation of those?

These results are interesting, we should definitely look into them. I would assume it is something in numba, but I would have to look at the SASS to say for sure. In any case, this should not block the PR. Can you create an issue noting this? Lets maybe have an EPIC issue comparing cuda.compute performance to CUB, and have this specific instance be a subissue

python/cuda_cccl/benchmarks/compute/environment.yml
danielfrg: Not 100% (i think) my plan was to have another `environment.yml` for cuda 13.

danielfrg:
Agree it should not block this one. I will create an epic issue to analyse the results.
I'll continue adding the other transform benchmarks on this PR unless you all prefer something else.

NaderAlAwar:
It's up to you. We can merge this or you can add the other benchmarks first

danielfrg:
Ok, then I will add a couple more here!

python/cuda_cccl/benchmarks/compute/analysis/migration_summary.md
shwina, line 1: Do we need to keep this in the repo?

python/cuda_cccl/benchmarks/compute/MIGRATION_STATUS.md
shwina, line 1: I think this is better tracked in an issue rather than in the source code itself

python/cuda_cccl/benchmarks/compute/analysis/migration_summary.md
danielfrg: No, this is just for me to keep track, i will delete eventually

python/cuda_cccl/benchmarks/compute/MIGRATION_STATUS.md
danielfrg: Agree, its AI generated. Just so i can keep track. I will delete eventually.

danielfrg:
This one is ready for some reviews. All the benchmarks run and most are inside the error tolerance we defined when I ran it on my 4090. Some that are not: `segmented_reduce/custom`, segmented_reduce/sum`, `segmented_sort/keys`, `transform/heavy`. 

Attached are the results of my run, if you open server it with `python -m http.server` you can see a dashboard that loads the output json files.

[results.tar.gz](https://github.com/user-attachments/files/25400787/results.tar.gz)

python/cuda_cccl/benchmarks/compute/histogram/even.py
```diff
@@ -0,0 +1,237 @@
+# Copyright (c) 2025, NVIDIA CORPORATION & AFFILIATES. ALL RIGHTS RESERVED.
```
NaderAlAwar: Important: should be 2026 for all new files (applies to other files as well)

python/cuda_cccl/benchmarks/compute/histogram/even.py
```diff
@@ -0,0 +1,237 @@
...
+# Type mapping: match C++ sample_types
+# Note: C++ uses int8_t, int16_t, etc. but Python histogram needs compatible types
+TYPE_MAP = {
+    "I8": np.int8,
+    "I16": np.int16,
+    "I32": np.int32,
+    "I64": np.int64,
+    "F32": np.float32,
+    "F64": np.float64,
+}
```
NaderAlAwar: Important: if this mapping is common across benchmarks, I would move it to `utils.py`

python/cuda_cccl/benchmarks/compute/histogram/even.py
```diff
@@ -0,0 +1,237 @@
...
+                    raw + float(lower_level), float(lower_level), float(upper_level)
+                ).astype(dtype)
+
+    return samples
+
+
+def bench_histogram_even(state: bench.State):
+    """
+    Benchmark histogram_even operation.
+    """
```
NaderAlAwar: Nit: this docstring feels unnecessary, lets remove it (same applies to other algorithms)

python/cuda_cccl/benchmarks/compute/histogram/even.py
```diff
@@ -0,0 +1,237 @@
...
+    # WORKAROUND: Clear caches to avoid bug where cached histogram objects
+    # are reused incorrectly when bin count changes. The cache key for
+    # np.ndarray only considers dtype, not shape/values, so h_num_output_levels
+    # arrays with different values but same dtype get the same cache key.
+    # This causes a histogram built for 32 bins to be reused for 2048 bins,
+    # leading to memory corruption.
+    # TODO: Fix in cuda.compute._caching to include array values in cache key
+    clear_all_caches()
+
+    # Get parameters from axes
```
NaderAlAwar: Nit: this comment and others below, like "Setup histogram parameters", "Allocate arrays", and others are unnecessary. We should remove them (same applies to other files)

python/cuda_cccl/benchmarks/compute/histogram/even.py
```diff
@@ -0,0 +1,237 @@
...
+    Benchmark histogram_even operation.
+    """
+    # WORKAROUND: Clear caches to avoid bug where cached histogram objects
+    # are reused incorrectly when bin count changes. The cache key for
+    # np.ndarray only considers dtype, not shape/values, so h_num_output_levels
+    # arrays with different values but same dtype get the same cache key.
+    # This causes a histogram built for 32 bins to be reused for 2048 bins,
+    # leading to memory corruption.
+    # TODO: Fix in cuda.compute._caching to include array values in cache key
+    clear_all_caches()
```
NaderAlAwar: Important: I believe we fixed this in #7657, could you remove the workaround and see if it works now?

python/cuda_cccl/benchmarks/compute/histogram/even.py
```diff
@@ -0,0 +1,237 @@
...
+    h_upper_level = np.array([upper_level], dtype=dtype)
+
+    # Build histogram operation
+    histogrammer = make_histogram_even(
+        d_samples=d_samples,
+        d_histogram=d_histogram,
+        h_num_output_levels=h_num_output_levels,
+        h_lower_level=h_lower_level,
+        h_upper_level=h_upper_level,
+        num_samples=num_elements,
```
NaderAlAwar: Nit: I am not a fan of using keyword arguments, let us remove them here and in other locations in this file (and in other files as well where they don't improve readability)

python/cuda_cccl/benchmarks/compute/histogram/even.py
```diff
@@ -0,0 +1,237 @@
...
+    d_samples = generate_samples_with_entropy(
+        num_elements, dtype, lower_level, upper_level, entropy_str, alloc_stream
+    )
+
+    # Output histogram (counter type is int32 in C++)
+    with alloc_stream:
+        d_histogram = cp.zeros(num_bins, dtype=np.int32)
+
+    # Synchronize to ensure data is ready
+    alloc_stream.synchronize()
```
NaderAlAwar: Important: this should not be necessary

python/cuda_cccl/benchmarks/compute/histogram/even.py
```diff
@@ -0,0 +1,237 @@
...
+            d_histogram=d_histogram,
+            h_num_output_levels=h_num_output_levels,
+            h_lower_level=h_lower_level,
+            h_upper_level=h_upper_level,
+            num_samples=num_elements,
+        )
+        cp.cuda.Device().synchronize()
+    except Exception as e:
+        state.skip(f"CUDA error during warmup: {e}")
+        return
```
NaderAlAwar: Important: I've been debating whether to keep this warmup run since it doesn't exist in the C++ benchmarks. For consistency, I would remove it (here and in other files)

python/cuda_cccl/benchmarks/compute/histogram/even.py
```diff
@@ -0,0 +1,237 @@
...
+            d_samples=d_samples,
+            d_histogram=d_histogram,
+            h_num_output_levels=h_num_output_levels,
+            h_lower_level=h_lower_level,
+            h_upper_level=h_upper_level,
+            num_samples=num_elements,
+            stream=launch.get_stream(),
+        )
+
+    state.exec(launcher)
```
NaderAlAwar: Important: you need to add `batched=False` to match the `no_batch` in C++. Please check the other benchmarks for similar behavior as well.

python/cuda_cccl/benchmarks/compute/histogram/even.py
```diff
@@ -0,0 +1,237 @@
...
+
+if __name__ == "__main__":
+    b = bench.register(bench_histogram_even)
+    b.set_name("base")  # Match C++ benchmark name
+
+    # Match C++ axes
+    b.add_string_axis("SampleT", list(TYPE_MAP.keys()))
+    b.add_int64_power_of_two_axis("Elements", range(16, 29, 4))  # [16, 20, 24, 28]
+    b.add_int64_axis("Bins", BINS_VALUES)  # [32, 128, 2048, 2097152]
+    b.add_string_axis("Entropy", ENTROPY_VALUES)
```
NaderAlAwar: Nit: I would inline the definition of `BINS_VALUES` and `ENTROPY_VALUES` here instead of defining them as global variables since they are note used anywhere else

python/cuda_cccl/benchmarks/compute/histogram/even.py
```diff
@@ -0,0 +1,237 @@
...
+            state.skip("Number of bins exceeds what SampleT can represent")
+            return
+
+    # Skip problematic configurations that cause cudaErrorIllegalAddress
+    # I8/I16 with large bin counts cause memory access issues
+    if dtype in (np.int8, np.int16) and num_bins >= 2048:
+        state.skip(
+            "Small integer types with large bin counts cause illegal memory access"
+        )
+        return
```
NaderAlAwar: Critical: if this error occurs only in Python, please file an issue so we can investigate

python/cuda_cccl/benchmarks/compute/histogram/even.py
```diff
@@ -0,0 +1,237 @@
...
+
+"""
+Python benchmark for histogram_even using cuda.compute.
+
+C++ equivalent: cub/benchmarks/bench/histogram/even.cu
+
+Notes:
+- The C++ benchmark uses Entropy axis to generate data with different bit distributions
+- For Python, we approximate this with random data in the appropriate range
+- Entropy 1.0 = uniform random, 0.201 = skewed towards lower values
```
NaderAlAwar: Important: we should try to match the C++ generation process more closely. According to gpt-5.3-codex, the C++ implementation does:

* Start with random values uniformly distributed over the valid range.
* For lower entropy, generate another uniform random array and do elementwise AND.
* Repeat k times (k depends on entropy level).
* More AND passes zero out more bits, so values collapse toward fewer patterns / smaller magnitudes.

Which should be doable in python. Could you verify if the above is true and try to implement something similar?

python/cuda_cccl/benchmarks/compute/histogram/even.py
```diff
@@ -0,0 +1,237 @@
...
+    state.exec(launcher)
+
+
+if __name__ == "__main__":
+    b = bench.register(bench_histogram_even)
+    b.set_name("base")  # Match C++ benchmark name
+
+    # Match C++ axes
+    b.add_string_axis("SampleT", list(TYPE_MAP.keys()))
+    b.add_int64_power_of_two_axis("Elements", range(16, 29, 4))  # [16, 20, 24, 28]
```
NaderAlAwar: Important: match the C++ version more closely
```suggestion
    b.add_int64_power_of_two_axis("Elements{io}", range(16, 29, 4))  # [16, 20, 24, 28]
```

python/cuda_cccl/benchmarks/compute/histogram/even.py
danielfrg: done

python/cuda_cccl/benchmarks/compute/histogram/even.py
danielfrg: done

python/cuda_cccl/benchmarks/compute/histogram/even.py
danielfrg: removed

python/cuda_cccl/benchmarks/compute/histogram/even.py
danielfrg: Yes, removed those.

python/cuda_cccl/benchmarks/compute/histogram/even.py
danielfrg: done

python/cuda_cccl/benchmarks/compute/histogram/even.py
danielfrg: done

python/cuda_cccl/benchmarks/compute/histogram/even.py
danielfrg: Look like this is not needed anymore

python/cuda_cccl/benchmarks/compute/histogram/even.py
danielfrg: Removed

python/cuda_cccl/benchmarks/compute/histogram/even.py
danielfrg: yes, made some changes to match this behaviour.

danielfrg:
@NaderAlAwar I think this is ready for you to take another look, i think everything is addressed

python/cuda_cccl/benchmarks/compute/histogram/even.py
```diff
@@ -0,0 +1,140 @@
...
+            state.skip("Number of bins exceeds what SampleT can represent")
+            return
+
+    if (
+        dtype in (np.float32, np.float64)
+        and num_bins >= 2048
+        and entropy_str == "0.201"
+    ):
+        state.skip("Skipping: Generates cudaErrorIllegalAddress")
+        return
```
NaderAlAwar: Critical: This is likely a bug in `cuda.compute`. We should not be skipping this here because we do not skip them in the C++ benchmark. Please file an issue with a reproducer so we can fix this.

python/cuda_cccl/benchmarks/compute/utils.py
```diff
@@ -0,0 +1,248 @@
+import cupy as cp
+import numpy as np
+
+import cuda.bench as bench
+
+TYPE_MAP = {
```
NaderAlAwar: Suggestion: for consistency with the below names, I would name this `ALL_TYPES`

python/cuda_cccl/benchmarks/compute/merge_sort/keys.py
```diff
@@ -0,0 +1,104 @@
...
+        )
+
+    state.exec(launcher, batched=False)
+
+
+if __name__ == "__main__":
+    b = bench.register(bench_merge_sort_keys)
+    b.set_name("base")
+
+    b.add_string_axis("T", list(TYPE_MAP.keys()))
```
NaderAlAwar: Suggestion: for consistency with C++

```suggestion
    b.add_string_axis("KeyT", list(TYPE_MAP.keys()))
```

python/cuda_cccl/benchmarks/compute/merge_sort/keys.py
```diff
@@ -0,0 +1,104 @@
...
+
+def bench_merge_sort_keys(state: bench.State):
+    type_str = state.get_string("T")
+    dtype = TYPE_MAP[type_str]
+    num_elements = int(state.get_int64("Elements{io}"))
+    entropy_str = state.get_string("Entropy")
+
+    if dtype.itemsize == 8 and num_elements >= 2**28:
+        state.skip("Skipping: large 64-bit key sizes hit allocator limits")
+        return
```
NaderAlAwar: Question: is this an allocator limit set by cupy? If so, we should find some way around it. This should be possible. Same applies to `pairs.cu` (and any other benchmark)

python/cuda_cccl/benchmarks/compute/histogram/even.py
```diff
@@ -0,0 +1,140 @@
...
+    h_lower_level = np.array([lower_level], dtype=dtype)
+    h_upper_level = np.array([upper_level], dtype=dtype)
+
+    histogrammer = make_histogram_even(
+        d_samples=d_samples,
+        d_histogram=d_histogram,
+        h_num_output_levels=h_num_output_levels,
+        h_lower_level=h_lower_level,
+        h_upper_level=h_upper_level,
+        num_samples=num_elements,
```
NaderAlAwar: Nit: for this and the calls to `histogrammer()` below, I would remove the keyword arguments, as they do not add much. Same applies to other benchmarks

python/cuda_cccl/benchmarks/compute/merge_sort/keys.py
```diff
@@ -0,0 +1,104 @@
...
+        d_in_keys=d_in_keys,
+        d_in_items=None,
+        d_out_keys=d_out_keys,
+        d_out_items=None,
+        op=OpKind.LESS,
+        num_items=num_elements,
+    )
+    if temp_storage_bytes <= 0:
+        state.skip("Skipping: invalid temp storage size")
+        return
```
NaderAlAwar: Important: I don't see this check in the C++ benchmark so I prefer to remove it. If there are cases where `temp_storage_bytes` is returning an invalid value we should file an issue. Same issue for other benchmarks

python/cuda_cccl/benchmarks/compute/merge_sort/keys.py
```diff
@@ -0,0 +1,104 @@
...
+    with alloc_stream:
+        temp_storage = cp.empty(temp_storage_bytes, dtype=np.uint8)
+
+    state.add_element_count(num_elements)
+    state.add_global_memory_reads(num_elements * d_in_keys.dtype.itemsize)
```
NaderAlAwar: Suggestion: for consistency with C++
```suggestion
    state.add_global_memory_reads(num_elements * d_in_keys.dtype.itemsizem, "Size")
```

python/cuda_cccl/benchmarks/compute/merge_sort/pairs.py
```diff
@@ -0,0 +1,123 @@
...
+    with alloc_stream:
+        if np.issubdtype(value_dtype, np.integer):
+            info = np.iinfo(value_dtype)
+            d_in_values = cp.random.randint(
+                int(info.min), int(info.max) + 1, size=num_elements, dtype=np.int64
+            ).astype(value_dtype)
+        else:
+            d_in_values = cp.random.uniform(-1, 1, size=num_elements).astype(
+                value_dtype
+            )
```
NaderAlAwar: Important: in the C++ benchmark, the the values are initialized to 0, so this can just be `cp.zeros` instead of rng.

python/cuda_cccl/benchmarks/compute/partition/three_way.py
```diff
@@ -0,0 +1,150 @@
...
+    dtype = TYPE_MAP[type_str]
+    num_elements = int(state.get_int64("Elements{io}"))
+    entropy_str = state.get_string("Entropy")
+
+    alloc_stream = as_cupy_stream(state.get_stream())
+
+    if np.issubdtype(dtype, np.integer):
+        info = np.iinfo(dtype)
+        min_val = (
+            0  # C++ uses min_val{} which is 0 for unsigned, could be min for signed
```
NaderAlAwar: Question: I don't understand the comment. Unless I'm mistaken, min_val is always 0 in C++?

python/cuda_cccl/benchmarks/compute/partition/three_way.py
```diff
@@ -0,0 +1,150 @@
...
+
+    alloc_stream.synchronize()
+
+    # Items where select_first_op(x) is true go to first partition
+    # Items where select_second_op(x) is true (and first is false) go to second partition
+    # Items where both are false go to unselected
+    #
+    # In C++ the predicates are:
+    #   select_op_1: x < left_border (selects first partition)
+    #   select_op_2: x < right_border (combined with !select_op_1, selects second partition)
```
NaderAlAwar: Suggestion: these comments feel redundant and should be removed

python/cuda_cccl/benchmarks/compute/partition/three_way.py
```diff
@@ -0,0 +1,150 @@
...
+
+    # Convert borders to the correct type for closure capture
+    left_thresh = dtype(left_border)
+    right_thresh = dtype(right_border)
+
+    # Use regular Python functions - cuda.compute JIT-compiles them internally
```
NaderAlAwar: Suggestion: this comment can be removed

python/cuda_cccl/benchmarks/compute/utils.py
danielfrg: done

python/cuda_cccl/benchmarks/compute/merge_sort/keys.py
danielfrg: done

python/cuda_cccl/benchmarks/compute/merge_sort/keys.py
danielfrg: Yes, sometimes it returns invalid values, i'll file an issue

python/cuda_cccl/benchmarks/compute/histogram/even.py
danielfrg: Created one here: https://github.com/NVIDIA/cccl/issues/7909

python/cuda_cccl/benchmarks/compute/merge_sort/keys.py
danielfrg: I thought i saw another error but i lost the logs. I removed and we get a regular `OutOfMemoryError: Out of memory allocating` error.

python/cuda_cccl/benchmarks/compute/merge_sort/keys.py
danielfrg: Done here https://github.com/NVIDIA/cccl/issues/7911 !

python/cuda_cccl/benchmarks/compute/merge_sort/keys.py
```diff
@@ -0,0 +1,97 @@
...
+        d_out_keys = cp.empty(num_elements, dtype=dtype)
+
+    alloc_stream.synchronize()
+
+    sorter = make_merge_sort(
+        d_in_keys=d_in_keys,
+        d_in_items=None,
+        d_out_keys=d_out_keys,
+        d_out_items=None,
+        op=OpKind.LESS,
```
NaderAlAwar: Nit: for this and all other benchmarks in the PR, please remove the keyword arguments. This also applies to the calls to sorter below. I think they make the code somewhat harder to read.

python/cuda_cccl/benchmarks/compute/select/flagged.py
```diff
@@ -0,0 +1,104 @@
...
+"""
+
+import sys
+from pathlib import Path
+
+sys.path.insert(0, str(Path(__file__).parent.parent))
+
+import cupy as cp
+import numpy as np
+from utils import ENTROPY_TO_PROB, as_cupy_stream, generate_data_with_entropy
```
NaderAlAwar: Critical: `ENTROPY_TO_PROB` is missing from utils.py

python/cuda_cccl/benchmarks/compute/select/if.py
```diff
@@ -0,0 +1,120 @@
...
+
+import sys
+from pathlib import Path
+
+sys.path.insert(0, str(Path(__file__).parent.parent))
+
+import cupy as cp
+import numpy as np
+from utils import (
+    ENTROPY_TO_PROB,
```
NaderAlAwar, line 28: Critical: `ENTROPY_TO_PROB` is missing from utils.py

python/cuda_cccl/benchmarks/compute/select/flagged.py
```diff
@@ -0,0 +1,104 @@
...
+        d_out_flags = cp.empty(selected_elements, dtype=np.uint8)
+        d_num_selected = cp.empty(1, dtype=np.uint64)
+
+    def flag_predicate(pair):
+        return np.uint8(pair[1] != 0)
+
+    flag_predicate.__annotations__ = {
+        "pair": zip_it.value_type,
+        "return": np.uint8,
+    }
```
NaderAlAwar: Important: if this can be removed, we should get rid of it

python/cuda_cccl/benchmarks/compute/segmented_sort/keys.py
```diff
@@ -0,0 +1,223 @@
...
+            except TypeError:
+                pass
+
+    try:
+        state.exec(launcher, batched=False, sync=True)
+    except TypeError:
+        state.exec(launcher, batched=False)
+
+
+def bench_segmented_sort_power(state: bench.State):
```
NaderAlAwar: Important: there is currently some code duplication between this one and the other function. Please try to unify them like the C++ benchmark does

python/cuda_cccl/benchmarks/compute/segmented_sort/keys.py
```diff
@@ -0,0 +1,223 @@
...
+    type_str = state.get_string("KeyT")
+    dtype = TYPE_MAP[type_str]
+    num_elements = int(state.get_int64("Elements{io}"))
+    num_segments = int(state.get_int64("Segments{io}"))
+    entropy_str = state.get_string("Entropy")
+
+    try:
```
NaderAlAwar: Important: we should not wrap the entire code block with a try catch. Try to localize the try catch to the exact part that does memory allocation.

python/cuda_cccl/benchmarks/compute/segmented_sort/keys.py
```diff
@@ -0,0 +1,223 @@
...
+            d_in_values,
+            d_out_values,
+            start_offsets,
+            end_offsets,
+            num_elements,
+            num_segments,
+        )
+    except (MemoryError, cp.cuda.memory.OutOfMemoryError):
+        state.skip("Skipping: out of memory.")
+        return
```
NaderAlAwar: Important: we must make sure to match what the C++ benchmark does when it runs out of memory. Is this doing that right now?

python/cuda_cccl/benchmarks/compute/segmented_sort/keys.py
```diff
@@ -0,0 +1,223 @@
...
+
+    try:
+        alloc_stream = as_cupy_stream(state.get_stream())
+        d_in_keys = generate_data_with_entropy(
+            num_elements, dtype, "1.000", alloc_stream
+        )
+        with alloc_stream:
+            d_out_keys = cp.empty(num_elements, dtype=dtype)
+            d_in_values = cp.arange(num_elements, dtype=dtype)
+            d_out_values = cp.empty(num_elements, dtype=dtype)
```
NaderAlAwar: Critical: this is a keys only benchmark, we should not be allocating any values

python/cuda_cccl/benchmarks/compute/select/if.py
```diff
@@ -0,0 +1,120 @@
...
+    dtype = TYPE_MAP[type_str]
+    num_elements = int(state.get_int64("Elements{io}"))
+    entropy_str = state.get_string("Entropy")
+
+    probability = ENTROPY_TO_PROB[entropy_str]
+    threshold = lerp_min_max(dtype, probability)
+
+    alloc_stream = as_cupy_stream(state.get_stream())
+
+    d_in = generate_data_with_entropy(num_elements, dtype, entropy_str, alloc_stream)
```
NaderAlAwar: Critical: the C++ benchmark does not use entropy to generate the data. It only uses it for the selection of the threshold. This could affect the selection ratio

python/cuda_cccl/benchmarks/compute/merge_sort/keys.py
```diff
@@ -0,0 +1,97 @@
...
+        )
+
+    state.exec(launcher, batched=False)
+
+
+if __name__ == "__main__":
+    b = bench.register(bench_merge_sort_keys)
+    b.set_name("base")
+
+    b.add_string_axis("KeyT", list(ALL_TYPES.keys()))
```
NaderAlAwar: Important:
```suggestion
    b.add_string_axis("T{ct}", list(ALL_TYPES.keys()))
```

python/cuda_cccl/benchmarks/compute/segmented_sort/keys.py
```diff
@@ -0,0 +1,223 @@
...
+        )
+    except (MemoryError, cp.cuda.memory.OutOfMemoryError):
+        state.skip("Skipping: out of memory.")
+        return
+
+
+if __name__ == "__main__":
+    b_power = bench.register(bench_segmented_sort_power)
+    b_power.set_name("power")
+    b_power.add_string_axis("KeyT", list(TYPE_MAP.keys()))
```
NaderAlAwar: Important:
```suggestion
    b_power.add_string_axis("T{ct}", list(TYPE_MAP.keys()))
```
this also appliues to the other benchmarks

python/cuda_cccl/benchmarks/compute/partition/three_way.py
```diff
@@ -0,0 +1,139 @@
...
+"""
+
+import sys
+from pathlib import Path
+
+sys.path.insert(0, str(Path(__file__).parent.parent))
+
+import cupy as cp
+import numpy as np
+from utils import ALL_TYPES, as_cupy_stream, generate_data_with_entropy
```
NaderAlAwar: Critical: the C++ benchmark uses `fundamental_types`, which is signed types only

python/cuda_cccl/benchmarks/compute/merge_sort/keys.py
```diff
@@ -0,0 +1,97 @@
...
+"""
+
+import sys
+from pathlib import Path
+
+sys.path.insert(0, str(Path(__file__).parent.parent))
+
+import cupy as cp
+import numpy as np
+from utils import ALL_TYPES, as_cupy_stream, generate_data_with_entropy
```
NaderAlAwar: Critical: the C++ benchmark uses `all_types`, which does not include unsigned types, while `ALL_TYPES` here does.

python/cuda_cccl/benchmarks/compute/reduce/sum.py
```diff
@@ -0,0 +1,85 @@
...
+    dtype = TYPE_MAP[type_str]
+    num_items = int(state.get_int64("Elements{io}"))
+
+    # Setup data - use random values like C++ generate()
+    alloc_stream = as_cupy_stream(state.get_stream())
+    with alloc_stream:
+        if np.issubdtype(dtype, np.integer):
+            d_in = cp.random.randint(0, 100, size=num_items, dtype=dtype)
+        else:
+            d_in = cp.random.random(num_items, dtype=dtype)
```
NaderAlAwar: Important: the C++ benchmarks do generate(num_items), which has a much larger range than (0,100). This pattern appears in multiple places in other benchmarks as well. Please make sure we use similar ranges in all cases.

python/cuda_cccl/benchmarks/compute/segmented_reduce/sum.py
```diff
@@ -0,0 +1,148 @@
...
+    # - small: SegmentSize 2^0 to 2^4
+    # - medium: SegmentSize 2^5 to 2^8
+    # - large: SegmentSize 2^9 to 2^16
+
+    # Small segments benchmark
+    b_small = bench.register(bench_segmented_reduce_sum)
+    b_small.set_name("small")
+    b_small.add_string_axis("T", list(TYPE_MAP.keys()))
+    b_small.add_int64_power_of_two_axis("Elements{io}", range(16, 29, 4))
+    b_small.add_int64_axis("SegmentSize", [2**i for i in range(0, 5)])
```
NaderAlAwar: Nit: please also use `add_int64_power_of_two_axis` for consistency with C++, same for below

python/cuda_cccl/benchmarks/compute/segmented_reduce/custom.py
```diff
@@ -0,0 +1,128 @@
...
+
+    state.exec(launcher, batched=False)
+
+
+if __name__ == "__main__":
+    b_small = bench.register(bench_segmented_reduce_custom)
+    b_small.set_name("small")
+    b_small.add_string_axis("T", list(TYPE_MAP.keys()))
+    b_small.add_int64_power_of_two_axis("Elements{io}", range(16, 29, 4))
+    b_small.add_int64_axis("SegmentSize", [2**i for i in range(0, 5)])
```
NaderAlAwar: Nit: please also use `add_int64_power_of_two_axis` for consistency with C++, same for below

python/cuda_cccl/benchmarks/compute/select/unique_by_key.py
```diff
@@ -0,0 +1,146 @@
...
+"""
+
+import sys
+from pathlib import Path
+
+sys.path.insert(0, str(Path(__file__).parent.parent))
+
+import cupy as cp
+import numpy as np
+from utils import ALL_TYPES, INTEGER_TYPES, as_cupy_stream, generate_key_segments
```
NaderAlAwar: Important: the C++ benchmark uses signed types only for values, see how `all_types` in C++ is defined in `nvbench_helper.cuh`.

python/cuda_cccl/benchmarks/compute/select/flagged.py
```diff
@@ -0,0 +1,104 @@
...
+        num_items=num_elements,
+    )
+    with alloc_stream:
+        temp_storage = cp.empty(temp_storage_bytes, dtype=np.uint8)
+
+    state.add_element_count(num_elements)
+    state.add_global_memory_reads(num_elements * d_in.dtype.itemsize)
+    state.add_global_memory_reads(num_elements * flags.dtype.itemsize)
+    state.add_global_memory_writes(selected_elements * d_out.dtype.itemsize)
+    state.add_global_memory_writes(selected_elements * d_out_flags.dtype.itemsize)
```
NaderAlAwar: Important: we do not write the flags back, this line should be removed

python/cuda_cccl/benchmarks/compute/transform/grayscale.py
```diff
@@ -0,0 +1,79 @@
...
+    def to_grayscale(rgb):
+        return w_r * rgb[0] + w_g * rgb[1] + w_b * rgb[2]
+
+    try:
+        alloc_stream = as_cupy_stream(state.get_stream())
+        with alloc_stream:
+            d_r = cp.random.random(num_elements).astype(dtype)
+            d_g = cp.random.random(num_elements).astype(dtype)
+            d_b = cp.random.random(num_elements).astype(dtype)
+            d_out = cp.empty(num_elements, dtype=dtype)
```
NaderAlAwar: Important: currently this uses a struct-of-arrays layout (SoA), while the C++ version uses array-of-structs (AoS). We should be consistent here. The correct way to define this is 

```
  @gpu_struct
  class RGB:
      r: np.float32  # or parameterize by dtype
      g: np.float32
      b: np.float32
```

Actually this is better to match the type parameterization in C++

```
gpu_struct({"r": dtype, "g": dtype, "b": dtype})
```

the you can generate the pixel data like so

```
d_pixels = cp.random.random((n, 3)).astype(dtype).view(RGB.dtype)
```

danielfrg:
This should be ready for another review

python/cuda_cccl/benchmarks/compute/select/flagged.py
```diff
@@ -0,0 +1,114 @@
...
+    #   reads: T * elements + bool * elements
+    #   writes: T * selected_elements + OffsetT * 1
+    state.add_element_count(num_elements)
+    state.add_global_memory_reads(num_elements * d_in.dtype.itemsize)
+    state.add_global_memory_reads(num_elements * np.dtype(np.bool_).itemsize)
+    state.add_global_memory_writes(selected_elements * d_out.dtype.itemsize)
+    state.add_global_memory_writes(d_num_selected.dtype.itemsize)
+
+    def launcher(launch: bench.Launch):
+        selector(
```
NaderAlAwar: Important: thinking about this some more, I think we should drop flagged.py from this PR. The corresponding C++ benchmark is measuring the flagged-selection path: separate value input + separate flag input. Python does not expose that path yet, so this file works around it by zipping `(value, flag)` together and running predicate-based `select` instead. That is a different API shape and a different implementation path than the C++ benchmark is exercising.

python/cuda_cccl/benchmarks/compute/segmented_reduce/custom.py
```diff
@@ -0,0 +1,121 @@
...
+        h_init=h_init,
+    )
+    with alloc_stream:
+        temp_storage = cp.empty(temp_storage_bytes, dtype=np.uint8)
+
+    state.add_element_count(actual_elements)
+    state.add_global_memory_reads(actual_elements * d_in.dtype.itemsize, "Size")
+    state.add_global_memory_writes(num_segments * d_out.dtype.itemsize)
+
+    def launcher(launch: bench.Launch):
```
NaderAlAwar: Important: thinking about this some more, I think we should drop segmented_reduce/custom.py and sum.py from this PR. The C++ benchmark here is specifically the fixed-size segmented-reduce path. In Python, we do not expose that path yet; the only available API is the generic offset-based segmented reduce. This therefore benchmarks a different algorithm/dispatch path from the one used in C++.

python/cuda_cccl/benchmarks/compute/partition/three_way.py
```diff
@@ -0,0 +1,138 @@
...
+        return x < right_thresh
+
+    partitioner = make_three_way_partition(
+        d_in=d_in,
+        d_first_part_out=d_first_part_out,
+        d_second_part_out=d_second_part_out,
+        d_unselected_out=d_unselected_out,
+        d_num_selected_out=d_num_selected_out,
+        select_first_part_op=select_first_part,
+        select_second_part_op=select_second_part,
```
NaderAlAwar: Important: for this and other files in this PR, please remove the keyword arguments.

python/cuda_cccl/benchmarks/compute/select/unique_by_key.py
```diff
@@ -0,0 +1,163 @@
...
+                    dtype=np.int64,
+                ).astype(value_dtype)
+        elif np.issubdtype(value_dtype, np.complexfloating):
+            real = cp.random.uniform(-1, 1, size=num_elements).astype(np.float32)
+            imag = cp.random.uniform(-1, 1, size=num_elements).astype(np.float32)
+            d_in_values = (real + 1j * imag).astype(value_dtype)
+        else:
+            d_in_values = cp.random.uniform(-1, 1, size=num_elements).astype(
+                value_dtype
+            )
```
NaderAlAwar: Important: this seems unnecessary as the C++ benchmark does not randomly generate the values, it default constructs them instead

python/cuda_cccl/benchmarks/compute/partition/three_way.py
```diff
@@ -0,0 +1,138 @@
...
+        num_items=num_elements,
+    )
+    with alloc_stream:
+        temp_storage = cp.empty(temp_storage_bytes, dtype=np.uint8)
+
+    state.add_element_count(num_elements)
+    state.add_global_memory_reads(num_elements * d_in.dtype.itemsize)
+    state.add_global_memory_writes(num_elements * d_in.dtype.itemsize)
+    # C++ reports add_global_memory_writes<offset_t>(1) — 1 element of offset type.
+    state.add_global_memory_writes(1 * d_num_selected_out.dtype.itemsize)
```
NaderAlAwar: Nit:
```suggestion
    state.add_global_memory_writes(d_num_selected_out.dtype.itemsize)
```

python/cuda_cccl/benchmarks/compute/radix_sort/pairs.py
```diff
@@ -0,0 +1,116 @@
...
+    num_elements = int(state.get_int64("Elements{io}"))
+    entropy_str = state.get_string("Entropy")
+
+    alloc_stream = as_cupy_stream(state.get_stream())
+
+    d_in_keys = generate_data_with_entropy(
+        num_elements, key_dtype, entropy_str, alloc_stream
+    )
+
+    d_in_values = generate_values(num_elements, value_dtype, alloc_stream)
```
NaderAlAwar: Important: for consistency with C++, I would remove this custom `generate_values()` function, and use `generate_data_with_entropy()` with entropy set to "1.000"

python/cuda_cccl/benchmarks/compute/reduce/deterministic.py
```diff
@@ -0,0 +1,85 @@
...
+        d_out = cp.empty(1, dtype=dtype)
+
+    h_init = np.zeros(1, dtype=dtype)
+
+    reducer = make_reduce_into(
+        d_in=d_in,
+        d_out=d_out,
+        op=OpKind.PLUS,
+        h_init=h_init,
+        determinism=Determinism.RUN_TO_RUN,
```
NaderAlAwar: Critical: the correct determinism is `GPU_TO_GPU`, which is currently not implemented. I would therefore remove this benchmark until we implement it in cuda.compute

python/cuda_cccl/benchmarks/compute/reduce/min.py
```diff
@@ -0,0 +1,85 @@
...
+        op=OpKind.MINIMUM,
+        num_items=num_items,
+        h_init=h_init,
+    )
+    with alloc_stream:
+        temp_storage = cp.empty(temp_storage_bytes, dtype=np.uint8)
+
+    state.add_element_count(num_items)
+    state.add_global_memory_reads(num_items * d_in.dtype.itemsize, "Size")
+    state.add_global_memory_writes(1 * d_out.dtype.itemsize)
```
NaderAlAwar: Nit:
```suggestion
    state.add_global_memory_writes(d_out.dtype.itemsize)
```

python/cuda_cccl/benchmarks/compute/reduce/custom.py
```diff
@@ -0,0 +1,82 @@
...
+
+    reducer = make_reduce_into(d_in=d_in, d_out=d_out, op=max_op, h_init=h_init)
+
+    temp_storage_bytes = reducer(
+        temp_storage=None,
+        d_in=d_in,
+        d_out=d_out,
+        op=max_op,
+        num_items=num_items,
+        h_init=h_init,
```
NaderAlAwar: Suggestion: please remove keyword arguments for all reduce benchmarks

python/cuda_cccl/benchmarks/compute/reduce/custom.py
```diff
@@ -0,0 +1,82 @@
...
+    )
+    with alloc_stream:
+        temp_storage = cp.empty(temp_storage_bytes, dtype=np.uint8)
+
+    state.add_element_count(num_items)
+    state.add_global_memory_reads(num_items * d_in.dtype.itemsize, "Size")
+    state.add_global_memory_writes(1 * d_out.dtype.itemsize)
```
NaderAlAwar: Nit:
```suggestion
    state.add_global_memory_writes(d_out.dtype.itemsize)
```

python/cuda_cccl/benchmarks/compute/reduce/sum.py
```diff
@@ -0,0 +1,80 @@
...
+        op=OpKind.PLUS,
+        num_items=num_items,
+        h_init=h_init,
+    )
+    with alloc_stream:
+        temp_storage = cp.empty(temp_storage_bytes, dtype=np.uint8)
+
+    state.add_element_count(num_items)
+    state.add_global_memory_reads(num_items * d_in.dtype.itemsize, "Size")
+    state.add_global_memory_writes(1 * d_out.dtype.itemsize)
```
NaderAlAwar: Nit:
```suggestion
    state.add_global_memory_writes(d_out.dtype.itemsize)
```

python/cuda_cccl/benchmarks/compute/scan/exclusive/custom.py
```diff
@@ -0,0 +1,83 @@
...
+    num_items = int(state.get_int64("Elements{io}"))
+
+    alloc_stream = as_cupy_stream(state.get_stream())
+    with alloc_stream:
+        d_in = generate_data_with_entropy(num_items, dtype, "1.000", alloc_stream)
+        d_out = cp.empty(num_items, dtype=dtype)
+
+    h_init = np.zeros(1, dtype=dtype)
+
+    scanner = make_exclusive_scan(d_in=d_in, d_out=d_out, op=max_op, init_value=h_init)
```
NaderAlAwar: Important: please remove keyword arguments

python/cuda_cccl/benchmarks/compute/transform/babelstream.py
```diff
@@ -0,0 +1,237 @@
...
+    _reset_pools()
+
+    try:
+        alloc_stream = as_cupy_stream(state.get_stream())
+        with alloc_stream:
+            c = cp.full(num_items, START_C, dtype=dtype)
+            b = cp.full(num_items, START_B, dtype=dtype)
+    except (MemoryError, cp.cuda.memory.OutOfMemoryError):
+        state.skip("Skipping: out of memory.")
+        return
```
NaderAlAwar, line 61: Important: we should mimic the C++ behavior when we run out of memory, same for other transform benchmarks

python/cuda_cccl/benchmarks/compute/transform/babelstream.py
```diff
@@ -0,0 +1,237 @@
...
+
+    scalar = dtype(START_SCALAR)
+
+    def mul_op(ci):
+        return ci * scalar
+
+    transform = cuda.compute.make_unary_transform(d_in=c, d_out=b, op=mul_op)
```
NaderAlAwar: Important: please remove keyword arguments

python/cuda_cccl/benchmarks/compute/select/flagged.py
danielfrg: Yeah I think i agree with that i was on the fence about this one. I'll remove them.

python/cuda_cccl/benchmarks/compute/segmented_reduce/custom.py
danielfrg: Removed

python/cuda_cccl/benchmarks/compute/select/flagged.py
danielfrg: Removed

python/cuda_cccl/benchmarks/compute/partition/three_way.py
danielfrg: done

python/cuda_cccl/benchmarks/compute/select/unique_by_key.py
danielfrg: done

python/cuda_cccl/benchmarks/compute/reduce/deterministic.py
danielfrg: removed

python/cuda_cccl/benchmarks/compute/reduce/custom.py
danielfrg: done

python/cuda_cccl/benchmarks/compute/scan/exclusive/custom.py
danielfrg: done

python/cuda_cccl/benchmarks/compute/transform/babelstream.py
danielfrg: done

python/cuda_cccl/benchmarks/compute/radix_sort/pairs.py
danielfrg: done

python/cuda_cccl/benchmarks/compute/segmented_reduce/sum.py
```diff
@@ -0,0 +1,140 @@
...
+            end_offsets,
+            h_init,
+            launch.get_stream(),
+        )
+
+    state.exec(launcher, batched=False)
+
+
+if __name__ == "__main__":
+    # Register three benchmarks matching C++ structure:
```
NaderAlAwar: Important: this benchmark still needs to be removed (like the `custom.py` you just deleted)

python/cuda_cccl/benchmarks/compute/segmented_sort/keys.py
```diff
@@ -0,0 +1,182 @@
...
+            try:
+                state.exec(exec_tag.sync, launcher, batched=False)
+                return
+            except TypeError:
+                pass
+
+    try:
+        state.exec(launcher, batched=False, sync=True)
+    except TypeError:
+        state.exec(launcher, batched=False)
```
NaderAlAwar: Important: 
```suggestion
    state.exec(launcher, batched=False, sync=True)
```

python/cuda_cccl/benchmarks/compute/transform/babelstream.py
```diff
@@ -0,0 +1,213 @@
...
+    Benchmark: b[i] = c[i] * scalar
+    Unary transform with scalar multiplication.
+    """
+    type_str = state.get_string("T{ct}")
+    dtype = TYPE_MAP[type_str]
+    num_items = int(state.get_int64("Elements{io}"))
+
+    _reset_pools()
+
+    try:
```
NaderAlAwar, line 55: Suggestion: move the try catch block to only surround the memory allocation parts (`cp.full`). Apply this to al other transform benchmarks as well.

python/cuda_cccl/benchmarks/compute/run_benchmarks.py
```diff
@@ -0,0 +1,495 @@
...
+# ============================================================================
+
+SCRIPT_DIR = Path(__file__).parent
+CCCL_ROOT = SCRIPT_DIR.parents[3]
+RESULTS_DIR = SCRIPT_DIR / "results"
+CUB_BENCH_DIR = CCCL_ROOT / "build" / "cub" / "bin"
+QUICK_CONFIG_FILE = SCRIPT_DIR / "quick_configs.yaml"
+
+# Supported benchmarks (Python implementations available)
+SUPPORTED_BENCHMARKS = [
```
NaderAlAwar, line 55: Important: this list still includes some benchmarks that were removed, like segmented_reduce/custom (and others). We should remove those

python/cuda_cccl/benchmarks/compute/run_benchmarks.py
```diff
@@ -0,0 +1,495 @@
...
+        "-b", "--benchmark", help="Run specific benchmark only [default: all]"
+    )
+    parser.add_argument("--py", action="store_true", help="Only run Python benchmarks")
+    parser.add_argument("--cpp", action="store_true", help="Only run C++ benchmarks")
+    parser.add_argument(
+        "-q",
+        "--quick",
+        action="store_true",
+        help="Run with reduced parameter set for fast testing",
+    )
```
NaderAlAwar, line 386: Suggestion: for simplicity, I would prefer to remove this option, and delete `quick_configs.yaml` to avoid clutter

python/cuda_cccl/benchmarks/compute/quick_configs.yaml
```diff
@@ -0,0 +1,162 @@
+# Quick mode configurations for fast benchmark testing
```
NaderAlAwar, line 1: SUggestion: delete this file after removing the `--quick` option

python/cuda_cccl/benchmarks/compute/run_benchmarks.py
```diff
@@ -0,0 +1,495 @@
...
+        if py_status and py_status.get("status") != "ok":
+            print(
+                "    Python status:  "
+                f"{py_status.get('status')} (exit {py_status.get('returncode')})"
+            )
+
+    print()
+    print("To compare results, run:")
+    for bench in benchmarks_to_run:
+        print(f"  python analysis/python_vs_cpp_summary.py -b {bench} -d {args.device}")
```
NaderAlAwar: Important: the analysis tooling seems to have been deleted. Can you go through this file and make sure it reflects the present state of the PR?

python/cuda_cccl/benchmarks/compute/pixi.lock
```diff
@@ -0,0 +1,1723 @@
+version: 6
```
NaderAlAwar: Important: delete this file, `pixi.toml` should be enough

python/cuda_cccl/benchmarks/compute/.opencode/commands/migration-status.md
```diff
@@ -0,0 +1,18 @@
+---
```
NaderAlAwar: Important: this file can be deleted now

danielfrg:
I think i addressed all the previous comments

python/cuda_cccl/benchmarks/compute/utils.py
```diff
@@ -0,0 +1,253 @@
...
+    "U8": np.uint8,
+    "U16": np.uint16,
+    "U32": np.uint32,
+    "U64": np.uint64,
+    "F32": np.float32,
+    "F64": np.float64,
+}
+
+SIGNED_TYPES = {k: ALL_TYPES[k] for k in ("I8", "I16", "I32", "I64", "F32", "F64")}
+INTEGER_TYPES = {k: ALL_TYPES[k] for k in ("I8", "I16", "I32", "I64")}
```
NaderAlAwar: Suggestion: this is a duplicate of `INTEGRAL_TYPES` below and can be removed

python/cuda_cccl/benchmarks/compute/segmented_reduce/sum.py
danielfrg: removed

python/cuda_cccl/benchmarks/compute/run_benchmarks.py
danielfrg: i would like to keep this because its quite useful if we want to run a small subset of benchmarks, maybe in CI we dont want to run it all but just a smaller subset. What if we refactor the --quick into something that allows to pass a config yml file instead?

danielfrg:
All done except one question i left about the --quick, i would like to keep it but we can probably refactor into a --config that allows multiple config for the benchmarks for diff scenarios?

python/cuda_cccl/benchmarks/compute/run_benchmarks.py
NaderAlAwar: On second thought, the CI idea is good. We can keep it as is, disregard my above comment

danielfrg: sounds good!

danielfrg:
Ok since you want to keep the --quick option now we should be mostly good to go.

These are the latest results on my dev machine:

CleanShot 2026-03-19 at 11 17 02@2x

[results.zip](https://github.com/user-attachments/files/26121363/results.zip)

I'll open up issues to track the performance of the transforms that we do not match at the moment

NaderAlAwar:
Great work @danielfrg! Could you also open up issues for `segmented_sort/keys` and `histogram/even`? Histogram being faster in python is suspicious.

PR #7342, opened 2026-01-23
Decouple `numba-cuda` from type system and other internals

shwina:
## Description

Closes https://github.com/NVIDIA/cccl/issues/7352. See parent issue #4610 for detailed rationale. Of all the PRs that will contribute to #4610, this one is by far the largest and most significant.

This PR decouples the majority of our internals from `numba-cuda`. In particular:

- Introduces a native type system that is independent of Numba
- Moves all JIT infrastructure to `_jit.py` 
- Decouples `@gpu_struct` from numba - registration of struct types with Numba is done at JIT time


python/cuda_cccl/cuda/compute/_jit.py
```diff
@@ -0,0 +1,716 @@
...
+
+        struct_val = cgutils.create_struct_proxy(toty)(context, builder)
+        for i, (field_name, field_type) in enumerate(field_spec.items()):
+            elem = builder.extract_value(val, i)
+            elem = context.cast(builder, elem, tuple_types[i], field_type)
+            setattr(struct_val, field_name, elem)
+
+        return struct_val._getvalue()
+
+    @lower_cast(_StructBase, StructType)
```
shwina, line 250: Note to reviewers: this is new in the struct logic. It  allows implicit conversions between structs with matching field names and layouts. This is important and useful in various situations where the users define a `gpu_struct` and operations that work on those, but we invoke the same operations with another `gpu_struct` with identical layout.

python/cuda_cccl/cuda/compute/_jit.py
```diff
@@ -0,0 +1,682 @@
...
+
+    if isinstance(output_type, cccl_types.TypeDescriptor):
+        numba_output_type = type_descriptor_to_numba(output_type)
+    else:
+        numba_output_type = output_type
+
+    sig = numba_output_type(*numba_input_types)
+    wrapped_op, wrapper_sig = create_op_void_ptr_wrapper(op, sig)
+    ltoir, _ = numba.cuda.compile(wrapped_op, sig=wrapper_sig, output="ltoir")
+    return ltoir, wrapped_op.__name__
```
shwina: Can we return a `bindings.Op` object here?

python/cuda_cccl/cuda/compute/_jit.py
```diff
@@ -0,0 +1,682 @@
...
+    sig = numba.types.void(*numba_input_types)
+    c_fn = numba.cfunc(sig)(fn)
+    return c_fn.ctypes
+
+
+__all__ = [
+    "compile_host_op",
+    "compile_iterator",
+    "compile_op",
+    "infer_signature",
```
shwina: I think we can remove this export

python/cuda_cccl/cuda/compute/_intrinsics.py
```diff
@@ -0,0 +1,166 @@
+# Copyright (c) 2025, NVIDIA CORPORATION & AFFILIATES. ALL RIGHTS RESERVED.
```
shwina: All license should be updated

NaderAlAwar (approved):

Issue #7332, opened 2026-01-23
[BUG]: serious warning in 32bit system


correaa:
### Is this a duplicate?

- [x] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cccl/issues) for this bug and that I agree to the [Code of Conduct](CODE_OF_CONDUCT.md)

### Type of Bug

Compile-time Error

### Component

General CCCL

### Describe the bug

I am not sure if you care about 32bit systems, but I get this warning deep in CCCL.

```
/builds/correaa/boost-multi/build/_deps/cccl-cmake-src/lib/cmake/libcudacxx/../../../libcudacxx/include/cuda/__memory/check_address.h:75:52: error: narrowing conversion of '18446744073709551615' from 'long long unsigned int' to 'uintptr_t' {aka 'unsigned int'} [-Wnarrowing]
   75 |   auto __limit = ::cuda::std::uintptr_t{UINTMAX_MAX} - static_cast<::cuda::std::uintptr_t>(__n);
```

When using this compiler

```
current binary directory: /builds/correaa/boost-multi/build
-- The CXX compiler identification is GNU 14.2.0
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
```

and this version of CCCL

```
-- Finding CCCL components: Thrust;CUB;libcudacxx
-- Found CUB: /builds/correaa/boost-multi/build/_deps/cccl-cmake-src/lib/cmake/cub/cub-config.cmake (found suitable version "3.3.0.0", minimum required is "3.3.0.0")
-- Found libcudacxx: /builds/correaa/boost-multi/build/_deps/cccl-cmake-src/lib/cmake/libcudacxx/libcudacxx-config.cmake (found suitable exact version "3.3.0.0")
-- Found Thrust: /builds/correaa/boost-multi/build/_deps/cccl-cmake-src/lib/cmake/thrust/thrust-config.cmake (found suitable exact version "3.3.0.0")
-- Found CCCL: /builds/correaa/boost-multi/build/_deps/cccl-cmake-src/lib/cmake/cccl/cccl-config.cmake (found version "3.3.0.0")
-- Found libcudacxx: /builds/correaa/boost-multi/build/_deps/cccl-cmake-src/lib/cmake/libcudacxx/libcudacxx-config.cmake (found version "3.3.0.0")
[...]

PR #7345, opened 2026-01-24
Enforce supported `long double` format

davebayer:
After thinking a bit more about #7332, I think we should make unknown/unsupported `long double` format a hard error. The `CCCL_DISABLE_LONG_DOUBLE_SUPPORT` macro can be used to disable `long double` support in CCCL.

The advantage is that now the user gets a message about what went wrong, not just getting some errors deep inside the libcu++'s internals.

### DIFF
--- a/libcudacxx/include/cuda/std/__cccl/dialect.h
+++ b/libcudacxx/include/cuda/std/__cccl/dialect.h
@@ -188,10 +188,10 @@
 
 // Fixme: replace the condition with (!_CCCL_DEVICE_COMPILATION())
 // FIXME: Enable this for clang-cuda in a followup
-#if !_CCCL_CUDA_COMPILATION() && !defined(CCCL_DISABLE_LONG_DOUBLE)
+#if !_CCCL_CUDA_COMPILATION() && !defined(CCCL_DISABLE_LONG_DOUBLE_SUPPORT)
 #  define _CCCL_HAS_LONG_DOUBLE() 1
-#else // ^^^ !_CCCL_CUDA_COMPILATION() ^^^ / vvv _CCCL_CUDA_COMPILATION() vvv
+#else // ^^^ has long double ^^^ / vvv no long double vvv
 #  define _CCCL_HAS_LONG_DOUBLE() 0
-#endif // ^^^ _CCCL_CUDA_COMPILATION() ^^^
+#endif // ^^^ no long double ^^^
 
 #endif // __CCCL_DIALECT_H


libcudacxx/include/cuda/std/__floating_point/format.h
```diff
@@ -62,22 +62,18 @@ template <class _Tp>
...
-    else // Unsupported long double format
-    {
-      return __fp_format::__invalid;
-    }
+#  if LDBL_MIN_EXP == -1021 && LDBL_MAX_EXP == 1024 && LDBL_MANT_DIG == 53
+    return __fp_format::__binary64;
+#  elif LDBL_MIN_EXP == -16381 && LDBL_MAX_EXP == 16384 && LDBL_MANT_DIG == 64
+    static_assert(sizeof(long double) == 16,
+                  "When the long double format is x86 80-bit extended floating point, CCCL requires the size of long "
+                  "double to be 16 bytes.");
```
bernhardmgruber, line 70: Q: Could this be a source breaking change on any of our **supported** platforms?

davebayer: When implementing the long double format detection, I thought it would be better to make the compilation fail once the user calls an API that works with `long double`. However, what I missed is that the `l` variants of cmath functions (for example `fabsl`) are not templated, thus all of the machinery is instantiated even just when the headers is included.

Currently, the `long double` format detection results in an invalid format if the size of `long double` is not `16`. But that also means that if your `long double` has 96-bits and you include `<cuda/std/cmath>` the compilation would fail. So, this doesn't change with this PR.

Another thing is that we don't allow `long double` with CUDA compilation yet.

So, the only combination for which it could be a source breaking change is C++ compilation with gcc with `-m96bit-long-double` flag (16 byte size is default on 64-bit systems) of source file that doesn't include `<cuda/std/__cmath/abs.h>`, but includes `<cuda/std/__floating_point/format.h>`.

I think this is very unlikely to happen.

Also, I'd like to point out that we don't currently test *any* of our `long double` functionality.

bernhardmgruber: I think that sounds rare enough. And IIUC, if a user would be broken by this PR, they could define `CCCL_DISABLE_LONG_DOUBLE` to make their code compile again. Unless they would rely on `long double` functionality, which is broken for their platform anyway. I think we are fine. Thx for the analysis!

libcudacxx/include/cuda/std/__cccl/dialect.h
```diff
@@ -188,7 +188,7 @@
 
 // Fixme: replace the condition with (!_CCCL_DEVICE_COMPILATION())
 // FIXME: Enable this for clang-cuda in a followup
-#if !_CCCL_CUDA_COMPILATION()
+#if !_CCCL_CUDA_COMPILATION() || defined(CCCL_DISABLE_LONG_DOUBLE)
 #  define _CCCL_HAS_LONG_DOUBLE() 1
```
miscco: I believe that condition is flipped?
```suggestion
#if !_CCCL_CUDA_COMPILATION() && !defined(CCCL_DISABLE_LONG_DOUBLE)
#  define _CCCL_HAS_LONG_DOUBLE() 1
```

libcudacxx/include/cuda/std/__floating_point/format.h
miscco: I love that we have 80 bit `long double` with 128 bit size requirement ^^

libcudacxx/include/cuda/std/__cccl/dialect.h
davebayer: Omg, my_brain.exe stopped working

davebayer:
I've renamed the macro to `CCCL_DISABLE_LONG_DOUBLE_SUPPORT`, so we are consistent with other macros disabling a support for certain type.

libcudacxx/include/cuda/std/__floating_point/format.h
```diff
@@ -62,22 +62,18 @@ template <class _Tp>
...
-    }
-    else if (LDBL_MIN_EXP == -16381 && LDBL_MAX_EXP == 16384 && LDBL_MANT_DIG == 113)
-    {
-      return __fp_format::__binary128;
-    }
-    else // Unsupported long double format
-    {
-      return __fp_format::__invalid;
-    }
+#  if LDBL_MIN_EXP == -1021 && LDBL_MAX_EXP == 1024 && LDBL_MANT_DIG == 53
```
fbusato, line 65: would not be better to use `if constexpr` for values and check for `#if defined(LDBL_MIN_EXP)` outside?

davebayer: I had it like that before, but there is no simple way to make the `static_assert(sizeof(long double) == 16)` work only for the fp80 branch, so I ended up using the preprocessor directives

PR #7346, opened 2026-01-25
Implement the new tuning API for deterministic (rfa) reduce dispatch

bernhardmgruber:
- [x] Merge before: #7334
- [x] Merge before: #7481
- [x] CUB tests pass
- [x] No SASS diff for `cub.bench.reduce.deterministic.base` on SM75/SM86/SM90

### DIFF
--- a/cub/benchmarks/bench/reduce/deterministic.cu
+++ b/cub/benchmarks/bench/reduce/deterministic.cu
@@ -14,7 +14,7 @@
 #if !TUNE_BASE
 struct policy_selector_t
 {
-  [[nodiscard]] constexpr auto operator()(::cuda::arch_id) const -> cub::detail::rfa::rfa_policy
+  [[nodiscard]] _CCCL_API constexpr auto operator()(::cuda::arch_id) const -> cub::detail::rfa::rfa_policy
   {
     return {{TUNE_THREADS_PER_BLOCK, TUNE_ITEMS_PER_THREAD, cub::BLOCK_REDUCE_RAKING},
             {TUNE_THREADS_PER_BLOCK, TUNE_ITEMS_PER_THREAD, cub::BLOCK_REDUCE_RAKING}};
@@ -46,7 +46,7 @@ void deterministic_sum(nvbench::state& state, nvbench::type_list<T>)
   std::size_t temp_storage_bytes{};
   // we explicitly provide template arguments to override accum_t
   cub::detail::rfa::dispatch<input_it_t, output_it_t, int, init_t, transform_t, accum_t>(
-    nullptr, temp_storage_bytes, d_in, d_out, elements, init_t{}, /* stream */ 0);
+    nullptr, temp_storage_bytes, d_in, d_out, elements, init_t{}, /* stream */ nullptr);
 
   thrust::device_vector<nvbench::uint8_t> temp_storage(temp_storage_bytes);
   auto* d_temp_storage = thrust::raw_pointer_cast(temp_storage.data());


cub/benchmarks/bench/reduce/deterministic.cu
```diff
@@ -12,32 +12,13 @@
...
   {
-    using ReducePolicy = AgentReducePolicy;
-
-    // SingleTilePolicy
-    using SingleTilePolicy = ReducePolicy;
-  };
-
-  using MaxPolicy = Policy350;
+    return {{TUNE_THREADS_PER_BLOCK, TUNE_ITEMS_PER_THREAD, cub::BLOCK_REDUCE_RAKING},
+            {TUNE_THREADS_PER_BLOCK, TUNE_ITEMS_PER_THREAD, cub::BLOCK_REDUCE_RAKING}};
```
oleksandr-pavlyk, line 20: Perhaps use struct field names to indicate which list applies to what sub-policy. If values start to differ, it would make it easier on reader to see which is which:


```suggestion
    return {.reduce = {TUNE_THREADS_PER_BLOCK, TUNE_ITEMS_PER_THREAD, cub::BLOCK_REDUCE_RAKING},
            .single_tile = {TUNE_THREADS_PER_BLOCK, TUNE_ITEMS_PER_THREAD, cub::BLOCK_REDUCE_RAKING}};
```

cub/benchmarks/bench/reduce/deterministic.cu
```diff
@@ -76,13 +44,25 @@ void deterministic_sum(nvbench::state& state, nvbench::type_list<T>)
...
 
   std::size_t temp_storage_bytes{};
-  dispatch_t::Dispatch(nullptr, temp_storage_bytes, d_in, d_out, elements, {}, 0);
+#if !TUNE_BASE
+  cub::detail::rfa::dispatch<input_it_t, output_it_t, int, init_t, transform_t, accum_t, policy_selector_t>(
+    nullptr, temp_storage_bytes, d_in, d_out, elements, {}, 0);
+#else
+  cub::detail::rfa::dispatch<input_it_t, output_it_t, int, init_t, transform_t, accum_t>(
+    nullptr, temp_storage_bytes, d_in, d_out, elements, {}, 0);
+#endif
```
oleksandr-pavlyk: Perhaps is preprocessor to define `auto rfa_dispatch` differently based on `TUNE_BASE` and use that throughout to reduce branching?


```suggestion
#if !TUNE_BASE
  auto rfa_dispatch = cub::detail::rfa::dispatch<input_it_t, output_it_t, int, init_t, transform_t, accum_t, policy_selector_t>;
#else
   auto rfa_dispatch = cub::detail::rfa::dispatch<input_it_t, output_it_t, int, init_t, transform_t, accum_t>;
#endif

    rfa_dispatch(nullptr, temp_storage_bytes, d_in, d_out, elements, {}, 0);
````

Then similarly for the actual invocation few lines below.

File: cub/benchmarks/bench/reduce/deterministic.cu (hunk 1 above)
Comment (bernhardmgruber) [reply cub/benchmarks/bench/reduce/deterministic.cu:20]: Sadly, this requires C++20 and I was told we won't upgrade the benchmarks yet :S

File: cub/benchmarks/bench/reduce/deterministic.cu (hunk 2 above)
Comment (bernhardmgruber) [reply cub/benchmarks/bench/reduce/deterministic.cu]: That's an interesting suggestion, but if I take the address of the template instantiation of `dispatch` when I need to later provide all function arguments (defaults no longer work). I didn't know that! I refactored it differently now.

File: cub/benchmarks/bench/reduce/deterministic.cu [hunk 3]
```diff
@@ -12,32 +12,13 @@
...
-  static constexpr cub::CacheLoadModifier LOAD_MODIFIER = cub::CacheLoadModifier::LOAD_DEFAULT;
-  constexpr static int ITEMS_PER_THREAD                 = TUNE_ITEMS_PER_THREAD;
-  constexpr static int BLOCK_THREADS                    = TUNE_THREADS_PER_BLOCK;
-};
-
-struct policy_hub_t
+struct policy_selector_t
 {
-  struct Policy350 : cub::ChainedPolicy<350, Policy350, Policy350>
+  [[nodiscard]] constexpr auto operator()(::cuda::arch_id) const -> cub::detail::rfa::rfa_policy
```
fbusato: question for my knowledge. Why this function is not marked with `_CCCL_*`?

cub/benchmarks/bench/reduce/deterministic.cu
```diff
@@ -76,13 +44,26 @@ void deterministic_sum(nvbench::state& state, nvbench::type_list<T>)
   state.add_global_memory_writes<T>(out.size());
 
   std::size_t temp_storage_bytes{};
-  dispatch_t::Dispatch(nullptr, temp_storage_bytes, d_in, d_out, elements, {}, 0);
+  // we explicitly provide template arguments to override accum_t
+  cub::detail::rfa::dispatch<input_it_t, output_it_t, int, init_t, transform_t, accum_t>(
+    nullptr, temp_storage_bytes, d_in, d_out, elements, init_t{}, /* stream */ 0);
```
fbusato:
```suggestion
    nullptr, temp_storage_bytes, d_in, d_out, elements, init_t{}, /* stream */ nullptr);
```
I prefer `nullptr` because it is a pointer

cub/cub/device/dispatch/kernels/kernel_reduce_deterministic.cuh
```diff
@@ -0,0 +1,249 @@
+// SPDX-FileCopyrightText: Copyright (c) 2024, NVIDIA CORPORATION. All rights reserved.
```
fbusato, line 1:
```suggestion
// SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION. All rights reserved.
```

cub/cub/device/dispatch/kernels/kernel_reduce_deterministic.cuh
```diff
@@ -0,0 +1,249 @@
...
+  constexpr int block_threads         = policy.block_threads;
+
+  using block_reduce_t = BlockReduce<AccumT, block_threads, policy.block_algorithm>;
+
+  // Shared memory storage
+  __shared__ typename block_reduce_t::TempStorage temp_storage;
+
+  using ftype              = typename AccumT::ftype;
+  constexpr int bin_length = AccumT::max_index + AccumT::max_fold;
+  const int tid            = block_threads * blockIdx.x + threadIdx.x;
```
fbusato, line 78: do we need to handle large input?

cub/cub/device/dispatch/kernels/kernel_reduce_deterministic.cuh
```diff
@@ -0,0 +1,249 @@
...
+
+  ftype* shared_bins = detail::rfa::get_shared_bin_array<ftype, bin_length>();
+
+  _CCCL_PRAGMA_UNROLL_FULL()
+  for (int index = threadIdx.x; index < bin_length; index += block_threads)
```
fbusato, line 83: we can rewrite this loop in a much better way: https://godbolt.org/z/dexnM8Y3a

fbusato: (probably out of scope for this PR)

cub/cub/device/dispatch/kernels/kernel_reduce_deterministic.cuh
```diff
@@ -0,0 +1,249 @@
...
+        thread_aggregate.renorm();
+        count = 0;
+      }
+    }
+  }
+
+  AccumT block_aggregate = block_reduce_t(temp_storage).Reduce(thread_aggregate, [](AccumT lhs, AccumT rhs) -> AccumT {
+    AccumT rtn = lhs;
+    rtn += rhs;
+    return rtn;
```
fbusato, line 134: why not just?
```suggestion
    return rhs + lhs;
```

cub/cub/device/dispatch/kernels/kernel_reduce_deterministic.cuh
```diff
@@ -0,0 +1,249 @@
...
+  });
+
+  // Output result
+  if (threadIdx.x == 0)
+  {
+    detail::uninitialized_copy_single(d_out + blockIdx.x, block_aggregate);
```
fbusato, line 140:
```suggestion
    cub::detail::uninitialized_copy_single(d_out + blockIdx.x, block_aggregate);
```

cub/cub/device/dispatch/tuning/tuning_reduce_deterministic.cuh
```diff
@@ -0,0 +1,108 @@
...
+
+  [[nodiscard]] _CCCL_API constexpr auto operator()(::cuda::arch_id arch) const -> rfa_policy
+  {
+    if (arch >= ::cuda::arch_id::sm_90)
+    {
+      // only tuned for float, fall through for other types
+      if (accum_t == type_t::float32)
+      {
+        // ipt_13.tpb_224  1.107188  1.009709  1.097114  1.316820
+        const auto scaled = scale_mem_bound(224, 13, accum_size);
```
fbusato, line 59: missing full qualification?

cub/benchmarks/bench/reduce/deterministic.cu
bernhardmgruber: It's only used in host code at runtime (by the CUB dispatch) and by the kernel at compile-time. However, adding `_CCCL_API` does not hurt.

cub/cub/device/dispatch/kernels/kernel_reduce_deterministic.cuh
bernhardmgruber: I did not add any new original content to this file, so the old copyright year should still apply to what is present in this file.

cub/cub/device/dispatch/kernels/kernel_reduce_deterministic.cuh
bernhardmgruber: I did not change any code in this file, just moved it from `kernel_reduce.cuh`. So I don't know. I guess it's fine.

cub/cub/device/dispatch/kernels/kernel_reduce_deterministic.cuh
bernhardmgruber: Great! Moved it to this issue: #7730

cub/cub/device/dispatch/kernels/kernel_reduce_deterministic.cuh
bernhardmgruber: Because `ReproducibleFloatingAccumulator` does not implement `operator+`. We could add it, but I don't want to mess with the kernel.

cub/cub/device/dispatch/tuning/tuning_reduce_deterministic.cuh
bernhardmgruber: We do not require full qualification outside libcu++ AFAIK

miscco: Image

cub/cub/device/dispatch/tuning/tuning_reduce_deterministic.cuh
```diff
@@ -0,0 +1,108 @@
...
+
+    if (arch >= ::cuda::arch_id::sm_60)
+    {
+      const auto scaled = scale_mem_bound(256, 16, accum_size);
+      return {{scaled.block_threads, scaled.items_per_thread, BLOCK_REDUCE_RAKING},
+              {scaled.block_threads, scaled.items_per_thread, BLOCK_REDUCE_RAKING}};
+    }
+
+    const auto scaled = scale_mem_bound(256, 20, accum_size);
+    return {{scaled.block_threads, scaled.items_per_thread, BLOCK_REDUCE_RAKING},
```
miscco, line 92: Just to be sure, are they *always* the same?

bernhardmgruber: Currently yes, but we could change this. It comes from the old `policy_hub`, where we had `using SingleTilePolicy = ReducePolicy;` everywhere.

davebayer (approved):

PR #7358, opened 2026-01-26
Assert before terminating on `throw` in device code

davebayer:
Currently, if an exception is thrown in device code, we just call `cuda::std::terminate`. There is basically no way to find out what went wrong.

This PR tries to change that and make a small step towards the _"Oh, I have an idea what went wrong.."_ goal by adding `_CCCL_ASSERT` before the `cuda::std::terminate()` call, providing a message with the name of the exception that would be thrown.

To make this possible, I had to change the implementation of the `_CCCL_THROW` macro, so we can convert the type name to string even if the type itself is not available (for example when compiling with NVRTC).

There is also another hidden change. If an assertion happens, the CUDA API calls will return `cudaErrorAssert` instead of the one returned after `__trap()` (I cannot remember which one is it, but it's different).

As the title says, this is an assertion, so if the user disables assertions in CCCL, just `cuda::std::terminate()` is called as usually.

### DIFF
--- a/docs/cccl/development/macro.rst
+++ b/docs/cccl/development/macro.rst
@@ -345,17 +345,17 @@ In C++23, the ``if consteval`` statement (`link <https://en.cppreference.com/w/c
 
 CUDA doesn't support exceptions in device code, however, sometimes we need to write host/device functions that use exceptions on host and ``__trap()`` on device. CCCL provides a set of macros that should be used in place of the standard C++ keywords to make the code compile in both, host and device code.
 
-+-----------------------------+-----------------------------------------------------------------------------------------------------------------+
-| ``_CCCL_TRY``               | Replacement for the ``try`` keyword                                                                             |
-+-----------------------------+-----------------------------------------------------------------------------------------------------------------+
-| ``_CCCL_CATCH (X)``         | Replacement for the ``catch (/*X*/)`` statement                                                                 |
-+-----------------------------+-----------------------------------------------------------------------------------------------------------------+
-| ``_CCCL_CATCH_ALL``         | Replacement for the ``catch (...)`` statement                                                                   |
-+-----------------------------+-----------------------------------------------------------------------------------------------------------------+
-| ``_CCCL_THROW(X, ...)``     | Replacement for the ``throw X(...)`` expression, ``X`` must be fully qualified type, without the leading ``::`` |
-+-----------------------------+-----------------------------------------------------------------------------------------------------------------+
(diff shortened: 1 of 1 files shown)


docs/cccl/development/macro.rst
```diff
@@ -345,17 +345,17 @@ In C++23, the ``if consteval`` statement (`link <https://en.cppreference.com/w/c
...
++-----------------------------+-----------------------------------------------------------------------------------------------------------------+
+| ``_CCCL_TRY``               | Replacement for the ``try`` keyword                                                                             |
++-----------------------------+-----------------------------------------------------------------------------------------------------------------+
+| ``_CCCL_CATCH (X)``         | Replacement for the ``catch (/*X*/)`` statement                                                                 |
++-----------------------------+-----------------------------------------------------------------------------------------------------------------+
+| ``_CCCL_CATCH_ALL``         | Replacement for the ``catch (...)`` statement                                                                   |
++-----------------------------+-----------------------------------------------------------------------------------------------------------------+
+| ``_CCCL_THROW(X, ...)``     | Replacement for the ``throw X(...)`` expression, ``X`` must be fully qualified type, without the leading ``::`` |
```
fbusato:
```suggestion
| ``_CCCL_THROW(X, ...)``     | Replacement for the ``throw X(...)`` expression. ``X`` must be a fully qualified type, without the leading ``::`` |
```

libcudacxx/include/cuda/std/__exception/exception_macros.h
```diff
@@ -88,17 +86,21 @@ _CCCL_END_NAMESPACE_CUDA_STD
...
+                        ({                                                                                       \
+                          ::fprintf(stderr,                                                                      \
+                                    "%s:%u: An instance of class %s would be thrown.\n  what():  %s\nAborted\n", \
+                                    __FILE__,                                                                    \
+                                    __LINE__,                                                                    \
+                                    #_TYPE,                                                                      \
+                                    (_TYPE(__VA_ARGS__)).what());                                                \
+                          ::fflush(stderr);                                                                      \
+                        }),                                                                                      \
+                        ({ _CCCL_ASSERT(false, "An instance of class " #_TYPE " would be thrown."); }))          \
```
fbusato, line 102: could you please do a quick search to see if the pattern `_CCCL_ASSERT, _CCCL_THROW;` is present somewhere?

davebayer: We don't. And it would be a mistake, because the behaviour would differ with and without assertions, you wouldn't be able to catch the exception.

PR #7359, opened 2026-01-26
Avoid compile errors from math function like macros

miscco:
Some libc implementation implement certain functions as macros

This might break our build because the definition of the function might clash with the macro and also we cannot qualify a macro, so ::isgreater will fail if isgreater is a macro.

To work around this we need to temporarily push / pop the macros and define the fallbacks accordingly.

This is fine because the functions themself can only be called through `cuda::std::isgreater` which avoids the macro issue

### DIFF
--- a/libcudacxx/include/cuda/std/__cmath/fpclassify.h
+++ b/libcudacxx/include/cuda/std/__cmath/fpclassify.h
@@ -70,6 +70,27 @@
 #  endif // !__FP_LOGBNAN_MIN
 #endif // !FP_ILOGBNAN
 
+#if !_CCCL_COMPILER(NVRTC)
+// This function may or may not be implemented as a function macro so we need to handle that case
+#  ifdef fpclassify
+
+template <class _Tp>
+[[nodiscard]] _CCCL_HOST_API int __cccl_fpclassify_runtime(_Tp __x) noexcept
+{
+  return fpclassify(__x);
+}
+#    pragma push_macro("fpclassify")
+#    undef fpclassify
+#    define _CCCL_POP_MACRO_fpclassify
+#  else // ^^^ Function Macro fpclassify ^^^ / vvv No function macro fpclassify vvv
+template <class _Tp>
+[[nodiscard]] _CCCL_HOST_API int __cccl_fpclassify_runtime(_Tp __x) noexcept
+{
+  return ::fpclassify(__x);
+}
+#  endif // No function macro fpclassify
+#endif // _CCCL_COMPILER(NVRTC)
+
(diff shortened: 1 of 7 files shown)


libcudacxx/include/cuda/std/__cmath/fpclassify.h
```diff
@@ -70,6 +70,25 @@
 #  endif // !__FP_LOGBNAN_MIN
 #endif // !FP_ILOGBNAN
 
+// This function may or may not be implemented as a function macro so we need to handle that case
+#ifdef fpclassify
```
davebayer: This is bad, just pushing/popping the macro doesn't solve the problem, users will have to `#undef` the macro to be able to use this.

davebayer (commented):
There must be a problem somewhere, because if you `#include <math.h>` for example, you should be including the C++ compatibility headers, not the C headers directly.

If this doesn't help, can't we just include the `<cmeow>` headers instead? There should always be the macros undefined I believe

libcudacxx/include/cuda/std/__cmath/fpclassify.h
miscco: I dont think so, because the only way the user can actually use our functions is through `cuda::std::meow(...)` which is a different token sequence than `meow(...)`

davebayer: See https://godbolt.org/z/eGMs4Grro

davebayer: + they should be able to use `using namespace cuda::std;` for example

miscco: I am fine with that

miscco: maybe we just need to undefine the macros?

davebayer: I think we need to find out what exactly happens. As I said, the compatibility headers should undefine the macros. But I think we can really try just including the `<cmeow>` headers, they will undefine the macros 100 %

miscco: I believe we already tried including `<cmath>` but that did not work. This should unblock holoscan to build I hope

davebayer (changes requested):
For now

miscco:
Closing this, we came to the conclusion that we cannot do anything here, because we do rely on C++ <math.h> to bring in the function definitions into the global namespace.

The issue at hand is that in the holoscan suite the C-math.h include has priority so we need to figure out the build system changes

PR #7364, opened 2026-01-27
Expose CUDA vector type traits

fbusato:
## Description

Several CUDA libraries use CUDA vector types and define their own type traits. The goal of the PR is to expose common type traits:

- `vector_type_t<T, N>`: (previously internal) corresponding vector type of `T` with size `N`.
- `scalar_type_t`: corresponding scalar type, forward to `tuple_element_t`.
- `is_vector_type_v`: determines if a type is a CUDA vector type.
- `vector_size_v`: size of a CUDAvector, , forward to `tuple_size_v`.

### DIFF
--- a/docs/libcudacxx/extended_api/type_traits/vector_types.rst
+++ b/docs/libcudacxx/extended_api/type_traits/vector_types.rst
@@ -32,7 +32,7 @@ Defined in the ``<cuda/type_traits>`` header.
 - ``longlong1``, ``longlong2``, ``longlong3``, ``longlong4``, and alignment variants in CUDA Toolkit 13.0+.
 - ``ulonglong1``, ``ulonglong2``, ``ulonglong3``, ``ulonglong4``, and alignment variants in CUDA Toolkit 13.0+.
 - ``float1``, ``float2``, ``float3``, ``float4``.
-- ``double1``, ``double2``, ``double3``, ``double4`` (and alignment variants in CUDA Toolkit 13.0+).
+- ``double1``, ``double2``, ``double3``, ``double4``, and alignment variants in CUDA Toolkit 13.0+.
 - ``dim3``.
 
 ``cuda::is_extended_fp_vector_type_v``


docs/libcudacxx/extended_api/type_traits/vector_types.rst
```diff
@@ -0,0 +1,117 @@
...
+- ``short1``, ``short2``, ``short3``, ``short4``.
+- ``ushort1``, ``ushort2``, ``ushort3``, ``ushort4``.
+- ``int1``, ``int2``, ``int3``, ``int4``.
+- ``uint1``, ``uint2``, ``uint3``, ``uint4``.
+- ``long1``, ``long2``, ``long3``, ``long4``, and alignment variants in CUDA Toolkit 13.0+.
+- ``ulong1``, ``ulong2``, ``ulong3``, ``ulong4``, and alignment variants in CUDA Toolkit 13.0+.
+- ``longlong1``, ``longlong2``, ``longlong3``, ``longlong4``, and alignment variants in CUDA Toolkit 13.0+.
+- ``ulonglong1``, ``ulonglong2``, ``ulonglong3``, ``ulonglong4``, and alignment variants in CUDA Toolkit 13.0+.
+- ``float1``, ``float2``, ``float3``, ``float4``.
+- ``double1``, ``double2``, ``double3``, ``double4`` (and alignment variants in CUDA Toolkit 13.0+).
```
bernhardmgruber: Please put the comment about the alignment variants consistently inside parentheses or not.

docs/libcudacxx/extended_api/type_traits.rst
```diff
@@ -22,3 +23,8 @@ Type traits
      - Tells whether a type is a floating point type
      - CCCL 3.0.0
      - CUDA 13.0
+
+   * - :ref:`Vector Type Traits <libcudacxx-extended-api-type_traits-vector_types>`
+     - Type traits for CUDA vector types
+     - CCCL 3.0.0
+     - CUDA 13.0
```
bernhardmgruber:
```suggestion
   * - :ref:`Vector Type Traits <libcudacxx-extended-api-type_traits-vector_types>`
     - Type traits for CUDA vector types
     - CCCL 3.3.0
     - CUDA 13.3
```

bernhardmgruber (approved):

libcudacxx/include/cuda/__type_traits/vector_type.h
davebayer, line 197: Wait, so we agreed that for CTK < 13 this should be a different type than for CTK >= 13? Because it was an internal trait before, we agreed that we don't care too much, but as this becomes public, I think it matters.

Shouldn't the trait have the alignment parameter than? This is the reason I didn't want to expose this, because it's unclear what should be the default.

I'd like to have a discussion about this

docs/libcudacxx/extended_api/type_traits/vector_types.rst
```diff
@@ -0,0 +1,117 @@
...
+.. code:: cuda
+
+   namespace cuda {
+
+   template <class T, cuda::std::size_t Size>
+   using vector_type_t = /* see below */;
+
+   } // namespace cuda
+
+``cuda::vector_type_t<T, Size>`` is a type alias that maps a scalar type ``T`` and a vector size ``Size`` to the corresponding CUDA vector type. If no valid vector type exists for the given combination, the type is ``void``. It supports both integral and floating-point vector types.
```
davebayer, line 73: The alignment behaviour is undocumented

libcudacxx/include/cuda/__type_traits/vector_type.h
bernhardmgruber: Sorry that I missed this. I looked at the changeset of this PR and it LGTM. I didn't know the internal implementation had an issue.

We should definitely aim for `vector_type_t<long, 4>` to be the same type across different CTKs. We could do this by defining `long4_32a` ourselves < 13.0. But maybe there is a different solution.

docs/libcudacxx/extended_api/type_traits/vector_types.rst
```diff
@@ -0,0 +1,117 @@
...
+.. code:: cuda
+
+   namespace cuda {
+
+   template <class VectorType>
+   using scalar_type_t = /* see below */;
+
+   } // namespace cuda
+
+``cuda::scalar_type_t<VectorType>`` extracts the scalar element type from a CUDA vector type. For example, ``cuda::scalar_type_t<int2>`` is ``int`` and ``cuda::scalar_type_t<float4>`` is ``float``. . It supports both integral and floating-point vector types.
```
davebayer, line 87: Shouldn't this be `cuda::element_type_t`? Shouldn't we provide the `struct` traits as well?

libcudacxx/include/cuda/__type_traits/vector_type.h
fbusato: I agree in principle, but I don't see space for alternatives.

- We cannot use `long4` in CTK 13 because is deprecated.
- We cannot use `long4_32a` in CTK 12 because it is not available.

What could be the implications? On the user side, I would assume it is handled symmetrically, with  `long4`/ `long4_32a` dispatch.

docs/libcudacxx/extended_api/type_traits/vector_types.rst
fbusato: I think `scalar_type_t` is more precise as we are dealing with numerical values. 
what is the use case for `struct` traits that cannot be achieved combining alias + availability traits (`constexpr bool`) ?

Issue #7092, opened 2026-01-06
Add `BlockTopK` that is based on AIR Top-K


elstehle:
The goal here is to have a block-level top-k implementation that we can use inside `DeviceSegmentedTopK` for small enough segments. We plan to have the `BlockTopK` implementation in the `detail` namespace for now until we feel confident that we can settle on the interface and expose it publicly.

PR #4584, opened 2025-04-29
move support for environments from `cuda::experimental` to `cuda::std::execution`

ericniebler:
## Description

@gevtushenko's work requires non-experimental support for unordered collections of key/value pairs, aka environments. The current working draft of C++26 has such utilities in the `std::execution` namespace.

- [`prop`](https://eel.is/c++draft/exec#prop): creates a singleton environment from a key/value pair
- [`env`](https://eel.is/c++draft/exec#env): creates an environment from a variadic pack of environments `envs...`, where `env{envs...}.query(Q)` returns `e.query(Q)` where `e` is the first environment in `envs` for which that expression is well-formed.

this PR adds `prop` and `env` to the `cuda::std::execution` namespace, as well as `get_env` for asking an object for its associated environment.

WIP needs docs and tests

## Checklist

- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #7384, opened 2026-01-27
Radix-selection based `BlockTopK` specialization

elstehle:
## Description

This PR adds the first specialized `BlockTopK` implementation based on radix selection.

**Note to reviewers:**
- The goal here is to have a block-level top-k implementation that we can use inside `DeviceSegmentedTopK` for small-enough segments.
- The **interface** for the **block-level primitive**, i.e., `BlockTopK`, itself is still evolving. It's not a concern for now. Determining the the exact interface and exposing `BlockTopK` publicly is follow-up work.
- I am collecting potential further improvements in https://github.com/NVIDIA/cccl/issues/7799

## How the algorithm works:

The algorithm has two stages: **(a) radix selection** identifies the bit-prefix of the k-th key, and **(b) partitioning** uses that prefix to scatter the top-k items into the first k output positions.

**(a) Radix selection** (`get_kth_key_prefix`): Processing bits from most-significant to least-significant, `RadixBits` at a time, the algorithm progressively narrows down which items could be the k-th element. In each pass, it builds a histogram of the current radix digit over only the *candidate* items (i.e., items whose more-significant bits match the prefix determined so far), then computes a prefix sum to find cumulative bucket counts. From these counts, it identifies the bucket into which the k-th candidate falls. All items in preceding buckets are guaranteed to be in the top-k ("selected"). All items in subsequent buckets are guaranteed *not* to be in the top-k. Items within the chosen bucket become the candidates for the next pass, and k is adjusted to be relative within these remaining candidates. If the number of remaining candidates exactly equals the remaining k, all candidates qualify and the algorithm short-circuits.

No data movement happens during this stage, items remain in their original registers throughout all passes. Only the histogram (in shared memory) is rebuilt each pass.

**(b) Partitioning** (`select_topk`): Once the bit-prefix of the k-th key is known, items are physically gathered into shared memory to produce the output. There are two cases:
- **All candidates qualify** (the remaining candidates exactly fill the remaining k slots, or ties are expanded): every item whose key prefix <= the k-th prefix is written into the output buffer using an atomic counter for write-coordination.
- **More candidates than remaining slots**: items with key prefix strictly less than the k-th prefix are scattered first (they are unconditionally top-k), then items with key prefix equal to the k-th prefix fill the remaining slots up to k.

Finally, each thread reads back its portion of the top-k items from shared memory. For key-value pairs, keys and values are scattered in two separate phases, reusing the same shared memory buffer.

### An example: 8-bit keys, 4-bit radix digits

The example is using hex digits (4 bits per pass) to give some intuition on how the algorithm works.

```
Input:  10 unsigned 8-bit keys (hex),  k = 5  (find the 5 smallest)

   B3  52  7A  B8  55  A4  74  58  B0  71
```

**Pass 1: inspect the upper hex digit (bits 7..4)**

Build a histogram of the upper nibble across **all** items (all are candidates initially), then take an inclusive prefix sum:

```
  Digit │ Items          │ Count │ Prefix Sum
  ──────┼────────────────┼───────┼───────────
    5   │ 52, 55, 58     │   3   │     3
    7   │ 7A, 74, 71     │   3   │     6
    A   │ A4             │   1   │     7
    B   │ B3, B8, B0     │   3   │    10
```

Find the bucket whose prefix-sum interval contains k = 5:

```
  Bucket '5': prefix_sum = 3         →  3 < 5   (too few)
  Bucket '7': prefix_sum = 6         →  6 ≥ 5   ← k falls here
```

Outcome:

```
  Selected  (bucket < '7'):  52, 55, 58       -> guaranteed top-5
  Candidates (bucket = '7'): 7A, 74, 71      -> need further refinement
  Rejected  (bucket > '7'):  A4, B3, B8, B0  -> not top-5

  total_selected = 3
  k_remaining    = 5 − 3 = 2
  kth_prefix     = 0x7·          (upper nibble resolved) - in the next pass, we only consider items with this prefix 
  prefix_mask    = 0xF0
```

**Pass 2: inspect the lower hex digit (bits 3..0), candidates only**

Build a histogram of the lower nibble among the 3 candidates **{7A, 74, 71}** only:

```
  Digit │ Items │ Count │ Prefix Sum
  ──────┼───────┼───────┼───────────
    1   │ 71    │   1   │     1
    4   │ 74    │   1   │     2
    A   │ 7A    │   1   │     3
```

Find the bucket for k\_remaining = 2:

```
  Bucket 1: prefix_sum = 1         →  1 < 2   (too few)
  Bucket 4: prefix_sum = 2         →  2 ≥ 2   ← k falls here
```

Outcome:

```
  Selected  (bucket < 4):  71               — guaranteed top-5
  Candidates (bucket = 4): 74               — sole remaining candidate
  Rejected  (bucket > 4):  7A               — not top-5

  total_selected = 3 + 1 = 4
  k_remaining    = 2 − 1 = 1
  candidates     = 1
```

**Short-circuit:** `candidates == k_remaining` (1 == 1), so the single remaining candidate (74) necessarily qualifies. No further passes needed.

```
  kth_prefix  = 0x74
  prefix_mask = 0xFF          (all bits resolved)
```

**Selection/Filtering: gather the top-5**

Since `total_selected + candidates == k` (4 + 1 = 5), *all* candidates qualify. Every item whose `(key & 0xFF) ≤ 0x74` is written to shared memory using an atomic counter for write-coordination:

```
  Shared-memory buffer (unordered):  52  55  58  71  74
```

Each thread reads back its portion of the output from shared memory.

**Summary of what each pass accomplishes:**

```
  Pass │ Radix digit │ Selected  │ Candidates │ Rejected │ k_remaining
  ─────┼─────────────┼───────────┼────────────┼──────────┼────────────
    1  │ upper hex   │ 52,55,58  │ 7A,74,71   │ 4 items  │ 2
    2  │ lower hex   │ 71        │ 74         │ 7A       │ 1 (= cands → done)
```

Each pass eliminates items that are provably not in the top-k (rejected) and promotes items that are provably in the top-k (selected), without moving any data. Only the histogram is rebuilt. After all passes, a single partitioning scatter produces the final result.

Closes https://github.com/NVIDIA/cccl/issues/7092.

<details>

<summary>Benchmark results on B200 (table) against `main` (using `BlockRadixSort`)</summary>

## [0] NVIDIA B200

|  KeyT{ct}  |  MaxSegmentSize{ct}  |  MaxNumSelected{ct}  |  Elements{io}  |  Entropy  |   Ref Time |   Ref Noise |   Cmp Time |   Cmp Noise |          Diff |   %Diff |  Status  |
|------------|----------------------|----------------------|----------------|-----------|------------|-------------|------------|-------------|---------------|---------|----------|
|    F32     |          64          |          8           |      2^28      |     1     |  38.300 ms |       0.01% |  22.235 ms |       0.01% | -16064.979 us | -41.95% |   FAST   |
|    F32     |          64          |          8           |      2^28      |   0.544   |  38.297 ms |       0.01% |  22.236 ms |       0.01% | -16060.992 us | -41.94% |   FAST   |
|    F32     |          64          |          8           |      2^28      |   0.201   |  38.298 ms |       0.01% |  22.236 ms |       0.01% | -16062.055 us | -41.94% |   FAST   |
|    F32     |          64          |          8           |      2^28      |     0     |  38.298 ms |       0.01% |  22.237 ms |       0.01% | -16060.796 us | -41.94% |   FAST   |
|    F32     |          64          |          16          |      2^28      |     1     |  38.319 ms |       0.01% |  22.208 ms |       0.01% | -16111.194 us | -42.05% |   FAST   |
|    F32     |          64          |          16          |      2^28      |   0.544   |  38.319 ms |       0.01% |  22.209 ms |       0.01% | -16110.596 us | -42.04% |   FAST   |
|    F32     |          64          |          16          |      2^28      |   0.201   |  38.319 ms |       0.01% |  22.208 ms |       0.01% | -16110.640 us | -42.04% |   FAST   |
|    F32     |          64          |          16          |      2^28      |     0     |  38.318 ms |       0.01% |  22.208 ms |       0.01% | -16109.546 us | -42.04% |   FAST   |
|    F32     |          64          |          32          |      2^28      |     1     |  38.386 ms |       0.01% |  22.161 ms |       0.01% | -16225.085 us | -42.27% |   FAST   |
|    F32     |          64          |          32          |      2^28      |   0.544   |  38.388 ms |       0.01% |  22.160 ms |       0.01% | -16227.911 us | -42.27% |   FAST   |
|    F32     |          64          |          32          |      2^28      |   0.201   |  38.386 ms |       0.01% |  22.160 ms |       0.01% | -16225.893 us | -42.27% |   FAST   |
|    F32     |          64          |          32          |      2^28      |     0     |  38.389 ms |       0.00% |  22.162 ms |       0.01% | -16227.566 us | -42.27% |   FAST   |
|    F32     |         128          |          8           |      2^28      |     1     |  19.256 ms |       0.01% |  11.303 ms |       0.01% |  -7952.999 us | -41.30% |   FAST   |
|    F32     |         128          |          8           |      2^28      |   0.544   |  19.256 ms |       0.01% |  11.302 ms |       0.01% |  -7953.960 us | -41.31% |   FAST   |
|    F32     |         128          |          8           |      2^28      |   0.201   |  19.256 ms |       0.01% |  11.302 ms |       0.01% |  -7953.992 us | -41.31% |   FAST   |
|    F32     |         128          |          8           |      2^28      |     0     |  19.255 ms |       0.01% |  11.302 ms |       0.01% |  -7952.808 us | -41.30% |   FAST   |
|    F32     |         128          |          16          |      2^28      |     1     |  19.264 ms |       0.01% |  11.291 ms |       0.01% |  -7973.516 us | -41.39% |   FAST   |
|    F32     |         128          |          16          |      2^28      |   0.544   |  19.265 ms |       0.01% |  11.292 ms |       0.01% |  -7973.287 us | -41.39% |   FAST   |
|    F32     |         128          |          16          |      2^28      |   0.201   |  19.264 ms |       0.01% |  11.290 ms |       0.01% |  -7974.915 us | -41.40% |   FAST   |
|    F32     |         128          |          16          |      2^28      |     0     |  19.265 ms |       0.01% |  11.292 ms |       0.01% |  -7973.760 us | -41.39% |   FAST   |
|    F32     |         128          |          32          |      2^28      |     1     |  19.293 ms |       0.01% |  11.325 ms |       0.01% |  -7968.653 us | -41.30% |   FAST   |
|    F32     |         128          |          32          |      2^28      |   0.544   |  19.292 ms |       0.01% |  11.324 ms |       0.01% |  -7968.978 us | -41.31% |   FAST   |
|    F32     |         128          |          32          |      2^28      |   0.201   |  19.294 ms |       0.01% |  11.324 ms |       0.01% |  -7969.603 us | -41.31% |   FAST   |
|    F32     |         128          |          32          |      2^28      |     0     |  19.293 ms |       0.01% |  11.324 ms |       0.01% |  -7969.286 us | -41.31% |   FAST   |
|    F32     |         256          |          8           |      2^28      |     1     |  10.311 ms |       0.01% |   5.751 ms |       0.02% |  -4559.354 us | -44.22% |   FAST   |
|    F32     |         256          |          8           |      2^28      |   0.544   |  10.309 ms |       0.01% |   5.751 ms |       0.02% |  -4558.456 us | -44.22% |   FAST   |
|    F32     |         256          |          8           |      2^28      |   0.201   |  10.310 ms |       0.01% |   5.751 ms |       0.02% |  -4558.817 us | -44.22% |   FAST   |
|    F32     |         256          |          8           |      2^28      |     0     |  10.309 ms |       0.01% |   5.751 ms |       0.02% |  -4558.186 us | -44.21% |   FAST   |
|    F32     |         256          |          16          |      2^28      |     1     |  10.321 ms |       0.01% |   5.724 ms |       0.02% |  -4596.405 us | -44.54% |   FAST   |
|    F32     |         256          |          16          |      2^28      |   0.544   |  10.322 ms |       0.01% |   5.724 ms |       0.02% |  -4597.624 us | -44.54% |   FAST   |
|    F32     |         256          |          16          |      2^28      |   0.201   |  10.321 ms |       0.01% |   5.724 ms |       0.02% |  -4597.239 us | -44.54% |   FAST   |
|    F32     |         256          |          16          |      2^28      |     0     |  10.322 ms |       0.01% |   5.724 ms |       0.02% |  -4597.923 us | -44.54% |   FAST   |
|    F32     |         256          |          32          |      2^28      |     1     |  10.350 ms |       0.01% |   5.757 ms |       0.02% |  -4592.951 us | -44.38% |   FAST   |
|    F32     |         256          |          32          |      2^28      |   0.544   |  10.349 ms |       0.01% |   5.757 ms |       0.02% |  -4591.825 us | -44.37% |   FAST   |
|    F32     |         256          |          32          |      2^28      |   0.201   |  10.349 ms |       0.01% |   5.756 ms |       0.02% |  -4592.447 us | -44.38% |   FAST   |
|    F32     |         256          |          32          |      2^28      |     0     |  10.349 ms |       0.01% |   5.757 ms |       0.02% |  -4592.075 us | -44.37% |   FAST   |
|    F32     |         256          |         128          |      2^28      |     1     |  10.415 ms |       0.01% |   5.790 ms |       0.02% |  -4625.012 us | -44.41% |   FAST   |
|    F32     |         256          |         128          |      2^28      |   0.544   |  10.416 ms |       0.01% |   5.789 ms |       0.02% |  -4626.536 us | -44.42% |   FAST   |
|    F32     |         256          |         128          |      2^28      |   0.201   |  10.415 ms |       0.01% |   5.789 ms |       0.02% |  -4626.136 us | -44.42% |   FAST   |
|    F32     |         256          |         128          |      2^28      |     0     |  10.416 ms |       0.01% |   5.790 ms |       0.02% |  -4626.015 us | -44.41% |   FAST   |
|    F32     |         512          |          8           |      2^28      |     1     |   5.376 ms |       0.02% |   3.068 ms |       0.04% |  -2307.423 us | -42.92% |   FAST   |
|    F32     |         512          |          8           |      2^28      |   0.544   |   5.376 ms |       0.02% |   3.068 ms |       0.04% |  -2307.279 us | -42.92% |   FAST   |
|    F32     |         512          |          8           |      2^28      |   0.201   |   5.376 ms |       0.02% |   3.068 ms |       0.04% |  -2307.956 us | -42.93% |   FAST   |
|    F32     |         512          |          8           |      2^28      |     0     |   5.376 ms |       0.02% |   3.068 ms |       0.04% |  -2307.335 us | -42.92% |   FAST   |
|    F32     |         512          |          16          |      2^28      |     1     |   5.379 ms |       0.02% |   3.092 ms |       0.04% |  -2286.462 us | -42.51% |   FAST   |
|    F32     |         512          |          16          |      2^28      |   0.544   |   5.379 ms |       0.02% |   3.092 ms |       0.03% |  -2286.520 us | -42.51% |   FAST   |
|    F32     |         512          |          16          |      2^28      |   0.201   |   5.379 ms |       0.02% |   3.092 ms |       0.03% |  -2286.960 us | -42.51% |   FAST   |
|    F32     |         512          |          16          |      2^28      |     0     |   5.379 ms |       0.02% |   3.092 ms |       0.03% |  -2286.624 us | -42.51% |   FAST   |
|    F32     |         512          |          32          |      2^28      |     1     |   5.387 ms |       0.02% |   3.123 ms |       0.03% |  -2263.563 us | -42.02% |   FAST   |
|    F32     |         512          |          32          |      2^28      |   0.544   |   5.388 ms |       0.02% |   3.123 ms |       0.04% |  -2264.293 us | -42.03% |   FAST   |
|    F32     |         512          |          32          |      2^28      |   0.201   |   5.387 ms |       0.02% |   3.123 ms |       0.04% |  -2263.766 us | -42.02% |   FAST   |
|    F32     |         512          |          32          |      2^28      |     0     |   5.388 ms |       0.02% |   3.123 ms |       0.04% |  -2264.094 us | -42.02% |   FAST   |
|    F32     |         512          |         128          |      2^28      |     1     |   5.420 ms |       0.02% |   3.110 ms |       0.04% |  -2309.788 us | -42.61% |   FAST   |
|    F32     |         512          |         128          |      2^28      |   0.544   |   5.420 ms |       0.02% |   3.110 ms |       0.03% |  -2309.677 us | -42.61% |   FAST   |
|    F32     |         512          |         128          |      2^28      |   0.201   |   5.420 ms |       0.02% |   3.111 ms |       0.03% |  -2309.907 us | -42.61% |   FAST   |
|    F32     |         512          |         128          |      2^28      |     0     |   5.420 ms |       0.02% |   3.110 ms |       0.04% |  -2309.517 us | -42.61% |   FAST   |
|    F32     |         1024         |          8           |      2^28      |     1     |   2.905 ms |       0.03% |   1.890 ms |       0.07% |  -1014.991 us | -34.94% |   FAST   |
|    F32     |         1024         |          8           |      2^28      |   0.544   |   2.905 ms |       0.04% |   1.890 ms |       0.06% |  -1015.096 us | -34.94% |   FAST   |
|    F32     |         1024         |          8           |      2^28      |   0.201   |   2.905 ms |       0.03% |   1.890 ms |       0.07% |  -1015.398 us | -34.95% |   FAST   |
|    F32     |         1024         |          8           |      2^28      |     0     |   2.905 ms |       0.03% |   1.890 ms |       0.06% |  -1014.862 us | -34.93% |   FAST   |
|    F32     |         1024         |          16          |      2^28      |     1     |   2.907 ms |       0.04% |   1.908 ms |       0.06% |   -998.828 us | -34.36% |   FAST   |
|    F32     |         1024         |          16          |      2^28      |   0.544   |   2.907 ms |       0.04% |   1.908 ms |       0.06% |   -998.986 us | -34.36% |   FAST   |
|    F32     |         1024         |          16          |      2^28      |   0.201   |   2.907 ms |       0.04% |   1.908 ms |       0.06% |   -999.171 us | -34.37% |   FAST   |
|    F32     |         1024         |          16          |      2^28      |     0     |   2.907 ms |       0.04% |   1.908 ms |       0.06% |   -998.759 us | -34.36% |   FAST   |
|    F32     |         1024         |          32          |      2^28      |     1     |   2.910 ms |       0.04% |   1.905 ms |       0.06% |  -1005.538 us | -34.55% |   FAST   |
|    F32     |         1024         |          32          |      2^28      |   0.544   |   2.910 ms |       0.03% |   1.905 ms |       0.06% |  -1005.656 us | -34.55% |   FAST   |
|    F32     |         1024         |          32          |      2^28      |   0.201   |   2.910 ms |       0.04% |   1.905 ms |       0.06% |  -1005.695 us | -34.55% |   FAST   |
|    F32     |         1024         |          32          |      2^28      |     0     |   2.910 ms |       0.03% |   1.905 ms |       0.06% |  -1005.581 us | -34.55% |   FAST   |
|    F32     |         1024         |         128          |      2^28      |     1     |   2.930 ms |       0.03% |   1.895 ms |       0.05% |  -1035.599 us | -35.34% |   FAST   |
|    F32     |         1024         |         128          |      2^28      |   0.544   |   2.930 ms |       0.04% |   1.895 ms |       0.05% |  -1035.448 us | -35.33% |   FAST   |
|    F32     |         1024         |         128          |      2^28      |   0.201   |   2.931 ms |       0.03% |   1.895 ms |       0.05% |  -1035.544 us | -35.34% |   FAST   |
|    F32     |         1024         |         128          |      2^28      |     0     |   2.930 ms |       0.03% |   1.895 ms |       0.06% |  -1035.171 us | -35.33% |   FAST   |
|    F32     |         1024         |         512          |      2^28      |     1     |   2.978 ms |       0.03% |   1.896 ms |       0.06% |  -1082.667 us | -36.35% |   FAST   |
|    F32     |         1024         |         512          |      2^28      |   0.544   |   2.979 ms |       0.04% |   1.895 ms |       0.06% |  -1083.437 us | -36.37% |   FAST   |
|    F32     |         1024         |         512          |      2^28      |   0.201   |   2.978 ms |       0.03% |   1.895 ms |       0.06% |  -1083.000 us | -36.37% |   FAST   |
|    F32     |         1024         |         512          |      2^28      |     0     |   2.979 ms |       0.04% |   1.895 ms |       0.05% |  -1083.066 us | -36.36% |   FAST   |

</details>

```

## [0] NVIDIA B200

|  KeyT{ct}  |  MaxSegmentSize{ct}  |  MaxNumSelected{ct}  |  Elements{io}  |  Entropy  |   Ref Time |   Ref Noise |   Cmp Time |   Cmp Noise |          Diff |   %Diff |  Status  |
|------------|----------------------|----------------------|----------------|-----------|------------|-------------|------------|-------------|---------------|---------|----------|
|    F32     |          64          |          8           |      2^28      |     1     |  38.300 ms |       0.01% |  22.235 ms |       0.01% | -16064.979 us | -41.95% |   FAST   |
|    F32     |          64          |          8           |      2^28      |   0.544   |  38.297 ms |       0.01% |  22.236 ms |       0.01% | -16060.992 us | -41.94% |   FAST   |
|    F32     |          64          |          8           |      2^28      |   0.201   |  38.298 ms |       0.01% |  22.236 ms |       0.01% | -16062.055 us | -41.94% |   FAST   |
|    F32     |          64          |          8           |      2^28      |     0     |  38.298 ms |       0.01% |  22.237 ms |       0.01% | -16060.796 us | -41.94% |   FAST   |
|    F32     |          64          |          16          |      2^28      |     1     |  38.319 ms |       0.01% |  22.208 ms |       0.01% | -16111.194 us | -42.05% |   FAST   |
|    F32     |          64          |          16          |      2^28      |   0.544   |  38.319 ms |       0.01% |  22.209 ms |       0.01% | -16110.596 us | -42.04% |   FAST   |
|    F32     |          64          |          16          |      2^28      |   0.201   |  38.319 ms |       0.01% |  22.208 ms |       0.01% | -16110.640 us | -42.04% |   FAST   |
|    F32     |          64          |          16          |      2^28      |     0     |  38.318 ms |       0.01% |  22.208 ms |       0.01% | -16109.546 us | -42.04% |   FAST   |
|    F32     |          64          |          32          |      2^28      |     1     |  38.386 ms |       0.01% |  22.161 ms |       0.01% | -16225.085 us | -42.27% |   FAST   |
|    F32     |          64          |          32          |      2^28      |   0.544   |  38.388 ms |       0.01% |  22.160 ms |       0.01% | -16227.911 us | -42.27% |   FAST   |
|    F32     |          64          |          32          |      2^28      |   0.201   |  38.386 ms |       0.01% |  22.160 ms |       0.01% | -16225.893 us | -42.27% |   FAST   |
|    F32     |          64          |          32          |      2^28      |     0     |  38.389 ms |       0.00% |  22.162 ms |       0.01% | -16227.566 us | -42.27% |   FAST   |
|    F32     |         128          |          8           |      2^28      |     1     |  19.256 ms |       0.01% |  11.303 ms |       0.01% |  -7952.999 us | -41.30% |   FAST   |
|    F32     |         128          |          8           |      2^28      |   0.544   |  19.256 ms |       0.01% |  11.302 ms |       0.01% |  -7953.960 us | -41.31% |   FAST   |
|    F32     |         128          |          8           |      2^28      |   0.201   |  19.256 ms |       0.01% |  11.302 ms |       0.01% |  -7953.992 us | -41.31% |   FAST   |
|    F32     |         128          |          8           |      2^28      |     0     |  19.255 ms |       0.01% |  11.302 ms |       0.01% |  -7952.808 us | -41.30% |   FAST   |
|    F32     |         128          |          16          |      2^28      |     1     |  19.264 ms |       0.01% |  11.291 ms |       0.01% |  -7973.516 us | -41.39% |   FAST   |
|    F32     |         128          |          16          |      2^28      |   0.544   |  19.265 ms |       0.01% |  11.292 ms |       0.01% |  -7973.287 us | -41.39% |   FAST   |
|    F32     |         128          |          16          |      2^28      |   0.201   |  19.264 ms |       0.01% |  11.290 ms |       0.01% |  -7974.915 us | -41.40% |   FAST   |
|    F32     |         128          |          16          |      2^28      |     0     |  19.265 ms |       0.01% |  11.292 ms |       0.01% |  -7973.760 us | -41.39% |   FAST   |
|    F32     |         128          |          32          |      2^28      |     1     |  19.293 ms |       0.01% |  11.325 ms |       0.01% |  -7968.653 us | -41.30% |   FAST   |
|    F32     |         128          |          32          |      2^28      |   0.544   |  19.292 ms |       0.01% |  11.324 ms |       0.01% |  -7968.978 us | -41.31% |   FAST   |
|    F32     |         128          |          32          |      2^28      |   0.201   |  19.294 ms |       0.01% |  11.324 ms |       0.01% |  -7969.603 us | -41.31% |   FAST   |
|    F32     |         128          |          32          |      2^28      |     0     |  19.293 ms |       0.01% |  11.324 ms |       0.01% |  -7969.286 us | -41.31% |   FAST   |
|    F32     |         256          |          8           |      2^28      |     1     |  10.311 ms |       0.01% |   5.751 ms |       0.02% |  -4559.354 us | -44.22% |   FAST   |
|    F32     |         256          |          8           |      2^28      |   0.544   |  10.309 ms |       0.01% |   5.751 ms |       0.02% |  -4558.456 us | -44.22% |   FAST   |
|    F32     |         256          |          8           |      2^28      |   0.201   |  10.310 ms |       0.01% |   5.751 ms |       0.02% |  -4558.817 us | -44.22% |   FAST   |
|    F32     |         256          |          8           |      2^28      |     0     |  10.309 ms |       0.01% |   5.751 ms |       0.02% |  -4558.186 us | -44.21% |   FAST   |
|    F32     |         256          |          16          |      2^28      |     1     |  10.321 ms |       0.01% |   5.724 ms |       0.02% |  -4596.405 us | -44.54% |   FAST   |
|    F32     |         256          |          16          |      2^28      |   0.544   |  10.322 ms |       0.01% |   5.724 ms |       0.02% |  -4597.624 us | -44.54% |   FAST   |
|    F32     |         256          |          16          |      2^28      |   0.201   |  10.321 ms |       0.01% |   5.724 ms |       0.02% |  -4597.239 us | -44.54% |   FAST   |
|    F32     |         256          |          16          |      2^28      |     0     |  10.322 ms |       0.01% |   5.724 ms |       0.02% |  -4597.923 us | -44.54% |   FAST   |
|    F32     |         256          |          32          |      2^28      |     1     |  10.350 ms |       0.01% |   5.757 ms |       0.02% |  -4592.951 us | -44.38% |   FAST   |
|    F32     |         256          |          32          |      2^28      |   0.544   |  10.349 ms |       0.01% |   5.757 ms |       0.02% |  -4591.825 us | -44.37% |   FAST   |
|    F32     |         256          |          32          |      2^28      |   0.201   |  10.349 ms |       0.01% |   5.756 ms |       0.02% |  -4592.447 us | -44.38% |   FAST   |
|    F32     |         256          |          32          |      2^28      |     0     |  10.349 ms |       0.01% |   5.757 ms |       0.02% |  -4592.075 us | -44.37% |   FAST   |
|    F32     |         256          |         128          |      2^28      |     1     |  10.415 ms |       0.01% |   5.790 ms |       0.02% |  -4625.012 us | -44.41% |   FAST   |
|    F32     |         256          |         128          |      2^28      |   0.544   |  10.416 ms |       0.01% |   5.789 ms |       0.02% |  -4626.536 us | -44.42% |   FAST   |
|    F32     |         256          |         128          |      2^28      |   0.201   |  10.415 ms |       0.01% |   5.789 ms |       0.02% |  -4626.136 us | -44.42% |   FAST   |
|    F32     |         256          |         128          |      2^28      |     0     |  10.416 ms |       0.01% |   5.790 ms |       0.02% |  -4626.015 us | -44.41% |   FAST   |
|    F32     |         512          |          8           |      2^28      |     1     |   5.376 ms |       0.02% |   3.068 ms |       0.04% |  -2307.423 us | -42.92% |   FAST   |
|    F32     |         512          |          8           |      2^28      |   0.544   |   5.376 ms |       0.02% |   3.068 ms |       0.04% |  -2307.279 us | -42.92% |   FAST   |
|    F32     |         512          |          8           |      2^28      |   0.201   |   5.376 ms |       0.02% |   3.068 ms |       0.04% |  -2307.956 us | -42.93% |   FAST   |
|    F32     |         512          |          8           |      2^28      |     0     |   5.376 ms |       0.02% |   3.068 ms |       0.04% |  -2307.335 us | -42.92% |   FAST   |
|    F32     |         512          |          16          |      2^28      |     1     |   5.379 ms |       0.02% |   3.092 ms |       0.04% |  -2286.462 us | -42.51% |   FAST   |
|    F32     |         512          |          16          |      2^28      |   0.544   |   5.379 ms |       0.02% |   3.092 ms |       0.03% |  -2286.520 us | -42.51% |   FAST   |
|    F32     |         512          |          16          |      2^28      |   0.201   |   5.379 ms |       0.02% |   3.092 ms |       0.03% |  -2286.960 us | -42.51% |   FAST   |
|    F32     |         512          |          16          |      2^28      |     0     |   5.379 ms |       0.02% |   3.092 ms |       0.03% |  -2286.624 us | -42.51% |   FAST   |
|    F32     |         512          |          32          |      2^28      |     1     |   5.387 ms |       0.02% |   3.123 ms |       0.03% |  -2263.563 us | -42.02% |   FAST   |
|    F32     |         512          |          32          |      2^28      |   0.544   |   5.388 ms |       0.02% |   3.123 ms |       0.04% |  -2264.293 us | -42.03% |   FAST   |
|    F32     |         512          |          32          |      2^28      |   0.201   |   5.387 ms |       0.02% |   3.123 ms |       0.04% |  -2263.766 us | -42.02% |   FAST   |
|    F32     |         512          |          32          |      2^28      |     0     |   5.388 ms |       0.02% |   3.123 ms |       0.04% |  -2264.094 us | -42.02% |   FAST   |
|    F32     |         512          |         128          |      2^28      |     1     |   5.420 ms |       0.02% |   3.110 ms |       0.04% |  -2309.788 us | -42.61% |   FAST   |
|    F32     |         512          |         128          |      2^28      |   0.544   |   5.420 ms |       0.02% |   3.110 ms |       0.03% |  -2309.677 us | -42.61% |   FAST   |
|    F32     |         512          |         128          |      2^28      |   0.201   |   5.420 ms |       0.02% |   3.111 ms |       0.03% |  -2309.907 us | -42.61% |   FAST   |
|    F32     |         512          |         128          |      2^28      |     0     |   5.420 ms |       0.02% |   3.110 ms |       0.04% |  -2309.517 us | -42.61% |   FAST   |
|    F32     |         1024         |          8           |      2^28      |     1     |   2.905 ms |       0.03% |   1.890 ms |       0.07% |  -1014.991 us | -34.94% |   FAST   |
|    F32     |         1024         |          8           |      2^28      |   0.544   |   2.905 ms |       0.04% |   1.890 ms |       0.06% |  -1015.096 us | -34.94% |   FAST   |
|    F32     |         1024         |          8           |      2^28      |   0.201   |   2.905 ms |       0.03% |   1.890 ms |       0.07% |  -1015.398 us | -34.95% |   FAST   |
|    F32     |         1024         |          8           |      2^28      |     0     |   2.905 ms |       0.03% |   1.890 ms |       0.06% |  -1014.862 us | -34.93% |   FAST   |
|    F32     |         1024         |          16          |      2^28      |     1     |   2.907 ms |       0.04% |   1.908 ms |       0.06% |   -998.828 us | -34.36% |   FAST   |
|    F32     |         1024         |          16          |      2^28      |   0.544   |   2.907 ms |       0.04% |   1.908 ms |       0.06% |   -998.986 us | -34.36% |   FAST   |
|    F32     |         1024         |          16          |      2^28      |   0.201   |   2.907 ms |       0.04% |   1.908 ms |       0.06% |   -999.171 us | -34.37% |   FAST   |
|    F32     |         1024         |          16          |      2^28      |     0     |   2.907 ms |       0.04% |   1.908 ms |       0.06% |   -998.759 us | -34.36% |   FAST   |
|    F32     |         1024         |          32          |      2^28      |     1     |   2.910 ms |       0.04% |   1.905 ms |       0.06% |  -1005.538 us | -34.55% |   FAST   |
|    F32     |         1024         |          32          |      2^28      |   0.544   |   2.910 ms |       0.03% |   1.905 ms |       0.06% |  -1005.656 us | -34.55% |   FAST   |
|    F32     |         1024         |          32          |      2^28      |   0.201   |   2.910 ms |       0.04% |   1.905 ms |       0.06% |  -1005.695 us | -34.55% |   FAST   |
|    F32     |         1024         |          32          |      2^28      |     0     |   2.910 ms |       0.03% |   1.905 ms |       0.06% |  -1005.581 us | -34.55% |   FAST   |
|    F32     |         1024         |         128          |      2^28      |     1     |   2.930 ms |       0.03% |   1.895 ms |       0.05% |  -1035.599 us | -35.34% |   FAST   |
|    F32     |         1024         |         128          |      2^28      |   0.544   |   2.930 ms |       0.04% |   1.895 ms |       0.05% |  -1035.448 us | -35.33% |   FAST   |
|    F32     |         1024         |         128          |      2^28      |   0.201   |   2.931 ms |       0.03% |   1.895 ms |       0.05% |  -1035.544 us | -35.34% |   FAST   |
|    F32     |         1024         |         128          |      2^28      |     0     |   2.930 ms |       0.03% |   1.895 ms |       0.06% |  -1035.171 us | -35.33% |   FAST   |
|    F32     |         1024         |         512          |      2^28      |     1     |   2.978 ms |       0.03% |   1.896 ms |       0.06% |  -1082.667 us | -36.35% |   FAST   |
|    F32     |         1024         |         512          |      2^28      |   0.544   |   2.979 ms |       0.04% |   1.895 ms |       0.06% |  -1083.437 us | -36.37% |   FAST   |
|    F32     |         1024         |         512          |      2^28      |   0.201   |   2.978 ms |       0.03% |   1.895 ms |       0.06% |  -1083.000 us | -36.37% |   FAST   |
|    F32     |         1024         |         512          |      2^28      |     0     |   2.979 ms |       0.04% |   1.895 ms |       0.05% |  -1083.066 us | -36.36% |   FAST   |
```

### DIFF
--- a/cub/benchmarks/bench/adjacent_difference/subtract_left.cu
+++ b/cub/benchmarks/bench/adjacent_difference/subtract_left.cu
@@ -9,19 +9,17 @@
 // %RANGE% TUNE_THREADS_PER_BLOCK tpb 128:1024:32
 
 #if !TUNE_BASE
-struct policy_hub_t
+struct policy_selector_t
 {
-  struct Policy500 : cub::ChainedPolicy<500, Policy500, Policy500>
+  _CCCL_API constexpr auto operator()(::cuda::arch_id) const
+    -> cub::detail::adjacent_difference::adjacent_difference_policy
   {
-    using AdjacentDifferencePolicy =
-      cub::AgentAdjacentDifferencePolicy<TUNE_THREADS_PER_BLOCK,
-                                         TUNE_ITEMS_PER_THREAD,
-                                         cub::BLOCK_LOAD_WARP_TRANSPOSE,
-                                         cub::LOAD_CA,
-                                         cub::BLOCK_STORE_WARP_TRANSPOSE>;
-  };
-
-  using MaxPolicy = Policy500;
+    return {TUNE_THREADS_PER_BLOCK,
+            TUNE_ITEMS_PER_THREAD,
+            cub::BLOCK_LOAD_WARP_TRANSPOSE,
+            cub::LOAD_CA,
+            cub::BLOCK_STORE_WARP_TRANSPOSE};
+  }
 };
 #endif // !TUNE_BASE
 
@@ -32,25 +30,6 @@ void left(nvbench::state& state, nvbench::type_list<T, OffsetT>)
   using output_it_t     = T*;
   using difference_op_t = ::cuda::std::minus<>;
   using offset_t        = cub::detail::choose_offset_t<OffsetT>;
-
-#if !TUNE_BASE
-  using dispatch_t = cub::DispatchAdjacentDifference<
-    input_it_t,
-    output_it_t,
-    difference_op_t,
-    offset_t,
-    cub::MayAlias::No,
-    cub::ReadOption::Left,
-    policy_hub_t>;
-#else
-  using dispatch_t = cub::DispatchAdjacentDifference<
-    input_it_t,
-    output_it_t,
-    difference_op_t,
-    offset_t,
-    cub::MayAlias::No,
-    cub::ReadOption::Left>;
-#endif // TUNE_BASE
 
   const auto elements         = static_cast<std::size_t>(state.get_int64("Elements{io}"));
   thrust::device_vector<T> in = generate(elements);
@@ -64,20 +43,40 @@ void left(nvbench::state& state, nvbench::type_list<T, OffsetT>)
   state.add_global_memory_writes<T>(elements);
 
   std::size_t temp_storage_bytes{};
-  dispatch_t::Dispatch(nullptr, temp_storage_bytes, d_in, d_out, static_cast<offset_t>(elements), difference_op_t{}, 0);
 
-  thrust::device_vector<std::uint8_t> temp_storage(temp_storage_bytes);
-  std::uint8_t* d_temp_storage = thrust::raw_pointer_cast(temp_storage.data());
-
-  state.exec(nvbench::exec_tag::gpu | nvbench::exec_tag::no_batch, [&](nvbench::launch& launch) {
-    dispatch_t::Dispatch(
-      d_temp_storage,
+  cub::detail::adjacent_difference::
+    dispatch<cub::MayAlias::No, cub::ReadOption::Left, input_it_t, output_it_t, offset_t, difference_op_t>(
+      nullptr,
       temp_storage_bytes,
       d_in,
       d_out,
       static_cast<offset_t>(elements),
       difference_op_t{},
-      launch.get_stream());
+      0
+#if !TUNE_BASE
+      ,
+      policy_selector_t{}
+#endif // TUNE_BASE
+    );
+
+  thrust::device_vector<std::uint8_t> temp_storage(temp_storage_bytes, thrust::no_init);
+  std::uint8_t* d_temp_storage = thrust::raw_pointer_cast(temp_storage.data());
+
+  state.exec(nvbench::exec_tag::gpu | nvbench::exec_tag::no_batch, [&](nvbench::launch& launch) {
+    cub::detail::adjacent_difference::
+      dispatch<cub::MayAlias::No, cub::ReadOption::Left, input_it_t, output_it_t, offset_t, difference_op_t>(
+        d_temp_storage,
+        temp_storage_bytes,
+        d_in,
+        d_out,
+        static_cast<offset_t>(elements),
+        difference_op_t{},
+        launch.get_stream()
+#if !TUNE_BASE
+          ,
+        policy_selector_t{}
+#endif // TUNE_BASE
+      );
   });
 }
 
--- a/cub/cub/device/device_adjacent_difference.cuh
+++ b/cub/cub/device/device_adjacent_difference.cuh
@@ -84,32 +84,6 @@ CUB_NAMESPACE_BEGIN
 //! @endrst
 struct DeviceAdjacentDifference
 {
-private:
-  template <MayAlias AliasOpt,
-            ReadOption ReadOpt,
-            typename NumItemsT,
-            typename InputIteratorT,
-            typename OutputIteratorT,
-            typename DifferenceOpT>
-  static CUB_RUNTIME_FUNCTION cudaError_t AdjacentDifference(
-    void* d_temp_storage,
-    size_t& temp_storage_bytes,
-    InputIteratorT d_input,
-    OutputIteratorT d_output,
-    NumItemsT num_items,
-    DifferenceOpT difference_op,
-    cudaStream_t stream)
-  {
-    using OffsetT = detail::choose_offset_t<NumItemsT>;
-
-    using DispatchT =
-      DispatchAdjacentDifference<InputIteratorT, OutputIteratorT, DifferenceOpT, OffsetT, AliasOpt, ReadOpt>;
-
-    return DispatchT::Dispatch(
-      d_temp_storage, temp_storage_bytes, d_input, d_output, static_cast<OffsetT>(num_items), difference_op, stream);
-  }
-
-public:
   //! @rst
   //! Subtracts the left element of each adjacent pair of elements residing within device-accessible memory
   //!
@@ -234,9 +208,9 @@ public:
     cudaStream_t stream         = 0)
   {
     _CCCL_NVTX_RANGE_SCOPE_IF(d_temp_storage, "cub::DeviceAdjacentDifference::SubtractLeftCopy");
-
-    return AdjacentDifference<MayAlias::No, ReadOption::Left>(
-      d_temp_storage, temp_storage_bytes, d_input, d_output, num_items, difference_op, stream);
+    using OffsetT = detail::choose_offset_t<NumItemsT>;
+    return detail::adjacent_difference::dispatch<MayAlias::No, ReadOption::Left>(
+      d_temp_storage, temp_storage_bytes, d_input, d_output, static_cast<OffsetT>(num_items), difference_op, stream);
   }
 
   //! @rst
@@ -345,9 +319,9 @@ public:
     cudaStream_t stream         = 0)
   {
     _CCCL_NVTX_RANGE_SCOPE_IF(d_temp_storage, "cub::DeviceAdjacentDifference::SubtractLeft");
-
-    return AdjacentDifference<MayAlias::Yes, ReadOption::Left>(
-      d_temp_storage, temp_storage_bytes, d_input, d_input, num_items, difference_op, stream);
+    using OffsetT = detail::choose_offset_t<NumItemsT>;
+    return detail::adjacent_difference::dispatch<MayAlias::Yes, ReadOption::Left>(
+      d_temp_storage, temp_storage_bytes, d_input, d_input, static_cast<OffsetT>(num_items), difference_op, stream);
   }
 
   //! @rst
@@ -475,9 +449,9 @@ public:
     cudaStream_t stream         = 0)
   {
     _CCCL_NVTX_RANGE_SCOPE_IF(d_temp_storage, "cub::DeviceAdjacentDifference::SubtractRightCopy");
-
-    return AdjacentDifference<MayAlias::No, ReadOption::Right>(
-      d_temp_storage, temp_storage_bytes, d_input, d_output, num_items, difference_op, stream);
+    using OffsetT = detail::choose_offset_t<NumItemsT>;
+    return detail::adjacent_difference::dispatch<MayAlias::No, ReadOption::Right>(
+      d_temp_storage, temp_storage_bytes, d_input, d_output, static_cast<OffsetT>(num_items), difference_op, stream);
   }
 
   //! @rst
@@ -575,9 +549,9 @@ public:
     cudaStream_t stream         = 0)
   {
     _CCCL_NVTX_RANGE_SCOPE_IF(d_temp_storage, "cub::DeviceAdjacentDifference::SubtractRight");
-
-    return AdjacentDifference<MayAlias::Yes, ReadOption::Right>(
-      d_temp_storage, temp_storage_bytes, d_input, d_input, num_items, difference_op, stream);
+    using OffsetT = detail::choose_offset_t<NumItemsT>;
+    return detail::adjacent_difference::dispatch<MayAlias::Yes, ReadOption::Right>(
+      d_temp_storage, temp_storage_bytes, d_input, d_input, static_cast<OffsetT>(num_items), difference_op, stream);
   }
 };
 
--- a/cub/cub/device/dispatch/dispatch_adjacent_difference.cuh
+++ b/cub/cub/device/dispatch/dispatch_adjacent_difference.cuh
@@ -313,12 +313,12 @@ struct DispatchAdjacentDifference
 
 namespace detail::adjacent_difference
 {
-template <typename InputIteratorT,
-          typename OutputIteratorT,
-          typename DifferenceOpT,
-          typename OffsetT,
-          MayAlias AliasOpt,
+template <MayAlias AliasOpt,
           ReadOption ReadOpt,
+          typename InputIteratorT,
+          typename OutputIteratorT,
+          typename OffsetT,
+          typename DifferenceOpT,
           typename PolicySelector        = policy_selector_from_types<InputIteratorT, AliasOpt == MayAlias::Yes>,
           typename KernelLauncherFactory = CUB_DETAIL_DEFAULT_KERNEL_LAUNCHER_FACTORY>
 #if _CCCL_HAS_CONCEPTS()
--- a/cub/cub/util_device.cuh
+++ b/cub/cub/util_device.cuh
@@ -82,7 +82,7 @@ CUB_RUNTIME_FUNCTION inline int CurrentDevice()
 
 //! @brief RAII helper which saves the current device and switches to the specified device on construction and switches
 //! to the saved device on destruction.
-class SwitchDevice
+class [[maybe_unused]] SwitchDevice
 {
   int target_device_;
   int original_device_;
@@ -278,39 +278,16 @@ CUB_RUNTIME_FUNCTION cudaError_t PtxVersionUncached(int& ptx_version)
 {
   // Instantiate `EmptyKernel<void>` in both host and device code to ensure
   // it can be called.
-  using EmptyKernelPtr                         = void (*)();
-  [[maybe_unused]] EmptyKernelPtr empty_kernel = detail::EmptyKernel<T>;
-
-  // Define a temporary macro that expands to the current target ptx version
-  // in device code.
-  // <nv/target> may provide an abstraction for this eventually. For now,
-  // we have to keep this usage of __CUDA_ARCH__.
-#  if _CCCL_CUDA_COMPILER(NVHPC)
-#    define CUB_TEMP_GET_PTX __builtin_current_device_sm()
-#  else // ^^^ _CCCL_CUDA_COMPILER(NVHPC) ^^^ / vvv !_CCCL_CUDA_COMPILER(NVHPC) vvv
-#    define CUB_TEMP_GET_PTX _CCCL_PTX_ARCH()
-#  endif // ^^^ !_CCCL_CUDA_COMPILER(NVHPC) ^^^
+  [[maybe_unused]] const auto empty_kernel = detail::EmptyKernel<T>;
 
   cudaError_t result = cudaSuccess;
-  NV_IF_TARGET(
-    NV_IS_HOST,
-    (cudaFuncAttributes empty_kernel_attrs;
-
-     result = CubDebug(cudaFuncGetAttributes(&empty_kernel_attrs, reinterpret_cast<void*>(empty_kernel)));
-
-     ptx_version = empty_kernel_attrs.ptxVersion * 10;),
-    // NV_IS_DEVICE
-    (
-      // This is necessary to ensure instantiation of EmptyKernel in device
-      // code. The `reinterpret_cast` is necessary to suppress a
-      // set-but-unused warnings. This is a meme now:
-      // https://twitter.com/blelbach/status/1222391615576100864
-      (void) reinterpret_cast<EmptyKernelPtr>(empty_kernel);
-
-      ptx_version = CUB_TEMP_GET_PTX;));
-
-#  undef CUB_TEMP_GET_PTX
-
+  NV_IF_ELSE_TARGET(NV_IS_HOST,
+                    ({
+                      cudaFuncAttributes empty_kernel_attrs;
+                      result      = CubDebug(cudaFuncGetAttributes(&empty_kernel_attrs, (const void*) empty_kernel));
+                      ptx_version = empty_kernel_attrs.ptxVersion * 10;
+                    }),
+                    ({ ptx_version = ::cuda::device::current_compute_capability().get() * 10; }));
   return result;
 }
 
@@ -320,7 +297,7 @@ CUB_RUNTIME_FUNCTION cudaError_t PtxVersionUncached(int& ptx_version)
 template <class T = void>
 _CCCL_HOST cudaError_t PtxVersionUncached(int& ptx_version, int device)
 {
-  [[maybe_unused]] SwitchDevice sd(device);
+  SwitchDevice sd(device);
   return PtxVersionUncached<T>(ptx_version);
 }
 
--- a/cudax/include/cuda/experimental/__stf/internal/async_resources_handle.cuh
+++ b/cudax/include/cuda/experimental/__stf/internal/async_resources_handle.cuh
@@ -37,6 +37,8 @@
 #include <mutex>
 #include <unordered_map>
 
+#include <cuda.h>
+
 namespace cuda::experimental::stf
 {
 class green_context_helper;
@@ -44,26 +46,50 @@ class green_context_helper;
 // Needed to set/get affinity
 class exec_place;
 
+/** Sentinel for "no stream" / empty slot. Distinct from any value returned by cuStreamGetId. */
+inline constexpr unsigned long long k_no_stream_id = static_cast<unsigned long long>(-1);
+
 /**
- * @brief A class to store a CUDA stream along with a few information to avoid CUDA queries
+ * @brief Returns the unique stream ID from the CUDA driver (cuStreamGetId).
+ * @param stream A valid CUDA stream, or nullptr.
+ * @return The stream's unique ID, or k_no_stream_id if stream is nullptr.
+ */
+inline unsigned long long get_stream_id(cudaStream_t stream)
+{
+  unsigned long long id = 0;
+  cuda_safe_call(cuStreamGetId(reinterpret_cast<CUstream>(stream), &id));
+  _CCCL_ASSERT(id != k_no_stream_id, "Internal error: cuStreamGetId returned k_no_stream_id");
+  return id;
+}
+
+/**
+ * @brief A class to store a CUDA stream along with metadata
  *
  * It contains
  *  - the stream itself,
- *  - a unique id (proper to CUDASTF, and only valid for streams in our pool, or equal to -1),
- *  - the pool associated to the unique ID, when valid
- *  - the device index in which the stream is
+ *  - the stream's unique ID from the CUDA driver (cuStreamGetId), or k_no_stream_id if no stream,
+ *  - the device index in which the stream resides
  */
 struct decorated_stream
 {
-  decorated_stream(cudaStream_t stream = nullptr, ::std::ptrdiff_t id = -1, int dev_id = -1)
+  decorated_stream() = default;
+
+  decorated_stream(cudaStream_t stream, unsigned long long id, int dev_id = -1)
       : stream(stream)
       , id(id)
       , dev_id(dev_id)
   {}
 
+  /** Construct from stream only; id is from cuStreamGetId, dev_id is -1 (filled lazily when needed). */
+  explicit decorated_stream(cudaStream_t stream)
+      : stream(stream)
+      , id(get_stream_id(stream))
+      , dev_id(-1)
+  {}
+
   cudaStream_t stream = nullptr;
-  // Unique ID (-1 if this is not part of our pool)
-  ::std::ptrdiff_t id = -1;
+  // Unique ID from cuStreamGetId (k_no_stream_id if no stream)
+  unsigned long long id = k_no_stream_id;
   // Device in which this stream resides
   int dev_id = -1;
 };
@@ -90,10 +116,10 @@ struct stream_pool
    * @brief stream_pool constructor taking a number of slots.
    *
    * Streams are created lazily only via next(place), which activates the place and calls place.create_stream().
-   * Slot dev_id is set from the created stream; the pool does not store a device id.
+   * Slot dev_id and id are set when the stream is created in next().
    */
   explicit stream_pool(size_t n)
-      : payload(n, decorated_stream(nullptr, -1, -1))
+      : payload(n, decorated_stream(nullptr, k_no_stream_id, -1))
   {}
 
   stream_pool(stream_pool&& rhs)
@@ -157,36 +183,6 @@ public:
 
 private:
   /**
-   * @brief A helper class to maintain a set of available IDs, and attributes IDs
-   */
-  class id_pool
-  {
-  public:
-    ~id_pool()
-    {
-      assert(released.load() == current.load());
-    }
-
-    ::std::ptrdiff_t get_unique_id(size_t cnt = 1)
-    {
-      // Use fetch_add to atomically increment current and return the previous value
-      return current.fetch_add(cnt);
-    }
-
-    void release_unique_id(::std::ptrdiff_t /* id */, size_t cnt = 1)
-    {
-      // Use fetch_add to atomically increment released
-      released.fetch_add(cnt);
-    }
-
-  private:
-    // next available ID
-    ::std::atomic<::std::ptrdiff_t> current{0};
-    // Number of IDs released, for bookkeeping
-    ::std::atomic<::std::ptrdiff_t> released{0};
-  };
-
-  /**
    * @brief This class implements a matrix to keep track of the previous
    * synchronization that occurred between each pair of streams in our pools.
    *
@@ -195,7 +191,7 @@ private:
    * ID) is implied by the previous synchronization, so it can be skipped thanks
    * to stream-ordering of operations.
    *
-   * This is implemented as a hash table where keys are pairs of IDs.
+   * Keys are pairs of stream IDs from cuStreamGetId.
    */
   class last_event_per_stream
   {
@@ -204,10 +200,10 @@ private:
     // located on stream "from" to stream "dst" (stream dst waits for the
     // event)
     // Returned value : boolean indicating if we can skip the synchronization
-    bool validate_sync_and_update(::std::ptrdiff_t dst, ::std::ptrdiff_t src, int event_id)
+    bool validate_sync_and_update(unsigned long long dst, unsigned long long src, int event_id)
     {
-      // If either of the streams is not from the pool, do not skip
-      if (dst == -1 || src == -1)
+      // If either of the streams has no valid id, do not skip
+      if (dst == k_no_stream_id || src == k_no_stream_id)
       {
         return false;
       }
@@ -232,10 +228,10 @@ private:
     }
 
   private:
-    // For each pair of unique IDs, we keep the last event id
-    ::std::unordered_map<::std::pair<::std::ptrdiff_t, ::std::ptrdiff_t>,
+    // For each pair of stream IDs (from cuStreamGetId), we keep the last event id
+    ::std::unordered_map<::std::pair<unsigned long long, unsigned long long>,
                          int,
-                         cuda::experimental::stf::hash<::std::pair<::std::ptrdiff_t, ::std::ptrdiff_t>>>
+                         cuda::experimental::stf::hash<::std::pair<unsigned long long, unsigned long long>>>
       interactions;
 
     ::std::mutex mtx;
@@ -295,7 +291,7 @@ private:
       for (auto i : each(n))
       {
         ::std::ignore = i;
-        new_payload.emplace_back(nullptr, ids.get_unique_id(), dev_id);
+        new_payload.emplace_back(nullptr, k_no_stream_id, dev_id);
       }
 
       ::std::lock_guard<::std::mutex> locker(p.mtx);
@@ -312,7 +308,6 @@ private:
       // Clean up outside the critical section
       for (auto& e : goner)
       {
-        ids.release_unique_id(e.id);
         if (e.stream)
         {
           cuda_safe_call(cudaStreamDestroy(e.stream));
@@ -321,9 +316,6 @@ private:
     }
 
   public:
-    // These are constructed and destroyed in reversed order
-    id_pool ids;
-
     // This memorize what was the last event used to synchronize a pair of streams
     last_event_per_stream cached_syncs;
 
@@ -359,19 +351,7 @@ public:
     return pimpl->get_device_stream_pool(dev_id, for_computation);
   }
 
-  ::std::ptrdiff_t get_unique_id(size_t cnt = 1)
-  {
-    assert(pimpl);
-    return pimpl->ids.get_unique_id(cnt);
-  }
-
-  void release_unique_id(::std::ptrdiff_t id, size_t cnt = 1)
-  {
-    assert(pimpl);
-    return pimpl->ids.release_unique_id(id, cnt);
-  }
-
-  bool validate_sync_and_update(::std::ptrdiff_t dst, ::std::ptrdiff_t src, int event_id)
+  bool validate_sync_and_update(unsigned long long dst, unsigned long long src, int event_id)
   {
     assert(pimpl);
     return pimpl->cached_syncs.validate_sync_and_update(dst, src, event_id);
@@ -445,54 +425,6 @@ public:
     return pimpl->affinity.top();
   }
 };
-
-//! @brief Registers a user-provided CUDA stream with asynchronous resources
-//!
-//! @details This optimization records a CUDA stream in the provided asynchronous resources handle,
-//! creating a decorated_stream object that encapsulates:
-//! - The original stream handle
-//! - A unique identifier for stream tracking
-//! - The associated device ID
-//!
-//! @param[in,out] async_resources Handle to asynchronous resources manager
-//! @param[in] user_stream Raw CUDA stream to register. Must be a valid stream.
-//!
-//! @return decorated_stream Object containing:
-//!         - Original stream handle
-//!         - Unique ID from async_resources
-//!         - Device ID associated with the stream
-//!
-//! @pre `user_stream` must be a valid CUDA stream created with `cudaStreamCreate` or equivalent
-//! @note This registration is an optimization to avoid repeated stream metadata lookups
-//!       in performance-critical code paths
-inline decorated_stream register_stream(async_resources_handle& async_resources, cudaStream_t user_stream)
-{
-  // Get a unique ID
-  const auto id    = async_resources.get_unique_id();
-  const int dev_id = get_device_from_stream(user_stream);
-
-  return decorated_stream(user_stream, id, dev_id);
-}
-
-//! @brief Unregisters a decorated CUDA stream from asynchronous resources
-//!
-//! @details Performs cleanup operations to release resources associated with a previously
-//! registered stream. This includes:
-//! - Releasing the unique ID back to the resource manager
-//! - Invalidating the decorated stream's internal ID
-//!
-//! @param[in,out] async_resources Handle to asynchronous resources manager
-//! @param[in,out] dstream Decorated stream to unregister. Its `id` will be set to -1.
-//!
-//! @pre `dstream.id` must be valid (≥ 0) before calling this function
-//! @post `dstream.id == -1` and associated resources are released
-//! @note Should be paired with register_stream() for proper resource management
-inline void unregister_stream(async_resources_handle& async_resources, decorated_stream& dstream)
-{
-  async_resources.release_unique_id(dstream.id);
-  // reset the decorated stream
-  dstream.id = -1;
-}
 
 #ifdef UNITTESTED_FILE
 /*
--- a/cudax/include/cuda/experimental/__stf/places/exec/cuda_stream.cuh
+++ b/cudax/include/cuda/experimental/__stf/places/exec/cuda_stream.cuh
@@ -114,7 +114,7 @@ public:
 inline exec_place_cuda_stream exec_place::cuda_stream(cudaStream_t stream)
 {
   int devid = get_device_from_stream(stream);
-  return exec_place_cuda_stream(decorated_stream(stream, -1, devid));
+  return exec_place_cuda_stream(decorated_stream(stream, get_stream_id(stream), devid));
 }
 
 inline exec_place_cuda_stream exec_place::cuda_stream(const decorated_stream& dstream)
--- a/cudax/include/cuda/experimental/__stf/places/places.cuh
+++ b/cudax/include/cuda/experimental/__stf/places/places.cuh
@@ -1218,6 +1218,7 @@ inline decorated_stream stream_pool::next(const exec_place& place)
   {
     exec_place_guard guard(place);
     result.stream = place.create_stream();
+    result.id     = get_stream_id(result.stream);
     result.dev_id = get_device_from_stream(result.stream);
   }
 
--- a/cudax/include/cuda/experimental/__stf/stream/internal/event_types.cuh
+++ b/cudax/include/cuda/experimental/__stf/stream/internal/event_types.cuh
@@ -255,7 +255,7 @@ public:
     return dstream;
   }
 
-  ::std::ptrdiff_t get_stream_id() const
+  unsigned long long get_stream_id() const
   {
     return dstream.id;
   }
@@ -397,7 +397,7 @@ private:
     for (const auto& e : prereq_in)
     {
       cudaStream_t stream;
-      ::std::ptrdiff_t stream_id = -1;
+      unsigned long long stream_id = 0;
       auto se   = reserved::handle<stream_and_event, reserved::handle_flags::non_null>(e, reserved::use_static_cast);
       stream    = se->get_stream();
       stream_id = se->get_stream_id();
@@ -415,7 +415,7 @@ private:
       if (stream_dev == devid)
       {
         //    fprintf(stderr, "Found matching device %d with stream %p\n", devid, stream);
-        return decorated_stream(stream, stream_id, devid);
+        return decorated_stream(stream, stream_id, static_cast<int>(stream_dev));
       }
     }
 
--- a/cudax/include/cuda/experimental/__stf/stream/stream_task.cuh
+++ b/cudax/include/cuda/experimental/__stf/stream/stream_task.cuh
@@ -156,7 +156,7 @@ public:
               auto se                    = reserved::handle<stream_and_event>(e, reserved::use_static_cast);
               decorated_stream candidate = se->get_decorated_stream();
 
-              if (candidate.id != -1)
+              if (candidate.id != k_no_stream_id)
               {
                 for (const decorated_stream& pool_s : pool)
                 {
--- a/cudax/test/green_context/green_ctx_smoke.cu
+++ b/cudax/test/green_context/green_ctx_smoke.cu
@@ -80,11 +80,20 @@ C2H_TEST("Green context", "[green_context]")
       auto id2 = ctx2.id();
 
       // Test that different contexts have different IDs
+#    if _CCCL_COMPILER(NVHPC, <, 25, 11)
+      CUDAX_REQUIRE(cuda::std::to_underlying(id1) != cuda::std::to_underlying(id2));
+#    else // ^^^ _CCCL_COMPILER(NVHPC, <, 25, 11) ^^^ / vvv !_CCCL_COMPILER(NVHPC, <, 25, 11) vvv
       CUDAX_REQUIRE(id1 != id2);
+#    endif // ^^^ !_CCCL_COMPILER(NVHPC, <, 25, 11) ^^^
 
       // Test that the same context returns the same ID when called multiple times
+#    if _CCCL_COMPILER(NVHPC, <, 25, 11)
+      CUDAX_REQUIRE(cuda::std::to_underlying(ctx1.id()) == cuda::std::to_underlying(id1));
+      CUDAX_REQUIRE(cuda::std::to_underlying(ctx2.id()) == cuda::std::to_underlying(id2));
+#    else // ^^^ _CCCL_COMPILER(NVHPC, <, 25, 11) ^^^ / vvv !_CCCL_COMPILER(NVHPC, <, 25, 11) vvv
       CUDAX_REQUIRE(ctx1.id() == id1);
       CUDAX_REQUIRE(ctx2.id() == id2);
+#    endif // ^^^ !_CCCL_COMPILER(NVHPC, <, 25, 11) ^^^
     }
   }
 #  endif // _CCCL_CTK_AT_LEAST(13, 0)
--- a/cudax/test/stf/places/cuda_stream_place.cu
+++ b/cudax/test/stf/places/cuda_stream_place.cu
@@ -53,29 +53,15 @@ int main()
   auto lX = ctx.logical_data(X);
   auto lY = ctx.logical_data(Y);
 
-  /* Compute Y = Y + alpha X */
+  /* Compute Y = Y + alpha X on the user stream */
   auto where = exec_place::cuda_stream(stream);
 
-  for (size_t iter = 0; iter < 10; iter++)
+  for (size_t iter = 0; iter < 20; iter++)
   {
     ctx.parallel_for(where, lX.shape(), lX.read(), lY.rw())->*[alpha] __device__(size_t i, auto x, auto y) {
       y(i) += alpha * x(i);
     };
   }
-
-  /* Associate the CUDA stream with a unique internal ID to speed up synchronizations */
-  auto rstream = register_stream(ctx.async_resources(), stream);
-  auto where2  = exec_place::cuda_stream(rstream);
-
-  for (size_t iter = 0; iter < 10; iter++)
-  {
-    ctx.parallel_for(where2, lX.shape(), lX.read(), lY.rw())->*[alpha] __device__(size_t i, auto x, auto y) {
-      y(i) += alpha * x(i);
-    };
-  }
-
-  // Remove the association
-  unregister_stream(ctx.async_resources(), rstream);
 
   ctx.finalize();
 
--- a/cudax/test/stream/stream_smoke.cu
+++ b/cudax/test/stream/stream_smoke.cu
@@ -151,20 +151,34 @@ C2H_CCCLRT_TEST("Stream ID", "[stream]")
   auto id2 = stream2.id();
 
   // Test that different streams have different IDs
+#if _CCCL_COMPILER(NVHPC, <, 25, 11)
+  CUDAX_REQUIRE(cuda::std::to_underlying(id1) != cuda::std::to_underlying(id2));
+#else // ^^^ _CCCL_COMPILER(NVHPC, <, 25, 11) ^^^ / vvv !_CCCL_COMPILER(NVHPC, <, 25, 11) vvv
   CUDAX_REQUIRE(id1 != id2);
+#endif // ^^^ !_CCCL_COMPILER(NVHPC, <, 25, 11) ^^^
 
   // Test that the same stream returns the same ID when called multiple times
+#if _CCCL_COMPILER(NVHPC, <, 25, 11)
+  CUDAX_REQUIRE(cuda::std::to_underlying(stream1.id()) == cuda::std::to_underlying(id1));
+  CUDAX_REQUIRE(cuda::std::to_underlying(stream2.id()) == cuda::std::to_underlying(id2));
+#else // ^^^ _CCCL_COMPILER(NVHPC, <, 25, 11) ^^^ / vvv !_CCCL_COMPILER(NVHPC, <, 25, 11) vvv
   CUDAX_REQUIRE(stream1.id() == id1);
   CUDAX_REQUIRE(stream2.id() == id2);
+#endif // ^^^ !_CCCL_COMPILER(NVHPC, <, 25, 11) ^^^
 
   {
     // Test that stream_ref also supports id()
     // NULL stream needs a device to be set
-    cudax::__ensure_current_device guard(cuda::device_ref{0});
-    cudax::stream_ref ref1(cudaStream_t{});
-    cudax::stream_ref ref2(stream1);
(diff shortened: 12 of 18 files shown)


cub/cub/block/block_topk.cuh
```diff
@@ -9,6 +9,8 @@
 
 #include <cub/config.cuh>
 
+#include <cstdint>
```
bernhardmgruber:
```suggestion
#include <cuda/std/cstdint>
```

miscco: Critical: Please move to after system header guard
```suggestion
#include <cuda/std/cstdint>
```

cub/cub/block/specializations/block_topk_air.cuh
```diff
@@ -0,0 +1,502 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#pragma once
+
+#include <cub/config.cuh>
+
+#include <cstdint>
+
```
miscco: Ditto needs to be `<cuda/std/cstdint>` and after system header guard

cub/cub/block/specializations/block_topk_air.cuh
```diff
@@ -0,0 +1,502 @@
...
+CUB_NAMESPACE_BEGIN
+
+namespace detail
+{
+template <typename SortKeyT>
+struct compare_key_prefix_op
+{
+  SortKeyT prefix_mask;
+  SortKeyT key_prefix;
+  _CCCL_HOST_DEVICE _CCCL_FORCEINLINE bool operator()(SortKeyT sort_key) const
```
miscco:
```suggestion
  [[nodiscard]] _CCCL_API _CCCL_FORCEINLINE constexpr bool operator()(SortKeyT sort_key) const noexcept
```

cub/cub/block/specializations/block_topk_air.cuh
```diff
@@ -0,0 +1,502 @@
...
+      {
+        storage.stage.passes.histogram[bin_idx] = thread_buckets[i];
+      }
+    }
+
+    __syncthreads();
+  }
+
+  // Identify the bucket that the k-th item falls into
+  _CCCL_DEVICE _CCCL_FORCEINLINE void choose_bucket(histo_counter_t k)
```
miscco: This is a very heavy use of force inline

Do we really need it everywhere?

cub/cub/block/specializations/block_topk_air.cuh
```diff
@@ -0,0 +1,502 @@
...
+      for (int i = 0; i < items_per_thread; ++i)
+      {
+        const bit_ordered_type key_prefix = unsigned_keys[i] & prefix_mask;
+
+        const bool is_selected  = key_prefix < kth_prefix;
+        const bool is_candidate = key_prefix == kth_prefix;
+        int item_class          = is_selected ? 0 : (is_candidate ? 1 : -1);
+
+        const bool is_valid = (IsFullTile || static_cast<int>(linear_tid) * items_per_thread + i < num_valid);
+        item_class          = is_valid ? item_class : -1;
```
miscco: I believe this should rather be
```suggestion
        const bool is_valid  = (IsFullTile || static_cast<int>(linear_tid) * items_per_thread + i < num_valid);
        const int item_class = is_valid ? is_selected ? 0 : (is_candidate ? 1 : -1) : -1;
```

cub/cub/block/specializations/block_topk_air.cuh
```diff
@@ -0,0 +1,502 @@
...
+
+        // Untwiddle the key before storing in shared memory
+        if constexpr (SelectDirection == detail::topk::select::max)
+        {
+          unsigned_keys[i] = bit_ordered_inversion::inverse(decomposer, unsigned_keys[i]);
+        }
+        unsigned_keys[i] = bit_ordered_conversion::from_bit_ordered(decomposer, unsigned_keys[i]);
+
+        if (item_class >= 0)
```
miscco: Nitpick: Those two branches of the big if are almost identical, except for small changes.

However, the variable names and the conditions are then quite different. This check here is effectively the same as `is_valid && is_candidate` which would be much closer to what is written above and in my opinion would improve readability of the algorithm

cub/cub/block/specializations/block_topk_air.cuh
elstehle: Thanks! Auto-includes 🙈

cub/cub/block/specializations/block_topk_air.cuh
elstehle: I had benchmarked it more extensively in `AgentTopK`, i.e., the implementation for single-problem `DeviceTopK`, where it had improved performance.

I will do another pass of benchmarking here too. I'm planning to defer it to https://github.com/NVIDIA/cccl/issues/7799, though, to be sure we're not hitting some unrelated bottleneck.

elstehle: I've added an item to https://github.com/NVIDIA/cccl/issues/7799.

cub/cub/block/specializations/block_topk_air.cuh
elstehle: Yeah, all of this (two branches & using `item_class`) has mostly been done as a performance concern, saving us a few instructions. I unified the branches and re-benchmarked. It gives ~2% slowdown in some cases. I will leave them unified for now and reevaluate performance in https://github.com/NVIDIA/cccl/issues/7799

cub/cub/block/specializations/block_topk_air.cuh
elstehle: I've unified the branches, focusing on readability over performance.

cub/cub/block/specializations/block_topk_air.cuh
```diff
@@ -0,0 +1,466 @@
...
+        } exchange;
+      } select;
+    } stage;
+  };
+
+  /// Shared storage reference
+  TempStorage_& storage;
+
+  /// Linear thread index
+  unsigned int linear_tid;
```
pauleonix:
```suggestion
  int linear_tid;
```
Suggestion: Do the cast once instead of at every use of the member variable.

cub/cub/block/specializations/block_topk_air.cuh
```diff
@@ -0,0 +1,466 @@
...
+    {
+      storage.stage.passes.histogram[histo_offset + threadIdx.x] = 0;
+    }
+  }
+
+  // Compute histogram over keys
+  template <bool IsFullTile, typename DigitExtractorT, typename FilterOpT>
+  _CCCL_DEVICE _CCCL_FORCEINLINE void compute_histograms(
+    const bit_ordered_type (&unsigned_keys)[items_per_thread],
+    int num_valid,
```
pauleonix:
```suggestion
    int valid_items,
```
Suggestion: This is the parameter name used in the `BlockScan` overloads I added (which I took from another existing CUB API).

cub/cub/block/specializations/block_topk_air.cuh
```diff
@@ -0,0 +1,466 @@
...
+  {
+    _CCCL_PRAGMA_UNROLL_FULL()
+    for (int i = 0; i < items_per_thread; ++i)
+    {
+      const auto item_index      = static_cast<int>(linear_tid) * items_per_thread + i;
+      const bit_ordered_type key = unsigned_keys[i];
+      if ((IsFullTile || item_index < num_valid) && filter_op(key))
+      {
+        const auto digit = digit_extractor.Digit(key);
+        atomicAdd(&storage.stage.passes.histogram[digit], histo_counter_t{1});
```
pauleonix, line 151: Question: Why are we not reusing the existing `cub::BlockHistogram` here?

cub/cub/block/specializations/block_topk_air.cuh
```diff
@@ -0,0 +1,466 @@
...
+    const int base = static_cast<int>(linear_tid) * buckets_per_thread;
+
+    _CCCL_PRAGMA_UNROLL_FULL()
+    for (int i = 0; i < buckets_per_thread; ++i)
+    {
+      const int bin_idx = base + i;
+      if (bin_idx < num_buckets)
+      {
+        const histo_counter_t prev = (bin_idx == 0) ? 0 : storage.stage.passes.histogram[bin_idx - 1];
+        const histo_counter_t cur  = storage.stage.passes.histogram[bin_idx];
```
pauleonix, line 197: I wonder if using `cub::BlockAdjacentDifference` with the data already in registers from `compute_bin_offsets()` would improve performance over writing all bin offsets to shared memory and reading all of them again twice. At least in cases where `buckets_per_thread > 1`, which - I guess - is the exception given the default to `1 << 8` bins? On the other hand using `cub::BlockAdjacentDifference` would warrant more synchronization (or smem use) as it could not rely on threads reading from and writing to the same position in smem (`storage.stage.passes.histogram[bin_idx]`).

cub/cub/block/specializations/block_topk_air.cuh
```diff
@@ -0,0 +1,466 @@
...
+    // The total number of selected items
+    total_selected = 0;
+
+    const int total_bits = end_bit - begin_bit;
+    const int num_passes = (total_bits > 0) ? ::cuda::ceil_div(total_bits, RadixBits) : 0;
+    for (int pass = 0; pass < num_passes; ++pass)
+    {
+      // Bit-range & mask of the current pass
+      const int pass_end_bit           = end_bit - pass * RadixBits;
+      const int pass_begin_bit         = ::cuda::std::max(pass_end_bit - RadixBits, begin_bit);
```
pauleonix:
```suggestion
      const int pass_begin_bit         = (::cuda::std::max) (pass_end_bit - RadixBits, begin_bit);
```
To avoid breaking due to `max` preprocessor macro being defined.

cub/cub/block/specializations/block_topk_air.cuh
```diff
@@ -0,0 +1,466 @@
...
+    for (int i = 0; i < buckets_per_thread; ++i)
+    {
+      const int bin_idx = base + i;
+      if (bin_idx < num_buckets)
+      {
+        storage.stage.passes.histogram[bin_idx] = thread_buckets[i];
+      }
+    }
+
+    __syncthreads();
```
pauleonix: This seems redundant with the synchronization done in `get_kth_key_prefix()`.
```suggestion
```

cub/cub/block/specializations/block_topk_air.cuh
```diff
@@ -0,0 +1,466 @@
...
+      if constexpr (SelectDirection == detail::topk::select::max)
+      {
+        unsigned_keys[i] = bit_ordered_inversion::inverse(decomposer, unsigned_keys[i]);
+      }
+      unsigned_keys[i] = bit_ordered_conversion::from_bit_ordered(decomposer, unsigned_keys[i]);
+
+      if (is_valid && (is_selected || is_candidate))
+      {
+        const histo_counter_t selected_offset = atomicAdd(&storage.stage.select.selected_offset[item_class], 1);
+        reinterpret_cast<bit_ordered_type*>(storage.stage.select.exchange.keys)[selected_offset] = unsigned_keys[i];
```
pauleonix:
```suggestion
        storage.stage.select.exchange.keys[selected_offset] = ::cuda::std::bit_cast<KeyT>(unsigned_keys[i]);
```

cub/cub/block/specializations/block_topk_air.cuh
```diff
@@ -0,0 +1,466 @@
...
+
+    // TODO (elstehle): Add support for custom decomposers
+    identity_decomposer_t decomposer;
+
+    // Get bit-twiddled sortkeys
+    bit_ordered_type(&unsigned_keys)[ItemsPerThread] = reinterpret_cast<bit_ordered_type(&)[ItemsPerThread]>(keys);
+    _CCCL_PRAGMA_UNROLL_FULL()
+    for (int i = 0; i < items_per_thread; ++i)
+    {
+      unsigned_keys[i] = bit_ordered_conversion::to_bit_ordered(decomposer, unsigned_keys[i]);
```
pauleonix: I expect the compiler to do the aliasing for you. So this should not increase register pressure.
```suggestion
    bit_ordered_type unsigned_keys[ItemsPerThread];
    _CCCL_PRAGMA_UNROLL_FULL()
    for (int i = 0; i < items_per_thread; ++i)
    {
      unsigned_keys[i] = bit_ordered_conversion::to_bit_ordered(decomposer, ::cuda::std::bit_cast<bit_ordered_type>(keys[i]));
```

cub/cub/block/specializations/block_topk_air.cuh
```diff
@@ -0,0 +1,466 @@
...
+
+#if defined(_CCCL_IMPLICIT_SYSTEM_HEADER_GCC)
+#  pragma GCC system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_CLANG)
+#  pragma clang system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_MSVC)
+#  pragma system_header
+#endif // no system header
+
+#include <cub/block/block_exchange.cuh>
```
pauleonix: Unused?
```suggestion
```

cub/cub/agent/agent_batched_topk.cuh
```diff
@@ -201,15 +202,15 @@ struct agent_batched_topk_worker_per_segment
     // Perform Block Top-K
     if constexpr (is_keys_only)
     {
-      const bool is_successful_dispatch =
-        detail::params::dispatch_discrete(select_directions, segment_id, [this, &thread_keys, k](auto direction_tag) {
+      const bool is_successful_dispatch = detail::params::dispatch_discrete(
```
fbusato: this is a function, should be 
```suggestion
      const bool is_successful_dispatch = cub::detail::params::dispatch_discrete(
```

cub/cub/block/block_topk.cuh
```diff
@@ -17,63 +17,86 @@
 #  pragma system_header
 #endif // no system header
```
fbusato, line 15: we should also update the Copyright

cub/cub/block/block_topk.cuh
```diff
@@ -17,63 +17,86 @@
...
 
+#include <cub/block/block_exchange.cuh>
 #include <cub/block/block_radix_sort.cuh>
+#include <cub/block/block_scan.cuh>
+#include <cub/block/specializations/block_topk_air.cuh>
+#include <cub/device/dispatch/dispatch_common.cuh>
 #include <cub/util_ptx.cuh>
 #include <cub/util_type.cuh>
 
+#include <cuda/std/cstdint>
```
fbusato:
```suggestion
#include <cub/block/specializations/block_topk_air.cuh>
#include <cub/device/dispatch/dispatch_common.cuh>
#include <cub/util_type.cuh>
```

cub/cub/block/block_topk.cuh
```diff
@@ -17,63 +17,86 @@
...
   };
 
 private:
-  TempStorage& temp_storage;
+  TempStorage& storage;
 
 public:
-  _CCCL_DEVICE _CCCL_FORCEINLINE block_topk(TempStorage& temp_storage)
-      : temp_storage(temp_storage)
+  _CCCL_DEVICE _CCCL_FORCEINLINE block_topk(TempStorage& storage)
```
fbusato: the branch looks behind `main`. We should use  `_CCCL_DEVICE` + `inline`

cub/cub/block/specializations/block_topk_air.cuh
```diff
@@ -0,0 +1,466 @@
...
+#include <cuda/std/cstdint>
+
+CUB_NAMESPACE_BEGIN
+
+namespace detail
+{
+template <typename SortKeyT>
+struct compare_key_prefix_op
+{
+  SortKeyT prefix_mask;
```
fbusato, line 35: I would add a static assertion to ensure the type is actually unsigned

cub/cub/block/specializations/block_topk_air.cuh
```diff
@@ -0,0 +1,466 @@
...
+{
+private:
+  // TODO (elstehle): Make this configurable
+  // Whether to include all items tied with the k-th key when selecting top-k
+  static constexpr bool expand_k_to_include_ties = false;
+
+  static constexpr int block_threads    = BlockThreads;
+  static constexpr int items_per_thread = ItemsPerThread;
+  static constexpr int tile_items       = block_threads * items_per_thread;
+  static constexpr int num_buckets      = (1 << RadixBits);
```
fbusato: this looks dangerous
```suggestion
  static constexpr int num_buckets      = int{1u << RadixBits};
```

cub/cub/block/specializations/block_topk_air.cuh
```diff
@@ -0,0 +1,466 @@
...
+  {
+    // Initialize histogram bin counts to zeros
+    int histo_offset = 0;
+
+    // Loop unrolling is beneficial for performance here
+    _CCCL_PRAGMA_UNROLL_FULL()
+    for (; histo_offset + block_threads <= num_buckets; histo_offset += block_threads)
+    {
+      storage.stage.passes.histogram[histo_offset + threadIdx.x] = 0;
+    }
```
fbusato, line 127: maybe we could use a stardard loop `for (int i = ...; )` and compute `histo_offset = num_buckets / block_threads` after that

cub/cub/block/specializations/block_topk_air.cuh
fbusato: strongly suggesting to use `cuda::atomic` here

cub/cub/block/specializations/block_topk_air.cuh
fbusato: is any value for `num_valid` valid? if not, we should add an assertion

cub/cub/block/specializations/block_topk_air.cuh
```diff
@@ -0,0 +1,466 @@
...
+  _CCCL_DEVICE _CCCL_FORCEINLINE void compute_histograms(
+    const bit_ordered_type (&unsigned_keys)[items_per_thread],
+    int num_valid,
+    DigitExtractorT digit_extractor,
+    FilterOpT filter_op)
+  {
+    _CCCL_PRAGMA_UNROLL_FULL()
+    for (int i = 0; i < items_per_thread; ++i)
+    {
+      const auto item_index      = static_cast<int>(linear_tid) * items_per_thread + i;
```
fbusato: `linear_tid` should be `int` natively

cub/cub/block/specializations/block_topk_air.cuh
```diff
@@ -0,0 +1,466 @@
...
+  // Compute prefix sum over buckets
+  _CCCL_DEVICE _CCCL_FORCEINLINE void compute_bin_offsets()
+  {
+    histo_counter_t thread_buckets[buckets_per_thread]{};
+    const int base = static_cast<int>(linear_tid) * buckets_per_thread;
+
+    _CCCL_PRAGMA_UNROLL_FULL()
+    for (int i = 0; i < buckets_per_thread; ++i)
+    {
+      const int bin_idx = base + i;
```
fbusato, line 165: we could also shift the `histogram` pointer instead by `base`

cub/cub/block/specializations/block_topk_air.cuh
```diff
@@ -0,0 +1,466 @@
...
+      if (bin_idx < num_buckets)
+      {
+        thread_buckets[i] = storage.stage.passes.histogram[bin_idx];
+      }
+    }
+
+    block_scan_t(storage.stage.passes.scan_temp_storage).InclusiveSum(thread_buckets, thread_buckets);
```
fbusato, line 172: please check that the code passes `compute-sanitizer` race check

cub/cub/block/specializations/block_topk_air.cuh
```diff
@@ -0,0 +1,466 @@
...
+  {
+    const int base = static_cast<int>(linear_tid) * buckets_per_thread;
+
+    _CCCL_PRAGMA_UNROLL_FULL()
+    for (int i = 0; i < buckets_per_thread; ++i)
+    {
+      const int bin_idx = base + i;
+      if (bin_idx < num_buckets)
+      {
+        const histo_counter_t prev = (bin_idx == 0) ? 0 : storage.stage.passes.histogram[bin_idx - 1];
```
fbusato, line 196: the same pattern occurs multiple times, consider using a reference for  `storage.stage.passes`

cub/cub/block/specializations/block_topk_air.cuh
```diff
@@ -0,0 +1,466 @@
...
+    int num_valid,
+    int begin_bit,
+    int end_bit,
+    int& total_selected,
+    int& num_candidates,
+    bit_ordered_type& kth_key_prefix,
+    bit_ordered_type& prefix_mask,
+    DecomposerT decomposer = DecomposerT{})
+  {
+    // We only consider candidates identified in the previous pass, i.e., ((sortkey & prefix_mask) == kth_prefix)
```
fbusato, line 232: there are many parameters, please add some preconditions with assertions

cub/cub/block/specializations/block_topk_air.cuh
```diff
@@ -0,0 +1,466 @@
...
+    // With each pass, we identify a wider prefix of the splitter key
+    kth_key_prefix = 0;
+    prefix_mask    = 0;
+
+    // The total number of selected items
+    total_selected = 0;
+
+    const int total_bits = end_bit - begin_bit;
+    const int num_passes = (total_bits > 0) ? ::cuda::ceil_div(total_bits, RadixBits) : 0;
```
fbusato:
```suggestion
    const int total_bits = ::max(end_bit - begin_bit, 0);
    const int num_passes = ::cuda::ceil_div(total_bits, RadixBits);
```

cub/cub/block/specializations/block_topk_air.cuh
```diff
@@ -0,0 +1,466 @@
...
+    for (int pass = 0; pass < num_passes; ++pass)
```
fbusato, line 242: non critical but consider a loop with a fixed number of iterations + unroll

cub/cub/block/specializations/block_topk_air.cuh
fbusato: if we know that we are using only integers, we can also use `::max`

cub/cub/block/specializations/block_topk_air.cuh
fbusato: the loop is pretty big, so it is hard to tell without benchmarking

cub/cub/block/specializations/block_topk_air.cuh
```diff
@@ -0,0 +1,466 @@
...
+    if ((!IsFullTile && k >= num_valid) || k >= tile_items)
+    {
+      return;
+    }
+
+    // TODO (elstehle): Add support for custom decomposers
+    identity_decomposer_t decomposer;
+
+    // Get bit-twiddled sortkeys
+    bit_ordered_type(&unsigned_keys)[ItemsPerThread] = reinterpret_cast<bit_ordered_type(&)[ItemsPerThread]>(keys);
```
fbusato, line 332: I would prefer `cuda::std::bit_cast` over potentially UB. Also, I would suggest `auto& unsigned_keys = `

cub/cub/block/specializations/block_topk_air.cuh
```diff
@@ -0,0 +1,466 @@
...
+    if (k <= 0)
+    {
+      return;
+    }
+
+    // TODO (elstehle): Short-circuit if begin_bit is constrained to be non-negative
+    if (begin_bit < 0)
+    {
+      begin_bit = 0;
+    }
```
fbusato:
```suggestion
    begin_bit = ::max(begin_bit, 0);
```

cub/cub/block/specializations/block_topk_air.cuh
```diff
@@ -0,0 +1,466 @@
...
+    if ((num_buckets % block_threads != 0) && (histo_offset + threadIdx.x < num_buckets))
+    {
+      storage.stage.passes.histogram[histo_offset + threadIdx.x] = 0;
+    }
+  }
+
+  // Compute histogram over keys
+  template <bool IsFullTile, typename DigitExtractorT, typename FilterOpT>
+  _CCCL_DEVICE _CCCL_FORCEINLINE void compute_histograms(
+    const bit_ordered_type (&unsigned_keys)[items_per_thread],
```
fbusato, line 138: would not be better to use `cuda::std::array` or `span`?

cub/cub/block/specializations/block_topk_air.cuh
```diff
@@ -0,0 +1,466 @@
...
+  SortKeyT prefix_mask;
+  SortKeyT key_prefix;
+  [[nodiscard]] _CCCL_API _CCCL_FORCEINLINE constexpr bool operator()(SortKeyT sort_key) const noexcept
+  {
+    return (sort_key & prefix_mask) == (key_prefix);
+  }
+};
+
+template <typename KeyT, int BlockThreads, int ItemsPerThread, typename ValueT = NullType, int RadixBits = 8>
+class block_topk_air
```
fbusato, line 56: I would describe the workflow/algorithm because it consists of several steps

cub/test/catch2_test_device_segmented_topk_keys.cu
```diff
@@ -125,7 +125,7 @@ C2H_TEST("DeviceBatchedTopK::{Min,Max}Keys work with small fixed-size segments",
 
   // Generate number of segments
   const segment_index_t num_segments = GENERATE_COPY(
-    values({segment_index_t{1}, segment_index_t{42}}), take(4, random(segment_index_t{1}, segment_index_t{10000})));
+    values({segment_index_t{1}, segment_index_t{42}}), take(4, random(segment_index_t{1}, segment_index_t{1000})));
```
fbusato, line 135: not related to this PR, but it looks that there are unused headers

cub/cub/block/block_topk.cuh
```diff
@@ -17,63 +17,86 @@
...
+
 CUB_NAMESPACE_BEGIN
 
 namespace detail
 {
+enum class block_topk_algorithm
+{
+  air_top_k
+};
+
```
fbusato: is it for public exposition? I don't see any other reference

cub/cub/block/specializations/block_topk_air.cuh
```diff
@@ -0,0 +1,466 @@
...
+    // TODO (elstehle): Add support for custom decomposers
+    identity_decomposer_t decomposer;
+
+    // Get bit-twiddled sortkeys
+    bit_ordered_type(&unsigned_keys)[ItemsPerThread] = reinterpret_cast<bit_ordered_type(&)[ItemsPerThread]>(keys);
+    _CCCL_PRAGMA_UNROLL_FULL()
+    for (int i = 0; i < items_per_thread; ++i)
+    {
+      unsigned_keys[i] = bit_ordered_conversion::to_bit_ordered(decomposer, unsigned_keys[i]);
+      unsigned_keys[i] = BaseDigitExtractor<KeyT>::ProcessFloatMinusZero(unsigned_keys[i]);
```
fbusato: do we reverse `-0.0f` or is it expected that it translates to `+0.0f`

cub/cub/block/specializations/block_topk_air.cuh
```diff
@@ -0,0 +1,466 @@
...
+      const bool is_selected  = key_prefix < kth_prefix;
+      const bool is_candidate = key_prefix == kth_prefix;
+
+      // We differentiate between candidates and selected only if not all candidates make it into the top-k items.
+      int item_class = (!select_all_candidates) && is_candidate ? 1 : 0;
+
+      // Untwiddle the key before storing in shared memory
+      if constexpr (SelectDirection == detail::topk::select::max)
+      {
+        unsigned_keys[i] = bit_ordered_inversion::inverse(decomposer, unsigned_keys[i]);
```
fbusato, line 433: is it applied to all items? even outside the candidate range?

cub/cub/block/specializations/block_topk_air.cuh
```diff
@@ -0,0 +1,466 @@
...
+    bit_ordered_type(&unsigned_keys)[ItemsPerThread] = reinterpret_cast<bit_ordered_type(&)[ItemsPerThread]>(keys);
+    _CCCL_PRAGMA_UNROLL_FULL()
+    for (int i = 0; i < items_per_thread; ++i)
+    {
+      unsigned_keys[i] = bit_ordered_conversion::to_bit_ordered(decomposer, unsigned_keys[i]);
+      unsigned_keys[i] = BaseDigitExtractor<KeyT>::ProcessFloatMinusZero(unsigned_keys[i]);
+      if constexpr (SelectDirection == detail::topk::select::max)
+      {
+        unsigned_keys[i] = bit_ordered_inversion::inverse(decomposer, unsigned_keys[i]);
+      }
```
fbusato: what happens if `k < num_valid` ? my concern is that `unsigned_keys` could contain garbage data (UB), and this affects the correctness of the logic for ranking

fbusato (changes requested):

cub/cub/block/block_topk.cuh
elstehle: This was in preparation for further specializations BlockTopK specializations but as it is not used in this PR, I dropped it

cub/cub/block/block_topk.cuh
elstehle: I must have missed this change. @fbusato, could you help me find the relevant PR, style guide, or resource that introduced the switch. In CUB's code base I could only find two occurrences of `_CCCL_DEVICE inline`. So, I'm probably missing something.

cub/cub/block/specializations/block_topk_air.cuh
elstehle: This was the canonical way from CUB's device histogram which gave 2% improvement over the previous version. Since I expect this part to be abstracted away in future anyways, I'd prefer to keep it this way for now to not have to re-benchmark.

cub/cub/block/specializations/block_topk_air.cuh
elstehle: Yes, generally agree. However, the data provision is planned to be abstracted away anyways in https://github.com/NVIDIA/cccl/issues/7799 to allow supporting both registers and shared memory. I added a note to support `cuda::std::array` to the issue and will leave this as follow-up work.

cub/cub/block/specializations/block_topk_air.cuh
elstehle: Thanks, Paul. Changed the naming.

elstehle: The only minor catch is that if `valid_items` is larger than `tile_size` that we obviously won't be processing more items than we have. Nontheless, I added an assert for expect reasonable values.

cub/cub/block/specializations/block_topk_air.cuh
elstehle: Could you elaborate on the benefit of that? Is there something beyond stylistic preference?

cub/cub/block/specializations/block_topk_air.cuh
elstehle: Good catch. Dropped it.

cub/cub/block/specializations/block_topk_air.cuh
elstehle: Yes, it's applied to oob items but we safeguard against doing anything with oob items subsequently:

```
const bool is_valid     = (IsFullTile || static_cast<int>(linear_tid) * items_per_thread + i < num_valid);
...
if (is_valid && (is_selected || is_candidate)){
// guarded use of unsigned_keys
}
```

cub/cub/block/specializations/block_topk_air.cuh
elstehle: Had run the `pairs` tests with both `racecheck` and `memcheck` and no errors/races.

cub/cub/block/specializations/block_topk_air.cuh
elstehle: > what happens if k < num_valid ?

`k < num_valid` is the expected case, so I want to make sure I understand your concern correctly. Is it that we are twiddling the oob keys unconditionally here? If so, we do guard against considering oob keys in relevant parts (e.g., histogram computation).

cub/cub/block/specializations/block_topk_air.cuh
elstehle: I've added this as follow-up work to https://github.com/NVIDIA/cccl/issues/7799. 

I'm all in for using `BlockHistogram` and `BlockAdjacentDifference` here. But in a next step this algorithm will be extended to take its inputs via shared memory, which would void the effort of temporarily switching to these building blocks.

cub/cub/block/specializations/block_topk_air.cuh
elstehle: Yes, that's a great observation about using the data already in registers and limiting shared memory exchange to just the first and last item of each thread. I've added it as follow-up work to https://github.com/NVIDIA/cccl/issues/7799

cub/cub/block/specializations/block_topk_air.cuh
elstehle: ...it's big and it has an early exit. But we can still give it a try. I will leave this as part of a benchmarking effort in https://github.com/NVIDIA/cccl/issues/7799.

cub/cub/block/specializations/block_topk_air.cuh
elstehle: Added a description.

cub/cub/block/specializations/block_topk_air.cuh
elstehle: Good catch, Federico. I had to change the logic here and add tests to verify we revert -0.0f back after the twiddling.

elstehle:
pre-commit.ci autofix

cub/cub/block/specializations/block_topk_air.cuh
fbusato: not stylistic, my (little) concern is that the compiler doesn't optimize the code and performance the sum multiple times. Up to you, non-critical

cub/cub/block/block_topk.cuh
fbusato: Sorry, I meant `_CCCL_DEVICE_API`, the PR is https://github.com/NVIDIA/cccl/pull/4584

cub/cub/block/specializations/block_topk_air.cuh
fbusato: yes, this is my concern. I see the guards, so it should be fine

fbusato (approved):

cub/cub/block/specializations/block_topk_air.cuh
```diff
@@ -0,0 +1,529 @@
...
+    int begin_bit,
+    int end_bit,
+    int& total_selected,
+    int& num_candidates,
+    bit_ordered_type& kth_key_prefix,
+    bit_ordered_type& prefix_mask,
+    DecomposerT decomposer = DecomposerT{})
+  {
+    // Preconditions
+    static constexpr int max_bit = int(sizeof(KeyT) * 8);
```
miscco: C:\cccl\cub\cub/block/specializations/block_topk_air.cuh(223): error #177-D: variable "max_bit" was declared but never referenced

elstehle: Thanks for saving me a lookup into the CI logs, Michael! 💚

cub/cub/block/block_topk.cuh
elstehle: I see. Thanks!

PR #7399, opened 2026-01-28
[c.parallel]: migrate transform to use jit templates instead of string based implementations

### LINKED ISSUE #7387: [BUG] c.parallel: Stateful ops don't work with `transform` (and some other algorithms)

NaderAlAwar:
## Description

closes #7387

Additionally, this PR makes changes to jit_templates to handle algorithms which could contain more than one type erased storage type. Prior to this, it was assumed that only one `storage_t` exists, but this change allows the caller to use a `tagged_arg` which contains a tag specifying the name of the storage type. This enables algorithms to define storage types with different sizes and alignments.

## Checklist

### DIFF
--- a/c/parallel/src/jit_templates/templates/operation.h
+++ b/c/parallel/src/jit_templates/templates/operation.h
@@ -295,6 +295,22 @@ __device__ {1} operator{2}(const {3} & lhs, const {3} & rhs)
           tagged_arg<RetStorageT, cccl_type_info> ret,
           tagged_arg<ArgStorageTs, cccl_type_info>... arguments)
   {
+    // We cannot use well-known operations with storage types as there
+    // is currently no way to tell whether multiple storage types,
+    // e.g., input and output, are the same type. This is necessary in
+    // order for operations like `negate` to work, as it requires both
+    // input and output to be the same type. For now, this check short
+    // circuits the specialization process for well-known operations
+    // with storage types. The code below that checks whether any of
+    // the arguments or the return type are storage types will
+    // currently not run, but is left here as it will be needed in the
+    // future.
+    if (ret.value.type == cccl_type_enum::CCCL_STORAGE
+        || ((arguments.value.type == cccl_type_enum::CCCL_STORAGE) || ...))
+    {
+      return cuda::std::nullopt;
+    }
+
     auto entry = well_known_operation_description(operation.type);
     if (!entry)
     {


c/parallel/src/jit_templates/traits.h
```diff
@@ -27,16 +27,68 @@ template <typename Tpl>
...
 struct specialization
 {
   std::string type_name;
   std::string aux_code = "";
 };
 
-template <typename Tag,
-          typename Traits,
-          typename... Args,
-          typename = Traits::template type<void, parameter_mapping<Args>::archetype...>>
```
griwes, line 39: I would like us to preserve the archetype part, either here or in the body of the function, so that if someone tries to pass in nonsensical arguments that are not accepted by the template in question, they still get an error here rather than at JIT time.

c/parallel/src/jit_templates/templates/operation.h
```diff
@@ -292,6 +292,11 @@ __device__ {1} operator{2}(const {3} & lhs, const {3} & rhs)
   template <typename, typename... Args>
   static cuda::std::optional<specialization> special(cccl_op_t operation, cccl_type_info ret, Args... arguments)
   {
+    if (ret.type == cccl_type_enum::CCCL_STORAGE || ((arguments.type == cccl_type_enum::CCCL_STORAGE) || ...))
+    {
+      return cuda::std::nullopt;
+    }
```
griwes: This is a bit of a regression; originally, @gevtushenko asked to also lower to `cuda::std::foo` for storage types, after giving them an `operator foo`, and that's handled within the `if (entry->symbol)` branch below. Can you see if that can be preserved here?

c/parallel/src/jit_templates/templates/operation.h
```diff
@@ -320,6 +325,15 @@ __device__ {1} operator{2}(const {3} & lhs, const {3} & rhs)
...
     return specialization{std::format(entry->name, cccl_type_enum_to_name(type_info_table[1]).c_str()), aux};
   }
+
+  template <typename Tag, typename RetStorageT, typename... ArgStorageTs>
+  static cuda::std::optional<specialization>
+  special(cccl_op_t operation,
+          tagged_arg<RetStorageT, cccl_type_info> ret,
+          tagged_arg<ArgStorageTs, cccl_type_info>... arguments)
+  {
+    return special<Tag>(operation, ret.value, arguments.value...);
```
griwes: If we preserve the behavior I mention above, the logic here should be reversed, i.e. `special` should accept anything, pass it through a `wrap` function that either returns the argument as is when it's tagged, or tags it with `storage_t` if it's not, and then the calls to `cccl_type_enum_to_name` above should use the tag for each argument, as the function above now has access to it.

c/parallel/src/jit_templates/traits.h
NaderAlAwar: Added it back

c/parallel/src/jit_templates/templates/operation.h
NaderAlAwar: Reverted the change, although now this test https://github.com/NVIDIA/cccl/blob/ce3629efcbe10160af85a7d537578a60884f978a/c/parallel/test/test_transform.cpp#L351 is failing since it expects input and output types to be the same, which is no longer the case now that we introduced different input and output storage types, which won't work since the op type is set to `NEGATE`

NaderAlAwar: We discussed this offline:
* The well known operators path with storage types will currently not work because there is no way for the C API to tell whether two storage types are the same or not. This is an issue because operators like `negate` require the input and output types to be the same
* It might be possible to use the upper bits of the type enum as a storage type discriminator, but this is beyond the scope of this PR
* We will open an issue to enhance the C API to tell whether two storage types are the same

For now I will keep my original change which bails out if a well known operator is used with storage types

PR #7405, opened 2026-01-29
part deux: make the abi of `__basic_any` compatible between c++17 and c++20

ericniebler:
## Description

part 2 of #7401. @bdice confirms that this, in addition to #7401, fixes the ABI issue in RMM.

## Checklist


bdice (approved):
Thank you! I confirmed this is necessary and sufficient with some testing in https://github.com/rapidsai/rmm/pull/2233.

miscco (approved):

PR #7406, opened 2026-01-29
Add versionadded annotations to CUB public APIs

cliffburdick:
## Description
Document when each public API was introduced using Sphinx versionadded directives. This helps users understand API availability when targeting specific CUDA Toolkit versions.

Version mappings:
- CCCL 2.2.0 -> CUDA Toolkit 12.3
- CCCL 3.2.0 -> CUDA Toolkit 13.2
- CCCL 3.3.0

For CCCL 3.3.0 there is no mapping to CTK version and is left empty for now.

Annotations added to:
- CUB block primitives (block_load, block_store, block_reduce, etc.)
- CUB device algorithms (device_reduce, device_scan, device_sort, etc.)
- CUB warp primitives (warp_reduce, warp_scan, warp_exchange, etc.)
- CUB iterators and grid utilities

The final output looks like this: 
image


cub/cub/block/block_adjacent_difference.cuh
```diff
@@ -266,6 +280,7 @@ public:
   //! ``{ [4,2,1,1], [1,1,1,1], [2,3,3,3], [3,4,1,4], ... }``.
   //! The corresponding output ``result`` in those threads will be
   //! ``{ [4,-2,-1,0], [0,0,0,0], [1,1,0,0], [0,1,-3,3], ... }``.
+  //!
   //! @endrst
```
bernhardmgruber, line 283: Suggestion: Can we revert these empty lines? I don't see a benefit of having them. Or do they fix a bug in the docs?

cub/cub/block/block_shuffle.cuh
```diff
@@ -108,12 +118,13 @@ public:
...
   //! @rst
-  //!
   //! Each *thread*\ :sub:`i` obtains the ``input`` provided by *thread*\ :sub:`i + distance`.
   //! The offset ``distance`` may be negative.
   //!
-  //! - @smemreuse
+  //! .. versionadded:: 2.2.0
+  //!    First appears in CUDA Toolkit 12.3.
   //!
+  //! - @smemreuse
```
bernhardmgruber, line 127: Why does the clause on SMEM reuse need to be after the versionadded block? Is this done on purpose? If not, I would change the order again to:
```suggestion
  //! - @smemreuse
  //!
  //! .. versionadded:: 2.2.0
  //!    First appears in CUDA Toolkit 12.3.
```
Comment applies to more places.

bernhardmgruber:
Except the above two comments, the changeset LGTM

cub/cub/block/block_shuffle.cuh
cliffburdick: I was trying to put the versionadded as close to the function description as possible so it's not buried in a wall of text. I can swap this if you want.

cub/cub/block/block_adjacent_difference.cuh
cliffburdick: I can remove them

cub/cub/block/block_shuffle.cuh
cliffburdick: I think it looks/flows better with the versionadded first since it's brief. If you look at some of these annotations they can be 20+ lines long and you lose the versionadded tag when reading it.

bernhardmgruber: Fine.

miscco (approved):
This looks amazing. Thanks a lot @cliffburdick 🎉

fbusato (approved):
thanks a lot 🙏

PR #7411, opened 2026-01-29
Move host stdlib wrappers to `<cuda/std/__host_stdlib>` directory

davebayer:
As there are more host stdlib headers that we would like to indirectly include (need to be wrapped by macro, checked, ...), I'd like to move all of these wrappers to a single location.

### DIFF
--- a/libcudacxx/include/cuda/std/__host_stdlib/math.h
+++ b/libcudacxx/include/cuda/std/__host_stdlib/math.h
@@ -1,0 +1,49 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of libcu++, the C++ Standard Library for your entire system,
+// under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef _CUDA_STD___HOST_STDLIB_MATH_H
+#define _CUDA_STD___HOST_STDLIB_MATH_H
+
+#include <cuda/std/detail/__config>
+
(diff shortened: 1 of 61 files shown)


libcudacxx/include/cuda/std/__host_stdlib/math.h
```diff
@@ -0,0 +1,49 @@
...
+    || defined(log10) || defined(log2) || defined(log1p) || defined(pow) || defined(sqrt) || defined(cbrt)          \
+    || defined(hypot) || defined(sin) || defined(cos) || defined(tan) || defined(asin) || defined(acos)             \
+    || defined(atan) || defined(atan2) || defined(sinh) || defined(cosh) || defined(tanh) || defined(asinh)         \
+    || defined(acosh) || defined(atanh) || defined(erf) || defined(erfc) || defined(tgamma) || defined(lgamma)      \
+    || defined(ceil) || defined(floor) || defined(trunc) || defined(round) || defined(lround) || defined(llround)   \
+    || defined(nearbyint) || defined(rint) || defined(lrint) || defined(llrint) || defined(frexp) || defined(ldexp) \
+    || defined(scalbn) || defined(scalbln) || defined(ilogb) || defined(logb) || defined(nextafter)                 \
+    || defined(nexttoward) || defined(copysign) || defined(fpclassify) || defined(isfinite) || defined(isinf)       \
+    || defined(isnan) || defined(isnormal) || defined(signbit) || defined(isgreater) || defined(isgreaterequal)     \
+    || defined(isless) || defined(islessequal) || defined(islessgreater) || defined(isunordered)
```
miscco, line 42: AFAIK only some of those are potentially macros: https://en.cppreference.com/w/c/header/math.html

We should be fine with just using those

davebayer: All of these can be defined as macros, see the docs

miscco: fine by me, doesnt hurt to be safe

PR #7418, opened 2026-01-29
Refactor `cub::ThreadStore`

bernhardmgruber:


cub/cub/thread/thread_store.cuh
```diff
@@ -326,14 +312,46 @@ ThreadStore(T* ptr, T val, detail::constant_t<MODIFIER> /*modifier*/, ::cuda::st
...
- * ThreadStore definition for generic modifiers
- */
 template <CacheStoreModifier MODIFIER, typename OutputIteratorT, typename T>
 _CCCL_DEVICE _CCCL_FORCEINLINE void ThreadStore(OutputIteratorT itr, T val)
 {
-  ThreadStore(
-    itr, val, detail::constant_v<MODIFIER>, detail::bool_constant_v<::cuda::std::is_pointer_v<OutputIteratorT>>);
+  if constexpr (!::cuda::std::is_pointer_v<OutputIteratorT> || MODIFIER == STORE_DEFAULT)
+  {
+    *itr = val;
```
miscco, line 321: It feels like this is kind of dangerous for some of the modifiers, should we at least warn here?

bernhardmgruber: I am following the original design. If we don't have a pointer all we can do is just store to the iterator and ignore the modifier. We can only apply the cache modifier if we have a pointer.

miscco: Can you add a todo to support `contiguous_iterator`

bernhardmgruber: Implemented it.

cub/cub/thread/thread_store.cuh
```diff
@@ -2,10 +2,8 @@
 // SPDX-FileCopyrightText: Copyright (c) 2011-2018, NVIDIA CORPORATION. All rights reserved.
```
srinivasyadav18: nit: 
```suggestion
// SPDX-FileCopyrightText: Copyright (c) 2011-2026, NVIDIA CORPORATION. All rights reserved.
```

cub/cub/thread/thread_store.cuh
```diff
@@ -128,11 +121,21 @@ struct iterate_thread_store<MAX, MAX>
...
   {}
 };
+
+template <CacheStoreModifier MODIFIER, typename T, size_t... Is>
+_CCCL_DEVICE _CCCL_FORCEINLINE void store_helper(T* ptr, T* vals, ::cuda::std::index_sequence<Is...>)
+{
+  (ThreadStore<MODIFIER>(ptr + Is, vals[Is]), ...);
+}
+
+template <CacheStoreModifier MODIFIER, typename T, size_t... Is>
```
srinivasyadav18: `MODIFIER` template parameter is not used. Do we need it ?

```suggestion
template <typename T, size_t... Is>
```

PR #7419, opened 2026-01-29
Refactor `cub::ThreadLoad`

bernhardmgruber:


cub/cub/thread/thread_load.cuh
```diff
@@ -342,10 +320,35 @@ ThreadLoad(T const* ptr, detail::constant_t<MODIFIER> /*modifier*/, ::cuda::std:
...
 template <CacheLoadModifier MODIFIER, typename RandomAccessIterator>
-_CCCL_DEVICE _CCCL_FORCEINLINE cub::detail::it_value_t<RandomAccessIterator> ThreadLoad(RandomAccessIterator itr)
+_CCCL_DEVICE _CCCL_FORCEINLINE detail::it_value_t<RandomAccessIterator> ThreadLoad(RandomAccessIterator itr)
 {
-  return ThreadLoad(
-    itr, detail::constant_v<MODIFIER>, detail::bool_constant_v<::cuda::std::is_pointer_v<RandomAccessIterator>>);
+  using T = detail::it_value_t<RandomAccessIterator>;
+  if constexpr (!::cuda::std::is_pointer_v<RandomAccessIterator> || MODIFIER == LOAD_DEFAULT)
+  {
+    return *itr;
```
miscco, line 329: Ditto should we warn that we ignore some modifiers?

bernhardmgruber: The previous implementation also just ignored it, so i would keep this behavior. I think the intent is to use `ThreadLoad<Tuning::LOAD_MODIFIER>(iter, ...)` and just hoping the implementation does the best effort.

PR #7439, opened 2026-01-30
Specialize `cuda::std::pointer_traits` for thrust smart pointers

miscco:
This replaces `thrust::raw_pointer_cast` with the standard conforming `cuda::std::to_address`

We enable usage of that by specializing `cuda::std::pointer_traits` for the thrust smart pointers.

NOTE: I did not yet deprecate `thrust::raw_pointer_cast` because we should give users, e.g rapids a way to update and its not yet available. I plan on deprecating it in the 3.4 time frame


thrust/thrust/detail/pointer.h
```diff
@@ -384,7 +403,126 @@ class pointer : public detail::pointer_base<Element, Tag, Reference, Derived>::t
...
+  using type = Ptr<T, Tag, ::cuda::std::add_lvalue_reference_t<T>, DerivedPtr<T, DerivedPtrTail...>>;
+};
+} // namespace detail
+
 /*! \} // memory_management
  */
 
 THRUST_NAMESPACE_END
+
+_CCCL_BEGIN_NAMESPACE_CUDA_STD
```
davebayer, line 397: This namespace shoulnd't be used outside of libcu++, we've already have some problems using it in cub

thrust/thrust/detail/pointer.h
```diff
@@ -397,7 +403,126 @@ class pointer : public detail::pointer_base<Element, Tag, Reference, Derived>::t
...
+  using difference_type = ptrdiff_t;
+
+  template <typename U>
+  struct rebind
+  {
+    using other = typename THRUST_NS_QUALIFIER::detail::thrust_pointer_rebind<pointer, U>::type;
+  };
+
+  // Backwards compatability with thrust::detail::pointer_traits
+  using raw_pointer = typename pointer::raw_pointer;
```
bernhardmgruber: Q: This is only used internally, right? So we can eventually remove this, right?

thrust/thrust/detail/pointer.h
```diff
@@ -397,7 +403,126 @@ class pointer : public detail::pointer_base<Element, Tag, Reference, Derived>::t
...
+  [[nodiscard]] _CCCL_API static constexpr raw_pointer to_address(const pointer iter) noexcept
+  {
+    return iter.get();
+  }
+
+  //! For backwards compatability with old pointer traits
+  [[nodiscard]] _CCCL_API static constexpr raw_pointer get(const pointer iter) noexcept
+  {
+    return iter.get();
+  }
```
bernhardmgruber: Why can't we replace all occurrences by `to_address`?

thrust/thrust/detail/pointer.h
miscco: yeah i hope so

thrust/thrust/detail/pointer.h
```diff
@@ -388,3 +393,38 @@ class pointer : public detail::pointer_base<Element, Tag, Reference, Derived>::t
  */
 
 THRUST_NAMESPACE_END
+
+// Specialize pointer traits for everything that has the raw_pointer alias
+template <typename Pointer>
+struct ::cuda::std::pointer_traits<Pointer, ::cuda::std::void_t<typename Pointer::raw_pointer>>
```
Jacobfaib: Consider specializing `std::pointer_traits` as well here?

thrust/thrust/device_ptr.h
```diff
@@ -221,3 +221,36 @@ THRUST_NAMESPACE_END
 
 #include <thrust/detail/device_ptr.inl>
 #include <thrust/detail/raw_pointer_cast.h>
+
+// Specialize pointer traits for everything that has the raw_pointer alias
+template <class T>
+struct ::cuda::std::pointer_traits<THRUST_NS_QUALIFIER::device_ptr<T>>
```
Jacobfaib: Same here for `std::pointer_traits`?

miscco: I definitely do not want to do that because that is a device pointer and the `std` traits have no dealing with that

thrust/thrust/detail/pointer.h
miscco: I am a bit on the fence, because the standard does not know about execution spaces, but with that I am fine

Jacobfaib: I think it would be useful for downstream users. In my experience it is common for downstream projects to use thrust fancy pointers, but not _yet_ `cuda::std::` utilities. I myself have (carefully) rolled my own specializations for `std::pointer_traits` for thrust fancy pointers, simply to have generic code that needs to introspect pointers Just Work(TM).

thrust/thrust/device_ptr.h
Jacobfaib: I'm not sure I follow, what would be the danger here?

bernhardmgruber:
I have the feeling this PR massively grew in size after I approved. Why was this necessary and why can't we ship the changes added later in a separate PR?

miscco:
Actually it is much smaller. 

It introduces `pointer_traits` specializations for the contiguous thrust iterators.

It then replaces `raw_pointer_cast` with `cuda::std::to_address` because that is the standard interface to retrieve a pointer from a smart pointer

The issue is that we use `raw_pointer_cast` a lot

thrust/thrust/detail/pointer.h
miscco: Doing it now

thrust/thrust/device_ptr.h
miscco: device only sequences should work with cuda::std:: types and not std::types

miscco:
Not to reviewers. all relevant product code changes are in https://github.com/NVIDIA/cccl/pull/7439/changes/f87d697a5df7fce80a6b298054dd50b6b4e0aa08

The rest is only search & replace

thrust/testing/cuda/partition.cu
```diff
@@ -3,7 +3,7 @@
 #include <thrust/iterator/discard_iterator.h>
 #include <thrust/partition.h>
 
-#include "thrust/detail/raw_pointer_cast.h"
+#include "cuda/std/memory"
```
bernhardmgruber:
```suggestion
#include <cuda/std/memory>
```

miscco:
> If there are no uses of `thrust::raw_pointer_cast` remaining, can we deprecate it?

I want to avoid that for now. `raw_pointer_cast` is heavily used by downstream projects and that would just break our CI for rapids and probably MatX until they upgrade to CCCL 3.4

I would say we target deprecation for the 3.5 timeframe

miscco:
I am seeing issues with our rapids CI that I need to investigate

PR #7441, opened 2026-01-30
Rewrite iterators to not depend on Numba

shwina:
## Description

Closes https://github.com/NVIDIA/cccl/issues/7353

This PR is the next major step towards dropping Numba as a required dependency of `cuda.compute`. Previously, all iterators were implemented in pure Python, with numba being used to JIT compile the definitions of the `advance` and `dereference` methods of all iterators. This PR reimplements all methods in C++, and uses NVRTC to JIT compile those into LTOIR. This does not affect the API/usage in any way.

The end effect of this PR is that numba is only ever imported in `_jit.py` (and the associated `_odr_helpers.py`). The only function exported by `_jit.py` is `to_jit_op_adapter`. This makes the library importable without numba, and also makes it trivial to raise if numba is not available in the future.

The important changes are:

* Some tooling to generate and JIT compile C++ with NVRTC (https://github.com/NVIDIA/cccl/pull/7441/changes/d16f6a813650f03a194e2e873fbe588df85e0218)
* Reimplementation of all iterator methods in C++ (https://github.com/NVIDIA/cccl/pull/7441/changes/eb9b0f2b228be17b4a91071ae973abbb988a75f6)

## Performance impact

The performance impact is neutral to slightly positive, with most benchmarks within noise. Attached here is a comparison of the pytest benchmarks between this branch ("byoc") and main:

[comparison.txt](https://github.com/user-attachments/files/25162551/comparison.txt)

And an AI-generated summary::

 ### Representative Benchmark Results

   #### 🚀 Iterator Benchmarks - Clear Improvements

   **bench_binary_transform_iterator** (time in μs, lower is better):
   ```
   Size      main                  byoc      Speedup
   ------------------------------------------------------
   10K       18.13                 15.57     1.16x (16% faster)
   100K      18.49                 15.22     1.21x (21% faster)
   1M        19.79                 16.30     1.21x (21% faster)
   ```

   **bench_unary_transform_iterator**:
   ```
   Size      main                  byoc      Speedup
   ------------------------------------------------------
   10K       15.51                 13.62     1.14x (14% faster)
   100K      15.38                 13.67     1.13x (13% faster)
   1M        16.52                 14.82     1.11x (11% faster)
   ```

   **bench_select_iterator**:
   ```
   Size      main                  byoc      Speedup
   ------------------------------------------------------
   10K       30.92                 29.17     1.06x (6% faster)
   100K      32.05                 29.68     1.08x (8% faster)
   1M        34.19                 32.09     1.07x (7% faster)
   ```

   #### ➡️  Pointer Benchmarks - Neutral Performance

   **bench_binary_transform_pointer**:
   ```
   Size      main                  byoc      Speedup
   ------------------------------------------------------
   10K       14.04                 13.50     1.04x (neutral)
   100K      13.81                 13.71     1.01x (neutral)
   1M        17.20                 17.23     1.00x (neutral)
   ```

   #### 📦 Struct Benchmarks - Neutral with Slight Regression

   **bench_reduce_struct**:
   ```
   Size      main                  byoc      Speedup
   ------------------------------------------------------
   10K       40.67                 41.08     0.99x (1% slower)
   100K      40.58                 41.24     0.98x (2% slower)
   1M        41.61                 42.95     0.97x (3% slower)
   ```

   **bench_upper_bound** (handles structs well):
   ```
   Size      main                  byoc      Speedup
   ------------------------------------------------------
   10K       23.96                 23.23     1.03x (neutral)
   100K      30.42                 27.87     1.09x (9% faster)
   1M        123.35                114.74    1.08x (8% faster)
   ```

   ### Key Takeaways

   ✅ **Iterator operations**: Consistent 6-21% speedups across the board
   ✅ **Pointer operations**: Performance maintained (within 1-4% variation)
   ✅ **Struct operations**: Mostly neutral, one 3% regression in reduce_struct
   ✅ **Scaling**: Benefits increase with problem size for search operations

## Checklist

### DIFF
--- a/python/cuda_cccl/tests/compute/test_scan.py
+++ b/python/cuda_cccl/tests/compute/test_scan.py
@@ -339,6 +339,7 @@ def test_reverse_input_iterator(monkeypatch):
     np.testing.assert_equal(d_output.get(), expected)
 
 
+@pytest.mark.no_verify_sass(reason="LDL/STL instructions emitted for this test.")
 def test_reverse_output_iterator():
     def add_op(a, b):
         return a + b
--- a/python/cuda_cccl/cuda/compute/iterators/_reverse.py
+++ b/python/cuda_cccl/cuda/compute/iterators/_reverse.py
@@ -10,7 +10,7 @@ from textwrap import dedent
 
 from .._bindings import Op, OpKind
 from .._cpp_compile import compile_cpp_to_ltoir
-from .._utils.protocols import is_device_array
+from .._utils.protocols import get_size, is_device_array
 from ._base import IteratorBase
 from ._common import CUDA_PREAMBLE, ensure_iterator
 
@@ -38,7 +38,7 @@ class ReverseIterator(IteratorBase):
             # TODO: this is probably incorrect behaviour. In C++, initializing
             # with a pointer to the end of the array is left to be done explicitly
             # by the user.
-            self._underlying = ensure_iterator(underlying) + (len(underlying) - 1)
+            self._underlying = ensure_iterator(underlying) + (get_size(underlying) - 1)
         else:
             self._underlying = ensure_iterator(underlying)
 


c/parallel/include/cccl/c/types.h
```diff
@@ -103,6 +103,11 @@ typedef struct cccl_op_t
   size_t size;
   size_t alignment;
   void* state;
+  // Extra LTOIR modules to link with this op (for composite ops that reference
+  // other compiled functions). Each entry is an LTOIR module.
```
shwina: Unnecessary comment.

python/cuda_cccl/cuda/compute/iterators/_base.py
```diff
@@ -0,0 +1,229 @@
+# Copyright (c) 2025, NVIDIA CORPORATION & AFFILIATES. ALL RIGHTS RESERVED.
```
shwina: Copyrights, everywhere.

python/cuda_cccl/cuda/compute/_utils/protocols.py
```diff
@@ -62,6 +62,21 @@ def get_shape(arr: DeviceArrayLike) -> Tuple[int]:
         return arr.__cuda_array_interface__["shape"]
 
 
+def get_size(arr: DeviceArrayLike) -> int:
+    """Get the total number of elements in an array."""
+    # Try fast path via .size attribute (works for NumPy, CuPy, PyTorch)
```
shwina: Not "NumPy"

python/cuda_cccl/cuda/compute/iterators/_base.py
```diff
@@ -0,0 +1,309 @@
...
+from typing import Hashable
+
+from .._bindings import Iterator, IteratorKind, IteratorState, Op, OpKind
+from .._caching import cache_with_registered_key_functions
+from ..types import TypeDescriptor
+
+
+class IteratorBase:
+    """
+    Base class for iterators that use C++ codegen.
```
shwina: Improve docstring.

python/cuda_cccl/cuda/compute/algorithms/_sort/_merge_sort.py
```diff
@@ -32,8 +31,8 @@ class _MergeSort:
 
     def __init__(
         self,
-        d_in_keys: DeviceArrayLike | IteratorBase,
-        d_in_items: DeviceArrayLike | IteratorBase | None,
+        d_in_keys: IteratorLike,
```
shwina: Call out in the docs what `IteratorLike` means.

python/cuda_cccl/cuda/compute/iterators/_base.py
```diff
@@ -0,0 +1,238 @@
...
+    Subclasses must implement:
+    - _make_advance_op() -> Op
+    - _make_input_deref_op() -> Op | None
+    - _make_output_deref_op() -> Op | None
+
+    These private methods create Op objects with compiled LTOIR.
+    Subclasses can use any compilation approach:
+    - C++ source via compile_cpp_source_to_ltoir() (most common)
+    - Pre-compiled LTOIR bytes
+    - Other compilation backends
```
shwina: Remove this bit.

python/cuda_cccl/cuda/compute/iterators/_codegen_utils.py
```diff
@@ -0,0 +1,302 @@
...
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Optional
+
+if TYPE_CHECKING:
+    from ._base import IteratorBase
+
+
+# ============================================================================
```
shwina: Not needed.

python/cuda_cccl/cuda/compute/iterators/_cache_modified.py
```diff
@@ -0,0 +1,162 @@
...
+        _, intrinsic = _CACHE_MODIFIERS[self._modifier]
+
+        # For types that don't support direct __ldcs, we need to use inline PTX
+        # __ldcs works for 4-byte and 8-byte types
+        if self._value_type.size in (4, 8):
+            body = dedent(f"""
+                auto* ptr = *static_cast<{cpp_type}**>(state);
+                *static_cast<{cpp_type}*>(result) = {intrinsic}(ptr);
+            """).strip()
+        else:
```
shwina: Does CUB do this?

python/cuda_cccl/cuda/compute/iterators/_base.py
```diff
@@ -0,0 +1,280 @@
...
+from typing import Hashable
+
+from .._bindings import Iterator, IteratorKind, IteratorState, Op
+from .._caching import cache_with_registered_key_functions
+from ..types import TypeDescriptor
+
+
+class IteratorBase:
+    """
+    Base class for all iterators.
```
shwina: Describe what an iterator is here instead.

c/parallel/src/nvrtc/command_list.h
```diff
@@ -326,11 +329,29 @@ struct nvrtc2_top_level
       }
     }
 
-    // Add all LTO-IR items to the linker
+    // Add all LTO-IR items to the linker (deduplicate identical blobs)
```
shwina, line 332: This bit is all vibe coded

c/parallel/include/cccl/c/types.h
shwina, line 1: Let's run compile benchmarks to make sure the dedup isn't hurting compile time too much.

python/cuda_cccl/cuda/compute/iterators/_base.py
```diff
@@ -0,0 +1,278 @@
...
+        """
+        raise NotImplementedError
+
+
+def _deterministic_suffix(kind: Hashable) -> str:
+    kind_str = str(kind)
+    return hashlib.sha256(kind_str.encode()).hexdigest()[:16]
+
+
+def compose_iterator_states(
```
shwina, line 221: Can we push this down to C++ or Cython?

python/cuda_cccl/cuda/compute/iterators/_counting.py
```diff
@@ -0,0 +1,113 @@
...
+
+import numpy as np
+
+from .._bindings import Op, OpKind
+from .._cpp_compile import compile_cpp_to_ltoir, cpp_type_from_descriptor
+from ..types import from_numpy_dtype
+from ._base import IteratorBase
+
+CUDA_PREAMBLE = """#include <cuda/std/cstdint>
+#include <cuda_fp16.h>
```
shwina: Comment fp16 out (and then all the includes). Does anything break?

python/cuda_cccl/cuda/compute/_jit.py
```diff
@@ -713,23 +509,51 @@ def _compile_op(op, input_types, output_type=None):
...
+    which allows proper deduplication during linking.
+    """
+    from ._caching import CachableFunction
+
+    cachable_op = CachableFunction(op)
+    return _compile_op_impl(cachable_op, tuple(input_types), output_type)
+
 
-    __slots__ = ["_func", "_cachable"]
+class _JitOp(OpAdapter):
```
NaderAlAwar: Question: why do we need another class in the inheritance layer? Can the methods implemented here be folded into `_StatelessOp`?

python/cuda_cccl/cuda/compute/types.py
```diff
@@ -216,6 +228,50 @@ def to_ctypes_type(td: TypeDescriptor):
...
+def _annotation_to_type_descriptor(annotation):
+    """
+    Convert a type annotation to a TypeDescriptor.
+
+    Handles:
+    - TypeDescriptor: returns as-is
+    - gpu_struct classes: returns their _type_descriptor
+    - numpy dtypes/types: converts via from_numpy_dtype
+    """
+    from .struct import _is_struct_type
```
NaderAlAwar, line 240: Suggestion: move this import to the top of the file if possible

python/cuda_cccl/cuda/compute/_cpp_compile.py
```diff
@@ -0,0 +1,120 @@
+# Copyright (c) 2025, NVIDIA CORPORATION & AFFILIATES. ALL RIGHTS RESERVED.
```
NaderAlAwar: Nit: should be 2026

python/cuda_cccl/cuda/compute/_cpp_compile.py
```diff
@@ -0,0 +1,120 @@
...
+from __future__ import annotations
+
+import functools
+
+from cuda.cccl import get_include_paths
+
+
+def _get_arch_string() -> str:
+    """Get the compute capability string for the current device."""
+    from cuda.core import Device
```
NaderAlAwar: Suggestion: move this and all cuda.core imports to the top of the file

python/cuda_cccl/cuda/compute/_cpp_compile.py
```diff
@@ -0,0 +1,120 @@
...
+
+    Example:
+        source = '''
+        extern "C" __device__ void my_add(void* a, void* b, void* result) {
+            *static_cast<int*>(result) = *static_cast<int*>(a) + *static_cast<int*>(b);
+        }
+        '''
+        ltoir = compile_cpp_to_ltoir(source)
+    """
+    from cuda.core import Program, ProgramOptions
```
NaderAlAwar: same as above

python/cuda_cccl/cuda/compute/_cpp_compile.py
```diff
@@ -0,0 +1,120 @@
...
+import functools
+
+from cuda.cccl import get_include_paths
+
+
+def _get_arch_string() -> str:
+    """Get the compute capability string for the current device."""
+    from cuda.core import Device
+
+    device = Device()
```
NaderAlAwar, line 22: Question: what happens on a GPU with multiple devices? I think this is something we have not considered in cuda.compute at all

python/cuda_cccl/cuda/compute/_cpp_compile.py
```diff
@@ -0,0 +1,120 @@
...
+
+    # Get include paths
+    include_paths = _get_include_paths()
+
+    # Configure compilation options for LTO
+    opts = ProgramOptions(
+        arch=arch,
+        relocatable_device_code=True,
+        link_time_optimization=True,
+        std="c++17",
```
NaderAlAwar: Suggestion: should just use `c++20`. More recent and less likely to break in the near future

python/cuda_cccl/cuda/compute/_cpp_compile.py
```diff
@@ -0,0 +1,120 @@
...
+    return result.code
+
+
+def cpp_type_from_descriptor(type_desc) -> str | None:
+    """
+    Get the C++ type name from a TypeDescriptor.
+    """
+    # Careful!!!
+    # For efficiency, this function returns None for non-primitive types.
+    # Callers must take care to handle that case appropriately.
```
NaderAlAwar: Suggestion: move this comment into the docstring

python/cuda_cccl/cuda/compute/_cpp_compile.py
```diff
@@ -0,0 +1,120 @@
...
+    from ._bindings import TypeEnum
```
NaderAlAwar: Also move this to the top of the file if possible

python/cuda_cccl/cuda/compute/_cpp_compile.py
```diff
@@ -0,0 +1,120 @@
...
+        TypeEnum.FLOAT32: "float",
+        TypeEnum.FLOAT64: "double",
+        TypeEnum.BOOLEAN: "bool",
+    }
+
+    if type_desc.info.typenum in type_map:
+        return type_map[type_desc.info.typenum]
+
+    # STORAGE types have no type name
+    return None
```
NaderAlAwar: Suggestion: replace with `return ... .get(..., None)`

python/cuda_cccl/cuda/compute/iterators/_base.py
NaderAlAwar: yeah switch to 2026

python/cuda_cccl/cuda/compute/iterators/_base.py
```diff
@@ -0,0 +1,278 @@
...
+    # - children property to return tuple of child iterators for dependency tracking
+    #
+    # Examples of such "compound" iterators include TransformIterator,
+    # PermutationIterator, ReverseIterator and ZipIterator.
+    #
+    # The base class provides public cached accessors:
+    #
+    # - get_advance_op() -> Op (cached)
+    # - get_input_deref_op() -> Op | None (cached)
+    # - get_output_deref_op() -> Op | None (cached)
```
NaderAlAwar, line 47: Suggestion: move this comment to docstring

python/cuda_cccl/cuda/compute/iterators/_cache_modified.py
```diff
@@ -0,0 +1,173 @@
+# Copyright (c) 2025, NVIDIA CORPORATION & AFFILIATES. ALL RIGHTS RESERVED.
```
NaderAlAwar: 2026 in all iterator files

python/cuda_cccl/cuda/compute/iterators/_cache_modified.py
```diff
@@ -0,0 +1,173 @@
...
+from .._cpp_compile import compile_cpp_to_ltoir, cpp_type_from_descriptor
+from .._utils.protocols import get_data_pointer, get_dtype
+from ..types import from_numpy_dtype
+from ._base import IteratorBase
+
+CUDA_PREAMBLE = """#include <cuda/std/cstdint>
+#include <cuda_fp16.h>
+#include <cuda/std/cstring>
+using namespace cuda::std;
+"""
```
NaderAlAwar: Suggestion: move this to common file to reuse across iterators

python/cuda_cccl/cuda/compute/iterators/_cache_modified.py
```diff
@@ -0,0 +1,173 @@
...
+        # These correspond to PTX instructions: ld.global.{modifier}.b{8,16,32,64,128}
+        if value_type.size not in (1, 2, 4, 8, 16):
+            raise ValueError(
+                f"CacheModifiedInputIterator only supports types of size 1, 2, 4, 8, or 16 bytes. "
+                f"Got type with size {value_type.size} bytes. "
+                f"This matches PTX cache-modified load instruction limitations."
+            )
+
+        # State is just the pointer (8 bytes on 64-bit)
+        import struct
```
NaderAlAwar: Suggestion: if possible move to file level

python/cuda_cccl/cuda/compute/iterators/_cache_modified.py
```diff
@@ -0,0 +1,173 @@
...
+
+    def _make_output_deref_op(self) -> Op | None:
+        # Cache-modified iterator is input-only
+        return None
+
+    def __add__(self, offset: int) -> "CacheModifiedInputIterator":
+        """Advance the iterator by offset elements."""
+        import struct
+
+        from .._utils.protocols import get_dtype
```
NaderAlAwar: Suggestion: move to file level if possible

python/cuda_cccl/cuda/compute/iterators/_cache_modified.py
```diff
@@ -0,0 +1,173 @@
...
+
+        dtype = get_dtype(self._array)
+        offset_ptr = self._ptr + offset * dtype.itemsize
+
+        # Create a new instance with the offset pointer
+        clone = CacheModifiedInputIterator.__new__(CacheModifiedInputIterator)
```
NaderAlAwar: Question: why not call the constructor directly?

python/cuda_cccl/cuda/compute/iterators/_discard.py
```diff
@@ -0,0 +1,114 @@
...
+            extra_ltoirs=[],
+        )
+
+    def __add__(self, offset: int) -> "DiscardIterator":
+        """Return a new DiscardIterator (stateless, so position doesn't matter)."""
+        # DiscardIterator is stateless - advance is a no-op
+        # If reference_iterator supports advance, advance it; otherwise use as-is
+        ref = self._reference_iterator
+        if isinstance(ref, IteratorBase) and hasattr(ref, "__add__"):
+            ref = ref + offset  # type: ignore[operator]
```
NaderAlAwar: Question: what is the point of advancing the ref iterator?

python/cuda_cccl/cuda/compute/iterators/_permutation.py
```diff
@@ -0,0 +1,233 @@
...
+from ._base import IteratorBase, compose_iterator_states
+
+CUDA_PREAMBLE = """#include <cuda/std/cstdint>
+#include <cuda_fp16.h>
+#include <cuda/std/cstring>
+using namespace cuda::std;
+"""
+
+if TYPE_CHECKING:
+    pass
```
NaderAlAwar: Suggestion: looks like this can be deleted

python/cuda_cccl/cuda/compute/iterators/_permutation.py
```diff
@@ -0,0 +1,233 @@
...
+
+
+def _ensure_iterator(obj):
+    """Wrap array in PointerIterator if needed."""
+    from ._pointer import PointerIterator
+
+    if isinstance(obj, IteratorBase):
+        return obj
+    if hasattr(obj, "__cuda_array_interface__"):
```
NaderAlAwar: Important: I believe we have a function that does this check

python/cuda_cccl/cuda/compute/iterators/_pointer.py
```diff
@@ -0,0 +1,205 @@
...
+from ._base import IteratorBase
+
+CUDA_PREAMBLE = """#include <cuda/std/cstdint>
+#include <cuda_fp16.h>
+#include <cuda/std/cstring>
+using namespace cuda::std;
+"""
+
+if TYPE_CHECKING:
+    pass
```
NaderAlAwar: Suggestion: delete this

python/cuda_cccl/cuda/compute/iterators/_pointer.py
```diff
@@ -0,0 +1,205 @@
...
+
+
+class PointerIterator(IteratorBase):
```
NaderAlAwar, line 21: Question: why is this class needed?

python/cuda_cccl/cuda/compute/iterators/_pointer.py
```diff
@@ -0,0 +1,205 @@
...
+        dtype = get_dtype(self._array)
+        offset_ptr = self._current_pointer() + offset * dtype.itemsize
+        return self._clone_with_pointer(offset_ptr)
+
+    def _current_pointer(self) -> int:
+        return int.from_bytes(self._state_bytes, sys.byteorder, signed=False)
+
+    def _clone_with_pointer(self, pointer_value: int):
+        """Clone this iterator with a different pointer value."""
+        clone = PointerIterator.__new__(PointerIterator)
```
NaderAlAwar: Important: If we cannot call the constructor here, lets create an init method in PointerIterator, or a make_* factory to avoid copying the logic in the constructor

python/cuda_cccl/cuda/compute/iterators/_zip.py
```diff
@@ -0,0 +1,227 @@
...
+
+def _ensure_iterator(obj):
+    """Wrap array in PointerIterator if needed."""
+    from ._pointer import PointerIterator
+
+    if isinstance(obj, IteratorBase):
+        return obj
+    if hasattr(obj, "__cuda_array_interface__"):
+        return PointerIterator(obj)
+    raise TypeError("ZipIterator requires iterators or device arrays")
```
NaderAlAwar: Important: this is duplicated, lets move it to a common file

python/cuda_cccl/cuda/compute/_cpp_compile.py
shwina: > a GPU with multiple devices

Could you clarify when this is the case?

python/cuda_cccl/cuda/compute/iterators/_base.py
shwina: `IteratorBase` shows up as a public type because it's in our type annotations. I don't think we (at this time) want to make public the details of subclassing it, so I kept it as code comments for now.

python/cuda_cccl/cuda/compute/iterators/_cache_modified.py
shwina: In general, the constructor may potentially do expensive pre-processing of its arguments that we can circumvent by creating an empty object and patching in its members. But in this situation (and the other place where we do `__new__`) it probably doesn't matter. So I changed to using the constructor instead.

python/cuda_cccl/cuda/compute/iterators/_discard.py
shwina: No point. I deleted it.

python/cuda_cccl/cuda/compute/iterators/_pointer.py
shwina: It's used by compound iterators (like zip, transform, reverse) as the inner iterator when it is a device array. FWIW, that's not new in this PR, we already do it ([example](https://github.com/NVIDIA/cccl/blob/99a51dd4a071d309cf44fd2a1678c426f84208a3/python/cuda_cccl/cuda/compute/iterators/_zip_iterator.py#L187-L188))

python/cuda_cccl/cuda/compute/iterators/_zip.py
shwina: Done.

python/cuda_cccl/cuda/compute/iterators/_pointer.py
shwina: I changed to call the constructor.

python/cuda_cccl/cuda/compute/types.py
shwina: We can't unfortunately, as we need `types.py` imported in `struct.py`.

python/cuda_cccl/cuda/compute/_jit.py
shwina: I agree. I removed it. The only common functionality was `get_return_type()` but it's I think it's fine to duplicate that as it's a one liner.

python/cuda_cccl/cuda/compute/iterators/_base.py
shwina: I left this as is for now. As we talked about earlier, this is host logic so we definitely can't JIT compile this with NVRTC. It's totally possible to push this down to Cython and ship as part of a pre-built extension, but we can revisit that idea if it ever becomes a problem. Currently the benchmarks show no compile time regressions.

c/parallel/include/cccl/c/types.h
shwina: Done (see PR description).

python/cuda_cccl/cuda/compute/iterators/_base.py
shwina: ✅

python/cuda_cccl/tests/compute/test_scan.py
```diff
@@ -339,6 +339,7 @@ def add_op(a, b):
     np.testing.assert_equal(d_output.get(), expected)
 
 
+@pytest.mark.no_verify_sass(reason="LDL/STL instructions emitted for this test.")
```
shwina, line 342: Note, this test is newly found to fail the LDL/STL check for CUDA 13.1 on H100.

I suspect this is due to https://github.com/NVIDIA/cccl/pull/7441/changes#diff-26033a79ec965190c915b3fe616dd6cba8eac2a7d556fb51f0f09d80e46ff35dR102, but we should follow up with an investigation.

shwina: https://github.com/NVIDIA/cccl/issues/7570

python/cuda_cccl/cuda/compute/iterators/_permutation.py
```diff
@@ -0,0 +1,213 @@
...
+
+
+class PermutationIterator(IteratorBase):
+    """
+    Iterator that accesses values through an index mapping.
+
+    At position i, yields values[indices[i]]. The indices iterator is advanced
+    normally, and the values iterator is accessed via random access using the
+    index value.
+    """
```
NaderAlAwar, line 24: Suggestion: for consistency with other iterator docstrings, add a link to https://nvidia.github.io/cccl/thrust/api/classthrust_1_1permutation__iterator.html and include example code (if we have it)

python/cuda_cccl/cuda/compute/_cpp_compile.py
NaderAlAwar: was a typo, I meant a machine with multiple devices. I think they would all have the same compute capability so it probably wouldn't matter here

python/cuda_cccl/cuda/compute/iterators/_permutation.py
shwina: I'll tack this on in the third PR in this series to save another CI cycle.

PR #7446, opened 2026-01-30
cub: change {Lower,Upper}Bound to accept iterator and number of elements.

### LINKED ISSUE #7445: Change cub::DeviceFind::{Lower,Upper}Bound to use iterator + count instead of iterator pairs

griwes:
## Description

Change the interface of `DeviceFind::{Lower,Upper}Bound` from accepting pairs of iterators to accepting a begin iterator and item count, to be more consistent with most of the other interfaces in CUB.

Resolves #7445.

## Checklist


cub/cub/detail/binary_search_helpers.cuh
```diff
@@ -15,31 +15,35 @@
...
+    using DifferenceT = ::cuda::std::iter_difference_t<RangeIteratorT>;
+    const auto last   = first + static_cast<DifferenceT>(num_items);
+
     ::cuda::std::get<1>(args) = Mode::Invoke(first, last, ::cuda::std::get<0>(args), op);
   }
 };
 
-template <typename Mode, typename RangeIteratorT, typename CompareOpT>
-_CCCL_HOST_DEVICE auto make_comp_wrapper(RangeIteratorT first, RangeIteratorT last, CompareOpT comp)
+template <typename Mode, typename RangeIteratorT, typename RangeNumItemsT, typename CompareOpT>
```
miscco, line 43: Blasphemeous question, why not have both interfaces and store `distance(first, last)` in the iterator based entry point?

cub/cub/device/device_find.cuh
```diff
@@ -167,48 +173,65 @@ struct DeviceFind
...
+    return DeviceFor::ForEachN(
       d_temp_storage,
       temp_storage_bytes,
-      ::cuda::make_zip_iterator(values_first, output),
-      ::cuda::make_zip_iterator(values_last, output + ::cuda::std::distance(values_first, values_last)),
-      detail::find::make_comp_wrapper<detail::find::lower_bound>(first, last, comp),
+      ::cuda::make_zip_iterator(d_values, d_output),
+      static_cast<ValuesOffsetT>(values_num_items),
+      detail::find::make_comp_wrapper<detail::find::lower_bound>(
+        d_range, static_cast<RangeOffsetT>(range_num_items), comp),
```
miscco, line 207: What if `values_num_items` is smaller than `range_num_items`?

Also we are using `values_num_items` in one place and `range_num_items` in the other

griwes: Yes, those are two separate ranges, we're doing for each over values, and letting each thread know about the whole range of input elements.

cub/cub/detail/binary_search_helpers.cuh
griwes: We require RAI, so this *is* straightforward for the user to do with just `last - first`.

PR #7447, opened 2026-02-02
Fix non default constructible input types test for cub::FindIf

gonidelis:
Fixes  #7337 test by

- Making `NonDefaultConstructible` actually non default constructible
- Making test input iterator raw so that vectorized path is taken `ConsumeTile(..., attempt_vectorization=true)` so test covers the faulty code.

The source code fix was correct. It's just that the added test was not covering it.

### DIFF
--- a/cub/test/catch2_test_device_find_if.cu
+++ b/cub/test/catch2_test_device_find_if.cu
@@ -237,7 +237,7 @@ struct index_to_value
   }
 };
 
-static_assert(cuda::std::is_default_constructible_v<NotDefaultConstructible> == false,
+static_assert(!cuda::std::is_default_constructible_v<NotDefaultConstructible>,
               "NotDefaultConstructible should not be default constructible");
 
 C2H_TEST("Device find_if works with non default constructible types", "[device][find_if]")


cub/test/catch2_test_device_find_if.cu
```diff
@@ -228,17 +228,12 @@ struct NotDefaultConstructible
...
 };
 
-struct converter
-{
-  __host__ __device__ constexpr NotDefaultConstructible operator()(const int val) const noexcept
-  {
-    return NotDefaultConstructible{val};
-  }
-};
+static_assert(cuda::std::is_default_constructible<NotDefaultConstructible>::value == false,
```
bernhardmgruber: Important: use the variable templates for traits
```suggestion
static_assert(cuda::std::is_default_constructible_v<NotDefaultConstructible> == false,
```

cub/test/catch2_test_device_find_if.cu
```diff
@@ -251,16 +246,26 @@ C2H_TEST("Device find_if works with non default constructible types", "[device][
...
-    find_if(c_it, thrust::raw_pointer_cast(out_result.data()), predicate, num_items);
-    REQUIRE(val_to_find == out_result[0]);
+    // fill with arbitrary values dont use c2h gen because NotDefaultConstructible is not default constructible
+    for (offset_t i = 0; i < num_items - 1; ++i)
+    {
+      d_vec[i] = NotDefaultConstructible{static_cast<int>(i)};
+    }
+
+    // set the value to find at the end
+    d_vec[num_items - 1] = NotDefaultConstructible{val_to_find};
```
bernhardmgruber: Important: This is equivalent to:
```suggestion
    // fill with arbitrary values dont use c2h gen because NotDefaultConstructible is not default constructible
    for (offset_t i = 0; i < num_items; ++i)
    {
      d_vec[i] = NotDefaultConstructible{static_cast<int>(i)};
    }
```

cub/test/catch2_test_device_find_if.cu
```diff
@@ -251,16 +246,26 @@ C2H_TEST("Device find_if works with non default constructible types", "[device][
...
   {
-    c2h::device_vector<offset_t> out_result(1, thrust::no_init);
-    auto predicate = thrust::detail::equal_to_value<NotDefaultConstructible>{NotDefaultConstructible{val_to_find}};
-    find_if(c_it, thrust::raw_pointer_cast(out_result.data()), predicate, num_items);
-    REQUIRE(val_to_find == out_result[0]);
+    // fill with arbitrary values dont use c2h gen because NotDefaultConstructible is not default constructible
+    for (offset_t i = 0; i < num_items - 1; ++i)
+    {
+      d_vec[i] = NotDefaultConstructible{static_cast<int>(i)};
+    }
```
bernhardmgruber: Important: this performs extremely bad since we need a `cudaMemcpy` for each vector element. Can we use `thrust::tabulate` to setup `d_vec`?

cub/test/catch2_test_device_find_if.cu
```diff
@@ -251,16 +246,26 @@ C2H_TEST("Device find_if works with non default constructible types", "[device][
       min_items,
       max_items,
     }));
-  const input_t val_to_find = static_cast<input_t>(num_items - 1);
+  const int val_to_find = static_cast<int>(num_items - 1);
```
srinivasyadav18: nit:
```suggestion
  const auto val_to_find = static_cast<int>(num_items - 1);
```

cub/test/catch2_test_device_find_if.cu
```diff
@@ -251,16 +246,26 @@ C2H_TEST("Device find_if works with non default constructible types", "[device][
...
+    // fill with arbitrary values dont use c2h gen because NotDefaultConstructible is not default constructible
+    for (offset_t i = 0; i < num_items - 1; ++i)
+    {
+      d_vec[i] = NotDefaultConstructible{static_cast<int>(i)};
+    }
+
+    // set the value to find at the end
+    d_vec[num_items - 1] = NotDefaultConstructible{val_to_find};
   }
+  auto it = thrust::raw_pointer_cast(d_vec.data());
```
srinivasyadav18: nit: could use latest pointer traits

```suggestion
  auto ptr = cuda::std::to_address(d_vec.begin());
```

cub/test/catch2_test_device_find_if.cu
```diff
@@ -251,16 +246,26 @@ C2H_TEST("Device find_if works with non default constructible types", "[device][
...
-  const input_t val_to_find = static_cast<input_t>(num_items - 1);
+  const int val_to_find = static_cast<int>(num_items - 1);
 
   CAPTURE(num_items, val_to_find);
 
-  // counting_iterator input
-  auto c_it = cuda::make_transform_iterator(cuda::make_counting_iterator(input_t{0}), converter{});
+  // raw device iterator to some device vector so that vectorized path is taken
+  c2h::device_vector<input_t> d_vec(num_items, NotDefaultConstructible(0));
   {
```
srinivasyadav18: question: Do we need open brace `{` here (and close brace below) ?

cub/test/catch2_test_device_find_if.cu
gonidelis: that call throws a compilation error. not sure if i should be using  `cuda::std::to_address(vec.data())` instead. Will skip for now.

cub/test/catch2_test_device_find_if.cu
```diff
@@ -228,17 +229,21 @@ struct NotDefaultConstructible
...
+template <typename OffsetT>
+struct index_to_value
 {
-  __host__ __device__ constexpr NotDefaultConstructible operator()(const int val) const noexcept
+  __host__ __device__ NotDefaultConstructible operator()(OffsetT i)
   {
-    return NotDefaultConstructible{val};
+    return NotDefaultConstructible{static_cast<int>(i)};
   }
 };
```
bernhardmgruber: Strong suggestion: since we cast to `int` anyway when constructing `NotDefaultConstructible`, we should just revert the entire generalization to `OffsetT` here.

cub/test/catch2_test_device_find_if.cu
```diff
@@ -228,17 +229,21 @@ struct NotDefaultConstructible
...
 
+static_assert(cuda::std::is_default_constructible_v<NotDefaultConstructible> == false,
+              "NotDefaultConstructible should not be default constructible");
+
 C2H_TEST("Device find_if works with non default constructible types", "[device][find_if]")
 {
-  using input_t  = int32_t;
+  using input_t  = NotDefaultConstructible;
   using offset_t = int32_t;
```
bernhardmgruber: Since we spell it `int` everywhere:
```suggestion
  using offset_t = int;
```

bernhardmgruber (approved):

cub/test/catch2_test_device_find_if.cu
```diff
@@ -228,18 +229,21 @@ struct NotDefaultConstructible
...
-  __host__ __device__ constexpr NotDefaultConstructible operator()(const int val) const noexcept
+  __host__ __device__ NotDefaultConstructible operator()(int i)
   {
-    return NotDefaultConstructible{val};
+    return NotDefaultConstructible{static_cast<int>(i)};
   }
 };
 
+static_assert(cuda::std::is_default_constructible_v<NotDefaultConstructible> == false,
+              "NotDefaultConstructible should not be default constructible");
```
miscco:
```suggestion
static_assert(!cuda::std::is_default_constructible_v<NotDefaultConstructible>,
              "NotDefaultConstructible should not be default constructible");
```

Issue #5635, opened 2025-08-22
[EPIC]: Env-based overloads implementation for all CUB device primitives


gonidelis:
For every existing CUB device overload add an env-based counterpart that uses the [new environment facilities](https://github.com/NVIDIA/cccl/blob/main/libcudacxx/include/cuda/std/__execution/env.h) from libcu++.

The goal is that we provide CUB device overloads that do not require explicit memory allocation which requires a 2-phase call. The existing CUB API looks like this:

```c++
void *d_temp_storage = nullptr;
size_t temp_storage_bytes = 0;

// First call: determine storage requirements
cub::DeviceScan::InclusiveSum(d_temp_storage, temp_storage_bytes, d_in, d_out, num_items, stream);

// Allocate storage
cudaMalloc(&d_temp_storage, temp_storage_bytes);

// Second call: actual algorithm execution
cub::DeviceScan::InclusiveSum(d_temp_storage, temp_storage_bytes, d_in, d_out, num_items, stream);
```

_Working PRs for reference_: https://github.com/NVIDIA/cccl/pull/4877 https://github.com/NVIDIA/cccl/pull/4985

When the new APIs are added CUB should allow for the users to call the primitives like this:

```c++
auto env = cuda::execution::require(cuda::execution::determinism::run_to_run);

// Single call: allocation and execution in one call
cub::DeviceReduce::Sum(input.begin(), output.begin(), input.size(), env);
```

The new `environment` argument should enable:

1. access to [`deterministic` requirements](https://github.com/NVIDIA/cccl/pull/4876) of the algorithm execution
2. flexibility on the [`memory resource`](https://nvidia.github.io/cccl/libcudacxx/extended_api/memory_resource.html) selection
3. tuning without reaching for dispatch layer
4. encapsulate the existing `stream` selection option

### cub::DeviceAdjacentDifference

- [ ] Add env-based cub::DeviceAdjacentDifference::SubtractLeftCopy()
- [ ] Add env-based cub::DeviceAdjacentDifference::SubtractLeft()
- [ ] Add env-based cub::DeviceAdjacentDifference::SubtractRightCopy()
- [ ] Add env-based cub::DeviceAdjacentDifference::SubtractRight()

### cub::DeviceCopy
- [ ] Add env-based cub::DeviceCopy::Batched()

### cub::DeviceMemcpy
- [ ] Add env-based cub::DeviceMemcpy::Batched()

### cub::DeviceFor

[...]

PR #7459, opened 2026-02-03
Add env based cub::DevicePartition and tests

gonidelis:
fixes https://github.com/NVIDIA/cccl/issues/5635?issue=NVIDIA%7Ccccl%7C7458

### DIFF
--- a/cub/cub/device/device_partition.cuh
+++ b/cub/cub/device/device_partition.cuh
@@ -371,7 +371,7 @@ public:
     // Dispatch with environment - handles all boilerplate
     return detail::dispatch_with_env(env, [&]([[maybe_unused]] auto tuning, void* storage, size_t& bytes, auto stream) {
       using tuning_t = decltype(tuning);
-      return partition_impl<tuning_t, InputIteratorT, FlagIterator, OutputIteratorT, NumSelectedIteratorT, NullType, offset_t>(
+      return partition_impl<tuning_t>(
         storage, bytes, d_in, d_flags, d_out, d_num_selected_out, static_cast<offset_t>(num_items), NullType{}, stream);
     });
   }
@@ -627,8 +627,16 @@ public:
     // Dispatch with environment - handles all boilerplate
     return detail::dispatch_with_env(env, [&]([[maybe_unused]] auto tuning, void* storage, size_t& bytes, auto stream) {
       using tuning_t = decltype(tuning);
-      return partition_impl<tuning_t, InputIteratorT, NullType*, OutputIteratorT, NumSelectedIteratorT, SelectOp, offset_t>(
-        storage, bytes, d_in, nullptr, d_out, d_num_selected_out, static_cast<offset_t>(num_items), select_op, stream);
+      return partition_impl<tuning_t>(
+        storage,
+        bytes,
+        d_in,
+        static_cast<NullType*>(nullptr),
+        d_out,
+        d_num_selected_out,
+        static_cast<offset_t>(num_items),
+        select_op,
+        stream);
     });
   }
 
--- a/cub/test/catch2_test_device_partition_env_api.cu
+++ b/cub/test/catch2_test_device_partition_env_api.cu
@@ -8,6 +8,9 @@
 
 #include <thrust/device_vector.h>
 
+#include <cuda/devices>
+#include <cuda/stream>
+
 #include <iostream>
 
 #include "catch2_test_device_select_common.cuh"
@@ -21,8 +24,8 @@ C2H_TEST("cub::DevicePartition::If accepts env with stream", "[partition][env]")
   auto num_selected = thrust::device_vector<int>(1);
   less_than_t<int> le{5};
 
-  cudaStream_t legacy_stream = 0;
-  cuda::stream_ref stream_ref{legacy_stream};
+  cuda::stream stream{cuda::devices[0]};
+  cuda::stream_ref stream_ref{stream};
   auto env = cuda::std::execution::env{stream_ref};
 
   auto error = cub::DevicePartition::If(input.begin(), output.begin(), num_selected.begin(), input.size(), le, env);
@@ -49,8 +52,8 @@ C2H_TEST("cub::DevicePartition::Flagged accepts env with stream", "[partition][e
   auto output       = thrust::device_vector<int>(8);
   auto num_selected = thrust::device_vector<int>(1);
 
-  cudaStream_t legacy_stream = 0;
-  cuda::stream_ref stream_ref{legacy_stream};
+  cuda::stream stream{cuda::devices[0]};
+  cuda::stream_ref stream_ref{stream};
   auto env = cuda::std::execution::env{stream_ref};
 
   auto error = cub::DevicePartition::Flagged(


cub/cub/device/device_partition.cuh
```diff
@@ -19,11 +19,40 @@
...
+{
+  [[nodiscard]] _CCCL_TRIVIAL_API constexpr auto query(const get_tuning_query_t&) const noexcept -> Derived
+  {
+    return static_cast<const Derived&>(*this);
+  }
+};
+
+struct default_tuning : tuning<default_tuning>
+{
+  template <class InputT, class FlagT, class OffsetT, bool DistinctPartitions, SelectImpl Impl>
```
gonidelis, line 49: @bernhardmgruber given that DevicePartition is being tuned via DeviceSelect this should be correct, right?

bernhardmgruber: I think this is fine for now. I will need to rewrite it anyway when we move to the new tuning API.

cub/cub/device/device_partition.cuh
```diff
@@ -19,11 +19,40 @@
...
+{};
+
+template <class Derived>
+struct tuning
+{
+  [[nodiscard]] _CCCL_TRIVIAL_API constexpr auto query(const get_tuning_query_t&) const noexcept -> Derived
+  {
+    return static_cast<const Derived&>(*this);
+  }
+};
```
NaderAlAwar, line 45: Suggestion: move this to a common header file to avoid duplication

cub/cub/device/device_partition.cuh
```diff
@@ -193,6 +279,142 @@ struct DevicePartition
...
+            typename FlagIterator,
+            typename OutputIteratorT,
+            typename NumSelectedIteratorT,
+            typename NumItemsT,
+            typename EnvT = ::cuda::std::execution::env<>,
+            typename ::cuda::std::enable_if_t<
+              ::cuda::std::is_integral_v<NumItemsT> && !::cuda::std::is_same_v<InputIteratorT, void*>
+                && !::cuda::std::is_same_v<FlagIterator, size_t&>,
+              int> = 0>
+  CUB_RUNTIME_FUNCTION _CCCL_FORCEINLINE static cudaError_t Flagged(
```
NaderAlAwar: Critical: this should be marked `[[nodiscard]]`

cub/cub/device/device_partition.cuh
```diff
@@ -351,6 +573,147 @@ struct DevicePartition
...
+  template <
+    typename InputIteratorT,
+    typename OutputIteratorT,
+    typename NumSelectedIteratorT,
+    typename SelectOp,
+    typename NumItemsT,
+    typename EnvT = ::cuda::std::execution::env<>,
+    typename ::cuda::std::
+      enable_if_t<::cuda::std::is_integral_v<NumItemsT> && !::cuda::std::is_same_v<InputIteratorT, void*>, int> = 0>
+  CUB_RUNTIME_FUNCTION _CCCL_FORCEINLINE static cudaError_t
```
NaderAlAwar: Critical: `[[nodiscard]]`

cub/test/catch2_test_device_partition_env_api.cu
```diff
@@ -0,0 +1,60 @@
...
+{
+  // example-begin partition-if-env-determinism
+  auto input        = thrust::device_vector<int>{1, 2, 3, 4, 5, 6, 7, 8};
+  auto output       = thrust::device_vector<int>(8);
+  auto num_selected = thrust::device_vector<int>(1);
+  less_than_t<int> le{5};
+
+  auto env = cuda::execution::require(cuda::execution::determinism::run_to_run);
+
+  auto error = cub::DevicePartition::If(input.begin(), output.begin(), num_selected.begin(), input.size(), le, env);
```
NaderAlAwar, line 31: Important: Should do the same check we have in the other env api tests

cub/test/catch2_test_device_partition_env_api.cu
```diff
@@ -0,0 +1,60 @@
...
+{
+  // example-begin partition-flagged-env-determinism
+  auto input        = thrust::device_vector<int>{1, 2, 3, 4, 5, 6, 7, 8};
+  auto flags        = thrust::device_vector<char>{1, 0, 0, 1, 0, 1, 1, 0};
+  auto output       = thrust::device_vector<int>(8);
+  auto num_selected = thrust::device_vector<int>(1);
+
+  auto env = cuda::execution::require(cuda::execution::determinism::run_to_run);
+
+  auto error = cub::DevicePartition::Flagged(
```
NaderAlAwar, line 59: Important: same as above

cub/cub/device/device_partition.cuh
gonidelis: I don't think this is needed. this portion of code is being holistically re-written from bernhard anyhow

bernhardmgruber: Also note that we need this piece of code for every CUB algorithm in the algorithm's namespace. So even if we moved those somewhere else, we would need define `get_tuning_query_t` and `tuning` in ~25 different namespaces.

bernhardmgruber: I am going to move this elsewhere when we go public with the tuning API.

gonidelis: https://github.com/NVIDIA/cccl/issues/7497

bernhardmgruber (commented):
Also for this PR, if determinism does not make sense for `DevicePartition` I would remove all code related to it.

Does determinism mean stable here? And if it would, do we have stable and unstable versions of the algorithm?

cub/cub/device/device_partition.cuh
```diff
@@ -196,6 +277,105 @@ struct DevicePartition
...
+  {
+    _CCCL_NVTX_RANGE_SCOPE("cub::DevicePartition::Flagged");
+
+    using offset_t = detail::choose_offset_t<NumItemsT>;
+
+    // Dispatch with environment - handles all boilerplate
+    return detail::dispatch_with_env(env, [&]([[maybe_unused]] auto tuning, void* storage, size_t& bytes, auto stream) {
+      using tuning_t = decltype(tuning);
+      return partition_impl<tuning_t, InputIteratorT, FlagIterator, OutputIteratorT, NumSelectedIteratorT, NullType, offset_t>(
+        storage, bytes, d_in, d_flags, d_out, d_num_selected_out, static_cast<offset_t>(num_items), NullType{}, stream);
```
bernhardmgruber: Suggestion: let the compiler do the work for you:
```suggestion
      return partition_impl<tuning_t>(
        storage, bytes, d_in, d_flags, d_out, d_num_selected_out, static_cast<offset_t>(num_items), NullType{}, stream);
```

cub/cub/device/device_partition.cuh
```diff
@@ -357,6 +537,101 @@ struct DevicePartition
...
+    _CCCL_NVTX_RANGE_SCOPE("cub::DevicePartition::If");
+
+    using offset_t = detail::choose_offset_t<NumItemsT>;
+
+    // Dispatch with environment - handles all boilerplate
+    return detail::dispatch_with_env(env, [&]([[maybe_unused]] auto tuning, void* storage, size_t& bytes, auto stream) {
+      using tuning_t = decltype(tuning);
+      return partition_impl<tuning_t, InputIteratorT, NullType*, OutputIteratorT, NumSelectedIteratorT, SelectOp, offset_t>(
+        storage, bytes, d_in, nullptr, d_out, d_num_selected_out, static_cast<offset_t>(num_items), select_op, stream);
+    });
```
bernhardmgruber:
```suggestion
      return partition_impl<tuning_t>(
        storage, bytes, d_in, nullptr, d_out, d_num_selected_out, static_cast<offset_t>(num_items), select_op, stream);
    });
```

cub/test/catch2_test_device_partition_env_api.cu
```diff
@@ -0,0 +1,71 @@
...
+  auto input        = thrust::device_vector<int>{1, 2, 3, 4, 5, 6, 7, 8};
+  auto output       = thrust::device_vector<int>(8);
+  auto num_selected = thrust::device_vector<int>(1);
+  less_than_t<int> le{5};
+
+  cudaStream_t legacy_stream = 0;
+  cuda::stream_ref stream_ref{legacy_stream};
+  auto env = cuda::std::execution::env{stream_ref};
+
+  auto error = cub::DevicePartition::If(input.begin(), output.begin(), num_selected.begin(), input.size(), le, env);
```
bernhardmgruber:
```suggestion
  cuda::stream_ref stream_ref{cuda::devices[0]};

  auto error = cub::DevicePartition::If(input.begin(), output.begin(), num_selected.begin(), input.size(), le, stream_ref);
```
Applies more times.

gonidelis: this doesn't compile I use this instead

```cpp
  cuda::stream stream{cuda::devices[0]};
```

gonidelis: I would like to keep the `env` thingy so users can see the coding paradigm

bernhardmgruber: Ah, so we can't have a `steram_ref` to a device, it needs to refer to an actual stream that was created for that device. Makes sense. My bad. Thx for trying!

PR #7461, opened 2026-02-03
Add env based DeviceRadixSort::SortPairs*() and tests

gonidelis:
fixes part of  https://github.com/NVIDIA/cccl/issues/7545

Adds env:
- `RadixSort::SortPairs()`
- `RadixSort::SortPairsDescending()`

It adds env based default API and DoublBuffer API but **not the decomposer APIs**. These will be added in subsequent PRs.

Adding RadixSort::SortKeys* in subsequent PR

**No determinsm requirements are imposed or queried for `RadixSort`. Any non-deterministic solution should be wrong**

### DIFF
--- a/cub/cub/device/device_radix_sort.cuh
+++ b/cub/cub/device/device_radix_sort.cuh
@@ -196,37 +196,6 @@ private:
       decomposer);
   }
 
-  template <typename TuningEnvT,
-            SortOrder Order,
-            typename KeyT,
-            typename ValueT,
-            typename OffsetT,
-            typename DecomposerT = detail::identity_decomposer_t>
-  CUB_RUNTIME_FUNCTION static cudaError_t sort_pairs_env_dispatch(
-    void* d_temp_storage,
-    size_t& temp_storage_bytes,
-    DoubleBuffer<KeyT>& d_keys,
-    DoubleBuffer<ValueT>& d_values,
-    OffsetT num_items,
-    int begin_bit,
-    int end_bit,
-    bool is_overwrite_okay,
-    cudaStream_t stream,
-    DecomposerT decomposer = {})
-  {
-    return detail::radix_sort::dispatch<Order>(
-      d_temp_storage,
-      temp_storage_bytes,
-      d_keys,
-      d_values,
-      static_cast<OffsetT>(num_items),
-      begin_bit,
-      end_bit,
-      is_overwrite_okay,
-      stream,
-      decomposer);
-  }
-
   // Name reported for NVTX ranges
   _CCCL_HOST_DEVICE static constexpr auto GetName() -> const char*
   {
@@ -384,8 +353,8 @@ public:
   //! @rst
   //! Sorts key-value pairs into ascending order using :math:`\approx 2N` auxiliary storage.
   //!
-  //! .. versionadded:: 3.2.0
-  //!    First appears in CUDA Toolkit 13.2.
+  //! .. versionadded:: 3.4.0
+  //!    First appears in CUDA Toolkit 13.4.
   //!
   //! This is an environment-based API that allows customization of:
   //!
@@ -477,22 +446,12 @@ public:
     using offset_t = detail::choose_offset_t<NumItemsT>;
 
     // Dispatch with environment - handles all boilerplate
-    constexpr bool is_overwrite_okay = false;
     DoubleBuffer<KeyT> d_keys(const_cast<KeyT*>(d_keys_in), d_keys_out);
     DoubleBuffer<ValueT> d_values(const_cast<ValueT*>(d_values_in), d_values_out);
 
     return detail::dispatch_with_env(env, [&]([[maybe_unused]] auto tuning, void* storage, size_t& bytes, auto stream) {
-      using tuning_t = decltype(tuning);
-      return sort_pairs_env_dispatch<tuning_t, SortOrder::Ascending>(
-        storage,
-        bytes,
-        d_keys,
-        d_values,
-        static_cast<offset_t>(num_items),
-        begin_bit,
-        end_bit,
-        is_overwrite_okay,
-        stream);
+      return detail::radix_sort::dispatch<SortOrder::Ascending>(
+        storage, bytes, d_keys, d_values, static_cast<offset_t>(num_items), begin_bit, end_bit, false, stream);
     });
   }
 
@@ -935,8 +894,8 @@ public:
   //! @rst
   //! Sorts key-value pairs into ascending order using :math:`\approx N` auxiliary storage.
   //!
-  //! .. versionadded:: 3.2.0
-  //!    First appears in CUDA Toolkit 13.2.
+  //! .. versionadded:: 3.4.0
+  //!    First appears in CUDA Toolkit 13.4.
   //!
   //! This is an environment-based API that allows customization of:
   //!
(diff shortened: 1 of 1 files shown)


cub/cub/device/device_radix_sort.cuh
```diff
@@ -333,6 +389,115 @@ public:
...
+        d_values_out,
+        static_cast<offset_t>(num_items),
+        begin_bit,
+        end_bit,
+        stream);
+    });
+  }
+
+  //! @rst
+  //! Sorts key-value pairs into ascending order using :math:`\approx 2N` auxiliary storage.
```
NaderAlAwar, line 459: Important: 
```suggestion
  //! Sorts key-value pairs into descending order using :math:`\approx 2N` auxiliary storage.
```

cub/test/catch2_test_device_radix_sort_env.cu
```diff
@@ -0,0 +1,254 @@
+// SPDX-FileCopyrightText: Copyright (c) 2025, NVIDIA CORPORATION. All rights reserved.
```
NaderAlAwar: Suggestion:
```suggestion
// SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION. All rights reserved.
```

cub/cub/device/device_radix_sort.cuh
```diff
@@ -333,6 +389,115 @@ public:
...
+            typename EnvT = ::cuda::std::execution::env<>,
+            typename ::cuda::std::enable_if_t<::cuda::std::is_integral_v<NumItemsT>, int> = 0>
+  [[nodiscard]] CUB_RUNTIME_FUNCTION _CCCL_FORCEINLINE static cudaError_t SortPairs(
+    const KeyT* d_keys_in,
+    KeyT* d_keys_out,
+    const ValueT* d_values_in,
+    ValueT* d_values_out,
+    NumItemsT num_items,
+    int begin_bit,
+    int end_bit,
```
NaderAlAwar: Important: I would add default values to these two, as it seems too verbose to force the users to specify begin and end bits if they want to use the env API

cub/cub/device/device_radix_sort.cuh
```diff
@@ -1204,6 +1369,115 @@ public:
...
+            typename EnvT = ::cuda::std::execution::env<>,
+            typename ::cuda::std::enable_if_t<::cuda::std::is_integral_v<NumItemsT>, int> = 0>
+  [[nodiscard]] CUB_RUNTIME_FUNCTION _CCCL_FORCEINLINE static cudaError_t SortPairsDescending(
+    const KeyT* d_keys_in,
+    KeyT* d_keys_out,
+    const ValueT* d_values_in,
+    ValueT* d_values_out,
+    NumItemsT num_items,
+    int begin_bit,
+    int end_bit,
```
NaderAlAwar: Important: same comment as above

cub/test/catch2_test_device_radix_sort_env.cu
```diff
@@ -0,0 +1,254 @@
...
+  REQUIRE(
+    cudaSuccess
+    == cub::DeviceRadixSort::SortPairs(
+      keys_in.data().get(),
+      keys_out.data().get(),
+      values_in.data().get(),
+      values_out.data().get(),
+      static_cast<int>(keys_in.size()),
+      0,
+      static_cast<int>(sizeof(int) * 8)));
```
NaderAlAwar: Suggestion: once we add defaults to begin and end bits, these two don't have to be specified. Same for other tests in this file

cub/test/catch2_test_device_radix_sort_env_api.cu
```diff
@@ -0,0 +1,63 @@
...
+  auto values_out = thrust::device_vector<int>(7);
+
+  auto error = cub::DeviceRadixSort::SortPairs(
+    keys_in.data().get(),
+    keys_out.data().get(),
+    values_in.data().get(),
+    values_out.data().get(),
+    static_cast<int>(keys_in.size()),
+    0,
+    sizeof(int) * 8);
```
NaderAlAwar: Suggestion: remove these two when we default these, same for other tests in this file

cub/test/catch2_test_device_radix_sort_env_api.cu
```diff
@@ -0,0 +1,63 @@
...
+
```
NaderAlAwar, line 27: Important: add a check of the error since this is an API test (same for other tests)

```
  if (error != cudaSuccess)
  {
    std::cerr << "cub::DeviceRadixSort::SortPairs failed with status: " << error << std::endl;
  }
```

cub/cub/device/device_radix_sort.cuh
```diff
@@ -1204,6 +1369,115 @@ public:
   //! .. versionadded:: 2.2.0
   //!    First appears in CUDA Toolkit 12.3.
   //!
```
bernhardmgruber: Important: the `versionadded` annotation should go to the next overload below the newly added function. Please also add a `versionadded` annotation for your new overload.

cub/cub/device/device_radix_sort.cuh
```diff
@@ -333,6 +389,115 @@ public:
   //! .. versionadded:: 2.2.0
   //!    First appears in CUDA Toolkit 12.3.
```
bernhardmgruber: Same here.

cub/test/catch2_test_device_radix_sort_env.cu
```diff
@@ -0,0 +1,254 @@
+// SPDX-FileCopyrightText: Copyright (c) 2025, NVIDIA CORPORATION. All rights reserved.
+// SPDX-License-Identifier: BSD-3-Clause
```
bernhardmgruber: Important: wrong license.

cub/cub/device/device_radix_sort.cuh
gonidelis: @NaderAlAwar unless i am missing sth, why? this is not a descending overload

cub/test/catch2_test_device_radix_sort_env.cu
gonidelis: where can i find more info about when to use `Apache` and when to use `BSD`?

bernhardmgruber: Here: https://github.com/NVIDIA/cccl/wiki/Cpp-Coding-Guidelines#licensing. Quoting:

> Any net new file should be licensed under Apache-2.0 WITH LLVM-exception.

Existing files seem to retain their license when making modifications. At least that's what everybody did when I started here ;) If you copy a large part of code to a new file, you should also retain the license from the old file IMO, since the code should keep the license it was written under. If you are unsure what to do, start a discussion.

cub/cub/device/device_radix_sort.cuh
NaderAlAwar: My mistake

cub/cub/device/device_radix_sort.cuh
```diff
@@ -330,8 +384,120 @@ public:
   //! @rst
   //! Sorts key-value pairs into ascending order using :math:`\approx 2N` auxiliary storage.
   //!
-  //! .. versionadded:: 2.2.0
-  //!    First appears in CUDA Toolkit 12.3.
+  //! .. versionadded:: 3.2.0
+  //!    First appears in CUDA Toolkit 13.2.
```
bernhardmgruber: Important: we already branched CCCL 3.3, so this PR currently targets 3.4
```suggestion
  //! .. versionadded:: 3.4.0
  //!    First appears in CUDA Toolkit 13.4.
```
Applies to the other newly added overloads as well.

cub/cub/device/device_radix_sort.cuh
```diff
@@ -173,6 +196,37 @@ private:
...
+      temp_storage_bytes,
+      d_keys,
+      d_values,
+      static_cast<OffsetT>(num_items),
+      begin_bit,
+      end_bit,
+      is_overwrite_okay,
+      stream,
+      decomposer);
+  }
```
bernhardmgruber: Suggestion: maybe we don't even need this function anymore and should just call `detail::radix_sort::dispatch` directly wherever `sort_pairs_env_dispatch` is called now.

bernhardmgruber (approved):

cub/cub/device/device_radix_sort.cuh
gonidelis: aren't we adding this overload in the next (3.2) release? or **is** 3.4 the next release?

cub/cub/device/device_radix_sort.cuh
gonidelis: makes more sense - yes

NaderAlAwar (approved):

Issue #6143, opened 2025-10-07
[FEA]: SoL on cub::Transform for custom/zip input and output iterator


Kh4ster:
### Is this a duplicate?

- [x] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cccl/issues) for this request and that I agree to the [Code of Conduct](CODE_OF_CONDUCT.md)

### Area

CUB

### Is your feature request related to a problem? Please describe.

In cuOpt's PDLP Solver (linear programming solver) we are dependent on cub::Transform.

We call it at every single iteration of the solver here: https://github.com/NVIDIA/cuopt/blob/branch-25.10/cpp/src/linear_programming/pdhg.cu#L197

And you can find the functor here: https://github.com/NVIDIA/cuopt/blob/branch-25.10/cpp/src/linear_programming/utils.cuh#L102

I wanted to make sure with the CCCL team that this case of input/output iterator is properly handled by cub::Transform to reach SoL.

### Describe the solution you'd like

An SoL cub::Transform for such iterator (which might already be the case)

bernhardmgruber:
Thank you for creating this issue. The main motivation for me was (after internal discussion) that the transformation writes to three output arrays, which is a use case we have considered in the past, but not implemented. I think we should derive a benchmark from the code you pointed to and add a new variant of `DeviceTransform` to writes to multiple output arrays.

PR #7473, opened 2026-02-03
Add `cub::DeviceTransform` N->M API entrypoint

bernhardmgruber:
Related to: #6143

### DIFF
--- a/cub/test/catch2_test_device_transform_api.cu
+++ b/cub/test/catch2_test_device_transform_api.cu
@@ -15,13 +15,13 @@ void test_transform_many_many_api()
 {
   // example-begin transform-many-many
   auto input1 = thrust::device_vector<int>{0, -1, 2, -3, 4, -5};
-  auto input2 = thrust::device_vector<float>{5.2f, 3.1f, -1.1f, 3.0f, 3.2f, 0.0f};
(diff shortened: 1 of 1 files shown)


cub/cub/device/device_transform.cuh
```diff
@@ -108,7 +108,110 @@ private:
...
+    cudaStream_t stream = nullptr)
+  {
+    if (d_temp_storage == nullptr)
+    {
+      temp_storage_bytes = 1;
+      return cudaSuccess;
+    }
+
+    return Transform(
+      ::cuda::std::move(inputs), ::cuda::std::move(outputs), num_items, ::cuda::std::move(transform_op), stream);
```
miscco, line 211: Nitpick: I believe we could directly forward to
```suggestion
    _CCCL_NVTX_RANGE_SCOPE("cub::DeviceTransform::Transform");
    return TransformInternal(
      ::cuda::std::move(inputs),
      ::cuda::std::move(outputs),
      num_items,
      detail::transform::always_true_predicate{},
      ::cuda::std::move(transform_op),
      ::cuda::stream_ref{stream});
```

bernhardmgruber: We could, but I always call the other API entrypoint first for the temp storage overloads. I'd like to keep it consistent.

PR #7489, opened 2026-02-04
Add docs for cccl-runtime 3.2 additions and cccl-runtime landing page

pciolkosz:
The cccl-runtime landing page right now is just a plain list of APIs. This PR adds some short description and CUDA Runtime interactions page.

It also expands the list of documented APIs with things added in 3.2

### DIFF
--- a/docs/libcudacxx/runtime/event.rst
+++ b/docs/libcudacxx/runtime/event.rst
@@ -3,7 +3,7 @@
 Events
 ======
 
-Event is a snapshot of execution state of a stream. It can be used to synchronize work submitted to a stream up to a certain point, estabilish dependency between streams or measure time passed between two events.
+Event is a snapshot of execution state of a stream. It can be used to synchronize work submitted to a stream up to a certain point, establish dependency between streams or measure time passed between two events.
 
 ``cuda::event_ref``
 --------------------------------------------------


docs/libcudacxx/runtime/buffer.rst
```diff
@@ -0,0 +1,218 @@
...
+
+While the buffer operates in stream order, it can also be constructed with a :ref:`synchronous_resource <libcudacxx-extended-api-memory-resources-synchronous-resource>`, in which case it will automatically use the :ref:`synchronous_resource_adapter <libcudacxx-extended-api-memory-resources-synchronous-adapter>` to wrap the provided resource.
+
+Availability: CCCL 3.2.0 / CUDA 13.2
+
+Example:
+
+.. code:: cpp
+
+   #include <cuda/buffer>
```
jrhemstad, line 26: suggestion: Let's use `literalinclude` to pull code snippets from real sample files that get compiled as part of our CI. This ensures our code samples never bit rot.

jrhemstad: This would be a good task for an agent.

miscco (approved):

docs/libcudacxx/runtime/buffer.rst
pciolkosz: Filed https://github.com/NVIDIA/cccl/issues/7560 to track this to not block the docs

PR #7491, opened 2026-02-04
Drop usage of pickle

miscco:
We are moving away from it because of security concerns, just use json

Addresses internal gitlab issue 2

### DIFF
--- a/python/cuda_cccl/cuda/coop/_caching.py
+++ b/python/cuda_cccl/cuda/coop/_caching.py
@@ -5,7 +5,6 @@
 import hashlib
 import json
 import os
-import pickle
 
 _ENABLE_CACHE = "CCCL_ENABLE_CACHE" in os.environ
 if _ENABLE_CACHE:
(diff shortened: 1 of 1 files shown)


python/cuda_cccl/cuda/coop/_caching.py
```diff
@@ -32,15 +31,15 @@ def cacher(*args, **kwargs):
...
                 with open(os.path.join(_CACHE_LOCATION, h), "rb") as f:
-                    out = pickle.load(f)
+                    out = json.load(f)
                 # return cache
                 return out
             else:
                 # compute output
                 out = func(*args, **kwargs)
                 # store to file
                 with open(os.path.join(_CACHE_LOCATION, h), "wb") as f:
```
shwina, line 41: I think we no longer need `"wb"` mode here - just `"w"` (ditto above with `"rb"` - just `"r"` should be used).

miscco: I would leave that as a followup

PR #7494, opened 2026-02-04
[STF] Fix equality operators for places

caugonnet:
## Description

Equality operator is broken for exec places and data places which might share the same affine data place for example, as this is typically the case with green contexts. We fix that here.

closes 

## Checklist

### DIFF
--- a/cudax/include/cuda/experimental/__stf/places/places.cuh
+++ b/cudax/include/cuda/experimental/__stf/places/places.cuh
@@ -1117,6 +1117,73 @@ private:
 };
 
 /**
+ * @brief RAII guard that activates an execution place and restores the previous one on destruction.
+ *
+ * This class provides a scoped mechanism for temporarily switching the active execution place.
+ * When constructed, it activates the given execution place (e.g., sets the current CUDA device).
+ * When destroyed, it restores the previous execution place that was active before construction.
+ *
+ * The guard is non-copyable and non-movable to ensure proper RAII semantics.
+ *
+ * @note This class only accepts `exec_place` objects. Implicit conversions from other types
+ *       (such as `data_place`) are explicitly disabled to prevent accidental misuse.
+ *
+ * Example usage:
+ * @code
+ * // Assume current device is 0
+ * {
+ *   exec_place_guard guard(exec_place::device(1));
+ *   // Device 1 is now active
+ *   // ... perform operations on device 1 ...
+ * }
+ * // Device 0 is restored
+ * @endcode
+ */
+class exec_place_guard
+{
+public:
+  /**
+   * @brief Constructs the guard and activates the given execution place.
+   *
+   * @param place The execution place to activate. Must be an `exec_place` object;
+   *              implicit conversions from other types are disabled.
+   */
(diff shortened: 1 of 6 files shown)


cudax/include/cuda/experimental/__stf/places/exec/green_context.cuh
```diff
@@ -474,6 +486,105 @@ inline data_place data_place::green_ctx(::std::shared_ptr<green_ctx_view> gc_vie
...
+
+  // Same green context data place should be equal
+  EXPECT(dp0a == dp0b);
+  EXPECT(!(dp0a != dp0b));
+
+  // Different green context data places should NOT be equal
+  EXPECT(dp0a != dp1);
+  EXPECT(!(dp0a == dp1));
+
+  // Green context data place should not be equal to regular device data place
```
caugonnet, line 580: Well this depends on how we create the green context actually .... so it may fail, unless we specfied it in the actor

caugonnet:
The comparison looks broken too, it probably relies on the fact that we have static objects for most places (host, device, ...) but it will not be reliable for green contexts. We can have two identical green contexts (same view) but different objects.

```
// To use in a ::std::map indexed by exec_place
  bool operator<(const exec_place& rhs) const
  {
    return pimpl < rhs.pimpl;
  }
```

cudax/include/cuda/experimental/__stf/places/exec/green_context.cuh
```diff
@@ -407,6 +407,18 @@ public:
       return *pool;
     }
 
+    bool operator==(const exec_place::impl& rhs) const override
```
miscco, line 425: Should those also implement !=?

cudax/include/cuda/experimental/__stf/places/exec/cuda_stream.cuh
```diff
@@ -69,6 +69,17 @@ public:
...
 
+    bool operator==(const exec_place::impl& rhs) const override
+    {
+      auto other = dynamic_cast<const impl*>(&rhs);
+      if (!other)
+      {
+        return false;
+      }
+      // Compare by stream handle
+      return dstream.stream == other->dstream.stream;
```
andralex: We could get fancy and use the feature that supports such cases quite nicely:

```suggestion
      if (auto other = dynamic_cast<const impl*>(&rhs); other)
      {
        return dstream.stream == other->dstream.stream;
      }
      // Compare by stream handle
      return false;
```

andralex:
lgtm

cudax/include/cuda/experimental/__stf/places/exec/green_context.cuh
caugonnet: I've now also added != and comparison methods like > <= >= ...

cudax/include/cuda/experimental/__stf/places/exec/green_context.cuh
caugonnet: this is actually ok, because this is using the "green-context specific" data place type

PR #7496, opened 2026-02-04
feat: add central equal to value predicate

thedevmystic:
This PR addresses:
  - centralized `equal_to_value<T>` predicate.
  - refactored tests to use the new predicates.


thedevmystic:
Hello, @bernhardmgruber! I think I might need some "vetting" to check my code is working perfectly before opening the PR for review. Thanks :)

libcudacxx/include/cuda/__functional/equal_to_value.h
```diff
@@ -0,0 +1,73 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-FileCopyrightText: Copyright (c) 2024 NVIDIA CORPORATION & AFFILIATES.
```
bernhardmgruber:
```suggestion
// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES.
```

libcudacxx/test/libcudacxx/std/algorithms/alg.modifying/alg.transform/pstl_binary_transform.cu
```diff
@@ -54,7 +47,7 @@ C2H_TEST("cuda::std::transform", "[parallel algorithm]")
       cuda::counting_iterator{0ull},
       output.begin(),
       cuda::std::minus<>{});
-    CHECK(thrust::all_of(output.begin(), output.end(), equal_to_one{}));
+    CHECK(thrust::all_of(output.begin(), output.end(), cuda::__equal_to_one{}));
```
bernhardmgruber: Suggestion: I think we can spell a `1` here:
```suggestion
    CHECK(thrust::all_of(output.begin(), output.end(), cuda::__equal_to_value{1}));
```
Applies to all occurrences below.

libcudacxx/include/cuda/__functional/equal_to_value.h
```diff
@@ -0,0 +1,73 @@
...
+/// @brief `__equal_to_one` is a function object that checks if a value is equal to one.
+/// It inherits from `__equal_to_value` and initializes it with the value `1`.
+/// @tparam _T The type of the value to be compared.
+template <typename _T>
+struct __equal_to_one : __equal_to_value<T>
+{
+  constexpr __equal_to_one() noexcept
+      : __equal_to_value<T>(static_cast<T>(1))
+  {}
+};
```
bernhardmgruber: Suggestion: this can easily be represented as `__equal_to_value{1}`, so I think there is no need for a dedicated functor.

libcudacxx/include/cuda/__functional/equal_to_value.h
```diff
@@ -0,0 +1,73 @@
...
+/// the default value of its type. It inherits from `__equal_to_value` and initializes
+/// it with the default-constructed value of type `_T`.
+/// @tparam _T The type of the value to be compared.
+template <typename _T>
+struct __equal_to_default : __equal_to_value<T>
+{
+  constexpr __equal_to_default() noexcept
+      : __equal_to_value<T>(T{})
+  {}
+};
```
bernhardmgruber: Suggestion: this is really just the same as `__equal_to_value{}`, so I think there is no need for a dedicated functor. Please remove it.

libcudacxx/include/cuda/__functional/equal_to_value.h
```diff
@@ -0,0 +1,73 @@
...
+  _T value_;
+
+  explicit constexpr __equal_to_value(const T& value) noexcept
+      : value_(value)
+  {}
+
+  __host__ __device__ constexpr bool operator()(const T& lhs) const noexcept
+  {
+    return lhs == value_;
+  }
```
bernhardmgruber:
```suggestion
  [[nodiscard]] _CCCL_API constexpr bool operator()(const T& lhs) const noexcept
  {
    return lhs == value_;
  }
```

libcudacxx/include/cuda/__functional/equal_to_value.h
```diff
@@ -0,0 +1,73 @@
...
+/// type of the value to be compared.
+/// @tparam _T The type of the value to be compared.
+template <typename _T>
+struct __equal_to_value
+{
+  _T value_;
+
+  explicit constexpr __equal_to_value(const T& value) noexcept
+      : value_(value)
+  {}
```
bernhardmgruber: Suggestion: Let's just rely on aggregate init:
```suggestion
```
Then we may need a deduction guide though after the class:
```c++
template <typename T>
__equal_to_value(T) -> __equal_to_value<T>;
```

libcudacxx/include/cuda/__functional/equal_to_value.h
```diff
@@ -0,0 +1,73 @@
...
+
+  __host__ __device__ constexpr bool operator()(const T& lhs) const noexcept
```
bernhardmgruber: There is one version of this functor in Thrust that relies on having a different type here. Let's try to retain this:
```suggestion
template <typename U>
  __host__ __device__ constexpr bool operator()(const U& lhs) const noexcept
```

bernhardmgruber (commented):
We are off to a good start, thx!

libcudacxx/include/cuda/__functional/equal_to_value.h
thedevmystic: Sorry, my copilot auto complete must've missed that. Thanks for the catch though 👍

thedevmystic:
Applied all suggestions! See if these needs any more tweaking or not. Thanks :)

libcudacxx/include/cuda/__functional/equal_to_value.h
```diff
@@ -0,0 +1,60 @@
...
+
+#include <cuda/std/__cccl/prologue.h>
+
+_CCCL_BEGIN_NAMESPACE_CUDA
+
+/// @brief `__equal_to_value` is a function object that checks if a value is equal to
+/// a specified value. It takes a single template parameter `_T`, which represents the
+/// type of the value to be compared.
+/// @tparam _T The type of the value to be compared.
+template <typename _T>
```
miscco: `_T` is a reserved name on windows, we should never use single letter macros for that reason.

Please replace with the common
```suggestion
template <typename _Tp>
```

Applies throughout

libcudacxx/include/cuda/__functional/equal_to_value.h
```diff
@@ -0,0 +1,60 @@
...
+#endif // no system header
+
+#include <cuda/std/__cccl/prologue.h>
+
+_CCCL_BEGIN_NAMESPACE_CUDA
+
+/// @brief `__equal_to_value` is a function object that checks if a value is equal to
+/// a specified value. It takes a single template parameter `_T`, which represents the
+/// type of the value to be compared.
+/// @tparam _T The type of the value to be compared.
```
miscco: We conventionally use
```suggestion
//! @brief `__equal_to_value` is a function object that checks if a given value is equal to
//! a stored one. 
//! @tparam _Tp The type of the value to be compared.
```

libcudacxx/include/cuda/__functional/equal_to_value.h
```diff
@@ -0,0 +1,60 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES.
+//
+//===----------------------------------------------------------------------===//
```
miscco, line 9: License is a difficult topic, but this is not part of libc++, so we should use
```suggestion
//===----------------------------------------------------------------------===//
//
// Part of libcu++, the C++ Standard Library for your entire system,
// under the Apache License v2.0 with LLVM Exceptions.
// See https://llvm.org/LICENSE.txt for license information.
// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES.
//
//===----------------------------------------------------------------------===//
```

libcudacxx/include/cuda/__functional/equal_to_value.h
```diff
@@ -0,0 +1,60 @@
...
+_CCCL_BEGIN_NAMESPACE_CUDA
+
+/// @brief `__equal_to_value` is a function object that checks if a value is equal to
+/// a specified value. It takes a single template parameter `_T`, which represents the
+/// type of the value to be compared.
+/// @tparam _T The type of the value to be compared.
+template <typename _T>
+struct __equal_to_value
+{
+  _T value_;
```
miscco: We should also uglify members and variables
```suggestion
  _Tp __value_;
```

Applies throughout

libcudacxx/include/cuda/__functional/equal_to_value.h
```diff
@@ -0,0 +1,60 @@
...
+
+  explicit constexpr __equal_to_value(const _T& value) noexcept
+      : value_(value)
+  {}
```
miscco: This is not correct, the value needs to be
```suggestion
  explicit constexpr __equal_to_value(const _Tp& __value) noexcept(::cuda::std::is_nothrow_copy_constructible_v<_Tp>)
      : __value_(__value)
  {}
```

libcudacxx/include/cuda/__functional/equal_to_value.h
```diff
@@ -0,0 +1,60 @@
...
+
+#if defined(_CCCL_IMPLICIT_SYSTEM_HEADER_GCC)
+#  pragma GCC system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_CLANG)
+#  pragma clang system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_MSVC)
+#  pragma system_header
+#endif // no system header
+
+#include <cuda/std/__cccl/prologue.h>
```
miscco, line 28:
```suggestion
#include <cuda/std/__type_traits/is_nothrow_copy_constructible.h>

#include <cuda/std/__cccl/prologue.h>
```

libcudacxx/include/cuda/__functional/equal_to_value.h
```diff
@@ -0,0 +1,60 @@
...
+  [[nodiscard]] _CCCL_API constexpr bool operator()(const _T& lhs) const noexcept
+  {
+    return lhs == value_;
+  }
+
+  template <typename _U>
+  [[nodiscard]] _CCCL_API constexpr bool operator()(const _U& lhs) const noexcept
+  {
+    return lhs == value_;
+  }
```
miscco: This is not fully correct, we need to ensure that this is nothrow comparable and convertible to bool.

Also we want combine both call operators to make the code less verbose. The `_Tp` case is handled fine by the template
```suggestion
  _CCCL_TEMPLATE(class _Up)
  _CCCL_REQUIRES(::cuda::std::__is_cpp17_equality_comparable_v<_Tp, _Up>)
  [[nodiscard]] _CCCL_API constexpr bool operator()(const _Up& __lhs) const noexcept(::cuda::std::__is_cpp17_nothrow_equality_comparable_v<_Tp, _Up>)
  {
    return static_cast<bool>(__value_ == __lhs);
  }
```

Note that `_CCCL_TEMPLATE` and `_CCCL_REQUIRES` are some horrible macro crimes we do to use concepts in C++20 and SFINAE in C++17

libcudacxx/include/cuda/__functional/equal_to_value.h
```diff
@@ -0,0 +1,60 @@
...
+};
+
+/// @brief Deduction guide for `__equal_to_value<_T>`.
+template <typename _T>
+__equal_to_value(_T) -> __equal_to_value<_T>;
```
miscco: We should not need a deduction guide for this in C++17
```suggestion
```

libcudacxx/include/cuda/__functional/equal_to_value.h
```diff
@@ -0,0 +1,60 @@
...
+#include <cuda/std/__cccl/prologue.h>
+
+_CCCL_BEGIN_NAMESPACE_CUDA
+
+/// @brief `__equal_to_value` is a function object that checks if a value is equal to
+/// a specified value. It takes a single template parameter `_T`, which represents the
+/// type of the value to be compared.
+/// @tparam _T The type of the value to be compared.
+template <typename _T>
+struct __equal_to_value
```
miscco: I am wondering whether we want to make this available to users? @bernhardmgruber thoughts

miscco (changes requested):
Thanks a lot for working on this 🎉 

There is a lot of folklore to be satisfied, that is not obvious.

We are also missing tests, please give `libcudacxx/test/libcudacxx/std/utilities/function.objects/comparisons/equal_to.pass.cpp` a look

The new test should live in `libcudacxx/test/libcudacxx/cuda/utilities/function.objects/comparisons/equal_to_value.pass.cpp`

You can verify whether the code builds fine by running

```
./ci/build_libcudacxx.sh
cd build
lit latest/libcudacxx/libcudacxx/test/libcudacxx/cuda/utilities/function.objects/comparisons/equal_to_value.pass.cpp
```

libcudacxx/include/cuda/__functional/equal_to_value.h
thedevmystic: Deduction guide was asked to be added by @bernhardmgruber.

libcudacxx/include/cuda/__functional/equal_to_value.h
thedevmystic: Okay no prob, whatever suits existing style!

libcudacxx/include/cuda/__functional/equal_to_value.h
thedevmystic: I only added this license boilerplate because existing files in cuda/__functional had the same license.

libcudacxx/include/cuda/__functional/equal_to_value.h
miscco: @bernhardmgruber explain yourself ;)

we should not need this

libcudacxx/include/cuda/__functional/equal_to_value.h
thedevmystic: Yeah, I can see. These macros do look very aweful for templates. But if this is the way to go then okay.

libcudacxx/include/cuda/__functional/equal_to_value.h
miscco: Please also test CTAD in the tests. you can do

```cpp
  [[maybe_unused]] __equal_to_value with_ctad{1.0f};
  static_assert(cuda::std::is_same_v<decltype(with_ctad), __equal_to_value<float>>);
```

libcudacxx/include/cuda/__functional/equal_to_value.h
miscco: Yeah they are something to get used to ^^

in C++17 they directly translate to 
```
tempalte<class _Up,
         ::cuda::std::enable_if_t<::cuda::std::__is_cpp17_equality_comparable_v<_Tp, _Up>, int> = 0>
```

The issue is less with a single constraint, but if there are multiple it gets much simpler to write

thedevmystic: So, they basically are like a reflection system. Somewhat similar to Unreal's one but with a catch that we have to do it manually. I can work with that, i have experience with it :)

thedevmystic: @miscco just a question should we use _CCCL_TEMPLATE(typename _Up)? As typename is favoured over class in this case?

miscco: in libcu++ we generally use `class`, but inthis case both is fine I believe

thedevmystic:
> Thanks a lot for working on this 🎉
> 
> There is a lot of folklore to be satisfied, that is not obvious.
> 
> We are also missing tests, please give `libcudacxx/test/libcudacxx/std/utilities/function.objects/comparisons/equal_to.pass.cpp` a look
> 
> The new test should live in `libcudacxx/test/libcudacxx/cuda/utilities/function.objects/comparisons/equal_to_value.pass.cpp`
> 
> You can verify whether the code builds fine by running
> 
> ```
> ./ci/build_libcudacxx.sh
> cd build
> lit latest/libcudacxx/libcudacxx/test/libcudacxx/cuda/utilities/function.objects/comparisons/equal_to_value.pass.cpp
> ```

Hey, @miscco I have a quick question I've wrote the `equal_to_value.pass.cpp` where to I link it? Like the tests directory looks like a maze tbh. Do I have to include it somewhere or regexp does it automatically?

Another off-topic question from my side, why do we add `.pass` in the names like `test_name.pass.cpp`, this one is only for curiosity.

libcudacxx/include/cuda/__functional/equal_to_value.h
bernhardmgruber: CTAD does not work with aggregate init before C++20. In C++17 it needs either a ctor on the class (making it not an aggregate) or a deduction guide.

libcudacxx/include/cuda/__functional/equal_to_value.h
bernhardmgruber: If we don't find any surprises, we can make it public.

miscco:
> Hey, @miscco I have a quick question I've wrote the `equal_to_value.pass.cpp` where to I link it? Like the tests directory looks like a maze tbh. Do I have to include it somewhere or regexp does it automatically?

You dont, lit the `LLVM Infrastructure Tester` will pick the right files up and compile just that file. So no linking and so on needed. That allows us to get really small binaries with almost no other code coming in accidentally.

The test folder reflect the broad category of the features as given in the standard. We adopted it for all the nonstandard extensions that you can find the `libcudacxx/cuda/` subfolder

> Another off-topic question from my side, why do we add `.pass` in the names like `test_name.pass.cpp`, this one is only for curiosity.

lit goes starts with a default folder that you pass it and then compiles each test with the right file ending. So `.pass.cpp` are tests that we expect to compile and run fine. `fail.cpp` are tests that we expect to *not* compile and `runfail.cpp` are tests we expect to compile but fail at runtime.

Finally there are also `compile.cpp` which are tests that we do not intend to run because they only test static properties of the code, e.g that `counting_iterator<int>::value_type` names the right type

cub/test/catch2_test_device_select_flagged_if.cu
```diff
@@ -220,7 +213,7 @@ C2H_TEST("DeviceSelect::FlaggedIf does not change input and is stable",
 
   // Ensure that we did not overwrite other elements
   const auto boundary = out.begin() + num_selected_out[0];
-  REQUIRE(thrust::all_of(c2h::device_policy, boundary, out.end(), equal_to_default_t{}));
+  REQUIRE(thrust::all_of(c2h::device_policy, boundary, out.end(), cuda::__equal_to_value{}));
```
miscco: This might not do what you expect it to do. 

For trivial types this will default-initialize the aggregate `__equal_to_value` which will leave __value_ default-initialized, which again for trivial types does nothing. So you are effectively reading uninitialized memory

I would strongly suggest to implement this properly as a separate struct

thrust/thrust/detail/internal_functional.h
```diff
@@ -56,24 +57,9 @@ struct predicate_to_integral
...
-      : rhs(rhs)
-  {}
-
-  template <typename T1>
-  _CCCL_HOST_DEVICE bool operator()(const T1& lhs) const
-  {
-    return lhs == rhs;
-  }
-};
+template <typename T>
```
miscco: I love that alias templates inhibit CTAD

miscco (changes requested):

cub/test/catch2_test_device_select_flagged_if.cu
thedevmystic: I did the same thing first, exact same one. With `__equal_to_default` but then @bernhardmgruber said:
> Suggestion: this is really just the same as __equal_to_value{}, so I think there is no need for a dedicated functor. Please remove it.

See 4th review from him.

thrust/thrust/detail/internal_functional.h
thedevmystic: What to do then?

thedevmystic:
Thank you very much for your time explaining that!

thedevmystic:
Added test files.

libcudacxx/include/cuda/__functional/equal_to_value.h
miscco: its something that easily slips through, not a blocker, but if I find it we should change it

cub/test/catch2_test_device_select_flagged_if.cu
miscco: @bernhardmgruber is going to buy a lot of beverages in the next meeting ;)

thrust/thrust/detail/internal_functional.h
miscco: just keep it as is

libcudacxx/test/libcudacxx/cuda/utilities/function.objects/comparisons/equal_to_value.pass.cpp
```diff
@@ -0,0 +1,92 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+// <cuda/__functional/equal_to_value.h>
+
+// equal_to_value
+
+#include <cuda/__functional/equal_to_value.h>
+#include <cuda/std/cassert>
+#include <cuda/std/type_traits>
```
miscco: Nitpick: we conventionally write tests as a user would write them, so
```suggestion
#include <cuda/functional>
#include <cuda/std/cassert>
#include <cuda/std/type_traits>
```

libcudacxx/test/libcudacxx/cuda/utilities/function.objects/comparisons/equal_to_value.pass.cpp
```diff
@@ -0,0 +1,92 @@
...
+
+#include "test_macros.h"
+
+struct dummy
+{
+  friend constexpr bool operator==(const dummy&, const dummy&)
+  {
+    return true;
+  }
+};
```
miscco: I would rather name that ComparisonMayThrow and give it a value so you can test two that arent equal

libcudacxx/test/libcudacxx/cuda/utilities/function.objects/comparisons/equal_to_value.pass.cpp
```diff
@@ -0,0 +1,92 @@
...
+
+int main(int, char**)
+{
+  // +ve value
```
miscco: I cannot decipher that easily. Please err on the side of too verbose. 

Unit tests should be immediately obvious

libcudacxx/test/libcudacxx/cuda/utilities/function.objects/comparisons/equal_to_value.pass.cpp
```diff
@@ -0,0 +1,92 @@
...
+    assert(eq(0));
+    assert(!eq(1));
+  }
+
+  // -ve value
+  {
+    const cuda::__equal_to_value<int> eq(-1);
+    assert(eq(-1));
+    assert(!eq(0));
+  }
```
miscco: I dont think we need more comparisons than one.

libcudacxx/test/libcudacxx/cuda/utilities/function.objects/comparisons/equal_to_value.pass.cpp
```diff
@@ -0,0 +1,92 @@
...
+};
+
+int main(int, char**)
+{
+  // +ve value
+  {
+    const cuda::__equal_to_value<int> eq(1);
+    assert(eq(1));
+    assert(!eq(2));
```
miscco: We should static assert:

1. that the return value of the operation is bool
2. the operation is noexcept if it should be

libcudacxx/test/libcudacxx/cuda/utilities/function.objects/comparisons/equal_to_value.pass.cpp
```diff
@@ -0,0 +1,92 @@
...
+struct dummy
+{
+  friend constexpr bool operator==(const dummy&, const dummy&)
+  {
+    return true;
+  }
+};
+
+int main(int, char**)
+{
```
miscco, line 34: Conventionally, we use a 
````
__host__ __device__ constexpr bool test() {...}
````

That way we can easily centralize the tests and then in main we can do

```
test();
static_assert(test());
```

Because we also want to make sure all of this works at compile time

**Review by miscco — CHANGES_REQUESTED** [review]

File: libcudacxx/include/cuda/__functional/equal_to_value.h [hunk 25]
```diff
@@ -0,0 +1,66 @@
...
+  _Tp __value_;
+
+  explicit constexpr __equal_to_value(const _Tp& __value) noexcept(::cuda::std::is_nothrow_copy_constructible_v<_Tp>)
+      : __value_(__value)
+  {}
+
+  [[nodiscard]] _CCCL_API constexpr bool operator()(const _Tp& __lhs) const noexcept
+  {
+    return static_cast<bool>(__lhs == __value_);
+  }
```
miscco: This still needs to go, because that is handled below and the noexcept is wrong

miscco:
@thedevmystic I am a bad person and just merged `cuda::std::count`, please also replace `__count_compare_eq`

miscco (changes requested):

cub/test/catch2_test_device_select_flagged_if.cu
thedevmystic: Well then send a coffee to me too, preferably via speed-post. So, it doesn't get cold. Just kidding ;)

libcudacxx/include/cuda/__functional/equal_to_value.h
thedevmystic: I already change it :)

cub/test/catch2_test_device_select_flagged_if.cu
thedevmystic: @miscco then what to do, should I rewrite it or not?

libcudacxx/test/libcudacxx/cuda/utilities/function.objects/comparisons/equal_to_value.pass.cpp
thedevmystic: Okay then, long live the verbose side!!

From now on, I'll try to be verbose.

libcudacxx/test/libcudacxx/cuda/utilities/function.objects/comparisons/equal_to_value.pass.cpp
thedevmystic: Well okay.

libcudacxx/test/libcudacxx/cuda/utilities/function.objects/comparisons/equal_to_value.pass.cpp
thedevmystic: I did that be because `libcudacxx/test/libcudacxx/std/utilities/function.objects/comparisons/equal_to.pass.cpp` also directly uses something like,
```cpp
assert(f(32,32);
```

libcudacxx/test/libcudacxx/cuda/utilities/function.objects/comparisons/equal_to_value.pass.cpp
thedevmystic: Again I did that because `equal_to.pass.cpp` directly tests in `int main(int, char**)`. If that's not the case then I'll rewrite the tests.

thedevmystic:
> @thedevmystic I am a bad person and just merged `cuda::std::count`, please also replace `__count_compare_eq`

Well that's really off-topic to this PR's content. But if you want, then I'll go ahead and replace it. 

**EDIT:**
I ran `grep -rn "__count_compare_eq" .` and got zero results. Where are they defined anyways?

libcudacxx/include/cuda/__functional/equal_to_value.h
thedevmystic: Sorry missed that. Gotta clean that up.

libcudacxx/test/libcudacxx/cuda/utilities/function.objects/comparisons/equal_to_value.pass.cpp
thedevmystic: Why `ComparisonMayThrow`? Even if we add something like:
```cpp
struct ComparisonObject
{
  int value;

  __host__ __device__ friend bool operator==(const ComparisonObject& lhs, const ComparisonObject& rhs)
  {
    return lhs.value == rhs.value;
  }
};
```

I ain't seeing any reason to name it `MayThrow` part because why it will ever throw?

thedevmystic:
Things done:
- add suggestions to test file (improved and modularized).
- remove unnecessary function in `equal_to_value.h`.

cub/test/catch2_test_device_select_flagged_if.cu
bernhardmgruber: `__equal_to_value{}` only behaves weirdly if the type has a constructor, which it shouldn't IMO :) @miscco is right if we consider the current state of the PR.

libcudacxx/include/cuda/__functional/equal_to_value.h
```diff
@@ -0,0 +1,61 @@
...
+//! @tparam _Tp The type of the value to be compared.
+template <typename _Tp>
+struct __equal_to_value
+{
+  _Tp __value_;
+
+  explicit constexpr __equal_to_value(const _Tp& __value) noexcept(::cuda::std::is_nothrow_copy_constructible_v<_Tp>)
+      : __value_(__value)
+  {}
+
```
bernhardmgruber: Question: Why do we need a constructor? It breaks aggregate initialization in some C++ versions IIRC. I would drop it.

Did you encounter issues with CTAD because of a missing ctor? Did the deduction guide not work in that case?

thedevmystic: Without it we can get into issues. While aggregate initialization might be useful in some cases, current approach with ctor is generally more robust in C++17+. Like we can control the `noexcept` and `explicit`, without it we might get an `implicit` conversion.

If I'm missing feel free to correct me.

cub/test/catch2_test_device_select_flagged_if.cu
thedevmystic: So, i should add a explicit `__equal_to_default` functor, @bernhardmgruber?

libcudacxx/include/cuda/__functional/equal_to_value.h
bernhardmgruber: I don't see the issues, please elaborate.

`noexcept`-ness does not matter for the ctor, if we don't have a ctor.

How can we prevent an implicit conversion? If I write `__equal_to_value<int> pred(3.2);` the code compiles fine and truncates. `__equal_to_value<int> pred{3.2};` does error as it should, but would also error if we had no ctor. `__equal_to_value<int> pred(3.2)` would even fail to compile with C++17 instead of truncate. It truncates again in C++20. So arguably the aggregate is safer in C++17.

I don't see any advantage of the ctor. But if you really insist, you can keep it, but then please add a default argument so `__equal_to_value{}` will create a functor comparing to a default constructed `_Tp`. Also, if you keep the ctor, then please remove the deduction guide again, since it's not needed.

thedevmystic: Well I can do whatever you want! I'm not a type of person who clings to their code. Whatever suits the project better, I can do.

Just tell me, with or without ctor (CTAD), @bernhardmgruber.

bernhardmgruber: Definitely with CTAD. But that should work without a deduction guide if you have a ctor.

thedevmystic:
Remove deduction guide and add default ctor. Now, `cuda::__equal_to_value{}` checks for equality against the default value.

miscco:
/ok to test [89a0bf3](https://github.com/NVIDIA/cccl/pull/7496/commits/89a0bf3479dd9e12cc5eb4621ee11ec8b5f1c6f7)

cub/test/catch2_test_device_select_flagged_if.cu
miscco: I just went without it, all current uses were in tests, so there we can just do `equal_value_to{T{}}`

miscco:
@thedevmystic I have pushed some changes and started CI

libcudacxx/include/cuda/std/__pstl/count.h
```diff
@@ -75,7 +63,7 @@ count([[maybe_unused]] const _Policy& __policy, _InputIterator __first, _InputIt
     const auto __count = ::cuda::std::distance(__first, __last);
     return __dispatch(
       __policy,
-      ::cuda::transform_iterator{::cuda::std::move(__first), __count_compare_eq<_Tp>{__value}},
+      ::cuda::transform_iterator{::cuda::std::move(__first), ::cuda::equal_to_value<_Tp>{__value}},
```
miscco, line 68: This should be fine because bool converts to integral just fine, we might want to bench

miscco:
/ok to test [f5b58bd](https://github.com/NVIDIA/cccl/pull/7496/commits/f5b58bd72f2245c4339b64f81f9835aac29d710a)[f5b58bd](https://github.com/NVIDIA/cccl/pull/7496/commits/f5b58bd72f2245c4339b64f81f9835aac29d710a)

miscco:
/ok to test [f5b58bd](https://github.com/NVIDIA/cccl/pull/7496/commits/f5b58bd72f2245c4339b64f81f9835aac29d710a)

libcudacxx/test/libcudacxx/cuda/utilities/function.objects/comparisons/equal_to_value.pass.cpp
```diff
@@ -0,0 +1,128 @@
...
+  // built-in types
+  constexpr cuda::equal_to_value<int> eq(42);
+  static_assert(noexcept(eq(42)) == true, "");
+  static_assert(eq(42) == true, "");
+
+  // user-defined types
+  constexpr ComparisonObject obj{42};
+  constexpr cuda::equal_to_value<ComparisonObject> eq_obj(obj);
+  static_assert(noexcept(eq_obj(obj)) == true, "");
```
miscco: Nitpick: This should test that a ppotentially throwing type is not wrapped in noexcept
```suggestion
  static_assert(!noexcept(eq_obj(obj)) == true, "");
```

libcudacxx/test/libcudacxx/cuda/utilities/function.objects/comparisons/equal_to_value.pass.cpp
```diff
@@ -0,0 +1,128 @@
...
+  static_assert(eq(c) == false, "");
+
+  return true;
+}
+
+// Test suit for CTAD
+__host__ __device__ constexpr bool test_ctad()
+{
+  // built-in types
+  constexpr cuda::equal_to_value eq(42);
```
miscco: Critical: We do not want to use `constexpr` variables here

Everything should just be plain variables and the compile time tests come through 
```
  test();
  static_assert(test());
```

libcudacxx/test/libcudacxx/cuda/utilities/function.objects/comparisons/equal_to_value.pass.cpp
```diff
@@ -0,0 +1,128 @@
...
+__host__ __device__ constexpr bool test()
+{
+  return test_numeric_types() && test_heterogeneous_comparisons() && test_user_defined_types() && test_ctad()
+      && test_noexcept();
+}
+
+int main(int, char**)
+{
+  test();
+  static_assert(test(), "");
```
miscco: We have C++17 only so
```suggestion
  static_assert(test());
```

The other occurences with `, ""` are remnants of us supporting C++11 / C++14

libcudacxx/test/libcudacxx/cuda/utilities/function.objects/comparisons/equal_to_value.pass.cpp
```diff
@@ -0,0 +1,128 @@
...
+  static_assert(eq_obj(obj) == true, "");
+
+  return true;
+}
+
+// Run all test suits
+__host__ __device__ constexpr bool test()
+{
+  return test_numeric_types() && test_heterogeneous_comparisons() && test_user_defined_types() && test_ctad()
+      && test_noexcept();
```
miscco: This can be simplified to:

```suggestion
  test_numeric_types();
  test_heterogeneous_comparisons();
  test_user_defined_types();
  test_ctad();
  test_noexcept();
  
  return true;
```

libcudacxx/test/libcudacxx/cuda/utilities/function.objects/comparisons/equal_to_value.pass.cpp
```diff
@@ -0,0 +1,128 @@
...
+  static_assert(noexcept(eq(42)) == true, "");
+  static_assert(eq(42) == true, "");
+
+  // user-defined types
+  constexpr ComparisonObject obj{42};
+  constexpr cuda::equal_to_value<ComparisonObject> eq_obj(obj);
+  static_assert(noexcept(eq_obj(obj)) == true, "");
+  static_assert(eq_obj(obj) == true, "");
+
+  return true;
```
miscco: Those subfunctions can be plain `void` functions. We only need the umbrella test() function to return true
```suggestion
```

miscco (changes requested):

libcudacxx/test/libcudacxx/cuda/utilities/function.objects/comparisons/equal_to_value.pass.cpp
thedevmystic: Great idea, tbh. We can drop the boolean and wrap it up in one function.

libcudacxx/test/libcudacxx/cuda/utilities/function.objects/comparisons/equal_to_value.pass.cpp
thedevmystic: Okay!

libcudacxx/test/libcudacxx/cuda/utilities/function.objects/comparisons/equal_to_value.pass.cpp
thedevmystic: When we remove `constexpr` from variables we get:
```
equal_to_value.pass.cpp:88:17: error: static assertion expression
      is not an integral constant expression
   88 |   static_assert(eq_obj(ComparisonObject{43}) == false);
      |                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
equal_to_value.pass.cpp:28:25: note: read of non-constexpr variable
      'eq_obj' is not allowed in a constant expression
   28 |     return lhs.value == rhs.value;
      |                         ^
../../../../../../include/cuda/__functional/equal_to_value.h:49:30: note:
      in call to 'operator==(ComparisonObject{43}, eq_obj.__value_)'
   49 |     return static_cast<bool>(__lhs == __value_);
      |                              ^~~~~~~~~~~~~~~~~
equal_to_value.pass.cpp:88:17: note: in call to
      'eq_obj.operator()<ComparisonObject, true, 0>(ComparisonObject{43})'
   88 |   static_assert(eq_obj(ComparisonObject{43}) == false);
      |                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
equal_to_value.pass.cpp:85:30: note: declared here
   85 |   const cuda::equal_to_value eq_obj(obj);
      |                              ^
equal_to_value.pass.cpp:97:17: error: static assertion expression
      is not an integral constant expression
   97 |   static_assert(eq(42) == true);
      |                 ^~~~~~~~~~~~~~

... Similar errors.
```

libcudacxx/test/libcudacxx/cuda/utilities/function.objects/comparisons/equal_to_value.pass.cpp
thedevmystic: This also fails:
```
equal_to_value.pass.cpp:102:17: error: static assertion failed due
      to requirement '!noexcept(eq_obj(obj)) == true'
  102 |   static_assert(!noexcept(eq_obj(obj)) == true);
      |                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

libcudacxx/test/libcudacxx/cuda/utilities/function.objects/comparisons/equal_to_value.pass.cpp
miscco: Yeah you should not static assert in the test

just a plain `assert`. Note that assert is not callable at compile time, so if the assert would trigger then the compilation fails

libcudacxx/test/libcudacxx/cuda/utilities/function.objects/comparisons/equal_to_value.pass.cpp
```diff
@@ -0,0 +1,128 @@
...
+};
+
+// Test suit for numeric types
+__host__ __device__ constexpr bool test_numeric_types()
+{
+  // integral values
+  {
+    constexpr cuda::equal_to_value<int> eq(1);
+    static_assert(eq(1) == true, "");
+    static_assert(eq(2) == false, "");
```
miscco: Tests need to be obvious at a glance, so please use longer names
```suggestion
    cuda::equal_to_value<int> eq_to_one{1};
    assert(eq_to_one(1));
    assert(!eq_to_one(2));
```

thedevmystic: Do we just need this there or here too?

```cpp
  constexpr cuda::equal_to_value<double> eq_to_pi(3.14);
  static_assert(eq_to_pi(3.14));
  static_assert(!eq_to_pi(2.71));
```

libcudacxx/test/libcudacxx/cuda/utilities/function.objects/comparisons/equal_to_value.pass.cpp
thedevmystic: Okay 👍

libcudacxx/test/libcudacxx/cuda/utilities/function.objects/comparisons/equal_to_value.pass.cpp
miscco: everywhere

libcudacxx/test/libcudacxx/cuda/utilities/function.objects/comparisons/equal_to_value.pass.cpp
```diff
@@ -0,0 +1,125 @@
...
+  assert(eq(c) == false);
+}
+
+// Test suit for CTAD
+__host__ __device__ constexpr void test_ctad()
+{
+  // built-in types
+  const cuda::equal_to_value eq(42);
+  auto is_same = ::cuda::std::is_same_v<decltype(eq), const cuda::equal_to_value<int>>;
+  assert(is_same);
```
miscco: Ah sorry, that type check should have stayed. Its a bit of a mess how to test, but we were actually bitten hard by it when the compiler helpfully constant folded away a function call that would segfault at runtime and we needed a lot of time to figure it out

libcudacxx/test/libcudacxx/cuda/utilities/function.objects/comparisons/equal_to_value.pass.cpp
```diff
@@ -0,0 +1,125 @@
...
+  test_ctad();
+  test_noexcept();
+
+  return true;
+}
+
+int main(int, char**)
+{
+  test();
+  assert(test());
```
miscco: This needs to be `static_assert(test());` 

Otherwise the compiler will not evaluate at compile time

miscco (approved):
I pushed minor fixes for the remaining issues

miscco:
/ok to test [d999fcc](https://github.com/NVIDIA/cccl/pull/7496/commits/d999fcce0aa0bd0d5531545bf659880da57ebab1)

NaderAlAwar (approved):
Benchmarks look good

thedevmystic:
So, we are moving it to public interface. Okay :)

libcudacxx/test/libcudacxx/cuda/utilities/function.objects/comparisons/equal_to_value.pass.cpp
thedevmystic: Sorry, forgot to rename when i replaced every static_assert with assert.

libcudacxx/test/libcudacxx/cuda/utilities/function.objects/comparisons/equal_to_value.pass.cpp
thedevmystic: Yeah, that's sucks. It was the same error that I got when i tried to simply move it to assert.

libcudacxx/test/libcudacxx/cuda/utilities/function.objects/comparisons/equal_to_value.pass.cpp
thedevmystic: @miscco you approved the changes, do I still have to rename them?

miscco:
/ok to test [099e1c6](https://github.com/NVIDIA/cccl/pull/7496/commits/099e1c6e603cf1ada34ee5812a709ab32de36695)

bernhardmgruber (approved):

miscco:
Thanks a lot @thedevmystic for implementing this great unification.

miscco:
/ok to test [d1914e5](https://github.com/NVIDIA/cccl/pull/7496/commits/d1914e5ca306511f867baaefd836239f8f7d72fa)

libcudacxx/include/cuda/__functional/equal_to_value.h
```diff
@@ -0,0 +1,57 @@
...
+    ::cuda::std::is_nothrow_copy_constructible_v<_Tp>)
+      : __value_(__value)
+  {}
+
+  _CCCL_TEMPLATE(class _Up)
+  _CCCL_REQUIRES(::cuda::std::__is_cpp17_equality_comparable_v<_Tp, _Up>)
+  [[nodiscard]] _CCCL_API constexpr bool operator()(const _Up& __lhs) const
+    noexcept(::cuda::std::__is_cpp17_nothrow_equality_comparable_v<_Tp, _Up>)
+  {
+    return static_cast<bool>(__lhs == __value_);
```
bernhardmgruber, line 49: Q: Some CI jobs are reporting:
```
equal_to_value.h:49:33: error: comparison between signed and unsigned integer expressions [-Werror=sign-compare]
```
I think we should suppress this warning here. @miscco what do you think?

miscco: We should actually just make sure the tests do the right thing. I believe the comparison there is valid and it is a potential bug, so I would not like to unilaterally suppress the warning in product code

miscco:
/ok to test [cac1c45](https://github.com/NVIDIA/cccl/pull/7496/commits/cac1c45401a16da68603bba821c585552ff91d65)

miscco:
/ok to test [2847125](https://github.com/NVIDIA/cccl/pull/7496/commits/28471251ce79bdb0e64b51d3cba38df996679e60)

miscco:
Thanks again for working on this.

welcome to the bottom of the stack, where all is joy 🎉

thedevmystic:
Thanks! Love to be at the bottom building things up!

Issue #7498, opened 2026-02-04
cuda.compute: Stateful operators are slow because they almost always need recompilation


shwina:
In https://github.com/NVIDIA/cccl/pull/7008, we added support for stateful ops in `cuda.compute`. However, it has been found to be slow (see for example https://github.com/scikit-hep/awkward/pull/3814#issuecomment-3835661158).

This is because, in practice, stateful ops are used something like this:

```python
def library_func(x, y, array):

    def stateful_op(x):
        # reference `array` here

    cuda.compute.transform(..., stateful_op, ...)
```

Above, we hope that the first call to `library_func` will trigger compilation of `stateful_op`, but successive calls will not.

But it is compiled on every call. This is because on every call, the op references a different global named `array`. The way `numba-cuda` works is that the pointer of the globally referenced `array` is captured as a constant before compilation. When the referenced array changes, it's treated as a different constant leading to recompilation.

## The solution

The solution is to use CCCL.c's native support for stateful operations. Stateful operations pass state to the library via an additional `void*` state parameter; so that stateful operators have the signature:

```c++
void (void* state, void* arg1, ..., void* argN, void* result_ptr)
```

(see [here](https://github.com/NVIDIA/cccl/blob/7942d8c8aa9b1dc85a9d2912b4d056adebabbb81/c/parallel/src/jit_templates/templates/operation.h#L62-L64))

So, on the Python side it's our job to take a stateful operator like:

```python
# example: stateful unary operator

def foo(x):
    return x + y[0]  # `y` is a global/closure
```

And translate it to something like:

```c++
void type_erased_f(void* state, void* x, void* result_ptr) {}
```

(for ODR reasons, all arguments must be declared `void*`).

To achieve this, we first do some Python-side AST manipulation to transform a function referencing external arrays into one that takes them as explciit arguments:

  ```python
  # given:
  def foo(x):
      return x + y[0]

  # transform it to:
  def foo(x, y):
      return x + y
  ```

We then use numba to define and compile the wrapper function with the correct signature (`void(void* state, void* x, void* result_ptr)`). Within the wrapper, we interpret the `void*` as a packed array of pointers to externally referenced arrays and unpack them into array objects. We then pass these to the transformed function above.

## State updates

On every invocation, the state (which holds pointers to externally referenced arrays) must be updated to hold the correct pointers. This introduces an API problem for `cuda.compute`. Consider the (object-based) API of `reduce_into`:

```python
# step 1: create algorithm object
[...]

PR #7500, opened 2026-02-04
cuda.compute: Use native CCCL.c support for stateful ops

shwina:
## Description

Closes https://github.com/NVIDIA/cccl/issues/7498.

See the issue above for design/rationale.

### Performance

As described in the issue, due to the overhead of inspecting/extracting state, there's an ~1us overhead introduced to every invocation. This can and should be eliminated for stateless operations.

## Checklist

### DIFF
--- a/python/cuda_cccl/cuda/compute/_jit.py
+++ b/python/cuda_cccl/cuda/compute/_jit.py
@@ -863,7 +863,7 @@ def _transform_function_ast(func: Callable, state_names: List[str]) -> Callable:
 
     Then the transformed function will be:
 
-        def func(x, state): return x + state[0]
+        def func(state, x): return x + state[0]
 
     Args:
         func: The original function


python/cuda_cccl/cuda/compute/_jit.py
```diff
@@ -692,36 +667,437 @@ def compile_iterator(it, io_kind: str):
...
+        operator_type=OpKind.STATELESS,
+        name=wrapped_op.__name__,
+        ltoir=ltoir,
+        state_alignment=1,
+        state=None,
+    )
+
+
+class _StatelessOp(OpAdapter):
+    """Adapter for stateles callables."""
```
shwina: typo

python/cuda_cccl/cuda/compute/_jit.py
```diff
@@ -692,36 +667,437 @@ def compile_iterator(it, io_kind: str):
...
+    Transform a function to add state arrays captured as globals or closures
+    as explicit parameters.
+
+    For example, if the function is:
+
+        def func(x): return x + state[0]  # state is a global array
+
+    Then the transformed function will be:
+
+        def func(x, state): return x + state[0]
```
shwina: Maybe `state` should go first.

python/cuda_cccl/cuda/compute/_jit.py
```diff
@@ -692,36 +667,437 @@ def compile_iterator(it, io_kind: str):
...
+        input_types: Tuple of TypeDescriptors for regular input arguments
+        state_arrays: Tuple of device arrays containing state
+        output_type: Optional TypeDescriptor for return value (inferred if None)
+
+    Returns:
+        Compiled Op object for C++ interop
+    """
+    import struct
+
     from . import types as cccl_types
```
shwina: Do these imports need to be here?

python/cuda_cccl/cuda/compute/op.py
```diff
@@ -101,6 +98,8 @@ def make_op_adapter(op) -> OpAdapter:
     Returns:
         A value with appropriate subtype of _BaseOp
     """
+    from ._jit import to_jit_op_adapter
```
shwina, line 101: Try to move this out because it's a non-trivial cost.

python/cuda_cccl/cuda/compute/algorithms/_reduce.py
shwina, line 1: What about other algorithms like `merge_sort`?

python/cuda_cccl/cuda/compute/_caching.py
```diff
@@ -137,22 +144,16 @@ def register(self, type_: type, key_function: Callable[[Any], Hashable]) -> None
...
     if hasattr(value, "py_func") and callable(value.py_func):
         return CachableFunction(value.py_func)
-    elif isinstance(value, DeviceArrayLike):
-        return _hash_device_array_like(value)
+    elif is_device_array(value):
+        # Ops with device arrays in globals/closures will be handled
+        # by stateful op machinery, which enables updating the state
+        # (pointers). Thus, we only cache on the dtype and shape of
+        # the referenced array, but not its pointer.
+        return (get_dtype(value), get_shape(value))
```
NaderAlAwar, line 156: Question: Remind me again why we cache on shape?

python/cuda_cccl/cuda/compute/_jit.py
```diff
@@ -692,36 +668,437 @@ def compile_iterator(it, io_kind: str):
...
+        Returns:
+            List of (name, array) tuples for detected device arrays
     """
-    Compile a host-side function (e.g., iterator advance) into a C function pointer.
+    state_arrays = []
+    code = func.__code__
+
+    for name in code.co_names:
+        val = func.__globals__.get(name)
+        if val is not None and hasattr(val, "__cuda_array_interface__"):
```
NaderAlAwar: Important: use the newly added `is_device_array` instead of hasattr

python/cuda_cccl/cuda/compute/_caching.py
shwina: Here's an (admittedly made up) example for when it could matter:

```python
import numpy as np, cupy as cp, cuda.compute as cc

def make_op(arr):
    def op(x):
        return x > len(arr)
    return op

d_in = cp.asarray([1, 2, 3])
d_out = cp.empty_like(d_in, dtype=bool)
op1 = make_op(cp.empty(1))  # len(arr) == 1
op2 = make_op(cp.empty(2))  # len(arr) == 2

>>> cc.unary_transform(d_in, d_out, op1, len(d_in)); print(d_out)
[False  True  True]

>>> cc.unary_transform(d_in, d_out, op2, len(d_in)); print(d_out)
[False False  True]

# without including the shape in the hash, the second result would also be `[False, True, True]`
```

shwina: Added this as a unit test.

python/cuda_cccl/cuda/compute/_jit.py
shwina: Done

python/cuda_cccl/cuda/compute/op.py
shwina: Unfortunately, `_jit` needs to import `OpAdapter` to subclass it. I think we should make it a protocol so that `_jit.py` won't need to import from here.

shwina: xref: https://github.com/NVIDIA/cccl/issues/7503

python/cuda_cccl/cuda/compute/algorithms/_reduce.py
shwina: Done

python/cuda_cccl/cuda/compute/_jit.py
shwina: No (fixed)

python/cuda_cccl/cuda/compute/_jit.py
shwina: Done

PR #7509, opened 2026-02-05
Use modern __syncthreads_or primitive

gonidelis:
Somehow this slipped through and I used an old version (sm20 dated) of `__syncthreads_or()`. 

It's coming from an internal CUDA header that looks like this.

https://github.com/zspace/system.cuda-12.6/blob/b3d92c837713411301ae5df45b0c7758df7d931e/include/sm_20_intrinsics.hpp#L90-L93

Anyways, this PR uses [the proper modern ](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html?highlight=__syncthreads_or#synchronization-functions)`syncthreads_or`


bernhardmgruber (approved):

pauleonix (approved):
LGTM

PR #7511, opened 2026-02-05
Simplify `DeviceTransform` vectorized policy

bernhardmgruber:
I removed the fallback values for prefetching, since we can just take the values from the neighboring prefetch policy now.

- [x] No SASS changes in `cub.bench.transform.babelstream.base` on SM `75;80;90;100`


bernhardmgruber:
> Looks much clearer

I think the old version was a remnant from when we still used a `variant` to hold one of the policies, so if we held the vectorized policy, we would need to have a prefetch policy inside. That's no longer needed.

Issue #7508, opened 2026-02-05
[DOC]: Ambiguity in .devcontainer README's Docker install step for WSL


acosmicflamingo:
### Is this a duplicate?

- [x] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cccl/issues) for this bug and that I agree to the [Code of Conduct](CODE_OF_CONDUCT.md)

### Is this for new documentation, or an update to existing docs?

Update

### Describe the incorrect/future/missing documentation

While setting up my new Windows workstation and following the .devcontainer README instructions, I found that the way it's currently written implies that installing the VSCode Remote Development Extension Pack would handle installing Docker on its own when in fact, it has to be installed. Perhaps the documentation was written to mean that Docker would automatically be picked up after the remote extension pack was installed.

I'm probably overanalyzing the instructions, but assuming there are others like me that would read it as such, I figured a change like this could be beneficial to merge to main:
```diff
diff --git a/.devcontainer/README.md b/.devcontainer/README.md
index ba6859312..f069ed22b 100644
--- a/.devcontainer/README.md
+++ b/.devcontainer/README.md
@@ -79,9 +79,10 @@ Apart from the initial setup of WSL, the process for using CCCL's Dev Containers
 - Windows OS that supports WSL 2 (Windows 11 or newer)
 - [Windows Subsystem for Linux v2 (WSL 2)](https://learn.microsoft.com/en-us/windows/wsl/install)
 - [Visual Studio Code](https://code.visualstudio.com/) (installed on Windows host)
+- [Docker](https://docs.docker.com/engine/install/)
+    - Required for VSCode / Dev Container integration
 - [VSCode Remote Development Extension Pack](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.vscode-remote-extensionpack) (installed on Windows host)
     - Includes [Dev Containers](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers) and [WSL](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-wsl) extensions
-- [Docker](https://docs.docker.com/engine/install/) - (Will be installed automatically by the Remote Development extension)
 
 #### GPU Prerequisites (only needed for executing tests that require a GPU)
 - Supported NVIDIA GPU
```

[...]

PR #7516, opened 2026-02-05
Clarify Docker install step for WSL in .devcontainer README

acosmicflamingo:
## Description

Addresses https://github.com/NVIDIA/cccl/issues/7508.

For WSL users, the .devcontainers' README is written in a way where it sounds like installing the VSCode Remote Development Extension Pack will automatically handle the Docker step, when in fact it will only handle automatically finding an already-installed Docker.

This PR clarifies the wording of the README so it explicitly states that Docker needs to be installed before installing the VSCode Remote Development Extension Pack.

## Checklist

### DIFF
--- a/.devcontainer/README.md
+++ b/.devcontainer/README.md
@@ -79,8 +79,8 @@ Apart from the initial setup of WSL, the process for using CCCL's Dev Containers
 - Windows OS that supports WSL 2 (Windows 11 or newer)
 - [Windows Subsystem for Linux v2 (WSL 2)](https://learn.microsoft.com/en-us/windows/wsl/install)
 - [Visual Studio Code](https://code.visualstudio.com/) (installed on Windows host)
-- [Docker](https://docs.docker.com/engine/install/)
-    - Required for VSCode / Dev Container integration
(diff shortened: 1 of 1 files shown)


acosmicflamingo:
I wonder if it'd also be beneficial to steer WSL users to installing the Ubuntu distro. Based on these [docs](https://docs.nvidia.com/cuda/wsl-user-guide/index.html) and the CUDA toolkit download center, it seems like Ubuntu is definitely the way to go.image

jrhemstad:
> I wonder if it'd also be beneficial to steer WSL users to installing the Ubuntu distro. Based on these [docs](https://docs.nvidia.com/cuda/wsl-user-guide/index.html) and the CUDA toolkit download center, it seems like Ubuntu is definitely the way to go.

It's always a delicate balance between trying to provide complete instructions in our own documentation vs deferring to other more authoritative documentation. As you've seen, the risk is that if we try and provide complete, detailed instructions, it's likely to eventually get stale. I think for this kind of guidance on what distro to select is better left to the docs at https://docs.nvidia.com/cuda/wsl-user-guide/index.html

acosmicflamingo:
Gotcha; yes I agree. Thanks for the context 😊

PR #7520, opened 2026-02-05
Fix pnputil to only restart NVIDIA display adapters [skip-matrix] [skip-vdc]

kkraus14:
## Summary
- The `pnputil /disable-device /class Display` and `pnputil /enable-device /class Display` commands in the GPU driver install script disable/enable **all** display devices, including non-NVIDIA adapters like the QEMU VGA Graphics Adapter.
- This replaces the blanket `/class Display` approach with a targeted `Get-PnpDevice` filter that only cycles NVIDIA display adapters by instance ID.

## Test plan
- [ ] Verify on a Windows CI runner with both NVIDIA GPU and QEMU VGA that only the NVIDIA adapter is restarted
- [ ] Verify driver mode changes still take effect correctly after the targeted restart


alliepiper:
(BTW -- these tags only work when they're in the commit message of the latest branch commit: `[skip-matrix] [skip-vdc]`)

PR #7530, opened 2026-02-06
Move delay constructor policies somewhere central

bernhardmgruber:
Since we are going to reuse them for a bunch of CUB algorithms, this PR extracts some delay constructor policies and stuff and moves it to a more central header.

Let me know if the new place is good. I also considered `tuning/common.cuh`.

### DIFF
--- a/c/parallel/src/radix_sort.cu
+++ b/c/parallel/src/radix_sort.cu
@@ -255,6 +255,7 @@ struct __align__({3}) values_storage_t {{
 {4}
 using device_radix_sort_policy = {5};
 using namespace cub;
+using namespace cub::detail;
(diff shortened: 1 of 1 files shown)


bernhardmgruber:
@griwes this probably affects you for the scan rewrite as well!

griwes:
> @griwes this probably affects you for the scan rewrite as well!

Aye, I have an almost identical patch set in my working branch, except with a separate header.

bernhardmgruber:
> > @griwes this probably affects you for the scan rewrite as well!
> 
> Aye, I have an almost identical patch set in my working branch, except with a separate header.

Great. Yours looks even better! Closing this PR in favor of: #7565

PR #7561, opened 2026-02-07
Remove recursion from __internal_is_address_from

dkolsen-pgi:
## Description

When running on pre-Hopper GPUs, a call to `__internal_is_address_from(ptr, cluster_shared)` would simply make a recursive call to `__internal_is_address_from(ptr, shared)`.  The recursion would stop there; there was no infinite recursion or large stack sizes.  But when compiling the GPU code with debug information and no optimization (`nvcc -G`), the recursive call would remain in the PTX and that would cause either ptxas or nvlink to be unable to calculate the correct stack size for the kernel.  That could result in a failed kernel if the default stack size is too small.

Avoid this problem by removing the recursive call in `__internal_is_address_from`.  Instead move the `case address_space::shared:` code to just after `case address_space::cluster_shared:` and have `case address_space::cluster_shared:` `[[fallthrough]]` to `case address_space::shared:` on pre-Hopper GPUs.

This fixes some stdpar tests when compiled with `nvc++ -g -stdpar` on pre-Hopper GPUs.  It fixes some CUDA applications compiled with `nvcc -G`, though I don't have any real-world examples.

Also fixes NVBug: 5880331

## Checklist

- [X] The documentation is up to date with these changes.

### DIFF
--- a/libcudacxx/include/cuda/__memory/address_space.h
+++ b/libcudacxx/include/cuda/__memory/address_space.h
@@ -169,7 +169,7 @@ enum class address_space
                       "  selp.u32 %0, 1, 0, p;\n\t"
                       "}\n\t" : "=r"(__ret) : "l"(__ptr));
          return static_cast<bool>(__ret);),
-        (/* Fall through to `case shared:` */;))
+        ([[fallthrough]]; /* to `case shared:` */))
 #  else // ^^^ _CCCL_CUDA_COMPILER(NVCC, <, 12, 3) || _CCCL_CUDA_COMPILER(NVRTC, <, 12, 3) ^^^ /
         // vvv !_CCCL_CUDA_COMPILER(NVCC, <, 12, 3) && !_CCCL_CUDA_COMPILER(NVRTC, <, 12, 3) vvv
       NV_IF_ELSE_TARGET(


dkolsen-pgi:
Here is an example of the problem:
```c++
#include <stdio.h>
#include <cuda/__memory/address_space.h>

__global__ void kernel(void *p) {
  if (cuda::device::__internal_is_address_from(p, cuda::device::address_space::cluster_shared)) {
    printf("Wrong answer!\n");
    asm("trap;");
  }
}

int main(int argc, char** argv) {
  kernel<<<32,32>>>(argv);
  if (cudaDeviceSynchronize() != cudaSuccess) {
    printf("Kernel failed\n");
  }
}
```

```
$ nvcc -arch=native -G -I/proj/cuda/cccl/main/libcudacxx/include test.cu
ptxas warning : Stack size for entry function '_Z6kernelPv' cannot be statically determined
```
The warning indicates that the kernel stack size might be wrong.  This test program is small enough that the default stack size isn't an actual problem.  But we have seen this warning followed by kernel failures due to stack overflow in larger programs.

I know that this change avoids the warning from ptxas or nvlink.  But I can't easily do rigorous regression testing, so I would appreciate someone on the CCCL team doing whatever testing is appropriate for this.

bernhardmgruber:
pre-commit.ci autofix

libcudacxx/include/cuda/__memory/address_space.h
```diff
@@ -203,7 +180,31 @@ enum class address_space
            _CCCL_ASSUME(__p); //
          } //
          return __p;),
-        (return ::cuda::device::__internal_is_address_from(__ptr, address_space::shared);))
+        (/* Fall through to `case shared:` */;))
+#  endif // ^^^ !_CCCL_CUDA_COMPILER(NVCC, <, 12, 3) && !_CCCL_CUDA_COMPILER(NVRTC, <, 12, 3) ^^^
+    }
+      [[fallthrough]];
```
davebayer:
```suggestion
```

libcudacxx/include/cuda/__memory/address_space.h
```diff
@@ -203,7 +180,31 @@ enum class address_space
            _CCCL_ASSUME(__p); //
          } //
          return __p;),
-        (return ::cuda::device::__internal_is_address_from(__ptr, address_space::shared);))
+        (/* Fall through to `case shared:` */;))
```
davebayer:
```suggestion
        ([[fallthrough]]; /* to `case shared:` */))
```

libcudacxx/include/cuda/__memory/address_space.h
```diff
@@ -192,7 +169,7 @@ enum class address_space
                       "  selp.u32 %0, 1, 0, p;\n\t"
                       "}\n\t" : "=r"(__ret) : "l"(__ptr));
          return static_cast<bool>(__ret);),
-        (return ::cuda::device::__internal_is_address_from(__ptr, address_space::shared);))
+        (/* Fall through to `case shared:` */;))
```
davebayer:
```suggestion
        ([[fallthrough]]; /* to `case shared:` */))
```

PR #6811, opened 2025-11-28
Integrate decoupled lookahead warpspeed scan

bernhardmgruber:
This integrates a MVP version of the warpspeed scan by @ahendriksen.

Here is a chronological **summary** of the many commits and MRs on GitLab that went into this PR:
* integrate prototype from @ahendriksen 
* update license
* integrate cub warp reductions
* support different input, output and accumulator types
* support different scan operators
* support any problem size and handle partial tiles
* enable exclusive scan and providing an initial value
* support unaligned input and output
* adjust number of stages based on max available smem
* enable PDL
* fallback when compiling for older PTX ISAs and SM arches
* ensure compilation on older CTKs and SM arches
* fallback for non-contiguous inputs, non trivially copyable types, ...
* move hard coded parameters into tuning policies
* ensure forward compatibility of binaries by ensuring at least one stage can run with <=48KiB SMEM, fallback otherwise
* use a single kernel symbol for old and new implementation to limit binary size impact
* rudimentary tuning by adjusting items per thread and number of stages based on input types
* extend support to all compilers supported by CCCL
* partial support for CCCL.C and JIT compilation (needs a bit of follow-up work)
* lots of cleanup
* tons of fixes for corner cases uncovered by CUB & Thrust tests

We are not entirely ready yet, here are a few open issues:

- [x] `SmemResource<InT> smemInOut` needs to be aligned for `OutputT` as well
- [x] ~~Improve performance on small problem sizes~~
- [x] Handle types with non-power-of-2 sizes. This is actively uncovering more bugs.
- [x] ~~Make the tests work under compute-sanitizer (still reports out of memory access, e.g. on `cub.test.device.scan.lid_0.types_2`)~~ Turns out `compute-sanitizer` has a bug when a masked bulk copy SMEM -> GMEM is used, reporting false positives.
- [x] `cccl.c.parallel.test.scan` is stuck and does not terminate
- [x] Update benchmark before merging
- [x] Create an issue for tracking the regression on small problem sizes we are taking with this PR.

The following unfinished tasks will be postponed beyond this PR:

- #6821
- #6800
- #6664
- #7440
- #7512

`cub.bench.scan.exclusive.sum.base` on B200:
```
## [0] NVIDIA B200

|  T{ct}  |  OffsetT{ct}  |  Elements{io}  |   Ref Time |   Ref Noise |   Cmp Time |   Cmp Noise |          Diff |   %Diff |  Status  |
[...]

PR #7565, opened 2026-02-08
Implement the new tuning API for `DeviceScan`

### LINKED ISSUE #6821: Make CCCL.C work again
This requires us to write and extend all the policy wrappers.

### LINKED ISSUE #7476: Refactor cccl.c scan to use tuning API
This issue can be closed by updating the `cub::DeviceScan` C++ API to use the new tuning policy API and refactoring cccl.c scan to use that.

### LINKED ISSUE #7521: Implement the new tuning API for `DeviceScan`

griwes:
## Description

Resolves #7521
Resolves #7476
Resolves #6821 

Ready for review, still planning to do SASS inspection in some crucial places.

Sidenote: this exact type of task seems to fit Codex really, really well.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -511,7 +511,7 @@ Does your project use CCCL? [Open a PR to add your project to this list!](https:
 - [Qiskit](https://github.com/Qiskit/qiskit-aer) - High performance simulator for quantum circuits
 - [QUDA](https://github.com/lattice/quda) - Lattice quantum chromodynamics (QCD) computations
 - [RAFT](https://github.com/rapidsai/raft) - Algorithms and primitives for machine learning
-- [SGLang](https://github.com/sgl-project/sglang) - LLM serving framework 
+- [SGLang](https://github.com/sgl-project/sglang) - LLM serving framework
 - [TensorFlow](https://github.com/tensorflow/tensorflow) - End-to-end platform for machine learning
 - [TensorRT](https://github.com/NVIDIA/TensorRT) - Deep learning inference
 - [TensorRT-LLM](https://github.com/NVIDIA/TensorRT-LLM) - Optimized LLM inference


c/parallel/src/scan.cu
```diff
@@ -384,31 +389,39 @@ CUresult cccl_device_scan(
...
+        indirect_arg_t{d_out},
+        indirect_arg_t{op},
+        init_arg,
+        static_cast<OffsetT>(num_items),
+        stream,
+        *static_cast<cub::detail::scan::policy_selector*>(build.runtime_policy),
+        scan::scan_kernel_source{build},
+        cub::detail::CudaDriverLauncherFactory{cu_device, build.cc});
+      error = static_cast<CUresult>(exec_status);
+    }
```
bernhardmgruber: Suggestion: can we try to unify the two branches by using something like:
```c++
std::conditional_t<std::is_same_v<InitValueT, cub::NullType>, cub::NullType, indirect_arg_t>{init}
```
I think the only difference is that we pass a null init, as `NullType` and otherwise wrapped in an `indirect_arg_t`

cub/benchmarks/bench/scan/policy_selector.h
```diff
@@ -0,0 +1,31 @@
...
+
+#include <cub/device/device_scan.cuh>
+
+// %RANGE% TUNE_ITEMS ipt 7:24:1
+// %RANGE% TUNE_THREADS tpb 128:1024:32
+// %RANGE% TUNE_MAGIC_NS ns 0:2048:4
+// %RANGE% TUNE_DELAY_CONSTRUCTOR_ID dcid 0:7:1
+// %RANGE% TUNE_L2_WRITE_LATENCY_NS l2w 0:1200:5
+// %RANGE% TUNE_TRANSPOSE trp 0:1:1
+// %RANGE% TUNE_LOAD ld 0:1:1
```
bernhardmgruber: Suggestion: Those `%RANGE%` definitions have no effect in my understanding, since `CMakeLists.txt` will only scan the benchmark source files for them. I would remove them here to not confuse anybody.

cub/benchmarks/bench/scan/policy_selector.h
```diff
@@ -0,0 +1,31 @@
+// SPDX-FileCopyrightText: Copyright (c) 2011-2026, NVIDIA CORPORATION. All rights reserved.
+// SPDX-License-Identifier: BSD-3
```
bernhardmgruber: Suggestion: Since this is a net new file with an arguable new content, I would just use the go-to license header for new files:

```suggestion
// SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
```

cub/cub/device/dispatch/tuning/tuning_radix_sort.cuh
```diff
@@ -1778,7 +1601,7 @@ struct policy_selector
     return ::cuda::std::max(value_size, key_size);
   }
 
-  [[nodiscard]] _CCCL_API constexpr auto make_onsweep_small_key_policy(const small_key_tuning_values& tuning) const
+  [[nodiscard]] _CCCL_API constexpr auto make_onesweep_small_key_policy(const small_key_tuning_values& tuning) const
```
bernhardmgruber, line 1705: Thx!

cub/cub/device/dispatch/dispatch_scan.cuh
```diff
@@ -444,4 +473,206 @@ struct DispatchScan
...
+  constexpr int init_kernel_threads = 128;
+  const int init_grid_size          = ::cuda::ceil_div(num_tiles, init_kernel_threads);
+
+#ifdef CUB_DEBUG_LOG
+  _CubLog("Invoking init_kernel<<<%d, %d, 0, %lld>>>()\n", init_grid_size, init_kernel_threads, (long long) stream);
+#endif // CUB_DEBUG_LOG
+
+  // Invoke init_kernel to initialize tile descriptors
+  launcher_factory(init_grid_size, init_kernel_threads, 0, stream, /* use_pdl */ true)
+    .doit(kernel_source.InitKernel(), tile_state, num_tiles);
```
bernhardmgruber: Important: We have been sloppy in the past checking the return code of `doit`. We should do it better here:

```suggestion
  if (const auto error = CubDebug(launcher_factory(init_grid_size, init_kernel_threads, 0, stream, /* use_pdl */ true)
    .doit(kernel_source.InitKernel(), tile_state, num_tiles))) { return error; }
```

cub/cub/device/dispatch/dispatch_scan.cuh
```diff
@@ -444,4 +473,206 @@ struct DispatchScan
...
+            scan_grid_size,
+            active_policy.block_threads,
+            (long long) stream,
+            active_policy.items_per_thread,
+            scan_sm_occupancy);
+#endif // CUB_DEBUG_LOG
+
+    // Invoke scan_kernel
+    launcher_factory(scan_grid_size, active_policy.block_threads, 0, stream, /* use_pdl */ true)
+      .doit(kernel_source.ScanKernel(), d_in, d_out, tile_state, start_tile, scan_op, init_value, num_items);
```
bernhardmgruber: Same here.

cub/cub/device/dispatch/dispatch_scan.cuh
```diff
@@ -444,4 +473,206 @@ struct DispatchScan
...
+  static_assert(::cuda::std::is_unsigned_v<OffsetT> && sizeof(OffsetT) >= 4,
+                "DispatchScan only supports unsigned offset types of at least 4-bytes");
+
+  ::cuda::arch_id arch_id{};
+  if (const auto error = CubDebug(launcher_factory.PtxArchId(arch_id)))
+  {
+    return error;
+  }
+
+  const scan_policy active_policy = policy_selector(arch_id);
```
bernhardmgruber: Important: Let's make it a habit to also log the selected policy:
```c++
#if !_CCCL_COMPILER(NVRTC) && defined(CUB_DEBUG_LOG)
  NV_IF_TARGET(NV_IS_HOST,
               (std::stringstream ss; ss << active_policy;
                _CubLog("Dispatching DeviceScan to arch %d with tuning: %s\n", (int) arch_id, ss.str().c_str());))
```

cub/cub/device/dispatch/tuning/tuning_scan.cuh
```diff
@@ -566,6 +631,418 @@ struct policy_hub
...
+{
+  int input_value_size;
+  int output_value_size;
+  int accum_size;
+  int offset_size;
+  type_t accum_type;
+  op_kind_t operation_t;
+  primitive_accum primitive_accum_t;
+  primitive_op primitive_op_t;
+  bool benchmark_match;
```
bernhardmgruber, line 876: Important remark: The detection whether the current workload, as indicated by the members of `policy_selector`, corresponds to the benchmark we tuned for is not something we want to eventually expose in the user facing API. So we really want to remove `benchmark_match` here before `policy_selector` is publicly exposed. It can stay for now.

cub/cub/device/dispatch/tuning/tuning_scan.cuh
```diff
@@ -566,6 +631,418 @@ struct policy_hub
...
+struct policy_selector
+{
+  int input_value_size;
+  int output_value_size;
+  int accum_size;
+  int offset_size;
+  type_t accum_type;
+  op_kind_t operation_t;
+  primitive_accum primitive_accum_t;
+  primitive_op primitive_op_t;
```
bernhardmgruber: Suggestion: I would remove those two data members and classify the accumulator and operation base don `accum_type` and `operation_t` inside `operator()(::cuda::arch_id arch)`. The operation should be easy. The definition of a primitive accumulator type is a bit annoying because it depends on `cub::Traits`. I don't really know what's best here :S

cub/benchmarks/bench/scan/policy_selector.h
griwes: This came from the comparison with radix sort, where you've also added these to the policy selector header :sweat_smile:

cub/cub/device/dispatch/tuning/tuning_scan.cuh
griwes: We'll need to figure out how to propagate this condition from c.parallel to CUB when we do that. I'll add a TODO here.

bernhardmgruber:
@griwes we just merged #6811, which also touches the scan tunings. This will probably create some more work for this PR. Issue #6821 also tracks making the new scan implementation available to CCCL.C. Do you think you can handle this as well?

bernhardmgruber:
@griwes I pulled out the delay constructor refactoring in #7668 so I can better stack my refactorings on top, in case this PR takes a bit longer (sorry again for the extra work with warpspeed!)

griwes:
Note, the warpspeed integration is still largely untested; I've added an rtxpro6000 test job to c.parallel and that will be the primary test right now. I'll lease a machine with a relevant GPU if that fails, or if there's anything that's clearly wrong to someone's eyes in review.

Edit: also seems I messed up some constexprness :sweat_smile:

cub/cub/device/dispatch/kernels/scan_warpspeed_policy.cuh
griwes, line 1: Note: I need this struct in both kernel_scan_warpspeed.cuh and tuning_scan.cuh, and cross-including them seems... bad. But putting it here is also icky, it ain't a kernel and it'd be the only file with non-conventional name in this folder. I could put it in, like, cub/detail, but that seems very far away from where it is actually used.

Suggestions? /cc @bernhardmgruber

griwes:
Okay, there were two big issues I addressed in the previous commit, I _think_ that what follows should just be problems of the nature of making compilers happy, save maybe for c.parallel (but I'm hoping the launch bounds fix fixes that also). This should be pretty much ready for post-warpspeed review.

griwes:
One last thing seems to be fundamentally wrong, I'm 99% sure that the reason is that I'm classifying types differently between host and device in c.parallel (c.parallel seeing just STORAGE, but the kernel seeing a trivial type and going with the wrong code path compared to how it is invoked). Working on fixing that.

ci/matrix.yaml
```diff
@@ -61,6 +61,8 @@ workflows:
     # c.parallel -- pinned to gcc13 on Linux to match python
     - {jobs: ['test'], project: 'cccl_c_parallel', ctk: '12.X', cxx: ['gcc13', 'msvc'], gpu: ['rtx2080']}
     - {jobs: ['test'], project: 'cccl_c_parallel', ctk: '13.X', cxx: ['gcc13', 'msvc'], gpu: ['rtx2080', 'l4', 'h100']}
+    # RTX PRO 6000 coverage (limited due to small number of runners):
+    - {jobs: ['test'], project: 'cccl_c_parallel', ctk: '13.X', cxx: ['gcc13'], gpu: ['rtxpro6000']}
```
alliepiper, line 65: Add the same coverage to nightly + weekly, I try to keep those a superset of PR builds.

cub/benchmarks/bench/scan/policy_selector.h
bernhardmgruber: Nice :D I'll drop them: #7736

cub/cub/device/dispatch/kernels/kernel_scan.cuh
```diff
@@ -192,23 +195,52 @@ __launch_bounds__(get_device_scan_launch_bounds<ChainedPolicyT, InputIteratorT,
...
+      policy.warpspeed.num_look_ahead_items,
+      policy.warpspeed.num_total_threads,
+      policy.warpspeed.items_per_thread,
+      policy.warpspeed.tile_size>;
+    NV_IF_TARGET(
+      NV_PROVIDES_SM_100, ({
+        auto scan_params = scanKernelParams<it_value_t<InputIteratorT>, it_value_t<OutputIteratorT>, AccumT>{
+          d_in, d_out, tile_state.lookahead, num_items, num_stages};
+        device_scan_lookahead_body<WarpspeedPolicyT, ForceInclusive, RealInitValueT>(scan_params, scan_op, init_value);
+      }));
```
bernhardmgruber: Suggestion: Since we just merged the warpspeed implementation, it's not on any release branch yet. We could entirely remove it from the legacy policy_hub and only support it on the new policy_selector. This may allow you to simplify some code, since `warpspeedKernelPolicy` may be obsolete and you could just pass the `PolicySelector` directly to `device_scan_lookahead_body` and us it there.

cub/benchmarks/bench/scan/policy_selector.h
```diff
@@ -0,0 +1,23 @@
...
+  {
+    return cub::detail::scan::make_mem_scaled_scan_policy(
+      TUNE_THREADS,
+      TUNE_ITEMS,
+      int{sizeof(AccumT)},
+      TUNE_LOAD_ALGORITHM,
+      TUNE_LOAD_MODIFIER,
+      TUNE_STORE_ALGORITHM,
+      cub::BLOCK_SCAN_WARP_SCANS,
+      cub::detail::delay_constructor_policy_from_type<delay_constructor_t>);
```
bernhardmgruber: I added a global constant `delay_constructor_policy` like I did in #7667 to replace this.

cub/cub/device/dispatch/kernels/kernel_scan.cuh
```diff
@@ -106,28 +111,26 @@ DeviceCompactInitKernel(ScanTileStateT tile_state, int num_tiles, NumSelectedIte
     *d_num_selected_out = 0;
   }
 }
-template <typename ChainedPolicyT, typename InputIteratorT, typename OutputIteratorT, typename AccumT>
-[[nodiscard]] _CCCL_DEVICE_API _CCCL_CONSTEVAL int get_device_scan_launch_bounds() noexcept
+template <typename PolicySelector, typename InputIteratorT, typename OutputIteratorT, typename AccumT>
+_CCCL_DEVICE constexpr int get_device_scan_launch_bounds() noexcept
```
bernhardmgruber: Q: Why is the drop of `[[nodiscard]]` and `_CCCL_DEVICE_API` needed? Can we just leave the signature?

cub/cub/device/dispatch/kernels/kernel_scan_warpspeed.cuh
```diff
@@ -804,8 +1022,7 @@ inline constexpr bool scan_use_warpspeed<Policy,
   && ::cuda::std::is_default_constructible_v<it_value_t<OutputIteratorT>>
   // need to fit one stage into 48KiB so binaries stay forward compatible with future GPUs
   && one_stage_fits_48KiB_SMEM<typename Policy::WarpspeedPolicy, InputIteratorT, OutputIteratorT, AccumT>();
```
bernhardmgruber: I think this is unused now. Can we remove it?

cub/cub/device/dispatch/kernels/scan_warpspeed_policy.cuh
bernhardmgruber: Just leave it here. It's fine. We should do a larger reorganization of files with CCCL 4.0 and move a lot more into a detail directory. All the dispatch, kernels and tuning files for example. They should all be clearly internal.

cub/cub/device/dispatch/kernels/scan_warpspeed_policy.cuh
```diff
@@ -0,0 +1,67 @@
...
+
+#include <cub/config.cuh>
+
+#include <cub/detail/warpspeed/squad/squad_desc.cuh>
+
+CUB_NAMESPACE_BEGIN
+
+namespace detail::scan
+{
+struct scan_warpspeed_policy
```
bernhardmgruber, line 18: Important: Please add `operator!=` and `operator<<`.

cub/cub/device/dispatch/kernels/kernel_scan_warpspeed.cuh
```diff
@@ -746,16 +847,135 @@ device_scan_init_lookahead_body(warpspeed::tile_state_t<AccumT>* tile_states, co
...
+  };
+
+  setup_scan_resources(
+    policy,
+    syncHandler,
+    smemAllocator,
+    res.smemInOut,
+    res.smemNextBlockIdx,
+    res.smemSumExclusiveCta,
+    res.smemSumThreadAndWarp);
```
bernhardmgruber, line 899: Remark: This seems like a massive duplication of the logic `allocResources` does. I am extremely worried this will render the codebase brittle and unmaintainable. We should really try to come up with a way to not duplicate so much logic.

griwes:
The setup of the resources is the same from the perspective of the kernel. The only thing that changes there is the ability to use runtime-sized types with the same logic.

If we drop that from this PR, I will need to duplicate a whole bunch of code between the paths for the host and for the device w.r.t. how the resources are set up for the purpose of computing the max number of stages and the dynamic shared memory necessary. Splitting the core logic of allocating the phases is risky, because now updating them would require correctly updating them both in the exact same way, and that's harder to maintain.

The changes you see there are crucial to actually ensure the logic matches.

cub/cub/device/dispatch/kernels/kernel_scan_warpspeed.cuh
griwes: This is, in fact, a **reduction** of the duplication. The only way to not have the parts that are duplicated duplicated is to entirely drop the use of typed resources and use the raw resources (so the path you're highlighting here) as the only code path. I... can do that, but that appears to me to be less desirable than what I have in the PR right now.

griwes: The reason it is like this is that the current code is all written in terms of types and their statically known sizes. We do not have that in c.parallel code paths, we need to use runtime values. So it's either this or all code paths use raw resources exclusively.

cub/cub/device/dispatch/kernels/kernel_scan.cuh
griwes: We can. This went through a few revisions and I lost it in the process, I'll restore it.

griwes:
The bottom line is that everything here is related to the new tuning API. Cutting out the changes around the resource setup and calculations means cutting out c.parallel, and actually getting c.parallel to be able to use the new toys is a good part of the reasoning for these changes in the first place.

griwes:
Last remaining real failure is SASS checks in non-scan c.parallel tests on sm120; I'll pull that out of this PR, together with the enablement of the config in CI, and post it separately.

cub/cub/device/dispatch/tuning/tuning_scan.cuh
```diff
@@ -564,80 +636,534 @@ struct policy_hub
...
+template <typename T>
+concept scan_policy_selector = policy_selector<T, scan_policy>;
+#endif // _CCCL_HAS_CONCEPTS()
+
+constexpr _CCCL_HOST_DEVICE delay_constructor_policy default_delay_constructor_policy(primitive_accum primitive)
+{
+  return primitive == primitive_accum::yes
+         ? delay_constructor_policy{delay_constructor_kind::fixed_delay, 350, 450}
+         : delay_constructor_policy{delay_constructor_kind::no_delay, 0, 450};
+}
```
bernhardmgruber: Important: I added this function in a different PR already. Please use it:
https://github.com/NVIDIA/cccl/blob/bfc82f0c26331e4b3bb1893fafdeb90890e460c7/cub/cub/detail/delay_constructor.cuh#L196-L203

cub/cub/device/dispatch/kernels/kernel_scan_warpspeed.cuh
```diff
@@ -53,76 +61,117 @@ struct scanKernelParams
...
-  // speedup compared to setting it equal to num_stages. Not sure why.
-  int numBlockIdxStages = numStages - 1;
-  // Ensure we have at least 1 stage
-  numBlockIdxStages = numBlockIdxStages < 1 ? 1 : numBlockIdxStages;
-
-  // We do not need too many sumExclusiveCta stages. The lookback warp is the
-  // bottleneck. As soon as it produces a new value, it will be consumed by the
-  // scanStore squad, releasing the stage.
-  int numSumExclusiveCtaStages = 2;
+  const auto counts = make_scan_stage_counts(numStages);
```
bernhardmgruber: Suggestion: could use a structured binding

cub/cub/device/dispatch/kernels/scan_warpspeed_policy.cuh
```diff
@@ -0,0 +1,88 @@
...
+{
+  static constexpr int num_squads = 5;
+
+  bool valid = false;
```
bernhardmgruber, line 22: Remark: we should probably introduce an algorithm enum like in DeviceTransform before all the policies go public. No changes need for now.

cub/cub/device/dispatch/dispatch_scan.cuh
```diff
@@ -406,20 +463,22 @@ struct DispatchScan
     return cudaSuccess;
   }
 
-#if __cccl_ptx_isa >= 860
-  template <typename ActivePolicyT>
-  CUB_RUNTIME_FUNCTION _CCCL_HOST _CCCL_FORCEINLINE cudaError_t __invoke_lookahead_algorithm(ActivePolicyT)
+#if _CCCL_CUDACC_AT_LEAST(12, 8)
```
bernhardmgruber: Question: why is this change needed? We should check for the PTX ISA we require IMO.

cub/cub/device/dispatch/dispatch_scan.cuh
```diff
@@ -445,19 +498,37 @@ struct DispatchScan
...
-    int smem_size    = smem_size_1_stage;
+    int smem_size    = detail::scan::smem_for_stages(
+      warpspeed_policy,
+      num_stages,
+      policy_selector.input_value_size,
+      policy_selector.input_value_alignment,
+      policy_selector.output_value_size,
+      policy_selector.output_value_alignment,
+      policy_selector.accum_size,
+      policy_selector.accum_alignment);
```
bernhardmgruber: Important: Please retain the compile-time check when possible. It helps a lot with development if we can turn on warpspeed unconditionally and just compile to see if we find any test failures etc.

```suggestion
    CUB_DETAIL_CONSTEXPR_ISH int smem_size    = detail::scan::smem_for_stages(
      warpspeed_policy,
      num_stages,
      policy_selector.input_value_size,
      policy_selector.input_value_alignment,
      policy_selector.output_value_size,
      policy_selector.output_value_alignment,
      policy_selector.accum_size,
      policy_selector.accum_alignment);
       CUB_DETAIL_STATIC_ISH_ASSERT(smem_size <= detail::max_smem_per_block); // this is ensured by scan_use_warpspeed
```
`smem_for_stages` is `constexpr`, so I think we just need to pull the `policy_getter` from `__invoke` into `__invoke_lookahead_algorithm` to get a `CUB_DETAIL_CONSTEXPR_ISH auto active_policy = policy_getter();`

cub/cub/device/dispatch/dispatch_scan.cuh
```diff
@@ -647,4 +865,145 @@ struct DispatchScan
...
+      d_out,
+      num_items,
+      scan_op,
+      init_value,
+      stream,
+      -1 /* ptx_version, not used actually */,
+      kernel_source,
+      launcher_factory}
+      .__invoke(policy_getter, policy_selector);
+  });
```
bernhardmgruber, line 1031: Remark: I wonder if it would have been easier to duplicate the logic from `DispatchScan` into the `dispatch` function and strip all warpspeed logic from `DispatchScan`. The warpspeed scan is not on a release branch yet, so it's fine if it's not reachable through `DispatchScan`.

cub/test/catch2_test_device_scan_env.cu
```diff
@@ -141,6 +129,13 @@ struct scan_tuning : cub::detail::scan::tuning<scan_tuning<BlockThreads>>
...
     using MaxPolicy = Policy500;
   };
+
+  template <class InputValueT, class OutputValueT, class AccumT, class OffsetT, class ScanOpT>
+  using selector =
+    cub::detail::scan::policy_selector_from_hub<policy_hub<InputValueT, OutputValueT, AccumT, OffsetT, ScanOpT>,
+                                                InputValueT,
+                                                OutputValueT,
+                                                AccumT>;
 };
```
bernhardmgruber, line 117: Suggestion: we can just rewrite `scan_tuning` to be only a policy selector (only have `operator()`).

cub/cub/device/device_scan.cuh
```diff
@@ -53,7 +53,7 @@ struct tuning
 struct default_tuning : tuning<default_tuning>
 {
   template <typename InputValueT, typename OutputValueT, typename AccumT, typename OffsetT, typename ScanOpT>
-  using fn = policy_hub<InputValueT, OutputValueT, AccumT, OffsetT, ScanOpT>;
+  using selector = policy_selector_from_types<InputValueT, OutputValueT, AccumT, OffsetT, ScanOpT>;
 };
```
bernhardmgruber: and drop this entirely

cub/cub/device/device_scan.cuh
```diff
@@ -137,14 +137,23 @@ struct DeviceScan
...
                                                     typename InitValueT::value_type>>;
 
-    using policy_t = typename scan_tuning_t::
-      template fn<detail::it_value_t<InputIteratorT>, detail::it_value_t<OutputIteratorT>, accum_t, offset_t, ScanOpT>;
+    using policy_selector_t = typename scan_tuning_t::template selector<
+      detail::it_value_t<InputIteratorT>,
+      detail::it_value_t<OutputIteratorT>,
+      accum_t,
+      offset_t,
+      ScanOpT>;
```
bernhardmgruber: and:
```suggestion
using default_policy_selector = policy_selector_from_types<InputValueT, OutputValueT, AccumT, OffsetT, ScanOpT>;
using policy_selector_t = ::cuda::std::execution::
__query_result_or_t<TuningEnvT, detail::scan::get_tuning_query_t, default_policy_selector>;

// Unsigned integer type for global offsets
using offset_t = detail::choose_offset_t<NumItemsT>;
```

bernhardmgruber:
I finished another review and I only have minor comments except for the wish to retain the static assert that one stage fits into SMEM. I am now waiting for confirmation that we don't see SASS changes.

bernhardmgruber:
I have been thinking a bit about how the check whether a single stage fits into 48KiB SMEM, and I wondered whether we actually need this check in CCCL.C. The main purpose of the check is to ensure forward compatibility of compiled binaries. So if you compile for `sm_100` today and run that binary in 10 years on a GPU that really only has 48KiB SMEM, it should still work. We don't need that guarantee for CCCL.C, since we don't keep around binaries.

The second reason we have this check is that a user could provide us with an input type, or an accumulator type (as dictated by the scan operator), that is so huge that we go beyond 48KiB SMEM even with a conservative tuning policy, and we should just fall back to the old scan, because it's not possible to run the warpspeed scan.

Now I wondered, is the set of types that CCCL.C will use open or closed? Because if we know all types that warpspeed scan will be used from CCCL.C, we can just test if it fits into SMEM in a unit test and entirely omit the entire compile time checking for CCCL.C. We would just drop the SMEM check from the  `scan_use_warpspeed` predicate. That would make this PR a lot simpler.

bernhardmgruber:
I just realized we still need the runtime computation to know how much SMEM we must request :S

cub/cub/device/dispatch/dispatch_scan.cuh
griwes: It's noise, backed it out.

griwes:
There *is* SASS changes. Here's a random assortment of kernels compared: https://gist.github.com/griwes/a94e3daf0d2b58faaeebea1932e0c1b0. I believe that there's a whole bunch of codegen artifacts here + some loss/gain of uniform instructions (presumably because the changes made it both easier and harder for the compiler to reason about uniformity...). I have not spotted any significant changes in the hot paths.

There's also two specific cases that seem to now be producing LMEM instructions, though as far as I can tell it's not in the hot loop either: https://gist.github.com/griwes/e0bc6107675b9a55fc3efabdc7244564.

bernhardmgruber:
While investigating the SASS changes, I noticed that some symbols in the CUB benchmarks contained the use of `policy_selector_from_hub`, which we should no longer see (its only use to support users directly accessing the dispatcher). I found out that those came from Thrust, so I ported the Thrust CUB backend to use `cub::detail::scan::dispatch` directly. But that is leading to test failures now in `thrust.test.scan`, which is super odd.

cub/cub/device/dispatch/kernels/kernel_scan.cuh
```diff
@@ -177,7 +180,7 @@ template <typename ChainedPolicyT,
           typename AccumT,
           bool ForceInclusive,
           typename RealInitValueT = typename InitValueT::value_type>
-__launch_bounds__(get_device_scan_launch_bounds<ChainedPolicyT, InputIteratorT, OutputIteratorT, AccumT>(), 1)
+__launch_bounds__(get_device_scan_launch_bounds<PolicySelector, InputIteratorT, OutputIteratorT, AccumT>())
```
bernhardmgruber: Critical: The drop of `, 1` causes many SASS changes. Adding it back makes some Thrust tests pass again and removes all SASS changes in `thrust.test.scan`.

bernhardmgruber: Fixed

bernhardmgruber:
There are now no SASS changes for `cub.bench.scan.exclusive.sum.base` on SM`75;80;86;90;100;120`

bernhardmgruber:
pre-commit.ci autofix

griwes:
pre-commit.ci autofix

PR #7567, opened 2026-02-09
[STF] Add stf_logical_data_with_place

caugonnet:
Make it possible to specify where data is located when creating a logical data with STF C interface

## Description

closes 

## Checklist

### DIFF
--- a/c/experimental/stf/test/test_logical_data_with_place.cu
+++ b/c/experimental/stf/test/test_logical_data_with_place.cu
@@ -21,7 +21,10 @@
 __global__ void scale_inplace(int n, float* data, float factor)
 {
   int i = blockIdx.x * blockDim.x + threadIdx.x;
-  if (i < n) { data[i] *= factor; }
+  if (i < n)
+  {
+    data[i] *= factor;
+  }
 }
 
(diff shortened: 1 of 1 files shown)


c/experimental/stf/src/stf.cu
```diff
@@ -44,12 +44,44 @@ cudaStream_t stf_fence(stf_ctx_handle ctx)
...
+      break;
+    case STF_DATA_PLACE_MANAGED:
+      cpp_dplace = cuda::experimental::stf::data_place::managed();
+      break;
+    case STF_DATA_PLACE_AFFINE:
+      cpp_dplace = cuda::experimental::stf::data_place::affine();
+      break;
+    default:
+      // Invalid data place - this should not happen with valid input
+      assert(false && "Invalid data_place kind");
```
bernhardmgruber, line 78: Suggestion: if `stf_logical_data_with_place` is a public API, or receives any user data through some public API that is potentially unsanitized, then I would use a stronger input validation here. Since this we seem to be in a C API and there is no error code to be returned, I think I would rather abort/exit/terminate the program here.

PR #7568, opened 2026-02-09
Fix operator identity and absorbing element for `char`

bernhardmgruber:
This PR adds the necessary checks to also provide operator properties if the element type is `char`. This fixes #6811 , which relied on `cub::detail::identity_v`.

### DIFF
--- a/libcudacxx/include/cuda/__functional/operator_properties.h
+++ b/libcudacxx/include/cuda/__functional/operator_properties.h
@@ -411,7 +411,7 @@ template <class _Op, class _Tp>
   using _Up = ::cuda::std::remove_cv_t<_Tp>;
   if constexpr (__is_cuda_std_plus_v<_Op>)
   {
-    if constexpr (::cuda::std::__cccl_is_integer_v<_Up>)
+    if constexpr (::cuda::std::__cccl_is_integer_v<_Up> || ::cuda::std::is_same_v<_Up, char>)
     {
       return _Up{};
     }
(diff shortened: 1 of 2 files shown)


libcudacxx/include/cuda/__functional/operator_properties.h
```diff
@@ -411,7 +411,7 @@ template <class _Op, class _Tp>
   using _Up = ::cuda::std::remove_cv_t<_Tp>;
   if constexpr (__is_cuda_std_plus_v<_Op>)
   {
-    if constexpr (::cuda::std::__cccl_is_integer_v<_Up>)
+    if constexpr (::cuda::std::__cccl_is_integer_v<_Up> || ::cuda::std::is_same_v<_Up, char>)
```
davebayer, line 414: Q: Why don't we handle all other character types then, too?

bernhardmgruber: `__cccl_is_integer_v` already handles `signed char` and `unsigned char`.

davebayer: I mean `char8_t`, `char16_t`, `char32_t` and `wchar_t`

bernhardmgruber: This PR is fine without them I think. If you want to add support for those, please go ahead. I wonder though whether anything in CCCL or CUB would lead to a fast kernel using one of those type :) I think we lack traits in all kinds of places.

davebayer: Personally, I don't need it for anything, but I don't like partial solutions. What makes `char` different from `char16_t` for example?

miscco: Question, why isnt this just `::cuda::std::is_integral_v`?

bernhardmgruber: Does covering `bool` make sense? if yes, @davebayer can you file a follow-up PR to make it `is_integral_v` please? Thx!

fbusato: brief note. I used `__cccl_is_integer_v` because identity and absorbing elements are mathematical concepts and apply to “arithmetic/compute” values, not chars. Anyway, the extension is fine.

PR #7571, opened 2026-02-09
Vectorize reduction also for trivially relocatable types

bernhardmgruber:
I want to reduce our use of `cub::is_primitive`, but switching to `cuda::std::trivially_copyable` is a breaking change for any user that specializes `cub::Traits` for any non-trivially copyable type. So let's use both and drop `cub::is_primitive` in CCCL 4.0.

### DIFF
--- a/cub/cub/agent/agent_reduce.cuh
+++ b/cub/cub/agent/agent_reduce.cuh
@@ -215,7 +215,7 @@ struct AgentReduceImpl
     (VECTOR_LOAD_LENGTH > 1) && (ITEMS_PER_THREAD % VECTOR_LOAD_LENGTH == 0)
     && (::cuda::std::is_pointer_v<InputIteratorT>)
     // TODO(bgruber): remove the check for is_primitive<ValueT> in CCCL 4.0
-    &&(is_primitive<InputT>::value || thrust::is_trivially_relocatable_v<InputT>);
+    &&(is_primitive<InputT>::value || THRUST_NS_QUALIFIER::is_trivially_relocatable_v<InputT>);
 
   static constexpr CacheLoadModifier LOAD_MODIFIER = AgentReducePolicy::LOAD_MODIFIER;
 


fbusato:
we still have the problem with `__half` and `__nv_bfloat16` + their compound types 😢

bernhardmgruber:
> we still have the problem with `__half` and `__nv_bfloat16` + their compound types 😢

Yeah, I still don't have a good solution for those except for removing `cub::Traits` and giving users a new opt-in. Or actually checking whether their half type is already trivially copyable.

cub/cub/agent/agent_reduce.cuh
```diff
@@ -208,9 +210,12 @@ struct AgentReduceImpl
 
   // Can vectorize according to the policy if the input iterator is a native
   // pointer to a primitive type
+  // TODO(bgruber): we should not check for `is_pointer_v` but `contiguous_iterator` and unwrap it
   static constexpr bool ATTEMPT_VECTORIZATION =
     (VECTOR_LOAD_LENGTH > 1) && (ITEMS_PER_THREAD % VECTOR_LOAD_LENGTH == 0)
-    && (::cuda::std::is_pointer_v<InputIteratorT>) && is_primitive<InputT>::value;
+    && (::cuda::std::is_pointer_v<InputIteratorT>)
+    // TODO(bgruber): remove the check for is_primitive<ValueT> in CCCL 4.0
+    &&(is_primitive<InputT>::value || THRUST_NS_QUALIFIER::is_trivially_relocatable_v<InputT>);
```
miscco, line 218: The title talks about `trivially copyable` but this references `trivially relocatable` should this accept both?

bernhardmgruber: I updated the PR title. Trivially relocatable is fine I think.

cub/cub/agent/agent_reduce.cuh
```diff
@@ -208,9 +210,12 @@ struct AgentReduceImpl
 
   // Can vectorize according to the policy if the input iterator is a native
   // pointer to a primitive type
+  // TODO(bgruber): we should not check for `is_pointer_v` but `contiguous_iterator` and unwrap it
   static constexpr bool ATTEMPT_VECTORIZATION =
     (VECTOR_LOAD_LENGTH > 1) && (ITEMS_PER_THREAD % VECTOR_LOAD_LENGTH == 0)
-    && (::cuda::std::is_pointer_v<InputIteratorT>) && is_primitive<InputT>::value;
+    && (::cuda::std::is_pointer_v<InputIteratorT>)
```
davebayer, line 216: Q: Shouldn't any _ContiguousRange_ be supported?

davebayer (approved):

cub/cub/agent/agent_reduce.cuh
bernhardmgruber: Yes, which is why I added a comment three lines above :)

PR #7572, opened 2026-02-09
Use operator properties for `reduce_by_key` identity

bernhardmgruber:
And refactor a little

### DIFF
--- a/thrust/thrust/system/cuda/detail/reduce_by_key.h
+++ b/thrust/thrust/system/cuda/detail/reduce_by_key.h
@@ -182,16 +182,7 @@ struct ReduceByKeyAgent
 
   // Whether or not the scan operation has a zero-valued identity value (true
   // if we're performing addition on a primitive type)
-  static constexpr int HAS_IDENTITY_ZERO = [] {
-    if constexpr (::cuda::has_identity_element_v<ReductionOp, value_type>)
-    {
-      return ::cuda::identity_element<ReductionOp, value_type>() == value_type{};
-    }
-    else
-    {
-      return false;
-    }
-  }();
+  static constexpr int HAS_IDENTITY_ZERO = ::cuda::identity_element<ReductionOp, value_type>() == 0;
 
   struct impl
   {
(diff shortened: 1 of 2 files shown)


thrust/thrust/system/cuda/detail/reduce_by_key.h
```diff
@@ -180,10 +180,18 @@ struct ReduceByKeyAgent
...
+  static constexpr int HAS_IDENTITY_ZERO = [] {
+    if constexpr (::cuda::has_identity_element_v<ReductionOp, value_type>)
+    {
+      return ::cuda::identity_element<ReductionOp, value_type>() == value_type{};
+    }
+    else
+    {
+      return false;
+    }
+  }();
```
bernhardmgruber: @fbusato I wonder whether we should make `__no_identity_element` just comparable to anything and return false. Then we could just write:

```suggestion
  static constexpr int HAS_IDENTITY_ZERO = ::cuda::identity_element<ReductionOp, value_type>() == value_type{};
```
Since `__no_identity_element{} != value_type{}`.

thrust/thrust/system/cuda/detail/reduce_by_key.h
bernhardmgruber, line 253: I was a bit surprised that both functions here have the same implementation. Could there be an error somewhere?

thrust/thrust/system/cuda/detail/reduce_by_key.h
fbusato: The idea is interesting, but I'm a bit puzzled about how `__no_identity_element` can be used. Are there situations where a user can make a wrong assumption about returning false instead of a compile error?

bernhardmgruber: I don't fully understand what you are getting at. In generic code for any `Op` an `T`, how would you test if the identity element is `T{}`?

fbusato: I'm not referring to testing the identity, but to the situation where the user needs SFINAE. On the other hand,  there is `has_identity_element_v`, so it should be fine.

fbusato (approved):

thrust/thrust/system/cuda/detail/reduce_by_key.h
bernhardmgruber: I implemented the operators now.

elstehle (approved):

PR #7575, opened 2026-02-09
Support more types in decoupled lookback fastpath

bernhardmgruber:
The goal is to reduce the use of `is_primitive<T>` in the long term.

### DIFF
--- a/cub/cub/util_arch.cuh
+++ b/cub/cub/util_arch.cuh
@@ -117,7 +117,7 @@ static constexpr ::cuda::std::size_t max_smem_per_block = 48 * 1024;
 
 // The size in bytes of the largest machine word that can be atomically read/written in a single instruction, so we can
 // use it to pass messages from one thread to another using strong loads (acquire) and stores (release).
-inline static constexpr int largest_atomic_message_size = 16;
+inline constexpr int largest_atomic_message_size = 16;
 
 struct scaling_result
 {


cub/cub/agent/single_pass_scan_operators.cuh
```diff
@@ -646,7 +653,14 @@ _CCCL_HOST_DEVICE _CCCL_FORCEINLINE cudaError_t tile_state_init(
 /**
  * Tile status interface.
  */
-template <typename T, bool SINGLE_WORD = detail::is_primitive<T>::value>
+template <typename T,
+          // TODO(bgruber): remove the check for is_primitive<T> in CCCL 4.0
+          bool SINGLE_WORD = detail::is_primitive<T>::value
```
fbusato: I would avoid macro style here

cub/cub/util_arch.cuh
```diff
@@ -115,6 +115,9 @@ inline constexpr bool prefer_conflict_over_padding = CUB_PTX_PREFER_CONFLICT_OVE
 // shared memory in a kernel. Note that dynamic shared memory may be larger than this amount.
 static constexpr ::cuda::std::size_t max_smem_per_block = 48 * 1024;
 
+// The size in bytes of the largest machine word that can be atomically read/written in a single instruction.
+inline static constexpr int largest_atomic_word_size = 16;
```
fbusato: I don't think this is true for all architectures

cub/cub/agent/single_pass_scan_operators.cuh
```diff
@@ -544,14 +546,19 @@ using default_no_delay_t             = default_no_delay_constructor_t::delay_t;
...
 
 template <class T>
 using default_delay_t = typename default_delay_constructor_t<T>::delay_t;
 
 template <class KeyT, class ValueT>
 using default_reduce_by_key_delay_constructor_t =
-  ::cuda::std::_If<is_primitive<ValueT>::value && (sizeof(ValueT) + sizeof(KeyT) < 16),
+  // TODO(bgruber): remove the check for is_primitive<ValueT> in CCCL 4.0
+  ::cuda::std::_If<(is_primitive<ValueT>::value || ::cuda::std::is_trivially_copyable_v<ValueT>)
+                     && (sizeof(ValueT) + sizeof(KeyT) < largest_atomic_word_size),
```
fbusato: `(sizeof(ValueT) + sizeof(KeyT) < largest_atomic_word_size)` or
`(sizeof(ValueT) + sizeof(KeyT) <= largest_atomic_word_size)`?

cub/cub/agent/single_pass_scan_operators.cuh
```diff
@@ -544,14 +546,19 @@ using default_no_delay_t             = default_no_delay_constructor_t::delay_t;
...
+                   default_no_delay_constructor_t>;
 
 template <class T>
 using default_delay_t = typename default_delay_constructor_t<T>::delay_t;
 
 template <class KeyT, class ValueT>
 using default_reduce_by_key_delay_constructor_t =
-  ::cuda::std::_If<is_primitive<ValueT>::value && (sizeof(ValueT) + sizeof(KeyT) < 16),
+  // TODO(bgruber): remove the check for is_primitive<ValueT> in CCCL 4.0
+  ::cuda::std::_If<(is_primitive<ValueT>::value || ::cuda::std::is_trivially_copyable_v<ValueT>)
```
fbusato: `conditional_t`?

cub/cub/util_arch.cuh
bernhardmgruber: I don't know TBH. We hardcoded 16 in the past and I know the build-in atomics like `__nv_atomic_load` support up to 16 bytes. Also, `cub::detail::store_release` etc. are also only implemented for up to 16 bytes. I think we are fine wrt. this PR, but we may revisit this in the future.

cub/cub/agent/single_pass_scan_operators.cuh
bernhardmgruber: I think it must be less than, since we also need to fit the status word into the 16 bytes.

cub/test/catch2_test_device_scan_invalid.cu
```diff
@@ -57,6 +56,9 @@ struct segment
   }
 };
 
+static_assert(!cub::detail::is_primitive_v<segment>);
+// static_assert(!cuda::std::is_trivially_copyable_v<segment>); // TODO(bgruber): why is this important?
```
bernhardmgruber, line 60: @pauleonix I need your input here please. Why does `segment` need to be a non-primitive type and what should this mean here? Because it is trivially copyable.

pauleonix: We wanted to explicitly test the non-primitive path.

bernhardmgruber: The changes in this PR make `segment` take the primitive path now, because it's trivially copyable. So I guess `segment` must be made non-trivially copyable by e.g. adding a non-trivial destructor or a copy ctor? Does that make sense?

bernhardmgruber: Confirmed offline.

cub/cub/util_arch.cuh
miscco: I believe this requires SM90 otherwise its only 8

bernhardmgruber: I think the name of the variable may not be accurate then. We have `ld.acquire.gpu.v2.u64` on SM70, and `ld.cg.v2.u64`+`__threadfence()` before that. (I am looking at the implementation of `cub::detail::load_acquire`). What this variable means is what is the largest message size in bytes that we can pass from one thread to another using a store release and a load acquire.

cub/cub/agent/single_pass_scan_operators.cuh
```diff
@@ -646,7 +653,14 @@ _CCCL_HOST_DEVICE _CCCL_FORCEINLINE cudaError_t tile_state_init(
 /**
  * Tile status interface.
  */
-template <typename T, bool SINGLE_WORD = detail::is_primitive<T>::value>
+template <typename T,
+          // TODO(bgruber): remove the check for is_primitive<T> in CCCL 4.0
+          bool single_word = detail::is_primitive<T>::value
```
pauleonix:
```suggestion
          bool SingleWord = detail::is_primitive<T>::value
```

cub/cub/agent/single_pass_scan_operators.cuh
```diff
@@ -1017,7 +1033,9 @@ struct ScanTileState<T, false>
  */
 template <typename ValueT,
           typename KeyT,
-          bool SINGLE_WORD = detail::is_primitive<ValueT>::value && (sizeof(ValueT) + sizeof(KeyT) < 16)>
+          // TODO(bgruber): remove the check for is_primitive<ValueT> in CCCL 4.0
+          bool single_word = (detail::is_primitive<ValueT>::value || ::cuda::std::is_trivially_copyable_v<ValueT>)
```
pauleonix:
```suggestion
          bool SingleWord = (detail::is_primitive<ValueT>::value || ::cuda::std::is_trivially_copyable_v<ValueT>)
```

pauleonix (approved):

cub/cub/util_arch.cuh
bernhardmgruber: Renamed to `largest_atomic_message_size`

cub/cub/util_arch.cuh
```diff
@@ -115,6 +115,10 @@ inline constexpr bool prefer_conflict_over_padding = CUB_PTX_PREFER_CONFLICT_OVE
 // shared memory in a kernel. Note that dynamic shared memory may be larger than this amount.
 static constexpr ::cuda::std::size_t max_smem_per_block = 48 * 1024;
 
+// The size in bytes of the largest machine word that can be atomically read/written in a single instruction, so we can
+// use it to pass messages from one thread to another using strong loads (acquire) and stores (release).
+inline static constexpr int largest_atomic_message_size = 16;
```
miscco: This will break NVHPC, we need the same workaround as for the warpspeed scan

miscco (changes requested):

cub/cub/util_arch.cuh
pauleonix: I think the `inline` here does nothing (`static` wins and makes more sense?). Looking at other constants in this header we do not seem to be very consistent about `inline` vs `static` constants?

bernhardmgruber: > we need the same workaround as for the warpspeed scan

No. warpspeed scan uses the compiler built-in atomics, whereas the classical decoupled lookback relies on inline PTX, which works fine on NVHPC.

bernhardmgruber: @pauleonix the `static` seems unnecessary here and should probably be removed.

bernhardmgruber:
```suggestion
inline constexpr int largest_atomic_message_size = 16;
```

PR #7603, opened 2026-02-10
[cudax] Implement _this_ hierarchy groups

### LINKED ISSUE #7618: [FEA]: Initial hierarchy groups implementation
Create an initial hierarchy groups implementation. We are interested in having:
- `thread_group`, a group that contains 1 thread,
- `warp_group`, a group that contains 1 warp,
- `block_group`, a group that contains 1 block,
- `cluster_group`, a group that contains 1 cluster and
- `grid_group`, a group that is an abstraction for the whole grid.

Each group should be specific to a given hierarchy.

davebayer:
Closes  #7618.

### DIFF
--- a/cudax/include/cuda/experimental/__hierarchy/grid_sync.cuh
+++ b/cudax/include/cuda/experimental/__hierarchy/grid_sync.cuh
@@ -79,15 +79,6 @@ _CCCL_DEVICE_API inline void __grid_sync()
                      : "=r"(__old_barrier_value)
                      : "l"(__bar_ptr), "r"(__nblocks)
                      : "memory");
-      }),
-      ({
-        ::__threadfence();
-        __old_barrier_value = ::atomicAdd(__bar_ptr, __nblocks);
-      }))
-
-    NV_IF_ELSE_TARGET(
-      NV_PROVIDES_SM_70,
-      ({
         unsigned __curr_barrier_value;
         do
         {
@@ -96,8 +87,11 @@ _CCCL_DEVICE_API inline void __grid_sync()
           !::cuda::experimental::__cg_imported::__grid_barrier_has_flipped(__old_barrier_value, __curr_barrier_value));
       }),
       ({
+        ::__threadfence();
+        __old_barrier_value = ::atomicAdd(__bar_ptr, __nblocks);
         while (!::cuda::experimental::__cg_imported::__grid_barrier_has_flipped(__old_barrier_value, *__bar_ptr))
-          ;
+        {
+        }
         ::__threadfence();
       }))
   }
--- a/cudax/test/hierarchy/group.cu
+++ b/cudax/test/hierarchy/group.cu
@@ -61,7 +61,8 @@ struct TestKernel
       CUDAX_REQUIRE(this_thread.rank(cuda::grid) == cuda::gpu_thread.rank(cuda::grid));
       this_thread.sync();
 
-      CUDAX_REQUIRE(sum(this_thread, array) == 6);
+      const auto result = sum(this_thread, array);
+      CUDAX_REQUIRE(result == 6);
     }
     {
       unsigned array[]{1, 2, 3};
@@ -71,7 +72,11 @@ struct TestKernel
       CUDAX_REQUIRE(this_warp.rank(cuda::grid) == cuda::warp.rank(cuda::grid));
       this_warp.sync();
 
-      CUDAX_REQUIRE(sum(this_warp, array) == 6 * cuda::gpu_thread.count(cuda::warp));
+      const auto result = sum(this_warp, array);
+      if (cuda::gpu_thread.rank(cuda::warp) == 0)
+      {
+        CUDAX_REQUIRE(result == 6 * cuda::gpu_thread.count(cuda::warp));
+      }
     }
     {
       unsigned array[]{1, 2, 3};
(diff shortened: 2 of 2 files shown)


cudax/include/cuda/experimental/__hierarchy/grid_sync.cuh
```diff
@@ -0,0 +1,113 @@
...
+      }),
+      ({
+        while (!::cuda::experimental::__cg_imported::__grid_barrier_has_flipped(__old_barrier_value, *__bar_ptr))
+          ;
+        ::__threadfence();
+      }))
+  }
+
+  // Wait for the thread 0 to finish the inter block synchronization.
+  ::__syncthreads();
```
pciolkosz: This needs to be `__barrier_sync(0)`, otherwise it won't work if `sync()` is called from a warp that was branched (obviously while all the threads correctly call it). It will generate outlined function that will textually align threads, but `__syncthreads()` will not and just hang.

We want to use the guarantees design later on to allow users to pass a guarantee that they won't branch warps calling this API, in which case we will use the more efficient `__syncthreads()`

cudax/include/cuda/experimental/__hierarchy/group.cuh
```diff
@@ -0,0 +1,264 @@
...
+};
+
+_CCCL_TEMPLATE(class _Hierarchy)
+_CCCL_REQUIRES(__is_or_has_hierarchy_member_v<_Hierarchy>)
+_CCCL_HOST_DEVICE grid_group(const _Hierarchy&)
+  -> grid_group<__hierarchy_type_of<_Hierarchy>, __this_hierarchy_group_kind>;
+
+_CCCL_TEMPLATE(class _HierarchyLike)
+_CCCL_REQUIRES(__is_or_has_hierarchy_member_v<_HierarchyLike>)
+[[nodiscard]] _CCCL_DEVICE_API auto this_thread(const _HierarchyLike& __hier_like) noexcept
```
pciolkosz, line 228: Stream of thoughts: Technically we don't have to carry over the choice to use `this_X` functions.

Options include:
- this_group(level), level.group(), some other variations of similar things
- Just constructors

But there is also no issue with it staying as is

cudax/include/cuda/experimental/__hierarchy/group.cuh
```diff
@@ -0,0 +1,264 @@
...
+  using __base_type::count_as;
+  using __base_type::hierarchy;
+
+#if _CCCL_CUDA_COMPILATION()
+  using __base_type::rank;
+  using __base_type::rank_as;
+
+  _CCCL_DEVICE_API void sync() noexcept
+  {
+    ::__syncthreads();
```
pciolkosz, line 162: Same here it should be `__barrier_sync(0)`

cudax/include/cuda/experimental/__hierarchy/group.cuh
```diff
@@ -0,0 +1,264 @@
...
+  _CCCL_REQUIRES(__is_hierarchy_level_v<_InLevel> _CCCL_AND(!::cuda::std::is_same_v<_Level2, grid_level>))
+  [[nodiscard]] _CCCL_DEVICE_API auto rank(const _InLevel& __in_level) const noexcept
+  {
+    return _Level{}.rank(__in_level);
+  }
+#endif // _CCCL_CUDA_COMPILATION()
+};
+
+template <class _Hierarchy>
+class thread_group<_Hierarchy, __this_hierarchy_group_kind> : __this_hierarchy_group_base<thread_level, _Hierarchy>
```
pciolkosz, line 97: Technically `thread_group` in CG was a general polymorphic type for all types of groups. It fits better with the naming scheme for it to be a single thread, but it will be confusing. We don't need to resolve it right now, but I wanted to note that down.

In general I don't know if we need that polymorphic type anymore or if we need a single thread group

davebayer: That's a good point. But we will need the `thread_group` for `cub::ThreadMeow` algorithms in the future.

As you say, I would deal with this later :)

cudax/include/cuda/experimental/__hierarchy/group.cuh
davebayer: Even though I'd love to use something more generic (as the options you listed), I find
```cpp
auto warp = cuda::this_warp(hierarchy);
```
more readable than any of:
```cpp
// sounds like we are grouping all warps in the hierarchy
auto warp = cuda::warp.group(hierarchy);

// I don't know what's going on at first glance 
auto warp = cuda::this_group(cuda::warp, hierarchy);

// it's not clear that it's the group of the current warp
cuda::warp_group warp{hierarchy};
```
I would love to make the `level.group(h)` work, but I don't find it easy to read and understand. Even `level.this_group(h)` doesn't seem better to me.

I would keep it as is for now :)

cudax/include/cuda/experimental/__hierarchy/grid_sync.cuh
```diff
@@ -0,0 +1,107 @@
...
+}
+
+[[nodiscard]] _CCCL_DEVICE_API inline bool __grid_barrier_has_flipped(unsigned __old, unsigned __curr) noexcept
+{
+  return (((__old ^ __curr) & 0x80000000) != 0);
+}
+
+_CCCL_DEVICE_API inline void __grid_sync()
+{
+  const auto __bar_ptr = &::cuda::experimental::__cg_imported::__grid_workspace_ptr()->__barrier_;
```
pciolkosz, line 57: Should we store this in the group?

CG was also validating it in case someone forgot to use the cooperative launch, so it wouldn't just fault on the device. We could also do validation, but maybe skip it if we construct the group from a config object with the cooperative launch option?

davebayer: I don't see the reason, why we should store the pointer when it's available through the envregs.

Regarding the validation, I moved it to the place where we obtain the pointer from the envregs :)

pciolkosz: We would need to check the generated code, but I believe the compiler can't store the result in registers and will need to generate a full constant bank read every time you call a sync, also do the validity check every time. I am pretty sure storing it in a register will end up being faster at the cost of extra registers

PR #7604, opened 2026-02-10
Incorrect scoped_range_in placement

seanbaxter:
The `optional<scoped_range_in<NVTXCCCLDomain>>` has to be inside the `NV_IS_HOST` macro. `~scoped_range_in` calls `domain::get<>`, which is defined like this:

```cpp
  template <typename D = global,
    typename std::enable_if<
      detail::is_c_string<decltype(D::name)>::value
    , int>::type = 0>
  NVTX3_NO_DISCARD static domain const& get() noexcept
  {
    static domain const d(D::name);
    return d;
  }
```

It initializes an object `d` with static storage duration. The problem is that the initializer is a call to a non-constexpr constructor:

```cpp
explicit domain(char const* name) noexcept : _domain{nvtxDomainCreateA(name)} {}
```

What we have is a dynamic initializer in a device module. That's not currently possible, so the build breaks. Moving the scoped_range_in object inside `NV_IS_HOST` stops this device module contamination.

### DIFF
--- a/libcudacxx/include/cuda/__nvtx/nvtx.h
+++ b/libcudacxx/include/cuda/__nvtx/nvtx.h
@@ -101,9 +101,9 @@ _CCCL_END_NAMESPACE_CUDA
 // into a dispatch region running only on the host, while preserving the semantic scope where the range is declared.
 #    define _CCCL_NVTX_RANGE_SCOPE_IF(condition, name)                                                                 \
       _CCCL_BEFORE_NVTX_RANGE_SCOPE(name)                                                                              \
-      ::cuda::std::optional<::nvtx3::v1::scoped_range_in<::cuda::detail::NVTXCCCLDomain>> __cuda_nvtx3_range;          \
       NV_IF_TARGET(                                                                                                    \
         NV_IS_HOST, ({                                                                                                 \
+          ::cuda::std::optional<::nvtx3::v1::scoped_range_in<::cuda::detail::NVTXCCCLDomain>> __cuda_nvtx3_range;      \
           static const ::nvtx3::v1::registered_string_in<::cuda::detail::NVTXCCCLDomain> __cuda_nvtx3_func_name{name}; \
           static const ::nvtx3::v1::event_attributes __cuda_nvtx3_func_attr{__cuda_nvtx3_func_name};                   \
           if (condition)                                                                                               \


libcudacxx/include/cuda/__nvtx/nvtx.h
```diff
@@ -101,9 +101,9 @@ _CCCL_END_NAMESPACE_CUDA
 // into a dispatch region running only on the host, while preserving the semantic scope where the range is declared.
 #    define _CCCL_NVTX_RANGE_SCOPE_IF(condition, name)                                                                 \
       _CCCL_BEFORE_NVTX_RANGE_SCOPE(name)                                                                              \
-      ::cuda::std::optional<::nvtx3::v1::scoped_range_in<::cuda::detail::NVTXCCCLDomain>> __cuda_nvtx3_range;          \
       NV_IF_TARGET(                                                                                                    \
         NV_IS_HOST, ({                                                                                                 \
+          ::cuda::std::optional<::nvtx3::v1::scoped_range_in<::cuda::detail::NVTXCCCLDomain>> __cuda_nvtx3_range;      \
           static const ::nvtx3::v1::registered_string_in<::cuda::detail::NVTXCCCLDomain> __cuda_nvtx3_func_name{name}; \
```
bernhardmgruber, line 107: Critical: Unfortunately, this will not work, since `__cuda_nvtx3_range` is a scope guard and it has to outlive the code generated by the `_CCCL_NVTX_RANGE_SCOPE_IF` macro. We cannot put it into any scope. `NV_IF_TARGET` does create a scope on NVHPC, since it compiles to `if target ... { ... /* range would be here */ }`

bernhardmgruber (changes requested):

libcudacxx/include/cuda/__nvtx/nvtx.h
seanbaxter: I don't think there is a problem. It expands to this:

```cpp
template <class RandomAccessIteratorT, class NumItemsT, class OpT>
  __attribute__((host)) __attribute__((device)) static cudaError_t
  ForEachCopyN(RandomAccessIteratorT first, NumItemsT num_items, OpT op, cudaStream_t stream = {})
  {
    if target ( nv::target::is_host) { { 
      ::cuda::std::optional<::nvtx3::v1::scoped_range_in<::cuda::detail::NVTXCCCLDomain>> __cuda_nvtx3_range; 
      static const ::nvtx3::v1::registered_string_in<::cuda::detail::NVTXCCCLDomain> __cuda_nvtx3_func_name{"cub::DeviceFor::ForEachCopyN"}; 
      static const ::nvtx3::v1::event_attributes __cuda_nvtx3_func_attr{__cuda_nvtx3_func_name}; 
      if (true) { 
        __cuda_nvtx3_range.emplace(__cuda_nvtx3_func_attr); 
      } 
    } } else {
    };
    return ForEachCopyNNoNVTX(first, num_items, op, stream);
  }
```

The `__cuda_nvtx3_range` outlives its uses.

bernhardmgruber: It needs to live until:
```c++
template <class RandomAccessIteratorT, class NumItemsT, class OpT>
  __attribute__((host)) __attribute__((device)) static cudaError_t
  ForEachCopyN(RandomAccessIteratorT first, NumItemsT num_items, OpT op, cudaStream_t stream = {})
  {
    if target ( nv::target::is_host) { { 
      ::cuda::std::optional<::nvtx3::v1::scoped_range_in<::cuda::detail::NVTXCCCLDomain>> __cuda_nvtx3_range; 
      static const ::nvtx3::v1::registered_string_in<::cuda::detail::NVTXCCCLDomain> __cuda_nvtx3_func_name{"cub::DeviceFor::ForEachCopyN"}; 
      static const ::nvtx3::v1::event_attributes __cuda_nvtx3_func_attr{__cuda_nvtx3_func_name}; 
      if (true) { 
        __cuda_nvtx3_range.emplace(__cuda_nvtx3_func_attr); 
      } 
    } } else {
    };
    return ForEachCopyNNoNVTX(first, num_items, op, stream);
    // HERE
  }
```
In order to correctly generate an NVTX range for the full function scope. Or am I missing something?

seanbaxter: It's ill-formed as it is now, so something has to change.

bernhardmgruber: Does #7649 fix your issue?

PR #7605, opened 2026-02-10
Don't use `inline` with glibcxx namespaces

davebayer:
The glibcxx namespace macros can expand to nothing, so it leaves plain `inline` somewhere in our code. We can remove it, because glibcxx defines the namespace as `inline` if needed inside the `c++config`


davebayer:
This doesn't make any sense

PR #7620, opened 2026-02-10
cuda.compute: Fix description of argument types in Sphinx documentation

shwina:
## Description

Currently, the rendered Sphinx documentation is inaccurate (and ugly) when it comes to argument types. For example, consider the documentation for [`reduce_into`](https://nvidia.github.io/cccl/python/compute_api.html#cuda.compute.algorithms.reduce_into):

Screenshot 2026-02-10 at 4 22 48 PM

* It's wrong: the type of `d_in` for example is not `DeviceArray | Any`.
* It's ugly: the type of the `op` argument is basically unreadable (how is the user supposed to known what `cuda.compute._bindings.OpKind` is?)

This PR introduces type aliases for operators and iterators that are documented, and makes it so that clicking on the argument type takes you to a description for that type:

Screenshot 2026-02-10 at 4 27 19 PM

## Checklist

### DIFF
--- a/python/cuda_cccl/cuda/compute/algorithms/_sort/_merge_sort.py
+++ b/python/cuda_cccl/cuda/compute/algorithms/_sort/_merge_sort.py
@@ -135,7 +135,7 @@ def make_merge_sort(
         d_in_items: Optional device array or iterator that contains each key's corresponding item
         d_out_keys: Device array to store the sorted keys
         d_out_items: Device array to store the sorted items
-        op: Callable or OpKind representing the comparison operator
+        op: The comparison operator for sorting. The signature is  ``(T, T) -> int8``, where ``T`` is the input data type.
 
     Returns:
         A callable object that can be used to perform the merge sort
@@ -171,7 +171,7 @@ def merge_sort(
         d_in_items: Device array or iterator containing the input sequence of items (optional)
         d_out_keys: Device array to store the sorted keys
         d_out_items: Device array to store the sorted items (optional)
-        op: Comparison operator for sorting
+        op: The comparison operator for sorting. The signature is  ``(T, T) -> int8``, where ``T`` is the input data type.
         num_items: Number of items to sort
         stream: CUDA stream for the operation (optional)
     """


python/cuda_cccl/cuda/compute/iterators/_zip.py
```diff
@@ -19,7 +19,18 @@ class ZipIterator(IteratorBase):
     """
     Iterator that zips multiple iterators together.
 
-    At each position, yields a tuple/struct of values from all underlying iterators.
+    At each position, yields a tuple of values from all underlying iterators.
```
shwina, line 22: @NaderAlAwar I promised to tack on this change (and in `_permutation.py`) in an older PR.

python/cuda_cccl/cuda/compute/_caching.py
```diff
@@ -48,13 +55,11 @@ def _key_for(value: Any) -> Hashable:
     # DeviceArrayLike is not a runtime-checkable protocol, so
     # we cannot isinstance() with it.
     if is_device_array(value):
-        return _KEY_FUNCTIONS[DeviceArrayLike](value)
+        return (_type_fqn(value), get_dtype(value))
```
shwina, line 58: @NaderAlAwar there's also this change we decided we should do (`DeviceArrayLike` is no longer a runtime checkable type, so there's no need for it to live in this dictionary).

python/cuda_cccl/cuda/compute/algorithms/_reduce.py
```diff
@@ -169,7 +173,9 @@ def reduce_into(
     Args:
         d_in: Device array or iterator containing the input sequence of data items
         d_out: Device array to store the result of the reduction
-        op: Binary reduction operator
+        op: Binary operator to apply.
+            The signature is ``(T, T) -> T``, where ``T`` is
+            the data type of the initial value ``h_init``.
```
NaderAlAwar, line 178: Important: this new description documentation is missing from `merge_sort`. Should be added there for consistency

Issue #7506, opened 2026-02-04
Add alignment argument to cuda::buffer


pciolkosz:
Currently buffer always passes `alignof(T)` to the `allocate` of the resource.

We should use the environment argument of the buffer constructors / `make_buffer` to allow `cuda::overalign_allocation(size_t)` that would communicate to the buffer construction that a different alignment value should be used instead.
In addition the buffer should store the alignment as a member variable and provide a query function for it.

This way you can do `cuda::buffer<byte>` aligned to some higher value for future casting or `cuda::buffer<float>` casted to float4.

PR #7623, opened 2026-02-11
[libcu++] Add explicit alignment argument to buffer construction

pciolkosz:
Closes https://github.com/NVIDIA/cccl/issues/7506

This PR adds `allocation_alignment` tag that can be passed to the buffer through env argument, which will be then passed to the resource on allocation and also stored in the buffer for later access and copy construction

### DIFF
--- a/libcudacxx/include/cuda/__memory_resource/allocation_alignment.h
+++ b/libcudacxx/include/cuda/__memory_resource/allocation_alignment.h
@@ -35,8 +35,9 @@
 
 #  include <cuda/std/__concepts/concept_macros.h>
 #  include <cuda/std/__concepts/convertible_to.h>
-#  include <cuda/std/__exception/throw_error.h>
+#  include <cuda/std/__exception/exception_macros.h>
 #  include <cuda/std/__execution/env.h>
+#  include <cuda/std/__host_stdlib/stdexcept>
 #  include <cuda/std/cstddef>
 
 #  include <cuda/std/__cccl/prologue.h>
@@ -95,9 +96,9 @@ __validate_allocation_alignment(::cuda::std::size_t __alignment, ::cuda::std::si
 {
   if (!__is_valid_allocation_alignment(__alignment, __min_alignment))
   {
-    ::cuda::std::__throw_invalid_argument(
-      "Invalid allocation alignment: must be a power of two and at least the "
-      "type's alignment.");
+    _CCCL_THROW(::std::invalid_argument,
+                "Invalid allocation alignment: must be a power of two and at least the "
+                "type's alignment.");
   }
 }
 


libcudacxx/test/libcudacxx/cuda/containers/buffer/constructor.cu
```diff
@@ -97,6 +105,14 @@ C2H_CCCLRT_TEST("cuda::buffer constructors", "[container][buffer]", test_types)
       CCCLRT_CHECK(check_offseted_pointer(buf.data()));
       CCCLRT_CHECK(buf.size() == 5);
     }
+
+    { // from size with no_init and allocation_alignment env (offset_by_alignment_resource verifies alignment)
+      const ::cuda::std::size_t alignment = ::cuda::mr::default_cuda_malloc_alignment / 2;
+      const auto env                      = ::cuda::std::execution::prop{::cuda::allocation_alignment, alignment};
```
jrhemstad, line 112: question: Is this the expected API most users would use? If so, this is an awfully verbose way to simply specify an allocation alignment.

pciolkosz: I plan to discuss some conventions around environment based APIs next week, so we can deep dive into this topic.

But for non-default alignment specifically, I consider it a ninja feature and I am not against it being verbose. But we could try to simplify it, for example having `::cuda::allocation_alignment` wrap the alignment value instead of relying on `::cuda::std::execution::prop`

jrhemstad: I wouldn't be too quick to assume that non-default alignment is a ninja feature. Many features essential for performance, like vectorized load/stores or TMA, require over-aligned allocations

libcudacxx/include/cuda/__container/buffer.h
```diff
@@ -135,6 +136,22 @@ class buffer
...
+  {
+    ::cuda::std::size_t __a;
+    if constexpr (::cuda::std::execution::__queryable_with<const _Env&, allocation_alignment_t>)
+    {
+      __a = ::cuda::allocation_alignment(__env);
+    }
+    else
+    {
+      __a = alignof(_Tp);
+    }
```
miscco: This should be
```suggestion
    const size_t __a= ::cuda::call_or(::cuda::allocation_alignment, alignof(_Tp), __env);
```

libcudacxx/include/cuda/__memory_resource/allocation_alignment.h
```diff
@@ -0,0 +1,93 @@
...
+  {
+    static_assert(noexcept(__env.query(*this)));
+    return __env.query(*this);
+  }
+
+  [[nodiscard]]
+  _CCCL_NODEBUG_API static constexpr auto query(::cuda::std::execution::forwarding_query_t) noexcept -> bool
+  {
+    return true;
+  }
```
miscco, line 78: The buffers provide the `alignment()` method. I believe we should make sure that we can pass a buffer to `allocation_alignment` and get the alignment out of it

libcudacxx/test/libcudacxx/cuda/containers/buffer/constructor.cu
```diff
@@ -14,14 +14,15 @@
...
 #include <cuda/std/array>
 #include <cuda/std/cassert>
+#include <cuda/std/cstddef>
 #include <cuda/std/initializer_list>
 #include <cuda/std/tuple>
 #include <cuda/std/type_traits>
 
 #include <stdexcept>
 
+#include "../test_resources.h"
```
miscco: This should not be relative, all tests should include the test/support folder
```suggestion
#include "test_resources.h"
```

libcudacxx/test/libcudacxx/cuda/containers/buffer/constructor.cu
miscco: I believe we default case should be no alignment specified as is done here. If a user wants to specify the alignment they should be prepared to jump through a few hoops

That said, I believe this isn't too bad

libcudacxx/test/libcudacxx/cuda/containers/buffer/constructor.cu
```diff
@@ -192,15 +208,29 @@ C2H_CCCLRT_TEST("cuda::buffer constructors", "[container][buffer]", test_types)
...
       const Buffer input{stream, resource, {T(1), T(42), T(1337), T(0), T(12), T(-1)}};
       Buffer buf(input);
+      CCCLRT_CHECK(buf.alignment() == input.alignment());
+      CCCLRT_CHECK(is_pointer_aligned(buf.data(), buf.alignment()));
       CCCLRT_CHECK(check_offseted_pointer(buf.data()));
       CCCLRT_CHECK(!buf.empty());
       CCCLRT_CHECK(equal_range(buf));
     }
+
+    { // copy construction preserves custom allocation_alignment
```
miscco, line 227: Important: I would also want to have tests that verify that we can use `cuda::allocation_alignment` on our buffer types and get the right result

libcudacxx/include/cuda/__container/buffer.h
```diff
@@ -135,6 +136,14 @@ class buffer
     return const_cast<__resource_t&>(__buf_.memory_resource());
   }
 
+  template <class _Env>
+  static size_t __alignment_from_env(const _Env& __env)
+  {
+    const ::cuda::std::size_t __a = ::cuda::__call_or(::cuda::allocation_alignment, alignof(_Tp), __env);
+    ::cuda::__validate_allocation_alignment(__a, alignof(_Tp));
+    return __a;
```
miscco: Nitpick: Would prefer a better name
```suggestion
    const auto __align = ::cuda::__call_or(::cuda::allocation_alignment, alignof(_Tp), __env);
    ::cuda::__validate_allocation_alignment(__align, alignof(_Tp));
    return __align;
```

libcudacxx/include/cuda/__container/buffer.h
```diff
@@ -507,6 +527,12 @@ class buffer
     return __buf_.size() == 0;
   }
 
+  //! @brief Returns the alignment used for the allocation.
+  [[nodiscard]] _CCCL_HIDE_FROM_ABI constexpr size_type alignment() const noexcept
```
miscco: Should that be
```suggestion
  [[nodiscard]] _CCCL_HOST_API constexpr size_type alignment() const noexcept
```

libcudacxx/test/libcudacxx/cuda/containers/uninitialized_async_buffer.cu
```diff
@@ -16,12 +16,14 @@
...
 #include <cuda/memory_resource>
 #include <cuda/std/cassert>
+#include <cuda/std/cstddef>
 #include <cuda/std/cstdint>
 #include <cuda/std/span>
 #include <cuda/std/type_traits>
 #include <cuda/std/utility>
 #include <cuda/stream>
 
+#include "test_resources.h"
```
miscco: the other tests have 
```suggestion
#include <test_resources.h>
```

miscco (approved):

PR #7625, opened 2026-02-11
cuda.compute: Enable directly passing LTO-IR for custom operators

shwina:
## Description

Closes https://github.com/NVIDIA/cccl/issues/7354.

This PR enables passing pre-compiled LTOIR directly to algorithms (and `TransformIterator`) and bypassing Numba JIT, via a `RawOp` type. A `RawOp` can be passed anywhere we currently accept a Python function/lambda:

```python
op = RawOp(ltoir=ltoir, name="add")
cuda.compute.reduce_into(d_input, d_output, op, num_items, h_init)
```

Importantly, the calling convention of the compiled device function must use all `void*` arguments (this is specified in the docstring of `RawOp` as well as the documentation):

```c++
void add(void* arg1, void* arg2, void* result);
```

## Checklist

### DIFF
--- a/python/cuda_cccl/cuda/compute/op.py
+++ b/python/cuda_cccl/cuda/compute/op.py
@@ -108,7 +108,6 @@ class RawOp(_OpAdapter):
     Args:
         name: The ABI name of the operator
         ltoir: bytes object containing the LTO-IR of the compiled operator
-        extra_ltoirs: Optional list of additional LTOIR modules that the main LTOIR depends on
         state: Optional bytes representing the operator's state
         state_alignment: Alignment requirement for the state bytes (default: 1)
         extra_ltoirs: Optional list of additional LTO-IRs to include during linking
@@ -137,15 +136,6 @@ class RawOp(_OpAdapter):
         state_alignment: int = 1,
         extra_ltoirs: list[bytes] | None = None,
     ):
-        """
-        Create an operator from LTO-IR.
-
-        Args:
-            ltoir: Compiled LTOIR bytecode from any language/compiler
-            name: Function name in the LTOIR
-            state: State bytes for stateful operators (default: b"")
-            state_alignment: Memory alignment for state (default: 1)
-        """
         self._ltoir = ltoir
         self._name = name
         self._state = state
--- a/python/cuda_cccl/tests/compute/examples/raw_op/cpp_stateful.py
+++ b/python/cuda_cccl/tests/compute/examples/raw_op/cpp_stateful.py
@@ -90,7 +90,7 @@ d_input = cp.array(h_input)
 
 # Allocate output arrays
 d_output = cp.empty(num_items, dtype=np.int32)
-d_num_selected = cp.zeros(1, dtype=np.int32)
+d_num_selected = cp.empty(1, dtype=np.int32)
 
 # Run select with the stateful operator
 cuda.compute.select(d_input, d_output, d_num_selected, select_op, num_items)
--- a/python/cuda_cccl/tests/compute/examples/raw_op/cpp_stateless.py
+++ b/python/cuda_cccl/tests/compute/examples/raw_op/cpp_stateless.py
@@ -55,7 +55,7 @@ multiply_op = RawOp(ltoir=ltoir_bytes, name="multiply_op")
 # Prepare test data
 h_input = np.array([1, 2, 3, 4, 5], dtype=np.int32)
 d_input = cp.array(h_input)
-d_output = cp.zeros(1, dtype=np.int32)
+d_output = cp.empty(1, dtype=np.int32)
 h_init = np.array(1, dtype=np.int32)
 
 # Use the custom operator with reduce_into
--- a/python/cuda_cccl/tests/compute/test_raw_op.py
+++ b/python/cuda_cccl/tests/compute/test_raw_op.py
@@ -10,7 +10,9 @@ import cupy as cp
 import numpy as np
 import pytest
 
+import cuda.compute
 from cuda.compute import types
+from cuda.compute._cpp_compile import _get_include_paths
 from cuda.compute.op import RawOp
 from cuda.core import Device, Program, ProgramOptions
 
@@ -26,12 +28,15 @@ def get_arch():
     return f"sm_{cc_major}{cc_minor}"
 
 
-def compile_to_ltoir(source: str, arch: str) -> bytes:
+def compile_to_ltoir(
+    source: str, arch: str, include_paths: list | None = None
+) -> bytes:
     """Compile C++ source to LTOIR using cuda.core."""
     opts = ProgramOptions(
         arch=arch,
         relocatable_device_code=True,
         link_time_optimization=True,
+        include_path=include_paths,
     )
     prog = Program(source, "c++", options=opts)
     return prog.compile("ltoir").code
@@ -45,7 +50,7 @@ def extract_function_name(source: str) -> str:
     raise ValueError("Could not extract function name from C++ source.")
 
 
-def make_cpp_op(source: str, name: str = None) -> RawOp:
+def make_cpp_op(source: str, name: str = None, include_paths=None) -> RawOp:
     """
     Compile C++ source to LTOIR and create a stateless RawOp.
 
@@ -60,7 +65,7 @@ def make_cpp_op(source: str, name: str = None) -> RawOp:
         name = extract_function_name(source)
 
     arch = get_arch()
-    ltoir = compile_to_ltoir(source, arch)
+    ltoir = compile_to_ltoir(source, arch, include_paths=include_paths)
 
     return RawOp(ltoir=ltoir, name=name)
 
@@ -108,11 +113,9 @@ def test_cpp_op_basic_add():
     num_items = 100
     h_input = np.arange(num_items, dtype=np.int32)
     d_input = cp.array(h_input)
-    d_output = cp.zeros(1, dtype=np.int32)
+    d_output = cp.empty(1, dtype=np.int32)
 
     # Use the custom op with reduce_into
-    import cuda.compute
-
     h_init = np.array(0, dtype=np.int32)
     cuda.compute.reduce_into(d_input, d_output, op, num_items, h_init)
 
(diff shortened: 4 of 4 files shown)


python/cuda_cccl/cuda/compute/op.py
```diff
@@ -89,6 +89,105 @@ def __hash__(self):
...
+            :language: python
+            :start-after: # example-begin
+
+    Args:
+        name: The ABI name of the operator
+        ltoir: bytes object containing the LTO-IR of the compiled operator
+        extra_ltoirs: Optional list of additional LTOIR modules that the main LTOIR depends on
+        state: Optional bytes representing the operator's state
+        state_alignment: Alignment requirement for the state bytes (default: 1)
+        extra_ltoirs: Optional list of additional LTO-IRs to include during linking
```
NaderAlAwar, line 113: Important: this seems to be duplicated from the constructor below. Also `extra_ltoirs` is mentioned twice

python/cuda_cccl/tests/compute/examples/raw_op/cpp_stateful.py
```diff
@@ -0,0 +1,119 @@
...
+)
+
+# Prepare test data: numbers 0 to 19
+num_items = 20
+h_input = np.arange(num_items, dtype=np.int32)
+d_input = cp.array(h_input)
+
+# Allocate output arrays
+d_output = cp.empty(num_items, dtype=np.int32)
+d_num_selected = cp.zeros(1, dtype=np.int32)
```
NaderAlAwar: Nit: this does not need to be zero initialized, you should use `cp.empty`. three_way_partition initializes it to 0 before calling the kernel

python/cuda_cccl/tests/compute/examples/raw_op/cpp_stateless.py
```diff
@@ -0,0 +1,70 @@
...
+arch = get_arch()
+ltoir_bytes = compile_cpp_to_ltoir(cpp_source, arch)
+
+# Create a RawOp from the LTOIR bytecode
+multiply_op = RawOp(ltoir=ltoir_bytes, name="multiply_op")
+
+# Prepare test data
+h_input = np.array([1, 2, 3, 4, 5], dtype=np.int32)
+d_input = cp.array(h_input)
+d_output = cp.zeros(1, dtype=np.int32)
```
NaderAlAwar: Nit: does not need to be zero initialized

python/cuda_cccl/tests/compute/test_raw_op.py
```diff
@@ -0,0 +1,514 @@
...
+    }
+    """
+
+    op = make_cpp_op(cpp_source, "add_op")
+
+    # Create test data
+    num_items = 100
+    h_input = np.arange(num_items, dtype=np.int32)
+    d_input = cp.array(h_input)
+    d_output = cp.zeros(1, dtype=np.int32)
```
NaderAlAwar: Nit: don't zero initialize, same with other tests below

python/cuda_cccl/tests/compute/test_raw_op.py
```diff
@@ -0,0 +1,514 @@
...
+
+    # Use the custom op with reduce_into
+    import cuda.compute
```
NaderAlAwar: Question: any reason to import this here? Same with other tests below

python/cuda_cccl/tests/compute/test_raw_op.py
```diff
@@ -0,0 +1,514 @@
...
+    assert result == expected, f"Expected {expected}, got {result}"
+
+
+def test_cpp_op_max():
+    """Test a C++ max operator with reduce_into."""
+    cpp_source = """
+    extern "C" __device__ void max_op(void* a, void* b, void* result) {
+        float va = *static_cast<float*>(a);
+        float vb = *static_cast<float*>(b);
+        *static_cast<float*>(result) = va > vb ? va : vb;
```
NaderAlAwar: Question: could we use `cuda::std::max()` here? We may need to add an include. Would be good to test and use it here if possible

python/cuda_cccl/cuda/compute/op.py
shwina: Removed the docstring from `__init__`.

python/cuda_cccl/tests/compute/test_raw_op.py
shwina: Yes - changed the test accordingly.

python/cuda_cccl/tests/compute/examples/raw_op/cpp_stateful.py
```diff
@@ -0,0 +1,119 @@
...
+from cuda.compute.op import RawOp
+from cuda.core import Device, Program, ProgramOptions
+
+
+def get_arch():
+    """Get the SM architecture string for the current device."""
+    device = Device()
+    device.set_current()
+    cc_major, cc_minor = device.compute_capability
+    return f"sm_{cc_major}{cc_minor}"
```
leofang, line 29: nits:
1. It is better to not hide `set_current()` inside a function; making it as the first line of the script (before `cp.zeros()`) would express the intent better
2. In fact, it can be as simple as `device = Device(); arch: str = device.arch`. Getting the CC (and arch, which is a shortcut of assembling the CC into a str) does not need any context to be set to current; initializing CUDA (via `Device()`) is enough.

python/cuda_cccl/tests/compute/examples/raw_op/cpp_stateful.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+def compile_cpp_to_ltoir(source: str, arch: str) -> bytes:
```
leofang, line 32: nit: to make super compact example we can omit `arch` here; if it is not passed to `ProgramOptions`, the current arch is assumed

python/cuda_cccl/tests/compute/examples/raw_op/cpp_stateful.py
```diff
@@ -0,0 +1,119 @@
...
+    )
+    prog = Program(source, "c++", options=opts)
+    return prog.compile("ltoir").code
+
+
+# Create a device counter initialized to 0
+d_counter = cp.zeros(1, dtype=np.int32)
+
+# Pack the counter pointer as state bytes
+counter_ptr = d_counter.__cuda_array_interface__["data"][0]
```
leofang, line 47: Note: (faster) alternative is `d_counter.data.ptr`

python/cuda_cccl/tests/compute/examples/raw_op/cpp_stateful.py
```diff
@@ -0,0 +1,119 @@
...
+"""
+
+# Compile C++ to LTOIR
+arch = get_arch()
+ltoir_bytes = compile_cpp_to_ltoir(cpp_source, arch)
+
+# Create a stateful RawOp with the state bytes
+select_op = RawOp(
+    ltoir=ltoir_bytes,
+    name="select_even_with_count",
```
leofang, line 81: nit: probably a good idea to add an inline comment that `name` needs to match that in `cpp_source`?

python/cuda_cccl/tests/compute/examples/raw_op/cpp_stateful.py
```diff
@@ -0,0 +1,119 @@
...
+    state=state_bytes,
```
leofang, line 82: Q: Why do we need a host state (`state_bytes`) and a device state (`d_counter`)? Would be good to explain their relation (and if they have potential lifetime/ownership issues).

python/cuda_cccl/tests/compute/examples/raw_op/cpp_stateful.py
```diff
@@ -0,0 +1,119 @@
...
+# Allocate output arrays
+d_output = cp.empty(num_items, dtype=np.int32)
+d_num_selected = cp.empty(1, dtype=np.int32)
+
+# Run select with the stateful operator
+cuda.compute.select(d_input, d_output, d_num_selected, select_op, num_items)
+
+# Get results
+num_selected = d_num_selected.get()[0]
+counter_value = d_counter.get()[0]
```
leofang, line 100: ditto, here we use the result from the device state (`d_counter`), but not the host state (`state_bytes`), which can be confusing

docs/python/compute.rst
```diff
@@ -286,6 +285,45 @@ To clear all caches and free memory:
...
+
+   **Required calling convention**: Compiled functions must use untyped pointers for
+   all parameters, with manual type casting inside the function. In C++, this means
+   all arguments (and the return value) must be passed as ``void*`` pointers:
+
+   .. code-block:: cpp
+
+      extern "C" __device__ void my_binary_op(void* a, void* b, void* result) {
+          *static_cast<int*>(result) = *static_cast<int*>(a) + *static_cast<int*>(b);
+      }
```
leofang, line 316: - It would be nice to give one or two concrete examples on how a naively written C++ device function should be type-erased to `void*`. Right now, this is not very clear on how scalar types and pointer types should be erased.
- It would be nice to add a `.. notes::` block below to explain to curious readers why type erasure is needed (IIRC it has to do with ODR issues)

python/cuda_cccl/cuda/compute/op.py
```diff
@@ -94,10 +89,121 @@ def __hash__(self):
...
+        - The provided LTO-IR must define a function with the specified name and the correct signature.
+        - The function must use untyped pointers for all parameters and return type. The function body
+          is responsible for correctly interpreting the pointer arguments based on the expected input and output types.
+          For stateless operators, the signature is
+
+             void func(void* arg1, void* arg2, ..., void* result)`
+
+          For stateful operators, the first parameter must be a pointer to the state:
+
+             void func(void* state, void* arg1, void* arg2, ...)
```
leofang, line 125: ditto; expanding the notes either here or in `compute.rst`

python/cuda_cccl/cuda/compute/op.py
```diff
@@ -94,10 +89,121 @@ def __hash__(self):
...
+
+    __slots__ = ["_ltoir", "_name", "_state", "_state_alignment", "_extra_ltoirs"]
+
+    def __init__(
+        self,
+        *,
+        ltoir: bytes,
+        name: str,
+        state: bytes = b"",
+        state_alignment: int = 1,
```
leofang, line 136: Q: We probably can set `state_alignment: int = None` and default to infer from `bytes` (say, `max(state, 16)`)?

python/cuda_cccl/tests/compute/conftest.py
leofang, line 1: The whole no-numba machinery is getting cooler and cooler! 🙂

docs/python/compute.rst
```diff
@@ -286,6 +285,45 @@ To clear all caches and free memory:
...
+
+   You must ensure that:
+
+   * All parameters are untyped pointers with manual casting in the function body
+   * Type casts match the actual data types passed at runtime
+   * For stateful operators, state is the first parameter (also an untyped pointer)
+   * State bytes have the correct layout and alignment
```
leofang, line 323: Also need to comment on the return value  being `void`

PR #7632, opened 2026-02-11
Complex acosh accuracy refinement

s-oboyle:
## Update the complex acosh function to avoid numerical issues.

The current complex acosh function loses accuracy in several places.
These mostly relate to over/underflow, and catastrophic cancellation for tough values.

This new version fixes these accuracy issues while improving perf.

## Perf
Using the math-teams standard math_bench test we have the following:

**Operations/SM/cycle:**
`cacosh():`
| --- | --- | --- | --- |
| fp64 | 0.1805 | 0.2535 | 1.40 | 
| fp32 | 0.4160 | 0.6057 | 1.46 |

## Correctness
Apart from the usual over/underflow suspects, there is also some very subtle intervals where accuracy gets badly thrown out, especially by catastrophic cancellation very close to `+-1`.

### GPU Correctness
```
GPU fp64:
Max ulp real error (4.837,0.01858) @ (0.0002312206021,0.003898933852)   (0x3f2e4e79d9260f37,0x3f6ff0a82c4a63fd)
        Ours = (0.003898924078,1.570565108)    Ref = (0.003898924078,1.570565108)
        Ours = (0x3f6ff0a2ecf2a7a4,0x3ff92108e0ee0513)               Ref = (0x3f6ff0a2ecf2a7a9,0x3ff92108e0ee0513)

Max ulp imag error (1.083,4.033) @ (2.014485587e+109,2.185144625e-84)   (0x56a127bb054c9f31,0x2e90fae48dad16d1)
        Ours = (252.3752862,1.084715938e-193)    Ref = (252.3752862,1.084715938e-193)
        Ours = (0x406f8c02582e1618,0x17dfac5cf0c70e8d)               Ref = (0x406f8c02582e1619,0x17dfac5cf0c70e89)
```

```
GPU fp32:
Max ulp real error (5.493,0.3926) @ (-0.0003716686915,0.003906234866)   (0xb9c2dc87,0x3b7fffbf)
        Ours = (0.003906223923,1.571167946)    Ref = (0.003906225087,1.571167946)
        Ours = (0x3b7fff90,0x3fc91c08)               Ref = (0x3b7fff95,0x3fc91c08)

Max ulp imag error (0.3869,4.068) @ (2.786598495e+30,1.932855031e+13)   (0x720caff1,0x558ca23b)
        Ours = (70.7955246,6.936250648e-18)    Ref = (70.7955246,6.936252303e-18)
        Ours = (0x428d974f,0x22ffe709)               Ref = (0x428d974f,0x22ffe70d)
```

### CPU Correctness
```
CPU fp64:
Max ulp real error (5.532,0.2167) @ (-3.854794742e-08,0.0624438834)     (0xbe64b1fd5340b800,0x3faff8a50a1a3c00)
        Ours = (0.06240337384,1.570796365)    Ref = (0.06240337384,1.570796365)
        Ours = (0x3faff355c3a29bc6,0x3ff921fb5e98057e)               Ref = (0x3faff355c3a29bc0,0x3ff921fb5e98057e)

Max ulp imag error (0.1839,3.671) @ (288934.8792,2.630363514e-178)      (0x4111a29b84468c00,0x1b110de4cc443800)
        Ours = (13.26710379,9.103655197e-184)    Ref = (13.26710379,9.103655197e-184)
        Ours = (0x402a88c1d3fb47b8,0x19eef2269dfd85fd)               Ref = (0x402a88c1d3fb47b8,0x19eef2269dfd8601)
```

```
CPU fp32:
Max ulp real error (4.414,0.282) @ (0.001004750957,-0.03124837391)      (0x3a83b1d9,0xbcfffc97)
        Ours = (0.0312432982,-1.569792032)    Ref = (0.03124330565,-1.569792032)
        Ours = (0x3cfff1f2,0xbfc8eef2)               Ref = (0x3cfff1f6,0xbfc8eef2)

Max ulp imag error (0.2415,4.52) @ (3.313251063e+26,4.860024734e+24)    (0x6b890869,0x6880a4cf)
        Ours = (61.75839615,0.01466739923)    Ref = (61.75839615,0.01466739457)
        Ours = (0x42770899,0x3c704f88)               Ref = (0x42770899,0x3c704f83)
```

### DIFF
--- a/libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
+++ b/libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
@@ -545,7 +545,7 @@ template <class _Tp>
   // real = (1.0 - eps) and imag ~ 0.0 but it's nowhere near as bad as in asinh.
   // We need a much smaller interval check than asinh in this case.
   // Bounds found experimentally.
-  constexpr _Tp __realx_close_bound = (is_same_v<_Tp, float>) ? _Tp{0.99f}; : _Tp{0.99999};
+  constexpr _Tp __realx_close_bound = (is_same_v<_Tp, float>) ? _Tp{0.99f} : _Tp{0.99999};
   constexpr _Tp __imagx_small_bound = (is_same_v<_Tp, float>) ? _Tp{1.0e-6f} : _Tp{1.0e-8};
 
   // Calculate the same thing, but in a more stable way on this interval.


libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
```diff
@@ -406,44 +406,282 @@ _CCCL_API inline complex<__half> asinh(const complex<__half>& __x) noexcept
...
+    if ((__x.real() == numeric_limits<_Tp>::infinity()) && ::cuda::std::isinf(__imagx))
     {
-      return complex<_Tp>(::cuda::std::abs(__x.imag()), __x.real());
+      return complex<_Tp>(numeric_limits<_Tp>::infinity(), ::cuda::std::copysign(_Tp{0.25} * __pi, __x.imag()));
+    }
+
+    // If z is (NaN,+inf), the result is (+inf,NaN).
+    if (::cuda::std::isnan(__realx) && ::cuda::std::isinf(__imagx))
+    {
+      return complex<_Tp>(numeric_limits<_Tp>::infinity(), NAN);
```
davebayer: Don't use `NAN`, please, use `numeric_limits<_Tp>::quiet_NaN()` instead

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
```diff
@@ -406,44 +406,282 @@ _CCCL_API inline complex<__half> asinh(const complex<__half>& __x) noexcept
...
+  // We need a much smaller interval check than asinh in this case.
+  // Bounds found experimentally.
+  _Tp __realx_close_bound = _Tp{0.99999};
+  _Tp __imagx_small_bound = _Tp{1.0e-8};
+
+  if constexpr (is_same_v<_Tp, float>)
+  {
+    __realx_close_bound = _Tp{0.99f};
+    __imagx_small_bound = _Tp{1.0e-6f};
+  }
```
davebayer: I think we can just do
```suggestion
  constexpr auto __realx_close_bound = (is_same_v<_Tp, float>) ? _Tp{0.99f}; : _Tp{0.99999};
  constexpr auto __imagx_small_bound = (is_same_v<_Tp, float>) ? _Tp{1.0e-6f} : _Tp{1.0e-8};
```

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
```diff
@@ -406,44 +406,282 @@ _CCCL_API inline complex<__half> asinh(const complex<__half>& __x) noexcept
...
+  const __cccl_asinh_sqrt_return_hilo<_Tp> __extended_sqrt_hilo =
+    ::cuda::std::__internal_double_Tp_sqrt_unsafe<_Tp>(__inner_most_term_hi, __inner_most_term_lo);
+
+  _Tp __extended_sqrt_hi = __extended_sqrt_hilo.__hi;
+  _Tp __extended_sqrt_lo = __extended_sqrt_hilo.__lo;
+
+  // 0.0, and some very particular values, do not survive this unsafe sqrt function.
+  // This case occurs when (1 + x^2) is zero or denormal. (and rsqrt(x)*rsqrt(x) become inf).
+  constexpr __uint_t __min_normal_bits = __uint_t{0x1} << __mant_nbits;
+  const _Tp __min_normal               = ::cuda::std::__fp_from_storage<_Tp>(__min_normal_bits);
```
davebayer: You can just use `::cuda::std::__fp_min<_Tp>()`

davebayer: or `numeric_limits<_Tp>::min()`

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
```diff
@@ -406,44 +406,282 @@ _CCCL_API inline complex<__half> asinh(const complex<__half>& __x) noexcept
...
+
+  // This reuses the sqrt calculated on CPU already in __recip_sqrt,
+  // And gets sqrt quickly on device using the rsqrt already calculated.
+#if _CCCL_CUDA_COMPILATION()
+  NV_IF_ELSE_TARGET(NV_IS_DEVICE,
+                    (__pos_evaluation_real = (__recip_sqrt * __inside_sqrt_term);),
+                    (__pos_evaluation_real = ::cuda::std::sqrt(__inside_sqrt_term);))
+#else
+  __pos_evaluation_real = ::cuda::std::sqrt(__inside_sqrt_term);
+#endif // _CCCL_CUDA_COMPILATION()
```
davebayer: NV target macros can be used always :)
```suggestion
  NV_IF_ELSE_TARGET(NV_IS_DEVICE,
                    (__pos_evaluation_real = (__recip_sqrt * __inside_sqrt_term);),
                    (__pos_evaluation_real = ::cuda::std::sqrt(__inside_sqrt_term);))
```

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
```diff
@@ -406,44 +406,282 @@ _CCCL_API inline complex<__half> asinh(const complex<__half>& __x) noexcept
...
+    // but not small enough that the acosh(x) ~ log(2x) estimate does
+    // not break down. We are not able to reduce this with a single simple reduction,
+    // so we do a fast/inlined frexp/ldexp:
+    const int32_t __exp_biased = static_cast<int32_t>(::cuda::std::__fp_get_storage(__max) >> __mant_nbits);
+
+    // Get a factor such that (__max * __exp_mul_factor) <= __max_allowed_exponent
+    const __uint_t __exp_reduce_factor =
+      (__uint_t((2 * __exp_max) + __max_allowed_exponent - __exp_biased) << __mant_nbits)
+      | __fp_explicit_bit_mask_of_v<_Tp>;
+    const _Tp __exp_mul_factor = ::cuda::std::__fp_from_storage<_Tp>(__exp_reduce_factor);
```
davebayer, line 500: Why don't you use `::cuda::std::__fp_set_exp_biased` here? Please, try to use as many of the `__fp_meow` functions as possible, if there is something missing, please add them/tell us to add them.

I'd like to see as few manual floating point types constructions as possible. They are hard to read and error prone.

miscco: We should make sure there is no perf difference here though

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
```diff
@@ -406,44 +406,282 @@ _CCCL_API inline complex<__half> asinh(const complex<__half>& __x) noexcept
...
 [[nodiscard]] _CCCL_API inline complex<_Tp> acosh(const complex<_Tp>& __x)
 {
-  constexpr _Tp __pi = __numbers<_Tp>::__pi();
-  if (::cuda::std::isinf(__x.real()))
+  // Uint of the same size as our fp type.
+  using __uint_t = __fp_storage_of_t<_Tp>;
+
+  constexpr int32_t __mant_nbits = __fp_mant_nbits_v<__fp_format_of_v<_Tp>>;
+  constexpr int32_t __exp_max    = __fp_exp_max_v<__fp_format_of_v<_Tp>>;
+  constexpr int32_t __exp_bias   = __fp_exp_bias_v<__fp_format_of_v<_Tp>>;
```
fbusato, line 409:
```suggestion
  constexpr int __mant_nbits = __fp_mant_nbits_v<__fp_format_of_v<_Tp>>;
  constexpr int __exp_max    = __fp_exp_max_v<__fp_format_of_v<_Tp>>;
  constexpr int __exp_bias   = __fp_exp_bias_v<__fp_format_of_v<_Tp>>;
```

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
```diff
@@ -406,44 +406,282 @@ _CCCL_API inline complex<__half> asinh(const complex<__half>& __x) noexcept
...
+
+  constexpr _Tp __pi  = __numbers<_Tp>::__pi();
+  constexpr _Tp __ln2 = __numbers<_Tp>::__ln2();
```
fbusato, line 412:
```suggestion
  constexpr auto __pi  = __numbers<_Tp>::__pi();
  constexpr auto __ln2 = __numbers<_Tp>::__ln2();
```

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
```diff
@@ -406,44 +406,282 @@ _CCCL_API inline complex<__half> asinh(const complex<__half>& __x) noexcept
...
+
+  _Tp __realx = ::cuda::std::fabs(__x.real());
+  _Tp __imagx = ::cuda::std::fabs(__x.imag());
+
+  // Special cases that do not pass through:
+  if (!::cuda::std::isfinite(__realx) || !::cuda::std::isfinite(__imagx))
   {
-    if (::cuda::std::isnan(__x.imag()))
+    // If z is (x,+inf) (for any finite x), the result is (+inf,pi/2).
+    if (::cuda::std::isfinite(__realx) && ::cuda::std::isinf(__imagx))
```
fbusato, line 421: not sure how the compiler optimizes these lines. We could avoid repeating the same functions, or even use `isfinite(__realx) + isfinite(__imagx)` and check the cases.

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
fbusato: we could also add `infinity` here as it is used multiple times

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
```diff
@@ -406,44 +406,282 @@ _CCCL_API inline complex<__half> asinh(const complex<__half>& __x) noexcept
...
+    const _Tp __imag_ret_value = (__x.real() == _Tp{1}) ? _Tp{0} : __pi;
+    return complex<_Tp>(_Tp{0}, ::cuda::std::copysign(__imag_ret_value, __x.imag()));
+  }
+
+  // The easiest place to account for large values appears to be here at the start.
+  // Use acosh(x) ~ log(2x) for large x.
+  // Get the largest exponent that passes through the algorithm without issue.
+  // This is ~(max_exponent / 4), with a small bias to make sure edge cases get caught
+  // ~254 for double, ~30 for float
+  constexpr int32_t __max_allowed_exponent = (__exp_max / 4) - 2;
```
fbusato, line 479:
```suggestion
  constexpr int __max_allowed_exponent = (__exp_max / 4) - 2;
```

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
```diff
@@ -406,44 +406,282 @@ _CCCL_API inline complex<__half> asinh(const complex<__half>& __x) noexcept
...
+    const _Tp __exp_mul_factor = ::cuda::std::__fp_from_storage<_Tp>(__exp_reduce_factor);
+
+    // Scale down to a working range.
+    __realx *= __exp_mul_factor;
+    __imagx *= __exp_mul_factor;
+
+    __x_big_factor = static_cast<_Tp>((__exp_biased - __exp_max) - __max_allowed_exponent) * __ln2;
+  }
+
+  const _Tp __diffx_m1 = (__realx - _Tp{1}) * (__realx + _Tp{1}) - __imagx * __imagx;
```
fbusato, line 509: there are several cases where we could use fma. Is intentional, right?

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
davebayer: I believe `auto` is the best option

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
s-oboyle: Missed that one, thanks.

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
s-oboyle: I had somewhat left it this in anticipation of some future needed bounds for __float128, or if half/bfloat16 want to use this template instead of calling the fp32 version.
But I'll change it for now if you think it's nicer.

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
s-oboyle: True, I forgot this wasn't the minimal denormal.

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
miscco: Nah, I would prefer if we would stay precise here, those are nontrival algorithms and I want to know what the exact constraints are

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
miscco: we could pull those out into variables?

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
s-oboyle: I mentioned before it would be nice to have some specific versions (for asinh, [here](https://github.com/NVIDIA/cccl/pull/6428#discussion_r2565176332) and [here](https://github.com/NVIDIA/cccl/pull/6428#discussion_r2565186089)), just  some more specific versions for only positive values. Would this be possible?

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
s-oboyle: Generally yes, but I'm not sure how good an idea it is. I'm somewhat assuming fp_contract is turned on, which is maybe a bad assumption.
Here, it doesn't matter how this gets contracted so I left it like this to let the compiler decide.
In other cases, it's things like `2.0*x + y`, where the result is exactly the same whether fma is used or not, so I again let the compiler decide.
Thoughts on if this is a good / bad thing?

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
s-oboyle: If CCCL is mapping these to the (eg) nvcc intrinsics then the compiler is usually very good at optimizing and pipeling these sequences (much better than I can do by hand usually).
They are just masking operations which GPU compilers have been optimizing since time began. (so I'm not sure I would do something like invoking the ALU with a '+' operation).

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
s-oboyle: I was trying to be fairly specific about these.
I don't like using `auto` in the middle of these nuanced mathematical algorithms where it adds an extra step for the maintainer (aka me) to have to keep in mind/maybe debug, if you are okay with me not using it I would like to leave it out.

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
s-oboyle: We had this question in the past [here](https://github.com/NVIDIA/cccl/pull/6428/changes#r2557351724) and [here](https://github.com/NVIDIA/cccl/pull/6428/changes#r2485937825) where there seemed to be little more nuance to it, does it make sense to you? It's a little above my CCCL-knowledge grade!

libcudacxx/test/libcudacxx/std/numerics/complex.number/complex.transcendentals/acosh.pass.cpp
```diff
@@ -88,8 +88,14 @@ __host__ __device__ void test_edges()
     }
     else if (cuda::std::isfinite(testcases[i].real()) && cuda::std::isnan(testcases[i].imag()))
     {
```
s-oboyle, line 90: @miscco from [https://en.cppreference.com/w/cpp/numeric/complex/acosh](https://en.cppreference.com/w/cpp/numeric/complex/acosh), there's an addendum about this corner case. (that points [here](https://open-std.org/JTC1/SC22/WG14/www/docs/n1892.htm#dr_471)) which does seem to have made it into IEEE: https://pubs.opengroup.org/onlinepubs/9799919799/
So I guess it stays?

s-oboyle:
/ok to test [283155d](https://github.com/NVIDIA/cccl/pull/7632/commits/283155dc066fca865f4b6fa6b1e6341e87597457)

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
miscco: I believe if there will be more type bounds, we should turn this into a constexpr function returning a pair and then desugaring via structured bindings

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
miscco: Those are all internal helpers, so we can have as many as we want :)

libcudacxx/test/libcudacxx/std/numerics/complex.number/complex.transcendentals/acosh.pass.cpp
miscco: I would split that into two clauses and link to the DR for the (0, NaN) case

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
miscco: I think I might have been overzealous there

Should be fine to just NV_IF_TARGET, especially if we are not using any runtime functions

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
s-oboyle: Fine by me, I might need someone else to write it though just to make sure it's formatted how you'd prefer.

libcudacxx/test/libcudacxx/std/numerics/complex.number/complex.transcendentals/acosh.pass.cpp
s-oboyle: This is the style it seems to be written in the other if/else clauses surrounding this one, I might leave it to you to fix it up as you see fit.

s-oboyle:
/ok to test [1672bfd](https://github.com/NVIDIA/cccl/pull/7632/commits/1672bfd746a8d0c6cfe3a77f6e9dc41acfda5eb1)

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
miscco: We want to be specific in those functions

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
miscco: I believe we should stay as simple as possible and let the compiler do its magic.

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
miscco: I am a bit scared by changing anything here, we should turn this into an issue, because these kind of operations happen a lot in the different new math functions

libcudacxx/include/cuda/std/__complex/inverse_hyperbolic_functions.h
```diff
@@ -406,44 +401,271 @@ _CCCL_API inline complex<__half> asinh(const complex<__half>& __x) noexcept
...
+  _Tp __inner_most_term_hi = __x_abs_sq_p1_sq_hi - _Tp{4} * __realx_sq_hi;
+  _Tp __inner_most_term_lo = ((__x_abs_sq_p1_sq_hi - __inner_most_term_hi) - _Tp{4} * __realx_sq_hi);
+  // lo parts, all needed:
+  __inner_most_term_lo += __x_abs_sq_p1_sq_lo - _Tp{4} * __realx_sq_lo;
+
+  // Differs from asinh slightly here, we can still get some cancellation when
+  // real = (1.0 - eps) and imag ~ 0.0 but it's nowhere near as bad as in asinh.
+  // We need a much smaller interval check than asinh in this case.
+  // Bounds found experimentally.
+  constexpr _Tp __realx_close_bound = (is_same_v<_Tp, float>) ? _Tp{0.99f}; : _Tp{0.99999};
```
davebayer, line 548:
```suggestion
  constexpr _Tp __realx_close_bound = (is_same_v<_Tp, float>) ? _Tp{0.99f} : _Tp{0.99999};
```

PR #7649, opened 2026-02-11
Only emit NVTX range ctor/dtor in host-only code

bernhardmgruber:


libcudacxx/include/cuda/__nvtx/nvtx.h
```diff
@@ -75,18 +75,40 @@
...
 #if _CCCL_HAS_NVTX3()
-#  include <cuda/std/optional>
-
 #  include <cuda/std/__cccl/prologue.h>
 
 _CCCL_BEGIN_NAMESPACE_CUDA
-namespace detail
-{
-struct NVTXCCCLDomain
+struct __nvtx_cccl_domain
```
miscco, line 81: nitpick, we conventionally use cccl as the prefix, so `__cccl_nvtx_domain`

bernhardmgruber: I noted a TODO

PR #7652, opened 2026-02-12
[cudax] Level queries for this hierarchy groups

davebayer:
This PR implements `level.query(group)` queries fro hierarchy groups.

### DIFF
--- a/libcudacxx/include/cuda/__hierarchy/hierarchy_level_base.h
+++ b/libcudacxx/include/cuda/__hierarchy/hierarchy_level_base.h
@@ -35,6 +35,10 @@
 #  include <cuda/std/__utility/integer_sequence.h>
 #  include <cuda/std/array>
 
+#  if defined(_CUDAX_HIERARCHY)
+#    include <cuda/experimental/__hierarchy/fwd.cuh>
+#  endif // _CUDAX_HIERARCHY
+
 #  include <cuda/std/__cccl/prologue.h>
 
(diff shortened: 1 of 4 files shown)


libcudacxx/include/cuda/__hierarchy/hierarchy_level_base.h
```diff
@@ -314,6 +318,72 @@ struct hierarchy_level_base
...
+  [[nodiscard]] _CCCL_API static constexpr _Tp count_as(const _Group& __group) noexcept
+  {
+    if constexpr (::cuda::std::is_same_v<_Level, typename _Group::level_type>)
+    {
+      return _Tp{1};
+    }
+    else
+    {
+      // todo: Pass __group.hierarchy() to the query.
+      return _Level::template count_as<_Tp>(typename _Group::level_type{} /*, __group.hierarchy()*/);
```
pciolkosz, line 334: Was passing a hierarchy difficult because we only store its type in the group?

pciolkosz (approved):

libcudacxx/include/cuda/__hierarchy/hierarchy_level_base.h
davebayer: No, it's because I haven't implemented yet querying e. g. `cuda::block.count(cuda::cluster, hierarchy)` when `cluster_level` is not part of the `hierarchy`. I will create an issue for that.

Issue #902, opened 2022-09-03
New indirection level for launch bounds


gevtushenko:
Specifying `__launch_bounds__` in the presence of RDC has proven to be troublesome and unreliable. We have to abstract it out so that launch bounds are not specified when RDC is enabled.

PR #7653, opened 2026-02-12
Optimize kernel functors launched by `cuda::launch`

davebayer:
Because we know whether the dimensions are defined at compile time, or not, we can use `__builtin_assume(...)` to pass the information about the builtin hierarchy variable values, so it can better optimize the whole kernel.

Note that this will not work with `nvc++` in CUDA mode.

It can make a huge difference, see https://godbolt.org/z/795oh9fdx


davebayer:
@pciolkosz, can you have a look, please?

libcudacxx/include/cuda/__launch/launch.h
```diff
@@ -47,9 +47,258 @@ _CCCL_BEGIN_NAMESPACE_CUDA
...
+#      undef _CCCL_CLUSTER_SIZE_IN_BLOCKS
+#      undef _CCCL_GRID_DIM_X
+#      undef _CCCL_GRID_DIM_Y
+#      undef _CCCL_GRID_DIM_Z
+#    endif // !_CCCL_CUDA_COMPILER(NVHPC)
+
 template <typename _Config, typename _Kernel, class... _Args>
 __global__ static void __kernel_launcher(const _CCCL_GRID_CONSTANT _Config __conf, _Kernel __kernel_fn, _Args... __args)
 {
+#    if !_CCCL_CUDA_COMPILER(NVHPC)
```
jrhemstad: question: Could/should we not use the same info for specifying `__launch_bounds__` on this kernel?

pciolkosz:
> It can make a huge difference, see https://godbolt.org/z/795oh9fdx

I think something is wrong with the queries implementation, the example above is using `cuda::gpu_thread.rank(cuda::grid, config)`, which should use the config argument to optimize the query without any assumes

davebayer:
> I think something is wrong with the queries implementation, the example above is using cuda::gpu_thread.rank(cuda::grid, config), which should use the config argument to optimize the query without any assumes

Yes, I will fix that, too. But I think we can give the compiler this info anyway, because the user can always use the old api.

fbusato:
@davebayer I'm favor of that but please also test the difference with sm_100+. Newer NVVM provides guarantees on the knowledge of the thread hieranchy ranges.

libcudacxx/include/cuda/__launch/launch.h
```diff
@@ -47,16 +47,331 @@ _CCCL_BEGIN_NAMESPACE_CUDA
 
 #  if _CCCL_CUDA_COMPILATION()
 
-template <typename _Config, typename _Kernel, class... _Args>
+#    if !_CCCL_CUDA_COMPILER(NVHPC)
+
+// clang-cuda replaces the variables with direct nvvm sreg calls, so we want to assume their return values directly.
```
fbusato, line 52: is it worth it?

libcudacxx/include/cuda/__launch/launch.h
```diff
@@ -47,16 +47,331 @@ _CCCL_BEGIN_NAMESPACE_CUDA
...
+
+  using _BlockDesc = typename _Hierarchy::template level_desc_type<block_level>;
+  using _BlockExts = typename _BlockDesc::extents_type;
+
+  if constexpr (_BlockExts::static_extent(0) != __dext)
+  {
+    _CCCL_ASSUME(_CCCL_BLOCK_DIM_X == _BlockExts::static_extent(0));
+    if constexpr (_BlockExts::static_extent(0) == 1)
+    {
+      _CCCL_ASSUME(_CCCL_THREAD_IDX_X == 0);
```
fbusato: can we just say
```cpp
_CCCL_ASSUME(_CCCL_THREAD_IDX_X < _BlockExts::static_extent(0));
```
?

davebayer: Yeah, works, thanks!

libcudacxx/include/cuda/__launch/launch.h
davebayer: I think it is, there are just some problems with lambdas, so I opened an issue on LLVM.

davebayer:
> @davebayer I'm favor of that but please also test the difference with sm_100+. Newer NVVM provides guarantees on the knowledge of the thread hieranchy ranges.

There is a difference even in some very simple cases like: https://godbolt.org/z/8PWjhP5Kc

We can always remove this in the future once it's not needed anymore.

libcudacxx/include/cuda/__launch/launch.h
```diff
@@ -47,16 +47,283 @@ _CCCL_BEGIN_NAMESPACE_CUDA
...
+#    if _CCCL_CUDA_COMPILER(NVCC)
+  && !__nv_is_extended_device_lambda_closure_type(_Kernel)
+#    endif
+  ;
+
+// We create 2 overloads, with/without the __launch_bounds__ attribute. We must use enable_if because of
+// "Pack template parameter must be the last template parameter for a variadic __global__ function template" error when
+// we try to use _CCCL_REQUIRES expression.
+template <class _Config, class _Kernel, class... _Args>
+__global__ static void __launch_bounds__(::cuda::__max_nthreads_per_block<_Config>())
```
jrhemstad: Check the CUB triple chevron launcher. I think we need to omit the `__launch_bounds__` when compiling with nvc++ and potentially when compiling with -rdc=true.

jrhemstad: See https://github.com/NVIDIA/cccl/issues/902

davebayer: I tried to compile `cuda::launch` (even the old one) with nvc++ in CUDA mode and the compilation fails, I need to investigate

davebayer: I introduced the `_CCCL_LAUNCH_BOUNDS(...)` macro that expands to `__launch_bounds__(...)` only if RDC is disabled

davebayer: > I tried to compile `cuda::launch` (even the old one) with nvc++ in CUDA mode and the compilation fails, I need to investigate

Okay, it was an unrelated issue, fixed by #7716

Issue #7654, opened 2026-02-12
Tuning policy selectors should use the compute capability instead of `cuda::arch_id`


bernhardmgruber:
We currently specify tunings in the new design like this (from `tuning_reduce.cuh`:
```c++
struct policy_selector {
  [[nodiscard]] _CCCL_API constexpr auto operator()(::cuda::arch_id arch) const -> reduce_policy {
    if (arch >= ::cuda::arch_id::sm_100)
      return {...};
    if (arch >= ::cuda::arch_id::sm_60)
      return {...};

    return {...}; // base case: SM50
  }
};
```
My rationale for using `arch_id` was that it allows us to differentiate arch-specific IDs from non-specific ones. E.g., we could provide a different tuning when we compile for `sm100a` (B200 only) vs. compiling for `sm100` (B200, and whatever sm103 is).

@davebayer pointed out to me a while back that `cuda::arch_id` was not intended to be comparable, and in fact, comparison should fail to compile. For some reason it worked and that's what I implemented. The logic above is wrong though in the presence of arch-specific ids, since for example `sm90a` has a higher integral value as `sm100`, so if we would select a policy for `sm90a` we would get the `sm100` tuning, which is surely wrong. This could be fixed by checking for arch specific ids, but this gets out of hand quickly:
```c++
struct policy_selector {
  [[nodiscard]] _CCCL_API constexpr auto operator()(::cuda::arch_id arch) const -> reduce_policy {
    // what if a future CTK introduces a new arch-specific id? all policy selectors would be broken.
    if (arch == ::cuda::arch_id::sm_100a)
      // now what? same as sm_100?
    if (arch == ::cuda::arch_id::sm_90a)
      // now what? same as sm_90?

    if (arch >= ::cuda::arch_id::sm_100)
      return {...};
    if (arch >= ::cuda::arch_id::sm_60)
      return {...};

    return {...}; // base case: SM50
  }
};
```
This is not how I want our users and ourselves to write tunings.

The alternative is to switch from `cuda::arch_id` to `cuda::compute_capability` (CC). A CC has the advantage that it is truly an ordered type, so the logic from the first snipped does actually work. But would we loose anything from not being able to provide a tuning to an arch-specific-compiled kernel?  Let's imagine we want to use a dedicated set of tuning values if we compile for `sm100a` exactly, and otherwise something else. We can write that with a compute capability as:
```c++
struct policy_selector {
[...]

PR #7655, opened 2026-02-12
Deprecate `<`, `<=`, `>` and `>=` operators for `cuda::arch_id`

### LINKED ISSUE #8709: [FEA]: Finalize CUB tuning policy selector API
Outcomes of today's meeting:
- We will use (and document) `cuda::compute_capability` inside tuning policy selectors.
- If specific archs are ever needed, the tuning policy selectors can take `cuda::arch_id`/some other type as the parameter and return the specific tuning. No steps towards handling archs instead of ccs will be taken unless it's crucial in the future.
- Compute capability literals might be considered in the future.
- We should deprecated + delete all operators for `arch_id` except for `==` and `!=`.
- We will keep the assertions in `cuda::to_arch_id` for now.

davebayer:
We don't want users to do:
```cpp
if (cuda::arch_id::sm_120 > cuda::arch_id::sm_120a) { /* ... */ }
```
They should get `compute_capability` of the `arch_id` and compare that instead.

We need to finish #7654 first, before merging this one.

### DIFF
--- a/libcudacxx/include/cuda/__device/arch_id.h
+++ b/libcudacxx/include/cuda/__device/arch_id.h
@@ -45,6 +45,34 @@ enum class arch_id : int
 #undef _CCCL_DEFINE_ARCH_SPECIFIC_ID
 };
 
+// todo: = delete these in 4.0.
+#define _CCCL_DEPRECATED_ARCH_ID_COMPARISONS(_OP)                                                                   \
+  CCCL_DEPRECATED_BECAUSE("Comparing cuda::arch_id using operator" _CCCL_TO_STRING(                                 \
+    _OP) " is deprecated and will be deleted in the next major release. Compare cuda::compute_capabilities of the " \
+         "given "                                                                                                   \
+         "cuda::arch_id instead.")
+[[nodiscard]] _CCCL_DEPRECATED_ARCH_ID_COMPARISONS(<) _CCCL_API constexpr bool
+operator<(arch_id __lhs, arch_id __rhs) noexcept
+{
+  return ::cuda::std::to_underlying(__lhs) < ::cuda::std::to_underlying(__rhs);
+}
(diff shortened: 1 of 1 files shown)


libcudacxx/include/cuda/__device/arch_id.h
```diff
@@ -45,6 +45,29 @@ enum class arch_id : int
 #undef _CCCL_DEFINE_ARCH_SPECIFIC_ID
 };
 
+// todo: = delete these in 4.0.
+#define _CCCL_DEPRECATED_ARCH_ID_COMPARISONS(_OP)                                                                     \
+  CCCL_DEPRECATED_BECAUSE("Comparing cuda::arch_id using operator" _CCCL_TO_STRING(                                   \
+    _OP) " is deprecated and will be deleted in next major release. Compare cuda::compute_capabilities of the given " \
```
bernhardmgruber:
```suggestion
    _OP) " is deprecated and will be deleted in the next major release. Compare cuda::compute_capabilities of the given " \
```

libcudacxx/include/cuda/__device/arch_id.h
```diff
@@ -45,6 +45,29 @@ enum class arch_id : int
...
+         "cuda::arch_id instead.")
+[[nodiscard]] _CCCL_DEPRECATED_ARCH_ID_COMPARISONS(<) _CCCL_API bool operator<(arch_id __lhs, arch_id __rhs) noexcept
```
bernhardmgruber: Suggestion: I think we need to mark the operators as `inline` to avoid multiple definitions, since we define the functions in a header file and they are not templates or inside a class.

davebayer: Ah, yeah, I've made those functions `constexpr`, which imply `inline`

bernhardmgruber: That works as well!

bernhardmgruber (approved):

PR #7656, opened 2026-02-12
Add functions for obtaining cuda arch list

davebayer:
This PR provides an internal way to obtain a the `__CUDA_ARCH_LIST__`.

It also changes the way we define architectures. It introduces the `_CCCL_KNOWN_PTX_ARCH_LIST`, which is a list of all PTX architectures known to CCCL. Similar macro is introduced for arch specific architectures.

This way we are sure we don't miss a new architecture in some of our APIs.


libcudacxx/include/cuda/std/__cccl/execution_space.h
```diff
@@ -71,4 +71,10 @@
 #  define _CCCL_TARGET_CONSTEXPR constexpr
 #endif // ^^^ !_CCCL_CUDA_COMPILER(NVHPC) ^^^
 
+//! @brief List of all known PTX architectures supported by this CCCL version.
+#define _CCCL_KNOWN_PTX_ARCH_LIST 60, 61, 62, 70, 75, 80, 86, 87, 88, 89, 90, 100, 103, 110, 120, 121
+
+//! @brief List of all known architecture specific architectures supported by this CCCL version.
+#define _CCCL_KNOWN_PTX_ARCH_SPECIFIC_LIST 90, 100, 103, 110, 120, 121
```
bernhardmgruber: Important: let's not call them PTX arches. They have nothing to do with PTX. They are SM architectures (or maybe compute capabilities). Given `__CUDA_ARCH__` and `__CUDA_ARCH_LIST__` are widely used macros, I would call those `_CCCL_KNOWN_CUDA_ARCH_LIST` or `_CCCL_KNOWN_CUDA_ARCHES`.

libcudacxx/include/cuda/__device/compute_capability.h
```diff
@@ -171,6 +173,45 @@ class compute_capability
...
+#  elif defined(NV_TARGET_SM_INTEGER_LIST)
+  return ::cuda::std::array{compute_capability{_Vs}...};
+#  else // ^^^ has arch list ^^^ / vvv no arch list vvv
+  static_assert(::cuda::std::__always_false_v<decltype(sizeof...(_Vs))>,
+                "This function can be instantiated only when __CUDA_ARCH_LIST__ or NV_TARGET_SM_INTEGER_LIST are "
+                "defined");
+#  endif // ^^^ no arch list ^^^
+}
+
+[[nodiscard]] _CCCL_API _CCCL_CONSTEVAL auto __cc_list() noexcept
```
bernhardmgruber: Suggestion: I think `__cc_list()` is too generic. How about `__target_compute_capabilities()`?

libcudacxx/include/cuda/__device/compute_capability.h
```diff
@@ -171,6 +173,45 @@ class compute_capability
...
+{
+  return ::cuda::std::array{compute_capability{_Vs}...};
+}
+
+[[nodiscard]] _CCCL_API _CCCL_CONSTEVAL auto __all_compute_capabilities() noexcept
+{
+  return ::cuda::__make_all_compute_capabilities<_CCCL_KNOWN_PTX_ARCH_LIST>();
+}
+
+#if _CCCL_CUDA_COMPILATION()
```
miscco, line 187: Is there a reason we need `__all_compute_capabilities` outside of cuda compilation? 
```suggestion
#if _CCCL_CUDA_COMPILATION()
template <int... _Vs>
[[nodiscard]] _CCCL_API _CCCL_CONSTEVAL auto __make_all_compute_capabilities() noexcept
{
  return ::cuda::std::array{compute_capability{_Vs}...};
}

[[nodiscard]] _CCCL_API _CCCL_CONSTEVAL auto __all_compute_capabilities() noexcept
{
  return ::cuda::__make_all_compute_capabilities<_CCCL_KNOWN_PTX_ARCH_LIST>();
}

```

libcudacxx/include/cuda/std/__cccl/execution_space.h
```diff
@@ -71,4 +71,10 @@
 #  define _CCCL_TARGET_CONSTEXPR constexpr
 #endif // ^^^ !_CCCL_CUDA_COMPILER(NVHPC) ^^^
 
+//! @brief List of all known PTX architectures supported by this CCCL version.
+#define _CCCL_KNOWN_PTX_ARCH_LIST 60, 61, 62, 70, 75, 80, 86, 87, 88, 89, 90, 100, 103, 110, 120, 121
```
miscco: No love for SM_53? 😿

libcudacxx/include/cuda/__device/arch_id.h
```diff
@@ -106,13 +79,10 @@ enum class arch_id : int
 {
   switch (__cc.get() * __arch_specific_id_multiplier)
   {
-    case ::cuda::std::to_underlying(arch_id::sm_90a):
-    case ::cuda::std::to_underlying(arch_id::sm_100a):
-    case ::cuda::std::to_underlying(arch_id::sm_103a):
-    case ::cuda::std::to_underlying(arch_id::sm_110a):
-    case ::cuda::std::to_underlying(arch_id::sm_120a):
-    case ::cuda::std::to_underlying(arch_id::sm_121a):
```
miscco, line 114: I dont remember exactly, why arent we considering the `f` families?

libcudacxx/include/cuda/std/__cccl/execution_space.h
davebayer: I agree, I just wanted to be consistent with `_CCCL_PTX_ARCH` macro that we already have :)

libcudacxx/include/cuda/std/__cccl/execution_space.h
davebayer: @pciolkosz's decision, seems like his GTX 950 is not working anymore :D

libcudacxx/include/cuda/__device/arch_id.h
davebayer: I am not sure, @pciolkosz what was the exact reason?

pciolkosz: Just keeping the scope reasonable, but we should add them as well at some point

libcudacxx/include/cuda/std/__cccl/execution_space.h
pciolkosz: I think we should also drop everything from 60 to 70 as its no longer officially supported, but its not urgent

Issue #7622, opened 2026-02-10
[BUG]: cuda.compute caching numpy ignores shape/values


danielfrg:
### Is this a duplicate?

- [x] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cccl/issues) for this bug and that I agree to the [Code of Conduct](CODE_OF_CONDUCT.md)

### Type of Bug

Runtime Error

### Component

cuda.compute (Python)

### Describe the bug

This happens on the histogram benchmark I am doing in https://github.com/NVIDIA/cccl/pull/7341

Looks like the caching in cuda.compute is not using the NumPy shapes/values so when it forced me to do `clear_all_caches()` when having a different set of benchmarks.

This is what Claude had to say:

The `cuda.compute` caching mechanism for `make_*` factory functions uses only the `dtype` of `np.ndarray` arguments when generating cache keys. This causes incorrect cache hits when the same dtype is used with different array values, leading to memory corruption and `cudaErrorIllegalAddress` errors.

## Affected Component

`python/cuda_cccl/cuda/compute/_caching.py` - Line 230-232:

```python
cache_with_registered_key_functions.register(
    np.ndarray, lambda arr: ("numpy.ndarray", arr.dtype)
)
```

### How to Reproduce

This happens on the histogram benchmark I am doing in https://github.com/NVIDIA/cccl/pull/7341

### Expected behavior

Probably should make the cache be aware of the dtypes and shapes

shwina:
Hmm, do you have a minimal repro of when this can result in erroneous results?  Looking at the histogram API it only accepts numpy scalars as inputs, so the shape shouldn't matter: https://nvidia.github.io/cccl/python/compute_api.html#cuda.compute.algorithms.histogram_even

We don't cache device arrays by their shape so that we don't unnecessarily have to recompile for new shapes (and ultimately the array sizes don't matter for caching purposes). However, I'm curious if/when this matters for host arrays.

danielfrg:
I got this small example to reproduce it:

```python
import sys
import cupy as cp
import numpy as np
from cuda.compute import clear_all_caches, make_histogram_even

CLEAN_CACHE = "--clean-cache" in sys.argv


def run_histogram(
    d_samples,
    d_histogram,
    h_num_output_levels,
    h_lower_level,
    h_upper_level,
    num_bins,
    num_elements,
):
    h_num_output_levels[0] = num_bins + 1
    h_lower_level[0] = 0
[...]

PR #7657, opened 2026-02-12
cuda.compute: Fix caching of histogram build objects, taking into consideration privatized smem strategy

shwina:
## Description

Closes https://github.com/NVIDIA/cccl/issues/7622

Histograms use the number of output levels, to determine at compile time, whether or not to use privatized shared memory. We should cache histogram objects while taking this into consideration.

## Checklist

### DIFF
--- a/python/cuda_cccl/cuda/compute/algorithms/_histogram.py
+++ b/python/cuda_cccl/cuda/compute/algorithms/_histogram.py
@@ -173,12 +173,12 @@ def make_histogram_even(
     upper_level_val = h_upper_level[0].item()
     level_dtype = h_lower_level.dtype
 
-    # bins <= 256 uses privatized smem strategy.
-    # We should include this information when caching
-    # compile time artifacts
+    # bins <= 256 uses privatized smem strategy. a different compile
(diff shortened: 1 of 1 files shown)


python/cuda_cccl/cuda/compute/algorithms/_histogram.py
```diff
@@ -137,13 +167,27 @@ def make_histogram_even(
...
+    num_output_levels_val = int(h_num_output_levels[0])
+    lower_level_val = h_lower_level[0].item()
+    upper_level_val = h_upper_level[0].item()
+    level_dtype = h_lower_level.dtype
+
+    # bins <= 256 uses privatized smem strategy.
+    # We should include this information when caching
+    # compile time artifacts
+    num_bins = num_output_levels_val - 1
+    uses_privatized_smem = num_bins <= 256
```
NaderAlAwar, line 181: Important: could you add a comment saying where in `dispatch_histogram.cuh` this came from?

shwina: Done

PR #7658, opened 2026-02-12
Add some high-level doc entrypoint to TMA

bernhardmgruber:
I have no good idea where to put it. It's a CCCL-wide topic. Let me know if you know a better place.


fbusato (approved):
makes sense. I would also add https://nvidia.github.io/cccl/libcudacxx/extended_api/tma.html

PR #7660, opened 2026-02-12
cuda.compute: Faster TransformIterator construction

shwina:
## Description

When constructing `TransformIterator`, we use numba type inference to figure the output type of the operator. This can be relatively expensive, and is relatively easily cached.

```python
# before:
In [4]: %timeit TransformIterator(arr, lambda x: x ** 2)
7.18 ms ± 10.6 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)

# after
In [6]: %timeit TransformIterator(arr, lambda x: x ** 2)
15.3 μs ± 14.3 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```

## Checklist

### DIFF
--- a/python/cuda_cccl/cuda/compute/_jit.py
+++ b/python/cuda_cccl/cuda/compute/_jit.py
@@ -11,7 +11,6 @@ import inspect
 import operator
 import struct
 import textwrap
-import uuid
 from types import new_class
 from typing import TYPE_CHECKING, Callable, Hashable, List, Tuple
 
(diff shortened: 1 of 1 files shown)


python/cuda_cccl/cuda/compute/_jit.py
```diff
@@ -451,22 +451,6 @@ def _infer_return_type(py_func, input_types):
     return _numba_type_to_type_descriptor(return_type)
 
 
-# -----------------------------------------------------------------------------
```
shwina, line 454: This is just stale code

PR #7666, opened 2026-02-13
Split the RLE tuning header

bernhardmgruber:
This makes further refactorings easier. I just moved code around, no functional change.

### DIFF
--- a/cub/cub/device/dispatch/tuning/common.cuh
+++ b/cub/cub/device/dispatch/tuning/common.cuh
@@ -146,19 +146,19 @@ enum class length_size
 };
 
 template <class T>
-constexpr primitive_key is_primitive_key()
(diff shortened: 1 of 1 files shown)


cub/cub/device/dispatch/tuning/common.cuh
```diff
@@ -119,5 +119,60 @@ template <typename It>
...
+enum class key_size
+{
+  _1,
+  _2,
+  _4,
+  _8,
+  _16,
+  unknown
+};
+enum class length_size
```
miscco, line 142: Should this actually be `offset_size`?

bernhardmgruber: I just moved it from the tuning headers without thinking. So I have no idea :D

PR #7669, opened 2026-02-13
Implement the new tuning API for `DeviceRleDispatch`

bernhardmgruber:
Depends on:

- [x] #7668
- [x] #7666
- [x] No SASS diff for `cub.bench.run_length_encode.non_trivial_runs.base` on SM `75;80;86;90;100`
- [x] Refactoring

Fixes: #7532

### DIFF
--- a/cub/cub/device/dispatch/tuning/tuning_rle_non_trivial_runs.cuh
+++ b/cub/cub/device/dispatch/tuning/tuning_rle_non_trivial_runs.cuh
@@ -378,23 +378,24 @@ struct policy_selector
   bool length_is_primitive;
   bool key_is_primitive; // TODO(bgruber): can probably be derived from key_type
 
+  _CCCL_API constexpr auto
+  make_default_policy(BlockLoadAlgorithm block_load_alg, int delay_ctor_key_size, CacheLoadModifier load_mod) const
+  {
+    const int nominal_4B_items_per_thread = 15;
+    const int items_per_thread =
+      ::cuda::std::clamp(nominal_4B_items_per_thread * 4 / key_size, 1, nominal_4B_items_per_thread);
+    return rle_non_trivial_runs_policy{
+      96,
+      items_per_thread,
+      block_load_alg,
+      load_mod,
+      true,
+      BLOCK_SCAN_WARP_SCANS,
+      default_reduce_by_key_delay_constructor_policy(delay_ctor_key_size, sizeof(int), true)};
+  }
+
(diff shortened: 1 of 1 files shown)


bernhardmgruber:
There are substantial SASS diffs, for example in
```
Function : void cub::_V_300400_SM_750_800_860_900_1000::detail::rle::DeviceRleSweepKernel<cub::_V_300400_SM_750_800_860_900_1000::detail::rle::non_trivial_runs::policy_selector_from_types<long, cuda::std::__4::complex<float>>, cuda::std::__4::complex<float> const*, long*, long*, long*, cub::_V_300400_SM_750_800_860_900_1000::ReduceByKeyScanTileState<long, int, true>, cuda::std::__4::equal_to<void>, int, long, cub::_V_300400_SM_750_800_860_900_1000::detail::rle::streaming_context<cuda::std::__4::complex<float> const*, long, long>>(cuda::std::__4::complex<float> const*, long*, long*, long*, cub::_V_300400_SM_750_800_860_900_1000::ReduceByKeyScanTileState<long, int, true>, cuda::std::__4::equal_to<void>, int, int, cub::_V_300400_SM_750_800_860_900_1000::detail::rle::streaming_context<cuda::std::__4::complex<float> const*, long, long>)
```
on SM75.

bernhardmgruber:
> There are substantial SASS diffs, for example in [...] on SM75.

With the change from #7733, I asked Cursor:

> The last two commits lead to SASS differences on the benchmark cub.bench.run_length_encode.non_trivial_runs.base for SM75. This was likely caused by the rewrite of the tuning information and dispatch layer of DeviceRleDispatch. Please find out why the SASS changed and fix the new tuning and dispatch code so the SASS change is gone. You are not allowed to revert to the old code.

and it indeed found the root cause. I am impressed.

cub/cub/device/dispatch/tuning/tuning_rle_non_trivial_runs.cuh
```diff
@@ -311,6 +327,286 @@ struct policy_hub
...
+      }
+
+      // no tuning for SM90, use a default policy
+      return make_default_policy(BLOCK_LOAD_WARP_TRANSPOSE, length_size, LOAD_DEFAULT);
+    }
+
+    if (arch >= ::cuda::arch_id::sm_86)
+    {
+      // TODO(bgruber): I think we want `LengthT` instead of `int`
+      return make_default_policy(BLOCK_LOAD_DIRECT, sizeof(int), LOAD_LDG);
```
bernhardmgruber, line 524: Retaining comment on old code

NaderAlAwar (approved):

PR #7675, opened 2026-02-13
Split CUB builds into per-launcher jobs.

### LINKED ISSUE #7670: [INFRA]: Address CUB / MSVC CI timeouts
### Is this a duplicate?

### Overview

CI jobs for MSVC + CUB + CTK 13.X timeout regularly when sccache misses are high, due to the wide range of CUDA arch families tested in 13.X and the combined expense of MSVC and nvcc.

## Proposed Fix

Split the cub build jobs, similar to the test jobs. Each targeted test will only depend on the corresponding build:

- `build_nolid` -> `test_nolid`
- `build_lid0` -> `test_lid0`
- `build_lid1` -> `test_lid1`
- `build_lid2` -> `test_lid2`

These are defined in ci/matrix.yaml.

- Each new test job will declare a dependency on the corresponding scoped build job.
- The CUB job map will map `test -> test_{nolid,lid{0,1,2}`.

### CMake

The CUB CMake will introduce new configuration options:

- `CUB_ENABLE_LAUNCH_NONE` (nolid)
- `CUB_ENABLE_LAUNCH_HOST` (lid0)
- `CUB_ENABLE_LAUNCH_DEVICE` (lid1)
- `CUB_ENABLE_LAUNCH_GRAPH` (lid2)

These will all be set to `ON` by default.

The existing `CUB_ENABLE_LAUNCH_VARIANTS` variable will be deprecated (print an appropriate warning if set to `OFF`) and if set to off, the lid1/lid2 options will be disabled. Any usage of this flag in presets, etc, will be replaced with the new options (disabling device + graph).

New presets will be added for the nolid/lidX build variants, and these will be used from the CI scripts. The existing "build everything" preset will still exist, and the specialized presets will only update the relevant `CUB_ENABLE_LAUNCH_*` flags. This is similar to the approach we currently have for splitting CUB test presets.

All test / example executables that are specialized for a launch technique will filtered based on these new variables. If they make no reference to a launch ID (eg. are general purpose / non-device-algorithm-tests that do not test across multiple launchers) they are classified as nolid -- if they are specialized across host / device / graph launch, they will be assigned to the appropriate group. Only the CMake test/example executables that are required for the configured launchers will be added to the build. All others will be skipped/dropped.

alliepiper:
Add nolid / lid_0 / lid_1 / lid_2 toggles:

    CUB_ENABLE_LAUNCH_NO_LAUNCHER
    CUB_ENABLE_LAUNCH_HOST_LAUNCHER
    CUB_ENABLE_LAUNCH_DEVICE_LAUNCHER
    CUB_ENABLE_LAUNCH_GRAPH_LAUNCHER

Updated presets and CI infra accordingly. This should fix #7670.

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -198,6 +198,122 @@
       }
     },
     {
+      "name": "cub-nolid",
+      "displayName": "CUB: No Launcher",
+      "inherits": "cub",
+      "cacheVariables": {
+        "CUB_ENABLE_LAUNCH_NO_LAUNCHER": true,
+        "CUB_ENABLE_LAUNCH_HOST_LAUNCHER": false,
+        "CUB_ENABLE_LAUNCH_DEVICE_LAUNCHER": false,
+        "CUB_ENABLE_LAUNCH_GRAPH_LAUNCHER": false
+      }
+    },
+    {
+      "name": "cub-nolid-cpp17",
+      "displayName": "CUB: No Launcher C++17",
+      "inherits": "cub-nolid",
+      "cacheVariables": {
+        "CMAKE_CXX_STANDARD": "17",
+        "CMAKE_CUDA_STANDARD": "17"
+      }
+    },
+    {
+      "name": "cub-nolid-cpp20",
+      "displayName": "CUB: No Launcher C++20",
+      "inherits": "cub-nolid",
+      "cacheVariables": {
+        "CMAKE_CXX_STANDARD": "20",
+        "CMAKE_CUDA_STANDARD": "20"
+      }
+    },
+    {
+      "name": "cub-lid0",
+      "displayName": "CUB: Host Launch",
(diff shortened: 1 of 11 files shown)


jrhemstad:
@alliepiper can you update the PR description and summarize what the dev-facing consequences of this will be?

alliepiper:
@jrhemstad None, really. The scripts will still do the same thing they did previously, ci just passes in some new options that limit cub builds to a subset of targets that match the existing test split.

Only things devs should notice is that msvc jobs stop timing out in ci 🙂

jrhemstad:
Ack, so the only other observable change is just new `BuildXLaunch` build jobs in CI?

alliepiper:
> Ack, so the only other observable change is just new `BuildXLaunch` build jobs in CI?

Correct.

Issue #7554, opened 2026-02-06
[FEA]: Relax restrictions on cross-system copy support for `cudax::copy_bytes(mdspan)`


gevtushenko:
### Area

CUDA Experimental (cudax)

### Is your feature request related to a problem? Please describe.

`cuda::copy_bytes()` currently supports multi-dimensional data only if they are contiguous (exhaustive) and with identical layout.

This issue can be closed by a PR that makes `cudax::copy_bytes` support any kind of layout in the input mdspans.

### Describe the solution you'd like

Extend `cuda::copy_bytes()` to support any predefined `mdspan` layout, including non-identical source/destination layouts.

The implementation will rely on CuTe to provide the optimal solution. The initial version will be added to cuda experimental.

PR #7676, opened 2026-02-13
Generalized `cudax::copy_bytes` for `mdspan`

fbusato:
closes https://github.com/NVIDIA/cccl/issues/7554 

## Description

`cuda::copy_bytes(mspan)` is currently limited to same layout, contigous mdspan.

This PR extends `copy_bytes` for mdspan-based host/device transfers so it works correctly and efficiently across contiguous and strided layouts, including different rank and mixed layout/order cases.
Common use cases are:  padded tensors, column-major views, subviews, and other `layout_stride` mappings.

**Main features:**

- Optimal (lowest) number of transfers.
- Compatibily with any strided layout, including edge cases, such as transposition. 
- Use `cuMemcpyBatchAsync` on CTK 13+.

The main challenge is use two different stride orderings for different purposes:

- **Stride-sorted order**: finding common contiguous regions / tile size.
- **Original order**: required for correct logical index mapping when source and destination stride permutations differ.

**Details:**

- If source/destination have the same stride order, iterating in stride-sorted order is safe and efficient.
- If stride orders differ, using sorted order for address generation can mismatch source/destination linearization. In that case, iterator addressing switches to original tensor order, while still using the computed tile size from layout analysis. This preserves correctness and still exploits contiguity where available.

### DIFF
--- a/libcudacxx/include/cuda/std/__mdspan/mdspan.h
+++ b/libcudacxx/include/cuda/std/__mdspan/mdspan.h
@@ -441,9 +441,9 @@ public:
   template <size_t... _Idxs>
   [[nodiscard]] _CCCL_API constexpr bool __check_size() const noexcept
   {
-    size_t __prod = 1;
     if constexpr (extents_type::rank() > 0) // MSVC raises a warning even with __r != extents_type::rank()
     {
+      size_t __prod = 1;
       for (size_t __r = 0; __r < extents_type::rank(); ++__r)
       {
         const auto __extent = static_cast<size_t>(mapping().extents().extent(__r));


libcudacxx/include/cuda/__mdspan/traits.h
```diff
@@ -0,0 +1,47 @@
...
+
+#include <cuda/std/__fwd/mdspan.h>
+#include <cuda/std/__type_traits/is_same.h>
+
+#include <cuda/std/__cccl/prologue.h>
+
+_CCCL_BEGIN_NAMESPACE_CUDA
+
+template <typename _LayoutPolicy>
+inline constexpr bool __is_layout_right_v = ::cuda::std::is_same_v<_LayoutPolicy, ::cuda::std::layout_right>;
```
davebayer: These traits should be moved to `<cuda/std/__fwd/mdspan.h>` and be named as `__is_cuda_std_layout_right_v`

cudax/include/cuda/experimental/__copy/mdspan_d2h_h2d.cuh
```diff
@@ -0,0 +1,215 @@
...
+  }
+  else
+  {
+    // __coalesce_right is not strictly needed. It is an optimization to reduce the number of modes.
+    using ::cuda::std::size_t;
+    const auto __src1       = ::cuda::experimental::__to_raw_tensor(__src);
+    const auto __dst1       = ::cuda::experimental::__to_raw_tensor(__dst);
+    const auto __src2       = ::cuda::experimental::__coalesce_right(__src1);
+    const auto __dst2       = ::cuda::experimental::__coalesce_right(__dst1);
+    const auto __src_sorted = ::cuda::experimental::__sort_by_stride_desc(__src1);
```
oleksandr-pavlyk: It appears `__src2` and `__dst2` are unused.

I suspect everywhere from line 134 to the end of this functions `__src1` and `__dst1` should be replaced with `__src2` and `__dst2`.

As is, if `__use_stride_order` is false, we do not use coalesced layouts at all. Suppose input and output were 5 dimensional, but could be coalesced to 2d layouts with different ordering. Presently, the code would execute an equivalent of 5-nested loop, but could be executing an equivalent of double-nested loop.

fbusato: yes, I added `__coalesce_right` at the very end of the testing. I just forgot to propagate them.

cudax/include/cuda/experimental/__copy_bytes/abs_integer.cuh
```diff
@@ -0,0 +1,57 @@
...
+
+namespace cuda::experimental
+{
+//! @brief Returns the absolute value of an integer. Identity for unsigned types.
+//!
+//! @param[in] __value Integer value
+//! @return Absolute value of @p __value
+_CCCL_TEMPLATE(class _Tp)
+_CCCL_REQUIRES(::cuda::std::__cccl_is_integer_v<_Tp>)
+[[nodiscard]] _CCCL_HOST_API constexpr _Tp __abs_integer(_Tp __value) noexcept
```
miscco: There is no reason this should be limited to host, and also this should work fine for nvrtc

cudax/include/cuda/experimental/__copy_bytes/mdspan_d2h_h2d.cuh
```diff
@@ -0,0 +1,237 @@
...
+          typename _AccessorPolicyOut>
+_CCCL_HOST_API void __copy_bytes_impl(
+  ::cuda::std::mdspan<_TpIn, _ExtentsIn, _LayoutPolicyIn, _AccessorPolicyIn> __src,
+  ::cuda::std::mdspan<_TpOut, _ExtentsOut, _LayoutPolicyOut, _AccessorPolicyOut> __dst,
+  [[maybe_unused]] __copy_direction __direction,
+  ::cuda::stream_ref __stream)
+{
+  namespace cudax = ::cuda::experimental;
+  static_assert(::cuda::std::is_same_v<::cuda::std::remove_cv_t<_TpIn>, ::cuda::std::remove_cv_t<_TpOut>>,
+                "TpIn and TpOut must be the same type");
```
miscco: Nitpick: I strongly prefer to add some prefix about where the assertion happened
```suggestion
                "cuda::copy_bytes: TpIn and TpOut must be the same type");
```

cudax/include/cuda/experimental/__copy_bytes/mdspan_d2h_h2d.cuh
```diff
@@ -0,0 +1,237 @@
...
+  using __default_accessor_out = ::cuda::std::default_accessor<_TpOut>;
+  static_assert(::cuda::std::is_convertible_v<_AccessorPolicyIn, __default_accessor_in>,
+                "AccessorPolicyIn must be convertible to cuda::std::default_accessor");
+  static_assert(::cuda::std::is_convertible_v<_AccessorPolicyOut, __default_accessor_out>,
+                "AccessorPolicyOut must be convertible to cuda::std::default_accessor");
+  if (__src.size() != __dst.size())
+  {
+    _CCCL_THROW(::std::invalid_argument, "mdspans must have the same size");
+  }
+  const auto __tensor_size = __src.size();
```
miscco, line 93: Please add some spacing, its hard to read as a block of text
```suggestion

  const auto __tensor_size = __src.size();
```

cudax/include/cuda/experimental/__copy_bytes/mdspan_d2h_h2d.cuh
```diff
@@ -0,0 +1,237 @@
...
+  }
+  if (!::cuda::std::is_sufficiently_aligned<alignof(_TpOut)>(__dst.data_handle()))
+  {
+    _CCCL_THROW(::std::invalid_argument, "destination mdspan must be sufficiently aligned");
+  }
+  if (cudax::__has_interleaved_stride_order(__dst))
+  {
+    _CCCL_THROW(::std::invalid_argument, "destination mdspan must not have interleaved stride order");
+  }
+  if (__tensor_size == 1)
```
miscco: This is the first check after precondition checks, please add a spacing
```suggestion

  if (__tensor_size == 1)
```

cudax/include/cuda/experimental/__copy_bytes/mdspan_d2h_h2d.cuh
```diff
@@ -0,0 +1,237 @@
...
+  {
+    ::cuda::__driver::__memcpyAsync(__dst.data_handle(), __src.data_handle(), sizeof(_TpIn), __stream.get());
+    return;
+  }
+  if constexpr (_ExtentsIn::rank() > 0 && _ExtentsOut::rank() > 0)
```
miscco, line 121: Critical: Is that really correct for all cases? AFAIK a rank 0 mdspan still has a single item it points to. 

Dont we need to check for `required_span_size()`?

cudax/include/cuda/experimental/__copy_bytes/mdspan_to_raw_tensor.cuh
```diff
@@ -0,0 +1,129 @@
...
+_CCCL_HOST_API constexpr __raw_tensor<_ExtentT, _StrideT, _Tp, _MaxRank>
+__to_raw_tensor(const ::cuda::std::mdspan<_Tp, _Extents, _LayoutPolicy, _AccessorPolicy>& __mdspan,
+                ::cuda::std::bool_constant<_RemoveExtent1> = {}) noexcept
+{
+  static_assert(_MaxRank >= _Extents::rank(), "_MaxRank must be at least _Extents::rank()");
+  using __raw_tensor_t = __raw_tensor<_ExtentT, _StrideT, _Tp, _MaxRank>;
+  using __extent_t     = typename __raw_tensor_t::__unsigned_extent_t;
+  using __rank_t       = typename _Extents::rank_type;
+  __raw_tensor_t __result{__mdspan.data_handle(), 0, {}, {}};
+  if constexpr (_Extents::rank() > 0)
```
miscco, line 82: Ditto: A zero rank mdspan can still point to a single item

cudax/include/cuda/experimental/__copy_bytes/mdspan_to_raw_tensor.cuh
```diff
@@ -0,0 +1,129 @@
...
+  {
+    ::cuda::std::size_t __r = 0;
+    for (::cuda::std::size_t __i = 0; __i < _Extents::rank(); ++__i)
```
miscco: Nitpick: This should use `rank_type` otherwise we might get conversion errors
```suggestion
    for (__rank_t __i = 0; __i != _Extents::rank(); ++__i)
```

cudax/include/cuda/experimental/__copy_bytes/mdspan_d2h_h2d.cuh
```diff
@@ -0,0 +1,237 @@
...
+    using __stride_t = ::cuda::std::common_type_t<cudax::__mdspan_stride_t<_ExtentsIn, _LayoutPolicyIn>,
+                                                  cudax::__mdspan_stride_t<_ExtentsOut, _LayoutPolicyOut>>;
+    constexpr auto __max_rank = ::cuda::std::max(_ExtentsIn::rank(), _ExtentsOut::rank());
+    const auto __src_raw      = cudax::__to_raw_tensor<__extent_t, __stride_t, __max_rank>(__src, __remove_extent1);
+    const auto __dst_raw      = cudax::__to_raw_tensor<__extent_t, __stride_t, __max_rank>(__dst, __remove_extent1);
+    if (!cudax::__same_extents(__src_raw, __dst_raw))
+    {
+      _CCCL_THROW(::std::invalid_argument, "mdspans must have the same extents (after removing singleton dimensions)");
+    }
+    auto __src_simplified = __src_raw;
```
miscco, line 135: Please also add some spacing here and when applicable

cudax/include/cuda/experimental/__copy_bytes/mdspan_to_raw_tensor.cuh
```diff
@@ -0,0 +1,129 @@
...
+//! @param[in] (tag) If true, remove extent-1 modes
+//! @return @ref __raw_tensor descriptor with data pointer, rank, extents, and strides
+template <typename _ExtentT,
+          typename _StrideT,
+          ::cuda::std::size_t _MaxRank,
+          typename _Tp,
+          typename _Extents,
+          typename _LayoutPolicy,
+          typename _AccessorPolicy,
+          bool _RemoveExtent1 = false>
```
miscco: Nitpick: Not a fan of mixing numbers and names, could we use `_RemoveExtentOne`

cudax/include/cuda/experimental/__copy_bytes/mdspan_to_raw_tensor.cuh
```diff
@@ -0,0 +1,129 @@
...
+//! Extent-1 modes are optionally removed when @p _RemoveExtent1 is true.
+//!
+//! @param[in] __mdspan Source mdspan view
+//! @param[in] (tag) If true, remove extent-1 modes
+//! @return @ref __raw_tensor descriptor with data pointer, rank, extents, and strides
+template <typename _Tp, typename _Extents, typename _LayoutPolicy, typename _AccessorPolicy, bool _RemoveExtent1 = false>
+[[nodiscard]]
+_CCCL_HOST_API constexpr __to_raw_tensor_t<_Extents::rank(), _Tp, _Extents, _LayoutPolicy>
+__to_raw_tensor(const ::cuda::std::mdspan<_Tp, _Extents, _LayoutPolicy, _AccessorPolicy>& __mdspan,
+                ::cuda::std::bool_constant<_RemoveExtent1> __tag = {}) noexcept
```
miscco: Question: Why are  we using tag dispatch here? we are defining the template arguments below anyway

cudax/include/cuda/experimental/__copy_bytes/memcpy_batch_tiles.cuh
```diff
@@ -0,0 +1,195 @@
...
+      , __contiguous_size_{__contiguous_size}
+  {
+    // Precomputes exclusive prefix products of extents so that each `operator()` call decomposes a flat index into
+    // multi-dimensional coordinates and computes the corresponding byte offset.
+    ::cuda::std::exclusive_scan(
+      __tensor.__extents.data(),
+      __tensor.__extents.data() + __tensor.__rank,
+      __extent_products_.data(),
+      __unsigned_extent_t{1},
+      ::cuda::std::multiplies<>{});
```
miscco, line 66: Question: Should this use a types `multiplies` to avoid integer promotion issues?

cudax/include/cuda/experimental/__copy_bytes/memcpy_batch_tiles.cuh
```diff
@@ -0,0 +1,195 @@
...
+      &__attribute_indices,
+      1, // __num_attributes
+      __stream.get());
+  };
+  constexpr size_t __max_tiles = 16;
+  if (__num_tiles > __max_tiles)
+  {
+    ::std::vector<const void*> __src_ptr_vector(__num_tiles);
+    ::std::vector<void*> __dst_ptr_vector(__num_tiles);
+    ::std::vector<size_t> __sizes(__num_tiles);
```
miscco: Nitpick: This is a heavy use for `std::vector` consider using `::cuda::std::get_temporary_buffer<void*>`

cudax/include/cuda/experimental/__copy_bytes/simplify_paired.cuh
```diff
@@ -0,0 +1,165 @@
...
+//! This helps to get a single logic for __tile_iterator_linearized
+//!
+//! @param[in] __input Raw tensor whose modes are reversed
+//! @return New raw tensor with extents and strides in reversed mode order
+template <typename _ExtentT, typename _StrideT, typename _Tp, ::cuda::std::size_t _MaxRank>
+[[nodiscard]] _CCCL_HOST_API __raw_tensor<_ExtentT, _StrideT, _Tp, _MaxRank>
+__reverse_modes(const __raw_tensor<_ExtentT, _StrideT, _Tp, _MaxRank>& __input) noexcept
+{
+  __raw_tensor<_ExtentT, _StrideT, _Tp, _MaxRank> __result{__input.__data, __input.__rank, {}, {}};
+  for (::cuda::std::size_t __i = 0; __i < __input.__rank; ++__i)
```
miscco: Ditto: Should use `rank_type`

cudax/include/cuda/experimental/__copy_bytes/simplify_paired.cuh
```diff
@@ -0,0 +1,165 @@
...
+                                            __raw_tensor<_ExtentT, _StrideT, _TpDst, _MaxRank>& __dst) noexcept
+{
+  namespace cudax = ::cuda::experimental;
+  using ::cuda::std::size_t;
+  using __unsigned_extent_t = ::cuda::std::make_unsigned_t<_ExtentT>;
+  using __mode_t            = ::cuda::std::tuple<__unsigned_extent_t, _StrideT, _StrideT>;
+  _CCCL_ASSERT(cudax::__same_extents(__src, __dst), "Source and destination tensors must have the same extents");
+  ::cuda::std::array<__mode_t, _MaxRank> __modes{};
+  const auto __rank = __src.__rank;
+  for (size_t __i = 0; __i < __rank; ++__i)
```
miscco: Ditto: rank_type

cudax/include/cuda/experimental/__copy_bytes/tensor_query.cuh
```diff
@@ -0,0 +1,190 @@
...
+//! check and may reject some layouts with distinct offsets.
+//!
+//! @param[in] __mdspan Mdspan view to inspect
+//! @return true if the layout has interleaved strides
+template <typename _Tp, typename _Extents, typename _LayoutPolicy, typename _AccessorPolicy>
+[[nodiscard]] _CCCL_HOST_API constexpr bool __has_interleaved_stride_order(
+  const ::cuda::std::mdspan<_Tp, _Extents, _LayoutPolicy, _AccessorPolicy>& __mdspan) noexcept
+{
+  if constexpr (_Extents::rank() > 0)
+  {
```
miscco, line 168: Please switch the conditions around to use an early return
```suggestion
  if constexpr (_Extents::rank() == 0)
  {
    return false;
  }
```

libcudacxx/include/cuda/__driver/driver_api.h
```diff
@@ -549,6 +574,18 @@ __pointerGetAttributeNoThrow(__pointer_attribute_value_type_t<_Attr>& __result,
...
+template <::CUpointer_attribute _Attr>
+[[nodiscard]] _CCCL_HOST_API __pointer_attribute_value_type_t<_Attr> __pointerGetAttribute(const void* __ptr)
+{
+  __pointer_attribute_value_type_t<_Attr> __result;
+  const auto __status = ::cuda::__driver::__pointerGetAttributeNoThrow<_Attr>(__result, __ptr);
+  if (__status != ::cudaSuccess)
+  {
+    _CCCL_THROW(::cuda::cuda_error, __status, "Failed to get attribute of a pointer");
+  }
+  return __result;
```
miscco: Important: Shouldnt this use `_CCCL_TRY_CUDA_API`

libcudacxx/include/cuda/__mdspan/traits.h
```diff
@@ -0,0 +1,47 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of the libcu++ Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef _CUDA___MDSPAN_TRAITS_H
```
miscco, line 10: I believe that file belongs in `cuda/std/__mdspan`

cudax/include/cuda/experimental/__copy_bytes/mdspan_d2h_h2d.cuh
oleksandr-pavlyk: If rank is zero, then `__tensor_size == 1` condition should have been entered. Perhaps in the `if constexpr` branch we should assert `__tensor_size > 1` to make it explicit.

cudax/include/cuda/experimental/__copy_bytes/tensor_query.cuh
```diff
@@ -0,0 +1,190 @@
...
+    namespace cudax       = ::cuda::experimental;
+    const auto __tensor   = cudax::__to_raw_tensor(__mdspan, __remove_extent1);
+    const auto __sorted   = cudax::__sort_by_stride(__tensor);
+    using __stride_t      = decltype(__sorted.__strides[0]);
+    const auto& __extents = __sorted.__extents;
+    const auto& __strides = __sorted.__strides;
+    for (::cuda::std::size_t __i = 0; __i < __sorted.__rank; ++__i)
```
oleksandr-pavlyk:
```suggestion
    using __rank_t       = typename _Extents::rank_type;
    for (__rank_t __i = 0; __i < __sorted.__rank; ++__i)
```

cudax/include/cuda/experimental/__copy_bytes/tensor_query.cuh
```diff
@@ -0,0 +1,190 @@
...
+    {
+      if (__extents[__i] > 1 && __strides[__i] == 0)
+      {
+        return true;
+      }
+    }
+    for (::cuda::std::size_t __i = 0; __i + 1 < __sorted.__rank; ++__i)
```
oleksandr-pavlyk:
```suggestion
    for (__rank_t __i = 0; __i + 1 < __sorted.__rank; ++__i)
```

cudax/include/cuda/experimental/__copy_bytes/tensor_query.cuh
```diff
@@ -0,0 +1,190 @@
...
+    {
+      const auto __extent = static_cast<__stride_t>(__extents[__i]);
+      if (__extent * cudax::__abs_integer(__strides[__i]) > cudax::__abs_integer(__strides[__i + 1]))
```
oleksandr-pavlyk, line 187: In NumPy it is possible to create 2D array which is interleaved, but where this condition is not satisfied.

```
In [1]: import numpy as np

In [2]: v = np.arange(6)

In [3]: swv = np.lib.stride_tricks.sliding_window_view(v, 3)

In [4]: swv.shape, swv.strides
Out[4]: ((4, 3), (8, 8))

In [5]: swv
Out[5]: 
array([[0, 1, 2],
       [1, 2, 3],
       [2, 3, 4],
       [3, 4, 5]])
``` 

For this reason I would replace this condition with

```suggestion
      if (__extent * cudax::__abs_integer(__strides[__i]) != cudax::__abs_integer(__strides[__i + 1]))
```

cudax/include/cuda/experimental/__copy_bytes/tensor_query.cuh
```diff
@@ -0,0 +1,190 @@
...
+          ::cuda::std::size_t _MaxRankOut>
+[[nodiscard]] _CCCL_HOST_API constexpr bool
+__same_extents(const __raw_tensor<_ExtentTIn, _StrideTIn, _TpIn, _MaxRankIn>& __tensor_in,
+               const __raw_tensor<_ExtentTOut, _StrideTOut, _TpOut, _MaxRankOut>& __tensor_out) noexcept
+{
+  if (__tensor_in.__rank != __tensor_out.__rank)
+  {
+    return false;
+  }
+  for (::cuda::std::size_t __i = 0; __i < __tensor_in.__rank; ++__i)
```
oleksandr-pavlyk: Perhaps use common rank type for both extents.

cudax/include/cuda/experimental/__copy_bytes/tensor_query.cuh
```diff
@@ -0,0 +1,190 @@
...
+//! @return true if strides[0..rank) are non-descending
+template <typename _Ep, typename _Sp, typename _Tp, ::cuda::std::size_t _MaxRank>
+[[nodiscard]] _CCCL_HOST_API bool __has_sorted_strides(const __raw_tensor<_Ep, _Sp, _Tp, _MaxRank>& __tensor) noexcept
+{
+  namespace cudax = ::cuda::experimental;
+  _CCCL_ASSERT(::cuda::in_range(__tensor.__rank, ::cuda::std::size_t{0}, _MaxRank), "Invalid tensor rank");
+  return ::cuda::std::is_sorted(
+    __tensor.__strides.cbegin(), __tensor.__strides.cbegin() + __tensor.__rank, [](auto __a, auto __b) {
+      return cudax::__abs_integer(__a) < cudax::__abs_integer(__b);
+    });
```
oleksandr-pavlyk, line 109: The implementation check for strictly ascending ordering. Either update the comment, or change the implementation to use `cudax::__abs_integer(__a) <= cudax::__abs_integer(__b);`

cudax/include/cuda/experimental/__copy_bytes/tensor_query.cuh
```diff
@@ -0,0 +1,190 @@
...
+  using __unsigned_extent_t = typename __raw_tensor<_ExtentT, _StrideT, _Tp, _MaxRank>::__unsigned_extent_t;
+  using __mode_t            = ::cuda::std::tuple<__unsigned_extent_t, _StrideT>;
+  const auto __rank         = __tensor.__rank;
+  ::cuda::std::array<__mode_t, _MaxRank> __modes{};
+  for (size_t __i = 0; __i < __rank; ++__i)
+  {
+    __modes[__i] = {__tensor.__extents[__i], __tensor.__strides[__i]};
+  }
+  ::cuda::std::stable_sort(__modes.begin(), __modes.begin() + __rank, [](const __mode_t& __lhs, const __mode_t& __rhs) {
+    return cudax::__abs_integer(::cuda::std::get<1>(__lhs)) < cudax::__abs_integer(::cuda::std::get<1>(__rhs));
```
oleksandr-pavlyk: The function does not impose total order on the modes, as it leaves unspecified what happens when absolute values of strides are equal. I guess because the sorting is stable, those entries remain in the same relative order as in the input. 

I assume this is OK, but wanted to call it out.

cudax/include/cuda/experimental/__copy_bytes/tensor_query.cuh
```diff
@@ -0,0 +1,190 @@
...
+[[nodiscard]] _CCCL_HOST_API constexpr __raw_tensor<_ExtentT, _StrideT, _Tp, _MaxRank>
+__sort_by_stride(const __raw_tensor<_ExtentT, _StrideT, _Tp, _MaxRank>& __tensor) noexcept
+{
+  namespace cudax = ::cuda::experimental;
+  using ::cuda::std::size_t;
+  using __unsigned_extent_t = typename __raw_tensor<_ExtentT, _StrideT, _Tp, _MaxRank>::__unsigned_extent_t;
+  using __mode_t            = ::cuda::std::tuple<__unsigned_extent_t, _StrideT>;
+  const auto __rank         = __tensor.__rank;
+  ::cuda::std::array<__mode_t, _MaxRank> __modes{};
+  for (size_t __i = 0; __i < __rank; ++__i)
```
oleksandr-pavlyk:
```suggestion
  for (decltype(__rank) __i = 0; __i < __rank; ++__i)
```

cudax/include/cuda/experimental/__copy_bytes/types.cuh
```diff
@@ -0,0 +1,57 @@
...
+namespace cuda::experimental
+{
+//! @brief Raw tensor descriptor with dynamic rank, extents, and strides.
+template <typename _ExtentT, typename _StrideT, typename _Tp, ::cuda::std::size_t _MaxRank>
+struct __raw_tensor
+{
+  using __unsigned_extent_t = ::cuda::std::make_unsigned_t<_ExtentT>;
+
+  _Tp* __data;
+  ::cuda::std::size_t __rank;
```
oleksandr-pavlyk: Would it ever make sense to make rank `uint32_t` instead of size_t?  Ranks are usually small.

libcudacxx/include/cuda/__driver/driver_api.h
```diff
@@ -341,6 +341,31 @@ _CCCL_HOST_API inline void __memcpyAsyncWithAttributes(
...
+
+_CCCL_HOST_API inline void __memcpyBatchAsync(
+  void** __dsts,
+  const void** __srcs,
+  const ::cuda::std::size_t* __sizes,
+  ::cuda::std::size_t __count,
+  ::CUmemcpyAttributes* __attributes,
+  ::cuda::std::size_t* __attribute_indices,
+  ::cuda::std::size_t __num_attributes,
+  ::CUstream __stream)
```
oleksandr-pavlyk, line 354: `cuMemcpyBatchAsync` documentation has a note that stream can not be legacy NULL stream: https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__MEM.html#group__CUDA__MEM_1g6f1ff58e3065df3eb4b573dba77ad31f

Should this be asserted before making the call?

cudax/include/cuda/experimental/__copy_bytes/simplify_paired.cuh
```diff
@@ -0,0 +1,165 @@
...
+  {
+    __modes[__i] = {__src.__extents[__i], __src.__strides[__i], __dst.__strides[__i]};
+  }
+  ::cuda::std::stable_sort(__modes.begin(), __modes.begin() + __rank, [](const __mode_t& __lhs, const __mode_t& __rhs) {
+    return cudax::__abs_integer(::cuda::std::get<2>(__lhs)) < cudax::__abs_integer(::cuda::std::get<2>(__rhs));
```
oleksandr-pavlyk: Again, the predicate does not impose total order.  If absolute values of destination strides are equal, should we further sort on absolute value of source strides?

cudax/include/cuda/experimental/__copy_bytes/memcpy_batch_tiles.cuh
```diff
@@ -0,0 +1,195 @@
...
+  [[nodiscard]] _CCCL_HOST_API _Tp* operator()(__unsigned_extent_t __tile_idx) const noexcept
+  {
+    const auto __index = __tile_idx * __contiguous_size_;
+    _StrideT __offset  = 0;
+    for (::cuda::std::size_t __i = 0; __i < __tensor_.__rank; ++__i)
+    {
+      const auto __coord = static_cast<_StrideT>((__index / __extent_products_[__i]) % __tensor_.__extents[__i]);
+      __offset += __coord * __tensor_.__strides[__i];
+    }
+    return __tensor_.__data + __offset;
```
oleksandr-pavlyk: This code performs integral division and modular operation even for 1D arrays, where `coord = static_cast<_StrideT>(__i);` could be used. Can this be avoided?

cudax/include/cuda/experimental/__copy_bytes/mdspan_to_raw_tensor.cuh
fbusato: already addressed by `__tensor_size == 1`

cudax/include/cuda/experimental/__copy_bytes/tensor_query.cuh
fbusato: `rank() > 0` is important to avoid compile-time errors. Secondly, even without compile errors, it saves compile time

libcudacxx/include/cuda/__driver/driver_api.h
fbusato: not sure, @pciolkosz  what is the convention in this case?

cudax/include/cuda/experimental/__copy_bytes/types.cuh
fbusato: the overhead on the host side is negligible and we would need to introduce explicit conversions.

cudax/include/cuda/experimental/__copy_bytes/memcpy_batch_tiles.cuh
fbusato: not really. All values are of the same type

cudax/include/cuda/experimental/__copy_bytes/memcpy_batch_tiles.cuh
fbusato: wait, `std::get_temporary_buffer` has been deprecated in C++20

cudax/include/cuda/experimental/__copy_bytes/tensor_query.cuh
oleksandr-pavlyk: I tried adding a test to cover such case.

<details>
<summary>Test source code</summary>

```cpp
TEST_CASE("copy_bytes sliding window view", "[copy_bytes][window]")
{
  constexpr size_t n = 6;
  constexpr size_t s = 2*n - 1;
  thrust::host_vector<int> vec(s, thrust::no_init);
  for(size_t i = 0; i < n; ++i) {
    vec[n - 1 + i] = i;
    if (i > 0) {
      vec[n - 1 - i] = i;
    }
  }

  thrust::host_vector<int> expected(n*n, thrust::no_init);
  for(size_t i = 0; i < n; ++i) {
    for(size_t j = 0; j < n; ++j) {
      expected[i * n + j] = static_cast<int>(std::max(i, j));
    }
  }

  using src_extents_t = cuda::std::extents<int, n, n>;
  using dst_extents_t = cuda::std::extents<int, n, n>;
  using src_mapping_t = cuda::std::layout_stride::mapping<src_extents_t>;
  using dst_mapping_t = cuda::std::layout_right::mapping<dst_extents_t>;

  cuda::std::array<int, 2> src_strides = {-1, 1};

  src_extents_t src_extents{};
  dst_extents_t dst_extents{};

  using T = int;
  thrust::device_vector<T> device_data(expected.size(), 0);
  {
    // host to device
    using host_mdspan_t   = cuda::host_mdspan<const T, src_extents_t, cuda::std::layout_stride>;
    using device_mdspan_t = cuda::device_mdspan<T, dst_extents_t, cuda::std::layout_right>;

    host_mdspan_t host_md(vec.data() + n - 1, src_mapping_t(src_extents, src_strides));

    device_mdspan_t device_md(thrust::raw_pointer_cast(device_data.data()), dst_mapping_t(dst_extents));

    cuda::experimental::copy_bytes(host_md, device_md, stream);
    stream.sync();
    CUDAX_REQUIRE(thrust::host_vector<T>(device_data) == expected);
  }
}
```
</details>

The test compiles when added at the bottom of `"cudax/test/copy_bytes/mdspan_d2h_h2_d.cu"`, but trips an assertion at run-time: 

```
cudax.test.copy: /home/opavlyk/repos/cccl_clean/lib/cmake/libcudacxx/../../../libcudacxx/include/cuda/std/__mdspan/layout_stride.h:308: mapping: Assertion `layout_stride::mapping ctor: the provided extents and strides lead to a non-unique mapping' failed.
```

Merging #7175 and replacing `cuda::std::layout_stride` with `cuda::layout_stride_relaxed` results in static assertion:

```
error: static assertion failed with "LayoutPolicyIn must be a predefined layout policy"
```

cudax/include/cuda/experimental/__copy_bytes/memcpy_batch_tiles.cuh
fbusato: I simply used `new/delete`

cudax/include/cuda/experimental/__copy_bytes/mdspan_to_raw_tensor.cuh
fbusato: remove entirely

cudax/include/cuda/experimental/__copy_bytes/mdspan_to_raw_tensor.cuh
fbusato: removed

cudax/include/cuda/experimental/__copy_bytes/memcpy_batch_tiles.cuh
fbusato: yes, also the first iteration can be improved

cudax/include/cuda/experimental/__copy_bytes/tensor_query.cuh
fbusato: that's a good point. I used stable sort for this reason.

cudax/include/cuda/experimental/__copy_bytes/tensor_query.cuh
fbusato: updated the comment

cudax/include/cuda/experimental/__copy_bytes/tensor_query.cuh
fbusato: updated all ranks type

libcudacxx/include/cuda/__mdspan/traits.h
fbusato: the issue is that I want to extend it to `layout_stride_relaxed`

cudax/include/cuda/experimental/__copy_bytes/simplify_paired.cuh
fbusato: that's a good idea. Sorted strides of source tensors could improve performance

cudax/include/cuda/experimental/__copy_bytes/tensor_query.cuh
fbusato: I didn't know that this is possible in NumPy. The new condition makes sense. I will add the test case.

Related to `layout_stride.h:308: mapping: Assertion `layout_stride::mapping ctor` this is a limitation of `layout_stride` in C++ standard. `layout_stride_relaxed` needs to the merged before testing it

fbusato: @oleksandr-pavlyk do you think it makes sense to enable interleaved layouts for destination tensors? this implies overlapping

miscco (approved):

cudax/include/cuda/experimental/__copy_bytes/tensor_query.cuh
oleksandr-pavlyk: I meant the source `mdspan` needs to support overlapping mappings, but the destination mapping must be unique to avoid race condition.

I did merge `layout_stride_relaxed` branch prior to testing, and the static assertion was tripped as reported.

fbusato: uhmm, weird. The check is only on the destination. Anyway, I will open a new PR for `layout_stride_relaxed` + your test

libcudacxx/include/cuda/__driver/driver_api.h
pciolkosz, line 331: Should we route this through `__memcpyBatchAsync`?

libcudacxx/include/cuda/__driver/driver_api.h
pciolkosz: I think we should pass the stream as is to the driver and then only if the call fails, we can check if this was NULL stream and provide better diagnostics.
This way if it end up being supported in the future we won't fail for no reason, also for performance reasons I think we should prefer just passing stuff through and then only diagnose when driver errors out, otherwise both we and the driver will perform exactly the same argument validation.
You would need to change this function to a noThrow variant

libcudacxx/include/cuda/__driver/driver_api.h
fbusato: yes, it is a good idea but for another PR

PR #7681, opened 2026-02-16
Fix max atomic size for NVHPC

### LINKED ISSUE #7650: [BUG]: Static assertion failure in atomic_ref for double scan on sm_120 (<=16 vs <16 size check mismatch)
### Is this a duplicate?

### Type of Bug

Compile-time Error

### Component

CUB

### Describe the bug

This test fails when targeting sm_120 architecture:
```
#include <execution>
#include <numeric>
#include <vector>

int32_t
main()
{
  std::vector<double> in(1000);
  std::vector<double> out(1000);
  auto orr = std::inclusive_scan(std::execution::par, in.begin(), in.end(), out.begin());
  return 0;
}
```
it fails with:
```
cuda/std/__atomic/functions/host.h", line 85: error: static assertion failed with "atomic_ref<T> where sizeof(T) > 8 is not supported on this system."
    _LIBCUDACXX_INT128_WARN(_Tp)
    ^
cuda/std/__atomic/functions/host.h", line 85: note: the final comparison was 16 < 16
    _LIBCUDACXX_INT128_WARN(_Tp)
    ^
```
On sm_120, we use `Policy1000`, which includes the `WarpspeedPolicy` structure. For type double, `scan_use_warpspeed` evaluates to true, leading to the instantiation of `atomic_ref<tile_state_t<double>>`. 
The function `storeTileAggregate` in  `cub/detail/warpspeed/look_ahead.cuh` contains:
```
  if constexpr (sizeof(tile_state_t<AccumT>) <= 16
```
Meanwhile, cuda/std/__atomic/functions/host.h enforces:
```
static_assert(sizeof(_Tp) < 16, "atomic_ref<T> where sizeof(T) > 8 is not supported on this system.");
```
Because `storeTileAggregate` allows `sizeof(tile_state_t<AccumT>) == 16`, it takes the `atomic_ref` path.
However, `atomic_ref` later asserts against `sizeof(T) < 16`, causing the build to fail.
This results in an inconsistent condition between these two checks (<= 16 vs < 16), leading to a static assertion failure when building for sm_120.

### How to Reproduce

```
nvc++ -stdpar --c++17  test_scan.cpp
```

### Expected behavior

Compilation should succeed, or the size conditions should be made consistent to prevent this mismatch.

### Reproduction link

_No response_

### Operating System

_No response_

### nvidia-smi output

_No response_

### NVCC version

_No response_

miscco:
We were hardcoding 16 bytes as the hard limit for the lookahead algorithm.

However, that is not correct for NVHPC which only supports 4 or 8 byte atomics

Also be safe and conditionally switch based on SM90 because that is required for 16 byte atomics

Fixes #7650

### DIFF
--- a/cub/cub/detail/warpspeed/look_ahead.cuh
+++ b/cub/cub/detail/warpspeed/look_ahead.cuh
@@ -30,10 +30,21 @@
 #  include <cuda/atomic>
 #endif // !defined(_CUDACC_DEVICE_ATOMIC_BUILTINS__) || _CCCL_COMPILER(MSVC)
 
+#include <nv/target>
+
 CUB_NAMESPACE_BEGIN
 
 namespace detail::warpspeed
 {
+[[nodiscard]] _CCCL_API _CCCL_CONSTEVAL size_t max_native_atomic_size() noexcept
+{
+#if _CCCL_CUDA_COMPILER(NVHPC)
+  return 8;
+#else // ^^^ _CCCL_CUDA_COMPILER(NVHPC) ^^^ / vvv !_CCCL_CUDA_COMPILER(NVHPC)  vvv
+  NV_IF_ELSE_TARGET(NV_PROVIDES_SM_90, (return 16;), (return 8;))
+#endif // !_CCCL_CUDA_COMPILER(NVHPC)
+}
+
 enum scan_state : ::cuda::std::uint32_t
 {
   empty          = 0,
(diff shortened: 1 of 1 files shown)


cub/cub/detail/warpspeed/look_ahead.cuh
```diff
@@ -30,10 +30,21 @@
 #  include <cuda/atomic>
 #endif // !defined(_CUDACC_DEVICE_ATOMIC_BUILTINS__) || _CCCL_COMPILER(MSVC)
 
+#include <nv/target>
+
 CUB_NAMESPACE_BEGIN
 
 namespace detail::warpspeed
 {
+[[nodiscard]] _CCCL_API _CCCL_CONSTEVAL size_t max_native_atomic_size() noexcept
```
davebayer:
```suggestion
[[nodiscard]] _CCCL_API _CCCL_CONSTEVAL ::cuda::std::size_t max_native_atomic_size() noexcept
```

davebayer (approved):

PR #7683, opened 2026-02-16
Try to better suppress deprecation warnings for NVHPC

miscco:


davebayer (approved):

cub/test/catch2_test_iterator.cu
```diff
@@ -21,7 +24,6 @@
 
 // We start suppressing deprecation warnings but do not stop at the end of the file. This suppresses warnings in the
 // compiler-generated `catch2_test_iterator.compute_120.cudafe1.stub.c`
```
davebayer: Shouldn't the suppresion stay where it was? You should move the comment, too

miscco: I just found that NVHPC still emits the deprecation warning, which could be the case because the warning happens deep inside CCCL, so the hope is that putting the deprecation suppression in front of any include that saw texture iterators it will work

PR #7686, opened 2026-02-16
Literal for `cuda::compute_capability`

davebayer:
I'd like to add a literal for `cuda::compute_capability`, because right now, if you want to compare compute capabilities, you need to do:
```cpp
if (cc >= cuda::compute_capability{90})
{
  // ...
}
```
which is very verbose. After this change, you can just do:
```cpp
using namespace cuda::compute_capability_literals;
if (cc >= 9.0_cc)
{
  // ...
}
```

There is just one question - should this be a user defined literals (`9.0_cc`) or do we want to make it a standard literal (`9.0cc`)? I'd prefer to get rid of the `_`, but I'm not sure if compilers would like it.

### DIFF
--- a/libcudacxx/include/cuda/__device/compute_capability.h
+++ b/libcudacxx/include/cuda/__device/compute_capability.h
@@ -22,6 +22,7 @@
 #endif // no system header
 
 #include <cuda/__fwd/devices.h>
+#include <cuda/std/__charconv/from_chars.h>
 #include <cuda/std/__fwd/format.h>
 #include <cuda/std/__utility/cmp.h>
 #include <cuda/std/__utility/to_underlying.h>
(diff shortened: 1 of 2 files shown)


jrhemstad:
While the benefits are obvious, I'm not sure they outweigh the risks. It seems like there's a pretty high chance of collision with user code that may have already defined a `_cc` literal?

davebayer:
> It seems like there's a pretty high chance of collision with user code that may have already defined a `_cc` literal?

That should be completely fine. The literal is hidden in `cuda::literals::compute_capability_literals` namespace, so you need to do:
```cpp
using namespace cuda::literals; // or
using namespace cuda::compute_capability_literals; // or
using namespace cuda::literals::compute_capability_literals;
```
to make the literals available in the code. This is exactly the same approach as what standard literals use.

This is definitely not source breaking for the user, but of course there might be collisions, if the user tries to use his and our literals in the same scope :)

davebayer:
After internal discussion, we decided that we don't want the literal

Issue #7249, opened 2026-01-15
[GH] Consolidate issue/pr type fields


alliepiper:
We currently have three ways to specify an issue type:

The organization-wide "Type" ([GH docs](https://github.com/orgs/community/discussions/139933), says it's opt-in, maybe we can opt-out?

<img width="552" height="704" alt="Image" src="https://github.com/user-attachments/assets/aef270e4-a847-4f37-9616-7c8649fb32c7" />

Our project's Type fields:

<img width="489" height="662" alt="Image" src="https://github.com/user-attachments/assets/d9e6dc1b-2bdd-4c18-aa19-f642f1b550d0" />

Our labels (ignore infra):

<img width="270" height="692" alt="Image" src="https://github.com/user-attachments/assets/d642e309-2ccd-4ae1-912c-4a4b43c84e2b" />

We should consolidate this, but first review the project_automation workflows to make sure we won't break anything there.

jrhemstad:
We should just use the native "Issue Type".

The only reason our project-specific Type field exists is because it predates the existence of the standard "Issue Type".

PR #7690, opened 2026-02-16
Update issue templates to use native issue types

jrhemstad:
closes: https://github.com/NVIDIA/cccl/issues/7249

Issue types didn't originally exist when we first created our GitHub project.

Now that they exist, use them as the source of truth and remove the old Project automation that would set the Project type field based on the label/title. 

As a result, I was able to just delete the `project_automation_on_issue_open.yml` workflow 
- [x] Update issue templates where appropriate to set issue type directly
- [x] Issue templates no longer set labels as they are no longer needed in favor of the native issue type
- [x] Delete `project_automation_on_issue_open.yml` workflow as it is no longer necessary
- [x] Delete "type" project field
- [x] Update issue template to add the issue to the GH Project directly using the `project` field (more deterministic than waiting on automation)
- [x] Update title of bug/fea/doc issue templates so they are always ordered first on the Issue Chooser.

### DIFF
--- a/.github/ISSUE_TEMPLATE/devex_ticket.yml
+++ b/.github/ISSUE_TEMPLATE/devex_ticket.yml
@@ -1,6 +1,7 @@
 name: DevEx Request
 description: Suggest improvements to GitHub project infrastructure and workflows.
 title: '[DEVEX]: '
+labels: ['devex']
 projects: ['NVIDIA/6']
 body:
   - type: markdown
--- a/.github/ISSUE_TEMPLATE/infra_ticket.yml
+++ b/.github/ISSUE_TEMPLATE/infra_ticket.yml
@@ -1,6 +1,7 @@
 name: Infra Ticket
 description: Request updates, report issues, etc. for CMake / GHA / CI / devcontainers.
 title: '[INFRA]: '
+labels: ['infra']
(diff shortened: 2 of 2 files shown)


.github/ISSUE_TEMPLATE/devex_ticket.yml
```diff
@@ -1,7 +1,7 @@
 name: DevEx Request
 description: Suggest improvements to GitHub project infrastructure and workflows.
 title: '[DEVEX]: '
-labels: ['devex']
+projects: ['NVIDIA/6']
```
jrhemstad, line 5:
```suggestion
labels: ['devex']
projects: ['NVIDIA/6']
```

.github/ISSUE_TEMPLATE/infra_ticket.yml
```diff
@@ -1,7 +1,7 @@
 name: Infra Ticket
 description: Request updates, report issues, etc. for CMake / GHA / CI / devcontainers.
 title: '[INFRA]: '
-labels: ['infra']
+projects: ['NVIDIA/6']
```
jrhemstad, line 5:
```suggestion
labels: ['infra']
projects: ['NVIDIA/6']
```

Issue #774, opened 2021-05-13
Add non-commutative reduction


brycelelbach:
We're missing a very useful form of reduction algorithm:

![image](https://user-images.githubusercontent.com/398194/118079414-07498d00-b36d-11eb-82f2-3c2bba6ed60a.png)

PR #7692, opened 2026-02-17
Implement `WarpReduceBatched`

### LINKED ISSUE #7564: [CUB] Implement MVP for WSPRO
Reference implementation hardcoded/manually unrolled the step and stride pattern. We have code samples from DevTech which use loops to generate the same structure as well as optimizations for the frequent swap/select operations. They are a great starting point for our implementation.

pauleonix:
## Description

Includes extensive testing and documentation/snippets (also tested).

closes #7564 

## Checklist


pauleonix:
~~Benchmarks still need to be adapted to also cover `*ToBlocked` interfaces.~~

cub/benchmarks/bench/reduce/warp_reduce_batched_sum.cu
```diff
@@ -0,0 +1,10 @@
+// SPDX-FileCopyrightText: Copyright (c) 2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
bernhardmgruber: Is this up-to-date? (maybe the code was indeed authored last yet)
```suggestion
// SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```

cub/benchmarks/bench/reduce/warp_reduce_batched_sum.cu
```diff
@@ -0,0 +1,10 @@
...
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include <nvbench_helper.cuh>
+
+using value_types = nvbench::type_list<int32_t, int64_t, float, double>;
```
bernhardmgruber: Important: if this benchmark is ever intended for tuning, please only compile for `TUNE_T` during benchmarking:

```suggestion
  #ifdef TUNE_T
  using value_types = nvbench::type_list<TUNE_T>;
  #else
  using value_types = nvbench::type_list<int32_t, int64_t, float, double>;
  #endif
```

However, I don't see any tuning parameters yet. Is this intended or is it still missing from the PR?

cub/benchmarks/bench/reduce/warp_reduce_batched_base.cuh
```diff
@@ -0,0 +1,241 @@
...
+
+#include <cuda_runtime_api.h>
+#include <device_side_benchmark.cuh>
+#include <nvbench_helper.cuh>
+
+/**
+ * @brief Benchmark operator for batched warp reduction
+ *
+ * Performs Batches reductions of LogicalWarpThreads elements each using WarpReduceBatched
+ */
```
bernhardmgruber: Important: Please use `//!` style doxygen comments

cub/cub/warp/warp_reduce_batched.cuh
```diff
@@ -0,0 +1,433 @@
...
+  // TODO: Should we allow LogicalWarpThreads = 1? (in which case everything is just no-op/copy)
+  static_assert(LogicalWarpThreads > 1 && LogicalWarpThreads <= detail::warp_threads,
+                "LogicalWarpThreads must be in the range [2, 32]");
+  static_assert(Batches >= 1, "Batches must be >= 1");
+
+#ifndef _CCCL_DOXYGEN_INVOKED // Do not document
+
+public:
+  /// Internal specialization.
+  using InternalWarpReduceBatched = detail::WarpReduceBatchedWspro<T, Batches, LogicalWarpThreads, SyncPhysicalWarp>;
```
bernhardmgruber: Q: Why do we even need `WarpReduceBatchedWspro` and can't just implement it here in this class?

cub/cub/warp/warp_reduce_batched.cuh
```diff
@@ -0,0 +1,433 @@
...
+#endif // _CCCL_DOXYGEN_INVOKED
+
+private:
+  //---------------------------------------------------------------------
+  // Constants and type definitions
+  //---------------------------------------------------------------------
+
+  static constexpr auto max_out_per_thread = ::cuda::ceil_div(Batches, LogicalWarpThreads);
+
+  /// Shared memory storage layout type for WarpReduceBatched
```
bernhardmgruber: Important: Please use a consistent documentation comment style
```suggestion
  //! Shared memory storage layout type for WarpReduceBatched
```
Applies to lots of places.

cub/cub/warp/specializations/warp_reduce_batched_wspro.cuh
```diff
@@ -0,0 +1,190 @@
...
+      : physical_lane_id(static_cast<int>(::cuda::ptx::get_sreg_laneid()))
+      , logical_lane_id(is_arch_warp ? physical_lane_id : physical_lane_id % LogicalWarpThreads)
+      , logical_warp_id(is_arch_warp ? 0 : (physical_lane_id / LogicalWarpThreads))
+      , member_mask(SyncPhysicalWarp ? 0xFFFFFFFFu : cub::WarpMask<LogicalWarpThreads>(logical_warp_id))
+  {}
+
+  //---------------------------------------------------------------------
+  // Batched reductions
+  //---------------------------------------------------------------------
+
```
bernhardmgruber: Suggestion: Such section comments do not help and rather hinder readability IMO. I would remove them.

cub/cub/warp/specializations/warp_reduce_batched_wspro.cuh
```diff
@@ -0,0 +1,190 @@
...
+  {
+    // Dispatch to more efficient intrinsics when applicable
+    constexpr bool can_use_redux =
+      ::cuda::std::is_integral_v<T> && sizeof(T) <= sizeof(unsigned)
+      && (is_cuda_minimum_maximum_v<ReductionOp, T> || is_cuda_std_plus_v<ReductionOp, T>
+          || is_cuda_std_bitwise_v<ReductionOp, T>);
+    // For 6 or more batches, WSPRO gives significantly better throughput, while latency is onlys lightly better for
+    // REDUX. For subwarps both throughput and latency are worse with REDUX independent of the number of batches. This
+    // might be a codegen issue.
+    constexpr bool redux_performs_better = is_arch_warp && Batches < 6;
```
bernhardmgruber: Suggestion: This looks a lot like a tuning parameter that should be exposed (e.g. at the class' template parameter list)

cub/cub/warp/specializations/warp_reduce_batched_wspro.cuh
```diff
@@ -0,0 +1,190 @@
...
+//! @tparam T
+//!   Data type being reduced
+//!
+//! @tparam Batches
+//!   Number of arrays to reduce in batch
+//!
+//! @tparam LogicalWarpThreads
+//!   Number of threads per logical warp (must be a power-of-two)
+template <typename T, int Batches, int LogicalWarpThreads, bool SyncPhysicalWarp>
+struct WarpReduceBatchedWspro
```
bernhardmgruber: Important: Please follow the naming convention for internal entities:

```suggestion
struct warp_reduce_batched_wspro
```
Applies all over this file.

cub/test/warp/catch2_test_warp_reduce_batched.cu
```diff
@@ -0,0 +1,778 @@
+// SPDX-FileCopyrightText: Copyright (c) 2025, NVIDIA CORPORATION. All rights reserved.
+// SPDX-License-Identifier: BSD-3-Clause
```
bernhardmgruber: Important: 
```suggestion
// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
```

cub/test/warp/catch2_test_warp_reduce_batched.cu
```diff
@@ -0,0 +1,778 @@
...
+  compute_host_reference(h_input_md, h_reference_span, reduction_op);
+
+  verify_results(h_reference, h_output);
+}
+
+/***********************************************************************************************************************
+ * Type Lists
+ **********************************************************************************************************************/
+
+// %PARAM% TEST_TYPES types 0:1:2
```
bernhardmgruber: Suggestion: Please put this at the top of the file

pauleonix:
@alliepiper @shwina @bernhardmgruber  I got rid of the benchmark and infra changes in this PR so they can be put into another, less time-critical one.

cub/cub/warp/specializations/warp_reduce_batched_wspro.cuh
pauleonix: You mean sth like just `MaxBatchesRedux = 6`? Or also a `MinThreadsRedux = 32`? I would opt for leaving exposition of this to a follow-up PR as it does not affect the interface and I foresee some bike-shedding about naming these.

cub/benchmarks/bench/reduce/warp_reduce_batched_sum.cu
pauleonix: Afaik we do not tune warp primitives/just expose tuning knobs to the user? Benchmarking got removed from this PR. Will follow up in new PR.

cub/benchmarks/bench/reduce/warp_reduce_batched_base.cuh
pauleonix: Seems like there should be no Doxygen at all in benchmarks... Benchmarking got removed from this PR. Will follow up in new PR.

cub/benchmarks/bench/reduce/warp_reduce_batched_sum.cu
pauleonix: Benchmarking got removed from this PR. Will follow up in new PR.

cub/cub/warp/warp_reduce_batched.cuh
pauleonix: It could be but it seemed nicer to me to split the API from the implementation this way.

pauleonix: I was also considering having the REDUX-based implementation as a second specialization, but opted to just copying what `WarpReduce` is doing in its `shuffle` specialization.

pauleonix: The other warp primitives have a smem implementation that is used for non-power-of-two logical warps. We decided not to support that at all b/c those old interfaces are pretty unintuitive (only a single logical warp per physical warp). We still decided to have a `TempStorage` to keep our options open for future changes in the perf/resource balance between smem and shuffles changes (I doubt it).

cub/cub/warp/specializations/warp_reduce_batched_wspro.cuh
```diff
@@ -0,0 +1,176 @@
...
+#include <cub/util_type.cuh>
+#include <cub/warp/specializations/warp_reduce_shfl.cuh>
+
+#include <cuda/__ptx/instructions/get_sreg.h>
+#include <cuda/__utility/static_for.h>
+#include <cuda/bit>
+#include <cuda/cmath>
+#include <cuda/std/__functional/operations.h>
+#include <cuda/std/array>
+#include <cuda/warp>
```
fbusato: please use precise headers

cub/cub/warp/specializations/warp_reduce_batched_wspro.cuh
```diff
@@ -0,0 +1,176 @@
...
+//!
+//! @tparam T
+//!   Data type being reduced
+//!
+//! @tparam Batches
+//!   Number of arrays to reduce in batch
+//!
+//! @tparam LogicalWarpThreads
+//!   Number of threads per logical warp (must be a power-of-two)
+template <typename T, int Batches, int LogicalWarpThreads, bool SyncPhysicalWarp>
```
fbusato, line 48: is there any constrain on `T`?

cub/cub/warp/specializations/warp_reduce_batched_wspro.cuh
```diff
@@ -0,0 +1,176 @@
...
+  //! Shared memory storage layout type
+  using TempStorage = NullType;
+
+  int physical_lane_id;
+  int logical_lane_id;
+  int logical_warp_id;
+
+  ::cuda::std::uint32_t member_mask;
+
+  _CCCL_DEVICE _CCCL_FORCEINLINE warp_reduce_batched_wspro(TempStorage& /*temp_storage*/)
```
fbusato: we now use `_CCCL_DEVICE_API`

cub/cub/warp/specializations/warp_reduce_batched_wspro.cuh
```diff
@@ -0,0 +1,176 @@
...
+      : physical_lane_id(static_cast<int>(::cuda::ptx::get_sreg_laneid()))
+      , logical_lane_id(is_arch_warp ? physical_lane_id : physical_lane_id % LogicalWarpThreads)
+      , logical_warp_id(is_arch_warp ? 0 : (physical_lane_id / LogicalWarpThreads))
+      , member_mask(SyncPhysicalWarp ? 0xFFFFFFFFu : cub::WarpMask<LogicalWarpThreads>(logical_warp_id))
+  {}
+
+  template <bool ToBlocked, typename InputT, typename OutputT, typename ReductionOp>
+  _CCCL_DEVICE _CCCL_FORCEINLINE void Reduce(const InputT& inputs, OutputT& outputs, ReductionOp reduction_op)
```
fbusato: there are no constraints for the template types

cub/cub/warp/specializations/warp_reduce_batched_wspro.cuh
```diff
@@ -0,0 +1,176 @@
...
+    ::cuda::std::array<T, max_out_per_thread> intermediate_outputs;
+
+    ::cuda::static_for<0, max_out_per_thread>([&](auto out_idx) {
+      constexpr auto first_idx      = out_idx * LogicalWarpThreads;
+      intermediate_outputs[out_idx] = RecurseReductionTree<ToBlocked, first_idx>(inputs, reduction_op);
+    });
+
+    ::cuda::static_for<0, max_out_per_thread>([&](auto out_idx) {
+      outputs[out_idx()] = intermediate_outputs[out_idx];
+    });
```
fbusato: maybe something that can be evaluated later, but `static_for` could be overkill compared to a simple pragma unroll

cub/cub/warp/specializations/warp_reduce_batched_wspro.cuh
```diff
@@ -0,0 +1,176 @@
...
+  static constexpr auto is_arch_warp = (LogicalWarpThreads == warp_threads);
+
+  static constexpr auto max_out_per_thread = ::cuda::ceil_div(Batches, LogicalWarpThreads);
+
+  //! Shared memory storage layout type
+  using TempStorage = NullType;
+
+  int physical_lane_id;
+  int logical_lane_id;
+  int logical_warp_id;
```
fbusato, line 70: these variables can also be moved to local variables in each function

cub/cub/warp/specializations/warp_reduce_batched_wspro.cuh
```diff
@@ -0,0 +1,176 @@
...
+      // Chose to use the last batch to avoid relying on default construction of T.
+      return inputs[Batches - 1];
+    }
+    else if constexpr (PrevStride == 1)
+    {
+      // Recursion base case
+      return inputs[base_idx];
+    }
+    // Explicit "else" branch needed to avoid compiler error on "% 0".
+    else
```
fbusato: this is the right choise also to improve compile time

cub/cub/warp/warp_reduce_batched.cuh
```diff
@@ -0,0 +1,429 @@
...
+CUB_NAMESPACE_BEGIN
+
+//! @rst
+//! The ``WarpReduceBatched`` class provides :ref:`collective <collective-primitives>` methods for performing
+//! batched parallel reductions of multiple arrays partitioned across a CUDA thread warp using the WSPRO algorithm.
+//!
+//! Overview
+//! ++++++++
+//!
+//! - Performs batched reductions of Batches arrays, each containing LogicalWarpThreads elements
```
fbusato: question. Does the order of the operations differs depending on `ReduceRedux` or `RecurseReductionTree` path?
we should clearly state that we have potential different values based on heuristics

cub/cub/warp/warp_reduce_batched.cuh
```diff
@@ -0,0 +1,429 @@
...
+//!
+//! @tparam LogicalWarpThreads
+//!   The number of threads per logical warp / elements per array. Must be a power-of-two in range [2, 32].
+//!   Default is the warp size of the targeted CUDA compute-capability (e.g., 32).
+//!
+template <typename T, int Batches, int LogicalWarpThreads = detail::warp_threads, bool SyncPhysicalWarp = false>
+class WarpReduceBatched
+{
+  static_assert(::cuda::is_power_of_two(LogicalWarpThreads), "LogicalWarpThreads must be a power of two");
+  // TODO: Should we allow LogicalWarpThreads = 1? (in which case everything is just no-op/copy)
```
fbusato: strongly suggested yes. The reason is that the user can have this class within a template code and non-trivial conditions on it. `LogicalWarpThreads = 1` will handle all cases at negligible cost on our side

cub/cub/warp/warp_reduce_batched.cuh
```diff
@@ -0,0 +1,429 @@
...
+  //! @return
+  //!   The reduction of the input values of the batch corresponding to the logical lane.
+  template <typename InputT, typename ReductionOp>
+  _CCCL_DEVICE _CCCL_FORCEINLINE T Reduce(const InputT& inputs, ReductionOp reduction_op)
+  {
+    static_assert(detail::is_fixed_size_random_access_range_v<InputT>,
+                  "InputT must support the subscript operator[] and have a compile-time size");
+    static_assert(detail::static_size_v<InputT> == Batches, "Input size must match Batches");
+    // These restrictions could be relaxed to allow type-conversions
+    static_assert(::cuda::std::is_same_v<::cuda::std::iter_value_t<InputT>, T>, "Input element type must match T");
```
fbusato: ah, got it. Makes sense to keep the constrains at higher level here

cub/cub/warp/warp_reduce_batched.cuh
```diff
@@ -0,0 +1,429 @@
...
+  //!
+  //! @param[out] outputs
+  //!   Statically-sized array-like container where thread i stores reductions sequentially:
+  //!   ``outputs[0]`` = result of batch i, ``outputs[1]`` = result of batch (i + LogicalWarpThreads), etc.
+  //!
+  //! @param[in] reduction_op
+  //!   Binary reduction operator
+  template <typename InputT, typename OutputT, typename ReductionOp>
+  _CCCL_DEVICE _CCCL_FORCEINLINE void ReduceToStriped(const InputT& inputs, OutputT& outputs, ReductionOp reduction_op)
+  {
```
fbusato, line 261: question. why `outputs` cannot be a return value?

cub/cub/warp/warp_reduce_batched.cuh
```diff
@@ -0,0 +1,429 @@
...
+                  "InputT must support the subscript operator[] and have a compile-time size");
+    static_assert(detail::is_fixed_size_random_access_range_v<OutputT>,
+                  "OutputT must support the subscript operator[] and have a compile-time size");
+    static_assert(detail::static_size_v<InputT> == Batches, "Input size must match Batches");
+    static_assert(detail::static_size_v<OutputT> == max_out_per_thread,
+                  "Output size must match ceil_div(Batches, LogicalWarpThreads)");
+    // These restrictions could be relaxed to allow type-conversions
+    static_assert(::cuda::std::is_same_v<::cuda::std::iter_value_t<InputT>, T>, "Input element type must match T");
+    static_assert(::cuda::std::is_same_v<::cuda::std::iter_value_t<OutputT>, T>, "Output element type must match T");
+
```
fbusato: these are repetitive. suggestion: use a function or trait for the constraints

fbusato (changes requested):

cub/cub/warp/warp_reduce_batched.cuh
pauleonix: REDUX currently only applies to operations on integers that are associative and commutative.

pauleonix: And even the newer FP32 min and max (not implemented yet) do not depend on ordering.

cub/cub/warp/specializations/warp_reduce_batched_wspro.cuh
pauleonix: What would be the advantage of that?

cub/cub/warp/warp_reduce_batched.cuh
pauleonix: I kept it in line with other CUB warp/block primitves that return more than one element per thread. On the other hand it causes a potentially unintuitive difference between `[Reduce|Sum]` (returning in the style of `WarpReduce`) and `[Reduce|Sum]To[Blocked|Striped]` which is your point, I guess? I'm fine with returning from all of them. It also gets rid of that additional copy loop.

pauleonix: One (minor?) reason to keep it this way is that it allows the user to use a statically-sized range of their choice. If we return `cuda::std::array` and the user needs to pass the results as a C-style array to another function, the user has to do the ugly copy loop.

cub/cub/warp/specializations/warp_reduce_batched_wspro.cuh
pauleonix: I added the missing `has_binary_call_operator` constraint. Other than that the only constraints on `T` are ones checked at the shuffle level (default constructible and not a pointer). Do you think I should duplicate them at this higher level?

cub/cub/warp/warp_reduce_batched.cuh
elstehle: > One (minor?) reason to keep it this way is that it allows the user to use a statically-sized range of their choice. If we return cuda::std::array and the user needs to pass the results as a C-style array to another function, the user has to do the ugly copy loop.

Sounds like a good justification to keep it as is. I wouldn't want to deviate from the rest of CUB's cooperative interfaces.

cub/cub/warp/specializations/warp_reduce_batched_wspro.cuh
fbusato: no, better a higher level. I saw this file before the other one

cub/cub/warp/warp_reduce_batched.cuh
fbusato: good point. This also remember me the work for `WarpReduce` that I have to recover one day

cub/cub/warp/specializations/warp_reduce_batched_wspro.cuh
```diff
@@ -0,0 +1,179 @@
...
+    {
+      constexpr auto stride             = PrevStride / 2;
+      constexpr auto offset_idx_striped = BaseIdxStriped + stride;
+
+      const auto base_value   = RecurseReductionTree<ToBlocked, BaseIdxStriped, stride>(inputs, reduction_op);
+      const auto offset_value = RecurseReductionTree<ToBlocked, offset_idx_striped, stride>(inputs, reduction_op);
+
+      // Each left lane exchanges its offset value against a right lane's base value
+      const auto is_left_lane = logical_lane_id % PrevStride < stride;
```
wmaxey:
```suggestion
      const auto is_left_lane = (logical_lane_id % PrevStride) < stride;
```

cub/cub/warp/warp_reduce_batched.cuh
fbusato: makes sense. we will have interface discussion when we will redesign the algorithm with thread hierarchy

cub/cub/warp/specializations/warp_reduce_batched_wspro.cuh
fbusato: no performance advantages. It improves readability because the variables are close to where they are used.

cub/cub/warp/specializations/warp_reduce_batched_wspro.cuh
```diff
@@ -0,0 +1,179 @@
...
+
+  template <bool ToBlocked, typename InputT, typename OutputT, typename ReductionOp>
+  _CCCL_DEVICE_API _CCCL_FORCEINLINE void Reduce(const InputT& inputs, OutputT& outputs, ReductionOp reduction_op)
+  {
+    // Dispatch to more efficient intrinsics when applicable
+    constexpr bool can_use_redux =
+      ::cuda::std::is_integral_v<T> && sizeof(T) <= sizeof(unsigned)
+      && (is_cuda_minimum_maximum_v<ReductionOp, T> || is_cuda_std_plus_v<ReductionOp, T>
+          || is_cuda_std_bitwise_v<ReductionOp, T>);
+    // For 6 or more batches, WSPRO gives significantly better throughput, while latency is onlys lightly better for
```
fbusato: it is now not consistent with `Batches <= 8`

cub/cub/warp/specializations/warp_reduce_batched_wspro.cuh
```diff
@@ -0,0 +1,179 @@
...
+    // REDUX. For subwarps both throughput and latency are worse with REDUX independent of the number of batches. This
+    // might be a codegen issue.
+    constexpr bool redux_performs_better = is_arch_warp && Batches <= 8;
```
fbusato, line 92: question. Does it mean that when we execute reduce on subwarp, WSPRO is always faster than the basic approach?

cub/cub/warp/specializations/warp_reduce_batched_wspro.cuh
```diff
@@ -0,0 +1,179 @@
...
+
+    // Needed in case of outputs aliasing inputs
+    ::cuda::std::array<T, max_out_per_thread> intermediate_outputs;
+
+    ::cuda::static_for<0, max_out_per_thread>([&](auto out_idx) {
+      constexpr auto first_idx      = out_idx * LogicalWarpThreads;
+      intermediate_outputs[out_idx] = RecurseReductionTree<ToBlocked, first_idx>(inputs, reduction_op);
+    });
+
+#pragma unroll
```
fbusato: use `_CCCL_PRAGMA_UNROLL_FULL()`

cub/cub/warp/specializations/warp_reduce_batched_wspro.cuh
```diff
@@ -0,0 +1,179 @@
...
+    constexpr bool redux_performs_better = is_arch_warp && Batches <= 8;
+    if constexpr (can_use_redux && redux_performs_better)
+    {
+      NV_IF_TARGET(NV_PROVIDES_SM_80, (ReduceRedux<ToBlocked>(inputs, outputs, reduction_op); return;))
+    }
+
+    // Needed in case of outputs aliasing inputs
+    ::cuda::std::array<T, max_out_per_thread> intermediate_outputs;
+
+    ::cuda::static_for<0, max_out_per_thread>([&](auto out_idx) {
```
fbusato, line 101: not critical for this PR, but I would suggest to use unroll directive instead of `static_for`

cub/cub/warp/specializations/warp_reduce_batched_wspro.cuh
```diff
@@ -0,0 +1,179 @@
...
+      const auto offset_value = RecurseReductionTree<ToBlocked, offset_idx_striped, stride>(inputs, reduction_op);
+
+      // Each left lane exchanges its offset value against a right lane's base value
+      const auto is_left_lane = logical_lane_id % PrevStride < stride;
+      const auto keep_value   = is_left_lane ? base_value : offset_value;
+      auto exch_value         = is_left_lane ? offset_value : base_value;
+      exch_value              = ::cuda::device::warp_shuffle_xor(exch_value, stride, member_mask);
+      // While the current implementation is possibly faster, another conditional swap here would allow for
+      // non-commutative reductions which might be useful for (segmented) scan operations.
+      return reduction_op(exch_value, keep_value);
```
fbusato:
```suggestion
      return static_cast<T>(reduction_op(exch_value, keep_value));
```

cub/cub/warp/warp_reduce_batched.cuh
```diff
@@ -0,0 +1,422 @@
...
+//! @endrst
+//!
+//! @tparam T
+//!   The reduction input/output element type
+//!
+//! @tparam Batches
+//!   The number of arrays to reduce in batch. Best performance when Batches = LogicalWarpThreads.
+//!
+//! @tparam LogicalWarpThreads
+//!   The number of threads per logical warp / elements per array. Must be a power-of-two in range [2, 32].
```
fbusato: inconsistent with `[1, 32]`

cub/cub/warp/warp_reduce_batched.cuh
```diff
@@ -0,0 +1,422 @@
...
+  //!
+  //! @param[in] inputs
+  //!   Statically-sized array-like container of Batches input values from calling thread
+  //!
+  //! @return
+  //!   The sum of the input values of the batch corresponding to the logical lane.
+  template <typename InputT>
+  _CCCL_DEVICE_API _CCCL_FORCEINLINE T Sum(const InputT& inputs)
+  {
+    return Reduce(inputs, ::cuda::std::plus<>{});
```
fbusato, line 359: probably for a later PR, but if the input is `short`, we keep performing the operation in `int32`, accumulating, then cast back

fbusato (changes requested):

cub/cub/warp/specializations/warp_reduce_batched_wspro.cuh
pauleonix: In my benchmarks, yes.

cub/cub/warp/specializations/warp_reduce_batched_wspro.cuh
pauleonix: For this particular one `static_for` is needed as the index is the starting point of the static recursion.

cub/cub/warp/specializations/warp_reduce_batched_wspro.cuh
pauleonix: I made the names more descriptive than in other places that have these, so I think it should not be a problem. I would like to avoid repeating the code that is initializing them in each member function.

cub/cub/warp/specializations/warp_reduce_batched_wspro.cuh
pauleonix: But I replaced all other occurrences now.

cub/cub/warp/warp_reduce_batched.cuh
pauleonix: AFAIK the current behavior is consistent with `WarpReduce`. But sure, we could improve both in this regard in the future.

cub/cub/warp/specializations/warp_reduce_batched_wspro.cuh
```diff
@@ -0,0 +1,187 @@
...
+//!
+//! @tparam T
+//!   Data type being reduced
+//!
+//! @tparam Batches
+//!   Number of arrays to reduce in batch
+//!
+//! @tparam LogicalWarpThreads
+//!   Number of threads per logical warp (must be a power-of-two)
+template <typename T, int Batches, int LogicalWarpThreads, bool SyncPhysicalWarp>
```
miscco, line 50: Question: I assume this is something that we will want to be able to tune. Is this done in the higher level algorithms?

cub/cub/warp/specializations/warp_reduce_batched_wspro.cuh
miscco: I believe that we should keep them here. @ahendriksen found that those variables can introduce a nontrivial amount of latency in a function

miscco: We also might want to reuse `SpecialRegisters`

cub/cub/warp/specializations/warp_reduce_batched_wspro.cuh
```diff
@@ -0,0 +1,187 @@
...
+  {}
+
+  template <bool ToBlocked, typename InputT, typename OutputT, typename ReductionOp>
+  _CCCL_DEVICE_API _CCCL_FORCEINLINE void Reduce(const InputT& inputs, OutputT& outputs, ReductionOp reduction_op)
+  {
+    // Dispatch to more efficient intrinsics when applicable
+    constexpr bool can_use_redux =
+      ::cuda::std::is_integral_v<T> && sizeof(T) <= sizeof(unsigned)
+      && (is_cuda_minimum_maximum_v<ReductionOp, T> || is_cuda_std_plus_v<ReductionOp, T>
+          || is_cuda_std_bitwise_v<ReductionOp, T>);
```
miscco: Nitpick: This is the second place we are determining this next to `warp_reduce_shfl`. Please move that definition into `thread_operators.cuh`

Critical: This is potentially failing to recognize qualified operators, please use `remove_cvref_t<ReductionOp>` instead of raw `ReductionOp`

cub/cub/warp/specializations/warp_reduce_batched_wspro.cuh
```diff
@@ -0,0 +1,187 @@
...
+    // For more batches, WSPRO gives significantly better throughput, while latency is onlys lightly better for
```
miscco:
```suggestion
    // For more batches, WSPRO gives significantly better throughput, while latency is only slightly better for
```

cub/cub/warp/warp_reduce_batched.cuh
```diff
@@ -0,0 +1,460 @@
...
+  //! @param[in] inputs
+  //!   Statically-sized input range holding ``Batches`` items
+  //!
+  //! @param[in] reduction_op
+  //!   Binary reduction operator
+  //!
+  //! @return
+  //!   The reduction of the input values of the batch corresponding to the logical lane.
+  template <typename InputT, typename ReductionOp>
+  _CCCL_DEVICE_API _CCCL_FORCEINLINE T Reduce(const InputT& inputs, ReductionOp reduction_op)
```
miscco: Nitpick: 
```suggestion
  [[nodiscard]] _CCCL_DEVICE_API _CCCL_FORCEINLINE T Reduce(const InputT& inputs, ReductionOp reduction_op)
```

cub/cub/warp/warp_reduce_batched.cuh
```diff
@@ -0,0 +1,460 @@
...
+  //! @param[out] outputs
+  //!   Statically-sized output range holding ``ceil_div(Batches, LogicalWarpThreads)`` items
+  //!
+  //! @param[in] reduction_op
+  //!   Binary reduction operator
+  template <typename InputT, typename OutputT, typename ReductionOp>
+  _CCCL_DEVICE_API _CCCL_FORCEINLINE void
+  ReduceToStriped(const InputT& inputs, OutputT& outputs, ReductionOp reduction_op)
+  {
+    (void) constraints<InputT, OutputT, ReductionOp>{};
```
miscco: Nitpick:
```suggestion
    [[maybe_unused]] constraints<InputT, OutputT, ReductionOp>{};
```

cub/cub/warp/warp_reduce_batched.cuh
```diff
@@ -0,0 +1,460 @@
...
+  //!   **[inferred]** The data type to be reduced having member
+  //!   ``operator[](int i)`` and must be statically-sized (``size()`` method or static array)
+  //!
+  //! @param[in] inputs
+  //!   Statically-sized input range holding ``Batches`` items
+  //!
+  //! @return
+  //!   The sum of the input values of the batch corresponding to the logical lane.
+  template <typename InputT>
+  _CCCL_DEVICE_API _CCCL_FORCEINLINE T Sum(const InputT& inputs)
```
miscco: Nitpick: 
```suggestion
  [[nodiscard]] _CCCL_DEVICE_API _CCCL_FORCEINLINE T Sum(const InputT& inputs)
```

cub/cub/warp/specializations/warp_reduce_batched_wspro.cuh
pauleonix: For both warp- and block-primitives we tend to leave perf-tuning up to the user (up to some default choices as a starting point). Afaik only device-primitives get the tuning treatment (where we try to find out good values for compile-time meta-parameters like items-per-thread). While device-primitive tuning already has the problem that not all possible custom functors can be tuned for, this is even more of an issue with warp- and block-primitives as we have no idea what else the user is doing in the rest of the kernel.

cub/cub/warp/warp_reduce_batched.cuh
pauleonix: This actually does not work with a constructor like this. I transformed it into a function so neither `(void)` nor `[[maybe_unsused]]` are needed anymore.

cub/test/warp/catch2_test_warp_reduce_batched_api.cu
```diff
@@ -0,0 +1,339 @@
...
+C2H_TEST("WarpReduceBatched::SumToStriped documentation kernel", "[warp][reduce][batched]")
+{
+  c2h::device_vector<int> d_out(20);
+
+  WarpReduceBatchedSumToStripedApiKernel<<<1, 8>>>(thrust::raw_pointer_cast(d_out.data()));
+  REQUIRE(cudaSuccess == cudaPeekAtLastError());
+  REQUIRE(cudaSuccess == cudaDeviceSynchronize());
+
+  c2h::host_vector<int> expected{-3, -1, 1, 3, 5, 1, 3, 5, 7, 9, 5, 7, 9, 11, 13, 9, 11, 13, 15, 17};
+  REQUIRE(expected == d_out);
```
elstehle: I assume this was supposed to test `WarpReduceBatchedSumToBlockedApiKernel`?

cub/cub/warp/specializations/warp_reduce_batched_wspro.cuh
```diff
@@ -0,0 +1,183 @@
...
+      constexpr auto offset_idx_striped = BaseIdxStriped + stride;
+
+      const auto base_value   = RecurseReductionTree<ToBlocked, BaseIdxStriped, stride>(inputs, reduction_op);
+      const auto offset_value = RecurseReductionTree<ToBlocked, offset_idx_striped, stride>(inputs, reduction_op);
+
+      // Each left lane exchanges its offset value against a right lane's base value
+      const auto is_left_lane = (logical_lane_id % PrevStride) < stride;
+      const auto keep_value   = is_left_lane ? base_value : offset_value;
+      auto exch_value         = is_left_lane ? offset_value : base_value;
+      exch_value              = ::cuda::device::warp_shuffle_xor(exch_value, stride, member_mask);
```
elstehle, line 176: suggestion: I think requiring commutative operators is something we'd want to point out in the documentation. We can drop that requirement, once we add a version that supports non-commutative operators.

cub/cub/warp/specializations/warp_reduce_batched_wspro.cuh
```diff
@@ -0,0 +1,183 @@
...
+//!
+//! @tparam LogicalWarpThreads
+//!   Number of threads per logical warp (must be a power-of-two)
+template <typename T, int Batches, int LogicalWarpThreads, bool SyncPhysicalWarp>
+struct warp_reduce_batched_wspro
+{
+  static_assert(::cuda::is_power_of_two(LogicalWarpThreads), "LogicalWarpThreads must be a power of two");
+  static_assert(LogicalWarpThreads > 0 && LogicalWarpThreads <= warp_threads,
+                "LogicalWarpThreads must be in the range [1, 32]");
+  static_assert(Batches >= 0, "Batches must be >= 0");
```
elstehle, line 56: question: Does a `Batches==0` invocation make sense? I would be concerned about expressions like `inputs[Batches - 1];`.

cub/cub/warp/warp_reduce_batched.cuh
```diff
@@ -0,0 +1,460 @@
...
+//! @tparam T
+//!   The reduction input/output element type
+//!
+//! @tparam Batches
+//!   The number of batches to reduce. Also corresponds to the size of the range of inputs for each thread.
+//!
+//! @tparam LogicalWarpThreads
+//!   <b>[optional]</b> The number of threads per "logical" warp (may be less than the number of
+//!   hardware warp threads but has to be a power of two).  Default is the warp size of the targeted CUDA
+//!   compute-capability (e.g., 32 threads for SM20).
```
elstehle: nit: SM20 feels old 🤣 ...we could mention something _slightly_ more recent, like SM80?
```suggestion
//!   compute-capability (e.g., 32 threads for SM80).
```

cub/cub/warp/warp_reduce_batched.cuh
```diff
@@ -0,0 +1,460 @@
...
+//!
+//! Overview
+//! ++++++++
+//!
+//! - A `reduction <http://en.wikipedia.org/wiki/Reduce_(higher-order_function)>`__ (or *fold*) uses a binary combining
+//!   operator to compute a single aggregate from a list of input elements.
+//! - Supports "logical" warps smaller than the physical warp size (e.g., logical warps of 8 threads)
+//! - This primitive performs batched reductions taking one item per batch per thread.
+//! - Results are distributed among the threads (given more batches than logical warp threads, either in a striped or
+//! blocked manner).
```
elstehle: nit:
```suggestion
//! - Results are distributed among the threads. When there are more batches than logical warp threads, results are distributed among threads in
   a striped or blocked manner.
```

cub/cub/warp/specializations/warp_reduce_batched_wspro.cuh
```diff
@@ -0,0 +1,183 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+//! @file
+//! cub::WarpReduceBatchedWspro provides WSPRO-based batched parallel reduction of items partitioned across a CUDA
```
elstehle:
```suggestion
//! cub::warp_reduce_batched_wspro provides WSPRO-based batched parallel reduction of items partitioned across a CUDA
```

cub/cub/warp/warp_reduce_batched.cuh
```diff
@@ -0,0 +1,460 @@
...
+//! - Uses special instructions when applicable (e.g., warp ``SHFL`` instructions)
+//! - Uses synchronization-free communication between warp lanes when applicable
+//! - Should generally give much better performance than sequential ``WarpReduce`` calls independent of blocked or
+//! striped output arrangements.
+//! - Achieves peak efficiency when the number of batches is a multiple of the number of logical warp threads.
+//! - For smaller than physical warp size logical warps, using ``SyncPhysicalWarp = true`` should in general give better
+//! performance than ``SyncPhysicalWarp = false``.
+//!   Note that it can cause a deadlock if not all non-exited logical warps from the same physical warp participate in
+//!   the reduction (due to branches).
+//! - Uneven number of batches are less efficient than the next higher even ones.
```
elstehle: nit:
```suggestion
//! - An uneven number of batches is less efficient than the next higher even ones.
```

elstehle (approved):
Looks good to me. Feel free to defer any feedback unrelated to tests, API contract (including its documentation), or correctness to follow-up work.                                                                                                                                                

@fbusato when you have some bandwidth, could you please help re-review? We have downstream users looking to start using this.

cub/test/warp/catch2_test_warp_reduce_batched_api.cu
elstehle: ...should be sufficient to rename `SumToStriped ` to `WarpReduceBatchedSumToBlockedApiKernel`.

cub/cub/warp/specializations/warp_reduce_batched_wspro.cuh
fbusato: I don't think `Batches==0` should be allowed. Otherwise, it is hard to define a valid input/output

cub/cub/warp/specializations/warp_reduce_batched_wspro.cuh
fbusato: we can use `cuda::is_commutative_v` https://nvidia.github.io/cccl/unstable/libcudacxx/extended_api/functional/operator_properties.html#libcudacxx-extended-api-functional-operator-properties

cub/cub/warp/specializations/warp_reduce_batched_wspro.cuh
fbusato: the current code is fine. Anyway, recent compiler versions are able to remove duplicate instantiations of these variables. So it is only a style choice

cub/cub/warp/specializations/warp_reduce_batched_wspro.cuh
```diff
@@ -0,0 +1,183 @@
...
+#include <cub/util_type.cuh>
+// For REDUX helper
+#include <cub/warp/specializations/warp_reduce_shfl.cuh>
+
+#include <cuda/__cmath/ceil_div.h>
+#include <cuda/__cmath/pow2.h>
+#include <cuda/__ptx/instructions/get_sreg.h>
+#include <cuda/__utility/static_for.h>
+#include <cuda/__warp/warp_shuffle.h>
+#include <cuda/std/__functional/operations.h>
```
fbusato: is this header used?

cub/cub/warp/specializations/warp_reduce_batched_wspro.cuh
```diff
@@ -0,0 +1,183 @@
...
+    {
+      outputs[i] = intermediate_outputs[i];
+    }
+  }
+
+  template <bool ToBlocked, typename InputT, typename OutputT, typename ReductionOp>
+  _CCCL_DEVICE_API _CCCL_FORCEINLINE void ReduceRedux(const InputT& inputs, OutputT& outputs, ReductionOp reduction_op)
+  {
+    // Needed to avoid compiler error on "/ 0" and improve compile time.
+    if constexpr (Batches != 0)
```
fbusato, line 115: I think we can remove this condition

cub/cub/warp/specializations/warp_reduce_batched_wspro.cuh
```diff
@@ -0,0 +1,183 @@
...
+    {
+      // Needed in case of outputs aliasing inputs
+      ::cuda::std::array<T, max_out_per_thread> intermediate_outputs;
+
+      // Can't use the full member mask given SyncPhysicalWarp==true because it affects the result of the reduction.
+      const auto reduce_mask = cub::WarpMask<LogicalWarpThreads>(logical_warp_id);
+      _CCCL_PRAGMA_UNROLL_FULL()
+      for (int i = 0; i < Batches; ++i)
+      {
+        auto result         = reduce_op_sync(inputs[i], reduce_mask, reduction_op);
```
fbusato:
```suggestion
        auto result         = cub::detail::reduce_op_sync(inputs[i], reduce_mask, reduction_op);
```

cub/cub/warp/warp_reduce_batched.cuh
```diff
@@ -0,0 +1,460 @@
...
+#endif // no system header
+
+#include <cub/detail/type_traits.cuh>
+#include <cub/util_type.cuh>
+#include <cub/warp/specializations/warp_reduce_batched_wspro.cuh>
+
+#include <cuda/__cmath/ceil_div.h>
+#include <cuda/__cmath/pow2.h>
+#include <cuda/std/__functional/operations.h>
+#include <cuda/std/__iterator/iterator_traits.h>
```
fbusato:
```suggestion
#include <cuda/std/__iterator/readable_traits.h>
```

cub/cub/warp/warp_reduce_batched.cuh
```diff
@@ -0,0 +1,460 @@
...
+//!   hardware warp threads but has to be a power of two).  Default is the warp size of the targeted CUDA
+//!   compute-capability (e.g., 32 threads for SM20).
+//!
+template <typename T, int Batches, int LogicalWarpThreads = detail::warp_threads, bool SyncPhysicalWarp = false>
+class WarpReduceBatched
+{
+  static_assert(::cuda::is_power_of_two(LogicalWarpThreads), "LogicalWarpThreads must be a power of two");
+  static_assert(LogicalWarpThreads > 0 && LogicalWarpThreads <= detail::warp_threads,
+                "LogicalWarpThreads must be in the range [1, 32]");
+  static_assert(Batches >= 0, "Batches must be >= 0");
```
fbusato, line 102: same here, `Batches > 0`

cub/test/warp/catch2_test_warp_reduce_batched_api.cu
fbusato: btw, this test case is duplicated from line 283.

cub/cub/warp/specializations/warp_reduce_batched_wspro.cuh
pauleonix: @Jacobfaib requested it so heavily templated code does not have to do special handling. In contrast to C-arrays, other static ranges can be empty. Naturally this is only supported for the more general `To[Striped|Blocked]` interface since `Reduce|Sum` need to return a `T`.

For WSPRO the support was trivial since the outermost loops go over `ceil_div(Batches, LogicalWarpThreads)` (i.e. those loops are not generated at all). REDUX needed one extra `if constexpr` to avoid division-by-zero compiler errors.

cub/cub/warp/specializations/warp_reduce_batched_wspro.cuh
```diff
@@ -0,0 +1,183 @@
...
+  }
+
+  template <bool ToBlocked, typename InputT, typename OutputT, typename ReductionOp>
+  _CCCL_DEVICE_API _CCCL_FORCEINLINE void ReduceRedux(const InputT& inputs, OutputT& outputs, ReductionOp reduction_op)
+  {
+    // Needed to avoid compiler error on "/ 0" and improve compile time.
+    if constexpr (Batches != 0)
+    {
+      // Needed in case of outputs aliasing inputs
+      ::cuda::std::array<T, max_out_per_thread> intermediate_outputs;
```
fbusato, line 118: should this variable be initialzied? it looks that the `outputs` can have undefined values

cub/cub/warp/warp_reduce_batched.cuh
pauleonix: I copied that from `WarpReduce` which is not that old 🤣 Agreed that it could be updated.

cub/cub/warp/specializations/warp_reduce_batched_wspro.cuh
```diff
@@ -0,0 +1,183 @@
...
+
+#include <cub/util_ptx.cuh>
+#include <cub/util_type.cuh>
+// For REDUX helper
+#include <cub/warp/specializations/warp_reduce_shfl.cuh>
+
+#include <cuda/__cmath/ceil_div.h>
+#include <cuda/__cmath/pow2.h>
+#include <cuda/__ptx/instructions/get_sreg.h>
+#include <cuda/__utility/static_for.h>
```
fbusato, line 30:
```suggestion
#include <cuda/__utility/static_for.h>
#include <cub/thread/thread_operators.cuh>
#include <cub/util_arch.cuh>
#include <cuda/std/cstdint>
```

cub/cub/warp/warp_reduce_batched.cuh
```diff
@@ -0,0 +1,460 @@
...
+//! @tparam Batches
+//!   The number of batches to reduce. Also corresponds to the size of the range of inputs for each thread.
+//!
+//! @tparam LogicalWarpThreads
+//!   <b>[optional]</b> The number of threads per "logical" warp (may be less than the number of
+//!   hardware warp threads but has to be a power of two).  Default is the warp size of the targeted CUDA
+//!   compute-capability (e.g., 32 threads for SM20).
+//!
+template <typename T, int Batches, int LogicalWarpThreads = detail::warp_threads, bool SyncPhysicalWarp = false>
+class WarpReduceBatched
```
fbusato, line 97: should not be `struct`?

fbusato (approved):
correctness, interface, style look good. Just a couple of minor suggestions.

cub/cub/warp/specializations/warp_reduce_batched_wspro.cuh
Jacobfaib: > requested it so heavily templated code does not have to do special handling. 

Right, for example suppose the user wants to use this algorithm over an ever shrinking region. They might generate such a loop at compile-time using `make_index_sequence()` which starts at 0 which they cannot change. We can make their lives easier by transparently handling the 0-sized cases.

> In contrast to C-arrays, other static ranges can be empty. 

For completeness, stuff like `[cuda::]std::array` is allowed to be 0 sized.

>  since Reduce|Sum need to return a T.

Nit: you could just return a `T{0}` for these like the usual numerical algorithms.

cub/cub/warp/specializations/warp_reduce_batched_wspro.cuh
pauleonix: Not directly anymore, since I put the REDUX check into `thread_operators.cuh`. Good catch!

cub/cub/warp/specializations/warp_reduce_batched_wspro.cuh
pauleonix: See https://github.com/NVIDIA/cccl/pull/7692#discussion_r3112451995

cub/cub/warp/specializations/warp_reduce_batched_wspro.cuh
pauleonix: I actually document it as having undefined value for positions that do not have a corresponding input batch, but I guess it should rather be initialized and documented as "invalid" instead. WSPRO does actually always write to all of the positions, but the REDUX implementation does not.

cub/cub/warp/warp_reduce_batched.cuh
pauleonix: `WarpReduce` is a `class` as well. While it does not make a big difference in the language, I would prefer using `struct` for POD structs.

cub/cub/warp/specializations/warp_reduce_batched_wspro.cuh
pauleonix: I think that would be more surprising. Generic code should use the generic interface. The simplified interface does not allow `Batches > LogicalWarpThreads` and so I do not see a problem with it als not allowing `Batches < 1`. It handles all (`Batches`, `LogicalWarpThreads`) combinations that fulfill `ceil_div(Batches, LogicalWarpThreads) == 1`.

cub/cub/warp/specializations/warp_reduce_batched_wspro.cuh
pauleonix: To me "reduction" implies that operators are commutative (related #774). See also [`std::reduce`](https://en.cppreference.com/cpp/algorithm/reduce):
> - The result is non-deterministic if the `binary_op` is not associative or not commutative (such as floating-point addition).

But I just realized the wikipedia article we link is about the general fold, not the parallel algorithm.

cub/cub/warp/specializations/warp_reduce_batched_wspro.cuh
```diff
@@ -0,0 +1,185 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+//! @file
+//! cub::warp_reduce_batched_wspro provides WSPRO-based batched parallel reduction of items partitioned across a CUDA
+//! thread warp.
```
oleksandr-pavlyk, line 6:
```suggestion
//! cub::warp_reduce_batched_wspro performs WSPRO-based batched parallel reduction of items partitioned 
//! across a CUDA warp.
```

cub/cub/warp/specializations/warp_reduce_batched_wspro.cuh
```diff
@@ -0,0 +1,185 @@
...
+#include <cuda/__warp/warp_shuffle.h>
+#include <cuda/std/array>
+#include <cuda/std/cstdint>
+
+CUB_NAMESPACE_BEGIN
+
+namespace detail
+{
+//! @brief WarpReduceBatchedWspro provides WSPRO (Warp Shuffle Parallel Reduction Optimization) based
+//!        batched parallel reduction of items partitioned across a CUDA thread warp.
```
oleksandr-pavlyk, line 40:
```suggestion
//!        batched parallel reduction of items partitioned across a CUDA warp.
```

cub/cub/warp/warp_reduce_batched.cuh
```diff
@@ -0,0 +1,467 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+//! @file
+//! @rst
+//! The ``cub::WarpReduceBatched`` class provides :ref:`collective <collective-primitives>` methods for
+//! performing batched parallel reductions of multiple arrays partitioned across a CUDA thread warp.
```
oleksandr-pavlyk, line 7:
```suggestion
//! performing batched parallel reductions of multiple arrays partitioned across a CUDA warp.
```

cub/cub/warp/warp_reduce_batched.cuh
```diff
@@ -0,0 +1,467 @@
...
+#include <cuda/__cmath/pow2.h>
+#include <cuda/std/__functional/operations.h>
+#include <cuda/std/__iterator/readable_traits.h>
+#include <cuda/std/__type_traits/is_same.h>
+
+CUB_NAMESPACE_BEGIN
+
+//! @rst
+//! The ``WarpReduceBatched`` class provides :ref:`collective <collective-primitives>` methods for performing
+//! batched parallel reductions of multiple batches of items partitioned across a CUDA thread warp.
```
oleksandr-pavlyk, line 36:
```suggestion
//! batched parallel reductions of multiple batches of items partitioned across a CUDA warp.
```

cub/test/warp/catch2_test_warp_reduce_batched_api.cu
```diff
@@ -0,0 +1,339 @@
...
+#include <cuda/__functional/maximum.h>
+#include <cuda/std/array>
+#include <cuda/std/span>
+
+#include <c2h/catch2_test_helper.h>
+
+__global__ __launch_bounds__(64) void WarpReduceBatchedOverviewKernel(int* out)
+{
+  // example-begin warp-reduce-batched-overview
+  using WarpReduceBatched = cub::WarpReduceBatched<int, 3>;
```
oleksandr-pavlyk, line 19:
```suggestion
  constexpr int num_batches = 3;
  using WarpReduceBatched = cub::WarpReduceBatched<int, num_batches>;
```

Consider naming the literal values to lower cognitive load of reading the example. 
Use the compile-time constant in `int thread_data[num_batches];` and in the later conditional:

```
  if (logical_lane_id < num_batches)
  {
    out[logical_warp_id * num_batches + logical_lane_id] = result;
  }
```

cub/test/warp/catch2_test_warp_reduce_batched_api.cu
```diff
@@ -0,0 +1,339 @@
...
+  REQUIRE(expected == d_out);
+}
+
+__global__ void WarpReduceBatchedReduceApiKernel(int* out)
+{
+  // example-begin warp-reduce-batched-reduce
+  // Can't allow for physical warp synchronization since only the first logical warp participates due to the
+  // conditional. The other threads (assuming there are more than 16 threads per block) can't exit early due to the
+  // barrier.
+  using WarpReduceBatched = cub::WarpReduceBatched<int, 3, 16>;
```
oleksandr-pavlyk, line 61: Ditto here.

cub/test/warp/catch2_test_warp_reduce_batched_api.cu
```diff
@@ -0,0 +1,339 @@
...
+
+  cuda::std::array<int, 5> thread_data{};
+  thread_data[0] = static_cast<int>(threadIdx.x) - 2;
+  thread_data[1] = static_cast<int>(threadIdx.x) - 1;
+  thread_data[2] = static_cast<int>(threadIdx.x);
+  thread_data[3] = static_cast<int>(threadIdx.x) + 1;
+  thread_data[4] = static_cast<int>(threadIdx.x) + 2;
+  // Use a static size spans to pick the first 3 elements as inputs and the last 2 elements as outputs (aliasing inputs)
+  cuda::std::span<int, 5> inputs{cuda::std::begin(thread_data), 5};
+  cuda::std::span<int, 3> results{cuda::std::begin(thread_data) + 2, 3};
```
oleksandr-pavlyk, line 268: Important: The comment describes the intent of picking last two elements, but `results` spans 3 elements in code. Please reconcile.

cub/test/warp/catch2_test_warp_reduce_batched_api.cu
```diff
@@ -0,0 +1,339 @@
...
+  __shared__ typename WarpReduceBatched::TempStorage temp_storage[4];
+
+  const int logical_warp_id = static_cast<int>(threadIdx.x) / 2;
+
+  cuda::std::array<int, 5> thread_data{};
+  thread_data[0] = static_cast<int>(threadIdx.x) - 2;
+  thread_data[1] = static_cast<int>(threadIdx.x) - 1;
+  thread_data[2] = static_cast<int>(threadIdx.x);
+  thread_data[3] = static_cast<int>(threadIdx.x) + 1;
+  thread_data[4] = static_cast<int>(threadIdx.x) + 2;
```
oleksandr-pavlyk, line 265: Assuming `constexpr int num_batches = 5`, could be shortened to

```c++
   const auto tid = static_cast<int>(threadIdx.x);
   cuda::std::array<int, num_batches> thread_data = {tid - 2, tid -1, tid, tid + 1, tid + 2};
```

cub/test/warp/catch2_test_warp_reduce_batched_api.cu
```diff
@@ -0,0 +1,339 @@
...
+  // Use a static size spans to pick the first 3 elements as inputs and the last 2 elements as outputs (aliasing inputs)
```
oleksandr-pavlyk, line 312: Same concern here, results is the span of 3 elements, but comment suggests output should be two elements. Did you mean `cuda::std::span<int, 2> results{cuda::std::begin(thread_data) + 3, 2};`?

pauleonix:
@oleksandr-pavlyk Will take care of these great points in a follow-up PR!

PR #7701, opened 2026-02-17
Provide `cuda::__ptr_alignment` (internal-only)

fbusato:
## Description

Retrieving the alignment of a pointer is a common utility in CCCL, as well as in other CUDA libraries.; see for example `cudax::hyperloglog` 

The PR adds `cuda::ptr_alignment(ptr, max_alignment)`  (`<cuda/memory>`) to get the alignment of a pointer (the largest power of two dividing its address), optionally capped by a maximum alignment.

### DIFF
--- a/docs/libcudacxx/extended_api/memory.rst
+++ b/docs/libcudacxx/extended_api/memory.rst
@@ -14,7 +14,6 @@ Memory
    memory/get_device_address
    memory/is_address_from
    memory/is_aligned
-   memory/ptr_alignment
    memory/ptr_rebind
    memory/ptr_in_range
    memory/ranges_overlap


libcudacxx/include/cuda/__memory/ptr_alignment.h
```diff
@@ -0,0 +1,59 @@
...
+
+//! @brief Get the alignment of a pointer, namely the largest power of two that divides the pointer address.
+//!
+//! @param  __ptr           the input pointer.
+//! @param  __max_alignment the maximum alignment to consider.
+//! @return The alignment of the pointer as a `size_t` value (always a power of two).
+//! @pre    __ptr is not null.
+//! @pre    __max_alignment is 0 or a power of two.
+[[nodiscard]] _CCCL_API inline ::cuda::std::size_t
+ptr_alignment(const void* __ptr, ::cuda::std::size_t __max_alignment = 0) noexcept
```
davebayer: I'm not sure if `__max_alignment = 0` is the right solution, the user would be able to pass `0`, too, which I think is not a good thing. I would prefer to have separate overloads:
```cpp
auto ptr_alignment(const void* __ptr);
auto ptr_alignment(const void* __ptr, ::cuda::std::size_t __max_alignment);
```

docs/libcudacxx/extended_api/memory/ptr_alignment.rst
```diff
@@ -0,0 +1,54 @@
...
+=======================
+
+Defined in the header ``<cuda/memory>``.
+
+.. code:: cuda
+
+   namespace cuda {
+
+   [[nodiscard]] __host__ __device__ inline
+   size_t ptr_alignment(const void* ptr, size_t max_alignment = 0) noexcept;
```
davebayer: I'm not sure if I understand the motivation of having this function. Why using `is_aligned` is not enough?

fbusato: suppose you have the following situation.
```cpp
auto align = ptr_alignment(ptr, 8);
switch (align) {
   case 1:
   case 2:
   ....
}

File: libcudacxx/include/cuda/__memory/ptr_alignment.h (hunk 1 above)
Comment (fbusato) [reply libcudacxx/include/cuda/__memory/ptr_alignment.h]: that's fine even if the benefit is very limited

File: docs/libcudacxx/extended_api/memory/ptr_alignment.rst (hunk 2 above)
Comment (davebayer) [reply docs/libcudacxx/extended_api/memory/ptr_alignment.rst]: I see, but is there enough motivation to publicly expose the function?

Comment (fbusato) [reply docs/libcudacxx/extended_api/memory/ptr_alignment.rst]: that's a good question. We could also use it only internally but, the situation described above is very common in CUDA. @bernhardmgruber, @miscco any thought?

Comment (bernhardmgruber) [reply docs/libcudacxx/extended_api/memory/ptr_alignment.rst]: I am not aware of a user requesting this function, so I would keep it internal. We may discover new uses that may mandate changing the API, so let's better not lock it down.

**Comment by fbusato** [comment]
alright. Thanks for the feedback. Made it internal.
Question: any thought about where we can put the documentation for the internal functionalities? just doxygen?

File: docs/libcudacxx/extended_api/memory/ptr_alignment.rst [hunk 3]
```diff
@@ -0,0 +1,57 @@
+.. _libcudacxx-extended-api-memory-ptr_alignment:
+
+``cuda::ptr_alignment``
```
davebayer: The docs should be removed

docs/libcudacxx/extended_api/memory.rst
```diff
@@ -14,6 +14,7 @@ Memory
    memory/get_device_address
    memory/is_address_from
    memory/is_aligned
+   memory/ptr_alignment
```
davebayer:
```suggestion
```

docs/libcudacxx/extended_api/memory.rst
```diff
@@ -53,6 +54,11 @@ Memory
      - CCCL 3.1.0
      - CUDA 13.1
 
+   * - :ref:`ptr_alignment <libcudacxx-extended-api-memory-ptr_alignment>`
+     - Get the alignment of a pointer
+     - CCCL 3.3.0
+     - CUDA 13.3
+
```
davebayer:
```suggestion
```

Issue #2242, opened 2024-08-14
[FEA]: (minor) formatting improvement


fbusato:
### Is this a duplicate?

- [X] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cccl/issues) for this request and that I agree to the [Code of Conduct](CODE_OF_CONDUCT.md)

### Area

General CCCL

### Is your feature request related to a problem? Please describe.

I noted that `.clang-format` misses the option `AlignConsecutiveAssignments` https://clang.llvm.org/docs/ClangFormatStyleOptions.html#alignconsecutiveassignments which greatly improves the readability of functions

Before
```
template <typename InputIteratorT,
          typename OutputIteratorT,
          typename NumSelectedIteratorT,
          typename OffsetT,
          typename SelectOp>
CUB_RUNTIME_FUNCTION _CCCL_FORCEINLINE static cudaError_t dispatch_select_if_wrapper(
  void* d_temp_storage,
  std::size_t& temp_storage_bytes,
  InputIteratorT d_in,
  OutputIteratorT d_out,
  NumSelectedIteratorT d_num_selected_out,
  OffsetT num_items,
  SelectOp select_op,
  cudaStream_t stream = 0)
{
```
after
```
template <typename InputIteratorT,
          typename OutputIteratorT,
          typename NumSelectedIteratorT,
          typename OffsetT,
          typename SelectOp>
CUB_RUNTIME_FUNCTION _CCCL_FORCEINLINE static cudaError_t dispatch_select_if_wrapper(
  void*                d_temp_storage,
  std::size_t&         temp_storage_bytes,
  InputIteratorT       d_in,
  OutputIteratorT      d_out,
  NumSelectedIteratorT d_num_selected_out,
  OffsetT              num_items,
  SelectOp             select_op,
  cudaStream_t         stream = 0)
{
```

### Describe the solution you'd like

add `AlignConsecutiveAssignments=Enabled` or similar options in `.clang-format`

bernhardmgruber:
If you plan to engage in challenging the current formatting, we may want to have a dedicated discussion. When I started in the team, I also discussed some formatting annoyances with @gevtushenko. Maybe it's time to have a general discussion and everybody can bring their wishes to the table.

bernhardmgruber:
> The point is **not** to promote _personal_ style preferences, but to improve readability and avoid potential bugs.

Unfortunately, readability can be very personal :D

PR #7704, opened 2026-02-18
`clang-format` Alignment for compound statements

fbusato:
Address https://github.com/NVIDIA/cccl/issues/2242

## Description

The PR adds two small improvements to the current code formatting:

- **Alignment for compound statements**, for example:
  ```cpp
  int x          = ...
  int long_name &= ...
  ```
- ~**Alignment for declarations**, for example~:

### DIFF
--- a/cub/cub/block/block_load_to_shared.cuh
+++ b/cub/cub/block/block_load_to_shared.cuh
@@ -160,8 +160,8 @@ private:
 
   _CCCL_DEVICE_API _CCCL_FORCEINLINE void __copy_aligned_async(char* smem_dst, const char* gmem_src, int num_bytes)
   {
-    for (int offset = linear_tid * detail::bulk_copy_min_align; offset < num_bytes;
-         offset += block_threads * detail::bulk_copy_min_align)
+    for (int offset  = linear_tid * detail::bulk_copy_min_align; offset < num_bytes;
+         offset     += block_threads * detail::bulk_copy_min_align)
     {
       [[maybe_unused]] const auto thread_src = gmem_src + offset;
       [[maybe_unused]] const auto thread_dst = smem_dst + offset;
@@ -176,8 +176,8 @@ private:
 
   _CCCL_DEVICE_API _CCCL_FORCEINLINE void __copy_aligned_fallback(char* smem_dst, const char* gmem_src, int num_bytes)
   {
-    for (int offset = linear_tid * detail::bulk_copy_min_align; offset < num_bytes;
-         offset += block_threads * detail::bulk_copy_min_align)
+    for (int offset  = linear_tid * detail::bulk_copy_min_align; offset < num_bytes;
+         offset     += block_threads * detail::bulk_copy_min_align)
     {
       const auto thread_src                       = gmem_src + offset;
       const auto thread_dst                       = smem_dst + offset;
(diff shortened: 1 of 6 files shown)


miscco:
Honestly I am also not too enthusiastic, but happy to follow consensus

fbusato:
> IMHO this doesn't improve the readability and makes it even worse..

🙁 I had the opposite impression. Maybe the subset that the PR shows in not very representative. Here I see a big difference https://github.com/NVIDIA/cccl/issues/2242.
Anyway, the idea is to get feedback and collect preferences.

davebayer:
What I don't like is that it does this:
```cpp
// 1. Good case
type1   a = 10;
longer1 b = 23;

// 2. short & long type name combination
short                       a = 10;
some_long_type_on_next_line b = 27;

// 3. short & long type name with long & short var name combination
short                       some_longer_variable_name = 10;
some_long_type_on_next_line b                         = 27;
```
I think `2.` is acceptable (but not better), but `3.` is just terrible and I'm afraid we would be getting a lot of them.

And I hope it was obvious from my previous review by not requiring changes, but I will happily follow the consensus, too :)

fbusato:
I definitely agree with your argument. My idea was to apply it only for functions. It degenerates very easy with long names. Unfortunately, it is not possible to provide penalty or knobs :(

fbusato:
We are getting very lucky https://github.com/llvm/llvm-project/issues/54220

fbusato:
removed **Alignment for declarations**. I will add it again when https://github.com/llvm/llvm-project/issues/54220 is included in an official LLVM release.

bernhardmgruber:
I am weakly against the formatting changes and will therefore not approve this PR. I personally think they don't improve readability. But if some other people think it's useful, I will not oppose.

fbusato:
@bernhardmgruber I removed the declaration alignment. Now the PR only includes the compound statements. We align assigned, I don't see any reason why we should not also align compound statements. Please take a look again.

bernhardmgruber:
> @bernhardmgruber I removed the declaration alignment. Now the PR only includes the compound statements. We align assigned, I don't see any reason why we should not also align compound statements. Please take a look again.

I believe I did look at the most recent changes, and I am still not convinced. But as I said, I am fine if anybody else is.

cub/cub/agent/agent_radix_sort_upsweep.cuh
```diff
@@ -421,8 +421,8 @@ struct AgentRadixSortUpsweep
 
     // Whole blocks
     _CCCL_PRAGMA_UNROLL_FULL()
-    for (int BIN_BASE = RADIX_DIGITS % BLOCK_THREADS; (BIN_BASE + BLOCK_THREADS) <= RADIX_DIGITS;
-         BIN_BASE += BLOCK_THREADS)
+    for (int BIN_BASE  = RADIX_DIGITS % BLOCK_THREADS; (BIN_BASE + BLOCK_THREADS) <= RADIX_DIGITS;
+         BIN_BASE     += BLOCK_THREADS)
```
miscco, line 425: That one is just funny :)

cub/cub/detail/rfa.cuh
```diff
@@ -324,29 +324,29 @@ private:
...
+      M           = primary(0);
+      ftype qd    = x * compression;
+      auto& ql    = get_bit_representation(qd);
+      ql         |= 1;
+      qd         += M;
+      primary(0)  = qd;
+      M          -= qd;
+      M          *= expansion * 0.5;
+      x          += M;
+      x          += M;
```
davebayer, line 336: Ugh, this is bad

PR #7713, opened 2026-02-18
Refactor cuda.coop to use maker factory functions.

tpn:
This PR refactors cuda.coop so that primitives are created outside the kernel via a function with the prefix `make_`.  This is in preparation for the upcoming 1.0 release, as well as setting the stage for a seamless single-phase implementation post 1.0 release.

### DIFF
--- a/python/cuda_cccl/cuda/coop/warp/_warp_reduce.py
+++ b/python/cuda_cccl/cuda/coop/warp/_warp_reduce.py
@@ -18,15 +18,18 @@ from .._types import (
 
 
 def make_reduce(dtype, binary_op, threads_in_warp=32, methods=None):
-    """Computes a warp-wide reduction for lane\ :sub:`0` using the specified binary reduction functor.
-    Each thread contributes one input element.
+    """
+    Creates a warp-wide reduction primitive for lane :sub:`0` using the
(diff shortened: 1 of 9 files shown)


python/cuda_cccl/cuda/coop/warp/_warp_reduce.py
```diff
@@ -34,7 +34,7 @@ def reduce(dtype, binary_op, threads_in_warp=32, methods=None):
             :start-after: example-begin imports
             :end-before: example-end imports
 
-        Below is the code snippet that demonstrates the usage of the ``reduce`` API:
+        Below is the code snippet that demonstrates the usage of the ``make_sum`` API:
```
shwina: Small mix up here maybe? (we reference `make_sum` in the docstring of `make_reduce`, and then reference `make_reduce` in the docstring of `make_sum`)

tpn: ~~Dang I thought I caught all of those.  Fixing...~~

Were you looking at the first commit or the changes overall?  The second commit should have fixed all these docstrings.

shwina: Ah gotcha - you're right. Apologies.

Issue #5684, opened 2025-08-27
`cuda.cccl.parallel`: Add `shuffle_iterator`


leofang:
We have `shuffle_iterator` in both Thrust and libcudacxx and it is pretty powerful. I'd like to use them in Python.

leofang:
#5679 is also relevant because the shuffle iterator needs to draw some random numbers, and at least in the case of `thrust::shuffle_iterator` we need to use one of the random number engines from Thrust.

PR #7721, opened 2026-02-19
cuda.compute: Add `ShuffleIterator` (redux)

shwina:
## Description

Closes https://github.com/NVIDIA/cccl/issues/5684.

Following the [refactor of iterators](https://github.com/NVIDIA/cccl/pull/7441/) to use C++, this PR implements `ShuffleIterator` using the new approach, superseding https://github.com/NVIDIA/cccl/pull/7062.

Like `thrust::shuffle_iterator`, the implementation uses `cuda::random_bijection`. _Unlike_ `thrust::shuffle_iterator`, it uses `cuda::pcg64` for the RNG, rather than `minstd_rand`. @RAMitchell to validate that this is not a bad idea :) 

By virtue of implementing the iterators in C++, we can reuse the relevant pieces from `libcudacxx`, unlike in #7062 where we basically reimplemented the algorithm.

### Validation

Using the same validation scripts as in the [previous PR](https://github.com/NVIDIA/cccl/pull/7062#issuecomment-3712125756), I can confirm that we get statistically uniform distributions (`p > 0.05`):

validation

## Checklist

### DIFF
--- a/python/cuda_cccl/cuda/compute/iterators/_shuffle.py
+++ b/python/cuda_cccl/cuda/compute/iterators/_shuffle.py
@@ -95,6 +95,13 @@ class ShuffleIterator(IteratorBase):
     def _make_input_deref_op(self) -> Op | None:
         symbol = self._make_input_deref_symbol()
 
+        # Note: a potential optimization is to avoid constructing
+        # `cuda::random_bijection` objects upon every dereference,
+        # instead constructing it once and using it as the state
+        # object. The tradeoff is that it would require a C++
+        # extension providing a constructor for
+        # `cuda::random_bijection` objects, since we would now be
+        # doing it on the host.  See discussion in #7721.
         source = dedent(f"""
             #include <cuda/__random/random_bijection.h>
             #include <cuda/__random/pcg_engine.h>


python/cuda_cccl/cuda/compute/iterators/_shuffle.py
```diff
@@ -0,0 +1,133 @@
...
+        """).strip()
+
+        return Op(
+            operator_type=OpKind.STATELESS,
+            name=symbol,
+            ltoir=compile_cpp_to_ltoir(source),
+            extra_ltoirs=[],
+        )
+
+    def _make_input_deref_op(self) -> Op | None:
```
shwina, line 95: @RAMitchell in particular, could you review the dereference method here? Thank you.

The `ShuffleState` type is defined above near line 20.

RAMitchell: LGTM

python/cuda_cccl/cuda/compute/iterators/_shuffle.py
```diff
@@ -0,0 +1,139 @@
...
+    def _make_advance_op(self) -> Op:
+        symbol = self._make_advance_symbol()
+
+        source = dedent(f"""
+            {CUDA_PREAMBLE}
+
+            {_SHUFFLE_STATE_STRUCT}
+
+            extern "C" __device__ void {symbol}(void* state, void* offset) {{
+                auto* s = static_cast<ShuffleState*>(state);
```
oleksandr-pavlyk:
```suggestion
                auto s = static_cast<ShuffleState*>(state);
```

python/cuda_cccl/cuda/compute/iterators/_shuffle.py
```diff
@@ -0,0 +1,139 @@
...
+                auto dist = *static_cast<uint64_t*>(offset);
+                s->current_index += static_cast<int64_t>(dist);
```
oleksandr-pavlyk: Can we do the following instead?

```suggestion
                auto dist = *static_cast<int64_t*>(offset);
                s->current_index += dist;
```

python/cuda_cccl/cuda/compute/iterators/_shuffle.py
```diff
@@ -0,0 +1,139 @@
...
+            }}
+        """).strip()
```
oleksandr-pavlyk: Is the purpose of using `strip()` to make debug prints nicer? Should `.strip()` be removed before merging?

python/cuda_cccl/cuda/compute/iterators/_shuffle.py
```diff
@@ -0,0 +1,139 @@
...
+
+        return Op(
+            operator_type=OpKind.STATELESS,
```
oleksandr-pavlyk, line 89: Did you mean `OpKind.STATEFUL` instead?

python/cuda_cccl/cuda/compute/iterators/_shuffle.py
```diff
@@ -0,0 +1,139 @@
...
+
+            extern "C" __device__ void {symbol}(void* state, void* result) {{
+                const auto* s = static_cast<const ShuffleState*>(state);
+                *static_cast<int64_t*>(result) = __shuffle_apply(
+                    s->num_items, s->seed, static_cast<uint64_t>(s->current_index));
+            }}
+        """).strip()
+
+        return Op(
+            operator_type=OpKind.STATELESS,
```
oleksandr-pavlyk, line 132: Did you mean `OpKind.STATEFUL` here?

python/cuda_cccl/cuda/compute/iterators/_shuffle.py
shwina: No -- iterator advance and dereference methods are stateless. The ops themselves hold no state. The state is held by the iterator itself and passed as an argument to these methods. 

See how we define `advance` and `dereference` for the C++ tests:

https://github.com/NVIDIA/cccl/blob/788fe7cc59b9b2d56b75e463a8121bae4c4e0297/c/parallel/test/test_util.h#L925-L926

Those calls to `make_operation` in turn call https://github.com/NVIDIA/cccl/blob/788fe7cc59b9b2d56b75e463a8121bae4c4e0297/c/parallel/test/test_util.h#L808-L810 (which creates stateless ops).

python/cuda_cccl/cuda/compute/iterators/_shuffle.py
shwina: Yes, I suppose we don't need it. Removed.

oleksandr-pavlyk (approved):

python/cuda_cccl/cuda/compute/iterators/_shuffle.py
djns99: Would it be better to make the round keys (rather than the seed) part of the state so we don't have to regenerate them each dereference?
Not sure a if theres a good way to implement this while reusing the existing implementation, unless `cuda::random_bijection` itself is trivially copyable and can just be the contents of the state object?

python/cuda_cccl/cuda/compute/iterators/_shuffle.py
```diff
@@ -0,0 +1,139 @@
...
+        self._seed = int(seed)
+        self._current_index = int(_current_index)
+
+        # State layout matches C++ ShuffleState:
+        #   int64_t  current_index  (offset 0,  size 8)
+        #   uint64_t num_items      (offset 8,  size 8)
+        #   uint64_t seed           (offset 16, size 8)
+        state_bytes = struct.pack(
+            "<qQQ", self._current_index, self._num_items, self._seed
+        )
```
leofang, line 65: For my own curiosity: IIRC you mentioned cuda.compute has moved away from using NumPy struct dtypes. Is there any functional difference?

python/cuda_cccl/cuda/compute/iterators/_shuffle.py
```diff
@@ -0,0 +1,139 @@
...
+            {_SHUFFLE_STATE_STRUCT}
+
+            // __noinline__ is required to prevent the compiler from merging
+            // this function's register usage into the calling kernel during LTO
+            // inlining. feistel_bijection constructs 24 round keys with
+            // UNROLL_FULL, which exhausts the kernel's register budget and
+            // causes spilling to local memory (LDL/STL instructions).
+            // Keeping it non-inlined gives it an isolated register frame.
+            __device__ __noinline__ int64_t __shuffle_apply(uint64_t num_items, uint64_t seed, uint64_t idx) {{
+                cuda::pcg64 rng(seed);
```
leofang, line 119: My wishlist is we could bring in our own RNG, but at this point I am asking too much! :D

python/cuda_cccl/cuda/compute/iterators/_shuffle.py
leofang: I think it is trivially copyable and the needed keys are finite:
https://github.com/NVIDIA/cccl/blob/788fe7cc59b9b2d56b75e463a8121bae4c4e0297/libcudacxx/include/cuda/__random/feistel_bijection.h#L44
One way to do this is to initialize the host state in `__init__` and make it part of `state_bytes`, but I dunno whether this would help improve perf.

shwina: Hmm, this would essentially need us to build a tiny C++ extension for Python, providing a constructor for `cuda::random_bijection` objects _or_ to replicate the construction logic using `ctypes`.

I'll leave a comment in the code that this is a potential optimization we can do, but I'd rather only deal with the complexity if absolutely necessary.

python/cuda_cccl/cuda/compute/iterators/_shuffle.py
shwina: Hmm, I don't think it was in this context. We have always used ctypes for managing iterator states (not NumPy).

We made an attempt to abandon the use of `@gpu_struct` for letting the user define struct types, in favor of just NumPy structs, but that didn't work out. That's a totally separate thing from this though.

leofang: Yes I recall now: https://github.com/NVIDIA/cccl/issues/5450, thanks!

python/cuda_cccl/cuda/compute/iterators/_shuffle.py
leofang: I think a host C++ extension is not necessary; the same `struct.pack()` trick can be extended to cover the keys too. But it is fine to punt on this.

shwina: > _or_ to replicate the construction logic using ctypes.

Without an extension, we'd still have this problem.

PR #7724, opened 2026-02-20
[CUB] Move helper functions out of `BlockLoadToShared`

pauleonix:
## Description
Given that the minimum alignment constant and the buffer size and alignment helper functions could be used in a potential future `WarpLoadToShared`, @elstehle and I decided to move them out from the class (helpers were static member functions) and into existing utility headers (feedback on the choice of headers is welcome). This also avoids the `primitive_t::template Foo<T>` uglyness in dependent contexts.

While the minimum alignment constant can remain in `cub::detail`, the helper functions should become public (moved from `cub::detail` into `cub`) in a future PR once we are happy with the interface.

## Checklist

### DIFF
--- a/cub/cub/util_device.cuh
+++ b/cub/cub/util_device.cuh
@@ -511,44 +511,6 @@ MaxPotentialDynamicSmemBytes(int& max_dyn_smem_bytes, KernelPtr kernel_ptr) noex
 
 namespace detail
 {
-// This should stay an implementation detail even when below functions become public.
-static constexpr int bulk_copy_min_align = 16;
-
-//! @brief Returns the alignment needed for the shared memory destination buffer of BlockLoadToShared.
-//! @tparam T
-//!   Value type to be loaded.
-template <typename T>
-_CCCL_HOST_DEVICE constexpr int LoadToSharedBufferAlignBytes()
-{
-  return (::cuda::std::max) (int{alignof(T)}, detail::bulk_copy_min_align);
-}
-
-//! @brief Returns the size needed for the shared memory destination buffer of BlockLoadToShared.
-//! @tparam T
-//!   Value type to be loaded.
-//! @tparam GmemAlign
-//!   Guaranteed alignment in bytes of the source range (both begin and end) in global memory
-//! @param[in] num_items
-//!   Size of the source range in global memory
-template <typename T, int GmemAlign = alignof(T)>
-_CCCL_HOST_DEVICE constexpr int LoadToSharedBufferSizeBytes(::cuda::std::size_t num_items)
-{
-  static_assert(::cuda::std::has_single_bit(unsigned{GmemAlign}));
-  static_assert(GmemAlign >= int{alignof(T)});
-  _CCCL_ASSERT(num_items <= ::cuda::std::size_t{::cuda::std::numeric_limits<int>::max()},
-               "num_items must fit into an int");
-  const int num_bytes = static_cast<int>(num_items) * int{sizeof(T)};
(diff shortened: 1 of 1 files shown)


cub/cub/util_arch.cuh
```diff
@@ -115,6 +115,8 @@ inline constexpr bool prefer_conflict_over_padding = CUB_PTX_PREFER_CONFLICT_OVE
 // shared memory in a kernel. Note that dynamic shared memory may be larger than this amount.
 static constexpr ::cuda::std::size_t max_smem_per_block = 48 * 1024;
 
+static constexpr int bulk_min_align = 16;
```
pauleonix: Put into `util_device.cuh` together with the helper functions to avoid including `util_arch.cuh` into it (compile time).

cub/cub/block/block_load_to_shared.cuh
```diff
@@ -234,26 +237,26 @@ public:
   //!
   //! @param[in] temp_storage
   //!   Reference to memory allocation having layout type TempStorage
-  _CCCL_DEVICE _CCCL_FORCEINLINE BlockLoadToShared(TempStorage& temp_storage)
+  _CCCL_DEVICE_API _CCCL_FORCEINLINE BlockLoadToShared(TempStorage& temp_storage)
```
pauleonix, line 240: I think this is only needed for default constructors (of which there are none anymore)? @miscco

cub/cub/block/block_load_to_shared.cuh
```diff
@@ -46,31 +47,31 @@ namespace detail
...
 {
 private:
   /// Constants
   static constexpr int block_threads = BlockDimX * BlockDimY * BlockDimZ;
-  // The alignment needed for cp.async.bulk and L1-skipping cp.async
-  static constexpr int minimum_align = 16;
 
   // Helper for fallback to gmem->reg->smem
-  struct alignas(minimum_align) vec_load_t
+  struct alignas(cub::detail::bulk_min_align) vec_load_t
```
bernhardmgruber: Suggestion:
```suggestion
  struct alignas(detail::bulk_min_align) vec_load_t
```
Applies multiple times.

cub/cub/block/block_load_to_shared.cuh
bernhardmgruber: You should use `_CCCL_DEVICE_API` liberally everywhere.

cub/cub/util_arch.cuh
bernhardmgruber: Suggestion: use a name closer to bulk copying or tma. E.g.:
```suggestion
static constexpr int bulk_copy_min_align = 16;
```

bernhardmgruber (approved):

cub/cub/block/block_load_to_shared.cuh
pauleonix: Ok, could we get this captured in the Coding Guidelines?

pauleonix: And does "everywhere" include private member functions (not part of the `API`)? I have applied it literally everywhere in this file now.

bernhardmgruber: > Ok, could we get this captured in the Coding Guidelines?

Already did that yesterday :)

bernhardmgruber: > And does "everywhere" include private member functions (not part of the `API`)? I have applied it literally everywhere in this file now.

Yes, I think it's intended to be used everywhere.

PR #7725, opened 2026-02-20
Replace `find_package(CUDAToolkit)` with `cccl_get_cudatoolkit()`

davebayer:
We have the `cccl_get_cudatoolkit()` macro that I believe should always be used instead of `find_package(CUDAToolkit)`.

### DIFF
--- a/c2h/CMakeLists.txt
+++ b/c2h/CMakeLists.txt
@@ -2,24 +2,31 @@ cmake_minimum_required(VERSION 3.21)
 
 project(C2H LANGUAGES CXX CUDA)
 
-cccl_get_catch2()
-
 # when CCCL is used from the CTK, C2H does not need to be rebuilt when making local changes to CCCL (faster iteration)
 option(C2H_USE_CCCL_FROM_CTK "Use CCCL from the CTK in c2h." OFF)
 
-if (C2H_USE_CCCL_FROM_CTK)
-  find_package(CUDAToolkit REQUIRED)
-else()
+# Get dependencies.
+cccl_get_catch2()
+if (NOT C2H_USE_CCCL_FROM_CTK)
   cccl_get_cccl()
-  find_package(CUDAToolkit)
+endif()
+cccl_get_cudatoolkit()
+
+set(has_curand OFF)
+if (TARGET CUDA::curand)
+  set(has_curand ON)
 endif()
 
-set(curand_default OFF)
-if (CUDA_curand_LIBRARY)
-  set(curand_default ON)
-endif()
+option(C2H_ENABLE_CURAND "Use CUDA CURAND library in c2h." ${has_curand})
 
-option(C2H_ENABLE_CURAND "Use CUDA CURAND library in c2h." ${curand_default})
+# Disable C2H_ENABLE_CURAND if cuRAND is unavailable.
+if (C2H_ENABLE_CURAND AND NOT has_curand)
+  message(
+    WARNING
+    "C2H_ENABLE_CURAND is requested, but CUDA::curand is unavailable. Disabling."
+  )
+  set(C2H_ENABLE_CURAND OFF)
+endif()
 
 add_library(
   cccl.c2h
--- a/cmake/CCCLCheckCudaArchitectures.cmake
+++ b/cmake/CCCLCheckCudaArchitectures.cmake
@@ -53,6 +53,7 @@ endfunction()
 
 # Query nvcc --help to determine which architectures are supported.
 function(_cccl_detect_nvcc_arch_support arches_var)
+  # cccl_get_cudatoolkit() is intentionally not used here.
   find_package(CUDAToolkit)
   if (NOT CUDAToolkit_FOUND)
     message(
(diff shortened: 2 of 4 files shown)


c2h/CMakeLists.txt
```diff
@@ -2,25 +2,22 @@ cmake_minimum_required(VERSION 3.21)
...
+# Get dependencies.
+cccl_get_catch2()
+if (NOT C2H_USE_CCCL_FROM_CTK)
   cccl_get_cccl()
-  find_package(CUDAToolkit)
 endif()
+cccl_get_cudatoolkit()
 
-set(curand_default OFF)
-if (CUDA_curand_LIBRARY)
```
davebayer, line 18: This variable is defined when the deprecated `find_package(CUDA)` is used. So, we were never using cuRAND by default.

libcudacxx/cmake/LibcudacxxInternalHeaderTesting.cmake
```diff
@@ -43,12 +39,6 @@ list(FILTER internal_headers EXCLUDE REGEX "__cuda/*")
 # generated cuda::ptx headers are not standalone
 list(FILTER internal_headers EXCLUDE REGEX "__ptx/instructions/generated")
 
-if ("NVHPC" STREQUAL "${CMAKE_CXX_COMPILER_ID}")
-  set(cudart_name NVHPC::CUDART)
-else()
-  set(cudart_name CUDA::cudart)
-endif()
```
davebayer, line 50: I believe we should be using the cuda runtime library that comes with the CUDA compiler, not NVHPC Toolkit. If the CUDA Toolkit and `nvcc` that comes with NVHPC Toolkit is used, CMake should correctly deduce that the `libcudart.so` from NVHPC Toolkit should be used.

libcudacxx/cmake/LibcudacxxPublicHeaderTestingHost.cmake
```diff
@@ -4,6 +4,8 @@
 # .inl files are not globbed for, because they are not supposed to be used as public
 # entrypoints.
 
+cccl_get_cudatoolkit()
```
davebayer, line 7: This dependency was missing.

c2h/CMakeLists.txt
```diff
@@ -2,25 +2,22 @@ cmake_minimum_required(VERSION 3.21)
...
-  set(curand_default ON)
+# Disable C2H_ENABLE_CURAND if cuRAND is unavailable.
+if (NOT TARGET CUDA::curand)
+  set(C2H_ENABLE_CURAND OFF)
```
alliepiper, line 28: This will lead to behavior where the user may enable curand, but then we silently build without it.


```suggestion
# Disable C2H_ENABLE_CURAND if cuRAND is unavailable.
if (C2H_ENABLE_CURAND AND NOT TARGET CUDA::curand)
  message(WARNING "C2H_ENABLE_CURAND is requested, but CUDA::curand is unavailable. Disabling.")
  set(C2H_ENABLE_CURAND OFF)
```

libcudacxx/cmake/LibcudacxxInternalHeaderTesting.cmake
alliepiper: Agreed -- we just use the CUDA:: target unconditionally everwhere else and haven't had any issues, so unless the libcu++ header tests are somehow special, this should be safe.

CMakeLists.txt
```diff
@@ -24,6 +24,7 @@ endif()
 project(CCCL LANGUAGES CXX)
 
 include(cmake/CCCLInstallRules.cmake)
+include(cmake/CCCLGetDependencies.cmake)
```
alliepiper: This was intentionally guarded behind TOPLEVEL checks to avoid polluting upstream projects.

```suggestion
```

CMakeLists.txt
```diff
@@ -92,7 +93,6 @@ if (CCCL_TOPLEVEL_PROJECT)
   include(cmake/CCCLConfigureTarget.cmake)
   include(cmake/CCCLEnsureMetaTargets.cmake)
   include(cmake/CCCLGenerateHeaderTests.cmake)
```
alliepiper, line 94:
```suggestion
  include(cmake/CCCLGenerateHeaderTests.cmake)
  include(cmake/CCCLGetDependencies.cmake)
```

cmake/CCCLCheckCudaArchitectures.cmake
```diff
@@ -53,7 +53,7 @@ endfunction()
 
 # Query nvcc --help to determine which architectures are supported.
 function(_cccl_detect_nvcc_arch_support arches_var)
-  find_package(CUDAToolkit)
+  cccl_get_cudatoolkit()
```
alliepiper: This was intentional, since this file must be included early in the config, before we setup most of the cccl utils. We should add a comment:


```suggestion
  # cccl_get_cudatoolkit() is intentionally unavailable here.
  find_package(CUDAToolkit)
```

PR #7731, opened 2026-02-21
Rename the docs header for the C++ libraries from "CUDA C++ Core Libraries" to "CCCL C++ Libraries"

brycelelbach:
Here's what the docs landing page looks like right now:

- Title: "CUDA Core Compute Libraries (CCCL)".
- Section: "CUDA C++ Core Libraries".
- Section: "CCCL Python Libraries".

This is inconsistent and makes it unclear what CCCL stands for. I mean, okay, the ambiguity *WAS* the reason I chose the name, but now it just means CUDA Core Compute Libraries.

### DIFF
--- a/docs/cpp.rst
+++ b/docs/cpp.rst
@@ -1,7 +1,7 @@
 .. _cccl-cpp-libraries:
 
-CUDA C++ Core Libraries
-=======================
+CCCL C++ Libraries
+==================
 
 .. toctree::
    :hidden:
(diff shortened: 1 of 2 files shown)


bernhardmgruber (commented):
Fine for me, but I would also want @jrhemstad's opinion.

docs/cpp.rst
```diff
@@ -1,7 +1,7 @@
 .. _cccl-cpp-libraries:
 
-CUDA C++ Core Libraries
-=======================
+CCCL C++ Libraries
```
miscco, line 3: No a fan, because this is libraries libraries

I would much rather have `CCCL C++` and `CCCL Python`

bernhardmgruber:
This change was applied as a drive-by in #11113

PR #7733, opened 2026-02-22
Explain SASS diffs in `AGENTS.md`

bernhardmgruber:
I used this a few times now to successfully ask Cursor to find the cause of SASS differences and fix them. I have no idea what good practices are for how to describe such stuff in AGENTS.md, so please let me know if this can be improved in any way.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -273,6 +273,54 @@ Test organization:
 
 ---
 
+## SASS Diffs
+
+Use this test when asked to check for SASS changes between commits, branches or a local changeset.
+
+### Goal
+
+Detect relevant changes in generated CUDA machine code (i.e. SASS) while filtering noise from addresses, symbols, metadata, etc.
+Any non-trivial change must be detected.
+
+### Inputs to establish
+
+* Compiled binary under test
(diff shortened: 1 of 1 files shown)


sleeepyjack:
Sorry, I'm late to the party. I feel like this is something that should be exposed as a separate skill in SKILLS.md like so:

```
Skill: CUDA Codegen SASS Verification

Description:
Compare SASS output before/after CUDA changes to validate performance stability.

Inputs:
- Baseline CUDA source
- Modified CUDA source
- Target architecture

Process:
1. Compile with nvcc -arch=...
2. Extract SASS via cuobjdump
3. Normalize output
4. Diff instruction classes
5. Report deltas
6. Flag register or occupancy regressions

Constraints:
- Must use same compiler flags
```

This way is slightly more structured as you can instruct an agent to specifically use that skill rather than letting it memoize the instructions from the AGENTS.md file when you instruct it to do a SASS comparison.

fbusato:
@sleeepyjack I raised the same issue. This is something that should definitely belong to a `SKILL.md`

bernhardmgruber:
@sleeepyjack thanks for sketching how this can be done. I opened an issue for this: #7829

PR #7735, opened 2026-02-22
Freestanding

gevtushenko:
## Description

closes https://github.com/NVIDIA/cccl_private/issues/622

##  Motivation

Host JIT compilation for CUB (e.g., `cub::DeviceReduce`) currently requires users to have system headers installed. This PR introduces a freestanding mode so that CUB device algorithms can be JIT-compiled without depending on any host standard library headers.

## What this PR does

Following the C++ standard [terminology](https://en.cppreference.com/w/cpp/freestanding.html), this PR introduces two new macros:

- `_CCCL_FREESTANDING()` — evaluates to 1 when compiling in a freestanding environment (no host standard library available)
- `_CCCL_HOSTED()` — evaluates to 1 when compiling in a hosted environment (inverse of freestanding)

Freestanding mode is activated when either:
- User defines `_CCCL_ENABLE_FREESTANDING`, or
- NVRTC compiler is detected (which has no access to host headers)

The `<cmath>` headers rely on too many global symbols. To reduce the amount of changes, this PR also adds `_CCCL_DISABLE_CMATH`, which disables most of them. Code paths that depend on <cmath> (e.g., RFA-based deterministic reduction) are disabled in freestanding mode.

## Checklist

### DIFF
--- a/libcudacxx/include/cuda/std/__string/constexpr_c_functions.h
+++ b/libcudacxx/include/cuda/std/__string/constexpr_c_functions.h
@@ -47,21 +47,23 @@ __cccl_strcpy_impl_constexpr(_CharT* _CCCL_RESTRICT __dst, const _CharT* _CCCL_R
   return __dst;
 }
 
+#if !_CCCL_COMPILER(NVRTC)
 template <class _CharT>
 _CCCL_HIDE_FROM_ABI _CCCL_HOST _CharT*
 __cccl_strcpy_impl_host(_CharT* _CCCL_RESTRICT __dst, const _CharT* _CCCL_RESTRICT __src) noexcept
 {
-#if _CCCL_HOSTED()
+#  if _CCCL_HOSTED()
   if constexpr (sizeof(_CharT) == 1)
   {
     return reinterpret_cast<_CharT*>(::strcpy(reinterpret_cast<char*>(__dst), reinterpret_cast<const char*>(__src)));
   }
   else
-#endif // _CCCL_HOSTED()
+#  endif // _CCCL_HOSTED()
   {
     return ::cuda::std::__cccl_strcpy_impl_constexpr(__dst, __src);
   }
 }
+#endif // !_CCCL_COMPILER(NVRTC)
 
 template <class _CharT>
 _CCCL_API constexpr _CharT* __cccl_strcpy(_CharT* _CCCL_RESTRICT __dst, const _CharT* _CCCL_RESTRICT __src) noexcept
@@ -94,22 +96,24 @@ __cccl_strncpy_impl_constexpr(_CharT* _CCCL_RESTRICT __dst, const _CharT* _CCCL_
   return __dst;
 }
 
+#if !_CCCL_COMPILER(NVRTC)
 template <class _CharT>
 _CCCL_HIDE_FROM_ABI _CCCL_HOST _CharT*
 __cccl_strncpy_impl_host(_CharT* _CCCL_RESTRICT __dst, const _CharT* _CCCL_RESTRICT __src, size_t __n) noexcept
 {
-#if _CCCL_HOSTED()
+#  if _CCCL_HOSTED()
   if constexpr (sizeof(_CharT) == 1)
   {
     return reinterpret_cast<_CharT*>(
       ::strncpy(reinterpret_cast<char*>(__dst), reinterpret_cast<const char*>(__src), __n));
   }
   else
-#endif // _CCCL_HOSTED()
+#  endif // _CCCL_HOSTED()
   {
     return ::cuda::std::__cccl_strncpy_impl_constexpr(__dst, __src, __n);
   }
 }
+#endif // !_CCCL_COMPILER(NVRTC)
 
 template <class _CharT>
 _CCCL_API constexpr _CharT*
@@ -135,20 +139,22 @@ template <class _CharT>
   return __len;
 }
 
+#if !_CCCL_COMPILER(NVRTC)
 template <class _CharT>
 [[nodiscard]] _CCCL_HIDE_FROM_ABI _CCCL_HOST size_t __cccl_strlen_impl_host(const _CharT* __ptr) noexcept
 {
-#if _CCCL_HOSTED()
+#  if _CCCL_HOSTED()
   if constexpr (sizeof(_CharT) == 1)
   {
     return ::strlen(reinterpret_cast<const char*>(__ptr));
   }
   else
-#endif // _CCCL_HOSTED()
+#  endif // _CCCL_HOSTED()
   {
     return ::cuda::std::__cccl_strlen_impl_constexpr(__ptr);
   }
 }
+#endif // !_CCCL_COMPILER(NVRTC)
 
 template <class _CharT>
 [[nodiscard]] _CCCL_API constexpr size_t __cccl_strlen(const _CharT* __ptr) noexcept
@@ -180,21 +186,23 @@ template <class _CharT>
   return (static_cast<_UCharT>(*__lhs) < static_cast<_UCharT>(*__rhs)) ? -1 : 1;
 }
 
+#if !_CCCL_COMPILER(NVRTC)
 template <class _CharT>
 [[nodiscard]] _CCCL_HIDE_FROM_ABI _CCCL_HOST int
 __cccl_strcmp_impl_host(const _CharT* __lhs, const _CharT* __rhs) noexcept
 {
-#if _CCCL_HOSTED()
+#  if _CCCL_HOSTED()
   if constexpr (sizeof(_CharT) == 1)
   {
     return ::strcmp(reinterpret_cast<const char*>(__lhs), reinterpret_cast<const char*>(__rhs));
   }
   else
-#endif // _CCCL_HOSTED()
+#  endif // _CCCL_HOSTED()
   {
     return ::cuda::std::__cccl_strcmp_impl_constexpr(__lhs, __rhs);
   }
 }
+#endif // !_CCCL_COMPILER(NVRTC)
 
 template <class _CharT>
 [[nodiscard]] _CCCL_API constexpr int __cccl_strcmp(const _CharT* __lhs, const _CharT* __rhs) noexcept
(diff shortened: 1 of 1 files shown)


davebayer:
Can't we rather `_CCCL_VERIFY(false, "Host implementation requires hosted evironment.")` in all of those cmath functions? That way we could still use them in `__device__` code.

Can we split this into multiple smaller PRs? This PR is for example changing how `cuda::devices` data are initialized and stored, which is something I'd like to see in a separate PR.

Also, I don't see the freestanding mode tested. I'd like to have at least public header testing for this mode.

cub/cub/agent/agent_radix_sort_onesweep.cuh
```diff
@@ -34,9 +34,9 @@
 #include <cuda/std/__type_traits/integral_constant.h>
 #include <cuda/std/__type_traits/is_same.h>
 
-#if !_CCCL_COMPILER(NVRTC)
+#if _CCCL_HOSTED()
```
miscco, line 37: I am really no fan of `_CCCL_HOSTED`, I would much rather have `!_CCCL_FREESTANDING` because that is a much better known term

cub/cub/detail/warpspeed/constant_assert.cuh
```diff
@@ -12,7 +12,7 @@
 #  pragma system_header
 #endif // no system header
 
-#include <cuda/std/cassert>
+#include <cuda/std/__cccl/assert.h>
```
miscco, line 15: This should not be necessary we *must* be able to include all `<cuda/std/...>` headers even in freestanding

cub/cub/detail/rfa.cuh
```diff
@@ -13,17 +13,19 @@
...
-#include <cuda/std/__bit/bit_cast.h>
-#include <cuda/std/__cmath/exponential_functions.h>
-#include <cuda/std/__cmath/isinf.h>
-#include <cuda/std/__type_traits/enable_if.h>
-#include <cuda/std/__type_traits/is_arithmetic.h>
-#include <cuda/std/__type_traits/is_floating_point.h>
-#include <cuda/std/__type_traits/is_same.h>
-#include <cuda/std/array>
-#include <cuda/std/climits>
+#ifndef _CCCL_DISABLE_CMATH
```
miscco, line 16: We need to have a discussion about the support matrix for the freestanding library

From our discussion I remember that this was meant to be limited to modern compilers where we can rely on the existence of the compiler builtins.

cub/cub/device/dispatch/tuning/tuning_radix_sort.cuh
```diff
@@ -25,7 +25,7 @@
 #include <cuda/__device/arch_id.h>
 #include <cuda/std/optional>
 
-#if !_CCCL_COMPILER(NVRTC)
+#if _CCCL_HOSTED()
 #  include <ostream>
 #endif
```
miscco, line 30: Missing comment
```suggestion
#endif // _CCCL_HOSTED()
```

cub/cub/device/dispatch/tuning/tuning_segmented_reduce.cuh
```diff
@@ -15,7 +15,7 @@
 
 #include <cub/device/dispatch/tuning/tuning_reduce.cuh>
 
-#if !_CCCL_COMPILER(NVRTC)
+#if _CCCL_HOSTED()
 #  include <ostream>
 #endif
```
miscco, line 20: We should unify them all
```suggestion
#if _CCCL_HOSTED()
#  include <ostream>
#endif // _CCCL_HOSTED()
```

cub/cub/device/dispatch/dispatch_reduce_deterministic.cuh
```diff
@@ -19,30 +19,32 @@
...
-#include <cuda/__type_traits/is_floating_point.h>
-#include <cuda/std/__algorithm/min.h>
-#include <cuda/std/__functional/identity.h>
-#include <cuda/std/__functional/invoke.h>
-#include <cuda/std/__functional/operations.h>
-#include <cuda/std/__type_traits/decay.h>
-#include <cuda/std/__type_traits/enable_if.h>
-#include <cuda/std/cstdint>
-#include <cuda/std/limits>
+#ifndef _CCCL_DISABLE_CMATH
```
miscco, line 22: I do not believe `_CCCL_DISABLE_CMATH` should be a thing

*IF* we cannot get cmath implemented in freestanding we should make it unavailable there, but there is no reason we should ever use that flag outside of freestanding

cub/cub/util_allocator.cuh
```diff
@@ -22,11 +22,13 @@
 #include <cub/util_debug.cuh>
 #include <cub/util_namespace.cuh>
 
-#include <cuda/std/__host_stdlib/math.h>
+#if _CCCL_HOSTED()
 
-#include <map>
-#include <mutex>
-#include <set>
+#  include <cuda/std/__host_stdlib/math.h>
```
miscco, line 27: Why was this added?
```suggestion
```

libcudacxx/include/cuda/__cmath/sincos.h
```diff
@@ -13,41 +13,43 @@
...
+
+#  if defined(_CCCL_IMPLICIT_SYSTEM_HEADER_GCC)
+#    pragma GCC system_header
+#  elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_CLANG)
+#    pragma clang system_header
+#  elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_MSVC)
+#    pragma system_header
+#  endif // no system header
+
+#  include <cuda/std/__cmath/trigonometric_functions.h>
```
miscco, line 26: Nitpick, this comes conventionally after the header guard
```suggestion
#  if defined(_CCCL_IMPLICIT_SYSTEM_HEADER_GCC)
#    pragma GCC system_header
#  elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_CLANG)
#    pragma clang system_header
#  elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_MSVC)
#    pragma system_header
#  endif // no system header

#ifndef _CCCL_DISABLE_CMATH

#  include <cuda/std/__cmath/trigonometric_functions.h>
```

libcudacxx/include/cuda/__device/physical_device.h
```diff
@@ -47,21 +50,27 @@ _CCCL_BEGIN_NAMESPACE_CUDA
...
 {
-  friend _CCCL_HOST_API inline ::std::unique_ptr<__physical_device[]>
+  friend _CCCL_HOST_API inline ::cuda::std::unique_ptr<__physical_device[]>
   __make_physical_devices(::cuda::std::size_t __device_count);
 
   ::CUdevice __device_{};
 
+#  if _CCCL_HOSTED()
   ::std::once_flag __primary_ctx_once_flag_{};
+#  endif // _CCCL_HOSTED()
```
miscco, line 60: We will need a different solution for that. Until then I would rather disable the whole feature

libcudacxx/include/cuda/__driver/driver_api.h
```diff
@@ -30,11 +30,14 @@
 #  include <cuda/std/__limits/numeric_limits.h>
 #  include <cuda/std/__type_traits/always_false.h>
 #  include <cuda/std/__type_traits/is_same.h>
-#  if _CCCL_OS(WINDOWS)
-#    include <windows.h>
-#  else
-#    include <dlfcn.h>
-#  endif
+
+#  if _CCCL_HOSTED()
```
miscco, line 34: Ditto, I believe in a freestanding implementation there is no driver

libcudacxx/include/cuda/std/__cccl/assert.h
```diff
@@ -27,9 +27,9 @@
 #include <cuda/std/__cccl/execution_space.h>
 #include <cuda/std/__cccl/preprocessor.h>
 
-#if !_CCCL_COMPILER(NVRTC)
+#if _CCCL_HOSTED()
 #  include <assert.h>
-#endif // !_CCCL_COMPILER(NVRTC)
+#endif // _CCCL_HOSTED()
```
miscco, line 32: We need to ensure that this header works with a freestanding implementation too.

Worst case we need to do some compiler magic / or define the macro to nothing

gevtushenko:
> Can't we rather `_CCCL_VERIFY(false, "Host implementation requires hosted evironment.")` in all of those cmath functions? That way we could still use them in `__device__` code.

The freestanding mode is implementation detail for now. I'd like to avoid spend a lot of time supporting freestanding math at the moment. I was thinking about enabling parts of cmath in freestanding as needed. Do you have any concerns about this approach?

> This PR is for example changing how `cuda::devices` data are initialized and stored, which is something I'd like to see in a separate PR.

Hm, changes in this PR shouldn't affect hosted mode. Do you see any change in behavior there? 


> Also, I don't see the freestanding mode tested. I'd like to have at least public header testing for this mode.

Agreed! The testing is a bit tricky, though. The goal is to support host JIT, which currently doesn't have any system headers. If I test with AOT compilation, it'll likely be very different setup (more built-ins available, some implicit system headers etc). My hope is to provide JIT compiler utility as a follow up, and test this mode as intended. Given that freestanding mode is not going to be publically available for now, the risk of not having tests is that I'd have to make a bit more changes when adding JIT compiler here, which doesn't look critical.

gevtushenko:
Suggestions:
- try dropping includes of `cuda/std/__cmath` on CUB end instead for now and see if that allows us to drop `_CCCL_DISABLE_CMATH`
- split the PR into 1. trivial changes (inclusion of iostream in cub) 2. non-trivial changes
- try providing a test, not necesserily integrated into CCCL, that'd show the scope of needed changes

cub/cub/agent/agent_radix_sort_onesweep.cuh
gevtushenko: Decided to go with `_CCCL_HOSTED`

PR #7776, opened 2026-02-24
Avoid use of class static variable in device function

miscco:
We should not use a class static member in a device function, because it is not defined there.

Luckily those tests do not need to be defined as `__device__` anyway

Fixes nvbug5925530

### DIFF
--- a/cub/test/catch2_test_device_segmented_reduce_large_offsets.cu
+++ b/cub/test/catch2_test_device_segmented_reduce_large_offsets.cu
@@ -194,13 +194,13 @@ struct dispatch_helper
   template <typename ActivePolicyT>
   CUB_RUNTIME_FUNCTION cudaError_t Invoke()
   {
-    thresholds = {ActivePolicyT::SmallReducePolicy::ITEMS_PER_TILE,
-                  ActivePolicyT::MediumReducePolicy::ITEMS_PER_TILE,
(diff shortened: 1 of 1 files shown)


cub/test/catch2_test_device_segmented_reduce_large_offsets.cu
```diff
@@ -192,15 +192,15 @@ struct dispatch_helper
   tuple_t thresholds{};
 
   template <typename ActivePolicyT>
-  CUB_RUNTIME_FUNCTION cudaError_t Invoke()
+  cudaError_t Invoke()
   {
     thresholds = {ActivePolicyT::SmallReducePolicy::ITEMS_PER_TILE,
                   ActivePolicyT::MediumReducePolicy::ITEMS_PER_TILE,
                   ActivePolicyT::ReducePolicy::BLOCK_THREADS * ActivePolicyT::ReducePolicy::ITEMS_PER_THREAD};
```
bernhardmgruber: Reading through the nvbug, maybe it's enough to just avoid ODR use:

```suggestion
thresholds = {+ActivePolicyT::SmallReducePolicy::ITEMS_PER_TILE,
+ActivePolicyT::MediumReducePolicy::ITEMS_PER_TILE,
+ActivePolicyT::ReducePolicy::BLOCK_THREADS * +ActivePolicyT::ReducePolicy::ITEMS_PER_THREAD};
```

PR #7792, opened 2026-02-25
Do not synchronize in `caching_allocator::{de}allocate`

miscco:
We should not need the additional synchronization.

### DIFF
--- a/nvbench_helper/nvbench_helper/nvbench_helper.cuh
+++ b/nvbench_helper/nvbench_helper/nvbench_helper.cuh
@@ -615,10 +615,10 @@ struct caching_allocator_t
     return result;
   }
 
-  void deallocate(::cuda::stream_ref, void* ptr, size_t num_bytes, size_t)
+  void deallocate(::cuda::stream_ref __stream, void* ptr, size_t num_bytes, size_t)
   {
+    __stream.sync();
     deallocate(static_cast<char*>(ptr), num_bytes);
-    // It is fine to not synchronize here, because there is no work done
   }
 #endif // THRUST_DEVICE_SYSTEM == THRUST_DEVICE_SYSTEM_CUDA
 


miscco:
> I don't believe this is correct. We should `.sync()` in `.deallocate()`, because if you just want to enqueue the deallocation in the stream, but you aren't yet done with the buffers, next allocation could pick up this buffer and overwrite your data.

yeah you are right, Its not only the pointer, but it *could* be that the operation has not yet finished

nvbench_helper/nvbench_helper/nvbench_helper.cuh
```diff
@@ -592,15 +592,33 @@ struct caching_allocator_t
...
-    void* __res = allocate(num_bytes);
-    __stream.sync();
-    return __res;
+    value_type* result{};
+    auto free_block = free_blocks.find(num_bytes);
+
+    if (free_block != free_blocks.end())
+    {
+      result = free_block->second;
+      free_blocks.erase(free_block);
```
davebayer, line 601:
```suggestion
      const auto [result, event] = free_block->second;
      free_blocks.erase(free_block);
      __stream.wait(event);
```

nvbench_helper/nvbench_helper/nvbench_helper.cuh
```diff
@@ -592,15 +592,33 @@ struct caching_allocator_t
...
+    allocated_blocks.insert(std::make_pair(result, num_bytes));
+    return result;
   }
 
-  void deallocate(::cuda::stream_ref __stream, void* ptr, size_t num_bytes, size_t)
+  void deallocate(::cuda::stream_ref __steam, void* ptr, size_t num_bytes, size_t)
   {
-    deallocate(static_cast<char*>(ptr), num_bytes);
     __stream.sync();
+    deallocate(static_cast<char*>(ptr), num_bytes);
```
davebayer, line 621:
```suggestion
    auto event = __stream.record_event();
    // store the event together with the pointer
    deallocate(static_cast<char*>(ptr), num_bytes);
```

nvbench_helper/nvbench_helper/nvbench_helper.cuh
miscco: nah I dont want to touch it any more

davebayer: Aaaah, it took me like 5 minutes to come up with this.. :sob:

nvbench_helper/nvbench_helper/nvbench_helper.cuh
```diff
@@ -592,15 +592,33 @@ struct caching_allocator_t
...
+    auto free_block = free_blocks.find(num_bytes);
+
+    if (free_block != free_blocks.end())
+    {
+      result = free_block->second;
+      free_blocks.erase(free_block);
+    }
+    else
+    {
+      const cudaError_t status = cudaMallocAsync(&result, num_bytes, __stream.get());
```
davebayer, line 605: I am not sure if you can allocate memory with `cudaMallocAsync` and then free it with `cudaFree`

miscco: Pretty sure you can

pciolkosz: You can: https://docs.nvidia.com/cuda/cuda-programming-guide/04-special-topics/stream-ordered-memory-allocation.html#freeing-memory
"Likewise, memory allocated with cudaMallocAsync can be freed with cudaFree()."

davebayer (approved):
But we should rework the allocator in the future

PR #7793, opened 2026-02-25
Add nvtx ranges to pstl algorithms

### LINKED ISSUE #7788: [FEA]: Provide NVTX ranges for PSTL algorithms
### Is this a duplicate?

### Area

libcu++

We are using the CUB backend, so we are getting some NVTX ranges, but it would be nice to get them similar to Thrust

We should add NVTC ranges to all parallel algorithms

_No response_

### Additional context

_No response_

miscco:
Fixes [FEA]: Provide NVTX ranges for PSTL algorithms Fixes #7788

### DIFF
--- a/libcudacxx/include/cuda/std/__pstl/copy_if.h
+++ b/libcudacxx/include/cuda/std/__pstl/copy_if.h
@@ -23,6 +23,7 @@
 
 #if !_CCCL_COMPILER(NVRTC)
 
+#  include <cuda/__nvtx/nvtx.h>
 #  include <cuda/std/__algorithm/copy_if.h>
 #  include <cuda/std/__concepts/concept_macros.h>
 #  include <cuda/std/__execution/policy.h>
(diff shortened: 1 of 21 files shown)


libcudacxx/include/cuda/std/__pstl/copy_if.h
```diff
@@ -67,6 +68,7 @@ _CCCL_HOST_API _OutputIterator copy_if(
     ::cuda::std::execution::__pstl_select_dispatch<::cuda::std::execution::__pstl_algorithm::__copy_if, _Policy>();
   if constexpr (::cuda::std::execution::__pstl_can_dispatch<decltype(__dispatch)>)
   {
+    _CCCL_NVTX_RANGE_SCOPE("cuda::std::copy_if");
```
bernhardmgruber, line 71: Suggestion: Add the range for the scope of the entire function, as we do in CUB and Thrust. Otherwise, it's weird to not see the API only if the range is not empty or when the CUDA backend is turned on.

miscco: I am not sure yet, because we do not offer NVTX ranges for serial algorithms and in that case we would have to add them there too

PR #7795, opened 2026-02-25
Add env SegmentedReduce (non fixed-size overloads)

gonidelis:
Split (1/2)

Adds env based overloads for non fixed-size segments `DeviceSegmentedReduce::*` algorithms

Merge before #8097  

**Segmented Reduce is inherently `run_to_run` deterministic thus this is the largest deterministic guarantee allowed. If you believe there at some point can be an a perf optimization that will ruin this contract let me know and we will remove this promise in this PR. Otherwise we stay bound to that.**

### DIFF
--- a/cub/cub/device/device_segmented_reduce.cuh
+++ b/cub/cub/device/device_segmented_reduce.cuh
@@ -104,7 +104,6 @@ private:
     using requested_determinism_t =
       ::cuda::std::execution::__query_result_or_t<requirements_t,
                                                   ::cuda::execution::determinism::__get_determinism_t,
-                                                  ::cuda::execution::determinism::not_guaranteed_t,
                                                   ::cuda::execution::determinism::run_to_run_t>;
 
     static_assert(!::cuda::std::is_same_v<requested_determinism_t, ::cuda::execution::determinism::gpu_to_gpu_t>,
@@ -121,6 +120,7 @@ private:
             ::cuda::std::execution::__query_result_or_t<decltype(tuning),
                                                         detail::segmented_reduce::get_tuning_query_t,
                                                         default_policy_selector_t>;
+          // TODO: in most cases we can just take the default AccumT and OffsetT. Refactor this
           return detail::segmented_reduce::dispatch<AccumT, OffsetT>(
             d_temp_storage,
             temp_storage_bytes,
--- a/cub/test/catch2_test_device_segmented_reduce_env_api.cu
+++ b/cub/test/catch2_test_device_segmented_reduce_env_api.cu
@@ -267,6 +267,176 @@ C2H_TEST("cub::DeviceSegmentedReduce::ArgMax env-based API", "[segmented_reduce]
   REQUIRE(h_out[2].value == 9);
 }
 
+C2H_TEST("cub::DeviceSegmentedReduce::Min accepts run_to_run determinism requirements", "[segmented_reduce][env]")
+{
+  int num_segments                     = 3;
+  thrust::device_vector<int> d_offsets = {0, 3, 3, 7};
+  auto d_offsets_it                    = thrust::raw_pointer_cast(d_offsets.data());
+  thrust::device_vector<int> d_in{8, 6, 7, 5, 3, 0, 9};
+  thrust::device_vector<int> d_out(3);
+
+  auto env = cuda::execution::require(cuda::execution::determinism::run_to_run);
+
+  auto error =
+    cub::DeviceSegmentedReduce::Min(d_in.begin(), d_out.begin(), num_segments, d_offsets_it, d_offsets_it + 1, env);
+  thrust::device_vector<int> expected{6, std::numeric_limits<int>::max(), 0};
+
+  REQUIRE(d_out == expected);
+  REQUIRE(error == cudaSuccess);
+}
+
+C2H_TEST("cub::DeviceSegmentedReduce::Min accepts not_guaranteed determinism requirements", "[segmented_reduce][env]")
+{
+  int num_segments                     = 3;
+  thrust::device_vector<int> d_offsets = {0, 3, 3, 7};
+  auto d_offsets_it                    = thrust::raw_pointer_cast(d_offsets.data());
+  thrust::device_vector<int> d_in{8, 6, 7, 5, 3, 0, 9};
+  thrust::device_vector<int> d_out(3);
+
+  auto env = cuda::execution::require(cuda::execution::determinism::not_guaranteed);
+
+  auto error =
+    cub::DeviceSegmentedReduce::Min(d_in.begin(), d_out.begin(), num_segments, d_offsets_it, d_offsets_it + 1, env);
+  thrust::device_vector<int> expected{6, std::numeric_limits<int>::max(), 0};
+
+  REQUIRE(d_out == expected);
+  REQUIRE(error == cudaSuccess);
+}
+
+C2H_TEST("cub::DeviceSegmentedReduce::Max accepts run_to_run determinism requirements", "[segmented_reduce][env]")
+{
+  int num_segments                     = 3;
+  thrust::device_vector<int> d_offsets = {0, 3, 3, 7};
+  auto d_offsets_it                    = thrust::raw_pointer_cast(d_offsets.data());
+  thrust::device_vector<int> d_in{8, 6, 7, 5, 3, 0, 9};
+  thrust::device_vector<int> d_out(3);
+
+  auto env = cuda::execution::require(cuda::execution::determinism::run_to_run);
+
+  auto error =
+    cub::DeviceSegmentedReduce::Max(d_in.begin(), d_out.begin(), num_segments, d_offsets_it, d_offsets_it + 1, env);
+  thrust::device_vector<int> expected{8, std::numeric_limits<int>::lowest(), 9};
+
+  REQUIRE(d_out == expected);
+  REQUIRE(error == cudaSuccess);
+}
+
+C2H_TEST("cub::DeviceSegmentedReduce::Max accepts not_guaranteed determinism requirements", "[segmented_reduce][env]")
+{
+  int num_segments                     = 3;
+  thrust::device_vector<int> d_offsets = {0, 3, 3, 7};
+  auto d_offsets_it                    = thrust::raw_pointer_cast(d_offsets.data());
+  thrust::device_vector<int> d_in{8, 6, 7, 5, 3, 0, 9};
+  thrust::device_vector<int> d_out(3);
+
+  auto env = cuda::execution::require(cuda::execution::determinism::not_guaranteed);
+
+  auto error =
+    cub::DeviceSegmentedReduce::Max(d_in.begin(), d_out.begin(), num_segments, d_offsets_it, d_offsets_it + 1, env);
+  thrust::device_vector<int> expected{8, std::numeric_limits<int>::lowest(), 9};
+
+  REQUIRE(d_out == expected);
+  REQUIRE(error == cudaSuccess);
+}
+
+C2H_TEST("cub::DeviceSegmentedReduce::ArgMin accepts run_to_run determinism requirements", "[segmented_reduce][env]")
+{
+  int num_segments                     = 3;
+  thrust::device_vector<int> d_offsets = {0, 3, 3, 7};
+  auto d_offsets_it                    = thrust::raw_pointer_cast(d_offsets.data());
+  thrust::device_vector<int> d_in{8, 6, 7, 5, 3, 0, 9};
+  thrust::device_vector<cub::KeyValuePair<int, int>> d_out(3);
+
+  auto env = cuda::execution::require(cuda::execution::determinism::run_to_run);
+
+  auto error =
+    cub::DeviceSegmentedReduce::ArgMin(d_in.begin(), d_out.begin(), num_segments, d_offsets_it, d_offsets_it + 1, env);
+
+  REQUIRE(error == cudaSuccess);
+
+  thrust::host_vector<cub::KeyValuePair<int, int>> h_out(d_out);
+  REQUIRE(h_out[0].key == 1);
+  REQUIRE(h_out[0].value == 6);
+  REQUIRE(h_out[1].key == 1);
+  REQUIRE(h_out[1].value == std::numeric_limits<int>::max());
+  REQUIRE(h_out[2].key == 2);
+  REQUIRE(h_out[2].value == 0);
+}
+
+C2H_TEST("cub::DeviceSegmentedReduce::ArgMin accepts not_guaranteed determinism requirements",
+         "[segmented_reduce][env]")
+{
+  int num_segments                     = 3;
+  thrust::device_vector<int> d_offsets = {0, 3, 3, 7};
+  auto d_offsets_it                    = thrust::raw_pointer_cast(d_offsets.data());
+  thrust::device_vector<int> d_in{8, 6, 7, 5, 3, 0, 9};
+  thrust::device_vector<cub::KeyValuePair<int, int>> d_out(3);
+
+  auto env = cuda::execution::require(cuda::execution::determinism::not_guaranteed);
+
+  auto error =
+    cub::DeviceSegmentedReduce::ArgMin(d_in.begin(), d_out.begin(), num_segments, d_offsets_it, d_offsets_it + 1, env);
+
+  REQUIRE(error == cudaSuccess);
+
+  thrust::host_vector<cub::KeyValuePair<int, int>> h_out(d_out);
+  REQUIRE(h_out[0].key == 1);
+  REQUIRE(h_out[0].value == 6);
+  REQUIRE(h_out[1].key == 1);
+  REQUIRE(h_out[1].value == std::numeric_limits<int>::max());
+  REQUIRE(h_out[2].key == 2);
+  REQUIRE(h_out[2].value == 0);
+}
+
+C2H_TEST("cub::DeviceSegmentedReduce::ArgMax accepts run_to_run determinism requirements", "[segmented_reduce][env]")
+{
+  int num_segments                     = 3;
+  thrust::device_vector<int> d_offsets = {0, 3, 3, 7};
+  auto d_offsets_it                    = thrust::raw_pointer_cast(d_offsets.data());
+  thrust::device_vector<int> d_in{8, 6, 7, 5, 3, 0, 9};
+  thrust::device_vector<cub::KeyValuePair<int, int>> d_out(3);
+
+  auto env = cuda::execution::require(cuda::execution::determinism::run_to_run);
+
+  auto error =
+    cub::DeviceSegmentedReduce::ArgMax(d_in.begin(), d_out.begin(), num_segments, d_offsets_it, d_offsets_it + 1, env);
+
+  REQUIRE(error == cudaSuccess);
+
+  thrust::host_vector<cub::KeyValuePair<int, int>> h_out(d_out);
+  REQUIRE(h_out[0].key == 0);
+  REQUIRE(h_out[0].value == 8);
+  REQUIRE(h_out[1].key == 1);
+  REQUIRE(h_out[1].value == std::numeric_limits<int>::lowest());
+  REQUIRE(h_out[2].key == 3);
+  REQUIRE(h_out[2].value == 9);
+}
+
+C2H_TEST("cub::DeviceSegmentedReduce::ArgMax accepts not_guaranteed determinism requirements",
+         "[segmented_reduce][env]")
+{
+  int num_segments                     = 3;
+  thrust::device_vector<int> d_offsets = {0, 3, 3, 7};
+  auto d_offsets_it                    = thrust::raw_pointer_cast(d_offsets.data());
+  thrust::device_vector<int> d_in{8, 6, 7, 5, 3, 0, 9};
+  thrust::device_vector<cub::KeyValuePair<int, int>> d_out(3);
+
+  auto env = cuda::execution::require(cuda::execution::determinism::not_guaranteed);
+
+  auto error =
+    cub::DeviceSegmentedReduce::ArgMax(d_in.begin(), d_out.begin(), num_segments, d_offsets_it, d_offsets_it + 1, env);
+
+  REQUIRE(error == cudaSuccess);
+
+  thrust::host_vector<cub::KeyValuePair<int, int>> h_out(d_out);
+  REQUIRE(h_out[0].key == 0);
+  REQUIRE(h_out[0].value == 8);
+  REQUIRE(h_out[1].key == 1);
+  REQUIRE(h_out[1].value == std::numeric_limits<int>::lowest());
+  REQUIRE(h_out[2].key == 3);
+  REQUIRE(h_out[2].value == 9);
+}
+
 C2H_TEST("cub::DeviceSegmentedReduce::Reduce accepts run_to_run determinism requirements", "[segmented_reduce][env]")
 {
   // example-begin segmented-reduce-reduce-env-determinism


cub/cub/device/device_segmented_reduce.cuh
```diff
@@ -213,6 +271,123 @@ struct DeviceSegmentedReduce
...
+  //!   **[optional]** Execution environment. Default is ``cuda::std::execution::env{}``.
+  //!   @endrst
+  template <typename InputIteratorT,
+            typename OutputIteratorT,
+            typename BeginOffsetIteratorT,
+            typename EndOffsetIteratorT,
+            typename ReductionOpT,
+            typename T,
+            typename      = ::cuda::std::void_t<typename ::cuda::std::iterator_traits<BeginOffsetIteratorT>::value_type,
+                                                typename ::cuda::std::iterator_traits<EndOffsetIteratorT>::value_type>,
```
bernhardmgruber: Question: This constraint looks super weird and I am not even sure it's correct :S Why is this necessary?

Suggestion: if you even need a constraint, can we just constrain `InputIteratorT` to not be `void*` in case we this is trying to prevent an ambiguity with the old overloads?

Applies to the other added overloads as well.

gonidelis:
I removed the helper underlying implementation function for fixed segment size overloads as it pre required knowledge of the `AccumT` and added extra logic that was unnecessary. Non fixed-size overloads still do use the `*_impl` function though

cub/cub/device/device_segmented_reduce.cuh
```diff
@@ -78,6 +79,63 @@ struct get_tuning_query_t
...
+
+    static_assert(::cuda::std::is_integral_v<OffsetT>, "Offset iterator value type should be integral.");
+    if constexpr (::cuda::std::is_integral_v<OffsetT>)
+    {
+      return detail::dispatch_with_env(
+        env, [&](auto tuning, void* d_temp_storage, size_t& temp_storage_bytes, cudaStream_t stream) {
+          using segmented_reduce_tuning_t = ::cuda::__call_result_or_t<
+            detail::segmented_reduce::get_tuning_query_t,
+            detail::segmented_reduce::policy_selector_from_types<AccumT, OffsetT, ReductionOpT>,
+            decltype(tuning)>;
```
bernhardmgruber: Important: querying the tuning environment should use `__query_result_or_t`:
```suggestion
          using default_policy_selector_t =
            detail::segmented_reduce::policy_selector_from_types<AccumT, OffsetT, ReductionOpT>;
          using policy_selector_t = ::cuda::__query_result_or_t<
            decltype(tuning),
            detail::segmented_reduce::get_tuning_query_t
            default_policy_selector_t>;
```
Also, let's use the term policy selector where it applies. The term tuning is too overloaded already.

cub/cub/device/device_segmented_reduce.cuh
```diff
@@ -78,6 +79,63 @@ struct get_tuning_query_t
...
+            temp_storage_bytes,
+            d_in,
+            d_out,
+            num_segments,
+            d_begin_offsets,
+            d_end_offsets,
+            reduction_op,
+            initial_value,
+            stream,
+            segmented_reduce_tuning_t{});
```
bernhardmgruber:
```suggestion
            default_policy_selector_t{});
```

cub/test/catch2_test_device_segmented_reduce_env.cu
```diff
@@ -1,14 +1,327 @@
-// SPDX-FileCopyrightText: Copyright (c) 2025, NVIDIA CORPORATION. All rights reserved.
-// SPDX-License-Identifier: BSD-3-Clause
+// SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
```
bernhardmgruber: Critical: we must not change the license of existing code. Please revert.

gonidelis: yeah i caught that later on sorry

gonidelis:
adding missing unit tests just now

cub/cub/device/device_segmented_reduce.cuh
```diff
@@ -685,15 +890,96 @@ struct DeviceSegmentedReduce
...
+      env, [&]([[maybe_unused]] auto tuning, void* d_temp_storage, size_t& temp_storage_bytes, cudaStream_t stream) {
+        return detail::reduce::
+          DispatchFixedSizeSegmentedReduce<InputIteratorT, OutputIteratorT, int, ::cuda::std::plus<>, output_t>::Dispatch(
+            d_temp_storage,
+            temp_storage_bytes,
+            d_in,
+            d_out,
+            num_segments,
+            segment_size,
+            ::cuda::std::plus{},
```
miscco: Question: This uses `plus<void>` and we have observed performance issues with this, because for smaller integer types it promotes. Shuld this rather be 
```suggestion
            ::cuda::std::plus<detail::it_value_t<InputIteratorT>>{},
```

cub/cub/device/device_segmented_reduce.cuh
```diff
@@ -808,6 +1094,106 @@ struct DeviceSegmentedReduce
...
+      BeginOffsetIteratorT d_begin_offsets,
+      EndOffsetIteratorT d_end_offsets,
+      EnvT env = {})
+  {
+    _CCCL_NVTX_RANGE_SCOPE("cub::DeviceSegmentedReduce::Min");
+
+    using OffsetT = detail::common_iterator_value_t<BeginOffsetIteratorT, EndOffsetIteratorT>;
+    using InputT  = detail::it_value_t<InputIteratorT>;
+    using init_t  = InputT;
+    using op_t    = ::cuda::minimum<>;
```
miscco, line 1018: Ditto: Should this rather be 
```suggestion
    using op_t    = ::cuda::minimum<InputT>;
```

cub/cub/device/device_segmented_reduce.cuh
```diff
@@ -78,6 +79,65 @@ struct get_tuning_query_t
...
+      ::cuda::std::execution::__query_result_or_t<requirements_t,
+                                                  ::cuda::execution::determinism::__get_determinism_t,
+                                                  ::cuda::execution::determinism::run_to_run_t>;
+
+    static_assert(!::cuda::std::is_same_v<requested_determinism_t, ::cuda::execution::determinism::gpu_to_gpu_t>,
+                  "gpu_to_gpu determinism is not supported for device segmented reductions ");
+
+    static_assert(::cuda::std::is_integral_v<OffsetT>, "Offset iterator value type should be integral.");
+    if constexpr (::cuda::std::is_integral_v<OffsetT>)
+    {
```
miscco, line 114: Question: Is this really needed anymore or is the static assert sufficient?

cub/cub/device/device_segmented_reduce.cuh
```diff
@@ -808,6 +1094,106 @@ struct DeviceSegmentedReduce
...
+    _CCCL_NVTX_RANGE_SCOPE("cub::DeviceSegmentedReduce::Min");
+
+    using OffsetT = detail::common_iterator_value_t<BeginOffsetIteratorT, EndOffsetIteratorT>;
+    using InputT  = detail::it_value_t<InputIteratorT>;
+    using init_t  = InputT;
+    using op_t    = ::cuda::minimum<>;
+    using AccumT  = ::cuda::std::__accumulator_t<op_t, cub::detail::it_value_t<InputIteratorT>, init_t>;
+
+    return segmented_reduce_impl<AccumT, OffsetT>(
+      d_in, d_out, num_segments, d_begin_offsets, d_end_offsets, op_t{}, ::cuda::std::numeric_limits<init_t>::max(), env);
```
miscco, line 1025: This implicitly requires `numeric_limits` to be specialized, which is commonly not the case. Should we assert that to avoid unfortunate compile issues?

cub/cub/device/device_segmented_reduce.cuh
```diff
@@ -894,6 +1280,87 @@ struct DeviceSegmentedReduce
...
+      env, [&]([[maybe_unused]] auto tuning, void* d_temp_storage, size_t& temp_storage_bytes, cudaStream_t stream) {
+        return detail::reduce::
+          DispatchFixedSizeSegmentedReduce<InputIteratorT, OutputIteratorT, int, ::cuda::minimum<>, input_t>::Dispatch(
+            d_temp_storage,
+            temp_storage_bytes,
+            d_in,
+            d_out,
+            num_segments,
+            segment_size,
+            ::cuda::minimum<>{},
```
miscco: Ditto: explicit 
```suggestion
            ::cuda::minimum<input_t>{},
```

cub/cub/device/device_segmented_reduce.cuh
bernhardmgruber: This is a @gevtushenko UX thing. He thinks that compiler error messages should stop at the first static assert and not continue compilation and produce errors for follow-up issues, generating a longer error report that can confuse users.

I don't believe in this rational. If a user wants the compiler to stop at the first error, they can just pass a flag like `-Wfatal-errors`. So I keep the `if constexpr` where it's alreay there and also don't mind if I see it in a review. I don't write those myself.

cub/cub/device/device_segmented_reduce.cuh
bernhardmgruber: Such changes should definitely go to separate PRs, since they change the status quo. AFAIK @gonidelis copies the setup for the dispatch call from the other non-env overloads.

cub/cub/device/device_segmented_reduce.cuh
bernhardmgruber: That's a good point! @gonidelis please also add this to the non-env overload:
```c++
static_assert(::cuda::std::numeric_limits<init_t>::is_specialized, "cuda::std::numeric_limits must be specialized for th input type");
```

gonidelis:
@miscco  https://github.com/NVIDIA/cccl/issues/7974#issue-4052456659 ok?

cub/cub/device/device_segmented_reduce.cuh
gonidelis: true ☝🏼 why do they change status quo?

cub/cub/device/device_segmented_reduce.cuh
gonidelis: done

cub/cub/device/device_segmented_reduce.cuh
gonidelis: resolving these per bernhard's suggestion and will handle in a separate PR

cub/test/catch2_test_device_segmented_reduce_env_api.cu
```diff
@@ -116,3 +117,286 @@ C2H_TEST("cub::DeviceSegmentedReduce::Sum accepts not_guaranteed determinism req
   REQUIRE(d_out == expected);
   REQUIRE(error == cudaSuccess);
 }
+
+C2H_TEST("cub::DeviceSegmentedReduce::Reduce env-based API", "[segmented_reduce][env]")
```
srinivasyadav18, line 123: where is `env` used in the `env API` tests ? If the focus here is just to show single-phase API with default `env` ? 

We use stream or memory resources in other algorithm env API tests to show the usage. Do we want to do same here as-well ?

cub/cub/device/device_segmented_reduce.cuh
miscco: For integer types `plus<>` introduces integer promotion, which e.g `plus<short>` does not.

So depending on the tested types, this can actually have some considerable performance implications

cub/cub/device/device_segmented_reduce.cuh
```diff
@@ -521,90 +704,11 @@ struct DeviceSegmentedReduce
     using OffsetT = detail::common_iterator_value_t<BeginOffsetIteratorT, EndOffsetIteratorT>;
     using OutputT = detail::non_void_value_t<OutputIteratorT, detail::it_value_t<InputIteratorT>>;
     using init_t  = OutputT;
-    using op_t    = ::cuda::std::plus<>;
+    using op_t    = ::cuda::std::plus<detail::it_value_t<InputIteratorT>>;
```
miscco: I thought we wanted to use `plus<>` and change it globally with proper perf investigations

cub/cub/device/device_segmented_reduce.cuh
```diff
@@ -862,34 +975,133 @@ struct DeviceSegmentedReduce
...
+    _CCCL_NVTX_RANGE_SCOPE("cub::DeviceSegmentedReduce::Min");
 
-    // `offset_t` a.k.a `SegmentSizeT` is fixed to `int` type now, but later can be changed to accept
-    // integral constant or larger integral types
-    using offset_t = int;
-    using input_t  = detail::it_value_t<InputIteratorT>;
+    using OffsetT = detail::common_iterator_value_t<BeginOffsetIteratorT, EndOffsetIteratorT>;
+    using InputT  = detail::it_value_t<InputIteratorT>;
+    using init_t  = InputT;
+    using op_t    = ::cuda::minimum<InputT>;
```
miscco: Ditto

cub/cub/device/device_segmented_reduce.cuh
bernhardmgruber: Critical: This is a breaking change, why is this necessary? I would revert this.

cub/cub/device/device_segmented_reduce.cuh
```diff
@@ -505,10 +690,8 @@ struct DeviceSegmentedReduce
             typename OutputIteratorT,
             typename BeginOffsetIteratorT,
             typename EndOffsetIteratorT,
-            typename      = ::cuda::std::void_t<typename ::cuda::std::iterator_traits<BeginOffsetIteratorT>::value_type,
-                                                typename ::cuda::std::iterator_traits<EndOffsetIteratorT>::value_type>,
```
bernhardmgruber, line 509: Q: This seems to implement a constraint that begin and end offset iterators have a `value_type`. Why is this no longer necessary?

cub/cub/device/device_segmented_reduce.cuh
```diff
@@ -794,6 +898,8 @@ struct DeviceSegmentedReduce
     using OffsetT = detail::common_iterator_value_t<BeginOffsetIteratorT, EndOffsetIteratorT>;
     using InputT  = detail::it_value_t<InputIteratorT>;
     using init_t  = InputT;
+    static_assert(::cuda::std::numeric_limits<init_t>::is_specialized,
+                  "numeric_limits must be specialized for the input value type");
```
bernhardmgruber, line 905: Remark: I guess this is ok since we use `::cuda::std::numeric_limits<init_t>:max()` below.

cub/cub/device/device_segmented_reduce.cuh
bernhardmgruber: Important: This is different then the non-env overload, which uses `minimum<>`. I think we should stay consistent with the old overload.

cub/cub/device/device_segmented_reduce.cuh
```diff
@@ -1279,6 +1620,116 @@ struct DeviceSegmentedReduce
...
+      BeginOffsetIteratorT d_begin_offsets,
+      EndOffsetIteratorT d_end_offsets,
+      EnvT env = {})
+  {
+    _CCCL_NVTX_RANGE_SCOPE("cub::DeviceSegmentedReduce::Max");
+
+    using OffsetT = detail::common_iterator_value_t<BeginOffsetIteratorT, EndOffsetIteratorT>;
+    using InputT  = cub::detail::it_value_t<InputIteratorT>;
+    using init_t  = InputT;
+    using op_t    = ::cuda::maximum<InputT>;
```
bernhardmgruber: Here as well.

cub/cub/device/device_segmented_reduce.cuh
```diff
@@ -78,6 +79,66 @@ struct get_tuning_query_t
...
+    ReductionOpT reduction_op,
+    InitT initial_value,
+    EnvT env)
+  {
+    using requirements_t = ::cuda::std::execution::
+      __query_result_or_t<EnvT, ::cuda::execution::__get_requirements_t, ::cuda::std::execution::env<>>;
+    using requested_determinism_t =
+      ::cuda::std::execution::__query_result_or_t<requirements_t,
+                                                  ::cuda::execution::determinism::__get_determinism_t,
+                                                  ::cuda::execution::determinism::run_to_run_t>;
```
pauleonix, line 107: Non-guaranteed might be missing

cub/cub/device/device_segmented_reduce.cuh
```diff
@@ -78,6 +79,66 @@ struct get_tuning_query_t
...
+
+    static_assert(!::cuda::std::is_same_v<requested_determinism_t, ::cuda::execution::determinism::gpu_to_gpu_t>,
+                  "gpu_to_gpu determinism is not supported for device segmented reductions ");
+
+    static_assert(::cuda::std::is_integral_v<OffsetT>, "Offset iterator value type should be integral.");
+    if constexpr (::cuda::std::is_integral_v<OffsetT>)
```
pauleonix, line 113: Not needed after static assert

cub/cub/device/device_segmented_reduce.cuh
```diff
@@ -78,6 +79,66 @@ struct get_tuning_query_t
...
+    {
+      return detail::dispatch_with_env(
+        env, [&](auto tuning, void* d_temp_storage, size_t& temp_storage_bytes, cudaStream_t stream) {
```
pauleonix, line 116: Is this on par with the current tuning infrastructure?

cub/cub/device/device_segmented_reduce.cuh
```diff
@@ -214,6 +275,130 @@ struct DeviceSegmentedReduce
...
+  //!   ``[d_begin_offsets, d_begin_offsets + num_segments)`` nor
+  //!   ``[d_end_offsets, d_end_offsets + num_segments)``.
+  //! - Can use a specific stream or cuda memory resource through the ``env`` parameter
+  //!
+  //! Snippet
+  //! +++++++++++++++++++++++++++++++++++++++++++++
+  //!
+  //! The code snippet below illustrates a custom min-reduction of a device vector of ``int`` data elements.
+  //!
+  //! .. literalinclude:: ../../../cub/test/catch2_test_device_segmented_reduce_env_api.cu
```
pauleonix: Relative path could be shorter (remove `../cub`)

cub/cub/device/device_segmented_reduce.cuh
```diff
@@ -78,6 +79,66 @@ struct get_tuning_query_t
...
+    {
+      return detail::dispatch_with_env(
+        env, [&](auto tuning, void* d_temp_storage, size_t& temp_storage_bytes, cudaStream_t stream) {
+          using default_policy_selector_t =
+            detail::segmented_reduce::policy_selector_from_types<AccumT, OffsetT, ReductionOpT>;
+          using policy_selector_t =
+            ::cuda::std::execution::__query_result_or_t<decltype(tuning),
+                                                        detail::segmented_reduce::get_tuning_query_t,
+                                                        default_policy_selector_t>;
+          return detail::segmented_reduce::dispatch<AccumT, OffsetT>(
```
pauleonix, line 124: Why does this differ from the old API which did not pass `AccumT` and `OffsetT` to dispatch?

cub/cub/device/device_segmented_reduce.cuh
```diff
@@ -428,8 +613,8 @@ struct DeviceSegmentedReduce
   //! @rst
   //! Computes a device-wide segmented sum using the addition (``+``) operator.
   //!
-  //! .. versionadded:: 2.2.0
-  //!    First appears in CUDA Toolkit 12.3.
+  //! .. versionadded:: 3.4.0
```
pauleonix: Maybe the diff is just funky, but m,ake sure the old API did not get overwritten

cub/cub/device/device_segmented_reduce.cuh
pauleonix: Is this needed here? @miscco

cub/cub/device/device_segmented_reduce.cuh
```diff
@@ -505,10 +690,8 @@ struct DeviceSegmentedReduce
             typename OutputIteratorT,
             typename BeginOffsetIteratorT,
             typename EndOffsetIteratorT,
-            typename      = ::cuda::std::void_t<typename ::cuda::std::iterator_traits<BeginOffsetIteratorT>::value_type,
-                                                typename ::cuda::std::iterator_traits<EndOffsetIteratorT>::value_type>,
             typename EnvT = ::cuda::std::execution::env<>>
```
pauleonix, line 696: Is above SFINAE not needed for the env version?

cub/cub/device/device_segmented_reduce.cuh
pauleonix: Again, does this need to be specified?

cub/cub/device/device_segmented_reduce.cuh
```diff
@@ -794,6 +898,8 @@ struct DeviceSegmentedReduce
     using OffsetT = detail::common_iterator_value_t<BeginOffsetIteratorT, EndOffsetIteratorT>;
     using InputT  = detail::it_value_t<InputIteratorT>;
     using init_t  = InputT;
+    static_assert(::cuda::std::numeric_limits<init_t>::is_specialized,
```
pauleonix, line 904: Should this be in this PR?

cub/cub/device/device_segmented_reduce.cuh
```diff
@@ -862,34 +975,133 @@ struct DeviceSegmentedReduce
...
+      cudaStream_t stream = 0)
+  {
+    _CCCL_NVTX_RANGE_SCOPE_IF(d_temp_storage, "cub::DeviceSegmentedReduce::Min");
+
+    // `offset_t` a.k.a `SegmentSizeT` is fixed to `int` type now, but later can be changed to accept
+    // integral constant or larger integral types
+    using offset_t = int;
+    using input_t  = detail::it_value_t<InputIteratorT>;
+
+    static_assert(::cuda::std::numeric_limits<input_t>::is_specialized,
```
pauleonix, line 1101: Should this be in this PR?

cub/cub/device/device_segmented_reduce.cuh
```diff
@@ -1018,6 +1230,8 @@ struct DeviceSegmentedReduce
     using InitT          = detail::reduce::empty_problem_init_t<OverrideAccumT>;
 
     static_assert(::cuda::std::is_same_v<int, OutputKeyT>, "Output key type must be int.");
+    static_assert(::cuda::std::numeric_limits<InputValueT>::is_specialized,
```
pauleonix, line 1236: Should this be in this PR?

cub/cub/device/device_segmented_reduce.cuh
pauleonix: No need to specialize?

cub/cub/device/device_segmented_reduce.cuh
```diff
@@ -1507,6 +1963,129 @@ struct DeviceSegmentedReduce
...
+  //!   (e.g., addition of floating point types) on the same GPU device.
+  //!   However, results for pseudo-associative reduction may be inconsistent
+  //!   from one device to a another device of a different compute-capability
+  //!   because CUB can employ different tile-sizing for different architectures.
+  //! - When input a contiguous sequence of segments, a single sequence
+  //!   ``segment_offsets`` (of length ``num_segments + 1``) can be aliased
+  //!   for both the ``d_begin_offsets`` and ``d_end_offsets`` parameters (where
+  //!   the latter is specified as ``segment_offsets + 1``).
+  //! - Can use a specific stream or cuda memory resource through the ``env`` parameter
+  //!
```
pauleonix, line 2004: Missing
```
- Let ``s`` be in ``[0, num_segments)``. The range
  //!   ``[d_out + d_begin_offsets[s], d_out + d_end_offsets[s])`` shall not
  //!   overlap ``[d_in + d_begin_offsets[s], d_in + d_end_offsets[s])``,
  //!   ``[d_begin_offsets, d_begin_offsets + num_segments)`` nor
  //!   ``[d_end_offsets, d_end_offsets + num_segments)``.
```
Check previous overload of ArgMin

cub/cub/device/device_segmented_reduce.cuh
```diff
@@ -20,6 +20,7 @@
 
 #include <cub/detail/choose_offset.cuh>
 #include <cub/detail/device_memory_resource.cuh>
+#include <cub/detail/env_dispatch.cuh>
```
pauleonix, line 23: Fix licence above

cub/test/catch2_test_device_segmented_reduce_env.cu
```diff
@@ -1,14 +1,496 @@
...
+DECLARE_LAUNCH_WRAPPER(cub::DeviceSegmentedReduce::ArgMin, device_segmented_reduce_argmin);
+DECLARE_LAUNCH_WRAPPER(cub::DeviceSegmentedReduce::ArgMax, device_segmented_reduce_argmax);
+
+// %PARAM% TEST_LAUNCH lid 0:1:2
+
+#include <cuda/__execution/require.h>
 
 #include <c2h/catch2_test_helper.h>
 
+namespace stdexec = cuda::std::execution;
```
pauleonix, line 30: Should this be `cuda::execution`?

cub/test/catch2_test_device_segmented_reduce_env.cu
```diff
@@ -1,14 +1,496 @@
...
+    == cub::DeviceSegmentedReduce::Reduce(
+      nullptr,
+      expected_bytes_allocated,
+      d_in.begin(),
+      d_out.begin(),
+      num_segments,
+      d_offsets_it,
+      d_offsets_it + 1,
+      ::cuda::std::plus<>{},
+      0));
```
pauleonix, line 164: Maybe have the 0 everywhere or nowhere?

cub/test/catch2_test_device_segmented_reduce_env.cu
```diff
@@ -1,14 +1,496 @@
...
+    == cub::DeviceSegmentedReduce::Max(
+      nullptr, expected_bytes_allocated, d_in.begin(), d_out.begin(), num_segments, d_offsets_it, d_offsets_it + 1));
+
+  auto env = stdexec::env{expected_allocation_size(expected_bytes_allocated)};
+
+  device_segmented_reduce_max(d_in.begin(), d_out.begin(), num_segments, d_offsets_it, d_offsets_it + 1, env);
+
+  thrust::device_vector<int> expected{8, 9, 2};
+  REQUIRE(d_out == expected);
+}
```
pauleonix, line 239: Missing argmin and argmax?

pauleonix: NVM, they are below, order could be better

cub/test/catch2_test_device_segmented_reduce_env.cu
```diff
@@ -1,14 +1,496 @@
...
+
+  REQUIRE(cudaSuccess == cudaStreamSynchronize(custom_stream));
+
+  thrust::device_vector<int> expected{8, 9, 2};
+  REQUIRE(d_out == expected);
+
+  REQUIRE(cudaSuccess == cudaStreamDestroy(custom_stream));
+}
+
+C2H_TEST("Device segmented argmin uses environment", "[segmented_reduce][device]")
```
pauleonix: This belongs higher above all custom stream tests.

cub/test/catch2_test_device_segmented_reduce_env_api.cu
```diff
@@ -116,3 +119,175 @@ C2H_TEST("cub::DeviceSegmentedReduce::Sum accepts not_guaranteed determinism req
...
+  REQUIRE(error == cudaSuccess);
+  REQUIRE(h_out[0].key == 0);
+  REQUIRE(h_out[0].value == 8);
+  REQUIRE(h_out[1].key == 1);
+  REQUIRE(h_out[1].value == std::numeric_limits<int>::lowest());
+  REQUIRE(h_out[2].key == 3);
+  REQUIRE(h_out[2].value == 9);
+}
+
+C2H_TEST("cub::DeviceSegmentedReduce::Reduce accepts run_to_run determinism requirements", "[segmented_reduce][env]")
```
pauleonix, line 440: This test should exist for other overloads as well

cub/test/catch2_test_device_segmented_reduce_env_api.cu
```diff
@@ -116,3 +119,175 @@ C2H_TEST("cub::DeviceSegmentedReduce::Sum accepts not_guaranteed determinism req
...
+  {
+    std::cerr << "cub::DeviceSegmentedReduce::Reduce failed with status: " << error << std::endl;
+  }
+  // example-end segmented-reduce-reduce-env
+
+  REQUIRE(d_out == expected);
+  REQUIRE(error == cudaSuccess);
+}
+
+C2H_TEST("cub::DeviceSegmentedReduce::Min env-based API", "[segmented_reduce][env]")
```
pauleonix, line 150: Is `Sum()` missing?

cub/cub/device/device_segmented_reduce.cuh
gonidelis: absolutely, pitfall during tide up

cub/cub/device/device_segmented_reduce.cuh
gonidelis: is it ok if we let it slide as a byproduct of this pr? @pauleonix

cub/cub/device/device_segmented_reduce.cuh
gonidelis: same as above. let is slide as a byproduct?

cub/cub/device/device_segmented_reduce.cuh
gonidelis: keeping it unresolved until we agree it's ok

cub/test/catch2_test_device_segmented_reduce_env.cu
gonidelis: it's needed, gives compilation error otherwise

cub/test/catch2_test_device_segmented_reduce_env.cu
gonidelis: yes it should :)

cub/cub/device/device_segmented_reduce.cuh
gonidelis: fixed

cub/cub/device/device_segmented_reduce.cuh
gonidelis: added back

cub/cub/device/device_segmented_reduce.cuh
```diff
@@ -214,6 +275,130 @@ struct DeviceSegmentedReduce
...
+  //! - When input a contiguous sequence of segments, a single sequence
+  //!   ``segment_offsets`` (of length ``num_segments + 1``) can be aliased
+  //!   for both the ``d_begin_offsets`` and ``d_end_offsets`` parameters (where
+  //!   the latter is specified as ``segment_offsets + 1``).
+  //! - Let ``s`` be in ``[0, num_segments)``. The range
+  //!   ``[d_out + d_begin_offsets[s], d_out + d_end_offsets[s])`` shall not
+  //!   overlap ``[d_in + d_begin_offsets[s], d_in + d_end_offsets[s])``,
+  //!   ``[d_begin_offsets, d_begin_offsets + num_segments)`` nor
+  //!   ``[d_end_offsets, d_end_offsets + num_segments)``.
+  //! - Can use a specific stream or cuda memory resource through the ``env`` parameter
```
pauleonix:
```suggestion
  //! - Can use a specific stream or cuda memory resource through the ``env`` parameter.
```

cub/cub/device/device_segmented_reduce.cuh
pauleonix: Sure, it does not significantly add to the review-workload. Maybe add it to the PR description.

cub/cub/device/device_segmented_reduce.cuh
gonidelis: i will choose to keep it for two reasons:
a) provides a clearer message that `if constexpr`
b) we already have it in non-env overloads from Bernhard's code

let's open a new pr to discuss separately if you insist

cub/cub/device/device_segmented_reduce.cuh
gonidelis: yes, `policy_selector_t` is the facility we use in the new dispatch layers

cub/cub/device/device_segmented_reduce.cuh
gonidelis: https://github.com/gonidelis/cccl/blob/94118f1042afc2e030a21ee3894f6d4b6f038d6c/cub/cub/device/dispatch/dispatch_segmented_reduce.cuh#L459-L460 

8-)

cub/test/catch2_test_device_segmented_reduce_env_api.cu
gonidelis: it's added already on upstream from previous pr

cub/cub/device/device_segmented_reduce.cuh
```diff
@@ -78,6 +79,67 @@ struct get_tuning_query_t
...
+    InitT initial_value,
+    EnvT env)
+  {
+    using requirements_t = ::cuda::std::execution::
+      __query_result_or_t<EnvT, ::cuda::execution::__get_requirements_t, ::cuda::std::execution::env<>>;
+    using requested_determinism_t =
+      ::cuda::std::execution::__query_result_or_t<requirements_t,
+                                                  ::cuda::execution::determinism::__get_determinism_t,
+                                                  ::cuda::execution::determinism::not_guaranteed_t,
+                                                  ::cuda::execution::determinism::run_to_run_t>;
```
bernhardmgruber, line 107: Important: I am confused. What should this code mean? This is asking for the result of querying `__get_determinism_t` with the argument `run_to_run_t` and using `not_guaranteed_t` if the query is invalid. This makes no sense to me.

cub/cub/device/device_segmented_reduce.cuh
bernhardmgruber: I think we want to refactor this at some point, but it's not so easy. For all public APIs except for `ArgMin|Max` we don't need to override `AccumT` and `OffsetT`, so we could use the same `= use_default` trick. But that does not work, since we need the replaced types for `policy_selector_from_types`.

I may address this when I pass over here again when wiring up all the tuning API exposure.

@gonidelis can you please just add a comment like:

```suggestion
          // TODO: in most cases we can just take the default AccumT and OffsetT. Refactor this
          return detail::segmented_reduce::dispatch<AccumT, OffsetT>(
```

cub/cub/device/device_segmented_reduce.cuh
bernhardmgruber: Works for me!

cub/cub/device/device_segmented_reduce.cuh
gonidelis: late nighters

PR #7798, opened 2026-02-26
Add env based `DeviceFor::*` algorithms

### LINKED ISSUE #7541: cub::DeviceFor
- [ ] Add env-based cub::DeviceFor::Bulk()
- [ ] Add env-based cub::DeviceFor::ForEach() 
- [ ] Add env-based cub::DeviceFor::ForEachN()
- [ ] Add env-based cub::DeviceFor::ForEachCopy()
- [ ] Add env-based cub::DeviceFor::ForEachCopyN()
- [ ] Add env-based cub::DeviceFor::ForEachInExtents()

gonidelis:
Adds env based `DeviceFor::*`

closed #7541 

**There are not deterministic guarantees imposed for the `DeviceFor::*` algorithms.**

### DIFF
--- a/cub/cub/device/device_for.cuh
+++ b/cub/cub/device/device_for.cuh
@@ -1290,8 +1290,8 @@ public:
   //! @return cudaError_t
   //!   error status
   _CCCL_TEMPLATE(typename LayoutMapping, typename OpType, typename EnvT = ::cuda::std::execution::env<>)
-  _CCCL_REQUIRES(::cuda::std::__is_cuda_std_layout_left_or_right_mapping_v<LayoutMapping>
-                   _CCCL_AND (!::cuda::std::is_convertible_v<EnvT, cudaStream_t>))
+  _CCCL_REQUIRES(::cuda::std::__is_cuda_std_layout_left_or_right_mapping_v<LayoutMapping> _CCCL_AND(
+    !::cuda::std::is_convertible_v<EnvT, cudaStream_t>))
   CUB_RUNTIME_FUNCTION static cudaError_t ForEachInLayout(const LayoutMapping& layout_mapping, OpType op, EnvT env = {})
   {
     using namespace cub::detail;
--- a/cub/test/catch2_test_device_for_env.cu
+++ b/cub/test/catch2_test_device_for_env.cu
@@ -145,4 +145,3 @@ C2H_TEST("DeviceFor::ForEachCopy env uses custom stream", "[for][env]")
   c2h::device_vector<int> expected_count{2};
   REQUIRE(count == expected_count);
 }
-
--- a/cub/test/catch2_test_device_segmented_scan_api.cu
+++ b/cub/test/catch2_test_device_segmented_scan_api.cu
@@ -303,9 +303,7 @@ C2H_TEST("cub::DeviceSegmentedScan::ExclusiveSegmentedScan API with two offsets 
    * over boolean values stored as bits in unsigned integer, where addition is bitwise XOR.
    * Each unsigned integer represents 32-long tuple of GF(2) values
    */
-  auto scan_op = [] __host__ __device__(unsigned v1, unsigned v2) -> unsigned {
-    return v1 ^ v2;
-  };
+  auto scan_op        = [] __host__ __device__(unsigned v1, unsigned v2) -> unsigned { return v1 ^ v2; };
   unsigned init_value = 0u;
 
   // 128 input elements
@@ -381,9 +379,7 @@ C2H_TEST("cub::DeviceSegmentedScan::InclusiveSegmentedScan API with three offset
   auto d_in  = input.begin();
   auto d_out = output.begin();
 
-  auto scan_op = [] __host__ __device__(float v1, float v2) noexcept -> float {
-    return cuda::maximum<>{}(v1, v2);
-  };
+  auto scan_op = [] __host__ __device__(float v1, float v2) noexcept -> float { return cuda::maximum<>{}(v1, v2); };
 
   void* temp_storage = nullptr;
   size_t temp_storage_bytes;


cub/cub/device/device_for.cuh
```diff
@@ -566,6 +570,63 @@ public:
...
+  //! @param[in] op
+  //!   Function object to apply to each index in the index space
+  //!
+  //! @param[in] env
+  //!   Execution environment
+  template <class ShapeT,
+            class OpT,
+            class EnvT,
+            ::cuda::std::enable_if_t<!::cuda::std::is_convertible_v<EnvT, cudaStream_t>, int> = 0>
+  [[nodiscard]] CUB_RUNTIME_FUNCTION static cudaError_t Bulk(ShapeT shape, OpT op, EnvT env)
```
bernhardmgruber: Suggestion: I think we can fully replace the overload:
```
template <class ShapeT, class OpT>
CUB_RUNTIME_FUNCTION static cudaError_t Bulk(ShapeT shape, OpT op, cudaStream_t stream = {})
```
above with the environment one here, so we don't need the `enable_if` and save some duplication.

Probably applies to all other new overloads as well.

gonidelis: love the idea

gonidelis:
One thing that might kill backwards compatibility from unifying the env overloads with the old ones is that we tend to make new overloads `[[nodiscard]]`. Adding this to the "new" env overloads (which just substitute the old) might pop compile erros to users out of the blue. @bernhardmgruber  @gevtushenko  should I avoid doing the overloads in this pr  `nodiscard`?

gonidelis:
todo: add docs about memory_resource (?)

bernhardmgruber:
> One thing that might kill backwards compatibility from unifying the env overloads with the old ones is that we tend to make new overloads `[[nodiscard]]`. Adding this to the "new" env overloads (which just substitute the old) might pop compile erros to users out of the blue. @bernhardmgruber @gevtushenko should I avoid doing the overloads in this pr `nodiscard`?

Good point. Please omit the `[[nodiscard]]` in this case.

cub/cub/device/device_for.cuh
```diff
@@ -199,7 +202,7 @@ public:
       return cudaSuccess;
     }
 
-    return Bulk(shape, op, stream);
+    return Bulk(shape, op, ::cuda::std::execution::env{::cuda::stream_ref{stream}});
```
bernhardmgruber: Important: This is not needed, since `cudaStream_t` supports the stream query for a stream. Please just pass the `stream` as it was passed before.

Applies to several other places as well.

cub/test/catch2_test_device_for_api.cu
```diff
@@ -78,7 +78,7 @@ C2H_TEST("Device bulk works without temporary storage", "[bulk][device]")
   c2h::device_vector<int> vec = {1, 2, 3, 4};
   square_t op{thrust::raw_pointer_cast(vec.data())};
 
-  cub::DeviceFor::Bulk(vec.size(), op);
+  (void) cub::DeviceFor::Bulk(vec.size(), op);
```
bernhardmgruber: Important: It seems error handling was forgotten here. Please just add a check like:

```suggestion
  REQUIRE(cub::DeviceFor::Bulk(vec.size(), op) ==cudaSuccess);
```
Applies more times below.

cub/test/catch2_test_device_for_env.cu
```diff
@@ -0,0 +1,215 @@
...
+{
+  auto vec = c2h::device_vector<int>{};
+  square_ref_op op{};
+
+  cuda::stream stream{cuda::devices[0]};
+  auto env = cuda::std::execution::env{cuda::stream_ref{stream}};
+
+  auto error = cub::DeviceFor::ForEachN(vec.begin(), 0, op, env);
+  REQUIRE(error == cudaSuccess);
+}
```
bernhardmgruber: Suggestion: I don't think we need to test empty inputs. The new tests we add in this PR should cover environment functionality and I think this is already covered by the test above.

I think we can drop every other test here in this file.

cub/test/catch2_test_device_for_env_api.cu
```diff
@@ -0,0 +1,171 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include "insert_nested_NVTX_range_guard.h"
+// above header needs to be included first
```
bernhardmgruber: Suggestion: This comment is not needed AFAIK, because clang-format handles this
```suggestion
```

cub/test/catch2_test_device_for_env_api.cu
```diff
@@ -0,0 +1,171 @@
...
+    {
+      atomicAdd(d_count, 1);
+    }
+  }
+};
+// example-end odd-count-env-t
+
+C2H_TEST("cub::DeviceFor::Bulk env-based API", "[for][env]")
+{
+  // example-begin bulk-env
```
bernhardmgruber, line 57: Important: I think those examples are not yet referred to from the documentation.

cub/test/catch2_test_device_for_api.cu
gonidelis: these are API tests that are being used as reference examples in literalincludes later on in the device headers. We should be verbal with the return code check and not use the c2h `REQUIRE` thingy inline.

gonidelis:
Since we replaced the `cudaStream_t stream` parameter with `EnvT env`, we broke backwards compatibility and callers that relied on implicit conversion from `nvbench::cuda_stream` to `cudaStream_t`. The `extents` benchmark showcases that. I see three options:

a) We go back to **only adding** env overloads (not converting old ones) for `DeviceFor`
b) We check internally if passed 3rd argument (`env`) is `nvbench::cuda_stream` and do the conversion explicitly inside the body
c) We break, and ask for people that were relying on implicit conversion to explicitly `.get_stream()`

bernhardmgruber:
> Since we replaced the `cudaStream_t stream` parameter with `EnvT env`, we broke backwards compatibility and callers that relied on implicit conversion from `nvbench::cuda_stream` to `cudaStream_t`. The `extents` benchmark showcases that.

That's interesting, because we hit the same case in `cub::DeviceTransform` with RMM's stream type. We solved this in #7266 and added tests in #7278. What we apparently missed are types that are convertible to `cudaStream_t` and are not copyable, because then you can't pass them to the environment parameter.

Let me ask @miscco

bernhardmgruber:
> Let me ask @miscco

Started an internal Slack thread.

In the meantime, I came up with a fix and pushed a commit.

cub/cub/device/device_for.cuh
```diff
@@ -1050,15 +1278,21 @@ public:
...
   //!
   //! @return cudaError_t
   //!   error status
-  _CCCL_TEMPLATE(typename LayoutMapping, typename OpType)
-  _CCCL_REQUIRES(::cuda::std::__is_cuda_std_layout_left_or_right_mapping_v<LayoutMapping>)
-  [[nodiscard]] CUB_RUNTIME_FUNCTION static cudaError_t
-  ForEachInLayout(const LayoutMapping& layout_mapping, OpType op, cudaStream_t stream = {})
+  _CCCL_TEMPLATE(typename LayoutMapping, typename OpType, typename EnvT = ::cuda::std::execution::env<>)
+  _CCCL_REQUIRES(::cuda::std::__is_cuda_std_layout_left_or_right_mapping_v<LayoutMapping>
+                   _CCCL_AND !::cuda::std::is_convertible_v<EnvT, cudaStream_t>)
```
bernhardmgruber:
```suggestion
                   _CCCL_AND (!::cuda::std::is_convertible_v<EnvT, cudaStream_t>))
```

miscco: yes concepts requires ! to be wrapped in braces

bernhardmgruber:
pre-commit.ci autofix

cub/cub/device/device_for.cuh
```diff
@@ -926,6 +1127,10 @@ public:
   //! .. versionadded:: 2.4.0
   //!    First appears in CUDA Toolkit 12.5.
```
gonidelis, line 1128: 😡😡😡😡😡

PR #7808, opened 2026-02-26
Implement `op_sat_overflow` arithmetic

davebayer:
Follow up of #3449

### DIFF
--- a/docs/libcudacxx/extended_api/numeric/add_sat_overflow.rst
+++ b/docs/libcudacxx/extended_api/numeric/add_sat_overflow.rst
@@ -37,11 +37,11 @@ The function ``cuda::add_sat_overflow`` performs saturating addition of two valu
 
 **Constraints**
 
-- ``T`` must be an integer type.
+- ``T`` must be an `integer type <https://eel.is/c++draft/basic.fundamental#1>`_.
 
 **Performance considerations**
 
-- Functionality is implemented by correcting the ``cuda::add_overflow`` result.
+- Functionality is implemented by correcting the ``cuda::add_overflow`` result in case of overflow/underflow.
 - Unsigned computations are generally faster than signed computations.
 
 Example
--- a/docs/libcudacxx/extended_api/numeric/div_sat_overflow.rst
+++ b/docs/libcudacxx/extended_api/numeric/div_sat_overflow.rst
@@ -41,11 +41,11 @@ The function ``cuda::div_sat_overflow`` performs saturating integer division of 
 
 **Constraints**
 
-- ``T`` must be an integer type.
+- ``T`` must be an `integer type <https://eel.is/c++draft/basic.fundamental#1>`_.
 
 **Performance considerations**
 
-- Functionality is implemented by correcting the ``cuda::div_overflow`` result.
+- Functionality is implemented by correcting the ``cuda::div_overflow`` result in case of overflow/underflow.
 - Unsigned computations are generally faster than signed computations.
 
 Example
--- a/docs/libcudacxx/extended_api/numeric/add_overflow.rst
+++ b/docs/libcudacxx/extended_api/numeric/add_overflow.rst
@@ -37,7 +37,7 @@ The function ``cuda::add_overflow`` performs addition of two values ``lhs`` and 
 
 **Constraints**
 
-- ``Result``, ``Lhs``, and ``Rhs`` must be integer types.
+- ``Result``, ``Lhs``, and ``Rhs`` must be `integer types <https://eel.is/c++draft/basic.fundamental#1>`_.
 
 **Performance considerations**
 
--- a/docs/libcudacxx/extended_api/numeric/div_overflow.rst
+++ b/docs/libcudacxx/extended_api/numeric/div_overflow.rst
@@ -41,7 +41,7 @@ The function ``cuda::div_overflow`` performs integer division of ``lhs`` by ``rh
 
 **Constraints**
 
-- ``Result``, ``Lhs``, and ``Rhs`` must be integer types.
+- ``Result``, ``Lhs``, and ``Rhs`` must be `integer types <https://eel.is/c++draft/basic.fundamental#1>`_.
 
 **Notes**
 
--- a/docs/libcudacxx/extended_api/numeric/mul_overflow.rst
+++ b/docs/libcudacxx/extended_api/numeric/mul_overflow.rst
@@ -37,7 +37,7 @@ The function ``cuda::mul_overflow`` performs multiplication of two values ``lhs`
 
 **Constraints**
 
-- ``Result``, ``Lhs``, and ``Rhs`` must be integer types.
+- ``Result``, ``Lhs``, and ``Rhs`` must be `integer types <https://eel.is/c++draft/basic.fundamental#1>`_.
 
 **Performance considerations**
 
--- a/docs/libcudacxx/extended_api/numeric/mul_sat_overflow.rst
+++ b/docs/libcudacxx/extended_api/numeric/mul_sat_overflow.rst
@@ -37,11 +37,11 @@ The function ``cuda::mul_sat_overflow`` performs saturating multiplication of tw
 
 **Constraints**
 
-- ``T`` must be an integer type.
+- ``T`` must be an `integer type <https://eel.is/c++draft/basic.fundamental#1>`_.
 
 **Performance considerations**
 
-- Functionality is implemented by correcting the ``cuda::mul_overflow`` result.
+- Functionality is implemented by correcting the ``cuda::mul_overflow`` result in case of overflow/underflow.
 - Unsigned computations are generally faster than signed computations.
 
 Example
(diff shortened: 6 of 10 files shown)


docs/libcudacxx/extended_api/numeric/add_sat_overflow.rst
```diff
@@ -0,0 +1,78 @@
...
+1. Returns an :ref:`overflow_result <libcudacxx-extended-api-numeric-overflow_result>` object containing the result of the saturating addition and a boolean indicating whether an overflow or underflow occurred.
+2. Returns ``true`` if an overflow or underflow occurred, ``false`` otherwise.
+
+**Constraints**
+
+- ``T`` must be an integer type.
+
+**Performance considerations**
+
+- Functionality is implemented by correcting the ``cuda::add_overflow`` result.
```
miscco: I am not sure that `correcting` is the proper word, but I dont know how to better phrase it 

Maybe add `in case of overflow`

docs/libcudacxx/extended_api/numeric.rst
```diff
@@ -33,6 +33,11 @@ Numeric
      - CCCL 3.0.0
      - CUDA 13.0
 
+   * - :ref:`cuda::overflow_cast <libcudacxx-extended-api-numeric-overflow_cast>`
+     - Casts a value with overflow checking
+     - CCCL 3.0.0
```
miscco, line 38:
```suggestion
     - CCCL 3.4.0
```

libcudacxx/include/cuda/__numeric/add_sat_overflow.h
```diff
@@ -0,0 +1,67 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES.
+//
+//===----------------------------------------------------------------------===//
```
miscco, line 8: Nitpick: I believe this should all be the libcu++ license

docs/libcudacxx/extended_api/numeric/div_sat_overflow.rst
```diff
@@ -0,0 +1,83 @@
...
+1. Returns an :ref:`overflow_result <libcudacxx-extended-api-numeric-overflow_result>` object containing the result of saturating integer division and a boolean flag indicating whether an overflow or underflow occurred.
+2. Returns ``true`` if an overflow or underflow occurred.
+
+**Preconditions**
+
+- ``rhs != 0``
+
+**Constraints**
+
+- ``T`` must be an integer type.
```
miscco: Critial: This might be confusing with `is_integral` I would change all those to:

```suggestion
- ``T`` must be a `standard integer type. <https://eel.is/c++draft/basic.fundamental#1>`__
```

docs/libcudacxx/extended_api/numeric.rst
davebayer: No, this is correct, we were missing docs for this function

libcudacxx/include/cuda/__numeric/add_sat_overflow.h
davebayer: No, these files are based on libc++'s test files

docs/libcudacxx/extended_api/numeric/div_sat_overflow.rst
davebayer: That's not quite correct, because we support `__int128` which is not a standard integer type. But I will add the link to cppreference.

miscco: Oh yeah It should probably be `extended standard integer types`

docs/libcudacxx/extended_api/numeric/add_sat_overflow.rst
```diff
@@ -0,0 +1,78 @@
...
+
+    #include <cuda/numeric>
+    #include <cuda/std/cassert>
+    #include <cuda/std/limits>
+
+    __global__ void kernel()
+    {
+        constexpr auto int_max = cuda::std::numeric_limits<int>::max();
+
+        const auto result = cuda::add_sat_overflow(1, int_max); // saturated
```
fbusato, line 60: we could add an example of underflow to show both behaviors

libcudacxx/include/cuda/__numeric/mul_sat_overflow.h
```diff
@@ -0,0 +1,67 @@
...
+_CCCL_REQUIRES(::cuda::std::__cccl_is_integer_v<_Tp>)
+[[nodiscard]] _CCCL_API constexpr overflow_result<_Tp> mul_sat_overflow(_Tp __x, _Tp __y) noexcept
+{
+  auto __result = ::cuda::mul_overflow(__x, __y);
+  if (__result.overflow)
+  {
+    if constexpr (::cuda::std::is_signed_v<_Tp>)
+    {
+      __result.value =
+        ((__x < 0) == (__y < 0)) ? ::cuda::std::numeric_limits<_Tp>::max() : ::cuda::std::numeric_limits<_Tp>::min();
```
fbusato, line 44: a more optimized code is `(x & y) < 0` (with `unsigned`)

libcudacxx/include/cuda/std/__numeric/mul_sat.h
```diff
@@ -35,20 +32,7 @@ _CCCL_TEMPLATE(class _Tp)
...
-    {
-      return ((__x < 0) == (__y < 0)) ? numeric_limits<_Tp>::max() : numeric_limits<_Tp>::min();
-    }
-    else
-    {
-      return numeric_limits<_Tp>::max();
-    }
-  }
-  return __result;
+  return ::cuda::mul_sat_overflow(__x, __y).value;
```
fbusato, line 35: I don't think it is the case. We should check if there is any change in the generated code

libcudacxx/test/libcudacxx/cuda/numeric/sat.overflow.arithmetic/add_sat_overflow.pass.cpp
```diff
@@ -0,0 +1,183 @@
...
+  test_add_sat_overflow<I>(I{-1}, minVal, minVal, true, zero_value); // saturated
+  test_add_sat_overflow<I>(I{-1}, maxVal, I{-1} + maxVal, false, zero_value);
+
+  test_add_sat_overflow<I>(I{0}, I{-1}, I{-1}, false, zero_value);
+  test_add_sat_overflow<I>(I{0}, I{0}, I{0}, false, zero_value);
+  test_add_sat_overflow<I>(I{0}, I{1}, I{1}, false, zero_value);
+  test_add_sat_overflow<I>(I{0}, minVal, minVal, false, zero_value);
+  test_add_sat_overflow<I>(I{0}, maxVal, maxVal, false, zero_value);
+
+  test_add_sat_overflow<I>(I{1}, I{-1}, I{0}, false, zero_value);
```
fbusato, line 66: I would prefer testing `a + b` and `b + a` instead of manually repeating test cases

libcudacxx/test/libcudacxx/cuda/numeric/sat.overflow.arithmetic/saturate_overflow_cast.pass.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+// template<class R, class T>
+// constexpr overflow_result<R> saturate_overflow_cast(T x) noexcept;                     // freestanding
+
+#include <cuda/numeric>
+#include <cuda/std/cassert>
+#include <cuda/std/limits>
+#include <cuda/std/type_traits>
+
+using IMIN = signed char;
```
fbusato, line 20: I understand that these are tests, but would be better to use `imin_t` and similar

fbusato (approved):

libcudacxx/include/cuda/__numeric/mul_sat_overflow.h
davebayer: It would have to be `^` with the values in the ternary swapped.

Anyway I would rather keep it like this. It's more readable. And it keeps the full picture for the optimizer, too, even though it currently doesn't produce better optimized code.

libcudacxx/test/libcudacxx/cuda/numeric/sat.overflow.arithmetic/add_sat_overflow.pass.cpp
davebayer: But then you need to track which combinations you already tried and which you didn't. Also, this structure would be different for non-commutative operations like - or /. I would keep it as is

libcudacxx/test/libcudacxx/cuda/numeric/sat.overflow.arithmetic/saturate_overflow_cast.pass.cpp
davebayer: I don't think it matters too much, we use `T` and `U` anyway. I can fix both sat and sat + overflow arithmetic tests in a follow up PR.

Issue #7464, opened 2026-02-03
Consistently use the new CUB tuning APIs internally


bernhardmgruber:
After the new tuning API has been (internally) implemented for a CUB algorithm and the correspond env-overloads are available, we can **replace all internal dispatch layer access by public CUB API calls**. This should mostly affect any remaining benchmarks or tests that still use a (legacy) policy_hub or target a (legacy) Dispatcher struct directly. The big chunk of this work will be benchmarks, but there are also some tests and Thrust calls to CUB. This will also ensure that all CUB algorithms correctly extract an eventual tuning from their environment.

 - [ ] Use the new tuning API internally for DeviceRadixSort
 - [ ] Use the new tuning API internally for DeviceReduce
 - [ ] Use the new tuning API internally for DeviceSegmentedReduce
 - [ ] Use the new tuning API internally for DeviceTransform
 - [ ] Use the new tuning API internally for nondeterministic reduce
 - [ ] Use the new tuning API internally for DispatchAdjacentDifference
 - [ ] Use the new tuning API internally for detail::for_each::dispatch_t
 - [ ] Use the new tuning API internally for DeviceRleDispatch
 - [ ] Use the new tuning API internally for detail::merge::dispatch_t
 - [ ] Use the new tuning API internally for deterministic (rfa) reduce dispatch
 - [ ] Use the new tuning API internally for DispatchReduceByKey
 - [ ] Use the new tuning API internally for DispatchStreamingReduceByKey
 - [ ] Use the new tuning API internally for detail::reduce::dispatch_streaming_arg_reduce_t
 - [ ] Use the new tuning API internally for DispatchBatchMemcpy
 - [ ] Use the new tuning API internally for DispatchHistogram
 - [ ] Use the new tuning API internally for DispatchMergeSort
 - [ ] Use the new tuning API internally for DeviceScan
 - [ ] Use the new tuning API internally for DispatchScanByKey
 - [ ] Use the new tuning API internally for DispatchSegmentedRadixSort
 - [ ] Use the new tuning API internally for DispatchFixedSizeSegmentedReduce
 - [ ] Use the new tuning API internally for DispatchSegmentedSort
 - [ ] Use the new tuning API internally for DispatchSelectIf
 - [ ] Use the new tuning API internally for DispatchThreeWayPartitionIf
 - [ ] Use the new tuning API internally for DispatchTopK
 - [ ] Use the new tuning API internally for DispatchUniqueByKey

[...]

PR #7810, opened 2026-02-26
Use the new tuning API internally for `detail::transform::dispatch`

bernhardmgruber:
- [x] Merge before: #7809 

Part of #7464

### DIFF
--- a/cub/benchmarks/bench/transform/common.h
+++ b/cub/benchmarks/bench/transform/common.h
@@ -89,9 +89,9 @@ void bench_transform(nvbench::state& state,
       output,
       num_items,
       transform_op,
-      cuda::std::execution::env{launch.get_stream().get_stream()
+      cuda::std::execution::env{::cuda::stream_ref{launch.get_stream().get_stream()}
 #if !TUNE_BASE
-                                  ,
+                                ,
                                 cuda::execution::__tune(policy_selector{})
 #endif // !TUNE_BASE
       });
--- a/cub/test/catch2_test_device_transform_env.cu
+++ b/cub/test/catch2_test_device_transform_env.cu
@@ -72,6 +72,10 @@ void check_graph_nodes_with_different_streams(F call_cub_api)
   {
     call_cub_api(with_get_stream_method{stream.get()});
   }
+  SECTION("environment with cuda::stream_ref")
+  {
+    call_cub_api(cuda::std::execution::env{cuda::stream_ref{stream}});
+  }
   SECTION("environment with prop with cudaStream_t")
   {
     // MSVC has trouble nesting two aggregate initializations with CTAD


cub/benchmarks/bench/transform/common.h
```diff
@@ -72,17 +73,16 @@ void bench_transform(nvbench::state& state,
...
-      launch.get_stream()
+      cuda::std::execution::env{launch.get_stream().get_stream()
 #if !TUNE_BASE
-        ,
-      policy_selector{}
+                                  ,
+                                cuda::execution::__tune(policy_selector{})
 #endif // !TUNE_BASE
-    );
+      });
```
bernhardmgruber: Q: Is this how we should correctly compose a stream and a tuning in an environment?

cub/cub/device/device_transform.cuh
```diff
@@ -48,8 +48,22 @@ struct ::cuda::proclaims_copyable_arguments<CUB_NS_QUALIFIER::detail::__return_c
...
+#if _CCCL_HAS_CONCEPTS()
+  requires transform_policy_selector<PolicySelector>
+#endif // _CCCL_HAS_CONCEPTS()
+struct tuning
+{
+  [[nodiscard]] _CCCL_TRIVIAL_API constexpr auto query(const get_tuning_query_t&) const noexcept -> PolicySelector
+  {
+    return static_cast<const PolicySelector&>(*this);
+  }
+};
```
bernhardmgruber, line 67: Suggestion: Because a policy selector for DeviceTransform needs to fulfill a concept anyhow, I wonder whether we should rather implement `get_tuning_query_t` more intelligently to look for a type inside the tuning environment that fulfills the concept. This way, users can avoid inheriting from `tuning<..>`.

cub/test/catch2_test_device_transform_env.cu
```diff
@@ -218,3 +218,47 @@ C2H_TEST("DeviceTransform::TransformStableArgumentAddresses custom stream", "[de
 
   CHECK(thrust::equal(result.begin(), result.end(), cuda::counting_iterator<type>{42 + 13}));
 }
+
+// use a policy selector that prescribes to run with exactly 8 threads per block and 3 items per thread
+// TODO(bgruber): can we get by without the base class?
+struct my_policy_selector : cub::detail::transform::tuning<my_policy_selector>
```
bernhardmgruber, line 228: ok, I get this error here:
```
   224 | struct my_policy_selector : cub::detail::transform::tuning<my_policy_selector>
        |                                                     ^~~~~~~~~~~~~~~~~~~~~~~~~~
  /home/coder/cccl/lib/cmake/cub/../../../cub/cub/device/device_transform.cuh:58:12: note: because 'my_policy_selector' does not satisfy 'transform_policy_selector'
     58 |   requires transform_policy_selector<PolicySelector>
        |            ^
  /home/coder/cccl/lib/cmake/cub/../../../cub/cub/device/dispatch/tuning/tuning_transform.cuh:202:37: note: because 'policy_selector<my_policy_selector, transform_policy>' evaluated to false
    202 | concept transform_policy_selector = policy_selector<T, transform_policy>;
        |                                     ^
  /home/coder/cccl/lib/cmake/cub/../../../cub/cub/util_device.cuh:900:5: note: because 'pol_sel(arch)' would be invalid: incomplete type in call to object of type 'my_policy_selector'
    900 |   { pol_sel(arch) } -> _CCCL_CONCEPT_VSTD::same_as<Policy>;
        |     ^
```
and I wonder how this can be solved. `my_policy_selector` is incomplete when passed to `cub::detail::transform::tuning` so the error is correct. So how should `cub::detail::transform::tuning` validate a concept on its template argument?

cub/cub/device/device_transform.cuh
bernhardmgruber: Will pursue this in a follow-up PR.

cub/test/catch2_test_device_transform_env.cu
bernhardmgruber: I have an idea for a follow-up PR.

cub/benchmarks/bench/transform/common.h
gevtushenko: yes

bernhardmgruber: I encountered a weird case and revisited this again now.

This works:
```c++
  cuda::stream_ref s = cudaStreamPerThread;
  cuda::std::execution::env env{s};
  cuda::stream_ref r = cuda::get_stream(env);
```
This does not (fails to compile):
```c++
  cudaStream_t s = cudaStreamPerThread;
  cuda::std::execution::env env{s};
  cudaStream_t r = cuda::get_stream(env);
```
I assume the reason is that `cuda::stream_ref` has the corresponding `query` member function. So we have to be careful. `launch.get_stream().get_stream()` returns a `cudaStream_t`, so the current code is wrong. We have to wrap `cuda::stream_ref{launch.get_stream().get_stream()}`.

PR #7811, opened 2026-02-26
Change mechanism for determining the right one-worker-per-segment policy in `DeviceBatchedTopK`

elstehle:
## Description

This PR updates the logic to determine the right policy for the one-worker-per-segment approach in `DeviceBatchedTopK`. Previously, we were selecting the first policy from the list of potential policies for which the corresponding agent would not exceed the available shared memory. Usually, these would be able to cover very large segments and, hence, have very large tile sizes. After this change, we select the policy that policy with the smallest tile size still able to cover the provided static upper bound for the segments that the user has provided us with.

### DIFF
--- a/cub/cub/device/dispatch/kernels/kernel_batched_topk.cuh
+++ b/cub/cub/device/dispatch/kernels/kernel_batched_topk.cuh
@@ -107,7 +107,7 @@ struct find_smallest_covering_policy_impl
                                                                                           // one
                                >;
 
-  static constexpr bool has_valid_policy = next_step::has_valid_policy;
+  static constexpr bool has_valid_policy = covers;
 };
 
 // Base case: end of tuple
--- a/cub/cub/device/dispatch/tuning/tuning_batched_topk.cuh
+++ b/cub/cub/device/dispatch/tuning/tuning_batched_topk.cuh
@@ -33,7 +33,7 @@ struct policy_hub
     // within available shared memory limits. The first policy that fits SMEM is taken. Policies must be ordered by
     // decreasing segment size.
     // TODO (elstehle): Consider making this a static constexpr array of a simple struct and implementing
-    // find_first_smem_fitting_policy_impl as a constexpr function.
+    // find_smallest_covering_policy as a constexpr function.
     using worker_per_segment_policies =
       ::cuda::std::tuple<agent_batched_topk_worker_per_segment_policy<256, 64, default_load_alg, default_store_alg>,
                          agent_batched_topk_worker_per_segment_policy<256, 32, default_load_alg, default_store_alg>,


cub/cub/device/dispatch/kernels/kernel_batched_topk.cuh
```diff
@@ -76,27 +79,68 @@ struct find_valid_policy_impl
   static constexpr bool has_valid_policy = fits ? true : next_step::has_valid_policy;
 };
 
-template <typename SegmentedTopKPolicy, template <typename...> class WorkerPerSegmentAgentT, typename... AgentParamsT>
+// Finds the last policy in a tuple of policies that can be instantiated within the shared memory limits
+// I.e., we want to find a policy for the one-worker-per-segment approach that is optimized for the segment size
+template <typename PoliciesT, ::cuda::std::int64_t Index, ::cuda::std::int64_t Count, int MaxSegmentSize>
+struct find_tightest_policy_impl
```
gevtushenko: suggestion: it's not exactly obvious from the names that the old one is searching for the first shared memory constrain match in the list and this one finds for the last tile size match in the list. Not necesserily for this PR, but we can consider finding names that reflect that better.

cub/cub/device/dispatch/kernels/kernel_batched_topk.cuh
```diff
@@ -76,27 +79,68 @@ struct find_valid_policy_impl
...
+  // If the current policy covers the segment, we check if the next one also still does.
+  // We want the last one that returns true for 'covers'.
+  using policy_t =
+    ::cuda::std::conditional_t<(covers && next_step::has_valid_policy),
+                               typename next_step::policy_t, // The next one is even tighter and still works
+                               ::cuda::std::conditional_t<covers, current_policy_t, void> // This is the tightest valid
+                                                                                          // one
+                               >;
+
+  static constexpr bool has_valid_policy = covers || next_step::has_valid_policy;
```
gevtushenko: suggestion: it's not clear when current policy doesn't cover for what next policy does. It's either hiding a bug or is unnecessary, please take a look.

cub/cub/device/dispatch/kernels/kernel_batched_topk.cuh
```diff
@@ -76,27 +79,68 @@ struct find_valid_policy_impl
...
+  static constexpr bool covers = (tile_size >= MaxSegmentSize);
+
+  // Peek at the next policy
+  using next_step = find_tightest_policy_impl<PoliciesT, Index + 1, Count, MaxSegmentSize>;
+
+  // Selection Logic:
+  // If the current policy covers the segment, we check if the next one also still does.
+  // We want the last one that returns true for 'covers'.
+  using policy_t =
+    ::cuda::std::conditional_t<(covers && next_step::has_valid_policy),
```
gevtushenko: suggestion: consider just
```suggestion
    ::cuda::std::conditional_t<next_step::has_valid_policy,
```

PR #7817, opened 2026-02-26
Avoid integer related warnings in CUB

miscco:
Working on pstl I got another warning about signed unsigned comparisons.

Also we should cast from `int` to OffsetType to avoid similar warnings

### DIFF
--- a/cub/cub/device/dispatch/dispatch_scan_by_key.cuh
+++ b/cub/cub/device/dispatch/dispatch_scan_by_key.cuh
@@ -166,7 +166,7 @@ CUB_DETAIL_KERNEL_ATTRIBUTES void DeviceScanByKeyInitKernel(
 
   const int tid           = blockDim.x * blockIdx.x + threadIdx.x;
   const OffsetT tile_base = static_cast<OffsetT>(tid) * items_per_tile;
-  if (tid < num_tiles)
+  if (tid > 0 && tid < num_tiles)
   {
     d_keys_prev_in[tid] = d_keys_in[tile_base - 1];
   }


cub/cub/device/dispatch/dispatch_scan_by_key.cuh
```diff
@@ -164,10 +164,9 @@ CUB_DETAIL_KERNEL_ATTRIBUTES void DeviceScanByKeyInitKernel(
   // Initialize tile status
   tile_state.InitializeStatus(num_tiles);
 
-  const unsigned tid      = threadIdx.x + blockDim.x * blockIdx.x;
+  const int tid           = blockDim.x * blockIdx.x + threadIdx.x;
   const OffsetT tile_base = static_cast<OffsetT>(tid) * items_per_tile;
-
-  if (tid > 0 && tid < num_tiles)
+  if (tid < num_tiles)
```
bernhardmgruber: In principle, `int` could be too small to hold `blockDim.x * blockIdx.x + threadIdx.x`. If we stayed with `unsigned`, then `num_tiles` (which is an `int`) would be promoted to `unsigned` (which should always be correct) and the comparison would fail correctly for any `tid` larger than `INT_MAX`.

If `tid` is an `int`, and `blockDim.x * blockIdx.x + threadIdx.x` is larger than `INT_MAX`, we could get a negative value that wrongly passes the check `tid < num_tiles`.

I therefore argue that we should keep `tid` as `unsigned` and make the comparison `tid < static_cast<unsigned>(num_tiles))`

cub/cub/device/dispatch/kernels/kernel_merge_sort.cuh
```diff
@@ -231,8 +231,7 @@ CUB_DETAIL_KERNEL_ATTRIBUTES void DeviceMergeSortPartitionKernel(
   OffsetT target_merged_tiles_number,
   int items_per_tile)
 {
-  OffsetT partition_idx = blockDim.x * blockIdx.x + threadIdx.x;
-
+  const OffsetT partition_idx = static_cast<OffsetT>(blockDim.x * blockIdx.x + threadIdx.x);
```
NaderAlAwar, line 234: Suggestion: for this and the other occurrence, it might be better to cast just one instead to prevent overflow
```suggestion
  const OffsetT partition_idx = static_cast<OffsetT>(blockDim.x) * blockIdx.x + threadIdx.x;
```

cub/cub/device/dispatch/dispatch_scan_by_key.cuh
miscco: We have a ton of places where this happens like this, I believe we should either use features such as `cuda::thread.rank(cuda::grid)` or make a concerted push to fix it globally.

Not that the [CUDA programming](https://docs.nvidia.com/cuda/cuda-programming-guide/02-basics/writing-cuda-kernels.html) guide writes:

```cpp
__global__ void vecAdd(float* A, float* B, float* C, int vectorLength)
{
    int workIndex = threadIdx.x + blockIdx.x*blockDim.x;
```

cub/cub/device/dispatch/kernels/kernel_merge_sort.cuh
miscco: That would have no effect. The multiplication would still promote

cub/cub/device/dispatch/dispatch_scan_by_key.cuh
bernhardmgruber: I am aware.

bernhardmgruber (approved):

cub/cub/device/dispatch/dispatch_scan_by_key.cuh
miscco: Added the tid > 0 to avoid overflows

elstehle (approved):

cub/cub/device/dispatch/kernels/kernel_merge_sort.cuh
davebayer: @NaderAlAwar is right here, `unsigned` needn't to be enough to store the desired size

cub/cub/device/dispatch/dispatch_scan_by_key.cuh
davebayer: Note that using `cuda::thread.rank(cuda::grid)` without the hierarchy parameter can be super slow, because we cannot assume that the block and grid are 1D only.

Also, computing thread rank in grid returns 64-bit unsigned int by default, which may introduce another overhead.

PR #7819, opened 2026-02-26
Enable launch dims assumptions for `clang` and `nvc++`

davebayer:
Actually I found out that we should be fine using the assumptions with `nvc++` in CUDA mode, because the dimensions don't depend on the target architecture.

I've also retested clang-cuda and it seems to be working fine. Maybe there are some circumstances under which the assumption is ignored, so I've locally suppressed the `-Wassume` warning.

### DIFF
--- a/libcudacxx/include/cuda/__launch/launch.h
+++ b/libcudacxx/include/cuda/__launch/launch.h
@@ -105,6 +105,9 @@ _CCCL_BEGIN_NAMESPACE_CUDA
 #      define _CCCL_GRID_DIM_Z                     gridDim.z
 #    endif // ^^^ !_CCCL_CUDA_COMPILER(CLANG) ^^^
 
+// clang-cuda sometimes warns about the assumption being ignored because it contains (potential) side-effects. We can
+// just suppress it, because in the worst case, the assumption will be just ignored.
+// note(dabayer): I haven't found out when exactly the assumption fails.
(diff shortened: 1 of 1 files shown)


libcudacxx/include/cuda/__launch/launch.h
```diff
@@ -48,64 +48,65 @@ _CCCL_BEGIN_NAMESPACE_CUDA
...
+#      define _CCCL_CLUSTER_IDX_Z                  ::__clusterIdx().z
+#      define _CCCL_CLUSTER_RELATIVE_BLOCK_RANK    ::__clusterRelativeBlockRank()
+#      define _CCCL_CLUSTER_SIZE_IN_BLOCKS         ::__clusterSizeInBlocks()
+#      define _CCCL_GRID_DIM_X                     gridDim.x
+#      define _CCCL_GRID_DIM_Y                     gridDim.y
+#      define _CCCL_GRID_DIM_Z                     gridDim.z
+#    endif // ^^^ !_CCCL_CUDA_COMPILER(CLANG) ^^^
+
+_CCCL_DIAG_PUSH
+_CCCL_DIAG_SUPPRESS_CLANG("-Wassume")
```
miscco, line 112: What is that warning about?

davebayer: I added a comment

PR #7822, opened 2026-02-26
Use `__block_size__` attribute in `cuda::launch`

davebayer:
This PR adds a new kernel functor launcher with `__block_size__` annotation.


leofang:
Q: is there a way we can test it works as intended?
- generate correct device code 
- give performance benefit

jrhemstad:
Agreed with @leofang. I'd like to find some way to add tests that this is actually doing what we want it to.

PR #7823, opened 2026-02-27
Optimized Device-to-Device Tensor Copy (`cudax`)

fbusato:
## Description

Provide an optimized version of **device-to-device copy between two multi-dimensional tensors** with compatible extents and arbitrary strides.
~The feature is experimentally based on cuTe (CUTLASS). We need to evaluate if it is possible to remove such dependency without reimplementing cuTe in CCCL before production.~  The functionality has been reimplemented without CUTLASS/cuTe.

The code contains the following optimizations:

1. Fast path for "empty" tensors and scalar.
2. Remove 1-size strides.
3. Stride sorting.
4. Optimize negative strides.
5. Tensor dimensions coalescing.
6. Dispatch contiguous tensors to `cub::DeviceTransform`.
7. Vectorization.
8. Tiling.
9. Common kernel optimizations: `__restrict__`, `__grid_constant__`.

To explore:

- Shared memory/TMA for tensor "transpose" use case.

The PR has been rebased from https://github.com/NVIDIA/cccl/pull/7676 (prerequisite) \.

The PR contains:

- [x] Execution code.
- [x] Unit tests: basic, edge cases, cuTe vectorization, from nvmath.
- [x] Benchmarks.
- [x] documentation.

Requires #8095

### DIFF
--- a/cudax/include/cuda/experimental/__copy/copy_contiguous.cuh
+++ b/cudax/include/cuda/experimental/__copy/copy_contiguous.cuh
@@ -26,6 +26,7 @@
 #include <cuda/__cmath/ceil_div.h>
 #include <cuda/__device/all_devices.h>
 #include <cuda/__device/arch_id.h>
+#include <cuda/__device/arch_traits.h>
 #include <cuda/__launch/configuration.h>
 #include <cuda/__launch/launch.h>
 #include <cuda/__stream/stream_ref.h>
@@ -195,8 +196,13 @@ _CCCL_HOST_API void __launch_copy_contiguous_kernel(
     const auto __inner_size      = __src.__extents[0];
     const auto __outer_size      = ::cuda::experimental::__total_size(__src) / __inner_size;
     const auto __num_inner_tiles = ::cuda::ceil_div(__inner_size, __tile_size);
-    const auto __config =
-      ::cuda::make_config(::cuda::block_dims<__block_size>(), ::cuda::grid_dims(dim3(__num_inner_tiles, __outer_size)));
+    constexpr auto __arch_limits = ::cuda::__common_arch_traits(::cuda::arch_id::sm_90);
+    _CCCL_ASSERT(__num_inner_tiles <= _ExtentT(__arch_limits.max_grid_dim_x),
+                 "grid x-dimension exceeds the maximum grid size");
+    _CCCL_ASSERT(__outer_size <= _ExtentT(__arch_limits.max_grid_dim_y),
+                 "grid y-dimension exceeds the maximum grid size");
+    const auto __grid_dims = ::dim3(static_cast<unsigned>(__num_inner_tiles), static_cast<unsigned>(__outer_size));
+    const auto __config    = ::cuda::make_config(::cuda::block_dims<__block_size>(), ::cuda::grid_dims(__grid_dims));
 
     const __tensor_coord_iterator<_ExtentT, _Rank> __coord_iter{__src.__extents};
     const auto __kernel = ::cuda::experimental::__copy_contiguous_kernel<


cudax/benchmarks/bench/copy/copy_bench.cu
```diff
@@ -0,0 +1,266 @@
...
+
+#include <cstdlib>
+
+#include <nvbench/nvbench.cuh>
+
+// GCC -Warray-bounds false positive for high-rank (20+) __raw_tensor instantiations
+#if _CCCL_COMPILER(GCC)
+_CCCL_DIAG_PUSH
+_CCCL_DIAG_SUPPRESS_GCC("-Warray-bounds")
+#endif // _CCCL_COMPILER(GCC)
```
miscco: Given that this is a test we can simplify to
```suggestion
_CCCL_DIAG_SUPPRESS_GCC("-Warray-bounds")
```

No need to restore

cudax/include/cuda/experimental/__copy/copy_contiguous.cuh
```diff
@@ -0,0 +1,187 @@
...
+inline constexpr int __bytes_in_flight = 64 * 1024; // 64KB
+
+//! @brief Compute the number of elements each thread copies for a given vector width.
+//!
+//! @param[in] __access_bytes Size in bytes of one vectorized access
+//! @return Number of elements per thread
+[[nodiscard]] _CCCL_HOST_API constexpr int __elem_per_thread(int __access_bytes) noexcept
+{
+  constexpr auto __threads_per_sm = 2048; // 2048 threads per SM
+  return (__bytes_in_flight / __access_bytes) / __threads_per_sm;
```
miscco: Question: Should this be `ceil_div`? We might round down in `(__bytes_in_flight / __access_bytes)`

cudax/include/cuda/experimental/__copy/copy_contiguous.cuh
```diff
@@ -0,0 +1,187 @@
...
+  constexpr int __block_size         = 256;
+  constexpr auto __elems_per_thread1 = cudax::__elem_per_thread(sizeof(_TpIn));
+  constexpr auto __tile_size         = __block_size * __elems_per_thread1;
+  const auto __inner_size            = __src.__extents[0];
+  const auto __outer_size            = cudax::__total_size(__src) / __inner_size;
+  const auto __num_inner_tiles       = ::cuda::ceil_div(__inner_size, __tile_size);
+  const auto __config =
+    ::cuda::make_config(::cuda::block_dims<__block_size>(), ::cuda::grid_dims(dim3(__num_inner_tiles, __outer_size)));
+
+  const __tensor_coord_iterator<_ExtentT, _Rank> __coord_iter(__src.__extents);
```
miscco: Important: Use list-initialization to disambiguate from function calls
```suggestion
  const __tensor_coord_iterator<_ExtentT, _Rank> __coord_iter{__src.__extents};
```

cudax/include/cuda/experimental/__copy/copy_contiguous.cuh
```diff
@@ -0,0 +1,187 @@
...
+_CCCL_HOST_API void __launch_copy_contiguous_kernel(
+  const __raw_tensor<_ExtentT, _StrideTIn, _TpIn, _Rank>& __src,
+  const __raw_tensor<_ExtentT, _StrideTOut, _TpOut, _Rank>& __dst,
+  ::cuda::stream_ref __stream,
+  const _SrcAccessor& __src_accessor = {},
+  const _DstAccessor& __dst_accessor = {})
+{
+  namespace cudax                    = ::cuda::experimental;
+  constexpr int __block_size         = 256;
+  constexpr auto __elems_per_thread1 = cudax::__elem_per_thread(sizeof(_TpIn));
```
miscco: I strongly dislike the use of shorthands.

It is really difficult to check whether we are actually fully qualifying or not. This is used twice here, so please use full qualification.

cudax/include/cuda/experimental/__copy/copy_optimized.cuh
```diff
@@ -0,0 +1,142 @@
...
+  _CCCL_GRID_CONSTANT const ::cuda::std::array<_StrideTOut, _Rank> __dst_strides,
+  _CCCL_GRID_CONSTANT const _DstAccessor __dst_accessor,
+  _CCCL_GRID_CONSTANT const __tensor_coord_iterator<_ExtentT, _Rank> __coord_iter,
+  _CCCL_GRID_CONSTANT const _ExtentT __tensor_size)
+{
+  const auto __idx    = ::cuda::gpu_thread.rank_as<_ExtentT>(::cuda::grid, __config);
+  const auto __stride = ::cuda::gpu_thread.count_as<_ExtentT>(::cuda::grid, __config);
+  const __partial_tensor __src{__src_ptr, __src_strides, __src_accessor};
+  const __partial_tensor __dst{__dst_ptr, __dst_strides, __dst_accessor};
+  for (auto __i = __idx; __i < __tensor_size; __i += __stride)
```
miscco, line 79: Question: Should this loop be unrolled? Also add please add some spacing
```suggestion

  for (auto __i = __idx; __i < __tensor_size; __i += __stride)
```

cudax/include/cuda/experimental/__copy/dispatch_by_vector.cuh
```diff
@@ -0,0 +1,143 @@
...
+//! @param[in] __op  Callable invoked with the reshaped source and destination tensors
+template <typename _ExtentT,
+          typename _StrideTIn,
+          typename _StrideTOut,
+          typename _TpIn,
+          typename _TpOut,
+          ::cuda::std::size_t _Rank,
+          typename _Op>
+_CCCL_HOST_API void __dispatch_by_vector_size(
+  const __raw_tensor<_ExtentT, _StrideTIn, _TpIn, _Rank>& __src,
```
miscco, line 81: Question: Is `__raw_tensor` a costly type or a light handle? If its rather light consider passing by value

cudax/include/cuda/experimental/__copy/dispatch_by_vector.cuh
```diff
@@ -0,0 +1,143 @@
...
+//! @return Maximum safe vector access width in bytes
+template <typename _SrcExtentT,
+          typename _SrcStrideT,
+          typename _TpSrc,
+          typename _DstExtentT,
+          typename _DstStrideT,
+          typename _TpDst,
+          ::cuda::std::size_t _MaxRank>
+[[nodiscard]] _CCCL_HOST_API ::cuda::std::size_t
+__vector_size_bytes(const __raw_tensor<_SrcExtentT, _SrcStrideT, _TpSrc, _MaxRank>& __src,
```
miscco, line 53: Question: Could this be `constexpr`?

cudax/include/cuda/experimental/__copy/dispatch_by_vector.cuh
```diff
@@ -0,0 +1,143 @@
...
+  _Op __op) noexcept
+{
+  namespace cudax              = ::cuda::experimental;
+  const auto __call_vectorized = [&](auto __const_vector_size) {
+    const auto __src_recast = cudax::__reshape_vectorized<__const_vector_size>(__src);
+    const auto __dst_recast = cudax::__reshape_vectorized<__const_vector_size>(__dst);
+    __op(__src_recast, __dst_recast);
+  };
+  const auto __vector_size_bytes = cudax::__vector_size_bytes(__src, __dst);
+#  if _CCCL_CTK_AT_LEAST(13, 0)
```
miscco, line 93: Important: This is checking the cuda toolkit version for something that looks more like a compiler check for vectorized operations.

Is this really relying on cuda toolkit headers?

cudax/include/cuda/experimental/__copy/dispatch_by_vector.cuh
```diff
@@ -0,0 +1,143 @@
...
+    {
+      __call_vectorized(__const_vector_size<2>);
+      return;
+    }
+  }
+  if constexpr (sizeof(_TpIn) <= 1)
+  {
+    __call_vectorized(__const_vector_size<1>);
+  }
+}
```
miscco, line 143: Critical: This function does nothing in this fallthrough and also does not communicate anything about that. 

If this is intentional we should at least have a comment here that makes clear that this is a valid code path

cudax/include/cuda/experimental/__copy/mdspan_d2d.cuh
```diff
@@ -0,0 +1,233 @@
...
+  }
+  if (cudax::__has_interleaved_stride_order(__dst))
+  {
+    _CCCL_THROW(::std::invalid_argument, "destination mdspan must not have interleaved stride order");
+  }
+  if (cudax::__may_overlap(__src, __dst))
+  {
+    _CCCL_THROW(::std::invalid_argument, "mdspans must not overlap in memory");
+  }
+  if (__tensor_size == 1)
```
miscco: Please add some spacing, so that it is more obvious that this is the end of the input validation
```suggestion

  if (__tensor_size == 1)
```

cudax/include/cuda/experimental/__copy/mdspan_d2d.cuh
```diff
@@ -0,0 +1,233 @@
...
+      __src_ptr += __src.mapping().offset();
+    }
+    if constexpr (::cuda::__is_layout_stride_relaxed_v<_LayoutPolicyOut>)
+    {
+      __dst_ptr += __dst.mapping().offset();
+    }
+    ::cuda::__driver::__memcpyAsync(__dst_ptr, __src_ptr, sizeof(_TpIn), __stream.get());
+    return;
+  }
+  if constexpr (_ExtentsIn::rank() > 0 && _ExtentsOut::rank() > 0)
```
miscco, line 147: Critical: It is not obvious from the control flow that the else case is handled above in the `__tensor_size == 1` case, especially given that one is `if constexpr` and the other not. 

Please add a comment for future you and me. I would also wonder if we should move the above clause into the else branch of the `if constexpr` and `_CCCL_ASSERT(__tensor_size == 1, "Invalid tensor size");` instead of the branch

cudax/include/cuda/experimental/__copy/mdspan_d2d.cuh
```diff
@@ -0,0 +1,233 @@
...
+      ::cuda::std::common_type_t<cudax::__mdspan_stride_t<_LayoutPolicyOut, decltype(__dst.mapping())>, int>;
+
+    constexpr auto __max_rank = ::cuda::std::max(_ExtentsIn::rank(), _ExtentsOut::rank());
+    const auto __src_raw      = cudax::__to_raw_tensor<__common_extent_t, __src_stride_t, __max_rank>(__src);
+    const auto __dst_raw      = cudax::__to_raw_tensor<__common_extent_t, __dst_stride_t, __max_rank>(__dst);
+    if (!cudax::__same_extents(__src_raw, __dst_raw))
+    {
+      _CCCL_THROW(::std::invalid_argument, "mdspans must have the same extents (after removing singleton dimensions)");
+    }
+    auto __src_simplified = __src_raw;
```
miscco, line 167: Nitpick: Some spacing
```suggestion

    auto __src_simplified = __src_raw;
```

cudax/include/cuda/experimental/__copy/mdspan_d2d.cuh
```diff
@@ -0,0 +1,233 @@
...
+    const auto __src_normalized = (__tile_size > 1) ? __src_simplified : cudax::__reverse_modes(__src_raw);
+    const auto __dst_normalized = (__tile_size > 1) ? __dst_simplified : cudax::__reverse_modes(__dst_raw);
+
+    _CCCL_ASSERT(__tensor_size % __tile_size == 0, "tensor size must be divisible by tile size");
+    const auto __inner_extent_bytes = __src_normalized.__extents[0] * sizeof(_TpIn);
+    // (1) contiguous case
+    if (__tile_size == __tensor_size)
+    {
+      _CCCL_TRY_CUDA_API(
+        ::cub::DeviceTransform::Transform,
```
miscco: Critical: This is unfortunately a bug, we need to use 
```suggestion
        CUB_NS_QUALIFIER::DeviceTransform::Transform,
```

Applies throughout

cudax/include/cuda/experimental/__copy/mdspan_d2d.cuh
```diff
@@ -0,0 +1,233 @@
...
+#  include <cuda/__functional/address_stability.h>
+#  include <cuda/__mdspan/host_device_mdspan.h>
+#  include <cuda/__mdspan/traits.h>
+#  include <cuda/__stream/stream_ref.h>
+#  include <cuda/std/__algorithm/max.h>
+#  include <cuda/std/__host_stdlib/stdexcept>
+#  include <cuda/std/__memory/is_sufficiently_aligned.h>
+#  include <cuda/std/__type_traits/common_type.h>
+#  include <cuda/std/__type_traits/is_convertible.h>
+#  include <cuda/std/__type_traits/is_same.h>
```
miscco, line 45: Ciritcal: This is missing some includes like 
```suggestion
#  include <cuda/std/__type_traits/is_same.h>
#  include <cuda/std/__type_traits/is_trivially_copyable.h>
#  include <cuda/std/__type_traits/conditional.h>
```

Please audit all uses of traits etc

cudax/include/cuda/experimental/__copy/tensor_copy_utils.cuh
```diff
@@ -0,0 +1,187 @@
...
+    const auto __stride = ::cuda::experimental::__abs_integer(__tensor.__strides[__i]);
+    if (__stride != 1)
+    {
+      const size_t __stride_bytes = static_cast<size_t>(__stride) * sizeof(_Tp);
+      __alignment                 = ::cuda::std::gcd(__alignment, __stride_bytes);
+    }
+  }
+  _CCCL_ASSERT(__alignment % sizeof(_Tp) == 0, "alignment is not a multiple of the element size");
+  // (3) Compute the number of items per vector over the contiguous mode
+  size_t __elem_alignment = __alignment / sizeof(_Tp);
```
miscco, line 86: Important: We are switching between raw `size_t` and `::cuda::std::size_t` please use the latter

cudax/include/cuda/experimental/__copy/tensor_copy_utils.cuh
```diff
@@ -0,0 +1,187 @@
...
+#if !_CCCL_COMPILER(NVRTC)
+
+#  include <cuda/__memory/ptr_alignment.h>
+#  include <cuda/__memory/ranges_overlap.h>
+#  include <cuda/__utility/in_range.h>
+#  include <cuda/std/__cstddef/types.h>
+#  include <cuda/std/__mdspan/mdspan.h>
+#  include <cuda/std/__memory/is_sufficiently_aligned.h>
+#  include <cuda/std/__numeric/gcd_lcm.h>
+
```
miscco, line 36: Critical: Missing some includes:
```suggestion
#  include <cuda/std/__type_traits/conditional.h>
#  include <cuda/std/__type_traits/is_const.h>
#  include <cuda/std/cstddef>

```

Please autit

cudax/include/cuda/experimental/__copy/tensor_copy_utils.cuh
```diff
@@ -0,0 +1,187 @@
...
+  using __vector_t = __reshape_vector_type<_VectorBytes, _Tp>;
+  using __rank_t   = typename __raw_tensor<_ExtentT, _StrideT, _Tp, _MaxRank>::__rank_t;
+  static_assert(_VectorBytes % sizeof(_Tp) == 0, "vector size must be a multiple of element size");
+  constexpr auto __elems_per_vector = _VectorBytes / sizeof(_Tp);
+  _CCCL_ASSERT(__tensor.__strides[0] == 1, "innermost mode must be contiguous");
+  _CCCL_ASSERT(__tensor.__extents[0] % __elems_per_vector == 0,
+               "innermost extent must be divisible by elements per vector");
+  _CCCL_ASSERT(::cuda::std::is_sufficiently_aligned<alignof(__vector_t)>(__tensor.__data),
+               "tensor data is not sufficiently aligned to the extents and strides");
+  const auto __data = reinterpret_cast<__vector_t*>(__tensor.__data);
```
miscco, line 122: Please add some spacing to increase readability
```suggestion

  constexpr auto __elems_per_vector = _VectorBytes / sizeof(_Tp);
  _CCCL_ASSERT(__tensor.__strides[0] == 1, "innermost mode must be contiguous");
  _CCCL_ASSERT(__tensor.__extents[0] % __elems_per_vector == 0,
               "innermost extent must be divisible by elements per vector");
  _CCCL_ASSERT(::cuda::std::is_sufficiently_aligned<alignof(__vector_t)>(__tensor.__data),
               "tensor data is not sufficiently aligned to the extents and strides");
               
  const auto __data = reinterpret_cast<__vector_t*>(__tensor.__data);
```

cudax/include/cuda/experimental/__copy/tensor_copy_utils.cuh
```diff
@@ -0,0 +1,187 @@
...
+  using __rank_t       = typename __raw_tensor_t::__rank_t;
+  _CCCL_ASSERT(::cuda::in_range(__tensor.__rank, size_t{1}, _MaxRank), "Invalid tensor rank");
+  if (__tensor.__strides[0] != 1)
+  {
+    return sizeof(_Tp);
+  }
+  // (1) pointer alignment
+  size_t __alignment = ::cuda::__ptr_alignment(__tensor.__data);
+  // (2) alignment over all strides
+  for (__rank_t __i = 0; __i < __tensor.__rank; ++__i)
```
miscco, line 75: Important: This might error out due to the dreaded "comparison of unsigned against 0"

Please use `__i != __tensor.__rank` to avoid that.

Applies throughout

cudax/include/cuda/experimental/__copy/tensor_copy_utils.cuh
```diff
@@ -0,0 +1,187 @@
...
+  return __result;
+}
+
+//! @brief Compute the total number of elements in a raw tensor.
+//!
+//! @param[in] __tensor Raw tensor descriptor
+//! @return Product of all extents
+template <typename _ExtentT, typename _StrideT, typename _Tp, ::cuda::std::size_t _MaxRank>
+[[nodiscard]]
+_CCCL_HOST_API _ExtentT __total_size(const __raw_tensor<_ExtentT, _StrideT, _Tp, _MaxRank>& __tensor) noexcept
```
miscco, line 142: Question: Should this be `constexpr`? That might help in case of statically known tensors

cudax/include/cuda/experimental/__copy/tensor_iterator.cuh
```diff
@@ -0,0 +1,201 @@
...
+  const ::cuda::std::array<_ExtentT, _Size>& __extents, ::cuda::std::size_t __start, ::cuda::std::size_t __end) noexcept
+{
+  using __fast_mod_div_t = ::cuda::fast_mod_div<_ExtentT>;
+  _CCCL_ASSERT(::cuda::in_range(__start, ::cuda::std::size_t{0}, _Size), "invalid __start");
+  _CCCL_ASSERT(::cuda::in_range(__end, ::cuda::std::size_t{0}, _Size), "invalid __end");
+  _ExtentT __size = 1;
+  for (auto __i = __start; __i < __end; __i++)
+  {
+    __size *= __extents[__i];
+  }
```
miscco, line 77: A kingdom for accumulate. No change requested

cudax/include/cuda/experimental/__copy/tensor_iterator.cuh
```diff
@@ -0,0 +1,201 @@
...
+  //! @return Array of coordinates into the tensor
+  [[nodiscard]] _CCCL_API ::cuda::std::array<_ExtentT, _Rank> operator()(_ExtentT __in_arraydex) const noexcept
+  {
+    if constexpr (_Rank == 1)
+    {
+      return ::cuda::std::array<_ExtentT, _Rank>{{__in_arraydex}};
+    }
+    else
+    {
+      ::cuda::std::array<_ExtentT, _Rank> __coords;
```
miscco: Important: A compiler might warn about use of uninitialized `__coords`

I believe the compiler should be smart enough to realize that we overwrite everything. Could you check whether there is any codegen issue from initializing the array?
```suggestion
      ::cuda::std::array<_ExtentT, _Rank> __coords{};
```

cudax/include/cuda/experimental/__copy/vector_access.cuh
```diff
@@ -0,0 +1,92 @@
...
+#if defined(_CCCL_IMPLICIT_SYSTEM_HEADER_GCC)
+#  pragma GCC system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_CLANG)
+#  pragma clang system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_MSVC)
+#  pragma system_header
+#endif // no system header
+
+#if !_CCCL_COMPILER(NVRTC)
+
```
miscco:
```suggestion
```

cudax/include/cuda/experimental/__copy/vector_access.cuh
```diff
@@ -0,0 +1,92 @@
...
+template <::cuda::std::size_t _VectorBytes>
+struct __vector_access
+{
+  using type = ::cuda::std::__make_nbit_uint_t<_VectorBytes * CHAR_BIT>;
+};
+
+template <>
+struct __vector_access<16>
+{
+  using type = ::uint4;
```
miscco: I believe this requires cuda tookit headers. Please add the explicit include and guard accordingly

cudax/include/cuda/experimental/__copy/vector_access.cuh
```diff
@@ -0,0 +1,92 @@
...
+[[nodiscard]] _CCCL_HOST_API inline ::cuda::std::size_t __max_gpu_arch_vector_size() noexcept
+{
+#  if _CCCL_CTK_AT_LEAST(13, 0)
+  const auto __dev_id = ::cuda::__driver::__cudevice_to_ordinal(::cuda::__driver::__ctxGetDevice());
+  const auto __dev    = ::cuda::devices[__dev_id];
+  const auto __major  = __dev.attribute<::cudaDevAttrComputeCapabilityMajor>();
+  return (__major >= 10) ? 32 : 16;
+#  else
+  return 16;
+#  endif
```
miscco: Missing comments
```suggestion
#  else // ^^^ _CCCL_CTK_AT_LEAST(13, 0) ^^^ / vvv _CCCL_CTK_BELOW(13, 0) vvv
  return 16;
#  endif // _CCCL_CTK_BELOW(13, 0)
```

cudax/include/cuda/experimental/__copy/vector_access.cuh
```diff
@@ -0,0 +1,92 @@
...
+
+template <::cuda::std::size_t _VectorBytes>
+using __vector_access_t = typename __vector_access<_VectorBytes>::type;
+
+//! @brief Query the maximum vector access width supported by the current GPU architecture.
+//!
+//! @return Maximum vector width in bytes (32 for SM >= 10.0, 16 otherwise)
+[[nodiscard]] _CCCL_HOST_API inline ::cuda::std::size_t __max_gpu_arch_vector_size() noexcept
+{
+#  if _CCCL_CTK_AT_LEAST(13, 0)
```
miscco, line 59: Critical: Do we really want to check the cuda toolkit version or the compiler version?

cudax/include/cuda/experimental/__copy/copy_contiguous.cuh
fbusato: this is purely heuristic. I need to think about but I don't see big differences

cudax/include/cuda/experimental/__copy/copy_optimized.cuh
fbusato: not explicitly. `__tensor_size` is dynamic, so it is a task for the compiler

cudax/include/cuda/experimental/__copy/dispatch_by_vector.cuh
fbusato: I would prefer by reference. It contains two `cuda::std::array` or `Rank` size, and `Rank` can be a nontrivial number.

cudax/include/cuda/experimental/__copy/dispatch_by_vector.cuh
fbusato: I think this is correct. It checks when the CTK introduces the 32-byte alignment vector types

cudax/include/cuda/experimental/__copy/dispatch_by_vector.cuh
fbusato: The dispatch is related to the number of bytes. It doesn't fallthrough because 1 is the minimum value. I will add a comment

cudax/include/cuda/experimental/__copy/vector_access.cuh
fbusato: yes, because vector types are defined in the CTK

cudax/include/cuda/experimental/__copy/dispatch_by_vector.cuh
fbusato: no, everything is dynamic here

cudax/include/cuda/experimental/__copy/copy_contiguous.cuh
fbusato: it doesn't affect correctness. I prefer the current version

cudax/include/cuda/experimental/__copy/dispatch_by_vector.cuh
fbusato: doubled-checked. This is correct, `ulong4_32a` and similar have been introduced in CTK 13.0. I will add a comment

cudax/include/cuda/experimental/__copy/dispatch_by_vector.cuh
fbusato: added a comment and a static assertion

cudax/include/cuda/experimental/__copy/mdspan_d2d.cuh
fbusato: forgot to run `libcudacxx-stype` SKILL

cudax/include/cuda/experimental/__copy/tensor_copy_utils.cuh
fbusato: it is not possible in this case because` __rank` is dynamic

cudax/include/cuda/experimental/__copy/tensor_copy_utils.cuh
fbusato: `size_t` is used multiple times so I introduced `using ::cuda::std::size_t;` in the local scope

cudax/include/cuda/experimental/__copy/tensor_copy_utils.cuh
fbusato: `__raw_tensor` is always dynamic

cudax/include/cuda/experimental/__copy/tensor_iterator.cuh
fbusato: nvcc generates identical code in both cases

cudax/include/cuda/experimental/__copy/mdspan_d2d.cuh
```diff
@@ -0,0 +1,242 @@
...
+    // check the preconditions for the vectorized case
+    constexpr bool __are_accessors_default_convertible =
+      ::cuda::std::is_convertible_v<_AccessorPolicyIn, ::cuda::std::default_accessor<_TpIn>>
+      && ::cuda::std::is_convertible_v<_AccessorPolicyOut, ::cuda::std::default_accessor<_TpOut>>;
+    constexpr bool __have_same_type =
+      ::cuda::std::is_same_v<::cuda::std::remove_cvref_t<_TpIn>, ::cuda::std::remove_cvref_t<_TpOut>>;
+    constexpr bool __are_trivial_copyable =
+      ::cuda::std::is_trivially_copyable_v<_TpIn> && ::cuda::std::is_trivially_copyable_v<_TpOut>;
+    constexpr bool __are_vectorizable =
+      sizeof(_TpIn) <= __max_vector_access && ::cuda::is_power_of_two(sizeof(_TpIn))
```
fbusato: probably this can be improved with

```suggestion
      alignof(_TpIn) <= __max_vector_access && (alignof(_TpIn) >= sizeof(_TpIn)
```

cudax/include/cuda/experimental/__copy/copy_contiguous.cuh
```diff
@@ -0,0 +1,186 @@
...
+      if (__tile_offset + __thread_id + __i < __remaining)
+      {
+        const auto __coord = __coord_iter(__base_idx + __i);
+        __dst(__coord)     = __src(__coord);
+      }
+    }
+  }
+}
+
+inline constexpr int __bytes_in_flight = 64 * 1024; // 64KB
```
NaderAlAwar: Question: we have `arch_to_min_bytes_in_flight` in tuning_transform.cuh where this exact value is repeated. Is it possible to reuse this here? Or at least have a common helper somewhere both can use?

cudax/include/cuda/experimental/__copy/mdspan_d2d.cuh
```diff
@@ -0,0 +1,242 @@
...
+      && __are_accessors_default_convertible && __have_same_type && __are_trivial_copyable;
```
NaderAlAwar: Suggestion: these trait checks are only used by the vectorized branches below, so I would move them closer to those use sites. It may also read better as a named helper/trait, e.g. __can_vectorize_copy_v, to make it clear these are preconditions for the custom vectorized path.

cudax/include/cuda/experimental/__copy/mdspan_d2d.cuh
```diff
@@ -0,0 +1,242 @@
...
+    auto __dst_ptr = __dst.data_handle();
+    if constexpr (::cuda::__is_layout_stride_relaxed_v<_LayoutPolicyIn>)
+    {
+      __src_ptr += __src.mapping().offset();
+    }
+    if constexpr (::cuda::__is_layout_stride_relaxed_v<_LayoutPolicyOut>)
+    {
+      __dst_ptr += __dst.mapping().offset();
+    }
+    ::cuda::__driver::__memcpyAsync(__dst_ptr, __src_ptr, sizeof(_TpIn), __stream.get());
```
NaderAlAwar, line 142: Question: is mixed element type copy intended to be supported by this API? The top-level constraint here is `is_convertible_v<_TpIn, _TpOut>`, which suggests yes, but this size-1 fast path uses a raw `memcpyAsync` of `sizeof(_TpIn)` bytes. If converting copies are supported, how is this path expected to handle cases like int -> float or differing source/destination sizes?

cudax/include/cuda/experimental/__copy/mdspan_d2d.cuh
```diff
@@ -0,0 +1,242 @@
...
+    if (static_cast<::cuda::std::size_t>(__tile_size) == __tensor_size)
+    {
+      _CCCL_TRY_CUDA_API(
+        CUB_NS_QUALIFIER::DeviceTransform::Transform,
+        "cub::DeviceTransform::Transform failed",
+        __src_simplified.__data,
+        __dst_simplified.__data,
+        __tensor_size,
+        ::cuda::proclaim_copyable_arguments(::cuda::std::identity{}),
+        __stream.get());
```
NaderAlAwar: Question: is copy intended to support arbitrary accessor semantics here, or only default-accessor-like cases? cub::DeviceTransform as written bypasses `accessor().access(...)`, which is fine for default_accessor/aligned_accessor-like accessors, but it would change behavior for other accessors. If only default-accessor-like accessors are meant to be supported, should that be constrained explicitly here, similar to copy_bytes:

```C++
  using __default_accessor_in  = ::cuda::std::default_accessor<_TpIn>;
  using __default_accessor_out = ::cuda::std::default_accessor<_TpOut>;
  static_assert(::cuda::std::is_convertible_v<_AccessorPolicyIn, __default_accessor_in>,
                "cudax::copy_bytes: AccessorPolicyIn must be convertible to cuda::std::default_accessor");
  static_assert(::cuda::std::is_convertible_v<_AccessorPolicyOut, __default_accessor_out>,
                "cudax::copy_bytes: AccessorPolicyOut must be convertible to cuda::std::default_accessor");
```

cudax/include/cuda/experimental/__copy/copy_contiguous.cuh
```diff
@@ -0,0 +1,186 @@
...
+      const auto __coord = __coord_iter(__base_idx + __i);
+      __dst(__coord)     = __src(__coord);
+    }
+  }
+  else
+  {
+    _CCCL_PRAGMA_UNROLL_FULL()
+    for (int __i = 0; __i < _TileSize; __i += __block_size)
+    {
+      if (__tile_offset + __thread_id + __i < __remaining)
```
NaderAlAwar: Suggestion: it feels wrong to add `__tile_offset` here since it was subtracted from `__remaining` earlier.

cudax/include/cuda/experimental/__copy/copy_contiguous.cuh
```diff
@@ -0,0 +1,186 @@
...
+
+//! @brief Compute the number of elements each thread copies for a given vector width.
+//!
+//! @param[in] __access_bytes Size in bytes of one vectorized access
+//! @return Number of elements per thread
+[[nodiscard]] _CCCL_HOST_API constexpr int __elem_per_thread(int __access_bytes) noexcept
+{
+  constexpr auto __threads_per_sm = 2048; // 2048 threads per SM
+  return (__bytes_in_flight / __access_bytes) / __threads_per_sm;
+}
```
NaderAlAwar: Important: it seems like this can return 0 when `__access_bytes` is greater than 32. We should maybe do something like

```C++
  return ::cuda::std::max((__bytes_in_flight / __access_bytes) / __threads_per_sm, 1);
```

cudax/include/cuda/experimental/__copy/tensor_copy_utils.cuh
```diff
@@ -0,0 +1,192 @@
...
+{
+  if (__a.size() == 0 || __b.size() == 0)
+  {
+    return false;
+  }
+  const auto* __a_begin = reinterpret_cast<const char*>(__a.data_handle());
+  const auto* __a_end   = __a_begin + __a.mapping().required_span_size() * sizeof(_Tp1);
+  const auto* __b_begin = reinterpret_cast<const char*>(__b.data_handle());
+  const auto* __b_end   = __b_begin + __b.mapping().required_span_size() * sizeof(_Tp2);
+  return ::cuda::ranges_overlap(__a_begin, __a_end, __b_begin, __b_end);
```
NaderAlAwar, line 186: Question: are negative strides meant to be supported here? If two mdspans overlap and one of them has a negative stride, this check seems to assume the view only touches memory at and after data_handle(). But a negative-stride view can also touch memory before data_handle(), so a real overlap could be missed.

cudax/include/cuda/experimental/__copy/copy_contiguous.cuh
fbusato: that's actually a good point. On the other hand, I'm worried that `arch_to_min_bytes_in_flight` is specifically tuned for `cub::DeviceTransform` . Also, this would mean introducing a policy chain to get the value at compile-time, which is pretty invasive. @bernhardmgruber any thought?

cudax/include/cuda/experimental/__copy/copy_contiguous.cuh
fbusato: great catch!

cudax/include/cuda/experimental/__copy/tensor_copy_utils.cuh
fbusato: `required_span_size()` cannot return a negative value. The only case with negative stride is `layout_stride_relaxed`, but `offset` must "compensate" for negative strides.
The main issue is that this conditions is not strictly evaluated in `layout_stride_relaxed`.

So this method is correct, but `layout_stride_relaxed` needs to be improved.

cudax/include/cuda/experimental/__copy/copy_contiguous.cuh
```diff
@@ -0,0 +1,230 @@
...
+//! @return Bytes-in-flight target (e.g. 12KB for V100, 16KB for A100, 48KB for H200, 64KB for B200)
+[[nodiscard]] _CCCL_HOST_API inline int __bytes_in_flight() noexcept
+{
+  const auto __dev_id = ::cuda::__driver::__cudevice_to_ordinal(::cuda::__driver::__ctxGetDevice());
+  const auto __dev    = ::cuda::devices[__dev_id];
+  const auto __major  = __dev.attribute<::cudaDevAttrComputeCapabilityMajor>();
+  const auto __minor  = __dev.attribute<::cudaDevAttrComputeCapabilityMinor>();
+  const auto __arch   = ::cuda::arch_id{__major * 10 + __minor};
+  return CUB_NS_QUALIFIER::detail::transform::arch_to_min_bytes_in_flight(__arch);
+}
```
fbusato, line 132: @pciolkosz would be nice to have a utility to avoid all these calls every time

cudax/test/copy/copy_common.cuh
```diff
@@ -0,0 +1,183 @@
...
+  thrust::device_vector<T> d_src(h_src.begin(), h_src.end());
+  thrust::device_vector<T> d_dst(dst_alloc, T{0});
+
+  using extents_t = cuda::std::dextents<int, Rank>;
+  using strides_t = cuda::dstrides<int, Rank>;
+  using mapping_t = cuda::layout_stride_relaxed::mapping<extents_t>;
+
+  extents_t ext(shape);
+  auto src_ptr = thrust::raw_pointer_cast(d_src.data()) + src_offset;
+  auto dst_ptr = thrust::raw_pointer_cast(d_dst.data()) + dst_offset;
```
NaderAlAwar: Important: I think this `layout_stride_relaxed` construction is inconsistent with the documented model. Here the pointer is shifted by `src_offset`/`dst_offset`, but the mapping itself is still created with `offset == 0`. The docs describe layout_stride_relaxed the other way around: keep the data pointer at the base, and store the compensation in `mapping.offset()` so `mapping(indices...) = offset + sum(index_i * stride_i)` remains nonnegative and `required_span_size()` reflects the actual span, especially for negative strides. See https://github.com/NVIDIA/cccl/blob/main/docs/libcudacxx/extended_api/mdspan/dlpack_to_mdspan.rst#semantics. As written, this helper seems to encode the offset in the pointer instead of the mapping.

cudax/include/cuda/experimental/__copy/mdspan_d2d.cuh
```diff
@@ -0,0 +1,251 @@
...
+    const auto __dst_raw      = cudax::__to_raw_tensor<__common_extent_t, __dst_stride_t, __max_rank>(__dst);
+    if (!cudax::__same_extents(__src_raw, __dst_raw))
+    {
+      _CCCL_THROW(::std::invalid_argument, "mdspans must have the same extents (after removing singleton dimensions)");
+    }
+
+    auto __src_simplified = __src_raw;
+    auto __dst_simplified = __dst_raw;
+    cudax::__sort_by_stride_paired(__src_simplified, __dst_simplified);
+    cudax::__flip_negative_strides_paired(__src_simplified, __dst_simplified);
```
NaderAlAwar, line 170: Question: is this also meant to handle cases where only one of src and dst has negative strides? At present, it seems that this mixed case will result in the negative stride surviving into the generic kernel path. It seems like `__partial_tensor` will compute a negative offset and pass it to `accessor.access()`, which takes `size_t`.

cudax/test/copy/copy_common.cuh
fbusato: good point. The PR was created before `layout_stride_relaxed` was merged.

cudax/include/cuda/experimental/__copy/mdspan_d2d.cuh
fbusato: this is expected. We cannot change the accessor interface which takes a `size_t`. Anyway, I will add a specific test case to check this path.

fbusato: the test is already present

https://github.com/fbusato/cccl/blob/5796e40ae1677d2125994a9db372cad55cf222a0/cudax/test/copy/copy_nvmath.cu#L131

PR #7826, opened 2026-02-27
Fix typo, looks like benchmarks are not build by default

miscco:
Also fix missing include


bernhardmgruber (approved):

bernhardmgruber:
> looks like benchmarks are not build by default

So, are they being built now?

bernhardmgruber:
> > looks like benchmarks are not build by default
> 
> So, are they being built now?

Maybe this will help: #7827

PR #7835, opened 2026-02-27
Use tuning policy as tag for tuning environment

bernhardmgruber:
- [x] Merge before: #7810
- [x] Merge before: #7565 (so we can migrate env unit tests using scan policies)

This proposes a simplification of how users can define their tunings.

### DIFF
--- a/cub/test/catch2_test_device_reduce_nondeterministic.cu
+++ b/cub/test/catch2_test_device_reduce_nondeterministic.cu
@@ -34,13 +34,15 @@ struct custom_policy_selector
 {
   _CCCL_API constexpr auto operator()(::cuda::arch_id) const -> cub::detail::reduce::reduce_policy
   {
-    auto rp = cub::detail::reduce::agent_reduce_policy{
+    const auto rp = cub::detail::reduce::agent_reduce_policy{
       BlockSize,
       ItemsPerThread,
       4,
-      cub::BlockReduceAlgorithm::BLOCK_REDUCE_WARP_REDUCTIONS_NONDETERMINISTIC,
+      cub::BlockReduceAlgorithm::BLOCK_REDUCE_WARP_REDUCTIONS,
       cub::CacheLoadModifier::LOAD_DEFAULT};
-    return {rp, rp, rp};
+    auto rp_nd            = rp;
+    rp_nd.block_algorithm = cub::BlockReduceAlgorithm::BLOCK_REDUCE_WARP_REDUCTIONS_NONDETERMINISTIC;
+    return {rp, rp, rp_nd};
   }
 };
 
@@ -95,6 +97,7 @@ C2H_TEST("Nondeterministic Device reduce works with float and double on gpu with
 
   const int num_items = GENERATE_COPY(values({0, 1, 20, 100, 2000, 1 << 20}));
 
+  // TODO(bgruber): not using cuda::std::plus falls back to run_to_run determinism. Is this intended?
   constexpr auto min_op = cuda::minimum<type>{};
   constexpr auto init   = cuda::std::numeric_limits<type>::max();
 


libcudacxx/include/cuda/__execution/tune.h
```diff
@@ -51,14 +52,16 @@ struct __get_tuning_t
...
 
-  // clang < 19 doesn't like this code
   // since all the tunings are stateless, let's ignore incoming parameters
-  ::cuda::std::execution::env<_Tunings...> __env{};
+  // we use the return type of the policy_selector as tag
+  ::cuda::std::execution::env<
+    ::cuda::std::execution::prop<decltype(_PolicySelectors{}(::cuda::arch_id{})), _PolicySelectors>...>
+    __env;
 
   return ::cuda::std::execution::prop{__get_tuning_t{}, __env};
```
miscco: I believe this might be easier to read as
```suggestion
  
  // we use the return type of the policy_selector as tag
  using tuning_env = ::cuda::std::execution::env<
    ::cuda::std::execution::prop<decltype(_PolicySelectors{}(::cuda::arch_id{})), _PolicySelectors>...>;
  
  return ::cuda::std::execution::prop{__get_tuning_t{}, tuning_env{}};
```

cub/cub/device/device_transform.cuh
```diff
@@ -105,8 +84,9 @@ private:
       ::cuda::std::is_same_v<Predicate, detail::transform::always_true_predicate>,
       ::cuda::std::tuple<RandomAccessIteratorsIn...>,
       RandomAccessIteratorOut>;
+
     using policy_selector = ::cuda::std::execution::
-      __query_result_or_t<tuning_env, detail::transform::get_tuning_query_t, default_policy_selector>;
+      __query_result_or_t<tuning_env, detail::transform::transform_policy, default_policy_selector>;
```
miscco, line 89: Oh that  is soo much better than what we had previously

bernhardmgruber: I like it a lot too!

miscco: Note that this currently relies on an internal type, so we *also* need a way for a user to retrieve the transform policy

bernhardmgruber: `transform_policy` will become a public type soon, so we should be fine!

cub/cub/device/device_reduce.cuh
```diff
@@ -183,16 +164,22 @@ private:
...
+    using accum_t  = ::cuda::std::
       __accumulator_t<ReductionOpT, ::cuda::std::invoke_result_t<TransformOpT, detail::it_value_t<InputIteratorT>>, T>;
-    using policy_t = typename reduce_tuning_t::template fn<accum_t, offset_t, ReductionOpT>;
+    using default_policy_selector = detail::rfa::policy_selector_from_types<accum_t>;
+    using policy_selector =
+      ::cuda::std::execution::__query_result_or_t<TuningEnvT, detail::rfa::rfa_policy, default_policy_selector>;
 
-    return detail::rfa::dispatch<InputIteratorT, OutputIteratorT, offset_t, T, TransformOpT, accum_t, policy_t>(
-      d_temp_storage, temp_storage_bytes, d_in, d_out, static_cast<offset_t>(num_items), init, stream, transform_op);
+    return detail::rfa::dispatch<InputIteratorT, OutputIteratorT, offset_t, T, TransformOpT, accum_t>(
```
miscco, line 173: Question: Do we actually need the explicit template arguments beyond `accum_t`?
```suggestion
    return detail::rfa::dispatch<accum_t>(
```

bernhardmgruber: I need to verify this case by case. There are some algorithms that want to override the accumulator that the dispatch function deduces.

bernhardmgruber: I will simplify this separately.

Issue #7465, opened 2026-02-03
[EPIC] Make the new CUB tuning API public


bernhardmgruber:
Depends on (but we can already start work in some places):

- [ ] #7165

We should first start with a solid design and write it down:

- [ ] #7671

For each CUB algorithm, if the tuning API has been implemented internally, we are ready to make it publicly accessible:

- [ ] Add a test (or verify there already is one) for each CUB algorithm overload specifying a custom tuning.
- [ ] Enable public access to tuning properties for CUB environment API overloads.
- [ ] Add or update documentation and examples on how the public tuning API works
- [ ] Deprecated all public CUB dispatchers and agent policies

Finally, in CCCL 4.0, we can remove all public dispatchers, which is tracked by #7466.

PR #7844, opened 2026-03-02
Implement the new tuning API for `DispatchSegmentedRadixSort`

bernhardmgruber:
- [x] No SASS changes for `cub.bench.segmented_radix_sort.keys.base` for SM`75;80;90;100`

Fixes: #7642

### DIFF
--- a/cub/cub/device/dispatch/dispatch_segmented_radix_sort.cuh
+++ b/cub/cub/device/dispatch/dispatch_segmented_radix_sort.cuh
@@ -694,7 +694,10 @@ CUB_RUNTIME_FUNCTION _CCCL_FORCEINLINE cudaError_t invoke_passes_segmented_radix
 
   if (d_temp_storage == nullptr)
   {
-    temp_storage_bytes = 1;
+    if (temp_storage_bytes == 0) // if alias_temporaries computed 0 bytes, make sure we request at least 1
+    {
+      temp_storage_bytes = 1;
+    }
     return cudaSuccess;
   }
 


cub/benchmarks/bench/segmented_radix_sort/keys.cu
```diff
@@ -17,10 +18,9 @@ void seg_radix_sort(nvbench::state& state,
   using offset_t          = OffsetT;
   using begin_offset_it_t = const offset_t*;
   using end_offset_it_t   = const offset_t*;
+  using segment_size_t    = offset_t;
   using key_t             = T;
   using value_t           = cub::NullType;
-  using dispatch_t        = cub::
-    DispatchSegmentedRadixSort<cub::SortOrder::Ascending, key_t, value_t, begin_offset_it_t, end_offset_it_t, offset_t>;
```
miscco, line 23: Nitpick: I would not say that removing the alias is an improvement

cub/cub/device/dispatch/kernels/kernel_segmented_radix_sort.cuh
```diff
@@ -86,17 +89,20 @@ namespace detail::radix_sort
...
           typename EndOffsetIteratorT,
           typename SegmentSizeT,
           typename DecomposerT = detail::identity_decomposer_t>
-__launch_bounds__(int((ALT_DIGIT_BITS) ? ChainedPolicyT::ActivePolicy::AltSegmentedPolicy::BLOCK_THREADS
-                                       : ChainedPolicyT::ActivePolicy::SegmentedPolicy::BLOCK_THREADS))
+#if _CCCL_HAS_CONCEPTS()
+  requires radix_sort_policy_selector<PolicySelector>
+#endif // _CCCL_HAS_CONCEPTS()
+__launch_bounds__(int(AltDigitBits ? PolicySelector{}(::cuda::arch_id{CUB_PTX_ARCH / 10}).alt_segmented.block_threads
+                                   : PolicySelector{}(::cuda::arch_id{CUB_PTX_ARCH / 10}).segmented.block_threads))
```
miscco: Nitpick: I believe it would be easier to read / more consistent throughout CUB if we would turn that into a function per algorithm. Depending on the number of different dispatch paths, there can be *many* different conditions here

cub/cub/device/dispatch/tuning/tuning_radix_sort.cuh
```diff
@@ -921,13 +921,108 @@ struct RadixSortPolicyWrapper<
...
+// TODO(bgruber): remove in CCCL 4.0 when we drop the radix sort dispatcher after publishing the tuning API
 template <typename PolicyT>
 _CCCL_HOST_DEVICE RadixSortPolicyWrapper<PolicyT> MakeRadixSortPolicyWrapper(PolicyT policy)
 {
   return RadixSortPolicyWrapper<PolicyT>{policy};
 }
 
+// TODO(bgruber): remove in CCCL 4.0 when we drop the radix sort dispatcher after publishing the tuning API
+template <typename LegacyActivePolicy>
+_CCCL_API constexpr auto convert_policy() -> radix_sort_policy
```
miscco, line 933: Question: When do we actually deprecate those? We should set a definite version where we mark those as deprecated so that users have a clear path forward before 4.0

cub/cub/device/dispatch/dispatch_segmented_radix_sort.cuh
```diff
@@ -633,6 +637,325 @@ struct DispatchSegmentedRadixSort
...
+  {
+    return error;
+  }
+
+  if (d_temp_storage == nullptr)
+  {
+    if (temp_storage_bytes == 0)
+    {
+      temp_storage_bytes = 1;
+    }
```
miscco: Nitpick: We should just unconditionally set this
```suggestion
    temp_storage_bytes = 1;
```

cub/cub/device/dispatch/kernels/kernel_segmented_radix_sort.cuh
```diff
@@ -145,8 +165,8 @@ __launch_bounds__(int((ALT_DIGIT_BITS) ? ChainedPolicyT::ActivePolicy::AltSegmen
     typename BlockDownsweepT::TempStorage downsweep;
     struct
     {
-      volatile SegmentSizeT reverse_counts_in[RADIX_DIGITS];
-      volatile SegmentSizeT reverse_counts_out[RADIX_DIGITS];
+      volatile SegmentSizeT reverse_counts_in[radix_digits];
+      volatile SegmentSizeT reverse_counts_out[radix_digits];
```
miscco, line 176: Question: This is the only place we use volatile for temporary storage, is this actually needed?

miscco (approved):

cub/cub/device/dispatch/tuning/tuning_radix_sort.cuh
bernhardmgruber: We can deprecate them when #7465 is done, because then users no longer need to rely on passing custom policy hubs to dispatchers. They can use the public tuning API and we can issue warnings if they use anything else.

cub/cub/device/dispatch/kernels/kernel_segmented_radix_sort.cuh
bernhardmgruber: OMG this is definitely a code smell. We should open an issue for that.

bernhardmgruber: #7846

cub/cub/device/dispatch/kernels/kernel_segmented_radix_sort.cuh
bernhardmgruber: Done.

cub/benchmarks/bench/segmented_radix_sort/keys.cu
bernhardmgruber: But we are no longer using the `DispatchSegmentedRadixSort` struct, but the new `dispatch` function. I can create a function pointer here, but then we loose deduction of all template arguments but the first two :S

cub/cub/device/dispatch/dispatch_segmented_radix_sort.cuh
bernhardmgruber: Turns out this is a bug :) Reverting

PR #7853, opened 2026-03-02
[libcu++] Change default pool getters to return memory_pool_ref&

pciolkosz:
Creating a `resource_ref` for a default memory pool creates a reference to a temporary returned from the default pool getter.
This PR changes the return type of those getters to `X_memory_pool_ref&` that references a static variable where we keep the pool.

### DIFF
--- a/libcudacxx/include/cuda/__memory_pool/device_memory_pool.h
+++ b/libcudacxx/include/cuda/__memory_pool/device_memory_pool.h
@@ -81,11 +81,11 @@ public:
 //! @brief  Returns the default ``cudaMemPool_t`` from the specified device.
 //! @throws cuda_error if retrieving the default ``cudaMemPool_t`` fails.
 //! @returns The default memory pool of the specified device.
-[[nodiscard]] inline device_memory_pool_ref device_default_memory_pool(::cuda::device_ref __device)
+[[nodiscard]] inline device_memory_pool_ref& device_default_memory_pool(::cuda::device_ref __device)
 {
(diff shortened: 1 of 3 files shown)


libcudacxx/include/cuda/__memory_pool/device_memory_pool.h
```diff
@@ -81,11 +81,11 @@ class device_memory_pool_ref : public __memory_pool_base
...
 //! @returns The default memory pool of the specified device.
-[[nodiscard]] inline device_memory_pool_ref device_default_memory_pool(::cuda::device_ref __device)
+[[nodiscard]] inline device_memory_pool_ref& device_default_memory_pool(::cuda::device_ref __device)
 {
-  static ::cudaMemPool_t __pool = ::cuda::__get_default_memory_pool(
-    ::CUmemLocation{::CU_MEM_LOCATION_TYPE_DEVICE, __device.get()}, ::CU_MEM_ALLOCATION_TYPE_PINNED);
-  return device_memory_pool_ref(__pool);
+  static device_memory_pool_ref __pool{::cuda::__get_default_memory_pool(
+    ::CUmemLocation{::CU_MEM_LOCATION_TYPE_DEVICE, __device.get()}, ::CU_MEM_ALLOCATION_TYPE_PINNED)};
+  return __pool;
```
davebayer, line 88: Does this mean that I can now overwrite the device default memory pool value?

miscco: Probably, but I would not care to much about that detail

PR #7857, opened 2026-03-03
Avoid GCC bug with dependent type template

miscco:
In our new policy definitions we often define a DefaultPolicy that is dependent on some other template parameters.

Older GCC, specifically gcc-11.2 cannot parse a type inheriting such a dependent type.

Luckily we can easily work around this by introducing an alias.

Fixes nvbug5935129

### DIFF
--- a/cub/cub/device/dispatch/tuning/tuning_reduce_by_key.cuh
+++ b/cub/cub/device/dispatch/tuning/tuning_reduce_by_key.cuh
@@ -882,8 +882,11 @@ struct policy_hub
                              default_reduce_by_key_delay_constructor_t<AccumT, int>>;
   };
 
+  // nvbug5935129: GCC-11.2 cannot directly use DefaultPolicy inside Policy500
+  using DefaultPolicy500 = DefaultPolicy<LOAD_LDG>;
+
   struct Policy500
-      : DefaultPolicy<LOAD_LDG>
+      : DefaultPolicy500
       , ChainedPolicy<500, Policy500, Policy500>
   {};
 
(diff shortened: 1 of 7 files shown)


cub/cub/device/dispatch/tuning/tuning_reduce_by_key.cuh
```diff
@@ -906,8 +909,11 @@ struct policy_hub
       decltype(select_agent_policy<sm80_tuning<KeyT, AccumT, is_primitive_op<ReductionOpT>()>>(0));
   };
 
+  // nvbug5935129: GCC-11.2 cannot directly use DefaultPolicy inside Policy860
+  using DefaultPolicy860 = DefaultPolicy<LOAD_LDG>;
+
   struct Policy860
-      : DefaultPolicy<LOAD_LDG>
+      : DefaultPolicy860
```
bernhardmgruber, line 916: Consider reusing the default policy from above:
```suggestion
struct Policy860
      : DefaultPolicy500
```
Applies to several more places in this PR.

bernhardmgruber (approved):

cub/cub/device/dispatch/tuning/tuning_reduce_by_key.cuh
miscco: I was really not sure whether we want that or not. 

Depending on tuning this could change, and this makes it very clear that it is a unique policy.

No hard feelings though if you prefer it, its your code

bernhardmgruber: Just leave it. Those are the legacy tunings anyway and we are going to remove them in CCCL 4.

PR #7859, opened 2026-03-03
Avoid compile issue with `__iset`

miscco:
Looks like GCCC8 & GCC9 both have issues which have shown themself in CI with the new pstl backend

Avoid it by expanding the workaround to also trigger with those compilers

Fixes nightly CI issues


pciolkosz:
I believe we need to back port this to 3.3, see failures in: https://github.com/NVIDIA/cccl/pull/7872, but I would appreciate another pair of eyes to confirm its the same issue

PR #7866, opened 2026-03-03
Detect non GNU getopt and error early

Jacobfaib:
## Description

Detect non "enhanced" `getopt` early in `.devcontainer/launch.sh` (for example, BSD `getopt`) and give a more informative error message. BSD `getopt` for instance does not support long options and will silently skip them leading to strange errors later down the line

## Checklist

### DIFF
--- a/.devcontainer/launch.sh
+++ b/.devcontainer/launch.sh
@@ -41,6 +41,22 @@ parse_options() {
     local -;
     set -euo pipefail;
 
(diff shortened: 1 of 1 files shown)


Jacobfaib:
Maybe a better approach would be to use the standard `getopt` builtin, but this requires manually parsing long options and doesn't support fancy options like `--option=value` (note the `=`)

PR #7868, opened 2026-03-03
Fix `cuda::dynamic_shared_memory` alignment

### LINKED ISSUE #7867: [BUG]: `cuda::dynamic_shared_memory(config)` may returned misaligned pointer
### Is this a duplicate?

### Type of Bug

Silent Failure

### Component

libcu++

### Describe the bug

When using `cuda::launch` with dynamic shared memory, we allocate the proper size, but we don't take into account the alignment.

In `cuda::dynamic_shared_memory(config)`, we do:
```cpp
{
  extern __shared__ unsigned char dyn_smem[];
  return *reinterpret_cast<T*>(dyn_smem);
}
```
but there is no guarantee that `dyn_smem` will have the right alignment.

### How to Reproduce

no-repro

### Expected behavior

`cuda::dynamic_shared_memory(config)` should return a view to properly aligned memory address.

### Reproduction link

_No response_

### Operating System

_No response_

### nvidia-smi output

_No response_

### NVCC version

_No response_

davebayer:
Fixes #7867.

### DIFF
--- a/libcudacxx/include/cuda/__launch/configuration.h
+++ b/libcudacxx/include/cuda/__launch/configuration.h
@@ -794,6 +794,9 @@ template <typename Dimensions, typename... Options>
 
 #  if _CCCL_CUDA_COMPILATION()
 
+template <class _Tp>
+extern __shared__ _Tp __cccl_device_dyn_smem[];
+
 template <class _Dims, class... _Opts>
 _CCCL_DEVICE_API decltype(auto) dynamic_shared_memory(const kernel_config<_Dims, _Opts...>& __config) noexcept
 {
@@ -801,8 +804,7 @@ _CCCL_DEVICE_API decltype(auto) dynamic_shared_memory(const kernel_config<_Dims,
   using _Opt  = ::cuda::std::remove_reference_t<decltype(__opt)>;
   static_assert(!::cuda::std::is_same_v<_Opt, __detail::option_not_found>,
                 "Dynamic shared memory option not found in the kernel configuration");
-  extern __shared__ unsigned char __cccl_device_dyn_smem[];
-  return __opt.__make_view(reinterpret_cast<typename _Opt::value_type*>(__cccl_device_dyn_smem));
+  return __opt.__make_view(__cccl_device_dyn_smem<typename _Opt::value_type>);
 }
 
 #  endif // _CCCL_CUDA_COMPILATION()


libcudacxx/include/cuda/__launch/configuration.h
```diff
@@ -794,15 +794,17 @@ template <typename Dimensions, typename... Options>
 
 #  if _CCCL_CUDA_COMPILATION()
 
+template <class _Tp>
+extern __shared__ _Tp __cccl_device_dyn_smem[];
```
fbusato, line 798: @bernhardmgruber had hard time to enforce shared memory alignment (he even wrote a guide on that). I'm not 100% sure that template variables work here. I would defer to Bernhard.

davebayer: It should be working fine, see https://godbolt.org/z/z6WWEzd5s

fbusato: This explain the problem in a practical way https://godbolt.org/z/e5P7dsGdq

davebayer: But what's the problem here? The output is right. See https://godbolt.org/z/EaY1bvP36

bernhardmgruber: The code above is fine if you never pass a type with an alignment higher than 16. If you do, the compiler correctly generates an `.align N` specifier into PTX, which is then lost in the backend writing the binary if you compile with `-G` or `-rdc=true` before nvcc 13.1.

See: https://godbolt.org/z/oW6Kcanc6

bernhardmgruber, line 798: Remark: it's not needed to pull the declaration outside the function `dynamic_shared_memory`.

bernhardmgruber: Here is the full (internal) story on dynamic SMEM: https://github.com/NVIDIA/cccl_private/wiki/Dynamic-shared-memory-alignment

bernhardmgruber: Remark: it is absolutely critical that CCCL never passes a type with an alignment larger than 16 bytes itself to this variable template.

davebayer: We could `static_assert(alignof(typename _Opt::value_type) <= 16)`. Is this something we want to enforce?

bernhardmgruber: Where would you add the static assert? It's fine if the user passes a type with higher alignment. The problem I am referring to is that any type with alignment > 16 will cause an increase of the static shared memory padding for the entire TU, which can impact occupancy. CCCL must not cause such a change, but if the user causes it, it's not our fault ;)

PR #7869, opened 2026-03-03
Enables sort_by_key with long double keys on the host

### LINKED ISSUE #7865: [BUG]: sort_by_key with a key type of long double does not compile
### Is this a duplicate?

### Type of Bug

Compile-time Error

### Component

Thrust

### Describe the bug

`sort_by_key` with a key type of `long double` does not compile

### How to Reproduce

See godbolt link below.   Compile and run the minimal reproducer below.  I used `host_vector` here, but also originally encountered the failure using `device_vector`

```
#define THRUST_DEVICE_SYSTEM 4
#include <thrust/host_vector.h>
#include <thrust/sort.h>

int main(int argc, char** argv)
{
    thrust::host_vector<int> a(10, 0);
    thrust::host_vector<long double> b(10, 0);

    // works
    thrust::sort_by_key(a.begin(), a.end(), b.begin());

    // doesn't work
    thrust::sort_by_key(b.begin(), b.end(), a.begin());

    return 0;
}
```

### Expected behavior

The code should compile, but instead errors with.  This worked with earlier versions of Thrust.

```
/opt/compiler-explorer/libs/cccl/trunk/thrust/thrust/system/detail/sequential/stable_radix_sort.h:536:3: error: type 'radix_sort_dispatcher<sizeof(KeyType)>' does not provide a call operator
  536 |   radix_sort_dispatcher<sizeof(KeyType)>()(exec, keys1, keys2, vals1, vals2, N);
      |   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

### Reproduction link

https://godbolt.org/z/qb48r3We4

### Operating System

_No response_

### nvidia-smi output

_No response_

### NVCC version

_No response_

andrewcorrigan:
## Enables `sort_by_key` with `long double` keys on the host

In terms of the approach: I disabled radix sort for this key type.  At first I tried to enable it by adding a specialization for `radix_sort_dispatcher` for the case of 16 byte keys.  But then, I also saw that there would be need  to be a specialization for `RadixEncoder` which would require a non-standard `uint128_t`.   Since this is all just for host code, I figured the best route would be to fall back on the less efficient code path.  In addition, I did not add the corresponding logic for the CUDA backend since it wouldn't be usable anyway.

Fixes #7865

- Sets `use_primitive_sort` to `false` for `long double`
- Implements `TestSortByKeyLongDouble` to enforce that it compiles and runs correctly on the host.
- Also fixes a typo in `_CCCL_ASSERT_IMPL_HOST` caused compilation errors when testing this on a Mac.

## Checklist

### DIFF
--- a/thrust/testing/sort_by_key.cu
+++ b/thrust/testing/sort_by_key.cu
@@ -140,8 +140,8 @@ DECLARE_UNITTEST(TestSortByKeyBoolDescending);
 
 void TestSortByKeyLongDouble()
 {
-  thrust::host_vector<long double> h_keys  = {10.0L, 9.0L, 8.0L, 7.0L, 6.0L, 5.0L, 4.0L, 3.0L, 2.0L, 1.0L};
-  thrust::host_vector<int> h_values = {1, 2, 3, 4, 5, 6, 7, 8, 9, 10};
+  thrust::host_vector<long double> h_keys    = {10.0L, 9.0L, 8.0L, 7.0L, 6.0L, 5.0L, 4.0L, 3.0L, 2.0L, 1.0L};
+  thrust::host_vector<int> h_values          = {1, 2, 3, 4, 5, 6, 7, 8, 9, 10};
   thrust::host_vector<int> h_values_expected = {10, 9, 8, 7, 6, 5, 4, 3, 2, 1};
 
   thrust::sort_by_key(h_keys.begin(), h_keys.end(), h_values.begin());


thrust/thrust/system/detail/sequential/sort.h
```diff
@@ -34,6 +34,7 @@ namespace sort_detail
 template <typename KeyType, typename Compare>
 inline constexpr bool use_primitive_sort =
   ::cuda::std::is_arithmetic_v<KeyType>
+  && !::cuda::std::is_same_v<KeyType, long double>
```
bernhardmgruber, line 38: Suggestion: Let's encode what the backend supports:
```suggestion
  // radix_sort_dispatcher only supports key sizes of 1, 2, 4, or 8 bytes
  && (sizeof(KeyType) == 1 || sizeof(KeyType) == 2 || sizeof(KeyType) == 4 || sizeof(KeyType) == 8)
```

andrewcorrigan: This change won't work in isolation and opens up a can of worms.

I had already tried something like this originally, but it leaves `use_primitive_sort` enabled on macOS/ARM (and probably Windows too) since `sizeof(long double) == 8`.   That leads to the more challenging issue, which is that we would need an overload of `RadixEncoder` for `long double`.  However, implementing that correctly seems not worth it, since it would need to handle both the 8 byte and 16 byte case and `std::uint128_t` is not a standard thing.  
```
template <>
struct RadixEncoder<long double>
{
  _CCCL_HOST_DEVICE auto operator()(long double x) const
  {
    if constexpr(sizeof(long double) == sizeof(std::uint64_t)) // same as double
    {
        union
        {
            long double f;
            std::uint64_t i;
        } u;
        u.f                = x;
        std::uint64_t mask = -static_cast<std::int64_t>(u.i >> 63) | (static_cast<std::uint64_t>(1) << 63);
        return u.i ^ mask;
    }
#if 0  // XXX can we assume __uint128_t and __int128_t are available?
    else if constexpr(sizeof(long double) == sizeof(__uint128_t))
    {
        union
        {
            long double f;
            __uint128_t i;
        } u;
        u.f                = x;
        __uint128_t mask = -static_cast<__int128_t>(u.i >> 127) | (static_cast<__uint128_t>(1) << 127);
        return u.i ^ mask;
    }
#endif
    else
    {
        // FAIL
    }
  }
};
```

I don't think it's a good idea to use non-standard types here, so I would only plan on implementing the first branch and error out otherwise.

I'll go with whatever you prefer.  Thank you.

bernhardmgruber: Can you add both conditions? I am not arguing for supporting `long double` in radix sort. I just want to future proof if we get more arithmetic types that are not supported by `radix_sort_dispatcher`

andrewcorrigan: Done.  It now adds,
```
  // RadixEncoder is not specialized for long double
  && !::cuda::std::is_same_v<KeyType, long double>
  // radix_sort_dispatcher only supports key sizes of 1, 2, 4, or 8 bytes
  && (sizeof(KeyType) == 1 || sizeof(KeyType) == 2 || sizeof(KeyType) == 4 || sizeof(KeyType) == 8)
```

bernhardmgruber:
pre-commit.ci autofix

libcudacxx/include/cuda/std/__cccl/assert.h
```diff
@@ -89,7 +89,7 @@ void __assert_fail(const char* __assertion, const char* __file, unsigned int __l
 #  if _CCCL_OS(APPLE)
 #    define _CCCL_ASSERT_IMPL_HOST(expression, message)      \
       _CCCL_BUILTIN_EXPECT(static_cast<bool>(expression), 1) \
-      ? (void) 0 : __assert_rtn(__func__, __FILE__, __LINE__, __message__)
+      ? (void) 0 : __assert_rtn(__func__, __FILE__, __LINE__, message)
```
miscco, line 92: 🙀

thrust/thrust/system/detail/sequential/sort.h
miscco: long double is really a can of worms, because it is not immediately clear how many bits it actually has, so I would just drop it

miscco (approved):

libcudacxx/include/cuda/std/__cccl/assert.h
bernhardmgruber: Yeah, seems we don't have a lot of Mac users :D

bernhardmgruber:
Thank you for the fix!

PR #7870, opened 2026-03-03
Don't assume CUDA is available for thrust single config

Jacobfaib:
## Description

Thrust single-config assumes CUDA is available as a device backend without first checking that it exists on the system. Without these changes, the user must manually specify `-DTHRUST_DEVICE_SYSTEM=CPP`, which we should be able to figure out on our own.

## Checklist

### DIFF
--- a/thrust/cmake/ThrustMultiConfig.cmake
+++ b/thrust/cmake/ThrustMultiConfig.cmake
@@ -128,9 +128,25 @@ function(thrust_configure_multiconfig)
     if (DEFINED THRUST_DEVICE_SYSTEM)
       set_property(CACHE THRUST_DEVICE_SYSTEM PROPERTY TYPE STRING)
     else()
+      # This is a chicken-egg problem. We have not run thrust_update_system_found_flags()
+      # yet (to set THRUST_CUDA_FOUND, THRUST_OMP_FOUND etc) because that requires us to
+      # find and configure thrust first via thrust_find_thrust(). But we cannot do that
+      # without first setting defaults here.
+      #
+      # So we make a best effort attempt at deducing suitable default values here.
+      find_package(CUDAToolkit QUIET)
+      find_package(OpenMP QUIET)
+      if (CUDAToolkit_FOUND)
+        set(thrust_device_system CUDA)
+      elseif (OpenMP_FOUND)
+        set(thrust_device_system OMP)
+      else()
+        set(thrust_device_system CPP)
+      endif()
+
       set(
         THRUST_DEVICE_SYSTEM
-        "CUDA"
+        "${thrust_device_system}"
         CACHE STRING
         "The targeted device system: ${THRUST_DEVICE_SYSTEM_OPTIONS}"
       )


bernhardmgruber (commented):
Looks reasonable to me, but I would like to have @alliepiper's opinion here.

bernhardmgruber (approved):

bernhardmgruber:
> Looks reasonable to me, but I would like to have @alliepiper's opinion here.

Ping

Jacobfaib:
@alliepiper ping

Jacobfaib:
> We can make the scripts / presets friendlier if non-CUDA Thrust builds are needed regularly.

This might be the better approach then. Perhaps we should define some CPU-only presets for developers that encode these variables. Based on my personal configure command, the following allowed me to configure successfully:
```bash
cmake_args=()
cmake_args+=(-DCMAKE_BUILD_TYPE=Debug)
cmake_args+=(-DLIBCUDACXX_ENABLE_CUDA=OFF)
cmake_args+=(-DLIBCUDACXX_ENABLE_LIBCUDACXX_TESTS=OFF)
cmake_args+=(-DCCCL_ENABLE_CUB=OFF)
cmake_args+=(-DTHRUST_MULTICONFIG_ENABLE_SYSTEM_TBB=OFF)
cmake_args+=(-DTHRUST_MULTICONFIG_ENABLE_SYSTEM_CUDA=OFF)
cmake_args+=(-DTHRUST_MULTICONFIG_ENABLE_SYSTEM_OMP=OFF)
cmake_args+=(-DTHRUST_ENABLE_HEADER_TESTING=OFF)
cmake_args+=(-DCCCL_ENABLE_CUB=OFF)
cmake_args+=(-Dcudax_ENABLE_CUDA=OFF)
cmake_args+=(-DC2H_ENABLE_CUDA=OFF)
cmake_args+=(-DCCCL_ENABLE_C_PARALLEL=OFF)
cmake_args+=(-DCCCL_ENABLE_C_EXPERIMENTAL_STF=OFF)
cmake_args+=(-DCCCL_ENABLE_EXAMPLES=OFF)
cmake_args+=(-DCCCL_ENABLE_NVBENCH_HELPER=OFF)
cmake_args+=(-DCCCL_ENABLE_CUDAX=OFF)

cmake --preset all-dev-debug "${cmake_args[@]}"
```

alliepiper:
> Based on my personal configure command, the following allowed me to configure successfully:

We encourage devs to use the scripts here instead of raw configures: https://github.com/NVIDIA/cccl/blob/main/CONTRIBUTING.md

```
$ ci/build_thrust.sh -configure -cmake-options "-DTHRUST_MULTICONFIG_ENABLE_SYSTEM_CUDA=OFF"
```

will effectively do the same (it leaves TBB/OMP on, but those are extremely cheap to build).

I wouldn't be opposed to adding a new flag to the thrust script if disabling backends is something we want to do regularly. In practice, we're usually focused on CUDA, and the C++ backends are relatively free next to CUDA, so we just leave them all on by default, since all are available in our devcontainers.

You can also just configure all of them with the defaults and then just build/test the subprojects. This is my usual workflow:

```bash
ci/build_thrust.sh -configure # release
ci/build_thrust.sh -configure -cmake-options "-DCMAKE_BUILD_TYPE=Debug" # debug
cd build/preset-latest

# Build/test a single host+device configuration
# thrust.[host_system].[device_system]
# cpp.cpp, omp.cuda, cpp.tbb, etc
ninja thrust.cpp.cuda 
ctest -R thrust.cpp.cuda
```

PR #7874, opened 2026-03-03
Implement the new tuning API for `DispatchSegmentedSort`

bernhardmgruber:
- [x] No SASS changes for `cub.bench.segmented_sort.keys.base` on SM `75;80;86`. ~~The large and small kernels are no longer emitted since their code paths are not taken. If you look into the git history, I forced emitting them to verify their SASS did not change, but the PR ultimately removes the kernels from the SASS entirely.~~ (reverted)

Note: the implementation of `DispatchSegmentedSort` also uses `DeviceThreeWayPartition`, which will be rewritten separately in #7900

Fixes: #7643
Fixes: #7477

### DIFF
--- a/c/parallel/src/radix_sort.cu
+++ b/c/parallel/src/radix_sort.cu
@@ -229,8 +229,8 @@ try
   const auto policy_sel = cub::detail::radix_sort::policy_selector{
     static_cast<int>(input_keys_it.value_type.size),
     // FIXME(bgruber): input_values_it.value_type.size is 4 when it represents cub::NullType, which is very odd
-    keys_only ? 0 : static_cast<int>(input_values_it.value_type.size), // TODO(bgruber): instead of 0 we should probably
-                                                                       // use int{sizeof(cub::NullType)}
+    // TODO(bgruber): instead of 0 we should probably use int{sizeof(cub::NullType)}
+    keys_only ? 0 : static_cast<int>(input_values_it.value_type.size),
     int{sizeof(OffsetT)},
     key_type};
 


c/parallel/src/segmented_sort.cu
```diff
@@ -888,35 +798,34 @@ CUresult cccl_device_segmented_sort_impl(
     cub::DoubleBuffer<indirect_arg_t> d_values_double_buffer(
       *static_cast<indirect_arg_t**>(&val_arg_in), *static_cast<indirect_arg_t**>(&val_arg_out));
 
-    auto exec_status = cub::DispatchSegmentedSort<
+    // TODO(bgruber): remove all template arguments except the first two (the others can be deduced)
```
bernhardmgruber, line 801: Addressed in #7900

c/parallel/src/segmented_sort.cu
```diff
@@ -769,35 +691,23 @@ __device__ consteval auto& three_way_partition_policy_generator() {{
...
-  using cub::detail::RuntimeRadixSortDownsweepAgentPolicy;
-  auto large_segment_policy = RuntimeRadixSortDownsweepAgentPolicy::from_json(runtime_policy, "LargeSegmentPolicy");
-
-  using cub::detail::RuntimeSubWarpMergeSortAgentPolicy;
-  auto small_segment_policy = RuntimeSubWarpMergeSortAgentPolicy::from_json(runtime_policy, "SmallSegmentPolicy");
-
-  auto medium_segment_policy = RuntimeSubWarpMergeSortAgentPolicy::from_json(runtime_policy, "MediumSegmentPolicy");
-
-  int partitioning_threshold = runtime_policy["PartitioningThreshold"].get<int>();
+  // TODO(bgruber): convert to the new tuning API
```
bernhardmgruber, line 694: Addressed in #7900

cub/cub/device/dispatch/dispatch_segmented_sort.cuh
```diff
@@ -896,4 +974,594 @@ private:
...
+  temporary_storage::alias<uint8_t>& device_partition_temp_storage,
+  temporary_storage::alias<local_segment_index_t>& large_and_medium_segments_indices,
+  temporary_storage::alias<local_segment_index_t>& small_segments_indices,
+  temporary_storage::alias<local_segment_index_t>& group_sizes,
+  KernelSource& kernel_source,
+  PartitionKernelSource& partition_kernel_source,
+  KernelLauncherFactory& launcher_factory,
+  typename PartitionPolicyHub::MaxPolicy partition_max_policy,
+  const segmented_sort_policy& active_policy,
+  GetFinalOutputOp&& get_final_output)
```
bernhardmgruber, line 1014: This was of course generated by Cursor, because no sane person would type this out. I also tried using a lambda inside the `dispatch` function below, but then it gets really long and is also unreadable. I will gladly take any suggestions on how to make this more readable. I even considered moving back to a struct with member functions, so we can share all the template parameters between the functions.

miscco: OIP

miscco: Why do we pass the individual aliases of the temporary storage individually?

miscco: Also why is there
```
DoubleBuffer<KeyT>& d_keys,
DoubleBuffer<ValueT>& d_values,
device_double_buffer<KeyT>& d_keys_double_buffer,
device_double_buffer<ValueT>& d_values_double_buffer,
```

bernhardmgruber: > Why do we pass the individual aliases of the temporary storage individually?

Because they have references internally to the main `temporary_storage_variable` outside the function.

> Also why is there ...

The first two are the user inputs, the last two are either pointing into those or into some auxiliary buffers depending on some settings.

bernhardmgruber:
I tried to simplify the code for about half an hour now, but couldn't beat the current line count or reduce the complexity :S

c/parallel/src/radix_sort.cu
```diff
@@ -229,7 +229,8 @@ try
   const auto policy_sel = cub::detail::radix_sort::policy_selector{
     static_cast<int>(input_keys_it.value_type.size),
     // FIXME(bgruber): input_values_it.value_type.size is 4 when it represents cub::NullType, which is very odd
-    keys_only ? 0 : static_cast<int>(input_values_it.value_type.size),
+    keys_only ? 0 : static_cast<int>(input_values_it.value_type.size), // TODO(bgruber): instead of 0 we should probably
+                                                                       // use int{sizeof(cub::NullType)}
```
miscco: Nitpick:
```suggestion
// FIXME(bgruber): input_values_it.value_type.size is 4 when it represents cub::NullType, which is very odd
// TODO(bgruber): instead of 0 we should probably use int{sizeof(cub::NullType)}
    keys_only ? 0 : static_cast<int>(input_values_it.value_type.size), 
```

cub/cub/device/dispatch/tuning/tuning_segmented_sort.cuh
```diff
@@ -15,12 +15,279 @@
...
 #include <cub/util_device.cuh>
 
+#include <cuda/__device/arch_id.h>
+#include <cuda/std/__algorithm/max.h>
+#include <cuda/std/__type_traits/is_same.h>
+#include <cuda/std/concepts>
+
+#if !_CCCL_COMPILER(NVRTC)
+#  include <ostream>
+#endif
```
miscco: Nitpick:
```suggestion
#endif // !_CCCL_COMPILER(NVRTC)
```

cub/cub/device/dispatch/dispatch_segmented_sort.cuh
```diff
@@ -35,25 +36,24 @@
 
 #include <nv/target>
 
+#if !_CCCL_COMPILER(NVRTC) && defined(CUB_DEBUG_LOG)
+#  include <sstream>
+#endif
```
miscco: Nitpick: 
```suggestion
#endif // !_CCCL_COMPILER(NVRTC) && defined(CUB_DEBUG_LOG)
```

cub/cub/device/dispatch/tuning/tuning_segmented_sort.cuh
```diff
@@ -15,12 +15,279 @@
...
+
+#if !_CCCL_COMPILER(NVRTC)
+#  include <ostream>
+#endif // !_CCCL_COMPILER(NVRTC)
+
 CUB_NAMESPACE_BEGIN
 
 namespace detail::segmented_sort
 {
+struct segmented_radix_sort_policy
```
NaderAlAwar, line 34: Nit: for reduce, we named the equivalent policy `agent_reduce_policy`. I think it would be more consistent to name this `agent_segmented_radix_sort_policy`.

cub/cub/device/dispatch/tuning/tuning_segmented_sort.cuh
```diff
@@ -15,12 +15,279 @@
...
+    return os
+        << "segmented_radix_sort_policy { .block_threads = " << p.block_threads
+        << ", .items_per_thread = " << p.items_per_thread << ", .load_algorithm = " << p.load_algorithm
+        << ", .load_modifier = " << p.load_modifier << ", .rank_algorithm = " << p.rank_algorithm
+        << ", .scan_algorithm = " << p.scan_algorithm << ", .radix_bits = " << p.radix_bits << " }";
+  }
+#endif // !_CCCL_COMPILER(NVRTC)
+};
+
+struct sub_warp_merge_sort_policy
```
NaderAlAwar, line 71: Nit: same as above, name this `agent_sub_warp_merge_sort_policy`

cub/cub/device/dispatch/tuning/tuning_segmented_sort.cuh
bernhardmgruber: It's actually the other way around. The name `agent_` should completely disappear. This will also happen before the tuning API goes public.

Issue #7875, opened 2026-03-04
[BUG]: Warnings with `-Wnull-dereference` cmake flag


nirandaperera:
### Is this a duplicate?

- [x] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cccl/issues) for this bug and that I agree to the [Code of Conduct](CODE_OF_CONDUCT.md)

### Type of Bug

Silent Failure

### Component

libcu++

### Describe the bug

Issue is described here https://gist.github.com/bdice/36ab8a1f3e6f851efb2e58bb3602d853

### How to Reproduce

build with 
```
        "-DCMAKE_CXX_FLAGS=-Wnull-dereference -Werror",
        "-DCMAKE_CUDA_FLAGS=-Xcompiler=-Wnull-dereference -Xcompiler=-Werror"
```

### Expected behavior

should not fail

### Operating System

ubuntu 24.04

miscco:
I believe this was a false positive and we disabled the warning in rapids

PR #7876, opened 2026-03-04
bypass null-deref warnings

nirandaperera:
## Description

Fixes https://github.com/NVIDIA/cccl/issues/7875 

Enables `-Wnull-dereference` (with `-Werror`) in the libcudacxx test suite and suppresses the resulting GCC false-positive warnings in the `__basic_any` type-erasure implementation.

### Why `-Wnull-dereference`?

`-Wnull-dereference` asks GCC to perform inter-procedural null-pointer flow analysis and warn (with `-Werror`, hard-fail) on any code path that can reach a dereference through a pointer that could be null. This catches a class of bugs that `-Wall` and `-Wextra` miss.

### Why the suppressions?

GCC's analysis cannot see through `__basic_any`'s type-erasure invariants. When `__query_interface` is inlined, GCC observes its `return nullptr` branch and concludes the result *might* be null at every call site, even those that are guarded by the design (e.g., `__iunknown` is always registered, virtual dispatch is only reached through a live object). This leads to false-positive errors in three locations:

| File | Location | Root cause |
|---|---|---|
| `virtcall.h` | `__virtcall` — `__vptr->__fn_(...)` | `__query_interface` inlined; GCC sees potential `nullptr` |
| `basic_any_value.h` | `reset()` / `type()` — `__query_interface(__iunknown())->...` | Same: `__iunknown` lookup inlined, nullable path visible |
| `rtti.h` | `__try_vptr_cast` — `rtti->__query_interface(...)` | `__query_interface(__iunknown())` inlined into slow-path cast |

Each suppression is scoped as tightly as possible — a `_CCCL_DIAG_PUSH` / `_CCCL_DIAG_SUPPRESS_GCC("-Wnull-dereference")` / `_CCCL_DIAG_POP` bracket around only the offending statement(s). `_CCCL_DIAG_SUPPRESS_GCC` expands to a no-op on all non-GCC compilers.

## Changes

- **`libcudacxx/test/CMakeLists.txt`** — adds `-Wnull-dereference` to `headertest_warning_levels_device` (NVIDIA/NVCC, Clang-as-device, and generic paths) and `headertest_warning_levels_host` (non-MSVC). MSVC paths are unchanged.
- **`libcudacxx/include/cuda/__utility/__basic_any/virtcall.h`** — suppresses false positive in `__virtcall` at the virtual dispatch call site.
- **`libcudacxx/include/cuda/__utility/__basic_any/basic_any_value.h`** — suppresses false positives in `reset()` and `type()` around the `__query_interface(__iunknown())` dereference chains.
- **`libcudacxx/include/cuda/__utility/__basic_any/rtti.h`** — suppresses false positive in `__try_vptr_cast` slow-path (down-cast / cross-cast) where `rtti` is known non-null by the `__iunknown` registration invariant.

## Checklist

### DIFF
--- a/.devcontainer/launch.sh
+++ b/.devcontainer/launch.sh
@@ -41,6 +41,22 @@ parse_options() {
     local -;
     set -euo pipefail;
 
+    # Must disable set -e temporarily. Per man getopt:
+    #
+    # [getopt -T] generates no output, and sets the error status to 4. Other
+    # implementations of getopt(1), and this version if the environment variable
+    # GETOPT_COMPATIBLE is set, will return '--' and error status 0.
+    set +e
+    getopt -T 2>&1 > /dev/null
+    getopt_ret=$?
+    set -e
+
+    if [[ "${getopt_ret}" != '4' ]]; then
+      echo "Must use enhanced (GNU) version of getopt which understand long options to use this script."
+      echo "Either your version of getopt does not support them or you have GETOPT_COMPATIBLE set."
+      exit 1
+    fi
+
     # Read the name of the variable in which to return unparsed arguments
     local UNPARSED="${!#}";
     # Splice the unparsed arguments variable name from the arguments list
--- a/c/parallel/test/test_segmented_sort.cpp
+++ b/c/parallel/test/test_segmented_sort.cpp
@@ -13,6 +13,7 @@
 #include <cstdlib>
 #include <optional> // std::optional
 #include <string>
+#include <type_traits>
 #include <vector>
 
 #include <cuda_runtime.h>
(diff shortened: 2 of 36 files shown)


pciolkosz:
Thanks for looking into it, it seems someone with cmake ownership needs to approve as well

libcudacxx/include/cuda/__utility/__basic_any/basic_any_value.h
```diff
@@ -348,9 +348,13 @@ struct _CCCL_TYPE_VISIBILITY_DEFAULT __basic_any : __basic_any_base<_Interface>
   {
     if (auto __vptr = __get_vptr())
     {
+      // GCC cannot prove __query_interface returns non-null for __iunknown (the invariant guarantees it)
+      _CCCL_DIAG_PUSH
+      _CCCL_DIAG_SUPPRESS_GCC("-Wnull-dereference")
```
miscco: Critical: some compilers have issues with warning suppression inside of class definitions. We should rather move the suppression outside to a single one at the top and one push at the bottom.

nirandaperera: @miscco what's the better practice here? Should I move the `PUSH` to the very top of the cpp file covering all methods? Or just cover these 2 methods?

Because, in the future, if some other warning needs to be suppressed, we should be able to see the existing warning suppressions.

miscco: Sorry, the message got lost. 

We should move the push to the top of the class and the pop after it

Issue #7877, opened 2026-03-04
[FEA]: Avoid EmptyKernel instantiation when possible


gevtushenko:
### Is this a duplicate?

- [x] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cccl/issues) for this request and that I agree to the [Code of Conduct](CODE_OF_CONDUCT.md)

### Area

CUB

### Is your feature request related to a problem? Please describe.

```c++
#include <cub/cub.cuh>
int main () {}
```

leads to:
```bash
cuobjdump -ptx a.out
.visible .entry _ZN3cub18CUB_300102_SM_12006detail11EmptyKernelIvEEvv()
{
   ret;
}
```

which leads to issues on user end when they use NVRTC and expect only one symbol in their TU. 

### Describe the solution you'd like

`EmptyKernel` should not be instantiated if parallel algorithms are not used.

PR #7878, opened 2026-03-04
Fix empty kernel

gevtushenko:
## Description

closes https://github.com/NVIDIA/cccl/issues/7877

This PR turns some of the CUB functions like `PtxVersion`  into templates to avoid instantiation of `EmptyKernel`. 

## Checklist

### DIFF
--- a/cub/test/cubin-check/CMakeLists.txt
+++ b/cub/test/cubin-check/CMakeLists.txt
@@ -34,6 +34,8 @@ function(cub_add_cubin_check_test source)
     ${target_name}
     PROPERTIES #
       CUDA_CUBIN_COMPILATION ON
+      # Some of our presets specify multiple architectures, so the hope here is to make the compilation a little faster.
+      # Given that empty kernel doesn't depend on the arch, that should be fine.
       CUDA_ARCHITECTURES "80-real"
   )
 
--- a/cub/test/ptx-json/CMakeLists.txt
+++ b/cub/test/ptx-json/CMakeLists.txt
@@ -59,6 +59,8 @@ function(cub_detail_ptx_json_add_test target_name_var source)
     ${target_name}
     PROPERTIES #
       CUDA_CUBIN_COMPILATION ON
+      # Some of our presets specify multiple architectures, so the hope here is to make the compilation a little faster.
+      # Given that ptx-json doesn't depend on the arch, that should be fine.
       CUDA_ARCHITECTURES "80-real"
   )
 


cub/test/cubin-check/CMakeLists.txt
```diff
@@ -0,0 +1,61 @@
...
+  message(STATUS "Skipping cubin-check tests for clang-cuda.")
+  return()
+endif()
+
+find_program(filecheck "FileCheck")
+
+if (NOT filecheck)
+  message(STATUS "Skipping cubin-check tests because FileCheck was not found.")
+  return()
+endif()
```
bernhardmgruber, line 16: Suggestion: This is inherently dangerous, because if `FileCheck` for some reason stops being available in our test containers, our unit tests continue to pass but we turn blind. I think we should require `FileCheck` and filter out more systems or platforms before the check, if required.

cub/test/cubin-check/CMakeLists.txt
```diff
@@ -0,0 +1,61 @@
...
+
+  add_library(${target_name} OBJECT "${source}")
+  cccl_configure_target(${target_name})
+  cccl_ensure_metatargets(${target_name})
+  target_link_libraries(${target_name} PRIVATE cub.compiler_interface)
+  set_target_properties(
+    ${target_name}
+    PROPERTIES #
+      CUDA_CUBIN_COMPILATION ON
+      CUDA_ARCHITECTURES "80-real"
```
bernhardmgruber, line 39: Question: Why do we have to set an architecture? Why can't we just use the one of the CI job?

cub/test/cubin-check/CMakeLists.txt
gevtushenko: Good point! There are more places that follow this pattern, so it'll be a larger change. It's safe to deliver the fix to the user and address FileCheck requirements separately in https://github.com/NVIDIA/cccl/issues/7881.

cub/test/cubin-check/CMakeLists.txt
gevtushenko: Some of our presets specify multiple architectures, so the hope here is to make the compilation a little faster. Given that empty kernel doesn't depend on the arch, that should be fine.

cub/test/cubin-check/CMakeLists.txt
bernhardmgruber: Alright. I was more thinking about just:

```suggestion
find_program(filecheck "FileCheck" REQUIRED)
```

cub/test/cubin-check/CMakeLists.txt
bernhardmgruber: Can you please add this rational as a comment here? Thx!

cub/test/cubin-check/CMakeLists.txt
gevtushenko: Yeah, I understand. I'm just not sure if the scope is limited to this. Do we have FileCheck in every place today, like DVS etc? I don't have an answer.

bernhardmgruber: @wmaxey raised similar concerns, so let's postpone the decision. It's tracked in #7881 anyways.

Issue #7871, opened 2026-03-03
Support <dlpack.h> include path in cuda/tma


cpcloud:
## Summary

`<cuda/tma>` currently pulls in DLPack via:

```cpp
#include <dlpack/dlpack.h>
```

For downstream consumers this can be awkward, because many projects vendor DLPack as a flat header (`dlpack.h`) rather than a `dlpack/` directory.

## Why this matters

In `cuda-python` (`cuda.core`) we are integrating `cuda::make_tma_descriptor` for TMA descriptor construction. We had to add an extra compatibility wrapper file at `dlpack/dlpack.h` that simply re-includes our existing vendored `dlpack.h`.

That wrapper is purely to satisfy include layout, not API/ABI behavior, and adds packaging/build maintenance overhead.

## Request

Would you consider one of these in CCCL:

1. Include `dlpack.h` directly in `make_tma_descriptor.h`, or
2. Support both include layouts with a fallback check, e.g.:

```cpp
#if __has_include(<dlpack.h>)
#  include <dlpack.h>
#elif __has_include(<dlpack/dlpack.h>)
#  include <dlpack/dlpack.h>
#else
#  error "DLPack header not found"
#endif
```

## Expected impact

- No functional behavior change in TMA descriptor logic.
- Better compatibility with downstreams that already provide `dlpack.h`.
- Avoids projects having to mirror headers into synthetic `dlpack/` trees.

Thanks for considering this.

fbusato:
Thanks for the suggestion. @cpcloud. I definitely agree with what you are proposing. I don't see problems by adding the dispatch based on `__has_include`. I will open a PR soon

PR #7884, opened 2026-03-04
Support DLPack inclusion for both `<dlpack/dlpack.h>` and `<dlpack.h>`

fbusato:
## Description

Close https://github.com/NVIDIA/cccl/issues/7871:  cuda-python (cuda.core) for `cuda::make_tma_descriptor` TMA descriptor construction.

### DIFF
--- a/libcudacxx/include/cuda/__internal/dlpack.h
+++ b/libcudacxx/include/cuda/__internal/dlpack.h
@@ -23,7 +23,11 @@
 
 #if _CCCL_HAS_DLPACK()
 
-#  include <dlpack/dlpack.h>
+#  if __has_include(<dlpack/dlpack.h>)
+#    include <dlpack/dlpack.h>
+#  elif __has_include(<dlpack.h>)
+#    include <dlpack.h>
+#  endif
 
 #  define _CCCL_DLPACK_AT_LEAST(_MAJOR, _MINOR) \
     (DLPACK_MAJOR_VERSION > (_MAJOR) || (DLPACK_MAJOR_VERSION == (_MAJOR) && DLPACK_MINOR_VERSION >= (_MINOR)))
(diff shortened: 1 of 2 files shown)


libcudacxx/include/cuda/__internal/dlpack.h
```diff
@@ -23,7 +23,11 @@
 
 #if _CCCL_HAS_DLPACK()
 
-#  include <dlpack/dlpack.h>
+#  if __has_include(<dlpack/dlpack.h>)
+#    include <dlpack/dlpack.h>
+#  elif __has_include(<dlpack.h>)
```
davebayer, line 28: Q: Shouldn't inclusion of `<dlpack.h>` be conditioned by some macro definition, since this is not the official way of including the header? For example `CCCL_USE_FLATTENED_DLPACK_INCLUDE`

fbusato: I don't think so. It looks a bit invasive. @cpcloud any preference?

cpcloud: Personally no preference, as long as there's some way we can write `#include <dlpack.h>`.

leofang: As a DLPack maintainer, I do not believe there is any "official way." In fact, we encourage each project to just vendor the header.

davebayer: By official, I mean all examples and tests in the repo use `#include <dlpack/dlpack.h>` and even if you install `libdlpack-dev` via `apt`, you get the file in `/usr/include/dlpack/dlpack.h`. So personally, I wouldn't expect `#include <dlpack.h>` to work.

PR #7887, opened 2026-03-05
Fix nvhpc `__builtin_(add|sub)_overflow` condition

davebayer:
The condition was wrong..


davebayer:
pre-commit.ci autofix

PR #7891, opened 2026-03-05
Disable `__builtin_mul_overflow` on ARM64

davebayer:


libcudacxx/include/cuda/__numeric/mul_overflow.h
```diff
@@ -53,6 +53,12 @@
 #  undef _CCCL_BUILTIN_MUL_OVERFLOW
 #endif // _CCCL_COMPILER(NVHPC, <, 26, 1)
 
+// On ARM64, using __builtin_mul_overflow can lead to `undefined reference to __muloti4` with clang and nvhpc, so let's
+// just disable the builtins on that platform.
+#if _CCCL_ARCH(ARM64) && (_CCCL_COMPILER(CLANG) || _CCCL_COMPILER(NVHPC))
+#  undef _CCCL_BUILTIN_MUL_OVERFLOW
```
miscco, line 59: Please define a macro that we can check in the implementations

fbusato: also, is this condition for nvhpc overlapping with the previous one?

davebayer: > Please define a macro that we can check in the implementations

What exactly do you mean?

> also, is this condition for nvhpc overlapping with the previous one?

Partially

miscco: I meant a _CCCL_ENABLE_BUILTIN_MUL_OVERFLOW() helper macro, the condition is getting complicated

PR #7899, opened 2026-03-05
[STF] Use cudaStreamGetId instead of manually assigning unique IDs to streams

caugonnet:
Stop manually assigning unique IDs to CUDA streams: we have had cu(da)StreamGetId since CUDA 12.0 so use it instead

## Description

closes 

## Checklist

### DIFF
--- a/cub/cub/device/device_radix_sort.cuh
+++ b/cub/cub/device/device_radix_sort.cuh
@@ -2477,6 +2477,96 @@ public:
   //! @rst
   //! Sorts keys into ascending order using :math:`\approx 2N` auxiliary storage.
   //!
+  //! .. versionadded:: 3.4.0
+  //!    First appears in CUDA Toolkit 13.4.
+  //!
+  //! This is an environment-based API that allows customization of:
+  //!
+  //! - Stream: Query via ``cuda::get_stream``
+  //! - Memory resource: Query via ``cuda::mr::get_memory_resource``
+  //!
+  //! - The contents of the input data are not altered by the sorting operation.
+  //! - Pointers to contiguous memory must be used; iterators are not currently
+  //!   supported.
+  //! - In-place operations are not supported. There must be no overlap between
+  //!   any of the provided ranges:
+  //!
+  //!   - ``[d_keys_in,  d_keys_in  + num_items)``
+  //!   - ``[d_keys_out, d_keys_out + num_items)``
+  //!
+  //! - An optional bit subrange ``[begin_bit, end_bit)`` of differentiating key
+  //!   bits can be specified. This can reduce overall sorting overhead and
+  //!   yield a corresponding performance improvement.
+  //!
+  //! Snippet
+  //! --------------------------------------------------
+  //!
+  //! The code snippet below illustrates the env-based sorting of keys:
+  //!
+  //! .. literalinclude:: ../../../cub/test/catch2_test_device_radix_sort_env_api.cu
+  //!     :language: c++
+  //!     :dedent:
+  //!     :start-after: example-begin radix-sort-keys-env
+  //!     :end-before: example-end radix-sort-keys-env
+  //!
+  //! @endrst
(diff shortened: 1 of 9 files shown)


cudax/include/cuda/experimental/__stf/internal/async_resources_handle.cuh
```diff
@@ -37,33 +37,62 @@
...
+inline unsigned long long get_stream_id(cudaStream_t stream)
+{
+  if (!stream)
+  {
+    return k_no_stream_id;
+  }
+  unsigned long long id = 0;
+  cuda_safe_call(cuStreamGetId(reinterpret_cast<CUstream>(stream), &id));
+  return id;
+}
```
davebayer, line 63: You can use `cuda::stream_ref::id()`

caugonnet: If this is just for cudax::stream_ref(stream).id() maybe it's adding not a small dependency just to avoid a single CUDA API call. Maybe the broader question is do i need "decorated_stream" which is a stream + some cached device id value + this id

cudax/include/cuda/experimental/__stf/internal/async_resources_handle.cuh
```diff
@@ -37,33 +37,62 @@
...
+
 namespace cuda::experimental::stf
 {
 class green_context_helper;
 
 // Needed to set/get affinity
 class exec_place;
 
+/** Sentinel for "no stream" / empty slot. Distinct from any value returned by cuStreamGetId. */
+inline constexpr unsigned long long k_no_stream_id = static_cast<unsigned long long>(-1);
```
andralex, line 50: I couldn't find a reference that -1 can't be a stream id... an overly cautious take would be:

```suggestion
inline const unsigned long long k_no_stream_id = [] {
  unsigned long long result;
  cuStreamGetId(nullptr, &result);
  return result;
}();
```

Not sure it's worth it.

caugonnet: it's going to mistake the sentinel with the null stream id. Maybe we just need to stop do this id computation lazily ?

cudax/include/cuda/experimental/__stf/internal/async_resources_handle.cuh
```diff
@@ -37,33 +37,62 @@
...
+/**
+ * @brief Returns the unique stream ID from the CUDA driver (cuStreamGetId).
+ * @param stream A valid CUDA stream, or nullptr.
+ * @return The stream's unique ID, or k_no_stream_id if stream is nullptr.
+ */
+inline unsigned long long get_stream_id(cudaStream_t stream)
+{
+  if (!stream)
+  {
+    return k_no_stream_id;
```
andralex: Hmm, this is not canon (and possibly confusing) seeing as the null stream is really the default process stream, which has id zero.

cudax/include/cuda/experimental/__stf/internal/async_resources_handle.cuh
```diff
@@ -37,33 +37,62 @@
...
+  }
+  unsigned long long id = 0;
+  cuda_safe_call(cuStreamGetId(reinterpret_cast<CUstream>(stream), &id));
```
andralex, line 60: Ah, one useful check to insert right after this:

```
assert(id != k_no_stream_id && "This is totally unexpected.");
```

caugonnet: added it

caugonnet:
A future extension to this PR will be to replace decorated_stream with a 1/ better name (annotated_stream) 2/ use CCCL's stream_ref facilities instead. As this is invasive we keep changes limited for now.

So i'm merging in the current state, and will consider it in future PRs

PR #7903, opened 2026-03-05
Improve `cub::PtxVersionUncached`

davebayer:


cub/cub/util_device.cuh
```diff
@@ -278,39 +278,16 @@ CUB_RUNTIME_FUNCTION cudaError_t PtxVersionUncached(int& ptx_version)
...
-     result = CubDebug(cudaFuncGetAttributes(&empty_kernel_attrs, reinterpret_cast<void*>(empty_kernel)));
-
-     ptx_version = empty_kernel_attrs.ptxVersion * 10;),
-    // NV_IS_DEVICE
-    (
-      // This is necessary to ensure instantiation of EmptyKernel in device
-      // code. The `reinterpret_cast` is necessary to suppress a
-      // set-but-unused warnings. This is a meme now:
-      // https://twitter.com/blelbach/status/1222391615576100864
-      (void) reinterpret_cast<EmptyKernelPtr>(empty_kernel);
```
bernhardmgruber, line 308: Q: Why is this workaround no longer needed?

davebayer: Because we force `EmptyKernel` instantiation in the upper section. And that one has `[[maybe_unused]]` attribute. I believe this just a pre c++17 relic

PR #7908, opened 2026-03-05
Add env RLE

### LINKED ISSUE #7547: Add env-based API for cub::DeviceRunLengthEncode
- [ ] Add env-based cub::DeviceRunLengthEncode::Encode()
- [ ] Add env-based cub::DeviceRunLengthEncode::NonTrivialRuns()

gonidelis:
fixes #7547

### DIFF
--- a/cub/test/catch2_test_device_run_length_encode_env.cu
+++ b/cub/test/catch2_test_device_run_length_encode_env.cu
@@ -1,5 +1,5 @@
 // SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
-// SPDX-License-Identifier: BSD-3-Clause
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 // Should precede any includes
 struct stream_registry_factory_t;
--- a/cub/test/catch2_test_device_run_length_encode_env_api.cu
+++ b/cub/test/catch2_test_device_run_length_encode_env_api.cu
@@ -1,5 +1,5 @@
 // SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
-// SPDX-License-Identifier: BSD-3-Clause
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 #include "insert_nested_NVTX_range_guard.h"
 


cub/cub/device/device_run_length_encode.cuh
```diff
@@ -203,6 +209,133 @@ struct DeviceRunLengthEncode
...
+  //!   @endrst
+  template <typename InputIteratorT,
+            typename UniqueOutputIteratorT,
+            typename LengthsOutputIteratorT,
+            typename NumRunsOutputIteratorT,
+            typename NumItemsT,
+            typename EnvT = ::cuda::std::execution::env<>,
+            typename ::cuda::std::enable_if_t<
+              ::cuda::std::is_integral_v<NumItemsT> && !::cuda::std::is_same_v<InputIteratorT, void*>,
+              int> = 0>
```
bernhardmgruber: No need to constrain `NumItemsT`
```suggestion
            ::cuda::std::enable_if_t<
              !::cuda::std::is_same_v<InputIteratorT, void*>,
              int> = 0>
```

cub/cub/device/device_run_length_encode.cuh
```diff
@@ -332,6 +465,120 @@ struct DeviceRunLengthEncode
...
+  //!   @endrst
+  template <typename InputIteratorT,
+            typename OffsetsOutputIteratorT,
+            typename LengthsOutputIteratorT,
+            typename NumRunsOutputIteratorT,
+            typename NumItemsT,
+            typename EnvT = ::cuda::std::execution::env<>,
+            typename ::cuda::std::enable_if_t<
+              ::cuda::std::is_integral_v<NumItemsT> && !::cuda::std::is_same_v<InputIteratorT, void*>,
+              int> = 0>
```
bernhardmgruber: Same here.

cub/cub/device/device_select.cuh
```diff
@@ -23,7 +23,6 @@
 #include <cub/device/dispatch/dispatch_select_if.cuh>
 #include <cub/device/dispatch/dispatch_unique_by_key.cuh>
 
-#include <cuda/__execution/determinism.h>
```
gonidelis, line 26: slipped in while on a workaround. let it pass it's a byfix

cub/cub/device/device_run_length_encode.cuh
```diff
@@ -332,6 +452,114 @@ struct DeviceRunLengthEncode
...
+    OffsetsOutputIteratorT d_offsets_out,
+    LengthsOutputIteratorT d_lengths_out,
+    NumRunsOutputIteratorT d_num_runs_out,
+    NumItemsT num_items,
+    EnvT env = {})
+  {
+    _CCCL_NVTX_RANGE_SCOPE("cub::DeviceRunLengthEncode::NonTrivialRuns");
+
+    using global_offset_t = detail::choose_signed_offset_t<NumItemsT>;
+    using equality_op     = ::cuda::std::equal_to<>;
```
miscco, line 548: I believe this is fine, because it always just returns a `bool` and does not promote integers

cub/cub/device/device_run_length_encode.cuh
```diff
@@ -203,6 +208,121 @@ struct DeviceRunLengthEncode
...
+    UniqueOutputIteratorT d_unique_out,
+    LengthsOutputIteratorT d_counts_out,
+    NumRunsOutputIteratorT d_num_runs_out,
+    NumItemsT num_items,
+    EnvT env = {})
+  {
+    _CCCL_NVTX_RANGE_SCOPE("cub::DeviceRunLengthEncode::Encode");
+
+    using equality_op              = ::cuda::std::equal_to<>;
+    using reduction_op             = ::cuda::std::plus<>;
```
miscco, line 301: Ditto: Should this rather be 
```suggestion
    using reduction_op             = ::cuda::std::plus<length_t>;
```

Otherwise this will always promote `offset_t` to a larger integer

cub/test/catch2_test_device_run_length_encode_env.cu
```diff
@@ -0,0 +1,233 @@
...
+TEST_CASE("DeviceRunLengthEncode::Encode uses custom stream", "[run_length_encode][device]")
+{
+  auto d_in           = c2h::device_vector<int>{0, 2, 2, 9, 5, 5, 5, 8};
+  auto d_unique_out   = c2h::device_vector<int>(8);
+  auto d_counts_out   = c2h::device_vector<int>(8);
+  auto d_num_runs_out = c2h::device_vector<int>(1);
+  int num_items       = static_cast<int>(d_in.size());
+
+  cudaStream_t custom_stream;
+  REQUIRE(cudaSuccess == cudaStreamCreate(&custom_stream));
```
bernhardmgruber, line 156: Suggestion: Please use `cuda::stream` so it gets higher test coverage. If you want to pass a `cudaStream_t`you can always call `.get()` (I think) on the `cuda::stream` to get the raw underlying stream.

This suggestion applies generally for all env-overload PRs.

cub/cub/device/device_run_length_encode.cuh
bernhardmgruber: The non-env overload where I guess this implementation comes from also uses `plus<>`. I agree that that's maybe not what we want, since `plus<length_t>` does not promote and influence the accumulator type. But this should be addressed in a separate PR.

cub/test/catch2_test_device_run_length_encode_env.cu
bernhardmgruber: @gonidelis since the PR auto-merged, please create a note or a tracking issue to replace all manual stream creation by `cuda::stream` in our unit tests.

PR #7913, opened 2026-03-06
Always use the latest devcontainer

miscco:
Updating the version step by step is cumbersome and leads to frequent issues


bernhardmgruber:
Won't automatic updates just let the CI fail spuriously and start blocking all PRs until we resolve any upgrade issues? I think we do want to control the upgrade ourselves.

jrhemstad:
> Don't automatic updates will just let the CI fail spuriously and start blocking all PRs until we resolve any upgrade issues? I think we do want to control the upgrade ourselves.

Agreed. I'd rather manually update the version at known, controlled points because implicitly switching to a new environment can cause just as many problems as it solves.

PR #7915, opened 2026-03-06
Support non-copyable stream types in `DeviceTransform`

bernhardmgruber:
This is a fix for another bug introduced in #6204

### DIFF
--- a/cub/cub/device/device_transform.cuh
+++ b/cub/cub/device/device_transform.cuh
@@ -182,7 +182,8 @@ public:
             typename... RandomAccessIteratorsOut,
             typename NumItemsT,
             typename TransformOp,
-            typename Env = ::cuda::std::execution::env<>>
+            typename Env = ::cuda::std::execution::env<>,
+            ::cuda::std::enable_if_t<!::cuda::std::is_convertible_v<Env, cudaStream_t>, int> = 0>
   CUB_RUNTIME_FUNCTION static cudaError_t Transform(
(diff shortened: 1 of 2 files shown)


cub/cub/device/device_transform.cuh
miscco, line 192: Strongly wondering whether it wouldnt be better to have an overload that forwards, but we should definitely make a definite decision on what we expect of environements

bernhardmgruber: I think taking the environment by `&&` would work as well here, yes. We should decide this.

cub/cub/device/device_transform.cuh
```diff
@@ -182,7 +182,8 @@ public:
             typename... RandomAccessIteratorsOut,
             typename NumItemsT,
             typename TransformOp,
-            typename Env = ::cuda::std::execution::env<>>
+            typename Env = ::cuda::std::execution::env<>,
+            ::cuda::std::enable_if_t<!::cuda::std::is_convertible_v<Env, cudaStream_t>, int> = 0>
```
pauleonix, line 186: Nit: Doesn't this make the docs look like one can't use the interface with `cudaStream_t` at all (since the new overload is not documented)?

bernhardmgruber: It kind of does. Let's fix this in a follow-up, since it concerns other CUB APIs as well.

PR #7906, opened 2026-03-05
[STF] Introduce a per-place container

caugonnet:
## Description

closes 

## Checklist

- [ ] New or existing tests cover these changes.
- [ ] The documentation is up to date with these changes.

PR #7917, opened 2026-03-06
[STF] Rework how execution place select their stream_pool

caugonnet:
## Description

To remove the dependency of places on async_resources_handle, we need to avoid using async_resources_handle in the place virtual methods. We therefore use a stream_pool container, that we will also store in the async_resources_handle.

This PR sits on top of the stf_per_place_container branch (PR #7906) which introduces place_indexed_container

closes 

## Checklist


caugonnet:
We now have a much better strategy, closing

Issue #7821, opened 2026-02-26
[BUG]: Spurious failures in cub.test.device.scan_alignment.lid_0


miscco:
### Is this a duplicate?

- [x] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cccl/issues) for this bug and that I agree to the [Code of Conduct](CODE_OF_CONDUCT.md)

### Type of Bug

Runtime Error

### Component

CUB

### Describe the bug

We are seeing [spurious ](https://github.com/NVIDIA/cccl/actions/runs/22452389642/job/65031381113?pr=7776#step:4:701) CI failures in ` cub.test.device.scan_alignment`

```
72/152 Test  #78: cub.test.device.scan_alignment.lid_0 ...........................***Failed    1.22 sec
  -- >> Running:
  	/home/coder/cccl/build/cuda13.1-gcc14/cub-lid0/bin/cub.test.device.scan_alignment.lid_0
  Randomness seeded to: 604188638
  
  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  cub.test.device.scan_alignment.lid_0 is a Catch2 v3.12.0 host application.
  Run with -? for options
  
  -------------------------------------------------------------------------------
  "Device scan works with all device interfaces"(0) - types_31 - 0
  -------------------------------------------------------------------------------
  /home/coder/cccl/cub/test/catch2_test_device_scan_alignment.cu:31
  ...............................................................................
  
  /home/coder/cccl/cub/test/catch2_test_launch_helper.h:173: FAILED:
    REQUIRE( temp_storage_bytes > 0 )
  with expansion:
    0 > 0
  with messages:
    num_items := 0
    offset := 51
```

We need to investigate where those are coming from and fix the issue

### How to Reproduce

Run the `cub.test.device.scan_alignment.lid_0` with the seed 604188638

### Expected behavior

The test should pass

bernhardmgruber:
There is a good chance this is coming from the warpspeed scan, so we must address this for the CCCL 3.4 release.

bernhardmgruber:
Using `cub.test.device.scan_alignment.lid_0 --rng-seed 604188638` I can reproduce this locally on my RTX 5090 as well.

bernhardmgruber:
One reason is that we don't handle `num_items == 0`, but I still see failures for `num_items == 1`.

PR #7921, opened 2026-03-06
Fix tiny problem sizes for warpspeed scan

bernhardmgruber:
Fixes: #7821

### DIFF
--- a/cub/cub/detail/warpspeed/squad/load_store.cuh
+++ b/cub/cub/detail/warpspeed/squad/load_store.cuh
@@ -236,7 +236,7 @@ squadStoreBulkSync(Squad squad, CpAsyncOobInfo<OutputT> cpAsyncOobInfo, const ::
 
     constexpr ::cuda::std::uint16_t byteMask  = 0xFFFF;
     const ::cuda::std::uint16_t byteMaskStart = byteMask << cpAsyncOobInfo.smemStartSkipBytes;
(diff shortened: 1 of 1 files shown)


cub/cub/detail/warpspeed/squad/load_store.cuh
```diff
@@ -236,7 +236,7 @@ squadStoreBulkSync(Squad squad, CpAsyncOobInfo<OutputT> cpAsyncOobInfo, const ::
 
     constexpr ::cuda::std::uint16_t byteMask  = 0xFFFF;
     const ::cuda::std::uint16_t byteMaskStart = byteMask << cpAsyncOobInfo.smemStartSkipBytes;
-    const ::cuda::std::uint16_t byteMaskEnd   = byteMask >> (16 - cpAsyncOobInfo.smemEndBytesAfter16BBoundary);
+    const ::cuda::std::uint16_t byteMaskEnd   = byteMask >> (16 - cpAsyncOobInfo.smemEndBytesAfter16BBoundary) % 16;
```
bernhardmgruber, line 239: @ahendriksen is there any smarter way to shift `byteMask` by `smemEndBytesAfter16BBoundary`, but leave it when it's 16? Would a predicated shift be faster?

ahendriksen: Can't think of a quicker way off the top of my head. I always hoped that the compiler would figure out the best way to compute all these values. 

A modulo 16 operation is just and `AND` by `~0xF`, which should be quicker than computing the predicate.

bernhardmgruber: Alright, thx!

PR #7922, opened 2026-03-06
[libcu++] Improve copy_bytes and dynamic_shared_memory documentation

pciolkosz:
`copy_configuration` was not mentioned in the `copy_bytes` doc and there was no mention of `non_portable` argument to `dynamic_shared_memory` option. This PR also tries to provide a bit more interesting examples for `copy_bytes` and `fill_bytes` beyond just packing pointers into a span. Showing some span manipulation might highlight why its better than `cudaMemcpy` on plain pointers

### DIFF
--- a/docs/libcudacxx/runtime/algorithm.rst
+++ b/docs/libcudacxx/runtime/algorithm.rst
@@ -12,10 +12,11 @@ The ``runtime`` part of the ``cuda/algorithm`` header provides stream-ordered, b
 
 Launch a byte-wise copy from source to destination on the provided stream.
 
+- Signature: ``copy_bytes(stream, src, dst, config = {})``
 - Overloads accept ``cuda::std::span``-convertible contiguous ranges or ``cuda::std::mdspan``-convertible multi-dimensional views.
 - Elements must be trivially copyable
 - ``cuda::std::mdspan``-convertible types must convert to an mdspan that is exhaustive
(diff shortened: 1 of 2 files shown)


docs/libcudacxx/runtime/algorithm.rst
```diff
@@ -12,23 +12,37 @@ The ``runtime`` part of the ``cuda/algorithm`` header provides stream-ordered, b
...
+     auto src_suffix = src.subspan(n / 2);
+     auto dst_suffix = dst.subspan(n / 2);
+
+     // Default behavior: enqueue a stream-ordered byte-wise copy on stream s.
+     cuda::copy_bytes(s, src_prefix, dst_prefix);
+
+     // Advanced behavior: customize source access order for this copy.
+     auto config = cuda::copy_configuration{
+       .src_access_order = cuda::source_access_order::during_api_call,
+     };
```
miscco, line 44: Question: are there any other configurations, we should at least name them here I believe

pciolkosz: I will work on improving links in these docs, so we keep this page brief in terms of just usage of the API and instead it links to the API reference for the full list of options.

pciolkosz: Filed https://github.com/NVIDIA/cccl/issues/7963

PR #7929, opened 2026-03-09
[STF] PIMPL my stream_pool

caugonnet:
## Description

To facilitate the design of independent place abstractions, we need to improve the stream_pool design

closes 

## Checklist

### DIFF
--- a/cudax/test/stf/examples/01-axpy-places.cu
+++ b/cudax/test/stf/examples/01-axpy-places.cu
@@ -84,9 +84,7 @@ void run()
 
   /* Compute Y = Y + alpha X */
   auto t = ctx.task(all_devs, handle_X.read(cdp), handle_Y.rw(cdp));
-  t->*[&](auto stream, auto sX, auto sY) {
-    // should be nullptr as we did not set a place
-    // fprintf(stderr, "t.get_stream() = %p\n", t.get_stream());
+  t->*[&](auto, auto sX, auto sY) {
     size_t grid_size = t.grid_dims().size();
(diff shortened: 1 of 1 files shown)


cudax/include/cuda/experimental/__stf/internal/async_resources_handle.cuh
```diff
@@ -243,34 +120,27 @@ private:
...
-      // stream_pool class to store CUDA streams for computation and another
-      // for data transfers)
-      for (auto& p : pool)
-      {
-        stream_pool_cleanup_internal(p.first);
-        stream_pool_cleanup_internal(p.second);
-      }
+      // stream_pool uses shared_ptr internally; streams are destroyed
+      // when the last copy of each pool goes away.
     }
```
miscco: I believe we can just drop that then?

caugonnet: yes indeed, thanks for catching it !

caugonnet:
```
  97/235 Test  #97: cudax.test.stf.examples.01-axpy-places ........................................Subprocess aborted***Exception:   1.39 sec
  terminate called after throwing an instance of 'cuda::experimental::stf::expecter::failure'
    what():  /home/coder/cccl/cudax/test/stf/examples/01-axpy-places.cu(108): 3.13468 < 0.0001 is false.
```
  
Some transient error had to be fixed ... unsure it was there before or not

=> fixed the example that was using the wrong stream

PR #7931, opened 2026-03-09
Expose PSTL algorithms through `<cuda/std/execution>`

miscco:
We discussed this internally and we are happy with the results of the parallel CUDA backend. So we want to expose this now rather than waiting for all algorithms to be implemented.

There are certain caveats:

* We require random access iterators for the CUDA backend

* We do not expose only a CUDA backend through `cuda::execution::gpu`. Standard execution policies will currently static_assert that there is a missing backend

* We do not provide any fallback serial implementation. This would be dangerous, because the serial implementation would naively run on host and not device.

- [x] Merge first: #8155


libcudacxx/benchmarks/bench/remove_copy/basic.cu
```diff
@@ -24,7 +24,7 @@ static void basic(nvbench::state& state, nvbench::type_list<T>)
 
   thrust::device_vector<T> in = generate(elements, bit_entropy::_1_000, T{0}, T{42});
   thrust::device_vector<T> out(elements, thrust::no_init);
-  const auto count = cuda::std::count(cuda::execution::__cub_par_unseq, in.begin(), in.end(), T{42});
+  const auto count = cuda::std::count(cuda::execution::par_unseq, in.begin(), in.end(), T{42});
```
davebayer: Will any execution policy other than parallel + unsequenced make sense on device? I feel like much more important piece information would be that it's a device execution policy, than that it's parallel and unsequenced.

Wouldn't `cuda::execution::device` and `cuda::execution::host` make more sense?

davebayer: We could document that these are always parallel and unsequenced

libcudacxx/include/cuda/__execution/policy.h
```diff
@@ -223,7 +223,7 @@ _CCCL_BEGIN_NAMESPACE_CUDA_EXECUTION
 using __cub_parallel_unsequenced_policy =
   ::cuda::std::execution::__execution_policy_base<::cuda::std::execution::__with_cuda_backend<static_cast<uint32_t>(
     ::cuda::std::execution::__execution_policy::__parallel_unsequenced)>()>;
-_CCCL_GLOBAL_CONSTANT __cub_parallel_unsequenced_policy __cub_par_unseq{};
+_CCCL_GLOBAL_CONSTANT __cub_parallel_unsequenced_policy par_unseq{};
```
bernhardmgruber: Q: Was this the final conclusion for the name of the CUDA GPU backend policy? There was a Slack thread recently, but I don't see this name mentioned there.

libcudacxx/benchmarks/bench/remove_copy/basic.cu
bernhardmgruber: I think we are aiming to be close to the std policies. But `cuda::execution::gpu` was discussed as well. There is a chance we are needing an OpenMP backend in the future, so a `host` policy doesn't make so much sense.

libcudacxx/include/cuda/__execution/policy.h
miscco: The plan was to use a policy that is close to the standard ones

Like there is thrust::device, but it doesnt really do the thing you expect iff your backend is TBB

libcudacxx/benchmarks/bench/remove_copy/basic.cu
miscco: I dont believe there is a simple  solution here. We could require the user to go through something like 

`cuda::std::execution::par_unseq.backend(CUDA)` but I am not sure whether that is actually better

libcudacxx/test/libcudacxx/std/algorithms/alg.modifying/alg.copy/pstl_copy_n.cu
```diff
@@ -94,7 +94,7 @@ C2H_TEST("cuda::std::copy_n", "[parallel algorithm]")
   {
     cuda::stream stream{cuda::device_ref{0}};
     cuda::device_memory_pool_ref device_resource = cuda::device_default_memory_pool(stream.device());
-    const auto policy = cuda::execution::__cub_par_unseq.with_memory_resource(device_resource).with_stream(stream);
+    const auto policy = cuda::execution::gpu.with_memory_resource(device_resource).with_stream(stream);
```
bernhardmgruber: Remark: looking at this line, it may sound a bit better as:
```suggestion
    const auto policy = cuda::execution::gpu.with_memory_resource(device_resource).on_stream(stream);
```
But I guess nobody wants to do another bulk rename?

miscco: There might be another rename coming but not that one

libcudacxx/include/cuda/std/algorithm
```diff
@@ -131,6 +131,42 @@
...
 #include <cuda/std/__algorithm/upper_bound.h>
 
+// parallel algorithms
+#if _CCCL_HAS_PSTL_BACKEND()
+#  include <cuda/std/__pstl/adjacent_find.h>
+#  include <cuda/std/__pstl/all_of.h>
+#  include <cuda/std/__pstl/any_of.h>
+#  include <cuda/std/__pstl/copy.h>
+#  include <cuda/std/__pstl/copy_if.h>
+#  include <cuda/std/__pstl/copy_n.h>
```
bernhardmgruber: Q: I thought many standard libraries would expose PSTL algorithms through the `<execution>` header and not `<algorithm>`. This would make the inclusion of `<algorithm>` cheaper.

bernhardmgruber: Discussed this with @miscco offline and it seems the C++ standard requires the overloads to be in `<algorithm>`. However, it may not be observable to the common user, since they need to include `<execution>` in addition to supply an execution policy.

bernhardmgruber:
@miscco could you please measure the compile-time of
```c++
#include <cuda/std/algorithm>
int main() {
  return cuda::std::min(0, 2);
}
```
before and after this PR? I would be curious how much of an impact pulling in most of CUB has ;)

libcudacxx/include/cuda/std/algorithm
jrhemstad: If it's not observable, then I would like to see exposing it in the `<execution>` header to avoid bloating `<algorithm>`.

miscco: I do not believe that is a correct statement. 

`<execution>` can include it all and be fine, but then `<algorithm>` would not have it.

The point is that the pstl headers pull effectively all of `<algorithm>`

bernhardmgruber: > <execution> can include it all and be fine, but then <algorithm> would not have it.

Why is the advantage of `<algorithm>` having an overload that cannot be called if a user does not also include `<execution>`?

> The point is that the pstl headers pull effectively all of <algorithm>

This is fine IMO, including a PSTL header can be more expensive.

miscco: Moved the exposure to `<cuda/std/execution>` in the hope of then being able to expose via a modularized access

miscco:
I checked the differences in compile times for the `<cuda/std/algorithm>`, `<cuda/std/numeric>` and `<cuda/std/execution>` header compared against main

Main:
* `<cuda/std/algorithm>`: 0.709372
* `<cuda/std/numeric>`:  0.545339
* `<cuda/std/execution>`:  1.004296

With PSTL
* `<cuda/std/algorithm>`: 2.118540 -> 300%
* `<cuda/std/numeric>`:  2.471399  ->  450%
* `<cuda/std/execution>`:  5.873960 ->587%

The reason that `<cuda/std/execution>` is hit so bad is that it *also* has to include all of the serial algorithms too

bernhardmgruber:
> Main:
> 
> * `<cuda/std/algorithm>`: 0.709372
> * `<cuda/std/numeric>`:  0.545339
> * `<cuda/std/execution>`:  1.004296
> 
> With PSTL
> 
> * `<cuda/std/algorithm>`: 2.118540 -> 300%
> * `<cuda/std/numeric>`:  2.471399  ->  450%
> * `<cuda/std/execution>`:  5.873960 ->587%

Thank you for providing these numbers! I think we should go ahead with the status quo and expose the PSTL through `<cuda/std/execution>` to not burden existing users including `<cuda/std/algorithm>` and `<cuda/std/numeric>`.

miscco:
We discussed this today in the our internal review meeting and decided to better align the customization with the other environment work

docs/libcudacxx/standard_api/algorithms_library.rst
```diff
@@ -31,3 +34,57 @@ Restrictions
...
+
+Parallel standard algorithms
+----------------------------
+
+CCCL provides an implementation for the standard `parallel algorithms library <http://www.eel.is/c++draft/algorithms.parallel>`_
+
+Currently the CUDA backend is the only supported backend. It can be selected by passing the `cuda::execution::gpu`
+execution policy to one of the supported algorithms. The CUDA backend requires the passed in sequences to reside in
+device accessible memory and the iterators into those sequences to be at least random access iterators. The CUDA backend
+is enabled if the program im compiled with a CUDA compile in CUDA mode.
```
davebayer:
```suggestion
is enabled if the program is compiled with a CUDA compiler in CUDA mode.
```

libcudacxx/include/cuda/std/__execution/policy.h
```diff
@@ -68,6 +70,126 @@ struct __execution_policy_base
...
+        ::cuda::std::forward<_Env>(__env), ::cuda::std::__get<_Indices>(this->__envs_)...};
+    }
+  }
+
+  //! @brief Prepend an environment to the current ones
+  template <class _Env>
+  [[nodiscard]] _CCCL_API constexpr auto with(_Env&& __env) const
+  {
+#if _CCCL_HAS_CTK() && !_CCCL_COMPILER(NVRTC)
+    if constexpr (__convertible_to_stream_ref<_Env> && (is_lvalue_reference_v<_Env> || is_pointer_v<_Env>) )
```
davebayer: Shouldn't the check for `cudaStream_t` be more precise?
```suggestion
    if constexpr (__convertible_to_stream_ref<_Env> && (is_lvalue_reference_v<_Env> || is_same_v<::cudaStream_t, remove_cvref_t<_Env>>) )
```

libcudacxx/include/cuda/std/__execution/policy.h
```diff
@@ -68,6 +70,126 @@ struct __execution_policy_base
...
+    else
+    {
+      return __execution_policy_base<_Policy, _Env, _Envs...>{
+        ::cuda::std::forward<_Env>(__env), ::cuda::std::__get<_Indices>(this->__envs_)...};
+    }
+  }
+
+  //! @brief Prepend an environment to the current ones
+  template <class _Env>
+  [[nodiscard]] _CCCL_API constexpr auto with(_Env&& __env) const
```
davebayer: Don't we want to separate all of the `if constexpr` cases in separate overloads?

libcudacxx/include/cuda/std/__execution/policy.h
```diff
@@ -68,6 +70,126 @@ struct __execution_policy_base
...
+  {
+#if _CCCL_HAS_CTK() && !_CCCL_COMPILER(NVRTC)
+    if constexpr (__convertible_to_stream_ref<_Env> && (is_lvalue_reference_v<_Env> || is_pointer_v<_Env>) )
+    { // streams are special in that they are their own environment, but we always want to store a stream_ref
+      // We must reject prvalue cuda::stream because they are not copyable
+      static_assert(!is_same_v<remove_cvref_t<_Env>, ::cuda::stream> || is_lvalue_reference_v<_Env>,
+                    "cuda::stream is not copyable. It must be passed as a cuda::stream_ref");
```
davebayer, line 103: Can't we just `= delete` that case?

libcudacxx/include/cuda/std/__fwd/execution_policy.h
```diff
@@ -63,9 +63,26 @@ template <uint32_t _Policy>
...
+  constexpr uint32_t __backend_mask{0xFFFF00FF};
+  constexpr uint32_t __new_policy = (_Policy & __backend_mask) | (static_cast<uint32_t>(_Backend) << 8);
+  return __new_policy;
+}
+
+//! @brief Checks whether a policy supports the cuda backend
+template <uint32_t _Policy, __execution_backend _Backend>
+[[nodiscard]] _CCCL_API constexpr bool __has_backend() noexcept
+{
+  constexpr uint32_t __backend_mask{0xFFFF00FF};
```
davebayer, line 82: Can we unify `__backend_mask` definition somewhere?

libcudacxx/include/cuda/std/__internal/pstl_config.h
```diff
@@ -27,6 +27,8 @@
 #define _CCCL_HAS_BACKEND_OMP()  0
 #define _CCCL_HAS_BACKEND_TBB()  0
 
+#define _CCCL_HAS_PSTL_BACKEND() _CCCL_HAS_BACKEND_CUDA() || _CCCL_HAS_BACKEND_OMP() || _CCCL_HAS_BACKEND_TBB()
```
davebayer:
```suggestion
#define _CCCL_HAS_PSTL_BACKEND() (_CCCL_HAS_BACKEND_CUDA() || _CCCL_HAS_BACKEND_OMP() || _CCCL_HAS_BACKEND_TBB())
```

libcudacxx/include/cuda/std/__execution/policy.h
miscco: No because a user could pass any resource. Also it would require a bunch of overloads

libcudacxx/include/cuda/std/__execution/policy.h
miscco: I believe its better to have a clear error message rather than rrying to figure out what went wrong

libcudacxx/include/cuda/std/__fwd/execution_policy.h
miscco: Its defined twice right now, so I believe we can just keep it for now

PR #7933, opened 2026-03-09
Improve this hierarchy groups

davebayer:


cudax/include/cuda/experimental/__hierarchy/this_group.cuh
```diff
@@ -109,19 +118,28 @@ public:
...
-  -> thread_group<__hierarchy_type_of<_Hierarchy>, __this_hierarchy_group_kind>;
+_CCCL_HOST_DEVICE this_thread(const _Hierarchy&) -> this_thread<__hierarchy_type_of<_Hierarchy>>;
+
+#if _CCCL_HAS_COOPERATIVE_GROUPS()
+_CCCL_HOST_DEVICE this_thread(const ::cooperative_groups::thread_block_tile<1, void>&) -> this_thread<>;
+#endif // _CCCL_HAS_COOPERATIVE_GROUPS()
 
 template <class _Hierarchy>
-class warp_group<_Hierarchy, __this_hierarchy_group_kind> : __this_hierarchy_group_base<warp_level, _Hierarchy>
+class this_warp : __this_group_base<warp_level, _Hierarchy>
```
pciolkosz, line 141: Should this be constructible from `thread_block_tile<32>`?

cudax/include/cuda/experimental/__hierarchy/this_group.cuh
```diff
@@ -109,19 +118,28 @@ public:
   using __base_type::rank;
   using __base_type::rank_as;
 
+#  if _CCCL_HAS_COOPERATIVE_GROUPS()
+  _CCCL_DEVICE_API this_thread(const ::cooperative_groups::thread_block_tile<1, void>&) noexcept
```
pciolkosz: Hmm, should this be a template `::cooperative_groups::thread_block_tile<1, T>`? Otherwise if someone just creates a tile it will say `thread_block` in place of T and there is an implicit conversion to the void templated `thread_block_tile`, but then it won't be able to implicitly convert to `this_thread`

pciolkosz (approved):

cudax/include/cuda/experimental/__hierarchy/fwd.cuh
```diff
@@ -21,49 +21,59 @@
 #  pragma system_header
 #endif // no system header
 
+// Q: Do we want to enable this by default, or do we want the user to define some macro to get the interoperability with
+//    cooperative groups?
+#if __has_include(<cooperative_groups.h>)
```
pciolkosz, line 26: I don't think I like this solution long term, but for now its fine

cudax/include/cuda/experimental/__hierarchy/fwd.cuh
```diff
@@ -21,49 +21,59 @@
...
+
 #include <cuda/std/__cccl/prologue.h>
 
 namespace cuda::experimental
 {
-// hierarchy group kinds
-
-class __this_hierarchy_group_kind
-{};
+using __implicit_hierarchy_t =
```
pciolkosz, line 39: Would it make sense to instead have the hierarchy query on a group return an optional and not have the implicit hierarchy? Mostly thinking out loud, because the implicit one is not the most useful and if the compiler can't optimize it out in some cases it might query a bunch of stuff for now reason

cudax/include/cuda/experimental/__hierarchy/fwd.cuh
davebayer: That's why I added the comment, I think the compatibility with cooperative groups should be included on demand by the user

cudax/include/cuda/experimental/__hierarchy/fwd.cuh
davebayer: From what I saw, the compiler is really good at eliminating the unnecessary special registers reads.

And what would it actually mean to return an `optional`? That a group is optionally part of some hierarchy? But the thing is that it's always part of some hierarchy. On top of that, the internals of the groups would have to use 2 types of queries (`unit.query(level)` and `unit.query(level, hierarchy)`) based on whether the hierarchy is implicit or not.

So I would much rather start with this solution and then see whether it's a actually problem

PR #7934, opened 2026-03-09
[STF] Move stream pools to execution places instead of async_resources_handle

caugonnet:
## Description

This improves how we can design independent place concepts rather than the akward need of an async_resources_handle objects which belongs more to STF internals

This relies on #7929

closes 

## Checklist


caugonnet:
Was merged as part of another PR

PR #7935, opened 2026-03-09
Make all algorithms in `cuda/std/algorithm` available through `cuda/std/algorithm.meow.h`

miscco:
We had a [poll](https://github.com/NVIDIA/cccl/discussions/7905) to see how users would like to access individual features of umbrella headers like `<cuda/std/algorithm>`

The vast majority preferred a pattern like `<cuda/std/algorithms/find.h>`

However, we can get away with finding a similar but different name if we use the pattern `umbrella_header.feature`

We can create a wrapper for the internal file and also instruct our IDE to ignore all files that match the pattern `[a-z_]*\.[a-z_]*` This works because we do not have any `.h` files in neither `<cuda/>` not `<cuda/std/>`

### DIFF
--- a/docs/libcudacxx/standard_api/algorithms_library.rst
+++ b/docs/libcudacxx/standard_api/algorithms_library.rst
@@ -22,6 +22,8 @@ Extensions
 
   - All supported algorithms are available from C++17 onwards.
   - All supported algorithms are constexpr, except the allocating ones.
+  - Because `<cuda/std/algorithm>` is a huge header with a considerable compile-time cost, we provide each algorithm
+    through a minimal subheader named e.g `<cuda/std/algorithm.find.h>`
 
 Restrictions
 ------------
@@ -32,7 +34,7 @@ Restrictions
     * ``inplace_merge``
     * ``nth_element``
     * ``sort``
-    * ``stable_partiin``
+    * ``stable_partition``
     * ``stable_sort``
 
 Parallel standard algorithms
--- a/libcudacxx/include/cuda/std/algorithm.adjacent_find.h
+++ b/libcudacxx/include/cuda/std/algorithm.adjacent_find.h
@@ -1,0 +1,26 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of libcu++, the C++ Standard Library for your entire system,
+// under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef _CUDA_STD_ALGORITHM_ALGORITHM_ADJACENT_FIND_H
+#define _CUDA_STD_ALGORITHM_ALGORITHM_ADJACENT_FIND_H
+
+#include <cuda/std/detail/__config>
+
+#if defined(_CCCL_IMPLICIT_SYSTEM_HEADER_GCC)
+#  pragma GCC system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_CLANG)
+#  pragma clang system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_MSVC)
+#  pragma system_header
+#endif // no system header
+
+#include <cuda/std/__algorithm/adjacent_find.h>
+
+#endif // _CUDA_STD_ALGORITHM_ALGORITHM_ADJACENT_FIND_H
--- a/libcudacxx/include/cuda/std/algorithm.ranges.find_if.h
+++ b/libcudacxx/include/cuda/std/algorithm.ranges.find_if.h
@@ -1,0 +1,26 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of libcu++, the C++ Standard Library for your entire system,
+// under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef _CUDA_STD_ALGORITHM_ALGORITHM_RANGES_FIND_IF_H
+#define _CUDA_STD_ALGORITHM_ALGORITHM_RANGES_FIND_IF_H
+
+#include <cuda/std/detail/__config>
+
+#if defined(_CCCL_IMPLICIT_SYSTEM_HEADER_GCC)
+#  pragma GCC system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_CLANG)
+#  pragma clang system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_MSVC)
+#  pragma system_header
+#endif // no system header
+
+#include <cuda/std/__algorithm/ranges_find_if.h>
+
+#endif // _CUDA_STD_ALGORITHM_ALGORITHM_RANGES_FIND_IF_H
--- a/libcudacxx/test/libcudacxx/libcxx/module_headers/cuda/std/algorithm/iter_swap.pass.cpp
+++ b/libcudacxx/test/libcudacxx/libcxx/module_headers/cuda/std/algorithm/iter_swap.pass.cpp
@@ -1,0 +1,44 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of libcu++, the C++ Standard Library for your entire system,
+// under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES.
+//
+//===----------------------------------------------------------------------===//
+
+#include <cuda/std/algorithm.iter_swap.h>
+#include <cuda/std/cassert>
+
+#include "test_macros.h"
+
+struct iter_swap_iter
+{
+  int v;
+};
+
+__host__ __device__ constexpr void iter_swap(iter_swap_iter* a, iter_swap_iter* b)
+{
+  cuda::std::swap(a->v, b->v);
+}
+
+__host__ __device__ constexpr bool test()
+{
+  int x = 1, y = 2;
+  cuda::std::iter_swap(&x, &y);
+  assert(x == 2 && y == 1);
+  iter_swap_iter p{3}, q{4};
+  cuda::std::iter_swap(&p, &q);
+  assert(p.v == 4 && q.v == 3);
+
+  return true;
+}
+
+int main(int, char**)
+{
+  test();
+  static_assert(test());
+
+  return 0;
+}
--- a/libcudacxx/test/libcudacxx/libcxx/module_headers/cuda/std/algorithm/search.pass.cpp
+++ b/libcudacxx/test/libcudacxx/libcxx/module_headers/cuda/std/algorithm/search.pass.cpp
@@ -1,0 +1,53 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of libcu++, the C++ Standard Library for your entire system,
+// under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES.
+//
+//===----------------------------------------------------------------------===//
+
+#include <cuda/std/algorithm.search.h>
+#include <cuda/std/cassert>
+
+#include "test_macros.h"
+
+struct equal_to
+{
+  __host__ __device__ constexpr bool operator()(int a, int b) const
+  {
+    return a == b;
+  }
+};
+
+struct searcher
+{
+  const int *pf, *pl;
+  template <class It>
+  __host__ __device__ constexpr cuda::std::pair<It, It> operator()(It f, It l) const
+  {
+    return cuda::std::__search(
+      f, l, pf, pl, equal_to{}, cuda::std::random_access_iterator_tag{}, cuda::std::random_access_iterator_tag{});
+  }
+};
+
+__host__ __device__ constexpr bool test()
+{
+  constexpr int h[] = {1, 2, 3, 4};
+  constexpr int n[] = {2, 3};
+  assert(cuda::std::search(h, h + 4, n, n + 2) == h + 1);
+  assert(cuda::std::search(h, h + 4, n, n + 2, equal_to{}) == h + 1);
+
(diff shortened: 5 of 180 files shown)


miscco:
I am proposing to keep the current folder structure intact and ~~only add symlinks~~ EDIT: discussed this internally and we need real files for windows

For me this greatly improves maintainability of our code base. Otherwise there are 500+ files in `<cuda/std/>` and no proper way of organizing them

bernhardmgruber (approved):

davebayer (approved):

bernhardmgruber (approved):

.devcontainer/cuda12.0-gcc10/devcontainer.json
```diff
@@ -63,7 +63,15 @@
...
-        "files.trimTrailingWhitespace": true
+        "files.exclude": {
+          "**/include/cuda/*.*": true,
+          "**/include/cuda/std/*.*": true
+        },
+        "files.trimTrailingWhitespace": true,
+        "search.exclude": {
+          "**/include/cuda/*.*": true,
+          "**/include/cuda/std/*.*": true
+        }
```
miscco: Note this one here might be a tad dangerous, it also removes the files for search paths in the VS-code IDE

We might want to drop that

jrhemstad:
> I am proposing to keep the current folder structure intact and only add symlinks

I don't think symlinks work on Windows, do they?

.devcontainer/cuda12.0-gcc10/devcontainer.json
jrhemstad: Can you elaborate on this? What was this added for?

jrhemstad:
Given this is a pretty extensive product facing change, I'd like more thoughts from @wmaxey @griwes @gevtushenko @brycelelbach @ericniebler

.devcontainer/cuda12.0-gcc10/devcontainer.json
miscco: These additional files add a considerable maintenance burden on the `cuda/std` folder. 

What is now relatively well contained will contain a few hundred headers soon. vs-code and other IDEs offer exclude files, where one can hide those files.

vs-code in particular has 2 options
* hide files from explorer view
* hide files from search view -> searching in the repository will not show those files

miscco:
> > I am proposing to keep the current folder structure intact and only add symlinks
> 
> I don't think symlinks work on Windows, do they?

They dont, moved to forwarding headers

jrhemstad (changes requested):
Some questions I'd like answered before we merge this:
- How will we test this? I think we should have tests that include each micro-header and ensure that was is supposed to be available in that header is indeed available.
- How will we document this? 
- I'd like to see an evaluation of how well this naming pattern will extend to other headers

miscco:
Here a file with the compile time cost of `<cuda/std/algorithm>` compared with all the individual subheaders:
[algorithm_include_cost.csv](https://github.com/user-attachments/files/26178719/algorithm_include_cost.csv)

100 algorithms about 80 take less than 1% of the include time of `<cuda/std/algorithm>`

Regarding the questions above
> How will we test this? I think we should have tests that include each micro-header and ensure that was is supposed to be available in that header is indeed available.

We would automatically test that each header can be included without issues. If we want to go the extra mile and use the modularized headers in the unit tests. That would ensure that they are working on their own.

> How will we document this?

Currently, I added blanket wording to our documentation of the form:
```
  - Because `<cuda/std/algorithm>` is a huge header with a considerable compile-time cost, we provide each algorithm
    through a minimal subheader named e.g `<cuda/std/algorithm.find>`
```

This is slightly complicated by the fact that we do not list individual features for standard headers.

> I'd like to see an evaluation of how well this naming pattern will extend to other headers

I intentionally choose a character that is not currently used in any of our source files.  Additionally it leans heavily on other popular languages like python that use `package.feature` for imports

In a perfect world we would just use C++20 modules and go on with our live, but that is not available to use currently

jrhemstad:
> I intentionally choose a character that is not currently used in any of our source files. Additionally it leans heavily on other popular languages like python that use package.feature for imports

This isn't quite what I meant. I meant an analysis of how this pattern would extend to other headers like `<cuda/std/type_traits>` or similar. I don't want to adopt a special pattern that only applies to the `<algorithm>` header and doesn't scale to other headers where we'd need to use a different pattern. That inconsistency would be a bad user experience. 

In other words, I want to be sure this isn't just a one-off solution for the algorithm header and want to think about this more holistically across the library. 

Concretely, I'd like to see a list of other headers we may want to split up like this and evaluate how well the pattern extends to other headers.

miscco:
> This isn't quite what I meant. I meant an analysis of how this pattern would extend to other headers like `<cuda/std/type_traits>` or similar. I don't want to adopt a special pattern that only applies to the `<algorithm>` header and doesn't scale to other headers where we'd need to use a different pattern. That inconsistency would be a bad user experience.

That is exactly what I meant. The chosen character does not appear in *any* source file name in the repository outside of the file extension

currently we only have `[a-z_]` as identifiers in the name and `.[h, cuh]` for the file ending.

As long as there is never a feature that we want to publicize named "h" and "cuh" we are fine.

The patter works perfectly fine with `<cuda/std/type_traits.remove_cvref>`  or any other currently used header name

bernhardmgruber:
I agree with @miscco here that `.` is probably a good choice.

> Concretely, I'd like to see a list of other headers we may want to split up like this and evaluate how well the pattern extends to other headers.

That is still valuable to have independently. And it may give us new insight after all. @miscco can we gather a list of headers that would benefit from splitting as well? Like `<type_traits>`, `<functional>`, `<cmath>`, `<iterator>`, `<utility>`, ... Maybe also `<atomic>`, `<chrono>`, `<memory>`, `<randpm`>, `<ranges>`. Would the `.` work everywhere and "look & feel" ok?

miscco:
> Concretely, I'd like to see a list of other headers we may want to split up like this and evaluate how well the pattern extends to other headers.

I believe the candidates are:
* `<cuda/cmath>`
* `<cuda/functional>`
* `<cuda/iterator>`
* `<cuda/memory>`
* `<cuda/numeric>`
* `<cuda/ptx>`
* `<cuda/random>`
* `<cuda/utility>`
* `<cuda/warp>`
* `<cuda/std/algorithm>`
* `<cuda/std/bit>`
* `<cuda/std/cmath>`
* `<cuda/std/concepts>`
* `<cuda/std/functional>`
* `<cuda/std/iterator>` -> We only want to expose iterators and `next` and friends
* `<cuda/std/memory>`
* `<cuda/std/numeric>`
* `<cuda/std/random>`
* `<cuda/std/ranges>`
* `<cuda/std/utility>`
* `<cuda/std/type_traits>`

bernhardmgruber:
> I think we should have tests that include each micro-header and ensure that was is supposed to be available in that header is indeed available.

Can we have some really minimal tests for that?

libcudacxx/test/libcudacxx/libcxx/module_headers/cuda/std/algorithm/sort_heap.pass.cpp
```diff
@@ -0,0 +1,48 @@
...
+  }
+};
+
+__host__ __device__ constexpr bool test()
+{
+  int a[] = {1, 2, 3};
+  cuda::std::push_heap(a, a + 1);
+  cuda::std::push_heap(a, a + 2);
+  cuda::std::push_heap(a, a + 3);
+  cuda::std::sort_heap(a, a + 3);
```
bernhardmgruber, line 30: Q: This file is about testing `sort_heap`, but it also uses `push_heap`. Is this fine, because we are expecting that to be available here?

libcudacxx/test/libcudacxx/libcxx/module_headers/cuda/std/algorithm/iter_swap.pass.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+struct iter_swap_iter
+{
+  int v;
+};
+
+__host__ __device__ constexpr void iter_swap(iter_swap_iter* a, iter_swap_iter* b)
+{
+  cuda::std::swap(a->v, b->v);
+}
```
bernhardmgruber, line 24: Suggestion: I don't think we need this part of the test.

libcudacxx/include/cuda/std/algorithm.ranges.find_if.h
bernhardmgruber, line 1: Suggestion: Should the header rather be called: `algorithm.ranges.find_if`?

libcudacxx/test/libcudacxx/libcxx/module_headers/cuda/std/algorithm/sort_heap.pass.cpp
miscco: hm I fixed that locally 🤔

miscco: yeah please refresh

libcudacxx/test/libcudacxx/libcxx/module_headers/cuda/std/algorithm/search.pass.cpp
```diff
@@ -0,0 +1,54 @@
...
+struct searcher
+{
+  const int *pf, *pl;
+  template <class It>
+  __host__ __device__ constexpr cuda::std::pair<It, It> operator()(It f, It l) const
+  {
+    return cuda::std::__search(
+      f, l, pf, pl, equal_to{}, cuda::std::random_access_iterator_tag{}, cuda::std::random_access_iterator_tag{});
+  }
+};
```
bernhardmgruber, line 33: Suggestion: restructure the test to not use `#include <cuda/std/utility>` and `cuda::std::pair`

miscco:
@jrhemstad I have added tests to verify everything work with just the modularized includes, and it even found a bug 🎉 

Any other questions?

libcudacxx/test/libcudacxx/libcxx/module_headers/cuda/std/algorithm/search.pass.cpp
miscco: That is not possible because the API requires pair

miscco: I believe we do not need the include though

libcudacxx/include/cuda/std/algorithm.ranges.find_if.h
miscco: That is aterribly good question

libcudacxx/test/libcudacxx/libcxx/module_headers/cuda/std/algorithm/iter_swap.pass.cpp
miscco: It does not hurt and it exercises machinery that the algorithm uses

libcudacxx/test/libcudacxx/libcxx/module_headers/cuda/std/algorithm/search.pass.cpp
bernhardmgruber: I am looking at cppreference here: https://en.cppreference.com/w/cpp/algorithm/search.html. And I don't see `pair` being used.

miscco: its the internal API that uses it

libcudacxx/include/cuda/std/algorithm.ranges.find_if.h
miscco: Changed to `algorithm.ranges.meow`

bernhardmgruber (approved):

jrhemstad (approved):

ericniebler:
i _really_ do not like the naming scheme for the files. tooling very much cares about extensions. i don't expect this to be a great experience in an IDE, for example. and if we modularize all the headers listed in [this comment](https://github.com/NVIDIA/cccl/pull/7935#issuecomment-4124590012) from @missco, then we'll have 100s of headers in `cuda/std/`.

> The vast majority preferred a pattern like `<cuda/std/algorithms/find.h>`

then why aren't we using that?

miscco:
> > The vast majority preferred a pattern like `<cuda/std/algorithms/find.h>`
> 
> then why aren't we using that?

The issue is that it is surprising and non-trivial. Note that we cannot do `<cuda/std/algorithm/find.h>` because `<cuda/std/algorithm>` is already a file not a folder. 

So for every header we need to find a similar but not equal name for the original header which is really awkward. It also breaks down with e.g. `<cuda/std/type_traits>` What would be a good name for a subfolder?

I would also like to note that we have a bazillion files in our CI that are named `meow.pass.cpp` or even `meow.compile.pass.cpp`

ericniebler:
> So for every header we need to find a similar but not equal name for the original header which is really awkward. 

so what? we can find the names. the cure you're suggesting is worse than the disease.

> It also breaks down with e.g. `<cuda/std/type_traits>` What would be a good name for a subfolder?

`<cuda/std/traits/is_integral.h>` seems fine to me. or we can use the stable section names from the standard: `<cuda/std/type.traits/is_integral.h>`

bernhardmgruber:
> > It also breaks down with e.g. `<cuda/std/type_traits>` What would be a good name for a subfolder?
> 
> `<cuda/std/traits/is_integral.h>` seems fine to me.

I disagree. I think discoverability of the single function headers is also very relevant for IDE usage, so if the user types `#include <cuda/std/type_traits` they should see in their auto-completion the `<cuda/std/type_traits>` header as well as all the single function headers.

Can't we just take the umbrella header and append a single character for the directory name? Like `<cuda/std/type_traits_/is_integral>`?

brycelelbach (commented):
I think we must do `meow.suffix` not `algorithm.meow`.

- Not having a suffix messes up syntax highlighting which is painful for users.
- It's probably also not great for AI agents. A suffix tells the AI agent what the file is without looking at it.

We should ideally use just one suffix for header names; let's go with whatever is most consistent with our current best practices. The options are `.h`, `.hpp`, `.cuh`; my understanding is that `.cuh` is considered best practice and what we try to use.

bernhardmgruber (approved):

libcudacxx/include/cuda/std/algorithm.adjacent_find.h
```diff
@@ -0,0 +1,26 @@
...
+// Part of libcu++, the C++ Standard Library for your entire system,
+// under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef _CUDA_STD_ALGORITHM_ALGORITHM_ADJACENT_FIND_H
+#define _CUDA_STD_ALGORITHM_ALGORITHM_ADJACENT_FIND_H
```
davebayer, line 12: Nitpick:
```suggestion
#ifndef _CUDA_STD_ALGORITHM_ADJACENT_FIND_H
#define _CUDA_STD_ALGORITHM_ADJACENT_FIND_H
```
applies to all of the files

davebayer (approved):

libcudacxx/include/cuda/std/algorithm.adjacent_find.h
miscco: I believe you did not refresh ^^

fbusato:
what is the plan to document this feature?

docs/libcudacxx/standard_api/algorithms_library.rst
miscco, line 1: @fbusato I documented it here

PR #7944, opened 2026-03-09
Reduce usage of `cub::DispatchReduce`

bernhardmgruber:


cub/benchmarks/bench/transform_reduce/sum.cu
```diff
@@ -35,34 +35,19 @@ struct square_t
   }
 };
 
+// TODO(bgruber): why do we even need a transpose iterator version? Is this just an artifact of a performance comparison
+// done for https://github.com/NVIDIA/cccl/pull/1091? Can we delete this?
 #define USE_TRANSPOSE_ITERATOR 0
```
bernhardmgruber, line 40: @gevtushenko I would love your opinion here!

gevtushenko: The intention was to have the code around for people to enable it and compare Reduce + transform iterators vs TransformReduce that was added at the time. I don't think it's relevant today, feel free to drop!

bernhardmgruber: Thx!

PR #7949, opened 2026-03-09
Use the new tuning API for `detail::radix_sort::dispatch`

### LINKED ISSUE #7947: Use the new tuning API internally for `detail::radix_sort::dispatch`

bernhardmgruber:
- [x] Merge before: #7835
- [x] No SASS changes for `cub.bench.radix_sort.keys.base` on SM`75;80;90;100`
- [x] No SASS changes for `cub.bench.radix_sort.pairs.base` on SM`75;80;90;100`

Fixes: #7947

### DIFF
--- a/cub/benchmarks/bench/radix_sort/keys.cu
+++ b/cub/benchmarks/bench/radix_sort/keys.cu
@@ -48,7 +48,7 @@ void radix_sort_keys(nvbench::state& state, nvbench::type_list<T, OffsetT>)
                         launch
 #if !TUNE_BASE
                         ,
-                        cuda::execution::__tune(policy_selector<key_t, value_t, offset_t>{})
+                        cuda::execution::tune(policy_selector<key_t, value_t, OffsetT>{})
 #endif // !TUNE_BASE
                           ))
         != cudaSuccess)
--- a/cub/benchmarks/bench/radix_sort/keys.cu
+++ b/cub/benchmarks/bench/radix_sort/keys.cu
@@ -14,7 +14,6 @@
 template <typename T, typename OffsetT>
 void radix_sort_keys(nvbench::state& state, nvbench::type_list<T, OffsetT>)
 {
-  using key_t   = T;
   using value_t = cub::NullType;
   if constexpr (!fits_in_default_shared_memory<T, value_t, OffsetT, cub::SortOrder::Ascending>())
   {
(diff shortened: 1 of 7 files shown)


cub/benchmarks/bench/segmented_radix_sort/keys.cu
```diff
@@ -9,18 +9,14 @@
...
+                    nvbench::type_list<T, OffsetT>,
                     const thrust::device_vector<OffsetT>& offsets,
                     bit_entropy entropy)
 {
-  constexpr bool is_overwrite_ok = false;
-
   using offset_t          = OffsetT;
   using begin_offset_it_t = const offset_t*;
   using end_offset_it_t   = const offset_t*;
-  using segment_size_t    = offset_t;
```
bernhardmgruber, line 21: This is a breaking change of the benchmark and I need feedback whether we are ok with that. We previously tested segmented radix sort with 32 and 64-bit offsets AND segment sizes. However, the public API allows 32 and 64-bit offsets but ONLY 32-bit segmented sizes. So the benchmark covered a use case that is not reachable through the public API. This PR now changes to the behavior (32 and 64-bit offset types but ONLY 32-bit segment size) that reflects the public API, because it is now using the public API.

cub/benchmarks/bench/radix_sort/keys.cu
```diff
@@ -14,82 +14,49 @@
...
-      policy_selector<KeyT>{}
+            ,
+            cuda::execution::__tune(policy_selector<key_t, value_t, offset_t>{})
 #endif // !TUNE_BASE
-    );
+          })
+        != cudaSuccess)
+    {
+      throw 42; // TODO(bgruber): what's the appropriate way to abort a benchmark?
+    }
```
bernhardmgruber: What to do here?

cub/benchmarks/bench/radix_sort/keys.cu
```diff
@@ -14,82 +14,47 @@
...
+                        launch
 #if !TUNE_BASE
-        ,
-      cub::detail::identity_decomposer_t{},
-      policy_selector<KeyT>{}
+                        ,
+                        cuda::execution::__tune(policy_selector<key_t, value_t, offset_t>{})
 #endif // !TUNE_BASE
-    );
+                          ))
```
bernhardmgruber: TODO: This deadlocks now during benchmark execution.

cub/cub/device/device_segmented_radix_sort.cuh
```diff
@@ -1997,21 +2015,27 @@ public:
...
+          d_values,
+          static_cast<::cuda::std::int64_t>(num_items),
+          static_cast<::cuda::std::int64_t>(num_segments),
+          d_begin_offsets,
+          d_end_offsets,
+          begin_bit,
+          end_bit,
+          false,
+          stream,
+          {},
```
miscco: I am always a bit scared of just `{}` as function arguments.

Can we add the type?

miscco (approved):

cub/benchmarks/bench/radix_sort/keys.cu
bernhardmgruber: Maybe this was also caused by #8901

cub/benchmarks/bench/radix_sort/keys.cu
bernhardmgruber: Using `_CCCL_TRY_CUDA_API`

cub/benchmarks/bench/segmented_radix_sort/keys.cu
bernhardmgruber: Moved this to a different PR

davebayer (approved):

PR #7965, opened 2026-03-10
Add env DeviceCopy::* and tests

### LINKED ISSUE #7539: cub::DeviceCopy
- [ ] Add env-based cub::DeviceCopy::Batched()

gonidelis:
fixes #7539 

This one is a bit peculiar.

- `DeviceCopy::Batched` uses `dispatch_with_env` like other CUB env overloads — it has its own kernel dispatch via `DispatchBatchMemcpy`.

 - `DeviceCopy::Copy` does not launch its own kernels, but it rather delegates entirely to `DeviceTransform::Transform` and `DeviceFor::ForEachInExtents`, both of which already accept an env. The env is forwarded directly through
  `detail::copy_mdspan::copy`, which is generalized from `cudaStream_t` to a templated `EnvT` parameter.

    - Because Copy delegates to APIs that already emit NVTX ranges, the Copy env overload does not emit its own `_CCCL_NVTX_RANGE_SCOPE` to avoid nested NVTX range violations.

    - mdspan Copy env tests are only in `_env_api.cu`, not `in _env.cu`. The `_env.cu` file defines `stream_registry_factory_t` as the `CUB_DETAIL_DEFAULT_KERNEL_LAUNCHER_FACTORY`, which intercepts kernel launches to verify env usage. Since mdspan Copy does not launch its own kernels, the factory intercepts at the wrong level (inside `Transform/ForEachInExtents`), causing nvcc compilation failures. The _env_api.cu tests (which don't use the factory) provide sufficient coverage for the mdspan Copy env path.

### DIFF
--- a/cub/test/catch2_test_device_copy_env_api.cu
+++ b/cub/test/catch2_test_device_copy_env_api.cu
@@ -33,13 +33,9 @@ C2H_TEST("cub::DeviceCopy::Batched accepts env with stream", "[copy][env]")
   const int* src_base = thrust::raw_pointer_cast(d_src.data());
   int* dst_base       = thrust::raw_pointer_cast(d_dst.data());
 
-  const int* h_input_ptrs[] = {src_base, src_base + range_size, src_base + 2 * range_size};
-  int* h_output_ptrs[]      = {dst_base, dst_base + range_size, dst_base + 2 * range_size};
-  int h_sizes[]             = {range_size, range_size, range_size};
-
-  thrust::device_vector<const int*> d_input_ptrs(h_input_ptrs, h_input_ptrs + num_ranges);
-  thrust::device_vector<int*> d_output_ptrs(h_output_ptrs, h_output_ptrs + num_ranges);
-  thrust::device_vector<int> d_sizes(h_sizes, h_sizes + num_ranges);
+  thrust::device_vector<const int*> d_input_ptrs{src_base, src_base + range_size, src_base + 2 * range_size};
+  thrust::device_vector<int*> d_output_ptrs{dst_base, dst_base + range_size, dst_base + 2 * range_size};
+  thrust::device_vector<int> d_sizes{range_size, range_size, range_size};
 
   cuda::stream stream{cuda::devices[0]};
   cuda::stream_ref stream_ref{stream};


cub/cub/device/device_copy.cuh
```diff
@@ -164,6 +166,85 @@ struct DeviceCopy
...
+  //!   The total number of range pairs
+  //!
+  //! @param[in] env
+  //!   **[optional]** Execution environment. Default is ``cuda::std::execution::env{}``.
+  //!   @endrst
+  template <typename InputIt,
+            typename OutputIt,
+            typename SizeIteratorT,
+            typename EnvT                                                          = ::cuda::std::execution::env<>,
+            ::cuda::std::enable_if_t<!::cuda::std::is_same_v<InputIt, void*>, int> = 0>
```
bernhardmgruber: Suggestion: We do not need a constraint here, since `Batched` here has 4-5 parameters, and the old overload has to 6-7, so they can never be ambiguous.


```suggestion
            typename EnvT = ::cuda::std::execution::env<>>
```

cub/cub/device/device_copy.cuh
```diff
@@ -277,6 +358,110 @@ struct DeviceCopy
...
+    _CCCL_ASSERT(mdspan_in.extents() == mdspan_out.extents(), "mdspan extents must be equal");
+    _CCCL_ASSERT((mdspan_in.data_handle() != nullptr && mdspan_out.data_handle() != nullptr) || mdspan_in.size() == 0,
+                 "mdspan data handle must not be nullptr if the size is not 0");
+    if (mdspan_in.size() != 0)
+    {
+      auto in_start  = mdspan_in.data_handle();
+      auto in_end    = in_start + mdspan_in.mapping().required_span_size();
+      auto out_start = mdspan_out.data_handle();
+      auto out_end   = out_start + mdspan_out.mapping().required_span_size();
+      _CCCL_ASSERT(!(in_end >= out_start && out_end >= in_start), "mdspan memory ranges must not overlap");
```
bernhardmgruber, line 457: Please retain the comment
```suggestion
      // TODO(fbusato): replace with __are_ptrs_overlapping
      _CCCL_ASSERT(!(in_end >= out_start && out_end >= in_start), "mdspan memory ranges must not overlap");
```

bernhardmgruber:
> Because Copy delegates to APIs that already emit NVTX ranges, the Copy env overload does not emit its own _CCCL_NVTX_RANGE_SCOPE to avoid nested NVTX range violations.

I think that's the wrong approach. We should emit an NVTX range at the first API entry into CUB. Later CUB algorithms should not emit NVTX ranges.

cub/cub/device/device_copy.cuh
```diff
@@ -277,6 +358,110 @@ struct DeviceCopy
...
+            typename Extents_Out,
+            typename Layout_Out,
+            typename Accessor_Out,
+            typename EnvT = ::cuda::std::execution::env<>>
+  [[nodiscard]] CUB_RUNTIME_FUNCTION static cudaError_t
+  Copy(::cuda::std::mdspan<T_In, Extents_In, Layout_In, Accessor_In> mdspan_in,
+       ::cuda::std::mdspan<T_Out, Extents_Out, Layout_Out, Accessor_Out> mdspan_out,
+       EnvT env = {})
+  {
+    // no nvtx range because Copy delegates to Transform/ForEachInExtents which emit their own NVTX ranges
```
bernhardmgruber: Critical: Please emit an NVTX range here.

cub/cub/device/dispatch/dispatch_copy_mdspan.cuh
```diff
@@ -66,11 +67,11 @@ copy(::cuda::std::mdspan<T_In, E_In, L_In, A_In> mdspan_in,
       mdspan_out.data_handle(),
       mdspan_in.size(),
       ::cuda::proclaim_copyable_arguments(::cuda::std::identity{}),
-      stream);
+      env);
```
bernhardmgruber: Suggestion: to avoid `cub::DeviceTransform` emitting an NVTX range here, maybe just call `TransformInternal` instead. If you rename `TransformInternal` yo `__trasnform_internal` you can make it `public`.

cub/cub/device/dispatch/dispatch_copy_mdspan.cuh
```diff
@@ -66,11 +67,11 @@ copy(::cuda::std::mdspan<T_In, E_In, L_In, A_In> mdspan_in,
       mdspan_out.data_handle(),
       mdspan_in.size(),
       ::cuda::proclaim_copyable_arguments(::cuda::std::identity{}),
```
bernhardmgruber: Suggestion: You can remove `proclaim_copyable_arguments` as well, since it's not needed.

cub/cub/device/dispatch/dispatch_copy_mdspan.cuh
```diff
@@ -66,11 +67,11 @@ copy(::cuda::std::mdspan<T_In, E_In, L_In, A_In> mdspan_in,
...
-      stream);
+      env);
   }
   // TODO (fbusato): add ForEachInLayout when mdspan_in and mdspan_out have compatible layouts
   // Compatible layouts could use more efficient iteration patterns
-  return cub::DeviceFor::ForEachInExtents(mdspan_in.extents(), copy_mdspan_t{mdspan_in, mdspan_out}, stream);
+  return cub::DeviceFor::ForEachInExtents(mdspan_in.extents(), copy_mdspan_t{mdspan_in, mdspan_out}, env);
```
bernhardmgruber: Suggestion: to avoid this API to emit an NVTX range, we have two options:

1. We add an internal version of `ForEachInExtents` that does not emit the range
2. We add a secret parameter to the `env` that is checked in `ForEachInExtents` and suppresses the NVTX range.

cub/cub/device/device_copy.cuh
gonidelis: ~how? comment explains why i can't~ you explained it right underneath

cub/cub/device/device_for.cuh
```diff
@@ -1318,6 +1318,31 @@ public:
...
+    for_each::op_wrapper_extents_t<OpType, extents_type, is_layout_right, fast_mod_array_t> op_wrapper{
+      op, extents, sub_sizes_div_array, extents_div_array};
+    using ShapeT = implicit_prom_t<extent_index_type>;
+    auto shape   = static_cast<ShapeT>(cub::detail::size(extents));
+    if (shape == 0)
+    {
+      return cudaSuccess;
+    }
+    auto stream = ::cuda::__call_or(::cuda::get_stream, ::cuda::stream_ref{cudaStream_t{}}, env);
+    return detail::for_each::dispatch<ShapeT>(shape, op_wrapper, stream.get());
```
bernhardmgruber: Important: Please avoid code duplication and refactor the existing overloads to share this new implementation detail.

I would make this function equivalent to the content of `ForEachInLayout` and call it `__for_each_in_extents` (no need for `_internal`). Then call it in `ForEachInLayout` and from `DeviceCopy::Copy`.

cub/test/catch2_test_device_copy_env.cu
```diff
@@ -0,0 +1,132 @@
...
+  auto output_it = thrust::make_transform_iterator(
+    iota, index_to_ptr<int>{thrust::raw_pointer_cast(d_dst.data()), thrust::raw_pointer_cast(d_offsets.data())});
+  auto sizes = thrust::make_transform_iterator(iota, get_size{thrust::raw_pointer_cast(d_offsets.data())});
+
+  REQUIRE(cudaSuccess == cub::DeviceCopy::Batched(input_it, output_it, sizes, num_ranges));
+
+  REQUIRE(d_dst == d_src);
+}
+
+#endif
```
miscco: Nitpick: 
```suggestion
#endif //  TEST_LAUNCH == 0
```

miscco (approved):

cub/test/catch2_test_device_copy_env_api.cu
```diff
@@ -0,0 +1,109 @@
...
+  cuda::stream_ref stream_ref{stream};
+  auto env = cuda::std::execution::env{stream_ref};
+
+  auto error = cub::DeviceCopy::Batched(input_it, output_it, sizes, num_ranges, env);
+  if (error != cudaSuccess)
+  {
+    std::cerr << "cub::DeviceCopy::Batched failed with status: " << error << std::endl;
+  }
+
+  thrust::device_vector<int> expected{10, 20, 30, 40, 50, 60};
```
bernhardmgruber: Suggestion: let's provide a more user-friendly and practical example, like we did for `DeviceMemcpy`.

cub/test/catch2_test_device_copy_env_api.cu
```diff
@@ -0,0 +1,97 @@
...
+  const int* src_base = thrust::raw_pointer_cast(d_src.data());
+  int* dst_base       = thrust::raw_pointer_cast(d_dst.data());
+
+  const int* h_input_ptrs[] = {src_base, src_base + range_size, src_base + 2 * range_size};
+  int* h_output_ptrs[]      = {dst_base, dst_base + range_size, dst_base + 2 * range_size};
+  int h_sizes[]             = {range_size, range_size, range_size};
+
+  thrust::device_vector<const int*> d_input_ptrs(h_input_ptrs, h_input_ptrs + num_ranges);
+  thrust::device_vector<int*> d_output_ptrs(h_output_ptrs, h_output_ptrs + num_ranges);
+  thrust::device_vector<int> d_sizes(h_sizes, h_sizes + num_ranges);
```
bernhardmgruber: Suggestion: Let's just init the device vectors directly:
```suggestion
  thrust::device_vector<const int*> d_input_ptrs{src_base, src_base + range_size, src_base + 2 * range_size};
  thrust::device_vector<int*> d_output_ptrs{dst_base, dst_base + range_size, dst_base + 2 * range_size};
  thrust::device_vector<int> d_sizes{range_size, range_size, range_size};
```

PR #7966, opened 2026-03-10
Add env DeviceMemcpy::Batched and tests

### LINKED ISSUE #7540: cub::DeviceMemcpy
- [ ] Add env-based cub::DeviceMemcpy::Batched()

gonidelis:
fixes #7540 

Adds environment overload for the only overload of `DeviceMemcpy`, `DeviceMemcpy::Batched` that is.

No deterministic guarantees are specified.

### DIFF
--- a/cub/test/catch2_test_device_memcpy_env_api.cu
+++ b/cub/test/catch2_test_device_memcpy_env_api.cu
@@ -7,8 +7,6 @@
 
 #include <thrust/detail/raw_pointer_cast.h>
 #include <thrust/device_vector.h>
-#include <thrust/iterator/counting_iterator.h>
-#include <thrust/iterator/transform_iterator.h>
 
 #include <cuda/devices>
 #include <cuda/stream>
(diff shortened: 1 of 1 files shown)


cub/test/catch2_test_device_memcpy_env.cu
```diff
@@ -0,0 +1,131 @@
...
+  }
+};
+
+#if TEST_LAUNCH == 0
+
+TEST_CASE("DeviceMemcpy::Batched works with default environment", "[memcpy][device]")
+{
+  // 3 buffers: [10, 20], [30, 40, 50], [60]
+  auto d_src     = c2h::device_vector<int>{10, 20, 30, 40, 50, 60};
+  auto d_dst     = c2h::device_vector<int>(6, 0);
```
bernhardmgruber: Suggestion:
```suggestion
  auto d_dst     = c2h::device_vector<int>(6);
```

cub/test/catch2_test_device_memcpy_env.cu
```diff
@@ -0,0 +1,131 @@
...
+  auto d_offsets = c2h::device_vector<int>{0, 2, 5, 6};
+
+  int num_buffers = 3;
+
+  thrust::counting_iterator<int> iota(0);
```
bernhardmgruber: Important: Please prefer cuda iterators over thrust iterators.

cub/test/catch2_test_device_memcpy_env.cu
```diff
@@ -0,0 +1,131 @@
...
+  auto input_it = thrust::make_transform_iterator(
+    iota, index_to_ptr<const int>{thrust::raw_pointer_cast(d_src.data()), thrust::raw_pointer_cast(d_offsets.data())});
+  auto output_it = thrust::make_transform_iterator(
+    iota, index_to_ptr<int>{thrust::raw_pointer_cast(d_dst.data()), thrust::raw_pointer_cast(d_offsets.data())});
+  auto sizes = thrust::make_transform_iterator(iota, get_size{thrust::raw_pointer_cast(d_offsets.data())});
+
+  cudaStream_t custom_stream;
+  REQUIRE(cudaSuccess == cudaStreamCreate(&custom_stream));
```
bernhardmgruber: Important: Please use `cuda::stream` to increase its test coverage

cub/test/catch2_test_device_memcpy_env.cu
```diff
@@ -0,0 +1,131 @@
...
+struct get_size
+{
+  const int* offsets;
+  __host__ __device__ __forceinline__ int operator()(int index) const
+  {
+    return (offsets[index + 1] - offsets[index]) * static_cast<int>(sizeof(int));
+  }
+};
+
+#if TEST_LAUNCH == 0
```
bernhardmgruber, line 49: Q: What prevents us from running the below unit test for launch id 1 and 2?

cub/test/catch2_test_device_memcpy_env_api.cu
```diff
@@ -0,0 +1,72 @@
...
+  auto env = cuda::std::execution::env{stream_ref};
+
+  auto error = cub::DeviceMemcpy::Batched(input_it, output_it, sizes, num_buffers, env);
+  if (error != cudaSuccess)
+  {
+    std::cerr << "cub::DeviceMemcpy::Batched failed with status: " << error << std::endl;
+  }
+
+  thrust::device_vector<int> expected{10, 20, 30, 40, 50, 60};
+  // example-end memcpy-batched-env
```
bernhardmgruber: Suggestion: I don't think this is a great API example. We should focus on the API's design of taking an iterator to pointers. Something like:
```c++
  auto d_src     = thrust::device_vector<int>{10, 20, 30, 40, 50, 60};
  auto d_src_pointers = thrust::device_vector<int*>{d_src[0], d_src[2], d_src[5]};
```
+ the raw pointer casts. You could even consider writing the results to two different buffers, to highlight the API's flexibility.

gonidelis: done

cub/test/catch2_test_device_memcpy_env.cu
gonidelis: providing explanation for public:

the launch wrapper for env tests expects last argument to be `env` but for `test_launch == 0` we are checking the cases were **env is not passed at all** so the wrapper always fails to compile due to last argument being anything other than `env`.

bernhardmgruber: Thx! We discussed this offline as well.

PR #7969, opened 2026-03-10
Add environment `DeviceMerge` 

### LINKED ISSUE #7543: Add env-based API for cub::DeviceMerge
- [x] Add env-based cub::DeviceMerge::MergeKeys()
- [ ] Add env-based cub::DeviceMerge::MergePairs()

gonidelis:
fixes #7543

### DIFF
--- a/cub/cub/device/device_merge.cuh
+++ b/cub/cub/device/device_merge.cuh
@@ -172,15 +172,16 @@ struct DeviceMerge
   //!   **[optional]** Execution environment. Default is ``cuda::std::execution::env{}``.
   //!   @endrst
   //! [strict weak ordering]: https://en.cppreference.com/w/cpp/concepts/strict_weak_order
-  template <typename KeyIteratorIn1,
-            typename KeyIteratorIn2,
-            typename KeyIteratorOut,
-            typename CompareOp            = ::cuda::std::less<>,
-            typename EnvT                 = ::cuda::std::execution::env<>,
-            ::cuda::std::enable_if_t<!::cuda::std::is_same_v<KeyIteratorIn1, void*>
-                                       && !::cuda::std::is_same_v<KeyIteratorIn1, ::cuda::std::nullptr_t>
-                                       && !::cuda::std::is_arithmetic_v<CompareOp>,
-                                     int> = 0>
+  template <
+    typename KeyIteratorIn1,
+    typename KeyIteratorIn2,
+    typename KeyIteratorOut,
+    typename CompareOp = ::cuda::std::less<>,
+    typename EnvT      = ::cuda::std::execution::env<>,
+    ::cuda::std::enable_if_t<
+      !::cuda::std::is_same_v<KeyIteratorIn1, void*> && !::cuda::std::is_same_v<KeyIteratorIn1, ::cuda::std::nullptr_t>
+        && ::cuda::std::indirect_binary_predicate<CompareOp, KeyIteratorIn1, KeyIteratorIn2>,
+      int> = 0>
   [[nodiscard]] CUB_RUNTIME_FUNCTION static cudaError_t MergeKeys(
     KeyIteratorIn1 keys_in1,
     ::cuda::std::int64_t num_keys1,
@@ -387,18 +388,19 @@ struct DeviceMerge
   //!   **[optional]** Execution environment. Default is ``cuda::std::execution::env{}``.
   //!   @endrst
   //! [strict weak ordering]: https://en.cppreference.com/w/cpp/concepts/strict_weak_order
-  template <typename KeyIteratorIn1,
-            typename ValueIteratorIn1,
-            typename KeyIteratorIn2,
-            typename ValueIteratorIn2,
-            typename KeyIteratorOut,
-            typename ValueIteratorOut,
-            typename CompareOp            = ::cuda::std::less<>,
-            typename EnvT                 = ::cuda::std::execution::env<>,
-            ::cuda::std::enable_if_t<!::cuda::std::is_same_v<KeyIteratorIn1, void*>
-                                       && !::cuda::std::is_same_v<KeyIteratorIn1, ::cuda::std::nullptr_t>
-                                       && !::cuda::std::is_arithmetic_v<CompareOp>,
-                                     int> = 0>
+  template <
+    typename KeyIteratorIn1,
+    typename ValueIteratorIn1,
+    typename KeyIteratorIn2,
+    typename ValueIteratorIn2,
+    typename KeyIteratorOut,
+    typename ValueIteratorOut,
+    typename CompareOp = ::cuda::std::less<>,
+    typename EnvT      = ::cuda::std::execution::env<>,
+    ::cuda::std::enable_if_t<
+      !::cuda::std::is_same_v<KeyIteratorIn1, void*> && !::cuda::std::is_same_v<KeyIteratorIn1, ::cuda::std::nullptr_t>
+        && ::cuda::std::indirect_binary_predicate<CompareOp, KeyIteratorIn1, KeyIteratorIn2>,
+      int> = 0>
   [[nodiscard]] CUB_RUNTIME_FUNCTION static cudaError_t MergePairs(
     KeyIteratorIn1 keys_in1,
     ValueIteratorIn1 values_in1,


cub/cub/device/device_merge.cuh
```diff
@@ -103,6 +106,100 @@ struct DeviceMerge
...
+  //!
+  //! [strict weak ordering]: https://en.cppreference.com/w/cpp/concepts/strict_weak_order
+  template <typename KeyIteratorIn1,
+            typename KeyIteratorIn2,
+            typename KeyIteratorOut,
+            typename CompareOp            = ::cuda::std::less<>,
+            typename EnvT                 = ::cuda::std::execution::env<>,
+            ::cuda::std::enable_if_t<!::cuda::std::is_same_v<KeyIteratorIn1, void*>
+                                       && !::cuda::std::is_same_v<KeyIteratorIn1, ::cuda::std::nullptr_t>,
+                                     int> = 0>
```
bernhardmgruber: Remark: that's an interesting constraint since it guards against using the old overload with `nullptr`. It's fine. You need it here I guess because you cannot constraint a second template argument since it's just `int64` here and a `size_t` from the other overload would convert to `int64` well.

cub/cub/device/device_merge.cuh
```diff
@@ -103,6 +106,100 @@ struct DeviceMerge
...
+    _CCCL_NVTX_RANGE_SCOPE("cub::DeviceMerge::MergeKeys");
+
+    using requirements_t = ::cuda::std::execution::
+      __query_result_or_t<EnvT, ::cuda::execution::__get_requirements_t, ::cuda::std::execution::env<>>;
+    using requested_determinism_t =
+      ::cuda::std::execution::__query_result_or_t<requirements_t,
+                                                  ::cuda::execution::determinism::__get_determinism_t,
+                                                  ::cuda::execution::determinism::run_to_run_t>;
+    static_assert(!::cuda::std::is_same_v<requested_determinism_t, ::cuda::execution::determinism::gpu_to_gpu_t>,
+                  "gpu_to_gpu determinism is not supported for unstable device merge");
```
bernhardmgruber: Important: I need to think about this, since an unstable algorithm could still be deterministic (even across multiple GPUs). How did you conclude that the merge path implementation is run_to_run deterministic?

cub/cub/device/device_merge.cuh
gonidelis: we still get an ambiguity:

```cpp
cub::DeviceMerge::MergeKeys(
    thrust::raw_pointer_cast(temp_storage.data()),  // KeyIteratorIn1 = char*
    temp_storage_bytes,                              // num_keys1 (size_t -> int64)
    keys1.begin(),                                   // KeyIteratorIn2
    static_cast<int>(keys1.size()),                  // num_keys2 (int -> int64)
    keys2.begin(),                                   // KeyIteratorOut
    static_cast<int>(keys2.size()),                  // CompareOp = int
    result.begin());                                 // EnvT = iterator
```

gonidelis: we need extra constraints

cub/cub/device/device_merge.cuh
gonidelis: stripped any refs about determinism

gonidelis:
After long discussion with @bernhardmgruber  we decided not to specify any deterministic guarantees for `DeviceMerge`

cub/test/catch2_test_device_merge_env.cu
```diff
@@ -0,0 +1,242 @@
...
+#include <thrust/device_vector.h>
+
+#include "catch2_test_env_launch_helper.h"
+
+DECLARE_LAUNCH_WRAPPER(cub::DeviceMerge::MergeKeys, merge_keys);
+DECLARE_LAUNCH_WRAPPER(cub::DeviceMerge::MergePairs, merge_pairs);
+
+// %PARAM% TEST_LAUNCH lid 0:1:2
+
+#include <cuda/__execution/require.h>
```
gonidelis, line 21: might not be needed

cub/cub/device/device_merge.cuh
```diff
@@ -103,6 +105,111 @@ struct DeviceMerge
...
+  //!   @endrst
+  //! [strict weak ordering]: https://en.cppreference.com/w/cpp/concepts/strict_weak_order
+  template <typename KeyIteratorIn1,
+            typename KeyIteratorIn2,
+            typename KeyIteratorOut,
+            typename CompareOp            = ::cuda::std::less<>,
+            typename EnvT                 = ::cuda::std::execution::env<>,
+            ::cuda::std::enable_if_t<!::cuda::std::is_same_v<KeyIteratorIn1, void*>
+                                       && !::cuda::std::is_same_v<KeyIteratorIn1, ::cuda::std::nullptr_t>
+                                       && !::cuda::std::is_arithmetic_v<CompareOp>,
```
miscco: This should rather be
```suggestion
                                       && ::cuda::std::indirect_binary_predicate<CompareOp, KeyIteratorIn1, KeyIteratorIn2>,
```

cub/cub/device/device_merge.cuh
```diff
@@ -103,6 +105,112 @@ struct DeviceMerge
...
+  template <
+    typename KeyIteratorIn1,
+    typename KeyIteratorIn2,
+    typename KeyIteratorOut,
+    typename CompareOp = ::cuda::std::less<>,
+    typename EnvT      = ::cuda::std::execution::env<>,
+    ::cuda::std::enable_if_t<
+      !::cuda::std::is_same_v<KeyIteratorIn1, void*> && !::cuda::std::is_same_v<KeyIteratorIn1, ::cuda::std::nullptr_t>
+        && ::cuda::std::indirect_binary_predicate<CompareOp, KeyIteratorIn1, KeyIteratorIn2>,
+      int> = 0>
```
miscco, line 184: Nitpick, I believe it would be easier to read if we put the different conditions into a separate enable_if
```suggestion
    ::cuda::std::enable_if_t<
      !::cuda::std::is_same_v<KeyIteratorIn1, void*> && !::cuda::std::is_same_v<KeyIteratorIn1, ::cuda::std::nullptr_t>, int> = 0,
   ::cuda::std::enable_if_t<::cuda::std::indirect_binary_predicate<CompareOp, KeyIteratorIn1, KeyIteratorIn2>,
      int> = 0>
```

More importantly, We should not need the void* check, because that should be part of the indirect binary predicate
```suggestion
    ::cuda::std::enable_if_t<::cuda::std::indirect_binary_predicate<CompareOp, KeyIteratorIn1, KeyIteratorIn2>,
      int> = 0>
```

Can you check whether that suffices?

cub/cub/device/device_merge.cuh
```diff
@@ -191,6 +299,139 @@ struct DeviceMerge
...
+    typename KeyIteratorIn2,
+    typename ValueIteratorIn2,
+    typename KeyIteratorOut,
+    typename ValueIteratorOut,
+    typename CompareOp = ::cuda::std::less<>,
+    typename EnvT      = ::cuda::std::execution::env<>,
+    ::cuda::std::enable_if_t<
+      !::cuda::std::is_same_v<KeyIteratorIn1, void*> && !::cuda::std::is_same_v<KeyIteratorIn1, ::cuda::std::nullptr_t>
+        && ::cuda::std::indirect_binary_predicate<CompareOp, KeyIteratorIn1, KeyIteratorIn2>,
+      int> = 0>
```
miscco, line 403: Ditto: try whether the indirect_binary_predicate is already sufficient

cub/cub/device/device_merge.cuh
bernhardmgruber: Here is a fun fact: our merge implementation is a bit sloppy and we convert all keys to the value type of the first iterator when loading from GMEM. See `merge::agent_t::consume_tile()`. However, I also don't want to put `indirect_binary_predicate<CompareOp, KeyIteratorIn1, KeyIteratorIn1>` into the public API since it's technically a bug and we may want to fix it without an API break. Can we not touch the comparison predicate with any constraint for now?

bernhardmgruber: More generally: I discussed with @gonidelis to review all constraints on env-overloads once they are done and properly refine them. Maybe you can help with that!

miscco: Fine by me

cub/cub/device/device_merge.cuh
```diff
@@ -13,9 +13,11 @@
 #  pragma system_header
 #endif // no system header
 
+#include <cub/detail/env_dispatch.cuh>
 #include <cub/device/dispatch/dispatch_merge.cuh>
 #include <cub/util_namespace.cuh>
 
+#include <cuda/__execution/require.h>
```
bernhardmgruber, line 20: I think this is unused

```suggestion
```

cub/cub/device/device_merge.cuh
bernhardmgruber: We can probably also leave the code as as, since the value type of `KeyIteratorIn1` must be convertible to the value type of `KeyIteratorIn2`, so if `indirect_binary_predicate<CompareOp, KeyIteratorIn1, KeyIteratorIn2>` is true, so is `indirect_binary_predicate<CompareOp, KeyIteratorIn1, KeyIteratorIn1>`.

PR #7970, opened 2026-03-10
Add DeviceScan::*ByKey env algorithms

### LINKED ISSUE #7548: cub::DeviceScan
- [x] [Add env-based cub::DeviceScan::ExclusiveSum()](https://github.com/NVIDIA/cccl/issues/5657)
- [x] [Add env-based cub::DeviceScan::ExclusiveScan()](https://github.com/NVIDIA/cccl/issues/5657)
- [ ] Add env-based cub::DeviceScan::InclusiveSum()
- [ ] Add env-based cub::DeviceScan::InclusiveScan()
- [ ] Add env-based cub::DeviceScan::InclusiveScanInit()
- [ ] Add env-based cub::DeviceScan::ExclusiveSumByKey()
- [ ] Add env-based cub::DeviceScan::ExclusiveScanByKey()
- [ ] Add env-based cub::DeviceScan::InclusiveSumByKey()
- [ ] Add env-based cub::DeviceScan::InclusiveScanByKey()

gonidelis:
partially fixes #7548 (one overload left)

### DIFF
--- a/cub/test/catch2_test_device_scan_by_key_env_api.cu
+++ b/cub/test/catch2_test_device_scan_by_key_env_api.cu
@@ -14,25 +14,6 @@
 
 #include <c2h/catch2_test_helper.h>
 
-C2H_TEST("cub::DeviceScan::ExclusiveSumByKey accepts environment", "[scan][by_key][env]")
-{
-  auto keys   = thrust::device_vector<int>{0, 0, 1, 1, 1, 2, 2};
-  auto input  = thrust::device_vector<int>{8, 6, 7, 5, 3, 0, 9};
-  auto output = thrust::device_vector<int>(7);
-
-  auto error =
-    cub::DeviceScan::ExclusiveSumByKey(keys.begin(), input.begin(), output.begin(), static_cast<int>(input.size()));
-  if (error != cudaSuccess)
-  {
-    std::cerr << "cub::DeviceScan::ExclusiveSumByKey failed with status: " << error << std::endl;
-  }
-
-  thrust::device_vector<int> expected{0, 8, 0, 7, 12, 0, 0};
-
-  REQUIRE(error == cudaSuccess);
-  REQUIRE(output == expected);
-}
-
 C2H_TEST("cub::DeviceScan::ExclusiveSumByKey accepts stream environment", "[scan][by_key][env]")
 {
   // example-begin exclusive-sum-by-key-env
@@ -53,27 +34,6 @@ C2H_TEST("cub::DeviceScan::ExclusiveSumByKey accepts stream environment", "[scan
 
   thrust::device_vector<float> expected{0.0f, 8.0f, 0.0f, 7.0f, 12.0f, 0.0f, 0.0f};
   // example-end exclusive-sum-by-key-env
-
-  REQUIRE(error == cudaSuccess);
-  REQUIRE(output == expected);
-}
-
-C2H_TEST("cub::DeviceScan::ExclusiveScanByKey accepts environment", "[scan][by_key][env]")
-{
-  auto op     = cuda::std::plus{};
-  auto keys   = thrust::device_vector<int>{0, 0, 1, 1, 1, 2, 2};
-  auto input  = thrust::device_vector<int>{8, 6, 7, 5, 3, 0, 9};
-  auto output = thrust::device_vector<int>(7);
-  auto init   = 0;
-
-  auto error = cub::DeviceScan::ExclusiveScanByKey(
-    keys.begin(), input.begin(), output.begin(), op, init, static_cast<int>(input.size()));
-  if (error != cudaSuccess)
-  {
-    std::cerr << "cub::DeviceScan::ExclusiveScanByKey failed with status: " << error << std::endl;
-  }
-
-  thrust::device_vector<int> expected{0, 8, 0, 7, 12, 0, 0};
 
   REQUIRE(error == cudaSuccess);
   REQUIRE(output == expected);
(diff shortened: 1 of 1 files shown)


cub/cub/device/device_scan.cuh
```diff
@@ -198,6 +201,46 @@ struct DeviceScan
...
   }
+  template <typename KeysInputIteratorT,
+            typename ValuesInputIteratorT,
+            typename ValuesOutputIteratorT,
+            typename EqualityOpT,
+            typename ScanOpT,
+            typename InitValueT,
+            typename NumItemsT,
+            typename EnvT>
+  CUB_RUNTIME_FUNCTION _CCCL_FORCEINLINE static cudaError_t scan_bykey_impl_env(
```
bernhardmgruber:
```suggestion
  CUB_RUNTIME_FUNCTION _CCCL_FORCEINLINE static cudaError_t scan_by_key_impl_env(
```

cub/cub/device/device_scan.cuh
```diff
@@ -198,6 +201,46 @@ struct DeviceScan
...
+    KeysInputIteratorT d_keys_in,
+    ValuesInputIteratorT d_values_in,
+    ValuesOutputIteratorT d_values_out,
+    EqualityOpT equality_op,
+    ScanOpT scan_op,
+    InitValueT init_value,
+    NumItemsT num_items,
+    EnvT env)
+  {
+    // Dispatch with environment - handles all boilerplate
```
bernhardmgruber: Suggestion: remove the comment. It does not add any value.

cub/cub/device/device_scan.cuh
```diff
@@ -2368,6 +2411,406 @@ struct DeviceScan
...
+  //!   @endrst
+  template <typename KeysInputIteratorT,
+            typename ValuesInputIteratorT,
+            typename ValuesOutputIteratorT,
+            typename EqualityOpT = ::cuda::std::equal_to<>,
+            typename NumItemsT   = uint32_t,
+            typename EnvT        = ::cuda::std::execution::env<>,
+            ::cuda::std::enable_if_t<
+              !::cuda::std::is_same_v<KeysInputIteratorT, void*> && !::cuda::std::is_null_pointer_v<KeysInputIteratorT>,
+              int> = 0>
```
bernhardmgruber, line 2453: Remark: first time I see `is_null_pointer_v`. I think we need to do an eventual pass over all env overloads and make all constraints uniform (follow the same logic, etc.).

cub/cub/device/device_scan.cuh
```diff
@@ -2368,6 +2411,406 @@ struct DeviceScan
...
+    KeysInputIteratorT d_keys_in,
+    ValuesInputIteratorT d_values_in,
+    ValuesOutputIteratorT d_values_out,
+    NumItemsT num_items,
+    EqualityOpT equality_op = EqualityOpT(),
+    EnvT env                = {})
+  {
+    _CCCL_NVTX_RANGE_SCOPE("cub::DeviceScan::ExclusiveSumByKey");
+
+    using InitT = cub::detail::it_value_t<ValuesInputIteratorT>;
```
bernhardmgruber: Important: Please follow the coding guidelines wrt. to the identifier naming:
```suggestion
    using init_t = cub::detail::it_value_t<ValuesInputIteratorT>;
```

cub/cub/device/device_scan.cuh
gonidelis: I get it it's weird but there is an ambiguity if we don't constrain it:

```cpp
 cub::DeviceScan::ExclusiveScanByKey(
        nullptr,                    // KeysInputIteratorT   = nullptr_t
        temp_bytes,                 // ValuesInputIteratorT = size_t
        d_keys,                     // ValuesOutputIteratorT = int*
        d_vals,                     // ScanOpT              = int*
        d_out,                      // InitValueT           = int*
        ::cuda::std::plus<int>{},   // NumItemsT            = plus<int>
        0,                          // EqualityOpT          = int
        100);                       // EnvT                 = int

```

for the sake of uniformity do you wanna use ` !is_same_v<KeysInputIteratorT, nullptr_t>` instead?

bernhardmgruber: Fine. Can we also constrain `ValuesInputIteratorT` to not be `size_t` like for other overloads in the past?

cub/cub/device/device_scan.cuh
```diff
@@ -2368,6 +2370,481 @@ struct DeviceScan
...
+    init_t init_value{};
+
+    return detail::dispatch_with_env(env, [&]([[maybe_unused]] auto tuning, void* storage, size_t& bytes, auto stream) {
+      using offset_t = detail::choose_offset_t<NumItemsT>;
+      return DispatchScanByKey<
+        KeysInputIteratorT,
+        ValuesInputIteratorT,
+        ValuesOutputIteratorT,
+        EqualityOpT,
+        ::cuda::std::plus<init_t>,
```
bernhardmgruber: Critical: The old overload used `::cuda::std::plus<>`, why do we need to change it? This may be a bug.

cub/cub/device/device_scan.cuh
```diff
@@ -2368,6 +2370,481 @@ struct DeviceScan
...
+
+    return detail::dispatch_with_env(env, [&]([[maybe_unused]] auto tuning, void* storage, size_t& bytes, auto stream) {
+      using offset_t = detail::choose_offset_t<NumItemsT>;
+      using value_t = cub::detail::it_value_t<ValuesInputIteratorT>;
+      return DispatchScanByKey<
+        KeysInputIteratorT,
+        ValuesInputIteratorT,
+        ValuesOutputIteratorT,
+        EqualityOpT,
+        ::cuda::std::plus<value_t>,
```
bernhardmgruber: Same here.

cub/cub/device/device_scan.cuh
gonidelis: @bernhardmgruber  @miscco  maybe i misunderstood, aren't the same issues for integer promotion of short to int applying? for the sake of performance that is

cub/cub/device/device_scan.cuh
gonidelis: done please resolve

cub/cub/device/device_scan.cuh
miscco: I believe the point is that we should use the pre-existing plus<> here and then properly migrate each algorithm as a whole

The env overloads should not change behavior of the algorithm

bernhardmgruber: Yes, if we touch the argument to `plus<>` here, we need to rebenchmark the algorithm and that is a lot more work than we need now. Let's get the env-overloads in without changing behavior.

cub/test/catch2_test_device_scan_by_key_env_api.cu
```diff
@@ -0,0 +1,197 @@
...
+  if (error != cudaSuccess)
+  {
+    std::cerr << "cub::DeviceScan::ExclusiveSumByKey failed with status: " << error << std::endl;
+  }
+
+  thrust::device_vector<int> expected{0, 8, 0, 7, 12, 0, 0};
+
+  REQUIRE(error == cudaSuccess);
+  REQUIRE(output == expected);
+}
```
bernhardmgruber: I am confused by this test and a few others. They don't have an `example-begin` etc. so they don't show up in any API docs? Plus, they seem to test the default environment, which is already covered in the other unit test file. Can we remove these tests? What are they for?

gonidelis: oversight fry mistake due to excessive test writing sorry - removed

PR #7973, opened 2026-03-10
fix NVbug 5967534, function name mismatch typo

chinasoul:
fix NVbug 5967534: function is named dispatch_streaming, not dispatch_streaming_reduce_by_key


miscco:
Thanks a lot for fixing this, I am a bit baffled how this came through, this suggest a major testing gap

bernhardmgruber:
I think this was the result of concurrently merging two PRs recently. Thx for the fix!

PR #7975, opened 2026-03-10
Drop more umbrella headers in cub

miscco:
This is dropping more of the umbrella headers in CUB

The remaining larger header is `<cuda/atomic>` which is really big. 

We should consider providing a simpler type for integrals only, which should improve most cases

### DIFF
--- a/cub/benchmarks/bench/for_each/extents.cu
+++ b/cub/benchmarks/bench/for_each/extents.cu
@@ -3,6 +3,7 @@
 
 #include <cub/device/device_for.cuh>
 
+#include <cuda/cmath>
 #include <cuda/std/mdspan>
 
 #include <nvbench_helper.cuh>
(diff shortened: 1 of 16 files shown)


bernhardmgruber (approved):

bernhardmgruber:
> The remaining larger header is `<cuda/atomic>` which is really big.

We have a `strong_store.cuh` etc. in CUB that is a lot more lightweight. We also have `__nv_atomic_load` etc. in nvcc 12.8+. We could consider adding a lightweight header for some of the common atomic load/store/add functions that we use in CUB so we can avoid including `<cuda/atomic>`.

bernhardmgruber (approved):

miscco:
> > The remaining larger header is `<cuda/atomic>` which is really big.
> 
> We have a `strong_store.cuh` etc. in CUB that is a lot more lightweight. We also have `__nv_atomic_load` etc. in nvcc 12.8+. We could consider adding a lightweight header for some of the common atomic load/store/add functions that we use in CUB so we can avoid including `<cuda/atomic>`.

Definitely, but this is something we should do in a followup PR, there will be enough transitive include issues popping up in this PR i fear

bernhardmgruber (approved):

davebayer (approved):

PR #7897, opened 2026-03-05
[STF] Avoid a deferred implementation of interpreted_execution_policy in places.cuh

caugonnet:
This avoids putting the implementation of an interpreted_execution_policy ctor in places.cuh which is totally unrelated

## Description

closes 

## Checklist

- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #7976, opened 2026-03-10
[STF] Refactor data_place to use polymorphic design

andralex:
## Summary

Replaces the ad-hoc tagged union design in `data_place` with a clean polymorphic architecture.

**Before:** `data_place` used an `int devid` plus separate `shared_ptr<composite_state>` and `shared_ptr<data_place_extension>` members — an awkward mix of enum-based type tags with extension pointers.

**After:**
- `data_place_interface` (new `data_place_interface.cuh`) — abstract virtual interface for all place types, with sensible defaults for `hash()`, `equals()`, `less_than()`, and `mem_create()`
- `data_place_impl.cuh` — concrete implementations: `data_place_host`, `data_place_managed`, `data_place_device`, `data_place_affine`, `data_place_device_auto`, `data_place_invalid`
- `data_place_composite` — defined in `places.cuh` after `exec_place_grid` (needed due to its dependency on that type)
- `data_place` — now holds a `shared_ptr<data_place_interface>` and delegates all operations to it
- Static singleton instances with no-op deleters for common types (host, managed, invalid) to avoid per-use allocations
- Removes `data_place_extension.cuh` — custom extensions now inherit directly from `data_place_interface`

## Test plan
- [ ] Build `cudax.test.stf` and verify no regressions

Made with [Cursor](https://cursor.com)

### DIFF
--- a/cudax/include/cuda/experimental/__stf/internal/logical_data.cuh
+++ b/cudax/include/cuda/experimental/__stf/internal/logical_data.cuh
@@ -1731,7 +1731,7 @@ inline void reserved::logical_data_untyped_impl::erase()
 
       data_instance& ref_instance  = get_data_instance(ref_id);
       const data_place& ref_dplace = ref_instance.get_dplace();
-      _CCCL_ASSERT(ref_dplace.is_concrete(), "ref_dplace must be resolved to a concrete place before erase");
+      _CCCL_ASSERT(ref_dplace.is_resolved(), "ref_dplace must be resolved before erase");
       auto e = ref_dplace.affine_exec_place();
       l.reconstruct_after_redux(ctx, ref_id, e, wb_prereqs);
 
--- a/cudax/include/cuda/experimental/__stf/places/data_place_impl.cuh
+++ b/cudax/include/cuda/experimental/__stf/places/data_place_impl.cuh
@@ -40,6 +40,11 @@ namespace cuda::experimental::stf
 class data_place_invalid final : public data_place_interface
 {
 public:
+  bool is_resolved() const override
+  {
+    return false;
+  }
+
   int get_device_ordinal() const override
   {
     return data_place_interface::invalid;
@@ -86,6 +91,11 @@ public:
 class data_place_host final : public data_place_interface
 {
 public:
+  bool is_resolved() const override
+  {
+    return true;
+  }
+
   int get_device_ordinal() const override
   {
     return data_place_interface::host;
@@ -149,6 +159,11 @@ public:
 class data_place_managed final : public data_place_interface
 {
 public:
+  bool is_resolved() const override
+  {
+    return true;
+  }
+
   int get_device_ordinal() const override
   {
     return data_place_interface::managed;
@@ -197,6 +212,11 @@ public:
 class data_place_device final : public data_place_interface
 {
 public:
+  bool is_resolved() const override
+  {
+    return true;
+  }
+
   explicit data_place_device(int device_id)
       : device_id_(device_id)
   {
@@ -294,6 +314,11 @@ private:
 class data_place_affine final : public data_place_interface
 {
 public:
+  bool is_resolved() const override
+  {
+    return false;
+  }
+
   int get_device_ordinal() const override
   {
     return data_place_interface::affine;
@@ -340,6 +365,11 @@ public:
 class data_place_device_auto final : public data_place_interface
 {
 public:
+  bool is_resolved() const override
+  {
+    return false;
+  }
+
   int get_device_ordinal() const override
   {
     return data_place_interface::device_auto;
--- a/cudax/include/cuda/experimental/__stf/places/data_place_interface.cuh
+++ b/cudax/include/cuda/experimental/__stf/places/data_place_interface.cuh
@@ -84,6 +84,16 @@ public:
   // === Core properties ===
 
   /**
+   * @brief Whether this place is fully resolved and ready for allocation
+   *
+   * Returns true for places that represent a concrete memory target:
+   * host, managed, device(N), composite, green_ctx, etc.
+   * Returns false for abstract/deferred places that need further
+   * resolution: invalid, affine, device_auto.
+   */
+  virtual bool is_resolved() const = 0;
+
+  /**
    * @brief Get the device ordinal for this place
    *
    * Returns:
--- a/cudax/include/cuda/experimental/__stf/places/exec/green_context.cuh
+++ b/cudax/include/cuda/experimental/__stf/places/exec/green_context.cuh
@@ -53,6 +53,11 @@ public:
       : view_(mv(view))
   {}
 
+  bool is_resolved() const override
+  {
+    return true;
+  }
+
   int get_device_ordinal() const override
   {
     return view_.devid;
@@ -451,10 +456,10 @@ UNITTEST("green context data_place equality")
   EXPECT(dp0a != dev0);
   EXPECT(!(dp0a == dev0));
 
-  // Green context data place should be concrete but not a plain device
-  EXPECT(dp0a.is_concrete());
+  // Green context data place should be resolved but not a plain device
+  EXPECT(dp0a.is_resolved());
   EXPECT(!dp0a.is_device());
-  EXPECT(dev0.is_concrete());
+  EXPECT(dev0.is_resolved());
   EXPECT(dev0.is_device());
 };
 
@@ -481,8 +486,8 @@ UNITTEST("green context exec_place equality with green_ctx_data_place flag")
   // Different green contexts should NOT be equal
   EXPECT(p0a != p1);
 
-  // Affine data place should be concrete but not a plain device when use_green_ctx_data_place=true
-  EXPECT(p0a.affine_data_place().is_concrete());
+  // Affine data place should be resolved but not a plain device when use_green_ctx_data_place=true
+  EXPECT(p0a.affine_data_place().is_resolved());
   EXPECT(!p0a.affine_data_place().is_device());
 };
 
--- a/cudax/include/cuda/experimental/__stf/places/places.cuh
+++ b/cudax/include/cuda/experimental/__stf/places/places.cuh
@@ -252,9 +252,9 @@ public:
     return typeid(ref) == typeid(data_place_device_auto);
   }
 
-  bool is_concrete() const
+  bool is_resolved() const
   {
-    return !is_invalid() && !is_affine() && !is_device_auto() && !is_composite();
+    return pimpl_->is_resolved();
   }
 
   ::std::string to_string() const
@@ -1585,6 +1585,11 @@ public:
       : grid_(mv(grid))
       , partitioner_func_(mv(partitioner_func))
   {}
+
+  bool is_resolved() const override
+  {
+    return true;
+  }
 
   int get_device_ordinal() const override
   {
--- a/cudax/include/cuda/experimental/__stf/internal/task.cuh
+++ b/cudax/include/cuda/experimental/__stf/internal/task.cuh
@@ -441,8 +441,7 @@ void dep_allocate(
 {
   auto& inst = d.get_data_instance(instance_id);
 
-  _CCCL_ASSERT(dplace.is_concrete() || dplace.is_composite(),
-               "dep_allocate requires a concrete or composite data_place (resolved upstream in acquire)");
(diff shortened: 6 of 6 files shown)


andralex:
pre-commit.ci autofix

andralex:
pre-commit.ci autofix

cudax/include/cuda/experimental/__stf/internal/logical_data.cuh
```diff
@@ -1732,7 +1731,9 @@ inline void reserved::logical_data_untyped_impl::erase()
 
       data_instance& ref_instance  = get_data_instance(ref_id);
       const data_place& ref_dplace = ref_instance.get_dplace();
-      auto e                       = ref_dplace.affine_exec_place();
+      _CCCL_ASSERT(!ref_dplace.is_affine() && !ref_dplace.is_invalid(),
```
miscco: Nitpick: Should there be a `is_concrete()` method?

cudax/include/cuda/experimental/__stf/places/data_place_impl.cuh
```diff
@@ -0,0 +1,412 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of CUDASTF in CUDA C++ Core Libraries,
+// under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-FileCopyrightText: Copyright (c) 2022-2024 NVIDIA CORPORATION & AFFILIATES.
```
miscco:
```suggestion
// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES.
```

cudax/include/cuda/experimental/__stf/places/data_place_interface.cuh
```diff
@@ -0,0 +1,258 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of CUDASTF in CUDA C++ Core Libraries,
+// under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-FileCopyrightText: Copyright (c) 2022-2024 NVIDIA CORPORATION & AFFILIATES.
```
miscco:
```suggestion
// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES.
```

cudax/include/cuda/experimental/__stf/internal/logical_data.cuh
caugonnet: This is not a bad idea because for example we are missing device_auto in the possible invalid values here, (even if i'll probably get rid of it)

```
    invalid     = ::std::numeric_limits<int>::min(),
    device_auto = -4,
    affine      = -3,
```

caugonnet: I've added it, and made the "choice" that composite places (which are a grid of places) are not a concrete place

cudax/include/cuda/experimental/__stf/places/exec/green_context.cuh
```diff
@@ -127,6 +129,14 @@ public:
       cuda_safe_call(cudaFreeAsync(ptr, stream));
     }
 
+    bool allocation_is_stream_ordered() const override
+    {
+      return true;
+    }
+
+    // Provide affine exec place implementation for data_place::affine_exec_place()
+    ::std::shared_ptr<void> get_affine_exec_impl() const override;
```
caugonnet: @andralex I can imagine the shared_ptr<void> is to resolve some cyclic dependencies ? Maybe it's worth a comment here because it feels strange

caugonnet: I see, later we have    

```
* Extensions override this to provide their affine exec_place implementation.
* Returns nullptr by default (non-extensions).
 * The returned shared_ptr should be castable to shared_ptr<exec_place::impl>.
```

cudax/include/cuda/experimental/__stf/places/exec/green_context.cuh
```diff
@@ -52,7 +52,7 @@ public:
   /**
    * @brief Extension implementation for green context data places
```
caugonnet: rename extension

cudax/include/cuda/experimental/__stf/places/data_place_interface.cuh
```diff
@@ -0,0 +1,271 @@
...
+   */
+  virtual bool is_composite() const
+  {
+    return false;
+  }
+
+  /**
+   * @brief Check if this is an extension-based place (green context, etc.)
+   */
+  virtual bool is_extension() const
```
caugonnet: this makes less sense now that everything is based on virtual methods, that's useful to make a difference with places that have an id ?

cudax/include/cuda/experimental/__stf/places/data_place_interface.cuh
```diff
@@ -0,0 +1,271 @@
...
+   */
+  virtual CUresult mem_create(CUmemGenericAllocationHandle*, size_t) const
+  {
+    return CUDA_ERROR_NOT_SUPPORTED;
+  }
+
+  // === Extension support ===
+
+  /**
+   * @brief Get the implementation for the affine exec_place (for extensions)
```
caugonnet: why for extensions ?

cudax/include/cuda/experimental/__stf/places/data_place_interface.cuh
```diff
@@ -0,0 +1,271 @@
...
+   *
+   * Extensions override this to provide their affine exec_place implementation.
+   * Returns nullptr by default (non-extensions).
+   * The returned shared_ptr should be castable to shared_ptr<exec_place::impl>.
+   */
+  virtual ::std::shared_ptr<void> get_affine_exec_impl() const
+  {
+    throw ::std::logic_error("get_affine_exec_impl() called on non-composite data_place");
```
caugonnet: is this "non concrete data place" instead ? in which case this would mean composite data places are concrete

cudax/include/cuda/experimental/__stf/places/data_place_interface.cuh
```diff
@@ -0,0 +1,271 @@
...
+namespace cuda::experimental::stf
+{
+// Forward declarations
+class exec_place;
+class exec_place_grid;
+class pos4;
+class dim4;
+
+//! Function type for computing executor placement from data coordinates
+using get_executor_func_t = pos4 (*)(pos4, dim4, dim4);
```
caugonnet, line 54: get_executor_func_t is really a poor name (not new to this PR)

cudax/include/cuda/experimental/__stf/places/data_place_interface.cuh
```diff
@@ -0,0 +1,271 @@
...
+   * @brief Special device ordinal values for non-device places
+   *
+   * Returned by get_device_ordinal() for places that don't correspond
+   * to a specific CUDA device.
+   */
+  enum ord : int
+  {
+    invalid     = ::std::numeric_limits<int>::min(),
+    composite   = -5,
+    device_auto = -4,
```
caugonnet, line 78: We need to remove data_place::device_auto and exec_place::device_auto as this is unmaintained in practice.

cudax/include/cuda/experimental/__stf/internal/logical_data.cuh
caugonnet: (#8006 to remove device_auto)

cudax/include/cuda/experimental/__stf/places/places.cuh
```diff
@@ -27,7 +27,8 @@
 #  pragma system_header
 #endif // no system header
 
-#include <cuda/experimental/__stf/places/data_place_extension.cuh>
+#include <cuda/experimental/__stf/internal/interpreted_execution_policy.cuh>
```
caugonnet, line 30: #7897 was supposed to make this include unnecessary

cudax/include/cuda/experimental/__stf/places/places.cuh
```diff
@@ -60,31 +61,49 @@ class exec_place_green_ctx;
...
-  // Constructors and factory functions below forward to this.
-  explicit data_place(int devid)
-      : devid(devid)
+  // Private constructor from interface pointer
+  explicit data_place(::std::shared_ptr<data_place_interface> impl)
+      : pimpl_(mv(impl))
   {}
 
+  template <typename T>
+  static ::std::shared_ptr<data_place_interface> make_static_instance()
```
caugonnet, line 77: Super cool ! This is a nice way to support places which are "names" such as devices, or ugpus which are identified by just device id and ugpu id

We probably need the same for exec_place

cudax/include/cuda/experimental/__stf/places/exec/green_context.cuh
andralex: Yah, and once we also change the design of `exec_place` this might change a bit.

cudax/include/cuda/experimental/__stf/places/data_place_interface.cuh
andralex: will destroy

cudax/include/cuda/experimental/__stf/localization/composite_slice.cuh
```diff
@@ -264,7 +264,7 @@ public:
       int item_dev = device_ordinal(item.place);
 
       // Physically allocate this block on the appropriate device/place
-      // Use the data_place's mem_create which delegates to extensions for custom behavior
+      // Use the data_place's mem_create which may implement custom behavior
       cuda_safe_call(item.place.mem_create(&item.alloc_handle, item.size));
```
caugonnet, line 268: should we have a better management of incompatible places not implementing mem_create ? (for another PR)

cudax/include/cuda/experimental/__stf/places/exec/green_context.cuh
```diff
@@ -40,136 +40,96 @@
...
-     * @brief Deallocate memory for green context
-     */
-    void deallocate(void* ptr, size_t /*size*/, cudaStream_t stream) const override
-    {
-      cuda_safe_call(cudaFreeAsync(ptr, stream));
-    }
+  void* allocate(::std::ptrdiff_t size, cudaStream_t stream) const override
+  {
+    void* result = nullptr;
+    cuda_safe_call(cudaSetDevice(view_.devid));
```
caugonnet, line 94: something smells, there is a set device, and nothing to set the device back. Allocate maybe already assume we have set the place ? I don't think so, but it looks like a bug

cudax/include/cuda/experimental/__stf/places/data_place_impl.cuh
```diff
@@ -0,0 +1,382 @@
...
+
+    SCOPE(exit)
+    {
+      if (prev_dev != device_id_)
+      {
+        cuda_safe_call(cudaSetDevice(prev_dev));
+      }
+    };
+
+    cuda_safe_call(cudaFreeAsync(ptr, stream));
```
caugonnet, line 290: in this case, we take care to set the device to call cudaFreeAsync, i'm not even sure it's needed. In the green context case we don't, but this might be on purpose, I need to check !

cudax/include/cuda/experimental/__stf/places/places.cuh
andralex: will add a commit for this

cudax/include/cuda/experimental/__stf/places/data_place_interface.cuh
andralex: I asked cursor for a better name. It said `partitioner_func_t`. Let's leave that to a future PR.

cudax/include/cuda/experimental/__stf/places/data_place_interface.cuh
andralex: Are you sure? It's used in a few places. LMK.

cudax/include/cuda/experimental/__stf/localization/composite_slice.cuh
andralex: You mean throwing vs returning an error code? Yah, we should.

Issue #7297, opened 2026-01-20
[FEA]: Thrust vector's should provide `emplace_back(Args...) -> T&`


bernhardmgruber:
### Is this a duplicate?

- [x] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cccl/issues) for this request and that I agree to the [Code of Conduct](CODE_OF_CONDUCT.md)

### Area

Thrust

### Is your feature request related to a problem? Please describe.

```c++
  thrust::host_vector<thrust::device_vector<int>> out_buffers;
  c2h::gen(C2H_SEED(1), out_buffers.emplace_back(42));
```
should compile.

### Describe the solution you'd like

Implement `emplace_back(Args...) -> T&` for thrust vectors

viralbhadeshiya:
@bernhardmgruber I like to work on this issue, if that's okay.

viralbhadeshiya:
Hey @bernhardmgruber or anyone, I have this done, but I want to test it and I can't build_thrust from main without error. Do we have stable version in main which should successfully build thrust test?

bernhardmgruber:
> I have this done, but I want to test it and I can't build_thrust from main without error. Do we have stable version in main which should successfully build thrust test?

`main` should be stable since it passed our CI. Please open a PR with your changes so I can test them against the CI.

miscco:
Are we really sure we want this?

AFAIK the reason we did not do this is because this allows people to use a massively inefficient API

bernhardmgruber:
> AFAIK the reason we did not do this is because this allows people to use a massively inefficient API

Only for device vectors, for which performance would be as bad as `operator[]`, which we also provide. I see your point though.

rbourgeois33:
Hello !
 I'd like to help with this issue. I see that the previous attempt (https://github.com/NVIDIA/cccl/pull/7981) has not been updated since March 12.

Is it okay if I work on this ? I do not want to step on anyone's toes. 

@bernhardmgruber @viralbhadeshiya

PR #7981, opened 2026-03-11
Implement emplace_back for thrust vector

viralbhadeshiya:
## Description
- added vector method for emplace_back for device and host vectors.

closes https://github.com/NVIDIA/cccl/issues/7297

## Checklist

### DIFF
--- a/thrust/testing/vector.cu
+++ b/thrust/testing/vector.cu
@@ -2,6 +2,7 @@
 
 #include <thrust/device_malloc_allocator.h>
 #include <thrust/sequence.h>
+#include <thrust/universal_vector.h>
 
 #include <initializer_list>
 #include <limits>
@@ -807,6 +808,20 @@ struct IntWithInit
   int value = 42;
 };
 
+struct ExplicitCtor
+{
+  int value;
+
+  _CCCL_HOST_DEVICE explicit ExplicitCtor(int v)
+      : value(v)
+  {}
+
+  _CCCL_HOST_DEVICE bool operator==(const ExplicitCtor& other) const
+  {
+    return value == other.value;
+  }
+};
+
 void TestVectorDefaultInitCtor()
 {
   // trivially-constructible type: just compilation test, since we cannot check that initialization was skipped
@@ -923,6 +938,19 @@ void TestVectorEmplaceBack()
     ASSERT_EQUAL(2, (int) dv[1]);
     ASSERT_EQUAL(3, (int) dv[2]);
   }
+
+  // universal_vector emplace_back
+  {
+    thrust::universal_vector<int> uv;
+    uv.emplace_back(1);
+    uv.emplace_back(2);
+    uv.emplace_back(3);
+
+    ASSERT_EQUAL(3lu, uv.size());
+    ASSERT_EQUAL(1, (int) uv[0]);
+    ASSERT_EQUAL(2, (int) uv[1]);
+    ASSERT_EQUAL(3, (int) uv[2]);
+  }
 }
 DECLARE_UNITTEST(TestVectorEmplaceBack);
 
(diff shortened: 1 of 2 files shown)


thrust/testing/vector.cu
```diff
@@ -895,3 +895,83 @@ void TestVectorNoInitResize()
...
+
+    ASSERT_EQUAL(3lu, hv.size());
+    ASSERT_EQUAL(1, hv[0]);
+    ASSERT_EQUAL(2, hv[1]);
+    ASSERT_EQUAL(3, hv[2]);
+  }
+
+  // device_vector emplace_back
+  {
+    thrust::device_vector<int> dv;
```
miscco, line 931: I believe we should also test universal_vector?

thrust/thrust/detail/vector_base.inl
```diff
@@ -585,6 +585,14 @@ void vector_base<T, Alloc>::push_back(const value_type& x)
   insert(end(), x);
 } // end vector_base::push_back()
 
+template <typename T, typename Alloc>
+template <typename... Args>
+typename vector_base<T, Alloc>::reference vector_base<T, Alloc>::emplace_back(Args&&... args)
+{
+  push_back(value_type(::cuda::std::forward<Args>(args)...));
+  return back();
+} // end vector_base::emplace_back()
```
miscco, line 594: Critical: This is inverted. We need to implement `push_back` through `emplace_back` and not the other way around

thrust/testing/vector.cu
```diff
@@ -895,3 +895,83 @@ void TestVectorNoInitResize()
...
+
+    ASSERT_EQUAL(1lu, hv.size());
+    ASSERT_EQUAL(42, ref);
+    ASSERT_EQUAL(42, hv[0]);
+  }
+
+  // device_vector emplace_back returns a reference to the inserted element
+  {
+    thrust::device_vector<int> dv;
+    auto ref = dv.emplace_back(42);
```
miscco: Critical: This is converting to a value.

Which I believe is working because we are returning a device_reference. I believe we should be a bit more explicit here:



```suggestion
    auto&& ref = dv.emplace_back(42);
    static_assert(cuda::std::is_same_v<decltype(ref), thrust::device_vector<int>::reference>);
```

thrust/testing/vector.cu
```diff
@@ -895,3 +895,83 @@ void TestVectorNoInitResize()
...
+DECLARE_UNITTEST(TestVectorEmplaceBackReturnsReference);
+
+void TestVectorEmplaceBackWithConstructorArgs()
+{
+  // This is the motivating use case from the issue:
+  // thrust::host_vector<thrust::device_vector<int>> out_buffers;
+  // out_buffers.emplace_back(42);
+  // should construct a device_vector<int> with 42 elements.
+  {
+    thrust::host_vector<thrust::device_vector<int>> out_buffers;
```
miscco, line 1026: Important: This should test a class type with an explicit constructor

thrust/testing/vector.cu
```diff
@@ -895,3 +895,83 @@ void TestVectorNoInitResize()
...
+  {
+    thrust::device_vector<int> dv;
+    auto ref = dv.emplace_back(42);
+
+    ASSERT_EQUAL(1lu, dv.size());
+    ASSERT_EQUAL(42, (int) ref);
+    ASSERT_EQUAL(42, (int) dv[0]);
+  }
+}
+DECLARE_UNITTEST(TestVectorEmplaceBackReturnsReference);
```
bernhardmgruber, line 991: Strong suggestion: if you use `DECLARE_VECTOR_UNITTEST` you'll get a bunch of different vector types to test for.

viralbhadeshiya: @bernhardmgruber you think it's good idea to use `DECLARE_VECTOR_UNITTEST` for all of the emplace_back tests?

thrust/testing/vector.cu
viralbhadeshiya: Done

thrust/thrust/detail/vector_base.inl
viralbhadeshiya: Done

thrust/testing/vector.cu
viralbhadeshiya: Done

thrust/testing/vector.cu
viralbhadeshiya: Done

thrust/testing/vector.cu
```diff
@@ -895,3 +910,134 @@ void TestVectorNoInitResize()
...
+}
+DECLARE_UNITTEST(TestVectorEmplaceBack);
+
+void TestVectorEmplaceBackReturnsReference()
+{
+  // host_vector emplace_back returns a reference to the inserted element
+  {
+    thrust::host_vector<int> hv;
+    auto& ref = hv.emplace_back(42);
+
```
miscco, line 963:
```suggestion
    auto&& ref = hv.emplace_back(42);
    static_assert(cuda::std::is_same_v<decltype(ref), thrust::host_vector<int>::reference>);

```

thrust/thrust/detail/vector_base.inl
```diff
@@ -582,9 +582,17 @@ _CCCL_HOST_DEVICE bool vector_base<T, Alloc>::empty() const
...
 {
-  insert(end(), x);
+  (void) emplace_back(x);
 } // end vector_base::push_back()
 
+template <typename T, typename Alloc>
+template <typename... Args>
+typename vector_base<T, Alloc>::reference vector_base<T, Alloc>::emplace_back(Args&&... args)
+{
+  insert(end(), value_type(::cuda::std::forward<Args>(args)...));
```
davebayer, line 592: This is not correct. `emplace_back` constructs the object in-place, so you should be using `emplace` instead of `insert`. If you use `insert`, the object is constructed on the stack and then moved to the end of the vector

thrust/testing/vector.cu
bernhardmgruber: Yes

PR #7983, opened 2026-03-11
Fix __nv_atomic_load guard from compiler to target based

### LINKED ISSUE #7982: [BUG]: sm_60 support broken for `cub::detail::find::agent_t`
### Is this a duplicate?

### Type of Bug

Compile-time Error

### Component

CUB

### Describe the bug

I am running into the following build failure while building for sm_60:
```
$PREFIX/include/rapids/cub/agent/agent_find.cuh(177): error: __nv_atomic_* functions are not supported on arch < sm_60.
          __nv_atomic_load(found_pos_ptr, &temp_storage.global_result, __NV_ATOMIC_RELAXED, __NV_THREAD_SCOPE_DEVICE);
                           ^
          detected during:
            instantiation of "void cub::detail::find::agent_t<AgentFindPolicy, InputIteratorT, OffsetT, PredicateT>::Process() [with AgentFindPolicy=cub::detail::find::agent_find_policy_t<128, 16, xgboost::FeatureType, 4, cub::LOAD_LDG, cub::detail::MemBoundScaling<128, 16, xgboost::FeatureType>>, InputIteratorT=const xgboost::FeatureType *, OffsetT=unsigned long long, PredicateT=xgboost::common::IsCatOp]" at line 59 of $PREFIX/include/rapids/cub/device/dispatch/dispatch_find.cuh
            instantiation of "void cub::detail::find::find_kernel<ChainedPolicyT,IteratorT,OffsetT,PredicateT>(IteratorT, OffsetT, OffsetT *, PredicateT) [with ChainedPolicyT=cub::detail::find::policy_hub_t<thrust::device_ptr<const xgboost::FeatureType>>::policy_300_t, IteratorT=const xgboost::FeatureType *, OffsetT=unsigned long long, PredicateT=xgboost::common::IsCatOp]" at line 117 of $PREFIX/include/rapids/cub/device/dispatch/dispatch_find.cuh
            instantiation of "cudaError_t cub::detail::find::dispatch_t<InputIteratorT, OutputIteratorT, OffsetT, PredicateT, PolicyHub>::Invoke<ActivePolicyT>() [with InputIteratorT=thrust::device_ptr<const xgboost::FeatureType>, OutputIteratorT=unsigned long long *, OffsetT=unsigned long long, PredicateT=xgboost::common::IsCatOp, PolicyHub=cub::detail::find::policy_hub_t<thrust::device_ptr<const xgboost::FeatureType>>, ActivePolicyT=cub::detail::find::policy_hub_t<thrust::device_ptr<const xgboost::FeatureType>>::policy_300_t]" at line 846 of $PREFIX/include/rapids/cub/util_device.cuh
            instantiation of "cudaError_t cub::ChainedPolicy<PolicyPtxVersion, PolicyT, PrevPolicyT>::find_and_invoke_policy<DevicePtxVersion,FunctorT>(FunctorT &) [with PolicyPtxVersion=300, PolicyT=cub::detail::find::policy_hub_t<thrust::device_ptr<const xgboost::FeatureType>>::policy_300_t, PrevPolicyT=cub::detail::find::policy_hub_t<thrust::device_ptr<const xgboost::FeatureType>>::policy_300_t, DevicePtxVersion=500, FunctorT=cub::detail::find::dispatch_t<thrust::device_ptr<const xgboost::FeatureType>, unsigned long long *, unsigned long long, xgboost::common::IsCatOp, cub::detail::find::policy_hub_t<thrust::device_ptr<const xgboost::FeatureType>>>]" at line 831 of $PREFIX/include/rapids/cub/util_device.cuh
            instantiation of "cudaError_t cub::ChainedPolicy<PolicyPtxVersion, PolicyT, PrevPolicyT>::runtime_arch_to_compiletime<ArchMult,CudaArches...,FunctorT>(int, FunctorT &) [with PolicyPtxVersion=300, PolicyT=cub::detail::find::policy_hub_t<thrust::device_ptr<const xgboost::FeatureType>>::policy_300_t, PrevPolicyT=cub::detail::find::policy_hub_t<thrust::device_ptr<const xgboost::FeatureType>>::policy_300_t, ArchMult=1, CudaArches=<500, 520, 600, 610, 700, 750, 800, 860, 890, 900, 1000, 1030, 1200, 1210>, FunctorT=cub::detail::find::dispatch_t<thrust::device_ptr<const xgboost::FeatureType>, unsigned long long *, unsigned long long, xgboost::common::IsCatOp, cub::detail::find::policy_hub_t<thrust::device_ptr<const xgboost::FeatureType>>>]" at line 796 of $PREFIX/include/rapids/cub/util_device.cuh
            [ 6 instantiation contexts not shown ]
            instantiation of "InputIt thrust::cuda_cub::find_if(thrust::cuda_cub::execution_policy<Derived> &, InputIt, InputIt, Predicate) [with Derived=thrust::cuda_cub::tag, InputIt=thrust::device_ptr<const xgboost::FeatureType>, Predicate=xgboost::common::IsCatOp]" at line 57 of $PREFIX/include/rapids/thrust/detail/find.inl
            instantiation of "InputIterator thrust::find_if(const thrust::detail::execution_policy_base<DerivedPolicy> &, InputIterator, InputIterator, Predicate) [with DerivedPolicy=thrust::cuda_cub::tag, InputIterator=thrust::device_ptr<const xgboost::FeatureType>, Predicate=xgboost::common::IsCatOp]" at line 34 of $PREFIX/include/rapids/thrust/system/detail/generic/logical.h
            instantiation of "__nv_bool thrust::system::detail::generic::any_of(thrust::execution_policy<ExecutionPolicy> &, InputIterator, InputIterator, Predicate) [with ExecutionPolicy=thrust::cuda_cub::tag, InputIterator=thrust::device_ptr<const xgboost::FeatureType>, Predicate=xgboost::common::IsCatOp]" at line 57 of $PREFIX/include/rapids/thrust/detail/logical.inl
            instantiation of "__nv_bool thrust::any_of(const thrust::detail::execution_policy_base<DerivedPolicy> &, InputIterator, InputIterator, Predicate) [with DerivedPolicy=thrust::cuda_cub::tag, InputIterator=thrust::device_ptr<const xgboost::FeatureType>, Predicate=xgboost::common::IsCatOp]" at line 96 of $PREFIX/include/rapids/thrust/detail/logical.inl
            instantiation of "__nv_bool thrust::any_of(InputIterator, InputIterator, Predicate) [with InputIterator=thrust::device_ptr<const xgboost::FeatureType>, Predicate=xgboost::common::IsCatOp]" at line 123 of $SRC_DIR/src/common/quantile.cuh
```

Found in https://github.com/rapidsai/xgboost-feedstock/pull/127#discussion_r2915430214

### How to Reproduce

No minimal reproducer yet. I will try to craft one using `thrust::find_if`.

### Expected behavior

The build should succeed for sm_60.

### Reproduction link

_No response_

### Operating System

_No response_

### nvidia-smi output

_No response_

### NVCC version

_No response_

gonidelis:
fixes #7982 

We used a rather modern atomic for our new cub::FindIf agent but we guarded it with merit to the compiler support. That does not suffice though as one might compile for `NVCC 12.9`  on `sm_50` which will throw as per the tracking issue.

We shift to using `__CUDACC_DEVICE_ATOMIC_BUILTINS__` which is a flag dedicated to atomic support instead.

### DIFF
--- a/libcudacxx/include/cuda/std/__cccl/builtin.h
+++ b/libcudacxx/include/cuda/std/__cccl/builtin.h
@@ -429,7 +429,7 @@
 #endif
 
 // The new __nv_atomic builtins are available when __CUDACC_DEVICE_ATOMIC_BUILTINS__ is defined
-#if defined(__CUDACC_DEVICE_ATOMIC_BUILTINS__) && !_CCCL_COMPILER(MSVC)
+#if defined(__CUDACC_DEVICE_ATOMIC_BUILTINS__) && _CCCL_PTX_ARCH() >= 600 && !_CCCL_COMPILER(MSVC)
 #  define _CCCL_HAS_NV_ATOMIC_BUILTINS() 1
 #else // ^^^ has intrinsics ^^^ / vvv no intrinsics
 #  define _CCCL_HAS_NV_ATOMIC_BUILTINS() 0


cub/cub/agent/agent_find.cuh
```diff
@@ -161,12 +161,12 @@ struct agent_t
       if (threadIdx.x == 0)
       {
         // __nv_atomic_load is a compiler build-in and compiles a lot faster
-#if _CCCL_CUDA_COMPILER(NVCC, >=, 12, 8)
+#if defined(_CUDACC_DEVICE_ATOMIC_BUILTINS__) && !_CCCL_COMPILER(MSVC)
```
gonidelis: is not using `ifdef` a deal breaker here?

cub/cub/agent/agent_find.cuh
elstehle: Don't we also need to guard the headers here?

cub/cub/agent/agent_find.cuh
miscco: 🙀 

This is a bug, we are using `_CUDACC_DEVICE_ATOMIC_BUILTINS__` but the builtin is defined with two leading underscores, so it should be `__CUDACC_DEVICE_ATOMIC_BUILTINS__`

Please fix all occurences you find

miscco: See https://docs.nvidia.com/cuda/cuda-programming-guide/05-appendices/cpp-language-extensions.html#cuda-feature-testing-macros

miscco: no we need it here because it does not allow `&&` afterwards

cub/cub/agent/agent_find.cuh
```diff
@@ -161,12 +161,12 @@ struct agent_t
       if (threadIdx.x == 0)
       {
         // __nv_atomic_load is a compiler build-in and compiles a lot faster
-#if _CCCL_CUDA_COMPILER(NVCC, >=, 12, 8)
+#if defined(__CUDACC_DEVICE_ATOMIC_BUILTINS__) && !_CCCL_COMPILER(MSVC)
```
davebayer: Can we move this check to our config and use something like `_CCCL_HAS_NV_ATOMIC_BUILTINS()` instead?

miscco: I used _CCCL_HAS_NV_ATOMIC_INTRINSICS()

cub/cub/agent/agent_find.cuh
gonidelis: i see pushed commits did you push the fix? @miscco

cub/cub/agent/agent_find.cuh
```diff
@@ -160,13 +162,13 @@ struct agent_t
       // Only one thread reads atomically and propagates it to other threads of the block through shared memory
       if (threadIdx.x == 0)
       {
+#if _CCCL_HAS_NV_ATOMIC_BUILTINS()
         // __nv_atomic_load is a compiler build-in and compiles a lot faster
-#if _CCCL_CUDA_COMPILER(NVCC, >=, 12, 8)
         __nv_atomic_load(found_pos_ptr, &temp_storage.global_result, __NV_ATOMIC_RELAXED, __NV_THREAD_SCOPE_DEVICE);
-#else
+#else // ^^^ _CCCL_HAS_NV_ATOMIC_BUILTINS() ^^^ / vvv !_CCCL_HAS_NV_ATOMIC_BUILTINS() vvv
```
davebayer, line 168: This is not sufficient, the `__CUDACC_DEVICE_ATOMIC_BUILTINS__` is always defined when the compiler has nv atomic builtins support, independently on the architecture. We should probably change the `_CCCL_HAS_NV_ATOMIC_BUILTINS()` macro definition to reflect that.

davebayer: See https://godbolt.org/z/v17rGf8b7

libcudacxx/include/cuda/std/__cccl/builtin.h
```diff
@@ -428,4 +428,11 @@
 #  define _CCCL_BUILTIN_STRLEN(...) __builtin_strlen(__VA_ARGS__)
 #endif
 
+// The new __nv_atomic builtins are available when __CUDACC_DEVICE_ATOMIC_BUILTINS__ is defined
+#if defined(__CUDACC_DEVICE_ATOMIC_BUILTINS__) && !_CCCL_COMPILER(MSVC)
```
davebayer:
```suggestion
#if defined(__CUDACC_DEVICE_ATOMIC_BUILTINS__) && __CUDA_ARCH__ >= 600 && !_CCCL_COMPILER(MSVC)
```

gonidelis: ~why?~  you explained it above
 ~and even still it shouldn't be `__CUDA_ARCH__ >= 500` ?~

gonidelis:
~since we got to that end why not just guard with the arch number per atomic? What's the point of `__CUDACC_DEVICE_ATOMIC_BUILTINS__` even?~

i am confused why was the bug disappearing without the extra arch check before? @davebayer

davebayer:
> i am confused why was the bug disappearing without the extra arch check before?

Because you had a typo in your implementation and instead of `__CUDACC_DEVICE_ATOMIC_BUILTINS__`, you were checking for `_CUDACC_DEVICE_ATOMIC_BUILTINS__` which was never defined, thus nv atomic builtins were always disabled

gonidelis:
> > i am confused why was the bug disappearing without the extra arch check before?
> 
> Because you had a typo in your implementation and instead of `__CUDACC_DEVICE_ATOMIC_BUILTINS__`, you were checking for `_CUDACC_DEVICE_ATOMIC_BUILTINS__` which was never defined, thus nv atomic builtins were always disabled

so the bug reappeared after we fixed the double underscore thingy?

cub/cub/agent/agent_find.cuh
miscco: This will happily break with NVC++ though

libcudacxx/include/cuda/std/__cccl/builtin.h
```diff
@@ -428,4 +428,11 @@
 #  define _CCCL_BUILTIN_STRLEN(...) __builtin_strlen(__VA_ARGS__)
 #endif
 
+// The new __nv_atomic builtins are available when __CUDACC_DEVICE_ATOMIC_BUILTINS__ is defined
+#if defined(__CUDACC_DEVICE_ATOMIC_BUILTINS__) && _CCCL_PTX_ARCH() >= 600 && !_CCCL_COMPILER(MSVC)
```
bernhardmgruber, line 432: Q: Why do we need to exclude MSVC here? I had to exclude it only in warpspeed scan because of compiler bugs. It should not generally be disabled IMO.

cub/cub/agent/agent_find.cuh
davebayer: Not true, but I agree that we should rather use `_CCCL_PTX_ARCH()`

bernhardmgruber: We officially don't support anything below sm_75, but I don't mind if you check for the `CUDA_ARCH >= 600` in `_CCCL_HAS_NV_ATOMIC_BUILTINS()`.

davebayer:
> so the bug reappeared after we fixed the double underscore thingy?

I believe so, but I haven't tested it

libcudacxx/include/cuda/std/__cccl/builtin.h
miscco: I believe for now the appropriate bugfix is to just fix the condition with the macro and then individually enable it for MSVC in a followup PR

bernhardmgruber: Fine. But we should not forget to enable the built-in atomics for MSVC where they work.

bernhardmgruber (approved):

libcudacxx/include/cuda/std/__cccl/builtin.h
gonidelis: i definitely copied over because of mighty legacy reasons that I wasn't aware

PR #7984, opened 2026-03-11
[libcu++] Add links to API reference docs in runtime section

pciolkosz:
The short description docs should link to the API docs in case a reader wants to learn more about specific APIs


docs/libcudacxx/runtime/buffer.rst
```diff
@@ -1,19 +1,28 @@
...
+.. _cuda_make_buffer: ../api/namespacecuda_1a8d909070d4cf758e776659b91e473a6f.html
+
 Buffer
 ======
 
 The buffer API provides a typed container allocated from memory resources. It handles stream-ordered allocation, initialization, and deallocation of memory.
 
-``cuda::buffer``
-----------------
+:cpp:any:`cuda::buffer`
```
miscco: Question: Shouldnt those be `:cpp:function:`?

PR #7988, opened 2026-03-11
Expand `DeviceSelect::Unique` to accept a predicate

miscco:
We want to use the public methods in the PSTL backend, so add an overload that takes a predicate.

### DIFF
--- a/cub/cub/device/device_select.cuh
+++ b/cub/cub/device/device_select.cuh
@@ -28,6 +28,8 @@
 #include <cuda/std/__functional/operations.h>
 #include <cuda/std/__type_traits/enable_if.h>
 #include <cuda/std/__type_traits/is_convertible.h>
+#include <cuda/std/__type_traits/is_same.h>
+#include <cuda/std/__type_traits/remove_cvref.h>
 #include <cuda/std/cstdint>
 
 CUB_NAMESPACE_BEGIN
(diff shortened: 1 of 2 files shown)


cub/cub/device/device_select.cuh
```diff
@@ -1502,6 +1504,138 @@ public:
     });
   }
 
+  //! @rst
+  //! Given an input sequence ``d_in`` having runs of consecutive equal-valued keys,
+  //! only the first key from each run is selectively copied to ``d_out``.
+  //! The total number of items selected is written to ``d_num_selected_out``.
+  //!
+  //! .. versionadded:: 2.2.0
+  //!    First appears in CUDA Toolkit 12.3.
```
bernhardmgruber:
```suggestion
  //! .. versionadded:: 3.4.0
  //!    First appears in CUDA Toolkit 13.4.
```

bernhardmgruber (commented):
Please also add tests for a predicate that is not the default one (`equal_to`).

bernhardmgruber:
@gonidelis please take note that we will need an env-overload for this new `Unique` overload as well!

elstehle:
> Please also add tests for a predicate that is not the default one (`equal_to`).

Agreed. If it helps: in `DeviceSelect::UniqueByKey`, which already does take a `equality_op` today, we're using:

```
template <typename T>
struct custom_equality_op
{
  T div_val;
  __host__ __device__ __forceinline__ bool operator()(const T& lhs, const T& rhs) const
  {
    return (lhs / div_val) == (rhs / div_val);
  }
};
```

elstehle (approved):
Apart from testing for an `equality_op` that's different from `cuda::std::equal_to`, this looks good.

cub/test/catch2_test_device_select_unique.cu
```diff
@@ -15,6 +15,15 @@
...
 #include <c2h/catch2_test_helper.h>
 
+struct fake_equal_to
+{
+  template <class T, class U>
+  [[nodiscard]] _CCCL_HOST_DEVICE constexpr bool operator()(const T& lhs, const U& rhs) const noexcept
+  {
+    return lhs == rhs;
+  }
+};
```
bernhardmgruber, line 26: Strong suggestion: Can we have a comparison predicate with different semantic? This way we would notice if the comparison predicate would not be taken into account somewhere. Like, someone forgot to call the predicate and just used `==` to compare.

bernhardmgruber (approved):

PR #7989, opened 2026-03-11
Avoid more "pointless comparison" warnings

miscco:
MSVC is really fond of those warnings. Add some more ternaries to avoid them.

Addresses parts of nvbug5970910


fbusato:
the most annoying and useless warning....

PR #7995, opened 2026-03-11
Diagnose iterators to inputs with `void` `value_type`

bernhardmgruber:
This detects some accidental use, notably passing a transform iterator with a __device__ callable and a deduced return type.

The specific internal bug report (on Slack) that this is solving did not occur before CCCL 3.3, since the vectorized tuning had the same threads per block as the other tunings. This changed with CCCL 3.3 and can thus lead to a launch failure, since the kernel's launch bounds will not match the host side launch config. We should thus backport this PR to 3.3.

### DIFF
--- a/cub/cub/device/dispatch/tuning/tuning_transform.cuh
+++ b/cub/cub/device/dispatch/tuning/tuning_transform.cuh
@@ -481,10 +481,10 @@ struct policy_selector_from_types<RequiresStableAddress,
                                   ::cuda::std::tuple<RandomAccessIteratorsIn...>,
                                   RandomAccessIteratorOut>
 {
-  // This can happen for multiple reasons. The user could pass an output iterator by accident, but it could also be a
-  // transform_iterator with a __device__ callable and a deduced return type (which is void in host code).
   static_assert((!::cuda::std::is_void_v<it_value_t<RandomAccessIteratorsIn>> && ...),
-                "Iterators for inputs must not have a value_type of void");
+                "Iterators for inputs must not have a value_type of void. This can happen for multiple reasons. You "
+                "could pass an output iterator by accident, but it could also be a transform_iterator with a "
+                "__device__ callable and a deduced return type (which is void in host code).");
 
   [[nodiscard]] _CCCL_API constexpr auto operator()(::cuda::arch_id arch) const -> transform_policy
   {


cub/cub/device/dispatch/tuning/tuning_transform.cuh
```diff
@@ -480,6 +481,11 @@ struct policy_selector_from_types<RequiresStableAddress,
                                   ::cuda::std::tuple<RandomAccessIteratorsIn...>,
                                   RandomAccessIteratorOut>
 {
+  // This can happen for multiple reasons. The user could pass an output iterator by accident, but it could also be a
+  // transform_iterator with a __device__ callable and a deduced return type (which is void in host code).
+  static_assert((!::cuda::std::is_void_v<it_value_t<RandomAccessIteratorsIn>> && ...),
+                "Iterators for inputs must not have a value_type of void");
+
```
miscco: This strongly suggests that we should require `proclaim_return_type`

bernhardmgruber: It's fine to just explicitly specify the return type. `proclaim_return_type` is for lambdas right?

miscco: yeah because that then lets us deduce the return type

miscco: Question: isnt this a breaking change, because AFAIK until now CUB allowed device lambdas to be passed just fine

miscco: Talking more internally, I believe querying the return type of a device function is indeed UB and we should be able to move forward with the change.

That said, I would love to extend the error message to point to this rather frequent issue

bernhardmgruber:
pre-commit.ci autofix

cub/test/test_device_transform_void_value_type_fail.cu
```diff
@@ -0,0 +1,108 @@
...
+  rhs_index_type second;
+};
+
+struct primitive_row_hasher
+{
+  size_type const* values;
+
+  __host__ __device__ static hash_value_type mix(hash_value_type x)
+  {
+    x ^= x >> 16;
```
pauleonix, line 41: Wouldn't it make sense to further simplify this test now that we know what caused the problem?

cub/test/test_device_transform_void_value_type_fail.cu
```diff
@@ -0,0 +1,108 @@
...
+};
+
+template <typename Hasher>
+struct masked_key_fn
+{
+  __host__ __device__ constexpr masked_key_fn(Hasher const& hasher)
+      : _hasher{hasher}
+  {}
+
+  __device__ __forceinline__ auto operator()(size_type i) const noexcept
```
pauleonix, line 62: Maybe add some comment marking the `auto` here as the issue?

PR #7980, opened 2026-03-10
Update libcudacxx header TU testing to reuse common cmake infra

jrhemstad:
## Description

closes #7979

Updates the libcudacxx header testing to use the common `cccl_generate_header_tests` machinery.

Issue #7920, opened 2026-03-06
[FEA]: Tool to Measure Header Hygiene (Transitive Includes & LOC)


jrhemstad:
### Is this a duplicate?

- [x] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cccl/issues) for this request and that I agree to the [Code of Conduct](CODE_OF_CONDUCT.md)

### Area

General CCCL

### Is your feature request related to a problem? Please describe.

Optimizing compile time is important for us, and a key part of this is header hygiene — ensuring public headers include only what they need. Currently, we have no quantitative measurement of "header hygiene" in the codebase.

**Problem:** We need a way to measure for each public header:
- How many other headers it transitively includes
- The number of lines of code (LOC) transitively pulled in

This measurement will help us identify problem areas and opportunities for improvement.

### Describe the solution you'd like

Create a script or tool (likely something based on clang tooling) that, when invoked on the cccl/ repo, recursively analyzes all public headers and produces for each:
- The count of other headers they transitively include
- The number of LOC they (transitively) include

The output should be provided in a standard format (e.g., CSV) for easy analysis.

### Additional context

The immediate goal is to collect baseline measurements. Longer term, we may want to automate regressions checks or hook this into CI.

fbusato:
we can use [clang-include-graph](https://github.com/bkryza/clang-include-graph) to generate the list (or graph) of all  transitively included headers, then use [cloc](https://github.com/AlDanial/cloc) or [tokei](https://github.com/XAMPPRocky/tokei) to get the number of lines.

I used a similar workflow for my c++ course to analyze the standard library.

<img width="892" height="821" alt="Image" src="https://github.com/user-attachments/assets/8ddd53e8-459b-42e7-9081-cfb93a2db164" />

PR #7996, opened 2026-03-12
Add a `profile_headers.sh` script to gather compile time per header in CCCL

jrhemstad:
## Description

Part of https://github.com/NVIDIA/cccl/issues/7920

Provides a new `ci/profile_headers.sh` script that hooks into our individual header TU testing to measure the compile each header. 

Merge after https://github.com/NVIDIA/cccl/pull/7980 

image


miscco:
I tried it locally and it does not include the internal headers from libc++ which would also be interesting

miscco:
@jrhemstad could you please add a third column that is just the number of occurences of that header in CCCL product code?

Its fine if there is a single header that is large, but if we have a header that is included a ton of times, we might want to make sure its small

should be fine to just grep for `include <$header>` in `cub/cub`, `thrust/thrust` and `libcudacxx/include`

jrhemstad:
Yeah, I was just working on something similar of trying a smarter ranking of the "cost" of a header by some function `f(compile time, number of times it is included)`

ci/profile_headers.sh
```diff
@@ -0,0 +1,126 @@
...
+status "Found ${#selected_tus[@]} header TUs."
+
+status "Counting direct include references across CCCL headers..."
+while IFS=$'\t' read -r header count; do
+  [[ -n "${header}" ]] || continue
+  included_by_count_map["${header}"]="${count}"
+done < <(
+  find \
+    "${repo_root}/cub" \
+    "${repo_root}/thrust" \
```
miscco: This also includes all tests, which massively distorts the results, please only greb in
```suggestion
    "${repo_root}/cub/cub" \
    "${repo_root}/thrust/thrust" \
```

ci/profile_headers.sh
```diff
@@ -0,0 +1,126 @@
...
+    "${repo_root}/cudax/include" \
+    -type f \
+    \( -name '*.h' -o -name '*.hpp' -o -name '*.cuh' -o ! -name '*.*' \) \
+    -print0 \
+    | xargs -0 grep -h -E '^[[:space:]]*#include[[:space:]]*[<"][^">]+[">]' \
+    | sed -E 's/^[[:space:]]*#include[[:space:]]*[<"]([^">]+)[">].*$/\1/' \
+    | sort \
+    | uniq -c \
+    | awk '{count=$1; $1=""; sub(/^ /, "", $0); print $0 "\t" count}'
+)
```
miscco: I would have expected something like 

`grep -R 'include <${header}>' cub/cub thrust/thrust libcudacxx/include | wc -l`

fbusato:
general question, how do we plan to handle the profiling results for different platform configurations?
there are many headers that depend on condition compilation (`#ifdef`) that could distort the evaluation

ci/profile_headers.sh
jrhemstad: That only counts for a single header. This counts for all headers at the same time (much faster).

jrhemstad:
> general question, how do we plan to handle the profiling results for different platform configurations? there are many headers that depend on condition compilation (`#ifdef`) that could distort the evaluation

Because I tied the header profiling to the cmake-driven per-header TU testing, we can reuse all of our common cmake machinery for configuring build environment like we would for any of our other test source files.

ci/profile_headers.sh
jrhemstad: Good catch.

ci/profile_headers.sh
jrhemstad: I realized counting direct include count like this is kinda meaningless.

Imagine `A.h` is expensive to compile and included only once in `B.h`, but `B.h` is included in 1000 other places. Counting this way would make it seem as though `A.h` is only included in 1 place when it is really included in 1000.

I've got an idea for a more accurate way to count this.

jrhemstad: image
I used `ctadvisor` to get the more accurate count of how many times a given header appears at least once in a public header.

miscco: I love it

miscco (approved):

PR #7999, opened 2026-03-12
Add env overloads for DeviceSegmentedRadixSort

### LINKED ISSUE #7549: cub::DeviceSegmentedRadixSort
- [ ] Add env-based cub::DeviceSegmentedRadixSort::SortPairs()
- [ ] Add env-based cub::DeviceSegmentedRadixSort::SortPairsDescending()
- [ ] Add env-based cub::DeviceSegmentedRadixSort::SortKeys()
- [ ] Add env-based cub::DeviceSegmentedRadixSort::SortKeysDescending()

gonidelis:
fixes #7549 

No deterministic guarantees. It's sorting, if it's not deterministic it's just wrong.

### DIFF
--- a/cub/test/catch2_test_device_segmented_radix_sort_env.cu
+++ b/cub/test/catch2_test_device_segmented_radix_sort_env.cu
@@ -333,3 +333,143 @@ TEST_CASE("DeviceSegmentedRadixSort::SortPairs uses custom stream", "[segmented_
 
   REQUIRE(cudaSuccess == cudaStreamDestroy(custom_stream));
 }
+
+TEST_CASE("DeviceSegmentedRadixSort::SortPairsDescending uses custom stream", "[segmented_radix_sort][device]")
+{
+  auto keys_in    = c2h::device_vector<int>{8, 6, 7, 5, 3, 0, 9};
+  auto keys_out   = c2h::device_vector<int>(7);
+  auto values_in  = c2h::device_vector<int>{0, 1, 2, 3, 4, 5, 6};
+  auto values_out = c2h::device_vector<int>(7);
+  auto offsets    = c2h::device_vector<int>{0, 3, 3, 7};
+
+  cudaStream_t custom_stream;
+  REQUIRE(cudaSuccess == cudaStreamCreate(&custom_stream));
+
+  size_t expected_bytes_allocated{};
+  REQUIRE(
+    cudaSuccess
+    == cub::DeviceSegmentedRadixSort::SortPairsDescending(
+      nullptr,
+      expected_bytes_allocated,
+      thrust::raw_pointer_cast(keys_in.data()),
+      thrust::raw_pointer_cast(keys_out.data()),
+      thrust::raw_pointer_cast(values_in.data()),
+      thrust::raw_pointer_cast(values_out.data()),
+      static_cast<::cuda::std::int64_t>(keys_in.size()),
+      static_cast<::cuda::std::int64_t>(3),
+      offsets.begin(),
+      offsets.begin() + 1));
+
+  auto stream_prop = stdexec::prop{cuda::get_stream_t{}, cuda::stream_ref{custom_stream}};
+  auto env         = stdexec::env{stream_prop, expected_allocation_size(expected_bytes_allocated)};
+
+  sort_pairs_descending(
+    thrust::raw_pointer_cast(keys_in.data()),
+    thrust::raw_pointer_cast(keys_out.data()),
+    thrust::raw_pointer_cast(values_in.data()),
+    thrust::raw_pointer_cast(values_out.data()),
+    static_cast<int>(keys_in.size()),
+    3,
+    offsets.begin(),
+    offsets.begin() + 1,
+    0,
+    static_cast<int>(sizeof(int) * 8),
+    env);
+
+  REQUIRE(cudaSuccess == cudaStreamSynchronize(custom_stream));
+
+  c2h::device_vector<int> expected_keys{8, 7, 6, 9, 5, 3, 0};
+  c2h::device_vector<int> expected_values{0, 2, 1, 6, 3, 4, 5};
+  REQUIRE(keys_out == expected_keys);
+  REQUIRE(values_out == expected_values);
+
+  REQUIRE(cudaSuccess == cudaStreamDestroy(custom_stream));
+}
+
+TEST_CASE("DeviceSegmentedRadixSort::SortKeys uses custom stream", "[segmented_radix_sort][device]")
+{
+  auto keys_in  = c2h::device_vector<int>{8, 6, 7, 5, 3, 0, 9};
+  auto keys_out = c2h::device_vector<int>(7);
+  auto offsets  = c2h::device_vector<int>{0, 3, 3, 7};
+
+  cudaStream_t custom_stream;
+  REQUIRE(cudaSuccess == cudaStreamCreate(&custom_stream));
+
+  size_t expected_bytes_allocated{};
+  REQUIRE(
+    cudaSuccess
+    == cub::DeviceSegmentedRadixSort::SortKeys(
+      nullptr,
+      expected_bytes_allocated,
+      thrust::raw_pointer_cast(keys_in.data()),
+      thrust::raw_pointer_cast(keys_out.data()),
+      static_cast<::cuda::std::int64_t>(keys_in.size()),
+      static_cast<::cuda::std::int64_t>(3),
+      offsets.begin(),
+      offsets.begin() + 1));
+
+  auto stream_prop = stdexec::prop{cuda::get_stream_t{}, cuda::stream_ref{custom_stream}};
+  auto env         = stdexec::env{stream_prop, expected_allocation_size(expected_bytes_allocated)};
+
+  sort_keys(
+    thrust::raw_pointer_cast(keys_in.data()),
+    thrust::raw_pointer_cast(keys_out.data()),
+    static_cast<int>(keys_in.size()),
+    3,
+    offsets.begin(),
+    offsets.begin() + 1,
+    0,
+    static_cast<int>(sizeof(int) * 8),
+    env);
+
+  REQUIRE(cudaSuccess == cudaStreamSynchronize(custom_stream));
+
+  c2h::device_vector<int> expected_keys{6, 7, 8, 0, 3, 5, 9};
+  REQUIRE(keys_out == expected_keys);
+
+  REQUIRE(cudaSuccess == cudaStreamDestroy(custom_stream));
+}
+
+TEST_CASE("DeviceSegmentedRadixSort::SortKeysDescending uses custom stream", "[segmented_radix_sort][device]")
+{
+  auto keys_in  = c2h::device_vector<int>{8, 6, 7, 5, 3, 0, 9};
+  auto keys_out = c2h::device_vector<int>(7);
+  auto offsets  = c2h::device_vector<int>{0, 3, 3, 7};
+
+  cudaStream_t custom_stream;
+  REQUIRE(cudaSuccess == cudaStreamCreate(&custom_stream));
+
+  size_t expected_bytes_allocated{};
+  REQUIRE(
+    cudaSuccess
+    == cub::DeviceSegmentedRadixSort::SortKeysDescending(
+      nullptr,
+      expected_bytes_allocated,
+      thrust::raw_pointer_cast(keys_in.data()),
+      thrust::raw_pointer_cast(keys_out.data()),
+      static_cast<::cuda::std::int64_t>(keys_in.size()),
+      static_cast<::cuda::std::int64_t>(3),
+      offsets.begin(),
+      offsets.begin() + 1));
+
+  auto stream_prop = stdexec::prop{cuda::get_stream_t{}, cuda::stream_ref{custom_stream}};
+  auto env         = stdexec::env{stream_prop, expected_allocation_size(expected_bytes_allocated)};
+
+  sort_keys_descending(
+    thrust::raw_pointer_cast(keys_in.data()),
+    thrust::raw_pointer_cast(keys_out.data()),
+    static_cast<int>(keys_in.size()),
+    3,
+    offsets.begin(),
+    offsets.begin() + 1,
+    0,
+    static_cast<int>(sizeof(int) * 8),
+    env);
+
+  REQUIRE(cudaSuccess == cudaStreamSynchronize(custom_stream));
+
+  c2h::device_vector<int> expected_keys{8, 7, 6, 9, 5, 3, 0};
+  REQUIRE(keys_out == expected_keys);
+
+  REQUIRE(cudaSuccess == cudaStreamDestroy(custom_stream));
+}


cub/test/catch2_test_device_segmented_radix_sort_env_api.cu
```diff
@@ -0,0 +1,159 @@
...
+    thrust::raw_pointer_cast(keys_out.data()),
+    thrust::raw_pointer_cast(values_in.data()),
+    thrust::raw_pointer_cast(values_out.data()),
+    static_cast<int>(keys_in.size()),
+    3,
+    offsets.begin(),
+    offsets.begin() + 1,
+    0,
+    sizeof(int) * 8,
+    env);
```
NaderAlAwar, line 80: Question: do we want to add a test where we do not explicitly pass an env? Other env API tests seem to have both kinds, explicit and implicit env objects.

gonidelis:
> Important: we are missing the DoubleBuffer overloads

@NaderAlAwar  will be added in subsequent pr for economy of reviewing effort per pr

cub/test/catch2_test_device_segmented_radix_sort_env_api.cu
gonidelis: i think it's fine. unit tests cover that case and users can see it throughout docs as soon as they start using our env overloads. don't wanna over bloat our test base for

cub/cub/device/device_segmented_radix_sort.cuh
```diff
@@ -419,6 +422,144 @@ public:
...
+  template <typename KeyT,
+            typename ValueT,
+            typename NumItemsT,
+            typename BeginOffsetIteratorT,
+            typename EndOffsetIteratorT,
+            typename EnvT                                                        = ::cuda::std::execution::env<>,
+            ::cuda::std::enable_if_t<::cuda::std::is_integral_v<NumItemsT>, int> = 0>
+  [[nodiscard]] CUB_RUNTIME_FUNCTION _CCCL_FORCEINLINE static cudaError_t SortPairs(
+    const KeyT* d_keys_in,
+    KeyT* d_keys_out,
```
bernhardmgruber: Q: Why do we need a constraint on this overload? Because of the second parameter `KeyT*`, it can't be ambiguous with the old overload, which has a `size_t&` here.

cub/cub/device/device_segmented_radix_sort.cuh
```diff
@@ -419,6 +422,144 @@ public:
...
+    const ValueT* d_values_in,
+    ValueT* d_values_out,
+    NumItemsT num_items,
+    NumItemsT num_segments,
```
bernhardmgruber: Important: the non-env overload uses `int64_t` for `num_items` and `num_segments`. Why do we make them template parameters here? I think we should retain `int64_t` here.

Applies to all newly added overloads.

cub/cub/device/device_segmented_radix_sort.cuh
```diff
@@ -419,6 +422,144 @@ public:
...
+    DoubleBuffer<ValueT> d_values(const_cast<ValueT*>(d_values_in), d_values_out);
+
+    return detail::dispatch_with_env(env, [&]([[maybe_unused]] auto tuning, void* storage, size_t& bytes, auto stream) {
+      return detail::radix_sort::dispatch<SortOrder::Ascending, SegmentSizeT>(
+        storage,
+        bytes,
+        d_keys,
+        d_values,
+        static_cast<::cuda::std::int64_t>(num_items),
+        static_cast<::cuda::std::int64_t>(num_segments),
```
bernhardmgruber: We can drop the cast after changing the parameter types. 

Applies to all newly added overloads.

cub/test/catch2_test_device_segmented_radix_sort_env_api.cu
bernhardmgruber: We have that test in the other file added in this PR.

pauleonix:
> No deterministic guarantees. It's sorting, if it's not deterministic it's just wrong.

Only because our radix-sort is stable by design. Unstable sort could be non-deterministic 😉

cub/cub/device/device_segmented_radix_sort.cuh
```diff
@@ -419,6 +422,139 @@ public:
...
+  //!   the latter is specified as ``segment_offsets + 1``).
+  //! - An optional bit subrange ``[begin_bit, end_bit)`` of differentiating key
+  //!   bits can be specified. This can reduce overall sorting overhead and
+  //!   yield a corresponding performance improvement.
+  //! - Note, the size of any segment may not exceed ``INT_MAX``. Please consider using ``DeviceSegmentedSort`` instead,
+  //!   if the size of at least one of your segments could exceed ``INT_MAX``.
+  //!
+  //! Snippet
+  //!
+  //! .. literalinclude:: ../../../cub/test/catch2_test_device_segmented_radix_sort_env_api.cu
```
pauleonix:
```suggestion
  //! .. literalinclude:: ../../test/catch2_test_device_segmented_radix_sort_env_api.cu
```

cub/cub/device/device_segmented_radix_sort.cuh
```diff
@@ -419,6 +422,139 @@ public:
...
+  //!
+  //! Sorts segments of key-value pairs into ascending order. (``~2N`` auxiliary storage required)
+  //!
+  //! .. versionadded:: 3.4.0
+  //!    First appears in CUDA Toolkit 13.4.
+  //!
+  //! This is an environment-based API that allows customization of:
+  //!
+  //! - Stream: Query via ``cuda::get_stream``
+  //! - Memory resource: Query via ``cuda::mr::get_memory_resource``
```
pauleonix: Not sure I like this better than the `- Can use a specific stream or cuda memory resource through the ``env`` parameter` I saw in the previous PR. I'm also not sure this list wont be rendered/read as part as the following list.

cub/cub/device/device_segmented_radix_sort.cuh
```diff
@@ -419,6 +422,139 @@ public:
...
+  //! - When input a contiguous sequence of segments, a single sequence
+  //!   ``segment_offsets`` (of length ``num_segments + 1``) can be aliased
+  //!   for both the ``d_begin_offsets`` and ``d_end_offsets`` parameters (where
+  //!   the latter is specified as ``segment_offsets + 1``).
+  //! - An optional bit subrange ``[begin_bit, end_bit)`` of differentiating key
+  //!   bits can be specified. This can reduce overall sorting overhead and
+  //!   yield a corresponding performance improvement.
+  //! - Note, the size of any segment may not exceed ``INT_MAX``. Please consider using ``DeviceSegmentedSort`` instead,
+  //!   if the size of at least one of your segments could exceed ``INT_MAX``.
+  //!
```
pauleonix, line 456: Maybe we should have a `@env` like the `@devicestorage` on the old APIs to avoid the repetition. Even if it needs multiple like `@env-with-guarantees`.

cub/cub/device/device_segmented_radix_sort.cuh
```diff
@@ -781,6 +917,139 @@ public:
...
+  //! - The contents of the input data are not altered by the sorting operation.
+  //! - When input a contiguous sequence of segments, a single sequence
+  //!   ``segment_offsets`` (of length ``num_segments + 1``) can be aliased
+  //!   for both the ``d_begin_offsets`` and ``d_end_offsets`` parameters (where
+  //!   the latter is specified as ``segment_offsets + 1``).
+  //! - An optional bit subrange ``[begin_bit, end_bit)`` of differentiating key
+  //!   bits can be specified. This can reduce overall sorting overhead and
+  //!   yield a corresponding performance improvement.
+  //! - Note, the size of any segment may not exceed ``INT_MAX``. Please consider using ``DeviceSegmentedSort`` instead,
```
pauleonix, line 963: Missing
```
  //! - Let ``in`` be one of ``{d_keys_in, d_values_in}`` and `out` be any of
  //!   ``{d_keys_out, d_values_out}``. The range ``[out, out + num_items)`` shall
  //!   not overlap ``[in, in + num_items)``,
  //!   ``[d_begin_offsets, d_begin_offsets + num_segments)`` nor
  //!   ``[d_end_offsets, d_end_offsets + num_segments)`` in any way.
```
and
```
  //! - Segments are not required to be contiguous. For all index values ``i``
  //!   outside the specified segments ``d_keys_in[i]``, ``d_values_in[i]``,
  //!   ``d_keys_out[i]``, ``d_values_out[i]`` will not be accessed nor modified.
```
Maybe also
```
  //! - @devicestorageNP For sorting using only ``O(P)`` temporary storage, see
  //!   the sorting interface using DoubleBuffer wrappers below.
```

cub/cub/device/device_segmented_radix_sort.cuh
```diff
@@ -419,6 +422,139 @@ public:
       stream);
   }
 
+  //! @rst
+  //! Overview
```
pauleonix, line 426: This "title" is new in comparison to the existing docs. If it stays, it looks like it should be underlined (rendered as a sub-title)
```suggestion
  //! Overview
  //! +++++++++++++++++++++++++++++++++++++++++++++
```
Same for `Snippet` below (which is not new but has this underline in existing docs)

cub/cub/device/device_segmented_radix_sort.cuh
```diff
@@ -1118,6 +1387,127 @@ public:
...
+  //!
+  //! - The contents of the input data are not altered by the sorting operation.
+  //! - When input a contiguous sequence of segments, a single sequence
+  //!   ``segment_offsets`` (of length ``num_segments + 1``) can be aliased
+  //!   for both the ``d_begin_offsets`` and ``d_end_offsets`` parameters (where
+  //!   the latter is specified as ``segment_offsets + 1``).
+  //! - An optional bit subrange ``[begin_bit, end_bit)`` of differentiating key
+  //!   bits can be specified. This can reduce overall sorting overhead and
+  //!   yield a corresponding performance improvement.
+  //! - Note, the size of any segment may not exceed ``INT_MAX``. Please consider using ``DeviceSegmentedSort`` instead,
```
pauleonix, line 1446: Missing some entries again. And the order of the ones not missing is changed as well.

cub/cub/device/device_segmented_radix_sort.cuh
```diff
@@ -1118,6 +1387,127 @@ public:
...
+  //!   The most-significant bit index (exclusive) needed for key comparison
+  //!   (e.g., ``sizeof(unsigned int) * 8``)
+  //!
+  //! @param[in] env
+  //!   **[optional]** Execution environment. Default is ``cuda::std::execution::env{}``.
+  template <typename KeyT,
+            typename BeginOffsetIteratorT,
+            typename EndOffsetIteratorT,
+            typename EnvT = ::cuda::std::execution::env<>>
+  [[nodiscard]] CUB_RUNTIME_FUNCTION _CCCL_FORCEINLINE static cudaError_t SortKeys(
```
pauleonix: Is `_CCCL_FORCEINLINE` necessary here? Existing overload does not have it.

cub/cub/device/device_segmented_radix_sort.cuh
```diff
@@ -1449,6 +1839,127 @@ public:
...
+  //!
+  //! - The contents of the input data are not altered by the sorting operation.
+  //! - When input a contiguous sequence of segments, a single sequence
+  //!   ``segment_offsets`` (of length ``num_segments + 1``) can be aliased
+  //!   for both the ``d_begin_offsets`` and ``d_end_offsets`` parameters (where
+  //!   the latter is specified as ``segment_offsets + 1``).
+  //! - An optional bit subrange ``[begin_bit, end_bit)`` of differentiating key
+  //!   bits can be specified. This can reduce overall sorting overhead and
+  //!   yield a corresponding performance improvement.
+  //! - Note, the size of any segment may not exceed ``INT_MAX``. Please consider using ``DeviceSegmentedSort`` instead,
```
pauleonix, line 1911: Missing entries again.

cub/test/catch2_test_device_segmented_radix_sort_env.cu
```diff
@@ -0,0 +1,335 @@
...
+      offsets.begin() + 1));
+
+  auto env = stdexec::env{expected_allocation_size(expected_bytes_allocated)};
+
+  sort_pairs_descending(
+    thrust::raw_pointer_cast(keys_in.data()),
+    thrust::raw_pointer_cast(keys_out.data()),
+    thrust::raw_pointer_cast(values_in.data()),
+    thrust::raw_pointer_cast(values_out.data()),
+    static_cast<int>(keys_in.size()),
```
pauleonix, line 199:
```suggestion
    static_cast<cuda::std::int64_t>(keys_in.size()),
```

cub/test/catch2_test_device_segmented_radix_sort_env.cu
```diff
@@ -0,0 +1,335 @@
...
+    offsets.begin() + 1,
+    0,
+    static_cast<int>(sizeof(int) * 8),
+    env);
+
+  c2h::device_vector<int> expected_keys{8, 7, 6, 9, 5, 3, 0};
+  REQUIRE(keys_out == expected_keys);
+}
+
+TEST_CASE("DeviceSegmentedRadixSort::SortPairs uses custom stream", "[segmented_radix_sort][device]")
```
pauleonix, line 285: Should this be tested for all overloads?

cub/test/catch2_test_device_segmented_radix_sort_env_api.cu
```diff
@@ -0,0 +1,159 @@
...
+  cuda::stream stream{cuda::devices[0]};
+  cuda::stream_ref stream_ref{stream};
+  auto env = cuda::std::execution::env{stream_ref};
+
+  auto error = cub::DeviceSegmentedRadixSort::SortPairs(
+    thrust::raw_pointer_cast(keys_in.data()),
+    thrust::raw_pointer_cast(keys_out.data()),
+    thrust::raw_pointer_cast(values_in.data()),
+    thrust::raw_pointer_cast(values_out.data()),
+    static_cast<int>(keys_in.size()),
```
pauleonix:
```suggestion
    static_cast<cuda::std::int64_t>(keys_in.size()),
```

cub/cub/device/device_segmented_radix_sort.cuh
bernhardmgruber: I considered suggesting the same! No need to do it in this PR though.

cub/cub/device/device_segmented_radix_sort.cuh
bernhardmgruber: Not necessary, please remove!

cub/test/catch2_test_device_segmented_radix_sort_env_api.cu
bernhardmgruber: Do we need a cast at all here? To suppress the unsigned -> signed warning, right?

cub/cub/device/device_segmented_radix_sort.cuh
gonidelis: changed it

cub/cub/device/device_segmented_radix_sort.cuh
gonidelis: this has already been suggested on a friday review meeting and it in my todo list. i will open an issue

cub/cub/device/device_segmented_radix_sort.cuh
gonidelis: thanks for the catch

PR #8002, opened 2026-03-12
Add enviornment overloads for `DeviceSegmentedSort::Keys`

gonidelis:
work towards resolving #7551 
 
This PR does not use the `DECLARE_LAUNCH_WRAPPER` / `stream_registry_factory_t` parameterized launch pattern (lid 0:1).

~`DeviceSegmentedSort` performs a `D->H memcpy` mid-dispatch via `launcher_factory.MemcpyAsync()` to read back partition group sizes before launching subsequent kernels. `stream_registry_factory_t` only implements kernel launch interception and does not provide `MemcpyAsync`, causing a compile error when used with this algorithm.~

As a byproduct it also implements `MemcpyAsync()` inside `catch2_test_env_launch_helper.h` so `DECLARE_LAUNCH_WRAPPER `can be used.

PR #8003, opened 2026-03-12
Add stable env `DeviceSegmentedSort::Pairs` 1/2

### LINKED ISSUE #7551: cub::DeviceSegmentedSort
- [ ] Add env-based cub::DeviceSegmentedSort::SortKeys()
- [ ] Add env-based cub::DeviceSegmentedSort::SortKeysDescending()
- [ ] Add env-based cub::DeviceSegmentedSort::StableSortKeys()
- [ ] Add env-based cub::DeviceSegmentedSort::StableSortKeysDescending()
- [ ] Add env-based cub::DeviceSegmentedSort::SortPairs()
- [ ] Add env-based cub::DeviceSegmentedSort::SortPairsDescending()
- [ ] Add env-based cub::DeviceSegmentedSort::StableSortPairs()
- [ ] Add env-based cub::DeviceSegmentedSort::StableSortPairsDescending()

gonidelis:
fixes #7551 

- [x] Follows exactly same logic as #8002 . Let's merge it after this one ^^^

### DIFF
--- a/cub/test/catch2_test_device_segmented_sort_pairs_env_api.cu
+++ b/cub/test/catch2_test_device_segmented_sort_pairs_env_api.cu
@@ -46,6 +46,7 @@ C2H_TEST("cub::DeviceSegmentedSort::StableSortPairs env-based API", "[segmented_
   thrust::device_vector<int> expected_keys{6, 7, 8, 0, 3, 5, 9};
   thrust::device_vector<int> expected_values{1, 2, 0, 5, 4, 3, 6};
   // example-end stable-sort-pairs-env
+  stream.sync();
 
   REQUIRE(error == cudaSuccess);
   REQUIRE(keys_out == expected_keys);
@@ -84,6 +85,7 @@ C2H_TEST("cub::DeviceSegmentedSort::StableSortPairsDescending env-based API", "[
   thrust::device_vector<int> expected_keys{8, 7, 6, 9, 5, 3, 0};
   thrust::device_vector<int> expected_values{0, 2, 1, 6, 3, 4, 5};
   // example-end stable-sort-pairs-descending-env
+  stream.sync();
 
   REQUIRE(error == cudaSuccess);
   REQUIRE(keys_out == expected_keys);
@@ -124,6 +126,7 @@ C2H_TEST("cub::DeviceSegmentedSort::StableSortPairs DoubleBuffer env-based API",
   thrust::device_vector<int> expected_keys{6, 7, 8, 0, 3, 5, 9};
   thrust::device_vector<int> expected_values{1, 2, 0, 5, 4, 3, 6};
   // example-end stable-sort-pairs-db-env
+  stream.sync();
 
   REQUIRE(error == cudaSuccess);
   thrust::device_vector<int> result_keys(d_keys.Current(), d_keys.Current() + 7);
@@ -168,6 +171,7 @@ C2H_TEST("cub::DeviceSegmentedSort::StableSortPairsDescending DoubleBuffer env-b
   thrust::device_vector<int> expected_keys{8, 7, 6, 9, 5, 3, 0};
   thrust::device_vector<int> expected_values{0, 2, 1, 6, 3, 4, 5};
   // example-end stable-sort-pairs-descending-db-env
+  stream.sync();
 
   REQUIRE(error == cudaSuccess);
   thrust::device_vector<int> result_keys(d_keys.Current(), d_keys.Current() + 7);


cub/test/catch2_test_device_segmented_sort_pairs_env.cu
NaderAlAwar, line 1: Important: can we add one representative MR/temp-storage env test here too? At the moment this only checks stream propagation, but these APIs also go through dispatch_with_env, so the memory-resource path is currently untested.

gonidelis: @NaderAlAwar isn't the mr tested by default in `lid_2` here? https://github.com/NVIDIA/cccl/blob/e1a7506d75e9a58b94974dcda8c27d29e9f6c9b5/cub/test/catch2_test_env_launch_helper.h#L466-L467

NaderAlAwar: That MR coverage exists for the `DECLARE_LAUNCH_WRAPPER` / `TEST_LAUNCH` path, but this file explicitly does not use that path, so it doesn’t apply here. This means that there is no `lid_2` variant here.

NaderAlAwar: Actually wait, I was wrong, I saw an older version of the PR, we do have `DECLARE_LAUNCH_WRAPPER` here. I still think we should not use `lid_2` for similar reasons to here https://github.com/NVIDIA/cccl/pull/8286#discussion_r3032805930

bernhardmgruber: @gonidelis please address this comment and change to `// %PARAM% TEST_LAUNCH lid 0:1`

cub/test/catch2_test_device_segmented_sort_pairs_env_api.cu
```diff
@@ -0,0 +1,173 @@
...
+
+  auto error = cub::DeviceSegmentedSort::StableSortPairs(
+    thrust::raw_pointer_cast(keys_in.data()),
+    thrust::raw_pointer_cast(keys_out.data()),
+    thrust::raw_pointer_cast(values_in.data()),
+    thrust::raw_pointer_cast(values_out.data()),
+    static_cast<int>(keys_in.size()),
+    2,
+    thrust::raw_pointer_cast(offsets.data()),
+    thrust::raw_pointer_cast(offsets.data()) + 1,
```
bernhardmgruber: I am confused here. We create a vector holding 3 offsets, then we say we have 2 segments, and then we pass a range consisting of a single element (begin iterator ` + 1). That does not make sense to me. Shouldn't we have a `+2` or `+3` here?

cub/test/catch2_test_device_segmented_sort_pairs_env.cu
gonidelis: done

cub/test/catch2_test_device_segmented_sort_pairs_env_api.cu
gonidelis: the offset vector defines the offset boundaries `[0,3),  [3,7)`, hence `num_segments = 2`. then for the end of **each** segment SegSort needs to look on the "next" +1 element of offset vector.

It's unfortunate cccl doesn't provide real compilable examples for SegmentedSort before these env examples I am adding here. I think I have an issue somewhere open to move the non-env examples to literalincludes and compile them. will dig up

https://github.com/gonidelis/cccl/blob/3beffdcebbf6bf15c7b84a88ee91be58e96c3aed/cub/cub/device/device_segmented_sort.cuh#L91-L92

Try:

```
...
  auto offsets    = thrust::device_vector<int>{0, 3, 3, 7, 7};
...
    thrust::raw_pointer_cast(offsets.data()) + 2,
...
```

to see that test still passes

I will separate the offsets.begin and offsets.end vectors to make it more readable

bernhardmgruber: I did some digging and I think I understood now how this is meant to be used. Sorry for the confusion! I think you can revert the last commit.

cub/cub/device/device_segmented_sort.cuh
```diff
@@ -18,10 +18,12 @@
 #endif // no system header
 
 #include <cub/detail/choose_offset.cuh>
+#include <cub/detail/env_dispatch.cuh>
 #include <cub/device/dispatch/dispatch_segmented_sort.cuh>
 #include <cub/util_namespace.cuh>
 
 #include <cuda/std/cstdint>
+#include <cuda/std/type_traits>
```
bernhardmgruber: Important: Please use the individual sub headers for the features you need. 

```suggestion
#include <cuda/std/__type_traits/enable_if.h>
```

cub/cub/device/device_segmented_sort.cuh
```diff
@@ -2319,6 +2321,142 @@ public:
...
+          d_temp_storage,
+          temp_storage_bytes,
+          d_keys,
+          d_values,
+          num_items,
+          num_segments,
+          d_begin_offsets,
+          d_end_offsets,
+          is_overwrite_okay,
+          stream);
```
bernhardmgruber: Important: this is equivalent to just calling `SortPairsNoNVTX`, which is shorter.

cub/cub/device/device_segmented_sort.cuh
```diff
@@ -2480,6 +2618,142 @@ public:
...
```
bernhardmgruber: Same here, use `SortPairsDescendingNoNVTX`

cub/cub/device/device_segmented_sort.cuh
bernhardmgruber: Please apply the more general refactoring as I did in: https://github.com/NVIDIA/cccl/pull/8002/changes/8861ee0ba41fddea8d21f04370aaa5499df42d06

cub/test/catch2_test_device_segmented_sort_pairs_env_api.cu
gonidelis: i feel the way i re-wrote is good to exist in our examples domain i ll keep it as is

cub/test/catch2_test_device_segmented_sort_pairs_env_api.cu
```diff
@@ -0,0 +1,177 @@
...
+    std::cerr << "cub::DeviceSegmentedSort::StableSortPairs failed with status: " << error << std::endl;
+  }
+
+  thrust::device_vector<int> expected_keys{6, 7, 8, 0, 3, 5, 9};
+  thrust::device_vector<int> expected_values{1, 2, 0, 5, 4, 3, 6};
+  // example-end stable-sort-pairs-env
+
+  REQUIRE(error == cudaSuccess);
+  REQUIRE(keys_out == expected_keys);
+  REQUIRE(values_out == expected_values);
```
bernhardmgruber, line 53: Important: please add a `stream.sync()` before touching data that was written by the CUB API call. Applies to all examples here using a custom stream.

PR #8005, opened 2026-03-12
Avoid including nvfp headers in `<cuda/std/limits>`

davebayer:
This should improve compile times of `<cuda/std/limits>`, which is modified to work with forward declarations of nvfp types exclusively without the need to include the full `<cuda_fpXY.h>` headers.

The idea is to put construction of all nvfp instances into an unevaluated context which is instantiated only once the function is actually used. It shifts the responsibility of including the full `<cuda_fpXY.h>` headers to the header's client (other header, user, ...), however, since they are using the type, they should be including the header anyway.

This requires us to write a bit more complicated code, for example:
```cpp
// old
template <class T>
struct __some_trait;
template <>
struct __some_trait<__half> { ... };

// new
template <class T, class U = T>
struct __some_trait;
template <class T>
struct __some_trait<__half, T> { ... };
```
but the benefits of reduced include times when the nvfp types are not actually needed are more valuable to us.

### DIFF
--- a/libcudacxx/test/libcudacxx/std/language.support/support.limits/limits/numeric.limits.members/common.h
+++ b/libcudacxx/test/libcudacxx/std/language.support/support.limits/limits/numeric.limits.members/common.h
@@ -14,8 +14,10 @@
 #include <disable_nvfp_conversions_and_operators.h>
 // clang-format on
 
-#include <cuda/std/__bit/bit_cast.h>
+#include <cuda/std/bit>
 #include <cuda/std/limits>
+
+#include "cuda_fp_types.h"
 
 template <class T>
 __host__ __device__ bool float_eq(T x, T y)
--- a/libcudacxx/include/cuda/std/__floating_point/format.h
+++ b/libcudacxx/include/cuda/std/__floating_point/format.h
@@ -21,7 +21,6 @@
 #  pragma system_header
 #endif // no system header
 
-#include <cuda/std/__floating_point/cuda_fp_types.h>
 #include <cuda/std/__fwd/fp.h>
 #include <cuda/std/__type_traits/is_same.h>
 #include <cuda/std/cfloat>
--- a/libcudacxx/include/cuda/std/__floating_point/storage.h
+++ b/libcudacxx/include/cuda/std/__floating_point/storage.h
@@ -22,7 +22,6 @@
 #endif // no system header
 
 #include <cuda/std/__bit/bit_cast.h>
-#include <cuda/std/__floating_point/cuda_fp_types.h>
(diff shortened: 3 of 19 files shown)


libcudacxx/test/libcudacxx/std/language.support/support.limits/limits/numeric.limits.members/traps.pass.cpp
```diff
@@ -13,9 +13,7 @@
 
 // traps
 
-#include <cuda/std/limits>
-
-#include "test_macros.h"
```
miscco, line 18: Please revert, we *always* include the appropriate header by convention

libcudacxx/test/libcudacxx/std/language.support/support.limits/limits/numeric.limits.members/common.h
```diff
@@ -14,9 +14,24 @@
...
 // clang-format on
 
-#include <cuda/std/__bit/bit_cast.h>
 #include <cuda/std/limits>
 
+// Check that we are not including nvfp headers in <cuda/std/limits> header.
+#if defined(__CUDA_FP16_H__) || defined(__CUDA_BF16_H__) || defined(__CUDA_FP8_H__) || defined(__CUDA_FP6_H__) \
+  || defined(__CUDA_FP4_H__)
+#  error "any of the nvfp headers was included by <cuda/std/limits>"
+#endif
```
miscco: Please move that into a separate test file

davebayer:
Please, don't merge before @jrhemstad approves this PR

miscco:
@jrhemstad I did a comparison of this PR with main and the results are
before:
```
cuda/std/limits, 376, 0.534079, 200.813785
```
after:
```
cuda/std/limits,376,0.071702,26.960061
```

Which is a sizeable improvement 🎉 

The full improvement  can be determined by running the profiling with the new commit and then determining the improvement via:
```
awk -F',' '{ if (FNR == NR) { diff[$1]-=$4} else { diff[$1]+=$4 } } END { for(header in diff) if (diff[header] > 0.1 || diff[header] < -0.1) print header, diff[header] }' profile_headers_all_before.csv profile_headers_all_after.csv | sort -g -k2 > profile_headers_all_diff_sorted.csv
```

This leads us to:
[results.txt](https://github.com/user-attachments/files/25977433/results.txt)

davebayer:
@jrhemstad can we merge this PR?

jrhemstad:
@davebayer please update the PR description with how this approach is working.

davebayer:
@jrhemstad I've updated the description :)

PR #8009, opened 2026-03-12
Make sure that `thrust::transform_iterator` has a non-void `value_type`

miscco:
We want to make sure that users do not fall into the trap of passing a `__device__` only function to `thrust::transform_iterator`

This will lead to a different `value_type` between host and device compilation, leading to multiple issues, e.g with algorithm tunings

### DIFF
--- a/thrust/thrust/iterator/transform_iterator.h
+++ b/thrust/thrust/iterator/transform_iterator.h
@@ -39,6 +39,7 @@
 #include <cuda/std/__memory/construct_at.h>
 #include <cuda/std/__type_traits/is_copy_assignable.h>
 #include <cuda/std/__type_traits/is_copy_constructible.h>
+#include <cuda/std/__type_traits/is_void.h>
 #include <cuda/std/__type_traits/remove_cvref.h>
 #include <cuda/std/__type_traits/remove_reference.h>
 
@@ -77,6 +78,10 @@ struct make_transform_iterator_base
 private:
   using reference  = replace_if_use_default<Reference, transform_iterator_reference<UnaryFunc, Iterator>>;
   using value_type = replace_if_use_default<Value, ::cuda::std::remove_cvref<reference>>;
+
+  static_assert(!::cuda::std::is_void_v<value_type>,
+                "thrust::transform_iterator requires a non-void value type. "
+                "A common error is using it with a __device__ only function.");
 
 public:
   using type =


thrust/thrust/iterator/transform_iterator.h
```diff
@@ -78,6 +79,10 @@ struct make_transform_iterator_base
   using reference  = replace_if_use_default<Reference, transform_iterator_reference<UnaryFunc, Iterator>>;
   using value_type = replace_if_use_default<Value, ::cuda::std::remove_cvref<reference>>;
 
+  static_assert(!::cuda::std::is_void_v<value_type>,
+                "thrust::transform_iterator requires a non-void value type. "
+                "A common error is using it with a __device__ only function.");
```
bernhardmgruber, line 84:
```suggestion
                "A common error is using it with a __device__ only function and a deduced return type.");
```

bernhardmgruber:
Thinking again about this, isn't this a huge breaking change? What if a user constructs a `transform_iterator` with a device only callable and a deduced return type, which is only used in device code? That's fine as long as no code ever asks for `transform_iterator::value_type` in host code. Of course the change proposed here is useful to detect buggy conditions, but it could break existing code that works by luck.

Should we move this PR to CCCL 4.0? or at least add a macro to disable the `static_assert`?

miscco:
Decided to drop it, because there are unclear issues wrt to user code

jrhemstad:
> We want to make sure that users do not fall into the trap of passing a __device__ only function to thrust::transform_iterator

This is incorrect, the purpose of this change was to avoid a `__device__` function with an `auto` return type. 

https://godbolt.org/z/sYf8qq8xv

This is similar to the guard we added to `cuda::std::invoke_result` to prevent attempting to query the return type of `__device__` lambdas in host code. 

https://github.com/NVIDIA/cccl/blob/2d6346a9e822aeb8224d8189fb59c0ad3c54a6d4/libcudacxx/include/cuda/std/__functional/invoke.h#L196

That said, as will all potentially breaking changes, we should add this with a macro people can define as an escape hatch.

miscco:
> This is similar to the guard we added to `cuda::std::invoke_result` to prevent attempting to query the return type of `__device__` lambdas in host code.

The issue we saw is that currently a user might have a device only lambda that is only ever invoked in a device path.

However, that static assert added would be evaluated both in the device *and* the host path, which would break technically valid user code.

I am not sure whether this is something we can just add outside of a major release, even if we provide an escape hatch

PR #8010, opened 2026-03-12
Properly support complex64 in nvbench_helper

bernhardmgruber:
We only supported `complex` (which is `complex<float>`)

### DIFF
--- a/cub/benchmarks/bench/transform/babelstream.cu
+++ b/cub/benchmarks/bench/transform/babelstream.cu
@@ -32,7 +32,7 @@ using element_types =
                      std::int16_t,
                      float,
                      double
-#  ifdef _CCCL_HAS_INT128()
+#  if _CCCL_HAS_INT128()
                      ,
                      __int128
 #  endif
--- a/cub/benchmarks/bench/transform/complex_cmp.cu
+++ b/cub/benchmarks/bench/transform/complex_cmp.cu
@@ -37,11 +37,11 @@ try
     return;
   }
 
-  thrust::device_vector<complex> in = generate(n);
+  thrust::device_vector<complex32> in = generate(n);
   thrust::device_vector<bool> out(n - 1);
 
   state.add_element_count(n);
-  state.add_global_memory_reads<complex>(n);
+  state.add_global_memory_reads<complex32>(n);
   state.add_global_memory_writes<bool>(n);
 
   // the complex comparison needs lots of compute and transform reads from overlapping input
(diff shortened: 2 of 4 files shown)


nvbench_helper/nvbench_helper/nvbench_helper.cuh
```diff
@@ -41,8 +42,8 @@ NVBENCH_DECLARE_TYPE_STRINGS(cuda::std::complex<__half>, "C16", "complex_half");
 NVBENCH_DECLARE_TYPE_STRINGS(__nv_bfloat16, "BF16", "bfloat16");
 NVBENCH_DECLARE_TYPE_STRINGS(cuda::std::complex<__nv_bfloat16>, "CB16", "complex_bfloat16");
 #endif
-NVBENCH_DECLARE_TYPE_STRINGS(complex, "C32", "complex32");
-NVBENCH_DECLARE_TYPE_STRINGS(cuda::std::complex<double>, "C64", "complex64");
+NVBENCH_DECLARE_TYPE_STRINGS(complex32, "C32", "complex32");
```
bernhardmgruber, line 45: This was actually a bug, because when tuning for `C32`, the tuning framework would inject `complex32` as type name and then fail to compile, because the alias some lines above was called `complex` and not `complex32`.

nvbench_helper/nvbench_helper/nvbench_helper.cuh
```diff
@@ -113,7 +114,7 @@ using all_types =
 #  endif
                      float,
                      double,
-                     complex>;
+                     complex32>;
```
bernhardmgruber, line 117: I am not adding `complex64` here to not increase benchmark times in general. The type is only uses in the warp reduction benchmarks.

nvbench_helper/nvbench_helper/nvbench_helper.cu
```diff
@@ -191,30 +191,29 @@ struct and_t
...
+    const T a_imag = a.imag();
 
-    const std::uint64_t result_real =
-      cuda::std::bit_cast<std::uint64_t>(a_real) & cuda::std::bit_cast<std::uint64_t>(b_real);
+    const T b_real = b.real();
+    const T b_imag = b.imag();
 
-    const std::uint64_t result_imag =
-      cuda::std::bit_cast<std::uint64_t>(a_imag) & cuda::std::bit_cast<std::uint64_t>(b_imag);
+    using uint_t           = std::conditional_t<sizeof(T) == 4, std::uint32_t, std::uint64_t>;
```
miscco, line 203:
```suggestion
    using uint_t           = cuda::std::conditional_t<sizeof(T) == 4, std::uint32_t, std::uint64_t>;
```

bernhardmgruber: Would require an additional include and this file is already using `<type_traits>`.

PR #7926, opened 2026-03-08
Add support for small, variable-size segments to `DeviceBatchedTopK`

elstehle:
Closes https://github.com/NVIDIA/cccl/issues/7839.

The PR is mostly concerned with extending tests for variable-size segments and secondarily lifting a few prior restrictions and guards for fixed-size segments.

PR #8013, opened 2026-03-12
Addresses a few follow-up comments for supporting variable-size segments in `DeviceSegmentedTopK`

elstehle:
Addresses follow-up comments from https://github.com/NVIDIA/cccl/pull/7926#pullrequestreview-3915042473

### DIFF
--- a/cub/cub/block/specializations/block_topk_air.cuh
+++ b/cub/cub/block/specializations/block_topk_air.cuh
@@ -237,13 +237,13 @@ private:
     // The total number of selected items
     total_selected = 0;
 
-    const int total_bits = (::max) (end_bit - begin_bit, 0);
+    const int total_bits = (::cuda::std::max) (end_bit - begin_bit, 0);
     const int num_passes = ::cuda::ceil_div(total_bits, RadixBits);
     for (int pass = 0; pass < num_passes; ++pass)
     {
       // Bit-range & mask of the current pass
       const int pass_end_bit           = end_bit - pass * RadixBits;
-      const int pass_begin_bit         = (::max) (pass_end_bit - RadixBits, begin_bit);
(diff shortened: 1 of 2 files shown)


cub/cub/agent/agent_batched_topk.cuh
```diff
@@ -150,8 +150,8 @@ struct agent_batched_topk_worker_per_segment
 
     // Resolve Segment Parameters
     const auto segment_size = segment_sizes.get_param(segment_id);
-    const auto k            = ::cuda::std::min(
-      k_param.get_param(segment_id), static_cast<decltype(k_param.get_param(segment_id))>(segment_size));
+    const auto k =
+      (::min) (k_param.get_param(segment_id), static_cast<decltype(k_param.get_param(segment_id))>(segment_size));
```
bernhardmgruber: Q: This confuses me. C++ only has `std::min`, and C has no `::min`. There is a `min` in POSIX, but we don't have that on Windows. Where does this function come from?

fbusato: `::min` is always available in device code. It could save compile time compared to `std::min` (only integer values)

bernhardmgruber: So it's a CUDA extension?

`std::min` works for FP types as well. I am not worried about the compile-time of calling `std::min` TBH.

elstehle: Yeah, I think that's a valid point from Bernhard. @fbusato, is it only compilation-time improvements, no runtime performance benefit to be expected?

elstehle: Since `::cuda::std::min` is generating the same sass as `::min`, I'll use `::cuda::std::min/max`:
https://ce.nvidia.com/z/1vbvMx

elstehle:
pre-commit.ci autofix

PR #8014, opened 2026-03-12
[cudax] Add replicate and join for streams

pciolkosz:
This PR adds two APIs operating on streams: `replicate` and `join`.

`replicate` takes a stream and creates a specified number of new streams that target the same logical device as the input stream. It's a useful API to create side streams from a given stream that will target the same device along with the same green context restrictions. There is a static version that returns an array that can be easily unpacked with structured bindings and a dynamic version that returns a vector. In case user wants to keep the created streams in the same object as the input stream, there is also `replicate_prepend`, that takes the input stream as an rvalue and includes it in the output at index 0.

`join` is an API that establishes dependencies across two ranges of streams (`from_range` and `to_range`). It ensures all streams in `to_range` will depend on all work already scheduled in `from_range`. There are also two `join` overloads that take a single `stream_ref` as `from` or `to`.
The implementation is a naive for loop that inserts all pairs of event dependencies inside. There is a way to do it smarter where all dependencies are reduced into one stream first and then propagated among streams in `to` range, but that introduces an extra dependency on the stream used for the reduction. The naive implementation was selected, because the user can still opt-in to the smarter behavior by calling `join` twice:
```
cudax::join(to_set, from_set[0]);
cudax::join(from_set[0], from_set);
```
We can decide to change the implementation to that later or request a accumulating event record from CUDA to implement the smart version that does not introduce the extra dependency.

Unfortunately event in CUDA is bound to a specific context, while join should work with set of streams across context. There is a special handling of that in `join`, where first time it is detected that `to` set contains streams from different devices an event is allocated for every stream, instead of reusing one across all streams. There could be a way to optimize it better in the future if needed

### DIFF
--- a/cudax/include/cuda/experimental/__stream/stream.cuh
+++ b/cudax/include/cuda/experimental/__stream/stream.cuh
@@ -265,14 +265,14 @@ inline void __join_impl(const _ToRange& __to_streams, const _FromRange& __from_s
   auto __first                 = stream_ref(*__to_begin);
   auto __first_device          = __first.device();
   bool __use_per_stream_events = false;
-  cuda::event __event(__first_device);
+  cuda::event __local_event(__first_device);
   for (const auto& __from_stream : __from_streams)
   {
     auto __from = stream_ref(__from_stream);
 
     if (!__use_per_stream_events)
     {
-      auto __status = ::cuda::__driver::__eventRecordNoThrow(__event.get(), __from.get());
+      auto __status = ::cuda::__driver::__eventRecordNoThrow(__local_event.get(), __from.get());
       if (__status == cudaSuccess)
       {
         for (const auto& __to_stream : __to_streams)
@@ -280,7 +280,7 @@ inline void __join_impl(const _ToRange& __to_streams, const _FromRange& __from_s
           auto __to = stream_ref(__to_stream);
           if (__to != __from)
           {
-            __to.wait(__event);
+            __to.wait(__local_event);
           }
         }
         continue;
--- a/cudax/include/cuda/experimental/__stream/stream.cuh
+++ b/cudax/include/cuda/experimental/__stream/stream.cuh
@@ -163,6 +163,21 @@ private:
   {}
 };
 
+template <size_t... _Idx>
+[[nodiscard]] inline auto __replicate_impl(logical_device __dev, ::cuda::std::index_sequence<_Idx...>)
+  -> ::cuda::std::array<stream, sizeof...(_Idx)>
+{
+  return ::cuda::std::array<stream, sizeof...(_Idx)>{((void) _Idx, stream(__dev))...};
+}
+
+template <size_t... _Idx>
+[[nodiscard]] inline auto
+__replicate_prepend_impl(stream&& __source, logical_device __dev, ::cuda::std::index_sequence<_Idx...>)
+  -> ::cuda::std::array<stream, sizeof...(_Idx) + 1>
+{
+  return ::cuda::std::array<stream, sizeof...(_Idx) + 1>{::cuda::std::move(__source), ((void) _Idx, stream(__dev))...};
+}
+
 template <size_t _Count>
 //! @brief Create a fixed-size group of streams on the same logical device as `__source`.
 //!
@@ -172,12 +187,7 @@ template <size_t _Count>
 [[nodiscard]] auto replicate(stream_ref __source) -> ::cuda::std::array<stream, _Count>
 {
   auto __dev = __source.logical_device();
-
-  auto __streams = [&]<size_t... _Idx>(::cuda::std::index_sequence<_Idx...>) {
-    return ::cuda::std::array<stream, _Count>{((void) _Idx, stream(__dev))...};
-  }(::cuda::std::make_index_sequence<_Count>{});
-
-  return __streams;
+  return __replicate_impl(__dev, ::cuda::std::make_index_sequence<_Count>{});
 }
 
 //! @brief Create a runtime-sized group of streams on the same logical device as `__source`.
@@ -228,12 +238,7 @@ template <size_t _Count>
 [[nodiscard]] inline auto replicate_prepend(stream&& __source) -> ::cuda::std::array<stream, _Count + 1>
 {
   auto __dev = __source.logical_device();
-
-  auto __streams = [&]<size_t... _Idx>(::cuda::std::index_sequence<_Idx...>) {
-    return ::cuda::std::array<stream, _Count + 1>{::cuda::std::move(__source), ((void) _Idx, stream(__dev))...};
-  }(::cuda::std::make_index_sequence<_Count>{});
-
-  return __streams;
+  return __replicate_prepend_impl(::cuda::std::move(__source), __dev, ::cuda::std::make_index_sequence<_Count>{});
 }
 
 template <class _Range>
--- a/cudax/include/cuda/experimental/__stream/stream.cuh
+++ b/cudax/include/cuda/experimental/__stream/stream.cuh
@@ -27,16 +27,15 @@
 #include <cuda/__driver/driver_api.h>
 #include <cuda/__runtime/api_wrapper.h>
 #include <cuda/__stream/invalid_stream.h>
-
-#include <cuda/experimental/__device/logical_device.cuh>
-#include <cuda/experimental/__stream/stream_ref.cuh> // IWYU pragma: export
-#include <cuda/experimental/__utility/ensure_current_device.cuh>
-
 #include <cuda/std/__ranges/concepts.h>
 #include <cuda/std/array>
 #include <cuda/std/span>
 #include <cuda/std/type_traits>
 #include <cuda/std/utility>
+
+#include <cuda/experimental/__device/logical_device.cuh>
+#include <cuda/experimental/__stream/stream_ref.cuh> // IWYU pragma: export
+#include <cuda/experimental/__utility/ensure_current_device.cuh>
 
 #include <vector>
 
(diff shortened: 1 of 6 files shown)


cudax/include/cuda/experimental/__stream/stream.cuh
```diff
@@ -154,6 +162,190 @@ private:
       : stream_ref(__handle)
   {}
 };
+
+template <size_t... _Idx>
+[[nodiscard]] inline auto __replicate_impl(logical_device __dev, ::cuda::std::index_sequence<_Idx...>)
+  -> ::cuda::std::array<stream, sizeof...(_Idx)>
+{
+  return ::cuda::std::array<stream, sizeof...(_Idx)>{((void) _Idx, stream(__dev))...};
```
caugonnet: So the new stream state is entirely defined by the logical_device __dev ? We drop flags and priority for example ? (we could query them from cudaStreamGetFlags if the goal is to have streams with the exact same properties)

pciolkosz: We don't have NULL stream, we don't use stream flags, because for us blocking stream does not exist.
For priority I initially queried it from the input stream, but I think usage of stream priorities is rare enough we can skip that and just ask users to change the stream priority manually. But I am not strongly opinionated on that

caugonnet: Me neither i don't have any strong opinion (nor do i ever use priorities), I misinterpreted the goal which i believed to be creating arrays of streams with identical properties, it's more identical context, be it green or primary ...

pciolkosz: My logic was that we don't establish dependencies in `replicate` anyway, `join` should be used instead, so the state of the resulting streams are not exactly the same and maybe for properties we can focus only on the most important ones. But we can still change that if it seems confusing

bdice (approved):
Seems fine to me. I think the static-sized `replicate` is not especially useful. If you wanted to cut the scope, I don't think that feature is strictly necessary.

pciolkosz:
> I think the static-sized replicate is not especially useful

I think its nice for structured bindings like:
auto [left, right] = cuda::replicate<2>(start);
So you end up with names for all created streams

bdice:
> So you end up with names for all created streams

Yeah, I see the intent. I would only want to name things if there were 4 or fewer, and generally if I'm forking/replicating a stream, I want to have some larger number of streams that wouldn't be suitable to name. It's fine to have the feature, I just don't expect to reach for it very often.

cudax/include/cuda/experimental/__stream/stream.cuh
```diff
@@ -154,6 +162,190 @@ private:
       : stream_ref(__handle)
   {}
 };
+
+template <size_t... _Idx>
+[[nodiscard]] inline auto __replicate_impl(logical_device __dev, ::cuda::std::index_sequence<_Idx...>)
```
davebayer:
```suggestion
template <::cuda::std::size_t... _Idx>
[[nodiscard]] _CCCL_HOST_API auto __replicate_impl(logical_device __dev, ::cuda::std::index_sequence<_Idx...>)
```
We shouldn't suppose to have types in global namespace available. `inline` is not needed when the function is a `template`. `_CCCL_HOST_API` is missing

Applies to other places, too

cudax/include/cuda/experimental/__stream/stream.cuh
```diff
@@ -154,6 +162,190 @@ private:
...
+{
+  return ::cuda::std::array<stream, sizeof...(_Idx) + 1>{::cuda::std::move(__source), ((void) _Idx, stream(__dev))...};
+}
+
+template <size_t _Count>
+//! @brief Create a fixed-size group of streams on the same logical device as `__source`.
+//!
+//! @note This function only creates streams; it does not add synchronization dependencies.
+//! @note The streams are created with the default priority, which can be changed by the user. // TODO expose stream
+//! attributes?
```
davebayer:
```suggestion

//! @brief Create a fixed-size group of streams on the same logical device as `__source`.
//!
//! @note This function only creates streams; it does not add synchronization dependencies.
//! @note The streams are created with the default priority, which can be changed by the user. // TODO expose stream
//! attributes?
template <::cuda::std::size_t _Count>
```

cudax/include/cuda/experimental/__stream/stream.cuh
```diff
@@ -154,6 +162,190 @@ private:
...
+[[nodiscard]] auto replicate(stream_ref __source) -> ::cuda::std::array<stream, _Count>
+{
+  auto __dev = __source.logical_device();
+  return __replicate_impl(__dev, ::cuda::std::make_index_sequence<_Count>{});
+}
+
+//! @brief Create a runtime-sized group of streams on the same logical device as `__source`.
+//!
+//! @note This function only creates streams; it does not add synchronization dependencies.
+[[nodiscard]] inline auto replicate(stream_ref __source, size_t __count) -> ::std::vector<stream>
```
davebayer: As discussed offline, I'm not a fan of dedicating these generic names to a very specific single purpose like this. I'd rather use something like `replicate_submitter`

cudax/include/cuda/experimental/__stream/stream.cuh
```diff
@@ -154,6 +162,190 @@ private:
...
+  return __streams;
+}
+
+template <size_t _Count>
+//! @brief Create a fixed-size group of streams on the same logical device as `__source` and prepend `__source` at
+//! index 0.
+//!
+//! @note This function only creates streams; it does not add synchronization dependencies.
+//! The returned array has size `_Count + 1` and owns all streams.
+[[nodiscard]] inline auto replicate_prepend(stream&& __source) -> ::cuda::std::array<stream, _Count + 1>
```
davebayer: I don't find this name intuitive

cudax/include/cuda/experimental/__stream/stream.cuh
```diff
@@ -24,13 +24,21 @@
...
 
 #include <cuda/__device/device_ref.h>
+#include <cuda/__driver/driver_api.h>
 #include <cuda/__runtime/api_wrapper.h>
 #include <cuda/__stream/invalid_stream.h>
+#include <cuda/std/__ranges/concepts.h>
+#include <cuda/std/array>
+#include <cuda/std/span>
+#include <cuda/std/type_traits>
+#include <cuda/std/utility>
```
miscco: Please modularize into the individual headers needed

cudax/include/cuda/experimental/__stream/stream.cuh
```diff
@@ -154,6 +162,190 @@ private:
       : stream_ref(__handle)
   {}
 };
+
+template <size_t... _Idx>
```
miscco: Critical: We do not want this in this header. 

This is not an implementation detail of a stream, but a feature built around it. Please move to a separate header inside `<cudax/__stream/...>`

cudax/include/cuda/experimental/__stream/stream.cuh
```diff
@@ -154,6 +162,190 @@ private:
...
+template <class _ToRange, class _FromRange>
+inline void __join_impl(const _ToRange& __to_streams, const _FromRange& __from_streams)
+{
+  auto __to_begin = ::cuda::std::ranges::begin(__to_streams);
+  if (__to_begin == ::cuda::std::ranges::end(__to_streams))
+  {
+    return;
+  }
+
+  auto __first                 = stream_ref(*__to_begin);
```
miscco: Nitpick: use list initilization
```suggestion
  auto __first                 = stream_ref{*__to_begin};
```

Appliesbelow

cudax/include/cuda/experimental/__stream/stream.cuh
```diff
@@ -154,6 +162,190 @@ private:
...
+      // switch to per-source-stream events; otherwise propagate the failure.
+      if (__from.device() != __first_device)
+      {
+        __use_per_stream_events = true;
+      }
+      else
+      {
+        _CCCL_THROW(::cuda::cuda_error, __status, "Failed to record an event");
+      }
+    }
```
miscco: Important: This function is really complicated because it crams the handling of the events into the whole logic.

Can we factor that out into a single function that either returns a 
`auto __from_event = cuda::event(__from);` or does the special event handling here?

cudax/include/cuda/experimental/__stream/stream.cuh
```diff
@@ -154,6 +162,190 @@ private:
...
+//! The returned array has size `_Count + 1` and owns all streams.
+[[nodiscard]] inline auto replicate_prepend(stream&& __source) -> ::cuda::std::array<stream, _Count + 1>
+{
+  auto __dev = __source.logical_device();
+  return __replicate_prepend_impl(::cuda::std::move(__source), __dev, ::cuda::std::make_index_sequence<_Count>{});
+}
+
+template <class _Range>
+_CCCL_CONCEPT __stream_join_range =
+  ::cuda::std::is_same_v<::cuda::std::ranges::range_value_t<const _Range&>, stream_ref>
```
miscco: Nitpick: Consider using `__one_of` from `cudax/include/cuda/experimental/__detail/type_traits.cuh`

cudax/include/cuda/experimental/__stream/stream_utils.cuh
```diff
@@ -0,0 +1,231 @@
...
+  return __streams;
+}
+
+//! @brief Create a fixed-size group of streams on the same logical device as `__source` and prepend `__source` at
+//! index 0.
+//!
+//! @note This function only creates streams; it does not add synchronization dependencies.
+//! The returned array has size `_Count + 1` and owns all streams.
+template <::cuda::std::size_t _Count>
+[[nodiscard]] _CCCL_HOST_API inline auto replicate_prepend(stream&& __source) -> ::cuda::std::array<stream, _Count + 1>
```
pciolkosz: replicate_into?

PR #8018, opened 2026-03-13
Add env overloads for Select::* in place algorithms

### LINKED ISSUE #7945: [FEA]: Add in-place env overloads for `DeviceSelect`
Going breadth first approach we only added env overloads for the standard overloads given that users can substitute their in place use cases with the standard one. Add env overloads for the in place ones for full coverage.

gonidelis:
Adds in-place `DeviceSelect` environment overloads 

fixes #7945

### DIFF
--- a/cub/cub/device/device_select.cuh
+++ b/cub/cub/device/device_select.cuh
@@ -334,7 +334,7 @@ public:
   //! The total number of items selected is written to ``d_num_selected_out``.
   //!
   //! .. versionadded:: 3.4.0
-  //!    First appears in CUDA Toolkit 12.4.
+  //!    First appears in CUDA Toolkit 13.4.
   //!
   //! This is an environment-based API that allows customization of:
   //!
@@ -448,8 +448,10 @@ public:
   //! - Stream: Query via ``cuda::get_stream``
   //! - Memory resource: Query via ``cuda::mr::get_memory_resource``
   //!
-  //! - Copies of the selected items are compacted in ``d_data`` and maintain
-  //!   their original relative ordering.
+  //! - The value type of ``d_flags`` must be castable to ``bool`` (e.g., ``bool``, ``char``, ``int``, etc.).
+  //! - Copies of the selected items are compacted in-place and maintain their original relative ordering.
+  //! - | The ``d_data`` may equal ``d_flags``. The range ``[d_data, d_data + num_items)`` shall not overlap
+  //!   | ``[d_flags, d_flags + num_items)`` in any other way.
   //!
   //!
   //! Snippet
@@ -531,7 +533,7 @@ public:
   //! The total number of items selected is written to ``d_num_selected_out``.
   //!
   //! .. versionadded:: 3.4.0
-  //!    First appears in CUDA Toolkit 12.4.
+  //!    First appears in CUDA Toolkit 13.4.
   //!
   //! This is an environment-based API that allows customization of:
   //!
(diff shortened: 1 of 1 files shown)


cub/test/catch2_test_device_select_env_api.cu
```diff
@@ -99,6 +99,89 @@ C2H_TEST("cub::DeviceSelect::FlaggedIf accepts env with stream", "[select][env]"
...
+  thrust::device_vector<int> expected_num_selected{4};
+  // example-end select-flagged-inplace-env
+
+  REQUIRE(error == cudaSuccess);
+  REQUIRE(num_selected == expected_num_selected);
+  data.resize(num_selected[0]);
+  REQUIRE(data == expected_output);
+}
+
+C2H_TEST("cub::DeviceSelect::If in-place accepts env with stream", "[select][env]")
```
miscco, line 129: We dopcument:
>  //! - | The range ``[d_out, d_out + *d_num_selected_out)`` shall not overlap
  //!   | ``[d_in, d_in + num_items)`` nor ``d_num_selected_out`` in any way.

Is that actually valid? Asking for someone that needs that

bernhardmgruber: `cub::DeviceSelect::If` has two overloads, an in-place one and one with separate source and destination iterators.

cub/cub/device/device_select.cuh
```diff
@@ -436,6 +436,96 @@ public:
...
+  //!    First appears in CUDA Toolkit 13.4.
+  //!
+  //! This is an environment-based API that allows customization of:
+  //!
+  //! - Stream: Query via ``cuda::get_stream``
+  //! - Memory resource: Query via ``cuda::mr::get_memory_resource``
+  //!
+  //! - Copies of the selected items are compacted in ``d_data`` and maintain
+  //!   their original relative ordering.
+  //!
```
NaderAlAwar, line 455: Important: the non env docs on lines 724-735 say this

```
  //! - The value type of ``d_flags`` must be castable to ``bool`` (e.g., ``bool``, ``char``, ``int``, etc.).
  //! - Copies of the selected items are compacted in-place and maintain their original relative ordering.
  //! - | The ``d_data`` may equal ``d_flags``. The range ``[d_data, d_data + num_items)`` shall not overlap
  //!   | ``[d_flags, d_flags + num_items)`` in any other way.
```

If these apply here, we should add them here as well

cub/cub/device/device_select.cuh
```diff
@@ -1280,6 +1459,110 @@ public:
...
+  //! This is an environment-based API that allows customization of:
+  //!
+  //! - Stream: Query via ``cuda::get_stream``
+  //! - Memory resource: Query via ``cuda::mr::get_memory_resource``
+  //!
+  //! - The expression ``select_op(flag)`` must be convertible to ``bool``,
+  //!   where the type of ``flag`` corresponds to the value type of ``FlagIterator``.
+  //! - Copies of the selected items are compacted in ``d_data`` and maintain
+  //!   their original relative ordering.
+  //!
```
NaderAlAwar, line 1482: Important: same as above, this is missing info from the non env overload on lines 1246-1248

```
  //! - | The ``d_data`` may equal ``d_flags``. The range ``[d_data, d_data + num_items)`` shall not overlap
  //!   | ``[d_flags, d_flags + num_items)`` in any other way.
```

cub/cub/device/device_select.cuh
NaderAlAwar: Suggestion: I think it's worth doing a driveby fix to change 12.4 -> 13.4 in the right places in this file

PR #8019, opened 2026-03-13
libcu++ Style `SKILL.md`

fbusato:
## Description

After experimenting for a while, I created this PR to propose the first `SKILL.md` for CCCL.

The SKILL targets the high-level coding "style" of libcu++. 

Tested mostly with Claude, the agent was able to follow most of the rules at first attempt. When it was unable to do so, I used the failure to improve the skill.

Unfortunately, Codex and Claude use two different directories for skills. I'm not sure if there is a general solution that can address the problem on both Linux and Windows.

My aspiration for the future is to prevent CI errors in non-standard cases (e.g. old GCC and MSVC) by explaining how to address them before the code is pushed.


.agent/skills/licudacxx-style/SKILL.md
```diff
@@ -0,0 +1,86 @@
+---
+name: libcudacxx-style
+description: Make the code in libcudacxx/include/cuda, cudax/include compliant with the coding style
```
miscco:
```suggestion
description: Make the code in libcudacxx/include, cudax/include compliant with the coding style 
```

.agent/skills/licudacxx-style/SKILL.md
```diff
@@ -0,0 +1,86 @@
...
+
+# libcudacxx Style
+
+## Naming style
+
+- Macros: macro style, e.g. `MY_MACRO`.
+- Template parameters: CamelCase, e.g. `MyParameter`.
+- All other symbols: snake style, e.g. `my_variable`.
+
+All non-public symbols must have a specific prefix:
```
miscco:
```suggestion
All non-public symbols must be C++ reserved identifiers:
```

.agent/skills/licudacxx-style/SKILL.md
```diff
@@ -0,0 +1,86 @@
...
+- All variables that can be evaluated at compile-time must use `constexpr`.
+- Consider using plural names for array, span, list, e.g. `int values[4]` instead of `int value[4]`.
+
+## Function
+
+Declaration/Definition:
+
+- All functions must be marked `_CCCL_HOST_API`, `_CCCL_DEVICE_API`, or `_CCCL_API`.
+- Non-template, non-`constexpr` functions must use `inline`.
+- All functions with a non-void return type must use `[[nodiscard]]`.
```
miscco: That is not the right constraint, functions that have side effect should not be nodiscard
```suggestion
- Most functions with a non-void return type shall use `[[nodiscard]]`. Exceptions are functions with known side effects, e.g. `cuda::std::copy`
```

.agent/skills/licudacxx-style/SKILL.md
```diff
@@ -0,0 +1,86 @@
...
+- All functions that don't throw exception must use `noexcept`. This can be omitted for `_CCCL_DEVICE_API`.
```
miscco: There is no benefit to ommiting it for device functions and maybe one day we will get device exceptions
```suggestion
- All functions that don't throw exception must use `noexcept`
```

.agent/skills/licudacxx-style/SKILL.md
```diff
@@ -0,0 +1,86 @@
...
+- `constexpr` must be used for all functions that don't depend on run-time features, e.g. pointers.
+- Return type must be explicit (non `auto`) when possible.
```
miscco: I am not sure whether that is consensus. 

I personally prefer this, but I have see a lot of people use auto and trailing return types
```suggestion
- If the return type is not explicit (`auto`), then a trailing return type is strongly preferred, e.g. `auto abs(float) -> float`
```

.agent/skills/licudacxx-style/SKILL.md
```diff
@@ -0,0 +1,86 @@
...
+
+Function call:
+
+- All functions must have fully qualified namespace starting from the global namespace, e.g. `::cuda::ceil_div`. This includes calls to functions defined in the same namespace, e.g. inside `cuda::`, call `::cuda::ceil_div(...)`, not `ceil_div(...)`.
```
miscco:
```suggestion
- All calls to free functions must be fully qualified starting from the global namespace, e.g. `::cuda::ceil_div`. This includes calls to functions defined in the same namespace, e.g. inside `cuda::`, call `::cuda::ceil_div(...)`, not `ceil_div(...)`. This does not apply to (static) member functions of classes.
```

.agent/skills/licudacxx-style/SKILL.md
```diff
@@ -0,0 +1,86 @@
...
+- Types must have fully qualified namespace, except when they are already declared in the current namespace.
+- This includes standard integer type aliases (`::cuda::std::size_t`, `::cuda::std::uintptr_t`, `::cuda::std::int32_t`, etc.) and any other `cuda::std` or standard library types. A local `using` declaration (e.g. `using ::cuda::std::size_t;`) is acceptable to avoid repetition within a function body.
+
+## Headers
+
+- All header inclusions must use the syntax `<header>`.
+- Files must include all headers related to the symbols that they are using.
+- No transitive header inclusion are allowed.
+- Unneeded headers must be removed.
+- The headers must be the most precise one, e.g. `#include <cuda/std/__type_traits/is_array>`.
```
miscco:
```suggestion
- The headers must be the most precise one, e.g. `#include <cuda/std/__type_traits/is_array.h>`.
```

.agent/skills/licudacxx-style/SKILL.md
```diff
@@ -0,0 +1,86 @@
...
+- Headers in `cuda/std/__cccl/` must not be included directly (they are provided by `__config` or the prologue/epilogue mechanism).
+
+- All headers must have the license boilerplate.
```
miscco:
```suggestion
- All headers must have the correct license. If the file is ported from LLVM libc++ then we *must* use the LLVM license.
```

.agent/skills/licudacxx-style/SKILL.md
```diff
@@ -0,0 +1,86 @@
...
+- Use Doxygen-style `//! @brief comments`.
+- All functions, except `_impl` functions, must have complete Doxygen documentation: `//! @brief`, `//! @param[in/out/in,out]` for every parameter, and `//! @return` for non-void functions.
+- The `@brief/@param/@return` description must accurately reflect the current functionality of the function.
+
+## General guidelines
+
+- The code must reuse `cuda/` or `cuda/std` functionalities as much as possible, including macros.
+- Try to use modern C++ as much as possible. The repository supports C++17 but many more recent functionalities have been backported with functions and macros.
+- Never allow lambda expression, in device or host-device code.
+- Protect host code with `#if !_CCCL_COMPILER(NVRTC)`.
```
miscco:
```suggestion
- Protect host only code with `#if !_CCCL_COMPILER(NVRTC)`.
```

.agent/skills/licudacxx-style/SKILL.md
```diff
@@ -0,0 +1,86 @@
...
+- `constexpr` must be used for all functions that don't depend on run-time features, e.g. pointers.
+- Return type must be explicit (non `auto`) when possible.
+
+Function call:
+
+- All functions must have fully qualified namespace starting from the global namespace, e.g. `::cuda::ceil_div`. This includes calls to functions defined in the same namespace, e.g. inside `cuda::`, call `::cuda::ceil_div(...)`, not `ceil_div(...)`.
+
+## Types
+
+- Types must have fully qualified namespace, except when they are already declared in the current namespace.
```
bernhardmgruber:
```suggestion
- Type names must be fully qualified, except when they are already declared in the current namespace.
```

.agent/skills/licudacxx-style/SKILL.md
```diff
@@ -0,0 +1,86 @@
...
+- This includes standard integer type aliases (`::cuda::std::size_t`, `::cuda::std::uintptr_t`, `::cuda::std::int32_t`, etc.) and any other `cuda::std` or standard library types. A local `using` declaration (e.g. `using ::cuda::std::size_t;`) is acceptable to avoid repetition within a function body.
+
+## Headers
+
+- All header inclusions must use the syntax `<header>`.
+- Files must include all headers related to the symbols that they are using.
+- No transitive header inclusion are allowed.
```
bernhardmgruber, line 51: Except for the config headers.

bernhardmgruber: I see you list that below, so we can ignore it here.

.agent/skills/licudacxx-style/SKILL.md
```diff
@@ -0,0 +1,86 @@
...
+#  pragma system_header
+#endif // no system header
+```
+- The last included header must be `#include <cuda/std/__cccl/prologue.h>` before the code, and `#include <cuda/std/__cccl/epilogue.h>` at the end of a file.
+
+## Comments
+
+- Commented code without a description is not allowed.
+- Use Doxygen-style `//! @brief comments`.
+- All functions, except `_impl` functions, must have complete Doxygen documentation: `//! @brief`, `//! @param[in/out/in,out]` for every parameter, and `//! @return` for non-void functions.
```
bernhardmgruber: I would also allow to skip documentation for any internal function, that is, a function starting with `__`.

.agent/skills/licudacxx-style/SKILL.md
```diff
@@ -0,0 +1,86 @@
...
+- The `@brief/@param/@return` description must accurately reflect the current functionality of the function.
+
+## General guidelines
+
+- The code must reuse `cuda/` or `cuda/std` functionalities as much as possible, including macros.
+- Try to use modern C++ as much as possible. The repository supports C++17 but many more recent functionalities have been backported with functions and macros.
+- Never allow lambda expression, in device or host-device code.
```
bernhardmgruber: Why? They work just fine and we use them in several places in CUB.

.agent/skills/licudacxx-style/SKILL.md
fbusato: definitely right. Let's see how LLMs digest it. The definition is not very precise

.agent/skills/licudacxx-style/SKILL.md
fbusato: good point.

.agent/skills/licudacxx-style/SKILL.md
fbusato: lambda expressions are not compatible with NVRTC

.agent/skills/licudacxx-style/SKILL.md
fbusato: why? I don't think adding documentation to internal functions is harmful. I find it extremely useful when I'm not familiar with the implementation details, preconditions, or parameters.

bernhardmgruber: I am in the camp that believes that if a function needs extensive documentation, it's interface is not well designed.  I am fine with adding a comment here and there, but requiring a full Doxygen block is too much IMO.

.agent/skills/licudacxx-style/SKILL.md
bernhardmgruber: They work fine in CUB and CCCL.C with `-default-device`.

.agent/skills/licudacxx-style/SKILL.md
fbusato: I had the same idea, but @davebayer made me change my mind.

.agent/skills/licudacxx-style/SKILL.md
fbusato: this requires an addition flag, I'm not sure we can only enforce that

.agent/skills/licudacxx-style/SKILL.md
```diff
@@ -20,7 +20,7 @@ All non-public symbols must be C++ reserved identifiers:
 
 ## Variables
 
-- All variables that are not modified must use `const`. This includes variables initialized by casts (`static_cast`, `reinterpret_cast`, `bit_cast`), function return values, and loop-invariant computations.
+- All variables that are not modified must use `const`. This includes variables initialized by casts (`static_cast`, `reinterpret_cast`, `bit_cast`), function return values, and loop-invariant computations. The only exception is function parameters.
```
miscco: Honestly, I would be really happy if we would start being const correct in function arguments

fbusato: I derived the rule looking at the code. 100% agree. The only minor concern is what we report in the documentation. I know that other code guidelines don't allow it to avoid confusion with overloads (`const` params don't have anything to do with overloads).

.claude/skills/licudacxx-style/SKILL.md
bernhardmgruber, line 1: Q: Do we really need to duplicate the content or can we just tell Claude to look into the other file? Or use a symlink?

fbusato: I really would like to avoid duplication. My understanding is that symlink doesn't work on all platforms.
@alliepiper any suggestion?

.agent/skills/licudacxx-style/SKILL.md
miscco: Yes we have a ton of tiny small helpers, so we definitely do not want to bloat our code with senseless documentation

.claude/skills/licudacxx-style/SKILL.md
miscco: AFAIK you can just point an agent to another file to read that and follow the instructions there

.agent/skills/licudacxx-style/SKILL.md
alliepiper: We have historically supported compilation without the experimental flag enabled. Changing this would be a significant change that requires discussion -- def out of scope here :)

.claude/skills/licudacxx-style/SKILL.md
alliepiper: You can tell it to read another file. I use the same approach across my projects -- main instructions in the generic AGENTS.md / .agents/skills/ dir, and then CLAUDE.md that just says "scan for all AGENTS.md files and use them as instructions. This repo keeps agent skills under .skills/agent/.

fbusato:
@Jacobfaib could you please take a look? I'm pretty sure there is a lot of overlapping. I would much prefer static-analysis over agent actions for accuracy and efficiency reasons.

Jacobfaib:
> I'm pretty sure there is a lot of overlapping. I would much prefer static-analysis over agent actions for accuracy and efficiency reasons.

Not sure how good LLM's have gotten at reading YAML, but perhaps you could instruct the agents to simply read `.clang-tidy`, `.clang-format` and infer most style-guide related preferences from there? `.clang-format` at the very least should be very instructive as it doesn't have a lot of implicit options.

`.clang-tidy` on the other hand has more of a "it's enabled unless mentioned" so LLM's might not pick up on these as easily.

fbusato:
> Not sure how good LLM's have gotten at reading YAML, but perhaps you could instruct the agents to simply read .clang-tidy, .clang-format and infer most style-guide related preferences from there? .clang-format at the very least should be very instructive as it doesn't have a lot of implicit options.

If we already have the rule in `.clang-tidy` or `.clang-format` make little sense to add to `SKILL.md`. The idea here is to cover all the aspects that cannot be enforced with clang tools

PR #8020, opened 2026-03-13
Add env ReduceByKey and TransformReduce

### LINKED ISSUE #7546: Add env-based API for cub::DeviceReduce
- [x] [Add env-based cub::DeviceReduce::Reduce()](https://github.com/NVIDIA/cccl/pull/4877)
- [x] [Add env-based cub::DeviceReduce::Sum()](https://github.com/NVIDIA/cccl/issues/4960)
- [x] [Add env-based cub::DeviceReduce::Min/Max/ArgMin/ArgMax](https://github.com/NVIDIA/cccl/issues/5121)
- [ ] Add env-based cub::DeviceReduce::ReduceByKey()
- [ ] Add env-based cub::DeviceReduce::TransformReduce()

gonidelis:
closes #7546 by adding the last two remaining overloads:

- `ReduceByKey`
- `TransformReduce`

For `DeviceReduce::TransformReduce` we maintain the deterministic guarantees (full determinism range) of `DeviceReduce::Reduce` as both go through RFA. The `transform`ation part happens before the reduction so the deterministic characteristics remain.

`DeviceReduce::ReduceByKey`provides `run_to_run` determinism by construction. `gpu_to_gpu` determinism is not supported because the algorithm is based on a reduce-by-segment scan operating on `(value, flag)` pairs with inner decoupled lookback. Different tile sizes across architectures change where partial aggregates are combined relative to segment boundaries, making architecture-independent results infeasible.

### DIFF
--- a/cub/test/catch2_test_device_reduce_env_api.cu
+++ b/cub/test/catch2_test_device_reduce_env_api.cu
@@ -511,17 +511,17 @@ C2H_TEST("cub::DeviceReduce::TransformReduce accepts stream", "[reduce][env]")
 C2H_TEST("cub::DeviceReduce::ReduceByKey accepts run_to_run determinism requirements", "[reduce][env]")
 {
   // example-begin reduce-by-key-env
-  auto keys_in        = thrust::device_vector<int>{0, 2, 2, 9, 5, 5, 5, 8};
-  auto values_in      = thrust::device_vector<int>{0, 7, 1, 6, 2, 5, 3, 4};
-  auto unique_out     = thrust::device_vector<int>(8);
-  auto aggregates_out = thrust::device_vector<int>(8);
-  auto num_runs_out   = thrust::device_vector<int>(1);
+  auto keys_in         = thrust::device_vector<int>{0, 2, 2, 9, 5, 5, 5, 8};
+  auto values_in       = thrust::device_vector<int>{0, 7, 1, 6, 2, 5, 3, 4};
+  auto unique_keys_out = thrust::device_vector<int>(5);
+  auto aggregates_out  = thrust::device_vector<int>(5);
+  auto num_runs_out    = thrust::device_vector<int>(1);
 
   auto env = cuda::execution::require(cuda::execution::determinism::run_to_run);
 
   auto error = cub::DeviceReduce::ReduceByKey(
     keys_in.begin(),
-    unique_out.begin(),
+    unique_keys_out.begin(),
     values_in.begin(),
     aggregates_out.begin(),
     num_runs_out.begin(),
@@ -539,26 +539,26 @@ C2H_TEST("cub::DeviceReduce::ReduceByKey accepts run_to_run determinism requirem
 
   REQUIRE(error == cudaSuccess);
   REQUIRE(num_runs_out[0] == 5);
-  unique_out.resize(5);
+  unique_keys_out.resize(5);
   aggregates_out.resize(5);
-  REQUIRE(unique_out == expected_keys);
+  REQUIRE(unique_keys_out == expected_keys);
   REQUIRE(aggregates_out == expected_aggregates);
 }
 
 C2H_TEST("cub::DeviceReduce::ReduceByKey accepts not_guaranteed determinism requirements", "[reduce][env]")
 {
   // example-begin reduce-by-key-env-non-determinism
-  auto keys_in        = thrust::device_vector<int>{0, 2, 2, 9, 5, 5, 5, 8};
-  auto values_in      = thrust::device_vector<int>{0, 7, 1, 6, 2, 5, 3, 4};
-  auto unique_out     = thrust::device_vector<int>(8);
-  auto aggregates_out = thrust::device_vector<int>(8);
-  auto num_runs_out   = thrust::device_vector<int>(1);
+  auto keys_in         = thrust::device_vector<int>{0, 2, 2, 9, 5, 5, 5, 8};
+  auto values_in       = thrust::device_vector<int>{0, 7, 1, 6, 2, 5, 3, 4};
+  auto unique_keys_out = thrust::device_vector<int>(5);
+  auto aggregates_out  = thrust::device_vector<int>(5);
+  auto num_runs_out    = thrust::device_vector<int>(1);
 
   auto env = cuda::execution::require(cuda::execution::determinism::not_guaranteed);
 
   auto error = cub::DeviceReduce::ReduceByKey(
     keys_in.begin(),
-    unique_out.begin(),
+    unique_keys_out.begin(),
     values_in.begin(),
     aggregates_out.begin(),
     num_runs_out.begin(),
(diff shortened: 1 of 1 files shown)


cub/cub/device/device_reduce.cuh
```diff
@@ -2114,6 +2114,287 @@ public:
...
+    EnvT env = {})
+  {
+    _CCCL_NVTX_RANGE_SCOPE("cub::DeviceReduce::TransformReduce");
+
+    static_assert(!::cuda::std::execution::__queryable_with<EnvT, ::cuda::execution::determinism::__get_determinism_t>,
+                  "Determinism should be used inside requires to have an effect.");
+    using requirements_t = ::cuda::std::execution::
+      __query_result_or_t<EnvT, ::cuda::execution::__get_requirements_t, ::cuda::std::execution::env<>>;
+    using default_determinism_t =
+      ::cuda::std::execution::__query_result_or_t<requirements_t, //
```
bernhardmgruber: Important: The implementation of `TransformReduce` is almost an identical copy of the env overload of `Reduce`. Please move this implementation into an internal function, like `__transform_reduce` and call it here. Then remove the implementation in `Reduce` and just call `__transform_reduce` with `::cuda::std::identity{}` as transform op.

cub/test/catch2_test_device_reduce_env_api.cu
```diff
@@ -351,3 +351,126 @@ C2H_TEST("cub::DeviceReduce::ArgMax accepts stream", "[reduce][env]")
...
+{
+  // example-begin transform-reduce-env-stream
+  auto op        = cuda::std::plus{};
+  auto transform = cuda::std::negate{};
+  auto input     = thrust::device_vector<float>{1.0f, 2.0f, 3.0f, 4.0f};
+  auto output    = thrust::device_vector<float>(1);
+  auto init      = 0.0f;
+
+  cudaStream_t legacy_stream = 0;
+  cuda::stream_ref stream_ref{legacy_stream};
```
bernhardmgruber: Important: Please create and pass a `cuda::stream` here. Let's not show the use of `cudaStream_t`

Applies to more examples.

cub/cub/device/device_reduce.cuh
```diff
@@ -235,6 +235,94 @@ private:
...
+            typename AccumT,
+            typename EnvT>
+  [[nodiscard]] CUB_RUNTIME_FUNCTION static cudaError_t __transform_reduce(
+    InputIteratorT d_in,
+    OutputIteratorT d_out,
+    NumItemsT num_items,
+    ReductionOpT reduction_op,
+    TransformOpT transform_op,
+    T init,
+    EnvT env)
```
bernhardmgruber: Strong suggestion: A good guideline for template parameter lists is to put the parameters that cannot be deduced from a function argument first. Here, `AccumT` cannot be deduced from any argument. If you put it first, you only need to specify a template argument for `AccumT` and let the compiler deduce the rest. This turns this:
```c++
    return __transform_reduce<InputIteratorT,
                              OutputIteratorT,
                              ReductionOpT,
                              ::cuda::std::identity,
                              T,
                              NumItemsT,
                              accum_t,
                              EnvT>(d_in, d_out, num_items, reduction_op, ::cuda::std::identity{}, init, env);
```
into this:
```c++
    return __transform_reduce<accum_t>(d_in, d_out, num_items, reduction_op, ::cuda::std::identity{}, init, env);
```

cub/test/catch2_test_device_reduce_env_api.cu
```diff
@@ -67,10 +69,11 @@ C2H_TEST("cub::DeviceReduce::Reduce accepts stream", "[reduce][env]")
...
   auto init   = 0.0f;
 
-  cudaStream_t legacy_stream = 0;
-  cuda::stream_ref stream_ref{legacy_stream};
+  cuda::stream stream{cuda::devices[0]};
+  cuda::stream_ref stream_ref{stream};
+  auto env = cuda::std::execution::env{stream_ref};
 
-  auto error = cub::DeviceReduce::Reduce(input.begin(), output.begin(), input.size(), op, init, stream_ref);
+  auto error = cub::DeviceReduce::Reduce(input.begin(), output.begin(), input.size(), op, init, env);
```
bernhardmgruber: Suggestion: No need to warp a `stream_ref` in an environment. A `stream_ref` *is* an environment already, you can pass it directly.

cub/test/catch2_test_device_reduce_env_api.cu
```diff
@@ -351,3 +353,126 @@ C2H_TEST("cub::DeviceReduce::ArgMax accepts stream", "[reduce][env]")
...
+  REQUIRE(output == expected);
+}
+
+C2H_TEST("cub::DeviceReduce::ReduceByKey accepts env", "[reduce][env]")
+{
+  // example-begin reduce-by-key-env
+  auto keys_in        = thrust::device_vector<int>{0, 2, 2, 9, 5, 5, 5, 8};
+  auto values_in      = thrust::device_vector<int>{0, 7, 1, 6, 2, 5, 3, 4};
+  auto unique_out     = thrust::device_vector<int>(8);
+  auto aggregates_out = thrust::device_vector<int>(8);
```
bernhardmgruber: Suggestion: I think we can just use the expected group count directly:
```suggestion
  auto unique_keys_out     = thrust::device_vector<int>(5);
  auto aggregates_out = thrust::device_vector<int>(5);
```
Plus a tiny rename for readability.

cub/cub/device/device_reduce.cuh
```diff
@@ -2114,6 +2136,233 @@ public:
...
+    OutputIteratorT d_out,
+    NumItemsT num_items,
+    ReductionOpT reduction_op,
+    TransformOpT transform_op,
+    T init,
+    EnvT env = {})
+  {
+    _CCCL_NVTX_RANGE_SCOPE("cub::DeviceReduce::TransformReduce");
+    using accum_t = ::cuda::std::
+      __accumulator_t<ReductionOpT, ::cuda::std::invoke_result_t<TransformOpT, detail::it_value_t<InputIteratorT>>, T>;
```
miscco, line 2234: Comment: Technically this should be `iter_reference_t` That said, our device_reference does crazy stuff and the reference should collapse to a value, so this should be fine

bernhardmgruber: Probably. But we use `it_value_t` everywhere I think. Or at least I have seen it being used a lot. I think the idea is to not pass a proxy reference to the reduction operator.

cub/test/catch2_test_device_reduce_env_api.cu
```diff
@@ -79,31 +79,14 @@ C2H_TEST("cub::DeviceReduce::Reduce accepts stream", "[reduce][env]")
   }
 
   thrust::device_vector<float> expected{6.0f};
+  stream.sync();
```
bernhardmgruber, line 82: Why do we need this sync here? AFAIK we don't have this in other API examples as well.

Appears a few more times below.

bernhardmgruber: @gonidelis please address this!

PR #8021, opened 2026-03-13
[cudax] Implement replicate and join for path_builder

pciolkosz:
PR https://github.com/NVIDIA/cccl/pull/8014 implements `replicate` and `join` for streams, this one does that for `path_builder`. The semantics is mostly the same as for streams, `replicate` creates a requested number of path builders in the same graph, `replicate_prepend` does the same but also add the input one to the output. `join` gathers all dependencies from one set of path builders and adds them to another set of path builders.
One observation was that we don't try to avoid duplicate dependencies in graph APIs and we possibly should take a look into that. `join` probably could use some duplicate dependency control, because the underlying driver API won't accept them. But that is a problem for another PR

### DIFF
--- a/cudax/include/cuda/experimental/__graph/path_builder.cuh
+++ b/cudax/include/cuda/experimental/__graph/path_builder.cuh
@@ -16,10 +16,8 @@
 #include <cuda/__runtime/api_wrapper.h>
 #include <cuda/std/__exception/cuda_error.h>
 #include <cuda/std/__exception/exception_macros.h>
-#include <cuda/std/__ranges/concepts.h>
 #include <cuda/std/span>
 #include <cuda/std/type_traits>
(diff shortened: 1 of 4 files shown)


cudax/include/cuda/experimental/__graph/path_builder.cuh
```diff
@@ -172,6 +176,14 @@ struct path_builder
       ...);
   }
 
+  //! \brief Add a range of dependency nodes to this path builder.
+  //! \param __deps The dependency node range to add.
+  template <size_t _Extent>
+  _CCCL_HOST_API void depends_on(::cuda::std::span<const cudaGraphNode_t, _Extent> __deps)
```
davebayer: Shouldn't this take a range instead of `span`?

PR #8026, opened 2026-03-13
Optimize `cuda::std::saturating_(add|sub)` for `120f+`

davebayer:
PTX ISA 9.2 introduced some `.sat` more variants of `add` and `sub` PTX instructions for 120-family architectures. We can use them in `cuda::std::saturating_add` and `cuda::std::saturating_sub`

### DIFF
--- a/libcudacxx/include/cuda/std/__numeric/saturating_add.h
+++ b/libcudacxx/include/cuda/std/__numeric/saturating_add.h
@@ -118,8 +118,32 @@ template <class _Tp>
 {
   if constexpr (is_signed_v<_Tp>)
   {
-    if constexpr (sizeof(_Tp) < sizeof(int32_t))
+    if constexpr (sizeof(_Tp) == sizeof(int8_t))
     {
+#  if __cccl_ptx_isa >= 920
+      NV_IF_TARGET(NV_HAS_FEATURE_SM_120f, ({
+                     // Use uint32_t because we want to avoid sign extension.
+                     uint32_t __result;
+                     asm("add.sat.s8x4 %0, %1, %2;"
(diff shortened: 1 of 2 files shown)


libcudacxx/include/cuda/std/__numeric/saturating_add.h
```diff
@@ -135,7 +153,43 @@ template <class _Tp>
   }
   else
   {
-    return ::cuda::add_sat_overflow(__x, __y).value;
+    if constexpr (sizeof(_Tp) == sizeof(uint8_t))
+    {
+#  if __cccl_ptx_isa >= 920
+      NV_IF_TARGET(NV_HAS_FEATURE_SM_120f, ({
+                     uint32_t __result;
+                     asm("add.sat.u8x4 %0, %1, %2;" : "=r"(__result) : "r"(uint32_t{__x}), "r"(uint32_t{__y}));
```
fbusato, line 167: I'm a bit puzzle about using packed saturation instructions. This approach could be slower than the standard one because it requires "unpacking" at the beginning and at the end

davebayer: These architectures provide dedicated sass instructions to implement these, are you concerned about the performance anyway? We should implement some benchmark to see what's the reality in the end

fbusato: I don't think creating benchmarks for a couple of SASS instructions make little sense. We know everything, latency and throughput.
My point is that such instructions have been introduce for packed data, not single value. This could introduce overhead instead of providing benefits. We just need to check the comparison with the current general approach, just it.

PR #8027, opened 2026-03-13
Fix family arch specific feature detection in `<nv/target>`

davebayer:
I believe the way we right now enable family arch specific features in `<nv/target>` is wrong. If we have a `100f` architecture, it supports only `100f` features. But if we have `103f` architecture, it supports `100f` and `103f`.

Currently, we only allow `103f` features and this PR fixes that.

### DIFF
--- a/libcudacxx/include/nv/detail/__target_macros
+++ b/libcudacxx/include/nv/detail/__target_macros
@@ -454,7 +454,7 @@
 // architecture-specific SM versions
 
 #  if defined(__CUDA_ARCH_SPECIFIC__)
-#    define _NV_TARGET_CHECK_ARCH_SPECIFIC(_MAJOR, _MINOR) (__CUDA_ARCH_SPECIFIC__ == _MAJOR##_MINOR##0)
+#    define _NV_TARGET_CHECK_ARCH_SPECIFIC(_MAJOR, _MINOR) (__CUDA_ARCH_SPECIFIC__ == (_MAJOR * 100 + _MINOR * 10))
 #  else
 #    define _NV_TARGET_CHECK_ARCH_SPECIFIC(_MAJOR, _MINOR) (0)
 #  endif


bernhardmgruber:
> if we have `103f` architecture, it supports `100f` and `103f`.

Are you sure? Reading this blog post: https://developer.nvidia.com/blog/nvidia-blackwell-and-nvidia-cuda-12-9-introduce-family-specific-architecture-features/#family-specific_feature_set_introduced_in_nvidia_blackwell:

> All devices within a family share the same major compute capability version. Family-specific features are guaranteed to be available in the same family of GPUs, which includes later GPUs of the same major compute capability and higher minor compute capability.

Especially the last clause indicates to me that `103f` means `sm_103` and anything later that still starts with `sm10x.` And not `sm_100`

davebayer:
> Are you sure?

I believe this is correct. This is how the internal `__CUDA_HAS_ARCH_FAMILY_SPECIFIC` macro is implemented.
```cpp
#define __CUDA_HAS_ARCH_FAMILY_SPECIFIC(N) ( (((N) == 101) && (__CUDA_ARCH_FAMILY_SPECIFIC__ == 1010)) || (((N) != 101) && (__CUDA_ARCH_FAMILY_SPECIFIC__ >= (N ##0) && __CUDA_ARCH_FAMILY_SPECIFIC__ < ((N ##0) + 100))))
```
There is some special treatment for `101` (`110`) architecture and then the macro checks if the architecture is greater or equal to the `N` an if it belongs to the same major compute capability (I think the implementation is wrong).

Also, many PTX ISA instruction descriptions talk about `sm_100f` or higher in the same family for example. If it was the other way, this note wouldn't make sense

davebayer:
> This is wrong. see https://docs.nvidia.com/cuda/cuda-programming-guide/05-appendices/compute-capabilities.html#compute-capabilities-family-specific-compatibility and the macro definition. We cannot say `103f ` is a superset of `100f`

As I understand it, it is. See a [demo using `add.sat.u8x4` instruction](https://ce.nvidia.com/z/Mxcvb7) which is available for `sm_120f` and [demo using `tcgen05.fence::before_thread_sync` instruction](https://ce.nvidia.com/z/n9KW1E) which is available for `sm_100f` and `sm_110f`.

fbusato:
This definitely true for [add.sat](https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#integer-arithmetic-instructions-add), but I would not generalize it

> sm_120f or higher in the same family

fbusato:
let me rephrase in a more formal way, otherwise the term "superset" is ambiguous:

- anything that works on `sm_100f` is expected to work on `sm_103f` or higher in the same family.
- the opposite is not true, `sm_103f` instructions are not supported on `sm_100f` in all cases.

davebayer:
> - the opposite is not true, sm_103f instructions are not supported on sm_100f in all cases.

Yes and that's what the implementation is doing (or at least it should):
```cpp
(__CUDA_ARCH_FAMILY_SPECIFIC__ >= _MAJOR##_MINOR##0 && __CUDA_ARCH_FAMILY_SPECIFIC__ < (_MAJOR + 1) * 100)
```

libcudacxx/include/nv/detail/__target_macros
```diff
@@ -510,10 +516,17 @@
 //----------------------------------------------------------------------------------------------------------------------
 // family-specific SM versions
 
+#  if defined(__CUDA_ARCH_FAMILY_SPECIFIC__)
+#    define _NV_TARGET_CHECK_FAMILY_SPECIFIC(_MAJOR, _MINOR) \
+      (__CUDA_ARCH_FAMILY_SPECIFIC__ >= _MAJOR##_MINOR##0 && __CUDA_ARCH_FAMILY_SPECIFIC__ < (_MAJOR + 1) * 100)
```
fbusato: suggestion. Use arithmetic operation instead of `_MAJOR##_MINOR##0`

libcudacxx/include/nv/detail/__target_macros
```diff
@@ -453,10 +453,16 @@
 //----------------------------------------------------------------------------------------------------------------------
 // architecture-specific SM versions
 
+#  if defined(__CUDA_ARCH_SPECIFIC__)
+#    define _NV_TARGET_CHECK_ARCH_SPECIFIC(_MAJOR, _MINOR) (__CUDA_ARCH_SPECIFIC__ == _MAJOR##_MINOR##0)
```
fbusato: suggestion, use arithmetic operations also here

PR #8029, opened 2026-03-13
Deprecate `cuda::std::rel_ops` in c++20

davebayer:
I accidently came across this

### DIFF
--- a/cudax/include/cuda/experimental/__execution/completion_behavior.cuh
+++ b/cudax/include/cuda/experimental/__execution/completion_behavior.cuh
@@ -50,24 +50,6 @@ enum class _CCCL_TYPE_VISIBILITY_DEFAULT completion_behavior : int
                     ///< `start()`.
 };
 
-#if _LIBCUDACXX_HAS_SPACESHIP_OPERATOR()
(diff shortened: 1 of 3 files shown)


cudax/include/cuda/experimental/__execution/completion_behavior.cuh
```diff
@@ -50,24 +50,6 @@ enum class _CCCL_TYPE_VISIBILITY_DEFAULT completion_behavior : int
...
-[[nodiscard]] _CCCL_API constexpr auto operator<(completion_behavior __a, completion_behavior __b) noexcept -> bool
-{
-  return static_cast<int>(__a) < static_cast<int>(__b);
-}
-[[nodiscard]] _CCCL_API constexpr auto operator==(completion_behavior __a, completion_behavior __b) noexcept -> bool
-{
-  return static_cast<int>(__a) == static_cast<int>(__b);
-}
-using namespace ::cuda::std::rel_ops;
-#endif
```
davebayer, line 69: As we found out with @bernhardmgruber, `enum class` is comparable by default

PR #8031, opened 2026-03-13
Minimize includes in `thrust/iterator_traits.h`

miscco:
We are currently including all of `<cuda/iterator>` which includes all of `<cuda/std/iterator>`

We can considerable reduce that by adding a header with forward declarations of the respective iterators we need.


thrust/thrust/iterator/iterator_traits.h
```diff
@@ -33,7 +33,9 @@
 #include <thrust/iterator/detail/minimum_system.h>
 #include <thrust/iterator/iterator_categories.h>
 
-#include <cuda/iterator>
+#include <cuda/__fwd/iterator.h>
+#include <cuda/std/__fwd/iterator.h>
+#include <cuda/std/__iterator/iterator_traits.h>
```
bernhardmgruber, line 38: This is the golden change, right?

miscco: yeah

miscco: the amount of transitive includes this found

PR #8032, opened 2026-03-13
Add `.sync_aligned()` for this groups

davebayer:
`.sync_aligned()` is an optimized version of `.sync()` that requires all threads that belong to the group to sync in the same textual context (calls `__syncthreads()` instead of `__barrier_sync(0)` for example).

Aligned synchronization can significantly decrease the amount of generated SASS. See the difference:
1. block synchronization: https://godbolt.org/z/P37xdEhf9
2. cluster synchronization: https://godbolt.org/z/MhsMcrq7W


cudax/include/cuda/experimental/__hierarchy/this_group.cuh
```diff
@@ -42,6 +42,109 @@ template <class _HierarchyLike>
...
+
+  // Synchronize with other blocks using the thread 0 in block.
+  if ((threadIdx.x | threadIdx.y | threadIdx.z) == 0)
+  {
+    const unsigned __expected = gridDim.x * gridDim.y * gridDim.z;
+    unsigned __nblocks        = 1;
+
+    if ((blockIdx.x | blockIdx.y | blockIdx.z) == 0)
+    {
+      __nblocks = unsigned{::cuda::std::numeric_limits<int>::min()} - (__expected - 1);
```
pciolkosz: I think the value here is incorrect, the overflow here will just result in 0xFFFFFFFF, not in 0x80000000, so some of the blocks will have that bit set in the old_barrier_value and never leave the barrier

davebayer: No, the problem was that I used the wrong barrier pointer 🙈

PR #8035, opened 2026-03-14
[STF] Refactor exec_place to unified grid model

andralex:
## Summary

This PR simplifies the `exec_place` hierarchy by removing unnecessary shell classes and introducing a cleaner RAII-based activation mechanism.

### Key Changes

- **Introduced `exec_place_scope`**: A movable RAII guard that manages execution context activation/deactivation. It holds the iteration state (current index) internally, eliminating the need for mutable state in `exec_place::impl`.

- **Removed `exec_place_host`**: The shell class is gone; `exec_place::host()` now returns a plain `exec_place`. This removes static type discrimination in favor of runtime `is_host()` checks where needed.

- **Removed `exec_place::iterator`**: Replaced range-based iteration with explicit index loops (`for (size_t i = 0; i < place.size(); ++i)`), which is clearer and avoids iterator complexity.

- **Simplified internal state**: `exec_place_scope` uses an empty `exec_place` (null pimpl) to represent inactive state instead of a separate `bool active_` flag. Removed `std::optional` wrapper from `task.cuh`.

- **Cleanup**: Removed dead code paths in `place_partition.cuh`, redundant operators in `exec_place_scope`, and applied pass-by-value + `mv()` idiom for sink parameters.

### Suggested Review Order

1. **`places.cuh`** — Core changes: `exec_place_scope` definition, removal of `exec_place_host`, simplified `exec_place` API
2. **`place_partition.cuh`** — Updated to use new API, dead code removal, parameter passing cleanup
3. **`launch.cuh`** — Updated iteration pattern (index-based loops)
4. **`task.cuh` / `acquire_release.cuh`** — Simplified `exec_place_scope` usage (no more `std::optional`)
5. **`parallel_for_scope.cuh` / `backend_ctx.cuh`** — Removed `exec_place_host` dependencies

### DIFF
--- a/cudax/include/cuda/experimental/__stf/places/places.cuh
+++ b/cudax/include/cuda/experimental/__stf/places/places.cuh
@@ -1323,7 +1323,8 @@ inline exec_place make_grid(::std::vector<exec_place> places, const dim4& dims)
 inline exec_place make_grid(::std::vector<exec_place> places)
 {
   _CCCL_ASSERT(!places.empty(), "invalid places");
-  return make_grid(mv(places), dim4(places.size(), 1, 1, 1));
+  const size_t n = places.size();
+  return make_grid(mv(places), dim4(n, 1, 1, 1));
 }
 
 // === data_place::affine_exec_place implementation ===


cudax/include/cuda/experimental/__stf/internal/launch.cuh
```diff
@@ -95,7 +95,7 @@ void cuda_launcher_graph(interpreted_spec interpreted_policy, Fun&& f, void** ar
 template <typename Fun, typename interpreted_spec, typename Arg>
 void launch_impl(interpreted_spec interpreted_policy, exec_place& p, Fun f, Arg arg, cudaStream_t stream, size_t rank)
 {
-  assert(!p.is_grid());
+  assert(p.size() == 1);
```
caugonnet: all these assertions should have been _CCCL_ASSERT in the first place

cudax/include/cuda/experimental/__stf/places/places.cuh
```diff
@@ -377,18 +367,19 @@ inline data_place from_index(size_t n);
...
  */
 class exec_place
 {
 public:
   /*
-   * @brief Using the pimpl idiom. Public because a number of classes inehrit from this.
+   * @brief Using the pimpl idiom. Public because a number of classes inherit from this.
    */
-  class impl
+  class impl : public ::std::enable_shared_from_this<impl>
```
caugonnet, line 382: i avoided this in data_place but maybe it's impossible to ignore here

cudax/include/cuda/experimental/__stf/places/places.cuh
```diff
@@ -507,25 +506,28 @@ public:
 
   exec_place() = default;
   exec_place(const data_place& affine)
-      : pimpl(affine.is_device() ? device(device_ordinal(affine)).pimpl : ::std::make_shared<impl>(affine))
+      : pimpl(affine.is_device() ? device(device_ordinal(affine)).pimpl : device_auto().pimpl)
```
caugonnet: device_auto() looks very surprising

cudax/include/cuda/experimental/__stf/places/places.cuh
```diff
@@ -507,25 +506,28 @@ public:
...
   {
-    _CCCL_ASSERT(pimpl->affine != data_place::host(),
-                 "To create an execution place for the host, use exec_place::host().");
+    _CCCL_ASSERT(!affine.is_host(), "To create an execution place for the host, use exec_place::host().");
```
caugonnet: (preexisting) this is a weird situation

cudax/include/cuda/experimental/__stf/places/places.cuh
```diff
@@ -598,20 +634,94 @@ public:
...
   }
 
   /**
-   * @brief Returns the `data_place` naturally associated with this execution place.
+   * @brief Set the current place for grid iteration
+   *
+   * Activates the place at the given index and saves state for later restoration.
    */
-  const data_place affine_data_place() const
+  void set_current_place(size_t idx)
```
caugonnet: should we have a grid iterator object in the exec_place::impl and leverage that instead ? If this is just a counter a new class is maybe unnecessary

cudax/include/cuda/experimental/__stf/places/places.cuh
```diff
@@ -626,70 +736,55 @@ public:
...
-   */
-  void deactivate(const exec_place& p) const
-  {
-    pimpl->deactivate(p);
-  }
-
   bool is_host() const
   {
-    return pimpl->is_host();
+    return affine_data_place().is_host();
```
caugonnet: this is wrong, one could make a device exec place where the memory is "backed" by pinned (host) memory, can't we rely on the dynamic_cast mambo jambo with exec_place_host instead ?

cudax/include/cuda/experimental/__stf/places/places.cuh
```diff
@@ -711,8 +806,8 @@ public:
   static exec_place green_ctx(const green_ctx_view& gc_view, bool use_green_ctx_data_place = false);
 #endif // _CCCL_CTK_AT_LEAST(12, 4)
 
-  static exec_place_cuda_stream cuda_stream(cudaStream_t stream);
-  static exec_place_cuda_stream cuda_stream(const decorated_stream& dstream);
+  static exec_place cuda_stream(cudaStream_t stream);
```
caugonnet, line 692: cool, no need for forward declarations

cudax/include/cuda/experimental/__stf/places/places.cuh
```diff
@@ -927,9 +1033,41 @@ inline exec_place_host exec_place::host()
...
+  {
+    return "device_auto";
+  }
+};
+
 inline exec_place exec_place::device_auto()
 {
-  return exec_place(data_place::device_auto());
+  static exec_place_device_auto_impl instance;
+  return ::std::shared_ptr<exec_place::impl>(&instance, [](exec_place::impl*) {});
```
caugonnet: can't we use the make_static_* factory style you had in data_place ?

cudax/include/cuda/experimental/__stf/places/places.cuh
```diff
@@ -1022,343 +1201,158 @@ UNITTEST("exec_place copyable")
...
-      : exec_place(mv(p))
-  {}
-
-  exec_place_grid(::std::vector<exec_place> p, const dim4& d)
-      : exec_place(::std::make_shared<impl>(mv(p), d))
-  {}
+private:
+  dim4 dims_;
+  ::std::vector<exec_place> places_;
+  mutable ::std::ptrdiff_t current_idx_ = -1;
```
caugonnet: should that move to exec_place::impl as an iterator class ?

cudax/include/cuda/experimental/__stf/internal/launch.cuh
andralex: done

cudax/include/cuda/experimental/__stf/places/places.cuh
andralex: Due to possible overheads? It's not too bad I think.

cudax/include/cuda/experimental/__stf/places/places.cuh
andralex: I deleted that ctor.

cudax/include/cuda/experimental/__stf/places/places.cuh
andralex: Agreed.

cudax/include/cuda/experimental/__stf/places/places.cuh
andralex: I'll look into it.

cudax/include/cuda/experimental/__stf/places/places.cuh
andralex: I added virtual functions `is_host` and `is_device` to the interface

cudax/include/cuda/experimental/__stf/places/places.cuh
andralex: done

cudax/include/cuda/experimental/__stf/places/place_partition.cuh
```diff
@@ -284,24 +295,33 @@ private:
...
+        // Multi-element grid: extract all places
+        for (size_t i = 0; i < place.size(); ++i)
+        {
+          sub_places.push_back(place.get_place(i));
+        }
+      }
+      else if (place.is_device())
+      {
+        // Scalar device place
+        sub_places.push_back(place);
```
caugonnet: This is odd

cudax/include/cuda/experimental/__stf/graph/interfaces/slice.cuh
```diff
@@ -87,8 +87,8 @@ public:
       return;
     }
 
-    exec_place_grid grid = memory_node.get_grid();
-    size_t total_size    = this->shape.size();
+    exec_place grid   = memory_node.affine_exec_place();
```
caugonnet, line 90: to verify

cudax/include/cuda/experimental/__stf/internal/backend_ctx.cuh
```diff
@@ -1082,13 +1082,6 @@ public:
     return reserved::launch_scope<Engine, thread_hierarchy_spec_t, Deps...>(self(), mv(spec), mv(e_place), mv(deps)...);
   }
 
-  /* Using ctx.launch with a host place */
-  template <typename... Deps>
-  auto launch(exec_place_host, task_dep<Deps>... deps)
-  {
-    return reserved::host_launch_scope<Engine, true, Deps...>(self(), mv(deps)...);
```
caugonnet, line 1107: where is this implemented now ?

cudax/include/cuda/experimental/__stf/internal/logical_data.cuh
```diff
@@ -1483,7 +1483,8 @@ public:
 
       exec_place e_place_n = memory_node.affine_exec_place();
 
-      auto saved_place = e_place_n.activate();
+      // Activate the execution place - automatically restores when active goes out of scope
+      auto active = e_place_n.activate();
```
caugonnet, line 1487: no way to avoid the RAII style now ?

cudax/include/cuda/experimental/__stf/localization/composite_slice.cuh
```diff
@@ -480,7 +476,8 @@ public:
       dim4 data_dims)
   {
     EXPECT(place.is_composite());
-    return cache.get(place.get_grid(), mapper, ::std::forward<F>(delinearize), total_size, elem_size, data_dims);
```
caugonnet, line 483: get_grid() == affine_exec_place() ?

cudax/include/cuda/experimental/__stf/places/place_partition.cuh
```diff
@@ -131,14 +131,13 @@ public:
    * @param grid Input execution place grid to partition
    * @param scope Partitioning granularity
    */
-  place_partition(async_resources_handle& handle, const exec_place_grid& grid, place_partition_scope scope)
+  place_partition(async_resources_handle& handle, exec_place grid, place_partition_scope scope)
```
caugonnet: the const was abusive ?

cudax/include/cuda/experimental/__stf/places/places.cuh
```diff
@@ -505,27 +498,32 @@ public:
     mutable stream_pool pool_data;
   };
 
-  exec_place() = default;
-  exec_place(const data_place& affine)
-      : pimpl(affine.is_device() ? device(device_ordinal(affine)).pimpl : ::std::make_shared<impl>(affine))
+  template <typename T>
+  static ::std::shared_ptr<impl> make_static_instance()
```
caugonnet, line 502: are we using it for devices ?

cudax/include/cuda/experimental/__stf/internal/parallel_for_scope.cuh
```diff
@@ -550,8 +550,8 @@ public:
     // If there is a partitioner, we ensure there is a proper affine data place for this execution place
     if constexpr (!::std::is_same_v<partitioner_t, null_partition>)
     {
-      // This is only meaningful for grid of places
-      if (e_place.is_grid())
+      // Grids need a composite data place
+      if (e_place.size() > 1)
```
miscco, line 554: Nitpick: In the context of mdspan and other standard C++ multidimensional access, I would have expected `rank()` as the name of that function

cudax/include/cuda/experimental/__stf/places/exec/cuda_stream.cuh
```diff
@@ -30,92 +30,79 @@
...
-        , dstream(_dstream)
-        , dummy_pool(_dstream)
-    {}
-
-    /* We set the current device to be the device on which the CUDA stream was created */
-    exec_place activate() const override
-    {
-      return exec_place::device(dstream.dev_id).activate();
-    }
+    EXPECT(idx == 0, "Index out of bounds for cuda_stream exec_place");
```
miscco: Should that be an `_CCCL_ASSERT`? 

We also have `_CCCL_VERIFY` for always on checks

cudax/include/cuda/experimental/__stf/places/exec/cuda_stream.cuh
```diff
@@ -30,92 +30,79 @@
...
+    return shared_from_this();
```
miscco, line 47:
```suggestion
    return ::std::shared_from_this();
```

cudax/include/cuda/experimental/__stf/places/exec/cuda_stream.cuh
```diff
@@ -30,92 +30,79 @@
...
+  mutable stream_pool dummy_pool_;
 };
 
-inline exec_place_cuda_stream exec_place::cuda_stream(cudaStream_t stream)
+inline exec_place exec_place::cuda_stream(cudaStream_t stream)
 {
   int devid = get_device_from_stream(stream);
-  return exec_place_cuda_stream(decorated_stream(stream, get_stream_id(stream), devid));
+  return exec_place(
+    ::std::make_shared<exec_place_cuda_stream_impl>(decorated_stream(stream, get_stream_id(stream), devid)));
```
miscco: Nitpick We prefer list-initialization to better disambiguate function calls for the reader
```suggestion
  return exec_place{
    ::std::make_shared<exec_place_cuda_stream_impl>(decorated_stream(stream, get_stream_id(stream), devid))};
```

cudax/include/cuda/experimental/__stf/places/exec/cuda_stream.cuh
```diff
@@ -30,92 +30,79 @@
...
 }
 
-inline exec_place_cuda_stream exec_place::cuda_stream(const decorated_stream& dstream)
+inline exec_place exec_place::cuda_stream(const decorated_stream& dstream)
 {
-  return exec_place_cuda_stream(dstream);
+  return exec_place(::std::make_shared<exec_place_cuda_stream_impl>(dstream));
```
miscco:
```suggestion
  return exec_place{::std::make_shared<exec_place_cuda_stream_impl>(dstream)};
```

cudax/include/cuda/experimental/__stf/places/places.cuh
```diff
@@ -398,98 +391,98 @@ public:
...
-    {
-      return "exec(" + affine.to_string() + ")";
-    }
+    /**
+     * @brief Get the impl of the sub-place at the given linear index
+     *
+     * For scalar places, idx must be 0 and returns shared_from_this().
+     * For grids, returns the impl of the stored sub-place.
+     */
+    virtual ::std::shared_ptr<impl> get_place(size_t idx);
```
miscco, line 420: Should this be pure virtual?

cudax/include/cuda/experimental/__stf/internal/backend_ctx.cuh
caugonnet: It looks like ctx.launch(exec_place::host(), ...) is no longer possible

cudax/include/cuda/experimental/__stf/places/exec/cuda_stream.cuh
caugonnet: shared_from_this() looks like a member function, not a free function

cudax/include/cuda/experimental/__stf/places/places.cuh
```diff
@@ -868,71 +929,106 @@ inline decorated_stream exec_place::getStream(bool for_computation) const
...
-                  "exec_place_host cannot add state; it would be sliced away.");
+    return "host";
   }
 };
 
-inline exec_place_host exec_place::host()
+inline exec_place exec_place::host()
 {
-  return exec_place_host();
+  return exec_place(make_static_instance<exec_place_host_impl>());
```
caugonnet, line 998: Should not we have an exec_place_host type, and a make_static_instance for the exec_place, instead of the impl ?

cudax/include/cuda/experimental/__stf/internal/parallel_for_scope.cuh
caugonnet: just my 2cents, but don't you find rank() is ambiguous with the rank/size semantic we have in SPMD programs ? (where rank is the position in the grid, not the dim)

cudax/include/cuda/experimental/__stf/internal/parallel_for_scope.cuh
```diff
@@ -656,13 +656,12 @@ public:
     }
 
     // Device land. Must use the supplemental if constexpr below to avoid compilation errors.
-    if constexpr (!::std::is_same_v<exec_place_t, exec_place_host> && is_extended_host_device_lambda_closure_type
```
caugonnet, line 659: there is a chance this breaks parallel_for on the host

cudax/include/cuda/experimental/__stf/places/place_partition.cuh
andralex: I think the loop works in all cases because for scalar devices the loop iterates exactly once and `get_place(0)` returns `shared_from_this()`. So I fixed by just using the loop.

cudax/include/cuda/experimental/__stf/graph/interfaces/slice.cuh
andralex: Seems correct. For composite data places, `affine_exec_place()` returns the execution grid - it goes through `get_affine_exec_impl()` which returns the stored `grid_` from data_place_composite. Will resolve for now.

cudax/include/cuda/experimental/__stf/internal/backend_ctx.cuh
andralex: Indeed. As per our discussion today, @caugonnet and I decided to simplify by removing the dedicated statically-typed classes (`exec_place_host`, `exec_place_device`, `exec_place_grid`, etc.). The previous design had a dual hierarchy: polymorphic implementations with virtual functions, plus pimpl facades with their own inheritance, requiring lots of forwarding and types without enough utility.

Now we have a cleaner separation: `host_launch` for host callbacks (uses `cudaLaunchHostFunc`/`cudaGraphAddHostNode`) and `launch` for device kernels. These are fundamentally different CUDA operations, so having distinct APIs makes sense.

cudax/include/cuda/experimental/__stf/internal/logical_data.cuh
andralex: Unscoped activation is possible via `get_impl()->activate(idx)` and `get_impl()->deactivate(prev, idx)` - these are public. But RAII is the intended pattern because it's safer and exception-safe. It's good that unscoped activation is more verbose because it's more dangerous!

cudax/include/cuda/experimental/__stf/localization/composite_slice.cuh
andralex: Yah, for composite data places they're equivalent. `data_place_composite::get_affine_exec_impl()` returns `grid_.get_impl()`, and `affine_exec_place()` wraps that into an `exec_place`. So `affine_exec_place()` is the correct replacement for the removed `get_grid()`.

cudax/include/cuda/experimental/__stf/places/place_partition.cuh
andralex: thanks, got back to `const`

cudax/include/cuda/experimental/__stf/places/places.cuh
andralex: Yah, though the impl is different by necessity (see `exec_place::device(int devid)`). We do save on allocations etc.

cudax/include/cuda/experimental/__stf/internal/parallel_for_scope.cuh
andralex: Arguably `size()` is the correct mdspan name here - it returns total element count. In `mdspan`, `rank()` means number of dimensions (e.g., 2 for a 2D grid), not element count. Also claude seems to have a similar point with @caugonnet 's about SPMD connotations.

cudax/include/cuda/experimental/__stf/places/exec/cuda_stream.cuh
andralex: will change, thanks

cudax/include/cuda/experimental/__stf/places/exec/cuda_stream.cuh
andralex: yah it's a member, resolving

cudax/include/cuda/experimental/__stf/places/exec/cuda_stream.cuh
andralex: done thanks love the attention to detail

cudax/include/cuda/experimental/__stf/places/exec/cuda_stream.cuh
andralex: Erledigt :)

cudax/include/cuda/experimental/__stf/places/places.cuh
andralex: I asked Claude to look at all implementations and make an assessment :). The base class default (`shared_from_this()`) works for dynamically allocated scalar places (`cuda_stream`, `green_context`) which should be the usual case. Only static instances and grids need overrides, which they already provide. Indeed, if someone creates a new static `impl` and forgets to override, `shared_from_this()` would break. But that's an advanced use case that should be approached with due trepidation.

cudax/include/cuda/experimental/__stf/places/places.cuh
andralex: Per discussion we repudiated `exec_place_host` for simplicity's sake (at the cost of more `if`s). We can add it back if necessary.

cudax/include/cuda/experimental/__stf/internal/parallel_for_scope.cuh
andralex: Actually we're better off now. The old code used static type dispatch to skip compiling device code when `exec_place_t` was `exec_place_host`. With `exec_place_host` removed, we rely on runtime `is_host()` checks and early returns (line 639-641). 

The device code IS compiled for host+device lambdas but not executed at runtime when the place is host. This should be safe because `__host__ __device__` lambdas must be valid for device compilation by definition.

andralex:
@miscco @caugonnet made one more pass minding reviews - thanks!

cudax/include/cuda/experimental/__stf/stream/stream_task.cuh
```diff
@@ -213,19 +206,29 @@ public:
...
+  /**
+   * @brief Activate a sub-place within the task's execution place grid
+   *
+   * Returns an exec_place_scope RAII guard. The sub-place is automatically
+   * deactivated when the guard is destroyed.
+   *
+   * @param p The position within the grid
+   * @return An exec_place_scope guard managing the activation lifetime
+   */
+  exec_place_scope activate_place(pos4 p)
```
caugonnet, line 218: is there a way to "early exit" on that exec_place_scope (eg. with a reset() method) ?

andralex: Currently you can early-exit by assigning an empty scope: `scope = exec_place_scope();`

But a dedicated `reset()` method would be cleaner. Want me to add it?

```cpp
void reset()
{
  if (place_.get_impl())
  {
    place_.get_impl()->deactivate(prev_, idx_);
    place_ = exec_place();
  }
}
```

Want me to add this method?

miscco (approved):

cudax/include/cuda/experimental/__stf/stream/stream_task.cuh
caugonnet: scope = exec_place_scope(); is quite a surprising API to deactivate :)

andralex: ok added reset()

PR #8045, opened 2026-03-16
[STF] Add resource import/export and unit tests

caugonnet:
- ctx_resource_set: export_resources(), import_resources() to move resource ownership between sets without releasing.
- backend_ctx: export_ctx_resources(), import_ctx_resources() on impl; export_resources(), import_resources() on backend_ctx_untyped.
- context: import_resources_from(context&) and private export_resources()/import_resources() for transferring resources from a child context (e.g. nested graph) to the parent.

Unit tests:
- context resources released on finalize: dummy callback-released resource, finalize (blocking), assert released.
- context resources released on finalize non blocking: user-provided stream, non-blocking finalize, sync stream, assert resource released after sync.
- context import_resources_from: add resource to child, import into parent, finalize parent, assert released.

## Description

closes 

## Checklist

### DIFF
--- a/cudax/include/cuda/experimental/__stf/internal/backend_ctx.cuh
+++ b/cudax/include/cuda/experimental/__stf/internal/backend_ctx.cuh
@@ -621,13 +621,13 @@ protected:
     }
 
     // Export all resources by moving them to a new ctx_resource_set
-    ctx_resource_set export_ctx_resources()
+    ctx_resource_set export_resources()
     {
       return ctx_resources.export_resources();
     }
 
     // Import all resources from another ctx_resource_set
-    void import_ctx_resources(ctx_resource_set&& other)
+    void import_resources(ctx_resource_set&& other)
     {
       ctx_resources.import_resources(mv(other));
     }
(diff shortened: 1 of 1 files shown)


caugonnet:
pre-commit.ci autofix

cudax/include/cuda/experimental/__stf/internal/backend_ctx.cuh
```diff
@@ -614,6 +614,18 @@ protected:
     {
       ctx_resources.add(mv(resource));
     }
+
+    // Export all resources by moving them to a new ctx_resource_set
+    ctx_resource_set export_ctx_resources()
```
miscco: Not a fan that this is differently named, but fine by me

caugonnet: @miscco you mean that we have a plural in ctx_ressources vs. ctx_ressource_set ?

miscco: I meant that this is not `export_resources`

caugonnet: Indeed this adds some unnecessary noise, especially if this to add "_ctx_" while it's a method added to the ctx members. I will change to have export_resources thanks ! (and same for import_resources)

PR #8055, opened 2026-03-17
Fix `MergeSort` uninit memory read sanitizer errors

### LINKED ISSUE #8054: [BUG]: Sanitizer errors in `MergeSort`
https://github.com/NVIDIA/cccl/actions/runs/23106493440/job/67117458183#step:4:2177

Weeklies reported uninitialized reads in `DeviceMergeSort`. Find source and either fix or suppress them

gonidelis:
fixes #8054 

Suppress sanitizer initialization check `initcheck` tests on three tests that (most probably) use padding and fail with uninit memory read. Pads are never initialized from `c2h::gen` and kick back.

@miscco is there any way to launch sanitizer just for this PR? It's kinda bulky to check locally in my ws.

### DIFF
--- a/cub/test/catch2_test_device_merge_sort.cu
+++ b/cub/test/catch2_test_device_merge_sort.cu
@@ -367,7 +367,7 @@ C2H_TEST("DeviceMergeSort::SortPairs works", "[merge][sort][device]", wide_key_t
 }
 
 C2H_TEST(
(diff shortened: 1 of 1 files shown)


gonidelis:
Update: I converted to draft. I don't feel safe about the repro. There are tests that at times when I don't patch `[skip-cs-initcheck]` they don't fail and other times they do. Sanitizer is very expensive to run and verification progress is slow.

PR #8066, opened 2026-03-17
Implement `static_count` hierarchy query

davebayer:
Currently, we only have `unit.count(level, hierarchy)`, which works in constexpr context, but only if all of the touched hierarchy levels have static extents. Otherwise the compilation fails. But in many places it would be extremely useful to just query the number of units and get `cuda::std::dynamic_extents` in case of some dimensions are being defined dynamically.

### DIFF
--- a/libcudacxx/test/libcudacxx/cuda/hierarchy/cluster_level/hierarchy_queries.pass.cpp
+++ b/libcudacxx/test/libcudacxx/cuda/hierarchy/cluster_level/hierarchy_queries.pass.cpp
@@ -47,13 +47,16 @@ __device__ void test_cluster(const Hierarchy& hier, const GridExts& grid_exts, c
     test_extents(exp, cuda::cluster, cuda::grid, hier);
   }
 
-  // 4. Test cuda::cluster.count(x, hier)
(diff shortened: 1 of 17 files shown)


libcudacxx/test/libcudacxx/cuda/hierarchy/cluster_level/hierarchy_queries.pass.cpp
```diff
@@ -47,13 +47,16 @@ __device__ void test_cluster(const Hierarchy& hier, const GridExts& grid_exts, c
...
+
+  // 5. Test cuda::cluster.count(x, hier)
   test_count(cuda::std::size_t{dims.z} * dims.y * dims.x, cuda::cluster, cuda::grid, hier);
 
-  // 5. test cuda::cluster.index(x, hier)
+  // 6. test cuda::cluster.index(x, hier)
   test_index(index, cuda::cluster, cuda::grid, hier);
 
-  // 6. Test cuda::cluster.rank(x, hier)
+  // 7. Test cuda::cluster.rank(x, hier)
```
miscco, line 59: Not a fan of that numbering

PR #8068, opened 2026-03-17
Refactor PSTL `temporary_storage`

miscco:
Currently we use `__call_or` to get a memory resource from the policy or a defaulted one.

However, that is suboptimal, because in the case that we want the default device memory pool we need to call a runtime function to get the device from the stream.

We can work around this by moving it into the temporary storage and make that take a memory resource.

this works now that we have fixed `get_default_device_memory_pool`

### DIFF
--- a/libcudacxx/include/cuda/std/__pstl/cuda/temporary_storage.h
+++ b/libcudacxx/include/cuda/std/__pstl/cuda/temporary_storage.h
@@ -112,7 +112,7 @@ class __temporary_storage
   }
 
   //! @brief Helper function to retrieve a memory resource from a policy
-  //!        In contrast to `__call_or` it does not require us to allways call .device() on the stream
+  //!        In contrast to `__call_or` it does not require us to always call .device() on the stream
   template <class _Policy>
   [[nodiscard]] _CCCL_HOST_API static ::cuda::mr::resource_ref<::cuda::mr::device_accessible>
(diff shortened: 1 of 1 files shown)


libcudacxx/include/cuda/std/__pstl/cuda/temporary_storage.h
```diff
@@ -105,22 +111,45 @@ class __temporary_storage
...
-    ::cuda::stream_ref __stream,
-    _Resource& __resource,
-    const size_t __num_bytes_storage,
-    const _Sizes... __elements_stored)
-      : __stream_(__stream)
-      , __resource_(__resource)
+  _CCCL_HOST_API
+  __temporary_storage(const _Policy& __policy, const size_t __num_bytes_storage, const _Sizes... __elements_stored)
+      : __stream_(::cuda::__call_or(::cuda::get_stream, ::cuda::stream_ref{cudaStreamPerThread}, __policy))
+      , __resource_(__get_memory_resource_or(__policy))
```
bernhardmgruber, line 140: Suggestion: an we avoid calling `cuda::get_stream` twice, if their is no memory resource?

miscco: We actually called it before that in the algorithm.

But in general I would assume get_stream to be effectively free

PR #8072, opened 2026-03-17
Improve groups test

davebayer:


cudax/test/hierarchy/group.cu
```diff
@@ -70,21 +85,26 @@ __device__ T sum(cudax::this_cluster<Hierarchy> group, T (&array)[N])
...
 
-                 if (cuda::gpu_thread.rank(cuda::block, group.hierarchy()) == 0)
+                 if (cuda::gpu_thread.rank(cuda::block, group.hierarchy()) == 0 && !is_root_thread)
                  {
-                   atomicAdd(dsmem, result);
+                   [[maybe_unused]] unsigned old;
+                   asm volatile("atom.relaxed.cluster.shared::cluster.add.s32 %0, [%1], %2;"
+                                : "=r"(old)
+                                : "l"(dsmem), "r"(result)
+                                : "memory");
```
miscco, line 101: I there a reason to not stay with the atomicAdd? I would prefer if we do not touch those individualy

davebayer: Yes, I can specify `.cluster` scope and `.shared::cluster` space, which is not possible with `atomicAdd`

PR #8073, opened 2026-03-17
Add clang-tidy

Jacobfaib:
## Description

Add a `clang-tidy` pass that runs over every created executable in CCCL. Run it via
```sh
# run clang-tidy over the whole repository
$ cmake --build build --target cccl.tidy
[0/212] clang-tidy build/all-dev-debug/thrust/testing/thrust.cpp.cpp/address_stability.cu.cpp
# Or only over thrust
$ cmake --build build --target thrust.tidy
# Or libcudacxx...
$ cmake --build build --target libcudacxx.tidy
...
```
A sub-target is also created per project, so `cub.tidy`, `libcudacxx.tidy`, `thrust.tidy`, etc. that will run `clang-tidy` over only that subset of the codebase. These sub-targets depend on the value of `${PROJECT_NAME}`, so whenever a new `project()` call is found, it will create a new sub-target if any `clang-tidy` targets are created in that directory.

The `cccl.tidy` target is only created for top-level builds. Crucially, the `cccl.tidy` target is _lazy_; if the host system has no `clang-tidy`, then an error is only emitted when you invoke `--target cccl.tidy`, not at configure-time. This allows seamless "opt-in" behavior.

Example of current errors generated: 
[clang_tidy_errors_sample.log](https://github.com/user-attachments/files/26069358/clang_tidy_errors_sample.log)

Please note that many errors (notably `bugprone-reserved-identifier` inside libcu++) can be locally silenced by a directory-local `.clang-tidy`, so please ignore them until the local `.clang-tidy` has been added.

## Checklist

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,0 +1,278 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024-2026 NVIDIA CORPORATION & AFFILIATES.
+#                         All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+---
+Checks:
+  # enable everything first
+  - 'performance-*'
+  - 'modernize-*'
+  - 'readability-*'
+  - 'clang-analyzer-*'
+  - 'clang-diagnostic-*'
+  - 'bugprone-*'
+  - 'misc-*'
+  - 'core-*'
+  - 'mpi-*'
+  - 'cert-*'
+  - 'portability-*'
+  - 'google-*'
+  - 'cppcoreguidelines-*'
+  - 'concurrency-*'
+  # TODO: BEGIN REMOVE ME
+  - '-*'
+  - 'performance-implicit-conversion-in-loop'
+  # END REMOVE ME
+  # HICPP is 99% aliased to other checks (mostly modernize-* and bugprone-*). We don't
+  # want to also enable it because then we need to duplicate the NOLINT. The only 2 checks
+  # that aren't aliases are:
+  #
+  # - hicpp-multiway-paths-covered. This check is useful for detecting degenerate if-else
+  #   branches, but the switch cases are IMO better handled by -Wswitch. To make -Wswitch
+  #   even stronger, we should ban default: cases entirely,
+  #
+  # - hicpp-signed-bitwise. This check is also covered by
+  #   clang.llvm.org/extra/clang-tidy/checks/clang-analyzer/core.BitwiseShift.html
+  #
+  # - 'hicpp-*'
+  #
+  # LLVM checks are extremely specific to the LLVM project and aren't suitable for
+  # downstream users.
+  #
+  # - 'llvm-*'
+  #
+  # then disable the stuff we don't want
+  - '-cert-dcl21-cpp'  # returning non-const from operator-- or operator++
+  - '-cert-dcl50-cpp'  # allow c-style variadics
+  # No reserved identifiers, both of these are aliased to bugprone-reserved-identifier,
+  # which we do enable. Leaving these enabled however, leads to needing to specify all
+  # three (bugprone-reserved-identifier, cert-dcl51-cpp, and cert-dcl37-c) in NOLINT lines
+  # which is a hassle. Since bugprone-reserved-identifier is enabled, the check still
+  # fires.
+  - '-cert-dcl51-cpp,-cert-dcl37-c,-cert-oop54-cpp'
+  # Covered by bugprone-throwing-static-initialization
+  - '-cert-err58-cpp'
+  - '-modernize-use-trailing-return-type'
+  - '-readability-function-cognitive-complexity'
+  - '-readability-implicit-bool-conversion'
+  - '-readability-braces-around-statements'
+  - '-readability-qualified-auto'
+  - '-readability-isolate-declaration'
+  - '-modernize-avoid-c-arrays'
+  - '-cppcoreguidelines-avoid-c-arrays'
+  - '-readability-named-parameter'
+  - '-readability-identifier-length'
+  - '-misc-non-private-member-variables-in-classes'
+  - '-bugprone-easily-swappable-parameters'
+  - '-bugprone-implicit-widening-of-multiplication-result'
+  - '-misc-include-cleaner'
+  - '-misc-header-include-cycle'
+  - '-modernize-macro-to-enum'
+  - '-cppcoreguidelines-macro-to-enum'
+  - '-misc-no-recursion'
+  - '-bugprone-dynamic-static-initializers'
+  - '-portability-avoid-pragma-once'
+  # Generally speaking if something is static then it has an express reason to be. 99% of
+  # the candidates identified by this check were because they just so happened not to
+  # touch any member variables, not because they are logically static. So we disable the
+  # check.
+  - '-readability-convert-member-functions-to-static'
+  # When running in a conda env, ignore options that may have been added by conda but are unused
+  - '-clang-diagnostic-unused-command-line-argument'
+  # Given
+  #
+  # std::make_pair<some_type, some_other_type>(...)
+  #
+  # Results in: error: for C++11-compatibility, use pair directly. But we don't care about
+  # C++11, and so we don't care about this warning.
+  - '-google-build-explicit-make-pair'
+  # This check is incredibly expensive for absolutely no reason, and since we a) use
+  # modern google-test and b) don't have clang-tidy enabled on our testing code, we don't
+  # need to enable it!
+  - '-google-upgrade-googletest-case'
+  # This warning will warn if you have redundant default-initializers for class
+  # members. For example:
+  #
+  # class Foo
+  # {
+  #   int x_{}; // WARNING: redundant initializer here
+  #
+  # public:
+  #   Foo(int x) : x_{x} { }
+  # };
+  #
+  # Since Foo can only ever be constructed with an explicit value for x_ via its constructor,
+  # the default initializer is technically redundant. However, if we change the definition
+  # of Foo to now allow a default ctor, then the initializer becomes non-redundant
+  # again. It is easier to just follow the rule of "always explicitly default initialize
+  # members" than to remember to change 2 places at once.
+  - '-readability-redundant-member-init'
+  # Alias for readability-enum-initial-value, disable this one because the readability-
+  # name is easier to understand, and we don't want to silence 2 things for the same
+  # warning.
+  - '-cert-int09-c'
+  # This one is potentially controversial. This check warns when iterating over unordered
+  # containers of pointers:
+  #
+  # {
+  #   int a = 1, b = 2;
+  #   std::unordered_set<int *> set = {&a, &b};
+  #
+  #   for (auto *i : set) { // iteration order not deterministic
+  #     f(i);
+  #   }
+  # }
+  #
+  # On the one hand, this is a clear case of non-determinism. But on the other hand, I
+  # feel it is obvious that a user does not care about the order of iteration
+  # because... they are using unordered containers!
+  - '-bugprone-nondeterministic-pointer-iteration-order'
+  - '-cppcoreguidelines-pro-type-reinterpret-cast'
+  # We don't use GSL
+  - '-cppcoreguidelines-owning-memory'
+  # Covered by modernize-use-override already
+  - '-cppcoreguidelines-explicit-virtual-functions'
+  # Covered by readability-magic-numbers
+  - '-cppcoreguidelines-avoid-magic-numbers'
+  # This check does more harm than good, as it appears to be very rudimentary. For
+  # example, it emits warnings saying that:
+  #
+  # #define DEFINE_OPERATOR(op)                            \
+  #   _CCCL_HOST_DEVICE custom_numeric operator op() const \
+  #   {                                                    \
+  #     return custom_numeric(op value[0]);                \
+  #   }
+  #
+  # Should be replaced by a template function, but this is not possible. We instead rely
+  # on best judgment of reviewers to catch macros that can be functions.
+  - '-cppcoreguidelines-macro-usage'
+  # This check does not understand more complex macro usage where brackets are not allowed
+  #
+  # error: macro replacement list should be enclosed in parentheses [bugprone-macro-parentheses]
+  # 36 | #define __THRUST_HOST_SYSTEM_INCLUDE(filename) <__THRUST_HOST_SYSTEM_ROOT/filename>
+  #    |                                                                           ^
+  #    |                                                                           (       )
+  #
+  # So better left disabled.
+  - '-bugprone-macro-parentheses'
+  # TODO: ironically enough, enable this someday
+  - '-google-readability-todo'
+  # Covered by modernize-avoid-c-style-cast, and also is buggier. It seems to fire on
+  # `_v`-style constexpr bools as well sometimes.
+  - '-google-readability-casting'
+  # Covered by performance-noexcept-move-constructor
+  - '-cppcoreguidelines-noexcept-move-operations'
+  # REVIEW ME: This warns about any usage of operator[], suggesting usage of .at() instead. Could
+  - '-cppcoreguidelines-pro-bounds-avoid-unchecked-container-access'
+  # Covered by bugprone-narrowing-conversions
+  - '-cppcoreguidelines-narrowing-conversions'
+  # Covered by misc-unconventional-assign-operator
+  - '-cppcoreguidelines-c-copy-assignment-signature'
+  # Covered by performance-noexcept-swap
+  - '-cppcoreguidelines-noexcept-swap'
+  # Covered by modernize-use-default-member-init
+  - '-cppcoreguidelines-use-default-member-init'
+  # clang-tidy documentation says that for a given file, it matches the closest
+  # .clang-tidy up the directory stack and applies the configuration from it. However, by
+  # "it" `clang-tidy` means the closest `.clang-tidy` to the *source* file, not the file
+  # that was included by the source file.
+  #
+  # So `cuda/std/foo.h` included by `cccl/foo/bar/baz.cpp` will never "match" against
+  # `cccl/libcudacxx/.clang-tidy` because `clang-tidy` finds `cccl/.clang-tidy` instead.
+  #
+  # So we disable this for now. In the future we should add a custom check that bans
+  # reserved identifiers except symbols from libcu++.
+  - '-bugprone-reserved-identifier'
+WarningsAsErrors: '*'
+HeaderFileExtensions:
+  - ''
+  - h
+  - hh
+  - hpp
+  - hxx
+  - cuh
+  - inl
+  - ipp
+ImplementationFileExtensions:
+  - c
+  - cc
+  - cpp
+  - cxx
+  - cu
+SystemHeaders: false
+ExtraArgsBefore:
+  # To match _CCCL_DOXYGEN_INVOKED
+  - '-D_CCCL_CLANG_TIDY_INVOKED=1'
+  - '-ftemplate-backtrace-limit=0'
+  - '-fmacro-backtrace-limit=0'
+ExtraArgs:
+  # These must come in ExtraArgs not ExtraArgsBefore, because they are overriding the
+  # effects of -Wall and -Werror, where the last flag "wins".
+  - '-Wno-unknown-warning-option'
+  - '-Wno-unknown-cuda-version'
+  - '-Wno-error=unused-command-line-argument'
+CheckOptions:
+  cert-dcl16-c.NewSuffixes: 'L;LL;LU;LLU'
+  cert-err33-c.CheckedFunctions: '::aligned_alloc;::asctime_s;::at_quick_exit;::atexit;::bsearch;::bsearch_s;::btowc;::c16rtomb;::c32rtomb;::calloc;::clock;::cnd_broadcast;::cnd_init;::cnd_signal;::cnd_timedwait;::cnd_wait;::ctime_s;::fclose;::fflush;::fgetc;::fgetpos;::fgets;::fgetwc;::fopen;::fopen_s;::fprintf;::fprintf_s;::fputc;::fputs;::fputwc;::fputws;::fread;::freopen;::freopen_s;::fscanf;::fscanf_s;::fseek;::fsetpos;::ftell;::fwprintf;::fwprintf_s;::fwrite;::fwscanf;::fwscanf_s;::getc;::getchar;::getenv;::getenv_s;::gets_s;::getwc;::getwchar;::gmtime;::gmtime_s;::localtime;::localtime_s;::malloc;::mbrtoc16;::mbrtoc32;::mbsrtowcs;::mbsrtowcs_s;::mbstowcs;::mbstowcs_s;::memchr;::mktime;::mtx_init;::mtx_lock;::mtx_timedlock;::mtx_trylock;::mtx_unlock;::printf_s;::putc;::putwc;::raise;::realloc;::remove;::rename;::scanf;::scanf_s;::setlocale;::setvbuf;::signal;::snprintf;::snprintf_s;::sprintf;::sprintf_s;::sscanf;::sscanf_s;::strchr;::strerror_s;::strftime;::strpbrk;::strrchr;::strstr;::strtod;::strtof;::strtoimax;::strtok;::strtok_s;::strtol;::strtold;::strtoll;::strtoul;::strtoull;::strtoumax;::strxfrm;::swprintf;::swprintf_s;::swscanf;::swscanf_s;::thrd_create;::thrd_detach;::thrd_join;::thrd_sleep;::time;::timespec_get;::tmpfile;::tmpfile_s;::tmpnam;::tmpnam_s;::tss_create;::tss_get;::tss_set;::ungetc;::ungetwc;::vfprintf;::vfprintf_s;::vfscanf;::vfscanf_s;::vfwprintf;::vfwprintf_s;::vfwscanf;::vfwscanf_s;::vprintf_s;::vscanf;::vscanf_s;::vsnprintf;::vsnprintf_s;::vsprintf;::vsprintf_s;::vsscanf;::vsscanf_s;::vswprintf;::vswprintf_s;::vswscanf;::vswscanf_s;::vwprintf_s;::vwscanf;::vwscanf_s;::wcrtomb;::wcschr;::wcsftime;::wcspbrk;::wcsrchr;::wcsrtombs;::wcsrtombs_s;::wcsstr;::wcstod;::wcstof;::wcstoimax;::wcstok;::wcstok_s;::wcstol;::wcstold;::wcstoll;::wcstombs;::wcstombs_s;::wcstoul;::wcstoull;::wcstoumax;::wcsxfrm;::wctob;::wctrans;::wctype;::wmemchr;::wprintf_s;::wscanf;::wscanf_s;'
+  llvm-else-after-return.WarnOnUnfixable: 'false'
+  cert-str34-c.DiagnoseSignedUnsignedCharComparisons: 'false'
+  cppcoreguidelines-non-private-member-variables-in-classes.IgnoreClassesWithAllMemberVariablesBeingPublic: 'true'
+  google-readability-braces-around-statements.ShortStatementLines: '1'
+  llvm-qualified-auto.AddConstToQualified: 'false'
+  llvm-else-after-return.WarnOnConditionVariables: 'false'
+  cert-oop54-cpp.WarnOnlyIfThisHasSuspiciousField: 'false'
+  performance-move-const-arg.CheckTriviallyCopyableMove: 'false'
+  performance-inefficient-string-concatenation.StrictMode: 'true'
+  readability-simplify-boolean-expr.ChainedConditionalReturn: 'true'
+  readability-simplify-boolean-expr.ChainedConditionalAssignment: 'true'
+  bugprone-dangling-handle.HandleClasses: '::cuda::std::span;::cuda::std::mdspan'
+  bugprone-misplaced-widening-cast.CheckImplicitCasts: 'true'
+  bugprone-unused-return-value.AllowCastToVoid: 'true'
+  readability-enum-initial-value.AllowExplicitZeroFirstInitialValue: 'false'
+  readability-enum-initial-value.AllowExplicitSequentialInitialValues: 'false'
+  readability-redundant-access-specifiers.CheckFirstDeclaration: 'true'
+  bugprone-lambda-function-name.IgnoreMacros: 'true'
+  # readability-identifier-naming.ClassCase: 'lower_case'
+  # readability-identifier-naming.UnionCase: 'lower_case'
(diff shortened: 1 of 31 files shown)


Jacobfaib:
Some follow-ups for (eventual) reviewers: `readability-identifier-naming` allows automating the naming scheme with `clang-tidy`. Each subproject follows _roughly_ the same style guide, but not always (`cudax/` for example is a bit all over the place).

I propose these checks are enabled in a follow-up, as it will be adding to an already massive diff in this PR.

Jacobfaib:
Also which CI job is best suited for this? I notice there isn't really a unified "lint" job (other than `pre-commit`). `clang-tidy` works only on what it sees. We don't need to run it for every combination of compiler and OS that CCCL supports, but it should run in a configuration with as many optional features enabled as possible.

.clang-tidy
```diff
@@ -0,0 +1,185 @@
...
+  - 'clang-analyzer*'
+  - 'clang-diagnostic-*'
+  - 'bugprone*'
+  - 'misc*'
+  - 'core*'
+  - 'mpi*'
+  - 'cert-*'
+  - 'portability-*'
+  - 'google*'
+  - 'cppcoreguidelines-pro-type-cstyle-cast'
```
fbusato: for consistency and following the clang-tidy guide

```suggestion
  - 'performance-*'
  - 'modernize-*'
  - 'readability-*'
  - 'clang-analyzer-*'
  - 'clang-diagnostic-*'
  - 'bugprone-*'
  - 'misc-*'
  - 'core-*'
  - 'mpi-*'
  - 'cert-*'
  - 'portability-*'
  - 'google-*'
  - 'cppcoreguidelines-*'
```

.clang-tidy
```diff
@@ -0,0 +1,185 @@
...
+  - 'modernize*'
+  - 'readability*'
+  - 'clang-analyzer*'
+  - 'clang-diagnostic-*'
+  - 'bugprone*'
+  - 'misc*'
+  - 'core*'
+  - 'mpi*'
+  - 'cert-*'
+  - 'portability-*'
```
fbusato, line 17: is it expected to not include `concurrency-*`,   `hicpp-*`, and `llvm-*`?

.clang-tidy
```diff
@@ -0,0 +1,185 @@
...
+  - '-cert-dcl51-cpp,-cert-dcl37-c,-cert-oop54-cpp'
+  - '-modernize-use-trailing-return-type'
+  - '-readability-function-cognitive-complexity'
+  - '-readability-implicit-bool-conversion'
+  - '-readability-braces-around-statements'
+  - '-readability-qualified-auto'
+  - '-readability-isolate-declaration'
+  - '-modernize-avoid-c-arrays'
+  - '-readability-named-parameter'
+  - '-readability-identifier-length'
```
fbusato, line 63: other suggestions:

- `readability-avoid-nested-conditional-operator` https://clang.llvm.org/extra/clang-tidy/checks/readability/avoid-nested-conditional-operator.html
- `performance-enum-size` https://clang.llvm.org/extra/clang-tidy/checks/performance/enum-size.html
- `cppcoreguidelines-pro-type-reinterpret-cast` https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines/pro-type-reinterpret-cast.html
- `readability-math-missing-parentheses` https://clang.llvm.org/extra/clang-tidy/checks/readability/math-missing-parentheses.html
- `misc-use-internal-linkage` https://clang.llvm.org/extra/clang-tidy/checks/misc/use-internal-linkage.html (false positive/unit test)
- `google-build-using-namespace` https://clang.llvm.org/extra/clang-tidy/checks/google/build-using-namespace.html, used in unit tests
- `cppcoreguidelines-owning-memory` https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines/owning-memory.html
- `modernize-use-designated-initializers` https://clang.llvm.org/extra/clang-tidy/checks/modernize/use-designated-initializers.html (we use c++17)

.clang-tidy
```diff
@@ -0,0 +1,185 @@
...
+  # readability-identifier-naming.PrivateMemberSuffix: '_'
+  # readability-identifier-naming.PrivateMethodCase: 'lower_case'
+  # readability-identifier-naming.PrivateMethodSuffix: '_'
+  # readability-identifier-naming.ProtectedMemberCase: 'lower_case'
+  # readability-identifier-naming.ProtectedMemberSuffix: '_'
+  # readability-identifier-naming.ProtectedMethodCase: 'lower_case'
+  # readability-identifier-naming.ProtectedMethodSuffix: '_'
+  # readability-identifier-naming.PublicMethodCase: 'lower_case'
+  # readability-identifier-naming.ScopedEnumCase: 'lower_case'
+  # readability-identifier-naming.ScopedEnumConstantCase: 'UPPER_CASE'
```
fbusato, line 262: this is only for historical reasons

.clang-tidy
```diff
@@ -0,0 +1,185 @@
...
+  # readability-identifier-naming.ClassCase: 'lower_case'
+  # readability-identifier-naming.UnionCase: 'lower_case'
+  # readability-identifier-naming.ClassConstantCase: 'UPPER_CASE'
+  # readability-identifier-naming.ClassIgnoredRegexp: 'tuple|has_.*|is_.*|as_.*|.*_tag|tag|.*_of'
+  # readability-identifier-naming.ConstantMemberCase: 'UPPER_CASE'
+  # readability-identifier-naming.ConstantMemberIgnoredRegexp: 'value'
+  # readability-identifier-naming.EnumCase: 'lower_case'
+  # readability-identifier-naming.EnumConstantCase: 'UPPER_CASE'
+  # readability-identifier-naming.FunctionCase: 'lower_case'
+  # readability-identifier-naming.GlobalConstantCase: 'UPPER_CASE'
```
fbusato, line 243: this is only for historical reasons

.clang-tidy
```diff
@@ -0,0 +1,185 @@
...
+  readability-redundant-access-specifiers.CheckFirstDeclaration: 'true'
+  bugprone-lambda-function-name.IgnoreMacros: 'true'
+  # readability-identifier-naming.ClassCase: 'lower_case'
+  # readability-identifier-naming.UnionCase: 'lower_case'
+  # readability-identifier-naming.ClassConstantCase: 'UPPER_CASE'
+  # readability-identifier-naming.ClassIgnoredRegexp: 'tuple|has_.*|is_.*|as_.*|.*_tag|tag|.*_of'
+  # readability-identifier-naming.ConstantMemberCase: 'UPPER_CASE'
+  # readability-identifier-naming.ConstantMemberIgnoredRegexp: 'value'
+  # readability-identifier-naming.EnumCase: 'lower_case'
+  # readability-identifier-naming.EnumConstantCase: 'UPPER_CASE'
```
fbusato, line 241: this is only for historical reasons

.clang-tidy
```diff
@@ -0,0 +1,185 @@
...
+  bugprone-unused-return-value.AllowCastToVoid: 'true'
+  readability-enum-initial-value.AllowExplicitZeroFirstInitialValue: 'false'
+  readability-enum-initial-value.AllowExplicitSequentialInitialValues: 'false'
+  readability-redundant-access-specifiers.CheckFirstDeclaration: 'true'
+  bugprone-lambda-function-name.IgnoreMacros: 'true'
+  # readability-identifier-naming.ClassCase: 'lower_case'
+  # readability-identifier-naming.UnionCase: 'lower_case'
+  # readability-identifier-naming.ClassConstantCase: 'UPPER_CASE'
+  # readability-identifier-naming.ClassIgnoredRegexp: 'tuple|has_.*|is_.*|as_.*|.*_tag|tag|.*_of'
+  # readability-identifier-naming.ConstantMemberCase: 'UPPER_CASE'
```
fbusato, line 238: this is only for historical reasons

.clang-tidy
```diff
@@ -0,0 +1,185 @@
...
+  bugprone-dangling-handle.HandleClasses: '::cuda::std::span;::cuda::std::mdspan'
+  bugprone-misplaced-widening-cast.CheckImplicitCasts: 'true'
+  bugprone-unused-return-value.AllowCastToVoid: 'true'
+  readability-enum-initial-value.AllowExplicitZeroFirstInitialValue: 'false'
+  readability-enum-initial-value.AllowExplicitSequentialInitialValues: 'false'
+  readability-redundant-access-specifiers.CheckFirstDeclaration: 'true'
+  bugprone-lambda-function-name.IgnoreMacros: 'true'
+  # readability-identifier-naming.ClassCase: 'lower_case'
+  # readability-identifier-naming.UnionCase: 'lower_case'
+  # readability-identifier-naming.ClassConstantCase: 'UPPER_CASE'
```
fbusato, line 236: this is only for historical reasons

.clang-tidy
miscco: I am wary of a lot of those warnings
* readability-avoid-nested-conditional-operator
* performance-enum-size
* cppcoreguidelines-pro-type-reinterpret-cast

.clang-tidy
Jacobfaib: Yes.

- `hicpp` is 99% aliased to other checks (mostly `modernize-*` and `bugprone-*`). We don't want to also enable it because then we need to duplicate the `NOLINT`[^1]. The only 2 checks that aren't aliases are:
  - [`hicpp-multiway-paths-covered`](https://clang.llvm.org/extra/clang-tidy/checks/hicpp/multiway-paths-covered.html). This check is useful for detecting degenerate `if-else` branches, but the switch cases are IMO better handled by `-Wswitch`. To make `-Wswitch` even stronger, we should ban `default:` cases entirely, I had added a `pre-commit` [check](https://github.com/nv-legate/legate/blob/main/scripts/pre-commit/no_default.py) to this effect in Legate previously which I can port over here.
  - [`hicpp-signed-bitwise`](https://clang.llvm.org/extra/clang-tidy/checks/hicpp/signed-bitwise.html) is also covered by https://clang.llvm.org/extra/clang-tidy/checks/clang-analyzer/core.BitwiseShift.html (undocumented, but true).

- `llvm-*` is extremely specific to LLVM, e.g. [llvm-twine-local](https://clang.llvm.org/extra/clang-tidy/checks/llvm/twine-local.html), [llvm-include-order](https://clang.llvm.org/extra/clang-tidy/checks/llvm/include-order.html), [llvm-use-new-mlir-op-builder](https://clang.llvm.org/extra/clang-tidy/checks/llvm/use-new-mlir-op-builder.html), and not really suitable for downstream projects.

- `concurrency-*` we can enable, I accidentally missed it.

I'll add comments explaining this to the `.clang-tidy`

[^1]: If `foo` and `bar` are aliased, `NOLINT(foo)` doesn't disable `bar`, you need to write `NOLINT(foo, bar)`, it's quite annoying. This is also why we disable most of the `cert-*` checks -- they are all covered by `bugprone-*` or `portability-*`.

.clang-tidy
Jacobfaib: - `readability-avoid-nested-conditional-operator`: you want to disable this? IMO this is a reasonable check, `condition1 ? (condition2 ? true1 : false1) : false2;` is quite unreadable, and is better expressed as `if`s...
- `performance-enum-size`: Would affect ABI, yes. If we don't care so much about ABI, then this is a no-brainer to enable. 
- `cppcoreguidelines-pro-type-reinterpret-cast`: Yes, we should disable this.
- `readability-math-missing-parentheses`: This is nice, IMO. It introduces no semantic changes (operator precedence is fixed by the standard, after all), but makes the code more understandable at a glance.
- `misc-use-internal-linkage`: This only fires in `.cpp`s, `clang-tidy` is smart enough not to warn for header files. Looking over the error log, I don't see any false positives.
- `cppcoreguidelines-owning-memory`: Yes we should disable this.
- `modernize-use-designated-initializers`: This will only fire if we enable C++20 mode. If we run `clang-tidy` in C++17 mode it emits no warnings.

Jacobfaib:
> We do have standard library components and need to account for that

The way this will be handled will be a separate `.clang-tidy` under `libcudacxx/`, I just haven't put it in yet. When emitting diagnostics, `clang-tidy` will find the closest `.clang-tidy` to the file that contains the code, so we can have different configurations for each package. For example, we will need to disable `bugprone-reserved-identifier` for it (actually, `bugprone-reserved-identifier` has an `Invert` option that flags names that are not prefixed with `__`).

.clang-tidy
Jacobfaib: I propose customizing `readability-identifier-naming` in a follow-up. I left all the options in there so we know what exists, but there will no doubt be a monumental amount of bike-shedding on this topic (not to mention a huge diff) so better to do this all in a separate PR.

.clang-tidy
bernhardmgruber: > This will only fire if we enable C++20 mode. If we run clang-tidy in C++17 mode it emits no warnings.

We compile all our code using C++17 and C++20 (and soon C++23), so I would avoid enabling warnings that force us to use featues which we cannot use in older language dialects.

.clang-tidy
bernhardmgruber: We do have a style guide, but I assume this could be a good change to discuss some of the rules. Also, we can likely not apply the style fixes all in one go. This will be a longer effort I believe.

bernhardmgruber:
> > We do have standard library components and need to account for that
> 
> The way this will be handled will be a separate `.clang-tidy` under `libcudacxx/`, I just haven't put it in yet. When emitting diagnostics, `clang-tidy` will find the closest `.clang-tidy` to the file that contains the code, so we can have different configurations for each package. For example, we will need to disable `bugprone-reserved-identifier` for it (actually, `bugprone-reserved-identifier` has an `Invert` option that flags names that are not prefixed with `__`).

CUB and Thrust also use `__` prefixes on some internal functions. At least I started using that. But a lot of other code uses a `detail` namespace. So we can actually run neither check before harmonizing to a single rule for internal functions :D

jrhemstad:
> Also which CI job is best suited for this? I notice there isn't really a unified "lint" job (other than `pre-commit`). `clang-tidy` works only on what it sees. We don't need to run it for every combination of compiler and OS that CCCL supports, but it should run in a configuration with as many optional features enabled as possible.

@alliepiper can you point @Jacobfaib in the right direction here?

.clang-tidy
Jacobfaib: That's ok, we just need to make sure the clang-tidy job runs only in C++17 mode (not a bad idea anyway to enforce minimum requirements).

If we want to be absolutely sure, we can even add `-std=c++17` to the extra flags that are appended for clang-tidy only. That should override any standard set by the build system.

Jacobfaib:
> CUB and Thrust also use __ prefixes on some internal functions. At least I started using that. But a lot of other code uses a detail namespace. So we can actually run neither check before harmonizing to a single rule for internal functions :D

So I misread `clang-tidy` when it said that it would match the closest `.clang-tidy` to a particular file. It matches the closest `.clang-tidy` to the _source_ file, not the file that was included by the source file. So `cuda/std/foo.h` included by `cccl/foo/bar/baz.cpp` will never "match" against `cccl/libcudacxx/.clang-tidy` because `clang-tidy` finds `cccl/.clang-tidy`.

To fix this, I think we just disable `bugprone-reserved-identifier` entirely for now and then I can add in a custom check that looks up the `.clang-tidy` based on the location of the symbol (or just hard-code that anything in `::cuda::std` is allowed `__`).

> CUB and Thrust also use __ prefixes on some internal functions.

My 2 cents would be that only `::cuda::std` is allowed the `__`. As low-level and special to NVIDIA as CUB and thrust are, they aren't really a "standard library" so `detail` namespace is appropriate. But will defer to others on this decision.

miscco:
> My 2 cents would be that only `::cuda::std` is allowed the `__`. As low-level and special to NVIDIA as CUB and thrust are, they aren't really a "standard library" so `detail` namespace is appropriate. But will defer to others on this decision.

I believe that is correct, albeit we are starting to use effectively all of CUB in the PSTL CUDA backend, so technically ...

alliepiper:
> Also which CI job is best suited for this?

@Jacobfaib We should put this in a new CI job. We'll need that level of control. As a first step, I propose adding new ci/tidy_<project>.sh scripts that use the existing common-ci-job infra to compile some new CMake presets, and some "tidy" jobs in ci/matrix.yaml that mirror the "build" job configs.

Take a look through the comments and docs in:

- ci/matrix.yaml (job defs, script locations and interfaces). See how the "build" job is setup (ignore the "build-lid0" etc variants, you won't need those), and use that as a template for defining a "tidy" job.
- `CMakePresets.json` is where you'll configure the tidy job configure, build, and (if needed?) ctest presets for the per-project, per-dialect builds.
- `ci/build_libcudacxx.sh` is the most straightforward build script to use as a template. You shouldn't need to mirror the complexity of say, `build_cub.sh` to get this working. We can look into splitting things up later if the runtime is bad enough.

----

Important: Use the `override` matrix in the `ci/matrix.yaml` file to test your new jobs without triggering a full 500+ job CI pipeline execution while iterating.

----

So:

1. New "tidy" job type def + pull_request/nightly/weekly jobs in `ci/matrix.yaml` (test with `override` matrix)
2. New config + build presets in `CMakePresets.json`
3. New `ci/tidy_<project>.sh` scripts per-project, following the `build_libcudacxx.sh` script as a template.

Jacobfaib:
Per offline discussion, the consensus was that trying to go whole-hog with all checks enabled and fixes applied would be a nightmare to review/downright impossible.

Instead, we should merge a minimal passing version of this PR in ASAP to get the `clang-tidy` machinery itself into CI and the build system, then slowly ratchet up the enabled checks until we have everything enabled/fixed.

To whit, this PR now enables only a single check, `performance-implicit-conversion-in-loop` (https://github.com/NVIDIA/cccl/pull/8073/changes#diff-0534891fbc8b89b4c521057e7e8dff70d3cbeb822cc9991a3abad77848c289f8R23) (which CCCL seems to be passing).

ci/matrix.yaml
```diff
@@ -561,6 +574,12 @@ projects:
   bisect:
     name: 'Bisect'
     stds: [17, 20]
+  # Used by the clang-tidy jobs. Does not really mean "all", because it does not name all
+  # projects, this is more "project-less" because clang-tidy will by default be invoked
+  # over all configured sub-projects.
+  all:
```
alliepiper: We need to split this up into per-project scripts from this list to work with our change-detection logic that determines which areas of the CCCL need to be built and tested.

thrust/thrust/iterator/zip_iterator.h
```diff
@@ -293,7 +294,8 @@ class _CCCL_DECLSPEC_EMPTY_BASES zip_iterator : public detail::make_zip_iterator
   inline _CCCL_HOST_DEVICE typename super_t::difference_type
   distance_to(const zip_iterator<OtherIteratorTuple>& other) const
   {
-    return ::cuda::std::get<0>(other.get_iterator_tuple()) - ::cuda::std::get<0>(get_iterator_tuple());
+    return ::cuda::std::distance(
+      ::cuda::std::get<0>(other.get_iterator_tuple()), ::cuda::std::get<0>(get_iterator_tuple()));
```
bernhardmgruber: Important: the arguments need to be swapped when using `distance`.

Jacobfaib: Yeah, fixed it in the next commit :)

cmake/CCCLAddTidyTarget.cmake
```diff
@@ -0,0 +1,175 @@
...
+  endif()
+
+  # Do not set to cache; multiple separate instances of CCCL in a build should not
+  # conflict.
+  set(CCCL_TIDY_INITIALIZED TRUE)
+  set(CCCL_TIDY_INITIALIZED TRUE PARENT_SCOPE)
+
+  if (CCCL_CLANG_TIDY)
+    add_custom_target(cccl.tidy COMMENT "clang-tidy CCCL")
+
```
miscco:
```suggestion
```

cub/test/internal/catch2_test_integer_utils.cu
```diff
@@ -25,11 +25,13 @@ using integral_types =
 using floating_point_types =
   c2h::type_list<float,
                  double
-#if TEST_HALF_T()
+// TODO(jfaibussowit): re-enable these, clang-tidy seems to choke on operator==() between
+// __half, __nv_bfloat16, and short below
+#if TEST_HALF_T() && !defined(_CCCL_CLANG_TIDY_INVOKED)
```
miscco, line 30: We technically should not rely on those operators, because a user might remove them.

We should revisit that piece of code in the future

cub/test/catch2_test_env_launch_helper.h
```diff
@@ -512,7 +512,21 @@ void launch(ActionT action, Args... args)
...
+#    if _CCCL_HAS_CDP()
   *d_error = action(args...);
+#    else
+  static_cast<void>(action);
+  (static_cast<void>(args), ...);
+  *d_error = cudaSuccess;
+#    endif
+#  else
+  *d_error = action(args...);
+#  endif
```
miscco: Please add comments to the end of conditionals
```suggestion
#  ifdef _CCCL_CLANG_TIDY_INVOKED
#    if _CCCL_HAS_CDP()
*d_error = action(args...);
#    else // ^^^  _CCCL_HAS_CDP() ^^^ / vvv ! _CCCL_HAS_CDP() vvv

  static_cast<void>(action);
  (static_cast<void>(args), ...);
  *d_error = cudaSuccess;

#    endif // ! _CCCL_HAS_CDP()
#  else ^^^ _CCCL_CLANG_TIDY_INVOKED ^^^ / vvv !_CCCL_CLANG_TIDY_INVOKED vvv
  *d_error = action(args...);
#  endif // !_CCCL_CLANG_TIDY_INVOKED
```

cudax/test/launch/launch_smoke.cu
```diff
@@ -7,16 +7,22 @@
...
-#include <cuda/memory>
 
-#include <cuda/experimental/graph.cuh>
-#include <cuda/experimental/kernel.cuh>
-#include <cuda/experimental/launch.cuh>
-#include <cuda/experimental/stream.cuh>
+// clang does not support __managed__ declarations, so clang-tidy produces spurious
+// errors. https://github.com/llvm/llvm-project/pull/149716 seemingly adds support for
+// __managed__ variables, but that PR seems to have stagnated.
+#ifndef _CCCL_CLANG_TIDY_INVOKED
```
miscco, line 14: I am wondering whether we should move that to cmake and glob out all the files that use managed memory

cudax/test/launch/launch_smoke.cu
```diff
@@ -386,3 +392,5 @@ C2H_TEST("Launch with default config", "")
 {
   test_default_config();
 }
+
+#endif
```
miscco:
```suggestion
#endif // _CCCL_CLANG_TIDY_INVOKED
```

thrust/benchmarks/bench/transform/zip_transform.cu
```diff
@@ -34,6 +34,10 @@ static void nstream_zip_transform(nvbench::state& state, nvbench::type_list<T>)
 
   const T scalar = startScalar;
   auto lambda    = cuda::proclaim_copyable_arguments([scalar] _CCCL_DEVICE(const T& ai, const T& bi, const T& ci) -> T {
+    // Needed to silence clangs -Wunused-lambda-capture. We cannot just remove it because other
+    // implementations (e.g. MSVC) will emit errors if we don't capture it. See discussion in
+    // https://reviews.llvm.org/D28467.
+    static_cast<void>(scalar);
```
miscco, line 40: Can we use `[[maybe_unused]]` in the lambda capture?

cudax/test/launch/launch_smoke.cu
Jacobfaib: Yes, we should. I considered doing that in this PR, but it ended up being quite an involved change so opted for this `#ifdef` stop-gap instead. I am actually not sure why this isn't already showing up in CI, do the `clang-cuda` jobs just not build `cudax` at the moment?

thrust/benchmarks/bench/transform/zip_transform.cu
Jacobfaib: No, you cannot yet apply attributes to the captured variables. The two places where attributes can be attached: 
```
[captures] front-attr ... (params) back-attr ... { body }
```
will apply the attributes to the type of the lambda and `operator()` respectively.

miscco: :(

ci/matrix.yaml
```diff
@@ -88,6 +88,16 @@ workflows:
...
+    # clang-tidy
+    #
+    # The precise value of sm is not important, but it is required for cmake to identify
+    # clang as a CUDA compiler (see
+    # https://discourse.cmake.org/t/cmake-cuda-clang-fails/8657/5).
+    #
+    # Standard being exactly "min" is required. clang-tidy may emit additional diagnostics
+    # for later C++ versions (for example, warning that you should use designated
+    # initializers in C++20 or higher).
+    - { jobs: ['build'], project: 'tidy', std: 'min', cxx: ['clang'], cudacxx: ['clang'], ctk: 'clang-cuda', sm: '75' }
```
alliepiper, line 101: This should also run in nightly/weekly. ~~I'd suggest using `minmax` for `std.` These jobs look cheap enough that we can spare the extra CPU runners to cover both standards.~~

ci/matrix.yaml
```diff
@@ -561,6 +571,9 @@ projects:
   bisect:
     name: 'Bisect'
     stds: [17, 20]
+  tidy:
+    name: 'clang-tidy'
+    stds: [17]
```
alliepiper, line 581: Why not 20?

cmake/CCCLAddTidyTarget.cmake
```diff
@@ -0,0 +1,186 @@
+include_guard(GLOBAL)
+
+find_program(CCCL_CLANG_TIDY clang-tidy)
+
+function(_cccl_tidy_create_target target_name project_name)
+  if (CCCL_CLANG_TIDY)
+    add_custom_target("${target_name}" COMMENT "clang-tidy ${project_name}")
+  else()
+    add_custom_target(
```
alliepiper: If the user requests tidy and its not available, the configure should fail. This branch should just be a message(FATAL_ERROR ...).

Ideally we'd have a top-level CCCL_ENABLE_CLANG_TIDY cmake option that controls whether or not these targets are added. If true, we find clang-tidy and build the targets. If not, it's a hard config error so the user knows about the missing tooling as early as possible.

CMakeLists.txt
```diff
@@ -23,6 +23,9 @@ endif()
 # Enable CXX so CMake can configure install paths
 project(CCCL LANGUAGES CXX)
 
+# Needed for clang-tidy so that relative paths can be disambiguated
+set(CCCL_TOPLEVEL_DIRECTORY "${CMAKE_CURRENT_LIST_DIR}")
```
alliepiper: Not needed -- `project(CCCL ...)` defines `CCCL_SOURCE_DIR` as a conventional way to get this path.

ci/matrix.yaml
Jacobfaib: It has to be the minimum. It's explained further up where the job is defined, but `clang-tidy` gates certain diagnostics behind standard version because they aren't suitable for lower standards.

For example, if you compile with `-std=c++20` `clang-tidy` will recommend you use designated initializers, or use ranges instead of iterator based algorithms. Since we need to support C++17, those wouldn't be appropriate to implement.

CMakeLists.txt
Jacobfaib: Does that still work for multiple concurrent CCCL's in a build? `project()` will put that stuff in the cache, I don't recall if it sets normal variables as well.

cmake/CCCLAddTidyTarget.cmake
Jacobfaib: I wanted to avoid a `CCCL_ENABLE_CLANG_TIDY` because for developers (i.e. top-level builds) it should always be enabled _if possible_. At the very least, it should not require developers who don't have `clang-tidy` to start passing `-DCCCL_ENABLE_CLANG_TIDY=OFF` to their builds all over the place. For users, this should never be enabled.

But I see your point, if you configured thinking `clang-tidy` exists but you need to wait until runtime/build-time to find that out, that's annoying.

This is also kind of the reason I added `CCCL_FORCE_CLANG_TIDY`, I needed a way to disambiguate between:

1. A dev, doing a dev build. If `clang-tidy` exists, it should be enabled and set up in case the dev wants to run it later. If not, it's not required (until later, if they try to run `clang-tidy`).
2. CI, doing the `clang-tidy` job/someone who requires `clang-tidy` to be working. `clang-tidy` should be enabled come hell or high water.

I could check `CCCL_FORCE_CLANG_TIDY` when I `find_program(clang-tidy)` and make it `REQUIRED` if `CCCL_FORCE_CLANG_TIDY` is truthy. Would this be acceptable?

Jacobfaib: Another alternative is that I make `CCCL_ENABLE_CLANG_TIDY` have 3 values:

1. `ON`: give me `clang-tidy` or give me death.
2. `ON_IF_SUPPORTED`: try to enable `clang-tidy`, but if that doesn't work, no hard error until you try to run it.
3. `OFF`: no `clang-tidy`.

The default value will depend on whether we think a developer build is in progress. If so, it is `ON_IF_SUPPORTED`, if not, it is `OFF`.

This way the conventional `-DCCCL_ENABLE_CLANG_TIDY=<bool>` works but we can have the developer build Just Work(TM) as well.

alliepiper: The reproducibility > convenience thing from the other PR applies here, too. Builds need to be deterministic. The options should just be OFF or ON, with defaults that make sense for a developer usecase and work in our devcontainers.

So:

`CCCL_ENABLE_CLANG_TIDY` -> default ON. This option should only be exposed in developer/CI builds which all should be using our devcontainers with the supported tooling available.

When OFF, we don't check for clang-tidy, and we skip adding the clang-tidy targets.

CMakeLists.txt
alliepiper: How are we ending up with multiple concurrent CCCLs?

Jacobfaib: Our CI isn't, but it's possible in the wild. For example, a project whose unrelated dependencies which depend on different versions of CCCL. `find_package(CCCL)` would try it's best to reuse these of course, but this can be defeated with `CMAKE_PREFIX_PATH` shenanigans and unsetting certain find-package variables.

It happened quite often in legate.

In any case, I switched to `CCCL_SOURCE_DIR`. clang-tidy is developer only so highly unlikely this will be a problem.

alliepiper: > unsetting certain find-package variables

Those people get what they deserve 🤣

Everything I've heard about legate sounds like the wild west tbh

> clang-tidy is developer only so highly unlikely this will be a problem

Agreed. This is the convention used throughout the codebase currently, so if we change this, we'd need to systemically update.

PR #8075, opened 2026-03-18
Extend `fast_mod_div` to any type

fbusato:
## Description

Currently `cuda::fast_mod_div` is limited by the largest integral type on a given platform / 2.
This PR extents the functionality to any integral type.

### DIFF
--- a/libcudacxx/include/cuda/__cmath/fast_modulo_division.h
+++ b/libcudacxx/include/cuda/__cmath/fast_modulo_division.h
@@ -77,7 +77,7 @@ public:
       const auto __k = __num_bits + __shift; // k: [N, 2*N-2]
       // __multiplier: ceil(2^k / divisor)
       //   computed as 2^k / divisor + (remainder != 0)
-      const auto __pow2_div = __divmod_pow2(__k, __u_divisor); 
+      const auto __pow2_div = __divmod_pow2(__k, __u_divisor);
       __multiplier          = __pow2_div.first + (__pow2_div.second != 0);
     }
     else
@@ -192,13 +192,13 @@ private:
     // If the result is >= __divisor, the divisor is subtracted and 1-bit is recorded in the quotient.
     for (int __bit = __power; __bit >= 0; --__bit)
     {
-      // __carry only matters for unsigned types with large divisors (MSB set: >= 2^(N-1))
-      const bool __carry = (__remainder >> (__num_bits - 1)) != 0; // __remainder / 2^N != 0
+      // __carry_flag only matters for unsigned types with large divisors (MSB set: >= 2^(N-1))
+      const bool __carry_flag = (__remainder >> (__num_bits - 1)) != 0; // __remainder / 2^N != 0
       __remainder <<= 1;
       __remainder |= (__bit == __power); // append the remainder bit
-      const bool __quotient_bit = __carry || (__remainder >= __divisor);
+      const bool __quotient_bit = __carry_flag || (__remainder >= __divisor);
       __quotient <<= 1; // shift
-      __quotient |= (__quotient_bit); // append the quotient bit
+      __quotient |= unsigned{__quotient_bit}; // append the quotient bit
       if (__quotient_bit)
       {
         __remainder -= __divisor;


libcudacxx/include/cuda/__cmath/fast_modulo_division.h
```diff
@@ -21,19 +21,17 @@
 #  pragma system_header
 #endif // no system header
 
-#include <cuda/__cmath/ceil_div.h>
 #include <cuda/__cmath/ilog.h>
 #include <cuda/__cmath/mul_hi.h>
 #include <cuda/__cmath/pow2.h>
+#include <cuda/std/__limits/numeric_limits.h>
```
miscco, line 27: We conventionally only include `<cuda/std/limits>`because that is a single feature header

libcudacxx/include/cuda/__cmath/fast_modulo_division.h
```diff
@@ -49,110 +47,166 @@ _CCCL_BEGIN_NAMESPACE_CUDA
...
     {
-      __shift            = ::cuda::ceil_ilog2(__divisor) - 1; // is_pow2(x) ? log2(x) - 1 : log2(x)
-      auto __k           = __num_bits_v<_Tp> + __shift; // k: [N, 2*N-2]
-      auto __multiplier1 = ::cuda::ceil_div(__larger_t{1} << __k, __divisor); // ceil(2^k / divisor)
-      __multiplier       = static_cast<__unsigned_t>(__multiplier1);
+      __shift        = ::cuda::ceil_ilog2(__divisor) - 1; // is_pow2(x) ? log2(x) : ceil(log2(x))
+      const auto __k = __num_bits + __shift; // k: [N, 2*N-2]
+      // __multiplier: ceil(2^k / divisor)
+      //   computed as 2^k / divisor + (remainder != 0)
+      const auto __pow2_div = __divmod_pow2(__k, __u_divisor); //
```
miscco: Important: I did not find that in the docs, since when is the instruction available?
```suggestion
      const auto __pow2_div = __divmod_pow2(__k, __u_divisor); 
```

Also the content of this file are available in a non-cuda compiler, so we need to make sure this works in all supported platforms

libcudacxx/test/libcudacxx/cuda/cmath/fast_modulo_division.pass.cpp
```diff
@@ -3,7 +3,7 @@
 // Part of the libcu++ Project, under the Apache License v2.0 with LLVM Exceptions.
 // See https://llvm.org/LICENSE.txt for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
-// SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES.
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES.
```
miscco: I believe we need to change the year range not the year
```suggestion
// SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES.
```

libcudacxx/include/cuda/__cmath/fast_modulo_division.h
fbusato: the function is implemented below

libcudacxx/include/cuda/__cmath/fast_modulo_division.h
fbusato: `cuda/std/limits` also includes `numeric_limits_ext.h` which is pretty large. I would prefer keeping the current header

miscco: No we already have a fix for that, please use the proper header here

davebayer: I am with @fbusato here, we should change that once the PR is merged, we will need to change it in far more places

libcudacxx/include/cuda/__cmath/fast_modulo_division.h
```diff
@@ -49,110 +47,166 @@ _CCCL_BEGIN_NAMESPACE_CUDA
...
 
   using __unsigned_t = ::cuda::std::make_unsigned_t<_Tp>;
 
 public:
   fast_mod_div() = delete;
 
+  //! @brief Constructs a fast_mod_div object from a divisor value
+  //! @param[in] __divisor1 The divisor value, must be positive
+  //! @pre \p __divisor1 must be positive
   _CCCL_API explicit fast_mod_div(_Tp __divisor1) noexcept
```
davebayer, line 67: Shouldn't I be able to create this type at compile time?
```suggestion
  _CCCL_API explicit constexpr fast_mod_div(_Tp __divisor1) noexcept
```

fbusato: Technically possible but makes little sense in practice. If we have a `constexpr` divisor, the functionality is not useful at all.

libcudacxx/include/cuda/__cmath/fast_modulo_division.h
miscco: I thought we found that with the forward declaration we can avoid the cost and `<cuda/std/limits>` becomes quite cheap?

fbusato:
> I thought we found that with the forward declaration we can avoid the cost and <cuda/std/limits> becomes quite cheap?

@davebayer as you fixed the problem, what currently is the best way to use `numeric_limits` only for integral types?

davebayer:
> @davebayer as you fixed the problem, what currently is the best way to use `numeric_limits` only for integral types?

Just include the `<cuda/std/__limits/numeric_limits.h>` header, it omits the extended fp types. I may collide them into 1 header in the future

PR #8080, opened 2026-03-18
Implement `is_root_rank` and `is_part_of` queries for groups

davebayer:
This PR adds:
- `x.is_root_rank(group)`: queries whether the `x`'s rank in the `group` is `0`
- `x.is_part_of(group)`: queries whether the `x` is part of the `group` (currently `x` is always part of the `group`)

### DIFF
--- a/cudax/test/hierarchy/group.cu
+++ b/cudax/test/hierarchy/group.cu
@@ -85,14 +85,14 @@ __device__ T sum(cudax::this_cluster<Hierarchy> group, T (&array)[N])
   NV_IF_TARGET(NV_PROVIDES_SM_90, ({
                  const auto dsmem = static_cast<T*>(__cluster_map_shared_rank(&smem.cluster_scratch, 0));
 
-                 const auto is_root_thread = (cuda::gpu_thread.rank(group) == 0);
-                 if (is_root_thread)
+                 if (cuda::gpu_thread.is_root_rank(group))
                  {
-                   smem.cluster_scratch = 0;
+                   smem.cluster_scratch = result;
                  }
                  group.sync_aligned();
 
-                 if (cuda::gpu_thread.rank(cuda::block, group.hierarchy()) == 0 && !is_root_thread)
+                 cudax::this_block this_block{group.hierarchy()};
+                 if (cuda::gpu_thread.is_root_rank(this_block) && !cuda::gpu_thread.is_root_rank(group))
                  {
                    [[maybe_unused]] unsigned old;
                    asm volatile("atom.relaxed.cluster.shared::cluster.add.s32 %0, [%1], %2;"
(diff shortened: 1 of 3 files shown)


jrhemstad:
Is `is_root_rank` necessary if we provide a `cg::invoke_one` equivalent API that just says "invoke this on one thread, I don't care which one"

In other words, I'm dubious of the need for a `is_root_rank` API.

davebayer:
> Is `is_root_rank` necessary if we provide a `cg::invoke_one` equivalent API that just says "invoke this on one thread, I don't care which one"
> 
> In other words, I'm dubious of the need for a `is_root_rank` API.

It is important for cooperative algorithms like reduction, when only 1 thread returns the valid result. I've updated the PR to use the query in more places in the test, so the motivation is more obvious.

If we find out later that it's not needed, I will happily remove it

cudax/test/hierarchy/group.cu
```diff
@@ -85,14 +85,14 @@ __device__ T sum(cudax::this_cluster<Hierarchy> group, T (&array)[N])
   NV_IF_TARGET(NV_PROVIDES_SM_90, ({
                  const auto dsmem = static_cast<T*>(__cluster_map_shared_rank(&smem.cluster_scratch, 0));
 
-                 const auto is_root_thread = (cuda::gpu_thread.rank(group) == 0);
-                 if (is_root_thread)
+                 if (cuda::gpu_thread.is_root_rank(group))
```
bernhardmgruber, line 88: In warpspeed, we called this the "leader".

```suggestion
                 if (cuda::gpu_thread.is_leader(group))
```
I think this term could fit here as well.

libcudacxx/include/cuda/__hierarchy/hierarchy_level_base.h
```diff
@@ -391,6 +391,20 @@ struct hierarchy_level_base
...
+  [[nodiscard]] _CCCL_API static constexpr bool is_root_rank(const _Group& __group) noexcept
+  {
+    return _Level::rank(__group) == 0;
+  }
+
+  _CCCL_TEMPLATE(class _Group)
+  _CCCL_REQUIRES(::cuda::experimental::__is_this_hierarchy_group_v<_Group>)
+  [[nodiscard]] _CCCL_API static constexpr bool is_part_of(const _Group& __group) noexcept
+  {
+    return true;
```
pciolkosz, line 406: Hmm, will this return true for something like `cuda::grid.is_part_of(this_thread)`?

pciolkosz (approved):
> If we find out later that it's not needed, I will happily remove it

I think its fine to leave it in for now in experimental version, but it might be true long term we need `invoke_one` instead

cudax/test/hierarchy/group.cu
davebayer: Hm, I proceed from MPI, where rank `0` is called the `root`, so I wanted to use this broadly known name. I like that `is_leader` is shorter, I was also considering `is_root`.

I will open on whether this query is needed and what should be the final name

libcudacxx/include/cuda/__hierarchy/hierarchy_level_base.h
davebayer: Yes for now, I will fix this for all group queries in a separate PR

PR #8082, opened 2026-03-18
Use `cuda::is_power_of_2` instead of `cuda::std::has_single_bit`

davebayer:
We often use `cuda::std::has_single_bit` to check whether a value is a power of 2, however the name is not straight forward and requires the type to be `unsigned`. I think we should rather use `cuda::is_power_of_2` instead in many of the cases, which much better describes what we are actually trying to do


fbusato (approved):
agree! `has_single_bit` is verbose

PR #8091, opened 2026-03-18
Implement `cuda::__is_valid_alignment`

davebayer:
It's a bit more robust and better describes the intention.

### DIFF
--- a/libcudacxx/test/libcudacxx/cuda/memory/is_valid_alignment.pass.cpp
+++ b/libcudacxx/test/libcudacxx/cuda/memory/is_valid_alignment.pass.cpp
@@ -33,7 +33,9 @@ __host__ __device__ constexpr void test()
 }
 
 struct alignas(512) OverAligned
-{};
+{
+  char data_[512];
+};
 
 __host__ __device__ constexpr bool test()
 {


cub/cub/device/dispatch/kernels/kernel_transform.cuh
```diff
@@ -63,7 +64,7 @@ namespace detail::transform
 template <typename T>
 _CCCL_HOST_DEVICE _CCCL_FORCEINLINE const char* round_down_ptr(const T* ptr, unsigned alignment)
 {
-  _CCCL_ASSERT(::cuda::is_power_of_two(alignment), "");
+  _CCCL_ASSERT(::cuda::__is_valid_alignment<T>(alignment), "");
```
fbusato: should not `is_sufficiently_aligned` better here?

libcudacxx/include/cuda/__memory/align_down.h
```diff
@@ -41,14 +42,13 @@
...
+[[nodiscard]] _CCCL_API _Tp* align_down(_Tp* __ptr, ::cuda::std::size_t __alignment) noexcept
 {
   using ::cuda::std::uintptr_t;
-  _CCCL_ASSERT(::cuda::is_power_of_two(__alignment), "alignment must be a power of two");
+  _CCCL_ASSERT(::cuda::__is_valid_alignment<_Tp>(__alignment), "invalid alignment");
   if constexpr (!::cuda::std::is_void_v<_Tp>)
   {
-    _CCCL_ASSERT(__alignment >= alignof(_Tp), "wrong alignment");
-    _CCCL_ASSERT(reinterpret_cast<uintptr_t>(__ptr) % alignof(_Tp) == 0, "ptr is not aligned");
+    _CCCL_ASSERT(::cuda::is_aligned(__ptr, alignof(_Tp)), "ptr is not aligned");
```
fbusato: `::cuda::is_aligned(__ptr)` could even be a new overload

libcudacxx/include/cuda/__memory/aligned_size.h
```diff
@@ -31,7 +31,7 @@ _CCCL_BEGIN_NAMESPACE_CUDA
 template <::cuda::std::size_t _Alignment>
 struct aligned_size_t
 {
-  static_assert(::cuda::is_power_of_two(_Alignment), "alignment must be a power of two");
+  static_assert(::cuda::__is_valid_alignment(_Alignment), "alignment must be a power of two");
```
fbusato: update the comment

libcudacxx/include/cuda/__memory/is_aligned.h
```diff
@@ -40,7 +40,7 @@ _CCCL_BEGIN_NAMESPACE_CUDA
 
 [[nodiscard]] _CCCL_API inline bool is_aligned(const void* __ptr, ::cuda::std::size_t __alignment) noexcept
 {
-  _CCCL_ASSERT(::cuda::is_power_of_two(__alignment), "alignment must be a power of two");
+  _CCCL_ASSERT(::cuda::__is_valid_alignment(__alignment), "alignment must be a power of two");
```
fbusato: ditto

libcudacxx/include/cuda/__memory/ptr_alignment.h
```diff
@@ -40,7 +40,7 @@ _CCCL_BEGIN_NAMESPACE_CUDA
 __ptr_alignment(const void* __ptr, ::cuda::std::size_t __max_alignment) noexcept
 {
   _CCCL_ASSERT(__ptr != nullptr, "ptr_alignment requires a non-null pointer");
-  _CCCL_ASSERT(::cuda::is_power_of_two(__max_alignment), "max_alignment must be a power of two");
+  _CCCL_ASSERT(::cuda::__is_valid_alignment(__max_alignment), "max_alignment must be a power of two");
```
fbusato: ditto

libcudacxx/include/cuda/std/__memory/align.h
```diff
@@ -37,7 +37,7 @@ _CCCL_BEGIN_NAMESPACE_CUDA_STD
 
 _CCCL_API inline void* align(size_t __alignment, size_t __size, void*& __ptr, size_t& __space)
 {
-  _CCCL_ASSERT(::cuda::is_power_of_two(__alignment), "cuda::std::align: alignment must be a power of two!");
+  _CCCL_ASSERT(::cuda::__is_valid_alignment(__alignment), "cuda::std::align: alignment must be a power of two!");
```
fbusato: ditto

cub/cub/device/dispatch/kernels/kernel_transform.cuh
bernhardmgruber: Critical: firstly, this is somewhat inaccurate, because `alignment` has nothing to do with `T` or `T`s alignment. This code checks whether `alignment` is a valid alignment value (a power of two) independent of any concrete type. Consider using:

```suggestion
  _CCCL_ASSERT(::cuda::__is_valid_alignment<void>(alignment), "");
```

Secondly, `aligment` can be smaller than `alignof(T)` here, which is fine. `__is_valid_alignment` will fail for that case.

davebayer: I will merge #8092 before this one

libcudacxx/include/cuda/__memory/align_down.h
davebayer: I think I would prefer the explicit `alignof(_Tp)` parameter

thrust/thrust/mr/pool_options.h
```diff
@@ -98,7 +99,7 @@ struct pool_options
     {
       return false;
     }
-    if (alignment != 0 && !::cuda::is_power_of_two(alignment))
+    if (::cuda::__is_valid_alignment(alignment))
```
dkolsen-pgi, line 102: @davebayer 

This change reverses the condition!  Before the `if` condition was true if the alignment was invalid.  Now the `if` condition is true if the alignment is valid.

This completely breaks NVC++ stdpar.  Every single test of ours is failing on `assert(m_options.validate());` in the `disjoint_unsynchronized_pool_resource` constructor in `<thrust/mr/disjoint_pool.h>`.  How did this make it through CI?  Even the basic smoke NVC++ stdpar smoke test should have caught this.

davebayer: I'm sorry, I completely missed that. #8124 should fix this. It might have passed the CI, because it uses `assert` (not `_CCCL_ASSERT`) and we build with `-O3`, which I believe forced `NDEBUG` to be defined. I will open an issue for further investigation.

bernhardmgruber: Nice! I completely missed that :D Thanks for reporting back so quickly!

miscco: @dkolsen-pgi the issue is that the smoke tests are currently compile only 
https://github.com/NVIDIA/cccl/actions/runs/23283614366/job/67702175012

I believe we might want to reconsider that given that the situation with the GPU runners has improved considerably

PR #8092, opened 2026-03-18
Remove `cub::detail::transform::round_down_ptr`

davebayer:
We can just use `cuda::align_down` instead

### DIFF
--- a/cub/cub/device/dispatch/kernels/kernel_transform.cuh
+++ b/cub/cub/device/dispatch/kernels/kernel_transform.cuh
@@ -25,6 +25,7 @@
 #include <cuda/__cmath/round_up.h>
 #include <cuda/__device/arch_id.h>
 #include <cuda/__memcpy_async/elect_one.h>
+#include <cuda/__memory/align_down.h>
 #include <cuda/__memory/aligned_size.h>
 #include <cuda/__ptx/instructions/cp_async_bulk.h>
 #include <cuda/__ptx/instructions/elect_sync.h>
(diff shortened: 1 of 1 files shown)


cub/cub/device/dispatch/kernels/kernel_transform.cuh
```diff
@@ -422,7 +415,7 @@ struct aligned_base_ptr
 template <typename T>
 _CCCL_HOST_DEVICE auto make_aligned_base_ptr(const T* ptr, int alignment) -> aligned_base_ptr<T>
 {
-  const char* base_ptr = round_down_ptr(ptr, alignment);
+  const char* base_ptr = ::cuda::align_down(ptr, alignment);
```
fbusato: the pointer type is `char*` not the input pointer type

bernhardmgruber: I am not sure we can call `::cuda::align_down` here, because it returns a `T*` and the result of aligning down a `T*` to a 16 or 128 byte address may not be a valid start of a `T`.

Consider an `int3*` at address `0x.....14`. Aligning it down to `0x.....10` as a `T*` does not point to the start of a `T`, since you moved the pointer 4 bytes, but the size of a `T` is 12.

Maybe it compiles fine, since we immediately turn the `T*` into a `char*` here, but I am not sure.

cub/cub/device/dispatch/kernels/kernel_transform.cuh
```diff
@@ -422,8 +415,9 @@ struct aligned_base_ptr
 template <typename T>
 _CCCL_HOST_DEVICE auto make_aligned_base_ptr(const T* ptr, int alignment) -> aligned_base_ptr<T>
 {
-  const char* base_ptr = round_down_ptr(ptr, alignment);
-  return aligned_base_ptr<T>{base_ptr, static_cast<int>(reinterpret_cast<const char*>(ptr) - base_ptr)};
+  const auto raw_ptr  = reinterpret_cast<const char*>(ptr);
+  const auto base_ptr = ::cuda::align_down(raw_ptr, alignment);
```
bernhardmgruber, line 418: Critical: this will not work when `alignment < alignof(T)`, which is a valid use case here.

PR #8097, opened 2026-03-18
Add fixed-size segment env overloads for DeviceSegmentedReduce

### LINKED ISSUE #7550: Add env-based API for cub::DeviceSegmentedReduce
Add env overloads for:

  - [ ] `cub::DeviceSegmentedReduce::Reduce`
  - [ ] `cub::DeviceSegmentedReduce::Sum`
  - [ ] `cub::DeviceSegmentedReduce::Min`
  - [ ] `cub::DeviceSegmentedReduce::Max`
  - [ ] `cub::DeviceSegmentedReduce::ArgMin`
  - [ ] `cub::DeviceSegmentedReduce::ArgMax`

gonidelis:
Split 2/2 for env `SegmentedReduce`

- [x] Adds fixed-size overloads. Ideally would be merged after #7795 
- [x] Merge first: #8335

closes #7550

### DIFF
--- a/cub/test/catch2_test_device_segmented_reduce_env_api.cu
+++ b/cub/test/catch2_test_device_segmented_reduce_env_api.cu
@@ -476,7 +476,7 @@ C2H_TEST("cub::DeviceSegmentedReduce::Reduce fixed-size env-based API", "[segmen
 
   if (error != cudaSuccess)
   {
-    std::cerr << "cub::DeviceSegmentedReduce::Reduce (fixed-size) failed with status: " << error << std::endl;
+    std::cerr << "cub::DeviceSegmentedReduce::Reduce (fixed-size) failed with status: " << error << '\n';
   }
   // example-end fixed-size-segmented-reduce-reduce-env
   stream.sync();
@@ -502,7 +502,7 @@ C2H_TEST("cub::DeviceSegmentedReduce::Sum fixed-size env-based API", "[segmented
 
   if (error != cudaSuccess)
   {
-    std::cerr << "cub::DeviceSegmentedReduce::Sum (fixed-size) failed with status: " << error << std::endl;
+    std::cerr << "cub::DeviceSegmentedReduce::Sum (fixed-size) failed with status: " << error << '\n';
   }
   // example-end fixed-size-segmented-reduce-sum-env
   stream.sync();
@@ -528,7 +528,7 @@ C2H_TEST("cub::DeviceSegmentedReduce::Min fixed-size env-based API", "[segmented
 
   if (error != cudaSuccess)
   {
-    std::cerr << "cub::DeviceSegmentedReduce::Min (fixed-size) failed with status: " << error << std::endl;
+    std::cerr << "cub::DeviceSegmentedReduce::Min (fixed-size) failed with status: " << error << '\n';
   }
   // example-end fixed-size-segmented-reduce-min-env
   stream.sync();
@@ -554,7 +554,7 @@ C2H_TEST("cub::DeviceSegmentedReduce::Max fixed-size env-based API", "[segmented
 
   if (error != cudaSuccess)
   {
-    std::cerr << "cub::DeviceSegmentedReduce::Max (fixed-size) failed with status: " << error << std::endl;
+    std::cerr << "cub::DeviceSegmentedReduce::Max (fixed-size) failed with status: " << error << '\n';
   }
   // example-end fixed-size-segmented-reduce-max-env
   stream.sync();
@@ -579,7 +579,7 @@ C2H_TEST("cub::DeviceSegmentedReduce::ArgMin fixed-size env-based API", "[segmen
 
   if (error != cudaSuccess)
   {
-    std::cerr << "cub::DeviceSegmentedReduce::ArgMin (fixed-size) failed with status: " << error << std::endl;
+    std::cerr << "cub::DeviceSegmentedReduce::ArgMin (fixed-size) failed with status: " << error << '\n';
   }
   // example-end fixed-size-segmented-reduce-argmin-env
   stream.sync();
@@ -604,7 +604,7 @@ C2H_TEST("cub::DeviceSegmentedReduce::ArgMax fixed-size env-based API", "[segmen
 
   if (error != cudaSuccess)
   {
-    std::cerr << "cub::DeviceSegmentedReduce::ArgMax (fixed-size) failed with status: " << error << std::endl;
+    std::cerr << "cub::DeviceSegmentedReduce::ArgMax (fixed-size) failed with status: " << error << '\n';
   }
   // example-end fixed-size-segmented-reduce-argmax-env
   stream.sync();


cub/cub/device/device_segmented_reduce.cuh
```diff
@@ -306,6 +362,111 @@ struct DeviceSegmentedReduce
...
+      ::cuda::std::execution::__query_result_or_t<requirements_t,
+                                                  ::cuda::execution::determinism::__get_determinism_t,
+                                                  ::cuda::execution::determinism::run_to_run_t>;
+
+    static_assert(!::cuda::std::is_same_v<requested_determinism_t, ::cuda::execution::determinism::gpu_to_gpu_t>,
+                  "gpu_to_gpu determinism is not supported for device segmented reductions ");
+
+    return detail::dispatch_with_env(
+      env, [&]([[maybe_unused]] auto tuning, void* d_temp_storage, size_t& temp_storage_bytes, cudaStream_t stream) {
+        return detail::reduce::DispatchFixedSizeSegmentedReduce<InputIteratorT, OutputIteratorT, int, ReductionOpT, T>::
```
bernhardmgruber: Important: please retain the comment:
```suggestion
    // `offset_t` a.k.a `SegmentSizeT` is fixed to `int` type now, but later can be changed to accept
    // integral constant or larger integral types
    using offset_t = int;
        return detail::reduce::DispatchFixedSizeSegmentedReduce<InputIteratorT, OutputIteratorT, offset_t, ReductionOpT, T>::
```

cub/cub/device/device_segmented_reduce.cuh
```diff
@@ -688,6 +768,87 @@ struct DeviceSegmentedReduce
...
+                                                  ::cuda::execution::determinism::__get_determinism_t,
+                                                  ::cuda::execution::determinism::run_to_run_t>;
+
+    static_assert(!::cuda::std::is_same_v<requested_determinism_t, ::cuda::execution::determinism::gpu_to_gpu_t>,
+                  "gpu_to_gpu determinism is not supported for device segmented reductions ");
+
+    return detail::dispatch_with_env(
+      env, [&]([[maybe_unused]] auto tuning, void* d_temp_storage, size_t& temp_storage_bytes, cudaStream_t stream) {
+        return detail::reduce::
+          DispatchFixedSizeSegmentedReduce<InputIteratorT, OutputIteratorT, int, ::cuda::std::plus<output_t>, output_t>::
```
bernhardmgruber: Same here.

cub/cub/device/device_segmented_reduce.cuh
```diff
@@ -899,6 +1066,90 @@ struct DeviceSegmentedReduce
...
+    static_assert(!::cuda::std::is_same_v<requested_determinism_t, ::cuda::execution::determinism::gpu_to_gpu_t>,
+                  "gpu_to_gpu determinism is not supported for device segmented reductions ");
+
+    static_assert(::cuda::std::numeric_limits<input_t>::is_specialized,
+                  "numeric_limits must be specialized for the input value type");
+
+    return detail::dispatch_with_env(
+      env, [&]([[maybe_unused]] auto tuning, void* d_temp_storage, size_t& temp_storage_bytes, cudaStream_t stream) {
+        return detail::reduce::
+          DispatchFixedSizeSegmentedReduce<InputIteratorT, OutputIteratorT, int, ::cuda::minimum<input_t>, input_t>::
```
bernhardmgruber: Same

cub/cub/device/device_segmented_reduce.cuh
```diff
@@ -899,6 +1066,90 @@ struct DeviceSegmentedReduce
...
+            Dispatch(d_temp_storage,
+                     temp_storage_bytes,
+                     d_in,
+                     d_out,
+                     num_segments,
+                     segment_size,
+                     ::cuda::minimum<input_t>{},
```
bernhardmgruber: Critical: this is a breaking change to the non-env overload. Please revert to:
```suggestion
                     ::cuda::minimum<>{},
```

cub/cub/device/device_segmented_reduce.cuh
```diff
@@ -688,6 +768,87 @@ struct DeviceSegmentedReduce
...
+      env, [&]([[maybe_unused]] auto tuning, void* d_temp_storage, size_t& temp_storage_bytes, cudaStream_t stream) {
+        return detail::reduce::
+          DispatchFixedSizeSegmentedReduce<InputIteratorT, OutputIteratorT, int, ::cuda::std::plus<output_t>, output_t>::
+            Dispatch(d_temp_storage,
+                     temp_storage_bytes,
+                     d_in,
+                     d_out,
+                     num_segments,
+                     segment_size,
+                     ::cuda::std::plus<output_t>{},
```
bernhardmgruber: Critical: this is breaking change compared with the env-overload. Please use:

```suggestion
                     ::cuda::std::plus{},
```

cub/cub/device/device_segmented_reduce.cuh
```diff
@@ -1159,6 +1414,118 @@ struct DeviceSegmentedReduce
...
+  //!   @rst
+  //!   **[optional]** Execution environment. Default is ``cuda::std::execution::env{}``.
+  //!   @endrst
+  template <typename InputIteratorT, typename OutputIteratorT, typename EnvT = ::cuda::std::execution::env<>>
+  [[nodiscard]] CUB_RUNTIME_FUNCTION static cudaError_t
+  ArgMin(InputIteratorT d_in, OutputIteratorT d_out, ::cuda::std::int64_t num_segments, int segment_size, EnvT env = {})
+  {
+    _CCCL_NVTX_RANGE_SCOPE("cub::DeviceSegmentedReduce::ArgMin");
+
+    using offset_t       = int;
```
bernhardmgruber:
```suggestion
    // `offset_t` a.k.a `SegmentSizeT` is fixed to `int` type now, but later can be changed to accept
    // integral constant or larger integral types
    using offset_t       = int;
```

cub/cub/device/device_segmented_reduce.cuh
```diff
@@ -1159,6 +1414,118 @@ struct DeviceSegmentedReduce
...
+
+    using arg_index_input_iterator_t =
+      THRUST_NS_QUALIFIER::transform_iterator<detail::reduce::generate_idx_value<InputIteratorT, output_value_t>,
+                                              THRUST_NS_QUALIFIER::counting_iterator<::cuda::std::int64_t>>;
+
+    arg_index_input_iterator_t d_indexed_in = THRUST_NS_QUALIFIER::make_transform_iterator(
+      THRUST_NS_QUALIFIER::counting_iterator<::cuda::std::int64_t>{0},
+      detail::reduce::generate_idx_value<InputIteratorT, output_value_t>(d_in, segment_size));
+
+    init_t initial_value{accum_t(1, ::cuda::std::numeric_limits<input_value_t>::max())};
```
bernhardmgruber: Important: please refactor this into a common function that is shared with the non-env `ArgMin` overload. To reduce duplication.

cub/cub/device/device_segmented_reduce.cuh
```diff
@@ -1359,6 +1731,89 @@ struct DeviceSegmentedReduce
...
+      env, [&]([[maybe_unused]] auto tuning, void* d_temp_storage, size_t& temp_storage_bytes, cudaStream_t stream) {
+        return detail::reduce::
+          DispatchFixedSizeSegmentedReduce<InputIteratorT, OutputIteratorT, int, ::cuda::maximum<input_t>, input_t>::
+            Dispatch(d_temp_storage,
+                     temp_storage_bytes,
+                     d_in,
+                     d_out,
+                     num_segments,
+                     segment_size,
+                     ::cuda::maximum<input_t>{},
```
bernhardmgruber:
```suggestion
                     ::cuda::maximum<>{},
```

cub/cub/device/device_segmented_reduce.cuh
```diff
@@ -1359,6 +1731,89 @@ struct DeviceSegmentedReduce
...
+
+    static_assert(!::cuda::std::is_same_v<requested_determinism_t, ::cuda::execution::determinism::gpu_to_gpu_t>,
+                  "gpu_to_gpu determinism is not supported for device segmented reductions ");
+    static_assert(::cuda::std::numeric_limits<input_t>::is_specialized,
+                  "numeric_limits must be specialized for the input value type");
+
+    return detail::dispatch_with_env(
+      env, [&]([[maybe_unused]] auto tuning, void* d_temp_storage, size_t& temp_storage_bytes, cudaStream_t stream) {
+        return detail::reduce::
+          DispatchFixedSizeSegmentedReduce<InputIteratorT, OutputIteratorT, int, ::cuda::maximum<input_t>, input_t>::
```
bernhardmgruber:
```suggestion
          DispatchFixedSizeSegmentedReduce<InputIteratorT, OutputIteratorT, int, ::cuda::maximum<>, input_t>::
```

cub/cub/device/device_segmented_reduce.cuh
```diff
@@ -1616,6 +2075,118 @@ struct DeviceSegmentedReduce
...
+  //!   @rst
+  //!   **[optional]** Execution environment. Default is ``cuda::std::execution::env{}``.
+  //!   @endrst
+  template <typename InputIteratorT, typename OutputIteratorT, typename EnvT = ::cuda::std::execution::env<>>
+  [[nodiscard]] CUB_RUNTIME_FUNCTION static cudaError_t
+  ArgMax(InputIteratorT d_in, OutputIteratorT d_out, ::cuda::std::int64_t num_segments, int segment_size, EnvT env = {})
+  {
+    _CCCL_NVTX_RANGE_SCOPE("cub::DeviceSegmentedReduce::ArgMax");
+
+    using offset_t       = int;
```
bernhardmgruber: Please also retain the original spelling and and comments:
```suggestion
    // Using common iterator value type is a breaking change, see:
    // https://github.com/NVIDIA/cccl/pull/414#discussion_r1330632615
    using OverrideOffsetT = int; // detail::common_iterator_value_t<BeginOffsetIteratorT, EndOffsetIteratorT>;
```

cub/cub/device/device_segmented_reduce.cuh
```diff
@@ -1616,6 +2075,118 @@ struct DeviceSegmentedReduce
...
+
+    using arg_index_input_iterator_t =
+      THRUST_NS_QUALIFIER::transform_iterator<detail::reduce::generate_idx_value<InputIteratorT, output_value_t>,
+                                              THRUST_NS_QUALIFIER::counting_iterator<::cuda::std::int64_t>>;
+
+    arg_index_input_iterator_t d_indexed_in = THRUST_NS_QUALIFIER::make_transform_iterator(
+      THRUST_NS_QUALIFIER::counting_iterator<::cuda::std::int64_t>{0},
+      detail::reduce::generate_idx_value<InputIteratorT, output_value_t>(d_in, segment_size));
+
+    init_t initial_value{accum_t(1, ::cuda::std::numeric_limits<input_value_t>::lowest())};
```
bernhardmgruber: Again, please refactor into a common function.

cub/cub/device/device_segmented_reduce.cuh
```diff
@@ -74,6 +74,104 @@ CUB_NAMESPACE_BEGIN
 struct DeviceSegmentedReduce
 {
 private:
+  template <typename InputIteratorT, typename OutputIteratorT>
+  CUB_RUNTIME_FUNCTION static cudaError_t fixed_size_arg_min_dispatch(
```
bernhardmgruber: Great work! Now try to unify this function with `fixed_size_arg_max_dispatch`. There are two different tokens that matter ;)

cub/cub/device/device_segmented_reduce.cuh
```diff
@@ -402,11 +500,6 @@ public:
   //!    First appears in CUDA Toolkit 13.2.
   //!
   //! - Does not support binary reduction operators that are non-commutative.
-  //! - Provides "run-to-run" determinism for pseudo-associative reduction
-  //!   (e.g., addition of floating point types) on the same GPU device.
-  //!   However, results for pseudo-associative reduction may be inconsistent
-  //!   from one device to a another device of a different compute-capability
-  //!   because CUB can employ different tile-sizing for different architectures.
```
bernhardmgruber, line 409: Q: Why is this comment deleted? Doesn't the fixed-size segmented reduction provide run-to-run determinism? Was the comment wrong?

cub/cub/device/device_segmented_reduce.cuh
```diff
@@ -486,6 +579,113 @@ public:
...
+                                                  ::cuda::execution::determinism::run_to_run_t>;
+
+    static_assert(!::cuda::std::is_same_v<requested_determinism_t, ::cuda::execution::determinism::gpu_to_gpu_t>,
+                  "gpu_to_gpu determinism is not supported for device segmented reductions ");
+
+    // `offset_t` a.k.a `SegmentSizeT` is fixed to `int` type now, but later can be changed to accept
+    // integral constant or larger integral types
+    return detail::dispatch_with_env(
+      env, [&]([[maybe_unused]] auto tuning, void* d_temp_storage, size_t& temp_storage_bytes, cudaStream_t stream) {
+        return detail::reduce::DispatchFixedSizeSegmentedReduce<InputIteratorT, OutputIteratorT, int, ReductionOpT, T>::
```
bernhardmgruber: The added comment would be easier to associate with an alias like this:
```suggestion
    // `offset_t` a.k.a `SegmentSizeT` is fixed to `int` type now, but later can be changed to accept
    // integral constant or larger integral types
    using offset_t = int;
    return detail::dispatch_with_env(
      env, [&]([[maybe_unused]] auto tuning, void* d_temp_storage, size_t& temp_storage_bytes, cudaStream_t stream) {
        return detail::reduce::DispatchFixedSizeSegmentedReduce<InputIteratorT, OutputIteratorT, offset_t, ReductionOpT, T>::
```
Please add it.

cub/cub/device/device_segmented_reduce.cuh
bernhardmgruber, line 697: Important: I think we still need to constrain this overload to distinguish it from this other one:
```c++
  template <typename InputIteratorT, typename OutputIteratorT>
  CUB_RUNTIME_FUNCTION static cudaError_t
  Sum(void* d_temp_storage,
      size_t& temp_storage_bytes,
      InputIteratorT d_in,
      OutputIteratorT d_out,
      ::cuda::std::int64_t num_segments,
      int segment_size,
      cudaStream_t stream = 0)
```
Let's add an `enable_if` that checks that `InputIteratorT` is not a `void*`.

cub/cub/device/device_segmented_reduce.cuh
```diff
@@ -789,6 +987,89 @@ public:
...
+
+    static_assert(!::cuda::std::is_same_v<requested_determinism_t, ::cuda::execution::determinism::gpu_to_gpu_t>,
+                  "gpu_to_gpu determinism is not supported for device segmented reductions ");
+
+    // `offset_t` a.k.a `SegmentSizeT` is fixed to `int` type now, but later can be changed to accept
+    // integral constant or larger integral types
+    return detail::dispatch_with_env(
+      env, [&]([[maybe_unused]] auto tuning, void* d_temp_storage, size_t& temp_storage_bytes, cudaStream_t stream) {
+        return detail::reduce::
+          DispatchFixedSizeSegmentedReduce<InputIteratorT, OutputIteratorT, int, ::cuda::std::plus<>, output_t>::Dispatch(
```
bernhardmgruber:
```suggestion
    // `offset_t` a.k.a `SegmentSizeT` is fixed to `int` type now, but later can be changed to accept
    // integral constant or larger integral types
    using offset_t = int;
    return detail::dispatch_with_env(
      env, [&]([[maybe_unused]] auto tuning, void* d_temp_storage, size_t& temp_storage_bytes, cudaStream_t stream) {
        return detail::reduce::
          DispatchFixedSizeSegmentedReduce<InputIteratorT, OutputIteratorT, offset_t, ::cuda::std::plus<>, output_t>::Dispatch(
```

cub/cub/device/device_segmented_reduce.cuh
```diff
@@ -1108,6 +1390,92 @@ public:
...
+
+    static_assert(::cuda::std::numeric_limits<input_t>::is_specialized,
+                  "numeric_limits must be specialized for the input value type");
+
+    // `offset_t` a.k.a `SegmentSizeT` is fixed to `int` type now, but later can be changed to accept
+    // integral constant or larger integral types
+    return detail::dispatch_with_env(
+      env, [&]([[maybe_unused]] auto tuning, void* d_temp_storage, size_t& temp_storage_bytes, cudaStream_t stream) {
+        return detail::reduce::
+          DispatchFixedSizeSegmentedReduce<InputIteratorT, OutputIteratorT, int, ::cuda::minimum<>, input_t>::Dispatch(
```
bernhardmgruber:
```suggestion
    // `offset_t` a.k.a `SegmentSizeT` is fixed to `int` type now, but later can be changed to accept
    // integral constant or larger integral types
    using offset_t = int;
    return detail::dispatch_with_env(
      env, [&]([[maybe_unused]] auto tuning, void* d_temp_storage, size_t& temp_storage_bytes, cudaStream_t stream) {
        return detail::reduce::
          DispatchFixedSizeSegmentedReduce<InputIteratorT, OutputIteratorT, offset_t, ::cuda::minimum<>, input_t>::Dispatch(
```

cub/cub/device/device_segmented_reduce.cuh
```diff
@@ -1815,6 +2215,91 @@ public:
...
+                  "gpu_to_gpu determinism is not supported for device segmented reductions ");
+    static_assert(::cuda::std::numeric_limits<input_t>::is_specialized,
+                  "numeric_limits must be specialized for the input value type");
+
+    // `offset_t` a.k.a `SegmentSizeT` is fixed to `int` type now, but later can be changed to accept
+    // integral constant or larger integral types
+    return detail::dispatch_with_env(
+      env, [&]([[maybe_unused]] auto tuning, void* d_temp_storage, size_t& temp_storage_bytes, cudaStream_t stream) {
+        return detail::reduce::
+          DispatchFixedSizeSegmentedReduce<InputIteratorT, OutputIteratorT, int, ::cuda::maximum<>, input_t>::Dispatch(
```
bernhardmgruber:
```suggestion
    // `offset_t` a.k.a `SegmentSizeT` is fixed to `int` type now, but later can be changed to accept
    // integral constant or larger integral types
    using offset_t = int;
    return detail::dispatch_with_env(
      env, [&]([[maybe_unused]] auto tuning, void* d_temp_storage, size_t& temp_storage_bytes, cudaStream_t stream) {
        return detail::reduce::
          DispatchFixedSizeSegmentedReduce<InputIteratorT, OutputIteratorT, offset_t, ::cuda::maximum<>, input_t>::Dispatch(
```

cub/test/catch2_test_device_segmented_reduce_env.cu
```diff
@@ -139,8 +139,230 @@ TEST_CASE("Device segmented argmax works with default environment", "[segmented_
...
 
+TEST_CASE("Device fixed-size segmented reduce uses custom stream", "[segmented_reduce][device]")
+{
+  int num_segments = 2;
+  int segment_size = 3;
+  thrust::device_vector<int> d_in{8, 6, 7, 5, 3, 0};
+  thrust::device_vector<int> d_out(2);
+
+  cudaStream_t custom_stream;
+  REQUIRE(cudaSuccess == cudaStreamCreate(&custom_stream));
```
bernhardmgruber: Important: Please use `cuda::stream` instead of manual resource management

cub/cub/device/device_segmented_reduce.cuh
gonidelis: this is not an env overload

cub/test/catch2_test_device_segmented_reduce_env.cu
gonidelis, line 24: @NaderAlAwar  @bernhardmgruber  do you think SegmentedReduce makes sense that it has cuda graph (lid2) tests.

https://github.com/gonidelis/cccl/blob/3d76549593283b964cf9d5463db13c72b96bf802/cub/cub/device/dispatch/dispatch_segmented_reduce.cuh#L268-L270

there is a very specific case where multiple kernels passes are being launched `>2B segments` i think.

gonidelis: answer: Only SegmentedSort is incompatible with cuda graph, all the other Segmented algorithms should be good to be tested with lid2

cub/cub/device/device_segmented_reduce.cuh
gonidelis: what would an ambiguous call look like before adding the constrain?

cub/cub/device/device_segmented_reduce.cuh
```diff
@@ -701,7 +855,7 @@ public:
     using OffsetT = detail::common_iterator_value_t<BeginOffsetIteratorT, EndOffsetIteratorT>;
     using OutputT = detail::non_void_value_t<OutputIteratorT, detail::it_value_t<InputIteratorT>>;
     using init_t  = OutputT;
-    using op_t    = ::cuda::std::plus<>;
+    using op_t    = ::cuda::std::plus<detail::it_value_t<InputIteratorT>>;
```
bernhardmgruber: Critical: This is a change in behavior. Please revert.

cub/cub/device/device_segmented_reduce.cuh
bernhardmgruber: I think this should be ambiguous:
```c++
size_t size;
Sum(nullptr, size, 0, 0, 0, 0);
```
However, if we would disambiguate the user would probably get an error later, since you can't get an `it_value_t<int>`. So we can also probably omit the `enable_if`. I can't construct a valid call at the moment that would be ambigous.

c2h/include/c2h/catch2_test_helper.h
```diff
@@ -521,7 +521,19 @@ auto Equals(const THRUST_NS_QUALIFIER::detail::vector_base<T, Alloc>& expected)
...
 
 #include <cuda/std/tuple>
+#include <cuda/std/utility>
+
 _CCCL_BEGIN_NAMESPACE_CUDA_STD
+// Constrained to arithmetic types to avoid nvcc SFINAE issues with non-streamable types (e.g. iterators)
+template <typename T1,
+          typename T2,
+          // allow serializing only pairs with arithmetic members otherwise normal_iterator tests complain
+          enable_if_t<::cuda::std::is_arithmetic_v<T1>&& ::cuda::std::is_arithmetic_v<T2>, int> = 0>
```
bernhardmgruber: I don't understand why this is necessary. If `operator<<` is called for a pair of iterators, and you can't serialize an iterator, then why does removing this overload allow the program to compile? what other `operator<<` is chosen instead?

Which test fails to compile if you remove the `enable_if`?

gonidelis: ci was failing in `libcudacxx/test/libcudacxx/std/algorithms/alg.modifying/alg.partitions/pstl_partition_copy.cu`  in L#67 before i add this sfinae

bernhardmgruber: Thx! I understand now what's happening. Catch2 was trying to find an `operator<<(osteam, pair<some thrust iterators>)`. Before, it didn't fine one and just did not generate code for printing the value. With the above change (and no constraint) it found one, but then the implementation of `operator<<` did not compile.

The trick here is, as you noticed, to make the `operator<<` SFINAE friendly. That means, that the `operator<<` will SFINAE itself out if it knows that its implementation would not compile. This way, a library like Catch2 can detect whether it can use the `operator<<` (by testing whether a call to that `operator<<` would compile).

I refined the constraint to use:
```c++
template <typename T1,
          typename T2,
          ::cuda::std::void_t<decltype(::cuda::std::declval<::std::ostream>()
                                       << ::cuda::std::declval<T1>() << ::cuda::std::declval<T2>())>* = nullptr>
::std::ostream& operator<<(::std::ostream& os, const pair<T1, T2>& pair)
```
This tests whether `T1` and `T2` can we written to an `ostream` as part of the interface of `operator<<`. If they can't be written, the `operator<<` is SFINAE-ed out.

Issue #7998, opened 2026-03-12
Add env-based API for decomposer overloads of cub::DeviceRadixSort


gonidelis:
- [ ] SortPairs with DecomposerT (`[pointers, doublebuffer] x [begin/end_bit, no bit range]`)
  - [ ] SortPairs with DecomposerT (same 4 variants)
  - [ ] SortPairsDescending with DecomposerT (same 4 variants)
  - [ ] SortKeys with DecomposerT (same 4 variants)
  - [ ] SortKeysDescending with DecomposerT (same 4 variants)

PR #8099, opened 2026-03-19
Add env-based decomposer overloads for DeviceRadixSort descending variants

gonidelis:
closes https://github.com/NVIDIA/cccl/issues/7998

### DIFF
--- a/cub/test/catch2_test_device_radix_sort_env.cu
+++ b/cub/test/catch2_test_device_radix_sort_env.cu
@@ -51,6 +51,21 @@ struct custom_pair_key_t
 {
   int key;
   int payload;
+
+  __host__ __device__ friend bool operator==(const custom_pair_key_t& a, const custom_pair_key_t& b)
+  {
+    return a.key == b.key && a.payload == b.payload;
+  }
+
+  __host__ __device__ friend bool operator!=(const custom_pair_key_t& a, const custom_pair_key_t& b)
+  {
+    return !(a == b);
+  }
+
+  friend std::ostream& operator<<(std::ostream& os, const custom_pair_key_t& cpk)
+  {
+    return os << "{" << cpk.key << ", " << cpk.payload << "}";
+  }
 };
 
 struct keys_decomposer_t
@@ -227,14 +242,10 @@ TEST_CASE("Device radix sort pairs decomposer works with default environment", "
       static_cast<int>(keys_in.size()),
       pairs_decomposer_t{}));
 
-  c2h::host_vector<custom_pair_key_t> h_keys_out(keys_out);
-  REQUIRE(h_keys_out[0].key == 1);
-  REQUIRE(h_keys_out[1].key == 2);
-  REQUIRE(h_keys_out[2].key == 3);
-  c2h::host_vector<int> h_values_out(values_out);
-  REQUIRE(h_values_out[0] == 1);
-  REQUIRE(h_values_out[1] == 2);
-  REQUIRE(h_values_out[2] == 0);
+  c2h::device_vector<custom_pair_key_t> expected_keys{{1, 200}, {2, 300}, {3, 100}};
+  REQUIRE(keys_out == expected_keys);
+  c2h::device_vector<int> expected_values{1, 2, 0};
+  REQUIRE(values_out == expected_values);
 }
 
 TEST_CASE("Device radix sort pairs decomposer with bits works with default environment", "[radix_sort][device]")
@@ -256,14 +267,10 @@ TEST_CASE("Device radix sort pairs decomposer with bits works with default envir
       0,
       sizeof(int) * 8));
 
-  c2h::host_vector<custom_pair_key_t> h_keys_out(keys_out);
-  REQUIRE(h_keys_out[0].key == 1);
-  REQUIRE(h_keys_out[1].key == 2);
-  REQUIRE(h_keys_out[2].key == 3);
-  c2h::host_vector<int> h_values_out(values_out);
-  REQUIRE(h_values_out[0] == 1);
-  REQUIRE(h_values_out[1] == 2);
-  REQUIRE(h_values_out[2] == 0);
+  c2h::device_vector<custom_pair_key_t> expected_keys{{1, 200}, {2, 300}, {3, 100}};
+  REQUIRE(keys_out == expected_keys);
+  c2h::device_vector<int> expected_values{1, 2, 0};
+  REQUIRE(values_out == expected_values);
 }
 
 TEST_CASE("Device radix sort keys descending decomposer+bits works with default environment", "[radix_sort][device]")
@@ -727,14 +734,10 @@ TEST_CASE("Device radix sort pairs decomposer uses custom stream", "[radix_sort]
 
   stream.sync();
 
-  c2h::host_vector<custom_pair_key_t> h_keys_out(keys_out);
-  REQUIRE(h_keys_out[0].key == 1);
-  REQUIRE(h_keys_out[1].key == 2);
-  REQUIRE(h_keys_out[2].key == 3);
-  c2h::host_vector<int> h_values_out(values_out);
-  REQUIRE(h_values_out[0] == 1);
-  REQUIRE(h_values_out[1] == 2);
-  REQUIRE(h_values_out[2] == 0);
+  c2h::device_vector<custom_pair_key_t> expected_keys{{1, 200}, {2, 300}, {3, 100}};
+  REQUIRE(keys_out == expected_keys);
+  c2h::device_vector<int> expected_values{1, 2, 0};
+  REQUIRE(values_out == expected_values);
 }
 
 #if TEST_LAUNCH == 0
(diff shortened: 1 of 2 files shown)


cub/cub/device/device_radix_sort.cuh
```diff
@@ -1939,7 +1939,7 @@ public:
             typename NumItemsT,
             typename EnvT = ::cuda::std::execution::env<>,
             typename ::cuda::std::enable_if_t<::cuda::std::is_integral_v<NumItemsT>, int> = 0>
-  [[nodiscard]] CUB_RUNTIME_FUNCTION _CCCL_FORCEINLINE static cudaError_t SortPairsDescending(
+  [[nodiscard]] CUB_RUNTIME_FUNCTION static cudaError_t SortPairsDescending(
```
bernhardmgruber: Q: Why do we need to remove `_CCCL_FORCEINLINE` here?

gonidelis: fixed 🤦

bernhardmgruber:
I rebased the PR to resolve conflicts.

cub/cub/device/device_radix_sort.cuh
```diff
@@ -125,57 +125,71 @@ CUB_NAMESPACE_BEGIN
...
-      stream,
-      decomposer);
+    using offset_t           = detail::choose_offset_t<NumItemsT>;
+    using decomposer_check_t = detail::radix::decomposer_check_t<KeyT, DecomposerT>;
+
+    static_assert(decomposer_check_t::value,
+                  "DecomposerT must be a callable object returning a tuple of references to "
+                  "arithmetic types");
+
+    if constexpr (decomposer_check_t::value)
```
gonidelis: why is the `static_assert(check); if constexpr(check){}` practice good? nvhpc hates the unreachable path after `if constexpr` pushing a nit-fix but why not drop the `if constexpr` through and through since static_assert guard it anyways?


```suggestion
    static_assert(decomposer_check_t::value,
                  "DecomposerT must be a callable object returning a tuple of references to "
                  "arithmetic types");
                  
          return detail::radix_sort::dispatch<Order>(
```
is that wrong?

bernhardmgruber: > why is the static_assert(check); if constexpr(check){} practice good?

It isn't IMO.

> why not drop the if constexpr through

The answer to that questions is 2m in front of you once you are back in HQ ;)

That person explained to me that by branching on the static assert condition, we can prevent the compiler from emitting follow-up errors, in case the assertion fails. This is supposed to reduce the output in case of compilation errors and should thus be user friendly. I don't buy this argument, because so far it has confused every developer and the error message in case of a compilation failure is long regardless.

I think the best course of action is to place a `requires decomposer_check_t::value` at the public API. But that's for a separate PR.

Feel free to remove the `if constexpr` or leave it. I have learned to ignore it.

cub/test/catch2_test_device_radix_sort_env.cu
```diff
@@ -263,56 +263,185 @@ TEST_CASE("Device radix sort pairs decomposer with bits works with default envir
...
-  REQUIRE(h_keys[2].key == 3);
-  auto& values = d_values.selector == 0 ? values_buf0 : values_buf1;
-  c2h::host_vector<int> h_values(values);
-  REQUIRE(h_values[0] == 1);
-  REQUIRE(h_values[1] == 2);
-  REQUIRE(h_values[2] == 0);
+  for (int i = 0; i < 7; ++i)
+  {
+    REQUIRE(static_cast<custom_key_t>(keys[i]).key == static_cast<custom_key_t>(expected[i]).key);
+  }
```
bernhardmgruber: Important: comparing element by element is slow, we should just compare the vectors directly:
```suggestion
    REQUIRE(keys == expected);
```
If that fails, we are maybe missing a comparison operator somewhere.

bernhardmgruber: Pushed a fix.

bernhardmgruber:
I rebased to resolve the conflict.

cub/test/catch2_test_device_radix_sort_env_api.cu
```diff
@@ -433,14 +439,223 @@ C2H_TEST("cub::DeviceRadixSort::SortKeys DB decomposer+bits env-based API", "[ra
...
+    keys_decomposer_t{},
+    0,
+    sizeof(int) * 8,
+    stream_ref);
+
+  // example-end radix-sort-keys-descending-decomposer-bits-env
+  stream.sync();
+
+  REQUIRE(error == cudaSuccess);
+  thrust::device_vector<custom_key_t> expected{{9}, {8}, {7}, {6}, {5}, {3}, {0}};
```
bernhardmgruber: Important: I think we should show the expected outcome in the documentation example:

```suggestion
  auto error = cub::DeviceRadixSort::SortKeysDescending(
    keys_in.data().get(),
    keys_out.data().get(),
    static_cast<int>(keys_in.size()),
    keys_decomposer_t{},
    0,
    sizeof(int) * 8,
    stream_ref);

  thrust::device_vector<custom_key_t> expected{{9}, {8}, {7}, {6}, {5}, {3}, {0}};
  // example-end radix-sort-keys-descending-decomposer-bits-env
  stream.sync();

  REQUIRE(error == cudaSuccess);
```

cub/test/catch2_test_device_radix_sort_env_api.cu
```diff
@@ -364,10 +379,7 @@ C2H_TEST("cub::DeviceRadixSort::SortKeys decomposer+bits env-based API", "[radix
 
   REQUIRE(error == cudaSuccess);
   thrust::device_vector<custom_key_t> expected{{0}, {3}, {5}, {6}, {7}, {8}, {9}};
-  for (int i = 0; i < 7; ++i)
-  {
-    REQUIRE(static_cast<custom_key_t>(keys_out[i]).key == static_cast<custom_key_t>(expected[i]).key);
-  }
+  REQUIRE(keys_out == expected);
```
gonidelis, line 381: @bernhardmgruber  giga thanks for these buncho loops --> `REQUIRE`

cub/test/catch2_test_device_radix_sort_env_api.cu
gonidelis: yes we should

cub/test/catch2_test_device_radix_sort_env_api.cu
bernhardmgruber: np

cub/test/catch2_test_device_radix_sort_env.cu
```diff
@@ -596,6 +704,126 @@ TEST_CASE("Device radix sort keys descending uses custom stream", "[radix_sort][
...
+  stream.sync();
+
+  c2h::host_vector<custom_pair_key_t> h_keys_out(keys_out);
+  REQUIRE(h_keys_out[0].key == 1);
+  REQUIRE(h_keys_out[1].key == 2);
+  REQUIRE(h_keys_out[2].key == 3);
+  c2h::host_vector<int> h_values_out(values_out);
+  REQUIRE(h_values_out[0] == 1);
+  REQUIRE(h_values_out[1] == 2);
+  REQUIRE(h_values_out[2] == 0);
```
bernhardmgruber: Please just compare with a vector of reference results
```suggestion
  c2h::device_vector<custom_pair_key_t> expected_keys{{1, 200}, {2, 300}, {3, 100}};
  REQUIRE(keys_out == expected_keys);
  c2h::device_vector<int> expected_values{1, 2, 0};
  REQUIRE(values_out == expected_values);
```

This should work if you add:
```c++

  __host__ __device__ friend bool operator==(const custom_pair_key_t& a, const custom_pair_key_t& b)
  {
    return a.key == b.key && a.payload == b.payload;
  }

  __host__ __device__ friend bool operator!=(const custom_pair_key_t& a, const custom_pair_key_t& b)
  {
    return !(a == b);
  }

  friend std::ostream& operator<<(std::ostream& os, const custom_pair_key_t& cpk)
  {
    return os << "{" << cpk.key << ", " << cpk.payload << "}";
  }
```
to `struct custom_pair_key_t`

gonidelis: added in env_api test too

cub/test/catch2_test_device_radix_sort_env.cu
```diff
@@ -596,6 +711,110 @@ TEST_CASE("Device radix sort keys descending uses custom stream", "[radix_sort][
...
+
+  auto& keys = d_keys.selector == 0 ? keys_buf0 : keys_buf1;
+  c2h::device_vector<custom_pair_key_t> expected_keys{{1, 200}, {2, 300}, {3, 100}};
+  REQUIRE(keys == expected_keys);
+  auto& values = d_values.selector == 0 ? values_buf0 : values_buf1;
+  c2h::device_vector<int> expected_values{1, 2, 0};
+  REQUIRE(values == expected_values);
+}
+
+#endif
```
bernhardmgruber, line 789:
```suggestion
#endif // TEST_LAUNCH == 0
```
for @miscco 's happiness

PR #8106, opened 2026-03-19
fix tests and replace exec() in compute/_jit.py

water0615:
## Description

closes 

fix tests and replace exec() in compute/_jit.py

## Checklist

### DIFF
--- a/cub/benchmarks/bench/topk/keys.cu
+++ b/cub/benchmarks/bench/topk/keys.cu
@@ -79,6 +79,7 @@ void topk_keys(nvbench::state& state, nvbench::type_list<KeyT, OffsetT, OutOffse
     static_cast<cub::NullType*>(nullptr),
     static_cast<offset_t>(elements),
     static_cast<out_offset_t>(selected_elements),
+    cub::detail::identity_decomposer_t{},
     0
 #if !TUNE_BASE
     ,
@@ -99,6 +100,7 @@ void topk_keys(nvbench::state& state, nvbench::type_list<KeyT, OffsetT, OutOffse
       static_cast<cub::NullType*>(nullptr),
       static_cast<offset_t>(elements),
       static_cast<out_offset_t>(selected_elements),
+      cub::detail::identity_decomposer_t{},
       launch.get_stream()
 #if !TUNE_BASE
         ,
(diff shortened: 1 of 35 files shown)


shwina:
Thanks for the PR, @water0615!

By its nature, `cuda.compute` accepts arbitrary user callables and should be run in trusted environments.

We don't want to be too restrictive in terms of the contents of the user-provided callables. For instance, the `__globals__` dict could contain e.g., other arbitrarily-named device functions. Thus, I think the restrictions in `_validate_source_safety` are a bit too strict and may prevent legitimate use-cases.

That being said, there are improvements we can make here in terms of validating, e.g., struct members to ensure they are identifiers. For example, I think the following patch would take us quite far in terms of better error reporting if someone passes something other than a identifier for the struct member name:

```diff
@@ -55,6 +55,13 @@ def gpu_struct(
     # At this point, field_dict must be a dict
     assert isinstance(field_dict, dict)

+    # Validate field names are valid Python identifiers before use in exec()
+    for key in field_dict:
+        if not isinstance(key, str) or not key.isidentifier():
+            raise ValueError(
+                f"gpu_struct field name {key!r} is not a valid Python identifier"
+            )
+
     # Normalize fields for storage on the struct class
     field_spec = {}
     for key, val in field_dict.items():
```

Do you think we could start there (and possibly some unit tests for the above)?

python/cuda_cccl/cuda/compute/_jit.py
```diff
@@ -183,22 +183,24 @@ def error_impl(struct_val, idx):
...
-            f"    {'if' if i == 0 else 'elif'} idx == {i}: return struct_val.{name}"
-            for i, name in enumerate(field_names_list)
-        )
-        exec(
-            f"def impl(struct_val, idx):\n{conditions}\n    else: raise IndexError('Index out of range')",
-            namespace := {},
-        )
-        return namespace["impl"]
+            def impl(struct_val, idx):
+                return getattr(struct_val, field_name)
```
shwina, line 190: I believe numba's type system will need to see the attribute access rather than access via `getattr`; so this may not work as intended.

Instead, we could assert that all field names are valid Python identifiers via `str.isidentifier()`.

shwina:
I'm going to go ahead and close this PR, as https://github.com/NVIDIA/cccl/pull/8235 addresses the issue without restricting the contents of user-provided callables.

PR #8109, opened 2026-03-19
Fix issues with thrust  benchmarks

miscco:
There were some issues with the thrust benchmarks copied over.

seems like they are not running properly in IC

### DIFF
--- a/thrust/benchmarks/bench/find_if/basic.cu
+++ b/thrust/benchmarks/bench/find_if/basic.cu
@@ -1,15 +1,17 @@
-// SPDX-FileCopyrightText: Copyright (c) 2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
-#include <thrust/detail/internal_functional.h>
 #include <thrust/device_vector.h>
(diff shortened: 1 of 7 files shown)


thrust/benchmarks/bench/find_if/basic.cu
```diff
@@ -24,13 +26,16 @@ void find_if(nvbench::state& state, nvbench::type_list<T>)
...
+                 thrust::find_if(policy(alloc, launch), dinput.begin(), dinput.end(), cuda::equal_to_value<T>{val}));
+             });
 }
 
-NVBENCH_BENCH_TYPES(find_if, NVBENCH_TYPE_AXES(fundamental_types))
+NVBENCH_BENCH_TYPES(basic, NVBENCH_TYPE_AXES(fundamental_types))
   .set_name("base")
   .add_int64_power_of_two_axis("Elements", nvbench::range(16, 28, 4))
-  .add_float64_axis("MismatchAt", std::vector{1.0, 0.5, 0.0});
+  .add_float64_axis("MismatchAt", std::vector{1.0, 0.5, 0.01});
```
bernhardmgruber, line 41: Q: Why is this necessary? We also use `0.0` in CUB, which means the very first element of the sequence contains the element to be found.

miscco: This is a new benchmark and I want that to be the same as the PSTL one which uses this

miscco: Updated all to use 0.0

PR #8115, opened 2026-03-20
[cudax] Refactor graph APIs to use CUDA Driver API

pciolkosz:
Currently graph APIs use CUDA Runtime, but we should bring it in line with the rest of the project and use the driver API

This PR also introduces experimental version of the driver API layer that uses type defines from a driver header, which was necessary for one of the graph functions. Once tested we can think about migrating the main driver layer in libcu++ to this new implementation

### DIFF
--- a/cudax/include/cuda/experimental/__driver/driver_api.cuh
+++ b/cudax/include/cuda/experimental/__driver/driver_api.cuh
@@ -4,7 +4,7 @@
 // under the Apache License v2.0 with LLVM Exceptions.
 // See https://llvm.org/LICENSE.txt for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
-// SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES.
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES.
 //
 //===----------------------------------------------------------------------===//
 
--- a/cudax/include/cuda/experimental/__graph/graph.cuh
+++ b/cudax/include/cuda/experimental/__graph/graph.cuh
@@ -4,7 +4,7 @@
 // under the Apache License v2.0 with LLVM Exceptions.
 // See https://llvm.org/LICENSE.txt for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
-// SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES.
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES.
 //
 //===----------------------------------------------------------------------===//
 
--- a/libcudacxx/include/cuda/__driver/driver_api.h
+++ b/libcudacxx/include/cuda/__driver/driver_api.h
@@ -4,7 +4,7 @@
 // under the Apache License v2.0 with LLVM Exceptions.
 // See https://llvm.org/LICENSE.txt for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
-// SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES.
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES.
 //
 //===----------------------------------------------------------------------===//
 
--- a/cudax/include/cuda/experimental/__graph/graph_builder.cuh
+++ b/cudax/include/cuda/experimental/__graph/graph_builder.cuh
@@ -4,7 +4,7 @@
 // under the Apache License v2.0 with LLVM Exceptions.
 // See https://llvm.org/LICENSE.txt for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
-// SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES.
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES.
 //
 //===----------------------------------------------------------------------===//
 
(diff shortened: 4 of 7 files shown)


cudax/include/cuda/experimental/__graph/graph.cuh
```diff
@@ -110,7 +111,8 @@ struct _CCCL_TYPE_VISIBILITY_DEFAULT graph
   {
     if (auto __exec = ::cuda::std::exchange(__exec_, nullptr))
     {
-      _CCCL_ASSERT_CUDA_API(cudaGraphExecDestroy, "cudaGraphDestroy failed", __exec);
+      [[maybe_unused]] auto __status = ::cuda::experimental::__driver::__graphExecDestroyNoThrow(__exec);
+      _CCCL_ASSERT(__status == cudaSuccess, "cuGraphExecDestroy failed");
```
bernhardmgruber, line 115: Q: Does the `\throws cuda::std::cuda_error if `cudaGraphExecDestroy` fails.` still apply?

libcudacxx/include/cuda/__driver/driver_api.h
```diff
@@ -95,7 +95,7 @@ _CCCL_SUPPRESS_DEPRECATED_POP
 //! @brief Makes the driver version from major and minor version.
 [[nodiscard]] _CCCL_HOST_API constexpr int __make_version(int __major, int __minor) noexcept
 {
-  _CCCL_ASSERT(__major >= 12, "invalid major CUDA Driver version");
+  _CCCL_ASSERT(__major >= 2, "invalid major CUDA Driver version");
```
bernhardmgruber, line 98: Q: Does this change result in forward compatible binaries? Meaning, if I compile this code into my binary and run it on a future GPU/driver in 10 years, will it still run?

cudax/include/cuda/experimental/__driver/driver_api.cuh
```diff
@@ -0,0 +1,364 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of CUDA Experimental in CUDA C++ Core Libraries,
+// under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES.
```
miscco:
```suggestion
// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES.
```

cudax/include/cuda/experimental/__driver/driver_api.cuh
```diff
@@ -0,0 +1,364 @@
...
+{
+  static auto __driver_fn = _CUDAX_GET_DRIVER_FUNCTION(cuGraphAddMemsetNode, 10, 0);
+  ::CUgraphNode __node{};
+  ::CUDA_MEMSET_NODE_PARAMS __params{};
+  __params.dst         = __dst;
+  __params.pitch       = __pitch;
+  __params.value       = __value;
+  __params.elementSize = __element_size;
+  __params.width       = __width;
+  __params.height      = __height;
```
miscco, line 66: Question: could we directly initialize the `CUDA_MEMSET_NODE_PARAMS`? Or is this relying on some undefined internal order?

libcudacxx/include/cuda/__driver/driver_api.h
pciolkosz: Yes, CUDA library is forever forward compatible and they never remove functions. The experimental version now requests an exact version and confirms the type with the PFN typedef, so it should remain stable with all future CUDA versions

cudax/include/cuda/experimental/__driver/driver_api.cuh
pciolkosz: With this many arguments I think I prefer to do that separately

cudax/include/cuda/experimental/__graph/graph.cuh
pciolkosz: Thanks, removed that note

cudax/include/cuda/experimental/__driver/driver_api.cuh
```diff
@@ -0,0 +1,364 @@
...
+
+#  include <cuda/std/__cccl/prologue.h>
+
+// Get a driver function pointer, casting to the PFN typedef for type safety.
+// Uses PFN_ typedefs from cudaTypedefs.h to avoid ABI mismatches caused by
+// #define'd version aliases in cuda.h (e.g. #define cuFoo cuFoo_v2).
+// The ## operator suppresses macro expansion of the function name, so this is
+// safe even for names that are #define'd to versioned variants.
+#  define _CUDAX_GET_DRIVER_FUNCTION(pfn_name, major, minor)  \
+    reinterpret_cast<PFN_##pfn_name##_v##major##0##minor##0>( \
```
davebayer, line 40: detail
```suggestion
    reinterpret_cast<::PFN_##pfn_name##_v##major##0##minor##0>( \
```

cudax/include/cuda/experimental/__driver/driver_api.cuh
```diff
@@ -0,0 +1,364 @@
...
+  ::CUdeviceptr __dst,
+  ::cuda::std::size_t __pitch,
+  unsigned int __value,
+  unsigned int __element_size,
+  ::cuda::std::size_t __width,
+  ::cuda::std::size_t __height)
+{
+  static auto __driver_fn = _CUDAX_GET_DRIVER_FUNCTION(cuGraphAddMemsetNode, 10, 0);
+  ::CUgraphNode __node{};
+  ::CUDA_MEMSET_NODE_PARAMS __params{};
```
davebayer, line 60: Q: wouldn't be better to just take `CUDA_MEMSET_NODE_PARAMS` as the parameter? I think it's easy to make a mistake, when a function takes that many parameters

pciolkosz: I have some different refactor in a follow up PR to move everything to `cuGraphAddNode` instead

cudax/include/cuda/experimental/__driver/driver_api.cuh
pciolkosz: Will fix in follow-up PR

PR #8116, opened 2026-03-20
[libcu++] Deprecate default alignment in type erased resource wrappers

pciolkosz:
Currently type erased resource wrappers have 16 as a default alignment argument. But CUDA resources have 256 as default alignment, so packing such resource into the wrapper changes the default alignment, which is potentially a footgun. Default alignment argument is just a convenience, the resource concept only specifies the alignment taking overload. We also don't expect users to call `allocate/deallocate` directly and use buffer or other types instead. We could add logic to the wrappers to store the default alignment in its state, but I doesn't seem worth the extra storage. Instead this PR deprecates the default alignment argument and instead explicit alignment argument is required with the type erased wrappers

### DIFF
--- a/libcudacxx/include/cuda/__memory_resource/any_resource.h
+++ b/libcudacxx/include/cuda/__memory_resource/any_resource.h
@@ -133,8 +133,8 @@ struct __ibasic_resource : __basic_interface<__ibasic_resource>
   }
 
   CCCL_DEPRECATED_BECAUSE("Specify an explicit alignment argument. The default alignment will be removed in a future "
-                           "release.")
-  _CCCL_PUBLIC_HOST_API void* allocate_sync(size_t __bytes)
+                          "release.") _CCCL_PUBLIC_HOST_API void*
+  allocate_sync(size_t __bytes)
   {
     return allocate_sync(__bytes, alignof(::cuda::std::max_align_t));
   }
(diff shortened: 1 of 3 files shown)


libcudacxx/include/cuda/__memory_resource/any_resource.h
```diff
@@ -543,7 +573,13 @@ class basic_any_resource
...
   //! @return `obj.allocate_sync(__size, __align)`, where `obj` is the wrapped
   //! object.
-  [[nodiscard]] void* allocate_sync(size_t __size, size_t __align = alignof(cuda::std::max_align_t));
+  [[nodiscard]] void* allocate_sync(size_t __size, size_t __align);
+
+  //! @brief Calls `allocate_sync` on the wrapped object with
+  //! `alignof(::cuda::std::max_align_t)` as the alignment.
+  //! @deprecated Specify an explicit alignment argument.
+  //! @pre `has_value()` is `true`.
+  [[deprecated]] [[nodiscard]] void* allocate_sync(size_t __size);
```
davebayer, line 582:
```suggestion
  [[nodiscard]] CCCL_DEPRECATED void* allocate_sync(size_t __size);
```

davebayer (changes requested):

libcudacxx/include/cuda/__memory_resource/any_resource.h
pciolkosz: These are just alternative declarations for doxygen, I think it should stay with `[[deprecated]]`

PR #8117, opened 2026-03-20
Avoid alias violations in `SHUFFLE_{UP, DOWN, INDEX}`

miscco:
We received an internal bug report about those functions.

The `reinterpret_cast` is a violation of the aliasing rules and will lead to miscompilations with compiles that enable TBAA-based optimizations via `-fstrict-aliasing` which is on by default at `-O2`

We can already observe this via clang-cuda when running the thrust max_element tests

### DIFF
--- a/cub/cub/util_ptx.cuh
+++ b/cub/cub/util_ptx.cuh
@@ -22,7 +22,10 @@
 #include <cub/util_debug.cuh>
 #include <cub/util_type.cuh>
 
+#include <cuda/__cmath/ceil_div.h>
 #include <cuda/__cmath/pow2.h>
+#include <cuda/__memory/alias_to_words.h>
+#include <cuda/__warp/warp_shuffle.h>
 
 CUB_NAMESPACE_BEGIN
 
@@ -211,28 +214,19 @@ template <int LOGICAL_WARP_THREADS, typename T>
 _CCCL_DEVICE _CCCL_FORCEINLINE T ShuffleUp(T input, int src_offset, int first_thread, unsigned int member_mask)
 {
   /// The 5-bit SHFL mask for logically splitting warps into sub-segments starts 8-bits up
-  constexpr int SHFL_C = (32 - LOGICAL_WARP_THREADS) << 8;
+  static constexpr int SHFL_C = (32 - LOGICAL_WARP_THREADS) << 8;
+  static constexpr int WORDS  = ::cuda::ceil_div(sizeof(T), sizeof(uint32_t));
 
-  using ShuffleWord = typename UnitWord<T>::ShuffleWord;
-
-  constexpr int WORDS = (sizeof(T) + sizeof(ShuffleWord) - 1) / sizeof(ShuffleWord);
-
-  T output;
-  ShuffleWord* output_alias = reinterpret_cast<ShuffleWord*>(&output);
-  ShuffleWord* input_alias  = reinterpret_cast<ShuffleWord*>(&input);
-
-  unsigned int shuffle_word;
-  shuffle_word    = SHFL_UP_SYNC((unsigned int) input_alias[0], src_offset, first_thread | SHFL_C, member_mask);
-  output_alias[0] = shuffle_word;
+  auto words = ::cuda::__alias_to_words(input);
 
   _CCCL_PRAGMA_UNROLL_FULL()
-  for (int WORD = 1; WORD < WORDS; ++WORD)
+  for (int WORD = 0; WORD < WORDS; ++WORD)
   {
-    shuffle_word       = SHFL_UP_SYNC((unsigned int) input_alias[WORD], src_offset, first_thread | SHFL_C, member_mask);
-    output_alias[WORD] = shuffle_word;
+    words[WORD] = SHFL_UP_SYNC(words[WORD], src_offset, first_thread | SHFL_C, member_mask);
   }
 
-  return output;
+  input = words.template __convert_back<T>();
+  return input;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


cub/cub/util_ptx.cuh
```diff
@@ -291,25 +291,23 @@ _CCCL_DEVICE _CCCL_FORCEINLINE T ShuffleDown(T input, int src_offset, int last_t
...
+  ShuffleWord output_alias[WORDS];
+  ShuffleWord input_alias[WORDS];
+  ::cuda::std::memcpy(input_alias, &input, sizeof(T));
 
   unsigned int shuffle_word;
-  shuffle_word    = SHFL_DOWN_SYNC((unsigned int) input_alias[0], src_offset, last_thread | SHFL_C, member_mask);
-  output_alias[0] = shuffle_word;
-
   _CCCL_PRAGMA_UNROLL_FULL()
-  for (int WORD = 1; WORD < WORDS; ++WORD)
```
miscco, line 307: Can somebody explain to me why the first iteration was pulled out of the loop? Given that it is `_CCCL_PRAGMA_UNROLL_FULL` I would assume that to be the same

davebayer (approved):

elstehle:
Can you please check for sass changes of device algorithms that are affected by this change, if any?

cub/cub/util_ptx.cuh
```diff
@@ -291,25 +291,23 @@ _CCCL_DEVICE _CCCL_FORCEINLINE T ShuffleDown(T input, int src_offset, int last_t
...
+  constexpr int WORDS = ::cuda::ceil_div(sizeof(T), sizeof(ShuffleWord));
 
-  constexpr int WORDS = (sizeof(T) + sizeof(ShuffleWord) - 1) / sizeof(ShuffleWord);
-
-  T output;
-  ShuffleWord* output_alias = reinterpret_cast<ShuffleWord*>(&output);
-  ShuffleWord* input_alias  = reinterpret_cast<ShuffleWord*>(&input);
+  ShuffleWord output_alias[WORDS];
+  ShuffleWord input_alias[WORDS];
+  ::cuda::std::memcpy(input_alias, &input, sizeof(T));
```
jrhemstad: Important: this changes require very thorough SASS inspection and performance benchmarking. Using memcpy inside device code like this is extremely likely to result in 1 byte loads/stores which are disastrous for performance.

davebayer: I thought this is basically just bitcasting the input to words, so I would expect no code to be generated here

fbusato: agree. If `input` is in registers, it translates to no op.

fbusato: important, use `sizeof(input)`, instead of `sizeof(T)`
```suggestion
  ::cuda::std::memcpy(input_alias, &input, sizeof(input));
```

fbusato:
> We received an internal bug report about those functions.
> The reinterpret_cast is a violation of the aliasing rules and will lead to miscompilations with compiles that enable TBAA-based optimizations via -fstrict-aliasing which is on by default at -O2

this practically breaks the vast majority of CUDA code. We need to find a systematic way to catch this problem at compile-time. We need to involve the compiler team, and on our side use static analysis, or simply searching for `reinterpret_cast`.

bernhardmgruber:
I swear I didn't make this up:

> or simply searching for `reinterpret_cast`.

```bash
~/dev/cccl$ git grep reinterpret_cast | wc -l
9224
```
image

cub/cub/util_ptx.cuh
pauleonix: Looking at the blame/commits my best guess is that Duane was debugging in 2015 and maybe wanted to put a breakpoint after the first iteration. The implementation started with the unrolled loop from 0, went to a recursive implementation, back to the loop and then to the loop starting at 1 in commits spanning only a few days. The commented out call to the recursive implementation was only removed years later in 2017.

cub/cub/util_ptx.cuh
pauleonix: Can't we use `cuda::std::bit_cast()` here? Edit: I think the answer is no because the alias could be bigger than the original in case the original has a size that is not a multiple of 4 bytes.

fbusato: we should..but we need to introduce `cuda::std::array`.

miscco: The issue is not bit_cast, but that input types might not be trivially copyable which the builtin actually checks against

I am experimenting with different solutions

davebayer: But if the type is not trivially copyable, isn't this whole thing an undefined behaviour?

fbusato:
side clarification. I'm pretty sure that many `reinterpret_cast` cases are perfectly valid in our codebase, e.g. `reinterpret_cast<size_t>`

miscco:
NOTE: this shows nontrivial amount of SASS changs for thrust::max_element, so we need to investigate how to properly do this

libcudacxx/include/cuda/__warp/warp_shuffle.h
```diff
@@ -73,19 +74,17 @@ template <int _Width = 32, typename _Tp, typename _Up = ::cuda::std::remove_cv_t
...
     {
-      __array[i] = ::cuda::ptx::shfl_sync_idx(__array[i], __pred, __src_lane, __clamp_segmask, __lane_mask);
+      __words[i] = ::cuda::ptx::shfl_sync_idx(__words[i], __pred, __src_lane, __clamp_segmask, __lane_mask);
     }
+
     warp_shuffle_result<_Up> __result;
     __result.pred = __pred;
-    ::cuda::std::memcpy(
-      static_cast<void*>(::cuda::std::addressof(__result.data)), static_cast<void*>(__array), sizeof(_Up));
+    __words.__assign_to(__result.data);
```
miscco, line 87: Note: We cannot directly assign here, as `_Up` might be an array

libcudacxx/include/cuda/__warp/warp_shuffle.h
```diff
@@ -217,19 +209,16 @@ template <int _Width = 32, typename _Tp, typename _Up = ::cuda::std::remove_cv_t
...
     for (int i = 0; i < __ratio; ++i)
     {
-      __array[i] = ::cuda::ptx::shfl_sync_bfly(__array[i], __pred, __xor_mask, __clamp_segmask, __lane_mask);
+      __words[i] = ::cuda::ptx::shfl_sync_bfly(__words[i], __pred, __xor_mask, __clamp_segmask, __lane_mask);
     }
     warp_shuffle_result<_Up> __result;
     __result.pred = __pred;
-    ::cuda::std::memcpy(
-      static_cast<void*>(::cuda::std::addressof(__result.data)), static_cast<void*>(__array), sizeof(_Up));
+    __words.__assign_to(__result.data);
```
miscco, line 221: Question, I am wondering whether we want to turn those into member functions of the `__alias_storage`

bernhardmgruber:
> I'm pretty sure that many `reinterpret_cast` cases are perfectly valid in our codebase, e.g. `reinterpret_cast<size_t>`

ok, now I want to know where a `reinterpret_cast<size_t>` would be valid, because I can't imagine a situation where it would be and no other integer type would be better suited (like `uintptr_t` etc.)

fbusato:
ahah, my fault `reinterpret_cast<uintptr_t>` is the right one

miscco:
dropping in favor of #8159

PR #8124, opened 2026-03-20
fix thrust mr validate alignment condition

davebayer:
Fixes problems introduced in #8091

### DIFF
--- a/thrust/thrust/mr/pool_options.h
+++ b/thrust/thrust/mr/pool_options.h
@@ -99,7 +99,10 @@ struct pool_options
     {
       return false;
     }
-    if (::cuda::__is_valid_alignment(alignment))
+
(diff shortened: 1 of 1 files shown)


pciolkosz (approved):

thrust/thrust/mr/pool_options.h
```diff
@@ -99,7 +99,7 @@ struct pool_options
     {
       return false;
     }
-    if (::cuda::__is_valid_alignment(alignment))
+    if (!::cuda::__is_valid_alignment(alignment))
```
davebayer: I've realized that this change is breaking, because `0` was considered a valid value. If the user doesn't set `options.alignment`, this method will return `false`. I will probably partially revert the change.

PR #8125, opened 2026-03-20
Optimized Device-to-Device Tensor Copy (cudax) - Transpose Case

fbusato:
## Description

Requires **Optimized Device-to-Device Tensor Copy (cudax)** #7823 before.

The PR adds the workflow to efficiently handle transpose cases by exploiting the shared memory.


cudax/include/cuda/experimental/__copy/copy_shared_memory.cuh
```diff
@@ -0,0 +1,359 @@
...
+//! @param[in]  __src_ptr               Pointer to source data
+//! @param[in]  __src_accessor          Accessor for reading source elements
+//! @param[out] __dst_ptr               Pointer to destination data
+//! @param[in]  __dst_accessor          Accessor for writing destination elements
+//! @param[in]  __grid_iter             Coordinate iterator for grid tile decomposition
+//! @param[in]  __grid_tile_src_strides Per-dimension source strides scaled by tile sizes
+//! @param[in]  __grid_tile_dst_strides Per-dimension destination strides scaled by tile sizes
+//! @param[in]  __tile_perm_iter        Coordinate iterator for src-permuted tile decomposition
+//! @param[in]  __src_perm_src_strides  Src-permuted source strides for loading
+//! @param[in]  __tile_perm_smem_strides Src-permuted shared memory strides for loading
```
NaderAlAwar, line 81: Nit: please add one space to all other param descriptions so that they are aligned

cudax/include/cuda/experimental/__copy/copy_shared_memory_utils.cuh
```diff
@@ -0,0 +1,218 @@
...
+  // - accumulate the product of the tile sizes until the shared memory limit is reached
+  // - the tile size (which is at block level) is capped at the warp size to get coalesced accesses
+  const auto __current_dev            = ::cuda::experimental::__current_device();
+  const size_t __max_shared_mem_bytes = __current_dev.attribute<::cudaDevAttrMaxSharedMemoryPerBlock>();
+  size_t __size_product               = 1;
+  int __tile_rank                     = 0;
+  for (size_t __r = 0; __r < __num_contiguous_dst; ++__r, ++__tile_rank)
+  {
+    const auto __tile_size_r = ::cuda::std::min(static_cast<size_t>(__dst.__extents[__r]), __max_tile_size);
+    if (__size_product * __tile_size_r * sizeof(_TpOut) > __max_shared_mem_bytes)
```
NaderAlAwar, line 132: Important: this shared-memory sizing seems inconsistent with the kernel’s actual staging type used in shared memory. This check uses `sizeof(_TpOut)` while the kernel allocates dynamic shared memory as `remove_cv_t<_TpIn>[]`, which means it can pick a tile that fits according to `sizeof(_TpOut)` but not according to the actual shared-memory footprint of the kernel when `sizeof(_TpIn) > sizeof(_TpOut)`. Is the intent to size this using the staged type instead? I wonder if there is an optimization opportunity where you can use the smaller of the two types for staging?

cudax/include/cuda/experimental/__copy/copy_shared_memory_utils.cuh
```diff
@@ -0,0 +1,218 @@
...
+  }
+
+  // * there are enough tiles to keep the GPU busy (at least one full wave across all SMs)
+  size_t __num_tiles = 1; // __num_tiles == number of blocks
+  for (__rank_t __r = 0; __r < __tile_rank; ++__r)
+  {
+    const auto __extent    = static_cast<size_t>(__dst.__extents[__r]);
+    const auto __tile_size = ::cuda::std::min(__extent, __max_tile_size);
+    __num_tiles *= ::cuda::ceil_div(__extent, __tile_size);
+  }
```
NaderAlAwar, line 152: Important: this occupancy check seems more conservative than the actual launch. Here `__num_tiles` is only multiplied over `__tile_rank`, i.e. the tiled prefix, but the launched `__grid_size` later is multiplied over all active dimensions, with untiled dimensions contributing `ceil_div(extent, 1)`. Is the intent here to estimate the real grid size, or only the tiled-prefix parallelism? As written, it looks like this can reject the shared-memory path unnecessarily.

cudax/include/cuda/experimental/__copy/copy_shared_memory_utils.cuh
```diff
@@ -0,0 +1,218 @@
...
+//!
+//! @param[in] __tile_total_bytes Shared memory required for one tile in bytes
+//! @return Thread block size
+[[nodiscard]] _CCCL_HOST_API inline int __find_thread_block_size(::cuda::std::size_t __tile_total_bytes) noexcept
+{
+  using ::cuda::std::size_t;
+  const auto __dev                      = ::cuda::experimental::__current_device();
+  const size_t __total_sm_threads       = __dev.attribute<::cudaDevAttrMaxThreadsPerMultiProcessor>();
+  const size_t __max_thread_block_size  = __dev.attribute<::cudaDevAttrMaxThreadsPerBlock>();
+  const size_t __total_shared_mem_bytes = __dev.attribute<::cudaDevAttrMaxSharedMemoryPerBlock>();
```
NaderAlAwar, line 207: Important: this should probably use `MaxSharedMemoryPerMultiprocessor` to match the behavior described in the docstring. The comment says this is dividing the SM thread budget across as many blocks as shared memory allows, but this is currently using `MaxSharedMemoryPerBlock`, which is a per-block limit rather than the SM-level shared-memory budget. As written, it seems to underestimate the number of blocks that can reside on an SM due to shared memory.

cudax/include/cuda/experimental/__copy/copy_shared_memory_utils.cuh
```diff
@@ -0,0 +1,218 @@
...
+  const auto __num_blocks               = ::cuda::ceil_div(__total_shared_mem_bytes, __tile_total_bytes);
+  const auto __thread_block_size =
+    ::cuda::std::min(::cuda::ceil_div(__total_sm_threads, __num_blocks), __max_thread_block_size);
```
NaderAlAwar, line 210: Question: should this just use normal integer division? It feels lik `ceil_div` can overestimate the number of allowed threads per block and lead to lower occupancy

cudax/include/cuda/experimental/__copy/copy_shared_memory_utils.cuh
```diff
@@ -0,0 +1,218 @@
...
+//! @return Thread block size
+[[nodiscard]] _CCCL_HOST_API inline int __find_thread_block_size(::cuda::std::size_t __tile_total_bytes) noexcept
+{
+  using ::cuda::std::size_t;
+  const auto __dev                      = ::cuda::experimental::__current_device();
+  const size_t __total_sm_threads       = __dev.attribute<::cudaDevAttrMaxThreadsPerMultiProcessor>();
+  const size_t __max_thread_block_size  = __dev.attribute<::cudaDevAttrMaxThreadsPerBlock>();
+  const size_t __total_shared_mem_bytes = __dev.attribute<::cudaDevAttrMaxSharedMemoryPerBlock>();
+  const auto __num_blocks               = ::cuda::ceil_div(__total_shared_mem_bytes, __tile_total_bytes);
+  const auto __thread_block_size =
```
NaderAlAwar, line 209: Question: have you considered rounding this to a multiple of 32? This kernel’s tiling is explicitly warp-oriented (`__max_tile_size == 32`), but this heuristic can produce arbitrary block sizes.

cudax/include/cuda/experimental/__copy/copy_shared_memory_utils.cuh
fbusato: use `sizeof(_TpIn)` makes totally sense, as shared memory is used only for the input type

cudax/include/cuda/experimental/__copy/copy_shared_memory_utils.cuh
fbusato: technically the kernel can run with any number of threads. The main issue is that partial warps can waste resources, so it makes sense to round to 32

cudax/include/cuda/experimental/__copy/copy_shared_memory_utils.cuh
fbusato: makes sense

cudax/include/cuda/experimental/__copy/copy_shared_memory_utils.cuh
fbusato: right, this check is too conservative. External dimensions, e.g. `4, 128, 128` with `128x128` tile, misses `x4` blocks.

cudax/include/cuda/experimental/__copy/copy_shared_memory_utils.cuh
fbusato: I also found another occurrence of the same issue

cudax/include/cuda/experimental/__copy/copy_shared_memory_utils.cuh
```diff
@@ -0,0 +1,219 @@
...
+//! @param[in] __tile_total_bytes Shared memory required for one tile in bytes
+//! @return Thread block size
+[[nodiscard]] _CCCL_HOST_API inline int __find_thread_block_size(::cuda::std::size_t __tile_total_bytes) noexcept
+{
+  using ::cuda::std::size_t;
+  const auto __dev                      = ::cuda::experimental::__current_device();
+  const size_t __total_sm_threads       = __dev.attribute<::cudaDevAttrMaxThreadsPerMultiProcessor>();
+  const size_t __max_thread_block_size  = __dev.attribute<::cudaDevAttrMaxThreadsPerBlock>();
+  const size_t __total_shared_mem_bytes = __dev.attribute<::cudaDevAttrMaxSharedMemoryPerMultiprocessor>();
+  const auto __num_blocks_per_sm        = ::cuda::ceil_div(__total_shared_mem_bytes, __tile_total_bytes);
```
NaderAlAwar: Suggestion: I believe this should also use normal integer division not `ceil_div` if it is supposed to represent resident blocks per SM by shared memory

cudax/include/cuda/experimental/__copy/copy_shared_memory_utils.cuh
```diff
@@ -0,0 +1,219 @@
...
+  const auto __thread_block_size = ::cuda::std::min(__total_sm_threads / __num_blocks_per_sm, __max_thread_block_size);
+  const auto __thread_block_size32 = ::cuda::round_up(__thread_block_size, /*warp size=*/size_t{32});
+  return static_cast<int>(__thread_block_size);
```
NaderAlAwar: Important: this should return the rounded up value
```suggestion
  return static_cast<int>(__thread_block_size32);
```

NaderAlAwar (approved):

cudax/include/cuda/experimental/__copy/copy_shared_memory.cuh
```diff
@@ -0,0 +1,359 @@
...
+
+  // Partial tile detection: is the current tile full or partial?
+  bool __is_full_tile = true;
+  _CCCL_PRAGMA_UNROLL_FULL()
+  for (int __k = 0; __k < __max_rank; ++__k)
+  {
+    const auto __block_start = __grid_coords[__k] * __tile_sizes[__k];
+    if (__block_start + __tile_sizes[__k] > __extents[__k])
+    {
+      __is_full_tile = false;
```
miscco, line 143: I believe we should break here
```suggestion
      __is_full_tile = false;
      break;
```

Jacobfaib, line 143:
```suggestion
      __is_full_tile = false;
      break;
```

cudax/include/cuda/experimental/__copy/copy_shared_memory.cuh
```diff
@@ -0,0 +1,359 @@
...
+  _StrideTIn __src_base    = 0;
+  _StrideTOut __dst_base   = 0;
+  _CCCL_PRAGMA_UNROLL_FULL()
+  for (int __k = 0; __k < __max_rank; ++__k)
+  {
+    __src_base += static_cast<_StrideTIn>(__grid_coords[__k]) * __grid_tile_src_strides[__k];
+    __dst_base += static_cast<_StrideTOut>(__grid_coords[__k]) * __grid_tile_dst_strides[__k];
+  }
+  __src_ptr += __src_base;
+  __dst_ptr += __dst_base;
```
Jacobfaib:
```suggestion
  {
    _StrideTIn __src_base    = 0;
    _StrideTOut __dst_base   = 0;
    _CCCL_PRAGMA_UNROLL_FULL()
    for (int __k = 0; __k < __max_rank; ++__k)
    {
      __src_base += static_cast<_StrideTIn>(__grid_coords[__k]) * __grid_tile_src_strides[__k];
      __dst_base += static_cast<_StrideTOut>(__grid_coords[__k]) * __grid_tile_dst_strides[__k];
  }
    __src_ptr += __src_base;
    __dst_ptr += __dst_base;
  }
```
`__src_base` and `__dst_base` don't appear to be used elsewhere, feel free to ignore if I've missed their use.

cudax/include/cuda/experimental/__copy/copy_shared_memory.cuh
```diff
@@ -0,0 +1,359 @@
...
+    // map the linear index to the multi-dimensional coordinates and copy the elements
+    for (auto __i = __tid; __i < __partial_tile_total; __i += __block_stride)
+    {
+      __tile_extent_t __linear = __i;
+      ::cuda::std::array<__tile_extent_t, __max_rank> __coords;
+      _CCCL_PRAGMA_UNROLL_FULL()
+      for (int __k = 0; __k < __max_rank; ++__k)
+      {
+        __coords[__k] = __linear % __partial_tile_sizes[__k];
+        __linear /= __partial_tile_sizes[__k];
```
Jacobfaib, line 215: Is it possible to use the the fast_div_mod here? Compilers will probably struggle on this

fbusato: unfortunately no, `__partial_tile_sizes` can be only be compute at runtime because depends on thread indexing

miscco (approved):

Issue #7639, opened 2026-02-11
Implement the new tuning API for `DispatchMergeSort`


bernhardmgruber:

jrhemstad:
Once this one is done, it'd be great to sync with RAPIDS folks like @bdice and @davidwendt to give them a better solution to control unrolling than what we did in https://github.com/NVIDIA/cccl/pull/4253

PR #8128, opened 2026-03-20
Implement the new tuning API for `DispatchMergeSort`

bernhardmgruber:
- [x] No SASS changes for `cub.bench.merge_sort.pairs.base` on SM120 (the only tuning is for SM60)

This PR also disables any support for the vsmem helper from CCCL.C, which is fine, as discussed internally, since it was incomplete and didn't work anyways. This will be revisited when we can JIT host code.

Fixes: #7639

### DIFF
--- a/cub/cub/device/dispatch/dispatch_merge_sort.cuh
+++ b/cub/cub/device/dispatch/dispatch_merge_sort.cuh
@@ -29,7 +29,7 @@
 
 #if !_CCCL_COMPILER(NVRTC) && defined(CUB_DEBUG_LOG)
 #  include <sstream>
-#endif
+#endif // !_CCCL_COMPILER(NVRTC) && defined(CUB_DEBUG_LOG)
 
 CUB_NAMESPACE_BEGIN
 
@@ -209,7 +209,7 @@ public:
       return cudaSuccess;
     }
 
-    const auto tile_size =
+    constexpr auto tile_size =
       detail::merge_sort::merge_sort_vsmem_helper_t<
         policy_getter<ActivePolicyT>,
         KeyInputIteratorT,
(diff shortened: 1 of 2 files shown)


cub/cub/device/dispatch/dispatch_merge_sort.cuh
```diff
@@ -95,15 +104,14 @@ template <typename KeyInputIteratorT,
...
-            typename PolicyHub::MaxPolicy,
+            detail::merge_sort::policy_selector_from_hub<PolicyHub>,
             KeyInputIteratorT,
             ValueInputIteratorT,
             KeyIteratorT,
             ValueIteratorT,
             OffsetT,
             CompareOpT>,
           typename KernelLauncherFactory = CUB_DETAIL_DEFAULT_KERNEL_LAUNCHER_FACTORY,
-          typename VSMemHelperT          = detail::merge_sort::VSMemHelper,
```
bernhardmgruber, line 106: We could consider this a breaking change, but this parameter was never intended to be customized by the user. It was only intended for CCCL.C.

cub/cub/device/dispatch/tuning/tuning_merge_sort.cuh
```diff
@@ -13,46 +13,41 @@
...
+#include <cub/block/block_load.cuh>
+#include <cub/block/block_store.cuh>
+#include <cub/device/dispatch/tuning/common.cuh>
+#include <cub/thread/thread_load.cuh>
 #include <cub/util_device.cuh>
+#include <cub/util_math.cuh>
 
-CUB_NAMESPACE_BEGIN
+#include <cuda/__device/arch_id.h>
+#include <cuda/std/concepts>
```
miscco: Is that really needed?

cub/cub/device/dispatch/dispatch_merge_sort.cuh
```diff
@@ -27,11 +27,15 @@
 #include <cuda/std/__type_traits/is_same.h>
 #include <cuda/std/cstdint>
 
+#if !_CCCL_COMPILER(NVRTC) && defined(CUB_DEBUG_LOG)
+#  include <sstream>
+#endif
```
miscco:
```suggestion
#endif // !_CCCL_COMPILER(NVRTC) && defined(CUB_DEBUG_LOG)
```

cub/cub/device/dispatch/dispatch_merge_sort.cuh
```diff
@@ -189,17 +209,17 @@ struct DispatchMergeSort
...
-      typename ActivePolicyT::MergeSortPolicy,
-      KeyInputIteratorT,
-      ValueInputIteratorT,
-      KeyIteratorT,
-      ValueIteratorT,
-      OffsetT,
-      CompareOpT,
-      KeyT,
-      ValueT>(wrapped_policy.MergeSort());
+    const auto tile_size =
```
miscco: Could that be `constexpr` or is that restricted due to CCCL.C

cub/cub/device/dispatch/dispatch_merge_sort.cuh
```diff
@@ -210,30 +230,31 @@ struct DispatchMergeSort
...
+        policy_getter<ActivePolicyT>,
         KeyInputIteratorT,
         ValueInputIteratorT,
         KeyIteratorT,
         ValueIteratorT,
         OffsetT,
         CompareOpT,
         KeyT,
-        ValueT>(wrapped_policy.MergeSort());
+        ValueT>::block_sort_agent_t>::vsmem_per_block;
```
miscco: Could we pull that out into a variable?

cub/cub/device/dispatch/dispatch_merge_sort.cuh
bernhardmgruber: It can I think.

cub/cub/device/dispatch/tuning/tuning_merge_sort.cuh
bernhardmgruber: We don't need `<concepts>` I think, since `_CCCL_HAS_CONCEPTS()` should be available regardless

cub/cub/device/dispatch/dispatch_merge_sort.cuh
bernhardmgruber: Refactored.

cub/cub/device/dispatch/dispatch_merge_sort.cuh
```diff
@@ -426,4 +438,218 @@ struct DispatchMergeSort
...
+  KeyT*                                  = nullptr /* for CCCL.C */,
+  ValueT*                                = nullptr /* for CCCL.C */) -> cudaError_t
+{
+  [[maybe_unused]] constexpr bool keys_only = ::cuda::std::is_same_v<ValueT, NullType>;
+
+  if (num_items == 0)
+  {
+    if (d_temp_storage == nullptr)
+    {
+      temp_storage_bytes = 0;
```
NaderAlAwar, line 485: Important: I think that this should be at least 1. I know existing behavior sets it to 0, but in general, even when we don't require any temporary storage, we return 1.

bernhardmgruber: I also think so. I retained it from the old implementation, but I will file a new PR to change it.

bernhardmgruber: #8152

PR #8131, opened 2026-03-22
unify float comparison helpers with isclose

### LINKED ISSUE #7662: Harmonize methods used for floating-point approximate equality testing across CCCL test suite
Presently, CCCL uses different tests for approximate equality of compute result against a reference value.

CCCL uses the following to test proximity to reference result

$$
\vert a - b \vert \le a_\{\mathrm{tol}} + r_{\mathrm{tol}} \cdot \mathrm{scale}(a, b)
$$

where $a_{\mathrm{tol}}$ and $r_{\mathrm{tol}}$ values used across different constituents of CCCL may differ. Moreover, $\mathrm{scale}(a, b)$ is sometimes defined as $\vert a \vert + \vert b \vert$, and sometimes as $\vert b \vert$.

The latter choice makes the testing function $\mathrm{isclose}(a, b)$ non-symmetric and should be avoided when testing value computed on GPU against value computed on CPU (as opposed to checking for reference value known a-priori).

[PEP-0485](https://peps.python.org/pep-0485/) proposes to use the following test, which is explicitly symmetric and has intuitive meaning of tolerance parameters:

$$
   \mathrm{isclose}(a, b) := (\vert a - b \vert \le \max( a_{\mathrm{tol}}, r_{\mathrm{tol}} \cdot \max( \vert a \vert, \vert b \vert )) )
$$ 

Furthermore, choice of tolerance parameters should depend on `std::numeric_limits<FpT>::epsilon()` and be based on Wilkinson-type error estimates derived for the pertinent algorithm.

NitishNaineni:
## Description

closes #7662

replaced 6 different float comparison implementations across cub/thrust/c2h with one `isclose` function using the PEP-0485 symmetric formula: `|a - b| <= max(a_tol, r_tol * max(|a|, |b|))`

what changed:
- added `c2h/include/c2h/isclose.h` with the canonical isclose function
- rewrote `REQUIRE_APPROX_EQ*` macros to use isclose
- replaced `CompareResults` float/double in test_util.h
- replaced `WithinRel` calls in check_results.cuh and thread_reduce tests
- replaced `require_almost_equal` in thrust complex tests
- replaced hand-rolled inline comparison in segmented scan tests

default tolerance is `1000 * numeric_limits<T>::epsilon()` which is derived from the type's precision instead of hardcoded magic numbers. callers can still pass custom tolerances when needed.

left per-algorithm wilkinson tolerance tuning and libcudacxx/cudax test unification for follow-ups.

## Checklist

### DIFF
--- a/c2h/include/c2h/check_results.cuh
+++ b/c2h/include/c2h/check_results.cuh
@@ -71,36 +71,28 @@ void verify_results(const c2h::host_vector<T>& expected_data, const c2h::host_ve
   else if constexpr (cuda::std::is_same_v<T, cuda::std::complex<__nv_bfloat16>>
                      || cuda::std::is_same_v<T, cuda::std::complex<__half>>)
   {
-    using real_t             = decltype(expected_data[0].real());
+    using real_t             = typename T::value_type;
     constexpr real_t rel_err = cuda::std::is_same_v<T, cuda::std::complex<__half>> ? real_t{0.08f} : real_t{0.2f};
     for (size_t i = 0; i < test_results.size(); ++i)
     {
-      bool close_real = isclose(expected_data[i].real(), test_results[i].real(), rel_err);
-      bool close_imag = isclose(expected_data[i].imag(), test_results[i].imag(), rel_err);
-      if (!close_real || !close_imag)
+      bool close = isclose(expected_data[i], test_results[i], rel_err);
+      if (!close)
       {
         INFO("index " << i);
       }
-      REQUIRE(close_real);
-      REQUIRE(close_imag);
+      REQUIRE(close);
     }
   }
   else if constexpr (cuda::std::__is_cuda_std_complex_v<T>)
   {
     for (size_t i = 0; i < test_results.size(); ++i)
     {
-      auto expected_real = expected_data[i].real();
-      auto test_real     = test_results[i].real();
-      auto expected_imag = expected_data[i].imag();
-      auto test_imag     = test_results[i].imag();
-      bool close_real    = isclose(expected_real, test_real);
-      bool close_imag    = isclose(expected_imag, test_imag);
-      if (!close_real || !close_imag)
+      bool close = isclose(expected_data[i], test_results[i]);
+      if (!close)
       {
         INFO("index " << i);
       }
-      REQUIRE(close_real);
-      REQUIRE(close_imag);
+      REQUIRE(close);
     }
   }
   else
--- a/c2h/include/c2h/check_results.cuh
+++ b/c2h/include/c2h/check_results.cuh
@@ -35,7 +35,7 @@ void verify_results(const c2h::host_vector<T>& expected_data, const c2h::host_ve
   }
   else if constexpr (cuda::std::is_same_v<T, __nv_bfloat16> || cuda::std::is_same_v<T, __half>)
   {
-    constexpr auto rel_err = cuda::std::is_same_v<T, __half> ? 0.08f : 0.2f;
+    constexpr T rel_err = cuda::std::is_same_v<T, __half> ? T{0.08f} : T{0.2f};
     REQUIRE_APPROX_EQ_EPSILON(expected_data, test_results, rel_err);
   }
   else if constexpr (cuda::std::is_same_v<T, float2>)
@@ -54,13 +54,12 @@ void verify_results(const c2h::host_vector<T>& expected_data, const c2h::host_ve
   }
   else if constexpr (cuda::std::is_same_v<T, __nv_bfloat162> || cuda::std::is_same_v<T, __half2>)
   {
-    constexpr auto rel_err = cuda::std::is_same_v<T, __half2> ? 0.08f : 0.2f;
+    using elem_t           = decltype(expected_data[0].x);
+    constexpr elem_t rel_err = cuda::std::is_same_v<T, __half2> ? elem_t{0.08f} : elem_t{0.2f};
     for (size_t i = 0; i < test_results.size(); ++i)
     {
-      bool close_x =
-        isclose(static_cast<float>(expected_data[i].x), static_cast<float>(test_results[i].x), rel_err);
-      bool close_y =
-        isclose(static_cast<float>(expected_data[i].y), static_cast<float>(test_results[i].y), rel_err);
+      bool close_x = isclose(expected_data[i].x, test_results[i].x, rel_err);
+      bool close_y = isclose(expected_data[i].y, test_results[i].y, rel_err);
       if (!close_x || !close_y)
       {
         INFO("index " << i);
@@ -72,15 +71,12 @@ void verify_results(const c2h::host_vector<T>& expected_data, const c2h::host_ve
   else if constexpr (cuda::std::is_same_v<T, cuda::std::complex<__nv_bfloat16>>
                      || cuda::std::is_same_v<T, cuda::std::complex<__half>>)
   {
-    constexpr auto rel_err = cuda::std::is_same_v<T, cuda::std::complex<__half>> ? 0.08f : 0.2f;
+    using real_t           = decltype(expected_data[0].real());
+    constexpr real_t rel_err = cuda::std::is_same_v<T, cuda::std::complex<__half>> ? real_t{0.08f} : real_t{0.2f};
     for (size_t i = 0; i < test_results.size(); ++i)
     {
-      auto expected_real = static_cast<float>(expected_data[i].real());
-      auto test_real     = static_cast<float>(test_results[i].real());
-      auto expected_imag = static_cast<float>(expected_data[i].imag());
-      auto test_imag     = static_cast<float>(test_results[i].imag());
-      bool close_real    = isclose(expected_real, test_real, rel_err);
-      bool close_imag    = isclose(expected_imag, test_imag, rel_err);
+      bool close_real = isclose(expected_data[i].real(), test_results[i].real(), rel_err);
+      bool close_imag = isclose(expected_data[i].imag(), test_results[i].imag(), rel_err);
       if (!close_real || !close_imag)
       {
         INFO("index " << i);
--- a/c2h/include/c2h/check_results.cuh
+++ b/c2h/include/c2h/check_results.cuh
@@ -11,7 +11,7 @@
 
 #include <test_util.h>
 
-#include <catch2/matchers/catch_matchers_floating_point.hpp>
+#include <c2h/isclose.h>
 
 template <typename T>
 void verify_results(const c2h::host_vector<T>& expected_data, const c2h::host_vector<T>& test_results)
@@ -42,8 +42,9 @@ void verify_results(const c2h::host_vector<T>& expected_data, const c2h::host_ve
   {
     for (size_t i = 0; i < test_results.size(); ++i)
     {
-      REQUIRE_THAT(expected_data[i].x, Catch::Matchers::WithinRel(test_results[i].x, 0.01f));
-      REQUIRE_THAT(expected_data[i].y, Catch::Matchers::WithinRel(test_results[i].y, 0.01f));
+      INFO("index " << i);
+      REQUIRE(isclose(expected_data[i].x, test_results[i].x, 0.01f));
+      REQUIRE(isclose(expected_data[i].y, test_results[i].y, 0.01f));
     }
   }
   else if constexpr (cuda::std::is_same_v<T, __nv_bfloat162> || cuda::std::is_same_v<T, __half2>)
@@ -51,8 +52,9 @@ void verify_results(const c2h::host_vector<T>& expected_data, const c2h::host_ve
     constexpr auto rel_err = cuda::std::is_same_v<T, __half2> ? 0.08f : 0.2f;
     for (size_t i = 0; i < test_results.size(); ++i)
     {
-      REQUIRE_THAT(expected_data[i].x, Catch::Matchers::WithinRel(test_results[i].x, rel_err));
-      REQUIRE_THAT(expected_data[i].y, Catch::Matchers::WithinRel(test_results[i].y, rel_err));
+      INFO("index " << i);
+      REQUIRE(isclose(static_cast<float>(expected_data[i].x), static_cast<float>(test_results[i].x), rel_err));
+      REQUIRE(isclose(static_cast<float>(expected_data[i].y), static_cast<float>(test_results[i].y), rel_err));
     }
   }
   else if constexpr (cuda::std::is_same_v<T, cuda::std::complex<__nv_bfloat16>>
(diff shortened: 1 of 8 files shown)


bernhardmgruber:
Thanks for this contribution! 

@oleksandr-pavlyk and @fbusato can you take a look at this?

c2h/include/c2h/isclose.h
```diff
@@ -0,0 +1,45 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION. All rights reserved.
+// SPDX-License-Identifier: BSD-3
```
oleksandr-pavlyk:
```suggestion
// SPDX-License-Identifier: BSD-3-Clause
```

I see that some headers in `c2h` use `BSD-3`, but the documented identifier is [BSD-3-Clause](https://spdx.org/licenses/BSD-3-Clause.html).

c2h/include/c2h/isclose.h
```diff
@@ -0,0 +1,45 @@
...
+  {
+    if (a == b)
+    {
+      return true;
+    }
+    return std::abs(a - b) <= std::max(a_tol, r_tol * std::max(std::abs(a), std::abs(b)));
+  }
+  else
+  {
+    return a == b;
```
oleksandr-pavlyk, line 34: If `isclose` is used on a complex type, it would use `a == b` branch, perhaps unexpectedly to the user. Perhaps it is worth adding `static_assert(std::is_integral_v<T>, "Non-integral type using exact comparison");`,  or perhaps a static assertion that type `T` does not expose `real` and `imag` methods.

c2h/include/c2h/isclose.h
```diff
@@ -0,0 +1,45 @@
...
+{
+  return isclose(a, b, r_tol, T(0));
+}
+
+template <typename T>
+bool isclose(T a, T b)
+{
+  if constexpr (std::is_floating_point_v<T>)
+  {
+    return isclose(a, b, T(1000) * std::numeric_limits<T>::epsilon(), T(0));
```
oleksandr-pavlyk: I would use powers of 2 here. `T( 1 << 10 )` would mean discrepancy of 10 binary [ULP](https://en.wikipedia.org/wiki/Unit_in_the_last_place)s is tolerated. 

10 binary units is about half of 23 total explicit mantissa units for single precision floating point numbers, but is all most of 11 total explicit half-precision explicit mantissa units.

Given that `std::is_floating_point_v<std::float16_t>` is `true`, perhaps we should use `T(1 << 8)` as default multiplier instead?

c2h/include/c2h/check_results.cuh
```diff
@@ -77,8 +80,9 @@ void verify_results(const c2h::host_vector<T>& expected_data, const c2h::host_ve
       auto test_real     = test_results[i].real();
       auto expected_imag = expected_data[i].imag();
       auto test_imag     = test_results[i].imag();
-      REQUIRE_THAT(expected_real, Catch::Matchers::WithinRel(test_real));
-      REQUIRE_THAT(expected_imag, Catch::Matchers::WithinRel(test_imag));
+      INFO("index " << i);
```
oleksandr-pavlyk: There is a lot of `INFO` added which would litter the output on failure in comparisons of large arrays.

I think it would be useful to save values of `isclose` checks, and place `INFO("index" << i);` in the branch executed only if `isclose` check is not met. `REQUIRE` would reuse the result of testing.

cub/test/thread_reduce/catch2_test_thread_reduce.cu
```diff
@@ -330,7 +323,8 @@ C2H_TEST("ThreadReduce Narrow PrecisionType Tests",
     auto reference_result =
       std::accumulate(h_in_float.begin(), h_in_float.begin() + num_items, operator_identity, std_reduce_op);
     run_thread_reduce_kernel(num_items, d_in, d_out, reduce_op);
-    verify_results(reference_result, float{c2h::host_vector<value_t>(d_out)[0]});
+    float test_result = float{c2h::host_vector<value_t>(d_out)[0]};
```
oleksandr-pavlyk: Nit:

```suggestion
    float test_result{c2h::host_vector<value_t>(d_out)[0]};
```

thrust/testing/catch2_test_complex.cu
```diff
@@ -42,30 +43,24 @@ struct other_floating_point_type<double>
...
 inline constexpr bool is_complex<std::complex<T>> = true;
 
-// Overload for complex types
 template <typename T1, typename T2>
 ::cuda::std::enable_if_t<is_complex<T1> && is_complex<T2>> require_almost_equal(const T1& a, const T2& b)
 {
-  CHECK(a.real() == Catch::Approx(b.real()).margin(DEFAULT_ABSOLUTE_TOL).epsilon(DEFAULT_RELATIVE_TOL));
-  CHECK(a.imag() == Catch::Approx(b.imag()).margin(DEFAULT_ABSOLUTE_TOL).epsilon(DEFAULT_RELATIVE_TOL));
+  CHECK(isclose(static_cast<double>(a.real()), static_cast<double>(b.real())));
+  CHECK(isclose(static_cast<double>(a.imag()), static_cast<double>(b.imag())));
```
oleksandr-pavlyk: Why cast to `double`?  The cast results in relative tolerance specific to double precision to be used when comparing complex values with less precise real/imaginary types. 

Is anything wrong with checking `isclose(a.rea(), b.real())` and `isclose(a.imag(), b.imag())`?

oleksandr-pavlyk: If the issue is that a common type is needed, perhaps use `std::common_type<T1, T2>`: https://en.cppreference.com/w/cpp/types/common_type.html

thrust/testing/catch2_test_complex.cu
```diff
@@ -42,30 +43,24 @@ struct other_floating_point_type<double>
...
 }
 
 template <typename T1, typename T2>
 ::cuda::std::enable_if_t<!is_complex<T1> && !is_complex<T2>> require_almost_equal(const T1& a, const T2& b)
 {
-  CHECK(a == Catch::Approx(b).margin(DEFAULT_ABSOLUTE_TOL).epsilon(DEFAULT_RELATIVE_TOL));
+  CHECK(isclose(static_cast<double>(a), static_cast<double>(b)));
```
oleksandr-pavlyk: Same question here.

c2h/include/c2h/isclose.h
```diff
@@ -0,0 +1,45 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION. All rights reserved.
+// SPDX-License-Identifier: BSD-3-Clause
```
bernhardmgruber: Important: Any net-new source file should be under `Apache-2.0 WITH LLVM-exception`
```suggestion
// SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
```

c2h/include/c2h/isclose.h
NitishNaineni: good point, switched to T(1 << 8)

c2h/include/c2h/isclose.h
NitishNaineni: i have added an assert

c2h/include/c2h/check_results.cuh
NitishNaineni: made INFO conditional on failed check

NitishNaineni:
reverted the thrust complex test changes for now. isclose lives in c2h which is only accessible from cub and c/parallel tests. thrust, libcudacxx, and cudax tests can't reach it. is there a shared location where a test utility like this could live so all test frameworks can use it?

c2h/include/c2h/isclose.h
```diff
@@ -0,0 +1,47 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#pragma once
+
+#include <algorithm>
```
fbusato: please use `cuda/std` headers

c2h/include/c2h/isclose.h
```diff
@@ -0,0 +1,47 @@
...
+#include <cmath>
+#include <limits>
+#include <type_traits>
+
+template <typename T>
+bool isclose(T a, T b, T r_tol, T a_tol)
+{
+  if constexpr (std::is_floating_point_v<T>)
+  {
+    if (a == b)
```
fbusato: this shortcut is not strictly needed

c2h/include/c2h/isclose.h
```diff
@@ -0,0 +1,47 @@
...
+
+#include <algorithm>
+#include <cmath>
+#include <limits>
+#include <type_traits>
+
+template <typename T>
+bool isclose(T a, T b, T r_tol, T a_tol)
+{
+  if constexpr (std::is_floating_point_v<T>)
```
fbusato: what about infinity and Nan?

c2h/include/c2h/isclose.h
```diff
@@ -0,0 +1,47 @@
...
+  {
+    static_assert(std::is_integral_v<T>, "isclose: unsupported type, expected floating point or integral");
+    return a == b;
+  }
+}
+
+template <typename T>
+bool isclose(T a, T b, T r_tol)
+{
+  return isclose(a, b, r_tol, T(0));
```
fbusato:
```suggestion
  return isclose(a, b, r_tol, T{});
```

c2h/include/c2h/isclose.h
fbusato: what about reduced-precision floating-point types? e.g. half, bfloat16

c2h/include/c2h/isclose.h
```diff
@@ -0,0 +1,47 @@
...
+}
+
+template <typename T>
+bool isclose(T a, T b)
+{
+  if constexpr (std::is_floating_point_v<T>)
+  {
+    return isclose(a, b, T(1 << 8) * std::numeric_limits<T>::epsilon(), T(0));
```
fbusato: `2^8 * e`  uhmm, for all types?

c2h/include/c2h/isclose.h
```diff
@@ -0,0 +1,47 @@
...
+
+#pragma once
+
+#include <algorithm>
+#include <cmath>
+#include <limits>
+#include <type_traits>
+
+template <typename T>
+bool isclose(T a, T b, T r_tol, T a_tol)
```
fbusato: please add an assertion to check that `r_tol` and `a_tol` are finite and positive

fbusato (changes requested):

c2h/include/c2h/isclose.h
NitishNaineni: handling inf and nan cases more explicitly and added support for half and bfloat16 in isclose via cuda::is_floating_point_v. the callers were previously converting them to float

c2h/include/c2h/isclose.h
NitishNaineni: scaled the multiplier to 1 << (digits/4) so it adapts per type. it's a reasonable default, callers with specific accuracy requirements should pass explicit tolerances. let me know if you want this handled differently

NitishNaineni:
addressed all feedback:
- switched to cuda/std headers
- replaced a == b shortcut with explicit NaN and inf checks
- use T{} instead of T(0)
- added assertions for r_tol and a_tol being finite and non-negative
- added support for half/bfloat16 via cuda::is_floating_point_v, removed float casts from callers
- scaled default tolerance per type using digits/4 instead of fixed 1 << 8

some of the changes like the default tolerance i am not sure about. let me know if these needs to be handled differently

c2h/include/c2h/isclose.h
```diff
@@ -0,0 +1,56 @@
...
+  return isclose(a, b, r_tol, T{});
+}
+
+template <typename T>
+bool isclose(T a, T b)
+{
+  if constexpr (cuda::is_floating_point_v<T>)
+  {
+    return isclose(
+      a, b, T(1 << (cuda::std::numeric_limits<T>::digits / 4)) * cuda::std::numeric_limits<T>::epsilon(), T{});
```
fbusato: prefer unsigned shift
```suggestion
      a, b, T(1u << (cuda::std::numeric_limits<T>::digits / 4)) * cuda::std::numeric_limits<T>::epsilon(), T{});
```

c2h/include/c2h/isclose.h
```diff
@@ -0,0 +1,56 @@
...
+#include <cuda/std/algorithm>
+#include <cuda/std/cassert>
+#include <cuda/std/cmath>
+#include <cuda/std/limits>
+#include <cuda/std/type_traits>
+
+template <typename T>
+bool isclose(T a, T b, T r_tol, T a_tol)
+{
+  if constexpr (cuda::is_floating_point_v<T>)
```
fbusato, line 17: could you please add a branch for complex number (`cuda::std::complex` and `cuda::complex`)

cub/test/test_util.h
```diff
@@ -1074,22 +1075,17 @@ int CompareResults(float* computed, float* reference, OffsetT len, bool verbose
...
       {
-        if (verbose)
-        {
-          std::cout
-            << "INCORRECT: [" << i << "]: "
-            << "(computed) " << CoutCast(computed[i]) << " != " << CoutCast(reference[i])
-            << " (difference:" << difference << ", fraction: " << fraction << ")";
-        }
-        return 1;
+        float difference = std::abs(computed[i] - reference[i]);
```
fbusato:
```suggestion
        float difference = cuda::std::abs(computed[i] - reference[i]);
```

c2h/include/c2h/isclose.h
fbusato: we use `r_tol` and `a_tol` only for floating-point types. Add `[[maybe_unused]]`

fbusato (changes requested):

NitishNaineni:
addressed latest feedback:
- unsigned shift 1u << for default tolerance
- cuda::std::abs in test_util.h
- [[maybe_unused]] on r_tol and a_tol                                                                                                                                                                                                                    
- added cuda::std::complex overloads for isclose, simplified complex branches in check_results.cuh to use them

c2h/include/c2h/isclose.h
```diff
@@ -46,11 +47,29 @@ bool isclose(T a, T b)
...
   }
   else
   {
     static_assert(cuda::std::is_integral_v<T>, "isclose: unsupported type, expected floating point or integral");
     return a == b;
   }
 }
+
+template <typename T>
+bool isclose(cuda::std::complex<T> a, cuda::std::complex<T> b, T r_tol, T a_tol)
```
fbusato, line 60: please add the `complex` case to the main function with `if constexpr`.

c2h/include/c2h/isclose.h
```diff
@@ -46,11 +47,29 @@ bool isclose(T a, T b)
...
+{
+  return isclose(a.real(), b.real(), r_tol, a_tol) && isclose(a.imag(), b.imag(), r_tol, a_tol);
```
fbusato, line 62: comparing real and imaginary parts separately is a weak condition. Please use `isclose(|a|, |b|)`

c2h/include/c2h/isclose.h
```diff
@@ -46,11 +47,29 @@ bool isclose(T a, T b)
...
+}
+
+template <typename T>
+bool isclose(cuda::std::complex<T> a, cuda::std::complex<T> b, T r_tol)
```
fbusato, line 66: please also handle `cuda::complex`

fbusato (changes requested):

PR #8132, opened 2026-03-22
Enable strict C++ compiler warnings and fix all violations

maxwbuckley:
## Summary

- Add 4 new strict warning flags beyond the existing `-Wall -Wextra -Werror`: `-Wshadow=local`, `-Wnon-virtual-dtor`, `-Wdouble-promotion`, `-Wformat=2`
- Fix all resulting warnings across CUB, Thrust, libcudacxx, cudax, nvbench_helper, and C Parallel (25+ files)
- Guard Clang-only `-Wgnu` / `-Wno-gnu-*` flags behind a compiler check to fix GCC9 builds
- `-Wpedantic` was evaluated but excluded — nvcc's preprocessor emits GCC-style line directives that trigger unavoidable pedantic warnings
- `-Wnull-dereference` was evaluated but excluded — GCC produces many false positives in system headers (e.g. `streambuf`) when inlining user code

### New warnings

| Warning | What it catches |
|---|---|
| `-Wshadow=local` | Local variable shadowing another local (dangerous scoping bugs) |
| `-Wnon-virtual-dtor` | Classes with virtual functions but non-virtual destructors |
| `-Wdouble-promotion` | Implicit `float` → `double` promotion (performance/precision bugs) |
| `-Wformat=2` | Format string security issues |

### Flags evaluated but excluded

| Warning | Why excluded |
|---|---|
| `-Wpedantic` | nvcc's preprocessor emits GCC-style line directives that trigger unavoidable warnings |
| `-Wnull-dereference` | GCC produces false positives in system headers (e.g. `<streambuf>`) when inlining user code |

### Fixes applied

**`-Wdouble-promotion` (implicit float→double):**
- `libcudacxx` nvbf16.h / nvfp16.h: wrap `__bfloat162float()` / `__half2float()` in `static_cast<double>()` in `complex<double>` constructors and assignments
- `libcudacxx` native_type.h: use `__float128{}` instead of `__float128{0.0}` to avoid double→\_\_float128 promotion
- `thrust` complex arithmetic.h: cast all mixed-type operands to `common_type` before arithmetic
- `thrust` complex.inl: same `common_type` cast pattern in equality operators
- `thrust` catch2_test_complex.cu: cast to `double` before CHECK comparisons of mixed float/double components
- `nvbench_helper` nvbench_helper.cu: fix float→double in `random_to_item_t` calls and distribution constructors
- `nvbench_helper` gen_uniform_distribution.cu: explicit casts in mixed-type arithmetic
- `cudax` 09-nbody-blocked.cu: cast float `elapsed` to double for printf
- `cub` test_allocator.cu, test_util.h, cpu_timer.h: explicit casts for printf and float comparisons

**`-Wshadow=local` (variable shadowing):**
- `cub` catch2_test_env_launch_helper.h: `args` → `inner_args` in nested lambdas
- `cub` catch2_test_device_transform_if.cu: `a`/`b` → `lhs`/`rhs`
- `cudax` test_bulk.cu, test_conditional.cu, test_stream_context.cu, test_visit.cu, launch_smoke.cu: renamed shadowed variables (`counter`→`ctr`, `vals`→`arr`, `config`→`smem_config`, etc.)
- `cudax` sections_2.cu: inner `s_bar` → `s_baz`
- `nvbench_helper` nvbench_helper.cu: `dist` → `float_dist` / `double_dist`
- `nvbench_helper` gen_entropy.cu: lambda parameter `entropy` → `level`

**`-Wnon-virtual-dtor`:**
- `c2h` catch2_test_helper.h: changed `~QuietMatchExpr() override` to `virtual ~QuietMatchExpr()` — base class `ITransientExpression` (Catch2) has no virtual destructor, so `override` is invalid

**Null-dereference hardening (defensive, not required by a flag):**
- `libcudacxx` basic_any_value.h, virtcall.h: added `_CCCL_ASSERT` + `_CCCL_ASSUME` for `__query_interface` return values
- `cudax` executable_graph_cache.cuh: added `_CCCL_ASSERT` for LRU cache iterator validity
- `cub` catch2_test_device_decoupled_look_back.cu: added `REQUIRE` for pointer validity

**CMake infrastructure:**
- Guarded `-Wgnu`, `-Wno-gnu-line-marker`, `-Wno-gnu-zero-variadic-macro-arguments` behind `if (Clang)` — GCC (especially GCC9) rejects unrecognized `-Wno-*` flags under `-Werror`

## Test plan

- [ ] CI matrix validation across all compiler/toolkit combinations
- [ ] CUB builds (GCC, Clang, ClangCUDA, nvcc) — zero new warnings
- [ ] Thrust builds (GCC, Clang, ClangCUDA, nvcc) — zero new warnings
- [ ] cudax builds (GCC, Clang) — zero new warnings
- [ ] libcudacxx builds — zero new warnings
- [ ] C Parallel builds — zero new warnings
- [ ] NVBench helper builds — zero new warnings

### DIFF
--- a/libcudacxx/include/cuda/__utility/__basic_any/basic_any_value.h
+++ b/libcudacxx/include/cuda/__utility/__basic_any/basic_any_value.h
@@ -21,12 +21,6 @@
 #  pragma system_header
 #endif // no system header
 
-// GCC's -Wnull-dereference fires false positives on __query_interface() return
-// values, which are guaranteed non-null by the library's invariants when called
-// on a valid vptr.
-_CCCL_DIAG_PUSH
-_CCCL_DIAG_SUPPRESS_GCC("-Wnull-dereference")
-
 #include <cuda/__utility/__basic_any/basic_any_base.h>
 #include <cuda/__utility/__basic_any/basic_any_fwd.h>
 #include <cuda/__utility/__basic_any/basic_any_ref.h>
@@ -354,9 +348,12 @@ public:
   {
     if (auto __vptr = __get_vptr())
     {
-      _CCCL_ASSERT(__vptr->__query_interface(__iunknown())->__cookie_ == 0xDEADBEEF,
+      auto* __unk = __vptr->__query_interface(__iunknown());
+      _CCCL_ASSERT(__unk != nullptr, "query_interface returned a null pointer");
+      _CCCL_ASSUME(__unk != nullptr);
+      _CCCL_ASSERT(__unk->__cookie_ == 0xDEADBEEF,
                    "query_interface returned a bad pointer to the __iunknown vtable");
-      __vptr->__query_interface(__iunknown())->__dtor_(__buffer_, __in_situ());
+      __unk->__dtor_(__buffer_, __in_situ());
       __release_();
     }
   }
@@ -367,9 +364,12 @@ public:
   {
     if (auto __vptr = __get_vptr())
     {
-      _CCCL_ASSERT(__vptr->__query_interface(__iunknown())->__cookie_ == 0xDEADBEEF,
+      auto* __unk = __vptr->__query_interface(__iunknown());
+      _CCCL_ASSERT(__unk != nullptr, "query_interface returned a null pointer");
+      _CCCL_ASSUME(__unk != nullptr);
+      _CCCL_ASSERT(__unk->__cookie_ == 0xDEADBEEF,
                    "query_interface returned a bad pointer to the __iunknown vtable");
-      return *__vptr->__query_interface(__iunknown())->__object_info_->__object_typeid_;
+      return *__unk->__object_info_->__object_typeid_;
     }
     return _CCCL_TYPEID(void);
   }
@@ -567,8 +567,6 @@ private:
 };
 
 _CCCL_END_NAMESPACE_CUDA
-
-_CCCL_DIAG_POP
 
 #include <cuda/std/__cccl/epilogue.h>
 
--- a/libcudacxx/include/cuda/__utility/__basic_any/basic_any_value.h
+++ b/libcudacxx/include/cuda/__utility/__basic_any/basic_any_value.h
@@ -351,8 +351,7 @@ public:
       auto* __unk = __vptr->__query_interface(__iunknown());
       _CCCL_ASSERT(__unk != nullptr, "query_interface returned a null pointer");
       _CCCL_ASSUME(__unk != nullptr);
-      _CCCL_ASSERT(__unk->__cookie_ == 0xDEADBEEF,
-                   "query_interface returned a bad pointer to the __iunknown vtable");
+      _CCCL_ASSERT(__unk->__cookie_ == 0xDEADBEEF, "query_interface returned a bad pointer to the __iunknown vtable");
       __unk->__dtor_(__buffer_, __in_situ());
       __release_();
     }
@@ -367,8 +366,7 @@ public:
       auto* __unk = __vptr->__query_interface(__iunknown());
       _CCCL_ASSERT(__unk != nullptr, "query_interface returned a null pointer");
       _CCCL_ASSUME(__unk != nullptr);
-      _CCCL_ASSERT(__unk->__cookie_ == 0xDEADBEEF,
-                   "query_interface returned a bad pointer to the __iunknown vtable");
+      _CCCL_ASSERT(__unk->__cookie_ == 0xDEADBEEF, "query_interface returned a bad pointer to the __iunknown vtable");
       return *__unk->__object_info_->__object_typeid_;
     }
     return _CCCL_TYPEID(void);
--- a/libcudacxx/include/cuda/__utility/__basic_any/basic_any_value.h
+++ b/libcudacxx/include/cuda/__utility/__basic_any/basic_any_value.h
@@ -21,6 +21,12 @@
 #  pragma system_header
 #endif // no system header
 
+// GCC's -Wnull-dereference fires false positives on __query_interface() return
+// values, which are guaranteed non-null by the library's invariants when called
+// on a valid vptr.
+_CCCL_DIAG_PUSH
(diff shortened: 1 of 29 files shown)


libcudacxx/include/cuda/__utility/__basic_any/basic_any_value.h
```diff
@@ -348,9 +348,12 @@ struct _CCCL_TYPE_VISIBILITY_DEFAULT __basic_any : __basic_any_base<_Interface>
...
     if (auto __vptr = __get_vptr())
     {
-      _CCCL_ASSERT(__vptr->__query_interface(__iunknown())->__cookie_ == 0xDEADBEEF,
+      auto* __unk = __vptr->__query_interface(__iunknown());
+      _CCCL_ASSERT(__unk != nullptr, "query_interface returned a null pointer");
+      _CCCL_ASSUME(__unk != nullptr);
+      _CCCL_ASSERT(__unk->__cookie_ == 0xDEADBEEF,
                    "query_interface returned a bad pointer to the __iunknown vtable");
-      __vptr->__query_interface(__iunknown())->__dtor_(__buffer_, __in_situ());
+      __unk->__dtor_(__buffer_, __in_situ());
```
bernhardmgruber, line 355: I am curious, why do we need the `_CCCL_ASSUME` here? Have you observed any regression otherwise?

thrust/testing/unittest/assertions.h
```diff
@@ -9,9 +9,31 @@
...
 
+namespace unittest::detail
+{
+// Promote both operands to their common type before comparing, avoiding
+// implicit float-to-double promotion warnings (-Wdouble-promotion).
+// For non-arithmetic types, compare directly with no cast.
+template <typename T1, typename T2>
+auto promote(const T1& a, const T2& b)
+{
+  if constexpr (std::is_arithmetic_v<std::decay_t<T1>> && std::is_arithmetic_v<std::decay_t<T2>>)
```
bernhardmgruber: Please use types from libcu++

```suggestion
  if constexpr (cuda::std::is_arithmetic_v<std::decay_t<T1>> && cuda::std::is_arithmetic_v<std::decay_t<T2>>)
```
Applies everywhere. This ensures we have good coverage of our own facilities.

libcudacxx/include/cuda/__utility/__basic_any/basic_any_value.h
miscco: I believe its just a "trust me compiler" that helps avoid followup warnings

thrust/testing/unittest/assertions.h
```diff
@@ -9,9 +9,31 @@
 
 #include <cuda/std/utility>
 
+#include <type_traits>
```
miscco: Critical: We are using our traits in new code
```suggestion
#include <cuda/std/type_traits>
```

thrust/testing/unittest/assertions.h
maxwbuckley: Done, thanks!

maxwbuckley: Done, thanks!

thrust/testing/unittest/assertions.h
maxwbuckley: Done, thanks!

miscco:
pre-commit.ci autofix

thrust/testing/unittest/assertions.h
```diff
@@ -7,11 +7,32 @@
...
+// Promote both operands to their common type before comparing, avoiding
+// implicit float-to-double promotion warnings (-Wdouble-promotion).
+// For non-arithmetic types, compare directly with no cast.
+template <typename T1, typename T2>
+auto promote(const T1& a, const T2& b)
+{
+  if constexpr (cuda::std::is_arithmetic_v<cuda::std::decay_t<T1>> && cuda::std::is_arithmetic_v<cuda::std::decay_t<T2>>)
+  {
+    using common_t = cuda::std::common_type_t<cuda::std::decay_t<T1>, cuda::std::decay_t<T2>>;
+    return std::pair<common_t, common_t>{static_cast<common_t>(a), static_cast<common_t>(b)};
```
miscco: This is still using `std::pair` and not `cuda::std::pair` Also I believe we can just rely on CTAD to get the right type
```suggestion
    return cuda::std::pair{static_cast<common_t>(a), static_cast<common_t>(b)};
```

thrust/testing/unittest/assertions.h
```diff
@@ -7,11 +7,32 @@
...
+  }
+  else
+  {
+    return std::pair<const T1&, const T2&>{a, b};
```
miscco: This does not do what you want it to do I believe, we want to store values not references

```suggestion
    return cuda::std::pair{a, b};
```

miscco (changes requested):

thrust/testing/unittest/assertions.h
maxwbuckley: Good catch, thank you! Fixed in d3db815 — switched to `cuda::std::pair` with CTAD. All three backends (CUDA, OMP, TBB) build and pass `unittest_tester` cleanly.

maxwbuckley: Good catch, thank you! Fixed in d3db815 — switched to cuda::std::pair with CTAD. All three backends (CUDA, OMP, TBB) build and pass unittest_tester cleanly.

thrust/testing/unittest/assertions.h
maxwbuckley: You're right — storing references there was wrong. Fixed in the same commit, now stores values via cuda::std::pair{a, b}. Thanks for catching this!

miscco (approved):

miscco:
/ok to test [d3db815](https://github.com/NVIDIA/cccl/pull/8132/commits/d3db81563712d14d40c540a97fdb3cad06dcae08)

PR #8139, opened 2026-03-23
[CUB]: Provide workaround to compilation errors in ptx-json with older nvrtc versions

### LINKED ISSUE #8138: [BUG]: `ptx-json/value.cuh` fails to compile on NVRTC < 12.5
### Is this a duplicate?

### Type of Bug

Compile-time Error

### Component

CUB

### Describe the bug

`cub/detail/ptx-json/value.cuh` uses concept-constrained non-type template parameters in partial specializations, which NVRTC versions prior to 12.5 do not support. This causes `NVRTC_ERROR_COMPILATION` when any algorithm compiled with `-DCUB_ENABLE_POLICY_PTX_JSON` is JIT-compiled on CUDA toolkit 12.4 or earlier.

The two partial specializations in `value.cuh` use concept-constrained NTTPs:

```cpp
template <a_value auto Nested>        // line 28
struct value<Nested> { ... };

template <cuda::std::integral auto V> // line 48
struct value<V> { ... };
```

NVRTC < 12.5 rejects these with:

```
error: a template argument list is not allowed in a declaration of a primary template
  struct value<Nested>
         ^
```

NVRTC does not recognize the concept constraint as disambiguating these from the primary template declaration.

### How to Reproduce

```cpp
// repro.cpp
// Compile: nvcc -x cu -std=c++20 repro.cpp -lnvrtc -lcuda -o repro && ./repro

#include <nvrtc.h>
#include <cuda.h>
#include <cstdio>
#include <cstdlib>

#define CHECK_NVRTC(expr) do {                                              \
  nvrtcResult res = (expr);                                                 \
  if (res != NVRTC_SUCCESS) {                                               \
    fprintf(stderr, "NVRTC error %d at %s:%d\n", res, __FILE__, __LINE__);  \
    exit(1);                                                                \
  }                                                                         \
} while(0)

const char* src = R"(
template <auto V>
struct value;

template <typename T>
concept integral_like = requires { T(0); };

// Fails on NVRTC < 12.5:
template <integral_like auto V>
struct value<V> {
    static constexpr auto val = V;
};
)";

int main() {
    int major, minor;
    nvrtcVersion(&major, &minor);
    printf("NVRTC version: %d.%d\n", major, minor);

    nvrtcProgram prog;
    CHECK_NVRTC(nvrtcCreateProgram(&prog, src, "test.cu", 0, nullptr, nullptr));

    const char* opts[] = {"--std=c++20"};
    nvrtcResult compileRes = nvrtcCompileProgram(prog, 1, opts);

    size_t logSize;
    CHECK_NVRTC(nvrtcGetProgramLogSize(prog, &logSize));
    if (logSize > 1) {
        char* log = new char[logSize];
        CHECK_NVRTC(nvrtcGetProgramLog(prog, log));
        printf("%s\n", log);
        delete[] log;
    }

    printf("%s\n", compileRes == NVRTC_SUCCESS ? "PASS" : "FAIL");
    nvrtcDestroyProgram(&prog);
    return compileRes != NVRTC_SUCCESS;
}
```

### Expected behavior

This should work for all versions of CTK 12.

### Reproduction link

_No response_

### Operating System

_No response_

### nvidia-smi output

_No response_

### NVCC version

_No response_

NaderAlAwar:
## Description

closes #8138

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -511,7 +511,7 @@ Does your project use CCCL? [Open a PR to add your project to this list!](https:
 - [Qiskit](https://github.com/Qiskit/qiskit-aer) - High performance simulator for quantum circuits
 - [QUDA](https://github.com/lattice/quda) - Lattice quantum chromodynamics (QCD) computations
 - [RAFT](https://github.com/rapidsai/raft) - Algorithms and primitives for machine learning
-- [SGLang](https://github.com/sgl-project/sglang) - LLM serving framework 
+- [SGLang](https://github.com/sgl-project/sglang) - LLM serving framework
 - [TensorFlow](https://github.com/tensorflow/tensorflow) - End-to-end platform for machine learning
 - [TensorRT](https://github.com/NVIDIA/TensorRT) - Deep learning inference
 - [TensorRT-LLM](https://github.com/NVIDIA/TensorRT-LLM) - Optimized LLM inference


cub/cub/detail/ptx-json/value.cuh
```diff
@@ -25,7 +25,14 @@ struct is_value<value<V>> : cuda::std::true_type
...
 concept a_value = is_value<T>::value;
 
+// NVRTC < 12.5 does not support concept-constrained non-type template parameters
+// in partial specializations. Use requires clauses as a workaround.
+#if _CCCL_COMPILER(NVRTC, <, 12, 5)
+template <auto Nested>
+  requires a_value<decltype(Nested)>
+#else
 template <a_value auto Nested>
+#endif
```
miscco, line 35: Nitpick: There is no difference between the two formulations, so maybe lets just do
```suggestion
template <auto Nested>
  requires a_value<decltype(Nested)>
```

bernhardmgruber: I don't mind too much, since it's dead code anyway. When the tuning API rollout completes, this code is gone.

PR #8140, opened 2026-03-23
Supress vectorization false positive werrors in permutation_iterator test

gonidelis:
fixes the first bug in the **2-Thrust** section of https://nvbugspro.nvidia.com/bug/5999870

Detailed explanation of what gcc is messing up can be found in my comment under the nvbug or [here](https://gist.github.com/gonidelis/77c08a9c0baf529196c439fe591934d0)  

It's basically a false positive where vectorizer is smart enough to generate guarded wide stores (16-byte vector(16) stores into a 4-byte buffer), but the warning pass is too dumb to see the guard that makes it unreachable

We suppress vectorization for `TestPermutationIteratorWithCountingIterator()` as this is the test that uses `general_copy.h` underneath which results in the bogus vectorization werror.

### DIFF
--- a/thrust/testing/permutation_iterator.cu
+++ b/thrust/testing/permutation_iterator.cu
@@ -8,6 +8,12 @@
 
 #include <unittest/unittest.h>
 
+#if _CCCL_COMPILER(GCC, >=, 11)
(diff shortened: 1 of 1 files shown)


gonidelis:
I don't see value in CI being green here as the error pops with GCC16. @ reviewers, do check with gcc16 and cuda 13.2 if you wanna verify fix is effective

PR #8147, opened 2026-03-24
Implement the new tuning API for `DispatchBatchMemcpy`

bernhardmgruber:
- [x] No SASS checks on `cub.bench.copy.memcpy.base` for SM75

Fixes: #7636

### DIFF
--- a/cub/benchmarks/bench/copy/memcpy.cu
+++ b/cub/benchmarks/bench/copy/memcpy.cu
@@ -78,7 +78,7 @@ struct offset_to_size_t
 
 struct policy_selector_t
 {
-  _CCCL_HOST_DEVICE constexpr auto operator()(::cuda::arch_id /*arch*/) const
+  [[nodiscard]] _CCCL_HOST_DEVICE constexpr auto operator()(::cuda::arch_id /*arch*/) const
     -> cub::detail::batch_memcpy::batch_memcpy_policy
   {
     return {


cub/cub/device/dispatch/tuning/tuning_batch_memcpy.cuh
```diff
@@ -14,81 +14,149 @@
...
-               block_delay_constructor_t>;
-
-    using AgentLargeBufferPolicyT =
-      agent_large_buffer_policy<LARGE_BUFFER_BLOCK_THREADS, LARGE_BUFFER_BYTES_PER_THREAD>;
-  };
-
-  using MaxPolicy = Policy700;
+    const auto large = large_buffer_policy{
+      256,
+      32,
```
Jacobfaib, line 137: It might be nice to give some names to these magic numbers (and below as well). Alternatively, adding inline comments like `/*the_paramter_name=*/`.

cub/cub/device/dispatch/tuning/tuning_batch_memcpy.cuh
```diff
@@ -14,81 +14,149 @@
...
+  [[nodiscard]] _CCCL_API constexpr friend bool
+  operator!=(const large_buffer_policy& lhs, const large_buffer_policy& rhs)
+  {
+    return !(lhs == rhs);
+  }
 
-  using buff_delay_constructor_t  = detail::default_delay_constructor_t<BufferOffsetT>;
-  using block_delay_constructor_t = detail::default_delay_constructor_t<BlockOffsetT>;
+#if !_CCCL_COMPILER(NVRTC)
+  friend ::std::ostream& operator<<(::std::ostream& os, const large_buffer_policy& policy)
```
Jacobfaib, line 92: Should this be `_CCCL_HOST_API` (and above, line 61)?

cub/benchmarks/bench/copy/memcpy.cu
```diff
@@ -75,43 +75,34 @@ struct offset_to_size_t
...
-  delay_constructor_t<TUNE_BUFF_MAGIC_NS, TUNE_BUFF_L2_WRITE_LATENCY_NS, TUNE_BUFF_DELAY_CONSTRUCTOR_ID>;
-using block_delay_constructor_t =
-  delay_constructor_t<TUNE_BLOCK_MAGIC_NS, TUNE_BLOCK_L2_WRITE_LATENCY_NS, TUNE_BLOCK_DELAY_CONSTRUCTOR_ID>;
-
-struct policy_hub_t
+
+struct policy_selector_t
 {
-  struct policy_t : cub::ChainedPolicy<500, policy_t, policy_t>
+  _CCCL_HOST_DEVICE constexpr auto operator()(::cuda::arch_id /*arch*/) const
```
Jacobfaib:
```suggestion
  [[nodiscard]] _CCCL_HOST_DEVICE constexpr auto operator()(::cuda::arch_id /*arch*/) const
```

cub/cub/device/dispatch/tuning/tuning_batch_memcpy.cuh
bernhardmgruber: Yeah, some IDEs offer nice help here, but I don't mind adding comments.

cub/cub/device/dispatch/tuning/tuning_batch_memcpy.cuh
bernhardmgruber: We don't have a policy when to apply that macro. I generally avoid it to decrease noise.

cub/test/catch2_test_device_batch_memcpy_custom_policy_hub.cu
bernhardmgruber, line 1: We can drop this test, since `DispatchBatchMemcpy` was not actually part of the public API, so we don't need to test passing a custom policy hub.

PR #8152, opened 2026-03-24
Fix temp storage for empty problem in merge sort

bernhardmgruber:


cub/cub/device/dispatch/dispatch_merge_sort.cuh
NaderAlAwar: Question: any reason not to fix the old path as well? Is it considered a breaking change?

bernhardmgruber: It's dead code, but fine. I fixed it there as well.

PR #8155, opened 2026-03-24
Ensure all execution policies are also environments

miscco:
We want to be able to pass additional information to execution policies.

We already settled on the sender & receiver approach with queries and key value pairs.

This refactors the current approach to turn `__execution_policy_base` into a proper environment. That has the advantage that we do not need to handle the cuda execution policy in a special way.

The only cuda specific code is when we want to ensure that streams are passed as `cuda::stream_ref` and memory resources as `cuda::mr::resource_ref`

This is currently failing some tests we had in place that relied on special handling of the `cuda` policy wrt to memory resources

I am not sure if that is something we want to change?

### DIFF
--- a/libcudacxx/include/cuda/std/__fwd/execution_policy.h
+++ b/libcudacxx/include/cuda/std/__fwd/execution_policy.h
@@ -48,13 +48,13 @@ enum __execution_backend : uint8_t
   // The backends we provide
   __none = 0,
 #if _CCCL_HAS_BACKEND_CUDA()
-  __cuda = 1 << 1,
+  __cuda = 1 << 0,
 #endif // _CCCL_HAS_BACKEND_CUDA()
 #if _CCCL_HAS_BACKEND_OMP()
-  __omp = 1 << 2,
+  __omp = 1 << 1,
 #endif // _CCCL_HAS_BACKEND_OMP()
 #if _CCCL_HAS_BACKEND_TBB()
-  __tbb = 1 << 3,
+  __tbb = 1 << 2,
 #endif // _CCCL_HAS_BACKEND_TBB()
 };
 


libcudacxx/include/cuda/std/__execution/policy.h
```diff
@@ -68,6 +75,58 @@ struct __execution_policy_base
...
+      return __execution_policy_base<_Policy, _Env, _Envs...>{
+        ::cuda::std::forward<_Env>(__env), ::cuda::std::__get<_Indices>(this->__envs_)...};
+    }
+  }
+
+  //! @brief Prepend an environment to the current ones
+  template <class _Env>
+  [[nodiscard]] _CCCL_API constexpr auto with(_Env&& __env) const
+  {
+    return __with(::cuda::std::forward<_Env>(__env), ::cuda::std::make_index_sequence<sizeof...(_Envs)>());
```
miscco: We want to expand this to also check for streams and resource_ref / any_resource

libcudacxx/include/cuda/std/__execution/policy.h
```diff
@@ -68,6 +75,58 @@ struct __execution_policy_base
...
+    { // We want to force the use of ::cuda::stream_ref
+      ::cuda::stream_ref __stream{__value};
+      return __with(prop{__tag, __stream}, ::cuda::std::make_index_sequence<sizeof...(_Envs)>());
+    }
+    else
+#endif // _CCCL_HAS_BACKEND_CUDA()
+      if constexpr (is_same_v<_Tag, ::cuda::mr::get_memory_resource_t>)
+      {
+        // TODO(miscco): If we support more than one backend, we need to change this to satisfy the backends needs
+        ::cuda::mr::resource_ref<::cuda::mr::device_accessible> __resource{__value};
```
miscco: We want to expand to be able to own the resource by switching on the reference type

rvalue_references -> any_resource
lvalue references -> resource_ref

miscco: Done

libcudacxx/test/libcudacxx/cuda/execution/execution_policy/get_stream.pass.cpp
```diff
@@ -34,23 +34,21 @@ void test(Policy pol)
...
     assert(cuda::__call_or(::cuda::get_stream, default_stream, pol_with_stream) == stream);
 
     using stream_policy_t = decltype(pol_with_stream);
     cuda::stream other_stream{cuda::device_ref{0}};
-    decltype(auto) pol_with_other_stream = pol_with_stream.with_stream(other_stream);
-    static_assert(cuda::std::is_same_v<decltype(pol_with_other_stream), stream_policy_t>);
+    decltype(auto) pol_with_other_stream = pol_with_stream.with(cuda::get_stream, other_stream);
 
     // The original stream remains unchanged
     assert(cuda::__call_or(::cuda::get_stream, default_stream, pol_with_stream) == stream);
```
miscco: We can shorten this to just `cuda::get_stream(pol_with_stream)`

libcudacxx/include/cuda/std/__execution/policy.h
```diff
@@ -68,6 +69,111 @@ struct __execution_policy_base
...
+    }
+    else if constexpr (::cuda::mr::resource<_Env>)
+    { // TODO(miscco): If we support more than one backend, we need to change this to satisfy the backends needs
+      static_assert(::cuda::mr::resource_with<_Env, ::cuda::mr::device_accessible>,
+                    "Memory resources need to provide device accessible memory");
+      static_assert(!is_const_v<_Env>, "A memory resource must be passed by non-const reference");
+      if constexpr (!is_lvalue_reference_v<_Env> && !::cuda::mr::__is_resource_ref<remove_cvref_t<_Env>>)
+      { // The user passed a prvalue, which indicates we should own the resource
+        return __with(prop{::cuda::mr::get_memory_resource,
+                           ::cuda::mr::any_resource<::cuda::mr::device_accessible> {
```
pciolkosz: If we template the policy anyway could we just keep the original resource type?

libcudacxx/include/cuda/std/__execution/policy.h
```diff
@@ -68,6 +69,111 @@ struct __execution_policy_base
...
+    }
+    else if constexpr (is_same_v<remove_cvref_t<_Tag>, ::cuda::mr::get_memory_resource_t>)
+    { // TODO(miscco): If we support more than one backend, we need to change this to satisfy the backends needs
+      static_assert(::cuda::mr::resource_with<_Value, ::cuda::mr::device_accessible>,
+                    "Memory resources need to provide device accessible memory");
+      static_assert(!is_const_v<_Value>, "A memory resource must be passed by non-const reference");
+      if constexpr (!is_lvalue_reference_v<_Value> && !::cuda::mr::__is_resource_ref<remove_cvref_t<_Value>>)
+      { // The user passed a prvalue, which indicates we should own the resource
+        return __with(prop{__tag,
+                           ::cuda::mr::any_resource<::cuda::mr::device_accessible> {
```
pciolkosz: Hmm, is it guaranteed to always be device_accessible? I know in general we wanted to support type erased wrappers with no properties, but not sure if it applies here too

pciolkosz (approved):

libcudacxx/test/libcudacxx/std/utilities/expol/policies.compile.pass.cpp
```diff
@@ -24,33 +24,6 @@
...
-  assert(!(execution::unseq == execution::par_unseq));
-
-  assert(execution::seq != execution::unseq);
-  assert(execution::par != execution::seq);
-  assert(execution::par_unseq != execution::par);
-  assert(execution::unseq != execution::par_unseq);
-
-  return true;
-}
-
```
bernhardmgruber, line 53: Q: Why do we need to remove these tests?

libcudacxx/include/cuda/std/__execution/policy.h
miscco: For now we only support the CUDA backend, so I would love to require that outright. 

That said, we might need to go a different route given that rapids might need resources without properties @bdice Is that correct?

libcudacxx/test/libcudacxx/std/utilities/expol/policies.compile.pass.cpp
miscco: With environments, this would require all properties attached to `with` to be equality comparable, which seems to be a huge requirement for no gain.

Thing about every tuning policy needing comparison operators

libcudacxx/test/libcudacxx/std/utilities/expol/environments.pass.cpp
```diff
@@ -0,0 +1,278 @@
...
+    cuda::stream stream{cuda::device_ref{0}};
+    static_assert(!cuda::std::__is_callable_v<cuda::get_stream_t, const Policy&>);
+    const auto new_policy = policy.with(cuda::get_stream, ::cuda::stream_ref{stream});
+    static_assert(cuda::std::__is_callable_v<cuda::get_stream_t, decltype(new_policy)>);
+    auto&& result = cuda::get_stream(new_policy);
+    static_assert(cuda::std::is_same_v<decltype(result), cuda::stream_ref&&>);
+    assert(stream == result);
+  }
+
+  { // the policy can take a cuStream_t
```
bernhardmgruber:
```suggestion
  { // the policy can take a cudaStream_t
```

libcudacxx/test/libcudacxx/std/utilities/expol/environments.pass.cpp
```diff
@@ -0,0 +1,278 @@
...
+    cuda::stream stream{cuda::device_ref{0}};
+    static_assert(!cuda::std::__is_callable_v<cuda::get_stream_t, const Policy&>);
+    const auto new_policy = policy.with(stream);
+    static_assert(cuda::std::__is_callable_v<cuda::get_stream_t, decltype(new_policy)>);
+    auto&& result = cuda::get_stream(new_policy);
+    static_assert(cuda::std::is_same_v<decltype(result), cuda::stream_ref&&>);
+    assert(stream == result);
+  }
+
+  { // the policy can take a cuStream_t as an environment
```
bernhardmgruber:
```suggestion
  { // the policy can take a cudaStream_t as an environment
```

libcudacxx/test/libcudacxx/std/utilities/expol/environments.pass.cpp
```diff
@@ -0,0 +1,278 @@
...
+    assert(result.value == 42);
+  }
+}
+
+bool test()
+{
+  test(cuda::std::execution::seq);
+  test(cuda::std::execution::par);
+  test(cuda::std::execution::par_unseq);
+  test(cuda::std::execution::unseq);
```
bernhardmgruber, line 265: Q: Why should the standard policies accept streams and memory resources?

libcudacxx/test/libcudacxx/std/utilities/expol/policies.compile.pass.cpp
bernhardmgruber: But the standard execution policies are not environments. Only `cuda::execution::gpu` is one.

miscco: That is true, but then we would have to make an exception for empty policies, which seems strange.

Also there is no requirement that the execution policies are comparable

https://en.cppreference.com/w/cpp/algorithm/execution_policy_tag_t.html

libcudacxx/test/libcudacxx/std/utilities/expol/environments.pass.cpp
miscco: Memory resources are universally useful also for potential host execution policies

regarding streams, its a bit more questionable, but I believe if we add a way to add a backend to an execution policy it will make sense

bernhardmgruber: I would really not allow this for any of the standard execution policies, even if we saw a use for this in the future. Users are going to do unimaginable things if they have a chance to ;)

libcudacxx/test/libcudacxx/std/utilities/expol/policies.compile.pass.cpp
bernhardmgruber: > Also there is no requirement that the execution policies are comparable

That's a fair point. Fine for me then!

libcudacxx/include/cuda/std/__fwd/execution_policy.h
davebayer, line 58:
```suggestion
#if _CCCL_HAS_BACKEND_CUDA()
__cuda = 1 << 0,
#endif // _CCCL_HAS_BACKEND_CUDA()
#if _CCCL_HAS_BACKEND_OMP()
__omp = 1 << 1,
#endif // _CCCL_HAS_BACKEND_OMP()
#if _CCCL_HAS_BACKEND_TBB()
__tbb = 1 << 2,
#endif // _CCCL_HAS_BACKEND_TBB()
```

libcudacxx/include/cuda/std/__pstl/cuda/temporary_storage.h
```diff
@@ -114,14 +116,27 @@ class __temporary_storage
...
   {
-    if constexpr (::cuda::std::__is_callable_v<::cuda::mr::get_memory_resource_t, const _Policy&>)
+    if constexpr (__is_callable_v<::cuda::mr::get_memory_resource_t, const _Policy&>)
     {
-      return ::cuda::mr::get_memory_resource(__policy);
+      const auto& __resource = ::cuda::mr::get_memory_resource(__policy);
+      using __resource_t     = remove_cvref_t<decltype(__resource)>;
+
+#  if !_CCCL_COMPILER(MSVC)
+      if constexpr (!::cuda::mr::resource_with<::cuda::mr::device_accessible>)
```
davebayer: Please, document why this is disallowed with msvc

miscco: This is just me fishing why it fails with MSVC

libcudacxx/include/cuda/std/__execution/policy.h
miscco: Dropped requirement to always have device_accessible memory resoures

libcudacxx/include/cuda/std/__execution/policy.h
miscco: nah we want to use resource_ref, that also makes it much simpler in the backend

Issue #8028, opened 2026-03-13
Allow to control SMEM check in warpspeed scan


bernhardmgruber:
The warpspeed scan implemenation checks whether it can fit the current workload (for the current tuning) into 48KiB SMEM, and if it can't, will fall back to the old scan implementation.

This is undesirable during tuning, because a random tuning may exceed the 48KiB limit and then benchmark the old scan implementation. We should rather fail to compile in this case and skip the tuning values.

A user providing a custom tuning to scan, may want to opt into ignoring the 48KiB SMEM limit, since they are willing to produce a binary that is not compatible with future CUDA architectures for the sake of higher performance. We are already aware of tunings that would yield better performance than the current CUB tuning, but we cannot implement them since a single stage would already exceed 48 KiB SMEM.

We should thus allow to disable the SMEM check or to enforce it without fallback.

PR #8145, opened 2026-03-24
Refactor warpspeed scan tuning

bernhardmgruber:
These are some refactorings following the restructuring to support CCCL.C and the new tuning API: #7565 

- [x] No SASS changes for `cub.bench.scan.exclusive.sum.base` on SM`75;80;86;90;100;120`

PR #8158, opened 2026-03-24
Let scan tuning policy choose warpspeed or not

bernhardmgruber:
The motivation for this change is that now users can pick the underlying implementation and enforce that choice, which leads to compilation errors if the chosen implementation is not viable (e.g. exceeds 48KiB SMEM, or exceeds registers, or the PTX ISO is too low).

This is also what we need for tuning. We want to enforce a certain scan implementation and fail to compile if the tuning framework selects parameters that are not viable.

- [x] Merge before: #8145
- [x] No SASS changes for `cub.bench.scan.exclusive.sum.base` on SM`75;80;86;90;100;120`

Fixes: #8028

### DIFF
--- a/cub/cub/device/dispatch/tuning/tuning_scan.cuh
+++ b/cub/cub/device/dispatch/tuning/tuning_scan.cuh
@@ -37,6 +37,7 @@
 #include <cuda/std/__type_traits/enable_if.h>
 #include <cuda/std/__type_traits/is_trivially_copy_constructible.h>
 #include <cuda/std/__type_traits/void_t.h>
+#include <cuda/std/optional>
 
 #if !_CCCL_COMPILER(NVRTC)
 #  include <ostream>
@@ -872,7 +873,8 @@ struct policy_selector
   // TODO(griwes): remove this field before policy_selector is publicly exposed
   bool benchmark_match;
 
-  _CCCL_API constexpr scan_warpspeed_policy get_warpspeed_policy(::cuda::arch_id arch) const
+  _CCCL_API constexpr auto get_warpspeed_policy(::cuda::arch_id arch) const
+    -> ::cuda::std::optional<scan_warpspeed_policy>
   {
     if (arch >= ::cuda::arch_id::sm_100)
     {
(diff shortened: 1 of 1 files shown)


cub/cub/device/dispatch/kernels/kernel_scan.cuh
```diff
@@ -180,44 +184,44 @@ template <typename PolicySelector,
...
+  _CCCL_GRID_CONSTANT const InputIteratorT d_in,
+  _CCCL_GRID_CONSTANT const OutputIteratorT d_out,
+  tile_state_kernel_arg_t<ScanTileState, AccumT> tile_state,
+  _CCCL_GRID_CONSTANT const int start_tile,
+  ScanOpT scan_op,
 // nvcc 12.0 gets stuck compiling some TUs like `cub.bench.scan.exclusive.sum.base`, so only enable for newer versions
 #if _CCCL_CUDACC_AT_LEAST(12, 8)
-    _CCCL_GRID_CONSTANT
+  _CCCL_GRID_CONSTANT
 #endif // _CCCL_CUDACC_AT_LEAST(12, 8)
```
miscco, line 196: Are we sure this is still needed, I thought that _CCCL_GRID_CONSTANT is only used above 12.8

bernhardmgruber:
I see SASS diffs for SM100 and SM120. Almost exclusively in the setup code around the SMEM allocator. More instructions are generated.

cub/cub/device/dispatch/kernels/kernel_scan.cuh
bernhardmgruber: The `DeviceScanKernel` entry point is also used for the old scan implementation, so yes, we still need that.

cub/cub/device/dispatch/kernels/kernel_scan_warpspeed.cuh
```diff
@@ -110,8 +110,9 @@ allocResources(warpspeed::SyncHandler& syncHandler, warpspeed::SmemAllocator& sm
     warpspeed::SmemResource<SumThreadAndWarpT>(syncHandler, smemAllocator, warpspeed::Stages{numStages}),
   };
 
+  constexpr auto policy = get_warpspeed_policy<PolicySelector>();
   setup_scan_resources(
-    get_warpspeed_policy<PolicySelector>(),
+    policy,
```
bernhardmgruber, line 115: This fixed the SASS regression. The tuning policy was computed at **runtime**. We should be careful in the future to always put a tuning policy into a `constexpr `variable before accessing it.

cub/cub/device/dispatch/kernels/kernel_scan.cuh
miscco: Oh I meant whether we need the conditional compilation

miscco (approved):

bernhardmgruber:
So, something here breaks CCCL.C.

miscco:
> So, something here breaks CCCL.C.

That looks like an invalid value passed

```
/home/coder/cccl/c/parallel/test/algorithm_execution.h:154: FAILED:
    {Unknown expression after the reported line}
  due to a fatal error condition:
    SIGFPE - Floating point error signal
 
```

bernhardmgruber:
I can't reproduce it locally. Seems it's in the old scan implementation.

bernhardmgruber:
Found it. I selected the warpspeed tuning even on older architectures, which led to a division by zero in the host code. Fixed.

PR #8159, opened 2026-03-24
[CUB] Replace `Shuffle(Up|Down|Index)` with `cuda::device::warp_shuffle`

fbusato:
## Description

Address the aliasing issue in:

-  #8117

Additionally, deprecate legacy shuffle functions in favor of new CUDA device APIs.
Notes: `SHFL` instruction counts are preserved, reduce benchmark kernels show significantly fewer instructions (up to 20%)

### DIFF
--- a/ci/bench.yaml
+++ b/ci/bench.yaml
@@ -29,16 +29,30 @@ benchmarks:
     # Examples:
     # - '^cub\.bench\.for_each\.base'
     # - '^cub\.bench\.reduce\.(sum|min)\.'
+    - '^cub\.bench\.reduce\.'
+    - '^cub\.bench\.scan\.'
+    - '^cub\.bench\.radix_sort\.'
+    - '^cub\.bench\.merge_sort\.'
+    - '^cub\.bench\.select\.'
+    - '^cub\.bench\.partition\.'
+    - '^cub\.bench\.run_length_encode\.'
+    - '^cub\.bench\.segmented_reduce\.'
+    - '^cub\.bench\.segmented_sort\.'
+    - '^cub\.bench\.segmented_radix_sort\.'
+    - '^cub\.bench\.topk\.'
+    - '^cub\.bench\.segmented_topk\.'
+    - '^cub\.bench\.transform_reduce\.'
+    - '^cub\.bench\.copy\.'
 
   # Select GPUs. These are limited and shared, be intentional and conservative.
   gpus:
     # - "t4"         # sm_75, 16 GB
     # - "rtx2080"    # sm_75,  8 GB
-    # - "rtxa6000"   # sm_86, 48 GB
+      - "rtxa6000"   # sm_86, 48 GB
     # - "l4"         # sm_89, 24 GB
     # - "rtx4090"    # sm_89, 24 GB
-    # - "h100"       # sm_90, 80 GB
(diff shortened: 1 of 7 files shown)


miscco (approved):

bernhardmgruber:
That's a very impactful change. I see it affects CUB's radix sort, reduce and scan. I am afraid we need to see benchmarks for all those algorithms on at least one architecture.

Q: Are the new `warp_shuffle_idx` functions implement the same on all CUDA architectures? If not, we may want to get an additional benchmark of one of the algorithms on an architecture where the shuffle implementation is different.

bernhardmgruber:
> `SHFL` instruction counts are preserved, reduce benchmark kernels show significantly fewer instructions (up to 20%)

Unfortunately, instruction count does not always translate to runtime improvement. For a change to our algorithms, there either needs to be no changes in SASS or a benchmark that shows the impact. I know it's annoying, but letting a regression slip in would be even worse.

We could try the new benchmark CI to do this ;)

fbusato:
@bernhardmgruber I ran the following set of benchmark set on SM86, SM100, SM120.
```
- '^cub\.bench\.reduce\.(sum|min|nondeterministic|deterministic|by_key)\.'
- '^cub\.bench\.scan\.exclusive\.(sum|by_key)\.'
- '^cub\.bench\.radix_sort\.keys\.'
- '^cub\.bench\.merge_sort\.keys\.'
- '^cub\.bench\.select\.if\.'
- '^cub\.bench\.partition\.if\.'
- '^cub\.bench\.run_length_encode\.'
- '^cub\.bench\.segmented_reduce\.sum\.'
- '^cub\.bench\.segmented_sort\.'
- '^cub\.bench\.segmented_radix_sort\.'
- '^cub\.bench\.topk\.keys\.'
- '^cub\.bench\.segmented_topk\.'
- '^cub\.bench\.transform_reduce\.'
- '^cub\.bench\.copy\.'
 ```
Better performance in most cases, up to 30% is the extreme ones. There are a few isolated regressions > 5%, but the number of improvements far exceed them.

bernhardmgruber:
The benchmarks are bit hard to read since there are so many of them. Also, the benchmarks were aborted after 6h so a lot of them are missing. This makes the analysis hard and incomplete.

`cub.bench.scan.exclusive.sum.base` looks good on H100. It's irrelevant on RTX PRO 6000 (uses warpspeed scan), and the benchmark on RTX A6000 did not finish, so no results.

The `cub.bench.scan.exclusive.by_key.base` result are mixed and have regressions larger than 2% for runs with >=2^24 elements. That's concerning. Some runs show up to 5% improvements. RTX PRO 6000 results are similar. A6000 again no results.

While it looks like the changes to `cub/cub/warp/specializations/warp_scan_shfl.cuh` could be somewhat fine and may look better with a different tuning, I can't judge this from the data available :S

Given this PR touches several other algorithms as well, can we split and scope down this changeset in a way that makes reviewing of the performance impact more manageable? We could split the changes to scan, reduce and radix sort for example.

The refactoring to `libcudacxx/include/cuda/__warp/warp_shuffle.h` looks like it would not change SASS. Can we send this in an extra PR and proof with a SASS diff, that no SASS changes?

PR #8160, opened 2026-03-24
Pin cuda-toolkit wheel to container's CTK major.minor in CI

### LINKED ISSUE #9377: Pin cuda-toolkit PyPI wheels in CI to match the container's exact CTK major.minor version

leofang:
## Summary

closes #9377

- Pin `cuda-toolkit` PyPI wheels in CI to match the container's exact CTK major.minor version (e.g. `cuda-toolkit==12.9.*`) instead of floating to the latest (e.g. `cuda-toolkit==12.*`)
- Uses pip's native `PIP_CONSTRAINT` mechanism — no changes to `pyproject.toml`, so end users are unaffected
- Applies to all 4 Python test scripts: `test_cuda_cccl_headers`, `test_cuda_compute`, `test_cuda_coop`, `test_cuda_cccl_examples`

## Motivation

The Python CI test scripts previously extracted only the **major** version from `nvcc` (e.g. `12`) and installed `cuda-toolkit==12.*`, which always resolved to the latest 12.x from PyPI regardless of the container's actual CTK version. This meant that even when running in a CTK 12.0 container, CI would install `cuda-toolkit` 12.9 wheels (as discovered in https://github.com/NVIDIA/cccl/pull/8139#discussion).

This masked issues like the nvrtc compiler bug in CTK 12.4, because the pip-installed nvrtc (latest) was always used instead of the container's version.

## How it works

Each test script now also extracts the `X.Y` version from `nvcc --version` and writes a pip constraint file:
```bash
cuda_version=$(nvcc --version | grep release | awk '{print $6}' | tr -d ',' | cut -d '.' -f 1-2 | cut -d 'V' -f 2)
export PIP_CONSTRAINT="${TMPDIR:-/tmp}/ctk-constraint.txt"
echo "cuda-toolkit==${cuda_version}.*" > "$PIP_CONSTRAINT"
```

`PIP_CONSTRAINT` is a standard pip environment variable that automatically constrains all subsequent `pip install` commands in the script.

## Test plan

- [ ] Verify Python CI jobs pass with this change (the constraint should be a no-op for jobs where the container CTK matches the latest wheels)
- [ ] Ideally test with a lower-bound CTK container (e.g. 12.0) to confirm that `cuda-toolkit==12.0.*` is installed instead of 12.9


ci/test_cuda_cccl_examples_python.sh
```diff
@@ -9,6 +9,10 @@ source "$ci_dir/pyenv_helper.sh"
 source "$ci_dir/util/python/common_arg_parser.sh"
 parse_python_args "$@"
 cuda_major_version=$(nvcc --version | grep release | awk '{print $6}' | tr -d ',' | cut -d '.' -f 1 | cut -d 'V' -f 2)
+# Pin cuda-toolkit wheels to match the container's CTK major.minor version
+cuda_version=$(nvcc --version | grep release | awk '{print $6}' | tr -d ',' | cut -d '.' -f 1-2 | cut -d 'V' -f 2)
```
leofang: nit: please consolidate these awk handling. Seems repetitive to me.

leofang:
/ok to test [7f3f7a1](https://github.com/NVIDIA/cccl/pull/8160/commits/7f3f7a131e86ba6f436d5b3201e31b84901fc3ab)

ci/test_cuda_cccl_examples_python.sh
```diff
@@ -8,7 +8,11 @@ source "$ci_dir/pyenv_helper.sh"
 # Parse common arguments
 source "$ci_dir/util/python/common_arg_parser.sh"
 parse_python_args "$@"
-cuda_major_version=$(nvcc --version | grep release | awk '{print $6}' | tr -d ',' | cut -d '.' -f 1 | cut -d 'V' -f 2)
+# Extract CTK major.minor version from nvcc and pin cuda-toolkit wheels to match
+cuda_version=$(nvcc --version | grep release | awk '{print $6}' | tr -d ',' | cut -d '.' -f 1-2 | cut -d 'V' -f 2)
+cuda_major_version=$(echo "$cuda_version" | cut -d '.' -f 1)
+export PIP_CONSTRAINT="${TMPDIR:-/tmp}/ctk-constraint.txt"
+echo "cuda-toolkit==${cuda_version}.*" > "$PIP_CONSTRAINT"
```
NaderAlAwar: Question: I may be missing something, but how does the constraint file get propagated to the pip install command?

shwina:
> Pin cuda-toolkit PyPI wheels in CI to match the container's exact CTK major.minor version (e.g. cuda-toolkit==12.9.*) instead of floating to the latest (e.g. cuda-toolkit==12.*)

I understand that this fixes things in our CI, but can you clarify whether this is a _temporary_ workaround? Longer term, I would prefer if no one is responsible for specifying a CUDA minor version; it's an additional burden that we essentially pass on to users.

It would be helpful if we can answer the following questions:

1. Why do we need to specify the minor version - shouldn't "minor version compatibility" ensure that even if I install CTK minor version 12.9, I can run with any 12.X driver?

2. Why does the system CTK affect us at all? We use pathfinder to load all the libraries required by CCCL; my understanding is that pathfinder will [prefer libraries in the pip env](https://nvidia.github.io/cuda-python/cuda-pathfinder/latest/generated/cuda.pathfinder.load_nvidia_dynamic_lib.html) over the system CTK.

NaderAlAwar:
> I understand that this fixes things in our CI, but can you clarify whether this is a temporary workaround? Longer term, I would prefer if no one is responsible for specifying a CUDA minor version; it's an additional burden that we essentially pass on to users.

> Why do we need to specify the minor version - shouldn't "minor version compatibility" ensure that even if I install CTK minor version 12.9, I can run with any 12.X driver?

I don't think this is a temporary workaround. My understanding is that this PR ensures that if we test with a 12.0 container, we use CTK 12.0. This would have helped us catch issues like https://github.com/NVIDIA/cccl/issues/8138. This will only help minor version compatibility, because we ensure that a larger set of minor versions work

shwina:
> if we test with a 12.0 container, we use CTK 12.0.

This is the problematic part. I think we should be able to use any 12.x on a 12.y container (or, my understanding of MVC is wrong, and I would love to understand what the gap is)

leofang:
For some reason I missed the notification. I think MVC does not work like that -- All CTK components must come from the same major.minor. MVC concerns CTK's version with respect to the UMD version. Without pinning all CTK components at major.minor, we have a potential mix-n-match (between CTK from wheels vs from container) which is not a supported use case. 

Either go with wheel pinning (to ensure wheel and container have the same major.minor), or follow CUDA Python and not use any CTK container at all (wheel alone decides the major.ninor).

leofang:
(The only outlier is nvJitLink, which is allowed to be higher than all other CTK's major.minor. I assumed this is well-known.)

shwina:
>  Without pinning all CTK components at major.minor, we have a potential mix-n-match (between CTK from wheels vs from container

If such a "mix-n-match" occurs, that is a bug somewhere in the CUDA Python ecosystem, correct? 

* CUDA Python libs should list, as dependencies, all CTK components they use 
* CUDA Python libs should use `pathfinder` to find the CTK components they use

Together, the above should ensure that system CTK in no way affects CUDA Python libs at runtime. 

--

My question is still:

> Why does the system CTK affect us at all?

NaderAlAwar:
@shwina in pure pip mode the system CTK does not affect us.

However, CuPy's default installation does not include a pip installed CTK. `pip install cupy-cuda12x` assumes a system CTK. CuPy also supports most minor versions, so 12.0+ and 13.0+. This means that there is a gap in our testing. I discussed this with @leofang and we agreed that cuda.compute can test the floor (i.e., 12.0 and 13.0) while CuPy can additionally test intermediate versions.

Without this PR, we are not testing 12.0 or 13.0. Instead, we install the latest CTK through pip. By pinning the minor version we guarantee that we test the floor.

The one missing thing is using the `sysctk` extra in our tests (this is for a future PR), but for now this at least tests the proper CTK minor version.

ci/windows/build_common_python.psm1
```diff
@@ -73,6 +73,26 @@ function Get-CudaMajor {
...
+        if (Test-Path $nvcc) {
+            $out = & $nvcc --version 2>&1
+            $text = ($out -join "`n")
+            if ($text -match 'release\s+(\d+\.\d+)') { return $Matches[1] }
+        }
+        # Fallback: parse major.minor from CUDA_PATH like ...\v13.0
+        $pathMatch = [regex]::Match($env:CUDA_PATH, 'v?(\d+\.\d+)')
+        if ($pathMatch.Success) { return $pathMatch.Groups[1].Value }
+    }
+    return '13.0'
```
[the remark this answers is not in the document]
NaderAlAwar: the default intentionally mirrors the existing Get-CudaMajor and CI containers always have nvcc (so the fallback is never hit)

miscco (approved):

wmaxey:
Not to hold up on submitting this, but most code in here looks repeated, can this functionality just be moved into a common script?

wmaxey (approved):

PR #8161, opened 2026-03-25
Doc issues fail CI

### LINKED ISSUE #7053: [BUG]: Doc issues do not fail CI
### Is this a duplicate?

### Type of Bug

Something else

### Component

Infrastructure

### Describe the bug

https://github.com/NVIDIA/cccl/pull/7052 fixed a number of bugs in CCCL docs that got through CI. We should make errors like these a CI failure. @vyasr suggested using `WARN_AS_ERROR=FAIL_ON_WARNINGS` in doxygen config and `-W` for `sphinx-build`. 

### How to Reproduce

Revert mentioned PR and see build logs

### Expected behavior

Bugs in docs do not get through CI 

### Reproduction link

_No response_

### Operating System

_No response_

### nvidia-smi output

_No response_

### NVCC version

_No response_

alliepiper:
## Summary

  Closes #7053. Makes the documentation build fail on both Doxygen and Sphinx
  warnings/errors, and fixes all existing issues so the build passes cleanly.

  ### Sphinx (`-W` warnings-as-errors)
  - Add `-W` to `SPHINXOPTS` in `gen_docs.bash`
  - Fix all undefined RST label references, missing toctree entries, title
    underline lengths, RST formatting errors, and duplicate labels
  - Add `suppress_warnings` in `conf.py` for `cpp.duplicate_declaration` and
    `docutils` — breathe architecture limitations that cannot be fixed in source
  - Skip generating API pages for symbols with unparseable C++ declarations
    (`_BREATHE_SKIP_SYMBOLS` in `auto_api_generator.py`)

  ### Doxygen (`WARN_AS_ERROR = FAIL_ON_WARNINGS`)
  - Unify warning settings across all 4 Doxyfiles (cub, libcudacxx, thrust, cudax)
  - Fix `gen_docs.bash` to check Doxygen exit codes from parallel builds
  - Fix `@param` name mismatches across STF base classes, cudax launch, and CUB
  - Add `_CCCL_DOXYGEN_INVOKED` stubs for CUB namespace macros
  - Remove orphaned `@endrst` directives in CUB device headers
  - Fix `@file` tags, stale `@param` entries, and misused `@name` tags
  - Hide recursive template inheritance from Doxygen (`graph_task`, `stream_task`,
    `task_dep`, `CompileTimeArray`)
  - Escape cross-project type references that Doxygen cannot resolve (each
    project's Doxygen build is currently isolated from the others)
  - Hide `_CCCL_DOXYGEN_INVOKED`-guard `::cuda::std::execution::env` defaults in
    `device_scan.cuh` template signatures
  - Exclude CUB `dispatch/` and `kernels/` directories (internal implementation)
  - Remove duplicate `tuple_element`/`tuple_size` docs from `thrust/pair.h`
    (authoritative docs remain in `tuple.h`)
  - Convert doc comments on excluded `detail` symbols to plain C++ comments
  - Restore `__hierarchy` to cudax Doxyfile INPUT (incorrectly removed)

  ### Verification

  Built docs from this branch and from the merge base with `gen_docs.bash` in
  a devcontainer, then diffed the HTML output. All differences are accounted for:

  - **29 CUB pages removed**: Internal dispatch/kernel classes and enums now
    excluded via `EXCLUDE_PATTERNS`. These were never public API.
  - **4 symbol pages removed**: `get_executor_func_t`, `group`,
    `has_property`, `property_with_value` — skipped due to unparseable C++
    declarations (breathe/Sphinx parser limitations on SFINAE expressions).
  - **2 thrust pages moved**: `tuple_element`/`tuple_size` moved from `pair`
    group to `tuple` group (removed duplicate, kept authoritative location).
  - **cudax hierarchy pages preserved**: `this_block`, `this_thread`,
    `this_warp`, `this_grid`, `this_cluster` all present after restoring
    `__hierarchy` to INPUT.
  - **No unexpected regressions** in generated documentation content.

  ## Test plan

  - [x] `docs/gen_docs.bash` passes with zero Doxygen errors and zero Sphinx warnings
  - [x] Diff of generated HTML against merge base shows no unexpected regressions

### DIFF
--- a/cub/cub/util_macro.cuh
+++ b/cub/cub/util_macro.cuh
@@ -27,7 +27,6 @@ CUB_NAMESPACE_BEGIN
 #  define CUB_DISABLE_KERNEL_VISIBILITY_WARNING_SUPPRESSION
 #endif
 
-
 /**
  * @def CUB_DISABLE_KERNEL_VISIBILITY_WARNING_SUPPRESSION
  * If defined, the default suppression of kernel visibility attribute warning is disabled.


alliepiper (commented):
/ok to test

cub/cub/util_namespace.cuh
```diff
@@ -31,6 +31,10 @@
 #  error CUB requires a definition of CUB_NS_QUALIFIER when CUB_NS_PREFIX/POSTFIX are defined.
 #endif
 
+#ifdef _CCCL_DOXYGEN_INVOKED
+#  define THRUST_CUB_WRAPPED_NAMESPACE
+#endif
```
Jacobfaib, line 36: I would put this (and any other such dummy defines) in the list of macros in `docs/cub/Doxyfile` to keep them all in one place.

cub/cub/util_vsmem.cuh
```diff
@@ -119,26 +119,20 @@ public:
...
   }
 
-  /**
-   * @brief Hints to discard modified cache lines of the used virtual shared memory.
-   * modified cache lines.
-   *
-   * @note Needs to be followed by `__syncthreads()` if the function returns true and the virtual shared memory is
-   * supposed to be reused after this function call.
-   */
+  // Hints to discard modified cache lines of the used virtual shared memory.
```
Jacobfaib: Why no longer doxygen-documented? Most of the other methods of this class are.

docs/_ext/auto_api_generator.py
```diff
@@ -1078,6 +1106,8 @@ def generate_namespace_api_page(project_name, items, title=None, doc_prefix=""):
         # Sort variables alphabetically
         items["variables"].sort(key=lambda x: x[0].lower())
         for name, refid in items["variables"]:
+            if name in _BREATHE_SKIP_SYMBOLS:
```
Jacobfaib, line 1116: Consider logging here as well (and above)?

docs/cub/Doxyfile
```diff
@@ -170,7 +170,9 @@ PREDEFINED             = __device__= \
 
 # Quiet mode
 QUIET                  = YES
-WARNINGS               = NO
+WARNINGS               = YES
+WARN_AS_ERROR          = FAIL_ON_WARNINGS
 WARN_IF_UNDOCUMENTED   = NO
 WARN_IF_DOC_ERROR      = YES
+WARN_IF_INCOMPLETE_DOC = NO
```
Jacobfaib, line 177: Why not this?

docs/gen_docs.bash
```diff
@@ -17,7 +17,8 @@ SCRIPT_PATH=$(cd $(dirname ${0}); pwd -P)
 cd $SCRIPT_PATH
 
 # Configuration
-SPHINXOPTS="${SPHINXOPTS:---keep-going}"
+# Keep going to surface all warnings; -W makes warnings fail the build.
+SPHINXOPTS="${SPHINXOPTS:---keep-going -W}"
```
Jacobfaib: Consider adding `--nitpicky` as well (warn about missing references, so `:func:does_not_exist()` or `\class does_not_exist` in doxygen raises warnings)

libcudacxx/include/cuda/__stream/stream_ref.h
```diff
@@ -105,8 +105,9 @@ class stream_ref
   //!
   //! @param __lhs The `stream_ref` to compare
   //! @return true if equal, false if unequal
-  [[nodiscard]] _CCCL_API friend bool operator==(const stream_ref& __lhs, const invalid_stream_t&) noexcept
+  [[nodiscard]] _CCCL_API friend bool operator==(const stream_ref& __lhs, const invalid_stream_t& __rhs) noexcept
   {
+    (void) __rhs;
```
Jacobfaib: I suppose we can let `clang-tidy` clean this up eventually, but should be `static_cast<void>(__rhs)`.

cub/cub/util_namespace.cuh
alliepiper: These are documented with @def, and need to be defined in the code to avoid warnings from doxygen when it tries to link the docs to the source location.

cub/cub/util_vsmem.cuh
alliepiper: Good catch -- this was to workaround some issues regarding undefined (to doxygen) symbols, but that ended up getting addressed in a different way and this can be reverted 👍

I'll put this back in the next push after this round of CI finishes.

docs/cub/Doxyfile
alliepiper: This warns when some, but not all, parameters have doxygen strings.

Like WARN_IF_UNDOCUMENTED, it's generally a nuisance imo -- we can add it later if there's a desire for it.

docs/gen_docs.bash
alliepiper: Sure, I'll give it shot and see what breaks 😄

libcudacxx/include/cuda/__stream/stream_ref.h
alliepiper: That'd be a discussion that's out of scope here. We use `(void) foo;` everywhere to silence unused param warnings by convention.

cudax/include/cuda/experimental/__stf/internal/data_interface.cuh
```diff
@@ -215,7 +216,7 @@ public:
    * @param instance_id The ID of the data instance
    * @return true if the instance was pinned, false otherwise
    */
-  virtual bool pin_host_memory(instance_id_t /*instance_id*/)
+  virtual bool pin_host_memory(instance_id_t instance_id)
```
caugonnet, line 219: this will issue an unused variable warning

cudax/include/cuda/experimental/__stf/internal/data_interface.cuh
```diff
@@ -310,11 +311,13 @@ public:
    *
    * @param ctx The backend context state
    * @param d The logical data_untyped
+   * @param tp The current task
    * @return The ID of the data instance for this logical data
    */
   template <typename backend_ctx_untyped>
-  instance_id_t get_default_instance_id(backend_ctx_untyped&, const logical_data_untyped& d, task& tp) const
+  instance_id_t get_default_instance_id(backend_ctx_untyped& ctx, const logical_data_untyped& d, task& tp) const
```
caugonnet, line 318: why is that better ?

cudax/include/cuda/experimental/__stf/internal/data_interface.cuh
bernhardmgruber, line 219:
```suggestion
  virtual bool pin_host_memory([[maybe_unused]] instance_id_t instance_id)
```
or `(void)instance_id` inside the function

cudax/include/cuda/experimental/__stf/internal/data_interface.cuh
bernhardmgruber: Because now there is a matching paramter to the `@param ctx` documentation comment

libcudacxx/include/cuda/__memory_resource/legacy_pinned_memory_resource.h
```diff
@@ -90,15 +90,17 @@ class legacy_pinned_memory_resource
 
   //! @brief Equality comparison with another \c legacy_pinned_memory_resource.
   //! @return Whether both \c legacy_pinned_memory_resource were constructed with the same flags.
-  [[nodiscard]] _CCCL_HOST_API constexpr bool operator==(legacy_pinned_memory_resource const&) const noexcept
+  [[nodiscard]] _CCCL_HOST_API constexpr bool operator==(legacy_pinned_memory_resource const& __other) const noexcept
   {
+    (void) __other;
```
bernhardmgruber: Q: Why do we need a parameter name here? There is no documentation for it.

cub/cub/device/device_scan.cuh
```diff
@@ -366,12 +366,18 @@ struct DeviceScan
...
   template <typename InputIteratorT,
             typename OutputIteratorT,
             typename NumItemsT,
-            typename EnvT = ::cuda::std::execution::env<>>
+            typename EnvT = // Doxygen cannot resolve ::cuda::std::execution::env
+#ifdef _CCCL_DOXYGEN_INVOKED
+            void
+#else
+            ::cuda::std::execution::env<>
+#endif
```
bernhardmgruber, line 379: This is very sad! Is there a way we can properly fix this? Environment overloads will show up everywhere!

docs/gen_docs.bash
alliepiper: ...maybe later.

● 11,057 warnings, dominated by:

  - ~4,500: Unresolved namespace/type references (::cuda, ::cuda::std, thrust, size_t, cudaError_t, cudaStream_t) — breathe emits C++ identifiers that
  Sphinx's cpp domain can't resolve because the types are defined in other Doxygen projects or CUDA headers
  - ~500: Unresolved detail namespace references
  - ~300: Undefined RST labels pointing to Doxygen-generated filenames (execution__policy_8h, device__vector_8h)
  - The rest: similar unresolvable cross-references

cub/cub/device/device_scan.cuh
alliepiper: There is a fix we can consider as part of the env rollout. I've filed a bug to track.

https://github.com/NVIDIA/cccl/issues/8176

alliepiper: tldr to save a click -- this fails because all of our projects use separate doxyfiles, and CUB's parser doesn't know about libcudacxx's symbols. Doxygen has some 'tagfile' mechanism to let the projects share cross-referencing info that would be great for a followup.

libcudacxx/include/cuda/__memory_resource/legacy_pinned_memory_resource.h
alliepiper: I ended up disabling claude subagents by the end of this branch because they kept doing unnecessary stuff like this. Apparently they don't have any real context from the session, don't read project config files, and generally like to ignore / not be aware of any custom/previous instructions.

So heads up if you start going down the subagent route. They're not great.

I'll double check whether this was another overeager robot and remove it if so.

alliepiper: It is indeed unnecessary, reverted.

cub/cub/util_namespace.cuh
alliepiper: Apologies to Nikolay for the accidental ping 😅

cub/cub/util_vsmem.cuh
alliepiper: Fixed.

docs/_ext/auto_api_generator.py
alliepiper: Fixed.

cudax/include/cuda/experimental/__stf/internal/data_interface.cuh
alliepiper: Fixed.

alliepiper: Fixed.

cub/cub/device/device_scan.cuh
miscco: Really? I thought that we explicitly defined the project dependency because of that

libcudacxx/include/cuda/__device/arch_traits.h
```diff
@@ -456,7 +456,7 @@ template <>
 
 //! @brief Gets the architecture traits for the given architecture id \c __id.
 //!
-//! @throws \c cuda::cuda_error if the \c __id is not a known architecture.
+//! @throws cuda::cuda_error if the \c __id is not a known architecture.
```
miscco, line 459: Could this be
```suggestion
//! @throws cuda::cuda_error if the @param __id is not a known architecture.
```

libcudacxx/include/cuda/__device/arch_traits.h
```diff
@@ -482,7 +482,7 @@ template <>
 
 //! @brief Gets the architecture traits for the given compute capability \c __cc.
 //!
-//! @throws \c cuda::cuda_error if the \c __cc doesn't have a corresponding architecture id.
+//! @throws cuda::cuda_error if the \c __cc doesn't have a corresponding architecture id.
```
miscco, line 485: Ditto:
```suggestion
//! @throws cuda::cuda_error if the @param __cc doesn't have a corresponding architecture id.
```

libcudacxx/include/cuda/__iterator/zip_function.h
```diff
@@ -109,7 +109,6 @@ class zip_function
 
 //! @brief Creates a @c zip_function from a function
 //! @tparam _Fn The functor to wrap
-//! @relates zip_iterator
```
miscco, line 112: This needs to stay, we want zip_function to show up with zip_iterator

cub/cub/device/device_scan.cuh
alliepiper: I couldn't find any cross-linking or dependencies between things at the doxygen level. Unless I missed a spot where we're patching this at a higher level, the tagfiles fix would be needed (AIUI, def not a doxy/sphinx expert).

libcudacxx/include/cuda/__device/arch_traits.h
alliepiper: I had claude run a quick test:

>   Tested in Doxygen 1.9.6. Using @param inside @throws causes:
  error: argument '__id' from the argument list of test_at_param has multiple @param documentation sections
  Doxygen interprets @param as starting a new parameter doc block, not an inline reference. Our current \c renders as monospace and is valid.

miscco: 😿

PR #8168, opened 2026-03-25
Add `explicit bool` operator for `cuda::device::lane_mask`

davebayer:
It can be really useful for example `assert(my_lane_mask & cuda::device::lane_mask::this_lane())`.

### DIFF
--- a/libcudacxx/include/cuda/__warp/lane_mask.h
+++ b/libcudacxx/include/cuda/__warp/lane_mask.h
@@ -63,6 +63,14 @@ public:
     return __value_;
   }
 
+  //! @brief Converts the lane mask to a bool.
+  //!
+  //! @return \c true if the mask is not zero, \c false otherwise.
+  _CCCL_DEVICE_API explicit constexpr operator bool() const noexcept
+  {
+    return __value_ != 0;
+  }
+
   //! @brief Returns a lane mask object with no lane bits set.
   //!
   //! @return A lane mask with no lane bits set.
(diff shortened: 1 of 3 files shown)


libcudacxx/include/cuda/__warp/lane_mask.h
```diff
@@ -63,6 +63,14 @@ class lane_mask
...
   }
 
+  //! @brief Converts the lane mask to a bool.
+  //!
+  //! @return \c true if the mask is not zero, \c false otherwise.
+  _CCCL_DEVICE_API explicit constexpr operator bool() const noexcept
+  {
+    return __value_ != 0;
+  }
+
```
miscco, line 73: I am not really convinced. A lane mask with no lanes set is still valid and this seems to suggest it is not

davebayer: It shouldn't mean that the mask is invalid. It should do the same thing as converting an `int` to `bool`. I will close the PR if you don't agree, we can just do `my_mask != lane_mask::none()`

miscco: I am totally fine to merge if there is support form others, just wanted to raise the question

bernhardmgruber:
I think we had the same discussion during standardizing `std::datapar::simd_mask`. I remember having seen functions like `if(all(mask))` and `if(any(mask))` etc., but I can't find them in the paper or wording at the moment. I believe having an `operator bool` was very controversial.

For lane mask, I would also argue against it.

PR #8169, opened 2026-03-25
Fix UB in `cuda::device::warp_match_all`

davebayer:


libcudacxx/include/cuda/__warp/warp_match_all.h
```diff
@@ -36,13 +36,15 @@ _CCCL_BEGIN_NAMESPACE_CUDA_DEVICE
...
 
-template <typename _Tp, typename _Up = ::cuda::std::remove_cv_t<_Tp>>
-[[nodiscard]] _CCCL_DEVICE_API bool warp_match_all(const _Tp& __data, lane_mask __lane_mask = lane_mask::all())
+template <class _Tp>
+[[nodiscard]] _CCCL_DEVICE_API bool warp_match_all(const _Tp& __data, lane_mask __lane_mask = lane_mask::all()) noexcept
 {
   _CCCL_ASSERT(__lane_mask != lane_mask::none(), "lane_mask must be non-zero");
-  constexpr int __ratio = ::cuda::ceil_div(sizeof(_Up), sizeof(uint32_t));
-  uint32_t __array[__ratio];
-  ::cuda::std::memcpy(__array, ::cuda::std::addressof(__data), sizeof(_Up));
```
davebayer, line 45: When `sizeof(_Tp)` is not a multiple of `4`, `__array` may contain some bytes that are not overwritten by `memcpy`. However, these bytes are compared in the loop, too.

miscco: The same applies for `warp_shuffle` please also fix

bernhardmgruber:
Can we do a quick SASS check whether there is no change for types where `sizeof(T)` is a multiple of 4?

davebayer:
> Can we do a quick SASS check whether there is no change for types where `sizeof(T)` is a multiple of 4?

@bernhardmgruber I haven't observed any significant changes, mostly the exact same code is generated. Just sometimes different registers are used. See https://godbolt.org/z/Yn36Yv99x

libcudacxx/include/cuda/__warp/warp_match_all.h
fbusato: no, warp shuffle instructions don't perform comparison

PR #8170, opened 2026-03-25
Refactor CCCL.C's histogram dispatch routines

bernhardmgruber:
* Move them from the dispatcher into a detail namespace
* Delete the unused/untested ones

No changes to functionality have been made


bernhardmgruber:
@NaderAlAwar, since you wrote #6915 initially, please have a look at this PR. I removed unused overloads since they are not tested either, so when refactoring everything to the new tuning API, I cannot test for breakages.

PR #8172, opened 2026-03-25
[CUB] Add detail::uninitialized_array

### LINKED ISSUE #7338: [FEA]: Provide a way to create an uninitialized array
### Is this a duplicate?

### Area

CUB

For vectorized code paths we often use something like `InputT input[items_per_thread];`

Because this is a plain C-array, the compiler will try to value-initialize the elements. However, this is almost never what we want because we want to fill it with data usually through a vectorized load.

We should provide a type that holds the array in a union, so that we can create it without spending time on initializing it.

Alternatively we could use `alignas(InputT) cuda::std::byte input[items_per_thread * sizeof(InputT)];` and then cast, but I am not sure what is cleaner

### Additional context

_No response_

charan-003:
Closes #7338

i assume uninitialized_array would look clean

it's reusable in many places in the transform kernel (and elsewhere in CUB) use this pattern, so using it avoids repeating the manual alignment + cast everytime


bernhardmgruber:
Thank you for the PR! I think we would want this new facility as an internal type in libcu++. It should also replace `thrust::cuda_cub::core::detail::uninitialized_array`. Can you please update the PR to reflect this? Thx!

miscco:
What is the use case for that? I have `libcudacxx/include/cuda/std/__pstl/cuda/temporary_storage.h`

otherwise we already have `cuda::__uninitialized_async_buffer` that does a lot of the things

cub/cub/detail/uninitialized_array.cuh
```diff
@@ -0,0 +1,54 @@
...
+CUB_NAMESPACE_BEGIN
+
+namespace detail
+{
+template <typename T, ::cuda::std::size_t N, ::cuda::std::size_t Alignment = alignof(T)>
+struct uninitialized_array
+{
+  using value_type           = T;
+  static constexpr auto size = ::cuda::std::integral_constant<::cuda::std::size_t, N>{};
+  alignas(Alignment) char data_[N * sizeof(T)];
```
miscco: This needs to be 
```suggestion
  alignas(Alignment) unsigned char data_[N * sizeof(T)];
```

Only `unsigned char` and `std::byte` are blessed by the standard

cub/cub/detail/uninitialized_array.cuh
```diff
@@ -0,0 +1,54 @@
...
+  _CCCL_HOST_DEVICE _CCCL_FORCEINLINE T* data()
+  {
+    return reinterpret_cast<T*>(data_);
+  }
+  _CCCL_HOST_DEVICE _CCCL_FORCEINLINE const T* data() const
+  {
+    return reinterpret_cast<const T*>(data_);
```
miscco: This is missing `cuda::std::launder` to tell the compiler we really know that there is an object in here

See `<cuda/std/inplace_vector>`

cub/cub/detail/uninitialized_array.cuh
```diff
@@ -0,0 +1,54 @@
...
+  }
+  _CCCL_HOST_DEVICE _CCCL_FORCEINLINE T& operator[](unsigned int idx)
+  {
+    return data()[idx];
```
miscco: This is missing a bounds check
```suggestion
    _CCCL_ASSERT(idx < N, "out of bounds access in uninitialized_array::operator[]");
    return data()[idx];
```

cub/cub/detail/uninitialized_array.cuh
```diff
@@ -0,0 +1,54 @@
...
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_CLANG)
+#  pragma clang system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_MSVC)
+#  pragma system_header
+#endif // no system header
+
+#include <cub/util_namespace.cuh>
+
+#include <cuda/std/cstddef>
+#include <cuda/std/type_traits>
```
miscco: Please only include the sub headers we need, its fine for `cstddef` but `type_traits` is too big

cub/cub/detail/uninitialized_array.cuh
```diff
@@ -0,0 +1,54 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION. All rights reserved.
+// SPDX-License-Identifier: BSD-3
+
+#pragma once
+
+#include <cub/config.cuh>
+
```
miscco: Please move this file into `libcudacxx/include/cuda/__memory/uninitialized_array`

Also please adopt the license and the config accoringly

cub/cub/detail/uninitialized_array.cuh
```diff
@@ -0,0 +1,54 @@
...
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_MSVC)
+#  pragma system_header
+#endif // no system header
+
+#include <cub/util_namespace.cuh>
+
+#include <cuda/std/cstddef>
+#include <cuda/std/type_traits>
+
+CUB_NAMESPACE_BEGIN
```
miscco: This needs to be
```suggestion
_CCCL_BEGIN_NAMESPACE_CUDA
```

cub/cub/detail/uninitialized_array.cuh
```diff
@@ -0,0 +1,54 @@
...
+
+namespace detail
+{
```
miscco: We can drop the detail namespace

cub/cub/detail/uninitialized_array.cuh
```diff
@@ -0,0 +1,54 @@
...
+template <typename T, ::cuda::std::size_t N, ::cuda::std::size_t Alignment = alignof(T)>
+struct uninitialized_array
```
miscco: Everything here should be `__ugly` or `_Ugly`
```suggestion
template <class _Tp, size_t _Size, size_t _Alignment = alignof(_Tp)>
struct __uninitialized_array
```

cub/cub/detail/uninitialized_array.cuh
miscco: Please structure this a bit differently so that we can easily see what is what
```suggestion
  alignas(_Alignment) unsigned char __data_[_Size * sizeof(_Tp)];
  
  using value_type           = _Tp;
```

cub/cub/detail/uninitialized_array.cuh
```diff
@@ -0,0 +1,54 @@
...
+{
+  using value_type           = T;
+  static constexpr auto size = ::cuda::std::integral_constant<::cuda::std::size_t, N>{};
```
miscco: What do we need size for?

cub/cub/detail/uninitialized_array.cuh
```diff
@@ -0,0 +1,54 @@
...
+  alignas(Alignment) char data_[N * sizeof(T)];
+  _CCCL_HOST_DEVICE _CCCL_FORCEINLINE T* data()
+  {
+    return reinterpret_cast<T*>(data_);
+  }
```
miscco: This should be reworded and also needs a `const overload`
```suggestion
  [[nodiscard]] _CCCL_API _Tp* data() noexcept
  {
    return ::cuda::std::launder(reinterpret_cast<_Tp*>(__data_));
  }
  
  [[nodiscard]] _CCCL_API const _Tp* data() const noexcept
  {
    return ::cuda::std::launder(reinterpret_cast<const _Tp*>(__data_));
  }
```

cub/cub/detail/uninitialized_array.cuh
```diff
@@ -0,0 +1,54 @@
...
+  {
+    return data()[idx];
+  }
+  _CCCL_HOST_DEVICE _CCCL_FORCEINLINE const T& operator[](unsigned int idx) const
+  {
+    return data()[idx];
+  }
+  _CCCL_HOST_DEVICE _CCCL_FORCEINLINE T (&as_array())[N]
+  {
+    return *reinterpret_cast<T(*)[N]>(data_);
```
miscco: I believe this is never used, we should drop it

cub/cub/detail/uninitialized_array.cuh
```diff
@@ -0,0 +1,54 @@
...
+    return reinterpret_cast<const T*>(data_);
+  }
+  _CCCL_HOST_DEVICE _CCCL_FORCEINLINE T& operator[](unsigned int idx)
+  {
+    return data()[idx];
+  }
+  _CCCL_HOST_DEVICE _CCCL_FORCEINLINE const T& operator[](unsigned int idx) const
+  {
+    return data()[idx];
+  }
```
miscco: Ditto
```suggestion
  [[nodiscard]] _CCCL_API _Tp& operator[](const size_t __idx) noexcept
  {
    return data()[__idx];
  }
  [[nodiscard]] _CCCL_API const _Tp& operator[](const size_t __idx) const noexcept
  {
    return data()[__idx];
  }
```

cub/test/catch2_test_detail_uninitialized_array.cu
miscco, line 1: This is an internal CCCL API, please move the test into `libcudacxx/test/libcudacxx/libcxx/memory/uninitialized_array.pass.cpp`

cub/test/catch2_test_detail_uninitialized_array.cu
```diff
@@ -0,0 +1,88 @@
...
+  construction_counter_t()
+  {
+    ++default_construction_calls;
+  }
+  int value{};
+};
+int construction_counter_t::default_construction_calls = 0;
+} // namespace
+
+C2H_TEST("detail::uninitialized_array size and alignment", "[util][detail][uninitialized_array]")
```
miscco: libcu++ uses lit as the testing infrastructure, please have a look at the surrounding files.

If you need any guidance, feel free to ask

cub/cub/detail/uninitialized_array.cuh
```diff
@@ -0,0 +1,54 @@
...
+
+namespace detail
+{
+template <typename T, ::cuda::std::size_t N, ::cuda::std::size_t Alignment = alignof(T)>
+struct uninitialized_array
+{
+  using value_type           = T;
+  static constexpr auto size = ::cuda::std::integral_constant<::cuda::std::size_t, N>{};
+  alignas(Alignment) char data_[N * sizeof(T)];
+  _CCCL_HOST_DEVICE _CCCL_FORCEINLINE T* data()
```
miscco: Question: This currently does not support any way to properly emplace any object into the array. That limmits us to is_trivially_{copy,move}_constructible types

cub/test/catch2_test_detail_uninitialized_array.cu
```diff
@@ -0,0 +1,88 @@
...
+
+C2H_TEST("detail::uninitialized_array custom alignment", "[util][detail][uninitialized_array]")
+{
+  using arr_t = cub::detail::uninitialized_array<int, 4, 32>;
+  STATIC_REQUIRE(alignof(arr_t) == 32);
+}
+
+C2H_TEST("detail::uninitialized_array element access", "[util][detail][uninitialized_array]")
+{
+  cub::detail::uninitialized_array<int, 4> arr{};
```
miscco: We should have tests for 
* basic integers
* float / double
* a type that is not trivially copy cconstructible
* a type that is not trivially move constructible
* a type that is not trivially constructible from another one

cub/cub/detail/uninitialized_array.cuh
```diff
@@ -0,0 +1,54 @@
...
+  {
+    return reinterpret_cast<T*>(data_);
+  }
+  _CCCL_HOST_DEVICE _CCCL_FORCEINLINE const T* data() const
```
miscco: Question: Should this have `begin()` and `end()` convenience functions? or are we fine with just `data()`

miscco (changes requested):
Thanks a lot for the interest. 

I agree with @bernhardmgruber that we want to move this into libcu++

unfortunately that requires a certain amount of changes, because the requirements are much stricter

charan-003:
> Thanks a lot for the interest.
> 
> I agree with @bernhardmgruber that we want to move this into libcu++
> 
> unfortunately that requires a certain amount of changes, because the requirements are much stricter

i will address the comments soon

From what I understand, the plan is to add `uninitialized_array` in libcu++, have CUB use it, and replace the duplicate in Thrust’s `cuda_cub::core::detail`. I’ll rework the PR .Thanks for the guidance.

cub/cub/detail/uninitialized_array.cuh
charan-003: i was following the [cccl/thrust/thrust/system/cuda/detail/core/util.h](https://github.com/NVIDIA/cccl/blob/b7ebe1e639950bc9a56e6c6a629afbe54d69b461/thrust/thrust/system/cuda/detail/core/util.h#L514)

well i removed it, we can get the size from the template parameter _Size directly if needed

cub/cub/detail/uninitialized_array.cuh
charan-003: i think `data() ` is sufficient for this. Adding `begin()` and `end()` would be iterator/ranges

please do let me know if we want to see `begin()/end()`

cub/test/catch2_test_detail_uninitialized_array.cu
charan-003: @miscco  for the non-trivial type tests should these be `.fail.cpp` tests that verify the `static_assert` which complain at compile time, or would you prefer a different approach?

miscco: it should be sufficient to e.g. use a type with an uninitialized data member and then assert that the member is properly zeroed out

charan-003:
before making further changes I want to make sure the implementation of uninitialized_array in the libcu++ is correct. 

Based on searching for all usages of uninitialized_array in the repo: https://github.com/search?q=repo%3ANVIDIA%2Fcccl%20uninitialized_array&type=code

I think the following changes are needed:

Replace the struct definition in [thrust/thrust/system/cuda/detail/core/util.h#L489-L523](https://github.com/NVIDIA/cccl/blob/b5b97b04de750df55a8281cf403c3e1d26929784/thrust/thrust/system/cuda/detail/core/util.h#L489-L523) with:

```cpp
template <class T, size_t N, size_t Alignment = alignof(T)>
using uninitialized_array = ::cuda::uninitialized_array<T, N, Alignment>;
```

Update [cub/cub/device/dispatch/kernels/kernel_transform.cuh#L277](https://github.com/NVIDIA/cccl/blob/b5b97b04de750df55a8281cf403c3e1d26929784/cub/cub/device/dispatch/kernels/kernel_transform.cuh#L277) to:

```cpp
using ::cuda::uninitialized_array;
```
Other usages in [set_operations.h](https://github.com/NVIDIA/cccl/blob/b5b97b04de750df55a8281cf403c3e1d26929784/thrust/thrust/system/cuda/detail/set_operations.h#L284) and[ reduce_by_key.h](https://github.com/NVIDIA/cccl/blob/b5b97b04de750df55a8281cf403c3e1d26929784/thrust/thrust/system/cuda/detail/reduce_by_key.h#L163) use `core::detail::uninitialized_array` which would continue to work transparently via util.h. Are these the right places to change, or am I missing anything?

libcudacxx/include/cuda/__memory/uninitialized_array.h
```diff
@@ -0,0 +1,65 @@
...
+#include <cuda/std/__new/launder.h>
+
+#include <cuda/std/__cccl/prologue.h>
+
+_CCCL_BEGIN_NAMESPACE_CUDA
+
+//! Aligned storage for _Tp elements (not constructed).
+//! Initialize before use with the `data()` method
+template <class _Tp, size_t _Size, size_t _Alignment = alignof(_Tp)>
+struct uninitialized_array
```
miscco: We do not want this to be publicly available, because it is a very dangerous feature. Please rename to
```suggestion
struct __uninitialized_array
```

libcudacxx/test/libcudacxx/libcxx/memory/uninitialized_array.pass.cpp
miscco, line 1: I believe this is an internal only feature we use. 

We should move this test to the other internal only tests, aka `‎libcudacxx/test/libcudacxx/libcxx/memory/uninitialized_array.pass.cpp`

miscco (changes requested):

thrust/thrust/system/cuda/detail/core/util.h
```diff
@@ -490,37 +491,7 @@ struct uninitialized
...
-  {
-    return data()[idx];
-  }
-
-  _CCCL_HOST_DEVICE T (&as_array())[N]
-  {
-    return static_cast<T(&)[N]>(data_);
-  }
-};
+using uninitialized_array = ::cuda::__uninitialized_array<T, N, Alignment>;
```
miscco: This is an internal helper type, I believe we should just use `::cuda::__uninitialized_array` internally

Could you please replace all internal uses and remove that type here?

charan-003:
@bernhardmgruber, I hope these changes will fix it.

cub/cub/device/dispatch/kernels/kernel_transform.cuh
```diff
@@ -275,10 +276,10 @@ _CCCL_DEVICE void transform_kernel_vectorized(
                                     && THRUST_NS_QUALIFIER::is_trivially_relocatable_v<output_t>;
 
   // if we can vectorize, we convert f's return type to the output type right away, so we can reinterpret later
-  using THRUST_NS_QUALIFIER::cuda_cub::core::detail::uninitialized_array;
+  using ::cuda::__uninitialized_array;
   using output_array_t                  = ::cuda::std::conditional_t<can_vectorize_store, output_t, result_t>;
   constexpr auto output_array_alignment = can_vectorize_store ? sizeof(output_t) * vec_size : alignof(result_t);
-  uninitialized_array<output_array_t, items_per_thread, output_array_alignment> output;
+  __uninitialized_array<output_array_t, items_per_thread, output_array_alignment> output;
```
bernhardmgruber: Suggestion: just use the type directly:
```suggestion
using output_array_t                  = ::cuda::std::conditional_t<can_vectorize_store, output_t, result_t>;
constexpr auto output_array_alignment = can_vectorize_store ? sizeof(output_t) * vec_size : alignof(result_t);
  ::cuda::__uninitialized_array<output_array_t, items_per_thread, output_array_alignment> output;
```

cub/cub/device/dispatch/kernels/kernel_transform.cuh
```diff
@@ -325,9 +326,9 @@ _CCCL_DEVICE void transform_kernel_vectorized(
       output[i] = f(inputs[i]...);
     }
   };
-  provide_array(uninitialized_array<it_value_t<RandomAccessIteratorsIn>,
-                                    items_per_thread,
-                                    sizeof(it_value_t<RandomAccessIteratorsIn>) * vec_size>{}...);
+  provide_array(__uninitialized_array<it_value_t<RandomAccessIteratorsIn>,
+                                      items_per_thread,
+                                      sizeof(it_value_t<RandomAccessIteratorsIn>) * vec_size>{}...);
```
bernhardmgruber: Same here

cub/cub/device/dispatch/tuning/common.cuh
```diff
@@ -51,7 +51,7 @@ template <typename T>
 inline constexpr auto classify_type = type_t::other;
 
 template <>
-inline constexpr auto classify_type<char> = ::cuda::std::is_signed_v<char> ? type_t::int8 : type_t::uint8;
+inline constexpr auto classify_type<char> = (::cuda::std::is_signed_v<char>) ? type_t::int8 : type_t::uint8;
```
bernhardmgruber: Q: Why is this change necessary?

bernhardmgruber (commented):
@miscco can you please review the libcu++ changes?

I also want to see a SASS diff for the fill benchmark of device transform. I can do that once we agree on the proposed changes.

cub/cub/device/dispatch/tuning/common.cuh
charan-003: I noticed several `expected '>'` parse errors and assumed wrapping the conditions would resolve them.

Would you like me to revert those changes? I also applied the same fix to a few other files.

fbusato:
> What is the use case for that? I have libcudacxx/include/cuda/std/__pstl/cuda/temporary_storage.h
> 
> otherwise we already have cuda::__uninitialized_async_buffer that does a lot of the things

@pciolkosz how much overlap we have in this direction?

charan-003:
@bernhardmgruber @miscco i've applied the suggested changes for the direct type
Please do let me know if any more changes are required
Thanks

bernhardmgruber (approved):

bernhardmgruber:
Waiting on another review from @miscco

miscco (approved):

miscco:
Looks good to me, sory @charan-003 that this took so long

libcudacxx/test/libcudacxx/libcxx/memory/uninitialized_array.pass.cpp
```diff
@@ -0,0 +1,111 @@
...
+  assert(arr_short[0] == 123);
+  arr_int[0] = 42;
+  assert(arr_int[0] == 42);
+  arr_long[0] = 100L;
+  assert(arr_long[0] == 100L);
+  arr_longlong[0] = 999LL;
+  assert(arr_longlong[0] == 999LL);
+}
+
+__host__ __device__ void test_floating_point()
```
miscco: Final comment: We need to replace open coded `__host__ __device__` with `TEST_FUNC` because of future tile support.

miscco (changes requested):

miscco (approved):

bernhardmgruber:
Thank you for contribution! Those small things really help to make the code base better over time!

PR #8173, opened 2026-03-25
Add shellcheck

Jacobfaib:
## Description

Add `shellcheck` to `pre-commit`.

The overwhelming majority of fixes in this PR were performed automatically, with
```sh
$ fd -e sh -e bash | shellcheck --format=diff | git apply
```

The main set of fixes (which account for almost all diffs) are:

- Use `${}` around variable names.
- [`SC2155`](https://www.shellcheck.net/wiki/SC2155): Declare and assign separately to avoid masking return values.
- [`SC2006`](https://www.shellcheck.net/wiki/SC2006): Use `$(...)` notation instead of legacy backticked `` `...` ``.
- [`SC2230`](https://www.shellcheck.net/wiki/SC2230): `which` is non-standard. Use builtin `command -v` instead.

## Checklist

### DIFF
--- a/.devcontainer/cccl-entrypoint.sh
+++ b/.devcontainer/cccl-entrypoint.sh
@@ -5,8 +5,10 @@
 set -e;
 
 if ! test -n "${DISABLE_SCCACHE:+x}" && test -n "${DEVCONTAINER_UTILS_ENABLE_SCCACHE_DIST:+x}" && ! test -n "${SCCACHE_DIST_URL:+x}"; then
-    export SCCACHE_DIST_URL="https://$(dpkg --print-architecture).$(uname -s | tr '[:upper:]' '[:lower:]').sccache.rapids.nvidia.com";
-    echo "export SCCACHE_DIST_URL=$SCCACHE_DIST_URL" >> ~/.bashrc;
+    arch_str="$(dpkg --print-architecture)"
+    uname_str="$(uname -s | tr '[:upper:]' '[:lower:]')"
+    export SCCACHE_DIST_URL="https://${arch_str}.${uname_str}.sccache.rapids.nvidia.com";
+    echo "export SCCACHE_DIST_URL=${SCCACHE_DIST_URL}" >> ~/.bashrc;
 fi
 
 if [[ -n "${GITHUB_ACTIONS:-}" ]]; then
--- a/.devcontainer/docker-entrypoint.sh
+++ b/.devcontainer/docker-entrypoint.sh
@@ -8,42 +8,48 @@
 : "${NEW_UID:=}";
 : "${NEW_GID:=}";
 
-eval "$(sed -n "s/${REMOTE_USER}:[^:]*:\([^:]*\):\([^:]*\):[^:]*:\([^:]*\).*/OLD_UID=\1;OLD_GID=\2;HOME_FOLDER=\3/p" /etc/passwd)";
-eval "$(sed -n "s/\([^:]*\):[^:]*:${NEW_UID}:.*/EXISTING_USER=\1/p" /etc/passwd)";
-eval "$(sed -n "s/\([^:]*\):[^:]*:${NEW_GID}:.*/EXISTING_GROUP=\1/p" /etc/group)";
+OLD_UID="$(sed -n "s/${REMOTE_USER}:[^:]*:\([^:]*\):\([^:]*\):[^:]*:\([^:]*\).*/\1/p" /etc/passwd)";
+OLD_GID="$(sed -n "s/${REMOTE_USER}:[^:]*:\([^:]*\):\([^:]*\):[^:]*:\([^:]*\).*/\2/p" /etc/passwd)";
+HOME_FOLDER="$(sed -n "s/${REMOTE_USER}:[^:]*:\([^:]*\):\([^:]*\):[^:]*:\([^:]*\).*/\3/p" /etc/passwd)";
+EXISTING_USER="$(sed -n "s/\([^:]*\):[^:]*:${NEW_UID}:.*/\1/p" /etc/passwd)";
+EXISTING_GROUP="$(sed -n "s/\([^:]*\):[^:]*:${NEW_GID}:.*/\1/p" /etc/group)";
 
-if [ -z "$OLD_UID" ]; then
-    echo "Remote user not found in /etc/passwd ($REMOTE_USER).";
-    exec "$(pwd)/.devcontainer/cccl-entrypoint.sh" "$@";
-elif [ "$OLD_UID" = "$NEW_UID" ] && [ "$OLD_GID" = "$NEW_GID" ]; then
-    echo "UIDs and GIDs are the same ($NEW_UID:$NEW_GID).";
+if [[ -z "${OLD_UID}" ]]; then
+    echo "Remote user not found in /etc/passwd (${REMOTE_USER}).";
+    entry_script="$(pwd)/.devcontainer/cccl-entrypoint.sh";
+    exec "${entry_script}" "$@";
+elif [[ "${OLD_UID}" = "${NEW_UID}" ]] && [[ "${OLD_GID}" = "${NEW_GID}" ]]; then
+    echo "UIDs and GIDs are the same (${NEW_UID}:${NEW_GID}).";
     # Even when IDs match, ensure we execute as the non-root REMOTE_USER so
     # gh and sccache use the mapped HOME (/home/coder) where ~/.aws is bind-mounted.
     export VIRTUAL_ENV=;
     export VIRTUAL_ENV_PROMPT=;
-    export HOME="$HOME_FOLDER";
-    export XDG_CACHE_HOME="$HOME_FOLDER/.cache";
-    export XDG_CONFIG_HOME="$HOME_FOLDER/.config";
-    export XDG_STATE_HOME="$HOME_FOLDER/.local/state";
-    export PYTHONHISTFILE="$HOME_FOLDER/.local/state/.python_history";
-    exec su -p "$REMOTE_USER" -- "$(pwd)/.devcontainer/cccl-entrypoint.sh" "$@";
-elif [ "$OLD_UID" != "$NEW_UID" ] && [ -n "$EXISTING_USER" ]; then
-    echo "User with UID exists ($EXISTING_USER=$NEW_UID).";
-    exec "$(pwd)/.devcontainer/cccl-entrypoint.sh" "$@";
+    export HOME="${HOME_FOLDER}";
+    export XDG_CACHE_HOME="${HOME_FOLDER}/.cache";
+    export XDG_CONFIG_HOME="${HOME_FOLDER}/.config";
+    export XDG_STATE_HOME="${HOME_FOLDER}/.local/state";
+    export PYTHONHISTFILE="${HOME_FOLDER}/.local/state/.python_history";
+    entry_script="$(pwd)/.devcontainer/cccl-entrypoint.sh"
+    exec su -p "${REMOTE_USER}" -- "${entry_script}" "$@";
+elif [[ "${OLD_UID}" != "${NEW_UID}" ]] && [[ -n "${EXISTING_USER}" ]]; then
+    echo "User with UID exists (${EXISTING_USER}=${NEW_UID}).";
+    entry_script="$(pwd)/.devcontainer/cccl-entrypoint.sh";
+    exec "${entry_script}" "$@";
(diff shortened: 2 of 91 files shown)


.devcontainer/cccl-entrypoint.sh
```diff
@@ -5,8 +5,10 @@
 set -e;
 
 if ! test -n "${DISABLE_SCCACHE:+x}" && test -n "${DEVCONTAINER_UTILS_ENABLE_SCCACHE_DIST:+x}" && ! test -n "${SCCACHE_DIST_URL:+x}"; then
-    export SCCACHE_DIST_URL="https://$(dpkg --print-architecture).$(uname -s | tr '[:upper:]' '[:lower:]').sccache.rapids.nvidia.com";
-    echo "export SCCACHE_DIST_URL=$SCCACHE_DIST_URL" >> ~/.bashrc;
+    arch_str="$(dpkg --print-architecture)"
+    uname_str="$(uname -s | tr '[:upper:]' '[:lower:]')"
+    export SCCACHE_DIST_URL="https://${arch_str}.${uname_str}.sccache.rapids.nvidia.com";
+    echo "export SCCACHE_DIST_URL=${SCCACHE_DIST_URL}" >> ~/.bashrc;
```
trxcllnt, line 11: So... I tend to think [SC2155](https://www.shellcheck.net/wiki/SC2155) is overly restrictive to the point of uselessness in most instances. Neither `dpkg --print-architecture` nor `uname -s` will fail here, so it's just noise for little benefit.

If we're worried about portability to non-debian OSs, we could replace `dpkg --print-architecture` with `uname -m | sed -e 's/x86_/amd/' -e 's/aarch/arm/'`.

.devcontainer/docker-entrypoint.sh
```diff
@@ -8,54 +8,60 @@
...
-    export HOME="$HOME_FOLDER";
-    export XDG_CACHE_HOME="$HOME_FOLDER/.cache";
-    export XDG_CONFIG_HOME="$HOME_FOLDER/.config";
-    export XDG_STATE_HOME="$HOME_FOLDER/.local/state";
-    export PYTHONHISTFILE="$HOME_FOLDER/.local/state/.python_history";
-    exec su -p "$REMOTE_USER" -- "$(pwd)/.devcontainer/cccl-entrypoint.sh" "$@";
-elif [ "$OLD_UID" != "$NEW_UID" ] && [ -n "$EXISTING_USER" ]; then
-    echo "User with UID exists ($EXISTING_USER=$NEW_UID).";
-    exec "$(pwd)/.devcontainer/cccl-entrypoint.sh" "$@";
+    export HOME="${HOME_FOLDER}";
```
trxcllnt, line 27: Did shellcheck do all the `s/$VAR/${VAR}/` transforms in this PR too?

Jacobfaib: Yes. You can get shellcheck to do it by adding the "style" option to the enabled checks. Then
```sh
$ fd -e sh -e bash | shellcheck --format=diff | git apply
```
(or `find . -name ...` if you prefer that) to apply them all.

.devcontainer/cccl-entrypoint.sh
Jacobfaib: Hmmm yeah I could disable it. I already disabled `SC2312` (very similar, warns about splitting subcommands inside substitution), but I figured SC2155 is simple enough to work around that it wasn't so bad.

ci/test_cub.sh
```diff
@@ -24,7 +24,7 @@ new_args=$("${ci_dir}/util/extract_switches.sh" \
   -compute-sanitizer-synccheck \
   -- "$@")
 
-eval set -- ${new_args}
+eval set -- "${new_args}"
```
trxcllnt: I _think_ I understand this change -- the word splitting happens in the call to eval instead of before it.

Does `new_args` contain other commands that should be evaluated at runtime? If not, I'd find something like this easier to follow:

```suggestion
declare -a new_args="($(<the command before this comment>))"
set -- "${new_args[@]}"
```

Jacobfaib: > Does new_args contain other commands that should be evaluated at runtime?

I don't think so. At the very least, all the command-line arguments should already have been shell-expanded and any unknown flags just lead to a crash. I'll fix up every instance where this pattern occurs

Jacobfaib: Hmm running into https://www.shellcheck.net/wiki/SC2207 now. I suppose I could do
```sh
IFS=" " read -ra new_args <<< "$(extract_switches.sh ... -- "$@")"
```
but man that is _ugly_. I think I will just silence the warning.

trxcllnt: From what I understand, `declare -a new_args="($(...))"` shouldn't trigger SC2207 since the declaration is quoted. Alternatively, `declare -a "new_args=($(...))"` should be equivalent, but it doesn't read as well to my eyes :sweat_smile:.

.devcontainer/docker-entrypoint.sh
trxcllnt: The reason I ask is that the style change makes it difficult to review the PR as-is. If all the other lint were applied separately, then the style change was made in a separate commit, that'd be easier to compare.

ci/test_cub.sh
```diff
@@ -22,9 +26,9 @@ new_args=$("${ci_dir}/util/extract_switches.sh" \
   -compute-sanitizer-racecheck \
   -compute-sanitizer-initcheck \
   -compute-sanitizer-synccheck \
-  -- "$@")
+  -- "$@"))
 
-eval set -- "${new_args}"
+set -- "${new_args[@]}"
```
trxcllnt: Any reason not to use `declare -a new_args="($(...))"`? Without it, the shell splits on spaces regardless of quoting rules:
```shell
$ strs=("a b" "cd" "e f")

$ ary1=($(echo "${strs[@]@Q}"))
$ echo "${ary1[@]}"
'a b' 'cd' 'e f'
$ echo "${ary1[@]@Q}"
''\''a' 'b'\''' ''\''cd'\''' ''\''e' 'f'\'''
$ echo "${#ary1[@]}"
5

$ declare -a ary2="($(echo "${strs[@]@Q}"))"
$ echo "${ary2[@]}"
a b cd e f
$ echo "${ary2[@]@Q}"
'a b' 'cd' 'e f'
$ echo "${#ary2[@]}"
3
```

`ary1` is a 5-element array whose elements include the single-tick quote chars, but `ary2` faithfully represents the original 3-element array.

I don't know what `extract_switches.sh` echos, but if its output includes quotes to preserve spaces in values, `declare -a new_args="($(...))"` is the correct form to use.

Jacobfaib: Ah sorry, yeah I can do that.

Jacobfaib: Hmm but this now trips up https://www.shellcheck.net/wiki/SC2178 (variable declared as array but assigned as string). Testing locally, your version does do The Right Thing(TM), so I guess I will just go ahead and silence `SC2178`? Unless you have another shell incantation up your sleeve that shellcheck wouldn't warn for.

Jacobfaib: Ah, there's actually a very subtle interaction here.
```sh
declare -a foo
foo="(a b c)"
> foo = "(a b c)"
```
is _not_ the same as
```sh
declare -a foo="(a b c)"
> foo = (a b c)
```
The latter will do the proper array splitting, while the former does not, it just stays a string. I don't know if this is a bashism, or perhaps an undocumented feature (my google-fu is failing me here), but given the sensitivity of it I might suggest reverting to the old form
```sh
new_args=(...)
```
and silencing `SC2207`. Or just going whole hog and doing `IFS=" " read -ra new_args <<< "$(extract_switches.sh ... -- "$@")"`... at least that works and produces no errors.

trxcllnt: > ```shell
> declare -a foo
> foo="(a b c)"
> ```

Yes, this is declaring an Array and then assigning a string as its first element. Any operation on an Array that doesn't use an explicit index or expansion will implicitly operate on the first element:

```shell
$ foo=('a b' 'cd')
$ echo "${foo@Q} - ${foo[0]@Q}"
'a b' - 'a b'

$ foo='e f'
$ echo "${foo@Q} - ${foo[0]@Q}"
'e f' - 'e f'
```

The `@Q` in `${VAR@Q}` is a [shell parameter expansion](https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html) that quotes `VAR` so it can be reused as input.

The form I mentioned shouldn't trigger SC2178, since we're not assigning a string as the first element of an Array, we're declaring an Array that contains the values using normal quoting rules.

> but given the sensitivity of it I might suggest reverting to the old form ... and silencing SC2207. Or just going whole hog and doing IFS=" " read -ra new_args <<< "$(extract_switches.sh ... -- "$@")"... at least that works and produces no errors.

See here, that also isn't equivalent to the form I'm suggesting:
```shell
# Incorrect
$ IFS=" " read -ra ary1 <<< "'a b' 'cd' 'e f'"
$ echo ${ary1[*]@Q}
''\''a' 'b'\''' ''\''cd'\''' ''\''e' 'f'\'''

# Correct
$ declare -a ary2="($(echo "'a b' 'cd' 'e f'"))"
$ echo ${ary2[*]@Q}
'a b' 'cd' 'e f'
```

In the `read -ra` above, we're telling `read` to split the input string on spaces. However `read` doesn't know the input string represents an Array whose _elements_ contain spaces, so the single tick quotes end up as part of the Array elements, rather than denoting the boundaries _between_ Array elements.

You can see this with the `@K` parameter expansion:
```shell
$ echo ${ary1[*]@K}
0 "'a" 1 "b'" 2 "'cd'" 3 "'e" 4 "f'"

$ echo ${ary2[*]@K}
0 "a b" 1 "cd" 2 "e f"
```

The statement `declare -a ary="('a b' 'cd' 'e f')"` is _not_ assigning a string to an Array, it is parsing the string to an Array using both the quoting _and_ word splitting rules, so spaces inside quotes (`'a<SPACE>b'`) are not treated the same as spaces between Array elements (`'a b'<SPACE>'cd'`).

Ok, so why does this matter? Let's say `extract_switches.sh` does this:
```
echo "--build '../path/with a/space'"
```

If we only split the output on spaces (via `read -ra` or `new_args=(...)`), we'll get an Array with three elements:
```shell
new_args[0] = "--build"
new_args[1] = "'../path/with"
new_args[2] = "a/space'"
```

And that would not be correct input if for example `new_args` is intended to be used as input to `cmake`:
```shell
cmake "--build" "'../path/with" "a/space'"
```
CMake sees the build dir is the string literal `'../path/with`, instead of `../path/with\ a/space`.

This may seem like a nit, but it's an important distinction, especially if `extract_switches.sh` output includes things like `$(pwd)`, where the absolute path may have spaces e.g. `/home/Users Full Name/cccl`.

Jacobfaib: The issue I found with `declare -a ary="(echo ...)"` is that it masks the return value of whatever is inside the array. If `extract_switches.sh` fails (and it definitely can, there are error checks within it), the value of `ary` will potentially be anything.

Sure it then _probably_ fails at `set --` but no guarantees (and if it does, that error message will be inscrutible, probably the classic `bash syntax error: line 0 col 1`).

The solution I ended up going with, going back to string + `eval` (which is needed to expand out the `""` inserted by `extract_switches`), seems like the least worst option, but am open to suggestions.

trxcllnt: Why not assign it to a string then declare the array?

```shell
new_args="$(...)"
declare -a new_args="($new_args)"
```

Jacobfaib: Can do that too, it seems to pass `shellcheck` without warning (and doesn't require `eval`!). Let me switch all the instances over to that then

.devcontainer/docker-entrypoint.sh
trxcllnt: I am going to try to replicate this PR locally without the `style` option, there's just too much noise from the variable transformations to effectively review any potential issues with logic changes.

trxcllnt: hm... shellcheck doesn't seem to work if I add `disable=SC2250` to the .shellcheckrc:
```shell
$ shellcheck --format=diff ./.devcontainer/verify_devcontainer.sh
Issues were detected, but none were auto-fixable. Use another format to see them.
```

It does produce a diff if I remove `disable=SC2250`. Is this a bug, or does it really expect me to fix each file manually?

PR #8183, opened 2026-03-26
Refactor `WarpExchangeShfl`

bernhardmgruber:
I wanted to cleanup the weird implementation but had some trouble getting good SASS generated in the end. I am not fully certain the result is a lot better than before, so I am fine with abandoning the PR as well.

Also, `WarpExchangeShfl` is not used *anywhere* in CUB. It can only be enabled via tuning, but we don't do that.

SASS diff: `ref_WarpExchangeShfl` (`93599e82b8`) vs `upstream/main` (`d0e21c2a5e`), sm_120

**Target:** `cub.test.warp.exchange_shfl` (40 `WARP_EXCHANGE_SHUFFLE` kernel instantiations — the only compiled artifact exercising this algorithm). All 40 test cases pass.

| Category | Count |
|---|---|
| Byte-identical | 8 |
| Pure register renaming | 4 |
| Same-opcode scheduling noise (reordered, identical instruction multiset) | 16 |
| Real opcode-count deltas | 12 |

All 12 "real" deltas are tiny — `{'NOP': ±1..3, 'MOV': ±1..8}` — a `MOV` count wobble tied to `.reuse` register-cache hints and instruction reordering, not a semantic difference.

### DIFF
--- a/cub/cub/warp/specializations/warp_exchange_shfl.cuh
+++ b/cub/cub/warp/specializations/warp_exchange_shfl.cuh
@@ -179,7 +179,10 @@ class WarpExchangeShfl
     v         = __shfl_xor_sync(mask, v, NUM_ENTRIES, LOGICAL_WARP_THREADS);
 
     constexpr int next_idx = IDX + 1 + ((IDX + 1) % NUM_ENTRIES == 0) * NUM_ENTRIES;
-    transpose_foreach<next_idx, NUM_ENTRIES>(vals, xor_bit_set, mask);
+    if constexpr (next_idx < NUM_ENTRIES)
+    {
+      transpose_foreach<NUM_ENTRIES, next_idx>(vals, xor_bit_set, mask);
+    }
   }
 
   template <int NUM_ENTRIES>
@@ -189,7 +192,7 @@ class WarpExchangeShfl
     if constexpr (NUM_ENTRIES != 0)
     {
       const bool xor_bit_set = lane_id & NUM_ENTRIES;
-      transpose_foreach<0, NUM_ENTRIES>(vals, xor_bit_set, mask);
+      transpose_foreach<NUM_ENTRIES, 0>(vals, xor_bit_set, mask);
 
       transpose<NUM_ENTRIES / 2>(vals, lane_id, mask);
     }
@@ -215,6 +218,8 @@ public:
   BlockedToStriped(const InputT (&input_items)[ITEMS_PER_THREAD], OutputT (&output_items)[ITEMS_PER_THREAD])
   {
     InputT vals[ITEMS_PER_THREAD];
+
+    _CCCL_PRAGMA_UNROLL_FULL()
     for (int i = 0; i < ITEMS_PER_THREAD; i++)
     {
       vals[i] = input_items[i];
@@ -222,6 +227,7 @@ public:
 
     transpose<ITEMS_PER_THREAD / 2>(lane_id, member_mask);
 
+    _CCCL_PRAGMA_UNROLL_FULL()
     for (int i = 0; i < ITEMS_PER_THREAD; i++)
     {
       output_items[i] = vals[i];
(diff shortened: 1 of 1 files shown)


cub/cub/warp/specializations/warp_exchange_shfl.cuh
```diff
@@ -260,8 +214,18 @@ public:
   _CCCL_DEVICE _CCCL_FORCEINLINE void
   BlockedToStriped(const InputT (&input_items)[ITEMS_PER_THREAD], OutputT (&output_items)[ITEMS_PER_THREAD])
   {
-    CompileTimeArray<OutputT, 0, ITEMS_PER_THREAD> arr{input_items, output_items};
-    arr.Transpose(lane_id, member_mask);
+    InputT vals[ITEMS_PER_THREAD];
+    for (int i = 0; i < ITEMS_PER_THREAD; i++)
```
miscco: We are commonly adding
```suggestion
    _CCCL_PRAGMA_UNROLL_FULL()
    for (int i = 0; i < ITEMS_PER_THREAD; i++)
```

cub/cub/warp/specializations/warp_exchange_shfl.cuh
```diff
@@ -260,8 +214,18 @@ public:
...
+    {
+      vals[i] = input_items[i];
+    }
+
+    transpose<ITEMS_PER_THREAD / 2>(lane_id, member_mask);
+
+    for (int i = 0; i < ITEMS_PER_THREAD; i++)
```
miscco:
```suggestion
    _CCCL_PRAGMA_UNROLL_FULL()
    for (int i = 0; i < ITEMS_PER_THREAD; i++)
```

cub/cub/warp/specializations/warp_exchange_shfl.cuh
```diff
@@ -34,212 +34,166 @@ class WarpExchangeShfl
...
+    v         = __shfl_xor_sync(mask, v, NUM_ENTRIES, LOGICAL_WARP_THREADS);
 
-  public:
-    _CCCL_DEVICE
-    CompileTimeArray(const InputT (&input_items)[ITEMS_PER_THREAD], OutputT (&output_items)[ITEMS_PER_THREAD])
-        : CompileTimeArray<OutputT, IDX + 1, SIZE>{input_items, output_items}
-        , val{input_items[IDX]}
-    {}
+    constexpr int next_idx = IDX + 1 + ((IDX + 1) % NUM_ENTRIES == 0) * NUM_ENTRIES;
+    transpose_foreach<next_idx, NUM_ENTRIES>(vals, xor_bit_set, mask);
```
miscco: Maybe I am stupid, but how does this function terminate?

bernhardmgruber: No, you are not. I indeed missed the termination condition. Thx!

cub/cub/warp/specializations/warp_exchange_shfl.cuh
```diff
@@ -34,215 +35,189 @@ class WarpExchangeShfl
...
   static constexpr bool IS_ARCH_WARP = LOGICAL_WARP_THREADS == warp_threads;
 
-  // Hide recursive template from Doxygen — it cannot handle self-referential inheritance.
-#ifndef _CCCL_DOXYGEN_INVOKED
-  // concrete recursion class
-  template <typename OutputT, int IDX, int SIZE>
-  class CompileTimeArray : protected CompileTimeArray<OutputT, IDX + 1, SIZE>
+  template <int NUM_ENTRIES, int IDX>
+  _CCCL_DEVICE _CCCL_FORCEINLINE void
+  transpose_foreach(InputT (&vals)[ITEMS_PER_THREAD], const bool xor_bit_set, const unsigned mask)
```
miscco, line 40: Should that be a private method?

bernhardmgruber: Yes, which it is.

miscco: Oh github did not show me 10 lines where the `public:` was hidden ^^

miscco (approved):
I believe the code is better to read / follow, So I am fine with merging especially given that its unused

PR #8186, opened 2026-03-26
Allow passing additional build options to the cmake build command

Jacobfaib:
## Description

Allows passing specific options through to `cmake --build` (such as tool specific ones like `-k 0`).

Currently this is only accessible by calling `build_preset` directly, since `configure_and_build_preset` will gobble any extra args and pass them to `configure_preset`.

Needed by https://github.com/NVIDIA/cccl/pull/8073

## Checklist


Jacobfaib:
Will be covered by https://github.com/NVIDIA/cccl/pull/8073

PR #8194, opened 2026-03-26
[STF] Replace type_name implementation with cuda::std::__pretty_nameof wrapper

caugonnet:
Remove the home-grown __PRETTY_FUNCTION__ parsing in traits.cuh and replace type_name<T> with a thin wrapper around cuda::std::__pretty_nameof<T>(), which is the existing robust CCCL facility for compile-time type name extraction.

## Description

closes 

## Checklist


caugonnet:
@andralex type_name returns a string_view and the operators that __pretty_nameof lack, and we can't really add such a method to get a string view to __pretty_nameof so I left it because it's indeed much more convenient.

PR #8206, opened 2026-03-28
Replace `[CUB|THRUST]_DETAIL_KERNEL_ATTRIBUTES` by `CCCL_DETAIL_KERNEL_ATTRIBUTES`

bernhardmgruber:


bernhardmgruber:
@miscco why is the macro `CCCL_DETAIL_KERNEL_ATTRIBUTES` not called `_CCCL_KERNEL_ATTRIBUTES`?

cub/cub/util_macro.cuh
```diff
@@ -23,10 +23,6 @@
 
 CUB_NAMESPACE_BEGIN
 
-#ifndef CUB_DETAIL_KERNEL_ATTRIBUTES
-#  define CUB_DETAIL_KERNEL_ATTRIBUTES CCCL_DETAIL_KERNEL_ATTRIBUTES
-#endif
```
davebayer, line 28: Could use define `CUB_DETAIL_KERNEL_ATTRIBUTES` himself to override the default behaviour? If yes, this PR will break this, because users will not define `CCCL_DETAIL_KERNEL_ATTRIBUTES`.

Also, if `CCCL_DETAIL_KERNEL_ATTRIBUTES` shouldn't be modified by the user, we shouldn't guard the definition by `#ifdef CCCL_DETAIL_KERNEL_ATTRIBUTES`

miscco:
> @miscco why is the macro `CCCL_DETAIL_KERNEL_ATTRIBUTES` not called `_CCCL_KERNEL_ATTRIBUTES`?

Probably coming from the fact that the other ones are public and we werent sure whether this is something we want to expose?

I am happy to make it private though

cub/cub/util_macro.cuh
bernhardmgruber: Since the macro has `DETAIL` in its name, I would argue it's not meant to be changed by the user. The README says:

> Any symbol whose name contains detail

is not part of the public API. Technically, we didn't spell `DETAIL`, but I think that macro belongs to us.

bernhardmgruber:
> > @miscco why is the macro `CCCL_DETAIL_KERNEL_ATTRIBUTES` not called `_CCCL_KERNEL_ATTRIBUTES`?
> 
> Probably coming from the fact that the other ones are public and we werent sure whether this is something we want to expose?

The other ones contain the term `DETAIL` and are thus for internal use. But we can be nice in case somebody relies on this and rename `CCCL_DETAIL_KERNEL_ATTRIBUTES` to `_CCCL_KERNEL_ATTRIBUTES` and allow the user to redefine it, adding a big comment that we will no longer allow re-definition at a later point. I will push those changes.

libcudacxx/include/cuda/std/__cccl/visibility.h
```diff
@@ -84,9 +84,11 @@
...
 #endif // !_CCCL_COMPILER(NVHPC)
 
-#if !defined(CCCL_DETAIL_KERNEL_ATTRIBUTES)
-#  define CCCL_DETAIL_KERNEL_ATTRIBUTES __global__ _CCCL_VISIBILITY_HIDDEN
-#endif // !CCCL_DETAIL_KERNEL_ATTRIBUTES
+// Note: we will allow the user to redefine _CCCL_KERNEL_ATTRIBUTES until at least CCCL 4.0, since they may have
+// redefined CUB_DETAIL_KERNEL_ATTRIBUTES or THRUST_DETAIL_KERNEL_ATTRIBUTES.
+#if !defined(_CCCL_KERNEL_ATTRIBUTES)
+#  define _CCCL_KERNEL_ATTRIBUTES __global__ _CCCL_VISIBILITY_HIDDEN
+#endif // !_CCCL_KERNEL_ATTRIBUTES
```
davebayer: I don't think this makes sense, because if someone is already using custom kernel attributes, they already define `(CCCL|CUB|THRUST)_DETAIL_KERNEL_ATTRIBUTES`. So, it would be a source breaking change.

I think we should either check for all of the definitions or just do the source breaking change and just always define `_CCCL_KERNEL_ATTRIBUTES`

davebayer: Please, update the docs for this cccl attribute

miscco:
yeah lets just use an internal name and deprecate the use of the old ones

libcudacxx/include/cuda/std/__cccl/visibility.h
bernhardmgruber: Here is my reasoning:

> if someone is already using custom kernel attributes, they already define `(CCCL|CUB|THRUST)_DETAIL_KERNEL_ATTRIBUTES`.

Which is a violation of our API guarantees, since the user is modifying an internal entity.

> So, it would be a source breaking change.

Which is fine, since the user already broke the contract.

> I think we should either check for all of the definitions or just do the source breaking change and just always define `_CCCL_KERNEL_ATTRIBUTES`

I was leaning towards always defining `_CCCL_KERNEL_ATTRIBUTES`, but then realized it does not cost us much to support users doing a bad thing for at least some migration time.

I chose now to error when the user defines `(CUB|THRUST)_DETAIL_KERNEL_ATTRIBUTES`, pointing them to `_CCCL_KERNEL_ATTRIBUTES` with a warning that we are taking that way in CCCL 4.0.

miscco (approved):

PR #8210, opened 2026-03-30
Add warp shuffle constraints

davebayer:
The implementation relies on the fact that the type is trivially copyable and default constructible, but doesn't actually check for these properties.


fbusato:
we didn't add this constraint on purpose. Otherwise, we are excluding data types where the assertion fails but the type is trivially copyable, e.g. `__half`, `__nv_bfloat16`, etc. I strongly suggest to revert it

PR #8211, opened 2026-03-30
[cudax] Return `cuda::std::optional` from group algorithms

davebayer:
Since only the `root` rank returns the proper value from collective algorithms, me and @bernhardmgruber, we think the result should be wrapped by `cuda::std::optional` and only the `root` thread should have the value initialized to prevent reads of invalid values.

### DIFF
--- a/cudax/test/hierarchy/group.cu
+++ b/cudax/test/hierarchy/group.cu
@@ -17,6 +17,7 @@
 #include <cuda/devices>
 #include <cuda/hierarchy>
 #include <cuda/launch>
+#include <cuda/std/optional>
 #include <cuda/std/utility>
 #include <cuda/stream>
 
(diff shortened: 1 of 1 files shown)


miscco:
I am slightly torn here.

I love my `optional` like anybody else, but shouldnt we just make sure that we broadcast to all threads?

cudax/test/hierarchy/group.cu
```diff
@@ -27,24 +28,25 @@
...
-__device__ T sum(cudax::this_warp<Hierarchy> group, T (&array)[N])
+__device__ cuda::std::optional<T> sum(cudax::this_warp<Hierarchy> group, T (&array)[N])
 {
   using WarpReduce = cub::WarpReduce<T>;
 
   __shared__ typename WarpReduce::TempStorage scratch;
 
   const auto partial = cub::ThreadReduce(array, cuda::std::plus<T>{});
-  return WarpReduce{scratch}.Sum(partial);
+  const auto result  = WarpReduce{scratch}.Sum(partial);
```
caugonnet, line 44: isn't it the Sum in WarpReduce that should return an optional ?

davebayer:
> I love my optional like anybody else, but shouldnt we just make sure that we broadcast to all threads?

I think we can provide an API doing that later :) We can even choose which one should be the default one

cudax/test/hierarchy/group.cu
davebayer: Oh, I don't want to change any of the current cub algorithms, this is just to give us some idea how the cooperative algorithms could look like

Issue #7475, opened 2026-02-03
Refactor cccl.c `histogram` to use new tuning API


jrhemstad:
This issue can be closed by updating the `cub::DeviceHistogram` C++ API to work with the new tuning API and updating the corresponding cccl.c `histogram` algorithm.

PR #8212, opened 2026-03-30
Implement the new tuning API for `DispatchHistogram`

bernhardmgruber:
- [x] No SASS changes in `cub.bench.histogram.even.base` for SM`75;90;100`
- [x] No SASS changes in `cub.bench.histogram.range.base` for SM`75;90;100`
- [x] No SASS changes in `cub.bench.histogram.multi.even.base` for SM`75;90;100`
- [x] No SASS changes in `cub.bench.histogram.multi.range.base` for SM`75;90;100`

The histogram dispatcher is unfortunately carrying some technical dept. If this PR turns out to be too hard to review, I can try to push a refactoring and rebase on top, to make the rewrite here smaller. Please let me know.

Fixes: #7637
Fixes: #7475

### DIFF
--- a/cub/cub/device/dispatch/tuning/tuning_histogram.cuh
+++ b/cub/cub/device/dispatch/tuning/tuning_histogram.cuh
@@ -56,21 +56,21 @@ enum class counter_size
 
 // TODO(bgruber): drop in CCCL 4.0
 template <class T>
-constexpr primitive_sample is_primitive_sample()
+_CCCL_API constexpr primitive_sample is_primitive_sample()
 {
   return is_primitive<T>::value ? primitive_sample::yes : primitive_sample::no;
 }
 
 // TODO(bgruber): drop in CCCL 4.0
 template <class CounterT>
-constexpr counter_size classify_counter_size()
+_CCCL_API constexpr counter_size classify_counter_size()
 {
   return sizeof(CounterT) == 4 ? counter_size::_4 : counter_size::unknown;
 }
 
 // TODO(bgruber): drop in CCCL 4.0
 template <class SampleT>
-constexpr sample_size classify_sample_size()
+_CCCL_API constexpr sample_size classify_sample_size()
 {
   return sizeof(SampleT) == 1 ? sample_size::_1 : sizeof(SampleT) == 2 ? sample_size::_2 : sample_size::unknown;
 }
@@ -167,7 +167,7 @@ struct policy_hub
   // TODO(bgruber): move inside t_scale in C++14
   static constexpr int v_scale = (sizeof(SampleT) + sizeof(int) - 1) / sizeof(int);
 
-  static constexpr int t_scale(int nominalItemsPerThread)
+  _CCCL_API static constexpr int t_scale(int nominalItemsPerThread)
   {
     return (::cuda::std::max) (nominalItemsPerThread / NumActiveChannels / v_scale, 1);
   }
@@ -185,7 +185,7 @@ struct policy_hub
   {
     // Use values from tuning if a specialization exists, otherwise pick Policy500
     template <typename Tuning>
-    static auto select_agent_policy(int)
+    _CCCL_API static auto select_agent_policy(int)
       -> AgentHistogramPolicy<Tuning::threads,
                               Tuning::items,
                               Tuning::load_algorithm,
@@ -195,7 +195,7 @@ struct policy_hub
                               Tuning::work_stealing>;
 
     template <typename Tuning>
-    static auto select_agent_policy(long) -> typename Policy500::AgentHistogramPolicyT;
+    _CCCL_API static auto select_agent_policy(long) -> typename Policy500::AgentHistogramPolicyT;
 
     using AgentHistogramPolicyT =
       decltype(select_agent_policy<
@@ -208,7 +208,7 @@ struct policy_hub
   {
     // Use values from tuning if a specialization exists, otherwise pick Policy900
     template <typename Tuning>
-    static auto select_agent_policy(int)
+    _CCCL_API static auto select_agent_policy(int)
       -> AgentHistogramPolicy<Tuning::threads,
                               Tuning::items,
                               Tuning::load_algorithm,
@@ -219,7 +219,7 @@ struct policy_hub
                               Tuning::vec_size>;
 
     template <typename Tuning>
-    static auto select_agent_policy(long) -> typename Policy900::AgentHistogramPolicyT;
+    _CCCL_API static auto select_agent_policy(long) -> typename Policy900::AgentHistogramPolicyT;
 
     using AgentHistogramPolicyT =
       decltype(select_agent_policy<


cub/cub/device/dispatch/kernels/kernel_histogram.cuh
```diff
@@ -329,21 +333,24 @@ struct Transforms
...
 
   // we trigger the sweep kernel only if we have a small number of remaining writes in this kernel
-  NV_IF_TARGET(NV_PROVIDES_SM_90,
-               (if (::cuda::std::reduce(num_output_bins_wrapper.begin(), num_output_bins_wrapper.end())
-                    <= ChainedPolicyT::ActivePolicy::pdl_trigger_next_launch_in_init_kernel_max_bin_count) {
-                 _CCCL_PDL_TRIGGER_NEXT_LAUNCH();
-               }));
+  NV_IF_TARGET(
+    NV_PROVIDES_SM_90,
+    (if (::cuda::std::reduce(num_output_bins_wrapper.begin(), num_output_bins_wrapper.end())
```
miscco, line 352: Technically this should be 
```suggestion
    (if (::cuda::std::accumulate(num_output_bins_wrapper.begin(), num_output_bins_wrapper.end())
```

Because it is meant to be run serially

cub/cub/device/dispatch/tuning/tuning_histogram.cuh
```diff
@@ -263,6 +231,148 @@ struct policy_hub
...
+  bool sample_is_primitive;
+  int sample_size;
+  int counter_size;
+  int sample_size_bytes;
+  int num_channels;
+  int num_active_channels;
+  bool is_even;
+
+private:
+  [[nodiscard]] _CCCL_API constexpr int t_scale(int nominal_items_per_thread) const
```
miscco, line 292: Can we have a better name?

cub/cub/device/dispatch/tuning/tuning_histogram.cuh
```diff
@@ -263,6 +231,148 @@ struct policy_hub
...
+  [[nodiscard]] _CCCL_API constexpr auto fallback_policy() const -> histogram_policy
+  {
+    return histogram_policy{384, t_scale(16), BLOCK_LOAD_DIRECT, LOAD_LDG, true, SMEM, false, 4, 0};
+  }
+
+public:
+  [[nodiscard]] _CCCL_API constexpr auto operator()(::cuda::arch_id arch) const -> histogram_policy
+  {
+    auto policy = fallback_policy();
+    if (arch >= ::cuda::arch_id::sm_90)
```
miscco: Nitpick: This goes contrary to the other policy selectors where we start with the newest policy and then move down.

Its also a bit confusing because the policy is not returned, so one has to look at the fields that are overwritten below.
Can we just keep the common order?

cub/cub/device/dispatch/dispatch_histogram.cuh
```diff
@@ -157,250 +157,237 @@ struct DeviceHistogramKernelSource
...
+  auto minus_one = ::cuda::proclaim_return_type<int>([](int levels) {
+    return levels - 1;
+  });
+  ::cuda::std::transform(
+    num_privatized_levels.begin(), num_privatized_levels.end(), num_privatized_bins_wrapper.begin(), minus_one);
+  ::cuda::std::transform(num_output_levels.begin(), num_output_levels.end(), num_output_bins_wrapper.begin(), minus_one);
 
-      int histogram_init_grid_dims =
-        (max_num_output_bins + histogram_init_block_threads - 1) / histogram_init_block_threads;
+  constexpr int histogram_init_block_threads = 256;
```
miscco, line 319: Question: Should this be something from the policy?

cub/cub/device/dispatch/dispatch_histogram.cuh
```diff
@@ -728,232 +681,371 @@ CUB_RUNTIME_FUNCTION _CCCL_FORCEINLINE static cudaError_t __dispatch_even_device
...
-    KernelLauncherFactory launcher_factory = {},
-    MaxPolicyT max_policy                  = {})
+    ::cuda::std::array<int, NUM_ACTIVE_CHANNELS> num_privatized_levels;
+    ::cuda::std::array<PrivatizedDecodeOpT, NUM_ACTIVE_CHANNELS> privatized_decode_op{};
+    ::cuda::std::array<OutputDecodeOpT, NUM_ACTIVE_CHANNELS> output_decode_op{};
+    int max_levels = num_output_levels[0];
+
+    for (int channel = 0; channel < NUM_ACTIVE_CHANNELS; ++channel)
+    {
+      num_privatized_levels[channel] = 257;
```
miscco, line 834: Is there a correlation to `PRIVATIZED_SMEM_BINS `?

cub/cub/device/dispatch/kernels/kernel_histogram.cuh
bernhardmgruber: Why? I just want to reduce a bunch of `int`s, so there is no need for that reduction to be performed left-to-right to preserve associativity.

cub/cub/device/dispatch/tuning/tuning_histogram.cuh
bernhardmgruber: The function was called `t_scale` in the old `policy_hub` and I frankly don't know what its meaning is. @gevtushenko moved it around in e3fba758, so the name seems to have been around for a while :S

cub/cub/device/dispatch/tuning/tuning_histogram.cuh
bernhardmgruber: ok, the state in this PR makes absolutely no sense. I wonder why it does not cause any SASS diffs then.

bernhardmgruber: ok, it should have worked, but it's very confusing. Fixed.

cub/cub/device/dispatch/dispatch_histogram.cuh
bernhardmgruber: Not at all, it's just a hardcoded parameter of the init kernel.

cub/cub/device/dispatch/dispatch_histogram.cuh
bernhardmgruber: I guess it's `PRIVATIZED_SMEM_BINS` (256) ` + 1`, since the number of levels is usually one higher than the number of bins. I can/should address this in a follow-up refactoring.

cub/cub/device/dispatch/dispatch_histogram.cuh
```diff
@@ -728,232 +681,371 @@ CUB_RUNTIME_FUNCTION _CCCL_FORCEINLINE static cudaError_t __dispatch_even_device
...
+      return cudaSuccess;
+    }
+  };
+
+public:
+  [[nodiscard]] _CCCL_API constexpr auto operator()(::cuda::arch_id arch_id) const -> histogram_policy
+  {
+    NV_IF_ELSE_TARGET(NV_IS_HOST,
+                      ({
+                        const int ptx_version = static_cast<int>(arch_id);
```
NaderAlAwar: Critical: this should be `static_cast<int>(arch_id) * 10`, like scan. `MaxPolicy::Invoke` is keyed on PTX versions (500/900/1000), not raw arch ids (50/90/100).

bernhardmgruber: Thanks a lot, that is definitely a bug!

PR #8213, opened 2026-03-30
Unconditionally provide `optional<T&>`

miscco:
It has been adopted in C++26, so we should unconditionally backport it.

While we are at it clean up the constructor constraints to make them a bit more readable


bernhardmgruber:
I vaguely remember there was a late fix to `optional<T&>` in LWG before it was voted into the standard. @miscco are we sure that our implementation corresponds to the latest working draft?

miscco:
> I vaguely remember there was a late fix to `optional<T&>` in LWG before it was voted into the standard. @miscco are we sure that our implementation corresponds to the latest working draft?

I believe that was related to assignments, but I checked the latest https://eel.is/c++draft/optional.optional.ref and that looks good

libcudacxx/include/cuda/std/version
```diff
@@ -85,9 +85,7 @@
 #define __cccl_lib_mdspan         202207L
 #define __cccl_lib_null_iterators 201304L
 #define __cccl_lib_optional       202110L
-#ifdef CCCL_ENABLE_OPTIONAL_REF
-#  define __cccl_lib_optional_ref 202602L
-#endif // CCCL_ENABLE_OPTIONAL_REF
+#define __cccl_lib_optional_ref   202602L
```
bernhardmgruber: I can't find the version macro value anywhere. The paper says `20XXXXL`. How do we get to the value `202602L` here?

miscco: That actually might have been a type. The feature test macro is supposed the mark the date when the feature was voted in. I need to find when that happened

miscco: The paper was accepted in June 2025 https://github.com/cplusplus/papers/issues/1661

libcudacxx/include/cuda/std/version
```diff
@@ -84,10 +84,7 @@
 // # define __cccl_lib_make_unique                          201304L
 #define __cccl_lib_mdspan         202207L
 #define __cccl_lib_null_iterators 201304L
-#define __cccl_lib_optional       202110L
-#ifdef CCCL_ENABLE_OPTIONAL_REF
-#  define __cccl_lib_optional_ref 202602L
-#endif // CCCL_ENABLE_OPTIONAL_REF
+#define __cccl_lib_optional       202506L
```
davebayer, line 87: There is also `__cpp_lib_freestanding_optional` being defined in the paper. And I believe our implementation is freestanding

davebayer (approved):

libcudacxx/include/cuda/std/version
miscco: I believe we cannot mark that as currently we rely on host <memory>

bernhardmgruber (approved):
cmake part is fine

PR #8218, opened 2026-03-30
Make pedantic warnings opt-in in ci/build_X scripts

### LINKED ISSUE #8217: Default behavior of `ci/build_X` scripts should not surface warnings nor cause warnings as errors
CCCL headers are implicitly treated as system headers which means users will never see warnings emitted from CCCL code.

However, the default behavior of our build/test scripts is to disable the system header annotation (https://github.com/NVIDIA/cccl/blob/84d100d3b35e364e208cc8811d1d5ff351ad55b4/libcudacxx/test/utils/libcudacxx/test/config.py#L1436).

We do this for our own CI testing so we can see and resolve any helpful warnings the compiler emits to improve our code.

However, this being the default behavior leads to internal bug reports of warnings promoted to errors that cause a lot of noise.

This issue can be closed by updating the build/test scripts and configuration such that invoking `build_X` does not emit warnings nor treat them as errors.

alliepiper:
Add CCCL_ENABLE_PRAGMA_SYSTEM_HEADER CMake option and --pedantic flag to ci/build_common.sh. Local builds now default to warnings suppressed and werror off. CI (GITHUB_ACTIONS) auto-enables pedantic mode. Direct cmake preset invocations retain existing defaults (werror on, system header pragma off). Lit test plumbing propagates the setting to Python config.

Fixes #8217


alliepiper:
pre-commit.ci autofix

PR #8222, opened 2026-03-31
Small warpspeed scan improvements

bernhardmgruber:
* Improve the diagnostic when we fail to fit into 24KiB SMEM
* Provide a better error message when uninitialized data is passed to the scan operator

### DIFF
--- a/cub/cub/device/dispatch/dispatch_scan.cuh
+++ b/cub/cub/device/dispatch/dispatch_scan.cuh
@@ -510,8 +510,8 @@ struct DispatchScan
   template <int SMemSizeForSingleStage>
   CUB_RUNTIME_FUNCTION static void __check_smem()
   {
-    CUB_DETAIL_STATIC_ISH_ASSERT(SMemSizeForSingleStage <= detail::max_smem_per_block,
-                                 "Single-stage warpspeed scan exceeds architecture independent SMEM (48KiB)");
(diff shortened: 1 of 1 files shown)


thrust/testing/scan.cu
```diff
@@ -734,7 +734,9 @@ struct composition_op_t
     permutation_t result;
     for (std::size_t i = 0; i < lhs.size(); i++)
     {
-      result[i] = rhs[lhs[i]];
+      const int sub = lhs[i];
+      _CCCL_ASSERT(sub >= 0 && sub < 5, "Permutation index out of range. Reading invalid data?");
+      result[i] = rhs[sub];
```
bernhardmgruber, line 739: This check is currently failing spuriously in the CI, because `rhs[lhs[i]]` reads out of bounds here sometimes. This change makes the failure deterministic and explicit.

cub/cub/device/dispatch/dispatch_scan.cuh
```diff
@@ -506,6 +506,14 @@ struct DispatchScan
   }
 
 #if __cccl_ptx_isa >= 860
+  // do check in separate function, so error message contains the required SMEM in the error novel
```
miscco, line 509: I knew C++ error messages are long, but that they classify as novels is something ^^

PR #8223, opened 2026-03-31
Don't specify `__launch_bounds__` together with  `__block_size__`

davebayer:
With nvcc before 13.1 it leads to ptxas errors as:
```bash
Conflicting directives: .maxntid and .reqntid cannot both be specified
```


davebayer:
> Question: Should this be conditional on the NVCC version?

No, specifying both doesn't make sense, nvcc 13.1+ just doesn't emit `.maxntid` in that case

PR #8224, opened 2026-03-31
[STF] move host_launch_deps from stf::reserved to stf namespace

caugonnet:
host_launch_deps is used directly in user-facing lambda signatures (e.g. scope->*[](host_launch_deps& deps) { ... }), so it should not be hidden in the reserved namespace.

## Description

closes 

## Checklist


andralex (approved):
good!

PR #8230, opened 2026-03-31
Fix STF compilation with `nvc++` as host compiler

davebayer:
Using `__NVCOMPILER` caused some serious issues with extended lambdas in this case..


caugonnet:
@davebayer how can we trust the CI if nvhpc is not enabled ?

davebayer:
> @davebayer how can we trust the CI if nvhpc is not enabled ?

I've tested it locally

davebayer:
> This looks good, but it's untested in the current CI afaik ?

I'm trying to finally get nvhpc back to the CI, so I need to patch these things 😅

caugonnet:
Sounds great, anyway at some point we must use faith :)

cudax/include/cuda/experimental/__stf/internal/launch.cuh
```diff
@@ -181,16 +181,16 @@ public:
   template <typename Fun>
   void operator->*(Fun&& f)
   {
-#  if __NVCOMPILER
+#  if _CCCL_CUDA_COMPILER(NVHPC)
```
miscco, line 184: I believe this is the wrong check as `__NVCOMPILER` checks only whether nvc++ is the host compiler
```suggestion
#  if _CCCL_COMPILER(NVHPC)
```

davebayer: And that's the problem that I'm fixing 😅

davebayer: Because the condition should only be true for NVC++ in CUDA mode

caugonnet: this can be host code too, not just device, if we launch on the host

miscco: But then shouldnt the title say `cuda compiler` ?

davebayer: > this can be host code too, not just device, if we launch on the host

But that's fine. The thing is that `nvc++` doesn't implement any of the `__nv_is_extended_meow_lambda_closure_type` traits, because it doesn't need them.

> But then shouldnt the title say `cuda compiler`?

No, because the problem occurs only when compiling CUDA source file with `nvcc` and `nvc++` as the host compiler. Then `__NVCOMPILER` is defined and we suppose that all lambdas are host/device, which is not true.

PR #8236, opened 2026-03-31
Prevent ADL from finding user-namespace functions in `kernel_scan_warp`

### LINKED ISSUE #8233: [BUG]: Unqualified `warpReduce` in `warpspeed` scan kernel causes ADL collision
### Is this a duplicate?

### Type of Bug

Compile-time Error

### Component

CUB

### Describe the bug

The warpspeed scan kernel (`kernel_scan_warpspeed.cuh`) [calls `warpReduce()`](https://github.com/NVIDIA/cccl/blob/060f9eb73a5920cbbc8e67872d6a8efe05ed6eab/cub/cub/device/dispatch/kernels/kernel_scan_warpspeed.cuh#L455) unqualified. When `ScanOpT` is a type from an external namespace that also defines a `warpReduce` function template, ADL pulls that function into the overload set and causes an ambiguity error.

This affects any downstream code that passes a custom scan operator to `thrust::inclusive_scan` / `cub::DeviceScan` when compiling for `sm_100+` (where the `warpspeed` kernel is selected).

Fix: qualify the call as `::cub::detail::scan::warpReduce` to prevent ADL.

### How to Reproduce

Minimal reproducer: pass a scan operator from a namespace that also contains a two-argument `warpReduce` function template. In practice, this is triggered by RAFT's [raft::add_op](https://github.com/rapidsai/raft/blob/eb74b0e6d23b59fb7803ff063b5cf1ee39038b34/cpp/include/raft/core/operators.hpp#L106) + [raft::warpReduce](https://github.com/rapidsai/raft/blob/eb74b0e6d23b59fb7803ff063b5cf1ee39038b34/cpp/include/raft/util/reduction.cuh#L49).

See https://github.com/rapidsai/cuvs/issues/1954.

### Expected behavior

Successful compilation.

### Reproduction link

_No response_

### Operating System

_No response_

### nvidia-smi output

_No response_

### NVCC version

_No response_

viclafargue:
Closes #8233.

### DIFF
--- a/cub/cub/device/dispatch/kernels/kernel_scan_warpspeed.cuh
+++ b/cub/cub/device/dispatch/kernels/kernel_scan_warpspeed.cuh
@@ -44,6 +44,7 @@ CUB_NAMESPACE_BEGIN
 
 namespace detail::scan
 {
+namespace __cub_detail  = CUB_NS_QUALIFIER::detail;
 namespace __scan_detail = CUB_NS_QUALIFIER::detail::scan;
 _CCCL_API constexpr warpspeed::SquadDesc squad_reduce(const scan_warpspeed_policy& policy)
 {
@@ -419,7 +420,7 @@ _CCCL_DEVICE_API _CCCL_FORCEINLINE void kernelBody(
         if (is_last_tile)
         {
           // TODO(bgruber): for operators where we know the identity we can probably optimize further here
-          regThreadSum = CUB_NS_QUALIFIER::detail::ThreadReducePartial(regInput, scan_op, valid_items_this_thread);
+          regThreadSum = __cub_detail::ThreadReducePartial(regInput, scan_op, valid_items_this_thread);
           regWarpSum   = __scan_detail::warpReducePartial(regThreadSum, scan_op, valid_threads_this_warp);
         }
         else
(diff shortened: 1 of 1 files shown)


cub/cub/device/dispatch/kernels/kernel_scan_warpspeed.cuh
```diff
@@ -419,12 +419,13 @@ _CCCL_DEVICE_API _CCCL_FORCEINLINE void kernelBody(
         {
           // TODO(bgruber): for operators where we know the identity we can probably optimize further here
           regThreadSum = ThreadReducePartial(regInput, scan_op, valid_items_this_thread);
-          regWarpSum   = warpReducePartial(regThreadSum, scan_op, valid_threads_this_warp);
+          regWarpSum =
+            CUB_NS_QUALIFIER::detail::scan::warpReducePartial(regThreadSum, scan_op, valid_threads_this_warp);
```
miscco: Important: I believe this is not sufficient, because we would still fail if the user ovevrloaded `ThreadReducePartial`

My suggestion would be to do `namespace __scan_detail = CUB_NS_QUALIFIER::detail::scan;` And then use `::__scan_detail::` throughout the file.

If that is too much work, I would be fine with merging this as a temporary fix to unblock you and then clean it up in a followup

viclafargue: Just added some protections for other ADL-related issues.

miscco:
/ok to test [f8516a4](https://github.com/NVIDIA/cccl/pull/8236/commits/f8516a4355c79e98f29b654cb2e4f3048900e4e4)

viclafargue:
Could someone trigger the CI again?

miscco:
/ok to test [c37e7cf](https://github.com/NVIDIA/cccl/pull/8236/commits/c37e7cf8bc7c18a51873f20aaec42c2c6156ed9f)

miscco:
Thanks a lot for reporting and fixing this ADL bug 🎉

Issue #8240, opened 2026-03-31
Add convenience functions for make_buffer to provide a shorthand for usage of the default mempools


pciolkosz:
The following creates a buffer:
```
auto buffer = cuda::make_buffer(stream, cuda::device_default_memory_pool(dev), 42, 0);
```
As a shorthand we should provide:
```
auto buffer = cuda::make_device_buffer(stream, dev, 42, 0);
```
this way usage of default pools becomes more convenient and seems like the default option for the users

miscco:
Question: how about `cuda::make_device_buffer(stream, 42, 0);` because the device is implicit in the stream

PR #8250, opened 2026-03-31
[libcu++] Add make_device_buffer, make_pinned_buffer, and make_managed_buffer

pciolkosz:
This PR adds `make_device/managed/pinned_buffer` as a shorthand for `make_buffer` with `device/managed/pinned_default_memory_pool`.

Also fly-by fix to add `_CCCL_HOST_API` to `make_buffer`.

Closes https://github.com/NVIDIA/cccl/issues/8240

### DIFF
--- a/libcudacxx/include/cuda/__container/make_buffer_with_pool.h
+++ b/libcudacxx/include/cuda/__container/make_buffer_with_pool.h
@@ -4,7 +4,7 @@
 // under the Apache License v2.0 with LLVM Exceptions.
 // See https://llvm.org/LICENSE.txt for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
-// SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES.
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES.
 //
 //===----------------------------------------------------------------------===//
 
--- a/libcudacxx/include/cuda/__container/buffer.h
+++ b/libcudacxx/include/cuda/__container/buffer.h
@@ -4,7 +4,7 @@
 // under the Apache License v2.0 with LLVM Exceptions.
 // See https://llvm.org/LICENSE.txt for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
-// SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES.
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES.
 //
 //===----------------------------------------------------------------------===//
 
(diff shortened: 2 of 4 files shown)


libcudacxx/include/cuda/__container/make_buffer_with_pool.h
```diff
@@ -0,0 +1,261 @@
...
+
+// ============================================================================
+// make_device_buffer — convenience wrappers using device_default_memory_pool
+// ============================================================================
+
+//! @brief Creates an empty buffer backed by the default device memory pool.
+//! @param __stream The stream used for allocation.
+//! @param __device The device whose default memory pool will be used.
+template <class _Tp, class _Env = ::cuda::std::execution::env<>>
+_CCCL_HOST_API auto make_device_buffer(stream_ref __stream, device_ref __device, const _Env& __env = {})
```
miscco: I believe there was another PR with a similar question: Is the `device` argument obolete given that stream and device must match?

If so should we verify that

libcudacxx/include/cuda/__container/make_buffer_with_pool.h
```diff
@@ -0,0 +1,261 @@
...
+make_pinned_buffer(stream_ref __stream, ::cuda::std::initializer_list<_Tp> __ilist, const _Env& __env = {})
+{
+  return ::cuda::make_buffer<_Tp>(__stream, ::cuda::pinned_default_memory_pool(), __ilist, __env);
+}
+
+//! @brief Creates a buffer from \p __range, backed by the default pinned memory pool.
+//! @param __stream The stream used for allocation and copy.
+//! @param __range The input range to be copied into the buffer.
+template <class _Tp, class _Range, class _Env = ::cuda::std::execution::env<>>
+_CCCL_HOST_API auto make_pinned_buffer(stream_ref __stream, _Range&& __range, const _Env& __env = {})
```
miscco: Critical: None of those APIs is constrained. We *really* need to add some constraints to disambiguate the different overloads

pciolkosz: I wanted to avoid repeating all the constraints, but you are right they could be ambiguous. I reworked these wrappers to just blindly forward arguments to `make_buffer`, the error message will be marginally worse, but at least we don't need to list all those overloads.
This approach doesn't work with the `initializer_list` overload, but I am questioning if its needed at all and maybe we should just deprecate it

libcudacxx/include/cuda/__container/make_buffer_with_pool.h
pciolkosz: The stream and device doesn't need to match. Stream is used for sequencing and device argument selects the device where the memory should reside.

We could provide another overload where the device is taken from the stream, it might be a good idea. I was a bit hesitant, because the pinned and managed variants only take a stream and the device variant not taking an explicit device could make it seem like the pinned and managed variants also are localized to the stream's device, where its not the case. I am 50/50 on this right now, but we can always add another overload later

PR #8251, opened 2026-03-31
`cuda::std::simd` core functionalities

fbusato:
Address https://github.com/NVIDIA/cccl/issues/30

## Description

Implement the core functionalities of C++26 `std::simd`, reference https://eel.is/c++draft/simd

The PR introduces the minimal set of functionalities that are self-consistent.
The main goal is to provide a C++26-compliant implementation, no specific optimizations have been introduced.

List of features:

- `cuda::std::simd` namespace and main header.
- ABI tags: `simd_abi::fixed_size_simple`, `simd_abi::native`,
- Type traits: `alignment`, `rebind`, `resize`.
- Exposition-only concepts and utilities.
- Flag types: `flags`, `flag_default`, `flag_convert`, `flag_aligned`, `flag_overaligned`.
- `basic_vec` and `basic_mask` classes.
- `fixed_size_simple` (custom ABI) specialization.

To do:

- [x] unit test.
- [ ] documentation.

PR #8252, opened 2026-03-31
`cuda::std::simd` load and store functionalities

fbusato:
## Description

Follow up PR of https://github.com/NVIDIA/cccl/pull/8251

- Add `partial_load`, `unchecked_load`, `partial_store`, `unchecked_store` free functions to `cuda::std::simd`.
- All overloads are provided: range, iterator+count, iterator+sentinel, each with and without mask.

To do:

- [x]  unit test.

### DIFF
--- a/libcudacxx/include/cuda/std/__simd/load.h
+++ b/libcudacxx/include/cuda/std/__simd/load.h
@@ -1,0 +1,355 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of libcu++ in the CUDA C++ Core Libraries,
+// under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef _CUDA_STD___SIMD_LOAD_H
+#define _CUDA_STD___SIMD_LOAD_H
+
+#include <cuda/std/detail/__config>
+
+#if defined(_CCCL_IMPLICIT_SYSTEM_HEADER_GCC)
+#  pragma GCC system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_CLANG)
+#  pragma clang system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_MSVC)
+#  pragma system_header
+#endif // no system header
+
+#include <cuda/__memory/is_valid_alignment.h>
+#include <cuda/__memory/ptr_rebind.h>
+#include <cuda/std/__algorithm/max.h>
+#include <cuda/std/__concepts/concept_macros.h>
+#include <cuda/std/__concepts/same_as.h>
+#include <cuda/std/__cstring/memcpy.h>
+#include <cuda/std/__iterator/concepts.h>
+#include <cuda/std/__iterator/distance.h>
+#include <cuda/std/__iterator/incrementable_traits.h>
+#include <cuda/std/__iterator/readable_traits.h>
+#include <cuda/std/__memory/assume_aligned.h>
+#include <cuda/std/__memory/pointer_traits.h>
+#include <cuda/std/__ranges/concepts.h>
+#include <cuda/std/__ranges/data.h>
+#include <cuda/std/__ranges/size.h>
+#include <cuda/std/__simd/basic_vec.h>
+#include <cuda/std/__simd/concepts.h>
+#include <cuda/std/__simd/flag.h>
+#include <cuda/std/__simd/utility.h>
+#include <cuda/std/__type_traits/remove_cvref.h>
+#include <cuda/std/__utility/cmp.h>
+#include <cuda/std/__utility/forward.h>
+
+#include <cuda/std/__cccl/prologue.h>
+
+_CCCL_BEGIN_NAMESPACE_CUDA_STD_SIMD
+
+// [simd.loadstore] helper: resolves default V template parameter for load functions
+// When _Vp = void (default), resolves to basic_vec<_Up>; otherwise uses the explicit _Vp
+template <typename _Vp, typename _Up>
+struct __load_vec_type
+{
+  using type = _Vp;
+};
+
+template <typename _Up>
+struct __load_vec_type<void, _Up>
+{
+  using type = basic_vec<_Up>;
+};
+
+template <typename _Vp, typename _Up>
+using __load_vec_t = typename __load_vec_type<_Vp, _Up>::type;
+
+template <typename _Result, typename _Up, typename... _Flags>
+_CCCL_HOST_DEVICE_API constexpr void
+__check_load_preconditions(const _Up* __ptr, flags<_Flags...>, const __simd_size_type __count = 1) noexcept
+{
+  using __value_t = typename _Result::value_type;
+  static_assert(same_as<remove_cvref_t<_Result>, _Result>, "V must not be a reference or cv-qualified type");
+
+  static_assert(__is_vectorizable_v<__value_t> && __is_enabled_abi_v<typename _Result::abi_type>,
+                "cuda::std::simd::load: V must be an enabled specialization of basic_vec");
+  static_assert(__is_vectorizable_v<_Up>, "range_value_t<R> must be a vectorizable type");
+
+  static_assert(__explicitly_convertible_to<_Up, __value_t>,
+                "cuda::std::simd::load: range_value_t<R> must satisfy explicitly-convertible-to<value_type>");
+
+  static_assert(__has_convert_flag_v<_Flags...> || __is_value_preserving_v<_Up, __value_t>,
+                "cuda::std::simd::load: Conversion from range_value_t<R> to value_type is not value-preserving; use "
+                "flag_convert");
+
+  _CCCL_ASSERT(__count == 0 || __ptr != nullptr, "cuda::std::simd::load: range data is nullptr");
+  ::cuda::std::simd::__assert_load_store_alignment<_Result, _Up, _Flags...>(__ptr);
+}
+
+// [simd.loadstore] helper: core partial load from pointer + count + mask
+template <typename _Result, typename _Up, typename... _Flags>
+[[nodiscard]] _CCCL_HOST_DEVICE_API constexpr _Result __partial_load_from_ptr(
+  const _Up* __ptr,
+  const __simd_size_type __count,
+  const typename _Result::mask_type& __mask,
+  flags<_Flags...> __flags = {}) noexcept
+{
+  using __value_t = typename _Result::value_type;
+  ::cuda::std::simd::__check_load_preconditions<_Result>(__ptr, __flags, __count);
+  constexpr auto __simd_size = _Result::__size;
+
+  _Result __result;
+  _CCCL_PRAGMA_UNROLL_FULL()
+  for (__simd_size_type __i = 0; __i < __simd_size; ++__i)
+  {
+    const auto __value = (__mask[__i] && __i < __count) ? static_cast<__value_t>(__ptr[__i]) : __value_t{};
+    __result.__set(__i, __value);
+  }
+  return __result;
+}
+
+template <typename _Result, typename _Up, typename... _Flags>
+[[nodiscard]] _CCCL_HOST_DEVICE_API constexpr _Result
+__full_load_from_ptr(const _Up* __ptr, const typename _Result::mask_type& __mask, flags<_Flags...> __flags) noexcept
+{
+  ::cuda::std::simd::__check_load_preconditions<_Result>(__ptr, __flags);
+  constexpr bool __has_aligned_flag = __has_aligned_flag_v<_Flags...>;
+
+  if constexpr (__has_aligned_flag || __has_overaligned_flag_v<_Flags...>)
+  {
+    _CCCL_IF_NOT_CONSTEVAL_DEFAULT
+    {
+      // minimum condition for pointer alignment
+      constexpr auto __base_alignment = __has_aligned_flag ? alignment_v<_Result, _Up> : alignof(_Up);
+      constexpr auto __ptr_alignment  = ::cuda::std::max(__base_alignment, __overaligned_value_v<_Flags...>);
+      constexpr auto __simd_size      = _Result::__size;
+      constexpr auto __data_size      = __simd_size * sizeof(_Up);
+
+      _Up __tmp[__simd_size]{};
+      // vectorized load from pointer
+      if constexpr (__is_cuda_vectorizable_v<_Up> && __simd_size > 1 && __ptr_alignment >= __data_size
+                    && ::cuda::__is_valid_alignment(__data_size))
+      {
+        struct alignas(__data_size) __aligned_t
+        {
+          _Up __data[__simd_size];
+        };
+        // nvcc performance bug: memcpy from pointer could not be vectorized
+        const auto __aligned_ptr = ::cuda::ptr_rebind<__aligned_t>(__ptr);
+        const auto __data        = *::cuda::std::assume_aligned<__ptr_alignment>(__aligned_ptr);
+        ::cuda::std::memcpy(&__tmp, &__data, sizeof(__tmp));
+      }
+      // rely on compiler vectorization
+      else
+      {
+        const auto __aligned_ptr = ::cuda::std::assume_aligned<__ptr_alignment>(__ptr);
+        _CCCL_PRAGMA_UNROLL_FULL()
+        for (__simd_size_type __i = 0; __i < __simd_size; ++__i)
+        {
+          __tmp[__i] = __aligned_ptr[__i];
+        }
+      }
+      using __value_t = typename _Result::value_type;
+      _Result __result;
+      _CCCL_PRAGMA_UNROLL_FULL()
+      for (__simd_size_type __i = 0; __i < __simd_size; ++__i)
+      {
+        const auto __value = (!__mask[__i]) ? __value_t{} : static_cast<__value_t>(__tmp[__i]);
+        __result.__set(__i, __value);
+      }
+      return __result;
+    }
+  }
+  return ::cuda::std::simd::__partial_load_from_ptr<_Result>(__ptr, _Result::__size, __mask, __flags);
+}
+
+//----------------------------------------------------------------------------------------------------------------------
+// [simd.loadstore] partial_load
+
+// partial_load: range, masked
+_CCCL_TEMPLATE(typename _Vp = void, typename _Range, typename... _Flags)
+_CCCL_REQUIRES(::cuda::std::ranges::contiguous_range<_Range> _CCCL_AND ::cuda::std::ranges::sized_range<_Range>)
+[[nodiscard]] _CCCL_HOST_DEVICE_API constexpr __load_vec_t<_Vp, ::cuda::std::ranges::range_value_t<_Range>>
+partial_load(_Range&& __r,
+             const typename __load_vec_t<_Vp, ::cuda::std::ranges::range_value_t<_Range>>::mask_type& __mask,
+             flags<_Flags...> __f = {})
+{
+  using __result_t        = __load_vec_t<_Vp, ::cuda::std::ranges::range_value_t<_Range>>;
+  const auto __range_size = ::cuda::std::ranges::size(__r);
+  _CCCL_ASSERT(::cuda::std::in_range<__simd_size_type>(__range_size),
+               "cuda::std::simd::partial_load: range size out of range");
+  const auto __size = static_cast<__simd_size_type>(__range_size);
+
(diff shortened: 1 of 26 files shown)


miscco:
@fbusato could you please squash on top of the first PR?

Its really difficult to review all of that with the base PR sitting on top and bringing in a bunch of changes

fbusato:
what about rebasing this PR with main after `cuda::std::simd` core functionalities #8251 is merged? I think starting reviewing this PR before the other one merged makes little sense

libcudacxx/include/cuda/std/__fwd/subrange.h
```diff
@@ -32,18 +32,24 @@ enum class _CCCL_TYPE_VISIBILITY_DEFAULT subrange_kind : bool
   sized
 };
 
+// nvcc+clang bug: ternary expression mixed with <> in a default template argument
+// "mismatched delimiters in default argument expression"
+template <class _Iter, class _Sent>
+inline constexpr subrange_kind __default_subrange_kind =
+  sized_sentinel_for<_Sent, _Iter> ? subrange_kind::sized : subrange_kind::unsized;
```
miscco: Critical: Please remove, this is not necessary anymore

libcudacxx/include/cuda/std/__simd/utility.h
```diff
@@ -150,9 +152,21 @@ _CCCL_API constexpr void __assert_load_store_alignment([[maybe_unused]] const _U
...
+    {
+      _CCCL_ASSERT(::cuda::is_aligned(__data, alignof(_Up)), "data is not aligned to alignof(_Up)");
+    }
   }
 }
 
+// TODO(fbusato): use ::cuda::is_trivially_copyable_v
+// used in load/store preconditions
+template <typename _TpIn>
+inline constexpr bool __is_cuda_vectoriazable_v =
```
NaderAlAwar: Nit: typo
```suggestion
inline constexpr bool __is_cuda_vectorizable_v =
```

libcudacxx/include/cuda/std/__simd/load.h
```diff
@@ -0,0 +1,356 @@
...
+      // EDG requires locals to be initialized for the function to remain constexpr
+#if __cpp_if_consteval >= 202106L || _CCCL_HAS_IF_CONSTEVAL_IN_CXX20()
+      _Up __tmp[__simd_size];
+#else
+      _Up __tmp[__simd_size] = {};
+#endif
+      // vectorized load from pointer
+      if constexpr (__is_cuda_vectoriazable_v<_Up> && __simd_size > 1 && __ptr_alignment >= __data_size)
+      {
+        struct alignas(__data_size) __aligned_t
```
NaderAlAwar, line 122: Important: `__data_size` is not always a valid alignment, so this can fail to compile. For example, `basic_vec<float, fixed_size<3>>` has `__simd_size == 3` and `sizeof(_Up) == 4`, so `__data_size == 12`, but `alignas(12)` is invalid.

libcudacxx/include/cuda/std/__simd/store.h
```diff
@@ -0,0 +1,361 @@
...
+#endif
+      _CCCL_PRAGMA_UNROLL_FULL()
+      for (__simd_size_type __i = 0; __i < __simd_size; ++__i)
+      {
+        __tmp[__i] = static_cast<_Up>(__v[__i]);
+      }
+      // vectorized store to pointer
+      if constexpr (__is_cuda_vectoriazable_v<_Up> && __simd_size > 1 && __ptr_alignment >= __data_size)
+      {
+        struct alignas(__data_size) __aligned_t
```
NaderAlAwar, line 116: Important: same comment as load

libcudacxx/include/cuda/std/__simd/load.h
```diff
@@ -0,0 +1,356 @@
...
+__full_load_from_ptr(const _Up* __ptr, const typename _Result::mask_type& __mask, flags<_Flags...> __flags) noexcept
+{
+  ::cuda::std::simd::__check_load_preconditions<_Result>(__ptr, __flags);
+
+  if constexpr (__has_aligned_flag_v<_Flags...> || __has_overaligned_flag_v<_Flags...>)
+  {
+    _CCCL_IF_NOT_CONSTEVAL_DEFAULT
+    {
+      constexpr auto __base_alignment = alignment_v<_Result, _Up>; // minimum condition for pointer alignment
+      constexpr auto __ptr_alignment  = ::cuda::std::max(__base_alignment, __overaligned_value_v<_Flags...>);
```
NaderAlAwar: Important: `alignment_v` returns the preferred alignment, not the minimum alignment like the comment suggests. At present, this can make `__ptr_alignment` is larger than the overaligned value specified by the user, causing the implementation to assume a stronger alignment than the user promised.

libcudacxx/include/cuda/std/__simd/store.h
```diff
@@ -0,0 +1,361 @@
...
+{
+  ::cuda::std::simd::__check_store_preconditions<_Tp, _Abi>(__ptr, __flags);
+  constexpr auto __simd_size = basic_vec<_Tp, _Abi>::__size;
+
+  if constexpr (__has_aligned_flag_v<_Flags...> || __has_overaligned_flag_v<_Flags...>)
+  {
+    _CCCL_IF_NOT_CONSTEVAL_DEFAULT
+    {
+      constexpr auto __base_alignment = alignment_v<basic_vec<_Tp, _Abi>, _Up>;
+      constexpr auto __ptr_alignment  = ::cuda::std::max(__base_alignment, __overaligned_value_v<_Flags...>);
```
NaderAlAwar: Important: same comment as load

libcudacxx/include/cuda/std/__simd/load.h
```diff
@@ -0,0 +1,356 @@
...
+  static_assert(__is_vectorizable_v<_Up>, "range_value_t<R> must be a vectorizable type");
+
+  static_assert(__explicitly_convertible_to<__value_t, _Up>,
+                "cuda::std::simd::load: range_value_t<R> must satisfy explicitly-convertible-to<value_type>");
+
+  static_assert(__has_convert_flag_v<_Flags...> || __is_value_preserving_v<_Up, __value_t>,
+                "cuda::std::simd::load: Conversion from range_value_t<R> to value_type is not value-preserving; use "
+                "flag_convert");
+
+  _CCCL_ASSERT(__ptr != nullptr, "cuda::std::simd::load: range data is nullptr");
```
NaderAlAwar: Question: should this permit `__ptr == nullptr` for `partial_load`/`partial_store` when `__count == 0`? The standard wording for partial load/store does not access `data()[i]` when `i < ranges::size(r)` is false.

libcudacxx/include/cuda/std/__simd/store.h
```diff
@@ -0,0 +1,361 @@
...
+_CCCL_TEMPLATE(typename _Tp, typename _Abi, typename _Range, typename... _Flags)
+_CCCL_REQUIRES(ranges::contiguous_range<_Range> _CCCL_AND ranges::sized_range<_Range> _CCCL_AND
+                 __explicitly_convertible_to<ranges::range_value_t<_Range>, _Tp>)
+_CCCL_API constexpr void partial_store(
+  const basic_vec<_Tp, _Abi>& __v,
+  _Range&& __r,
+  const typename basic_vec<_Tp, _Abi>::mask_type& __mask,
+  flags<_Flags...> __f = {})
+{
+  static_assert(indirectly_writable<ranges::iterator_t<_Range>, ranges::range_value_t<_Range>>,
```
NaderAlAwar: Important: this should be a constraint instead of a static assert, as specified by the standard

libcudacxx/include/cuda/std/__simd/store.h
```diff
@@ -0,0 +1,361 @@
...
+_CCCL_TEMPLATE(typename _Tp, typename _Abi, typename _Ip, typename... _Flags)
+_CCCL_REQUIRES(contiguous_iterator<_Ip> _CCCL_AND __explicitly_convertible_to<iter_value_t<_Ip>, _Tp>)
+_CCCL_API constexpr void partial_store(
+  const basic_vec<_Tp, _Abi>& __v,
+  const _Ip __first,
+  const iter_difference_t<_Ip> __n,
+  const typename basic_vec<_Tp, _Abi>::mask_type& __mask,
+  flags<_Flags...> __f = {})
+{
+  static_assert(indirectly_writable<_Ip, iter_value_t<_Ip>>,
```
NaderAlAwar: Important: same as above, should be a constraint

libcudacxx/include/cuda/std/__simd/store.h
```diff
@@ -0,0 +1,361 @@
...
+_CCCL_REQUIRES(contiguous_iterator<_Ip> _CCCL_AND sized_sentinel_for<_Sp, _Ip> _CCCL_AND
+                 __explicitly_convertible_to<iter_value_t<_Ip>, _Tp>)
+_CCCL_API constexpr void partial_store(
+  const basic_vec<_Tp, _Abi>& __v,
+  const _Ip __first,
+  const _Sp __last,
+  const typename basic_vec<_Tp, _Abi>::mask_type& __mask,
+  flags<_Flags...> __f = {})
+{
+  static_assert(indirectly_writable<_Ip, iter_value_t<_Ip>>,
```
NaderAlAwar: Important: same as above, should be a constraint

libcudacxx/include/cuda/std/__simd/load.h
```diff
@@ -0,0 +1,356 @@
...
+_CCCL_API constexpr void __check_load_preconditions(const _Up* __ptr, flags<_Flags...>) noexcept
+{
+  using __value_t = typename _Result::value_type;
+  static_assert(same_as<remove_cvref_t<_Result>, _Result>, "V must not be a reference or cv-qualified type");
+
+  static_assert(__is_vectorizable_v<__value_t> && __is_enabled_abi_v<typename _Result::abi_type>,
+                "cuda::std::simd::load: V must be an enabled specialization of basic_vec");
+  static_assert(__is_vectorizable_v<_Up>, "range_value_t<R> must be a vectorizable type");
+
+  static_assert(__explicitly_convertible_to<__value_t, _Up>,
```
NaderAlAwar: Important: this checks that `__value_t` is convertible to `_Up`, but it should be the other way around

libcudacxx/include/cuda/std/__simd/store.h
```diff
@@ -0,0 +1,361 @@
...
+#include <cuda/std/__cccl/prologue.h>
+
+_CCCL_BEGIN_NAMESPACE_CUDA_STD_SIMD
+
+template <typename _Tp, typename _Abi, typename _Up, typename... _Flags>
+_CCCL_API constexpr void __check_store_preconditions(_Up* const __ptr, flags<_Flags...>) noexcept
+{
+  static_assert(__is_vectorizable_v<_Up>, "cuda::std::simd::store: range_value_t<R> must be a vectorizable type");
+
+  static_assert(__explicitly_convertible_to<_Up, _Tp>,
```
NaderAlAwar: Important: this checks that `_Up` is convertible to `_Tp`, but it should be the other way around

libcudacxx/test/simd_codegen/CMakeLists.txt
```diff
@@ -0,0 +1,100 @@
+add_custom_target(libcudacxx.test.simd.ptx)
```
NaderAlAwar, line 11: Important: this should be added to `CMakePresets.json` to the libcudacxx build preset because at present it is not executed unless explicitly built

libcudacxx/test/simd_codegen/load_store_i8.cu
```diff
@@ -0,0 +1,37 @@
...
+
+/*
+
+; SM8X-LABEL: .visible .entry {{.*}}test_load_i8_8{{.*}}(
+; SM8X: {{.*}}ld.global.v4.u16{{.*}}
+; SM8X: {{.*}}st.global.v4.u16{{.*}}
+
+; SM10X-LABEL: .visible .entry {{.*}}test_load_i8_16{{.*}}(
+; SM10X: {{.*}}ld.global.v4.b32{{.*}}
+; SM10X: {{.*}}st.global.v4.b32{{.*}}
```
NaderAlAwar: Important: tested this locally and it emitted `u32` instead of `b32`. Should both be allowed here?

libcudacxx/include/cuda/std/__simd/load.h
fbusato: `__is_cuda_vectoriazable_v` protects against this case

fbusato: also, `alignas(12)` is not valid. The compiler doesn't allow it

fbusato: found the failing case. It happens when the flag/pointer are overaligned compared to the SIMD vector size

libcudacxx/include/cuda/std/__simd/load.h
fbusato: I don't have a strong opinion on that. We can support  `__count == 0`

libcudacxx/include/cuda/std/__simd/load.h
fbusato: `alignment_v` is implementation defined and depends on the ABI, so it is not problem. The second path probably can be improved to avoid potential UB

libcudacxx/test/simd_codegen/load_store_i8.cu
fbusato: weird. Probably difference nvcc versions generate slightly different code

libcudacxx/include/cuda/std/__simd/store.h
```diff
@@ -149,19 +152,23 @@ __full_store_to_ptr(const basic_vec<_Tp, _Abi>& __v, _Up* const __ptr, flags<_Fl
 //----------------------------------------------------------------------------------------------------------------------
 // [simd.loadstore] partial_store
 
+template <typename _Tp, typename _Range, bool = (ranges::contiguous_range<_Range> && ranges::sized_range<_Range>)>
+inline constexpr bool __partial_store_range_constraints_v = false;
+
+template <typename _Tp, typename _Range>
+inline constexpr bool __partial_store_range_constraints_v<_Tp, _Range, true> =
+  indirectly_writable<ranges::iterator_t<_Range>, ranges::range_value_t<_Range>> //
+  && __explicitly_convertible_to<_Tp, ranges::range_value_t<_Range>>;
```
miscco: Question: Why not define this as a `_CCCL_REQUIRES_EXPR` tjhat would give us short circuit behavior out f the box

libcudacxx/include/cuda/std/__simd/store.h
```diff
@@ -149,19 +152,23 @@ __full_store_to_ptr(const basic_vec<_Tp, _Abi>& __v, _Up* const __ptr, flags<_Fl
 //----------------------------------------------------------------------------------------------------------------------
 // [simd.loadstore] partial_store
 
+template <typename _Tp, typename _Range, bool = (ranges::contiguous_range<_Range> && ranges::sized_range<_Range>)>
```
miscco: Please always fully qualify

libcudacxx/include/cuda/std/__simd/store.h
NaderAlAwar: This still has the issue where the `alignment_v` could return a larger value than the alignment guaranteed by the user. This was fixed in load, but still needs to be fixed here

NaderAlAwar (approved):

libcudacxx/include/cuda/std/__simd/store.h
fbusato: it is more to reduce verbosity, but if we lose short circuiting I will add it back

libcudacxx/include/cuda/std/__simd/store.h
fbusato: wait, this is not clear to me. `ranges::contiguous_range` and `ranges::sized_range` are concepts, not functions, and we are in the `cuda::std::simd` namespace so it should be fine

libcudacxx/include/cuda/std/__fwd/subrange.h
fbusato: reverted

fbusato:
@NaderAlAwar @miscco I addressed the comment, but I also took the opportunity to relax the alignment (implementation defined). Please double-check the latest commit c583361

libcudacxx/include/cuda/std/__simd/store.h
miscco: Yes but a user could do `using namespace std` and the we get conflicts with `std::ranges::meow`

fbusato: I don't think so. For the namespace resolution rules, `ranges::contiguous_range` can only match `cuda::std::ranges` and not `std::ranges`. I also tried to reproduce it and I don't see problems https://godbolt.org/z/bs53MT3Mv. Please correct me if I'm wrong

miscco: I believe that this is too dangerous.

Currently we have the rule:
"Either do not qualify if its a type/trait in the same namespace or fully qualify"

This loosens that to 
"Either do not qualify if its a type/trait in the same namespace or its a type/trait in an enclosing namespace or fully qualify"

There is no hurt in fully qualifying and it makes validating expectations much easier

fbusato: ook, let me also add it to the SKILL

libcudacxx/include/cuda/std/__simd/load.h
```diff
@@ -0,0 +1,361 @@
...
+      constexpr auto __simd_size      = _Result::__size;
+      constexpr auto __data_size      = __simd_size * sizeof(_Up);
+
+      // When _CCCL_IF_NOT_CONSTEVAL falls back to __builtin_is_constant_evaluated(),
+      // EDG requires locals to be initialized for the function to remain constexpr
+#if __cpp_if_consteval >= 202106L || _CCCL_HAS_IF_CONSTEVAL_IN_CXX20()
+      _Up __tmp[__simd_size];
+#else
+      _Up __tmp[__simd_size] = {};
+#endif
```
miscco: Question: I am 99% sure the compiler just elides the initialization. We talked about this in the context of using fold expression and your investigation showed no codegen for initializing arrays

libcudacxx/include/cuda/std/__simd/load.h
```diff
@@ -0,0 +1,361 @@
...
+      // vectorized load from pointer
+      if constexpr (__is_cuda_vectorizable_v<_Up> && __simd_size > 1 && __ptr_alignment >= __data_size
+                    && ::cuda::__is_valid_alignment(__data_size))
+      {
+        struct alignas(__data_size) __aligned_t
+        {
+          char __data[__data_size];
+        };
```
miscco, line 125: We do have `__uninitialized_array` added recently, we should consider using it

libcudacxx/include/cuda/std/__simd/load.h
miscco: We could also just use `__uninitialized_array<_Up, __simd_size>` here

libcudacxx/include/cuda/std/__simd/store.h
```diff
@@ -0,0 +1,369 @@
...
+      constexpr auto __ptr_alignment  = ::cuda::std::max(__base_alignment, __overaligned_value_v<_Flags...>);
+      constexpr auto __data_size      = __simd_size * sizeof(_Up);
+
+      // When _CCCL_IF_NOT_CONSTEVAL falls back to __builtin_is_constant_evaluated(),
+      // EDG requires locals to be initialized for the function to remain constexpr
+#if __cpp_if_consteval >= 202106L || _CCCL_HAS_IF_CONSTEVAL_IN_CXX20()
+      _Up __tmp[__simd_size];
+#else
+      _Up __tmp[__simd_size] = {};
+#endif
```
miscco: Ditto `__uninitialized_array`

libcudacxx/include/cuda/std/__simd/load.h
fbusato: this is the only one that I'm not 100% sure, because `alignas` is in a difference position and nvcc could behave differently

fbusato:
`__uninitialized_array` didn't work for compile-time expressions. Reverted

libcudacxx/include/cuda/std/__simd/load.h
```diff
@@ -0,0 +1,355 @@
...
+};
+
+template <typename _Up>
+struct __load_vec_type<void, _Up>
+{
+  using type = basic_vec<_Up>;
+};
+
+template <typename _Vp, typename _Up>
+using __load_vec_t = typename __load_vec_type<_Vp, _Up>::type;
```
miscco: Nitpick: might be easier to read with
```suggestion
template <typename _Vp, typename _Up>
using __load_vec_t = conditional_t<is_void_v<_Vp>, basic_vec<_Up>, _Vp>;
```

miscco (approved):

PR #8253, opened 2026-03-31
`cuda::std::simd` Reduction functionalities

fbusato:
## Description

Follow up PR of https://github.com/NVIDIA/cccl/pull/8251

- Add `reduce`, `reduce_min`, `reduce_max` with/without mask.
- Mask reductions: `all_of`, `any_of`, `none_of`, `reduce_count`, `reduce_min_index`, `reduce_max_index` plus scalar bool overloads

To do:

- [x] Mask reductions are free functions. Probably the best solution is to moved them to basic_mask as friend functions.
- [x]  unit test.


libcudacxx/include/cuda/std/__simd/reductions.h
```diff
@@ -0,0 +1,352 @@
...
+_CCCL_BEGIN_NAMESPACE_CUDA_STD_SIMD
+// [simd.expos], reduction-binary-operation concept
+
+template <typename _BinaryOp, typename _Tp>
+_CCCL_CONCEPT __reduction_binary_operation = _CCCL_REQUIRES_EXPR(
+  (_BinaryOp, _Tp), const _BinaryOp __binary_op, const vec<_Tp, 1> __v)(_Same_as(vec<_Tp, 1>) __binary_op(__v, __v));
+
+template <typename _Tp, typename _BinaryOp>
+inline constexpr bool __is_nothrow_reduction_binary_operation_v = noexcept(
+  ::cuda::std::declval<const _BinaryOp&>()(::cuda::std::declval<vec<_Tp, 1>>(), ::cuda::std::declval<vec<_Tp, 1>>()));
```
miscco: Critical: This requires the expression to be valid. We need to check `__reduction_binary_operation` first. Also the template arguments are flipped compared to `__reduction_binary_operation `
```suggestion
template <typename _BinaryOp, typename _Tp, bool = __reduction_binary_operation<_BinaryOp, _Tp>>
inline constexpr bool __is_nothrow_reduction_binary_operation_v = false;

template <typename _BinaryOp, typename _Tp>
inline constexpr bool __is_nothrow_reduction_binary_operation_v<_BinaryOp, _Tp, true> = noexcept(
  ::cuda::std::declval<const _BinaryOp&>()(::cuda::std::declval<vec<_Tp, 1>>(), ::cuda::std::declval<vec<_Tp, 1>>()));
```

libcudacxx/include/cuda/std/__simd/reductions.h
```diff
@@ -0,0 +1,352 @@
...
+
+template <typename _Tp, typename _BinaryOp>
+inline constexpr bool __is_reduce_default_supported_operation_v =
+  is_same_v<_BinaryOp, plus<>> //
+  || is_same_v<_BinaryOp, multiplies<>> //
+  || is_same_v<_BinaryOp, bit_and<>> //
+  || is_same_v<_BinaryOp, bit_or<>> //
+  || is_same_v<_BinaryOp, bit_xor<>>;
```
miscco: Quesion: What about the specialized e.g `plus<_Tp>` Should those also be supported?

miscco: Followup: What about `::std::plus`

libcudacxx/include/cuda/std/__simd/reductions.h
```diff
@@ -0,0 +1,352 @@
...
+
+template <typename _Tp, typename _BinaryOp>
+[[nodiscard]] _CCCL_API constexpr _Tp __default_identity_element() noexcept
```
miscco, line 65: Dont we have those already at home?

libcudacxx/include/cuda/std/__simd/reductions.h
```diff
@@ -0,0 +1,352 @@
...
+  }
+}
+
+// [simd.reductions], reduce
+
+_CCCL_TEMPLATE(typename _Tp, typename _Abi, typename _BinaryOperation = plus<>)
+_CCCL_REQUIRES(__reduction_binary_operation<_BinaryOperation, _Tp>)
+[[nodiscard]] _CCCL_API constexpr _Tp reduce(
+  const basic_vec<_Tp, _Abi>& __x,
+  _BinaryOperation __binary_op = plus<>{}) noexcept(__is_nothrow_reduction_binary_operation_v<_Tp, _BinaryOperation>)
```
miscco: Critical: This wont default the `_BinaryOperation` if that is not `plus`
```suggestion
  _BinaryOperation __binary_op = {}) noexcept(__is_nothrow_reduction_binary_operation_v<_Tp, _BinaryOperation>)
```

miscco: Nitpick: In such cases where `clang-format` does such a terrible job, I prefer to force a newline by adding a `//` in front of the `noexcept`

libcudacxx/include/cuda/std/__simd/reductions.h
```diff
@@ -0,0 +1,352 @@
...
+{
+  vec<_Tp, 1> __result{__x[0]};
+  _CCCL_PRAGMA_UNROLL_FULL()
+  for (__simd_size_type __i = 1; __i < __x.__size; ++__i)
+  {
+    __result = __binary_op(__result, vec<_Tp, 1>{__x[__i]});
+  }
```
miscco, line 100: Important: This has come up quite often. I believe this should just be a fold over the compile time size

libcudacxx/include/cuda/std/__simd/reductions.h
```diff
@@ -0,0 +1,352 @@
...
+
+// We need two overloads:
+// 1) An argument for identity_element is provided for the invocation
+// 2) unless BinaryOperation is one of plus<>, multiplies<>, bit_and<>, bit_or<>, or bit_xor<>
+_CCCL_TEMPLATE(typename _Tp, typename _Abi, typename _BinaryOperation)
+_CCCL_REQUIRES(__reduction_binary_operation<_BinaryOperation, _Tp>)
+[[nodiscard]] _CCCL_API constexpr _Tp reduce(
+  const basic_vec<_Tp, _Abi>& __x,
+  const typename basic_vec<_Tp, _Abi>::mask_type& __mask,
+  _BinaryOperation __binary_op,
```
miscco: Important: This is missing the default for `__binary_op`

libcudacxx/include/cuda/std/__simd/reductions.h
```diff
@@ -0,0 +1,352 @@
...
+}
+
+_CCCL_TEMPLATE(typename _Tp, typename _Abi, typename _BinaryOperation = plus<>)
+_CCCL_REQUIRES(__reduction_binary_operation<_BinaryOperation, _Tp> _CCCL_AND
+                 __is_reduce_default_supported_operation_v<_Tp, _BinaryOperation>)
+[[nodiscard]] _CCCL_API constexpr _Tp
+reduce(const basic_vec<_Tp, _Abi>& __x,
+       const typename basic_vec<_Tp, _Abi>::mask_type& __mask,
+       const _BinaryOperation __binary_op =
+         plus<>{}) noexcept(__is_nothrow_reduction_binary_operation_v<_Tp, _BinaryOperation>)
```
miscco: Critical: wrong default
```suggestion
         {}) noexcept(__is_nothrow_reduction_binary_operation_v<_Tp, _BinaryOperation>)
```

libcudacxx/include/cuda/std/__simd/reductions.h
```diff
@@ -0,0 +1,352 @@
...
+{
+  return ::cuda::std::simd::reduce(
+    __x, __mask, __binary_op, ::cuda::std::simd::__default_identity_element<_Tp, _BinaryOperation>());
+}
+
```
miscco, line 139: Question: I am missing the reductions specified in https://eel.is/c%2B%2Bdraft/simd#reductions-10

Is this intentional?

libcudacxx/include/cuda/std/__simd/reductions.h
```diff
@@ -0,0 +1,352 @@
...
+{
+  auto __result = __x[0];
+  _CCCL_PRAGMA_UNROLL_FULL()
+  for (__simd_size_type __i = 1; __i < __x.__size; ++__i)
+  {
+    const auto __val = __x[__i];
+    if (!(__result < __val))
+    {
+      __result = __val;
+    }
```
miscco: This will overwrite __result in case of equality. Why isnt this just
```suggestion
    if (__val < __result))
    {
      __result = __val;
    }
```

libcudacxx/include/cuda/std/__simd/reductions.h
```diff
@@ -0,0 +1,352 @@
...
+_CCCL_TEMPLATE(typename _Tp, typename _Abi)
+_CCCL_REQUIRES(totally_ordered<_Tp>)
+[[nodiscard]] _CCCL_API constexpr _Tp reduce_max(const basic_vec<_Tp, _Abi>& __x) noexcept
+{
+  auto __result = __x[0];
+  _CCCL_PRAGMA_UNROLL_FULL()
+  for (__simd_size_type __i = 1; __i < __x.__size; ++__i)
+  {
+    const auto __val = __x[__i];
+    if (!(__val < __result))
```
miscco: Ditto simplify
```suggestion
    if (__result < __val)
```

Will stop mentioning

libcudacxx/include/cuda/std/__simd/reductions.h
```diff
@@ -0,0 +1,352 @@
...
+reduce_min(const basic_vec<_Tp, _Abi>& __x, const typename basic_vec<_Tp, _Abi>::mask_type& __mask) noexcept
+{
+  auto __result = numeric_limits<_Tp>::max();
+  _CCCL_PRAGMA_UNROLL_FULL()
+  for (__simd_size_type __i = 0; __i < __x.__size; ++__i)
+  {
+    if (__mask[__i])
+    {
+      const auto __val = __x[__i];
+      if (!(__result < __val))
```
miscco:
```suggestion
      if (__val < __result))
```

libcudacxx/include/cuda/std/__simd/reductions.h
```diff
@@ -0,0 +1,352 @@
...
+  _CCCL_PRAGMA_UNROLL_FULL()
+  for (__simd_size_type __i = 0; __i < __k.__size; ++__i)
+  {
+    if (__k[__i])
+    {
+      return __i;
+    }
+  }
+  _CCCL_UNREACHABLE();
+  return 0;
```
miscco: Important: This is only a precondition. We should return an invalid `__simd_size_type`, e.g `__simd_size_type(-1)` and _CCCL_ASSERT at the end. 

Currently this does twice the work
```suggestion
  _CCCL_PRAGMA_UNROLL_FULL()
  for (__simd_size_type __i = 0; __i < __k.__size; ++__i)
  {
    if (__k[__i])
    {
      return __i;
    }
  }
  _CCCL_ASSERT(false, "No bits are set");
  return __simd_size_tpe(-1);
```

libcudacxx/test/libcudacxx/std/numerics/simd/simd.reductions/mask_reductions.pass.cpp
miscco, line 1: Please split into individual tests, we can have `all_of`, `none_of` and `any_of` in the same file, but `reduce` and `reduce_count` are different APIs that should be in different files

We really want to make sure every test is as small and simple as possible. Putting multiple APIs in the same test just increases mental load on the reader

miscco (changes requested):

libcudacxx/include/cuda/std/__simd/reductions.h
fbusato: great catch!

libcudacxx/include/cuda/std/__simd/reductions.h
fbusato: specialized operator, e.g `plus<_Tp>` , are not supported by the c++ specification.  
About `::std::plus`, I'm always a bit concerned to mix cuda and std libraries

libcudacxx/include/cuda/std/__simd/reductions.h
fbusato: technically yes, but the SIMD ones cover fewer cases. I have to think about if we can merge them

libcudacxx/include/cuda/std/__simd/reductions.h
fbusato: I think this is correct. The default value of  `__binary_op` is part of the second overload

libcudacxx/include/cuda/std/__simd/reductions.h
fbusato: got it. only the template parameter has a default type

libcudacxx/include/cuda/std/__simd/reductions.h
fbusato: the c++ specification states this form `(!(__result < __val))`. I would not change it

libcudacxx/include/cuda/std/__simd/reductions.h
fbusato: same here

libcudacxx/include/cuda/std/__simd/reductions.h
fbusato: same here

libcudacxx/include/cuda/std/__simd/reductions.h
fbusato: but if the precondition is violated, should not be UB?

libcudacxx/include/cuda/std/__simd/reductions.h
fbusato: I don't entirely agree. The loop has a fixed number of iterations. Pragma unroll generates the same come of manual/compile-time unrolling.
Compile-time fold would just add pressure on compile time. I would postpose this modification until I have stronger motivations

libcudacxx/include/cuda/std/__simd/reductions.h
fbusato: the original code was right. The calling code uses `<Tp, BinaryOp>`. I need to revert/adapt the change

libcudacxx/include/cuda/std/__simd/reductions.h
fbusato: no, I think it is a bad idea. `cuda::identity_element`  can be also specialized by the user

libcudacxx/include/cuda/std/__simd/reductions.h
fbusato: see https://eel.is/c%2B%2Bdraft/simd#reductions-16

libcudacxx/include/cuda/std/__simd/reductions.h
fbusato: `::std::plus` is practically never used in CCCL. Anyway,  I added `operations_traits.h` to handle the traits for `cuda::std::plus`/`std::plus`, etc.

miscco: We dont use it, but users might its more of a followup for a different issue

libcudacxx/include/cuda/std/__simd/reductions.h
miscco: This is still not correct, you cannot initialize any `_BinaryOperation` with `plus<>{}` you need just `{}`

libcudacxx/include/cuda/std/__simd/reductions.h
miscco: Please file an editorial bug, the formulation makes no sense

libcudacxx/include/cuda/std/__simd/reductions.h
miscco: Its UB that is why I added the _CCCL_ASSERT(false)

libcudacxx/include/cuda/std/__simd/reductions.h
fbusato: created https://github.com/cplusplus/draft/issues/8991

libcudacxx/include/cuda/std/__simd/reductions.h
fbusato: I don't think so. [eel.is/c%2B%2Bdraft/simd#reductions-10](https://eel.is/c%2B%2Bdraft/simd#reductions-10) states:

> template<class T, class BinaryOperation = plus<>>
>  constexpr T reduce(const T& x, BinaryOperation binary_op = {});

which is present at line 91

libcudacxx/include/cuda/std/__functional/operations_traits.h
```diff
@@ -0,0 +1,80 @@
...
+#  pragma system_header
+#endif // no system header
+
+#include <cuda/std/__functional/operations.h>
+#include <cuda/std/__type_traits/is_same.h>
+
+// TODO(fbusato): move to _CCCL_HOSTED()
+#if !_CCCL_COMPILER(NVRTC)
+#  include <functional>
+#endif // !_CCCL_COMPILER(NVRTC)
```
miscco: Critical: We should not include `<functional>` for this. All of that should be forward declarations

libcudacxx/include/cuda/std/__functional/operations_traits.h
```diff
@@ -0,0 +1,80 @@
...
+#include <cuda/std/__cccl/prologue.h>
+
+_CCCL_BEGIN_NAMESPACE_CUDA_STD
+
+template <typename _BinaryOp>
+inline constexpr bool __is_plus_op_v =
+  is_same_v<_BinaryOp, plus<>>
+#if !_CCCL_COMPILER(NVRTC)
+  || is_same_v<_BinaryOp, ::std::plus<>>
+#endif // !_CCCL_COMPILER(NVRTC)
```
miscco: Important: Use `_CCCL_HOSTED()`

miscco (approved):

PR #8254, opened 2026-04-01
Recover `warp_shuffle` original behavior (revert #8210)

fbusato:
## Description

#8210 enforces additional constraints to the type allowed in `cuda::device::warp_shuffle_*`, namely default contractible and trivially copyable.  

While this is conceptually correct, the new constraints prevent using warp shuffle instructions with types that practically satisfy these property but where the type traits fail, e.g. `__half`, `__nv_bfloat16`, other reduced precision floating-point types, composition of them like array and structures. Enforcing such constraints prevent using them in many context, e.g. CUB.

This PR reverts the original behavior until we don't find a reliable way to prevent the problem.

### DIFF
--- a/libcudacxx/include/cuda/__warp/warp_shuffle.h
+++ b/libcudacxx/include/cuda/__warp/warp_shuffle.h
@@ -31,6 +31,7 @@
 #    include <cuda/std/__memory/addressof.h>
 #    include <cuda/std/__type_traits/enable_if.h>
 #    include <cuda/std/__type_traits/integral_constant.h>
+#    include <cuda/std/__type_traits/is_default_constructible.h>
 #    include <cuda/std/__type_traits/is_pointer.h>
 #    include <cuda/std/cstdint>
 
@@ -55,6 +56,7 @@ template <int _Width = 32, typename _Tp, typename _Up = ::cuda::std::remove_cv_t
 [[nodiscard]] _CCCL_DEVICE_API warp_shuffle_result<_Up> warp_shuffle_idx(
   const _Tp& __data, int __src_lane, uint32_t __lane_mask = 0xFFFFFFFF, ::cuda::std::integral_constant<int, _Width> = {})
 {
+  static_assert(::cuda::std::is_default_constructible_v<_Tp>, "_Tp must be default constructible");
   constexpr auto __warp_size   = 32u;
   constexpr bool __is_void_ptr = ::cuda::std::is_same_v<_Up, void*> || ::cuda::std::is_same_v<_Up, const void*>;
   static_assert(!::cuda::std::is_pointer_v<_Up> || __is_void_ptr,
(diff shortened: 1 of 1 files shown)


libcudacxx/include/cuda/__warp/warp_shuffle.h
```diff
@@ -60,9 +55,6 @@ template <int _Width = 32, typename _Tp, typename _Up = ::cuda::std::remove_cv_t
 [[nodiscard]] _CCCL_DEVICE_API warp_shuffle_result<_Up> warp_shuffle_idx(
   const _Tp& __data, int __src_lane, uint32_t __lane_mask = 0xFFFFFFFF, ::cuda::std::integral_constant<int, _Width> = {})
 {
-  static_assert(::cuda::std::is_default_constructible_v<_Tp>, "_Tp must be default constructible");
```
jrhemstad, line 63: question: Instead of wholesale removing these checks, should we just add explicit exceptions for known types? With the ability to allow people to proclaim types as valid for use with these APIs?

davebayer: As @fbusato explained to me, the problem is that if there is a `struct` containing `__half`, it won't be trivially copyable.. However, we do the same think for `cuda::std::bit_cast` and noone has complained yet.

But I think we should keep at least the requirement on default constructibility.

miscco: I would second default_constructability, because that is a much clearer error message than what a C++ compiler generates 5 lines below

jrhemstad: I'd recommend taking a look at what we did in cuCollections by offering a `is_bitwise_comparable` custom trait. By default, we use `has_unique_object_representation<T>`, but that is `false` for floating-point values due to NaNs. However, for the majority of use cases that doesn't matter, and so we allow an escape hatch of specializing `is_bitwise_comparable` to opt-in. We emit a helpful diagnostic when this situation arises pointing people towards specializing `is_bitwise_comparable`.

We could do something similar here. 

https://github.com/NVIDIA/cuCollections/blob/6477be2182668015f9a91e3a0bb7e248eceecd09/include/cuco/utility/traits.hpp#L24-L60

fbusato: the idea is a bit invasive but nice. The problem affects other warp instructions as well, so this solution applies to all of them. We can specialize the new type traits for reduced precision floating points + array. 

I opened an RFE for the compiler nvbug 5497120 a while ago. We can rely on the proposed solution until we don't get an official workaround.

fbusato: the funny aspect is that `has_unique_object_representation<T>` recognizes `__half`, `__nv_bfloat16` as unique object representation, while this is not the case

fbusato:
merging for now. We will switch to `cuda::is_trivially_copyable` when ready https://github.com/NVIDIA/cccl/pull/8265

Issue #8242, opened 2026-03-31
[BUG]: `cub::DeviceScan::ExclusiveScan` and `InclusiveScan` produce incorrect results with custom binary operator


shrshi:
### Is this a duplicate?

- [x] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cccl/issues) for this bug and that I agree to the [Code of Conduct](CODE_OF_CONDUCT.md)

### Type of Bug

Runtime Error

### Component

CUB

## Environment
- **GPU**: NVIDIA RTX PRO 6000 Blackwell Server Edition (Compute Capability 12.0)
- **Driver**: 580.95.05
- **CUDA**: 13.1 (V13.1.115)
- **CCCL Version**: 3.4.0 (CCCL_VERSION 3004000)
- **OS**: Ubuntu 24.04.4 LTS

## Description

`cub::DeviceScan::ExclusiveScan` produces incorrect results when using a custom binary operator that propagates values conditionally. 

## Additional Notes

- `cub::DeviceScan::InclusiveScan` exhibits the same bug at the same positions

## Minimal Reproducible Example

```cpp
#include <cub/cub.cuh>
#include <cuda_runtime.h>
#include <numeric>
#include <vector>
#include <cstdio>

// Custom operator that propagates the last "write" value through "read" values
template <typename StackSymbolT>
struct PropagateLastWrite {
  __host__ __device__ StackSymbolT operator()(StackSymbolT const& lhs,
                                               StackSymbolT const& rhs) const
  {
    bool is_rhs_read  = (rhs == read_symbol);
    bool is_lhs_write = (lhs != read_symbol);
    return (is_rhs_read && is_lhs_write) ? lhs : rhs;
  }
  StackSymbolT read_symbol;
};

int main()
{
  constexpr size_t num_elements     = 8160;
  constexpr char read_symbol        = 'x';
  constexpr char empty_stack_symbol = '_';

  // Create input - mostly read_symbol with scattered write values
  std::vector<char> h_input(num_elements, read_symbol);

  // Key write values that trigger the bug
  h_input[8073] = '[';
  h_input[8074] = '{';
  h_input[8075] = '[';
  h_input[8076] = '{';  // Last '{' before positions 8077-8147 (all 'x')
  h_input[8148] = '_';
  h_input[8151] = '{';
  h_input[8152] = '_';
  h_input[8154] = '[';
  h_input[8155] = '_';
  h_input[8157] = '[';
[...]

PR #8259, opened 2026-04-01
Fix cub ExclusiveScan logic for associative-only operator

davidwendt:
## Description

Fixes the kernel_scan_warpspeed logic to reorder the parameters to the input `scan_op` to support associative operators that are not necessarily commutative. The current logic works fine only if the `scan_op` is both associative and commutative.

Closes https://github.com/NVIDIA/cccl/issues/8242

## Checklist

### DIFF
--- a/cub/test/catch2_test_device_exclusive_scan_noncommutative.cu
+++ b/cub/test/catch2_test_device_exclusive_scan_noncommutative.cu
@@ -27,7 +27,8 @@
  *    Ref: https://en.wikipedia.org/wiki/Monoid
  */
 
-/* PropagateLastWrite: an associative, non-commutative operator over a symbol alphabet.
+/**
+ * PropagateLastWrite: an associative, non-commutative operator over a symbol alphabet.
  *
  * Given a read_symbol r, the operator is:
  *     op(lhs, rhs) = lhs  if  rhs == r  and  lhs != r
@@ -56,7 +57,6 @@
 
 namespace impl
 {
-
 // bicyclic monoid operator: associative and non-commutative
 template <typename UnsignedIntegralT>
 struct bicyclic_monoid_op
@@ -119,7 +119,6 @@ struct populate_sparse_write_input
     return (idx % static_cast<size_t>(write_period) == 0) ? write_symbol : read_symbol;
   }
 };
-
 } // namespace impl
 
 // Sizes chosen to cover: single-element, small, medium, the original bug-report size (8160),
@@ -178,7 +177,7 @@ C2H_TEST("Device exclusive scan works with PropagateLastWrite operator", "[scan]
   // in the wrong order when combining the inter-tile exclusive prefix with the intra-tile prefix,
   // producing incorrect results for non-commutative operators like this one.
 
-  using symbol_t = char;
+  using symbol_t                        = char;
   constexpr symbol_t read_symbol        = 'x';
   constexpr symbol_t empty_stack_symbol = '_'; // init_value for the exclusive scan
   constexpr symbol_t write_symbol       = '[';
@@ -196,9 +195,7 @@ C2H_TEST("Device exclusive scan works with PropagateLastWrite operator", "[scan]
 
   c2h::device_vector<symbol_t> input(num_items);
   thrust::tabulate(
-    input.begin(),
-    input.end(),
-    impl::populate_sparse_write_input<symbol_t>{read_symbol, write_symbol, write_period});
+    input.begin(), input.end(), impl::populate_sparse_write_input<symbol_t>{read_symbol, write_symbol, write_period});
   c2h::device_vector<symbol_t> output(num_items);
 
   symbol_t* d_input  = thrust::raw_pointer_cast(input.data());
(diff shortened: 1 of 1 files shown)


davidwendt:
> @davidwendt what operation did you catch this on, was it "just" floats?

No, a custom operator. The linked issue has all the details: https://github.com/NVIDIA/cccl/issues/8242
I'm trying to get Claude to create me a cub test based on the issue.

shrshi:
@davidwendt we saw the same bug in inclusive scan too. Does your change fix that code path as well?

davidwendt:
> @davidwendt we saw the same bug in inclusive scan too. Does your change fix that code path as well?

Yes. I tested with your draft PR and everything passed.

davidwendt:
Added a 300 line unit test for a one line code change.
I asked claude to generate the test code for this.

cub/test/catch2_test_device_exclusive_scan_noncommutative.cu
bdice, line 1: The existing test is fine. I wondered if we could do something with a simpler associative but noncommutative operator. Consider $f( (a, b), (c, d) ) \to (a, d)$. We can write that pretty simply by packing two values into the upper/lower bits of a `uint32`.

No need to change this -- just explored the idea and wanted to share the result.

Something like this (untested):

```cpp

#include <cub/device/device_scan.cuh>

#include <catch2_test_device_scan.cuh>

namespace
{
struct op_t
{
  __host__ __device__ uint32_t operator()(uint32_t lhs, uint32_t rhs) const
  {
    return (lhs & 0xffff0000u) | (rhs & 0x0000ffffu);
  }
};
} // namespace

C2H_TEST("Device exclusive scan works with associative non-commutative uint32_t operator", "[scan][device]")
{
  using T = uint32_t;

  constexpr int num_items = 50'000;
  constexpr T init_value  = 0;

  c2h::host_vector<T> h_input(num_items);
  for (int i = 0; i < num_items; ++i)
  {
    h_input[i] = (static_cast<T>((17 * i + 1) % 251) << 16) | static_cast<T>((29 * i + 7) % 241);
  }

  c2h::device_vector<T> input(h_input);
  c2h::device_vector<T> output(num_items);

  T* d_input  = thrust::raw_pointer_cast(input.data());
  T* d_output = thrust::raw_pointer_cast(output.data());

  size_t tmp_size{};
  REQUIRE(cudaSuccess == cub::DeviceScan::ExclusiveScan(nullptr, tmp_size, d_input, d_output, op_t{}, init_value, num_items));
  REQUIRE(tmp_size > 0);

  using cuda::std::byte;
  c2h::device_vector<byte> tmp(tmp_size);
  REQUIRE(cudaSuccess
          == cub::DeviceScan::ExclusiveScan(
            thrust::raw_pointer_cast(tmp.data()), tmp_size, d_input, d_output, op_t{}, init_value, num_items));

  c2h::host_vector<T> h_output(output);
  c2h::host_vector<T> h_expected(num_items);
  compute_exclusive_scan_reference(h_input.cbegin(), h_input.cend(), h_expected.begin(), init_value, op_t{});

  REQUIRE(h_expected == h_output);
}
```

miscco:
Dont worry about the rapids CI I would tackle that next week

miscco:
> Added a 300 line unit test for a one line code change.

Welcome to CCCL where 99% is fighting CI and compilers

cub/test/catch2_test_device_exclusive_scan_noncommutative.cu
davidwendt: Sorry, this did not trigger the error when I ran it locally.

davidwendt:
Any chance this fix could be applied to CCCL 3.4?
I'm not sure the process for this kind of request.

miscco:
> Any chance this fix could be applied to CCCL 3.4? 

Do you mean 3.3? Which would be the next release candidate

cub/test/catch2_test_device_exclusive_scan_noncommutative.cu
bdice: Ah, darn. Sorry.

miscco:
/ok to test [783fe52](https://github.com/NVIDIA/cccl/pull/8259/commits/783fe52dfbb6883be97d707c5c60f436d474cb11)

bdice:
This is fine to merge to `main` (3.4), we are already pulling CCCL from `main`. I'll just update the commit we're using once this goes in. That's why I need to get RMM fixed up before we update the RAPIDS pinning of CCCL.

(Update: see https://github.com/rapidsai/rmm/pull/2343 for that fix)

davidwendt:
> > Any chance this fix could be applied to CCCL 3.4?
> 
> Do you mean 3.3? Which would be the next release candidate

No. 3.4 is what we are pulling right now and where the error is. The error is not in 3.3.

bernhardmgruber:
Btw, we do have a test for scanning non-commutative operators, but the input set is very small, so it probably didn't hit the case fixed by this PR: https://github.com/NVIDIA/cccl/blob/bb8a5b9e26010f932042a302bddbccea8621819a/thrust/testing/scan.cu#L743-L789

davidwendt:
Can this still be merged despite the `Build RAPIDS` failures?

bdice:
> Can this still be merged despite the `Build RAPIDS` failures?

Yes, those are nonblocking. I think I can merge since CI is passing and @miscco has approved.

PR #3183, opened 2024-12-17
Ensure `cuda::std::bit_cast` works with vector types

miscco:
We want to make sure we can use `bit_cast` with suitable vector or sum types like
* `cuda::std::array`
* c-arrays
* cuda vector types like float2

Unfortunately we cannot use it with the extended floating point vector types, because they are not trivially copyable and also not trivially copy assignable

PR #8265, opened 2026-04-01
`cuda::is_trivially_copyable`

fbusato:
## Description

Followup of the discussion in

- https://github.com/NVIDIA/cccl/pull/8254

The PR introduces `cuda::is_trivially_copyable_relaxed` to support types that are actually trivially copyable but not recognized by the C++ `std::is_trivially_copyable`  type trait. 

The new trait supports:

- trivially copyable types.
- extended floating-point types.
- extended floating-point vector types.
- Raw arrays, `cuda::std::array`.
- `cuda::std::pair`.
- `cuda::std::tuple`.
- Nested compositions of the above cases.

### Potentially affected paths by `std::is_trivially_copyable`

unsupported cases:

- `cuda/std/__atomic/types/base.h`:39
- `cuda/std/__atomic/types/reference.h`:39
- `cuda/__memcpy_async/memcpy_async_barrier.h`:61
- `cuda/__memcpy_async/memcpy_async_tx.h`:58
- `cuda/__container/buffer.h`:111
- `cuda/__algorithm/copy.h`:73
- `cuda/__algorithm/fill.h`:42
- `cuda/std/string_view`:147

fallback to slower path:

- `cuda/std/__algorithm/copy.h`:102 
- `cuda/std/__algorithm/copy_backward.h`:50
- `cuda/std/__algorithm/move.h`:54
- `cuda/std/__algorithm/move_backward.h`:53
- `cuda/std/__bit/bit_cast.h`:55
- `cub/device/dispatch/kernels/kernel_histogram.cuh`:81
- `cub/detail/uninitialized_copy.cuh`:33
- `thrust/type_traits/is_trivially_relocatable.h`:213
- `cudax/include/cuda/experimental/__kernel/kernel_ref.cuh`:54

Require https://github.com/NVIDIA/cccl/pull/8347


miscco (changes requested):
I am not too excited about this.

We need to be really careful here because the compiler may also break in some of those cases.

I did some experiments in #3183

bernhardmgruber:
I wanted to start a discussion for some time to add such a trait, but I would really keep this feature internally for now until it is sufficiently baked. I would really like this feature to make `thrust::is_trivially_relocatable` and `THRUST_PROCLAIM_TRIVIALLY_RELOCATABLE` obsolete, because those two are often used wrongly, so we should make sure the new trait here can replace it.

docs/libcudacxx/extended_api/type_traits/is_trivially_copyable_relaxed.rst
```diff
@@ -0,0 +1,90 @@
...
+- ``cuda::std::array<T, N>``: relaxed trivially copyable when ``T`` is.
+- ``cuda::std::pair<T1, T2>``: relaxed trivially copyable when both ``T1`` and ``T2`` are.
+- ``cuda::std::tuple<Ts...>``: relaxed trivially copyable when all ``Ts...`` are.
+
+``const``, ``volatile``, and ``const volatile`` qualifications are handled transparently.
+
+Custom Specialization
+---------------------
+
+Users may specialize ``cuda::is_trivially_copyable_relaxed`` for their own types whose memory representation is safe to copy
```
davebayer: This is inconsistent with `cuda::is_floating_point`, there we say they should specialize the `_v` trait

docs/libcudacxx/extended_api/type_traits/is_trivially_copyable_relaxed.rst
```diff
@@ -0,0 +1,90 @@
...
+with ``memcpy`` but that the compiler does not consider trivially copyable.
```
davebayer: Isn't this just UB?

davebayer (changes requested):
I am not a fan of this trait. We bend C++ rules to fix poorly designed nvfp types. We would have to basically change every use of `is_trivially_copyable` and `is_trivially_copy_constructible` to this trait to make it work consistently in libcu++.

I don't think this is a good idea, we should rather insist of the nvfp types being fixed.

docs/libcudacxx/extended_api/type_traits/is_trivially_copyable_relaxed.rst
fbusato: not for the types that we care about. Said that, the user could provide an object that triggers UB. I can highlight this point in the documentation but we cannot do anything to explicitly prevent it.

docs/libcudacxx/extended_api/type_traits/is_trivially_copyable_relaxed.rst
fbusato: done

fbusato:
> We need to be really careful here because the compiler may also break in some of those cases.

> I did some experiments in https://github.com/NVIDIA/cccl/pull/3183

Integrated and extended the tests that you point out. Everything works

fbusato:
> I wanted to start a discussion for some time to add such a trait, but I would really keep this feature internally for now until it is sufficiently baked. I would really like this feature to make thrust::is_trivially_relocatable and THRUST_PROCLAIM_TRIVIALLY_RELOCATABLE obsolete, because those two are often used wrongly, so we should make sure the new trait here can replace it.

the actual issue is that the user can extend the trait to custom types. It needs to be public

fbusato:
> I am not a fan of this trait. We bend C++ rules to fix poorly designed nvfp types. We would have to basically change every use of is_trivially_copyable and is_trivially_copy_constructible to this trait to make it work consistently in libcu++.
> I don't think this is a good idea, we should rather insist of the nvfp types being fixed.

there are no plan for that. nvfp types are not trivially copyable for optimization purposes.

jrhemstad:
> we should rather insist of the nvfp types being fixed.

This isn't going to happen and we shouldn't delude ourselves into thinking it ever will.

docs/libcudacxx/extended_api/type_traits/is_trivially_copyable_relaxed.rst
```diff
@@ -0,0 +1,110 @@
...
+A type ``T`` satisfies ``cuda::is_trivially_copyable_relaxed`` if any of the following holds:
+
+- ``T`` is trivially copyable.
+- ``T`` is an extended floating-point scalar type (e.g. ``__half``, ``__nv_bfloat16``, ``__nv_fp8_e4m3``).
+- ``T`` is an extended floating-point vector type (e.g. ``__half2``, ``__nv_bfloat162``, ``__nv_fp8x2_e4m3``).
+
+The trait also propagates through composite types:
+
+- C-style arrays: ``T[N]`` and ``T[]`` are relaxed trivially copyable when ``T`` is.
+- ``cuda::std::array<T, N>``: relaxed trivially copyable when ``T`` is.
```
gonidelis:
```suggestion
- ``cuda::std::array<T, N>``: relaxed trivially copyable when ``T`` is also trivially copyable.
```

maybe?

docs/libcudacxx/extended_api/type_traits/is_trivially_copyable_relaxed.rst
```diff
@@ -0,0 +1,110 @@
...
+- ``cuda::std::pair<T1, T2>``: relaxed trivially copyable when both ``T1`` and ``T2`` are and the object has no padding.
```
gonidelis: why do we exclude pairs that have padding since this passes? https://godbolt.org/z/d856xqroj isn't it equivalent?

docs/libcudacxx/extended_api/type_traits/is_trivially_copyable_relaxed.rst
```diff
@@ -0,0 +1,110 @@
...
+- ``cuda::std::tuple<Ts...>``: relaxed trivially copyable when all ``Ts...`` are and the object has no padding.
+
+``const`` qualification is handled transparently, while ``volatile`` is compiler dependent.
+
+.. note::
+
+  The type trait cannot determine if a structure (``struct`` or ``class``) contains extended floating-point types, and thus it recognizes the type as *trivially copyable*. The user must manually specialize the type trait for such types.
```
gonidelis: > The type trait cannot determine if a structure (``struct`` or ``class``) contains extended floating-point types

we could determine if type contains extended fp types with structured bindings https://en.cppreference.com/w/cpp/language/structured_binding.html 

credit to @gevtushenko

docs/libcudacxx/extended_api/type_traits/is_trivially_copyable_relaxed.rst
```diff
@@ -0,0 +1,110 @@
...
+
+Custom Specialization
+---------------------
+
+Users may specialize ``cuda::is_trivially_copyable_relaxed_v`` for types whose semantics allow copying with ``memcpy``, but which the compiler does not consider to be trivially copyable.
+
+A `trivially copyable <https://en.cppreference.com/w/cpp/language/classes.html>`__ class is a class that
+
+- has at least one eligible copy constructor, move constructor, copy assignment operator, or move assignment operator,
+- each eligible copy constructor is trivial
```
gonidelis:
```suggestion
- each of its eligible copy constructors is trivial
```

libcudacxx/include/cuda/__type_traits/is_trivially_copyable_relaxed.h
```diff
@@ -0,0 +1,72 @@
...
+#include <cuda/std/__type_traits/is_trivially_copyable.h>
+#include <cuda/std/__type_traits/remove_const.h>
+
+#include <cuda/std/__cccl/prologue.h>
+
+_CCCL_BEGIN_NAMESPACE_CUDA
+
+//! Users are allowed to specialize this variable template for their own types
+template <typename _Tp>
+constexpr bool is_trivially_copyable_relaxed_v =
```
gonidelis: What are the arguments for adding the `_relaxed` suffix

libcudacxx/include/cuda/__type_traits/is_trivially_copyable.h
```diff
@@ -0,0 +1,72 @@
...
+
+_CCCL_BEGIN_NAMESPACE_CUDA
+
+//! Users are allowed to specialize this variable template for their own types
+template <typename _Tp>
+constexpr bool is_trivially_copyable_relaxed_v =
+  ::cuda::std::is_trivially_copyable_v<::cuda::std::remove_const_t<_Tp>>
+  || ::cuda::std::__is_extended_floating_point_v<::cuda::std::remove_const_t<_Tp>>
+#if _CCCL_HAS_CTK()
+  || ::cuda::is_extended_fp_vector_type_v<::cuda::std::remove_const_t<_Tp>>
```
gonidelis: Why is_vector_type.h guarded by `_CCCL_HAS_CTK()` in the first place?

This question stems from the original question: why is there a difference between `is_extended_fp_vector_type_v` and `__is_extended_floating_point_v` to begin with?

docs/libcudacxx/extended_api/type_traits/is_trivially_copyable_relaxed.rst
fbusato: padding description is wrong. Thanks!

docs/libcudacxx/extended_api/type_traits/is_trivially_copyable_relaxed.rst
fbusato: working on that :)

libcudacxx/include/cuda/__type_traits/is_trivially_copyable_relaxed.h
fbusato: I'm totally fine in removing it. looks too verbose

libcudacxx/include/cuda/__type_traits/is_trivially_copyable.h
fbusato: vector types (floating-point and integrals) are only defined in the CUDA toolkit headers, not at compile-level. So this line looks correct

libcudacxx/test/libcudacxx/cuda/type_traits/is_trivially_copyable.aggr.pass.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+#include <cuda/type_traits>
+
+template <class _Tp>
+__host__ __device__ void test_is_trivially_copyable()
+{
+  static_assert(cuda::is_trivially_copyable<_Tp>::value);
+  static_assert(cuda::is_trivially_copyable<const _Tp>::value);
+  static_assert(cuda::is_trivially_copyable_v<_Tp>);
+  static_assert(cuda::is_trivially_copyable_v<const _Tp>);
```
miscco: Critical: We must ensure that this type does not only satisfy the trait, but can also be used in e.g bit_cast and memcpy without the compiler throwing up

miscco (changes requested):

docs/libcudacxx/extended_api/type_traits/is_trivially_copyable.rst
```diff
@@ -0,0 +1,103 @@
...
+- has at least one eligible copy constructor, move constructor, copy assignment operator, or move assignment operator,
+- each of its eligible copy constructors is trivial
+- each eligible move constructor is trivial
+- each eligible copy assignment operator is trivial
+- each eligible move assignment operator is trivial, and
+- has a non-deleted trivial destructor.
+
+.. warning::
+
+    The user is responsible for ensuring that the type is actually trivially copyable when specializing this variable template. Otherwise, the behavior is undefined.
```
bernhardmgruber: Suggestion: I think this information is misleading, because a user opting into `cuda::is_trivially_copyable_v` gives us permission to ignore any of these requirements. Also, the user does not need to uphold any of these requirements. I would just drop this part.

docs/libcudacxx/extended_api/type_traits/is_trivially_copyable.rst
```diff
@@ -0,0 +1,103 @@
...
+   static_assert(cuda::is_trivially_copyable_v<__half2>);
+
+   // Padding-free composite types containing extended floating-point types
+   static_assert(cuda::is_trivially_copyable_v<__half[4]>);
+   static_assert(cuda::is_trivially_copyable_v<cuda::std::array<__half, 4>>);
+   static_assert(cuda::is_trivially_copyable_v<cuda::std::pair<__half, __half>>);
+   static_assert(cuda::is_trivially_copyable_v<cuda::std::tuple<__half, __half>>);
+
+   // Composites with padding are not trivially copyable
+   static_assert(!cuda::is_trivially_copyable_v<cuda::std::pair<__half, int>>);
```
bernhardmgruber: Q: Why not? I would expect them to be.

fbusato: artifact from the previous version. thanks

docs/libcudacxx/extended_api/type_traits/is_trivially_copyable.rst
miscco: I would still strongly prefer this to not be a user-definable trait

fbusato: makes sense. It is too low level to expose specialization to user.

libcudacxx/test/libcudacxx/cuda/type_traits/is_trivially_copyable.aggr.pass.cpp
fbusato: my plan was to update bit_cast in a second PR, but it also makes sense to add everything together.

libcudacxx/include/cuda/std/__bit/bit_cast.h
```diff
@@ -42,37 +41,45 @@ _CCCL_BEGIN_NAMESPACE_CUDA_STD
...
-#  endif // _CCCL_COMPILER(GCC, >=, 8)
+#endif // _CCCL_COMPILER(GCC, >=, 8)
+
+_CCCL_TEMPLATE(class _To, class _From)
+_CCCL_REQUIRES((sizeof(_To) == sizeof(_From)) _CCCL_AND(::cuda::is_trivially_copyable_v<_To>)
+                 _CCCL_AND(::cuda::is_trivially_copyable_v<_From>))
+[[nodiscard]] _CCCL_API inline _CCCL_CONSTEXPR_BIT_CAST _To bit_cast(const _From& __from) noexcept
+{
+#if defined(_CCCL_BUILTIN_BIT_CAST)
+  if constexpr (::cuda::std::is_trivially_copyable_v<_To> && ::cuda::std::is_trivially_copyable_v<_From>)
```
fbusato, line 73: here the trick. We know that `__nv_bfloat2` and `__half2` are not trivially copyable but, we also know, that these types are not available in constant expressions so we can dispatch depending on this property

libcudacxx/include/cuda/__type_traits/is_trivially_copyable.h
```diff
@@ -0,0 +1,84 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
```
miscco: Should be the libcu++ license

libcudacxx/include/cuda/__type_traits/is_trivially_copyable.h
```diff
@@ -0,0 +1,84 @@
...
+#include <cuda/std/__type_traits/is_extended_floating_point.h>
+#include <cuda/std/__type_traits/is_trivially_copyable.h>
+#include <cuda/std/__type_traits/remove_const.h>
+
+#include <cuda/std/__cccl/prologue.h>
+
+_CCCL_BEGIN_NAMESPACE_CUDA
+
+template <typename _Tp, typename = void>
+constexpr bool __is_aggregate_trivially_copyable_v = false;
```
miscco: Critical: All those need to be `inline constexpr`

libcudacxx/include/cuda/std/__bit/bit_cast.h
```diff
@@ -42,37 +41,48 @@ _CCCL_BEGIN_NAMESPACE_CUDA_STD
...
+  {
+    return _CCCL_BUILTIN_BIT_CAST(_To, __from);
+  }
+  else
+  {
+    return ::cuda::std::__bit_cast_memcpy<_To>(__from);
+  }
+#else // ^^^ _CCCL_BUILTIN_BIT_CAST ^^^ / vvv !_CCCL_BUILTIN_BIT_CAST vvv
+  return ::cuda::std::__bit_cast_memcpy<_To>(__from);
 #endif // !_CCCL_BUILTIN_BIT_CAST
```
miscco: We can shorten this
```suggestion
#if defined(_CCCL_BUILTIN_BIT_CAST)
  if constexpr (::cuda::std::is_trivially_copyable_v<_To> && ::cuda::std::is_trivially_copyable_v<_From>)
  {
    return _CCCL_BUILTIN_BIT_CAST(_To, __from);
  }
  else
#endif // _CCCL_BUILTIN_BIT_CAST
  {
    return ::cuda::std::__bit_cast_memcpy<_To>(__from);
  }
```

libcudacxx/include/cuda/std/__type_traits/aggregate_members.h
```diff
@@ -0,0 +1,163 @@
...
+
+_CCCL_DIAG_PUSH
+_CCCL_DIAG_SUPPRESS_CLANG("-Wmissing-field-initializers")
+
+_CCCL_BEGIN_NAMESPACE_CUDA_STD
+
+#if defined(_CCCL_BUILTIN_STRUCTURED_BINDING_SIZE)
+
+// Returns the number of aggregate members, or `-1` if the type is not an aggregate.
+template <typename _Tp, ::cuda::std::enable_if_t<_CCCL_BUILTIN_STRUCTURED_BINDING_SIZE(_Tp) >= 0, int> = 0>
```
miscco: Critical: This is not defined in libcu++ but only in cudax currently.

We need to move the builtin definition into libcu++

libcudacxx/include/cuda/std/__type_traits/aggregate_members.h
```diff
@@ -0,0 +1,163 @@
...
+constexpr int __aggregate_arity_v = _CCCL_BUILTIN_STRUCTURED_BINDING_SIZE(_Tp);
```
miscco: Critical: missing `inline`
```suggestion
inline constexpr int __aggregate_arity_v = _CCCL_BUILTIN_STRUCTURED_BINDING_SIZE(_Tp);
```

libcudacxx/include/cuda/std/__type_traits/aggregate_members.h
```diff
@@ -0,0 +1,163 @@
...
+// T is not an aggregate, return 1
+template <typename _Tp>
+struct __aggregate_arity_impl<_Tp, false>
+{
+  _CCCL_API auto operator()() const -> char*;
+};
+
+// Returns the number of aggregate members, or `-1` if the type is not an aggregate.
+template <typename _Tp>
+constexpr int __aggregate_arity_v = int{sizeof(*__aggregate_arity_impl<_Tp>{}())} - 2;
```
miscco: Critical: missing `inline`
```suggestion
inline constexpr int __aggregate_arity_v = int{sizeof(*__aggregate_arity_impl<_Tp>{}())} - 2;
```

libcudacxx/include/cuda/std/__type_traits/aggregate_members.h
miscco, line 1: Please move this out of the PR.

This is a nontrivial piece of code and the current implementation seems really costly.

I want this in a separate PR so that we can properly review it

libcudacxx/test/libcudacxx/cuda/type_traits/is_trivially_copyable.aggr.pass.cpp
```diff
@@ -0,0 +1,114 @@
...
+}
```
miscco: Important: This should test that we can perform a bit_cast from this type to a similarly sized type

libcudacxx/test/libcudacxx/cuda/type_traits/is_trivially_copyable.pass.cpp
```diff
@@ -0,0 +1,129 @@
...
+
+#include "test_macros.h"
+
+template <class T>
+__host__ __device__ void test_is_trivially_copyable()
+{
+  static_assert(cuda::is_trivially_copyable<T>::value);
+  static_assert(cuda::is_trivially_copyable<const T>::value);
+  static_assert(cuda::is_trivially_copyable_v<T>);
+  static_assert(cuda::is_trivially_copyable_v<const T>);
```
miscco: Missing tests for bitcast back and forth

miscco (changes requested):

libcudacxx/test/libcudacxx/cuda/type_traits/is_trivially_copyable.pass.cpp
fbusato: I added many new tests to `bit_cast` unit tests

libcudacxx/include/cuda/__type_traits/is_trivially_copyable.h
fbusato: why? `inline` is redundant here. ODR allows multiple identical definitions of template variables across translation units. Is a specific limitation of nvcc?

libcudacxx/test/libcudacxx/cuda/type_traits/is_trivially_copyable.aggr.pass.cpp
fbusato: I added many new tests to bit_cast unit tests

libcudacxx/include/cuda/std/__type_traits/aggregate_members.h
fbusato: Fine, I based the implementation on the code from stdexec. Eric already review it

libcudacxx/include/cuda/__type_traits/is_trivially_copyable.h
miscco: We never want multiple definitions of the same variable template ever, thats why every variable template is an inline variable

libcudacxx/include/cuda/std/__type_traits/aggregate_members.h
miscco: I am not against merging it, but I want it in a separate PR

libcudacxx/include/cuda/std/__bit/bit_cast.h
```diff
@@ -42,37 +41,43 @@ _CCCL_BEGIN_NAMESPACE_CUDA_STD
...
-          enable_if_t<is_trivially_copyable_v<_To> || __is_extended_floating_point_v<_To>, int>     = 0,
-          enable_if_t<is_trivially_copyable_v<_From> || __is_extended_floating_point_v<_From>, int> = 0>
-[[nodiscard]] _CCCL_API inline _CCCL_CONSTEXPR_BIT_CAST _To bit_cast(const _From& __from) noexcept
+template <class _To, class _From>
+[[nodiscard]] _CCCL_API inline _To __bit_cast_memcpy(const _From& __from) noexcept
 {
-#if defined(_CCCL_BUILTIN_BIT_CAST)
-  return _CCCL_BUILTIN_BIT_CAST(_To, __from);
-#else // ^^^ _CCCL_BUILTIN_BIT_CAST ^^^ / vvv !_CCCL_BUILTIN_BIT_CAST vvv
-  static_assert(is_trivially_default_constructible_v<_To>,
```
fbusato, line 62: C++ specification does not impose this constrain https://eel.is/c++draft/bit.cast. `is_trivially_default_constructible_v` is too strict. `cuda::std::complex` fails for example.

We only need to check if `_To __temp;` compiles. The best match is C++20 `std::default_initializable`

libcudacxx/include/cuda/std/__type_traits/aggregate_members.h
fbusato: created a new PR https://github.com/NVIDIA/cccl/pull/8347

miscco:
@fbusato I checked the standard and it turns out that C++23 made complex trivially copyable for standard floating point types, so we should only specialize for `__half` and `__nv_bfloat16`

See https://github.com/NVIDIA/cccl/pull/8356

fbusato:
- entirely refactored the implementation relying on `__aggregate_all_of_v`. 
- replaced `cuda::std::is_trivially_copyable` with `cuda::is_trivially_copyable` in all CCCL library paths (but not in tests/benchmarks)

libcudacxx/test/libcudacxx/cuda/type_traits/is_trivially_copyable.mem.pass.cpp
```diff
@@ -0,0 +1,323 @@
+//===----------------------------------------------------------------------===//
```
fbusato, line 1: I'm inclined to remove this file because it is very similar with `bit_cast.pass.cpp`

docs/libcudacxx/extended_api/type_traits/is_trivially_copyable.rst
```diff
@@ -0,0 +1,72 @@
...
+   template <typename T>
+   constexpr bool is_trivially_copyable_v = /* see below */;
+
+   template <typename T>
+   using is_trivially_copyable = cuda::std::bool_constant<is_trivially_copyable_v<T>>;
+
+   } // namespace cuda
+
+``cuda::is_trivially_copyable_v`` trait evaluates if a type can be copied by copying its underlying bytes.
+It extends ``cuda::std::is_trivially_copyable`` to also recognize CUDA extended floating-point vector types as trivially copyable.
```
Jacobfaib, line 21: Should also mention whether it is legal for users to specialize the trait, and give an example of how to do that if so. Given the definition above it implies that users should specialize `is_trivially_copyable_v` instead of `is_trivially_copyable`, usually it is the other way around.

docs/libcudacxx/extended_api/type_traits/is_trivially_copyable.rst
```diff
@@ -0,0 +1,72 @@
...
+
+``cuda::is_trivially_copyable_v<T>`` relies on ``cuda::std::is_trivially_copyable`` but adds support for CUDA-specific types.
```
Jacobfaib: If it _is_ legal to specialize, should elaborate whether the user should specialize `cuda::std::is_trivially_copyable` or `cuda::is_trivially_copyable`

docs/libcudacxx/extended_api/type_traits/is_trivially_copyable.rst
```diff
@@ -0,0 +1,72 @@
...
+
+The trait also propagates through composite types:
+
+- C-style arrays: ``T[N]`` and ``T[]`` are trivially copyable when ``T`` is.
+- ``cuda::std::array<T, N>``: trivially copyable when ``T`` is.
+- ``cuda::std::pair<T1, T2>``: trivially copyable when both ``T1`` and ``T2`` are.
+- ``cuda::std::tuple<Ts...>``: trivially copyable when all ``Ts...`` are.
+- ``cuda::std::complex<T>``: trivially copyable when ``T`` is.
+- ``cuda::complex<T>``: trivially copyable when ``T`` is.
+- Aggregates (types that can be initialized with a braced initializer list ``{}``): trivially copyable when all their members are.
```
Jacobfaib: Nit: since you mention initialization, you technically mean types that are _aggregate initializeable_, not true aggregate types. True aggregates have much stricter requirements.

docs/libcudacxx/extended_api/type_traits/is_trivially_copyable.rst
```diff
@@ -0,0 +1,72 @@
...
+
+  - On MSVC, recursive data-member inspection is not supported beyond the first level.
+
+``const`` qualification is handled transparently, while ``volatile`` is compiler dependent.
+
+Examples
+--------
+
+.. code:: cuda
```
Jacobfaib, line 45: Important: Could you `.. literalinclude::` this instead from the test file? That way it doesn't bit-rot.

libcudacxx/include/cuda/__type_traits/is_trivially_copyable.h
```diff
@@ -0,0 +1,109 @@
...
+    ::cuda::std::__aggregate_all_of_v<__is_trivially_copyable_callable, _Tp>;
+
+//----------------------------------------------------------------------------------------------------------------------
+// public traits
+
+template <typename _Tp>
+inline constexpr bool is_trivially_copyable_v = __is_trivially_copyable_v<_Tp>;
+
+template <typename _Tp>
+inline constexpr bool is_trivially_copyable_v<const _Tp> = is_trivially_copyable_v<_Tp>;
```
Jacobfaib, line 100: No `volatile`?

libcudacxx/include/cuda/__type_traits/is_trivially_copyable.h
```diff
@@ -0,0 +1,109 @@
...
+_CCCL_BEGIN_NAMESPACE_CUDA
+
+template <typename _Tp, typename = void>
+inline constexpr bool __is_aggregate_trivially_copyable_v = false;
+
+template <typename _Tp>
+inline constexpr bool __is_trivially_copyable_v =
+  ::cuda::std::is_trivially_copyable_v<_Tp> || __is_aggregate_trivially_copyable_v<_Tp>;
+
+#if _CCCL_HAS_NVFP16()
```
Jacobfaib, line 47: Should this explicit specialization be expanded to all of the extended floating-point types as well?

libcudacxx/test/libcudacxx/cuda/type_traits/is_trivially_copyable.basic_types.pass.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+#include <cuda_bf16.h>
+#include <cuda_fp16.h>
+
+#include <cuda/type_traits>
+
+#include "test_macros.h"
+
+template <class T>
+__host__ __device__ void test_is_trivially_copyable()
```
Jacobfaib: critical: `TEST_FUNC`

libcudacxx/test/libcudacxx/cuda/type_traits/is_trivially_copyable.basic_types.pass.cpp
Jacobfaib, line 1: Important: can you add tests that the trait works even when it is user-specialized?

libcudacxx/include/cuda/std/__bit/bit_cast.h
Jacobfaib: Should this be `::cuda::is_trivially_copyable_v`?

docs/libcudacxx/extended_api/type_traits/is_trivially_copyable.rst
fbusato: specializing `is_trivially_copyable_v` could be too far. The only reason we are proving it is to overcome CUDA data type limitations. I don't see a good motivation for user to specialize it

docs/libcudacxx/extended_api/type_traits/is_trivially_copyable.rst
fbusato: true, but I wanted to give users an idea on what an aggregate is. Maybe I can pointed out the cppref documentation

libcudacxx/include/cuda/__type_traits/is_trivially_copyable.h
fbusato: `volatile` is compile-specific and for this reason out-of-scope

libcudacxx/include/cuda/__type_traits/is_trivially_copyable.h
fbusato: no, only what is not handled by `cuda::std::is_trivially_copyable`

docs/libcudacxx/extended_api/type_traits/is_trivially_copyable.rst
fbusato: in general, we provide a godbolt link

libcudacxx/include/cuda/std/__bit/bit_cast.h
fbusato: no, we can call `_CCCL_BUILTIN_BIT_CAST` only in this case

docs/libcudacxx/extended_api/type_traits/is_trivially_copyable.rst
fbusato: @miscco  @davebayer  let me know if you have a different opinion

Jacobfaib: > I don't see a good motivation for user to specialize it

Even so, I still think the definition snippet should show the "usual" way these traits are defined, i.e.
```c++
template <typename T>
struct my_trait : bool_constant<...> { };

template <typename T>
inline constexper bool my_trait_v = my_trait<T>::value;
```
Even if that is not technically true, we don't want to give the impression that specializing the `_v` form is what actually matters.

docs/libcudacxx/extended_api/type_traits/is_trivially_copyable.rst
```diff
@@ -0,0 +1,75 @@
...
+
+   template <typename T>
+   using is_trivially_copyable = cuda::std::bool_constant<is_trivially_copyable_v<T>>;
+
+   } // namespace cuda
+
+``cuda::is_trivially_copyable_v`` trait evaluates if a type can be copied by copying its underlying bytes.
+It extends ``cuda::std::is_trivially_copyable`` to also recognize CUDA extended floating-point vector types as trivially copyable.
+
+``cuda::is_trivially_copyable_v<T>`` relies on ``cuda::std::is_trivially_copyable`` but adds support for CUDA-specific types.
```
pciolkosz: Note: This sentence and the previous one seem a bit redundant

libcudacxx/test/libcudacxx/std/numerics/bit/bit.cast/bit_cast_test_helpers.h
```diff
@@ -29,6 +29,7 @@ TEST_FUNC void test_roundtrip_through_buffer(T from)
   Buffer middle  = cuda::std::bit_cast<Buffer>(from);
   T to           = cuda::std::bit_cast<T>(middle);
   Buffer middle2 = cuda::std::bit_cast<Buffer>(to);
+  unused(middle2);
```
miscco: Nitpick: I would prefer `[[maybe_unused]]`

miscco (approved):

docs/libcudacxx/extended_api/type_traits/is_trivially_copyable.rst
miscco: NO we should never advertise that we use a bool_constant as the source of truth. IN contrary we must make clear that the inline variable is the source of truth

miscco (approved):

bernhardmgruber (approved):

c/parallel/src/transform.cu
```diff
@@ -327,7 +328,7 @@ static_assert(device_transform_policy()(detail::current_tuning_cc()) == {9}, "Ho
   build_ptr->cache                      = new transform::cache();
 
   // avoid new and delete which requires the allocated and freed types to match
-  static_assert(std::is_trivially_copyable_v<decltype(policy_sel)>);
+  static_assert(::cuda::is_trivially_copyable_v<decltype(policy_sel)>);
```
davebayer, line 331:
```suggestion
  static_assert(cuda::is_trivially_copyable_v<decltype(policy_sel)>);
```

c/parallel/src/transform.cu
```diff
@@ -528,7 +529,7 @@ static_assert(device_transform_policy()(detail::current_tuning_cc()) == {12}, "H
...
```
davebayer, line 532:
```suggestion
  static_assert(cuda::is_trivially_copyable_v<decltype(policy_sel)>);
```

davebayer (approved):
with a guilty conscience

c/parallel/src/transform.cu
fbusato: it was set as auto-merge...I need to open a new PR

PR #8266, opened 2026-04-01
[STF] Revamp STF C API: opaque handles for exec_place and data_place

caugonnet:
Replace tagged-union types (stf_exec_place, stf_data_place with enums and sub-structs) with opaque pointer handles (stf_exec_place_handle, stf_data_place_handle) backed by heap-allocated C++ objects.

- Add factory functions: stf_exec_place_host(), stf_data_place_device(), etc.
- Add lifecycle management: stf_exec_place_destroy(), stf_data_place_destroy()
- Add property queries: stf_exec_place_is_host(), stf_data_place_to_string(), etc.
- Add grid construction: stf_exec_place_grid_from_devices(), stf_exec_place_grid_create()
- Add composite placement: stf_data_place_composite()
- Update all STF functions to accept opaque handles directly
- Update test_places.cpp and test_logical_data_with_place.cu for new API

Made-with: Cursor

## Description

closes 

## Checklist

### DIFF
--- a/docs/404_helper.inc.html
+++ b/docs/404_helper.inc.html
@@ -36,24 +36,26 @@
                 const pagelist = text.split(",")
 
                 const options = {
-                includeScore: true,
-                distance: 100,
-                threshold: 0.8,
-                ignoreLocation: true
+                  includeScore: true,
+                  threshold: 0.8,
+                  ignoreLocation: false
                 }
 
                 // cccl/device_vector.html => 'cccl 'device 'vector
                 console.log("query:" + searchString)
                 const fuse = new Fuse(pagelist, options)
-                const result = fuse.search(searchString, {limit: 10})
+                // Include the ending token in the search to prioritize short matches
+                const result = fuse.search(searchString + "(end)", {limit: 10})
 
                 const listItem = '<li><a href=\"{2}{0}\">{0}</a> - Score: {1}</li>'.replace(/\{2\}/g, rootUrl)
                 document.getElementById('linkGen').innerHTML = ""
                 result.forEach(element => {
-                  document.getElementById('linkGen').innerHTML += listItem.replace(/\{0\}/g, element.item).replace(/\{1\}/g, element.score.toFixed(2))
+                  const targetUri = element.item.replace("(end)", "")
+                  const targetScore = element.score.toFixed(2)
+                  document.getElementById('linkGen').innerHTML += listItem.replace(/\{0\}/g, targetUri).replace(/\{1\}/g, targetScore)
                   // The match is near perfect, just link directly to it.
-                  if (element.score <= 0.02 ) {
-                    window.location.href = rootUrl + element.item
+                  if (targetScore <= 0.02 ) {
+                    window.location.href = rootUrl + targetUri
                   }
                 });
             })
--- a/docs/scrape_docs.bash
+++ b/docs/scrape_docs.bash
@@ -15,7 +15,8 @@ cd $SCRIPT_PATH
 
 pages=$(
         cd $path_to_docs;
-        find ./ -iname "*.html" -printf '/%P,'
+        # Embed a token to artificially limit search results
+        find ./ -iname "*.html" -printf '/%P(end),'
     )
 
 echo "$pages" > $path_to_docs/pagelist.txt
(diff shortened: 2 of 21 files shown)


c/experimental/stf/test/test_places.cpp
```diff
@@ -91,8 +92,9 @@ C2H_TEST("empty stf tasks", "[task]")
   stf_task_create(ctx, &t4);
   stf_task_set_symbol(t4, "T4");
   stf_task_add_dep(t4, lY, STF_READ);
-  auto d_place_dev0 = make_device_data_place(0);
-  stf_task_add_dep_with_dplace(t4, lZ, STF_RW, &d_place_dev0);
+  stf_data_place_handle d_place_dev0 = stf_data_place_device(0);
+  stf_task_add_dep_with_dplace(t4, lZ, STF_RW, d_place_dev0);
+  stf_data_place_destroy(d_place_dev0);
```
NaderAlAwar, line 100: Suggestion: tasks t1-t4 are never destroyed. It would make sense to add `stf_task_destroy`. This may apply to other tests as well

c/experimental/stf/src/stf.cu
```diff
@@ -8,66 +8,198 @@
...
+}
 } // namespace
 
 extern "C" {
 
-/* Convert the C-API stf_data_place to a C++ data_place object */
-static data_place to_data_place(const stf_data_place* data_p)
+stf_exec_place_handle stf_exec_place_host(void)
+{
+  return reinterpret_cast<stf_exec_place_handle>(new exec_place(exec_place::host()));
```
NaderAlAwar: Important: All factory functions can throw C++ exceptions from operator new across the `extern "C"` boundary. In cccl.c we wrap these with `try-catch` blocks to convert exceptions to `CUresult` error codes

c/experimental/stf/src/stf.cu
```diff
@@ -8,66 +8,198 @@
...
+  _CCCL_ASSERT(h != nullptr, "data_place handle must not be null");
+  return device_ordinal(*reinterpret_cast<data_place*>(h));
+}
+
+const char* stf_data_place_to_string(stf_data_place_handle h)
+{
+  _CCCL_ASSERT(h != nullptr, "data_place handle must not be null");
+  static thread_local ::std::string s;
+  s = reinterpret_cast<data_place*>(h)->to_string();
+  return s.c_str();
```
andralex, line 340: This is not working. It will return a dangling pointer. I'll work on a fix.

andralex: take that back, string is thread-local so it should be fine

c/experimental/stf/src/stf.cu
andralex: OK will do

c/experimental/stf/test/test_places.cpp
andralex: you got it

c/experimental/stf/test/test_cuda_kernel.cu
```diff
@@ -38,14 +38,11 @@ C2H_TEST("axpy with stf cuda_kernel", "[cuda_kernel]")
...
+  stf_ctx_handle ctx = stf_ctx_create();
+  REQUIRE(ctx != nullptr);
 
-  stf_logical_data_handle lX, lY;
-
-  double *X, *Y;
-  X = (double*) malloc(N * sizeof(double));
-  Y = (double*) malloc(N * sizeof(double));
+  double* X = (double*) malloc(N * sizeof(double));
+  double* Y = (double*) malloc(N * sizeof(double));
```
oleksandr-pavlyk, line 45: We need to use 

```
 std::unique_ptr<void, decltype(&free)> X_owner(malloc(N * sizeof(double)), free);
 std::unique_ptr<void, decltype(&free)> Y_owner(malloc(N * sizeof(double)), free);
 double *X = (double *) X_owner.get();
 double *Y = (double *) Y_owner.get();
```

and remove `free(X); free(Y);` lines at the end of the test to avoid memory leak on any test failure.

oleksandr-pavlyk: If tests are written as C, perhaps `CHECK` should be used instead of `REQUIRE`, and every check should be following by goto to a clean-up section.

PR #8267, opened 2026-04-01
Reenable nvhpc CI for cub and cudax

davebayer:


cub/cub/device/device_radix_sort.cuh
```diff
@@ -688,27 +688,23 @@ public:
...
-            d_values,
-            static_cast<offset_t>(num_items),
-            begin_bit,
-            end_bit,
-            is_overwrite_okay,
-            stream,
-            decomposer);
-        });
-    }
+    return detail::dispatch_with_env(env, [&]([[maybe_unused]] auto tuning, void* storage, size_t& bytes, auto stream) {
```
davebayer: We `static_assert` on `decomposer_check_t::value == true` above, so I removed the `if` completely

cudax/cmake/cudaxHeaderTesting.cmake
```diff
@@ -87,7 +89,10 @@ function(cudax_add_header_test label definitions)
       # the following line removed:
       HEADER_TEMPLATE "${cudax_SOURCE_DIR}/cmake/header_test.in.cu"
     )
-    target_link_libraries(${headertest_target} PUBLIC cudax.compiler_interface)
+    target_link_libraries(
+      ${headertest_target}
+      PUBLIC cudax.compiler_interface CUDA::cuda_driver
```
davebayer, line 94: STF uses driver APIs, but didn't link cuda driver

miscco (approved):

cub/cub/device/device_radix_sort.cuh
miscco: Please revert, that is meant to ensure that we do not instantiate all the machinery and create good error messages

miscco: I believe we can simply add `_CCCL_UNREACHABLE()` after the if

miscco: See https://github.com/NVIDIA/cccl/pull/7684/changes/3aae23b240968e075996baaa86b8cbdd3bf2f131

cudax/cmake/cudaxHeaderTesting.cmake
Jacobfaib: `cudax_stf_configure_target()` defined in `cudax/cmake/cudaxSTFConfigureTarget.cmake` should already do most of the things listed here (including linking against relevant CUDA targets). Could we use that instead here?

Also, shouldn't the `CUDA::cuda_driver` already be getting linked transitively by `libcudacxx` etc, since that also uses the driver these days?

davebayer: > `cudax_stf_configure_target()` defined in `cudax/cmake/cudaxSTFConfigureTarget.cmake` should already do most of the things listed here (including linking against relevant CUDA targets). Could we use that instead here?

I will try 

> Also, shouldn't the `CUDA::cuda_driver` already be getting linked transitively by `libcudacxx` etc, since that also uses the driver these days?

No, libcu++ dynamically links `libcuda.so` :)

davebayer: > `cudax_stf_configure_target()` defined in `cudax/cmake/cudaxSTFConfigureTarget.cmake` should already do most of the things listed here (including linking against relevant CUDA targets). Could we use that instead here?

I realized this wouldn't be that easy, because header tests don't use `cudax_stf_configure_target()` for them. I would prefer to keep this change, we can change how the header tests are configured in a separate PR

alliepiper: Likely would be fine to just pass the headertest targets to `cudax_stf_configure_target`. The goal of those functions is to avoid having to scatter these dependencies around, and instead have a single source of truth for what each project needs.

PR #8270, opened 2026-04-01
`cuda::is_bitwise_comparable`

fbusato:
## Description

Followup of the discussion in

- https://github.com/NVIDIA/cccl/pull/8254

The PR introduces `cuda::is_bitwise_comparable` to:

1. Allow user specification.
2. Disable comparison of types that are actually not bit bitwise comparable, namely extended floating-point types.

The new trait relies on `has_unique_object_representations_v` but is disabled the following cases:

- extended floating-point types.
- extended floating-point vector types.
- Raw arrays, `cuda::std::array` of extended floating-point.
- `cuda::std::pair` of extended floating-point.
- `cuda::std::tuple` of extended floating-point.
- Nested compositions of the above cases.

### Potentially affected paths by `cuda::std::is_bitwise_comparable`:

- `cuda::device::warp_match_all`


docs/libcudacxx/extended_api/type_traits/is_bitwise_comparable.rst
```diff
@@ -0,0 +1,73 @@
...
+   template <>
+   constexpr bool cuda::is_bitwise_comparable_v<MyType> = true;
+
+   static_assert(cuda::is_bitwise_comparable_v<MyType>);
+   static_assert(cuda::is_bitwise_comparable_v<const MyType>);
+
+
+... warning::
+
+    Users are responsible for ensuring that the type is actually bitwise comparable when specializing this variable template. Otherwise, the behavior is undefined.
```
jrhemstad: suggestion: We should mention that users should take extra care with custom types that may contain padding bits.

fbusato: thanks. I will probably add a list of conditions to "avoid"

miscco (approved):

bernhardmgruber (approved):

PR #8271, opened 2026-04-01
[cudax] Implement more CUDA graphs APIs

pciolkosz:
This PR throws away all graph node insertion APIs and instead adds __graphAddNode. The polymorphic cuGraphAddNode is the recommended way to add graph nodes and other APIs will get deprecated in the future.

It also adds a few new APIs:
- child graph
- copy/fill bytes
- conditional node
- event record/wait

### DIFF
--- a/cudax/include/cuda/experimental/__graph/conditional_node.cuh
+++ b/cudax/include/cuda/experimental/__graph/conditional_node.cuh
@@ -128,11 +128,11 @@ __make_conditional_node(path_builder& __pb, conditional_handle __handle, ::CUgra
 //! run conditionally before the parent graph is instantiated.
 //!
 //! \param __pb          Path builder to insert the node into.
-//! \param __default_val Initial handle value (1 = execute, 0 = skip). Ignored when
+//! \param __default_val Initial handle value (true = execute, false = skip). Ignored when
 //!                      \p __handle is provided.
 //! \return A `conditional_node_result` containing the node ref, body graph, and handle.
 //! \throws cuda::std::cuda_error if node creation fails.
-_CCCL_HOST_API inline conditional_node_result make_if_node(path_builder& __pb, unsigned int __default_val = 1)
+_CCCL_HOST_API inline conditional_node_result make_if_node(path_builder& __pb, bool __default_val = true)
 {
   conditional_handle __handle{__pb.get_graph(), __default_val};
   return __make_conditional_node(__pb, __handle, ::CU_GRAPH_COND_TYPE_IF);
@@ -155,14 +155,14 @@ _CCCL_HOST_API inline conditional_node_result make_if_node(path_builder& __pb, c
 //! is non-zero at the start of each iteration (including the first).
 //!
 //! The caller must populate the returned `body_graph` before instantiating the parent
-//! graph. The body is responsible for calling `handle.set_value(0) or handle.disable()`
+//! graph. The body is responsible for calling `handle.set_value(false)` or `handle.disable()`
 //! to terminate the loop.
 //!
 //! \param __pb          Path builder to insert the node into.
-//! \param __default_val Initial handle value (1 = enter loop, 0 = skip).
+//! \param __default_val Initial handle value (true = enter loop, false = skip).
 //! \return A `conditional_node_result` containing the node ref, body graph, and handle.
 //! \throws cuda::std::cuda_error if node creation fails.
-_CCCL_HOST_API inline conditional_node_result make_while_node(path_builder& __pb, unsigned int __default_val = 1)
+_CCCL_HOST_API inline conditional_node_result make_while_node(path_builder& __pb, bool __default_val = true)
 {
   conditional_handle __handle{__pb.get_graph(), __default_val};
   return __make_conditional_node(__pb, __handle, ::CU_GRAPH_COND_TYPE_WHILE);


cudax/test/graph/graph_node_ops_smoke.cu
```diff
@@ -0,0 +1,500 @@
...
+  // Host callback increments it.
+  cudax::host_launch(pb, [ptr]() {
+    *ptr += 1;
+  });
+
+  // Kernel verifies the final value is 43.
+  cudax::launch(pb, test::one_thread_dims, test::verify_n<43>{}, ptr);
+
+  auto exec = g.instantiate();
+  exec.launch(s);
```
caugonnet, line 260: You should have a test that verifies we can create a host callback like this and launch it multiple time (to ensure we properly destroy argument payload)

pciolkosz: Thanks! I added a test case like that

cudax/include/cuda/experimental/__graph/conditional_node.cuh
```diff
@@ -0,0 +1,187 @@
...
+  //! \throws cuda::std::cuda_error if `cudaGraphConditionalHandleCreate` fails.
+  _CCCL_HOST_API explicit conditional_handle(graph_builder_ref __graph, unsigned int __default_val = 1)
+      : __handle_(::cuda::experimental::__driver::__graphConditionalHandleCreate(
+          __graph.get(), __graph.get_device().primary_context(), __default_val, ::cudaGraphCondAssignDefault))
+  {}
+
+  //! \brief Sets the runtime value of the conditional handle from device code.
+  //!
+  //! \param __value Non-zero to execute the body, zero to skip.
+  _CCCL_DEVICE void set_value(unsigned int __value) const noexcept
```
davebayer:
```suggestion
  //! \param __default_val Initial value of the handle (1 = execute body, 0 = skip).
  //! \throws cuda::std::cuda_error if `cudaGraphConditionalHandleCreate` fails.
  _CCCL_HOST_API explicit conditional_handle(graph_builder_ref __graph, bool __default_val = true)
      : __handle_(::cuda::experimental::__driver::__graphConditionalHandleCreate(
          __graph.get(), __graph.get_device().primary_context(), __default_val, ::cudaGraphCondAssignDefault))
  {}

  //! \brief Sets the runtime value of the conditional handle from device code.
  //!
  //! \param __value Non-zero to execute the body, zero to skip.
  _CCCL_DEVICE void set_value(bool __value) const noexcept
```

cudax/include/cuda/experimental/__graph/conditional_node.cuh
```diff
@@ -0,0 +1,187 @@
...
+  {
+    ::cudaGraphSetConditional(__handle_, __value);
+  }
```
davebayer: Shouldn't we guard this with `_CCCL_HAS_DEVICE_RUNTIME()`?

libcudacxx/include/cuda/__device/device_ref.h
```diff
@@ -128,6 +128,12 @@ class device_ref
   //! @brief Initializes the primary context of the device.
   _CCCL_HOST_API void init() const; // implemented in <cuda/__device/physical_device.h> to avoid circular dependency
 
+  //! @brief Retrieve the primary context of this device.
+  //!
+  //! @return The primary CUDA context for this device.
+  [[nodiscard]] _CCCL_HOST_API ::CUcontext primary_context() const; // implemented in <cuda/__device/physical_device.h>
+                                                                    // to avoid circular dependency
```
davebayer: Wait, we are exposing this?

cudax/include/cuda/experimental/__driver/driver_api.cuh
```diff
@@ -36,145 +36,59 @@
...
 
-[[nodiscard]] _CCCL_HOST_API inline ::CUgraphNode __graphAddMemcpyNode1D(
-  ::CUgraph __graph,
-  const ::CUgraphNode* __deps,
-  ::cuda::std::size_t __ndeps,
-  ::CUdeviceptr __dst,
-  ::CUdeviceptr __src,
-  ::cuda::std::size_t __byte_count)
+[[nodiscard]] _CCCL_HOST_API inline ::CUgraphNode __graphAddNode(
+  ::CUgraph __graph, const ::CUgraphNode* __deps, ::cuda::std::size_t __ndeps, ::CUgraphNodeParams* __params)
```
Jacobfaib, line 50: Somewhat unrelated to the changes in this PR, but if you are going to version the CUDA driver calls, you also need to version any structure parameters like this explicitly. `CUgraphNodeParams` has a `_v1` and `_v2` IIRC

libcudacxx/include/cuda/__device/device_ref.h
pciolkosz: Good point, made it `__primary_context` for now, we can think about exposing it later

cudax/include/cuda/experimental/__driver/driver_api.cuh
pciolkosz: I don't think `CUgraphNodeParams` was versioned yet, but I will keep it in mind, some substructures of it could be versioned

cudax/include/cuda/experimental/__graph/conditional_node.cuh
pciolkosz: This is a syscall so should be fine I think

PR #8269, opened 2026-04-01
Add public any_cast for type-erased memory resources

bdice:
## Description

closes https://github.com/NVIDIA/cccl/issues/7994

Add `cuda::mr::any_cast<Resource>` overloads (mutable and const pointer variants) for all 4 type-erased memory resource wrappers: `any_resource`, `any_synchronous_resource`, `resource_ref`, and `synchronous_resource_ref`. This enables users to recover the concrete resource type from type-erased wrappers, analogous to `std::any_cast` for `std::any`.

Each overload is constrained with `resource_with` / `synchronous_resource_with` to ensure the cast target satisfies both the resource concept and all required properties, and delegates to the internal `__any_cast` on the `__basic_any` base class.

## Checklist
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #8272, opened 2026-04-02
[libcu++] Add resource_cast and dynamic_resource_cast

pciolkosz:
This PR adds `resource_cast` and `dynamic_resource_cast` wrapper for internal `__any_cast` and `__dynamic_any_cast` specialized for resource wrappers.

I also added validation to `__dynamic_any_cast` implementation, which was right now succeeding regardless of what the destination interface was. Now it confirms destination interface is a subset of the original vtable. Tests were added to `__basic_any` testing to confirm this behavior

### DIFF
--- a/libcudacxx/test/libcudacxx/cuda/utility/basic_any.pass.cpp
+++ b/libcudacxx/test/libcudacxx/cuda/utility/basic_any.pass.cpp
@@ -726,11 +726,9 @@ _CCCL_HOST void test_iset_dynamic_cast()
 
 int main(int, char**)
 {
-  NV_IF_TARGET(NV_IS_HOST,
-               (test_basic_any<SmallType>(); test_basic_any<LargeType>();
+  NV_IF_TARGET(NV_IS_HOST, (test_basic_any<SmallType>(); test_basic_any<LargeType>();))
 #if !_CCCL_COMPILER(NVRTC)
-                test_iset_dynamic_cast();
+  NV_IF_TARGET(NV_IS_HOST, (test_iset_dynamic_cast();))
 #endif // !_CCCL_COMPILER(NVRTC)
-                ))
   return 0;
 }


bdice:
I did a bit of comparison between #8269 and this PR.
- Concept constraints. #8269 constrains every overload with `resource_with` / `synchronous_resource_with`. This PR's `resource_cast` is unconstrained, any `_Tp` is accepted. I wonder if this should be a compile-time error for types that don't satisfy the resource concept.
- Does this PR have enough Doxygen? #8269 has a bit more, I'm not sure what is expected.
- Test gaps compared to #8269:
  - `nullptr` input (covered by `__any_cast` internally, but seems untested in the public API)
  - moved-from wrapper returns `nullptr`
  - I'd like explicit tests that succeed or fail for derived-to-base casts, we have lots of MR inheritance in RMM
- Naming. `cuda::mr::any_cast` mirrors `std::any_cast`. Does this function have similar semantics to `std::any_cast`?

pciolkosz:
>Concept constraints. https://github.com/NVIDIA/cccl/pull/8269 constrains every overload with resource_with / synchronous_resource_with. This PR's resource_cast is unconstrained, any _Tp is accepted. I wonder if this should be a compile-time error for types that don't satisfy the resource concept.

I added static_asserts inside

>Does this PR have enough Doxygen? https://github.com/NVIDIA/cccl/pull/8269 has a bit more, I'm not sure what is expected.

I think its fine

>Test gaps compared to https://github.com/NVIDIA/cccl/pull/8269:
nullptr input (covered by __any_cast internally, but seems untested in the public API)
moved-from wrapper returns nullptr
I'd like explicit tests that succeed or fail for derived-to-base casts, we have lots of MR inheritance in RMM

Added test cases

>Naming. cuda::mr::any_cast mirrors std::any_cast. Does this function have similar semantics to std::any_cast?

I think because we also have the `dynamic_resource_cast` that has no equivalent in stl and the interface is restricted only to specifically resource wrappers I would stay with `cuda::mr::resource_cast` and `cuda::mr::dynamic_resource_cast`

PR #8276, opened 2026-04-02
Replace thrust::discard_iterator with cuda::discard_iterator

Jacobfaib:
## Description

Replaces `thrust::discard_iterator` with `cuda::discard_iterator`

## Checklist

### DIFF
--- a/thrust/thrust/iterator/permutation_iterator.h
+++ b/thrust/thrust/iterator/permutation_iterator.h
@@ -32,6 +32,9 @@
 #include <thrust/iterator/iterator_adaptor.h>
 #include <thrust/iterator/iterator_traits.h>
 
+#include <cuda/std/__type_traits/is_same.h>
+#include <cuda/std/__type_traits/remove_cvref.h>
+
 THRUST_NAMESPACE_BEGIN
 
 template <typename, typename>
@@ -48,7 +51,7 @@ struct make_permutation_iterator_base
   // We need this to make proxy iterators work because those have a void reference/value type
   using iterator_value_t =
     ::cuda::std::conditional_t<::cuda::std::is_same_v<it_value_t<ElementIterator>, void>,
-                               ::cuda::std::iter_value_t<ElementIterator>,
+                               ::cuda::std::remove_cvref_t<decltype(*::cuda::std::declval<ElementIterator>())>,
                                it_value_t<ElementIterator>>;
 
   using iterator_reference_t =


thrust/testing/discard_iterator.cu
miscco, line 1: please revert as those are the original `thrust::discard_iterator` tests

Jacobfaib: Done

cub/cub/device/device_transform.cuh
```diff
@@ -433,7 +434,7 @@ struct DeviceTransform
   {
     static_assert(::cuda::std::is_invocable_v<Generator>, "The passed generator must be a nullary function object");
     static_assert(
-      ::cuda::std::is_assignable_v<detail::it_reference_t<RandomAccessIteratorOut>,
+      ::cuda::std::is_assignable_v<::cuda::std::iter_reference_t<RandomAccessIteratorOut>,
```
bernhardmgruber, line 437: Important: This is a change in semantics, why is this necessary?

See: https://github.com/NVIDIA/cccl/blob/0969ad0fc449b91d6ba8bbe4a3f3d12e574a12c4/cub/cub/util_type.cuh#L54-L69

cub/test/catch2_test_nvrtc.cu
```diff
@@ -92,7 +92,7 @@ TEST_CASE("Test nvrtc", "[test][nvrtc]")
 
     #include <thrust/iterator/constant_iterator.h>
     #include <thrust/iterator/counting_iterator.h>
-    #include <thrust/iterator/discard_iterator.h>
+    #include <cuda/iterator>
```
bernhardmgruber, line 95: Important: Please revert. We want to continue to test the old iterator with NVRTC.

cub/cub/device/dispatch/kernels/kernel_scan_warpspeed.cuh
```diff
@@ -935,8 +936,8 @@ _CCCL_API constexpr bool use_warpspeed(
 template <typename InputIteratorT, typename OutputIteratorT, typename AccumT>
 _CCCL_API constexpr bool use_warpspeed(const scan_warpspeed_policy& policy)
 {
-  using InputT  = it_value_t<InputIteratorT>;
-  using OutputT = it_value_t<OutputIteratorT>;
+  using InputT  = non_void_value_t<InputIteratorT, it_value_t<OutputIteratorT>>;
+  using OutputT = non_void_value_t<OutputIteratorT, it_value_t<InputIteratorT>>;
```
bernhardmgruber, line 940: Important: This is probably wrong, since we cannot just take the traits of the input type in case we have no output type.

Why is this necessary? Which test is failing?

Jacobfaib: `cuda::discard_iterator::value_type` is `void`, while `thrust::discard_iterator::value_type` is not. Many CUB algorithms seemingly don't handle this properly and/or had special handling for `thrust::discard_iterator` (or other proxy iterators) and so failed to compile with the new discard iterator.

cub/cub/device/device_transform.cuh
Jacobfaib: See the other comment. There is a change in `value_type` between the iterators which is not handled by CUB.

Jacobfaib:
@bernhardmgruber @miscco given that the `cuda::discard_iterator` has some fundamental mismatches with `thrust::discard_iterator`, I'm not sure we can wholesale replace the former with the latter, especially because CUB doesn't play nice with it.

I propose we table this PR until CCCL 4.0.

bernhardmgruber:
> @bernhardmgruber @miscco given that the `cuda::discard_iterator` has some fundamental mismatches with `thrust::discard_iterator`, I'm not sure we can wholesale replace the former with the latter, especially because CUB doesn't play nice with it.
> 
> I propose we table this PR until CCCL 4.0.

We should solve it before CCCL 4.0 I think, because we want to deprecate Thrust iterators several releases before 4.0 and also be sure that CUB works when users start feeding it with CUDA iterators.

Can we slim down this PR and merge pieces which are working?

bernhardmgruber:
Looking through the changeset, can we upstream some of the unit test changes?

Jacobfaib:
Ah sorry these got lost in my emails. We can resurrect some of the changes but it would require a lot of silence macros. Let me resurrect this PR and see what can be saved

PR #8278, opened 2026-04-02
Reenable nvhpc CI for thrust

davebayer:


ci/matrix.yaml
miscco, line 1: There are some other run that are still disabled below, please reenable

PR #8059, opened 2026-03-17
Add env overloads for InclusiveSum and future ExclusiveScan

gonidelis:
Adds miscellaneous env overloads for `InclusiveSum` and `ExclusiveScan` with `FutureValue`

PR #8280, opened 2026-04-02
Improve DeviceScan env test coverage and clean up stale code

gonidelis:
Taking the opportunity from [a recent review from one of my env PRs](https://github.com/NVIDIA/cccl/pull/8059#discussion_r3011241854) I took the liberty to clean up the scan tests mainly for two reasons:

1. use non-identity `init` values to make sure they are factored in properly
2. apply scan/sum ops in more than 1 elements (which in `exclusive*` it's just the `init` value) to make sure without loss of generality that the results is correct for the full input range and not just the edge case
3. clean-up some dead leftover code
4. Add `block_size_recording_iterator_t` for `ExclusiveSum` tuning test since `ExclusiveSum` doesn't accept a custom op, we use a custom input iterator that records `blockDim.x` as a side effect, allowing us to verify the tuning policy is applied
5. Guard (previously unguarded) `blockDim.x` writes in `block_size_check_t` with `if (threadIdx.x == 0)`

### DIFF
--- a/cub/test/catch2_test_device_scan.cuh
+++ b/cub/test/catch2_test_device_scan.cuh
@@ -168,7 +168,7 @@ void compute_inclusive_scan_by_key_reference(
     host_values.cbegin(), host_keys.cbegin(), result_out_it, scan_op, equality_op, num_items);
 }
 
-struct block_size_recording_iterator_t
+struct block_size_recording_constant_iterator
 {
   using value_type        = int;
   using reference         = int;
@@ -180,7 +180,7 @@ struct block_size_recording_iterator_t
   int* block_size_ptr;
   difference_type offset;
 
-  __host__ __device__ block_size_recording_iterator_t(int val, int* bs_ptr, difference_type off = 0)
+  __host__ __device__ block_size_recording_constant_iterator(int val, int* bs_ptr, difference_type off = 0)
       : value(val)
       , block_size_ptr(bs_ptr)
       , offset(off)
@@ -204,28 +204,28 @@ struct block_size_recording_iterator_t
     return value;
   }
 
-  __host__ __device__ block_size_recording_iterator_t operator+(difference_type n) const
+  __host__ __device__ block_size_recording_constant_iterator operator+(difference_type n) const
   {
     return {value, block_size_ptr, offset + n};
   }
 
-  __host__ __device__ block_size_recording_iterator_t& operator+=(difference_type n)
+  __host__ __device__ block_size_recording_constant_iterator& operator+=(difference_type n)
   {
     offset += n;
     return *this;
   }
 
-  __host__ __device__ difference_type operator-(const block_size_recording_iterator_t& other) const
+  __host__ __device__ difference_type operator-(const block_size_recording_constant_iterator& other) const
   {
     return offset - other.offset;
   }
 
-  __host__ __device__ bool operator==(const block_size_recording_iterator_t& other) const
+  __host__ __device__ bool operator==(const block_size_recording_constant_iterator& other) const
   {
     return offset == other.offset;
   }
 
-  __host__ __device__ bool operator!=(const block_size_recording_iterator_t& other) const
+  __host__ __device__ bool operator!=(const block_size_recording_constant_iterator& other) const
   {
     return offset != other.offset;
   }
--- a/cub/test/catch2_test_device_scan_env.cu
+++ b/cub/test/catch2_test_device_scan_env.cu
@@ -153,7 +153,7 @@ C2H_TEST("Device scan exclusive-sum can be tuned", "[scan][device]", block_sizes
   c2h::device_vector<int> d_block_size(1, 0);
   // use block_size_recording_iterator to embed blockDim info in the input type and query after
   // since ExclusiveSum can not take a custom scan_op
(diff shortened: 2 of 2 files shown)


cub/test/catch2_test_device_scan_env.cu
```diff
@@ -82,21 +80,15 @@ TEST_CASE("Device scan exclusive sum works with default environment", "[sum][dev
 {
   using num_items_t = int;
   using value_t     = int;
-  using offset_t    = cub::detail::choose_offset_t<num_items_t>;
-
-  int current_device{};
-  REQUIRE(cudaSuccess == cudaGetDevice(&current_device));
```
gonidelis, line 88: @srinivasyadav18 what was the reason for having this? what does it check? blocking my own PR till your approval.

cub/test/catch2_test_device_scan_env.cu
```diff
@@ -82,21 +80,15 @@ TEST_CASE("Device scan exclusive sum works with default environment", "[sum][dev
...
 
-  int ptx_version{};
-  REQUIRE(cudaSuccess == cub::PtxVersion(ptx_version, current_device));
-
```
gonidelis, line 92: same as above @srinivasyadav18

cub/test/catch2_test_device_scan_env.cu
```diff
@@ -138,15 +130,16 @@ C2H_TEST("Device scan exclusive-scan can be tuned", "[scan][device]", block_size
...
   // We are expecting that `unrelated_tuning` is ignored
   auto env = cuda::execution::__tune(scan_tuning<target_block_size>{}, unrelated_tuning{});
 
-  REQUIRE(cudaSuccess == cub::DeviceScan::ExclusiveScan(d_in, d_out.begin(), block_size_check, 0, num_items, env));
+  REQUIRE(cudaSuccess == cub::DeviceScan::ExclusiveScan(d_in, d_out.begin(), block_size_check, init, num_items, env));
 
   for (int i = 0; i < num_items; i++)
   {
-    REQUIRE(d_out[i] == i);
+    REQUIRE(d_out[i] == i + init);
```
gonidelis: use a thrust::equals(d_out, counting_iterator(init))

gonidelis: don't teach llms how to write bad cuda code

cub/test/catch2_test_device_scan_env.cu
```diff
@@ -57,22 +57,20 @@ TEST_CASE("Device scan exclusive scan works with default environment", "[scan][d
...
-  REQUIRE(cudaSuccess == cudaGetDevice(&current_device));
-
   cuda::arch_id arch_id;
   REQUIRE(cudaSuccess == cub::detail::ptx_arch_id(arch_id));
   const auto target_block_size = selector_t{}(arch_id).block_threads;
 
-  num_items_t num_items = 1;
+  num_items_t num_items = 2;
   c2h::device_vector<int> d_block_size(1);
   block_size_check_t block_size_check{thrust::raw_pointer_cast(d_block_size.data())};
```
gonidelis, line 70: modify the functor with `if threadId.x == 0` so there is no data race

cub/test/catch2_test_device_scan_env.cu
```diff
@@ -168,6 +161,9 @@ C2H_TEST("Device scan exclusive-sum can be tuned", "[scan][device]", block_sizes
   {
     REQUIRE(d_out[i] == i);
   }
+
```
gonidelis: _we could use the `stream_registry_factory_t` for that but that might end up getting too involved_

let's just create a custom `IputTypeT` for d_in that stores `blockDim.x` info internally and we can poll it from there

cub/test/catch2_test_device_scan_env.cu
```diff
@@ -368,6 +361,6 @@ C2H_TEST("Device scan inclusive-scan-init uses environment", "[scan][device]")
 
   for (int i = 0; i < num_items; i++)
   {
-    REQUIRE(d_out[i] == (i + 1 + 10.0f));
+    REQUIRE(d_out[i] == (i + 1 + init));
```
gonidelis: use thrust::equal as above

cub/test/catch2_test_device_scan_env.cu
```diff
@@ -292,7 +285,7 @@ C2H_TEST("Device scan exclusive-scan uses environment", "[scan][device]")
 
   for (int i = 0; i < num_items; i++)
   {
-    REQUIRE(d_out[i] == i);
+    REQUIRE(d_out[i] == i + init);
```
gonidelis: same, use `thrust::equal`

cub/test/catch2_test_device_scan.cuh
```diff
@@ -167,3 +167,66 @@ void compute_inclusive_scan_by_key_reference(
   compute_inclusive_scan_by_key_reference(
     host_values.cbegin(), host_keys.cbegin(), result_out_it, scan_op, equality_op, num_items);
 }
+
+struct block_size_recording_iterator_t
```
bernhardmgruber: Suggestion: consider renaming this to `block_size_recording_constant_iterator[_t]`. This helps understanding more quickly what this class is about.

bernhardmgruber (approved):

PR #8281, opened 2026-04-02
Enable `performance-*` in `clang-tidy`.

Jacobfaib:
## Description

closes 

## Checklist

### DIFF
--- a/c/parallel/test/test_util.h
+++ b/c/parallel/test/test_util.h
@@ -917,7 +917,8 @@ struct name_source_t
 };
 
 template <class ValueT, class StateT>
-iterator_t<ValueT, StateT> make_iterator(name_source_t state, operation_t advance, operation_t dereference)
+iterator_t<ValueT, StateT>
+make_iterator(const name_source_t& state, const operation_t& advance, const operation_t& dereference)
 {
   iterator_t<ValueT, StateT> it;
   it.state_name                = state.name;
@@ -977,7 +978,10 @@ inline std::tuple<std::string, std::string, std::string> make_random_access_iter
 
 template <class ValueT>
 iterator_t<ValueT, random_access_iterator_state_t<ValueT>> make_random_access_iterator(
-  iterator_kind kind, std::string_view value_type, std::string prefix = "", std::string transform = "")
+  iterator_kind kind,
+  std::string_view value_type,
+  std::string prefix    = "",
+  std::string transform = "") // NOLINT(performance-unnecessary-value-param)
 {
   std::string iterator_state_name = std::format("{0}state_t", prefix);
   std::string advance_fn_name     = std::format("{0}advance", prefix);
--- a/cub/cub/agent/agent_scan.cuh
+++ b/cub/cub/agent/agent_scan.cuh
@@ -244,8 +244,11 @@ struct AgentScan
   //---------------------------------------------------------------------
 
   template <bool Inclusive = IS_INCLUSIVE>
-  _CCCL_DEVICE _CCCL_FORCEINLINE void
-  ScanFirstTile(AccumT (&items)[ITEMS_PER_THREAD], InitValueT init_value, ScanOpT scan_op, AccumT& block_aggregate)
+  _CCCL_DEVICE _CCCL_FORCEINLINE void ScanFirstTile(
+    AccumT (&items)[ITEMS_PER_THREAD],
+    InitValueT init_value, // NOLINT(performance-unnecessary-value-param)
+    ScanOpT scan_op,
+    AccumT& block_aggregate)
   {
     BlockScanT blockScan(temp_storage.scan_storage.scan);
     if constexpr (Inclusive)
@@ -303,7 +306,11 @@ struct AgentScan
    *   Initial value to seed the exclusive scan
    */
   _CCCL_DEVICE _CCCL_FORCEINLINE AgentScan(
-    TempStorage& temp_storage, InputIteratorT d_in, OutputIteratorT d_out, ScanOpT scan_op, InitValueT init_value)
+    TempStorage& temp_storage,
+    InputIteratorT d_in,
+    OutputIteratorT d_out,
+    ScanOpT scan_op,
+    InitValueT init_value) // NOLINT(performance-unnecessary-value-param)
       : temp_storage(temp_storage.Alias())
       , d_in(d_in)
       , d_out(d_out)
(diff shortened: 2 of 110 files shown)


thrust/examples/arbitrary_transformation.cu
```diff
@@ -72,10 +72,10 @@ int main()
                    arbitrary_functor1());
 
   // print the output
-  std::cout << "Tuple functor" << std::endl;
+  std::cout << "Tuple functor" << '\n';
```
davebayer, line 75: do we actually care about these? I feel like they make it harder to read

libcudacxx/include/cuda/std/__tuple_dir/tuple.h
```diff
@@ -284,7 +284,7 @@ class _CCCL_TYPE_VISIBILITY_DEFAULT tuple
     __base_.swap(__t.__base_);
   }
 
-  _CCCL_API friend void swap(tuple& __t, tuple& __u)
+  _CCCL_API friend void swap(tuple& __t, tuple& __u) noexcept
   {
     __t.swap(__u);
   }
```
davebayer: I don't think this is correct

libcudacxx/include/cuda/std/__ranges/compressed_movable_box.h
```diff
@@ -874,7 +874,7 @@ struct _CCCL_DECLSPEC_EMPTY_BASES __compressed_movable_box<_Elem1, _Elem2, _Elem
   }
 
   _CCCL_EXEC_CHECK_DISABLE
-  _CCCL_API friend constexpr void swap(__compressed_movable_box& __x, __compressed_movable_box& __y)
+  _CCCL_API friend constexpr void swap(__compressed_movable_box& __x, __compressed_movable_box& __y) noexcept
```
davebayer: I'm not sure whether this is correct, @miscco can you have a look, please

Jacobfaib:
> I'm not sure if we can just slap noexcept to many of these functions, we must be extremely careful, it can lead to program termination..

For move constructors/assignment operators at least, unless these functions are incredibly carefully implemented to have strong exception guarantees (which it does not appear a lot of them are), program termination is arguably preferable to leaving zombie objects.

Similarly with swap which usually do multi-step swapping of members. In those cases, is it really preferable (unless we are otherwise constrained by the standard) to propagate the exception at the expense of leaving the object half swapped-from?

libcudacxx/include/cuda/__algorithm/copy.h
```diff
@@ -37,7 +38,7 @@
 _CCCL_BEGIN_NAMESPACE_CUDA
 
 //! @brief Source access order for copy_bytes
-enum class source_access_order
+enum class source_access_order : ::cuda::std::uint8_t
```
davebayer, line 41: This an ABI break

libcudacxx/include/cuda/__stream/invalid_stream.h
```diff
@@ -35,7 +35,7 @@ _CCCL_GLOBAL_CONSTANT invalid_stream_t invalid_stream{};
 
 [[nodiscard]] _CCCL_API _CCCL_FORCEINLINE ::cudaStream_t __invalid_stream() noexcept
 {
-  return reinterpret_cast<::cudaStream_t>(~0ull);
+  return ::cudaStream_t{};
```
davebayer: This is not correct, please, revert this

Jacobfaib: Ah good catch, I missed the `~` in `~0ull` when reading this. Fixed.

Jacobfaib:
Reviewers:

Please review this PR **by commit**. Each commit is split up into fixes for each specific check. This is so that any checks that we decide we don't want can be quickly and easily disabled/reverted.

By far and away though, the largest commit is the last one, performance-unnecessary-value-param so please take extra care when reviewing that one.

To silence that check you can either:

- Make arguments a const-ref
- Move or forward the argument later on
- Silence the check

My policy for fixing this was as follows:

- If the API is usually used on device (or is device only), silence the check. Promoting by-value arguments to references might lead to missed optimizations from nvcc on device as it may not be able to prove that pointers are restricted. Also overwhelming majority of on-device data structures are PODs so move makes no difference for perf (not to mention correctness issues if e.g. init values come from inside a container to be modified).
- If the API is host-only, or majority host, move or forward the arguments.
- If moving is not possible or "hard" (big function, easy to inadvertently use after move), silence the diagnostic. In extremely rare cases (test code), where I could prove that it was host-only, I made it a const-ref.

The upshot is that this was almost never fixed via const-ref. The vast majority of violations in CUB are silenced, and a lot of them are fixed via moves in thrust.

bernhardmgruber:
@Jacobfaib This work is great! However, the PR is too long and the changes to critical to review in one go. Please split it up.

* `performance-avoid-endl` is less critical but a lot of changes -> 1 PR
* `noexcept` move and swap is super critical, we need to think about every line -> 1 PR
* `enum-size` is ABI breaking in some places and has performance implications, may need benchmarks -> 1 PR
* `faster-string-find`, `for-range-copy`, `inefficient-vector-operations` are trivial -> 1 PR and easy review
* `performance-unnecessary-value-param` is very noisy -> 1 PR
* the other view could go to yet another PR

c/parallel/test/algorithm_execution.h
```diff
@@ -189,7 +189,7 @@ struct BuildResultDeleter
   {
     if (status != CUDA_SUCCESS)
     {
-      std::cerr << "Clean-up call returned status " << status << std::endl;
+      std::cerr << "Clean-up call returned status " << status << '\n';
```
caugonnet, line 192: why is this a better thing ?

c/parallel/test/test_util.h
```diff
@@ -868,7 +868,7 @@ struct iterator_t
   }
 };
 
-enum class iterator_kind
+enum class iterator_kind : std::uint8_t
```
caugonnet, line 871: cuda::std::uint8_t ?

c/parallel/test/algorithm_execution.h
davebayer: Because `std::endl` also flushes the stream which adds some overhead, `'\n' doesn't

Jacobfaib:
@bernhardmgruber https://github.com/NVIDIA/cccl/pull/8300 is ready for review. It splits out the trivial checks and has some cmake/.clang-tidy changes so should go first. Once that merges I will do the others.

PR #8283, opened 2026-04-02
Add additional warning flags for dev builds not enabled by Wall or Wextra

Jacobfaib:
## Description

This crucially also adds
- `-ftemplate-backtrace-limit=0`, so CI no longer truncates template expansions in error logs.
- `-fmacro-backtrace-limit=0`, similarly but for macro expansions.

## Checklist

### DIFF
--- a/cudax/test/execution/test_stream_context.cu
+++ b/cudax/test/execution/test_stream_context.cu
@@ -14,6 +14,8 @@
 // Then include the test helpers
 #include <thrust/equal.h>
 
+#include <cuda/std/cstddef>
+
 #include <cuda/experimental/container.cuh>
 #include <cuda/experimental/memory_resource.cuh>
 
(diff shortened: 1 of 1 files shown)


libcudacxx/include/cuda/std/__cccl/attributes.h
```diff
@@ -78,7 +78,7 @@
 // NVCC does not properly respect [[assume()]], so use __builtin_assume, see nvbug5458663
 #if _CCCL_CUDA_COMPILER(NVCC) && _CCCL_DEVICE_COMPILATION()
 #  define _CCCL_ASSUME(...) __builtin_assume(__VA_ARGS__)
-#elif _CCCL_HAS_CPP_ATTRIBUTE(assume)
+#elif _CCCL_HAS_CPP_ATTRIBUTE(assume) && (_CCCL_STD_VER >= 2023)
 #  define _CCCL_ASSUME(...) [[assume(__VA_ARGS__)]]
```
bernhardmgruber, line 82: Q: Why shouldn't we use the attribute if the compiler supports it before C++23?

Jacobfaib: `clang` warns with `-Wc++23-compat`. I consider this a bug, TBH, that it allows you to use `[[assume]]` but tells you off for doing so.

bernhardmgruber: Do we care about the `-Wc++23-compat` warning?

Jacobfaib: Until we start comprehensively testing C++23, I think we should heed the warning. Our CI matrix only goes to C++20 for now.

Note this warning actually only fired due to a previous version of this PR where I also had `-pedantic` enabled, but that resulted in _so_ many warnings that I turned it off and hope that someone else eventually becomes motivated enough to fix that :)

Issue #8277, opened 2026-04-02
[cuda.compute]: Update `segmented_reduce` to accept `max_segment_size` argument


shwina:
As a follow-up to https://github.com/NVIDIA/cccl/pull/7718

PR #8284, opened 2026-04-02
Expose `max_segment_size` guarantee in cuda.compute

shwina:
## Description

Closes https://github.com/NVIDIA/cccl/issues/8277.

Adds an optional `max_segment_size` parameter to `segmented_reduce` which can be used by the implementation to dispatch to the optimal kernel for the given segment size.

An informal benchmark to show speedup from specifying the `max_segment_size` ([code](https://gist.github.com/shwina/3823d422b4dfcc9bf2f4151a0a85523a)). As expected, it only really makes a difference for small and medium segment sizes, as large segments benefit from the default (block per segment):

```
  seg_size      num_segs    no hint (ms)     hint (ms)   speedup
--------------------------------------------------------------------
         1    67,108,864           73.03          1.85    39.44x
         4    16,777,216           18.16          0.70    25.97x
        16     4,194,304            4.57          0.41    11.05x
        64     1,048,576            1.16          0.33     3.47x
       256       262,144            0.35          0.32     1.09x
     1,024        65,536            0.31          0.31     1.00x
     4,096        16,384            0.31          0.31     1.00x
    16,384         4,096            0.31          0.31     1.00x
    65,536         1,024            0.31          0.31     1.00x
```

## Checklist

### DIFF
--- a/python/cuda_cccl/benchmarks/compute/quick_configs.yaml
+++ b/python/cuda_cccl/benchmarks/compute/quick_configs.yaml
@@ -117,13 +117,6 @@ segmented_sort/keys:
       "Elements{io}": 22
       "MaxSegmentSize": 10
 
-segmented_reduce/sum:
-  benchmarks:
-    small:
-      "T{ct}": "I32"
-      "Elements{io}": 16
-      "SegmentSize": 1
-
 segmented_reduce/variable_sum:
   benchmarks:
     variable_default:
--- a/python/cuda_cccl/benchmarks/compute/run_benchmarks.py
+++ b/python/cuda_cccl/benchmarks/compute/run_benchmarks.py
@@ -74,15 +74,27 @@ SUPPORTED_BENCHMARKS = [
     "merge_sort/keys",
     "merge_sort/pairs",
     "segmented_sort/keys",
-    "segmented_reduce/sum",
     "segmented_reduce/variable_sum",
     "partition/three_way",
 ]
 
 # Axes that use power-of-two values (need [pow2] suffix for nvbench)
 # These are the base names (without {ct}/{io} suffixes)
-POW2_AXES_CPP = {"Elements", "GuaranteedMaxSegSize", "MaxSegSize", "MaxSegmentSize", "SegmentSize", "Segments"}
-POW2_AXES_PY = {"Elements", "GuaranteedMaxSegSize", "MaxSegSize", "MaxSegmentSize", "Segments"}
+POW2_AXES_CPP = {
+    "Elements",
+    "GuaranteedMaxSegSize",
+    "MaxSegSize",
+    "MaxSegmentSize",
+    "SegmentSize",
+    "Segments",
+}
+POW2_AXES_PY = {
+    "Elements",
+    "GuaranteedMaxSegSize",
+    "MaxSegSize",
+    "MaxSegmentSize",
+    "Segments",
+}
 
 # Axis name mappings from C++ to Python.
 # Keep type axes in their C++ form (`{ct}`) to match benchmark axis names.


python/cuda_cccl/cuda/compute/algorithms/_segmented_reduce.py
```diff
@@ -164,6 +170,7 @@ def segmented_reduce(
     op: Operator,
     h_init: np.ndarray | GpuStruct,
     num_segments: int,
+    max_segment_size: int | None = None,
```
NaderAlAwar, line 173: Question: Did you consider making this a kwarg, like we did for determinism in reduce? My worry with this approach is that the user would have to specifythe max_segment_size to `None` if they want to pass a strema

c/parallel/include/cccl/c/segmented_reduce.h
```diff
@@ -72,11 +72,12 @@ CCCL_C_API CUresult cccl_device_segmented_reduce(
   size_t* temp_storage_bytes,
   cccl_iterator_t d_in,
   cccl_iterator_t d_out,
-  uint64_t num_offsets,
+  uint64_t num_segments,
   cccl_iterator_t start_offset_in,
   cccl_iterator_t end_offset_in,
   cccl_op_t op,
   cccl_value_t init,
+  size_t max_segment_size,
```
NaderAlAwar, line 80: Important: please add cccl.c tests that exercise the new `max_segment_size` paths

python/cuda_cccl/benchmarks/compute/segmented_reduce/sum.py
```diff
@@ -0,0 +1,142 @@
...
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+"""
+Python benchmark for segmented_reduce (sum) with fixed-size segments.
+
+C++ equivalent: cub/benchmarks/bench/segmented_reduce/sum.cu (uses base.cuh)
+
+Notes:
+- Implements three sub-benchmarks: small, medium, large (by SegmentSize)
+- The C++ equivalent uses DispatchFixedSizeSegmentedReduce; the Python API uses
```
NaderAlAwar: Important: I would remove this benchmark because it is meant to showcase the fixed size segmented reduce which we do not expose. We want to show parity with C++ performance, and at present this benchmark will likely not do that.

python/cuda_cccl/benchmarks/compute/segmented_reduce/variable_sum.py
NaderAlAwar, line 1: Important: we should add this benchmark to `run_benchmarks.py` and `quick_configs.yaml`

python/cuda_cccl/benchmarks/compute/segmented_reduce/variable_sum.py
```diff
@@ -0,0 +1,165 @@
...
+
+import sys
+from pathlib import Path
+
+sys.path.insert(0, str(Path(__file__).parent.parent))
+
+import cupy as cp
+import numpy as np
+from utils import (
+    FUNDAMENTAL_TYPES as TYPE_MAP,
```
NaderAlAwar: Important: the C++ benchmark only uses int32/64 and float32/64, so we should match the exact types used there instead of all of them. You can see how we do this in nondeterministic.py:

```python
TYPE_MAP = {k: _ALL_TYPES[k] for k in ("I32", "I64", "F32", "F64")}
```

python/cuda_cccl/benchmarks/compute/segmented_reduce/sum.py
shwina: Hmm - we have https://github.com/NVIDIA/cccl/blob/main/cub/benchmarks/bench/segmented_reduce/sum.cu to compare with, correct?

python/cuda_cccl/cuda/compute/algorithms/_segmented_reduce.py
shwina: It is a `kwarg` (defaulted to `None`)?

NaderAlAwar: I meant `**kwargs`-style handling to avoid cases like this, where the use now has to pass `None` if they want to pass a stream:

```
  segmented_reduce(..., num_segments, None, my_stream)
```

I'm not sure what the best way is, just thinking out loud

shwina: Ah I see, kwarg only arguments are an option but it's a stylistic choice we should apply across the codebase. Right now if they want to pass a stream, skipping the `max_segment_size`, they can do:

```python
  segmented_reduce(..., num_segments, stream=my_stream)
```

NaderAlAwar: My concern is that by default, you can do:

```python
  segmented_reduce(..., num_segments)
```

For most other algorithms, if you want to pass a stream you just pass it positionally as the last argument, but in this case, you have to also specify `None` then pass a stream or use a kwarg

```python
  segmented_reduce(..., num_segments, None, my_stream)
  segmented_reduce(..., num_segments, stream=my_stream)
```

which is inconsistent with most other algorithms. That said, there are a few other algorithms that have the same issue like `lower/upper_bound` with the `comp` argument and `radix_sort` with the `begin` and `end` bit arguments.

I agree that it is a stylistic choice and this is an area we have not spent too much time looking into yet, so just something to keep in mind

shwina: https://github.com/NVIDIA/cccl/issues/8296

c/parallel/include/cccl/c/segmented_reduce.h
shwina: Done

python/cuda_cccl/benchmarks/compute/segmented_reduce/variable_sum.py
shwina: Fixed

python/cuda_cccl/benchmarks/compute/segmented_reduce/sum.py
NaderAlAwar: Yes, that’s my concern. `cub/benchmarks/bench/segmented_reduce/sum.cu` benchmarks the fixed-size segmented reduce path, but Python currently only exposes the offsets-based segmented reduce path, even with `max_segment_size`. So the Python `segmented_reduce/sum.py` benchmark would not be an apples-to-apples comparison with the C++ benchmark and could be misleading from a parity/performance standpoint.

Issue #6123, opened 2025-10-03
Support a custom comparison predicate in `cub::DeviceReduce::ArgMin|Max`.


bernhardmgruber:
This will be needede to port ` thrust/extrema.h` (e.g. `thrust::min_element`) to `cub::DeviceReduce`.

~~This will require that we come up with a different way to reduce without initial element. The current initial element is determined using `::cuda::std::numeric_limits` which only works for small set of types and only if the comparison predicate is `<`.~~ This is only needed for empty problems (zero elements).

PR #8285, opened 2026-04-02
Support a custom comparison operator in `DeviceReduce::ArgMin|Max`

bernhardmgruber:
Fixes: #6123

### DIFF
--- a/cub/test/catch2_test_device_reduce.cu
+++ b/cub/test/catch2_test_device_reduce.cu
@@ -126,7 +126,7 @@ C2H_TEST("Device reduce works with all device interfaces", "[reduce][device]", f
   }
   auto d_in_it = thrust::raw_pointer_cast(in_items.data());
 
-  CAPTURE(c2h::type_name<item_t>(), c2h::type_name<output_t>(), num_items, in_items);
+  CAPTURE(c2h::type_name<item_t>(), c2h::type_name<output_t>(), num_items);
 
 #if TEST_TYPES != 4
   SECTION("reduce")


cub/cub/device/device_reduce.cuh
```diff
@@ -1031,58 +1023,66 @@ public:
...
   //!   @endrst
-  template <typename InputIteratorT, typename ExtremumOutIteratorT, typename IndexOutIteratorT>
+  template <
+    typename InputIteratorT,
+    typename ExtremumOutIteratorT,
+    typename IndexOutIteratorT,
+    typename CompareOpT,
+    // TODO(bgruber): this constraint is not accurate, since the implementation will compare the value types of
+    // ExtremumOutIteratorT, which is wrong IMO
+    ::cuda::std::enable_if_t<::cuda::std::indirectly_comparable<InputIteratorT, InputIteratorT, CompareOpT>, int> = 0>
```
bernhardmgruber, line 1072: Instead of `InputIteratorT` we should use `non_void_value_t<ExtremumOutIteratorT, it_value_t<InputIteratorT>>`, but that just "feels" wrong here. But this is what the implementation does. What do the reviewers think?

I think the implementation should actually be changed to compare the input values, not the converted ones.

bernhardmgruber:
> Suggestion: the issue being closed mentions ArgMax as well in the title, but this PR only appears to add public custom-comparator overloads and test coverage for ArgMin.

So I temporarily added a new overload for `ArgMax` as well, but then I noticed that the implementation is actually identical to `ArgMin(..., std::not_fn(compare_op), ...)` and then I wondered whether a simple negation of the predicate deserves another public API overload.

Also, while `ArgMin` without a comparison operator defaults to `std::less`, what should `ArgMax` default to? `std::max_element` defaults to `std::less` to find the maximum, but maybe that's irritating to some users on a first glance. I thought not adding `ArgMax` would just avoid the confusion.

Finally, I considered naming the new overload not `ArgMin` but something like `ArgReduce`, but that doesn't make any sense either, since the user does not specify the reduction, but the comparison predicate. Maybe we should just call the overload `ArgExtremum` to distinguish it from `ArgMin`. It generalizes both `ArgMin` and `ArgMax`. @NaderAlAwar let me know what you think!

> I would either create a separate issue for ArgMax or expose the custom comparator overload as well.

As my last paragraph points out, the new overload actually generalizes over both, `ArgMin` and `ArgMax`, so no more work should be necessary. An `ArgMax` with a custom comparison is essentially calling `ArgMin` with that operator.

NaderAlAwar:
@bernhardmgruber those are good points, I hadn't considered that. Looking into this some more, since the standard library and Thrust already expose comparator overloads for both `min_element` and `max_element`, I think matching that symmetry here would be less surprising to users than only exposing a comparator overload on `ArgMin`. Since the public CUB API already has both `ArgMin` and `ArgMax`, I would expect custom-comparator
  support to be available on both as well.

My worry about `ArgExtremum` is that it's name may be less familiar to users which could lead to them avoiding using it.

> std::max_element defaults to std::less to find the maximum, but maybe that's irritating to some users on a first glance.

I do agree that this is a little confusing. I don't feel too strongly about this either way since I have not used this extensively but my intuition would be to stay consistent with existing standards unless we believe they are broken in some way.

cub/cub/device/dispatch/dispatch_streaming_reduce.cuh
```diff
@@ -143,47 +157,49 @@ struct local_to_global_op
...
-  typename PolicySelector =
-    reduce::policy_selector_from_types<KeyValuePair<PerPartitionOffsetT, InitT>, PerPartitionOffsetT, ReductionOpT>>
+template <typename PerPartitionOffsetT,
+          typename InputIteratorT,
+          typename ExtremumOutIteratorT,
+          typename IndexOutIteratorT,
+          typename GlobalOffsetT,
+          typename ReductionOpT,
+          typename PolicySelector = policy_selector_from_types<
+            KeyValuePair<PerPartitionOffsetT, non_void_value_t<ExtremumOutIteratorT, it_value_t<InputIteratorT>>>,
```
miscco, line 167: Question: If we already add new behavior, do we want to go with `cuda::std::pair` instead of `KeyValuePair`

cub/cub/device/dispatch/dispatch_streaming_reduce.cuh
```diff
@@ -209,18 +225,31 @@ CUB_RUNTIME_FUNCTION _CCCL_FORCEINLINE cudaError_t dispatch_streaming_arg_reduce
...
+    else if constexpr (::cuda::std::is_same_v<ReductionOpT, arg_max>)
+    {
+      return ::cuda::std::numeric_limits<input_value_t>::lowest();
+    }
+    else
+    {
+      return input_value_t{};
+    }
+  }());
+  auto initial_value = empty_problem_init_t<per_partition_accum_t>{{PerPartitionOffsetT{1}, empty_problem_extremum}};
```
miscco, line 249: I am really unhappy that we actually need an initial value

cub/cub/device/device_reduce.cuh
```diff
@@ -948,18 +933,45 @@ public:
...
+    void* d_temp_storage,
+    size_t& temp_storage_bytes,
+    InputIteratorT d_in,
+    ExtremumOutIteratorT d_min_out,
+    IndexOutIteratorT d_index_out,
+    ::cuda::std::int64_t num_items,
+    CompareOpT compare_op,
+    cudaStream_t stream)
+  {
+    using PerPartitionOffsetT = int; // used by the kernel to index within one partition
```
miscco, line 948: Can we use the more explicit `int32_t`?

cub/cub/device/device_reduce.cuh
miscco: I do not follow why that constraint is wrong? We want to ensure that the input sequence is comparable with the passed operator. Why should we compare the ExtremumOutputIteratorT

cub/cub/device/device_reduce.cuh
```diff
@@ -1031,56 +1051,137 @@ public:
...
+    void* d_temp_storage,
+    size_t& temp_storage_bytes,
+    InputIteratorT d_in,
+    ExtremumOutIteratorT d_min_out,
+    IndexOutIteratorT d_index_out,
+    ::cuda::std::int64_t num_items,
+    cudaStream_t stream = 0)
+  {
+    return ArgMin(
+      d_temp_storage, temp_storage_bytes, d_in, d_min_out, d_index_out, num_items, ::cuda::std::less{}, stream);
```
miscco, line 1104: Critical: We should use the typed `cuda::std::less<T>` to avoid integer promotions
```suggestion
      d_temp_storage, temp_storage_bytes, d_in, d_min_out, d_index_out, num_items, ::cuda::std::less<iter_value_t<InputIterator>>{}, stream);
```

cub/cub/device/device_reduce.cuh
```diff
@@ -1649,58 +1714,66 @@ public:
...
+    void* d_temp_storage,
+    size_t& temp_storage_bytes,
+    InputIteratorT d_in,
+    ExtremumOutIteratorT d_max_out,
+    IndexOutIteratorT d_index_out,
+    ::cuda::std::int64_t num_items,
+    cudaStream_t stream = 0)
+  {
+    return ArgMax(
+      d_temp_storage, temp_storage_bytes, d_in, d_max_out, d_index_out, num_items, ::cuda::std::less{}, stream);
```
miscco, line 1777: Ditto use typed less

cub/cub/thread/thread_operators.cuh
```diff
@@ -106,36 +106,29 @@ struct ArgMin
...
   {
-    if ((b.second > a.second) || ((a.second == b.second) && (b.first < a.first)))
+    const auto& less = static_cast<const ValueLessThen&>(*this);
+    if (less(b.second, a.second) || (!less(a.second, b.second) && b.first < a.first))
     {
       return b;
     }
 
     return a;
   }
```
miscco, line 124: Important: Inheritance is almost always worse than making it a member.

cub/cub/thread/thread_operators.cuh
```diff
@@ -144,6 +137,29 @@ struct arg_min
...
 
+template <typename ValueLessThen>
+arg_less(ValueLessThen) -> arg_less<ValueLessThen>;
+
+/// @brief Arg min functor (keeps the value and offset of the first occurrence of the smallest item)
+using arg_min = arg_less<::cuda::std::less<>>;
+
+//! @brief Binary functor swapping the arguments to ``operator()`` before forwarding to an inner functor
+template <typename Predicate>
+struct swap_args : Predicate
```
miscco, line 148: Question: Why arent we just using `not_fn`

cub/cub/device/dispatch/dispatch_streaming_reduce.cuh
bernhardmgruber: I tried this and it became way too much work. We should do this later.

cub/cub/device/dispatch/dispatch_streaming_reduce.cuh
bernhardmgruber: It's only needed for the case where the user passes `num_items == 0`.

cub/cub/device/device_reduce.cuh
bernhardmgruber: We could but I am retaining the original code here. I would not touch it.

cub/cub/thread/thread_operators.cuh
bernhardmgruber: But with a member, we are not getting EBCO. But maybe that's not so important here.

cub/cub/thread/thread_operators.cuh
bernhardmgruber: Because `non_fun(less{})` is not the same as `greater{}`, it's `greater_equal{}`. It should actually not matter, since we are returning the first element that matches the predicate. But I felt swapping arguments is more true.

cub/cub/device/device_reduce.cuh
bernhardmgruber: The reduction implementation does not call `compare_op(d_in[i], d_in[j])`, it calls something like:
```c++
  using input_value_t     = it_value_t<InputIteratorT>;
  using accum_t = non_void_value_t<ExtremumOutIteratorT, input_value_t>;
  accum_t a = d_in[i];
  accum_t b = d_in[j];
  compare_op(a, b);
```
So it performs a conversion of the input value to the output iterator's `value_type` before comparing. That can be a totally different type.

I think this is a bug itself, but outside the scope of this PR.

cub/cub/device/device_reduce.cuh
bernhardmgruber: I don't think this matters here. The code previously (in `struct ArgMin`) just used `b.value < a.value`. Now I am doing (in `struct arg_less`) `less(b.second, a.second)`. In both cases, the inputs are taking without conversion. `::cuda::std::less<iter_value_t<InputIterator>>` would force a conversion of the comparison result to `bool`, while `::cuda::std::less` would later just rely on the contextual conversion to `bool` inside the `if` statement. No integer promotion should take place anywhere.

cub/cub/thread/thread_operators.cuh
bernhardmgruber: Hmm, I think if I move to a data member, aggregate init would no longer work with the deduction guide in C++17. This can be worked around of course. Do you insist on this change, or can I save myself 43s of typing?

bernhardmgruber:
```
          Start  65: cub.test.device.reduce.lid_0.types_0
   59/177 Test  #65: cub.test.device.reduce.lid_0.types_0 ...........................   Passed  3285.96 sec
```
Seems a bit excessive: https://github.com/NVIDIA/cccl/actions/runs/24074951429/job/70228571613?pr=8285

cub/cub/device/dispatch/dispatch_streaming_reduce.cuh
miscco: I would love for us to change the implementation so that in the legacy API without a comparison operator we do the return value thing and for the new API we only return indices, which in that case can just be `0`

cub/cub/thread/thread_operators.cuh
miscco: we can keep it as is

cub/test/catch2_test_device_reduce.cu
```diff
@@ -112,6 +126,8 @@ C2H_TEST("Device reduce works with all device interfaces", "[reduce][device]", f
   }
   auto d_in_it = thrust::raw_pointer_cast(in_items.data());
 
+  CAPTURE(c2h::type_name<item_t>(), c2h::type_name<output_t>(), num_items, in_items);
```
bernhardmgruber: Critical: capturing `in_items` produces a string representation of the device-side input vector by memcpy-ing its items one by one, which causes the 1h+ test execution :D

bernhardmgruber:
> Seems a bit excessive: https://github.com/NVIDIA/cccl/actions/runs/24074951429/job/70228571613?pr=8285

Solved

cub/test/catch2_test_device_reduce.cu
miscco: 🙀

PR #8286, opened 2026-04-03
Add nonstable env `DeviceSegmentedSort::Pairs`  2/2

gonidelis:
Split 2/2 from from https://github.com/NVIDIA/cccl/pull/8003 for easier review

### DIFF
--- a/cub/test/catch2_test_device_segmented_sort_pairs_env.cu
+++ b/cub/test/catch2_test_device_segmented_sort_pairs_env.cu
@@ -11,6 +11,9 @@ struct stream_registry_factory_t;
 
 #include <thrust/device_vector.h>
 
+#include <cuda/devices>
+#include <cuda/stream>
+
 #include "catch2_test_env_launch_helper.h"
 
 DECLARE_LAUNCH_WRAPPER(cub::DeviceSegmentedSort::StableSortPairs, stable_sort_pairs);
@@ -204,8 +207,7 @@ TEST_CASE("DeviceSegmentedSort::SortPairs nonstable uses custom stream", "[segme
   auto values_out = c2h::device_vector<int>(7);
   auto offsets    = c2h::device_vector<int>{0, 3, 7};
 
-  cudaStream_t custom_stream;
-  REQUIRE(cudaSuccess == cudaStreamCreate(&custom_stream));
+  cuda::stream stream{cuda::devices[0]};
 
   size_t expected_bytes_allocated{};
   REQUIRE(
@@ -222,7 +224,7 @@ TEST_CASE("DeviceSegmentedSort::SortPairs nonstable uses custom stream", "[segme
       thrust::raw_pointer_cast(offsets.data()),
       thrust::raw_pointer_cast(offsets.data()) + 1));
 
-  auto stream_prop = stdexec::prop{cuda::get_stream_t{}, cuda::stream_ref{custom_stream}};
+  auto stream_prop = stdexec::prop{cuda::get_stream_t{}, cuda::stream_ref{stream}};
   auto env         = stdexec::env{stream_prop, expected_allocation_size(expected_bytes_allocated)};
 
   sort_pairs(
@@ -236,13 +238,12 @@ TEST_CASE("DeviceSegmentedSort::SortPairs nonstable uses custom stream", "[segme
     thrust::raw_pointer_cast(offsets.data()) + 1,
     env);
 
-  REQUIRE(cudaSuccess == cudaStreamSynchronize(custom_stream));
+  stream.sync();
 
   c2h::device_vector<int> expected_keys{6, 7, 8, 0, 3, 5, 9};
   c2h::device_vector<int> expected_values{1, 2, 0, 5, 4, 3, 6};
   REQUIRE(keys_out == expected_keys);
   REQUIRE(values_out == expected_values);
-  REQUIRE(cudaSuccess == cudaStreamDestroy(custom_stream));
 }
 
 TEST_CASE("DeviceSegmentedSort::SortPairsDescending nonstable uses custom stream", "[segmented_sort][pairs][device]")
@@ -253,8 +254,7 @@ TEST_CASE("DeviceSegmentedSort::SortPairsDescending nonstable uses custom stream
   auto values_out = c2h::device_vector<int>(7);
   auto offsets    = c2h::device_vector<int>{0, 3, 7};
 
-  cudaStream_t custom_stream;
-  REQUIRE(cudaSuccess == cudaStreamCreate(&custom_stream));
+  cuda::stream stream{cuda::devices[0]};
 
   size_t expected_bytes_allocated{};
   REQUIRE(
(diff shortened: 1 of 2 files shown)


cub/test/catch2_test_device_segmented_sort_pairs_env.cu
```diff
@@ -0,0 +1,250 @@
...
+#include <cub/device/device_segmented_sort.cuh>
+
+#include <thrust/device_vector.h>
+
+#include "catch2_test_env_launch_helper.h"
+
+DECLARE_LAUNCH_WRAPPER(cub::DeviceSegmentedSort::SortPairs, sort_pairs);
+DECLARE_LAUNCH_WRAPPER(cub::DeviceSegmentedSort::SortPairsDescending, sort_pairs_descending);
+
+// %PARAM% TEST_LAUNCH lid 0:1:2
```
NaderAlAwar: Important: we should keep this at `lid 0:1` for now. The current lid_2 coverage here only uses 2 segments, so it stays below partitioning_threshold and exercises the fallback path which is graph capturable. `DeviceSegmentedSort` still switches to a partitioning path for larger segment counts, and that path still does a host `MemcpyAsync` followed by `SyncStream(stream)`, so this does not yet show general graph-capture support. The existing segmented sort tests in `cub/test/catch2_test_device_segmented_sort_pairs.cu:10` and `cub/test/catch2_test_device_segmented_sort_keys.cu:17` still keep lid_2 disabled for that reason, so this file should stay consistent unless we also prove the partitioning path is capture-safe.

gonidelis: good catch

cub/test/catch2_test_device_segmented_sort_pairs_env.cu
NaderAlAwar, line 1: Important: please add a custom-stream env test in this file, similar to the sibling env suites for segmented radix sort / merge sort (see https://github.com/NVIDIA/cccl/blob/6710f54e8def2788662c5ac95cb80d8f81d2be6c/cub/test/catch2_test_device_merge_sort_env.cu#L232 for example). The new cases here only pass `expected_allocation_size(...)`, so they verify temp-storage plumbing but not that the launch-helper path / segmented sort dispatch actually
  honors `cuda::get_stream_t`. We still have stream-bearing examples in `_env_api.cu`, but those don’t give the same coverage as the helper-based env

cub/cub/device/device_segmented_sort.cuh
```diff
@@ -1562,6 +1564,141 @@ public:
...
+          temp_storage_bytes,
+          d_keys,
+          d_values,
+          num_items,
+          num_segments,
+          d_begin_offsets,
+          d_end_offsets,
+          is_overwrite_okay,
+          stream);
+      });
```
NaderAlAwar, line 2794: Important: echoing bernhard's comment here https://github.com/NVIDIA/cccl/pull/8003#discussion_r3051067834, we can reuse existing code by calling `SortPairsNoNVTX`. Same applies below

cub/test/catch2_test_device_segmented_sort_pairs_env_api.cu
```diff
@@ -0,0 +1,172 @@
...
+    std::cerr << "cub::DeviceSegmentedSort::SortPairs failed with status: " << error << std::endl;
+  }
+
+  thrust::device_vector<int> expected_keys{6, 7, 8, 0, 3, 5, 9};
+  thrust::device_vector<int> expected_values{1, 2, 0, 5, 4, 3, 6};
+  // example-end sort-pairs-env
+
+  REQUIRE(error == cudaSuccess);
+  REQUIRE(keys_out == expected_keys);
+  REQUIRE(values_out == expected_values);
```
bernhardmgruber, line 214: So, since I finally understood when we need a `stream.sync()` I can now start complaining about where they are missing ;)

Important: we need

```suggestion
  stream.sync();
  REQUIRE(keys_out == expected_keys);
  REQUIRE(values_out == expected_values);
```
here, because the `cub::DeviceSegmentedSort::SortPairs` runs on a different stream than `operator==` here.

@pauleonix I think you correctly spotted this on a previous PR and I argued against it. I am sorry, because I believe you were right and I didn't correctly understand what you meant!

bernhardmgruber: We need the syncs everywhere CUB is run on a custom stream and we perform a comparison of `device_vector` later.

gonidelis:
I repushed all changes clean after refactoring that was done in #8003 with the current reviews incorporated.

cub/test/catch2_test_device_segmented_sort_pairs_env.cu
```diff
@@ -77,8 +79,323 @@ TEST_CASE("DeviceSegmentedSort::StableSortPairsDescending works with default env
...
+TEST_CASE("DeviceSegmentedSort::SortPairs nonstable uses custom stream", "[segmented_sort][pairs][device]")
+{
+  auto keys_in    = c2h::device_vector<int>{8, 6, 7, 5, 3, 0, 9};
+  auto keys_out   = c2h::device_vector<int>(7);
+  auto values_in  = c2h::device_vector<int>{0, 1, 2, 3, 4, 5, 6};
+  auto values_out = c2h::device_vector<int>(7);
+  auto offsets    = c2h::device_vector<int>{0, 3, 7};
+
+  cudaStream_t custom_stream;
+  REQUIRE(cudaSuccess == cudaStreamCreate(&custom_stream));
```
bernhardmgruber: Important: please use `cuda::stream` here for all the tests instead of the C APIs.

cub/test/catch2_test_device_segmented_sort_pairs_env_api.cu
```diff
@@ -179,3 +179,169 @@ C2H_TEST("cub::DeviceSegmentedSort::StableSortPairsDescending DoubleBuffer env-b
...
+  auto error = cub::DeviceSegmentedSort::SortPairs(
+    thrust::raw_pointer_cast(keys_in.data()),
+    thrust::raw_pointer_cast(keys_out.data()),
+    thrust::raw_pointer_cast(values_in.data()),
+    thrust::raw_pointer_cast(values_out.data()),
+    static_cast<int>(keys_in.size()),
+    2,
+    thrust::raw_pointer_cast(offsets_begin.data()),
+    thrust::raw_pointer_cast(offsets_end.data()),
+    env);
```
bernhardmgruber: Strong suggestion: pass the stream directly instead of wrapping it in an environment.
```suggestion
  cuda::stream stream{cuda::devices[0]};
  cuda::stream_ref stream_ref{stream};

  auto error = cub::DeviceSegmentedSort::SortPairs(
    thrust::raw_pointer_cast(keys_in.data()),
    thrust::raw_pointer_cast(keys_out.data()),
    thrust::raw_pointer_cast(values_in.data()),
    thrust::raw_pointer_cast(values_out.data()),
    static_cast<int>(keys_in.size()),
    2,
    thrust::raw_pointer_cast(offsets_begin.data()),
    thrust::raw_pointer_cast(offsets_end.data()),
    stream_ref);
```
Applies everywhere here

cub/test/catch2_test_device_segmented_sort_pairs_env_api.cu
pauleonix: Credits go to @gevtushenko. He noted this during Code Review Hour.

PR #8290, opened 2026-04-03
[TRIVIAL] Review fixup for #8266: RAII malloc buffers in test_cuda_kernel

andralex:
Follow-up from merged STF C API work: address review feedback on `test_cuda_kernel.cu`.

Wrap the host `malloc` buffers for the axpy test in `std::unique_ptr<void, decltype(&free)>` so `REQUIRE` failures cannot leak memory. Remove explicit `free(X)` / `free(Y)` at the end of the test.

Includes `REQUIRE` that allocation succeeded (`get() != nullptr`).

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -20,7 +20,12 @@ Checks:
   - 'concurrency-*'
   # TODO: BEGIN REMOVE ME
   - '-*'
-  - 'performance-implicit-conversion-in-loop'
+  - 'performance-*'
+  - '-performance-noexcept-move-constructor'
+  - '-performance-noexcept-swap'
+  - '-performance-enum-size'
+  - '-performance-unnecessary-copy-initialization'
+  - '-performance-unnecessary-value-param'
   # END REMOVE ME
   # HICPP is 99% aliased to other checks (mostly modernize-* and bugprone-*). We don't
   # want to also enable it because then we need to duplicate the NOLINT. The only 2 checks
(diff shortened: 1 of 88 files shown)


c/experimental/stf/test/test_cuda_kernel.cu
```diff
@@ -41,8 +44,12 @@ C2H_TEST("axpy with stf cuda_kernel", "[cuda_kernel]")
   stf_ctx_handle ctx = stf_ctx_create();
   REQUIRE(ctx != nullptr);
 
-  double* X = (double*) malloc(N * sizeof(double));
-  double* Y = (double*) malloc(N * sizeof(double));
+  std::unique_ptr<void, decltype(&free)> X_owner(malloc(N * sizeof(double)), free);
```
jrhemstad: suggestion: Can't these just be `std::vector` to avoid the custom logic here?

davebayer: Q: why don't you just use `new`/`delete`? Or `std::make_unique`

caugonnet: I'm going to vote for this option too, and avoid all complicated stuffs

c/experimental/stf/test/test_token.cpp
```diff
@@ -20,9 +22,9 @@ C2H_TEST("stf token", "[token]")
   stf_ctx_handle ctx = stf_ctx_create();
   REQUIRE(ctx != nullptr);
 
-  float* X = (float*) malloc(N * sizeof(float));
-  float* Y = (float*) malloc(N * sizeof(float));
-  float* Z = (float*) malloc(N * sizeof(float));
+  [[maybe_unused]] std::vector<float> X(N);
```
caugonnet: this is actually unused

PR #8300, opened 2026-04-07
Enable some `clang-tidy` performance checks

Jacobfaib:
## Description

Enable `clang-tidy` checks:

- `performance-faster-string-find`
- `performance-for-range-copy`
- `performance-inefficient-vector-operation`
- `performance-no-int-to-ptr`
- `performance-move-const-arg`

## Checklist

### DIFF
--- a/ci/matrix.yaml
+++ b/ci/matrix.yaml
@@ -21,7 +21,6 @@ workflows:
   #       args: '--preset libcudacxx --lit-tests "cuda/utility/basic_any.pass.cpp"' }
   #
   override:
-    - { jobs: ['build'], project: 'tidy', std: 'min', cxx: ['clang'], cudacxx: ['clang'], ctk: 'clang-cuda', sm: '75' }
 
   pull_request:
     # Old CTK: Oldest/newest supported host compilers:


libcudacxx/include/cuda/__memory/align_down.h
```diff
@@ -61,7 +61,8 @@ template <typename _Tp>
   using _Up                = ::cuda::std::remove_cv_t<_Tp>;
   const auto __char_ptr    = reinterpret_cast<char*>(const_cast<_Up*>(__ptr));
   const auto __tmp         = static_cast<uintptr_t>(__alignment - 1);
-  const auto __aligned_ptr = reinterpret_cast<char*>(reinterpret_cast<uintptr_t>(__ptr) & ~__tmp);
+  const auto __aligned_ptr = reinterpret_cast<char*>( // NOLINT(performance-no-int-to-ptr)
+    reinterpret_cast<uintptr_t>(__ptr) & ~__tmp);
```
bernhardmgruber, line 65: Suggestion: this may be more readable:

```suggestion
  // NOLINTNEXTLINE(performance-no-int-to-ptr)
  const auto __aligned_ptr = reinterpret_cast<char*>(reinterpret_cast<uintptr_t>(__ptr) & ~__tmp);
```
Applies more often in this PR.

ci/matrix.yaml
```diff
@@ -21,6 +21,7 @@ workflows:
   #       args: '--preset libcudacxx --lit-tests "cuda/utility/basic_any.pass.cpp"' }
   #
   override:
+    - { jobs: ['build'], project: 'tidy', std: 'min', cxx: ['clang'], cudacxx: ['clang'], ctk: 'clang-cuda', sm: '75' }
```
bernhardmgruber: Please revert before merging.

Jacobfaib: Yep

libcudacxx/include/cuda/__memory/align_down.h
Jacobfaib: Unfortunately, this won't work. `NOLINTNEXTLINE` applies to the literal next line, not the lexical next line. You'd have to do something like
```c++
const auto foo =
  // NOLINTNEXTLINE(...
  reinterpret_cast(...);
```
The other alternative is to do a `NOLINTBEGIN` and `NOLINTEND`, but I prefer scoping `NOLINT`s precisely to the expression we want it to apply to to avoid hiding other issues.

Jacobfaib:
> Where there any useful discoveries with performance-no-int-to-ptr? I wonder whether we should just disable the check to avoid a lot of clutter.

So it is possible to resolve `no-int-to-ptr` by rewriting:
```c++
char *orig_ptr;

uintptr_t aligned_ptr = orig_ptr + ...;
return (T*)aligned_ptr;
```
to
```c++
char *orig_ptr;

uintptr_t aligned_ptr = ...;
uintptr_t bias = aligned_ptr - orig_ptr;
return orig_ptr + bias;
```
and this may help optimizers keep track of pointer provenance. Let me try to do this rewrite for the memory alignment methods (it seems they already do something like that).

For the `cudaStream_t` casts, there is nothing we can do but silence the warning.

libcudacxx/include/cuda/__memory/align_down.h
bernhardmgruber: Alright then. Let's keep what's in this PR.

bernhardmgruber:
I understand the usefulness of the check. I just wonder how much it helps us.

> Let me try to do this rewrite for the memory alignment methods

Mind that any change here will require you to submit a bunch of benchmarks or SASS diffs to proof we won't run into a regression. You may want to address this in a separate PR.

Jacobfaib:
> Mind that any change here will require you to submit a bunch of benchmarks or SASS diffs to proof we won't run into a regression. You may want to address this in a separate PR.

Ah fair point. I had compared the output on godbolt to verify it is equivalent (https://godbolt.org/z/354fP7n9E), but that is by no means exhaustive. So let me do that in a follow-up then.

PR #8303, opened 2026-04-07
`performance-noexcept-swap`

Jacobfaib:
## Description

closes 

## Checklist

### DIFF
--- a/libcudacxx/include/cuda/std/__cccl/builtin.h
+++ b/libcudacxx/include/cuda/std/__cccl/builtin.h
@@ -409,6 +409,10 @@
 #  define _CCCL_BUILTIN_TYPE_PACK_ELEMENT(...) __type_pack_element<__VA_ARGS__>
 #endif // _CCCL_HAS_BUILTIN(__type_pack_element)
 
+#if _CCCL_HAS_BUILTIN(__is_complete_type)
+#  define _CCCL_BUILTIN_IS_COMPLETE_TYPE(...) __is_complete_type(__VA_ARGS__)
+#endif // _CCCL_HAS_BUILTIN(__is_complete_type)
+
 // NVCC prior to 12.2 have trouble with pack expansion into __type_pack_element in an alias template
 #if _CCCL_CUDACC_BELOW(12, 2)
 #  undef _CCCL_BUILTIN_TYPE_PACK_ELEMENT
--- a/libcudacxx/include/cuda/std/__tuple_dir/tuple_leaf.h
+++ b/libcudacxx/include/cuda/std/__tuple_dir/tuple_leaf.h
@@ -440,7 +440,19 @@ struct _CCCL_DECLSPEC_EMPTY_BASES __tuple_impl<__tuple_indices<_Indx...>, _Tp...
   _CCCL_HIDE_FROM_ABI __tuple_impl& operator=(__tuple_impl&&)      = default;
 
   // Using a fold exppression here breaks nvrtc
-  _CCCL_API inline void swap(__tuple_impl& __t) noexcept(__fold_and_v<is_nothrow_swappable_v<_Tp>...>)
+  _CCCL_API inline void swap(__tuple_impl& __t)
+  // NVCC 12.0.X has a bug where it instantiates friend functions eagerly. This leads to errors
+  // because the friend swap() in tuple causes this swap() to (transitively) be instantiated
+  // regardless of whether it is called or not.
+  //
+  // When using tuples of incomplete types this causes errors with is_nothrow_swappable (or
+  // rather, any is_swappable trait) as they require the types to be complete. In this case we
+  // need to lazily instantiate these templates so we can short-circuit if _Tp is incomplete.
+#if _CCCL_CUDA_COMPILER(NVCC, <, 12, 1)
+    noexcept(__fold_and_v<__is_complete_and_nothrow_swappable_v<remove_cvref_t<_Tp>>...>)
+#else // ^^^  _CCCL_CUDA_COMPILER(NVCC, <, 12, 1) ^^^ / vvv  _CCCL_CUDA_COMPILER(NVCC, >=, 12, 1) vvv
+    noexcept(__fold_and_v<is_nothrow_swappable_v<_Tp>...>)
+#endif //  _CCCL_CUDA_COMPILER(NVCC, <, 12, 1)
   {
     (__tuple_leaf<_Indx, _Tp>::swap(static_cast<__tuple_leaf<_Indx, _Tp>&>(__t)), ...);
   }
--- a/libcudacxx/include/cuda/std/__type_traits/is_complete.h
+++ b/libcudacxx/include/cuda/std/__type_traits/is_complete.h
@@ -1,0 +1,53 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of libcu++, the C++ Standard Library for your entire system,
+// under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef _CUDA_STD___TYPE_TRAITS_IS_COMPLETE_H
+#define _CUDA_STD___TYPE_TRAITS_IS_COMPLETE_H
+
+#include <cuda/std/detail/__config>
+
+#if defined(_CCCL_IMPLICIT_SYSTEM_HEADER_GCC)
+#  pragma GCC system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_CLANG)
+#  pragma clang system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_MSVC)
+#  pragma system_header
+#endif // no system header
+
+#include <cuda/std/__type_traits/void_t.h> // IWYU pragma: keep
+
+#include <cuda/std/__cccl/prologue.h>
+
+_CCCL_BEGIN_NAMESPACE_CUDA_STD
+
+#ifdef _CCCL_BUILTIN_IS_COMPLETE_TYPE
+
+template <typename T>
+inline constexpr bool __is_complete_v = _CCCL_BUILTIN_IS_COMPLETE_TYPE(T);
+
+#else // ^^^  _CCCL_BUILTIN_IS_COMPLETE_TYPE ^^^ / vvv no builtin vvv
+// Must be a SFINAE trait instead of
+//
+// template <typename T>
+// inline constexpr __is_complete_v = sizeof(T) > 0;
+//
+// Because older NVCC doesn't even allow you to utter the phrase sizeof(T) if T is incomplete
+template <typename T, typename = void>
+inline constexpr bool __is_complete_v = false;
+
+template <typename T>
+inline constexpr bool __is_complete_v<T, void_t<decltype(sizeof(T))>> = (sizeof(T) > 0);
+#endif // ^^^ no builtin ^^^
+
+_CCCL_END_NAMESPACE_CUDA_STD
+
+#include <cuda/std/__cccl/epilogue.h>
+
+#endif // _CUDA_STD___TYPE_TRAITS_IS_COMPLETE_H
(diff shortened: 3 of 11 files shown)


Jacobfaib:
> I remember that I had hard time with noexcept and compiler compatibility. @miscco should take a look at it

Yes I am struggling at the moment with a particular edge case of
```c++
struct IncompleteType;

cuda::std::tuple<IncompleteType&> t;
```
This is causing compilation errors because the `noexcept` specification is being eagerly computed, via `is_nothrow_swappable_v<IncompleteType&>`. However, computing `is_nothrow_swappable` with an incomplete type is undefined behavior according to the standard.

I guess I might have to resort to the usual method of deferring this via some dummy argument:
```c++
template <typename DummyT = char>
void swap(tuple& tup, DummyT = DummyT{}) noexcept(noexcept(...))
```
But it seems a little icky to do this on the public interface. I also don't understand how libc++ is getting around this, their implementation seems quite similar to ours but they don't have the error.

miscco:
> However, computing `is_nothrow_swappable` with an incomplete type is undefined behavior according to the standard.

I thought we were trying hard to make that work, will look at this on a friday

Jacobfaib:
> I thought we were trying hard to make that work, will look at this on a friday

The root problem is that older nvcc (or perhaps it's GCC) don't even allow you to write `sizeof(T)` if `T` is incomplete. See https://godbolt.org/z/9P64PE9s8 for an example.

libcudacxx/include/cuda/std/__type_traits/is_complete.h
```diff
@@ -0,0 +1,53 @@
...
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_CLANG)
+#  pragma clang system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_MSVC)
+#  pragma system_header
+#endif // no system header
+
+#include <cuda/std/__type_traits/integral_constant.h>
+#if !_CCCL_HAS_BUILTIN(__is_complete_type)
+#  include <cuda/std/__type_traits/void_t.h>
+#endif
```
miscco:
```suggestion
#  include <cuda/std/__type_traits/void_t.h>
```

libcudacxx/include/cuda/std/__tuple_dir/tuple_leaf.h
```diff
@@ -46,6 +46,12 @@
 #include <cuda/std/__utility/move.h>
 #include <cuda/std/__utility/swap.h>
 
+#if _CCCL_CUDA_COMPILER(NVCC, <, 12, 1)
+#  define _CCCL_BROKEN_SIZEOF_INCOMPLETE_TYPE
+#  include <cuda/std/__type_traits/conjunction.h>
+#  include <cuda/std/__type_traits/is_complete.h>
+#endif
```
miscco: We should just include those headers, they are minimal for a reason

libcudacxx/include/cuda/std/__tuple_dir/tuple_leaf.h
```diff
@@ -440,12 +446,29 @@ struct _CCCL_DECLSPEC_EMPTY_BASES __tuple_impl<__tuple_indices<_Indx...>, _Tp...
...
+  //
+  // When using tuples of incomplete types this causes errors with is_nothrow_swappable (or
+  // rather, any is_swappable trait) as they require the types to be complete. In this case we
+  // need to lazily instantiate these templates so we can short-circuit if _Tp is incomplete.
+  //
+  // This is also why we must use conjunction. __fold_and_v is eager, while conjunction is
+  // lazy.
+#ifdef _CCCL_BROKEN_SIZEOF_INCOMPLETE_TYPE
+    noexcept(conjunction_v<__is_complete<remove_cvref_t<_Tp>>..., is_nothrow_swappable<_Tp>...>)
+#else
```
miscco: Critical: Also I am not sure whether `__is_complete` is universally available.

Important: Can we have a simpler trait like

```
template<class _Tp, class =  void>
inline constexpr bool __is_complete_and_nothrow_swappable = false;

template<class _Tp>
inline constexpr bool __is_complete_and_nothrow_swappable<_Tp, void_t<decltype(sizeof(_Tp))>> =  is_nothrow_swappable<_Tp>;
```

libcudacxx/include/cuda/std/__type_traits/is_complete.h
Jacobfaib: I reworked this bit to use the usual builtin handling (adding `_CCCL_BUILTIN_IS_COMPLETE_TYPE`). Do you still want me to make this change? `void_t` is only used if we don't have the builtin.

libcudacxx/include/cuda/std/__tuple_dir/tuple_leaf.h
Jacobfaib: > Critical: Also I am not sure whether __is_complete is universally available.

It is. If we have the builtin, we use that. If not, we use the `void_t` version.

> Important: Can we have a simpler trait like

Yes, where should it be defined?

miscco: can either put it next to swappable or whereever it is needed

libcudacxx/include/cuda/std/__type_traits/is_complete.h
```diff
@@ -0,0 +1,62 @@
...
+#  pragma clang system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_MSVC)
+#  pragma system_header
+#endif // no system header
+
+#include <cuda/std/__type_traits/integral_constant.h>
+
+#ifndef _CCCL_BUILTIN_IS_COMPLETE_TYPE
+#  include <cuda/std/__type_traits/void_t.h>
+#endif
```
miscco: Please include unconditionally

libcudacxx/include/cuda/std/__type_traits/is_complete.h
```diff
@@ -0,0 +1,62 @@
...
+
+#include <cuda/std/__cccl/prologue.h>
+
+_CCCL_BEGIN_NAMESPACE_CUDA_STD
+
+#ifdef _CCCL_BUILTIN_IS_COMPLETE_TYPE
+template <typename T>
+struct __is_complete : bool_constant<_CCCL_BUILTIN_IS_COMPLETE_TYPE(T)>
+{};
+#else
```
miscco: This is an internal trait, we should never rely on structs if we can avoid it, so we should always define `inline constexpr bool __is_complete_v` I believe we never need the struct version

libcudacxx/include/cuda/std/__cccl/builtin.h
```diff
@@ -409,6 +409,10 @@
 #  define _CCCL_BUILTIN_TYPE_PACK_ELEMENT(...) __type_pack_element<__VA_ARGS__>
 #endif // _CCCL_HAS_BUILTIN(__type_pack_element)
 
+#if _CCCL_HAS_BUILTIN(__is_complete_type)
+#  define _CCCL_BUILTIN_IS_COMPLETE_TYPE(...) __is_complete_type(__VA_ARGS__)
+#endif
```
miscco: We try really hard to have no uncommented `#endif`
```suggestion
#endif // _CCCL_HAS_BUILTIN(__is_complete_type)
```

libcudacxx/include/cuda/std/__tuple_dir/tuple_leaf.h
```diff
@@ -440,7 +440,19 @@ struct _CCCL_DECLSPEC_EMPTY_BASES __tuple_impl<__tuple_indices<_Indx...>, _Tp...
...
+  // regardless of whether it is called or not.
+  //
+  // When using tuples of incomplete types this causes errors with is_nothrow_swappable (or
+  // rather, any is_swappable trait) as they require the types to be complete. In this case we
+  // need to lazily instantiate these templates so we can short-circuit if _Tp is incomplete.
+#if _CCCL_CUDA_COMPILER(NVCC, <, 12, 1)
+    noexcept(__fold_and_v<__is_complete_and_nothrow_swappable_v<remove_cvref_t<_Tp>>...>)
+#else
+    noexcept(__fold_and_v<is_nothrow_swappable_v<_Tp>...>)
+#endif
```
miscco:
```suggestion
#if _CCCL_CUDA_COMPILER(NVCC, <, 12, 1)
    noexcept(__fold_and_v<__is_complete_and_nothrow_swappable_v<remove_cvref_t<_Tp>>...>)
#else // ^^^  _CCCL_CUDA_COMPILER(NVCC, <, 12, 1) ^^^ / vvv  _CCCL_CUDA_COMPILER(NVCC, >=, 12, 1) vvv
    noexcept(__fold_and_v<is_nothrow_swappable_v<_Tp>...>)
#endif //  _CCCL_CUDA_COMPILER(NVCC, <, 12, 1)
```

libcudacxx/include/cuda/std/__type_traits/is_complete.h
```diff
@@ -0,0 +1,50 @@
...
+#include <cuda/std/__type_traits/void_t.h> // IWYU pragma: keep
+
+#include <cuda/std/__cccl/prologue.h>
+
+_CCCL_BEGIN_NAMESPACE_CUDA_STD
+
+#ifdef _CCCL_BUILTIN_IS_COMPLETE_TYPE
+template <typename T>
+inline constexpr bool __is_complete_v = _CCCL_BUILTIN_IS_COMPLETE_TYPE(T);
+#else
```
miscco:
```suggestion
#ifdef _CCCL_BUILTIN_IS_COMPLETE_TYPE

template <typename T>
inline constexpr bool __is_complete_v = _CCCL_BUILTIN_IS_COMPLETE_TYPE(T);

#else // ^^^  _CCCL_BUILTIN_IS_COMPLETE_TYPE ^^^ / vvv no builtin vvv
```

libcudacxx/include/cuda/std/__type_traits/is_complete.h
```diff
@@ -0,0 +1,50 @@
...
+// template <typename T>
+// inline constexpr __is_complete_v = sizeof(T) > 0;
+//
+// Because older NVCC doesn't even allow you to utter the phrase sizeof(T) if T is incomplete
+template <typename T, typename = void>
+inline constexpr bool __is_complete_v = false;
+
+template <typename T>
+inline constexpr bool __is_complete_v<T, void_t<decltype(sizeof(T))>> = (sizeof(T) > 0);
+#endif
```
miscco:
```suggestion
#endif // ^^^ no builtin ^^^
```

libcudacxx/include/cuda/std/__type_traits/is_complete.h
```diff
@@ -0,0 +1,50 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-FileCopyrightText: Copyright (c) 2023 NVIDIA CORPORATION & AFFILIATES.
+//
+//===----------------------------------------------------------------------===//
```
miscco, line 9: This should have the libcu++ license for 2026

thrust/thrust/detail/contiguous_storage.h
```diff
@@ -100,8 +101,30 @@ class contiguous_storage
 
   _CCCL_HOST_DEVICE void deallocate() noexcept;
 
+private:
+  struct is_nothrow_swappable
+  {
+    static constexpr bool value()
```
bernhardmgruber: Suggestion: can we just omit the struct?

```suggestion
    static constexpr bool is_nothrow_swappable()
```

bernhardmgruber (commented):
Thrust changes are fine

thrust/thrust/detail/contiguous_storage.h
Jacobfaib: Yes, good point.

libcudacxx/include/cuda/std/__type_traits/is_complete.h
Jacobfaib: Ah, hadn't noticed that. I copied this file originally over from another one under `cuda/std`, and I am noticing now that many such files have the LLVM license banner. Is this a copy-paste error or are these truly separately licensed?

libcudacxx/include/cuda/std/__cccl/builtin.h
Jacobfaib: This should really be automated...

libcudacxx/include/cuda/std/__type_traits/is_complete.h
miscco: Its more a technicality LLVM has no copyright for files that are fully written by us.

So every file that was not ported from llvm should have the libcu++ license

PR #8304, opened 2026-04-07
`performance-enum-size`

Jacobfaib:
## Description

closes 

## Checklist

### DIFF
--- a/cub/cub/agent/agent_radix_sort_onesweep.cuh
+++ b/cub/cub/agent/agent_radix_sort_onesweep.cuh
@@ -33,6 +33,7 @@
 #include <cuda/std/__type_traits/conditional.h>
 #include <cuda/std/__type_traits/integral_constant.h>
 #include <cuda/std/__type_traits/is_same.h>
+#include <cuda/std/__utility/to_underlying.h>
 #include <cuda/std/cstdint>
 
 #if !_CCCL_COMPILER(NVRTC)
@@ -69,7 +70,7 @@ inline ::std::ostream& operator<<(::std::ostream& os, RadixSortStoreAlgorithm al
     case RADIX_SORT_STORE_ALIGNED:
       return os << "RADIX_SORT_STORE_ALIGNED";
     default:
-      return os << "<unknown RadixSortStoreAlgorithm: " << static_cast<int>(algo) << ">";
+      return os << "<unknown RadixSortStoreAlgorithm: " << ::cuda::std::to_underlying(algo) << ">";
   }
 }
 #endif // !_CCCL_COMPILER(NVRTC) && !_CCCL_DOXYGEN_INVOKED


cub/cub/agent/agent_radix_sort_onesweep.cuh
```diff
@@ -45,7 +46,7 @@ CUB_NAMESPACE_BEGIN
  * memory. Currently applies only to writing 4B keys in full tiles; in all other cases,
  * RADIX_SORT_STORE_DIRECT is used.
  */
-enum RadixSortStoreAlgorithm
+enum RadixSortStoreAlgorithm : ::cuda::std::uint8_t
```
fbusato, line 50: no, this wrong. This can introduce a breaking change. Please revert for public members

fbusato: secondly, it has no performance benefits. It only adds verbosity

Jacobfaib: > no, this wrong. This can introduce a breaking change. Please revert for public members

Hmmm I'm not sure I follow. Are the underlying types of enums considered a public part of the interface in CUB? The usual contract for enums is that only the names are considered public, and everything else (the specific values, number of values, and types) is considered to be abstracted away.

fbusato: this can affects integer conversion behavior, overload resolution, `sizeof`, ABI, etc. 
Enums are considered a public part of the interface in CUB

Jacobfaib: OK to clarify, I should revert all of the public enum underlying size for CUB?

I was under the impression that CUB does not make promises about ABI (only libcu++). Also vis-a-vis overload resolution, I assume you mean implicit conversion of enums to their underlying type:
```c++
void foo(int);
void foo(unsigned);

enum Enum {};
enum UEnum : unsigned {};

foo(Enum{}); // calls foo(int);
foo(UEnum{}); // calls foo(unsigned);
```
In the above case, what about `enum class`? Those would require a specific cast anyways, so overload resolution and integer conversion is always explicit. We could change the underlying type without breaking anything (though if you'd rather err on the safe side we can disable the check for those too).

fbusato: This modification puts CUB in a dangerous position only for stylistic reasons. I'm not again it but please defer all the public enum changes to CCCL 4

gevtushenko: Even if we put ABI question aside, public enums are used in strange ways ([1](https://github.com/oneapi-src/SYCLomatic/blob/f350c9d684edb1268f5e8fe9eea0666a680060fd/clang/lib/DPCT/RulesLangLib/MapNamesLangLib.cpp#L35), [2](https://github.com/cesmix-mit/MDP/blob/746e4f4aead5911ddeda77b0d2c117e3b70cc5c4/src/C%2B%2B/Core/hipcub/backend/rocprim/block/block_load.hpp#L54)). Without careful consideration, it's hard to prove that this won't break someone. 

The benefit of the change is subtle as well. Since most of these are used at compile time, we won't get any performance improvement while getting more verbose code.

cub/cub/agent/single_pass_scan_operators.cuh
```diff
@@ -94,7 +95,7 @@ struct BlockScanRunningPrefixOp
 /**
  * Enumerations of tile status
  */
-enum ScanTileStatus
```
gevtushenko, line 97: question: this is a rare case when we use enum at runtime. Does this affect codegen in any way?

Jacobfaib: It would affect codegen in the same way that switching a variable from `int` to `uint8_t` will. Basically, for all of these underyling type changes, it makes all of these enums smaller (clang-tidy will always suggest the smallest possible type, given the enum values[^1]).

[^1]: This is not strictly true. If you have what looks like a flag enum (i.e. one whose members you intend to be XOR-ing), then `clang-tidy` will compute the largest possible value you can make and suggest a type capable of holding that.

Jacobfaib:
@gevtushenko @fbusato @miscco @bernhardmgruber @davebayer given how controversial this check is, the overhead of needing to undo all the CUB changes (and likely others), and not wanting to give myself carpel tunnel I think I will just disable this check entirely for now. 

We could revisit it come CCCL 4.0, but in the interest of moving https://github.com/NVIDIA/cccl/pull/8305 and other PRs forward I think it's probably for the best to let sleeping dogs lie for this one. Any objections? Otherwise I will close this PR.

PR #8305, opened 2026-04-07
`performance-unnecessary-value-param`

Jacobfaib:
## Description

To silence this check you can either:

- Make arguments a const-ref
- Move or forward the argument later on
- Silence the check

My policy for fixing this was as follows:

- If the API is usually used on device (or is device only), silence the check. Promoting by-value arguments to references might lead to missed optimizations from nvcc on device as it may not be able to prove that pointers are restricted. Also overwhelming majority of on-device data structures are PODs so move makes no difference for perf (not to mention correctness issues if e.g. init values come from inside a container to be modified).
- If the API is host-only, or majority host, move or forward the arguments.
- If moving is not possible or "hard" (big function, easy to inadvertently use after move), silence the diagnostic. In extremely rare cases (test code), where I could prove that it was host-only, I made it a const-ref.

The upshot is that this was almost never fixed via const-ref. The vast majority of violations in CUB are silenced, and a lot of them are fixed via moves in thrust.

## Checklist

### DIFF
--- a/thrust/thrust/detail/reduce.inl
+++ b/thrust/thrust/detail/reduce.inl
@@ -20,8 +20,7 @@
 
 #include <cuda/std/__utility/move.h>
 
-// Include all active backend system implementations (generic, sequential, host
-// and device)
+// Include all active backend system implementations (generic, sequential, host and device)
 #include <thrust/system/detail/generic/reduce.h>
 #include <thrust/system/detail/generic/reduce_by_key.h>
 #include <thrust/system/detail/sequential/reduce.h>
--- a/thrust/thrust/system/cuda/detail/scan_by_key.h
+++ b/thrust/thrust/system/cuda/detail/scan_by_key.h
@@ -336,18 +336,11 @@ ValOutputIt _CCCL_HOST_DEVICE exclusive_scan_by_key(
        value_first,
        value_result,
        ::cuda::std::distance(key_first, key_last),
-       init,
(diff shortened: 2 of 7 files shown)


fbusato (commented):
not sure if it is worth for CUB. The PR applies `NOLINT` everywhere and use `move` which just adds verbosity

miscco:
I believe this PR is too big to properly review. 

We should split this up into smaller parts per subproject. Also I second @jrhemstad that using references is sometimes a loss, especially for types that are cheap to copy

Jacobfaib:
@jrhemstad @miscco see the PR description, I took great care to not use references anywhere that would reasonably be used on device more often than not (e.g. CUB, or the thrust kernels). These are all silenced.

But yeah, given how much of these diffs are just `// NOLINT`, it seems like we can probably get away with disabling the check.

wmaxey:
> I believe this PR is too big to properly review.
> 
> We should split this up into smaller parts per subproject. Also I second @jrhemstad that using references is sometimes a loss, especially for types that are cheap to copy

I think this is mostly concisely targeting Thrust. I like the idea from a correctness point of view, but understand the difficulty with actually narrowing the check to just host API. I won't immediately say no to the bulk of this.

Jacobfaib:
Punting to possibly CCCL 4.0 since there seems to be no traction in favor for now.

PR #8306, opened 2026-04-07
`performance-unnecessary-copy-initialization`

Jacobfaib:
## Description

closes 

## Checklist

### DIFF
--- a/thrust/testing/permutation_iterator.cu
+++ b/thrust/testing/permutation_iterator.cu
@@ -243,7 +243,7 @@ void TestPermutationIteratorHostDeviceScatter()
   // scatter device->host
   thrust::copy(d_source.begin(), d_source.end(), p_h_output);
 
-  HostVector href(dref);
+  HostVector href = dref;
   ASSERT_EQUAL(h_output, href);
 }
 DECLARE_UNITTEST(TestPermutationIteratorHostDeviceScatter);
--- a/thrust/testing/unittest/testframework.h
+++ b/thrust/testing/unittest/testframework.h
@@ -474,71 +474,71 @@ public:
   TEST##UnitTest TEST##Instance
 
 // Macro to create instances of a test for several array sizes.
-#define DECLARE_SIZED_UNITTEST(TEST)                \
-  class TEST##UnitTest : public UnitTest            \
-  {                                                 \
-  public:                                           \
-    TEST##UnitTest()                                \
-        : UnitTest(#TEST)                           \
-    {}                                              \
-    void run()                                      \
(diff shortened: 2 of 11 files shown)


thrust/testing/unittest/testframework.h
```diff
@@ -474,71 +474,71 @@ class UnitTestDriver
...
+    TEST##UnitTest()                                       \
+        : UnitTest(#TEST)                                  \
+    {}                                                     \
+    void run()                                             \
+    {                                                      \
+      const std::vector<size_t>& sizes = get_test_sizes(); \
+      for (size_t i = 0; i != sizes.size(); ++i)           \
+      {                                                    \
+        TEST(sizes[i]);                                    \
+      }                                                    \
```
bernhardmgruber, line 490: Suggestion: just use a range-for here. Also in the places below.

thrust/testing/permutation_iterator.cu
```diff
@@ -243,7 +243,9 @@ void TestPermutationIteratorHostDeviceScatter()
   // scatter device->host
   thrust::copy(d_source.begin(), d_source.end(), p_h_output);
 
-  HostVector href(dref);
+  // This actually seems wrong, in fact this whole test seems wrong. HostVector and
+  // DeviceVector are both host_vectors. It seems like they aren't supposed to be.
+  HostVector href = dref; // NOLINT(performance-unnecessary-copy-initialization)
```
bernhardmgruber: I think so. #8312

miscco: This is the case when we do not have a CUDA backend.

For the host backends host_vector == device__vector, but not for the CUDA backend

```suggestion
  HostVector href = dref; // NOLINT(performance-unnecessary-copy-initialization)
```

Jacobfaib: No I mean, in this test `HostVector` and `DeviceVector` were _literally_ defined as the same type:
```c++
using HostVector     = thrust::host_vector<T>;
using DeviceVector   = thrust::host_vector<T>;
```
But I see this seems to have been fixed by https://github.com/NVIDIA/cccl/pull/8312 just now :)

thrust/testing/unittest/testframework.h
Jacobfaib: That will be done by the `modernize-*` pass for `clang-tidy`. It took enormous effort and willpower not to go around fixing such things in this pass :(

bernhardmgruber: Fine.

PR #8308, opened 2026-04-07
[STF] Use move ownership for child graph nodes with mem-alloc nodes

caugonnet:
cudaGraphAddChildGraphNode always clones the child graph into the parent. When the child graph contains memory allocation or free nodes (e.g. from cudaMallocAsync during stream capture), CUDA rejects the clone with CUDA_ERROR_NOT_SUPPORTED.

CTK 13+ provides cudaGraphChildGraphOwnershipMove via cudaGraphAddNode
+ cudaGraphNodeParams, which transfers ownership of the child graph to the parent instead of cloning. This is required when stream-captured task bodies call APIs that internally use cudaMallocAsync (such as cuda.compute / CUB device algorithms).

For CTK < 13, the old cudaGraphAddChildGraphNode (clone) path is kept.

Adds a C++ unit test that calls cudaMallocAsync/cudaFreeAsync inside a graph_ctx task body to exercise the move-ownership code path.

Made-with: Cursor

## Description

closes 

## Checklist

### DIFF
--- a/cudax/include/cuda/experimental/__stf/graph/graph_task.cuh
+++ b/cudax/include/cuda/experimental/__stf/graph/graph_task.cuh
@@ -224,11 +224,10 @@ public:
         // Move ownership so child graphs with memory alloc/free nodes
         // (e.g. from cudaMallocAsync during stream capture) are accepted.
         cudaGraphNodeParams nodeParams = {};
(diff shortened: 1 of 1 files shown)


caugonnet:
pre-commit.ci autofix

PR #8311, opened 2026-04-07
Implement the new tuning API for `DispatchSelectIf`

bernhardmgruber:
- [x] No SASS diffs on `cub.bench.select.if.base` for SM`75;80;86;90;100`
- [x] No SASS diffs on `cub.bench.select.flagged.base` for SM`75;80;86;90;100`
- [x] No SASS diffs on `cub.bench.select.unique.base` for SM`75;80;86;90;100`
- [x] No SASS diffs on `cub.bench.partition.if.base` for SM`75;80;86;90;100`
- [x] No SASS diffs on `cub.bench.partition.flagged.base` for SM`75;80;86;90;100`

Fixes: #7644

### DIFF
--- a/cub/cub/device/dispatch/tuning/tuning_select_if.cuh
+++ b/cub/cub/device/dispatch/tuning/tuning_select_if.cuh
@@ -27,6 +27,10 @@
 #include <cuda/std/__type_traits/is_same.h>
 #include <cuda/std/optional>
 
+#if _CCCL_COMPILER(MSVC)
+#  include <ciso646>
+#endif // _CCCL_COMPILER(MSVC)
+
 #if !_CCCL_COMPILER(NVRTC)
 #  include <ostream>
 #endif // !_CCCL_COMPILER(NVRTC)


cub/cub/device/dispatch/tuning/tuning_select_if.cuh
```diff
@@ -17,44 +17,58 @@
...
 #include <cub/block/block_scan.cuh>
+#include <cub/detail/delay_constructor.cuh>
+#include <cub/device/dispatch/tuning/common.cuh>
 #include <cub/util_device.cuh>
 #include <cub/util_math.cuh>
 #include <cub/util_type.cuh>
 
 #include <cuda/std/__algorithm/clamp.h>
 #include <cuda/std/__type_traits/is_same.h>
+#include <cuda/std/concepts>
```
miscco: probably unused

cub/cub/device/dispatch/tuning/tuning_select_if.cuh
```diff
@@ -1564,6 +1590,983 @@ struct policy_hub
...
+      BLOCK_SCAN_WARP_SCANS,
+      delay_constructor_policy{delay_constructor_kind::fixed_delay, 350, 450}};
+  }
+
+  [[nodiscard]] _CCCL_API constexpr auto make_scaled_policy(
+    int threads, int nominal_4b_items, BlockLoadAlgorithm la, CacheLoadModifier lm, delay_constructor_policy delay) const
+    -> select_if_policy
+  {
+    const int ipt = nominal_4B_items_to_items(nominal_4b_items, input_size_bytes);
+    return select_if_policy{threads, ipt, la, lm, BLOCK_SCAN_WARP_SCANS, delay};
```
miscco: Nitpick: above this is named 
```suggestion
    const int items_per_thread = nominal_4B_items_to_items(nominal_4b_items, input_size_bytes);
    return select_if_policy{threads, items_per_thread, la, lm, BLOCK_SCAN_WARP_SCANS, delay};
```

cub/cub/device/dispatch/tuning/tuning_select_if.cuh
```diff
@@ -1564,6 +1590,983 @@ struct policy_hub
...
+      {
+        return select_if_policy{
+          256,
+          5,
+          BLOCK_LOAD_WARP_TRANSPOSE,
+          LOAD_DEFAULT,
+          BLOCK_SCAN_WARP_SCANS,
+          delay_constructor_policy{delay_constructor_kind::fixed_delay, 400, 1090}};
+      }
+      if (has_flags && keep_rejects)
```
miscco, line 1706: Question: This is the same as the above condition, is that intentional?

cub/cub/device/dispatch/tuning/tuning_select_if.cuh
```diff
@@ -1564,6 +1590,983 @@ struct policy_hub
...
+    return select_if_policy{threads, ipt, la, lm, BLOCK_SCAN_WARP_SCANS, delay};
+  }
+
+  [[nodiscard]] _CCCL_API constexpr auto get_sm80_tuning(bool has_flags, bool keep_rejects) const -> select_if_policy
+  {
+    // before SM100, we only tuned for int32, but we always take these tunings independently of the offset type size
+
+    if (input_type == type_t::int128 || input_type == type_t::uint128)
+    {
+      if (!has_flags && !keep_rejects)
```
miscco: Suggestion: the flags are really error prone because it is easy to miss whether there is a `!` or not. 

We should consider introducing an enumeration for that and then switch over that

cub/cub/device/dispatch/tuning/tuning_select_if.cuh
```diff
@@ -1564,6 +1590,983 @@ struct policy_hub
...
+    }
+    return default_policy(LOAD_DEFAULT);
+  }
+
+  [[nodiscard]] _CCCL_API constexpr auto get_sm100_tuning(bool has_flags, bool keep_rejects, bool may_alias) const
+    -> ::cuda::std::optional<select_if_policy>
+  {
+    if (!input_is_primitive)
+    {
+      return {};
```
miscco: What is the meaning of no tuning, will we then fall back into the one from the previous architecture?

cub/cub/device/dispatch/dispatch_select_if.cuh
```diff
@@ -828,6 +868,275 @@ struct DispatchSelectIf
...
+  SelectOpT select_op,
+  EqualityOpT equality_op,
+  OffsetT num_items,
+  cudaStream_t stream,
+  [[maybe_unused]] PolicySelector policy_selector,
+  KernelLauncherFactory launcher_factory)
+{
+  static_assert(::cuda::std::is_empty_v<PolicyGetter>);
+
+  using per_partition_offset_t                     = detail::select::per_partition_offset_t;
```
miscco: Question is this an signed integer?

cub/cub/device/dispatch/dispatch_select_if.cuh
```diff
@@ -828,6 +868,275 @@ struct DispatchSelectIf
...
+  using num_total_items_t                          = OffsetT;
```
miscco: Question: What is the purpose of this alias, it feels like `OffsetT` is perfectly fine

cub/cub/device/dispatch/tuning/tuning_select_if.cuh
bernhardmgruber: It shouldn't. The conditions are:
```c++
      if (!has_flags && !keep_rejects) { .. }
      if (has_flags && !keep_rejects) { .. }
      if (!has_flags && keep_rejects) { .. }
      if (has_flags && keep_rejects) { .. }
```
so basically all combinations.

We would really need pattern matching here :)

cub/cub/device/dispatch/tuning/tuning_select_if.cuh
bernhardmgruber: How about `not`?

cub/cub/device/dispatch/tuning/tuning_select_if.cuh
bernhardmgruber: Yes, added a comment.

cub/cub/device/dispatch/dispatch_select_if.cuh
bernhardmgruber: It's `::cuda::std::int32_t`.

cub/cub/device/dispatch/dispatch_select_if.cuh
bernhardmgruber: There is none, I just retained it from the old dispatcher. Refactored it out.

cub/cub/device/dispatch/tuning/tuning_select_if.cuh
miscco: Oh I meant that the policy is the same in both clauses

bernhardmgruber: I just checked the old policy hub and the same tuning is indeed used for both versions.

PR #8100, opened 2026-03-19
Add env-based API for DoubleBuffer cub::DeviceSegmentedRadixSort

gonidelis:
closes https://github.com/NVIDIA/cccl/issues/8017

PR #8313, opened 2026-04-07
Add DoubleBuffer env overloads for `DeviceSegmentedRadixSort::SortPairs`

gonidelis:
Splits https://github.com/NVIDIA/cccl/pull/8100 1/2

### DIFF
--- a/cub/test/catch2_test_device_segmented_radix_sort_env_api.cu
+++ b/cub/test/catch2_test_device_segmented_radix_sort_env_api.cu
@@ -87,7 +87,7 @@ C2H_TEST("cub::DeviceSegmentedRadixSort::SortPairsDescending env with stream", "
   thrust::device_vector<int> expected_keys{8, 7, 6, 9, 5, 3, 0};
   thrust::device_vector<int> expected_values{0, 2, 1, 6, 3, 4, 5};
   // example-end segmented-radix-sort-pairs-descending-env
-  stream.sync()
+  stream.sync();
 
   REQUIRE(error == cudaSuccess);
   REQUIRE(keys_out == expected_keys);
@@ -156,7 +156,7 @@ C2H_TEST("cub::DeviceSegmentedRadixSort::SortKeysDescending env with stream", "[
 
   thrust::device_vector<int> expected_keys{8, 7, 6, 9, 5, 3, 0};
   // example-end segmented-radix-sort-keys-descending-env
-  stream.sync()
+  stream.sync();
 
   REQUIRE(error == cudaSuccess);
   REQUIRE(keys_out == expected_keys);


cub/test/catch2_test_device_segmented_radix_sort_env_api.cu
```diff
@@ -157,3 +157,97 @@ C2H_TEST("cub::DeviceSegmentedRadixSort::SortKeysDescending env with stream", "[
...
+  if (error != cudaSuccess)
+  {
+    std::cerr << "cub::DeviceSegmentedRadixSort::SortPairs (DoubleBuffer) failed with status: " << error << std::endl;
+  }
+
+  thrust::device_vector<int> expected_keys{6, 7, 8, 0, 3, 5, 9};
+  thrust::device_vector<int> expected_values{1, 2, 0, 5, 4, 3, 6};
+  // example-end segmented-radix-sort-pairs-db-env
+
+  stream.sync();
```
bernhardmgruber, line 262: Important: This sync should not be needed and is irritating. Please remove it.

cub/test/catch2_test_device_segmented_radix_sort_env_api.cu
```diff
@@ -157,3 +157,97 @@ C2H_TEST("cub::DeviceSegmentedRadixSort::SortKeysDescending env with stream", "[
...
+  {
+    std::cerr
+      << "cub::DeviceSegmentedRadixSort::SortPairsDescending (DoubleBuffer) failed with status: " << error << std::endl;
+  }
+
+  thrust::device_vector<int> expected_keys{8, 7, 6, 9, 5, 3, 0};
+  thrust::device_vector<int> expected_values{0, 2, 1, 6, 3, 4, 5};
+  // example-end segmented-radix-sort-pairs-descending-db-env
+
+  stream.sync();
```
bernhardmgruber, line 310: Same here.

cub/test/catch2_test_device_segmented_radix_sort_env_api.cu
gonidelis: need to sync because the comparison is in a different stream than the cub api call

cub/test/catch2_test_device_segmented_radix_sort_env_api.cu
gonidelis: need to sync because the comparison is in a different stream than the cub api call

cub/test/catch2_test_device_segmented_radix_sort_env_api.cu
bernhardmgruber: Discussed offline. Agreed.

cub/test/catch2_test_device_segmented_radix_sort_env_api.cu
bernhardmgruber: Agreed

Issue #8316, opened 2026-04-07
any_resource(resource_ref) construction breaks with proxy types due to missing non-template overload


bdice:
## Summary

tl;dr There is a discrepancy between the _documented_ API of `any_resource` and its actual implementation. [The docs say](https://nvidia.github.io/cccl/unstable/libcudacxx/api/classcuda_1_1mr_1_1basic__any__resource.html#_CPPv4I_13_ResourceKindDpEN4cuda2mr18basic_any_resource18basic_any_resourceE18basic_resource_refI10_OtherKindDp16_OtherPropertiesE) there is a constructor `any_resource(resource_ref)`, but passing an object that _converts_ to a `resource_ref` doesn't work. The _real_ constructor behaves differently than the documented constructor.

## Details

`any_resource<Props...>` can be constructed from a `resource_ref<Props...>` in pure C++, but this only works incidentally through the inherited forwarding constructor `__basic_any(_Tp&&)`. Any proxy type wrapping `resource_ref` that provides `operator T&()` (e.g. Cython's `__Pyx_FakeReference`) breaks template argument deduction, because the SFINAE constraint `__satisfies<decay_t<_Tp>, _Interface>` evaluates against the proxy type rather than `resource_ref`.

The documented API in `basic_any_resource` (line 506 of `any_resource.h`) declares:

```cpp
template <_ResourceKind _OtherKind, class... _OtherProperties>
basic_any_resource(basic_resource_ref<_OtherKind, _OtherProperties...> __res);
```

But the real `any_resource` struct at line 268 only has:

```cpp
_CCCL_DELEGATE_CONSTRUCTORS(any_resource, ::cuda::__basic_any, __iasync_resource<_Properties...>);
```

which expands to `using __base::__base`, inheriting only the constrained forwarding constructor. There is no dedicated non-template overload for `resource_ref`.

## Impact

This affects Cython (used by RAPIDS RMM, cuDF, etc.) which wraps C++ intermediate values in `__Pyx_FakeReference<T>` — a proxy with `operator T&()`. Constructing `any_resource` from a `resource_ref` obtained from Cython's `optional::value()` fails at compile time.

More generally, any codegen or proxy pattern that stores a `resource_ref` in a wrapper with `operator T&()` will hit the same issue.

## Proposed Fix

[...]

Issue #7397, opened 2026-01-28
[BUG]: `-std=c++XX` flag forms part of ABI for `__basic_any`


wence-:
### Is this a duplicate?

- [x] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cccl/issues) for this bug and that I agree to the [Code of Conduct](CODE_OF_CONDUCT.md)

### Type of Bug

Runtime Error

### Component

libcu++

### Describe the bug

The new memory resources use a custom type erased implementation in `cuda/__utility/basic_any`.

The ABI (and RTTI) of these types changes depending on the C++ standard selected when compiling things.

For example, a consequence is that if I have `libA.so` compiled with `-std=c++17`, and then executable `B` (linking to `libA.so`), compiled with `-std=c++20` then I cannot use the types across the DSO boundary.

### How to Reproduce

```sh
$ mkdir build
$ cmake -DCMAKE_BUILD_TYPE=Debug -DCMAKE_EXPORT_COMPILE_COMMANDS=1 -G Ninja  .. -DBUILD_SHARED_LIBS=ON
[...]
$ ninja
$ ./bug
bug: /home/coder/doodles/c++/cccl/build/_deps/cccl-src/lib/cmake/libcudacxx/../../../libcudacxx/include/cuda/__utility/__basic_any/virtual_ptrs.h:45: operator const _VTable*: Assertion `bad vtable cast detected' failed.
Aborted (core dumped)
[ble: exit 134]
```

```c++
// bug.cxx
#include <cuda/memory_resource>
#include <cuda/memory>
#include <cuda/devices>
#include <cuda/buffer>
#include <cuda/stream_ref>
#include <cuda_runtime_api.h>
#include "devbuf.h"

int main(void) {
  auto const &dev = cuda::devices[0];
  dev.init();
  auto mr = cuda::device_default_memory_pool(dev);

  auto stream = cuda::stream{dev};
  auto stream_ref = cuda::stream_ref{stream};

  auto buf = device_buffer(10, stream, mr);
}
```
```c++
// devbuf.h
#include <cuda/buffer>
#include <cuda/devices>
#include <cuda/memory>
#include <cuda/memory_resource>
#include <cuda/stream_ref>

struct device_buffer {
  device_buffer(std::size_t size, cuda::stream_ref stream,
                cuda::mr::resource_ref<cuda::mr::device_accessible> ref);

  ~device_buffer() noexcept;

[...]

PR #8320, opened 2026-04-07
Add non-template ref-to-value converting constructor to __basic_any

bdice:
## Description

closes https://github.com/NVIDIA/cccl/issues/8316

Add a non-template converting constructor `__basic_any(__basic_any<_Interface&> const&)` to the `__basic_any` value specialization, enabling implicit conversion from proxy types that wrap a `__basic_any` reference type.

The `__basic_any` value specialization provides a forwarding constructor `__basic_any(_Tp&&)` and template converting constructors `__basic_any(__basic_any<_OtherInterface> const&)`. When a proxy type wraps a `__basic_any<I&>` (e.g. a `resource_ref`) and provides `operator __basic_any<I&>&()`, neither path works: the forwarding constructor deduces the proxy type which doesn't satisfy the interface, and the template converting constructors can't deduce `_OtherInterface` from the proxy since implicit user-defined conversions are not considered during template argument deduction.

This occurs concretely with Cython's `__Pyx_FakeReference<T>`, which wraps intermediate C++ expression results and provides `operator T&()`. Constructing `any_resource<P...>` from a Cython-wrapped `resource_ref<P...>` fails at compile time.

The fix adds a non-template constructor for the same-interface ref-to-value conversion. Because the parameter is non-template, the compiler considers implicit conversion sequences through the proxy's `operator T&()`. The existing template converting constructor gains an additional `!same_as<_OtherInterface, _Interface&>` constraint to avoid ambiguity.

This single fix in `__basic_any` covers all downstream types: `any_resource`, `any_synchronous_resource`, and any future `__basic_any`-based value types.

**Testing:** A new test was added based on the minimal reproducer in the issue. RMM Cython bindings build and pass tests using direct `any_resource` construction from `resource_ref` without workaround helpers.

## Checklist


libcudacxx/include/cuda/__utility/__basic_any/basic_any_value.h
```diff
@@ -160,13 +160,25 @@ struct _CCCL_TYPE_VISIBILITY_DEFAULT __basic_any : __basic_any_base<_Interface>
   //! `__icopyable<>`.
   //! @post `has_value() == __other.has_value()`.
   _CCCL_TEMPLATE(class _OtherInterface)
-  _CCCL_REQUIRES((!::cuda::std::same_as<_OtherInterface, _Interface>)
-                   _CCCL_AND __any_convertible_to<__basic_any<_OtherInterface> const&, __basic_any>)
+  _CCCL_REQUIRES(
+    (!::cuda::std::same_as<_OtherInterface, _Interface>) _CCCL_AND(!::cuda::std::same_as<_OtherInterface, _Interface&>)
```
ericniebler: could be simplified to
```suggestion
    (!::cuda::std::same_as<_OtherInterface&, _Interface&>)
```

libcudacxx/include/cuda/__utility/__basic_any/basic_any_value.h
```diff
@@ -160,13 +160,25 @@ struct _CCCL_TYPE_VISIBILITY_DEFAULT __basic_any : __basic_any_base<_Interface>
...
+  //! @brief Non-template converting constructor from the corresponding
+  //! reference type `__basic_any<_Interface&>`. This enables implicit
+  //! conversion from proxy types (e.g. Cython's __Pyx_FakeReference)
+  //! that wrap a `__basic_any<_Interface&>` and provide
+  //! `operator __basic_any<_Interface&>&()`, since non-template parameters
+  //! participate in implicit conversion sequences.
+  _CCCL_API __basic_any(__basic_any<_Interface&> const& __other)
+  {
+    __convert_from(__other);
+  }
```
miscco: Thanks for the fix, please also add the corresponding move constructor and test for rvalue proxy

libcudacxx/test/libcudacxx/cuda/memory_resource/any_resource/any_resource.cu
```diff
@@ -348,4 +348,83 @@ TEST_CASE("regression test for NVIDIA/cccl#8037", "[container][resource]")
...
+  {
+    cuda::mr::any_resource<cuda::mr::host_accessible> any{ref};
+    // Extra parens prevent Catch2's expression decomposition, which triggers
+    // an nvcc bug with auto NTTP deduction in __satisfies during ADL.
+    CHECK((any == ref));
+  }
+
+#  if !_CCCL_CUDA_COMPILER(NVCC, <, 12, 9)
+  // nvcc before CTK 12.9 has a bug where auto NTTP deduction failures in
+  // __satisfies are hard errors instead of SFINAE during overload resolution.
```
bdice, line 386: Just raising a flag here. It seems like CUDA 12.0 didn't work with these changes. I wasn't able to come to a good workaround, but maybe someone else will have ideas. This is the MRE (sadly not very minimal): https://gist.github.com/bdice/0704486a9d32bceb782acd931c823123

I can't give a Godbolt reproducer link because it requires the changes from this PR...

libcudacxx/test/libcudacxx/cuda/memory_resource/any_resource/any_resource.cu
```diff
@@ -348,4 +348,83 @@ TEST_CASE("regression test for NVIDIA/cccl#8037", "[container][resource]")
...
+{
+  host_device_resource mr;
+  cuda::mr::resource_ref<cuda::mr::host_accessible> ref{mr};
+
+  SECTION("direct construction from resource_ref")
+  {
+    cuda::mr::any_resource<cuda::mr::host_accessible> any{ref};
+    // Extra parens prevent Catch2's expression decomposition, which triggers
+    // an nvcc bug with auto NTTP deduction in __satisfies during ADL.
+    CHECK((any == ref));
```
bdice, line 381: Also flagging this for reviewers' thoughts -- Claude suggested this and I didn't have a great alternative in mind.

libcudacxx/include/cuda/__utility/__basic_any/basic_any_value.h
```diff
@@ -160,13 +160,33 @@ struct _CCCL_TYPE_VISIBILITY_DEFAULT __basic_any : __basic_any_base<_Interface>
...
+  //! participate in implicit conversion sequences.
+  //! @pre `_Interface` must extend `__icopyable<>` (converting a ref to
+  //! a value requires copying the referenced object).
+  template <bool _Copyable = __copyable, ::cuda::std::enable_if_t<_Copyable, int> = 0>
+  _CCCL_API __basic_any(__basic_any<_Interface&> const& __other)
+  {
+    __convert_from(__other);
+  }
+
+  template <bool _Copyable = __copyable, ::cuda::std::enable_if_t<_Copyable, int> = 0>
```
pciolkosz, line 184: Should this be _CCCL_REQUIRES?

bdice: No, this was intentional. The `__basic_any` class uses raw `enable_if_t` instead of `_CCCL_TEMPLATE`/`_CCCL_REQUIRES` to maintain ABI compatibility across C++17 and C++20. The `_CCCL_REQUIRES` macro expands to a `requires`-clause in C++20 but to additional SFINAE template parameters in C++17, which produces different mangled names depending on the dialect. This caused runtime failures (vtable cast assertions) when `any_resource` was used across DSO boundaries compiled with different `-std=` flags. See #7397 for the original bug report and #7401 / #7405 for the fix.

libcudacxx/test/libcudacxx/cuda/memory_resource/any_resource/any_resource.cu
ericniebler: this is a fine workaround, thanks for the explanatory comment.

libcudacxx/include/cuda/__utility/__basic_any/basic_any_value.h
bdice: Done in 69b44f5.

libcudacxx/test/libcudacxx/cuda/memory_resource/any_resource/any_resource.cu
ericniebler: seems reasonable to exclude the tests that trigger a since-fixed compiler bug. particularly one that is hard to work around, like this one.

EDIT: i wonder if it is the same issue that i worked around here: https://github.com/NVIDIA/cccl/blob/main/libcudacxx/include/cuda/__utility/__basic_any/basic_any_value.h#L254-L255

libcudacxx/include/cuda/__utility/__basic_any/basic_any_value.h
ericniebler: we only needed to use `enable_if` for the functions whose ptrs actually end up in the pseudo-vtable (the ones marked `_CCCL_PUBLIC_API`). for these constructors, `_CCCL_REQUIRES` would not cause the ABI issue described in #7397, and it would be consistent with the other constructors in this class.

PR #8323, opened 2026-04-08
Add `DoubleBuffer` env overloads for `DeviceSegmentedRadixSort::SortKeys`

gonidelis:
Split 2/2 from #8100

### DIFF
--- a/cub/test/catch2_test_device_segmented_radix_sort_env_api.cu
+++ b/cub/test/catch2_test_device_segmented_radix_sort_env_api.cu
@@ -157,3 +157,65 @@ C2H_TEST("cub::DeviceSegmentedRadixSort::SortKeysDescending env with stream", "[
   REQUIRE(error == cudaSuccess);
   REQUIRE(keys_out == expected_keys);
 }
+
+C2H_TEST("cub::DeviceSegmentedRadixSort::SortKeys DoubleBuffer env with stream", "[segmented_radix_sort][env]")
+{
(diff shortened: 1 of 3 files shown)


cub/test/catch2_test_device_segmented_radix_sort_env_api.cu
```diff
@@ -157,3 +157,65 @@ C2H_TEST("cub::DeviceSegmentedRadixSort::SortKeysDescending env with stream", "[
...
+  if (error != cudaSuccess)
+  {
+    std::cerr
+      << "cub::DeviceSegmentedRadixSort::SortKeysDescending (DoubleBuffer) failed with status: " << error << std::endl;
+  }
+
+  thrust::device_vector<int> expected_keys{8, 7, 6, 9, 5, 3, 0};
+  // example-end segmented-radix-sort-keys-descending-db-env
+
+  stream.sync();
```
bernhardmgruber, line 215: Same.

gonidelis: we need the sync because these are happening on a different stream 

https://github.com/gonidelis/cccl/blob/382332973276125ff2a8fa6185b34ebdca3218ea/cub/test/catch2_test_device_segmented_radix_sort_env_api.cu#L186-L188

than the cub API call

bernhardmgruber: Yes. Discussed offline.

PR #8324, opened 2026-04-08
Disable extreme distribution tests that is failing in CI

miscco:
I could reproduce failures locally on blackwell devices

Fixes nvbug6007539

### DIFF
--- a/libcudacxx/test/support/random_utilities/test_distribution.h
+++ b/libcudacxx/test/support/random_utilities/test_distribution.h
@@ -197,7 +197,7 @@ __host__ __device__ bool test_eval(const typename D::param_type param, CDF cdf)
     {
       auto dist_val  = d1(g_1, param);
       auto dist2_val = d2(g_2);
-      assert(dist_val == dist2_val);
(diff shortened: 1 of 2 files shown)


jrhemstad:
> I could reproduce failures locally on blackwell devices

Does that not concern us?

libcudacxx/test/support/random_utilities/test_distribution.h
```diff
@@ -197,7 +197,7 @@ __host__ __device__ bool test_eval(const typename D::param_type param, CDF cdf)
     {
       auto dist_val  = d1(g_1, param);
       auto dist2_val = d2(g_2);
-      assert(dist_val == dist2_val);
+      assert((dist_val == dist2_val) || (cuda::std::isnan(dist_val) && cuda::std::isnan(dist2_val)));
```
miscco, line 200: @jrhemstad I investigated with @RAMitchell and it turns out that the extreme value distribution generates NaN on device, which compare unequal

davebayer (approved):

PR #8333, opened 2026-04-08
[libcudacxx] Add support for wider types in fill_bytes

pciolkosz:
## Summary
- Generalize `fill_bytes` to accept `uint16_t` and `uint32_t` fill values in addition to `uint8_t`, matching the CUDA driver's `cuMemsetD8Async`/`cuMemsetD16Async`/`cuMemsetD32Async` capabilities.
- Template `__fill_bytes_impl` on the fill value type, with a `static_assert` restricting to 1, 2, or 4 byte values.
- For wider fill values, validate that the destination size in bytes is a multiple of the fill value size.

## Test plan
- [ ] Existing `fill_bytes` tests continue to pass (uint8_t path unchanged)
- [ ] Header tests build across all compiler configurations
- [ ] Add tests for `uint16_t` and `uint32_t` fill patterns

### DIFF
--- a/libcudacxx/include/cuda/__algorithm/fill.h
+++ b/libcudacxx/include/cuda/__algorithm/fill.h
@@ -4,7 +4,7 @@
 // under the Apache License v2.0 with LLVM Exceptions.
 // See https://llvm.org/LICENSE.txt for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
-// SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES.
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES.
 //
 //===----------------------------------------------------------------------===//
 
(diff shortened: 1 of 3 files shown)


libcudacxx/include/cuda/__algorithm/fill.h
```diff
@@ -36,21 +36,40 @@ _CCCL_BEGIN_NAMESPACE_CUDA
...
+//! @param __value The fill pattern. Each element-sized unit of the destination is set to this value.
+//!
+//! When `sizeof(_ValueTy) > 1`, the destination size in bytes must be a multiple of `sizeof(_ValueTy)`.
+template <typename _ValueTy, typename _DstTy, ::cuda::std::size_t _DstSize>
+_CCCL_HOST_API void __fill_bytes_impl(stream_ref __stream, ::cuda::std::span<_DstTy, _DstSize> __dst, _ValueTy __value)
 {
   static_assert(!::cuda::std::is_const_v<_DstTy>, "Fill destination can't be const");
   static_assert(::cuda::std::is_trivially_copyable_v<_DstTy>);
+  static_assert(sizeof(_ValueTy) == 1 || sizeof(_ValueTy) == 2 || sizeof(_ValueTy) == 4,
+                "Fill value must be 1, 2, or 4 bytes (matching CUDA driver memset support)");
```
Jacobfaib: This check is already done in `__memsetAsync()` (though the message there could be improved).

libcudacxx/include/cuda/__algorithm/fill.h
```diff
@@ -91,6 +131,24 @@ _CCCL_HOST_API void fill_bytes(stream_ref __stream, _DstTy&& __dst, ::cuda::std:
     __stream, __as_mdspan(launch_transform(__stream, ::cuda::std::forward<_DstTy>(__dst))), __value);
 }
 
+//! @overload
+_CCCL_TEMPLATE(typename _DstTy)
+_CCCL_REQUIRES(__mdspannable<transformed_device_argument_t<_DstTy>>)
+_CCCL_HOST_API void fill_bytes(stream_ref __stream, _DstTy&& __dst, ::cuda::std::uint16_t __value)
```
davebayer, line 135: Q: why do we need these functions when we will get `cuda::std::fill` with gpu execution policy in this release?

pciolkosz: Because this is not exactly the `std::` interface, this is direct replacement for async memcpy. I think this works better for a lot of use cases

fbusato:
a natural extension is to support any `mdspan` in a similar way of `cudax::copy_bytes`. I will open a new issue for that

libcudacxx/include/cuda/__algorithm/fill.h
```diff
@@ -91,6 +129,24 @@ _CCCL_HOST_API void fill_bytes(stream_ref __stream, _DstTy&& __dst, ::cuda::std:
     __stream, __as_mdspan(launch_transform(__stream, ::cuda::std::forward<_DstTy>(__dst))), __value);
 }
```
davebayer, line 130: Shouldn't we provide an overload that takes `cuda::std::byte`?

davebayer (changes requested):
I've just realized - this is a source breaking change. See https://godbolt.org/z/Enq6G4zvG

Could we provide something like `cuda::byte2` and `cuda::byte4` implemented as:
```cpp
enum class byte2 : unsigned short {};
enum class byte4 : unsigned {};
```
which could be passed as:
```cpp
cuda::fill_bytes(stream, dst, cuda::std::byte{0xff});
cuda::fill_bytes(stream, dst, cuda::byte2{0xf0f0});
cuda::fill_bytes(stream, dst, cuda::byte4{0xff00'ff00};
```
Or something else?

libcudacxx/include/cuda/__algorithm/fill.h
```diff
@@ -63,16 +80,19 @@ _CCCL_HOST_API void __fill_bytes_impl(stream_ref __stream,
...
+//! The fill value can be 1, 2, or 4 bytes wide. The destination size in bytes must be a multiple of
+//! the fill value size.
+//!
+//! @param __stream Stream that the fill should be inserted into
 //! @param __dst Destination memory to fill
-//! @param __value Value to fill into every byte in the destination
+//! @param __value Value to fill into the destination (1, 2, or 4 bytes)
 _CCCL_TEMPLATE(typename _DstTy)
 _CCCL_REQUIRES(__spannable<transformed_device_argument_t<_DstTy>>)
 _CCCL_HOST_API void fill_bytes(stream_ref __stream, _DstTy&& __dst, ::cuda::std::uint8_t __value)
```
miscco, line 98: Question: What about signed integer types? Should we provide overloads for those?

libcudacxx/include/cuda/__algorithm/fill.h
```diff
@@ -36,21 +36,38 @@ _CCCL_BEGIN_NAMESPACE_CUDA
...
+//!
+//! When `sizeof(_ValueTy) > 1`, the destination size in bytes must be a multiple of `sizeof(_ValueTy)`.
+template <typename _ValueTy, typename _DstTy, ::cuda::std::size_t _DstSize>
+_CCCL_HOST_API void __fill_bytes_impl(stream_ref __stream, ::cuda::std::span<_DstTy, _DstSize> __dst, _ValueTy __value)
 {
   static_assert(!::cuda::std::is_const_v<_DstTy>, "Fill destination can't be const");
   static_assert(::cuda::std::is_trivially_copyable_v<_DstTy>);
 
+  const auto __size_bytes = __dst.size_bytes();
+  if constexpr (sizeof(_ValueTy) > 1)
```
miscco, line 54: Shouldnt we check that sizeof(_ValueTy ) <= sizeof(uint32_t)

libcudacxx/include/cuda/__algorithm/fill.h
```diff
@@ -81,6 +101,24 @@ _CCCL_HOST_API void fill_bytes(stream_ref __stream, _DstTy&& __dst, ::cuda::std:
     __stream, ::cuda::std::span(launch_transform(__stream, ::cuda::std::forward<_DstTy>(__dst))), __value);
 }
 
+//! @overload
+_CCCL_TEMPLATE(typename _DstTy)
+_CCCL_REQUIRES(__spannable<transformed_device_argument_t<_DstTy>>)
+_CCCL_HOST_API void fill_bytes(stream_ref __stream, _DstTy&& __dst, ::cuda::std::uint16_t __value)
```
miscco, line 107: Important: I believe we should be able to relax the constraints here. 

We should be able to accept all trivially_copyable types (probably with unique_bit_representation) of size [8, 32] and just `bit_cast` to the right unsigned integer type

That would allow us to condense everything into just one function. 

We could / should also collapse the `__spannable` / `__mdspannable` constraint into an `if constexpr` chain with a proper error message at the end

PR #8346, opened 2026-04-09
Remove taplo-pre-commit hook (no aarch64 wheels)

### LINKED ISSUE #8345: Replace taplo-pre-commit with check-toml (taplo has no Linux aarch64 wheels)
## Problem

`taplo-pre-commit` (v0.9.3) does not ship pre-built wheels for Linux aarch64. When `pre-commit run -a` is invoked on an aarch64 host, the hook falls back to building from source — automatically downloading and installing a Rust toolchain via puccinialin — then fails because the source tarball is missing `crates/taplo-lsp/Cargo.toml`:

```
Computed rustc target triple: aarch64-unknown-linux-gnu
Rust not found, installing into a temporary directory
...
error: failed to load manifest for workspace member `.../crates/taplo-cli`
  Caused by: failed to read `.../crates/taplo-lsp/Cargo.toml`
  Caused by: No such file or directory (os error 2)
```

This means contributors on Linux aarch64 cannot run pre-commit at all.

## Proposal

Remove the `taplo-pre-commit` hook entry from `.pre-commit-config.yaml` and rely on the existing `check-toml` hook (from `pre-commit-hooks`, already configured on line 22) for TOML syntax validation. `check-toml` is a pure-Python hook that works on every platform without a Rust toolchain.

`check-toml` validates TOML syntax but does not format. If TOML formatting enforcement is desired in the future, it could be added as a separate CI step using a pinned taplo binary that supports aarch64, rather than through a pre-commit hook that requires compilation.

bdice:
## Description

closes #8345

`taplo-pre-commit` (v0.9.3) does not ship pre-built wheels for Linux aarch64. When `pre-commit run -a` is invoked on an aarch64 host, the hook falls back to building from source — automatically installing a Rust toolchain — then fails because the source tarball is incomplete. This removes the hook entirely, relying on the existing `check-toml` hook for TOML syntax validation.

## Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -59,16 +59,6 @@ repos:
     - id: ruff  # linter
     - id: ruff-format  # formatter
 
-  # TOML lint & format
-  - repo: https://github.com/ComPWA/taplo-pre-commit
-    rev: v0.9.3
-    hooks:
-      # See https://github.com/NVIDIA/cccl/issues/3426
-      # - id: taplo-lint
-      #   exclude: "^docs/"
-      - id: taplo-format
-        exclude: "^docs/"
-
   # CMake formatting
   - repo: https://github.com/BlankSpruce/gersemi
     rev: 0.23.1


rwgk (approved):
Yay! (sorry for the trouble; this was an unsuccesful attempt to do something that was meant to be mainstream)

elstehle (approved):

PR #8347, opened 2026-04-09
`__aggregate_all_of` Internal utility to recursively apply a predicate to aggregate members

fbusato:
## Description

`__aggregate_all_of` is useful to extract properties for custom traits.

Extract from:

- https://github.com/NVIDIA/cccl/pull/8265


fbusato:
@miscco @davebayer @gevtushenko considering the complexity, the impact on compile time, and a couple of compatibility problems with MSVC, I'm not 100% sure it worth to add it.

details:

- compile time: the trait adds an initializer one by one, until it reaches the maximum.
- MSVC doesn't support arrays and nested structures.

wmaxey:
>     * MSVC doesn't support arrays and nested structures.

I remember vaguely finding this trick before and decided to write it off based on this. I think we should wait for proper reflection unless this immediately solves problems of our own making internally where we can control what goes through the machinery.

If we can get a guarantee that this will only be used with well supported input types, then I'd consider approving this. Otherwise leaving a footgun laying around concerns me.

miscco (approved):
Moved to inline variables

PR #8349, opened 2026-04-09
Fix typo in .clang-tidy that left noexcept-swap disabled

Jacobfaib:
## Description

closes 

## Checklist

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -21,7 +21,6 @@ Checks:
   # TODO: BEGIN REMOVE ME
   - '-*'
   - 'performance-*'
-  - '-performance-noexcept-swap'
   - '-performance-enum-size'
   - '-performance-unnecessary-value-param'
   # END REMOVE ME


libcudacxx/include/cuda/std/__tuple_dir/tuple.h
```diff
@@ -43,6 +43,7 @@
 
 _CCCL_BEGIN_NAMESPACE_CUDA_STD
 
+// dummy edit for CI
```
bernhardmgruber: You can also just `git commit --amend --no-edit` to change the last commit's SHA and then force push.

Jacobfaib: The point was to get CI to think that C++ files were changed (the clang-tidy job only runs if C++ sources were changed), not to add a commit to trigger CI.

alliepiper (approved):

PR #8351, opened 2026-04-09
Fix and improve vector type traits

davebayer:
We are missing the `template <class T> struct trait { ... };` trait, we only expose the `trait_v` or `trait_t` variants. This PR fixes that.

I've also rewritten the code so we don't pull in tuple element/size. This fixes a bug, because we don't currently implement tuple protocol for extended nvfp vector types, thus `scalar_type_t` wasn't working properly.

This PR also prevents inclusion of `<cuda_fpXY.h>` headers in these traits.


docs/libcudacxx/extended_api/type_traits/vector_types.rst
```diff
@@ -5,19 +5,22 @@ Vector Type Traits
...
+   struct is_vector_type : cuda::std::bool_constant</* see below */> {};
+
    template <class T>
-   inline constexpr bool is_vector_type_v = /* see below */;
+   inline constexpr bool is_vector_type_v = is_vector_type<T>::value;
 
    } // namespace cuda
 
-``cuda::is_vector_type_v<T>`` is ``true`` if ``T`` is a CUDA vector type, such as ``int2``, ``float4``, or ``dim3``.
+``cuda::is_vector_type<T>`` is ``true`` if ``T`` is a cv-qualified CUDA vector type, such as ``int2``, ``float4``, or ``dim3``.
```
davebayer, line 23: `is_trait`s always work for cv-qualified types

docs/libcudacxx/extended_api/type_traits/vector_types.rst
```diff
@@ -58,33 +64,45 @@ Defined in the ``<cuda/type_traits>`` header.
...
+     using type = /* see below */;
+   };
+
+   template <class T>
+   using scalar_type_t = typename scalar_type<T>::type;
 
    } // namespace cuda
 
-``cuda::scalar_type_t<VectorType>`` extracts the scalar element type from a CUDA vector type. For example, ``cuda::scalar_type_t<int2>`` is ``int`` and ``cuda::scalar_type_t<float4>`` is ``float``. . It supports both integral and floating-point vector types.
+``cuda::scalar_type<T>::type`` extracts the scalar element type from a cv-qualified CUDA vector type. For example, ``cuda::scalar_type<int2>::type`` is ``int`` and ``cuda::scalar_type<float4>::type`` is ``float``. . It supports both integral and floating-point vector types. If ``T`` is not a vector type, the ``type`` is ``void``.
```
davebayer, line 105: I've made this consistent with `vector_type` and `type` is `void` if `_Tp` is not a known vector type

libcudacxx/include/cuda/__type_traits/vector_type.h
```diff
@@ -24,458 +24,398 @@
...
 
 #  include <cuda/std/__cstddef/types.h>
-#  include <cuda/std/__floating_point/cuda_fp_types.h>
-#  include <cuda/std/__tuple_dir/vector_types.h>
+#  include <cuda/std/__type_traits/integral_constant.h>
 #  include <cuda/std/__type_traits/is_same.h>
 
-#  if !_CCCL_CUDA_COMPILATION()
-#    include <vector_types.h>
-#  endif // !_CCCL_CUDA_COMPILATION()
```
davebayer, line 33: This header is included by `<cuda_runtime.h>` which we include always if available

docs/libcudacxx/extended_api/type_traits/vector_types.rst
```diff
@@ -5,19 +5,22 @@ Vector Type Traits
...
 .. code:: cuda
 
    namespace cuda {
 
+   template <class T, cuda::std::size_t Size>
+   struct is_vector_type : cuda::std::bool_constant</* see below */> {};
+
    template <class T>
-   inline constexpr bool is_vector_type_v = /* see below */;
+   inline constexpr bool is_vector_type_v = is_vector_type<T>::value;
```
miscco, line 19: We do not want that at all, users should always specialize the variable template not instantiate a struct

miscco (changes requested):

docs/libcudacxx/extended_api/type_traits/vector_types.rst
davebayer: But these are not for users to specialize

libcudacxx/include/cuda/__type_traits/scalar_type.h
```diff
@@ -0,0 +1,467 @@
...
+};
+template <class _Tp>
+struct _CCCL_TYPE_VISIBILITY_DEFAULT scalar_type<const _Tp> : scalar_type<_Tp>
+{};
+template <class _Tp>
+struct _CCCL_TYPE_VISIBILITY_DEFAULT scalar_type<volatile _Tp> : scalar_type<_Tp>
+{};
+template <class _Tp>
+struct _CCCL_TYPE_VISIBILITY_DEFAULT scalar_type<const volatile _Tp> : scalar_type<_Tp>
+{};
```
miscco, line 44: Those should be variable templates not structs

docs/libcudacxx/extended_api/type_traits/vector_types.rst
miscco: The way it is written would suggest that a user could specialize them

Also we never want those as types, but only as variable templates

davebayer: We don't specify it anywherer. We probably should.

I mean all type traits are provided as `struct` and variable/alias template. We do that for `cuda::is_floating_point`, too. I think we should stay consistent.

libcudacxx/include/cuda/__type_traits/scalar_type.h
davebayer: I don't follow, can you elaborate?

docs/libcudacxx/extended_api/type_traits/vector_types.rst
miscco: we should not provide structs the are completely unnecessary

libcudacxx/include/cuda/__type_traits/scalar_type.h
miscco: we should not use structs for traits in new code

davebayer: But if I don't use `struct`, I would need the type to be fully defined, which would require inclusion of `<cuda_fpXY.h>` headers. Note that this trait returns type, thus I cannot use template variable

docs/libcudacxx/extended_api/type_traits/vector_types.rst
davebayer: So why do we provide `cuda::is_floating_point`? How would you explain to the user that all type traits have `struct` version, except this one? What makes this one special?

fbusato: I kind agree with @miscco. We are introducing many structures for no benefit. I would just introduce an alias (like in `cuda::is_floating_point`). These traits should not be specialized by the user.

docs/libcudacxx/extended_api/type_traits/vector_types.rst
```diff
@@ -58,33 +64,45 @@ Defined in the ``<cuda/type_traits>`` header.
...
 -----------------------
 
 .. code:: cuda
 
    namespace cuda {
 
-   template <class VectorType>
-   using scalar_type_t = /* see below */;
+   template <class T>
+   struct scalar_type
```
miscco, line 95: Should this requires that T is a vector type?

libcudacxx/include/cuda/__type_traits/is_vector_type.h
```diff
@@ -22,209 +22,30 @@
...
-
-template <typename _Tp>
-inline constexpr bool is_extended_fp_vector_type_v = false;
-
-#  if _CCCL_HAS_NVFP16()
-template <>
-inline constexpr bool is_extended_fp_vector_type_v<::__half2> = true;
-#  endif // _CCCL_HAS_NVFP16()
+template <class _Tp>
+inline constexpr bool is_vector_type_v = (vector_size_v<_Tp> > 0);
```
miscco:
```suggestion
inline constexpr bool is_vector_type_v = (vector_size_v<_Tp> != 0);
```

libcudacxx/include/cuda/__type_traits/scalar_type.h
```diff
@@ -0,0 +1,467 @@
...
+#  include <cuda/std/__cccl/prologue.h>
+
+_CCCL_BEGIN_NAMESPACE_CUDA
+_CCCL_SUPPRESS_DEPRECATED_PUSH
+
+template <class _Tp>
+struct _CCCL_TYPE_VISIBILITY_DEFAULT scalar_type
+{
+  using type = void;
+};
```
miscco: Should this type just be empty?
```suggestion
{};
```

libcudacxx/include/cuda/__type_traits/vector_type.h
```diff
@@ -24,458 +24,397 @@
...
-    else if constexpr (_Size == 3)
-    {
-      return ::long3{};
-    }
-    else if constexpr (_Size == 4)
-    {
+struct _CCCL_TYPE_VISIBILITY_DEFAULT vector_type
+{
+  using type = void;
+};
```
miscco, line 40: We should just keep it empty
```suggestion
{};
```

davebayer: We cannot do that unfortunately. This is already documented and released :/

miscco (approved):

fbusato (approved):
that's a nice refactoring. It avoid a lot of included header code

PR #8354, opened 2026-04-10
additional constrains for `layout_stride_relaxed`

fbusato:
## Description

@NaderAlAwar found that `layout_stride_relaxed` could produce a layout with a "negative" size that is not compensated by the offset. 
This PR adds several assertion to prevent the problem and covers the testing of these cases.

### DIFF
--- a/libcudacxx/include/cuda/__mdspan/layout_stride_relaxed.h
+++ b/libcudacxx/include/cuda/__mdspan/layout_stride_relaxed.h
@@ -261,9 +261,11 @@ public:
       // The dot product of indices and strides is linear.
       // Thus, over all valid indices, the max value of the dot product is achieved at the extrema: either the min
       // index (0) if the stride is negative, or the max index (extent(r) - 1) if the stride is non-negative.
-      // For non-negative stride: contribution is (extent - 1) * stride
-      // For negative stride: contribution is 0 (max achieved at index 0)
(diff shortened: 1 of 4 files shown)


libcudacxx/include/cuda/__mdspan/layout_stride_relaxed.h
```diff
@@ -261,25 +261,44 @@ class _CCCL_DECLSPEC_EMPTY_BASES layout_stride_relaxed::mapping
...
       {
         const auto __ext = extents().extent(__r);
         if (__ext == index_type{0})
         {
           return index_type{0};
         }
-        _CCCL_ASSERT(!::cuda::overflow_cast<index_type>(::cuda::uabs(strides().stride(__r))),
+        const auto __stride_val = strides().stride(__r);
+        _CCCL_ASSERT(!::cuda::overflow_cast<index_type>(::cuda::uabs(__stride_val)),
                      "layout_stride_relaxed::mapping: stride is out of range");
```
davebayer:
```suggestion
        _CCCL_ASSERT(::cuda::std::in_range<index_type>(__stride_val),
"layout_stride_relaxed::mapping: stride is out of range");
```
applies to other places, too

miscco (approved):

davebayer (approved):

PR #8355, opened 2026-04-10
[cub]: implement utilities for policy selection

davebayer:
This PR implements several utilities for policy selection. The main advantage is that they can potentially return arch-specific and family-specific architecture ids.

### DIFF
--- a/cub/cub/device/dispatch/kernels/kernel_for_each.cuh
+++ b/cub/cub/device/dispatch/kernels/kernel_for_each.cuh
@@ -17,6 +17,7 @@
 #include <cub/detail/mdspan_utils.cuh> // is_sub_size_static
 #include <cub/detail/type_traits.cuh> // implicit_prom_t
 #include <cub/device/dispatch/tuning/tuning_for.cuh>
+#include <cub/util_arch.cuh>
 
 #include <cuda/std/__type_traits/enable_if.h>
 #include <cuda/std/__type_traits/integral_constant.h>
@@ -92,7 +93,7 @@ template <class PolicySelector, class OffsetT, class OpT>
 #endif // _CCCL_HAS_CONCEPTS()
 _CCCL_KERNEL_ATTRIBUTES void dynamic_kernel(_CCCL_GRID_CONSTANT const OffsetT num_items, OpT op)
 {
-  static constexpr for_policy policy = PolicySelector{}(::cuda::arch_id{CUB_PTX_ARCH / 10});
+  static constexpr for_policy policy = current_policy<PolicySelector>();
   using agent_policy_t               = policy_t<policy.block_threads, policy.items_per_thread>;
   using agent_t                      = agent_block_striped_t<agent_policy_t, OffsetT, OpT>;
 
@@ -118,10 +119,10 @@ template <class PolicySelector, class OffsetT, class OpT>
   requires for_policy_selector<PolicySelector>
 #endif // _CCCL_HAS_CONCEPTS()
 _CCCL_KERNEL_ATTRIBUTES //
-__launch_bounds__(int(PolicySelector{}(::cuda::arch_id{CUB_PTX_ARCH / 10}).block_threads)) //
+__launch_bounds__(int(current_policy<PolicySelector>().block_threads)) //
   void static_kernel(_CCCL_GRID_CONSTANT const OffsetT num_items, OpT op)
 {
-  static constexpr for_policy policy = PolicySelector{}(::cuda::arch_id{CUB_PTX_ARCH / 10});
+  static constexpr for_policy policy = current_policy<PolicySelector>();
   using agent_policy_t               = policy_t<policy.block_threads, policy.items_per_thread>;
   using agent_t                      = agent_block_striped_t<agent_policy_t, OffsetT, OpT>;
 
--- a/cub/cub/util_arch.cuh
+++ b/cub/cub/util_arch.cuh
@@ -25,6 +25,8 @@
 
 #include <cuda/__cmath/ceil_div.h>
 #include <cuda/__cmath/round_up.h>
+#include <cuda/__device/arch_id.h>
+#include <cuda/__device/compute_capability.h>
 #include <cuda/std/__algorithm/clamp.h>
 #include <cuda/std/__algorithm/max.h>
 #include <cuda/std/__algorithm/min.h>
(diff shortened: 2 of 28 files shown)


cub/cub/device/dispatch/kernels/kernel_for_each.cuh
```diff
@@ -92,7 +93,7 @@ template <class PolicySelector, class OffsetT, class OpT>
 #endif // _CCCL_HAS_CONCEPTS()
 _CCCL_KERNEL_ATTRIBUTES void dynamic_kernel(_CCCL_GRID_CONSTANT const OffsetT num_items, OpT op)
 {
-  static constexpr for_policy policy = PolicySelector{}(::cuda::arch_id{CUB_PTX_ARCH / 10});
+  static constexpr for_policy policy = CUB_NS_QUALIFIER::detail::select_policy_for_current_arch<PolicySelector>();
```
miscco: This is not really an improvement in my opinion

davebayer: As I said in the description, now the policy selector gets info whether the target architecture is `a` or `f`, too

bernhardmgruber: I actually like this new approach! It hides the abomination that `::cuda::arch_id{CUB_PTX_ARCH / 10}` is. I think it's a bit too wordy. Full qualification is fine, but usually not what we do in CUB.

Given this will be used in every kernel, I think we could allow us a more concise name that people have to lookup when they see it the first time:

```suggestion
  static constexpr for_policy policy = CUB_NS_QUALIFIER::detail::current_policy<PolicySelector>();
```

cub/cub/util_arch.cuh
```diff
@@ -175,6 +177,42 @@ struct NoScaling
...
+  // else
+  // {
+  //   return PolicySelector{}(::cuda::compute_capability{arch_id});
+  // }
+  return PolicySelector{}(::cuda::arch_id{::cuda::compute_capability{arch_id}.get()});
+}
+
+// Selects the tuning policy for the current_tuning_arch() architecture.
+template <class PolicySelector>
+[[nodiscard]] _CCCL_API constexpr auto select_policy_for_current_arch()
```
bernhardmgruber: Suggestion: I would love if this function was device-only, since it will give wrong results when called from host code (where the current compilation-pass' arch does not exist):
```suggestion
template <class PolicySelector>
[[nodiscard]] _CCCL_DEVICE_API constexpr auto select_policy_for_current_arch()
```

cub/cub/util_arch.cuh
```diff
@@ -175,6 +177,42 @@ struct NoScaling
...
+  // if constexpr (::cuda::std::is_invocable_v<PolicySelector, ::cuda::arch_id>)
+  // {
+  //   return PolicySelector{}(arch_id);
+  // }
+  // else
+  // {
+  //   return PolicySelector{}(::cuda::compute_capability{arch_id});
+  // }
+  return PolicySelector{}(::cuda::arch_id{::cuda::compute_capability{arch_id}.get()});
+}
```
bernhardmgruber, line 211: Suggestion: I would not need this overload, since the policy selector is essentially a callable that turns an `arch_id` into a tuning policy. This function just handles the creation of an instance from a type, which for me is too little reason for it to exist.

About the handling of policies that accept compute capabilities instead of arch ids: My goal is that we can just switch to compute capabilities entirely and never implement support for arch_ids. No user asked to support arch_ids to handle a and f arches.

bernhardmgruber (approved):

bernhardmgruber:
CCCL.C seems to fail to build with:
```
  FAILED: [code=1] _parent_cccl/c/parallel/CMakeFiles/cccl.c.parallel.dir/src/scan.cu.o
  /usr/local/cuda/bin/nvcc -forward-unknown-to-host-compiler -ccbin=/opt/rh/gcc-toolset-13/root/usr/bin/g++ -DCCCL_C_EXPERIMENTAL=1 -DCUB_DEFINE_RUNTIME_POLICIES -DCUB_DISABLE_CDP=1 -DNVRTC_GET_TYPE_NAME=1 -D_CCCL_NO_SYSTEM_HEADER -Dcccl_c_parallel_EXPORTS -I/workspace/c/parallel/include -I/workspace/c/parallel/src -I/workspace/lib/cmake/cub/../../../cub -I/workspace/lib/cmake/libcudacxx/../../../libcudacxx/include -I/workspace/lib/cmake/thrust/../../../thrust -isystem /usr/local/cuda/include -isystem /workspace/python/cuda_cccl/build/cp313-cp313-linux_x86_64/_deps/json-src/include -O3 -DNDEBUG -std=c++20 --generate-code=arch=compute_52,code=[compute_52,sm_52] -Xcompiler=-fPIC --extended-lambda -Xcudafe=--display_error_number -Wno-deprecated-gpu-targets -Xcudafe=--promote_warnings -Wreorder -Xcompiler=-Werror -Xcompiler=-Wall -Xcompiler=-Wextra -Xcompiler=-Wreorder -Xcompiler=-Winit-self -Xcompiler=-Woverloaded-virtual -Xcompiler=-Wcast-qual -Xcompiler=-Wpointer-arith -Xcompiler=-Wunused-local-typedefs -Xcompiler=-Wvla -Xcompiler=-Wimplicit-fallthrough -Xcompiler=-fdiagnostics-show-template-tree -Xcompiler=-Wignored-qualifiers -Xcompiler=-Wmissing-field-initializers -Xcompiler=-Wsign-compare -Xcompiler=-Walloc-zero -Xcompiler=-Wtsan -Xcompiler=-Wenum-conversion -Xcompiler=-Wpacked -Xcompiler=-ftemplate-backtrace-limit=0 -Xcompiler=-Wno-gnu-line-marker -Xcompiler=-Wno-gnu-zero-variadic-macro-arguments -Xcompiler=-Wno-unused-function -Xcompiler=-Wno-noexcept-type -MD -MT _parent_cccl/c/parallel/CMakeFiles/cccl.c.parallel.dir/src/scan.cu.o -MF _parent_cccl/c/parallel/CMakeFiles/cccl.c.parallel.dir/src/scan.cu.o.d -x cu -c /workspace/c/parallel/src/scan.cu -o _parent_cccl/c/parallel/CMakeFiles/cccl.c.parallel.dir/src/scan.cu.o
  /workspace/lib/cmake/libcudacxx/../../../libcudacxx/include/cuda/__device/arch_id.h(185): error: static assertion failed with "unknown CUDA architecture"
      static_assert(__is_known_cc, "unknown CUDA architecture");
      ^
```
The problem is that nvcc 12.9 is used for compilation and SM52 (`--generate-code=arch=compute_52,code=[compute_52,sm_52]`) is used as target, which is officially unsupported by CCCL and thus we cannot convert `520` into a `cuda::arch_id`. Can a python/CCCL.C maintainer please fix this in the build system? We should compile for at least SM75.

PR #8356, opened 2026-04-10
Ensure `cuda::std::complex` is trivially copyable when possible

miscco:
With C++23  `std::complex` has been changed to be trivially copyable

Adopt those changes for our implementations.

Unfortunately, we cannot make `__half` of `__nv_bfloat16` work because we do not controll those

### DIFF
--- a/libcudacxx/include/cuda/std/__complex/complex.h
+++ b/libcudacxx/include/cuda/std/__complex/complex.h
@@ -31,6 +31,11 @@
 #include <cuda/std/__type_traits/is_constructible.h>
 #include <cuda/std/__type_traits/is_floating_point.h>
 #include <cuda/std/__type_traits/is_integral.h>
+#include <cuda/std/__type_traits/is_nothrow_copy_assignable.h>
+#include <cuda/std/__type_traits/is_nothrow_copy_constructible.h>
+#include <cuda/std/__type_traits/is_nothrow_default_constructible.h>
+#include <cuda/std/__type_traits/is_nothrow_move_assignable.h>
+#include <cuda/std/__type_traits/is_nothrow_move_constructible.h>
 #include <cuda/std/__type_traits/is_same.h>
 #include <cuda/std/cmath>
 #include <cuda/std/cstdint>
@@ -76,10 +81,17 @@ class _CCCL_TYPE_VISIBILITY_DEFAULT _LIBCUDACXX_COMPLEX_ALIGNAS complex
 public:
   using value_type = _Tp;
 
-  _CCCL_API constexpr complex(const value_type& __re = value_type(), const value_type& __im = value_type())
+  _CCCL_API constexpr complex(const value_type& __re = value_type(),
+                              const value_type& __im = value_type()) noexcept(is_nothrow_default_constructible_v<_Tp>)
       : __re_(__re)
       , __im_(__im)
   {}
+
+  _CCCL_HIDE_FROM_ABI constexpr complex(const complex&) noexcept(is_nothrow_copy_constructible_v<_Tp>) = default;
+  _CCCL_HIDE_FROM_ABI constexpr complex(complex&&) noexcept(is_nothrow_move_constructible_v<_Tp>)      = default;
+
+  _CCCL_HIDE_FROM_ABI constexpr complex& operator=(const complex&) noexcept(is_nothrow_copy_assignable_v<_Tp>) = default;
+  _CCCL_HIDE_FROM_ABI constexpr complex& operator=(complex&&) noexcept(is_nothrow_move_assignable_v<_Tp>) = default;
 
   template <class _Up, enable_if_t<__cccl_internal::__is_non_narrowing_convertible<_Tp, _Up>::value, int> = 0>
   _CCCL_API constexpr complex(const complex<_Up>& __c)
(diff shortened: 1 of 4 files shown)


libcudacxx/include/cuda/std/__complex/complex.h
```diff
@@ -76,11 +81,18 @@ class _CCCL_TYPE_VISIBILITY_DEFAULT _LIBCUDACXX_COMPLEX_ALIGNAS complex
 public:
   using value_type = _Tp;
 
-  _CCCL_API constexpr complex(const value_type& __re = value_type(), const value_type& __im = value_type())
+  _CCCL_API constexpr complex(const value_type& __re = value_type(),
+                              const value_type& __im = value_type()) noexcept(is_nothrow_default_constructible_v<_Tp>)
       : __re_(__re)
       , __im_(__im)
   {}
```
davebayer, line 88: I don't think the condition in `noexcept` is needed, see requirements on [numeric types](https://en.cppreference.com/w/cpp/named_req/NumericType.html)

libcudacxx/include/cuda/std/__complex/complex.h
```diff
@@ -76,11 +81,18 @@ class _CCCL_TYPE_VISIBILITY_DEFAULT _LIBCUDACXX_COMPLEX_ALIGNAS complex
...
 
+  constexpr complex(const complex&) noexcept(is_nothrow_copy_constructible_v<_Tp>) = default;
+  constexpr complex(complex&&) noexcept(is_nothrow_move_constructible_v<_Tp>)      = default;
+
+  constexpr complex& operator=(const complex&) noexcept(is_nothrow_copy_assignable_v<_Tp>) = default;
+  constexpr complex& operator=(complex&&) noexcept(is_nothrow_move_assignable_v<_Tp>)      = default;
```
davebayer:
```suggestion
  _CCCL_HIDE_FROM_ABI constexpr complex(const complex&) noexcept(is_nothrow_copy_constructible_v<_Tp>) = default;
  _CCCL_HIDE_FROM_ABI constexpr complex(complex&&) noexcept(is_nothrow_move_constructible_v<_Tp>)      = default;

  _CCCL_HIDE_FROM_ABI constexpr complex& operator=(const complex&) noexcept(is_nothrow_copy_assignable_v<_Tp>) = default;
  _CCCL_HIDE_FROM_ABI constexpr complex& operator=(complex&&) noexcept(is_nothrow_move_assignable_v<_Tp>)      = default;
```

libcudacxx/include/cuda/std/__complex/complex.h
miscco: I want to be safe because users are users and may specialize accordingly

davebayer: But that's UB:
*The behavior is unspecified (and may fail to compile) if T is not a cv-unqualified standard(until C++23) floating-point type and undefined if T is not [NumericType](https://en.cppreference.com/w/cpp/named_req/NumericType.html).*

libcudacxx/include/cuda/std/__complex/nvbf16.h
```diff
@@ -114,10 +114,18 @@ class _CCCL_TYPE_VISIBILITY_DEFAULT _CCCL_ALIGNAS(alignof(__nv_bfloat162)) compl
 public:
   using value_type = __nv_bfloat16;
 
-  _CCCL_API inline complex(const value_type& __re = value_type(), const value_type& __im = value_type())
+  _CCCL_API inline complex(const value_type& __re = value_type(), const value_type& __im = value_type()) noexcept
       : __repr_(__re, __im)
   {}
 
+#  if !_CCCL_COMPILER(GCC, <, 8) // Old GCC considers those as deleted
```
fbusato, line 121: 😮‍💨

libcudacxx/test/libcudacxx/std/numerics/complex.number/complex/traits.pass.cpp
```diff
@@ -0,0 +1,71 @@
...
+  static_assert(cuda::std::is_trivially_move_assignable_v<C> == cuda::std::is_floating_point_v<T>);
+  static_assert(cuda::std::is_nothrow_move_assignable_v<C>);
+
+  static_assert(cuda::std::is_trivially_destructible_v<C>);
+  static_assert(cuda::std::is_trivially_copyable_v<C> == cuda::std::is_floating_point_v<T>);
+}
+
+int main(int, char**)
+{
+  test<float>();
```
fbusato, line 56: question. what about `__float128`?

libcudacxx/include/cuda/std/__complex/complex.h
fbusato: I don't have a strong opinion on that, but `noexcept` is what we expect by this type even if it is not strictly required

fbusato (approved):

libcudacxx/test/libcudacxx/std/numerics/complex.number/complex/traits.pass.cpp
miscco: support for `__float128` is sketchy and we rarely test it outside of explicit floating point facilities

PR #8357, opened 2026-04-10
Don't guard [[assume]] behind C++23

Jacobfaib:
Only quite old or technically unsupported clangs warn for it.

## Checklist


miscco (approved):

davebayer (approved):

bernhardmgruber (approved):

fbusato:
@Jacobfaib  I don't think it is correct. `[[assume]]` is not really supported in device code < C++23 with nvcc. It is ignored, which can significantly affect performance. Please revert or improve the preprocessing

Jacobfaib:
There is already additional handling for `nvcc` in the lines immediately preceding the diff. I added the `&& (_CCCL_STD_VER >= 2023)` in #8283 but folks wanted the old behavior back (having `[[assume]]` before C++23). So I reverted it here.

PR #8359, opened 2026-04-10
Do not use open coded `__host__ __device__` in tests

miscco:
We might want to use different attributes when tile comes around


davebayer (approved):
LGTM, but I'd like more opinions on this

libcudacxx/test/libcudacxx/cuda/annotated_ptr/access_property.pass.cpp
```diff
@@ -11,7 +11,7 @@
 #include "utils.h"
 
 template <typename T>
-__host__ __device__ __noinline__ void test_global_implicit_property(T ap, cudaAccessProperty cp)
+_CCCL_API __noinline__ void test_global_implicit_property(T ap, cudaAccessProperty cp)
```
jrhemstad: question: `_CCCL_API` doesn't feel very descriptive. Is there another macro name we could use that is more self-describing?

miscco: Its the one we use internally. 

The issue is that we would otherwise make sure every file includes the "test_macros.h" file

fbusato: I would prefer an alternative solution if possible
`_CCCL_API` has additional attributes. My concern is that this could even change the behavior in edge cases.
```cpp
#  define _CCCL_API        _CCCL_HOST_DEVICE _CCCL_VISIBILITY_HIDDEN _CCCL_EXCLUDE_FROM_EXPLICIT_INSTANTIATION
```

wmaxey: I'd prefer if our tests implement the macros separately from CCCL. This provides some coverage in case there are differences.

miscco: yeah it might just be a tad more work because not all tests include `test_macros.h` but we can do that

miscco: Moved to `TEST_API` and `TEST_DEVICE_API`

libcudacxx/test/support/test_macros.h
```diff
@@ -24,6 +24,10 @@
 #define TEST_DIAG_SUPPRESS_MSVC(...)  _CCCL_DIAG_SUPPRESS_MSVC(__VA_ARGS__)
 #define TEST_NV_DIAG_SUPPRESS(...)    _CCCL_BEGIN_NV_DIAG_SUPPRESS(__VA_ARGS__)
 
+// Use the CCCL host device function
+#define TEST_API        _CCCL_HOST_DEVICE
+#define TEST_DEVICE_API _CCCL_DEVICE
```
bernhardmgruber: Suggestion: I would prefer to have those macros named either `TEST_HOST_DEVICE` etc., or something that does not involve `API`, like `TEST_FUNC`.

Are we planning to add any other attributes except execution space annotations? Then I think I would prefer `TEST_[HOST_]DEVICE`, or just `TEST_HD`.

bernhardmgruber (approved):

libcudacxx/test/support/test_macros.h
miscco: It will also container Tile, so I went with TEST_FUNC

davebayer (approved):

PR #8365, opened 2026-04-10
[docs] Update cuda::maximum/minimum signatures to match actual API

### LINKED ISSUE #8350: [DOC]: `cuda::maximum` and `cuda::minimum` docs differ from API
### Is this a duplicate?

Update

The docs make it look like these functors take inputs by value, but looking at the header they are taking them by reference instead. 

https://nvidia.github.io/cccl/unstable/libcudacxx/extended_api/functional/maximum_minimum.html#libcudacxx-extended-api-functional-maximum-minimum

edenfunf:
## Description

closes #8350

The documentation for `cuda::maximum` and `cuda::minimum` did not match the actual API in several ways:

| | Docs (before) | Actual API |
|---|---|---|
| Parameters | by value (`T a, T b`) | by const reference (`const T& a, const T& b`) |
| Qualifier | `inline` | `constexpr` |
| Exception spec | *(not shown)* | `noexcept(/* conditional */)` |
| FP behavior | *(not documented)* | Uses `fmax`/`fmin` (NaN = missing data) |

**Changes:**

- Updated all four `operator()` signatures (typed + `void` specializations, for both `maximum` and `minimum`) to use `const T&` params, `constexpr`, and `noexcept(/* see below */)`
- Added "Defined in the header" line (consistent with other docs in this directory)
- Added "Floating-Point Behavior" section explaining that `fmax`/`fmin` is used for floating-point types, as requested by @fbusato in the issue comments

## Checklist

### DIFF
--- a/docs/libcudacxx/extended_api/functional/maximum_minimum.rst
+++ b/docs/libcudacxx/extended_api/functional/maximum_minimum.rst
@@ -3,39 +3,46 @@
 ``cuda::maximum`` and ``cuda::minimum``
 =======================================
 
+Defined in the header ``<cuda/functional>``.
+
 .. code:: cuda
 
     template <typename T>
     struct maximum {
-        [[nodiscard]] __host__ __device__ inline
-        T operator()(T a, T b) const;
+        [[nodiscard]] __host__ __device__ constexpr
+        T operator()(const T& a, const T& b) const noexcept(/* see below */);
     };
 
     template <>
     struct maximum<void> {
         template <typename T1, typename T2>
-        [[nodiscard]] __host__ __device__ inline
-        cuda::std::common_type_t<T1, T2> operator()(T1 a, T2 b) const;
+        [[nodiscard]] __host__ __device__ constexpr
+        cuda::std::common_type_t<T1, T2> operator()(const T1& a, const T2& b) const noexcept(/* see below */);
     };
 
     template <typename T>
     struct minimum {
-        [[nodiscard]] __host__ __device__ inline
-        T operator()(T a, T b) const;
+        [[nodiscard]] __host__ __device__ constexpr
+        T operator()(const T& a, const T& b) const noexcept(/* see below */);
     };
 
     template <>
     struct minimum<void> {
         template <typename T1, typename T2>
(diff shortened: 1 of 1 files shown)


docs/libcudacxx/extended_api/functional/maximum_minimum.rst
```diff
@@ -3,40 +3,47 @@
...
+Function objects for performing maximum and minimum operations. The ``operator()`` is ``noexcept`` when the comparison between the values is also ``noexcept``.
 
 .. note::
 
     Differently from ``std::plus`` and other functional operators, ``cuda::maximum`` and ``cuda::minimum`` specialized for ``void`` returns ``cuda::std::common_type_t`` and not the implicit promotion
 
+Floating-Point Behavior
+-----------------------
+
+For floating-point types (and extended floating-point types), ``cuda::maximum`` uses ``cuda::std::fmax`` and ``cuda::minimum`` uses ``cuda::std::fmin`` instead of the comparison operator. This means that ``NaN`` arguments are treated as missing data: ``cuda::maximum{}(NaN, 1.0)`` returns ``1.0``, and ``cuda::minimum{}(NaN, 1.0)`` returns ``1.0``.
```
fbusato: please also add a note to explain that the two operators are also commutative with floating-point types.

fbusato: > This means that ``NaN`` arguments are treated as missing data

This is true in CUDA because we don't have signaling NaN, but potentially not right for other implementations (e.g. non-IEEE754). I would just write that it follows `std::fmax/fmin` specification.

fbusato (changes requested):

edenfunf:
@fbusato  Thanks for the review! I've pushed a follow-up commit addressing both points:

- Replaced the NaN-specific claim with a reference to the `std::fmax`/`std::fmin` specification
- Added a note about commutativity for floating-point types

Please take another look when you get a chance.

fbusato (approved):
looks good.

fbusato:
thanks for the contribution, @edenfunf !!

PR #8366, opened 2026-04-10
Allow specifying a difference type for cuda::counting_iterator

Jacobfaib:
## Description

Allow an explicit difference type to be set for `counting_iterator`

## Checklist

### DIFF
--- a/libcudacxx/include/cuda/__fwd/iterator.h
+++ b/libcudacxx/include/cuda/__fwd/iterator.h
@@ -22,10 +22,13 @@
 #endif // no system header
 
 #include <cuda/__fwd/random.h>
+#include <cuda/std/__concepts/arithmetic.h>
 #include <cuda/std/__concepts/copyable.h>
+#include <cuda/std/__cstddef/types.h>
 #include <cuda/std/__iterator/concepts.h>
 #include <cuda/std/__iterator/incrementable_traits.h>
 #include <cuda/std/__type_traits/enable_if.h>
+#include <cuda/std/__type_traits/type_identity.h>
 #include <cuda/std/cstdint>
 
 #include <cuda/std/__cccl/prologue.h>
@@ -35,13 +38,49 @@ _CCCL_BEGIN_NAMESPACE_CUDA
 template <class _Tp, class _Index = ::cuda::std::ptrdiff_t>
 class constant_iterator;
 
+template <class _Tp>
+[[nodiscard]] _CCCL_API _CCCL_CONSTEVAL auto __get_wider_signed() noexcept
+{
+  if constexpr (sizeof(_Tp) < sizeof(int))
+  {
+    return ::cuda::std::type_identity<int>{};
+  }
+  else if constexpr (sizeof(_Tp) < sizeof(long))
+  {
+    return ::cuda::std::type_identity<long>{};
+  }
+#if _CCCL_HAS_INT128()
+  else if constexpr (sizeof(_Tp) < sizeof(long long))
+  {
+    return ::cuda::std::type_identity<long long>{};
+  }
+  else // if constexpr (sizeof(_Start) < sizeof(__int128_t))
+  {
+    return ::cuda::std::type_identity<__int128_t>{};
+  }
+#else // ^^^ _CCCL_HAS_INT128() ^^^ / vvv !_CCCL_HAS_INT128() vvv
+  else // if constexpr (sizeof(_Start) < sizeof(long long))
+  {
+    return ::cuda::std::type_identity<long long>{};
+  }
+#endif // _CCCL_HAS_INT128()
+}
+
+template <class _Start>
+using _IotaDiffT = typename ::cuda::std::conditional_t<
+  (!::cuda::std::integral<_Start> || sizeof(::cuda::std::iter_difference_t<_Start>) > sizeof(_Start)),
+  ::cuda::std::type_identity<::cuda::std::iter_difference_t<_Start>>,
+  decltype(::cuda::__get_wider_signed<_Start>())>::type;
+
 #if _CCCL_HAS_CONCEPTS()
-template <::cuda::std::weakly_incrementable _Start>
+template <::cuda::std::weakly_incrementable _Start, ::cuda::std::signed_integral _DiffT = _IotaDiffT<_Start>>
   requires ::cuda::std::copyable<_Start>
 #else // ^^^ _CCCL_HAS_CONCEPTS() ^^^ / vvv !_CCCL_HAS_CONCEPTS() vvv
 template <class _Start,
+          class _DiffT                                                             = _IotaDiffT<_Start>,
           ::cuda::std::enable_if_t<::cuda::std::weakly_incrementable<_Start>, int> = 0,
-          ::cuda::std::enable_if_t<::cuda::std::copyable<_Start>, int>             = 0>
+          ::cuda::std::enable_if_t<::cuda::std::copyable<_Start>, int>             = 0,
+          ::cuda::std::enable_if_t<::cuda::std::signed_integral<_DiffT>, int>      = 0>
 #endif // ^^^ !_CCCL_HAS_CONCEPTS() ^^^
 class counting_iterator;
 
--- a/libcudacxx/include/cuda/__iterator/counting_iterator.h
+++ b/libcudacxx/include/cuda/__iterator/counting_iterator.h
@@ -47,6 +47,7 @@
 #include <cuda/std/__type_traits/always_false.h>
 #include <cuda/std/__type_traits/conditional.h>
 #include <cuda/std/__type_traits/enable_if.h>
+#include <cuda/std/__type_traits/is_comparable.h>
 #include <cuda/std/__type_traits/is_constructible.h>
 #include <cuda/std/__type_traits/is_nothrow_copy_constructible.h>
 #include <cuda/std/__type_traits/is_nothrow_move_constructible.h>
@@ -65,49 +66,6 @@ _CCCL_BEGIN_NAMESPACE_CUDA
 //! @{
 
 //! @cond
-
-template <class _Int>
-struct __get_wider_signed
-{
-  _CCCL_API inline static auto __call() noexcept
-  {
-    if constexpr (sizeof(_Int) < sizeof(short))
-    {
-      return ::cuda::std::type_identity<short>{};
-    }
-    else if constexpr (sizeof(_Int) < sizeof(int))
-    {
-      return ::cuda::std::type_identity<int>{};
-    }
-    else if constexpr (sizeof(_Int) < sizeof(long))
-    {
-      return ::cuda::std::type_identity<long>{};
-    }
-#if _CCCL_HAS_INT128()
-    else if constexpr (sizeof(_Int) < sizeof(long long))
-    {
-      return ::cuda::std::type_identity<long long>{};
-    }
-    else // if constexpr (sizeof(_Int) < sizeof(__int128_t))
-    {
-      return ::cuda::std::type_identity<__int128_t>{};
-    }
-#else // ^^^ _CCCL_HAS_INT128() ^^^ / vvv !_CCCL_HAS_INT128() vvv
-    else // if constexpr (sizeof(_Int) < sizeof(long long))
-    {
-      return ::cuda::std::type_identity<long long>{};
-    }
-#endif // _CCCL_HAS_INT128()
-  }
-
-  using type = typename decltype(__call())::type;
-};
-
-template <class _Start>
-using _IotaDiffT = typename ::cuda::std::conditional_t<
-  (!::cuda::std::integral<_Start> || sizeof(::cuda::std::iter_difference_t<_Start>) > sizeof(_Start)),
-  ::cuda::std::type_identity<::cuda::std::iter_difference_t<_Start>>,
-  __get_wider_signed<_Start>>::type;
 
 template <class _Iter>
 _CCCL_CONCEPT __decrementable = _CCCL_REQUIRES_EXPR((_Iter), _Iter __iter)(
@@ -164,12 +122,14 @@ struct __counting_iterator_category<_Tp, ::cuda::std::enable_if_t<::cuda::std::i
 //! std::copy(iter, iter + vec.size(), vec.begin());
 //! @endcode
 #if _CCCL_HAS_CONCEPTS()
-template <::cuda::std::weakly_incrementable _Start>
+template <::cuda::std::weakly_incrementable _Start, ::cuda::std::signed_integral _DiffT>
   requires ::cuda::std::copyable<_Start>
 #else // ^^^ _CCCL_HAS_CONCEPTS() ^^^ / vvv !_CCCL_HAS_CONCEPTS() vvv
 template <class _Start,
+          class _DiffT,
           ::cuda::std::enable_if_t<::cuda::std::weakly_incrementable<_Start>, int>,
-          ::cuda::std::enable_if_t<::cuda::std::copyable<_Start>, int>>
+          ::cuda::std::enable_if_t<::cuda::std::copyable<_Start>, int>,
+          ::cuda::std::enable_if_t<::cuda::std::signed_integral<_DiffT>, int>>
 #endif // ^^^ !_CCCL_HAS_CONCEPTS() ^^^
 class counting_iterator : public __counting_iterator_category<_Start>
 {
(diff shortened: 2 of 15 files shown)


libcudacxx/include/cuda/__iterator/counting_iterator.h
```diff
@@ -394,81 +407,78 @@ class counting_iterator : public __counting_iterator_category<_Start>
...
   {
-    return __x.__value_ <=> __y.__value_;
+    return __x.__value_ <=> __y.__get_value();
   }
 #else // ^^^ _LIBCUDACXX_HAS_SPACESHIP_OPERATOR() ^^^ / vvv !_LIBCUDACXX_HAS_SPACESHIP_OPERATOR() vvv
   //! @brief Compares two @c counting_iterator for less than.
   //! @return True if stored values compare less than
-  _CCCL_TEMPLATE(class _Start2 = _Start)
-  _CCCL_REQUIRES(::cuda::std::totally_ordered<_Start2>)
+  _CCCL_TEMPLATE(class _Start2, class _DiffT2)
```
miscco, line 448: The `_Start2` cludge is there because this would otherwise not be a depdentent type

If we have `_DiffT2` as a template argument it is unnecessary

libcudacxx/include/cuda/__iterator/counting_iterator.h
```diff
@@ -494,36 +504,36 @@ _CCCL_BEGIN_NAMESPACE_STD
...
 }
 
-template <class _Start>
-[[nodiscard]] _CCCL_HOST_API constexpr ::cuda::counting_iterator<_Start>
-next(::cuda::counting_iterator<_Start> __iter,
-     ::cuda::_IotaDiffT<_Start> __n = 1) noexcept(::cuda::std::__integer_like<_Start>)
+template <class _Start, class _DiffT>
+[[nodiscard]] _CCCL_HOST_API constexpr ::cuda::counting_iterator<_Start, _DiffT>
+next(::cuda::counting_iterator<_Start, _DiffT> __iter,
+     ::cuda::std::type_identity_t<_DiffT> __n = 1) noexcept(::cuda::std::__integer_like<_Start>)
```
miscco, line 529: Is there a reason we use `::cuda::std::type_identity_t<_DiffT>` here? There should be no ambigouity we need to break

libcudacxx/test/libcudacxx/cuda/iterators/counting_iterator/compare.pass.cpp
miscco, line 1: We also need to change the member types and iterator traits tests to check that all the aliases etc are valid

libcudacxx/include/cuda/__iterator/counting_iterator.h
Jacobfaib: Yeah, user could pass
```c++
next(counting_iterator<int, ptrdiff_t>..., int{1})
```
and it should work. Note that standard technically says that the second argument should be `iterator_traits<It>::difference_type` so we prefetch that by type-identity-ing to `DiffT` right away.

libcudacxx/test/libcudacxx/cuda/iterators/counting_iterator/compare.pass.cpp
Jacobfaib: Yeah, working on that now

libcudacxx/include/cuda/__fwd/iterator.h
```diff
@@ -35,13 +38,46 @@ _CCCL_BEGIN_NAMESPACE_CUDA
...
+  }
+#endif // _CCCL_HAS_INT128()
+}
+
+template <class _IntT>
+using _IotaDiffT = typename decltype(::cuda::__get_wider_signed<_IntT>())::type;
+
 #if _CCCL_HAS_CONCEPTS()
-template <::cuda::std::weakly_incrementable _Start>
+template <::cuda::std::weakly_incrementable _Start, ::cuda::std::signed_integral _DiffT = _IotaDiffT<_Start>>
```
Jacobfaib, line 76: @miscco I considered relaxing this to just `cuda::std::integral` to allow an unsigned difference type, but this would IMO make all of the operator+/- etc SFINAE/concept wrangling even more egregious. Do you agree?

libcudacxx/include/cuda/__iterator/counting_iterator.h
Jacobfaib: Yeah but I kept it in because it may be useful to compare counting iterators that have different value-types (since we already allow comparisons across different difference types). For example:
```c++
counting_iterator<int32_t> it1;
counting_iterator<int64_t, short> it2;

it1 < it2
```
What do you think?

libcudacxx/include/cuda/__fwd/iterator.h
```diff
@@ -35,13 +38,46 @@ _CCCL_BEGIN_NAMESPACE_CUDA
...
 class constant_iterator;
 
+template <class _Tp>
+[[nodiscard]] _CCCL_API _CCCL_CONSTEVAL auto __get_wider_signed() noexcept
+{
+  if constexpr (sizeof(_Tp) < sizeof(int))
+  {
+    return ::cuda::std::type_identity<int>{};
+  }
+  else if constexpr (sizeof(_Tp) < sizeof(long))
```
Jacobfaib, line 48: @miscco also why not use the sized integer types here? `int32_t`, `int64_t`, etc.

miscco: its ported from libc++ but if I had to guess its windows ^^

libcudacxx/include/cuda/__fwd/iterator.h
miscco: No that does not work. All iterator concepts require signed difference type

cub/benchmarks/bench/segmented_topk/keys.cu
```diff
@@ -52,8 +52,8 @@ void fixed_seg_size_topk_keys(
   constexpr auto max_num_total_items = ::cuda::std::numeric_limits<::cuda::std::int32_t>::max();
 
   // Iterator types
-  using key_input_it_t  = cuda::strided_iterator<cuda::counting_iterator<KeyT*>>;
-  using key_output_it_t = cuda::strided_iterator<cuda::counting_iterator<KeyT*>>;
+  using key_input_it_t  = cuda::strided_iterator<cuda::counting_iterator<KeyT*>, cuda::std::ptrdiff_t>;
+  using key_output_it_t = cuda::strided_iterator<cuda::counting_iterator<KeyT*>, cuda::std::ptrdiff_t>;
```
miscco: I dont think this should be necessary or are you trying to measure a potential performance benefit?

libcudacxx/include/cuda/__iterator/counting_iterator.h
```diff
@@ -211,6 +187,43 @@ class counting_iterator : public __counting_iterator_category<_Start>
...
+  _CCCL_TEMPLATE(class _DiffT2)
+  _CCCL_REQUIRES((!::cuda::std::same_as<_DiffT, _DiffT2>) )
+  _CCCL_API constexpr counting_iterator& operator=(const counting_iterator<_Start, _DiffT2>&) = delete;
+
+  //! @brief Assignment between counting iterators of differing difference types is explicitly
+  //! deleted. If such a conversion is intended, use the copy or move constructors
+  _CCCL_TEMPLATE(class _DiffT2)
+  _CCCL_REQUIRES((!::cuda::std::same_as<_DiffT, _DiffT2>) )
+  _CCCL_API constexpr counting_iterator& operator=(counting_iterator<_Start, _DiffT2>&&) = delete;
+
```
miscco, line 227: I believe that is an unnecessary burden. There should be no difference between constructing and assigning.

In both cases we should use `overflow_cast` to see whether there was an overflow when converting

libcudacxx/include/cuda/__iterator/counting_iterator.h
```diff
@@ -394,81 +407,78 @@ class counting_iterator : public __counting_iterator_category<_Start>
...
-  //! @return True if the stored values compare equal
-  _CCCL_TEMPLATE(class _Start2 = _Start)
-  _CCCL_REQUIRES(::cuda::std::equality_comparable<_Start2>)
+  _CCCL_TEMPLATE(class _Start2, class _DiffT2)
+  _CCCL_REQUIRES(::cuda::std::equality_comparable_with<_Start, _Start2>)
   [[nodiscard]] _CCCL_API friend constexpr bool
-  operator==(const counting_iterator& __x, const counting_iterator& __y) noexcept(
-    noexcept(::cuda::std::declval<const _Start2&>() == ::cuda::std::declval<const _Start2&>()))
+  operator==(const counting_iterator& __x, const counting_iterator<_Start2, _DiffT2>& __y) noexcept(
+    noexcept(::cuda::std::declval<const _Start&>() == ::cuda::std::declval<const _Start2&>()))
```
miscco: We have `__is_cpp17_nothrow_equality_comparable_v` in `is_comparable.h` Which checks exactly that but is much better readable

libcudacxx/include/cuda/__iterator/counting_iterator.h
```diff
@@ -394,81 +407,78 @@ class counting_iterator : public __counting_iterator_category<_Start>
...
   //! @return True if the stored values do not compare equal
-  _CCCL_TEMPLATE(class _Start2 = _Start)
-  _CCCL_REQUIRES(::cuda::std::equality_comparable<_Start2>)
+  _CCCL_TEMPLATE(class _Start2, class _DiffT2)
+  _CCCL_REQUIRES(::cuda::std::equality_comparable_with<_Start, _Start2>)
   [[nodiscard]] _CCCL_API friend constexpr bool
-  operator!=(const counting_iterator& __x, const counting_iterator& __y) noexcept(
-    noexcept(::cuda::std::declval<const _Start2&>() != ::cuda::std::declval<const _Start2&>()))
+  operator!=(const counting_iterator& __x, const counting_iterator<_Start2, _DiffT2>& __y) noexcept(
+    noexcept(::cuda::std::declval<const _Start&>() != ::cuda::std::declval<const _Start2&>()))
```
miscco, line 427: I am strongly considering just changing this to `!operator==`

libcudacxx/test/libcudacxx/cuda/iterators/counting_iterator/minus.pass.cpp
```diff
@@ -28,6 +28,13 @@ using IntDiffT = long long;
 using IntDiffT = long;
 #endif
 
+_CCCL_DIAG_PUSH
+// MSVC complains that converting a difference type (ptrdiff_t) to
+// SomeInt::diifference_type (int) for operator-() is not allowed. This would require
```
miscco:
```suggestion
// SomeInt::difference_type (int) for operator-() is not allowed. This would require
```

libcudacxx/test/libcudacxx/cuda/iterators/counting_iterator/minus.pass.cpp
```diff
@@ -28,6 +28,13 @@ using IntDiffT = long long;
...
+// changing SomeInt::difference_type, but that is used in other tests to ensure that the
+// counting_iterator's difference type is chosen properly.
+_CCCL_DIAG_SUPPRESS_MSVC(4244)
```
miscco: This is a valid warning, we need to make sure that we never trigger that warning by appropriately casting around

libcudacxx/test/libcudacxx/cuda/iterators/counting_iterator/minus_eq.pass.cpp
```diff
@@ -13,10 +13,18 @@
...
 
 #include "test_macros.h"
 #include "types.h"
 
+_CCCL_DIAG_PUSH
+// MSVC complains that converting a difference type (ptrdiff_t) to
+// SomeInt::diifference_type (int) for operator-=() is not allowed. This would require
+// changing SomeInt::difference_type, but that is used in other tests to ensure that the
+// counting_iterator's difference type is chosen properly.
+_CCCL_DIAG_SUPPRESS_MSVC(4244)
```
miscco: Ditto, need to cast inside the method

libcudacxx/test/libcudacxx/std/ranges/range.factories/range.iota.view/types.h
```diff
@@ -67,11 +67,21 @@ struct SomeInt
     lhs.value_ += rhs;
     return lhs;
   }
+  __host__ __device__ friend constexpr SomeInt& operator+=(SomeInt& lhs, cuda::std::ptrdiff_t rhs)
+  {
+    lhs.value_ += static_cast<int>(rhs);
+    return lhs;
+  }
```
miscco: Lets not change those, we need to make sure it works as specified and iota_view always has the right difference type, so it must work

libcudacxx/test/libcudacxx/std/ranges/range.factories/range.iota.view/iterator/member_typedefs.compile.pass.cpp
```diff
@@ -201,7 +201,7 @@ __host__ __device__ void test()
     static_assert(cuda::std::same_as<Iter::iterator_concept, cuda::std::input_iterator_tag>);
     static_assert(!HasIteratorCategory<cuda::std::ranges::iota_view<NotIncrementable>>);
     static_assert(cuda::std::same_as<Iter::value_type, NotIncrementable>);
-    static_assert(cuda::std::same_as<Iter::difference_type, int>);
+    static_assert(cuda::std::same_as<Iter::difference_type, cuda::std::ptrdiff_t>);
```
miscco: This should not have changed

miscco (changes requested):

cub/benchmarks/bench/segmented_topk/keys.cu
Jacobfaib: Yeah I don't really understand why it was necessary either, but I was getting compile-errors down the line saying that `strided_iterator<counting_iterator<...>, ptrdiff_t>` was not convertible to `strided_iterator<counting_iterator<...>, __int128>`. I think this was because of the changes you had in `_IotaDiffT` that made it so that it always chose what as in `__get_wider_signed()` instead of considering `iter_difference_t`. I'll revert that. 

All of the MSVC related warnings are as a result of this change as well.

libcudacxx/include/cuda/__iterator/counting_iterator.h
Jacobfaib: The point of deleting these was to force the user to make the conversion explicit, because converting to another difference type is bugprone. So the user should not be able to e.g.
```c++
counting_iterator<float *, ptrdiff_t> x;
counting_iterator<float *, int8_t> y = x;

y += 127; // oops
```
You can't make assignment operators `explicit`, so they should have to explicitly opt into this via explicit construction:
```c++
counting_iterator<float *, ptrdiff_t> x;
counting_iterator<float *, int8_t> y = counting_iterator<float *, int8_t>{x};

y += 127; // well, you asked for it
```

libcudacxx/test/libcudacxx/cuda/iterators/counting_iterator/minus.pass.cpp
Jacobfaib: The cast would need to appear inside `counting_iterator::operator-()`, specifically:
```c++
if constexpr (::cuda::std::__integer_like<_Start> && !::cuda::std::__signed_integer_like<_Start>)
    {
      if (__n >= difference_type(0))
      {
        __value_ -= static_cast<_Start>(__n);
      }
      else
      {
        __value_ += static_cast<_Start>(-__n);
      }
    }
    else if constexpr (::cuda::std::__signed_integer_like<_Start>)
    {
      __value_ -= static_cast<_Start>(__n);
    }
    else
    {
      __value_ -= __n; // << ---- HERE
    }
```
I don't think that is a correct change because now you no longer test that the iterator operators properly exercise their `requires` clauses.

`SomeInt` defined the types of its arithmetic operations as taking specifically `int` instead of a more appropriate wider difference type. I consider this a "user bug" in the definition of `SomeInt`. The code still works, but MSVC (and `clang-tidy` FWIW) warn about integer truncation. We should not cast this away as it would hide that warning from the user.

libcudacxx/test/libcudacxx/std/ranges/range.factories/range.iota.view/iterator/member_typedefs.compile.pass.cpp
Jacobfaib: Was as a result of the `IotaDiffT` change that has now been reverted

libcudacxx/include/cuda/__iterator/counting_iterator.h
miscco: Those and the other comparisons also fall back to operator< so we can use the same shortcut as above

Jacobfaib: I'm not sure what you mean, it already calls `__y < __x`

miscco: I meant the `__is_cpp17_equality_comparable_v`

Jacobfaib: Switched all to using the trait

miscco (approved):

PR #3530, opened 2025-01-27
Adds support for large num items to `DeviceMerge`

elstehle:
## Description

Closes https://github.com/NVIDIA/cccl/issues/3134

Decided to go with single offset type template instantiation:
- Performance for merging pairs with *wider offset type* is, on average, even *better* than for 32 bit wide type
- Choosing the offset type statically is challenging, since we take the size of the LHS and the size of the RHS, even if both are `int`, the output size may require 64-bit indexing
- We need to compile only a single kernel independent of the user-provided offset type

### Performance summary for merge.pairs on H100:
 
|    | i64/i32 time | i64/i32 time (2^28) |
|-----|--------------|---------------------|
| min |       74.39% |              74.39% |
| max |      105.08% |             105.08% |
| avg |       99.98% |              99.42% |

### Performance summary for merge.keys on H100:
| x   | i64/i32 time | i64/i32 time (2^28) |
|-----|--------------|---------------------|
| min |       98.98% |              99.98% |
| max |      107.28% |             107.28% |
| avg |      100.91% |             101.30% |

PR #8381, opened 2026-04-10
Replace `detail::merge::dispatch` by CUB's public API

### LINKED ISSUE #7955: Use the new tuning API internally for `detail::merge::dispatch_t`

bernhardmgruber:
* Use the new tuning API internally where possible
* Drop 32 bit offset types in benchmarks (public API just uses int64)
* Drop index type dispatch in Thrust (public API just uses int64)

This PR does change the SASS of `cub.bench.merge.pairs.base` (tested on SM120), but this is expected since we discard all 32-bit offset type benchmarks. Those are not needed, since the public API does not allow to compile a 32-bit kernel.

This PR will also take away Thrust's ability to compile a 32-bit offset type kernel, which is not needed since the 64-bit offset type version is on par: #3530

Fixes: #7955

### DIFF
--- a/cub/benchmarks/bench/merge/keys.cu
+++ b/cub/benchmarks/bench/merge/keys.cu
@@ -49,7 +49,7 @@ void keys(nvbench::state& state, nvbench::type_list<KeyT>)
       launch
 #if !TUNE_BASE
       ,
-      cuda::execution::__tune(policy_selector<key_t, value_t, offset_t>{})
+      cuda::execution::__tune(bench_policy_selector<key_t>{})
 #endif // !TUNE_BASE
     );
     _CCCL_TRY_CUDA_API(
--- a/cub/benchmarks/bench/merge/keys.cu
+++ b/cub/benchmarks/bench/merge/keys.cu
@@ -18,24 +18,10 @@
 // %RANGE% TUNE_ITEMS_PER_THREAD ipt 7:24:1
 // %RANGE% TUNE_THREADS_PER_BLOCK_POW2 tpb 6:10:1
 
-#if !TUNE_BASE
-struct bench_policy_selector
+template <typename KeyT>
+void keys(nvbench::state& state, nvbench::type_list<KeyT>)
 {
-  _CCCL_API constexpr auto operator()(::cuda::arch_id /*arch*/) const -> cub::detail::merge::merge_policy
-  {
-    return cub::detail::merge::merge_policy{
-      TUNE_THREADS_PER_BLOCK,
-      cub::Nominal4BItemsToItems<KeyT>(TUNE_ITEMS_PER_THREAD),
-      TUNE_LOAD_MODIFIER,
-      TUNE_STORE_ALGORITHM,
-      TUNE_USE_BL2SH};
-  }
-};
-#endif // !TUNE_BASE
-
-template <typename KeyT, typename OffsetT>
-void keys(nvbench::state& state, nvbench::type_list<KeyT, OffsetT>)
-{
+  using offset_t     = int64_t;
   using compare_op_t = less_t;
 
   // Retrieve axis parameters
@@ -46,7 +32,7 @@ void keys(nvbench::state& state, nvbench::type_list<KeyT, OffsetT>)
   const auto num_items_rhs  = elements - num_items_lhs;
   auto [keys_lhs, keys_rhs] = generate_lhs_rhs<KeyT>(num_items_lhs, num_items_rhs, entropy);
 
-  thrust::device_vector<KeyT> keys_out(elements);
+  thrust::device_vector<KeyT> keys_out(elements, thrust::no_init);
   KeyT* d_keys_lhs = thrust::raw_pointer_cast(keys_lhs.data());
   KeyT* d_keys_rhs = thrust::raw_pointer_cast(keys_rhs.data());
   KeyT* d_keys_out = thrust::raw_pointer_cast(keys_out.data());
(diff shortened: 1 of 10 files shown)


cub/cub/detail/env_dispatch.cuh
```diff
@@ -74,6 +74,18 @@ CUB_RUNTIME_FUNCTION static cudaError_t dispatch_with_env(EnvT env, AlgorithmCal
   return (error != cudaSuccess) ? error : deallocate_error;
 }
 //! @endcond
+
+template <typename DefaultPolicySelector, typename EnvT, typename AlgorithmCallable>
+CUB_RUNTIME_FUNCTION static cudaError_t dispatch_with_env_and_tuning(EnvT env, AlgorithmCallable&& algorithm_callable)
```
bernhardmgruber, line 79: I don't have a good name for this function yet, but I think it will at some point replace `dispatch_with_env` entirely. But this rollout will take time.

cub/test/catch2_test_device_reduce_env.cu
```diff
@@ -47,18 +47,6 @@ struct block_size_check_t
...
-  int* ptr;
-
-  template <class ActivePolicyT>
-  cudaError_t Invoke()
-  {
-    *ptr = ActivePolicyT::SingleTilePolicy::BLOCK_THREADS;
-    return cudaSuccess;
-  }
-};
-
```
bernhardmgruber, line 61: Drive-by fix. This was unused.

nvbench_helper/nvbench_helper/nvbench_helper.cuh
```diff
@@ -18,6 +18,10 @@
 #  include <cuda/stream>
 #endif // THRUST_DEVICE_SYSTEM == THRUST_DEVICE_SYSTEM_CUDA
 
+#include <cuda/__memory_resource/any_resource.h>
+#include <cuda/__stream/stream_ref.h>
+#include <cuda/std/__execution/env.h>
```
miscco: Critical: Those headers are already included above and should not be included without a CUDA system, or at least CTK headers

nvbench_helper/nvbench_helper/nvbench_helper.cuh
```diff
@@ -711,4 +711,14 @@ auto policy(caching_allocator_t&, nvbench::launch&)
...
+
+// Returns an environment for benchmarking using alloc as MR, launch's stream, and any additional envs passed in.
+template <typename... MoreEnvs>
+auto cub_bench_env(caching_allocator_t& alloc, nvbench::launch& launch, MoreEnvs... envs)
+{
+  return cuda::std::execution::env{
+    ::cuda::stream_ref{launch.get_stream().get_stream()},
+    ::cuda::std::execution::prop{cuda::mr::get_memory_resource, ::cuda::mr::resource_ref<>{alloc}},
+    envs...};
+}
```
NaderAlAwar, line 724: Important: this will not compile when `THRUST_DEVICE_SYSTEM=CPP` because the relevant header files will not be included

cub/test/catch2_test_device_merge_no_unroll.cu
```diff
@@ -88,37 +88,6 @@ C2H_TEST("DeviceMerge::MergeKeys large key types", "[merge][device]", c2h::type_
...
-        s1,
-        k2,
-        value_nullptr,
-        s2,
-        r,
-        value_nullptr,
-        co,
-        cudaStream_t{0},
-        fixed_policy_selector{});
+      REQUIRE(cub::DeviceMerge::MergeKeys(k1, s1, k2, s2, r, co, fixed_policy_selector{}) == cudaSuccess);
```
NaderAlAwar: Critical: `fixed_policy_selector{}` needs to be passed through `cuda::execution::__tune()`, not directly as the last argument to `DeviceMerge::MergeKeys`. After this change, `fixed_policy_selector` will be treated as an env not a tuning, which breaks the intent of this test, which is to test the fallback vsmem path.

bernhardmgruber: Thx a lot for catching this!

NaderAlAwar (approved):

cub/benchmarks/bench/merge/keys.cu
```diff
@@ -56,51 +42,26 @@ void keys(nvbench::state& state, nvbench::type_list<KeyT, OffsetT>)
...
+  caching_allocator_t alloc;
+  state.exec(nvbench::exec_tag::gpu | nvbench::exec_tag::no_batch, [&](nvbench::launch& launch) {
+    auto env = cub_bench_env(
+      alloc,
+      launch
 #if !TUNE_BASE
-    ,
-    bench_policy_selector{}
+      ,
+      cuda::execution::__tune(policy_selector<key_t, value_t, offset_t>{})
```
bernhardmgruber: Important: 

```suggestion
      cuda::execution::__tune(bench_policy_selector<key_t>{})
```

cub/benchmarks/bench/merge/pairs.cu
```diff
@@ -66,49 +52,29 @@ void pairs(nvbench::state& state, nvbench::type_list<KeyT, ValueT, OffsetT>)
...
+  caching_allocator_t alloc;
+  state.exec(nvbench::exec_tag::gpu | nvbench::exec_tag::no_batch, [&](nvbench::launch& launch) {
+    auto env = cub_bench_env(
+      alloc,
+      launch
 #if !TUNE_BASE
-    ,
-    bench_policy_selector{}
+      ,
+      cuda::execution::__tune(policy_selector<key_t, value_t, offset_t>{})
```
bernhardmgruber:
```suggestion
      cuda::execution::__tune(bench_policy_selector<key_t>{})
```

miscco (approved):

PR #8386, opened 2026-04-13
Use `cuda::ptx` functions in `cuda::for_each_canceled`

davebayer:
We can simplify the implementation a bit by using already existing functions from `cuda::ptx` namespace.

### DIFF
--- a/libcudacxx/include/cuda/__functional/for_each_canceled.h
+++ b/libcudacxx/include/cuda/__functional/for_each_canceled.h
@@ -23,6 +23,7 @@
 
 #if _CCCL_CUDA_COMPILATION()
 
+#  include <cuda/__ptx/instructions/clusterlaunchcontrol.h>
 #  include <cuda/std/__functional/invoke.h>
 #  include <cuda/std/__utility/move.h>
 #  include <cuda/std/__utility/unreachable.h>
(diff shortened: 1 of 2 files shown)


libcudacxx/include/cuda/__functional/for_each_canceled.h
```diff
@@ -148,39 +96,27 @@ _CCCL_DEVICE_API void __for_each_canceled_block_sm100(::dim3 __block_idx, bool _
...
-        // required since each thread block calls this API at most once.
-        break;
-      }
+      // Invalidating mbarrier and synchronizing before exiting not
+      // required since each thread block calls this API at most once.
+      break;
     }
 
-    // Read new thread block dimensions
-    ::dim3 __b(::cuda::device::__cluster_get_dim<0>(__result), 1, 1);
```
davebayer, line 174: @gonzalobg is this really correct? Shouldn't the new block index default to `{::cuda::device::__cluster_get_dim<0>(__result), 0, 0}` since it's an index and not dimensions?

bernhardmgruber: I would intuitively agree, but then we could argue that we didn't define what value an the dimension of `__b` should have, if it's `>= __ThreadBlockRank`. In other words, if I call `for_each_canceled_block<1>` we do not specify what `dim3.y and dim3.z` are for the argument of the passed unary function.

I would have also chosen zero though instead of one. Maybe @gonzalobg can explain.

PR #8392, opened 2026-04-13
Implement `ranges::views::drop_while_view`

Jacobfaib:
## Description

See https://github.com/NVIDIA/cccl/issues/93

## Checklist

### DIFF
--- a/libcudacxx/include/cuda/std/__ranges/drop_while_view.h
+++ b/libcudacxx/include/cuda/std/__ranges/drop_while_view.h
@@ -1,0 +1,176 @@
+// -*- C++ -*-
+//===----------------------------------------------------------------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef _CUDA_STD___RANGES_DROP_WHILE_VIEW_H
+#define _CUDA_STD___RANGES_DROP_WHILE_VIEW_H
+
+#include <cuda/std/detail/__config>
+
+#if defined(_CCCL_IMPLICIT_SYSTEM_HEADER_GCC)
+#  pragma GCC system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_CLANG)
+#  pragma clang system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_MSVC)
+#  pragma system_header
+#endif // no system header
+
+#include <cuda/std/__algorithm/ranges_find_if_not.h>
+#include <cuda/std/__concepts/constructible.h>
+#include <cuda/std/__functional/bind_back.h>
+#include <cuda/std/__functional/reference_wrapper.h>
+#include <cuda/std/__iterator/concepts.h>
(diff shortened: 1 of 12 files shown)


libcudacxx/include/cuda/std/__ranges/drop_while_view.h
```diff
@@ -0,0 +1,176 @@
...
+  [[nodiscard]] _CCCL_API constexpr const _Pred& pred() const
+  {
+    return *__pred_;
+  }
+
+  [[nodiscard]] _CCCL_API constexpr auto begin()
+  {
+    _CCCL_ASSERT(__pred_.__has_value(),
+                 "drop_while_view needs to have a non-empty predicate before calling begin() -- did a previous "
+                 "assignment to this drop_while_view fail?");
```
miscco, line 112: I belkieve that error message was not the best
```suggestion
    _CCCL_ASSERT(__pred_.__has_value(),
                 "drop_while_view needs to have a valid predicate before calling begin() -- did a previous "
                 "assignment to this drop_while_view fail?");
```

libcudacxx/test/libcudacxx/std/ranges/range.adaptors/range.drop.while/base.pass.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+// constexpr V base() const & requires copy_constructible<V> { return base_; }
+// constexpr V base() && { return cuda::std::move(base_); }
+
+#include <cuda/std/cassert>
+#include <cuda/std/ranges>
+#include <cuda/std/type_traits>
+#include <cuda/std/utility>
+
+#include "MoveOnly.h"
```
miscco, line 18: Please include `"test_macros.h` in all files, I was lazy before

Jacobfaib: Added

libcudacxx/include/cuda/std/__ranges/drop_while_view.h
```diff
@@ -0,0 +1,176 @@
+// -*- C++ -*-
+//===----------------------------------------------------------------------===//
+//
+// Part of libcu++, the C++ Standard Library for your entire system,
+// under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
```
miscco, line 9: I believe this needs to be the LLVM license in all other files too

libcudacxx/test/libcudacxx/std/ranges/range.adaptors/range.drop.while/adaptor.pass.cpp
```diff
@@ -0,0 +1,181 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of libcu++, the C++ Standard Library for your entire system,
```
miscco: We need to use the LLVM license in all of those files, seems like I messed up at some point 

At the same time it might be derived enough, but lets just keep the LLVM one

libcudacxx/test/libcudacxx/std/ranges/range.adaptors/range.drop.while/pred.pass.cpp
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  return true;
+}
+
+int main(int, char**)
+{
+  test();
+#if TEST_STD_VER >= 2020
+  static_assert(test(), "");
```
miscco: Nitpick: can you run clang-tidy ?

miscco (approved):

Issue #7978, opened 2026-03-10
[cuda.compute]: Investigate presence of LDL/STL instructions in kernels in CUB and cuda.compute


NaderAlAwar:
In cuda.compute and c.parallel, we assert that the LDL and STL instructions are not present in the compiled kernel for performance reasons. We currently have to disable these checks for certain combinations of CTK versions + gpu arch + algorithm inputs.

We need to do a deeper dive into what is causing these instructions to appear and re-enable the checks. We can assume that the C++ versions of the algorithms do not contain these instructions as they have been tuned, which should automatically disqualify tunings that do result in those instructions. In order to narrow down what is causing these instructions to appear, we should look at the parts of cuda.compute compilation process that differ from the C++ compilation process:

1. Kernel template instantiation: we could be using different template arguments in some instances than in C++ due to type erasure. In some cases, things like `const` might be missing, which might not have a huge effect but it is still good to be consistent between the two.
2. Selecting the right tuning: the process for selecting the tuning differs between C++ and c.parallel (which cuda.compute depends on) as C++ relies fully on compile-time information while c.parallel finds the tuning at run-time. We need to ensure that the same policies are being selected for the same input types.
3. Different compilers: in cuda.compute, we use numba-cuda to compile the operators, nvrtc to compile the instantiated kernels, and LTO to generate the final compiled kernel. This is different from C++ where we only use NVCC. If we can find instances where the same code is being compiled differently, we should file bug reports.

Update: it appears that these instructions can appear even in CUB, so our investigation should include that as well

bernhardmgruber:
FYI: The warpspeed scan implementation may generate LDL and STL instructions due to a very high number of items per thread. I think it's 255 for `int8`. Taking the hit from local memory is cheaper than communicating more partial results through the lookback. So we cannot generally assume that an optimized kernel will not contain LDL/STL instructions.

PR #8393, opened 2026-04-13
Disable all LDL and STL sass checks by default

NaderAlAwar:
This PR removes disables all the checks for LDL and STL instructions in SASS by default. The checks can still be turned on by modifying the flags in the source. We are doing this because these checks fail somewhat randomly and typically block CI, and our default response is to disable them for specific instances.

The reason we are disabling all these checks now is that we have recently added benchmarks to cuda.compute which showed that performance was equivalent to C++ https://github.com/NVIDIA/cccl/pull/7341#issuecomment-4091403018 (except for a few cases for which we opened issues).

In the future, we plan on investigating why LDL and STL instructions appear in our kernels https://github.com/NVIDIA/cccl/issues/7978

### DIFF
--- a/c/parallel/test/algorithm_execution.h
+++ b/c/parallel/test/algorithm_execution.h
@@ -20,6 +20,8 @@
 #include <c2h/catch2_test_helper.h>
 #include <cccl/c/types.h>
 
+inline constexpr bool check_ldl_stl_in_sass = false;
+
 template <int device_id_ = 0>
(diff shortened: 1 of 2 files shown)


bernhardmgruber (approved):

shwina (approved):
Enthusiastic approval

NaderAlAwar:
pre-commit.ci autofix

Issue #8310, opened 2026-04-07
[FEA]: Allow hierarchy queries for levels that are not part of the hierarchy


davebayer:
I'd like to be able to write:
```cpp
auto v1 = cuda::gpu_thread./*query*/(cuda::warp, hierarchy);
auto v2 = cuda::warp./*query*/(cuda::block, hierarchy);
auto v3 = cuda::gpu_thread./*query*/(cuda::cluster, hierarchy);
auto v3 = cuda::cluster./*query*/(cuda::grid, hierarchy);
```
even when `warp`/`cluster` are not part of the hierarchy. This will simplify implementation of group queries.

PR #8394, opened 2026-04-13
Refactor and improve `extents_as` queries

davebayer:
This PR refactors the `extents_as` implementation and adds additional support for `warp_level` queries in a hierarchy and `cluster_level` queries in a hierarchy when `cluster_level` is not part of the hierarchy. This change allows us to use queries with hierarchy parameter inside groups implementation.

Partially implements #8310. 2 more PRs will follow with `index` and `rank` implementations.

### DIFF
--- a/libcudacxx/include/cuda/__hierarchy/queries/extents.h
+++ b/libcudacxx/include/cuda/__hierarchy/queries/extents.h
@@ -1,0 +1,359 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of libcu++, the C++ Standard Library for your entire system,
+// under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef _CUDA___HIERARCHY_QUERIES_EXTENTS_H
+#define _CUDA___HIERARCHY_QUERIES_EXTENTS_H
+
+#include <cuda/std/detail/__config>
+
+#if defined(_CCCL_IMPLICIT_SYSTEM_HEADER_GCC)
+#  pragma GCC system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_CLANG)
+#  pragma clang system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_MSVC)
+#  pragma system_header
+#endif // no system header
+
+#if _CCCL_HAS_CTK()
+
+#  include <cuda/__cmath/ceil_div.h>
+#  include <cuda/__fwd/hierarchy.h>
+#  include <cuda/__hierarchy/traits.h>
+#  include <cuda/std/__algorithm/max.h>
+#  include <cuda/std/__cstddef/types.h>
+#  include <cuda/std/__mdspan/extents.h>
+#  include <cuda/std/__utility/integer_sequence.h>
+#  include <cuda/std/array>
+
+#  include <cuda/std/__cccl/prologue.h>
+
+_CCCL_BEGIN_NAMESPACE_CUDA
+
+// helpers
+
+[[nodiscard]] _CCCL_API _CCCL_CONSTEVAL ::cuda::std::size_t
+__hierarchy_static_extents_mul_helper(::cuda::std::size_t __lhs, ::cuda::std::size_t __rhs) noexcept
+{
+  if (__lhs == ::cuda::std::dynamic_extent || __rhs == ::cuda::std::dynamic_extent)
+  {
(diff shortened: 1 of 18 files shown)


libcudacxx/test/libcudacxx/cuda/hierarchy/cluster_level/hierarchy_query_signatures.compile.pass.cpp
```diff
@@ -70,14 +73,17 @@ __device__ void test_query_as_signatures(const Level& level, const Hierarchy& hi
...
 
-  // 4. Test cuda::cluster_level::index_as(x, hier) signature.
-  static_assert(
-    cuda::std::is_same_v<cuda::hierarchy_query_result<T>, decltype(cuda::cluster_level::index_as<T>(level, hier))>);
-  static_assert(noexcept(cuda::cluster_level::index_as<T>(level, hier)));
-
-  // 5. Test cuda::cluster_level::rank_as(x, hier) signature.
-  static_assert(cuda::std::is_same_v<T, decltype(cuda::cluster_level::rank_as<T>(level, hier))>);
-  static_assert(noexcept(cuda::cluster_level::rank_as<T>(level, hier)));
+  if constexpr (Hierarchy::has_level(cuda::cluster))
```
pciolkosz, line 76: Shouldn't this be over cluster queries?

libcudacxx/include/cuda/__hierarchy/queries/extents.h
```diff
@@ -0,0 +1,348 @@
...
+  [[nodiscard]] _CCCL_API static constexpr auto __call(const _Hierarchy& __hier) noexcept
+  {
+    auto __block_exts = __extents_query<thread_level, block_level>::template __call<_Tp>(__hier);
+    using _BlockExts  = decltype(__block_exts);
+
+    if constexpr (_BlockExts::rank_dynamic() == 0)
+    {
+      constexpr auto __static_thread_count =
+        _BlockExts::static_extent(0) * _BlockExts::static_extent(1) * _BlockExts::static_extent(2);
+      constexpr auto __static_warp_count = ::cuda::ceil_div(__static_thread_count, 32);
```
pciolkosz: I would add a note here that it only works if x dim is larger than 32, maybe even assert?

libcudacxx/include/cuda/__hierarchy/queries/extents.h
```diff
@@ -0,0 +1,348 @@
...
+      {
+        return __curr_exts;
+      }
+      else
+      {
+        const auto __next_exts = __extents_query<block_level, _Level>::template __call<_Tp>(__hier);
+        return ::cuda::__hierarchy_extents_mul(__curr_exts, __next_exts);
+      }
+    }
+    else
```
pciolkosz, line 297: I think we can merge if and else if __curr_exts come from a lambda or a helper

libcudacxx/test/libcudacxx/cuda/hierarchy/warp_level/hierarchy_query_signatures.compile.pass.cpp
```diff
@@ -0,0 +1,133 @@
...
+  static_assert(cuda::std::is_same_v<cuda::std::size_t, decltype(cuda::warp_level::static_count(level, hier))>);
+  static_assert(noexcept(cuda::warp_level::static_count(level, hier)));
+
+  // 5. Test cuda::warp_level::count(x, hier) signature.
+  static_assert(cuda::std::is_same_v<cuda::std::size_t, decltype(cuda::warp_level::count(level, hier))>);
+  static_assert(noexcept(cuda::warp_level::count(level, hier)));
+
+  if constexpr (Hierarchy::has_level(cuda::cluster))
+  {
+    // 6. Test cuda::warp_level::index(x, hier) signature.
```
pciolkosz, line 48: Should there be extents test here?

libcudacxx/test/libcudacxx/cuda/hierarchy/cluster_level/hierarchy_query_signatures.compile.pass.cpp
davebayer: I extend the support to allow cluster queries even when cluster level is not part of the hierarchy, so this is correct

libcudacxx/test/libcudacxx/cuda/hierarchy/warp_level/hierarchy_query_signatures.compile.pass.cpp
davebayer: It's implemented in `3.`, `index` and `rank` queries are not yet supported with warp and cluster is problematic when it's not part of the hierarchy.

I will fix this in next 2 PRs

libcudacxx/include/cuda/__hierarchy/queries/extents.h
davebayer: It's not that simple, I'd rather keep it as is if you don't mind

miscco (approved):
Looks great from my part

pciolkosz (approved):

PR #8395, opened 2026-04-13
[CUB] Replace `Shuffle(Up|Down|Index)` with cuda::device::warp_shuffle - RadixSort only

fbusato:
## Description

Split

- https://github.com/NVIDIA/cccl/pull/8159

The change only affects RadixSort routines. This helps to isolate potential performance regressions


miscco (commented):
The code changes look good to me. However, it looks like there were no benchmark results

bernhardmgruber:
Perf results look ok on sm120, but not on the other architectures.

fbusato:
> Perf results look ok on sm120, but not on the other architectures.

It is fun that I just checked the SASS for Radix Sort on SM90, SM100, SM120 for all benchmark types (`int8_t`, `int16_t`, `int32_t`, `int64_t`, `int128_t`, `float`, `double`). No difference, they are identical.

If we are not able to get stable performance comparison even for identical SASS, we have a problem (@gevtushenko )

Coincidence: https://github.com/NVIDIA/nvbench/issues/316#issuecomment-4246363558

fbusato:
@bernhardmgruber can we merge this PR? no sass diff

ci/bench.yaml
bernhardmgruber, line 1: Please revert changes to this file.

bernhardmgruber (approved):

PR #8398, opened 2026-04-13
Switch from pyenv to uv

shwina:
## Description

Unlike `pyenv`, which downloads and builds Python from source on each job, `uv` simply downloads pre-built Python binaries. Largely this should improve our CI QOL by having fewer network timeouts etc., There should also be a small improvement in Python CI times (although nothing dramatic).

## Checklist


NaderAlAwar (approved):
Python part looks good

alliepiper (approved):

PR #8404, opened 2026-04-14
[libcu++] Make memory resources work with get_memory_resource when wrapped in environment

pciolkosz:
Right now when an API takes an environment, in order to use `get_memory_resource` on that environment it needs to be wrapped in `cuda::std::execution::prop`. It would be nice to put the resource directly into it, to make it work this PR adds a CRTP base class for resources that implements `query(get_memory_resource)` that allows it to work with the environment machinery.

### DIFF
--- a/libcudacxx/include/cuda/__memory_resource/memory_resource_base.h
+++ b/libcudacxx/include/cuda/__memory_resource/memory_resource_base.h
@@ -1,0 +1,53 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of libcu++, the C++ Standard Library for your entire system,
+// under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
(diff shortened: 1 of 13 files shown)


libcudacxx/include/cuda/__memory_resource/memory_resource_base.h
```diff
@@ -0,0 +1,53 @@
...
+_CCCL_BEGIN_NAMESPACE_CUDA_MR
+
+//! @brief CRTP base class for memory resources.
+//!
+//! When a resource inherits from `memory_resource_base<Derived>`, it provides a
+//! `query(get_memory_resource_t)` method that returns a const reference to itself.
+//! This enables the resource to be discovered inside a composed
+//! `cuda::std::execution::env` by the `get_memory_resource` customization point.
+template <class _Derived>
+struct memory_resource_base
```
miscco, line 39: I feel this is overkill. It would be much better to just add the query to the individual resources

miscco (changes requested):

miscco (approved):

PR #8405, opened 2026-04-14
Cleanup `thrust::pointer_traits` a bit more

miscco:
We can rely on the standard `cuda::std::pointer_traits` for `element_type` and `difference_type`


davebayer (approved):

bernhardmgruber (approved):

bernhardmgruber:
@miscco I really like that you take some time to clean up the guts of Thrust here! It will make the implementation more robust and grounded on well-defined semantics.

PR #8406, opened 2026-04-14
Refactor and improve `index_as` queries

davebayer:
Follow up to #8394.

### DIFF
--- a/libcudacxx/include/cuda/__hierarchy/queries/index.h
+++ b/libcudacxx/include/cuda/__hierarchy/queries/index.h
@@ -1,0 +1,266 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of libcu++, the C++ Standard Library for your entire system,
+// under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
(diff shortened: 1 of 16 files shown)


libcudacxx/include/cuda/__hierarchy/queries/index.h
```diff
@@ -0,0 +1,266 @@
...
+  }
+};
+
+template <>
+struct __index_query_native<thread_level, warp_level>
+{
+  template <class _Tp>
+  [[nodiscard]] _CCCL_DEVICE_API static hierarchy_query_result<_Tp> __call() noexcept
+  {
+    return {static_cast<_Tp>(::cuda::ptx::get_sreg_laneid()), 0, 0};
```
pciolkosz, line 83: I would add a todo to check if its worth using PTX here, it might make some compiler optimizations difficult

pciolkosz (approved):

libcudacxx/include/cuda/__hierarchy/queries/index.h
davebayer: I will add comment in next PR, thanks

PR #8413, opened 2026-04-14
[docs] Fix incorrect code snippets in memory resource documentation

### LINKED ISSUE #6843: [DOC]: Misleading snippets in memory resource docs
### Is this a duplicate?

New

https://nvidia.github.io/cccl/libcudacxx/extended_api/memory_resource/resource_ref.html seems to suggest that one can use `cuda::mr::resource_ref<>` with an empty property list. However, it must in fact be a non-empty list.

Also on that page, `cuda::mr::get_property` needs to take a `required_alignment{}` (not the type). And maybe that needs to be the free function `get_property`?

The `buggy_allocate_with_alignment` is buggy because `cuda::has_property` takes two template arguments, the type to query for properties and the property.

On https://nvidia.github.io/cccl/libcudacxx/extended_api/memory_resource/resource.html

```
struct valid_resource {
  void* allocate_sync(std::size_t, std::size_t) { return nullptr; }
  void deallocate_sync(void*, std::size_t, std::size_t) noexcept {}
  bool operator==(const valid_resource&) const { return true; }
  // NOTE: C++20 thankfully added operator rewrite rules so defining operator!= is not required.
  // However, if compiled with C++14 / C++17, operator != must also be defined.
  bool operator!=(const valid_resource&) const { return false; }
};
static_assert(cuda::mr::resource<valid_resource>, "");
```

Fails because `valid_resource` only satisfies `cuda::mr::synchronous_resource`.

_No response_

edenfunf:
## What

Fix several incorrect code examples in the libcu++ memory resource documentation that would produce compile errors if copied verbatim.

### resource.rst

The first `valid_resource` example (under the `cuda::synchronous_resource` concept section) only defines `allocate_sync`/`deallocate_sync`, which satisfies `cuda::mr::synchronous_resource` but **not** `cuda::mr::resource` (which additionally requires stream-ordered `allocate`/`deallocate`). The `static_assert` was using the wrong concept.

**Before:**
```cpp
struct valid_resource {
  void* allocate_sync(std::size_t, std::size_t) { return nullptr; }
  void deallocate_sync(void*, std::size_t, std::size_t) noexcept {}
  ...
};
static_assert(cuda::mr::resource<valid_resource>, "");  // fails to compile
```

**After:**
```cpp
static_assert(cuda::mr::synchronous_resource<valid_resource>, "");
```

### wrappers.rst

Three issues in the `resource_ref` examples:

1. `required_alignment` was defined as an empty struct with no `value_type`, so `get_property` would return `void` — unusable as the `align` argument. Added `using value_type = std::size_t;`.

2. `cuda::mr::get_property` does not exist; `get_property` is found via ADL. Fixed to use the unqualified call.

3. `required_alignment` (a type name) was passed where an instance is required. Fixed to `required_alignment{}`.

4. The comment `"BUG: This will always be false"` for `cuda::has_property<required_alignment>` was misleading. The actual problem is that `cuda::has_property` requires two template arguments (resource type + property type); with one argument the expression is ill-formed. Updated the comment accordingly.

## Why

Fixes #6843 — users copying these snippets would get compile errors unrelated to the concept being demonstrated.

## How

Documentation-only changes to two `.rst` files. No header or test modifications.

## Test

Doc-only change; no compilation required. The fixes bring the snippets in line with the actual API (verified against the headers in `libcudacxx/include/cuda/__memory_resource/`) and the patterns used in the existing test suite (e.g., `resource_ref/properties.pass.cpp`).

### DIFF
--- a/docs/libcudacxx/extended_api/memory_resource/wrappers.rst
+++ b/docs/libcudacxx/extended_api/memory_resource/wrappers.rst
@@ -66,19 +66,26 @@ lifetime.
 
 .. code:: cpp
 
-   struct required_alignment{};
+   struct required_alignment {
+       using value_type = std::size_t;
+   };
    void* do_allocate_with_alignment(cuda::mr::resource_ref<required_alignment> resource, cuda::stream_ref stream, std::size_t size) {
-       return resource.allocate(stream, size, cuda::mr::get_property(resource, required_alignment));
+       return resource.allocate(stream, size, get_property(resource, required_alignment{}));
    }
 
 However, the type erasure comes with the cost that arbitrary properties cannot be queried from either wrapper:
 
 .. code:: cpp
 
-   struct required_alignment{};
+   struct required_alignment {
+       using value_type = std::size_t;
+   };
    void* buggy_allocate_with_alignment(cuda::mr::resource_ref<> resource, cuda::stream_ref stream, std::size_t size) {
-       if constexpr (cuda::has_property<required_alignment>) { // BUG: This will always be false
-           return resource.allocate(stream, size, cuda::mr::get_property(resource, required_alignment));
+       // BUG: cuda::has_property requires two template arguments: the resource type and the property.
+       // This will not compile. The correct form is cuda::has_property<decltype(resource), required_alignment>,
+       // but even then it evaluates to false because resource_ref<> was not constrained with required_alignment.
+       if constexpr (cuda::has_property<required_alignment>) {
+           return resource.allocate(stream, size, get_property(resource, required_alignment{}));
        } else {
            return resource.allocate(stream, size, my_default_alignment);
        }
--- a/docs/libcudacxx/extended_api/memory_resource/wrappers.rst
+++ b/docs/libcudacxx/extended_api/memory_resource/wrappers.rst
@@ -81,10 +81,7 @@ However, the type erasure comes with the cost that arbitrary properties cannot b
        using value_type = std::size_t;
    };
    void* buggy_allocate_with_alignment(cuda::mr::resource_ref<> resource, cuda::stream_ref stream, std::size_t size) {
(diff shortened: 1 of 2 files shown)


docs/libcudacxx/extended_api/memory_resource/wrappers.rst
```diff
@@ -66,19 +66,26 @@ lifetime.
...
+   struct required_alignment {
+       using value_type = std::size_t;
+   };
    void* buggy_allocate_with_alignment(cuda::mr::resource_ref<> resource, cuda::stream_ref stream, std::size_t size) {
-       if constexpr (cuda::has_property<required_alignment>) { // BUG: This will always be false
-           return resource.allocate(stream, size, cuda::mr::get_property(resource, required_alignment));
+       // BUG: cuda::has_property requires two template arguments: the resource type and the property.
+       // This will not compile. The correct form is cuda::has_property<decltype(resource), required_alignment>,
+       // but even then it evaluates to false because resource_ref<> was not constrained with required_alignment.
+       if constexpr (cuda::has_property<required_alignment>) {
```
pciolkosz: I would just correct the example to use `has_property` correctly, to focus on showing that the property check will always be false

pciolkosz (commented):
thank you for looking into it @edenfunf

docs/libcudacxx/extended_api/memory_resource/wrappers.rst
edenfunf: Good point — updated. The example now uses cuda::has_property<decltype(resource), required_alignment> (correct two-argument form), keeping the focus on the actual lesson: even with the right syntax, the check is always false because resource_ref<> carries no property constraints at compile time.

pciolkosz (approved):
Looks good, thanks again

Issue #8207, opened 2026-03-28
[BUG] ScanTileState static_assert fails for types with is_primitive=true and sizeof >= 16


leofang:
## Summary

After #7575 (commit b9a0994c1c, "Support more types in decoupled lookback fastpath"), `cub::DeviceScan::InclusiveSum` fails to compile for types that are marked as primitive via `cub::NumericTraits` but have `sizeof(T) >= 16`, such as `thrust::complex<double>`.

## Root cause

#7575 changed the default template parameter of `ScanTileState` to include `is_primitive<T>` in the `SingleWord` condition:

```cpp
// single_pass_scan_operators.cuh, line ~657
bool SingleWord = detail::is_primitive<T>::value
                   || ::cuda::std::is_trivially_copyable_v<T>
                   && sizeof(T) < detail::largest_atomic_message_size
```

Due to C++ operator precedence (`&&` binds tighter than `||`), this evaluates as:

```cpp
is_primitive<T>::value || (is_trivially_copyable_v<T> && sizeof(T) < 16)
```

When `is_primitive<T>::value` is `true`, `SingleWord` is unconditionally `true` regardless of `sizeof(T)`. This selects the `ScanTileState<T, true>` specialization, which then hits:

```cpp
static_assert(sizeof(TileDescriptor) <= detail::largest_atomic_message_size);
```

For `thrust::complex<double>` (`sizeof == 16`), `TileDescriptor = {StatusWord, T}` exceeds 16 bytes.

## Who is affected

Projects like [CuPy](https://github.com/cupy/cupy) specialize `cub::NumericTraits` for complex types to enable CUB device algorithms:

```cpp
template <>
struct NumericTraits<thrust::complex<double>>
    : BaseTraits<FLOATING_POINT, true, unsigned long long,
                 thrust::complex<double>> {};
```

The second `BaseTraits` template parameter (`true`) marks the type as "primitive". This pattern worked on all prior CCCL releases and breaks after #7575.

## Minimal reproducer

```cpp
#include <thrust/complex.h>
#include <cuda/std/limits>

// Trait specializations to enable CUB device algorithms on complex types
// (same pattern used by CuPy)
namespace cuda {
namespace std {
template <>
class numeric_limits<thrust::complex<double>> {
  public:
    static __host__ __device__ thrust::complex<double> max() noexcept {
        return thrust::complex<double>(
            cuda::std::numeric_limits<double>::max(),
            cuda::std::numeric_limits<double>::max());
    }
[...]

PR #8414, opened 2026-04-14
Assert that primitive types' `UnsignedBits` have the same size

bernhardmgruber:
Spin off from: #8207

### DIFF
--- a/cub/cub/util_type.cuh
+++ b/cub/cub/util_type.cuh
@@ -831,6 +831,7 @@ namespace detail
 {
 struct is_primitive_impl;
 
+// case for _CATEGORY = NOT_A_NUMBER, or _PRIMITIVE = false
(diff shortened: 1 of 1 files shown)


davebayer (approved):
Q: Shouldn't we force the same alignment, too? It should be fine for bit casting without the check, but `*reinterpret_cast<T*>(...)` could crash

bernhardmgruber:
> Q: Shouldn't we force the same alignment, too? It should be fine for bit casting without the check, but `*reinterpret_cast<T*>(...)` could crash

Maybe. But maybe we are over restrictive then. We never actually specify what a user can do with `cub::Traits<T>::UnsignedBits`. I think the idea was that it is a type that can hold all bits of `T` for radix sort. I don't know whether aliasing was actually intended. However, I see that the top k implementation reinterpret_casts a key to its unsigned bits, so we seem to require it. Not sure what to do. I am afraid of breaking a lot of users.

PR #8419, opened 2026-04-14
[libcu++] Add support for family-specific architectures

davebayer:
This PR adds support for family specific architectures in libcu++ (`sm_XYf`)

### DIFF
--- a/libcudacxx/include/cuda/__device/arch_id.h
+++ b/libcudacxx/include/cuda/__device/arch_id.h
@@ -37,29 +37,41 @@ _CCCL_BEGIN_NAMESPACE_CUDA
 //! capability. For example, sm_90 and sm_90a have the same compute capability, but the identifier is different.
 enum class arch_id : int
 {
-#define _CCCL_DEFINE_ARCH_ID(_CC)          sm_##_CC = _CC,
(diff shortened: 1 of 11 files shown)


bernhardmgruber (approved):

libcudacxx/include/cuda/__device/arch_id.h
```diff
@@ -37,29 +37,41 @@ _CCCL_BEGIN_NAMESPACE_CUDA
 //! capability. For example, sm_90 and sm_90a have the same compute capability, but the identifier is different.
 enum class arch_id : int
 {
-#define _CCCL_DEFINE_ARCH_ID(_CC)          sm_##_CC = _CC,
-#define _CCCL_DEFINE_ARCH_SPECIFIC_ID(_CC) sm_##_CC##a = _CC * __arch_specific_id_multiplier,
+#define _CCCL_DEFINE_ARCH_ID(_CC)            sm_##_CC = _CC,
+#define _CCCL_DEFINE_FAMILY_SPECIFIC_ID(_CC) sm_##_CC##f = _CC + __family_specific_id_offset,
+#define _CCCL_DEFINE_ARCH_SPECIFIC_ID(_CC)   sm_##_CC##a = _CC + __arch_specific_id_offset,
```
davebayer, line 42: This is ABI break, but I don't think there is any other way to fit the data here. I could keep the `a` variants to use `*` and `f` variants to use `+`

PR #8421, opened 2026-04-14
New tuning API for DeviceUniqueByKey

### LINKED ISSUE #7479: Refactor cccl.c unique_by_key to new tuning API

### LINKED ISSUE #7647: Implement the new tuning API for `DispatchUniqueByKey`

griwes:
## Description

Implements the new, improved, tuning API for UniqueByKey.

Resolves #7647.
Resolves #7479.

## Checklist


griwes:
SASS verification pending, posted already primarily for the purpose of CI runs.

cub/cub/device/dispatch/tuning/tuning_unique_by_key.cuh
```diff
@@ -862,6 +863,685 @@ struct policy_hub
...
+              switch (value_size)
+              {
+                case 1:
+                  // ipt_12.tpb_512.trp_0.ld_0.ns_948.dcid_5.l2w_955 1.121279  1.000000  1.114566  1.43765
+                  return {512,
+                          12,
+                          BLOCK_LOAD_DIRECT,
+                          LOAD_DEFAULT,
+                          BLOCK_SCAN_WARP_SCANS,
+                          delay_constructor_policy_from_type<exponential_backon_jitter_window_constructor_t<948, 955>>};
```
bernhardmgruber, line 894: :D :D :D :D

cub/cub/device/dispatch/tuning/tuning_unique_by_key.cuh
```diff
@@ -862,6 +863,685 @@ struct policy_hub
...
+                case 2:
+                  // ipt_14.tpb_512.trp_0.ld_0.ns_1228.dcid_7.l2w_320 1.151229  1.007229  1.151131  1.443520
+                  return {512,
+                          14,
+                          BLOCK_LOAD_DIRECT,
+                          LOAD_DEFAULT,
+                          BLOCK_SCAN_WARP_SCANS,
+                          delay_constructor_policy_from_type<exponential_backon_constructor_t<1228, 320>>};
```
bernhardmgruber, line 902: Suggestion: I would prefer to construct the `delay_constructor_policy` directly here instead of creating the type and converting it.

```suggestion
                          delay_constructor_policy{delay_constructor_kind::exponential_backon, 1228, 320};
```
I had this generated in the past as well, but Claude could easily rewrite this.

cub/cub/device/dispatch/tuning/tuning_unique_by_key.cuh
```diff
@@ -862,6 +863,685 @@ struct policy_hub
...
+                case 4:
+                  // todo(gonidelis): tuning performs very well for medium input size, regresses for large input sizes.
+                  // find better tuning.
+                  // ipt_14.tpb_512.trp_0.ld_0.ns_408.dcid_7.l2w_960 1.136333  0.995833  1.144371  1.448687
+                  return {512,
+                          14,
+                          BLOCK_LOAD_WARP_TRANSPOSE,
+                          LOAD_DEFAULT,
+                          BLOCK_SCAN_WARP_SCANS,
+                          delay_constructor_policy_from_type<no_delay_constructor_t<1045>>};
```
bernhardmgruber, line 992: In this case, the sm100 tuning was not sufficiently better than the older sm90 tuning, so @gonidelis added the TODO comment and we fell back to the older tuning like this:
```
template <class KeyT, class ValueT>
struct sm100_tuning<KeyT, ValueT, primitive_key::yes, primitive_val::yes, key_size::_4, val_size::_4>
    : sm90_tuning<KeyT, ValueT, primitive_key::yes, primitive_val::yes, key_size::_4, val_size::_4> { ... }
```
This "fallback" behavior is less visible here now. The tuning values here are from the SM90 tuning, which is correct, but it's not immediately obvious.

Suggestion: could we create a function `get_sm100_tuning() -> optional<unique_by_key_policy>` that would not return a policy for this case. And outside in the `operator()` we would do:
```c++
if (arch >= ::cuda::arch_id::sm_100) {
  if (auto opt_tuning = get_sm100_tuning())
    return *opt_tuning;
}
if (arch >= ::cuda::arch_id::sm_90) {
  if (auto opt_tuning = get_sm90_tuning())
    return *opt_tuning;
}
...
```
This would split up the `operator()` a bit and would better encode this fallback behavior.

cub/cub/device/dispatch/tuning/tuning_unique_by_key.cuh
```diff
@@ -862,6 +863,685 @@ struct policy_hub
...
+          delay_constructor_policy_from_type<typename policy_t::detail::delay_constructor_t>};
+}
+
+template <typename PolicyHub>
+struct policy_selector_from_hub
+{
+  [[nodiscard]] _CCCL_API constexpr auto operator()(::cuda::arch_id arch) const -> unique_by_key_policy
+  {
+    (void) arch;
+    using UniqueByKeyPolicyT = typename PolicyHub::MaxPolicy::UniqueByKeyPolicyT;
```
bernhardmgruber, line 1530: Important: Reaching out to `MaxPolicy` is only correct in device code, so let's make sure we are not calling this on the host:

```suggestion
  [[nodiscard]] _CCCL_DEVICE_API constexpr auto operator()(::cuda::arch_id arch) const -> unique_by_key_policy
  {
    (void) arch;
    using UniqueByKeyPolicyT = typename PolicyHub::MaxPolicy::UniqueByKeyPolicyT;
```
Edit: I think this is later used in host code in the dispatcher. Please rewrite this to do the `MaxPolicy::Invoke` thing and extract the policy in the closure type. I think you did this correctly in other PRs, so you know what I mean ;)

cub/cub/device/device_select.cuh
```diff
@@ -36,11 +36,9 @@ CUB_NAMESPACE_BEGIN
 
 namespace detail::unique_by_key_tuning
 {
-// TODO(bgruber): drop this after rewriting to the new tuning API
 struct get_tuning_query_t
```
bernhardmgruber, line 40: Remark: I don't mind if the comment is deleted, but the TODO is not addressed yet. This will be done with #7464, which will properly extract the tuning policies from the passed environments. But out of scope for this PR.

cub/cub/device/dispatch/tuning/tuning_unique_by_key.cuh
griwes: I believe I fixed the dispatch path with the c.parallel changes, but yeah, I'll mark this function correctly; good catch.

griwes:
SASS comparison is showing no diffs.

cub/cub/device/dispatch/tuning/tuning_unique_by_key.cuh
griwes: I think :D :D :D :D :D :D is more appropriate, given the number of opening braces in that hunk :P

bernhardmgruber (approved):

PR #5143, opened 2025-07-03
Add env-based overloads for DeviceReduce::(Arg)MinMax

gonidelis:
fixes #5121 

Docs and tests added

PR #8428, opened 2026-04-15
Fix bugs in DeviceReduce::ArgMax env overloads and add missing tests

gonidelis:
Fixing critical oversights in one of my [earliest](https://github.com/NVIDIA/cccl/pull/5143) env PRs for `ArgMIn/Max` 

  - Fix missing `d_max_out` arg in `ArgMax compare_op` env overload's `__arg_min_env` call                                                           
  - Fix user-provided env silently ignored, hardcoded env{} passed instead
  - Add missing `enable_if_t<!indirectly_comparable>` SFINAE guard on `ArgMax no-compare_op` env overload                                          
  - Add launch-wrapper tests (lid 0/1/2) for Min, Max, ArgMin, ArgMax with and without compare_op
  - Move `stream.sync()` after example-end markers in env_api tests to hide from doc snippet

### DIFF
--- a/cub/cub/device/device_reduce.cuh
+++ b/cub/cub/device/device_reduce.cuh
@@ -1992,8 +1992,7 @@ public:
     EnvT env = {})
   {
     _CCCL_NVTX_RANGE_SCOPE("cub::DeviceReduce::ArgMax");
-    return __arg_min_env(
-      d_in, d_index_out, num_items, detail::arg_less{detail::swap_args{compare_op}}, ::cuda::std::execution::env{});
+    return __arg_min_env(d_in, d_max_out, d_index_out, num_items, detail::arg_less{detail::swap_args{compare_op}}, env);
   }
 
   //! @overload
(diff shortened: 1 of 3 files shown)


cub/cub/device/device_reduce.cuh
```diff
@@ -1992,16 +1992,18 @@ public:
     EnvT env = {})
   {
     _CCCL_NVTX_RANGE_SCOPE("cub::DeviceReduce::ArgMax");
-    return __arg_min_env(
-      d_in, d_index_out, num_items, detail::arg_less{detail::swap_args{compare_op}}, ::cuda::std::execution::env{});
+    return __arg_min_env(d_in, d_max_out, d_index_out, num_items, detail::arg_less{detail::swap_args{compare_op}}, env);
```
bernhardmgruber, line 1995: Nice. I think I introduced this in #8285. What a fail :D

bernhardmgruber (approved):
I think this is more of a cleanup of my sloppy work in #8285 . Thank you so much for that!

PR #8429, opened 2026-04-15
Fix nvcc compile error: concept-id inside enable_if_t NTTP

gonidelis:
fixes https://nvbugspro.nvidia.com/bug/6079906

Replaces raw `enable_if_t`s with `_CCCL_REQUIRES` across 5 affected overloads in `device_reduce.cuh`.

This was a latent bug in the `ArgMin/ArgMax` overloads that used `enable_if_t<indirectly_comparable<...>>` directly instead of going through the `_CCCL_TEMPLATE/_CCCL_REQUIRES` macros. Older nvcc versions only diagnosed this at template instantiation time, and no test in the libcudacxx or cub suites ever instantiated these particular overloads (they require passing a custom comparator). Newer nvcc now diagnoses it eagerly at parse time, which surfaced the bug.

### DIFF
--- a/cub/cub/device/device_reduce.cuh
+++ b/cub/cub/device/device_reduce.cuh
@@ -1062,14 +1062,10 @@ public:
   //!   @rst
   //!   **[optional]** CUDA stream to launch kernels within. Default is stream\ :sub:`0`.
   //!   @endrst
-  template <
-    typename InputIteratorT,
-    typename ExtremumOutIteratorT,
-    typename IndexOutIteratorT,
-    typename CompareOpT,
-    // TODO(bgruber): this constraint is not accurate, since the implementation will compare the value types of
-    // ExtremumOutIteratorT, which is wrong IMO
-    ::cuda::std::enable_if_t<::cuda::std::indirectly_comparable<InputIteratorT, InputIteratorT, CompareOpT>, int> = 0>
+  // TODO(bgruber): this constraint is not accurate, since the implementation will compare the value types of
+  // ExtremumOutIteratorT, which is wrong IMO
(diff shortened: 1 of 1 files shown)


bernhardmgruber (approved):
Thx for the improvement! I still can't see how the original code would trigger the error:
```
error: type/value mismatch at argument 1 in template parameter list for 'template<bool _Bp, class _Tp> using cuda::std::__4::enable_if_t = typename cuda::std::__4::enable_if::type'
 1069 |     typename CompareOpT,
```
In any case, this PR is an improvement.

miscco (changes requested):
I would prefer that we better understand the underlying issue rather than just changing to a different formulation.

The classical `enable_if` SFINAE should work in this case. What I could imagine is that our choice of default type `, int> = 0>` clashes with other overloads,'

Nevertheless this seems more like a bug on the compiler side rather than a bug on our side

gonidelis:
i reproed and it works with this fix. Supposedly, in the concept path the compiler sees the enable if<normat_iter>> thingy as an nttp concept id

________________________________

gonidelis:
@miscco 

Can you elaborate on why the classical `enable_if` should work still when concepts enabled?

do you oppose in getting this merged regardless?

gonidelis:
Do not merge. This https://gist.github.com/gonidelis/ee5d0c3811d475f6930c718f1e5fc061 compiles with clang and not with most recent dev nvcc which proves it most probably is a compiler bug and not CCCL

miscco (approved):

bernhardmgruber:
> I would prefer that we better understand the underlying issue rather than just changing to a different formulation.

It's a GCC bug: https://gcc.godbolt.org/z/81P41Wa68

miscco:
So the actual issue is the defaulted template argument for the concept

PR #8432, opened 2026-04-15
Add env overload for DevicePartition::If three-way 

gonidelis:
missed it during #7459

### DIFF
--- a/cub/test/catch2_test_device_partition_env.cu
+++ b/cub/test/catch2_test_device_partition_env.cu
@@ -30,10 +30,6 @@ struct greater_than_t
 {
   T compare;
 
-  explicit __host__ greater_than_t(T compare)
-      : compare(compare)
-  {}
-
   __host__ __device__ bool operator()(const T& a) const
   {
     return a > compare;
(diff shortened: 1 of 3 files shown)


cub/test/catch2_test_device_partition_env.cu
```diff
@@ -25,6 +25,21 @@ DECLARE_LAUNCH_WRAPPER(cub::DevicePartition::Flagged, device_partition_flagged);
...
 
+template <typename T>
+struct greater_than_t
+{
+  T compare;
+
+  explicit __host__ greater_than_t(T compare)
+      : compare(compare)
+  {}
+
```
bernhardmgruber: Suggestion: rely on CTAD
```suggestion
```

cub/test/catch2_test_device_partition_env_api.cu
```diff
@@ -15,6 +15,21 @@
...
 
+template <typename T>
+struct greater_than_t
+{
+  T compare;
+
+  explicit __host__ greater_than_t(T compare)
+      : compare(compare)
+  {}
+
```
bernhardmgruber:
```suggestion
```

bernhardmgruber (approved):

PR #5335, opened 2025-07-22
Add env-based cub::DeviceSelect algorithms

gonidelis:
partially fixes #7552 

Adds `Select::If` and `Select::Flagged`

`cub::DeviceSelect::FlaggedIf()` `cub::DeviceSelect::Unique()` and  `cub::DeviceSelect::UniqueByKey()` to be added in a subsequent PR.

PR #8434, opened 2026-04-15
Add env overload for DeviceSelect::Unique with custom `EquialityOpT`

gonidelis:
slipped while #5335 

Adds `DeviceSelect::Unique` that accepts op. The previous `Unique` that was added didn't accept any operators

### DIFF
--- a/cub/test/catch2_test_device_select_env_api.cu
+++ b/cub/test/catch2_test_device_select_env_api.cu
@@ -216,15 +216,13 @@ C2H_TEST("cub::DeviceSelect::Unique with custom equality_op accepts env with str
   auto output       = thrust::device_vector<int>(8);
   auto num_selected = thrust::device_vector<int>(1);
 
-  auto eq_mod3 = [] __host__ __device__(int a, int b) {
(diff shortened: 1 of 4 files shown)


bernhardmgruber (approved):

cub/test/catch2_test_device_select_env_api.cu
```diff
@@ -208,6 +208,41 @@ C2H_TEST("cub::DeviceSelect::Unique accepts env with stream", "[select][env]")
   REQUIRE(num_selected == expected_num_selected);
 }
 
+C2H_TEST("cub::DeviceSelect::Unique with custom equality_op accepts env with stream", "[select][env]")
+{
+  // example-begin select-unique-eqop-env
+  // Unique by modulo 3 — consecutive elements are "equal" if they have the same remainder mod 3
```
bernhardmgruber: Remark: Would sound better in my head, but I am not a native speaker
```suggestion
  // Unique modulo 3 — consecutive elements are "equal" if they have the same remainder mod 3
```

PR #8435, opened 2026-04-15
Refactor and improve `count_as` queries

davebayer:
Similar to #8394, just for `count_as`.

### DIFF
--- a/libcudacxx/include/cuda/__hierarchy/queries/count.h
+++ b/libcudacxx/include/cuda/__hierarchy/queries/count.h
@@ -1,0 +1,113 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of libcu++, the C++ Standard Library for your entire system,
+// under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef _CUDA___HIERARCHY_QUERIES_COUNT_H
+#define _CUDA___HIERARCHY_QUERIES_COUNT_H
+
+#include <cuda/std/detail/__config>
+
+#if defined(_CCCL_IMPLICIT_SYSTEM_HEADER_GCC)
+#  pragma GCC system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_CLANG)
+#  pragma clang system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_MSVC)
+#  pragma system_header
+#endif // no system header
+
+#if _CCCL_HAS_CTK()
+
+#  include <cuda/__fwd/hierarchy.h>
+#  include <cuda/std/__cstddef/types.h>
+
+#  include <cuda/std/__cccl/prologue.h>
+
+_CCCL_BEGIN_NAMESPACE_CUDA
+
+// native hierarchy queries
+
+#  if _CCCL_CUDA_COMPILATION()
+
+// cudafe++ makes the queries (that are device only) return void when compiling for host, which causes host compilers
+// to warn about applying [[nodiscard]] to a function that returns void.
+_CCCL_DIAG_PUSH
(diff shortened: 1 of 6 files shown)


libcudacxx/include/cuda/__hierarchy/queries/count.h
```diff
@@ -0,0 +1,113 @@
...
+
+#if defined(_CCCL_IMPLICIT_SYSTEM_HEADER_GCC)
+#  pragma GCC system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_CLANG)
+#  pragma clang system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_MSVC)
+#  pragma system_header
+#endif // no system header
+
+#if _CCCL_HAS_CTK()
```
Jacobfaib, line 24: Does this actually need the CTK? You already gate off any CUDA specific things behind `_CCCL_CUDA_COMPILATION()`

libcudacxx/include/cuda/__hierarchy/queries/count.h
```diff
@@ -0,0 +1,113 @@
...
+template <class _Unit, class _Level>
+struct __count_query_native
+{
+  template <class _Tp>
+  [[nodiscard]] _CCCL_DEVICE_API static _Tp __call() noexcept
+  {
+    const auto __exts = __extents_query_native<_Unit, _Level>::template __call<_Tp>();
+
+    _Tp __ret = 1;
+    for (::cuda::std::size_t __i = 0; __i < __exts.rank(); ++__i)
```
Jacobfaib, line 55: Assuming `__exts` is an mdspan, `rank()` is not guaranteed to be `size_t`. Technically it should be `decltype(__exts)::rank_type`.

libcudacxx/include/cuda/__hierarchy/queries/count.h
```diff
@@ -0,0 +1,113 @@
...
+template <class _Unit, class _Level>
+struct __count_query
+{
+  template <class _Tp, class _Hierarchy>
+  [[nodiscard]] _CCCL_API static constexpr _Tp __call(const _Hierarchy& __hier) noexcept
+  {
+    const auto __exts = __extents_query<_Unit, _Level>::template __call<_Tp>(__hier);
+
+    _Tp __ret = 1;
+    for (::cuda::std::size_t __i = 0; __i < __exts.rank(); ++__i)
```
Jacobfaib, line 99: Same here with rank.

libcudacxx/include/cuda/__hierarchy/queries/index.h
```diff
@@ -80,6 +80,8 @@ struct __index_query_native<thread_level, warp_level>
   template <class _Tp>
   [[nodiscard]] _CCCL_DEVICE_API static hierarchy_query_result<_Tp> __call() noexcept
   {
+    // todo(dabayer): Is it worth using uda::ptx::get_sreg_laneid() here? Doesn't it prevent some other optimizations
```
Jacobfaib, line 83:
```suggestion
    // TODO(dabayer): Is it worth using cuda::ptx::get_sreg_laneid() here? Doesn't it prevent some other optimizations
```

libcudacxx/include/cuda/__hierarchy/queries/count.h
davebayer: From what I see, `std::extents` always define `rank_type` to `size_t`, see https://en.cppreference.com/w/cpp/container/mdspan/extents.html

Also, this is not coming from mdspan :)

libcudacxx/include/cuda/__hierarchy/queries/count.h
davebayer: You can `cuda::launch` even from a C++ source file, so all we need is CTK in that case. And the dimensions related queries don't need CUDA compilation

libcudacxx/include/cuda/__hierarchy/queries/index.h
davebayer: Oops, I will fix this in next PR, I don't want to rerun the CI because of it

miscco (approved):

libcudacxx/include/cuda/__hierarchy/queries/index.h
davebayer: Fixed in #8478

PR #8438, opened 2026-04-15
Enable clang-tidy modernize checks

Jacobfaib:
## Checklist

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -23,7 +23,41 @@ Checks:
   - 'performance-*'
   - '-performance-enum-size'
   - '-performance-unnecessary-value-param'
+  - 'modernize-*'
+  - '-modernize-use-std-numbers'
+  - '-modernize-deprecated-headers'
+  - '-modernize-min-max-use-initializer-list'
+  - '-modernize-use-emplace'
+  - '-modernize-make-unique'
+  - '-modernize-use-using'
+  - '-modernize-use-integer-sign-comparison'
+  - '-modernize-concat-nested-namespaces'
+  - '-modernize-use-nodiscard'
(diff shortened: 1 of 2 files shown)


.clang-tidy
```diff
@@ -23,7 +23,41 @@ Checks:
...
+  - '-modernize-type-traits'
+  - '-modernize-use-nullptr'
+  - '-modernize-return-braced-init-list'
+  - '-modernize-use-auto'
+  - '-modernize-redundant-void-arg'
+  - '-modernize-use-equals-delete'
+  - '-modernize-pass-by-value'
+  - '-modernize-loop-convert'
+  - '-modernize-unary-static-assert'
+  - '-modernize-use-override'
```
bernhardmgruber, line 48: Q: Are we going to have a discussion about which checks we want to enable, or is was this choice already made? I would enable a lot of these.

bernhardmgruber (approved):

.clang-tidy
Jacobfaib: They will all be enabled.

For each of the `clang-tidy` checks, the procedure is to first open a big umbrella PR in draft mode (in this case https://github.com/NVIDIA/cccl/pull/8344) and fix _all_ the errors in that, splitting each check into a separate commit. Once CI is green, I open a setup PR (this one) that enables the category but disables all the checks that need fixing. Once the setup PR goes in, I cherry-pick every commit from the umbrella PR into a separate branch and PR which also re-enables that particular check.

Jacobfaib: I could do it all in one shot, but I figure you'd rather not review... 1906 files in one go, even if they are split (rather painstakingly!!) by commit :)

PR #8449, opened 2026-04-15
modernize-use-integer-sign-comparison

Jacobfaib:
## Description

closes 

## Checklist

### DIFF
--- a/cub/cub/warp/specializations/warp_scan_smem.cuh
+++ b/cub/cub/warp/specializations/warp_scan_smem.cuh
@@ -32,6 +32,7 @@
 #include <cuda/std/__functional/operations.h>
 #include <cuda/std/__type_traits/integral_constant.h>
 #include <cuda/std/__type_traits/is_integral.h>
+#include <cuda/std/__utility/cmp.h>
 
 CUB_NAMESPACE_BEGIN
 namespace detail
@@ -111,7 +112,7 @@ struct WarpScanSmem
     __syncwarp(member_mask);
 
     // Update partial if addend is in range
-    if (HAS_IDENTITY || (lane_id >= OFFSET))
+    if (HAS_IDENTITY || ::cuda::std::cmp_greater_equal(lane_id, OFFSET))
     {
       T addend = (T) ThreadLoad<LOAD_VOLATILE>(&temp_storage[HALF_WARP_THREADS + lane_id - OFFSET]);
       partial  = scan_op(addend, partial);
@@ -240,7 +241,7 @@ struct WarpScanSmem
     __syncwarp(member_mask);
 
     // Iterate scan steps
-    if (static_cast<int>(lane_id) < valid_items)
+    if (::cuda::std::cmp_less(lane_id, valid_items))
     {
       inclusive_output = input;
     }
@@ -260,7 +261,7 @@ struct WarpScanSmem
       __syncwarp(member_mask);
       T addend = temp_storage[HALF_WARP_THREADS + lane_id - offset];
       // Update partial if addend is in range
-      if ((lane_id >= offset) && (static_cast<int>(lane_id) < valid_items))
+      if (::cuda::std::cmp_greater_equal(lane_id, offset) && ::cuda::std::cmp_less(lane_id, valid_items))
       {
         inclusive_output = scan_op(addend, inclusive_output);
       }
@@ -481,7 +482,7 @@ struct WarpScanSmem
     if constexpr (::cuda::std::is_integral_v<T> && cub::detail::is_cuda_std_plus_v<ScanOpT, T>)
     {
       // initial value presumed 0
-      if (static_cast<int>(lane_id) < valid_items)
+      if (::cuda::std::cmp_less(lane_id, valid_items))
       {
         exclusive = inclusive - input;
       }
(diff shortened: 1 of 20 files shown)


fbusato (changes requested):
I'm not in favor of that. This can change the performance. If we know that two values are positive (and within the range), we can compare `int` and `unsigned`

Jacobfaib:
> I'm not in favor of that. This can change the performance. If we know that two values are positive (and within the range), we can compare int and unsigned

I think in that case we should assert that at compile-time by using appropriate types. So storing strictly positive values in unsigned types etc

Jacobfaib:
Since most of these are in CUB-land, I would ask that @bernhardmgruber or other CUB maintainers please take a close look at this PR and let me know which values could be made unsigned. I see that a lot of CUB tuning parameters just use `int` (for example `BlockRakingLayout::RAKING_THREADS`) when these can be made unsigned.

libcudacxx/include/cuda/std/__thread/threading_support_cuda.h
```diff
@@ -31,12 +32,12 @@ _CCCL_BEGIN_NAMESPACE_CUDA_STD
...
-_CCCL_API inline void __cccl_thread_sleep_for(::cuda::std::chrono::nanoseconds __ns){
-  NV_IF_TARGET(NV_PROVIDES_SM_70, ({
-                 auto const __step = __ns.count();
-                 _CCCL_ASSERT(__step < numeric_limits<unsigned>::max(), "invalid nanoseconds count");
-                 ::__nanosleep((unsigned) __step);
-               }))}
+_CCCL_API inline void __cccl_thread_sleep_for(::cuda::std::chrono::nanoseconds __ns){NV_IF_TARGET(
+  NV_PROVIDES_SM_70, ({
+    auto const __step = __ns.count();
+    _CCCL_ASSERT(::cuda::std::cmp_less(__step, numeric_limits<unsigned>::max()), "invalid nanoseconds count");
```
davebayer:
```suggestion
    _CCCL_ASSERT(::cuda::std::in_range<unsigned>(__step)), "invalid nanoseconds count");
```

cub/cub/warp/specializations/warp_scan_smem.cuh
```diff
@@ -260,7 +261,7 @@ struct WarpScanSmem
       __syncwarp(member_mask);
       T addend = temp_storage[HALF_WARP_THREADS + lane_id - offset];
       // Update partial if addend is in range
-      if ((lane_id >= offset) && (static_cast<int>(lane_id) < valid_items))
+      if (::cuda::std::cmp_greater_equal(lane_id, offset) && ::cuda::std::cmp_less(lane_id, valid_items))
```
davebayer, line 264: I agree with @fbusato, we should be careful about changes in cub. `lane_id` will always be in range `[0;31)`, we should be fine here

Jacobfaib: OK I await decisions then on which values can have their types changed to `unsigned`, otherwise I will just `// NOLINT` everything in CUB.

libcudacxx/include/cuda/std/__thread/threading_support_cuda.h
```diff
@@ -31,12 +32,12 @@ _CCCL_BEGIN_NAMESPACE_CUDA_STD
...
-_CCCL_API inline void __cccl_thread_sleep_for(::cuda::std::chrono::nanoseconds __ns){
-  NV_IF_TARGET(NV_PROVIDES_SM_70, ({
-                 auto const __step = __ns.count();
-                 _CCCL_ASSERT(__step < numeric_limits<unsigned>::max(), "invalid nanoseconds count");
-                 ::__nanosleep((unsigned) __step);
-               }))}
+_CCCL_API inline void __cccl_thread_sleep_for(::cuda::std::chrono::nanoseconds __ns){NV_IF_TARGET(
+  NV_PROVIDES_SM_70, ({
+    auto const __step = __ns.count();
+    _CCCL_ASSERT(::cuda::std::in_range<unsigned>(__step)), "invalid nanoseconds count");
```
davebayer:
```suggestion
    _CCCL_ASSERT(::cuda::std::in_range<unsigned>(__step), "invalid nanoseconds count");
```

miscco (changes requested):
I am also not in favor off those changes.

I believe those a generally a win in user code, but we have to also consider other constraints than user code.
1. Does this affect runtime performance in a tight loop?
2. Does this affect compile times? We are a template heavy library and using `_Width <= __warp_size` is much faster to compile than `::cuda::std::cmp_less_equal(_Width, __warp_size)`
3. Does this affect debug experience? An integer comparison is just an instruction, but now we have to traverse into a nontrivial function. 

So if we can get away with a simple cast, because we know that it will not overflow it is generally

bernhardmgruber:
> I would ask that @bernhardmgruber or other CUB maintainers please take a close look at this PR and let me know which values could be made unsigned.

I had a quick look and agree with @fbusato and @miscco that these changes are highly dangerous to affect performance in unpredictable ways. Any change to kernel code must be thoroughly benchmarked.

I would leave this clang-tidy check disabled for now.

Jacobfaib:
Tabling for CCCL 4.0

PR #8451, opened 2026-04-15
[libcu++] Deprecate the default alignment of shared_resource

### LINKED ISSUE #8063: [BUG] Type-erasing a memory resource changes default allocation alignment from 256 to 16 bytes
## Description

Concrete memory resources (`device_memory_pool`, `legacy_pinned_memory_resource`, etc.) default to `default_cuda_malloc_alignment` (256 bytes) when callers omit the alignment argument. However, when one of these resources is type-erased into `any_resource` / `any_synchronous_resource`, wrapped in `shared_resource`, or adapted with `synchronous_resource_adapter`, the default alignment silently changes — or disappears entirely.

I noticed this when attempting to call `shared_resource`'s `allocate(stream, bytes)` method, only to realize that alignment had to be specified explicitly (there is no two-argument overload with a default alignment).

```cpp
// Calls allocate_sync(bytes, 256) — correct
pool.allocate_sync(bytes);

// Calls allocate_sync(bytes, 16) — silent behavior change
any_synchronous_resource<> any{pool};
any.allocate_sync(bytes);
```

## Background

The codebase uses two alignment constants:
- `default_cuda_malloc_alignment` (256) — required for device-accessible memory
- `default_cuda_malloc_host_alignment` / `alignof(max_align_t)` (~16) — sufficient for host-only memory

Host-only resources like `legacy_pinned_memory_resource` technically only need 16-byte alignment on the host side. But type-erased wrappers (`any_resource`, `shared_resource`, etc.) don't know what they hold — the underlying resource could target device memory, host memory, or both. Defaulting to 16 bytes is unsafe because it silently under-aligns device allocations. **Note that most of the underlying resources that only call CUDA runtime APIs will simply ignore 16-byte alignment requests and will _always_ give 256-byte aligned pointers, but that isn't safe to assume if the user has implemented a custom device allocator like RMM's pool or arena!** Defaulting to 256 bytes is always safe: it satisfies both host and device requirements, at the cost of potentially wasted alignment padding for host-only resources.

Type-erased and wrapper types should default to 256 bytes since they do not know the requirements of the underlying resource. I would accept a solution where resources with only `host_accessible` properties (and no `device_accessible` property) use 16-byte alignments by default, but I think that's potentially messy because some resources might be `device_accessible` at runtime though that might not be statically known.

## Affected locations

**Current host-friendly default (16) should be device-friendly (256):**
- `any_resource.h:116` — `__ibasic_resource::allocate_sync`
- `any_resource.h:123` — `__ibasic_resource::deallocate_sync`
- `any_resource.h:143` — `__ibasic_async_resource::allocate` (2-arg overload)
- `any_resource.h:155` — `__ibasic_async_resource::deallocate` (3-arg overload)
- `shared_resource.h:188` — `shared_resource::allocate_sync`
- `shared_resource.h:197` — `shared_resource::deallocate_sync`

**No default and no convenience overload:**
- `synchronous_resource_adapter.h:67` — `allocate` (no 2-arg overload)
- `synchronous_resource_adapter.h:79` — `allocate_sync` (no default alignment)
- `synchronous_resource_adapter.h:85` — `deallocate` (no 3-arg overload)
- `synchronous_resource_adapter.h:98` — `deallocate_sync` (no default alignment)
- `shared_resource.h:212` — `allocate` (no 2-arg overload)
- `shared_resource.h:227` — `deallocate` (no 3-arg overload)

## Proposed fix

1. Replace `alignof(::cuda::std::max_align_t)` defaults with `::cuda::mr::default_cuda_malloc_alignment` in `any_resource.h` and `shared_resource.h`.
2. Add default `= ::cuda::mr::default_cuda_malloc_alignment` to `allocate_sync` and `deallocate_sync` in `synchronous_resource_adapter.h`.
3. Add 2-arg `allocate(stream, bytes)` and 3-arg `deallocate(stream, ptr, bytes)` convenience overloads to `shared_resource` and `synchronous_resource_adapter`, matching the overloads already present on `__memory_pool_base` and `__ibasic_async_resource`.

edenfunf:
## What

`shared_resource` is the last type-erased resource wrapper that still picks an alignment on the caller's behalf. This deprecates its two-argument `allocate_sync` / `deallocate_sync` overloads, matching what #8116 already did for `any_resource` and `resource_ref`.

**Before:**
```cpp
// Silently allocates with alignof(max_align_t), no diagnostic
shared.allocate_sync(bytes);
```

**After:**
```cpp
// warning: Specify an explicit alignment argument. The default alignment
//          will be removed in a future release.
shared.allocate_sync(bytes);

// The supported spelling
shared.allocate_sync(bytes, alignment);
```

## Why

This revision replaces an earlier one that changed the wrapper defaults from `alignof(max_align_t)` to `default_cuda_malloc_alignment`. Per the review discussion, that was the wrong angle: a wrapper holds any resource, not necessarily a CUDA-compatible one, so it cannot know the alignment requirements of what it holds and should not choose a value at all. #8116 already started moving the wrappers in that direction, and `shared_resource` was simply missed.

Deprecating rather than changing the value means no behavior change for existing code. Callers that omit the alignment keep what they get today and now get a diagnostic pointing at the explicit spelling.

## How

- Remove the `= alignof(::cuda::std::max_align_t)` defaults from `shared_resource::allocate_sync` and `deallocate_sync`.
- Add deprecated two-argument overloads that forward `alignof(::cuda::std::max_align_t)`, using the same deprecation message as the other wrappers.
- The stream-ordered `allocate` / `deallocate` overloads and `synchronous_resource_adapter` already require an explicit alignment and are left untouched.

`any_resource.h` is unchanged from `main` in this revision.

## Scope note

This does not change the effective alignment of any call today. The discrepancy described in #8063 goes away once the deprecated overloads are removed, not with this PR, hence `Refs` rather than `Fixes`. Happy to follow up with the removal whenever you consider the deprecation period over.

## Test

Verified locally on Windows (MSVC 19.50, CUDA 13.3, sm_120):

- `cuda/memory_resource` lit suite: 23/23 pass.
- C2H tests `shared_resource`, `synchronous_adapter`, `shared_block_ptr`, `buffer.constructor`, `buffer.copy`: all pass.
- `shared_resource.h` compiles standalone via the `internal_headers` target.
- Explicit-alignment calls compile clean under `-Werror=all-warnings`; omitting the alignment emits the same C4996 text as the `any_resource` equivalent.

Not covered locally: GCC and Clang, which take the `__attribute__((deprecated(MSG)))` branch of `CCCL_DEPRECATED_BECAUSE`. The attribute placement is copied verbatim from the existing overloads in `any_resource.h`.

There are no non-test callers of `cuda::mr::shared_resource` in the repository, so nothing internal trips the new deprecation.

Refs #8063

### DIFF
--- a/libcudacxx/include/cuda/__memory_resource/shared_resource.h
+++ b/libcudacxx/include/cuda/__memory_resource/shared_resource.h
@@ -187,7 +187,7 @@ struct shared_resource
   //! @param __bytes The size in bytes of the allocation.
   //! @param __alignment The requested alignment of the allocation.
   //! @return Pointer to the newly allocated memory
-  [[nodiscard]] void* allocate_sync(size_t __bytes, size_t __alignment = alignof(::cuda::std::max_align_t))
+  [[nodiscard]] void* allocate_sync(size_t __bytes, size_t __alignment = ::cuda::mr::default_cuda_malloc_alignment)
   {
     return __control_block->__resource.allocate_sync(__bytes, __alignment);
   }
@@ -196,7 +196,8 @@ struct shared_resource
   //! @param __ptr Pointer to be deallocated. Must have been allocated through a call to `allocate` or `allocate_sync`
   //! @param __bytes The number of bytes that was passed to the allocation call that returned \p __ptr.
   //! @param __alignment The alignment that was passed to the allocation call that returned \p __ptr.
-  void deallocate_sync(void* __ptr, size_t __bytes, size_t __alignment = alignof(::cuda::std::max_align_t)) noexcept
+  void
+  deallocate_sync(void* __ptr, size_t __bytes, size_t __alignment = ::cuda::mr::default_cuda_malloc_alignment) noexcept
   {
     __control_block->__resource.deallocate_sync(__ptr, __bytes, __alignment);
   }
@@ -212,7 +213,8 @@ struct shared_resource
   //! operation has completed.
   _CCCL_TEMPLATE(class _ThisResource = _Resource)
   _CCCL_REQUIRES(::cuda::mr::resource<_ThisResource>)
-  [[nodiscard]] void* allocate(::cuda::stream_ref __stream, size_t __bytes, size_t __alignment)
+  [[nodiscard]] void*
+  allocate(::cuda::stream_ref __stream, size_t __bytes, size_t __alignment = ::cuda::mr::default_cuda_malloc_alignment)
   {
     return this->__control_block->__resource.allocate(__stream, __bytes, __alignment);
   }
@@ -228,7 +230,10 @@ struct shared_resource
   //! operation has completed.
   _CCCL_TEMPLATE(class _ThisResource = _Resource)
   _CCCL_REQUIRES(::cuda::mr::resource<_ThisResource>)
-  void deallocate(::cuda::stream_ref __stream, void* __ptr, size_t __bytes, size_t __alignment) noexcept
+  void deallocate(::cuda::stream_ref __stream,
+                  void* __ptr,
+                  size_t __bytes,
+                  size_t __alignment = ::cuda::mr::default_cuda_malloc_alignment) noexcept
   {
     this->__control_block->__resource.deallocate(__stream, __ptr, __bytes, __alignment);
   }
--- a/libcudacxx/include/cuda/__memory_resource/shared_resource.h
+++ b/libcudacxx/include/cuda/__memory_resource/shared_resource.h
@@ -160,20 +160,44 @@ struct shared_resource
   //! @param __bytes The size in bytes of the allocation.
   //! @param __alignment The requested alignment of the allocation.
   //! @return Pointer to the newly allocated memory
-  [[nodiscard]] _CCCL_HOST_API void*
-  allocate_sync(size_t __bytes, size_t __alignment = alignof(::cuda::std::max_align_t))
+  [[nodiscard]] _CCCL_HOST_API void* allocate_sync(size_t __bytes, size_t __alignment)
   {
     return __block_.__payload().allocate_sync(__bytes, __alignment);
+  }
+
+  //! @brief Allocate memory of size at least \p __bytes using the stored resource with
+  //! `alignof(::cuda::std::max_align_t)` as the alignment.
+  //! @param __bytes The size in bytes of the allocation.
+  //! @return Pointer to the newly allocated memory
+  //! @deprecated Specify an explicit alignment argument. A wrapper cannot know the alignment
+  //! requirements of the resource it holds.
+  CCCL_DEPRECATED_BECAUSE("Specify an explicit alignment argument. The default alignment will be removed in a future "
+                          "release.") _CCCL_HOST_API void*
+  allocate_sync(size_t __bytes)
+  {
+    return allocate_sync(__bytes, alignof(::cuda::std::max_align_t));
   }
 
   //! @brief Deallocate memory pointed to by \p __ptr using the stored resource.
   //! @param __ptr Pointer to be deallocated. Must have been allocated through a call to `allocate` or `allocate_sync`
   //! @param __bytes The number of bytes that was passed to the allocation call that returned \p __ptr.
   //! @param __alignment The alignment that was passed to the allocation call that returned \p __ptr.
-  _CCCL_HOST_API void
-  deallocate_sync(void* __ptr, size_t __bytes, size_t __alignment = alignof(::cuda::std::max_align_t)) noexcept
+  _CCCL_HOST_API void deallocate_sync(void* __ptr, size_t __bytes, size_t __alignment) noexcept
   {
     __block_.__payload().deallocate_sync(__ptr, __bytes, __alignment);
+  }
+
+  //! @brief Deallocate memory pointed to by \p __ptr using the stored resource with
+  //! `alignof(::cuda::std::max_align_t)` as the alignment.
+  //! @param __ptr Pointer to be deallocated. Must have been allocated through a call to `allocate` or `allocate_sync`
+  //! @param __bytes The number of bytes that was passed to the allocation call that returned \p __ptr.
+  //! @deprecated Specify an explicit alignment argument. A wrapper cannot know the alignment
+  //! requirements of the resource it holds.
+  CCCL_DEPRECATED_BECAUSE("Specify an explicit alignment argument. The default alignment will be removed in a future "
+                          "release.") _CCCL_HOST_API void
+  deallocate_sync(void* __ptr, size_t __bytes) noexcept
+  {
+    deallocate_sync(__ptr, __bytes, alignof(::cuda::std::max_align_t));
   }
 
   //! @brief Enqueues an allocation of memory of size at least \p __bytes using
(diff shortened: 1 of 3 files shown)


edenfunf:
@griwes  Hi, could you please take a look at this PR when you have time?

This fixes the default alignment used by deprecated no-alignment overloads
in type-erased memory resource wrappers.

Currently, type-erased wrappers (e.g. `any_resource`, `resource_ref`, `shared_resource`)
default to `alignof(::cuda::std::max_align_t)` (16 bytes), while concrete resources
(e.g. `device_memory_pool`) use `::cuda::mr::default_cuda_malloc_alignment` (256 bytes).

This creates a behavioral mismatch and can lead to under-aligned allocations
when using the type-erased interfaces with CUDA APIs expecting 256-byte alignment.

This PR:
- updates all deprecated no-alignment overloads to use `default_cuda_malloc_alignment`
- aligns behavior between concrete and type-erased resources
- applies consistent defaults across sync/async wrappers

The change only affects deprecated overloads and default parameters.

Tested:
- Compiles cleanly (CUDA 12.9, MSVC 19.50, sm_89)

Fixes #8063

Would really appreciate your feedback, thanks!

edenfunf:
@griwes Hi, could you please take a look at this PR when you have time?

This fixes the default alignment used by deprecated no-alignment overloads
in type-erased memory resource wrappers.

Currently, type-erased wrappers (e.g. any_resource, resource_ref, shared_resource)
default to alignof(::cuda::std::max_align_t) (16 bytes), while concrete resources
(e.g. device_memory_pool) use ::cuda::mr::default_cuda_malloc_alignment (256 bytes).

This creates a behavioral mismatch and can lead to under-aligned allocations
when using the type-erased interfaces with CUDA APIs expecting 256-byte alignment.

This PR:

updates all deprecated no-alignment overloads to use default_cuda_malloc_alignment
aligns behavior between concrete and type-erased resources
applies consistent defaults across sync/async wrappers
The change only affects deprecated overloads and default parameters.

Tested:

Compiles cleanly (CUDA 12.9, MSVC 19.50, sm_89)
Fixes https://github.com/NVIDIA/cccl/issues/8063

Would really appreciate your feedback, thanks!

libcudacxx/include/cuda/__memory_resource/shared_resource.h
```diff
@@ -187,7 +187,8 @@ struct shared_resource
   //! operation has completed.
   _CCCL_TEMPLATE(class _ThisResource = _Resource)
   _CCCL_REQUIRES(::cuda::mr::resource<_ThisResource>)
-  [[nodiscard]] _CCCL_HOST_API void* allocate(::cuda::stream_ref __stream, size_t __bytes, size_t __alignment)
+  [[nodiscard]] _CCCL_HOST_API void*
+  allocate(::cuda::stream_ref __stream, size_t __bytes, size_t __alignment = ::cuda::mr::default_cuda_malloc_alignment)
```
Jacobfaib: This is an API change. We must not change the default here.

libcudacxx/include/cuda/__memory_resource/shared_resource.h
```diff
@@ -203,7 +204,11 @@ struct shared_resource
   //! operation has completed.
   _CCCL_TEMPLATE(class _ThisResource = _Resource)
   _CCCL_REQUIRES(::cuda::mr::resource<_ThisResource>)
-  _CCCL_HOST_API void deallocate(::cuda::stream_ref __stream, void* __ptr, size_t __bytes, size_t __alignment) noexcept
+  _CCCL_HOST_API void deallocate(
+    ::cuda::stream_ref __stream,
+    void* __ptr,
+    size_t __bytes,
+    size_t __alignment = ::cuda::mr::default_cuda_malloc_alignment) noexcept
```
Jacobfaib: Likewise, the default argument is an API change (and below)

Jacobfaib (changes requested):
I agree with the premise of the bug, but not sure this is the best way to solve it. `any_resource` is well... any resource at all, not necessarily a CUDA compatible memory resource. Enforcing CUDA alignment on it is probably the wrong angle.

I believe what @pciolkosz will say is that the user should hold a `device_memory_resource` if they want to have this alignment (but then `cuda::buffer` doesn't and potentially suffers the same problem here). The alternative is that the resources have a `virtual size_t alignment() const` member function which they use to populate the default?

pciolkosz:
I think default alignment might have been a mistake in general. Only a small subset of APIs should call `allocate/deallocate` directly and within that small subset even smaller subset (or nobody?) should skip the alignment argument. I remember there was a discussion that wrapping the resources with the type erased wrappers changes the default alignment and we concluded that wrappers in general should not have a default, because its hard to say what the default is on the underlying resource. We could expose a trait like `resource::default_alignment`, but as I said, I think the use case is so small I don't think we should care. So I think going forward we should just not provide any defaults for alignment arguments in any resource wrappers and that should be fine. Specific resources can provide defaults, since that is safe and well defined

miscco:
> So I think going forward we should just not provide any defaults for alignment arguments in any resource wrappers and that should be fine. Specific resources can provide defaults, since that is safe and well defined

Yeah I believe the original intention was to be a drop in replacement, without any code changes which probably was not the right call. 

We gravitate towards explicit APIs for a reason and this is a good example why.

libcudacxx/include/cuda/__memory_resource/shared_resource.h
```diff
@@ -160,22 +160,46 @@ struct shared_resource
...
+  //! @param __bytes The size in bytes of the allocation.
+  //! @return Pointer to the newly allocated memory
+  //! @deprecated Specify an explicit alignment argument. A wrapper cannot know the alignment
+  //! requirements of the resource it holds.
+  CCCL_DEPRECATED_BECAUSE("Specify an explicit alignment argument. The default alignment will be removed in a future "
+                          "release.") _CCCL_HOST_API void*
+  allocate_sync(size_t __bytes)
+  {
+    return allocate_sync(__bytes, alignof(::cuda::std::max_align_t));
+  }
```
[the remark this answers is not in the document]
edenfunf: Intentional, see the discussion above. These are deprecated and kept behavior-compatible until removal.

edenfunf:
Thanks all. Rewritten accordingly.

- `any_resource.h` is back to `main`.
- After #8116, `shared_resource::allocate_sync` / `deallocate_sync` were the only
  wrapper entry points still picking a value. They are now deprecated the same
  way and still forward `alignof(max_align_t)`, so no behavior change.
- Left #8063 as `Refs` since nothing changes alignment today. Happy to follow up
  with the removal.

Local run (MSVC 19.50, CUDA 13.3, sm_120):

- `cuda/memory_resource` lit suite 23/23
- `shared_resource`, `synchronous_adapter` and `buffer` C2H tests pass
- untested on GCC/Clang

pciolkosz (approved):
Looks good, thanks

Jacobfaib (approved):

Jacobfaib:
/ok to test [8e1cb27](https://github.com/NVIDIA/cccl/pull/8451/commits/8e1cb27d62f0f6b0f1f9b898b13d91222c04f1cf)

Issue #3701, opened 2025-02-05
Track future deprecations and breaking changes for CCCL 4.0


bernhardmgruber:
This issue tracks entities in CCCL that we want to deprecate in the future. Those should be addressed during the CCCL 3.x cycle, with removal at a later stage. Feel free to add items to this list.

**General**

- [ ] Any API retuning an error code should be `[[nodiscard]]`

**CUB**:

- [x] #4165
- [ ] `WarpReduce` should use/return accumulator type
- [ ] `WarpReduce` drop edge cases (to discuss), e.g. dynamic num. items with multiple logical warps
- [ ] Drop `CUB_RUNTIME_FUNCTION` and replace it by `_CCCL_CDP_API`. Drop `CUB_DISABLE_CDP` and `CUB_RDC_ENABLED`
- [ ] Consider replacing `cub::DeviceScan::InclusiveScanInit` by a new overload of `InclusiveScan`. We didn't do this initially to not break any users of the other overloads and to not add the initial value after the stream at the end.
- [ ] Assuming the public tuning API has landed, move `cub/device/dispatch` into `cub/device/detail/dispatch`. There is no need anymore for anyone to include a dispatch header or anything below it.
- [ ] #6609
- [ ] Move non-public headers and directories to `cub/detail`, for example `cub/device/dispatch/kernel` and `cub/device/dispatch/tuning`.

**Thrust**:

- [ ] #3862
- [ ] #34
- [ ] #3993
- [ ] #5837
- [ ] #6637
- [ ] `thrust::[make_][integer|index]_sequence` in favor of libcu++
- [ ] Maybe the entire error code framework `thrust/thrust/system/detail/error_*`
- [ ] Maybe `thrust::system::cuda::errc`
- [ ] `thrust::mr::validator` (move to detail)
- [ ] `thrust/detail/preprocessor.h` and `THRUST_PP*` macros
- [ ] `thrust::tuple_for_each`
- [ ] parts of thrust `memory_algorithms.h`
- [ ] thrust `event_error.h`
- [ ] `thrust::has_difference_type` in `advance.inl` (move to detail)
- [ ] Review and deprecate macros from `thrust/thrust/system/cuda/config.h`

**libcu++**

- [ ] #3477
- [ ] Remove `<cuda/discard_memory>`, deprecated in #5672
- [ ] Remove `<cuda/stream_ref>`, deprecated in #6266
- [ ] #6225
[...]

PR #8452, opened 2026-04-15
modernize-use-nodiscard

Jacobfaib:
## Description

closes 

## Checklist

### DIFF
--- a/cudax/examples/vector.cuh
+++ b/cudax/examples/vector.cuh
@@ -107,7 +107,7 @@ private:
       __v_.sync_device_to_host(__str_, _Kind);
     }
 
-    ::cuda::std::span<_Ty> transformed_argument() const
+    [[nodiscard]] ::cuda::std::span<_Ty> transformed_argument() const
     {
       return {__v_.__d_.data().get(), __v_.__d_.size()};
     }
--- a/thrust/examples/repeated_range.cu
+++ b/thrust/examples/repeated_range.cu
@@ -49,12 +49,12 @@ public:
       , repeats(repeats)
   {}
 
-  iterator begin() const
+  [[nodiscard]] iterator begin() const
   {
     return PermutationIterator(first, TransformIterator(CountingIterator(0), repeat_functor(repeats)));
   }
 
-  iterator end() const
+  [[nodiscard]] iterator end() const
   {
     return begin() + repeats * (last - first);
   }
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -28,7 +28,6 @@ Checks:
   - '-modernize-use-using'
   - '-modernize-use-integer-sign-comparison'
   - '-modernize-concat-nested-namespaces'
-  - '-modernize-use-nodiscard'
   - '-modernize-use-default-member-init'
   - '-modernize-use-equals-default'
   - '-modernize-use-bool-literals'
--- a/c/parallel/test/algorithm_execution.h
+++ b/c/parallel/test/algorithm_execution.h
@@ -55,27 +55,27 @@ public:
     return singleton;
   }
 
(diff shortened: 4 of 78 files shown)


fbusato (approved):
looks good. It is important to not add `[[nodiscard]]` to member function or free-function that can change internal or parameter states

NaderAlAwar (approved):

cudax/examples/vector.cuh
```diff
@@ -107,7 +107,7 @@ private:
       __v_.sync_device_to_host(__str_, _Kind);
     }
 
-    ::cuda::std::span<_Ty> transformed_argument() const
+    [[nodiscard]] ::cuda::std::span<_Ty> transformed_argument() const
```
davebayer, line 110: Do we want the attribute even in the example? I'm not sure whether it's desired

thrust/examples/repeated_range.cu
```diff
@@ -49,12 +49,12 @@ public:
...
   {}
 
-  iterator begin() const
+  [[nodiscard]] iterator begin() const
   {
     return PermutationIterator(first, TransformIterator(CountingIterator(0), repeat_functor(repeats)));
   }
 
-  iterator end() const
+  [[nodiscard]] iterator end() const
```
davebayer, line 57: Same

cudax/examples/vector.cuh
fbusato: wait, do we have fully qualified names in examples ?!

Jacobfaib: I would be fiddly to disable this just for examples. I guess it's not the _end_ of the world on the grounds that `[[nodiscard]]` is good hygiene and that our examples should proselytize this when possible, but if folks are strongly against it I can go about trying to disable it.

But beware it may end up with `// NOLINT`, which IMHO is the worse option for examples.

davebayer: I'm probably fine with the change, just wanted to have a discussion about it :)

bernhardmgruber:
I think @gevtushenko concluded in the past that adding `[[nodiscard]]` to any public API is a source breaking change for users compiling with `-Werror`. At least that's the reason we are not adding nodiscard to CUB APIs until CCCL 4.0. This PR is adding nodiscard to public API functions. I would like to have @gevtushenko's opinion on the proposed changes.

miscco (approved):

gevtushenko:
> I think @gevtushenko concluded in the past that adding `[[nodiscard]]` to any public API is a source breaking change for users compiling with `-Werror`. At least that's the reason we are not adding nodiscard to CUB APIs until CCCL 4.0. This PR is adding nodiscard to public API functions. I would like to have @gevtushenko's opinion on the proposed changes.

Thanks for pushing on this. I agree `[[nodiscard]]` is a good hygiene improvement and I'd like to see more of it in the codebase.

That said, we cannot retrofit it onto existing interfaces. Adding `[[nodiscard]]` to an already-released function is a source-breaking change for any downstream user compiling with warnings-as-errors, which is a pretty common setup. unfortunately, not checking the return value of CUB APIs is fairly common. So retrofitting `[[nodiscard]]` onto those APIs would very likely break someone on a minor version update. 

Let's adopt `[[nodiscard]]` as the default for *new* interfaces going forward, and leave existing declarations as is until CCCL 4.0. It might be a good idea to start a tracker of what breaking changes we want to make and include `[[nodiscard]]` there.

Jacobfaib:
Tabling until CCCL 4.0

bernhardmgruber:
> It might be a good idea to start a tracker of what breaking changes we want to make and include `[[nodiscard]]` there.

I am a year ahead: #3701

PR #8454, opened 2026-04-15
modernize-use-equals-default

Jacobfaib:
## Description

closes 

## Checklist

### DIFF
--- a/c/parallel/test/test_radix_sort.cpp
+++ b/c/parallel/test/test_radix_sort.cpp
@@ -37,7 +37,7 @@ struct TestParameters
   static constexpr bool m_descending     = descending;
   static constexpr bool m_overwrite_okay = overwrite_okay;
 
-  constexpr TestParameters() {}
+  constexpr TestParameters() = default;
 
   bool is_descending() const
   {
--- a/cub/cub/util_type.cuh
+++ b/cub/cub/util_type.cuh
@@ -701,8 +701,8 @@ struct KeyValuePair
   using Key   = _Key; ///< Key data type
   using Value = _Value; ///< Value data type
 
-  Key key{}; ///< Item key
-  Value value{}; ///< Item value
+  Key key; ///< Item key
+  Value value; ///< Item value
 
   /// Constructor
   _CCCL_FORCEINLINE KeyValuePair() = default;
--- a/cub/cub/util_type.cuh
+++ b/cub/cub/util_type.cuh
@@ -701,11 +701,11 @@ struct KeyValuePair
   using Key   = _Key; ///< Key data type
   using Value = _Value; ///< Value data type
 
-  Key key; ///< Item key
-  Value value; ///< Item value
+  Key key{}; ///< Item key
+  Value value{}; ///< Item value
 
   /// Constructor
-  _CCCL_HOST_DEVICE _CCCL_FORCEINLINE KeyValuePair() {}
+  _CCCL_FORCEINLINE KeyValuePair() = default;
 
   /// Constructor
   _CCCL_HOST_DEVICE _CCCL_FORCEINLINE KeyValuePair(Key const& key, Value const& value)
--- a/libcudacxx/include/cuda/std/__tuple_dir/tuple_leaf.h
+++ b/libcudacxx/include/cuda/std/__tuple_dir/tuple_leaf.h
@@ -105,8 +105,11 @@ class __tuple_leaf<_Ip, _Hp, __tuple_leaf_specialization::__default>
 #endif // !_CCCL_BUILTIN_REFERENCE_CONSTRUCTS_FROM_TEMPORARY
 
 public:
+  // The compiler-generated constructor would not value-initialize trivial types, but the
+  // standard requires that we do.
   _CCCL_EXEC_CHECK_DISABLE
-  _CCCL_API constexpr __tuple_leaf() noexcept(is_nothrow_default_constructible_v<_Hp>)
+  _CCCL_API constexpr __tuple_leaf() // NOLINT(modernize-use-equals-default)
+    noexcept(is_nothrow_default_constructible_v<_Hp>)
       : __value_()
   {}
 
@@ -227,14 +230,19 @@ public:
   _CCCL_EXEC_CHECK_DISABLE
   __tuple_leaf(__tuple_leaf&& __t) = default;
 
+  // Do not use = default here. The value type may be a reference, in which case the defaulted
+  // assignment constructor is implicitly deleted (a quirk in the C++ spec meant).
   _CCCL_EXEC_CHECK_DISABLE
-  _CCCL_API constexpr __tuple_leaf& operator=(const __tuple_leaf& __t) noexcept
+  _CCCL_API constexpr __tuple_leaf& operator=(const __tuple_leaf& __t) noexcept // NOLINT(modernize-use-equals-default)
   {
     __value_ = __t.__value_;
     return *this;
   }
+
+  // Do not use = default here. The value type may be a reference, in which case the defaulted
+  // assignment constructor is implicitly deleted (a quirk in the C++ spec meant).
   _CCCL_EXEC_CHECK_DISABLE
-  _CCCL_API constexpr __tuple_leaf& operator=(__tuple_leaf&& __t) noexcept
+  _CCCL_API constexpr __tuple_leaf& operator=(__tuple_leaf&& __t) noexcept // NOLINT(modernize-use-equals-default)
   {
     __value_ = ::cuda::std::move(__t.__value_);
     return *this;
@@ -267,8 +275,11 @@ template <size_t _Ip, class _Hp>
 class __tuple_leaf<_Ip, _Hp, __tuple_leaf_specialization::__empty_non_final> : private remove_const_t<_Hp>
 {
 public:
+  // The compiler-generated constructor would not value-initialize trivial types, but the
+  // standard requires that we do.
   _CCCL_EXEC_CHECK_DISABLE
-  _CCCL_API constexpr __tuple_leaf() noexcept(is_nothrow_default_constructible_v<_Hp>)
+  _CCCL_API constexpr __tuple_leaf() // NOLINT(modernize-use-equals-default)
+    noexcept(is_nothrow_default_constructible_v<_Hp>)
       : _Hp()
   {}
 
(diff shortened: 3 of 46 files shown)


c/parallel/test/test_radix_sort.cpp
```diff
@@ -37,7 +37,7 @@ struct TestParameters
   static constexpr bool m_descending     = descending;
   static constexpr bool m_overwrite_okay = overwrite_okay;
 
-  constexpr TestParameters() {}
+  constexpr TestParameters() = default;
```
fbusato, line 40: this actually change the behavior and could affect the performance. `= default` means default/zero-initialization of data members. 
Did you already check if the PR affects this behavior?

fbusato (changes requested):

c/parallel/test/test_radix_sort.cpp
Jacobfaib: For the purpose of implicitly initializing members, `{}` and `= default` are identical https://godbolt.org/z/rr4hndsrs.

davebayer: I think it just changes the result of `cuda::std::is_trivially_default_constructible`, other than that, I think the behaviour should be identical

davebayer: Unless it's somehow important for the test, I think this change is fine

libcudacxx/include/cuda/std/__tuple_dir/tuple_leaf.h
```diff
@@ -106,9 +106,7 @@ class __tuple_leaf<_Ip, _Hp, __tuple_leaf_specialization::__default>
 
 public:
   _CCCL_EXEC_CHECK_DISABLE
-  _CCCL_API constexpr __tuple_leaf() noexcept(is_nothrow_default_constructible_v<_Hp>)
-      : __value_()
-  {}
+  _CCCL_HIDE_FROM_ABI constexpr __tuple_leaf() noexcept(is_nothrow_default_constructible_v<_Hp>) = default;
```
miscco: This is wrong, we *must* value initialize the element in tuple_leaf

libcudacxx/include/cuda/std/__tuple_dir/tuple_leaf.h
```diff
@@ -268,9 +271,7 @@ class __tuple_leaf<_Ip, _Hp, __tuple_leaf_specialization::__empty_non_final> : p
 {
 public:
   _CCCL_EXEC_CHECK_DISABLE
-  _CCCL_API constexpr __tuple_leaf() noexcept(is_nothrow_default_constructible_v<_Hp>)
-      : _Hp()
-  {}
+  _CCCL_HIDE_FROM_ABI constexpr __tuple_leaf() noexcept(is_nothrow_default_constructible_v<_Hp>) = default;
```
miscco: Ditto, we *must* value initialize even if the class is empty

libcudacxx/include/cuda/std/__tuple_dir/tuple_leaf.h
```diff
@@ -370,9 +371,7 @@ struct _CCCL_DECLSPEC_EMPTY_BASES __tuple_impl<__tuple_indices<_Indx...>, _Tp...
 {
   using _Constraints = __tuple_constraints<_Tp...>;
 
-  _CCCL_API constexpr __tuple_impl() noexcept(_Constraints::__nothrow_default_constructible)
-      : __tuple_leaf<_Indx, _Tp>()...
-  {}
+  _CCCL_HIDE_FROM_ABI constexpr __tuple_impl() noexcept(_Constraints::__nothrow_default_constructible) = default;
```
miscco: Needs to be reverted

miscco (changes requested):

libcudacxx/include/cuda/std/__tuple_dir/tuple_leaf.h
davebayer: Why don't we default initialize the value instead
```cpp
_Hp __value_{};
```
Then we could `= default` the constructor

miscco: This is icky, my understanding is that there is *no* requirement on `_Hp` to be default constructible

So when we write `_Hp __value_{};` it *must* be default constructible. However, we SFINAE all uses of the default constructor out in the `tuple` implementation, so we never use the default constructor and all is good.

miscco: To be sure, here is the difference:
With default member initializer
https://godbolt.org/z/9aqodYYne
Without:
https://godbolt.org/z/ddo8zs6T8

Jacobfaib: TIL, I was under the impression that compilers would not evaluate member initializers until a constructor is called (implicitly or explicitly), but I guess this is not the case? Would be interested to hear from language lawyers why this distinction exists.

miscco: Dont you love this language?

Jacobfaib: Ah wait, does this apply? I amended your second example https://godbolt.org/z/vYEqqj9oT. `tuple_leaf` can use `= default` so long as we don't initialize the member.

miscco: No because them it does not initialize the member if it is trivial

miscco (approved):

fbusato (approved):

cub/cub/util_type.cuh
```diff
@@ -708,7 +708,7 @@ struct KeyValuePair
   Value value; ///< Item value
 
   /// Constructor
-  _CCCL_HOST_DEVICE _CCCL_FORCEINLINE KeyValuePair() {}
+  _CCCL_FORCEINLINE KeyValuePair() = default;
```
miscco, line 708: This is also a functional change. This will omit initialization of the members if they are trivial

miscco (changes requested):

cub/cub/util_type.cuh
Jacobfaib: Fixed

miscco (approved):

cub/cub/util_type.cuh
```diff
@@ -704,11 +704,11 @@ struct KeyValuePair
   using Key   = _Key; ///< Key data type
   using Value = _Value; ///< Value data type
 
-  Key key; ///< Item key
-  Value value; ///< Item value
+  Key key{}; ///< Item key
+  Value value{}; ///< Item value
```
bernhardmgruber: Critical: this performs value initialization for trivial types, which requires additional code to be generated and will thus likely affect performance. Please revert or benchmark all algorithms that involve `KeyValuePair`.

bernhardmgruber (changes requested):

cub/cub/util_type.cuh
Jacobfaib: Sorry was a result of https://github.com/NVIDIA/cccl/pull/8454/changes#r3109162186 which was actually incorrect. Will revert

cub/cub/util_type.cuh
Jacobfaib: Should have checked on godbolt but this is actually incorrect. The original constructor does not initialize the members either https://godbolt.org/z/PTWMxYTj5, so reverting this.

thrust/examples/uninitialized_vector.cu
```diff
@@ -47,6 +47,11 @@ struct uninitialized_allocator : thrust::device_allocator<T>
   }
 };
 
+template <typename T>
+uninitialized_allocator<T>::uninitialized_allocator() = default;
+template <typename T>
+uninitialized_allocator<T>::~uninitialized_allocator() = default;
+
```
bernhardmgruber, line 54: Q: Why do we need an out of line definition of a defaulted special member here? Why can't we just `= default` them right at their declaration?

thrust/testing/allocator.cu
```diff
@@ -121,6 +121,12 @@ struct my_allocator_with_custom_destroy
   std::allocator<T> use_me_to_alloc;
 };
 
+template <typename T>
+my_allocator_with_custom_destroy<T>::my_allocator_with_custom_destroy() = default;
+
+template <typename T>
+my_allocator_with_custom_destroy<T>::~my_allocator_with_custom_destroy() = default;
+
```
bernhardmgruber, line 129: Same here.

thrust/testing/allocator.cu
```diff
@@ -170,6 +176,12 @@ struct my_minimal_allocator
   std::allocator<T> use_me_to_alloc;
 };
 
+template <typename T>
+my_minimal_allocator<T>::my_minimal_allocator() = default;
+
+template <typename T>
+my_minimal_allocator<T>::~my_minimal_allocator() = default;
+
```
bernhardmgruber, line 184: And here

thrust/testing/vector_allocators.cu
```diff
@@ -15,7 +15,7 @@ public:
       : state(i)
   {}
 
-  ~stateful_allocator() {}
+  _CCCL_HOST ~stateful_allocator() {} // NOLINT(modernize-use-equals-default)
```
bernhardmgruber, line 18: Q: Why do we need `_CCCL_HOST` here? I would omit it.

thrust/examples/uninitialized_vector.cu
Jacobfaib: This is the way to declare a defaulted function that is `__host__` or `__device__` only. `nvcc` does not honor
```c++
class Foo {
  __host__ Foo() = default;
};
```
and warns that it will ignore `__host__` (conceivably the `= default` bit is "stronger"??)

bernhardmgruber: Why do we need to make the function host or device only?

thrust/testing/vector_allocators.cu
Jacobfaib: Otherwise we get many many execution space warnings about calling a host-only function from a host-device function due to the destruction of the `std::allocator` member. My best guess is that nvcc changes the way in which member object destructors are emitting for functions that are `= default` vs `{}`. Ultimately they call the same code, but it seems to change whether nvcc deduces them as host-only or host-device. If you specify as {} for host-only, then nvcc will not warn. But if you `= default` for host-only (and properly do it), then nvcc still tries to call host-device versions. It's very strange.

thrust/examples/uninitialized_vector.cu
Jacobfaib: Not sure, but that's the way they were before. I didn't want to change them.

bernhardmgruber (approved):

PR #8455, opened 2026-04-15
modernize-use-bool-literals

Jacobfaib:
## Description

closes 

## Checklist

### DIFF
--- a/libcudacxx/include/cuda/std/__limits/numeric_limits.h
+++ b/libcudacxx/include/cuda/std/__limits/numeric_limits.h
@@ -268,11 +268,11 @@ public:
   static constexpr int radix       = 2;
   _CCCL_API static constexpr type epsilon() noexcept
   {
-    return type(0);
+    return false;
   }
   _CCCL_API static constexpr type round_error() noexcept
   {
-    return type(0);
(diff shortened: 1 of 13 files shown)


fbusato (approved):

bernhardmgruber (approved):

libcudacxx/include/cuda/std/__limits/numeric_limits.h
```diff
@@ -268,11 +268,11 @@ class __numeric_limits_impl<bool, __numeric_limits_type::__bool>
   static constexpr int radix       = 2;
   _CCCL_API static constexpr type epsilon() noexcept
   {
-    return type(0);
+    return false;
```
miscco, line 271: This is not correct, please revert. Those are integer types

Jacobfaib: Is it? This in the bool specialization for numeric limits. Type is bool, so type(0) is equal to false

miscco: Sorry I thought that was the integer specialization

PR #8458, opened 2026-04-15
modernize-return-braced-init-list

Jacobfaib:
## Description

closes 

## Checklist

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -38,7 +38,6 @@ Checks:
   - '-modernize-use-bool-literals'
   - '-modernize-type-traits'
   - '-modernize-use-nullptr'
-  - '-modernize-return-braced-init-list'
   - '-modernize-use-auto'
   - '-modernize-redundant-void-arg'
   - '-modernize-use-equals-delete'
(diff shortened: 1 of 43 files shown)


fbusato (commented):
don't have a strong opinion on that, slightly in favor. My concern is that I'm 100% that all compilers accept the new style

miscco (changes requested):
I am not in favor of this. IN our heavily templated code this might trigger especially the older compilers like GCC-7 which are not fully tested in PR CI

I am all for using `{}` for constructors, but we should retain the type

Jacobfaib:
> I am not in favor of this. IN our heavily templated code this might trigger especially the older compilers like GCC-7 which are not fully tested in PR CI
>
> I am all for using {} for constructors, but we should retain the type

`clang-tidy` will only suggest replacement with `{}` when the return-type is unambiguously clear, i.e. when it is spelled out explicitly in the return type. If you take a look at the changes in this PR, you'll note that it does not specify a replacement for any templated types, only when the type is fully resolved.

Is there a mechanism where we can test older compilers in CI?

Jacobfaib:
Since there seems to be no concerted votes for this feature, I will go ahead and close this PR and leave the check disabled.

PR #8459, opened 2026-04-15
modernize-use-auto

Jacobfaib:
## Description

closes 

## Checklist

### DIFF
--- a/cub/benchmarks/bench/histogram/even.cu
+++ b/cub/benchmarks/bench/histogram/even.cu
@@ -58,7 +58,7 @@ static void even(nvbench::state& state, nvbench::type_list<SampleT, CounterT, Of
   }
 
   const SampleT lower_level = 0;
-  const SampleT upper_level = get_upper_level<SampleT>(num_bins, elements);
+  const auto upper_level    = get_upper_level<SampleT>(num_bins, elements);
 
   thrust::device_vector<SampleT> input = generate(elements, entropy, lower_level, upper_level);
   thrust::device_vector<CounterT> hist(num_bins);
@@ -70,7 +70,7 @@ static void even(nvbench::state& state, nvbench::type_list<SampleT, CounterT, Of
   std::size_t temp_storage_bytes{};
 
   cuda::std::bool_constant<sizeof(SampleT) == 1> is_byte_sample;
-  OffsetT num_row_pixels     = static_cast<OffsetT>(elements);
+  auto num_row_pixels        = static_cast<OffsetT>(elements);
   OffsetT num_rows           = 1;
   OffsetT row_stride_samples = num_row_pixels;
 
--- a/cub/benchmarks/bench/histogram/multi/even.cu
+++ b/cub/benchmarks/bench/histogram/multi/even.cu
@@ -60,7 +60,7 @@ static void even(nvbench::state& state, nvbench::type_list<SampleT, CounterT, Of
   }
 
   const SampleT lower_level_r = 0;
-  const SampleT upper_level_r = get_upper_level<SampleT>(num_bins, elements);
+  const auto upper_level_r    = get_upper_level<SampleT>(num_bins, elements);
   const SampleT lower_level_g = lower_level_r;
   const SampleT upper_level_g = upper_level_r;
   const SampleT lower_level_b = lower_level_g;
@@ -80,7 +80,7 @@ static void even(nvbench::state& state, nvbench::type_list<SampleT, CounterT, Of
   std::size_t temp_storage_bytes{};
 
   cuda::std::bool_constant<sizeof(SampleT) == 1> is_byte_sample;
-  OffsetT num_row_pixels     = static_cast<OffsetT>(elements);
+  auto num_row_pixels        = static_cast<OffsetT>(elements);
   OffsetT num_rows           = 1;
   OffsetT row_stride_samples = num_row_pixels;
 
--- a/cub/benchmarks/bench/histogram/multi/range.cu
+++ b/cub/benchmarks/bench/histogram/multi/range.cu
@@ -52,7 +52,7 @@ static void range(nvbench::state& state, nvbench::type_list<SampleT, CounterT, O
   const int num_levels_b = num_levels_g;
 
   const SampleT lower_level = 0;
-  const SampleT upper_level = get_upper_level<SampleT>(num_bins, elements);
+  const auto upper_level    = get_upper_level<SampleT>(num_bins, elements);
 
   SampleT step = (upper_level - lower_level) / num_bins;
   thrust::device_vector<SampleT> levels_r(num_bins + 1);
@@ -80,7 +80,7 @@ static void range(nvbench::state& state, nvbench::type_list<SampleT, CounterT, O
   std::size_t temp_storage_bytes{};
 
   cuda::std::bool_constant<sizeof(SampleT) == 1> is_byte_sample;
-  OffsetT num_row_pixels     = static_cast<OffsetT>(elements);
+  auto num_row_pixels        = static_cast<OffsetT>(elements);
   OffsetT num_rows           = 1;
   OffsetT row_stride_samples = num_row_pixels;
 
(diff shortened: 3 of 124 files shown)


fbusato (commented):
I'm generally in favor of that, but my feeling is that @gevtushenko and @miscco  could have a different opinion

Jacobfaib:
> I'm generally in favor of that, but my feeling is that @gevtushenko and @miscco could have a different opinion

To perhaps allay some fears here, `clang-tidy` will only suggest using `auto` when the type is already named on the same line:

- Casts: `Foo f = static_cast<Foo>(x)` -> `auto f = static_cast<Foo>(x)`
- Construction: `Foo f = Foo{...};` -> `auto f = Foo{...};`
- New expressions: `Foo *f = new Foo{...};` -> `auto *f = new Foo{...};`

Or specifically for standard library iterators:
```c++
std::vector<int>::iterator it = v.begin();
// becomes
auto it = v.begin();
```

cub/cub/block/block_radix_sort.cuh
```diff
@@ -387,7 +387,7 @@ private:
     ::cuda::std::bool_constant<KEYS_ONLY> is_keys_only,
     DecomposerT decomposer = {})
   {
-    bit_ordered_type(&unsigned_keys)[ItemsPerThread] = reinterpret_cast<bit_ordered_type(&)[ItemsPerThread]>(keys);
+    auto(&unsigned_keys)[ItemsPerThread] = reinterpret_cast<bit_ordered_type(&)[ItemsPerThread]>(keys);
```
bernhardmgruber, line 390: Remark: haven't seen a reinterpretation as a reference to an array in a long time

cub/cub/device/dispatch/dispatch_radix_sort.cuh
```diff
@@ -556,8 +556,8 @@ private:
     // for testing purposes, one portion is <= 2**28 elements
     const PortionOffsetT PORTION_SIZE = ((1 << 28) - 1) / ONESWEEP_TILE_ITEMS * ONESWEEP_TILE_ITEMS;
     int num_passes                    = ::cuda::ceil_div(end_bit - begin_bit, RADIX_BITS);
-    OffsetT num_portions              = static_cast<OffsetT>(::cuda::ceil_div(num_items, PORTION_SIZE));
-    PortionOffsetT max_num_blocks     = ::cuda::ceil_div(
+    auto num_portions                 = static_cast<OffsetT>(::cuda::ceil_div(num_items, PORTION_SIZE));
+    auto max_num_blocks               = ::cuda::ceil_div(
       static_cast<int>(::cuda::std::min(num_items, static_cast<OffsetT>(PORTION_SIZE))), ONESWEEP_TILE_ITEMS);
```
bernhardmgruber, line 561: This looks a bit like clang-tidy knows that `PortionOffsetT` is `int`. Can we reflect this in the cast?
```suggestion
    auto max_num_blocks               = ::cuda::ceil_div(
static_cast<PortionOffsetT>(::cuda::std::min(num_items, static_cast<OffsetT>(PORTION_SIZE))), ONESWEEP_TILE_ITEMS);
```

cub/examples/block/example_block_radix_sort.cu
```diff
@@ -140,9 +140,9 @@ void Test()
   constexpr int TILE_SIZE = BLOCK_THREADS * ITEMS_PER_THREAD;
 
   // Allocate host arrays
-  Key* h_in          = new Key[TILE_SIZE * g_grid_size];
-  Key* h_reference   = new Key[TILE_SIZE * g_grid_size];
-  clock_t* h_elapsed = new clock_t[g_grid_size];
+  Key* h_in        = new Key[TILE_SIZE * g_grid_size];
+  Key* h_reference = new Key[TILE_SIZE * g_grid_size];
```
bernhardmgruber: Q: Why no `auto` here?

```suggestion
  auto* h_in        = new Key[TILE_SIZE * g_grid_size];
  auto* h_reference = new Key[TILE_SIZE * g_grid_size];
```

cub/examples/device/example_device_partition_flagged.cu
```diff
@@ -134,9 +134,9 @@ int main(int argc, char** argv)
   CubDebugExit(args.DeviceInit());
 
   // Allocate host arrays
-  int* h_in              = new int[num_items];
-  int* h_reference       = new int[num_items];
-  unsigned char* h_flags = new unsigned char[num_items];
+  int* h_in        = new int[num_items];
+  int* h_reference = new int[num_items];
```
bernhardmgruber: Same here

bernhardmgruber (commented):
This PR is like 95% fine and I prefer the refactoring. I looked through a third or so.

cub/examples/block/example_block_radix_sort.cu
Jacobfaib: Sooo we are starting to run into one unfortunate quirck for `clang-tidy`... it is a tangled mess of heuristics. A lot of checks are quite expensive, and `clang-tidy` employs a vast corpus of heuristics to keep performance acceptable. My guess is this TU was too complicated/long and `modernize-use-auto` gave up trying to process it.

We will run into this a lot more frequently when we enable the Big Daddy `misc-const-correctness`. Touching a seemingly separate part of the code will trip some hidden lever in `clang-tidy` and make it start emitting missed diagnostics in another part.

cub/cub/device/dispatch/dispatch_radix_sort.cuh
Jacobfaib: Hmmm I am confused, the `int` cast already existed before, you want to change that to `PortionOffsetT` as well?

cub/benchmarks/bench/histogram/multi/even.cu
```diff
@@ -60,7 +60,7 @@ static void even(nvbench::state& state, nvbench::type_list<SampleT, CounterT, Of
   }
 
   const SampleT lower_level_r = 0;
-  const SampleT upper_level_r = get_upper_level<SampleT>(num_bins, elements);
+  const auto upper_level_r    = get_upper_level<SampleT>(num_bins, elements);
```
miscco, line 63: I am against this one, it breaks alignnment

cub/benchmarks/bench/histogram/multi/range.cu
```diff
@@ -52,7 +52,7 @@ static void range(nvbench::state& state, nvbench::type_list<SampleT, CounterT, O
   const int num_levels_b = num_levels_g;
 
   const SampleT lower_level = 0;
-  const SampleT upper_level = get_upper_level<SampleT>(num_bins, elements);
+  const auto upper_level    = get_upper_level<SampleT>(num_bins, elements);
```
miscco, line 55: Those also break alignment

cub/benchmarks/bench/histogram/even.cu
```diff
@@ -70,7 +70,7 @@ static void even(nvbench::state& state, nvbench::type_list<SampleT, CounterT, Of
   std::size_t temp_storage_bytes{};
 
   cuda::std::bool_constant<sizeof(SampleT) == 1> is_byte_sample;
-  OffsetT num_row_pixels     = static_cast<OffsetT>(elements);
+  auto num_row_pixels        = static_cast<OffsetT>(elements);
```
miscco, line 73: Ditto breaks alignment

miscco (commented):
I am really not sure about this. Auto is often usefull if we have a `static_cast` on the other side of an assignment

But in many cases this creates visual noise, when we have multiple variables and one of them suddenly becomes `auto`

Often its also not easy to keep the type in min for more complex code, so I am not sure whether its a win

cub/benchmarks/bench/histogram/multi/even.cu
Jacobfaib: What if I made them all `auto`? And
```c++
const SampleT lower_level_r = 0;
// to
const auto lower_level_r = SamepleT{0};
```
I can silence them with `// NOLINT` but I'd rather avoid that if there are other code changes we can make to satisfy the linter.

fbusato: we could enforce variable alignment 😁 https://clang.llvm.org/docs/ClangFormatStyleOptions.html#alignconsecutivedeclarations

Jacobfaib:
Since there seems to be no concerted votes for this feature, I will go ahead and close this PR and leave the check disabled.

PR #8462, opened 2026-04-15
modernize-pass-by-value

Jacobfaib:
## Description

closes 

## Checklist

### DIFF
--- a/thrust/thrust/mr/disjoint_pool.h
+++ b/thrust/thrust/mr/disjoint_pool.h
@@ -223,8 +223,8 @@ private:
 
   struct pool
   {
-    _CCCL_HOST pool(const pointer_vector& free)
-        : free_blocks(free)
+    _CCCL_HOST pool(pointer_vector free)
+        : free_blocks(::cuda::std::move(free))
         , previous_allocated_count(0)
     {}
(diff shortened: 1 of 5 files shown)


davebayer (approved):

thrust/thrust/mr/disjoint_pool.h
```diff
@@ -223,8 +223,8 @@ class disjoint_unsynchronized_pool_resource final
 
   struct pool
   {
-    _CCCL_HOST pool(const pointer_vector& free)
-        : free_blocks(free)
+    _CCCL_HOST pool(pointer_vector free)
+        : free_blocks(::cuda::std::move(free))
```
davebayer, line 227: Are we fine with this change? I think it should be fine

Jacobfaib: I don't see how it could change things. A copy is made either way (either in the ctor argument, or into the member from the const-ref).

davebayer: Yeah, I know, it's just that we are changing the signature of a public function. I think it should be fine, too, just wanted to point it out

PR #8465, opened 2026-04-15
modernize-use-override

Jacobfaib:
## Description

closes 

## Checklist

### DIFF
--- a/thrust/thrust/device_allocator.h
+++ b/thrust/thrust/device_allocator.h
@@ -67,13 +67,13 @@ public:
       : m_upstream(upstream)
   {}
 
-  [[nodiscard]] _CCCL_HOST virtual pointer
+  [[nodiscard]] _CCCL_HOST pointer
   do_allocate(std::size_t bytes, std::size_t alignment = THRUST_MR_DEFAULT_ALIGNMENT) override
   {
     return pointer(m_upstream->do_allocate(bytes, alignment).get());
   }
 
-  _CCCL_HOST virtual void do_deallocate(pointer p, std::size_t bytes, std::size_t alignment) override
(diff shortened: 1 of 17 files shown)


bernhardmgruber (approved):

fbusato (approved):

thrust/thrust/device_allocator.h
```diff
@@ -67,13 +67,13 @@ class device_ptr_memory_resource final : public thrust::mr::memory_resource<devi
...
 
-  [[nodiscard]] _CCCL_HOST virtual pointer
+  [[nodiscard]] _CCCL_HOST pointer
   do_allocate(std::size_t bytes, std::size_t alignment = THRUST_MR_DEFAULT_ALIGNMENT) override
   {
     return pointer(m_upstream->do_allocate(bytes, alignment).get());
   }
 
-  _CCCL_HOST virtual void do_deallocate(pointer p, std::size_t bytes, std::size_t alignment) override
+  _CCCL_HOST void do_deallocate(pointer p, std::size_t bytes, std::size_t alignment) override
```
davebayer, line 76: Removing `virtual` is breaking the current behaviour, isn't it? Or was this a bug?

davebayer (changes requested):
Removing `virtual` is breaking change in many cases. Are we fine with that?

thrust/thrust/device_allocator.h
Jacobfaib: No, `override` implies `virtual`. It actually makes the check stronger here, since for `override` the compiler first checks that you are overriding _something_. `virtual` on its own might lead to inadvertently hiding virtual functions in case the base changes.

davebayer: Whaaaat ?! I didn't know that! Thanks for clarification

davebayer (approved):

PR #8468, opened 2026-04-15
Remove hardcoded c++17 dialect for benchmarks

NaderAlAwar:
This changes benchmark-related CMake targets to use the user-configured `CMAKE_CXX_STANDARD` / `CMAKE_CUDA_STANDARD` instead of hardcoding `DIALECT 17`.

Before this change, several benchmark paths always forced c++17 even when the build was configured for C++20. Tests already inherit the configured dialect, so this makes benchmarks consistent with the rest of the repo.


oleksandr-pavlyk (approved):

bernhardmgruber (approved):

miscco (approved):

miscco:
pre-commit.ci autofix

PR #8471, opened 2026-04-15
Deprecate `thrust::integer_sequence` and friends

bernhardmgruber:


thrust/thrust/type_traits/integer_sequence.h
```diff
@@ -150,10 +158,11 @@ struct merge_and_renumber_reversed_integer_sequences_impl<integer_sequence<T, Is
  *  \see make_reversed_index_sequence
  */
 template <typename T, ::cuda::std::size_t N>
-using make_reversed_integer_sequence = typename detail::make_reversed_integer_sequence_impl<T, N>::type;
+using make_reversed_integer_sequence CCCL_DEPRECATED = typename detail::make_reversed_integer_sequence_impl<T, N>::type;
```
miscco: There is no equivalent for this in the standard and we are using it internally, that was the reason I did not rip it out yet

miscco (changes requested):

thrust/thrust/type_traits/integer_sequence.h
davebayer: Where? I haven't found any use

miscco: 🤔 I cannot find the uses anymore, could be that we refactored them out

miscco (approved):

thrust/thrust/type_traits/integer_sequence.h
bernhardmgruber: I refactored the last use away a couple days ago, which is why I made this PR now.

thrust/thrust/type_traits/integer_sequence.h
```diff
@@ -162,27 +171,32 @@ using make_reversed_integer_sequence = typename detail::make_reversed_integer_se
  *  \see make_reversed_index_sequence
  */
 template <::cuda::std::size_t N>
-using make_reversed_index_sequence = make_reversed_integer_sequence<::cuda::std::size_t, N>;
+using make_reversed_index_sequence
+  CCCL_DEPRECATED = typename detail::make_reversed_integer_sequence_impl<::cuda::std::size_t, N>::type;
```
davebayer: Do we want to add a message that no replacement is planned?

davebayer (approved):

thrust/thrust/type_traits/integer_sequence.h
bernhardmgruber: Fine

PR #8472, opened 2026-04-15
`cuda::std::simd` Iterators

fbusato:
## Description

Follow up PR of 

- https://github.com/NVIDIA/cccl/pull/8251

Implement `cuda::std::simd` iterator class and methods `[simd.iterator]`


libcudacxx/include/cuda/std/__simd/iterator.h
```diff
@@ -0,0 +1,264 @@
...
+// [simd.iterator], class template __simd_iterator
+template <typename _Vp>
+class __simd_iterator
+{
+  _Vp* __data_               = nullptr;
+  __simd_size_type __offset_ = 0;
+
+  _CCCL_API constexpr __simd_iterator(_Vp& __d, const __simd_size_type __off) noexcept
+      : __data_{::cuda::std::addressof(__d)}
+      , __offset_{__off}
```
miscco: Nitpick: We conventionally name equal things equal
```suggestion
  _CCCL_API constexpr __simd_iterator(_Vp& __data, const __simd_size_type __offset) noexcept
      : __data_{::cuda::std::addressof(__data)}
      , __offset_{__offset}
```

libcudacxx/include/cuda/std/__simd/iterator.h
```diff
@@ -0,0 +1,264 @@
...
+  _CCCL_TEMPLATE(typename _Up = remove_const_t<_Vp>)
+  _CCCL_REQUIRES(is_same_v<const _Up, _Vp>)
+  _CCCL_API constexpr __simd_iterator(const __simd_iterator<_Up>& __i) noexcept
+      : __data_{__i.__data_}
+      , __offset_{__i.__offset_}
+  {}
+
+  // the C++ specification does not require a noexcept for most operators, but the iterator only operates on pointer and
+  // integer type
+
```
miscco: We can just drop that. `noexcept` is QoI
```suggestion
```

libcudacxx/include/cuda/std/__simd/iterator.h
```diff
@@ -0,0 +1,264 @@
...
+  [[nodiscard]] _CCCL_API constexpr value_type operator*() const noexcept
+  {
+    _CCCL_ASSERT(__data_ != nullptr, "cuda::std::simd::__simd_iterator: data is nullptr");
+    return (*__data_)[__offset_];
+  }
+
+  _CCCL_API constexpr __simd_iterator& operator++() noexcept
+  {
+    return *this += 1;
```
miscco: Important: We should properly spell those out. The operators are really simple and this makes it simpler for the compiler and also avoids the need for the debugger to step into
```suggestion
    ++__offset_;
    _CCCL_ASSERT(::cuda::in_range(__offset_, __simd_size_type{0}, _Vp::__size),
                 "cuda::std::simd::__simd_iterator: offset is out of range");
    return *this;
```

libcudacxx/include/cuda/std/__simd/iterator.h
```diff
@@ -0,0 +1,264 @@
...
+  }
+
+  _CCCL_API constexpr __simd_iterator operator++(int) noexcept
+  {
+    __simd_iterator __tmp = *this;
+    *this += 1;
+    return __tmp;
```
miscco: Ditto: Spell out
```suggestion
    __simd_iterator __tmp = *this;
    ++__offset_;
    _CCCL_ASSERT(::cuda::in_range(__offset_, __simd_size_type{0}, _Vp::__size),
                 "cuda::std::simd::__simd_iterator: offset is out of range");
    return __tmp;
```

libcudacxx/include/cuda/std/__simd/iterator.h
```diff
@@ -0,0 +1,264 @@
...
+  }
+
+  _CCCL_API constexpr __simd_iterator& operator--() noexcept
+  {
+    return *this -= 1;
```
miscco: Spell out applies throughout

libcudacxx/include/cuda/std/__simd/iterator.h
```diff
@@ -0,0 +1,264 @@
...
+
+  // [simd.iterator] comparisons
+
+  [[nodiscard]] _CCCL_API friend constexpr bool
+  operator==(const __simd_iterator& __a, const __simd_iterator& __b) noexcept
+  {
+    return __a.__data_ == __b.__data_ && __a.__offset_ == __b.__offset_;
+  }
+
+  [[nodiscard]] _CCCL_API friend constexpr bool operator==(const __simd_iterator __i, default_sentinel_t) noexcept
```
miscco, line 158: Critical: This is missing `&` for all the sentinel comparisons.

We should be consistent and either use `const __simd_iterator ` *or* `const __simd_iterator&` throughout.

libcudacxx/include/cuda/std/__simd/iterator.h
```diff
@@ -0,0 +1,264 @@
...
+    return __i == __s;
+  }
+#endif // _CCCL_STD_VER <= 2017
+
+#if _LIBCUDACXX_HAS_SPACESHIP_OPERATOR()
+  [[nodiscard]]
+  _CCCL_API friend constexpr auto operator<=>(const __simd_iterator __a, const __simd_iterator __b) noexcept
+  {
+    _CCCL_ASSERT(__a.__data_ == __b.__data_, "cuda::std::simd::__simd_iterator: iterators refer to different objects");
+    _CCCL_ASSERT(__a.__data_ != nullptr, "cuda::std::simd::__simd_iterator: default-constructed iterator");
```
miscco: Critical: I believe this is not correct. It is perfectly fine to compare two default constructed iterators.
```suggestion
```

libcudacxx/include/cuda/std/__simd/iterator.h
```diff
@@ -0,0 +1,264 @@
...
+    return __a.__offset_ <=> __b.__offset_;
+  }
+#else // ^^^ _LIBCUDACXX_HAS_SPACESHIP_OPERATOR() ^^^ / vvv !_LIBCUDACXX_HAS_SPACESHIP_OPERATOR() vvv
+  [[nodiscard]]
+  _CCCL_API friend constexpr bool operator<(const __simd_iterator& __a, const __simd_iterator& __b) noexcept
+  {
+    _CCCL_ASSERT(__a.__data_ == __b.__data_, "cuda::std::simd::__simd_iterator: iterators refer to different objects");
+    _CCCL_ASSERT(__a.__data_ != nullptr, "cuda::std::simd::__simd_iterator: default-constructed iterator");
```
miscco: Critical: It is valid to compare default constructed iterators
```suggestion
```

libcudacxx/include/cuda/std/__simd/iterator.h
```diff
@@ -0,0 +1,264 @@
...
+    return __a.__offset_ < __b.__offset_;
+  }
+
+  [[nodiscard]] _CCCL_API friend constexpr bool
+  operator>(const __simd_iterator& __a, const __simd_iterator& __b) noexcept
+  {
+    return __b < __a;
```
miscco: Important: Simple enough to spell out
```suggestion    
    _CCCL_ASSERT(__a.__data_ == __b.__data_, "cuda::std::simd::__simd_iterator: iterators refer to different objects");
    return __b.__offset_ < __a.__offset_;
```

libcudacxx/include/cuda/std/__simd/iterator.h
```diff
@@ -0,0 +1,264 @@
...
+_CCCL_BEGIN_NAMESPACE_CUDA_STD_SIMD
+
+// [simd.iterator], class template __simd_iterator
+template <typename _Vp>
+class __simd_iterator
+{
+  _Vp* __data_               = nullptr;
+  __simd_size_type __offset_ = 0;
+
+  _CCCL_API constexpr __simd_iterator(_Vp& __d, const __simd_size_type __off) noexcept
```
miscco: Nitpick: We might want a second overload that takes a `_Vp*` so that we can use this internally

libcudacxx/include/cuda/std/__simd/iterator.h
```diff
@@ -0,0 +1,264 @@
...
+  [[nodiscard]] _CCCL_API friend constexpr __simd_iterator
+  operator-(__simd_iterator __i, const difference_type __n) noexcept
+  {
+    return __i -= __n;
+  }
+
+  [[nodiscard]] _CCCL_API friend constexpr difference_type
+  operator-(const __simd_iterator __a, const __simd_iterator __b) noexcept
+  {
+    _CCCL_ASSERT(__a.__data_ != nullptr, "cuda::std::simd::__simd_iterator: data is nullptr");
```
miscco: Critical: It is valid to compare default constructed iterators
```suggestion
```

libcudacxx/include/cuda/std/__simd/iterator.h
```diff
@@ -0,0 +1,264 @@
...
+    _CCCL_ASSERT(__a.__data_ == __b.__data_, "cuda::std::simd::__simd_iterator: iterators refer to different objects");
+    return __a.__offset_ - __b.__offset_;
+  }
+
+  [[nodiscard]] _CCCL_API friend constexpr difference_type
+  operator-(const __simd_iterator __i, default_sentinel_t) noexcept
+  {
+    _CCCL_ASSERT(__i.__data_ != nullptr, "cuda::std::simd::__simd_iterator: data is nullptr");
```
miscco:
```suggestion
```

libcudacxx/include/cuda/std/__simd/iterator.h
```diff
@@ -0,0 +1,264 @@
...
+    return __i.__offset_ - _Vp::__size;
+  }
+
+  [[nodiscard]] _CCCL_API friend constexpr difference_type
+  operator-(default_sentinel_t, const __simd_iterator __i) noexcept
+  {
+    _CCCL_ASSERT(__i.__data_ != nullptr, "cuda::std::simd::__simd_iterator: data is nullptr");
```
miscco:
```suggestion
```

libcudacxx/test/libcudacxx/std/numerics/simd/simd.iterator/begin_end.pass.cpp
```diff
@@ -0,0 +1,214 @@
...
+// constexpr default_sentinel_t end() const noexcept;
+// constexpr default_sentinel_t cend() const noexcept;
+
+#include <cuda/std/__simd_>
+#include <cuda/std/cassert>
+#include <cuda/std/cstddef>
+#include <cuda/std/iterator>
+#include <cuda/std/type_traits>
+
+#include "../simd_test_utils.h"
```
miscco, line 26: Important: We should also test against the host STL if we can

See the iterator traits checks for the cuda iterators

libcudacxx/test/libcudacxx/std/numerics/simd/simd.iterator/begin_end.pass.cpp
miscco, line 1: Nitpick: please split this up into individual tests for the respective operations. 

That reduces the compile time per test and also makes it much simpler to pinpoint issues

miscco (changes requested):

libcudacxx/include/cuda/std/__simd/iterator.h
fbusato: that's a good idea. Something I can also add to the SKILL

libcudacxx/include/cuda/std/__simd/iterator.h
fbusato: I would prefer to follow the c++ spec by using `const __simd_iterator`

libcudacxx/test/libcudacxx/std/numerics/simd/simd.iterator/begin_end.pass.cpp
fbusato: I reorganized `begin_end.pass.cpp` and split the unit tests

libcudacxx/include/cuda/std/__simd/iterator.h
miscco: Then please remove the `&` from the others

libcudacxx/test/libcudacxx/std/numerics/simd/simd.iterator/arithmetic.pass.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+template <typename T, int N>
+TEST_FUNC constexpr void test_increment_decrement()
+{
+  using Vec = simd::basic_vec<T, simd::fixed_size<N>>;
+  Vec vec   = make_iota_vec<T, N>();
+
+  // pre-increment
+  auto it   = vec.begin();
+  auto& ref = ++it;
```
miscco: Important: This is missing a check that we return the right type
```suggestion
  auto&& ref = ++it;
  static_assert(cuda::std::is_same_v<decltype(ref), Iter&>);
```

Applies throughout

libcudacxx/test/libcudacxx/std/numerics/simd/simd.iterator/arithmetic.pass.cpp
```diff
@@ -0,0 +1,138 @@
...
+  assert(&ref == &it);
```
miscco: Nitpick: use  `cuda::std::addressof` as `operator&` might have a different  meaning

libcudacxx/test/libcudacxx/std/numerics/simd/simd.iterator/arithmetic.pass.cpp
```diff
@@ -0,0 +1,138 @@
...
+//----------------------------------------------------------------------------------------------------------------------
+// increment and decrement
+
+template <typename T, int N>
+TEST_FUNC constexpr void test_increment_decrement()
+{
+  using Vec = simd::basic_vec<T, simd::fixed_size<N>>;
+  Vec vec   = make_iota_vec<T, N>();
+
+  // pre-increment
```
miscco: Nitpick: I would strongly suggest using braces to separate the different conditions.

libcudacxx/test/libcudacxx/std/numerics/simd/simd.iterator/arithmetic.pass.cpp
```diff
@@ -0,0 +1,138 @@
...
+  assert(&ref == &it);
+  if constexpr (N > 1)
+  {
+    assert(*it == static_cast<T>(1));
+  }
+
+  // post-increment
+  auto it2 = vec.begin();
+  auto old = it2++;
+  assert(*old == static_cast<T>(0));
```
miscco: Nitpick: the proper check would be
```suggestion
  assert(old == it);
```

libcudacxx/test/libcudacxx/std/numerics/simd/simd.iterator/arithmetic.pass.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+template <typename T, int N>
+TEST_FUNC constexpr void test_arithmetic()
+{
+  using Vec = simd::basic_vec<T, simd::fixed_size<N>>;
+  Vec vec   = make_iota_vec<T, N>();
+
+  auto it = vec.begin();
+
+  auto it2 = it + N;
```
miscco, line 152: Ditto missing type check
```suggestion
  auto&& it2 = it + N;
  static_assert(cuda::std::is_same_v<decltype(it2), Iter>);
```

libcudacxx/test/libcudacxx/std/numerics/simd/simd.iterator/arithmetic.pass.cpp
miscco: Also we should check `noexcept` for all operations

libcudacxx/test/libcudacxx/std/numerics/simd/simd.iterator/begin_end.pass.cpp
```diff
@@ -0,0 +1,111 @@
...
+  assert(cuda::std::begin(obj) != cuda::std::end(obj));
+  assert(cuda::std::begin(const_obj) != cuda::std::end(const_obj));
+  assert(cuda::std::cbegin(const_obj) != cuda::std::cend(const_obj));
+
+  int count = 0;
+  for (auto it = obj.begin(); it != obj.end(); ++it)
+  {
+    ++count;
+  }
+  assert(count == N);
```
miscco: Question: Should this just be `cuda::std::distance`?

libcudacxx/test/libcudacxx/std/numerics/simd/simd.iterator/comparison.pass.cpp
```diff
@@ -0,0 +1,98 @@
...
+template <typename T, int N>
+TEST_FUNC constexpr void test_comparisons()
+{
+  using Vec = simd::basic_vec<T, simd::fixed_size<N>>;
+  Vec vec   = make_iota_vec<T, N>();
+
+  auto a = vec.begin();
+  auto b = vec.begin();
+
+  assert(a == b);
```
miscco, line 52: Important: Missing return type check and noexcept

libcudacxx/test/libcudacxx/std/numerics/simd/simd.iterator/comparison.pass.cpp
```diff
@@ -0,0 +1,98 @@
...
+// sentinel comparisons and noexcept
+
+template <typename T, int N>
+TEST_FUNC constexpr void test_sentinel()
+{
+  using Vec = simd::basic_vec<T, simd::fixed_size<N>>;
+  Vec vec{};
+
+  auto it = vec.begin();
+  assert(!(it == cuda::std::default_sentinel));
```
miscco: Ditto Missing return type and noexcept  check

libcudacxx/test/libcudacxx/std/numerics/simd/simd.iterator/comparison.pass.cpp
```diff
@@ -0,0 +1,98 @@
...
+  assert(it != cuda::std::default_sentinel);
+
+  it += N;
```
miscco, line 100: I believe it would be clearer to do
```suggestion
  auto end = vec.end();
```

libcudacxx/test/libcudacxx/std/numerics/simd/simd.iterator/concept.compile.pass.cpp
```diff
@@ -0,0 +1,53 @@
...
+
+template <typename Iter>
+TEST_FUNC void check_iterator_concepts()
+{
+  using Sentinel = cuda::std::default_sentinel_t;
+
+  static_assert(cuda::std::input_iterator<Iter>);
+  static_assert(cuda::std::forward_iterator<Iter>);
+  static_assert(cuda::std::bidirectional_iterator<Iter>);
+  static_assert(cuda::std::random_access_iterator<Iter>);
```
miscco, line 32: I believe this is missing `cuda::std::contiguous_iterator`

libcudacxx/test/libcudacxx/std/numerics/simd/simd.iterator/derefence.pass.cpp
```diff
@@ -0,0 +1,58 @@
...
+TEST_FUNC constexpr void test_dereference()
+{
+  using Vec = simd::basic_vec<T, simd::fixed_size<N>>;
+  Vec vec   = make_iota_vec<T, N>();
+
+  auto it = vec.begin();
+  for (int i = 0; i < N; ++i)
+  {
+    assert(*it == static_cast<T>(i));
+    assert(it[0] == static_cast<T>(i));
```
miscco, line 40: Missing return type and noexcept

libcudacxx/test/libcudacxx/std/numerics/simd/simd.iterator/iterator_traits.compile.pass.cpp
```diff
@@ -0,0 +1,63 @@
...
+#include "test_macros.h"
+
+namespace simd = cuda::std::simd;
+
+template <typename Iter, typename ValueType>
+TEST_FUNC void check_iter_traits()
+{
+  static_assert(cuda::std::is_same_v<typename Iter::value_type, ValueType>);
+  static_assert(cuda::std::is_same_v<typename Iter::iterator_category, cuda::std::input_iterator_tag>);
+  static_assert(cuda::std::is_same_v<typename Iter::difference_type, cuda::std::ptrdiff_t>);
```
miscco, line 37: This is missing `reference`, `pointer` and preferably `iterator_concept`

miscco (changes requested):

libcudacxx/test/libcudacxx/std/numerics/simd/simd.iterator/begin_end.pass.cpp
fbusato: yes, good idea

libcudacxx/test/libcudacxx/std/numerics/simd/simd.iterator/iterator_traits.compile.pass.cpp
```diff
@@ -0,0 +1,71 @@
...
+{
+  using Traits = cuda::std::iterator_traits<Iter>;
+
+  static_assert(cuda::std::is_same_v<typename Iter::value_type, ValueType>);
+  static_assert(cuda::std::is_same_v<typename Iter::iterator_category, cuda::std::input_iterator_tag>);
+  static_assert(cuda::std::is_same_v<typename Iter::iterator_concept, cuda::std::random_access_iterator_tag>);
+  static_assert(cuda::std::is_same_v<typename Iter::difference_type, cuda::std::ptrdiff_t>);
+
+  static_assert(cuda::std::is_same_v<typename Traits::value_type, ValueType>);
+  static_assert(cuda::std::is_same_v<typename Traits::iterator_category, cuda::std::input_iterator_tag>);
```
miscco, line 40: Nitpick: missing `iterator_concept`

miscco, line 40: Important: We ware providing those iterators in all supported standard modes on host and device. If someone wants to use them together with the host STL iterator facilities they will have a bad time. 

We should consider overloading `::std::next`, `::std::prev`, `::std::distance` and `::std::advance` as we do for the cuda iterators

libcudacxx/include/cuda/std/__simd/iterator.h
```diff
@@ -258,6 +260,59 @@ class __simd_iterator
...
+  using __iter            = simd::__simd_iterator<_Vp>;
+  using iterator_concept  = typename __iter::iterator_concept;
+  using iterator_category = typename __iter::iterator_category;
+  using value_type        = typename __iter::value_type;
+  using difference_type   = typename __iter::difference_type;
+  using pointer           = void;
+  using reference         = value_type;
+};
+
+_CCCL_END_NAMESPACE_CUDA_STD
```
miscco: This should not be necessary, what was the reason for adding it?

fbusato: this is for required for `iterator_concept`. Is there any alternative?

miscco: You can just define it inside the iterator

fbusato: not really, the following code doesn't work if we don't specialize `iterator_traits`
```cpp
  using Vec    = simd::basic_vec<T, simd::fixed_size<N>>;
  using Iter    = typename Vec::iterator;
  using Traits = cuda::std::iterator_traits<Iter>;
  static_assert(is_same_v<typename Traits::iterator_concept, cuda::std::random_access_iterator_tag>);
```

miscco: Sorry that was me, I always forget that its a shitshow

miscco (approved):

miscco (approved):

PR #8473, opened 2026-04-15
Replace `detail::merge_sort::dispatch` by CUB's public API

### LINKED ISSUE #8376: Use the new tuning API internally for `detail::merge_sort::dispatch`

bernhardmgruber:
Fixes: #8376

- [x] Merge before: #8381
- [x] No SASS changes in `thrust.test.sort` on SM120
- [x] No SASS changes in `cub.bench.merge_sort.pairs.base` on SM120

### DIFF
--- a/cub/test/catch2_test_device_reduce_env.cu
+++ b/cub/test/catch2_test_device_reduce_env.cu
@@ -127,11 +127,12 @@ struct unrelated_tuning
   }
 };
 
-using block_sizes = c2h::type_list<cuda::std::integral_constant<int, 32>, cuda::std::integral_constant<int, 64>>;
+using block_sizes =
+  c2h::type_list<cuda::std::integral_constant<unsigned int, 32>, cuda::std::integral_constant<unsigned int, 64>>;
 
 C2H_TEST("Device reduce can be tuned", "[reduce][device]", block_sizes)
 {
-  constexpr int target_block_size = c2h::get<0, TestType>::value;
+  constexpr unsigned int target_block_size = c2h::get<0, TestType>::value;
   c2h::device_vector<unsigned int> d_block_size(1);
   block_size_check_t block_size_check{thrust::raw_pointer_cast(d_block_size.data())};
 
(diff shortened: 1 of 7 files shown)


cub/test/catch2_test_device_merge_sort_env_api.cu
```diff
@@ -5,15 +5,54 @@
...
+struct unrelated_tuning
+{
+  // should never be called
+  auto operator()(cuda::arch_id /*arch*/) const -> unrelated_policy
+  {
+    throw 1337;
+  }
+};
+
+struct block_size_compare_t
```
oleksandr-pavlyk: This operator assumes that only one CTA is going to be launched. Perhaps an assertion in the if branch should be added to make that assumption explicit

oleksandr-pavlyk (approved):

cub/test/catch2_test_device_merge_sort_env_api.cu
bernhardmgruber: Right. I added `atomicMax` to reduce the `blockDim` in case there would be multiple blocks.

cub/test/catch2_test_device_reduce_env.cu
```diff
@@ -41,13 +41,14 @@ namespace stdexec = cuda::std::execution;
...
   __device__ int operator()(int a, int b)
   {
     if (threadIdx.x == 0)
     {
-      *ptr = blockDim.x;
+      // use an atomic operation to write the block dim in case multiple blocks are launched
+      atomicMax(ptr, blockDim.x);
     }
     return a + b;
   }
```
bernhardmgruber, line 54: Remark: This is a drive-by fix for consistency.

cub/test/catch2_test_env_launch_helper.h
```diff
@@ -512,8 +512,11 @@ void launch(ActionT action, Args... args)
     REQUIRE(cudaSuccess == cudaStreamDestroy(stream));
   }
 
-  size_t expected_bytes_allocated = fixed_env.query(get_expected_allocation_size_t{});
-  REQUIRE(expected_bytes_allocated == bytes_allocated);
+  if constexpr (cuda::std::execution::__queryable_with<env_t, get_expected_allocation_size_t>)
+  {
+    const size_t expected_bytes_allocated = fixed_env.query(get_expected_allocation_size_t{});
+    REQUIRE(expected_bytes_allocated == bytes_allocated);
+  }
```
bernhardmgruber, line 519: By making the `expected_allocation_size` check in the launch wrapper optional, we can also use the launch wrappers to test whether e.g. tunings were applied. At least for `TEST_LAUNCH != 1`.

@gonidelis I think you ran into the same issue, which is why many tests are guarded by `#if TEST_LAUNCH == 0`. Just FYI.

miscco (approved):

PR #8474, opened 2026-04-15
Add env overloads for inplace DeviceScan

gonidelis:
leftover from #7548

### DIFF
--- a/cub/cub/device/device_scan.cuh
+++ b/cub/cub/device/device_scan.cuh
@@ -423,7 +423,8 @@ struct DeviceScan
 #else
             ::cuda::std::execution::env<>
 #endif
-            >
+            ,
+            typename = ::cuda::std::enable_if_t<::cuda::std::is_integral_v<NumItemsT>>>
   [[nodiscard]] CUB_RUNTIME_FUNCTION static cudaError_t
   ExclusiveSum(InputIteratorT d_in, OutputIteratorT d_out, NumItemsT num_items, EnvT env = {})
   {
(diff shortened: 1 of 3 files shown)


cub/cub/device/device_scan.cuh
```diff
@@ -423,7 +423,8 @@ struct DeviceScan
 #else
             ::cuda::std::execution::env<>
 #endif
-            >
+            ,
+            typename = ::cuda::std::enable_if_t<::cuda::std::is_integral_v<NumItemsT>>>
```
bernhardmgruber: Important: please move the `enable_if` from a default argument to an actual parameter. This makes SFINAE more robust. For a reason I forgot :S @miscco ?
```suggestion
            ::cuda::std::enable_if_t<::cuda::std::is_integral_v<NumItemsT>, int> = 0>
```
Applies to other overloads as well.

bernhardmgruber (commented):
Please fix the enable_if, otherwise LGTM

cub/cub/device/device_scan.cuh
gonidelis: is it because defaulted enable_if's with opposite conditions could still create ambiguities? 

```
template <typename T, typename = enable_if_t< is_integral_v<T>>>  void f(T);
template <typename T, typename = enable_if_t<!is_integral_v<T>>>  void f(T);  // redefinition error
```

bernhardmgruber: Oh yes! That's the reason! Well done!

bernhardmgruber (approved):

PR #8475, opened 2026-04-16
`cuda::std::simd` Complex

fbusato:
## Description

Follow up PR of

- https://github.com/NVIDIA/cccl/pull/8251

The PR addresses:

- [[simd.complex.math]](https://eel.is/c++draft/simd#complex.math)
- [[simd.complex.access]](https://eel.is/c++draft/simd#complex.access)
- [[simd.ctor]](https://eel.is/c++draft/simd#ctor) (complex)


libcudacxx/include/cuda/std/__simd/specializations/fixed_size_mask.h
```diff
@@ -0,0 +1,175 @@
...
+  template <typename _Generator, __simd_size_type... _Is>
+  [[nodiscard]] _CCCL_API static constexpr _MaskStorage
+  __generate_init(_Generator&& __g, integer_sequence<__simd_size_type, _Is...>)
+  {
+#if _CCCL_STD_VER >= 2020
+    _MaskStorage __result;
+    ((__result.__data[_Is] = static_cast<bool>(__g(integral_constant<__simd_size_type, _Is>()))), ...);
+    return __result;
+#else // ^^^ C++20 ^^^ / vvv C++17 vvv
+    return _MaskStorage{{ static_cast<bool>(__g(integral_constant<__simd_size_type, _Is>()))... }};
```
Jacobfaib, line 85: Does this make a difference?

libcudacxx/include/cuda/std/__simd/specializations/fixed_size_mask.h
```diff
@@ -0,0 +1,175 @@
...
+#endif // _CCCL_STD_VER < 2020
+  }
+
+  template <typename _Generator>
+  [[nodiscard]] _CCCL_API static constexpr _MaskStorage __generate(_Generator&& __g)
+  {
+    return __generate_init(__g, make_integer_sequence<__simd_size_type, _Np>());
```
Jacobfaib, line 92:
```suggestion
    return __generate_init(__g, ::cuda::std::make_integer_sequence<__simd_size_type, _Np>());
```
(there were a few others I also noticed)

libcudacxx/include/cuda/std/__simd/specializations/fixed_size_vec.h
```diff
@@ -0,0 +1,361 @@
...
+  template <typename _Generator, __simd_size_type... _Is>
+  [[nodiscard]] _CCCL_API static constexpr _SimdStorage
+  __generate_init(_Generator&& __g, integer_sequence<__simd_size_type, _Is...>)
+  {
+#if _CCCL_STD_VER >= 2020
+    _SimdStorage __result;
+    ((__result.__data[_Is] = __g(integral_constant<__simd_size_type, _Is>())), ...);
+    return __result;
+#else // ^^^ C++20 ^^^ / vvv C++17 vvv
+    return _SimdStorage{{ __g(integral_constant<__simd_size_type, _Is>())... }};
```
Jacobfaib, line 93: Ditto here

libcudacxx/include/cuda/std/__simd/concepts.h
```diff
@@ -0,0 +1,264 @@
...
+template <typename _Tp>
+_CCCL_CONCEPT __has_shift_left_size = _CCCL_REQUIRES_EXPR((_Tp), _Tp __t)((__t << __simd_size_type{}));
+
+template <typename _Tp>
+_CCCL_CONCEPT __has_shift_right_size = _CCCL_REQUIRES_EXPR((_Tp), _Tp __t)((__t >> __simd_size_type{}));
+
+// [simd.comparison], comparison operator constraints
+
+template <typename _Tp>
+_CCCL_CONCEPT __has_equal_to = _CCCL_REQUIRES_EXPR((_Tp), _Tp __a, _Tp __b)((__a == __b));
```
Jacobfaib, line 243: Could this be `__is_cpp17_equality_comparable_v` in `libcudacxx/include/cuda/std/__type_traits/is_comparable.h` instead? If so, we should consider moving all of these to `is_comparable.h`, currently I think we only have `==` and `<`.

libcudacxx/include/cuda/std/__simd/exposition.h
```diff
@@ -0,0 +1,73 @@
...
+
+template <typename _Tp>
+inline constexpr bool __is_complex_vectorizable_v<::cuda::std::complex<_Tp>> = ::cuda::is_floating_point_v<_Tp>;
+
+// [simd.expos], vectorizable types:
+// all standard integer types, character types, and the types float and double ([basic.fundamental]);
+// std::float16_t, std::float32_t, and std::float64_t if defined ([basic.extended.fp]); and
+// complex<T> where T is a vectorizable floating-point type.
+template <typename _Tp>
+inline constexpr bool __is_vectorizable_v =
```
Jacobfaib, line 55: Should these be proper concepts?

libcudacxx/include/cuda/std/__simd_
Jacobfaib, line 1: `cuda/std/simd`?

fbusato: we will switch to `cuda/std/simd` after all library is covered

libcudacxx/include/cuda/std/__simd/exposition.h
fbusato: the c++ specification doesn't propose it as a concept. I don't thing it makes any different for the compile time

libcudacxx/include/cuda/std/__simd/specializations/fixed_size_mask.h
fbusato: we are already in the `cuda::std` namespace. This is not needed

libcudacxx/include/cuda/std/__simd/specializations/fixed_size_mask.h
fbusato: unfortunately the two paths are needed. The C++17 version doesn't work with C++20

libcudacxx/include/cuda/std/__simd/complex_math.h
```diff
@@ -0,0 +1,515 @@
...
+  [[nodiscard]] _CCCL_API constexpr auto operator()(const _Tp& __z) const noexcept
+  {
+    return __z.imag();
+  }
+};
+
+struct __fn_abs
+{
+  template <typename _Tp>
+  [[nodiscard]] _CCCL_API constexpr auto operator()(const _Tp& __z) const
```
Jacobfaib:
```suggestion
  [[nodiscard]] _CCCL_API constexpr auto operator()(const _Tp& __z) const noexcept
```

Jacobfaib: Actually I think all of these can be made `noexcept`.

libcudacxx/include/cuda/std/__simd/basic_vec.h
```diff
@@ -202,8 +220,18 @@ class basic_vec<_Tp, _Abi, enable_if_t<__is_vectorizable_v<_Tp> && __is_enabled_
...
+  // [simd.ctor] complex constructor
+  _CCCL_TEMPLATE(typename _Up = _Tp)
+  _CCCL_REQUIRES(__is_cuda_std_complex_v<_Up>)
+  _CCCL_API constexpr basic_vec(const rebind_t<typename _Up::value_type, basic_vec>& __reals,
+                                const rebind_t<typename _Up::value_type, basic_vec>& __imags = {}) noexcept
+  {
+    _CCCL_PRAGMA_UNROLL_FULL()
+    for (__simd_size_type __i = 0; __i < __size; ++__i)
+    {
+      __s_.__set(__i, value_type(__reals[__i], __imags[__i]));
```
Jacobfaib: Is the use of `()` vs `{}` constructor intentional? I.e. are you intentionally allowing narrowing conversion with `()` (and below as well)?

libcudacxx/include/cuda/std/__simd/complex_math.h
fbusato: that's a good point, but it makes no difference because they are used within the construction which is without `noexcept`. Also, the underlying operations don't have `noexcept`

libcudacxx/include/cuda/std/__simd/basic_vec.h
fbusato: I don't think so. I prefer `{}` as you suggested

libcudacxx/include/cuda/std/__simd/complex_math.h
Jacobfaib: I checked a few (`abs()`, `arg()`, etc), and even if the top-level ones aren't `noexcept`, they all seem to call `noexcept` functions (e.g. `arg()` calls `atan2()` which is `noexcept`), so those need to be fixed up at some point, but the main thing is I think we are OK to just slap `noexcept` on all of these.

fbusato: I would prefer to have another `noexcept` PR before making this modification. Would do you like to open it?

libcudacxx/include/cuda/std/__simd/concepts.h
fbusato: I moved the traits to `cuda/std/__type_traits/operations.h`. `__has_equal_to` is more precise and also more efficient because it is expressed with concepts (in C++20)

libcudacxx/include/cuda/std/__simd/basic_vec.h
```diff
@@ -155,7 +172,8 @@ class basic_vec<_Tp, _Abi, enable_if_t<__is_vectorizable_v<_Tp> && __is_enabled_
   // [simd.ctor] converting constructor from basic_vec<U, UAbi> (explicit overload)
   _CCCL_TEMPLATE(typename _Up, typename _UAbi)
   _CCCL_REQUIRES((__simd_size_v<_Up, _UAbi> == __size) _CCCL_AND(__explicitly_convertible_to<_Up, value_type>)
-                   _CCCL_AND(__is_vec_ctor_explicit<_Up, value_type>))
+                   _CCCL_AND(__is_vec_ctor_explicit<_Up, value_type>)
+                     _CCCL_AND(!__is_complex_real_type_v<value_type, _Up>))
```
miscco: Nitpick: If this needs to be short-circuiting, we should move this into a `_CCCL_CONCEPT` otherwise, I believe it would be clearer to put this into a `static constexpr bool __can_explicitly_convert<_Up, _UAbi>` member variable.

Those complex conditions are really hard to read

libcudacxx/include/cuda/std/__simd/basic_vec.h
```diff
@@ -218,12 +236,22 @@ class basic_vec<_Tp, _Abi, enable_if_t<__is_vectorizable_v<_Tp> && __is_enabled_
...
+  _CCCL_REQUIRES(__is_cuda_std_complex_v<_Up>)
+  _CCCL_API constexpr basic_vec(const rebind_t<typename _Up::value_type, basic_vec>& __reals,
+                                const rebind_t<typename _Up::value_type, basic_vec>& __imags = {}) noexcept
+  {
+    _CCCL_PRAGMA_UNROLL_FULL()
+    for (__simd_size_type __i = 0; __i < __size; ++__i)
+    {
+      __s_.__set(__i, value_type{__reals[__i], __imags[__i]});
+    }
+  }
```
miscco: Important: We conventionally use `_Up`  as a *different* type than `_Tp`

If this is meant to properly SFINAE out with `_Up` as a dependent type, I would rather have this named `_Tp2` which we commonly use if that is meant to just make it dependent

libcudacxx/include/cuda/std/__simd/exposition.h
```diff
@@ -38,12 +40,25 @@ _CCCL_BEGIN_NAMESPACE_CUDA_STD_SIMD
 template <size_t _Bytes>
 using __integer_from = __make_nbit_int_t<_Bytes * 8, true>;
 
+// complex<T> where T is a vectorizable floating-point type
+template <typename _Tp>
+inline constexpr bool __is_complex_vectorizable_v = false;
+
+template <typename _Tp>
+inline constexpr bool __is_complex_vectorizable_v<::cuda::std::complex<_Tp>> = ::cuda::is_floating_point_v<_Tp>;
+
```
miscco, line 57: Question: I believe we should also allow for `::std::complex` to be vectorizable.

We would need to add `_CCCL_EXEC_CHECK_DISABLE` to the constructor.

libcudacxx/include/cuda/std/__simd/exposition.h
```diff
@@ -38,12 +40,25 @@ _CCCL_BEGIN_NAMESPACE_CUDA_STD_SIMD
...
+// complex<T> where T is a vectorizable floating-point type.
 template <typename _Tp>
 inline constexpr bool __is_vectorizable_v =
-  __is_extended_arithmetic_v<_Tp> && !is_same_v<_Tp, bool> && !is_const_v<_Tp> && !is_volatile_v<_Tp>;
+  (__is_extended_arithmetic_v<_Tp> || __is_complex_vectorizable_v<_Tp>)
+  && !is_same_v<_Tp, bool> && !is_const_v<_Tp> && !is_volatile_v<_Tp>;
+
+// [simd.expos], simd-complex-value-type
+template <typename _Tp>
+using __simd_complex_value_type = typename _Tp::value_type;
```
miscco: Nitpick: We conventionally suffix alias templates like this with `_t` like `remove_reference_t`

libcudacxx/test/libcudacxx/std/numerics/simd/simd.complex/complex_math_trig.pass.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+template <typename T, int N>
+TEST_FUNC void test_type()
+{
+  test_trig<T, N>();
+  test_hyperbolic<T, N>();
+}
+
+DEFINE_COMPLEX_TEST_NONCONSTEXPR()
+DEFINE_BASIC_VEC_TEST_RUNTIME()
```
miscco: Important: The unit tests should be as simple as possible and easily comprehensible. These macros make it much harder to follow what is actually tested, especially in review.

I do not believe we gain anything by replacing a 6 line function with a macro. Please inline

libcudacxx/test/libcudacxx/std/numerics/simd/simd.vec.class/ctor.pass.cpp
```diff
@@ -258,19 +218,12 @@ template <typename T, typename U, int N>
...
+  auto arr  = make_iota_array<U, N>();
 
   static_assert(!noexcept(Vec(arr, simd::flag_convert)));
 
   Vec vec(arr, simd::flag_convert);
-  for (int i = 0; i < N; ++i)
-  {
-    assert(vec[i] == static_cast<T>(static_cast<U>(i + 1)));
-  }
+  assert((vec == make_iota_array<T, N>()));
```
miscco: Nitpick: Next time please move these cleanups into a separate PR

libcudacxx/test/libcudacxx/std/numerics/simd/simd_test_utils.h
```diff
@@ -95,6 +97,57 @@ struct is_less_than_index
 template <int Bytes>
 using integer_from_t = cuda::std::__make_nbit_int_t<Bytes * 8, true>;
 
+//----------------------------------------------------------------------------------------------------------------------
+// Approximate floating-point comparison for SIMD complex math tests
+
+// even if SIMD applies mathematical operations for each component, the compilers could still performance different
```
miscco:
```suggestion
// even if SIMD applies mathematical operations for each component, the compilers could still perform different
```

miscco (changes requested):

libcudacxx/include/cuda/std/__simd/complex_math.h
```diff
@@ -0,0 +1,523 @@
...
+  {
+    return ::cuda::std::log10(__z);
+  }
+};
+
+struct __fn_sqrt
+{
+  template <typename _Tp>
+  [[nodiscard]] _CCCL_API constexpr auto operator()(const _Tp& __z) const noexcept
```
miscco, line 137: This one is `noexcept`, please add to all functions, as we only deal with arithmetic types

libcudacxx/include/cuda/std/__simd/complex_math.h
```diff
@@ -0,0 +1,523 @@
...
+// [simd.complex.math], unary complex functions returning real-valued result
+
+_CCCL_TEMPLATE(typename _Tp, typename _Abi)
+_CCCL_REQUIRES(__is_complex_vectorizable_v<_Tp>)
+[[nodiscard]] _CCCL_API constexpr rebind_t<__simd_complex_value_type<_Tp>, basic_vec<_Tp, _Abi>>
+real(const basic_vec<_Tp, _Abi>& __v) noexcept
+{
+  using __vec_t    = basic_vec<_Tp, _Abi>;
+  using __result_t = rebind_t<__simd_complex_value_type<_Tp>, __vec_t>;
+  return __result_t{__gen_complex_apply_unary<__vec_t, __fn_real>{__v}};
```
miscco, line 289: Important: I am a bit concerned that this will generate suboptimal codegen because it might force a copy of the input `__v` into the generator

Have you compare the codegen to what a simple dispatch via a `index_sequence`?

I had something like this in mind
```suggestion
  return __result_t{__v[__simd_size_type{_Idx}].real()...};
```

Note that this would also be more efficient compile time wise as we do not need to instantiate a lot of the types, but only one `index_sequence`

libcudacxx/include/cuda/std/__simd/basic_vec.h
```diff
@@ -237,11 +265,55 @@ class basic_vec<_Tp, _Abi, enable_if_t<__is_vectorizable_v<_Tp> && __is_enabled_
...
-  // constexpr real-type imag() const noexcept;
-  // constexpr void real(const real-type& __v) noexcept;
-  // constexpr void imag(const real-type& __v) noexcept;
+  // [simd.complex.access], basic_vec complex accessors
+
+  _CCCL_TEMPLATE(typename _Up = _Tp)
+  _CCCL_REQUIRES(__is_cuda_std_complex_v<_Up>)
+  [[nodiscard]] _CCCL_API constexpr rebind_t<typename _Up::value_type, basic_vec> real() const noexcept
+  {
+    rebind_t<typename _Up::value_type, basic_vec> __ret;
```
miscco: Important: This will trigger warnings about uninitialized `__ret` in constexpr evaluation

libcudacxx/include/cuda/std/__simd/basic_vec.h
fbusato: agree. good idea

libcudacxx/include/cuda/std/__simd/complex_math.h
fbusato: I added `noexcept` to all functions that call other `noexcept` functions. The first problem is with the underlying functions

libcudacxx/include/cuda/std/__simd/basic_vec.h
fbusato: I don't think so because `rebind_t<typename _Up::value_type, basic_vec>` is a SIMD vector type, which has a default constructor

libcudacxx/include/cuda/std/__simd/complex_math.h
fbusato: tested. There is no SASS differences.
I'm not really in favor of the modification. The proposed approach requires creating a new function, creating a `cuda::std::array` (value by value initialization is not possible), then inside the ctr there is a loop with a fixed num of iterations.
The generator approach is the conventional way to initialize a SIMD vector. Values are capture by reference (as lambda), everything is inlined, and we use `index_sequence` in the ctr. I would prefer this approach

libcudacxx/include/cuda/std/__simd/basic_vec.h
```diff
@@ -169,11 +174,18 @@ class basic_vec<_Tp, _Abi, enable_if_t<__is_vectorizable_v<_Tp> && __is_enabled_
...
   {}
 
+  template <typename _Up>
+  static inline constexpr bool __can_explicitly_convert_v =
+    __is_vec_ctor_explicit<_Up, value_type> && !__is_complex_real_type_v<value_type, _Up>;
+
+  template <typename _Up>
+  static inline constexpr bool __can_implicitly_convert_v =
+    !__is_vec_ctor_explicit<_Up, value_type> && !__is_complex_real_type_v<value_type, _Up>;
+
```
miscco, line 193: This is backwards, we should check if something is implicitly convertible. 

It is explicitly convertible if it is convertible but not implicit convertible, see how tuple does it

libcudacxx/include/cuda/std/__simd/exposition.h
```diff
@@ -35,6 +35,9 @@
 
 _CCCL_BEGIN_NAMESPACE_CUDA_STD_SIMD
 
+#if _CCCL_HOSTED()
+using ::std::complex;
+#endif // _CCCL_HOSTED()
```
miscco: Critical: we cannot pull in `::std::complex` into the whole `cuda::std::simd` namespace . PLease always spell it out

miscco (changes requested):

libcudacxx/include/cuda/std/__simd/exposition.h
fbusato: this is really bad :(

libcudacxx/include/cuda/std/__simd/exposition.h
```diff
@@ -31,19 +33,46 @@
 
 #include <cuda/std/__cccl/prologue.h>
 
+#if _CCCL_HOSTED()
+using ::std::complex;
+#endif // _CCCL_HOSTED()
```
miscco: This is still pulling in `::std::complex`

miscco (changes requested):

libcudacxx/include/cuda/std/__simd/exposition.h
fbusato: just realized that it is already declared in `cuda/std/__fwd/complex.h`

libcudacxx/include/cuda/std/__simd/exposition.h
```diff
@@ -38,12 +40,35 @@ _CCCL_BEGIN_NAMESPACE_CUDA_STD_SIMD
...
+
+template <typename _Tp>
+inline constexpr bool __is_complex_vectorizable_v<::cuda::complex<_Tp>> = true;
+
+#if _CCCL_HOSTED()
+
+template <typename _Tp>
+inline constexpr bool __is_complex_vectorizable_v<::std::complex<_Tp>> = true;
+
+#endif // _CCCL_HOSTED()
```
miscco: I believe this should rather be
```suggestion
#if _CCCL_HAS_HOST_STD_LIB()

template <typename _Tp>
inline constexpr bool __is_complex_vectorizable_v<::std::complex<_Tp>> = true;

#endif // _CCCL_HAS_HOST_STD_LIB()
```

libcudacxx/test/libcudacxx/std/numerics/simd/simd.traits/resize.pass.cpp
```diff
@@ -86,6 +87,10 @@ TEST_FUNC void test()
   test_resize_mask_all<char>();
   test_resize_mask_all<int>();
   test_resize_mask_all<double>();
+  test_resize_mask_all<cuda::std::complex<float>>();
+#if _CCCL_HAS_INT128()
+  test_resize_mask_all<cuda::std::complex<double>>();
+#endif // _CCCL_HAS_INT128()
```
miscco, line 100: Question: I do not see tests for `::std::complex` is that intended?

libcudacxx/test/libcudacxx/std/numerics/simd/simd.complex/complex_access.pass.cpp
```diff
@@ -0,0 +1,159 @@
...
+#include "test_macros.h"
+
+namespace simd = cuda::std::simd;
+
+//----------------------------------------------------------------------------------------------------------------------
+// complex constructor
+
+template <typename T, int N>
+TEST_FUNC constexpr void test_complex_ctor()
+{
```
miscco: I believe we should have one more indirection to test `::std::complex`

miscco: We can replace `T` with `{::cuda}::std::complex<T>` and get `T` back from its `value_type`

miscco (approved):

PR #8476, opened 2026-04-16
[libcu++] Add first, last and subspan to buffer

pciolkosz:
Working on an example I realized it would be nice to create those subspans directly from a buffer without a need to crate a span first.

I could see `first` and `last` to be a bit vague and maybe `subspan` without the others would be sufficient, but I decided to add all 3


miscco (approved):
Looks great thanks a lot

PR #8477, opened 2026-04-16
Implement `cuda::arch`

davebayer:
This PR implements `cuda::arch` type that could be used as the parameter passed to cub tunings.


cub/cub/device/dispatch/tuning/tuning_merge.cuh
```diff
@@ -105,7 +106,7 @@ struct policy_selector
         512, ipt_800_plus, LOAD_DEFAULT, BLOCK_STORE_WARP_TRANSPOSE, value_size == 0 ? use_bl2sh_keys : use_bl2sh_pairs};
     }
 
-    if (arch >= ::cuda::arch_id::sm_60)
+    if (arch.provides(::cuda::arch_id::sm_60))
```
davebayer, line 109: I think this looks good, @bernhardmgruber what do you think?

libcudacxx/include/cuda/__arch_
```diff
@@ -0,0 +1,29 @@
...
+// Part of libcu++, the C++ Standard Library for your entire system,
+// under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef _CUDA_ARCH
+#define _CUDA_ARCH
```
davebayer, line 12: I think it would be worth expose arch-only in a separate header, so the user needn't to include whole `<cuda/devices>` in case he just needs `cuda::arch` in his kernels.

libcudacxx/include/cuda/__arch/arch.h
```diff
@@ -0,0 +1,120 @@
...
+#include <cuda/std/__cccl/prologue.h>
+
+_CCCL_BEGIN_NAMESPACE_CUDA
+
+class arch
+{
+  arch_id __id_;
+
+public:
+  _CCCL_HIDE_FROM_ABI arch() = default;
```
miscco: Do we really want this to be uninitialized?

libcudacxx/include/cuda/__fwd/arch.h
```diff
@@ -0,0 +1,40 @@
...
+
+#include <cuda/std/__fwd/span.h>
+
+#include <cuda/std/__cccl/prologue.h>
+
+_CCCL_BEGIN_NAMESPACE_CUDA
+
+struct arch_traits_t;
+class compute_capability;
+enum class arch_id : int;
```
miscco:
```suggestion
struct arch_traits_t;

class compute_capability;

enum class arch_id : int;
```

libcudacxx/include/cuda/__arch_
miscco: I would prefer to keep this in a private `__arch_` for at least a release

davebayer: No problem with that!

libcudacxx/include/cuda/__arch/arch.h
davebayer: Aaah, I am not sure, making it `arch_id __id_{}` would probably be better

libcudacxx/include/cuda/__arch/arch.h
```diff
@@ -0,0 +1,120 @@
...
+#include <cuda/__arch/arch_id.h>
+#include <cuda/__arch/arch_traits.h>
+#include <cuda/__device/compute_capability.h>
+#include <cuda/__fwd/arch.h>
+
+#include <cuda/std/__cccl/prologue.h>
+
+_CCCL_BEGIN_NAMESPACE_CUDA
+
+class arch
```
pciolkosz, line 33: Seems to me that all of these things could be just part of arch_traits

PR #8484, opened 2026-04-16
[cccl.c] Split build step into compile + load

### LINKED ISSUE #8410: [cccl.c]: Add infrastructure to support ahead-of-time compilation (AoT)
## Background

Currently, each algorithm has a `cccl_device_<algo>_build` function that compiles a kernel via NVRTC and returns a build result struct holding the loaded CUlibrary and related state.

To support ahead-of-time (AoT) compilation — pre-compiling kernels and saving them to disk for use in a different process or on a different machine — the Python layer needs to serialize and deserialize these build result structs. This requires the C layer to expose enough metadata in the structs to make that possible.

## Required changes

  1. New fields in all `*_build_result_t` structs:
    - `int cc` — the compute capability the kernel was compiled for (encoded as major*10 + minor)
    - `size_t runtime_policy_size` — size of the opaque `runtime_policy` blob, so it can be round-tripped through serialization
    - Per-kernel char* lowered name fields — the mangled CUDA kernel names produced by NVRTC, needed to resolve kernels from a cubin via `cuLibraryGetKernel` during deserialization
  2. Cross-CC build support — when a kernel is compiled for a target CC that doesn't match the current device (e.g. compiling for SM 9.0 on an SM 8.6 machine), `cuLibraryLoadData` returns
  `CUDA_ERROR_NO_BINARY_FOR_GPU`. Currently this is a fatal error. The build functions should be updated to treat this case as success — returning the cubin and lowered names without a loaded CUlibrary — so that the
  result can be serialized and shipped to a matching device.

# Motivation

These changes are purely additive to the C structs and transparent to existing callers. They unblock the Python layer to implement `save()` / `load_algorithm()` for pre-compiled kernel distribution (e.g. shipping pre-compiled kernels in a Python wheel that works across a range of GPU architectures).

shwina:
## Description

Closes https://github.com/NVIDIA/cccl/issues/8410.

## Summary of changes

* This PR splits the existing `build` API of each algorithm into separate `compile` and `load` APIs. A third API called `link_ltoir` is also introduced (explained below).

* The `build` API doesn't go away - it just calls `compile` + `load`.

* The build struct of each algorithm is changed/extended:
  - Rather than always holding `cubin`, a build struct can now hold either a `cubin` or `ltoir`
  - The names of the kernels contained in the `cubin`/`ltoir` are included in the build struct

### Motivation - ahead-of-time compilation

The changes above are motivated by the need to do ahead-of-time compilation of CUB device-wide algorithms. Two AoT modes will be supported:

* **Full** - operator is known at compile-time
* **Partial** - only kernel is pre-compiled, operator is linked at runtime

### How it works

Here is how the three APIs `compile`, `link_ltoir` and `load` can be used together to support both AoT modes:

* `compile`: does the NVRTC compilation of the kernel. Depending on the operator provided, it can do one of two things:
  - Full compilation: if a valid operator is provided, compiles and links the kernel with the operator to produce a CUBIN
  - Partial compilation: if an empty operator is provided (contains no LTO-IR), compiles the kernel with an unresolved extern function - the result is kernel LTO-IR

* `link_ltoir`: useful in the partial compilation case - links the user provided LTO-IR with the kernel LTO-IR to produce a CUBIN

* `load`: Takes the cubin produced by `compile` or `link_ltoir` and calls `cuLibraryLoadData + cuLibraryGetKernel` for each kernel. This is the only step that requires a GPU. The lowered kernel names were previously required to be written in the build struct (in `compile`) so that they can be used here.

The flowchart below shows both the full AoT (left) and partial AoT (right) paths:

 ```mermaid
  flowchart TD
      A["kernel + operator"] --> B["compile()"] --> C["cubin"]
      D["kernel only"] --> E["compile()"] --> F["kernel LTOIR"] --> G["link_ltoir() + operator LTOIR"] --> C
      C --> H["load()"] --> I["execute"]
  ```

### Performance

The PR includes a test (which I suppose can be deleted later) that measures the speedup that can be expected from full and partial AoT:

```
$ ./build/cccl-c-parallel/bin/cccl.c.parallel.test.bench_aot

     --- AoT latency benchmark  (SM 89, N=1048576) ---
       algorithm                                JIT             link_ltoir               Full AoT
       --------------------------------------------------------------------------------------------
       reduce (int32)                    941.9 ms      252.076 ms (   4x)     1.799 ms ( 523x)
       scan (int32, excl.)              1337.6 ms      161.633 ms (   8x)     3.013 ms ( 444x)
       [merge_sort breakdown] link=343.375 ms  load=0.165 ms  exec+sync=2.072 ms (total=345.612 ms)
       merge_sort (int32, keys)          911.7 ms      345.612 ms (   3x)     1.918 ms ( 475x)
       --------------------------------------------------------------------------------------------
```

I note that, depending on the algorithm, the link step can be significantly expensive.

### Next steps

As a follow-up to this PR, we will provide functions for serializing and deserializing the result of the `compile` step to/from disk. This will complete the AoT support.

## Checklist


c/parallel/include/cccl/c/binary_search.h
```diff
@@ -29,6 +29,8 @@ typedef struct cccl_device_binary_search_build_result_t
   size_t cubin_size;
   CUlibrary library;
   CUkernel kernel;
+  // Lowered (mangled) kernel name, heap-allocated, freed by cccl_device_binary_search_cleanup():
```
shwina, line 32: Could do without these comments.

c/parallel/src/scan.cu
```diff
@@ -438,14 +435,43 @@ static_assert(device_scan_policy()(::cuda::arch_id{{CUB_PTX_ARCH / 10}}) == {6},
   build_ptr->init_kind                  = init_kind;
   build_ptr->description_bytes_per_tile = description_bytes_per_tile;
   build_ptr->payload_bytes_per_tile     = payload_bytes_per_tile;
-  build_ptr->runtime_policy             = new cub::detail::scan::policy_selector{policy_sel};
+  build_ptr->runtime_policy             = std::malloc(sizeof(cub::detail::scan::policy_selector));
+  build_ptr->runtime_policy_size        = sizeof(cub::detail::scan::policy_selector);
+  std::memcpy(build_ptr->runtime_policy, &policy_sel, sizeof(cub::detail::scan::policy_selector));
```
bernhardmgruber: Q: Why can't we use `new` to allocate the policy selector anymore?

shwina: We can keep the allocation/deallocation pattern the same. I undid this change.

c/parallel/src/binary_search.cu
```diff
@@ -241,12 +241,10 @@ void binary_search_kernel({1} d_data, OffsetT num_data, {3} d_values, OffsetT nu
...
-  cuLibraryLoadData(&build_ptr->library, result.data.get(), nullptr, nullptr, 0, nullptr, nullptr, 0);
-  check(cuLibraryGetKernel(&build_ptr->kernel, build_ptr->library, lowered_name.c_str()));
-
-  build_ptr->cc         = cc;
-  build_ptr->cubin      = (void*) result.data.release();
-  build_ptr->cubin_size = result.size;
+  build_ptr->cc                  = cc;
+  build_ptr->cubin               = (void*) result.data.release();
+  build_ptr->cubin_size          = result.size;
+  build_ptr->kernel_lowered_name = strdup(lowered_name.c_str());
```
NaderAlAwar, line 247: Important: we should not use `strdup` (applies to all other files in the PR) since it is POSIX, not standard C++. I would suggest something like

```C++
char* duplicate_c_string(std::string_view s)
  {
    auto p = std::make_unique<char[]>(s.size() + 1);
    std::memcpy(p.get(), s.data(), s.size());
    p[s.size()] = '\0';
    return p.release();
  }
```

and use that everywhere

c/parallel/src/binary_search.cu
```diff
@@ -328,7 +381,11 @@ try
   }
 
   std::unique_ptr<char[]> cubin(reinterpret_cast<char*>(build_ptr->cubin));
-  check(cuLibraryUnload(build_ptr->library));
+  std::free(build_ptr->kernel_lowered_name);
```
NaderAlAwar, line 384: Important: after replacing strdup, we should use something similar to the `std::unique_ptr<char[]>` pattern above

c/parallel/src/reduce.cu
```diff
@@ -18,8 +18,10 @@
 #include <cuda/std/functional> // ::cuda::std::identity
 #include <cuda/std/variant>
 
+#include <cstring> // strdup, free, memcpy
 #include <format>
 #include <memory>
+#include <new> // std::nothrow
```
NaderAlAwar, line 24: Suggestion: this seems like it's unused, should be removed.

NaderAlAwar (commented):
You did not make the same changes to `for`, is that intentional?

c/parallel/src/scan.cu
bernhardmgruber: Thx!

NaderAlAwar (approved):

c/parallel/src/util/nvjitlink.h
```diff
@@ -0,0 +1,85 @@
...
+#include <format>
+#include <memory>
+#include <utility>
+
+#include <nvrtc/nvjitlink_helper.h>
+#include <util/errors.h>
+
+// Links LTO-IR blobs via nvJitLink → returns SASS cubin (PTX fallback if SASS unavailable).
+// Caller owns the returned buffer.
+inline std::pair<std::unique_ptr<char[]>, size_t>
```
Jacobfaib:
```suggestion
[[nodiscard]] inline std::pair<std::unique_ptr<char[]>, size_t>
```
New API should generally try and use `[[nodiscard]]`

c/parallel/src/histogram.cu
```diff
@@ -388,18 +387,89 @@ static_assert(device_histogram_policy()(detail::current_tuning_cc()) == {4}, "Ho
   build_ptr->num_active_channels = num_active_channels;
   build_ptr->may_overflow = false; // This is set in cccl_device_histogram_even_impl so that kernel source can access
                                    // it later.
-  build_ptr->runtime_policy = new cub::detail::histogram::policy_selector{policy_sel};
+  build_ptr->runtime_policy            = new cub::detail::histogram::policy_selector{policy_sel};
+  build_ptr->runtime_policy_size       = sizeof(cub::detail::histogram::policy_selector);
+  build_ptr->init_kernel_lowered_name  = duplicate_c_string(init_kernel_lowered_name);
+  build_ptr->sweep_kernel_lowered_name = duplicate_c_string(sweep_kernel_lowered_name);
```
Jacobfaib: If any of these throw, `build_ptr->runtime_policy` will leak

c/parallel/src/merge_sort.cu
```diff
@@ -318,43 +323,159 @@ static_assert(device_merge_sort_policy()(detail::current_tuning_cc()) == {10}, "
...
-      ->get_name({merge_kernel_name, merge_kernel_lowered_name})
-      ->link_program()
-      ->add_link_list(linkable_list)
-      ->finalize_program();
+      ->get_name({merge_kernel_name, merge_kernel_lowered_name});
+
+  build_ptr->cc                             = cc.get();
+  build_ptr->key_type                       = input_keys_it.value_type;
+  build_ptr->item_type                      = input_items_it.value_type;
+  build_ptr->runtime_policy                 = new cub::detail::merge_sort::policy_selector{policy_sel};
```
Jacobfaib: Ditto with leak on error here. I might even move all of these to the complete bottom of the function if possible, or hold them in a temporary RAII-aware C++ object until the very end.

The way I've seen such C wrappers over C++ done in past projects is something like so:
```c++
struct c_object
{
	foo *ptr;
  bar mem;
};

// RAII-safe wrapper over the C version
class cpp_wrapper
{
  std::unique_ptr<foo> ptr;
  bar mem;

  c_object to_c_object()
  {
    c_object ret;

	  ret.foo = ptr.release();
    try {
      ret.mem = mem;
    } catch (...) {
      delete ret.foo;
      throw;
    }
    return ret;
  }
};

c_object frobnicate()
{
  try {
    cpp_wrapper wrap;

    // ...

    // Only at the last moment do we convert to the returned object
    return cpp_wrapper.to_c_object();
  } catch (...) {
    // report error
    return {};
  }
}
```
This helps keep the surface area of "dangerous" code quite low, because pointers are kept in RAII containers until the very last moment.

c/parallel/src/merge_sort.cu
```diff
@@ -318,43 +323,135 @@ static_assert(device_merge_sort_policy()(detail::current_tuning_cc()) == {10}, "
...
+    libcudacxx_path,
+    ctk_path,
+    config);
+  if (r != CUDA_SUCCESS)
+  {
+    return r;
+  }
+  if (build_ptr->cubin == nullptr)
+  {
+    return CUDA_SUCCESS;
```
oleksandr-pavlyk: Shouldn't this (call to compile did not result in meaningful `cubin`) be an error?

shwina:
@coderabbitai full review

c/parallel/include/cccl/c/three_way_partition.h
```diff
@@ -28,10 +28,16 @@ typedef struct cccl_device_three_way_partition_build_result_t
...
+  void* kernel_ltoir;
+  size_t kernel_ltoir_size;
   CUlibrary library;
   CUkernel three_way_partition_init_kernel;
   CUkernel three_way_partition_kernel;
   void* runtime_policy;
+  size_t runtime_policy_size;
+  // Lowered (mangled) kernel names, heap-allocated, freed by cccl_device_three_way_partition_cleanup():
+  char* three_way_partition_init_kernel_lowered_name;
+  char* three_way_partition_kernel_lowered_name;
```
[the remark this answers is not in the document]
shwina: I think it makes sense for the new members `kernel_ltoir` and `kernel_ltoir_size` to follow `cubin` and `cubin_size`. I could be convinced otherwise. Will let someone else on the team comment.

c/parallel/src/segmented_reduce.cu
```diff
@@ -239,50 +241,144 @@ static_assert(
...
+CUresult cccl_device_segmented_reduce_load(cccl_device_segmented_reduce_build_result_t* build_ptr)
+try
+{
+  if (build_ptr == nullptr || build_ptr->cubin == nullptr || build_ptr->cubin_size == 0)
+  {
+    return CUDA_ERROR_INVALID_VALUE;
+  }
+  check(cuLibraryLoadData(&build_ptr->library, build_ptr->cubin, nullptr, nullptr, 0, nullptr, nullptr, 0));
+  check(cuLibraryGetKernel(
+    &build_ptr->segmented_reduce_kernel, build_ptr->library, build_ptr->segmented_reduce_kernel_lowered_name));
```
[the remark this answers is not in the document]
shwina: build/build_ex aren't involved in the AoT workflow (that path is compile + link_ltoir + load). Am I understanding incorrectly?

c/parallel/include/cccl/c/three_way_partition.h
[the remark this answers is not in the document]
NaderAlAwar: I agree with Ashwin's point

c/parallel/src/segmented_sort.cu
```diff
@@ -859,3 +965,42 @@ catch (const std::exception& exc)
...
+  build_ptr->kernel_ltoir_size = 0;
+  build_ptr->cubin             = (void*) cubin.release();
+  build_ptr->cubin_size        = cubin_size;
+  return CUDA_SUCCESS;
+}
+catch (const std::exception& exc)
+{
+  printf("\nEXCEPTION in cccl_device_segmented_sort_link_ltoir(): %s\n", exc.what());
+  return CUDA_ERROR_UNKNOWN;
+}
```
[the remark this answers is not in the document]
shwina: No longer relevant.

c/parallel/src/three_way_partition.cu
```diff
@@ -261,41 +269,141 @@ static_assert(
...
+  return CUDA_SUCCESS;
+}
+catch (const std::exception& exc)
+{
+  fflush(stderr);
+  printf("\nEXCEPTION in cccl_device_three_way_partition_load(): %s\n", exc.what());
+  fflush(stdout);
+
+  return CUDA_ERROR_UNKNOWN;
+}
```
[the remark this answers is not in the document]
shwina: Fixed.

shwina:
@coderabbitai review

shwina:
@coderabbitai pr review

shwina:
@coderabbitai review

shwina:
@coderabbitai review

c/parallel/include/cccl/c/for.h
```diff
@@ -26,10 +26,12 @@ CCCL_C_EXTERN_C_BEGIN
...
   int cc;
-  void* cubin;
-  size_t cubin_size;
+  void* payload;
+  size_t payload_size;
+  cccl_payload_kind_t payload_kind;
   CUlibrary library;
   CUkernel static_kernel;
+  char* static_kernel_lowered_name;
 } cccl_device_for_build_result_t;
```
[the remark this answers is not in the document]
shwina: There's no `runtime_policy` for `for`.

c/parallel/test/test_bench_aot.cpp
shwina, line 1: We probably want to delete this test eventually.

c/parallel/src/reduce.cu
```diff
@@ -322,16 +326,20 @@ static_assert(device_reduce_nd_policy()(detail::current_tuning_cc()) == {10},
...
 
   appender.append_operation(op);
   appender.add_iterator_definition(input_it);
   appender.add_iterator_definition(output_it);
 
-  nvrtc_link_result result =
-    begin_linking_nvrtc_program(num_lto_args, lopts)
+  // kernel-only mode: extract kernel LTOIR without linking the operator in.
+  // Only custom ops (non-empty name) with no LTOIR trigger this; well-known ops (name="") do not.
+  const bool kernel_only = (op.code_size == 0) && (op.name != nullptr) && (op.name[0] != '\0');
```
NaderAlAwar: Suggestion: can we move this check to a common header, something like `is_custom_op`? That way we hide the details and the associated comments

c/parallel/src/reduce.cu
```diff
@@ -341,51 +349,144 @@ static_assert(device_reduce_nd_policy()(detail::current_tuning_cc()) == {10},
...
 }
 
+CUresult cccl_device_reduce_load(cccl_device_reduce_build_result_t* build)
+try
+{
+  if (build == nullptr || build->payload == nullptr || build->payload_size == 0
+      || build->payload_kind != CCCL_PAYLOAD_CUBIN || build->single_tile_kernel_lowered_name == nullptr
+      || build->single_tile_kernel_lowered_name[0] == '\0' || build->single_tile_second_kernel_lowered_name == nullptr
+      || build->single_tile_second_kernel_lowered_name[0] == '\0' || build->reduction_kernel_lowered_name == nullptr
+      || build->reduction_kernel_lowered_name[0] == '\0')
```
NaderAlAwar, line 437: Question: why do we check specifically for `'\0'`? Isn't a `nullptr` check enough?

c/parallel/include/cccl/c/reduce.h
NaderAlAwar, line 60: Question: now that we have the `_compile` API, is there a need to keep the old `_build` API around?

c/parallel/include/cccl/c/binary_search.h
```diff
@@ -71,6 +88,19 @@ CCCL_C_API CUresult cccl_device_binary_search(
   cccl_op_t op,
   CUstream stream);
 
+CCCL_C_API CUresult cccl_device_binary_search_link_ltoir(
```
NaderAlAwar, line 91: Critical: this function is not defined anywhere

python/cuda_cccl/cuda/compute/_bindings_impl.pyx
```diff
@@ -2421,16 +2421,19 @@ cdef class DeviceBinarySearchBuildResult:
             )
 
     def _get_cubin(self):
-        return _binary_search_cubin_bytes(&self.build_data)
+        return PyBytes_FromStringAndSize(
+            <const char*>self.build_data.transform.payload,
+            self.build_data.transform.payload_size
+        )
```
NaderAlAwar: Critical: this should keep using `_binary_serach_cubin_bytes()` because v1 nests `transform.payload`

c/parallel/src/reduce.cu
```diff
@@ -172,7 +176,7 @@ struct reduce_kernel_source
 struct reduce_iterator_tag;
 struct reduction_operation_tag;
 
-CUresult cccl_device_reduce_build_ex(
+CUresult cccl_device_reduce_compile(
```
NaderAlAwar, line 181: Suggestion: the new `_compile` API should probably zero initialize all the fields it does not set, such as the library and kernel that `_load` later sets. This was not needed for `_build` because it initialized all those fields

c/parallel/src/reduce.cu
```diff
@@ -341,51 +349,144 @@ static_assert(device_reduce_nd_policy()(detail::current_tuning_cc()) == {10},
...
 
   build->cc               = cc_major * 10 + cc_minor;
-  build->cubin            = (void*) result.data.release();
-  build->cubin_size       = result.size;
   build->accumulator_size = accum_t.size;
   build->determinism      = determinism;
   if (build_nondeterministic)
   {
-    build->runtime_policy = new cub::detail::reduce_nondeterministic::policy_selector{policy_sel_nd};
+    build->runtime_policy      = policy_nd.release();
```
NaderAlAwar: Important: we should not call `release()` here. Later in the function we do `get_program_ltoir` and `add_link_list`, which could throw, so we get memory leaks and `build` is only partially initialized. The right way to do this is to move the calls to `release()` to the end. This also applies to most of the other algorithms as well

c/parallel/src/reduce.cu
```diff
@@ -341,51 +349,144 @@ static_assert(device_reduce_nd_policy()(detail::current_tuning_cc()) == {10},
...
+  }
+  else
+  {
+    build->runtime_policy      = policy.release();
+    build->runtime_policy_size = sizeof(cub::detail::reduce::policy_selector);
+  }
+  build->single_tile_kernel_lowered_name        = single_tile_name.release();
+  build->single_tile_second_kernel_lowered_name = single_tile_second_name.release();
+  build->reduction_kernel_lowered_name          = reduction_name.release();
+  build->nondeterministic_kernel_lowered_name   = nondeterministic_name.release();
```
NaderAlAwar: Important: same as above regarding `release()`

c/parallel/src/reduce.cu
```diff
@@ -522,7 +623,18 @@ try
   {
     policy.reset(static_cast<cub::detail::reduce::policy_selector*>(build_ptr->runtime_policy));
   }
```
NaderAlAwar: Question: is `runtime_policy` intended to be deserialized as an opaque byte buffer using `runtime_policy_size`, or always reconstructed as a real `policy_selector` object? This typed cleanup is only safe for objects allocated with `new`; if AoT deserialization restores `runtime_policy` with raw allocation + byte copy, cleanup should use the same raw-storage ownership model as transform (`free`).

c/parallel/src/reduce.cu
```diff
@@ -565,3 +677,49 @@ CUresult cccl_device_reduce_build(
...
+  }
+  const int cc_major = build_ptr->cc / 10;
+  const int cc_minor = build_ptr->cc % 10;
+  std::vector<const void*> all_blobs;
+  std::vector<size_t> all_sizes;
+  if (build_ptr->payload != nullptr && build_ptr->payload_size > 0 && build_ptr->payload_kind == CCCL_PAYLOAD_LTOIR)
+  {
+    all_blobs.push_back(build_ptr->payload);
+    all_sizes.push_back(build_ptr->payload_size);
+  }
```
NaderAlAwar: Important: it makes more sense to just reject payloads that are not `CCCL_PAYLOAD_LTOIR`, so we can add something like this to the start of the function, then we don't need to check this condition here

```cpp
  if (build_ptr->payload == nullptr ||
      build_ptr->payload_size == 0 ||
      build_ptr->payload_kind != CCCL_PAYLOAD_LTOIR)
  {
    return CUDA_ERROR_INVALID_VALUE;
  }
```
This applies to other algorithms as well

c/parallel/src/reduce.cu
```diff
@@ -341,51 +349,144 @@ static_assert(device_reduce_nd_policy()(detail::current_tuning_cc()) == {10},
...
+  return CUDA_SUCCESS;
+}
+catch (const std::exception& exc)
+{
+  fflush(stderr);
+  printf("\nEXCEPTION in cccl_device_reduce_load(): %s\n", exc.what());
+  fflush(stdout);
+  return CUDA_ERROR_UNKNOWN;
+}
```
NaderAlAwar, line 472: Important: load error reporting is inconsistent. For example, scan does this

```cpp
  CUresult status =
    cuLibraryLoadData(&build_ptr->library, build_ptr->payload, nullptr, nullptr, 0, nullptr, nullptr, 0);
  if (status != CUDA_SUCCESS)
  {
    return status;
  }
```

but here we wrap the call to `cuLibraryLoadData` with `check()`, which throws and we set the error to `CUDA_ERROR_UNKNOWN`. I like the scan approach better because it gives potentially better error codes, e.g. `CUDA_ERROR_NO_BINARY_FOR_GPU`.

c/parallel/src/reduce.cu
shwina: Done in https://github.com/NVIDIA/cccl/pull/8484/changes/421874d32ed1959174f80821151b8bd4e8655015

c/parallel/src/reduce.cu
shwina: Done in https://github.com/NVIDIA/cccl/pull/8484/commits/010a83bc66dd59762deeffde97b36bd3e5d63044

c/parallel/src/reduce.cu
shwina: Done in https://github.com/NVIDIA/cccl/pull/8484/commits/9d76a67e00a9eb665e3948a306182820f6e01175

c/parallel/src/reduce.cu
shwina: Moved `release`s to the end in https://github.com/NVIDIA/cccl/pull/8484/commits/1588172e9aeb552150f105228cd5802fe8dc758f

c/parallel/src/reduce.cu
shwina: An empty string isn't valid here either.

c/parallel/include/cccl/c/reduce.h
shwina: Originally it was for convenience (less change on the Python side). But now, we also have `v2` that doesn't support `compile`/`load`. So it makes sense to keep around for a bit longer.

c/parallel/include/cccl/c/binary_search.h
shwina: Fixed in https://github.com/NVIDIA/cccl/pull/8484/commits/fb8083482d78b5d567766510763a28bed98d32fb.

python/cuda_cccl/cuda/compute/_bindings_impl.pyx
shwina: https://github.com/NVIDIA/cccl/pull/8484/changes/2eddaf60bafbe2021878e585d2d84a35d4a26c3a

c/parallel/src/reduce.cu
shwina: Part of https://github.com/NVIDIA/cccl/pull/8484/changes/1588172e9aeb552150f105228cd5802fe8dc758f

c/parallel/src/reduce.cu
shwina: Part of https://github.com/NVIDIA/cccl/pull/8484/changes/1588172e9aeb552150f105228cd5802fe8dc758f

c/parallel/src/binary_search.cu
NaderAlAwar, line 228: Important: this disables the LTOIR path for binary search. If the user originally passed in an op with `op.name != "" && op.code_size == 0` this will overwrite it.

c/parallel/src/reduce.cu
```diff
@@ -341,51 +350,159 @@ static_assert(device_reduce_nd_policy()(detail::current_tuning_cc()) == {10},
...
+  {
+    void operator()(void* p) const
+    {
+      std::free(p);
+    }
+  };
+  static_assert(::cuda::is_trivially_copyable_v<cub::detail::reduce::policy_selector>);
+  static_assert(::cuda::is_trivially_copyable_v<cub::detail::reduce_nondeterministic::policy_selector>);
+  const size_t policy_size = build_nondeterministic ? sizeof(policy_sel_nd) : sizeof(policy_sel);
+  std::unique_ptr<void, free_deleter> policy_ptr(std::malloc(policy_size));
```
NaderAlAwar, line 365: Important: `std::malloc` could return `nullptr`, we should check for this and abort return an error. Applies to all algorithms

c/parallel/src/three_way_partition.cu
```diff
@@ -246,10 +253,12 @@ static_assert(
 
   cccl::detail::extend_args_with_build_config(args, config);
 
+  const bool kernel_only = is_custom_op(select_first_part_op) && is_custom_op(select_second_part_op);
+
```
NaderAlAwar, line 261: Important: we should either support cases were only one of the ops was provided or explicitly reject those cases.

c/parallel/src/reduce.cu
```diff
@@ -341,51 +350,159 @@ static_assert(device_reduce_nd_policy()(detail::current_tuning_cc()) == {10},
...
+    cub_path,
+    thrust_path,
+    libcudacxx_path,
+    ctk_path,
+    config);
+  if (r != CUDA_SUCCESS)
+  {
+    return r;
+  }
+  return cccl_device_reduce_load(build);
```
NaderAlAwar: Important: prior to this PR, the cubin was automatically freed if loading in `build_ex` failed through RAII. Now, if load fails, we will get a memory leak. We should check the return code and cleanup for consistency. Something like this to make sure to return the load error code (instead of overwriting it):

```cpp
  r = cccl_device_reduce_load(build);
  if (r != CUDA_SUCCESS)
  {
    CUresult cleanup_r = cccl_device_reduce_cleanup(build);
    return cleanup_r == CUDA_SUCCESS ? r : cleanup_r;
  }

  return CUDA_SUCCESS;
```

NaderAlAwar (approved):

PR #8487, opened 2026-04-16
Enable hierarchy query tests for nvrtc

davebayer:


libcudacxx/test/libcudacxx/cuda/hierarchy/warp_level/hierarchy_queries.pass.cpp
```diff
@@ -280,6 +279,11 @@ void test()
 
 int main(int, char**)
 {
-  NV_IF_ELSE_TARGET(NV_IS_HOST, (test();), (test_device();))
+  NV_IF_ELSE_TARGET(NV_IS_HOST,
+                    ({
+                      cuda_thread_count = 32;
```
davebayer, line 284: @pciolkosz, I had to set the thread count to 32 to make this test pass, I'm not sure whether the assertion on block size being at least 32 for `gpu_thread.count(warp, hierarchy)` is a good thing..

pciolkosz (approved):

PR #8492, opened 2026-04-16
Replace `detail::adjacent_difference::dispatch` by CUB's public API

### LINKED ISSUE #7952: Use the new tuning API internally for `detail::adjacent_difference::dispatch`

bernhardmgruber:
- [x] Merge before: #8473

Fixes: #7952

### DIFF
--- a/thrust/thrust/system/cuda/detail/adjacent_difference.h
+++ b/thrust/thrust/system/cuda/detail/adjacent_difference.h
@@ -30,6 +30,7 @@
 #  include <thrust/type_traits/is_contiguous_iterator.h>
 #  include <thrust/type_traits/unwrap_contiguous_iterator.h>
 
+#  include <cuda/__memory/check_address.h>
 #  include <cuda/std/__functional/operations.h>
 #  include <cuda/std/__iterator/distance.h>
 #  include <cuda/std/__type_traits/is_pointer.h>
@@ -50,7 +51,7 @@ namespace cuda_cub
 {
 namespace __adjacent_difference
 {
-template <bool Comparable, class InputIt, class OutputIt, class BinaryOp>
+template <class InputIt, class OutputIt, class BinaryOp>
 cudaError_t THRUST_RUNTIME_FUNCTION doit_step(
   void* d_temp_storage,
   size_t& temp_storage_bytes,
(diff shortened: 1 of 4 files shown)


thrust/thrust/system/cuda/detail/adjacent_difference.h
```diff
@@ -65,49 +65,40 @@ cudaError_t THRUST_RUNTIME_FUNCTION doit_step(
...
+        (d_temp_storage, temp_storage_bytes, first, result, num_items_fixed, binary_op, stream));
+    }
+    return status;
+  }
+  else
+  {
+    cudaError_t status;
+    THRUST_INDEX_TYPE_DISPATCH(
+      status,
+      cub::DeviceAdjacentDifference::SubtractLeftCopy,
```
NaderAlAwar, line 97: Critical: Non-comparable iterator cases are now treated as non-overlapping and routed through `SubtractLeftCopy`, which changes existing Thrust behavior. `Comparable == false` only means this code cannot prove aliasing via direct pointer comparison; it does not mean input and output cannot refer to the same range. Before this change, Thrust used the conservative alias-safe path when it could not prove non-aliasing. After this change, previously supported in-place
`thrust::adjacent_difference(first, last, first, ...)` cases for fancy/random-access iterators that alias the same underlying storage but do not unwrap to comparable raw pointers can now hit the non-overlap API and fall into undefined behavior.

bernhardmgruber: You mean:
```suggestion
      cub::DeviceAdjacentDifference::SubtractLeft,
```
?

bernhardmgruber: Hmm, no. That doesn't compile. It seems Thrust exercising a use case that is not exposed in the CUB public API. We can either have an in-place `SubtractLeft` accepting a single iterator or a non-overlapping `SubtractLeftCopy` with two different iterators.

bernhardmgruber: I reverted to calling the dispatch directly again.

NaderAlAwar (approved):

miscco (approved):

miscco (approved):

PR #8493, opened 2026-04-16
Segmented scan use new tuning api

oleksandr-pavlyk:
## Description

This PR builds on top of #6712, but includes changes to use new tuning API and `policy_selector`.

closes 

## Checklist


cub/cub/device/dispatch/tuning/tuning_segmented_scan.cuh
```diff
@@ -26,10 +30,389 @@
...
+    return thread_segmented_scan_policy{policy_t::BLOCK_THREADS, policy_t::ITEMS_PER_THREAD, policy_t::load_modifier};
+  }
+};
+
+template <typename PolicyT>
+CUB_RUNTIME_FUNCTION thread_segmented_scan_policy_wrapper<PolicyT>
+make_thread_segmented_scan_policy_wrapper(PolicyT policy)
+{
+  return thread_segmented_scan_policy_wrapper<PolicyT>{policy};
+}
```
bernhardmgruber: This code should no longer be necessary, since you can just use the ordinary policies in host code.

cub/cub/device/dispatch/tuning/tuning_segmented_scan.cuh
```diff
@@ -71,6 +473,101 @@ CUB_RUNTIME_FUNCTION segmented_scan_policy_wrapper<PolicyT> make_segmented_scan_
   return segmented_scan_policy_wrapper<PolicyT>{policy};
 }
```
bernhardmgruber: Also those, should no longer be needed.

cub/cub/device/dispatch/dispatch_segmented_scan.cuh
```diff
@@ -66,126 +106,238 @@ struct device_segmented_scan_kernel_source
...
+        warp_policy::block_threads,
+        warp_policy::items_per_thread,
+        warp_policy::load_modifier,
+        warp_load,
+        warp_store,
+        warp_policy::max_segments_per_warp},
+      thread_segmented_scan_policy{
+        thread_policy::block_threads, thread_policy::items_per_thread, thread_policy::load_modifier}};
+  }
+};
```
bernhardmgruber: Should not be necessary, since segmented scan is not a public API. You can entirely replace any policy_hub by a policy_selector.

cub/cub/device/dispatch/dispatch_segmented_scan.cuh
```diff
@@ -66,126 +106,238 @@ struct device_segmented_scan_kernel_source
...
-  /// Offsets to beginning of each segment in the input sequence
-  BeginOffsetIteratorInputT d_input_begin_offsets;
+    if (d_temp_storage == nullptr)
+    {
+      temp_storage_bytes = 1;
+      return cudaSuccess;
+    }
 
-  /// Offsets to end of each segment in the input sequence
-  EndOffsetIteratorInputT d_input_end_offsets;
```
bernhardmgruber, line 134: If you move the `active_policy.CheckLoadModifier();` into the device code, you can omit the `dispatch_arch` entirely I think, which is simpler and compiles faster. We do the same in `DeviceScanKernel`.

cub/cub/device/dispatch/dispatch_segmented_scan.cuh
```diff
@@ -66,126 +106,238 @@ struct device_segmented_scan_kernel_source
...
-              d_input_begin_offsets,
-              d_input_end_offsets,
-              d_output_begin_offsets,
-              static_cast<OffsetT>(num_current_segments),
-              scan_op,
-              init_value);
+      switch (worker_choice)
+      {
+        case worker::block:
+          launcher.doit(
```
bernhardmgruber: Important: `doit` returns an error code that needs to be checked.

cub/cub/device/device_segmented_scan.cuh
```diff
@@ -187,26 +188,28 @@ struct DeviceSegmentedScan
...
-        d_in,
-        d_out,
-        num_segments,
-        d_in_begin_offsets,
-        d_in_end_offsets,
-        d_in_begin_offsets,
-        scan_op,
-        detail::InputValue<init_value_t>(init_value),
-        stream);
+      detail::InputValue<init_value_t>>(
```
bernhardmgruber: Suggestion: you can probably remove a lot of the explicit template arguments, since they can be deduced from the function arguments.

bernhardmgruber (commented):
Here is some quick feedback

cub/cub/device/dispatch/dispatch_segmented_scan.cuh
oleksandr-pavlyk: Moving it to device code makes sense. I would still need `dispatch_arch`, however, because of use of `active_policy.block`, `active_policy.warp`, and `active_policy.thread` to extract `workers_per_block, block_size` for active policy.

cub/cub/device/dispatch/tuning/tuning_segmented_scan.cuh
oleksandr-pavlyk: Great idea! Lots of lines of code removed

cub/cub/device/dispatch/tuning/tuning_segmented_scan.cuh
oleksandr-pavlyk: removed

cub/cub/device/dispatch/dispatch_segmented_scan.cuh
oleksandr-pavlyk: It was only used in `catch_test_device_segmented_scan_multi_segment.cu`. I changed the test, which calls into dispatch to test single-segment-per-worker vs. multiple-segments-per-worker to use policy selector and removed this code. 👍🏻

PR #8500, opened 2026-04-17
[cudax] Extend mapping result properties

davebayer:
This PR extends mapping result's properties by specifying:
- `MappingResult::is_always_exhaustive()`: a compile-time constant specifying whether a the mapping can produce threads that are not part of the group object (optional participation).
- `MappingResult::is_always_contiguous()`: a compile-time constant specifying whether the mapping produces a contiguous range of units (for example puts in a group threads `0` and `1`, but not `0` and `2`).

There properties are later used to optimize the group. queries/operations. I've made `cudax::group_by` templated on whether it may result in a non-exhaustive group by passing the `cudax::non_exhaustive` tag to the constructor. That way both static/dynamic versions of `cudax::group_by` can cut-off all excess threads that don't create a full group. The exhaustive version `assert`s on not creating a partial group.

I've written a proper test for `cudax::group_by`, too.

In next PR, I'd like to make the constructor take a base group instead of hierarchy + level parameters.

### DIFF
--- a/cudax/include/cuda/experimental/__hierarchy/mappings.cuh
+++ b/cudax/include/cuda/experimental/__hierarchy/mappings.cuh
@@ -36,13 +36,20 @@
 
 namespace cuda::experimental
 {
+struct non_exhaustive_t
+{
+  _CCCL_HIDE_FROM_ABI explicit non_exhaustive_t() = default;
+};
+
+_CCCL_DEVICE constexpr non_exhaustive_t non_exhaustive;
+
 // Requirements on mappings:
 // - must be copyable
 // - must implement `map(_Unit, _Level, _Hierarchy)` method that returns an object that satisfies the
 //   `__group_mapping_result` concept
 
 // todo(dabayer): do we want to add stride parameter?
-template <::cuda::std::size_t _Np = ::cuda::std::dynamic_extent>
+template <::cuda::std::size_t _Np, bool _IsExhaustive>
 class group_by
 {
   static_assert(_Np != 0, "_Np must not be zero");
@@ -56,6 +63,11 @@ public:
     unsigned __group_rank_;
     unsigned __rank_;
 
+    [[nodiscard]] _CCCL_DEVICE_API static constexpr __mapping_result __invalid(unsigned __group_count) noexcept
+    {
+      return {__group_count, 0xffff'ffffu, 0xffff'ffffu};
+    }
+
     [[nodiscard]] _CCCL_DEVICE_API static constexpr ::cuda::std::size_t static_group_count() noexcept
     {
       return _NGroups;
@@ -68,6 +80,10 @@ public:
 
     [[nodiscard]] _CCCL_DEVICE_API unsigned group_rank() const noexcept
     {
+      if constexpr (!_IsExhaustive)
+      {
+        _CCCL_ASSERT(is_valid(), "getting group rank of thread that is not part of the group is UB");
+      }
       return __group_rank_;
     }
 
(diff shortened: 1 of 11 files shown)


cudax/include/cuda/experimental/__hierarchy/mappings.cuh
```diff
@@ -36,13 +36,20 @@
 
 namespace cuda::experimental
 {
+struct non_exhaustive_t
+{
+  _CCCL_HIDE_FROM_ABI explicit non_exhaustive_t() = default;
+};
+
+_CCCL_GLOBAL_CONSTANT non_exhaustive_t non_exhaustive;
```
pciolkosz, line 44: Does this need to be global or is it enough to be constexpr device?

cudax/include/cuda/experimental/__hierarchy/mappings.cuh
```diff
@@ -86,48 +98,110 @@ public:
...
+      if constexpr (__static_nunits != ::cuda::std::dynamic_extent)
+      {
+        static_assert(__static_nunits % _Np == 0, "group_by mapping _IsExhaustive precondition violation");
+      }
+      else
+      {
+        _CCCL_ASSERT(__nunits % _Np == 0, "group_by mapping _IsExhaustive precondition violation");
+      }
+    }
+    else if (__nunits % _Np != 0)
```
pciolkosz, line 172: I would either constantly use count or _Np, its sometimes mixed now

cudax/include/cuda/experimental/__hierarchy/mappings.cuh
```diff
@@ -86,48 +98,110 @@ public:
...
+    {
+      --__ret.__group_count_;
```
pciolkosz, line 174: I am not sure we should subtract here, since we already done floor division

cudax/include/cuda/experimental/__hierarchy/mappings.cuh
```diff
@@ -56,6 +63,11 @@ public:
     unsigned __group_rank_;
     unsigned __rank_;
 
+    [[nodiscard]] _CCCL_DEVICE_API static constexpr __mapping_result __invalid() noexcept
+    {
+      return {0xffff'ffffu, 0xffff'ffffu, 0xffff'ffffu};
```
pciolkosz, line 68: I was wondering if for a non-participating thread it would make sense to keep the general group information like count and group count, but just have invalid rank?

cudax/include/cuda/experimental/__hierarchy/mappings.cuh
```diff
@@ -162,28 +258,81 @@ public:
...
+        return __mapping_result::__invalid();
+      }
+    }
     return __ret;
   }
 };
 
 _CCCL_HOST_DEVICE group_by(unsigned) -> group_by<::cuda::std::dynamic_extent>;
 
+_CCCL_HOST_DEVICE group_by(unsigned, non_exhaustive_t) -> group_by<::cuda::std::dynamic_extent, false>;
```
pciolkosz, line 316: I think the constructor is just device

pciolkosz (approved):

cudax/include/cuda/experimental/__hierarchy/mappings.cuh
davebayer: That's because `count()` returns unsigned, but `_Np` is `cuda::std::size_t`, so I want to avoid casting

cudax/include/cuda/experimental/__hierarchy/mappings.cuh
davebayer: Oh yeah, you are right, thanks

cudax/include/cuda/experimental/__hierarchy/mappings.cuh
davebayer: I was also thinking about it. But the only thing that makes sense is group count, because all of the other values are specific to the group the unit is in.

I've made the them queryable for group count

cudax/include/cuda/experimental/__hierarchy/mappings.cuh
davebayer: `__device__ constexpr` is fine

PR #8505, opened 2026-04-17
Prepare testing in tile mode for libcu++

miscco:
tile programing has been anounced and we want to be ready for the user to use libcu++ features in tile mode

This adds basic functionality to build libcu++ in tile mode once the next CTK is actually released.

Note that there are a number of additional changes needed to make some features work and that some features are not going to be suppported in the future


libcudacxx/include/cuda/std/__cccl/compiler.h
```diff
@@ -111,6 +111,12 @@
 #  define _CCCL_CUDA_COMPILATION() 0
 #endif // ^^^ not compiling .cu file ^^^
 
+#ifdef __CUDACC_TILE__
```
davebayer, line 114:
```suggestion
#if defined(__CUDACC_TILE__)
```
I think we always use `defined`, I don't even know why :D

libcudacxx/include/cuda/std/__cccl/execution_space.h
```diff
@@ -28,10 +28,16 @@
...
+#    define _CCCL_TILE __tile__
+#  else // ^^^ __CUDACC_TILE__ ^^^ / vvv !__CUDACC_TILE__ vvv
+#    define _CCCL_TILE
+#  endif // !__CUDACC_TILE__
 #else // ^^^ _CCCL_CUDA_COMPILATION ^^^ / vvv !_CCCL_CUDA_COMPILATION vvv
 #  define _CCCL_HOST
 #  define _CCCL_DEVICE
 #  define _CCCL_HOST_DEVICE
+#  define _CCCL_TILE
 #endif // !_CCCL_CUDA_COMPILATION
```
davebayer, line 42:
```suggestion
#else // ^^^ _CCCL_CUDA_COMPILATION ^^^ / vvv !_CCCL_CUDA_COMPILATION vvv
#  define _CCCL_HOST
#  define _CCCL_DEVICE
#  define _CCCL_HOST_DEVICE
#endif // !_CCCL_CUDA_COMPILATION

#if _CCCL_TILE_COMPILATION()
#  define _CCCL_TILE __tile__
#else // ^^^ _CCCL_TILE_COMPILATION() ^^^ / vvv !_CCCL_TILE_COMPILATION() vvv
#  define _CCCL_TILE
#endif // ^^^ !_CCCL_TILE_COMPILATION() ^^^
```

CMakeLists.txt
```diff
@@ -62,6 +62,23 @@ else()
...
 endif()
 
+option(CCCL_ENABLE_TILE "Enable tile support" OFF)
+if (CCCL_ENABLE_TILE) # TODO(miscco) sccache currently does not support tile
+  if (CMAKE_CUDA_COMPILER_VERSION VERSION_LESS_EQUAL 13.2)
+    message(
+      FATAL_ERROR
+      "tile programs are not supported in NVCC until after 13.2"
+    )
+  endif()
```
davebayer, line 72: We should check that the CUDA compiler is `nvcc`

fbusato: also, I would suggest the opposite condition `VERSION_GREATER_EQUAL 13.3`

libcudacxx/include/cuda/std/__cccl/compiler.h
miscco: I believe there will be a lot of clang-tidy issues because that prefers ifdef when possible

CMakeLists.txt
miscco: but we do not know the next version of NVCC

libcudacxx/include/cuda/std/__cccl/compiler.h
davebayer: lol, okay

libcudacxx/test/libcudacxx/CMakeLists.txt
```diff
@@ -120,6 +120,14 @@ string(
   " -DLIBCUDACXX_IGNORE_DEPRECATED_ABI"
 )
 
+# enable tile support
+if (CCCL_ENABLE_TILE)
+  string(APPEND LIBCUDACXX_TEST_COMPILER_FLAGS " --enable-tile")
+  set(LIBCUDACXX_ENABLE_TILE True)
+else()
+  set(LIBCUDACXX_ENABLE_TILE False)
```
bernhardmgruber, line 128: Q: Why do we need the extra `LIBCUDACXX_ENABLE_TILE`? Can we just use `CCCL_ENABLE_TILE` all the way?

libcudacxx/test/CMakeLists.txt
```diff
@@ -60,6 +60,16 @@ if ("${CMAKE_CUDA_COMPILER_ID}" STREQUAL "NVIDIA")
...
     )
   endif()
+
+  if (
+    CCCL_ENABLE_TILE
+    AND "${CMAKE_CUDA_COMPILER_VERSION}" VERSION_LESS_EQUAL "13.2"
+  )
+    message(
+      FATAL_ERROR
+      "tile programs are not supported in NVCC until after 13.2 but found NVCC ${CMAKE_CUDA_COMPILER_VERSION}"
```
bernhardmgruber, line 70: I needed 3 attempts to parse this sentence. Can we do:
```suggestion
      "tile programs are not supported in NVCC <= 13.2. NVCC version is ${CMAKE_CUDA_COMPILER_VERSION}"
```

bernhardmgruber (approved):
PR is fine actually

libcudacxx/test/libcudacxx/CMakeLists.txt
miscco: The issue is that in our cmake setup an option is set to `OFF`

The lit machinery only understands `True, False`

PR #8522, opened 2026-04-20
[CUB] Delegate large segments to Multi-CTA Implementation in `DeviceSegmentedTopK`

### LINKED ISSUE #8364: Delegate large segments to Multi-CTA Implementation in `DeviceSegmentedTopK`
The goal here is to extend the existing kernel, which processes a batch of small segments, such that large segments that cannot be processed by a thread block are pushed into a global queue that will get picked up by a subsequent kernel that assigns a (presumably load-balanced) number of CTAs to each of those segments. 

**Tasks:**
- [ ] Get acquainted with the current `agent_batched_topk_worker_per_segment` that currently assumes one-cta-per-segment
- [ ] Allocate memory for a queue that will hold the segment indices of the "large" segments
- [ ] Allocate memory for an atomic counter that's used to coordinate writes to such queue
- [ ] Fuse scan over queue to be done by the last CTA (determined via atomic counter)
- [ ] ...

_Depends on:_
- [ ] Interdependent co-design with #8360

pauleonix:
## Description

This PR just handles the preparation for the multi-worker/large-segment agent/kernel that @elstehle is working on.

It implements a queue for the existing single-worker agent to queue tiles that are too large for it to handle and adds an epilogue that does a plus-scan on the segment sizes (in units of tiles of the multi-worker agent) from that queue. The resulting offsets from this fused scan will be used by the multi-worker agent for load-balancing-search.

In case we know that all items are too big for the single-worker agent, we avoid launching it and directly dispatch a SOL device scan for the load-balancing search offsets.

Due to the partial nature of this PR, the large segment handling (ids and offsets) can't really be tested yet.

closes #8364 

## Checklist

### DIFF
--- a/cub/cub/agent/agent_batched_topk.cuh
+++ b/cub/cub/agent/agent_batched_topk.cuh
@@ -1,4 +1,4 @@
-// SPDX-FileCopyrightText: Copyright (c) 2025, NVIDIA CORPORATION. All rights reserved.
+// SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 #pragma once
--- a/cub/cub/agent/agent_batched_topk.cuh
+++ b/cub/cub/agent/agent_batched_topk.cuh
@@ -23,11 +23,26 @@
 #include <cub/util_type.cuh>
 
 #include <cuda/__cmath/ceil_div.h>
+#include <cuda/std/cstdint>
 
 CUB_NAMESPACE_BEGIN
 
 namespace detail::batched_topk
 {
+// Atomic counters used by the small-segment kernel to (a) enqueue large segments into the large-segment work queue
+// and (b) elect the last block to run the epilogue scan over the queued tile counts. `alignas(128)` isolates each
+// counter on its own cache line for performance.
+struct alignas(128) batched_topk_counters
+{
+  // Number of segments enqueued in the large-segment work queue. Atomically incremented (by 1) by the first thread
+  // of each block that decides its segment is large.
+  unsigned long long large_segments_count;
+
+  // Block retirement counter. Each block atomically increments by 1 when it has finished processing its segment, and
+  // the block that observes `gridDim.x - 1` runs the epilogue on the queued large segments tile counts.
+  alignas(128) unsigned retirement_count;
+};
+
 template <typename PolicyGetter, // TODO(bgruber): pass worker_policy as NTTP in C++20
           typename KeyInputItItT,
           typename KeyOutputItItT,
@@ -54,13 +69,12 @@ struct agent_batched_topk_worker_per_segment
 
   static constexpr worker_policy active_policy = PolicyGetter{}();
 
-  static constexpr int block_threads             = active_policy.block_threads;
-  static constexpr int items_per_thread          = active_policy.items_per_thread;
-  static constexpr int tile_size                 = block_threads * items_per_thread;
-  static constexpr int epilogue_items_per_thread = active_policy.epilogue_items_per_thread;
-  static constexpr int epilogue_tile_size        = block_threads * epilogue_items_per_thread;
-  // TODO (elstehle): This is just a placeholder for now, we will need to specialize this for the large segment agent.
-  static constexpr int large_segment_agent_tile_size = tile_size;
+  static constexpr int block_threads                 = active_policy.block_threads;
+  static constexpr int items_per_thread              = active_policy.items_per_thread;
+  static constexpr int tile_size                     = block_threads * items_per_thread;
+  static constexpr int epilogue_items_per_thread     = active_policy.epilogue.items_per_thread;
+  static constexpr int epilogue_tile_size            = block_threads * epilogue_items_per_thread;
+  static constexpr int large_segment_agent_tile_size = active_policy.epilogue.multi_worker_tile_size;
 
   static constexpr bool only_small_segments = params::static_max_value_v<SegmentSizeParameterT> <= tile_size;
 
@@ -81,10 +95,10 @@ struct agent_batched_topk_worker_per_segment
   using block_store_vals_t = BlockStore<value_t, block_threads, items_per_thread, active_policy.store_algorithm>;
 
   using block_load_epilogue_t =
-    BlockLoad<segment_size_val_t, block_threads, epilogue_items_per_thread, active_policy.epilogue_load_algorithm>;
-  using block_scan_epilogue_t = BlockScan<int, block_threads, active_policy.epilogue_scan_algorithm>;
+    BlockLoad<segment_size_val_t, block_threads, epilogue_items_per_thread, active_policy.epilogue.load_algorithm>;
+  using block_scan_epilogue_t = BlockScan<int, block_threads, active_policy.epilogue.scan_algorithm>;
   using block_store_epilogue_t =
-    BlockStore<segment_size_val_t, block_threads, epilogue_items_per_thread, active_policy.epilogue_store_algorithm>;
+    BlockStore<segment_size_val_t, block_threads, epilogue_items_per_thread, active_policy.epilogue.store_algorithm>;
 
   // -------------------------------------------------------------------------
   // Shared Memory Storage
@@ -118,9 +132,7 @@ struct agent_batched_topk_worker_per_segment
   KParameterT k_param;
   SelectDirectionParameterT select_directions;
   NumSegmentsParameterT num_segments;
-  num_segments_val_t* d_large_segments_counter;
-  // Currently we use int for segment_id = blockIdx.x, so int has be enough.
-  ::cuda::std::uint32_t* d_retirement_counter;
+  batched_topk_counters* d_counters;
   num_segments_val_t* d_large_segments_ids;
   // Assuming the large segment agent will also use int for tile_id = blockIdx.x.
   ::cuda::std::int64_t* d_large_segments_tile_offsets;
@@ -137,8 +149,7 @@ struct agent_batched_topk_worker_per_segment
     KParameterT k_param,
     SelectDirectionParameterT select_directions,
     NumSegmentsParameterT num_segments,
-    num_segments_val_t* d_large_segments_counter,
-    ::cuda::std::uint32_t* d_retirement_counter,
+    batched_topk_counters* d_counters,
     num_segments_val_t* d_large_segments_ids,
     ::cuda::std::int64_t* d_large_segments_tile_offsets)
       : temp_storage(temp_storage.Alias())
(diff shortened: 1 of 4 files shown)


cub/cub/agent/agent_batched_topk.cuh
```diff
@@ -150,107 +191,179 @@ struct agent_batched_topk_worker_per_segment
...
+    _CCCL_PRAGMA_NOUNROLL()
+    for (int large_segment_offset = 0; large_segment_offset < num_large_segments;
+         large_segment_offset += epilogue_tile_size)
     {
+      segment_size_val_t segment_sizes[epilogue_items_per_thread];
+      block_load_epilogue_t(temp_storage.load_epilogue)
+        .Load(
+          d_large_segments_sizes + large_segment_offset, segment_sizes, num_large_segments - large_segment_offset, 0);
+      __syncthreads();
+      block_scan_epilogue_t(temp_storage.scan_epilogue).ExclusiveSum(segment_sizes, segment_sizes, prefix_callback_op);
```
pauleonix: Instead of segment size, scan over the number of CTAs needed for each segment.

cub/cub/agent/agent_batched_topk.cuh
```diff
@@ -150,107 +191,179 @@ struct agent_batched_topk_worker_per_segment
...
+      if constexpr (!is_keys_only)
+      {
+        __syncthreads();
+        auto block_vals_out = d_value_segments_out_it[segment_id];
+
+        block_store_vals_t(temp_storage.store_vals).Store(block_vals_out, thread_values, k);
+      }
+    }
+
+    bool is_last_block = false;
```
pauleonix: Do not compile when there are no large segments.

cub/cub/agent/agent_batched_topk.cuh
```diff
@@ -108,7 +122,16 @@ struct agent_batched_topk_worker_per_segment
...
   NumSegmentsParameterT num_segments;
-
+  // Currently we use int for segment_id, so int has be enough for these atomic counters.
+  int* d_retirement_count;
+  int* d_large_segment_queue;
+  key_it_t* d_key_large_segments_it;
+  key_it_t* d_key_large_segments_out_it;
+  value_it_t* d_value_large_segments_it;
+  value_it_t* d_value_large_segments_out_it;
+  [[maybe_unused]] segment_size_val_t* d_large_segments_sizes;
```
pauleonix: `d_large_segment_offsets` for scan results always needed.

cub/cub/agent/agent_batched_topk.cuh
```diff
@@ -108,7 +122,16 @@ struct agent_batched_topk_worker_per_segment
...
   SelectDirectionParameterT select_directions;
   NumSegmentsParameterT num_segments;
-
+  // Currently we use int for segment_id, so int has be enough for these atomic counters.
+  int* d_retirement_count;
+  int* d_large_segment_queue;
+  key_it_t* d_key_large_segments_it;
+  key_it_t* d_key_large_segments_out_it;
+  value_it_t* d_value_large_segments_it;
+  value_it_t* d_value_large_segments_out_it;
```
pauleonix: Only write out large segment ids.

cub/cub/agent/agent_batched_topk.cuh
```diff
@@ -22,10 +22,27 @@
...
 namespace detail::batched_topk
 {
+// Atomic counters used by the small-segment kernel to (a) enqueue large segments into the large-segment work queue
+// and (b) elect the last block to run the epilogue scan over the queued tile counts. `alignas(128)` isolates each
+// counter on its own cache line for performance.
+struct alignas(128) batched_topk_counters
+{
+  // Number of segments enqueued in the large-segment work queue. Atomically incremented (by 1) by the first thread
+  // of each block that decides its segment is large.
+  unsigned long long large_segments_count;
```
bernhardmgruber: Remark: I would have intuitively written it like:

```suggestion
struct batched_topk_counters
{
  // Number of segments enqueued in the large-segment work queue. Atomically incremented (by 1) by the first thread
  // of each block that decides its segment is large.
  alignas(128) unsigned long long large_segments_count;
```
since we care about the alignment of the integer and not the struct.

cub/cub/device/dispatch/tuning/tuning_batched_topk.cuh
```diff
@@ -23,13 +24,27 @@
...
+struct epilogue_policy
+{
+  int items_per_thread;
+  // TODO (elstehle): worker-agent needs to know tile size of multi-worker agent but it isn't a tuning parameter for the
+  // worker-agent. Depending on how the tuning for the multi-worker agent is implemented, the way this is passed to the
+  // worker-agent might need to be changed.
+  int multi_worker_tile_size;
+  BlockLoadAlgorithm load_algorithm;
+  BlockScanAlgorithm scan_algorithm;
+  BlockStoreAlgorithm store_algorithm;
```
bernhardmgruber: Important: Please swap the two members for consistency with other tuning policies. See also: https://docs.google.com/spreadsheets/d/1EUdwvmv2n7-WRp30-lt7CQxyu8r6atdfJJx3O6Pf1Ag

pauleonix: @elstehle Should we rename `worker_policy` to `medium_segment`? That would be in tune with `segmented_sort_policy`'s sub-policies.

pauleonix: On the other hand we might not want to have "segment" in the name because this is also for the batched API.

cub/cub/device/dispatch/tuning/tuning_batched_topk.cuh
```diff
@@ -23,13 +24,27 @@
 CUB_NAMESPACE_BEGIN
 namespace detail::batched_topk
 {
+struct epilogue_policy
+{
+  int items_per_thread;
+  // TODO (elstehle): worker-agent needs to know tile size of multi-worker agent but it isn't a tuning parameter for the
+  // worker-agent. Depending on how the tuning for the multi-worker agent is implemented, the way this is passed to the
+  // worker-agent might need to be changed.
+  int multi_worker_tile_size;
```
pauleonix: Add wrapping policy including large segment agent tuning parameters.

cub/cub/device/dispatch/tuning/tuning_batched_topk.cuh
elstehle: I think I would generally be ok to call something "segment", even in the batched context, as I think it's more intuitive to describe a "sub-problem of a batch". I simply cannot find a good term that would be similarly intuitive (e.g., "sub-problem", "task", "range", "buffer", ... 🤷).

As for the specific question on naming `worker_policy`. I don't have a preference or answer to that yet. The whole tuning parameter space here is still in flux. I'm ok to call it medium, but I don't want to commit to having that public yet.

cub/cub/agent/agent_batched_topk.cuh
```diff
@@ -140,107 +195,171 @@ struct agent_batched_topk_worker_per_segment
...
+        return;
+      }
+      const auto num_large_segments = d_counters->large_segments_count;
+      // For tracking the running total across tiles (loop iterations).
+      const auto prefix_callback_op =
+        [running_total = segment_size_val_t{0}](segment_size_val_t block_aggregate) mutable {
+          auto old_running_total = running_total;
+          running_total += block_aggregate;
+          return old_running_total;
+        };
```
elstehle, line 349: This is an interesting approach, it hadn't crossed my mind before. Would it be worth pointing out that this gets invoked by the full warp (to make sure we don't mess up with a future change) and only the return value of lane 0 is applied as the block prefix?

> The functor will be invoked by the first warp of threads in the block, however only the return value from lane0 is applied as the block-wide prefix. Can be stateful.

cub/cub/agent/agent_batched_topk.cuh
```diff
@@ -17,15 +17,34 @@
...
+// Atomic counters used by the small-segment kernel to (a) enqueue large segments into the large-segment work queue
+// and (b) elect the last block to run the epilogue scan over the queued tile counts. `alignas(128)` isolates each
+// counter on its own cache line for performance.
+template <class NumTilesT>
+struct batched_topk_counters
+{
+  using tile_count_t = detail::choose_offset_t<NumTilesT>;
+  // Number of segments enqueued in the large-segment work queue. Atomically incremented (by 1) by the first thread
+  // of each block that decides its segment is large.
+  alignas(128) tile_count_t large_segments_count;
```
elstehle: This one is actually counting the number segments, right. So, it should actually be like a SegmentIndexT?

cub/cub/agent/agent_batched_topk.cuh
```diff
@@ -17,15 +17,34 @@
...
+#include <cuda/__cmath/ceil_div.h>
+
 CUB_NAMESPACE_BEGIN
 
 namespace detail::batched_topk
 {
+// Atomic counters used by the small-segment kernel to (a) enqueue large segments into the large-segment work queue
+// and (b) elect the last block to run the epilogue scan over the queued tile counts. `alignas(128)` isolates each
+// counter on its own cache line for performance.
+template <class NumTilesT>
```
elstehle: It seems we are passing in the type used to index into segments here, which would also better fit its use for `large_segments_count`. So we probably want to rename this appropriately? Or, asked differently, what was the motivation to call this `NumTiles`.

cub/cub/agent/agent_batched_topk.cuh
```diff
@@ -17,15 +17,34 @@
...
+struct batched_topk_counters
+{
+  using tile_count_t = detail::choose_offset_t<NumTilesT>;
+  // Number of segments enqueued in the large-segment work queue. Atomically incremented (by 1) by the first thread
+  // of each block that decides its segment is large.
+  alignas(128) tile_count_t large_segments_count;
+
+  // Block retirement counter. Each block atomically increments by 1 when it has finished processing its segment, and
+  // the block that observes `gridDim.x - 1` runs the epilogue on the queued large segments tile counts.
+  alignas(128) unsigned retirement_count;
```
elstehle, line 50: This one does count the number of thread blocks. I think having this fixed to 32-bit type should be ok, and we can take special care, like having a grid-stride loop or streaming kernel invocations, if we wanted to support more segments than `INT_MAX`. But maybe, let's leave a comment so future AI agents have context when debugging :-)

elstehle (approved):

cub/cub/device/dispatch/tuning/tuning_batched_topk.cuh
pauleonix: Then let's leave it as is for now.

cub/cub/agent/agent_batched_topk.cuh
pauleonix: ~~The point is that the id can go up to `num_segments - 1` so it should use the same type as `num_segments`. I assumed that that type will be customizable in the future (currently it is hardcoded to 64bit in the guarantees interface).~~ I see what you mean.

cub/cub/agent/agent_batched_topk.cuh
pauleonix: It comes from a snippet in the docs 😆 But I think I only went with it b/c `ExclusiveSum()` has no overload taking an initial value. I was thinking about taking the easier to read and potentially cheaper to compile way and use `ExclusiveScan()` with initial value and aggregate. But there the aggregate is broadcasted which could cost extra, so I will leave it this way.

PR #8523, opened 2026-04-20
[cudax] Implement public groups synchronizers

### LINKED ISSUE #8516: [FEA]: Implement `cudax::barrier_synchronizer` synchronizer
I'd like to have a public barrier synchronizer:
```cpp
// creates a barrier synchronizer that uses the specified barriers for synchronization
cudax::barrier_synchronizer synchronizer{/*span-of-barriers*/};
```

### LINKED ISSUE #8517: [FEA]: Implement `cudax::lane_synchronizer` synchronizer
I'd like to have a synchronizer that uses `__syncwarp(/*mask*/)` for thread synchronization inside a warp:
```cpp
cudax::lane_synchronizer synchronizer{};
```

davebayer:
Closes #8516 and #8517.

### DIFF
--- a/cudax/include/cuda/experimental/__group/synchronizer/lane_synchronizer.cuh
+++ b/cudax/include/cuda/experimental/__group/synchronizer/lane_synchronizer.cuh
@@ -1,0 +1,99 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of CUDA Experimental in CUDA C++ Core Libraries,
+// under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef _CUDA_EXPERIMENTAL___GROUP_SYNCHRONIZER_LANE_SYNCHRONIZER_CUH
+#define _CUDA_EXPERIMENTAL___GROUP_SYNCHRONIZER_LANE_SYNCHRONIZER_CUH
+
+#include <cuda/std/detail/__config>
+
+#if defined(_CCCL_IMPLICIT_SYSTEM_HEADER_GCC)
+#  pragma GCC system_header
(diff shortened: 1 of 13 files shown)


cudax/include/cuda/experimental/__group/synchronizer/lane_synchronizer.cuh
```diff
@@ -0,0 +1,99 @@
...
+    }
+    else
+    {
+      _CCCL_ASSERT(__is_supported_count<_Level>(__mapping_result.count()),
+                   "unsupported count for cuda::lane_synchronizer");
+    }
+
+    const auto __mask  = (1u << __mapping_result.count()) - 1;
+    const auto __shift = (__mapping_result.group_rank() * __mapping_result.count()) % 32;
+    return {__mask << __shift};
```
pciolkosz, line 90: Should this assert the mapping is contiguous?

cudax/test/group/synchronizer/barrier_synchronizer.cu
```diff
@@ -0,0 +1,127 @@
...
+#include <cuda/std/cstddef>
+#include <cuda/std/type_traits>
+#include <cuda/stream>
+
+#include <cuda/experimental/group.cuh>
+
+#include "testing.cuh"
+
+template <class Level, class Config>
+__device__ void test_barrier_synchronizer(const Level& level, Config config)
```
pciolkosz, line 24: Should the synchronizer test unit test actual synchronization of threads not through a group?

pciolkosz (approved):

cudax/include/cuda/experimental/__group/synchronizer/lane_synchronizer.cuh
davebayer: It takes just group_by mapping for now which is always contiguous. This is a temporary solution anyway, I need to reiterate over this and come up with some clear constraints on when this synchronizer is allowed.

For example when we are grouping threads in a warp, the mapping even needn't to be contiguous

cudax/test/group/synchronizer/barrier_synchronizer.cu
davebayer: Probably yes. It is a bit more complicated, because in next PR, the `make_instance(...)` takes the parent group as a parameter, so we are basically replicating the whole behaviour of the group constructor.

My plan is to extend the tests later, now I just need the most basic stuff to work

PR #8524, opened 2026-04-20
Fix `<cuda/barrier>` `clang-tidy` warnings

davebayer:
I've came across some `clang-tidy` warnings when including and instantiating types from `<cuda/barrier>`. Not sure why the `typedef` warning hasn't appeared earlier.

### DIFF
--- a/libcudacxx/include/cuda/barrier
+++ b/libcudacxx/include/cuda/barrier
@@ -46,7 +46,7 @@
 // We need to forward-declare both CUtensorMap_st (the struct) and CUtensorMap
 // (the typedef):
 struct CUtensorMap_st;
-typedef struct CUtensorMap_st CUtensorMap;
+typedef struct CUtensorMap_st CUtensorMap; // NOLINT(modernize-use-using)
 
 #include <cuda/std/__cccl/prologue.h>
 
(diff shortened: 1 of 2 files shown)


libcudacxx/include/cuda/barrier
```diff
@@ -46,7 +46,7 @@
 // We need to forward-declare both CUtensorMap_st (the struct) and CUtensorMap
 // (the typedef):
 struct CUtensorMap_st;
-typedef struct CUtensorMap_st CUtensorMap;
+typedef struct CUtensorMap_st CUtensorMap; // NOLINT(modernize-use-using)
```
miscco, line 49: Couldnt this just be 
```suggestion
using CUtensorMap = CUtensorMap_st;
```

We just forward declared in the line above

davebayer: Oh, it could definitely be defined like that, I'm just not sure what happens when `<cuda.h>` defines the type alias with `typedef` syntax, I'm afraid compilers will go crazy about it

miscco: Fine by me

miscco (approved):

PR #8526, opened 2026-04-20
Use environment based CUB APIs when possible in PSTL

miscco:
Some of our PSTL algorithms can use the new environment based CUB APIs

This gives us the advantage that the user can pass tunings / guarantees along with the execution policy.

### DIFF
--- a/libcudacxx/include/cuda/std/__pstl/cuda/copy_if.h
+++ b/libcudacxx/include/cuda/std/__pstl/cuda/copy_if.h
@@ -71,7 +71,7 @@ struct __pstl_dispatch<__pstl_algorithm::__copy_if, __execution_backend::__cuda>
     using _OffsetType = iter_difference_t<_InputIterator>;
     _OffsetType __ret;
 
-    auto __stream = ::cuda::__call_or(::cuda::get_stream, ::cuda::stream_ref{cudaStreamPerThread}, __policy);
+    auto __stream = ::cuda::__call_or(::cuda::get_stream, ::cuda::stream_ref{cudaStream_t{}}, __policy);
 
     // Determine temporary device storage requirements
     void* __temp_storage = nullptr;
--- a/libcudacxx/include/cuda/std/__pstl/cuda/transform_reduce.h
+++ b/libcudacxx/include/cuda/std/__pstl/cuda/transform_reduce.h
@@ -92,7 +92,7 @@ struct __pstl_dispatch<__pstl_algorithm::__transform_reduce, __execution_backend
       __init);
 
     // Allocate memory for result
-    auto __stream = ::cuda::__call_or(::cuda::get_stream, ::cuda::stream_ref{cudaStreamPerThread}, __policy);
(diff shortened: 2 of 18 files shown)


libcudacxx/include/cuda/std/__pstl/cuda/copy_if.h
```diff
@@ -71,7 +71,7 @@ struct __pstl_dispatch<__pstl_algorithm::__copy_if, __execution_backend::__cuda>
     using _OffsetType = iter_difference_t<_InputIterator>;
     _OffsetType __ret;
 
-    auto __stream = ::cuda::__call_or(::cuda::get_stream, ::cuda::stream_ref{cudaStreamPerThread}, __policy);
+    auto __stream = ::cuda::__call_or(::cuda::get_stream, ::cuda::stream_ref{cudaStream_t{}}, __policy);
```
miscco, line 74: @bernhardmgruber using an open coded stream seems to be very error prone. I believe we should put that into a separate function in `get_stream.h` so that all algorithms in CCCL use the same default stream when possible?

bernhardmgruber: I agree! But I am certain we have different defaults in different places. CUB's public API has a lot of `cudaSteam_t stream = nullptr` parameters. `git grep` shows me 35 instances of `cudaStream_t{}` and 25 `cudaStreamPerThread{}`.

We should definitely use the same default for PSTL, which hasn't shipped yet. So we can at least make a common function for PSTL.

libcudacxx/include/cuda/std/__pstl/cuda/transform_reduce.h
```diff
@@ -92,7 +92,7 @@ struct __pstl_dispatch<__pstl_algorithm::__transform_reduce, __execution_backend
       __init);
 
     // Allocate memory for result
-    auto __stream = ::cuda::__call_or(::cuda::get_stream, ::cuda::stream_ref{cudaStreamPerThread}, __policy);
+    auto __stream = ::cuda::__call_or(::cuda::get_stream, ::cuda::stream_ref{cudaStream_t{}}, __policy);
```
bernhardmgruber, line 95: @pciolkosz which stream is actually better as a default? The per thread or the per stream one?

bernhardmgruber (approved):

libcudacxx/include/cuda/std/__pstl/cuda/copy_if.h
miscco: The problem is that PSTL *must* use the same stream if we want to use the CUB environment APIs

Otherwise us synchronizing after the algorithm wont have any effect

bernhardmgruber: Oh, I see the problem. I think we use the null stream everywhere at the CUB public API, so that's what we should use in the PSTL then as well.

PR #8533, opened 2026-04-20
Remove MSVC warning 4848 because its unused

Jacobfaib:
## Description

MSVC error 4848 is silenced often enough that it warrants being silenced globally. For each such macro, we already ensure that it is only enabled when we are sure the compiler supports it properly.

## Checklist

### DIFF
--- a/cudax/include/cuda/experimental/__execution/prologue.cuh
+++ b/cudax/include/cuda/experimental/__execution/prologue.cuh
@@ -23,7 +23,6 @@
 _CCCL_DIAG_PUSH
 _CCCL_DIAG_SUPPRESS_GCC("-Wsubobject-linkage")
 _CCCL_DIAG_SUPPRESS_CLANG("-Wunused-value")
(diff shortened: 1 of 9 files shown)


miscco (changes requested):
We should drop that warning. 

The `[[no_unique_address]]` emulation is disabled for MSVC because it is completely broken in conjunction with NVCC

Jacobfaib:
> We should drop that warning.

Drop as in silence it? Or drop as in ignore it from command-line? Note that this warning is not just about `no_unique_address`, it also applies to other C++ attributes like `[[likely]]`, ``[[nodiscard("reason")]]]`` or `[[assume]]`.

miscco:
We can drop the warning because we should never use the attribute on windows

Jacobfaib:
But we use the others (like `[[assume]]`) and MSVC would warn about them without the silence no? (Though I don't think MSVC supports `[[assume]]` yet)

Jacobfaib:
Removed the silence

miscco (approved):

PR #8538, opened 2026-04-20
Implement the new tuning API for `detail::batched_topk::dispatch_batched_topk`

### LINKED ISSUE #8479: Implement the new tuning API for `detail::batched_topk::dispatch_batched_topk`

bernhardmgruber:
- [x] No SASS changes for `cub.bench.segmented_topk.keys.base` for SM120

Fixes: #8479

### DIFF
--- a/cub/benchmarks/bench/segmented_topk/keys.cu
+++ b/cub/benchmarks/bench/segmented_topk/keys.cu
@@ -109,7 +109,7 @@ void fixed_seg_size_topk_keys(
     segment_sizes,
     k,
     select_directions,
-    num_segments_uniform_t{num_segments},
+    num_segments_uniform_t{static_cast<::cuda::std::int64_t>(num_segments)},
     total_num_items,
     nullptr
 #if !TUNE_BASE
@@ -133,7 +133,7 @@ void fixed_seg_size_topk_keys(
       segment_sizes,
       k,
       select_directions,
-      num_segments_uniform_t{num_segments},
+      num_segments_uniform_t{static_cast<::cuda::std::int64_t>(num_segments)},
       total_num_items,
       launch.get_stream()
 #if !TUNE_BASE
--- a/cub/benchmarks/bench/segmented_topk/keys.cu
+++ b/cub/benchmarks/bench/segmented_topk/keys.cu
@@ -12,33 +12,36 @@
 // %RANGE% TUNE_THREADS_PER_BLOCK tpb 128:1024:32
 // %RANGE% TUNE_BLOCK_LOAD_ALGORITHM ld 0:2:1
 
-#if TUNE_BLOCK_LOAD_ALGORITHM == 0
-#  define TUNE_LOAD_ALGORITHM cub::BLOCK_LOAD_DIRECT
-#elif TUNE_BLOCK_LOAD_ALGORITHM == 1
-#  define TUNE_LOAD_ALGORITHM cub::BLOCK_LOAD_WARP_TRANSPOSE
-#elif TUNE_BLOCK_LOAD_ALGORITHM == 2
-#  define TUNE_LOAD_ALGORITHM cub::BLOCK_LOAD_VECTORIZE
-#endif
-
 #if !TUNE_BASE
-template <class KeyInT, class OffsetT>
-struct policy_hub_t
+struct tuned_policy_selector
 {
-  struct policy_t : cub::ChainedPolicy<300, policy_t, policy_t>
+  [[nodiscard]] _CCCL_API constexpr auto operator()(::cuda::arch_id /*arch*/) const
+    -> cub::detail::batched_topk::batched_topk_policy
   {
-    static constexpr int items_per_thread = TUNE_ITEMS_PER_THREAD;
-
-    static constexpr int bits_per_pass = cub::detail::topk::calc_bits_per_pass<KeyInT>();
-
-    using topk_policy_t =
-      cub::detail::topk::AgentTopKPolicy<TUNE_THREADS_PER_BLOCK,
-                                         items_per_thread,
-                                         bits_per_pass,
-                                         TUNE_LOAD_ALGORITHM,
-                                         cub::BLOCK_SCAN_WARP_SCANS>;
-  };
-
-  using MaxPolicy = policy_t;
+    // Single-entry policy chain driven by the tuning knobs.
+    constexpr auto store_alg = cub::BLOCK_STORE_WARP_TRANSPOSE;
+#  if TUNE_BLOCK_LOAD_ALGORITHM == 0
+    constexpr auto load_alg = cub::BLOCK_LOAD_DIRECT;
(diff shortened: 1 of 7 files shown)


cub/benchmarks/bench/segmented_topk/keys.cu
```diff
@@ -12,33 +12,36 @@
...
-
-    static constexpr int bits_per_pass = cub::detail::topk::calc_bits_per_pass<KeyInT>();
-
-    using topk_policy_t =
-      cub::detail::topk::AgentTopKPolicy<TUNE_THREADS_PER_BLOCK,
-                                         items_per_thread,
-                                         bits_per_pass,
-                                         TUNE_LOAD_ALGORITHM,
-                                         cub::BLOCK_SCAN_WARP_SCANS>;
-  };
```
bernhardmgruber, line 39: Remark: I think this was just wrong before?

cub/cub/device/dispatch/tuning/tuning_batched_topk.cuh
```diff
@@ -16,35 +16,108 @@
...
 
-#include <cuda/std/tuple>
+#include <cuda/__device/arch_id.h>
+#include <cuda/std/__host_stdlib/ostream>
+#include <cuda/std/array>
 
 CUB_NAMESPACE_BEGIN
 namespace detail::batched_topk
 {
+struct agent_batched_topk_policy
```
bernhardmgruber: TODO: rename

cub/cub/device/dispatch/kernels/kernel_batched_topk.cuh
```diff
@@ -18,140 +18,82 @@
...
-
-  // Select result:
-  // If 'fits' is true, we stop here.
-  // If 'fits' is false, we take the result from 'next_step'.
-  using policy_t = ::cuda::std::conditional_t<fits, current_policy_t, typename next_step::policy_t>;
-
-  // Whether there's a valid policy that we can instantiate the agent with such that the agent's shared memory doesn't
-  // exceed the static shared memory limimt
-  static constexpr bool has_valid_policy = fits ? true : next_step::has_valid_policy;
-};
```
bernhardmgruber, line 83: Remark: This code was entirely unused.

cub/cub/device/dispatch/kernels/kernel_batched_topk.cuh
```diff
@@ -18,140 +18,82 @@
...
+          return active_policy.worker_per_segment_policies[Index];
+        }
+      };
+      using candidate_agent_t  = agent_batched_topk_worker_per_segment<policy_getter_17, AgentParamsT...>;
+      constexpr bool covers    = tile_size >= max_segment_size;
+      constexpr bool fits_smem = sizeof(typename candidate_agent_t::TempStorage) <= max_smem_per_block;
+      constexpr int next       = find_index<Index + 1>();
+      if constexpr (covers && fits_smem)
+      {
+        return next >= 0 ? next : Index;
```
bernhardmgruber, line 65: Important remark: I reviewed the initial implementation claude generated and noticed it did not include the SMEM check (`fits_smem`), so I instructed it to add it. I only later noticed that the entire code around `find_first_smem_fitting_policy_impl` was actually unused. So the logic implemented here differs from the original implementation, since it finds the smallest covering policy that ALSO fits into SMEM. I think this is an improvement, but I am fine removing this extra again.

cub/cub/device/dispatch/dispatch_batched_topk.cuh
```diff
@@ -145,193 +145,82 @@ template <typename KeyInputItItT,
...
+  ValueOutputItItT d_value_segments_out_it,
+  SegmentSizeParameterT segment_sizes,
+  KParameterT k,
+  SelectDirectionParameterT select_directions,
+  NumSegmentsParameterT num_segments,
+  [[maybe_unused]] TotalNumItemsGuaranteeT total_num_items_guarantee,
+  cudaStream_t stream                             = nullptr,
+  [[maybe_unused]] PolicySelector policy_selector = {})
 {
-  using offset_t = ::cuda::std::int64_t;
```
bernhardmgruber, line 154: Remark: this was also unused

cub/benchmarks/bench/segmented_topk/keys.cu
elstehle: Yeah, I think so. I think before we tune, we want to make sure to make `bits_per_pass` a tuning parameter, but that's future work.

cub/cub/device/dispatch/tuning/tuning_batched_topk.cuh
```diff
@@ -13,38 +13,111 @@
...
+        << ", .load_algorithm = " << p.load_algorithm << ", .store_algorithm = " << p.store_algorithm << " }";
+  }
+#endif // !_CCCL_COMPILER(NVRTC)
+};
+
+struct batched_topk_policy
+{
+  // The list of per-segment agent policies is ordered by decreasing tile size. At compile time, the smallest policy
+  // whose tile size still covers the upper bound of the segment size is selected.
+  ::cuda::std::array<worker_policy, 6> worker_per_segment_policies;
```
elstehle, line 58: question: How would we envision this to work with varying number of `worker_per_segment_policies`. E.g., if `sm_120` had 6, but `sm_80` had only 4 `worker_per_segment_policies`?

cub/cub/device/dispatch/kernels/kernel_batched_topk.cuh
elstehle: Thanks for pointing this out. In my mind, I assumed that as tile sizes are decreasing, I assumed strictly non-increasing shared memory requirements. That will probably hold for some time to come, but I think the extra check won't hurt.

cub/cub/device/dispatch/kernels/kernel_batched_topk.cuh
elstehle: Yeah, the motivation for `find_first_smem_fitting_policy` is that it will be needed once we support larger-than-what-fits-into-one-worker-per-segment segments, as we want to figure out how large of a segment size we can support "in-place" by a single worker. During the PR review meeting where we reviewed the PR that made `find_smallest_covering_policy` superseed `find_first_smem_fitting_policy`, we agreed to keep this part around until that's the case. But I'm also ok to copy it back in, once needed.

elstehle (approved):
Thanks a lot for taking this on! Changes look good to me.

cub/cub/device/dispatch/tuning/tuning_batched_topk.cuh
bernhardmgruber: We can't in the current design. We would need a container of dynamic size that works in `constexpr` land. If you can give me some maximum, we could use `cuda::std::inplace_vector`. Otherwise we are going to have a hard time :S

PR #8542, opened 2026-04-21
[libcu++] Add total_global_memory device attribute

pciolkosz:
Working on https://github.com/NVIDIA/cccl/pull/8541 I encountered a need for this simple wrapper over `cuDeviceTotalMem`

### DIFF
--- a/libcudacxx/include/cuda/__device/attributes.h
+++ b/libcudacxx/include/cuda/__device/attributes.h
@@ -759,6 +759,18 @@ using host_memory_pools_supported_t = __dev_attr<::cudaDevAttrHostMemoryPoolsSup
 static constexpr host_memory_pools_supported_t host_memory_pools_supported{};
 #  endif // ^^^ _CCCL_CTK_AT_LEAST(13, 0) ^^^
 
+// Total global memory available on the device in bytes
+struct total_global_memory_t
+{
+  using type = ::cuda::std::size_t;
+
+  [[nodiscard]] _CCCL_HOST_API type operator()(device_ref __dev) const
+  {
+    return ::cuda::__driver::__deviceTotalMem(__dev.get());
+  }
+};
+static constexpr total_global_memory_t total_global_memory{};
+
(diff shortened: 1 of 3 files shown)


libcudacxx/include/cuda/__device/device_ref.h
```diff
@@ -142,6 +142,13 @@ class device_ref
                                                                       // <cuda/__device/physical_device.h> to avoid
                                                                       // circular dependency
 
+  //! @brief Retrieve the total global memory of this device in bytes.
+  //!
+  //! @return The total global memory of this device in bytes.
+  [[nodiscard]] _CCCL_HOST_API ::cuda::std::size_t total_global_memory() const; // implemented in
+                                                                                // <cuda/__device/physical_device.h>
+                                                                                // to avoid circular dependency
```
davebayer, line 150: Q: Shouldn't this be just an attribute in `cuda::device_attributes` namespace?

davebayer (commented):
Just one question

libcudacxx/include/cuda/__device/device_ref.h
pciolkosz: It could be, I am not sure if for things that have no cuda runtime attributes we should add them as attributes or separate APIs. I would be fine both ways, not sure what pros and cons are

davebayer: We already do that for `compute_capability`. I don't think this case is any different. Honestly, I'm not even sure why this API isn't just another attribute. Maybe because the attributes return `int`, but `size_t` is needed? Anyway, I really think this should go to attributes.

I'm not even sure whether `device_ref::name()` was the right choice, maybe it should have gone to the attributes, too

pciolkosz: I switched to an attribute, I think it makes sense

miscco (approved):

davebayer (approved):

PR #8543, opened 2026-04-21
[docs] Hide cub::DeviceReduce::__arg_min_env and fix ArgMin @overload

gonidelis:
internal helper `__arg_min_env` was by mistake exposed to the docs.

fixes `ArgMin` docs along the way

### DIFF
--- a/cub/cub/device/device_reduce.cuh
+++ b/cub/cub/device/device_reduce.cuh
@@ -1087,9 +1087,10 @@ public:
   //! @rst
   //! .. versionadded:: 2.2.0
   //!    First appears in CUDA Toolkit 12.3.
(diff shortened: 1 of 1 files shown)


bernhardmgruber:
Thx for the fix! I think @pauleonix was faster in #8536. But you also moved an ` @endrst` in this PR. Can you please rebase on `main` so we can integrate that fix as well? Thx!

bernhardmgruber (approved):

bernhardmgruber:
Turns out `git` took care of that ;) No action needed then!

gonidelis:
> Turns out git took care of that ;) No action needed then!

how?
image

PR #8549, opened 2026-04-21
[Tile] Disable `__builtin_expect` until its supported in tile

miscco:
We utilize `__builtin_expect` in our assert implementation. As this is purely an optimization hint for the compiler disable it until it is supported

### DIFF
--- a/libcudacxx/include/cuda/std/__cccl/builtin.h
+++ b/libcudacxx/include/cuda/std/__cccl/builtin.h
@@ -117,6 +117,11 @@
 #  define _CCCL_BUILTIN_EXPECT(_EXPR, _VAL) (_EXPR)
 #endif // ^^^ no __builtin_expect ^^^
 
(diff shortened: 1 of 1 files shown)


bernhardmgruber (approved):

libcudacxx/include/cuda/std/__cccl/builtin.h
```diff
@@ -117,6 +117,11 @@
 #  define _CCCL_BUILTIN_EXPECT(_EXPR, _VAL) (_EXPR)
 #endif // ^^^ no __builtin_expect ^^^
 
+#if _CCCL_TILE_COMPILATION() // TODO(miscco):  __builtin_expect is unsupported in tile mode
```
fbusato: same here, would be great to track the compiler feature with an nvbug

fbusato (approved):

libcudacxx/include/cuda/std/__cccl/builtin.h
miscco: Filed https://nvbugspro.nvidia.com/bug/6100927

PR #8550, opened 2026-04-21
[Tile] Disable `__builtin_assume` until it is supported in tile

miscco:
It is not supported in tile mode. Given that it is a pure compiler optimization hint just disable it for now

### DIFF
--- a/libcudacxx/include/cuda/std/__cccl/attributes.h
+++ b/libcudacxx/include/cuda/std/__cccl/attributes.h
@@ -84,6 +84,11 @@
 #  define _CCCL_ASSUME(...) _CCCL_BUILTIN_ASSUME(__VA_ARGS__)
 #endif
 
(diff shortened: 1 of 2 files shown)


bernhardmgruber (approved):

libcudacxx/include/cuda/std/__cccl/attributes.h
```diff
@@ -84,6 +84,11 @@
 #  define _CCCL_ASSUME(...) _CCCL_BUILTIN_ASSUME(__VA_ARGS__)
 #endif
 
+#if _CCCL_TILE_COMPILATION() // __builtin_assume is not supported in tile mode
```
fbusato: is there any nvbug associated with it?

fbusato (approved):

libcudacxx/include/cuda/std/__cccl/attributes.h
miscco: Filed https://nvbugspro.nvidia.com/bug/6100910

PR #8564, opened 2026-04-21
Add a freestanding mode to allow building without any host standard library support

miscco:
We want to generalize our use of the host libraries and explore building without them

### DIFF
--- a/cub/cub/iterator/cache_modified_input_iterator.cuh
+++ b/cub/cub/iterator/cache_modified_input_iterator.cuh
@@ -104,13 +104,13 @@ public:
   /// The type of a reference to an element the iterator can point to
   using reference = ValueType;
 
-#if _CCCL_COMPILER(NVRTC)
+#if _CCCL_FREESTANDING()
   using iterator_category = ::cuda::std::random_access_iterator_tag;
-#else // ^^^ _CCCL_COMPILER(NVRTC) ^^^ // vvv !_CCCL_COMPILER(NVRTC) vvv
+#else // ^^^ _CCCL_FREESTANDING() ^^^ // vvv _CCCL_HOSTED() vvv
   using iterator_category =
     THRUST_NS_QUALIFIER::detail::iterator_facade_category_t<THRUST_NS_QUALIFIER::device_system_tag,
                                                             THRUST_NS_QUALIFIER::random_access_traversal_tag>;
-#endif // _CCCL_COMPILER(NVRTC)
+#endif // _CCCL_HOSTED()
 
 public:
   /// Wrapped native pointer
@@ -207,12 +207,12 @@ public:
   }
 
   /// ostream operator
-#if !_CCCL_COMPILER(NVRTC)
+#if _CCCL_HOSTED()
   friend ::std::ostream& operator<<(::std::ostream& os, const self_type& /*itr*/)
   {
     return os;
   }
-#endif // !_CCCL_COMPILER(NVRTC)
+#endif // _CCCL_HOSTED()
 };
 
 namespace detail
--- a/libcudacxx/include/cuda/std/__cccl/host_std_lib.h
+++ b/libcudacxx/include/cuda/std/__cccl/host_std_lib.h
@@ -34,28 +34,6 @@
 #  include <ciso646>
 #endif // ^^^ __has_include(<ciso646>) ^^^
 
-#define _CCCL_HOST_STD_LIB_MAKE_VERSION(_MAJOR, _MINOR) ((_MAJOR) * 100 + (_MINOR))
-#define _CCCL_HOST_STD_LIB(...)                         _CCCL_VERSION_COMPARE(_CCCL_HOST_STD_LIB_, _CCCL_HOST_STD_LIB_##__VA_ARGS__)
-
-#if defined(_MSVC_STL_VERSION)
-#  undef _CCCL_HOST_STD_LIB_STL
-#  define _CCCL_HOST_STD_LIB_STL() (_MSVC_STL_VERSION, 0)
-#elif defined(__GLIBCXX__)
-#  undef _CCCL_HOST_STD_LIB_LIBSTDCXX
-#  define _CCCL_HOST_STD_LIB_LIBSTDCXX() (_GLIBCXX_RELEASE, 0)
-#elif defined(_LIBCPP_VERSION)
-#  undef _CCCL_HOST_STD_LIB_LIBCXX
-// since llvm-16, the version scheme has been changed from MMppp to MMmmpp
-#  if _LIBCPP_VERSION / 10000 < 2
-#    define _CCCL_HOST_STD_LIB_LIBCXX() (_LIBCPP_VERSION / 1000, 0)
-#  else
-#    define _CCCL_HOST_STD_LIB_LIBCXX() (_LIBCPP_VERSION / 10000, (_LIBCPP_VERSION / 100) % 100)
-#  endif
-#endif // ^^^ _LIBCPP_VERSION ^^^
-
-#define _CCCL_HAS_HOST_STD_LIB() \
-  (_CCCL_HOST_STD_LIB(LIBSTDCXX) || _CCCL_HOST_STD_LIB(LIBCXX) || _CCCL_HOST_STD_LIB(STL))
-
 // Freestanding environment detection
 // NVRTC is treated as freestanding since it has no access to the host standard library
 #if defined(_CCCL_ENABLE_FREESTANDING) || _CCCL_COMPILER(NVRTC)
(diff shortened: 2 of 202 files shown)


libcudacxx/include/cuda/std/__cccl/host_std_lib.h
```diff
@@ -56,4 +56,15 @@
...
+// Freestanding environment detection
+// NVRTC is treated as freestanding since it has no access to the host standard library
+#if defined(_CCCL_ENABLE_FREESTANDING) || _CCCL_COMPILER(NVRTC)
+#  define _CCCL_FREESTANDING() 1
+#  define _CCCL_NO_TYPEID
+#else
+#  define _CCCL_FREESTANDING() 0
+#endif
+
+#define _CCCL_HOSTED() (!_CCCL_FREESTANDING())
```
bernhardmgruber, line 68: Suggestion: I would strongly prefer to have a single macro, either `_CCCL_FREESTANDING` and `_CCCL_HOSTED`.

bernhardmgruber (commented):
CUB and Thrust changes LGTM

cub/cub/iterator/cache_modified_input_iterator.cuh
```diff
@@ -104,13 +104,13 @@ public:
...
-#if _CCCL_COMPILER(NVRTC)
+#if _CCCL_FREESTANDING()
   using iterator_category = ::cuda::std::random_access_iterator_tag;
-#else // ^^^ _CCCL_COMPILER(NVRTC) ^^^ // vvv !_CCCL_COMPILER(NVRTC) vvv
+#else // ^^^ _CCCL_FREESTANDING() ^^^ // vvv _CCCL_HOSTED() vvv
   using iterator_category =
     THRUST_NS_QUALIFIER::detail::iterator_facade_category_t<THRUST_NS_QUALIFIER::device_system_tag,
                                                             THRUST_NS_QUALIFIER::random_access_traversal_tag>;
-#endif // _CCCL_COMPILER(NVRTC)
+#endif // _CCCL_HOSTED()
```
davebayer, line 113: I think @bernhardmgruber and we should just have one. Since we use `_CCCL_HOSTED()` almost everywhere, I would keep just that one

libcudacxx/include/cuda/std/__utility/move.h
```diff
@@ -39,6 +39,12 @@
 #  define _CCCL_HAS_BUILTIN_STD_MOVE() 1
 #endif // _CCCL_CUDA_COMPILER(NVCC) && _CCCL_DEVICE_COMPILATION()
 
+// We cannot use host features if the user requires `_CCCL_ENABLE_FREESTANDING`
+#ifdef _CCCL_ENABLE_FREESTANDING
```
davebayer, line 43:
```suggestion
#if !_CCCL_HOSTED()
```

libcudacxx/include/cuda/std/__cccl/host_std_lib.h
```diff
@@ -56,4 +56,15 @@
 #define _CCCL_HAS_HOST_STD_LIB() \
   (_CCCL_HOST_STD_LIB(LIBSTDCXX) || _CCCL_HOST_STD_LIB(LIBCXX) || _CCCL_HOST_STD_LIB(STL))
```
davebayer, line 57: Shouldn't we disable `_CCCL_HAS_HOST_STD_LIB()` when compiling in freestanding mode, too?

cub/cub/iterator/cache_modified_input_iterator.cuh
miscco: There are currently still places where it is breaking

libcudacxx/include/cuda/std/__utility/move.h
miscco: This was intentional, because `NVRTC` defines a compiler builtin for that, so we only disable when the user explicitly uses `_CCCL_ENABLE_FREESTANDING`

libcudacxx/include/cuda/std/__cccl/host_std_lib.h
miscco: I believe it should be disabled by none of the macros being picked up because we do not include any of the files that would give us the host STL detection

libcudacxx/include/cuda/std/__cccl/host_std_lib.h
miscco: That was my perspective too, but it seems that `hosted` is a term of art in the standard, so I believe its actually not too bad to have them

libcudacxx/include/cuda/std/__utility/move.h
davebayer: Oh yeah, so maybe we should add a comment :)

libcudacxx/include/cuda/std/__cccl/host_std_lib.h
bernhardmgruber: I know. So can we skip `CCCL_FREESTANDING()` then?

libcudacxx/include/cuda/std/__utility/move.h
davebayer: Maybe `_CCCL_HOSTED() && !_CCCL_COMPILER(NVRTC)` would be a better solution?

miscco: I believe we should keep it as is

cub/cub/iterator/cache_modified_input_iterator.cuh
miscco: Oh sorry, I misread you are talking about `_CCCL_FREESTANDING()` and `_CCCL_HOSTED()` not the code in that snipped?

libcudacxx/include/cuda/std/__cccl/host_std_lib.h
miscco: Honestly I believe that `!_CCCL_SOMETHING` is actually a really easy to make mistake and I would say that having two macros does not hurt

fbusato: I'm also in favor of `_CCCL_HOSTED`. Otherwise, we have the same readability problem of !NVRTC

libcudacxx/include/cuda/std/__cccl/host_std_lib.h
```diff
@@ -56,4 +56,15 @@
...
 
+// Freestanding environment detection
+// NVRTC is treated as freestanding since it has no access to the host standard library
+#if defined(_CCCL_ENABLE_FREESTANDING) || _CCCL_COMPILER(NVRTC)
+#  define _CCCL_FREESTANDING() 1
```
fbusato, line 40: there are a log of changes, would also make sense to split the PR between _CCCL_FREESTANDING and _CCCL_HOST

libcudacxx/include/cuda/std/__cccl/host_std_lib.h
davebayer: I've already solved one nvbug, because of a similar reason. I though some libstc++ macro will only be defined when compiling against libstdc++ - that was a wrong assumption, because the project defined some of those macros themselves :D

So, I'd rather disable it manually, so it's not our problem users doing dangerous things

libcudacxx/include/cuda/std/__utility/forward.h
```diff
@@ -38,15 +38,21 @@
 #  define _CCCL_HAS_BUILTIN_STD_FORWARD() 1
 #endif // _CCCL_CUDA_COMPILER(NVCC) && _CCCL_DEVICE_COMPILATION()
 
+// We cannot use host features if the user requires `_CCCL_ENABLE_FREESTANDING`
+#ifdef _CCCL_ENABLE_FREESTANDING
+#  undef _CCCL_HAS_BUILTIN_STD_FORWARD
+#  define _CCCL_HAS_BUILTIN_STD_FORWARD() 0
+#endif // _CCCL_ENABLE_FREESTANDING
```
davebayer: I'm sorry, but to it is not obvious to me that this means `_CCCL_FREESTANDING() && !_CCCL_COMPILER(NVRTC)`. I'd like this to be changed

libcudacxx/include/cuda/std/__cccl/host_std_lib.h
davebayer: Can we go with only `_CCCL_HOSTED`, then?

miscco: Decided to keep both

libcudacxx/include/cuda/std/__cccl/host_std_lib.h
miscco: Updated it

davebayer (approved):

PR #8565, opened 2026-04-21
Replace `detail::for_each::dispatch` by CUB's public API

### LINKED ISSUE #7953: Use the new tuning API internally for `detail::for_each::dispatch_t`

bernhardmgruber:
- [x] No SASS changes in `cub.bench.for_each.base.base` for SM120
- [x] No SASS changes in `cub.bench.for_each.copy.base` for SM120
- [x] No SASS changes in `cub.bench.for_each.extents.base` for SM120

Fixes: #7953

### DIFF
--- a/cub/cub/device/device_for.cuh
+++ b/cub/cub/device/device_for.cuh
@@ -44,6 +44,8 @@ struct DeviceFor
   template <class OffsetT, class OpT, class RandomAccessIteratorT>
   struct __op_wrapper_t
   {
+    static_assert(::cuda::std::is_integral_v<OffsetT>);
+
     RandomAccessIteratorT input;
     OpT op;
 
@@ -62,6 +64,8 @@ struct DeviceFor
   template <class OffsetT, class OpT, class T>
   struct __op_wrapper_vectorized_t
   {
+    static_assert(::cuda::std::is_integral_v<OffsetT>);
+
     const T* input; // Raw pointer to the input data
     OpT op; // User-provided operator
     OffsetT partially_filled_vector_id; // Index of the vector that doesn't have all elements
--- a/cub/cub/device/device_for.cuh
+++ b/cub/cub/device/device_for.cuh
@@ -21,6 +21,7 @@
 #include <thrust/type_traits/unwrap_contiguous_iterator.h>
 
 #include <cuda/__cmath/ceil_div.h>
+#include <cuda/__execution/tune.h>
 #include <cuda/__functional/call_or.h>
 #include <cuda/__stream/get_stream.h>
 #include <cuda/std/__concepts/concept_macros.h>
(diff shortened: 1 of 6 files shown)


cub/cub/device/device_for.cuh
```diff
@@ -37,102 +38,120 @@
...
-template <class OffsetT, class OpT, class RandomAccessIteratorT>
-struct op_wrapper_t
+struct DeviceFor
 {
-  RandomAccessIteratorT input;
-  OpT op;
-
-  _CCCL_DEVICE _CCCL_FORCEINLINE void operator()(OffsetT i)
+  //! `__op_wrapper_t` turns bulk into a for-each operation by wrapping the user-provided unary operator.
+  template <class OffsetT, class OpT, class RandomAccessIteratorT>
```
oleksandr-pavlyk, line 44: Is there a need to constrain the type with some concept check (if concepts are available)?

cub/cub/device/device_for.cuh
```diff
@@ -37,102 +38,120 @@
...
       if (::cuda::std::is_sufficiently_aligned<alignof(typename wrapped_op_t::vector_t)>(unwrapped_first))
       { // Vectorize loads
-        const OffsetT num_vec_items = ::cuda::ceil_div(num_items, wrapped_op_t::vec_size);
-
-        return detail::for_each::dispatch<OffsetT, wrapped_op_t>(
+        const NumItemsT num_vec_items = ::cuda::ceil_div(num_items, wrapped_op_t::vec_size);
+        return __bulk(
           num_vec_items,
           wrapped_op_t{
             unwrapped_first, op, num_items % wrapped_op_t::vec_size ? num_vec_items - 1 : num_vec_items, num_items},
```
oleksandr-pavlyk: Perhaps just use ordinary integer division instead?

```suggestion
            unwrapped_first, op, (num_items / wrapped_op_t::vec_size), num_items},
```

Hope compiler can optimize repeated divisions away.

fbusato: probably yes, the original code looks wrong

cub/cub/device/device_for.cuh
fbusato: do you mean

- `OffsetT` integral type?
- `OpT` get a single arg
- `RandomAccessIteratorT` is a random access iter?

yes, it is a good practice

oleksandr-pavlyk (approved):

cub/cub/device/device_for.cuh
bernhardmgruber: I added a static assert that the offset type is integral. I think we may not add checks for `OpT` and the iterators, since CCCL.C may type erase them to `indirect_iterator_t` and `indirect_value_t`, which are not iterators.

cub/cub/device/device_for.cuh
bernhardmgruber: I think the original code is fine. It computes the index of the for iteration where we find a partial vector. If `num_items` is divisible without remainder, there is no partial vector so we use `num_vec_items` as index, which is never reached by DeviceFor. Otherwise, it's the last index.

I think @oleksandr-pavlyk correctly spotted that this is the same as integer dividing `num_items / vec_size`.

cub/cub/device/device_for.cuh
```diff
@@ -37,102 +38,121 @@
...
-} // namespace detail::for_each
 
-struct DeviceFor
-{
-private:
-  template <bool UseVectorization, class RandomAccessOrContiguousIteratorT, class OffsetT, class OpT>
+  template <bool AllowCopy = false, class RandomAccessIteratorT, class NumItemsT, class OpT, class EnvT>
   CUB_RUNTIME_FUNCTION static cudaError_t
-  for_each_n(RandomAccessOrContiguousIteratorT first, OffsetT num_items, OpT op, cudaStream_t stream)
+  __for_each_n(RandomAccessIteratorT first, NumItemsT num_items, OpT op, EnvT env)
```
Jacobfaib, line 118: Does this (and the other changes) need to be `__ugly`? Can you not make it `private` instead? I thought `__ugly` syntax was for libcudacxx only.

cub/cub/device/device_for.cuh
```diff
@@ -37,102 +38,121 @@
...
   {
-    if constexpr (UseVectorization)
+    // We tried to detect if we can still use vectorization from the non-Copy CUB APIs, but it's disabled for now:
```
Jacobfaib, line 120: Can you elaborate in the comment why it's disabled?

cub/cub/device/device_for.cuh
```diff
@@ -37,102 +38,121 @@
...
-      // Fallback to non-vectorized version
-      return for_each_n<false>(first, num_items, op, stream);
     }
-    else
+
+    return __bulk(num_items, __op_wrapper_t<NumItemsT, OpT, RandomAccessIteratorT>{first, op}, env);
+  }
+
+  template <class RandomAccessIteratorT, class NumItemsT, class OpT>
+  CUB_RUNTIME_FUNCTION static cudaError_t __for_each_n(
```
Jacobfaib: These internal functions should at least be `[[nodiscard]]`. (`__bulk()` and the others as well while we're here).

cub/cub/device/device_for.cuh
```diff
@@ -37,102 +38,121 @@
...
- */
-template <class OffsetT, class OpT, class RandomAccessIteratorT>
-struct op_wrapper_t
+struct DeviceFor
 {
-  RandomAccessIteratorT input;
-  OpT op;
+  //! `__op_wrapper_t` turns bulk into a for-each operation by wrapping the user-provided unary operator.
+  template <class OffsetT, class OpT, class RandomAccessIteratorT>
+  struct __op_wrapper_t
```
Jacobfaib, line 45: Can you make this class (and the other one) `private:`?

cub/cub/device/device_for.cuh
```diff
@@ -37,102 +38,121 @@
...
+        const vector_t vec = *reinterpret_cast<const vector_t*>(input + vec_size * i);
 
-      _CCCL_PRAGMA_UNROLL_FULL()
-      for (int j = 0; j < vec_size; j++)
-      {
-        (void) op(*(reinterpret_cast<const T*>(&vec) + j));
+        _CCCL_PRAGMA_UNROLL_FULL()
+        for (int j = 0; j < vec_size; j++)
+        {
+          (void) op(*(reinterpret_cast<const T*>(&vec) + j));
```
Jacobfaib, line 98: Nit: consider moving this cast outside of the loop

cub/cub/device/device_for.cuh
```diff
@@ -37,102 +38,121 @@
...
-    // Surrounding `Bulk` call doesn't invoke this operator on invalid indices, so we don't need to
-    // check for out-of-bounds access here.
-    if (i != partially_filled_vector_id)
-    { // Case of fully filled vector
-      const vector_t vec = *reinterpret_cast<const vector_t*>(input + vec_size * i);
+    _CCCL_DEVICE _CCCL_FORCEINLINE void operator()(OffsetT i)
+    {
+      // Surrounding `Bulk` call doesn't invoke this operator on invalid indices, so we don't need to
+      // check for out-of-bounds access here.
+      if (i != partially_filled_vector_id)
```
Jacobfaib: Nit: consider reversing polarity of this `if`. In general, positive assertions tend to be easier to understand than contrapositives, i.e. `if (foo && bar)` is easier than `if (!foo || !bar)`.

cub/cub/device/device_for.cuh
bernhardmgruber: Unfortunately not, because when they can no longer appear in the signature of a kernel (nvcc will complain)

cub/cub/device/device_for.cuh
bernhardmgruber: Fine. This will add unrelated refactorings to this PR however.

cub/cub/device/device_for.cuh
bernhardmgruber: I prefer the code like it is, since it uses the cast locally where it's needed.

cub/cub/device/device_for.cuh
bernhardmgruber: This is still somewhat of an open debate. Historically, we have only uglified libcudacxx. However, all the libraries converge more and more, and I would also like to have more universal rules across CCCL. I think uglification is the way to go to harden against macros etc. and it's also shorter than having a detail namespace. Others think it creates an additional barrier for contributors.

cub/cub/device/device_for.cuh
bernhardmgruber: I retained the comment (and generally all implementation here) from the previous code. I think @gevtushenko ran into troubles detecting whether we can copy the argument to the user-provided op. and disabled vectorization for safety.

cub/cub/device/device_for.cuh
Jacobfaib: Hmmm this is a good point. I guess we probably should address this before I turn on the readability checks for clang-tidy.

cub/cub/device/device_for.cuh
Jacobfaib: @gevtushenko can you confirm this is why? It's not a huge deal but would be nice to have this context in the code for future maintainers to decide if it's still warranted

Jacobfaib (approved):

cub/cub/device/device_for.cuh
bernhardmgruber: The rollout of this check will take a long time, because even a library like CUB has a different naming convention at the public API vs. for everything internal. And that cannot be separated at the file system level (we have detail functions in public headers).

miscco (approved):

PR #8566, opened 2026-04-21
Extend _LIBCUDACXX_HIDDEN_FRIEND_NAMESPACE to also cover MSVC

Jacobfaib:
## Description

Use `LIBCUDACXX_HIDDEN_FRIEND_NAMESPACE` as workaround for broken MSVC parser

## Checklist


miscco (changes requested):
I am not against this per se, but then we need to remove the separate msvc workaround namespace

Jacobfaib:
> I am not against this per se, but then we need to remove the separate msvc workaround namespace

Those are only introduced in the ranges PRs, so depending on which of these merges first I either remove it from here or from there

miscco (approved):

PR #8602, opened 2026-04-21
Fix reduce tunings

srinivasyadav18:
## Description

Current tuning reduce doesn't use `AccumT`, which is needed for `scale_mem_bound`, this PR fixes by adding `AccumT` to `PolicySelector`. 

Additionally it also fixes `TUNE_ITEMS_PER_THREAD ` naming convention by replacing with `TUNE_NOMINAL_4B_ITEMS_PER_THREAD` which shows that ITEMS_PER_THREAD are indeed and pre-scaled before passing onto tuning policy.


cub/benchmarks/bench/reduce/base.cuh
```diff
@@ -11,12 +11,15 @@
 #include <nvbench_helper.cuh>
 
 #if !TUNE_BASE
+template <typename AccumT>
 struct policy_selector
 {
```
oleksandr-pavlyk, line 16: Policy selector must be non-templated. It could have `int accum_size;` member though passed to it on construction. 

This is what segmented scan does.

srinivasyadav18: IIUC, `policy_selector` in public `tuning_*.cuh` should be non-templated. But, in benchmarks/tunings in `cub/bench` can be templated. For e.g. in [scan](https://github.com/NVIDIA/cccl/blob/main/cub/benchmarks/bench/scan/policy_selector.h#L15).

bernhardmgruber: It's fine here. Only the `policy_selector` in the CUB tuning header should have no template parameters, so CCCL.C can use it without types.

bernhardmgruber (approved):
Thx for catching this oversight introduced in #6544!

oleksandr-pavlyk (approved):

miscco (approved):

srinivasyadav18:
Closing, as its fixed in https://github.com/NVIDIA/cccl/pull/8759/

PR #8603, opened 2026-04-21
Implement std::common_iterator

Jacobfaib:
## Description

See #93 

## Checklist

### DIFF
--- a/libcudacxx/include/cuda/std/__iterator/common_iterator.h
+++ b/libcudacxx/include/cuda/std/__iterator/common_iterator.h
@@ -4,7 +4,7 @@
 // Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
 // See https://llvm.org/LICENSE.txt for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
-// SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES
 //
 //===----------------------------------------------------------------------===//
 
@@ -91,7 +91,7 @@ class common_iterator
   {
     iter_value_t<_Iter> __value_;
 
-    _CCCL_API constexpr const iter_value_t<_Iter>* operator->() const noexcept
+    [[nodiscard]] _CCCL_API constexpr const iter_value_t<_Iter>* operator->() const noexcept
     {
       return ::cuda::std::addressof(__value_);
     }
@@ -152,7 +152,7 @@ public:
   }
 
   _CCCL_EXEC_CHECK_DISABLE
-  _CCCL_API constexpr decltype(auto) operator*()
+  [[nodiscard]] _CCCL_API constexpr decltype(auto) operator*()
   {
     _CCCL_ASSERT(__hold_.__holds_first(), "Attempted to dereference a non-dereferenceable common_iterator");
     return *__hold_.__first_;
--- a/libcudacxx/include/cuda/std/__iterator/common_iterator.h
+++ b/libcudacxx/include/cuda/std/__iterator/common_iterator.h
@@ -101,7 +101,7 @@ class common_iterator
   {
     iter_value_t<_Iter> __value_;
 
-    _CCCL_API constexpr const iter_value_t<_Iter>& operator*() const noexcept
+    [[nodiscard]] _CCCL_API constexpr const iter_value_t<_Iter>& operator*() const noexcept
     {
       return __value_;
     }
@@ -136,7 +136,7 @@ public:
   {}
 
   _CCCL_TEMPLATE(class _I2, class _S2)
-  _CCCL_REQUIRES(convertible_to<const _I2&, _Iter>&& convertible_to<const _S2&, _Sent>)
+  _CCCL_REQUIRES(convertible_to<const _I2&, _Iter> _CCCL_AND convertible_to<const _S2&, _Sent>)
   _CCCL_API constexpr common_iterator(const common_iterator<_I2, _S2>& __other) noexcept(
     is_nothrow_constructible_v<_Iter, const _I2&> && is_nothrow_constructible_v<_Sent, const _S2&>)
       : __hold_{__other.__get_hold()}
@@ -161,7 +161,7 @@ public:
   _CCCL_EXEC_CHECK_DISABLE
   _CCCL_TEMPLATE(class _I2 = _Iter)
   _CCCL_REQUIRES(__dereferenceable<const _I2>)
-  _CCCL_API constexpr decltype(auto) operator*() const
+  [[nodiscard]] _CCCL_API constexpr decltype(auto) operator*() const
   {
     _CCCL_ASSERT(__hold_.__holds_first(), "Attempted to dereference a non-dereferenceable common_iterator");
     return *__hold_.__first_;
@@ -172,7 +172,7 @@ public:
   _CCCL_REQUIRES(indirectly_readable<const _I2> _CCCL_AND(
     __has_const_arrow<_I2> || is_reference_v<iter_reference_t<_I2>>
     || constructible_from<iter_value_t<_I2>, iter_reference_t<_I2>>))
-  _CCCL_API constexpr decltype(auto) operator->() const
+  [[nodiscard]] _CCCL_API constexpr decltype(auto) operator->() const
   {
     _CCCL_ASSERT(__hold_.__holds_first(), "Attempted to dereference a non-dereferenceable common_iterator");
     if constexpr (__has_const_arrow<_Iter>)
@@ -225,10 +225,11 @@ public:
   }
 
   _CCCL_EXEC_CHECK_DISABLE
-  template <class _I2, class _S2>
-  _CCCL_API friend constexpr auto operator==(const common_iterator& __x, const common_iterator<_I2, _S2>& __y)
-    _CCCL_TRAILING_REQUIRES(bool)(sentinel_for<_S2, _Iter>&& sentinel_for<_Sent, _I2>
-                                  && (!equality_comparable_with<_Iter, _I2>) )
+  _CCCL_TEMPLATE(class _I2, class _S2)
+  _CCCL_REQUIRES(
+    sentinel_for<_S2, _Iter> _CCCL_AND sentinel_for<_Sent, _I2> _CCCL_AND(!equality_comparable_with<_Iter, _I2>))
+  [[nodiscard]] _CCCL_API friend constexpr bool
+  operator==(const common_iterator& __x, const common_iterator<_I2, _S2>& __y)
   {
     auto& __y_hold = __y.__get_hold();
 
@@ -253,9 +254,9 @@ public:
 
 #if _CCCL_STD_VER <= 2017
   _CCCL_EXEC_CHECK_DISABLE
-  template <class _I2 = _Iter>
-  _CCCL_API friend constexpr auto operator==(const common_iterator& __x, const common_iterator& __y)
-    _CCCL_TRAILING_REQUIRES(bool)((!equality_comparable<_I2>) )
+  _CCCL_TEMPLATE(class _I2 = _Iter)
+  _CCCL_REQUIRES((!equality_comparable<_I2>) )
+  [[nodiscard]] _CCCL_API friend constexpr bool operator==(const common_iterator& __x, const common_iterator& __y)
   {
     auto& __y_hold = __y.__get_hold();
 
@@ -278,27 +279,29 @@ public:
     return __x.__hold_.__second_ == __y_hold.__first_;
   }
 
-  template <class _I2 = _Iter>
-  _CCCL_API friend constexpr auto operator!=(const common_iterator& __x, const common_iterator& __y)
-    _CCCL_TRAILING_REQUIRES(bool)((!equality_comparable<_I2>) )
+  _CCCL_TEMPLATE(class _I2 = _Iter)
+  _CCCL_REQUIRES((!equality_comparable<_I2>) )
+  [[nodiscard]] _CCCL_API friend constexpr bool operator!=(const common_iterator& __x, const common_iterator& __y)
   {
     return !(__x == __y);
   }
 
-  template <class _I2, class _S2>
-  _CCCL_API friend constexpr auto operator!=(const common_iterator& __x, const common_iterator<_I2, _S2>& __y)
-    _CCCL_TRAILING_REQUIRES(bool)(sentinel_for<_S2, _Iter>&& sentinel_for<_Sent, _I2>
-                                  && (!equality_comparable_with<_Iter, _I2>) )
+  _CCCL_TEMPLATE(class _I2, class _S2)
+  _CCCL_REQUIRES(
+    sentinel_for<_S2, _Iter> _CCCL_AND sentinel_for<_Sent, _I2> _CCCL_AND(!equality_comparable_with<_Iter, _I2>))
+  [[nodiscard]] _CCCL_API friend constexpr bool
+  operator!=(const common_iterator& __x, const common_iterator<_I2, _S2>& __y)
   {
     return !(__x == __y);
   }
 #endif // _CCCL_STD_VER <= 2017
 
   _CCCL_EXEC_CHECK_DISABLE
-  template <class _I2, class _S2>
-  _CCCL_API friend constexpr auto operator==(const common_iterator& __x, const common_iterator<_I2, _S2>& __y)
-    _CCCL_TRAILING_REQUIRES(bool)(
-      sentinel_for<_S2, _Iter>&& sentinel_for<_Sent, _I2>&& equality_comparable_with<_Iter, _I2>)
+  _CCCL_TEMPLATE(class _I2, class _S2)
+  _CCCL_REQUIRES(
+    sentinel_for<_S2, _Iter> _CCCL_AND sentinel_for<_Sent, _I2> _CCCL_AND equality_comparable_with<_Iter, _I2>)
+  [[nodiscard]] _CCCL_API friend constexpr bool
+  operator==(const common_iterator& __x, const common_iterator<_I2, _S2>& __y)
   {
     auto& __y_hold = __y.__get_hold();
 
@@ -324,9 +327,9 @@ public:
   }
 #if _CCCL_STD_VER <= 2017
   _CCCL_EXEC_CHECK_DISABLE
-  template <class _I2 = _Iter>
-  _CCCL_API friend constexpr auto operator==(const common_iterator& __x, const common_iterator& __y)
-    _CCCL_TRAILING_REQUIRES(bool)(equality_comparable<_I2>)
+  _CCCL_TEMPLATE(class _I2 = _Iter)
+  _CCCL_REQUIRES(equality_comparable<_I2>)
+  [[nodiscard]] _CCCL_API friend constexpr bool operator==(const common_iterator& __x, const common_iterator& __y)
   {
     auto& __y_hold = __y.__get_hold();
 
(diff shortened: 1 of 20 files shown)


libcudacxx/include/cuda/std/__iterator/common_iterator.h
```diff
@@ -0,0 +1,502 @@
+// -*- C++ -*-
+//===----------------------------------------------------------------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES
```
miscco:
```suggestion
// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES
```

libcudacxx/include/cuda/std/__iterator/common_iterator.h
```diff
@@ -0,0 +1,502 @@
...
+  } && indirectly_readable<_Iter> && constructible_from<iter_value_t<_Iter>, iter_reference_t<_Iter>>
+  && move_constructible<iter_value_t<_Iter>>;
+#else // ^^^ _CCCL_HAS_CONCEPTS() ^^^ / vvv !_CCCL_HAS_CONCEPTS() vvv
+template <class _Iter>
+_CCCL_CONCEPT_FRAGMENT(
+  __postfix_can_reference_,
+  requires(_Iter& __it)(requires(__dereferenceable<decltype(__it++)>), requires(__can_reference<decltype(*__it++)>)));
+
+template <class _Iter>
+_CCCL_CONCEPT __postfix_can_reference = _CCCL_FRAGMENT(__postfix_can_reference_, _Iter);
```
miscco: We nowadays have the `_CCCL_REQUIRES_EXPR` macro that cleans a lot of this up.

The only place where we need to be careful is concepts that we want to be able to subsume, e.g. the iterator concepts

libcudacxx/include/cuda/std/__iterator/common_iterator.h
```diff
@@ -0,0 +1,502 @@
...
+          enable_if_t<sentinel_for<_Sent, _Iter>, int>                  = 0,
+          enable_if_t<(!same_as<_Iter, _Sent> && copyable<_Iter>), int> = 0>
+#endif // !_CCCL_HAS_CONCEPTS()
+class common_iterator
+{
+  struct __proxy
+  {
+    iter_value_t<_Iter> __value_;
+
+    _CCCL_API constexpr const iter_value_t<_Iter>* operator->() const noexcept
```
miscco: Applies throughout
```suggestion
    [[nodiscard]] _CCCL_API constexpr const iter_value_t<_Iter>* operator->() const noexcept
```

libcudacxx/include/cuda/std/__iterator/common_iterator.h
```diff
@@ -0,0 +1,502 @@
...
+  _CCCL_REQUIRES(convertible_to<const _I2&, _Iter> _CCCL_AND convertible_to<const _S2&, _Sent> _CCCL_AND
+                   assignable_from<_Iter&, const _I2&> _CCCL_AND assignable_from<_Sent&, const _S2&>)
+  _CCCL_API constexpr common_iterator& operator=(const common_iterator<_I2, _S2>& __other)
+  {
+    __hold_ = __other.__get_hold();
+    return *this;
+  }
+
+  _CCCL_EXEC_CHECK_DISABLE
+  _CCCL_API constexpr decltype(auto) operator*()
```
miscco:
```suggestion
  [[nodiscard]] _CCCL_API constexpr decltype(auto) operator*()
```

libcudacxx/include/cuda/std/__iterator/common_iterator.h
```diff
@@ -0,0 +1,502 @@
...
+      return __hold_.__first_++;
+    }
+    _CCCL_UNREACHABLE();
+  }
+
+  _CCCL_EXEC_CHECK_DISABLE
+  template <class _I2, class _S2>
+  _CCCL_API friend constexpr auto operator==(const common_iterator& __x, const common_iterator<_I2, _S2>& __y)
+    _CCCL_TRAILING_REQUIRES(bool)(sentinel_for<_S2, _Iter>&& sentinel_for<_Sent, _I2>
+                                  && (!equality_comparable_with<_Iter, _I2>) )
```
miscco: In other iterators we did not need to use the trailing workaround anymore, can we try and see whether its still required?

libcudacxx/include/cuda/std/__iterator/common_iterator.h
```diff
@@ -0,0 +1,471 @@
...
+template <input_or_output_iterator _Iter, sentinel_for<_Iter> _Sent>
+  requires(!same_as<_Iter, _Sent> && copyable<_Iter>)
+#else // ^^^ _CCCL_HAS_CONCEPTS() ^^^ / vvv !_CCCL_HAS_CONCEPTS() vvv
+template <class _Iter,
+          class _Sent,
+          enable_if_t<input_or_output_iterator<_Iter>, int>             = 0,
+          enable_if_t<sentinel_for<_Sent, _Iter>, int>                  = 0,
+          enable_if_t<(!same_as<_Iter, _Sent> && copyable<_Iter>), int> = 0>
+#endif // !_CCCL_HAS_CONCEPTS()
+class common_iterator
```
miscco: Important: we probably want to move that into `__fwd/iterator.h`

libcudacxx/include/cuda/std/__iterator/common_iterator.h
```diff
@@ -0,0 +1,471 @@
...
+#else // ^^^ _CCCL_HAS_CONCEPTS() ^^^ / vvv !_CCCL_HAS_CONCEPTS() vvv
+  _CCCL_TEMPLATE(class _I2 = _Iter)
+  _CCCL_REQUIRES(default_initializable<_I2>)
+  _CCCL_API constexpr common_iterator() noexcept(is_nothrow_default_constructible_v<_I2>) {}
+#endif // !_CCCL_HAS_CONCEPTS()
+
+  _CCCL_HIDE_FROM_ABI constexpr common_iterator(const common_iterator&)            = default;
+  _CCCL_HIDE_FROM_ABI constexpr common_iterator& operator=(const common_iterator&) = default;
+  _CCCL_HIDE_FROM_ABI constexpr common_iterator(common_iterator&&)                 = default;
+  _CCCL_HIDE_FROM_ABI constexpr common_iterator& operator=(common_iterator&&)      = default;
```
miscco, line 110: Question: Do we care enough to add the `noexcept()` clauses?

libcudacxx/include/cuda/std/__iterator/common_iterator.h
```diff
@@ -0,0 +1,471 @@
...
+template <input_iterator _Iter, class _Sent>
+struct iterator_traits<common_iterator<_Iter, _Sent>>
+#else // ^^^ _CCCL_HAS_CONCEPTS() ^^^ / vvv !_CCCL_HAS_CONCEPTS() vvv
+template <class _Iter, class _Sent>
+struct iterator_traits<common_iterator<_Iter, _Sent>, enable_if_t<input_iterator<_Iter>>>
+#endif // !_CCCL_HAS_CONCEPTS()
+{
+  using iterator_concept  = conditional_t<forward_iterator<_Iter>, forward_iterator_tag, input_iterator_tag>;
+  using iterator_category = conditional_t<__denotes_forward_iter<_Iter>, forward_iterator_tag, input_iterator_tag>;
+  using pointer           = typename __arrow_type_or_void<_Iter, _Sent>::type;
```
miscco: I believe we could just do
```suggestion
  using pointer           = __iterator_traits_member_pointer_or_arrow_or_void<common_iterator<_Iter, _Sent>>;
```

Because it does not define `pointer` as an alias

libcudacxx/include/cuda/std/__iterator/common_iterator.h
```diff
@@ -0,0 +1,471 @@
...
+struct __arrow_type_or_void<_Iter, _Sent, enable_if_t<__common_iter_has_ptr_op<_Iter, _Sent>>>
+{
+  using type _CCCL_NODEBUG = decltype(::cuda::std::declval<const common_iterator<_Iter, _Sent>&>().operator->());
+};
+
+#if _CCCL_COMPILER(GCC) // GCC breaks with a circular definition here
+template <class _Iter, class _Sent>
+struct __is_primary_std_template<common_iterator<_Iter, _Sent>> : true_type
+{};
+#endif // _CCCL_COMPILER(GCC)
```
miscco, line 431: Lets check if that is still needed

libcudacxx/include/cuda/std/__iterator/common_iterator.h
```diff
@@ -0,0 +1,471 @@
...
+
+#if _CCCL_HAS_CONCEPTS()
+template <input_iterator _Iter, class _Sent>
+struct iterator_traits<common_iterator<_Iter, _Sent>>
+#else // ^^^ _CCCL_HAS_CONCEPTS() ^^^ / vvv !_CCCL_HAS_CONCEPTS() vvv
+template <class _Iter, class _Sent>
+struct iterator_traits<common_iterator<_Iter, _Sent>, enable_if_t<input_iterator<_Iter>>>
+#endif // !_CCCL_HAS_CONCEPTS()
```
miscco: I just unified the definition
```suggestion
template <class _Iter, class _Sent>
struct iterator_traits<common_iterator<_Iter, _Sent>, enable_if_t<input_iterator<_Iter>>>
```

libcudacxx/include/cuda/std/__iterator/common_iterator.h
```diff
@@ -0,0 +1,471 @@
...
+template <class _Iter, class _Sent>
+struct incrementable_traits<common_iterator<_Iter, _Sent>>
+{
+  using difference_type = iter_difference_t<_Iter>;
+};
+
+template <class _Iter>
+_CCCL_CONCEPT __denotes_forward_iter = _CCCL_REQUIRES_EXPR((_Iter), )(
+  typename(typename iterator_traits<_Iter>::iterator_category),
+  requires(derived_from<typename iterator_traits<_Iter>::iterator_category, forward_iterator_tag>));
```
miscco: This is `__has_iterator_category_convertible_to<_Iter, forward_iterator_tag>`

libcudacxx/include/cuda/std/__iterator/variant_like.h
```diff
@@ -0,0 +1,459 @@
...
+class __variant_like : public __vl_destruct_base<_Tp, _Up>
+{
+public:
+  _CCCL_DELEGATE_CONSTRUCTORS(__variant_like, __vl_destruct_base, _Tp, _Up)
+
+#if _CCCL_STD_VER >= 2020
+  _CCCL_HIDE_FROM_ABI constexpr __variant_like(const __variant_like&)
+    requires is_trivially_copy_constructible_v<_Tp> && is_trivially_copy_constructible_v<_Up>
+  = default;
+#endif // _CCCL_STD_VER >= 2020
```
miscco:
```suggestion
#if _CCCL_HAS_CONCEPTS()
  _CCCL_HIDE_FROM_ABI constexpr __variant_like(const __variant_like&)
    requires is_trivially_copy_constructible_v<_Tp> && is_trivially_copy_constructible_v<_Up>
  = default;
#endif // _CCCL_HAS_CONCEPTS()
```

libcudacxx/include/cuda/std/__iterator/variant_like.h
```diff
@@ -0,0 +1,459 @@
...
+      case __variant_like_state::__nothing:
+        break;
+    }
+  }
+
+#if _CCCL_STD_VER >= 2020
+  _CCCL_HIDE_FROM_ABI constexpr __variant_like(__variant_like&&)
+    requires is_trivially_move_constructible_v<_Tp> && is_trivially_move_constructible_v<_Up>
+  = default;
+#endif // _CCCL_STD_VER >= 2020
```
miscco:
```suggestion
#if _CCCL_HAS_CONCEPTS()
  _CCCL_HIDE_FROM_ABI constexpr __variant_like(__variant_like&&)
    requires is_trivially_move_constructible_v<_Tp> && is_trivially_move_constructible_v<_Up>
  = default;
#endif // _CCCL_HAS_CONCEPTS()
```

libcudacxx/test/libcudacxx/std/iterators/predef.iterators/iterators.common/assign.pass.cpp
```diff
@@ -0,0 +1,166 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES
```
miscco:
```suggestion
// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES
```

Applies throughout

libcudacxx/test/libcudacxx/std/iterators/predef.iterators/iterators.common/ctor.sentinel.pass.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+#if TEST_STD_VER >= 2020
+  static_assert(test<cpp17_input_iterator<int*>>());
+  static_assert(test<forward_iterator<int*>>());
+  static_assert(test<bidirectional_iterator<int*>>());
+  static_assert(test<random_access_iterator<int*>>());
+  static_assert(test<contiguous_iterator<int*>>());
+  static_assert(test<int*>());
+  static_assert(test<const int*>());
+#endif
```
miscco: I see the other test does not have that constraint. can we get rid of it?

```suggestion
#endif // TEST_STD_VER >= 2020
```

libcudacxx/test/libcudacxx/std/iterators/predef.iterators/iterators.common/deref.pass.cpp
```diff
@@ -0,0 +1,161 @@
...
+    auto iter1       = simple_iterator<int*>(buffer);
+    auto commonIter1 = cuda::std::common_iterator<decltype(iter1), sentinel_type<int*>>(iter1);
+    auto commonSent1 =
+      cuda::std::common_iterator<decltype(iter1), sentinel_type<int*>>(sentinel_type<int*>{buffer + 8});
+
+    const auto iter2       = simple_iterator<int*>(buffer);
+    const auto commonIter2 = cuda::std::common_iterator<decltype(iter1), sentinel_type<int*>>(iter1);
+
+    assert(*iter1 == 1);
+    assert(*commonIter1 == 1);
```
miscco, line 38: This is missing a check that validates the return value of `operator*` through a static asser. We can also add a noexcept check

libcudacxx/test/libcudacxx/std/iterators/predef.iterators/iterators.common/minus.pass.cpp
```diff
@@ -0,0 +1,71 @@
...
+TEST_FUNC constexpr bool test()
+{
+  int buffer[8] = {1, 2, 3, 4, 5, 6, 7, 8};
+
+  {
+    auto iter1       = random_access_iterator<int*>(buffer);
+    auto commonIter1 = cuda::std::common_iterator<decltype(iter1), sized_sentinel_type<int*>>(iter1);
+    auto commonSent1 =
+      cuda::std::common_iterator<decltype(iter1), sized_sentinel_type<int*>>(sized_sentinel_type<int*>{buffer + 8});
+    assert(commonIter1 - commonSent1 == -8);
```
miscco, line 41: Ditto missing noexcept + return type validation

libcudacxx/test/libcudacxx/std/iterators/predef.iterators/iterators.common/plus_plus.pass.cpp
```diff
@@ -0,0 +1,201 @@
...
+  // Reference: http://eel.is/c++draft/iterators.common#common.iter.nav-5
+  // Case 2: can-reference
+  {
+    int buffer[8]    = {1, 2, 3, 4, 5, 6, 7, 8};
+    auto iter1       = simple_iterator<int*>(buffer);
+    auto commonIter1 = cuda::std::common_iterator<decltype(iter1), sentinel_type<int*>>(iter1);
+    auto commonSent1 =
+      cuda::std::common_iterator<decltype(iter1), sentinel_type<int*>>(sentinel_type<int*>{buffer + 8});
+
+    assert(*(commonIter1++) == 1);
```
miscco, line 35: Ditto

miscco:
We should also specialize in `thrust/thrust/iterator/iterator_traits.h`
```cpp
template <class Iter, class Sent>
struct iterator_system<::cuda::std::common_iterator<Iter, Sent>> : iterator_system<Iter>
{};
template <class Iter, class Sent>
struct iterator_traversal<::cuda::std::common_iterator<Iter, Sent>> : iterator_traversal<Iter>
{};
```

libcudacxx/include/cuda/std/__iterator/common_iterator.h
Jacobfaib: That would require putting `__iterator/concepts.h` and several type-traits in the fwd declarations as well, which could get hefty. Currently it's just `<cuda/std/__cccl/prologue.h>` in there. Are you sure?

miscco: I mean we already include those in the `<cuda/__wd/iterator.h>` so I am fine with that

libcudacxx/include/cuda/std/__iterator/common_iterator.h
Jacobfaib: `= default` takes care of this for us already. `constexpr` is actually also superfluous, but I guess we keep that there to ensure that compiler do it. I believe adding these is what ends up tripping up the older compilers (perhaps by adding them explicitly they are evaluated more eagerly leading to bugs).

libcudacxx/include/cuda/std/__iterator/common_iterator.h
Jacobfaib: Yeah, much better. Am always on the hunt for stuff that has already been implemented in other traits!

libcudacxx/include/cuda/std/__iterator/common_iterator.h
Jacobfaib: Commented it out, let's see

libcudacxx/include/cuda/std/__iterator/variant_like.h
Jacobfaib: Fixed (everywhere else too). Actually I had meant to ask, how are these not duplicate definitions when concepts are enabled? We have
```c++
_CCCL_HIDE_FROM_ABI constexpr __variant_like(const __variant_like&)
  ...
= default;
```
and then
```c++
CCCL_API _CCCL_CONSTEXPR_CXX20 __variant_like(const __variant_like& __other) 
  ...
{
  ...
}
```
right below it. Shouldn't these be mutually exclusive?

libcudacxx/test/libcudacxx/std/iterators/predef.iterators/iterators.common/ctor.sentinel.pass.cpp
Jacobfaib: No, GCC 7 and 11 didn't like it. The __variant_like constructors/destructors aren't constexpr because of the construct_at shenanigans. I'm actually not sure why it only complained for this one.

libcudacxx/include/cuda/std/__iterator/common_iterator.h
Jacobfaib: I guess still needed?
```
/home/coder/cccl/lib/cmake/libcudacxx/../../../libcudacxx/include/cuda/std/__type_traits/is_primary_template.h:56:83: error: ‘iterator_traits’ was not declared in this scope; did you mean ‘std::iterator_traits’?
     56 | using __is_primary_cccl_template = _IsValidExpansion<__test_for_primary_template, iterator_traits<_Iter>>;
        |                                                                                   ^~~~~~~~~~~~~~~
        |                                                                                   std::iterator_traits
  In file included from /usr/include/c++/13/bits/stl_algobase.h:61,
                   from /usr/include/c++/13/vector:62,
                   from /home/coder/cccl/c/experimental/stf/test/test_logical_data.cpp:11:
  /usr/include/c++/13/bits/cpp_type_traits.h:470:29: note: ‘std::iterator_traits’ declared here
    470 |   template<typename> struct iterator_traits;
        |                             ^~~~~~~~~~~~~~~
```
(https://github.com/NVIDIA/cccl/actions/runs/24892595022/job/72888704418?pr=8603)

libcudacxx/test/libcudacxx/std/iterators/predef.iterators/iterators.common/deref.pass.cpp
Jacobfaib: Fixed, added return type checks, `noexcept` will do next

libcudacxx/include/cuda/std/__iterator/common_iterator.h
Jacobfaib: Ah I think I might have diagnosed the root problem. This (and the original GCC problem) are the results of partial include cycles. The new `__fwd/iterator.h` now transitively includes `cuda/std/__type_traits/is_primary_template.h` which means that the contents of that are placed *above* the forward declarations of `iterator_traits` and `__is_primary_std_template`.

I don't believe we can solve this by locally pre-declaring `iterator_traits` and friends either, because `iterator_traits` has a default argument in the second parameter. It's OK to (re) predeclare it as
```c++
template <class, class>
class iterator_traits;
```
_If_ we didn't use the default value immediately in the `is_primary_template` code:
```c++
using ... = ... iterator_traits<It>; // requires knowing that second arg has default
```
which if `__fwd/iterator.h` is included first, it wouldn't.

Jacobfaib: So unfortunately, I think the fix is that we might need a separate `__fwd` file for `common_iterator`.

miscco: fine by me

libcudacxx/include/cuda/std/__iterator/common_iterator.h
miscco: The point is that `=default` is not correct about `noexcept` in most compilers we support ;)

libcudacxx/include/cuda/std/__iterator/common_iterator.h
```diff
@@ -0,0 +1,451 @@
...
+  iter_move(const common_iterator& __i) noexcept(noexcept(::cuda::std::ranges::iter_move(declval<const _Iter&>())))
+  {
+    _CCCL_ASSERT(__i.__hold_.__holds_first(), "Attempted to iter_move a non-dereferenceable common_iterator");
+    return ::cuda::std::ranges::iter_move(__i.__hold_.__first_);
+  }
+
+  _CCCL_TEMPLATE(class _I2, class _S2)
+  _CCCL_REQUIRES(indirectly_swappable<_I2, _Iter>)
+  _CCCL_API friend constexpr void iter_swap(const common_iterator& __x, const common_iterator<_I2, _S2>& __y) noexcept(
+    noexcept(::cuda::std::ranges::iter_swap(::cuda::std::declval<const _Iter&>(), ::cuda::std::declval<const _I2&>())))
```
miscco, line 400: I believe we have `__noexcept_swappable<_Iter, _I2>` We should add the same for `iter_move` at some point in the future

miscco (approved):

Issue #8433, opened 2026-04-15
Add env-based API for cub::SegmentedScan


gonidelis:

gonidelis:
close by #8614

PR #8609, opened 2026-04-21
Workaround `subrange.h`  nvcc+clang bug

fbusato:
## Description

Refactor default template argument for subrange class to resolve nvcc+clang bug.

PR #8610, opened 2026-04-22
Add env overloads and tests for DeviceSegmentedScan (non separate offsets)

gonidelis:
fixes first half of #8433 

Adds env overloads for non-separate offsets cub::DeviceSegmentedScan APIs


gonidelis:
wow fails for cuda 13.0 and not for 13.1 - looks like some sorta frontend error

bernhardmgruber:
> wow fails for cuda 13.0 and not for 13.1 - looks like some sorta frontend error

That's an unrelated bug. PR is up: #8609

miscco:
No a bug, but a sccache preprocessor issue

miscco (approved):

PR #8613, opened 2026-04-22
[cudax] Implement generic group in level queries

### LINKED ISSUE #8591: [FEA]: Add support for `group.query(level)`
Currently, this is implemented only for _this_ groups. We want all groups to support this.

davebayer:
Implements #8591.

### DIFF
--- a/cudax/include/cuda/experimental/__group/group.cuh
+++ b/cudax/include/cuda/experimental/__group/group.cuh
@@ -23,6 +23,8 @@
 
 #include <cuda/__bit/bitmask.h>
 #include <cuda/__cmath/pow2.h>
+#include <cuda/__hierarchy/queries/count.h>
+#include <cuda/__hierarchy/queries/rank.h>
 #include <cuda/barrier>
 #include <cuda/hierarchy>
 #include <cuda/std/__concepts/concept_macros.h>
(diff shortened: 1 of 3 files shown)


cudax/include/cuda/experimental/__group/group.cuh
```diff
@@ -177,6 +179,45 @@ public:
...
+      __ret *= __count_query<level_type, _InLevel>::template __call<_Tp>(__hier_);
+    }
+    return __ret;
+  }
+
+  _CCCL_TEMPLATE(class _InLevel)
+  _CCCL_REQUIRES(__is_hierarchy_level_v<_InLevel>)
+  [[nodiscard]] _CCCL_DEVICE_API constexpr ::cuda::std::size_t count(const _InLevel& __in_level) const noexcept
+  {
+    return count_as<::cuda::std::size_t>(__in_level);
```
pciolkosz, line 199: Hmm, should it be smarter about the default? I think it should be level dependent, we already have `__product_type` in level type

pciolkosz (approved):

cudax/include/cuda/experimental/__group/group.cuh
davebayer: I will fix that in #8637

PR #8619, opened 2026-04-22
Fix cudax with clang-15

davebayer:


cudax/include/cuda/experimental/__places/localized_array.cuh
```diff
@@ -91,7 +91,8 @@ public:
 
     CUmemAllocationProp prop = {};
     prop.type                = CU_MEM_ALLOCATION_TYPE_PINNED;
-    prop.location            = {.type = CU_MEM_LOCATION_TYPE_DEVICE, .id = dev};
```
davebayer, line 94: This is not even standard in C++17

davebayer:
fixed directly in #8620

PR #8495, opened 2026-04-16
Replace `detail::scan::dispatch` by CUB's public API

bernhardmgruber:
- [x] Merge before: #8473
- [x] Postponed:  #6576
- [x] No SASS changes for `cub.bench.exclusive.sum.baes` for SM`75,80,90,100,120`

Fixes: #8373

## Summary by CodeRabbit

* **Refactor**
  * Simplified device scan benchmarks with modernized internal API patterns.
  * Device scan dispatch refactored to support flexible runtime policy selection for improved tuning capabilities.
  * Expanded test coverage for inclusive and exclusive scan initialization operations.

* **Chores**
  * Added implementation notes marking areas for future optimization work.

Issue #7641, opened 2026-02-11
Implement the new tuning API for `DispatchFixedSizeSegmentedReduce`


bernhardmgruber:
Should wait for

- [x] #7718

bernhardmgruber:
@srinivasyadav18 told me that he plans to entirely get rid of `DispatchFixedSizeSegmentedReduce` and integrate it into `segmented_reduce::dispatch`, which would render this issue obsolete. I will keep it open until the restructuring is done.

Issue #7165, opened 2026-01-12
[EPIC] Implement the new tuning API for CUB


bernhardmgruber:
We designed a new tuning API for CUB with the goal to simplify the internal implementation of the CUB dispatchers, improve the CCCL.C integration, and eventually allow users to supply custom tunings to CUB algorithms. The design and first implementation to test the feasibility are documented here: #6368. The entire design is described in this document: https://docs.google.com/document/u/1/d/1KtrSsMk2LAXSB5cdAUnmn3BMow5y6B8BhC5-XVA434g

This issue tracks the implementation status across CUB.

The new tuning API has been implemented for:

 - [ ] DispatchAdjacentDifference
 - [ ] DispatchBatchMemcpy
 - [ ] detail::for_each::dispatch_t
 - [ ] DispatchHistogram
 - [ ] detail::merge::dispatch_t
 - [ ] DispatchMergeSort
 - [ ] DispatchRadixSort
 - [x] #6544
 - [ ] DispatchReduceByKey
 - [ ] DispatchReduceDeterministic
 - [ ] DispatchReduceNondeterministic
 - [ ] DeviceRleDispatch
 - [ ] DispatchScan
 - [ ] DispatchScanByKey
 - [ ] DispatchSegmentedRadixSort
 - [ ] DispatchSegmentedReduce
 - [ ] DispatchFixedSizeSegmentedReduce
 - [ ] DispatchSegmentedSort
 - [ ] DispatchSelectIf
 - [ ] detail::reduce::dispatch_streaming_arg_reduce_t
 - [ ] DispatchStreamingReduceByKey
 - [ ] DispatchThreeWayPartitionIf
 - [ ] DispatchTopK
 - [ ] #6914
 - [ ] DispatchUniqueByKey

Further tasks:

- [ ] #6677
- [ ] Change/Extend all CUB algorithms to provide access to the new tuning API
- [ ] Replace all internal dispatch layer access by public CUB API calls including the new tuning API

We concluded to not allow public access to the tuning API for now. We can however use the new tuning API already internally. This should give us time to iron out any issued with the design. After all above tasks are complete, we should:

- [ ] Enable public access to tuning properties for CUB environment API overloads
- [ ] Deprecated all public CUB dispatchers (dropped at the next major release)

bernhardmgruber:
I have tried to approach the reduce by key dispatchers from a few angles, but it seems we have to refactor #7531, #7532 and #7533 together, since they interdepend on each other.

[...]

PR #8623, opened 2026-04-22
Document public CUB tuning API

### LINKED ISSUE #8593: Document public tuning of CUB algorithms
We should document how tunings can be specified by our users to CUB algorithms as part of the public CUB tuning documentation (not the developer guide, which is handled by #8592). We should explain policy selectors and tuning policies, as well as the mechanics to incorporate them into environments that are passed to CUB algorithms.

bernhardmgruber:
I let claude rewrite the documentation and did a lot of manual editing on top. Here is a summary by claude claude:

  - Updated `docs/cub/developer/device_scope.rst` to document the new dispatch architecture: `policy_selector` + `dispatch_arch` replacing `ChainedPolicy`/`MaxPolicy`/`PolicyHub`                                   
  - Created `docs/cub/policy_selectors.rst` explaining how users define and pass custom policy selectors via `cuda::execution::tune`                                                                                 
  - Updated `docs/cub/tuning.rst` so benchmarks call the public CUB API with an environment instead of instantiating dispatch structs with custom policy hubs                                                        
                                                                                                                                                                                                                     
  This follows the roll-out of the new tuning API across ~20 CUB dispatchers (#7165). The old docs described `ChainedPolicy` linked lists, the `MaxPolicy` kernel trick, and template-specialization-based tunings.  
  The new architecture uses stateless `constexpr` callables returning plain aggregate policy structs, `dispatch_arch` for runtime-to-compile-time arch dispatch, and kernels querying                                
  `PolicySelector{}(arch_id{CUB_PTX_ARCH / 10})` directly

After some design reconsideration, I changed all uses of the term architecture by compute capability.

<details>
  <summary>Prompts:</summary>
  
Please look at the github issue https://github.com/NVIDIA/cccl/issues/7165.
It has many sub issues attached that are named something like "Implement the new tuning API for".
I paste the list here as well: 

Implement the new tuning API for DispatchAdjacentDifference #7522
bernhardmgruber
Implement the new tuning API for detail::for_each::dispatch_t #7525
bernhardmgruber
Implement the new tuning API for DeviceRleDispatch #7532
bernhardmgruber
Implement the new tuning API for detail::merge::dispatch_t #7638
bernhardmgruber
Implement the new tuning API for detail::reduce::dispatch_streaming_arg_reduce_t #7645
bernhardmgruber
Implement the new tuning API for DispatchReduceByKey #7531
bernhardmgruber
Implement the new tuning API for detail::find::dispatch_t #7812
bernhardmgruber
Implement the new tuning API for DispatchSegmentedRadixSort #7642
bernhardmgruber
Implement the new tuning API for DispatchStreamingReduceByKey #7533
bernhardmgruber
Implement the new tuning API for DispatchTopK #7648
bernhardmgruber
Implement the new tuning API for DispatchSegmentedSort #7643
bernhardmgruber
Implement the new tuning API for DispatchThreeWayPartitionIf #7646
bernhardmgruber
Implement the new tuning API for DeviceScan #7521
griwes
Implement the new tuning API for DispatchMergeSort #7639
bernhardmgruber
Implement the new tuning API for DispatchBatchMemcpy #7636
bernhardmgruber
Implement the new tuning API for DispatchHistogram #7637
bernhardmgruber
Implement the new tuning API for DispatchSelectIf #7644
bernhardmgruber
Implement the new tuning API for DispatchScanByKey #7640
griwes
Implement the new tuning API for DispatchUniqueByKey #7647
griwes
Implement the new tuning API for detail::batched_topk::dispatch_batched_topk #8479
bernhardmgruber
Implement the new tuning API for DispatchFixedSizeSegmentedReduce #7641
srinivasyadav18
Implement the new tuning API for detail::segmented_scan::dispatch_segmented_scan #8424
oleksandr-pavlyk

Go to each of the linked issues and find the linked pull request. if it's merged, get the git SHA of the commit that resulted from the merge. Please print the list of commit SHAs.

Now, please look at the documentation page docs/cub/developer/device_scope.rst. It documents how the CUB device level API looks like, and how it calls through the dispatch layer to a kernel. It also documents how tuning policies work, etc. This documentation is mostly obsolete based on all the pull requests above. Please look at the changes of all the git SHAs above and try to understand how the new dispatch logic works. Then please update the device_scope.rst documentation page to reflect the new architecture of CUB.

Before all the changes, I drafted a document describing the new API (a design document). I placed a copy for you into ./new_api.md. Please read that document as well. It should help you understand the changes to the documentation.

Please also update docs/cub/tuning.rst, the section titled Policy Hub. It should discuss policy selectors instead. Instead of showing how to instantiate the Dispatch struct, show how to pass a policy selector to the public CUB API. An example can be found in this PR: https://github.com/NVIDIA/cccl/pull/8495. Just show calling the public API like `cub::DeviceScan::InclusiveScan(..., /* env = */ cuda::execution::tune(policy_selector{}));`. Please also update the sections titled Main Body to call the public API instead. Also the section titled Creating tuning policies.

Please move the section "Policy Selector" to a new page between the TOC entries "CUB Benchmarks" and "CUB Tunings". Add a link to that page where the section was.

I have reworded some things. See the last two commits. I have not touched the device_scope.rst changes yet. They go to a different PR. Please check what is added by the last two commits only and suggest improvements and reword where necessary.

Just apply the changes, I can look at the diff
  
</details>

I also let claude rewrite the developer documentation, but that will come as a separate PR.

- [x] Merge before: #8728 

Fixes: #8593

### DIFF
--- a/docs/cub/policy_selectors.rst
+++ b/docs/cub/policy_selectors.rst
@@ -13,7 +13,7 @@ e.g. :code:`cub::ReducePolicy`, which must be returned by a policy selector pass
 CUB provides default policy selectors providing tunings for known architectures and workloads,
 which are not publicly accessible to users, but used internally by CUB to select tunings.
 Users can override CUB's policy selector for a given algorithm
-by passing a custom policy selector through the algorithm's environment parameter,
+by passing a custom policy selector through the algorithm's environment parameter.
 
 For a description of how policy selectors are used internally in CUB's dispatch layer and kernels
 see the corresponding :ref:`developer documentation <cub-developer-guide-device-scope>`.
@@ -55,9 +55,9 @@ taking a ``cuda::arch_id`` and returning the algorithm's policy struct:
       }
     };
 
-The policy selector must be stateless (:coda:`std::is_empty_v<T>` must be :coda:`true`),
+The policy selector must be stateless (:code:`std::is_empty_v<T>` must be :code:`true`),
 since only its type will be passed to a kernel and it will be default-constructed in device code.
-It can be a class template, but then only a full specialization can be passed to cub,
+It can be a class template, but then only a full specialization can be passed to CUB,
 e.g., :code:`my_reduce_tuning<float>`.
 The implementation can use branches and helper functions arbitrarily,
 as long as they can be evaluated at compile-time.
@@ -73,7 +73,7 @@ and each algorithm will pick the one with the matching policy struct.
 This is useful if the same environment is reused across several algorithm calls.
 
 The policy structs themselves aim to be simple aggregates without special member functions (constructors, assignment operators, etc.),
-thus supporting C++20 designated initializers (i.e., the syntax :code:`{ .block_threads = 512, ... }`.
+thus supporting C++20 designated initializers (i.e., the syntax :code:`{ .block_threads = 512, ... }`).
 They may occasionally contain member functions that compute derived values from the contained tuning values.
 They support comparison for (in-)equality and serialization using :code:`operator<<`.
 All policy structs are public types and will try to evolve in a non-breaking way,
@@ -104,7 +104,7 @@ Multiple tunings for different algorithms can be combined in a single environmen
     cub::DeviceScan::ExclusiveSum(d_in, d_out, num_items, env);
 
 An environment can carry further properties like a stream or a memory resource.
-Policy selectors can simply by added to those:
+Policy selectors can simply be added to those:
 
 .. code:: c++
 
--- a/docs/cub/policy_selectors.rst
+++ b/docs/cub/policy_selectors.rst
@@ -74,9 +74,9 @@ Multiple policy selectors returning different policy structs can be passed as pa
 and each algorithm will pick the one with the matching policy struct.
 This is useful if the same environment is reused across several algorithm calls.
 
-The policy structs themselves are simple aggregates without special member functions (no constructors, assignment operators, etc.).
+The policy structs themselves are simple semiregular aggregates.
 They support C++20 designated initializers (i.e., the syntax :code:`{ .block_threads = 512, ... }`),
-support comparison for (in-)equality and serialization using :code:`operator<<`.
+comparison for (in-)equality, and serialization using :code:`operator<<`.
 They may occasionally contain member functions that compute derived values from the contained tuning values.
 All policy structs are public types and will evolve in a non-breaking way, at least during minor releases,
 by only adding new data members at the end of the struct.
--- a/docs/cub/policy_selectors.rst
+++ b/docs/cub/policy_selectors.rst
@@ -1,0 +1,126 @@
+.. _cub-policy-selectors:
+
+CUB Tuning Policy Selectors
+================================================================================
+
+Policy selectors are the mechanism by which CUB's device algorithms and kernels select
+tuning parameters for a given GPU architecture and workload.
+A policy selector is a stateless callable that maps a :code:`cuda::arch_id` to a policy struct
+containing the tuning values for that architecture.
+Each set of CUB algorithms using a common underlying implementation defines a common policy struct,
+e.g. :code:`cub::ReducePolicy`, which must be returned by a policy selector passed to those algorithms.
+
+CUB provides default policy selectors providing tunings for known architectures and workloads,
+which are not publicly accessible to users, but used internally by CUB to select tunings.
+Users can override CUB's policy selector for a given algorithm
+by passing a custom policy selector through the algorithm's environment parameter,
+
+For a description of how policy selectors are used internally in CUB's dispatch layer and kernels
+see the corresponding :ref:`developer documentation <cub-developer-guide-device-scope>`.
+
+Defining a policy selector
+--------------------------------------------------------------------------------
+
+A policy selector is any type with a :code:`__host__`,  :code:`__device__`, and :code:`constexpr` call operator
+taking a ``cuda::arch_id`` and returning the algorithm's policy struct:
+
+..
+    TODO(bgruber): we have to update this example again after the reduce tuning API has been actually released
+
+.. code:: c++
+
+    template <typename T>
+    struct my_reduce_tuning {
+      __host__ __device__ constexpr auto operator()(cuda::arch_id arch) const -> cub::ReducePolicy {
+        // tuning for Hopper and later
+        if (arch >= cuda::arch_id::sm_90) {
+          const auto policy = cub::detail::agent_reduce_policy{
+            .block_threads = 512,
+            .items_per_thread = 64 / sizeof(T), // 8 double, 16 float, 32 half_t, ...
+            .vector_load_length = 2,
+            .block_algorithm = cub::BLOCK_REDUCE_WARP_REDUCTIONS,
+            .load_modifier = cub::LOAD_DEFAULT
+          };
+          return {policy, policy, policy, policy};
+        }
+        // fallback for older architectures
+        const auto policy = cub::detail::agent_reduce_policy{
+          .block_threads = 256,
+          .items_per_thread = 12,
+          .vector_load_length = 1,
+          .block_algorithm = cub::BLOCK_REDUCE_WARP_REDUCTIONS,
+          .load_modifier = cub::LOAD_DEFAULT
+        };
+        return {policy, policy, policy, policy};
+      }
+    };
+
+The policy selector must be stateless (:coda:`std::is_empty_v<T>` must be :coda:`true`),
+since only its type will be passed to a kernel and it will be default-constructed in device code.
+It can be a class template, but then only a full specialization can be passed to cub,
+e.g., :code:`my_reduce_tuning<float>`.
+The implementation can use branches and helper functions arbitrarily,
+as long as they can be evaluated at compile-time.
+The returned tuning policy can, and probably should, contain different values for different CUDA architectures or workloads.
+
+Each CUB algorithm with an environment parameter searches the environment for a policy selector returning a matching policy struct.
+If one is found, the policy selector will be used to determine the tuning values for host-side dispatch and kernel compilation.
+Each CUB algorithm documents the policy struct to which it responds.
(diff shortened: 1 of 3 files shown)


docs/cub/policy_selectors.rst
```diff
@@ -0,0 +1,126 @@
...
+
+Policy selectors are the mechanism by which CUB's device algorithms and kernels select
+tuning parameters for a given GPU architecture and workload.
+A policy selector is a stateless callable that maps a :code:`cuda::arch_id` to a policy struct
+containing the tuning values for that architecture.
+Each set of CUB algorithms using a common underlying implementation defines a common policy struct,
+e.g. :code:`cub::ReducePolicy`, which must be returned by a policy selector passed to those algorithms.
+
+CUB provides default policy selectors providing tunings for known architectures and workloads,
+which are not publicly accessible to users, but used internally by CUB to select tunings.
```
Jacobfaib: This sentence repeats "provides"/"providing" and "tunings" quite a bit, it's a little tricky to read. I would suggest rewording this slightly.

docs/cub/policy_selectors.rst
```diff
@@ -0,0 +1,126 @@
...
+Users can override CUB's policy selector for a given algorithm
+by passing a custom policy selector through the algorithm's environment parameter.
+
+For a description of how policy selectors are used internally in CUB's dispatch layer and kernels
+see the corresponding :ref:`developer documentation <cub-developer-guide-device-scope>`.
+
+Defining a policy selector
+--------------------------------------------------------------------------------
+
+A policy selector is any type with a :code:`__host__`,  :code:`__device__`, and :code:`constexpr` call operator
```
Jacobfaib: Does the call operator have to be `__host__ __device__` or can you have separate call operators for `__host__` and `__device__`? This should clarify this (and ideally show an example if the latter).

Should also clarify whether the call operator must be `const` or not. I suspect it probably doesn't matter because these are just used as stateless functors, but the example has `const` so it kind of implies it's required.

docs/cub/policy_selectors.rst
```diff
@@ -0,0 +1,126 @@
...
+
+The policy selector must be stateless (:code:`std::is_empty_v<T>` must be :code:`true`),
+since only its type will be passed to a kernel and it will be default-constructed in device code.
+It can be a class template, but then only a full specialization can be passed to CUB,
+e.g., :code:`my_reduce_tuning<float>`.
+The implementation can use branches and helper functions arbitrarily,
+as long as they can be evaluated at compile-time.
+The returned tuning policy can, and probably should, contain different values for different CUDA architectures or workloads.
+
+Each CUB algorithm with an environment parameter searches the environment for a policy selector returning a matching policy struct.
```
Jacobfaib, line 68: Consider explaining how this matching is actually done. In particular:

1. Does the user need to do any additional work (for example declare any additional specializations) to enable this matching?
2. Does the user need to modify the environment itself to make it match?

docs/cub/policy_selectors.rst
```diff
@@ -0,0 +1,126 @@
...
+Each CUB algorithm documents the policy struct to which it responds.
+If CUB does not find a matching policy selector in the environment, it falls back to its internal default policy selector.
+
+Multiple policy selectors returning different policy structs can be passed as part of the same environment,
+and each algorithm will pick the one with the matching policy struct.
+This is useful if the same environment is reused across several algorithm calls.
+
+The policy structs themselves aim to be simple aggregates without special member functions (constructors, assignment operators, etc.),
+thus supporting C++20 designated initializers (i.e., the syntax :code:`{ .block_threads = 512, ... }`).
+They may occasionally contain member functions that compute derived values from the contained tuning values.
```
Jacobfaib, line 80: Whether they contain member functions is irrelevant for users, the only guaranteed interaction they have is with the constructor, the existence of any other functions should be considered an implementation detail.

docs/cub/policy_selectors.rst
```diff
@@ -0,0 +1,126 @@
...
+They support comparison for (in-)equality and serialization using :code:`operator<<`.
+All policy structs are public types and will try to evolve in a non-breaking way,
+by only adding new data members at the end of the struct.
```
Jacobfaib: This is a little wishy-washy. Committing to backwards compat _is_ relevant, but how we do that is not. Perhaps this is best left out, since it's not clear whether we can expect these types to be fully API or ABI compatible.

docs/cub/policy_selectors.rst
```diff
@@ -0,0 +1,126 @@
...
+
+CUB Tuning Policy Selectors
+================================================================================
+
+Policy selectors are the mechanism by which CUB's device algorithms and kernels select
+tuning parameters for a given GPU architecture and workload.
+A policy selector is a stateless callable that maps a :code:`cuda::arch_id` to a policy struct
+containing the tuning values for that architecture.
+Each set of CUB algorithms using a common underlying implementation defines a common policy struct,
+e.g. :code:`cub::ReducePolicy`, which must be returned by a policy selector passed to those algorithms.
```
davebayer: Can we wait with this PR until we agree on how exactly the tuning policy selector should look like?

docs/cub/policy_selectors.rst
```diff
@@ -0,0 +1,126 @@
...
+
+    // alternatively, if we want to extend an env `other_env`
+    auto env = cuda::std::execution::env{
+      other_env,
+      cuda::execution::tune(my_reduce_tuning{})
+    };
+    cub::DeviceReduce::Reduce(d_in, d_out, num_items, op, init, env);
+
+CUB's benchmarks also make heavy use of policy selectors for tuning.
+For more details on authoring benchmarks and their policy selectors, see :ref:`cub-tuning`.
```
Jacobfaib, line 129: This seems like developer-facing text, not user-facing. Consider rewording this to the effect of "for detailed examples of policy selectors, see benchmarks...". But perhaps better if we can point to examples instead. Benchmarks aren't so pedagogically useful as they tend to be pretty dense.

davebayer (changes requested):

docs/cub/policy_selectors.rst
bernhardmgruber: Alright, I see the necessity to wait on that design decision.

docs/cub/tuning.rst
```diff
@@ -554,68 +544,66 @@ was thus compiled with ``-DTUNE_ITEMS_PER_THREAD=19 -DTUNE_THREADS_PER_BLOCK=512
 The meaning of these values is specific to the benchmark definition,
 and we have to check the benchmark’s source code for how they are applied.
 Equally named tuning parameters may not translate to different benchmarks (please double check).
-These tuning parameters are then typically used to create a policy hub,
-which is passed to the algorithm’s dispatcher, as :ref:`sketched above <cub-tuning-authoring-benchmarks>`,
-and repeated here:
+
+As a user of CUB,
+such a new set of tuning parameters (i.e. a variant) is then typically used to define a policy selector,
```
Jacobfaib: Some wonky text wrapping

docs/cub/policy_selectors.rst
bernhardmgruber: There must be a single one being `__host__ __device__ constexpr` and `const` annotated.

docs/cub/policy_selectors.rst
bernhardmgruber: Policy structs are fully public, so any member function they have is part of the public API. I am trying to minimize such member functions, so maybe I can eventually get away with none.

docs/cub/policy_selectors.rst
bernhardmgruber: I think it is relevant how we plan to extend then, because users need to know in which ways they are allowed to use policy structs. I rephrased the paragraph a little.

docs/cub/policy_selectors.rst
bernhardmgruber: Reworded.

docs/cub/policy_selectors.rst
bernhardmgruber: Here I would say the user does not need to know exactly how the machinery works. I provide a few examples later that show how a user composes such an environment. Furthermore, @gevtushenko will write ample documentation about how environments and their properties will work (#7712) ;)

docs/cub/policy_selectors.rst
bernhardmgruber: I wanted to link to the benchmark documentation here because it provides further examples of how policy selectors can be used, and emphasize that they take the role of tuning there. Do you think this would be better:

> Further examples for how to use policy selectors can be found in the tuning documentation.

docs/cub/tuning.rst
bernhardmgruber: Reworded.

bernhardmgruber:
@davebayer I made the changes to use the term compute capability. Please re-review. Thx!

docs/cub/policy_selectors.rst
```diff
@@ -0,0 +1,129 @@
+.. _cub-policy-selectors:
+
+CUB Tuning Policy Selectors
+================================================================================
+
+Policy selectors are the mechanism by which CUB's device algorithms and kernels select
+tuning parameters for a given workload and GPU compute capability.
```
jrhemstad, line 7: suggestion: I think we need a higher-level preamble here that explains what tuning parameters are in the first place. If this is intended to be the starting point for users to learn about user-defined tuning, we need some higher level prose that explains what tuning parameters are, how they intersect with algorithm implementations, etc.

bernhardmgruber: There is a lot in the Tuning documentation we already have: https://nvidia.github.io/cccl/unstable/cub/tuning.html. But maybe we need another rework to streamline how the information is presented. The current docs (without this PR) are very focused on the tuning process. This PR mainly tries to introduce the new tuning aspect of CUB's API.

bernhardmgruber: Let's address this as a follow-up: #8731

davebayer (approved):

PR #8625, opened 2026-04-22
[Tile] Try to use `assert` instead of `__trap` for tile mode

miscco:
Trap is currently not available in tile mode.

See nvbug6072716

### DIFF
--- a/libcudacxx/include/cuda/std/__exception/terminate.h
+++ b/libcudacxx/include/cuda/std/__exception/terminate.h
@@ -22,6 +22,10 @@
 #  pragma system_header
 #endif // no system header
 
+#if _CCCL_TILE_COMPILATION()
+#  include <cassert>
+#endif // !_CCCL_TILE_COMPILATION()
+
 #if !_CCCL_COMPILER(NVRTC)
 #  include <stdlib.h>
 #endif // !_CCCL_COMPILER(NVRTC)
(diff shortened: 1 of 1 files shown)


libcudacxx/include/cuda/std/__exception/terminate.h
```diff
@@ -22,6 +22,10 @@
 #  pragma system_header
 #endif // no system header
 
+#if _CCCL_TILE_COMPILATION()
+#  include <cassert>
+#endif // !_CCCL_TILE_COMPILATION()
```
davebayer, line 27: This won't work for nvrtc. Can't we just include `<cuda/std/cassert>`?

davebayer (changes requested):

libcudacxx/include/cuda/std/__exception/terminate.h
miscco: 🤦 yes we should

libcudacxx/include/cuda/std/__exception/terminate.h
```diff
@@ -22,6 +22,10 @@
 #  pragma system_header
 #endif // no system header
 
+#if _CCCL_TILE_COMPILATION()
+#  include <cassert>
```
davebayer:
```suggestion
#  include <cuda/std/cassert>
```

davebayer (approved):

PR #8626, opened 2026-04-22
[Tile] Avoid uses of `threadIdx` and related CUDA builtins

miscco:
In tile mode we cannot access any of threadIdx or gridDim or ...

We also cannot use compiler builtins such as `__isGlobal` because those are recognized as device functions

### DIFF
--- a/libcudacxx/test/libcudacxx/cuda/functional/proclaim_return_type.pass.cpp
+++ b/libcudacxx/test/libcudacxx/cuda/functional/proclaim_return_type.pass.cpp
@@ -8,8 +8,6 @@
 //
 //===----------------------------------------------------------------------===//
 
-// UNSUPPORTED: gcc-4
-
 #include <cuda/functional>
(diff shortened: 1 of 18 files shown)


libcudacxx/test/libcudacxx/cuda/functional/proclaim_return_type.pass.cpp
```diff
@@ -20,10 +18,16 @@
 template <typename ReturnT>
 TEST_FUNC void test_lambda_return_type()
 {
-  // Ensure type can be queried from cuda::std::invoke_result_t
+// Ensure type can be queried from cuda::std::invoke_result_t
+#  if _CCCL_TILE_COMPILATION()
+  auto d_lm = [] _CCCL_TILE() -> ReturnT {
```
davebayer, line 23: This looks very confusing :D

davebayer (approved):

PR #8635, opened 2026-04-22
Fix cudax hash __load_chunk odd-alignment bug

### LINKED ISSUE #8634: [BUG]: cudax __load_chunk reads wrong bytes for odd-aligned spans in hash functions
### Is this a duplicate?

### Type of Bug

Silent Failure

### Component

CUDA Experimental (cudax)

### Describe the bug

`cuda::experimental::cuco::__load_chunk` reads from the span's base
pointer instead of the indexed pointer in its 1-byte-aligned fallback
branch, so any span-based hash whose start address is odd and whose
length spans more than one chunk silently re-reads the first
`sizeof(_Tp)` bytes for every chunk.

https://github.com/NVIDIA/cccl/blob/09094af138841ef521de1adbbdd18ab8b3dad47b/cudax/include/cuda/experimental/__cuco/__hash_functions/utils.cuh#L40-L66

The three aligned branches copy from `__ptr`; the fallback copies from
`__bytes` (the unindexed base). For any `__index >= 1` the load ignores
the index. Scalar-key paths go through `_Byte_holder` +
`cuda::std::bit_cast` and do not call `__load_chunk`, so they are
unaffected.

All cudax span hash overloads are affected:
`hash<K, xxhash_32>(span)`, `hash<K, xxhash_64>(span)`,
`hash<K, murmurhash3_32>(span)`,
`hash<K, murmurhash3_x86_128>(span)`,
`hash<K, murmurhash3_x64_128>(span)`.

Fix is two characters:

```diff
--- a/cudax/include/cuda/experimental/__cuco/__hash_functions/utils.cuh
+++ b/cudax/include/cuda/experimental/__cuco/__hash_functions/utils.cuh
@@ -60,7 +60,7 @@
   }
   else
   {
-    ::cuda::std::memcpy(&__chunk, __bytes, sizeof(_Tp));
+    ::cuda::std::memcpy(&__chunk, __ptr,   sizeof(_Tp));
   }
```

### How to Reproduce

Host-only; no CUDA compiler needed because `hash<...>::operator()` is
`_CCCL_HOST_DEVICE`.

```cpp
#include <cuda/experimental/__cuco/hash_functions.cuh>
#include <cuda/std/cstddef>
#include <cuda/std/span>
#include <cstdio>

namespace cx = cuda::experimental::cuco;

int main() {
    alignas(8) static cuda::std::byte raw[64];
    for (int i = 0; i < 64; ++i)
        raw[i] = cuda::std::byte{static_cast<unsigned char>(i + 1)};

    constexpr std::size_t len = 40;
    cuda::std::byte*              odd_src = raw + 1;
    alignas(8) static cuda::std::byte aligned_copy[64];
    for (std::size_t i = 0; i < len; ++i) aligned_copy[i] = odd_src[i];

    cx::hash<cuda::std::byte, cx::hash_algorithm::murmurhash3_x64_128> h{0};
    auto h_odd     = h(cuda::std::span<cuda::std::byte>{odd_src,      len});
    auto h_aligned = h(cuda::std::span<cuda::std::byte>{aligned_copy, len});

    std::printf("match? %s\n", h_odd == h_aligned ? "yes" : "NO (bug)");
}
```

### Expected behavior

The same bytes should hash to the same value regardless of source
address.

### Reproduction link

https://godbolt.org/z/vxvb7sffG

### Operating System

_No response_

### nvidia-smi output

_No response_

### NVCC version

_No response_

sleeepyjack:
Fixes #8634

### DIFF
--- a/cudax/include/cuda/experimental/__cuco/__hash_functions/utils.cuh
+++ b/cudax/include/cuda/experimental/__cuco/__hash_functions/utils.cuh
@@ -60,7 +60,7 @@ template <typename _Tp, typename _Extent>
   }
   else
   {
-    ::cuda::std::memcpy(&__chunk, __bytes, sizeof(_Tp));
+    ::cuda::std::memcpy(&__chunk, __ptr, sizeof(_Tp));
   }
   return __chunk;
 }


sleeepyjack:
/ok to test [4e6acdd](https://github.com/NVIDIA/cccl/pull/8635/commits/4e6acddd028cac902b861b176037384a4facc8ab)

davebayer (approved):

PR #8636, opened 2026-04-22
[cudax] Improve `cudax::cuco::__load_chunk`

davebayer:
After I saw #8635, I realized that we can actually make the function better by reusing our alignment machinery.

### DIFF
--- a/cudax/include/cuda/experimental/__cuco/__hash_functions/utils.cuh
+++ b/cudax/include/cuda/experimental/__cuco/__hash_functions/utils.cuh
@@ -21,9 +21,10 @@
 #  pragma system_header
 #endif // no system header
 
+#include <cuda/__memory/is_aligned.h>
+#include <cuda/std/__cstring/memcpy.h>
 #include <cuda/std/__memory/assume_aligned.h>
 #include <cuda/std/cstddef>
-#include <cuda/std/cstring>
 
 #include <cuda/std/__cccl/prologue.h>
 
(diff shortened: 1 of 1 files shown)


cudax/include/cuda/experimental/__cuco/__hash_functions/utils.cuh
```diff
@@ -37,24 +38,22 @@ namespace cuda::experimental::cuco
...
-  auto __ptr     = __bytes + __index * sizeof(_Tp);
-  auto __uintptr = reinterpret_cast<::cuda::std::uintptr_t>(__ptr);
-
   static_assert(sizeof(_Tp) == 4 || sizeof(_Tp) == 8, "__load_chunk must be used with types of size 4 or 8 bytes");
 
-  if (alignof(_Tp) == 8 && ((__uintptr % 8) == 0))
+  const auto __ptr = __bytes + __index * sizeof(_Tp);
+
+  _Tp __chunk;
+  if (alignof(_Tp) == 8 && ::cuda::is_aligned(__ptr, 8))
```
fbusato: what about using `if constexpr`?

fbusato: same for `alignof(_Tp) == 4`

davebayer: We `static_assert` that the type is `4` or `8` bytes wide. I don't know, I didn't want to overcomplicate it :D But yeah, I can add `if constexpr`

fbusato (approved):

PR #8638, opened 2026-04-22
cub: unify the style and hygiene of `policy_selector_from_hub`

### LINKED ISSUE #8494: [BUG]:  error: expression must have a constant value   __launch_bounds__ & cannot call non-constexpr function "__builtin_is_device_code" (declared implicitly)       NV_IF_ELSE_TARGET(NV_IS_HOST,
### Is this a duplicate?

### Type of Bug

Compile-time Error

### Component

CUB

### Describe the bug

```
#include <thrust/copy.h>
#include <thrust/device_vector.h>
#include <thrust/scan.h>

int main()
{
  int keys[]   = {0, 0, 0, 1, 1, 2, 2, 2, 2, 3, 4, 4, 5, 5, 5};
  int values[] = {2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2};

  int N = sizeof(keys) / sizeof(int);

  thrust::device_vector<int> d_keys(keys, keys + N);
  thrust::device_vector<int> d_values(values, values + N);
  thrust::device_vector<int> d_output(N);

  thrust::inclusive_scan_by_key(d_keys.begin(), d_keys.end(), d_values.begin(), d_output.begin());

  return 0;
}

```
This test has started failing after commit 7976e6d with:
```
cub/device/dispatch/dispatch_scan_by_key.cuh", line 128: error: expression must have a constant value
  __launch_bounds__(int(PolicySelector{}(::cuda::arch_id{CUB_PTX_ARCH / 10}).block_threads))
```
and
```
cub/device/dispatch/tuning/tuning_scan_by_key.cuh", line 1069: note: cannot call non-constexpr function "__builtin_is_device_code" (declared implicitly)
      NV_IF_ELSE_TARGET(NV_IS_HOST,
```

### How to Reproduce

```
nvc++  -cuda  -std=c++17 scan_by_key.cu
```

### Expected behavior

This test should compile find after the fix.

### Reproduction link

_No response_

### Operating System

_No response_

### nvidia-smi output

_No response_

### NVCC version

_No response_

griwes:
## Description

Resolves #8494.

I've extended it past just scan_by_key - there's been a few other algorithms that used the nvc++-problematic construction, they should now be in the clear. Also, since the `operator()` of each `policy_selector_from_hub` is now only valid in device code, I've made them all device only, and made `current_policy` also device only, so that attempts to invalidly invoke those will result in reasonably early error messages.

## Checklist


davebayer (dismissed):

bernhardmgruber (dismissed):

griwes:
I am going to widen this to migrate all algorithms from doing `Invoke` within their `policy_selector_from_hub`s, and to use `current_policy` for device side policy getting. That will get us:
1. unified style of hub -> selector conversions across the algorithms, with proper device hygiene where appropriate, and
2. will eliminate the same class of an NVC++ bug from other algorithms.

griwes:
@bernhardmgruber ping for review?

cub/cub/device/dispatch/kernels/kernel_merge_sort.cuh
```diff
@@ -24,11 +24,36 @@ CUB_NAMESPACE_BEGIN
...
+struct device_fallback_policy_getter
+{
+  _CCCL_DEVICE_API _CCCL_FORCEINLINE constexpr auto operator()() const
+  {
+    merge_sort_policy policy = DefaultPolicyGetter{}();
+    policy.block_threads     = 64;
+    policy.items_per_thread  = 1;
+    return policy;
+  }
+};
```
bernhardmgruber: Q: Why can't we just use a single struct definition and use `_CCCL_API` for the `operator()`?

cub/cub/device/dispatch/dispatch_histogram.cuh
bernhardmgruber: Suggestion: I think this type is even unused. Please remove it.

cub/cub/device/dispatch/dispatch_histogram.cuh
```diff
@@ -750,38 +750,16 @@ _CCCL_API constexpr auto convert_policy() -> histogram_policy
...
-                        const int ptx_version = static_cast<int>(arch_id) * 10;
-                        histogram_policy policy{};
-                        extract_policy_dispatch_t dispatch{policy};
-                        MaxPolicy::Invoke(ptx_version, dispatch);
-                        return policy;
-                      }),
-                      ({ return convert_policy<typename MaxPolicy::ActivePolicy>(); }));
+    return convert_policy<typename MaxPolicy::ActivePolicy>();
   }
 };
```
bernhardmgruber: Critical: I don't think this is valid. `policy_selector_from_max_policy` will later be used in host code as well, so `convert_policy<typename MaxPolicy::ActivePolicy>()` is wrong. We need to retain the `extract_policy_dispatch_t`.

griwes: Ah, yes, you're right; I've first done a mechanical replacement of everything, and when things failed, I undid this one but only partially. Will fix.

cub/cub/device/dispatch/kernels/kernel_merge_sort.cuh
griwes: This one we can. I think I was being too careful here, but I THINK that this one is probably fine being a single type with exec check disable.

cub/cub/device/dispatch/dispatch_segmented_sort.cuh
```diff
@@ -521,7 +534,7 @@ struct DispatchSegmentedSort
         return error;
       }
 
-      detail::three_way_partition::dispatch(
+      DispatchThreeWayPartitionIfT::Dispatch(
```
bernhardmgruber, line 537: Important: Is it really necessary to roll back this change? This makes it harder to deprecate `DispatchThreeWayPartitionIf` in a few weeks :S Using `detail::three_way_partition::dispatch` must have worked since it passed all tests and caused no SASS differences.

Why can't we convert the `PartitionPolicyHub` to a policy selector here and call ` detail::three_way_partition::dispatch`?

bernhardmgruber (commented):
There is one issue remaining, the rest LGTM

cub/cub/device/dispatch/dispatch_segmented_sort.cuh
griwes: TWP does the exact problematic for nvc++ thing that scan by key is doing: https://github.com/NVIDIA/cccl/blob/2b7188d2b5e4f2671f68078abc7d842e6b1e3d22/cub/cub/device/dispatch/kernels/kernel_three_way_partition.cuh#L120. So if we preserved the host side of the selector, with target checks, then we will keep the original bug alive within TWP.

So yeah, it worked, but it will break w/ nvc++ as the device compiler.

bernhardmgruber (approved):

cub/cub/device/dispatch/dispatch_segmented_sort.cuh
bernhardmgruber: *sad face*

PR #8639, opened 2026-04-22
Segmented scan multisegment part 1

### LINKED ISSUE #6783: [CUB]: Support processing multiple segments per block in agent_segment_scan

### LINKED ISSUE #8424: Implement the new tuning API for `detail::segmented_scan::dispatch_segmented_scan`

oleksandr-pavlyk:
## Description

This PR builds on top of #6712 and adds two commits
  - First one removes support for thread-granularity agent and kernel
  - Second removes support for warp-granularity agent and kernel

This should reduce the number of lines changed significantly.

---

Closes #6783 
Closes #8424

  - extend `agent_segmented_scan` with ability to to consume multiple segments per block
     - Introduce new agent template parameter `MaxSegmentsPerBlock` which controls the size of `OffsetT` array in shared memory dedicated to storing inclusive-scan of segment sizes. The actual number of segments to process is passed dynamically to `consume_ranges` method. It can not exceed `MaxSegmentsPerBlock`, however.
     - Multiple-segments are processed by reconstructing head-flags while loading values using `multi_segment_iterator` class, which is read- and write- transform iterator.  
        - Read transformation reconstructs head-flag and forms value-flag pair, used as `AccumT` by `BlockScan`, and  
        - write-transform unpacks value-flag pair and only writes the value to the output iterator.
     - Mapping from `element_id` to `segment_id` and `relative_offset` (or rank) is encapsulated into searcher types.
     - Three searcher types are implemented:
        - When segment sizes are all the same, `segment_id = element_id / segment_size`.
        - Generically, binary search in the array of inclusive-scan of segment sizes is performed. Two implementations (types) are provided:
           - When static bound on array size of known (as it is in our case), branch-less binary search provides best performance
           - When this bound is unknown, `cuda::std::upper_bound` can be used for larger number of sizes, and linear search using dynamic for-loop for  smaller sizes
     - While computing inclusive-scan of segment sizes, we check if all sizes to be processed by block are the same, and use fixed-size searcher, otherwise use branch-less generic searcher.
     - When dynamic number of segments equals 1, simpler `consume_range` call is made, which does not reconstruct head-flags 

   - Dispatch layer is modified to take 2 additional parameters (with default values), the number of segments to be processed by a worker (defaults to 1), and the type of worker requested (defaults to block).


cub/cub/device/dispatch/tuning/tuning_segmented_scan.cuh
```diff
@@ -13,95 +13,155 @@
...
+      large_values ? BLOCK_LOAD_WARP_TRANSPOSE_TIMESLICED : BLOCK_LOAD_WARP_TRANSPOSE;
+    const auto scan_transposed_blockstore =
+      large_values ? BLOCK_STORE_WARP_TRANSPOSE_TIMESLICED : BLOCK_STORE_WARP_TRANSPOSE;
+
+    constexpr int nominal_block_threads    = 128;
+    constexpr int nominal_items_per_thread = 9;
+    constexpr int max_segments_per_block   = 512;
+
+    const int align          = accum_size;
+    const int augmented_size = ::cuda::round_up(accum_size, align);
```
oleksandr-pavlyk: In code as in, `augmented_size` is always equal to `accum_size`. The policy selector should either also store `int{alignof(AccumT)}`, or we should simplify the code.

cub/cub/device/dispatch/dispatch_segmented_scan.cuh
```diff
@@ -66,213 +73,194 @@ struct device_segmented_scan_kernel_source
...
   static_assert(::cuda::std::is_integral_v<OffsetT> && sizeof(OffsetT) >= 4 && sizeof(OffsetT) <= 8,
                 "dispatch_segmented_scan only supports integral offset types of 4- or 8-bytes");
 
-  /// Device-accessible allocation of temporary storage.  When nullptr, the
-  /// required allocation size is written to \p temp_storage_bytes and no work
-  /// is done.
-  void* d_temp_storage;
+  if (num_segments <= 0)
+  {
+    return cudaErrorUnknown;
```
oleksandr-pavlyk:
```suggestion
    return (num_segments == 0) ? cudaSuccess : cudaErrorUnknown;
```

The error should be signaled only for strictly negative value.

cub/cub/device/dispatch/dispatch_segmented_scan.cuh
```diff
@@ -66,213 +73,194 @@ struct device_segmented_scan_kernel_source
...
-  /// Offsets to end of each segment in the input sequence
-  EndOffsetIteratorInputT d_input_end_offsets;
+  const auto [workers_per_block, block_size, normalized_spw] =
+    [&](worker selector) -> ::cuda::std::tuple<int, int, int> {
+    switch (selector)
+    {
+      case worker::block: {
+        const auto bw = active_policy.block;
+        _CCCL_ASSERT(num_segments_per_worker <= bw.max_segments, "Number of segments per block exceeds maximum value");
+        return {bw.WorkersPerBlock(), bw.BlockThreads(), ::cuda::std::min(num_segments_per_worker, bw.max_segments)};
```
oleksandr-pavlyk: Should we account for the possibility that `bw.max_segments` may be negative? Perhaps the line `num_segments_per_worker = (::cuda::std::max) (num_segments_per_worker, 1);` should be moved after and applied `normalized_spw` instead

cub/cub/device/device_segmented_scan.cuh
```diff
@@ -2276,26 +2197,20 @@ struct DeviceSegmentedScan
...
-                                     d_in,
-                                     d_out,
-                                     num_segments,
-                                     d_in_begin_offsets,
-                                     d_in_end_offsets,
-                                     d_in_begin_offsets,
-                                     scan_op,
-                                     detail::InputValue<InitValueT>(init_value),
-                                     stream);
+    return cub::detail::segmented_scan::dispatch<ForceInclusive::Yes, accum_t>(
```
bernhardmgruber: I think the overriding accumulator type provided here is actually the one that `dispatch` would deduce anyway. I think it can be omitted. Calling `cub::detail::segmented_scan::dispatch<ForceInclusive::Yes>` should be fine.

Please double check, maybe I am missing something. Please also check that for the remaining overloads.

cub/test/catch2_test_device_segmented_scan.cu
```diff
@@ -246,10 +246,12 @@ C2H_TEST("Device segmented_scan works with all device interfaces", "[segmented][
...
   {
-    using op_t    = cuda::std::plus<>;
-    using accum_t = cuda::std::__accumulator_t<op_t, input_t, input_t>;
+    using op_t              = cuda::std::plus<>;
+    using unwrapped_input_t = typename cuda::std::iterator_traits<decltype(unwrap_it(d_in_it))>::value_type;
+    using accum_t           = cuda::std::__accumulator_t<op_t, unwrapped_input_t, unwrapped_input_t>;
+    using h_accum_t         = cuda::std::__accumulator_t<op_t, input_t, input_t>;
 
-    INFO("Accum type: " << typeid(accum_t).name());
+    INFO("Unwrapped accum type: " << typeid(accum_t).name() << ", host accum_t: " << typeid(h_accum_t).name());
```
bernhardmgruber: May be easier:
```suggestion
    CAPTURE(c2h::type_name<accum_t>(), c2h::type_name<h_accum_t>());
```

cub/test/catch2_test_device_segmented_scan_multi_segment.cu
```diff
@@ -0,0 +1,693 @@
...
+  return cuda::ceil_div(max_input_bytes, elem_bytes);
+}
+
+using integral_types = c2h::type_list<std::int32_t, std::int64_t, std::uint32_t, std::uint64_t>;
+
+using itp_list =
+  c2h::type_list<std::integral_constant<int, 1>,
+                 std::integral_constant<int, 2>,
+                 std::integral_constant<int, 3>,
+                 std::integral_constant<int, 8>>;
```
bernhardmgruber: Please prefer utilities from `cuda::std::` to increase our test coverage.

cub/test/catch2_test_device_segmented_scan_multi_segment.cu
```diff
@@ -0,0 +1,693 @@
...
+    d_offsets,
+    d_offsets + 1,
+    d_out_offsets,
+    scan_op,
+    init_value,
+    segments_per_worker,
+    worker_choice,
+    stream);
+
+  c2h::device_vector<cuda::std::uint8_t> temp_storage(temp_storage_bytes);
```
bernhardmgruber:
```suggestion
  c2h::device_vector<cuda::std::uint8_t> temp_storage(temp_storage_bytes, thrust::no_init);
```

cub/cub/device/dispatch/dispatch_segmented_scan.cuh
```diff
@@ -66,213 +73,194 @@ struct device_segmented_scan_kernel_source
...
-
-  CUB_RUNTIME_FUNCTION static constexpr size_t AccumSize()
-  {
-    return sizeof(AccumT);
-  }
+      EnforceInclusive == ForceInclusive::Yes>);
 };
 
+// select the accumulator type using an overload set, so __accumulator_t is not instantiated when an overriding
+// accumulator type is present. This matches the reduce dispatch pattern and is needed by CCCL.C.
```
bernhardmgruber: Suggestion: that workaround is actually no needed, since CCCL.C does not expose segmented scan. And if it will, we may already have host code jitting.

cub/cub/device/dispatch/dispatch_segmented_scan.cuh
```diff
@@ -66,213 +73,194 @@ struct device_segmented_scan_kernel_source
...
+  EndOffsetIteratorInputT input_end_offsets,
+  BeginOffsetIteratorOutputT output_begin_offsets,
+  ScanOpT scan_op,
+  InitValueT init_value,
+  int num_segments_per_worker,
+  worker worker_choice,
+  cudaStream_t stream,
+  PolicySelector policy_selector         = {},
+  KernelSource kernel_source             = {},
+  KernelLauncherFactory launcher_factory = {})
```
bernhardmgruber, line 130: Remark: Also those would not be necessary since there is no CCCL.C exposure. But you can leave them.

cub/cub/device/dispatch/dispatch_segmented_scan.cuh
bernhardmgruber: We should at least put an assertion in. You can also detect the situation and error if you want.

cub/cub/device/dispatch/dispatch_segmented_scan.cuh
```diff
@@ -66,213 +73,194 @@ struct device_segmented_scan_kernel_source
...
+          kernel_source.segmented_scan_kernel(),
+          d_in,
+          d_out,
+          input_begin_offsets,
+          input_end_offsets,
+          output_begin_offsets,
+          segment_count,
+          scan_op,
+          init_value,
+          normalized_spw);
```
bernhardmgruber:
```suggestion
        error = CubDebug(launcher.doit(
          kernel_source.segmented_scan_kernel(),
          d_in,
          d_out,
          input_begin_offsets,
          input_end_offsets,
          output_begin_offsets,
          segment_count,
          scan_op,
          init_value,
          normalized_spw));
```

bernhardmgruber: I would also generally follow the pattern of
```c++
if (const auto error = CubDebug(func()))
  return error;
```
It leaves less room for mistakes like forgetting to check an error code, or checking the error code from a wrong API call etc.

Should be applied a few times below.

cub/cub/device/dispatch/dispatch_segmented_scan.cuh
```diff
@@ -66,213 +73,194 @@ struct device_segmented_scan_kernel_source
...
-  OutputIteratorT d_out;
+  if (d_temp_storage == nullptr)
+  {
+    temp_storage_bytes = 1;
+    return cudaSuccess;
+  }
 
-  /// The number of segments that comprise the segmented scan data
-  ::cuda::std::int64_t num_segments;
+  active_policy.CheckLoadModifier();
```
bernhardmgruber: Suggestion: can we inline the content of `CheckLoadModifier();` and avoid member functions on policy structs? I want them to be as simple as possible since they will become part of the public API.

cub/cub/device/dispatch/tuning/tuning_segmented_scan.cuh
```diff
@@ -13,95 +13,155 @@
...
+
+  CUB_RUNTIME_FUNCTION constexpr int BlockThreads() const
+  {
+    return block_threads;
+  }
+
+  CUB_RUNTIME_FUNCTION constexpr int WorkersPerBlock() const
+  {
+    return 1;
+  }
```
bernhardmgruber: Suggestion: can we make those free functions instead? or just inline their behavior. The goal is to have no member functions on policy structs, unless there is a really strong reason for them.

cub/cub/device/dispatch/tuning/tuning_segmented_scan.cuh
bernhardmgruber: Feel free to store additional values!

cub/cub/device/dispatch/tuning/tuning_segmented_scan.cuh
```diff
@@ -13,95 +13,155 @@
...
-      AccumT,
-      scan_transposed_load,
+    const bool large_values = accum_size > 128;
+    const auto scan_transposed_blockload =
+      large_values ? BLOCK_LOAD_WARP_TRANSPOSE_TIMESLICED : BLOCK_LOAD_WARP_TRANSPOSE;
+    const auto scan_transposed_blockstore =
+      large_values ? BLOCK_STORE_WARP_TRANSPOSE_TIMESLICED : BLOCK_STORE_WARP_TRANSPOSE;
+
+    constexpr int nominal_block_threads    = 128;
+    constexpr int nominal_items_per_thread = 9;
```
bernhardmgruber, line 111: Question: This value used to be 15 in the original code. Is this change expected?

cub/cub/device/dispatch/kernels/kernel_segmented_scan.cuh
```diff
@@ -42,35 +619,75 @@ __launch_bounds__(int(ChainedPolicyT::ActivePolicy::segmented_scan_policy_t::BLO
...
+    policy.items_per_thread,
+    policy.load_algorithm,
+    policy.load_modifier,
+    policy.store_algorithm,
+    policy.scan_algorithm,
+    policy.max_segments>;
 
-  __shared__ typename agent_segmented_scan_t::TempStorage temp_storage;
+  using agent_t =
+    agent_segmented_scan<policy_t, InputIteratorT, OutputIteratorT, OffsetT, ScanOpT, ActualInitValueT, AccumT, ForceInclusive>;
```
bernhardmgruber: Suggestion: the workaround I sometimes use here is to just wrap the tuning policy in a nullary callable and pass its type to the agent:

```suggestion
  struct policy_getter() {
    constexpr auto operator()() {
      return policy;
    }
  };

  using agent_t =
    agent_segmented_scan<policy_getter, InputIteratorT, OutputIteratorT, OffsetT, ScanOpT, ActualInitValueT, AccumT, ForceInclusive>;
```
The agent can then access the policy values like `PolicyGetter().block.value`.

bernhardmgruber (commented):
Here is some high level review

cub/cub/device/dispatch/tuning/tuning_segmented_scan.cuh
oleksandr-pavlyk: My local benchmarking kept pointing me to 9 as being an optimal value. 

This should only matter for very short types due to scaling.  I assuming tuning will follow, so I would prefer to keep this value until we find evidence to change it.

cub/cub/device/dispatch/kernels/kernel_segmented_scan.cuh
oleksandr-pavlyk: This way we can remove agent policy structs entirely. I suppose agent should constrain the `policy_getter` type with a concept check.

cub/cub/device/dispatch/dispatch_segmented_scan.cuh
oleksandr-pavlyk: I understand you are suggesting to remove `OverrideAccumT` template parameter entirely, remove `select_accum_t` and always use `__accumulator_t`?

cub/cub/device/dispatch/kernels/kernel_segmented_scan.cuh
bernhardmgruber: > we can remove agent policy structs entirely

That's also a goal. However, we can't pass values returned from the policy selector through template parameters in C++17. We need to wait until 20.

cub/cub/device/dispatch/tuning/tuning_segmented_scan.cuh
bernhardmgruber: Alright, so we are consciously changing the tunings here, thx!

cub/benchmarks/bench/segmented_scan/base.cuh
```diff
@@ -0,0 +1,228 @@
...
+  using offset_t       = cub::detail::choose_offset_t<OffsetT>;
+  using offset_it      = const offset_t*;
+
+#if !TUNE_BASE
+  // IMPORTANT: Make sure not to hurt occupancy
+  // since shared memory is allocated per warp, scale it down so that total amount of shared memory
+  // per CTA is the same as in block-level segmented scan
+  constexpr auto max_segments_per_warp =
+    ::cuda::ceil_div(TUNE_MAX_SEGMENTS_PER_BLOCK * cub::detail::warp_threads, TUNE_THREADS);
+  using policy_t = policy_selector_t<TUNE_THREADS, TUNE_ITEMS, TUNE_MAX_SEGMENTS_PER_BLOCK, max_segments_per_warp>;
```
oleksandr-pavlyk:
```suggestion
  using policy_t = policy_selector_t<TUNE_THREADS, TUNE_ITEMS, TUNE_MAX_SEGMENTS_PER_BLOCK>;
```

cub/benchmarks/bench/segmented_scan/base.cuh
```diff
@@ -0,0 +1,228 @@
...
+#  endif // TUNE_TRANSPOSE
+
+#  if TUNE_LOAD == 0
+#    define TUNE_LOAD_MODIFIER cub::LOAD_DEFAULT
+#  elif TUNE_LOAD == 1
+#    define TUNE_LOAD_MODIFIER cub::LOAD_CA
+#  endif // TUNE_LOAD
+
+template <int BlockThreads, int ItemsPerThread, int MaxSegmentsPerBlock, int MaxSegmentsPerWarp>
+struct policy_selector_t
```
oleksandr-pavlyk:
```suggestion
template <int BlockThreads, int ItemsPerThread, int MaxSegmentsPerBlock>
struct policy_selector_t
```

cub/cub/device/dispatch/dispatch_segmented_scan.cuh
```diff
@@ -66,213 +73,184 @@ struct device_segmented_scan_kernel_source
...
+  _CCCL_ASSERT(num_segments_per_worker > 0, "Number of segments per worker parameter must be positive");
 
-  /// Offsets to beginning of each segment in the output sequence
-  BeginOffsetIteratorOutputT d_output_begin_offsets;
+  const auto [workers_per_block, block_size, normalized_spw] =
+    [&](worker selector) -> ::cuda::std::tuple<int, int, int> {
+    switch (selector)
+    {
+      case worker::block: {
+        const auto bw = active_policy.block;
```
NaderAlAwar: Nit: bw is hard to read. Something like `block_policy` would be better IMO

cub/cub/device/dispatch/dispatch_segmented_scan.cuh
```diff
@@ -66,213 +73,184 @@ struct device_segmented_scan_kernel_source
...
+  InputIteratorT d_in,
+  OutputIteratorT d_out,
+  ::cuda::std::int64_t num_segments,
+  BeginOffsetIteratorInputT input_begin_offsets,
+  EndOffsetIteratorInputT input_end_offsets,
+  BeginOffsetIteratorOutputT output_begin_offsets,
+  ScanOpT scan_op,
+  InitValueT init_value,
+  int num_segments_per_worker,
+  worker worker_choice,
```
NaderAlAwar, line 126: Question: why are these passed as arguments here? I would have thought they're internal tuning parameters

cub/cub/device/dispatch/kernels/kernel_segmented_scan.cuh
```diff
@@ -42,35 +618,81 @@ __launch_bounds__(int(ChainedPolicyT::ActivePolicy::segmented_scan_policy_t::BLO
...
-    ActualInitValueT,
-    AccumT,
-    ForceInclusive>;
+  struct policy_getter
+  {
+    constexpr auto operator()() const
+    {
+      return policy;
+    }
+  };
```
NaderAlAwar, line 634: Question: can you use `cub::detail::device_policy_getter`?

cub/cub/device/dispatch/tuning/tuning_segmented_scan.cuh
```diff
@@ -13,95 +13,148 @@
...
+    return !(lhs == rhs);
+  }
+
+#if !_CCCL_COMPILER(NVRTC)
+  friend ::std::ostream& operator<<(::std::ostream& os, const block_segmented_scan_policy& policy)
+  {
+    return os
+        << "block_segmented_scan_policy { .block_threads = " << policy.block_threads
+        << ", .items_per_thread = " << policy.items_per_thread << ", .load_algorithm = "
+        << ", .load_modifier = " << policy.load_modifier << policy.load_algorithm
```
NaderAlAwar: Important: the printing of load algorithm and modifier is currently malformed

cub/cub/device/dispatch/tuning/tuning_segmented_scan.cuh
oleksandr-pavlyk: @bernhardmgruber SASS of `cub.bench.segmented_scan.custom.base` shows LDL/STL instruction only for `int128_t`, and the same for `cub.bench.segmented_scan.sum.base` (checked manually). 

There are 34 local load/stores per kernel instance for int128_t for each offset type:

```
(cccl) coder ➜ ~/cccl $ grep LDL segmented_scan_sum_base.sass | wc
     68     544    8024
(cccl) coder ➜ ~/cccl $ grep STL segmented_scan_sum_base.sass | wc
     68     544    8024
```

cub/cub/device/dispatch/kernels/kernel_segmented_scan.cuh
```diff
@@ -13,15 +17,584 @@
...
+      worker_prefix_callback_t prefix_callback_op{exclusive_prefix, offsets_scan_op};
+
+      for (unsigned chunk_id = 0; chunk_id < n_chunks; ++chunk_id)
+      {
+        const unsigned work_id = chunk_id * block_threads + tid;
+
+        // TODO: use BlockLoad to load
+        const OffsetT input_segment_begin = (work_id < n_segments) ? inp_idx_begin_it[work_id] : 0;
+        const OffsetT input_segment_end   = (work_id < n_segments) ? inp_idx_end_it[work_id] : 0;
+        const OffsetT segment_size        = input_segment_end - input_segment_begin;
```
NaderAlAwar: Critical: for cases where `input_segment_end < input_segment_begin`, this multi-segment path no longer matches the single-segment behavior. The single-segment path clamps inverted ranges to empty via `max(end, begin) - begin`, but here we subtract directly

cub/cub/device/dispatch/kernels/kernel_segmented_scan.cuh
```diff
@@ -13,15 +17,584 @@
...
+  _CCCL_DEVICE _CCCL_FORCEINLINE void consume_ranges_chunked_impl(
+    const SearcherT& searcher,
+    InputBeginOffsetIteratorT inp_idx_begin_it,
+    OutputBeginOffsetIteratorT out_idx_begin_it,
+    OffsetT items_per_block)
+  {
+    const OffsetT n_chunks = ::cuda::ceil_div(items_per_block, tile_items);
+
+    using augmented_scan_op_t = multi_segment_helpers::schwarz_scan_op<ScanOpT, AccumT>;
+    using augmented_init_value_t =
```
NaderAlAwar: Suggestion: this is unused

cub/cub/device/dispatch/tuning/tuning_segmented_scan.cuh
```diff
@@ -13,95 +13,148 @@
...
-    using segmented_scan_policy_t = agent_segmented_scan_policy_t<
-      128,
-      15,
-      AccumT,
-      scan_transposed_load,
+    const bool large_values = accum_size > 128;
+    const auto scan_transposed_blockload =
+      large_values ? BLOCK_LOAD_WARP_TRANSPOSE_TIMESLICED : BLOCK_LOAD_WARP_TRANSPOSE;
+    const auto scan_transposed_blockstore =
+      large_values ? BLOCK_STORE_WARP_TRANSPOSE_TIMESLICED : BLOCK_STORE_WARP_TRANSPOSE;
```
NaderAlAwar: Question: would it make more sense to determine large_values after you calculate `augmented_size` below?

cub/cub/device/dispatch/kernels/kernel_segmented_scan.cuh
oleksandr-pavlyk: We could, but I think we have shaving a bit of compile time here. With `cub::detail::policy_getter` or `cub::detail::device_policy_getter` we would need to reevaluate `policy_selector::operator()` call. 

Here we need that value to extract agent parameters. I don't have a strong opinion here.

cub/cub/device/dispatch/kernels/kernel_segmented_scan.cuh
oleksandr-pavlyk: <strike>Thanks for spotting this. I modified lambda function to declare return type as `augmented_init_value_t`.</strike>

**Update**: actually removed it as suggested.

cub/cub/device/dispatch/dispatch_segmented_scan.cuh
oleksandr-pavlyk: Great question. 

In #6712, the choice of `worker::block` is most appropriate when segments are sufficiently large; as segment sizes become smaller, it becomes beneficial to use `worker::warp` or `worker::thread`. There is no automation to decide on which worker to choose from the algorithm's input data. The same applies to the number of segments processed by the chosen worker.

So I made it an argument to dispatch to test and benchmark choice of worker and segments per worker.

I think this needs to be revisited, but doing so is not in the scope of this PR. Also see https://github.com/NVIDIA/cccl/pull/6712#issuecomment-4215577790

cub/cub/device/dispatch/kernels/kernel_warp_segmented_scan.cuh
```diff
@@ -0,0 +1,667 @@
...
+          typename ScanOpT,
+          typename InitValueT,
+          typename AccumT,
+          bool ForceInclusive,
+          typename ActualInitValueT = typename InitValueT::value_type>
+#if _CCCL_HAS_CONCEPTS()
+  requires segmented_scan_policy_selector<PolicySelector>
+#endif // _CCCL_HAS_CONCEPTS()
+__launch_bounds__(int(PolicySelector{}(::cuda::arch_id{CUB_PTX_ARCH / 10}).warp.block_threads))
+  _CCCL_KERNEL_ATTRIBUTES void device_warp_segmented_scan_kernel(
```
NaderAlAwar: Question: This is not called anywhere in this PR currently. Do you intend to keep it?

cub/cub/detail/segmented_scan_helpers.cuh
```diff
@@ -0,0 +1,492 @@
...
+  IterT m_it;
+  OffsetT m_start;
+  SearcherT m_searcher;
+  BeginOffsetIterT m_it_idx_begin;
+  ReadTransformT m_read_transform_fn;
+  WriteTransformT m_write_transform_fn;
+
+  using iterator_concept      = ::cuda::std::random_access_iterator_tag;
+  using iterator_category     = ::cuda::std::random_access_iterator_tag;
+  using underlying_value_type = ::cuda::std::iter_value_t<IterT>;
```
NaderAlAwar, line 385: Suggestion: this may overly constrain the output iterator type. This adapter is currently used for both inputs and outputs, but the store path only needs to write through the output iterator. Discard or transform output iterators could work theoretically, but might be rejected here because the adapter requires a readable `iter_value_t<IterT>`. Is there a way to relax this, maybe by splitting the input/output adapters?

cub/test/catch2_test_device_segmented_scan_multi_segment.cu
NaderAlAwar, line 1: Suggestion: could we add a test where `end_offset < begin_offset` for one segment, if that is supported? From what I can see it should be treated as an empty segment

cub/test/catch2_test_device_segmented_scan_multi_segment.cu
```diff
@@ -0,0 +1,689 @@
...
+  const unsigned num_items  = offsets.back();
+
+  const auto num_output = out_offsets.back();
+
+  c2h::device_vector<value_t> input(num_items);
+  c2h::device_vector<value_t> output(num_output);
+
+  thrust::tabulate(input.begin(), input.end(), cuda::std::identity{});
+
+  constexpr int segments_per_worker = 2;
```
NaderAlAwar, line 633: Suggestion: could we use randomly generated values, up to a value where `segments_per_worker > block_threads`? This will test the path where the loop in `consume_ranges` processes segment offsets in multiple chunks and has to carry the prefix across iterations.

cub/cub/device/dispatch/kernels/kernel_segmented_scan.cuh
```diff
@@ -42,35 +616,81 @@ __launch_bounds__(int(ChainedPolicyT::ActivePolicy::segmented_scan_policy_t::BLO
...
+  _CCCL_ASSERT(num_segments_per_worker <= policy.block.max_segments,
+               "Requested number of segments to be processed by block exceeds compile-time maximum");
+
+  const auto work_id = num_segments_per_worker * blockIdx.x;
+
+  agent_t agent(temp_storage, d_in, d_out, scan_op, _init_value);
+
+  if constexpr (policy.block.max_segments == 1)
+  {
+    _CCCL_ASSERT(num_segments_per_worker == 1, "Inconsistent parameters in device_warp_segmented_scan_kernel");
```
NaderAlAwar: Nit:
```suggestion
    _CCCL_ASSERT(num_segments_per_worker == 1, "Inconsistent parameters in device_segmented_scan_kernel");
```

NaderAlAwar (approved):

cub/cub/device/dispatch/kernels/kernel_warp_segmented_scan.cuh
oleksandr-pavlyk: This file was meant to be removed in the commit. I will amend and force push.

cub/cub/detail/segmented_scan_helpers.cuh
oleksandr-pavlyk: Great point, @NaderAlAwar !

I will start by adding tests that use `transform_iterator` for `d_input`, and `transform_output_iterator` for `d_output` and fix this by introducing `multi_segmented_input_iterator` and multi_segmented_output_iterator`. 

Right now, I am forced to pass some unused write-transform projection functor instance for multi-segmented iterator meant for reading input, and some unused read-transform packer functor for segmented iterator meant for writing output.

The change would make the intent clearer even though it would increase number of lines added.

cub/test/catch2_test_device_segmented_scan_multi_segment.cu
oleksandr-pavlyk: To carry the prefix across iterations the total number of elements in `segments_per_worker` needs to exceed `threads_in_worker * items_per_thread`.  Also `segments_per_worker` should not exceed `max_segments` values in the policy.

That said, using `GENERATE` is a good idea.

cub/cub/detail/segmented_scan_helpers.cuh
```diff
@@ -0,0 +1,492 @@
...
+#include <cuda/std/type_traits>
+
+CUB_NAMESPACE_BEGIN
+
+namespace detail::segmented_scan
+{
+namespace multi_segment_helpers
+{
+template <typename ValueT, typename FlagT = bool>
+using augmented_value_t = ::cuda::std::tuple<ValueT, FlagT>;
```
bernhardmgruber: Suggestion: could this just be a struct? `tuple` is relatively expensive to compile.

cub/cub/detail/segmented_scan_helpers.cuh
```diff
@@ -0,0 +1,492 @@
...
+_CCCL_DEVICE _CCCL_FORCEINLINE constexpr ValueT get_value(augmented_value_t<ValueT, FlagT> fv) noexcept
+{
+  return ::cuda::std::get<0>(fv);
+}
+
+template <typename ValueT, typename FlagT>
+_CCCL_DEVICE _CCCL_FORCEINLINE constexpr augmented_value_t<ValueT, FlagT> make_value_flag(ValueT v, FlagT f) noexcept
+{
+  return {v, f};
+}
```
bernhardmgruber, line 56: Remark: These functions kind of suggest that `augmented_value_t` should just be a struct with a `flag` and `value` data member.

cub/cub/detail/segmented_scan_helpers.cuh
```diff
@@ -0,0 +1,492 @@
...
+  }
+
+  template <typename Op                   = BinaryOpT,
+            ::cuda::std::enable_if_t<!::cuda::std::is_invocable_v<const Op&, const ValueT&, ValueT>
+                                       && ::cuda::std::is_invocable_v<Op&, const ValueT&, ValueT>,
+                                     int> = 0>
+  _CCCL_DEVICE _CCCL_FORCEINLINE fv_t operator()(fv_t o1, fv_t o2)
+  {
+    return call_impl(scan_op, o1, o2);
+  }
```
bernhardmgruber: Suggestion: Let's just mark the `scan_op` mutable and avoid the overloads with complex constraints:


```suggestion
  mutable BinaryOpT scan_op;
  
  _CCCL_DEVICE _CCCL_FORCEINLINE fv_t operator()(fv_t o1, fv_t o2) const
  {
    if (get_flag(o2))
    {
      return o2;
    }
    const auto o2_value    = get_value(o2);
    const auto o1_value    = get_value(o1);
    const ValueT res_value = scan_op(o1_value, o2_value);

    return make_value_flag(res_value, get_flag(o1));
  }
```

cub/cub/detail/segmented_scan_helpers.cuh
```diff
@@ -0,0 +1,492 @@
...
+
+  template <typename Op = ScanOp,
+            ::cuda::std::enable_if_t<
+              !::cuda::std::is_invocable_v<const Op&, const V&, V> && ::cuda::std::is_invocable_v<Op&, const V&, V>,
+              int> = 0>
+  _CCCL_HOST_DEVICE _CCCL_FORCEINLINE constexpr auto operator()(V v, F f)
+  {
+    return impl(op, v, f);
+  }
```
bernhardmgruber: Same here

cub/cub/detail/segmented_scan_helpers.cuh
```diff
@@ -0,0 +1,492 @@
...
+    WriteTransformT m_write_fn;
+
+    _CCCL_DEVICE _CCCL_FORCEINLINE explicit __mapping_proxy(
+      IterT it, OffsetT offset, bool head_flag, ReadTransformT read_fn, WriteTransformT write_fn)
+        : m_it(it)
+        , m_offset(offset)
+        , m_head_flag(head_flag)
+        , m_read_fn(read_fn)
+        , m_write_fn(write_fn)
+    {}
```
bernhardmgruber: Suggestion: Please delete the constructor and just aggregate init.

cub/cub/detail/segmented_scan_helpers.cuh
```diff
@@ -0,0 +1,492 @@
...
+      return m_read_fn(m_it[m_offset], m_head_flag);
+    }
+
+    template <
+      typename Op = ReadTransformT,
+      cuda::std::enable_if_t<!::cuda::std::is_invocable_v<const Op&, const underlying_value_type&, bool>, int> = 0>
+    _CCCL_DEVICE _CCCL_FORCEINLINE operator value_type()
+    {
+      return m_read_fn(m_it[m_offset], m_head_flag);
+    }
```
bernhardmgruber: Again: just one overload and mark `m_read_fn` as`mutable`.

Why do we even need to handle non-`const` `operator()`? Is this for backward compatibility? IMO, a non-`const` operator is a user-side bug.

cub/cub/device/dispatch/kernels/kernel_segmented_scan.cuh
```diff
@@ -13,15 +17,582 @@
...
+  // Use cub::NullType means no initial value is provided
+  static constexpr bool has_init = !::cuda::std::is_same_v<InitValueT, NullType>;
+  // We are relying on either initial value being `NullType`
+  // or the ForceInclusive tag to be true for inclusive scan
+  // to get picked up.
+  static constexpr bool is_inclusive    = ForceInclusive || !has_init;
+  static constexpr int block_threads    = SegmentedScanPolicyGetterT{}().block.block_threads;
+  static constexpr int items_per_thread = SegmentedScanPolicyGetterT{}().block.items_per_thread;
+  static constexpr int tile_items       = block_threads * items_per_thread;
+  static constexpr int max_segments     = SegmentedScanPolicyGetterT{}().block.max_segments;
```
bernhardmgruber: Suggestion: I would extract the policy into a separate constant for slightly better readability:
```c++
static constexpr /* type */ policy = SegmentedScanPolicyGetterT{}();
```

cub/cub/detail/segmented_scan_helpers.cuh
oleksandr-pavlyk: I implemented your suggested to use `mutable` qualifier.

cub/cub/detail/segmented_scan_helpers.cuh
oleksandr-pavlyk: I started by adding a test that use `transform_iterator` for input, and `transform_output_iterator` for output. It failed to compile for the reason you stated. 

I split the multi-segmented iterator into `multi_segmented_input_iterator` and `multi_segmented_output_iterator` which made the test pass. 

Thank you for the suggestion, @NaderAlAwar !

cub/test/catch2_test_device_segmented_scan_multi_segment.cu
oleksandr-pavlyk: I left this unchanged for now. There are other tests that vary the number of segments per worker.

cub/test/catch2_test_device_segmented_scan_multi_segment.cu
oleksandr-pavlyk: Test was added

cub/cub/detail/segmented_scan_helpers.cuh
```diff
@@ -0,0 +1,543 @@
...
+_CCCL_DEVICE _CCCL_FORCEINLINE constexpr ValueT get_value(augmented_value_t<ValueT, FlagT> fv) noexcept
+{
+  return fv.value;
+}
+
+template <typename ValueT, typename FlagT>
+_CCCL_DEVICE _CCCL_FORCEINLINE constexpr augmented_value_t<ValueT, FlagT> make_value_flag(ValueT v, FlagT f) noexcept
+{
+  return {v, f};
+}
```
bernhardmgruber, line 56: Suggestion: I think those functions are obsolete. Let's just access the data members and ctor directly.

For template argument deduction to work in C++17, you could add a deduction guide:
```c++
template <typename ValueT, typename FlagT = bool>
augmented_value_t(ValueT, FlagT) -> augmented_value_t<ValueT, FlagT>;
```

cub/cub/detail/segmented_scan_helpers.cuh
```diff
@@ -0,0 +1,543 @@
...
+    return {segment_id, relative_offset};
+  }
+};
+
+template <typename IterT,
+          typename OffsetT,
+          typename SearcherT,
+          typename BeginOffsetIterT,
+          typename ReadTransformT,
+          unsigned int LinearBinarySearchThreshold = 18>
```
bernhardmgruber, line 374: Important: This parameter is unused. Should it be deleted?

cub/cub/detail/segmented_scan_helpers.cuh
```diff
@@ -0,0 +1,543 @@
...
+  {
+    V res = v;
+    if (f)
+    {
+      res = op_ref(init_v, v);
+    }
+    return make_value_flag(res, f);
+  }
+
+public:
```
bernhardmgruber, line 132: Important: Seems unused

```suggestion

```

cub/cub/detail/segmented_scan_helpers.cuh
```diff
@@ -0,0 +1,543 @@
...
+struct multi_segmented_input_iterator
+{
+  IterT m_it;
+  OffsetT m_start;
+  SearcherT m_searcher;
+  BeginOffsetIterT m_it_idx_begin;
+  mutable ReadTransformT m_read_transform_fn;
+
+  using iterator_concept      = ::cuda::std::random_access_iterator_tag;
+  using iterator_category     = ::cuda::std::random_access_iterator_tag;
```
bernhardmgruber, line 384: Remark: `multi_segmented_input_iterator` is technically an input iterator, because it does not return `value_type&` for `operator*`. But that may degrade the iterator's performance and since this type is not public I can gloss over the non conformance.

cub/cub/detail/segmented_scan_helpers.cuh
oleksandr-pavlyk: Reworked header file to use simple struct. Removed `#include <cuda/std/tuple>`

cub/cub/detail/segmented_scan_helpers.cuh
oleksandr-pavlyk: Used `mutable` as suggested. I thought supporting callable objects with non-const qualified call operator was necessary as a user-friendly gesture. Let me know if you think we should rather not and I should remove use of `mutable`.

cub/cub/detail/segmented_scan_helpers.cuh
oleksandr-pavlyk: Deleted in favor of aggregate initialziation.

cub/cub/device/dispatch/kernels/kernel_segmented_scan.cuh
oleksandr-pavlyk: Done. Used `static constexpr auto agent_policy = SegmentedScanPolicyGetterT{}().block;`.

cub/cub/device/dispatch/kernels/kernel_segmented_scan.cuh
```diff
@@ -13,15 +17,584 @@
...
+#include <cuda/std/span>
+
 CUB_NAMESPACE_BEGIN
 
 namespace detail::segmented_scan
 {
-template <typename ChainedPolicyT,
+/******************************************************************************
+ * Thread block abstractions
+ ******************************************************************************/
```
bernhardmgruber, line 45: Remark: I dislike such section comments since the grouping they introduce eventually gets out of sync with the code.

Also: what is a thread block abstraction? That sounds like cooperative groups or hierarchies.

cub/cub/device/dispatch/kernels/kernel_segmented_scan.cuh
```diff
@@ -13,15 +17,584 @@
...
+//!   Nullary callable type for getting segmented_scan_policy
+//!
+//! @tparam InputIteratorT
+//!   Random-access input iterator type
+//!
+//! @tparam OutputIteratorT
+//!   Random-access output iterator type
+//!
+//! @tparam OffsetT
+//!   Signed integer type for global offsets
```
bernhardmgruber, line 60: Suggestion: if the signedness of the offset type is critical, consider adding a `static_assert` below inside `agent_segmented_scan`.

cub/cub/device/dispatch/kernels/kernel_segmented_scan.cuh
```diff
@@ -13,15 +17,584 @@
...
+  // Input iterator wrapper type (for applying cache modifier)
+  // Wrap the native input pointer with CacheModifiedInputIterator
+  // or directly use the supplied input iterator type
+  using wrapped_input_iterator_t =
+    ::cuda::std::conditional_t<::cuda::std::is_pointer_v<InputIteratorT>,
+                               CacheModifiedInputIterator<agent_policy.load_modifier, input_t, OffsetT>,
+                               InputIteratorT>;
+
+  // Constants
+
```
bernhardmgruber, line 98: Remark: not needed since we already have 
```c++
  //---------------------------------------------------------------------
  // Types and constants
  //---------------------------------------------------------------------
```
above.

cub/cub/device/dispatch/kernels/kernel_segmented_scan.cuh
```diff
@@ -13,15 +17,584 @@
...
+  // Use cub::NullType means no initial value is provided
```
bernhardmgruber, line 99:
```suggestion
  // Using cub::NullType means no initial value is provided
```

cub/cub/device/dispatch/kernels/kernel_segmented_scan.cuh
```diff
@@ -13,15 +17,584 @@
...
+  // We are relying on either initial value being `NullType`
+  // or the ForceInclusive tag to be true for inclusive scan
+  // to get picked up.
+  static constexpr bool is_inclusive    = ForceInclusive || !has_init;
+  static constexpr int block_threads    = agent_policy.block_threads;
+  static constexpr int items_per_thread = agent_policy.items_per_thread;
+  static constexpr int tile_items       = block_threads * items_per_thread;
+  static constexpr int max_segments     = agent_policy.max_segments;
+
+private:
```
bernhardmgruber, line 110: Suggestion: Does anything above need to be public? Maybe everything before could also be private.

cub/cub/device/dispatch/kernels/kernel_segmented_scan.cuh
```diff
@@ -13,15 +17,584 @@
...
+
+public:
+  // Alias wrapper allowing storage to be unioned
+  using TempStorage = Uninitialized<_TempStorage>;
+
+  _TempStorage& temp_storage; ///< Reference to temp_storage
+  wrapped_input_iterator_t d_in; ///< Input data
+  OutputIteratorT d_out; ///< Output data
+  ScanOpT scan_op; ///< Binary associative scan operator
+  InitValueT initial_value; ///< The initial value element for ScanOpT
```
bernhardmgruber, line 171: Q: Since you provide a constructor, do these members need to be public?

cub/cub/device/dispatch/kernels/kernel_segmented_scan.cuh
```diff
@@ -13,15 +17,584 @@
...
+  //! @param d_in
+  //!   Input data
+  //!
+  //! @param d_out
+  //!   Output data
+  //!
+  //! @param scan_op
+  //!   Binary scan operator
+  //!
+  //! @param init_value
```
bernhardmgruber, line 189:
```suggestion
  //! @param initial_value
```

cub/cub/device/dispatch/kernels/kernel_segmented_scan.cuh
```diff
@@ -13,15 +17,584 @@
...
+  //!   Initial value to seed the exclusive scan
+  //!
```
bernhardmgruber, line 191: Suggestion: the added value of this documentation block is very little, can we delete it?

cub/cub/device/dispatch/kernels/kernel_segmented_scan.cuh
```diff
@@ -13,15 +17,584 @@
...
+  //! @param inp_idx_begin
+  //!   Index of start of the segment in input array
+  //!
+  //! @param inp_idx_end
+  //!  Index of end of the segment in input array
+  //!
+  //! @param out_idx_begin
+  //!  Index of start of the segment's prefix scan result in the output array
+  //!
+  _CCCL_DEVICE _CCCL_FORCEINLINE void consume_range(OffsetT inp_idx_begin, OffsetT inp_idx_end, OffsetT out_idx_begin)
```
bernhardmgruber, line 212: Q: The documentation says this function scans a single segment. But the function name does not reflect this. Can we improve the name of the function? Like:

```suggestion
  //! @brief Scan one segment of values
  //!
  //! @param inp_idx_begin
  //!   Index of start of the segment in input array
  //!
  //! @param inp_idx_end
  //!  Index of end of the segment in input array
  //!
  //! @param out_idx_begin
  //!  Index of start of the segment's prefix scan result in the output array
  //!
  _CCCL_DEVICE _CCCL_FORCEINLINE void scan_segment(OffsetT elem_idx_begin, OffsetT elem_idx_end, OffsetT out_idx_begin)
```

cub/cub/device/dispatch/kernels/kernel_segmented_scan.cuh
```diff
@@ -13,15 +17,584 @@
...
+  {
+    const OffsetT segment_items = ::cuda::std::max(inp_idx_end, inp_idx_begin) - inp_idx_begin;
+    const OffsetT n_chunks      = ::cuda::ceil_div(segment_items, tile_items);
+
+    AccumT exclusive_prefix{};
+    worker_prefix_callback_t prefix_op{exclusive_prefix, scan_op};
+
+    for (OffsetT chunk_id = 0; chunk_id < n_chunks;)
```
bernhardmgruber, line 220: Remark: non-canonical for loops are always a bit surprising. No change requested.

cub/cub/device/dispatch/kernels/kernel_segmented_scan.cuh
```diff
@@ -13,15 +17,584 @@
...
+      }
+    }
+  };
+
+  //! @brief Scan dynamically given number of segments of values
+  template <typename InputBeginOffsetIteratorT,
+            typename InputEndOffsetIteratorT,
+            typename OutputBeginOffsetIteratorT,
+            ::cuda::std::size_t NumSegments = max_segments,
+            class                           = ::cuda::std::enable_if_t<(NumSegments > 1)>>
```
bernhardmgruber, line 279: Important: Please formulate constraints in a way that forces the constraint's evaluation before the compiler can build the function signature. This avoids ambiguity errors in general (not sure if this would actually happen here, but it is good practice).
```suggestion
            ::cuda::std::enable_if_t<(NumSegments > 1), int> = 0>
```

cub/cub/device/dispatch/kernels/kernel_segmented_scan.cuh
```diff
@@ -13,15 +17,584 @@
...
+  _CCCL_DEVICE _CCCL_FORCEINLINE void consume_ranges(
+    InputBeginOffsetIteratorT inp_idx_begin_it,
+    InputEndOffsetIteratorT inp_idx_end_it,
+    OutputBeginOffsetIteratorT out_idx_begin_it,
```
bernhardmgruber, line 283: Suggestion: I guess `inp` does not mean interpolate here. Does it mean `input`? If it does, consider spelling it out. Or consider calling it `element`, or something else.

cub/cub/device/dispatch/kernels/kernel_segmented_scan.cuh
```diff
@@ -13,15 +17,584 @@
...
+    int n_segments)
+  {
+    static_assert(::cuda::std::is_convertible_v<::cuda::std::iter_value_t<InputBeginOffsetIteratorT>, OffsetT>,
+                  "Unexpected iterator type");
+    static_assert(::cuda::std::is_convertible_v<::cuda::std::iter_value_t<InputEndOffsetIteratorT>, OffsetT>,
+                  "Unexpected iterator type");
+    static_assert(::cuda::std::is_convertible_v<::cuda::std::iter_value_t<OutputBeginOffsetIteratorT>, OffsetT>,
+                  "Unexpected iterator type");
```
bernhardmgruber, line 291: Suggestion: I think this should probably use `iter_reference_t`, since you want to know whether `inp_idx_begin_it[i]` (which returns a the iterator's reference) is convertible.

cub/cub/device/dispatch/kernels/kernel_segmented_scan.cuh
```diff
@@ -13,15 +17,584 @@
...
+
+    augmented_accum_t exclusive_prefix{};
+    worker_prefix_callback_t prefix_op{exclusive_prefix, augmented_scan_op};
+
+    using multi_segment_helpers::multi_segmented_input_iterator;
+    using multi_segment_helpers::multi_segmented_output_iterator;
+    using multi_segment_helpers::packer;
+    using multi_segment_helpers::packer_iv;
+    using multi_segment_helpers::projector;
+    using multi_segment_helpers::projector_iv;
```
bernhardmgruber, line 436: Suggestion: maybe we should just remove the namespace `multi_segment_helpers` and spill it's content into the enclosing namespace?

cub/cub/device/dispatch/kernels/kernel_segmented_scan.cuh
```diff
@@ -33,7 +606,10 @@ template <typename ChainedPolicyT,
           typename AccumT,
           bool ForceInclusive,
           typename ActualInitValueT = typename InitValueT::value_type>
-__launch_bounds__(int(ChainedPolicyT::ActivePolicy::segmented_scan_policy_t::BLOCK_THREADS))
+#if _CCCL_HAS_CONCEPTS()
+  requires segmented_scan_policy_selector<PolicySelector>
+#endif // _CCCL_HAS_CONCEPTS()
+__launch_bounds__(int(PolicySelector{}(::cuda::arch_id{CUB_PTX_ARCH / 10}).block.block_threads))
```
bernhardmgruber, line 612: Important: Please use `current_policy`:
```suggestion
__launch_bounds__(current_policy<PolicySelector>().block.block_threads)
```

cub/cub/device/dispatch/kernels/kernel_segmented_scan.cuh
```diff
@@ -42,35 +618,81 @@ __launch_bounds__(int(ChainedPolicyT::ActivePolicy::segmented_scan_policy_t::BLO
     _CCCL_GRID_CONSTANT const BeginOffsetIteratorOutputT begin_offset_d_out,
     _CCCL_GRID_CONSTANT const OffsetT n_segments,
     _CCCL_GRID_CONSTANT const ScanOpT scan_op,
-    _CCCL_GRID_CONSTANT const InitValueT init_value)
+    _CCCL_GRID_CONSTANT const InitValueT init_value,
+    _CCCL_GRID_CONSTANT const int num_segments_per_worker)
 {
-  using segmented_scan_policy_t = typename ChainedPolicyT::ActivePolicy::segmented_scan_policy_t;
+  static constexpr auto policy = PolicySelector{}(::cuda::arch_id{CUB_PTX_ARCH / 10});
```
bernhardmgruber, line 624:
```suggestion
  static constexpr auto policy = current_policy<PolicySelector>();
```

cub/test/catch2_test_device_segmented_scan_multi_segment.cu
bernhardmgruber, line 1: Q: I don't see launch id parameterization (`// %PARAM% TEST_LAUNCH lid 0:1:2`). Does the new segmented implementation not support being called from device code or being graph captured?

bernhardmgruber:
I did a high level pass without trying to understand what the code is doing. Here are a few more comments. Since @NaderAlAwar already approved, there is no need to wait for my further approval.

PR #8641, opened 2026-04-22
Add a script for producing a JSON manifest of cccl headers

wmaxey:
## Description

Dumps the output of an install folder for CCCL. Used for packaging and manifests.


ci/util/manifest.sh
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
+set -u
```
Jacobfaib, line 6:
```suggestion
set -oeu pipefail
```

ci/util/manifest.sh
```diff
@@ -0,0 +1,27 @@
...
+read -r -d '' prog << 'EOF'
+split(.,"\n")
+  | select(length > 2)
+  | [{(.[]) : {"type": "header"}}]
+  | add
+  | $manifest + { "files" : . }
+EOF
+
+find "$path" -wholename "*include/*" -type f -printf "%P\n" \
+  | jq -s --raw-input --argjson manifest "$manifest" "$prog" > $2
```
Jacobfaib:
```suggestion
  | jq -s --raw-input --argjson manifest "$manifest" "$prog" > "${outfile}"
```

ci/util/manifest.sh
```diff
@@ -0,0 +1,27 @@
...
+
+read -r -d '' prog << 'EOF'
+split(.,"\n")
+  | select(length > 2)
+  | [{(.[]) : {"type": "header"}}]
+  | add
+  | $manifest + { "files" : . }
+EOF
+
+find "$path" -wholename "*include/*" -type f -printf "%P\n" \
```
Jacobfaib:
```suggestion
find "$path" -wholename '*include/*' -type f -printf '%P\n' \
```
Or do you need the shell to expand these?

ci/util/manifest.sh
wmaxey: Can't bail on error because of the way `read` works with the unquoted heredoc. 

Best we can do is enforce unassigned variables unless we rewrite this using printf.

Jacobfaib: Maybe add a comment above the `set` line then explaining why we cannot enable `eo pipefail` then. It's the first thing a future dev might think to do when looking at the file.

alliepiper: Agreed, I regularly upgrade those settings out of habit and the comment would be helpful.

alliepiper (approved):

PR #8646, opened 2026-04-23
Add clang-21 to CI

davebayer:


ci/matrix.yaml
```diff
@@ -357,7 +361,7 @@ devcontainer_version: '26.06'
 
 # Compiler versions used for the cuda99.X internal builds:
 cuda99_gcc_version: 14
-cuda99_clang_version: 20
+cuda99_clang_version: 21
```
alliepiper, line 361: This will need to be coordinated with @wmaxey

alliepiper (approved):

ci/matrix.yaml
wmaxey: Done!

bernhardmgruber (approved):

wmaxey (approved):

Jacobfaib (approved):

PR #8651, opened 2026-04-23
Improve `iterator_traits` implementation

miscco:
Our `iterator_traits` implementation has been a thorn in my side forever. It was completely split between a concepts and C++17 implementation which led to a ton of code duplication.

Even worse, it also led to a bug in the C++17 implementation, where we relied on C++20 concept subsumption.

This made it so that we could not use `cuda::std::iterator_traits` for thrust iterators.

Wih that fixed, we can now always use `cuda::std::iterator_traits` for thrust

### DIFF
--- a/libcudacxx/include/cuda/std/__iterator/iterator_traits.h
+++ b/libcudacxx/include/cuda/std/__iterator/iterator_traits.h
@@ -79,47 +79,27 @@ struct __cccl_std_contiguous_iterator_tag_exists : __cccl_type_is_defined<struct
 
 _CCCL_BEGIN_NAMESPACE_CUDA_STD
 
-#if _CCCL_HAS_CONCEPTS()
-
 template <class _Tp>
 using __with_reference = _Tp&;
 
 template <class _Tp>
-concept __can_reference = requires { typename __with_reference<_Tp>; };
-
-template <class _Tp>
-concept __dereferenceable = requires(_Tp& __t) {
-  { *__t } -> __can_reference; // not required to be equality-preserving
-};
-
-// [iterator.traits]
-template <__dereferenceable _Tp>
-using iter_reference_t = decltype(*declval<_Tp&>());
-
-#else // ^^^ _CCCL_HAS_CONCEPTS() ^^^ // vvv _CCCL_HAS_CONCEPTS() vvv
-
-template <class _Tp>
-using __with_reference = _Tp&;
-
-template <class _Tp>
-_CCCL_CONCEPT_FRAGMENT(__can_reference_, requires()(typename(__with_reference<_Tp>)));
-
-template <class _Tp>
-_CCCL_CONCEPT __can_reference = _CCCL_FRAGMENT(__can_reference_, _Tp);
+_CCCL_CONCEPT __can_reference = _CCCL_REQUIRES_EXPR((_Tp))(typename(__with_reference<_Tp>));
 
 _CCCL_DIAG_PUSH
 _CCCL_DIAG_SUPPRESS_CLANG("-Wvoid-ptr-dereference")
+
 template <class _Tp>
-_CCCL_CONCEPT_FRAGMENT(__dereferenceable_, requires(_Tp& __t)(requires(__can_reference<decltype(*__t)>)));
+_CCCL_CONCEPT __dereferenceable = _CCCL_REQUIRES_EXPR((_Tp), _Tp& __t)(requires(__can_reference<decltype(*__t)>));
+
 _CCCL_DIAG_POP
 
-template <class _Tp>
-_CCCL_CONCEPT __dereferenceable = _CCCL_FRAGMENT(__dereferenceable_, _Tp);
-
 // [iterator.traits]
+#if _CCCL_HAS_CONCEPTS()
+template <__dereferenceable _Tp>
+using iter_reference_t = decltype(*::cuda::std::declval<_Tp&>());
+#else // ^^^ _CCCL_HAS_CONCEPTS() ^^^ // vvv _CCCL_HAS_CONCEPTS() vvv
 template <class _Tp>
-using iter_reference_t = enable_if_t<__dereferenceable<_Tp>, decltype(*declval<_Tp&>())>;
-
+using iter_reference_t = enable_if_t<__dereferenceable<_Tp>, decltype(*::cuda::std::declval<_Tp&>())>;
 #endif // _CCCL_HAS_CONCEPTS()
 
 #if _CCCL_COMPILER(NVRTC)
@@ -223,7 +203,8 @@ _CCCL_API inline auto __iter_concept_fn(_Iter, __priority_tag<0>)
                  random_access_iterator_tag>;
 
 template <class _Iter>
-using __iter_concept_t = decltype(::cuda::std::__iter_concept_fn<_Iter>(declval<_Iter>(), __priority_tag<3>{}));
+using __iter_concept_t =
+  decltype(::cuda::std::__iter_concept_fn<_Iter>(::cuda::std::declval<_Iter>(), __priority_tag<3>{}));
 
 template <class _Iter, class = void>
 struct __iter_concept_cache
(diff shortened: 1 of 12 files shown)


libcudacxx/include/cuda/std/__iterator/iterator_traits.h
```diff
@@ -247,525 +227,263 @@ _CCCL_CONCEPT __has_member_pointer = _CCCL_REQUIRES_EXPR((_Tp))(typename(typenam
...
-  using type = typename _Tp::pointer;
-};
-
-template <class _Tp>
-_CCCL_CONCEPT __cpp17_iterator_missing_members =
-  !__specialized_from_std<_Tp> && !__specifies_members<_Tp> && __iterator_traits_detail::__cpp17_iterator<_Tp>;
-
-template <class _Tp>
-_CCCL_CONCEPT __cpp17_input_iterator_missing_members =
-  __cpp17_iterator_missing_members<_Tp> && __iterator_traits_detail::__cpp17_input_iterator<_Tp>;
```
miscco, line 597: I believe that this was the ultimate issue that lead the the inconsistency

This direct translation of the C++20 concept definition requires that `__cpp17_input_iterator_missing_members ` subsumes `__cpp17_iterator_missing_members` which it does in C++20, but its not a thing in C++17

libcudacxx/test/libcudacxx/std/iterators/iterator.primitives/iterator.traits/cxx20_iterator_traits.compile.pass.cpp
```diff
@@ -179,6 +179,77 @@ static_assert(cuda::std::same_as<LegacyInputTraits::reference, LegacyInput::refe
 static_assert(cuda::std::same_as<LegacyInputTraits::pointer, void>);
 static_assert(!has_iterator_concept_v<LegacyInputTraits>);
 
+struct LegacyInputArrow
```
miscco, line 182: We never tested the pointer type of the legacy-cpp17-input-iterator

thrust/thrust/iterator/zip_iterator.h
```diff
@@ -343,7 +343,7 @@ THRUST_NAMESPACE_END
 // The reason is that libcu++ backported the C++20 range iterator machinery to C++17, but C++17 has slightly different
 // language rules, especially regarding `void`. We deemed to it too hard to work around the issues.
 #if _CCCL_STD_VER < 2020
-_CCCL_BEGIN_NAMESPACE_CUDA_STD
+_CCCL_BEGIN_NAMESPACE_STD
```
miscco: Drive by fix to make it work with host STL, lets see what NVRTC says

libcudacxx/include/cuda/std/__iterator/iterator_traits.h
```diff
@@ -79,47 +79,27 @@ struct __cccl_std_contiguous_iterator_tag_exists : __cccl_type_is_defined<struct
...
 
 _CCCL_DIAG_PUSH
 _CCCL_DIAG_SUPPRESS_CLANG("-Wvoid-ptr-dereference")
-template <class _Tp>
-_CCCL_CONCEPT_FRAGMENT(__dereferenceable_, requires(_Tp& __t)(requires(__can_reference<decltype(*__t)>)));
-_CCCL_DIAG_POP
 
 template <class _Tp>
-_CCCL_CONCEPT __dereferenceable = _CCCL_FRAGMENT(__dereferenceable_, _Tp);
+_CCCL_CONCEPT __dereferenceable = _CCCL_REQUIRES_EXPR((_Tp), _Tp& __t)(requires(__can_reference<decltype(*__t)>));
```
Jacobfaib, line 92: Why not using `_Satisfies()`?
```suggestion
_CCCL_CONCEPT __dereferenceable = _CCCL_REQUIRES_EXPR((_Tp), _Tp& __t)
(
  _Satisfies(__can_reference) *__t
);
```

libcudacxx/include/cuda/std/__iterator/iterator_traits.h
```diff
@@ -247,525 +228,263 @@ _CCCL_CONCEPT __has_member_pointer = _CCCL_REQUIRES_EXPR((_Tp))(typename(typenam
...
-  typename common_reference_t<decltype(*__i++)&&, typename indirectly_readable_traits<_Ip>::value_type&>;
-  requires signed_integral<typename incrementable_traits<_Ip>::difference_type>;
-};
+// [iterator.traits#concept:cpp17-iterator]
+template <class _Iter>
+_CCCL_CONCEPT __cpp17_iterator = _CCCL_REQUIRES_EXPR((_Iter), _Iter __i)(
+  requires(copyable<_Iter>),
+  requires(__can_reference<decltype(*__i)>),
+  requires(same_as<decltype(++__i), _Iter&>),
+  requires(__can_reference<decltype(*__i++)>));
```
Jacobfaib: Same here, could be `_Satisfies()` no?

libcudacxx/include/cuda/std/__iterator/iterator_traits.h
```diff
@@ -79,47 +79,27 @@ struct __cccl_std_contiguous_iterator_tag_exists : __cccl_type_is_defined<struct
...
 
 // [iterator.traits]
+#if _CCCL_HAS_CONCEPTS()
+template <__dereferenceable _Tp>
+using iter_reference_t = decltype(*::cuda::std::declval<_Tp&>());
+#else // ^^^ _CCCL_HAS_CONCEPTS() ^^^ // vvv _CCCL_HAS_CONCEPTS() vvv
 template <class _Tp>
-using iter_reference_t = enable_if_t<__dereferenceable<_Tp>, decltype(*declval<_Tp&>())>;
-
+using iter_reference_t = enable_if_t<__dereferenceable<_Tp>, decltype(*::cuda::std::declval<_Tp&>())>;
```
ericniebler, line 102: could this just be:
```c++
template <class _Tp>
using iter_reference_t = decltype(*::cuda::std::declval<_Tp&>());
```
?

libcudacxx/include/cuda/std/__iterator/iterator_traits.h
miscco: Oh I always forget that this exists

libcudacxx/include/cuda/std/__iterator/iterator_traits.h
miscco: I think so, I am not sure about the cases with `convertible_to`

libcudacxx/include/cuda/std/__iterator/iterator_traits.h
miscco: no because it guards against bad `operator*` like ones that return `void`

libcudacxx/include/cuda/std/__iterator/iterator_traits.h
bernhardmgruber: So upgrading to C++20 would have helped after all! :D

thrust/examples/stream_compaction.cu
```diff
@@ -21,7 +21,7 @@ struct is_odd
 template <typename Iterator>
 void print_range(const std::string& name, Iterator first, Iterator last)
 {
-  using T = typename std::iterator_traits<Iterator>::value_type;
+  using T = cuda::std::iter_value_t<Iterator>;
```
bernhardmgruber, line 25: Please also remove the include for `<iterator>`

thrust/examples/transform_iterator.cu
```diff
@@ -48,7 +48,7 @@ struct simple_negate
 template <typename Iterator>
 void print_range(const std::string& name, Iterator first, Iterator last)
 {
-  using T = typename std::iterator_traits<Iterator>::value_type;
+  using T = cuda::std::iter_value_t<Iterator>;
```
bernhardmgruber, line 52: Here as well

thrust/thrust/iterator/iterator_traits.h
```diff
@@ -76,12 +76,8 @@ struct lazy_trait
 //! at compile-time. You can specialize cuda::std::iterator_traits for your own iterator types if needed.
 //! deprecated [Since 3.0]
 template <typename T>
-using iterator_traits CCCL_DEPRECATED_BECAUSE("Use cuda::std::iterator_traits instead") =
-// FIXME(bgruber): switching to ::cuda::std::iterator_traits<T> breaks some tests, e.g. cub.test.device_merge_sort.lid_1
-#if _CCCL_COMPILER(NVRTC)
-  ::cuda
-#endif // _CCCL_COMPILER(NVRTC)
```
bernhardmgruber, line 83: YESSS !!

thrust/thrust/iterator/zip_iterator.h
bernhardmgruber: I think we should rather stay with `cuda::std::iterator_traits` or specialize both, `cuda::std` and `std`.

miscco: Actually I believe that this was a manifestation of the bug

davebayer (approved):

PR #8652, opened 2026-04-23
Make `cuda::execution::tune` public

bernhardmgruber:
- [ ] ~~Ideally, merge the docs before, but not necessary: #8623~~


miscco (approved):
Congratulations, that is an important milestone

davebayer (approved):
We should fix this after we agree on what's actually passed to the tuning

Issue #8644, opened 2026-04-23
[BUG] `catch2_test_device_copy_batched.cu` fails for `sm_120`


davebayer:
This test fails when ran on `sm_120` with CTK 13.2:
```sh
7/184 Test  #13: cub.test.device.copy_batched.lid_0 .............................***Failed    0.52 sec
  -- >> Running:
  	/home/coder/cccl/build/cuda13.2-gcc14/cub-lid0/bin/cub.test.device.copy_batched.lid_0
  Randomness seeded to: 2567978899
  cub.test.device.copy_batched.lid_0: /home/coder/cccl/lib/cmake/thrust/../../../thrust/thrust/detail/allocator/temporary_allocator.h:96: deallocate: Assertion `Exception thrown in deallocate' failed.
  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  cub.test.device.copy_batched.lid_0 is a Catch2 v3.12.0 host application.
  Run with -? for options
  -------------------------------------------------------------------------------
  "DeviceMemcpy::Batched works for a very large number of ranges"(0) - types_233
  - 0
  -------------------------------------------------------------------------------
  /home/coder/cccl/cub/test/catch2_test_device_copy_batched.cu:233
  ...............................................................................
  /home/coder/cccl/cub/test/catch2_test_device_copy_batched.cu:248: FAILED:
    {Unknown expression after the reported line}
  due to a fatal error condition:
    SIGABRT - Abort (abnormal termination) signal
  ===============================================================================
  test cases:  3 |  2 passed | 1 failed
  assertions: 47 | 46 passed | 1 failed
```
I've disabled this test to be run for now on this architecture. We are still compiling, though.

bernhardmgruber:
The line:
```
/home/coder/cccl/lib/cmake/thrust/../../../thrust/thrust/detail/allocator/temporary_allocator.h:96: deallocate: Assertion `Exception thrown in deallocate' failed.
```
could be a hint.

bernhardmgruber:
I can reproduce it in the nvcc 13.2 + gcc 14 devcontainer on my RTX 5090. I cannot reproduce it with nvcc 13.3 and clang 20 on the same GPU.

bernhardmgruber:
```
========= Invalid __global__ write of size 6144 bytes
[...]

PR #8655, opened 2026-04-23
Improve diagnostics of Thrust's `temporary_allocator`

bernhardmgruber:
I initially had a hard time debugging #8644 because `thrust::detail::temporary_allocator` would swallow the error message by CUDA. This PR tries to make such errors visible.

### DIFF
--- a/thrust/thrust/detail/allocator/temporary_allocator.h
+++ b/thrust/thrust/detail/allocator/temporary_allocator.h
@@ -27,8 +27,7 @@
 
 #include <nv/target>
 
-#include <exception>
-#include <iostream>
+#include <cstdio>
 
 #if _CCCL_CUDA_COMPILATION() && _CCCL_DEVICE_COMPILATION()
 #  include <thrust/system/cuda/detail/terminate.h>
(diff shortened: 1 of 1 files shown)


thrust/thrust/detail/allocator/temporary_allocator.h
```diff
@@ -91,13 +94,22 @@ class temporary_allocator : public thrust::detail::tagged_allocator<T, System, t
...
     }
+    // Swallow all exceptions to maintain noexcept contract per C++ allocator requirements.
+    // Deallocate must be noexcept to be safe in destructors and during exception unwinding.
+    // Clear CUDA error state and leak the memory rather than propagating exception.
+    // Memory is leaked, but this matches standard allocator behavior when deallocation fails.
+    _CCCL_CATCH (std::exception & e)
+    {
+      // if we have some context about the thrown exception, log it
+      std::cerr << "Exception thrown in thrust::detail::temporary_allocator::deallocate: " << e.what() << '\n';
+      _CCCL_ASSERT(false, "Exception thrown in thrust::detail::temporary_allocator::deallocate");
```
bernhardmgruber, line 106: I guess `std::cerr` is not everybody's favorit. Should we rather go with `fprintf(stderr, ...)` or something else? The importance for me was to make the error visible.

miscco: I believe fprintf has the smaller include footprint

We should also not need to include exeption here it should already be there

davebayer: Yes, I think we should use `fprintf`, too

davebayer (approved):

PR #8657, opened 2026-04-23
Add a killswitch for warpspeed scan

bernhardmgruber:
We are still seeing bugs with the warpspeed scan implementation, especially wrt. to codegen issues. Let's add a kill switch so users can turn off the implementation (and fall back to the old scan implementation) in case they encounter issues after we shipped it.

- [x] Verify the fix actually works.

### DIFF
--- a/cub/cub/device/dispatch/tuning/tuning_scan.cuh
+++ b/cub/cub/device/dispatch/tuning/tuning_scan.cuh
@@ -959,8 +959,10 @@ struct policy_selector
     // We need `cuda::std::is_constant_evaluated` for the compile-time SMEM computation. And we need PTX ISA 8.6.
     // MSVC + nvcc < 13.1 just fails to compile `cub.test.device.scan.lid_1.types_0` with `Internal error` and nothing
     // else.
+    // The macro `CCCL_DISABLE_WARPSPEED_SCAN` will be left in as a kill-switch for users in case they find any bugs
+    // after we shipped the implementation. TODO(bgruber): remove CCCL_DISABLE_WARPSPEED_SCAN in CCCL 4.0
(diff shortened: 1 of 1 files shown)


cub/cub/device/dispatch/tuning/tuning_scan.cuh
```diff
@@ -959,8 +959,10 @@ struct policy_selector
     // We need `cuda::std::is_constant_evaluated` for the compile-time SMEM computation. And we need PTX ISA 8.6.
     // MSVC + nvcc < 13.1 just fails to compile `cub.test.device.scan.lid_1.types_0` with `Internal error` and nothing
     // else.
+    // The macro `CCCL_DISABLE_WARPSPEED_SCAN` will be left in as a kill-switch for users in case they find any bugs
+    // after we shipped the implementation. TODO(bgruber): remove CCCL_DISABLE_WARPSPEED_SCAN in CCCL 4.0
 #if __cccl_ptx_isa < 860 || !defined(_CCCL_BUILTIN_IS_CONSTANT_EVALUATED) \
-  || ((_CCCL_COMPILER(MSVC) && _CCCL_CUDA_COMPILER(NVCC, <, 13, 1)))
+  || ((_CCCL_COMPILER(MSVC) && _CCCL_CUDA_COMPILER(NVCC, <, 13, 1))) || defined(CCCL_DISABLE_WARPSPEED_SCAN)
```
davebayer, line 965: Please update the macros documentation

davebayer (approved):

miscco (approved):

cub/cub/device/dispatch/tuning/tuning_scan.cuh
bernhardmgruber: I want to keep this macro hidden. Ideally, nobody should ever use it. I only want to reveal its existence if we get a bug report for a CCCL version that already shipped to production and we can no longer fix it.

PR #8668, opened 2026-04-24
[Tile] Disable tests that rely on a return statements inside loops

miscco:
We have a lot of tests that use an conditional return inside a loop. That is currently not supported.

Mark those tests as XFAIL so that we know when it works again


libcudacxx/test/libcudacxx/libcxx/numerics/floating.point/constants/denorm_min.pass.cpp
```diff
@@ -6,7 +6,6 @@
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 // SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES.
 //
-//===----------------------------------------------------------------------===//
```
davebayer, line 9: I don't think you've put here what you wanted

miscco:
Oh that was a merge conflict ^^

PR #8669, opened 2026-04-24
[Tile] Avoid using `asm` optimizations for saturating arithmetics in tile mode

miscco:
We rely on ptx or asm statements for improved runtime performance. 

Disable those optimizations in tile mode


davebayer (approved):
Noooo, my optimizations 😭

PR #8674, opened 2026-04-24
[STF] Fix stream_ctx under CUDA graph capture

caugonnet:
When an STF `stream_ctx` is constructed from a user stream that is already capturing (e.g. opened by `cudaStreamBeginCapture(..., Relaxed)` or `wp.ScopedCapture`), two problems prevented task submission from completing the capture cleanly:

1. In `task::acquire`, the context's `start_events` were only merged as prerequisites when the task had *zero* dependencies. A task with only `write()`-mode dependencies has no *input* predecessor and therefore was never linked to the context's start event. The first pool stream picked for such a task was then synchronized via `cudaStreamWaitEvent` against an event that was never recorded inside the capture, producing `cudaErrorStreamCaptureIsolation` at submission or finalize time.

   Fix: compute a `has_input_dep` flag in the existing dependency loop (true for `read` / `rw` / `reduce` / `relaxed` accesses, including after mode merging). If the task has no input dependency and the context has start events, merge them as prerequisites. This makes every task's first pool stream fork from the user stream inside the capture, so its subsequent work is correctly captured.

2. Passing a user-supplied `async_resources_handle` together with a capturing user stream is still unsafe: the handle's pool very likely carries streams with residual uncaptured work from prior contexts, and CUDA will reject folding them into the ongoing capture. Reject this configuration early with a clear `EXPECT` message.

To implement the check cleanly without losing move semantics on the handle, `backend_ctx_untyped::impl` now records `user_provided_handle = bool(async_resources)` in its member initializer list *before* the handle is moved in. `stream_ctx(cudaStream_t, async_resources_handle)` then inspects `state().user_provided_handle` in its body and, if set, asserts that `user_stream` is not currently capturing.

Adds `cudax/test/stf/local_stf/legacy_to_stf_in_capture.cu`, which runs a diamond-DAG workload twice under an explicit `cudaStreamBeginCapture` in relaxed mode -- once with hand-wired event-based dependencies, once through `stream_ctx` + tokens -- compares the instantiated graphs, validates the host-side result, and also checks that constructing `stream_ctx(capture_stream, handle)` during an active capture triggers the new assertion.

Made-with: Cursor

## Description

closes 

## Checklist


cudax/include/cuda/experimental/__stf/internal/backend_ctx.cuh
```diff
@@ -116,6 +116,10 @@ protected:
     impl(async_resources_handle async_resources = async_resources_handle())
         : auto_scheduler(reserved::scheduler::make(getenv("CUDASTF_SCHEDULE")))
         , auto_reorderer(reserved::reorderer::make(getenv("CUDASTF_TASK_ORDER")))
+        // Record whether the handle was supplied by the caller *before* we
+        // move it into ``async_resources``. Relies on declaration order:
+        // ``user_provided_handle`` is declared before ``async_resources``.
+        , user_provided_handle(bool(async_resources))
```
andralex, line 122: no need for cast here I thinka

cudax/include/cuda/experimental/__stf/internal/acquire_release.cuh
```diff
@@ -125,6 +130,17 @@ inline event_list task::acquire(backend_ctx_untyped& ctx)
     size_t d_index = it - task_deps.begin();
     pimpl->unskipped_indexes.push_back(::std::make_pair(d_index, mode));
 
+    // Track whether this task actually reads from existing data. Modes that
+    // read (and therefore chain back to previous writers / the context start
+    // events through ``fetch_data``) include read, rw, reduce and relaxed.
+    // Write-only modes (write, reduce_no_init) do not.
+    if (!has_input_dep
+        && (mode == access_mode::read || mode == access_mode::rw || mode == access_mode::reduce
+            || mode == access_mode::relaxed))
```
andralex: as discussed there's the question whether this should be just `mode != access_mode::write`

cudax/test/stf/local_stf/legacy_to_stf_in_capture.cu
```diff
@@ -0,0 +1,425 @@
...
+
+using namespace cuda::experimental::stf;
+
+__global__ void initA(double* d_ptrA, size_t N)
+{
+  size_t tid      = blockIdx.x * blockDim.x + threadIdx.x;
+  size_t nthreads = blockDim.x * gridDim.x;
+  for (size_t i = tid; i < N; i += nthreads)
+  {
+    d_ptrA[i] = sin((double) i);
```
andralex:
```suggestion
    d_ptrA[i] = sin(i);
```

cudax/test/stf/local_stf/legacy_to_stf_in_capture.cu
```diff
@@ -0,0 +1,425 @@
...
+  }
+}
+
+__global__ void initB(double* d_ptrB, size_t N)
+{
+  size_t tid      = blockIdx.x * blockDim.x + threadIdx.x;
+  size_t nthreads = blockDim.x * gridDim.x;
+  for (size_t i = tid; i < N; i += nthreads)
+  {
+    d_ptrB[i] = cos((double) i);
```
andralex:
```suggestion
    d_ptrB[i] = cos(i);
```

andralex (approved):
lgtm, added a lil improvement

andralex (approved):

c/experimental/stf/test/test_host_launch.cu
```diff
@@ -101,8 +101,8 @@ C2H_TEST("host_launch with stream context", "[host_launch]")
   stf_host_launch_submit(h, verify_callback);
   stf_host_launch_destroy(h);
 
-  stf_logical_data_destroy(lData);
   stf_ctx_finalize(ctx);
+  stf_logical_data_destroy(lData);
```
caugonnet, line 104: this is weird that we do this

cudax/test/stf/local_stf/legacy_to_stf_in_capture.cu
```diff
@@ -0,0 +1,429 @@
...
+ * @file
+ *
+ * @brief Exercise building an STF context on a caller-provided stream that is
+ *        currently in CUDA graph capture mode.
+ *
+ * The flow is:
+ *
+ *   1. Create an explicit CUDA stream.
+ *   2. Start a capture on it with ``cudaStreamCaptureModeRelaxed``.
+ *   3. Construct a ``stream_ctx`` bound to that captured stream and submit a
```
caugonnet, line 21: fix text it's not a "diamond"

c/experimental/stf/test/test_host_launch.cu
andralex: Here's what Cursor said:

I think the surprising part is that `stf_logical_data_destroy(lData)` happens before `stf_ctx_finalize(ctx)`.

`stf_host_launch_submit(h, verify_callback)` only enqueues the host callback; it does not make the callback run synchronously. The callback is still pending until the context is drained, which happens here at `stf_ctx_finalize(ctx)`. Since that host callback has a dependency on `lData`, destroying the logical data before finalizing the context means we are destroying a handle while queued work still depends on it.

So the safer lifetime ordering is:

```cpp
stf_ctx_finalize(ctx);
stf_logical_data_destroy(lData);
```

That keeps the logical data alive until the queued host callback has actually run. The broader issue may be that the C API lifetime contract is unclear here: a user might reasonably expect stf_logical_data_destroy to be valid after destroying the task/host-launch handle, but queued asynchronous work can still hold references until the context is finalized.

andralex: Cursor does have a point, i.e. the context should be created before the logical data.

caugonnet: this should not be true, there are ref counters (through shared ptrs), the logical data itself has no real meaning once we have finalized. Both should work

cudax/test/stf/local_stf/legacy_to_stf_in_capture.cu
caugonnet: Commit f8b7fdd90d4dd38449a53637c28a455169375b33 fixes that in the stf_c_api where this PR probably comes from. I will refresh the patch

cudax/include/cuda/experimental/__stf/internal/acquire_release.cuh
caugonnet: @andralex i'm not sure i understand you did a lot of modifications in the helper for this, but we don't use it eventually ?

cudax/include/cuda/experimental/__stf/internal/backend_ctx.cuh
caugonnet: Cursor ... but i don't like the overall style so much, it's a lot of mess for just a safety check

cudax/test/stf/local_stf/legacy_to_stf_in_capture.cu
caugonnet: done

cudax/include/cuda/experimental/__stf/internal/acquire_release.cuh
andralex: I left this alone; the changes in the helper just improve its safety (disallows invalid states) and indirectly makes this test here correct. Otherwise it might fail in weird cases.

andralex (approved):
time to get this show on the road!

caugonnet:
I actually found a bug in "has_input_dep" when we have only tokens ... a partial revert is needed

cudax/include/cuda/experimental/__stf/internal/acquire_release.cuh
```diff
@@ -154,9 +170,16 @@ inline event_list task::acquire(backend_ctx_untyped& ctx)
...
+  // A task without any "input" dependency (ie. a dep whose access mode reads
+  // from existing data) has no runtime prerequisites chained back from previous
+  // tasks, so it would otherwise run concurrently with (or before) the context
+  // entry events. Merge the context's start events in that case so that such
+  // tasks still depend on the context's entry point. This is required for
+  // correctness when the context stream is participating in a CUDA graph
+  // capture: without this, tasks dispatched to pool streams would never issue
+  // the ``cudaStreamWaitEvent`` that forks the pool stream into the capture,
+  // and later work on these pool streams would be considered uncaptured.
+  if (!has_input_dep && ctx.has_start_events())
```
caugonnet: this is broken when we have input data as token + a capture on-going, otherwise we would manipulate a stream from the pool without a "fork" from the capturing stream

caugonnet: I've applied the fix here

cudax/include/cuda/experimental/__places/stream_pool.cuh
```diff
@@ -49,6 +49,20 @@ class exec_place;
...
+  }
+
+  auto capture_status = cudaStreamCaptureStatusNone;
+  if (cudaStreamIsCapturing(stream, &capture_status) == cudaSuccess && capture_status != cudaStreamCaptureStatusNone)
+  {
+    // cudaStreamGetDevice/cuStreamGetCtx are not permitted while the stream is
+    // participating in capture. Use the active device, which is the device on
+    // which the capture is being constructed.
+    return cuda_try<cudaGetDevice>();
+  }
```
davebayer, line 64: @pciolkosz shouldn't we do something similar in `cuda::stream_ref`?

cudax/include/cuda/experimental/__stf/internal/constants.cuh
```diff
@@ -34,27 +34,51 @@ namespace cuda::experimental::stf
...
-  relaxed = 4, /* operator ? */
-  reduce  = 8, // overwrite the content of the logical data (if any) with the result of the reduction (equivalent to
-              // write)
+  none           = 0,
+  read           = 1,
+  write          = 2,
+  rw             = 3, // READ + WRITE
+  relaxed        = 4, /* operator ? */
+  reduce         = 9, // READ + reduction; overwrite the content of the logical data (if any) with the reduction result
   reduce_no_init = 16, // special case where the reduction will accumulate into the existing content (equivalent to rw)
```
davebayer, line 43: Suggestion:
```suggestion
  none           = 0b0000,
  read           = 0b0001,
  write          = 0b0010,
  rw             = 0b0011, // READ + WRITE
  relaxed        = 0b0100, /* operator ? */
  reduce         = 0b1001, // READ + reduction; overwrite the content of the logical data (if any) with the reduction result
  reduce_no_init = 0b1111, // special case where the reduction will accumulate into the existing content (equivalent to rw)
```

cudax/include/cuda/experimental/__stf/internal/backend_ctx.cuh
```diff
@@ -349,6 +353,13 @@ protected:
     ::std::atomic<size_t> total_finished_task_cnt = 0;
 #endif
 
+    // True iff the ``async_resources_handle`` was supplied by the caller at
```
davebayer, line 356:
```suggestion
    // True if the ``async_resources_handle`` was supplied by the caller at
```

davebayer (approved):

bernhardmgruber (approved):
Just looked at cmake changes

PR #8679, opened 2026-04-24
Fixup nightly tests for GCC's 8 through 12 for filter_view

Jacobfaib:
## Description

closes 

## Checklist

### DIFF
--- a/libcudacxx/test/libcudacxx/std/ranges/range.adaptors/range.filter/iterator/compare.pass.cpp
+++ b/libcudacxx/test/libcudacxx/std/ranges/range.adaptors/range.filter/iterator/compare.pass.cpp
@@ -7,12 +7,12 @@
 //
 //===----------------------------------------------------------------------===//
 
-// (clang-14 || gcc-12 || msvc-19.39) in C++20 tries to erroneously instantiate a bunch of
+// (clang-14 || gcc-8 through gcc-12 || msvc-19.39) in C++20 tries to erroneously instantiate a bunch of
 // default constructors that don't exist because it evaluates the class initializers before
 // considering the default constructors requirements clause. It's not possible to selectively
 // disable them in this file like the others, so we just disable the compiler entirely.
 
-// UNSUPPORTED: (clang-14 || gcc-12 || msvc-19.39) && c++20
+// UNSUPPORTED: (clang-14 || gcc-8 || gcc-9 || gcc-10 || gcc-11 || gcc-12 || msvc-19.39) && c++20
 
 // friend constexpr bool operator==(iterator const&, iterator const&)
 //  requires equality_comparable<iterator_t<V>>
--- a/ci/matrix.yaml
+++ b/ci/matrix.yaml
@@ -21,6 +21,9 @@ workflows:
   #       args: '--preset libcudacxx --lit-tests "cuda/utility/basic_any.pass.cpp"' }
   #
   override:
+    - {jobs: ['build'], std: 'all', ctk: '12.0', cxx: ['gcc7', 'gcc8', 'gcc9', 'gcc10', 'gcc11', 'gcc12']}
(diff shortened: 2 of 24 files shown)


libcudacxx/test/libcudacxx/std/ranges/range.adaptors/range.filter/iterator/compare.pass.cpp
```diff
@@ -7,12 +7,12 @@
...
 //===----------------------------------------------------------------------===//
 
-// (clang-14 || gcc-12 || msvc-19.39) in C++20 tries to erroneously instantiate a bunch of
+// (clang-14 || gcc-8 through gcc-12 || msvc-19.39) in C++20 tries to erroneously instantiate a bunch of
 // default constructors that don't exist because it evaluates the class initializers before
 // considering the default constructors requirements clause. It's not possible to selectively
 // disable them in this file like the others, so we just disable the compiler entirely.
 
-// UNSUPPORTED: (clang-14 || gcc-12 || msvc-19.39) && c++20
+// UNSUPPORTED: (clang-14 || gcc-8 || gcc-9 || gcc-10 || gcc-11 || gcc-12 || msvc-19.39) && c++20
```
miscco, line 15: I believe this is the wrong approach. We should *not* use class initializers if this breaks supported compilers

miscco (changes requested):
Please drop the offending class initializers

libcudacxx/test/libcudacxx/std/ranges/range.adaptors/range.filter/iterator/compare.pass.cpp
Jacobfaib: The offending lines are the constructor `filter_view::__base_`.
```c++
class filter_view : public view_interface<filter_view<_View, _Pred>>
{
  _CCCL_NO_UNIQUE_ADDRESS _View __base_{}; // <---- here
  _CCCL_NO_UNIQUE_ADDRESS __movable_box<_Pred> __pred_;
```
These versions of GCC and clang were ignoring the requirements in our default constructors and seemingly instantiating these anyways.

If I got rid of the initializer, or change it to `_View __base_ = _View()` or `= {}`, now some of the GCCs complain that `__base_` is uninitialized (I guess this is true if the base class uses pointers -- even if it has a default ctor).

I even tried putting it in a separate class
```c++
template <typename _T>
struct __default_ctor_box
{
  _T __value;

  template <typename _U = _T, enable_if_t<is_default_constructible_v<_U>, int> = 0>
  __default_ctor_box() : __value() { }
};

class filter_view : public view_interface<filter_view<_View, _Pred>>
{
  _CCCL_NO_UNIQUE_ADDRESS __default_ctor_box<_View> __base_;
  _CCCL_NO_UNIQUE_ADDRESS __movable_box<_Pred> __pred_;
```
but it still complained. I could not find a combination of things that passed all compilers without defining these default ctors. Am very open to clever ideas though if you have them.

miscco: Just define the default constructor and initialize appropriately inside

Jacobfaib: We do. The compilers just ignore it. https://github.com/NVIDIA/cccl/blob/526d40fdc0b34265789e6fbadb91ad63e60f9650/libcudacxx/include/cuda/std/__ranges/filter_view.h#L293

Jacobfaib: (And we can't remove = default in C++20 because the standard requires it to be trivially default constructable https://eel.is/c++draft/range.filter.view)

Jacobfaib:
closing in favor of https://github.com/NVIDIA/cccl/pull/8680

PR #8680, opened 2026-04-24
Fix `filter_view` initialization issues

miscco:
Currently we used member initializer with `filter_view`

However, that breaks some compilers in C++20, so just use the plain default constructor if possible

### DIFF
--- a/libcudacxx/test/libcudacxx/std/ranges/range.adaptors/range.filter/base.pass.cpp
+++ b/libcudacxx/test/libcudacxx/std/ranges/range.adaptors/range.filter/base.pass.cpp
@@ -19,16 +19,9 @@
 
 struct Range : cuda::std::ranges::view_base
 {
-  // (clang-14 || gcc-12 || msvc-19.39) in C++20 tries to erroneously instantiate a bunch of
-  // default constructors that don't exist because it evaluates the class initializers before
-  // considering the default constructors requirements clause.
-#if (TEST_COMPILER(CLANG, ==, 14) || TEST_COMPILER(GCC, ==, 12) || TEST_COMPILER(MSVC, <, 19, 44)) \
-  && (TEST_STD_VER == 2020)
-  TEST_FUNC constexpr explicit Range()
+  TEST_FUNC constexpr Range()
       : Range{nullptr, nullptr}
   {}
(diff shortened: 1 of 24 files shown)


libcudacxx/test/libcudacxx/std/ranges/range.adaptors/range.filter/base.pass.cpp
```diff
@@ -19,16 +19,9 @@
 
 struct Range : cuda::std::ranges::view_base
 {
-  // (clang-14 || gcc-12 || msvc-19.39) in C++20 tries to erroneously instantiate a bunch of
-  // default constructors that don't exist because it evaluates the class initializers before
-  // considering the default constructors requirements clause.
-#if (TEST_COMPILER(CLANG, ==, 14) || TEST_COMPILER(GCC, ==, 12) || TEST_COMPILER(MSVC, <, 19, 44)) \
-  && (TEST_STD_VER == 2020)
-  TEST_FUNC constexpr explicit Range()
+  TEST_FUNC constexpr Range()
```
Jacobfaib, line 22: This ctor only exists to work around the buggy compilers, if it works we should remove it.

libcudacxx/test/libcudacxx/std/ranges/range.adaptors/range.filter/begin.pass.cpp
```diff
@@ -28,16 +21,9 @@ struct Range : cuda::std::ranges::view_base
   using Iterator = forward_iterator<int*>;
   using Sentinel = sentinel_wrapper<Iterator>;
 
-  // (clang-14 || gcc-12 || msvc-19.39) in C++20 tries to erroneously instantiate a bunch of
-  // default constructors that don't exist because it evaluates the class initializers before
-  // considering the default constructors requirements clause.
-#if (TEST_COMPILER(CLANG, ==, 14) || TEST_COMPILER(GCC, ==, 12) || TEST_COMPILER(MSVC, <, 19, 44)) \
-  && (TEST_STD_VER == 2020)
   TEST_FUNC constexpr explicit Range()
```
Jacobfaib: Same here. Basically anywhere you see either `Range`, or `InputRange` default ctors in any of these files, it's because it was a workaround for the compilers. We can remove them all

davebayer (approved):

Issue #8633, opened 2026-04-22
cub::DeviceSelect::UniqueByKey might be ambiguous


gonidelis:
` cub::DeviceSelect::UniqueByKey` might be ambiguous after the adding the env overloads.

- [ ] solve the ambiguity 
- [ ] find out why wasn't it caught from tests

Candidate call that showcases the ambiguity:

```cpp
int* d_keys_in          = nullptr;
int* d_values_in        = nullptr;
int* d_keys_out         = nullptr;
int* d_values_out       = nullptr;
int* d_num_selected_out = nullptr;
int num_items           = 0;

auto env = cuda::execution::require(cuda::execution::determinism::run_to_run);

auto a = cub::DeviceSelect::UniqueByKey(
  d_keys_in, d_values_in, d_keys_out, d_values_out, d_num_selected_out, num_items, env);
```

@bernhardmgruber  for viz

bernhardmgruber:
I think we above call is invalid and should be:
```c++
auto a = cub::DeviceSelect::UniqueByKey(
  d_keys_in, d_values_in, d_keys_out, d_values_out, d_num_selected_out, num_items, ::cuda::std::equal_to<>, env);
```
We can try to provide a better error message though by asserting that `EnvT` is not an indirect predicate.

PR #8684, opened 2026-04-25
Add missing env Select::UniqueByKey overload and tests

gonidelis:
solving #8633 I found that there is a missing env overload, it's the one that has `EquialityOp` defaulted.

This solves ambiguity and adds the overload

**Ambiguities analysis beteween all UniqueByKey overloads (env and non-env) and how they separate:**

  1. 6 args (..., num_items)

  - only env-no-equality overload can match.

  2. 7 args (..., num_items, X)

  - env-with-equality matches iff X is an indirect_binary_predicate.
  - env-no-equality matches iff X is not.
  - mutually exclusive.

  3. 8 args **<--- most dangerous one**

  - overlap region is env-with-equality vs non-env-no-equality.
  - I tested the dangerous shape; compiler picks one (usually env if arg1 isn’t void* exactly), not ambiguous.

  4. 9 args

  - non-env-no-equality (with stream arg) vs non-env-with-equality (without stream arg).
  - disambiguated by `is_convertible_v<arg9, cudaStream_t>` gate.

  5. 10 args

  - only non-env-with-equality.

### DIFF
--- a/cub/cub/device/device_select.cuh
+++ b/cub/cub/device/device_select.cuh
@@ -360,7 +360,13 @@ public:
     typename OutputIteratorT,
     typename NumSelectedIteratorT,
     typename NumItemsT,
-    typename EnvT                 = ::cuda::std::execution::env<>,
+    typename EnvT = // Doxygen cannot resolve ::cuda::std::execution::env
+#ifdef _CCCL_DOXYGEN_INVOKED
+    void
+#else
+    ::cuda::std::execution::env<>
+#endif
+    ,
     ::cuda::std::enable_if_t<::cuda::std::is_integral_v<NumItemsT> && !::cuda::std::is_same_v<InputIteratorT, void*>
                                && !::cuda::std::is_same_v<FlagIterator, size_t&>,
                              int> = 0>
@@ -458,7 +464,13 @@ public:
             typename FlagIterator,
             typename NumSelectedIteratorT,
             typename NumItemsT,
-            typename EnvT                 = ::cuda::std::execution::env<>,
+            typename EnvT = // Doxygen cannot resolve ::cuda::std::execution::env
+#ifdef _CCCL_DOXYGEN_INVOKED
+            void
+#else
+            ::cuda::std::execution::env<>
+#endif
+            ,
             ::cuda::std::enable_if_t<::cuda::std::is_integral_v<NumItemsT> && !::cuda::std::is_same_v<IteratorT, void*>
                                        && !::cuda::std::is_same_v<FlagIterator, size_t&>,
                                      int> = 0>
@@ -559,7 +571,13 @@ public:
     typename NumSelectedIteratorT,
     typename SelectOp,
     typename NumItemsT,
-    typename EnvT                 = ::cuda::std::execution::env<>,
+    typename EnvT = // Doxygen cannot resolve ::cuda::std::execution::env
+#ifdef _CCCL_DOXYGEN_INVOKED
+    void
+#else
+    ::cuda::std::execution::env<>
+#endif
+    ,
     ::cuda::std::enable_if_t<::cuda::std::is_integral_v<NumItemsT> && !::cuda::std::is_same_v<InputIteratorT, void*>,
                              int> = 0>
   [[nodiscard]] CUB_RUNTIME_FUNCTION _CCCL_FORCEINLINE static cudaError_t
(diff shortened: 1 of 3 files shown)


cub/cub/device/device_select.cuh
```diff
@@ -1812,6 +1865,127 @@ public:
...
+        bytes,
+        d_keys_in,
+        d_values_in,
+        d_keys_out,
+        d_values_out,
+        d_num_selected_out,
+        ::cuda::std::equal_to<>{},
+        static_cast<offset_t>(num_items),
+        stream);
+    });
```
bernhardmgruber: Suggestion: let's do it like the non-env overload to highlight that this overload just adds a parameter:

```suggestion
    return UniqueByKey(
      d_keys_in,
      d_values_in,
      d_keys_out,
      d_values_out,
      d_num_selected_out,
      num_items,
      ::cuda::std::equal_to<>{},
      stream);
```

cub/cub/device/device_select.cuh
```diff
@@ -360,7 +360,13 @@ public:
...
     typename NumSelectedIteratorT,
     typename NumItemsT,
-    typename EnvT                 = ::cuda::std::execution::env<>,
+    typename EnvT = // Doxygen cannot resolve ::cuda::std::execution::env
+#ifdef _CCCL_DOXYGEN_INVOKED
+    void
+#else
+    ::cuda::std::execution::env<>
+#endif
+    ,
```
bernhardmgruber, line 369: Important: Please file this as a separate PR!

cub/cub/device/device_select.cuh
```diff
@@ -1789,8 +1842,8 @@ public:
     ValueOutputIteratorT d_values_out,
     NumSelectedIteratorT d_num_selected_out,
     NumItemsT num_items,
-    EqualityOpT equality_op = {},
-    EnvT env                = {})
+    EqualityOpT equality_op,
+    EnvT env = {})
```
bernhardmgruber, line 1847: I think we defaulted `equality_op` on purpose to only add a single new env overload instead of two.

bernhardmgruber (changes requested):
After a second inspection, I don't believe we missed an overload. I think we designed the new env-overload on purpose this way. Any user providing an env also needs to provide an equality operator.

cub/cub/device/device_select.cuh
```diff
@@ -1777,10 +1825,15 @@ public:
...
-                               && !::cuda::std::is_convertible_v<EqualityOpT, cudaStream_t>,
+    typename EqualityOpT,
+    typename EnvT = // Doxygen cannot resolve ::cuda::std::execution::env
+#ifdef _CCCL_DOXYGEN_INVOKED
+    void
+#else
+    ::cuda::std::execution::env<>
+#endif
+    ,
+    ::cuda::std::enable_if_t<::cuda::std::indirect_binary_predicate<EqualityOpT, KeyInputIteratorT, KeyInputIteratorT>,
```
NaderAlAwar: Important: we seem to have dropped the old constraints completely. Unless there is a good reason to do so, we should add them back

gonidelis:
> Any user providing an env also needs to provide an equality operator.

why? @bernhardmgruber

cub/cub/device/device_select.cuh
gonidelis: i will for all the other merged algorithms. there is no need to avoid it in this one i think

cub/cub/device/device_select.cuh
gonidelis: but why? i think what happened is here i added only flagged and if https://github.com/NVIDIA/cccl/pull/5335 

and then here github.com/NVIDIA/cccl/pull/8018 i added only the inplace. 

that leaves the algorithm added here out of the joint Ven diagram of the two PRs

i don't see any reason for not adding it

bernhardmgruber: > but why?

To have a smaller API surface. You can call the API like:
```c++
UniqueByKey(..., num_items); // no customization
UniqueByKey(..., num_items, cuda::std::equals_t<>, env); // custom env
UniqueByKey(..., num_items, comp);  // custom compare
UniqueByKey(..., num_items, comp, env);  // custom compare and custom env
```
Adding the new overload would allow to call it like:
```c++
UniqueByKey(..., num_items, env); // custom env
```
I think this small convenience would not justify a full overload.

But if you really want to have the new overload, I am also not blocking it.

bernhardmgruber (approved):

PR #8685, opened 2026-04-25
feat(thrust): mark complex<> members constexpr where safe

### LINKED ISSUE #485: [FEA]: Expand `thrust::complex` to be usable at compile time
### Is this a duplicate?

- [X] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cccl/issues) for this request and that I agree to the [Code of Conduct](CODE_OF_CONDUCT.md)

### Area

Thrust

With the rework of `thrust::complex` to be based upon `cuda::std::complex` we can leverage the later ones `constexpr` support

We should expand existing operators / methods with the appropriate constexpr qualifier.

This would also require us to rework the way we internally test `thrust::complex`

Usually if there was a `void test_something()` function previously one would split it up into a `constexpr bool test_something_impl()` and then use 
```cpp
void test_something() {
    test_something_impl();

#if TEST_STD_VER >= 14
    static_assert(test_something_imple(), "");
#endif // TEST_STD_VER >= 14
}
```

_No response_

### Additional context

_No response_

MukundaKatta:
## Summary
Closes #485. Marks `thrust::complex` constructors, accessors, arithmetic and equality operators `constexpr` where the bodies are constexpr-correct (skipping `abs/arg/norm/polar` which pull in non-constexpr math). Storage now uses default-member-initializers so the defaulted `complex<T>` ctor is itself a constant expression.

## Test plan
- Added `static_assert` battery in `thrust/testing/catch2_test_complex.cu` covering construction, arithmetic, equality, compound assignment, and setters for both `float` and `double`.
- Verified host-only compile with Apple clang 17 (C++17/20/23) and g++ (C++17). CUDA build was NOT verified locally; reviewers should confirm via CI on `nvcc`/`nvhpc`.

## Concerns / open questions
1. Maintainer @miscco indicated in 2024 a preference for routing users to `cuda::std::complex` over investing in `thrust::complex`. Confirming this work is still wanted before going beyond draft.
2. `thrust::complex<T> v;` (default-init) previously left storage uninitialized; the new default-member-initializers zero it. Aligns with `cuda::std::complex` but is a subtle behavior change worth a reviewer note.

### DIFF
--- a/thrust/testing/catch2_test_complex.cu
+++ b/thrust/testing/catch2_test_complex.cu
@@ -689,3 +689,170 @@ TEMPLATE_LIST_TEST_CASE("ComplexExplicitConstruction", "[complex]", FloatingPoin
   const thrust::complex<T> input(42.0, 1337.0);
   [[maybe_unused]] const user_complex result = thrust::exp(input);
 }
+
+// Compile-time tests for thrust::complex (issue #485). These exercise the
+// constexpr-marked constructors, accessors, arithmetic operators and equality
+// operators of thrust::complex<T> via static_assert. They are independent of
+// the runtime test cases above and run at compile time only.
+namespace
+{
+template <typename T>
+constexpr bool test_complex_constexpr_construction()
+{
+  using C = thrust::complex<T>;
+
+  // Default construction zero-initializes.
+  static_assert(C().real() == T(0), "default ctor real");
+  static_assert(C().imag() == T(0), "default ctor imag");
+
+  // Construct from real only.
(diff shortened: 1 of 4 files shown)


miscco:
@MukundaKatta we have recently made a lot of progress in improving our `cuda::std::complex` implementation to have a higher accuracy. 

I do not believe that we want to put any work into `thrust::complex`, but will most likely deprecate it within the 3.X release cycle and hopefully remove with 4.0

PR #8691, opened 2026-04-27
[thrust] Fix `thrust::transform` test for `gcc-15`

davebayer:
According to the comment, this looks like an issue we've dealt before. I've found out that this test is problematic only when `T` is `signed char`, so I've just disabled this test for that type.

### DIFF
--- a/thrust/testing/catch2_test_transform.cu
+++ b/thrust/testing/catch2_test_transform.cu
@@ -11,10 +11,18 @@
 #include "unittest/random.h"
 #include "unittest/special_types.h"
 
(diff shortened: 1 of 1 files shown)


bernhardmgruber (commented):
Can you try whether annotating the function where the GCC error happens with `THRUST_DISABLE_BROKEN_GCC_VECTORIZER`?

davebayer:
> Can you try whether annotating the function where the GCC error happens with `THRUST_DISABLE_BROKEN_GCC_VECTORIZER`?

Works, thanks for the suggestion!

bernhardmgruber (approved):

bernhardmgruber:
I am backporting this to 3.3 because of NVBug 6166900

PR #8694, opened 2026-04-27
Refactor tuning block size checks

bernhardmgruber:


cub/test/catch2_test_device_merge_env.cu
```diff
@@ -24,19 +24,7 @@ DECLARE_LAUNCH_WRAPPER(cub::DeviceMerge::MergePairs, merge_pairs);
...
 
-struct block_size_extracting_less_t
-{
-  int* ptr;
-
-  __device__ bool operator()(const int& lhs, const int& rhs) const
-  {
-    if (threadIdx.x == 0)
-    {
-      atomicMin(ptr, blockDim.x);
```
davebayer, line 35: The new `block_size_extracting_op` uses `atomicMax`, was the original behaviour wrong?

bernhardmgruber: No, it was fine as well. Some code before initialized a device vector with 2048 and then used `atomicMin` to update that to the block size. The new code starts with a device vector of 0 and used `atomicMax`. In both cases, we start with an integer that cannot be a block size.

davebayer (approved):

PR #8700, opened 2026-04-27
Implement parallel `cuda::std::is_heap_until` / `is_heap`

### LINKED ISSUE #7761: [FEA]: Implement CUDA backend for parallel `cuda::std::is_heap`
### Is this a duplicate?

### Area

libcu++

We want to implement a CUDA backend for the following algorithms:

* `std::is_heap` see https://en.cppreference.com/w/cpp/algorithm/is_heap.html
* `std::is_heap_until` see https://en.cppreference.com/w/cpp/algorithm/is_heap_until.html

Not clear how to implement this yet.

It is possible that we might skip this

_No response_

### Additional context

_No response_

edenfunf:
## Description

closes #7761

This implements the heap-checking algorithms for the cuda backend.

* `std::is_heap_until` see https://en.cppreference.com/w/cpp/algorithm/is_heap_until
* `std::is_heap`       see https://en.cppreference.com/w/cpp/algorithm/is_heap

The implementation reuses the existing `__find_if` PSTL dispatch over a
`cuda::counting_iterator` paired with an index-aware predicate that
evaluates `comp(base[(i - 1) / 2], base[i])`. No new dispatch machinery
is required: `is_heap_until` is "find first heap-property violation",
which maps directly onto find_if.

`is_heap` is implemented in terms of `is_heap_until == last`, so the two
stay in lock-step semantically.

Tests follow the `pstl_is_sorted_*` shape and exercise empty range,
valid heap, and broken heap (with violation injected at child index 42
whose parent is index 20), over `all_types`
(uint16/int32/uint64/int128/float/double/half/bfloat16), with raw
pointer and `random_access_iterator` inputs, default `less<>` and custom
`greater<>` (min-heap) comparator, across the four standard sections
(default stream / provided stream / memory_resource / both).

Benchmark mirrors `is_partitioned/basic.cu`, parameterised on element
count and violation position.

## Test results

Local run on RTX 5070 + CUDA 12.9 + MSVC 19.50 (sm_89, Release):

\`\`\`
libcudacxx.test.std.algorithms.alg.sorting.alg.heap.operations.is.heap.pstl_is_heap
  All tests passed (160 assertions in 8 test cases)
libcudacxx.test.std.algorithms.alg.sorting.alg.heap.operations.is.heap.pstl_is_heap_comp
  All tests passed (160 assertions in 8 test cases)
libcudacxx.test.std.algorithms.alg.sorting.alg.heap.operations.is.heap.pstl_is_heap_until
  All tests passed (224 assertions in 8 test cases)
libcudacxx.test.std.algorithms.alg.sorting.alg.heap.operations.is.heap.pstl_is_heap_until_comp
  All tests passed (224 assertions in 8 test cases)
\`\`\`

Total: **768 assertions across 32 test cases, 0 failures**.

The benchmark target (\`libcudacxx.bench.is_heap_until.basic.base\`)
also builds and runs end-to-end (NVBench output verified on a small
config).

## Checklist

### DIFF
--- a/libcudacxx/benchmarks/bench/is_heap/basic.cu
+++ b/libcudacxx/benchmarks/bench/is_heap/basic.cu
@@ -34,7 +34,7 @@ static void basic(nvbench::state& state, nvbench::type_list<T>)
 {
   const auto elements        = static_cast<std::size_t>(state.get_int64("Elements"));
   const auto violation_frac  = state.get_float64("ViolationAt");
-  const auto violation_point = ::cuda::std::clamp<std::size_t>(elements * violation_frac, 0ull, elements - 1);
+  const auto violation_point = cuda::std::clamp<std::size_t>(elements * violation_frac, 0ull, elements - 1);
 
   thrust::device_vector<T> dinput(elements, thrust::no_init);
   prepare_input(dinput, violation_point);
@@ -60,7 +60,7 @@ static void with_predicate(nvbench::state& state, nvbench::type_list<T>)
 {
   const auto elements        = static_cast<std::size_t>(state.get_int64("Elements"));
   const auto violation_frac  = state.get_float64("ViolationAt");
-  const auto violation_point = ::cuda::std::clamp<std::size_t>(elements * violation_frac, 0ull, elements - 1);
+  const auto violation_point = cuda::std::clamp<std::size_t>(elements * violation_frac, 0ull, elements - 1);
 
   thrust::device_vector<T> dinput(elements, thrust::no_init);
   prepare_input(dinput, violation_point);
--- a/libcudacxx/benchmarks/bench/is_heap/basic.cu
+++ b/libcudacxx/benchmarks/bench/is_heap/basic.cu
@@ -46,7 +46,7 @@ static void basic(nvbench::state& state, nvbench::type_list<T>)
 
   state.exec(nvbench::exec_tag::gpu | nvbench::exec_tag::no_batch | nvbench::exec_tag::sync,
              [&](nvbench::launch& launch) {
-               do_not_optimize(cuda::std::is_heap_until(cuda_policy(alloc, launch), dinput.begin(), dinput.end()));
+               do_not_optimize(cuda::std::is_heap(cuda_policy(alloc, launch), dinput.begin(), dinput.end()));
              });
 }
 
@@ -70,11 +70,11 @@ static void with_predicate(nvbench::state& state, nvbench::type_list<T>)
 
   caching_allocator_t alloc{};
 
-  state.exec(
-    nvbench::exec_tag::gpu | nvbench::exec_tag::no_batch | nvbench::exec_tag::sync, [&](nvbench::launch& launch) {
-      do_not_optimize(
-        cuda::std::is_heap_until(cuda_policy(alloc, launch), dinput.begin(), dinput.end(), cuda::std::less<>{}));
-    });
+  state.exec(nvbench::exec_tag::gpu | nvbench::exec_tag::no_batch | nvbench::exec_tag::sync,
+             [&](nvbench::launch& launch) {
+               do_not_optimize(
+                 cuda::std::is_heap(cuda_policy(alloc, launch), dinput.begin(), dinput.end(), cuda::std::less<>{}));
+             });
 }
 
 NVBENCH_BENCH_TYPES(with_predicate, NVBENCH_TYPE_AXES(fundamental_types))
--- a/libcudacxx/benchmarks/bench/is_heap_until/basic.cu
+++ b/libcudacxx/benchmarks/bench/is_heap_until/basic.cu
@@ -34,7 +34,7 @@ static void basic(nvbench::state& state, nvbench::type_list<T>)
 {
   const auto elements        = static_cast<std::size_t>(state.get_int64("Elements"));
   const auto violation_frac  = state.get_float64("ViolationAt");
-  const auto violation_point = ::cuda::std::clamp<std::size_t>(elements * violation_frac, 0ull, elements - 1);
+  const auto violation_point = cuda::std::clamp<std::size_t>(elements * violation_frac, 0ull, elements - 1);
 
   thrust::device_vector<T> dinput(elements, thrust::no_init);
   prepare_input(dinput, violation_point);
@@ -60,7 +60,7 @@ static void with_predicate(nvbench::state& state, nvbench::type_list<T>)
 {
   const auto elements        = static_cast<std::size_t>(state.get_int64("Elements"));
   const auto violation_frac  = state.get_float64("ViolationAt");
-  const auto violation_point = ::cuda::std::clamp<std::size_t>(elements * violation_frac, 0ull, elements - 1);
+  const auto violation_point = cuda::std::clamp<std::size_t>(elements * violation_frac, 0ull, elements - 1);
 
   thrust::device_vector<T> dinput(elements, thrust::no_init);
   prepare_input(dinput, violation_point);
--- a/libcudacxx/benchmarks/bench/is_heap_until/basic.cu
+++ b/libcudacxx/benchmarks/bench/is_heap_until/basic.cu
@@ -1,0 +1,83 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of libcu++, the C++ Standard Library for your entire system,
+// under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES.
+//
+//===----------------------------------------------------------------------===//
+
+#include <thrust/device_vector.h>
+#include <thrust/fill.h>
+
+#include <cuda/functional>
+#include <cuda/std/execution>
+#include <cuda/stream>
+
+#include "nvbench_helper.cuh"
+
+// All-zero is a valid heap; setting one element to 1 forces a violation at
+// that child index since its parent is still 0.
+template <typename T>
+static void prepare_input(thrust::device_vector<T>& d, std::size_t violation_point)
+{
+  thrust::fill(d.begin(), d.end(), T{0});
+  if (violation_point >= 1 && violation_point < d.size())
+  {
+    d[violation_point] = T{1};
+  }
+}
+
+template <typename T>
+static void basic(nvbench::state& state, nvbench::type_list<T>)
+{
+  const auto elements        = static_cast<std::size_t>(state.get_int64("Elements"));
+  const auto violation_frac  = state.get_float64("ViolationAt");
+  const auto violation_point = ::cuda::std::clamp<std::size_t>(elements * violation_frac, 0ull, elements - 1);
+
+  thrust::device_vector<T> dinput(elements, thrust::no_init);
+  prepare_input(dinput, violation_point);
+
+  state.add_global_memory_reads<T>(2 * elements);
+  state.add_global_memory_writes<size_t>(1);
+
+  caching_allocator_t alloc{};
+
+  state.exec(nvbench::exec_tag::gpu | nvbench::exec_tag::no_batch | nvbench::exec_tag::sync,
+             [&](nvbench::launch& launch) {
+               do_not_optimize(cuda::std::is_heap_until(cuda_policy(alloc, launch), dinput.begin(), dinput.end()));
+             });
+}
+
+NVBENCH_BENCH_TYPES(basic, NVBENCH_TYPE_AXES(fundamental_types))
+  .set_name("base")
+  .add_int64_power_of_two_axis("Elements", nvbench::range(16, 28, 4))
+  .add_float64_axis("ViolationAt", std::vector{1.0, 0.5, 0.01});
+
+template <typename T>
+static void with_predicate(nvbench::state& state, nvbench::type_list<T>)
+{
+  const auto elements        = static_cast<std::size_t>(state.get_int64("Elements"));
+  const auto violation_frac  = state.get_float64("ViolationAt");
+  const auto violation_point = ::cuda::std::clamp<std::size_t>(elements * violation_frac, 0ull, elements - 1);
+
+  thrust::device_vector<T> dinput(elements, thrust::no_init);
+  prepare_input(dinput, violation_point);
+
+  state.add_global_memory_reads<T>(2 * elements);
+  state.add_global_memory_writes<size_t>(1);
+
+  caching_allocator_t alloc{};
+
+  state.exec(
+    nvbench::exec_tag::gpu | nvbench::exec_tag::no_batch | nvbench::exec_tag::sync, [&](nvbench::launch& launch) {
+      do_not_optimize(
+        cuda::std::is_heap_until(cuda_policy(alloc, launch), dinput.begin(), dinput.end(), cuda::std::less<>{}));
+    });
+}
+
+NVBENCH_BENCH_TYPES(with_predicate, NVBENCH_TYPE_AXES(fundamental_types))
+  .set_name("with_predicate")
+  .add_int64_power_of_two_axis("Elements", nvbench::range(16, 28, 4))
+  .add_float64_axis("ViolationAt", std::vector{1.0, 0.5, 0.01});
--- a/libcudacxx/include/cuda/std/__pstl/is_heap.h
+++ b/libcudacxx/include/cuda/std/__pstl/is_heap.h
@@ -24,7 +24,6 @@
 #if !_CCCL_COMPILER(NVRTC)
 
 #  include <cuda/__iterator/counting_iterator.h>
-#  include <cuda/__iterator/transform_iterator.h>
 #  include <cuda/__iterator/zip_function.h>
 #  include <cuda/__iterator/zip_iterator.h>
 #  include <cuda/__nvtx/nvtx.h>
@@ -36,7 +35,7 @@
 #  include <cuda/std/__iterator/distance.h>
 #  include <cuda/std/__iterator/iterator_traits.h>
 #  include <cuda/std/__pstl/dispatch.h>
-#  include <cuda/std/__pstl/is_heap_until.h> // for __is_heap_parent_at_fn
+#  include <cuda/std/__pstl/is_heap_until.h>
 #  include <cuda/std/__type_traits/always_false.h>
 #  include <cuda/std/__type_traits/is_execution_policy.h>
 #  include <cuda/std/__utility/move.h>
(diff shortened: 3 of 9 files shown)


libcudacxx/include/cuda/std/__pstl/is_heap.h
```diff
@@ -0,0 +1,56 @@
...
+_CCCL_BEGIN_NAMESPACE_CUDA_STD
+
+_CCCL_BEGIN_NAMESPACE_ARCH_DEPENDENT
+
+_CCCL_TEMPLATE(class _Policy, class _RandomAccessIterator, class _Compare = less<>)
+_CCCL_REQUIRES(__has_random_access_traversal<_RandomAccessIterator> _CCCL_AND is_execution_policy_v<_Policy>)
+[[nodiscard]] _CCCL_HOST_API bool
+is_heap(const _Policy& __policy, _RandomAccessIterator __first, _RandomAccessIterator __last, _Compare __comp = {})
+{
+  return ::cuda::std::is_heap_until(__policy, __first, __last, ::cuda::std::move(__comp)) == __last;
```
miscco: Critical: This is definitely a reasonable approach, but we want to give the user the right information.

That is why even similar algorithms are not implemented in terms of others but open coded. The reason being is that we can 
* set the right nvtx range for the algorithm
* give the right error message. 

Currently if something would go wrong this would talk about `is_heap_until` and not `is_heap` which might be confusing to users

libcudacxx/include/cuda/std/__pstl/is_heap_until.h
```diff
@@ -0,0 +1,117 @@
...
+_CCCL_BEGIN_NAMESPACE_CUDA_STD
+
+// Returns true at the first child index `i` (1 <= i < n) where the parent-child
+// max-heap invariant is broken, i.e. comp(base[(i-1)/2], base[i]) holds.
+template <class _RandomAccessIterator, class _Compare>
+struct __is_heap_until_fn
+{
+  _RandomAccessIterator __base_;
+  _Compare __comp_;
+
```
miscco, line 57: Please add an explicit constructor. We do have some occasions where CTAD fails with some of the older supported compiler combinations and its generally not worth the hassle

libcudacxx/include/cuda/std/__pstl/is_heap_until.h
```diff
@@ -0,0 +1,117 @@
...
+    // Find the first heap-property violation in the index range [1, n).
+    // The result's dereferenced value is the violating child index, or n
+    // if the range is a heap. `__first + *__it` returns __last in the
+    // latter case.
+    const auto __it = __dispatch(
+      __policy,
+      ::cuda::counting_iterator{__diff_t(1)},
+      ::cuda::counting_iterator{__n},
+      __is_heap_until_fn<_RandomAccessIterator, _Compare>{__first, ::cuda::std::move(__comp)});
+    return __first + (*__it);
```
miscco: Nitpick: usually use `__ret` or `__res`

libcudacxx/test/libcudacxx/std/algorithms/alg.sorting/alg.heap.operations/is.heap/pstl_is_heap_until.cu
```diff
@@ -0,0 +1,106 @@
...
+
+inline constexpr int size = 1000;
+
+template <class Policy, class T>
+void test_is_heap_until(const Policy& policy, c2h::device_vector<T>& input)
+{
+  { // empty should not access anything
+    const auto res = cuda::std::is_heap_until(policy, static_cast<T*>(nullptr), static_cast<T*>(nullptr));
+    CHECK(res == nullptr);
+  }
```
miscco, line 39: Important: We should also make sure that sequences of length 1 do not access anything.

We can do so by passing in a host range which would segfault when tried to access on device

libcudacxx/test/libcudacxx/std/algorithms/alg.sorting/alg.heap.operations/is.heap/pstl_is_heap.cu
```diff
@@ -0,0 +1,102 @@
...
+template <class Policy, class T>
+void test_is_heap(const Policy& policy, c2h::device_vector<T>& input)
+{
+  { // empty should not access anything
+    const auto res = cuda::std::is_heap(policy, static_cast<T*>(nullptr), static_cast<T*>(nullptr));
+    CHECK(res);
+  }
+
+  // Strictly decreasing range is a valid max-heap.
+  thrust::sequence(input.begin(), input.end(), static_cast<T>(size), static_cast<T>(-1));
```
miscco: Nitpick: I believe I recently changed it so that we can just do 
```suggestion
  thrust::sequence(input.begin(), input.end(), size, -1);
```

libcudacxx/include/cuda/std/__pstl/is_heap_until.h
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    // Find the first heap-property violation in the index range [1, n).
+    // The result's dereferenced value is the violating child index, or n
+    // if the range is a heap. `__first + *__it` returns __last in the
+    // latter case.
+    const auto __it = __dispatch(
+      __policy,
+      ::cuda::counting_iterator{__diff_t(1)},
+      ::cuda::counting_iterator{__n},
```
miscco: Suggestion: `is_heap` has horrible access patterns that are definitely not cache friendly. However, I would like to see whether it improves throughput if we *also* load the input sequence here via a zip_iterator.

Its fine if that does not help, but we should at least check

miscco (changes requested):
Thanks a lot for the contribution, this looks great 🎉 

I have some issues with fully delegating the implementation of `is_heap`

It does make sense to ensure the implementation is always in lockstep, but it also delivers a worse user experience regarding error messages and tooling. We generally put the onus on us to ensure that we update accordingly. 

I would also like to explore whether we can get some better behavior by loading the input sequence in the `find_if` dispatch.

libcudacxx/include/cuda/std/__pstl/is_heap.h
edenfunf: Done in fe21648b — `is_heap` is now standalone with its own `__find_if` dispatch, its own `_CCCL_NVTX_RANGE_SCOPE("cuda::std::is_heap")`, and its own `static_assert` message, mirroring the `is_sorted` / `is_sorted_until` split.

libcudacxx/include/cuda/std/__pstl/is_heap_until.h
edenfunf: Added in 8e99e31950 — `__is_heap_parent_at_fn` now has an explicit `_CCCL_HOST_DEVICE` constructor.

libcudacxx/include/cuda/std/__pstl/is_heap_until.h
edenfunf: Renamed to `__result` in fe21648b, matching the convention used in `is_sorted`, `mismatch`, and `is_partitioned`.

libcudacxx/test/libcudacxx/std/algorithms/alg.sorting/alg.heap.operations/is.heap/pstl_is_heap_until.cu
edenfunf: Added in 8e99e31950 to all four `pstl_is_heap*.cu` files. The size-1 case now passes a host pointer:

```cpp
{ // size-1 should not access anything; using a host pointer would segfault on device
  T host_value{};
  const auto res = cuda::std::is_heap_until(policy, &host_value, &host_value + 1);
  CHECK(res == &host_value + 1);
}
```
Any device-side access would segfault, so this guards the n < 2 short-circuit against future regressions.

libcudacxx/test/libcudacxx/std/algorithms/alg.sorting/alg.heap.operations/is.heap/pstl_is_heap.cu
edenfunf: Done in 8e99e31950 — applied to all four test files (both the `(size, -1)` decreasing form and the `(0)` increasing form).

libcudacxx/include/cuda/std/__pstl/is_heap_until.h
edenfunf: Implemented in fe21648b — parent and child loads are now exposed to the dispatch via `zip_iterator{transform_iterator{counting, parent_at}, first + 1}`, and the predicate is just `zip_function{comp}` with no indexing inside it.

A/B run with the same benchmark, swapping only the two PSTL headers between the original counting-iterator + index-aware predicate (v1) and the current zip_iterator version (v2). RTX 5070, N = 16M, T = `int32_t`:

| ViolationAt    | v1 GPU time       | v1 BWUtil | v2 GPU time       | v2 BWUtil | Δ throughput |
|:---------------|------------------:|----------:|------------------:|----------:|-------------:|
| 1.0 (full scan)| 243.6 µs          | 82.0 %    | 229.2 µs          | **87.1 %**| **+6.3 %**   |
| 0.5 (mid)      | 153.3 µs          | —         | 146.3 µs          | —         | +4.8 %       |
| 0.01 (early)   |  76.1 µs          | —         |  69.3 µs          | —         | +8.9 %       |

Modest but consistent across all three regimes, so it does help — keeping the zip_iterator form.

edenfunf:
Hi @miscco, thanks for taking the time to go through this so carefully — really useful feedback. Pushed two follow-up commits to address everything:

**fe21648b** is the bigger one: `is_heap` is now standalone (its own dispatch, NVTX scope and error message, no longer routing through `is_heap_until`), and the parent/child loads in `is_heap_until` are now exposed to `find_if` through `zip_iterator{transform_iterator{counting, parent_at}, first + 1}` so the predicate is just a plain `zip_function{comp}`.

**8e99e31950** covers the smaller items: explicit constructor on `__is_heap_parent_at_fn`, the size-1 host-pointer guard you suggested in all four `pstl_is_heap*.cu` files, and the `static_cast<T>(...)` args removed from the `thrust::sequence` calls.

On the zip_iterator question — I was curious about it too, so I ran an A/B with the same bench, swapping only the two PSTL headers between the two versions. On RTX 5070, N = 16M, `int32_t`, the iterator-level form is consistently faster across full-scan / mid / early-exit by roughly 5–9 %, with full-scan SOL bandwidth utilization going from 82 % to 87 %. Full numbers are in the corresponding inline reply.

Local C2H tests are green here (896 assertions across 32 test cases, 0 failures, RTX 5070 + CUDA 12.9 + MSVC 19.50, sm_89 Release).

Let me know if anything else needs tweaking — happy to iterate!

libcudacxx/benchmarks/bench/is_heap_until/basic.cu
```diff
@@ -0,0 +1,83 @@
...
+static void basic(nvbench::state& state, nvbench::type_list<T>)
+{
+  const auto elements        = static_cast<std::size_t>(state.get_int64("Elements"));
+  const auto violation_frac  = state.get_float64("ViolationAt");
+  const auto violation_point = ::cuda::std::clamp<std::size_t>(elements * violation_frac, 0ull, elements - 1);
+
+  thrust::device_vector<T> dinput(elements, thrust::no_init);
+  prepare_input(dinput, violation_point);
+
+  state.add_global_memory_reads<T>(2 * elements);
```
miscco: Important: This is not correct. The algorithm returns early, so we only ever read `violation_point` elements

libcudacxx/benchmarks/bench/is_heap_until/basic.cu
```diff
@@ -0,0 +1,83 @@
...
+static void with_predicate(nvbench::state& state, nvbench::type_list<T>)
+{
+  const auto elements        = static_cast<std::size_t>(state.get_int64("Elements"));
+  const auto violation_frac  = state.get_float64("ViolationAt");
+  const auto violation_point = ::cuda::std::clamp<std::size_t>(elements * violation_frac, 0ull, elements - 1);
+
+  thrust::device_vector<T> dinput(elements, thrust::no_init);
+  prepare_input(dinput, violation_point);
+
+  state.add_global_memory_reads<T>(2 * elements);
```
miscco: Ditto read only 2 * validation_point elements

libcudacxx/include/cuda/std/__pstl/is_heap.h
```diff
@@ -0,0 +1,104 @@
...
+
+    auto __parent_at = ::cuda::transform_iterator{
+      ::cuda::counting_iterator{__diff_t(0)}, __is_heap_parent_at_fn<_RandomAccessIterator>{__first}};
+
+    auto __result = __dispatch(
+      __policy,
+      ::cuda::zip_iterator{__parent_at, __first + 1},
+      ::cuda::zip_iterator{__parent_at + (__n - 1), __last},
+      ::cuda::zip_function{::cuda::std::move(__comp)});
+    return ::cuda::std::get<1>(__result.__iterators()) == __last;
```
miscco: This is a bit more complex than we need it to be. The access at `(__i - 1) / 2` will almost always be a global read

With that we can only load the current iterator
```suggestion
    const auto __result = __dispatch(
      __policy,
      ::cuda::zip_iterator{::cuda::counting_iterator{__diff_t(1)}, __first + 1},
      ::cuda::zip_iterator{::cuda::counting_iterator{__n}, __last},
      ::cuda::zip_function{__is_heap_until_fn<_RandomAccessIterator, _Compare>{__first, ::cuda::std::move(__comp)}});
    return ::cuda::std::get<1>(__result.__iterators());
```

I find this to be equivalent with minimal improvements
[comparison_new.log](https://github.com/user-attachments/files/27157616/comparison_new.log)

miscco (commented):
Thanks a lot for the improvements.

When running some benchmarks I found an issue with the stats in the current benchmark. This is due to the fact that we return early in `find_if`

I also added a `is_heap` benchmark because we want all algorithms to be 

Running the benchmarks I see a nice improvement to the first implementation with just `counting_iterator`
[comparison_orig.log](https://github.com/user-attachments/files/27157648/comparison_orig.log)

miscco:
/ok to test [e1bd997](https://github.com/NVIDIA/cccl/pull/8700/commits/e1bd997570046469e4b16fa95d4963da48bdae0e)

libcudacxx/include/cuda/std/__pstl/is_heap_until.h
```diff
@@ -0,0 +1,124 @@
...
+
+// Returns true at the first child index `i` (1 <= i < n) where the parent-child
+// max-heap invariant is broken, i.e. comp(base[(i-1)/2], base[i]) holds.
+template <class _RandomAccessIterator, class _Compare>
+struct __is_heap_until_fn
+{
+  _RandomAccessIterator __base_;
+  _Compare __comp_;
+
+  _CCCL_HOST_DEVICE explicit __is_heap_until_fn(_RandomAccessIterator __base, _Compare __comp)
```
miscco: Nitpick: This should be `_CCCL_API`, which is more or less the same with some additional attributes added, so that it is marked as hidden when possible. Also it can just be constexpr
```suggestion
  _CCCL_API explicit constexpr __is_heap_until_fn(_RandomAccessIterator __base, _Compare __comp)
```

miscco (approved):
Final nitpick

miscco:
/ok to test [c639654](https://github.com/NVIDIA/cccl/pull/8700/commits/c63965449aee939950b32b8f27f93f1576981702)

libcudacxx/benchmarks/bench/is_heap_until/basic.cu
miscco: This is a bit funny: We do read 2 elements per comparison, so we should read `2 * violation_point`

However, the benchmark still tells me we reached 120% bandwidth, which suggest that it does assume both reads are coalesced

edenfunf:
Thanks a ton @miscco! Really appreciate you not just reviewing but actually going in and pushing the simplification and bench fixes yourself — the single `zip_iterator` form is way cleaner than what I had, and I clearly missed the early-exit accounting in the benchmark.

For context, I'm a high school senior currently learning my way around  CUDA, so apologies for the rougher edges in this one. I'll definitely be more thorough with checks before opening the next PR. Thanks again for taking the time to walk through everything!

miscco:
> For context, I'm a high school senior currently learning my way around CUDA, so apologies for the rougher edges in this one. I'll definitely be more thorough with checks before opening the next PR. Thanks again for taking the time to walk through everything!

I want to make clear that this has been an excellent contribution. The changes you applied were correct and fine.

This is one of the more complex codebases out there, so being able to make a significant contribution is something you can be proud of

edenfunf:
> > For context, I'm a high school senior currently learning my way around CUDA, so apologies for the rougher edges in this one. I'll definitely be more thorough with checks before opening the next PR. Thanks again for taking the time to walk through everything!
> 
> I want to make clear that this has been an excellent contribution. The changes you applied were correct and fine.
> 
> This is one of the more complex codebases out there, so being able to make a significant contribution is something you can be proud of

Thanks @miscco, that genuinely means a lot. Picking up the conventions through your feedback has taught me more than I expected — really looking forward to working on the next one.

libcudacxx/benchmarks/bench/is_heap/basic.cu
```diff
@@ -0,0 +1,83 @@
...
+    d[violation_point] = T{1};
+  }
+}
+
+template <typename T>
+static void basic(nvbench::state& state, nvbench::type_list<T>)
+{
+  const auto elements        = static_cast<std::size_t>(state.get_int64("Elements"));
+  const auto violation_frac  = state.get_float64("ViolationAt");
+  const auto violation_point = ::cuda::std::clamp<std::size_t>(elements * violation_frac, 0ull, elements - 1);
```
NaderAlAwar: Nit: we typically don't use the leading `::` before `cuda` in `.cu` files. We only fully qualify the namespace `cuda` in header files. So this should be
```suggestion
  const auto violation_point = cuda::std::clamp<std::size_t>(elements * violation_frac, 0ull, elements - 1);
```
This also applies to other places in this file and the other benchmark file added

NaderAlAwar (approved):
Really great work @edenfunf ! Left a minor non blocking comment that would be good to address before merging

libcudacxx/benchmarks/bench/is_heap/basic.cu
edenfunf: Thanks! Addressed in b2de463d — dropped the leading `::` from both bench files (the four `cuda::std::clamp` call sites). I'll keep that convention in mind for future `.cu` files.

miscco: To give a little more context. we *must* use full qualification in product code, aka `::cuda::std::`, because someone could have a nested `namespace myproject::cuda` somewhere that would otherwise be ambiguous.

On the other hand that cannot happen in our tests. So there we make sure that everything works also without full qualification

edenfunf: Thanks for the context @miscco — good to know it's actually a guard, not just style. Learned something new!

miscco:
Thanks a lot for this great contribution 🎉

PR #8701, opened 2026-04-27
`libcuda-ptx` Set device-side compilation-only

fbusato:
We recently encountered "false" positive errors in the compilation of `libcuda-ptx` tests, e.g. (nvbug 6064037).

The main issue derives from the full compilation of `libcuda-ptx` tests as single translation units, while they target device-code only. This trigger linker errors that are outside the scope of `libcuda-ptx` instructions.

This PR adds the flags `-rdc=true -dc` to skip full TU compilation only for `libcuda-ptx` tests.


miscco (approved):

davebayer:
@fbusato it seems there are problems with `.pragma "enable_smem_spilling"` in RDC mode. Do you think we should disable the pragma for these cases? I'm just unsure whether it's desired, because it would get disabled without the user getting any warning

fbusato:
@davebayer IIRC `enable_smem_spilling` was manually added, right? I would prefer to disable this test

davebayer:
> `enable_smem_spilling` was manually added, right?

Yes

> I would prefer to disable this test

No problem with that

bernhardmgruber:
> I would prefer to disable this test

How is the test run then? Can it be run at all now? Maybe we should just delete it, or move it out of the PTX tests to a different directory that will run the test? Or at least add a comment why it is disabled.

fbusato:
@bernhardmgruber I added a comment. This is a temporary fix until the compiler addresses the problem

PR #8704, opened 2026-04-27
`cuda::std::simd` Bit

fbusato:
## Description

Implement [[simd.bit]](https://eel.is/c++draft/simd.bit)


libcudacxx/include/cuda/std/__simd/bit.h
```diff
@@ -0,0 +1,380 @@
...
+template <typename _Vp>
+struct __simd_has_single_bit_generator
+{
+  using __result_t = bool;
+  const _Vp& __v_;
+
+  template <typename _Idx>
+  [[nodiscard]] _CCCL_API constexpr __result_t operator()(_Idx) const noexcept
+  {
+    return ::cuda::std::has_single_bit(__v_[_Idx::value]);
```
miscco, line 94: I am strongly wondering, whether it would be possible to use a generic class that takes a function pointer to e.g `cuda::std::has_single_bit<typename _Vp::value_type>` and then forwards to that

libcudacxx/include/cuda/std/__simd/bit.h
```diff
@@ -0,0 +1,380 @@
...
+{
+  using __result_t = typename _Vp0::value_type;
+  const _Vp0& __v0_;
+  const _Vp1& __v1_;
+
+  template <typename _Idx>
+  [[nodiscard]] _CCCL_API constexpr __result_t operator()(_Idx) const noexcept
+  {
+    _CCCL_ASSERT(::cuda::std::in_range<int>(__v1_[_Idx::value]), "rotr: count is out of range");
+    return ::cuda::std::rotr(__v0_[_Idx::value], static_cast<int>(__v1_[_Idx::value]));
```
miscco, line 124: Shouldnt this be in the range of `[0, sizeof(T))`

libcudacxx/include/cuda/std/__simd/bit.h
```diff
@@ -0,0 +1,380 @@
...
+
+_CCCL_TEMPLATE(typename _Tp0,
+               typename _Abi0,
+               typename _Tp1,
+               typename _Abi1,
+               typename _Vp0 = basic_vec<_Tp0, _Abi0>,
+               typename _Vp1 = basic_vec<_Tp1, _Abi1>)
+_CCCL_REQUIRES(__simd_is_valid_rotate_v<_Tp0, _Abi0, _Tp1, _Abi1>)
+[[nodiscard]] _CCCL_API constexpr _Vp0
+rotl(const basic_vec<_Tp0, _Abi0>& __v0, const basic_vec<_Tp1, _Abi1>& __v1) noexcept
```
miscco, line 280: Question: Would it be simpler to read to use 
```suggestion
rotl(const _Vp0& __v0, const _Vp1& __v1) noexcept
```

Not really sure though

libcudacxx/include/cuda/std/__simd/bit.h
```diff
@@ -0,0 +1,380 @@
...
+
+//----------------------------------------------------------------------------------------------------------------------
+// [simd.bit], bit manipulation
+
+_CCCL_TEMPLATE(typename _Tp, typename _Abi, typename _Vp = basic_vec<_Tp, _Abi>)
+_CCCL_REQUIRES(is_integral_v<_Tp>)
+[[nodiscard]] _CCCL_API constexpr _Vp byteswap(const basic_vec<_Tp, _Abi>& __v) noexcept
+{
+  return _Vp{__simd_byteswap_generator<_Vp>{__v}};
+}
```
miscco: Style: I am not too enthusiastic about the defaulted template argument we could just do
```suggestion
_CCCL_TEMPLATE(typename _Tp, typename _Abi)
_CCCL_REQUIRES(is_integral_v<_Tp>)
[[nodiscard]] _CCCL_API constexpr auto byteswap(const basic_vec<_Tp, _Abi>& __v) noexcept
{
  return basic_vec<_Tp, _Abi>{__simd_byteswap_generator{__v}};
}
```

libcudacxx/include/cuda/std/__simd/bit.h
```diff
@@ -0,0 +1,380 @@
...
+
+_CCCL_TEMPLATE(typename _Tp, typename _Abi, typename _Vp = basic_vec<_Tp, _Abi>)
+_CCCL_REQUIRES(__cccl_is_unsigned_integer_v<_Tp>)
+[[nodiscard]] _CCCL_API constexpr _Vp bit_ceil(const basic_vec<_Tp, _Abi>& __v) noexcept
+{
+  return _Vp{__simd_bit_ceil_generator<_Vp>{__v}};
```
miscco: Suggestion: I believe we should be able to deduce `_Vp` for all the generators

libcudacxx/include/cuda/std/__simd/bit.h
fbusato: it would be possible but I'm concerned about performance. I'm in favor of the current approach even if it is verbose

libcudacxx/include/cuda/std/__simd/bit.h
fbusato: I think this is right. `cuda::std::rotr` already have the range checks inside and we need to check that the overflow doesn't happen before calling the function

miscco (approved):

PR #8712, opened 2026-04-28
Expose `CCCL_OS()` macros

### LINKED ISSUE #8711: [FEA]: Expose OS Macros
### Is this a duplicate?

### Area

libcu++

See #8247 

`CCCL_OS()`

_No response_

### Additional context

_No response_

Jacobfaib:
## Description

closes #8711 

## Checklist

### DIFF
--- a/docs/_ext/auto_api_generator.py
+++ b/docs/_ext/auto_api_generator.py
@@ -421,14 +421,22 @@ def extract_doxygen_items(xml_dir):
     return items
 
 
+# Unlike other normal symbols, doxygen treats them as floating in the global namespace
+# (which, technically, is correct). As such, they aren't namespaced, and the name of the
+# generated files would just be <name of file containing macro>_<refid>, which is nearly
+# impossible to name in hand-written rst code.
+#
+# We therefore perform special mangling of the generated file name and prepend this prefix
+# to it.
+MACRO_FILE_NAME_PREFIX = "macro_"
+
+
 def format_macro_doc_filename(macro_name):
     """Create a stable filename for a generated macro page. Macros are not namespace
     members so their refid's (and generated pages) are just the name of the file and a
     hash. This makes it impossible to refer to macros in hand-written rst code."""
-    safe_name = "".join(
-        char if char.isalnum() or char == "_" else "_" for char in macro_name
-    )
-    return f"macro_{safe_name}"
+    macro_name = macro_name.casefold().replace(" ", "_")
+    return f"{MACRO_FILE_NAME_PREFIX}{macro_name}"
 
 
 def extract_doxygen_classes(xml_dir, project_name=None):
@@ -1247,7 +1255,7 @@ def generate_api_docs(app, config):
                 "struct*.rst",
                 "group*.rst",
                 "namespace*.rst",
-                "macro_*.rst",
+                f"{MACRO_FILE_NAME_PREFIX}*.rst",
             ]:
                 for file in api_dir.glob(pattern):
                     file.unlink()
--- a/docs/_ext/auto_api_generator.py
+++ b/docs/_ext/auto_api_generator.py
@@ -207,6 +207,7 @@ def extract_doxygen_items(xml_dir):
         "typedefs": [],
         "enums": [],
         "variables": [],
+        "macros": [],
         "function_groups": {},  # Group functions by name for overloads
         "groups": [],  # Doxygen groups
     }
@@ -323,6 +324,47 @@ def extract_doxygen_items(xml_dir):
 
                 items["typedefs"].append((qualified_name, typedef_refid))
 
+        # Extract documented macros from file compounds. Doxygen records macros as
+        # file-level "define" members, not namespace members.
+        seen_macros = set()
+        for compound in root.findall('.//compound[@kind="file"]'):
+            refid = compound.get("refid")
+            if not refid:
+                continue
+
+            file_xml_file = xml_path / f"{refid}.xml"
+            if not file_xml_file.exists():
+                continue
+
+            try:
+                file_tree = ET.parse(file_xml_file)
+                file_root = file_tree.getroot()
+                for memberdef in file_root.findall('.//memberdef[@kind="define"]'):
+                    macro_refid = memberdef.get("id")
+                    name_elem = memberdef.find("name")
+                    if macro_refid is None or name_elem is None or not name_elem.text:
+                        continue
+
+                    has_description = any(
+                        "".join(description.itertext()).strip()
+                        for description in (
+                            memberdef.find("briefdescription"),
+                            memberdef.find("detaileddescription"),
+                        )
+                        if description is not None
+                    )
+                    if not has_description:
+                        continue
+
+                    macro_key = (name_elem.text, macro_refid)
+                    if macro_key in seen_macros:
+                        continue
+
+                    seen_macros.add(macro_key)
+                    items["macros"].append((name_elem.text, macro_refid))
+            except Exception as e:
+                logger.debug(f"Failed to parse macros from {file_xml_file}: {e}")
+
         # Extract functions, typedefs, enums, and variables from namespaces
         for namespace_compound in namespace_compounds:
             namespace_name = namespace_compound.find("name").text
@@ -377,6 +419,16 @@ def extract_doxygen_items(xml_dir):
         logger.warning(f"Failed to parse Doxygen XML: {e}")
 
     return items
+
+
+def format_macro_doc_filename(macro_name):
+    """Create a stable filename for a generated macro page. Macros are not namespace
+    members so their refid's (and generated pages) are just the name of the file and a
+    hash. This makes it impossible to refer to macros in hand-written rst code."""
+    safe_name = "".join(
+        char if char.isalnum() or char == "_" else "_" for char in macro_name
+    )
+    return f"macro_{safe_name}"
 
 
 def extract_doxygen_classes(xml_dir, project_name=None):
@@ -533,15 +585,20 @@ def generate_group_index_page(group_name, group_refid, project_name, xml_dir, ap
                         member_name = memberdef.find("name")
                         if member_name is None:
                             continue
+                        member_kind = memberdef.get("kind")
                         member_refid = memberdef.get("id")
-                        member_rst_file = Path(api_dir) / f"{member_refid}.rst"
+                        member_filename = (
+                            format_macro_doc_filename(member_name.text)
+                            if member_kind == "define"
+                            else member_refid
+                        )
+                        member_rst_file = Path(api_dir) / f"{member_filename}.rst"
                         # only add refids for exists pages.
                         # Refids corresponding to overloads
                         # do not have associated RST file
                         if not member_rst_file.exists():
                             continue
-                        member_kind = memberdef.get("kind")
-                        members.append((member_kind, member_name.text, member_refid))
+                        members.append((member_kind, member_name.text, member_filename))
         except Exception as e:
             logger.warning(f"Failed to parse group XML {group_xml_file}: {e}")
 
@@ -631,7 +688,7 @@ def generate_member_api_page(
     overload_refids=None,
     xml_dir=None,
 ):
-    """Generate RST content for a single function/typedef/enum/variable API page."""
+    """Generate RST content for a single function/typedef/enum/variable/macro API page."""
     content = []
 
     # Add marker comment for auto-generated files
@@ -646,13 +703,16 @@ def generate_member_api_page(
         "typedef": "doxygentypedef",
         "enum": "doxygenenum",
         "variable": "doxygenvariable",
+        "macro": "doxygendefine",
     }
 
     directive = directive_map.get(member_type, "doxygenfunction")
 
     # For thrust and cub, we need to use the namespace-qualified name
     # For cudax, the member_name already includes the namespace
-    if project_name in ["thrust", "cub"]:
+    if member_type == "macro":
+        qualified_name = member_name
+    elif project_name in ["thrust", "cub"]:
         # If the member_name doesn't already include the namespace, add it
         if "::" not in member_name:
             qualified_name = f"{project_name}::{member_name}"
@@ -1121,6 +1181,24 @@ def generate_namespace_api_page(project_name, items, title=None, doc_prefix=""):
             content.append(format_doc_reference(name, refid, doc_prefix))
         content.append("")
 
+    # Macros section
+    if items["macros"]:
+        content.append("Macros")
+        content.append("~~~~~~")
+        content.append("")
+
+        # Sort macros alphabetically
+        items["macros"].sort(key=lambda x: x[0].lower())
+        for name, refid in items["macros"]:
+            if name in _BREATHE_SKIP_SYMBOLS:
+                logger.info(
+                    f"Skipping macro reference {name} (in _BREATHE_SKIP_SYMBOLS)"
+                )
+                continue
+            filename = format_macro_doc_filename(name)
+            content.append(format_doc_reference(name, filename, doc_prefix))
+        content.append("")
+
     # Keep API reference pages as leaf nodes in sidebar navigation.
     # Symbol/group/category pages are still discoverable on-page and via search.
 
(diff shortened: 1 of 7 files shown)


docs/libcudacxx/extended_api/os.rst
```diff
@@ -0,0 +1,53 @@
+.. _libcudacxx-extended-api-os:
```
fbusato: the scope of these macros are pretty small. What about if we use a single file for all of them?

docs/libcudacxx/extended_api/os.rst
```diff
@@ -0,0 +1,53 @@
...
+For supported operating systems, the macro expands to an implementation-defined truthy
+value if the current operating system matches, or falsey otherwise. Supported values for
+``os`` (and their corresponding operating systems) include:
+
+- WINDOWS: Windows, either in 32-bit or 64-bit mode.
+- LINUX: Any kind of Linux installation. Note that other unix-based operating systems will
+         also match against this.
+- ANDROID: Android operating system.
+- QNX: QNX real-time operating system.
+- APPLE: macOS (Intel or Apple Silicon).
```
fbusato: we don't actually support macOS. not really sure about adding it

libcudacxx/include/cuda/os
```diff
@@ -0,0 +1,56 @@
...
+//!
+//! Passing any other value will result in an undefined expansion, which may or may not be
+//! diagnosed by the compiler.
+//!
+//! @note Some operating systems may satisfy multiple conditions. For example macOS and Android
+//! satisfy both `APPLE`/`ANDROID` and `LINUX`. To detect just one, users are encouraged to
+//! use:
+//!
+//! @code
+//! #if CCCL_OS(LINUX)
```
fbusato: do we need the full example in the code documentation?

libcudacxx/include/cuda/os
```diff
@@ -0,0 +1,56 @@
...
+//! - LINUX
+//! - ANDROID
+//! - QNX
+//! - APPLE
+//!
+//! Passing any other value will result in an undefined expansion, which may or may not be
+//! diagnosed by the compiler.
+//!
+//! @note Some operating systems may satisfy multiple conditions. For example macOS and Android
+//! satisfy both `APPLE`/`ANDROID` and `LINUX`. To detect just one, users are encouraged to
```
Jacobfaib: Android is matched because it also defines `__linux__`, but Apple is matched because of `__LP64__` which smells like a bug (the comment in `__cccl/os.h` indicates that `__LP64__` is supposed to work around NVRTC). I guess it _kinda_ works because Darwin is based on an ancient BSD, but perhaps this should be tightened up before we publicly release this feature.

libcudacxx/include/cuda/os
Jacobfaib: This is actually something I was confused about. How much of the doxygen docstring needs to be repeated in the rst? I'd rather keep most of this in doxygen, but I noticed other APIs seem to repeat themselves in both places.

libcudacxx/include/cuda/os
miscco, line 1: I am not too exited about the new file. I believe that if we want to expose it it should just be available everywhere through the `cccl_config`

libcudacxx/include/cuda/os
```diff
@@ -0,0 +1,56 @@
...
+//!   // Will be visible only on Android
+//! #endif
+//!
+//! #if CCCL_OS(LINUX) && !CCCL_OS(APPLE) && !CCCL_OS(ANDROID)
+//!   // Only visible on Linux
+//! #endif
+//! @endcode
+//!
+//! @return 1 if the specified OS is begin compiled for, 0 otherwise.
+#define CCCL_OS(__os__) _CCCL_OS(__os__)
```
davebayer: Critical: if the user defines `WINDOWS` or any other of these macros, this implementation would expand the macro. We should copy the `_CCCL_OS` implementation here

docs/libcudacxx/extended_api/os.rst
davebayer:
```suggestion
- ``WINDOWS``: Windows, either in 32-bit or 64-bit mode.
- ``LINUX``: Any kind of Linux installation. Note that other unix-based operating systems will
         also match against this.
- ``ANDROID``: Android operating system.
- ``QNX``: QNX real-time operating system.
- ``APPLE``: macOS (Intel or Apple Silicon).
```

davebayer (changes requested):
I agree with @miscco, I don't think we should introduce a separate public header just because of a single macro

libcudacxx/include/cuda/os
fbusato: this should be also added to the unit test

libcudacxx/include/cuda/os
Jacobfaib: You mean `cuda/__cccl_config`? Or `cuda/std/detail/__config`?

libcudacxx/include/cuda/os
Jacobfaib: Should we even have both implementations? Or should we just rename `_CCCL_OS()` to `CCCL_OS()`? I guess the benefit of the private form is that we can test new operating system support without claiming support for it in `CCCL_OS()`, but we can achieve the same effect by just not documenting those operating systems before they are ready for support.

Jacobfaib: > this should be also added to the unit test

Added

docs/libcudacxx/extended_api/os.rst
Jacobfaib: There will eventually be quite a few of them (compiler macros, host/device detection etc.) such that a single file might be quite long (though on the _other_ hand, I do find that the current docs requires clicking through a substantial number of links to find documentation).

libcudacxx/include/cuda/os
fbusato: > I am not too exited about the new file. I believe that if we want to expose it it should just be available everywhere through the cccl_config

not sure about that. Many libraries, e.g. [Boost.Predef](https://www.boost.org/doc/libs/latest/libs/predef/doc/index.html), provide a specific header for macros. 
`__cccl_config` should be used internally but no publicly.
I'm think at the use case where a user only want to include the macro header.

docs/libcudacxx/extended_api/macros.rst
```diff
@@ -0,0 +1,60 @@
+.. _libcudacxx-extended-api-macros:
+
+======
+Macros
+======
+
+CCCL provides a set of convenience macros for detecting various system and compile-time
```
fbusato, line 7: nit. don't manually add line break

docs/libcudacxx/extended_api/macros.rst
```diff
@@ -0,0 +1,60 @@
...
+For supported operating systems, the macro expands to an implementation-defined truthy
+value if the current operating system matches, or falsey otherwise. Supported values for
+``os`` (and their corresponding operating systems) include:
+
+- ``WINDOWS``: Windows, either in 32-bit or 64-bit mode.
+- ``LINUX``: Any kind of Linux installation. Note that other unix-based operating systems will
+         also match against this.
+- ``ANDROID``: Android operating system.
+- ``QNX``: QNX real-time operating system.
+- ``APPLE``: macOS (Intel or Apple Silicon).
```
fbusato: should we add a note to state that CCCL and CTK don't support macOS?

libcudacxx/include/cuda/os
jrhemstad: We shouldn't be repeating ourselves and instead be using `literal-include` to reuse code snippets in docs.

Jacobfaib: Ended up having to add some logic to the auto-api generator to also generate documentation for macros, so now we can just link to the doxygen generated docs from rst.

docs/libcudacxx/extended_api/macros.rst
Jacobfaib: This won't actually introduce any line-breaks in the generated documentation if you are worried about this. E.g. 
image

docs/libcudacxx/extended_api/macros.rst
Jacobfaib: @jrhemstad? IMO the machinery is there, and plenty of people do use thrust from macOS (at least to prototype), I don't know if having macos detection support necessarily forces us into implicitly supporting it.

But I'm happy to add a note about CTK and CCCL not "officially" supporting it.

libcudacxx/include/cuda/os
miscco: I mean it should not be its own header, and people could use `<cuda/std/version>` if they want something small

Jacobfaib:
@miscco @fbusato @davebayer 

Should we even have both implementations, `CCCL_OS()` and `_CCCL_OS()`? Or should we just rename `_CCCL_OS()` to `CCCL_OS()`? I guess the benefit of the private form is that we can test new operating system support without claiming support for it in `CCCL_OS()`, but we can achieve the same effect by just not documenting those operating systems before they are ready for "full" support.

docs/libcudacxx/extended_api/macros.rst
jrhemstad: The machinery may be there, but we don't really have a way to test it. I'm uncomfortable advertising any feature that we can't test.

docs/_ext/auto_api_generator.py
```diff
@@ -323,6 +324,47 @@ def is_nested_and_parent_exists(name, all_classes, all_structs):
 
                 items["typedefs"].append((qualified_name, typedef_refid))
 
+        # Extract documented macros from file compounds. Doxygen records macros as
+        # file-level "define" members, not namespace members.
+        seen_macros = set()
```
miscco, line 329: We should probably exclude `_UGLY`macros

docs/libcudacxx/extended_api/macros.rst
Jacobfaib: Hmmm I could add a note that macOS detection is "best effort" or something like that? No widespread testing but some minimal support? Something along the lines of

> ``APPLE``: ... Users should note that CCCL and the CUDA toolkit do not officially support macOS and that support for this platform is "best effort".

docs/_ext/auto_api_generator.py
Jacobfaib: Hmmmm I would rather do this by skipping the doxygen strings with
```c++
#ifndef _CCCL_DOXYGEN_INVOKED
// @brief Our internal macro
#define _INTERNAL_MACRO(...)
#endif
```
It's not obvious that _this_ file is the reason a particular thing ends up not being documented (actually realizing that the `CCCL_OS()` docstring wasn't appearing due to lack of macro support here took me... a good long while), so I'd rather handle it in the source.

WDYT?

docs/libcudacxx/extended_api/macros.rst
fbusato: I know, the minor issue is in the readability of source documentation with difference width

fbusato:
> I agree with @miscco, I don't think we should introduce a separate public header just because of a single macro

we will definitely have other macros. Anyway, I'm fine with @miscco idea of using `cuda/std/version` is they only need macros.

> Should we even have both implementations, CCCL_OS() and _CCCL_OS()? Or should we just rename _CCCL_OS() to CCCL_OS()?

I'm strongly in favor of keeping both. In this way, we can apply internal workaround, test new OSes, etc. independently, without breaking user code

docs/_ext/auto_api_generator.py
miscco: fine by me

miscco:
> @miscco @fbusato @davebayer
> 
> Should we even have both implementations, `CCCL_OS()` and `_CCCL_OS()`? Or should we just rename `_CCCL_OS()` to `CCCL_OS()`? 

This is a hard NO.

We need to retain the ability to change / break these macros at will. Worst case now we need to duplicate the old implementation. Otherwise we would need to wait for a major CTK release

davebayer (approved):
libcu++ part seems fine to me. I'm not that skilled with docs generation, can someone else review that, please?

docs/libcudacxx/extended_api/macros.rst
Jacobfaib: @jrhemstad thoughts ^^? Or should I just remove any claim of support for it?

docs/_ext/auto_api_generator.py
```diff
@@ -1289,6 +1376,17 @@ def generate_api_docs(app, config):
                 f.write(content)
             logger.info(f"Generated variable API page: {output_file}")
 
+        # Generate individual pages for macros
+        for name, refid in items["macros"]:
```
fbusato, line 1380: it looks too invasive to create a page for each macro and to automate it with doxygen. I would prefer to keep the documentation aligned with all other extended API

Jacobfaib: Hmmm I'm not sure I follow. The extended API all also each get their own pages, listed in the API reference (https://nvidia.github.io/cccl/unstable/libcudacxx/api/index.html). This is what this particular code fragment is doing. It adds a section
```rst
Macros
~~~~~~

- :doc:`CCCL_OS <link_to_manpage...>`
```
at the bottom of that list.

I also add a bespoke page that groups all the macros together (the `extended_api/macros.rst` page), but at the moment there isn't really much to say about the macros that isn't already explained in the manpage, so I just link to that. Once we add more then `macros.rst` might have more to explain.

Jacobfaib: Also prior to the changes in `auto_api_generator.py` we were also not generating any manpages for macros at all (I guess CCCL never publicly exported macros in docs). Doxygen was generating the XML for it, but sphinx wouldn't pick it up because we wouldn't generate the corresponding rst version of it.

We will need these changes as we expose more of the macros, since we shouldn't repeat the docstring twice.

fbusato: I meant that the other extended API uses manually documentation writing, instead of Doxygen, to better control the rendering. I would not use `auto_api_generator.py` only for them.
We don't have Doxygen documentation in most of libcu++ code, we keep the document ion on in the `docs/` side.

Anyway, I don't think it a blocker. From my experience, manual documentation is much better than Doxygen. Maybe in the future, when we will have a new method for the documentation, we can move everything to automated, but again it could be invasive from the code point of view

Jacobfaib: Ah I see what you mean. Yeah that's fair, doxygen doesn't really give you that much control. For reference, this is what it produces for the current docstring
image

fbusato: looks good, but `__os__` is weird in the documentation

miscco: I believe that is fine given that we need it for correctness

fbusato:
I just saw the resulting page here https://nvidia.github.io/cccl/unstable/libcudacxx/api/macro_cccl_os.html#c.CCCL_OS
Honestly I don't like that. The page is unrelated to the other extended APIs. Second, `__os__` looks a reserved keyword, not a user symbol

Jacobfaib:
> The page is unrelated to the other extended APIs.

Where else should it go? `CCCL_OS()` is a symbol in the libcudacxx API, and it is not standardized.

> Second, __os__ looks a reserved keyword, not a user symbol

Hmmm OK I can change it to just `os` if you prefer that.

fbusato:
> Where else should it go? CCCL_OS() is a symbol in the libcudacxx API, and it is not standardized.

Let me show an example. See the difference between `CCCL_OS` and `bitmask` https://nvidia.github.io/cccl/unstable/libcudacxx/extended_api/bit/bitmask.html#libcudacxx-extended-api-bit-bitmask.
I don't see a good reason to have `CCCL_OS` in a difference position, difference left panel

Jacobfaib:
I can make it so that there is a `extended_api/macros/cccl_os.rst` and a table in `extended_api/macros.rst` that links to it, but then I would need to recreate the documentation for `CCCL_OS` in that file.

I thought the whole point was that we didn't need to regurgitate the whole documentation in both the `.rst` and the source?

fbusato:
> I thought the whole point was that we didn't need to regurgitate the whole documentation in both the .rst and the source?

this is the reason I would keep the full documentation in `.rst` instead of source code. Not a blocker but I think we should be consistent for the documentation, otherwise users will see different styles depending on what is documented

Jacobfaib:
https://github.com/NVIDIA/cccl/pull/8797

PR #8719, opened 2026-04-29
[HostJit] Add a test that compiles a simple program including `<cuda/std/initializer_list>`

miscco:
We want to ensure that we can include basic CCCL headers like `<cuda/std/initialilzer_list>`

### DIFF
--- a/c/parallel/test/freestanding/test_basic_cccl_header.cpp
+++ b/c/parallel/test/freestanding/test_basic_cccl_header.cpp
@@ -1,0 +1,73 @@
+//===----------------------------------------------------------------------===//
+// test_basic_cccl_header.cpp
+//
+// Smoke-test for HostJIT compiler infrastructure (issue #7743).
+// JIT-compiles a minimal host+device CUDA source, runs it, and verifies
+// the result — without relying on any system CUDA headers at JIT time.
+//===----------------------------------------------------------------------===//
+
+#include <cassert>
+#include <cstdio>
+
+#include <cuda_runtime.h>
+
+#include <hostjit/config.hpp>
+#include <hostjit/jit_compiler.hpp>
+
+static const char* k_source = R"(
+
+#include <cuda/std/initializer_list>
+
+#include <cuda_runtime.h>
+
+#ifdef _WIN32
+#  define EXPORT __declspec(dllexport)
+#else
+#  define EXPORT __attribute__((visibility("default")))
+#endif
+
+__global__ void device_kernel(int* ptr)
+{
+  ::cuda::std::initializer_list<::std::size_t> meow {42ull, 1337ull};
+  *ptr = static_cast<int>(::std::move(*meow.begin()));
+}
+
+extern "C" EXPORT void host_entry(int* ptr)
+{
+  device_kernel<<<1, 1>>>(ptr);
+}
+)";
+
+int main()
+{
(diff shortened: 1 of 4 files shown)


c/parallel/test/freestanding/test_basic_cccl_header.cpp
```diff
@@ -0,0 +1,73 @@
+//===----------------------------------------------------------------------===//
+// test_basic_cccl_header.cpp
+//
+// Smoke-test for HostJIT compiler infrastructure (issue #7743).
+// JIT-compiles a minimal host+device CUDA source, runs it, and verifies
+// the result — without relying on any system CUDA headers at JIT time.
+//===----------------------------------------------------------------------===//
```
bernhardmgruber, line 9: Q: Does this file need a license header?

c/parallel/src/hostjit/compiler.cpp
```diff
@@ -775,6 +758,7 @@ class CUDACompiler::Impl
     arg_strings.push_back("-D__HOSTJIT_DEVICE_COMPILATION__=1");
     arg_strings.push_back("-DNDEBUG");
     arg_strings.push_back("-DCCCL_DISABLE_CTK_COMPATIBILITY_CHECK");
+    arg_strings.push_back("-D_CCCL_ENABLE_FREESTANDING=1");
```
miscco, line 761: @gevtushenko I am really not sure about this one, but otherwise the compiler is unable to find basic headers like `<assert.h>`

miscco: Discussed internally that its indeed needed

c/parallel/test/freestanding/test_basic_cccl_header.cpp
```diff
@@ -0,0 +1,75 @@
...
+#ifdef _WIN32
+#  define EXPORT __declspec(dllexport)
+#else
+#  define EXPORT __attribute__((visibility("default")))
+#endif
+
+__global__ void device_kernel(int* ptr)
+{
+  ::cuda::std::initializer_list<::std::size_t> meow {42ull, 1337ull};
+  *ptr = static_cast<int>(::std::move(*meow.begin()));
```
bernhardmgruber, line 34: Q: Is that `std::move` on purpose? Should it be `cuda::std::move`?

bernhardmgruber (approved):

shwina:
Hmm, the `hostjit` jobs did not run as part of CI for this PR. @alliepiper any ideas why?

https://github.com/NVIDIA/cccl/blob/6d80568c5bad806394a36f9f62e225e32ebf55cc/ci/project_files_and_dependencies.yaml#L117

c/parallel/test/freestanding/test_basic_cccl_header.cpp
miscco: yes because technically `<cuda/std/initializer_list>` does not include `cuda::std::move`

We just want that `::std::move` is always available

miscco:
> Hmm, the `hostjit` jobs did not run as part of CI for this PR. @alliepiper any ideas why?
> 
> https://github.com/NVIDIA/cccl/blob/6d80568c5bad806394a36f9f62e225e32ebf55cc/ci/project_files_and_dependencies.yaml#L117

maybe because I moved the tests one level down?

PR #8721, opened 2026-04-29
[cub]: Always use `select_policy` to invoke the policy selector

davebayer:
This PR rewrites all `policy_selector{}(arch_id)` invocations to use `cub::detail::select_policy<PolicySelector>(arch_id)` instead. This is necessary to allow us to change the tuning selector invocation mechanism.

### DIFF
--- a/cub/cub/device/dispatch/dispatch_reduce.cuh
+++ b/cub/cub/device/dispatch/dispatch_reduce.cuh
@@ -26,6 +26,7 @@
 #include <cub/device/dispatch/kernels/kernel_reduce.cuh>
 #include <cub/device/dispatch/tuning/tuning_reduce.cuh>
 #include <cub/grid/grid_even_share.cuh>
+#include <cub/util_arch.cuh>
 #include <cub/util_debug.cuh>
 #include <cub/util_device.cuh>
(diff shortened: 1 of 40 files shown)


bernhardmgruber:
CCCL.C may want to have a word with you in host code, where we violate the documented requirements of policy selectors to be stateless :)

cub/cub/device/dispatch/dispatch_reduce.cuh
```diff
@@ -778,7 +779,7 @@ CUB_RUNTIME_FUNCTION _CCCL_FORCEINLINE auto dispatch(
     return error;
   }
 
-  const reduce_policy active_policy = policy_selector(arch_id);
+  const reduce_policy active_policy = select_policy<PolicySelector>(arch_id);
```
bernhardmgruber, line 782: Critical: This will break CCCL.C, which depends on using stateful policy selectors in host code, and injecting stateless policy selectors for the device code using the `kernel_source`.

I did not document this behavior anywhere, since it is obsolete if we can get the host code JIT-ing to work. Let's wait with this PR until we get more experience with the JIT.

bernhardmgruber (changes requested):

PR #8722, opened 2026-04-29
[HostJit] Enable HostJit to compile `cub::DeviceAdjacentDifference`

miscco:
We need some additional work to ensure that we can actually compile cub kernels


miscco:
Closed in favor of #8801

PR #8723, opened 2026-04-29
[HostJit] Add basic implementations of string functions

miscco:
We need implementations of e.g `strcpy` for CCCL.

Those are defined in the host string.h We can use the compiler builtins for them just fine. This reduces code changes needed in libcu++ considerably and is also simply correct.

## Description

closes 

## Checklist


bernhardmgruber (approved):

bernhardmgruber:
So, libcudacxx is now also adding a C library? :D

davebayer (approved):
Reimplementing STL never gets old :)

PR #8726, opened 2026-04-29
[HostJit] Add `_CCCL_HOSTJIT` macro

miscco:
We need to know when we are explicitly compiling for freestanding and not just for NVRTC

Also move the macro to the right place

### DIFF
--- a/docs/cccl/development/macro.rst
+++ b/docs/cccl/development/macro.rst
@@ -496,3 +496,58 @@ Usage example:
     _CCCL_DIAG_SUPPRESS_GCC("-Wattributes")
     // code ..
     _CCCL_DIAG_POP
+
+----
+
+Freestanding support
+--------------------------
+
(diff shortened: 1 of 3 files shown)


davebayer (approved):

bernhardmgruber (approved):

jrhemstad:
How is `_CCCL_HOSTJIT` different from `_CCCL_FREESTANDING`?

Can we add some docs to the developer guide on the definition of these new macros?

miscco:
> How is `_CCCL_HOSTJIT` different from `_CCCL_FREESTANDING`?
> 
> Can we add some docs to the developer guide on the definition of these new macros?

`_CCCL_FREESTANDING` also includes `NVRTC` and this is explicitly for only host jit compilation

miscco:
> _CCCL_FREESTANDING

Added some documentation

PR #8728, opened 2026-04-29
[cub] Replace `arch_id` with `compute_capability`

### LINKED ISSUE #7654: Tuning policy selectors should use the compute capability instead of `cuda::arch_id`

### LINKED ISSUE #8709: [FEA]: Finalize CUB tuning policy selector API

davebayer:


c/parallel/src/merge_sort.cu
```diff
@@ -235,8 +235,7 @@ try
   const auto policy_sel = cub::detail::merge_sort::policy_selector{static_cast<int>(input_keys_it.value_type.size)};
 
   // TODO(bgruber): drop this if tuning policies become formattable
-  std::stringstream policy_sel_str;
-  policy_sel_str << policy_sel(cuda::to_arch_id(cuda::compute_capability{cc_major, cc_minor}));
+  policy_sel_str << policy_sel(cc);
```
bernhardmgruber, line 239: Q: Where did the stream `policy_sel_str` go?

davebayer: :facepalm:

cub/benchmarks/bench/transform/common.h
```diff
@@ -28,10 +28,11 @@
 #if !TUNE_BASE
 struct policy_selector
 {
-  _CCCL_API constexpr auto operator()(cuda::arch_id) const -> cub::detail::transform::transform_policy
+  _CCCL_API constexpr auto operator()(cuda::compute_capability) const -> cub::detail::transform::transform_policy
   {
     const int min_bytes_in_flight =
-      cub::detail::transform::arch_to_min_bytes_in_flight(::cuda::arch_id{__CUDA_ARCH_LIST__ / 10}) + TUNE_BIF_BIAS;
+      cub::detail::transform::cc_to_min_bytes_in_flight(::cuda::compute_capability{__CUDA_ARCH_LIST__ / 10})
+      + TUNE_BIF_BIAS;
```
bernhardmgruber: Suggestion: I am pretty certain this can be:

```suggestion
  _CCCL_API constexpr auto operator()(cuda::compute_capability cc) const -> cub::detail::transform::transform_policy
{
const int min_bytes_in_flight =
      cub::detail::transform::cc_to_min_bytes_in_flight(cc) + TUNE_BIF_BIAS;
```

cub/benchmarks/bench/transform/common.h
```diff
@@ -62,11 +63,12 @@ struct policy_selector
...
-    constexpr auto algorithm   = cub::detail::transform::Algorithm::ublkcp;
-    auto policy                = cub::detail::transform::async_copy_policy{};
-    policy.block_threads       = TUNE_THREADS;
-    policy.bulk_copy_alignment = cub::detail::transform::bulk_copy_alignment(::cuda::arch_id{__CUDA_ARCH_LIST__ / 10});
-    policy.unroll_factor       = TUNE_UNROLL_FACTOR;
+    constexpr auto algorithm = cub::detail::transform::Algorithm::ublkcp;
+    auto policy              = cub::detail::transform::async_copy_policy{};
+    policy.block_threads     = TUNE_THREADS;
+    policy.bulk_copy_alignment =
+      cub::detail::transform::bulk_copy_alignment(::cuda::compute_capability{__CUDA_ARCH_LIST__ / 10});
```
bernhardmgruber:
```suggestion
      cub::detail::transform::bulk_copy_alignment(cc);
```

cub/cub/detail/launcher/cuda_runtime.cuh
```diff
@@ -36,9 +38,9 @@ struct TripleChevronFactory
   }
 
   template <class T = void>
-  CUB_RUNTIME_FUNCTION ::cudaError_t PtxArchId(::cuda::arch_id& arch_id) const
+  CUB_RUNTIME_FUNCTION ::cudaError_t PtxCC(::cuda::compute_capability& cc) const
   {
-    return ptx_arch_id<T>(arch_id);
+    return ptx_cc<T>(cc);
   }
```
bernhardmgruber: Remark: I advocated for `PtxCC` on Slack since we will use the function so often and it's internal. We should consider to give it a longer name still like `PtxCompCap`. I think `PtxComputeCapability` is a bit too long.

cub/cub/detail/arch_dispatch.cuh
```diff
@@ -24,37 +24,38 @@ CUB_NAMESPACE_BEGIN
...
 #  if _CCCL_STD_VER < 2020
 template <typename PolicySelector, size_t N>
 _CCCL_API constexpr auto find_lowest_arch_with_same_policy(
-  PolicySelector policy_selector, size_t i, const ::cuda::std::array<::cuda::arch_id, N>& all_arches) -> ::cuda::arch_id
+  PolicySelector policy_selector, size_t i, const ::cuda::std::array<::cuda::compute_capability, N>& all_ccs)
+  -> ::cuda::compute_capability
 {
-  const auto policy = policy_selector(all_arches[i]);
-  while (i > 0 && policy_selector(all_arches[i - 1]) == policy)
+  const auto policy = policy_selector(all_ccs[i]);
```
bernhardmgruber, line 53: Remark: I just bought an electrical vehicle and read `ccs` as `Combined Charging System` in my head :D

No change requested.

cub/cub/detail/arch_dispatch.cuh
```diff
@@ -66,60 +67,62 @@ struct lowest_arch_resolver<ArchMult, ::cuda::std::integer_sequence<int, CudaArc
...
+    return all_ccs[i];
   }
 
-  static constexpr ::cuda::arch_id lowest_arch_with_same_policy[sizeof...(Is)] = {find_lowest(Is)...};
+  static constexpr ::cuda::compute_capability lowest_arch_with_same_policy[sizeof...(Is)]{find_lowest(Is)...};
 };
 #  endif // if _CCCL_STD_VER < 2020
 
 template <int ArchMult, int... CudaArches, typename PolicySelector, typename FunctorT, size_t... Is>
 CUB_RUNTIME_FUNCTION _CCCL_FORCEINLINE cudaError_t dispatch_to_arch_list(
```
bernhardmgruber: Suggestion: Consider renaming the function to `dispatch_to_cc_list`.

cub/cub/detail/launcher/cuda_runtime.cuh
davebayer: I've renamed it to `PtxComputeCap`

cub/cub/detail/arch_dispatch.cuh
```diff
@@ -169,10 +175,11 @@ dispatch_arch(PolicySelector policy_selector, ::cuda::arch_id device_arch, F&& f
 // if we are compiling CCCL.C with runtime policies, we cannot query the policy hub at compile time
 _CCCL_EXEC_CHECK_DISABLE
 template <typename PolicySelector, typename F>
-_CCCL_API _CCCL_FORCEINLINE cudaError_t dispatch_arch(PolicySelector policy_selector, ::cuda::arch_id device_arch, F&& f)
+_CCCL_API _CCCL_FORCEINLINE cudaError_t
+dispatch_arch(PolicySelector policy_selector, ::cuda::compute_capability device_cc, F&& f)
```
bernhardmgruber: Suggestion: Can we also rename the function to `dispatch_comp_cap` and then may the file to `comp_cap_dispatch` etc.? So all is consistently saying CC and not arch.

cub/cub/device/dispatch/kernels/kernel_reduce.cuh
```diff
@@ -18,7 +18,7 @@
 #include <cub/grid/grid_even_share.cuh>
 #include <cub/util_arch.cuh>
 
-#include <cuda/__device/arch_id.h>
+#include <cuda/__device/compute_capability.h>
```
bernhardmgruber, line 21:
```suggestion
```

cub/cub/device/dispatch/kernels/kernel_reduce_deterministic.cuh
```diff
@@ -20,7 +20,7 @@
 
 #include <thrust/type_traits/unwrap_contiguous_iterator.h>
 
-#include <cuda/__device/arch_id.h>
+#include <cuda/__device/compute_capability.h>
```
bernhardmgruber, line 23:
```suggestion
```

cub/cub/device/dispatch/kernels/kernel_scan_warpspeed.cuh
```diff
@@ -29,7 +29,7 @@
 #include <cub/warp/warp_scan.cuh>
 
 #include <cuda/__cmath/ceil_div.h>
-#include <cuda/__device/arch_id.h>
+#include <cuda/__device/compute_capability.h>
```
bernhardmgruber, line 32:
```suggestion
```

cub/cub/device/dispatch/kernels/kernel_segmented_radix_sort.cuh
```diff
@@ -20,7 +20,7 @@
 #include <cub/device/dispatch/tuning/tuning_radix_sort.cuh>
 #include <cub/util_arch.cuh>
 
-#include <cuda/__device/arch_id.h>
+#include <cuda/__device/compute_capability.h>
```
bernhardmgruber, line 23:
```suggestion
```

cub/cub/device/dispatch/kernels/kernel_segmented_reduce.cuh
```diff
@@ -19,7 +19,7 @@
 #include <cub/iterator/arg_index_input_iterator.cuh>
 #include <cub/util_arch.cuh>
 
-#include <cuda/__device/arch_id.h>
+#include <cuda/__device/compute_capability.h>
```
bernhardmgruber, line 22:
```suggestion
```

cub/cub/device/dispatch/kernels/kernel_transform.cuh
```diff
@@ -24,7 +24,7 @@
 
 #include <cuda/__cmath/pow2.h>
 #include <cuda/__cmath/round_up.h>
-#include <cuda/__device/arch_id.h>
+#include <cuda/__device/compute_capability.h>
```
bernhardmgruber, line 27:
```suggestion
```

cub/cub/device/dispatch/tuning/tuning_histogram.cuh
```diff
@@ -293,9 +293,9 @@ private:
   }
 
 public:
-  [[nodiscard]] _CCCL_API constexpr auto operator()(::cuda::arch_id arch) const -> histogram_policy
+  [[nodiscard]] _CCCL_API constexpr auto operator()(::cuda::compute_capability cc) const -> histogram_policy
   {
-    if (arch >= ::cuda::arch_id::sm_100)
+    if (cc >= ::cuda::compute_capability{100})
```
bernhardmgruber: Suggestion: I would prefer to spell it like:
```suggestion
    if (cc >= ::cuda::compute_capability{10, 0})
```
If you agree as well, can you apply this change EVERYWHERE? :)

cub/cub/device/dispatch/tuning/tuning_radix_sort.cuh
```diff
@@ -2041,7 +2041,7 @@ struct policy_selector
         alt_segmented};
     }
 
-    if (static_cast<int>(arch) >= 62) // TODO(bgruber): add ::cuda::arch_id::sm_62
+    if (cc >= ::cuda::compute_capability{62}) // TODO(bgruber): add ::cuda::compute_capability{62}
```
bernhardmgruber:
```suggestion
    if (cc >= ::cuda::compute_capability{62})
```

cub/cub/device/dispatch/dispatch_adjacent_difference.cuh
```diff
@@ -338,19 +338,19 @@ CUB_RUNTIME_FUNCTION _CCCL_FORCEINLINE auto dispatch(
...
 
-  const adjacent_difference_policy active_policy = policy_selector(arch_id);
+  const adjacent_difference_policy active_policy = policy_selector(cc);
 #if _CCCL_HOSTED() && defined(CUB_DEBUG_LOG)
   NV_IF_TARGET(
     NV_IS_HOST, ({
       ::std::stringstream ss;
       ss << active_policy;
-      _CubLog("Dispatching DeviceAdjacentDifference to arch %d with tuning: %s\n", (int) arch_id, ss.str().c_str());
+      _CubLog("Dispatching DeviceAdjacentDifference to arch %d with tuning: %s\n", cc.get(), ss.str().c_str());
```
bernhardmgruber: Strong suggestion: Since the log statement is user facing and I would like to promote it more in the future (so users can debug what tunings apply), we should format this better:
```suggestion
      _CubLog("Dispatching DeviceAdjacentDifference to compute capability %d.%d with tuning: %s\n", cc.major(), cc.minor(), ss.str().c_str());
```
Applies to all dispatch functions.

cub/benchmarks/bench/reduce/arg_extrema.cu
```diff
@@ -18,7 +18,8 @@
 #if !TUNE_BASE
 struct tuned_policy_selector
 {
-  [[nodiscard]] _CCCL_API constexpr auto operator()(::cuda::arch_id) const -> cub::detail::reduce::reduce_policy
+  [[nodiscard]] _CCCL_API constexpr auto operator()(::cuda::compute_capability) const
```
miscco: Nitpick: 
```suggestion
  [[nodiscard]] _CCCL_API constexpr auto operator()(cuda::compute_capability) const
```

cub/benchmarks/bench/reduce/base.cuh
```diff
@@ -13,7 +13,7 @@
 #if !TUNE_BASE
 struct policy_selector
 {
-  _CCCL_API constexpr auto operator()(cuda::arch_id) const -> ::cub::reduce_policy
+  _CCCL_API constexpr auto operator()(cuda::compute_capability) const -> ::cub::reduce_policy
```
miscco: Nitpick:
```suggestion
   [[nodiscard]] _CCCL_API constexpr auto operator()(cuda::compute_capability) const -> ::cub::reduce_policy
```

cub/benchmarks/bench/reduce/nondeterministic.cu
```diff
@@ -15,7 +15,7 @@
 #if !TUNE_BASE
 struct policy_selector
 {
-  _CCCL_API constexpr auto operator()(cuda::arch_id) const -> ::cub::reduce_policy
+  _CCCL_API constexpr auto operator()(cuda::compute_capability) const -> ::cub::reduce_policy
```
miscco: Nitpick:
```suggestion
  [[nodiscard]] _CCCL_API constexpr auto operator()(cuda::compute_capability) const -> ::cub::reduce_policy
```

cub/benchmarks/bench/scan/exclusive/by_key.cu
```diff
@@ -18,7 +18,7 @@
 #if !TUNE_BASE
 struct bench_scan_by_key_policy_selector
 {
-  [[nodiscard]] _CCCL_API constexpr auto operator()(::cuda::arch_id /*arch*/) const
+  [[nodiscard]] _CCCL_API constexpr auto operator()(::cuda::compute_capability) const
```
miscco:
```suggestion
  [[nodiscard]] _CCCL_API constexpr auto operator()(cuda::compute_capability) const
```

cub/benchmarks/bench/run_length_encode/non_trivial_runs.cu
```diff
@@ -18,7 +18,7 @@
 #if !TUNE_BASE
 struct bench_rle_policy_selector
 {
-  [[nodiscard]] constexpr auto operator()(::cuda::arch_id /*arch*/) const
+  [[nodiscard]] constexpr auto operator()(::cuda::compute_capability) const
```
miscco:
```suggestion
  [[nodiscard]] constexpr auto operator()(cuda::compute_capability) const
```

miscco (approved):

bernhardmgruber (approved):

cub/cub/device/dispatch/tuning/tuning_histogram.cuh
davebayer: I don't have a strong opinion :) I've fixed it everywhere

davebayer:
pre-commit.ci autofix

cub/cub/detail/launcher/cuda_driver.cuh
```diff
@@ -84,20 +83,20 @@ struct CudaDriverLauncherFactory
 
   ::cudaError_t PtxVersion(int& version) const
   {
-    version = cc * 10;
+    version = cc_ * 10;
     return cudaSuccess;
   }
```
bernhardmgruber, line 88: Suggestion: that function should become obsolete by this PR and can be removed (in a follow-up PR)

cub/test/catch2_test_arch_dispatch.cu
```diff
@@ -15,11 +15,11 @@
 #  define CUDA_SM_LIST_SCALE 10
 #endif
 
-using cuda::arch_id;
+using cuda::compute_capability;
 
 struct a_policy
 {
-  arch_id value;
+  int value;
```
bernhardmgruber, line 22: Suggestion: Please add a comment to turn the `int` into a `cuda::compute_capability` once #8730 is fixed. Can we done after this PR. Or fix it right way when solving #8730.

cub/test/catch2_test_arch_dispatch.cu
```diff
@@ -34,23 +34,24 @@ struct a_policy
...
   {
-    return a_policy{id};
+    return a_policy{cc.get()};
   }
 };
 
 #ifdef CUDA_SM_LIST
-template <arch_id SelectedPolicyArch, int... ArchList>
+template <int SelectedPolicyCC, int... ArchList>
 void check_arch_is_in_list()
```
bernhardmgruber, line 45: Suggestion: also transition the names used in this file, like:

```suggestion
void check_cc_is_in_list()
```
Can we done in a follow-up.

Also consider renaming the file. Also consider renaming `arch_dispatch.cuh`.

bernhardmgruber (approved):

cub/test/catch2_test_arch_dispatch.cu
davebayer: Implemented by https://github.com/NVIDIA/cccl/pull/8860

PR #8732, opened 2026-04-29
Add launch ID parameterization to segmented scan multi-segment test

### LINKED ISSUE #8717: Ensure segmented scan multi-segment implementation can be called from device code or be graph captured
Q: I don't see launch id parameterization (`// %PARAM% TEST_LAUNCH lid 0:1:2`). Does the new segmented implementation not support being called from device code or being graph captured?

_Originally posted by @bernhardmgruber in https://github.com/NVIDIA/cccl/pull/8639#discussion_r3155006958_

oleksandr-pavlyk:
## Description

closes #8717 

Add `// %PARAM% TEST_LAUNCH lid 0:1:2` and required include.

## Checklist

### DIFF
--- a/cub/test/catch2_test_device_segmented_scan_multi_segment.cu
+++ b/cub/test/catch2_test_device_segmented_scan_multi_segment.cu
@@ -17,7 +17,7 @@
 #include <vector>
 
 #include "catch2_test_device_scan.cuh"
-#include <c2h/catch2_test_helper.h>
+#include "catch2_test_launch_helper.h"
 
 // %PARAM% TEST_LAUNCH lid 0:1:2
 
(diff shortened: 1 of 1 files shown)


bernhardmgruber (changes requested):
Critical: compiling the test multiple times with different launch IDs only has an effect when the CUB APIs are wrapped behind launch wrappers, that invoke the CUB API in different scenarios (host, device, CUDA graph). Look for `DECLARE_LAUNCH_WRAPPER` in other tests.

Looking at the source, I am also worried that the unit tests don't call the public CUB API anywhere. I think this should probably be fixed first.

oleksandr-pavlyk:
> Looking at the source, I am also worried that the unit tests don't call the public CUB API anywhere. I think this should probably be fixed first.

This particular test file calls into dispatch to exercise `max_segments > 1` code path. Public CUB API uses `max_segments = 1`.

If, because of this, this file should not parametrize over launch ID we could close this PR, otherwise I intend to add use of `DECLARE_LAUNCH_WRAPPER`

bernhardmgruber (approved):

bernhardmgruber:
> This particular test file calls into dispatch to exercise `max_segments > 1` code path. Public CUB API uses `max_segments = 1`.

I assume we are going to add support calling the dispatch function with `max_segments > 1` with any public API at some point, so let's merge this.

PR #8733, opened 2026-04-29
Allows moving the token returned from `BlockLoadToShared`. 

elstehle:
## Description

Allows moving the token returned from `BlockLoadToShared`. 

Thanks to @pauleonix, who originally ran into and made the proposal.

### DIFF
--- a/cub/cub/block/block_load_to_shared.cuh
+++ b/cub/cub/block/block_load_to_shared.cuh
@@ -241,6 +241,10 @@ private:
     token_impl(const token_impl&)            = delete;
     token_impl& operator=(const token_impl&) = delete;
     // NOLINTEND(modernize-use-equals-delete)
+
+  public:
+    token_impl(token_impl&&)            = default;
+    token_impl& operator=(token_impl&&) = default;
   };
 
 public:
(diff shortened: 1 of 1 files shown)


bernhardmgruber (approved):

oleksandr-pavlyk:
Suggestion: perhaps add a test where such move is necessary.

**Update**: we already have such a test in `kernel_dyn_smem_dst` (see [catch2_test_block_load_to_shared.cu](https://github.com/NVIDIA/cccl/blob/main/cub/test/catch2_test_block_load_to_shared.cu#L147-L148)):

```
  // also test separate Commit and Wait calls with token passing here
  auto token = block_load2sh.Commit();
  block_load2sh.Wait(::cuda::std::move(token));
```

but it seems contrived, and likely optimized out by compiler. I was wondering if there is a more meaningful example/test where there is a some work done in between commit and wait steps.

pauleonix (approved):
LGTM

pauleonix:
The token has no data in it, so it should be completely optimized out in any case. The point is just to have the interface enforce ordering of calls to member functions by the user. When I ran into this problem it was code that only used `BlockLoadToShared` conditionally (tuning), so it needed to be moved out of a conditional block (into a `cuda::std::optional`).

oleksandr-pavlyk (approved):

cub/cub/block/block_load_to_shared.cuh
```diff
@@ -241,6 +241,10 @@ private:
     token_impl(const token_impl&)            = delete;
     token_impl& operator=(const token_impl&) = delete;
     // NOLINTEND(modernize-use-equals-delete)
+
+  public:
```
miscco: nitpick: move the deletion to the public section

miscco (approved):

PR #8735, opened 2026-04-29
Implement parallel cuda::std::lexicographical_compare

### LINKED ISSUE #7752: [FEA]: Implement CUDA backend for parallel `cuda::std::lexicographical_compare`
### Is this a duplicate?

### Area

libcu++

We want to implement a CUDA backend for the following algorithms:

* `std::lexicoraphical_compare` see https://en.cppreference.com/w/cpp/algorithm/lexicographical_compare.html

Hopefully similar to mismatch

_No response_

### Additional context

_No response_

edenfunf:
Resolves #7752. Implements the CUDA backend for `cuda::std::lexicographical_compare`.

## Design

Following the suggestion on #7752: a `cuda::zip_transform_iterator` produces an
`int8_t` per pair (`-1` LESS, `+1` GREATER, `0` EQUIV), `cub::DeviceFind::FindIf`
finds the first non-zero index with early termination, then a 1-element
`cub::DeviceReduce::TransformReduce` (with `cuda::std::identity{}` + `plus<int8_t>{}`
over `int8_t{0}`) reads the state at that position back to host. Final answer is
`state < 0`.

Spec complexity bound (≤ `2 * min(N1, N2)` `comp` calls) is met: Pass 1 short-circuits
at `2k` worst-case for early-termination index `k`; Pass 2 adds 1; total ≤ `2 * min - 1`.

I prototyped four value-recovery variants locally (1-element FindIf, 1-element
TransformReduce, custom 1-thread kernel, two-FindIf baseline). TransformReduce-identity
won consistently by 5–30% in p50 because cub's 1-element reduce path has lower launch
overhead than 1-element FindIf in this regime. Happy to switch shape if you'd prefer
something different.

## Performance (RTX 5070 / sm_120, CUDA 12.9)

Realistic workload — two independent random sequences (≈always diverges at k=0):

| N    | T   | this PR | full-scan reduce baseline |
|----:|----:|--------:|--------------------------:|
| 16M  | int32 | 103 µs |   274 µs |
| 64M  | int32 | 100 µs |   930 µs |
| 64M  | int64 | 110 µs |  1755 µs |

Worst case for the early-termination design (late divergence and equal-all at
scale): within 5–10% of the no-early-termination baseline.

## Tests

- `pstl_lexicographical_compare.cu` (default `less`) — 480 assertions
- `pstl_lexicographical_compare_comp.cu` (custom `Compare`) — 192 assertions

Both cover the `all_types` matrix × 4 SECTIONs (default stream / provided stream /
provided memory_resource / both), plus edge cases: empty vs empty, empty/non-empty
mixed, single-element ranges (equal/less/greater — exercises the `n=1` path through
both passes), prefix in either direction, divergence at `k=0/N/2/N-1`,
`random_access_iterator` wrapper, and `greater<>`/`less<>` duality
(`lex_cmp(s2, s1, greater) == lex_cmp(s1, s2, less)`).

`bench/lexicographical_compare/basic.cu` mirrors `is_heap_until/basic.cu`'s shape
with `2 * violation_point` memory accounting and `ViolationAt ∈ {1.0, 0.5, 0.01}`.

### DIFF
--- a/libcudacxx/include/cuda/std/__pstl/cuda/lexicographical_compare.h
+++ b/libcudacxx/include/cuda/std/__pstl/cuda/lexicographical_compare.h
@@ -1,0 +1,306 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of libcu++, the C++ Standard Library for your entire system,
+// under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef _CUDA_STD___PSTL_CUDA_LEXICOGRAPHICAL_COMPARE_H
+#define _CUDA_STD___PSTL_CUDA_LEXICOGRAPHICAL_COMPARE_H
+
+#include <cuda/std/detail/__config>
+
+#if defined(_CCCL_IMPLICIT_SYSTEM_HEADER_GCC)
+#  pragma GCC system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_CLANG)
+#  pragma clang system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_MSVC)
+#  pragma system_header
+#endif // no system header
+
+#if _CCCL_HAS_BACKEND_CUDA()
+
+_CCCL_DIAG_PUSH
+_CCCL_DIAG_SUPPRESS_CLANG("-Wshadow")
+_CCCL_DIAG_SUPPRESS_CLANG("-Wunused-local-typedef")
+_CCCL_DIAG_SUPPRESS_GCC("-Wattributes")
+_CCCL_DIAG_SUPPRESS_NVHPC(attribute_requires_external_linkage)
+
+#  include <cub/device/device_find.cuh>
+#  include <cub/device/device_reduce.cuh>
+
+_CCCL_DIAG_POP
+
+#  include <cuda/__execution/policy.h>
+#  include <cuda/__functional/call_or.h>
+#  include <cuda/__iterator/zip_transform_iterator.h>
+#  include <cuda/__stream/get_stream.h>
+#  include <cuda/__stream/stream_ref.h>
+#  include <cuda/std/__algorithm/lexicographical_compare.h>
+#  include <cuda/std/__algorithm/max.h>
+#  include <cuda/std/__algorithm/min.h>
+#  include <cuda/std/__exception/cuda_error.h>
+#  include <cuda/std/__exception/exception_macros.h>
+#  include <cuda/std/__execution/env.h>
+#  include <cuda/std/__execution/policy.h>
+#  include <cuda/std/__functional/operations.h>
+#  include <cuda/std/__iterator/concepts.h>
+#  include <cuda/std/__iterator/distance.h>
+#  include <cuda/std/__iterator/iterator_traits.h>
+#  include <cuda/std/__memory/addressof.h>
+#  include <cuda/std/__pstl/cuda/temporary_storage.h>
+#  include <cuda/std/__pstl/dispatch.h>
+#  include <cuda/std/__type_traits/always_false.h>
+#  include <cuda/std/__type_traits/common_type.h>
+#  include <cuda/std/__type_traits/remove_cvref.h>
+#  include <cuda/std/__utility/move.h>
+
+#  include <cuda/std/__cccl/prologue.h>
+
+_CCCL_BEGIN_NAMESPACE_CUDA_STD_EXECUTION
+
+// Tri-valued result of comparing a pair (a, b) drawn from the two input ranges
+// under `comp`. The underlying values are chosen so that a `0` accumulator with
+// `plus` recovers the sign directly: a negative result means `a` orders before `b`.
+enum class __lex_ordering : int
+{
+  __less    = -1,
+  __equal   = 0,
+  __greater = 1,
+};
+
+// Maps a pair (a, b) to its `__lex_ordering`. Used as the transform of a
+// `cuda::zip_transform_iterator`, so the per-element state is computed lazily by
+// the device kernels that walk the iterator.
+template <class _Compare>
+struct __lex_state_fn
+{
+  _Compare __comp_;
+
+  _CCCL_API explicit constexpr __lex_state_fn(_Compare __comp)
+      : __comp_(::cuda::std::move(__comp))
+  {}
+
+  template <class _Tp, class _Up>
+  [[nodiscard]] _CCCL_DEVICE_API constexpr __lex_ordering operator()(const _Tp& __a, const _Up& __b) const
+  {
+    if (__comp_(__a, __b))
+    {
+      return __lex_ordering::__less;
+    }
+    if (__comp_(__b, __a))
+    {
+      return __lex_ordering::__greater;
+    }
+    return __lex_ordering::__equal;
+  }
+};
+
+// Predicate handed to the find pass: locates the first non-equivalent pair.
+struct __lex_non_equal
+{
+  [[nodiscard]] _CCCL_DEVICE_API constexpr bool operator()(__lex_ordering __state) const noexcept
+  {
+    return __state != __lex_ordering::__equal;
+  }
+};
+
+// Transform handed to the read-back pass: surfaces the ordering as its underlying
+// signed value so the 1-element reduction returns it unchanged.
+struct __lex_ordering_to_int
+{
+  [[nodiscard]] _CCCL_DEVICE_API constexpr int operator()(__lex_ordering __state) const noexcept
+  {
+    return static_cast<int>(__state);
+  }
+};
+
+_CCCL_BEGIN_NAMESPACE_ARCH_DEPENDENT
+
+template <>
+struct __pstl_dispatch<__pstl_algorithm::__lexicographical_compare, __execution_backend::__cuda>
+{
+  template <class _Policy, class _InputIter1, class _InputIter2, class _Compare>
+  [[nodiscard]] _CCCL_HOST_API static bool __par_impl(
+    const _Policy& __policy,
+    _InputIter1 __first1,
+    _InputIter1 __last1,
+    _InputIter2 __first2,
+    _InputIter2 __last2,
+    _Compare __comp)
+  {
+    if (__first1 == __last1)
+    {
+      return __first2 != __last2;
+    }
+    if (__first2 == __last2)
+    {
+      return false;
+    }
+
+    using _OffsetType = common_type_t<iter_difference_t<_InputIter1>, iter_difference_t<_InputIter2>>;
+    const auto __n1   = static_cast<_OffsetType>(::cuda::std::distance(__first1, __last1));
+    const auto __n2   = static_cast<_OffsetType>(::cuda::std::distance(__first2, __last2));
+    const auto __n    = ::cuda::std::min(__n1, __n2);
+
+    // Lazy per-element ordering over the two input ranges, capped at the common
+    // length so neither side reads past its own end.
+    auto __state_first = ::cuda::zip_transform_iterator{
+      __lex_state_fn<_Compare>{::cuda::std::move(__comp)}, ::cuda::std::move(__first1), ::cuda::std::move(__first2)};
+
+    constexpr __lex_ordering_to_int __to_int{};
+    constexpr ::cuda::std::plus<int> __plus{};
+
+    // Size the temporary storage for both passes up front and allocate it once.
+    // The find pass over `__n` elements dominates the 1-element read-back, so the
+    // shared region is reused by both cub calls; only the two 1-element result
+    // slots are extra.
+    size_t __find_bytes = 0;
+    _CCCL_TRY_CUDA_API(
+      CUB_NS_QUALIFIER::DeviceFind::FindIf,
+      "__pstl_cuda_lexicographical_compare: determining temporary storage for cub::DeviceFind::FindIf failed",
+      static_cast<void*>(nullptr),
+      __find_bytes,
+      __state_first,
+      static_cast<_OffsetType*>(nullptr),
+      __lex_non_equal{},
+      __n);
+
+    size_t __reduce_bytes = 0;
+    _CCCL_TRY_CUDA_API(
+      CUB_NS_QUALIFIER::DeviceReduce::TransformReduce,
+      "__pstl_cuda_lexicographical_compare: determining temporary storage for cub::DeviceReduce::TransformReduce "
+      "failed",
+      static_cast<void*>(nullptr),
+      __reduce_bytes,
+      __state_first,
+      static_cast<int*>(nullptr),
+      _OffsetType{1},
+      __plus,
+      __to_int,
+      int{0});
+
+    const size_t __temp_bytes = ::cuda::std::max(__find_bytes, __reduce_bytes);
+
+    auto __stream = ::cuda::__call_or(::cuda::get_stream, ::cuda::stream_ref{cudaStream_t{}}, __policy);
+
+    // Single allocation: slot<0> = find offset, slot<1> = read-back state, plus the
+    // shared algorithm scratch region sized for whichever pass needs more.
+    __temporary_storage<_OffsetType, int> __storage{__policy, __temp_bytes, 1, 1};
+
+    // Pass 1 (early-terminating): index of the first non-equivalent pair.
+    _OffsetType __k;
+    _CCCL_TRY_CUDA_API(
+      CUB_NS_QUALIFIER::DeviceFind::FindIf,
+      "__pstl_cuda_lexicographical_compare: kernel launch of cub::DeviceFind::FindIf failed",
+      __storage.__get_temp_storage(),
+      __find_bytes,
+      __state_first,
+      __storage.template __get_ptr<0>(),
+      __lex_non_equal{},
+      __n,
+      __stream.get());
+    _CCCL_TRY_CUDA_API(
+      ::cudaMemcpyAsync,
+      "__pstl_cuda_lexicographical_compare: copy of find result from device to host failed",
+      ::cuda::std::addressof(__k),
+      __storage.template __get_ptr<0>(),
+      sizeof(_OffsetType),
+      ::cudaMemcpyDefault,
+      __stream.get());
+    __stream.sync();
+
+    // No divergence within the common prefix: the shorter range is "less".
+    if (__k == __n)
+    {
+      return __n1 < __n2;
+    }
+
+    // Pass 2: read back the ordering at the divergence position via a 1-element
+    // transform_reduce, reusing the same scratch region as the find pass.
+    int __state;
+    _CCCL_TRY_CUDA_API(
+      CUB_NS_QUALIFIER::DeviceReduce::TransformReduce,
+      "__pstl_cuda_lexicographical_compare: kernel launch of cub::DeviceReduce::TransformReduce failed",
+      __storage.__get_temp_storage(),
+      __reduce_bytes,
+      __state_first + __k,
+      __storage.template __get_ptr<1>(),
+      _OffsetType{1},
+      __plus,
+      __to_int,
+      int{0},
+      __stream.get());
+    _CCCL_TRY_CUDA_API(
+      ::cudaMemcpyAsync,
+      "__pstl_cuda_lexicographical_compare: copy of read-back result from device to host failed",
+      ::cuda::std::addressof(__state),
+      __storage.template __get_ptr<1>(),
+      sizeof(int),
+      ::cudaMemcpyDefault,
+      __stream.get());
+    __stream.sync();
+
+    return __state < 0;
+  }
+
+  _CCCL_TEMPLATE(class _Policy, class _InputIter1, class _InputIter2, class _Compare)
+  _CCCL_REQUIRES(__has_forward_traversal<_InputIter1> _CCCL_AND __has_forward_traversal<_InputIter2>)
+  [[nodiscard]] _CCCL_HOST_API bool operator()(
+    [[maybe_unused]] const _Policy& __policy,
+    _InputIter1 __first1,
+    _InputIter1 __last1,
+    _InputIter2 __first2,
+    _InputIter2 __last2,
+    _Compare __comp) const
+  {
+    if constexpr (__has_random_access_traversal<_InputIter1> && __has_random_access_traversal<_InputIter2>)
+    {
+      try
+      {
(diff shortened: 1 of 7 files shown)


libcudacxx/include/cuda/std/__pstl/lexicographical_compare.h
```diff
@@ -0,0 +1,173 @@
...
+struct __lex_state_fn
+{
+  _Compare __comp_;
+
+  _CCCL_API explicit constexpr __lex_state_fn(_Compare __comp)
+      : __comp_(::cuda::std::move(__comp))
+  {}
+
+  template <class _Tp, class _Up>
+  [[nodiscard]] _CCCL_DEVICE_API constexpr int8_t operator()(const _Tp& __a, const _Up& __b) const
```
miscco: @fbusato I believe using plain `int` might be more efficient because we do not need to unpack the `int8_t` thoughts?

edenfunf:
Tested locally on RTX 5070 / sm_120, CUDA 12.9, with both variants
implemented identically except for the state type. Per-call wall-clock
p50 (5 warmup + 50 reps via cudaEvent pairs); ratio = `int` / `int8_t`
so values < 1.0 mean `int` wins.

| N    | T     | workload      | int8_t p50 | int p50 | ratio |
|----:|----:|---------------|-----------:|--------:|------:|
| 64M  | int32 | diverge@k=10  | 101.3 µs   |  79.6 µs | 0.785 |
| 16M  | int64 | diverge@k=10  | 120.6 µs   |  78.3 µs | 0.649 |
| 64M  | int64 | diverge@k=10  | 144.6 µs   | 117.3 µs | 0.811 |
| 16M  | int32 | diverge@k=N/2 | 242.4 µs   | 225.0 µs | 0.928 |
| 64M  | int32 | rand-all-diff |  77.4 µs   |  78.2 µs | 1.010 |
| 64M  | int32 | equal-all     | 916.7 µs   | 928.4 µs | 1.013 |

Honest disclosure: small-N + equal-all (where Pass 2 is skipped because
`k == n`) shows `int8_t` slightly faster by ~10 µs absolute (e.g., N=1M
i32 equal-all: int8_t 44 µs vs int 55 µs). That regime is launch-bound
either way, and the gain on the hot path (early-divergence + large N,
where Pass 2's 1-element TransformReduce dominates) is the meaningful
trade.

Pushed the swap as 15c236d6e3.

edenfunf:
Hi @miscco — gentle ping when you have a moment, I think this is ready for another look.

The int8_t → int suggestion is done (15c236d6), with the before/after numbers in the table above. I also re-checked locally on a 5070 (CUDA 12.9): both tests pass, plus a 50M cross-check against std::lexicographical_compare across int/int64/uint8/float/double and the edge cases, and memcheck came back clean.

Happy to make any changes — please let me know if anything's still blocking. Thanks!

miscco:
/ok to test [15c236d](https://github.com/NVIDIA/cccl/pull/8735/commits/15c236d6e320e688a48e3e05fe20ac0e2eff0eeb)

libcudacxx/include/cuda/std/__pstl/lexicographical_compare.h
```diff
@@ -0,0 +1,172 @@
...
+  _CCCL_API explicit constexpr __lex_state_fn(_Compare __comp)
+      : __comp_(::cuda::std::move(__comp))
+  {}
+
+  template <class _Tp, class _Up>
+  [[nodiscard]] _CCCL_DEVICE_API constexpr int operator()(const _Tp& __a, const _Up& __b) const
+  {
+    if (__comp_(__a, __b))
+    {
+      return -1;
```
miscco: Important: We avoid magical numbers as much as possible. Please create an enum class with members `__less`, `__equal`, `__greater`

libcudacxx/include/cuda/std/__pstl/lexicographical_compare.h
```diff
@@ -0,0 +1,172 @@
...
+  _Compare __comp = {})
+{
+  static_assert(indirect_strict_weak_order<_Compare, _InputIter1, _InputIter2>,
+                "cuda::std::lexicographical_compare: Compare must satisfy "
+                "indirect_strict_weak_order<Compare, InputIter1, InputIter2>");
+
+  [[maybe_unused]] auto __find_dispatch =
+    ::cuda::std::execution::__pstl_select_dispatch<::cuda::std::execution::__pstl_algorithm::__find_if, _Policy>();
+  [[maybe_unused]] auto __reduce_dispatch =
+    ::cuda::std::execution::__pstl_select_dispatch<::cuda::std::execution::__pstl_algorithm::__transform_reduce,
```
miscco: The current approach will allocate twice, because we need temporary storage for find_if and also reduce. 

However, we can coalesce both allocations into one because we know how much storage we need. We can also reuse the temp storage allocation for both reduce and find and only have one additional int sized one for the return value.

libcudacxx/include/cuda/std/__pstl/lexicographical_compare.h
```diff
@@ -0,0 +1,172 @@
...
+    const auto __n  = ::cuda::std::min(__n1, __n2);
+
+    // Lazy per-element state (-1 / 0 / +1) over the two input ranges, capped at
+    // the common length so neither side reads past its own end.
+    auto __state_first = ::cuda::zip_transform_iterator{
+      __lex_state_fn<_Compare>{::cuda::std::move(__comp)}, ::cuda::std::move(__first1), ::cuda::std::move(__first2)};
+    auto __state_last = __state_first + __n;
+
+    // Pass 1 (early-terminating): index of the first non-equivalent pair.
+    auto __k_iter = __find_dispatch(__policy, ::cuda::std::move(__state_first), __state_last, __lex_non_zero{});
```
miscco: We will need to implement this for cuda to coalesce allocations

miscco (commented):
Thanks a lot for the work. I believe we want to coalesce the two allocations we currently perform into a single one

libcudacxx/include/cuda/std/__pstl/lexicographical_compare.h
edenfunf: Done in `8b90fc46` — added an `__lex_ordering` enum class (`__less`/`__equal`/`__greater`). Kept the underlying -1/0/1 so the 1-element read-back recovers the sign with a plain `plus<int>`.

libcudacxx/include/cuda/std/__pstl/lexicographical_compare.h
edenfunf: Done. One `__temporary_storage<_OffsetType, int>` now: slot 0 = find offset, slot 1 = read-back state, shared scratch sized `max(find_bytes, reduce_bytes)`. Both passes run against the same `__get_temp_storage()`, so it is a single allocation instead of two.

libcudacxx/include/cuda/std/__pstl/lexicographical_compare.h
edenfunf: Yep — composing the generic find_if/transform_reduce dispatches cannot share scratch, so I added a dedicated `__pstl/cuda/lexicographical_compare.h` backend and switched the frontend to select it. Serial fallback unchanged.

edenfunf:
@miscco thanks for the review! All three points are addressed in `8b90fc46`:

- magic numbers are now an `__lex_ordering` enum class
- the CUDA path moved into a dedicated `__pstl/cuda/lexicographical_compare.h` backend that only allocates once — both passes share a single `__temporary_storage`, sized to the larger of the two

Re-ran everything locally (RTX 5070, CUDA 12.9): tests pass and memcheck is clean. Let me know if anything else needs changing.

libcudacxx/include/cuda/std/__pstl/cuda/lexicographical_compare.h
```diff
@@ -0,0 +1,306 @@
...
+
+// Maps a pair (a, b) to its `__lex_ordering`. Used as the transform of a
+// `cuda::zip_transform_iterator`, so the per-element state is computed lazily by
+// the device kernels that walk the iterator.
+template <class _Compare>
+struct __lex_state_fn
+{
+  _Compare __comp_;
+
+  _CCCL_API explicit constexpr __lex_state_fn(_Compare __comp)
```
miscco:
```suggestion
  _CCCL_API explicit constexpr __lex_state_fn(_Compare __comp) noexcept(is_nothrow_move_constructible_v<_Compare>)
```

libcudacxx/include/cuda/std/__pstl/cuda/lexicographical_compare.h
```diff
@@ -0,0 +1,306 @@
...
+    if (__comp_(__a, __b))
+    {
+      return __lex_ordering::__less;
+    }
+    if (__comp_(__b, __a))
+    {
+      return __lex_ordering::__greater;
+    }
+    return __lex_ordering::__equal;
+  }
```
miscco, line 112: Nitpick: There might be pathological cases where `_Compare` has a non-const call operator

It would be safer to also have a non-const call operator

libcudacxx/include/cuda/std/__pstl/cuda/lexicographical_compare.h
```diff
@@ -0,0 +1,306 @@
...
+    }
+    if (__first2 == __last2)
+    {
+      return false;
+    }
+
+    using _OffsetType = common_type_t<iter_difference_t<_InputIter1>, iter_difference_t<_InputIter2>>;
+    const auto __n1   = static_cast<_OffsetType>(::cuda::std::distance(__first1, __last1));
+    const auto __n2   = static_cast<_OffsetType>(::cuda::std::distance(__first2, __last2));
+    const auto __n    = ::cuda::std::min(__n1, __n2);
```
miscco: Nitpick: We prefer more descriptive variable names like 
```suggestion
    const auto __count1   = static_cast<_OffsetType>(::cuda::std::distance(__first1, __last1));
    const auto __count2   = static_cast<_OffsetType>(::cuda::std::distance(__first2, __last2));
    const auto __count     = ::cuda::std::min(__count1, __count2);
```

libcudacxx/include/cuda/std/__pstl/cuda/lexicographical_compare.h
```diff
@@ -0,0 +1,306 @@
...
+      __stream.get());
+    _CCCL_TRY_CUDA_API(
+      ::cudaMemcpyAsync,
+      "__pstl_cuda_lexicographical_compare: copy of find result from device to host failed",
+      ::cuda::std::addressof(__k),
+      __storage.template __get_ptr<0>(),
+      sizeof(_OffsetType),
+      ::cudaMemcpyDefault,
+      __stream.get());
+    __stream.sync();
```
miscco: Suggestion: I believe we can actually optimize further here and completely drop reduce.

Rather than copying _OffsetT back to host to launch a reduce which copies it back to device we can use a custom kernel that does

```c++
_CCCL_KERNEL_ATTRIBUTES void __lex_result(...) {
    // No divergence within the common prefix: the shorter range is "less".
    if (__offset == __count)
    {
      *__offset = static_cast<_OffsetT>(__count1 < __count2);
    }
    else 
    {
      *__offset = static_cast<_OffsetT>(*(__state_first + __offset) == __lex_ordering::__less);
    }
}
```

Then we only need to copy that result back, likely saving one `cudaMemcypAsync` call

miscco (approved):

miscco (changes requested):
Sorry approved accidentally

libcudacxx/include/cuda/std/__pstl/cuda/lexicographical_compare.h
edenfunf: Applied, thanks.

libcudacxx/include/cuda/std/__pstl/cuda/lexicographical_compare.h
edenfunf: Added a non-const `operator()` overload alongside the const one.

libcudacxx/include/cuda/std/__pstl/cuda/lexicographical_compare.h
edenfunf: Done, renamed to `__count1`/`__count2`/`__count`.

libcudacxx/include/cuda/std/__pstl/cuda/lexicographical_compare.h
edenfunf: Dropped the reduce — a single-thread kernel now resolves the result in place from the find index, so it is just one memcpy back instead of a second launch + copy. Pushed in `9fa18810`.

edenfunf:
No worries! Another look whenever you have time would be great. Thanks again!

libcudacxx/include/cuda/std/__pstl/cuda/lexicographical_compare.h
```diff
@@ -108,15 +120,20 @@ struct __lex_non_equal
...
+//   - otherwise the ordering at the divergence position decides.
+template <class _StateIter, class _OffsetType>
+_CCCL_KERNEL_ATTRIBUTES void __lexicographical_compare_result_kernel(
+  _StateIter __state_first, _OffsetType __count, bool __shorter_is_less, _OffsetType* __result)
 {
-  [[nodiscard]] _CCCL_DEVICE_API constexpr int operator()(__lex_ordering __state) const noexcept
-  {
-    return static_cast<int>(__state);
-  }
-};
```
miscco: Critical: This is a kernel, so it needs to live inside the `_CCCL_BEGIN_NAMESPACE_ARCH_DEPENDENT` namespace

libcudacxx/include/cuda/std/__pstl/cuda/lexicographical_compare.h
```diff
@@ -0,0 +1,292 @@
...
+//   - if no divergence was found within the common prefix, the shorter range is
+//     "less" (`__shorter_is_less`);
+//   - otherwise the ordering at the divergence position decides.
+template <class _StateIter, class _OffsetType>
+_CCCL_KERNEL_ATTRIBUTES void __lexicographical_compare_result_kernel(
+  _StateIter __state_first, _OffsetType __count, bool __shorter_is_less, _OffsetType* __result)
+{
+  const _OffsetType __k = *__result;
+  *__result =
+    static_cast<_OffsetType>((__k == __count) ? __shorter_is_less : (*(__state_first + __k) == __lex_ordering::__less));
```
miscco: Lets use slightly more descriptive names
```suggestion
  const _OffsetType __offset = *__result;
  *__result =
    static_cast<_OffsetType>((__offset == __count) ? __shorter_is_less : (*(__state_first + __offset ) == __lex_ordering::__less));
```

libcudacxx/include/cuda/std/__pstl/cuda/lexicographical_compare.h
davebayer: Do we need the `const` one now for anything?

libcudacxx/include/cuda/std/__pstl/cuda/lexicographical_compare.h
```diff
@@ -0,0 +1,292 @@
...
+    // Pass 2: resolve the answer in place on device, so only one value is copied
+    // back to the host (no second reduce launch, no offset round-trip).
+    const bool __shorter_is_less = (__count1 < __count2);
+    const void* __kernel =
+      reinterpret_cast<const void*>(&__lexicographical_compare_result_kernel<decltype(__state_first), _OffsetType>);
+    void* __kernel_args[] = {
+      const_cast<void*>(reinterpret_cast<const void*>(::cuda::std::addressof(__state_first))),
+      const_cast<void*>(reinterpret_cast<const void*>(::cuda::std::addressof(__count))),
+      const_cast<void*>(reinterpret_cast<const void*>(::cuda::std::addressof(__shorter_is_less))),
+      const_cast<void*>(reinterpret_cast<const void*>(::cuda::std::addressof(__result_ptr)))};
```
miscco: I believe we should be able to just `static_cast<void*>` and not require reinterpret_cast
```suggestion
      static_cast<void*>(::cuda::std::addressof(__state_first)),
      static_cast<void*>(::cuda::std::addressof(__count)),
      static_cast<void*>(::cuda::std::addressof(__shorter_is_less)),
      static_cast<void*>(::cuda::std::addressof(__result_ptr)));
```

libcudacxx/include/cuda/std/__pstl/cuda/lexicographical_compare.h
```diff
@@ -0,0 +1,292 @@
...
+      {
+        if (__err.status() == cudaErrorMemoryAllocation)
+        {
+          _CCCL_THROW(::std::bad_alloc);
+        }
+        else
+        {
+          throw __err;
+        }
+      }
```
davebayer, line 272:
```suggestion
      _CCCL_TRY
      {
        return __par_impl(
          __policy,
          ::cuda::std::move(__first1),
          ::cuda::std::move(__last1),
          ::cuda::std::move(__first2),
          ::cuda::std::move(__last2),
          ::cuda::std::move(__comp));
      }
      _CCCL_CATCH (const ::cuda::cuda_error& __err)
      {
        if (__err.status() == ::cudaErrorMemoryAllocation)
        {
          _CCCL_THROW(::std::bad_alloc);
        }
        else
        {
          _CCCL_RETHROW;
        }
      }
```

libcudacxx/include/cuda/std/__pstl/cuda/lexicographical_compare.h
```diff
@@ -0,0 +1,292 @@
...
+    using _OffsetType   = common_type_t<iter_difference_t<_InputIter1>, iter_difference_t<_InputIter2>>;
+    const auto __count1 = static_cast<_OffsetType>(::cuda::std::distance(__first1, __last1));
+    const auto __count2 = static_cast<_OffsetType>(::cuda::std::distance(__first2, __last2));
+    const auto __count  = ::cuda::std::min(__count1, __count2);
+
+    // Lazy per-element ordering over the two input ranges, capped at the common
+    // length so neither side reads past its own end.
+    auto __state_first = ::cuda::zip_transform_iterator{
+      __lex_state_fn<_Compare>{::cuda::std::move(__comp)}, ::cuda::std::move(__first1), ::cuda::std::move(__first2)};
+
```
davebayer, line 171: We need to set the context of the stream by constructing an `__ensure_current_context` object here, that will set the current device for us

libcudacxx/include/cuda/std/__pstl/cuda/lexicographical_compare.h
davebayer: I would just go with C-style cast in these cases
```suggestion
    const auto __kernel = (const void*)&__lexicographical_compare_result_kernel<decltype(__state_first), _OffsetType>;
    void* __kernel_args[] = {
      (void*)::cuda::std::addressof(__state_first)),
      (void*)::cuda::std::addressof(__count)),
      (void*)::cuda::std::addressof(__shorter_is_less)),
      (void*)::cuda::std::addressof(__result_ptr)));
```

libcudacxx/include/cuda/std/__pstl/cuda/lexicographical_compare.h
miscco: We should keep it because it lets the compiler optimize better

bernhardmgruber:
Looking at the implementation from a bird’s-eye view, I was thinking whether we could avoid calling an extra kernel if we would extend `DeviceFind` with some kind of finalizer action when the right position is found. It turns out that's not so easy, because `DeviceFind` uses atomics in SMEM and GMEM to perform (integral) reductions when it starts to find indices where the predicate is true, so we would need to replace this by some other form of reduction that reduces index and value, since the blocks in the find kernel don't actually know with index will be the global first one (this is resolved in L2).

I think the current approach with using `DeviceFind` to find the mismatch position and then loading the two values again for the comparison is a bit wasteful, but doing better is a LOT harder. I think this is fine. We can keep a tracking issue around to revisit this.

miscco:
yeah I believe we could do better if we extended `FindIf` but I believe  that is out of scope of this PR

libcudacxx/include/cuda/std/__pstl/cuda/lexicographical_compare.h
edenfunf: Done, moved it inside.

libcudacxx/include/cuda/std/__pstl/cuda/lexicographical_compare.h
edenfunf: Done, renamed to `__offset`.

libcudacxx/include/cuda/std/__pstl/cuda/lexicographical_compare.h
edenfunf: Done. Dropped the reinterpret_cast/const_cast pair and made the launched locals non-const so `static_cast<void*>` works directly.

edenfunf: Went with @miscco's `static_cast<void*>` on the args. Kept `reinterpret_cast` only on the kernel function pointer, since `static_cast` can't convert a function pointer to `void*` (matches the `cudaLaunchKernel` idiom in cudax). Can switch to C-style throughout if you'd prefer.

libcudacxx/include/cuda/std/__pstl/cuda/lexicographical_compare.h
edenfunf: The named `_CCCL_CATCH(...)` expands to an unbalanced brace in the no-exception (device) pass and isn't used anywhere else in the tree; the sibling PSTL cuda backends (partition/find_if/transform_reduce) all use a plain try/catch. Kept the raw try/catch for consistency but took the rest: `::cudaErrorMemoryAllocation` and a plain `throw`. Happy to revisit if you'd prefer pushing on the macro.

libcudacxx/include/cuda/std/__pstl/cuda/lexicographical_compare.h
edenfunf: Added, right after getting the stream so the device stays current through the allocation, launches and copy-back.

edenfunf:
@bernhardmgruber @miscco agreed, extending FindIf is out of scope here. Happy to keep a tracking issue around to revisit avoiding the second load later.

miscco:
Sorry  I totally dropped the ball on this. Ping me if I did not respond on monday

edenfunf:
> Sorry I totally dropped the ball on this. Ping me if I did not respond on monday

No worries at all! Hope you have a great weekend :>

edenfunf:
@miscco When you have a chance, could you please take a look? Thanks!

miscco:
/ok to test [0f03425](https://github.com/NVIDIA/cccl/pull/8735/commits/0f03425743658748e94f9f71b92e0d1d464496ee)

miscco (approved):
Looks good to me, @davebayer any comment on cuda launch here?

PR #8740, opened 2026-04-29
`cuda::std::simd` Math

fbusato:
## Description

Implement https://eel.is/c++draft/simd#math

Notable differences compared to the standard:

- Several functions are not `constexpr` because the corresponding scalar version in `cuda/std/cmath` is also non-`constexpr`: `exp`, `log`, `sqrt`, `hypot`, `pow`, `fma`, `ceil`, `floor`, `round`, `trunc`, `lround`, `llround`, `ldexp`, `scalbn`, `scalbln`, `nextafter`, `frexp`, `remquo`, and `modf`.
- The following functions are not implemented because  the corresponding scalar version is not implemented in `cuda/std/cmath`: `assoc_laguerre`, `assoc_legendre`, `sph_legendre`, `hermite`, `laguerre`, `legendre`, `sph_bessel`, `sph_neumann`, `beta`, `comp_ellint_*`, `cyl_bessel_*`, `cyl_neumann`, `ellint_*`, `expint`, and `riemann_zeta`.


libcudacxx/include/cuda/std/__simd/math/abs.h
```diff
@@ -0,0 +1,93 @@
...
+{
+  using __result_t = typename _Result::value_type;
+
+  const _Vp& __x_;
+
+  template <typename _Ip>
+  [[nodiscard]] _CCCL_API constexpr __result_t operator()(_Ip) const noexcept
+  {
+    const auto __x = __x_[_Ip::value];
+    if constexpr (is_integral_v<__result_t>)
```
miscco: Nitpick: This could optimize for unsigned

libcudacxx/include/cuda/std/__simd/math/abs.h
```diff
@@ -0,0 +1,93 @@
...
+    {
+      _CCCL_ASSERT(__x >= -numeric_limits<__result_t>::max(),
+                   "cuda::std::simd::abs precondition: each element must be greater than the minimum value");
+      return (__x < __result_t{0}) ? static_cast<__result_t>(-__x) : __x;
+    }
+    else
+    {
+      return static_cast<__result_t>(::cuda::std::fabs(__x));
```
miscco: Important: I would be deeply troubled if there would be a cast needed here

libcudacxx/include/cuda/std/__simd/math/classification.h
```diff
@@ -0,0 +1,80 @@
...
+#include <cuda/std/__cmath/isnan.h>
+#include <cuda/std/__cmath/isnormal.h>
+#include <cuda/std/__cmath/signbit.h>
+#include <cuda/std/__simd/math/common.h>
+
+#include <cuda/std/__cccl/prologue.h>
+
+_CCCL_BEGIN_NAMESPACE_CUDA_STD_SIMD
+
+_CCCL_SIMD_MATH_UNARY_GENERATOR(fpclassify);
```
miscco, line 36: Please explore just taking a function pointer of the right overload

libcudacxx/include/cuda/std/__simd/math/common.h
```diff
@@ -0,0 +1,219 @@
...
+
+_CCCL_BEGIN_NAMESPACE_CUDA_STD_SIMD
+
+// concept simd-vec-type, exposition only
+template <typename _Tp, typename _Up = remove_cvref_t<_Tp>>
+_CCCL_CONCEPT __is_simd_vec_type_v = _CCCL_REQUIRES_EXPR((_Up))(
+  typename(typename _Up::value_type),
+  typename(typename _Up::abi_type),
+  requires(is_same_v<_Up, basic_vec<typename _Up::value_type, typename _Up::abi_type>>),
+  requires(is_default_constructible_v<_Up>));
```
miscco, line 45: Important: This seems like the wrong place for such a generic concept

fbusato: it is used only once in this file. If it will be used in more places, I will move it

libcudacxx/include/cuda/std/__simd/math/classification.h
fbusato: math APIs have overloads which could lead to bugs. Secondly, the performance could be bad. I'm not really in favor of taking these risks.

miscco: :+1: Would still love to explore this at a later date

miscco:
@fbusato This has merge conflics, can you please rebase

libcudacxx/include/cuda/std/__simd/math/abs.h
```diff
@@ -0,0 +1,100 @@
...
+  {
+    const auto __x = __x_[_Ip::value];
+    if constexpr (is_unsigned_v<__result_t>)
+    {
+      return __x;
+    }
+    else if constexpr (is_integral_v<__result_t>)
+    {
+      _CCCL_ASSERT(__x >= -numeric_limits<__result_t>::max(),
+                   "cuda::std::simd::abs precondition: each element must be greater than the minimum value");
```
miscco, line 50: I am not sure whether this is the right precondition. Shouldnt this rather be 
```suggestion
      _CCCL_ASSERT(__x > numeric_limits<__result_t>::min(),
                   "cuda::std::simd::abs precondition: each element must be greater than the minimum value");
```

libcudacxx/include/cuda/std/__simd/math/abs.h
```diff
@@ -0,0 +1,100 @@
...
+_CCCL_REQUIRES(is_integral_v<_Tp> _CCCL_AND is_signed_v<_Tp>)
+[[nodiscard]] _CCCL_HOST_DEVICE_API constexpr auto abs(const basic_vec<_Tp, _Abi>& __x) noexcept
+{
+  using __vec_t = basic_vec<_Tp, _Abi>;
+  return __vec_t{__simd_abs_generator<__vec_t, __vec_t>{__x}};
+}
+
+// floating point
+_CCCL_TEMPLATE(typename _Vp)
+_CCCL_REQUIRES(__is_math_floating_point_v<_Vp>)
```
miscco: Important: `__is_math_floating_point_v` is not a good name, because it does not relate to anything SIMD

If I read that in a codebase I would assume it relates to any T

Can we please change the name so that is makes it clear that this is a SIMD vector of floating point types?

libcudacxx/include/cuda/std/__simd/math/common.h
```diff
@@ -0,0 +1,222 @@
...
+// concept simd-floating-point, exposition only
+template <typename _Tp>
+_CCCL_CONCEPT __is_simd_floating_point_v = _CCCL_REQUIRES_EXPR(
+  (_Tp))(typename(typename _Tp::value_type),
+         requires(__is_simd_vec_type_v<_Tp>),
+         requires(::cuda::is_floating_point_v<typename _Tp::value_type>));
+
+// concept math-floating-point, exposition only
+template <typename _Tp>
+_CCCL_CONCEPT __is_math_floating_point_v = _CCCL_REQUIRES_EXPR(
```
miscco: I would really love a better name

libcudacxx/test/libcudacxx/std/numerics/simd/simd.math/classification.pass.cpp
```diff
@@ -0,0 +1,75 @@
...
+    assert(classes[i] == cuda::std::fpclassify(vec[i]));
+    assert(finite[i] == cuda::std::isfinite(vec[i]));
+    assert(inf[i] == cuda::std::isinf(vec[i]));
+    assert(nan[i] == cuda::std::isnan(vec[i]));
+    assert(normal[i] == cuda::std::isnormal(vec[i]));
+    assert(signs[i] == cuda::std::signbit(vec[i]));
+  }
+}
+
+DEFINE_SIMD_MATH_FLOATING_TEST()
```
miscco: Nitpick: We could test whether those are constexpr, but I dont really see that much value

temp.build.yml
miscco, line 1: I believe this is an accidental add and should be removed

libcudacxx/include/cuda/std/__simd/math/abs.h
fbusato: I agree with you, but the specification is https://eel.is/c++draft/simd#lib:abs,simd_
image

libcudacxx/include/cuda/std/__simd/math/abs.h
fbusato: good point. it is confusing

libcudacxx/test/libcudacxx/std/numerics/simd/simd.math/classification.pass.cpp
fbusato: it makes sense for consistency

libcudacxx/include/cuda/std/__simd/math/abs.h
miscco: fin by me

miscco (approved):

PR #8743, opened 2026-04-29
Add `<cuda/execution>` umbrella and partial headers

bernhardmgruber:


bernhardmgruber:
@miscco should I still add a test for the umbrella header itself?

libcudacxx/include/cuda/execution
```diff
@@ -0,0 +1,31 @@
...
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_MSVC)
+#  pragma system_header
+#endif // no system header
+
+#include <cuda/__execution/determinism.h>
+#include <cuda/__execution/output_ordering.h>
+#include <cuda/__execution/policy.h>
+#include <cuda/__execution/require.h>
+#include <cuda/__execution/tune.h>
+#include <cuda/std/execution>
```
miscco, line 29: This makes me wonder whether we just want `<cuda/tune>`

miscco (approved):

libcudacxx/include/cuda/execution
bernhardmgruber: This is not a bad idea. Let's wait a little, since we may want to expose the tuning policies for PSTL as well and then rediscuss where they should live etc.

PR #8744, opened 2026-04-29
Implement `ranges::zip_view`

Jacobfaib:
## Description

See #93 

## Checklist

### DIFF
--- a/libcudacxx/include/cuda/__iterator/zip_common.h
+++ b/libcudacxx/include/cuda/__iterator/zip_common.h
@@ -21,6 +21,7 @@
 #endif // no system header
 
 #include <cuda/__fwd/iterator.h>
+#include <cuda/std/__algorithm/ranges_min_element.h>
 #include <cuda/std/__concepts/constructible.h>
 #include <cuda/std/__fwd/pair.h>
 #include <cuda/std/__fwd/tuple.h>
@@ -29,11 +30,16 @@
 #include <cuda/std/__iterator/iter_swap.h>
 #include <cuda/std/__iterator/iterator_traits.h>
 #include <cuda/std/__iterator/readable_traits.h>
+#include <cuda/std/__tuple_dir/get.h>
+#include <cuda/std/__tuple_dir/tuple_element.h>
+#include <cuda/std/__tuple_dir/tuple_size.h>
 #include <cuda/std/__type_traits/is_nothrow_default_constructible.h>
 #include <cuda/std/__type_traits/is_nothrow_move_constructible.h>
+#include <cuda/std/__type_traits/remove_cvref.h>
 #include <cuda/std/__type_traits/remove_reference.h>
 #include <cuda/std/__type_traits/void_t.h>
 #include <cuda/std/__utility/declval.h>
+#include <cuda/std/__utility/integer_sequence.h>
 
 #include <cuda/std/__cccl/prologue.h>
 
@@ -103,7 +109,7 @@ struct __zip_op_star
 
   _CCCL_EXEC_CHECK_DISABLE
   template <class... _Iterators>
-  [[nodiscard]] _CCCL_API constexpr reference<_Iterators...> operator()(const _Iterators&... __iters) const
+  [[nodiscard]] _CCCL_API constexpr auto operator()(const _Iterators&... __iters) const
     noexcept(noexcept(reference<_Iterators...>{*__iters...}))
   {
     return reference<_Iterators...>{*__iters...};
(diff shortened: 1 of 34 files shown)


libcudacxx/include/cuda/std/__ranges/zip_view.h
```diff
@@ -0,0 +1,960 @@
...
+{
+  using iterator_category = input_iterator_tag;
+};
+
+template <bool _Const, class... _Views>
+using __zv_iter_category_base =
+  conditional_t<__zip_all_forward<_Const, _Views...>, __zv_iter_category_base_tag, __zv_iter_category_base_none>;
+
+template <bool _Const, class... _Views>
+class __zip_iterator : public __zv_iter_category_base<_Const, _Views...>
```
miscco: I still believe we should explore how to reuse `cuda::zip_iterator` as much as possible

There is the issue with alias templates not supporting CTAD but that can be worked around with explicit template arguments and we could make the sentinel actually work with `::cuda::zip_iterator` if that helps deduction

miscco:
Please run the full compiler matrix against this one

Old compilers are notoriously bad with zip_view

Jacobfaib:
@miscco https://github.com/NVIDIA/cccl/actions/runs/25381896067

Modulo one tiny fix (older GCC not understanding a conversion operator in the test code) it passes. There are some other failures, but they appear unrelated (compilation errors in CUB, static assertions failing with the new SIMD code, and issues in the module headers).

Jacobfaib:
@coderabbitai full review

libcudacxx/include/cuda/std/__ranges/zip_view.h
```diff
@@ -155,7 +155,8 @@ struct __zv_functors
   template <class _Diff, class _Tuple1, class _Tuple2, size_t _Zero, size_t... _Indices>
   [[nodiscard]] _CCCL_API static constexpr _Diff
   __iter_op_minus(const _Tuple1& __tuple1, const _Tuple2& __tuple2, index_sequence<_Zero, _Indices...>) noexcept(
-    noexcept(((::cuda::std::get<_Indices>(__tuple1) - ::cuda::std::get<_Indices>(__tuple2)) && ...)))
+    noexcept(::cuda::std::get<0>(__tuple1) - ::cuda::std::get<0>(__tuple2))
+    && noexcept(((::cuda::std::get<_Indices>(__tuple1) - ::cuda::std::get<_Indices>(__tuple2)) && ...)))
```
miscco: `cuda::std::get` should be noexcept. So we can have a helper template that checks `declval<const _Iter1&>() - declval<const _Iter2&>()` and use that with `tuple_element_t<_Indices, _Tuple1>`

libcudacxx/include/cuda/std/__ranges/zip_view.h
```diff
@@ -0,0 +1,463 @@
...
+    return ((::cuda::std::get<_Indices>(__tuple1) == ::cuda::std::get<_Indices>(__tuple2)) || ...);
+  }
+
+  template <class _Tuple1, class _Tuple2>
+  [[nodiscard]] _CCCL_API static constexpr bool
+  __iter_op_eq(const _Tuple1& __tuple1, const _Tuple2& __tuple2) noexcept(noexcept(
+    __zv_functors::__iter_op_eq(__tuple1, __tuple2, make_index_sequence<tuple_size_v<remove_cvref_t<_Tuple1>>>{})))
+  {
+    return __zv_functors::__iter_op_eq(__tuple1, __tuple2, make_index_sequence<tuple_size_v<remove_cvref_t<_Tuple1>>>{});
+  }
```
miscco: This is the third time we define this. Once in `cuda::zip_iterator`, then in `cuda::zip_transform_iterator` and now here, we should move all of this into `<cuda/__iterator/zip_common.h>`

libcudacxx/include/cuda/std/__ranges/zip_view.h
```diff
@@ -0,0 +1,463 @@
...
+    template <class _Diff>
+    [[nodiscard]] _CCCL_API constexpr bool operator()(const _Diff& __x, const _Diff& __y) const
+      noexcept(noexcept(__op_comp_abs::__abs(__x) < __op_comp_abs::__abs(__y)))
+    {
+      return __op_comp_abs::__abs(__x) < __op_comp_abs::__abs(__y);
+    }
+  };
+
+  template <class _Tp, class _Up>
+  static constexpr bool __is_nothrow_differentiable = noexcept(declval<const _Tp&>() - declval<const _Up&>());
```
miscco: I would name this `__nothrow_distance`
Lets also move this into `__zip_iter_constraints`

libcudacxx/include/cuda/std/__ranges/zip_view.h
```diff
@@ -0,0 +1,463 @@
...
+    (convertible_to<sentinel_t<_Views>, sentinel_t<__maybe_const<_OtherConst, _Views>>> && ...);
+
+  _CCCL_TEMPLATE(bool _OtherConst = _Const)
+  _CCCL_REQUIRES(_OtherConst _CCCL_AND __all_convertible<_OtherConst>)
+  _CCCL_API constexpr __zip_sentinel(__zip_sentinel<!_OtherConst, _Views...> __i)
+      : __end_{::cuda::std::move(__i.__end_)}
+  {}
+
+  template <class... _Iters>
+  static constexpr bool __is_iterator = same_as<__iterator<_Iters...>, __iterator<iterator_t<_Views>...>>;
```
miscco, line 284: I believe this should rather be named `__is_common`

libcudacxx/include/cuda/std/__ranges/zip_view.h
```diff
@@ -0,0 +1,463 @@
...
+
+  template <class... _Iters>
+  static constexpr bool __is_const_iterator = same_as<__iterator<_Iters...>, __iterator<iterator_t<const _Views>...>>;
```
miscco, line 287: Ditto should rather be `__is_common_const`

libcudacxx/include/cuda/std/__ranges/zip_view.h
```diff
@@ -0,0 +1,463 @@
...
+    // compute instead of needing to instantiate the other types
+    (sizeof...(_Iters) == sizeof...(_Views)) && (__is_const_iterator<_Iters...> || __is_iterator<_Iters...>);
+
+  // tie-breaker in case const and non-const iterators are the same
+  template <class... _Iters>
+  static constexpr bool __use_const_iterator = __is_const_iterator<_Iters...> && !__is_iterator<_Iters...>;
+
+  template <bool _OtherConst>
+  static constexpr bool __all_sentinel =
+    (sentinel_for<sentinel_t<__maybe_const<_Const, _Views>>, iterator_t<__maybe_const<_OtherConst, _Views>>> && ...);
```
miscco, line 301: I believe we should also move those into `__zip_iter_constraints` or `__zip_sentinel_constraints` There were issues with older compilers and static constexpr bool variables of variadic arguments

libcudacxx/include/cuda/std/__ranges/zip_view.h
```diff
@@ -0,0 +1,463 @@
...
+
+  template <class... _Iters, bool _OtherConst = __use_const_iterator<_Iters...>>
+  [[nodiscard]] friend _CCCL_API constexpr auto operator==(const __iterator<_Iters...>& __x, const __zip_sentinel& __y)
+    _CCCL_TRAILING_REQUIRES(bool)(__is_compatible_iterator<_Iters...>&& __all_sentinel<_OtherConst>)
+  {
+    return __zv_functors::__iter_op_eq(__x.__iterators(), __y.__end_);
+  }
```
miscco, line 308: I do not know why this was necessary initially, but I believe if we move the constraint into `__zip_sentinel_constraints` or similar, then we can just use the usual `_CCCL_REQUIRES`

libcudacxx/include/cuda/std/__ranges/zip_view.h
```diff
@@ -0,0 +1,463 @@
...
+  template <class... _Ranges>
+  static constexpr bool __all_input = (input_range<_Ranges> && ...);
+
+  _CCCL_TEMPLATE(class... _Ranges)
+  _CCCL_REQUIRES((sizeof...(_Ranges) > 0) _CCCL_AND __all_input<_Ranges...>)
+  [[nodiscard]] _CCCL_API constexpr auto operator()(_Ranges&&... __rs) const
+    noexcept(noexcept(zip_view<all_t<_Ranges&&>...>{::cuda::std::forward<_Ranges>(__rs)...}))
+      -> decltype(zip_view<all_t<_Ranges&&>...>(::cuda::std::forward<_Ranges>(__rs)...))
+  {
+    return /*------*/ zip_view<all_t<_Ranges>...>{::cuda::std::forward<_Ranges>(__rs)...};
```
miscco: I believe we should just drop the comment, it does not align in our formatting anyhow
```suggestion
    return zip_view<all_t<_Ranges>...>{::cuda::std::forward<_Ranges>(__rs)...};
```

libcudacxx/test/libcudacxx/std/ranges/range.adaptors/range.zip/iterator/arithmetic.pass.cpp
```diff
@@ -0,0 +1,161 @@
...
+constexpr bool canPlusEqual<T, U, cuda::std::void_t<decltype(cuda::std::declval<T&>() += cuda::std::declval<U&>())>> =
+  true;
+
+template <class T, class U, class = void>
+constexpr bool canMinusEqual = false;
+
+template <class T, class U>
+constexpr bool canMinusEqual<T, U, cuda::std::void_t<decltype(cuda::std::declval<T&>() -= cuda::std::declval<U&>())>> =
+  true;
+#endif // TEST_STD_VER <= 2017
```
miscco: When I initially wrote those tests, we did not have the concept macros, I believe we should just either use inline variables or the concept macros

libcudacxx/include/cuda/std/__ranges/zip_view.h
Jacobfaib: Hmm, I don't it's checking whether its a common iterator, rather it is checking whether the incoming iterators are exactly the iterator type.

libcudacxx/include/cuda/std/__ranges/zip_view.h
Jacobfaib: MSVC still doesn't like it unfortunately.

libcudacxx/include/cuda/std/__ranges/zip_view.h
miscco: yeah a common view is a view where sentinel and iterator are the same type

Jacobfaib: Right but this is checking whether those iterators are the _current_ sentinel's expected iterator type, not whether sentinel and iterator are the same type (indeed they aren't, at this point I cannot remember why, but we couldn't just reuse `zip_iterator` for the sentinel).

I guess a better name would be `__is_my_iterator` or `__is_my_const_iterator`.

libcudacxx/include/cuda/__iterator/zip_common.h
```diff
@@ -137,13 +143,118 @@ struct __zip_iter_move
...
     return __iter_move_ret<_Iterators...>{::cuda::std::ranges::__iter_move_cpo{}(__iters)...};
   }
 };
 
+struct __zip_op_eq
+{
+  // Extra level of indirection needed because GCC7 and older clang don't allow you to use
+  // member functions in noexcept() clauses. We also can't use
+  // __is_cpp17_nothrow_equality_comparable_v because the tuple-like type passed to these
+  // functions might not implement operator==().
```
miscco, line 158: I dont understand this, without operator== we cannot instantiate this type?

Shouldnt this be constrained then?

libcudacxx/include/cuda/__iterator/zip_common.h
```diff
@@ -137,13 +143,118 @@ struct __zip_iter_move
...
+    return ((::cuda::std::get<_Indices>(__tuple1) == ::cuda::std::get<_Indices>(__tuple2)) || ...);
+  }
+
+  template <class _Tuple1, class _Tuple2, ::cuda::std::size_t... _Indices>
+  [[nodiscard]] _CCCL_API constexpr bool
+  operator()(const _Tuple1& __tuple1, const _Tuple2& __tuple2, ::cuda::std::index_sequence<_Indices...> __seq) const
+    noexcept(noexcept(::cuda::__zip_op_eq::__do_it(__tuple1, __tuple2, __seq)))
+  {
+    return ::cuda::__zip_op_eq::__do_it(__tuple1, __tuple2, __seq);
+  }
```
miscco, line 173: Why arent we just directly forwarding to `__do_it`?

libcudacxx/include/cuda/std/__ranges/zip_view.h
```diff
@@ -0,0 +1,403 @@
...
+
+  template <bool _OtherConst>
+  static constexpr bool __all_sized_sentinel =
+    (sized_sentinel_for<sentinel_t<__maybe_const<_Const, _Views>>, iterator_t<__maybe_const<_OtherConst, _Views>>>
+     && ...);
+
+  template <bool _OtherConst>
+  using _OtherDiff _CCCL_NODEBUG = common_type_t<range_difference_t<__maybe_const<_OtherConst, _Views>>...>;
+
+  template <class... _Iters, bool _OtherConst = __use_const_iterator<_Iters...>>
```
miscco, line 341: Question: Do we still need the trailing requires?

miscco: I checked and we do

miscco (approved):
Looks good

miscco (approved):

libcudacxx/include/cuda/__iterator/zip_common.h
Jacobfaib: We only care that the sub-types of the tuple-like implements `operator==`, we don't care that the enclosing type does (because we only compare the internal values). So for example the following class should work
```c++
struct NoOperatorEqual {
  int *x, *y, *z; // iterators
  
  bool operator==(const NoOperatorEqual&) = delete;
};
```
`__is_cpp17_nothrow_equality_comparable_v<NoOperatorEqual>` would fail, but we would still be able to do
```c++
get<0>(outer1) == get<0>(outer2) && ...;
```
so it's over constrained. We can use `__is_cpp17_nothrow_equality_comparable_v<tuple_type_t<NoOperatorEqual, 0>>` etc but at that point it's easier to just write the `noexcept(noexcept(...))` expression.

Issue #7671, opened 2026-02-13
Design the API surface of the CUB tuning API


bernhardmgruber:
Before we make any tuning API features public we should carefully review all CUB algorithm's policies, their names and data members. The main goal is to include enough information to allow future evolution (e.g. instead of a `bool is_primitive_type` we should rather store a type enum and derive that information). We should carefully consider which algorithms can share the same policies (e.g. RLE encode and ReduceByKey), and which shouldn't, and how single API entrypoints should handle multiple policies in the background (e.g. rfa and ordinary DeviceReduce have different tuning policies but the same API entrypoint). Furthermore, tuning policies may be specified not only at the CUB API but also for APIs in Thrust or libcu++, so we should put them into a central place. We should aim to have a complete picture about the API surface before going public.

bernhardmgruber:
Here is a list of all current tuning policies and there data members: https://docs.google.com/spreadsheets/d/1EUdwvmv2n7-WRp30-lt7CQxyu8r6atdfJJx3O6Pf1Ag

bernhardmgruber:
I think we need to essentially solve how the public tuning API for `DeviceReduce` will look like: #7577. That is, how policy selectors interplay with requirements at the public API that change the underlying dispatch/kernel/algorithm.

bernhardmgruber:
Based on the following PRs:

- #8745
- #8600
- #8853
- #8863

which serve as examples, here is my current design proposal (summarized by claude based on the above PRs):

## Summary and open questions

All four PRs make CUB's internal tuning policies public so users can customize algorithm behavior per compute capability. The mechanism is:             
                        
1. A policy struct (in the `cub` namespace) holds tuning knobs (threads per block, items per thread, algorithm choice, etc.)                                                                                             
2. Users write a policy selector — a callable taking `cuda::compute_capability` and returning the policy struct                                                                                                      
3. The selector is passed via `cuda::execution::tune(selector)` as an environment to the existing `DeviceXxx` API methods                                                              

## New public entities for `cub::DeviceMergeSort`

| Entity | Data Members |
|--------|--------------|
| `cub::MergeSortPolicy` | `threads_per_block`, `items_per_thread`, `load_algorithm`, `load_modifier`, `store_algorithm` |

Example:
```c++
struct PolicySelector
{
  __host__ __device__ constexpr auto operator()(cuda::compute_capability cc) const -> cub::MergeSortPolicy
  {
    return {.threads_per_block = 256,
[...]

PR #8708, opened 2026-04-28
Unify always_true and always_false into libcu++

rmalani-nv:
## Summary

Closes #7456

This PR introduces unified `cuda::always_true_t` and `cuda::always_false_t` function object types in libcu++, consolidating the previously scattered definitions across CUB, CUDAX, and test code.

### Changes

- **New header**: `cuda/__functional/always_true_false.h` defines `cuda::always_true_t` and `cuda::always_false_t` structs with variadic `operator()`, plus `cuda::always_true` and `cuda::always_false` inline variables
- **New compile-time variable**: `cuda::std::__always_true_v` added alongside existing `__always_false_v` for `static_assert` use
- **CUB unification**: `cub::detail::transform::always_true_predicate` is now a type alias for `cuda::always_true_t`, preserving backward compatibility and CUB's critical `is_same_v` optimization paths
- **`proclaims_copyable_arguments`** specialized for both new types in `address_stability.h`
- **Test cleanup**: CUB test files now use the unified types via type aliases instead of duplicated struct definitions
- **CUDAX STF**: Replaced local `reserved::always_false` with `cuda::std::__always_false_v`
- **C parallel**: Updated NVRTC kernel name strings to reflect the resolved type

### Key design decisions

1. **Type alias for backward compat**: `always_true_predicate` remains available as a name in CUB via `using`, so existing code continues to compile
2. **`is_same_v` checks preserved**: Since `always_true_predicate` is now an alias for `cuda::always_true_t`, all compile-time type checks (used for vectorization optimization) work transparently
3. **Separate concerns**: Runtime function objects (`cuda::always_true_t/always_false_t`) vs compile-time variable templates (`__always_true_v/__always_false_v`) remain distinct, serving different purposes

## Test plan

- [ ] CUB transform tests pass (critical optimization path)
- [ ] CUB select/partition tests pass with unified types
- [ ] PSTL algorithms using `always_true_predicate` default work correctly
- [ ] C parallel NVRTC kernel name resolution works

PR #8745, opened 2026-04-29
Allow public tuning of `cub::DeviceTransform`

### LINKED ISSUE #5570: Verify `DeviceTransform` is using the bulk copy path
I have been asked too many times by users of `cub::DeviceTransform`:

How can I assert that it uses the fast path (i.e. the bulk copy)?

So let's add some mechanism, maybe as part of the environment API overload, to make compilation fail if the fast path is not taken.

### LINKED ISSUE #8585: Productize the `cub::DeviceTransform` tuning API

bernhardmgruber:
- [x] rename memcpy_async to ldgsts
- [x] finish designing unroll_factor
- [x] remove bulk_copy_alignment from the tuning policy
- [x] Merge first: #8757
- [x] Merge first: #8728 
- [x] Merge first: #8835
- [x] Merge first: #8836
- [x] #7671 is sufficiently approved to proceed with this PR. The remaining discussion can continue here.
- [x] Bikeshedding
- [x] ~~Should we split the `TransformAsyncCopyPolicy` into two, one for `ldgsts` and one for `ublkcp`? Then they could evolve independently. I personally see little benefit, since we can extend `TransformAsyncCopyPolicy` for anything we need for bulk copy and ignore the filed for ldgsts. I am also certain that there will be no further development for the ldgsts kernel. Answer: Seems the reviewers were fine
- [x] Are we fine with `TransformAlgorithm::ldgsts` and `TransformAlgorithm::ublkcp`? It was suggested to pick names like `copy_async` or `bulk_copy`. Or include the term `tma`. I don't mind, but please make two consistent suggestions if you care. Answer: Seems the reviewers were fine

Fixes: #8585
Fixes: #5570

## New public entities for `cub::DeviceTransform`

| Entity | Enumerators / Data Members |
|--------|--------------|
| `cub::TransformAlgorithm` | `prefetch`, `vectorized`, `ldgsts`, `ublkcp` |
| `cub::TransformPrefetchPolicy` | `threads_per_block`, `items_per_thread_no_input`, `min_items_per_thread`, `max_items_per_thread`, `prefetch_byte_stride`, `unroll_factor` |
| `cub::TransformVectorizedPolicy` | `threads_per_block`, `items_per_thread`, `vec_size` |
| `cub::TransformAsyncCopyPolicy` | `threads_per_block`, `min_items_per_thread`, `max_items_per_thread`, `unroll_factor` |
| `cub::TransformPolicy` | `min_bytes_in_flight`, `algorithm` (`TransformAlgorithm`), `prefetch` (`TransformPrefetchPolicy`), `vectorized` (`TransformVectorizedPolicy`), `async_copy` (`TransformAsyncCopyPolicy`) |

### DIFF
--- a/cub/cub/detail/cc_dispatch.cuh
+++ b/cub/cub/detail/cc_dispatch.cuh
@@ -112,9 +112,7 @@ CUB_RUNTIME_FUNCTION _CCCL_FORCEINLINE cudaError_t dispatch_to_cc_list(
   // This causes f to be only instantiated for each distinct policy, since the same policy for different arches results
   // in the same integral_constant type passed to f
   using policy_t = decltype(policy_selector(::cuda::compute_capability{}));
-  (...,
-   (device_cc == all_ccs[Is] ? (e = f(policy_constant<policy_t, policy_selector(all_ccs[Is])>{}))
-                             : cudaSuccess));
+  (..., (device_cc == all_ccs[Is] ? (e = f(policy_constant<policy_t, policy_selector(all_ccs[Is])>{})) : cudaSuccess));
 #  else // if _CCCL_STD_VER >= 2020
   // In C++17, we have to collapse architectures with the same policies ourselves, so we instantiate call_for_cc once
   // per policy on the lowest CC which produces the same policy
--- a/cub/cub/detail/cc_dispatch.cuh
+++ b/cub/cub/detail/cc_dispatch.cuh
@@ -84,7 +84,7 @@ struct lowest_cc_resolver<::cuda::std::integer_sequence<int, CudaCcs...>, Policy
 
 // GCC below 12 ICEs in some cases when creating an integral_constant holding a policy
 #  if _CCCL_STD_VER >= 2020 && _CCCL_COMPILER(GCC, <, 12)
-template <auto P>
+template <typename Tp, Tp P>
 struct policy_constant
 {
   _CCCL_API constexpr auto operator()() const noexcept
@@ -93,8 +93,8 @@ struct policy_constant
   }
 };
 #  else // _CCCL_STD_VER >= 2020 && _CCCL_COMPILER(GCC, <, 12)
-template <auto P>
-using policy_constant = ::cuda::std::integral_constant<decltype(P), P>;
+template <typename Tp, Tp P> // using <auto P> will miscompile on GCC 12
+using policy_constant = ::cuda::std::integral_constant<Tp, P>;
 #  endif // _CCCL_STD_VER >= 2020 && _CCCL_COMPILER(GCC, <, 12)
 
 template <typename PolicySelector, typename FunctorT, size_t... Is>
@@ -111,8 +111,9 @@ CUB_RUNTIME_FUNCTION _CCCL_FORCEINLINE cudaError_t dispatch_to_cc_list(
   // In C++20, we just create an integral_constant holding the policy, because policies are structural types in C++20.
   // This causes f to be only instantiated for each distinct policy, since the same policy for different arches results
   // in the same integral_constant type passed to f
+  using policy_t = decltype(policy_selector(::cuda::compute_capability{}));
   (...,
-   (device_cc == all_ccs[Is] ? (e = f(policy_constant<policy_selector(all_ccs[Is])>{}))
+   (device_cc == all_ccs[Is] ? (e = f(policy_constant<policy_t, policy_selector(all_ccs[Is])>{}))
                              : cudaSuccess));
 #  else // if _CCCL_STD_VER >= 2020
   // In C++17, we have to collapse architectures with the same policies ourselves, so we instantiate call_for_cc once
--- a/cub/cub/detail/cc_dispatch.cuh
+++ b/cub/cub/detail/cc_dispatch.cuh
@@ -82,6 +82,21 @@ struct lowest_cc_resolver<::cuda::std::integer_sequence<int, CudaCcs...>, Policy
 };
 #  endif // if _CCCL_STD_VER < 2020
 
+// GCC below 12 ICEs in some cases when creating an integral_constant holding a policy
+#  if _CCCL_STD_VER >= 2020 && _CCCL_COMPILER(GCC, <, 12)
+template <auto P>
+struct policy_constant
+{
+  _CCCL_API constexpr auto operator()() const noexcept
+  {
+    return P;
+  }
+};
+#  else // _CCCL_STD_VER >= 2020 && _CCCL_COMPILER(GCC, <, 12)
+template <auto P>
+using policy_constant = ::cuda::std::integral_constant<decltype(P), P>;
+#  endif // _CCCL_STD_VER >= 2020 && _CCCL_COMPILER(GCC, <, 12)
+
 template <typename PolicySelector, typename FunctorT, size_t... Is>
 CUB_RUNTIME_FUNCTION _CCCL_FORCEINLINE cudaError_t dispatch_to_cc_list(
   PolicySelector policy_selector, ::cuda::compute_capability device_cc, FunctorT&& f, ::cuda::std::index_sequence<Is...>)
@@ -96,9 +111,8 @@ CUB_RUNTIME_FUNCTION _CCCL_FORCEINLINE cudaError_t dispatch_to_cc_list(
   // In C++20, we just create an integral_constant holding the policy, because policies are structural types in C++20.
   // This causes f to be only instantiated for each distinct policy, since the same policy for different arches results
   // in the same integral_constant type passed to f
-  using policy_t = decltype(policy_selector(::cuda::compute_capability{}));
   (...,
-   (device_cc == all_ccs[Is] ? (e = f(::cuda::std::integral_constant<policy_t, policy_selector(all_ccs[Is])>{}))
+   (device_cc == all_ccs[Is] ? (e = f(policy_constant<policy_selector(all_ccs[Is])>{}))
                              : cudaSuccess));
 #  else // if _CCCL_STD_VER >= 2020
   // In C++17, we have to collapse architectures with the same policies ourselves, so we instantiate call_for_cc once
--- a/cub/cub/device/dispatch/tuning/tuning_transform.cuh
+++ b/cub/cub/device/dispatch/tuning/tuning_transform.cuh
@@ -502,21 +502,5 @@ struct policy_selector_from_types<RequiresStableAddress,
     return policies(cc);
   }
 };
-
-struct always_true_predicate
-{
-  template <typename... Ts>
-  _CCCL_HOST_DEVICE constexpr bool operator()(Ts&&...) const
-  {
-    return true;
-  }
-};
 } // namespace detail::transform
 CUB_NAMESPACE_END
-
-namespace cuda
-{
-template <>
-struct proclaims_copyable_arguments<CUB_NS_QUALIFIER::detail::transform::always_true_predicate> : ::cuda::std::true_type
-{};
-} // namespace cuda
--- a/cub/cub/device/dispatch/tuning/tuning_transform.cuh
+++ b/cub/cub/device/dispatch/tuning/tuning_transform.cuh
@@ -30,51 +30,56 @@
 #include <cuda/std/tuple>
 
 CUB_NAMESPACE_BEGIN
-namespace detail::transform
-{
-enum class Algorithm
+
+//! Backend algorithms for @ref DeviceTransform.
+enum class TransformAlgorithm
 {
   // We previously had a fallback algorithm that would use cub::DeviceFor. Benchmarks showed that the prefetch algorithm
   // is always superior to that fallback, so it was removed.
-  prefetch,
-  vectorized,
-  memcpy_async,
-  ublkcp
+  prefetch, //!< Uses a transform kernel that relies on prefetching the memory of all contiguous iterators.
+  vectorized, //!< Uses a transform kernel that uses load and store vectorization.
+  ldgsts, //!< Uses a transform kernel that relies on cp.async/LDGSTS to stage data into shared memory.
+  ublkcp //!< Uses a transform kernel that relies on cp.async.bulk/UBLKCP to stage data into shared memory.
 };
 
 #if _CCCL_HOSTED()
-inline ::std::ostream& operator<<(::std::ostream& os, const Algorithm& algorithm)
+inline ::std::ostream& operator<<(::std::ostream& os, const TransformAlgorithm& algorithm)
 {
   switch (algorithm)
   {
-    case Algorithm::prefetch:
-      return os << "Algorithm::prefetch";
-    case Algorithm::vectorized:
-      return os << "Algorithm::vectorized";
-    case Algorithm::memcpy_async:
-      return os << "Algorithm::memcpy_async";
-    case Algorithm::ublkcp:
-      return os << "Algorithm::ublkcp";
+    case TransformAlgorithm::prefetch:
+      return os << "TransformAlgorithm::prefetch";
+    case TransformAlgorithm::vectorized:
+      return os << "TransformAlgorithm::vectorized";
+    case TransformAlgorithm::ldgsts:
+      return os << "TransformAlgorithm::ldgsts";
+    case TransformAlgorithm::ublkcp:
+      return os << "TransformAlgorithm::ublkcp";
     default:
-      return os << "Algorithm::<unknown>";
+      return os << "TransformAlgorithm::<unknown>";
   }
 }
 #endif // _CCCL_HOSTED()
 
-struct prefetch_policy
+//! The prefetch sub-policy for @ref TransformPolicy.
+struct TransformPrefetchPolicy
 {
-  int threads_per_block;
+  int threads_per_block; //!< Number of threads in a CUDA block
   // items per tile are determined at runtime. these (inclusive) bounds allow overriding that value via a tuning policy
-  int items_per_thread_no_input = 2; // when there are no input iterators, the kernel is just filling
-  int min_items_per_thread      = 1;
-  int max_items_per_thread      = 32;
-  int prefetch_byte_stride      = 128; // somewhat of a cache line size
+  int items_per_thread_no_input = 2; //!< When there are no iterators as inputs, the kernel is just filling. This is the
+                                     //!< number of items written per thread in this case.
+  int min_items_per_thread = 1; //!< Minimum number of items per thread (inclusive)
+  int max_items_per_thread = 32; //!< Maximum number of items per thread (inclusive)
+  int prefetch_byte_stride = 128; //!< The stride in bytes to issue prefetch requests to memory. Corresponds somewhat to
+                                  //!< the size of a cache line.
   // ahendriksen: various unrolling yields less <1% gains at much higher compile-time cost, so prevent unrolling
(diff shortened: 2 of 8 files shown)


cub/test/catch2_test_device_transform_env_api.cu
```diff
@@ -0,0 +1,56 @@
...
+#include <iostream>
+
+#include <c2h/catch2_test_helper.h>
+
+#if _CCCL_STD_VER >= 2020
+
+// example-begin transform-policy-selector
+struct MyTransformPolicySelector
+{
+  __host__ __device__ constexpr auto operator()(cuda::compute_capability /*cc*/) const -> cub::TransformPolicy
```
NaderAlAwar, line 27: Important: this will fail to build because we are using `cuda::compute_capability` instead of `arch_id`

cub/cub/device/dispatch/tuning/tuning_transform.cuh
```diff
@@ -407,10 +406,10 @@ struct policy_selector
     else if (arch >= ::cuda::arch_id::sm_80)
     {
       const int block_threads = 256;
-      const auto prefetch     = prefetch_policy{block_threads};
+      const auto prefetch     = TransformPrefetchPolicy{block_threads};
       const auto vectorized =
         tuned_vectorized_policy(arch, ::cuda::std::max(1, output.value_type_size), no_input_streams);
-      const auto async = async_copy_policy{block_threads, ldgsts_size_and_align};
+      const auto async = TransformAsyncCopyPolicy{block_threads, ldgsts_size_and_align};
```
NaderAlAwar: Critical: after removing `bulk_copy_alignment` from `TransformAsyncCopyPolicy`, passing `ldgsts_size_and_align` to the constructor here will set `min_items_per_thread`

cub/cub/device/dispatch/tuning/tuning_transform.cuh
```diff
@@ -29,184 +29,183 @@
...
-  prefetch_policy prefetch;
-  vectorized_policy vectorized;
-  async_copy_policy async_copy;
-
-  [[nodiscard]] _CCCL_API constexpr friend bool operator==(const transform_policy& lhs, const transform_policy& rhs)
+  int min_bytes_in_flight; //!< Minimum number of bytes in flight per SM to reach by scaling the items per thread. Has
+                           //!< no effect if algorithm is @p vectorized.
+  TransformAlgorithm algorithm; //!< The transform algorithm to use by the kernel
+  TransformPrefetchPolicy prefetch; //!< Sub-policy for the prefetch algorithm. Only used when @p algorithm is @p
+                                    //!< prefetch or when @p algorithm is @p vectorized and the input pointrs are not
```
NaderAlAwar: Nit:

```suggestion
                                    //!< prefetch or when @p algorithm is @p vectorized and the input pointers are not
```

cub/cub/device/dispatch/tuning/tuning_transform.cuh
```diff
@@ -29,184 +29,183 @@
...
+enum class TransformAlgorithm
 {
   // We previously had a fallback algorithm that would use cub::DeviceFor. Benchmarks showed that the prefetch algorithm
   // is always superior to that fallback, so it was removed.
-  prefetch,
-  vectorized,
-  memcpy_async,
-  ublkcp
+  prefetch, //!< Uses a transform kernel that relies on prefetching the memory of all contiguous iterators.
+  vectorized, //!< Uses a transform kernel that used load and store vectorization.
```
NaderAlAwar: Nit:

```suggestion
  vectorized, //!< Uses a transform kernel that uses load and store vectorization.
```

cub/test/catch2_test_device_transform_env_api.cu
bernhardmgruber: It works now since  #8728 was merged.

bernhardmgruber:
ok, what is happening here:
```
 /home/coder/cccl/cub/test/catch2_test_device_transform_env_api.cu:23:9: error: use of GNU old-style field designator extension [-Werror,-Wgnu-designator]
     23 | return {min_bytes_in_flight: 64 * 1024, algorithm: cub::TransformAlgorithm::prefetch, prefetch: {block_threads: 256, }, vectorized: {}, async_copy: {}}; 
        |         ^~~~~~~~~~~~~~~~~~~~
        |         .min_bytes_in_flight = 
  /home/coder/cccl/cub/test/catch2_test_device_transform_env_api.cu:23:41: error: use of GNU old-style field designator extension [-Werror,-Wgnu-designator]
     23 | return {min_bytes_in_flight: 64 * 1024, algorithm: cub::TransformAlgorithm::prefetch, prefetch: {block_threads: 256, }, vectorized: {}, async_copy: {}}; 
        |                                         ^~~~~~~~~~
        |                                         .algorithm = 
  /home/coder/cccl/cub/test/catch2_test_device_transform_env_api.cu:23:87: error: use of GNU old-style field designator extension [-Werror,-Wgnu-designator]
     23 | return {min_bytes_in_flight: 64 * 1024, algorithm: cub::TransformAlgorithm::prefetch, prefetch: {block_threads: 256, }, vectorized: {}, async_copy: {}}; 
        |                                                                                       ^~~~~~~~~
        |                                                                                       .prefetch = 
  /home/coder/cccl/cub/test/catch2_test_device_transform_env_api.cu:23:98: error: use of GNU old-style field designator extension [-Werror,-Wgnu-designator]
     23 | return {min_bytes_in_flight: 64 * 1024, algorithm: cub::TransformAlgorithm::prefetch, prefetch: {block_threads: 256, }, vectorized: {}, async_copy: {}}; 
        |                                                                                                  ^~~~~~~~~~~~~~
        |                                                                                                  .block_threads = 
  /home/coder/cccl/cub/test/catch2_test_device_transform_env_api.cu:23:121: error: use of GNU old-style field designator extension [-Werror,-Wgnu-designator]
     23 | return {min_bytes_in_flight: 64 * 1024, algorithm: cub::TransformAlgorithm::prefetch, prefetch: {block_threads: 256, }, vectorized: {}, async_copy: {}}; 
        |                                                                                                                         ^~~~~~~~~~~
        |                                                                                                                         .vectorized = 
  /home/coder/cccl/cub/test/catch2_test_device_transform_env_api.cu:23:137: error: use of GNU old-style field designator extension [-Werror,-Wgnu-designator]
     23 | return {min_bytes_in_flight: 64 * 1024, algorithm: cub::TransformAlgorithm::prefetch, prefetch: {block_threads: 256, }, vectorized: {}, async_copy: {}}; 
        |                                                                                                                                         ^~~~~~~~~~~
        |                                                                                                                                         .async_copy = 
  6 errors generated.
```
The corresponding source code:
```c++
    return {.min_bytes_in_flight = 64 * 1024,
            .algorithm           = cub::TransformAlgorithm::prefetch,
            .prefetch            = {.block_threads = 256},
            .vectorized          = {},
            .async_copy          = {}};
```
Does nvcc translate designated initializers into a GNU extension syntax?

cub/cub/detail/cc_dispatch.cuh
```diff
@@ -107,8 +107,8 @@ struct policy_constant
   }
 };
 #  else // _CCCL_STD_VER >= 2020 && _CCCL_COMPILER(GCC, <, 12)
-template <auto P>
-using policy_constant = ::cuda::std::integral_constant<decltype(P), P>;
+template <typename Tp, Tp P> // using <auto P> will miscompile on GCC 12
+using policy_constant = ::cuda::std::integral_constant<Tp, P>;
```
miscco, line 97: Could you please verify that the GCC-11 workaround is still needed with the explicit NTTP passed around?

cub/cub/device/dispatch/tuning/tuning_transform.cuh
```diff
@@ -499,6 +498,21 @@ struct policy_selector_from_types<RequiresStableAddress,
     return policies(cc);
   }
 };
-} // namespace detail::transform
 
+struct always_true_predicate
```
oleksandr-pavlyk: Question: can we use ``cuda::always_true_t``?

bernhardmgruber: When it lands, sure! PR is still open: #8708

cub/cub/detail/cc_dispatch.cuh
bernhardmgruber: I don't follow entirely, sorry. I added a GCC-11 workaround because it couldn't compile something inside `cuda::std::integral_constant`. How should I roll back that workaround without using `cuda::std::integral_constant`?

bernhardmgruber:
Converting to draft to prevent accidental merging until  #7671 is resolved.

cub/cub/device/dispatch/tuning/tuning_transform.cuh
bernhardmgruber: merged in the meantime, dropping it here.

cub/cub/detail/cc_dispatch.cuh
```diff
@@ -82,6 +82,21 @@ struct lowest_cc_resolver<::cuda::std::integer_sequence<int, CudaCcs...>, Policy
...
+// GCC below 12 ICEs in some cases when creating an integral_constant holding a policy
+#  if _CCCL_STD_VER >= 2020 && _CCCL_COMPILER(GCC, <, 12)
+template <typename Tp, Tp P>
+struct policy_constant
+{
+  _CCCL_API constexpr auto operator()() const noexcept
+  {
+    return P;
+  }
+};
```
bernhardmgruber, line 94: TODO: test again whether we need this.

cub/cub/detail/cc_dispatch.cuh
bernhardmgruber: I just confirmed again that GCC 11 dies at `integral_constant`:
```
/home/coder/cccl/lib/cmake/libcudacxx/../../../libcudacxx/include/cuda/std/__type_traits/integral_constant.h:28:61: internal compiler error: in finish_member_declaration, at cp/semantics.c:3408
     28 | struct _CCCL_TYPE_VISIBILITY_DEFAULT integral_constant
        |                                                             ^
```

cub/cub/detail/cc_dispatch.cuh
bernhardmgruber: Yes we need it.

cub/cub/device/dispatch/tuning/tuning_transform.cuh
```diff
@@ -30,51 +30,56 @@
...
-      return os << "Algorithm::ublkcp";
+    case TransformAlgorithm::prefetch:
+      return os << "TransformAlgorithm::prefetch";
+    case TransformAlgorithm::vectorized:
+      return os << "TransformAlgorithm::vectorized";
+    case TransformAlgorithm::ldgsts:
+      return os << "TransformAlgorithm::ldgsts";
+    case TransformAlgorithm::ublkcp:
+      return os << "TransformAlgorithm::ublkcp";
     default:
```
Jacobfaib, line 58: Don't add a `default:` case here (or in any switch-over-enum in general). If nothing matches a switch-case, it will fall through anyways, whilst adding a `default` case will stop compilers warning about uncovered cases. So for example:
```c++
enum Enum {
  FOO, BAR
};

// No compiler warning
switch (e) {
  case Enum::FOO: // ...
  default: // ...
  // oops, BAR not handled
}

// Compiler will warn "unhandled case BAR"
switch (e) {
  case Enum::FOO: // ...
  // oops, BAR not handled
}
```

cub/cub/device/dispatch/tuning/tuning_transform.cuh
```diff
@@ -83,121 +88,129 @@ struct prefetch_policy
...
 #if _CCCL_HOSTED()
-  friend ::std::ostream& operator<<(::std::ostream& os, const prefetch_policy& policy)
+  friend ::std::ostream& operator<<(::std::ostream& os, const TransformPrefetchPolicy& policy)
   {
     return os
-        << "prefetch_policy { .threads_per_block = " << policy.threads_per_block << ", .items_per_thread_no_input = "
-        << policy.items_per_thread_no_input << ", .min_items_per_thread = " << policy.min_items_per_thread
-        << ", .max_items_per_thread = " << policy.max_items_per_thread << ", .prefetch_byte_stride = "
-        << policy.prefetch_byte_stride << ", .unroll_factor = " << policy.unroll_factor << " }";
+        << "TransformPrefetchPolicy { .threads_per_block = " << policy.threads_per_block
```
Jacobfaib, line 100: Nit: it may be worth adding a local `// clang-tidy: off` here and formatting this line by line:
```c++
os << "prefetch_policy {"
   << ".threads_per_block = " << policy.threads_per_block << ", "
   << ...
```
the current for is... quite unreadable and easy to miss added members :)

cub/cub/device/dispatch/tuning/tuning_transform.cuh
```diff
@@ -83,121 +88,129 @@ struct prefetch_policy
...
-  int threads_per_block;
-  int items_per_thread;
-  int vec_size;
+  int threads_per_block; //!< Number of threads in a CUDA block
+  int items_per_thread; //!< Number of items processed per thread. Must be a multiple of vec_size.
+  int vec_size; //!< Number of elements loaded/stored per vectorized access. Must evenly divide items_per_thread.
 
   [[nodiscard]] _CCCL_HOST_DEVICE_API constexpr friend bool
-  operator==(const vectorized_policy& lhs, const vectorized_policy& rhs)
+  operator==(const TransformVectorizedPolicy& lhs, const TransformVectorizedPolicy& rhs)
```
Jacobfaib: Nit: add `noexcept` while we are here, can't hurt (and below)

cub/cub/device/dispatch/tuning/tuning_transform.cuh
```diff
@@ -83,121 +88,129 @@ struct prefetch_policy
...
 #endif // _CCCL_HOSTED()
 };
 
+namespace detail::transform
+{
 #if _CCCL_HAS_CONCEPTS()
 template <typename T>
-concept transform_policy_selector = policy_selector<T, transform_policy>;
+concept transform_policy_selector = policy_selector<T, TransformPolicy>;
 #endif // _CCCL_HAS_CONCEPTS()
```
Jacobfaib, line 214: Semi important: You could use the `CCCL_CONCEPT` macros here and expose these concepts unconditionally.

cub/cub/device/dispatch/tuning/tuning_transform.cuh
Jacobfaib: `inline constexpr`. My reading of https://eel.is/c++draft/dcl.constexpr#1.sentence-3 is that implicit `inline` is only for static constexpr class members.

cub/cub/device/dispatch/tuning/tuning_transform.cuh
bernhardmgruber: Hmm, but maybe I want to print "TransformAlgorithm::<unknown>" if the user casts an unknown integer to my enum? Or is that bad practice as well?

I know what you are referring to above, but that applies when the `default:` does not contain any code itself.

cub/cub/device/dispatch/tuning/tuning_transform.cuh
bernhardmgruber: I would rather try to have clang-format be enabled as much as possible. Also, do people still add data members by hand and not have an agent do it? ;)

cub/cub/device/dispatch/tuning/tuning_transform.cuh
Jacobfaib: > Hmm, but maybe I want to print "TransformAlgorithm::" if the user casts an unknown integer to my enum?
 
You should handle this outside the switch. If no case matches in a switch-case, it simply skips the entire expression as if you had a `default: break;`.

cub/cub/device/dispatch/tuning/tuning_transform.cuh
bernhardmgruber: I believe I tried to convert `policy_selector` to the concept emulation macros in the past and failed. I forgot why. Maybe it was the check whether `operator()` is a constant expression, but I removed that check by now.

cub/cub/device/dispatch/tuning/tuning_transform.cuh
```diff
@@ -30,51 +30,56 @@
...
+                                     //!< number of items written per thread in this case.
+  int min_items_per_thread = 1; //!< Minimum number of items per thread (inclusive)
+  int max_items_per_thread = 32; //!< Maximum number of items per thread (inclusive)
+  int prefetch_byte_stride = 128; //!< The stride in bytes to issue prefetch requests to memory. Corresponds somewhat to
+                                  //!< the size of a cache line.
   // ahendriksen: various unrolling yields less <1% gains at much higher compile-time cost, so prevent unrolling
   // bgruber: but A6000 and H100 show small gains without pragma, so omitting pragma
-  int unroll_factor = -1; // -1 means we leave it to the compiler
+  int unroll_factor = 0; //!< For any value >1, the unroll factor for the transformation loop in the kernel. The value 0
+                         //!< retains the compiler's default unrolling by specifying no unroll pragma. 1 prevents no
```
pauleonix:
```suggestion
                         //!< retains the compiler's default unrolling by specifying no unroll pragma. 1 prevents
```

cub/test/catch2_test_device_transform_env_api.cu
```diff
@@ -0,0 +1,64 @@
...
+#  endif // _CCCL_COMPILER(CLANG)
+
+// example-begin transform-policy-selector
+struct TransformPolicySelector
+{
+  __host__ __device__ constexpr auto operator()(cuda::compute_capability /*cc*/) const -> cub::TransformPolicy
+  {
+    return {.min_bytes_in_flight = 64 * 1024,
+            .algorithm           = cub::TransformAlgorithm::prefetch,
+            .prefetch            = {.threads_per_block = 256},
```
pauleonix, line 31: I assume that if we want to retune sth in the future we will just modify our policy-selector while the tuning parameter defaults implicit in the policy can just stay the same even if that means they are bad?

pauleonix (approved):

cub/cub/device/dispatch/tuning/tuning_transform.cuh
bernhardmgruber: Please show a code example. What scalable condition should it put into the conditional
```c++
  switch (algorithm)
  {
    case TransformAlgorithm::prefetch:
      return os << "TransformAlgorithm::prefetch";
    case TransformAlgorithm::vectorized:
      return os << "TransformAlgorithm::vectorized";
    case TransformAlgorithm::ldgsts:
      return os << "TransformAlgorithm::ldgsts";
    case TransformAlgorithm::ublkcp:
      return os << "TransformAlgorithm::ublkcp";
  }
  if (/* HERE */) {
      return os << "TransformAlgorithm::<unknown>";
  }
```
Without mentioning all of the above enumerators again?

cub/test/catch2_test_device_transform_env_api.cu
bernhardmgruber: We can now change ANYTHING we want in the future. Users will now have the ability to set their own tuning parameters, so any change we do that regresses a user can be reverted by the user specifying the old tuning policy manually.

davebayer (approved):

PR #8751, opened 2026-04-30
Handle unexpected struct_name in cuda.coop kernel codegen

randomizedcoder:
## Description

In `python/cuda_cccl/cuda/coop/_types.py`, `Algorithm._kernel_source()` assigns `provide_alloc_version`, `storage`, and `sync` only inside an if/elif chain that checks whether `self.struct_name` starts with `"Warp"` or `"Block"`. If neither matches, all three variables are used unbound on subsequent lines, raising `UnboundLocalError`.

Currently only "Warp" and "Block" prefixes exist, so this doesn't trigger in practice. The fix adds an `else` clause with a clear `ValueError` so the failure mode is explicit rather than a confusing `UnboundLocalError`.

### Change

```diff
             elif self.struct_name.startswith("Block"):
                 provide_alloc_version = True
                 storage = "__shared__ temp_storage_t temp_storage;"
                 sync = "__syncthreads();"
+            else:
+                raise ValueError(f"Unsupported struct_name: {self.struct_name!r}, expected 'Warp*' or 'Block*'")
```

Found via pylint 3.3.7 (E0606: possibly-used-before-assignment).

## Checklist

### DIFF
--- a/python/cuda_cccl/cuda/coop/_types.py
+++ b/python/cuda_cccl/cuda/coop/_types.py
@@ -719,6 +719,8 @@ class Algorithm:
                 provide_alloc_version = True
                 storage = "__shared__ temp_storage_t temp_storage;"
                 sync = "__syncthreads();"
+            else:
+                raise ValueError(f"Unsupported struct_name: {self.struct_name!r}, expected 'Warp*' or 'Block*'")
 
             buf = StringIO()
             w = buf.write


randomizedcoder:
Closing this as obsolete — the code it patches no longer exists in the repo.

This PR added a missing `else: raise ValueError(...)` to `get_lto_ir` in `python/cuda_cccl/cuda/coop/_types.py`, guarding against an unexpected `struct_name` falling through with `provide_alloc_version`/`storage`/`sync` unbound.

Since this PR was opened, that module was renamed to `cuda/coop/_experimental/_types.py` (#8788) and the entire `cuda.coop` module was then removed from `cuda-cccl` in #10104 ("Remove cuda.coop._experimental from cuda-cccl"). The `get_lto_ir` / `struct_name` / `provide_alloc_version` symbols no longer appear anywhere in the tree, so there's nothing left to fix here. Closing.

PR #8752, opened 2026-04-30
Provide simplified cuda device atomics

miscco:
We use need atomic stores loads and adds in CUB

However, using a full <atomic> is a bit of an overkill for that.

This introduces a simplified function that should do the same trick without pulling in all of atomic


miscco:
does not seem worth it currently

PR #8755, opened 2026-04-30
[cudax] Change the group mapping application logic

davebayer:
This PR changes the way mappings are applied during the mapping phase of group creation. I found out it would be better if the mappings worked as transformers that take the previous mapping result and return a modified mapping result. This is a step towards allowing combining multiple mappings together to produce 1 group.

### DIFF
--- a/cudax/test/group/synchronizer/lane_synchronizer.cu
+++ b/cudax/test/group/synchronizer/lane_synchronizer.cu
@@ -33,11 +33,12 @@ __device__ void test_lane_synchronizer(const Level& level, Config config)
   // Test make_instance(...).
   {
     const auto parent_group = cudax::make_this_group(level, config);
+    const ThreadsInWarpMappingResult prev_mapping_result;
 
     const cudax::group_by mapping{2};
     const Synchronizer synchronizer{};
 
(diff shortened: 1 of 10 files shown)


cudax/test/group/synchronizer/lane_synchronizer.cu
```diff
@@ -33,11 +33,12 @@ __device__ void test_lane_synchronizer(const Level& level, Config config)
   // Test make_instance(...).
   {
     const auto parent_group = cudax::make_this_group(level, config);
+    const ThreadsInWarpMappingResult prev_mapping_result;
```
pciolkosz, line 36: I don't get why is this always thread in warp mapping, while the parent varies. Applies in other tests too

pciolkosz (approved):
A bit confused on the testing, but the tests pass, so it probably just confusion and I will approve

cudax/test/group/synchronizer/lane_synchronizer.cu
davebayer: Because I'm just grouping threads within a warp in this test. I agree, it's a bit confusing

PR #8762, opened 2026-04-30
[HostJit] Avoid calling host clock features

miscco:
We do not have native clock support in HostJit


libcudacxx/include/cuda/std/ctime
```diff
@@ -32,7 +32,8 @@
 #if _CCCL_FREESTANDING()
 #  define TIME_UTC 1
 
-using time_t = long long int;
+using clock_t = long long int;
```
davebayer, line 36: NVRTC defines `clock_t`

davebayer (approved):

PR #8427, opened 2026-04-15
[thrust] Single-pass `is_partitioned` via adjacent zip_iterator

edenfunf:
## What

Replaces the two-pass `thrust::is_partitioned` CUDA implementation with a single-pass version that mirrors `cuda::std::is_partitioned` from the PSTL.

**Before:**
```cpp
ItemsIt boundary = cuda_cub::find_if_not(policy, first, last, predicate);
ItemsIt end      = cuda_cub::find_if(policy, boundary, last, predicate);
return end == last;
```
Two kernel launches in the fully-partitioned case (the common fast path).

**After:**
```cpp
const auto first_zip = ::cuda::make_zip_iterator(first, first + 1);
const auto last_zip  = ::cuda::make_zip_iterator(last, last);
const auto result    = cuda_cub::find_if(policy, first_zip, last_zip,
                         __is_partitioned_fn<Predicate>{predicate});
return ::cuda::std::get<1>(result.__iterators()) == last;
```
One kernel launch, scanning adjacent pairs for a "false → true" transition.

## Why

`thrust::is_partitioned` was benchmarked as significantly slower than `cuda::std::is_partitioned` (up to ~33 % slower for `int8_t` at large N), even though both check the same invariant. The root cause is the extra kernel launch:

- `find_if_not` scans the whole range to find the boundary.
- `find_if` then scans from the boundary to the end.

The PSTL avoids this by zipping adjacent elements and scanning once for any `(!pred(a[i]) && pred(a[i+1]))` pair. This PR ports that approach to the Thrust CUDA backend.

Closes #8085.

## How

- Added `__is_partitioned_fn<Predicate>` functor (host+device callable) that takes a `cuda::zip_iterator` value and returns `true` on a partitioning violation.
- Used `cuda::make_zip_iterator(first, first+1)` / `cuda::make_zip_iterator(last, last)`. Because `zip_iterator::operator-` returns the minimum component distance, the synthetic range has exactly `n-1` elements — the number of adjacent pairs.
- The return check uses `::cuda::std::get<1>(result.__iterators()) == last` (not `result == last_zip`) to handle the `n==1` edge case correctly: when `n==1` the zip distance is 0, `find_if_n` returns `first_zip`, and `get<1>(first_zip) == last` evaluates to `true` as expected.
- Pattern directly mirrors `libcudacxx/include/cuda/std/__pstl/is_partitioned.h`.
[...]

PR #8769, opened 2026-04-30
[thrust] Fix is_partitioned with non-const predicates

### LINKED ISSUE #8714: [BUG]: is_partitioned fails with error: no instance of function template "TestUtils::NonConstAdapter<F>::operator()
### Is this a duplicate?

### Type of Bug

Compile-time Error

### Component

Thrust

### Describe the bug

```
#include <execution>
#include <algorithm>
#include <vector>

struct test_non_const
{
  template <typename Policy, typename Iterator>
  void
  operator()(Policy&& exec, Iterator iter)
  {
    auto is_even = [&](int v) {
      uint32_t i = (uint32_t)v;
      return i % 2 == 0;
    };
    invoke_if(exec, [&]() { std::is_partitioned(exec, iter, iter, non_const(is_even)); });
  }
};


int32_t
main()
{
  const std::size_t max_n = 1000000;
  std::vector<int> in;
  in.reserve(max_n);
  const bool actual = std::is_partitioned(std::execution::par_unseq, in.begin(), in.begin() + 100, test_non_const());
  return 0;
}

```
This test fails with:
```
thrust/system/cuda/detail/partition.h", line 383: error: no instance of function template "TestUtils::NonConstAdapter<F>::operator() [with F=lambda [](TestUtils::float64_t)->bool]" matches the argument list and object (the object has type qualifiers that prevent a match)
            argument types are: (int)
            object type is: const TestUtils::NonConstAdapter<lambda [](TestUtils::float64_t)->bool>
      const bool lhs = pred_(thrust::raw_reference_cast(::cuda::std::get<0>(tuple)));
                       ^
```

### How to Reproduce

```
nvc++ -stdpar --c++17  test_is_partitioned.cpp
```

### Expected behavior

This test should compile fine after the fix.

### Reproduction link

_No response_

### Operating System

_No response_

### nvidia-smi output

_No response_

### NVCC version

_No response_

edenfunf:
## What

`thrust::is_partitioned` on the CUDA backend rejects predicates whose `operator()` is not const-qualified.

The compile error reported in #8714:

```
.../partition.h(383): error: no instance of function template "MyPred::operator()" matches
  the argument list and object (the object has type qualifiers that prevent a match)
  const bool lhs = pred_(thrust::raw_reference_cast(::cuda::std::get<0>(tuple)));
                   ^
```

## Why

The single-pass implementation merged in #8427 introduced an internal wrapper `__is_partitioned_fn` whose `operator()` is const-qualified. Inside a const member function, the `pred_` data member is treated as `const Predicate`, so any user predicate whose own `operator()` is non-const cannot be invoked.

Thrust does not require predicates to be const-callable (`thrust/testing/cuda/unique.cu` already exercises this for `unique`). The two-pass implementation prior to #8427 forwarded the predicate directly to `find_if`/`find_if_not` without an intermediate wrapper, so non-const predicates worked.

Closes #8714.

## How

- **Fix**: drop the trailing `const` on `__is_partitioned_fn::operator()`. The wrapper is constructed locally and consumed only by `cuda_cub::find_if`, which takes the predicate by value and stores it in a non-const `agent_t::predicate` member. No call site requires the wrapper itself to be const-callable.

- **Comment**: a short note above the operator explains why the `const` is intentionally absent, to discourage it from being added back during a future cleanup.

- **Test**: `TestIsPartitionedWithNonConstPredicate` in `thrust/testing/cuda/is_partitioned.cu`, using the same `NonConstAdapter` pattern already established at `thrust/testing/cuda/unique.cu:455`. Two assertions exercise both branches of the wrapper's `!lhs && rhs` logic (partitioned and not-partitioned inputs). The pre-existing `is_even<T>` predicate template is hoisted out of `#ifdef THRUST_TEST_DEVICE_SIDE` so the new test can reuse it.

## Test

Built locally with the `thrust-cpp17` preset (CUDA 13.0, gcc14, sm_80):

```
ninja thrust.cpp.cuda.test.cuda.is_partitioned thrust.cpp.cuda.test.is_partitioned
```

All three resulting executables compile clean (`cdp_0`, `cdp_1`, generic), zero warnings.

Reverting just the fix and rebuilding confirms the new regression test fails to compile with the exact error reported in #8714, so it serves as a true regression guard.

### DIFF
--- a/thrust/testing/cuda/is_partitioned.cu
+++ b/thrust/testing/cuda/is_partitioned.cu
@@ -2,15 +2,9 @@
 #include <thrust/functional.h>
 #include <thrust/partition.h>
 
-#include <unittest/unittest.h>
+#include <cuda/std/utility>
 
-#ifdef THRUST_TEST_DEVICE_SIDE
-template <typename ExecutionPolicy, typename Iterator, typename Predicate, typename Iterator2>
-__global__ void
-is_partitioned_kernel(ExecutionPolicy exec, Iterator first, Iterator last, Predicate pred, Iterator2 result)
-{
-  *result = thrust::is_partitioned(exec, first, last, pred);
-}
+#include <unittest/unittest.h>
 
 template <typename T>
 struct is_even
@@ -20,6 +14,14 @@ struct is_even
     return ((int) x % 2) == 0;
   }
 };
+
+#ifdef THRUST_TEST_DEVICE_SIDE
+template <typename ExecutionPolicy, typename Iterator, typename Predicate, typename Iterator2>
+__global__ void
+is_partitioned_kernel(ExecutionPolicy exec, Iterator first, Iterator last, Predicate pred, Iterator2 result)
+{
+  *result = thrust::is_partitioned(exec, first, last, pred);
+}
 
 template <typename ExecutionPolicy>
 void TestIsPartitionedDevice(ExecutionPolicy exec)
@@ -110,3 +112,38 @@ void TestIsPartitionedCudaStreams()
   cudaStreamDestroy(s);
 }
 DECLARE_UNITTEST(TestIsPartitionedCudaStreams);
+
+// Wraps a callable with a non-const operator() so the predicate becomes
+// non-const-callable. is_partitioned must accept such predicates;
+// do not add const to operator() — that would defeat the test.
+template <typename F>
+struct NonConstAdapter
+{
+  F f;
+
+  NonConstAdapter(const F& func)
+      : f(func)
+  {}
+
+  template <typename... Args>
+  _CCCL_HOST_DEVICE auto operator()(Args&&... args) -> decltype(f(cuda::std::forward<Args>(args)...))
+  {
+    return f(cuda::std::forward<Args>(args)...);
+  }
+};
+
+void TestIsPartitionedWithNonConstPredicate()
+{
+  thrust::device_vector<int> partitioned   = {0, 2, 4, 1, 3, 5};
+  thrust::device_vector<int> unpartitioned = {0, 1, 2, 3};
+
+  using Adapter = NonConstAdapter<is_even<int>>;
+
+  ASSERT_EQUAL_QUIET(
+    true, thrust::is_partitioned(thrust::cuda::par, partitioned.begin(), partitioned.end(), Adapter{is_even<int>{}}));
+
+  ASSERT_EQUAL_QUIET(
+    false,
+    thrust::is_partitioned(thrust::cuda::par, unpartitioned.begin(), unpartitioned.end(), Adapter{is_even<int>{}}));
+}
+DECLARE_UNITTEST(TestIsPartitionedWithNonConstPredicate);
--- a/thrust/testing/cuda/is_partitioned.cu
+++ b/thrust/testing/cuda/is_partitioned.cu
@@ -2,9 +2,15 @@
 #include <thrust/functional.h>
 #include <thrust/partition.h>
 
-#include <cuda/std/utility>
-
 #include <unittest/unittest.h>
+
+#ifdef THRUST_TEST_DEVICE_SIDE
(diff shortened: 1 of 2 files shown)


thrust/testing/cuda/is_partitioned.cu
```diff
@@ -110,3 +112,38 @@ void TestIsPartitionedCudaStreams()
...
+  NonConstAdapter(const F& func)
+      : f(func)
+  {}
+
+  template <typename... Args>
+  _CCCL_HOST_DEVICE auto operator()(Args&&... args) -> decltype(f(cuda::std::forward<Args>(args)...))
+  {
+    return f(cuda::std::forward<Args>(args)...);
+  }
+};
```
bernhardmgruber: Suggestion: we can just use something like:
```c++
template <typename T>
struct is_even_non_const
{
  _CCCL_HOST_DEVICE bool operator()(T x) // no const
  {
    return ((int) x % 2) == 0;
  }
};
```

thrust/testing/cuda/is_partitioned.cu
```diff
@@ -21,6 +15,14 @@ struct is_even
...
 };
 
+#ifdef THRUST_TEST_DEVICE_SIDE
+template <typename ExecutionPolicy, typename Iterator, typename Predicate, typename Iterator2>
+__global__ void
+is_partitioned_kernel(ExecutionPolicy exec, Iterator first, Iterator last, Predicate pred, Iterator2 result)
+{
+  *result = thrust::is_partitioned(exec, first, last, pred);
+}
+
```
bernhardmgruber: Q: Why do we need to move this code?

edenfunf: The kernel's absolute line numbers shifted because I hoisted `is_even<T>` above the `#ifdef` so the new test could reuse it. The kernel itself wasn't restructured (its content and its position relative to the `#ifdef` are unchanged). With your suggestion above (`is_even_non_const` as a self-contained predicate), the hoisting is no longer needed and the kernel is back at its original lines in the latest commit.

thrust/testing/cuda/is_partitioned.cu
edenfunf: Thanks, applied in the latest commit.

bernhardmgruber (approved):

miscco (approved):

edenfunf:
MSVC build hit nvcc OOM in unrelated files looks like a CI flake. Mind re-running the failed jobs? Thanks!

thrust/thrust/system/cuda/detail/partition.h
miscco, line 389: I believe we should have both the const and the non-const overloads
```suggestion
  // Not const-qualified: a const operator() would propagate const onto pred_
  // and reject predicates whose own operator() is non-const (Thrust permits these).
template <class Tuple>
  [[nodiscard]] _CCCL_HOST_DEVICE bool operator()(const Tuple& tuple)
{
const bool lhs = pred_(thrust::raw_reference_cast(::cuda::std::get<0>(tuple)));
const bool rhs = pred_(thrust::raw_reference_cast(::cuda::std::get<1>(tuple)));
return !lhs && rhs;
}
};

template <class Tuple>
  [[nodiscard]] _CCCL_HOST_DEVICE bool operator()(const Tuple& tuple) const
{
const bool lhs = pred_(thrust::raw_reference_cast(::cuda::std::get<0>(tuple)));
const bool rhs = pred_(thrust::raw_reference_cast(::cuda::std::get<1>(tuple)));
return !lhs && rhs;
}
};
```

bernhardmgruber: I would be fine with the status quo of the PR, but I also don't mind if we add both overloads.

bernhardmgruber:
@edenfunf thank you for the fix!

PR #8771, opened 2026-04-30
Fix invalid C++ syntax: remove leading :: after struct in proclaims_copyable_arguments specializations

### LINKED ISSUE #8833: CUB: ill-formed class-head name in device_transform.cuh breaks GCC (CUDA 13.2)
## Bug type
Compile-time Error

## Affected component
CUB

## Bug description

Commit [1e3b0b4](https://github.com/NVIDIA/cccl/commit/1e3b0b4ce47acdf7fa87674535c4b1a5573aa6f9)
PR: https://github.com/NVIDIA/cccl/pull/6844

("Avoid use of cccl namespace macros in cub") removed `_CCCL_BEGIN_NAMESPACE_CUDA` /
`_CCCL_END_NAMESPACE_CUDA` macros and replaced them with inline explicit qualification.
The resulting code in `cub/device/device_transform.cuh:44` uses `::cuda::` as the
class-head name in a template specialization:

```cpp
struct ::cuda::proclaims_copyable_arguments<CUB_NS_QUALIFIER::detail::__return_constant<T>>
    : ::cuda::std::true_type
```

This is **ill-formed C++**. The C++ standard forbids a globally-qualified name (`::cuda::`)
as the class-head name of a struct/class definition or template specialization. The leading
`::` is valid in expression and type-specifier contexts, but **not in a class-head name**.

NVCC on x86 silently accepts this as a non-conforming extension, but **GCC on ARM/SBSA
treats it as a hard error**, breaking any downstream project that includes `<cub/cub.cuh>`
when building with CUDA 13.2+ on SBSA.

## Error message

```
cub/device/device_transform.cuh:44:134:
error: global qualification of class name is invalid before ':' token
   44 | struct ::cuda::proclaims_copyable_arguments<...> : ::cuda::std::true_type
      |                                                                          ^
error: expected '{' before ':' token
```

## Reproducer

Run the shell script below — it creates two CMake projects and a Dockerfile, then
launches `docker build` using the exact image that ships the buggy headers.
The good project compiles cleanly; the bad project fails with the GCC error above.

```bash
#!/bin/bash
set -e

rm -rf /tmp/cccl_reproducer_bad /tmp/cccl_reproducer_good
mkdir -p /tmp/cccl_reproducer_bad /tmp/cccl_reproducer_good

cat << 'REPRODUCER_BAD_EOF' > /tmp/cccl_reproducer_bad/reproducer.cpp
#include <type_traits>

namespace cuda {
    template <typename T>
    struct proclaims_copyable_arguments : std::false_type {};
}
namespace detail {
    template <typename T> struct __return_constant {};
}

// ill-formed: leading :: forbidden in class-head name
// mirrors cub/device/device_transform.cuh:44 after commit 1e3b0b4
template <typename T>
struct ::cuda::proclaims_copyable_arguments<detail::__return_constant<T>>
    : std::true_type {};

int main() {}
REPRODUCER_BAD_EOF

cat << 'REPRODUCER_GOOD_EOF' > /tmp/cccl_reproducer_good/reproducer.cpp
#include <type_traits>

namespace cuda {
    template <typename T>
    struct proclaims_copyable_arguments : std::false_type {};
}
namespace detail {
    template <typename T> struct __return_constant {};
}

// correct: relative qualification in class-head name
// one-character fix: remove leading :: from class-head name
template <typename T>
struct cuda::proclaims_copyable_arguments<detail::__return_constant<T>>
    : std::true_type {};

int main() {}
REPRODUCER_GOOD_EOF

cat << 'CMAKE_EOF' > /tmp/cccl_reproducer_bad/CMakeLists.txt
cmake_minimum_required(VERSION 3.20)
project(cccl_reproducer_bad CXX)
set(CMAKE_CXX_STANDARD 17)
set(CMAKE_CXX_STANDARD_REQUIRED ON)
message(STATUS "Compiler: ${CMAKE_CXX_COMPILER_ID} ${CMAKE_CXX_COMPILER_VERSION}")
find_package(CUDA REQUIRED)
add_executable(reproducer reproducer.cpp)
CMAKE_EOF

cat << 'CMAKE_GOOD_EOF' > /tmp/cccl_reproducer_good/CMakeLists.txt
cmake_minimum_required(VERSION 3.20)
project(cccl_reproducer_good CXX)
set(CMAKE_CXX_STANDARD 17)
set(CMAKE_CXX_STANDARD_REQUIRED ON)
message(STATUS "Compiler: ${CMAKE_CXX_COMPILER_ID} ${CMAKE_CXX_COMPILER_VERSION}")
find_package(CUDA REQUIRED)
add_executable(reproducer reproducer.cpp)
CMAKE_GOOD_EOF

cat > /tmp/Dockerfile.cccl_reproducer << 'DOCKERFILE_EOF'
FROM nvcr.io/nvidia/cuda-dl-base:26.04-cuda13.2-devel-ubuntu24.04

COPY ./cccl_reproducer_good /cccl_reproducer_good
COPY ./cccl_reproducer_bad  /cccl_reproducer_bad

RUN apt-get update -qq && apt-get install -y cmake build-essential

RUN echo -e "[32m ==Device Transform Cuh:44-50== [0m"     && echo -e "[33m $(sed -n '40,50p' /usr/local/cuda/include/cccl/cub/device/device_transform.cuh) [0m"

WORKDIR /cccl_reproducer_good
RUN cmake -B /cccl_build_good -S /cccl_reproducer_good && cmake --build /cccl_build_good

WORKDIR /cccl_reproducer_bad
RUN cmake -B /cccl_build_bad -S /cccl_reproducer_bad && cmake --build /cccl_build_bad
DOCKERFILE_EOF

docker build --progress=plain -t cccl_reproducer -f /tmp/Dockerfile.cccl_reproducer /tmp
```

## Suggested fix

```diff
- struct ::cuda::proclaims_copyable_arguments<CUB_NS_QUALIFIER::detail::__return_constant<T>>
+ struct cuda::proclaims_copyable_arguments<CUB_NS_QUALIFIER::detail::__return_constant<T>>
      : ::cuda::std::true_type
```

Same fix applies to any other class-head names introduced by the same commit
that use `::cuda::` or `::cuda::std::` as the leading qualifier.

## Environment

| | |
|---|---|
| CCCL version | Introduced in commit 1e3b0b4, present in CUDA 13.2 |
| Affected file | `cub/device/device_transform.cuh:44` |
| Failing compiler | GCC 13+  |
| Reproducer image | `nvcr.io/nvidia/cuda-dl-base:26.04-cuda13.2-devel-ubuntu24.04` |

lahwaacz:
## Description

Using `struct ::cuda::proclaims_copyable_arguments<...>` (global qualification after the struct keyword) is rejected by GCC up to 15.2:

    error: global qualification of class name is invalid before ‘:’ token

Other compilers such as Clang accept this. When compiling through nvcc, cudafe++ rewrites the code into valid forms before passing to the host compiler, so this was never caught in CI. However, downstream projects that include CCCL headers directly in C++ translation units compiled by GCC hit build failures.

Refs: microsoft/onnxruntime#28023

Small reproducer: https://godbolt.org/z/ssjavqveT

Fixes: #8833

## Checklist

### DIFF
--- a/cub/cub/device/device_transform.cuh
+++ b/cub/cub/device/device_transform.cuh
@@ -42,9 +42,11 @@ struct __return_constant
 } // namespace detail
 CUB_NAMESPACE_END
 
+namespace cuda {
 template <typename T>
-struct cuda::proclaims_copyable_arguments<CUB_NS_QUALIFIER::detail::__return_constant<T>> : ::cuda::std::true_type
+struct proclaims_copyable_arguments<CUB_NS_QUALIFIER::detail::__return_constant<T>> : ::cuda::std::true_type
 {};
+} // namespace cuda
 
 CUB_NAMESPACE_BEGIN
 //! DeviceTransform provides device-wide, parallel operations for transforming elements tuple-wise from multiple input
--- a/cub/cub/device/device_transform.cuh
+++ b/cub/cub/device/device_transform.cuh
@@ -43,7 +43,7 @@ struct __return_constant
 CUB_NAMESPACE_END
 
 template <typename T>
-struct ::cuda::proclaims_copyable_arguments<CUB_NS_QUALIFIER::detail::__return_constant<T>> : ::cuda::std::true_type
+struct cuda::proclaims_copyable_arguments<CUB_NS_QUALIFIER::detail::__return_constant<T>> : ::cuda::std::true_type
 {};
 
 CUB_NAMESPACE_BEGIN
--- a/cub/cub/device/device_transform.cuh
+++ b/cub/cub/device/device_transform.cuh
@@ -42,7 +42,8 @@ struct __return_constant
 } // namespace detail
 CUB_NAMESPACE_END
 
-namespace cuda {
+namespace cuda
+{
 template <typename T>
 struct proclaims_copyable_arguments<CUB_NS_QUALIFIER::detail::__return_constant<T>> : ::cuda::std::true_type
 {};
--- a/cub/cub/device/dispatch/tuning/tuning_transform.cuh
+++ b/cub/cub/device/dispatch/tuning/tuning_transform.cuh
@@ -42,10 +42,10 @@ struct always_true_predicate
 } // namespace detail::transform
 CUB_NAMESPACE_END
 
-namespace cuda {
+namespace cuda
+{
 template <>
-struct proclaims_copyable_arguments<CUB_NS_QUALIFIER::detail::transform::always_true_predicate>
-    : ::cuda::std::true_type
+struct proclaims_copyable_arguments<CUB_NS_QUALIFIER::detail::transform::always_true_predicate> : ::cuda::std::true_type
 {};
 } // namespace cuda
 
--- a/cub/cub/device/dispatch/tuning/tuning_transform.cuh
+++ b/cub/cub/device/dispatch/tuning/tuning_transform.cuh
@@ -43,7 +43,7 @@ struct always_true_predicate
 CUB_NAMESPACE_END
 
 template <>
-struct ::cuda::proclaims_copyable_arguments<CUB_NS_QUALIFIER::detail::transform::always_true_predicate>
+struct cuda::proclaims_copyable_arguments<CUB_NS_QUALIFIER::detail::transform::always_true_predicate>
     : ::cuda::std::true_type
 {};
 
--- a/cub/cub/device/dispatch/tuning/tuning_transform.cuh
+++ b/cub/cub/device/dispatch/tuning/tuning_transform.cuh
@@ -42,10 +42,12 @@ struct always_true_predicate
 } // namespace detail::transform
 CUB_NAMESPACE_END
 
+namespace cuda {
 template <>
-struct cuda::proclaims_copyable_arguments<CUB_NS_QUALIFIER::detail::transform::always_true_predicate>
+struct proclaims_copyable_arguments<CUB_NS_QUALIFIER::detail::transform::always_true_predicate>
     : ::cuda::std::true_type
 {};
+} // namespace cuda
 
 CUB_NAMESPACE_BEGIN
 namespace detail::transform


cub/cub/device/dispatch/tuning/tuning_transform.cuh
```diff
@@ -43,7 +43,7 @@ struct always_true_predicate
 CUB_NAMESPACE_END
 
 template <>
-struct ::cuda::proclaims_copyable_arguments<CUB_NS_QUALIFIER::detail::transform::always_true_predicate>
+struct cuda::proclaims_copyable_arguments<CUB_NS_QUALIFIER::detail::transform::always_true_predicate>
     : ::cuda::std::true_type
 {};
```
miscco, line 49: Please wrap this in `_CCCL_BEGIN_NAMESPACE_CUDA` and `_CCCL_END_NAMESPACE_CUDA`

We *must* not use unqualifies `cuda::` in product code due to user defined `namespace cuda`

miscco (changes requested):

cub/cub/device/dispatch/tuning/tuning_transform.cuh
lahwaacz: Hmm, so https://github.com/NVIDIA/cccl/commit/1e3b0b4ce47acdf7fa87674535c4b1a5573aa6f9 should be reverted?

lahwaacz: cc @davebayer

miscco: Honestly yes, neither CUB not Thrust claim STL level conformance / hardening, so if a user defines an unfortunate macro they are on their own.

bernhardmgruber:
I think we have been bouncing back and forth between syntax here.

> Fix invalid C++ syntax

AFAIK, the syntax in CCCL is conforming and you are observing a GCC bug: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=66892

However, we cannot change GCC so I am fine with any resolution that solves the works around the issue with GCC.

cub/cub/device/dispatch/tuning/tuning_transform.cuh
davebayer: No, we cannot use `_CCCL_(BEGIN|END)_NAMESPACE_CUDA` in cub, because of libcu++'s prologue/epilogue checks that break for pstl. There are 2 options:
1. we can use `namespace cuda { ... }` directly
2. we can remove the leading `::`

I would go for option 1.

miscco: Lets try version 1, there as a bug in older NVCC about specializing variables in inline namespaces, but lets see whether it has been fixed

cub/cub/device/dispatch/tuning/tuning_transform.cuh
```diff
@@ -43,7 +43,7 @@ struct always_true_predicate
 CUB_NAMESPACE_END
 
 template <>
```
davebayer, line 47:
```suggestion
namespace cuda {
template <>
```

cub/cub/device/dispatch/tuning/tuning_transform.cuh
```diff
@@ -43,7 +43,7 @@ struct always_true_predicate
...
-struct ::cuda::proclaims_copyable_arguments<CUB_NS_QUALIFIER::detail::transform::always_true_predicate>
+struct cuda::proclaims_copyable_arguments<CUB_NS_QUALIFIER::detail::transform::always_true_predicate>
```
davebayer:
```suggestion
struct proclaims_copyable_arguments<CUB_NS_QUALIFIER::detail::transform::always_true_predicate>
```

cub/cub/device/dispatch/tuning/tuning_transform.cuh
davebayer, line 49:
```suggestion
{};
} // namespace cuda
```

cub/cub/device/device_transform.cuh
```diff
@@ -43,7 +43,7 @@ struct __return_constant
 CUB_NAMESPACE_END
 
 template <typename T>
```
davebayer, line 47:
```suggestion
namespace cuda {
template <typename T>
```

cub/cub/device/device_transform.cuh
```diff
@@ -43,7 +43,7 @@ struct __return_constant
...
-struct ::cuda::proclaims_copyable_arguments<CUB_NS_QUALIFIER::detail::__return_constant<T>> : ::cuda::std::true_type
+struct cuda::proclaims_copyable_arguments<CUB_NS_QUALIFIER::detail::__return_constant<T>> : ::cuda::std::true_type
```
davebayer:
```suggestion
struct proclaims_copyable_arguments<CUB_NS_QUALIFIER::detail::__return_constant<T>> : ::cuda::std::true_type
```

cub/cub/device/device_transform.cuh
```diff
@@ -43,7 +43,7 @@ struct __return_constant
...
 {};
```
davebayer, line 49:
```suggestion
{};
} // namespace cuda
```

vacu9708:
I confirmed this code works.

```c++
namespace cuda {
template <typename T>
struct proclaims_copyable_arguments<CUB_NS_QUALIFIER::detail::__return_constant<T>> : ::cuda::std::true_type
{};
} // namespace cuda
```

bernhardmgruber (approved):

bernhardmgruber:
pre-commit.ci autofix

bernhardmgruber (approved):

bernhardmgruber:
@miscco please re-review.

davebayer (approved):

miscco (approved):

PR #8775, opened 2026-04-30
`cuda::std::simd` Optimize Arithmetic Floating-point Operations

fbusato:
## Description

This PR introduces the following optimizations and checks:

- **`HADD2`, `HMUL2`, `HFMA` for Half and Bfloat16**: There are no single-element variants, so here we only check that all operators relying on them generate the expected number of instructions of that type.
- `F32x2` Blackwell SM100: Ensure that the following operations are mapped to `F32x2` instructions:

  - Plus `+`
  - Minus `-`
  - Unary minus `(-)`
  - Increment `++`
  - Decrement `--`


libcudacxx/include/cuda/std/__simd/specializations/fixed_size_storage.h
```diff
@@ -0,0 +1,68 @@
...
+template <typename _Tp, __simd_size_type _Np>
+struct __simd_storage<_Tp, __fixed_size<_Np>>
+{
+  using value_type = _Tp;
+
+  _Tp __data[_Np]{};
+
+  _CCCL_HIDE_FROM_ABI constexpr __simd_storage()                                 = default;
+  _CCCL_HIDE_FROM_ABI constexpr __simd_storage(const __simd_storage&)            = default;
+  _CCCL_HIDE_FROM_ABI constexpr __simd_storage& operator=(const __simd_storage&) = default;
```
Jacobfaib: You can omit these, no? You don't have any other special functions defined, and you have inline initializers, so no need to define these. Unless you specifically want to prohibit move ctors and assignment.

libcudacxx/include/cuda/std/__simd/specializations/fixed_size_vec.h
```diff
@@ -104,6 +72,20 @@ struct __simd_operations<_Tp, __fixed_size<_Np>>
...
   {
+#if _CCCL_HAS_SIMD_F32X2()
+    _CCCL_IF_NOT_CONSTEVAL_DEFAULT
+    {
+      if constexpr (is_same_v<_Tp, float> && _Np >= 2)
+      {
+        // clang-format off
+        NV_IF_TARGET(NV_IS_EXACTLY_SM_100,
+                     (constexpr _SimdStorage __one = __broadcast(1.0f);
+                      __s                      = ::cuda::std::simd::__plus_f32x2(__s, __one);
```
Jacobfaib:
```suggestion
                      __s                          = ::cuda::std::simd::__plus_f32x2(__s, __one);
```
(I think, it's hard to tell if this lines up from the suggestion box)

libcudacxx/include/cuda/std/__simd/specializations/fp32x2_intrinsics.h
```diff
@@ -0,0 +1,255 @@
...
+
+#if defined(_CCCL_IMPLICIT_SYSTEM_HEADER_GCC)
+#  pragma GCC system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_CLANG)
+#  pragma clang system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_MSVC)
+#  pragma system_header
+#endif // no system header
+
+#if _CCCL_CUDA_COMPILER(NVCC, >=, 12, 8) || (__cccl_ptx_isa >= 860ULL)
```
Jacobfaib: You reuse these conditions (nvcc-12.8+ and PTX ISA 8.6+) quite a lot below, consider making them separate defines that you then `#undef` at the end of the file:
```c++
#define _CCCL_SIMD_HAVE_NATIVE_FOO() _CCCL_CUDA_COMPILER(NVCC, >=, 12, 8)
#define _CCCL_SIMD_HAVE_PTX_8_6() (__cccl_ptx_isa >= 860ULL)
```
Guards against typos and makes the `ifdef`s easier to read when you add more conditions.

libcudacxx/include/cuda/std/__simd/specializations/fp32x2_intrinsics.h
```diff
@@ -0,0 +1,255 @@
...
+#  define _CCCL_HAS_SIMD_F32X2() 1
+#else
+#  define _CCCL_HAS_SIMD_F32X2() 0
+#endif // _CCCL_CUDA_COMPILER(NVCC, >=, 12, 8) || (__cccl_ptx_isa >= 860ULL)
+
+#if _CCCL_HAS_SIMD_F32X2()
```
Jacobfaib, line 27: Does the _entire_ file need to be gated behind this? I notice that only a few functions specifically need it, and for those you can just add an extra `#else` clause that does the naive algorithm.

This saves downstream users also needing to gate their code behind `_CCCL_HAS_SIMD_F32X2()` as the symbols always exist.

libcudacxx/test/atomic_codegen/dump_and_check.bash
```diff
@@ -1,9 +1,11 @@
...
-## Usage: dump_and_check test.a test.cu PREFIX
-input_archive="$1"
-input_testfile="$2"
-input_prefix="$3"
+## Usage: dump_and_check test.a test.cu PREFIXES [cuobjdump-mode]
+input_archive=$1
+input_testfile=$2
+input_prefix=$3
+dump_mode=${4:---dump-ptx}
+filecheck=${FILECHECK:-FileCheck}
```
Jacobfaib: These should be quoted (and ideally `{}`-ed as well) to avoid argument splitting

libcudacxx/test/simd_codegen/CMakeLists.txt
Jacobfaib, line 1: Do cmake files not have a license comment?

libcudacxx/test/simd_codegen/CMakeLists.txt
```diff
@@ -0,0 +1,72 @@
...
+find_program(cuobjdump "cuobjdump" REQUIRED)
+find_program(bash "bash" REQUIRED)
+
+set(libcudacxx_simd_codegen_tests)
+if (NOT "NVHPC" STREQUAL "${CMAKE_CXX_COMPILER_ID}")
+  file(GLOB libcudacxx_simd_codegen_tests "*.cu")
+endif()
+
+set(simd_codegen_cuda_archs 80 90)
+if (CMAKE_CUDA_COMPILER_VERSION VERSION_GREATER_EQUAL 12.9)
```
Jacobfaib: Don't assume `nvcc` here. Can also be `clang-cuda`

libcudacxx/test/simd_codegen/decrement_f32x2.cu
Jacobfaib, line 1: License header?

libcudacxx/test/simd_codegen/decrement_f32x2.cu
```diff
@@ -0,0 +1,31 @@
+#include <cuda/std/__simd_> // IWYU pragma: keep
+#include <cuda/std/array>
+
+#if _CCCL_HAS_SIMD_F32X2()
```
Jacobfaib: You are already skipping these tests in cmake no? The
```cmake
 if (test_name MATCHES "_f32x2$" AND NOT arch EQUAL 100)
      continue()
    endif()
```
bit

libcudacxx/test/simd_codegen/CMakeLists.txt
fbusato: I don't think checking the SASS code for `clang-cuda` is critical

libcudacxx/test/simd_codegen/decrement_f32x2.cu
fbusato: I followed `atomic_codegen`. I defer it to @wmaxey

libcudacxx/include/cuda/std/__simd/specializations/fp32x2_intrinsics.h
fbusato: This is intentional. Users should never use this file

libcudacxx/test/simd_codegen/CMakeLists.txt
fbusato: `atomic_codegen` doesn't have a license, but yes better to add it

libcudacxx/include/cuda/std/__simd/specializations/fixed_size_float_vec.h
```diff
@@ -0,0 +1,128 @@
...
+  {
+#if _CCCL_HAS_SIMD_F32X2()
+    _CCCL_IF_NOT_CONSTEVAL_DEFAULT
+    {
+      // clang-format off
+      NV_IF_TARGET(NV_IS_EXACTLY_SM_100,
+                   (constexpr _SimdStorage __one = __base::__broadcast(1.0f);
+                    __s                          = ::cuda::std::simd::__plus_f32x2(__s, __one);
+                    return;))
+      // clang-format on
```
davebayer: Nitpick: I'd rather see this than `// clang-format off`
```suggestion
      NV_IF_TARGET(NV_IS_EXACTLY_SM_100,
                   ({
                       constexpr _SimdStorage __one = __base::__broadcast(1.0f);
                       __s                          = ::cuda::std::simd::__plus_f32x2(__s, __one);
                       return;
                    }))
```

libcudacxx/include/cuda/std/__simd/specializations/fp32x2_intrinsics.h
```diff
@@ -0,0 +1,254 @@
...
+  {
+    ::cuda::std::simd::__add_f32x2(
+      __lhs.__data[__i],
+      __lhs.__data[__i + 1],
+      __rhs.__data[__i],
+      __rhs.__data[__i + 1],
+      __result.__data[__i],
+      __result.__data[__i + 1]);
+  }
+  if (_Np % 2 != 0)
```
davebayer:
```suggestion
  if constexpr (_Np % 2 != 0)
```

libcudacxx/include/cuda/std/__simd/specializations/fp32x2_intrinsics.h
```diff
@@ -0,0 +1,254 @@
...
+
+_CCCL_BEGIN_NAMESPACE_CUDA_STD_SIMD
+
+_CCCL_DEVICE_API inline void __add_f32x2(
+  const float __lhs1,
+  const float __lhs2,
+  const float __rhs1,
+  const float __rhs2,
+  float& __result1,
+  float& __result2) noexcept
```
davebayer: Can't the signature rather be this to avoid references:
```suggestion
_CCCL_DEVICE_API inline ::float2 __add_f32x2(
  const ::float2 __lhs,
  const ::float2 __rhs) noexcept
```

libcudacxx/include/cuda/std/__simd/specializations/fp32x2_intrinsics.h
```diff
@@ -0,0 +1,254 @@
...
+
+#if defined(_CCCL_IMPLICIT_SYSTEM_HEADER_GCC)
+#  pragma GCC system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_CLANG)
+#  pragma clang system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_MSVC)
+#  pragma system_header
+#endif // no system header
+
+#define _CCCL_HAS_SIMD_F32X2_INTRINSICS() (_CCCL_CUDACC_AT_LEAST(12, 8) && _CCCL_HAS_CTK())
```
davebayer:
```suggestion
#define _CCCL_HAS_SIMD_F32X2_INTRINSICS() _CCCL_CUDACC_AT_LEAST(12, 8)
```
Should be sufficient

libcudacxx/include/cuda/std/__simd/specializations/fixed_size_float_vec.h
miscco: strong + 1

libcudacxx/include/cuda/std/__simd/specializations/fixed_size_vec.h
```diff
@@ -354,6 +327,13 @@ struct __simd_operations<_Tp, __fixed_size<_Np>>
     return __result;
   }
 };
+
+// Default path (no optimizations)
+template <typename _Tp, __simd_size_type _Np>
+struct __simd_operations<_Tp, __fixed_size<_Np>, enable_if_t<!__is_fixed_size_float_v<_Tp, _Np>>>
+    : __fixed_size_operations<_Tp, _Np>
+{};
```
miscco, line 330: Why do we need to explicitly specify the default path with a special type? SFINAE should pick the specialization if applicable

libcudacxx/include/cuda/std/__simd/specializations/fp32x2_intrinsics.h
miscco, line 1: Question: What about float4 / double4

Same for double / `__half` / `__nv_bfloat16` I believe there should also be optimizations for those.

Do we offer intrinsics for those?

miscco (changes requested):
I would like to push this PR to the back of the review queue

This is effectively doing the compilers job trying to vectorize operations on a single type.

We should talk with the backend team to see what it the compiler needs to recognize those operations out of the box and not handwrite the optimized asm

libcudacxx/include/cuda/std/__simd/specializations/fp32x2_intrinsics.h
fbusato: std::simd only support "native" types (we also support extended types). So ` float4 / double4` are out of scope, they are `basic_vec<float, 4>`. 

> Same for double / __half / __nv_bfloat16 I believe there should also be optimizations for those.

HW doesn't have DFMAx2 for double. While for `__half` / `__nv_bfloat16` the compiler already vectorize them because there are no scalar version of them (i.e. there is not `HADD`, there is `HADD2` instead). This is easy because two __half` / `__nv_bfloat16` requires a single register (32-bit)

fbusato:
> I would like to push this PR to the back of the review queue
>
> This is effectively doing the compilers job trying to vectorize operations on a single type.
>
> We should talk with the backend team to see what it the compiler needs to recognize those operations out of the box and not handwrite the optimized asm

Discussed offline. The proposed optimization makes sense. Currently, the compiler doesn't perform it automatically

libcudacxx/include/cuda/std/__simd/specializations/fp32x2_intrinsics.h
fbusato: nope, clang doesn't work with new instrinsics

libcudacxx/include/cuda/std/__simd/specializations/fixed_size_vec.h
fbusato: SFINAE here is required otherwise `basic_vec<float, 2>` would match both structures. On the other hand, I can avoid SFINAE and add a new enum parameter

libcudacxx/include/cuda/std/__simd/specializations/fixed_size_float_vec.h
```diff
@@ -0,0 +1,128 @@
...
+template <__simd_size_type _Np>
+struct __simd_operations<float, __fixed_size<_Np>, __simd_operations_kind::__fixed_size_float>
+    : __fixed_size_operations<float, _Np>
+{
+  using __base       = __fixed_size_operations<float, _Np>;
+  using _SimdStorage = __simd_storage<float, __fixed_size<_Np>>;
+
+  _CCCL_API static constexpr void __increment(_SimdStorage& __s) noexcept
+  {
+#if _CCCL_HAS_SIMD_F32X2()
```
miscco: Question: This whole specialization is only useful when `_CCCL_HAS_SIMD_F32X2()` is true.

Should we just guard the whole thing on it?

libcudacxx/include/cuda/std/__simd/specializations/fp32x2_intrinsics.h
```diff
@@ -0,0 +1,208 @@
...
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_CLANG)
+#  pragma clang system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_MSVC)
+#  pragma system_header
+#endif // no system header
+
+#define _CCCL_HAS_SIMD_F32X2_INTRINSICS() (_CCCL_CUDACC_AT_LEAST(12, 8) && _CCCL_HAS_CTK() && !_CCCL_COMPILER(CLANG))
+#define _CCCL_HAS_SIMD_F32X2_PTX()        (__cccl_ptx_isa >= 860ULL)
+
+#define _CCCL_HAS_SIMD_F32X2() (_CCCL_HAS_SIMD_F32X2_INTRINSICS() || _CCCL_HAS_SIMD_F32X2_PTX())
```
miscco: I believe these definitions belong into `libcudacxx/include/cuda/std/__internal/features.h`

libcudacxx/test/simd_codegen/minus_f32x2.cu
```diff
@@ -0,0 +1,39 @@
...
+  out[0] = result[0];
+  out[1] = result[1];
+  out[2] = result[2];
+  out[3] = result[3];
+}
+
+/*
+
+; SMXX-LABEL: {{[[:space:]]*}}Function : test_operator_minus_f32_4
+; SM100: {{.*FADD2.*}}
```
miscco, line 36: Thanks a lot for adding the codegen tests, love seeing more of them

miscco (approved):
I have some minor nitpicks:

Please also add a ling to the nvbug, so that we know when to disable the specialization

libcudacxx/include/cuda/std/__simd/specializations/fixed_size_float_vec.h
fbusato: that's a good point. I need to check that everything compiler correctly

PR #8777, opened 2026-05-01
[libcu++] Add arch_traits for sm_5x

pciolkosz:
We still support 12.X CTK and it supports sm_5x. We should add traits for those architectures until we drop 12.X support


Jacobfaib (approved):

davebayer (approved):

bernhardmgruber:
> We still support 12.X CTK and it supports sm_5x. We should add traits for those architectures until we drop 12.X support

It's still important to point out that CCCL does not officially support anything below sm_75. We are just not trying to break support for older architectures. This PR is not a sign that we are starting to properly support architectures < sm_75 again, we are just trying to cause less noise when we are refactoring several internal facilities related to `cuda::compute_capability` and `cuda::arch_id`.

PR #8792, opened 2026-05-04
[cudax] Add support for generic thread groups within warp and cluster

davebayer:


cudax/test/group/cooperative_algorithm.cu
```diff
@@ -116,46 +116,60 @@ __device__ cuda::std::optional<T> sum(cudax::this_cluster<Hierarchy> group, T (&
...
-  using MappingResult = typename Group::__mapping_result_type;
-
-  constexpr auto ngroups = MappingResult::static_group_count();
-  static_assert(ngroups != cuda::std::dynamic_extent, "group count must be statically known");
-
-  __shared__ T group_sums[ngroups];
+  using Level = typename Group::level_type;
 
-  if (!Unit{}.is_part_of(group))
+  if constexpr (cuda::std::is_same_v<Level, cuda::warp_level>)
```
miscco, line 124: Question: Does this also handle cluster level fine or was the comment outdated

cudax/test/group/cooperative_algorithm.cu
```diff
@@ -116,46 +116,60 @@ __device__ cuda::std::optional<T> sum(cudax::this_cluster<Hierarchy> group, T (&
...
   {
-    return cuda::std::nullopt;
+    const auto result_unit = sum(cudax::this_thread{group.hierarchy()}, array);
+
+    // todo(dabayer): Implement fallback for cc < 80.
+    T result;
+    NV_IF_TARGET(NV_PROVIDES_SM_80,
+                 ({ result = __reduce_add_sync(group.__synchronizer_instance().__lane_mask(), result_unit.value()); }))
+    return (cuda::gpu_thread.is_root_rank(group)) ? cuda::std::optional{result} : cuda::std::nullopt;
```
miscco, line 132: Question: That comment suggests the code path is not valid for SM < 80, we should at least assert that to ensure we do not forget it once this goes out of experimentall

cudax/test/group/cooperative_algorithm.cu
```diff
@@ -116,46 +116,60 @@ __device__ cuda::std::optional<T> sum(cudax::this_cluster<Hierarchy> group, T (&
...
-  if (!Unit{}.is_part_of(group))
+  if constexpr (cuda::std::is_same_v<Level, cuda::warp_level>)
   {
-    return cuda::std::nullopt;
+    const auto result_unit = sum(cudax::this_thread{group.hierarchy()}, array);
+
+    // todo(dabayer): Implement fallback for cc < 80.
+    T result;
+    NV_IF_TARGET(NV_PROVIDES_SM_80,
+                 ({ result = __reduce_add_sync(group.__synchronizer_instance().__lane_mask(), result_unit.value()); }))
```
miscco, line 131: Question: Cannot we use `ThreadReduce` here just fine?

It should use the `__reduce_add_sync` optimization when applicable

cudax/test/group/cooperative_algorithm.cu
```diff
@@ -116,46 +116,60 @@ __device__ cuda::std::optional<T> sum(cudax::this_cluster<Hierarchy> group, T (&
...
-  // todo(dabayer): Replace by group.rank(level) once this query is available.
-  const auto group_rank = group.__mapping_result().group_rank();
+    constexpr auto ngroups = MappingResult::static_group_count();
+    static_assert(ngroups != cuda::std::dynamic_extent, "group count must be statically known");
 
-  if (cuda::gpu_thread.is_root_rank(group))
-  {
-    group_sums[group_rank] = 0;
-  }
+    __shared__ T group_sums[ngroups];
```
pciolkosz, line 142: Hmm, I was thinking if in actual interface shared memory should come from the user instead, for example here if only one group calls this we waste shared memory

davebayer: This is correct, but I would say that this is outside of this epic's scope, so I just went with statically shared memory allocations inside the algorithms

davebayer:
This PR is no longer relevant after `cudax::coop::reduce` was implemented

Issue #8318, opened 2026-04-07
[FEA]: Prototype `DeviceSegmentedReduce`-like algorithm for groups


davebayer:
Create a prototype of planned `cub::DeviceSegmentedReduce` algorithm in groups test.

PR #8793, opened 2026-05-04
[cudax] Prototype of segmented reduce algorithm using groups

davebayer:
Implements #8318

### DIFF
--- a/cudax/test/group/segmented_algorithm.cu
+++ b/cudax/test/group/segmented_algorithm.cu
@@ -1,0 +1,114 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of CUDA Experimental in CUDA C++ Core Libraries,
+// under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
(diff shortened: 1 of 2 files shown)


cudax/test/group/segmented_algorithm.cu
```diff
@@ -0,0 +1,114 @@
...
+
+C2H_TEST("Segmented algorithm", "[group]")
+{
+  const auto device = cuda::devices[0];
+
+  const cuda::stream stream{device};
+
+  auto in  = cuda::make_device_buffer<int>(stream, device, 1024, cuda::no_init);
+  auto out = cuda::make_device_buffer<int>(stream, device, 8, cuda::no_init);
+  cuda::std::fill(cuda::execution::gpu.with(stream), in.begin(), in.end(), 1);
```
pciolkosz: Why separate fill instead of `make_device_buffer` with init?

pciolkosz (approved):

cudax/test/group/segmented_algorithm.cu
davebayer: Aaah, that's because I filled the buffer in a different way before

PR #8795, opened 2026-05-04
[HostJit] Do not use `std::once_flag` in hosted mode

miscco:
We cannot use the host STL `::std::once_flag` because there is none. We also cannot emulate it because for that we would need `clock` and other OS facilities.

We do the next best thing and just check whether the members are already set.

This wont work cleanly in a multithreaded environement but for HostJit we are fine

### DIFF
--- a/libcudacxx/include/cuda/__device/physical_device.h
+++ b/libcudacxx/include/cuda/__device/physical_device.h
@@ -27,13 +27,14 @@
 #  include <cuda/__driver/driver_api.h>
 #  include <cuda/__fwd/devices.h>
 #  include <cuda/std/__cstddef/types.h>
-#  include <cuda/std/__host_stdlib/memory>
+#  include <cuda/std/__memory/unique_ptr.h>
+#  include <cuda/std/cassert>
 #  include <cuda/std/span>
 #  include <cuda/std/string_view>
 
-#  include <cassert>
-#  include <mutex>
-#  include <vector>
+#  if _CCCL_HOSTED()
+#    include <mutex>
+#  endif // _CCCL_HOSTED()
 
 #  include <cuda/std/__cccl/prologue.h>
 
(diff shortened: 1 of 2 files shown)


libcudacxx/include/cuda/__device/physical_device.h
```diff
@@ -47,21 +48,69 @@ _CCCL_BEGIN_NAMESPACE_CUDA
...
+  _CCCL_HOST_API void __set_name()
+  {
+    const auto __id = ::cuda::__driver::__cudevice_to_ordinal(__device_);
+    ::cuda::__driver::__deviceGetName(__name_, __max_name_length, __id);
+    __name_length_ = ::cuda::std::char_traits<char>::length(__name_);
+  }
+
+  _CCCL_HOST_API void __set_peers()
+  {
+    const auto __count = static_cast<int>(::cuda::__physical_devices().size());
```
Jacobfaib, line 83: Why convert to `int` here?

libcudacxx/include/cuda/__device/physical_device.h
```diff
@@ -47,21 +48,69 @@ _CCCL_BEGIN_NAMESPACE_CUDA
...
+    const auto __id    = ::cuda::__driver::__cudevice_to_ordinal(__device_);
+
+    // This overallocates, but given that we are talking about `device_ref` this is fine
+    __peers_.reset(static_cast<device_ref*>(::operator new[](sizeof(device_ref) * __count)));
```
Jacobfaib, line 87: Store the new array in a temporary `unique_ptr` and reset at the end to preserve a strong exception guarantee. If `has_peer_access_to()` (or any other line) throws, then we will leave `__peers_` in a half-constructed form.

Also, do we need to fully qualify `new[]` here? Can we not just do

```suggestion
    __peers_.reset(new device_ref *[__count]);
```

libcudacxx/include/cuda/__device/physical_device.h
```diff
@@ -47,21 +48,69 @@ _CCCL_BEGIN_NAMESPACE_CUDA
...
+    size_t __num_peers = 0;
```
Jacobfaib, line 88: Nit: this is very similar to the member, `__num_peers_`. Suggest renaming to `__n_peers` or something like that to avoid typo bugs

libcudacxx/include/cuda/__device/physical_device.h
```diff
@@ -79,59 +128,54 @@ class __physical_device
   //! @return A reference to the primary context for this device.
   [[nodiscard]] _CCCL_HOST_API ::CUcontext __primary_context()
   {
+#  if _CCCL_HOSTED()
     ::std::call_once(__primary_ctx_once_flag_, [this]() {
       __primary_ctx_ = ::cuda::__driver::__primaryCtxRetain(__device_);
     });
+#  else // ^^^ _CCCL_HOSTED() ^^^ / vvv _CCCL_FREESTANDING() vvv
+    if (!__primary_ctx_)
```
Jacobfaib, line 136: You could `atomic_ref` and `compare_exchange_weak()` this.

libcudacxx/include/cuda/__memory_pool/memory_pool_base.h
```diff
@@ -37,6 +37,10 @@
 #  include <cuda/std/__host_stdlib/stdexcept>
 #  include <cuda/std/cstddef>
 
+#  if _CCCL_HOSTED()
+#    include <vector>
```
Jacobfaib, line 41: Do you need to surround uses of `std::vector` with this as well?

libcudacxx/include/cuda/__device/physical_device.h
miscco: Pre-existing, but I believe because ` ::CUdevice` is an alias for `int`

libcudacxx/include/cuda/__device/physical_device.h
miscco: We cannot because that requires `device_ref` to be default constructible

libcudacxx/include/cuda/__device/physical_device.h
miscco: Its actually more of a convention in constructors that we name the input the same as the member variable without the trailing `_`

libcudacxx/include/cuda/__device/physical_device.h
miscco: We cannot because hosted does not have that. If we had atomics I would have just implemented `once_flag`

libcudacxx/include/cuda/__memory_pool/memory_pool_base.h
miscco: It worked without any other changed so I would keep it as is

libcudacxx/include/cuda/__device/physical_device.h
miscco: Also there is nothing to preserve here, if the query of devices throws, the program ends

davebayer (approved):

libcudacxx/include/cuda/__device/physical_device.h
Jacobfaib: > Also there is nothing to preserve here, if the query of devices throws, the program ends

Not necessarily. These are all stored in static local variables in functions. There's no guarantee when such a function may be called, and whether the user is capable of handling the error, no?

miscco: I believe this is still save, we only assign `__num_peers_` at the end of the function, so worst case there will be no peers

bernhardmgruber:
Could we just rely on a function local `static` variable instead of a `once_flag`? What's the advantage of a `once_flag`?

miscco:
> Could we just rely on a function local `static` variable instead of a `once_flag`? What's the advantage of a `once_flag`?

Unfortunately not. The issue is that a function local static would be *per class* and not *per instance*

So for example the `__peers_` would be determined for the first device this is called on and the just copied for all other devices.

PR #8799, opened 2026-05-04
[libcu++] Replace `cuda::__bad_any_cast` with `std::bad_any_cast`

davebayer:
All of the host standard libraries come with `<any>` header, thus we should just throw `std::bad_any_cast` instead of the internal `cuda::__bad_any_cast`.

### DIFF
--- a/libcudacxx/include/cuda/__utility/__basic_any/basic_any_ref.h
+++ b/libcudacxx/include/cuda/__utility/__basic_any/basic_any_ref.h
@@ -29,6 +29,8 @@
 #include <cuda/__utility/__basic_any/rtti.h>
 #include <cuda/std/__concepts/concept_macros.h>
 #include <cuda/std/__concepts/same_as.h>
+#include <cuda/std/__exception/exception_macros.h>
+#include <cuda/std/__host_stdlib/any>
 #include <cuda/std/__type_traits/is_const.h>
 #include <cuda/std/__type_traits/maybe_const.h>
 #include <cuda/std/__type_traits/remove_const.h>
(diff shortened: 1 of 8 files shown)


libcudacxx/include/cuda/__utility/__basic_any/basic_any_ref.h
```diff
@@ -210,7 +212,7 @@ struct _CCCL_DECLSPEC_EMPTY_BASES __basic_any<__ireference<_Interface>>
     using __src_interface_t _CCCL_NODEBUG_ALIAS = typename ::cuda::std::remove_reference_t<_SrcCvAny>::interface_type;
     if (!__from.has_value())
     {
-      __throw_bad_any_cast();
+      _CCCL_THROW(std::bad_any_cast);
```
Jacobfaib:
```suggestion
      _CCCL_THROW(::std::bad_any_cast);
```
And below as well. I guess it _shouldn't_ make a difference as we don't have a `bad_any_cast`, but better to be defensive?

libcudacxx/include/cuda/std/__host_stdlib/any
```diff
@@ -0,0 +1,28 @@
...
+#if defined(_CCCL_IMPLICIT_SYSTEM_HEADER_GCC)
+#  pragma GCC system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_CLANG)
+#  pragma clang system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_MSVC)
+#  pragma system_header
+#endif // no system header
+
+#if _CCCL_HOSTED()
+#  include <any>
```
miscco, line 25: Did we check whether `<typeinfo>` would be smaller?

miscco (commented):
Question about which host header to include

PR #8801, opened 2026-05-04
[HostJit] Add tests for compiling CUB APIs with hostjit

miscco:
This adds 3 tests for different CUB APIs

### DIFF
--- a/c/parallel/test/freestanding/test_cub_device_adjacent_difference.cpp
+++ b/c/parallel/test/freestanding/test_cub_device_adjacent_difference.cpp
@@ -1,0 +1,137 @@
+#include <iostream>
+#include <numeric>
+
+#include <cuda_runtime.h>
+
+#include "test_util.h"
+#include <hostjit/config.hpp>
+#include <hostjit/jit_compiler.hpp>
+
+static const char* cuda_source = R"(
+#include <cuda_runtime.h>
+#include <cub/device/device_adjacent_difference.cuh>
+#include <cuda/std/functional>
+
+#ifdef _WIN32
+#define HOSTJIT_EXPORT __declspec(dllexport)
+#else
+#define HOSTJIT_EXPORT
+#endif
+
+extern "C" HOSTJIT_EXPORT int adjacentDifference(
+    const int* d_input, int* d_output, int num_items) {
+    void* d_temp_storage = nullptr;
+    size_t temp_storage_bytes = 0;
+
+    // Query temp storage
+    cudaError_t err = cub::DeviceAdjacentDifference::SubtractLeftCopy(
+        d_temp_storage, temp_storage_bytes, d_input, d_output, num_items,
+        cuda::std::minus<int>{});
+    if (err != cudaSuccess) return -1;
+
+    // Allocate temp storage
+    err = cudaMalloc(&d_temp_storage, temp_storage_bytes);
+    if (err != cudaSuccess) return -2;
+
+    // Run adjacent difference
+    err = cub::DeviceAdjacentDifference::SubtractLeftCopy(
+        d_temp_storage, temp_storage_bytes, d_input, d_output, num_items,
+        cuda::std::minus<int>{});
+    if (err != cudaSuccess) {
+        cudaFree(d_temp_storage);
+        return -3;
+    }
+
+    err = cudaDeviceSynchronize();
+    cudaFree(d_temp_storage);
+    return (err == cudaSuccess) ? 0 : -4;
+}
+)";
+
+int main()
+{
+  // Detect Clang/CUDA configuration from the build environment
+  auto config = hostjit::detectDefaultConfig();
+
+  hostjit::JITCompiler compiler(config);
+  if (!compiler.compile(cuda_source))
+  {
+    std::fprintf(stderr, "HostJIT compilation failed:\n%s\n", compiler.getLastError().c_str());
+    return 1;
+  }
+
+  auto adjacentDiff = compiler.getFunction<int (*)(const int*, int*, int)>("adjacentDifference");
+  if (!adjacentDiff)
+  {
+    std::cerr << "Failed to get function: " << compiler.getLastError() << "\n";
+    return 1;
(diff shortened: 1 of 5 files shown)


c/parallel/test/freestanding/test_cub_device_adjacent_difference.cpp
```diff
@@ -0,0 +1,147 @@
...
+#define CUDA_CHECK(call)                                                                                       \
+  do                                                                                                           \
+  {                                                                                                            \
+    cudaError_t err = call;                                                                                    \
+    if (err != cudaSuccess)                                                                                    \
+    {                                                                                                          \
+      std::cerr << "CUDA error at " << __FILE__ << ":" << __LINE__ << ": " << cudaGetErrorString(err) << "\n"; \
+      return 1;                                                                                                \
+    }                                                                                                          \
+  } while (0)
```
bernhardmgruber: Suggestion: put this into a common test utility header

c/parallel/test/freestanding/test_cub_device_adjacent_difference.cpp
```diff
@@ -0,0 +1,147 @@
...
+  CUDA_CHECK(cudaMalloc(&d_output, N * sizeof(int)));
+  CUDA_CHECK(cudaMemcpy(d_input, h_input.data(), N * sizeof(int), cudaMemcpyHostToDevice));
+
+  std::cout << "Computing adjacent differences of " << N << " integers...\n";
+  int status = adjacentDiff(d_input, d_output, N);
+  if (status != 0)
+  {
+    std::cerr << "adjacentDifference failed with status: " << status << "\n";
+    CUDA_CHECK(cudaFree(d_input));
+    CUDA_CHECK(cudaFree(d_output));
```
bernhardmgruber, line 95: Suggestion: I think we cannot use Thrust vectors here because it's a plain C++ file. Can we just use a `cuda::buffer` here instead of the raw allocation? And if not, can we just add a `unique_ptr` using `cudaFree` as deleter in a central test utility header?

c/parallel/test/freestanding/test_cub_device_adjacent_difference.cpp
```diff
@@ -0,0 +1,147 @@
...
+    std::cout << (i ? ", " : "") << h_output[i];
+  }
+  std::cout << "]\n";
+
+  std::cout << "Expected: [";
+  for (int i = 0; i < N; i++)
+  {
+    std::cout << (i ? ", " : "") << expected[i];
+  }
+  std::cout << "]\n";
```
bernhardmgruber, line 121: Suggestion: looks like we should have a test utility for printing the content of a vector. Could we just rely on `std::format` for that? There should be no reason these tests here must compile for C++17. We should be able to rely on C++20.

c/parallel/test/freestanding/test_cub_device_reduce_custom_op.cpp
```diff
@@ -0,0 +1,243 @@
...
+  {                                                                                                            \
+    cudaError_t err = call;                                                                                    \
+    if (err != cudaSuccess)                                                                                    \
+    {                                                                                                          \
+      std::cerr << "CUDA error at " << __FILE__ << ":" << __LINE__ << ": " << cudaGetErrorString(err) << "\n"; \
+      return 1;                                                                                                \
+    }                                                                                                          \
+  } while (0)
+
+#ifdef USE_ADD_OP
```
bernhardmgruber: Q: Why isn't the test compiled twice with a macro, or better a constant, that switches between the add and min operator? We should enable all the tests we have.

c/parallel/test/freestanding/test_cub_device_adjacent_difference.cpp
miscco: why shouldnt we be able to use thrust vectors? this is calling CUB kernels

c/parallel/test/freestanding/test_cub_device_reduce_custom_op.cpp
miscco: Honestly I am not sure, will probably split

c/parallel/test/freestanding/test_cub_device_adjacent_difference.cpp
bernhardmgruber: Because thrust vector's cannot be compiled by a non-CUDA compiler, since they bring in CUDA C++ syntax.

miscco: I doubt CUB can be compiled with a non-CUDA compiler

c/parallel/test/freestanding/test_cub_device_reduce_custom_op.cpp
miscco: Done

c/parallel/test/freestanding/test_cub_device_adjacent_difference.cpp
miscco: Will postpone that for a followup

bernhardmgruber (approved):

c/parallel/test/freestanding/test_cub_device_adjacent_difference.cpp
bernhardmgruber: You are only using CUB in the JIT-ed portion of the code. I am talking about the code right here, that is pre compiled.

miscco: question, where is this using `thrust::vector` ? this is using `std::vector`

bernhardmgruber: Nowhere. Sorry if I wasn't clear enough. I am suggesting to replace the raw allocation using `cudaMalloc` by a smart pointer or container. We cannot use `thrust::vector` here, so I suggested to use `cuda::buffer`.

PR #8802, opened 2026-05-04
[libcu++] Add shared versions of memory pool owning objects

### LINKED ISSUE #8827: Add shared ownership versions of memory pool types

pciolkosz:
This PR adds `shared_*_memory_pool` types, which are memory pools with shared ownership semantics. Currently `shared_resource` can be used with a memory pool type to have shared ownership of a pool, but it has some downsides:
1. User needs to know the `shared_resource` wrapper exist, if they just reach for the memory pool type it won't work out of the box with `any_resource`, `buffer` etc.
2. Member functions of the pool need to be accessed with `.res.get.member()`, the new types have the members available immediately
3. Shared resource is a bit more difficult to construct.

This PR also removes `release` member function from `*_memory_pool_ref` types that was added there by mistake and didn't make much sense.
It also adds `no_init` constructors to `*_memory_pool` types to align with the shared counterparts and provide a way to have an object you move a pool into at some later point.

The implementation moved the ref counting bits of `shared_resource` to an internal class, so it could be reused for these new shared pool types. There is a bit of duplication because of that, the deleter holds another copy of the memory pool pointer, but it allowed to keep the implementation simpler

## Release Notes

* **New Features**
  * Added shared-ownership memory pool variants: `shared_device_memory_pool`, `shared_managed_memory_pool`, and `shared_pinned_memory_pool` enabling copyable shared ownership semantics
  * Added support for creating empty memory pools via `no_init` constructor  
  * Added `release()` method to existing pool types for explicit ownership transfer

* **Tests**
  * Added comprehensive test coverage for new shared memory pool functionality


libcudacxx/include/cuda/__memory_pool/shared_device_memory_pool.h
```diff
@@ -0,0 +1,109 @@
...
+//! Shared device memory pool
+//! -------------------------
+//!
+//! ``shared_device_memory_pool`` provides shared ownership of a device memory
+//! pool. It is copyable and each copy shares the same underlying
+//! ``cudaMemPool_t`` via reference counting. This makes it usable with
+//! ``any_resource`` and other contexts that require copyable resources.
+//!
+//! @endrst
+struct shared_device_memory_pool : __shared_memory_pool_base<shared_device_memory_pool>
```
Jacobfaib: Nit: not sure if CCCL has this in the style-guide but `struct` is usually reserved for PODs or type-traits. Types with access visibility, or operators, constructors and other such C++-isms should use `class`.

libcudacxx/include/cuda/__memory_pool/shared_managed_memory_pool.h
```diff
@@ -0,0 +1,107 @@
...
+//! @endrst
+struct shared_managed_memory_pool : __shared_memory_pool_base<shared_managed_memory_pool>
+{
+  _CCCL_EXEC_CHECK_DISABLE
+  _CCCL_HOST_API shared_managed_memory_pool(const shared_managed_memory_pool& __other)
+      : __shared_memory_pool_base(__other)
+  {}
+  _CCCL_EXEC_CHECK_DISABLE
+  _CCCL_HOST_API shared_managed_memory_pool(shared_managed_memory_pool&& __other)
+      : __shared_memory_pool_base(static_cast<__shared_memory_pool_base&&>(__other))
```
Jacobfaib:
```suggestion
      : __shared_memory_pool_base(::cuda::std::move(__other))
```

libcudacxx/include/cuda/__memory_pool/shared_managed_memory_pool.h
```diff
@@ -0,0 +1,107 @@
...
+  {}
+  shared_managed_memory_pool& operator=(const shared_managed_memory_pool&) = default;
+  shared_managed_memory_pool& operator=(shared_managed_memory_pool&&)      = default;
+  ~shared_managed_memory_pool()                                            = default;
```
Jacobfaib: You don't need to explicitly define this, the base class already has a public one
```suggestion
```

libcudacxx/include/cuda/__memory_pool/shared_memory_pool_base.h
```diff
@@ -0,0 +1,117 @@
...
+
+  _CCCL_EXEC_CHECK_DISABLE
+  _CCCL_HOST_API __shared_memory_pool_base(__shared_memory_pool_base&& __other) noexcept
+      : __memory_pool_base(__other.__pool_)
+      , __ref_(::cuda::std::move(__other.__ref_))
+  {}
+
+  __shared_memory_pool_base& operator=(const __shared_memory_pool_base&) = default;
+  __shared_memory_pool_base& operator=(__shared_memory_pool_base&&)      = default;
+  ~__shared_memory_pool_base()                                           = default;
```
Jacobfaib: Likewise you don't need to explicitly define this, the base classes will have one if needed.

libcudacxx/include/cuda/__memory_pool/shared_managed_memory_pool.h
```diff
@@ -0,0 +1,107 @@
...
+
+  //! @brief Constructs a shared managed memory pool from an existing native handle.
+  //! @param __pool The ``cudaMemPool_t`` to take shared ownership of.
+  _CCCL_HOST_API static shared_managed_memory_pool from_native_handle(::cudaMemPool_t __pool) noexcept
+  {
+    return shared_managed_memory_pool(__pool);
+  }
+
+  _CCCL_HOST_API friend constexpr void
+  get_property(shared_managed_memory_pool const&, ::cuda::mr::device_accessible) noexcept
```
Jacobfaib: Nit: use of east and west `const` is inconsistent (though I guess this is a problem for `clang-tidy`?)

libcudacxx/include/cuda/__memory_pool/shared_pinned_memory_pool.h
```diff
@@ -0,0 +1,126 @@
...
+
+#    if _CCCL_CTK_AT_LEAST(13, 0)
+  //! @brief Constructs a shared pinned memory pool.
+  //! @param __properties Optional pool creation properties.
+  _CCCL_HOST_API explicit shared_pinned_memory_pool(memory_pool_properties __properties = {})
+      : __shared_memory_pool_base(__create_cuda_mempool(
+          __properties, ::CUmemLocation{::CU_MEM_LOCATION_TYPE_HOST, 0}, ::CU_MEM_ALLOCATION_TYPE_PINNED))
+  {
+    enable_access_from(cuda::devices);
+  }
```
Jacobfaib: Important: let's call the other constructor from this one instead of duplicating the logic.

libcudacxx/include/cuda/__memory_resource/shared_block_ptr.h
```diff
@@ -0,0 +1,163 @@
...
+_CCCL_BEGIN_NAMESPACE_CUDA_MR
+
+//! @brief A ref-counted control block holding a payload and an atomic reference count.
+template <class _Payload>
+struct __shared_control_block
+{
+  template <class... _Args>
+  _CCCL_HOST_API explicit __shared_control_block(_Args&&... __args)
+      : __payload(::cuda::std::forward<_Args>(__args)...)
+      , __ref_count(1)
```
Jacobfaib: Important: prefer to do static initialization like this in the variable declaration:
```c++
:cuda::std::atomic<int> __ref_count{1};
```
If you add other constructors you can never forget to do it, and it will also correctly initialize if you aggregate initialize the type via `{}`

libcudacxx/include/cuda/__memory_resource/shared_block_ptr.h
```diff
@@ -0,0 +1,163 @@
...
+{
+  using __block_t = __shared_control_block<_Payload>;
+
+  __block_t* __block_;
+
+public:
+  //! @brief Constructs a null ``__shared_block_ptr`` with no control block.
+  _CCCL_HOST_API __shared_block_ptr() noexcept
+      : __block_(nullptr)
+  {}
```
Jacobfaib: Initialize `__block_ = nullptr` then you can simply `= default` this constructor.

libcudacxx/include/cuda/__memory_resource/shared_block_ptr.h
```diff
@@ -0,0 +1,163 @@
...
+  _CCCL_HOST_API __shared_block_ptr(__shared_block_ptr&& __other) noexcept
+      : __block_(::cuda::std::exchange(__other.__block_, nullptr))
+  {}
+
+  _CCCL_HOST_API __shared_block_ptr& operator=(const __shared_block_ptr& __other) noexcept
+  {
+    if (this != &__other)
+    {
+      __shared_block_ptr(__other).swap(*this);
+    }
```
Jacobfaib: Important: you don't need to guard behind `this != &other`. Copy and swap does the right thing for self assignment.

libcudacxx/include/cuda/__memory_resource/shared_block_ptr.h
```diff
@@ -0,0 +1,163 @@
...
+    return *this;
+  }
+
+  _CCCL_HOST_API __shared_block_ptr& operator=(__shared_block_ptr&& __other) noexcept
+  {
+    if (this != &__other)
```
Jacobfaib: Same here with guard not required.

libcudacxx/include/cuda/__memory_resource/shared_block_ptr.h
```diff
@@ -0,0 +1,163 @@
...
+    }
+    return *this;
+  }
+
+  _CCCL_HOST_API ~__shared_block_ptr()
+  {
+    if (__block_ && __block_->__ref_count.fetch_sub(1, ::cuda::std::memory_order_acq_rel) == 1)
+    {
+      delete __block_;
+    }
```
Jacobfaib: Critical: There is a subtle but potentially significant performance optimization we can do here by making this `memory_order_release` and putting an `acquire` threadfence in the deletion path:
```suggestion
    if (__block_ && __block_->__ref_count.fetch_sub(1, ::cuda::std::memory_order_release) == 1)
    {
      ::cuda::std::atomic_thread_fence(::cuda::std::memory_order_acquire);
      delete __block_;
    }
```
`acq_rel` (or `seq_cst`, there's no difference in this scenario) would make it so that all thread read/writes need to be strictly ordered against each other, but that's actually overkill. If thread 1 and 2 both decrement to a non-zero refcount we don't care what order they do the decrement in, so long as their writes are flushed by the time thread 0 decrements it completely.

So only the thread actually _performing_ the deletion of the control block needs to be strictly ordered after prior thread writes. All other threads only need to ensure their read/writes are flushed.

libcudacxx/include/cuda/__memory_resource/shared_block_ptr.h
```diff
@@ -0,0 +1,163 @@
...
+  }
+
+  _CCCL_HOST_API friend void swap(__shared_block_ptr& __lhs, __shared_block_ptr& __rhs) noexcept
+  {
+    __lhs.swap(__rhs);
+  }
+
+  [[nodiscard]] _CCCL_HOST_API _Payload& __payload() noexcept
+  {
+    return __block_->__payload;
```
Jacobfaib, line 124: Nit: Add a `_CCCL_ASSERT(__block_)` to these to aide debugging

libcudacxx/include/cuda/__memory_resource/shared_block_ptr.h
```diff
@@ -0,0 +1,163 @@
...
+  {
+    return __lhs.__block_ == __rhs.__block_;
+  }
+
+#if _CCCL_STD_VER <= 2017
+  [[nodiscard]] _CCCL_HOST_API friend bool
+  operator!=(const __shared_block_ptr& __lhs, const __shared_block_ptr& __rhs) noexcept
+  {
+    return __lhs.__block_ != __rhs.__block_;
+  }
```
Jacobfaib, line 149: Important: `shared_ptr` and friends also implement the other relational operators. We probably don't need all of them, but at the very least we should:
1. Implement operators against `nullptr_t`, so we can say `__shard_block_ptr == nullptr`
2. Implement `operator<` so we can put this in `std::map`
3. Implement `std::hash` so we can put this in a `std::unordered_map`

libcudacxx/include/cuda/__memory_resource/shared_resource.h
```diff
@@ -66,50 +62,38 @@ struct shared_resource
...
-      : __control_block(__other.__control_block)
-  {
-    if (__control_block)
-    {
-      __control_block->__ref_count.fetch_add(1, ::cuda::std::memory_order_relaxed);
-    }
-  }
+  _CCCL_HOST_API shared_resource(const shared_resource& __other) noexcept
+      : __block_(__other.__block_)
+  {}
```
Jacobfaib, line 73: Important: can be `= default` since `__block_` implements copy constructors.

libcudacxx/include/cuda/__memory_resource/shared_resource.h
```diff
@@ -66,50 +62,38 @@ struct shared_resource
...
 
   //! @brief Move-constructs a \c shared_resource assuming ownership of the resource stored
   //! in \c __other.
   //! @param __other The \c shared_resource object to move from.
   //! @post \c __other is left in a valid but unspecified state.
-  shared_resource(shared_resource&& __other) noexcept
-      : __control_block(::cuda::std::exchange(__other.__control_block, nullptr))
+  _CCCL_HOST_API shared_resource(shared_resource&& __other) noexcept
+      : __block_(::cuda::std::move(__other.__block_))
   {}
```
Jacobfaib, line 81: Important: can be `= default` as well.

libcudacxx/include/cuda/__memory_resource/shared_resource.h
```diff
@@ -66,50 +62,38 @@ struct shared_resource
...
   //! @brief Releases the reference held by this \c shared_resource object. If this is the
   //! last reference to the wrapped resource, the resource is deleted.
-  ~shared_resource()
-  {
-    if (__control_block && __control_block->__ref_count.fetch_sub(1, ::cuda::std::memory_order_acq_rel) == 1)
-    {
-      delete __control_block;
-    }
-  }
+  ~shared_resource() = default;
```
Jacobfaib: Important: no need to define this

libcudacxx/include/cuda/__memory_resource/shared_resource.h
```diff
@@ -66,50 +62,38 @@ struct shared_resource
...
 
   //! @brief Copy-assigns from \c __other. Self-assignment is a no-op. Otherwise, the reference
   //! held by this \c shared_resource object is released and a new reference is acquired to the
   //! wrapped resource of \c __other, if any.
   //! @param __other The \c shared_resource object to copy from.
-  shared_resource& operator=(const shared_resource& __other) noexcept
+  _CCCL_HOST_API shared_resource& operator=(const shared_resource& __other) noexcept
   {
     if (this != &__other)
```
Jacobfaib: Same as above, don't need to guard here

libcudacxx/include/cuda/__memory_resource/shared_block_ptr.h
pciolkosz: This an internal type, I would rather add that stuff on case by case basis when its needed

libcudacxx/include/cuda/__memory_resource/shared_resource.h
pciolkosz: If I do it nvcc 12 gets confused in execution space checks

libcudacxx/include/cuda/__memory_resource/shared_resource.h
pciolkosz: If I do it nvcc 12 gets confused in execution space checks

libcudacxx/include/cuda/__memory_pool/shared_pinned_memory_pool.h
```diff
@@ -0,0 +1,117 @@
...
+
+#  include <cuda/__memory_pool/pinned_memory_pool.h>
+#  include <cuda/__memory_pool/shared_memory_pool_base.h>
+#  include <cuda/__memory_resource/properties.h>
+
+#  include <cuda/std/__cccl/prologue.h>
+
+_CCCL_BEGIN_NAMESPACE_CUDA
+
+#  if _CCCL_CTK_AT_LEAST(12, 9)
```
bdice, line 34: Why 12.9 here? I think pinned pools have been supported since 12.6.

libcudacxx/test/libcudacxx/cuda/memory_resource/resources/memory_pools.cu
```diff
@@ -57,27 +57,24 @@ static_assert(cuda::std::is_default_constructible<cuda::pinned_memory_pool>::val
...
+PoolType construct_pool(cuda::memory_pool_properties props = {})
 {
   if constexpr (cuda::std::is_same_v<PoolType, cuda::device_memory_pool>)
   {
-    return cuda::device_memory_pool(device_id, props);
+    return cuda::device_memory_pool(0, props);
   }
-  else
-  {
 #if _CCCL_CTK_AT_LEAST(12, 9)
```
bdice, line 66: 12.6? Same in a few other places.

bdice (approved):
Generally looks good to me. I had one question about the supported CUDA version for pinned pools. I am okay with requiring 12.9 -- we already require 12.9 for all CUDA 12 compilation in RAPIDS -- just wanted to verify the intent and make sure I wasn't missing something.

libcudacxx/include/cuda/__memory_pool/shared_pinned_memory_pool.h
pciolkosz: There was some issue where the attribute to check for pinned pool support was not added until 12.9 and we needed that for validation. I decided to bump the requirement for the whole thing, I think we are far enough into 13.X cycle that 12.6 vs 12.9 shouldn't matter that much

bdice: Good, thanks!

pciolkosz:
@coderabbitai full review

libcudacxx/include/cuda/__memory_pool/shared_device_memory_pool.h
```diff
@@ -0,0 +1,104 @@
...
+
+  using default_queries = ::cuda::mr::properties_list<::cuda::mr::device_accessible>;
+
+private:
+  _CCCL_HOST_API explicit shared_device_memory_pool(::cudaMemPool_t __pool) noexcept
+      : __shared_memory_pool_base(__pool)
+  {}
+};
+
+static_assert(::cuda::mr::resource_with<shared_device_memory_pool, ::cuda::mr::device_accessible>);
```
davebayer: This should just be part of the tests

libcudacxx/include/cuda/__memory_pool/shared_managed_memory_pool.h
```diff
@@ -0,0 +1,104 @@
...
+  using default_queries = ::cuda::mr::properties_list<::cuda::mr::device_accessible, ::cuda::mr::host_accessible>;
+
+private:
+  _CCCL_HOST_API explicit shared_managed_memory_pool(::cudaMemPool_t __pool) noexcept
+      : __shared_memory_pool_base(__pool)
+  {}
+};
+
+static_assert(::cuda::mr::resource_with<shared_managed_memory_pool, ::cuda::mr::device_accessible>);
+static_assert(::cuda::mr::resource_with<shared_managed_memory_pool, ::cuda::mr::host_accessible>);
```
davebayer: ditto

libcudacxx/include/cuda/__memory_pool/shared_memory_pool_base.h
```diff
@@ -0,0 +1,116 @@
...
+
+  _CCCL_HOST_API __pool_destroyer(__pool_destroyer&& __other) noexcept
+      : __pool(::cuda::std::exchange(__other.__pool, nullptr))
+  {}
+
+  _CCCL_HOST_API ~__pool_destroyer() noexcept
+  {
+    if (__pool != nullptr)
+    {
+      ::cuda::__driver::__mempoolDestroy(__pool);
```
davebayer: Can't `__mempoolDestroy` throw? This function is marked as `noexcept`, we should add a non-throwing version

libcudacxx/include/cuda/__memory_pool/shared_memory_pool_base.h
```diff
@@ -0,0 +1,116 @@
...
+  ::cuda::mr::__shared_block_ptr<__pool_destroyer> __ref_;
+
+protected:
+  _CCCL_HOST_API explicit __shared_memory_pool_base(no_init_t) noexcept
+      : __memory_pool_base(::cudaMemPool_t{})
+  {}
+
+  _CCCL_HOST_API explicit __shared_memory_pool_base(::cudaMemPool_t __pool)
+      : __memory_pool_base(__pool)
+      , __ref_(__pool)
```
davebayer, line 87: We duplicate the stored handles? We store `cudaMempool_t` once in `__memory_pool_base` and the second time in `__shared_block_ptr`? Can't we do better?

libcudacxx/include/cuda/__memory_pool/shared_pinned_memory_pool.h
```diff
@@ -0,0 +1,117 @@
...
+  using default_queries = ::cuda::mr::properties_list<::cuda::mr::device_accessible, ::cuda::mr::host_accessible>;
+
+private:
+  _CCCL_HOST_API explicit shared_pinned_memory_pool(::cudaMemPool_t __pool) noexcept
+      : __shared_memory_pool_base(__pool)
+  {}
+};
+
+static_assert(::cuda::mr::resource_with<shared_pinned_memory_pool, ::cuda::mr::device_accessible>);
+static_assert(::cuda::mr::resource_with<shared_pinned_memory_pool, ::cuda::mr::host_accessible>);
```
davebayer: ditto

libcudacxx/include/cuda/__memory_resource/shared_block_ptr.h
```diff
@@ -0,0 +1,157 @@
...
+//! @brief A ref-counted smart pointer to a ``__shared_control_block<_Payload>``.
+//!
+//! Manages shared ownership of a heap-allocated control block. When the last
+//! ``__shared_block_ptr`` referencing a block is destroyed, the block (and its
+//! payload) are deleted.
+//!
+//! Used as the common ref-counting mechanism for ``shared_resource`` and the
+//! shared memory pool types.
+template <class _Payload>
+class __shared_block_ptr
```
davebayer, line 60: I'd like to see this type being properly tested

davebayer (changes requested):

libcudacxx/include/cuda/__memory_pool/shared_memory_pool_base.h
pciolkosz: I implemented it first without the duplication, but it made the implementation much messier, I don't think it's worth it. You either use only the copy in the shared_block and then you have to reimplement half of `__memory_pool_base` or you only use the one in the pool base, but you have to do some weird destruction pattern, where you have an explicit destroy call for the shared_block. Both are workable solutions, but the code becomes either redundant or difficult to follow

libcudacxx/include/cuda/__memory_resource/shared_block_ptr.h
```diff
@@ -0,0 +1,157 @@
...
+{
+  using __block_t = __shared_control_block<_Payload>;
+
+  __block_t* __block_ = nullptr;
+
+public:
+  //! @brief Constructs a null ``__shared_block_ptr`` with no control block.
+  __shared_block_ptr() = default;
```
davebayer:
```suggestion
  _CCCL_HIDE_FROM_ABI __shared_block_ptr() = default;
```

davebayer (approved):

PR #5176, opened 2025-07-08
Fix overflow in offset calculation in transform kernel

bernhardmgruber:
- [x] Merge before: #5175

PR #8803, opened 2026-05-05
Fix DeviceTransform byte-offset overflow when num_items*sizeof(T) > 4Gb

### LINKED ISSUE #8800: [BUG]: thrust::scatter silently corrupts data beyond 4 GiB byte offset due to 32-bit arithmetic overflow
### Is this a duplicate?

### Type of Bug

Silent Failure

### Component

Thrust

### Describe the bug

`thrust::scatter` produces incorrect results when scattering from source arrays containing more than 2^30 elements of 4-byte types (or any configuration where `element_count × sizeof(T) ≥ 4 GiB`). The operation silently reads from wrong source indices due to 32-bit integer overflow in byte-offset calculations, causing values at `source[2^30 + k]` to be incorrectly read as `source[k]`.

**This is a silent data corruption bug** - no error or warning is produced, but results are incorrect.

### How to Reproduce

A complete standalone reproducer is provided.

  Conditions Required

  1. Large array: Use thrust::scatter with > 2^30 elements of 4-byte types (or > 2^29 for 8-byte types)
  2. Cross 4 GiB boundary: Ensure scatter reads from indices beyond 2^30
  3. CCCL v3.2.0: Bug exists in current CCCL (likely earlier versions too)

  Simple Reproduction
```c++
  // 1. Create a large source array (> 1 billion uint32_t elements)
  constexpr auto idx = 1u << 30;
  constexpr auto num_elements = idx + 1;
  thrust::device_vector<uint32_t> source(num_elements);
  thrust::sequence(source.begin(), source.end(), 1000);  // source[i] = 1000 + i

  // 2. Create identity scatter map
  thrust::device_vector<uint32_t> map(num_elements);
  thrust::sequence(map.begin(), map.end(), 0);  // map[i] = i

  // 3. Scatter
  thrust::device_vector<uint32_t> output(num_elements);
  thrust::scatter(source.begin(), source.end(), map.begin(), output.begin());

  // 4. Check value at 2^30 (1,073,741,824)
  assert(output[idx] == 1000u + idx);
```
[scatterBillionElementTest.cpp](https://github.com/user-attachments/files/27370136/scatterBillionElementTest.cpp)

### Expected behavior

After calling scatter with the above repro:  
Expected: output[2^30] = 1000 + 2^30 = 1,073,742,824  
Actual:   output[2^30] = 1000 (reads from source[0] instead!)  

### Reproduction link

_No response_

### Operating System

Ubuntu 24.04.3 LTS

### nvidia-smi output
```
+-----------------------------------------------------------------------------------------+
| NVIDIA-SMI 575.57.08              Driver Version: 575.57.08      CUDA Version: 12.9     |
|-----------------------------------------+------------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id          Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |           Memory-Usage | GPU-Util  Compute M. |
|                                         |                        |               MIG M. |
|=========================================+========================+======================|
|   0  NVIDIA A100 80GB PCIe          Off |   00000000:C1:00.0 Off |                    0 |
| N/A   39C    P0             47W /  300W |      14MiB /  81920MiB |      0%      Default |
|                                         |                        |             Disabled |
+-----------------------------------------+------------------------+----------------------+
```
### NVCC version

nvcc: NVIDIA (R) Cuda compiler driver
Copyright (c) 2005-2025 NVIDIA Corporation
Built on Tue_May_27_02:21:03_PDT_2025
Cuda compilation tools, release 12.9, V12.9.86
Build cuda_12.9.r12.9/compiler.36037853_0

gonidelis:
fixes https://github.com/NVIDIA/cccl/issues/8800

~_adding more context and tests in a bit - don't merge_~ my full research path in the comments

When we addd TMA in `__transform_internal`, when the dispatcher picks `int32` OffsetT and the byte product exceeds `UINT_MAX`, `transform_kernel_ublkcp`'s deliberate `U32` multiply (`offset * unsigned{sizeof(T)}`) wraps and reads from wrong tile addresses, corrupting tail outputs.

### DIFF
--- a/cub/cub/device/device_transform.cuh
+++ b/cub/cub/device/device_transform.cuh
@@ -97,11 +97,16 @@ struct DeviceTransform
 
     // Even when num_items fits in offset_t, the byte product (num_items * sizeof(T)) used for
     // pointer arithmetic inside the kernel may overflow 32-bit math. Promote to int64_t in that case.
+    // Only the input iterators are considered: the kernel's buggy expression
+    // `aligned_ptr.ptr + offset * unsigned{sizeof(T)}` (in transform_kernel_ublkcp) is on the input
+    // bulk-copy side; output writes use the output iterator's own arithmetic. Including the output
+    // iterator's value_type would also break compilation for write-only output iterators whose
+    // value_type is `void` (e.g. transform_output_iterator, tabulate_output_iterator).
     // todo(giannis): consider moving this to a helper function under choose_offset.cuh ?
-    if constexpr (sizeof(offset_t) < sizeof(::cuda::std::int64_t))
+    if constexpr (sizeof(offset_t) < sizeof(::cuda::std::int64_t) && sizeof...(RandomAccessIteratorsIn) > 0)
     {
-      constexpr ::cuda::std::size_t max_value_size = ::cuda::std::max(
-        {sizeof(detail::it_value_t<RandomAccessIteratorsIn>)..., sizeof(detail::it_value_t<RandomAccessIteratorOut>)});
+      constexpr ::cuda::std::size_t max_value_size =
+        ::cuda::std::max({sizeof(detail::it_value_t<RandomAccessIteratorsIn>)...});
       if (static_cast<::cuda::std::uint64_t>(num_items) * max_value_size > 0xFFFFFFFFull)
       {
         return detail::transform::dispatch<StableAddress>(
--- a/cub/cub/device/device_transform.cuh
+++ b/cub/cub/device/device_transform.cuh
@@ -16,16 +16,13 @@
 #include <cub/detail/choose_offset.cuh>
 #include <cub/device/dispatch/dispatch_transform.cuh>
 #include <cub/util_namespace.cuh>
-#include <cub/util_type.cuh>
 
 #include <cuda/__execution/tune.h>
 #include <cuda/__functional/address_stability.h>
 #include <cuda/__functional/call_or.h>
 #include <cuda/__iterator/zip_iterator.h>
 #include <cuda/__stream/get_stream.h>
-#include <cuda/std/__algorithm/max.h>
 #include <cuda/std/__execution/env.h>
-#include <cuda/std/cstdint>
 #include <cuda/std/tuple>
 
 CUB_NAMESPACE_BEGIN
@@ -94,31 +91,6 @@ struct DeviceTransform
 #if _CCCL_HAS_CONCEPTS()
     static_assert(detail::transform::transform_policy_selector<policy_selector>);
 #endif // _CCCL_HAS_CONCEPTS()
-
-    // Even when num_items fits in offset_t, the byte product (num_items * sizeof(T)) used for
-    // pointer arithmetic inside the kernel may overflow 32-bit math. Promote to int64_t in that case.
-    // Only the input iterators are considered: the kernel's buggy expression
-    // `aligned_ptr.ptr + offset * unsigned{sizeof(T)}` (in transform_kernel_ublkcp) is on the input
-    // bulk-copy side; output writes use the output iterator's own arithmetic. Including the output
-    // iterator's value_type would also break compilation for write-only output iterators whose
-    // value_type is `void` (e.g. transform_output_iterator, tabulate_output_iterator).
-    // todo(giannis): consider moving this to a helper function under choose_offset.cuh ?
-    if constexpr (sizeof(offset_t) < sizeof(::cuda::std::int64_t) && sizeof...(RandomAccessIteratorsIn) > 0)
-    {
-      constexpr ::cuda::std::size_t max_value_size =
-        ::cuda::std::max({sizeof(detail::it_value_t<RandomAccessIteratorsIn>)...});
-      if (static_cast<::cuda::std::uint64_t>(num_items) * max_value_size > 0xFFFFFFFFull)
-      {
-        return detail::transform::dispatch<StableAddress>(
-          ::cuda::std::move(inputs),
-          ::cuda::std::move(output),
-          static_cast<::cuda::std::int64_t>(num_items),
-          ::cuda::std::move(predicate),
-          ::cuda::std::move(transform_op),
-          stream,
-          policy_selector{});
-      }
-    }
 
     return detail::transform::dispatch<StableAddress>(
       ::cuda::std::move(inputs),
--- a/cub/cub/device/device_transform.cuh
+++ b/cub/cub/device/device_transform.cuh
@@ -16,13 +16,16 @@
 #include <cub/detail/choose_offset.cuh>
 #include <cub/device/dispatch/dispatch_transform.cuh>
 #include <cub/util_namespace.cuh>
+#include <cub/util_type.cuh>
 
 #include <cuda/__execution/tune.h>
 #include <cuda/__functional/address_stability.h>
 #include <cuda/__functional/call_or.h>
 #include <cuda/__iterator/zip_iterator.h>
 #include <cuda/__stream/get_stream.h>
+#include <cuda/std/__algorithm/max.h>
 #include <cuda/std/__execution/env.h>
+#include <cuda/std/cstdint>
 #include <cuda/std/tuple>
 
 CUB_NAMESPACE_BEGIN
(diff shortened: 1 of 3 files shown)


cub/cub/device/device_transform.cuh
```diff
@@ -92,6 +95,26 @@ struct DeviceTransform
     static_assert(detail::transform::transform_policy_selector<policy_selector>);
 #endif // _CCCL_HAS_CONCEPTS()
 
+    // Even when num_items fits in offset_t, the byte product (num_items * sizeof(T)) used for
+    // pointer arithmetic inside the kernel may overflow 32-bit math. Promote to int64_t in that case.
+    // todo(giannis): consider moving this to a helper function under choose_offset.cuh ?
```
gonidelis: @elstehle got an opinion?

pauleonix: I think we should just remove that cast to `unsigned`. This current fix does compile two kernels instead of one. And if the compiler would do that pointer arithmetic for us, it would use `size_t` anyway. @bernhardmgruber 

For reference: If the code were using `BlockLoadToShared` (as future uses of UBLKCPY should do), a `span<T>` would need to be passed to the primitive and therefore any sane user would let the compiler do the pointer arithmetic on `T*` instead, so the bug would be avoided.

I fear you will have to do some benchmarking to make sure that my proposed change really does not pessimize perf.

pauleonix: If it does, we will need the ~~guarantees API~~ argument framwork to opt into the the 32b arithmetic.

gonidelis: running benchmark now

bernhardmgruber:
Thanks for talking a stab at this @gonidelis! We should:

* add a test/extend the tests to exposes the bug. I assume you tested whether the test you added shows the crash.
* fix the bug in a minimal way. I think we should just ensure that any offset computation in the kernel does not exceed the value range of the used integral types.
* Find out when the bug was introduced and decide whether we should backport the fix

IIRC, the performance of `DeviceTransform` is pretty much the same for I32 and I64 offset types and I considered in the past to just hardcode it to I64 or U64. Maybe we should do that as well (potentially as a follow-up).

bernhardmgruber:
I think #5176 already tried to solve this problem, but not sufficiently. But we should be able to adapt it's test to use `unsigned int` instead of `unsigned short` as value type.

cub/test/catch2_test_device_transform.cu
```diff
@@ -139,6 +140,41 @@ catch (const std::bad_alloc&)
...
+// math, corrupting tail outputs. The "large input" test above uses unsigned short (2 bytes), which
+// keeps the byte product under 4 GiB even at int32 num_items max — so it does not cover this regime.
+// This test does, with a 4-byte type pushed past 2^30 elements.
+C2H_TEST("DeviceTransform::Transform byte-product exceeds 4 GiB",
+         "[device][transform][skip-cs-initcheck][skip-cs-racecheck][skip-cs-synccheck]",
+         offset_types)
+try
+{
+  using type     = std::uint32_t;
+  using offset_t = c2h::get<0, TestType>;
```
bernhardmgruber: Important: Please just change the above test to use `uint32_t` as value type and run the test per offset type and with a negative and positive offset added to the num items to cover cases below and above 4GiB.

The test here somewhat duplicates the goal of the above test and is actually better than the above test. So let's just have one good test.

gonidelis: i am embedding this test to the one you added in #5176

gonidelis:
This compares the babelstream benchmark but with a 12byte (non-power-of-2) input type I introduced so that `offset * sizeof(T)` can't lower to a shift and a `IMAD` or `IMAD.WIDE` is actually emitted. I ran for `2^24`, and `2^28` elements. It compares `upstream/main` and Paul's fix (current top of this PR) of stripping the init-cast.
 
clean cells (with not-significant noise) show no changes in wall time.

```md
['main_tri.json', 'paul_tri.json']
# mul

## [0] NVIDIA RTX PRO 6000 Blackwell Workstation Edition

|   Tri   |      I32      |      2^24      | 275.925 us |       0.63% | 276.029 us |       0.63% |  0.104 us |   0.04% |   SAME   |
|   Tri   |      I32      |      2^28      |   4.411 ms |       0.16% |   4.411 ms |       0.15% | -0.113 us |  -0.00% |   SAME   |
|   Tri   |      I64      |      2^24      | 277.843 us |       1.37% | 277.758 us |       1.39% | -0.085 us |  -0.03% |   SAME   |
|   Tri   |      I64      |      2^28      |   4.411 ms |       0.17% |   4.412 ms |       0.16% |  0.246 us |   0.01% |   SAME   |

# add

## [0] NVIDIA RTX PRO 6000 Blackwell Workstation Edition

|   Tri   |      I32      |      2^24      | 402.657 us |       0.44% | 402.678 us |       0.48% |  0.022 us |   0.01% |   SAME   |
|   Tri   |      I32      |      2^28      |   6.352 ms |       0.12% |   6.352 ms |       0.13% |  0.024 us |   0.00% |   SAME   |
|   Tri   |      I64      |      2^24      | 401.844 us |       0.43% | 401.793 us |       0.43% | -0.051 us |  -0.01% |   SAME   |
|   Tri   |      I64      |      2^28      |   6.348 ms |       0.14% |   6.347 ms |       0.14% | -0.344 us |  -0.01% |   SAME   |

# triad

## [0] NVIDIA RTX PRO 6000 Blackwell Workstation Edition

|  T{ct}  |  OffsetT{ct}  |  Elements{io}  |   Ref Time |   Ref Noise |   Cmp Time |   Cmp Noise |       Diff |   %Diff |  Status  |
|---------|---------------|----------------|------------|-------------|------------|-------------|------------|---------|----------|
|   Tri   |      I32      |      2^24      | 403.980 us |       0.42% | 403.990 us |       0.44% |   0.010 us |   0.00% |   SAME   |
|   Tri   |      I32      |      2^28      |   6.363 ms |       0.12% |   6.363 ms |       0.12% |   0.185 us |   0.00% |   SAME   |
|   Tri   |      I64      |      2^24      | 403.565 us |       0.49% | 400.606 us |       3.34% |  -2.959 us |  -0.73% |   FAST   |
|   Tri   |      I64      |      2^28      |   7.110 ms |       7.58% |   7.598 ms |      11.49% | 487.543 us |   6.86% |   SAME   |

# nstream

## [0] NVIDIA RTX PRO 6000 Blackwell Workstation Edition

|  T{ct}  |  OffsetT{ct}  |  Elements{io}  |   Ref Time |   Ref Noise |   Cmp Time |   Cmp Noise |       Diff |   %Diff |  Status  |
|---------|---------------|----------------|------------|-------------|------------|-------------|------------|---------|----------|
|   Tri   |      I32      |      2^24      | 514.497 us |       1.39% | 509.182 us |       3.81% |  -5.315 us |  -1.03% |   SAME   |
|   Tri   |      I32      |      2^28      |   9.080 ms |       7.22% |   9.503 ms |      10.09% | 423.068 us |   4.66% |   SAME   |
|   Tri   |      I64      |      2^24      | 529.110 us |       0.41% | 529.070 us |       0.40% |  -0.040 us |  -0.01% |   SAME   |
|   Tri   |      I64      |      2^28      |   8.350 ms |       0.12% |   8.350 ms |       0.12% |  -0.190 us |  -0.00% |   SAME   |

# Summary

- Total Matches: 16
  - Pass    (diff <= min_noise): 15
  - Unknown (infinite noise):    0
  - Failure (diff > min_noise):  1

```

the raw json files to check for yourselves are here https://gist.github.com/gonidelis/df895e885a00ba0b529f002805d7344d

gonidelis:
with the cast, the compiler emitted an `UIMAD + UIADD3.64` pair (2 instructions) for each byte-offset computation; without the cast, it emits a single `UIMAD.WIDE`. Net codegen is **80 instructions shorter, equivalent cycle cost** (`UIMAD.WIDE` ≈ same total cycles as the pair it replaces).

_* For power-of-2 `sizeof(T)`, SASS is byte-identical between the two_ 

## Opcode-histogram diff (notable changes)

| Opcode | main | paul's suggestion | Δ | meaning |
|---|---|---|---|---|
| `UIMAD.WIDE` | 40 | 120 | **+80** | 32×32→64 multiply (single instruction) |
| `UIADD3.64` | 368 | 288 | **−80** | 64-bit add — no longer needed when wide-multiply absorbs it |
| `UMOV` | 516 | 473 | −43 | constant moves |
| `USHF.L.U32` | 336 | 304 | −32 | 32-bit left shift |
| `UIMAD` | 132 | 124 | −8 | 32×32→32 multiply |
| `NOP` | 704 | 727 | +23 | pipeline padding |


source sass files and usage histograms to see yourself https://gist.github.com/gonidelis/010ef6d7b44f3ad722ec10ef224603d7

cub/test/catch2_test_device_transform.cu
```diff
@@ -110,17 +110,22 @@ struct times_seven
   }
 };
 
-C2H_TEST("DeviceTransform::Transform with large input",
+// Exercises the 32-bit byte-offset overflow regime in transform_kernel_ublkcp (NVIDIA/cccl#8800).
+// num_items = 2^30 +/- a few thread blocks: combined with sizeof(type) = 4, byte product
+// straddles 4 GiB (negative delta -> just under, positive -> just over). Both deltas fit in I32
+// and I64 offset types. The earlier version of this test (PR #5176) used unsigned short, which
+// kept the byte product under 4 GiB even at I32 num_items max and missed the bug.
```
bernhardmgruber: No need to explain what we did in the past. Let's keep the part of the comment that explains what the current test aims for
```suggestion
// Exercises the 32-bit byte-offset overflow regime in transform_kernel_ublkcp (NVIDIA/cccl#8800).
// num_items = 2^30 +/- a few thread blocks: combined with sizeof(type) = 4, byte product
// straddles 4 GiB (negative delta -> just under, positive -> just over). Both deltas fit in I32
// and I64 offset types.
```

bernhardmgruber (approved):

bernhardmgruber:
Backports: #8806, #8807, #8808

PR #8810, opened 2026-05-05
[Tile] Improve overlapping range check for `cuda::std::copy`

miscco:
We are currently directly checking whether `__first < __result` which is technically UB if the two pointers to not originate from the same object / array

The tile compiler has better insight into the variables, so it fails to compile the construct we have with the chained conditions. 

Pull it out into the function to both cleanup the code and make tile work

### DIFF
--- a/libcudacxx/include/cuda/std/__algorithm/copy.h
+++ b/libcudacxx/include/cuda/std/__algorithm/copy.h
@@ -86,13 +86,22 @@ _CCCL_EXEC_CHECK_DISABLE
 template <class _Tp, class _Up>
 _CCCL_API constexpr bool __constexpr_tail_overlap(_Tp* __first, _Up* __needle, [[maybe_unused]] _Tp* __last)
 {
+#if !_CCCL_TILE_COMPILATION() // pointer values cannot be compared
+  if (!::cuda::std::is_constant_evaluated())
+  {
+    return __first < __needle;
+  }
+  else
+#endif // !_CCCL_TILE_COMPILATION()
+  {
 #if defined(_CCCL_BUILTIN_CONSTANT_P)
-  NV_IF_ELSE_TARGET(NV_IS_HOST,
-                    (return _CCCL_BUILTIN_CONSTANT_P(__first < __needle) && __first < __needle;),
-                    (return __constexpr_tail_overlap_fallback(__first, __needle, __last);))
+    NV_IF_ELSE_TARGET(NV_IS_HOST,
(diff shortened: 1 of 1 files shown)


libcudacxx/include/cuda/std/__algorithm/copy.h
```diff
@@ -86,13 +86,22 @@ _CCCL_EXEC_CHECK_DISABLE
...
-  NV_IF_ELSE_TARGET(NV_IS_HOST,
-                    (return _CCCL_BUILTIN_CONSTANT_P(__first < __needle) && __first < __needle;),
-                    (return __constexpr_tail_overlap_fallback(__first, __needle, __last);))
+    NV_IF_ELSE_TARGET(NV_IS_HOST,
+                      (return _CCCL_BUILTIN_CONSTANT_P(__first < __needle) && __first < __needle;),
+                      (return __constexpr_tail_overlap_fallback(__first, __needle, __last);))
 #else // ^^^ _CCCL_BUILTIN_CONSTANT_P ^^^ / vvv !_CCCL_BUILTIN_CONSTANT_P vvv
-  return __constexpr_tail_overlap_fallback(__first, __needle, __last);
+    return __constexpr_tail_overlap_fallback(__first, __needle, __last);
 #endif // !_CCCL_BUILTIN_CONSTANT_P
```
Jacobfaib, line 103: This is very difficult to read. Can you perhaps just duplicate the logic inside each `#if - #else` block? Something like so
```c++
#if CCCL_TILE
if (foo) {
  foo();
} else {
  bar();
}
#else
foo();
#endif
```

miscco: This is the way we do conditional compilation branches with if statements all over the place 

also the whole other part is super ugly so I really do nto want to repeat it

davebayer:
Can't `cuda::ranges_overlap` be used in this case?

libcudacxx/include/cuda/std/__algorithm/copy.h
Jacobfaib: > This is the way we do conditional compilation branches with if statements all over the place

Yeah, I know, it's murder on the eyes. I would strongly advise untangling it, even if there is some duplication, but won't hold the PR over it.

Jacobfaib (approved):

miscco:
> Can't `cuda::ranges_overlap` be used in this case?

It brings in all of `<functional>` so I *really* do not want to use it anywhere honestly

PR #8813, opened 2026-05-05
[Tile] Avoid call to `delete[]` in unique_ptr

miscco:
We cannot use it anyhow, but that breaks including headers, even if we do not use it

Instead terminate if we ever encounter it to ensure that we do not silently leak memory

### DIFF
--- a/libcudacxx/include/cuda/std/__memory/unique_ptr.h
+++ b/libcudacxx/include/cuda/std/__memory/unique_ptr.h
@@ -93,7 +93,11 @@ struct _CCCL_TYPE_VISIBILITY_DEFAULT default_delete<_Tp[]>
   operator()(_Up* __ptr) const noexcept
   {
     static_assert(sizeof(_Up) >= 0, "cannot delete an incomplete type");
+#if _CCCL_TILE_COMPILATION()
+    _CCCL_ASSERT(false, "Cannot call delete[] in a tile program");
+#else // ^^^ _CCCL_TILE_COMPILATION() ^^^ / vvv !_CCCL_TILE_COMPILATION() vvv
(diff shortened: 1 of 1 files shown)


libcudacxx/include/cuda/std/__memory/unique_ptr.h
```diff
@@ -93,7 +94,11 @@ struct _CCCL_TYPE_VISIBILITY_DEFAULT default_delete<_Tp[]>
   operator()(_Up* __ptr) const noexcept
   {
     static_assert(sizeof(_Up) >= 0, "cannot delete an incomplete type");
+#if _CCCL_TILE_COMPILATION()
+    ::cuda::std::terminate();
```
Jacobfaib: Important: needs a comment explaining this termination. Or better yet, do the diagnostic with a `_CCCL_VERIFY(false, "A very useful error message");`.

libcudacxx/include/cuda/std/__memory/unique_ptr.h
```diff
@@ -93,7 +93,11 @@ struct _CCCL_TYPE_VISIBILITY_DEFAULT default_delete<_Tp[]>
   operator()(_Up* __ptr) const noexcept
   {
     static_assert(sizeof(_Up) >= 0, "cannot delete an incomplete type");
+#if _CCCL_TILE_COMPILATION()
+    _CCCL_ASSERT(false, "Cannot call delete[] in a tile program");
```
davebayer:
```suggestion
    _CCCL_VERIFY(false, "Cannot call delete[] in a tile program");
```

davebayer (approved):

PR #8819, opened 2026-05-05
[Tile] Disable more tests that rely on return statement inside loop

miscco:
Currently return statements in loops are not allowed and sometimes the compiler trips over them.

Disable 2 tests until that is fixed


bernhardmgruber (approved):

Jacobfaib (approved):
Is there anything left that is supported in tile mode???? It seems basically the whole language is unsupported

davebayer (approved):

Issue #8175, opened 2026-03-25
Final passthrough for all env APIs


gonidelis:
After closing epic #5635 we should do a final pass on all overloads to double verify the following:

- Each non-env overload has a corresponding env overload
- The documentation for all env-overloads renders correctly
- The documentation for all env-overloads follows a consistent style for the description of the environment parameter, code example, etc.
- The env-overload docs should have the same information as their non-env counterparts
- The `versionadded` comments in the documentation reflect the ship vehicle
- The constraints on all env overloads (`enable_if`s) are as strict as necessary to resolve any ambiguities
- Verify where we need a `stream.sync()` before any `REQUIRES(keys = expected)`
- Move default environment tests in a separate api test file use `tab-set-code` like `block-radix-sort` sphynx examples. 3 tabs: default, custom stream, custom env
- Disable lid2/CUDA graph for algorithms (mostly segmented) that perform host-side synchronization ([explanation](https://github.com/NVIDIA/cccl/pull/8286#discussion_r3032805930)).
- Use `stream_ref` explicitly as a last argument in env_api tests and not through an environment wrap as we want to show to the users that we can pass a stream_ref directly.
- `literalinclude` code snippets in the docs need the full path `../../../cub/test/<...>`. Check that they appear in the generated docs.

PR #8829, opened 2026-05-05
Final env-passthrough 1/4

gonidelis:

- `DeviceSegmentedSort`
- `DeviceRadixSort`
- `DeviceSegmentedReduce`
- `DeviceMemCpy`
- `DeviceMerge`

no serious issues found, fixes are predominantly in the docs and have to do with snippet paths (even for non-env algorithms) and documentation conformity between non-env and env counterparts.

It also mainly makes the api tests use the `stream_ref` directly and not wrapped to an env before passing it to the algorithm.

### DIFF
--- a/cub/cub/device/device_segmented_reduce.cuh
+++ b/cub/cub/device/device_segmented_reduce.cuh
@@ -311,7 +311,13 @@ public:
   //!
   //! The code snippet below illustrates a custom min-reduction of a device vector of ``int`` data elements.
   //!
-  //! .. literalinclude:: ../../test/catch2_test_device_segmented_reduce_api.cu
+  //! .. literalinclude:: ../../../cub/test/catch2_test_device_segmented_reduce_api.cu
+  //!     :language: c++
+  //!     :dedent:
+  //!     :start-after: example-begin segmented-reduce-custommin
+  //!     :end-before: example-end segmented-reduce-custommin
+  //!
+  //! .. literalinclude:: ../../../cub/test/catch2_test_device_segmented_reduce_api.cu
   //!     :language: c++
   //!     :dedent:
   //!     :start-after: example-begin segmented-reduce-reduce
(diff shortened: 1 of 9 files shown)


cub/cub/device/device_segmented_reduce.cuh
```diff
@@ -631,13 +643,7 @@ public:
   //!    First appears in CUDA Toolkit 13.4.
   //!
   //! - Does not support binary reduction operators that are non-commutative.
-  //! - Provides "run-to-run" determinism for pseudo-associative reduction
-  //!   (e.g., addition of floating point types) on the same GPU device.
-  //!   However, results for pseudo-associative reduction may be inconsistent
-  //!   from one device to a another device of a different compute-capability
-  //!   because CUB can employ different tile-sizing for different architectures.
```
bernhardmgruber, line 638: Why does that comment no longer apply?

cub/cub/device/device_segmented_reduce.cuh
```diff
@@ -1642,18 +1630,13 @@ public:
...
   //!     ``d_out[i].value`` and its offset in that segment is written to ``d_out[i].key``.
-  //!   - The ``{1, std::numeric_limits<T>::max()}`` tuple is produced for zero-length inputs
+  //!   - The ``{1, ::cuda::std::numeric_limits<T>::max()}`` tuple is produced for zero-length inputs
   //!
-  //! - Does not support ``<`` operators that are non-commutative.
-  //! - Provides "run-to-run" determinism for pseudo-associative reduction
-  //!   (e.g., addition of floating point types) on the same GPU device.
-  //!   However, results for pseudo-associative reduction may be inconsistent
-  //!   from one device to a another device of a different compute-capability
-  //!   because CUB can employ different tile-sizing for different architectures.
```
bernhardmgruber, line 1652: For `ArgMin` I would agree that there is no need to talk about determinism since no numeric should be involved.

cub/cub/device/device_segmented_reduce.cuh
```diff
@@ -2201,15 +2182,13 @@ public:
   }
 
   //! @rst
-  //! Computes a device-wide segmented maximum using the greater-than (``>``) operator
-  //! and a fixed segment size.
```
bernhardmgruber, line 2205: Q: Why is the "and a fixed segment size" removed?

cub/cub/device/device_segmented_reduce.cuh
```diff
@@ -1214,8 +1211,7 @@ public:
   }
 
   //! @rst
-  //! Computes a device-wide segmented minimum using the less-than (``<``) operator
-  //! and a fixed segment size.
```
bernhardmgruber, line 1218: Same here

bernhardmgruber (commented):
LGTM, just a few questions:

bernhardmgruber (approved):

gonidelis:
duh... no need for backporting it's already included in 3.4.x as it was merged before branching off

PR #8835, opened 2026-05-05
Rename `BlockThreads` to `ThreadsPerBlock`

bernhardmgruber:
After some internal discussion in the context of how we name tuning policy members, we concluded that the term "threads per block" was more comprehensible than "block threads". This PR:

* Renames exact matches of `BlockThreads` in template parameters to `ThreadsPerBlock`
* The use of `BlockThreads` in a single `static_assert`
* All matches of `BlockThreads()` functions in tuning policy wrappers (internal stuff in CUB for CCCL.C)
* The remaining uses of `BlockThreads`

One commit each.

### DIFF
--- a/cub/cub/agent/agent_batch_memcpy.cuh
+++ b/cub/cub/agent/agent_batch_memcpy.cuh
@@ -448,7 +448,7 @@ private:
 /**
  * Parameterizable tuning policy type for AgentBatchMemcpy
  */
-template <uint32_t BlockThreads,
+template <uint32_t ThreadsPerBlock,
           uint32_t BuffersPerThread,
           uint32_t TlevBytesPerThread,
           bool PreferPow2Bits,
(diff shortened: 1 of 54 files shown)


pciolkosz (approved):

cub/cub/agent/agent_batch_memcpy.cuh
```diff
@@ -460,7 +460,7 @@ template <uint32_t BlockThreads,
 struct AgentBatchMemcpyPolicy
 {
   /// Threads per thread block
-  static constexpr uint32_t BLOCK_THREADS = BlockThreads;
+  static constexpr uint32_t BLOCK_THREADS = ThreadsPerBlock;
```
davebayer, line 463: These are part of the public API and we cannot change those, right?
```suggestion
  static constexpr uint32_t THREADS_PER_BLOCK = ThreadsPerBlock;
```

davebayer (approved):

cub/cub/agent/agent_reduce.cuh
```diff
@@ -95,7 +95,7 @@ struct AgentWarpReducePolicy
   static constexpr int VECTOR_LOAD_LENGTH = VectorLoadLength;
 
   /// Number of threads per block
-  static constexpr int BLOCK_THREADS = BlockThreads;
+  static constexpr int BLOCK_THREADS = ThreadsPerBlock;
```
miscco, line 98: I assume this is going into a follow up

miscco (approved):

cub/cub/agent/agent_batch_memcpy.cuh
bernhardmgruber: All agent policies in CUB are part of the public API, so yes, the names of their members cannot be changed. However, they are all going to be deprecated when we replace them by the new tuning policies.

PR #8843, opened 2026-05-06
Add a `_CCCL_VISIBILTY_EXPORT` macro

miscco:
We use this in CCCL.C a lot and it very close to our `_CCCL_VISIBILITY_DEFAULT`

### DIFF
--- a/libcudacxx/include/cuda/std/__cccl/visibility.h
+++ b/libcudacxx/include/cuda/std/__cccl/visibility.h
@@ -52,6 +52,14 @@
 #  define _CCCL_VISIBILITY_DEFAULT __attribute__((__visibility__("default")))
 #endif // !_CCCL_COMPILER(NVRTC)
 
+#if _CCCL_COMPILER(MSVC)
+#  define _CCCL_VISIBILITY_EXPORT __declspec(dllexport)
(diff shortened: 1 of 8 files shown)


libcudacxx/include/cuda/std/__cccl/visibility.h
```diff
@@ -52,6 +52,14 @@
 #  define _CCCL_VISIBILITY_DEFAULT __attribute__((__visibility__("default")))
 #endif // !_CCCL_COMPILER(NVRTC)
 
+#if _CCCL_COMPILER(MSVC)
+#  define _CCCL_VISIBILITY_EXPORT __declspec(dllexport)
+#elif _CCCL_COMPILER(NVRTC) // ^^^ _CCCL_COMPILER(MSVC) ^^^ / vvv _CCCL_COMPILER(NVRTC) vvv
+#  define _CCCL_VISIBILITY_EXPORT
+#else // ^^^ _CCCL_COMPILER(NVRTC) ^^^ / vvv !_CCCL_COMPILER(NVRTC) vvv
+#  define _CCCL_VISIBILITY_EXPORT __attribute__((__visibility__("default")))
```
Jacobfaib: I would define this in terms of the default visibility macro for Unix. For unix systems symbol visibility is the only import/export system that exists so the coupling is semantically correct. A visible symbol _is_ an exported symbol.

bernhardmgruber (approved):

bernhardmgruber (approved):

PR #8849, opened 2026-05-06
[libcu++] Rework our handling of `iterator_category`

### LINKED ISSUE #8640: [BUG]: `cuda::transform_iterator` silently downgrades `iterator_category` → 2.5–3.3× slowdowns for thrust algorithms that dispatch on it
### Is this a duplicate?

### Type of Bug

Performance

### Component

General CCCL

## Problem

`cuda::transform_iterator` deliberately sets its classic `iterator_category` member to `input_iterator_tag` whenever the wrapped functor returns by value, even when the underlying iterator is random-access. This is spec-compliant for C++20 `std::ranges::transform_view::iterator`, but it silently breaks downstream Thrust algorithms that dispatch on `std::iterator_traits<It>::iterator_category` (e.g. `thrust::copy` → `thrust::scatter` → `thrust::make_permutation_iterator`), causing them to fall off their CUB bulk / vectorized fast paths.

In cuDF we observed a **2.5–3.3× GPU-side slowdown** after migrating two call sites from `thrust::make_transform_iterator` to `cuda::transform_iterator`. One-line revert restored parity with `thrust::`: https://github.com/rapidsai/cudf/pull/14162.

## Observed impact (cuDF scatter regression, RTX 6000 / CUDA 13.1)

Benchmark: `cudf::scatter` with `COPYING_NVBENCH -b scatter`. Baseline `origin/main` uses `thrust::make_transform_iterator`. The regressing branch uses `cuda::transform_iterator` over this functor:

```cpp
template <typename MapType>
struct index_converter {
  __device__ MapType operator()(MapType in) const { return ((in % n_rows) + n_rows) % n_rows; }
  size_type n_rows;
};
```

The transformed iterator flows into `thrust::scatter(..., map_it, ...)`, which the generic implementation expands to
`thrust::copy(first, last, thrust::make_permutation_iterator(output, map_it))`.

`thrust::make_permutation_iterator` takes the minimum traversal of its two constituents; with `map_it` downgraded to input, the permutation iterator becomes input, `thrust::copy` loses its CUB bulk fast path, and the scatter kernel falls onto a materializing fallback — costing **~170 ns/row of extra device work** and extra device allocations (causing new RMM-pool OOMs at medium sizes):

| num_rows | num_cols | thrust (ref) | cuda (cmp) | GPU Δ | %Diff |
|---:|---:|---:|---:|---:|---:|
| 262 144 | 1 | 24.40 µs | 70.66 µs | +46.3 µs | **+189.6 %** |
| 2 097 152 | 1 | 159.34 µs | 512.57 µs | +353.2 µs | **+221.7 %** |
| 16 777 216 | 1 | 1.227 ms | 4.062 ms | +2.835 ms | **+231.1 %** |
| 262 144 | 8 | 171.12 µs | 463.55 µs | +292.4 µs | **+170.9 %** |
| 2 097 152 | 8 | 1.256 ms | 3.414 ms | +2.158 ms | **+171.9 %** |
| 16 777 216 | 8 | 10.18 ms | 27.27 ms | +17.09 ms | **+167.9 %** |

### How to Reproduce

1. Check out branch in https://github.com/rapidsai/cudf/pull/14162, but commit 2076570046683348807e34655ef9bca7fda1c9a9.
2. Build `COPYING_NVBENCH`.
3. `COPYING_NVBENCH -b scatter --json broken.json` on any NVIDIA GPU.
4. Checkout the latest commit on that branch and run .
5. `COPYING_NVBENCH -b scatter --json fixed.json` on any NVIDIA GPU.
6. `nvbench_compare.py broken.json fixed.json`

### Expected behavior

There should be significant change in the performance when switching from `thrust::transform_iterator` vs `cuda::transform_iterator`.

### Reproduction link

_No response_

### Operating System

_No response_

### nvidia-smi output

_No response_

### NVCC version

_No response_

miscco:
We often only look for `iterator_category` to decide whether we can use more efficient implementations.

This does not play well with the C++20 iterators, which often have `input_iterator` as their `iterator_category` and the actual information stored in `iterator_concept`

Consequently we check whether something has `iterator_concept` defined and if so we use the higher of `iterator_concept` and `iterator_category`

### DIFF
--- a/libcudacxx/include/cuda/std/__algorithm/equal.h
+++ b/libcudacxx/include/cuda/std/__algorithm/equal.h
@@ -103,8 +103,8 @@ equal(_InputIterator1 __first1,
     __first2,
     __last2,
     __pred,
-    typename iterator_traits<_InputIterator1>::iterator_category(),
-    typename iterator_traits<_InputIterator2>::iterator_category());
+    __iterator_traits_category_or_concept_t<_InputIterator1>(),
+    __iterator_traits_category_or_concept_t<_InputIterator2>());
 }
 
 template <class _InputIterator1, class _InputIterator2>
@@ -117,8 +117,8 @@ equal(_InputIterator1 __first1, _InputIterator1 __last1, _InputIterator2 __first
     __first2,
     __last2,
     __equal_to{},
-    typename iterator_traits<_InputIterator1>::iterator_category(),
-    typename iterator_traits<_InputIterator2>::iterator_category());
+    __iterator_traits_category_or_concept_t<_InputIterator1>(),
+    __iterator_traits_category_or_concept_t<_InputIterator2>());
 }
 
 _CCCL_END_NAMESPACE_CUDA_STD
--- a/libcudacxx/include/cuda/std/__algorithm/fill.h
+++ b/libcudacxx/include/cuda/std/__algorithm/fill.h
@@ -48,7 +48,7 @@ __fill(_RandomAccessIterator __first, _RandomAccessIterator __last, const _Tp& _
 template <class _ForwardIterator, class _Tp>
 _CCCL_API constexpr void fill(_ForwardIterator __first, _ForwardIterator __last, const _Tp& __value_)
 {
-  ::cuda::std::__fill(__first, __last, __value_, typename iterator_traits<_ForwardIterator>::iterator_category());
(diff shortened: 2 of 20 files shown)


miscco (commented):
Note: I still need to add more tests, but I want to see whether things break first

Jacobfaib (approved):

bernhardmgruber (approved):

PR #8859, opened 2026-05-07
CUDA runtime unit test

### LINKED ISSUE #8649: Add a unit test that `cudaMallocManaged` is available and working
Our QA sometimes uses obscure machines coupled with experimental drivers to test CCCL, resulting in bug reports that take a non-trivial amount of time to determine whether there is a bug in CCCL or the CUDA runtime/driver, or anything below that.

We should add a few rudimentary tests that ensure the CUDA implementation is fundamentally working, so we have an easier way to detect if the problem is with CCCL or CUDA on that system.

The latest incarnation is NVBug 5739038, which seems to not support `cudaMallocManaged` correctly, leading to a lot of Thrust test failures (which use `thrust::universal_vector`). So we should at least add a test whether that API behaves as expected. But also:

* Can we allocate and deallocate GPU memory
* Can we run a kernel
* Can we access pinned memory on the host, device memory on the device
* Can we access managed/pinned memory as expected.

Ideally, I want those tests to run before any Thrust, CUB or libcu++ tests.

charan-003:
Closes #8649 

NVBug 5739038: CUDA runtime unit tests for CCCL.
Tests that ensure the CUDA implementation is fundamentally working


charan-003:
Hi @bernhardmgruber 

This a draft implementation to check if the approach works before expanding it further.

Checking device availability `(cudaGetDeviceCount, cudaSetDevice, cudaGetDeviceProperties)`
 
 If this looks good, i will expand the coverage to include:
- GPU memory alloc/free
- Pinned memory access
- Stream operations
- Multi-GPU basics

Please do let me know if i need to change something
Thanks!!

charan-003:
Hi @bernhardmgruber,
I wanted to follow up on this PR. Please let me know if im going in right direction 
Thanks !!

test/cuda_smoke/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
...
+cccl_get_cudatoolkit()
+
+set(_smoke_target cccl.test.cuda_runtime_smoke)
+
+cccl_add_executable(${_smoke_target} SOURCES cuda_runtime_smoke.cu ADD_CTEST)
+
+target_link_libraries(
+  ${_smoke_target}
+  PRIVATE cccl.c2h.main CUDA::cudart_static
+)
```
bernhardmgruber: Let's maybe just write the test without `c2h` and just `Catch2`


```suggestion
cccl_get_catch2()
cccl_get_cudatoolkit()

set(_smoke_target cccl.test.cuda_runtime_smoke)

cccl_add_executable(${_smoke_target} SOURCES cuda_runtime_smoke.cu ADD_CTEST)

target_link_libraries(
  ${_smoke_target}
  PRIVATE Catch2::Catch2WithMain
)
```

bernhardmgruber:
I am curious what the CI will do now :)

charan-003:
Also, this only gates CUB, libcudacxx, cudax and Thrust. OMP/TBB tests run independently

charan-003:
Fixed CI bug where the smoke binary wasn't making it into the test tarballs

Tried on A100 + V100: smoke test passes when the build includes the right CUDA architectures, if not, 209 (cudaErrorNoKernelImageForDevice) runtime error occurs with wrong arch flags

CMakeLists.txt
```diff
@@ -114,6 +114,46 @@ if (CCCL_TOPLEVEL_PROJECT)
...
+  AND
+    (
+      CCCL_ENABLE_CUB
+      OR CCCL_ENABLE_THRUST
+      OR CCCL_ENABLE_LIBCUDACXX
+      OR CCCL_ENABLE_CUDAX
+    )
+)
+  set(CCCL_CUDA_SMOKE_ENABLED ON)
+endif()
```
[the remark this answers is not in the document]
charan-003: The current gating on top-level `CCCL_ENABLE_*`  flags is intentional...
The subproject `_ENABLE_TESTING` flags are undefined at this scope until `add_subdirectory()` runs (lines 157-163). We can't check variables that don't exist yet.

bernhardmgruber:
CI logs look like its working:
```
 Running command: timeout 5.5h ctest --output-log /home/coder/cccl/build/cuda13.2-llvm21/cub-lid0/ctest.log --preset=cub-lid0
  Test project /home/coder/cccl/build/cuda13.2-llvm21/cub-lid0
          Start 239: cccl.test.cuda_runtime_smoke
    1/188 Test #239: cccl.test.cuda_runtime_smoke ...................................   Passed    0.41 sec
          Start   7: cub.test.device.adjacent_difference_custom_policy_hub.lid_0
    2/188 Test   #7: cub.test.device.adjacent_difference_custom_policy_hub.lid_0 ....   Passed    0.19 sec
```

test/cuda_smoke/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
+# Registered as a CTest setup fixture (${CCCL_CUDA_SMOKE_FIXTURE}); library
+# tests get FIXTURES_REQUIRED via the deferred walker in
+# cmake/CCCLCudaSmokeFixture.cmake, which emits ${CMAKE_BINARY_DIR}/CTestCustom.cmake.
+#
+# Declare CUDA inside this subproject only so the root project() can stay
+# CXX-only (external consumers via add_subdirectory(cccl) see no change).
+project(CCCL_CUDA_SMOKE LANGUAGES CXX CUDA)
```
bernhardmgruber: Question: Can we just use `CUDA` here?
```suggestion
# Declare CUDA inside this subproject only so the root project() can stay
# CXX-only (external consumers via add_subdirectory(cccl) see no change).
project(CCCL_CUDA_SMOKE LANGUAGES CUDA)
```

test/cuda_smoke/cuda_runtime_smoke.cu
```diff
@@ -0,0 +1,86 @@
...
+
+  cudaDeviceProp props{};
+  CUDART_REQUIRE(cudaGetDeviceProperties(&props, 0));
+  REQUIRE(props.name[0] != '\0');
+
+  REQUIRE(cudaGetLastError() == cudaSuccess);
+}
+
+// Covers the NVBug 5739038 failure : host write -> kernel transform -> host read-back
+// via a managed pointer. Fails here instead of producing dozens of thrust/cub failures.
```
bernhardmgruber: Suggestion: I think I would not need the comment.

```suggestion
```

bernhardmgruber:
@alliepiper can you please review the cmake and script changes? They seem a bit invasive, so maybe there is a simpler approach.

test/cuda_smoke/cuda_runtime_smoke.cu
charan-003: Thanks for the comments, @bernhardmgruber! I will go ahead and remove those lines once I receive feedback from @alliepiper as well... so I can address everything together. After that, I'll focus on expanding the coverage for the remaining tests.

alliepiper (changes requested):
The cmake approach is overly complex and fragile. There are global side effects to doing things like shadowing add_test, and injecting a fixture to run the smoke test on every ctest invocation of a CCCL target is too noisy.

I suggest something like:

1. build the smoke test. The current test/cuda_smoke works. I'd drop the [.smoke] tag, since the goal of running the executable is always to run those tests AIUI.
2. Provide a single CMake option to enable/disable the smoke test. Default to OFF.
3. Enable in the base CMake preset so all configs inherit. Override / set to OFF for the installation presets only.
4. Update `ci/build_common.sh` to invoke the smoke test when running gpu tests (https://github.com/NVIDIA/cccl/blob/main/ci/build_common.sh#L414-L416)

This way it runs once per CI (or QA) script invocation, is available to run as a one-off in any preset build, and leaves a much smaller footprint in the build system.

charan-003:
@alliepiper Thanks for the feedback.
If this looks good, I'll expand the test coverage to cover the remaining checks from the issue (device memory alloc/free, pinned memory, etc.).

ci/build_common.sh
```diff
@@ -413,6 +413,10 @@ function test_preset()
 
     if $GPU_REQUIRED; then
         fail_if_no_gpu
+        local smoke_bin="${BUILD_DIR}/${PRESET}/bin/cccl.test.cuda_runtime_smoke"
+        if [[ -x "$smoke_bin" ]]; then
+            run_ci_timed_command "CUDA smoke ${BUILD_NAME}" "$smoke_bin" || return $?
+        fi
```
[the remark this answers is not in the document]
charan-003: `CCCL_ENABLE_CUDA_SMOKE_TESTS` defaults to OFF, so there are valid GPU-testing configs where the binary isn't built
If the binary happens to be there, run it as a fail-fast check... if not proceed normally

charan-003:
@bernhardmgruber Should the CUDA smoke test stay in CTest, or only run from CI when there’s a GPU?
If it stays in CTest, is excluding it from thrust-cpu the right fix?


```          Start 423: cccl.test.cuda_runtime_smoke
  423/423 Test #423: cccl.test.cuda_runtime_smoke ......................................***Failed    0.00 sec
  Randomness seeded to: 1631046513
  
  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  cccl.test.cuda_runtime_smoke is a Catch2 v3.12.0 host application.
  Run with -? for options
  
  -------------------------------------------------------------------------------
  cudaMallocManaged round-trip works
  -------------------------------------------------------------------------------
  /home/coder/cccl/test/cuda_smoke/cuda_runtime_smoke.cu:40
  ...............................................................................
  
  /home/coder/cccl/test/cuda_smoke/cuda_runtime_smoke.cu:45: FAILED:
    REQUIRE( (cudaDeviceGetAttribute(&managed_supported, cudaDevAttrManagedMemory, 0)) == cudaSuccess )
  with expansion:
    35 == 0
  
  -------------------------------------------------------------------------------
  CUDA device is available
  -------------------------------------------------------------------------------
  /home/coder/cccl/test/cuda_smoke/cuda_runtime_smoke.cu:25
  ...............................................................................
  
  /home/coder/cccl/test/cuda_smoke/cuda_runtime_smoke.cu:28: FAILED:
    REQUIRE( (cudaGetDeviceCount(&device_count)) == cudaSuccess )
  with expansion:
    35 == 0
  
  ===============================================================================
  test cases: 2 | 2 failed
  assertions: 2 | 2 failed
  
  
  Errors while running CTest```

**Comment by bernhardmgruber** [comment]
> @bernhardmgruber Should the CUDA smoke test stay in CTest, or only run from CI when there’s a GPU? If it stays in CTest, is excluding it from thrust-cpu the right fix?

@alliepiper and @wmaxey do you have opinions on this? I believe I have observed QA using their own scripts to run our unit tests, bypassing the CI test scripts. So I am inclinded to have CTest also run them, but definitely disable them when there is no GPU target active (no CUB, or no Thrust with CUDA, etc.).

File: ci/build_common.sh (hunk 5 above)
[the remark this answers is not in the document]
Comment (alliepiper) [reply ci/build_common.sh]: I agree with parts of the review here -- These scripts are only used with the cmake CI presets, and all of the CI presets will enable the smoke test if they need it. If GPU_REQUIRED is true, we should expect that the smoke test exists and passes.

The issue is in how the thrust CPU test job is implemented. We should be guarding that better in our CI scripts here: https://github.com/NVIDIA/cccl/blob/main/ci/test_thrust.sh#L51-L64

```
if $CPU_ONLY; then
  PRESETS=("thrust-cpu")
  GPU_REQUIRED=false
elif $GPU_ONLY; then
  PRESETS=("thrust-gpu")
  GPU_REQUIRED=true
else
  PRESETS=("thrust")
  GPU_REQUIRED=true
fi

for PRESET in "${PRESETS[@]}"; do
  test_preset "Thrust (${PRESET})" "${PRESET}" "${GPU_REQUIRED}"
done
```

Two issues:

1. GPU_REQUIRED should be per preset, not global.
2. if neither CPU_ONLY and GPU_ONLY are specified, both `thrust-cpu` and `thrust-gpu` targets should be run, not just the `thrust` preset. They provide the same coverage, but splitting them will let us apply the GPU_REQUIRED flag correctly.

Would you mind fixing that here, and making this check required? Allow the user to set a CCCL_SKIP_CI_SMOKE env var to bypass the check if they really need to for a one-off case.

alliepiper:
> > @bernhardmgruber Should the CUDA smoke test stay in CTest, or only run from CI when there’s a GPU? If it stays in CTest, is excluding it from thrust-cpu the right fix?
> 
> @alliepiper and @wmaxey do you have opinions on this? I believe I have observed QA using their own scripts to run our unit tests, bypassing the CI test scripts. So I am inclinded to have CTest also run them, but definitely disable them when there is no GPU target active (no CUB, or no Thrust with CUDA, etc.).

https://github.com/NVIDIA/cccl/pull/8859#discussion_r3312452993 for the scripting fix regarding thrust-cpu.

For SWQA, this will be a teaching/documentation problem. We don't want CTest to run it because there's no clean way to implement it that wouldn't make it a nuisance.

bernhardmgruber (commented):
I am happy with the state of the PR. @alliepiper can you please also approve for the script changes?

charan-003:
Thanks. Right now it only covers `Device availability`
once this gets approved, i will follow the same pattern and cover:
-  GPU memory alloc/dealloc
- Managed/pinned memory
- pinned memory on host and device

please do let me know if we want to see something else as well

bernhardmgruber:
@charan-003 those additions would be great, but maybe let's get this PR merged first and propose the additions as a follow up!

alliepiper (approved):

bernhardmgruber:
Thank you for the contribution @charan-003 ! :)

PR #8861, opened 2026-05-07
[cub] Simplify arch dispatch

### LINKED ISSUE #7654: Tuning policy selectors should use the compute capability instead of `cuda::arch_id`

davebayer:
This PR uses `cuda::__target_compute_capabilities` function to simplify cub's arch dispatch. I've also added a test for `cuda::__target_compute_capabilities`, so we are sure it does what we think it should do.

### DIFF
--- a/cub/cub/detail/arch_dispatch.cuh
+++ b/cub/cub/detail/arch_dispatch.cuh
@@ -45,32 +45,18 @@ struct device_policy_getter : PolicySelector
 
 #if !defined(CUB_DEFINE_RUNTIME_POLICIES) && !_CCCL_COMPILER(NVRTC)
 #  if _CCCL_STD_VER < 2020
-template <typename PolicySelector, size_t N>
-_CCCL_API constexpr auto find_lowest_cc_with_same_policy(
-  PolicySelector policy_selector, size_t i, const ::cuda::std::array<::cuda::compute_capability, N>& all_ccs)
-  -> ::cuda::compute_capability
-{
-  const auto policy = policy_selector(all_ccs[i]);
-  while (i > 0 && policy_selector(all_ccs[i - 1]) == policy)
-  {
-    --i;
-  }
-  return all_ccs[i];
-}
-
-template <int CcMult, typename CudaArchSeq, typename PolicySelector, size_t... Is>
+template <typename CudaArchSeq, typename PolicySelector, size_t... Is>
 struct lowest_cc_resolver;
 
 // we keep the compile-time build up of the mapping table outside a template parameterized by a user-provided callable
-template <int CcMult, int... CudaCcs, typename PolicySelector, size_t... Is>
-struct lowest_cc_resolver<CcMult, ::cuda::std::integer_sequence<int, CudaCcs...>, PolicySelector, Is...>
+template <int... CudaCcs, typename PolicySelector, size_t... Is>
+struct lowest_cc_resolver<::cuda::std::integer_sequence<int, CudaCcs...>, PolicySelector, Is...>
 {
   static_assert(sizeof...(CudaCcs) == sizeof...(Is));
 
   using policy_t = decltype(PolicySelector{}(::cuda::compute_capability{}));
 
-  static constexpr ::cuda::compute_capability all_ccs[sizeof...(Is)]{
-    ::cuda::compute_capability{(CudaCcs * CcMult) / 10}...};
+  static constexpr ::cuda::compute_capability all_ccs[sizeof...(Is)]{::cuda::compute_capability{CudaCcs}...};
 
   // GCC 7 has issues reusing the constexpr array of tuning policies in find_lowest below (it loses the constexpr-ness)
 #    if _CCCL_COMPILER(GCC, >=, 8)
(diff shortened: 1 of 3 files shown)


miscco (approved):

cub/cub/detail/arch_dispatch.cuh
```diff
@@ -45,32 +45,18 @@ struct device_policy_getter : PolicySelector
...
-  PolicySelector policy_selector, size_t i, const ::cuda::std::array<::cuda::compute_capability, N>& all_ccs)
-  -> ::cuda::compute_capability
-{
-  const auto policy = policy_selector(all_ccs[i]);
-  while (i > 0 && policy_selector(all_ccs[i - 1]) == policy)
-  {
-    --i;
-  }
-  return all_ccs[i];
-}
```
bernhardmgruber, line 59: Me: Wait, this was always unused?

Dave: always has been

*shot fired*

bernhardmgruber (approved):
I love it!

cub/cub/detail/arch_dispatch.cuh
davebayer: Yeah, I think we should rather remove it, if we ever need it later for some reason, it can be always found it the repo history

bernhardmgruber: I think I added this while prototyping and just completely forgot about deleting it. I don't need it!

PR #8862, opened 2026-05-07
[cub] Add future arch test

davebayer:
This PR adds a compile test that checks whether CUB can be used with a future (unknown) architecture. I had to do some awful hacks to make it work.

### DIFF
--- a/cub/test/test_future_arch.cu
+++ b/cub/test/test_future_arch.cu
@@ -1,0 +1,34 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+// This is a compile-only test that checks whether cub can handle a future (unknown) architecture. Currently, works only
+// with nvcc.
+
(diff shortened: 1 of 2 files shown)


cub/test/test_future_arch.cu
```diff
@@ -0,0 +1,34 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+// This is a compile-only test that checks whether cub can handle a future (unknown) architecture. Currently, works only
+// with nvcc.
+
+#if defined(__NVCC__)
+#  define TEST_CUDA_ARCH 1000000
```
bernhardmgruber: Q: Are we depending on the magnitude of this integer anywhere? I would have chosen a smaller value. SM120 has a value of `1200`, so I guess `9900` (`sm_990`) would have been large enough for a while.

davebayer: Sounds reasonable

bernhardmgruber (approved):

PR #8826, opened 2026-05-05
Use the new tuning API internally for `detail::reduce[_nd]::dispatch[_nd]`

bernhardmgruber:
- [x] Merge before: #8851
- [x] #8884
- [x] #8897
- [x] No SASS changes on `cub.bench.reduce.sum.base` on SM`75;100`
- [x] No SASS changes on `cub.bench.reduce.nondeterministic.base` on SM`75;100`
- [x] No SASS changes on `cub.bench.reduce.deterministic.base` on SM`75;86;90`
- [x] No SASS changes on `cub.bench.reduce.arg_extrema.base` on SM`75;86;90` (diffed with #8897)

Fixes: #7948
Fixes: #7951
Fixes: #7960
Fixes: #7957

PR #8863, opened 2026-05-07
Allow public tuning of non-ByKey `cub::DeviceReduce`

### LINKED ISSUE #7577: [FEA]: Productize `cub::DeviceReduce` tuning API
### Is this a duplicate?

### Area

CUB

Current tuning API for reduce is implementation detail, so unavailable to users. 

We need public reduce tuning API (TODO: populate @bernhardmgruber)

_No response_

### Additional context

_No response_

bernhardmgruber:
- [x] #8826
- [x] add more `Device reduce can be tuned` tests that cover the two other determinisms
- [x] #7671 is sufficiently approved to proceed with this PR. The remaining discussion can continue here.
- [x] #9349
- [x] #9639
- [x] Bikeshedding
- [x] Remove the `My` from the policy selector names in the API example
- [x] We should not need to provide multiple policy selectors to `DeviceReduce::Sum` and rather use a struct with appropriate sub members, like `DeviceTransform`. There is not actually a difference between the two cases (as I tried to argue).
- [x] Deprecate the agent policy
- [x] Rename `block_algorithm` to `reduce_algorithm`.

Fixes: #7577

### DIFF
--- a/docs/cub/tuning.rst
+++ b/docs/cub/tuning.rst
@@ -43,13 +43,13 @@ taking a ``cuda::compute_capability`` and returning the algorithm's policy struc
 .. code:: c++
 
     template <typename T>
-    struct CustomReducePolicySelector {
+    struct my_reduce_tuning {
       __host__ __device__ constexpr auto operator()(cuda::compute_capability cc) const -> cub::ReducePolicy {
         // tuning for Hopper and later
         if (cc >= cuda::compute_capability(9, 0)) {
           const auto pass = cub::ReducePassPolicy{
             .threads_per_block = 512,
-            .items_per_thread = 64 / sizeof(T), // 8 double, 16 float, 32 half_t, ...
+            .items_per_thread = std::max(64 / sizeof(T), 1), // 8 double, 16 float, 32 half_t, ...
             .vec_size = 2,
             .reduce_algorithm = cub::BLOCK_REDUCE_WARP_REDUCTIONS,
             .load_modifier = cub::LOAD_DEFAULT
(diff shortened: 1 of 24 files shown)


docs/cub/tuning.rst
```diff
@@ -40,34 +40,31 @@ Defining a policy selector
...
 taking a ``cuda::compute_capability`` and returning the algorithm's policy struct:
 
-..
-    TODO(bgruber): we have to update this example again after the reduce tuning API has been actually released
-
 .. code:: c++
 
     template <typename T>
-    struct my_reduce_tuning {
+    struct CustomReducePolicySelector {
```
NaderAlAwar: Important: other locations in this document still use `my_reduce_tuning` instead of `CustomReducePolicySelector`

docs/cub/tuning.rst
```diff
@@ -40,34 +40,31 @@ Defining a policy selector
...
       __host__ __device__ constexpr auto operator()(cuda::compute_capability cc) const -> cub::ReducePolicy {
         // tuning for Hopper and later
         if (cc >= cuda::compute_capability(9, 0)) {
-          const auto policy = cub::detail::agent_reduce_policy{
+          const auto pass = cub::ReducePassPolicy{
             .threads_per_block = 512,
             .items_per_thread = 64 / sizeof(T), // 8 double, 16 float, 32 half_t, ...
```
NaderAlAwar: Suggestion: this should be clamped to at least 1

NaderAlAwar (approved):
Important: we still need to update `device_scope.rst` since it uses reduce policy tuning as an example. Line 247 should say `multi_tile` instead of reduce and line 458 should return `cub::ReducePolicy`

docs/cub/tuning.rst
bernhardmgruber: I'll name it back to `my_reduce_tuning`

PR #8866, opened 2026-05-07
[libcu++] Make `cuda::compute_capability` structural type

### LINKED ISSUE #8730: [BUG/FEA]: `cuda::compute_capability` is not a structural type
We cannot pass `cuda::compute_capability` as a template parameter in C+++20 because the type is not a _structural type_. This is unfortunate, because that type is conceptually a strongly typed `int`. The only thing preventing it from being structural is that the `int __cc_;` data member is private.

I propose to make the data member public. Given the prefix `__` it is still an implementation detail from a user's perspective. This is also what `std::array` does, which has a public member array that is an implementation detail.

davebayer:
Fixes #8730

### DIFF
--- a/libcudacxx/test/libcudacxx/cuda/ccclrt/device/compute_capability.pass.cpp
+++ b/libcudacxx/test/libcudacxx/cuda/ccclrt/device/compute_capability.pass.cpp
@@ -233,6 +233,14 @@ TEST_FUNC constexpr bool test()
     assert(!(cc2 > cc1));
   }
 
+  // 12. Test that cuda::compute_capability is a structural type.
+#if _CCCL_STD_VER >= 2020
+  {
+    [[maybe_unused]] constexpr auto val =
+      cuda::std::integral_constant<cuda::compute_capability, cuda::compute_capability{100}>{};
+  }
+#endif // _CCCL_STD_VER >= 2020
+
   return true;
 }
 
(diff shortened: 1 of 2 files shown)


bernhardmgruber (approved):

Jacobfaib (approved):

libcudacxx/test/libcudacxx/cuda/ccclrt/device/compute_capability.pass.cpp
```diff
@@ -233,6 +233,13 @@ TEST_FUNC constexpr bool test()
     assert(!(cc2 > cc1));
   }
 
+  // 12. Test that cuda::compute_capability is a structural type.
+#if _CCCL_STD_VER >= 2020
+  {
+    using T [[maybe_unused]] = cuda::std::integral_constant<cuda::compute_capability, cuda::compute_capability{100}>;
```
bernhardmgruber: Suggestion: consider:

```suggestion
    [[maybe_unused]] constexpr auto val = cuda::std::integral_constant<cuda::compute_capability, cuda::compute_capability{100}>;
```
I think alias declarations may not be required to be resolved by the compiler immediately, so template instantiation could be delayed.

davebayer: From what I've tested, it actually failed before I moved the `public:` label, so I believe it is instantiated. I've used your suggestion anyway.

PR #8868, opened 2026-05-07
[libcu++] Ensure that cuda iterators work with all host only types

### LINKED ISSUE #8382: [BUG]: `cuda::transform_iterator` triggers nvcc warning 20011/20013 in mixed host/device TUs
### Is this a duplicate?

### Type of Bug

Compile-time Error

### Component

libcu++

### Describe the bug

`cuda::transform_iterator` uses an internal helper (`__compressed_movable_box`) that defines its destructor as both `__host__` and `__device__`. Because of that, the CUDA compiler (`nvcc`) ends up checking *all* types in the same `.cu` file as if they might run on both the CPU and GPU.

This becomes a problem when you also use types that are meant to work only on the CPU (like `std::unique_ptr` or `std::vector`). Even though you’re not trying to use them on the GPU, nvcc still analyzes them in that context and can raise misleading or false errors.

On the other hand, `thrust::transform_iterator` avoids this issue because its special member functions (like the destructor) are just defaulted (`= default`). That doesn’t force them to be treated as both host and device code, so nvcc doesn’t over-analyze unrelated types.

### How to Reproduce

I haven't put together a minimal repro yet, but the issue is fairly easy to reproduce within cuDF. If you replace the `thrust::transform_iterator` use cases in https://github.com/rapidsai/cudf/blob/main/cpp/include/cudf/column/column_device_view.cuh with `cuda::transform_iterator` and build cudf, the error will show up.

### Expected behavior

The build should pass.

miscco:
We had reports that `__compressed_movable_box` was not friendly to host only types that were not trivially destructible

This ensures that we add the proper execution suppression macros and also adds a bunch of tests to verify that the cuda iterators work with host only types

Fixes #8382

### DIFF
--- a/libcudacxx/include/cuda/__iterator/constant_iterator.h
+++ b/libcudacxx/include/cuda/__iterator/constant_iterator.h
@@ -148,6 +148,7 @@ public:
   }
 
   //! @brief Increments the stored index
+  _CCCL_EXEC_CHECK_DISABLE
   _CCCL_API constexpr constant_iterator& operator++() noexcept
   {
     ++__index();
@@ -164,6 +165,7 @@ public:
   }
 
   //! @brief Decrements the stored index
+  _CCCL_EXEC_CHECK_DISABLE
   _CCCL_API constexpr constant_iterator& operator--() noexcept
   {
     if constexpr (::cuda::std::is_signed_v<_Index> || !::cuda::std::is_integral_v<_Index>)
@@ -189,6 +191,7 @@ public:
 
   //! @brief Advances a @c constant_iterator by a given number of elements
   //! @param __n The amount of elements to advance
+  _CCCL_EXEC_CHECK_DISABLE
   _CCCL_API constexpr constant_iterator& operator+=(difference_type __n) noexcept
   {
     if constexpr (::cuda::std::is_signed_v<_Index> || !::cuda::std::is_integral_v<_Index>)
@@ -202,6 +205,7 @@ public:
   //! @brief Creates a copy of a @c constant_iterator advanced by a given number of elements
   //! @param __iter The @c constant_iterator to advance
   //! @param __n The amount of elements to advance
+  _CCCL_EXEC_CHECK_DISABLE
   template <int = 0> // Must be template, or the compiler complains about a nonliteral return type
   [[nodiscard]] _CCCL_API friend constexpr constant_iterator operator+(
     const constant_iterator& __iter, difference_type __n) noexcept(::cuda::std::is_nothrow_copy_constructible_v<_Tp>)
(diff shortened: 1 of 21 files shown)


Jacobfaib (approved):

davebayer (approved):

libcudacxx/test/libcudacxx/cuda/iterators/shuffle_iterator/host_only_types.pass.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+struct host_bijection
+{
+  using index_type = uint32_t;
+
+  host_bijection() {}
+  host_bijection(const host_bijection&) {}
+  host_bijection(host_bijection&&) {}
+  host_bijection& operator=(const host_bijection&) {}
+  host_bijection& operator=(host_bijection&&) {}
```
davebayer: Clang is unhappy about those, applies to other files, too
```suggestion
  host_bijection& operator=(const host_bijection&) { return *this; }
  host_bijection& operator=(host_bijection&&) { return *this; }
```

miscco: :facepalm: Lets just quickly add them ^^

bernhardmgruber (approved):

libcudacxx/include/cuda/std/__ranges/compressed_movable_box.h
```diff
@@ -202,6 +204,8 @@ struct _CCCL_DECLSPEC_EMPTY_BASES __compressed_box_storage
     _CCCL_API constexpr __storage(in_place_t, _Args&&... __args) noexcept(is_nothrow_constructible_v<_Tp, _Args...>)
         : __val_(::cuda::std::forward<_Args>(__args)...)
     {}
+
+    _CCCL_EXEC_CHECK_DISABLE
```
Jacobfaib, line 208: Might as well remove these if you blanket disable in the file

Jacobfaib (approved):

libcudacxx/include/cuda/std/__ranges/compressed_movable_box.h
miscco: I was wondering about that too, but wanted to be more explicit

libcudacxx/include/cuda/__iterator/constant_iterator.h
```diff
@@ -22,6 +22,7 @@
 #endif // no system header
 
 #include <cuda/__fwd/iterator.h>
+#include <cuda/std/__concepts/constructible.h>
```
davebayer, line 25: Is this used?

libcudacxx/include/cuda/__iterator/shuffle_iterator.h
```diff
@@ -44,6 +44,9 @@
 
 _CCCL_BEGIN_NAMESPACE_CUDA
 
+_CCCL_BEGIN_NV_DIAG_SUPPRESS(20011) // calling a __host__ function from a __host__ __device__ function shuffle_iterator
+                                    // is not allowed
```
davebayer, line 48: `_CCCL_EXEC_CHECK_DISABLE` are not sufficient in this file?

davebayer (approved):
lgtm

libcudacxx/include/cuda/__iterator/shuffle_iterator.h
miscco: Unfortunately not, I dont really know why, but I am getting the copy constructor flagged

libcudacxx/include/cuda/__iterator/constant_iterator.h
miscco: yeah it was not included properly

PR #8873, opened 2026-05-07
`cuda::std::simd`  Optimize small integer operations

fbusato:
## Description

The PR introduces the following optimizations:

- Bitwise And/Or/Xor: single operation, any GPU arch:

  - u16xN -> N/2 ops
  - u8xN  -> N/4 ops
- `VIADD` (SIMD vector add), extended to `++x`, `--x`, `(-x)`, `x - y`:

  - u16xN -> N/2 ops, SM90+
  - u8xN  -> N/4 ops, SM120f

Verified that the compiler is not able to automatically generate them, tested with CTK 13.2.


libcudacxx/include/cuda/std/__simd/specializations/fixed_size_integral_vec.h
```diff
@@ -0,0 +1,268 @@
...
+  __is_fixed_size_small_integral_v<_Tp, _Np> ? __simd_operations_small_integral : __simd_operations_kind::__default;
+
+// Simd operations for fixed_size ABI with small integral element types.
+template <typename _Tp, __simd_size_type _Np>
+struct __simd_operations<_Tp, __fixed_size<_Np>, __simd_operations_small_integral> : __fixed_size_operations<_Tp, _Np>
+{
+  using __base           = __fixed_size_operations<_Tp, _Np>;
+  using __simd_storage_t = __simd_storage<_Tp, __fixed_size<_Np>>;
+
+  static_assert(alignof(__simd_storage_t) >= alignof(uint32_t));
```
Jacobfaib, line 78: Could use a comment explaining why, `alignof(uint32_t)` is magic

libcudacxx/include/cuda/std/__simd/specializations/vadd_intrinsics.h
```diff
@@ -0,0 +1,127 @@
...
+
+#include <cuda/std/__cccl/prologue.h>
+
+_CCCL_BEGIN_NAMESPACE_CUDA_STD_SIMD
+
+[[nodiscard]] _CCCL_DEVICE_API inline uint32_t __vadd2_u16(const uint32_t __lhs, const uint32_t __rhs) noexcept
+{
+  NV_IF_TARGET(NV_PROVIDES_SM_90,
+               (return ::__vadd2(__lhs, __rhs);), //
+               (_CCCL_ASSERT(false, "Unsupported architecture"); return uint32_t{};));
```
Jacobfaib: Important: use `_CCCL_VERIFY()` instead, it fires regardless of whether debugging is enabled.

libcudacxx/include/cuda/std/__simd/specializations/vadd_intrinsics.h
```diff
@@ -0,0 +1,127 @@
...
+  return __result;
+}
+
+#if _CCCL_HAS_SIMD_VADD_8B()
+
+[[nodiscard]] _CCCL_DEVICE_API inline uint32_t __vadd4_u8(const uint32_t __lhs, const uint32_t __rhs) noexcept
+{
+#  if _CCCL_HAS_SIMD_VADD_8B_INTRINSICS()
+  return ::__vadd4(__lhs, __rhs);
+#  else
```
Jacobfaib: I believe @miscco would want a `// ^^^ / vvv` comment here :)

libcudacxx/include/cuda/std/__simd/type_traits.h
```diff
@@ -32,16 +32,23 @@
 
 _CCCL_BEGIN_NAMESPACE_CUDA_STD_SIMD
 
+#if _CCCL_CTK_AT_LEAST(12, 9)
+inline constexpr size_t __optimal_cuda_alignment = 32;
+#else
+inline constexpr size_t __optimal_cuda_alignment = 16;
+#endif // _CCCL_CTK_AT_LEAST(12, 9)
```
Jacobfaib:
```suggestion
inline constexpr size_t __optimal_cuda_alignment = _CCCL_CTK_AT_LEAST(12, 9) ? 32 : 16;
```
No need to involve `#if` and `#endif` here

libcudacxx/test/libcudacxx/std/numerics/simd/simd.traits/alignment.pass.cpp
```diff
@@ -23,27 +23,33 @@
 
 namespace simd = cuda::std::simd;
 
-template <typename T, int N, size_t ExpectedAlign = alignof(T) * N>
+#if _CCCL_CTK_AT_LEAST(12, 9)
+inline constexpr size_t optimal_cuda_alignment = 32;
+#else
+inline constexpr size_t optimal_cuda_alignment = 16;
+#endif // _CCCL_CTK_AT_LEAST(12, 9)
```
Jacobfaib: Same here with ternary operator

libcudacxx/include/cuda/std/__simd/specializations/fixed_size_storage.h
```diff
@@ -36,9 +37,16 @@ struct __fixed_size
...
+// However, SIMD storage could spill from register to cache/memory. This could break the alignment of the data for
+// vectorized instructions. For this reason, we align the SIMD storage to at least 8 bytes (max SIMD instruction size).
+// 8 bytes is a negligible constraint in case of spilling.
+template <typename _Tp, __simd_size_type _Np, size_t _TotalAlignment = alignof(_Tp) * _Np>
+inline constexpr size_t __simd_storage_alignment_v = ::cuda::std::max(alignof(_Tp), size_t{8});
+
 // Element-per-slot simd storage for fixed_size ABI
 template <typename _Tp, __simd_size_type _Np>
-struct __simd_storage<_Tp, __fixed_size<_Np>>
+struct alignas(__simd_storage_alignment_v<_Tp, _Np>) __simd_storage<_Tp, __fixed_size<_Np>>
```
Jacobfaib: I can't tell from the tests but should probably be a test of the alignment of this structure? Since it is explicitly specified? Is that done in `libcudacxx/test/libcudacxx/std/numerics/simd/simd.traits/alignment.pass.cpp`?

Jacobfaib (approved):

libcudacxx/include/cuda/std/__simd/specializations/vadd_intrinsics.h
fbusato: it should work as long as the function is not compiled outside SM90+

libcudacxx/include/cuda/std/__simd/type_traits.h
fbusato: I never saw this usage of this macro. I will try it, thanks

libcudacxx/include/cuda/std/__simd/specializations/fixed_size_storage.h
fbusato: actually, this has been ported from #8252. I don't think this should be part of the unit test because it is an implementation-defined optmization. The standard doesn't mandate anything for the alignment.

libcudacxx/include/cuda/std/__simd/specializations/vadd_intrinsics.h
Jacobfaib: Yeah but `_CCCL_ASSERT()` does nothing (it expands empty) if `-DNDEBUG` is defined, and so this just becomes `return uint32_t{}`. That code-path is _always_ wrong, no matter what happens and so we should abort if it is ever entered. `_CCCL_VERIFY()` is like `_CCCL_ASSERT()` except it never goes away, it always checks the condition.

libcudacxx/include/cuda/std/__simd/specializations/fixed_size_storage.h
fbusato: let me port your last two comments on the other PR

Jacobfaib: > The standard doesn't mandate anything for the alignment.

Perhaps not, but we would unexpectedly lose performance if this alignment went away no? And trying to debug why that perf was lost is going to be very tricky.

libcudacxx/include/cuda/std/__simd/type_traits.h
Jacobfaib: All of the `_CCCL` macros expand to a boolean expression or integer literal. You could even, if so inclined, use them in regular code:
```c++
if (_CCCL_CTK_AT_LEAST(12, 9)) {
  frobnicate();
}
```

libcudacxx/include/cuda/std/__simd/specializations/fixed_size_integral_vec.h
```diff
@@ -0,0 +1,252 @@
...
+      const auto __rhs_u = ::cuda::std::simd::__to_unsigned_storage(__rhs);
+      _CCCL_PRAGMA_UNROLL_FULL()
+      for (__simd_size_type __i = 0; __i < __usize; ++__i)
+      {
+        __result_u[__i] = __lhs_u[__i] & __rhs_u[__i];
+      }
+      return ::cuda::std::simd::__copy_from_unsigned_storage<__simd_storage_t>(__result_u);
+    }
+    return __fixed_size_operations<_Tp, _Np>::__bitwise_and(__lhs, __rhs);
+  }
```
miscco: Question: I believe all those binary operations are identical, so we could have a macro that takes a `__bitwise_and` name and an operation `&` and synthesizes those binary operations

libcudacxx/include/cuda/std/__simd/specializations/fixed_size_integral_vec.h
```diff
@@ -0,0 +1,252 @@
...
+      if constexpr (sizeof(_Tp) == 2)
+      {
+        NV_IF_TARGET(NV_PROVIDES_SM_90, (__s = __plus(__s, __minus_one); return;))
+      }
+#  if _CCCL_HAS_SIMD_8BIT()
+      else
+      {
+        NV_IF_TARGET(NV_HAS_FEATURE_SM_120f, (__s = __plus(__s, __minus_one); return;))
+      }
+#  endif // _CCCL_HAS_SIMD_8BIT()
```
miscco, line 142: Critical: The branch is not constrained, I believe this is meant to work with `sizeof(_Tp) == 1`

libcudacxx/include/cuda/std/__simd/specializations/simd_intrinsics_array.h
```diff
@@ -0,0 +1,133 @@
...
+
+template <typename _Tp, __simd_size_type _Np>
+inline constexpr size_t __simd_storage_copy_size_u32<__simd_storage<_Tp, __fixed_size<_Np>>> = _Np * sizeof(_Tp);
+
+template <typename _SimdStorage, typename _SimdStorageU32 = __simd_storage_u32_t<_SimdStorage>>
+[[nodiscard]] _CCCL_API constexpr _SimdStorageU32 __to_unsigned_storage(const _SimdStorage& __s) noexcept
+{
+  _SimdStorageU32 __tmp{};
+  const auto __input_data = ::cuda::std::assume_aligned<alignof(uint32_t)>(__s.__data);
+  ::cuda::std::memcpy(__tmp.data(), __input_data, __simd_storage_copy_size_u32<_SimdStorage>);
```
miscco, line 64: Question: This is effectively a bit_cast, should we try and use `cuda::std::bit_cast` instead?

libcudacxx/include/cuda/std/__simd/type_traits.h
miscco: Wrong comment, that clause is
```suggestion
#endif // _CCCL_CTK_BELOW(12, 9)
```

libcudacxx/test/libcudacxx/std/numerics/simd/simd.traits/alignment.pass.cpp
miscco: Ditto that clause is
```suggestion
#endif // _CCCL_CTK_BELOW(12, 9)
```

libcudacxx/include/cuda/std/__simd/specializations/simd_intrinsics_array.h
fbusato: we cannot here, because `__simd_storage_copy_size_u32<_SimdStorage>` could be different from the input or output size

libcudacxx/include/cuda/std/__simd/specializations/fixed_size_integral_vec.h
fbusato: could be. It is only  repeated  three times so the benefits is not huge

libcudacxx/include/cuda/std/__simd/specializations/fixed_size_integral_vec.h
fbusato: not really, the code can be executed only with `sizeof(_Tp) == 1` or `sizeof(_Tp) == 2`. Anyway, adding the condition makes the code clearer

PR #8876, opened 2026-05-08
Fix wrong order of `NV_DISPATCH_TARGET` clauses

bernhardmgruber:
Fixes nvbug: 6156669


thrust/testing/scan.cu
```diff
@@ -15,7 +15,7 @@
 #include <numeric>
 
 #include <unittest/unittest.h>
-
+#if 0
```
oleksandr-pavlyk: Critical: Was this section (L18-L792) disabled intentionally?

bernhardmgruber: Absolutely not. Thx! Must have been brain-afk when commit this.

oleksandr-pavlyk (approved):

davebayer (approved):

PR #8880, opened 2026-05-08
Use the new tuning API internally for `detail::select::dispatch` and `DeviceSelect`

### LINKED ISSUE #8379: Use the new tuning API internally for `detail::select::dispatch`

bernhardmgruber:
- [x] #8896
- [x] No SASS changes in `cub.bench.select.flagged.base` on SM`75;80;86;90;100`
- [x] No SASS changes in `cub.bench.select.if.base` on SM`75;80;86;90;100`
- [x] No SASS changes in `cub.bench.select.unique.base` on SM`75;80;86;90;100`
- [x] No SASS changes in `thrust.test.unique` on SM120 (quick sanity check)
- [x] No SASS changes in `thrust.test.copy` on SM120 (quick sanity check)

Fixes: #8379

### DIFF
--- a/cub/benchmarks/bench/select/flagged.cu
+++ b/cub/benchmarks/bench/select/flagged.cu
@@ -19,75 +19,29 @@
 // %RANGE% TUNE_L2_WRITE_LATENCY_NS l2w 0:1200:5
 
 #if !TUNE_BASE
-#  if TUNE_TRANSPOSE == 0
-#    define TUNE_LOAD_ALGORITHM cub::BLOCK_LOAD_DIRECT
-#  else // TUNE_TRANSPOSE == 1
-#    define TUNE_LOAD_ALGORITHM cub::BLOCK_LOAD_WARP_TRANSPOSE
-#  endif // TUNE_TRANSPOSE
-
-#  if TUNE_LOAD == 0
-#    define TUNE_LOAD_MODIFIER cub::LOAD_DEFAULT
-#  else // TUNE_LOAD == 1
-#    define TUNE_LOAD_MODIFIER cub::LOAD_CA
-#  endif // TUNE_LOAD
-
 template <typename InputT>
-struct policy_hub_t
+struct bench_policy_selector
 {
-  struct policy_t : cub::ChainedPolicy<300, policy_t, policy_t>
+  [[nodiscard]] _CCCL_API constexpr auto operator()(cuda::compute_capability) const
+    -> cub::detail::select::select_if_policy
   {
-    using SelectIfPolicyT =
-      cub::AgentSelectIfPolicy<TUNE_THREADS_PER_BLOCK,
-                               TUNE_ITEMS_PER_THREAD,
-                               TUNE_LOAD_ALGORITHM,
-                               TUNE_LOAD_MODIFIER,
-                               cub::BLOCK_SCAN_WARP_SCANS,
-                               delay_constructor_t>;
-  };
-
-  using MaxPolicy = policy_t;
+    return {TUNE_THREADS_PER_BLOCK,
+            TUNE_ITEMS_PER_THREAD,
+            (TUNE_TRANSPOSE == 0 ? cub::BLOCK_LOAD_DIRECT : cub::BLOCK_LOAD_WARP_TRANSPOSE),
+            (TUNE_LOAD == 0 ? cub::LOAD_DEFAULT : cub::LOAD_CA),
+            cub::BLOCK_SCAN_WARP_SCANS,
+            delay_constructor_policy};
+  }
 };
 #endif // !TUNE_BASE
 
-template <typename T, typename OffsetT, typename InPlace>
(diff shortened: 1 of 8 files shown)


cub/benchmarks/bench/select/unique.cu
```diff
@@ -19,136 +19,83 @@
...
-
-// The implementation of DeviceSelect for 64-bit offset types uses a streaming approach, where it runs multiple passes
-// using a 32-bit offset type, so we only need to test one (to save time for tuning and the benchmark CI).
-using select_offset_types = nvbench::type_list<int64_t>;
-
-NVBENCH_BENCH_TYPES(unique, NVBENCH_TYPE_AXES(fundamental_types, select_offset_types, is_in_place))
+NVBENCH_BENCH_TYPES(unique, NVBENCH_TYPE_AXES(fundamental_types))
   .set_name("base")
-  .set_type_axes_names({"T{ct}", "OffsetT{ct}", "InPlace{ct}"})
+  .set_type_axes_names({"T{ct}"})
```
bernhardmgruber: I am dropping the `InPlace` option here, since there is no public API for it and we also have no tunings for it.

oleksandr-pavlyk:
Compilation error for some compiler/standard combination:

```
    /home/coder/cccl/lib/cmake/thrust/../../../thrust/thrust/system/cuda/detail/remove.h:41:13: error: no template named 'copy_if' in namespace 'thrust::cuda_cub::detail'; did you mean simply 'copy_if'?
```

bernhardmgruber:
There are SASS diffs for `cub.bench.select.unique.base` on all SM versions

cub/benchmarks/bench/select/flagged.cu
```diff
@@ -19,73 +19,25 @@
...
-  };
-
-  using MaxPolicy = policy_t;
+    return {TUNE_THREADS_PER_BLOCK,
+            TUNE_ITEMS_PER_THREAD,
+            (TUNE_TRANSPOSE == 0 ? cub::BLOCK_LOAD_DIRECT : cub::BLOCK_LOAD_WARP_TRANSPOSE),
+            (TUNE_LOAD == 0 ? cub::LOAD_DEFAULT : cub::LOAD_CA),
+            cub::BLOCK_SCAN_WARP_SCANS,
+            delay_constructor_policy};
+  }
```
miscco, line 34: I love this

bernhardmgruber:
> There are SASS diffs for `cub.bench.select.unique.base` on all SM versions

They were caused by removing the InPlace benchmarks. I split out the commit dropping them and then diffed that commit with HEAD. No SASS changes now.

thrust/thrust/system/cuda/detail/unique.h
```diff
@@ -218,9 +136,8 @@ template <class Derived, class ForwardIt, class BinaryPred>
 ForwardIt _CCCL_HOST_DEVICE
 unique(execution_policy<Derived>& policy, ForwardIt first, ForwardIt last, BinaryPred binary_pred)
 {
-  THRUST_CDP_DISPATCH(
-    (return detail::select_unique<cub::SelectImpl::SelectPotentiallyInPlace>(policy, first, last, first, binary_pred);),
```
bernhardmgruber, line 222: TODO: Thrust needs an in-place unique, for which there is not CUB public API. We should add it and revert the removal of this use case from the benchmarks

cub/benchmarks/bench/select/unique.cu
bernhardmgruber: Added back.

thrust/thrust/system/cuda/detail/unique.h
bernhardmgruber: Added in #8896

cub/benchmarks/bench/select/flagged.cu
```diff
@@ -98,38 +52,52 @@ void select(nvbench::state& state, nvbench::type_list<T, OffsetT, InPlace>)
...
+    {
+      _CCCL_TRY_CUDA_API(
+        cub::DeviceSelect::Flagged,
+        "DeviceSelect::Flagged failed",
+        d_in,
+        d_flags,
+        d_num_selected,
+        static_cast<offset_t>(elements),
+        env);
+    }
```
NaderAlAwar, line 88: Critical: for all new in place variant benchmarks, the input values will change between runs. This will affect the performance reported. We should maintain the original values and copy them in each iteration. We can use `nvbench::exec_tag::timer` to avoid timing the copy

bernhardmgruber: That's a valid observation! Now I wonder what the original benchmark's use case was? For the in-place versions it measured the in-place implementation but on distinct input and output ranges, a case not reachable with the public CUB API.

miscco (approved):

davebayer (approved):

PR #8881, opened 2026-05-08
[CUB] Do not use raw kernels

miscco:
we should not use raw `__global__` but always use `_CCCL_KERNEL_ATTRIBUTES`


bernhardmgruber (approved):

miscco:
closing in favor of #8882

PR #8885, opened 2026-05-08
Specialize `basic_common_reference` and `common_type` for tuples

Jacobfaib:
## Description

Requires https://github.com/NVIDIA/cccl/pull/8886

Add the C++23 `common_type` specialization for tuples (https://en.cppreference.com/cpp/utility/tuple/common_type).

This is needed for `zip_view` such that `const`-qualified ranges may be used.
```c++
const std::vector<int> v;

// Currently fails to compile because cuda::std::views::zip() does not model input_range
// because cuda::std::common_reference_with<ref_t&&, val_t&>, where:
//
// ref_t = cuda::std::iter_reference_t<iter_t>; // cuda::std::tuple<const int &, const int &>
// val_t = cuda::std::iter_value_t<iter_t>; // cuda::std::tuple<int, int>
//
// Is not satisfied
cuda::std::views::zip(v, v) | cuda::std::views::transform(...);
```

## Checklist

### DIFF
--- a/libcudacxx/include/cuda/std/__tuple_dir/tuple.h
+++ b/libcudacxx/include/cuda/std/__tuple_dir/tuple.h
@@ -29,6 +29,7 @@
 #include <cuda/std/__tuple_dir/tuple_leaf.h>
 #include <cuda/std/__tuple_dir/tuple_size.h>
 #include <cuda/std/__tuple_dir/tuple_types.h>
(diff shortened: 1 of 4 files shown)


Jacobfaib:
Folding this PR into https://github.com/NVIDIA/cccl/pull/8886 as this one requires 8886 anyways.

PR #8886, opened 2026-05-08
Add C++23 lvalue reference conversion constructors for tuple

Jacobfaib:
## Description

Add C++23 lvalue conversion constructors to `cuda::std::tuple`. This is _also_ needed by `zip_view` as, in conjunction with https://github.com/NVIDIA/cccl/pull/8885, it now means that in:
```c++
std::vector<int> v;

cuda::std::views::zip(v, v) | cuda::std::views::transform(...);
// Internally above, the zip_view iterator creates and holds a `tuple<int, int>`
// which it passes to the transform as `tuple<int, int> &`. The transform, however,
// expects to construct a `tuple<int&, int&>` which is _legal_ from a reference but
// currently cannot be done
```

## Checklist

### DIFF
--- a/libcudacxx/test/libcudacxx/std/utilities/tuple/tuple.tuple/tuple.cnstr/convert_non_const_copy.pass.cpp
+++ b/libcudacxx/test/libcudacxx/std/utilities/tuple/tuple.tuple/tuple.cnstr/convert_non_const_copy.pass.cpp
@@ -174,14 +174,7 @@ TEST_FUNC constexpr bool test()
   {
     cuda::std::tuple<TracedCopyMove, TracedCopyMove> t1{};
     cuda::std::tuple<TracedCopyMove, TracedCopyMove> t2{t1};
-    // NOTE: LLVM's test has this as `assert(nonConstCopyCtrCalled())`, but I think this is
-    // wrong. One of the constraints on the lvalue-ref constructor in P2165 is that:
-    //
-    // - different-from<UTuple, tuple> [range.utility.helpers] is true
-    //
-    // So given equal objects, the t2{t1} constructor call should instead select the
-    // copy-constructor `tuple(const tuple& other)` (which our implementation does).
-    assert(!nonConstCopyCtrCalled(cuda::std::get<0>(t2)));
+    assert(nonConstCopyCtrCalled(cuda::std::get<0>(t2)));
   }
 
   // These two test points cause gcc to ICE, because it cannot follow the (intentionally)
--- a/libcudacxx/test/support/copy_move_types.h
+++ b/libcudacxx/test/support/copy_move_types.h
@@ -1,0 +1,489 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef TEST_STD_UTILITIES_TUPLE_CNSTR_TYPES_H
+#define TEST_STD_UTILITIES_TUPLE_CNSTR_TYPES_H
+
+#include <cuda/std/tuple>
+#include <cuda/std/type_traits>
+#include <cuda/std/utility>
+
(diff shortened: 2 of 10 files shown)


libcudacxx/test/libcudacxx/std/utilities/tuple/tuple.tuple/tuple.cnstr/convert_non_const_copy.pass.cpp
```diff
@@ -0,0 +1,204 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
```
miscco, line 5: Missing our copyright, same other test

libcudacxx/test/support/copy_move_types.h
```diff
@@ -0,0 +1,489 @@
...
+      , alloc_constructed(true)
+  {}
+#endif
+};
+
+#if !TEST_COMPILER(NVRTC)
+template <>
+struct cuda::std::uses_allocator<MutableMove, test_allocator<int>> : cuda::std::true_type
+{};
+#endif
```
miscco:
```suggestion
#endif // !TEST_COMPILER(NVRTC) applies througout
```

libcudacxx/test/support/copy_move_types.h
```diff
@@ -0,0 +1,489 @@
...
+  {}
+  constexpr MutableMove(MutableMove&&)                 = default;
+  TEST_FUNC constexpr MutableMove(const MutableMove&&) = delete;
+
+#if !TEST_COMPILER(NVRTC)
+  TEST_FUNC constexpr MutableMove(cuda::std::allocator_arg_t, const test_allocator<int>&, MutableMove&& o)
+      : val(o.val)
+      , alloc_constructed(true)
+  {}
+#endif
```
miscco: What is the issue here? Please add some explanation

Jacobfaib: NVRTC doesn't like the entire contents of `test_allocator.h`, which also does not appear to have been `cuda::std::`-ified either, it also includes all the normal std library headers instead of ours. I guess it was copied over from LLVM but never used anywhere. Instead of fixing `test_allocator.h` and opening that whole can of worms in this PR I just ifdefed them out.

Actually on thinking on it, I might just remove these bits from here instead of ifdefing them out, the `allocator_arg_t` versions aren't used anywhere (neither in the test nor in our API), because I didn't add the `allocator_arg_t` aware constructors. What do you think?

miscco:
Please combine this with #8885 to fully implement [P2165](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p2165r4.pdf)

libcudacxx/test/support/copy_move_types.h
miscco: We should test them if we can, more test coverage is always better

Jacobfaib:
> Please combine this with https://github.com/NVIDIA/cccl/pull/8885 to fully implement [P2165](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p2165r4.pdf)

Discussed offline, but this PR (with or without 8885) does not fully implement P2165, only the subset required for `ranges::zip_view` (https://github.com/NVIDIA/cccl/pull/8744). Notably missing are:

1. The `tuple` <=> `pair` equivalence.
2. The `allocator_arg_t` constructors for foreign tuples.
4. `operator==()` for foreign tuples.
3. The _tuple-like_ concept.
5. `tuple_cat`, `apply()` and `make_from_tuple()` for _tuple-like_.
6. The updated `basic_common_reference` and `common_type` specializations for _tuple-like_ objects.

libcudacxx/test/support/copy_move_types.h
Jacobfaib: There is nothing to test. The `allocator_arg_t` constructors for foreign tuples are not implemented in the PR. I would like to leave this as a follow-up that implements P2165 in its entirety, as the constructors may need some additional work for foreign tuples and tuple-like objects.

PR #8887, opened 2026-05-08
Use `_CCCL_KERNEL_ATTRIBUTES` in `launch.h`

KyleFromNVIDIA:
## Description

Replace `__global__` with `_CCCL_KERNEL_ATTRIBUTES` in `libcudacxx/include/cuda/__launch/launch.h`.

Follow-up to https://github.com/NVIDIA/cccl/pull/8882.

## Checklist

### DIFF
--- a/libcudacxx/include/cuda/__launch/launch.h
+++ b/libcudacxx/include/cuda/__launch/launch.h
@@ -290,9 +290,12 @@ inline constexpr bool __invoke_kernel_functor_with_config_v =
 // 2. With __launch_bounds__ for non-cluster launches with compile-time known block size.
 // 3. Fallback without any attributes.
 
+// GCC falsely throws the following error without the below suppression:
+//
+// libcudacxx/include/cuda/__launch/launch.h:311:119: error: '__visibility__' attribute ignored [-Werror=attributes]
+
 _CCCL_DIAG_PUSH
 _CCCL_DIAG_SUPPRESS_GCC("-Wattributes")
-_CCCL_DIAG_SUPPRESS_CLANG("-Wattributes")
 
 template <class _Config, class _Kernel, class... _Args>
 _CCCL_KERNEL_ATTRIBUTES void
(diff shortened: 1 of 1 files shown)


libcudacxx/include/cuda/__launch/launch.h
```diff
@@ -290,8 +290,12 @@ inline constexpr bool __invoke_kernel_functor_with_config_v =
 // 2. With __launch_bounds__ for non-cluster launches with compile-time known block size.
 // 3. Fallback without any attributes.
 
+_CCCL_DIAG_PUSH
+_CCCL_DIAG_SUPPRESS_GCC("-Wattributes")
+_CCCL_DIAG_SUPPRESS_CLANG("-Wattributes")
```
Jacobfaib: Can you add a comment in the source that includes the reported warning? Something like
```c++
// GCC and clang warn with:
//
// /truncated/path/to/file: error: bla bla bla
// ... foo()
//     ^~~ No bar here
```
Show the full relevant error (including any source snippets printed by the compiler), but truncate paths and big template messes where needed.

So that future readers aren't wondering what the problem was.

Jacobfaib (approved):

libcudacxx/include/cuda/__launch/launch.h
bdice: The issue is scoped to some nvcc versions from 12.0 to some boundary before 12.9 (12.0 bad, 12.9 good). I'm figuring out the affected version range now. I also don't think clang is affected, possibly only GCC. Then we can tighten the scope of this even further.

KyleFromNVIDIA:
We're still running into cudafe++ problems:

https://github.com/NVIDIA/cccl/actions/runs/25579889959/job/75096300437?pr=8887#step:4:2830

Nothing I have done has been able to resolve this error. I am at a loss. Closing.

libcudacxx/include/cuda/__launch/launch.h
bdice: This is closed and not planning to revisit at this time -- but for historical record, the fix appears to be in CUDA 12.6. Earlier than 12.6 fails in this way, and 12.6+ appear to be working.

PR #8888, opened 2026-05-09
[infra] Add `clang-22` CUDA build job to nightly/weekly CI

davebayer:
LLVM 22.1 is out, so I thought we could add it to our nightly/weekly CI runs as `clang-cuda-next` run to catch potential issues early. It brings CUDA 12.9 support.

Waiting for llvm22 + cuda-ext devcontainers: https://github.com/rapidsai/devcontainers/pull/695


bernhardmgruber (approved):

alliepiper (approved):

libcudacxx/include/cuda/std/span
```diff
@@ -613,10 +613,10 @@ _CCCL_DEDUCTION_GUIDE_ATTRIBUTES span(const array<_Tp, _Sz>&) -> span<const _Tp,
 
 #if _CCCL_HAS_HOST_STD_LIB()
 template <class _Tp, size_t _Sz>
-_CCCL_HOST_DEVICE span(::std::array<_Tp, _Sz>&) -> span<_Tp, _Sz>;
+_CCCL_DEDUCTION_GUIDE_ATTRIBUTES span(::std::array<_Tp, _Sz>&) -> span<_Tp, _Sz>;
```
Jacobfaib, line 621: There are a lot of other places where this is just `_CCCL_API` (for example the ranges). Is that a problem?

libcudacxx/test/libcudacxx/std/utilities/memory/smartptr/unique.ptr/unique.ptr.create/make_unique_for_overwrite.pass.cpp
```diff
@@ -24,13 +24,21 @@
...
+#include <cuda/std/type_traits>
 #include <cuda/std/utility>
 
 #include "test_macros.h"
 
+template <class T, class Void, class... Args>
+inline constexpr bool HasMakeUniqueForOverwriteImpl = false;
 template <class T, class... Args>
-_CCCL_CONCEPT HasMakeUniqueForOverwrite = _CCCL_REQUIRES_EXPR((T, variadic Args), T t, Args&&... args)(
-  (cuda::std::make_unique_for_overwrite<T>(cuda::std::forward<Args>(args)...)));
```
Jacobfaib, line 33: What was the issue with the original concept macro formulation?

libcudacxx/include/cuda/std/span
davebayer: I've changed all deduction guide attributes in #6964, so none should use `_CCCL_HOST_DEVICE` anymore :)

libcudacxx/test/libcudacxx/std/utilities/memory/smartptr/unique.ptr/unique.ptr.create/make_unique_for_overwrite.pass.cpp
davebayer: `clang-22` in CUDA mode failed to evaluate it correctly in C++17 for some reason

ci/matrix.yaml
```diff
@@ -231,7 +231,9 @@ workflows:
     - {jobs: ['build'], cxx: 'nvhpc',      ctk: 'nvhpc',      std: 'all', project: ['libcudacxx', 'cub', 'thrust', 'cudax', 'stdpar'], cpu: ['amd64', 'arm64']}
     # clang-cuda
     - {jobs: ['build'], cudacxx: 'clang', ctk: 'clang-cuda', cxx: 'clang-cuda', std: 'all', sm: '75;80;90;100;120'}
+    - {jobs: ['build'], cudacxx: 'clang', ctk: 'clang_preview-cuda', cxx: 'clang_preview-cuda', std: 'all', sm: '75;80;90;100;120'}
```
miscco, line 241: Nitpick: I would prefer if we would not use `_` here but rather `clang-cuda-preview`

miscco (approved):

PR #8890, opened 2026-05-09
[libcu++] Element access methods/operators for host accessible buffers

### LINKED ISSUE #8889: [FEA]: Element access methods/operators for host accessible buffers
When rewriting some CUDA examples using our `cuda::buffer`, I've found out that it's painful to use this type without `.at`, `operator[]`, `front` and `back` methods/operators. I'd like to be able to use them directly on host side when the buffer is host accessible.

davebayer:
Fixes #8889.

### DIFF
--- a/libcudacxx/include/cuda/__container/buffer.h
+++ b/libcudacxx/include/cuda/__container/buffer.h
@@ -477,6 +477,92 @@ public:
     return __buf_.data();
   }
 
+  //! @brief Gets a reference to the element at index `__i`.
+  //! @throw std::out_of_range if `__i < size()`.
+  _CCCL_TEMPLATE(bool _IsHostAccessible = ::cuda::mr::__is_host_accessible<_Properties...>)
+  _CCCL_REQUIRES(_IsHostAccessible)
+  [[nodiscard]] _CCCL_HOST_API _Tp& at(size_type __i)
+  {
+    if (__i >= size())
+    {
+      _CCCL_THROW(::std::out_of_range, "__i must be less than size()");
+    }
+    return data()[__i];
+  }
+
+  //! @brief Gets a reference to the element at index `__i`.
+  //! @throw std::out_of_range if `__i < size()`.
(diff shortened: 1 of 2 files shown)


libcudacxx/include/cuda/__container/buffer.h
```diff
@@ -477,6 +477,92 @@ class buffer
     return __buf_.data();
   }
 
+  //! @brief Gets a reference to the element at index `__i`.
+  //! @throw std::out_of_range if `__i < size()`.
```
Jacobfaib, line 481: Question: should we be adding `@since <version>` to any newly added API? Or is there some changelog document we should be adding to that tracks these additions?

libcudacxx/include/cuda/__container/buffer.h
```diff
@@ -477,6 +477,92 @@ class buffer
...
+  _CCCL_TEMPLATE(bool _IsHostAccessible = ::cuda::mr::__is_host_accessible<_Properties...>)
+  _CCCL_REQUIRES(_IsHostAccessible)
```
Jacobfaib, line 484: Does doxygen properly translate the requires macros into a readable form? Or should documentation also spell out that this is only accessible if the underlying resource is host accessible?

Jacobfaib (approved):

libcudacxx/include/cuda/__container/buffer.h
bernhardmgruber: > Question: should we be adding `@since <version>` to any newly added API?

Yes please.

bernhardmgruber (approved):

miscco (approved):
With the `@versionadded`

libcudacxx/include/cuda/__container/buffer.h
davebayer: I don't know, I've updated the docs to explicitly say that the functionality is available only when the buffer is host accessible

PR #8891, opened 2026-05-10
STF: extend cuda_try output-parameter inference (first/last + ambiguity rejection)

andralex:
## Summary

Today, calls to single-output CUDA APIs through STF's `cuda_try` look like this:

```cpp
unsigned long long id;
cuda_try(cuStreamGetId(stream, &id));

cudaGraphNode_t node;
cuda_try(cudaGraphAddEmptyNode(&node, graph, deps.data(), deps.size()));

int can_access;
cuda_try(cudaDeviceCanAccessPeer(&can_access, dev, peer));
```

Three lines of bookkeeping per call, with a temporary that has to be named, default-initialized correctly, and then thrown away after one use. The pattern is mechanical, error-prone (e.g. uninitialized variable on the failure path), and obscures what the code is actually doing.

This PR generalizes the templated `cuda_try<fun>(args...)` overload so that all three of the above become one-liners:

```cpp
auto id          = cuda_try<cuStreamGetId>(stream);
auto node        = cuda_try<cudaGraphAddEmptyNode>(graph, deps.data(), deps.size());
auto can_access  = cuda_try<cudaDeviceCanAccessPeer>(dev, peer);
```

The helper picks the right call shape at compile time based on the function's signature, with no runtime cost and no loss of error reporting (the same `cuda_exception` is thrown on failure, with the same source location).

## What changed

`cuda_try<fun>(args...)` now selects, in order, from three forms:

1. **Direct.** If `fun(args...)` is invocable, call it and check the status. Returns `void`.
2. **First-parameter output.** Otherwise, if `fun`'s first parameter is a non-`const` pointer (the CUDA convention for output parameters) and `fun(&result, args...)` is invocable, materialize `result`, call, and return it. Matches `cudaStreamCreate`, `cudaGraphAddEmptyNode`, `cudaDeviceCanAccessPeer`, ...
3. **Last-parameter output.** Otherwise, if `fun`'s last parameter is a non-`const` pointer and `fun(args..., &result)` is invocable, materialize `result`, call, and return it. Matches the driver-API "getter" convention: `cuStreamGetId`, `cuCtxGetId`, `cuStreamGetCtx`, ...

Form (3) is the new piece - previously only forms (1) and (2) were supported, which is why driver-API getters had to fall back to the verbose pattern.

### Compile-time ambiguity rejection

Some CUDA functions have non-`const` pointer parameters in *both* the first and the last position (e.g. `cudaMemGetInfo(size_t* free, size_t* total)`). For those, supplying a single `size_t*` is consistent with either interpretation, with different effects. Rather than silently picking one, `cuda_try` now refuses to compile such calls:

```
error: static assertion failed: Ambiguous cuda_try: both first- and last-output forms apply;
       call the function explicitly to disambiguate.
```

The user is then expected to fall back to the explicit `cuda_try(cudaMemGetInfo(&free, &total))` form, which is unambiguous.

The zero-argument case (`cuda_try<fun>()`) is exempt from the assertion because the synthesized call `fun(&result)` is identical for both interpretations.

A negative compile-time test (`cudax/test/stf/static_error_checks/cuda_try_ambiguous.cu`) exercises the assertion using `cudaMemGetInfo`, and is wired into the existing static-error test infrastructure.

### Documentation

The doc comment on `cuda_try<fun>` now spells out:

- the three call shapes and their selection order,
- the ambiguity rule and its zero-argument exception,
- known limitations (overloaded names like `cudaMalloc` are unsupported and must use the runtime-status overload),
- worked examples for each form.

### Call-site migrations

A handful of safe, single-output call sites in `__stf` and `__places` are migrated to the new shorthand to demonstrate the ergonomics and exercise both new forms in CI:

- `cudax/include/cuda/experimental/__places/exec/green_context.cuh` -- `cuCtxGetId`
- `cudax/include/cuda/experimental/__places/localized_array.cuh` -- `cudaDeviceCanAccessPeer`
- `cudax/include/cuda/experimental/__places/stream_pool.cuh` -- `cudaStreamGetDevice`, `cuStreamGetCtx`, `cuStreamGetId`
- `cudax/include/cuda/experimental/__stf/graph/graph_ctx.cuh` -- `cudaGraphAddEmptyNode`

Sites involving overloaded CUDA functions (`cudaMalloc` family) or in/out parameters were intentionally left alone.

## What is *not* changed

- The runtime-status overload `cuda_try(cudaSomething(...))` is unchanged and remains the right tool whenever the new shorthand cannot be used (overloaded functions, multiple outputs, ambiguous signatures).
- No public ABI/API changes outside the `cuda_try` template.
- No behavior change on the failure path: same `cuda_exception`, same source location.

## Test plan

- [x] Builds locally with `cudax-cpp17` preset, `CMAKE_CUDA_ARCHITECTURES=86`.
- [x] Existing `cuda_try` unit tests (`cuda_try1`, `cuda_try2`) pass.
- [x] New negative compile-time test `cuda_try_ambiguous.cu` fails to compile with the expected `static_assert` message and is wired into the static-error CMake test list.
- [ ] Full STF + cudax CI on this PR.


caugonnet:
I guess this belongs to CCCL utilities at some point, but it's nice !

caugonnet (approved):

alliepiper (approved):

PR #8892, opened 2026-05-11
Variable segment size TopK benchmark

gevtushenko:
## Description

closes https://github.com/NVIDIA/cccl_private/issues/668

This PR adds a benchmark to variable segment size version of TopK. A few things of note:

- not all important workloads are supported today, so parts of `max_segment_size_list` are ifdef-ed out
- if patterns turn out to be usable outside of TopK context, we'll move them into nvbench helper

Data distribution based on the pattern:

### random

Screenshot 2026-05-10 at 7 01 20 PM

### quantized random

Screenshot 2026-05-10 at 7 02 17 PM

### relu quantized

Screenshot 2026-05-10 at 7 02 49 PM

### tie heavy

Screenshot 2026-05-10 at 7 03 25 PM

### pivot tie

Screenshot 2026-05-10 at 7 04 18 PM

## Checklist


cub/benchmarks/bench/segmented_topk/variable/keys.cu
```diff
@@ -0,0 +1,259 @@
...
+      //  |
+      //  |
+      //  |
+      //  |                         |
+      //  ----------------------------
+      //  1.0                       2.0
+      case pattern_kind::pivot_tie: {
+        const auto seg_size = static_cast<int>(d_seg_sizes[idx / MaxSegmentSize]);
+        return (j >= seg_size - gt_count) ? 2.f : 1.f;
+      }
```
NaderAlAwar, line 143: Question: Is this intentionally placing the 2.0 tail at the end of each live segment rather than at the end of the full `MaxSegmentSize` row? It differs from the issue’s Python `pivot_tie`, where shorter segments might not see the 2.0 values. This looks like the right behavior but I wanted to check the discrepancy

NaderAlAwar (approved):

cub/benchmarks/bench/segmented_topk/variable/keys.cu
```diff
@@ -0,0 +1,259 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include <cub/detail/choose_offset.cuh>
+#include <cub/device/dispatch/dispatch_batched_topk.cuh>
+
+#include <thrust/device_vector.h>
+#include <thrust/tabulate.h>
```
oleksandr-pavlyk, line 9:
```suggestion
#include <thrust/device_vector.h>
#include <thrust/reduce.h>
#include <thrust/tabulate.h>
```

Add include to use `thrust::reduce` without relying on transitive include.

cub/benchmarks/bench/segmented_topk/variable/keys.cu
```diff
@@ -0,0 +1,259 @@
...
+  auto d_keys         = thrust::device_vector<float>{num_keys, thrust::no_init};
+
+  // gt_count == "greater-than count": number of 2.0 values placed at the tail of each segment's live region.
+  constexpr int gt_count = cuda::std::max(1, cuda::std::min(K / 4, MaxSegmentSize / 8));
+
+  thrust::tabulate(d_keys.begin(), d_keys.end(), [pattern, d_seg_sizes] __device__(std::size_t idx) -> float {
+    auto quantize = [](float base) -> float {
+      const auto scaled = static_cast<double>(base) * 32.0;
+      const auto r      = cuda::std::rint(scaled);
+      return static_cast<float>(r / 32.0);
```
oleksandr-pavlyk: We could avoid using double precision altogether by subtracting integer part from the base, and only scaling the difference.

```suggestion
      const auto r      = cuda::std::rint(base);
      const auto scaled_fr = std::rint((base - r) * 32.0f);
      return r + (scaled_fr / 32.0f);
```

cub/benchmarks/bench/segmented_topk/variable/keys.cu
```diff
@@ -0,0 +1,259 @@
...
+    };
+
+    auto random_value = [](unsigned long long idx) -> float {
+      cuda::pcg64 rng(42);
+      rng.discard(idx);
+      cuda::std::normal_distribution<float> normal(0.f, 1.f);
+      return normal(rng);
```
oleksandr-pavlyk, line 80: Each thread needs to discard words used by other threads with preceding IDXs. But how many words are consumed to sample from `normal`? If it is two words (likely), perhaps we should do `rng.discard(2*idx);` instead.

The main value of getting it right in this benchmark is education, though.

cub/benchmarks/bench/segmented_topk/variable/keys.cu
```diff
@@ -0,0 +1,259 @@
...
+  16384,
+  32768,
+  65536,
+  131072,
+  262144,
+  524288,
+  1048576
+#endif
+  >;
+
```
oleksandr-pavlyk, line 254:
```suggestion
using max_segment_size_list = nvbench::enum_type_list< //
  512,
  1024,
  2048,
  4096,
  8192
#if 0 // need these, waiting for implementation to catch up
  ,
  16384,
  32768,
  65536,
  131072,
  262144,
  524288,
  1048576
#endif
  >;
```

Clang-tidy complained about this:

```
  Error while processing /home/coder/cccl/cub/benchmarks/bench/segmented_topk/variable/keys.cu.
  /home/coder/cccl/cub/benchmarks/bench/segmented_topk/variable/keys.cu:251:3: error: expected expression [clang-diagnostic-error]
    251 |   >;
        |   ^
  /home/coder/cccl/cub/benchmarks/bench/segmented_topk/variable/keys.cu:255:83: error: use of undeclared identifier 'max_segment_size_list' [clang-diagnostic-error]
    255 | NVBENCH_BENCH_TYPES(variable_seg_size_topk_keys, NVBENCH_TYPE_AXES(key_type_list, max_segment_size_list, k_list))
        |                                                                                   ^
```

cub/benchmarks/bench/segmented_topk/variable/keys.cu
gevtushenko: Thank you for checking on this! Yes, that's intentional. Original Python code is from fixed-segment-size code, so it was a bug that I fixed in C++ version only.

cub/benchmarks/bench/segmented_topk/variable/keys.cu
gevtushenko: Good catch, thanks!

PR #8894, opened 2026-05-11
[cudax] Implement `binary_partition` mapping for threads within warp

davebayer:
This PR implements `binary_partition` mapping that splits each group into 2 parts based on whether a predicate function returns `true` or `false`. It is currently implemented only for threads within warp.

From what I've seen, cooperative groups use `__ballot_sync` for voting, however I used newer and more flexible `__match_any_sync`. The generated SASS is almost identical, but I'm not sure whether there isn't some overhead introduced when using the `MATCH.ANY` instruction.

### DIFF
--- a/cudax/include/cuda/experimental/__group/mapping/binary_partition.cuh
+++ b/cudax/include/cuda/experimental/__group/mapping/binary_partition.cuh
@@ -1,0 +1,102 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of CUDA Experimental in CUDA C++ Core Libraries,
+// under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef _CUDA_EXPERIMENTAL___GROUP_MAPPING_BINARY_PARTITION_CUH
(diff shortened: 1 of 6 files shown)


miscco:
I believe @gonzalobg would know

cudax/include/cuda/experimental/__group/mapping/binary_partition.cuh
```diff
@@ -0,0 +1,133 @@
...
+    constexpr auto __static_ngroups =
+      (__static_prev_ngroups != ::cuda::std::dynamic_extent)
+        ? (__static_prev_ngroups * 2)
+        : ::cuda::std::dynamic_extent;
+
+    using _MappingResult =
+      __mapping_result<__static_ngroups, ::cuda::std::dynamic_extent, _PrevMappingResult::is_always_exhaustive(), false>;
+
+    bool __pred;
+    unsigned __group_rank = ~0u;
```
miscco: Can we use explicitly sized integers?

cudax/include/cuda/experimental/__group/mapping/binary_partition.cuh
```diff
@@ -0,0 +1,133 @@
...
+    constexpr auto __static_prev_ngroups = _PrevMappingResult::static_group_count();
+    constexpr auto __static_ngroups =
+      (__static_prev_ngroups != ::cuda::std::dynamic_extent)
+        ? (__static_prev_ngroups * 2)
+        : ::cuda::std::dynamic_extent;
+
+    using _MappingResult =
+      __mapping_result<__static_ngroups, ::cuda::std::dynamic_extent, _PrevMappingResult::is_always_exhaustive(), false>;
+
+    bool __pred;
```
miscco: We should have a default

```suggestion
    bool __pred = false; 
```

miscco (approved):

PR #8897, opened 2026-05-11
ArgMin/Max fixes and improvements

bernhardmgruber:
There were a couple of oversights in #8285, which are fixed by this PR:

* arguments to `dispatch_streaming_arg_reduce` are passed in the wrong order in the `arg_extrema.cu` benchmark
* `arg_less` wasn't a good name since it does not return bool. `arg_min` would be right, but that name is taken. So it's `arg_reduce_op`.
* Call `__arg_min` directly in some device level APIs to avoid one additional function call

### DIFF
--- a/cub/cub/device/device_reduce.cuh
+++ b/cub/cub/device/device_reduce.cuh
@@ -881,7 +881,7 @@ private:
       d_min_out,
       d_index_out,
       static_cast<GlobalOffsetT>(num_items),
-      detail::arg_less{compare_op},
+      detail::arg_reduce_op{compare_op},
       stream);
   }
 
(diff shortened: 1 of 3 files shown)


cub/cub/device/device_reduce.cuh
```diff
@@ -1898,7 +1893,7 @@ public:
     EnvT env = {})
   {
     _CCCL_NVTX_RANGE_SCOPE("cub::DeviceReduce::ArgMax");
-    return __arg_min_env(d_in, d_max_out, d_index_out, num_items, detail::arg_less{detail::swap_args{compare_op}}, env);
```
bernhardmgruber, line 1901: Wrapping in `arg_less` here was too early as well.

miscco (approved):

davebayer (approved):

Issue #6371, opened 2025-10-29
[FEA]: Host-side logging system


leofang:
### Is this a duplicate?

- [x] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cccl/issues) for this request and that I agree to the [Code of Conduct](CODE_OF_CONDUCT.md)

### Area

General CCCL

### Is your feature request related to a problem? Please describe.

Today, math libraries such as cuBLAS, cuSolver, cuFFT, and others use a unified logging system (whose implementation is internal-only) that supports controls via both environment variables and APIs (both are documented and public-facing), allowing for in-memory and on-disk logging. It is very useful for debugging, performance tracking, rich error msgs, and algorithm selection transparency.

Host-side logging should not require recompilation to enable or disable, and the logging code should always be present in the binary with aforementioned runtime control. For header-only use cases, we could provide a macro to turn off the logging completely if desired. 

(CCCL) Developer education for logging practices will be needed (ex: when to warn; "debug" vs "error"; ...)

Examples of valuable log include but are not limited to: tuning policy selection, input arguments (pointer addresses, sizes, types, ...), temporary memory allocations, memory resource usage, and stream addresses, which are not easily accessible through existing developer tools like Nsys or NCU.

miscco:
I am torn here.

Logging is amazing and I believe it would add tremendous value to our users and ourself ;)

However, logging is a nontrivial endeavor that carries a *lot* of additional burden when it comes to binary size and potentially runtime performance.

Because of that I would always require the user to explicitly enable any for of logging through a build flag

bernhardmgruber:
> However, logging is a nontrivial endeavor that carries a _lot_ of additional burden when it comes to binary size and potentially runtime performance.

I am actually not so worried about that if we are careful. Adding some logging capabilities would probably add like 1k-10k strings to CCCL (that's a few strings per CUB/Thrust API). Let's assume those are 100 chars long. That's up to about 1MiB. This feels still small in comparison to all the kernel binaries for various SM architectures.

Regarding runtime cost, we are talking about an (atomic) load of a flag (the log level) and a branch. Maybe a bit more since we would need to get the env variable on the first time we ask for the log level. That should be very cheap on the same code path that schedules a kernel, potentially does an allocation, and schedules an NVTX range (enabled by default).

I think it would be reasonable to pay that, even in a default build. But we should (runtime) turn off logging by default. And we should have an option to compile out the logging code.

[...]

PR #8899, opened 2026-05-11
Always log selected tuning policies

bernhardmgruber:
They were missing in some dispatch functions.

### DIFF
--- a/cub/cub/device/dispatch/dispatch_reduce_deterministic.cuh
+++ b/cub/cub/device/dispatch/dispatch_reduce_deterministic.cuh
@@ -32,6 +32,7 @@
 #include <cuda/std/__functional/identity.h>
 #include <cuda/std/__functional/invoke.h>
 #include <cuda/std/__functional/operations.h>
+#include <cuda/std/__host_stdlib/sstream>
 #include <cuda/std/__type_traits/decay.h>
 #include <cuda/std/__type_traits/enable_if.h>
 #include <cuda/std/cstdint>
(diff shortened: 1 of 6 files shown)


cub/cub/device/dispatch/dispatch_reduce_deterministic.cuh
```diff
@@ -346,6 +347,18 @@ CUB_RUNTIME_FUNCTION _CCCL_FORCEINLINE cudaError_t dispatch(
...
+#if _CCCL_HOSTED() && defined(CUB_DEBUG_LOG)
+  NV_IF_TARGET(NV_IS_HOST, ({
+                 std::stringstream ss;
+                 ss << active_policy;
+                 _CubLog("Dispatching DeviceReduceDeterministic to compute capability %d.%d with tuning: %s\n",
+                         cc.major_cap(),
+                         cc.minor_cap(),
+                         ss.str().c_str());
+               }))
+#endif // _CCCL_HOSTED() && defined(CUB_DEBUG_LOG)
```
davebayer, line 360: Suggestion: We should make a function that does the logging for us, we repeat ourselves many times

davebayer (approved):

cub/cub/device/dispatch/dispatch_reduce_deterministic.cuh
leofang: xref https://github.com/NVIDIA/cccl/issues/6371

bernhardmgruber: > xref #6371

I am aware and I want this soonish, because I want people to be able to easily print the selected tunings.

bernhardmgruber: > Suggestion: We should make a function that does the logging for us, we repeat ourselves many times

Yeah, but I would like to defer this. Possibly until #6371 is resolved. Here is a tracking issue for now: #8900

PR #8914, opened 2026-05-12
[STF] Fix graph capture integration

caugonnet:
Rework graph_task<>'s stream-capture path to use
cudaStreamBeginCaptureToGraph on CTK 12.3+, replacing the previous
"capture into a fresh per-task graph and embed it as a child-graph
node" scheme. The legacy scheme is kept verbatim for CTK 12.0-12.2.
Motivation: conditional graph nodes created inside a task body (e.g.
Warp's wp.capture_while, available from CTK 12.4) cannot be embedded
through a child-graph node -- CUDA refuses to splice the captured
graph back into ctx_graph. Capturing straight into ctx_graph lets
those nodes land at the top level where they belong.

Implementation notes:
* Two new private helpers encapsulate the new path:
  - begin_capture_into_ctx_graph() wraps
    cudaStreamBeginCaptureToGraph in relaxed mode.
  - end_capture_into_ctx_graph_and_emit_done() snapshots the capture
    frontier *before* cudaStreamEndCapture (the deps_out pointer is
    invalidated by any subsequent capture call), ends the capture,
    and always adds exactly one empty "done" node into ctx_graph
    that depends on that frontier. Emitting a single done node
    unconditionally preserves the done_nodes.size() == captures
    invariant and avoids fragile empty/single-tail detection
    (cudaStreamGetCaptureInfo can report inherited input deps as
    the frontier when no work was actually captured).
* ctx_graph is mutated for the full capture interval
  (BeginCaptureToGraph .. user work .. GetCaptureInfo .. EndCapture
  .. AddEmptyNode), which spans start()/end_uncleared(). graph_mutex
  is now held across that boundary by moving a unique_lock from
  start() into a new capture_lock_ member and back into a local in
  end_uncleared(), so RAII releases the lock even if user code
  throws. As a consequence graph_task<> becomes move-only.
* lock_ctx_graph() now returns an explicit
  std::unique_lock<std::mutex> instead of auto, so the existing
  [[nodiscard]] actually has effect.

Also fix an unrelated event leak in
stackable_ctx_impl::pop_after_finalize(): when a graph_scope is used
purely for token ordering (pushed_data is empty), the loop that
chains finalize_prereqs through pushed_data left the parent context
with nothing to wait on, so the parent could synchronize its own
(empty) submitted_stream while the body graph was still running on
support_stream. Forward finalize_prereqs as dangling events on the
parent backend (when it tracks them) so insert_fence picks them up
on the next task or finalize().

closes 

## Checklist


caugonnet:
@coderabbitai full review

cudax/include/cuda/experimental/__stf/graph/graph_task.cuh
```diff
@@ -81,7 +81,7 @@ public:
 
   graph_task& start()
   {
-    ::std::lock_guard<::std::mutex> lock(graph_mutex);
+    auto lock = lock_ctx_graph();
```
andralex, line 84: noice

cudax/include/cuda/experimental/__stf/graph/graph_task.cuh
```diff
@@ -104,9 +104,23 @@ public:
...
       capture_stream = get_exec_place().getStream(true).stream;
-      // Use relaxed capture mode to allow capturing workloads that lazily initialize
-      // resources (e.g., set up memory pools)
+#if _CCCL_CTK_AT_LEAST(12, 3)
+      // New path: capture directly into ctx_graph via cudaStreamBeginCaptureToGraph.
+      // ctx_graph is mutated for the full capture interval, so graph_mutex must
+      // stay held across the start()/end_uncleared() boundary; transfer ownership
+      // into the task-owned capture_lock_ member here.
+      begin_capture_into_ctx_graph(capture_stream, ctx_graph, ready_dependencies);
+      capture_lock_ = ::std::move(lock);
```
andralex:
```suggestion
      capture_lock_ = mv(lock);
```

cudax/include/cuda/experimental/__stf/graph/graph_task.cuh
```diff
@@ -123,13 +137,38 @@ public:
...
   {
+#if _CCCL_CTK_AT_LEAST(12, 3)
+    // On the capture path, graph_mutex was acquired in start() and lives in
+    // capture_lock_. Move it into a local unique_lock so RAII releases it at
+    // the end of this function. On the non-capture path, acquire a fresh lock
+    // here just like the legacy code did.
+    ::std::unique_lock<::std::mutex> lock;
+    if (is_capture_enabled())
+    {
+      lock = ::std::move(capture_lock_);
```
andralex: Generally it's better to use `mv` whenever within the generous aegis of cudastf (better checking, nice diagnosis in case of misuse).

cudax/include/cuda/experimental/__stf/graph/graph_task.cuh
```diff
@@ -123,13 +137,38 @@ public:
...
+    }
+    else
+    {
+      lock = lock_ctx_graph();
+    }
```
andralex:
```suggestion
    ::std::unique_lock<::std::mutex> lock = is_capture_enabled()
      ? mv(capture_lock_)
      : lock_ctx_graph();
```

cudax/include/cuda/experimental/__stf/graph/graph_task.cuh
```diff
@@ -123,13 +137,38 @@ public:
...
+#else // _CCCL_CTK_AT_LEAST(12, 3)
+      // Legacy path: end capture into a fresh per-task graph and let the
+      // child-graph embed branch below splice it into ctx_graph.
       cudaGraph_t childGraph = nullptr;
       cuda_safe_call(cudaStreamEndCapture(capture_stream, &childGraph));
       set_child_graph(childGraph);
+#endif // _CCCL_CTK_AT_LEAST(12, 3)
     }
 
     cudaGraphNode_t n;
```
andralex: Where is this used?

cudax/include/cuda/experimental/__stf/graph/graph_task.cuh
```diff
@@ -494,6 +547,64 @@ private:
...
+#  else // _CCCL_CTK_AT_LEAST(13, 0)
+    cuda_safe_call(cudaStreamGetCaptureInfo_v2(s, &status, nullptr, nullptr, &deps_out, &ndeps));
+#  endif // _CCCL_CTK_AT_LEAST(13, 0)
+    ::std::vector<cudaGraphNode_t> frontier(deps_out, deps_out + ndeps);
+
+    // Ignore the returned graph handle: with BeginCaptureToGraph it is the
+    // caller-supplied ctx_graph; do not make correctness depend on identity.
+    cudaGraph_t captured = nullptr;
+    cuda_safe_call(cudaStreamEndCapture(s, &captured));
+    (void) captured;
```
andralex: Spaceship-era version: `::std::ignore = captured;`. Also `[[maybe_unused]]` on the declaration. Not sure which I like more.
Ironically in this case it's not needed at all...

cudax/include/cuda/experimental/__stf/graph/graph_task.cuh
```diff
@@ -494,6 +547,64 @@ private:
...
+
+    cudaGraphNode_t done_node = nullptr;
+#  if _CCCL_CTK_AT_LEAST(13, 0)
+    cudaGraphNodeParams params = {};
+    params.type                = cudaGraphNodeTypeEmpty;
+    cuda_safe_call(cudaGraphAddNode(&done_node, ctx_graph, frontier.data(), nullptr, frontier.size(), &params));
+#  else // _CCCL_CTK_AT_LEAST(13, 0)
+    cuda_safe_call(cudaGraphAddEmptyNode(&done_node, ctx_graph, frontier.data(), frontier.size()));
+#  endif // _CCCL_CTK_AT_LEAST(13, 0)
```
andralex, line 593: I guess no need to use different APIs here - just use `cudaGraphAddEmptyNode` everywhere.


```suggestion
  cuda_safe_call(cudaGraphAddEmptyNode(&done_node, ctx_graph, frontier.data(), frontier.size()));
```

cudax/include/cuda/experimental/__stf/graph/graph_task.cuh
caugonnet: indeed i've moved it where it's used

cudax/include/cuda/experimental/__stf/graph/graph_task.cuh
caugonnet: The polymorphic API is the newest version, and was introduced in CUDA 12.2, and in CUDA 13 the number of parameters also changed so it's ok-ish to have this, even if the right future-proof API is the one in cudaGraphAddNode

cudax/include/cuda/experimental/__stf/graph/graph_task.cuh
caugonnet: [[maybe_unused]] is more consistent here

cudax/include/cuda/experimental/__stf/stackable/stackable_ctx_impl.cuh
```diff
@@ -839,6 +839,24 @@ public:
...
+      // The parent's eventual fence/finalize then synchronizes its own
+      // (essentially empty) submitted_stream while the body graph is
+      // still running. Registering as a dangling event funnels into
+      // `insert_fence`, which any subsequent task or `finalize()` will
+      // pick up.
+      auto& parent_backend = parent_ctx.get_backend();
+      if (finalize_prereqs.size() > 0 && parent_backend.track_dangling_events())
+      {
+        parent_backend.get_state().add_dangling_events(parent_backend, finalize_prereqs);
+      }
```
andralex:
```suggestion
      if (!finalize_prereqs.empty())
      {
        auto& parent_backend = parent_ctx.get_backend();
        if (parent_backend.track_dangling_events())
        {
          parent_backend.get_state().add_dangling_events(parent_backend, finalize_prereqs);
        }
      }
```

andralex: or your favorite:


```suggestion
      if (!finalize_prereqs.empty())
      {
        if (auto& parent_backend = parent_ctx.get_backend(); parent_backend.track_dangling_events())
        {
          parent_backend.get_state().add_dangling_events(parent_backend, finalize_prereqs);
        }
      }
```

andralex (approved):
lgtm, see a few nits

PR #8916, opened 2026-05-12
[libcu++] Fix return type for `count` and `rank` hierarchy queries

### LINKED ISSUE #8637: [BUG] Hierarchy `count` and `rank` queries always return `size_t`
They should return `typename _Level::__product_type`.

davebayer:
Fixes #8637.


miscco (approved):

libcudacxx/include/cuda/__hierarchy/hierarchy_levels.h
```diff
@@ -24,15 +24,64 @@
...
+  using __product_type  = ::cuda::std::uint32_t;
+  using __allowed_above = __allowed_levels<grid_level, cluster_level>;
+  using __allowed_below = __allowed_levels<thread_level>;
+
+  using __next_native_level = cluster_level;
+};
+
+struct _CCCL_DECLSPEC_EMPTY_BASES cluster_level : __native_hierarchy_level_base<cluster_level>
+{
+  using __product_type  = ::cuda::std::uint32_t;
```
miscco, line 65: We might want to consider moving that to
```suggestion
  using __product_type  = ::cuda::std::uint64_t;
```

davebayer: That's not necessary, because the number of threads within a cluster can never overflow `uint32_t`

miscco: I mean worst case we can change it

PR #8919, opened 2026-05-12
[STF] Properly destroy CUDA streams and do not try to initialize CUDA while capturing

caugonnet:
Destroy pool-owned streams with the stream pool and initialize the CUDA runtime only once so consecutive stream_ctx instances on a caller stream serialize without explicit synchronization.

## Description

closes 

## Checklist


cudax/include/cuda/experimental/__stf/internal/backend_ctx.cuh
```diff
@@ -43,6 +43,7 @@
 
 #include <atomic>
 #include <fstream>
+#include <mutex>
```
caugonnet, line 46: remove ?

caugonnet: actually not, it's used

cudax/include/cuda/experimental/__places/stream_pool.cuh
```diff
@@ -163,11 +163,35 @@ class stream_pool
...
+      if (externally_owned)
+      {
+        return;
+      }
+
+      for (auto& ds : payload)
+      {
+        if (ds.stream != nullptr)
+        {
+          [[maybe_unused]] cudaError_t err = cudaStreamDestroy(ds.stream);
```
andralex:
```suggestion
          std::ignore = cudaStreamDestroy(ds.stream);
```

or perhaps:


```suggestion
          cudaError_t err = cudaStreamDestroy(ds.stream);
          if (err)
          {
            fprintf(stderr, "CUDA error while destroying streams: %s (%s) at %s:%d\n", cudaGetErrorName(err), cudaGetErrorString(err), __FILE__, __LINE__);
          }
```

cudax/test/stf/local_stf/stream_ctx_lifetime_btb.cu
```diff
@@ -0,0 +1,205 @@
...
+constexpr long long BUSY_CYCLES = 5'000'000;
+
+__global__ void slow_set_kernel(int* slice, int n, int value, long long ns)
+{
+  const int tid = blockIdx.x * blockDim.x + threadIdx.x;
+  if (tid >= n)
+  {
+    return;
+  }
+  long long start = clock64();
```
andralex:
```suggestion
  const long long start = clock64();
```

cudax/test/stf/local_stf/stream_ctx_lifetime_btb.cu
```diff
@@ -0,0 +1,205 @@
...
+  while (clock64() - start < ns)
+  {
+    // busy wait
+  }
```
andralex:
```suggestion
  unsigned remaining = ns;
  while (remaining > 0) {
    const unsigned chunk = remaining > 1000000u ? 1000000u : remaining;
    __nanosleep(chunk);
    remaining -= chunk;
  }
```

cudax/test/stf/local_stf/stream_ctx_lifetime_btb.cu
```diff
@@ -0,0 +1,205 @@
...
+bool has_mismatch(const std::vector<int>& values, int expected)
+{
+  for (int value : values)
+  {
+    if (value != expected)
+    {
+      return true;
+    }
+  }
+  return false;
```
andralex:
```suggestion
  return ::std::any_of(values.begin(), values.end(), [=](int x) { return x != expected; });
```

andralex (approved):
lgtm % nits

cudax/include/cuda/experimental/__places/stream_pool.cuh
caugonnet: CUDA teardown can happen in weird time, and we may in legit situations where we try to destroy streams while CUDA has been deinit. So i'm not applying. i'm simply going to cast to void, and add a noexcept

cudax/test/stf/local_stf/stream_ctx_lifetime_btb.cu
caugonnet: clock64() reads the GPU special register clock. The compiler treats that as an operation with observable changing state, not as a pure constant expression. Since the loop condition re-reads clock64() every iteration, it cannot legally collapse the loop to nothing.

But i'm renaming to cycles instead of ns

cudax/include/cuda/experimental/__places/stream_pool.cuh
caugonnet: we actually ignore the result because it can "legitimately" fail if we are during CUDA teardown

alliepiper (approved):

caugonnet:
Need to be investigated
```
  The following tests FAILED:
  	208 - cudax.test.stf.threads.axpy-threads-graph (Subprocess aborted)

```

Issue #8528, opened 2026-04-20
[BUG] CUB DeviceScanKernel out-of-bounds global write in torch.isin with BFloat16 on sm120


bernhardmgruber:
Using an internal devcontainer for pytorch, some people are seeing an error with `cub::DeviceScan` running the test:
```
compute-sanitizer pytest -v test/test_meta.py -k test_dispatch_meta_outplace_isin_cuda_bfloat16
...
========= Invalid __global__ write of size 32 bytes
=========     at void at_cuda_detail::cub::detail::scan::DeviceScanKernel<...policy_selector_from_types<int, long, int, unsigned int, cuda::std::__4::plus<void>>, const int *, long *, at_cuda_detail::cub::ScanTileState<int, (bool)1>, cuda::std::__4::plus<void>, at_cuda_detail::cub::NullType, unsigned int, int, (bool)0, at_cuda_detail::cub::NullType>(...)+0x6a50
=========     by thread (151,0,0) in block (0,0,0)
=========     Access to 0x15482f6402a00 is out of bounds
=========     and is 249597593119744 bytes after the nearest allocation at 0x7180fe603e00 of size 1 bytes
```
Host backtrace (simplified):
```
cudaLaunchKernelExC
  → cub::detail::scan::dispatch (InclusiveScan, policy: {128 threads, 24 items, BlockLoad=4, BlockStore=4, BlockScan=2, warpspeed={true, 4, 4, 63}})
  → cub::DeviceScan::InclusiveScan<int const*, long*, plus<void>, int>
  → at::cuda::cub::inclusive_scan<int const*, long*, plus<void>, 1073741824>
  → at::native::internal::compute_unique<BFloat16>
  → at::native::_unique_cuda
  → torch.isin (isin_sorting)
```

NVBug: 6088993

bernhardmgruber:
Reproducer for ToT nvcc (`cuda_dev.bugfix_main.37795374_0`) using CCCL 19ca1f31ee4e9c7bfe2bcd2aacd73eb51f40ab2c. Does not reproduce on nvcc 13.2.
```c++
// Build: 
//   nvcc -Icub -Ithrust -Ilibcudacxx/include -arch=sm_100 repro_scan.cu -o repro_scan 
// Run with sanitizer: 
//   compute-sanitizer ./repro_scan 
 
#include <cub/device/device_scan.cuh> 
#include <cuda_runtime.h> 
 
#include <cuda/std/functional> 
 
#include <cassert> 
[...]

PR #8922, opened 2026-05-12
Disable warpspeed scan on sm120 with nvcc < 13.4

bernhardmgruber:
There seems to be a codegen bug in nvcc when targeting GB20x GPUs (sm120), so let's disable warpspeed scan until this is resolved. See also #8528.

### DIFF
--- a/cub/cub/device/dispatch/tuning/tuning_scan.cuh
+++ b/cub/cub/device/dispatch/tuning/tuning_scan.cuh
@@ -981,7 +981,8 @@ struct policy_selector
     return {};
   }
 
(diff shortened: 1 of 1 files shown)


miscco (approved):

davebayer (approved):

bernhardmgruber:
It turns out this caused a break in CCCL.C, which the CI did not detect at this PR: https://github.com/NVIDIA/cccl/actions/runs/25811557957/job/75830204599?pr=8965#step:4:946

bdice:
This bug is also blocking cuOpt adoption of CUDA 13.2, reproducible on RTX PRO 6000 Blackwell and DGX Spark. Thanks for the fix. I'll update RAPIDS to use a newer CCCL commit accordingly.

PR #8851, opened 2026-05-06
Refactor `DeviceReduce` helpers

bernhardmgruber:
- [x] Merge before: #8809

PR #8924, opened 2026-05-12
Fix nightly GCC8 build regressions in bit_cast and DeviceReduce

alliepiper:
## Summary

Two unrelated regressions surfaced together in the [2026-05-12 nightly run on `main`](https://github.com/NVIDIA/cccl/actions/runs/25713127213). Both block all GCC builds on older CTKs (12.0, 12.X) for libcu++, and all GCC 8 CUB builds.

### libcu++: `bit_cast<__float128>` static_assert fires

`__bit_cast_memcpy` is reached on GCC 8/9/10 because those host compilers don't provide `__builtin_bit_cast`. Commit 8b28e1d244 (#8265) replaced `is_trivially_default_constructible_v` with the C++20-style `default_initializable` concept; the concept emulation evaluates **false** for `__float128` in C++17 mode, even though the built-in scalar is default-initializable.

Restore an equivalent type trait (`is_default_constructible_v`), which queries the compiler builtin directly and works on every supported GCC. The `_CCCL_COMPILER(GCC, <=, 7)` gate is no longer needed and is removed.

### CUB: `DeviceReduce::reduce_impl` flags `reduction_op` as unused

Commit b60f063594 (#8851) merged the three determinism overloads into a single `if constexpr` template. The `__gpu_to_gpu` branch dispatches to RFA, which hardcodes `deterministic_sum_t<accum_t>` internally and does not accept a reduction operator. GCC 8's `-Werror=unused-but-set-parameter` flags `reduction_op` because that branch never references it.

The parameter must remain in the signature (the other two branches use it). The invariant — enforced upstream by `__transform_reduce`'s `float_double_plus` check — is acknowledged with `(void) reduction_op` plus a one-line note inside the branch.

### CI override

`ci/matrix.yaml` carries a temporary `workflows.override` reproducing the failures on PR CI:
- libcu++ build on GCC 8/9/10 × CTK 12.0/12.X × all stds
- CUB build on GCC 8 × CTK 12.0/12.X × C++17

**Reset to empty before merging.**

## Test plan

- [ ] Override-matrix PR CI passes on this branch
- [ ] Reset `workflows.override` to empty
- [ ] Full PR CI passes after override reset

### DIFF
--- a/libcudacxx/include/cuda/std/__bit/bit_cast.h
+++ b/libcudacxx/include/cuda/std/__bit/bit_cast.h
@@ -23,7 +23,7 @@
 
 #include <cuda/__type_traits/is_trivially_copyable.h>
 #include <cuda/std/__concepts/concept_macros.h>
-#include <cuda/std/__concepts/constructible.h>
+#include <cuda/std/__type_traits/is_default_constructible.h>
 #include <cuda/std/cstring>
 
 #include <cuda/std/__cccl/prologue.h>
@@ -51,10 +51,8 @@ _CCCL_DIAG_SUPPRESS_GCC("-Wclass-memaccess")
 template <class _To, class _From>
 [[nodiscard]] _CCCL_API inline _To __bit_cast_memcpy(const _From& __from) noexcept
 {
-#if !_CCCL_COMPILER(GCC, <=, 7)
-  static_assert(::cuda::std::default_initializable<_To>,
-                "bit_cast memcpy fallback requires the destination type to be default initializable");
-#endif // !_CCCL_COMPILER(GCC, <=, 7)
+  static_assert(::cuda::std::is_default_constructible_v<_To>,
+                "bit_cast memcpy fallback requires the destination type to be default constructible");
   _To __temp;
   ::cuda::std::memcpy(&__temp, &__from, sizeof(_To));
   return __temp;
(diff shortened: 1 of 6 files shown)


miscco (approved):

fbusato (changes requested):
I would prefer `#if !_CCCL_COMPILER(GCC, <=, 8)`. `cuda::std::default_initializable` is more precise than `is_default_constructible`

miscco:
/ok to test [b40e516](https://github.com/NVIDIA/cccl/pull/8924/commits/b40e516de19a07524a860d8dfc34e3dd644cb00d)

libcudacxx/include/cuda/std/__bit/bit_cast.h
```diff
@@ -51,10 +51,8 @@ _CCCL_DIAG_SUPPRESS_GCC("-Wclass-memaccess")
 template <class _To, class _From>
 [[nodiscard]] _CCCL_API inline _To __bit_cast_memcpy(const _From& __from) noexcept
 {
-#if !_CCCL_COMPILER(GCC, <=, 7)
-  static_assert(::cuda::std::default_initializable<_To>,
-                "bit_cast memcpy fallback requires the destination type to be default initializable");
-#endif // !_CCCL_COMPILER(GCC, <=, 7)
+  static_assert(::cuda::std::is_default_constructible_v<_To>,
+                "bit_cast memcpy fallback requires the destination type to be default constructible");
```
fbusato:
```suggestion
#if !_CCCL_COMPILER(GCC, <=, 8)
  static_assert(::cuda::std::default_initializable<_To>,
                "bit_cast memcpy fallback requires the destination type to be default constructible");
#endif // !_CCCL_COMPILER(GCC, <=, 8)
```

fbusato (changes requested):

fbusato (approved):

PR #8928, opened 2026-05-12
[c2h] Make catch macros work on device

davebayer:
Catch 2 test macros are unsupported in device code, so we have other alternatives like `CUDAX_REQUIRE` or `CCCLRT_REQUIRE` that should work in both host and device code.

I don't like this, so I found a way to allow us using the original macros directly. If `CATCH_CONFIG_PREFIX_ALL` macro is defined, catch 2 prepends these macros with `CATCH_` and the non-prefixed versions are not defined. We can then define those manually and define our own behaviour in device code.

 The output for failed tests looks something like:
 ```
/home/dabayer/cccl/cudax/test/group/synchronizer/lane_synchronizer.cu:48:
    CHECK(synchronizer_instance.__lane_mask_ != (this_lane_mask | another_lane_mask)) failed
    block [0, 0, 0], thread [4, 3, 0]
```
I've also implemented some extensions, such as:
```cpp
REQUIRE_CUDA(cuSomeDriverCall(...));
CHECK_CUDA(cuSomeDriverCall(...));

REQUIRE_CUDART(cudaSomeCudaRuntimeCall(...));
CHECK_CUDART(cudaSomeCudaRuntimeCall(...));
```

### DIFF
--- a/c2h/include/c2h/catch2_test_macros.h
+++ b/c2h/include/c2h/catch2_test_macros.h
@@ -1,5 +1,5 @@
-// SPDX-FileCopyrightText: Copyright (c) 2011-2022, NVIDIA CORPORATION. All rights reserved.
-// SPDX-License-Identifier: BSD-3-Clause
+// SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 #pragma once
 
@@ -55,19 +55,9 @@
                       }                                                                \
                     }))
 
-#define REQUIRE_THROWS(...)                                                           \
-  NV_IF_ELSE_TARGET(NV_IS_HOST, (CATCH_REQUIRE_THROWS(__VA_ARGS__);), ({              \
-                      __VA_ARGS__;                                                    \
-                      C2H_INTERNAL_DEVICE_TEST_PRINT("REQUIRE_THROWS", #__VA_ARGS__); \
-                      ::__trap();                                                     \
-                    }))
-#define REQUIRE_THROWS_AS(EXPR, TYPE)                                                   \
-  NV_IF_ELSE_TARGET(NV_IS_HOST, (CATCH_REQUIRE_THROWS_AS((EXPR), TYPE);), ({            \
-                      EXPR;                                                             \
-                      C2H_INTERNAL_DEVICE_TEST_PRINT("REQUIRE_THROWS_AS", #EXPR #TYPE); \
-                      ::__trap();                                                       \
-                    }))
-#define REQUIRE_NOTHROW(...) NV_IF_ELSE_TARGET(NV_IS_HOST, (CATCH_REQUIRE_NOTHROW(__VA_ARGS__);), (__VA_ARGS__;))
+#define REQUIRE_THROWS(...)    CATCH_REQUIRE_THROWS(__VA_ARGS__)
+#define REQUIRE_THROWS_AS(...) CATCH_REQUIRE_THROWS_AS(__VA_ARGS__)
+#define REQUIRE_NOTHROW(...)   NV_IF_ELSE_TARGET(NV_IS_HOST, (CATCH_REQUIRE_NOTHROW(__VA_ARGS__);), (__VA_ARGS__;))
 
 #define CHECK(...)                                                             \
   NV_IF_ELSE_TARGET(NV_IS_HOST, (CATCH_CHECK(__VA_ARGS__);), ({                \
@@ -107,8 +97,11 @@
 #define REGISTER_TEST_CASE(...)  CATCH_REGISTER_TEST_CASE(__VA_ARGS__)
 #define SECTION(...)             CATCH_SECTION(__VA_ARGS__)
 #define DYNAMIC_SECTION(...)     CATCH_DYNAMIC_SECTION(__VA_ARGS__)
-#define FAIL(...) \
-  NV_IF_ELSE_TARGET(NV_IS_HOST, (CATCH_FAIL(__VA_ARGS__);), ({ C2H_INTERNAL_DEVICE_TEST_PRINT("FAIL", #__VA_ARGS__); }))
+#define FAIL(...)                                                           \
+  NV_IF_ELSE_TARGET(NV_IS_HOST, (CATCH_FAIL(__VA_ARGS__);), ({              \
+                      C2H_INTERNAL_DEVICE_TEST_PRINT("FAIL", #__VA_ARGS__); \
+                      ::__trap();                                           \
+                    }))
 #define FAIL_CHECK(...) CATCH_FAIL_CHECK(__VA_ARGS__)
 #define SUCCEED(...)    CATCH_SUCCEED(__VA_ARGS__)
 #define SKIP(...)       CATCH_SKIP(__VA_ARGS__)
--- a/c2h/include/c2h/catch2_test_macros.h
+++ b/c2h/include/c2h/catch2_test_macros.h
@@ -10,6 +10,7 @@
 #include <catch2/catch_message.hpp>
 #include <catch2/catch_template_test_macros.hpp>
 #include <catch2/catch_test_macros.hpp>
+#include <catch2/matchers/catch_matchers.hpp>
 
 // This file implements Catch2's test macros that work both in host and device code. We globally define the
 // CATCH_CONFIG_PREFIX_ALL macro to force Catch2 to prepend it's macros with CATCH_ prefix. That allows us to implement
@@ -33,6 +34,8 @@
 
 #define C2H_INTERNAL_DEVICE_TEST_PRINT(KIND, ...) C2H_INTERNAL_DEVICE_PRINTF(KIND "(%s) failed", "" #__VA_ARGS__)
 
+// <catch2/catch2_test_macros.hpp>
+
 #define REQUIRE(...)                                                            \
   NV_IF_ELSE_TARGET(NV_IS_HOST, (CATCH_REQUIRE(__VA_ARGS__);), ({               \
                       if (!(__VA_ARGS__))                                       \
@@ -52,13 +55,13 @@
 
 #define REQUIRE_THROWS(...)                                                          \
   NV_IF_ELSE_TARGET(NV_IS_HOST, (CATCH_REQUIRE_THROWS(__VA_ARGS__);), ({             \
-                      (void) __VA_ARGS__;                                            \
+                      __VA_ARGS__;                                                   \
                       C2H_INTERNAL_DEVICE_TEST_PRINT("REQUIRE_THROWS", __VA_ARGS__); \
                       ::__trap();                                                    \
                     }))
 #define REQUIRE_THROWS_AS(EXPR, TYPE)                                                  \
-  NV_IF_ELSE_TARGET(NV_IS_HOST, (CATCH_REQUIRE_THROWS_AS(EXPR, TYPE);), ({             \
-                      (void) EXPR;                                                     \
+  NV_IF_ELSE_TARGET(NV_IS_HOST, (CATCH_REQUIRE_THROWS_AS((EXPR), TYPE);), ({           \
+                      EXPR;                                                            \
                       C2H_INTERNAL_DEVICE_TEST_PRINT("REQUIRE_THROWS_AS", EXPR, TYPE); \
                       ::__trap();                                                      \
                     }))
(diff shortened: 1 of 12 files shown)


cudax/test/common/testing.cuh
```diff
@@ -32,56 +32,13 @@ namespace cudax_async = cuda::experimental::execution; // NOLINT: misc-unused-al
...
+#define CUDAX_CHECK(condition) CHECK(condition)
 
-#define CUDAX_FAIL(message) /*                                                                   */ \
-  NV_IF_ELSE_TARGET(NV_IS_DEVICE, /*                                                             */ \
-                    (cudax_require_impl(false, message, __FILE__, __LINE__, __PRETTY_FUNCTION__);), \
-                    (FAIL(message);))
+#define CUDAX_FAIL(message) FAIL(message)
 
-#define CUDAX_CHECK_FALSE(condition) CUDAX_CHECK(!(condition))
+#define CUDAX_CHECK_FALSE(condition) CHECK_FALSE(condition)
```
davebayer, line 41: I will remove these in a separate PR

libcudacxx/test/libcudacxx/cuda/ccclrt/common/testing.cuh
```diff
@@ -26,40 +26,19 @@
...
-#define CCCLRT_REQUIRE_DEVICE(condition) \
-  ccclrt_require_impl(condition, #condition, __FILE__, __LINE__, __PRETTY_FUNCTION__);
-#define CCCLRT_CHECK_DEVICE(condition) \
-  ccclrt_require_impl(condition, #condition, __FILE__, __LINE__, __PRETTY_FUNCTION__);
-#define CCCLRT_FAIL_DEVICE(message)          ccclrt_require_impl(false, message, __FILE__, __LINE__, __PRETTY_FUNCTION__);
-#define CCCLRT_CHECK_FALSE_DEVICE(condition) CCCLRT_CHECK_DEVICE(!(condition))
+#define CCCLRT_REQUIRE_DEVICE(condition)     REQUIRE_DEVICE(condition)
+#define CCCLRT_CHECK_DEVICE(condition)       CHECK(condition)
+#define CCCLRT_FAIL_DEVICE(message)          FAIL(message)
+#define CCCLRT_CHECK_FALSE_DEVICE(condition) CHECK_FALSE(condition)
```
davebayer, line 41: ditto

cub/test/catch2_test_device_radix_sort_custom.cu
```diff
@@ -499,12 +499,12 @@ struct decomposer_t
 };
 // example-end custom-type
 
-static __host__ std::ostream& operator<<(std::ostream& os, const custom_t& self)
+__host__ std::ostream& operator<<(std::ostream& os, const custom_t& self)
```
davebayer, line 502: I got warnings about these functions being unused 🤷

c2h/include/c2h/catch2_test_macros.h
```diff
@@ -0,0 +1,186 @@
...
+// CATCH_CONFIG_PREFIX_ALL macro to force Catch2 to prepend it's macros with CATCH_ prefix. That allows us to implement
+// the non-prefixed versions ourselves.
+//
+// In host code, we just use the CATCH_-prefixed variant, in device code we implement the functionality, so it
+// corresponds the desired functionality.
+//
+// Only a subset of the Catch2's macro are provided. If needed, feel free to extend the support. Host-only macros can
+// be determined by missing NV_IF_ELSE_TARGET wrapper and immediate dispatch to CATCH_-prefixed variant.
+
+// We must pass the COND as a cstring parameter, because it might contain the '%' character that would break the printf
```
Jacobfaib, line 25: If `COND` is a c-string, when why not embed it directly in the string? Or are you anticipating it being passed as e.g. `string.c_str()`?

c2h/include/c2h/catch2_test_macros.h
```diff
@@ -0,0 +1,186 @@
...
+#define TEMPLATE_PRODUCT_TEST_CASE_SIG(...)        CATCH_TEMPLATE_PRODUCT_TEST_CASE_SIG(__VA_ARGS__)
+#define TEMPLATE_PRODUCT_TEST_CASE_METHOD(...)     CATCH_TEMPLATE_PRODUCT_TEST_CASE_METHOD(__VA_ARGS__)
+#define TEMPLATE_PRODUCT_TEST_CASE_METHOD_SIG(...) CATCH_TEMPLATE_PRODUCT_TEST_CASE_METHOD_SIG(__VA_ARGS__)
+#define TEMPLATE_LIST_TEST_CASE(...)               CATCH_TEMPLATE_LIST_TEST_CASE(__VA_ARGS__)
+#define TEMPLATE_LIST_TEST_CASE_METHOD(...)        CATCH_TEMPLATE_LIST_TEST_CASE_METHOD(__VA_ARGS__)
+
+// <catch2/matchers/catch_matchers.hpp>
+
+#define REQUIRE_THROWS_WITH(...)    CATCH_REQUIRE_THROWS_WITH(__VA_ARGS__)
+#define REQUIRE_THROWS_MATCHES(...) CATCH_REQUIRE_THROWS_MATCHES(__VA_ARGS__)
```
Jacobfaib, line 144: In my humble (and correct!) opinion, `REQUIRE_THROWS_MATCHES()` is the only proper way to check exceptions thrown as it checks both the type and expected message so there is no ambiguity on which exception is being captured. 

Perhaps we can take this time to fix up some defaults and just delete the other macros.

c2h/include/c2h/catch2_test_macros.h
```diff
@@ -0,0 +1,186 @@
...
+                    }))
+#define REQUIRE_THROWS_AS(EXPR, TYPE)                                                   \
+  NV_IF_ELSE_TARGET(NV_IS_HOST, (CATCH_REQUIRE_THROWS_AS((EXPR), TYPE);), ({            \
+                      EXPR;                                                             \
+                      C2H_INTERNAL_DEVICE_TEST_PRINT("REQUIRE_THROWS_AS", #EXPR #TYPE); \
+                      ::__trap();                                                       \
+                    }))
+#define REQUIRE_NOTHROW(...) NV_IF_ELSE_TARGET(NV_IS_HOST, (CATCH_REQUIRE_NOTHROW(__VA_ARGS__);), (__VA_ARGS__;))
+
+#define CHECK(...)                                                             \
```
Jacobfaib, line 62: Similarly, we should consider removing `CHECK()` and replacing all instances with `REQUIRE()`. I am struggling to think of a good scenario where you would want test execution to continue _after_ initial failure.

cub/test/catch2_test_device_radix_sort_custom.cu
Jacobfaib: IMO then we should remove them if they really are unused.

c2h/include/c2h/catch2_test_macros.h
pciolkosz: I think there could be value in having more information in the log from some external testing like CI or QA, so you might not have to reproduce the issue yourself, while REQUIRE would not give you enough information to figure out the failure. I wouldn't say its common and I can see using CHECK too often lead to a more messy log with multiple lines of repeated same failure for example. So its very case by case, but I think there is space for both of these macros

c2h/include/c2h/catch2_test_macros.h
davebayer: Seems that my comment is not expressive enough. The problem is that we use `printf` for printing the output. And the `%` is the introductory character for parameter formatting. And that can cause problems, because if the `COND` contains `%`, it will do the formatting instead of just printing the character.

So to prevent this, I pass it as the argument to `%s` parameter. I am open to any better solution :D

c2h/include/c2h/catch2_test_macros.h
davebayer: We definitely can! Let's move this discussion to slack/new issue. This PR really only aims to provide host + device working alternatives

c2h/include/c2h/catch2_test_macros.h
davebayer: Agreed! But maybe someone else would have a valid use case for that

cub/test/catch2_test_device_radix_sort_custom.cu
davebayer: They are used, just not during device pass and that makes `cicc` complain

c2h/include/c2h/catch2_test_macros.h
```diff
@@ -0,0 +1,186 @@
...
+    blockIdx.y,                                                                                                  \
+    blockIdx.z,                                                                                                  \
+    threadIdx.x,                                                                                                 \
+    threadIdx.y,                                                                                                 \
+    threadIdx.z)
+
+// <catch2/catch2_test_macros.hpp>
+
+#define REQUIRE(...)                                                             \
+  NV_IF_ELSE_TARGET(NV_IS_HOST, (CATCH_REQUIRE(__VA_ARGS__);), ({                \
```
pciolkosz, line 42: Don't you need to handle clang-coda here?

pciolkosz (approved):

c2h/include/c2h/catch2_test_macros.h
davebayer: It seems to be working correctly in most cases, for other problematic cases I added `REQUIRE_DEVICE`

c2h/include/c2h/catch2_test_macros.h
Jacobfaib: > o you might not have to reproduce the issue yourself, while REQUIRE would not give you enough information to figure out the failure.

Ehhh, overall I think this ends up being rare in practice.

I see it similarly to C++ compiler error messages. Yes, once in a blue moon seeing the 30 different overloads of `operator<<` that GCC failed to match (and Very Helpful(TM) description for each on why they didn't match) might help you debug the failure, but usually all you want to see is that you accidentally typed `bool is_less = foo << bar` when you meant `bool is_less = foo < bar`, and that's the first error message in that whole stack.

More often than not, `CHECK()` leads to red herring errors, e.g. when you have
```c++
CHECK(container.size() >= n);
// a lot of other unrelated code...
CHECK(container[n] == foo); // oops
```
And now you are left trying parse the crash log instead.

Jacobfaib (approved):

alliepiper (approved):
CMake signoff

c2h/include/c2h/catch2_test_macros.h
```diff
@@ -0,0 +1,189 @@
+// SPDX-FileCopyrightText: Copyright (c) 2011-2022, NVIDIA CORPORATION. All rights reserved.
+// SPDX-License-Identifier: BSD-3-Clause
```
bernhardmgruber: Suggestion: this seems to be new logic, so consider:

```suggestion
// SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
```

c2h/include/c2h/catch2_test_macros.h
```diff
@@ -0,0 +1,186 @@
...
+                        ::__trap();                                                    \
+                      }                                                                \
+                    }))
+
+#define REQUIRE_THROWS(...)                                                           \
+  NV_IF_ELSE_TARGET(NV_IS_HOST, (CATCH_REQUIRE_THROWS(__VA_ARGS__);), ({              \
+                      __VA_ARGS__;                                                    \
+                      C2H_INTERNAL_DEVICE_TEST_PRINT("REQUIRE_THROWS", #__VA_ARGS__); \
+                      ::__trap();                                                     \
+                    }))
```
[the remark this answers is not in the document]
bernhardmgruber: Suggestion: Should we just statically assert/ fail to compile if you put a `REQUIRE_THROWS` in device code?

c2h/include/c2h/catch2_test_macros.h
bernhardmgruber: I use `CHECK()` a lot where I want to validate a bunch of values after I ran a test. For example, I do a `SortPairs` and afterwards validate if the keys and values correspond to two reference arrays. I want to see both failures if they mismatch. Please keep the `CHECK()` macro.

c2h/include/c2h/catch2_test_macros.h
bernhardmgruber: I would strongly advise to stay with the exact semantics of Catch2. If `REQUIRE_THROWS_MATCHES` is indeed better, we should just use an agent to rewrite the tests not using it.

If we change the meaning of macros between C2H and Catch2, it will create a lot of confusion for maintainers.

bernhardmgruber (approved):
I am fine with this, but I want to point out that this may create additional maintenance work when upgrading Catch2.

c2h/include/c2h/catch2_test_macros.h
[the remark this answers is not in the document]
davebayer: I've removed device support from these, it probably doesn't make sense to provide them. We can reimplement them later if needed

PR #8944, opened 2026-05-12
[Tile] Mark `seed_seq` as `_CCCL_HOST_DEVICE_API`

miscco:
It relies on dynamic memory, which is not supported in tile mode


libcudacxx/include/cuda/__random/pcg_engine.h
```diff
@@ -151,8 +151,8 @@ class __pcg_uint128_fallback
   }
 
 #if _CCCL_STD_VER <= 2017
-  [[nodiscard]] _CCCL_API constexpr friend bool
-  operator!=(__pcg_uint128_fallback __lhs, __pcg_uint128_fallback __rhs) noexcept
+  [[nodiscard]]
+  _CCCL_API constexpr friend bool operator!=(__pcg_uint128_fallback __lhs, __pcg_uint128_fallback __rhs) noexcept
```
fbusato, line 155: this formatting change is weird...

libcudacxx/include/cuda/std/__random/seed_seq.h
```diff
@@ -106,7 +106,7 @@ class _CCCL_TYPE_VISIBILITY_DEFAULT seed_seq
 
   /// @brief Returns the number of seed values stored.
   /// @return Number of internal seed values (may be 0).
-  [[nodiscard]] _CCCL_API constexpr size_t size() const noexcept
+  [[nodiscard]] _CCCL_HOST_DEVICE_API constexpr size_t size() const noexcept
```
fbusato, line 109: what is the reason to change it here? or is more to provide the same execution space for all functions in the class?

fbusato (approved):

bernhardmgruber (approved):

libcudacxx/include/cuda/std/__random/seed_seq.h
miscco: seed_seq relies on dynamic allocations, which is not allowed in tile mode

miscco: So yeah I want the whole class to have the same execution space, because we cannot construct it in tile mode and I am wary whether we can then use it there

PR #8948, opened 2026-05-12
Add CCCL workflow skills and helper agents

### LINKED ISSUE #7829: Move SASS diff instructions from AGENTS.md to SKILLS.md
I feel like this is something that should be exposed as a separate skill in SKILLS.md like so:

```
Skill: CUDA Codegen SASS Verification

Description:
Compare SASS output before/after CUDA changes to validate performance stability.

Inputs:
- Baseline CUDA source
- Modified CUDA source
- Target architecture

Process:
1. Compile with nvcc -arch=...
2. Extract SASS via cuobjdump
3. Normalize output
4. Diff instruction classes
5. Report deltas
6. Flag register or occupancy regressions

Constraints:
- Must use same compiler flags
```

This way is slightly more structured as you can instruct an agent to specifically use that skill rather than letting it memoize the instructions from the AGENTS.md file when you instruct it to do a SASS comparison.

_Originally posted by @sleeepyjack in https://github.com/NVIDIA/cccl/issues/7733#issuecomment-3970302419_

### LINKED ISSUE #8149: [DOC]: Infrastructure documentation and agent tooling
### Is this a duplicate?

Update

The infrastructure docs need a refresh and perhaps some new agent skills for common tasks.

_No response_

alliepiper:
closes https://github.com/NVIDIA/cccl/issues/8149 

See [this new doc](https://github.com/alliepiper/cccl/blob/1be52f83f9110c9248df440e27cdc87204c68846/docs/maintainers/how_tos/use_cccl_agent_skills.rst) for details.


.agent/skills/libcudacxx-style/SKILL.md
```diff
@@ -1,105 +0,0 @@
...
-All non-public symbols must be C++ reserved identifiers:
-
-- `_` for macros and template parameters, e.g. `_MY_MACRO`., `_MyParameter`.
-- `__` for all other symbols, e.g. `__my_variable`.
-
-- Avoid single letter names for template parameters. Wrong: `_T`, correct: `_Tp`.
-
-## Variables
-
-- All variables that are not modified must use `const`. This includes variables initialized by casts (`static_cast`, `reinterpret_cast`, `bit_cast`), function return values, and loop-invariant computations.
```
fbusato, line 23: it looks some details have been lost moving to the new organization

.agent/skills/cccl-libcudacxx/references/style/testing.md
```diff
@@ -0,0 +1,65 @@
...
+// RUN: %{run}
+
+#include <cuda/std/...>
+// test body
+```
+
+The `RUN:` lines invoke the lit substitution variables defined by the test suite
+configuration.
+
+## Doxygen in production headers
```
fbusato: why doxygen is in testing?

.agent/skills/cccl-libcudacxx/references/style/templates.md
```diff
@@ -0,0 +1,46 @@
...
+This makes the return type visible at the point of declaration without requiring the
+reader to parse the full signature.
+
+## SFINAE and `enable_if`
+
+Prefer `_CCCL_REQUIRES` or concept-based constraints when available. When falling back to
+`enable_if`, place it in the trailing return type or as a defaulted non-type template
+parameter — never in the function parameter list.
+
+## Modern C++
```
fbusato: this is not strictly related to templates

.agent/skills/libcudacxx-style/SKILL.md
alliepiper: Restored and checked for other drops, this was the only bit that seems missing. lmk if you notice anything else.

Move the other commented bits to more appropriate locations (naming.md and a new doxygen.md).

alliepiper:
@coderabbitai We don't run markdown linters. Please don't spam about guidelines we don't actually enforce.

jrhemstad (changes requested):
I took a look through many of the skills and I have some fundamental questions about what goes in documentation vs skills. My core concern is drift between what our skills say and what our documentation says. Any agent will start with a copy of our docs available, so skills should be something net new on top of the docs that would go beyond something that would be appropriate to put into docs.

jrhemstad:
To start, I'd like to see a reference doc added to give maintainers an entrypoint into understanding what skills we have available and how to use them. 

My preference would be to add a "How to use CCCL maintainer skills" added under Maintainer Docs > How Tos

alliepiper:
> I took a look through many of the skills and I have some fundamental questions about what goes in documentation vs skills. My core concern is drift between what our skills say and what our documentation says. Any agent will start with a copy of our docs available, so skills should be something net new on top of the docs that would go beyond something that would be appropriate to put into docs.

@jrhemstad The point is to avoid having the agent load the full docs into memory to get basic usage info for every-day tasks. The skills are organized into progressively loaded "agent cheat sheets" for the CLI utilities that we want to encourage it to use. It saves significant context over long sessions to not clutter the initial context with repeated searches for the same docs on build/test etc, and cuts down on tool calls / startup time as well.

It also includes references to where the relevant docs, scripts, and config files are for each subject, so more detailed info can be easily looked up if needed.

We can revise where the line is between "cheat sheet" and "lookup later", but there's value in teaching it the progressively loaded skills over just saying "here's our docs; read them and figure it out yourself" every time.

The drift concern isn't trivial, but it's also something that you can just ask an agent to review and update periodically. Same with moving agent skill info into docs. Once these are merged, you can easily ask claude to sort these out however you please. I cannot guess how you want these structured, if you have strong opinions that block merging I encourage a followup commit that addresses them.

**I vote for merging this now and shuffling these around to taste later**. It would be very helpful to have these available to other development work while sorting this out.

alliepiper:
> To start, I'd like to see a reference doc added to give maintainers an entrypoint into understanding what skills we have available and how to use them.
> 
> My preference would be to add a "How to use CCCL maintainer skills" added under Maintainer Docs > How Tos

https://github.com/alliepiper/cccl/blob/2001aeee140e4aa2f13f9ef558953a3eb665f424/.agent/skills/cccl-README.md has this information already. I can push a fix to the pre-commit that moves it to the docs.

.agent/agents/cccl-ci-fetch-failures.md
```diff
@@ -0,0 +1,92 @@
+---
+name: cccl-ci-fetch-failures
+description: "Fetch failed jobs from a CCCL CI run — given a PR# or run ID, returns TSV of `<job-id>\\t<full-name>\\t<grouping-hint>` at a caller-specified path. Handles `gh api --paginate` slurp gotcha. Non-interactive, read-only. Called by `cccl-triage`."
+model: haiku
+color: cyan
+tools: Bash, Read
```
brycelelbach, line 6: Can you make any skills like this that take a conceptual argument have the `argument-hint` metadata? It's not part of the agentskills.io standard, but it is recognized by claude and most agent harnesses.

There are various conventions for placeholders for the arguments; claude officially supports `$name`, but I tend to use `<name>`.

An example: https://github.com/evo-hq/evo/blob/main/plugins/evo/skills/optimize/SKILL.md

Use syntax to denote defaults (`foo=5`), optional arguments (`[foo=<N>]`, etc. 

I would also suggest always having a section that enumerates the arguments.

.agent/agents/cccl-ci-fetch-failures.md
```diff
@@ -0,0 +1,92 @@
...
+---
+
+You are a non-interactive read-only `cccl-ci-fetch-failures` agent. The caller has a PR number or workflow run ID and wants a TSV of failed jobs for downstream summarization or override-matrix generation. You never modify files beyond writing the named output TSV and a raw-API scratch file, never call `AskUserQuestion`, never spawn subagents.
```
brycelelbach, line 9: Instead of forbidding AskUserQuestion explicitly (something that is specific to particular harnesses), describe the behavior you don't want "never use a tool that queries the human interactively for input".

I think it's better to have a separate constraints section that enumerates:
- What the agent can do.
- What the agent cannot do.
- What the agent should ask permission to do.

.agent/agents/cccl-ci-fetch-failures.md
```diff
@@ -0,0 +1,92 @@
...
+```
+gh api repos/NVIDIA/cccl/actions/runs/<RUN_ID>/jobs?per_page=100 --paginate > <scratch>/jobs_raw.json
+```
+
+`--paginate` concatenates objects; subsequent `jq` needs `-s`.
+
+### 3. Extract failures
+
+```
+jq -s -r '[.[].jobs[] | select(.conclusion == "failure")] | .[] | [.id, .name] | @tsv' \
```
brycelelbach, line 44: Put this into a script inside of the skill. Complex scripting is an easy thing for agents to trip up on, even when you give them verbatim examples.

.agent/agents/cccl-ci-fetch-failures.md
```diff
@@ -0,0 +1,92 @@
+---
```
brycelelbach, line 1: Why structure these as agents instead of skills?

.agent/agents/cccl-ci-fetch-failures.md
```diff
@@ -0,0 +1,92 @@
...
+
+run_id: <RUN_ID>
+total_failures: <N>
+
+tally:
+  <toolchain>|<project>|<variant>: <count>
+  ...
+
+output_path: <output>
+```
```
brycelelbach, line 78: I would add a script that verifies the "schema" of output to ensure it is correct and consistent; then ask the agent/skill to check the output with the script.

.agent/agents/cccl-ci-fetch-failures.md
```diff
@@ -0,0 +1,92 @@
...
+
+## Stop conditions
+
+- Missing `pr:` and `run:` → `STATUS: UNDER_BRIEFED`.
+- No failed jobs → `STATUS: NO_FAILURES`.
+- `gh api` non-zero exit → return raw stderr, `STATUS: UNDER_BRIEFED`.
+
+## Hard prohibitions
+
+- No `AskUserQuestion`. Not available; not applicable.
```
brycelelbach, line 88: The easiest way to prevent this is to just run the harness in non-interactive mode. I do agree it's good to have a rule about this though.

.agent/agents/cccl-ci-fetch-failures.md
alliepiper: I think that'd make a great followup PR!

.agent/agents/cccl-ci-fetch-failures.md
alliepiper: We can def soften those -- the goal is to route things through the /cccl-clarify skill instead of having to decide "end turn with question? use harness prompt as a one-off? is the question already in my context? (I found this greatly improves interactions with claude as a personal skill, esp when combined with other skills)

.agent/agents/cccl-ci-fetch-failures.md
brycelelbach, line 1: I would structure this as a plugin (see https://github.com/evo-hq/evo/tree/main/plugins/evo for an example).

That gives you namespacing (e.g. `/cccl:bench` instead of `/cccl-bench`) and makes it easier for someone to install these skills/agents into their harness if they're working outside of the repo.

It also helps with version management, which can be important as some harnesses "cache" skills and the cache can get stale.

alliepiper: I considered plugins and the complexity didn't seem worthwhile. If you'd like to take that on, please go ahead :)

alliepiper: They read a 20MB+ logs and output about 20 lines. Rather do that in haiku (or eventually, a script as a first pass)

.agent/agents/cccl-ci-fetch-failures.md
alliepiper: > I would

You may! 😄

alliepiper:
Here's the updated users guide, rewritten toward a more specific audience:

https://github.com/alliepiper/cccl/blob/1be52f83f9110c9248df440e27cdc87204c68846/docs/maintainers/how_tos/use_cccl_agent_skills.rst

docs/maintainers/how_tos/use_cccl_agent_skills.rst
```diff
@@ -0,0 +1,313 @@
...
+  through real CCCL work.
+
+What the skills do (and don't)
+------------------------------
+
+The skills encode entry points and project conventions — they do not
+replace engineering judgment. A skill knows that ``cccl-infra`` regenerates
+devcontainers from ``ci/matrix.yaml`` rather than hand-editing each
+``devcontainer.json``; it does not know whether bumping CTK 13.3 is the
+right move this week. That part stays with you.
```
jrhemstad, line 24: suggestion: This is an example of overly-verbose-LLM noise that I don't think is needed in this document.

docs/maintainers/how_tos/use_cccl_agent_skills.rst
```diff
@@ -0,0 +1,313 @@
...
+
+1. Have the agent load ``cccl`` first — it carries the intent → skill
+   routing table. This should happen automatically after the first prompt,
+   but can be invoked manually if needed using the client's skill load command
+   (eg ``/cccl`` on Claude).
+2. Describe what you're actually trying to do, in plain language.
+   ``"Triage PR #9001"``, ``"resplit this branch"``, ``"bisect this
+   segfault"`` are all the agent needs to pick the right entry.
+3. Approve the mutating actions (commits, pushes, PR writes, CI
+   triggers) when the skill asks.
```
jrhemstad, line 309: This is more what I was looking for for a How To guide. I think this document could effectively just be boiled down to something that looks more like this.

docs/maintainers/how_tos/use_cccl_agent_skills.rst
```diff
@@ -0,0 +1,313 @@
+CCCL Agent Skills
```
jrhemstad, line 1: suggestion: These docs are currently a bit too verbose and sprinkled with LLM-y over explanation or conversation that just adds visual noise.  I'd like to see it tightened up and made more focused. 

Part of the problem is that this doc attempts to be both a reference guide and a how-to document.

A reference document answers "What skills are available and what do they do"

A how-to document answers "How do I use the skills to accomplish a task"

More importantly, it's not clear to me, does someone using these skills need to even know about each and every individual skill? 

For example, consider one of the example prompts and explanations 

>    *"PR #8965 is failing in CI on the libcudacxx jobs for cuda13.2/gcc14.
   Figure out why, fix it, commit with override tags so we don't re-run
   the green half of the matrix, push, mark ready."*

>   ``cccl-triage`` (cluster CI failures and summarize representatives) →
   you engineer the fix → ``cccl-ci-overrides`` (minimal override
   snippet) → ``cccl-commit`` (chunk walk, optional test gate, message
   draft) → ``cccl-pr`` (push, ``/ok to test``, ready). The fix arrives
   with an override that re-runs only the jobs that matter.

That prompt doesn't say anything about what skills to use nor even guarantees the agent will actually use the skills in the way the explanation claims. So I'm not sure what this documentation actually accomplishes? I don't mean that to be critical, but everything is so new here that I just legitimately do not know.

.agent/skills/cccl-c/SKILL.md
```diff
@@ -0,0 +1,102 @@
...
+│   ├── src/                   — CUDA/C++ implementation (one .cu per algorithm)
+│   │   ├── util/              — shared context, error, type, tuning utilities
+│   │   ├── nvrtc/             — NVRTC / nvJitLink helpers
+│   │   ├── jit_templates/     — JIT type-wrapper template system (see below)
+│   │   └── hostjit/           — optional LLVM-backed host JIT (optional, ~20 min build)
+│   ├── test/                  — CTest-based C++ tests (one per algorithm)
+│   │   └── freestanding/      — header-isolation + bitcode tests
+│   └── cmake/                 — CParallelHeaderTesting.cmake
+└── experimental/stf/          — C bindings for the STF (stream task framework) backend
+```
```
jrhemstad, line 33: This is a good example picked at random of broader theme I'm worried about: Hardcoding information (like directory structure) that is likely to change and could be trivially found through other means (like running `ls`) feels like a recipe for bit rot and a non-delightful experience with agents working from stale or incorrect information.

docs/maintainers/how_tos/use_cccl_agent_skills.rst
alliepiper: My goal with this document was to give devs an overview of what the skills can concretely do in their workflows, rather than just saying "agents can do stuff now". I find the remainder of the document compelling and informative.

> That prompt doesn't say anything about what skills to use nor even guarantees the agent will actually use the skills in the way the explanation claims. So I'm not sure what this documentation actually accomplishes?

The user doesn't have to invoke any specific skills. They can use plain language, and the skills will be progressively loaded in on demand (the description field in the header is where trigger conditions are specified).

This example is meant to say "If you ask this, the agent will deploy these skills to perform these actions (`cccl-triage` is a skill, etc. The user doesn't *have* to invoke them manually, but they can).

.agent/skills/cccl-c/SKILL.md
alliepiper: > Claude, review the cccl- skills and agents against the current repo, and fix drift/omissions.

With frontier models and harnesses, the skills are self-healing and self-maintaining.

docs/maintainers/how_tos/use_cccl_agent_skills.rst
alliepiper: It's fair expectation setting IMO. Agent capabilities are genuinely surprising, and I don't know if new users will intuitively understand where the line is.

docs/maintainers/how_tos/use_cccl_agent_skills.rst
alliepiper: Then this boils down to "we have agent skills that do stuff" and I'd question the point of this doc entirely.

.agent/skills/cccl-libcudacxx/references/style/testing.md
```diff
@@ -0,0 +1,138 @@
...
+
+| Category   | Location                       | Runner                        |
+|------------|--------------------------------|-------------------------------|
+| Lit tests  | `libcudacxx/test/` (recursive) | `llvm-lit` / CCCL lit wrapper |
+| Unit tests | `libcudacxx/test/` (CMake targets) | CTest                     |
+
+Run lit tests via `cccl-test` (the underlying `build_and_test_targets.sh` script takes a
+`--lit-tests` flag). Unit tests run as normal CTest targets.
+
+## Directory layout
```
fbusato, line 13:
```suggestion
## Test purpose

libcudacxx tests validate library functionality. In particular:

- Edge cases.
- Input and output types.
- Exception behavior.
- Runtime and constant-evaluation behavior.
- Device and host behavior.

## Directory layout
```

.agent/skills/cccl-libcudacxx/references/style/testing.md
```diff
@@ -0,0 +1,138 @@
...
+
+int main(int, char**) {
+  // ... compile-time + runtime checks ...
+  return 0;
+}
+```
+
+Rules:
+
+- Always include top-level headers — never internal `__`-prefixed headers.
```
fbusato, line 58:
```suggestion
- All tests must have the correct license banner.
- Always include top-level headers — never internal `__`-prefixed headers.
```

.agent/skills/cccl-libcudacxx/references/style/templates.md
```diff
@@ -0,0 +1,35 @@
...
+- Avoid single-letter names. `_T` and `_U` are prohibited; `_Tp` and `_Up` are the
+  conventional replacements.
+
+## `constexpr` functions
+
+Use `constexpr` for all functions that do not depend on run-time features (raw pointers,
+device memory, system calls). The default should be `constexpr`; opt out only when a
+specific run-time dependency prevents it.
+
+## Trailing return types
```
fbusato, line 16:
```suggestion
- Use `_CCCL_CONSTEVAL` when a function can only be evaluated at compile time.

## Trailing return types
```

.agent/skills/cccl-libcudacxx/references/style/templates.md
```diff
@@ -0,0 +1,35 @@
...
+- Use `CamelCase` prefixed with `_` for all template parameters: `_Tp`, `_Up`, `_Key`,
+  `_Value`, `_Alloc`.
+- Avoid single-letter names. `_T` and `_U` are prohibited; `_Tp` and `_Up` are the
+  conventional replacements.
+
+## `constexpr` functions
+
+Use `constexpr` for all functions that do not depend on run-time features (raw pointers,
+device memory, system calls). The default should be `constexpr`; opt out only when a
+specific run-time dependency prevents it.
```
fbusato, line 14:
```suggestion
- Use `constexpr` for all functions that do not depend on run-time features (raw pointers, device memory, system calls). The default should be `constexpr`; opt out only when a specific run-time dependency prevents it.
```

.agent/skills/cccl-libcudacxx/references/style/headers.md
```diff
@@ -0,0 +1,93 @@
...
+- `libcudacxx/include/cuda/atomic` → `_CUDA_ATOMIC`
+
+The closing `#endif` always carries a comment repeating the guard name:
+```cpp
+#endif // _CUDA_STD_ATOMIC
+```
+
+## License selection
+
+| Directory                                                | License                               |
```
fbusato, line 88:
```suggestion
All headers must have the correct license banner.

| Directory                                                | License                               |
```

.agent/skills/cccl-precommit/SKILL.md
```diff
@@ -0,0 +1,102 @@
...
+  run", "clang-format failed", "pre-commit hook failed", "fix formatting".
+---
+
+# cccl-precommit
+
+Reference and orientation for CCCL's pre-commit setup. Configuration lives
+in `.pre-commit-config.yaml`; tool settings (ruff, codespell, mypy) live in
+the root `pyproject.toml`; CMake formatter settings are in `.gersemirc`.
+
+## Hook inventory
```
fbusato, line 16: is it necessary to list all hooks?

.agent/skills/cccl-sass-diff/SKILL.md
```diff
@@ -0,0 +1,31 @@
...
+
+- Target/library being built.
+- SM architectures (detect, offer, confirm).
+- Baseline + candidate refs.
+- SASS (`cuobjdump -sass`) or PTX (`cuobjdump -ptx`).
+
+## Workflow
+
+1. Build both versions with the same arches and flags.
+2. Dump disassembly to `/tmp/claude/<sessionid>/{baseline,candidate}.sass`.
```
fbusato, line 22: this should not be specific to claude

.agent/skills/cccl-sass-diff/SKILL.md
```diff
@@ -0,0 +1,31 @@
...
+3. Normalize both identically: strip addresses, build IDs, paths, timestamps, whitespace; drop empty/comment lines.
+4. `diff -u` the normalized listings.
+5. Classify — ignore register renames with identical opcodes/operands, label renumbering, formatting-only
+   differences.
+6. Report top 5 non-trivial regions: kernel name, change type (opcode, memory-access size, register count delta,
+   control-flow), normalized line numbers, plain-language interpretation. Or: "only noise detected".
```
fbusato, line 28: this should be a script otherwise it will be very token inefficient (called every time)

.agent/skills/cccl-test/SKILL.md
```diff
@@ -0,0 +1,89 @@
+---
```
fbusato, line 1: not 100% sure about this skill. I would expect the developer to focus on a single sub-project at the time. I would suggest to move test description per sub-project instead of globally

.agent/skills/cccl_detail-cpp-macros/SKILL.md
```diff
@@ -0,0 +1,137 @@
+---
+description: |
```
fbusato, line 2: not sure if this SKILL is really needed. Agents already explore near code. Also, we could directly pointed out to the position, without repeating everything

.agent/skills/cccl-cub/SKILL.md
```diff
@@ -0,0 +1,121 @@
+---
+description: |
```
fbusato, line 2: this SKILL only replies to questions, it doesn't perform actions. Not sure if really needed. Agents already can do that

.agent/skills/cccl-build/references/arch-flag.md
```diff
@@ -0,0 +1,22 @@
+# Architecture flag forms
```
fbusato, line 1: this doesn't look specific to CCCL. LLMs should be already aware of it

fbusato (changes requested):
provided a few suggestions. My main concern is that the SKILLs are used as knowledge base, not to perform actions. Secondly, some of them just list paths, code, tools, which an agent should already be aware of. Both add little value compared to agent initialization, and the risk is to just pollute the context

.agent/skills/cccl-sass-diff/SKILL.md
alliepiper: This is an existing procedure in `main`, just reinserted into the new system. I agree that this is a good candidate for improvement, but not something that needs to be addressed in this PR.

.agent/skills/cccl-test/SKILL.md
alliepiper: The testing instructions are the same across projects, minus slightly different flags for running lit single targets, so a single skill makes sense.

This is also more discoverable IMO:

Agent needs to test, sees a test skill , vs
Agent needs to test, and has to know to dig a few levels deep into the project structure hierarchy skills to just find "ci/util/build_and_test_target.sh" no matter which path they take.

Happy to explore further, but think this can be held off for a followup.

.agent/skills/cccl_detail-cpp-macros/SKILL.md
alliepiper: Agreed that this is low value.

alliepiper:
> My main concern is that the SKILLs are used as knowledge base, not to perform actions.

The knowledge base aspect is intentional. Skill are not only for actions, they serve the same purpose as AGENTS.md / CLAUDE.md: To orient the agent within the repo, inform it of conventions, teach it the tools and workflows. This is knowledge-base work.

> Secondly, some of them just list paths, code, tools, which an agent should already be aware of. Both add little value compared to agent initialization, and the risk is to just pollute the context.

This is *how* the agent becomes aware of these things. Otherwise it needs to explore the codebase and research everything from scratch every session. The alternative is to list it all in AGENTS.md -- which *will* pollute the context. Moving it to skills allows the harness to progressively load only the bits they need.

docs/maintainers/how_tos/use_cccl_agent_skills.rst
jrhemstad: > My goal with this document was to give devs an overview of what the skills can concretely do in their workflows

This is valuable, but I think this is ultimately a different document. This would be appropriate in an "Example Prompts for Maintainers" reference document. The skills are effectively an implementation detail of the prompts.

PR #8951, opened 2026-05-12
`cuda::std::simd` F32x2 cleanup/refactoring

fbusato:
## Description

This PR only focuses on cleanup/refactoring. No functional changes.

- Split F32x2 file in two: intrinsics, array level. This will help to keep the simd optimizations organized with next PRs.
- Same for codegen test.
- Add a couple of missing headers
- Full qualifications for one function
- Add `NV_IF_TARGET` for `asm` code + `_CCCL_VERIFY`


libcudacxx/include/cuda/std/__simd/specializations/fp32x2_intrinsics.h
```diff
@@ -26,178 +26,133 @@
...
-[[nodiscard]] _CCCL_DEVICE_API inline ::float2 __add_f32x2(const ::float2 __lhs, const ::float2 __rhs) noexcept
+[[nodiscard]] _CCCL_DEVICE_API inline ::float2
+__add_f32x2([[maybe_unused]] const ::float2 __lhs, [[maybe_unused]] const ::float2 __rhs) noexcept
 {
   ::float2 __result{};
 #  if _CCCL_HAS_SIMD_F32X2_INTRINSICS()
-  NV_IF_TARGET(NV_IS_EXACTLY_SM_100, (__result = ::__fadd2_rn(__lhs, __rhs);))
+  NV_IF_TARGET(NV_IS_EXACTLY_SM_100,
+               (__result = ::__fadd2_rn(__lhs, __rhs);),
+               (_CCCL_VERIFY(false, "cuda::std::simd::__add_f32x2: Unsupported architecture");))
```
davebayer: Shouldn't this be:
```suggestion
  NV_IF_TARGET(NV_PROVIDES_SM_100,
               (__result = ::__fadd2_rn(__lhs, __rhs);),
               (_CCCL_VERIFY(false, "cuda::std::simd::__add_f32x2: Unsupported architecture");))
```

libcudacxx/include/cuda/std/__simd/specializations/fp32x2_intrinsics.h
```diff
@@ -26,178 +26,133 @@
...
-      : "f"(__lhs.x), "f"(__lhs.y), "f"(__rhs.x), "f"(__rhs.y));
+  NV_IF_TARGET(
+    NV_IS_EXACTLY_SM_100,
+    (asm("{"
+         ".reg .b64 __lhs, __rhs, __result;"
+         "mov.b64 __lhs, {%2, %3};"
+         "mov.b64 __rhs, {%4, %5};"
+         "add.f32x2 __result, __lhs, __rhs;"
+         "mov.b64 {%0, %1}, __result;"
+         "}" //
```
davebayer, line 52: Can't we do just
```suggestion
    (asm("add.f32x2 {%0, %1}, {%2, %3}, {%4, %5};"
```

libcudacxx/include/cuda/std/__simd/specializations/fp32x2_intrinsics.h
fbusato: no, FP32x2 is not supported in SM120

libcudacxx/include/cuda/std/__simd/specializations/fp32x2_intrinsics.h
fbusato: without `NV_IF_TARGET` the code fails to compile with SM != 100, see https://godbolt.org/z/TqWEszv3a

libcudacxx/include/cuda/std/__simd/specializations/fp32x2_intrinsics.h
davebayer: It is, it's sm100 intrinsics, thus usable with any cc >= 100. See https://godbolt.org/z/7exGf9n37

libcudacxx/include/cuda/std/__simd/specializations/fp32x2_intrinsics.h
davebayer: I meant keeping the `NV_IF_TARGET` but making it a one liner instead of 6 :D

fbusato: I tried but ptxas or clang complained about it. I will try again

libcudacxx/include/cuda/std/__simd/specializations/fp32x2_intrinsics.h
fbusato: the only GPU architecture where it is supported by HW is SM100. It is emulated on the other ones

libcudacxx/include/cuda/std/__simd/specializations/fp32x2_intrinsics.h
fbusato: it doesn't work. `add.f32x2` requires `.b64` not two `f` inputs

libcudacxx/include/cuda/std/__simd/specializations/fp32x2_intrinsics.h
davebayer: But do we care that it's emulated? I mean what if a newer architecture supports fp32x2 operations natively as well? If it doesn't hurt performance, I would rather have sm100+ always using the intrinsic

libcudacxx/include/cuda/std/__internal/features.h
```diff
@@ -103,8 +103,13 @@
 
 #define _CCCL_HAS_SIMD_F32X2_INTRINSICS() (_CCCL_CUDACC_AT_LEAST(12, 8) && _CCCL_HAS_CTK() && !_CCCL_COMPILER(CLANG))
```
davebayer:
```suggestion
#define _CCCL_HAS_SIMD_F32X2_INTRINSICS() (_CCCL_CUDACC_AT_LEAST(12, 8) && _CCCL_HAS_CTK() && !_CCCL_CUDA_COMPILER(CLANG))
```
Must be fixed everywhere

libcudacxx/include/cuda/std/__simd/math/fma.h
```diff
@@ -0,0 +1,53 @@
...
+  return __simd_fma_impl(__x_vec, __y_vec, __z_vec); // ADL
+}
+
+_CCCL_END_NAMESPACE_CUDA_STD_SIMD
+
+//----------------------------------------------------------------------------------------------------------------------
+
+_CCCL_BEGIN_NAMESPACE_CUDA_STD
+
+using simd::fma;
```
Jacobfaib, line 47:
```suggestion
using ::cuda::std::simd::fma;
```
Technically function not type resolution so better safe than sorry?

libcudacxx/include/cuda/std/__simd/specializations/fixed_size_float_vec.h
```diff
@@ -105,10 +109,20 @@ struct __simd_operations<float, __fixed_size<_Np>, __simd_operations_kind::__fix
...
   }
+
+  [[nodiscard]] _CCCL_HOST_DEVICE_API static _SimdStorage
+  __fma(const _SimdStorage& __lhs, const _SimdStorage& __rhs, const _SimdStorage& __add) noexcept
+  {
+    _CCCL_IF_NOT_CONSTEVAL_DEFAULT
+    {
+      NV_IF_TARGET(NV_PROVIDES_SM_100, (return ::cuda::std::simd::__fma_f32x2(__lhs, __rhs, __add);))
+    }
+    return __base::__fma(__lhs, __rhs, __add);
```
Jacobfaib, line 124: Should these not also be fully qualified? I admit that the double underscore in `__base` invites trouble for any users doing that themselves, but you never know...

libcudacxx/include/cuda/std/__simd/specializations/fp32x2_intrinsics.h
```diff
@@ -26,178 +26,133 @@
...
 #  include <cuda/std/__cccl/prologue.h>
 
 _CCCL_BEGIN_NAMESPACE_CUDA_STD_SIMD
 
-[[nodiscard]] _CCCL_DEVICE_API inline ::float2 __add_f32x2(const ::float2 __lhs, const ::float2 __rhs) noexcept
+[[nodiscard]] _CCCL_DEVICE_API inline ::float2
+__add_f32x2([[maybe_unused]] const ::float2 __lhs, [[maybe_unused]] const ::float2 __rhs) noexcept
 {
   ::float2 __result{};
 #  if _CCCL_HAS_SIMD_F32X2_INTRINSICS()
```
Jacobfaib, line 39: If I'm reading this correctly, we have `_CCCL_HAS_SIMD_F32X2_INTRINSICS()` here, but we are still unable to satisfy this without blackwell? The previous version of the code seems to have been able to handle this, but could be misreading. The diffs are hard to parse here.

libcudacxx/include/cuda/std/__simd/specializations/fixed_size_float_vec.h
fbusato: `__base` is an alias for the base class, so it is not a problem

libcudacxx/include/cuda/std/__simd/specializations/fp32x2_intrinsics.h
fbusato: `F32X2` requires Blackwell. However, the previous code only targeted SM100, missing SM103. This PR fixes it

libcudacxx/include/cuda/std/__simd/math/fma.h
fbusato: the `using` declaration makes it unambiguous here. also, @miscco already checked this aspect.

davebayer (approved):

PR #8954, opened 2026-05-12
Fix `mdspan` `layout_stride` ABI failure in MSVC

### LINKED ISSUE #8953: [BUG]: mdspan layout_stride no_unique_address sizeof check fails on MSVC — [[no_unique_address]] is a no-op for ABI compat
### Type of Bug
Runtime Error

### Component
libcu++

### Describe the bug

`cuda/containers/views/mdspan/layout_stride/no_unique_address.pass.cpp` fails at line 34 (device-side assertion) on MSVC:

```
no_unique_address.pass.cpp:34: block: [0,0,0], thread: [0,0,0]
Assertion `sizeof(Mapping) == expected_size` failed.
```

The test computes `expected_size` assuming `[[no_unique_address]]` shrinks the empty extents/strides members of `layout_stride::mapping`. MSVC treats standard `[[no_unique_address]]` as a no-op for ABI compatibility (only `[[msvc::no_unique_address]]` takes effect), so `sizeof(Mapping)` is larger than expected and the assertion fires.

The sibling test `cuda/containers/views/mdspan/mdspan/no_unique_address.pass.cpp` already carries `// UNSUPPORTED: msvc // See nvbug5272086` for the same reason — the `layout_stride` variant needs the same treatment, and the underlying `layout_stride::mapping` likely needs `_CCCL_NO_UNIQUE_ADDRESS` (or an equivalent macro routing to `[[msvc::no_unique_address]]` under MSVC) to make sizes match across toolchains.

Reproduces on every MSVC libcu++ Test job: CTK12.9/MSVC14.44 and CTK13.2/MSVC14.50, C++20, RTX4090 and H100. Discovered while enabling libcu++ MSVC GPU jobs in #5725, which marks the test `// UNSUPPORTED: msvc` to unblock; this issue tracks both the test and the underlying mapping layout.

### How to Reproduce

1. Check out PR #5725.
2. Build and run `libcudacxx.test.lit` against MSVC (CTK12.9/MSVC14.44 or CTK13.2/MSVC14.50, C++20).

### Expected behavior

`sizeof(layout_stride::mapping<extents>)` on MSVC matches the size on clang/gcc — empty extents/strides members occupy no additional storage — and the test passes across all supported toolchains. Related: nvbug5272086 (sibling `mdspan/mdspan/no_unique_address` test).

### Operating System

Windows Server (CI runner image)

### NVCC version

CTK 12.9 (MSVC 14.44.35207) and CTK 13.2 (MSVC 14.50.35717).

fbusato:
## Description

Fixes https://github.com/NVIDIA/cccl/issues/8953

The `sizeof()` failure is due to ABI mismatch between MSVC/NVCC for double inheritance of the same empty class.

Fix: add an extra level of indirection + specialization for rank=0

### DIFF
--- a/libcudacxx/include/cuda/std/__mdspan/layout_stride.h
+++ b/libcudacxx/include/cuda/std/__mdspan/layout_stride.h
@@ -50,6 +50,25 @@ _CCCL_BEGIN_NAMESPACE_CUDA_STD
 
 namespace __layout_stride_detail
 {
+template <class _Extents, class _StrideArray, bool = (_Extents::rank() == 0)>
+struct _CCCL_DECLSPEC_EMPTY_BASES __mapping_base : __mdspan_ebco<_Extents, _StrideArray>
+{
+  using __base = __mdspan_ebco<_Extents, _StrideArray>;
+  using __base::__base;
+};
+
+template <class _Extents, class _StrideArray>
+struct _CCCL_DECLSPEC_EMPTY_BASES __mapping_base<_Extents, _StrideArray, true> : __mdspan_ebco<_Extents>
+{
+  using __base = __mdspan_ebco<_Extents>;
+  using __base::__base;
+
+  _CCCL_API constexpr __mapping_base(const _Extents& __ext,
+                                     const _StrideArray&) noexcept(is_nothrow_constructible_v<__base, const _Extents&>)
+      : __base(__ext)
+  {}
+};
+
 template <class _StridedLayoutMapping, class _Extents>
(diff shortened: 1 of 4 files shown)


miscco (changes requested):
We have expressedly moved all of mdspan to using empty base class optimization. 

If the issue still appears that is a bug in our implementation

miscco (approved):
I dont have time to properly investigate why this makes a difference, so lets get it into 3.4

PR #8956, opened 2026-05-13
[libcu++] Run host and device testing in parallel

davebayer:
I had an idea that we could execute host test path while asynchronously testing stuff on device. This can lead to some small speedups when testing with `lit`, especially when the tests are a bit larger.

I've also how the return value on device is being returned to avoid WAW conflicts.

### DIFF
--- a/libcudacxx/test/libcudacxx/force_include.h
+++ b/libcudacxx/test/libcudacxx/force_include.h
@@ -54,15 +54,21 @@ __tile__
 int cuda_thread_count = 1;
 int cuda_cluster_size = 1;
 
+__device__ int fake_main_kernel_ret = 0;
+
 #ifdef __CUDACC_TILE__
 __tile_global__
 #else // ^^^ __CUDACC_TILE__ ^^^ / vvv !__CUDACC_TILE__ vvv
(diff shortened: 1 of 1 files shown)


libcudacxx/test/libcudacxx/force_include.h
```diff
@@ -54,15 +54,21 @@ __tile__
 int cuda_thread_count = 1;
 int cuda_cluster_size = 1;
 
+__device__ int fake_main_kernel_ret = 0;
```
davebayer, line 57: @miscco will this break tile?

miscco: unfortunately yes

miscco (changes requested):
I dont think this is a good idea

Individual test time is completely irrelevant due to the fact that our tests are really small and we have a bazillion tests

davebayer:
> I dont think this is a good idea
> 
> Individual test time is completely irrelevant due to the fact that our tests are really small and we have a bazillion tests

Yeah, I've also realized that sometimes we rely on the host part being launched before the device part.. I'll close the issue

PR #8977, opened 2026-05-13
Add `constexpr` to `cuda::sincos` when possible

fbusato:
Address nvc++ warning related to `if consteval` in a non-constant context.


fbusato:
we could also consider backports

davebayer (changes requested):
We don't even have our `cuda::std::sin`/`cuda::std::cos` `constexpr, I would rather keep this non-constexpr for now

fbusato:
my intuition was because `sin/cos` are c++ standard functions, we keep them aligned. Anyway, if we have the suppression for the warning, I can close this PR

fbusato:
btw, would be great to have `sin/cos` constexpr, because `std::simd::sin/cos` also requires it

davebayer:
> btw, would be great to have `sin/cos` constexpr, because `std::simd::sin/cos` also requires it

Yeah sure, but I think we should go the easy way and implement `sincos` by separate `sin` and `cos` during constant evaluation anyway, it will be much simpler to maintain

PR #8979, opened 2026-05-14
Final env-passthrough 2/4

gonidelis:

  - DeviceAdjacentDifference
  - DeviceCopy
  - DevicePartition
  - DeviceRunLengthEncode
  - DeviceFind

the most serious issues have to do with defaults in adjacentdifference and ambiguities they were created.

I chose to start introducing non-env api file tests to check that the non-env APIs work in their minimal form. That is when default args are not explicitly passed. These files will facilitate as ground for extracting the example snippets from non env overloads from doxygen to literalincludes later on.

### DIFF
--- a/cub/test/catch2_test_device_adjacent_difference_api.cu
+++ b/cub/test/catch2_test_device_adjacent_difference_api.cu
@@ -1,0 +1,54 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include "insert_nested_NVTX_range_guard.h"
+
+#include <cub/device/device_adjacent_difference.cuh>
+
+#include <c2h/catch2_test_helper.h>
+
+// Guard: the legacy memory-size query call with all defaults (no explicit difference_op,
+// no explicit stream) must resolve unambiguously to the legacy temp-storage overload
+// when the env passthrough overload is also visible. If the env overload's SFINAE
+// is too loose, this becomes "ambiguous overload" or silently dispatches to env.
+
+C2H_TEST("DeviceAdjacentDifference::SubtractLeftCopy legacy size-query is unambiguous", "[adjacent_difference][device]")
+{
+  int* d_in    = nullptr;
+  int* d_out   = nullptr;
+  size_t bytes = 0;
+  int n        = 0;
+
+  REQUIRE(cudaSuccess == cub::DeviceAdjacentDifference::SubtractLeftCopy(nullptr, bytes, d_in, d_out, n));
+}
+
+C2H_TEST("DeviceAdjacentDifference::SubtractLeft legacy size-query is unambiguous", "[adjacent_difference][device]")
+{
+  int* d_in    = nullptr;
+  size_t bytes = 0;
+  int n        = 0;
+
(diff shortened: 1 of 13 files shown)


cub/test/catch2_test_device_adjacent_difference_api.cu
```diff
@@ -0,0 +1,54 @@
...
+
+C2H_TEST("DeviceAdjacentDifference::SubtractLeftCopy legacy size-query is unambiguous", "[adjacent_difference][device]")
+{
+  int* d_in    = nullptr;
+  int* d_out   = nullptr;
+  size_t bytes = 0;
+  int n        = 0;
+
+  REQUIRE(cudaSuccess == cub::DeviceAdjacentDifference::SubtractLeftCopy(nullptr, bytes, d_in, d_out, n));
+}
```
bernhardmgruber, line 23: Important: Those tests don't look like API examples. So they should go into the files `cub/test/catch2_test_device_adjacent_difference_subtract_*.cu`

cub/test/catch2_test_device_copy_api.cu
```diff
@@ -0,0 +1,32 @@
...
+{
+  // DeviceCopy::Batched takes iterator-of-iterators for input/output ranges.
+  int** in                        = nullptr;
+  int** out                       = nullptr;
+  size_t* sizes                   = nullptr;
+  size_t bytes                    = 0;
+  ::cuda::std::int64_t num_ranges = 0;
+
+  REQUIRE(cudaSuccess == cub::DeviceCopy::Batched(nullptr, bytes, in, out, sizes, num_ranges));
+}
```
bernhardmgruber, line 27: Same here, this is not an API example test.

Applies a few more times below.

cub/test/catch2_test_device_partition_api.cu
```diff
@@ -0,0 +1,71 @@
...
+
+#include <c2h/catch2_test_helper.h>
+
+struct always_true_t
+{
+  __host__ __device__ bool operator()(int) const
+  {
+    return true;
+  }
+};
```
bernhardmgruber: Important: Please use ` ::cuda::always_true`

cub/test/catch2_test_device_adjacent_difference_api.cu
gonidelis: 1. doesn't have to be an example to go under the `_api.cu` test. It's literally what the name suggest, an API test, not an example (which the file name doesn't even suggest).

2. it's a good starting point to have for moving old docs examples into literalincludes

please let's keep it as is

bernhardmgruber: > 1. doesn't have to be an example to go under the `_api.cu` test. It's literally what the name suggest, an API test, not an example (which the file name doesn't even suggest).

This is contrary to my understanding. `_api.cu` contain only API examples which are referenced from the documentation. They do not contain any functional tests.

Functional tests are always in files like `catch2_test_device_algorithmname.cu` (with some variations). You also added tests named `catch2_test_device_algorithmname_env.cu`, which test whether the handling of the execution environment works correctly, with little or no focus on algorithmic correctness.

Please elaborate why we need to break the convention now, or where it has been broken before.

gonidelis:
critical ambiguity resolved in last commit

cub/test/catch2_test_device_adjacent_difference_api.cu
bernhardmgruber: Let's continue here: #9078

bernhardmgruber (approved):

PR #8981, opened 2026-05-14
Final env-passthrough 3/4

gonidelis:

  - DeviceFor
  - DeviceTransform
  - DeviceSelect
  - DeviceSegmentedReduce
  - DeviceSegmentedRadix sort

### DIFF
--- a/cub/cub/device/device_segmented_reduce.cuh
+++ b/cub/cub/device/device_segmented_reduce.cuh
@@ -1013,6 +1013,8 @@ public:
       int segment_size,
       cudaStream_t stream = nullptr)
   {
+    static_assert(!::cuda::std::is_same_v<InputIteratorT, void*>,
+                  "InputIteratorT must be a real iterator; void* has no iterator_traits::value_type.");
     _CCCL_NVTX_RANGE_SCOPE_IF(d_temp_storage, "cub::DeviceSegmentedReduce::Sum");
     using init_t = detail::non_void_value_t<OutputIteratorT, detail::it_value_t<InputIteratorT>>;
     return fixed_size_impl(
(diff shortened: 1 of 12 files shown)


cub/cub/device/device_segmented_reduce.cuh
```diff
@@ -1003,9 +1003,7 @@ public:
   //!   @rst
   //!   **[optional]** CUDA stream to launch kernels within. Default is stream\ :sub:`0`.
   //!   @endrst
-  template <typename InputIteratorT,
-            typename OutputIteratorT,
-            ::cuda::std::enable_if_t<!::cuda::std::is_same_v<InputIteratorT, void*>, int> = 0>
```
fbusato, line 1008: I'm not sure I understand the change here. What happens now if `InputIteratorT` is `void*`? any `static_assert` in `detail::` traits?

gonidelis: We've discussed with Bernhard that I keep only the SFINAEs if they are actually disambiguating the new env ovreloads from their old non-env counterparts so that we keep the signatures at their minimal form. This SFINAE was just offering a better diagnostic so I removed it (and just added it back as a `static_assert` in the function body so it doesn't bloat the signature).

cub/cub/device/device_segmented_reduce.cuh
```diff
@@ -1015,6 +1013,8 @@ public:
       int segment_size,
       cudaStream_t stream = nullptr)
   {
+    static_assert(!::cuda::std::is_same_v<InputIteratorT, void*>,
+                  "InputIteratorT must be a real iterator; void* has no iterator_traits::value_type.");
```
bernhardmgruber, line 1017: Remark: Hmm. I am not sure how I feel about this. We should not need this, since a user should never get into this overload with a `void*` input iterator. But then, it doesn't hurt either.

I would personally remove this, but no action needed.

bernhardmgruber (commented):
We may need to revisit this after we decided on #9078

bernhardmgruber (approved):

cub/cub/device/device_segmented_reduce.cuh
gonidelis: @fbusato  all good with that? can i resolve?

PR #8982, opened 2026-05-14
Add env DeviceTopK without temp_storage args

### LINKED ISSUE #8969: `cub::DeviceTopK` does not have env-overloads handling memory allocation
I just saw that cub::DeviceTopK does not yet have env-overloads without
```c++
    void* d_temp_storage,
    size_t& temp_storage_bytes,
```
that handle the allocation using a memory resource from the environment.

The existing overloads do take environments instead of streams, so more options in addition to a stream can be specified, but they still rely on user-side allocation handling.

Please add env-overloads that use a memory resource from the env (or the default one) to handle allocation.

gonidelis:
fixes https://github.com/NVIDIA/cccl/issues/8969


cub/cub/device/device_topk.cuh
```diff
@@ -433,6 +544,123 @@ struct DeviceTopK
...
+            typename KeyOutputIteratorT,
+            typename ValueInputIteratorT,
+            typename ValueOutputIteratorT,
+            typename NumItemsT,
+            typename NumOutItemsT,
+            typename DecomposerT,
+            typename EnvT = ::cuda::std::execution::env<>>
+  [[nodiscard]] CUB_RUNTIME_FUNCTION static //
+    ::cuda::std::enable_if_t<detail::radix::is_valid_decomposer<detail::it_value_t<KeyInputIteratorT>, DecomposerT>,
+                             cudaError_t>
```
bernhardmgruber: Suggestion: Please put requirements (`enable_if`) into the template head instead of using the return type.

cub/cub/device/device_topk.cuh
```diff
@@ -681,35 +1019,248 @@ struct DeviceTopK
...
```
bernhardmgruber: Same here.

cub/test/catch2_test_device_topk_env.cu
```diff
@@ -184,4 +187,165 @@ C2H_TEST("DeviceTopK::MinPairs can be tuned", "[topk][device]", block_sizes)
...
+{
+  using T = c2h::get<0, TestType>;
+
+  const int num_items = 256;
+  thrust::host_vector<T> h_in(num_items);
+  for (int i = 0; i < num_items; ++i)
+  {
+    h_in[i] = static_cast<T>((i * 1664525 + 1013904223) % 251);
+  }
+  thrust::device_vector<T> d_in    = h_in;
```
bernhardmgruber: Important: in ordinary unit tests, please use `c2h::host_vector` etc.

cub/test/catch2_test_device_topk_env.cu
```diff
@@ -184,4 +187,165 @@ C2H_TEST("DeviceTopK::MinPairs can be tuned", "[topk][device]", block_sizes)
...
+    std::sort(sorted.begin(), sorted.end());
+  }
+  sorted.resize(static_cast<size_t>(k));
+  return sorted;
+}
+} // namespace
+
+using topk_element_types = c2h::type_list<int8_t, int16_t, int32_t, uint32_t, int64_t, float, double>;
+
+C2H_TEST("DeviceTopK::MaxKeys env-alloc returns correct top K", "[topk][env]", topk_element_types)
```
bernhardmgruber: Q: Why do we need to cover different element types when just testing whether the env-overload works correctly? We should not test the topk implementation itself, those tests already exist elsewhere. The current state adds a lot of compile-time for little gain IMO.

cub/test/catch2_test_device_topk_env_api.cu
```diff
@@ -0,0 +1,332 @@
...
+  auto env = cuda::std::execution::env{
+    cuda::execution::require(cuda::execution::determinism::not_guaranteed, cuda::execution::output_ordering::unsorted),
+    stream_ref};
+
+  auto error = cub::DeviceTopK::MaxKeys(d_in.begin(), d_out.begin(), static_cast<int>(d_in.size()), k, env);
+  if (error != cudaSuccess)
+  {
+    std::cerr << "cub::DeviceTopK::MaxKeys failed with status: " << error << '\n';
+  }
+  // example-end topk-max-keys-env
```
bernhardmgruber, line 65: Important: we should list at least one possible expected outcome in the documentation. Like:

```suggestion
  auto error = cub::DeviceTopK::MaxKeys(d_in.begin(), d_out.begin(), static_cast<int>(d_in.size()), k, env);
  if (error != cudaSuccess)
  {
    std::cerr << "cub::DeviceTopK::MaxKeys failed with status: " << error << '\n';
  }
  thrust::device_vector<int> expected{9, 8, 7}; // possibly in different order
  // example-end topk-max-keys-env
```

cub/cub/device/device_topk.cuh
gonidelis: I just followed the pattern from the pre-existing overloads.

https://github.com/gonidelis/cccl/blob/b0a851953a4f087fb3eba9d4f80a315c7d6c99a6/cub/cub/device/device_topk.cuh#L1834-L1837

I know it's uncanny but are you sure you want to move it?

bernhardmgruber: I am aware. It's generally a better practice to use the template head instead of the return type. But it's just a suggestion. If you don't want to, it will not block the PR.

bernhardmgruber: You may also file a separate PR (or ask claude to do it) and rewrite the old overloads.

cub/test/catch2_test_device_topk_env.cu
```diff
@@ -184,4 +186,154 @@ C2H_TEST("DeviceTopK::MinPairs can be tuned", "[topk][device]", block_sizes)
...
+c2h::host_vector<int> sorted_top_k(const c2h::host_vector<int>& h_in, int k, bool largest)
+{
+  auto sorted = h_in;
+  if (largest)
+  {
+    std::sort(sorted.begin(), sorted.end(), cuda::std::greater<int>{});
+  }
+  else
+  {
+    std::sort(sorted.begin(), sorted.end());
```
bernhardmgruber: Suggestion: here is a chance to flex with some algorithm knowledge:
```suggestion
    std::partial_sort(sorted.begin(), sorted.begin() + k, sorted.end(), cuda::std::greater<int>{});
  }
  else
  {
    std::partial_sort(sorted.begin(), sorted.begin() + k, sorted.end());
```

cub/test/catch2_test_device_topk_env.cu
```diff
@@ -184,4 +186,154 @@ C2H_TEST("DeviceTopK::MinPairs can be tuned", "[topk][device]", block_sizes)
...
+
+C2H_TEST("DeviceTopK::MaxKeys env-alloc returns correct top K", "[topk][env]")
+{
+  const int num_items = 256;
+  c2h::host_vector<int> h_in(num_items);
+  for (int i = 0; i < num_items; ++i)
+  {
+    h_in[i] = static_cast<int>((i * 1664525 + 1013904223) % 251);
+  }
+  c2h::device_vector<int> d_in    = h_in;
```
bernhardmgruber: Important: I have found those magic numbers in a few unrelated tests, so I am a bit surprised why those are needed here. Please explain if there is a good reason.

Otherwise, I would suggest just using the standard `gen` function used everywhere:
```suggestion
  c2h::device_vector<int> d_in(num_items);
  c2h::gen(C2H_SEED(1), d_in);
  c2h::host_vector<int> h_in    = d_in;
```
Applies to other tests as well.

cub/test/catch2_test_device_topk_env.cu
```diff
@@ -184,4 +186,154 @@ C2H_TEST("DeviceTopK::MinPairs can be tuned", "[topk][device]", block_sizes)
...
+  c2h::device_vector<int> d_out_k = c2h::device_vector<int>(8);
+
+  auto env = topk_requirements();
+  REQUIRE(cudaSuccess == cub::DeviceTopK::MaxKeys(d_in.begin(), d_out_k.begin(), num_items, 8, env));
+
+  c2h::host_vector<int> h_out = d_out_k;
+  std::sort(h_out.begin(), h_out.end(), cuda::std::greater<int>{});
```
bernhardmgruber, line 220: Suggestion: use `partial_sort`.

cub/test/catch2_test_device_topk_env_api.cu
```diff
@@ -0,0 +1,332 @@
...
+  // example-begin topk-max-keys-env
+  auto d_in  = thrust::device_vector<int>{8, 6, 7, 5, 3, 0, 9, 1, 4, 2};
+  auto d_out = thrust::device_vector<int>(3);
+  int k      = 3;
+
+  cuda::stream stream{cuda::devices[0]};
+  cuda::stream_ref stream_ref{stream};
+  auto env = cuda::std::execution::env{
+    cuda::execution::require(cuda::execution::determinism::not_guaranteed, cuda::execution::output_ordering::unsorted),
+    stream_ref};
```
bernhardmgruber, line 57: Mild suggestion: We may want to force a line break here:
```suggestion
  auto env = cuda::std::execution::env{
    cuda::execution::require(cuda::execution::determinism::not_guaranteed, //
    cuda::execution::output_ordering::unsorted),
    stream_ref};
```

bernhardmgruber (commented):
Mostly LGTM, a few more suggestions:

cub/cub/device/device_topk.cuh
gonidelis: no i changed it, s'all good

bernhardmgruber (approved):

PR #8984, opened 2026-05-14
Add sm120 Warpspeed scan regression test

bdice:
## Description

Adds a minimal regression test for the sm120 NVCC Warpspeed scan codegen issue guarded by #8922.

The test exercises a one-element `DeviceScan::ExclusiveScan<int*, int*>`, which fails under compute-sanitizer before #8922 when Warpspeed scan is selected on sm120 with NVCC 13.2, and passes once #8922 disables that path for NVCC < 13.4.

## Checklist


bdice:
Closing this for now. I could not produce a functional regression test that fails without compute-sanitizer: the reduced scan cases I tried pass in normal execution before #8922 and only fail under compute-sanitizer. Given #8528, it is also unclear whether the compute-sanitizer reports are actionable failures or possible false positives/noise around the underlying compiler issue. We should avoid landing a test that only asserts sanitizer behavior until we have a normal runtime reproducer or a clearer sanitizer-backed failure mode.

PR #8437, opened 2026-04-15
[cccl.c] Introduce minimal compilation support for host freestanding mode

shwina:
## Description

Closes #8401 

This PR introduces the minimal infrastructure for compiling and testing CCCL in "host freestanding" mode. The infrastructure being introduced suffices to compile the following code using the included JIT compiler:

```c++
std::string source = R"(
__global__ void device(int* ptr) {
  *ptr = 42;
}
void host(int* ptr) {
  device<<<1, 1>>>(ptr);
}
)";
```

A CI job has been added which builds the host JIT infrastructure for a specific combination of arch/compiler/CUDA. Eventually, we will want the entire matrix to build with host JIT, but during initial iteration this is sufficient:

Screenshot 2026-04-24 at 7 54 09 AM

Note that an important consideration/blocker is that nvfatbin (a dependency for host JIT) is available only on CUDA 12.4+.

## Checklist

- [ ] New or existing tests cover these changes.
- [ ] The documentation is up to date with these changes.

PR #8985, opened 2026-05-14
CCCL C v2

shwina:
## Overview

This PR introduces `cccl.c.parallel.v2`, a reimplementation of `cccl.c.parallel` with an identical C API, built on the host-JIT infrastructure that landed in #8437.

v1 and v2 ship side-by-side so we can iterate on v2 without disrupting downstream consumers. The eventual plan is to retire v1 once v2 reaches parity in CI coverage.

## Scope of changes

**New project: `c/parallel.v2/`** — mirrors the layout of `c/parallel/`:
- `include/cccl/c/*.h` — same public headers as v1, identical signatures. The only divergences are in the `build_result_t` structs (different internal bookkeeping fields — `jit_compiler*` instead of `CUlibrary`/`CUkernel`s, etc.).
- `src/*.cu` — Implementation of algorithms. Each one is a thin shell that invokes `CubCall` (described below) and then assigns the result into the `build_result_t`.
- `src/hostjit/` — the JIT infrastructure brought over from #8484.
- `src/util/build_utils.h` — shared utilities
- `test/freestanding/` — v2-specific JIT-infrastructure tests (compile path, header availability, etc.).

**Shared test tree.** Both backends compile the same `test_X.cpp` sources under `c/parallel/test/`; v2's `CMakeLists.txt` globs them and compiles with `-DCCCL_C_PARALLEL_V2=1`, producing distinct test binaries (`cccl.c.parallel.test.X` vs `cccl.c.parallel.v2.test.X`).

## How v2 works

Each algorithm in v2 routes through one builder, `CubCall`, which **generates and JIT-compiles a C++ host wrapper around a CUB device function** given a list of typed argument "tags." The caller describes what they want symbolically; `CubCall` produces the source string, links in any user-provided bitcode for ops/iterators, runs the JIT compiler, and hands back a function pointer.

```cpp
auto result = CubCall::from("cub/device/device_X.cuh")  // header to include
                .run("cub::DeviceX::Y")                 // CUB function to call
                .name("cccl_jit_X")                     // exported symbol
                .with(/* arg tags */)                   // describe the args
                .compile(cc_major, cc_minor, config, ctk_root, cccl_include_path);

// result.compiler  -> opaque JITCompiler* (caller owns)
// result.fn_ptr    -> function pointer to the exported wrapper
// result.cubin     -> for SASS inspection
```

### `CubCall::with(...)` vocabulary

Each tag controls both the parameter emitted in the JIT'd function signature and the value emitted at the CUB call site.

| Tag | JIT-function parameter | CUB call-site value |
| --- | --- | --- |
| `temp_storage` | `void* d_temp_storage` | `d_temp_storage` |
| `temp_bytes` | `size_t* temp_storage_bytes` | `*temp_storage_bytes` |
| `num_items` (also `num_segments`, `num_needles`, `num_haystack`) | `unsigned long long <name>` | `(unsigned long long)<name>` |
| `stream` | `void* stream` | `(cudaStream_t)stream` |
| `env_stream` | `void* stream` | `::cuda::std::execution::env{::cuda::stream_ref{(cudaStream_t)stream}}` |
| `in(iter)` | `void* d_in_N` | iterator-typed `in_N` (setup_code unpacks it) |
| `out(iter)` | `void* d_out_N` | iterator-typed `out_N` |
| `op` (`cccl_op_t` directly) | `void* op_N_state` | binary functor `op_N` invoking user bitcode |
| `cmp(op)` | `void* cmp_N_state` | comparison functor returning `bool` |
| `unary_op(op, in_t, out_t)` | `void* op_N_state` | unary functor `op_N` |
| `for_each_op(op)` | `void* op_N_state` | functor adapting user `op(T*)` to CUB's `op(T&)` |
| `cccl_value_t` | `void* val_N_ptr` | accumulator-typed value copied from the ptr |
| `future_val(t)` | `void* future_N_param` | `cub::FutureValue<accum_t>(...)` |
| `typed_scalar(t, name)` | `void* <name>_ptr` | by-value POD copied from the ptr |
| `force_accum_type(t)` | *(none)* | *(no code emitted; overrides accum-type resolution)* |

### What `CubCall` produces under the hood

1. Walks the args, figures out the accumulator type, and builds a preamble (extern decls, functor structs, storage-type aliases for unknown user types).
2. Emits a self-contained CUDA source like:

    ```cpp
    #include <cub/device/device_X.cuh>
    /* preamble: extern decls, functor structs, type aliases */

    extern "C" EXPORT int cccl_jit_X(/* params from tags */)
    {
        /* setup_code: build iterators, copy state bytes, etc. */
        cudaError_t err = cub::DeviceX::Y(/* cub_args from tags */);
        return (int)err;
    }
    ```

3. Sets up the `CompilerConfig` (CCCL include paths, CTK root, extra cub/thrust includes, flags from `cccl_build_config`).
4. Spins up a `BitcodeCollector`, walks the args again, and registers each user op/iterator's bitcode with it so the JIT linker resolves the `extern "C" __device__` symbols.
5. Hands the source to `JITCompiler` (Clang-based), gets the cubin and the function pointer for the exported symbol back.

Each algorithm's `_build_ex` collapses to ~5 lines of `CubCall` configuration plus a few `build_ptr->X` field assignments. All the source-generation, bitcode wiring, JIT-config plumbing, and JIT-compile orchestration lives once in `CubCall` — never in the per-algorithm file.

## Testing and CI

`v1` and `v2` shared a unified test tree at `c/parallel/test/`. The v2 build globs the `test_*.cpp` files from there and compiles them with `-DCCCL_C_PARALLEL_V2=1`. The CMake setup produces two distinct binaries per test file (`cccl.c.parallel.test.X` and `cccl.c.parallel.v2.test.X`) so both are runnable through `ctest`.

**JIT-infrastructure coverage** lives in `c/parallel.v2/test/freestanding/` — small Catch2 binaries that exercise the JIT compile path directly, separate from any C-parallel algorithm. These are v2-specific.

**CI** is intentionally limited at this stage. Two rows in `ci/matrix.yaml` (mirrored into `pull_request_lite`):

- `cccl_c_parallel_v2`: CTK 13.X / gcc13 / RTX 2080. Builds and runs all 13 ctest binaries plus `freestanding/`.

This is a single point of coverage per project. The goal of this PR's CI is to keep v2 green during day-to-day development; broader coverage is deferred to the follow-up matrix expansion described below.

## Next steps

**CI matrix expansion** (the largest piece of remaining work). Right now `pull_request` runs v2 against exactly one configuration; the goal is parity with v1's existing coverage.

**Outstanding review questions** (Nader's review comments tagged `Question:`):

- **Nondeterministic reduce** (`reduce.cu`). `cccl_device_reduce_nondeterministic` currently forwards to the deterministic path; v1 had a real dispatch into `cub::detail::reduce_nondeterministic::dispatch`. v2 can plumb this through `cub::DeviceReduce::Reduce`'s env-based overload with a `not_guaranteed` determinism requirement — same shape as the env trick we already use for `binary_search`.
- **`__half` re-enable** (`test_radix_sort.cpp`, `test_histogram.cpp`). These tests currently skip fp16 in v2 builds. Pending an audit of v2's fp16 paths.
- **Wheel naming convention** (`get_wheel_artifact_name.sh`). v2 uses a `-v2` suffix to avoid clobbering; longer-term we may want a more deliberate naming convention as we approach v1 retirement.

**Further DRY across algorithm files**. Each `_build_ex` still has identical boilerplate around the `CubCall` invocation (path parsing, MergedBuildConfig construction, exception boundary, common `build_result_t` field setup). A `BuildContext` helper plus a `finalize_jit_build()` template plus an `api_boundary()` exception-wrapper could collapse another ~150 lines across the twelve files. Deferred to keep this PR's scope contained.

**Eventual v1 retirement**. The medium-term plan, contingent on CI parity and a soak period: remove `c/parallel/`, drop the `CCCL_C_PARALLEL_V2` macro from the test tree, rename `c/parallel.v2/` to `c/parallel/`, and drop the v1/v2 split in the Python bindings. None of that needs to happen in this PR; the layout chosen here makes the eventual move mostly mechanical.

## Checklist


shwina:
/ok to test https://github.com/NVIDIA/cccl/pull/8985/changes/c45b7805029fe9cd4bc3a1e510d378c685d0541c

shwina:
/ok to test https://github.com/NVIDIA/cccl/pull/8985/changes/3e4ed01e762e2b785f2be68ffac31361cc50c7b7

shwina:
/ok to test https://github.com/NVIDIA/cccl/pull/8985/commits/299e41133df29d203188edc4dddcd0f69cf0297d

shwina:
/ok to test https://github.com/NVIDIA/cccl/pull/8985/changes/f844126ebfd99deb5d05a99ad2ecd5c30e2b1354

shwina:
/ok to test https://github.com/NVIDIA/cccl/pull/8985/changes/db8ea0c84fc307bae6b423eda09a3ea3f121caf8

shwina:
/ok to test https://github.com/NVIDIA/cccl/pull/8985/commits/0c377ebb4439f75ad8a33ee2242831d37a6ce706

shwina:
/ok to test https://github.com/NVIDIA/cccl/pull/8985/commits/1b6cff30150fe2a33d2f96d766c5a9f7e3d57a1a

shwina:
/ok to test https://github.com/NVIDIA/cccl/pull/8985/changes/db76019d85f397c101bb74c11e26a17b2d2e1ccb

shwina:
/ok to test https://github.com/NVIDIA/cccl/pull/8985/changes/8f9b8a49082afd0c8d73bf2d2570334520f18b3b

c/parallel.v2/src/hostjit/include/hostjit/cuda_minimal/stubs/cstddef
```diff
@@ -3,7 +3,8 @@
 #ifndef _HOSTJIT_CSTDDEF
 #define _HOSTJIT_CSTDDEF
 
-#include <stddef.h>
+#include "stddef.h"
```
miscco: Why are we moving to relative includes?

c/parallel.v2/src/hostjit/include/hostjit/cuda_minimal/stubs/string.h
```diff
@@ -0,0 +1,27 @@
...
+int memcmp(const void*, const void*, size_t);
+void* memmove(void*, const void*, size_t);
+void* memchr(const void*, int, size_t);
+size_t strlen(const char*);
+int strcmp(const char*, const char*);
+int strncmp(const char*, const char*, size_t);
+char* strcpy(char*, const char*);
+char* strncpy(char*, const char*, size_t);
+char* strchr(const char*, int);
+char* strrchr(const char*, int);
```
miscco: I did provide implementations for those functions, why drop them?

shwina:
/ok to test https://github.com/NVIDIA/cccl/pull/8985/commits/5db72b1f9fc6f5bf42dc4d07689e31b8b3218064

c/parallel.v2/src/hostjit/include/hostjit/cuda_minimal/stubs/cstddef
shwina: Fixed.

c/parallel.v2/src/hostjit/include/hostjit/cuda_minimal/stubs/string.h
shwina: Agent cherry picked this file from `freestanding` branch and started there. Fixed.

shwina:
/ok to test https://github.com/NVIDIA/cccl/pull/8985/changes/7a565b33907f2d069f3f9e0c780d2334e8646997

shwina:
/ok to test https://github.com/NVIDIA/cccl/pull/8985/changes/696dabf3c3b3e188d3c560c1dbd46920dfd78872

shwina:
/ok to test https://github.com/NVIDIA/cccl/pull/8985/commits/98320a82fda4d2918908181aa1ee605e377ee278

shwina:
/ok to test https://github.com/NVIDIA/cccl/pull/8985/commits/266fa7673ba20b99e62468479dae43cac2813c34

c/parallel.v2/src/util/runtime_policy_readme.md
bernhardmgruber, line 1: I removed this file from CCCL.C a while ago, so I suspect the work in this PR is not up-2-date. Please check for any recent modifications in CCCL.C and whether those were carried forward to CCCL.C v2.

c/parallel.v2/src/util/indirect_arg.h
```diff
@@ -0,0 +1,119 @@
...
+//
+// We cannot use `nullptr` because the pointer is passed to a driver API that
+// requires the size of empty arguments to be 1 (not 0), meaning it will attempt
+// to copy a byte from the address.
+//
+// Initially, we initialized `ptr` to `this`, but this triggered the dangling
+// pointer warning in GCC. To avoid this, we use a global variable instead.
+static inline char _global_storage = 0;
+
+struct indirect_arg_t
```
bernhardmgruber: Q: Why do we still need `indirect_arg_t`? This was required for type erasure so we can pre-compile the host dispatch of CUB, which should no longer be needed with a JIT compiler.

c/parallel.v2/src/util/tuning.h
```diff
@@ -0,0 +1,13 @@
...
+// under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES.
+//
+//===----------------------------------------------------------------------===//
+
+#pragma once
+
+int nominal_4b_items_to_items(int nominal_4b_items_per_thread, int key_size);
```
bernhardmgruber: Important: Please just use this function from CUB's `util_math.cuh` instead of duplicating it.

c/parallel.v2/src/for.cu
```diff
@@ -0,0 +1,369 @@
...
+  src += "#include <cuda_runtime.h>\n";
+  src += "#include <cuda_fp16.h>\n";
+  src += "#include <cub/agent/agent_for.cuh>\n";
+  src += "#include <climits>\n\n";
+
+  src += "#ifdef _WIN32\n"
+         "#define EXPORT __declspec(dllexport)\n"
+         "#else\n"
+         "#define EXPORT __attribute__((visibility(\"default\")))\n"
+         "#endif\n\n";
```
bernhardmgruber: Important: Please use raw string literals for readability. Applies everywhere in this PR.

c/parallel.v2/src/for.cu
```diff
@@ -0,0 +1,369 @@
...
+  // Generate source
+  std::string cuda_source = make_for_source(d_data, op);
+  if (const char* dump_path = std::getenv("FOR_DUMP_SOURCE"))
+  {
+    std::ofstream f(dump_path);
+    f << cuda_source;
+  }
+
+  // Compile
+  auto* compiler = new JITCompiler(jit_config);
```
bernhardmgruber: Critical: Please use a `unique_ptr` to manage memory resources.

c/parallel.v2/src/merge_sort.cu
```diff
@@ -0,0 +1,242 @@
...
+  const char* thrust_path,
+  const char* libcudacxx_path,
+  const char* ctk_path,
+  cccl_build_config* config)
+try
+{
+  if (d_out_keys.type == CCCL_ITERATOR || d_out_items.type == CCCL_ITERATOR)
+  {
+    fprintf(stderr, "\nERROR in cccl_device_merge_sort_build(): merge sort output cannot be an iterator\n");
+    return CUDA_ERROR_UNKNOWN;
```
bernhardmgruber, line 53: Question: why? CUB supports this. Where is this limitation coming from?

shwina:
/ok to test https://github.com/NVIDIA/cccl/pull/8985/commits/f777f3ee975cac7a449498d2c61442410cbeb7a4

c/parallel.v2/src/merge_sort.cu
shwina: I'd recommend unblocking this in a follow up, as `v1` does not support this (for whatever reason). If we support it in `v2` we'll have to add new tests etc.,

c/parallel.v2/src/for.cu
shwina: Done.

c/parallel.v2/src/for.cu
shwina: Done.

c/parallel.v2/src/util/tuning.h
shwina: I just removed this as it was unused.

c/parallel.v2/src/util/indirect_arg.h
shwina: Removed.

c/parallel.v2/src/util/runtime_policy_readme.md
shwina: Removed.

c/parallel.v2/src/merge_sort.cu
bernhardmgruber: I see. Fine for me then!

c/parallel.v2/include/cccl/c/reduce.h
```diff
@@ -0,0 +1,91 @@
...
+
+CCCL_C_EXTERN_C_BEGIN
+
+typedef struct cccl_device_reduce_build_result_t
+{
+  int cc;
+  void* cubin;
+  size_t cubin_size;
+  void* jit_compiler; // hostjit::JITCompiler*
+  void* reduce_fn; // Function pointer: int(*)(void*, size_t*, void*, void*, unsigned long long, void*)
```
NaderAlAwar: Nit: the comment is missing an extra `void*` at the end

c/parallel.v2/include/cccl/c/reduce.h
```diff
@@ -0,0 +1,91 @@
...
+// Part of CUDA Experimental in CUDA Core Compute Libraries,
+// under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-FileCopyrightText: Copyright (c) 2024 NVIDIA CORPORATION & AFFILIATES.
+//
+//===----------------------------------------------------------------------===//
+
+#pragma once
+
```
NaderAlAwar, line 13: Suggestion: add `// NOLINTBEGIN(modernize-use-using)` here so you don't run into clang-tidy issues, applies to other headers as well

c/parallel.v2/src/reduce.cu
```diff
@@ -0,0 +1,258 @@
...
+CUresult cccl_device_reduce(
+  cccl_device_reduce_build_result_t build,
+  void* d_temp_storage,
+  size_t* temp_storage_bytes,
+  cccl_iterator_t d_in,
+  cccl_iterator_t d_out,
+  uint64_t num_items,
+  cccl_op_t op,
+  cccl_value_t init,
+  CUstream /*stream*/)
```
NaderAlAwar: Critical: this stream should not be ignored

c/parallel.v2/src/segmented_reduce.cu
```diff
@@ -0,0 +1,178 @@
...
+  void* d_temp_storage,
+  size_t* temp_storage_bytes,
+  cccl_iterator_t d_in,
+  cccl_iterator_t d_out,
+  uint64_t num_segments,
+  cccl_iterator_t start_offset,
+  cccl_iterator_t end_offset,
+  cccl_op_t op,
+  cccl_value_t init,
+  CUstream /*stream*/)
```
NaderAlAwar: Critical: this stream should not be ignored

c/parallel.v2/src/binary_search.cu
```diff
@@ -0,0 +1,353 @@
...
+
+CUresult cccl_device_binary_search(
+  cccl_device_binary_search_build_result_t build,
+  cccl_iterator_t d_data,
+  uint64_t num_items,
+  cccl_iterator_t d_values,
+  uint64_t num_values,
+  cccl_iterator_t d_out,
+  cccl_op_t op,
+  CUstream /*stream*/)
```
NaderAlAwar: Critical: this stream should not be ignored

c/parallel.v2/src/for.cu
```diff
@@ -0,0 +1,376 @@
...
+  return CUDA_SUCCESS;
+}
+catch (const std::exception& exc)
+{
+  fprintf(stderr, "\nEXCEPTION in cccl_device_for_build(): %s\n", exc.what());
+  return CUDA_ERROR_UNKNOWN;
+}
+
+CUresult cccl_device_for(
+  cccl_device_for_build_result_t build, cccl_iterator_t d_data, uint64_t num_items, cccl_op_t op, CUstream /*stream*/)
```
NaderAlAwar: Critical: this stream should not be ignored

c/parallel.v2/src/merge_sort.cu
```diff
@@ -0,0 +1,242 @@
...
+    {
+      return CUDA_ERROR_INVALID_VALUE;
+    }
+
+    int status;
+    // Dispatch to the correct function arity based on whether the current call
+    // has items.  The build function compiles either SortKeysCopy (7-arg) or
+    // SortPairsCopy (9-arg); both the build and the run must agree on which
+    // variant is being used (null items → keys, non-null → pairs).
+    const bool has_items = !(d_in_items.type == CCCL_POINTER && d_in_items.state == nullptr);
```
NaderAlAwar: Important: instead of re-checking whether it has items or not in the run function, store this boolean in the build struct

c/parallel.v2/src/reduce.cu
```diff
@@ -0,0 +1,258 @@
...
+    init,
+    determinism,
+    cc_major,
+    cc_minor,
+    cub_path,
+    thrust_path,
+    libcudacxx_path,
+    ctk_path,
+    nullptr);
+}
```
NaderAlAwar, line 167: Question: can we also support nondeterministic reduce?

c/parallel.v2/src/util/types.h
```diff
@@ -0,0 +1,109 @@
...
+//
+// Part of CUDA Experimental in CUDA C++ Core Libraries,
+// under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-FileCopyrightText: Copyright (c) 2024 NVIDIA CORPORATION & AFFILIATES.
+//
+//===----------------------------------------------------------------------===//
+
+#pragma once
```
NaderAlAwar: Suggestion: the contents of this file are unused so it can be deleted

c/parallel.v2/src/util/errors.h
```diff
@@ -0,0 +1,17 @@
...
+//
+// Part of CUDA Experimental in CUDA C++ Core Libraries,
+// under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-FileCopyrightText: Copyright (c) 2024 NVIDIA CORPORATION & AFFILIATES.
+//
+//===----------------------------------------------------------------------===//
+
+#pragma once
```
NaderAlAwar: Suggestion: the contents of this file are unused so it can be deleted

c/parallel.v2/src/util/errors.cpp
```diff
@@ -0,0 +1,31 @@
...
+//
+// Part of CUDA Experimental in CUDA C++ Core Libraries,
+// under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-FileCopyrightText: Copyright (c) 2024 NVIDIA CORPORATION & AFFILIATES.
+//
+//===----------------------------------------------------------------------===//
+
+#include "errors.h"
```
NaderAlAwar: Suggestion: the contents of this file are unused so it can be deleted

c/parallel.v2/src/binary_search.cu
```diff
@@ -0,0 +1,353 @@
...
+#include <hostjit/jit_compiler.hpp>
+#include <util/build_utils.h>
+
+using namespace hostjit;
+using namespace hostjit::codegen;
+
+// d_data_state, num_items, d_values_state, num_values, d_out_state, op_state
+using binary_search_fn_t = int (*)(void*, unsigned long long, void*, unsigned long long, void*, void*);
+
+static std::string make_binary_search_source(
```
NaderAlAwar: Question: why not use CubCall? Same applies for for.cu, radix_sort.cu, segmented_sort.cu, and histogram.cu

c/parallel.v2/test/test_scan.cpp
```diff
@@ -0,0 +1,815 @@
...
+  value_t<T> init{T{42}};
+
+  // Test _ex version with custom build configuration
+  cccl_build_config config;
+  const char* extra_flags[]      = {"-DTEST_IDENTITY_ENABLED"};
+  const char* extra_dirs[]       = {TEST_INCLUDE_PATH};
+  config.extra_compile_flags     = extra_flags;
+  config.num_extra_compile_flags = 1;
+  config.extra_include_dirs      = extra_dirs;
+  config.num_extra_include_dirs  = 1;
```
NaderAlAwar: Critical: this is not initialized properly, specifically `enable_pch` and `verbose` are not initialized. The reduce tests initialize this properly in struct reduce_build_ex, we should do the same here.

c/parallel.v2/test/test_transform.cpp
```diff
@@ -0,0 +1,824 @@
...
+  pointer_t<T> output_ptr(num_items);
+
+  // Test _ex version with custom build configuration
+  cccl_build_config config;
+  const char* extra_flags[]      = {"-DTEST_IDENTITY_ENABLED"};
+  const char* extra_dirs[]       = {TEST_INCLUDE_PATH};
+  config.extra_compile_flags     = extra_flags;
+  config.num_extra_compile_flags = 1;
+  config.extra_include_dirs      = extra_dirs;
+  config.num_extra_include_dirs  = 1;
```
NaderAlAwar: Critical: this is not initialized properly, specifically `enable_pch` and `verbose` are not initialized. The reduce tests initialize this properly in struct reduce_build_ex, we should do the same here.

c/parallel.v2/test/test_merge_sort.cpp
```diff
@@ -0,0 +1,708 @@
...
+  pointer_t<int> output_items_ptr;
+
+  // Test _ex version with custom build configuration
+  cccl_build_config config;
+  const char* extra_flags[]      = {"-DTEST_IDENTITY_ENABLED"};
+  const char* extra_dirs[]       = {TEST_INCLUDE_PATH};
+  config.extra_compile_flags     = extra_flags;
+  config.num_extra_compile_flags = 1;
+  config.extra_include_dirs      = extra_dirs;
+  config.num_extra_include_dirs  = 1;
```
NaderAlAwar: Critical: this is not initialized properly, specifically `enable_pch` and `verbose` are not initialized. The reduce tests initialize this properly in struct reduce_build_ex, we should do the same here.

c/parallel.v2/test/test_unique_by_key.cpp
```diff
@@ -0,0 +1,934 @@
...
+  pointer_t<std::size_t> output_num_selected_ptr(1);
+
+  // Test _ex version with custom build configuration
+  cccl_build_config config;
+  const char* extra_flags[]      = {"-DTEST_IDENTITY_ENABLED"};
+  const char* extra_dirs[]       = {TEST_INCLUDE_PATH};
+  config.extra_compile_flags     = extra_flags;
+  config.num_extra_compile_flags = 1;
+  config.extra_include_dirs      = extra_dirs;
+  config.num_extra_include_dirs  = 1;
```
NaderAlAwar: Critical: this is not initialized properly, specifically `enable_pch` and `verbose` are not initialized. The reduce tests initialize this properly in struct reduce_build_ex, we should do the same here.

c/parallel.v2/test/test_for.cpp
```diff
@@ -0,0 +1,339 @@
...
+  pointer_t<T> input_ptr(input);
+
+  // Test _ex version with custom build configuration
+  cccl_build_config config;
+  const char* extra_flags[]      = {"-DTEST_IDENTITY_ENABLED"};
+  const char* extra_dirs[]       = {TEST_INCLUDE_PATH};
+  config.extra_compile_flags     = extra_flags;
+  config.num_extra_compile_flags = 1;
+  config.extra_include_dirs      = extra_dirs;
+  config.num_extra_include_dirs  = 1;
```
NaderAlAwar: Critical: this is not initialized properly, specifically `enable_pch` and `verbose` are not initialized. The reduce tests initialize this properly in struct reduce_build_ex, we should do the same here.

c/parallel.v2/test/test_three_way_partition.cpp
```diff
@@ -0,0 +1,519 @@
...
+                    first_part,
+                    second_part,
+                    unselected)
+        == std::tie(other.num_items_in_first_part,
+                    other.num_items_in_second_part,
+                    other.num_unselected_items,
+                    other.first_part,
+                    other.second_part,
+                    other.unselected);
+  }
```
NaderAlAwar: Important: this compares the full output buffers, but we only want to compare up to `num_selected_in_first_part`, `num_selected_in_second_part`, and `num_unselected`

c/parallel.v2/test/test_unique_by_key.cpp
```diff
@@ -0,0 +1,934 @@
...
+      output_num_selected_ptr,
+      op,
+      num_items,
+      CU_STREAM_LEGACY));
+
+  // Verify results
+  size_t num_selected;
+  cudaMemcpy(&num_selected, static_cast<void*>(output_num_selected_ptr.ptr), sizeof(size_t), cudaMemcpyDeviceToHost);
+  REQUIRE(num_selected > 0);
+  REQUIRE(num_selected <= num_items);
```
NaderAlAwar: Important: we should also verify keys and values

c/parallel.v2/test/test_binary_search.cpp
```diff
@@ -0,0 +1,191 @@
...
+  }
+};
+
+struct binary_search_run
+{
+  template <typename... Ts>
+  CUresult operator()(BuildResultT build, void*, std::size_t*, cccl_binary_search_mode_t, Ts... args) const noexcept
+  {
+    return cccl_device_binary_search(build, args...);
+  }
```
NaderAlAwar: Important: this ignores the temp storage query, so this will execute and leave temp storage bytes at 0. We should do what `transform` does: set a dummy nonzero scratch size when storage is null, return success for the query call, and only invoke `cccl_device_binary_search` when scratch storage is provided

c/parallel.v2/test/test_unique_by_key.cpp
```diff
@@ -0,0 +1,934 @@
...
+  for (int i = 0; i < num_selected; ++i)
+  {
+    output_pairs.emplace_back(host_output_keys[i], host_output_values[i]);
+  }
+
+  REQUIRE(input_pairs == output_pairs);
+}
+
+struct UniqueByKey_Iterators_Fixture_Tag;
+C2H_TEST("DeviceMergeSort::SortPairs works with input and output iterators", "[merge_sort]")
```
NaderAlAwar: Nit:
```suggestion
C2H_TEST("DeviceMergeSort::SortPairs works with input and output iterators", "[unique_by_key]")
```

c/parallel.v2/test/test_unique_by_key.cpp
```diff
@@ -0,0 +1,934 @@
...
+
+  bool operator==(const large_key_pair& other) const
+  {
+    return a == other.a;
+  }
+};
+
+C2H_TEST("DeviceSelect::UniqueByKey fails to build for large types due to no vsmem", "[device][select_unique_by_key]")
+{
+  SKIP("v2 handles large types via a different memory path; the v1-only no-vsmem failure no longer applies");
```
NaderAlAwar: Important: if this passes now, we should change this test to verify that large types work

c/parallel.v2/test/test_reduce.cpp
```diff
@@ -0,0 +1,587 @@
...
+  reduce(input_ptr, output_ptr, num_items, op, init, CCCL_RUN_TO_RUN, cache_opt, test_key);
+
+  const T output   = output_ptr[0];
+  const T expected = std::accumulate(input.begin(), input.end(), init.value);
+  REQUIRE(output == expected);
+}
+
+struct Reduce_FloatingPointTypes_Fixture_Tag;
+using floating_point_types = c2h::type_list<
+#if _CCCL_HAS_NVFP16() && 0 // Disable for now
```
NaderAlAwar: Question: why is `__half` disabled? This also applies to some other tests

ci/build_cuda_cccl_python_v2.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+# Thin wrapper around build_cuda_cccl_python.sh that builds the cuda_cccl
+# wheel against the HostJIT-based cccl.c.parallel.v2 library instead of the
+# legacy NVRTC v1 library. The shared build script honors CCCL_PYTHON_USE_V2.
+set -euo pipefail
+
+ci_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+export CCCL_PYTHON_USE_V2=1
+exec "$ci_dir/build_cuda_cccl_python.sh" "$@"
```
NaderAlAwar: Question: is it an issue that the wheel will have the same name for both V2 and V1? At present both go through `get_wheel_artifact_name.sh` which does not distinguish between the two

ci/matrix.yaml
```diff
@@ -143,8 +145,8 @@ workflows:
     - {project: 'packaging', jobs: ['install']}
     # NVBench Helper testing:
     - {project: 'nvbench_helper', jobs: ['test'], ctk: '13.X', cxx: ['gcc', 'clang'], gpu: 'rtx2080'}
-    # c.parallel with HostJIT
-    - {jobs: ['test'], project: 'cccl_c_parallel_hostjit', ctk: '13.X', cxx: ['gcc13'], gpu: 'rtx2080'}
+    # c.parallel v2 (HostJIT-based)
+    - {jobs: ['test'], project: 'cccl_c_parallel_v2', ctk: '13.X', cxx: ['gcc13'], gpu: 'rtx2080'}
```
NaderAlAwar, line 149: Important: `python_v2` is declared as a lite dependent of `cccl_c_parallel_v2` in `project_files_and_dependencies.yaml`, so a `c/parallel.v2/` change can produce `LITE_BUILD=python_v2`. But `pull_request_lite` does not contain any `python_v2` row, so the workflow builder has no lite job to select and the Python v2 coverage is silently skipped for v2-only changes. We should add a `python_v2` row here, probably matching the full PR row, so the declared lite dependency actually runs

ci/project_files_and_dependencies.yaml
```diff
@@ -114,15 +114,30 @@ projects:
...
-    matrix_project: "cccl_c_parallel_hostjit"
-    lite_dependencies: [libcudacxx_public]
-    full_dependencies: [cccl_c_parallel_public]
+  cccl_c_parallel_v2:
+    name: "CCCL C Parallel Library v2 (HostJIT)"
+    matrix_project: "cccl_c_parallel_v2"
+    # v2 depends on libcudacxx, cub, and thrust headers (it JIT-compiles
+    # CUB's host+device code via HostJIT). Any change to those should trigger
+    # v2 to run.
+    lite_dependencies: [libcudacxx_public, cub_public, thrust_public]
```
NaderAlAwar: Important: this should also depend on `c2h` like cccl_c_parallel_internal

ci/project_files_and_dependencies.yaml
```diff
@@ -114,15 +114,30 @@ projects:
...
+    # cccl_c_parallel_v2 already pulls in libcudacxx/cub/thrust, so listing
+    # it here transitively triggers python_v2 on any of those upstream
+    # changes too. Direct includes catch Python-only edits.
+    lite_dependencies: [cccl_c_parallel_v2]
+    full_dependencies: []
     include_regexes:
-      - "c/parallel/src/hostjit/"
-      - "ci/build_cccl_c_parallel_hostjit\\.sh"
-      - "ci/test_cccl_c_parallel_hostjit\\.sh"
+      - "python/cuda_cccl/"
```
NaderAlAwar: Question: python_v1 just does `python` and `pyproject.toml`. Should we do the same here?

python/cuda_cccl/tests/compute/conftest.py
```diff
@@ -123,10 +123,43 @@ def guarded_import(name, *args, **kwargs):
...
+        from cuda.compute._build_info import USING_V2  # type: ignore[import-not-found]
+    except ImportError:
+        return False
+    return bool(USING_V2)
+
+
+# Individual tests known to crash on the v2 backend that don't match the
+# stateful/fp16 substring rules below. Match is on `item.name` (parametrized
+# id, e.g. "test_foo[int32]") OR on the bare function name. Add a one-line
+# reason for each so it's clear why it's deferred rather than fixed.
```
NaderAlAwar: Suggestion: we don't seem to skip any stateful or fp16 testsm we should remove that from the comment

python/cuda_cccl/pyproject.toml
```diff
@@ -3,7 +3,10 @@
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 [build-system]
-requires = ["scikit-build-core>=0.10", "setuptools_scm", "cython"]
+# cmake>=3.27 is needed for FindCUDAToolkit's CUDA::nvfatbin /
+# CUDA::nvfatbin_static targets, which the v2 (HostJIT) backend links
+# against. Listed here so isolated builds (pip's default) pick it up.
+requires = ["scikit-build-core>=0.10", "setuptools_scm", "cython", "cmake>=3.27"]
```
NaderAlAwar: Suggestion: `scikit-build` below still accepts cmake>=3.21 so it should probably be updated too. Same for CMakeLists.txt, it should be minimum 3.27

c/parallel.v2/src/reduce.cu
shwina: Hmm, the issue is that the env overload of `cub::DeviceReduce::Reduce` doesn't accept `temp_storage` - it materializes temp internally via the env's `memory_resource`. There's no public CUB API that takes both env *and* temp storage, and v2 exposes the two-step API to callers for parity with v1.

Leaving this as a follow-up so we land parity first. Until then, callers asking for `CCCL_NOT_GUARANTEED` get correct (deterministic) results, just not the faster algorithm.

python/cuda_cccl/pyproject.toml
shwina: Done.

python/cuda_cccl/tests/compute/conftest.py
shwina: Done.

ci/project_files_and_dependencies.yaml
shwina: Done.

ci/project_files_and_dependencies.yaml
shwina: Fixed.

ci/matrix.yaml
shwina: Done.

ci/build_cuda_cccl_python_v2.sh
shwina: Fixed - nice catch! Now we produce separate artifacts for v1/v2.

c/parallel.v2/test/test_reduce.cpp
shwina: Enabled it.

c/parallel.v2/test/test_unique_by_key.cpp
shwina: It does pass now and added a test.

c/parallel.v2/test/test_three_way_partition.cpp
shwina: Good catch. Fixed

c/parallel.v2/test/test_for.cpp
shwina: Fixed and ditto everywhere,

c/parallel.v2/src/util/errors.cpp
shwina: Removed. Ditto everywhere.

c/parallel.v2/include/cccl/c/reduce.h
shwina: Fixed.

c/parallel.v2/include/cccl/c/reduce.h
shwina: Done.

c/parallel.v2/src/reduce.cu
shwina: Fixed. Ditto everywhere.

c/parallel.v2/src/merge_sort.cu
shwina: Done.

c/parallel.v2/include/cccl/c/binary_search.h
```diff
@@ -0,0 +1,78 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of CUDA Experimental in CUDA C++ Core Libraries,
+// under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-FileCopyrightText: Copyright (c) 2024 NVIDIA CORPORATION & AFFILIATES.
```
miscco: Applies throughout
```suggestion
// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES.
```

c/parallel.v2/include/cccl/c/binary_search.h
```diff
@@ -0,0 +1,78 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+#pragma once
+// NOLINTBEGIN(modernize-use-using)
```
miscco, line 12: 😿

c/parallel.v2/src/hostjit/codegen/iterators.cpp
```diff
@@ -0,0 +1,269 @@
...
+    // host pass. They accumulate into `_delta` rather than calling the
+    // device-only `advance` bitcode. `operator*` (device-only) applies the
+    // accumulated `_delta` to a copy of state via `advance`, then derefs.
+    // `alignas({5})` matches the iterator's declared state alignment so the
+    // user-supplied advance/dereference (which casts state as a pointer/etc.)
+    // sees properly-aligned memory.
+    result.preamble += std::format(
+      "struct alignas({5}) {0} {{\n"
+      "  using value_type = {1};\n"
+      "  using difference_type = long long;\n"
```
bernhardmgruber: Important: Please use raw string literals

Comment applies everywhere.

c/parallel.v2/include/cccl/c/binary_search.h
```diff
@@ -0,0 +1,78 @@
...
+// Extended version with build configuration
+CCCL_C_API CUresult cccl_device_binary_search_build_ex(
+  cccl_device_binary_search_build_result_t* build,
+  cccl_binary_search_mode_t mode,
+  cccl_iterator_t d_data,
+  cccl_iterator_t d_values,
+  cccl_iterator_t d_out,
+  cccl_op_t op,
+  int cc_major,
+  int cc_minor,
```
bernhardmgruber, line 58: Q: We have `cuda::compute_capability` as a strong type. IIUC, the code here is a public API. Should we use a strong type here as well? Could we reuse `cuda::compute_capability`?

c/parallel.v2/include/cccl/c/extern_c.h
bernhardmgruber, line 1: Question: This file is a duplication of the one in CCCL.C v1. Can we reuse it? Or is the point that we want to straight delete the v1 directory at some point?

c/parallel.v2/src/util/build_utils.h
```diff
@@ -0,0 +1,183 @@
...
+// (e.g. /usr/local/cuda/targets/x86_64-linux/include), backs up to the real
+// toolkit root so callers find `nvvm/libdevice/libdevice.10.bc`.
+inline std::string parse_ctk_root(const char* ctk_path)
+{
+  if (!ctk_path || ctk_path[0] == '\0')
+  {
+    return {};
+  }
+  std::string p = ctk_path;
+  if (p.substr(0, 2) == "-I")
```
bernhardmgruber:
```suggestion
  if (p.starts_with("-I"))
```
Applies to all such string operations across this PR.

c/parallel.v2/src/util/build_utils.h
```diff
@@ -0,0 +1,183 @@
...
+  auto strip_dash_I = [](const char* in) -> std::string {
+    if (!in || in[0] == '\0')
+    {
+      return {};
+    }
+    std::string p = in;
+    if (p.size() >= 2 && p.substr(0, 2) == "-I")
+    {
+      p = p.substr(2);
+    }
```
bernhardmgruber: Seems duplicate to the logic two functions above. Please unify.

c/parallel.v2/src/util/build_utils.h
```diff
@@ -0,0 +1,183 @@
...
+    auto add = [&](const char* p) {
+      if (!p || p[0] == '\0')
+      {
+        return;
+      }
+      std::string s = p;
+      if (s.size() >= 2 && s.substr(0, 2) == "-I")
+      {
+        s = s.substr(2);
+      }
```
bernhardmgruber: I believe I have seen this before. This looks vibe coded by now.

c/parallel.v2/src/util/build_utils.h
```diff
@@ -0,0 +1,183 @@
...
+      owned_strs_.push_back(std::move(s));
```
bernhardmgruber: Suggestion: `owned_strs_.reserve` the number of CUB and Thrust paths to avoid reallocations.

c/parallel.v2/src/util/build_utils.h
```diff
@@ -0,0 +1,183 @@
...
+{
+  if (cubin.empty())
+  {
+    out_cubin = nullptr;
+    out_size  = 0;
+    return;
+  }
+  auto buf = std::make_unique<char[]>(cubin.size());
+  std::memcpy(buf.get(), cubin.data(), cubin.size());
+  out_cubin = buf.release();
```
bernhardmgruber: Suggestion: I don't see the point of using `std::make_unique` here, since we call `delete[]` in the function below. I would just call `new`.

c/parallel.v2/src/binary_search.cu
```diff
@@ -0,0 +1,146 @@
...
+try
+{
+  std::string cccl_include_str  = cccl::detail::parse_cccl_include_path(libcudacxx_path);
+  std::string ctk_root_str      = cccl::detail::parse_ctk_root(ctk_path);
+  const char* cccl_include_path = cccl_include_str.empty() ? nullptr : cccl_include_str.c_str();
+  const char* ctk_root          = ctk_root_str.empty() ? nullptr : ctk_root_str.c_str();
+  cccl::detail::MergedBuildConfig merged(config, cub_path, thrust_path);
+
+  const char* find_fn =
+    (mode == CCCL_BINARY_SEARCH_LOWER_BOUND) ? "cub::DeviceFind::LowerBound" : "cub::DeviceFind::UpperBound";
```
bernhardmgruber, line 46: Important: if `CCCL_BINARY_SEARCH_LOWER_BOUND` is an enumerator, it should be spelled lower_case.

c/parallel.v2/src/binary_search.cu
```diff
@@ -0,0 +1,146 @@
...
+  cccl_iterator_t d_data,
+  uint64_t num_items,
+  cccl_iterator_t d_values,
+  uint64_t num_values,
+  cccl_iterator_t d_out,
+  cccl_op_t op,
+  CUstream stream)
+{
+  try
+  {
```
bernhardmgruber, line 82: Suggestion: The previous function used a function try block, why not here?

```suggestion
try
{
```

c/parallel.v2/src/binary_search.cu
```diff
@@ -0,0 +1,146 @@
...
+    auto fn = reinterpret_cast<binary_search_fn_t>(build.binary_search_fn);
+    if (!fn)
+    {
+      return CUDA_ERROR_INVALID_VALUE;
+    }
```
bernhardmgruber: Question and mild suggestion: In my understanding, a `reinterpret_cast` does not alter the representation in memory, so why are we doing the null check after the cast? It seems like something was intended here, that C++ does not provide. If this check is validating an API precondition, I would put it first:

```suggestion
    if (!build.binary_search_fn)
    {
      return CUDA_ERROR_INVALID_VALUE;
    }
    auto fn = reinterpret_cast<binary_search_fn_t>(build.binary_search_fn);
```

c/parallel.v2/src/binary_search.cu
```diff
@@ -0,0 +1,146 @@
...
+
+    int status =
+      fn(d_data.state, num_items, d_values.state, num_values, d_out.state, op.state, reinterpret_cast<void*>(stream));
```
bernhardmgruber:
```suggestion
    const int status =
      fn(d_data.state, num_items, d_values.state, num_values, d_out.state, op.state, reinterpret_cast<void*>(stream));
```

c/parallel.v2/src/histogram.cu
```diff
@@ -0,0 +1,213 @@
...
+//   fn(temp, temp_bytes,
+//      d_samples,                  // input iterator state
+//      d_histogram,                // output pointer (counter_t*)
+//      &num_levels,                // int (host pointer)
+//      &lower_level, &upper_level, // level_t (host pointer)
+//      &num_row_pixels,            // long long (host pointer)
+//      &num_rows,                  // long long (host pointer)
+//      &row_stride_bytes,          // size_t (host-precomputed: row_stride_samples * sizeof(sample_t))
+//      stream)
+using histogram_fn_t = int (*)(void*, size_t*, void*, void*, void*, void*, void*, void*, void*, void*, void*);
```
bernhardmgruber, line 29: Suggestion: This mapping of JIT produced call signatures and the definition here look brittle. I assume if they don't match up, we get runtime crashes. Is there any way we can validate if they match?

c/parallel/test/test_three_way_partition.cpp
```diff
@@ -127,18 +127,15 @@ struct three_way_partition_result_t
...
-                    other.unselected);
+    if (num_items_in_first_part != other.num_items_in_first_part
+        || num_items_in_second_part != other.num_items_in_second_part
+        || num_unselected_items != other.num_unselected_items)
+    {
+      return false;
+    }
+    return std::equal(first_part.begin(), first_part.begin() + num_items_in_first_part, other.first_part.begin())
+        && std::equal(second_part.begin(), second_part.begin() + num_items_in_second_part, other.second_part.begin())
+        && std::equal(unselected.begin(), unselected.begin() + num_unselected_items, other.unselected.begin());
```
bernhardmgruber, line 138: Question: Why is this rewrite required? The code before looked fine to me.

bernhardmgruber (commented):
I looked at a few implementation files and headers

c/parallel/test/test_three_way_partition.cpp
shwina: https://github.com/NVIDIA/cccl/pull/8985/#discussion_r3284257286

shwina: Huh - that's annoying. I meant to link `https://github.com/NVIDIA/cccl/pull/8985/#discussion_r3284257286`.

python/cuda_cccl/CMakeLists.txt
```diff
@@ -2,7 +2,7 @@
 #
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
-cmake_minimum_required(VERSION 3.21...3.31 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.27...3.31 FATAL_ERROR)
```
NaderAlAwar: Question: Should this minimum be CMake 3.30 instead of 3.27? The v2 HostJIT CMake links `CUDA::nvfatbin` / `CUDA::nvfatbin_static`, and CMake’s `FindCUDAToolkit` docs list the `nvFatBin` imported targets as new in CMake 3.30: https://cmake.org/cmake/help/v3.30/module/FindCUDAToolkit.html#nvfatbin. CMake 3.27 added `CUDA_FATBIN_COMPILATION`, but this PR is using the `FindCUDAToolkit` imported targets, not that property.

c/parallel.v2/CMakeLists.txt
```diff
@@ -0,0 +1,109 @@
+cmake_minimum_required(VERSION 3.21)
```
NaderAlAwar: Important: this should be 3.30 as well I believe

ci/project_files_and_dependencies.yaml
NaderAlAwar, line 200: Important: we should add `cccl_c_parallel_v2` here as well

c/parallel.v2/include/cccl/c/binary_search.h
NaderAlAwar: I don't think that's possible because this is a C API

c/parallel.v2/src/reduce.cu
```diff
@@ -0,0 +1,249 @@
...
+  std::string cccl_include_str;
+  if (libcudacxx_path && libcudacxx_path[0] != '\0')
+  {
+    cccl_include_str = libcudacxx_path;
+    if (cccl_include_str.substr(0, 2) == "-I")
+    {
+      cccl_include_str = cccl_include_str.substr(2);
+    }
+    cccl_include_path = cccl_include_str.c_str();
+  }
```
NaderAlAwar: Important: this looks like it duplicates `build_utils.h`, it should reuse that logic instead

c/parallel.v2/src/radix_sort.cu
```diff
@@ -0,0 +1,266 @@
...
+        &begin_bit,
+        &end_bit,
+        reinterpret_cast<void*>(stream));
+    }
+
+    if (selector)
+    {
+      // Copy variant always writes to d_keys_out (= d_buffers[1] in DoubleBuffer mode).
+      *selector = is_overwrite_okay ? 1 : 0;
+    }
```
NaderAlAwar: Critical: this breaks the behavior in cccl.c v1. When `is_overwrite_okay` is true, we should construct a `DoubleBuffer` then call the device radix sort function. Afterwards, we should retrieve the selector from the double buffer object we passed to the function call

c/parallel.v2/src/segmented_sort.cu
```diff
@@ -0,0 +1,252 @@
...
+        start_offset_in.state,
+        end_offset_in.state,
+        reinterpret_cast<void*>(stream));
+    }
+
+    if (selector)
+    {
+      // Copy variant always writes to d_keys_out (= d_buffers[1] in DoubleBuffer mode).
+      *selector = is_overwrite_okay ? 1 : 0;
+    }
```
NaderAlAwar: Critical: same issue as radix_sort regarding `DoubleBuffer`

c/parallel/test/test_three_way_partition.cpp
bernhardmgruber: Oh, I missed that. Thanks!

c/parallel.v2/src/binary_search.cu
shwina: This is a pre-existing convention in v1: https://github.com/NVIDIA/cccl/blob/52931dd9a79f2159c8c30b7852898636f8d17f0f/c/parallel/include/cccl/c/types.h#L172. Do you want me to change it in this PR?

c/parallel.v2/src/binary_search.cu
shwina: ✅

c/parallel.v2/src/radix_sort.cu
shwina: Addressed this by:

- introducing a `double_buffer` tag for `CubCall`
- adding an overload for `CubCall::compile` that accepts a list of functions to compile so that we compile a wrapper for both the copy overload and double-buffer overload of `DeviceRadixSort::Sort*` - previously the assumption was that only a single wrapper was produced and compiled

c/parallel.v2/src/segmented_sort.cu
shwina: Same resolution as above.

c/parallel.v2/src/reduce.cu
shwina: ✅

shwina:
I removed all Python changes from this PR. Will come in a follow-up (the scope of Python changes increased to be non-trivial).

c/parallel.v2/src/hostjit/codegen/iterators.cpp
shwina: Done.

c/parallel.v2/include/cccl/c/extern_c.h
shwina: > Or is the point that we want to straight delete the v1 directory at some point?

Pretty much yeah. Other headers are also almost identical.

c/parallel.v2/src/util/build_utils.h
shwina: Done.

c/parallel.v2/src/util/build_utils.h
shwina: Done.

c/parallel.v2/src/util/build_utils.h
shwina: Done.

c/parallel.v2/src/util/build_utils.h
shwina: Done.

c/parallel.v2/src/binary_search.cu
shwina: (and in other places that we could use function try).

c/parallel.v2/src/binary_search.cu
shwina: Done.

c/parallel.v2/src/binary_search.cu
shwina: Done.

c/parallel.v2/src/histogram.cu
shwina: I agree there's some brittleness here - although any such mistakes _should_ easily be caught by tests. I'm open to suggestions for how to address this (I'm sure some template metaprogramming solution exists).

But if it's OK I'd prefer not to block merging on this.

python/cuda_cccl/CMakeLists.txt
shwina: I've punted on Python changes to a follow-up PR, but yes we can bump the min version.

c/parallel.v2/CMakeLists.txt
shwina: Done.

ci/project_files_and_dependencies.yaml
shwina: Done.

gevtushenko (approved):

bernhardmgruber:
pre-commit.ci autofix

c/parallel.v2/src/binary_search.cu
bernhardmgruber: Let's leave it for now.

c/parallel.v2/src/histogram.cu
bernhardmgruber: Agreed. No need to block the PR on it.

ci/matrix.yaml
```diff
@@ -20,8 +20,6 @@ workflows:
   #   - { jobs: ['run_gpu'], project: 'target', ctk: ['12.X', '13.X'], cxx: ['gcc', 'clang'], gpu: 'rtx2080',
   #       args: '--preset libcudacxx --lit-tests "cuda/utility/basic_any.pass.cpp"' }
   #
-  override:
-
```
alliepiper, line 24: Don't delete this, leave the stub in place please,

alliepiper (approved):
Approved -- only a trivial change needed

ci/matrix.yaml
```diff
@@ -20,8 +20,6 @@ workflows:
   #   - { jobs: ['run_gpu'], project: 'target', ctk: ['12.X', '13.X'], cxx: ['gcc', 'clang'], gpu: 'rtx2080',
   #       args: '--preset libcudacxx --lit-tests "cuda/utility/basic_any.pass.cpp"' }
   #
```
alliepiper, line 22:
```suggestion
  #
  override:

```

shwina:
/ok to test https://github.com/NVIDIA/cccl/pull/8985/commits/648364f13d0b7e5909ac1fbd08e6b60567959a98

PR #8991, opened 2026-05-14
`cuda::simd` Add `saturating_add/sub`

fbusato:
## Description

Introduce SIMD `saturating_add/sub` to use `VIADD.S8x4.ISAT`, `VIADD.U8x4.ISAT`, `VIADD.S16x2.ISAT`, and `VIADD.16x2.ISAT` on SM107f and SM120f.


libcudacxx/include/cuda/__simd/simd_intrinsics_array.h
```diff
@@ -0,0 +1,83 @@
...
+
+#  include <cuda/std/__cccl/prologue.h>
+
+_CCCL_BEGIN_NAMESPACE_CUDA_SIMD
+
+//----------------------------------------------------------------------------------------------------------------------
+// device-only functions
+
+template <typename _Tp, ::cuda::std::size_t _Np>
+[[nodiscard]] _CCCL_DEVICE_API constexpr ::cuda::std::simd::__array_u32_t<_Np> __vadd_sat_16bit_x2(
```
miscco, line 39: Important, I believe the separation in intrinsics / array intrinsics is not the right one. 

I believe we should categorize by function, so everything related to `add_sat` gets into its own file

fbusato: I following the same exact organization of `cuda/std/__simd`. We can always reorganize it later

griwes (approved):
Reviewed throughout. LGTM once you merge main in and adapt to the changes in the codegen cmakelists.

libcudacxx/include/cuda/std/__internal/features.h
```diff
@@ -115,6 +115,13 @@
 #define _CCCL_HAS_SIMD_8BIT() \
   ((_CCCL_HAS_SIMD_8BIT_PTX() || _CCCL_HAS_SIMD_8BIT_INTRINSICS()) && !_CCCL_TILE_COMPILATION())
 
+// TODO(fbusato): CTK 13.4 produces non-optimal code for SIMD SAT intrinsics
+#define _CCCL_HAS_SIMD_SAT_INTRINSICS() 0
```
miscco, line 119: Question: how does this work with clang-cuda?

miscco (approved):

libcudacxx/include/cuda/std/__internal/features.h
fbusato: clang-cuda doesn't support these intrinsics yet

davebayer:
Can we revert this PR and wait until [P2956R2](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2026/p2956r2.html) is accepted to C++ standard? I don't think this is a feature someone was asking for right now and we will have to go through deprecating + removing it from `cuda::` namespace basically for no reason

PR #8994, opened 2026-05-14
`cuda::simd` Add `abs_diff`

fbusato:
## Description

Introduce SIMD `abs_diff` to compute the vectorize absolute difference and use `VABSDIFF`.

Requires:

- https://github.com/NVIDIA/cccl/pull/8991


libcudacxx/include/cuda/std/__internal/features.h
```diff
@@ -103,8 +103,21 @@
...
+#define _CCCL_HAS_SIMD_8BIT_INTRINSICS() 0 // TODO(fbusato): CTK 13.2 produces non-optimal code for 8-bit SIMD instrs.
+#define _CCCL_HAS_SIMD_8BIT_PTX()        (__cccl_ptx_isa >= 920ULL)
+#define _CCCL_HAS_SIMD_8BIT() \
+  (_CCCL_HAS_SIMD_8BIT_PTX() || _CCCL_HAS_SIMD_8BIT_INTRINSICS()) && !_CCCL_TILE_COMPILATION()
+
+// TODO(fbusato): CTK 13.2 produces non-optimal code for SIMD SAT intrinsics
+#define _CCCL_HAS_SIMD_SAT_INTRINSICS() 0
+#define _CCCL_HAS_SIMD_SAT_PTX()        (__cccl_ptx_isa >= 920ULL)
+#define _CCCL_HAS_SIMD_SAT() \
+  (_CCCL_HAS_SIMD_SAT_PTX() || _CCCL_HAS_SIMD_SAT_INTRINSICS()) && _CCCL_CUDA_COMPILATION() && !_CCCL_TILE_COMPILATION()
```
Jacobfaib: Important: wrap the whole expression in `()` (and below), to avoid any unintended operator precedence shenanigans.

libcudacxx/include/cuda/std/__simd/specializations/fixed_size_integral_vec.h
```diff
@@ -0,0 +1,252 @@
...
+  __is_fixed_size_small_integral_v<_Tp, _Np> ? __simd_operations_small_integral : __simd_operations_kind::__default;
+
+// Simd operations for fixed_size ABI with small integral element types.
+template <typename _Tp, __simd_size_type _Np>
+struct __simd_operations<_Tp, __fixed_size<_Np>, __simd_operations_small_integral> : __fixed_size_operations<_Tp, _Np>
+{
+  using __base           = __fixed_size_operations<_Tp, _Np>;
+  using __simd_storage_t = __simd_storage<_Tp, __fixed_size<_Np>>;
+
+  // all computation is done on uint32_t, so the alignment must be at least the alignment of uint32_t
```
Jacobfaib, line 77: This comment would also be useful above when you introduce `__is_fixed_size_small_integral_v`. The choice of `uint32_t` is not immediately apparent there.

libcudacxx/include/cuda/std/__simd/specializations/simd_intrinsics.h
```diff
@@ -0,0 +1,93 @@
...
+{
+  NV_IF_TARGET(NV_PROVIDES_SM_90,
+               (return ::__vadd2(__lhs, __rhs);), //
+               (_CCCL_VERIFY(false, "cuda::std::simd::__vadd_u16x2: Unsupported architecture"); return uint32_t{};));
+}
+
+[[nodiscard]] _CCCL_DEVICE_API inline uint32_t
+__vadd_s16x2([[maybe_unused]] const uint32_t __lhs, [[maybe_unused]] const uint32_t __rhs) noexcept
+{
+  // prevent MSVC warning
```
Jacobfaib: What prevents the MSVC warning? 

Also minor nit, but you could decalre `uint32_t __result{}` outside the `NV_IF_TARGET`, then you can also move the `return __result` outside and keep a single path through the whole function in either case.

libcudacxx/include/cuda/std/__simd/specializations/simd_intrinsics_array.h
```diff
@@ -0,0 +1,133 @@
...
+
+template <typename _SimdStorage>
+inline constexpr size_t __simd_storage_size_u32 = 0;
+
+template <typename _Tp, __simd_size_type _Np>
+inline constexpr size_t __simd_storage_size_u32<__simd_storage<_Tp, __fixed_size<_Np>>> =
+  ::cuda::ceil_div(_Np, sizeof(uint32_t) / sizeof(_Tp));
+
+template <size_t _Np>
+using __array_u32_t = array<uint32_t, _Np>;
```
Jacobfaib, line 49: IMO this `using` decl does not help readability that much. It is useful to the reader that `__array_u32_t` is actually a `cuda::std::array`. I would spell out `array<uint32_t, _NP>` at use-sites.

As time has gone on I've begun to subscribe more and more to the Linux kernel coding style for typedefs (https://www.kernel.org/doc/html/v5.0/process/coding-style.html#typedefs). The TL;DR: is that typedefs should only be used when:

1. You wish to actively and completely obfuscate the type. I.e. the user must not know any other information about the type than what you allow.
2. The underlying type is completely irrelevant to its use. The common example here are the underlying types for `uin32_t` etc.

libcudacxx/include/cuda/std/__simd/type_traits.h
```diff
@@ -32,16 +32,23 @@
 
 _CCCL_BEGIN_NAMESPACE_CUDA_STD_SIMD
 
+#if _CCCL_CTK_AT_LEAST(12, 9)
+inline constexpr size_t __optimal_cuda_alignment = 32;
+#else
+inline constexpr size_t __optimal_cuda_alignment = 16;
+#endif // _CCCL_CTK_AT_LEAST(12, 9)
```
Jacobfaib: You use could ternary here.

libcudacxx/include/cuda/std/__simd/specializations/fixed_size_integral_vec.h
fbusato: the problem is this PR (and other simd PRs) are "stacked" on top of several ones. IIRC I already addressed this point in the first one (I will double-check).

libcudacxx/include/cuda/std/__simd/specializations/simd_intrinsics.h
fbusato: MSVC is not very meaningful, I will remove it. Good point about `__result`

libcudacxx/include/cuda/std/__simd/type_traits.h
fbusato: also here, I already fixed it in other PR

libcudacxx/include/cuda/std/__simd/specializations/simd_intrinsics_array.h
fbusato: tried. I find `__array_u32_t` easier to read, even because it is repeated several times in the code

libcudacxx/include/cuda/std/__simd/specializations/fixed_size_integral_vec.h
fbusato: yes, already addressed here https://github.com/NVIDIA/cccl/pull/8873

docs/libcudacxx/extended_api/simd/saturating_add.rst
miscco, line 1: Question: The title talks about `abs_diff` but this also adds `saturating_add`. Why is that in this PR and only `saturating_add`

libcudacxx/include/cuda/__simd/simd_intrinsics_array.h
```diff
@@ -0,0 +1,113 @@
...
+//
+// Part of libcu++ in the CUDA C++ Core Libraries,
+// under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef _CUDA___SIMD_SIMD_INTRINSICS_ARRAY_H
```
miscco, line 11: This file has a very confusing name

libcudacxx/include/cuda/__simd/abs_diff.h
```diff
@@ -0,0 +1,100 @@
...
+        NV_IF_TARGET(NV_IS_DEVICE, ({
+                       using __unsigned_storage_t = ::cuda::std::simd::__simd_storage_u32_t<_ResultStorage>;
+                       constexpr __unsigned_storage_t __c_u{};
+                       const auto __lhs_u    = ::cuda::std::simd::__to_unsigned_storage(__lhs);
+                       const auto __rhs_u    = ::cuda::std::simd::__to_unsigned_storage(__rhs);
+                       const auto __result_u = ::cuda::simd::__vabsdiff_8bit_x4<_Tp>(__lhs_u, __rhs_u, __c_u);
+                       return ::cuda::std::simd::__copy_from_unsigned_storage<_ResultStorage>(__result_u);
+                     }));
+      }
+      // 16-bit path (vabsdiff2) generates worst code in most cases
```
miscco:
```suggestion
      // 16-bit path (vabsdiff2) generates worse code in most cases
```

libcudacxx/include/cuda/__simd/simd_intrinsics.h
```diff
@@ -31,73 +31,109 @@
...
+                 return __result;
+               }))
+  _CCCL_VERIFY(false, "cuda::simd::__vabsdiff_u8x4: Unsupported architecture");
+  return ::cuda::std::uint32_t{};
+}
+
+[[nodiscard]] _CCCL_DEVICE_API inline ::cuda::std::uint32_t __vabsdiff_s8x4(
+  [[maybe_unused]] const ::cuda::std::uint32_t __lhs,
+  [[maybe_unused]] const ::cuda::std::uint32_t __rhs,
+  [[maybe_unused]] const ::cuda::std::uint32_t __c) noexcept
```
miscco, line 121: Question: Shouldnt those be signed integers?

fbusato: I don't think so. I used `uint32_t` to express a sequence of bit (packed values)

libcudacxx/include/cuda/__simd/abs_diff.h
```diff
@@ -0,0 +1,100 @@
...
+    _CCCL_PRAGMA_UNROLL_FULL()
+    for (::cuda::std::simd::__simd_size_type __i = 0; __i < __size; ++__i)
+    {
+      const __result_value_t __lhs_i = static_cast<__result_value_t>(__lhs.__data[__i]);
+      const __result_value_t __rhs_i = static_cast<__result_value_t>(__rhs.__data[__i]);
+      const bool __is_less           = __lhs.__data[__i] < __rhs.__data[__i];
+      __result.__data[__i]           = __is_less ? (__rhs_i - __lhs_i) : (__lhs_i - __rhs_i);
+    }
+    return __result;
+  }
```
davebayer: Why does the fallback computes everything per element and doesn't use `simd` operations?

We could just do vectorized min/max + vectorized subtraction

davebayer:
And I believe we should provide a non-simd variant, too

libcudacxx/include/cuda/__simd/abs_diff.h
fbusato: uhmm, my concern is the duplicate computation. Let me check the generated SASS

fbusato: didn't work directly (higher number of SASS instrs.). I had to add a specialized path to generate `VIADD C, A, -B` + I found a compiler bug for the 32-bit case

libcudacxx/include/cuda/__simd/abs_diff.h
```diff
@@ -0,0 +1,93 @@
...
+  using __result_type _CCCL_NODEBUG = ::cuda::std::simd::basic_vec<::cuda::std::make_unsigned_t<_Tp>, _Abi>;
+#if _CCCL_HAS_SIMD_VABSDIFF()
+  _CCCL_IF_NOT_CONSTEVAL_DEFAULT
+  {
+    if constexpr (sizeof(_Tp) == sizeof(::cuda::std::uint8_t))
+    {
+      NV_IF_TARGET(NV_IS_DEVICE,
+                   (return __simd_abs_diff_impl(
+                             __lhs, __rhs, __abs_diff_operation<_Tp>{}, static_cast<__result_type*>(nullptr));)) // ADL
+    }
```
miscco: Question: Why is this using ADL? we know the namespace this is in?

fbusato: `__simd_abs_diff_impl` is a friend function within `basic_vec`. The reason is that `std::simd` does not allow storing results with `operator[](index)` (only reading).

libcudacxx/include/cuda/__simd/simd_intrinsics.h
```diff
@@ -31,73 +31,109 @@
...
+[[nodiscard]] _CCCL_DEVICE_API inline ::cuda::std::uint32_t __vabsdiff_u8x4(
+  [[maybe_unused]] const ::cuda::std::uint32_t __lhs,
+  [[maybe_unused]] const ::cuda::std::uint32_t __rhs,
+  [[maybe_unused]] const ::cuda::std::uint32_t __c) noexcept
+{
+  NV_IF_TARGET(NV_IS_DEVICE, ({
+                 ::cuda::std::uint32_t __result{};
+                 asm("vabsdiff4.u32.u32.u32 %0, %1, %2, %3;" : "=r"(__result) : "r"(__lhs), "r"(__rhs), "r"(__c));
+                 return __result;
+               }))
```
davebayer, line 113: Why don't you use `__vabsdiffu4`?

libcudacxx/include/cuda/__simd/simd_intrinsics.h
```diff
@@ -31,73 +31,109 @@
...
+  [[maybe_unused]] const ::cuda::std::uint32_t __rhs,
+  [[maybe_unused]] const ::cuda::std::uint32_t __c) noexcept
+{
+  NV_IF_TARGET(NV_IS_DEVICE, ({
+                 ::cuda::std::uint32_t __result{};
+                 asm("vabsdiff4.u32.s32.s32 %0, %1, %2, %3;" : "=r"(__result) : "r"(__lhs), "r"(__rhs), "r"(__c));
+                 return __result;
+               }))
+  _CCCL_VERIFY(false, "cuda::simd::__vabsdiff_s8x4: Unsupported architecture");
+  return ::cuda::std::uint32_t{};
```
davebayer, line 129: Same here, can't we use `__vabsdiffs4`?

fbusato: discussed offine. Builtins are not efficient

davebayer (approved):

PR #8999, opened 2026-05-15
[cub] Replace `assert` with `CHECK` or `REQUIRE` in tests

davebayer:
Using `assert` outside of libcu++ for testing is not safe, because we compile with optimizations enabled and that meanst `NDEBUG` is defined, thus `assert` expands to noop. This PR replaces them with catch2 test macros.


cub/test/internal/catch2_test_integer_utils.cu
```diff
@@ -1,142 +0,0 @@
-// SPDX-FileCopyrightText: Copyright (c) 2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
-// SPDX-License-Identifier: BSD-3-Clause
-#include <cub/detail/integer_utils.cuh>
```
davebayer, line 3: I've removed the source file & test. The functionality is not used anywhere in cub, plus it was never actually tested, because the `assert` expanded to noop instead of the test due to missing `#undef NDEBUG`

bernhardmgruber (approved):

pciolkosz (approved):

cub/test/catch2_test_block_radix_sort_custom.cu
```diff
@@ -82,9 +80,9 @@ __global__ void sort_keys()
      }};
   // example-end keys
 
-  assert(thread_keys[threadIdx.x][0] == expected_output[threadIdx.x][0]);
-  assert(thread_keys[threadIdx.x][1] == expected_output[threadIdx.x][1]);
-  assert(thread_keys[threadIdx.x][2] == expected_output[threadIdx.x][2]);
+  CHECK(thread_keys[threadIdx.x][0] == expected_output[threadIdx.x][0]);
+  CHECK(thread_keys[threadIdx.x][1] == expected_output[threadIdx.x][1]);
+  CHECK(thread_keys[threadIdx.x][2] == expected_output[threadIdx.x][2]);
```
davebayer: @gevtushenko is strongly for using `REQUIRE` here and in the rest of the PR

bernhardmgruber: I remain against. As a person that has to figure out what's going on in case of a test failure, I need the largest context possible of meaningful information.

That means, I don't want to see dependent failures, like the failed comparison with an expected result after a failed call to radix sort. That call to radix sort needs a `REQUIRE` of course. But a comparison of the results keys and values needs a `CHECK` since in case of a mismatch in the result, I want to see the mismatch of the keys **and** values. This may allow me to deduce better, what kind of failure happened, without first modifying the test code and recompiling.

@gevtushenko what would be the advantage of `REQUIRE` here? I don't see it.

pauleonix: To add some context from the code review hour: According to @davebayer `CHECK` in device code is currently implemented to do the same as `REQUIRE` because anything else would be hard to implement. So with the current implementation you get `REQUIRE` semantics anyway and @gevtushenko just argued that this state is confusing and that `CHECK` should not compile in device code if it can't provide the correct semantics.

bernhardmgruber: I see. I didn't see this was device code. Fine then!

PR #9004, opened 2026-05-15
[Tile] Avoid returns inside loops in `mdspan`

miscco:
tile does not allow returns inside of loops currently

So avoid them by querying the current state


libcudacxx/include/cuda/__mdspan/layout_stride_relaxed.h
```diff
@@ -65,13 +65,19 @@ template <class _StridedMapping>
...
   for (_RankType __r = 0; __r != __rank; ++__r)
   {
     if (__mapping.extents().extent(__r) == 0)
     {
-      return typename _StridedMapping::index_type{0};
+      __extent_is_zero = true;
+      break;
     }
   }
+  if (__extent_is_zero)
```
davebayer, line 77: This is just so stupid

davebayer (approved):

PR #9005, opened 2026-05-15
[Tile] Make `arch_id` `_CCCL_HOST_DEVICE_API`

miscco:
It relies on returns inside switch statements, which is currently not supported


davebayer (approved):

libcudacxx/test/libcudacxx/cuda/ccclrt/device/is_specific_arch.pass.cpp
```diff
@@ -8,6 +8,9 @@
 //
 //===----------------------------------------------------------------------===//
 
+// UNSUPPORTED: enable-tile
+// error: return in loop statement is not supported
```
davebayer, line 12: Is this the right error description?

PR #9013, opened 2026-05-15
[Tile] Mark `simd` as `_CCCL_HOST_DEVICE`

miscco:
it relies on asm and manuall loop unrolling a lot


davebayer (approved):

fbusato:
not entirely agree. We could disable unrolling pragma macro only in Tile. The code uses `asm` only for the optimized path which can be disabled

PR #9023, opened 2026-05-15
[Tile] Avoid internal usage of CPOs

miscco:
We cannot use CPOs in tile programs, because they live in device memory and its currently not possible to have `__device__ __tile__` variables

So avoid this by constructing the empty struct instead of accessing the global instance


miscco:
@coderabbitai full review

fbusato (approved):

davebayer (approved):
Image

PR #9027, opened 2026-05-15
[Tile] Avoid return statements in loops in `bitset`

miscco:
tile does not allow return statements in loops.


fbusato (approved):
hopefully there will not be regressions...

davebayer (approved):

PR #9036, opened 2026-05-15
Complex tanh (and thus tan) accuracy refinement

s-oboyle:
## Update the complex tanh function to avoid numerical issues.

The current complex tanh function has problems with both accuracy and over/underflow
This new version fixes these accuracy issues while giving a speedup.

## Perf
Testing on H100 we see decent speedup. 
Using the math-teams standard math_bench test we have the following (Similar results on a 5080):

**Operations/SM/cycle:**
`ctanh():`
| --- | --- | --- | --- |
| fp64 | 0.2163 | 0.3565 | 1.65 | 
| fp32 | 0.6407 | 0.8284 | 1.29 |

## Correctness
Apart from the usual over/underflow suspects, there are also some intervals where catastrophic cancellation causes accuracy loss.

### GPU Correctness
```
GPU fp64:
Max ulp real error (8.316,3.366) @ (7.270392412e-15,1.507304668)        (0x3d005f188b8eb261,0x3ff81deb7fd12fae)                       
        Ours = (1.805960114e-12,15.72893075)    Ref = (1.805960114e-12,15.72893075)                                                   
        Ours = (0x3d7fc55240728a71,0x402f753669549df9)               Ref = (0x3d7fc55240728a69,0x402f753669549df6)                    
                                                                                                                                      
Max ulp imag error (2.948,5.811) @ (3.333738963e-05,-2.171464686e+77)   (0x3f017a77fc5190dc,0xcffe0148bb9741c2) 
        Ours = (6.580654335e-05,0.9868921372)    Ref = (6.580654335e-05,0.9868921372)                                                 
        Ours = (0x3f114033ce64fb7a,0x3fef949ed1bfde0f)               Ref = (0x3f114033ce64fb7d,0x3fef949ed1bfde15)
```

```
GPU fp32:
Max ulp real error (8.016,3.672) @ (-3.960616141e-06,1.310337782)       (0xb684e56f,0x3fa7b926)
        Ours = (-5.97211656e-05,3.752166986)    Ref = (-5.97211947e-05,3.75216794)                                                    
        Ours = (0xb87a7d1d,0x40702381)               Ref = (0xb87a7d25,0x40702385)                                                    
                                                                                                                                      
Max ulp imag error (0.8914,5.721) @ (-3.473282576,2.940004844e+34)      (0xc05e4a43,0x78b53119)                 
        Ours = (-1.000158906,0.001917456975)    Ref = (-1.000158787,0.001917456277)    
        Ours = (0xbf800535,0x3afb532e)               Ref = (0xbf800534,0x3afb5328) 
```

### CPU Correctness
```
CPU fp64:
Max ulp real error (6.902,3.774) @ (7.979126863e-06,1.830677559)        (0x3ee0bbc2d87c7800,0x3ffd4a748d384c00)
        Ours = (0.0001208383555,-3.760891926)    Ref = (0.0001208383555,-3.760891926)          
        Ours = (0x3f1fad5323efb6fd,0xc00e164e8194b79d)               Ref = (0x3f1fad5323efb704,0xc00e164e8194b7a1)
                                                                                                                                      
Max ulp imag error (2.924,5.354) @ (7.979126851e-06,1.83067756) (0x3ee0bbc2d80da400,0x3ffd4a748da45a00)
        Ours = (0.0001208383539,-3.760891902)    Ref = (0.0001208383539,-3.760891902)  
        Ours = (0x3f1fad531cd4ca44,0xc00e164e7e6282af)               Ref = (0x3f1fad531cd4ca47,0xc00e164e7e6282b4)   
```

```
CPU fp32:
Max ulp real error (6.136,2.442) @ (-0.03077560104,-1.829890966)        (0xbcfc1d1d,0xbfea39de)
        Ours = (-0.4624662101,3.719158411)    Ref = (-0.4624663889,3.719158888)                                                       
        Ours = (0xbeecc85f,0x406e06b1)               Ref = (0xbeecc865,0x406e06b3)                                                    
                                                                                                                                      
Max ulp imag error (1.907,5.369) @ (0.004123169463,1.157435281e+13)     (0x3b871ba7,0x55286dd3)                 
        Ours = (0.008117808029,0.9842792749)    Ref = (0.008117809892,0.984279573)     
        Ours = (0x3c05008e,0x3f7bf9ba)               Ref = (0x3c050090,0x3f7bf9bf)  
```


s-oboyle:
/ok to test [0e1ebee](https://github.com/NVIDIA/cccl/pull/9036/commits/0e1ebee774ccafc960cc5bd508e891a3d33775fc)

libcudacxx/include/cuda/std/__complex/hyperbolic_functions.h
```diff
@@ -331,33 +331,117 @@ _CCCL_API inline complex<__nv_bfloat16> cosh(const complex<__nv_bfloat16>& __x)
...
 template <class _Tp>
-[[nodiscard]] _CCCL_API inline complex<_Tp> tanh(const complex<_Tp>& __x)
+[[nodiscard]] _CCCL_API inline complex<_Tp> tanh(const complex<_Tp>& __x) noexcept
 {
-  if (::cuda::std::isinf(__x.real()))
+  const _Tp __realx = ::cuda::std::fabs(__x.real());
+  _Tp __imagx       = __x.imag();
+
+  // __realx == inf overrides __imagx == nan:
+  if (__realx == numeric_limits<_Tp>::infinity())
```
miscco: I am wondering whether this should just be
```suggestion
  if (::cuda::std::isinf(__realx))
```

libcudacxx/include/cuda/std/__complex/hyperbolic_functions.h
```diff
@@ -331,33 +331,117 @@ _CCCL_API inline complex<__nv_bfloat16> cosh(const complex<__nv_bfloat16>& __x)
...
   {
-    return complex<_Tp>(__2rsh > _Tp(0) ? _Tp(1) : _Tp(-1), __2i > _Tp(0) ? _Tp(0) : _Tp(-0.));
+    // Set some very specific values to help with range reduction.
+    const _Tp __clamp_huge_values_bound = is_same_v<_Tp, double> ? static_cast<_Tp>(648.0) : static_cast<_Tp>(58.0f);
+
+    // Use two intervals to reduce values. Some values will be reduced twice.
+    const _Tp __large_interval_bound = is_same_v<_Tp, double> ? static_cast<_Tp>(334.0) : static_cast<_Tp>(34.0f);
+
+    // Has very low ulp error for c-r exp(__large_interval_subtract_factor).
+    const _Tp __large_interval_subtract_factor =
```
miscco: Nitpick: I believe those could all be `constexpr`

libcudacxx/include/cuda/std/__complex/hyperbolic_functions.h
```diff
@@ -331,33 +331,117 @@ _CCCL_API inline complex<__nv_bfloat16> cosh(const complex<__nv_bfloat16>& __x)
...
+    const _Tp __small_interval_mul_factor =
+      is_same_v<_Tp, double> ? static_cast<_Tp>(4.2808424752052536879e-137) : static_cast<_Tp>(2.6770937897e-11f);
+
+    // Clamp huge values, doesn't change result.
+    if (__realx_reduced > __clamp_huge_values_bound)
+    {
+      __realx_reduced = __clamp_huge_values_bound;
+    }
+
+    if (__realx_reduced >= __large_interval_bound)
```
miscco: Question: Should those rather be `else if` or is `__clamp_huge_values_bound` larger than `__large_interval_bound`

s-oboyle: It's a little hard to see in the templated code, but these all do (necessarily) cascade together (so very large values will hit all three of those `if`'s). Big values are painful :)

libcudacxx/include/cuda/std/__complex/hyperbolic_functions.h
s-oboyle: True

libcudacxx/include/cuda/std/__complex/hyperbolic_functions.h
s-oboyle: Ah, poorly named value on my part, `__realx` should be `__realx_abs` or somesuch.

libcudacxx/include/cuda/std/__complex/hyperbolic_functions.h
s-oboyle: Another check for extra high values might also be needed actually....

libcudacxx/include/cuda/std/__complex/hyperbolic_functions.h
```diff
@@ -331,33 +331,118 @@ _CCCL_API inline complex<__nv_bfloat16> cosh(const complex<__nv_bfloat16>& __x)
 // tanh
 
 template <class _Tp>
-[[nodiscard]] _CCCL_API inline complex<_Tp> tanh(const complex<_Tp>& __x)
+[[nodiscard]] _CCCL_API inline complex<_Tp> tanh(const complex<_Tp>& __x) noexcept
```
fbusato, line 336: very likely that we need to move it to `_CCCL_HOST_DEVICE_API`.

`_CCCL_API ` also means supporting Tile

libcudacxx/include/cuda/std/__complex/hyperbolic_functions.h
```diff
@@ -331,33 +331,118 @@ _CCCL_API inline complex<__nv_bfloat16> cosh(const complex<__nv_bfloat16>& __x)
...
+  // __real_x_abs == inf overrides __imag_x == nan:
+  if (__real_x_abs == numeric_limits<_Tp>::infinity())
   {
-    if (!::cuda::std::isfinite(__x.imag()))
-    {
-      return complex<_Tp>(::cuda::std::copysign(_Tp(1), __x.real()), _Tp(0));
-    }
-    return complex<_Tp>(::cuda::std::copysign(_Tp(1), __x.real()),
-                        ::cuda::std::copysign(_Tp(0), ::cuda::std::sin(_Tp(2) * __x.imag())));
+    __imag_x = ::cuda::std::copysign(_Tp{0}, __imag_x);
```
fbusato, line 344: I'm not sure I understand this point. Where the sign of 0 is meaningful in the code?

libcudacxx/include/cuda/std/__complex/hyperbolic_functions.h
```diff
@@ -331,33 +331,118 @@ _CCCL_API inline complex<__nv_bfloat16> cosh(const complex<__nv_bfloat16>& __x)
...
-  _Tp __d(::cuda::std::cosh(__2r) + __2i_cos);
-  _Tp __2rsh(::cuda::std::sinh(__2r));
-  if (::cuda::std::isinf(__2rsh) && ::cuda::std::isinf(__d))
+
+  const auto [__sin_imag, __cos_imag] = ::cuda::sincos(__imag_x);
+
+  // We can avoid under/overflow issues by scaling large real values
+  // appropriately and correcting afterwards.
+  // floor(log(MAX_FLOAT) / 4)
+  constexpr _Tp __reduction_needed_bound = is_same_v<_Tp, double> ? static_cast<_Tp>(177.0) : static_cast<_Tp>(22.0f);
```
fbusato: do we need `static_cast` here? it should not be a problem at least for the first branch. The second one depends if this function supports extended floating-point types or not

fbusato: apply to several other points in the code

libcudacxx/include/cuda/std/__complex/hyperbolic_functions.h
```diff
@@ -331,33 +331,118 @@ _CCCL_API inline complex<__nv_bfloat16> cosh(const complex<__nv_bfloat16>& __x)
...
+
+    if (__real_x_reduced > __reduction_needed_bound)
+    {
+      __imag_x_mul_factor *= __small_interval_mul_factor;
+      __real_x_reduced = __real_x_reduced + (__small_interval_subtract_factor * _Tp{0.5});
+    }
   }
-  return complex<_Tp>(__2rsh / __d, __2i_sin / __d);
+
+  const _Tp __expm1_2x = ::cuda::std::expm1(_Tp{2} * __real_x_reduced);
```
fbusato, line 408: do we need additional headers for `::cuda::std::expm1`?

libcudacxx/include/cuda/std/__complex/hyperbolic_functions.h
```diff
@@ -331,33 +331,118 @@ _CCCL_API inline complex<__nv_bfloat16> cosh(const complex<__nv_bfloat16>& __x)
...
+  const _Tp __cos_sq_4 = _Tp{4} * __cos_imag * __cos_imag;
+  const _Tp __denom    = ::cuda::std::fma(__expm1_2x, __expm1_2x, ::cuda::std::fma(__cos_sq_4, __expm1_2x, __cos_sq_4));
+  const _Tp __numer_real = ::cuda::std::fma(__expm1_2x, __expm1_2x, _Tp{2} * __expm1_2x);
```
fbusato, line 411: also `fma` is not covered by headers

libcudacxx/include/cuda/std/__complex/hyperbolic_functions.h
```diff
@@ -331,33 +331,118 @@ _CCCL_API inline complex<__nv_bfloat16> cosh(const complex<__nv_bfloat16>& __x)
...
+  }
+
+  const _Tp __ans_real = __numer_real * __denom_recip;
+  return complex<_Tp>{::cuda::std::copysign(__ans_real, __x.real()), __ans_imag};
 }
 
+// We have performance issues with extended floating point types
+#if _LIBCUDACXX_HAS_NVFP16()
+template <>
+_CCCL_API inline complex<__half> tanh(const complex<__half>& __x) noexcept
```
fbusato, line 434:
```suggestion
_CCCL_API inline complex<::__half> tanh(const complex<::__half>& __x) noexcept
```
applies to other cases

fbusato (changes requested):

libcudacxx/include/cuda/std/__complex/hyperbolic_functions.h
s-oboyle: You are quite correct, something has gone awry translating my usual non-template code into this template code. It needs a little more baking it seems.

libcudacxx/include/cuda/std/__complex/hyperbolic_functions.h
s-oboyle: This is true, and happy to remove, it was more just for keeping in line with the other functions. But I do like removing them.

libcudacxx/include/cuda/std/__complex/hyperbolic_functions.h
s-oboyle: Ah, so the `copysign` does matter, because we overwrite the answer at the end with this value sometimes:
``` 
if (__imag_x == _Tp{0})
  {
    __ans_imag = __imag_x;
  }
```
Without this we get a incorrectly signed 0 for some values eg: `tan(inf, -0)` returns `(1, 0)` instead of `(1, -0)`.

libcudacxx/include/cuda/std/__complex/hyperbolic_functions.h
s-oboyle: Worked this out with some constant tweaking, is better.

libcudacxx/include/cuda/std/__complex/hyperbolic_functions.h
s-oboyle: It seems to be included from elsewhere, but I added it anyway

libcudacxx/include/cuda/std/__complex/hyperbolic_functions.h
s-oboyle: same.

libcudacxx/include/cuda/std/__complex/hyperbolic_functions.h
s-oboyle: I'd rather update all these functions at the same time, I don't think main has this change yet?

libcudacxx/include/cuda/std/__complex/hyperbolic_functions.h
s-oboyle: Can we do this in one go across all maths functions instead? This is the last complex function that needed updating so we can do it straight after.

s-oboyle:
/ok to test [25c2be5](https://github.com/NVIDIA/cccl/pull/9036/commits/25c2be5896d04412bc82f0e9df25ec708f92ed6a)

libcudacxx/include/cuda/std/__complex/hyperbolic_functions.h
fbusato: yes, the change is already in the main branch. 80% of libcu++ functionalities support Tile

libcudacxx/include/cuda/std/__complex/hyperbolic_functions.h
fbusato: definitely. No problem with that. Just raising the issue

fbusato (approved):

libcudacxx/include/cuda/std/__complex/hyperbolic_functions.h
oleksandr-pavlyk: If `__imag_x` is NaN, the mapping to signed 0 depends on the value of sign-bit of NaN. 

This may be withing the spec, but please check. Python Array API says that sign of zero imaginary part is not specified: https://data-apis.org/array-api/draft/API_specification/generated/array_api.tanh.html

s-oboyle: It seems to be the same as the C++ spec ([https://en.cppreference.com/cpp/numeric/complex/tanh](https://en.cppreference.com/cpp/numeric/complex/tanh)) which has been explicitly tested against. Good to know about the python ones though.

PR #9040, opened 2026-05-15
Hardcode `DeviceTransform` offset type to I64 or U64

### LINKED ISSUE #8805: Consider hardcoding a 64-bit offset type for `DeviceTransform`
`cub::DeviceTransform` currently uses a `I32` or `I64` as offset type, depending on the offset type provided by the user. Using just `I64` would make the implementation simpler and should prevent 32-bit overflow bugs.

This performance difference between the two offset types was benchmarked a while ago on Hopper by @elstehle in #2714 and we saw differences of 0 - 3%. Several optimizations and changes have been made to `DeviceTransform` in the meantime, so I would suggest to revisit the issue, especially also benchmarking on Blackwell.

bernhardmgruber:
Fixes: #8805

### DIFF
--- a/cub/cub/device/dispatch/dispatch_transform.cuh
+++ b/cub/cub/device/dispatch/dispatch_transform.cuh
@@ -175,7 +175,7 @@ CUB_RUNTIME_FUNCTION _CCCL_FORCEINLINE int
 spread_out_items_per_thread(Offset num_items, Policy policy, int items_per_thread, int sm_count, int max_occupancy)
 {
   const int items_per_thread_evenly_spread = static_cast<int>(
-    (::cuda::std::min) (Offset{items_per_thread},
+    (::cuda::std::min) (static_cast<Offset>(items_per_thread),
                         ::cuda::ceil_div(num_items, sm_count * policy.threads_per_block * max_occupancy)));
   const int items_per_thread_clamped =
     ::cuda::std::clamp(items_per_thread_evenly_spread, policy.min_items_per_thread, policy.max_items_per_thread);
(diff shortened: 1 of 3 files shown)


miscco:
Should this go into 3.4?

cub/cub/device/dispatch/dispatch_transform.cuh
```diff
@@ -556,9 +556,9 @@ CUB_RUNTIME_FUNCTION _CCCL_FORCEINLINE static cudaError_t dispatch(
   KernelSource kernel_source             = {},
   KernelLauncherFactory launcher_factory = {})
 {
-  static_assert(
-    ::cuda::std::is_same_v<Offset, ::cuda::std::int32_t> || ::cuda::std::is_same_v<Offset, ::cuda::std::int64_t>,
-    "cub::DeviceTransform is only tested and tuned for 32-bit or 64-bit signed offset types");
+  // static_assert(
+  //   ::cuda::std::is_same_v<Offset, ::cuda::std::int32_t> || ::cuda::std::is_same_v<Offset, ::cuda::std::int64_t>,
+  //   "cub::DeviceTransform is only tested and tuned for 32-bit or 64-bit signed offset types");
```
miscco, line 561: Shouldnt this still be checking that we only use int64?

miscco (approved):

cub/cub/device/dispatch/dispatch_transform.cuh
bernhardmgruber: It will eventually just go away, since I want to replace the template parameter `Offset` by just the hardcoded 64-bit type that wins the race. I still need data for Blackwell and probably Ampere.

bernhardmgruber:
> Should this go into 3.4?

It's fine to ship later.

bernhardmgruber:
Forgot about this PR. Closing in favor of #9127

PR #9044, opened 2026-05-16
[libcu++] Fix default make_shared_resource construction

### LINKED ISSUE #9043: make_shared_resource<T>() default-constructs an empty shared_resource after shared_block_ptr rewrite
## Summary

`cuda::mr::make_shared_resource<T>()` no longer default-constructs `T` when called with no constructor arguments. Instead, the resulting `cuda::mr::shared_resource<T>` contains an empty `__shared_block_ptr`, so using the resource later asserts with `dereferencing a null __shared_block_ptr` in assertion-enabled builds.

This broke RAPIDS RMM's `cuda_async_managed_memory_resource`, which constructs its implementation with:

```cpp
cuda::mr::make_shared_resource<detail::cuda_async_managed_memory_resource_impl>()
```

First bad commit:
`207502e57019cefacf6f21d3bb6045aeebef2a3e`

Introduced by #8802.

## Minimal Reproducer

```cpp
#include <cuda/memory_resource>

#include <cstddef>
#include <iostream>

struct default_constructible_resource {
  default_constructible_resource() { ++constructed; }

  void* allocate_sync(std::size_t, std::size_t) { return this; }
  void deallocate_sync(void*, std::size_t, std::size_t) noexcept {}

  friend bool operator==(default_constructible_resource const&,
                         default_constructible_resource const&) noexcept
  {
    return true;
  }

  friend bool operator!=(default_constructible_resource const&,
                         default_constructible_resource const&) noexcept
  {
    return false;
  }

  static inline int constructed = 0;
};

int main()
{
  auto resource = cuda::mr::make_shared_resource<default_constructible_resource>();
  (void) resource;
  std::cout << "Constructed " << default_constructible_resource::constructed << " times.";
  return default_constructible_resource::constructed == 1 ? 0 : 1;
}
```

Expected: exits `0` because `default_constructible_resource` was constructed.
This is the behavior of CCCL 3.3.2: https://godbolt.org/z/vMGof94Px

Actual at `207502e57019cefacf6f21d3bb6045aeebef2a3e`: exits `1` because `default_constructible_resource` was not constructed.
This is the current behavior of CCCL trunk: https://godbolt.org/z/sGefhK1Kq

## Assertion-Enabled Symptom

If the reproducer calls `resource.allocate_sync(...)` and is built with `-DCCCL_ENABLE_ASSERTIONS`, it aborts with:

```text
cuda/__memory_resource/shared_block_ptr.h:123: __payload: Assertion `dereferencing a null __shared_block_ptr' failed.
```

## Suspected Cause

After `207502e57019cefacf6f21d3bb6045aeebef2a3e`, `shared_resource` stores:

```cpp
__shared_block_ptr<_Resource> __block_;
```

and constructs it with:

```cpp
: __block_(::cuda::std::forward<_Args>(__args)...)
```

When `_Args...` is empty, that selects `__shared_block_ptr()` default construction, which intentionally creates a null pointer, rather than creating a control block containing a default-constructed `_Resource`.

A zero-argument `make_shared_resource<T>()` should still construct a `T`.

bdice:
## Description

closes #9043

Adds coverage for zero-argument `cuda::mr::make_shared_resource<T>()` with a default-constructible resource, and fixes `shared_resource<T>(in_place_type<T>)` so it constructs a `T` even when no constructor arguments are provided.

The fix preserves null default construction for `__shared_block_ptr<T>` by adding an explicit in-place payload construction path.

Tested with:

```bash
CCCL_BUILD_INFIX=opencode cmake --preset libcudacxx-cpp17 -DCMAKE_CUDA_ARCHITECTURES=native
cmake --build /home/coder/cccl/build/opencode/libcudacxx-cpp17 --target libcudacxx.test.cuda.memory_resource.shared_resource -j
ctest --test-dir /home/coder/cccl/build/opencode/libcudacxx-cpp17 -R '^libcudacxx\.test\.cuda\.memory_resource\.shared_resource$' --output-on-failure
```

## Checklist

### DIFF
--- a/libcudacxx/include/cuda/__memory_resource/shared_block_ptr.h
+++ b/libcudacxx/include/cuda/__memory_resource/shared_block_ptr.h
@@ -76,7 +76,8 @@ public:
 
   //! @brief Constructs a new control block, forwarding arguments to the payload.
   _CCCL_TEMPLATE(class _Arg, class... _Rest)
-  _CCCL_REQUIRES((!::cuda::std::is_same_v<::cuda::std::remove_cvref_t<_Arg>, __shared_block_ptr>) )
+  _CCCL_REQUIRES((!::cuda::std::is_same_v<::cuda::std::remove_cvref_t<_Arg>, __shared_block_ptr>)
+                 && (!::cuda::std::is_same_v<::cuda::std::remove_cvref_t<_Arg>, ::cuda::std::in_place_type_t<_Payload>>) )
   _CCCL_HOST_API explicit __shared_block_ptr(_Arg&& __arg, _Rest&&... __rest)
       : __block_(new __block_t(::cuda::std::forward<_Arg>(__arg), ::cuda::std::forward<_Rest>(__rest)...))
   {}
--- a/libcudacxx/include/cuda/__memory_resource/shared_block_ptr.h
+++ b/libcudacxx/include/cuda/__memory_resource/shared_block_ptr.h
@@ -76,8 +76,9 @@ public:
 
   //! @brief Constructs a new control block, forwarding arguments to the payload.
   _CCCL_TEMPLATE(class _Arg, class... _Rest)
-  _CCCL_REQUIRES((!::cuda::std::is_same_v<::cuda::std::remove_cvref_t<_Arg>, __shared_block_ptr>)
-                 && (!::cuda::std::is_same_v<::cuda::std::remove_cvref_t<_Arg>, ::cuda::std::in_place_type_t<_Payload>>) )
+  _CCCL_REQUIRES(
+    (!::cuda::std::is_same_v<::cuda::std::remove_cvref_t<_Arg>, __shared_block_ptr>)
+    && (!::cuda::std::is_same_v<::cuda::std::remove_cvref_t<_Arg>, ::cuda::std::in_place_type_t<_Payload>>) )
   _CCCL_HOST_API explicit __shared_block_ptr(_Arg&& __arg, _Rest&&... __rest)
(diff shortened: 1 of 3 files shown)


Jacobfaib (approved):

libcudacxx/include/cuda/__memory_resource/shared_block_ptr.h
```diff
@@ -67,9 +68,17 @@ class __shared_block_ptr
...
+
   //! @brief Constructs a new control block, forwarding arguments to the payload.
   _CCCL_TEMPLATE(class _Arg, class... _Rest)
-  _CCCL_REQUIRES((!::cuda::std::is_same_v<::cuda::std::remove_cvref_t<_Arg>, __shared_block_ptr>) )
+  _CCCL_REQUIRES(
+    (!::cuda::std::is_same_v<::cuda::std::remove_cvref_t<_Arg>, __shared_block_ptr>)
+    && (!::cuda::std::is_same_v<::cuda::std::remove_cvref_t<_Arg>, ::cuda::std::in_place_type_t<_Payload>>) )
   _CCCL_HOST_API explicit __shared_block_ptr(_Arg&& __arg, _Rest&&... __rest)
       : __block_(new __block_t(::cuda::std::forward<_Arg>(__arg), ::cuda::std::forward<_Rest>(__rest)...))
   {}
```
miscco, line 84: I believe the additional constraint should not be necessary. We should probably just require everybody to use `in_place`

bdice: Happy to do this in a follow-up but we need to land the backport first to get CI unblocked for RAPIDS.

GCC 7 jobs failed without the additional constraint (ambiguous with two forwarding constructors).

PR #9053, opened 2026-05-18
Improve argument names in cub/util_math

pauleonix:
## Description
The size in bytes is part of the function names already and does not need to be repeated in the argument name. Especially not with differing capitalization which can cause confusion due to the regular interpretation of "B" = bytes and "b" = bits.

Since this PR only modifies argument names, there is no need to change call sites/this is a non-breaking change even for the public functions.

## Checklist


bernhardmgruber (approved):

pauleonix:
pre-commit.ci autofix

PR #9056, opened 2026-05-18
Vectorize contiguous iterators in `cub::BlockLoad`/`Store`

bernhardmgruber:


cub/cub/block/block_store.cuh
```diff
@@ -812,10 +812,9 @@ public:
     }
     else if constexpr (Algorithm == BLOCK_STORE_VECTORIZE)
     {
-      // FIXME(bgruber): we should test for contiguous iterator here
-      if constexpr (::cuda::std::is_pointer_v<OutputIteratorT>)
+      if constexpr (::cuda::std::contiguous_iterator<OutputIteratorT> && ::cuda::std::__can_to_address<OutputIteratorT>)
       {
-        StoreDirectBlockedVectorized(linear_tid, block_itr, items);
+        StoreDirectBlockedVectorized(linear_tid, ::cuda::std::to_address(block_itr), items);
```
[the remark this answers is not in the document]
miscco: Disregard: `__can_to_address` also works with non-C++20 contiguous iterators such as `thrust::pointer`

cub/cub/block/block_load.cuh
```diff
@@ -980,10 +980,10 @@ public:
       {
         InternalLoadDirectBlockedVectorized<RandomAccessIterator::__modifier>(linear_tid, block_src_it.ptr, dst_items);
       }
-      // FIXME(bgruber): we should test for contiguous iterator here
-      else if constexpr (::cuda::std::is_pointer_v<RandomAccessIterator>)
+      else if constexpr (::cuda::std::contiguous_iterator<RandomAccessIterator>
```
miscco, line 983: Important: We should also be able to vectorize loads of synthesizing iterators such as `counting_iterator` and `constant_iterator`

We probably want a trait for that though. Can you add a comment / issue

miscco (approved):

cub/cub/block/block_load.cuh
bernhardmgruber: I don't think this is a good idea, or I don't understand what you have in mind. 

Vectorized loading means I can check the iterators alignment and reinterpret it as a vector type for the purpose of issuing wider load instructions to fill an array of values in registers.

If I have a counting or constant iterator, I can just loop over it serially and fill my register array. I don't see the advantage of using vector types to do that.

PR #9058, opened 2026-05-18
Enable bugprone

Jacobfaib:
## Description

closes 

## Checklist


libcudacxx/include/cuda/__memory_resource/get_property.h
```diff
@@ -114,6 +114,10 @@ _CCCL_BEGIN_NAMESPACE_CPO(__forward_property)
 template <class _Derived, class _Upstream>
 struct __fn
 {
+private:
+  __fn() = default;
+  friend _Derived;
+
```
miscco, line 120: This is a bug because it turns everything private

libcudacxx/include/cuda/__numeric/add_overflow.h
```diff
@@ -346,11 +346,11 @@ _CCCL_API constexpr overflow_result<_ActualResult> add_overflow(const _Lhs __lhs
   else
   {
     // skip checks in cmp_less, cmp_greater, uabs
-    if constexpr (is_unsigned_v<_Lhs> && is_signed_v<_Lhs>)
+    if constexpr (is_unsigned_v<_Lhs> && is_signed_v<_Rhs>)
     {
       _CCCL_ASSUME(__rhs < 0);
     }
-    else if constexpr (is_unsigned_v<_Rhs> && is_signed_v<_Rhs>)
+    if constexpr (is_unsigned_v<_Rhs> && is_signed_v<_Lhs>)
```
miscco, line 353: This looks like a bug fix, please pull out into separate PR @davebayer for viz

libcudacxx/include/cuda/__utility/__basic_any/basic_any_ptr.h
```diff
@@ -172,7 +172,7 @@ struct _CCCL_TYPE_VISIBILITY_DEFAULT __basic_any<_Interface*>
   {
     __vptr_for<interface_type> __vptr = ::cuda::__get_vtable_ptr_for<interface_type, _Vp>();
     __ref_.__set_ref(__obj ? __vptr : nullptr, __obj);
-    return *static_cast<::cuda::std::__maybe_const<__is_const_ptr, _Vp>**>(static_cast<void*>(&__ref_.__optr_));
+    return *reinterpret_cast<::cuda::std::__maybe_const<__is_const_ptr, _Vp>**>(&__ref_.__optr_);
```
miscco, line 175: Critical: This is not the same, please revert

libcudacxx/include/cuda/__utility/__basic_any/basic_any_value.h
```diff
@@ -86,8 +86,9 @@ struct _CCCL_TYPE_VISIBILITY_DEFAULT __basic_any : __basic_any_base<_Interface>
                  // The following constraint seems to trip up a lot of compilers, so leave it out.
                  // _CCCL_AND ::cuda::std::constructible_from<_Up, _Tp>
                  _CCCL_AND __satisfies<_Up, _Interface>)
-  _CCCL_HOST_DEVICE_API __basic_any(_Tp&& __value) noexcept(
-    __is_small<_Up, __movable>(__size_, __align_) && ::cuda::std::is_nothrow_constructible_v<_Up, _Tp>)
+  _CCCL_HOST_DEVICE_API __basic_any( // NOLINT(bugprone-forwarding-reference-overload)
```
miscco, line 89: Please disable the `bugprone-forwarding-reference-overload` we use that all the time

libcudacxx/include/cuda/std/__algorithm/min.h
```diff
@@ -31,14 +31,15 @@ _CCCL_BEGIN_NAMESPACE_CUDA_STD
 
 _CCCL_EXEC_CHECK_DISABLE
 template <class _Tp, class _Compare>
-[[nodiscard]] _CCCL_API constexpr const _Tp& min(const _Tp& __a, const _Tp& __b, _Compare __comp)
+[[nodiscard]] _CCCL_API constexpr const _Tp&
+min(const _Tp& __a _CCCL_LIFETIMEBOUND, const _Tp& __b _CCCL_LIFETIMEBOUND, _Compare __comp)
```
miscco, line 35: Please pull that out

libcudacxx/include/cuda/std/__cmath/logarithms.h
```diff
@@ -232,6 +232,8 @@ template <class _Tp>
 [[nodiscard]] _CCCL_API inline constexpr int __ilogb_impl(_Tp __x) noexcept
 {
   const auto __fp = ::cuda::std::fpclassify(__x);
+  // FP_ILOGB0 and FP_ILOGBNAN may have the same value, but the cases are semantically distinct.
+  // NOLINTBEGIN(bugprone-branch-clone)
```
miscco, line 236: Strongly considering dropping this

miscco (changes requested):

miscco:
Please also drop the sizeof check

libcudacxx/include/cuda/__numeric/add_overflow.h
Jacobfaib: @miscco This PR is not for review (at least not in its current state). It's the umbrella PR that collects all of the changes needed for the whole category. Once CI passes:

1. I split each check into separate PRs.
2. I revert all changes in this PR and enable the category in the root `.clang-tidy`, while disabling all the checks that need fixes.
3. This PR (in reduced form) merges.
4. Each separate fix PR then removes the disable and implements the fixes. These sub-PR's is where these kinds of comments should be posted, they'll just get lost here.

Jacobfaib: This is what this PR will (eventually) look like https://github.com/NVIDIA/cccl/pull/8438.

libcudacxx/include/cuda/__utility/__basic_any/basic_any_ptr.h
Jacobfaib: So I deliberated quite a bit in this change when I saw the clang-tidy diagnostic, and it is my understanding that in this scenario, they are the same. But happy to be proved wrong.

Based on my reading of section 5.2.10, reinterpret_cast is equivalent to a double static_cast through void if and only if you cast a type there and back again (the bilbo baggins rule), but never inspect the intermediate.

This is what any does, as it ensures the type is cast to the original type with RTTI.

But the code is complex, so maybe I missed something.

PR #9064, opened 2026-05-18
`cuda::simd` Dot product

fbusato:
## Description

Introduce SIMD `idot` to compute the dot product of two integer vectors. 
The operation maps to `IDP4`/`IDP2` HW instructions.

### Use cases

- INT8 quantized inference
- Mixed-precision INT16 × INT8 GEMM/conv
- Image processing (convolution, sum of squared differences, histogram).
- Checksum algorithms (e.g. CRC)

Requires:

- https://github.com/NVIDIA/cccl/pull/8994


libcudacxx/include/cuda/__simd/simd_intrinsics_array.h
```diff
@@ -0,0 +1,207 @@
...
+    {
+      __result = ::cuda::simd::__dp4a_s8x4_u8x4(__lhs_u[__i], __rhs_u[__i], __result);
+    }
+    else
+    {
+      __result = ::cuda::simd::__dp4a_s8x4_s8x4(__lhs_u[__i], __rhs_u[__i], __result);
+    }
+  }
+  return __result;
+}
```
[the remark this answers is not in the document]
miscco: That is actually a goood comment, I believe we should static assert that `_AccumT` is integral and has size 4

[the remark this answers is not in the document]
fbusato: Addressed in 94b9aefff4. Both packed intrinsic helpers now require a 32-bit integral accumulator, while the public `idot` dispatch still permits wider accumulators through the scalar fallback. Verified with the 13-configuration compiler matrix and SM80/SM120 IDP2A/IDP4A codegen checks.

libcudacxx/test/libcudacxx/std/numerics/simd/simd.non_std/idot.pass.cpp
```diff
@@ -0,0 +1,207 @@
...
+{
+  {
+    cuda::std::array<int8_t, 4> lhs_values{-8, -3, 2, 7};
+    cuda::std::array<int8_t, 4> rhs_values{4, -5, 6, -7};
+    test_values<int8_t, int8_t, int32_t, 4>(lhs_values, rhs_values, int32_t{11});
+  }
+  {
+    cuda::std::array<uint8_t, 7> lhs_values{1, 2, 3, 4, 5, 6, 7};
+    cuda::std::array<uint8_t, 7> rhs_values{8, 7, 6, 5, 4, 3, 2};
+    test_values<uint8_t, uint8_t, uint32_t, 7>(lhs_values, rhs_values, uint32_t{13});
```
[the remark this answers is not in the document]
fbusato: Addressed in 70655829dd. The seven-element unsigned DP4A case now sets high bits in both operands (including 128, 200, and 255), so incorrect signed intrinsic selection changes the result while tail handling remains covered. The focused runtime/constexpr lit test and pre-commit checks pass.

fbusato:
@miscco @davebayer the current PR covers the *integer* dot product `cuda::simd::idot()`.
On the other hand, would make sense to generalize it to other types, i.e. floating-point types, to cover more cases and optimization opportunities. The function would be named `cuda::simd::dot()`. What do you think about? (C++ also uses `inner_product` for similar purposes)

docs/libcudacxx/extended_api/simd/idot.rst
```diff
@@ -0,0 +1,81 @@
...
+
+   } // namespace cuda::simd
+
+The function ``cuda::simd::idot`` computes the dot product of two integer ``cuda::std::simd::basic_vec`` objects and adds the result to an accumulator.
+
+For each element ``i`` in the input vectors, the result is equivalent to:
+
+.. code:: cuda
+
+   acc += static_cast<AccT>(lhs[i]) * static_cast<AccT>(rhs[i])
```
miscco: I am a bit puzzled about the docuumentation here. Is `Acc` meant to be a scalar or a simd vector?

If its a scalar the current formulation is a bit odd

docs/libcudacxx/extended_api/simd/idot.rst
```diff
@@ -0,0 +1,81 @@
...
+.. code:: cuda
+
+   namespace cuda::simd {
+
+   template <class T, class U, class Abi, class AccT>
+   [[nodiscard]] __host__ __device__ constexpr
+   AccT idot(
+     const cuda::std::simd::basic_vec<T, Abi>& lhs,
+     const cuda::std::simd::basic_vec<U, Abi>& rhs,
+     AccT acc) noexcept;
```
miscco: I am a bit unhappy about the accumulator being passed explicitly, because the same code could be written as

`acc + idot(veca, vecb)` but at the same time it would be more flexible because we do not hard-code `+` in that expression.

Could it be that the idea bout the accumulator is to force the return type?

docs/libcudacxx/extended_api/simd/idot.rst
fbusato: yes, it is a scalar. Do you mean I should add the `for` loop?

docs/libcudacxx/extended_api/simd/idot.rst
fbusato: ah, `AccT acc` the name is confusing! It should have instead the same meaning of `init` in `std::accumulate` https://en.cppreference.com/cpp/algorithm/accumulate. I have to replace it

libcudacxx/include/cuda/__simd/idot.h
```diff
@@ -0,0 +1,124 @@
...
+  [[nodiscard]] _CCCL_HOST_DEVICE_API constexpr _AccumT
+  operator()(const _LhsStorage& __lhs, const _RhsStorage& __rhs, const _AccumT __init) const noexcept
+  {
+#if _CCCL_HAS_SIMD_IDOT()
+    _CCCL_IF_NOT_CONSTEVAL_DEFAULT
+    {
+      using ::cuda::std::is_signed_v;
+      using ::cuda::std::is_unsigned_v;
+      constexpr bool __is_unsigned_dot   = is_unsigned_v<_Tp> && is_unsigned_v<_Up> && is_unsigned_v<_AccumT>;
+      constexpr bool __is_signed_dot     = (is_signed_v<_Tp> || is_signed_v<_Up>) && is_signed_v<_AccumT>;
```
davebayer: Is this correct? Won't unsigned input values like `0xff` get sign-extended to `0xffff'ffff` instead of staying `0xff`?

davebayer (changes requested):

libcudacxx/include/cuda/__simd/idot.h
fbusato: yes, it looks correct. Not sure where the sign-extend could happen.  `_Tp`, `_Up` are the original scalar types, while `_LhsStorage` and `_RhsStorage` (`__lhs`, `__rhs`) are the actual SIMD storage, e.g. packed `U8` values

davebayer: Ah yeah, you are right

docs/libcudacxx/extended_api/simd/idot.rst
```diff
@@ -0,0 +1,85 @@
...
+- ``T``, ``U``, and ``AccT`` must be integer types.
+
+**Performance considerations**
+
+- Packed 8-bit input vectors with compatible 32-bit accumulators use ``IDP4A`` on ``SM61`` and newer device targets.
+- Packed 16-bit by 8-bit input vectors with compatible 32-bit accumulators use ``IDP2A`` on ``SM61`` and newer device
+  targets.
+- A compatible accumulator is unsigned when both inputs are unsigned and signed when either input is signed.
+- Other integer input and accumulator combinations use the scalar fallback.
+
```
davebayer: I miss a section describing overflow behaviour - what if the dot product + `init` overflows. Well defined for unsigned types, but undefined for signed integer types?

docs/libcudacxx/extended_api/simd/idot.rst
```diff
@@ -0,0 +1,85 @@
...
+.. code:: cuda
+
+   namespace cuda::simd {
+
+   template <class T, class U, class Abi, class AccT>
+   [[nodiscard]] __host__ __device__ constexpr
+   AccT idot(
+     const cuda::std::simd::basic_vec<T, Abi>& lhs,
+     const cuda::std::simd::basic_vec<U, Abi>& rhs,
+     AccT init) noexcept;
```
davebayer: Would it make sense to provide an overload that would omit the `init` parameter? It would make sense in my opinion

docs/libcudacxx/extended_api/simd/idot.rst
```diff
@@ -0,0 +1,85 @@
...
+    for (size_t i = 0; i < lhs.size(); ++i) {
+        acc += static_cast<AccT>(lhs[i]) * static_cast<AccT>(rhs[i]);
+    }
+    return acc;
+
+**Parameters**
+
+- ``lhs``: The left-hand side input vector.
+- ``rhs``: The right-hand side input vector.
+- ``acc``: The initial accumulator value.
```
davebayer:
```suggestion
- ``init``: The initial accumulator value.
```

libcudacxx/include/cuda/__simd/idot.h
```diff
@@ -0,0 +1,124 @@
...
+    _AccumT __result      = __init;
+    _CCCL_PRAGMA_UNROLL_FULL()
+    for (::cuda::std::simd::__simd_size_type __i = 0; __i < __size; ++__i)
+    {
+      const auto __lhs_value = static_cast<_AccumT>(__lhs.__data[__i]);
+      const auto __rhs_value = static_cast<_AccumT>(__rhs.__data[__i]);
+      const auto __product   = static_cast<_AccumT>(__lhs_value * __rhs_value);
+      __result               = __result + __product;
+    }
+    return __result;
```
davebayer: Why is the fallback implementation not vectorized? We can do vectorized multiplication and vectorized add reduction

docs/libcudacxx/extended_api/simd/idot.rst
fbusato: what would be the return type in this case? `std::common_type`? `decltype(a0*b0+...)`? we had a similar problem with CUB/Thrust

libcudacxx/include/cuda/__simd/idot.h
fbusato: what do you mean? SIMD vector operations instead of scalar for loop?

docs/libcudacxx/extended_api/simd/idot.rst
fbusato: Overflow behavior can never happen in the cases of PTX and CUDA math intrinsics. I would keep the standard C++ behavior for the generic path

libcudacxx/include/cuda/__simd/idot.h
davebayer: Yes

docs/libcudacxx/extended_api/simd/idot.rst
davebayer: Okay, I see, we can add it later if needed

docs/libcudacxx/extended_api/simd/idot.rst
davebayer: Yeah, I just feel like we should be explicit about this

libcudacxx/include/cuda/__simd/idot.h
fbusato: it is not (directly) possible. Here we need to convert to `_AccumT`, then multiply/add. Vectorization would truncate the results.

docs/libcudacxx/extended_api/simd/idot.rst
miscco: I am still not sure whether this does not restrict the algorithm.

In most cases the additional 0 will just be noise and you cannot use any other operation than plus.

The only situation where this might be valid is overflow protection, but also that only partially.

Not a hill to die though

libcudacxx/include/cuda/__simd/simd_intrinsics.h
```diff
@@ -31,73 +31,291 @@
...
+  NV_IF_TARGET(
+    NV_PROVIDES_SM_61,
+    (return ::__dp4a(static_cast<::cuda::std::int32_t>(__lhs), static_cast<::cuda::std::int32_t>(__rhs), __init);))
+#    elif _CCCL_HAS_SIMD_IDOT_PTX()
+  NV_IF_TARGET(NV_PROVIDES_SM_61, ({
+                 ::cuda::std::int32_t __result{};
+                 asm("dp4a.s32.s32 %0, %1, %2, %3;" : "=r"(__result) : "r"(__lhs), "r"(__rhs), "r"(__init));
+                 return __result;
+               }))
+#    endif // _CCCL_HAS_SIMD_IDOT_INTRINSICS() || _CCCL_HAS_SIMD_IDOT_PTX()
```
davebayer, line 137: I would prefer not to care about pre sm_61 architectures to be honest

fbusato: definitely, we can just say device code

docs/libcudacxx/extended_api/simd/idot.rst
fbusato: I agree that would make sense to always start from 0 implicitly. On the other hand, the main reason is useful is to set the return/accumulation type without writing the template parameter explicitly, which I think it is not really intuitive.

fbusato: what about an overload with `common_type`?

miscco: nah, lets keep it as is

docs/libcudacxx/extended_api/simd/dot.rst
```diff
@@ -0,0 +1,91 @@
...
+
+        cuda::std::array<float, 4> lhs_values{1.0f, 2.0f, 3.0f, 4.0f};
+        cuda::std::array<float, 4> rhs_values{5.0f, 6.0f, 7.0f, 8.0f};
+        vec_t lhs(lhs_values);
+        vec_t rhs(rhs_values);
+
+        float result = cuda::simd::dot(lhs, rhs, 10.0f);
+
+        assert(result == 80.0f);
+    }
```
miscco, line 90: I would like to have an example where we intentionally widen the init type for smaller integers

libcudacxx/include/cuda/__simd/dot.h
```diff
@@ -0,0 +1,119 @@
...
+//! @brief Computes a dot product and adds it to an accumulator.
+//! @param[in] __lhs The left-hand side input vector.
+//! @param[in] __rhs The right-hand side input vector.
+//! @param[in] __init The initial accumulator value.
+//! @return The accumulator plus the dot product of the input vectors.
+template <typename _Tp, typename _Up, typename _Abi, typename _AccumT>
+[[nodiscard]] _CCCL_HOST_DEVICE_API constexpr _AccumT
+dot(const ::cuda::std::simd::basic_vec<_Tp, _Abi>& __lhs,
+    const ::cuda::std::simd::basic_vec<_Up, _Abi>& __rhs,
+    const _AccumT __init) noexcept
```
miscco: We discussed this internally, Could we default `_AccumT` to `_Tp` and then default `__init = {}`

I believe that would give us the best of both worlds

libcudacxx/include/cuda/__simd/simd_intrinsics_array.h
```diff
@@ -79,6 +80,105 @@ template <typename _Tp, ::cuda::std::size_t _Np>
...
+      __result = ::cuda::simd::__dp4a_u8x4_s8x4(__lhs_u[__i], __rhs_u[__i], __result);
+    }
+    else if constexpr (::cuda::std::is_unsigned_v<_Up>)
+    {
+      __result = ::cuda::simd::__dp4a_s8x4_u8x4(__lhs_u[__i], __rhs_u[__i], __result);
+    }
+    else
+    {
+      __result = ::cuda::simd::__dp4a_s8x4_s8x4(__lhs_u[__i], __rhs_u[__i], __result);
+    }
```
griwes: Normally I don't like nesting, but here I think it'd help to go with `if (is_unsigned_v<_Tp>) { if (is_unsigned_v<_Up>) { ...`, because then you don't have to mentally track all four options across all the branches.

libcudacxx/include/cuda/__simd/simd_intrinsics_array.h
```diff
@@ -79,6 +80,105 @@ template <typename _Tp, ::cuda::std::size_t _Np>
...
+  static_assert(::cuda::std::is_integral_v<_AccumT> && sizeof(_AccumT) == 4,
+                "cuda::simd::__dp2a_16bit_x2_8bit_x4: dot intrinsic path requires a 32-bit integral accumulator");
+  static_assert(_N16Bit == 2 * _N8Bit || _N16Bit + 1 == 2 * _N8Bit,
+                "cuda::simd::__dp2a_16bit_x2_8bit_x4: Input vector sizes must match");
+  _AccumT __result = __init;
+  _CCCL_PRAGMA_UNROLL_FULL()
+  for (::cuda::std::size_t __i = 0; __i < _N16Bit; ++__i)
+  {
+    const auto __rhs_u = __rhs_u8[__i / 2];
+    if constexpr (::cuda::std::is_unsigned_v<_Tp> && ::cuda::std::is_unsigned_v<_Up>)
```
griwes: Same here, except here the branches are longer and it's easy for the earlier conditions to scroll off.

griwes (commented):
Could the filecheck tests maybe use the `%PARAM%` machinery to avoid restating the same function, just with different types, over and over again? Seems to me the pattern is fairly regular and the "tuple"-ish `%PARAM%` values I've added for atomics would let you easily specify both the type and the instruction segment at the same time.

libcudacxx/include/cuda/__simd/simd_intrinsics_array.h
fbusato: ok, let me try

fbusato: yes, much better now

fbusato:
> Could the filecheck tests maybe use the %PARAM% machinery to avoid restating the same function, just with different types, over and over again? Seems to me the pattern is fairly regular and the "tuple"-ish %PARAM% values I've added for atomics would let you easily specify both the type and the instruction segment at the same time.

now, the codegen checks are not repeated. On the other hand, I find the `%PARAM%` section a bit hard to read, but that's fine overall

griwes (approved):

miscco (approved):

davebayer (approved):

PR #9065, opened 2026-05-19
Env passthrough 4/4

### LINKED ISSUE #8175: Final passthrough for all env APIs

gonidelis:
closes #8175 

examines 

- `DeviceMergeSort`
- `DeviceHistogram`
- `DeviceScan`
- `DeviceSegmentedScan`

ambiguities found. important to backport.

`DeviceTopK` is still open so we cannot examine it here in order not to create conflicts.

### DIFF
--- a/cub/cub/device/device_merge_sort.cuh
+++ b/cub/cub/device/device_merge_sort.cuh
@@ -20,7 +20,6 @@
 
 #include <cuda/__execution/require.h>
 #include <cuda/std/__execution/env.h>
-
 CUB_NAMESPACE_BEGIN
 
 /**
--- a/cub/cub/device/device_merge_sort.cuh
+++ b/cub/cub/device/device_merge_sort.cuh
@@ -20,6 +20,7 @@
 
 #include <cuda/__execution/require.h>
 #include <cuda/std/__execution/env.h>
+
 CUB_NAMESPACE_BEGIN
 
 /**
--- a/cub/cub/device/device_scan.cuh
+++ b/cub/cub/device/device_scan.cuh
@@ -769,13 +769,15 @@ struct DeviceScan
   //!   @rst
   //!   **[optional]** Execution environment. Default is ``cuda::std::execution::env{}``.
   //!   @endrst
-  template <typename InputIteratorT,
-            typename OutputIteratorT,
-            typename ScanOpT,
-            typename InitValueT,
-            typename NumItemsT,
-            typename EnvT                                                        = ::cuda::std::execution::env<>,
-            ::cuda::std::enable_if_t<::cuda::std::is_integral_v<NumItemsT>, int> = 0>
+  template <
+    typename InputIteratorT,
+    typename OutputIteratorT,
+    typename ScanOpT,
+    typename InitValueT,
+    typename NumItemsT,
+    typename EnvT                 = ::cuda::std::execution::env<>,
+    ::cuda::std::enable_if_t<::cuda::std::is_integral_v<NumItemsT> && !::cuda::std::is_same_v<OutputIteratorT, size_t>,
+                             int> = 0>
   [[nodiscard]] CUB_RUNTIME_FUNCTION static cudaError_t ExclusiveScan(
     InputIteratorT d_in,
     OutputIteratorT d_out,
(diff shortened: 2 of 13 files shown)


cub/cub/device/device_merge_sort.cuh
```diff
@@ -20,7 +20,6 @@
 
 #include <cuda/__execution/require.h>
 #include <cuda/std/__execution/env.h>
-
```
bernhardmgruber, line 23: Suggestion: I would retain the empty line here.

cub/cub/device/device_scan.cuh
```diff
@@ -769,13 +769,15 @@ struct DeviceScan
...
-            ::cuda::std::enable_if_t<::cuda::std::is_integral_v<NumItemsT>, int> = 0>
+  template <
+    typename InputIteratorT,
+    typename OutputIteratorT,
+    typename ScanOpT,
+    typename InitValueT,
+    typename NumItemsT,
+    typename EnvT                 = ::cuda::std::execution::env<>,
+    ::cuda::std::enable_if_t<::cuda::std::is_integral_v<NumItemsT> && !::cuda::std::is_same_v<OutputIteratorT, size_t>,
+                             int> = 0>
```
bernhardmgruber: Suggestion: I think @miscco stated his preference of having each constraint separately. I am fine with the version here. Just FYI.
```suggestion
    ::cuda::std::enable_if_t<::cuda::std::is_integral_v<NumItemsT>, int> = 0>,
    ::cuda::std::enable_if_t<!::cuda::std::is_same_v<OutputIteratorT, size_t>, int> = 0
```

cub/test/catch2_test_device_segmented_scan_env_api.cu
```diff
@@ -54,13 +55,13 @@ C2H_TEST("cub::DeviceSegmentedScan::ExclusiveSegmentedScan accepts stream", "[se
   cuda::stream_ref stream_ref{stream};
 
   auto error = cub::DeviceSegmentedScan::ExclusiveSegmentedScan(
-    d_in.begin(), d_out.begin(), d_offsets_it, d_offsets_it + 1, num_segments, ::cuda::std::plus<>{}, 100, stream_ref);
+    d_in.begin(), d_out.begin(), d_offsets_it, d_offsets_it + 1, num_segments, cuda::maximum<>{}, 4, stream_ref);
```
bernhardmgruber, line 58: Question: I am curious what prompted the change from `plus` to `maximum`?

cub/test/catch2_test_device_scan_api.cu
```diff
@@ -38,3 +38,131 @@ C2H_TEST("Device inclusive scan works", "[scan][device]")
 
   REQUIRE(expected == out);
 }
+
+C2H_TEST("cub::DeviceScan::ExclusiveSum non-env API", "[scan][device]")
```
bernhardmgruber: Suggestion: the tests look a bit dull without knowing what they cover. Let's rename the tests so something like:

```suggestion
C2H_TEST("cub::DeviceScan::ExclusiveSum non-env overload is not ambiguous", "[scan][device]")
```
Applies to all tests here.

cub/cub/device/device_scan.cuh
gonidelis: nah i like mike's thing more

cub/cub/device/device_merge_sort.cuh
gonidelis: Done — empty line restored.

cub/test/catch2_test_device_segmented_scan_env_api.cu
gonidelis: just super safety guard. Old test (plus, init=100) would silently pass if dispatch dropped op/init and fell through to a segemntedscansum. also use non cualified cuda namespace as part

cub/test/catch2_test_device_histogram_api.cu
```diff
@@ -0,0 +1,123 @@
...
+#include <c2h/catch2_test_helper.h>
+
+C2H_TEST("cub::DeviceHistogram::HistogramEven non-env overload is not ambiguous", "[histogram][device]")
+{
+  thrust::device_vector<int> samples(1);
+  thrust::device_vector<int> histogram(1);
+  size_t temp_storage_bytes = 0;
+  cub::DeviceHistogram::HistogramEven(
+    nullptr, temp_storage_bytes, samples.begin(), thrust::raw_pointer_cast(histogram.data()), 2, 0, 10, 1);
+}
```
bernhardmgruber, line 19: May need to be revisited depending on the outcome of #9078

bernhardmgruber (approved):

PR #9068, opened 2026-05-19
[cub] Fix compilation warnings with `nvc++`

davebayer:
This PR fixes compilation warnings emitted during CUB compilation with `nvc++`.


cub/cub/detail/fast_modulo_division.cuh
```diff
@@ -106,7 +106,16 @@ multiply_extract_higher_bits(T value, R multiplier)
...
+#if _CCCL_CUDA_COMPILER(NVHPC)
+        ? [value, multiplier](){
+          unsigned long long result;
+          asm("mul.hi.u64 %0, %1, %2;" : "=l"(result) : "l"(value), "l"(multiplier));
+          return static_cast<unsigned_t>(result);
+        }()
+#else // ^^^ _CCCL_CUDA_COMPILER(NVHPC) ^^^ / vvv !_CCCL_CUDA_COMPILER(NVHPC) vvv
         ? static_cast<unsigned_t>(__umul64hi(value, multiplier))
+#endif // ^^^ !_CCCL_CUDA_COMPILER(NVHPC) ^^^
         : static_cast<unsigned_t>((static_cast<larger_t>(value) * multiplier) >> NumBits);}));
```
miscco, line 119: Critical: We use conditional compilation inside an NV_IF_TARGET. you need to wrap outside of it

miscco (changes requested):

cub/test/catch2_test_device_transform.cu
```diff
@@ -726,8 +726,12 @@ __global__ void unrelated_kernel()
 {
   __shared__ int ssmem; // 4 bytes
   extern __shared__ int dsmem[]; // aligned to 16 by default, so 12 bytes padding needed
-  asm("" : "+r"(ssmem));
-  asm("" : "+r"(dsmem[0]));
+  if (threadIdx.x == 0)
+  {
+    ssmem = 0;
+  }
```
bernhardmgruber, line 732: Suggestion: can you please add a comment why this is necessary? And that it's because of nvc++.

libcudacxx/include/cuda/__cmath/mul_hi.h
```diff
@@ -100,7 +100,16 @@ _CCCL_API constexpr _Tp mul_hi(_Tp __lhs, _Tp __rhs) noexcept
       {
         [[maybe_unused]] const auto __lhs1 = static_cast<unsigned long long>(__lhs);
         [[maybe_unused]] const auto __rhs1 = static_cast<unsigned long long>(__rhs);
+        // nvc++ doesn't implement __umul64hi and crashes, so we replace it with inline PTX
```
bernhardmgruber, line 103: Suggestion: Consider filing a bug report

cub/cub/agent/agent_histogram.cuh
```diff
@@ -307,9 +307,15 @@ struct AgentHistogram
         {
           if (bins[pixel] >= 0)
           {
+            // nvc++ has a bug that it does not treat atomicAdd_block as a builtin and ptxas fails due to unresolved
+            // symbol.
+#if _CCCL_CUDA_COMPILER(NVHPC)
+            atomicAdd(privatized_histograms[ch] + bins[pixel], accumulator);
```
bernhardmgruber, line 313: Q: Why hasn't this come up yet? Or: why hasn't this been reported by the nvc++ team yet? What change makes this a problem now? I want to understand where this is coming from.

libcudacxx/include/cuda/__cmath/mul_hi.h
davebayer: I've already told @dkolsen-pgi

cub/cub/agent/agent_histogram.cuh
davebayer: Probably because they don't use histogram. And we don't use it anywhere else

jrhemstad (changes requested):
I'd like to discuss the motivation for this before adding additional complexity to CUB's source code.

PR #9069, opened 2026-05-19
[libcu++] Explicitly mark `__half` and `__nv_bfloat16` as trivially_copyable

miscco:
Currently we only mark the vector types as trivially_copyable. However, for the machinery to work we actually need the type to be defined.

Rather than forcing internal headers to include `cuda_fp16.h` we should just specialize the trait directly

### DIFF
--- a/libcudacxx/include/cuda/__type_traits/is_trivially_copyable.h
+++ b/libcudacxx/include/cuda/__type_traits/is_trivially_copyable.h
@@ -47,12 +47,17 @@ inline constexpr bool __is_trivially_copyable_v =
 #if _CCCL_HAS_NVFP16()
 
 template <>
+inline constexpr bool __is_trivially_copyable_v<::__half> = true;
+
+template <>
 inline constexpr bool __is_trivially_copyable_v<::__half2> = true;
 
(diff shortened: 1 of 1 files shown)


bernhardmgruber (approved):

bernhardmgruber:
@miscco I think @coderabbitai has a point, we may want to add a test that just statically asserts that the trivially copyable holds.

fbusato (changes requested):
`__half` and `__nv_bfloat16` are already trivially copyable

miscco:
> `__half` and `__nv_bfloat16` are already trivially copyable

that is correct, but that relies on the compiler sussing it out and it does not work if e.g `cuda_fp16.h` is not included

fbusato (approved):

davebayer (approved):

miscco:
> @miscco I think @coderabbitai has a point, we may want to add a test that just statically asserts that the trivially copyable holds.

there is already a test, the issue was that the test includes `cuda_fp16.h` and friends and I want that trait to work with them being forward declared.

This is more of an evil use case though

PR #9072, opened 2026-05-19
[Tile] Improve support for `complex`, `__half` and `__nv_bfloat16`

miscco:
`complex` is not properly supported currently due to most math functions not being available. 

However, we can get basic suport by enabling just the right ones.

Also mark cmath functions that are not supported as `_CCCL_HOST_DEVICE_API`


docs/libcudacxx/tile.rst
```diff
@@ -105,7 +105,8 @@ CUDA extended floating point types
 
 Tile programs treat the CUDA extended floating point types as compiler builtin types.
 This disallows accessing their internals which we require internally.
-Support for extended floating point types such as ``__half``, ``__nv_bfloat16`` is disabled in tile mode.
+Support for extended floating point types such as ``__half``, ``__nv_bfloat16`` is only partial in libcu++.
+Support for extended floating point types of size 8, 6 and 4 is disabled.
```
fbusato:
```suggestion
Support for extended floating point types of size 8, 6 and 4-bit is disabled.
```

libcudacxx/test/libcudacxx/std/numerics/bit/bit.cast/bit_cast.trivially_copyable.pass.cpp
```diff
@@ -84,7 +84,7 @@ TEST_FUNC bool tests()
     test_roundtrip_through_buffer(i);
   }
 
-#if _LIBCUDACXX_HAS_NVFP16()
+#if _LIBCUDACXX_HAS_NVFP16() && !_CCCL_TILE_COMPILATION()
```
fbusato, line 87: wondering if makes  sense to add `_CCCL_TILE_COMPILATION()` to `_LIBCUDACXX_HAS_NVFP16`

fbusato (approved):

libcudacxx/test/libcudacxx/std/numerics/bit/bit.cast/bit_cast.trivially_copyable.pass.cpp
miscco: No because we do want to support them, but we do not have support for many math functions

bernhardmgruber (approved):

PR #8875, opened 2026-05-08
Argument annotation framework for segmented algorithms

pciolkosz:
## Summary

Introduces a flexible argument annotation framework that allows users to annotate algorithm parameters with compile-time and runtime bounds, enabling dispatch-layer optimizations without combinatorial explosion of function overloads.

This is a prototype implementation of [Option 1 from #7495](https://github.com/NVIDIA/cccl/issues/7495): annotating parameters with wrappers on the parameter itself.

Resolves #7495

### Three argument wrappers + plain values

- **Plain values** work as-is — no wrapper required. A scalar is a single value, a pointer/span is a collection.
- **`argument::deferred{indirect_value, bounds...} and argument::deferred_sequence{sequence, bounds...}`** — device-resident value not accessible at API call time. User explicitly specifies if the argument holds a single value or a sequence with the wrapper type. Host-side dispatch can only use bounds, not actual values.
- This explicit separation of value vs sequence was applied to other wrappers as well for consistency, even if in those cases the type system could be used to infer it.
- **`argument::constant<Value>` and `argument::constant_sequence<Seq>** — compile-time constant. Supports scalars and arrays in the sequence case (e.g., `argument::constant_sequence<std::array{128, 256, 512}>`). Bounds are derived from the value itself.
- **`argument::immediate{value, bounds...}` and `argument::immediate_sequence{seq, bounds...}`** — runtime value with optional static and/or runtime bounds. Stores any type (scalar, span, pointer, etc.), but for multiple values `immediate_sequence` should be used.

### Bounds are separate types created via factory functions

- **`cuda::argument::bounds<1, 8>()`** — static bounds, type deduced from NTTPs (no need to spell the value type)
- **`cuda::argument::bounds(rt_min, rt_max)`** — runtime bounds, type deduced via CTAD
- Both can be provided: `argument::dynamic{k, argument::bounds<1, 256>(), argument::bounds(50, 200)}`
- Static bounds are part of the wrapper's type for compile-time specialization
- Runtime bounds are stored as a member with full-range defaults

### Unified access via `argument::traits` and free functions

- **`argument::traits<T>`** — type-level access: `is_deferred`, `value_type`, `element_type`, `static_min`, `static_max`. Works in `if constexpr` and `static_assert`.
- **`argument::unwrap(arg)`** — extracts the raw value from any wrapper or passes plain values through
- **`argument::static_min(arg)`** / **`argument::static_max(arg)`** — free functions for static bounds
[...]

PR #9074, opened 2026-05-20
[cub] Replace cub parameter framework with cuda::argument

pciolkosz:
This PR replaces most of the functionality in `segmented_params.cuh` with `cuda::argument` wrappers from https://github.com/NVIDIA/cccl/pull/8875. This PR contains the other one, since it's not merged yet.

There are two things that were left from the original implementation, the static dispatch over bounded set of values and `get_param` that either gets item from a sequence at a given index or returns a uniform value depending on the argument. Both of those things were more fitting for a cub-specific functionality, but its not set in stone

### DIFF
--- a/cub/benchmarks/bench/segmented_topk/fixed/keys.cu
+++ b/cub/benchmarks/bench/segmented_topk/fixed/keys.cu
@@ -4,7 +4,7 @@
 #include <cub/detail/choose_offset.cuh>
 #include <cub/device/dispatch/dispatch_batched_topk.cuh>
 
-#include <cuda/argument>
+#include <cuda/__argument_>
 #include <cuda/iterator>
 
 #include <nvbench_helper.cuh>
@@ -51,7 +51,7 @@ void fixed_seg_size_topk_keys(
   const auto selected_elements = static_cast<::cuda::std::ptrdiff_t>(MaxNumSelected);
   const auto num_segments      = ::cuda::std::max<std::size_t>(1, (max_elements / segment_size));
   const auto elements          = num_segments * segment_size;
-  const auto total_num_items   = ::cuda::argument::__immediate{static_cast<::cuda::std::int64_t>(elements)};
+  const auto total_num_items   = ::cuda::__argument::__immediate{static_cast<::cuda::std::int64_t>(elements)};
   const bit_entropy entropy    = str_to_entropy(state.get_string("Entropy"));
 
   // Skip workloads where k exceeds the segment size
@@ -68,8 +68,8 @@ void fixed_seg_size_topk_keys(
   auto d_keys_in      = cuda::make_strided_iterator(cuda::make_counting_iterator(d_keys_in_ptr), segment_size);
   auto d_keys_out     = cuda::make_strided_iterator(cuda::make_counting_iterator(d_keys_out_ptr), selected_elements);
 
-  auto segment_sizes    = ::cuda::argument::__constant<MaxSegmentSize>{};
-  auto k                = ::cuda::argument::__constant<MaxNumSelected>{};
+  auto segment_sizes    = ::cuda::__argument::__constant<MaxSegmentSize>{};
+  auto k                = ::cuda::__argument::__constant<MaxNumSelected>{};
   auto select_direction = cub::detail::topk::select::max;
 
   state.add_element_count(elements, "NumElements");
@@ -99,7 +99,7 @@ void fixed_seg_size_topk_keys(
       segment_sizes,
       k,
       select_direction,
-      ::cuda::argument::__immediate{static_cast<::cuda::std::int64_t>(num_segments)},
+      ::cuda::__argument::__immediate{static_cast<::cuda::std::int64_t>(num_segments)},
       total_num_items,
       env);
   });
--- a/cub/benchmarks/bench/segmented_topk/fixed/keys.cu
+++ b/cub/benchmarks/bench/segmented_topk/fixed/keys.cu
@@ -4,6 +4,7 @@
 #include <cub/detail/choose_offset.cuh>
 #include <cub/device/dispatch/dispatch_batched_topk.cuh>
 
+#include <cuda/argument>
 #include <cuda/iterator>
 
 #include <nvbench_helper.cuh>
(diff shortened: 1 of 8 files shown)


cub/benchmarks/bench/segmented_topk/fixed/keys.cu
```diff
@@ -87,9 +68,9 @@ void fixed_seg_size_topk_keys(
   auto d_keys_in      = cuda::make_strided_iterator(cuda::make_counting_iterator(d_keys_in_ptr), segment_size);
   auto d_keys_out     = cuda::make_strided_iterator(cuda::make_counting_iterator(d_keys_out_ptr), selected_elements);
 
-  auto segment_sizes     = seg_size_t{};
-  auto k                 = k_value_t{};
-  auto select_directions = select_direction_value_t{};
+  auto segment_sizes    = ::cuda::__argument::__constant<MaxSegmentSize>{};
+  auto k                = ::cuda::__argument::__constant<MaxNumSelected>{};
+  auto select_direction = cub::detail::topk::select::max;
```
elstehle: important: I think we want to switch to making this a compile-time option:
```suggestion
  auto select_direction = ::cuda::__argument::__constant<cub::detail::topk::select::max>{};
```

cub/benchmarks/bench/segmented_topk/fixed/keys.cu
```diff
@@ -44,33 +45,13 @@ void fixed_seg_size_topk_keys(
...
-    cub::detail::batched_topk::total_num_items_guarantee<min_num_total_items, max_num_total_items>;
-
   // Retrieve axis parameters
   const auto max_elements      = static_cast<size_t>(state.get_int64("Elements{io}"));
   const auto segment_size      = static_cast<::cuda::std::ptrdiff_t>(MaxSegmentSize);
   const auto selected_elements = static_cast<::cuda::std::ptrdiff_t>(MaxNumSelected);
   const auto num_segments      = ::cuda::std::max<std::size_t>(1, (max_elements / segment_size));
   const auto elements          = num_segments * segment_size;
-  const auto total_num_items   = total_num_items_guarantee_t{static_cast<::cuda::std::int64_t>(elements)};
+  const auto total_num_items   = ::cuda::__argument::__immediate{static_cast<::cuda::std::int64_t>(elements)};
```
elstehle, line 54: note to myself: I think this should become part of the guarantees API, as, in the device interface, there is no concrete argument that could be annotated.

cub/cub/agent/agent_batched_topk.cuh
```diff
@@ -72,8 +73,8 @@ struct agent_batched_topk_worker_per_segment
   using key_t   = it_value_t<key_it_t>;
   using value_t = it_value_t<value_it_t>;
 
-  using segment_size_val_t = typename SegmentSizeParameterT::value_type;
-  using num_segments_val_t = typename NumSegmentsParameterT::value_type;
+  using segment_size_val_t = typename ::cuda::__argument::__traits<SegmentSizeParameterT>::element_type;
+  using num_segments_val_t = typename ::cuda::__argument::__traits<NumSegmentsParameterT>::element_type;
```
elstehle, line 77: Note to myself: I think we want to narrower the index/offset/size types to the more narrow of the two: the static upper bound or the element type.

cub/cub/agent/agent_batched_topk.cuh
```diff
@@ -190,16 +191,16 @@ struct agent_batched_topk_worker_per_segment
...
     // TODO (elstehle): consider skipping boundary check if we can safely assume the right grid dimensions
-    if (segment_id >= num_segments.get_param(0))
+    if (segment_id >= params::get_param(num_segments, 0))
     {
       return;
     }
 
-    constexpr bool is_full_tile = params::has_single_static_value_v<SegmentSizeParameterT>
-                               && params::static_min_value_v<SegmentSizeParameterT> == tile_size;
+    constexpr bool is_full_tile = ::cuda::__argument::__traits<SegmentSizeParameterT>::is_constant
```
elstehle, line 199: question: For something like an `__immediate` where we have sharp bounds, i.e., `lowest=max`, we wouldn't hit the full_tile branch, right? Do you think this is something we should somehow cover or do you think we can expect users to be always be using `__constant`? I guess there could be a scenario where users themselves do have something like bounds template parameters and would not check for narrow bounds before instantiating something with `__immediate`?

cub/cub/device/dispatch/dispatch_batched_topk.cuh
```diff
@@ -44,103 +46,25 @@ CUB_NAMESPACE_BEGIN
...
-  // Create default ctor, 1 param ctor taking min, 2 param ctor taking min/max
-  total_num_items_guarantee() = default;
-
-  _CCCL_HOST_DEVICE total_num_items_guarantee(value_type num_items)
-      : min_num_items(num_items)
-      , max_num_items(num_items)
-  {}
+  return params::uniform_discrete_param<detail::topk::select, detail::topk::select::max, detail::topk::select::min>{
+    dir};
+}
```
elstehle, line 65: important: I think the compile-time-only option (i.e., `__constant`) is probably the most important one.

```suggestion
// Uniform (compile-time): __constant<Dir> → single-option uniform_discrete_param (specialized to one direction)
template <detail::topk::select Dir>
_CCCL_HOST_DEVICE auto wrap_select_direction(::cuda::__argument::__constant<Dir>)
{
  return params::uniform_discrete_param<detail::topk::select, Dir>{Dir};
}

// Uniform: single enum value → uniform_discrete_param
_CCCL_HOST_DEVICE inline auto wrap_select_direction(detail::topk::select dir)
{
  return params::uniform_discrete_param<detail::topk::select, detail::topk::select::max, detail::topk::select::min>{
    dir};
}

```

cub/benchmarks/bench/segmented_topk/variable/keys.cu
```diff
@@ -171,20 +172,17 @@ void variable_seg_size_topk_keys(nvbench::state& state,
...
-  cub::detail::batched_topk::segment_size_per_segment<const cuda::std::int64_t*, 1, MaxSegmentSize> segment_sizes_param{
-    thrust::raw_pointer_cast(d_segment_sizes.data())};
-  cub::detail::batched_topk::k_static<K> k_param{};
-  cub::detail::batched_topk::select_direction_static<cub::detail::topk::select::max> select_directions{};
-  cub::detail::batched_topk::num_segments_uniform<> num_segments_uniform_param{
-    static_cast<cuda::std::int64_t>(num_segments)};
+  auto segment_sizes_param = ::cuda::__argument::__immediate_sequence{
+    thrust::raw_pointer_cast(d_segment_sizes.data()), ::cuda::__argument::__bounds<1, MaxSegmentSize>()};
+  auto k_param            = ::cuda::__argument::__constant<K>{};
+  auto select_direction   = cub::detail::topk::select::max;
```
elstehle:
```suggestion
  auto select_direction   = ::cuda::__argument::__constant<cub::detail::topk::select::max>{};
```

cub/cub/detail/segmented_params.cuh
```diff
@@ -13,127 +13,77 @@
...
+// =====================================================================
+// get_param — unified segment parameter access
+// =====================================================================
+
+//! @brief Returns the value of an argument for a given segment index.
+//! For single-value arguments, the index is ignored and the value is returned directly.
+//! For per-segment arguments, returns the element at the given index.
+_CCCL_TEMPLATE(class _Tp)
+_CCCL_REQUIRES((!::cuda::__argument::__is_wrapper_v<::cuda::std::remove_cvref_t<_Tp>>) )
+_CCCL_HOST_DEVICE constexpr auto get_param(_Tp&& __arg, [[maybe_unused]] size_t __index) noexcept
```
elstehle: I think we may want to support narrower segment index types, like 32-bit types, for potentially better code gen, when actually indexing into a `_sequence` parameter? (Same would apply throughout `get_param` helpers).
```suggestion
_CCCL_HOST_DEVICE constexpr auto get_param(_Tp&& __arg, [[maybe_unused]] SegmentIndexT __index) noexcept
```

elstehle (approved):
I think this looks much cleaner with the argument annotation framework replacements. Thanks for working on this!

cub/cub/agent/agent_batched_topk.cuh
pciolkosz: I think we could make `__immediate` smarter and report `is_constant` when it has static bounds where `lower == higher`. I will take a look into it in a follow-up PR.

cub/benchmarks/bench/segmented_topk/fixed/keys.cu
elstehle: PR: https://github.com/NVIDIA/cccl/pull/9278

PR #9076, opened 2026-05-20
[cudax] Disable cuco hashers `__int128` test for nvcc 12.0 + gcc combination

davebayer:


cudax/test/cuco/utility/test_hashers.cu
```diff
@@ -224,20 +224,24 @@ TEST_CASE("Test Hasher's on host and device", "")
     test_xxhash32{}();
     test_xxhash64{}();
     test_murmurhash3_32{}();
-#if _CCCL_HAS_INT128()
+
+    // nvcc 12.0 crashes for int128 tests when using gcc as the host compiler
+#if _CCCL_HAS_INT128() && !(_CCCL_CUDA_COMPILER(NVCC, ==, 12, 0) && _CCCL_COMPILER(GCC))
```
bernhardmgruber, line 229: Q: I thought we shouldn't compare a compiler version with `==` since it will fail unless the compiler has a matching patch version.

Should this be?

```suggestion
#if _CCCL_HAS_INT128() && !(_CCCL_CUDA_COMPILER(NVCC, <, 12, 1) && _CCCL_COMPILER(GCC))
```

davebayer: No, the compiler macros ignore build versions, we just check the major + minor version

bernhardmgruber (approved):

PR #9088, opened 2026-05-21
Refactor DeviceReduce dispatch logic

bernhardmgruber:
- [x] @NaderAlAwar: Critical: this breaks `Sum(uint8_t*, uint8_t*, ...)`. so that it doesn't compile anymore. Need to add a few more tests and fix some dispatching logic.

### DIFF
--- a/cub/cub/device/device_reduce.cuh
+++ b/cub/cub/device/device_reduce.cuh
@@ -111,57 +111,54 @@ private:
     using accum_t  = ::cuda::std::
       __accumulator_t<ReductionOpT, ::cuda::std::invoke_result_t<TransformOpT, detail::it_value_t<InputIteratorT>>, T>;
 
-    return detail::dispatch_with_env(env, [&]([[maybe_unused]] auto tuning, void* storage, size_t& bytes, auto stream) {
-      using tuning_env_t = decltype(tuning);
-
-      if constexpr (Determinism == ::cuda::execution::determinism::__determinism_t::__gpu_to_gpu)
-      {
-        // Only instantiated with `plus<float|double>`; RFA hardcodes `deterministic_sum_t<accum_t>`.
-        (void) reduction_op;
-        using default_policy_selector = detail::rfa::policy_selector_from_types<accum_t>;
-        using policy_selector =
-          ::cuda::std::execution::__query_result_or_t<tuning_env_t, detail::rfa::rfa_policy, default_policy_selector>;
-        return detail::rfa::dispatch<InputIteratorT, OutputIteratorT, offset_t, T, TransformOpT, accum_t>(
-          storage, bytes, d_in, d_out, static_cast<offset_t>(num_items), init, stream, transform_op, policy_selector{});
-      }
-      else if constexpr (Determinism == ::cuda::execution::determinism::__determinism_t::__not_guaranteed)
-      {
-        using default_policy_selector =
-          detail::reduce_nondeterministic::policy_selector_from_types<accum_t, offset_t, ReductionOpT>;
-        using policy_selector =
-          ::cuda::std::execution::__query_result_or_t<tuning_env_t,
-                                                      detail::reduce_nondeterministic::reduce_nondeterministic_policy,
-                                                      default_policy_selector>;
-        return detail::reduce_nondeterministic::dispatch<accum_t>(
-          storage,
-          bytes,
-          d_in,
-          THRUST_NS_QUALIFIER::unwrap_contiguous_iterator(d_out),
-          static_cast<offset_t>(num_items),
-          reduction_op,
-          init,
-          stream,
-          transform_op,
-          policy_selector{});
-      }
-      else
-      {
-        using default_policy_selector = detail::reduce::policy_selector_from_types<accum_t, offset_t, ReductionOpT>;
-        using policy_selector         = ::cuda::std::execution::
-          __query_result_or_t<tuning_env_t, detail::reduce::reduce_policy, default_policy_selector>;
-        return detail::reduce::dispatch<accum_t>(
-          storage,
-          bytes,
-          d_in,
-          d_out,
-          static_cast<offset_t>(num_items),
-          reduction_op,
-          init,
-          stream,
-          transform_op,
-          policy_selector{});
-      }
-    });
+    if constexpr (Determinism == ::cuda::execution::determinism::__determinism_t::__gpu_to_gpu)
+    {
+      // Only instantiated with `plus<float|double>`; RFA hardcodes `deterministic_sum_t<accum_t>`.
+      (void) reduction_op;
+      using default_policy_selector = detail::rfa::policy_selector_from_types<accum_t>;
+      return detail::dispatch_with_env_and_tuning<default_policy_selector>(
+        env, [&](auto policy_selector, void* storage, size_t& bytes, cudaStream_t stream) {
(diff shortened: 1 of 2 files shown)


tpn (approved):
Reviewed the DeviceReduce dispatch refactor. No blocking issues from me; the CI summary is green.

cub/cub/device/device_reduce.cuh
```diff
@@ -522,37 +545,10 @@ public:
...
-      ::cuda::std::conditional_t<no_determinism && !(is_contiguous_fallback && is_4b_or_greater),
-                                 ::cuda::execution::determinism::run_to_run_t,
-                                 default_determinism_t>;
-
-    using InitT = OutputT;
-
-    return reduce_impl(
-      d_in, d_out, num_items, ::cuda::std::plus<>{}, ::cuda::std::identity{}, InitT{}, determinism_t{}, env);
+    using accum_t = ::cuda::std::__accumulator_t<::cuda::std::plus<>, cub::detail::it_value_t<InputIteratorT>, OutputT>;
+    return __transform_reduce<accum_t>(
```
NaderAlAwar, line 566: Critical: this breaks `Sum(uint8_t*, uint8_t*, ...)` so that it doesn't compile anymore. Prior to this change, we fell back to run to run determinism for output types < 4 bytes. But `__transform_reduce` checks `AccumT` instead of `OutputT`. For `uint8_t`, `AccumT` is `int` due to integer promotion, so the kernel will attempt to do `atomicAdd(unsigned char*, int)` which doesn't compile

NaderAlAwar: This issue already existed for Reduce() before this PR, but sum had its own detection logic

Thinking about this some more, it is probably okay to merge this PR and I'll create a fix for `__transform_reduce` to check OutputT instead of AccumT, or you can driveby fix it if you want, but the fix is slightly involved and I would like to add tests to verify the behavior. We basically need to check two things:

1. `OutputT == AccumT`
2. `atomicAdd()` is supported for that type. This is mostly handled by `is_4b_or_greater` but it occurs to me now that there is no overload for `long long`.

I'm going to approve, let me know if you prefer fixing it here or I can fix it after this gets merged

NaderAlAwar (approved):

cub/cub/device/device_reduce.cuh
bernhardmgruber: > Critical: this breaks Sum(uint8_t*, uint8_t*, ...) so that it doesn't compile anymore. 

Why do the unit tests pass then? We cover `uint8` in the tests  in `catch2_test_device_reduce.cu`. Is this about the deterministic version? I only see floating point types in `catch2_test_device_reduce_deterministic.cu`. Seems we are missing tests here then.

> the fix is slightly involved and I would like to add tests to verify the behavior

Can you please do this for me then? I think you will get this done faster then me ;)

You can just push to this PR!

cub/cub/device/device_reduce.cuh
```diff
@@ -187,27 +184,25 @@ private:
...
-    using accum_t = AccumT;
-
     constexpr auto gpu_gpu_determinism =
       ::cuda::std::is_same_v<default_determinism_t, ::cuda::execution::determinism::gpu_to_gpu_t>;
 
     // integral types are always gpu-to-gpu deterministic if reduction operator is a simple cuda binary
     // operator, so fallback to run-to-run determinism
     constexpr auto integral_fallback =
-      gpu_gpu_determinism && ::cuda::std::is_integral_v<accum_t> && (detail::is_cuda_binary_operator<ReductionOpT>);
+      gpu_gpu_determinism && ::cuda::std::is_integral_v<AccumT> && (detail::is_cuda_binary_operator<ReductionOpT>);
```
miscco, line 197: This looks wrong given that we calculate `accum_t` above. Why is this not using the calculated one and do we need that actually

cub/cub/device/device_reduce.cuh
```diff
@@ -522,37 +545,10 @@ public:
...
-    using determinism_t =
-      ::cuda::std::conditional_t<no_determinism && !(is_contiguous_fallback && is_4b_or_greater),
-                                 ::cuda::execution::determinism::run_to_run_t,
-                                 default_determinism_t>;
-
-    using InitT = OutputT;
-
-    return reduce_impl(
-      d_in, d_out, num_items, ::cuda::std::plus<>{}, ::cuda::std::identity{}, InitT{}, determinism_t{}, env);
+    using accum_t = ::cuda::std::__accumulator_t<::cuda::std::plus<>, cub::detail::it_value_t<InputIteratorT>, OutputT>;
```
bernhardmgruber, line 565: I am highly confused by this actually. But I see why `claude` generated this. We used to have `using InitT = OutputT;` before this change, which makes absolutely no sense. The initial value (and it's type) is either provided by the user or we determine it from the accumulator type. We may also use the input type for it. But definitely not the output type.

cub/cub/device/device_reduce.cuh
bernhardmgruber: Within this function, the change is fine, since we had above: `using accum_t = AccumT;`.

It does indeed change something as @NaderAlAwar pointed out.

The backstory is: I saw some duplication and asked `claude` to refactor it. Apparently that changes behavior now between the `Sum` and the other APIs of `DeviceReduce`. But that shouldn't be the case IMO. Calling `Sum` should be the same like calling `Reduce(..., std::plus{}, ...)`, otherwise we have a logic error somewhere. I am counting on @NaderAlAwar to figure out the correct functionality here.

cub/cub/device/device_reduce.cuh
NaderAlAwar: > Why do the unit tests pass then? We cover uint8 in the tests in catch2_test_device_reduce.cu. Is this about the deterministic version? I only see floating point types in catch2_test_device_reduce_deterministic.cu. Seems we are missing tests here then.

This is specifically about the nondeterministic implementation. The existing `uint8_t` tests pass because they test the older/default Sum API, not the env overload with not_guaranteed.

In the env path, `Sum(uint8_t*, uint8_t*, ...)` deduces `AccumT == int` due to integer promotion, while `OutputT == uint8_t`. The nondeterministic implementation atomically accumulates directly into `d_out`, so it cannot be used when `OutputT != AccumT`.

The change I just pushed adds a check whether `AccumT == OutputT`. If not, fall back to run to run determinism. Before this change, the helper only checked the accumulator size, so `uint8_t` output with `AccumT == int` incorrectly selected the atomic path and tried to instantiate an atomic add into the `uint8_t` output location using an int accumulator. That fails to compile because the atomic path treats `d_out[0]` as an `AccumT` atomic object, but in this case `d_out[0]` is actually a `uint8_t`.

I added a fix so that this case falls back to run-to-run determinism, and added env tests covering both Reduce and Sum with uint8_t output.

cub/cub/device/device_reduce.cuh
NaderAlAwar: Fixed, see https://github.com/NVIDIA/cccl/pull/9088#discussion_r3305884183

cub/cub/device/device_reduce.cuh
NaderAlAwar: `DeviceReduce::Sum` does not have an initial value for the reduction unlike `Reduce`, so we always set that as 0 with a type we select. It might make more sense to have it be `InputT`, but this could be a breaking change

NaderAlAwar:
pre-commit ci autofix

NaderAlAwar:
pre-commit.ci autofix

PR #9094, opened 2026-05-21
Codestyle: Exclude unit test files from full qualification

bernhardmgruber:


fbusato:
my plan is to create another SKILL for tests. Mixing everything together is a bit confusing. The styles between library and tests are very difference

bernhardmgruber:
> my plan is to create another SKILL for tests. Mixing everything together is a bit confusing. The styles between library and tests are very difference

That's a very good idea! Can you do it soon? I get Coderabbit complain a lot on my tests :D

fbusato:
sure, I can open a PR by today

fbusato:
created https://github.com/NVIDIA/cccl/pull/9100.

bernhardmgruber:
> created #9100.

Closing in favor of the new issue

PR #9095, opened 2026-05-21
Use the new tuning API internally for `detail::batched_topk::dispatch`

### LINKED ISSUE #8482: Use the new tuning API internally for `detail::batched_topk::dispatch_batched_topk`

bernhardmgruber:
- [x] No SASS changes for `cub.bench.segmented_topk.fixed.base` on SM120
- [x] No SASS changes for `cub.bench.segmented_topk.variable.base` on SM120

Fixes: #8482

### DIFF
--- a/cub/benchmarks/bench/segmented_topk/fixed/keys.cu
+++ b/cub/benchmarks/bench/segmented_topk/fixed/keys.cu
@@ -107,6 +107,7 @@ void fixed_seg_size_topk_keys(
       cuda::execution::tune(tuned_policy_selector{})
 #endif // !TUNE_BASE
     );
+    // TODO(bgruber): call the public API once available
     _CCCL_TRY_CUDA_API(
       cub::detail::batched_topk::dispatch_with_env,
       "batched topk failed",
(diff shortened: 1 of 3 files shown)


bernhardmgruber:
@elstehle Does the changeset proposed by this PR even make sense at this point? It seems there is no public API for batched topk yet, but I am going through benchmarks and move them from the two step allocation to the environment overload that does the allocation automatically.

Should I instead just leave a note in the benchmark to update them later?

elstehle:
> @elstehle Does the changeset proposed by this PR even make sense at this point? It seems there is no public API for batched topk yet, but I am going through benchmarks and move them from the two step allocation to the environment overload that does the allocation automatically.
> 
> Should I instead just leave a note in the benchmark to update them later?

If the question is whether it's worth introducing the `dispatch_with_env` wrapper for the time being to simplify the benchmarks, then I'm good moving ahead. I just want to be thoughtful of exposing anything publicly yet or committing to the exact set of tuning parameters, as these things are still in flux.

elstehle (approved):

bernhardmgruber:
> If the question is whether it's worth introducing the `dispatch_with_env` wrapper for the time being to simplify the benchmarks, then I'm good moving ahead.

Yes, that is the question.

>  I just want to be thoughtful of exposing anything publicly yet or committing to the exact set of tuning parameters, as these things are still in flux.

That's fine, nothing public yet!

PR #9099, opened 2026-05-21
[codex] libcudacxx: relax bit_cast fallback construction check

tpn:
## Summary
- replace the `cuda::std::bit_cast` memcpy fallback assertion from `default_initializable<_To>` to `is_default_constructible_v<_To>`
- add `__float128` / `__uint128_t` coverage to the existing `bit_cast` roundtrip test

## Root Cause
The memcpy fallback only needs to default-construct a temporary destination object before copying bytes into it. `default_initializable` is stricter than that and can reject compiler extended floating-point types such as `__float128`, causing the fallback path to fail even though the destination type is default constructible.

## Validation
- `g++ -std=gnu++20 -Ilibcudacxx/include -Ilibcudacxx/test/support libcudacxx/test/libcudacxx/std/numerics/bit/bit.cast/bit_cast.pass.cpp -o /tmp/cccl-bit-cast-pass-gcc && /tmp/cccl-bit-cast-pass-gcc`
- `clang++ -std=gnu++20 -Ilibcudacxx/include -Ilibcudacxx/test/support libcudacxx/test/libcudacxx/std/numerics/bit/bit.cast/bit_cast.pass.cpp -o /tmp/cccl-bit-cast-pass-clang && /tmp/cccl-bit-cast-pass-clang`
- `nvcc -std=c++20 -Ilibcudacxx/include -Ilibcudacxx/test/support -x cu libcudacxx/test/libcudacxx/std/numerics/bit/bit.cast/bit_cast.pass.cpp -o /tmp/cccl-bit-cast-pass-nvcc && /tmp/cccl-bit-cast-pass-nvcc`
- commit hook suite, including `clang-format` and `codespell`

### DIFF
--- a/libcudacxx/include/cuda/std/__bit/bit_cast.h
+++ b/libcudacxx/include/cuda/std/__bit/bit_cast.h
@@ -23,7 +23,7 @@
 
 #include <cuda/__type_traits/is_trivially_copyable.h>
 #include <cuda/std/__concepts/concept_macros.h>
-#include <cuda/std/__concepts/constructible.h>
+#include <cuda/std/__type_traits/is_default_constructible.h>
 #include <cuda/std/cstring>
 
 #include <cuda/std/__cccl/prologue.h>
@@ -52,7 +52,7 @@ template <class _To, class _From>
 [[nodiscard]] _CCCL_API inline _To __bit_cast_memcpy(const _From& __from) noexcept
 {
 #if !_CCCL_COMPILER(GCC, <=, 8)
-  static_assert(::cuda::std::default_initializable<_To>,
+  static_assert(::cuda::std::is_default_constructible_v<_To>,
                 "bit_cast memcpy fallback requires the destination type to be default constructible");
 #endif // !_CCCL_COMPILER(GCC, <=, 8)
   _To __temp;
--- a/libcudacxx/test/libcudacxx/std/numerics/bit/bit.cast/bit_cast.pass.cpp
+++ b/libcudacxx/test/libcudacxx/std/numerics/bit/bit.cast/bit_cast.pass.cpp
@@ -285,6 +285,18 @@ TEST_FUNC bool tests()
     test_roundtrip_through<long long>(i);
   }
 
+#if _CCCL_HAS_FLOAT128()
+  {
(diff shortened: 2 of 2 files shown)


fbusato:
It looks like a hallucination.

Tested locally with `__float128` / `__uint128_t` on several configurations across clang, gcc, cuda-12, nvrtc, nvc++, etc. Everything passes. Also, `default_initializable<_To>` is the right trait, and stricter than `is_default_constructible_v` on purpose. 
Where are you finding the failure?

tpn:
Thanks for checking. The original signal came from an internal NVIDIA DVS/NVBugs report, so I do not want to paste the internal links/log URLs into a public PR.

I did reproduce the relevant condition locally on the public checkout:

- With `g++ -std=gnu++20`, CCCL reports `cuda::std::default_initializable<__float128> == false` while `cuda::std::is_default_constructible_v<__float128> == true`.
- Directly instantiating the fallback path fails on `origin/main`:
  `cuda::std::__bit_cast_memcpy<__float128>(__uint128_t{})` hits the `default_initializable` static assertion in `bit_cast.h`.
- The same direct fallback instantiation passes on this PR.

You are also right that ordinary `bit_cast<__float128>` tests can pass on compilers that use `__builtin_bit_cast`, because they bypass the memcpy fallback. I pushed `1eb080f4b0` to tighten the regression test so it explicitly instantiates `__bit_cast_memcpy<__float128>` under the existing `_CCCL_HAS_FLOAT128()` guard.

libcudacxx/include/cuda/std/__bit/bit_cast.h
```diff
@@ -52,7 +52,7 @@ template <class _To, class _From>
 [[nodiscard]] _CCCL_API inline _To __bit_cast_memcpy(const _From& __from) noexcept
 {
 #if !_CCCL_COMPILER(GCC, <=, 8)
-  static_assert(::cuda::std::default_initializable<_To>,
+  static_assert(::cuda::std::is_default_constructible_v<_To>,
```
bernhardmgruber, line 55: I have seen this being fixed elsewhere. Is the PR up to date?

tpn:
Closed; obsolete.

PR #9105, opened 2026-05-22
[libcu++] Do not require `default_initializable` for bit_cast

miscco:
The additional constraints in `default_initializable` namely that `new _Tp` is well formed are not necessary here. We only need it because we default construct a `T` via `T _temp;`

Fixes https://nvbugspro.nvidia.com/bug/6194162


bernhardmgruber (approved):

davebayer (approved):

fbusato:
still disagree (fine for a quick fix of that anyway).

nvbug 6194162 reports the problem for GCC 8.5 which was already addressed in https://github.com/NVIDIA/cccl/pull/8924. It could be just the condition with the macro `#if !_CCCL_COMPILER(GCC, <=, 8)` -> `#if !_CCCL_COMPILER(GCC, <, 9)`

PR #9112, opened 2026-05-22
[CUB] Fix invalid condition to exclude 128 bit integers

miscco:
Looks like a simple typo, the condition should be or and not and, because that is not possible


davebayer (approved):

davebayer:
pre-commit.ci autofix

PR #9119, opened 2026-05-23
Asserting current device and CUB stream matches

### LINKED ISSUE #7782: Validate current device and CUB stream matches
All CUB algorithms accept a user provided stream argument. We assume in many places that the current CUDA device matches to where the stream is going. For example, when we ask for the PTX architecture or the available SMEM on the current device. If the current device is not the one where kernels will be launched via the user-provided stream, errors may happen, like launching a kernel with more dynamic SMEM than is available on the device to where the stream goes to.

We should add a check, at least an assertion, to all CUB dispatching codes that the stream launches kernels to the same device as the current CUDA device.

thom-gg:
## Description

closes #7782 

Adding an assertion to all the dispatching codes to ensure current device and CUB stream matches. Calling it at the very beginning of each of the dispatch functions, hence the number of modified files

The assertion itself uses `cudaStreamGetDevice` which was introduced in CTK 12.8 so it's guarded by the macro _CCCL_CTK_AT_LEAST(12,8). 

I'm new to the project so unsure if there is a better place to call the assertion rather than doing it in every dispatch file, also unsure if the assertion should be put in the  `cub/cub/util_device.cuh` file like i did or elsewhere, please tell me if this issue should be addressed differently and i'll try to do it !

## Checklist


cub/cub/util_device.cuh
```diff
@@ -456,6 +456,35 @@ CUB_RUNTIME_FUNCTION inline cudaError_t SyncStream([[maybe_unused]] cudaStream_t
...
+CUB_RUNTIME_FUNCTION _CCCL_FORCEINLINE cudaError_t validate_stream_device(cudaStream_t stream)
+{
+  cudaError_t error = cudaSuccess;
+#  if _CCCL_CTK_AT_LEAST(12, 8)
+  int streamDevice;
+  error = cudaStreamGetDevice(stream, &streamDevice);
+  if (error != cudaSuccess)
+  {
+    return error;
+  }
```
bernhardmgruber: Suggestion: Let's not reuse the `error` variable:

```suggestion
  if (const auto error = cudaStreamGetDevice(stream, &streamDevice);)
  {
    return error;
  }
```

cub/cub/util_device.cuh
```diff
@@ -456,6 +456,35 @@ CUB_RUNTIME_FUNCTION inline cudaError_t SyncStream([[maybe_unused]] cudaStream_t
...
+  int currentDevice;
+  error = cudaGetDevice(&currentDevice);
+  if (error != cudaSuccess)
+  {
+    return error;
+  }
```
bernhardmgruber:
```suggestion
  if (const auto error = cudaGetDevice(&currentDevice);)
  {
    return error;
  }
```

cub/cub/util_device.cuh
```diff
@@ -456,6 +456,35 @@ CUB_RUNTIME_FUNCTION inline cudaError_t SyncStream([[maybe_unused]] cudaStream_t
...
+  _CCCL_ASSERT(currentDevice == streamDevice, "current device must match CUB stream device");
+  if (currentDevice != streamDevice)
+  {
+    return cudaErrorInvalidDevice;
+  }
+#  endif // _CCCL_CTK_AT_LEAST(12,8)
+  return error;
```
bernhardmgruber:
```suggestion
  return cudaSuccess;
```

cub/cub/util_device.cuh
```diff
@@ -456,6 +456,35 @@ CUB_RUNTIME_FUNCTION inline cudaError_t SyncStream([[maybe_unused]] cudaStream_t
   NV_IF_ELSE_TARGET(NV_IS_HOST, (return CubDebug(cudaStreamSynchronize(stream));), (return cudaErrorNotSupported;))
 }
 
+namespace detail
+{
+// Validates stream's device is current device, when CTK >= 12.8, otherwise does nothing.
+CUB_RUNTIME_FUNCTION _CCCL_FORCEINLINE cudaError_t validate_stream_device(cudaStream_t stream)
+{
+  cudaError_t error = cudaSuccess;
+#  if _CCCL_CTK_AT_LEAST(12, 8)
```
bernhardmgruber: Important: sometimes users violate our API requirements, but their software ran fine for a long time. They would be upset if we suddenly enforce requirements, causing their software to break. Let's add a macro to disable this new feature:
```suggestion
#  if _CCCL_CTK_AT_LEAST(12, 8) && !defined(CCCL_DISABLE_STREAM_DEVICE_CHECK)
```
If possible, add a unit test that calls a simple algorithm like `DeviceFor` with a stream and a different current device and define the `CCCL_DISABLE_STREAM_DEVICE_CHECK` macro, to see whether the escape hatch works.

bernhardmgruber (commented):
Thanks a lot for this contribution! Please add a unit test to at least one algorithm calling it with a stream that does not match the current device. This test must be written in a way that it also works if there is only one GPU/device in the system (just succeeding is fine I think). I can try it briefly on my machine where I have two GPUs.

cub/cub/util_device.cuh
```diff
@@ -456,6 +456,35 @@ CUB_RUNTIME_FUNCTION inline cudaError_t SyncStream([[maybe_unused]] cudaStream_t
...
+  int streamDevice;
+  error = cudaStreamGetDevice(stream, &streamDevice);
```
davebayer: We can make this function work even before CTK 12.8 using CUDA Driver API. We already have this implemented for `cuda::stream_ref`. It should look as:
```cpp
{
  ::CUdevice current_device;
  if (const auto error = ::cuda::__driver::__ctxGetDeviceNoThrow(current_device); error != cudaSuccess)
  {
    return error;
  }

  ::CUcontext stream_ctx;
  if (const auto error = ::cuda::__driver::__streamGetCtxNoThrow(stream_ctx, stream); error != cudaSuccess)
  {
    return error;
  }

  if (const auto error = ::cuda::__driver::__ctxPushNoThrow(stream_ctx); error != cudaSuccess)
  {
    return error;
  }

  ::CUdevice stream_device;
  if (const auto error = ::cuda::__driver::__ctxGetDeviceNoThrow(stream_device); error != cudaSuccess)
  {
    return error;
  }

  if (const auto error = ::cuda::__driver::__ctxPopNoThrow(); error != cudaSuccess)
  {
    return error;
  }

  _CCCL_ASSERT(current_device == stream_device, "current device must match CUB stream device");
}
```
The only problem is that we need to add `__meowNoThrow` variants of all context-related driver APIs to `<cuda/__driver/driver_api.h>`.

If you don't feel comfortable doing this, I will make a follow up PR after this one is merged :)

cub/cub/util_device.cuh
davebayer: Critical: Since this is an assertion, we need to make sure all of the CUDA Runtime/Driver calls are done only when assertions are enabled, because they won't get optimized out and can introduce some unwanted overhead.

davebayer (changes requested):

thom-gg:
Hi, thanks to you both for the feedbacks:

- i added the noThrow variants for the driver api methods
- used them in validate_stream_device to support all CTK version and not only >=12.8
- stopped re-using the error variables
- added a macro CCCL_DISABLE_STREAM_DEVICE_CHECK to disable the check
- guarded the check in behind a CCCL_ENABLE_ASSERTIONS macro
- wrote 2 unit tests launching DeviceFor on a stream from another device than the current one, one test that should fail and one that defines CCCL_DISABLE_STREAM_DEVICE_CHECK and should therefore skip the check and succeed. I wasn't able to run the tests since i only have one gpu, but i compiled them though. if there is less than 2 gpus the tests are skipped.

happy to keep modifying this if needed :)

cub/test/catch2_test_device_for_api.cu
```diff
@@ -421,4 +421,31 @@ C2H_TEST("DeviceFor::ForEachCopy legacy size-query is unambiguous", "[for][devic
...
+
+  if (num_devices < 2)
+  {
+    SKIP("Test requires at least 2 CUDA devices");
+  }
+
+  REQUIRE(cudaSetDevice(1) == cudaSuccess);
+  cudaStream_t stream_on_device_1;
+  REQUIRE(cudaStreamCreate(&stream_on_device_1) == cudaSuccess);
+  REQUIRE(cudaSetDevice(0) == cudaSuccess);
```
bernhardmgruber: Suggestion: We have a modern CUDA runtime API now. From the top of my head:
```suggestion
  if (cuda::devices.size() < 2)
  {
    SKIP("Test requires at least 2 CUDA devices");
  }

  cuda::stream stream_on_device_1(cuda::devices[1]);
  REQUIRE(cudaSetDevice(0) == cudaSuccess); // not sure if this is still required, please check
```

cub/test/catch2_test_device_for.cu
```diff
@@ -15,6 +15,8 @@
 #include "catch2_test_launch_helper.h"
 #include <c2h/catch2_test_helper.h>
 
+#define CCCL_DISABLE_STREAM_DEVICE_CHECK
+
```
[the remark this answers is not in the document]
bernhardmgruber: I agree. Please just create a new file, like `catch2_test_device_for_stream.cu` or so, and move the new test there. Thx!

thom-gg:
Hi, implemented your suggestions @bernhardmgruber :

- tests use the modern cuda runtime api cuda::devices, cuda::stream etc, and as you mentioned it's not necessary to set back to device 0 anymore since the stream constructor uses `__ensure_current_context` which sets state back on destruction

- i moved the test containing `#define CCCL_DISABLE_STREAM_DEVICE_CHECK` to another test file to avoid messing with the other tests

- the CI highlighted that dispatch functions are `CUB_RUNTIME_FUNCTION` which can be host and device functions, while the driver functions are host only, so i added a check using NV_IF_TARGET(NV_IS_HOST in the validate_stream_device function. this seems to be why some builds of CUB and Thrust failed

- i'm quite confused why builds of libcudaxx were failing tho, the errors mention issues in `libcudacxx/test/libcudacxx/std/numerics/simd/simd.complex/simd_test_utils.h` which as far as i see are not using the driver api (which is the only part i touched in libcudaxx, and even there i only added new functions so it shouldnt affect existing tests)

tests are taking forever to build locally on my laptop (like +2hrs per target) so i only tried on one target (cuda 12.0 gcc 12). if you have any ideas about the libcudaxx fail i'll take it otherwise i'll try to replicate tomorrow

cub/cub/util_device.cuh
```diff
@@ -456,6 +457,49 @@ CUB_RUNTIME_FUNCTION inline cudaError_t SyncStream([[maybe_unused]] cudaStream_t
   NV_IF_ELSE_TARGET(NV_IS_HOST, (return CubDebug(cudaStreamSynchronize(stream));), (return cudaErrorNotSupported;))
 }
 
+namespace detail
+{
+// Validates stream's device is current device. Does nothing if CCCL_DISABLE_STREAM_DEVICE_CHECK or
+// !CCCL_DISABLE_STREAM_DEVICE_CHECK or is being called from device
```
bernhardmgruber:
```suggestion
// Validates stream's device is current device. Does nothing if CCCL_DISABLE_STREAM_DEVICE_CHECK is
// defined or when being called from device code.
```

cub/cub/util_device.cuh
```diff
@@ -456,6 +457,49 @@ CUB_RUNTIME_FUNCTION inline cudaError_t SyncStream([[maybe_unused]] cudaStream_t
...
+[[nodiscard]] CUB_RUNTIME_FUNCTION _CCCL_FORCEINLINE cudaError_t validate_stream_device(cudaStream_t stream) noexcept
+{
+  NV_IF_TARGET(
+    NV_IS_HOST,
+    (
+
+#  if CCCL_ENABLE_ASSERTIONS && !defined(CCCL_DISABLE_STREAM_DEVICE_CHECK)
```
bernhardmgruber: Important: We cannot put preprocessor directives inside macro argument lists. Please move then outside `NV_IF_TARGET(...)`

cub/cub/util_device.cuh
```diff
@@ -456,6 +457,49 @@ CUB_RUNTIME_FUNCTION inline cudaError_t SyncStream([[maybe_unused]] cudaStream_t
...
 
+namespace detail
+{
+// Validates stream's device is current device. Does nothing if CCCL_DISABLE_STREAM_DEVICE_CHECK or
+// !CCCL_DISABLE_STREAM_DEVICE_CHECK or is being called from device
+[[nodiscard]] CUB_RUNTIME_FUNCTION _CCCL_FORCEINLINE cudaError_t validate_stream_device(cudaStream_t stream) noexcept
+{
+  NV_IF_TARGET(
+    NV_IS_HOST,
+    (
```
bernhardmgruber: Suggestion: If you wrap the code passed into a block statement, you get a lot better formatting:

```suggestion
  NV_IF_TARGET(
    NV_IS_HOST,
    ({
```
and the closing `}` later on.

thom-gg:
Hi, fixed what you commented about NV_IF_TARGET, and tests that failed in the previous CI run now run correctly locally, both for CUB and libcudacxx, everything should be good :)

cub/test/catch2_test_device_for_api.cu
```diff
@@ -421,4 +423,25 @@ C2H_TEST("DeviceFor::ForEachCopy legacy size-query is unambiguous", "[for][devic
...
+
+  // example-begin for-each-n-wo-temp-storage
+  c2h::device_vector<int> vec = {1, 2, 3, 4};
+  square_ref_t op{};
+
+  auto result = cub::DeviceFor::ForEachN(vec.begin(), vec.size(), op, stream_on_device_1);
+  REQUIRE(result == cudaErrorInvalidDevice);
+  // example-end for-each-n-wo-temp-storage
+
+  REQUIRE(cudaStreamDestroy(stream_on_device_1) == cudaSuccess);
```
bernhardmgruber: Important: just pass the a `stream_ref`:
```suggestion
  cuda::stream stream_on_device_1(cuda::devices[1]);

  // example-begin for-each-n-wo-temp-storage
  c2h::device_vector<int> vec = {1, 2, 3, 4};
  square_ref_t op{};

  auto result = cub::DeviceFor::ForEachN(vec.begin(), vec.size(), op, cuda::stream_ref{stream_on_device_1});
  REQUIRE(result == cudaErrorInvalidDevice);
  // example-end for-each-n-wo-temp-storage
```

cub/test/catch2_test_device_for_stream.cu
```diff
@@ -0,0 +1,62 @@
...
+C2H_TEST("Device for each n with stream from another device works when using CCCL_DISABLE_STREAM_DEVICE_CHECK",
+         "[for][device]")
+{
+  if (cuda::devices.size() < 2)
+  {
+    SKIP("Test requires at least 2 CUDA devices");
+  }
+
+  cuda::stream stream_on_device_1_wrapper(cuda::devices[1]);
+  cudaStream_t stream_on_device_1 = stream_on_device_1_wrapper.release();
```
bernhardmgruber: Same here

bernhardmgruber (approved):

thom-gg:
- 6 out of the 8 checks that failed were because the tests using SKIP when there was only one GPU returned exit code 4, and the CI consider everything other than exit code 0 a failure, so i replaced it with SUCCEED, it now returns code 0

- the two other checks have tests (`libcudacxx.test.cuda.containers.buffer.transform` , `cudax.example.async_buffer_add` and `cudax.example.cub_reduce` ) that are failing because the device context is invalid at the beginning of the test. It's a non-deterministic behavior, that seems to depend on the order of execution of tests inside a same file (for the libcudacxx test at least). 

This behavior is there on main branch as well, it can be replicated by calling `::CUdevice current_device = ::cuda::__driver::__ctxGetDevice();` before the first call to `thrust::sequence` in `libcudacxx.test.cuda.containers.buffer.transform`, it fails on some runs.

So either there is an issue with the contexts in these files (even on main branch), or it's just something like lazy initialization and the validate_stream_device should be called later to ensure device context is valid at that time

Edit: if i move the validation a little bit further in the dispatch functions, after calling PtxComputeCap or PtxVersion, it fixes the issue since PtxVersion calls some runtime api function and i think it takes care of initializing the device context if needed. 
```cpp
::cuda::compute_capability cc{};
if (const auto error = CubDebug(launcher_factory.PtxComputeCap(cc)))
{
  return error;
}

if (const auto error = CubDebug(detail::validate_stream_device(stream)))
{
  return error;
}
```
I can do that for every dispatch function, however i don't think it's a solid way of fixing that issue, is there a method i could call at the beginning of my validation to ensure device context initialization that wouldn't add overhead ?

bernhardmgruber:
@davebayer can you help out with answering questions on the device context initialization? Thx!

davebayer:
> the two other checks have tests (libcudacxx.test.cuda.containers.buffer.transform , cudax.example.async_buffer_add and cudax.example.cub_reduce ) that are failing because the device context is invalid at the beginning of the test. It's a non-deterministic behavior, that seems to depend on the order of execution of tests inside a same file (for the libcudacxx test at least).

The problem is that in libcu++, we delay primary context initialization (compared to CUDA Runtime, that does it whenever a CUDART API is called).

> if i move the validation a little bit further in the dispatch functions, after calling PtxComputeCap or PtxVersion, it fixes the issue since PtxVersion calls some runtime api function and i think it takes care of initializing the device context if needed.

Yes, let's do that

cub/cub/util_device.cuh
```diff
@@ -456,6 +457,62 @@ CUB_RUNTIME_FUNCTION inline cudaError_t SyncStream([[maybe_unused]] cudaStream_t
...
+        return error;
+      }
+
+      if (const auto error = ::cuda::__driver::__ctxPushNoThrow(stream_ctx); error != cudaSuccess)
+      {
+        return error;
+      }
+
+      ::CUdevice stream_device;
+      const auto get_device_error = ::cuda::__driver::__ctxGetDeviceNoThrow(stream_device);
```
davebayer, line 516: Please use `::cuda::__driver::__streamGetDevice` in CTK 13.0+

thom-gg:
Hello, alright thanks @davebayer for the feedback, should be good now !

thom-gg:
hey @davebayer  as you advised i used streamGetDevice(noThrow) for CTK>=13.0 which uses the driver method cuStreamGetDevice, and that makes some tests such as the ones in `cub.test.device.transform_env.lid_0` fail because cuStreamGetDevice is apparently not allowed during a stream graph capture

the workflow i was using before and that i kept for CTK <13.0 (__streamGetCtxNoThrow -> __ctxPushNoThrow -> __ctxGetDeviceNoThrow -> __ctxPopNoThrow) seems to work even during a stream capture

so should i remove the cuStreamGetDevice part and just use the regular workflow?

thom-gg:
> hey @davebayer as you advised i used streamGetDevice(noThrow) for CTK>=13.0 which uses the driver method cuStreamGetDevice, and that makes some tests such as the ones in `cub.test.device.transform_env.lid_0` fail because cuStreamGetDevice is apparently not allowed during a stream graph capture
> 
> the workflow i was using before and that i kept for CTK <13.0 (__streamGetCtxNoThrow -> __ctxPushNoThrow -> __ctxGetDeviceNoThrow -> __ctxPopNoThrow) seems to work even during a stream capture
> 
> so should i remove the cuStreamGetDevice part and just use the regular workflow?

@davebayer hello ! would it be possible to get an update regarding this please :) ?

davebayer:
Hi @thom-gg, I'm sorry, I missed the pings. Yes, please, remove that `cuStreamGetDevice` path, thank you!

thom-gg:
> Hi @thom-gg, I'm sorry, I missed the pings. Yes, please, remove that `cuStreamGetDevice` path, thank you!

hey, no problem, done :)

thom-gg:
> /ok to test [9c6a53f](https://github.com/NVIDIA/cccl/commit/9c6a53f19a7498d2c28c1e3d2a031b172408288a)

it's still going to fail, i'm fixing the MSVC errors right now (edit: done)

Issue #7455, opened 2026-02-03
Unify common functors and optionally expose them


bernhardmgruber:
We repeatedly define common functors all over CCCL, like `__return_constant` (at least 2 occurrences)., `equal_to_default_t` (4 occurrences), `always_true_predicate`, `equal_to_value`, `equal_to_t`, `always_false_t`, `always_true_t`, etc. Sometimes, the definitions are slightly different. We should pick a general enough version of those and put them somewhere central, like `<cuda/functional>`. They can remain internal, but we may also consider making them public.

The motivation is twofold: cleaner code, but also more opportunity for optmization. `cub::DeviceTransform` for example can optimize if the predicate is exactly `cub::detail::transform::always_true_predicate`. Any other predicate type (with the same definition) will not lead to an optimal code path.

thedevmystic:
Hello! I'm interested in working on this issue. Can you give me some brief information about this and what the expectations are?

miscco:
@thedevmystic thanks a lot for your interest 

Please give #7456 or #7457 a look. Those are some of the relevant sub-issues for relevant types we want to unify

thedevmystic:
I'm thinking of handling #7457 first, if it's fine with cccl.

Just few questions,
- can i document functions with doxygen or not?
- Do I've to implement and change the usage in one PR or not?

The second question is because I'm thinking implementing `cuda/__functional/equal_to_value.h` first in a single PR, then slowly refactoring its usage in individual PRs. So, PRs stays small and easy to review.

### TL;DR
**My plan:**
- implement `cuda/__functional/equal_to_value.h` (one PR).
- refactor its usage (other individual PRs).

bernhardmgruber:
@thedevmystic plesae add `cuda/__functional/equal_to_value.h` and the implementation as an internal functor for now, and replace all uses of it across CCCL in a single PR. The reason is that this gives the new implementation the necessary test coverage.

thedevmystic:
Okay! Just a heads up @bernhardmgruber, that these:

- equal_to_one
- equal_to_val
- equal_to_value
- equal_to_default_t
- equal_to_t

Are used at various files, so the PR will likely be like 10 or even more files changed.

KyleWang02:
[...]

PR #9120, opened 2026-05-23
[FEA] Unify is_even functor

### LINKED ISSUE #9638: [FEA] Unify `is_even` functors

thedevmystic:
This is a continuation of #7455.

This PR focuses on unifying `is_even` functor across CCCL testing and benchmarking environment.

## Quick Summary of Changed
- Moved `is_even` logic to `c2h/include/c2h/operator.cuh` and enclosed it in c2h namespace.
- changed scattered functors to utilize unified one.

## Note
This does not unify `cub/*` files, because they uses `custom_type`. If we have to unify them, then I'll.

(Comment from @bernhardmgruber in #7455 specified duplicates in cub/*)

Also, this is an issue from my end, my system has integrated M2 chip and testing CCCL on it, is either impossible or painfully slow. So, changes can contain few errors.

Fixes: #9638

### DIFF
--- a/cub/test/catch2_test_device_select_api.cu
+++ b/cub/test/catch2_test_device_select_api.cu
@@ -9,19 +9,11 @@
 #include <thrust/equal.h>
 #include <thrust/memory.h>
 
+#include <cuda/functional>
+
 #include <cstddef>
 
 #include <c2h/catch2_test_helper.h>
-
-// example-begin segmented-select-iseven
-struct is_even_t
-{
-  __host__ __device__ bool operator()(int flag) const
-  {
-    return !(flag % 2);
-  }
-};
-// example-end segmented-select-iseven
 
 C2H_TEST("cub::DeviceSelect::FlaggedIf works with int data elements", "[select][device]")
 {
@@ -31,7 +23,7 @@ C2H_TEST("cub::DeviceSelect::FlaggedIf works with int data elements", "[select][
   c2h::device_vector<int> d_flags = {8, 6, 7, 5, 3, 0, 9, 3};
   c2h::device_vector<int> d_out(num_items);
   c2h::device_vector<int> d_num_selected_out(num_items);
-  is_even_t is_even{};
+  cuda::__is_even is_even{};
 
   // Determine temporary device storage requirements
   size_t temp_storage_bytes = 0;
@@ -74,7 +66,7 @@ C2H_TEST("cub::DeviceSelect::FlaggedIf in-place works with int data elements", "
   c2h::device_vector<int> d_data  = {0, 1, 2, 3, 4, 5, 6, 7};
   c2h::device_vector<int> d_flags = {8, 6, 7, 5, 3, 0, 9, 3};
   c2h::device_vector<int> d_num_selected_out(num_items);
-  is_even_t is_even{};
+  cuda::__is_even is_even{};
 
   // Determine temporary device storage requirements
   size_t temp_storage_bytes = 0;
--- a/cub/test/catch2_test_device_select_api.cu
+++ b/cub/test/catch2_test_device_select_api.cu
@@ -15,16 +15,6 @@
 
 #include "cub_test_macros.h"
 
-// example-begin segmented-select-iseven
-struct is_even_t
-{
-  __host__ __device__ bool operator()(int flag) const
-  {
-    return !(flag % 2);
-  }
-};
-// example-end segmented-select-iseven
-
 CUB_TEST("cub::DeviceSelect::FlaggedIf works with int data elements", "[select][device]", CUB_SMALL)
 {
   // example-begin segmented-select-flaggedif
--- a/cub/test/catch2_test_device_select_api.cu
+++ b/cub/test/catch2_test_device_select_api.cu
@@ -9,6 +9,8 @@
 #include <thrust/equal.h>
 #include <thrust/memory.h>
 
+#include <cuda/functional>
+
 #include <cstddef>
 
 #include "cub_test_macros.h"
@@ -31,7 +33,7 @@ CUB_TEST("cub::DeviceSelect::FlaggedIf works with int data elements", "[select][
   c2h::device_vector<int> d_flags = {8, 6, 7, 5, 3, 0, 9, 3};
   c2h::device_vector<int> d_out(num_items);
   c2h::device_vector<int> d_num_selected_out(num_items);
-  is_even_t is_even{};
+  cuda::__is_even is_even{};
 
   // Determine temporary device storage requirements
   size_t temp_storage_bytes = 0;
@@ -74,7 +76,7 @@ CUB_TEST("cub::DeviceSelect::FlaggedIf in-place works with int data elements", "
   c2h::device_vector<int> d_data  = {0, 1, 2, 3, 4, 5, 6, 7};
   c2h::device_vector<int> d_flags = {8, 6, 7, 5, 3, 0, 9, 3};
   c2h::device_vector<int> d_num_selected_out(num_items);
-  is_even_t is_even{};
+  cuda::__is_even is_even{};
 
   // Determine temporary device storage requirements
   size_t temp_storage_bytes = 0;
--- a/cub/test/catch2_test_device_select_flagged_if.cu
+++ b/cub/test/catch2_test_device_select_flagged_if.cu
@@ -55,13 +55,9 @@ DECLARE_LAUNCH_WRAPPER(cub::DeviceSelect::FlaggedIf, select_flagged_if);
 
 using custom_t = c2h::custom_type_t<c2h::equal_comparable_t>;
 
-struct is_even_t
+struct is_even_t : cuda::__is_even
 {
-  template <typename T>
-  __host__ __device__ bool operator()(T const& elem) const
-  {
-    return cuda::__is_even{}(elem);
-  }
+  using __is_even::operator();
 
   template <template <typename> class... Policies>
   __host__ __device__ bool operator()(c2h::custom_type_t<Policies> elem) const
--- a/cub/test/catch2_test_device_select_flagged_if.cu
+++ b/cub/test/catch2_test_device_select_flagged_if.cu
@@ -66,7 +66,7 @@ struct is_even_t
   template <template <typename> class... Policies>
   __host__ __device__ bool operator()(c2h::custom_type_t<Policies> elem) const
   {
-   return cuda::__is_even<int>{}(elem.key);
+    return cuda::__is_even<int>{}(elem.key);
   }
 };
 
--- a/cub/test/catch2_test_device_select_flagged_if.cu
+++ b/cub/test/catch2_test_device_select_flagged_if.cu
@@ -10,6 +10,7 @@
 #include <thrust/logical.h>
 
 #include <cuda/devices>
+#include <cuda/functional>
 #include <cuda/iterator>
 #include <cuda/std/execution>
 
@@ -54,21 +55,18 @@ DECLARE_LAUNCH_WRAPPER(cub::DeviceSelect::FlaggedIf, select_flagged_if);
 
 using custom_t = c2h::custom_type_t<c2h::equal_comparable_t>;
 
-template <typename T>
 struct is_even_t
 {
+  template <typename T>
   __host__ __device__ bool operator()(T const& elem) const
   {
-    return !(elem % 2);
+    return cuda::__is_even<T>{}(elem);
   }
-};
 
-template <>
-struct is_even_t<custom_t>
-{
-  __host__ __device__ bool operator()(custom_t elem) const
+  template <template <typename> class... Policies>
+  __host__ __device__ bool operator()(c2h::custom_type_t<Policies> elem) const
   {
-    return !(elem.key % 2);
+   return cuda::__is_even<int>{}(elem.key);
   }
 };
 
--- a/cub/test/catch2_test_device_select_flagged_if.cu
+++ b/cub/test/catch2_test_device_select_flagged_if.cu
@@ -60,13 +60,13 @@ struct is_even_t
   template <typename T>
   __host__ __device__ bool operator()(T const& elem) const
   {
-    return cuda::__is_even<T>{}(elem);
+    return cuda::__is_even{}(elem);
   }
 
   template <template <typename> class... Policies>
   __host__ __device__ bool operator()(c2h::custom_type_t<Policies> elem) const
   {
-    return cuda::__is_even<int>{}(elem.key);
+    return cuda::__is_even{}(elem.key);
   }
 };
 
--- a/cub/test/catch2_test_device_select_flagged_if.cu
+++ b/cub/test/catch2_test_device_select_flagged_if.cu
@@ -170,7 +170,7 @@ C2H_TEST("DeviceSelect::FlaggedIf does not change input and is stable",
   c2h::device_vector<input_type> out(num_items);
   c2h::gen(C2H_SEED(2), in);
 
-  is_even_t<flag_type> is_even{};
+  is_even_t is_even{};
 
   c2h::device_vector<flag_type> flags(num_items);
   c2h::gen(C2H_SEED(1), flags);
@@ -211,7 +211,7 @@ C2H_TEST("DeviceSelect::FlaggedIf works with user provided memory and environmen
   c2h::device_vector<input_type> out(num_items, thrust::default_init);
   c2h::gen(C2H_SEED(2), in);
 
-  is_even_t<flag_type> is_even{};
+  is_even_t is_even{};
 
   c2h::device_vector<flag_type> flags(num_items, thrust::default_init);
   c2h::gen(C2H_SEED(1), flags);
@@ -322,7 +322,7 @@ C2H_TEST("DeviceSelect::FlaggedIf works in place with user provided memory and e
   c2h::device_vector<input_type> in(num_items, thrust::default_init);
   c2h::gen(C2H_SEED(2), in);
 
-  is_even_t<flag_type> is_even{};
+  is_even_t is_even{};
 
   c2h::device_vector<flag_type> flags(num_items, thrust::default_init);
   c2h::gen(C2H_SEED(1), flags);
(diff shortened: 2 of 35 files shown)


thedevmystic:
Most issues that the AI pointed out was already there, but if we gotta fix them, tell me.

thedevmystic:
@bernhardmgruber anything i can do?

c2h/include/c2h/operator.cuh
```diff
@@ -102,3 +103,34 @@ inline const c2h::custom_type_t<Policies...> identity_v<cuda::minimum<>, c2h::cu
...
+namespace c2h
+{
+/***********************************************************************************************************************
+ * Operator is_even check
+ **********************************************************************************************************************/
+
+using complex32 = cuda::std::complex<float>;
+using complex64 = cuda::std::complex<double>;
+
+struct is_even_fn
```
bernhardmgruber: Important: the suffix `_fn` is not established. Please stick with `_t`, or just omit a suffix and the global constant below. I prefer the latter.

c2h/include/c2h/operator.cuh
```diff
@@ -102,3 +103,34 @@ inline const c2h::custom_type_t<Policies...> identity_v<cuda::minimum<>, c2h::cu
...
 {};
+
+namespace c2h
+{
+/***********************************************************************************************************************
+ * Operator is_even check
+ **********************************************************************************************************************/
+
+using complex32 = cuda::std::complex<float>;
+using complex64 = cuda::std::complex<double>;
```
bernhardmgruber: Suggestion: Please avoid adding more aliases to `c2h` and just use the full type names below.

cub/test/catch2_test_device_select_api.cu
```diff
@@ -43,7 +33,7 @@ C2H_TEST("cub::DeviceSelect::FlaggedIf works with int data elements", "[select][
     d_out.begin(),
     d_num_selected_out.data(),
     num_items,
-    is_even);
+    c2h::is_even);
```
bernhardmgruber: Remark: Addressing the above comment would make this:
```suggestion
    c2h::is_even{});
```
which is fine I think.

c2h/include/c2h/operator.cuh
bernhardmgruber: @miscco where would be a good place in libcu++ to add this functor? I don't want it in C2H (which is for unit tests), since it's used in benchmarks.

miscco: We can put it into `cuda/__functional/is_even.h`

miscco: The question would be whether we want to make it public or not. 

If its public we should name it `is_even` if not then just the ugly `__is_even`

bernhardmgruber: I would stay with the ugly version for now. Nobody is asking for this functor.

thedevmystic: okay 👍

thedevmystic: so we gotta put is_even in `cuda/__functional/is_even.h` as __is_even. am I right? @bernhardmgruber? ~we also have other functors like square_t etc, we could use a single header file as something like `internal_functors.h` or something, but anyways whats the plan forward?~ (since I realized we use explicit one functor per file).

bernhardmgruber: > so we gotta put is_even in cuda/__functional/is_even.h as __is_even. am I right?

yes

bernhardmgruber: The longer rational is that I want to have as many functors in libcu++ as possible, since they are neither CUB nor Thrust specific. C2H headers are for unit testing only, nvbench_helper is benchmark only. So they all only cover a part of the code base. Only libcu++ is universally available.

thedevmystic: @bernhardmgruber I noticed that original codebase didnt guard against floating points and used simple `static_cast<int>(value) % 2 == 0` mainly, this means 2.2 returns true while mathematically it should return false. What should I do? Should I go along without guarding, or add guard (using is_integral), or return false directly for floating points (well with checking for numbers like 2.0)?

bernhardmgruber: Do we need `is_even` for floating point types? If not, I would just constrain `T` on the `operator()` to an integral. Then we also don't need the cast I think.

thedevmystic: We have this in `copy_if/basic.cu`:

```cpp
  __device__ constexpr bool operator()(const complex32& val) const noexcept
  {
    return static_cast<int>(val.real()) % 2 == 0;
  }

  __device__ constexpr bool operator()(const complex64& val) const noexcept
  {
    return static_cast<int>(val.real()) % 2 == 0;
  }
```

And `val.real()` is floating point. So, It is needed, I guess. Am I right, @bernhardmgruber?

bernhardmgruber: I checked those two benchmarks and they use:
```
NVBENCH_BENCH_TYPES(basic, NVBENCH_TYPE_AXES(fundamental_types))
```
and `fundamental_types` does not include complex. I think we can just remove the handling of `complex[32|64]` here.

thedevmystic: got it!

bernhardmgruber: Filed a PR: #9873

thedevmystic: A quick question, @bernhardmgruber. The file 'catch2_test_device_select_flagged_if.cu' does this:

```cpp
using custom_t = c2h::custom_type_t<c2h::equal_comparable_t>;

template <>
struct is_even_t<custom_t>
{
  __host__ __device__ bool operator()(custom_t elem) const
  {
    return !(elem.key % 2);
  }
};
```

But I have constrained it to integrals:

```cpp
struct __is_even
{
  _CCCL_TEMPLATE(class _Tp)
  _CCCL_REQUIRES(::cuda::std::is_integral_v<_Tp>)
  [[nodiscard]] _CCCL_API _CCCL_HOST_DEVICE constexpr bool operator()(const _Tp& __value) const noexcept
  {
    return (__value & 1) == 0;
  }
};
```

This contradicts. What's should be done?

bernhardmgruber: Can you add an overload to `is_even_t` for `custom_t`?

thedevmystic: I wrote this, does this satisfy?

```cpp
using custom_t = c2h::custom_type_t<c2h::equal_comparable_t>;

struct is_even_t
{
  template <typename T>
  __host__ __device__ bool operator()(T const& elem) const
  {
    return cuda::__is_even<T>{}(elem);
  }

  __host__ __device__ bool operator()(custom_t elem) const
  {
   return cuda::__is_even<int>{}(elem.key);
  }
};
```

bernhardmgruber: Let's do:
```c++
struct is_even_t
{
  template <typename T>
  __host__ __device__ bool operator()(T const& elem) const
  {
    return cuda::__is_even<T>{}(elem);
  }

  template <template <typename> class... Policies>
  __host__ __device__ bool operator()(custom_type_t<Policies> elem) const
  {
   return cuda::__is_even<int>{}(elem.key);
  }
};
```

thedevmystic:
I added the requested changes.

bernhardmgruber:
Please rebase your branch

thedevmystic:
Done :+1:

libcudacxx/include/cuda/__functional/is_even.h
```diff
@@ -0,0 +1,56 @@
...
+#  pragma system_header
+#endif // no system header
+
+#include <cuda/std/__type_traits/is_integral.h>
+
+#include <cuda/std/__cccl/prologue.h>
+
+_CCCL_BEGIN_NAMESPACE_CUDA
+
+//! @brief `is_even` is a function object that checks if a value is even (const).
```
bernhardmgruber:
```suggestion
//! @brief `is_even` is a function object that checks if an integral value is even.
```

libcudacxx/include/cuda/__functional/is_even.h
```diff
@@ -0,0 +1,56 @@
...
+//! @brief `is_even_no_const` is a function object that checks if a value is even (non-const).
+struct __is_even_no_const
+{
+  _CCCL_TEMPLATE(class _Tp)
+  _CCCL_REQUIRES(::cuda::std::is_integral_v<_Tp>)
+  [[nodiscard]] _CCCL_API _CCCL_HOST_DEVICE constexpr bool operator()(const _Tp& __value) /* no const */ noexcept
+  {
+    return (__value & 1) == 0;
+  }
+};
```
bernhardmgruber: Important: this is just used in a single case. Let's not generalize this version. Remove it and keep it in the one test where it's used.

libcudacxx/benchmarks/bench/copy_if/basic.cu
bernhardmgruber: Important: Please remove. Also check whether there are more occurrences that have not been deleted.

thrust/benchmarks/bench/remove_if/basic.cu
bernhardmgruber: Important: remove

thedevmystic:
@bernhardmgruber, in file `libcudacxx/test/libcudacxx/std/numerics/simd/simd_test_utils.h` I wrapped that is_even to just use `cuda::__is_even`.

My reasoning was that since it is a utils file we can just wrap. Tell me if we want to remove but removing it will cause to update multiple simd related. I suggest keeping it, but what's the call.

thedevmystic:
I think that's all.

cub/test/catch2_test_device_select_flagged_if.cu
```diff
@@ -54,21 +55,18 @@ DECLARE_LAUNCH_WRAPPER(cub::DeviceSelect::FlaggedIf, select_flagged_if);
...
-struct is_even_t<custom_t>
-{
-  __host__ __device__ bool operator()(custom_t elem) const
+  template <template <typename> class... Policies>
+  __host__ __device__ bool operator()(c2h::custom_type_t<Policies> elem) const
   {
-    return !(elem.key % 2);
+    return cuda::__is_even<int>{}(elem.key);
   }
 };
```
bernhardmgruber, line 69: Suggestion:
```c++
struct is_even_t : cuda::__is_even
{
  using __is_even::operator();

  template <template <typename> class... Policies>
  __host__ __device__ bool operator()(c2h::custom_type_t<Policies> elem) const
  {
    return cuda::__is_even<int>{}(elem.key);
  }
};
```

libcudacxx/benchmarks/bench/copy_if/basic.cu
```diff
@@ -41,7 +33,8 @@ static void basic(nvbench::state& state, nvbench::type_list<T>)
 
   state.exec(
     nvbench::exec_tag::gpu | nvbench::exec_tag::no_batch | nvbench::exec_tag::sync, [&](nvbench::launch& launch) {
-      do_not_optimize(cuda::std::copy_if(cuda_policy(alloc, launch), in.begin(), in.end(), out.begin(), is_even{}));
+      do_not_optimize(
+        cuda::std::copy_if(cuda_policy(alloc, launch), in.begin(), in.end(), out.begin(), cuda::__is_even<T>{}));
```
bernhardmgruber: Q: How does this even compile? `__is_even` does not take a template argument. Please update all occurrences.
```suggestion
      do_not_optimize(
        cuda::std::copy_if(cuda_policy(alloc, launch), in.begin(), in.end(), out.begin(), cuda::__is_even{}));
```

thedevmystic: sorry, my bad.

thedevmystic:
I think I fixed the template argument problem.

libcudacxx/benchmarks/bench/copy_if/basic.cu
bernhardmgruber: no worries!

thedevmystic:
I see this gives error:
```cpp
struct is_even_t : cuda::__is_even
{
  using __is_even::operator();

  template <template <typename> class... Policies>
  __host__ __device__ bool operator()(c2h::custom_type_t<Policies> elem) const
  {
    return cuda::__is_even<int>{}(elem.key);
  }
};
```
With error:
```
parameter pack Policies was referenced but not expanded.
```

bernhardmgruber:
> I see this gives error:
> parameter pack Policies was referenced but not expanded.

So expand it: `c2h::custom_type_t<Policies...>` ;)

thedevmystic:
okay, just felt like notifying, lol! Also after seeing libcu++ errors it seems like it uses it is passing some ref-qualified value via indirect_unary_predicate in partition function, so I'll add remove_cvref in it.

thedevmystic:
it seems like indirect_unary_predicate is failing because `__is_even` is not invocable with `thrust::device_reference<T>>` but the test file passes it. So, it seems like we have to drop the `is_integral` constraint. Should I proceed with it?

bernhardmgruber:
> it seems like indirect_unary_predicate is failing because `__is_even` is not invocable with `thrust::device_reference<T>>` but the test file passes it. So, it seems like we have to drop the `is_integral` constraint. Should I proceed with it?

Hmm. Fine. Let's try removing it.

thedevmystic:
removed it and rebased

thedevmystic:
Sorry for vanishing, lol! I synced the PR. Let me fix the issues too.

thedevmystic:
Multiple SIMD related tests failed. Is it due to using `cuda::__is_even`?

```cpp
#ifndef SIMD_TEST_UTILS_H
#define SIMD_TEST_UTILS_H

+#include <cuda/functional>
 #include <cuda/std/__simd_>
 #include <cuda/std/array>
 #include <cuda/std/cmath>
@@ -53,7 +54,7 @@ struct is_even
   template <typename I>
   TEST_HOST_DEVICE_FUNC constexpr bool operator()(I i) const noexcept
   {
-    return i % 2 == 0;
+    return cuda::__is_even(i);
   }
 };

```

this is the only change in simd region of tests and PR's scope. Wanna change it back, @bernhardmgruber?

thedevmystic:
Fixed:
- Unused template `T` var in many tests
- Cub library incorrect declaration
- SIMD Utils issue

I may have a question @bernhardmgruber. Why does Unused variable declaration passes on many tasks but fails on few, shouldn't it fail on all tasks of that particular sub-library?

libcudacxx/include/cuda/__functional/is_even.h
```diff
@@ -0,0 +1,42 @@
...
+
+_CCCL_BEGIN_NAMESPACE_CUDA
+
+//! @brief `__is_even` is a function object that checks if an integral value is even.
+struct __is_even
+{
+  template <class _Tp>
+  [[nodiscard]] _CCCL_API constexpr bool operator()(const _Tp& __value) const noexcept
+  {
+    return (static_cast<long long>(__value) & 1) == 0;
```
griwes: This is one of those cases where I wonder if `% 2 == 0` would be better, honestly. If we just say `% 2 == 0`,  any reasonable compiler is going to optimize it down to the same code as `(... & 1) == 0`. Also, any reason for this to be `long long` and not just `int`? Probably shouldn't go smaller, but `int` in the "native-ish word size" does actually fit here.

thedevmystic: Well, both `(... & 1)` and `(... % 2)` works. I generally use `(... & 1)`, so I used it here too, and prior reviewers didn't asked to change it. and for using `long long` instead of `int` is because prior test cases' `is_even` functors used a collection of `int` and `long long`, so I simply choose the biggest size to fit all of them. and using `long long` doesn't cause any issue and will work perfectly fine.

bernhardmgruber: Oh, if we can, we should definitely use
```suggestion
    return __value % 2 == 0;
```
Do we even need the cast?

bernhardmgruber:
> Why does Unused variable declaration passes on many tasks but fails on few, shouldn't it fail on all tasks of that particular sub-library?

`¯\_(ツ)_/¯`

thedevmystic:
updated to use modulo.

libcudacxx/include/cuda/__functional/is_even.h
```diff
@@ -0,0 +1,42 @@
...
+#  pragma clang system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_MSVC)
+#  pragma system_header
+#endif // no system header
+
+#include <cuda/std/__cccl/prologue.h>
+
+_CCCL_BEGIN_NAMESPACE_CUDA
+
+//! @brief `__is_even` is a function object that checks if an integral value is even.
```
bernhardmgruber, line 28: Q: Do we need the `integral` here? The `% 2` should work for floats as well now, right?

thrust/testing/cuda/is_partitioned.cu
```diff
@@ -111,12 +104,11 @@ void TestIsPartitionedCudaStreams()
...
 
-template <typename T>
-struct is_even_non_const
+struct is_even_no_const
 {
-  _CCCL_HOST_DEVICE bool operator()(T x) // no const
+  [[nodiscard]] _CCCL_HOST_DEVICE_API constexpr bool operator()(const int x) noexcept
   {
-    return ((int) x % 2) == 0;
+    return (x & 1) == 0;
```
bernhardmgruber: Suggestion: Let's stay with `% 2` here as well:

```suggestion
    return x % 2 == 0;
```

bernhardmgruber (commented):
Two more nits, otherwise LGTM

libcudacxx/include/cuda/__functional/is_even.h
thedevmystic: No it will not work. Unless we `static_cast` value to `int`.

Here `a` is a `float`:
```text
code.cpp:5:20: error: invalid operands to binary expression ('float' and 'int')
    5 |   bool is_even = a % 2 == 0;
      |                  ~ ^ ~
1 error generated.
```

And again if we cast 4.45 (which is not even) will be falsely reported even.

thrust/testing/cuda/is_partitioned.cu
thedevmystic: okay

libcudacxx/include/cuda/__functional/is_even.h
bernhardmgruber: Right. Sorry for the brain split.

bernhardmgruber (approved):

thedevmystic:
fixed a missing header file in `cuda/std/algorithm/partition.pass.cpp` that was causing last errors in libcudaxx.

thedevmystic:
Is there anything else to add?

PR #9123, opened 2026-05-25
Distribute `std::tuple_(size|element)` specializations to implementation files

davebayer:
I want to implement tuple protocol for `integer_sequence`, but there is a non-trivial circular dependency, because we specialize all relevant `std::` traits in one file, which includes `make_index_sequence`.

So, I decided to decentralize the specializations and move them to the places where we specialize the `cuda::std::` traits. 

Note that we already specialize for example `std::formatter` inside the implementation files without any problems.


miscco (approved):
Looks good, I believe with the added forward declarations this is now the better approach

PR #9124, opened 2026-05-25
use `std::scoped_lock`

### LINKED ISSUE #8632: [FEA] Use `std::scoped_lock` instead of `std::lock_guard`
`clang-tidy`'s `modernize-use-scoped-lock` option suggests using `std::scoped_lock` instead of `std::lock_guard`. All host compilers implement that type, so we should be able to that one.

I'm just not sure if there are any drawbacks to using `std::scoped_lock`.

charan-003:
Use `std::scoped_lock` instead of `std::lock_guard`

Closes ##8632


davebayer (approved):
LGTM, thank you for the contribution!

bernhardmgruber (approved):

miscco (approved):

PR #9127, opened 2026-05-26
Hardcode `DeviceTransform` `OffsetT` to `int64`

### LINKED ISSUE #8805: Consider hardcoding a 64-bit offset type for `DeviceTransform`

bernhardmgruber:
This change is going to cause a tiny regression (benchmarks in #8805, important ones repeated below). However, it:

* halfs compilation time for `thrust::transform` (and any other Thrust call using `DeviceTransform`, because we got rid of the index type dispatch
* halfs compilation time of all `DeviceTransform` tests
* halfs benchmark time for `DeviceTransform`
* doubles tuning throughput for `DeviceTransform`

I am gladly eating a small regression for these benefits. Here are the regressions:

A100 (1/200 workloads regressed):
```
# triad

## [0] NVIDIA A100-SXM4-80GB

|  T{ct}  |  OffsetT{ct}  |  Elements{io}  |   Ref Time |   Ref Noise |   Cmp Time |   Cmp Noise |     Diff |   %Diff |  Status  |
|---------|---------------|----------------|------------|-------------|------------|-------------|----------|---------|----------|
|   F64   |      I64      |      2^24      | 243.144 us |       1.02% | 245.588 us |       0.38% | 2.444 us |   1.01% |  🔴 SLOW  |
```

H200 (1/200 workloads regressed):
```
# nstream

## [0] NVIDIA H200 NVL

|   F32   |      I64      |      2^20      |   9.963 us |       1.47% |  10.443 us |       6.24% |  0.480 us |   4.82% |  🔴 SLOW  |
```

B200 (4/200 workloads regressed):
```
# mul

## [0] NVIDIA B200

|   I8    |      I32      |      2^28      | 116.430 us |       0.54% | 117.871 us |       0.70% |  1.442 us |   1.24% |  🔴 SLOW  |

# triad

## [0] NVIDIA B200

|  T{ct}  |  OffsetT{ct}  |  Elements{io}  |   Ref Time |   Ref Noise |   Cmp Time |   Cmp Noise |     Diff |   %Diff |  Status  |
|---------|---------------|----------------|------------|-------------|------------|-------------|----------|---------|----------|
|   I8    |      I32      |      2^28      | 153.406 us |       0.32% | 155.685 us |       0.25% | 2.280 us |   1.49% |  🔴 SLOW  |

# nstream

## [0] NVIDIA B200

|  T{ct}  |  OffsetT{ct}  |  Elements{io}  |   Ref Time |   Ref Noise |   Cmp Time |   Cmp Noise |     Diff |   %Diff |  Status  |
|---------|---------------|----------------|------------|-------------|------------|-------------|----------|---------|----------|
|   I8    |      I32      |      2^28      | 208.966 us |       0.18% | 212.496 us |       0.36% | 3.531 us |   1.69% |  🔴 SLOW  |
|   I16   |      I32      |      2^28      | 317.133 us |       0.23% | 322.522 us |       0.29% | 5.389 us |   1.70% |  🔴 SLOW  |
```

Let me know if we need more benchmarks, like on non-data-center-grade GPUs.

Fixes: #8805

### DIFF
--- a/cub/benchmarks/bench/transform/babelstream.cu
+++ b/cub/benchmarks/bench/transform/babelstream.cu
@@ -55,16 +55,11 @@ inline constexpr auto startScalar = -2; // BabelStream: 0.4
 
 static_assert(startA == (startA + startB + startScalar * startC), "nstream must have a consistent workload");
 
-template <typename T, typename OffsetT>
-static void mul(nvbench::state& state, nvbench::type_list<T, OffsetT>)
+template <typename T>
+static void mul(nvbench::state& state, nvbench::type_list<T>)
 try
 {
   const auto n = state.get_int64("Elements{io}");
-  if (sizeof(OffsetT) == 4 && n > std::numeric_limits<OffsetT>::max())
-  {
-    state.skip("Skipping: input size exceeds 32-bit offset type capacity.");
-    return;
-  }
 
   thrust::device_vector<T> b(n, startB);
   thrust::device_vector<T> c(n, startC);
@@ -74,31 +69,25 @@ try
   state.add_global_memory_writes<T>(n);
 
   const T scalar = startScalar;
-  bench_transform(
-    state, ::cuda::std::tuple{c.begin()}, b.begin(), static_cast<OffsetT>(n), [=] _CCCL_DEVICE(const T& ci) {
-      return ci * scalar;
-    });
+  bench_transform(state, ::cuda::std::tuple{c.begin()}, b.begin(), n, [=] _CCCL_DEVICE(const T& ci) {
+    return ci * scalar;
+  });
 }
 catch (const std::bad_alloc&)
 {
   state.skip("Skipping: out of memory.");
 }
 
-NVBENCH_BENCH_TYPES(mul, NVBENCH_TYPE_AXES(element_types, offset_types))
+NVBENCH_BENCH_TYPES(mul, NVBENCH_TYPE_AXES(element_types))
   .set_name("mul")
-  .set_type_axes_names({"T{ct}", "OffsetT{ct}"})
+  .set_type_axes_names({"T{ct}"})
   .add_int64_power_of_two_axis("Elements{io}", array_size_powers);
 
-template <typename T, typename OffsetT>
-static void add(nvbench::state& state, nvbench::type_list<T, OffsetT>)
+template <typename T>
+static void add(nvbench::state& state, nvbench::type_list<T>)
 try
 {
   const auto n = state.get_int64("Elements{io}");
-  if (sizeof(OffsetT) == 4 && n > std::numeric_limits<OffsetT>::max())
-  {
-    state.skip("Skipping: input size exceeds 32-bit offset type capacity.");
-    return;
-  }
 
   thrust::device_vector<T> a(n, startA);
   thrust::device_vector<T> b(n, startB);
@@ -108,11 +97,7 @@ try
   state.add_global_memory_reads<T>(2 * n);
   state.add_global_memory_writes<T>(n);
   bench_transform(
-    state,
-    ::cuda::std::tuple{a.begin(), b.begin()},
-    c.begin(),
-    static_cast<OffsetT>(n),
-    [] _CCCL_DEVICE(const T& ai, const T& bi) -> T {
+    state, ::cuda::std::tuple{a.begin(), b.begin()}, c.begin(), n, [] _CCCL_DEVICE(const T& ai, const T& bi) -> T {
       return ai + bi;
     });
 }
@@ -121,21 +106,16 @@ catch (const std::bad_alloc&)
   state.skip("Skipping: out of memory.");
 }
 
-NVBENCH_BENCH_TYPES(add, NVBENCH_TYPE_AXES(element_types, offset_types))
+NVBENCH_BENCH_TYPES(add, NVBENCH_TYPE_AXES(element_types))
   .set_name("add")
-  .set_type_axes_names({"T{ct}", "OffsetT{ct}"})
+  .set_type_axes_names({"T{ct}"})
   .add_int64_power_of_two_axis("Elements{io}", array_size_powers);
 
-template <typename T, typename OffsetT>
-static void triad(nvbench::state& state, nvbench::type_list<T, OffsetT>)
+template <typename T>
+static void triad(nvbench::state& state, nvbench::type_list<T>)
 try
 {
   const auto n = state.get_int64("Elements{io}");
-  if (sizeof(OffsetT) == 4 && n > std::numeric_limits<OffsetT>::max())
-  {
-    state.skip("Skipping: input size exceeds 32-bit offset type capacity.");
-    return;
-  }
 
   thrust::device_vector<T> a(n, startA);
   thrust::device_vector<T> b(n, startB);
@@ -146,11 +126,7 @@ try
   state.add_global_memory_writes<T>(n);
   const T scalar = startScalar;
   bench_transform(
-    state,
-    ::cuda::std::tuple{b.begin(), c.begin()},
-    a.begin(),
-    static_cast<OffsetT>(n),
-    [=] _CCCL_DEVICE(const T& bi, const T& ci) {
+    state, ::cuda::std::tuple{b.begin(), c.begin()}, a.begin(), n, [=] _CCCL_DEVICE(const T& bi, const T& ci) {
       return bi + scalar * ci;
     });
 }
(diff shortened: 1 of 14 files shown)


cub/benchmarks/bench/transform/common.h
```diff
@@ -75,11 +75,11 @@ struct policy_selector
 };
 #endif // !TUNE_BASE
 
-template <typename OffsetT, typename... RandomAccessIteratorsIn, typename RandomAccessIteratorOut, typename TransformOp>
+template <typename... RandomAccessIteratorsIn, typename RandomAccessIteratorOut, typename TransformOp>
 void bench_transform(nvbench::state& state,
                      ::cuda::std::tuple<RandomAccessIteratorsIn...> inputs,
                      RandomAccessIteratorOut output,
-                     OffsetT num_items,
+                     ::cuda::std::size_t num_items,
```
miscco: Shouldnt this be `int64_t`?

cub/cub/device/dispatch/kernels/kernel_transform.cuh
```diff
@@ -134,7 +134,7 @@ _CCCL_DEVICE void transform_kernel_prefetch(
 {
   const int tile_size   = ThreadsPerBlock * num_elem_per_thread;
   const Offset offset   = static_cast<Offset>(blockIdx.x) * tile_size;
-  const int valid_items = static_cast<int>((::cuda::std::min) (num_items - offset, Offset{tile_size}));
+  const int valid_items = static_cast<int>((::cuda::std::min) (num_items - offset, static_cast<Offset>(tile_size)));
```
miscco, line 137: Why do we still have a template `Offset` when it should always be `int64_t`?

cub/benchmarks/bench/transform/babelstream.cu
```diff
@@ -74,31 +69,25 @@ try
   state.add_global_memory_writes<T>(n);
 
   const T scalar = startScalar;
-  bench_transform(
-    state, ::cuda::std::tuple{c.begin()}, b.begin(), static_cast<OffsetT>(n), [=] _CCCL_DEVICE(const T& ci) {
-      return ci * scalar;
-    });
+  bench_transform(state, ::cuda::std::tuple{c.begin()}, b.begin(), n, [=] _CCCL_DEVICE(const T& ci) {
+    return ci * scalar;
+  });
```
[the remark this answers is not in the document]
bernhardmgruber: Benchmarks can use lambdas

cub/cub/device/dispatch/dispatch_transform.cuh
miscco, line 173: Ditto: Why is `Offset` still needed as a template argument?

miscco (approved):
Looks good to me, the regressions are small enough IMHO to be ignored.

I would also be curious if we could get rid of them by tuning with just the 64 bit offset

cub/benchmarks/bench/transform/common.h
bernhardmgruber: I changed the public API again to use a templated `NumItemsT` parameter, so a user does not suddenly get conversion warnings. I only cast to `int64` before calling the dispatch layer now.

bernhardmgruber: Changed the occurrence here to `int64`

cub/cub/device/dispatch/kernels/kernel_transform.cuh
bernhardmgruber: I left the dispatch and kernel as is for now, since I want to play a bit with`uint64_t` in the future. It did give a speedup in some runs, so I want to know why and whether we can extract that speedup somehow #8805 by using unsigned types in a few places in the kernel.

cub/cub/device/dispatch/dispatch_transform.cuh
bernhardmgruber: Same reason as above

cub/test/catch2_test_device_transform.cu
```diff
@@ -29,15 +29,12 @@ DECLARE_LAUNCH_WRAPPER(cub::DeviceTransform::TransformStableArgumentAddresses, t
...
-
 C2H_TEST("DeviceTransform::Transform BabelStream add",
          "[device][transform]",
-         c2h::type_list<std::uint8_t, std::uint16_t, std::uint32_t, std::uint64_t, uchar3>,
-         offset_types)
+         c2h::type_list<std::uint8_t, std::uint16_t, std::uint32_t, std::uint64_t, uchar3>)
 {
   using type     = c2h::get<0, TestType>;
-  using offset_t = c2h::get<1, TestType>;
+  using offset_t = ::cuda::std::int64_t;
```
NaderAlAwar: Nit:
```suggestion
  using offset_t = cuda::std::int64_t;
```

cub/test/catch2_test_device_transform_if.cu
```diff
@@ -13,15 +13,12 @@
...
-
 C2H_TEST("DeviceTransform::TransformIf conditional BabelStream add",
          "[device][transform_if]",
-         c2h::type_list<std::uint8_t, std::uint16_t, std::uint32_t, std::uint64_t>,
-         offset_types)
+         c2h::type_list<std::uint8_t, std::uint16_t, std::uint32_t, std::uint64_t>)
 {
   using type     = c2h::get<0, TestType>;
-  using offset_t = c2h::get<1, TestType>;
+  using offset_t = ::cuda::std::int64_t;
```
NaderAlAwar, line 21: Nit:
```suggestion
  using offset_t = cuda::std::int64_t;
```

cub/benchmarks/bench/transform/babelstream.cu
```diff
@@ -74,31 +69,25 @@ try
   state.add_global_memory_writes<T>(n);
 
   const T scalar = startScalar;
-  bench_transform(
-    state, ::cuda::std::tuple{c.begin()}, b.begin(), static_cast<OffsetT>(n), [=] _CCCL_DEVICE(const T& ci) {
-      return ci * scalar;
-    });
+  bench_transform(state, ::cuda::std::tuple{c.begin()}, b.begin(), n, [=] _CCCL_DEVICE(const T& ci) {
```
NaderAlAwar: Nit: also applies to other locations in this source file and other benchmarks
```suggestion
  bench_transform(state, cuda::std::tuple{c.begin()}, b.begin(), n, [=] _CCCL_DEVICE(const T& ci) {
```

thrust/thrust/system/cuda/detail/transform.h
```diff
@@ -148,19 +148,13 @@ OutputIt _CCCL_HOST_DEVICE_API _CCCL_FORCEINLINE cub_transform_many(
   if constexpr (::cuda::proclaims_copyable_arguments<Predicate>::value
                 && ::cuda::proclaims_copyable_arguments<TransformOp>::value)
   {
-    THRUST_INDEX_TYPE_DISPATCH(
-      status,
-      (CUB_NS_QUALIFIER::DeviceTransform::TransformIf),
-      num_items,
-      (firsts, result, num_items_fixed, pred, transform_op, cuda_cub::stream(policy)));
+    status = CUB_NS_QUALIFIER::DeviceTransform::TransformIf(
```
NaderAlAwar, line 154: Critical: this changes behavior. `THRUST_INDEX_TYPE_DISPATCH` also called `_THRUST_INDEX_TYPE_DISPATCH_GUARD_UNDERFLOW` which would throw an exception if `num_items` was negative.

cub/cub/device/device_transform.cuh
```diff
@@ -70,14 +70,17 @@ struct DeviceTransform
...
   {
-    using choose_offset_t = detail::choose_signed_offset<NumItemsT>;
-    using offset_t        = typename choose_offset_t::type;
-
-    // Check if the number of items exceeds the range covered by the selected signed offset type
-    if (const cudaError_t error = choose_offset_t::is_exceeding_offset_type(num_items); error != cudaSuccess)
+    // We use int64_t internally, since it's faster than uint64_t and similar to a 32-bit offset type. See
+    // https://github.com/NVIDIA/cccl/issues/8805 for data. We need to check if it can hold the value passed by the
+    // user.
+    using offset_t = ::cuda::std::int64_t;
```
NaderAlAwar, line 76: Suggestion: `choose_signed_offset<NumItemsT>` rejected certain types for `NumItemsT`, like `bool` for instance, so this changes behavior. It would be good to add a static assert for what types are allowed.

bernhardmgruber: I switched back to using `choose_signed_offset` but retained hardcoding `offset_t`.

thrust/thrust/system/cuda/detail/transform.h
bernhardmgruber: I was kind of leaning into: This cannot happen, since a `last - first` iterator difference cannot be negative (it would violate the precondition that `first <= last`). But then it does not hurt to have the check. I added it back.

bernhardmgruber:
@NaderAlAwar thanks for the review and the good findings! I addressed your comments. Please have another look!

NaderAlAwar (approved):

PR #9128, opened 2026-05-26
Make more warpspeed scan stage counts tunable

bernhardmgruber:


cub/cub/device/dispatch/kernels/kernel_scan_warpspeed.cuh
```diff
@@ -102,7 +102,12 @@ allocResources(warpspeed::SyncHandler& syncHandler, warpspeed::SmemAllocator& sm
   using InOutT            = typename ScanResourcesT::InOutT;
   using SumThreadAndWarpT = typename ScanResourcesT::SumThreadAndWarpT;
 
-  const auto [num_block_idx_stages, num_sum_exclusive_cta_stages] = make_scan_stage_counts(numStages);
+  constexpr auto policy = get_warpspeed_policy<PolicySelector>();
+
+  const int num_block_idx_stages =
+    policy.block_idx_stages > 0 ? policy.block_idx_stages : ::cuda::std::max(1, numStages + policy.block_idx_stages);
+  const int num_sum_exclusive_cta_stages =
+    policy.lookback_stages > 0 ? policy.lookback_stages : ::cuda::std::max(1, numStages + policy.lookback_stages);
```
NaderAlAwar, line 110: Suggestion: this is duplicated from tuning_scan.cuh. Is it possible to move this to a common header? Would help avoid issues later if the logic changes

NaderAlAwar (approved):

cub/cub/device/dispatch/kernels/kernel_scan_warpspeed.cuh
bernhardmgruber: Yeah, there is unfortunately a lot more duplication here, which is needed to support CCCL.C. The entire logic of `smem_for_stages` is a duplication of the kernel setup code, but written in a way to be `constexpr` and type erased. I can't wait for the host JIT approach to ditch all of that.

So I have no intention to address your comment here, it's a wasted effort IMO. We will drop this all the second we can stop supporting CCCL.C.

cub/cub/device/dispatch/tuning/tuning_scan.cuh
```diff
@@ -590,6 +590,19 @@ struct scan_warpspeed_policy
...
+  // Therefore, a value of 0 just takes the number of stages.
+  // TODO(bgruber): should we rather have two values? A stage bias (additive) and a stage scale (multiplicative)?
+
+  // We do not need too many stages for lookback since the lookback warp is the bottleneck. As soon as it produces a new
+  // value, it will be consumed by the scanStore squad, releasing the stage. So just always use 2 stages.
+  int lookback_stages = 2; // TODO(bgruber): make this tunable
+
+  // If one less than the number of stages, we find a small speedup compared to setting it equal to num_stages. Not sure
+  // why.
+  int block_idx_stages = -1; // TODO(bgruber): make this tunable
```
bernhardmgruber:
```suggestion
  int lookback_stages = 2;

  // If one less than the number of stages, we find a small speedup compared to setting it equal to num_stages. Not sure
  // why.
  int block_idx_stages = -1;
```

PR #9129, opened 2026-05-26
[libcu++] Implement tuple protocol for `integer_sequence`

### LINKED ISSUE #10376: [FEA]: P1789R3 Library Support for Expansion Statements
This issue can be closed by implementing [P1789R3](https://wg21.link/P1789R3)

davebayer:
This PR implements tuple protocol for `integer_sequence` ([P1789R3](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2025/p1789r3.pdf)) from C++26

### DIFF
--- a/libcudacxx/include/cuda/std/__utility/integer_sequence.h
+++ b/libcudacxx/include/cuda/std/__utility/integer_sequence.h
@@ -3,7 +3,7 @@
 // Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
 // See https://llvm.org/LICENSE.txt for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
-// SPDX-FileCopyrightText: Copyright (c) 2023 NVIDIA CORPORATION & AFFILIATES.
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES.
 //
 //===----------------------------------------------------------------------===//
 
@@ -20,8 +20,11 @@
 #  pragma system_header
 #endif // no system header
 
+#include <cuda/std/__cstddef/types.h>
+#include <cuda/std/__tuple_dir/tuple_element.h>
+#include <cuda/std/__tuple_dir/tuple_size.h>
+#include <cuda/std/__type_traits/integral_constant.h>
 #include <cuda/std/__type_traits/is_integral.h>
-#include <cuda/std/cstddef>
 
 #include <cuda/std/__cccl/prologue.h>
 
@@ -242,7 +245,63 @@ using make_index_sequence = make_integer_sequence<size_t, _Np>;
 template <class... _Tp>
 using index_sequence_for = make_index_sequence<sizeof...(_Tp)>;
 
+// specialize cuda::std::tuple_size and cuda::std::tuple_element for cuda::std::integer_sequence
+
+template <class _Tp, _Tp... _Indices>
+struct tuple_size<integer_sequence<_Tp, _Indices...>> : integral_constant<size_t, sizeof...(_Indices)>
+{};
+
+template <size_t _Ip, class _Tp, _Tp... _Indices>
+struct tuple_element<_Ip, integer_sequence<_Tp, _Indices...>>
+{
+  static_assert(_Ip < sizeof...(_Indices),
+                "Index out of bounds in cuda::std::tuple_element<> (cuda::std::integer_sequence)");
+  using type = _Tp;
+};
+
+template <size_t _Ip, class _Tp, _Tp... _Indices>
+struct tuple_element<_Ip, const integer_sequence<_Tp, _Indices...>>
+{
+  static_assert(_Ip < sizeof...(_Indices),
+                "Index out of bounds in cuda::std::tuple_element<> (const cuda::std::integer_sequence)");
+  using type = _Tp;
+};
+
+template <size_t _Ip, class _Tp, _Tp... _Indices>
+[[nodiscard]] _CCCL_API constexpr _Tp get(integer_sequence<_Tp, _Indices...>) noexcept
+{
+  static_assert(_Ip < sizeof...(_Indices), "Index out of bounds in cuda::std::get<> (cuda::std::integer_sequence)");
+  constexpr _Tp __indices[]{_Indices...};
+  return __indices[_Ip];
+}
+
 _CCCL_END_NAMESPACE_CUDA_STD
+
+// tuple protocol for cuda::std::integer_sequence
+
+_CCCL_BEGIN_NAMESPACE_STD
+
+template <class _Tp, _Tp... _Indices>
+struct tuple_size<::cuda::std::integer_sequence<_Tp, _Indices...>>
+    : ::cuda::std::integral_constant<::cuda::std::size_t, sizeof...(_Indices)>
+{};
+
+template <::cuda::std::size_t _Ip, class _Tp, _Tp... _Indices>
+struct tuple_element<_Ip, ::cuda::std::integer_sequence<_Tp, _Indices...>>
+{
+  static_assert(_Ip < sizeof...(_Indices), "Index out of bounds in std::tuple_element<> (cuda::std::integer_sequence)");
+  using type = _Tp;
+};
+
+template <::cuda::std::size_t _Ip, class _Tp, _Tp... _Indices>
+struct tuple_element<_Ip, const ::cuda::std::integer_sequence<_Tp, _Indices...>>
+{
+  static_assert(_Ip < sizeof...(_Indices),
+                "Index out of bounds in std::tuple_element<> (const cuda::std::integer_sequence)");
+  using type = _Tp;
+};
+
+_CCCL_END_NAMESPACE_STD
 
 #include <cuda/std/__cccl/epilogue.h>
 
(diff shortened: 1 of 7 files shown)


cudax/test/CMakeLists.txt
```diff
@@ -104,6 +104,13 @@ cudax_add_catch2_test(test_target copy
 cudax_add_catch2_test(test_target cuco
     cuco/utility/test_hashers.cu
 )
+# nvcc 12.0 fails to compile 128-bit int cuco hashers test, so we disable the int128 support in CCCL
```
Jacobfaib: Surely just ifdefing out the offending tests as before is better no? Rather than disabling the entire feature wholesale.

libcudacxx/include/cuda/std/__utility/integer_sequence.h
```diff
@@ -242,8 +245,64 @@ using make_index_sequence = make_integer_sequence<size_t, _Np>;
 template <class... _Tp>
 using index_sequence_for = make_index_sequence<sizeof...(_Tp)>;
 
+// specialize cuda::std::tuple_size and cuda::std::tuple_element for cuda::std::integer_sequnence
```
Jacobfaib:
```suggestion
// specialize cuda::std::tuple_size and cuda::std::tuple_element for cuda::std::integer_sequence
```
(and elsewhere)

libcudacxx/include/cuda/std/__utility/integer_sequence.h
```diff
@@ -242,8 +245,64 @@ using make_index_sequence = make_integer_sequence<size_t, _Np>;
...
+template <size_t _Ip, class _Tp, _Tp... _Indices>
+struct tuple_element<_Ip, const integer_sequence<_Tp, _Indices...>>
+{
+  static_assert(_Ip < sizeof...(_Indices),
+                "Index out of bounds in cuda::std::tuple_element<> (const cuda::std::integer_sequence)");
+  using type = _Tp;
+};
+
+template <size_t _Ip, class _Tp, _Tp... _Indices>
+[[nodiscard]] _CCCL_HOST_DEVICE_API constexpr _Tp get(integer_sequence<_Tp, _Indices...>) noexcept
```
Jacobfaib: This should work with tile no?
```suggestion
[[nodiscard]] _CCCL_API constexpr _Tp get(integer_sequence<_Tp, _Indices...>) noexcept
```

libcudacxx/include/cuda/std/__utility/integer_sequence.h
```diff
@@ -242,8 +245,64 @@ using make_index_sequence = make_integer_sequence<size_t, _Np>;
...
+};
+
+template <::cuda::std::size_t _Ip, class _Tp, _Tp... _Indices>
+struct tuple_element<_Ip, const ::cuda::std::integer_sequence<_Tp, _Indices...>>
+{
+  static_assert(_Ip < sizeof...(_Indices),
+                "Index out of bounds in std::tuple_element<> (const cuda::std::integer_sequence)");
+  using type = _Tp;
+};
+
```
Jacobfaib, line 303: Why no specialization of `std::get`?

libcudacxx/test/libcudacxx/std/utilities/intseq/intseq.binding/tuple_interface.pass.cpp
```diff
@@ -0,0 +1,74 @@
...
+// template<size_t I, class T, T... Values>
+//   constexpr T get(integer_sequence<T, Values...>) noexcept;
+
+#include <cuda/std/cassert>
+#include <cuda/std/concepts>
+#include <cuda/std/tuple>
+#include <cuda/std/type_traits>
+#include <cuda/std/utility>
+
+__host__ __device__ constexpr bool test()
```
Jacobfaib: This should test the `std::` versions as well since we specialize them.

cudax/test/CMakeLists.txt
davebayer: This only touches nvcc 12.0. We've tried ifdefing out parts of the code and the bug still appears. Honestly, I don't want to spend time fixing this, when this is in cudax and will most likely never meet 12.0 in production

libcudacxx/include/cuda/std/__utility/integer_sequence.h
davebayer: Because it doesn't make sense, the get function must be implemented in the same namespace as the type it's implemented for. The structured bindings use ADL to find the correct implementation

libcudacxx/include/cuda/std/__utility/integer_sequence.h
```diff
@@ -242,8 +245,64 @@ using make_index_sequence = make_integer_sequence<size_t, _Np>;
...
+template <size_t _Ip, class _Tp, _Tp... _Indices>
+struct tuple_element<_Ip, const integer_sequence<_Tp, _Indices...>>
+{
+  static_assert(_Ip < sizeof...(_Indices),
+                "Index out of bounds in cuda::std::tuple_element<> (const cuda::std::integer_sequence)");
+  using type = _Tp;
+};
```
miscco, line 268: Why do we need to specialize the `const` case too?

I believe that should work through `tuple_element` and if it does not we would also need `volatile` and `const volatile` specializations

libcudacxx/include/cuda/std/__utility/integer_sequence.h
miscco: yeah no reason this would not work in tile

libcudacxx/test/libcudacxx/std/utilities/intseq/intseq.binding/tuple_interface.pass.cpp
miscco:
```suggestion
TEST_FUNC constexpr bool test()
```

libcudacxx/test/libcudacxx/std/utilities/intseq/intseq.binding/tuple_interface.pass.cpp
```diff
@@ -0,0 +1,74 @@
...
+  static_assert(cuda::std::is_same_v<cuda::std::tuple_element_t<3, Seq4>, int>);
+
+  static_assert(cuda::std::is_same_v<cuda::std::tuple_element_t<0, const Seq4>, int>);
+  static_assert(cuda::std::is_same_v<cuda::std::tuple_element_t<1, const Seq4>, int>);
+  static_assert(cuda::std::is_same_v<cuda::std::tuple_element_t<2, const Seq4>, int>);
+  static_assert(cuda::std::is_same_v<cuda::std::tuple_element_t<3, const Seq4>, int>);
+
+  // cuda::std::get
+  static_assert(cuda::std::is_same_v<decltype(cuda::std::get<0>(Seq4{})), int>);
+  static_assert(noexcept(cuda::std::get<0>(Seq4{})));
```
miscco: Please also test unqualified `get`

libcudacxx/test/libcudacxx/std/utilities/intseq/intseq.binding/tuple_interface.pass.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+#include <cuda/std/cassert>
+#include <cuda/std/concepts>
+#include <cuda/std/tuple>
+#include <cuda/std/type_traits>
+#include <cuda/std/utility>
+
+__host__ __device__ constexpr bool test()
+{
+  // cuda::std::tuple_size_v
```
miscco: Second also testing std types

libcudacxx/include/cuda/std/__utility/integer_sequence.h
Jacobfaib: Yeah but if I call qualified std::get it should work too, no?

libcudacxx/include/cuda/std/__utility/integer_sequence.h
davebayer: Yes, it's because the `type` is `const`-unqualified

libcudacxx/include/cuda/std/__utility/integer_sequence.h
davebayer: No, we specialize `std::tuple_(size|element)` types only to make structured bindings work. We can have a discussion to make `std::get` support our types in a separate thread/issue. But I don't think it makes too much sense

Jacobfaib (approved):

libcudacxx/include/cuda/std/__tuple_dir/tuple_like.h
```diff
@@ -83,6 +84,9 @@ inline constexpr bool __tuple_like_ext<::cuda::std::ranges::subrange<_Ip, _Sp, _
 template <class... _Tp>
 inline constexpr bool __tuple_like_ext<__tuple_types<_Tp...>> = true;
 
+template <class _Tp, _Tp... _Vs>
+inline constexpr bool __tuple_like_ext<integer_sequence<_Tp, _Vs...>> = true;
+
```
miscco: This needs to be dropped, or we would introduce a tuple constructor from `integer_sequence` that is AFAIK not intentional

miscco (changes requested):

miscco (approved):

PR #9130, opened 2026-05-26
[libcu++] Change `inplace_vector` return type for `try_meow` methods

davebayer:
This PR implements part of [P3981R2](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2026/p3981r2.html) that changes `inplace_vector` return types for `try_meow` methods.

I've not updated the feature testing macro value, because there is another paper that we don't currently implement and they share the same macro value.

This is a source breaking change, but it's consistent what other standard libraries do when a feature is changed before the standard release.

### DIFF
--- a/libcudacxx/include/cuda/std/inplace_vector
+++ b/libcudacxx/include/cuda/std/inplace_vector
@@ -44,6 +44,7 @@
 #include <cuda/std/__memory/uninitialized_algorithms.h>
 #include <cuda/std/__new/device_new.h>
 #include <cuda/std/__new/launder.h>
+#include <cuda/std/__optional/optional_ref.h>
 #include <cuda/std/__ranges/access.h>
 #include <cuda/std/__ranges/concepts.h>
 #include <cuda/std/__ranges/from_range.h>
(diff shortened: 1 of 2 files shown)


libcudacxx/include/cuda/std/inplace_vector
```diff
@@ -44,6 +44,7 @@
 #include <cuda/std/__memory/uninitialized_algorithms.h>
 #include <cuda/std/__new/device_new.h>
 #include <cuda/std/__new/launder.h>
+#include <cuda/std/__optional/optional_ref.h>
```
davebayer, line 47: @miscco should I only include this header or rather full `<cuda/std/optional>`?

libcudacxx/include/cuda/std/inplace_vector
```diff
@@ -1406,34 +1407,33 @@ public:
...
 
   template <class... _Args>
-  _CCCL_API constexpr pointer try_emplace_back(_Args&&... __args) noexcept(is_nothrow_constructible_v<_Tp, _Args...>)
+  _CCCL_API constexpr optional<reference>
+  try_emplace_back(_Args&&... __args) noexcept(is_nothrow_constructible_v<_Tp, _Args...>)
   {
     if (this->size() == _Capacity)
     {
-      return nullptr;
+      return nullopt;
```
miscco: This might break tile mode, because `nullopt` is a device variable and not a `tile` variable. Please replace by `{}`

libcudacxx/include/cuda/std/inplace_vector
miscco: I believe this should be fine if we are absolutely sure it will always only be a reference

miscco (approved):

PR #9136, opened 2026-05-26
[STF] Deduplicate cudaStreamIsCapturing helper

andralex:
## Summary

Follow-up on top of #8919 that came out of a post-merge review.

Three independent copies of the `cudaStreamIsCapturing(stream, &s); s != None` pattern had grown across cudax: two inline in `places.cuh` (`get_device_from_stream`, `get_stream_id`) and one as a private `stream_ctx::is_capturing` helper. They all do the same thing -- ask "is this stream currently part of a CUDA graph capture?" -- which is the *only* driver query that is itself capture-safe and so deserves a single named, documented home.

Introduce `cuda::experimental::places::is_stream_capturing(cudaStream_t)` in `__places/stream_pool.cuh`, document why it is the one query you can safely call mid-capture, and re-export it into the `stf` namespace via `stf_places_extended_exports.cuh`. Replace the three call sites. No behavior change; the new helper does **not** silently swallow the `cudaStreamIsCapturing` return value the way the old `cap_err == cudaSuccess` paths did, because that error is always fatal in practice and the existing `cuda_try` machinery already gives the right diagnostic.

### History

Originally this PR also wrapped `cudaFree(0)` in a process-wide gate to skip redundant runtime-init calls on the hot path. Dropped per @caugonnet's review: CUDA contexts can be torn down and recreated (some Python stacks do this), and a process-wide `static bool` would lie about that, potentially leaving subsequent contexts uninitialized. The micro-optimization isn't worth that risk.

## Test plan

- [ ] CI green -- pure refactor of the capture-query, exercised by the existing in-capture coverage (`legacy_to_stf_in_capture.cu` etc.).

### DIFF
--- a/cudax/include/cuda/experimental/__stf/stream/stream_ctx.cuh
+++ b/cudax/include/cuda/experimental/__stf/stream/stream_ctx.cuh
@@ -142,12 +142,15 @@ public:
       : backend_ctx<stream_ctx>(::std::make_shared<impl>(mv(handle)))
   {}
   stream_ctx(cudaStream_t user_stream, async_resources_handle handle = async_resources_handle(nullptr))
-      : backend_ctx<stream_ctx>(::std::make_shared<impl>(mv(handle), !is_capturing(user_stream)))
+      : backend_ctx<stream_ctx>(::std::make_shared<impl>(mv(handle), !is_stream_capturing(user_stream)))
   {
-    // A valid user stream means the CUDA runtime has already been initialized.
-    // If that stream is currently capturing, avoid making the backend
-    // constructor issue any additional runtime initialization calls that could
-    // be incompatible with the capture.
+    // If ``user_stream`` is currently capturing a CUDA graph, avoid issuing
+    // any CUDA runtime initialization calls in the backend constructor: under
+    // ``cudaStreamCaptureModeThreadLocal`` / ``Global`` such calls are
+    // rejected with ``cudaErrorStreamCaptureUnsupported`` and invalidate the
+    // in-progress capture. (The null/default stream is treated as
+    // ``not-capturing`` by ``cudaStreamIsCapturing``, so it goes through the
+    // normal init path.)
 
     // When the caller supplies their own ``async_resources_handle``, its
     // stream pool is very likely already populated with streams that carry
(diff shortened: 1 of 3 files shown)


cudax/include/cuda/experimental/__stf/internal/backend_ctx.cuh
```diff
@@ -125,11 +125,36 @@ protected:
...
+        // double-checked-lock on the steady-state path, which is exactly
+        // what we're trying to remove.
+        //
+        // The capture-safety gate is still upstream of this branch (callers
+        // pass ``initialize_cuda_runtime = false`` for capturing streams),
+        // because ``cudaFree(0)`` under
+        // ``cudaStreamCaptureModeThreadLocal`` / ``Global`` is rejected with
+        // ``cudaErrorStreamCaptureUnsupported`` *and* invalidates the
+        // in-flight capture.
+        static bool cuda_runtime_initialized = false;
```
caugonnet: I thought about that too, but this is replacing a very cheap operation we do all the time by some static state that might have bad consequences. We sometimes entirely destroy CUDA contexts (eg. this is what some python stacks do), I am unsure things would work if we don't "initialize" CUDA again but it's probably too defensive ...

cudax/include/cuda/experimental/__stf/stream/stream_ctx.cuh
```diff
@@ -142,12 +142,15 @@ public:
...
   {}
   stream_ctx(cudaStream_t user_stream, async_resources_handle handle = async_resources_handle(nullptr))
-      : backend_ctx<stream_ctx>(::std::make_shared<impl>(mv(handle), !is_capturing(user_stream)))
+      : backend_ctx<stream_ctx>(::std::make_shared<impl>(mv(handle), !is_stream_capturing(user_stream)))
   {
-    // A valid user stream means the CUDA runtime has already been initialized.
-    // If that stream is currently capturing, avoid making the backend
-    // constructor issue any additional runtime initialization calls that could
-    // be incompatible with the capture.
+    // If ``user_stream`` is currently capturing a CUDA graph, avoid issuing
```
caugonnet, line 147: I think the previous comment was more to the point, no need to initialize CUDA while capture because it is illegal and useless

caugonnet:
I think we can revert one comment modification, otherwise it's fine, cool to factorize code !

andralex:
@caugonnet good call -- I hadn't thought about Python stacks tearing down and recreating CUDA contexts. Dropped the static-bool gate (commit 77cfef883e); the PR is now just the is_stream_capturing dedupe. Updated the title and description accordingly.

caugonnet (approved):

cudax/include/cuda/experimental/__stf/stream/stream_ctx.cuh
andralex: OK will undo

PR #9137, opened 2026-05-27
`cudax::copy(mdspan)` Optimize shared memory cases

fbusato:
## Description

The PR optimizes `cudax::copy(mdspan)` for the following cases:

- Extend permutation of the destination tensor to exploit coalesced memory accesses on a wider set of cases. Previously, the shared memory path was enabled only when the destination tensor is contiguous.
- Add Xor swizzle for 2D cases.
- Simplified `fast_mod_div` code to use less registers.

### DIFF
--- a/cudax/benchmarks/bench/copy/copy_bench.cu
+++ b/cudax/benchmarks/bench/copy/copy_bench.cu
@@ -9,6 +9,8 @@
 
 #include <cuda/experimental/copy.cuh>
 
+#include <cstddef>
+#include <cstdint>
 #include <cstdlib>
 
 #include <nvbench/nvbench.cuh>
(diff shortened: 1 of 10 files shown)


cudax/include/cuda/experimental/__copy/copy_shared_memory_utils.cuh
```diff
@@ -89,107 +108,162 @@ __num_contiguous_dimensions(const __raw_tensor<_ExtentT, _StrideT, _Tp, _MaxRank
...
+    const auto __stride = ::cuda::experimental::__abs_integer(__tensor.__strides[__perm_i]);
+    if (__stride != __expected_stride) // input tensor not contiguous
+    {
+      break;
+    }
+
+    if (__result.__tile_sizes[__perm_i] == 1) // first time we see this dimension
+    {
+      const auto __tile_size             = ::cuda::std::min(__extent, __max_tile_size);
+      const auto __tile_total_size_bytes = __result.__tile_total_size * __tile_size * sizeof(_Tp);
```
NaderAlAwar: critical: this can over or under estimate shared memory requirements because `sizeof(_Tp)` depends on which tensor is passed to this helper. The shared memory buffer is always allocated using the source dtype, so this check should use the source/shared memory dtype for both source and destination runs. We should add a mixed dtype test that covers the new shared memory tiled path

cudax/benchmarks/bench/copy/copy_bench.cu
```diff
@@ -75,189 +76,256 @@ void bench_copy(nvbench::state& state,
...
 /***********************************************************************************************************************
- * Negative strides benchmarks
+ * Transpose benchmark
  **********************************************************************************************************************/
 
-// src: (70,90,80,80):(-576000,-6400,-80,-1), offset=alloc-1
-// dst: (70,90,80,80):(-576000,-6400,-80,-1), offset=alloc-1
-void negative_strides_0(nvbench::state& state)
+// src: (8192,8192):(8192,1)
+// dst: (8192,8192):(1,8192)
```
NaderAlAwar: Suggestion: this comment is stale, and shuold be updated to reflect the actual shapes + strides (32768 instead of 8192)

cudax/benchmarks/bench/copy/copy_bench.cu
```diff
@@ -75,189 +76,256 @@ void bench_copy(nvbench::state& state,
...
 /***********************************************************************************************************************
- * Negative strides benchmarks
+ * Transpose benchmark
  **********************************************************************************************************************/
 
-// src: (70,90,80,80):(-576000,-6400,-80,-1), offset=alloc-1
-// dst: (70,90,80,80):(-576000,-6400,-80,-1), offset=alloc-1
-void negative_strides_0(nvbench::state& state)
+// src: (32768,32768):(32768,1)
+// dst: (32768,32768):(1,32768)
```
NaderAlAwar: Nit: the src and dst strides are actually reversed according to the code below

NaderAlAwar (approved):

cudax/include/cuda/experimental/__copy/copy_shared_memory.cuh
```diff
@@ -54,18 +52,38 @@
...
+{
+  if constexpr (_UseXorSwizzle)
+  {
+    static_assert(__max_tile_size == 32, "XOR shared-memory swizzle assumes 32 banks and 32-element tile modes");
+    constexpr __tile_extent_t __swizzle_tile_size = __max_tile_size * __max_tile_size;
+    const auto __outer                            = __offset / __swizzle_tile_size;
+    const auto __inner                            = __offset - (__outer * __swizzle_tile_size);
+    const auto __row                              = __inner / __max_tile_size;
+    const auto __col                              = __inner - (__row * __max_tile_size);
+    return __outer * __swizzle_tile_size + __row * __max_tile_size + (__col ^ __row);
```
oleksandr-pavlyk: We compute `__outer * __swizzle_tile_size` twice. Same for `__row * __max_tile_size`. 

Question: Does saving these in `const auto` variable have any downsides?

fbusato: the compiler should be able to merge them because the pattern is very simple. Anyway, better to consolidate them

srinivasyadav18 (approved):

PR #9138, opened 2026-05-27
[STF] [TRIVIAL] Rename decorated_stream to augmented_stream

andralex:
## Summary

Pure rename: ``decorated_stream`` → ``augmented_stream`` across cudax/STF (12 files, 58 type uses + the ``get_decorated_stream`` accessor).

The struct bundles a ``cudaStream_t`` with its pre-resolved driver identity (``cuStreamGetId``) and its home device, so downstream STF -- the stream-reuse logic in ``stream_task``, the ``(src_id, dst_id)``-keyed sync-skip cache in ``async_resources_handle``, the stream event helpers -- can use it as a value type and as a cache key without re-paying the driver round-trip every time.

"Decorated" is technically accurate (decorator pattern) but uninformative: it doesn't tell the reader *what* the decoration is for. "Augmented" -- same family as "augmented data" / "augmented reality" -- makes the central fact explicit: this is a stream that has been *augmented* with extra, pre-computed information so it can stand in for the bare ``cudaStream_t`` in places where the bare handle alone would not be enough. The doc comment on the struct now spells this out.

### Scope

- Type: ``decorated_stream`` → ``augmented_stream``
- Accessor: ``get_decorated_stream`` → ``get_augmented_stream``
- One local variable (``decorated_s`` in ``slice.cuh``) renamed because it explicitly mirrored the type name
- Doc-comment mentions of "decorated stream" updated to match
- The ``dstream`` local-variable abbreviation used throughout STF is **left untouched on purpose**: it's a project-wide convention orthogonal to the type name, and renaming it would balloon the diff for no clarity gain

### Test plan

- [ ] CI green -- pure rename, no behavior change; if it builds, it works

### DIFF
--- a/cudax/include/cuda/experimental/__places/exec/cuda_stream.cuh
+++ b/cudax/include/cuda/experimental/__places/exec/cuda_stream.cuh
@@ -35,7 +35,7 @@ namespace cuda::experimental::places
 class exec_place_cuda_stream_impl : public exec_place::impl
 {
 public:
-  exec_place_cuda_stream_impl(const decorated_stream& dstream)
+  exec_place_cuda_stream_impl(const augmented_stream& dstream)
       : exec_place::impl(data_place::device(dstream.dev_id))
       , dstream_(dstream)
       , dummy_pool_(dstream)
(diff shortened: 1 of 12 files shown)


caugonnet:
Why not changing dstream variables to astream for example ?

bernhardmgruber (approved):

PR #8190, opened 2026-03-26
[STF] Move unstable_unique from STF to generic cudax utility

caugonnet:
unstable_unique is a generic algorithm with no STF dependency. Move it to cuda::experimental namespace under __utility/, replace the STF UNITTEST with a Catch2 test, and update the sole STF consumer (event_types.cuh) to use the new location directly.

## Description

closes 

## Checklist

- [x] New or existing tests cover these changes.
- [ ] The documentation is up to date with these changes.

PR #9139, opened 2026-05-27
[STF] Don't terminate test binary when a UNITTEST source path is missing

andralex:
## Summary

The header-unittest dispatcher in ``unittest.cuh`` decides whether to run each registered UNITTEST by comparing the call site's ``std::source_location::current()`` with the test binary's ``UNITTESTED_FILE`` via:

```cpp
if (!::std::filesystem::equivalent(loc.file_name(), UNITTESTED_FILE)) { ... }
```

``loc.file_name()`` is the **source path the compiler saw**, baked into the binary as a string literal. If the underlying header is later moved or renamed, that recorded path may no longer exist on disk -- a perfectly normal situation for any binary that wasn't rebuilt after the move.

The throwing overload of ``std::filesystem::equivalent`` raises ``filesystem_error`` ("No such file or directory") in that case. The try/catch around the user lambda below the check is *inside* the same if-branch and so cannot catch it; the exception propagates out of the dispatcher and the entire test binary calls ``std::terminate`` -- masking every other unittest in the same binary and producing a core dump.

### Concrete repro

This is exactly what happens today on a tree that still has a build directory predating PR #8190 (which moved ``unstable_unique.cuh`` out of ``__stf/utility/``):

```text
$ ./build/cudax/bin/cudax.test.stf.unittest_headers.__stf.internal.context
terminate called after throwing an instance of 'std::filesystem::__cxx11::filesystem_error'
  what():  filesystem error: cannot check file equivalence: No such file or directory
       [/.../__stf/utility/unstable_unique.cuh] [/.../__stf/internal/context.cuh]
Aborted (core dumped)
```

The old ``__stf/utility/unstable_unique.cuh`` contained a ``UNITTEST(...)`` block, so the baked-in source location survives in any ``.o`` built before the move. After the move, the path doesn't exist, ``equivalent`` throws, and the binary dies before running anything else.

## Fix

Switch to the non-throwing overload (the one that takes an ``std::error_code``) and treat any error as "not equivalent" -- i.e. skip just that unittest, run the rest:

```cpp
::std::error_code __equiv_ec;
if (!::std::filesystem::equivalent(loc.file_name(), UNITTESTED_FILE, __equiv_ec)) { ... }
```

Identical semantics on the happy path. No more ``std::terminate`` from a stale binary.

## Test plan

- [ ] CI green on the existing unittest_headers targets (everything that worked before still works).
- [ ] Manual: a binary built against an older tree can now be re-run against a checkout where headers have moved without crashing; it correctly skips the moved UNITTEST and runs the rest. (Not added as an automated test because it would require artificially producing a stale binary.)

### DIFF
--- a/cudax/include/cuda/experimental/__stf/utility/unittest.cuh
+++ b/cudax/include/cuda/experimental/__stf/utility/unittest.cuh
@@ -31,6 +31,7 @@
 #include <cuda/experimental/__stf/utility/traits.cuh>
 
 #include <filesystem>
+#include <set>
 
 #ifndef _CCCL_DOXYGEN_INVOKED // Do not document
 // One level of macro indirection is required in order to resolve __COUNTER__,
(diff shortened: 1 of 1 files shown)


andralex:
Closing -- this turned out to be a stale-binary issue, not a framework bug.

The crash that motivated this PR (``std::terminate`` from ``std::filesystem::equivalent`` in the UNITTEST dispatcher) reproduced only on test binaries built **before** ``459e81a530`` ([STF] Move unstable_unique from STF to generic cudax utility, #8190). That commit deleted ``cudax/include/cuda/experimental/__stf/utility/unstable_unique.cuh``, but the old header contained a ``UNITTEST(...)`` block whose ``loc.file_name()`` path was baked into every ``.o`` that transitively included it. After the move, the path no longer exists on disk, so the throwing overload of ``std::filesystem::equivalent`` raises ``filesystem_error``; the dispatcher's ``try``/``catch`` doesn't cover that throw, the exception escapes, and the binary aborts.

Verified empirically:

| binary                                                  | mtime           | occurrences of dead path |
| ------------------------------------------------------- | --------------- | ------------------------ |
| ``build/cudax/bin/...__stf.internal.context``           | Apr 1  21:57    | 1                        |
| ``build/cudax-cpp20/bin/...__stf.internal.context``     | May 26 23:13    | 0                        |

Rebuilding (or wiping the stale build tree) clears the symptom permanently. No source code in the current tree produces the dead path.

The proposed change (non-throwing ``equivalent`` + dedup stderr diagnostic) would gracefully handle this for any future header move/rename, but that's pure defensive code for a self-inflicted "old binary against new tree" situation. Not worth the extra surface area on the dispatcher. If this pattern ever becomes a recurring footgun for contributors, the patch is two-stanza small and easy to revive from the branch.

PR #9144, opened 2026-05-27
[cub] Harden histogram benchmarks and add bin-by-bin verification

robobryce:
## Summary

Hardens the four `cub.bench.histogram.{even,range,multi.even,multi.range}.base` NVBench benchmarks (under `cub/benchmarks/bench/histogram/`) against pattern-matching tunings and silent dispatch failures, and adds an opt-out bin-by-bin correctness verifier that runs outside the timed region. Also extends and adds CTest coverage in `cub/test/`.

The four benchmarks each call into `cub::DeviceHistogram::HistogramEven` / `HistogramRange` (single-channel) or `MultiHistogramEven` / `MultiHistogramRange` (3-active-channel of 4) over a sweep of input sizes, bin counts, sample types, and entropy values, and report a bandwidth in bytes/sec via NVBench's `GlobalMem BW` column.

## Why

When you tune CUB's histogram code paths, NVBench measures elapsed time inside `state.exec` and reports bandwidth as `bytes_processed / elapsed`. Two failure modes can quietly inflate the reported bandwidth:

1. **The dispatch returns a non-`cudaSuccess` error code**, but the benchmark drops the return value on the floor. The kernel never launched, so the elapsed time is small, so the reported bandwidth is large.
2. **The kernel completes with `cudaSuccess` but writes wrong counts** into the output histogram. For example, a partition mask that drops samples that should have landed in another partition's write set produces a non-empty histogram with the right shape but the wrong values. CTest cases that aggregate to a sum-of-counts check pass; the bench reports inflated bandwidth because effectively only a fraction of the input was processed.

There is also a third class — uniform-spacing detection fast paths and tunings that key off power-of-two element/bin counts — where the bench measures a code path real users do not exercise, so the reported bandwidth overstates production behavior.

### Six commits, each scoped narrowly

1. **`[cub] Harden histogram benchmarks against pattern-matching tunings`** — single-channel `even.cu` / `range.cu`. Quadratic-spaced range levels (still strictly monotonic across `[lower_level, upper_level]`) so `DispatchRange` stays on the `SearchTransform` path instead of collapsing to the uniform-detection fast path. Replaces the power-of-two `Elements{io}` axis with non-power-of-two sizes. Switches to NVBench's manual-timer `exec_tag` and calls `cudaCtxResetPersistingL2Cache()` outside the timed window, since cold-cache eviction does not demote persistence-marked addresses set via `cudaStreamSetAttribute` / `cudaAccessPolicyWindow`.

2. **`[cub] Extend histogram bench/test hardening to multi-channel and add cache coverage`** — same three changes for `multi/even.cu` and `multi/range.cu`. Replaces 128/2048 with 100/2000 in the `Bins` axis on all four benches so tunings that hard-code on power-of-two bin counts cannot use those shortcuts. In `cub/test/catch2_test_device_histogram.cu`, `setup_bin_levels_for_range` now perturbs interior levels by ±`min_bin_width/4` (alternating sign), falling back to uniform when the type is too tight (e.g. byte-sample with 256 levels). Adds a new `cub/test/catch2_test_device_histogram_thread_local_cache.cu` with three Catch2 cases targeting the `thread_local` detection-stream / detection-buffer cache in `dispatch_range`: sequential calls across multiple user streams, four-thread concurrent calls on the same device, and a single-thread cross-device case that skips when fewer than two GPUs are present.

3. **`[cub] Force histogram bench persisting-L2 reservation to zero each iteration`** — adds `cudaDeviceSetLimit(cudaLimitPersistingL2CacheSize, 0)` immediately before every timed iteration, so a prior bench in the same NVBench process can't leave the reservation bumped. The default is 0; this hardcodes 0 instead of relying on the default.

4. **`[cub] Use jittered-uniform histogram bench levels instead of quadratic`** — quadratic spacing produced bin widths spanning ~2n× (last bin vs first), which is not representative of typical workloads. Jittered uniform spacing (±25% of step, fixed `mt19937` seed for reproducibility) keeps consecutive widths within ~3× while still defeating the uniform-spacing detection fast path.

5. **`[cub] Add 16384 and 60000 bin sizes to histogram benchmarks`** — covers the 10k–65k bin tier between 2000 and 2097152 across all four benches.

6. **`[cub] Verify histogram benchmark output bin-by-bin against a reference`** — the verifier described below.

### How the verifier works

Each benchmark cell builds the strided pixel-major sample buffer and the output histograms exactly as before. Then, before `state.exec`:

1. The dispatch (`HistogramEven` / `HistogramRange` / `MultiHistogramEven` / `MultiHistogramRange`) is called once with explicit return-code checking. A non-`cudaSuccess` return throws `std::runtime_error` and the cell is reported as failed; NVBench prints the message in the per-cell log and emits no bandwidth row for that cell.

2. The produced histogram(s) are compared bin-by-bin against an independent reference. The reference is computed on the same device, but with `thrust::for_each` + a global `atomicAdd` per pixel per active channel — the EVEN reference uses the closed-form bin index `(sample - lo) * num_bins / (hi - lo)` (in double); the RANGE reference uses `cub::UpperBound` on the per-channel level array. Neither reference shares any kernel with `cub::DeviceHistogram`, so a bug that produces a wrong-but-plausible histogram from the optimized path is caught.

3. The compare reports both the total count (`opt_total` vs `ref_total`) and the first mismatched bin index with both values, so a failure log identifies the channel and bin to investigate.

The warmup, the reference build, and the comparison all run outside `state.exec`, so the timed region only contains the dispatch under test.

### Disabling the verifier at run time

The verifier is on by default. To skip it (e.g. during a tuning sweep where only relative throughput matters and the verifier's wall-clock cost is undesirable), set `CUB_BENCH_HISTOGRAM_VERIFY` to `0`, `false`, `no`, or `off` (case-insensitive) in the environment when invoking the benchmark binary. The flag is read once per process via a function-local static, so toggling has no per-cell overhead.

## Cost

The verifier's cost on the timed region (NVBench's reported `histogram` metric — geometric mean across all four binaries' `GlobalMem BW (bytes/sec)` rows, in `GiB/s`) is zero within measurement noise. Wall-clock benchmark runtime grows by ~14% (median ~7 s on a ~52 s baseline run) because the on-device reference loops over every sample once per cell.

Hardware: NVIDIA B200, CUDA driver 580.126.09, CUDA toolkit 12.8.93, host Linux 6.8 with 30 CPUs / 180 GiB RAM. Five interleaved runs of `cub/benchmarks/bench/histogram/run_histogram_benchmark.py` each (interleaved to factor out NVBench-iteration-count run-to-run variance):

| | Mean `histogram` (GiB/s) | Median wall-clock (s) |
|---|---|---|
| baseline (no verifier) | 210.87 | 52.0 |
| verifier on (default) | 210.88 | 59.4 |
| verifier off (`CUB_BENCH_HISTOGRAM_VERIFY=0`) | 210.78 | 52.8 |

The timed-region metric is unchanged across all three modes within ±0.05% (well inside the bench's natural noise of ±0.05–0.10%). The wall-clock overhead from running the verifier is ~7 s; with the verifier disabled, wall-clock is statistically indistinguishable from baseline.

## Validation

Built and ran the full sweep on the B200 host. With the verifier on the unmodified branch, all 160 benchmark cells (40 cells × 4 binaries) pass; the reported `histogram` metric is unchanged within noise.

To confirm the verifier triggers when the optimized output is actually wrong, I rebased a known-bad change onto this branch — a CUB optimization attempt that introduced a partition-mask bug that silently dropped most input samples on multi-channel `Bins=60000` configurations. With the verifier in place, that change is rejected: 44 cells across the four binaries fail, with messages like

```
Run:  [1/1] base [Device=0 SampleT{ct}=I32 CounterT{ct}=I32 OffsetT{ct}=I32 Elements{io}=268435456 Bins=60000 Entropy=0.201]
Fail: Unexpected error: bench correctness check [multi.range]: channel=0 total opt=1209247 ref=268435456;
      first mismatched bin=0 (opt=763903 ref=169668305)
```

The message reports both the total-count mismatch (`1.2M` vs `268M` — only ~0.45% of samples landed in the optimized histogram) and the first mismatched bin's per-bin counts. Without the verifier (`CUB_BENCH_HISTOGRAM_VERIFY=0`), the same change reports `histogram=893.25` GiB/s — a ~4.2× bandwidth speedup that is entirely an artifact of the dropped samples — exactly the failure mode this PR is intended to prevent.

## Test plan

- [x] `ci/util/build_and_test_targets.sh --preset cub-cpp20 --build-targets "cub.bench.histogram.even.base cub.bench.histogram.range.base cub.bench.histogram.multi.even.base cub.bench.histogram.multi.range.base"` builds clean.
- [x] `ctest -R 'cub\.test\.device\.histogram(\.lid_0|_api\.lid_0|_env\.lid_0|_thread_local_cache\.lid_0)$'` passes (including the new `cub.test.device.histogram_thread_local_cache.lid_0`).
- [x] All 160 benchmark cells (40 cells × 4 binaries) pass the bin-by-bin verifier on the unmodified branch; reported `histogram` metric is unchanged within noise.
- [x] `CUB_BENCH_HISTOGRAM_VERIFY=0` skips the verifier; reported `histogram` metric is unchanged within noise; wall-clock is statistically indistinguishable from baseline.
- [x] Cherry-picking a known-bad change onto this branch causes 44 cells to fail with explicit bin-by-bin mismatch messages when the verifier is on, and "passes" with an inflated `histogram` value when the verifier is off (demonstrating the failure mode the verifier exists to prevent).

### DIFF
--- a/cub/benchmarks/bench/histogram/even.cu
+++ b/cub/benchmarks/bench/histogram/even.cu
@@ -46,10 +46,14 @@ static void even(nvbench::state& state, nvbench::type_list<SampleT, CounterT, Of
   state.add_global_memory_writes<CounterT>(num_bins);
 
   caching_allocator_t alloc;
-  // Demote persisting L2 lines outside the timed window so that no
-  // cudaAccessPolicyWindow set by the dispatcher can carry across iterations.
+  // Force the persisting-L2 reservation back to 0 and demote any persisting
+  // lines outside the timed window, so neither cudaAccessPolicyWindow nor a
+  // bumped cudaLimitPersistingL2CacheSize can carry across iterations. The
+  // default reservation is 0; hardcoding 0 also clears any pollution left by
+  // a prior benchmark in the same nvbench process.
   state.exec(nvbench::exec_tag::gpu | nvbench::exec_tag::no_batch | nvbench::exec_tag::timer,
              [&](nvbench::launch& launch, auto& timer) {
+               cudaDeviceSetLimit(cudaLimitPersistingL2CacheSize, 0);
                cudaCtxResetPersistingL2Cache();
                timer.start();
                auto env = cub_bench_env(
--- a/cub/benchmarks/bench/histogram/even.cu
+++ b/cub/benchmarks/bench/histogram/even.cu
@@ -46,26 +46,32 @@ static void even(nvbench::state& state, nvbench::type_list<SampleT, CounterT, Of
   state.add_global_memory_writes<CounterT>(num_bins);
 
   caching_allocator_t alloc;
-  state.exec(nvbench::exec_tag::gpu | nvbench::exec_tag::no_batch, [&](nvbench::launch& launch) {
-    auto env = cub_bench_env(
-      alloc,
-      launch
+  // Demote persisting L2 lines outside the timed window so that no
+  // cudaAccessPolicyWindow set by the dispatcher can carry across iterations.
+  state.exec(nvbench::exec_tag::gpu | nvbench::exec_tag::no_batch | nvbench::exec_tag::timer,
+             [&](nvbench::launch& launch, auto& timer) {
+               cudaCtxResetPersistingL2Cache();
+               timer.start();
+               auto env = cub_bench_env(
+                 alloc,
+                 launch
 #if !TUNE_BASE
-      ,
-      cuda::execution::tune(bench_policy_selector<key_t, 1, 1>{})
+                 ,
+                 cuda::execution::tune(bench_policy_selector<key_t, 1, 1>{})
 #endif // !TUNE_BASE
-    );
-    _CCCL_TRY_CUDA_API(
-      cub::DeviceHistogram::HistogramEven,
-      "HistogramEven failed",
-      d_input,
-      d_histogram,
-      num_levels,
-      lower_level,
-      upper_level,
-      static_cast<OffsetT>(elements),
-      env);
-  });
+               );
+               _CCCL_TRY_CUDA_API(
+                 cub::DeviceHistogram::HistogramEven,
+                 "HistogramEven failed",
+                 d_input,
+                 d_histogram,
+                 num_levels,
+                 lower_level,
+                 upper_level,
+                 static_cast<OffsetT>(elements),
+                 env);
+               timer.stop();
+             });
 }
 
 using counter_types     = nvbench::type_list<int32_t>;
@@ -80,6 +86,6 @@ using sample_types = nvbench::type_list<int8_t, int16_t, int32_t, int64_t, float
 NVBENCH_BENCH_TYPES(even, NVBENCH_TYPE_AXES(sample_types, counter_types, some_offset_types))
   .set_name("base")
   .set_type_axes_names({"SampleT{ct}", "CounterT{ct}", "OffsetT{ct}"})
-  .add_int64_power_of_two_axis("Elements{io}", nvbench::range(16, 28, 4))
+  .add_int64_axis("Elements{io}", {100'000, 1 << 20, 20'000'000, 1 << 28})
   .add_int64_axis("Bins", {32, 128, 2048, 2097152})
   .add_string_axis("Entropy", {"0.201", "1.000"});
--- a/cub/benchmarks/bench/histogram/even.cu
+++ b/cub/benchmarks/bench/histogram/even.cu
@@ -22,17 +22,18 @@ static void even(nvbench::state& state, nvbench::type_list<SampleT, CounterT, Of
   const auto num_bins  = state.get_int64("Bins");
   const int num_levels = static_cast<int>(num_bins) + 1;
 
-  // Skip invalid configurations where LevelT (= SampleT) cannot represent the number of bins
-  if constexpr (cuda::std::is_integral_v<SampleT>)
+  // Skip invalid configurations where the SampleT range can't hold enough
+  // strictly-monotonic levels: bins + 1 levels require bins + 1 distinct
+  // SampleT values, and the bench's `[get_lower_level, get_upper_level]`
+  // range spans at most `max_representable_bins<SampleT>() + 1` distinct
+  // values.
+  if (num_bins > max_representable_bins<SampleT>())
   {
-    if (num_bins > static_cast<int64_t>(cuda::std::numeric_limits<SampleT>::max()))
-    {
-      state.skip("Number of bins exceeds what LevelT (= SampleT) can represent");
-      return;
-    }
+    state.skip("Number of bins exceeds what SampleT can represent");
+    return;
   }
 
-  const SampleT lower_level = 0;
+  const SampleT lower_level = get_lower_level<SampleT>();
   const SampleT upper_level = get_upper_level<SampleT>(num_bins, elements);
 
   thrust::device_vector<SampleT> input = generate(elements, entropy, lower_level, upper_level);
--- a/cub/benchmarks/bench/histogram/even.cu
+++ b/cub/benchmarks/bench/histogram/even.cu
@@ -87,5 +87,5 @@ NVBENCH_BENCH_TYPES(even, NVBENCH_TYPE_AXES(sample_types, counter_types, some_of
   .set_name("base")
   .set_type_axes_names({"SampleT{ct}", "CounterT{ct}", "OffsetT{ct}"})
   .add_int64_axis("Elements{io}", {100'000, 1 << 20, 20'000'000, 1 << 28})
-  .add_int64_axis("Bins", {32, 128, 2048, 2097152})
+  .add_int64_axis("Bins", {32, 100, 2000, 2097152})
   .add_string_axis("Entropy", {"0.201", "1.000"});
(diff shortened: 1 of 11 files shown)


cub/benchmarks/bench/histogram/range.cu
```diff
@@ -44,26 +66,80 @@ static void range(nvbench::state& state, nvbench::type_list<SampleT, CounterT, O
...
+
+  // Force the persisting-L2 reservation back to 0 and demote any persisting
+  // lines outside the timed window, so neither cudaAccessPolicyWindow nor a
+  // bumped cudaLimitPersistingL2CacheSize can carry across iterations. The
+  // default reservation is 0; hardcoding 0 also clears any pollution left by
+  // a prior benchmark in the same nvbench process.
+  state.exec(nvbench::exec_tag::gpu | nvbench::exec_tag::no_batch | nvbench::exec_tag::timer,
+             [&](nvbench::launch& launch, auto& timer) {
+               cudaDeviceSetLimit(cudaLimitPersistingL2CacheSize, 0);
+               cudaCtxResetPersistingL2Cache();
```
jrhemstad, line 137: This shouldn't be repeated in every benchmark. This is the kind of thing that should be in nvbench.

cub/test/catch2_test_device_histogram_thread_local_cache.cu
```diff
@@ -0,0 +1,261 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+// Exercises the thread_local detection-stream / detection-buffer cache used by
+// dispatch_range's uniform-levels detection path: sequential calls on multiple
+// user streams, concurrent calls from multiple threads, and the cross-device
+// hazard where the cache is bound to the device current at first call.
```
jrhemstad, line 7: These tests feel really ad hoc and not clear these are needed.

cub/benchmarks/bench/histogram/even.cu
```diff
@@ -45,27 +45,87 @@ static void even(nvbench::state& state, nvbench::type_list<SampleT, CounterT, Of
   state.add_global_memory_reads<SampleT>(elements);
   state.add_global_memory_writes<CounterT>(num_bins);
 
+  // Warmup + correctness check: run HistogramEven once outside `state.exec`,
```
jrhemstad, line 51: nvbench already does warmups. Manual warmups should not be added.

brycelelbach: This is not a warmup. It's a correctness check to validate that the histogram is actually correctly executed.

The unit tests only test with a problem size of up to 1024, and doesn't provide sufficient code coverage.

cub/test/catch2_test_device_histogram_thread_local_cache.cu
brycelelbach: Without these checks, an AI agent optimizing this code will attempt to cache the CUDA streams and other state in a thread local.

cub/benchmarks/bench/histogram/range.cu
brycelelbach: Sure, but it's not valid for every NVBench test, some may not want the reset. I'd be happy to add it upstream.

Without this agents will cheat with L2 persistence on the benchmark.

cub/benchmarks/bench/histogram/multi/even.cu
```diff
@@ -58,27 +58,94 @@ static void even(nvbench::state& state, nvbench::type_list<SampleT, CounterT, Of
   state.add_global_memory_reads<SampleT>(elements * num_active_channels);
   state.add_global_memory_writes<CounterT>(num_bins * num_active_channels);
 
+  // Warmup + correctness check: run MultiHistogramEven once outside
+  // `state.exec`, checking the dispatch return code, then verify each
+  // channel's histogram bin-by-bin against an independent reference.
+  // Skipped when CUB_BENCH_HISTOGRAM_VERIFY=0|false|no|off.
+  if (bench_correctness_checks_enabled())
```
miscco, line 78: Question: This is also costly to compile, should this be an `if constexpr`?

cub/benchmarks/bench/histogram/multi/even.cu
```diff
@@ -58,27 +58,94 @@ static void even(nvbench::state& state, nvbench::type_list<SampleT, CounterT, Of
...
+  // Force the persisting-L2 reservation back to 0 and demote any persisting
+  // lines outside the timed window, so neither cudaAccessPolicyWindow nor a
+  // bumped cudaLimitPersistingL2CacheSize can carry across iterations. The
+  // default reservation is 0; hardcoding 0 also clears any pollution left by
+  // a prior benchmark in the same nvbench process.
+  state.exec(nvbench::exec_tag::gpu | nvbench::exec_tag::no_batch | nvbench::exec_tag::timer,
+             [&](nvbench::launch& launch, auto& timer) {
+               cudaDeviceSetLimit(cudaLimitPersistingL2CacheSize, 0);
+               cudaCtxResetPersistingL2Cache();
+               timer.start();
```
miscco, line 141: We should have a helper function for that to use it in other / all benchmarks

brycelelbach:
Superseded by #10555

PR #9148, opened 2026-05-27
[places] Align cyclic_shape::size with iteration cardinality

fallintoplace:
## What changed
- make `cyclic_shape::size()` use ceil division so it matches the iterator cardinality for strided ranges
- add a regression test for the `[0, 7)` with stride `2` case

## Why
`cyclic_shape::size()` was using floor division while `cyclic_shape::end()` computed the past-the-end index with ceil division.

That breaks a basic invariant for strided tails. For `[0, 7)` with stride `2`, iteration yields `{0, 2, 4, 6}` (4 elements) but `size()` reported `3`.

## Impact
This makes `size()` agree with the number of elements produced by iteration for cyclic shapes with stragglers.

## Validation
- added a focused header unittest for the odd-tail case
- attempted `cmake --preset cudax-cpp17`, but validation is blocked in this environment because `nvcc` / `CUDAToolkit` is not available


miscco (approved):

miscco:
/ok to test [eb407f9](https://github.com/NVIDIA/cccl/pull/9148/commits/eb407f9b8ee5e3d701c6db9079e88cfeef77ad18)

fallintoplace:
Hi! Just checking in on this PR. It looks like the required checks are passing and the changes were approved, but there’s still one optional RAPIDS check failing. Is that failure considered blocking here, or is there anything I should update on my side before this can move forward? @miscco

caugonnet:
@fallintoplace It looks like we forgot to hit the enable auto-merge after test passed ... I'm passing CI now and will merge then, thanks for your fix and for the reminder !

fallintoplace:
@caugonnet Thank you. Have a nice Sunday.

PR #9149, opened 2026-05-27
[places] Guard empty tiled mdspan shape size

fallintoplace:
## What changed
- return `0` early from `tiled_mdspan_shape::size()` when the original shape is empty
- add a regression test covering the empty-shape case

## Why
`tiled_mdspan_shape::size()` computes `tile_per_part` and then uses `(tile_per_part - 1)` in unsigned arithmetic.

For empty input, `original_shape.size()` is `0`, so `ntiles == 0` and `tile_per_part == 0`. That makes `(tile_per_part - 1)` underflow and can turn an empty shape into a huge nonzero size.

## Impact
Empty tiled mdspan shapes now report `0` elements instead of underflowing.

## Validation
- added a focused header unittest for the empty-shape case
- attempted `cmake --preset cudax-cpp17`, but validation is blocked in this environment because `nvcc` / `CUDAToolkit` is not available

### DIFF
--- a/cudax/include/cuda/experimental/__places/partitions/tiled_partition.cuh
+++ b/cudax/include/cuda/experimental/__places/partitions/tiled_partition.cuh
@@ -51,6 +51,10 @@ public:
     _CCCL_ASSERT(mdspan_shape_t::rank() == 1, "Tiled mdspan shape only implemented in 1D yet");
 
     const size_t n = original_shape.size();
+    if (n == 0)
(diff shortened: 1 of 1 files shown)


miscco:
/ok to test [1e2d0c6](https://github.com/NVIDIA/cccl/pull/9149/commits/1e2d0c6a7f0aec704b637b6290fa323f406cfee3)

PR #9150, opened 2026-05-28
[STF] Migrate __stf/utility/ from cuda_safe_call to cuda_try

andralex:
## Summary

Second PR in the series migrating production STF headers off `cuda_safe_call` (abort-on-failure) and onto `cuda_try` (throw-on-failure), so callers (Python wrappers, exception-aware control flow) can recover from CUDA errors instead of having the process aborted.

Twelve call sites across three headers in `cudax/include/cuda/experimental/__stf/utility/`. After this PR, zero `cuda_safe_call` calls remain in that subtree outside of `cuda_safe_call.cuh` itself (the implementation).

Every pattern here was already demonstrated by PR #9147, so reviewer overhead per site should be small.

### `occupancy.cuh`

- **`compute_occupancy`** — hoisted the cache write to *after* the CUDA call. Before, `auto& result = occupancy_cache[key]` inserted a default-initialized entry into the cache *before* invoking the CUDA query, so a thrown `cuda_try` would have left zero-init data behind for any subsequent lookup. Now the result is computed into a stack-local first and assigned into the cache only on success.
- **`compute_kernel_limits`** — `cudaFuncGetAttributes` switched to the templated `cuda_try<F>(args...)` form so the result can be `const`-initialized.

### `memory.cuh`

- **`allocateHostMemory` / `allocateManagedMemory` pool wipes** — rewrote the "wipe pool when too full" `for`-loops as pop-then-free `while`-loops. With the naive substitution, a `cudaFreeHost` / `cudaFree` mid-loop failure would have left the pool full of pointers that were already freed, and the next call would have re-entered the wipe and double-freed them. The new pattern leaks at most the in-flight pointer (already removed from the pool), never causes a double-free.
- **`allocateHostMemory` / `allocateManagedMemory`** — `cudaMallocHost` / `cudaMallocManaged` switched to templated form, fused with the return.
- **Three callback launchers** (`cudaLaunchHostFunc` × 2 and `cudaGraphAddHostNode`) — each previously did `new std::pair<size_t, void*>(...)` and passed the raw pointer to the launch. If the launch returned an error, `cuda_safe_call` aborted; if we'd done a naive `cuda_try` conversion, the launch failure would have leaked the heap pair forever. Rewrote with `std::make_unique` + `args.release()` *after* the `cuda_try` succeeds, so a failed launch unwinds cleanly.
- **`address_is_pinned`** — `cudaPointerGetAttributes` → templated form.
- **`unpin_memory`** — `cudaGetLastError` → direct `cuda_try`.

### `constant_logical_data.cuh`

- **`cached_value::get()`** — between the `cached.emplace(...)` and the `cudaStreamSynchronize`, added `SCOPE(fail) { cached.erase(it); };`. If the sync throws, the partially-fetched cache entry would otherwise persist and corrupt any subsequent `get()` call (which would find the cache entry and return the unsynchronized value). `unordered_map::erase` by iterator is `noexcept`, so the SCOPE body is safe.

## Includes added

- `memory.cuh`: `<memory>` (for `std::make_unique`) and `utility/scope_guard.cuh`.
- `constant_logical_data.cuh`: `utility/scope_guard.cuh`.

(Both `scope_guard.cuh` includes are explicit rather than transitive, matching the style established in PR #9147.)

## Test plan

- [ ] CI green (`/ok to test` in comment below)
- [ ] No behavioral change on the success path


cudax/include/cuda/experimental/__stf/utility/memory.cuh
```diff
@@ -94,26 +100,28 @@ inline void* allocateHostMemory(size_t sz)
...
+      const auto it     = pool.begin();
+      void* const entry = it->second;
+      pool.erase(it);
+      cuda_try(cudaFree(entry));
     }
-    pool.clear();
   }
-  cuda_safe_call(cudaMallocManaged(&result, sz));
-  return result;
+  return cuda_try<cudaMallocManaged>(sz);
```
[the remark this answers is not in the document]
andralex: Fixed in 4f3a72acb9 — `allocateManagedMemory` now calls `cuda_try(cudaMallocManaged(&p, sz, cudaMemAttachGlobal))` so the attachment flags are explicit.

cudax/include/cuda/experimental/__stf/utility/memory.cuh
```diff
@@ -65,25 +67,35 @@ enum : size_t
...
+    // cudaFreeHost so that a thrown cuda_try leaks at most the in-flight
+    // pointer (already removed from the pool, so no double-free risk on
+    // the next call).
+    while (!pool.empty())
     {
-      cuda_safe_call(cudaFreeHost(entry.second));
+      const auto it     = pool.begin();
+      void* const entry = it->second;
+      pool.erase(it);
+      cuda_try(cudaFreeHost(entry));
```
caugonnet: not `cuda_try<cudaFreeHost>(entry) `?

cudax/include/cuda/experimental/__stf/utility/memory.cuh
```diff
@@ -65,25 +67,35 @@ enum : size_t
...
   }
-  cuda_safe_call(cudaMallocHost(&result, sz));
-  return result;
+  // Note: cannot use the templated ``cuda_try<cudaMallocHost>(sz)`` form
+  // here -- ``cudaMallocHost`` is a C++ overload set in cuda_runtime.h
+  // (templated wrapper around the C API), so ``decltype(fun)`` cannot
+  // resolve it. Use the runtime-status form instead.
+  void* p = nullptr;
+  cuda_try(cudaMallocHost(&p, sz));
+  return p;
```
caugonnet, line 98: return `cuda_try<cudaMallocHost>(sz)` ?

cudax/include/cuda/experimental/__stf/utility/memory.cuh
```diff
@@ -198,14 +222,17 @@ inline void deallocateHostMemory(void* p, size_t sz, cudaStream_t stream)
  */
 inline void deallocateManagedMemory(void* p, size_t sz, cudaStream_t stream)
 {
-  cuda_safe_call(cudaLaunchHostFunc(
+  // Same pattern as deallocateHostMemory(stream-ordered overload).
+  auto args = ::std::make_unique<::std::pair<size_t, void*>>(sz, p);
```
caugonnet, line 261: I'm not sure about these ones ! The arg is deleted in the callback with a delete. Maybe it's good but it is suspicious

cudax/include/cuda/experimental/__stf/utility/memory.cuh
andralex: good point, done

cudax/include/cuda/experimental/__stf/utility/memory.cuh
andralex: As we discussed offline, `cudaMallocHost` is a template that overloads a function and therefore its name alone is ambiguous. Artifices such as `cuda_try<cudaMallocHost<float>>` can be sometimes done but they're more trouble than it's worth.

cudax/include/cuda/experimental/__stf/utility/memory.cuh
andralex: Thanks for bringing this point up. This PR preserves existing semantics with the proviso that it throws instead of terminating, which is a more difficult task (you don't care about deallocating memory when aborting).

Previous code:
- allocate a pair with `new`
- pass a pointer to it to the callback
- if the callback fails to launch the program is terminated
- if the callback works it will delete the pair, thus cleaning things up

New code:
- allocate via a `unique_ptr`
- launch the callback
- if launching the callback fails, the `unique_ptr` deletes the memory on the way out
- if launching the callback succeeds, `unique_ptr::release` relinquishes ownership and the callback will delete the pair

caugonnet: Would not relying on ctx.add_resource simpler ? While idiomatic and safer, this adds a lot of noise in the code ...

This is what is done in parallel_for for instance 

```
    // For graph contexts, use deferred cleanup via ctx_resource (needed for graph replay)
    // For stream contexts, delete immediately in callback (better memory efficiency)
    if constexpr (::std::is_same_v<context, graph_ctx>)
    {
      auto resource = ::std::make_shared<parallel_for_args_resource<args_t>>(args);
      ctx.add_resource(mv(resource));
    }

    // The function which the host callback will execute
    auto host_func = [](void* untyped_args) {
      auto p = static_cast<decltype(args)>(untyped_args);
      
      auto& data               = ::std::get<0>(*p);
      const size_t n           = ::std::get<1>(*p);
      Fun& f                   = ::std::get<2>(*p);
      const sub_shape_t& shape = ::std::get<3>(*p);
    
      // deps_ops_t are pairs of data instance type, and a reduction operator,
      // this gets only the data instance types (eg. slice<double>)
      auto explode_coords = [&](size_t i, auto&&... data) {
        auto h = [&](auto&&... coords) {
          f(::std::forward<decltype(coords)>(coords)..., ::std::forward<decltype(data)>(data)...);
        };
        ::std::apply(h, shape.index_to_coords(i));
      };
  
      // Finally we get to do the workload on every 1D item of the shape
      for (size_t i = 0; i < n; ++i)
      {
        ::std::apply(explode_coords, ::std::tuple_cat(::std::make_tuple(i), data));
      }

      // For stream contexts, delete immediately (no replay risk)
      // For graph contexts, resource system handles cleanup (avoid use-after-free on replay)
      if constexpr (!::std::is_same_v<context, graph_ctx>)
      {
        delete p;
      }
    };

```

cudax/include/cuda/experimental/__stf/utility/memory.cuh
```diff
@@ -198,14 +222,17 @@ inline void deallocateHostMemory(void* p, size_t sz, cudaStream_t stream)
  */
 inline void deallocateManagedMemory(void* p, size_t sz, cudaStream_t stream)
 {
-  cuda_safe_call(cudaLaunchHostFunc(
+  // Same pattern as deallocateHostMemory(stream-ordered overload).
```
caugonnet: (unnecessary comment)

andralex: done

caugonnet (approved):

PR #9151, opened 2026-05-28
[STF] Use direct C++ linkage guards in C API header

caugonnet:
## Description

Simplify the STF C API header by using a local `#ifdef __cplusplus` / `extern "C"` wrapper directly in `stf.h`.

This avoids pulling in `cccl/c/extern_c.h` from the C parallel API solely for the linkage macros, so the STF C API does not need a new dependency on `cccl.c.parallel`.

## Testing

Not run; this is a header-only include/linkage cleanup.

### DIFF
--- a/c/experimental/stf/CMakeLists.txt
+++ b/c/experimental/stf/CMakeLists.txt
@@ -70,7 +70,11 @@ target_compile_options(
 
 target_include_directories(
   cccl.c.experimental.stf
-  PUBLIC "include"
+  PUBLIC #
+    "include"
+    # Pulls in cccl/c/extern_c.h, the canonical CCCL_C_EXTERN_C_BEGIN/_END macros
+    # shared with cccl.c.parallel.
+    "${CCCL_SOURCE_DIR}/c/parallel/include"
   PRIVATE "src"
 )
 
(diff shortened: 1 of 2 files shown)


miscco (approved):

c/experimental/stf/CMakeLists.txt
```diff
@@ -70,7 +70,11 @@ target_compile_options(
...
 target_include_directories(
   cccl.c.experimental.stf
-  PUBLIC "include"
+  PUBLIC #
+    "include"
+    # Pulls in cccl/c/extern_c.h, the canonical CCCL_C_EXTERN_C_BEGIN/_END macros
+    # shared with cccl.c.parallel.
+    "${CCCL_SOURCE_DIR}/c/parallel/include"
   PRIVATE "src"
 )
```
bernhardmgruber, line 75: This include seems a bit far fetched and creates a new dependency now, which we may need to encode in the CI scripts. @alliepiper for vis.

Maybe we should consider moving `<cccl/c/extern_c.h>` somewhere else instead.

caugonnet:
As per @bernhardmgruber suggestion, we need to make sure we have proper dependencies : there is currently no dependency between cccl_c_parallel and cccl_c_stf

```
  cccl_c_stf:
    name: "CCCL C CUDASTF Library"
    matrix_project: "cccl_c_stf"
    lite_dependencies: [libcudacxx_public, cudax_public, c2h]
    full_dependencies: []
    include_regexes:  ["c/experimental/stf/"]
```

caugonnet:
I have circled to a ridiculous PR so i'm closing :)

PR #9154, opened 2026-05-28
[cudax] Initial `cudax::coop::reduce` prototype

### LINKED ISSUE #9156: [FEA]: Initial `cudax::coop::reduce` prototype
Create a simplified implementation that takes `cudax::this_(thread|warp|block)` as the parameter.

davebayer:
This PR implements `cudax::coop::reduce` function prototype for `cudax::this_(thread|warp|block)`. The function dispatches to `cub::` equivalents for now and allocates all of the shared memory itself.

### DIFF
--- a/cudax/include/cuda/experimental/__coop/reduce.cuh
+++ b/cudax/include/cuda/experimental/__coop/reduce.cuh
@@ -1,0 +1,98 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of CUDA Experimental in CUDA C++ Core Libraries,
+// under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef _CUDA_EXPERIMENTAL___COOP_REDUCE_CUH
+#define _CUDA_EXPERIMENTAL___COOP_REDUCE_CUH
+
+#include <cuda/std/detail/__config>
+
+#if defined(_CCCL_IMPLICIT_SYSTEM_HEADER_GCC)
+#  pragma GCC system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_CLANG)
+#  pragma clang system_header
+#elif defined(_CCCL_IMPLICIT_SYSTEM_HEADER_MSVC)
+#  pragma system_header
+#endif // no system header
+
+#include <cub/block/block_reduce.cuh>
+#include <cub/thread/thread_reduce.cuh>
+#include <cub/warp/warp_reduce.cuh>
+
+#include <cuda/std/__cstddef/types.h>
+#include <cuda/std/__functional/operations.h>
+#include <cuda/std/optional>
+
+#include <cuda/experimental/group.cuh>
+
+#include <cuda/std/__cccl/prologue.h>
+
+#if !defined(_CCCL_DOXYGEN_INVOKED)
+
+namespace cuda::experimental::coop
+{
+template <class _Hierarchy, class _Tp, ::cuda::std::size_t _Np, class _RedFn>
+[[nodiscard]] _CCCL_DEVICE_API ::cuda::std::optional<_Tp>
+__reduce_impl(this_thread<_Hierarchy>, _Tp (&__thread_data)[_Np], _RedFn __red_fn)
+{
+  return ::cub::ThreadReduce(__thread_data, __red_fn);
+}
+
+template <class _Hierarchy, class _Tp, ::cuda::std::size_t _Np, class _RedFn>
+[[nodiscard]] _CCCL_DEVICE_API ::cuda::std::optional<_Tp>
+__reduce_impl(this_warp<_Hierarchy> __group, _Tp (&__thread_data)[_Np], _RedFn __red_fn)
+{
+  using _WarpReduce = ::cub::WarpReduce<_Tp>;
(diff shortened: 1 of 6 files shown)


cub/cub/warp/warp_reduce.cuh
```diff
@@ -504,7 +504,7 @@ public:
   [[nodiscard]] _CCCL_DEVICE _CCCL_FORCEINLINE T Reduce(const InputType& input, ReductionOp reduction_op)
   {
     auto thread_reduction = cub::ThreadReduce(input, reduction_op);
-    return WarpReduce<T, LogicalWarpThreads>::Reduce(thread_reduction, LogicalWarpThreads, reduction_op);
+    return WarpReduce<T, LogicalWarpThreads>::Reduce(thread_reduction, reduction_op);
```
miscco, line 507: Please move into separate PR and rebase once merged

cudax/cmake/cudaxBuildCompilerTargets.cmake
```diff
@@ -35,7 +35,7 @@ if ("Clang" STREQUAL "${CMAKE_CXX_COMPILER_ID}")
   append_option_if_available("-Wno-deprecated-copy" cxx_compile_options)
 endif()
 
-list(APPEND cxx_compile_definitions CCCL_ENABLE_ASSERTIONS)
+list(APPEND cxx_compile_definitions CCCL_ENABLE_ASSERTIONS _CUDAX_GROUP)
```
miscco: I would love a tad more descriptive macro like `_CUDAX_ENABLE_GROUP`

cudax/include/cuda/experimental/__coop/reduce.cuh
```diff
@@ -0,0 +1,98 @@
...
+
+template <class _Hierarchy, class _Tp, ::cuda::std::size_t _Np, class _RedFn>
+[[nodiscard]] _CCCL_DEVICE_API ::cuda::std::optional<_Tp>
+__reduce_impl(this_warp<_Hierarchy> __group, _Tp (&__thread_data)[_Np], _RedFn __red_fn)
+{
+  using _WarpReduce = ::cub::WarpReduce<_Tp>;
+  __shared__ typename _WarpReduce::TempStorage __scratch;
+
+  const auto __result = _WarpReduce{__scratch}.Reduce(__thread_data, __red_fn);
+  return (gpu_thread.is_root_rank(__group)) ? ::cuda::std::optional{__result} : ::cuda::std::nullopt;
```
miscco, line 55: I am confused by this line. Is this because the value is only valid in the leader thread? Should we broadcast it rather than diverging further?

Otherwise, why do we even compute it if its not desired

cudax/test/coop/reduce/this_thread.cu
```diff
@@ -0,0 +1,254 @@
...
+  template <class Config, int NumItems, class T, class RedOp>
+  __device__ void operator()(
+    Config config,
+    cuda::std::integral_constant<int, NumItems>,
+    const T* __restrict__ d_in,
+    T* __restrict__ d_out,
+    RedOp red_op)
+  {
+    T thread_data[NumItems];
+#pragma unroll
```
miscco: Shouldnt this just use the CCCL macros for that?

cudax/test/coop/reduce/this_thread.cu
```diff
@@ -0,0 +1,254 @@
...
+};
+
+template <>
+struct operator_to_std<cuda::maximum<>>
+{
+  using type = cuda::maximum<>;
+};
+
+template <class CudaT>
+using operator_to_std_t = typename operator_to_std<CudaT>::type;
```
miscco: Is this used anywhere?

cub/cub/warp/warp_reduce.cuh
davebayer: Implemented by #9153

cudax/cmake/cudaxBuildCompilerTargets.cmake
davebayer: Yeah, I thought the same thing

davebayer: I extracted this to a separate PR #9180

cudax/include/cuda/experimental/__coop/reduce.cuh
davebayer: This is because only the root rank has the correct value. CUB does the same thing, but returns garbage for non-root ranks. I think returning optional is the better way.

We've already discussed this before and we agreed that we would start like this and add an API that would also broadcast the result in the future.

cudax/test/coop/reduce/this_thread.cu
davebayer: I'll just get rid of the pragma completely, it's irrelevant for the test

cudax/test/coop/reduce/this_thread.cu
```diff
@@ -0,0 +1,203 @@
...
+constexpr int max_size  = 16;
+constexpr int num_seeds = 10;
+
+/***********************************************************************************************************************
+ * Test cases
+ **********************************************************************************************************************/
+
+_CCCL_DIAG_SUPPRESS_MSVC(4244) // warning C4244: '=': conversion from 'int' to '_Tp', possible loss of data
+
+C2H_TEST("reduce/this_thread Integral Type Tests", "[reduce][this_thread]", integral_type_list, operator_integral_list)
```
pciolkosz, line 165: The tests look very similar across the hierarchy levels, could it be just a template? The test body is mostly the same, the dispatch is almost the same

pciolkosz (approved):

cudax/test/coop/reduce/this_thread.cu
davebayer: It's simplified a bit for now, I will need to extend the tests in the future to be more specific for each level, so I'd rather duplicate the code for now

PR #9161, opened 2026-05-28
[STF] Add C API context wait helper

caugonnet:
Expose a blocking readback helper so C callers can observe logical data without finalizing the context.

## Description

closes 

## Checklist

### DIFF
--- a/c/experimental/stf/src/stf.cu
+++ b/c/experimental/stf/src/stf.cu
@@ -11,6 +11,7 @@
 #include <cuda/experimental/places.cuh>
 #include <cuda/experimental/stf.cuh>
 
+#include <algorithm>
 #include <cstddef>
 #include <cstdio>
 #include <cstring>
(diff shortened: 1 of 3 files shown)


andralex (approved):
lgtm

c/experimental/stf/src/stf.cu
```diff
@@ -433,6 +434,40 @@ cudaStream_t stf_fence(stf_ctx_handle ctx)
...
+    builder.add_deps(task_dep_untyped(*ld_ptr, access_mode::read));
+    builder.set_symbol("wait");
+    builder->*[dst, cap](reserved::host_launch_deps& deps) {
+      auto data      = deps.get<slice<char>>(0);
+      size_t copy_sz = ::std::min(cap, static_cast<size_t>(data.extent(0)));
+      ::std::memcpy(dst, data.data_handle(), copy_sz);
+    };
+
+    cudaStream_t fence_stream = context_ptr->fence();
+    cuda_safe_call(cudaStreamSynchronize(fence_stream));
```
NaderAlAwar, line 474: Suggestion: this aborts on error. It would probably make more sense to return an error since the return type of htis function is `int`.

c/experimental/stf/src/stf.cu
```diff
@@ -433,6 +434,40 @@ cudaStream_t stf_fence(stf_ctx_handle ctx)
...
+    void* dst  = out;
+    size_t cap = size;
+
+    auto builder = context_ptr->host_launch();
+    builder.add_deps(task_dep_untyped(*ld_ptr, access_mode::read));
+    builder.set_symbol("wait");
+    builder->*[dst, cap](reserved::host_launch_deps& deps) {
+      auto data      = deps.get<slice<char>>(0);
+      size_t copy_sz = ::std::min(cap, static_cast<size_t>(data.extent(0)));
+      ::std::memcpy(dst, data.data_handle(), copy_sz);
```
NaderAlAwar, line 470: Important: if `dst` and `data.data_handle()` overlap, this is UB. Consider guarding this call

caugonnet: That would not happen, but yeah let's add some assertion

c/experimental/stf/src/stf.cu
caugonnet: @andralex cuda_try(cudaStreamSynchronize(fence_stream) would put some value we can use in the catch block ?

caugonnet: @NaderAlAwar while that's true, returning a cudaError_t would probably create some precedent compared to other methods in our C API. This would probably be part of a bigger PR which would replace many _CCCL_ASSERT by proper error checking with "invalid value" returned for example.

c/experimental/stf/src/stf.cu
caugonnet: done

NaderAlAwar (approved):

PR #9162, opened 2026-05-28
[STF] Add extended C context creation API

caugonnet:
Expose stream/backend/resource options for C STF contexts so callers can reuse async resources and bind user streams explicitly.

## Description

closes 

## Checklist


c/experimental/stf/test/test_stream_ctx_override.cu
```diff
@@ -0,0 +1,307 @@
...
+  }
+  int acc = 0;
+  // Busy loop to keep the kernel resident on the SM for a while.
+  for (int i = 0; i < iters; ++i)
+  {
+    acc += (i * 1103515245 + 12345) & 0x7fffffff;
+  }
+  // Side-effect: ensure the compiler does not elide the loop, then commit
+  // the intended `value` (independent of acc) to the buffer.
+  arr[tid] = value + (acc & 0);
```
NaderAlAwar: Important: `acc & 0` will always return 0, so the above loop can be optimized away

c/experimental/stf/include/cccl/c/experimental/stf/stf.h
```diff
@@ -345,6 +345,101 @@ stf_ctx_handle stf_ctx_create(void);
...
+
+typedef struct stf_ctx_options
+{
+  stf_backend_kind backend; //!< Backend selector (default: STF_BACKEND_STREAM)
+  int has_stream; //!< 0: no caller stream; non-zero: inherit `stream`
+  cudaStream_t stream; //!< Caller-owned stream (used iff `has_stream != 0`).
+                       //!< `cudaStream_t` is a pointer; `nullptr` is the NULL stream,
+                       //!< not a sentinel -- use `has_stream` to say "no stream".
+  stf_async_resources_handle handle; //!< Shared resources handle, or NULL for "create fresh"
+} stf_ctx_options;
```
NaderAlAwar, line 427: Question: if you append stuff here it would break ABI. Is that an intentional choice?

c/experimental/stf/include/cccl/c/experimental/stf/stf.h
```diff
@@ -345,6 +345,101 @@ stf_ctx_handle stf_ctx_create(void);
...
+//! for (int i = 0; i < N; ++i) {
+//!   stf_ctx_options opts = {0};
+//!   opts.backend    = STF_BACKEND_GRAPH;
+//!   opts.has_stream = 1;           // opt-in to caller stream binding
+//!   opts.stream     = user_stream; // may be 0 for the default/NULL stream
+//!   opts.handle     = h;
+//!   stf_ctx_handle ctx = stf_ctx_create_ex(&opts);
+//!   // ... submit tasks ...
+//!   stf_ctx_finalize(ctx);
+//! }
```
NaderAlAwar, line 458: Question: my understanding is that contexts created with `has_stream = 1` make `stf_ctx_finalize(ctx)` non-blocking, so should this example call `cudaStreamSynchronize(user_stream)` after `stf_ctx_finalize(ctx)` before observing results or destroying the shared stf_async_resources_handle? Otherwise the example appears to rely on `stf_ctx_finalize` completing all queued work, which is not true for caller-owned stream contexts.

c/experimental/stf/include/cccl/c/experimental/stf/stf.h
caugonnet: Are we supposed to have some ABI stability across versions ?

c/experimental/stf/include/cccl/c/experimental/stf/stf.h
caugonnet: That's a good observation, and i'm realizing the (blocking) sync is needed with all users of the async resource handle before it can be destructed. We have enforced it in all case, including in C++, but we need to document this (in C++ too)

c/experimental/stf/include/cccl/c/experimental/stf/stf.h
NaderAlAwar: Not necessarily, was just curious

c/experimental/stf/test/test_stream_ctx_override.cu
caugonnet: well spotted, fixed

andralex (approved):
lgtm

c/experimental/stf/include/cccl/c/experimental/stf/stf.h
caugonnet: I've documented this, here and in the async resource doc in general, thanks a lot @NaderAlAwar

NaderAlAwar (approved):

PR #9163, opened 2026-05-28
Only enabled PDL for PTX/SASS supporting it

### LINKED ISSUE #9134: [BUG] Unconditionally enabling PDL for any kernel is a bug
Programmatic Dependent Launch (PDL) enables a kernel to overlap into the previous one on the same stream when the second kernel is launched with PDL enabled (by launching with the `programmaticStreamSerializationAllowed` attribute). In order to make this save and properly ensure that data written by the first kernel is visible to the secondary one, the secondary kernel needs to execute `cudaGridDependencySynchronize`/`ACQBULK`, which blocks the kernel until the first kernel's `MEMBAR` has completed, and all writes are visible.

CCCL implements `ACQBULK` as:
```c++
#  define _CCCL_PDL_GRID_DEPENDENCY_SYNC() NV_IF_TARGET(NV_PROVIDES_SM_90, ::cudaGridDependencySynchronize();)
```
and for any kernel with the right synchronizations in place, we *unconditionally* enable PDL when launching the kernel.

The use of `cudaTriggerProgrammaticLaunchCompletion`/`PRE-EXIT` can allow the secondary kernel to overlap more, but is otherwise irrelevant here. What's important is that in the absence of `PRE-EXIT`, the execution of `EXIT` will trigger the launch of the next kernel, which is then followed by the execution of the first kernels `MEMBAR`.

Now consider a user compiling CCCL to PTX for `sm_80` (any arch before `sm_90`, which is the first arch that supports PDL). All PDL instructions are compiled out because they are guarded under `NV_IF_TARGET(NV_PROVIDES_SM_90, ...)`. And then running this binary on `sm_90`. The driver will JIT compile the `sm_80` PTX without ACQBULK to `sm_90` and the host code (which was *unconditionally* compiled with PDL enabled) will now launch kernels with PDL enabled. A secondary kernel will be overlapped into the first one, as triggered by the first kernel's `EXIT` instruction. However, because the secondary kernel does not execute `ACQBULK` , its execution overlaps with the first kernel's `MEMBAR` and global memory changes by the first kernel may not be globally visible yet. This may result in race conditions, when the secondary kernel reads data that was written by the previous kernel.

This issue was discovered in llama.cpp: https://github.com/ggml-org/llama.cpp/pull/22522#issuecomment-4505600876 and reported to me on Slack. Upon further discussion, we noticed that CCCL is also affected.

The PDL roll-out in CCCL is tracked by #3115 and PDL is currently only used by `DeviceFind`, `DeviceHistogram`, `DeviceMergeSort`, `DeviceScan`, and `DeviceTransform`. The fix is to disable PDL in the host side kernel launcher when the compute capability we compiled for is below 9.0.

bernhardmgruber:
The changes in this PR assert that PDL is only enabled with PTX/SASS supporting it. The assertions are not triggered by any CI run, but I can hit them locally by e.g. compiling for SM80 and running on SM120 (e.g. `cub.test.device.transform.lid_0`).

Then, a fix is proposed by guarding any PDL use by the compute capability of the PTX used to launch.

Fixes: #9134

### DIFF
--- a/cub/cub/detail/launcher/cuda_driver.cuh
+++ b/cub/cub/detail/launcher/cuda_driver.cuh
@@ -79,8 +79,9 @@ struct CudaDriverLauncherFactory
   {
     if (dependent_launch)
     {
+      // note: assumes that cc_ holds the current device's AND the PTX's compute capability
       _CCCL_ASSERT((::cuda::compute_capability{cc_} >= ::cuda::compute_capability{9, 0}),
-                   "Enabling PDL for a kernel launch requires SM90+ PTX/SASS");
+                   "Enabling PDL for a kernel launch requires CC 9.0+ PTX/SASS when running on SM90+");
     }
   }
 
--- a/cub/cub/detail/launcher/cuda_driver.cuh
+++ b/cub/cub/detail/launcher/cuda_driver.cuh
@@ -75,7 +75,7 @@ struct CudaDriverLauncher
 
 struct CudaDriverLauncherFactory
 {
-  CUB_RUNTIME_FUNCTION void __assert_pdl_allowed(bool dependent_launch) const
+  CUB_RUNTIME_FUNCTION inline void __assert_pdl_allowed(bool dependent_launch) const
   {
     if (dependent_launch)
     {
--- a/cub/cub/detail/launcher/cuda_driver.cuh
+++ b/cub/cub/detail/launcher/cuda_driver.cuh
@@ -75,9 +75,19 @@ struct CudaDriverLauncher
 
 struct CudaDriverLauncherFactory
 {
+  CUB_RUNTIME_FUNCTION void __assert_pdl_allowed(bool dependent_launch) const
+  {
+    if (dependent_launch)
+    {
+      _CCCL_ASSERT((::cuda::compute_capability{cc_} >= ::cuda::compute_capability{9, 0}),
+                   "Enabling PDL for a kernel launch requires SM90+ PTX/SASS");
+    }
+  }
+
   CudaDriverLauncher
   operator()(dim3 grid, dim3 block, unsigned int shared_mem, ::CUstream stream, bool dependent_launch = false) const
   {
+    __assert_pdl_allowed(dependent_launch);
     return CudaDriverLauncher{grid, block, shared_mem, stream, dependent_launch};
   }
 
--- a/cub/cub/detail/launcher/cuda_driver.cuh
+++ b/cub/cub/detail/launcher/cuda_driver.cuh
@@ -75,7 +75,7 @@ struct CudaDriverLauncher
 
 struct CudaDriverLauncherFactory
 {
-  CUB_RUNTIME_FUNCTION inline void __assert_pdl_allowed(bool dependent_launch) const
+  CUB_RUNTIME_FUNCTION void __assert_pdl_allowed(bool dependent_launch) const
   {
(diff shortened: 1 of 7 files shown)


bernhardmgruber:
Let's see how far back we can backport this.

cub/cub/detail/launcher/cuda_driver.cuh
```diff
@@ -75,9 +75,20 @@ struct CudaDriverLauncher
 
 struct CudaDriverLauncherFactory
 {
+  CUB_RUNTIME_FUNCTION inline void __assert_pdl_allowed(bool dependent_launch) const
```
davebayer: Nitpick:
```suggestion
  CUB_RUNTIME_FUNCTION void __assert_pdl_allowed(bool dependent_launch) const
```

cub/cub/detail/launcher/cuda_runtime.cuh
```diff
@@ -25,9 +25,26 @@ namespace detail
 {
 struct TripleChevronFactory
 {
+  CUB_RUNTIME_FUNCTION inline void __assert_pdl_allowed(bool dependent_launch) const
```
davebayer:
```suggestion
  CUB_RUNTIME_FUNCTION void __assert_pdl_allowed(bool dependent_launch) const
```

cub/cub/device/dispatch/dispatch_find.cuh
```diff
@@ -190,17 +191,23 @@ CUB_RUNTIME_FUNCTION _CCCL_FORCEINLINE cudaError_t dispatch(
...
-  if (const auto error = CubDebug(THRUST_NS_QUALIFIER::cuda_cub::detail::triple_chevron(
-                                    findif_grid_size, active_policy.threads_per_block, 0, stream, true)
-                                    .doit(kernel_ptr, d_in_unwrapped, num_items, found_pos_ptr, predicate)))
+  if (const auto error = CubDebug(
+        THRUST_NS_QUALIFIER::cuda_cub::detail::triple_chevron(
+          findif_grid_size,
+          active_policy.threads_per_block,
+          0,
+          stream,
+          /* dependent_launch */ cc >= ::cuda::compute_capability{9, 0})
```
davebayer, line 200: Can we rather introduce:
```cpp
enum class enable_dependent_launch : bool {};
```
and do:

```suggestion
          enable_dependent_launch{cc >= ::cuda::compute_capability{9, 0}})
```
I hate that we need the comment

cub/cub/detail/launcher/cuda_driver.cuh
bernhardmgruber: I added it because CodeRabbit told me to, because @fbusato put that into the coding guidelines:

> Per coding guidelines, non-template, non-constexpr functions must use the inline keyword. This function qualifies since it's marked with CUB_RUNTIME_FUNCTION but is neither a template nor constexpr.

I think we should exclude member functions here.

cub/cub/device/dispatch/dispatch_find.cuh
bernhardmgruber: I just tried that and found it a bit cumbersome. Where should we put the definition of this enum? It's needed in Thrust and CUB.

davebayer (approved):

bernhardmgruber:
Manual backport to 3.3: #9188

PR #9168, opened 2026-05-28
Refactor warpspeed scan 1/2

bernhardmgruber:
This is the first part of a few cleanup commits that should improve the readability of the warpspeed scan implementation.

AI generated summary:
                                                                                                                                    
  - Replaced the large kernelBody free function with an agent_warpspeed_scan struct that holds the kernel parameters as member data
  - Broke the monolithic function body into focused member functions:                                                                                                                                                
    - load_next_tile_index — loads the next tile index into shared memory (scheduler squad)
    - load_current_tile — bulk-loads the current tile from global to shared memory (load squad)                                                                                                                      
    - lookback — performs the decoupled lookback for prefix propagation (lookback squad)                                                                                                                             
    - reduce_tile — loads tile from smem, reduces across threads/warps/squad, stores aggregates (reduce squad)                                                                                                       
    - scan_and_store_tile — performs the inclusive/exclusive scan and stores results back to global memory (scan-store squad)                                                                                        
    - run — main loop orchestrating the pipeline across squads                                                                                                                                                       
  - Eliminated the scan_and_store lambda (which couldn't capture structured bindings in C++17) in favor of a proper template member function with IsLastTile as a template parameter                                                                                                                                                                                                                                                                                           
  - smem_resource.cuh: Renamed popStage() to nextStage() for clarity                                                                                                                                                 



I tried pedantically to not cause any SASS changes, but extracting the `reduce_tile` function somehow let the compiler elide an `ISETP.GE.U64.AND P3, PT, R14, 0xf80, PT ;` instruction, so the kernel is now exactly one instruction shorter. A few register names changed as well, but the instructions are otherwise identical.


cub/cub/detail/warpspeed/squad/squad_desc.cuh
```diff
@@ -47,14 +47,12 @@ struct SquadDesc
     return 32 * warpCount();
   }
 
-  [[nodiscard]] _CCCL_HOST_DEVICE_API friend constexpr bool
-  operator==(const SquadDesc& lhs, const SquadDesc& rhs) noexcept
```
bernhardmgruber, line 51: This is needed to avoid ODR use of squad descriptions which are `static constexpr` members of a class, since otherwise we get the error that those are not accessible in `__device__` code.

cub/cub/detail/warpspeed/squad/squad.cuh
```diff
@@ -69,10 +69,10 @@ struct Squad : SquadDesc
...
   }
 
-  _CCCL_DEVICE_API void syncThreads()
+  _CCCL_DEVICE_API void syncThreads() const
   {
     // barrier 0 is reserved for __syncthreads(). We use barrier ids 1, ...
-    int barrierIdx = (int) this->mSquadIdx + 1;
+    const int barrierIdx = this->mSquadIdx + 1;
 
     __barrier_sync_count(barrierIdx, this->threadCount());
```
davebayer, line 77: Unrelated to the PR: Why do we use unaligned version of the `__barrier_sync`? From what I see in the main loop, we always synchronize the warps on the same line in the code, so we should be able to use the aligned version, which produces a bit less code.

See the comparison: https://godbolt.org/z/vPGbT9W7c

cub/cub/device/dispatch/kernels/kernel_scan_warpspeed.cuh
```diff
@@ -303,491 +286,549 @@ _CCCL_DEVICE_API _CCCL_FORCEINLINE void kernelBody(
...
+      else
+      {
+        regThreadSum = CUB_NS_QUALIFIER::ThreadReduce(regInput, scan_op);
+        regWarpSum   = __scan_detail::warpReduce(regThreadSum, scan_op);
+      }
     }
 
     ////////////////////////////////////////////////////////////////////////////////
-    // Get next tile index from shared memory (all squads)
+    // Store warp sump to shared memory
```
davebayer:
```suggestion
    // Store warp sum to shared memory
```

cub/cub/detail/warpspeed/squad/squad.cuh
davebayer: Ah, I'm probably wrong, it's not that all threads that are part of the barrier must call the instruction uniformly, but rather the whole CTA :(

cub/cub/device/dispatch/kernels/kernel_scan_warpspeed.cuh
```diff
@@ -303,491 +286,549 @@ _CCCL_DEVICE_API _CCCL_FORCEINLINE void kernelBody(
...
-    warpspeed::CpAsyncOobInfo loadInfo =
-      warpspeed::prepareCpAsyncOob(const_cast<InputT*>(params.ptrIn) + idxTileBase, valid_items);
+  template <typename PhaseInOutT, typename PhaseSumT>
+  _CCCL_DEVICE_API _CCCL_FORCEINLINE void reduce_tile(
+    const warpspeed::Squad& squad,
+    PhaseInOutT& phaseInOutRW,
+    PhaseSumT& phaseSumThreadAndWarpW,
+    int valid_items,
+    bool is_first_tile,
+    bool is_last_tile, // TODO(bgruber): should we dispatch on is_last_tile outside this function and compile it twice?
```
miscco, line 343: I believe in the long run we should have an enumeration like
```cpp
enum class good_name{
  __full_tile,
  __first_tile,
  __last_tile,
  __only_tile,
};
```

cub/cub/device/dispatch/kernels/kernel_scan_warpspeed.cuh
```diff
@@ -303,491 +286,549 @@ _CCCL_DEVICE_API _CCCL_FORCEINLINE void kernelBody(
...
-    const bool is_last_tile = valid_items < tile_size;
-    warpspeed::CpAsyncOobInfo loadInfo =
-      warpspeed::prepareCpAsyncOob(const_cast<InputT*>(params.ptrIn) + idxTileBase, valid_items);
+  template <typename PhaseInOutT, typename PhaseSumT>
+  _CCCL_DEVICE_API _CCCL_FORCEINLINE void reduce_tile(
+    const warpspeed::Squad& squad,
+    PhaseInOutT& phaseInOutRW,
+    PhaseSumT& phaseSumThreadAndWarpW,
+    int valid_items,
+    bool is_first_tile,
```
miscco, line 342: Should those rather be template arguments?

cub/cub/device/dispatch/kernels/kernel_scan_warpspeed.cuh
bernhardmgruber: Possibly!

cub/cub/device/dispatch/kernels/kernel_scan_warpspeed.cuh
bernhardmgruber: This is exactly what the next line implies:

> // TODO(bgruber): should we dispatch on is_last_tile outside this function and compile it twice?

I believe we may want to test this, but not in this PR.

miscco (approved):

